PR #4, opened 2016-08-07
Miscellaneous fixes for ArchLinux

L3n41c:
* Make ArchLinux image bootable
* Add the possibility to customize the ArchLinux packages to be installed
* Fix ArchLinux image build on non-ArchLinux host

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -6,7 +6,6 @@ import argparse, sys, os, subprocess, uuid, platform, tempfile, shutil, time, ha
 # TODO
 # - squashfs root
 # - volatile images
-# - fix archlinux
 # - make debian/ubuntu images bootable
 # - work on device nodes
 # - allow passing env vars
@@ -467,12 +466,22 @@ def install_arch(args, workspace, run_build_script):
     subprocess.run(["pacman-key", "--init"], check=True)
     subprocess.run(["pacman-key", "--populate", "archlinux"], check=True)
 
-    cmdline = ["pacstrap",
-               "-c",
-               "-d",
-               workspace + "/root"]
+    with open(os.path.join(workspace, "pacman.conf"), "w") as f:
+        f.write("[options]\n")
+        f.write("HookDir     = /no_hook/\n")
+        f.write("HoldPkg     = pacman glibc\n")
+        f.write("Architecture = auto\n")
+        f.write("CheckSpace\n")
+        f.write("SigLevel    = Required DatabaseOptional\n")
+        f.write("\n")
+        f.write("[core]\n")
+        f.write("Server = https://mirrors.kernel.org/archlinux/$repo/os/$arch\n")
+        f.write("\n")
+        f.write("[extra]\n")
+        f.write("Server = https://mirrors.kernel.org/archlinux/$repo/os/$arch\n")
 
-    c = subprocess.run(["pacman", "-Sg", "base"], stdout=subprocess.PIPE, universal_newlines=True, check=True)
+    subprocess.run(["pacman", "--config", os.path.join(workspace, "pacman.conf"), "-Sy"], check=True)
+    c = subprocess.run(["pacman", "--config", os.path.join(workspace, "pacman.conf"), "-Sg", "base"], stdout=subprocess.PIPE, universal_newlines=True, check=True)
     packages = set(c.stdout.split())
     packages.remove("base")
 
@@ -502,9 +511,12 @@ def install_arch(args, workspace, run_build_script):
     if run_build_script and args.build_packages is not None:
         packages |= set(args.build_packages)
 
-    cmdline.extend(list(packages))
-
-    cmdline.extend(["--hookdir", "/no_hook"])
+    cmdline = ["pacstrap",
+               "-C", os.path.join(workspace, "pacman.conf"),
+               "-c",
+               "-d",
+               workspace + "/root"] + \
+               list(packages)
 
     subprocess.run(cmdline, check=True)
 
--- a/mkosi
+++ b/mkosi
@@ -470,9 +470,41 @@ def install_arch(args, workspace, run_build_script):
     cmdline = ["pacstrap",
                "-c",
                "-d",
-               "-i",
-               workspace + "/root",
-               "base"]
+               workspace + "/root"]
+
+    c = subprocess.run(["pacman", "-Sg", "base"], stdout=subprocess.PIPE, universal_newlines=True, check=True)
+    packages = set(c.stdout.split())
+    packages.remove("base")
+
+    packages -= {"cryptsetup",
+                 "device-mapper",
+                 "dhcpcd",
+                 "e2fsprogs",
+                 "jfsutils",
+                 "lvm2",
+                 "mdadm",
+                 "netctl",
+                 "pcmciautils",
+                 "reiserfsprogs",
+                 "xfsprogs"}
+
+    if args.bootable:
+        if args.output_format == OutputFormat.raw_gpt:
+            packages.add("e2fsprogs")
+        elif args.output_format == OutputFormat.raw_btrfs:
+            packages.add("btrfs-progs")
+    else:
+        packages.remove("linux")
+
+    if args.packages is not None:
+        packages |= set(args.packages)
+
+    if run_build_script and args.build_packages is not None:
+        packages |= set(args.build_packages)
+
+    cmdline.extend(list(packages))
+
+    cmdline.extend(["--hookdir", "/no_hook"])
 
     subprocess.run(cmdline, check=True)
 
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -107,6 +107,7 @@ def create_image(args, workspace):
     print_step("Creating partition table...")
 
     f = tempfile.NamedTemporaryFile(dir = os.path.dirname(args.output), prefix='.mkosi-')
+    subprocess.run(["chattr", "+C", f.name], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, check=True)
     f.truncate(image_size(args))
```
poettering: makes sense to add. And the patch is fine I figure, though I personally would always have tried to do the trivial ioctl this issues directly from python instead of shelling out to chattr...

mkosi
```diff
@@ -469,6 +469,40 @@ def install_distribution(args, workspace, run_build_script):
...
+                    os.path.join(workspace, "root", "boot/efi/initramfs-linux.img"))
+
+    with open(os.path.join(workspace, "root", "boot/efi/loader/loader.conf"), "w") as f:
+        f.write("default arch\n")
+
+    with open(os.path.join(workspace, "root", "boot/efi/loader/entries/arch.conf"), "w") as f:
+        f.write("title   Arch Linux\n")
+        f.write("linux   /vmlinuz-linux\n")
+        f.write("initrd  /initramfs-linux.img\n")
+        f.write("options quiet rw\n")
```
poettering: I really wished arch could just use kernel-install so that we don't have to write things manually like this...

mkosi
```diff
@@ -469,6 +469,40 @@ def install_distribution(args, workspace, run_build_script):
...
+    subprocess.run(["arch-chroot", os.path.join(workspace, "root"), "mkinitcpio", "-p", "linux"], check=True)
+
+    shutil.copyfile(os.path.join(workspace, "root", "boot/vmlinuz-linux"),
+                    os.path.join(workspace, "root", "boot/efi/vmlinuz-linux"))
+
+    shutil.copyfile(os.path.join(workspace, "root", "boot/initramfs-linux.img"),
+                    os.path.join(workspace, "root", "boot/efi/initramfs-linux.img"))
+
+    with open(os.path.join(workspace, "root", "boot/efi/loader/loader.conf"), "w") as f:
+        f.write("default arch\n")
```
poettering: this is redundant, no? I mean, why set a default if we only have a single entry anyway...

mkosi
```diff
@@ -457,6 +457,34 @@ def install_arch(args, workspace, run_build_script):
...
+
+    if args.bootable:
+        with open(os.path.join(workspace, "root", "etc/systemd/network/wired.network"), "w") as f:
+            f.write("[Match]\n")
+            f.write("Name=en*\n")
+            f.write("\n")
+            f.write("[Network]\n")
+            f.write("DHCP=yes\n")
+
     print_step("Installing ArchLinux complete.")
```
poettering: Not sure how I feel about this patch. The goal with mkosi was actualy to change as little as possible from the actual image, instead of patching around too widely.

here's an idea: maybe we should start shipping something like wired.network by default in systemd, but ask downstream distros to split it out in separate packages? That way it would be a simple matter of installing that additional package to enable automatic networkd coverage of everything ethernet. And in the case of mkosi we'd add that to the default packet set? Would that work for ArchLinux?

poettering: i added an RFE for this idea to https://github.com/systemd/systemd/issues/3998 now.

mkosi
```diff
@@ -457,6 +457,34 @@ def install_arch(args, workspace, run_build_script):
 
     subprocess.run(cmdline, check=True)
 
+    subprocess.run(["arch-chroot", os.path.join(workspace, "root"), "systemctl", "enable", "systemd-networkd"], check=True)
+    subprocess.run(["arch-chroot", os.path.join(workspace, "root"), "systemctl", "enable", "systemd-resolved"], check=True)
```
poettering: hmpf, I don't like the mixing of different chrooting tools (i.e. nspawn vs arch-chroot), and invoking code from inside the image too much.

I'd feel a lot more comfortable if this would use systemctl from the host with its --root= switch.

poettering: also, please do this in one line. "systemct enable" takes as many arguments as you like.

mkosi
```diff
@@ -457,6 +457,34 @@ def install_arch(args, workspace, run_build_script):
...
+        with open(nsswitch_new, "w") as new:
+            for line in old:
+                if line.startswith("hosts:"):
+                    new.write(" ".join(["resolve" if x == "dns" else x for x in line.split(" ")]))
+                else:
+                    new.write(line)
+
+    shutil.copystat(nsswitch_old, nsswitch_new)
+    os.remove(nsswitch_old)
+    shutil.move(nsswitch_new, nsswitch_old)
```
poettering: i really don't like patching of config files like this I must say, but I have no better idea either... But could you please move this into a separate function, so that we can call it it also for the other distros that need this (specifically: fedora). Same for the resolv.conf handling.

mkosi
poettering: for now it would be OK to add this with the code you proposed as long as the systemd RFE is not implemented. But please use Type=ether as match, and make this a separate function, so that we can reuse this for the fedora case.

poettering:
the rest looks good to me.

Thanks for working on this! Much appreciated!

mkosi
lhuard1A: I wish I could find a function to change file attributes in the python standard library.
Unfortunately, without that, invoking `chattr` was the simplest way.

mkosi
lhuard1A: Fixed.
I’m now using `kernel-install` to populate the ESP.

mkosi
lhuard1A: Fixed.

mkosi
lhuard1A: I’ve moved inside an `enable_networkd(…)` function everything which is related to networkd activation on ArchLinux:
* Enable `systemd-networkd` and `systemd-resolved` units;
* Replace `dns` by `resolve` in `nsswitch.conf`;
* Symlink `/etc/resolv.conf`.

I don’t know if it would make sense to split this function further because not all steps are needed on all distributions.

mkosi
lhuard1A: I’ve put the creation of the `*.network` config file in the same function as the one that enables networkd (`enable_networkd(…)`).
Would you need to have a dedicated one instead?

Also, for some reasons, using `Type=ether` instead of `Name=en*` doesn’t work on ArchLinux. The interface remains unconfigured:

```
[root@localhost ~]# networkctl list
IDX LINK             TYPE               OPERATIONAL SETUP     
  1 lo               loopback           carrier     unmanaged 
  2 ens3             ether              off         unmanaged 

2 links listed.
```

According to `systemd.network(5)` man page, the `Type=` value has to match the `DEVTYPE` udev property, but I have the feeling that this property is not set:

```
[root@localhost ~]# udevadm info /sys/devices/pci0000\:00/0000\:00\:03.0/net/ens3
P: /devices/pci0000:00/0000:00:03.0/net/ens3
E: DEVPATH=/devices/pci0000:00/0000:00:03.0/net/ens3
E: ID_BUS=pci
E: ID_MODEL_FROM_DATABASE=82540EM Gigabit Ethernet Controller (QEMU Virtual Machine)
E: ID_MODEL_ID=0x100e
E: ID_NET_DRIVER=e1000
E: ID_NET_LINK_FILE=/usr/lib/systemd/network/99-default.link
E: ID_NET_NAME=ens3
E: ID_NET_NAME_MAC=enx525400123456
E: ID_NET_NAME_PATH=enp0s3
E: ID_NET_NAME_SLOT=ens3
E: ID_PATH=pci-0000:00:03.0
E: ID_PATH_TAG=pci-0000_00_03_0
E: ID_PCI_CLASS_FROM_DATABASE=Network controller
E: ID_PCI_SUBCLASS_FROM_DATABASE=Ethernet controller
E: ID_VENDOR_FROM_DATABASE=Intel Corporation
E: ID_VENDOR_ID=0x8086
E: IFINDEX=2
E: INTERFACE=ens3
E: SUBSYSTEM=net
E: SYSTEMD_ALIAS=/sys/subsystem/net/devices/ens3 /sys/subsystem/net/devices/ens3
E: TAGS=:systemd:
E: USEC_INITIALIZED=2057125
```

The systemd version that I used was 231. I haven’t checked the udev rules yet to figure out whether this is a systemd regression or an arch-specific issue.

mkosi
poettering: well, python exposes ioctl() in the "fnctl" module afaik. All you have to do is invoke it on an fd, with the FS_IOC_SETFLAGS arg as first param, plus the bit field to set as second param...

PR #12, opened 2016-09-01
Mirror selection support and Fedora build improvements

davide125:
This PR includes three stacked changesets:
- generate a temporary dnf.conf (akin as what's done for arch) instead of relying on the system dnf config; this allows mkosi to build Fedora images on e.g. CentOS hosts
- add an option to specify a custom mirror (useful for us as we'd rather use our local one)
- if available, use a local copy of the Fedora gpg key instead of fetching it from upstream

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -394,8 +394,23 @@ def enable_networkd(workspace):
 def install_fedora(args, workspace, run_build_script):
     print_step("Installing Fedora...")
 
+    with open(os.path.join(workspace, "dnf.conf"), "w") as f:
+        f.write("[main]\n")
+        f.write("gpgcheck=1\n")
+        f.write("\n")
+        f.write("[fedora]\n")
+        f.write("name=Fedora %s - base\n" % args.release)
+        f.write("baseurl=https://mirrors.kernel.org/fedora/releases/%s/Everything/x86_64/os/\n" % args.release)
+        f.write("gpgkey=https://getfedora.org/static/81B46521.txt\n")
+        f.write("\n")
+        f.write("[updates]\n")
+        f.write("name=Fedora %s - updates\n" % args.release)
+        f.write("baseurl=https://mirrors.kernel.org/fedora/updates/%s/x86_64/\n" % args.release)
+        f.write("gpgkey=https://getfedora.org/static/81B46521.txt\n")
+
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -394,8 +394,23 @@ def enable_networkd(workspace):
...
+        f.write("\n")
+        f.write("[fedora]\n")
+        f.write("name=Fedora %s - base\n" % args.release)
+        f.write("baseurl=https://mirrors.kernel.org/fedora/releases/%s/Everything/x86_64/os/\n" % args.release)
+        f.write("gpgkey=https://getfedora.org/static/81B46521.txt\n")
+        f.write("\n")
+        f.write("[updates]\n")
+        f.write("name=Fedora %s - updates\n" % args.release)
+        f.write("baseurl=https://mirrors.kernel.org/fedora/updates/%s/x86_64/\n" % args.release)
+        f.write("gpgkey=https://getfedora.org/static/81B46521.txt\n")
```
poettering: btw, i think it would be nicer if we'd start using multine strings for this.

I can't say I like the hardcoding of the x86_64 thing, we should probably make it match the host arch... but then again I can't say I care too much for now, after all pretty much everything real is x86_64 these days...

mkosi
```diff
@@ -1246,6 +1268,16 @@ def load_args():
...
             args.release = "yakkety"
 
+    if args.mirror is None:
+        if args.distribution == Distribution.fedora:
+            args.mirror = "https://mirrors.kernel.org/fedora"
+        elif args.distribution == Distribution.debian:
+            args.mirror = "http://httpredir.debian.org/debian"
+        elif args.distribution == Distribution.ubuntu:
+            args.mirror = "http://archive.ubuntu.com/ubuntu"
+        elif Distribution.arch:
```
kovetskiy: I guess there should be `elif args.distribution == Distribution.arch`

davide125: sorry, I'll fix that

mkosi
davide125: Given that we abort early on if `platform.machine() != "x86_64"`, I think this is fine for now.

mkosi
davide125: #13 should take care of this

PR #16, opened 2016-09-15
Install dbus and libpam-systemd with debootstrap

### LINKED ISSUE #15: Debian/Ubuntu: dbus not installed
`debootstrap` does not install `dbus`, probably because it only installs required dependencies of the `--include`d packages (`systemd` recommends `dbus`, but doesn’t require it). This means that the new image can’t be controlled with `machinectl shell` or `machinectl login`.

lucaswerkmeister:
Both packages are recommended by the Debian systemd package, but debootstrap does not install Recommends.

Installing dbus fixes #15; the addition of libpam-systemd was suggested by @stefan-it in that same issue.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -454,7 +454,7 @@ def install_debian_or_ubuntu(args, workspace, run_build_script, mirror):
     cmdline = ["debootstrap",
                "--verbose",
                "--variant=minbase",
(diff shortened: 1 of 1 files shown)


lucaswerkmeister:
Rebased onto #17.

PR #17, opened 2016-09-22
Do not install usrmerge on Debian/Ubuntu images

fsateler:
This reverts commit a7f5575b1a49cee812e3b4f6f105b2d8ce111bee

Turns out debootstrap now has native support for merged /usr, and will likely be default (for the supported targets) in the near future. 

Thus disable explicitly installing usrmerge, unbreaking the generation of debian < stretch and ubuntu images.


fsateler:
cc @stefan-it

keszybz:
Yeah, the debootstrap option is a nicer solution.

PR #19, opened 2016-09-28
debian: do not fail if root/etc/dracut.conf.d already exists

fsateler:
This can happen if dracut is asked to be installed by the user

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -472,7 +472,7 @@ def install_debian_or_ubuntu(args, workspace, run_build_script, mirror):
     subprocess.run(cmdline, check=True)
 
     # Work around debian bug #835628
-    os.makedirs(os.path.join(workspace, "root/etc/dracut.conf.d"))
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -472,7 +472,10 @@ def install_debian_or_ubuntu(args, workspace, run_build_script, mirror):
     subprocess.run(cmdline, check=True)
 
     # Work around debian bug #835628
-    os.makedirs(os.path.join(workspace, "root/etc/dracut.conf.d"))
+    try:
+        os.makedirs(os.path.join(workspace, "root/etc/dracut.conf.d"))
```
keszybz: This would work, but it's nicer to use exist_ok=True parameter instead.

fsateler: doh! of course. Will fix

fsateler:
pushed the change to using `exist_ok=True`

PR #26, opened 2016-10-06
Allow setting root password in the image

keszybz:
Those images are supposed to be used locally for testing, so it
should be as easy to log in as possible.

As a special case, --password= completely deletes the password.
In other words, it's not possible to set an empty password.

Note: the obvious thing of calling /bin/passwd does not work because
audit or some other thing trips it up.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -4,6 +4,7 @@ import argparse
 import configparser
 import contextlib
 import ctypes, ctypes.util
+import crypt
 import hashlib
 import os
 import platform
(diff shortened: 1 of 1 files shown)


poettering:
lgtm, but curious about the point @lucaswerkmeister raised?

lucaswerkmeister:
It’s also worth noting that this will create an MD5 hash, since that’s all that openssl can generate. But according to [this serverfault answer](http://serverfault.com/a/330072/343005), Python includes a module that can do the job, no need for subprocesses:
```python
import crypt;

password = crypt.crypt(args.password, crypt.mksalt(crypt.METHOD_SHA512)))
```
Apparently, this is for Python 3.3+ – is that okay for mkosi?

keszybz:
Repushed with `crypt.crypt()`. That's a much better idea.

PR #28, opened 2016-10-08
Modprobe filesystems before mounting

g2p:
Due to security hardening, mount needs the relevant
filesystem modules to be loaded first.


g2p:
Never mind, this isn't generally required.

PR #31, opened 2016-10-12
Debian fixes

fsateler:
- Debootstrap does not have a proper resolver so delay installing packages until we have apt.
- Force not starting daemons with policy-rc.d
- Allow running apt with network access

Note that because debootstrap uses the same dir as apt for its cache, the `--cache` is not busted by delaying the installs :)

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -528,8 +528,20 @@ def install_debian_or_ubuntu(args, workspace, run_build_script, mirror):
         extra_packages += ["linux-image-amd64", "dracut"]
 
     if extra_packages:
+        # Debian policy is to start daemons by default.
+        # The policy-rc.d script can be used choose which ones to start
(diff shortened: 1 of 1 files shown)


fsateler:
Yeah, [it is](http://sources.debian.net/src/init-system-helpers/1.45/doc/README.policy-rc.d/). This program is run by the packaging helpers to determine if it should start the given daemon.

keszybz:
Can you put this comment in the commit message, or maybe even better in a comment in the code? The mechanism is not obvious otherwise.

fsateler:
>
> Can you put this comment in the commit message, or maybe even better in a
> comment in the code? The mechanism is not obvious otherwise.
>

Sure.  Will do.

-- 

Saludos,
Felipe Sateler

fsateler:
Force pushed a version with the comment

PR #37, opened 2016-11-07
fix mirror for Fedora 25, improve keys handling

davide125:
- look for Fedora 25 in the right place (`development/`, not `releases/`)
- add keys for all Fedora releases

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -13,6 +13,7 @@ import subprocess
 import sys
 import tempfile
 import time
+import urllib.request
 import uuid
 from enum import Enum
 
@@ -58,6 +59,13 @@ else:
     sys.exit(1)
 
 CLONE_NEWNS = 0x00020000
+
+FEDORA_KEYS_MAP = {
+  "23": "34EC9CBA",
+  "24": "81B46521",
+  "25": "FDB19C98",
+  "26": "64DAB85D",
+}
 
 def unshare(flags):
     libc = ctypes.CDLL(ctypes.util.find_library("c"), use_errno=True)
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -427,16 +434,21 @@ def install_fedora(args, workspace, run_build_script):
     if os.path.exists(gpg_key):
         gpg_key = "file://%s" % gpg_key
     else:
-        gpg_key = "https://getfedora.org/static/81B46521.txt"
+        gpg_key = "https://getfedora.org/static/{key}.txt".format(key=FEDORA_KEYS_MAP[args.release])
```
keszybz: If there's just one argument, using named keys seems overkill. I don't think this is an improvement.

mkosi
```diff
@@ -427,16 +434,21 @@ def install_fedora(args, workspace, run_build_script):
...
     if args.mirror:
-        release_url = "baseurl={.mirror}/releases/{.release}/Everything/x86_64/os/".format(args)
-        updates_url = "baseurl={.mirror}/updates/{.release}/x86_64/".format(args)
+        if args.release == "25":
+            section = "development"
+        else:
+            section = "releases"
+
+        release_url = "baseurl={args.mirror}/{section}/{args.release}/Everything/x86_64/os/".format(args=args, section=section)
+        updates_url = "baseurl={args.mirror}/updates/{args.release}/x86_64/".format(args=args)
```
keszybz: That's probably better than what was before..., but it feels suboptimal. By the time this mkosi is released, Fedora will likely have released F25, and this will be broken. Maybe we should do a query of the URL (using `requests` or `urllib`), and pick the URL that works?

mkosi
davide125: Happy to change it back, I mostly did it for consistency with the rest.

mkosi
davide125: I couldn't find a definitive place for this, so I guess we'll just have to try `releases/` first and fallback to `development/` if that fails, probably checking for something like `$section/$release/Everything/$arch/os/media.repo`.

PR #38, opened 2016-11-14
Ignore chattr failure due to missing file system support

lucaswerkmeister:
This makes it possible to run mkosi in a tmpfs file system.

---

As I understand it, `chattr +C` (no copy on write) is only an optimization on COW filesystems, so if it can’t be set, that shouldn’t be a problem. With this patch, I was able to build a fully functional image on `/tmp` (tmpfs).

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -129,7 +129,8 @@ def create_image(args, workspace):
     print_step("Creating partition table...")
 
     f = tempfile.NamedTemporaryFile(dir = os.path.dirname(args.output), prefix='.mkosi-')
-    subprocess.run(["chattr", "+C", f.name], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, check=True)
+    # disable copy-on-write if applicable on filesystem
+    subprocess.run(["chattr", "+C", f.name], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, check=False)
     f.truncate(image_size(args))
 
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -129,7 +130,13 @@ def create_image(args, workspace):
...
 
     f = tempfile.NamedTemporaryFile(dir = os.path.dirname(args.output), prefix='.mkosi-')
-    subprocess.run(["chattr", "+C", f.name], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, check=True)
+    try:
+        subprocess.run(["chattr", "+C", f.name], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+    except subprocess.CalledProcessError as e:
+        if e.stderr.startswith(("chattr: " + os.strerror(errno.ENOTTY)).encode()) or e.stderr.startswith(("chattr: " + os.strerror(errno.ENOTSUP)).encode()):
+            pass # file system does not support "no copy on write" attribute
+        else:
+            raise e
```
lucaswerkmeister: I guess that wouldn’t hurt either… should a warning be printed to the console, or the error be ignored completely silently?

poettering: yes let's ignore the return value, please!

lucaswerkmeister: Alright, that makes the patch much smaller :) pushed

PR #39, opened 2016-11-22
add support for generating squashfs and dm-verity disk images

poettering:
And a couple of other fixes

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,8 @@ The following output formats are supported:
 
 * Raw *GPT* disk image, with btrfs as root (*raw_btrfs*)
 
+* Raw *GPT* disk image, with squashfs as read-only root (*raw_squashfs*)
+
 * Plain directory, containing the *OS* tree (*directory*)
 
 * btrfs subvolume, with separate subvolumes for `/var`, `/home`,
@@ -245,5 +247,5 @@ It is usually easiest to use the distribution package.
 When not using distribution packages, for example, on *Fedora* you need:
 
 ```bash
-dnf install python3 debootstrap arch-install-scripts xz btrfs-progs dosfstools edk2-ovmf
+dnf install arch-install-scripts btrfs-progs debootstrap dosfstools edk2-ovmf squashfs-tools gnupg python3 tar xz
 ```
--- a/README.md
+++ b/README.md
@@ -169,6 +169,9 @@ they exist in the local directory:
   next to image files it boots, for additional container
   runtime settings.
 
+* `mkosi.cache` may be a directory. If so, it is automatically used as
+  package download cache, in order to speed repeated runs of the tool.
+
 All these files are optional.
 
 Note that the location of all these files may also be
--- a/README.md
+++ b/README.md
@@ -30,6 +30,9 @@ options are available:
 
 * Separate partitions for `/srv` and `/home` may be added in
 
+* A dm-verity partition may be added in that adds runtime integrity
+  data for the root partition
+
 # Compatibility
 
 Generated images are *legacy-free*. This means only *GPT* disk
@@ -250,5 +253,5 @@ It is usually easiest to use the distribution package.
 When not using distribution packages, for example, on *Fedora* you need:
 
 ```bash
-dnf install arch-install-scripts btrfs-progs debootstrap dosfstools edk2-ovmf squashfs-tools gnupg python3 tar xz
+dnf install arch-install-scripts btrfs-progs debootstrap dosfstools edk2-ovmf squashfs-tools gnupg python3 tar veritysetup xz
 ```
--- a/mkosi
+++ b/mkosi
@@ -199,7 +199,7 @@ def create_image(args, workspace):
             run_sfdisk = True
 
     if args.output_format != OutputFormat.raw_squashfs:
-        table += 'type={}, name="Root Partition"\n'.format(GPT_ROOT_NATIVE)
+        table += 'type={}, name="Root Partition", attrs="{}"\n'.format(GPT_ROOT_NATIVE, "GUID:60" if args.read_only and args.output_format == OutputFormat.raw_gpt else "")
         run_sfdisk = True
 
     args.root_partno = pn
--- a/mkosi
+++ b/mkosi
@@ -20,8 +20,8 @@ from enum import Enum
 __version__ = '1'
 
 # TODO
-# - squashfs root
 # - volatile images
+# - set gpt read-only bit
 # - make debian/ubuntu images bootable
 # - work on device nodes
 # - allow passing env vars
--- a/mkosi
+++ b/mkosi
@@ -117,14 +117,14 @@ def setup_workspace(args):
 
 def btrfs_subvol_create(path, mode=0o755):
     m = os.umask(~mode & 0o7777)
-    subprocess.run(["btrfs", "subvol", "create", path], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, check=True)
+    subprocess.run(["btrfs", "subvol", "create", path], check=True)
     os.umask(m)
 
-def btrfs_subvol_delete(path, mode=0o755):
+def btrfs_subvol_delete(path):
     subprocess.run(["btrfs", "subvol", "delete", path], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, check=True)
 
 def btrfs_subvol_make_ro(path, b=True):
-    subprocess.run(["btrfs", "property", "set", path, "ro", "true" if b else "false"], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, check=True)
+    subprocess.run(["btrfs", "property", "set", path, "ro", "true" if b else "false"], check=True)
 
 def image_size(args):
     size = GPT_HEADER_SIZE + GPT_FOOTER_SIZE
--- a/mkosi
+++ b/mkosi
@@ -25,7 +25,6 @@ __version__ = '1'
 # - make debian/ubuntu images bootable
 # - work on device nodes
 # - allow passing env vars
-# - rework cache management to use mkosi.cache by default in the project dir
 
 class OutputFormat(Enum):
     raw_gpt = 1
@@ -1512,6 +1511,14 @@ def find_extra(args):
         else:
             args.extra_trees.append("mkosi.extra")
 
+def find_cache(args):
+
+    if args.cache_path is not None:
+        return
+
+    if os.path.exists("mkosi.cache/"):
+        args.cache_path = "mkosi.cache/" + args.distribution.name + "~" + args.release
+
 def find_build_script(args):
     if args.build_script is not None:
         return
@@ -1579,6 +1586,8 @@ def load_args():
             args.release = "unstable"
         elif args.distribution == Distribution.ubuntu:
             args.release = "yakkety"
+
+    find_cache(args)
 
     if args.mirror is None:
         if args.distribution == Distribution.fedora:
--- a/mkosi
+++ b/mkosi
@@ -280,6 +280,10 @@ def mount_bind(what, where):
     os.makedirs(where, 0o755, True)
     subprocess.run(["mount", "--bind", what, where], check=True)
 
+def mount_tmpfs(where):
+    os.makedirs(where, 0o755, True)
+    subprocess.run(["mount", "tmpfs", "-t", "tmpfs", where], check=True)
+
 @contextlib.contextmanager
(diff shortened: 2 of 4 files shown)


mkosi
```diff
@@ -350,7 +350,7 @@ def umount(where):
     # Ignore failures and error messages
     subprocess.run(["umount", "-n", where], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
 
-def prepare_tree(args, workspace):
+def prepare_tree(args, workspace, run_build_script):
```
lucaswerkmeister: Shouldn’t the `prepare_tree` call in line 1587 also be updated to propagate the `run_build_script` parameter?

poettering: indeed!

poettering:
I force pushed a new version now that fixes the issue @lucaswerkmeister pointed out. Please have a look.

(This was actually just a "git add -p" mistake... A later commit got this right, but it should already have been in the earlier commit, and is now)

lucaswerkmeister:
Thanks, I must’ve missed the later commit then.

Perhaps update the README to add the `squashfs-tools` dependency?

poettering:
> Perhaps update the README to add the squashfs-tools dependency?

Ah, of course! Will add!

poettering:
Force pushed a new version that also update the README with the "squashfs-tools" dep. Please have a look!

mkosi
```diff
@@ -849,8 +881,98 @@ def make_tar(args, workspace):
...
+    else:
+        old_table, last_partition_sector = read_partition_table(loopdev)
+
+    args.root_size = roundup512(os.stat(squashfs.name).st_size)
+    new_size = last_partition_sector + args.root_size + GPT_FOOTER_SIZE
+
+    print_step("Resizing disk image to {}...".format(format_bytes(new_size)))
+
+    os.truncate(raw.name, new_size)
+    subprocess.run(["losetup", "--set-capacity", loopdev])
```
poettering: indeed!

poettering:
Pushed another version, adding this missing check

lucaswerkmeister:
Okay, now I can’t find anything else to complain about :)

poettering:
thanks for the review! Would you be interested in privileges so that you can press the green merge button? It's a good feeling to merge code, I can tell you! ;-)

lucaswerkmeister:
Wow, quite an unexpected honor, thanks!

lucaswerkmeister:
@keszybz Does `-t raw_btrfs --read-only` fail as well? (I can’t test it myself right now, need to reboot first.)

keszybz:
```
$ sudo ../mkosi/mkosi --password=foobar -t raw_btrfs --read-only --force
...
‣ Detaching image file completed.
  File "../mkosi/mkosi", line 1891, in <module>
    main()
  File "../mkosi/mkosi", line 1887, in main
    build_stuff(args)
  File "../mkosi/mkosi", line 1833, in build_stuff
    raw, tar = build_image(args, workspace, run_build_script=True)
  File "../mkosi/mkosi", line 1786, in build_image
    make_read_only(args, workspace.name)
  File "../mkosi/mkosi", line 891, in make_read_only
    btrfs_subvol_make_ro(os.path.join(workspace, "root"))
  File "../mkosi/mkosi", line 128, in btrfs_subvol_make_ro
    subprocess.run(["btrfs", "property", "set", path, "ro", "true" if b else "false"], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, check=True)
  File "/usr/lib64/python3.5/subprocess.py", line 711, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['btrfs', 'property', 'set', '/var/tmp/mkosi-89aev0qf/root', 'ro', 'true']' returned non-zero exit status 1
```

But that's a different error: setting ro itself fails.

mkosi
```diff
@@ -30,6 +30,7 @@ __version__ = '1'
 class OutputFormat(Enum):
     raw_gpt = 1
     raw_btrfs = 2
+    raw_squashfs = 6
```
poettering: I didn't want to change all the lines in the enum... But I figure it would be fine...

keszybz:
On Tue, Nov 29, 2016 at 08:37:16AM -0800, Lennart Poettering wrote:
> poettering commented on this pull request.
> 
> 
> 
> > @@ -30,6 +30,7 @@ __version__ = '1'
>  class OutputFormat(Enum):
>      raw_gpt = 1
>      raw_btrfs = 2
> +    raw_squashfs = 6
> 
> I didn't want to change all the lines in the enum... But I figure it would be fine...

It's not a public API, doesn't have to be stable.

keszybz:
> But that's a different error: setting ro itself fails.

Yep, removing the redirection of stderr to /dev/null shows the following:
```
‣ Marking root subvolume read-only...
ERROR: object is not a btrfs object: /var/tmp/mkosi-gg_oyucb/root
usage: btrfs property set [-t <type>] <object> <name> <value>

    Sets a property on a btrfs object.

    Please see the help of 'btrfs property get' for a description of
    objects and object types.

‣ Detaching image file...
‣ Detaching image file completed.
  File "../mkosi/mkosi", line 1891, in <module>
    main()
  File "../mkosi/mkosi", line 1887, in main
    build_stuff(args)
  File "../mkosi/mkosi", line 1833, in build_stuff
    raw, tar = build_image(args, workspace, run_build_script=True)
  File "../mkosi/mkosi", line 1786, in build_image
    make_read_only(args, workspace.name)
  File "../mkosi/mkosi", line 891, in make_read_only
    btrfs_subvol_make_ro(os.path.join(workspace, "root"))
  File "../mkosi/mkosi", line 128, in btrfs_subvol_make_ro
    subprocess.run(["btrfs", "property", "set", path, "ro", "true" if b else "false"], check=True)
  File "/usr/lib64/python3.5/subprocess.py", line 711, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['btrfs', 'property', 'set', '/var/tmp/mkosi-gg_oyucb/root', 'ro', 'true']' returned non-zero exit status 1
```

poettering:
So, currently the development image and the final image always are built the same way. This means if your build script expects a writable image, and squashfs is used, then things blow up.

Not entirely sure what the best approach is. We could:

 a. Build the development image with ext4, and only the final image with squashfs
 b. We could use overlayfs to make the build tree writable even if the backend file system actually isn't
 c. we could not do anything, and pass the problem to the caller: if he wants to build read-only images, he needs to make sure to build with a build tree in /var/tmp

I don't think c would be a terrible choice. But then again it's certainly not the most comfortable solution... And I figure mkosi is supposed to be comfortable, so maybe we should not require this.

Option a. has the drawback that it would require sizing the ext4 partition to something during build, even though the nice feature of squashfs after all is that the size of the root disk is minimal and automatically sized.

I am mostly leaning towards option b right now I must say. It appears like a nice way out: the development image is as close as possible to the final image, but the source tree could be made writable this way. I am usually not a big fan of overlayfs, but in this case it seems OK, as we have a pretty good idea about the kind of write access we'll see on the directory.

I'll be working on adding support for that, but I think this should probably be a different PR... The current version of the PR should work fine as long as you have no build script in place, or your build script builds out-of-tree to /tmp or so.

keszybz:
@poettering I pushed two commits to your branch to elucidate and fix the issue from https://github.com/systemd/mkosi/pull/39#issuecomment-263625891. I hope that you're OK with me doing that. But maybe my fix is wrong, please drop my commit if so.

Current state is that `--read-only` is broken, and so is compilation on the squashfs image, which is also readonly. But this is a regression with this PR, the same command (`sudo ../mkosi/mkosi --password=foobar -t raw_btrfs --read-only --force`) works fine with mkosi master.

> The current version of the PR should work fine as long as you have no build script in place, or your build script builds out-of-tree to /tmp or so.

I was just using systemd as a test case (the default build script).

lucaswerkmeister:
I agree that c isn’t a terrible choice in principle, but unless we can somehow detect the “read-only file system” errors and print a better error message (“please change your build script to operate in /var/tmp”), it’s not very user friendly. I agree that option b sounds better.

> But this is a regression with this PR

Looks like it was introduced in db0d4bf5a9d20c0a447d7e954ef64ed379c44909 (“add support for creating squashfs images”, the main commit, near the end of the diff), perhaps it would be better to squash the fix commit into that one?

poettering:
OK, I force pushed a new version now. It should fix the pointed out issues (I merged @keszybz's commit into my own though), and makes sure that building works now even with squashfs. I went for the overlay path, but to make this nice I added some new tricks to nspawn, which are waiting in https://github.com/systemd/systemd/pull/4694. Hence, when building read-only images we now depend on a yet-to-be-released systemd version for now (or actually even a yet-to-be-merged PR for a yet-to-be-released systemd version).

poettering:
I force pushed a new version now that adds dm-verity support on top, please have a look!

poettering:
There's now support for making use of this dm-verity info in systemd, waiting in https://github.com/systemd/systemd/pull/4795.

mkosi
```diff
@@ -956,42 +973,102 @@ def read_partition_table(loopdev):
...
+        return
+
+    print_step("Inserting squashfs root partition...")
+    args.root_size = insert_partition(args, workspace, raw, loopdev, args.root_partno, squashfs, "Root Partition", GPT_ROOT_NATIVE)
+    print_step("Inserting squashfs root partition completed.")
+
+def make_verity(args, workspace, loopdev, run_build_script):
+
+    if run_build_script or not args.verity:
+        return None, None, None
```
keszybz: This returns 3 args here...

mkosi
```diff
@@ -956,42 +973,102 @@ def read_partition_table(loopdev):
...
+    for line in c.stdout.decode("utf-8").split('\n'):
+        if line.startswith("Root hash:"):
+            root_hash = line[10:].strip()
 
-    print_step("Inserting squashfs file system completed.")
+    assert root_hash is not None
+
+    print_step("Generating verity hashes completed.")
+
+    return f, root_hash
```
keszybz: ... and two here.

Fails with:
```
    verity, root_hash = make_verity(args, workspace.name, loopdev, run_build_script)
ValueError: too many values to unpack (expected 2)
```

mkosi
```diff
@@ -1009,6 +1086,19 @@ def xz_output(args, raw):
 
     return f
 
+def write_root_hash_file(args, root_hash):
+    if root_hash is None:
+        return None
+
+    print_step("Writing .roothash file...")
+
+    f = tempfile.NamedTemporaryFile(mode = "w+b", dir=os.path.dirname(args.output_root_hash_file), prefix='.mkosi')
```
keszybz: `mode='w+b'` is the standard python style.

mkosi
```diff
@@ -1767,7 +1892,7 @@ def build_image(args, workspace, run_build_script):
     # If there's no build script set, there's no point in executing
     # the build script iteration. Let's quite early.
     if args.build_script is None and run_build_script:
-        return (None, None)
+        return (None, None, None)
```
keszybz: This really calls for a `namedtuple`, but let's do that in a cleanup PR once this one lands.

mkosi
```diff
@@ -761,6 +761,23 @@ def install_distribution(args, workspace, run_build_script):
...
+    # is either initialized and commited (if /etc is writable) or the
+    # image runs with a transient machine ID, that changes on each
+    # boot (if the image is read-only).
+
+    print_step("Resetting machine ID...")
+
+    machine_id = os.path.join(workspace, 'root', 'etc/machine-id')
+
+    os.unlink(machine_id)
+    f = open(machine_id, "w+b");
```
keszybz: I'd prefer `open(machine_id, "w+b").close()`.

(While it's true that with cpython, `del` causes `close` to be called immediately, but in principle somebody could use `pypy3` to run this, at least once it starts supporting modern python syntax again. So, let's keep up the illusion of portability and not require destructors to be called immediately ;))

README.md
```diff
@@ -245,5 +253,5 @@ It is usually easiest to use the distribution package.
 When not using distribution packages, for example, on *Fedora* you need:
 
 ```bash
-dnf install python3 debootstrap arch-install-scripts xz btrfs-progs dosfstools edk2-ovmf
+dnf install arch-install-scripts btrfs-progs debootstrap dosfstools edk2-ovmf squashfs-tools gnupg python3 tar xz
```
keszybz: Also `veritysetup`.

poettering:
thanks for the review! force pushed a new version that fixes all issues you pointed out!

PR #40, opened 2016-11-26
Several improvements over Debian/Ubuntu image creation and support for multiple mkosi.defaults files

zaro:
Most of the patches fix/improve in the Debian/Ubuntu image building and are almost self explanatory .
1c0e243  - add support for reading multiple mkosi.default files from a mkosi.defaults.d directory allowing overriding/adding of settings.
I am trying to port few existing docker images  to mkosi. And having the option to have a base image  upon which to build is very important as it removes a lot of duplication. Since mkosi currently cannot  inherit/include  other images, the closest thing would be to include/inherit configuration . This way there can be a base mkosi.default , symlinked in each separate images mkosi.defaults.d dir, and having one specific for the image that adds packages.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1398,6 +1398,8 @@ def load_args():
             args.mirror = "http://httpredir.debian.org/debian"
         elif args.distribution == Distribution.ubuntu:
             args.mirror = "http://archive.ubuntu.com/ubuntu"
+            if platform.machine() == "aarch64":
+                args.mirror = "http://ports.ubuntu.com/"
         elif args.distribution == Distribution.arch:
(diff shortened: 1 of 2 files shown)


mkosi
```diff
@@ -1301,13 +1302,44 @@ def load_defaults(args):
     config.optionxform = str
     config.read_file(f)
 
+    # this is used only for validation
```
fsateler: Validation of what?

zaro: If the section/key name are valid, so that if we encounter and invalid section or key name the file name containing it can be printed.

Basically All files are read up in a options dict, and  then it is used in **load_defaults** to set values in the real args object.

PR #46, opened 2016-12-09
mkosi: add support for running postinst scripts and other fixes

poettering:
This allows running a script "mkosi.postinst" as last step before considering the image built final, inside the environment. This is useful to do arbitrary final adjustments. In particular, when working on systemd it's useful to regenerate the initrd after dropping in the systemd built artifacts, so that they also are updated in the initrd.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -738,20 +738,28 @@ def install_distribution(args, workspace, run_build_script):
 
     install[args.distribution](args, workspace, run_build_script)
 
-@complete_step('Resetting machine ID')
-def reset_machine_id(args, workspace):
+def reset_machine_id(args, workspace, run_build_script):
     """Make /etc/machine-id an empty file.
 
     This way, on the next boot is either initialized and commited (if /etc is
     writable) or the image runs with a transient machine ID, that changes on
     each boot (if the image is read-only).
     """
-    machine_id = os.path.join(workspace, 'root', 'etc/machine-id')
-    os.unlink(machine_id)
-    open(machine_id, "w+b").close()
 
-def set_root_password(args, workspace):
+    if run_build_script:
+        return
+
+    with complete_step('Resetting machine ID'):
+        machine_id = os.path.join(workspace, 'root', 'etc/machine-id')
+        os.unlink(machine_id)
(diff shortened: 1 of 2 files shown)


lucaswerkmeister:
Looks mostly okay to me, just two comments on the commit message of the main commit:

> This adds an additional scripting hook to the build script. We look for
"mkosi.postinst". If it existsthe script will be run as last step of the
final build.

There’s a missing space in “existsthe”, and the wording is a bit confusing IMO since the script is run for both images (build and final), with a different first argument to indicate the difference.

The new file should probably also be mentioned in the `README`, though I guess you could leave that to a separate PR.

mkosi
```diff
@@ -739,19 +739,27 @@ def install_distribution(args, workspace, run_build_script):
...
+def reset_machine_id(args, workspace, run_build_script):
     """Make /etc/machine-id an empty file.
 
     This way, on the next boot is either initialized and commited (if /etc is
     writable) or the image runs with a transient machine ID, that changes on
     each boot (if the image is read-only).
     """
+
+    if run_build_script:
+        return
```
keszybz: This conflicts with using `complete_step` as a decorator. We shouldn't print anything if we skip the step.

mkosi
```diff
@@ -1891,10 +1891,9 @@ def run_build_script(args, workspace, raw):
         dest = os.path.join(workspace, "dest")
         os.mkdir(dest, 0o755)
 
-        directory = os.path.join(workspace, "root") if raw is None else "--image=" + raw.name
         cmdline = ["systemd-nspawn",
                    '--quiet',
-                   "--directory", directory,
+                   "--directory=" + os.path.join(workspace, "root") if raw is None else "--image=" + raw.name,
```
keszybz: Oops, I messed that up. But maybe it's better to keep it split out:
```python
target = os.path.join(workspace, "root") if raw is None else "--image=" + raw.name
cmdline = [...
           target,
           ...]
```
?

lucaswerkmeister:
I force-pushed a tiny typo fix in the `README`, rest looks good to me.

Issue #42, opened 2016-11-30
Is there any use for luks support?


ingobecker:
I have created a fork and added basic [luks support](https://github.com/ingobecker/mkosi/commit/8a1c3f18b2108b7f13c82243660098d746c9d6e6)  in order to boostrap images that are fulldiskencrypted. I'm using mkosi to bootstrap images which i can later copy on bare metal machines (or VMs). I'm aware of the fact, that systemd-nspawn doesn't support luks partitions and i'm not sure if it makes any sense to use encrypted partitions in containers. I just wanted to mention my fork and ask if it makes sense beyond bootstraping bare metal images. If so, i could clean up my code and create a pull request.

poettering:
I think LUKS support would be great to have (and direct LUKS support in nspawn would be great too, and dm-verity support would be great too). So please submit a PR!

ingobecker:
There are still some parts missing like encrypting swap partitions, supplying luks format parameters and a way to include keyfiles into the resulting image (which may not be needed once systemd-nspawn is able to handle luks partitions). I'm working on it and i will have a look for debian support as well, when i'm done the named features. I will create a PR once i think arch support is working in a stable manner.

poettering:
I wouldn't make too many things configurable, in fact I'd make this a single option that can have three values:

a) `off` — no encryption is done
b) `data` — the root partition is not encrypted (except on btrfs), but the /home and /srv partitions are, including swap
c) `all` — all partitions are encrypted, including swap

I am not convinced there's really value in configuring things in more detail.

And regarding the swap, I think we should probably never use LUKS for that by default. Instead, have an extra GPT partition flag. If that's set on a swap partition we should set up a raw dm-crypt with a random key on boot and invoke mkswap accordinly on it.

ingobecker:
Ok, so how do you think the key should be supplyed?
At the moment, you can setup a passphrase by directly supplying it as a commandline argument when invoking mkosi or by setting it inside a mkosi.default file. It is also possible to insert a passphrase interactively while the image is created. There is also the possibility to use a keyfile. Which way do you think is appropriate?

There are also distinct keys to setup for the different partitions. Do you think all partitions should be using the same passphrase/keyfile?

What should be done about encryption details? Should we just go with the defaults, crypsetup is using?

ingobecker:
Another option would be to have a keyfile created automatically.

poettering:
[...]

PR #49, opened 2016-12-14
Add LUKS support

poettering:
This is a rebased, improved version of @ingobecker's LUKS patches, posted as #43. I made a number of changes over the original PR (besides the actual rebasing and some bug fixing):

* there is an addition to NEWS now
* i dropped the bootloader bits for now. I have been working on adding root luks auto-discovery to systemd, and I'd really prefer if that's the way forward. Note that the "mkosi.postinst" stuff is flexible enough to add in the support locally in projects
* I made sure LUKS support also works for squashfs and btrfs root devices
* We now keep the LUKS devices open as long as necessary, instead of closing/reopening them frequently. After all this is a pretty slow operation.
* I reworked the relevant cryptsetup commands to use --batch-mode

@ingobecker, please have a look. I hope it's OK that I updated your PR. I left you in as commiter.

Fixes: #42

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -5,16 +5,19 @@ import configparser
 import contextlib
 import ctypes, ctypes.util
 import crypt
+import getpass
 import hashlib
 import os
 import platform
 import shutil
+import stat
 import subprocess
 import sys
 import tempfile
 import time
 import urllib.request
 import uuid
+
 from enum import Enum
 
 __version__ = '1'
@@ -247,6 +250,9 @@ def attach_image_loopback(args, raw):
             subprocess.run(["losetup", "--detach", loopdev], check=True)
 
 def partition(loopdev, partno):
+    if partno is None:
+        return None
+
     return loopdev + "p" + str(partno)
 
 def prepare_swap(args, loopdev):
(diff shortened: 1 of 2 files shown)


mkosi
```diff
@@ -1626,6 +1756,30 @@ def find_postinst_script(args):
...
+def find_passphrase(args):
+
+    if args.encrypt is None:
+        args.passphrase = None
+        return
+
+    try:
+        passphrase_mode = os.stat('mkosi.passphrase').st_mode & (stat.S_IRWXU | stat.S_IRWXG | stat.S_IRWXO)
+        if (passphrase_mode & stat.S_IRWXU > 0o600) or (passphrase_mode & (stat.S_IRWXG | stat.S_IRWXO) > 0):
+            print_step("Permissions of 'mkosi.passphrase' of '{}' are too open. Please make sure that the file is only accessible by its owner. Aborting.".format(oct(passphrase_mode)))
```
lucaswerkmeister: I don’t really love this phrasing – it sounds like the solution is to `chmod` the file, but it would be better to generate a new passphrase (giving it the right permissions from the start), since an adversary might already have read the current passphrase file. Is there a good way to express that without making the error message too long?

Also, all the other `sys.exit`s are preceded by `sys.stderr.write` instead of `print_step` – error messages aren’t steps. (There’s enough of them that it might be a good idea to make a `die(message)` function, but that’s for another PR.)

poettering: indeed! I force pushed a new version now that rewords the message. I hope this one is better, please have a look. I also change this to use sys.stderr.write() like the rest of the code.

thanks for the review!

poettering:
Feel free to click the green merge button when you approve. Our workflow has been that we generally merge the stuff after a single positive review!

/me does that now.

PR #51, opened 2016-12-16
generate unified kernel+initrd images that include the root hash as last step of the build + other improvements

poettering:
This adds support for generating combined kernel+initrd image files as last step of the boot that also contain the root hash if verity is enabled. These kernel+initrd images are unified blobs that can be signed as one, in order to implement a complete SecureBoot chain.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1088,7 +1088,7 @@ def read_partition_table(loopdev):
 
         table.append(stripped)
 
-        name, rest = stripped.split(":")
+        name, rest = stripped.split(":", 1)
         fields = rest.split(",")
 
         start = None
--- a/mkosi
+++ b/mkosi
@@ -24,7 +24,6 @@ __version__ = '1'
 
 # TODO
 # - volatile images
-# - set gpt read-only bit
 # - make debian/ubuntu images bootable
 # - work on device nodes
 # - allow passing env vars
--- a/mkosi
+++ b/mkosi
@@ -214,7 +214,7 @@ def create_image(args, workspace):
                 run_sfdisk = True
 
         if args.output_format != OutputFormat.raw_squashfs:
-            table += 'type={}, name="Root Partition", attrs="{}"\n'.format(GPT_ROOT_NATIVE, "GUID:60" if args.read_only and args.output_format == OutputFormat.raw_gpt else "")
+            table += 'type={}, attrs={}, name="Root Partition"\n'.format(GPT_ROOT_NATIVE, "GUID:60" if args.read_only and args.output_format != OutputFormat.raw_btrfs else "")
             run_sfdisk = True
 
         args.root_partno = pn
@@ -1137,7 +1137,7 @@ def insert_partition(args, workspace, raw, loopdev, partno, blob, name, type_uui
     if uuid is not None:
         table += "uuid=" + str(uuid) + ", "
 
-    table += 'size={}, type={}, name="{}", attrs="GUID:60"\n'.format((blob_size + luks_extra) // 512, type_uuid, name)
+    table += 'size={}, type={}, attrs=GUID:60, name="{}"\n'.format((blob_size + luks_extra) // 512, type_uuid, name)
 
     print(table)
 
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -1228,6 +1255,58 @@ def patch_root_uuid(args, loopdev, root_hash):
...
+    # menu. These "unified" single-file images have the benefit that
+    # they can be signed like normal EFI binaries, and can encode
+    # everything necessary to boot a specific root device, including
+    # the root hash.
+
+    if not args.bootable:
+        return
+
+    if args.distribution != Distribution.fedora:
+        return
```
lucaswerkmeister: Perhaps it would be cleaner to put everything below in an `install_unified_kernel_fedora` function and make `install_unified_kernel` delegate to the appropriate function for the distribution? Arch Linux, at least, will probably look quite different (use `mkinitcpio` instead of `dracut` (or `kernel-install`?) and then patch together the executable with `objcopy`).

mkosi
```diff
@@ -1228,6 +1255,58 @@ def patch_root_uuid(args, loopdev, root_hash):
...
+                          "--no-hostonly",
+                          "--uefi",
+                          "--kver", kver.name,
+                          "--kernel-cmdline", cmdline ]
+
+                # Temporary fix until dracut includes these in the image anyway
+                dracut += [ "-i", "/usr/lib/systemd/system/systemd-volatile-root.service", "/usr/lib/systemd/system/systemd-volatile-root.service",
+                            "-i", "/usr/lib/systemd/systemd-volatile-root", "/usr/lib/systemd/systemd-volatile-root",
+                            "-i", "/usr/lib/systemd/systemd-veritysetup", "/usr/lib/systemd/systemd-veritysetup",
+                            "-i", "/usr/lib/systemd/system-generators/systemd-veritysetup-generator", "/usr/lib/systemd/system-generators/systemd-veritysetup-generator" ]
```
lucaswerkmeister: You could loop over the four files to avoid specifying the long path twice each time, but meh, this works too.

mkosi
poettering: hmm, i am a bit conservative on adding abstraction code without actually working on at least two backends... Hence I just hacked this up as it is, and then leave the abstraction work to the one who actually adds archlinux support...

(btw, I think it would be good if Arch's mkinitcpio could natively support a `--uefi` switch or so, the way dracut does. I think think objcopy thing is something that better belongs in the initrd implementation than in mkosi. but anyway, this is something for the arch maintainers to figure out)

mkosi
poettering: well, it's mostly a hack, and I kinda hope this can go away quickly, when dracut gets updated to include these things in the "no-host-only" mode anyway. Hence, I wanted it to look a bit like a hack ;-)

mkosi
```diff
@@ -1228,6 +1255,58 @@ def patch_root_uuid(args, loopdev, root_hash):
...
+    if args.distribution != Distribution.fedora:
+        return
+
+    with complete_step("Generating combined kernel + initrd boot file"):
+
+        cmdline = args.kernel_commandline
+        if root_hash is not None:
+            cmdline += " roothash=" + root_hash
+
+            for kver in os.scandir(os.path.join(workspace, "root", "usr/lib/modules")):
```
poettering: ouch, indeed!

mkosi
poettering: i changed this now a bit and are now using list repitition (i.e. `*2`) to avoid the double paths. I think this looks a bit nicer now!

(In an ideal world we'd use dracut's `-I` switch, instead of `-i` which would remove the need for us to keep repeating each path twice. Unfortunately, I couldn't get it to work — see https://github.com/dracutdevs/dracut/issues/187)

poettering:
I force pushed a new version now, which hopefully addresses all issues raised, with the exception of the unified-kernel generalization bits. Let's leave that for the patch that adds similar support for archlinux...

I also added another patch on top that shows the kernel cmdline to use in the summary output

lucaswerkmeister:
Alright, looks good to me now, let’s press the green button.

lucaswerkmeister:
By the way, this doesn’t add any new dependencies on the host system, correct? Dracut is called inside the image.

poettering:
> By the way, this doesn’t add any new dependencies on the host system, correct? Dracut is called inside the image.

Correct!

mkosi
fsateler: The debian bootable builds also use dracut, so I suppose this can work as-is on debian too?

poettering: @fsateler could be. happy to merge a patch that extends the check if you can verify that it works correctly!

PR #53, opened 2016-12-20
basic UEFI SecureBoot setup

poettering:
Optioanally sign all resulting EFI binaries


README.md
```diff
@@ -197,6 +197,11 @@ they exist in the local directory:
   must have an access mode of 0600 or less. If this file does not
   exist and encryption is requested the user is queried instead.
 
+* `mkosi.secure-boot.crt` and `mkosi.secure-boot.key` may contain a
```
lucaswerkmeister: I think this should be “a**n** X509 certificate” (pronunciation of X509 starts with vowel).

poettering:
force pushed a new version, that fixes the issues pointed out

PR #55, opened 2016-12-23
fixes to make -t directory work again

poettering:
Apparently this has not been tested in a while, restore working support for -t directory

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -579,8 +579,8 @@ def prepare_tree(args, workspace, run_build_script):
             cmdline.write("\n")
 
     if run_build_script:
-        os.mkdir(os.path.join(workspace, "root", "root"))
-        os.mkdir(os.path.join(workspace, "root", "root/dest"))
+        os.mkdir(os.path.join(workspace, "root", "root"), 0o750)
+        os.mkdir(os.path.join(workspace, "root", "root/dest"), 0o755)
 
 def patch_file(filepath, line_rewriter):
     temp_new_filepath = filepath + ".tmp.new"
--- a/mkosi
+++ b/mkosi
@@ -2291,10 +2291,6 @@ def build_image(args, workspace, run_build_script):
 
     tar = make_tar(args, workspace.name, run_build_script)
 
-    # Remove the root directory again, if this is the build script build
-    if run_build_script:
-        unlink_try_hard(os.path.join(workspace.name, "root"))
-
     return raw, tar, root_hash
 
 def var_tmp(workspace):
@@ -2362,6 +2358,11 @@ def build_stuff(args):
 
     if tar is not None:
         del tar
+
+    if args.build_script is not None:
+        with complete_step("Removing artifacts from development build"):
+            unlink_try_hard(os.path.join(workspace.name, "root"))
+            unlink_try_hard(os.path.join(workspace.name, "var-tmp"))
 
     raw, tar, root_hash = build_image(args, workspace, run_build_script=False)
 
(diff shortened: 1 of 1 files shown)


stefan-it:
I'm not sure if it's a bug or feature, but when calling:

```bash
./mkosi -d arch -t directory -o /var/lib/machines/arch-dev --bootable=false
```

`mkosi` recognizes the `mkosi.default` file (a few months ago this was not the case).

poettering:
@stefan-it no, it always did it. It loads the defaults from mkosi.defaults if it exists, and if you specify something on the cmdline then that takes precedence...

stefan-it:
I used the current branch for this PR and used `arch` as distribution and `directory` as format. Then `mkosi` tried to merge the packages for Arch Linux with the packages for the `mkosi.default` file :(

I ran `mkosi` from the root folder of the `mkosi` repository, so the directory creation failed, because `mkosi` wanted to install some Fedora packages specified in `mkosi.default`.

poettering:
@stefan-it well, but that was always the behaviour: we use the "mkosi.default" file for the defaults, and whatever you specify on the command line overrides individual options, but only those. The configuration in effect is hence the combination of the defaults hardcoded in the script, individually overriden by the defaults in your "mkosi.default" file in the working directory, which in turn are overriden by whatever you specify on the command line. This was always the behaviour.

Of course, the resulting combination might be contradictory ultimately: if your mkosi.default configures debian as distro and picks debian packages, but on the command line you then switch to fedora, the debian packages will still be used, and don't make sense anymore.

I figure we could improve that, and when picking a different distro on the command line we also invalidate the package list, but that'd be completely new behaviour, and unrelated to this PR, and hence discussed elsewhere. I am not against changing the behaviour in this regard, but that's really in no way connected to this very PR, and it would be *new* behaviour, as mkosi never worked like that so far.

(An alternative would be to introduce a new Packages.fedora= and BuildPackages.debian= syntax or so, to permit per-distro-type package lists in a single mkosi defualt file, but I am not sure I like the concept...)

mkosi
```diff
@@ -507,18 +511,22 @@ def mount_api_vfs(args, workspace):
...
-
-    if args.cache_path is None:
-        return
 
     # We can't do this in mount_image() yet, as /var itself might have to be created as a subvolume first
-    if args.distribution == Distribution.fedora:
-        mount_bind(args.cache_path, os.path.join(workspace, "root", "var/cache/dnf"))
-    elif args.distribution in (Distribution.debian, Distribution.ubuntu):
-        mount_bind(args.cache_path, os.path.join(workspace, "root", "var/cache/apt/archives"))
+    with complete_step('Mounting Package Cache'):
```
fsateler: This would announce the mounting of the package cache even if there is no such cache.

lucaswerkmeister: Perhaps swap the `with` with the `if` below? The message would still be incorrect if the distribution is Arch, but I think that would be acceptable.

lucaswerkmeister:
> (An alternative would be to introduce a new Packages.fedora= and BuildPackages.debian= syntax or so, to permit per-distro-type package lists in a single mkosi defualt file, but I am not sure I like the concept...)

I like the approach in the systemd repository – if I understand correctly, you’re meant to use the additional files in `.mkosi` as default files in place of the normal one:
```sh
mkosi --default .mkosi/mkosi.arch
```
Sure, there’s some duplication, but there’s also at least one more distro-specific option (`Cache`), and adding `Cache.arch` as well feels a bit silly.

poettering:
I finally pushed a new version of the patch, fixing the issue @fsateler pointed out.

PR #57, opened 2017-01-09
Fix uninitialsed *partno fields

zaro:
Currently building image with directory output fails with :
```
AttributeError: 'Namespace' object has no attribute 'root_partno'
```

I don't know if this is the best place to initialize the *partno fields,  but it works.


lucaswerkmeister:
> Currently building image with directory output fails with :

#55 claims to fix that too, but the fixes look completely different…

zaro:
Just tested the #55, and the problem is gone.

So I guess you can ignore this PR.

PR #59, opened 2017-01-31
Add basic bootable Ubuntu image support

AsciiWolf:
This pull request adds basic support for creating bootable Ubuntu images into mkosi. It uses _initramfs-tools_ instead of _dracut_ because dracut is [not supported](https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1660587) by Ubuntu kernel packages.

I did not test any advanced things like crypto, but a basic btrfs image seems to be working correctly when booted using nspawn. There are [some errors](https://gist.github.com/anonymous/401cc790fd4ad633ce2c6ba618715b9e) being shown when trying to boot a squashfs image using nspawn (and the image seems to be booted read-only), but I am not sure if the errors are related to mkosi or are some local issues with my system.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -781,7 +781,11 @@ def install_debian_or_ubuntu(args, workspace, run_build_script, mirror):
         f.write("hostonly=no")
 
     if args.bootable:
-        extra_packages += ["linux-image-amd64", "dracut"]
+        if args.distribution == Distribution.ubuntu:
+            # Dracut is not supported in Ubuntu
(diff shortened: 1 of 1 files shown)


fsateler:
@AsciiWolf Bootable refers to being able to boot in a virtual machine. Can you boot it using `qemu`? Last time I tried I could not make it work.

AsciiWolf:
@fsateler You are right, I just tried to boot it using qemu and sadly it doesn't seem to work.

PR #60, opened 2017-02-23
mkosi: documentation fix for systemd-nspawn invocation

docent-net:
Missing '-D' parameter for directory name


poettering:
lgtm! thanks!

PR #61, opened 2017-02-24
Invoke the reset_machine_id() function after install_boot_loader()

### LINKED ISSUE #47: Purging the machine-id is incompatible with bootloader installation
Commit f5cafba17214c02f9f80bc5283a165054f6dc828 removed machine-id. The next step, installation of the kernel in boot, fails:
```
‣ Installing boot loader...
Could not determine your machine ID from /etc/machine-id.
Please run 'systemd-machine-id-setup' as root. See man:machine-id(5)
```
Those options are fundamentally incompatible. Should we disable machine-id purging if bootable=yes?

AsciiWolf:
Fixes #47.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -2265,8 +2265,8 @@ def build_image(args, workspace, run_build_script):
 
                 with mount_cache(args, workspace.name):
                     install_distribution(args, workspace.name, run_build_script)
-                    reset_machine_id(args, workspace.name, run_build_script)
                     install_boot_loader(args, workspace.name)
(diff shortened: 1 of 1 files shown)


lucaswerkmeister:
Hm, I think I like this suggestion by @keszybz in the issue:

> Or even move reset_machine_id until just before the image is unmounted.

Would there be a downside to that?

lucaswerkmeister:
Closing in favor of #62, sorry – but still, thanks for your PR!

PR #62, opened 2017-02-26
Fixes to reset_machine_id

### LINKED ISSUE #47: Purging the machine-id is incompatible with bootloader installation

fsateler:
Fixes #47 , and really resets the machine id in case dbus was installed.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -2265,7 +2265,6 @@ def build_image(args, workspace, run_build_script):
 
                 with mount_cache(args, workspace.name):
                     install_distribution(args, workspace.name, run_build_script)
-                    reset_machine_id(args, workspace.name, run_build_script)
                     install_boot_loader(args, workspace.name)
                     install_extra_trees(args, workspace.name)
                     install_build_src(args, workspace.name, run_build_script)
@@ -2273,6 +2272,7 @@ def build_image(args, workspace, run_build_script):
                     set_root_password(args, workspace.name, run_build_script)
                     run_postinst_script(args, workspace.name, run_build_script)
 
+                reset_machine_id(args, workspace.name, run_build_script)
                 make_read_only(args, workspace.name)
 
             squashfs = make_squashfs(args, workspace.name)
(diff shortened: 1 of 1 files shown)


AsciiWolf:
Yep, this PR is much better than my quick fix. Thanks!

mkosi
```diff
@@ -937,6 +937,10 @@ def reset_machine_id(args, workspace, run_build_script):
         machine_id = os.path.join(workspace, 'root', 'etc/machine-id')
         os.unlink(machine_id)
         open(machine_id, "w+b").close()
+        dbus_machine_id = os.path.join(workspace, 'root', 'var/lib/dbus/machine-id')
+        if os.path.exists(dbus_machine_id):
+            os.unlink(dbus_machine_id)
+            os.symlink('../../../etc/machine-id', dbus_machine_id)
```
poettering: hmm, smells like TOCTOU... I think it would be much nicer if we'd just unlink the thing, and if that results in a file not found exception we should just eat that up...

fsateler: It seems reasonable to me to assume we are the only ones modifying the image, since we unshared the mount namespace and thus nobody else would be seeing this file. Moreover, I did not want to create the symlink if the file did not exist in the first place.

If you still prefer the unlink unconditionally approach, I can post a follow up PR, but I don't think it is necessary.

poettering: true, it's pretty safe to assume that. Whenever I see something like this it immediately raises my eyebrows though, and I'd always avoid any raised eyebrows for this, given how simply it is.

And yes, I think it makes sense not to create the symlink if the file doesn't exist anyway. But that's orthogonal to whether we unlink first or not:

```python
unlinked = False
try:
    os.unlink(dbus_machine_id)
    unlinked = True
except FileNotFoundError:
    pass

if unlinked:
    os.symlink(....)
``` 
or something along those lines

fsateler: OK, I'll post a follow up PR soon.

PR #65, opened 2017-02-26
Add options to run scripts with network access

lucaswerkmeister:
(Note: this PR is based on #64.)

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -975,7 +975,7 @@ def run_postinst_script(args, workspace, run_build_script):
         shutil.copy2(args.postinst_script,
                      os.path.join(workspace, "root", "root/postinst"))
 
-        run_workspace_command(args, workspace, "/root/postinst", "build" if run_build_script else "final")
+        run_workspace_command(args, workspace, "/root/postinst", "build" if run_build_script else "final", network=args.with_network)
         os.unlink(os.path.join(workspace, "root", "root/postinst"))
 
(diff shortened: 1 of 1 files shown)


lucaswerkmeister:
Note: this was intended as a fix to systemd/systemd#5324, but apparently doesn’t fix that issue – I still get the error even without `--private-network` – so I should probably investigate that some more before this is merged.

keszybz:
If it works correctly, this should be merged anyway. It useful functionality, even if it doesn't fix your issue. Do you plan to amend this patch, or is it OK as is?

poettering:
hmm, do we we really need the ability to switch on networking for postinst and build independently? I'd prefer if we could keep the exposed settings minimal, hence a single boolean enabling it in both is a lot more attractive to me...

keszybz:
Right, a single `--with-network` that applies to the whole build...

lucaswerkmeister:
Well, I thought requiring network access in `postinst` would probably be even rarer than in `build`, so `--build-with-network` ought to be enough for most projects… but I can rework this to a single `--with-network` too, sure.

lucaswerkmeister:
Alright, new version pushed. Here’s a simple `mkosi.build` script to test the change, by the way:
```bash
#!/bin/bash

exec 3<>/dev/udp/8.8.8.8/53
```
`sudo mkosi` should fail, `sudo mkosi --with-network` shouldn’t.

PR #66, opened 2017-02-28
Add support for openSUSE distros

fbuihuu:


mkosi
```diff
@@ -912,12 +915,105 @@ SigLevel    = Required DatabaseOptional
...
+        "tumbleweed" : "http://download.opensuse.org/%s/repo/oss/",
+        "42"         : "http://download.opensuse.org/distribution/leap/%s/repo/oss/",
+        "13"         : "http://download.opensuse.org/distribution/%s/repo/oss/",
+    }
+
+    opensuse_update_repository_map = {
+        "tumbleweed" : "http://download.opensuse.org/update/%s/",
+        "42"         : "http://download.opensuse.org/update/leap/%s/oss/",
+        "13"         : "http://download.opensuse.org/update/%s/",
+    }
```
poettering: hmm, any chance this can be done more future-proof? i.e. if "14" appears that we don't have to update mkosi?

mkosi
```diff
@@ -912,12 +915,105 @@ SigLevel    = Required DatabaseOptional
...
+
+    if extra_packages:
+        subprocess.run(cmdline + extra_packages, check=True)
+
+    #
+    # Disable packages caching in the image that was enabled
+    # previously to populate the package cache.
+    #
+    subprocess.run(["zypper", "--root", root, "modifyrepo", "-K", "Main"], check=True)
+    subprocess.run(["zypper", "--root", root, "modifyrepo", "-K", "Updates"], check=True)
```
poettering: for the fedora one I tried to invoke dnf as little as possible. Can't you merge the four calls to zypper above, or doe they really have to be done separately?

poettering:
looks good to me, just two questions

lucaswerkmeister:
I think the README could use a few more changes (unfortunately, GitHub won’t let me comment on lines outside the direct diff range):

- Any distribution that packages `zypper` may be used to build *openSUSE* images.
- Currently, *Fedora* packages all four tools. (@poettering: does it?)
  - If Fedora does package `zypper`, the package should also be added to the list under Requirements.

poettering:
> Currently, Fedora packages all four tools. (@poettering: does it?)

It doesn't AFAICS. only arch and debian/ubuntu may be bootstrapped with packaged tools afaics.

mkosi
fsateler: Alternatively, for these two and the first addrepo call, couldn't they be replaced with writing a text file?

mkosi
fbuihuu: Hmm 13.x is legacy and there won't ever be "14". New stuff will probably be a Leap distro so we could make it the default. I'm reworking this part.

mkosi
fbuihuu: @poettering, I don't think it's possible to merge all calls to zypper in one call, but I'm far from a zypper expert so I can be wrong. OTOH I'm not sure why you want to limit the number of calls as it will definitively makes the code harder to parse in this case.

fbuihuu:
@lucaswerkmeister  I'll update the README to include your first suggestion, thanks.

fbuihuu:
BTW, I worked on a similar project a couple years ago.

You might be interested: https://github.com/fbuihuu/installer

mkosi
poettering: well, my original intention with mkosi was really to patch around in the generated images as little as possible, and hence to minimize the active foreign code run in inside of it.

In an ideal world we'd not even run any code inside of the container environment for building things, and would just extract RPMs directly. Unfortunately though RPM doesn't work like that, and postinsts snippets and so on force us to run things in the container env while building. I still think so we should absolutely minimize how often we do that, and hence do as little namespace transitions as we can

fbuihuu: sorry I can't think of a way to shrink the num of calls to zypper and creating manually the repo config files and then later patching them seems really not a good plan to me and it's a different way to patch the generated image anyways.

fbuihuu:
please find a new attempt.

PR #67, opened 2017-02-28
reset_machine_id: ask for forgiveness, not permission

fsateler:
Avoid TOCTOU by always deleting and handling the FileNotFoundError

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1027,8 +1027,11 @@ def reset_machine_id(args, workspace, run_build_script):
         os.unlink(machine_id)
         open(machine_id, "w+b").close()
         dbus_machine_id = os.path.join(workspace, 'root', 'var/lib/dbus/machine-id')
-        if os.path.exists(dbus_machine_id):
+        try:
             os.unlink(dbus_machine_id)
+        except FileNotFoundError:
+            pass
+        else:
             os.symlink('../../../etc/machine-id', dbus_machine_id)
 
 def set_root_password(args, workspace, run_build_script):


mkosi
```diff
@@ -1027,8 +1027,14 @@ def reset_machine_id(args, workspace, run_build_script):
...
         dbus_machine_id = os.path.join(workspace, 'root', 'var/lib/dbus/machine-id')
-        if os.path.exists(dbus_machine_id):
+        has_dbus_machine_id = False
+        try:
             os.unlink(dbus_machine_id)
+            has_dbus_machine_id = True
+        except FileNotFoundError:
+            pass
+
+        if has_dbus_machine_id:
```
keszybz: OK. But why not use `try..else`, and get rid of the boolean?

fsateler: I'm not sure what the `else` would be for. I could lump the symlink into the same try block though..

```python
try:
  os.unlink(dbus_machine_id)
except FileNotFoundError:
  pass
else:
  os.symlink('../../../etc/machine-id', dbus_machine_id)
```

poettering: i figure that was my bad suggestion. I forgot that you can do the "else" syntax when catching exceptions...

fsateler: Actually, what I want is to create the symlink only if the file already existed. I do not want to create it if we didn't remove it. That would imply creating the `/var/lib/dbus` dir if it doesn't exist, which I don't think we should be doing. That leaves us with either this PR or:

```python
try:
  os.unlink(dbus_machine_id)
  os.symlink('../../../etc/machine-id', dbus_machine_id)
except FileNotFoundError:
  pass
```

keszybz: My code does what you described.

poettering: @fsateler no. the "else" branch of a "try" statement is only executed when no exception is thrown. The difference between the version you are now proposing and @keszybz's is that in your version  "os.symlink()" throwing FileNotFoundError is ignored, while in @keszybz's version it is not. And I think @keszybz' version is right here: we should only ignore the errors we really want to ignore, and that's the ones from os.unlink().

fsateler: Oh, its clear I did not know how the `else` clause works. Force pushed a new version using that.

Thanks for the patience!

PR #71, opened 2017-03-15
Use https for openSUSE repo urls

AsciiWolf:


AsciiWolf:
@lucaswerkmeister To be honest, I don't know. The HTTPS repos seem to work now, but you are right it could be a problem in the future.

keszybz:
With Let's Encrypt, it's hard not to be able to afford a certificate.

lucaswerkmeister:
Sure, but currently they’re on DigiCert, which afaik is not free.

keszybz:
Right, but you can switch at any time. And in this day and age, HTTPS is the norm. I'm pretty sure we should change all http to https wherever supported. And where unsupported, we should ask for it to be supported.

PR #80, opened 2017-05-09
use correct name for initramfs (not initrd)

cmtm:


poettering:
uh oh. "initramfs" isn't any better. After all this is implemented as "tmpfs" file system these days, not as "ramfs". 

We discussed this amongsome people a while back and came to the conclusion that inittmpfs would be worse than either initrd and initramfs, even though the latter two are slightly misleading. We hence agreed to simply stick to the original name, which is "initrd", and be done with it.

I hope that makes sense!

PR #85, opened 2017-05-18
support for incremental image building

poettering:
Let's speed up building images, for usage in source trees such as systemd's, by caching the image version right before copying in the project source or build tree.


mkosi
```diff
@@ -1642,7 +1770,7 @@ def parse_args():
     group = parser.add_argument_group("Output")
     group.add_argument('-t', "--format", dest='output_format', choices=OutputFormat.__members__, help='Output Format')
     group.add_argument('-o', "--output", help='Output image path', metavar='PATH')
-    group.add_argument('-f', "--force", action='store_true', help='Remove existing image file before operation')
+    group.add_argument('-f', "--force", action=ForceAction, nargs=0, default=False, help='Remove existing image file before operation')
```
keszybz: Wouldn't it be possible to use `action='count'` instead?

.gitignore
```diff
@@ -1,15 +1,18 @@
+*.cache1
+*.cache2
```
keszybz: What about somewhat more descriptive names? `.cache-pre-dev`, `.cache-pre-inst`? I think this would be more feature-proof, in case different stages are added.

I don't grok the meaning of `-f`. I'd expect that `-i` means that the cached images will be used. Why do you need to specify `-f` in addition? Also, `-ff` negates `-f`, which is counter-intuitive.

Looking at the bigger picture, it seems that the incremental builds should become the default (just because they're much more pleasant). We have the following (orthogonal) choices: 1. use cached images if available or force regeneration, 2. leave cached images around for the next run or not, 3. remove cached images or let them be. I think `--incremental` shoudl control option 1., not sure about the rest. But `--force` seems to be a bad fit for controlling 2. and 3.

mkosi
```diff
@@ -550,13 +607,20 @@ def prepare_tree(args, workspace, run_build_script):
             pass
 
     if args.output_format in (OutputFormat.subvolume, OutputFormat.raw_btrfs):
+
+        if cached and args.outputformat is Outputformat.raw_btrfs:
```
keszybz: typo

mkosi
poettering: ah, i wasn't aware that existed. Will fix

.gitignore
poettering: currently, running "mkosi" without "-f" (nor "-i") means: build if output.raw doesn't exist yet, error out if "output.raw" already exists. running "mkosi" with "-f" (but without "-i") means: build output.raw, possibly removing it first if it already exists.

Now, the new "-i" doesn't change that behaviour: whether -f is specified once or no times just has the effect whether output.raw is removed first or not. However if you now specify it twice then we'll not only remove output.raw but also output.raw.cache1 and output.raw.cache2.

Summary:

 1. mkosi -i → neither delete/replace output.raw nor output.raw.cache[12], use the latter if they exist
 2. mkosi -if → delete/replace output.raw, but not output.raw.cache[12], use the latter if they exist
 3. mkosi -iff → delete/replace output.raw and output.raw.cache[12], don't use the latter

Does that make sense?

poettering:
> Looking at the bigger picture, it seems that the incremental builds should become the default (just because they're much more pleasant).

I am not sure about that. Yes, "-i" is useful for developer style usage (i.e. me, the systemd hacker, wanting to try out his new code change), but for all other uses it's not the top choice. A CI system, an embedded developer, and so on should not use it. Moreover it has some weird semantics regarding reproducability: if you use it the distro version becomes "frozen in time" until you remove the cache files... It's not an option without side effect hence, and I am not sure that's a good thing... Moreover the way it is implemented right now we won't automatically detect whether mkosi.defaults (or any other of the mkosi.xyz files) has changed, and hence using the incremental mode means you really need to know how to use it and when to rebuild in full.

> We have the following (orthogonal) choices: 1. use cached images if available or force regeneration, 2. leave cached images around for the next run or not, 3. remove cached images or let them be. I think --incremental shoudl control option 1., not sure about the rest. But --force seems to be a bad fit for controlling 2. and 3.

`--incremental` currently doesn't control 1. actually (or not alone) the way I implemented it right now. Instead:

`--incremental` controls whether to use/generate cache files at all. If not specified we get the exact same behaviour as we always did. --force then controls how much we "destroy" while building: nothing, the final output, or the final output and the cache files at the same time. Now, some artificial possibilities are not covered by this: i.e. the option of using but not generating cache files, or the option of removing cache files but not the output files. But I think that's fine, as I couldn't find any realistic usecase where you'd want that...

does that make sense?

poettering:
I force pushed a new version now. I left the "-i" and "-f" behaviour in place but fixed all other things. Please have a look

poettering:
will force push a new version shortly with your coding style suggestions fixed

poettering:
done, please have a look!

Issue #79, opened 2017-04-26
mkosi fails on Python 3.4


eliasp:
`mkosi` can't be executed on Python 3.4-based distributions (e.g. current Debian stable 8.7 which only ships 3.4.2) due to various Python 3.5 specific items:

- [Line 636](https://github.com/systemd/mkosi/blob/0910ed5dbc3b089703b4a888d58ab96c4bace5f7/mkosi#636):
    **Code:**
    ```
    cmdline += ['--', *cmd]
    ```
    **Error**: SyntaxError: can use starred expression only as assignment target
- [Line 724](https://github.com/systemd/mkosi/blob/0910ed5dbc3b089703b4a888d58ab96c4bace5f7/mkosi#L724)
    **Code:**
    ```
    "--disablerepo=*",
    *repos,
    "--setopt=keepcache=1",
    ```
    **Error**: SyntaxError: can use starred expression only as assignment target
- [Line 125](https://github.com/systemd/mkosi/blob/0910ed5dbc3b089703b4a888d58ab96c4bace5f7/mkosi#L125)
    **Code:**
    ```
    subprocess.run(["mount", "--make-rslave", "/"], check=True)
    ```
    **Error**: AttributeError: 'module' object has no attribute 'run'
    **Note**: The `run()` method in `subprocess` was introduced in Python 3.5

Either `setup.py` needs to be adjusted to check for Python >= 3.5 (leaving Distributions without Python 3.5 unsupported) or the offending items need to be changed to also work with Python < 3.5.

keszybz:
The first two are rather trivial, and we could replace `run` with `check_call`. So it's probably worth the trouble to support Debian.

eliasp:
…on the other hand: Debian 8.x still has only systemd-215 - is this sufficient at all for what `mkosi` does using `systemd-nspawn`?

fsateler:
Moreover, debian stretch is coming out ["really soon now"](https://lists.debian.org/debian-devel-announce/2017/04/msg00013.html):

> At the time of writing, 143 RC bugs affected stretch. Just one of them is marked as a blocker and unfixed in sid

 I'd just make 3.5 required.

lucaswerkmeister:
Well, we’ve also wanted to release mkosi v2 for a while now (#68), so perhaps we could at least make this release Python 3.4-compatible.

keszybz:
[...]

PR #86, opened 2017-05-18
require python 3.5 as minimum version

poettering:
Fixes: #79

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -22,6 +22,9 @@ from enum import Enum
 
 __version__ = '1'
 
+if sys.version_info < (3, 5):
+    sys.exit("Sorry, we need at least Python 3.5.")
+
 # TODO
 # - volatile images
 # - make debian/ubuntu images bootable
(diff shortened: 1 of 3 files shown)


mkosi
```diff
@@ -22,6 +22,9 @@ from enum import Enum
 
 __version__ = '1'
 
+if sys.version_info[1] < 5:
```
keszybz: That's unpythonic, e.g. python 2.7 passes that test. (And a hypothetical python 4.0 would also fail the test...). I'd prefer a check like:
```python
if sys.version_info < (3, 5):
```

poettering: well, we specify "python3" in the shebang header, hence the major is already checked, i only checked the minor...

I am not sure I fully grok tuple comparisons in python3, for example, is (3, 5, 0) < (3, 5) or is (3, 5) < (3, 5, 0)? I am not keen on using expressions whose full effect I really don't grok...

keszybz: They are sorted in "dictionary" order, i.e. pairs of elements in the same position are compared, and if they are all equal, the longer sequence is later. This comparison is a standard Python idiom, in particular used quite often for versions ;)

poettering:
posted a new version now. Please have a look

PR #94, opened 2017-06-22
Optionally ignore all untracked files in git ls-files

keszybz:
It is very easy for a large file to end up in the built image. Either
the package cache, or some other already built image, or the new
*.cache-* files are large enough to overflow the available space in
the destination image.

Current approach of blacklisting certain files with --exclude just
doesn't work: it depends on file naming in the project being built.
For example we had '--exclude=/.mkosi*', which in systemd excludes the
various mkosi configuration files which should be distributed. On the
other hand, writing an exclude filter for all possible image and cache
names is infeasible.

If somebody is adding new files to the project being built, they can
always use 'git add -N' to include them in the mkosi image without
adding them to the staging area.

This is the missing piece I need to get #85 to work without jumping through hoops.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1145,9 +1145,11 @@ def install_extra_trees(args, workspace):
         for d in args.extra_trees:
             enumerate_and_copy(d, os.path.join(workspace, "root"))
 
-def copy_git_files(src, dest):
-    c = subprocess.run(['git', 'ls-files', '-z', '--others', '--cached',
-                        '--exclude-standard', '--exclude', '/.mkosi-*'],
+def copy_git_files(src, dest, *, git_files):
+    what_files = ['--exclude-standard', '--cached']
+    if git_files == 'others':
+        what_files += ['--others']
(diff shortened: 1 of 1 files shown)


poettering:
Hmm, any chance we can make this an option rather than the default? I regularly put test files in my devel tree and want them to show up in the mkosi image so that I can test them

keszybz:
And what about `git add -N`, wouldn't that work for you?

poettering:
> And what about git add -N, wouldn't that work for you?

Well, sure, but I think we should add everything by default, already to keep the behaviour between git and non-git source dirs minimal... I mean, the fact mkosi can take benefit of git magic doesn't mean it should totally change its default semantics, if you follow what I mean?

mkosi is after all also intended to be used without any source tree at all, just by invoking it on some mkosi.default file...

keszybz:
Hm, I never used `mkosi` outside of a repo, so I'm biased.

Anyway, updated to make it optional.

PR #98, opened 2017-06-28
don't patch nsswitch if resolve is already present

### LINKED ISSUE #97: don't patch nsswitch on Arch guest
Arch already comes with nss_resolve enabled
https://git.archlinux.org/svntogit/packages.git/tree/trunk/nsswitch.conf?h=packages/filesystem

the patching in 
https://github.com/systemd/mkosi/blob/master/mkosi#L678
results in `hosts: files mymachines resolve [!UNAVAIL=return] resolve myhostname`

Ideally it should not patch nsswitch if resolve is already enabled.

gdamjan:
the default nsswitch in ArchLinux (and probably other distros in future)
already has:

`hosts: files mymachines resolve [!UNAVAIL=return] dns myhostname`

so don't rewrite `dns` with `resolve` as it then becomes:

`hosts: files mymachines resolve [!UNAVAIL=return] resolve myhostname`

fixes #97

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -666,6 +666,13 @@ def patch_file(filepath, line_rewriter):
     os.remove(filepath)
     shutil.move(temp_new_filepath, filepath)
 
+def fix_hosts_line_in_nsswitch(line):
+    if line.startswith("hosts:"):
+        sources = line.split(" ")
+        if 'resolve' not in sources:
+            return " ".join(["resolve" if w == "dns" else w for w in sources])
+    return line
+
 def enable_networkd(workspace):
     subprocess.run(["systemctl",
                     "--root", os.path.join(workspace, "root"),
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -675,8 +675,14 @@ def enable_networkd(workspace):
...
     os.symlink("../usr/lib/systemd/resolv.conf", os.path.join(workspace, "root", "etc/resolv.conf"))
 
-    patch_file(os.path.join(workspace, "root", "etc/nsswitch.conf"),
-               lambda line: " ".join(["resolve" if w == "dns" else w for w in line.split(" ")]) if line.startswith("hosts:") else line)
+    def fix_hosts_in_nsswitch(line):
+        if line.startswith("hosts:"):
+            sources = line.split(" ")
+            if 'resolve' not in sources:
+                return " ".join(["resolve" if w == "dns" else w for w in sources])
+        return line
```
keszybz: There's no reason for this to be an nested function. Please move it to the top level.

If there's nothing to do, it'd be nicer to just do nothing, instead of rewriting the file with identical contents.

gdamjan: I'll move it top-level sure.

To not rewrite the file if not needed it would need two passes I guess? 
One to check if it's there, second to rewrite

gdamjan:
@keszybz to not touch the file if not needed the patch_file would have to be changed.
Open another issue for that perhaps?

keszybz:
Looking at it again, I'm not sure if the optimization is worth the trouble. Other uses of `patch_file` should always make some changes, so optimizing just for `/etc/nsswitch.conf` probably isn't important enough.

PR #100, opened 2017-06-28
Adjust note about Fedora and package manager tools

Conan-Kudo:
Since Fedora 26, all four tools are available in Fedora, as Zypper was added for Fedora 26.


lucaswerkmeister:
Merged, thanks!

PR #102, opened 2017-06-29
Fix FileNotFound error while running with -i

slaykovsky:
Hi!

I've tried to run mkosi like this:

`mkosi -d fedora -t directory -o fedora -b false -i`

And I've got a traceback instead.

So this change fixes it.

Signed-off-by: Aleksei Slaikovskii <aslaikov@redhat.com>

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1649,7 +1649,7 @@ def calculate_signature(args, checksum):
 
 def save_cache(args, workspace, raw, cache_path):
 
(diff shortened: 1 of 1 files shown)


keszybz:
What was the traceback?

The patch looks wrong, because the cache can be used not only with raw images.

slaykovsky:
@keszybz yes, totally wrong. Never mind :)

PR #105, opened 2017-06-30
Added optional hostname assignment

slaykovsky:


mkosi
```diff
@@ -557,6 +557,26 @@ def mount_image(args, workspace, loopdev, root_dev, home_dev, srv_dev, root_read
...
+    with open(hostname_path, "w+") as f:
+        f.write("{}\n".format(args.hostname))
+
+    if os.path.isfile(hosts_path):
+        with open(hosts_path, "a") as f:
+            f.write("127.0.0.1\t{}\n".format(args.hostname))
+    else:
+        with open(hosts_path, "w+") as f:
+            f.write("127.0.0.1\tlocalhost\n")
+            f.write("127.0.0.1\t{}\n".format(args.hostname))
```
keszybz: There's no need to write `/etc/hosts`, `nss-myhostname` or `nss-resolved` cover this better.

gdamjan:
I'd change the commit message that's it's an optional assignment :)

mkosi
```diff
@@ -1814,6 +1835,7 @@ def parse_args():
     group.add_argument('-C', "--directory", help='Change to specified directory before doing anything', metavar='PATH')
     group.add_argument("--default", dest='default_path', help='Read configuration data from file', metavar='PATH')
     group.add_argument("--kernel-commandline", help='Set the kernel command line (only bootable images)')
+    group.add_argument("--hostname", help="Set hostname")
```
poettering: this really should have a counterpart in the mkosi.default file I think, like most our options have.

mkosi
```diff
@@ -557,6 +557,26 @@ def mount_image(args, workspace, loopdev, root_dev, home_dev, srv_dev, root_read
 
             umount(root)
 
+@complete_step("Assigning hostname")
```
poettering: btw, we don't use Signed-Off-By on commits to our projects, that's a kernel thing

mkosi
```diff
@@ -557,6 +557,17 @@ def mount_image(args, workspace, loopdev, root_dev, home_dev, srv_dev, root_read
...
+def assign_hostname(args, workspace):
+    if not args.hostname:
+        return
+
+    root = os.path.join(workspace, "root")
+    hostname_path = os.path.join(root, "etc/hostname")
+    
+    with open(hostname_path, "w+") as f:
```
dteleguin: If an image is used with `systemd-nspawn`/`machinectl`, having static hostname will prevent `systemd` from setting transient hostname to machine name - it will always remain e.g. `localhost` despite the `-M` flag.

If this is by design (@poettering ?), then we might probably need to unset static hostname (i.e. remove `/etc/hostname`) if not specified explicitly. In Fedora and Mageia, `/etc/hostname` for "localhost" is installed by default via the `systemd` RPM package.

fsateler: Debootstrap will try to copy `/etc/hostname` from the host, so this affects debian-based distros too.

poettering: > If this is by design (@poettering ?)
Yes, this is by design.

And yupp, we should remove /etc/hostname if none is set, I am very sure. Would love to see this PR updated for that.

slaykovsky:
Sorry, didn't have a time, will try to find some this week.

mkosi
slaykovsky: Done

slaykovsky:
@poettering @keszybz guys, could you please take a look?

mkosi
dteleguin: I think the logic should be the following:
* if neither `--hostname` option nor `Hostname` config key are given, do nothing (leave whatever has been installed as per `/etc/hostname`);
* if either `--hostname` or `Hostname` are specified with an empty value, reset hostname (i.e. remove `/etc/hostname`);
* if a non-empty value is given, set the hostname via `/etc/hostname`.

@poettering WDYT?

mkosi
```diff
@@ -564,6 +564,18 @@ def mount_image(args, workspace, loopdev, root_dev, home_dev, srv_dev, root_read
...
+@complete_step("Assigning hostname")
+def assign_hostname(args, workspace):
+    root = os.path.join(workspace, "root")
+    hostname_path = os.path.join(root, "etc/hostname")
+    
+    if not args.hostname:
+        if os.path.isfile(hostname_path):
+            os.remove(hostname_path)
+    else:
+        with open(hostname_path, "w+") as f:
```
poettering: given that there's a good chance that /etc/hostname is already a symlink to /run/systemd/resolve/resolv.conf I think it would make sense to explicitly remove the file before we write to it, otherwise the open() would just follow the symlink and that's not what we want here...

mkosi
poettering: I think we should remove /etc/hostname if it is a regular file, the way the patch currently does it... That way we can take benefit of nspawns and systemd's hostname logic in a much nicer way: nspawn will initialize the container's hostname to the container's image/directory name (with suffixes removed), and if /etc/hostname isn#t there, then systemd will leave it that way, which is a good thing. If we'd however leave /etc/hostname in place, in particular with a default value of "localhost" or so, then this nspawn feature will never have any effect...

Also note that when systemd boots up and sees the hostname uninitialized, it will make sure it is initialized, and for that has a fallback built-in of "localhost" (but this can be changed by distros if they like at compile time, and fedora for example does, and changes this to "fedora"). Hence, leaving in /etc/hostname that only carries a default value is not only redudant in all cases but also hurtful in others...

Hence I think the way the patch does it right now is great.

(But yeah, you do have a point, and I actually would agree with you, if distros would get things right, and not drop in /etc/hostname by default in the first place. If they wouldn't there would of course be no point in deleting it...)

mkosi
slaykovsky: Thanks! You're right, I didn't thought about symlink :)

poettering:
BTW, please consider merging your three patches into a single one, to make this easier to review. We usually try to merge PRs in a "perfect" form, i.e. where they consist of individual logical commits, not commits that reflect development history, if you understand what I mean. 

Note that this isn't too important, I can also wield them into a single one when I merge this, but usually it's nicer to already have them in the PR that way...

mkosi
dteleguin: /etc/hostname is a symlink to /run/systemd/resolve/resolv.conf? WAT? :)

slaykovsky: @dteleguin didn't make attention to this one yesterday :) Yes, that's not right

I think @poettering just was doing some different business by the time of comment :) But anyway I agree to remove /etc/hostname in any cases (when nothing is given and before assignment).

poettering:
thanks! merged!

mkosi
poettering: uh, oh, i indeed was doing other stuff then... Sorry, I was very confused... Anyway the new patch looks great, and I merged it now.

slaykovsky:
Thank you!

PR #106, opened 2017-06-30
RFE: add "mkosi.skeleton/" subdir concept, which is copied into the OS tree before dnf and friends are invoked

gdamjan:
I'd like to allow for a custom `pacman.conf` when setting up an Arch image.

What should the override file be named? 

- [ ] `pacman,conf`
- [ ] `mkosi.pacman`
- [ ] `mkosi.pacman.conf`
- [ ] something else?

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -911,6 +911,28 @@ def install_debian(args, workspace, run_build_script):
 def install_ubuntu(args, workspace, run_build_script):
     install_debian_or_ubuntu(args, workspace, run_build_script, args.mirror)
 
+def make_pacman_conf(args, server):
+    return """\
+[options]
+LogFile     = /dev/null
+HookDir     = /no_hook/
+HoldPkg     = pacman glibc
+Architecture = auto
+UseSyslog
+Color
+CheckSpace
+SigLevel    = Required DatabaseOptional
+
+[core]
+{server}
+
+[extra]
+{server}
+
+[community]
+{server}
+""".format(args=args, server=server)
+
 @complete_step('Installing Arch Linux')
 def install_arch(args, workspace, run_build_script):
     if args.release is not None:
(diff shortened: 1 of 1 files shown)


keszybz:
Hm, wouldn't it be enough to put it in `mkosi.extra/`?

gdamjan:
no,
this is for the initial pacstrap that operates on the host (before the guest is even a thing). 

that pacman.conf is put in /var/tmp/<temp-dir>/pacman.conf

gdamjan:
OTOH. 
I could check for `mkosi.extra/etc/pacman.conf` and use it here?
that way the same pacman.conf file would be used for both the pacstrap and in the image.

gdamjan:
ps.
anyone knows what the idea was behind `HookDir     = /no_hook/` ?

poettering:
I'd very much prefer if we could keep this generic, with as little per-distro hooks as we can get away with. 

`mkosi.extra` is copied over the image after the package manager ran. I figure we could add a new dir `mkosi.skeleton` or so, which permits definining a directory tree that is used as skeleton, i.e. before we invoke the package manager. THen you could drop your pacman.conf in there, but everybody else could benefit of that too!

Does that make sense?

gdamjan:
So for Arch the file would be `mkosi.skeleton/pacman.conf` and it would go to the `/var/tmp/mkosi-XXXXXX/` directory?

Then the generated pacman.conf would need to be checked in the tmp dir perhaps?

That's fine by me. But I don't know what the other distros need.

poettering:
> mkosi.skeleton/pacman.conf

It would be `mkosi.skeleton/etc/pacman.conf` I figure, since that config file should end up below `/etc/`, no?

poettering:
> That's fine by me. But I don't know what the other distros need.

Well, not sure what they need either, but I do see benefit in having that generic, and I think it could be useful, not just for dropping in files pre-package-manager, but also for adjusting perms of some dirs before the package manager takes possossion of them or so.

gdamjan:
that would work I guess.

OTOH, pacstrap doesn't care where pacman.conf is. And for putting pacman.conf in the image, we have mkosi.extra.

Tangential. 
I'm thinking of actually just installing `pacman` with `pacstrap` and then installing the rest in `mkosi.postinst`. Reason beeing even the mkosi limited package list is not enough. Dunno if this is someting of upstream interest.

poettering:
Note that "postinst" is run *very* late, i.e. after mkosi.build or anything like that... (not sure if you care about that though)

lucasdemarchi:
> Well, not sure what they need either, but I do see benefit in having that generic, and I think it could be useful, not just for dropping in files pre-package-manager, but also for adjusting perms of some dirs before the package manager takes possossion of them or so.

Agreed. On Fedora this could be used to enable other repositories for example... copr comes to mind.
@gdamjan are you going to respin this PR or could I take a look on it?

gdamjan:
@lucasdemarchi please take it

I'm kind of strapped with time currently anyway

fsateler:
FWIW, this would be useful for debian as well. While debootstrap does not take a config file, apt does, and we use apt to install any extra packages, so the no-config part only applies to the bare minimum.

PR #109, opened 2017-07-02
Fix zypper calls (opensuse)

sebix:
Tested on Tumbleweed.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1028,7 +1028,7 @@ def install_opensuse(args, workspace, run_build_script):
     #
     # Install the "minimal" package set.
     #
-    subprocess.run(cmdline + ["-t", "pattern", "minimal_base"], check=True)
+    subprocess.run(cmdline + ["patterns-base-minimal_base"], check=True)
 
     #
     # Now install the additional packages if necessary.
--- a/mkosi
+++ b/mkosi
@@ -2287,7 +2287,7 @@ def load_args():
             if platform.machine() == "aarch64":
                 args.mirror = "http://mirror.archlinuxarm.org"
         elif args.distribution == Distribution.opensuse:
-            args.mirror = "https://download.opensuse.org"
+            args.mirror = "http://download.opensuse.org"
 
     if args.bootable:
         if args.distribution == Distribution.ubuntu:


mkosi
```diff
@@ -2287,7 +2287,7 @@ def load_args():
             if platform.machine() == "aarch64":
                 args.mirror = "http://mirror.archlinuxarm.org"
         elif args.distribution == Distribution.opensuse:
-            args.mirror = "https://download.opensuse.org"
+            args.mirror = "http://download.opensuse.org"
```
keszybz: https works fine on Fedora: c.f. #6231. What error are you getting?

poettering:
can you explain what the precise issue?

@fbuihuu can you comment on this patch?

mkosi
sebix: It's this one:
```
‣ Installing openSUSE...
Adding repository 'Main' ...............................................................................................................................[done]
Repository 'Main' successfully added

URI         : https://download.opensuse.org/tumbleweed/repo/oss/
Enabled     : Yes                                               
GPG Check   : Yes                                               
Autorefresh : No                                                
Priority    : 99 (default priority)                             

Repository priorities are without effect. All enabled repositories share the same priority.
['zypper', '--root', '/var/tmp/mkosi-zsj8e3kx/root', 'addrepo', '-ck', 'https://download.opensuse.org/tumbleweed/repo/oss/', 'Main']
Adding repository 'Updates' -------------------------------------------------------------------------------------------------------------------------------[\]
Download (curl) error for 'https://download.opensuse.org/update/tumbleweed/repodata/repomd.xml':
Error code:  Bad URL
Error message: Protocol "http" not supported or disabled in libcurl

Abort, retry, ignore? [a/r/i/...? shows all options] (a): 
```

mkosi
```diff
@@ -1028,7 +1028,7 @@ def install_opensuse(args, workspace, run_build_script):
     #
     # Install the "minimal" package set.
     #
-    subprocess.run(cmdline + ["-t", "pattern", "minimal_base"], check=True)
+    subprocess.run(cmdline + ["patterns-base-minimal_base"], check=True)
```
fbuihuu: It looks like  we lost the visibility flag. I reported this issue and it should be fixed soon.

So this change isn't correct IMHO and you should have reported this issue to the openSUSE community instead.

mkosi
fbuihuu: Running `zypper --root /tmp/mkosi-zsj8e3kx/root addrepo -ck https://download.opensuse.org/tumbleweed/repo/oss/ Main` here (TW) works fine.

So there might be something wrong with your setup.

Could you try to run the same command and if it indeed fails then ask some help to the openSUSE community ?

mkosi
poettering: So you are saying the issues this PR intends to fixed are located somewhere else, and we can close this PR hence?

fbuihuu: I think so.

poettering: OK, closing this one then, as I understand you pinged the right people.

sebix: > I reported this issue and it should be fixed soon.

Do you have a link to a bugreport or something else? I don't have the necessary pattern on TW.

fbuihuu: no sorry, I reported the issue internally and apparently it's not fixed yet. OTOH, I would suggest you to open a bug if you want to track it down.

sebix: Done here: https://bugzilla.opensuse.org/show_bug.cgi?id=1049997

sebix: This has been fixed now.

keszybz: Great, thanks.

mkosi
sebix: @fbuihuu See this thread for an explanation why this does not work: https://forums.opensuse.org/showthread.php/527052-Cached-files-in-home-cache-kioexec-krun-_-How-to-avoid?p=2838516#post2838516

mkosi
juaningan: I continuing to need this patch for both leap and tumbleweed

mkosi
sebix: @fbuihuu And also see this very detailed post: https://progress.opensuse.org/news/40

mkosi
sebix: @juaningan 

For TW I don't see this problem:
```
> zypper se -s -t pattern minimal
Loading repository data...
Reading installed packages...

S | Name         | Type    | Version       | Arch   | Repository             
--+--------------+---------+---------------+--------+------------------------
i | minimal_base | pattern | 20170319-15.1 | x86_64 | openSUSE-Tumbleweed-Oss
v | minimal_base | pattern | 20170319-15.1 | i586   | openSUSE-Tumbleweed-Oss
```
The package in Virtualization/mkosi still has this patch in btw because at the time v3 has been released that was not fixed.

And is Leap supported by mkosi?

juaningan: Same error for both releases:

```
# mkosi -d opensuse -r tumbleweed
# mkosi -d opensuse -r 42.3
```
```
(...)
Building repository 'Main' cache .....................................................................................................................................................................[done]
Building repository 'Updates' cache ..................................................................................................................................................................[done]
Loading repository data...                                                                            
Reading installed packages...                      
'minimal_base' not found in package names. Trying capabilities.                                                                                                                                             No provider of 'minimal_base' found.
‣ Unmounting Package Cache...                      
‣ Unmounting image...                              
‣ Unmounting image complete.                       
‣ Detaching image file...                          
  File "/home/juanjop/mkosi/mkosi", line 3390, in <module>                                            
    main()                                         
  File "/home/juanjop/mkosi/mkosi", line 3380, in main                                                
    build_stuff(args)                              
  File "/home/juanjop/mkosi/mkosi", line 3284, in build_stuff                                         
  File "/home/juanjop/mkosi/mkosi", line 3133, in build_image                                         
    install_distribution(args, workspace.name, run_build_script, cached)                              
  File "/home/juanjop/mkosi/mkosi", line 1442, in install_distribution                                
    install[args.distribution](args, workspace, run_build_script)                                     
  File "/usr/lib/python3.6/contextlib.py", line 52, in inner                                          
    return func(*args, **kwds)                     
  File "/home/juanjop/mkosi/mkosi", line 1381, in install_opensuse                                    
    run(cmdline + ["-t", "pattern", "minimal_base"], check=True)                                      
  File "/usr/lib/python3.6/subprocess.py", line 418, in run                                           
    output=stdout, stderr=stderr)                  
subprocess.CalledProcessError: Command '['zypper', '--root', '/var/tmp/mkosi-9ww544b1/root', '--gpg-auto-import-keys', 'install', '-y', '--no-recommends', '-t', 'pattern', 'minimal_base']' returned non-zero exit status 104.
```

keszybz:
OK, it seems that we should merge at least the first one.
> Fix opensuse repo URL

Based on  https://progress.opensuse.org/news/40, https works sometimes, depending on the mirror you get, but doesn't othertimes (because some mirrors use http, and downgrading from https to http does not work). Also, there are no plans to enable official https support.

> Fix opensuse pattern installation 

This is fixed in TW, but not in other versions. Maybe we should just merge this patch for now, just to make things work ootb, and revisit this after the pattern syntax is supported everywhere.

sebix:
Thanks for reopening. I merged in master and fixed the conflict.

sebix:
The URL has already been fixed in the meantime in 2421b2b3 (that was btw 3 months after I proposed it here - that would have "prevented" #137 and #173).

keszybz:
Part of the reason is that https://github.com/systemd/mkosi/commit/2421b2b3c17a609a75d45672def27174462dbdc0 has a much better commit message.

PR #112, opened 2017-07-10
Mageia Linux support

dteleguin:
Current limitations:

- No support for `--read-only` until Mageia gets systemd-233;
- for bootable images: due to old dracut, the kernel command line should contain explicit root definition, e.g. `--kernel-commandline="root=LABEL=root"`. Mageia 6 will get updated dracut soon after release;
- if building an image with LVM2 support (`--packages lvm2`), the boot process will hang for ~2min in initrd due to dracut bug (Mageia specific). To fix this, boot into an image and regenerate unified EFI kernel with `dracut --hostonly ...` To avoid generating OpenSSH keys, polluting `/var/log` etc. boot into single user mode (use EFI boot screen);
- Mageia doesn't (yet) publish GPG keys on the web, so in order to install Mageia from a non-Mageia Linux one needs to have `/etc/pki/rpm-gpg/RPM-GPG-KEY-Mageia`. Not sure how to deal with this; I'll ask Mageia guys to publish the key, and if it's not possible, we can either disable GPG checks or automate key retrieval from the MIT server:
```
gpg --keyserver pgp.mit.edu --recv-keys 80420F66
```

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -846,17 +846,13 @@ def install_mageia(args, workspace, run_build_script):
 #        gpg_key = "https://getfedora.org/static/%s.txt" % FEDORA_KEYS_MAP[args.release]
 
     if args.mirror:
-        baseurl = "{args.mirror}/distrib/{args.release}/x86_64/media/core/release/".format(args=args)
-        if not check_if_url_exists("%s/media.repo" % baseurl):
-            baseurl = "{args.mirror}/distrib/cauldron/x86_64/media/core/release/".format(args=args)
-
-        release_url = "baseurl=%s" % baseurl
-        updates_url = "baseurl={args.mirror}/distrib/{args.release}/x86_64/media/core/updates/".format(args=args)
+        baseurl = "{args.mirror}/distrib/{args.release}/x86_64/media/core/".format(args=args)
+        release_url = "baseurl=%s/release/" % baseurl
+        updates_url = "baseurl=%s/updates/" % baseurl
     else:
-        release_url = ("mirrorlist=https://www.mageia.org/mirrorlist/?" +
-                       "release={args.release}&arch=x86_64&section=core&repo=release".format(args=args))
-        updates_url = ("mirrorlist=https://www.mageia.org/mirrorlist/?" +
-                       "release={args.release}&arch=x86_64&section=core&repo=updates".format(args=args))
+        baseurl = "https://www.mageia.org/mirrorlist/?release={args.release}&arch=x86_64&section=core".format(args=args)
+        release_url = "mirrorlist=%s&repo=release" % baseurl
+        updates_url = "mirrorlist=%s&repo=updates" % baseurl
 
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -832,6 +833,99 @@ gpgkey={gpg_key}
...
+               "passwd"])
+
+    if args.packages is not None:
+        cmdline.extend(args.packages)
+
+    if run_build_script and args.build_packages is not None:
+        cmdline.extend(args.build_packages)
+
+    if args.bootable:
+        cmdline.extend(["kernel", "binutils"])
```
Conan-Kudo: There are several possible `kernel` flavors that could be picked. I'd suggest picking a flavor and sticking with it here. Probably `kernel-server-latest` is the right choice here, as it works the best across a wide variety of targets.

dteleguin:
@Conan-Kudo good catch with `kernel`, thx! also our `basesystem-minimal` works much better than the semi-copypasted set of Fedora packages.
@poettering I think we should wait a bit until the situation with GPG keys clarifies.

mkosi
```diff
@@ -832,6 +833,99 @@ gpgkey={gpg_key}
...
+
+        # Temporary hack: dracut only adds crypto support to the initrd, if the cryptsetup binary is installed
+        if args.encrypt or args.verity:
+            cmdline.append("cryptsetup")
+
+        if args.output_format == OutputFormat.raw_gpt:
+            cmdline.append("e2fsprogs")
+
+        if args.output_format == OutputFormat.raw_btrfs:
+            cmdline.append("btrfs-progs")
```
poettering: hmm, any chance we can at least unify the actual dnf command invocation? i.e. add a new function invoke_dnf() or so, which takes the distro-specific stuff as argument, and unifies the stuff shared between fedora + mageia otherwise?

poettering:
looks pretty OK to me, but please, let's add a new helper function for invoking dnf, shared between the fedora and mageia codepaths

mkosi
dteleguin: @poettering sure we can factor out the DNF stuff. Do you think we should move DNF config generation into invoke_dnf() too? we can pass config data as a Python dictionary, for eaxmple:
```
config = dict(
    gpg = True,
    repos = [
        dict(
            id = 'mageia',
            name = 'Mageia Core Release',
            url ='...',
            gpgkey = '...'
        )
     ]
)
```
Otherwise we could leave config generation in the install_* functions.

poettering: dunno, up to you. I wouldn't overarchitect things, and whatever is shorter usually wins in my eyes... not sure whether pressng everything into a dict to immediately decode it from that again would shorten the code or make it longer...

dteleguin: OK, let the things be simple for now, I'll leave config generation outside invoke_dnf(). Anyway, we can revisit this in the future.

poettering:
BTW, you might want to add a brief mention of Mageia support to README.md, too

PR #114, opened 2017-07-12
mkosi: support reusing builddirs between subsequent mkosi runs

poettering:
With this change, mkosi will look for "mkosi.builddir" when invoked.
When this directory exists it is mounted into the build container and
may be used as location for out-of-tree builds, which can be shared
between multiple invocations to speed up the building process.

This is particularly useful when used in conjunction with incremental
building (-i), as with this in place building OS images using mkosi is
only a little slower than just building the project natively on the host
now.

"mkosi clean -f" will empty the builddir if it exists now too, in
addition to the incremental image trees.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ they exist in the local directory:
   is loaded in the same manner adding/overriding the values specified in
   `mkosi.default`.
 
-* `mkosi.extra` may be a directory. If this exists all files
+* `mkosi.extra/` may be a directory. If this exists all files
   contained in it are copied over the directory tree of the
   image after the *OS* was installed. This may be used to add in
   additional files to an image, on top of what the
@@ -189,8 +189,23 @@ they exist in the local directory:
   next to image files it boots, for additional container
   runtime settings.
 
-* `mkosi.cache` may be a directory. If so, it is automatically used as
+* `mkosi.cache/` may be a directory. If so, it is automatically used as
   package download cache, in order to speed repeated runs of the tool.
+
+* `mkosi.builddir/` may be a directory. If so, it is automatically
+  used as out-of-tree build directory, if the build commands in the
+  `mkosi.build` script support it. Specifically, this directory will
+  be mounted into the build countainer, and the `$BUILDDIR`
+  environment variable will be set to it when the build script is
+  invoked. The build script may then use this directory as build
+  directory, for automake-style or ninja-style out-of-tree
+  builds. This speeds up builds considerably, in particular when
+  `mkosi` is used in incremental mode (`-i`): not only the disk images
+  but also the build tree is reused between subsequent
+  invocations. Note that if this directory does not exist the
+  `$BUILDDIR` environment variable is not set, and it is up to build
+  script to decide whether to do in in-tree or an out-of-tree build,
+  and which build directory to use.
 
 * `mkosi.passphrase` may be a passphrase file to use when LUKS
   encryption is selected. It should contain the passphrase literally,
--- a/mkosi
+++ b/mkosi
@@ -1845,6 +1845,7 @@ def parse_args():
     group.add_argument("--extra-tree", action='append', dest='extra_trees', help='Copy an extra tree on top of image', metavar='PATH')
     group.add_argument("--build-script", help='Build script to run inside image', metavar='PATH')
     group.add_argument("--build-sources", help='Path for sources to build', metavar='PATH')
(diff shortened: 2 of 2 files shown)


poettering:
Here's a PR that adds support for this to systemd:

https://github.com/systemd/systemd/pull/6349

mkosi
```diff
@@ -2640,6 +2667,10 @@ def run_build_script(args, workspace, raw):
         else:
             cmdline.append("--chdir=/root")
 
+        if args.build_dir is not None:
+            cmdline.append("--setenv=BUILDDIR=/root/build")
```
fsateler: Shouldn't this be always set, even if no build_dir? This would break `$BUILDDIR`-using scripts if one accidentally nukes `mkosi.builddir`

poettering: well, i wanted that scripts can use the fact whether $BUILDDIR is set or not to decide whether they want to do an in-tree or out-of-tree build, and I think that's really relevant, as doing out-of-tree builds is often broken in projects, as it's "new-style" and some projects aren't up for that yet...

I'd claim that removing mkosi.builddir/ is fine if it has the same effects as removing mkosi.default: i.e. break stuff...

fsateler: > i wanted that scripts can use the fact whether $BUILDDIR is set or not to decide whether they want to do an in-tree or out-of-tree build, 

That can still be decided by `[ "`realpath $BUILDDIR`" = "`realpath .`" ]`.

> I'd claim that removing mkosi.builddir/ is fine if it has the same effects as removing mkosi.default: i.e. break stuff...

Right, but then you have to make sure the dir always exist in your git repo via a .gitkeep file or such, and gitignore the rest. If BUILDDIR were always set, builds would just not be cached.
It's not the end of the world, but it is an annoyance.

poettering: Hmm, so what would you suggest we set $BUILDDIR to if mkosi.builddir isn't set?

I mean, meson & co don't support in-tree builds, hence for them you'd probably want to se BUILDDIR=$SRCDIR/build or something like that... But automake supports in-tree builds just fine, hence then you'd want to set BUILDDIR=$SRCDIR... Hence, what would mkosi do? force things to be out-of-tree? for things to be in-tree then? I am sure we'd break half of the existing projects by either choice...

Hence I think it's best to leave this choice to the build scripts. For example in https://github.com/systemd/systemd/pull/6349 I opted for setting BUILDDIR=$SRCDIR/build if $BUILDDIR isn't set. But I am also sure for an automake project I'd want to do this differently...

Moreover, the check involving realpath isn't really pretty at all I must say...

fsateler: > Hence, what would mkosi do? force things to be out-of-tree? for things to be in-tree then? I am sure we'd break half of the existing projects by either choice...

Indeed, that sounds like a good argument against setting it unconditionally. Maybe this should be documented in the README though. Appending "`$BUILDDIR` will not be set if this directory is not present" or so to make it explicit should be enough I think.

README.md
```diff
@@ -187,9 +187,20 @@ they exist in the local directory:
...
   runtime settings.
 
-* `mkosi.cache` may be a directory. If so, it is automatically used as
+* `mkosi.cache/` may be a directory. If so, it is automatically used as
   package download cache, in order to speed repeated runs of the tool.
 
+* `mkosi.builddir/` may be a directory. If so, it is automatically
+  used as out-of-tree build directory, if the build commands in the
+  `mkosi.build` script support it. Specifically, this directory will
+  be mounted into the build countainer, and the $BUILDDIR environment
```
fsateler: Enclose $BUILDDIR in backquotes.

README.md
```diff
@@ -187,9 +187,20 @@ they exist in the local directory:
...
+  variable will be set to it when the build script is invoked. The
+  build script may then use this directory as build directory, for
+  automake-style or ninja-style out-of-tree builds. This speeds up
+  builds considerably, in particular when `mkosi` is used in
+  incremental mode (`-i`): not only the disk images but also the build
+  tree is reused between subsequent invocations.
```
fsateler: As mentioned in the discussion, I think explicitly noting that `$BUILDDIR` will not be set if the dir is empty would be nice.

poettering:
Excellent! Thanks for the review! I have now force pushed a new version, only with the documentation changes suggested.

poettering:
given the positive review of the code bits, will merge now. thanks!

PR #115, opened 2017-07-14
Ignore temporary directories while copying

lucasdemarchi:
If we call mkosi passing --build-sources with a directory that contains
the temporary .mkosi-* directories, it will fall in a inifinite
recursion.

This is the case in which we are have the mkosi.* files in a
subdirectory of the git repository and pass the git root dir as argument
to --build-sources.


lucasdemarchi:
There's also the argument that in this case it should use git, but mkosi runs `git ls-files` in the CWD rather than where the source is. Maybe we could also force the autodetection to check for .git in the directory pointed by build-sources?

PR #116, opened 2017-07-15
Always turn workspace/root into a mount point

op3:
pacstrap wants workspace/root to be a mountpoint (see e. g. [this bug report](https://bugs.archlinux.org/task/46137)). At the moment, this is only the case when using `Format=raw_gpt,` `raw_btrfs` or `subvolume`. When using `Format=tar`, `directory` or `raw_squashfs`, this is not the case, which causes pacstrap to manually execute `mount $workspace/root $workspace/root --bind`. As this is not done recursively, any directory mounted below workspace/root, such as the package cache (from mkosi.cache or Cache=), is ignored and therefore unavailable, effectively disabling the package cache function of mkosi.

This commit preemptively turns workspace/root into a mountpoint, preventing such problems.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -607,7 +607,7 @@ def prepare_tree(args, workspace, run_build_script, cached):
         btrfs_subvol_create(os.path.join(workspace, "root"))
     else:
         try:
-            os.mkdir(os.path.join(workspace, "root"))
+            mount_bind(os.path.join(workspace, "root"), os.path.join(workspace, "root"))
         except FileExistsError:
             pass
 


poettering:
looks OK, but could you add a brief comment to the relevant lines that explains this? Otherwise the next one stumbling over that line might drop the mount again, because it otherwise is unnecessary.

lucasdemarchi:
Unfortunately it doesn't really work.

```sh
‣ Linking image file...
  File "/bin/mkosi", line 2865, in <module>
    main()
  File "/bin/mkosi", line 2861, in main
    build_stuff(args)
  File "/bin/mkosi", line 2824, in build_stuff
    tar.name if tar is not None else None)
  File "/bin/mkosi", line 1731, in link_output
    os.rename(os.path.join(workspace, "root"), args.output)
OSError: [Errno 16] Device or resource busy: '/home/lucas/p/mavlink-router/tests/osi/.mkosi-zc11wg0y/root' -> '/home/lucas/p/mavlink-router/tests/osi/rootfs'
  File "/usr/lib64/python3.6/weakref.py", line 624, in _exitfunc
    f()
  File "/usr/lib64/python3.6/weakref.py", line 548, in __call__
  File "/usr/lib64/python3.6/tempfile.py", line 797, in _cleanup
    _shutil.rmtree(name)
  File "/usr/lib64/python3.6/shutil.py", line 480, in rmtree
  File "/usr/lib64/python3.6/shutil.py", line 422, in _rmtree_safe_fd
    onerror(os.rmdir, fullname, sys.exc_info())
  File "/usr/lib64/python3.6/shutil.py", line 420, in _rmtree_safe_fd
    os.rmdir(name, dir_fd=topfd)
OSError: [Errno 16] Device or resource busy: 'root'
```

This is an alternative approach, moving it to mount_cache():
https://github.com/lucasdemarchi/mkosi/commit/524b993e2572de46473cd73c5db00d8a51833f6a

PR #119, opened 2017-07-17
mkosi: add support for doing CentOS images

poettering:
This is very similar to Fedora/Mageia support, but uses yum rather than
dnf, as CentOS doesn't have that by default.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -57,6 +57,7 @@ class Distribution(Enum):
     arch = 4
     opensuse = 5
     mageia = 6
+    centos = 7
 
 GPT_ROOT_X86           = uuid.UUID("44479540f29741b29af7d131d5f0458a")
 GPT_ROOT_X86_64        = uuid.UUID("4f68bce3e8cd4db196e7fbcaf984b709")
@@ -522,6 +523,7 @@ def mount_loop(args, dev, where, read_only=False):
     subprocess.run(["mount", "-n", dev, where, options], check=True)
 
 def mount_bind(what, where):
+    os.makedirs(what, 0o755, True)
     os.makedirs(where, 0o755, True)
     subprocess.run(["mount", "--bind", what, where], check=True)
 
@@ -590,6 +592,10 @@ def mount_cache(args, workspace):
     with complete_step('Mounting Package Cache'):
         if args.distribution in (Distribution.fedora, Distribution.mageia):
             mount_bind(args.cache_path, os.path.join(workspace, "root", "var/cache/dnf"))
+        elif args.distribution == Distribution.centos:
+            # We mount both the YUM and the DNF cache in this case, as YUM might just be redirected to DNF even if we invoke the former
+            mount_bind(os.path.join(args.cache_path, "yum"), os.path.join(workspace, "root", "var/cache/yum"))
+            mount_bind(os.path.join(args.cache_path, "dnf"), os.path.join(workspace, "root", "var/cache/dnf"))
         elif args.distribution in (Distribution.debian, Distribution.ubuntu):
             mount_bind(args.cache_path, os.path.join(workspace, "root", "var/cache/apt/archives"))
         elif args.distribution == Distribution.arch:
@@ -600,7 +606,7 @@ def mount_cache(args, workspace):
         yield
     finally:
         with complete_step('Unmounting Package Cache'):
-            for d in ("var/cache/dnf", "var/cache/apt/archives", "var/cache/pacman/pkg", "var/cache/zypp/packages"):
+            for d in ("var/cache/dnf", "var/cache/yum", "var/cache/apt/archives", "var/cache/pacman/pkg", "var/cache/zypp/packages"):
                 umount(os.path.join(workspace, "root", d))
 
 def umount(where):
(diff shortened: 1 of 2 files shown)


mkosi
```diff
@@ -882,6 +888,78 @@ gpgkey={gpg_key}
...
+gpgkey={gpg_key}
+""".format(args=args,
+           gpg_key=gpg_key,
+           release_url=release_url,
+           updates_url=updates_url))
+
+    repos = ["--enablerepo=" + repo for repo in (args.repositories if args.repositories else ["base", "updates"])]
+
+    root = os.path.join(workspace, "root")
+    cmdline = ["yum",
```
dteleguin: As the command line for yum and dnf is identical, I think we could use invoke_dnf() here by introducing a parameter with default value:
```
def invoke_dnf(args, workspace, repositories, base_packages, boot_packages, cmd="dnf"):
```

mkosi
```diff
@@ -882,6 +888,78 @@ gpgkey={gpg_key}
...
+               "--config=" + os.path.join(workspace, "yum.conf"),
+               "--best",
+               "--allowerasing",
+               "--releasever=" + args.release,
+               "--installroot=" + root,
+               "--disablerepo=*",
+               *repos,
+               "--setopt=keepcache=1",
+               "--setopt=install_weak_deps=0"]
+
```
dteleguin: As the command line for yum and dnf is identical, I think we could use invoke_dnf() here by introducing a parameter with default value:
```
def invoke_dnf(args, workspace, repositories, base_packages, boot_packages, cmd="dnf"):
```

Conan-Kudo: It's not quite the same. I know that `install_weak_deps` isn't the correct option here. I think for Yum, it's `requires_policy`, but I don't recall what the params are.

poettering:
I have reworked this now so that we use the same function for both yum and dnf. Maybe we have to special case yum sooner or later, but for now this should be good enough.

Please have a look!

mkosi
```diff
@@ -747,14 +753,14 @@ def disable_kernel_install(args, workspace):
...
     repos = ["--enablerepo=" + repo for repo in repositories]
 
     root = os.path.join(workspace, "root")
-    cmdline = ["dnf",
+    cmdline = [command,
                "-y",
-               "--config=" + os.path.join(workspace, "dnf.conf"),
+               "--config=" + os.path.join(workspace, command + ".conf"),
                "--best",
                "--allowerasing",
```
Conan-Kudo: `--best` and `--allowerasing` do not exist for Yum. Not setting the weakdeps policy for DNF makes the images much fatter, too.

dteleguin: Perhaps my suggestion to merge yum/dnf was a bit premature :-\

Conan-Kudo:
@poettering You're also going to want to detect the `yum-deprecated` binary and use it if it exists for your `yum` function. Both Fedora and Mageia ship `yum` packages with the binary named `yum-deprecated`.

poettering:
OK, I force pushed a new version which uses "dnf" if it exists to install centos, and reverts to yum if it doesn't. I tested this against yum-deprecated to make sure it works. (The tool will never look for yum-deprecated though)

willronchetti:
Which CentOS version are you testing this on? I'm running a minimal CentOS 7 install and this does not appear to work on my machine. Running 'mkosi' fails to detect the distribution, and specifying CentOS (with -d) fails at the sfdisk command, as sfdisk is run with the option '--color=never' which supposedly does not exist.

poettering:
@willronchetti i didn't test this on centos. I tested building centos images on fedora, that's all. I think it would be useful to make mkosi work on centos as well, but that's kinda outside of the scope of this PR so far.

willronchetti:
Which version of Fedora are you using? I'm on Fedora 26 (x86_64) and am finding that there is a problem with dnf, as evidenced in the following error messages I'm seeing (trying to create CentOS image).

...
Mounting API VFS complete.
Repository updates is listed more than once in the configuration
Failed to synchronize cache for repo 'base', disabling.
Failed to synchronize cache for repo 'updates', disabling.
No package systemd available.
No package system-release available.
No package passwd available.
...
Error: Unable to find a match

Relevant dnf command is on line 991, but I suspect the root of the problem is not there. Dnf appears to work fine for me as I've installed the dependencies and my internet is working.

willronchetti:
@poettering Ah, I've found my issue is that I'd like to build a UEFI bootable CentOS image. A regular CentOS image works just fine. Spent some time trying to build in support for bootable CentOS with no luck. Any plans to support that in the future?

poettering:
> @poettering Ah, I've found my issue is that I'd like to build a UEFI bootable CentOS image. A regular CentOS image works just fine. Spent some time trying to build in support for bootable CentOS with no luck. Any plans to support that in the future?

RHEL/CentOS don't bring the necessary bits to my knowledge. Hence, it's not really up to me to support that, we'd need sd-boot to be available in centos/rhell somehow before this could be made working

mkosi
```diff
@@ -884,9 +893,106 @@ gpgkey={gpg_key}
...
+gpgkey={gpg_key}
+""".format(args=args,
+           gpg_key=gpg_key,
+           release_url=release_url,
+           updates_url=updates_url))
+
+    invoke_dnf_or_yum(args, workspace,
+                      args.repositories if args.repositories else ["base", "updates"],
+                      ["systemd", "centos-release", "passwd"],
+                      ["kernel", "systemd-udev", "binutils"],
```
NicolasT: Trying to build a CentOS 7 image on a Fedora 29 host. This `systemd-udev` package can't be found in the (default) CentOS repositories, so image building fails. Where should this package be available from?

godfryd: I have the same problem. What about this missing systemd-udev package for centos?

PR #122, opened 2017-07-17
Some assorted additions, including new "mkosi shell" and "mkosi boot" commands

poettering:


mkosi
```diff
@@ -1945,8 +1956,11 @@ def unlink_try_hard(path):
 
 def empty_directory(path):
 
-    for f in os.listdir(path):
-        unlink_try_hard(os.path.join(path, f))
+    try:
+        for f in os.listdir(path):
+            unlink_try_hard(os.path.join(path, f))
+    except:
```
fsateler: Swallowing all exceptions is a code smell. Why is it a good idea to presume everything worked when it didn't?

mkosi
```diff
@@ -2830,25 +2865,46 @@ def check_root():
...
         die("Must be invoked as root.")
 
+def run_shell(args):
+    target = "--directory=" + args.output if args.output_format in (OutputFormat.directory, OutputFormat.subvolume) else "--image=" + args.output
+
+    cmdline = ["systemd-nspawn",
+               target]
+
+    if args.verb == "boot":
+        cmdline += ('--boot',)
```
fsateler: It would be nice to have an option to run another command instead of `/bin/sh` (which is the very limited dash in debian-based systems)

mkosi
```diff
@@ -2830,25 +2865,46 @@ def check_root():
     if os.getuid() != 0:
         die("Must be invoked as root.")
 
+def run_shell(args):
+    target = "--directory=" + args.output if args.output_format in (OutputFormat.directory, OutputFormat.subvolume) else "--image=" + args.output
+
+    cmdline = ["systemd-nspawn",
```
fsateler: I would add `--as-pid2`, `--quiet` and possibly `--register=no`, to prevent conflicts with other instances.

mkosi
keszybz: Also, `except:` should almost never be used. It's a Python gotcha, soemthing that seems natural, but is the wrong thing to do in almost all circumstances, because it catches InterruptExcetion, so ^C doesn't work. If you want to catch "everything", `except Exception` is the way to go. (The exception to this rule is interactive shells, where you **need** to catch ^C, and return to the prompt.)

mkosi
poettering: I thought about those, and explicitly decided not to use them:

 1. `--as-pid2` is the right choice for most commands, but not so much for a shell, as shells generally have some basic support in place for being run as PID 1. Moreover, running them as PID 1 has the benefit that we can later on start the whole system, by doing "exec /usr/lib/systemd/system". Or in other words: if we run the shell this way it is fully equivalent to init=/bin/sh on a physical system, and that's a good thing:
2. Regarding `--quiet`: this is an interactive shell, hence I think we really should tell the user how he can exit it. I mean, there's a reason systemd-nspawn is by default not quiet: so that people know how to exit the thing, and this applies in this case as much as it applies for the general nspawn usecase.
3. Regarding `--register=no`: This is an interactive boot, and I think it's a really good thing that we can connetc from the outside to it. It's at least the way I usually work here: do something on the console of the container, and then get another shell in it via "machinectl shell" so that i can invoke strace or gdb or whatever else to supervise what I am doing there. By doing `--register=no` we'd take that away.

Or to say this all differntly: this is really about interactive shells, and the usecase is debugging: "please give me a quick shell in what i am working on here". The three options you suggested are the right choice for automatic build processes (and that's why we use them there), but not for these commands. I hope that makes sense?

mkosi
poettering: Yeah, we should add that eventually. I didn't want to do that for now, since doing arbitrary commands should really imply --as-pid2, while invoking an interactive shell shouldn't (see my comments above), and I didn't want to get into that mess... UNIX semantics on this aren't pretty, and we already failed to do this right on nspawn itself, as --as-pid2 is not the implied default for non-interactive stuff...

poettering:
I force pushed a new version now. It changes the try/except line to be more strict. I also changed "shell" and "boot" to pass the remaining command line on to systemd-nspawn. This means you can now invoke "mkosi -i shell /bin/bash" if you like, and it will do the right thing.

I didn't change any of the nspawn params though, for the reasons explained above.

poettering:
@fsateler any chance you can give this a second look?

mkosi
fsateler: > Regarding --register=no: This is an interactive boot, and I think it's a really good thing that we can connetc from the outside to it. It's at least the way I usually work here: do something on the console of the container, and then get another shell in it via "machinectl shell" so that i can invoke strace or gdb or whatever else to supervise what I am doing there. By doing --register=no we'd take that away.

Hmm. A problem is that this makes it impossible to run this command simultaneously on different images in the default output path, since you can't have two machines with the same name `image`

(https://github.com/systemd/systemd/issues/6347 would prevent this even with `--register=no` but we can fix that one in systemd)

mkosi
```diff
@@ -2011,6 +2026,10 @@ def unlink_output(args):
             with complete_step('Clearing out build directory'):
                 empty_directory(args.build_dir)
 
+        if args.cache_path is not None:
+            with complete_step('Clearing out package cache'):
+                empty_directory(args.cache_path)
```
fsateler: I'm not so sure about this one. Unlike the build dir and the mkosi cache, packages are uniquely named: either you want file foo_something.deb or you don't: the contents of foo_something.deb cannot change (and if they do, the relevant apt/rpm command would reject the invalid signature). At least in the deb world, that holds true, and I think it holds true in the rpm world too. I have no idea about arch.

Therefore, I think cleaning the package cache should not be mixed with clearing mkosi stuff or build stuff.

poettering:
OK, I force pushed a new version now that addresses @fsateler point. Can you take another look? I now moved the package cache deletion logic under an extra `--force`.

mkosi
lucaswerkmeister: Arch package file names also contain the version and should therefore be unique as well. On the other hand, hiding this behind two/three `--force`s seems acceptable to me.

PR #123, opened 2017-07-18
Use git clone

lucasdemarchi:


poettering:
Hmm, can this even work correctly?

It was my intention that "mkosi -if" can be used when hacking arbitrary projects to run whatever is in the current working directory. That means all untracked files should be copied too. Or to say this differently: .gitignore should be used as the only blacklist for copying, and everything that isn't blacklisted needs to be copied over. Unless I am mistaken using "git clone" would mean we'd only copy what is already added to the git archive, right? i.e. if i create a so far untrack new source file and want to quickly test it, it would not be included in the copy, right? I am not sure I like that I must say...

lucasdemarchi:
See that in the code I added there's also a call to git ls-files and another to diff. This way modified, added and untracked files can be copied in addition to the clone.

lucasdemarchi:
I'm doing some corner cases tests to ensure it works. Already got one bug.

fsateler:
@lucasdemarchi AFAICS this explicitly only copies files known to git and not untracked ones... you should add `--others` to the ls-files invocation to include those.

lucasdemarchi:
I force pushed some changes:
- We can't pass `--local` to git clone if the output is not of type directory since it would cross filesystem binary. Now we just don't pass the option and let git figure it out (it does)
- Fixed some corner cases with git-ls-files or git-diff returning an empty list
- Fixed the --git-files=others switch to automatically ignore our temporary files

lucasdemarchi:
@fsateler this was already the case before. If you were using `--use-git-files=yes` you'd need to pass `--git-files=others` in order to copy untracked files.

lucasdemarchi:
@poettering While doing this I thought about an alternative: we could bind mount the src directory and even make it readonly. Could this be an option?

poettering:
> @poettering While doing this I thought about an alternative: we could bind mount the src directory and even make it readonly. Could this be an option?

This was suggested before, but I'd like to avoid that, since none of the build artifacts that might be lying around should end up in the container I think (simply because the container pretty likely has a different distro than the host?), and vice versa. After all, in-tree building is still pretty common, and even if we do out-of-tree building like meson/ninja want you to, most people place the build artifacts in a subdir "build" that shouldn't appear there either...

Hence I am pretty sure we should operate on a copy, and it should have everything .gitignore hides removed...

Does that make sense?

lucasdemarchi:
> Does that make sense?

humn... maybe. Still not entirely convinced an option to just bind mount the src dir would not be doable for the simple cases of using the newly introduced builddir.

In the scenario of using git, this PR should fix the cases pointed out in the commit message then.

mkosi
```diff
@@ -1289,15 +1289,29 @@ def install_extra_trees(args, workspace, for_cache):
...
     files = {x.decode("utf-8") for x in c.stdout.rstrip(b'\0').split(b'\0')}
 
+    # everything that's modified and about to be committed
+    c = subprocess.run(['git', '-C', src, 'diff', '--cached', '--name-only', '-z'],
+                       stdout=subprocess.PIPE,
+                       universal_newlines=False,
+                       check=True)
+    files |= {x.decode("utf-8") for x in c.stdout.rstrip(b'\0').split(b'\0')}
+    files.discard('')
+    print(files)
```
poettering: do you really intent to print this here? or is this just debug left-over?

lucasdemarchi: Debug left-over, thanks.

lucasdemarchi: Fixed with a force push.

PR #134, opened 2017-08-01
Use 'cp' instead of python shutil to copy tees

### LINKED ISSUE #132: mkosi -i raises OSError with Format=directory when trying to copy /var/run/minissdpd.sock from cache
To reproduce:

mkosi.default
```
[Distribution]
Distribution=debian
Release=stretch

[Output]
Format=directory

[Packages]
Packages=
```

mkosi.build
```
touch $DESTDIR/share/test
```
Checked out from master (40df6a3), ran:
```
mkosi -i
```
Traceback:
```
  File "./mkosi", line 2883, in <module>
    main()
  File "./mkosi", line 2879, in main
    build_stuff(args)
  File "./mkosi", line 2825, in build_stuff
    raw, tar, root_hash = build_image(args, workspace, run_build_script=True)
  File "./mkosi", line 2680, in build_image
    cached = reuse_cache_tree(args, workspace.name, run_build_script, for_cache, cached)
  File "./mkosi", line 2644, in reuse_cache_tree
    enumerate_and_copy(fname, os.path.join(workspace, "root"))
  File "./mkosi", line 1270, in enumerate_and_copy
    enumerate_and_copy(source, dest, suffix + "/" + entry.name)
  File "./mkosi", line 1270, in enumerate_and_copy
    enumerate_and_copy(source, dest, suffix + "/" + entry.name)
  File "./mkosi", line 1277, in enumerate_and_copy
    shutil.copy(entry.path, dest_path, follow_symlinks=False)
  File "/usr/lib/python3.5/shutil.py", line 241, in copy
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/lib/python3.5/shutil.py", line 120, in copyfile
    with open(src, 'rb') as fsrc:
OSError: [Errno 6] No such device or address: 'image.cache-pre-dev/var/run/minissdpd.sock'
```

Host os is debian sid.

DanielG:
Fix #132

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1259,24 +1259,8 @@ def install_boot_loader(args, workspace, cached):
         if args.distribution == Distribution.opensuse:
             install_boot_loader_opensuse(args, workspace)
 
-def enumerate_and_copy(source, dest, suffix = ""):
-    for entry in os.scandir(source + suffix):
-        dest_path = dest + suffix + "/" + entry.name
-
-        if entry.is_dir():
-            os.makedirs(dest_path,
-                        mode=entry.stat(follow_symlinks=False).st_mode & 0o7777,
-                        exist_ok=True)
-            enumerate_and_copy(source, dest, suffix + "/" + entry.name)
-        else:
-            try:
-                os.unlink(dest_path)
-            except:
-                pass
-
-            shutil.copy(entry.path, dest_path, follow_symlinks=False)
-
-        shutil.copystat(entry.path, dest_path, follow_symlinks=False)
+def enumerate_and_copy(source, dest):
+    subprocess.run(["cp", "--reflink=auto", "--recursive", "--no-dereference", "--preserve=all", "--no-target-directory", source, dest], check=True)
 
 def install_extra_trees(args, workspace, for_cache):
     if args.extra_trees is None:
--- a/mkosi
+++ b/mkosi
@@ -1329,7 +1329,7 @@ def install_build_src(args, workspace, run_build_script, for_cache):
             if use_git:
                 copy_git_files(args.build_sources, target, git_files=args.git_files)
             else:
-                ignore = shutil.ignore_patterns('.git', '.mkosi-*')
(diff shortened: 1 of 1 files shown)


dsmcfarl:
@DanielG I tested your branch. The use of cp works well and handles the device files. The only part of #132 that it did not fix was my comment about adding '*.cache-pre-*', to the ignore_patterns so it still hangs on install_build_src when it tries to copy /dev/full from image.cache-pre-dev which it should not be copying in the first place.

replacing line 1332 with:

    ignore = shutil.ignore_patterns('.git', '.mkosi-*', '*.cache-pre-*')

fixes that.

My only other suggestion (take it or leave it) is to replace:

   subprocess.run(["cp", "--reflink=auto", "-aPT", source, dest], check=True)

with:

    subprocess.run(["cp", "--reflink=auto", "--no-dereference", "--preserve=all", "--recursive", "--no-target-directory", source, dest], check=True)

to make it more explicit when reading the code.

If you'd rather me put up a pull request to address these, let me know.

DanielG:
Ah yes, I was using an out of tree output dir so I wasn't running into that particular problem. Should be fixed now.

poettering:
Hmpf, I am not convinced that shelling out to "cp" is really the best option. This is not supposed to be a shell script, but a proper python program, and we we shouldn't shell out to anything unless we have really really good reasons to (for example, mke2fs makes sense to shell out out...)

DanielG:
Unfortunately python's `shutil` just can't deal with special files such as named pipes and sockets at all, see [shutils.py:113](https://github.com/python/cpython/blob/master/Lib/shutil.py#L113), whereas `cp` does that very well.

poettering:
Hmm, so I thought about this. I still don't like it, but doing a proper cp implementation in python is indeed a lot of work, in particular if we want to make sure it can handle all the corner cases (ACLs, …) correctly.

Hence, let's merge this.

fthiery:
How about using rsync instead of cp ? Do you think rsync could cover all corner cases too ? Asking this because of #166

fthiery:
@DanielG would you mind testing your use case with https://github.com/systemd/mkosi/pull/169 ? It uses rsync over cp.

Thanks

PR #135, opened 2017-08-02
Don't swallow stderr of subprocesses

DanielG:
Swallowing stderr makes debugging unnecessarily hard and given that `mkosi` is already rather verbose this shouldn't be too much noise.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -150,7 +150,7 @@ def btrfs_subvol_create(path, mode=0o755):
     os.umask(m)
 
 def btrfs_subvol_delete(path):
-    subprocess.run(["btrfs", "subvol", "delete", path], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, check=True)
+    subprocess.run(["btrfs", "subvol", "delete", path], stdout=subprocess.DEVNULL, check=True)
 
 def btrfs_subvol_make_ro(path, b=True):
     subprocess.run(["btrfs", "property", "set", path, "ro", "true" if b else "false"], check=True)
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -176,7 +176,7 @@ def image_size(args):
 def disable_cow(path):
     """Disable copy-on-write if applicable on filesystem"""
 
-    subprocess.run(["chattr", "+C", path], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, check=False)
+    subprocess.run(["chattr", "+C", path], stdout=subprocess.DEVNULL, check=False)
```
poettering: this is swallowed for a reason: "+C" is not available on anything but btrfs at the moment, and we really don't want to see the error about it, as it's an optimization only anyway

mkosi
```diff
@@ -605,7 +605,7 @@ def mount_cache(args, workspace):
 
 def umount(where):
     # Ignore failures and error messages
-    subprocess.run(["umount", "-n", where], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+    subprocess.run(["umount", "-n", where], stdout=subprocess.DEVNULL)
```
poettering: we ignore errors on this one for a reason too: we call it for stuff that might very well not be a mount point, and we let umount figure that out, but are not interested in its error message...

mkosi
```diff
@@ -150,7 +150,7 @@ def btrfs_subvol_create(path, mode=0o755):
     os.umask(m)
 
 def btrfs_subvol_delete(path):
-    subprocess.run(["btrfs", "subvol", "delete", path], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, check=True)
+    subprocess.run(["btrfs", "subvol", "delete", path], stdout=subprocess.DEVNULL, check=True)
```
poettering: we invoke this on non-btrfs file systems, too, and we really don't want to show the error messages this causes to the user. If people run ext4 and this call hence necessarily fails they really shouldn't be alerted.

And do note that we try a number of different ways to remove a dir, one of them is this, and if it fails it doesn't really matter too much, we'll catch the error later on in a different way...

PR #136, opened 2017-08-02
Fix btrfs_subvol_delete by deleting recursively

DanielG:
`btrfs subvol delete` errors out when the subvolume still contains
anything so we have to delete the contents first. However the subvolumes
we have to delete usually themselves contain subvolumes so we can't just
use `rm -r`. What we do instead is go through the directory depth first
using `find` and then try to delete each entry with `rm -d` first and if
that fails check if the entry is a directory and if so assume it's a
subvolume and fall back to `btrfs subvol delete`.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -150,6 +150,7 @@ def btrfs_subvol_create(path, mode=0o755):
     os.umask(m)
 
 def btrfs_subvol_delete(path):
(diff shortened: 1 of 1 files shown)


poettering:
Again, this is not a shell script. I agree with the goal, but let's do this properly please and not just glue together shell commands.

Issue #139, opened 2017-08-22
Debian fails to build due to `apt-get` errors


kBite:
Debian fails to build due to `apt-get` errors. Interestingly Ubuntu does not.

---
`mkosi.default`
```
[Distribution]
Distribution=debian
Release=jessie

[Packages]
WithNetwork=true
```
`# mkosi --default mkosi.default -o debian-8`
```
DISTRIBUTION:
          Distribution: debian
               Release: jessie
                Mirror: http://httpredir.debian.org/debian

OUTPUT:
         Output Format: raw_gpt
                Output: /root/mkosi/debian-8/debian-8

PACKAGES:
         Build Sources: /root/mkosi/debian-8
  Scripts with network: yes

PARTITIONS:
        Root Partition: 1.0G
```
```
WARNING: The following packages cannot be authenticated!
  libcap-ng0 libdbus-1-3 libexpat1 dbus libpam-systemd
E: There are problems and -y was used without --force-yes
```

---
`mkosi.default`
```
[Distribution]
Distribution=debian
Release=jessie

[Packages]
WithNetwork=true
```
`# mkosi --default mkosi.default -o debian-8`
```
DISTRIBUTION:
          Distribution: debian
               Release: jessie
                Mirror: http://httpredir.debian.org/debian

OUTPUT:
         Output Format: directory
                Output: /root/mkosi/debian-8/debian-8

PACKAGES:
         Build Sources: /root/mkosi/debian-8
  Scripts with network: yes
```
```
...
WARNING: The following packages cannot be authenticated!                           
  libcap-ng0 libdbus-1-3 libexpat1 dbus libpam-systemd                             
E: There are problems and -y was used without --force-yes 
...
```
`mkosi.default`
```
[Distribution]
Distribution=debian
Release=stretch

[Packages]
WithNetwork=true
```
`# mkosi --default mkosi.default -o debian-9`
```
DISTRIBUTION:
          Distribution: debian
               Release: stretch
                Mirror: http://httpredir.debian.org/debian

OUTPUT:
         Output Format: directory
                Output: /root/mkosi/debian-9/debian-9

PACKAGES:

         Build Sources: /root/mkosi/debian-9
  Scripts with network: yes
```
```
...
WARNING: The following packages cannot be authenticated!                           
  libdbus-1-3 libexpat1 dbus libpam-systemd                                        
E: There were unauthenticated packages and -y was used without --allow-unauthenticated
...
```
---
It's reproducible with
[...]

PR #140, opened 2017-08-22
[WIP] allow apt-get to install unauthenticated packages

kBite:
There might be a better solution (see https://github.com/systemd/mkosi/issues/139#issuecomment-324106303)
---
As described in issue #139 Debian fails due to 'apt-get' errors which are caused by unauthenticated packages: libdbus-1-3 libexpat1 dbus libpam-systemd

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -970,7 +970,7 @@ def install_debian_or_ubuntu(args, workspace, run_build_script, mirror):
                     'path-exclude /usr/share/linda/*\n',
                 ])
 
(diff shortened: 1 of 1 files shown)


fsateler:
I don't think this is an appropriate change. Allowing unauthenticated packages is a massive security hole. It's best to work out why apt is rejecting the archive keys in the first place.

kBite:
I was thinking the same and found a hint; see https://github.com/systemd/mkosi/issues/139#issuecomment-324106303.

I'm closing this PR myself.

poettering:
OK, closing as suggested

PR #141, opened 2017-08-22
set new/current DebianGeoMirror

kBite:
To avoid problems deb.debian.org should be used.

httpredir.debian.org is discontinued since 2017-02-17; see:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=817155


fsateler:
In theory, httpredir should point to the same place as deb, but I figure we should not stick to the old name,

PR #147, opened 2017-09-06
Only run build script if present in args, fixes #143

### LINKED ISSUE #143: Traceback: TypeError: chmod: path should be string, bytes, os.PathLike or integer, not NoneType
I use mkosi git master (currently at commit a499812) on Fedora 26.

The traceback happens when running `mkosi -if`:
```
‣ Installing cache copy ...        
    main()                         
  File "/usr/bin/mkosi", line 2900, in main                           
    build_stuff(args)              
  File "/usr/bin/mkosi", line 2833, in build_stuff                    
    args.cache_pre_dev)            
  File "/usr/bin/mkosi", line 1751, in save_cache                     
    os.chmod(raw, 0o666 & ~args.original_umask)                       
TypeError: chmod: path should be string, bytes, os.PathLike or integer, not NoneType
```

My mkosi.default:
```
[Distribution]
Distribution=fedora
Release=26

[Output]
ReadOnly=yes

[Validation]
Password=something-else
```

fthiery:


fthiery:
I agree, but it will still crash with non raw formats. The problem is that cache_path seems to be statically filled (i.e. even if the build step was not completed).

```
def save_cache(args, workspace, raw, cache_path):

    if cache_path is None or raw is None:
        return

    with complete_step('Installing cache copy ',
                       'Successfully installed cache copy ' + cache_path):

        if args.output_format in (OutputFormat.raw_btrfs, OutputFormat.raw_gpt):
            os.chmod(raw, 0o666 & ~args.original_umask)
            shutil.move(raw, cache_path)
        else:
            shutil.move(os.path.join(workspace, "root"), cache_path)
```

fthiery:
While it doesn't crash anymore, the pre-inst image is not used as cache if no build script is used: the time needed is exactly the same with and without it (time mkosi -if)

fthiery:
What is the point of generating the cache images all the time ? I.e. if we ask for cache (mkosi -i) i would assume it would only be built once; instead, reuse-cache_image always returns if entering the cache generation, where i would have expected something like "build cache if not already present".

```def reuse_cache_image(args, workspace, run_build_script, for_cache):
    print("In reuse cache image")

    if not args.incremental:
        return None, False
    if for_cache:
        return None, False
    if args.output_format not in (OutputFormat.raw_gpt, OutputFormat.raw_btrfs):
        return None, False
```

fsateler:
I can't retest right now but I think cache usage is broken if there is no dev image. That is, if you remove the pre-dev image, no cache is used. I think the behavior you are mentioning is related to this.

fthiery:
Yes, image cache is completely broken if there is no build script present; i made a few changes to fix this. Cache images will only be generated once (they will need to be deleted if conditions such as BuildPackages change, not sure how to do that).

I tested my use case with and without a build script, seems to work, but now it fails to build without caching...

fthiery:
I believe it should work now; if anyone can test too, if it works for you i could merge all commits and resubmit a pull request to clean it.

I must say it's not easy to debug, some more logging could help and some variables could be made a little clearer.

fthiery:
I rebased the PR, please comment

fsateler:
Could you please unmerge the branch that fixes #153 ? Each PR should have a single purpose...

fthiery:
Sorry this is unintentional, i didn't make a dedicated branch. I'll try to fix this

fthiery:
I'm very sorry bit i will have to open a new pull request since i merged all the other workbranches into my master (i didn't plan to work on more issues). Sorry for the mess

fthiery:
Here it is https://github.com/systemd/mkosi/pull/155

PR #148, opened 2017-09-08
Support package cache on arch despite have no Release defined

### LINKED ISSUE #146: Unable to use package cache on Arch unless a bogus Release is set
Using the following mkosi.default, alongside to a cache folder mkosi.cache, running mkosi results in an error being related to the lack of release in the [Distribution] key.

```
$ mkdir mkosi.cache
$ cat mkosi.default
[Distribution]
Distribution=arch
$ mkosi
    main()
  File "/usr/bin/mkosi", line 2885, in main
    args = load_args()
  File "/usr/bin/mkosi", line 2403, in load_args
    find_cache(args)
  File "/usr/bin/mkosi", line 2269, in find_cache
    args.cache_path = "mkosi.cache/" + args.distribution.name + "~" + args.release
TypeError: must be str, not NoneType
```
Arch being a rolling distro, there is no such thing as a release (except maybe for the date ?) Checking https://github.com/systemd/systemd/blob/master/.mkosi/mkosi.arch does not reveal any specific syntax that would allow the use of the cache folder.

Putting in a random Release, like "Release=2017-09-06" does skip the aforementioned error, should we automatically try to fill the Release key of mkosi.default in the case of arch ? Or just tolerate the absence of args.distribution.name ?

fthiery:
Arch being a rolling distro, it does not have a release, so it will crash when trying to determine the package cache folder name; this commit only appends the "~release" string if present. Fixes #146

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -2266,7 +2266,8 @@ def find_cache(args):
         return
 
     if os.path.exists("mkosi.cache/"):
-        args.cache_path = "mkosi.cache/" + args.distribution.name + "~" + args.release
+        release = "~" + args.release if args.distribution.name != "arch" else ""
+        args.cache_path = "mkosi.cache/" + args.distribution.name + release
 
 def find_build_script(args):
     if args.build_script is not None:
--- a/mkosi
+++ b/mkosi
@@ -2266,8 +2266,8 @@ def find_cache(args):
         return
 
     if os.path.exists("mkosi.cache/"):
-        release = "~" + args.release if args.distribution.name != "arch" else ""
-        args.cache_path = "mkosi.cache/" + args.distribution.name + release
+        args.cache_path = "mkosi.cache/" + args.distribution.name
+        args.cache_path += "~" + args.release if args.release is not None else ""
 
 def find_build_script(args):
     if args.build_script is not None:


mkosi
```diff
@@ -2266,7 +2266,8 @@ def find_cache(args):
         return
 
     if os.path.exists("mkosi.cache/"):
-        args.cache_path = "mkosi.cache/" + args.distribution.name + "~" + args.release
+        release = "~" + args.release if args.distribution.name != "arch" else ""
```
poettering: why check explicitly for "arch" here? it appears much nicer to me to just check whether `args.release is not None` or so? i.e. check for what you actually want to check for instead of some secondary condition...

fthiery: Agreed, building other distros should fail/refuse to build before that if they are missing the release arg.

poettering:
BTW, we genreally prefer merging "perfect" PRs, i.e. where each individual commit reflects a logical step, and not a historical one. i.e. no commits in PRs that fix up previous commits in the same PR please.

I'll squash the two commits now when merging, but please next time it would be nice to keep that in mind!

fthiery:
Sure, sorry for that one, i'll keep this in mind next time.

PR #152, opened 2017-09-08
Do not copy raw images into image if not running from a git project

### LINKED ISSUE #151: image cache files are being copied into the target image if not running from a git project
That one's pretty fun:

```
            if use_git:
            else:
                ignore = shutil.ignore_patterns('.git', '.mkosi-*')
                shutil.copytree(args.build_sources, target, symlinks=True, ignore=ignore)
```
This means that in a non-git folder, raw images are being copied into it, resulting in an out of space error.

```
$ ls
$ image.raw.cache-pre-dev  image.raw.cache-pre-inst  mkosi.build  mkosi.default
$ sudo mkosi -if
...
...
[Errno 28] No space left on device'
```

fthiery:
…fixes #151

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1360,7 +1360,7 @@ def install_build_src(args, workspace, run_build_script, for_cache):
             if use_git:
                 copy_git_files(args.build_sources, target, git_files=args.git_files)
             else:
-                ignore = shutil.ignore_patterns('.git', '.mkosi-*')
+                ignore = shutil.ignore_patterns('.git', '.mkosi-*', '*.raw*')
                 shutil.copytree(args.build_sources, target, symlinks=True, ignore=ignore)
 
 def install_build_dest(args, workspace, run_build_script, for_cache):


fsateler:
Hmm. This would prevent using mkosi for any project that uses files that match `*.raw*`. This is a valid pattern that might be reasonably used by any project. What is the exact problem you are trying to fix?

fsateler:
Sorry, I missed the longer description in #151. I think a better fix would be to either tighten this match to match only the known files, or instead rename the cache files to `.mkosi-$foo`

fthiery:
What do you mean by "known files" ? Files that config allows to predict are expected (like Output arg content + .cache-pre-dev / .cache-pre-inst suffixes) ? Does not work well if multiple configs are available.

While i agree with you that cache files could get hidden, we shouldn't hide image.raw...

In the git case, all ignoring is done by the provided .gitignore:
```
*.cache-pre-dev
*.cache-pre-inst
/.mkosi-*
/SHA256SUMS
/SHA256SUMS.gpg
/__pycache__
/build
/dist
/image
/image.raw
/image.raw.xz
/image.roothash
/image.tar.xz
/mkosi.build
/mkosi.cache
/mkosi.egg-info
/mkosi.extra
/mkosi.nspawn
```
(@poettering is there a reason why you didn't you use /mkosi.* ?)

Given the .gitignore above, i think it would be reasonable to at least include all items in the shutil filter.

```
                ignore = shutil.ignore_patterns('.git', '.mkosi-*', '*.cache-pre-dev', '*.cache-pre-inst', 'mkosi.*', ...)
```

But this does not solve the image.raw issue; another option could be to always output files into a new folder, e.g. mkosi.output (like buildroot / yocto); it would then be ignored like all other mkosi.* folders now (if we use the mkosi.* filter or adding mkosi.output/ to the gitignore and filter list).

I think the output folder approach is the best/easiest one, but also breaking things.

Btw: i don't really care about the git-less scenario, i just discovered it by accident while testing, and i confused me so i think it should be fixed. Also, using mkosi from another CVS would fail too.

poettering:
hmm, I figure `*.cache-pre-dev` and `*.cache-pre-inst` should be added to what we ignore implicitly and by default.

other than that I think that *.raw and stuff should better be added to a project's .gitignore rather than be ignored explicitly by us. After all ".raw" might not refer to files we generate, as the suffix is used in completely different contexts too (for example image tools use it to drop raw pixel data somewhere). Hence, maybe we should implicitly ignore all files which are configured as output for the current operation but not just silently: if we notice that they aren't ignored by git's own ignoring logic too, we should complain loudly and suggest people add them git .gitignore.

fthiery:
AFAIK unless you commit the raw file, when running from a git project, the file won't get copied even if it's not in the gitignore.

This issue and PR concerns only non-git cases, where .. well, you can't add the file to .gitignore :)

What about the mkosi.output/ folder suggestion (along with ignoring it by default) ? I know it breaks things but it's the most KISS solution IMO.

poettering:
@fthiery hmm, it might make sense indeed to drop the generated output into mkosi.output/ if such a directory exists, instead of dropping it into the current directory. By discovering the dir automatically we can maintain compatibility...

fthiery:
Okay, i'll try again with the folder approach ; i also think that some log should display that mkosi is currently copying a very large file (e.g. over 500 MB), just in case the user didn't create a mkosi.output/ folder.

PR #154, opened 2017-09-08
Do not setup systemd-networkd on Arch if networkmanager is used

### LINKED ISSUE #153: Support NetworkManager on Arch images
Currently, systemd-networkd is unconditionally used when installing an Arch system; i am deploying NetworkManager inside my image, so i don't want networkd or resolved to be setup.

fthiery:
networkd is currently unconditionally used and setup; fixes #153

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -714,6 +714,12 @@ Type=ether
 DHCP=yes
 """)
 
+def enable_networkmanager(workspace):
+    subprocess.run(["systemctl",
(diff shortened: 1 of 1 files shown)


poettering:
can't say I like this overly specific checks. yuck... the networkd magic is bad enough, but making this even more complex i really don't like. 

but meh, i'll merge it anyway

PR #155, opened 2017-09-08
Rework image cache logic to support buildscript-less situations and improve speed

fthiery:
- building with -i (image caching) was not working at all if there was no mkosi.build script present (#143)
- pre-dev and pre-inst images were generated every time, reducing the usefulness of caching

Until a change detection mecanism is implemented (that would e.g. hash the Packages/BuildPackages arguments), the image caches must be removed manually.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -260,12 +260,18 @@ def reuse_cache_image(args, workspace, run_build_script, for_cache):
 
     if not args.incremental:
         return None, False
-    if for_cache:
-        return None, False
     if args.output_format not in (OutputFormat.raw_gpt, OutputFormat.raw_btrfs):
         return None, False
 
     fname = args.cache_pre_dev if run_build_script else args.cache_pre_inst
+    if for_cache:
+        if fname and os.path.exists(fname):
+            # Cache already generated, skip generation, note that manually removing the exising cache images is
+            # necessary if Packages or BuildPackages change
+            return None, True
+        else:
+            return None, False
+
     if fname is None:
         return None, False
 
@@ -1885,7 +1891,7 @@ def calculate_signature(args, checksum):
 
 def save_cache(args, workspace, raw, cache_path):
 
-    if cache_path is None:
+    if cache_path is None or raw is None:
         return
 
     with complete_step('Installing cache copy ',
@@ -1966,7 +1972,7 @@ def print_output_size(args):
         st = os.stat(args.output)
         print_step("Resulting image size is " + format_bytes(st.st_size) + ", consumes " + format_bytes(st.st_blocks * 512) + ".")
 
-def setup_cache(args):
+def setup_package_cache(args):
     with complete_step('Setting up package cache',
                        'Setting up package cache {} complete') as output:
         if args.cache_path is None:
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -2819,34 +2829,38 @@ def build_stuff(args):
...
-
-    # Run the image builder for the first (develpoment) stage in preparation for the build script
-    raw, tar, root_hash = build_image(args, workspace, run_build_script=True)
-
-    run_build_script(args, workspace.name, raw)
-    remove_artifacts(args, workspace.name, raw, tar, run_build_script=True)
+        if raw:
+            save_cache(args,
+                       workspace.name,
+                       raw.name if raw is not None else None,
```
fsateler: `raw` was already determined to be truthy here, so you could just use `raw.name`.

mkosi
```diff
@@ -2819,34 +2829,38 @@ def build_stuff(args):
     # always the same
     args.machine_id = uuid.uuid4().hex
 
-    cache = setup_cache(args)
+    package_cache = setup_package_cache(args)
```
fsateler: This variable seems to be unused. It predates your change but since you are already modifying this line...

fthiery:
Thanks for taking a look, i included your suggestions. To be honest i don't fully understand the logic either, partly because of the way the functions are called/nested.

fthiery:
To automatically re-generate the pre-dev and pre-inst images, i was thinking of hashing the args (including Packages and BuildPackages) and use it to name the image (e.g. mediacoder-image.raw.cache-pre-dev-0a4d55a8d778e5022fab701977c5d840bbc486d0), so that the corresponding image is used automatically.

Should i try to include such a feature in this PR ?

poettering:
I must say, I am not sure I follow what this PR is about. The image pre-cache logic exists, so that repeatedly building the same source tree doesn't imply we always start building the image totally from scratch, but can reuse the image we prepared for the last build iteration already. i.e. instead of prepping and image, doing "./configure && make && make install" in it, then prepping a second image, and copying what the make commands generated earlier into it, we can just focus on the configure/make/copy and avoid the prepping.

Now, if I understand correctly you are trying to make this work without build script  (i.e. without `mkosi.build`), but that means we are in a completely different scenario from what the caching was built for initially: if you don't have a build script, then you don't build any sources, and the whole two-phase logic becomes just a single one, with the configure/make/copy steps removed. But if you do that there's no point in caching really, is there?, as the cache would just end up being used as-is without further modifications?

Can you elaborate?

fthiery:
I my build-less case i used the mkosi.postinst to perform the actual customizations, and in that case the image caching was not used and it took way longer. With these patches the postinst is applied to the mediacoder-image.raw.cache-pre-inst cache image, speeding up the process.

I use the postinst script to create users, enable/disable services, setup locale, install custom prebuilt packages. Rebuilding the base image is pretty slow, so i think it is still useful to support -i without caching (ex: for producing distro spins).

Now my project uses an actual build script, so i'm kind of unaffected, but i recall that the cache logic wasn't working at all (and would be fixed by your PR here https://github.com/systemd/mkosi/pull/162) so this PR was about fixing the error and making caching actually work.

PR #158, opened 2017-09-29
157 support other official arch kernels

fthiery:
If the user specifies a linux kernel among the other officially suported ones in the packages list (linux-hardened, linux-lts, linux-zen), mkosi will not install the standard linux package anymore.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -18,6 +18,7 @@ import tempfile
 import time
 import urllib.request
 import uuid
+import glob
 
 try:
     import argcomplete
@@ -1138,9 +1139,26 @@ SigLevel    = Required DatabaseOptional
 """.format(args=args, server=server))
 
     subprocess.run(["pacman", "--color", "never", "--config", os.path.join(workspace, "pacman.conf"), "-Sy"], check=True)
+    # determine base packages list from base metapackage
     c = subprocess.run(["pacman", "--color", "never", "--config", os.path.join(workspace, "pacman.conf"), "-Sg", "base"], stdout=subprocess.PIPE, universal_newlines=True, check=True)
     packages = set(c.stdout.split())
     packages.remove("base")
+
+    official_kernel_packages = [
+        "linux",
+        "linux-lts",
+        "linux-hardened",
+        "linux-zen"
+    ]
+
+    kernel_packages = {"linux"}
+    if args.packages is not None:
+        kernel_packages = set.intersection(set(args.packages), set(official_kernel_packages))
+        # prefer user-specified packages over implicit base kernel
+        if kernel_packages and "linux" not in kernel_packages:
+            packages.remove("linux")
+        if len(kernel_packages) > 1:
+            print('Warning, more than one kernel will be installed: %s' % (' '.join(kernel_packages)))
 
     packages -= {"cryptsetup",
                  "device-mapper",
@@ -1162,8 +1180,7 @@ SigLevel    = Required DatabaseOptional
         elif args.output_format == OutputFormat.raw_btrfs:
             packages.add("btrfs-progs")
     else:
-        if "linux" in packages:
-            packages.remove("linux")
+        packages -= kernel_packages
 
     if args.packages is not None:
         packages |= set(args.packages)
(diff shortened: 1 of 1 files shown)


fsateler:
Other distributions would benefit from allowing changing the installed kernel. Could this be made generic for other distros too?

fthiery:
Well, not really generic; 
- on debian/ubuntu we can override linux-image-amd64 by any linux-image-* if present, 
- on Arch it's a bit dangerous since there kernel packages are only prefixed by "linux-", and many other packages use the same prefix (e.g. linux-tools-meta), so we must resort to some static list
- as for fedora, it seems that a singgle release is supported per version (am i correct ?); as for arch, many other packages are prefixed with "kernel-"
- others i don't know, any tips appreciated

I can work on that but
- what should we do if mupltiple kernels are defined in the package list ? should we allow multiple kernels ? If so, we could install all of them as EFI boot stub, but which should be the default ? (in my branch i just take the last package name, and select the first vmlinuz file i find to install)
- to prevent the whitelist thing, maybe each distro should look at the "Replaces" (or equivalent) for all packages that could match the prefix, but i'm afraid it could make things slower

fthiery:
My previous comment about the "replaces" doesn't work with linux-lts on arch for instance:
```
$ LANG=C pacman -Si linux-lts
Repository      : core
Name            : linux-lts
Version         : 4.9.52-1
Description     : The Linux-lts kernel and modules
Architecture    : x86_64
URL             : https://www.kernel.org/
Licenses        : GPL2
Groups          : None
Provides        : None
Depends On      : coreutils  linux-firmware  kmod  mkinitcpio>=0.7
Optional Deps   : crda: to set the correct wireless channels of your country
Conflicts With  : None
Replaces        : None
Download Size   : 59.76 MiB
Installed Size  : 82.40 MiB
Packager        : Andreas Radke <andyrtr@archlinux.org>
Build Date      : Wed Sep 27 21:49:29 2017
Validated By    : MD5 Sum  SHA-256 Sum  Signature
```

"Replaces" is None, so there's no alternative to a whitelist (at least on Arch).

mkosi
```diff
@@ -1240,7 +1255,7 @@ def install_boot_loader_arch(args, workspace):
     kernel_version = next(filter(lambda x: x[0].isdigit(), os.listdir(os.path.join(workspace, "root", "lib/modules"))))
 
     run_workspace_command(args, workspace,
-                      "/usr/bin/kernel-install", "add", kernel_version, "/boot/vmlinuz-linux")
+                      "/usr/bin/kernel-install", "add", kernel_version, glob.glob("/boot/vmlinuz-*")[0])
```
poettering: hmm, this is ugly, given that it silently picks whatever is first in the globbing list... could you hide this in a new function find_vmlinuz() or so, which hides this, and prints a warning if the kernel image name is not unique, and indicates which one of the options it is picking? (even better: make this deterministic, by sorting the list of kernel images)

mkosi
```diff
@@ -1023,10 +1024,24 @@ SigLevel    = Required DatabaseOptional
...
+        "linux-hardened",
+        "linux-zen"
+    ]
+
+    kernel_package = "linux"
+    for p in other_official_kernel_packages:
+        if p in packages:
+            kernel_package = p
+    if kernel_package != "linux":
+        packages.remove("linux")
```
poettering: hmm, i figure python gurus could write this in a prettier way using the set logic they have, i.e. by determining the intersection of both lists or so... but given that I am not guru enough for knowing how to write that off the top of my head I'll take this as it is.

fthiery: I reworked the logic to support multiple kernels (and not just pick the first one), to keep "linux" if the user specified it explicitly, and use sets/intersection instead of the basic loop i used.

mkosi
fthiery: Did so, as for the warning, i'm using print() as there is no proper logging setup.

fthiery:
Please let me know whenever the changes suit you, at which point i can squash the bugs into one

poettering:
yeah, looks pretty ok now, but please, always make the PR "merge-ready" before review, i.e. squash patches and stuff, so that the PR only contains individual commits for logical steps, and no historical cruft, and things are nice and easy to review and bisect later on.

fthiery:
I rebased and squashed commits. Every time i will adjust i will do so, sorry.

PR #160, opened 2017-10-06
mkosi: unbreak the squashfs mode, when used together with build trees

poettering:
We need to create the mount point for the build tree early on, so that
it is included in the read-only squashfs image, and we can mount the
host's builddir into it.

Without this patch squashfs builds with builddir usage fail.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -681,6 +681,9 @@ def prepare_tree(args, workspace, run_build_script, cached):
         os.mkdir(os.path.join(workspace, "root", "root"), 0o750)
         os.mkdir(os.path.join(workspace, "root", "root/dest"), 0o755)
 
+        if args.build_dir is not None:
+            os.mkdir(os.path.join(workspace, "root", "root/build"), 0o755)
+
 def patch_file(filepath, line_rewriter):
     temp_new_filepath = filepath + ".tmp.new"
 


lucaswerkmeister:
> Without this patch squashfs builds with builddir usage fail.

Hm, I can’t reproduce this failure. `sudo ~/git/mkosi/mkosi -t raw_squashfs` in a project with `mkosi.default` and `mkosi.build` should work (that is, not work), right? (`~/git/mkosi` was on 6ae448696500ffa1df528951bc716e06fce51f54 at the time, to test #161. The root password worked, and the build script was also run, the resulting program was installed in the image.)

poettering:
You need a project that uses the build script logic *and* the builddir stuff *and* outputs into squashfs ro. for example, try to use the systemd git repo, create "mkosi.builddir" in it and use `-t raw_squashfs`, then you should be able to trigger the issue

poettering:
anyway, you already approved it. I am taking the liberty now to merge it given the approval, as the bug makes hacking on systemd hard for me. Or to say this differently: no takebacksies ;-)

lucaswerkmeister:
Oh, I see, I didn’t try with an out-of-tree build dir. Anyways, this is fine to merge :)

PR #162, opened 2017-10-10
fix incremental mode when used without build script

poettering:
A fix for #143. I don't think -i without build script makes sense, but we should handle it gracefully.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -2655,6 +2655,9 @@ def load_args():
         if args.secure_boot_certificate is None:
             die("UEFI SecureBoot enabled, but couldn't find certificate. (Consider placing it in mkosi.secure-boot.crt?)")
 
+    if args.incremental and args.build_script is None:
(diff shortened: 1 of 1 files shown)


fthiery:
As said in my original PR, i think there is a valid usecase for -i support without build script: it's producing distribution spins (e.g. pre-customize it with non-standard packages, create users, define locales, ...). You don't build anything but there is a trial-and-error approach that can be improved a lot by supporting caching.

swilmet:
Yes I wanted to construct my container incrementally, starting simple (since I was also learning mkosi at the same time), and then adding more and more stuff for my needs, but one thing at a time to test if everything works so far. But I stopped using mkosi because the caching didn't work and it was too slow to execute everything each time (installing all the base packages etc, even if the packages are already downloaded).

keszybz:
Replaced by #155.

PR #163, opened 2017-10-10
README: document the debian/ubuntu keyring situation explicitly

poettering:
I figure the mkosi distro packages should add both deps explicitly to
avoid any confusion.

Fixes: #139


fsateler:
I'm sorry for merging this too quickly. It turns out that `ubuntu-keyring` does not exist in debian, but it does in Ubuntu (yay divergence?). Filed #164 as a followup

PR #165, opened 2017-10-10
mkosi.output/ directory support

### LINKED ISSUE #151: image cache files are being copied into the target image if not running from a git project

poettering:
Let's fix #151.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,24 +159,35 @@ they exist in the local directory:
   additional files to an image, on top of what the
   distribution includes in its packages.
 
-* `mkosi.build` may be an executable script. If it exists the
-  image will be built twice: the first iteration will be the
-  *development* image, the second iteration will be the
-  *final* image. The *development* image is used to build the
-  project in the current working directory (the *source*
-  tree). For that the whole directory is copied into the
-  image, along with the mkosi.build build script. The script
-  is then invoked inside the image (via `systemd-nspawn`), with
-  `$SRCDIR` pointing to the *source* tree. `$DESTDIR` points to a
-  directory where the script should place any files generated
-  it would like to end up in the *final* image. Note that
-  `make`/`automake` based build systems generally honour `$DESTDIR`,
-  thus making it very natural to build *source* trees from the
-  build script. After the *development* image was built and the
-  build script ran inside of it, it is removed again. After
-  that the *final* image is built, without any *source* tree or
-  build script copied in. However, this time the contents of
-  `$DESTDIR` is added into the image.
+* `mkosi.build` may be an executable script. If it exists the image
+  will be built twice: the first iteration will be the *development*
+  image, the second iteration will be the *final* image. The
+  *development* image is used to build the project in the current
+  working directory (the *source* tree). For that the whole directory
+  is copied into the image, along with the mkosi.build build
+  script. The script is then invoked inside the image (via
+  `systemd-nspawn`), with `$SRCDIR` pointing to the *source*
+  tree. `$DESTDIR` points to a directory where the script should place
+  any files generated it would like to end up in the *final*
+  image. Note that `make`/`automake` based build systems generally
+  honour `$DESTDIR`, thus making it very natural to build *source*
+  trees from the build script. After the *development* image was built
+  and the build script ran inside of it, it is removed again. After
+  that the *final* image is built, without any *source* tree or build
+  script copied in. However, this time the contents of `$DESTDIR` are
+  added into the image.
+
+  When the source tree is copied into the *build* image, all files are
+  copied, except for `mkosi.builddir/`, `mkosi.cache/` and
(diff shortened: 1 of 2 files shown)


fthiery:
Just tested, works as advertised.

mkosi
```diff
@@ -2576,6 +2588,14 @@ def load_args():
         else:
             args.output = "image"
 
+    if args.output_dir is not None:
+        args.output_dir = os.path.abspath(args.output_dir)
+
+        if "/" not in args.output:
+            args.output = args.output_dir + "/" + args.output
+        else:
+            sys.stderr.write("WARNING: Ignoring configured output directory as output file is a qualified path.");
```
keszybz: Too much C ;) The semicolon is not necessary. Also `\n` is missing.

README.md
```diff
@@ -159,24 +159,35 @@ they exist in the local directory:
...
+  script. The script is then invoked inside the image (via
+  `systemd-nspawn`), with `$SRCDIR` pointing to the *source*
+  tree. `$DESTDIR` points to a directory where the script should place
+  any files generated it would like to end up in the *final*
+  image. Note that `make`/`automake` based build systems generally
+  honour `$DESTDIR`, thus making it very natural to build *source*
+  trees from the build script. After the *development* image was built
+  and the build script ran inside of it, it is removed again. After
+  that the *final* image is built, without any *source* tree or build
+  script copied in. However, this time the contents of `$DESTDIR` is
```
keszybz: s/is/are/

README.md
```diff
@@ -159,24 +159,35 @@ they exist in the local directory:
...
+  When the source tree is copied into the *build* image, all files are
+  copied, except for `mkosi.builddir/`, `mkosi.cache/` and
+  `mkosi.output/`. That said, `.gitignore` is respected if the source
+  tree is a `git` checkout. If multiple different images shall be
+  built from the same source tree it's essential to exclude their
+  output files from this copy operation, as otherwise a version of an
+  image built earlier might be included in a later build, which is
+  usually not intended. An alternative to excluding these built images
+  via `.gitignore` entries is making use of the `mkosi.output/`
+  directory (see below), which is an easy way to exclude all images
```
keszybz: Doesn't sound right, maybe "an easy way to exclude all build artifacts."?

README.md
```diff
@@ -233,6 +244,19 @@ they exist in the local directory:
   support is enabled. All EFI binaries included in the image's ESP are
   signed with this key, as a late step in the build process.
 
+* `mkosi.output/` may be a directory. If it exists, and the image
```
keszybz: Hmm, I think description starts in the wrong place, because it doesn't take into account the case when `--output` is used. What about:

> An output directory may be specifed using the `--output` option, or if that option is not used and `mkosi.output/` directory exists, it will be used as the output directory automatically. If an output directory is used, all outputs will be written to this directory, unless their name is explicitly configured to a path that contains at least one slash (i.e. plain file names still end up in the output directory, but anything resembling a path is used as is). This applies to all build artifacts (the image itself, the root hash file in case Verity is used, the checksum and its signature if enabled, and the nspawn settings file if any). This setting is particularly useful if multiple different images shall be built from the same working directory, as otherwise the build result of a preceeding run might be copied into a build image as part of the
source tree (see above).

mkosi
```diff
@@ -2795,13 +2823,25 @@ def reuse_cache_tree(args, workspace, run_build_script, for_cache, cached):
 
     return True
 
+def make_output_dir(args):
+    """Create the output directory if set and not existing yet"""
+    if args.output_dir is None:
+        return
+
+    try:
+        os.mkdir(args.output_dir, 0o755)
```
keszybz: Maybe add a comment here that `mkdirs` is not appropriate because we only want to create the final component.

README.md
poettering: > An output directory may be specifed using the --output option

But that's misleading. With --output= you specify the output image path, not the dir to place it in. Except when we do -t directory the result will be a file, not a directory after all.

Also, I'd really like to keep this in the list of .mkos-* files and directories, instead of moving this elsewhere.

poettering:
I force pushed a new version now, which makes all suggested changes, except the paragraph rewording proposed by @keszybz, because I am not convinced it would be an improvement.

I'd prefer if we could merge this as-is, and then we can tweak that paragraph later on in a separate PR?

PR #168, opened 2017-10-11
fix unreliable vmlinuz file detection

### LINKED ISSUE #167: Sometimes vmlinuz is not properly detected (Arch)
Since https://github.com/systemd/mkosi/pull/158 sometimes the vmlinuz file is not detected properly, due to improper use of lstrip.

```
In [1]: t = ['/var/tmp/mkosi-wb80oqrv/root/boot/vmlinuz-linux']

In [2]: t[0]
Out[2]: '/var/tmp/mkosi-wb80oqrv/root/boot/vmlinuz-linux'

In [3]: t[0].lstrip('/var/tmp/mkosi-wb80oqrv/root')
Out[3]: 'linuz-linux'

### PR DESCRIPTION
**Pull request by fthiery**
Due to improper use of lstrip, it was somehow working but not reliably, this fixes #167

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1373,8 +1373,13 @@ def find_kernel_file(workspace_root, pattern):
     # Look for the vmlinuz file in the workspace
     workspace_pattern = os.path.join(workspace_root, pattern.lstrip('/'))
     kernel_files = sorted(glob.glob(workspace_pattern))
-    # the path kernel-install expects is within the workspace reference as it is run from within the container
-    kernel_file = '/' + (kernel_files[0].lstrip(workspace_root))
+    kernel_file = kernel_files[0]
+    # The path the kernel-install script expects is within the workspace reference as it is run from within the container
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: mkosi [hunk 1]
```diff
@@ -1374,7 +1374,7 @@ def find_kernel_file(workspace_root, pattern):
     workspace_pattern = os.path.join(workspace_root, pattern.lstrip('/'))
     kernel_files = sorted(glob.glob(workspace_pattern))
     # the path kernel-install expects is within the workspace reference as it is run from within the container
-    kernel_file = '/' + (kernel_files[0].lstrip(workspace_root))
+    kernel_file = kernel_files[0].replace(workspace_root, '')
```
lucaswerkmeister: This is also fragile if the `workspace_root` appears again in the path later. Some combination of `startswith` and then a splice operator is probably the safest option.

fthiery: Good point, please check my changes

Issue #166, opened 2017-10-11
Tree copy fails if build produces local files to folders that are a symbolic link (e.g. /lib > /usr/lib on Arch)


fthiery:
My build script installs files into /root/dest/lib/firmware, and mkosi crashes in that case.

```
\u2023 Running build script...
...
Copy file /lib/firmware/fw.bin to /root/dest/lib/firmware/fw.bin
...
\u2023 Running build script complete.
\u2023 Copying in build tree...
cp: cannot overwrite non-directory '/var/tmp/mkosi-qs39r9yw/root/lib' with directory '/var/tmp/mkosi-qs39r9yw/dest/lib'
\u2023 Unmounting Package Cache...
...
\u2023 Detaching image file complete.
Traceback (most recent call last):
  File "/usr/bin/mkosi", line 3072, in <module>
    main()
  File "/usr/bin/mkosi", line 3068, in main
    build_stuff(args)
  File "/usr/bin/mkosi", line 3020, in build_stuff
    raw, tar, root_hash = build_image(args, workspace, run_build_script=False)
  File "/usr/bin/mkosi", line 2871, in build_image
    install_build_dest(args, workspace.name, run_build_script, for_cache)
  File "/usr/bin/mkosi", line 1512, in install_build_dest
    enumerate_and_copy(os.path.join(workspace, "dest"), os.path.join(workspace, "root"))
  File "/usr/bin/mkosi", line 1430, in enumerate_and_copy
    subprocess.run(["cp", "--reflink=auto", "--recursive", "--no-dereference", "--preserve=all", "--no-target-directory", source, dest], check=True)
  File "/usr/lib/python3.6/subprocess.py", line 418, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['cp', '--reflink=auto', '--recursive', '--no-dereference', '--preserve=all', '--no-target-directory', '/var/tmp/mkosi-qs39r9yw/dest', '/var/tmp/mkosi-qs39r9yw/root']' returned non-zero exit status 1.
```

To reproduce, use the following as mkosi.build script:

```
mkdir -p /root/dest/lib/
echo "test" > /root/dest/lib/testfile
```

fthiery:
Adding some debug shows that cp refuses to copy into lib because (at least on arch) /lib is a symlink to /usr/lib:
```
Copy /var/tmp/mkosi-kktbf0bs/dest to /var/tmp/mkosi-kktbf0bs/root
[cp running]
[...]

PR #169, opened 2017-10-11
use rsync instead of cp in order to support copying into symlink folders

fthiery:
Some distros (e.g. Arch, see #166) have symlinks on / (e.g. /lib > /usr/lib), and cp is unable to copy into symlink folders. This PR replaces cp by rsync, which supports copying into symlink folders.


keszybz:
#183 implements copying internally instead. Let's close this.

PR #170, opened 2017-10-11
Fix building base Arch image if no additional package was requested

fthiery:
This fixes a regression induced by NetworkManager support (#153), if no packages are defined (at all).

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1196,10 +1196,10 @@ SigLevel    = Required DatabaseOptional
 
     subprocess.run(cmdline, check=True)
 
-    if not "networkmanager" in args.packages:
-        enable_networkd(workspace)
-    else:
+    if args.packages and "networkmanager" in args.packages:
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -1196,7 +1196,7 @@ SigLevel    = Required DatabaseOptional
 
     subprocess.run(cmdline, check=True)
 
-    if not "networkmanager" in args.packages:
+    if not args.packages or not "networkmanager" in args.packages:
```
lucaswerkmeister: I feel like this would be cleaner the other way around:
```python
if args.packages and "networkmanager" in args.packages:
    enable_networkmanager(workspace)
else:
    enable_networkd(workspace)
```

fthiery:
Agreed, much simpler now, thanks for the suggestion.

lucaswerkmeister:
Great, thanks for the fix!

Issue #159, opened 2017-10-01
OpenSuse fails: Failed to cache repo


Nudin:
Building an OpenSuse-Image (on Arch) fails with:
```
Download (curl) error for 'https://download.opensuse.org/update/tumbleweed/repodata/repomd.xml':
Error code:  Bad URL
Error message: Protocol "http" not supported or disabled in libcurl
```
This can be fixed by overriding mirror to not use https but http. (See also #109 & #137) but it still fails then with:

```
Building repository 'Main' cache ................................................................................[done]
Error building the cache:
[|] Failed to cache repo (4).                                                                                          
Warning: The metadata cache needs to be built for the 'Main' repository. You can run 'zypper refresh' as root to do this.
Disabling repository 'Main'.
Building repository 'Updates' cache .............................................................................[done]
Error building the cache:
[|] Failed to cache repo (127).                                                                                        
Warning: The metadata cache needs to be built for the 'Updates' repository. You can run 'zypper refresh' as root to do this.
Disabling repository 'Updates'.
Some of the repositories have not been refreshed because of an error.
Loading repository data...
Reading installed packages...
'patterns-base-minimal_base' not found in package names. Trying capabilities.
No provider of 'patterns-base-minimal_base' found.
```

poettering:
Hmm, @fbuihuu is our OpenSuse guru. @fbuihuu any idea what is going on here?

fbuihuu:
> Hmm, @fbuihuu is our OpenSuse guru

Oh hum... not really ;)

It looks a duplicate of issue #137

Nudin:
Hi,

but if this is only a duplicate of #137 shouldn't it be fixed by  #173? But as mentioned above, I did try that and it removed the first error-message but the second one still happened…

lucaswerkmeister:
Hm, I’m also seeing this error.
```
Building repository 'Updates' cache ............[done]
Error building the cache:
[|] Failed to cache repo (127).
Warning: The metadata cache needs to be built for the 'Updates' repository. You can run 'zypper refresh' as root to do this.
Disabling repository 'Updates'.
Some of the repositories have not been refreshed because of an error.
Loading repository data...
Reading installed packages...
'minimal_base' not found in package names. Trying capabilities.
No provider of 'minimal_base' found.
```
I’m not seeing any further error message that would indicate *why* Updates could not be refreshed, though. But let’s reopen this, looks like #137/#173 only covered the first part.

dermarens:
[...]

Issue #137, opened 2017-08-10
opensuse: bad URL


reverendhomer:
Hi, I tried to make an opensuse image under fedora 26 and got this error message:

```
Download (curl) error for 'https://download.opensuse.org/update/tumbleweed/repodata/repomd.xml':
Error code:  Bad URL
Error message: Protocol "http" not supported or disabled in libcurl
```

mkosi from git master. How can I fix that?

poettering:
Hmm, very weird... does curl work on the cmdline?

reverendhomer:
Yup
```
λ curl https://download.opensuse.org/update/tumbleweed/repodata/repomd.xml
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>302 Found</title>
</head><body>
<h1>Found</h1>
<p>The document has moved <a href="http://download.opensuse.org/repositories/openSUSE:/Factory:/Update/standard/repodata/repomd.xml">here</a>.</p>
<hr>
<address>Apache/2.2.34 (Linux/SUSE) Server at download.opensuse.org Port 443</address>
</body></html>

```

poettering:
Hmm @fbuihuu any idea about this one, given that this is in the opensuse code?

keszybz:
I suspect this is caused by the redirect to plain http from an https URL.

fbuihuu:
It seems that curl(1) doesn't like such redirection anymore.

There seems to be some WIP to stop redirecting https to http, see https://github.com/openSUSE/mirrorbrain/issues/3 but for the time being we could fix that simply by using http:// for the mirrors as there seems to be no additional value in https:// (the rpms are signed by gpg keys already, so is the repodata

PR #173, opened 2017-10-12
openSUSE: don't use https to access mirrors for now

fbuihuu:
Some openSUSE mirrors redirect https to http and this seems to make
curl(1) unhappy now.

There's a request to fix that on mirror side:
https://github.com/openSUSE/mirrorbrain/issues/3 but for the time
being switch to http to access all mirrors.

There seems to be no additional values in https:// as the rpms are
signed by gpg keys already, so is the repodata anyways.

Fixes: #137
Fixes: #159

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -2574,7 +2574,7 @@ def load_args():
             if platform.machine() == "aarch64":
                 args.mirror = "http://mirror.archlinuxarm.org"
         elif args.distribution == Distribution.opensuse:
-            args.mirror = "https://download.opensuse.org"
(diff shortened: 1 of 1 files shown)


lucaswerkmeister:
Just for the record, I strongly object to the statement that there’s “no additional value in HTTPS”. For example:

* Mirrors that support HTTP/2 are likely to only offer it on HTTPS (even though this restriction was first introduced by browsers).
* Using HTTPS hides the list of packages you’re installing.

However, that’s no reason not to merge this.

Issue #142, opened 2017-08-29
Source copying is borked when there are modified symlinks


fsateler:
May be related to #123? Steps to reproduce:

```sh
% git clone https://github.com/systemd/systemd.git
% cd systemd
% ln -sf .mkosi/mkosi.debian mkosi.default
% sudo mkosi
# snip lots of output
‣ Copying in build script and sources...
Cloning into '/var/tmp/mkosi-3nz7dw74/root/root/src'...
warning: --depth is ignored in local clones; use file:// instead.
done.
‣ Unmounting Package Cache...
‣ Unmounting Package Cache complete.
‣ Unmounting image...
‣ Unmounting image complete.
‣ Detaching image file...
‣ Detaching image file complete.
Traceback (most recent call last):
  File "/usr/bin/mkosi", line 2904, in <module>
    main()
  File "/usr/bin/mkosi", line 2900, in main
    build_stuff(args)
  File "/usr/bin/mkosi", line 2846, in build_stuff
    raw, tar, root_hash = build_image(args, workspace, run_build_script=True)
  File "/usr/bin/mkosi", line 2706, in build_image
    install_build_src(args, workspace.name, run_build_script, for_cache)
  File "/usr/bin/mkosi", line 1361, in install_build_src
    copy_git_files(args.build_sources, target, git_files=args.git_files)
  File "/usr/bin/mkosi", line 1339, in copy_git_files
    shutil.copy2(src_path, dest_path, follow_symlinks=False)
  File "/usr/lib/python3.5/shutil.py", line 257, in copy2
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/lib/python3.5/shutil.py", line 118, in copyfile
    os.symlink(os.readlink(src), dst)
FileExistsError: [Errno 17] File exists: '.mkosi/mkosi.debian' -> '/var/tmp/mkosi-3nz7dw74/root/root/src/mkosi.default'
```

cc @lucasdemarchi

lucasdemarchi:
Humn... yeah, sorry about that, it seems I broke it. I'll take a look on how to fix it.

TrumanLing:
Same problem +1

TrumanLing:
@fsateler @lucasdemarchi @poettering can we check whether file is symbolic link? if yes, delete it then copy it.

lucasdemarchi:
@TrumanLing what I was thinking but never got to implement (sorry, my bad) was to have a try/catch and use a force flag on the symlink creation.

PR #174, opened 2017-10-12
Avoid conflict when copying symbolic links

TrumanLing:
Fix bug #142: avoid symbolic link conflict.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1436,7 +1436,10 @@ def copy_git_files(src, dest, *, git_files):
         directory = os.path.dirname(dest_path)
         os.makedirs(directory, exist_ok=True)
 
-        shutil.copy2(src_path, dest_path, follow_symlinks=False)
+        if os.path.islink(src_path):
+            subprocess.run(['ln', '-nfs', src_path, dest_path])
+        else:
+            shutil.copy2(src_path, dest_path, follow_symlinks=False)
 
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
LGTM, thanks!

lucasdemarchi:
Thinking again, there's one corner case. If previously `dest_path` was a directory and is now a link, the `os.remove()` call will fail and raise an exception. So you may want to try copying and handle the cases in which it fails, or handle the exception on the `os.remove()`

TrumanLing:
@lucasdemarchi I have updated my modification via "ls -nfs", which is your original idea, I think.

mkosi
```diff
@@ -1436,7 +1436,10 @@ def copy_git_files(src, dest, *, git_files):
         directory = os.path.dirname(dest_path)
         os.makedirs(directory, exist_ok=True)
 
-        shutil.copy2(src_path, dest_path, follow_symlinks=False)
+        if os.path.islink(src_path):
+            subprocess.run(['ls','-nfs', src_path, dest_path])
```
keszybz: `ln` not `ls`!

Also a space is missing before `'-nfs'`.

TrumanLing:
@keszybz I am sorry for my handwriting mistake, and I have updated it.

keszybz:
#183 implements copying internally instead. Let's close this. I'll add a comment in the other PR why.

PR #180, opened 2017-10-25
add "mkosi qemu" verb

poettering:
This PR makes things work on Fedora only so far. On other distributions the qemu/kvm paths might differ and the location of the UEFI firmware blobs. Basic infrastructure is in place to deal with that, but it is not populated with the specific requirements and paths for the other distributions.

I think it would make a lot of sense to merge this for Fedora initially, and then extend this with the right paths for the other distributions with later PRs.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -3297,13 +3297,7 @@ def run_shell(args):
     if args.cmdline:
         cmdline += ('--', *args.cmdline)
 
-    try:
-        subprocess.run(cmdline, check=True)
-    except subprocess.CalledProcessError as e:
-        if e.returncode >= 0:
-            sys.exit(e.returncode)
-
-        sys.exit(255)
+    os.execvp(cmdline[0], cmdline)
 
 def main():
     args = load_args()
--- a/mkosi
+++ b/mkosi
@@ -2135,8 +2135,8 @@ def parse_args():
     parser = argparse.ArgumentParser(description='Build Legacy-Free OS Images', add_help=False)
 
     group = parser.add_argument_group("Commands")
-    group.add_argument("verb", choices=("build", "clean", "help", "summary", "shell", "boot"), nargs='?', default="build", help='Operation to execute')
-    group.add_argument("cmdline", nargs=argparse.REMAINDER, help="The command line to use for 'shell' and 'boot'")
+    group.add_argument("verb", choices=("build", "clean", "help", "summary", "shell", "boot", "qemu"), nargs='?', default="build", help='Operation to execute')
+    group.add_argument("cmdline", nargs=argparse.REMAINDER, help="The command line to use for 'shell', 'boot', 'qemu'")
     group.add_argument('-h', '--help', action='help', help="Show this help")
     group.add_argument('--version', action='version', version='%(prog)s ' + __version__)
 
@@ -2713,8 +2713,8 @@ def load_args():
     find_passphrase(args)
     find_secure_boot(args)
 
-    if args.cmdline and args.verb not in ('shell', 'boot'):
-        die("Additional parameters only accepted for 'shell' and 'boot' invocations.")
+    if args.cmdline and args.verb not in ('shell', 'boot', 'qemu'):
+        die("Additional parameters only accepted for 'shell', 'boot', 'qemu' invocations.")
 
     args.force = args.force_count > 0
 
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -2745,12 +2747,16 @@ def load_args():
         if args.secure_boot_certificate is None:
             die("UEFI SecureBoot enabled, but couldn't find certificate. (Consider placing it in mkosi.secure-boot.crt?)")
 
-    if args.verb in ("shell", "boot"):
+    if args.verb in ("shell", "boot", "qemu"):
```
fsateler: This seems redundant with the check below

mkosi
```diff
@@ -3145,17 +3151,63 @@ def run_shell(args):
...
+    else:
+        found = None
+
+    if found is None:
+        found = shutil.which("qemu")
+
+    if found is None:
+        found = shutil.which("qemu-kvm")
+        if found is None:
+            die("Couldn't find QEMU/KVM binary")
```
fsateler: I think it would be nice to have a command `--qemu-binary` option to be able to override the command.

mkosi
```diff
@@ -3145,17 +3151,63 @@ def run_shell(args):
...
+
+        cmdline = [ found ]
+    else:
+        cmdline = [ found, "-machine", "accel=kvm" ]
+
+    found = None
+    for i in ( '/usr/share/edk2/ovmf/OVMF_CODE.fd', ):
```
fsateler: A more pythonic way to do this would be:

```python
paths = ('/usr/share/edk2/ovmf/OVMF_CODE.fd',)
found = next((p for p in paths if os.path.exists(p)), None)
```

keszybz: Please write this as:
```python
for path in ('/usr/share/edk2/ovmf/OVMF_CODE.fd',):
    if os.path.exists(path):
        break
else:
     die("Couldn't find OVMF UEFI firmware blob.")
```

mkosi
```diff
@@ -3147,17 +3151,63 @@ def run_shell(args):
...
+    cmdline += [ "-bios", found,
+                 "-smp", "2",
+                 "-m", "1024",
+                 "-drive", "format=raw,file=" + args.output,
+                 *args.cmdline ]
+
+    try:
+        subprocess.run(cmdline, check=True)
+    except subprocess.CalledProcessError as e:
+        if e.returncode >= 0:
```
keszybz: `> 0`?

mkosi
keszybz: Some later PR.

mkosi
poettering: Hmm? not following?

I mean, the branch contains two checks, the tar check and the xz check. The latter is not redundant. The first one somewhat is, but thinking about the alternatives I like this way much better, and it outputs a friendlier, morespecific message anyway this way...

mkosi
poettering: we want to propagate the return code including when it is 0... not following here?

mkosi
fsateler: Ah, indeed the compression is independent from the output format. Sorry for the noise.

poettering:
force pushed a new version. PTAL. Addressed all of @keszybz points, with the exception of the retval thing, but see discussion above...

mkosi
keszybz: I don't understand how this is supposed to work. I'd expect that if subprocess.run() is successful, exit(0) is called, and exit(n), where 0 < n <= 255 is called. Calling exit(0) when CalledProcessError was thrown seems wrong. But the docs for CalledProcessError specify that it is only called when the command returns a non-zero exit status, so this check seems redundant anyway.

I'd expect something like this:
```python
sys.exit(subprocess.run(cmdline).returncode)
```
but maybe I'm missing the reason to mangle the return code?

poettering: if the invoked process fails and aborts due to a signal, then `CalledProcessError.returncode` will be negative. If it exited with a non-zero exit code, it will be positive. In either case we want mkosi to fail with a non-zero exit code. The code you are looking at hence propagates the returncode as is if it is positive, and if it is negative returns 255 (simply because there's no better value to return really, as propagating signals as signals probably doesn't make too much sense...)

poettering: thinking about this, I figure it would be nicer for the case where we are just supposed to execute qemu or nspawn, to use plain execve() instead of forking something off. After all, we are little more than a preparator for the qemu/nspawn command line we'll eventually invoke.

poettering:
Force pushed a new version now, that invokes qemu und the final nspawn of "mkosi shell" and "mkosi boot" through os.execvp(), replacing mkosi in memory. I am pretty sure this is much nicer, as the caller can then process the original exit codes without us having to (incompletely) proxy them.

PTAL!

mkosi
```diff
@@ -3297,13 +3297,7 @@ def run_shell(args):
...
         cmdline += ('--', *args.cmdline)
 
-    try:
-        subprocess.run(cmdline, check=True)
-    except subprocess.CalledProcessError as e:
-        if e.returncode >= 0:
-            sys.exit(e.returncode)
-
-        sys.exit(255)
+    os.execvp(cmdline[0], cmdline)
```
keszybz: Nice!

q2dg:
README.md should be updated accordingly, shouldn't it?
Thanks!

PR #183, opened 2017-10-30
Copy tree

### LINKED ISSUE #142: Source copying is borked when there are modified symlinks

keszybz:
Fixes #142, replaces #169, #174.

Hmm, this doesn't handle the case where a directory is replaced by a file. It's probably not that common, and should be easy to fix, so I'm submitting this incomplete version for now. Device nodes and special files are also not supported, but could be added easily.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -121,6 +121,18 @@ def roundup512(x):
 def print_step(text):
     sys.stderr.write("‣ \033[0;1;39m" + text + "\033[0m\n")
 
+def mkdir_last(path, mode=0o777):
+    """Create directory path
+
+    Only the final component will be created, so this is different than mkdirs().
+    """
+    try:
+        os.mkdir(path, mode)
+    except FileExistsError:
+        if not os.path.isdir(path):
+            raise
+    return path
+
 @contextlib.contextmanager
 def complete_step(text, text2=None):
     print_step(text + '...')
@@ -634,10 +646,7 @@ def prepare_tree(args, workspace, run_build_script, cached):
     if args.output_format == OutputFormat.subvolume:
         btrfs_subvol_create(os.path.join(workspace, "root"))
     else:
-        try:
-            os.mkdir(os.path.join(workspace, "root"))
-        except FileExistsError:
-            pass
+        mkdir_last(os.path.join(workspace, "root"))
 
     if args.output_format in (OutputFormat.subvolume, OutputFormat.raw_btrfs):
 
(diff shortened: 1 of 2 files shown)


lucasdemarchi:
I tested it here and it works, although I don't follow why it's better than the other proposed fixes.

poettering:
@lucasdemarchi btw, i just invited you to become an mkosi maintainer. if you see a patch that you like, and are sufficiently sure it's good, then please just go ahead and merge it.

poettering:
I figure we should start splitting out parts of the huge script sooner or later, I figure it's getting quite long, and for a number of the concepts we have now a very good case could be made for "this is library code, and should live in its own python source". Not entirely sure how to properly install that tough? should we maybe switch to meson for this too?

poettering:
(as you can see, my python-fu is limited, i never hacked on a project larger than a single script)

keszybz:
> I don't follow why it's better than the other proposed fixes.

Doing things internally is generally preferred:
- tiny bit faster
- usually better error handling, because the caller can access more information
- good control over features

The problem with doing things internally is when a lot of code is required. Current implementation is a bit verbose, but it implements reflinks, and should handle all the corner cases exactly as we want them handled. Ideally, at least some of this stuff would be moved into the python stdlib, and then our footprint would be reduced. But even without that, I think it's worth it.

poettering:
Hmm, so something is wrong with this. After merging this I now get corrupted files (always at the end of files, where some extra rubbish appears to be appended) when generated the systemd mkosi file (which is btrfs by default)...

Issue #186, opened 2017-11-02
UTF-8 locale file not found in Arch Linux


johnlinp:
### Reproducing Steps

1. Create a `mkosi.default` like this:

```
[Distribution]
Distribution=arch

[Output]
Format=raw_btrfs
Bootable=yes

[Partitions]
RootSize=2G

[Packages]
Cache=/var/cache/pacman/pkg/
Packages=
        python
```

2. `mkosi`

3. `systemd-nspawn -bi image.raw`

4. Try to set locale to `en_US.UTF-8` but failed:

```
[root@image ~]# export LC_CTYPE=en_US.UTF-8
-bash: warning: setlocale: LC_CTYPE: cannot change locale (en_US.UTF-8): No such file or directory
```

5. Check the preferred encoding with `python`:

```
[root@image ~]# python -c 'import locale; print(locale.getpreferredencoding())'
ANSI_X3.4-1968
```

The last step is for `meson` because `meson` needs the encoding to be "UTF-8".

PR #187, opened 2017-11-02
mkosi: generate default UTF-8 locale for Arch Linux

johnlinp:
Fixes: #186

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1201,6 +1201,11 @@ SigLevel    = Required DatabaseOptional
     else:
         enable_networkd(workspace)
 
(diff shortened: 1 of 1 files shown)


poettering:
grrr. A pity that arch doesn't get this right out-of-the-box

poettering:
hmm, this approach leaves locale-gen.service around in the image. I'd prefer if we didn't. Can't we instead just invoke localegen through nspawn, after running the package manager?

johnlinp:
@poettering Thank you for the comments. I will work on it.

johnlinp:
In the new patch, I use `run_workspace_command()` to execute `locale-gen` instead of creating a service unit. Please take a look at it. Thanks.

PR #188, opened 2017-11-03
Fix ubuntu builds

NeilW:
Fix ubuntu builds by binding in the host resolver when using the host network namespace

Update the default distribution to a supported version.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -741,7 +741,10 @@ def run_workspace_command(args, workspace, *cmd, network=False, env={}):
                "--register=no",
                "--bind=" + var_tmp(workspace) + ":/var/tmp" ]
 
-    if not network:
+    if network:
+        # If we're using the host network namespace, use the same resolver
+        cmdline += ["--bind-ro=/etc/resolv.conf"]
+    else:
         cmdline += ["--private-network"]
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -741,7 +741,10 @@ def run_workspace_command(args, workspace, *cmd, network=False, env={}):
                "--register=no",
                "--bind=" + var_tmp(workspace) + ":/var/tmp" ]
 
-    if not network:
+    if network:
+        # If we're using the host network namespace, use the same resolver
+        cmdline += ["--bind-ro=/etc/resolv.conf"]
```
lucaswerkmeister: Hm, what does this do if `/etc/resolv.conf` is a symlink?

NeilW: Well here the links are followed on both sides and mounted in the appropriate place. 

/dev/vda1 on /run/systemd/resolve/stub-resolv.conf type ext4 (ro,relatime,seclab
el,data=ordered)

lucaswerkmeister: Hm, looks like it indeed work.

juaningan: I think that same change is missing also for build scipt

PR #190, opened 2017-11-08
Add support to bmap-tools

lucasdemarchi:
bmap-tools (https://github.com/intel/bmap-tools) allows us to write
images to physical disks like USB drives without having to write the
entire image. It rather writes only the used blocks.  With this we
reduce the time to write to a disk when we manually set the size of the
partitions.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1986,6 +1986,19 @@ def calculate_signature(args, checksum):
 
     return f
 
+def calculate_bmap(args, raw):
+    if args.output_format not in (OutputFormat.raw_gpt, OutputFormat.raw_btrfs):
+        return None
+
+    with complete_step('Creating BMAP file'):
+        f = tempfile.NamedTemporaryFile(mode="w+", prefix=".mkosi-", encoding="utf-8",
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -2054,6 +2076,7 @@ def parse_args():
     group.add_argument("--checksum", action='store_true', help='Write SHA256SUMS file')
     group.add_argument("--sign", action='store_true', help='Write and sign SHA256SUMS file')
     group.add_argument("--key", help='GPG key to use for signing')
+    group.add_argument("--bmap", action='store_true', help='Write .bmap file')
```
poettering: one nitpick, this is pretty brief, could you extend this help text a tiny bit, and say something like "Write block map file (.bmap) for bmaptool usage".

Otherwise it's very unclear what that actually is, but with this change there's "bmaptool" referenced which is much easier googleable...

lucasdemarchi: Done, thanks.

lucasdemarchi:
v2 force pushed. I also changed it so it only creates a bmap file for raw_gpt and raw_btrfs as I think it doesn't make sense for the other formats.

Issue #189, opened 2017-11-06
using self-compiled packages or Fedora copr


fedelibre:
I'm currently using the postinst script to download a repository and build the binary of a software which is not packaged in Fedora. I don't like much this approach, as there's no cache and the download is always performed every time I run mkosi.

So I decided to write a spec and built an RPM package. How can I let mkosi use this local package?
Another option would be uploading the RPM to Fedora Copr. But I don't think mkosi can install packages from unofficial repositories, right?

fsateler:
You can add the package to mkosi.extras. Or you could have it pre-installed there.

fedelibre:
Having it preinstalled in mkosi.extra was the first choice I made. Then I preferred building the binary during the postinst, as I don't like adding binaries in my project repository.

When you say "add the package to mkosi.extras", it means that I should just copy the RPM file somewhere and then it's up to the user to install it?
It's not what I want.

I think I'll stay with the postinst script.

Installing a package from Copr would be cool.

fedelibre:
This morining I woke up with the idea of downloading the .rpm file within `mkosi.extra/` ("gitignoring" .rpm files and using a Makefile to download the files needed) and then installing the .rpm in the mkosi.postinst.

But skeleton tree is way better. I'll wait for the pull request to be merged. Thanks @lucasdemarchi !

PR #192, opened 2017-11-08
Add support to skeleton trees

lucasdemarchi:
It's already possible to pass extra trees to copy over the final tree
after installing all packages. However in some cases it is desirable to
copy the tree before installing packages or running anything in the
image.  One of the cases is if we want to enable additional repositories
for the package manager or want to configure it in a particular way.

Now mkosi can use a mkosi.skeleton directory (or by passing a
--skeleton-tree argument) in the same way it currently supports extra
trees, however copying them before running the "install_distro" phase.

To allow changing the package manager configuration distros may need
more tweaks to point the package manager to the right files. Right now
this has been tested with Fedora 27 by adding
`mkosi.skeleton/etc/yum.repos.d/rpmfusion-free.repo` and installing a
package from that repo.

Closes #106.
May solve #189.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -2593,11 +2593,11 @@ def find_nspawn_settings(args):
         args.nspawn_settings = "mkosi.nspawn"
 
 def find_extra(args):
-    if os.path.exists("mkosi.extra"):
+    if os.path.isdir("mkosi.extra"):
         args.extra_trees.append("mkosi.extra")
 
 def find_skeleton(args):
-    if os.path.exists("mkosi.skeleton"):
+    if os.path.isdir("mkosi.skeleton"):
         args.skeleton_trees.append("mkosi.skeleton")
 
 def find_cache(args):
--- a/mkosi
+++ b/mkosi
@@ -1574,6 +1574,14 @@ def install_extra_trees(args, workspace, for_cache):
         for d in args.extra_trees:
             copy(d, os.path.join(workspace, "root"))
 
+def install_skeleton_trees(args, workspace, for_cache):
+    if args.skeleton_trees is None:
+        return
+
+    with complete_step('Copying in skeleton file trees'):
+        for d in args.skeleton_trees:
+            copy(d, os.path.join(workspace, "root"))
+
 def copy_git_files(src, dest, *, git_files):
     run(['git', 'clone', '--depth=1', '--recursive', '--shallow-submodules', src, dest],
         check=True)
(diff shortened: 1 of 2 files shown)


lucasdemarchi:
One shortcoming that I noticed is preserving file ownership from the extra tree. Usually we will want the file on final tree to be owned by root, but that means we should store mkosi.skeleton/* as owned by root. This is bad when we want to keep the mkosi configuration under git control.  Possible solutions that I can imagine:

1. For each extra and skeleton tree, support a `/mkosi-tree.conf` or something like that which configures the user and extra params to be applied for that tree. 
2. Add a `preinst` script that can fixup any perm and do more things in the tree before installing packages

mkosi
```diff
@@ -1438,6 +1438,17 @@ def install_extra_trees(args, workspace, for_cache):
         for d in args.extra_trees:
             enumerate_and_copy(d, os.path.join(workspace, "root"))
 
+def install_skeleton_trees(args, workspace, for_cache):
+    if args.skeleton_trees is None:
+        return
+
+    if for_cache:
+        return
```
poettering: why this bit? I think this should be included in the cache files really, otherwise the skeleton tree would not be taken into account when we invoke mkosi in incremental mode, but it really should be in that case, so that we can properly cache the result of the first dnf/apt-get run

mkosi
```diff
@@ -1438,6 +1438,17 @@ def install_extra_trees(args, workspace, for_cache):
...
+
+    with complete_step('Copying in skeleton file trees'):
+        for d in args.skeleton_trees:
+            enumerate_and_copy(d, os.path.join(workspace, "root"))
```
poettering: now that I merged @keszybz copy() rework, could you update this bit accordingly?

mkosi
```diff
@@ -2715,6 +2741,11 @@ def load_args():
         for i in range(len(args.extra_trees)):
             args.extra_trees[i] = os.path.abspath(args.extra_trees[i])
 
+    if args.skeleton_trees is not None:
+        for i in range(len(args.skeleton_trees)):
+            args.skeleton_trees[i] = os.path.abspath(args.skeleton_trees[i])
+
+
```
poettering: spurious double newline here...

poettering:
also add a short entry about the skeleton tree to the list in README.md!

poettering:
hmm, regarding the access mode/ownership issue: never thought about this. And it's a bummer, given that git stores neither ownership nor the full access mode. And this applies to both mkosi.skeleton/ and mkosi.extra/... Yuck...

What about this: 

1. declare that all files and dirs copied in through either of these two dirs, will be owned by root, and have one of 0755, 0555, 0644, 0444 as access mode, depending on whether the files/dirs have both x and w bits set in the source, or just the the x bit, or just the w bit, or neither. 
2. add a concept of "mkosi.extra.tar" and "mkosi.skeleton.tar" for the more complex cases, which are tarballs (uncompressed) which are like the dirs, but can adjust access mode/ownership.

I figure for most cases the dirs would suffice, but for other cases the tarballs should be workable enough

mkosi
lucasdemarchi: I misread what for_cache was and just copied what was there for extra_tree.

lucasdemarchi:
I think this is ready for review again, but I will make some additional tests if I didn't screw up something in my search and replace. The skeleton tree is working for me, both using a directory (thus files will be owned by root) or as a tar archive (preserving ownership).

Note that what I said in https://github.com/systemd/mkosi/pull/192#issuecomment-342871633 ("final tree to be owned by root, but that means we should store mkosi.skeleton/* as owned by root." ) was already not true anymore, probably as the result of rebasing this branch on top of the new copy() operations.

Issue #185, opened 2017-11-01
ubuntu 'artful' fails to install - Temporary failure resolving 'archive.ubuntu.com'


NeilW:
Building Ubuntu 'artful' on Fedora 26 with

`sudo mkosi/mkosi --distribution ubuntu       --release artful       --output artful.tar.xz   --format tar       build`

builds the debootstrap phase and then fails with

```
Host and machine ids are equal (1d97ae2831c44890b68ba67e8b283bff): refusing to link journals
Reading package lists... Done
Building dependency tree... Done
The following additional packages will be installed:
  libdbus-1-3 libexpat1
Suggested packages:
  default-dbus-session-bus | dbus-session-bus
The following NEW packages will be installed:
  dbus libdbus-1-3 libexpat1 libpam-systemd
0 upgraded, 4 newly installed, 0 to remove and 0 not upgraded.
Need to get 492 kB of archives.
After this operation, 1758 kB of additional disk space will be used.
Err:1 http://archive.ubuntu.com/ubuntu artful/main amd64 libdbus-1-3 amd64 1.10.22-1ubuntu1
  Temporary failure resolving 'archive.ubuntu.com'
Err:2 http://archive.ubuntu.com/ubuntu artful/main amd64 libexpat1 amd64 2.2.3-1
  Temporary failure resolving 'archive.ubuntu.com'
Err:3 http://archive.ubuntu.com/ubuntu artful/main amd64 dbus amd64 1.10.22-1ubuntu1
  Temporary failure resolving 'archive.ubuntu.com'
Err:4 http://archive.ubuntu.com/ubuntu artful/main amd64 libpam-systemd amd64 234-2ubuntu12
  Temporary failure resolving 'archive.ubuntu.com'
W: Download is performed unsandboxed as root as file '/var/cache/apt/archives/partial/libdbus-1-3_1.10.22-1ubuntu1_amd64.deb' couldn't be accessed by user '_apt'. - pkgAcquire::Run (13: Permission denied)
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/d/dbus/libdbus-1-3_1.10.22-1ubuntu1_amd64.deb  Temporary failure resolving 'archive.ubuntu.com'
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/e/expat/libexpat1_2.2.3-1_amd64.deb  Temporary failure resolving 'archive.ubuntu.com'
[...]

PR #193, opened 2017-11-08
Use host resolver with host network for build script

lucaswerkmeister:
This is essentially a copy of f4ee420a00, but for the build script, since @juaningan pointed out in a comment in #188 that we need it there as well.

---

Note: I’m not sure if this is actually necessary – I haven’t been able to build a scenario that fails without this change yet. @NeilW or @juaningan can you look into this perhaps?

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -3113,7 +3113,10 @@ def run_build_script(args, workspace, raw):
             cmdline.append("--setenv=BUILDDIR=/root/build")
             cmdline.append("--bind=" + args.build_dir + ":/root/build")
 
-        if not args.with_network:
+        if args.with_network:
+            # If we're using the host network namespace, use the same resolver
+            cmdline.append("--bind-ro=/etc/resolv.conf")
+        else:
             cmdline.append("--private-network")
 
         cmdline.append("/root/" + os.path.basename(args.build_script))


mkosi
```diff
@@ -3113,7 +3113,10 @@ def run_build_script(args, workspace, raw):
             cmdline.append("--setenv=BUILDDIR=/root/build")
             cmdline.append("--bind=" + args.build_dir + ":/root/build")
 
-        if not args.with_network:
+        if args.with_network:
+            # If we're using the host network namespace, use the same resolver
+            cmdline.append("--bind-ro=/etc/resolv.conf")
```
lucaswerkmeister: Side note: I changed this from `cmdline += […]` to `cmdline.append(…)` to match the surrounding code in `run_build_script`, but I suppose at some point we should go over the whole program to reduce inconsistencies like this (e. g. the multiple ways to format strings we use), and perhaps look into some CI that makes sure we stick to one style.

keszybz:
@NeilW a refactoring would be great, but can you rebase that and submit as PR?

poettering:
hmm, what i don't get: nspawn should do that anyway, no? i mean, the logic in nspawn is there to always use the host resolv.conf. Are you saying that doesn't work?

keszybz:
Hmm, I merged this, assuming that https://github.com/systemd/mkosi/commit/f4ee420a00d138cfa3e5c965f39d0aef72e70233 is correct. But yeah, it seems systemd-nspawn should do this.

keszybz:
Ah, #185 was on F26. Maybe /etc/resolv.conf wasn't handled like that back then?

PR #196, opened 2017-11-14
Allow loose permissions on password file

keszybz:


mkosi
```diff
@@ -2644,7 +2644,7 @@ def find_password(args):
     try:
         require_private_file('mkosi.rootpw', 'root password')
 
-        with open('mkosi.rootpw', 'r', encoding="utf-8") as f:
```
poettering: are you sure this is the right thing to do? i.e. dropping the encoding= bit means we read some binary strings from it, not proper unicode strings. Unless I have misunderstood something there this doesn't look right...

poettering:
hmm, some testing suggests you are right on the encoding thing... maybe i got confused by a python2 thing?

PR #198, opened 2017-11-16
mkosi: make sure when copying files we unlink existing matching files…

poettering:
… first

Previously, if a file already existed before, we'd open it for write and
write the new file into the same file. If the old file was larger than
the new file we'd not truncate it, so that in that case the resulting
file was a combination of the new small file plus the old suffix.

This one confused the hell out of me...


keszybz:
Pfff, thanks for fixing this.

PR #206, opened 2017-11-29
mkosi: make -i mode snappier on reflink-capable file systems

poettering:
Let's start out with a reflink copy, even if we don't actually want CoW
for disk images ultimately.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -183,6 +183,14 @@ def copy_fd(oldfd, newfd):
         shutil.copyfileobj(open(oldfd, 'rb', closefd=False),
                            open(newfd, 'wb', closefd=False))
 
+def copy_file_object(oldobject, newobject):
+    try:
+        _reflink(oldobject.fileno(), newobject.fileno())
+    except OSError as e:
+        if e.errno not in {errno.EXDEV, errno.EOPNOTSUPP}:
+            raise
+        shutil.copyfileobj(oldobject, newobject)
+
 def copy_file(oldpath, newpath):
     with open_close(oldpath, os.O_RDONLY) as oldfd:
         st = os.stat(oldfd)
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -415,8 +415,22 @@ def reuse_cache_image(args, workspace, run_build_script, for_cache):
         with source:
             f = tempfile.NamedTemporaryFile(dir = os.path.dirname(args.output), prefix='.mkosi-')
             output.append(f)
+
+            # So on one hand we want CoW off, since this stuff will
+            # have a lots of random write accesses. On the other we
```
reverendhomer: Just a little nitpick: typo here. "a lot of" or "lots of" would be right.

poettering:
I force pushed a new version now that fixes the typo @reverendhomer pointed out. PTAL!

mkosi
```diff
@@ -415,8 +415,22 @@ def reuse_cache_image(args, workspace, run_build_script, for_cache):
...
+            # CoW but later changes do not result in CoW anymore.
+
             disable_cow(f.name)
-            shutil.copyfileobj(source, f)
+            try:
+                _reflink(source.fileno(), f.fileno())
+            except OSError as e:
+                if e.errno not in {errno.EXDEV, errno.EOPNOTSUPP}:
+                    raise
+                shutil.copyfileobj(source, f)
```
lucaswerkmeister: Couldn’t this use the `_copy_file` function? It looks like it does mostly the same thing, and I’m not sure if the slight differences are intended.

mkosi
```diff
@@ -415,8 +415,22 @@ def reuse_cache_image(args, workspace, run_build_script, for_cache):
...
             f = tempfile.NamedTemporaryFile(dir = os.path.dirname(args.output), prefix='.mkosi-')
             output.append(f)
+
+            # So on one hand we want CoW off, since this stuff will
+            # have a lot of random write accesses. On the other we
+            # want the copy to be snappy, hence we do want CoW. Let's
+            # ask for both, and let the kernel figure things out:
+            # let's turn off CoW on the file, but start with a CoW
+            # copy. On btrfs that works: the initial copy is made as
+            # CoW but later changes do not result in CoW anymore.
```
lucaswerkmeister: I found it a bit confusing that this long block of documentation precedes just a single call to `_reflink` – where is all this “turn off CoW but start with CoW” logic hiding? :) IIUC that’s just the behavior of the `FICLONE` ioctl, so I think some of this documentation should be placed on `_reflink` instead.

poettering: not quite. we first invoke "disable_cow()" and then follow up with "_reflink()" right after. it's that combination of first turning off cow, and then doing a reflink (which is the mother of all cow concepts) after all that deserves the comment.

mkosi
poettering: copy_file() takes a path, _copy_file() takes fd integers, but we need python file objects here...

But you are right, we should probably split this into a function of its own. Will do.

poettering:
Force pushed a new version now, which turns the copy stuff into a function of its own. please have a look. it's a bit nasty that we currently have so many different slightly different flavours of copy file routines, but cleaning that up is probably work for another PR

mkosi
lucaswerkmeister: Wait, how the fuck did I not see that `disable_cow` call? :D

Ben Mendelsohn playing Orson Krennic in “Rogue One: A Star Wars Story”, in the scene on Scarif just after the rebel attack starts, yelling at the officers surrounding him: “Are we blind? Deploy the garrison!”

PR #211, opened 2018-01-04
Re-enable kernel install scripts

lucasdemarchi:
After doing the distro installation we'd better re-enable the install
scripts from the distro so we can continue to support scripts that
install kernels following the Bootloader Spec (even though we prefer
a unified image here).

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -890,7 +890,10 @@ def check_if_url_exists(url):
     except:
         return False
 
+kernel_install_files = ()
+
 def disable_kernel_install(args, workspace):
+    global kernel_install_files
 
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
A little more information: I've been using mkosi to generate (minimal) images to test kernel changes. In this scenario I always want to add my custom modified kernel to boot either on qemu or real machine. For that I use https://github.com/lucasdemarchi/toolbox/blob/master/custom-kernel-gen which will call kernel-install. I wouldn't mind if the kernel-install hooks would generate a unified image, but right now on Fedora it creates systemd-boot entries.

PR #215, opened 2018-01-07
Add the path where the UEFI firmware blob stands on Arch Linux

L3n41c:
as shipped with the `ovmf` package.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -3384,7 +3384,8 @@ def run_shell(args):
 def run_qemu(args):
 
     # Look for the right qemu command line to use
-    ARCH_BINARIES = { 'x86_64' : 'qemu-system-x86_64' }
+    ARCH_BINARIES = { 'x86_64' : 'qemu-system-x86_64',
+                      'i386'   : 'qemu-system-i386'}
     arch_binary = ARCH_BINARIES.get(platform.machine(), None)
     for cmdline in ([arch_binary, '-machine', 'accel=kvm'],
                     ['qemu', '-machine', 'accel=kvm'],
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -3396,7 +3396,7 @@ def run_qemu(args):
         die("Couldn't find QEMU/KVM binary")
 
     # Look for UEFI firmware blob
-    for firmware in ('/usr/share/edk2/ovmf/OVMF_CODE.fd',):
+    for firmware in ('/usr/share/edk2/ovmf/OVMF_CODE.fd','/usr/share/ovmf/ovmf_code_x64.bin'):
```
poettering: hmm, that file appears to be appropriate only for x86-64, no? i figure this should be conditionalized appropriately...

L3n41c: Indeed.
The `ovmf` package comes with two firmware blobs.
I’m now selecting the right one depending on the architecture.

lucaswerkmeister:
Hm, on my system the package includes different paths:
```sh
$ pacman -Ql ovmf
ovmf /usr/
ovmf /usr/share/
ovmf /usr/share/licenses/
ovmf /usr/share/licenses/ovmf/
ovmf /usr/share/licenses/ovmf/License.txt
ovmf /usr/share/ovmf/
ovmf /usr/share/ovmf/x64/
ovmf /usr/share/ovmf/x64/OVMF_CODE.fd
ovmf /usr/share/ovmf/x64/OVMF_VARS.fd
```

poettering:
@lucaswerkmeister can you submit those paths as a PR?

lucaswerkmeister:
@poettering #240

PR #216, opened 2018-01-07
mkosi qemu doesn’t need to be run as root

L3n41c:
On my setup, once I’ve changed the owner of the image generated by `mkosi -b` (launched as root), `mkosi qemu` can successfully be run with a non-privileged user.


poettering:
hm, so this would break "mkosi -bf", i.e. the combination of building an image and running qemu with it. there's something missing here

PR #220, opened 2018-01-23
mkosi: btrfs_subvol_delete with mounted subvolumes

gdiscry:
Recursively deleting a subvolume does not work if an ancestor subvolume
is explicitly mounted (e.g. with `mount -o subvolid=<ID>`).

In such a case, the paths returned by `btrfs subvol list -o` are
specific to the BTRFS filesystem and have no relation with the mount
point.  An incorrect path is then constructed for the children and their
deletion will fail.

Instead, we retrieve the ID of the subvolumes below the current one and
query their path relative to the current subvolume with `btrfs subvol
show -r`.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -317,24 +317,19 @@ def btrfs_subvol_create(path: str, mode: int=0o755) -> None:
     os.umask(m)
 
 def btrfs_subvol_delete(path: str) -> None:
-    # Extract the path of the subvolume relative to the filesystem
-    c = run(["btrfs", "subvol", "show", path],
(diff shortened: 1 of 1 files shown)


keszybz:
Updated to resolve conficts.

keszybz:
Unfortunately this fails on older btrfs. btrfs-progs-4.11.1-3.fc27.x86_64 is bad, but btrfs-progs-4.15.1.fc28 seems good enough. What kind of backwards compat do we want to keep?

DaanDeMeyer:

PR #221, opened 2018-01-23
mkosi: try hard to cleanup the workspace

### LINKED ISSUE #182: mkosi won't clean up btrfs snapshots
I'm running mkosi to build a btrfs snapshot debian image.

It's trying to clean up the build directory (`.mkosi-sbcfbsy1/root`) by deleting it with `os.rmdir`, which fails because `/home` and `/var` (among others) are btrfs subvolumes within the build root.

It should be calling `btrfs subvolume delete` instead of `os.rmdir` when using btrfs subvolumes to do the build.

The command I ran:
```
$ mkosi -t subvolume -o debian-base -r stable -d debian
```

The btrfs subvolume listing after the build:

```
root@drcid /var/lib/sandbox $ btrfs subvolume list .
ID 288 gen 19556 top level 5 path var/lib/sandbox
...
ID 300 gen 19536 top level 288 path .mkosi-sbcfbsy1/root
ID 301 gen 19536 top level 300 path .mkosi-sbcfbsy1/root/home
ID 302 gen 18611 top level 300 path .mkosi-sbcfbsy1/root/srv
ID 303 gen 19536 top level 300 path .mkosi-sbcfbsy1/root/var
ID 304 gen 18613 top level 303 path .mkosi-sbcfbsy1/root/var/tmp
ID 305 gen 18614 top level 303 path .mkosi-sbcfbsy1/root/var/lib/machines
```

I was using [master](https://github.com/systemd/mkosi/commit/7f82d90f349ac618c41e43d7dd666eb411a4b08f). [Full logs](https://gist.github.com/zfjagann/41f4b439f4d54ddde3d563f04fe6d9e5)

gdiscry:
The workspace is a `tempfile.TemporaryDirectory` and its cleanup fails
if it cannot delete a file or directory contained in the workspace.

This is the case when an exception occurs during the build of a BTRFS
subvolume image. `tempfile.TemporaryDirectory` cannot delete the
subvolumes and the workspace remains.

The workspace is now an instance of `Workspace`, a class that wraps a
`tempfile.TemporaryDirectory`. When the workspace instance is deleted,
its contents are first deleted with unlink_try_hard and the temporary
directory can finally be cleaned up.

Fixes #182

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -23,6 +23,7 @@ import sys
 import tempfile
 import urllib.request
 import uuid
+import weakref
 
 try:
     import argcomplete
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -268,13 +269,27 @@ def init_namespace(args):
...
-        d = tempfile.TemporaryDirectory(dir='/var/tmp', prefix='mkosi-')
+        d = Workspace(dir='/var/tmp', prefix='mkosi-')
 
     print_step("Temporary workspace in " + d.name + " is now set up.")
     return d
 
+class Workspace(object):
+    def __init__(self, suffix=None, prefix=None, dir=None):
+        self._tempdir = tempfile.TemporaryDirectory(suffix=suffix, prefix=prefix, dir=dir)
+        weakref.finalize(self, self._cleanup, self._tempdir)
```
keszybz: Hmm, can you explain why `weakref` is used instead of just defining a `__del__` method?

keszybz: (sorry for the late review)

gdiscry: I think I started from the tempfile.TemporaryDirectory implementation, but I removed the `__enter__` and `__exit__` methods as they were not used. `__del__` might indeed be enough.

keszybz: I think the nicest way to implement this would be subclass TemporaryDirectory and only override .cleanup() to call `empty_directory()` first.

DaanDeMeyer:

PR #223, opened 2018-01-24
Improve the Arch Linux installation

gdiscry:
I had a few issues with the Arch Linux installation.

The custom `pacman.conf` file does not place all the configuration options under the workspace/root directory. Furthermore, several calls to pacman-key and pacman do not use the custom configuration file and touch the host system.

The package selection also has issues. It's supposed to use a slimmed down version of the `base` group, excluding a few packages that are not used by the image. However, some packages are still included even though they are not needed. For example, the kernel is installed even when the image is not bootable and even though the code looks like it's trying to exclude it.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1329,24 +1329,25 @@ def install_arch(args, workspace, run_build_script):
     if args.release is not None:
         sys.stderr.write("Distribution release specification is not supported for Arch Linux, ignoring.\n")
 
-    keyring = "archlinux"
-
-    if platform.machine() == "aarch64":
-        keyring += "arm"
-
-    run(["pacman-key", "--nocolor", "--init"], check=True)
-    run(["pacman-key", "--nocolor", "--populate", keyring], check=True)
-
     if platform.machine() == "aarch64":
         server = "Server = {}/$arch/$repo".format(args.mirror)
     else:
         server = "Server = {}/$repo/os/$arch".format(args.mirror)
 
-    with open(os.path.join(workspace, "pacman.conf"), "w") as f:
+    root = os.path.join(workspace, "root")
+    # Create base layout for pacman and pacman-key
+    os.makedirs(os.path.join(root, "var/lib/pacman"), 0o755, exist_ok=True)
+    os.makedirs(os.path.join(root, "etc/pacman.d/gnupg"), 0o755, exist_ok=True)
+
+    pacman_conf = os.path.join(workspace, "pacman.conf")
+    with open(pacman_conf, "w") as f:
         f.write("""\
 [options]
+RootDir     = {root}
 LogFile     = /dev/null
-HookDir     = /no_hook/
+CacheDir    = {root}/var/cache/pacman/pkg/
+GPGDir      = {root}/etc/pacman.d/gnupg/
+HookDir     = {root}/etc/pacman.d/hooks/
 HoldPkg     = pacman glibc
 Architecture = auto
 UseSyslog
@@ -1362,11 +1363,38 @@ SigLevel    = Required DatabaseOptional
 
 [community]
 {server}
-""".format(args=args, server=server))
+""".format(server=server, root=root))
 
-    run(["pacman", "--color", "never", "--config", os.path.join(workspace, "pacman.conf"), "-Sy"], check=True)
-    # determine base packages list from base metapackage
-    c = run(["pacman", "--color", "never", "--config", os.path.join(workspace, "pacman.conf"), "-Sg", "base"], stdout=PIPE, universal_newlines=True, check=True)
+    def run_pacman(args, **kwargs):
+        cmdline = [
+            "pacman",
+            "--noconfirm",
+            "--color", "never",
+            "--config", pacman_conf,
+        ]
+        return run(cmdline + args, **kwargs, check=True)
+
+    def run_pacman_key(args):
+        cmdline = [
+            "pacman-key",
+            "--nocolor",
+            "--config", pacman_conf,
+        ]
+        return run(cmdline + args, check=True)
+
+    def run_pacstrap(packages):
+        cmdline = ["pacstrap", "-C", pacman_conf, "-dGM", root]
+        run(cmdline + list(packages), check=True)
+
+    keyring = "archlinux"
+    if platform.machine() == "aarch64":
+        keyring += "arm"
+    run_pacman_key(["--init"])
+    run_pacman_key(["--populate", keyring])
+
+    run_pacman(["-Sy"])
+    # determine base packages list from base group
+    c = run_pacman(["-Sg", "base"], stdout=PIPE, universal_newlines=True)
     packages = set(c.stdout.split())
     packages.remove("base")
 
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -1258,24 +1258,28 @@ def install_arch(args, workspace, run_build_script):
...
         server = "Server = {}/$repo/os/$arch".format(args.mirror)
 
-    with open(os.path.join(workspace, "pacman.conf"), "w") as f:
+    root = os.path.join(workspace, "root")
+    # Create base layout for pacman-key
+    os.makedirs(os.path.join(root, "var/lib/pacman"), 0o755, exist_ok=True)
+    os.makedirs(os.path.join(root, "etc/pacman.d/gnupg"), 0o755, exist_ok=True)
+    # Fix permissions when subvolumes are used
+    os.chmod(os.path.join(root, "var/lib"), 0o755)
+    os.chmod(os.path.join(root, "var/tmp"), 0o1777)
```
lucasdemarchi: this breaks when output format is not a subvolume. Is this really fixing something? In prepare_tree() this is already the permission used.

gdiscry: I'll look into it.

The `chmod`s are just there to fix a few warnings during the installation and could be dropped or better handled.

The makedirs however is necessary to set up the keyring inside the image.

gdiscry:
So I've completely dropped the `chmod`s because while there are some warnings during the installation of the packages in a subvolume image, the final permissions are correct.

The `makedir`s remain. The first is required to run `pacman` before `pacstrap` (that would create it) and without the second, `/etc` is created with a 777 mode (when `gpg` is called by `pacman-key` I think).

The failure when unmounting `workspace/root` was caused by `pacman-key` launching an instance of `gpg-agent` holding some sockets open inside `workspace/root/etc/pacman.d/gnupg`. Once we are finished with `pacman-key`, we can kill the newly created `gpg-agent`.

poettering:
I am not an arch user myself. @lucasdemarchi does this look good to you now? if so, let's get this merged

lucasdemarchi:
I'll test again on kmod to see if nothings breaks.

lucasdemarchi:
This is still failing for me on kmod. The config I'm using is this: https://git.kernel.org/pub/scm/utils/kernel/kmod/kmod.git/tree/testsuite/mkosi/mkosi.arch?id=aca4eca103d6699543f7ed663335c28c2e9908bb

The error is:
```
‣ Mounting image...
mount: /var/tmp/mkosi-yibw6sx5/root: /dev/loop0p1 already mounted on /var/tmp/mkosi-yibw6sx5/root.
‣ Detaching image file...
  File "/usr/bin/mkosi", line 3621, in <module>
    main()
  File "/usr/bin/mkosi", line 3611, in main
    build_stuff(args)
  File "/usr/bin/mkosi", line 3445, in build_stuff
    raw, tar, root_hash = build_image(args, workspace, run_build_script=True, for_cache=True)
  File "/usr/bin/mkosi", line 3341, in build_image
    with mount_image(args, workspace.name, loopdev, encrypted_root, encrypted_home, encrypted_srv, root_read_only=True):
  File "/usr/lib64/python3.6/contextlib.py", line 81, in __enter__
    return next(self.gen)
  File "/usr/bin/mkosi", line 697, in mount_image
    mount_loop(args, root_dev, root, root_read_only)
  File "/usr/bin/mkosi", line 676, in mount_loop
    run(["mount", "-n", dev, where, options], check=True)
  File "/usr/lib64/python3.6/subprocess.py", line 418, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['mount', '-n', '/dev/loop0p1', '/var/tmp/mkosi-yibw6sx5/root', '-odiscard,ro']' returned non-zero exit status 32.
  File "/usr/lib64/python3.6/weakref.py", line 624, in _exitfunc
    f()
  File "/usr/lib64/python3.6/weakref.py", line 548, in __call__
  File "/usr/lib64/python3.6/tempfile.py", line 797, in _cleanup
    _shutil.rmtree(name)
  File "/usr/lib64/python3.6/shutil.py", line 480, in rmtree
  File "/usr/lib64/python3.6/shutil.py", line 422, in _rmtree_safe_fd
    onerror(os.rmdir, fullname, sys.exc_info())
  File "/usr/lib64/python3.6/shutil.py", line 420, in _rmtree_safe_fd
    os.rmdir(name, dir_fd=topfd)
OSError: [Errno 16] Device or resource busy: 'root'
make: *** [Makefile:3009: mkosi] Error 1

```

gdiscry:
I could not reproduce the error with the configuration used by the kmod test. I've rebased my commits just in case but the build is still successful.

@lucaswerkmeister could you run a build with these lines at the end of `mount_image` instead of just `umount(root)` ?

```python
            c = run(['find', root], stdout=PIPE, universal_newlines=True)
            run(['lsof'] + c.stdout.splitlines())
            umount(root)
```

lucaswerkmeister:
I think you mean @lucasdemarchi? :)

gdiscry:
Indeed…

lucasdemarchi:
```
COMMAND     PID USER   FD   TYPE             DEVICE SIZE/OFF   NODE NAME
dirmngr   23178 root    3u  unix 0xffff8fa4706df800      0t0 307440 /var/tmp/mkosi-cctj8xxw/root/etc/pacman.d/gnupg/S.dirmngr type=STREAM
gpg-agent 23185 root    3u  unix 0xffff8fa4ef722400      0t0 305985 /var/tmp/mkosi-cctj8xxw/root/etc/pacman.d/gnupg/S.gpg-agent type=STREAM
gpg-agent 23185 root    4u  unix 0xffff8fa4ef723c00      0t0 305986 /var/tmp/mkosi-cctj8xxw/root/etc/pacman.d/gnupg/S.gpg-agent.extra type=STREAM
gpg-agent 23185 root    5u  unix 0xffff8fa4ef720000      0t0 305987 /var/tmp/mkosi-cctj8xxw/root/etc/pacman.d/gnupg/S.gpg-agent.browser type=STREAM
gpg-agent 23185 root    6u  unix 0xffff8fa4ef721c00      0t0 305988 /var/tmp/mkosi-cctj8xxw/root/etc/pacman.d/gnupg/S.gpg-agent.ssh type=STREAM
```

along with several warnings like:
```
lsof: WARNING: can't stat() fuse.gvfsd-fuse file system /run/user/1000/gvfs
      Output information may be incomplete.
```

lucasdemarchi:
Just to make it clear... I'm running these commands on Fedora as host with pacman 5.0.2 installed and arch-install-scripts-18 installed, which seems to be the latest versions on Arch as well.

gdiscry:
Thanks for the information.

The lsof warnings can be ignored but as with the initial version, gpg-agent is still running and keeping a few sockets open. There is nothing specific to pacman or arch-install-script beyond pacman-key calling gpg with a homedir inside our image and gpg-agent remaining after the installation. This regression is entirely caused by the updated GPGDir entry in pacman.conf.

However, gpg-agent is supposed to be killed by 

```python
    # Kill the gpg-agent started by pacman-key
    run(['gpg-connect-agent', '--homedir', os.path.join(root, 'etc/pacman.d/gnupg'), 'KILLAGENT', '/bye'])
```

and you should see `OK closing connection` after the output of pacman-key.

Could you add `check=True` to that call ? I didn't want to be too strict there if everything works, but my guess would be that gpg-connect-agent fails to kill gpg-agent in your case.

I'll try to setup a Fedora machine so I can test locally. Luckily I know just the right tool for that ;)

gdiscry:
While testing, I noticed that pacman can also invoke gpg if the keyring is not up to date anymore. I moved killing gpg-agent to the end of the function, after all the calls to pacman-key, pacman and pacstrap.

What's weird is that I could not reproduce the problem inside a Fedora host with a simple Arch image.

keszybz:
This thread reminded me to update arch-install-scripts to the latest version in Fedora. kmod builds fine with that and this PR, let's merge.

Issue #82, opened 2017-05-17
Error when trying to make Debian image without specifying the release


AsciiWolf:
This strange error appears if I try to make Debian image using mkosi under Debian unstable without specifying the release manually:

![error](https://cloud.githubusercontent.com/assets/2339930/26170573/3622e2e0-3b42-11e7-8457-90aec5839193.png)

It looks like the default _args.release_ variable is not correctly parsed. This bug is not present under Ubuntu 16.04 on my another machine.

Here is the full traceback:

```
Traceback (most recent call last):
  File "/usr/local/bin/mkosi", line 2538, in <module>
    main()
  File "/usr/local/bin/mkosi", line 2534, in main
    build_stuff(args)
  File "/usr/local/bin/mkosi", line 2486, in build_stuff
    raw, tar, root_hash = build_image(args, workspace, run_build_script=False)
  File "/usr/local/bin/mkosi", line 2384, in build_image
    install_distribution(args, workspace.name, run_build_script)
  File "/usr/local/bin/mkosi", line 1015, in install_distribution
    install[args.distribution](args, workspace, run_build_script)
  File "/usr/lib/python3.5/contextlib.py", line 30, in inner
    return func(*args, **kwds)
  File "/usr/local/bin/mkosi", line 833, in install_debian
    install_debian_or_ubuntu(args, workspace, run_build_script, args.mirror)
  File "/usr/local/bin/mkosi", line 778, in install_debian_or_ubuntu
    subprocess.run(cmdline, check=True)
  File "/usr/lib/python3.5/subprocess.py", line 398, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['debootstrap', '--verbose', '--merged-usr', '--variant=minbase', '--include=systemd-sysv', '--exclude=sysv-rc,initscripts,startpar,lsb-base,insserv', '--components=main', '"9"', '/var/tmp/mkosi-488qxli3/root', 'http://httpredir.debian.org/debian']' returned non-zero exit status 1
```

fsateler:
Do you have a `mkosi.default` lying around? I have seen that happen where `/u/s/debootstrap/scripts/23` does not exist (23 being the fedora release in the `mkosi.default` file)

AsciiWolf:
@fsateler I don't.

reverendhomer:
I have the same issue. I've checked this subprocess by running it manually. `debootstrap` reports `E: unrecognized or invalid option --merged-usr`.

[...]

PR #224, opened 2018-01-24
fix release name detection on Debian hosts

hfcg-hcf:
This fixes release name detection on Debian hosts by extracting the correct codename (jessie, stretch, etc. ) from the VERSION field in the ```/etc/os-release``` file.
If this sounds to brittle, we could replace it by a dictionary lookup with the version numbers and codenames: ``` {"8": "jessie", 9": "stretch"}```.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -2313,14 +2313,27 @@ def detect_distribution():
 
     id = None
     version_id = None
+    codename = None
 
     for ln in f:
         if ln.startswith("ID="):
             id = ln[3:].strip()
         if ln.startswith("VERSION_ID="):
             version_id = ln[11:].strip()
+        if ln.startswith("VERSION="):
+            # extract Debain release codename
+            version_str = ln[8:].strip()
+            open_paren_pos = version_str.index("(")
+            close_paren_pos = version_str.index(")")
+
+            if open_paren_pos and close_paren_pos:
+                codename = version_str[(open_paren_pos+1):close_paren_pos]
 
     d = Distribution.__members__.get(id, None)
+
+    if d == Distribution.debian and codename:
+        version_id = codename
+
     return d, version_id
 
 def unlink_try_hard(path):


lucasdemarchi:
what's this **fixing** exactly? It seems it's just preferring a codename rather than version number (Debian-only). Or am I missing something?

fsateler:
@lucasdemarchi the problem is that debootstrap does not understand version numbers (#82)

hfcg-hcf:
@lucasdemarchi If you run mkosi on the latest Debian stable release (Jessie), by default (without the fedora mkosi.default file) it will try to get the current version from ```/etc/os-release```, extract the version number from the ```VERSION_ID``` field (9), pass that value to the ```debootstrap``` command and finally fail with the following error:
```
  File "mkosi", line 3442, in <module>
    main()
  File "mkosi", line 3432, in main
    build_stuff(args)
  File "mkosi", line 3330, in build_stuff
  File "mkosi", line 3178, in build_image
    install_distribution(args, workspace.name, run_build_script, cached)
  File "mkosi", line 1464, in install_distribution
    install[args.distribution](args, workspace, run_build_script)
  File "/opt/python3.6.2/lib/python3.6/contextlib.py", line 52, in inner
    return func(*args, **kwds)
  File "mkosi", line 1250, in install_debian
    install_debian_or_ubuntu(args, workspace, run_build_script, args.mirror)
  File "mkosi", line 1196, in install_debian_or_ubuntu
    run(cmdline, check=True)
  File "/opt/python3.6.2/lib/python3.6/subprocess.py", line 418, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['debootstrap', '--verbose', '--merged-usr', '--variant=minbase', '--include=systemd-sysv', '--exclude=sysv-rc,initscripts,startpar,lsb-base,insserv', '--components=main', '"9"', '/var/tmp/mkosi-u7l9_kio/root', 'http://deb.debian.org/debian']' returned non-zero exit status 1.
```
Like @fsateler commented before, ```debootstrap``` expects release code names, not version numbers.

mkosi
```diff
@@ -2313,14 +2313,27 @@ def detect_distribution():
...
         if ln.startswith("VERSION_ID="):
             version_id = ln[11:].strip()
+        if ln.startswith("VERSION="):
+            # extract Debain release codename
+            version_str = ln[8:].strip()
+            open_paren_pos = version_str.index("(")
+            close_paren_pos = version_str.index(")")
+
+            if open_paren_pos and close_paren_pos:
+                codename = version_str[(open_paren_pos+1):close_paren_pos]
```
poettering: hmm, doesn't debian set VERSION_CODENAME= for this? this was added explicitly for this purpose... if they don't set this right now, I think we should still look for it and use preferably it if it's set, and also somebody should file a bug requesting them to add this for cases like this...

mkosi
```diff
@@ -2313,14 +2313,27 @@ def detect_distribution():
...
     for ln in f:
         if ln.startswith("ID="):
             id = ln[3:].strip()
         if ln.startswith("VERSION_ID="):
             version_id = ln[11:].strip()
+        if ln.startswith("VERSION="):
+            # extract Debain release codename
+            version_str = ln[8:].strip()
+            open_paren_pos = version_str.index("(")
+            close_paren_pos = version_str.index(")")
```
poettering: there must be a more pythonic way to do something like this... in particular as it might do weird stuff if ) appears before (...

@keszybz can you suggest a pythonic way to solve this?

mkosi
poettering: see docs about this:

https://www.freedesktop.org/software/systemd/man/os-release.html#VERSION_CODENAME=

hfcg-hcf:
Thanks for the feedback. Updated the PR by adding the preferred ```VERSION_CODENAME``` field and using an hopefully more robust way to extract the codename from the ```VERSION``` field.

mkosi
```diff
@@ -2313,14 +2314,33 @@ def detect_distribution():
...
 
     for ln in f:
         if ln.startswith("ID="):
             id = ln[3:].strip()
         if ln.startswith("VERSION_ID="):
             version_id = ln[11:].strip()
+        if ln.startswith("VERSION_CODENAME="):
+            version_codename = ln[17:].strip()
+        if ln.startswith("VERSION="):
+            # extract Debain release codename
```
lucaswerkmeister: s/Debain/Debian/ :)

mkosi
```diff
@@ -2313,14 +2314,33 @@ def detect_distribution():
...
+            version_str = ln[8:].strip()
+            debian_codename_re = r'\((.*?)\)'
+
+            codename_list = re.findall(debian_codename_re, version_str)
+            if len(codename_list) == 1:
+                extracted_codename = codename_list[0]
 
     d = Distribution.__members__.get(id, None)
+
+    if d == Distribution.debian and (version_codename or extracted_codename):
```
lucaswerkmeister: perhaps a short comment after this to explain why it’s necessary? e. g. “debootstrap needs codenames, not version numbers”

mkosi
hfcg-hcf: :) Updated

mkosi
hfcg-hcf: Added

mkosi
```diff
@@ -2313,14 +2314,34 @@ def detect_distribution():
...
+        if ln.startswith("VERSION_CODENAME="):
+            version_codename = ln[17:].strip()
+        if ln.startswith("VERSION="):
+            # extract Debian release codename
+            version_str = ln[8:].strip()
+            debian_codename_re = r'\((.*?)\)'
+
+            codename_list = re.findall(debian_codename_re, version_str)
+            if len(codename_list) == 1:
+                extracted_codename = codename_list[0]
```
keszybz: Does this actually work? I see:
```
VERSION="17.04 (Zesty Zapus)"
VERSION="17.10 (Artful Aardvark)"
```
and my debian-unstable container doens't even have a `VERSION` field.

But even if the field is present, the result will be `"Zesty Zapus"`, but we need `zesty`, no?

hfcg-hcf: > Does this actually work?

Yes, for Debian. Tested on Stretch (stable) and Buster (unstable). It's true that Buster install doesn't have the ```VERSION``` field (and ```VERSION_ID``` is also missing) in the ```/etc/os-release``` but in that case ```mkosi``` already falls back into ```unstable``` - see ```load_args``` function.

Regarding Ubuntu, I have not tested but this PR shouldn't change anything on Canonical's distribution. Should I also have a look there?

PR #226, opened 2018-01-30
Clear Linux support

cmarcelo:
Add support to creating Clear Linux images.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1569,7 +1569,7 @@ def install_boot_loader_debian(args, workspace):
 def install_boot_loader_opensuse(args, workspace):
     install_boot_loader_debian(args, workspace)
 
-def install_boot_loader(args, workspace, cached):
+def install_boot_loader(args, workspace, loopdev, cached):
     if not args.bootable:
         return
 
@@ -3182,7 +3182,7 @@ def build_image(args, workspace, run_build_script, for_cache=False):
                     cached = reuse_cache_tree(args, workspace.name, run_build_script, for_cache, cached)
                     install_skeleton_trees(args, workspace.name, for_cache)
                     install_distribution(args, workspace.name, run_build_script, cached)
-                    install_boot_loader(args, workspace.name, cached)
+                    install_boot_loader(args, workspace.name, loopdev, cached)
 
                     install_extra_trees(args, workspace.name, for_cache)
                     install_build_src(args, workspace.name, run_build_script, for_cache)
--- a/mkosi
+++ b/mkosi
@@ -1482,7 +1482,10 @@ def reset_machine_id(args, workspace, run_build_script, for_cache):
 
     with complete_step('Resetting machine ID'):
         machine_id = os.path.join(workspace, 'root', 'etc/machine-id')
-        os.unlink(machine_id)
+        try:
+            os.unlink(machine_id)
+        except FileNotFoundError:
+            pass
         open(machine_id, "w+b").close()
         dbus_machine_id = os.path.join(workspace, 'root', 'var/lib/dbus/machine-id')
         try:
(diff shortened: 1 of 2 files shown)


lucasdemarchi:
LGTM. @poettering, do you want to take a look?

mkosi
```diff
@@ -982,6 +983,70 @@ def invoke_dnf(args, workspace, repositories, base_packages, boot_packages, conf
...
+        packages += ['kernel-native']
+
+    # Since swupd-extract program is usually installed with "go get",
+    # try to find it in some other locations besides PATH.
+    swupd_extract = None
+    sudo_user = os.getenv("SUDO_USER")
+    if sudo_user:
+        sudo_user_program = os.path.expanduser("~{}/go/bin/swupd-extract".format(sudo_user))
+        if os.path.exists(sudo_user_program):
+            swupd_extract = sudo_user_program
```
poettering: uh, oh, this sudo stuff here is … wild …

are we sure we want this bit? That just appears too magic to me...

mkosi
```diff
@@ -982,6 +983,70 @@ def invoke_dnf(args, workspace, repositories, base_packages, boot_packages, conf
...
+    if swupd_extract is None:
+        user_program = os.path.expanduser("~/go/bin/swupd-extract")
```
poettering: is that path in any way standardized? can't clearlinux just install this stuff like a normal binary like everybody else?

lucasdemarchi: The problem is actually installing on other distros so I can generate a Clear image from my Fedora host. The path is standard for go programs.

mkosi
lucasdemarchi: The path is standard for go programs. I find it pretty convenient for not having to edit PATH for a single go program I have on my machine. Maybe move this to a `def find_go_program()`?

poettering: hmpf, i find automatism that reach across user identity boundaries like this a bit too magic and dangerous for my tastes... Can't we find a different solution for this? Maybe add a new setting to mkosi.default that basically alters the search paths we use for all our tools? think:

```
ExtraSearchPath=$HOME/go/bin:$SUDO_HOME/go/bin
```

That way, you could make things work for your case automatically, but we wouldn't have to hardcode this into the normal logic. And it would be generically useful, and could cover all other external tools we need.

Of course, `$SUDO_HOME` isn't actually defined by sudo currently, but I think it would be fine to fake it in in our code if it's not set. I mean, faking such env vars isn't conceptually a new thing, given the xdg basedir stuff which in theory one day we could support the same way...

What do you think?

poettering: and of course, ExtraSearchPath= should be somewhat "smart", i.e. first split the specified string at colons, and then replace the env vars, and if any env vars can't be replaced because they are not set currently, and not faked either, then skip the whole item. i.e. be smarter than what the shell would do in this case, where it would replace unset env vars by the empty string...

poettering:
besides the search path logic patch looks great btw...

mkosi
lucasdemarchi: humn... that looks good to me. @cmarcelo ?

cmarcelo:
Updated the Clear Linux support without the path finding magic.

Will add ExtraSearchPaths in another PR.

DaanDeMeyer:
@cmarcelo I'm having trouble getting Clear Linux to boot in qemu with the current master branch. Did this ever work correctly for you? I'm wondering what could have changed (on Clear Linux or mkosi) to cause this regression but I feel like I've tried everything so I'm running out of ideas.

lucasdemarchi:
It worked for me, yes. It's been a while since I last used it though

lucasdemarchi:
It worked for me, yes. It's been a while since I last used it though

PR #229, opened 2018-02-01
Add "ExtraSearchPaths=" and --extra-search-paths

cmarcelo:
This variable keeps a colon-delimited list of paths to be prepended to
PATH in the context of mkosi execution. Unlike shell variable
expansion, if any path refers to an unset variable, that path will be
ignored.

Besides the environment variables, the variable SUDO_HOME can be used
to refer to the home directory of the user calling mkosi with
sudo. This allows settings like

    ExtraSearchPaths=$SUDO_HOME/go/bin

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -2262,13 +2262,16 @@ def setup_package_cache(args):
 
     return d
 
-class PackageAction(argparse.Action):
+class ListAction(argparse.Action):
     def __call__(self, parser, namespace, values, option_string=None):
         l = getattr(namespace, self.dest)
         if l is None:
             l = []
-        l.extend(values.split(","))
+        l.extend(values.split(self.delimiter))
         setattr(namespace, self.dest, l)
+
+class CommaDelimitedListAction(ListAction):
+    delimiter = ","
 
 def parse_args():
     parser = argparse.ArgumentParser(description='Build Legacy-Free OS Images', add_help=False)
@@ -2283,7 +2286,7 @@ def parse_args():
     group.add_argument('-d', "--distribution", choices=Distribution.__members__, help='Distribution to install')
     group.add_argument('-r', "--release", help='Distribution release to install')
     group.add_argument('-m', "--mirror", help='Distribution mirror to use')
-    group.add_argument("--repositories", action=PackageAction, dest='repositories', help='Repositories to use', metavar='REPOS')
+    group.add_argument("--repositories", action=CommaDelimitedListAction, dest='repositories', help='Repositories to use', metavar='REPOS')
 
     group = parser.add_argument_group("Output")
     group.add_argument('-t', "--format", dest='output_format', choices=OutputFormat.__members__, help='Output Format')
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -3498,6 +3511,44 @@ def run_qemu(args):
...
+    environ = os.environ.copy()
+    # Add a fake SUDO_HOME variable to allow non-root users specify
+    # paths in their home when using mkosi via sudo.
+    sudo_user = os.getenv("SUDO_USER")
+    if sudo_user:
+        environ["SUDO_HOME"] = os.path.expanduser("~{}".format(sudo_user))
+
+    # Don't use os.path.expandvars because it treats unset variables
+    # as empty. Here entries that refer to unset variables are ignored.
+    vars_pattern = re.compile(r'\$(\w+|\{[^}]*\})', re.ASCII)
```
poettering: urks... regexes... 

Why not just use python template strings? i.e. this stuff: https://www.python.org/dev/peps/pep-0292/

cmarcelo: No reason except I forgot about those, just changed to use it.

mkosi
```diff
@@ -3498,6 +3512,34 @@ def run_qemu(args):
...
+def expand_paths(input_paths):
+    if not input_paths:
+        return []
+
+    environ = os.environ.copy()
+    # Add a fake SUDO_HOME variable to allow non-root users specify
+    # paths in their home when using mkosi via sudo.
+    sudo_user = os.getenv("SUDO_USER")
+    if sudo_user:
+        environ["SUDO_HOME"] = os.path.expanduser("~{}".format(sudo_user))
```
poettering: i think we should leave SUDO_HOME set the way it is if it already is set, and only add it if it isn't...

mkosi
```diff
@@ -2335,6 +2337,9 @@ def parse_args():
     group.add_argument("--bmap", action='store_true', help='Write block map file (.bmap) for bmaptool usage (only raw_gpt, raw_btrfs)')
     group.add_argument("--password", help='Set the root password')
 
+    group = parser.add_argument_group("Host configuration")
+    group.add_argument("--extra-search-paths", help="List of colon-separated paths to look for programs before looking in PATH")
```
poettering: hmm, so extra_search_paths is initially a string, right? and then magically becomes a list... I don't like this very much I must say. I think it would be much nicer if we'd maintain it as list all the way.

Also I think we should keep extending the list. i.e. could you please rework this to work like the PackageAction thing, except we split by colon rather than comma for this? It think this would be tons nicer and more systematic.

i.e. it's fine to resolve the env vars in it later, but please let's at least stick to one type for the thing for the whole run

mkosi
```diff
@@ -3498,6 +3512,34 @@ def run_qemu(args):
...
+            paths.append(path)
+        except KeyError:
+            # Skip path if it uses a variable not defined.
+            pass
+    return paths
+
+def prepend_environ_path(paths):
+    real_path = os.getenv("PATH", None)
+    if real_path is not None:
+        paths.append(real_path)
```
poettering: i don't like this tbh... append to this list a string that contains multiple colon-separate elements is just ugly...

I'd much prefer if we'd append the original PATH env var to the end of the new env var as a string, and never add it to the list if you follow what I mean... maybe something like this:

```python
if len(paths) == 0:
    return

original_path = os.getenv("PATH", None)
new_path = ":".join(paths)

if original_path is not None:
    environ["PATH"] = original_path + ":" + new_path
else
    environ["PATH"] = new_path
```

But maybe there's a more pythonic way to do this? @keszybz might be able to help?

mkosi
cmarcelo: Rework to use a list action (added a separate commit as groundwork). See if that is more in tune with what you are thinking.

mkosi
cmarcelo: Done.

mkosi
cmarcelo: Done like your suggestion.

cmarcelo:
This is missing accepting multiple lines in config file. Will do that in a bit.

cmarcelo:
> This is missing accepting multiple lines in config file. Will do that in a bit.

Done.

poettering:
excellent! thanks!

PR #230, opened 2018-02-06
hostname setting fixes, and fixes to make kernel-install masking/unmasking work

poettering:
Without this mkosi is borked for my own usecase, i.e. build systemd for test purposes...


lucasdemarchi:
LGTM

poettering:
OK, let's merge it then!

PR #231, opened 2018-02-07
Add XFS root support

randomvariable:
Some may want XFS root file systems, particularly when using overlay2 FS for runc and friends on RH
derivatives.

Root size is 1300Mib in this case because of additional metadata


mkosi
```diff
@@ -56,6 +56,14 @@ class OutputFormat(Enum):
     directory = 4
     subvolume = 5
     tar = 6
+    raw_xfs = 7
+
+RAW_FORMATS = [
+  OutputFormat.raw_gpt,
```
poettering: hmm, if we add this i think it would make sense to also rename raw_gpt to raw_ext4 then, otherwise it gets too confusing...(maybe in a separate commit either before or after this one) there should be some compat however, so that when the user uses the old name we do the right thing still, even tough we'd only advertise and document the new name

mkosi
```diff
@@ -2050,7 +2068,7 @@ def secure_boot_sign(args, workspace, run_build_script, for_cache):
                 os.rename(p + ".signed", p)
 
 def xz_output(args, raw):
-    if args.output_format not in (OutputFormat.raw_btrfs, OutputFormat.raw_gpt, OutputFormat.raw_squashfs):
+    if args.output_format not in (RAW_FORMATS):
```
poettering: there shouldn't be () around this, no?

mkosi
```diff
@@ -2176,7 +2194,7 @@ def link_output(args, workspace, raw, tar):
                        'Successfully linked ' + args.output):
         if args.output_format in (OutputFormat.directory, OutputFormat.subvolume):
             os.rename(os.path.join(workspace, "root"), args.output)
-        elif args.output_format in (OutputFormat.raw_btrfs, OutputFormat.raw_gpt, OutputFormat.raw_squashfs):
+        elif args.output_format in (RAW_FORMATS):
```
poettering: no (), right?

mkosi
```diff
@@ -2927,7 +2945,7 @@ def load_args():
             die("Directory, subvolume and tar images cannot be booted.")
 
     if args.encrypt is not None:
-        if args.output_format not in (OutputFormat.raw_gpt, OutputFormat.raw_btrfs, OutputFormat.raw_squashfs):
+        if args.output_format not in (RAW_FORMATS):
```
poettering: as above

mkosi
```diff
@@ -2940,7 +2958,7 @@ def load_args():
         args.checksum = True
 
     if args.output is None:
-        if args.output_format in (OutputFormat.raw_gpt, OutputFormat.raw_btrfs, OutputFormat.raw_squashfs):
+        if args.output_format in (RAW_FORMATS):
```
poettering: as above

mkosi
```diff
@@ -56,6 +56,14 @@ class OutputFormat(Enum):
...
     subvolume = 5
     tar = 6
+    raw_xfs = 7
+
+RAW_FORMATS = [
+  OutputFormat.raw_gpt,
+  OutputFormat.raw_btrfs,
+  OutputFormat.raw_squashfs,
+  OutputFormat.raw_xfs
+]
```
poettering: not that i was a python guru, but wouldn't it be nicer to use a tuple rather than a list for this? after all this is supposed to be immutable... i.e. [] → ()

mkosi
```diff
@@ -3052,7 +3073,7 @@ def load_args():
             die("Sorry, can't acquire shell in or boot an XZ compressed image.")
 
     if args.verb == "qemu":
-        if args.output_format not in (OutputFormat.raw_gpt, OutputFormat.raw_btrfs, OutputFormat.raw_squashfs):
+        if args.output_format not in (RAW_FORMATS):
```
poettering: as above

mkosi
```diff
@@ -3121,18 +3142,18 @@ def print_summary(args):
...
 
-    if args.output_format in (OutputFormat.raw_gpt, OutputFormat.raw_btrfs, OutputFormat.raw_squashfs, OutputFormat.tar):
+    if args.output_format in (OutputFormat.raw_gpt, OutputFormat.raw_btrfs, OutputFormat.raw_squashfs, OutputFormat.raw_xfs, OutputFormat.tar):
         sys.stderr.write("        XZ Compression: " + yes_no(args.xz) + "\n")
 
     sys.stderr.write("            Encryption: " + none_to_no(args.encrypt) + "\n")
     sys.stderr.write("                Verity: " + yes_no(args.verity) + "\n")
 
-    if args.output_format in (OutputFormat.raw_gpt, OutputFormat.raw_btrfs, OutputFormat.raw_squashfs):
+    if args.output_format in (RAW_FORMATS):
```
poettering: as above, and for the rest tooo

mkosi
randomvariable: I was thinking this too, but wasn't sure if ext4 on `raw_gpt` had some special meaning for a specific purpose.

Have renamed everything to `raw_ext4`, including the command help, though `raw_gpt` is still in
the enum with the same value, so shouldn't break anything for anyone.

mkosi
randomvariable: Yes. Am totally not a Python guru. I've fixed these up throughout.

mkosi
```diff
@@ -2965,8 +2966,8 @@ def load_args():
             die("Directory, subvolume and tar images cannot be booted.")
 
     if args.encrypt is not None:
-        if args.output_format not in (OutputFormat.raw_gpt, OutputFormat.raw_btrfs, OutputFormat.raw_squashfs):
-            die("Encryption is only supported for raw gpt, btrfs or squashfs images.")
+        if args.output_format not in RAW_FORMATS:
```
keszybz: Is RAW_FORMATS defines somewhere?

mkosi
```diff
@@ -58,6 +58,15 @@ class OutputFormat(Enum):
     directory = 4
     subvolume = 5
     tar = 6
+    raw_xfs = 7
+
+RAW_RW_FS_FORMATS = (
+  OutputFormat.raw_ext4,
+  OutputFormat.raw_btrfs,
+  OutputFormat.raw_xfs
```
keszybz: Please use normal indentation.

mkosi
```diff
@@ -58,6 +58,15 @@ class OutputFormat(Enum):
...
+)
+
+RAW_FORMATS = RAW_RW_FS_FORMATS + (OutputFormat.raw_squashfs,)
```
keszybz: You could also write that as 
```
RAW_FORMATS = *RAW_RW_FS_FORMATS, OutputFormat.raw_squashfs
```
(but your version is OK too).

mkosi
```diff
@@ -3063,6 +3082,9 @@ def load_args():
     if args.output_format in (OutputFormat.raw_ext4, OutputFormat.raw_btrfs) and args.root_size is None:
         args.root_size = 1024*1024*1024
 
+    if args.output_format == (OutputFormat.raw_xfs) and args.root_size is None:
```
keszybz: Please drop () here.

mkosi
```diff
@@ -3160,18 +3182,18 @@ def print_summary(args):
     sys.stderr.write("Output nspawn Settings: " + none_to_na(args.output_nspawn_settings if args.nspawn_settings is not None else None) + "\n")
     sys.stderr.write("           Incremental: " + yes_no(args.incremental) + "\n")
 
-    if args.output_format in (OutputFormat.raw_ext4, OutputFormat.raw_btrfs, OutputFormat.raw_squashfs, OutputFormat.subvolume):
+    if args.output_format in RAW_FORMATS + (OutputFormat.subvolume,):
```
keszybz: Maybe `... in (*RAW_FORMATS, OutputFormat.subvolume)` (here and below)?

mkosi
randomvariable: Back on line [67](https://github.com/systemd/mkosi/pull/231/files#diff-6980c629516e4165bfaf96be8e5003cfR69)

randomvariable:
Hope I've addressed the badness now. Sorry for the Python fails.

mkosi
randomvariable: And now moved into the right commit.

poettering:
Oh, I just noticed that the README needs some updating for this too, to indicate that this is supported now. Any chance you can add that?

PR #235, opened 2018-02-10
Rename "Post Installation Script" to "Postinstall script"

keszybz:
The short version is pretty well established in packaging world, and
anyway "post-installation" or even "postinstallation" would be more
correct than "post installation" as an adjective.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1591,7 +1591,7 @@ def run_postinst_script(args, workspace, run_build_script, for_cache):
     if for_cache:
         return
 
-    with complete_step('Running post installation script'):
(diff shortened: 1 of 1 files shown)


poettering:
hmm, can we at least use "post-install" rather than "postinstall"? the latter sounds so … german in its simple concatenation way... ;-)

keszybz:
I think that would trip people up, because it's not clear whether to capitalize the name or not (and I think we don't want a dash in the option name). Spellings like "postinstallation" and "postinstall" seem to be at least as common as "post-install", so I think that's OK. Also, German is good ;)

PR #237, opened 2018-02-12
README.md: Update for new raw_xfs, and renamed raw_ext4 options

randomvariable:
Matches changes in #231

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,9 +8,11 @@ bells and whistles.
 
 The following output formats are supported:
 
-* Raw *GPT* disk image, with ext4 as root (*raw_gpt*)
+* Raw *GPT* disk image, with ext4 as root (*raw_ext4*)
 
 * Raw *GPT* disk image, with btrfs as root (*raw_btrfs*)
+
+* Raw *GPT* disk image, with xfs as root (*raw_xfs*)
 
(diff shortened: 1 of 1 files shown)


poettering:
excellent! thanks!

lucaswerkmeister:
I think some packages might also be missing now from the “requirements” section (whichever package contains `mkfs.xfs`, and possibly also the same for `mkfs.ext4`?), but I don’t use Fedora, so I don’t know which packages that would be.

randomvariable:
We renamed `raw_gpt` to `raw_ext4` for consistency, but otherwise there's no change. `e2fsprogs` covers the `mkfs.ext4`.
`mkfs.xfs` is covered by `xfsprogs`. I've tested Fedora [[1]](https://apps.fedoraproject.org/packages/xfsprogs), and could be wrong, but it seems it's the same package name elsewhere [[2](https://www.archlinux.org/packages/core/x86_64/xfsprogs/),[3](https://packages.debian.org/stretch/xfsprogs)].

lucaswerkmeister:
> `e2fsprogs` covers `mkfs.ext4`.

Yes, but as far as I can tell, it’s not in the README. (Not a new problem, just something I didn’t notice before.)

I’ll post a PR to add those two packages to the list (thanks a lot for the investigation on `xfsprogs`!).

PR #241, opened 2018-02-13
mkosi: fix to enable Fedora rawhide support

cjp256:
The Fedora repos require using the release version of 'rawhide' in the URLs,
while mkosi currently forces you to select a specific version number, e.g. 28.
You can not install Fedora '28', but you can install 'rawhide', using the gpg
keys for release version '28'.

These changes:
(1) Assume the latest (known) version when --release=rawhide, but warn the user
(2) Allow user to override a specific rawhide version with the option
--release=rawhide-<version>

Signed-off-by: Chris Patterson <cjp256@gmail.com>

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1076,16 +1076,22 @@ ensure that you have openssl program in your system.
 def install_fedora(args, workspace, run_build_script):
     if args.release == 'rawhide':
         last = sorted(FEDORA_KEYS_MAP)[-1]
-        die('Use numerical release for Fedora, not "rawhide"\n' +
-            '(rawhide was {} when this mkosi version was released)'.format(last))
+        warn('Assuming rawhide is version {} -- '.format(last) +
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -1076,16 +1076,22 @@ ensure that you have openssl program in your system.
 def install_fedora(args, workspace, run_build_script):
     if args.release == 'rawhide':
         last = sorted(FEDORA_KEYS_MAP)[-1]
-        die('Use numerical release for Fedora, not "rawhide"\n' +
-            '(rawhide was {} when this mkosi version was released)'.format(last))
+        warn('Assuming rawhide is version {} -- '.format(last) +
+             'You may specify otherwise with --release=rawhide-<version>')
+        args.releasever = last
+    elif args.release.startswith('rawhide-'):
```
keszybz: This is a bit of a hack, but it should work. We can always tweak the interface later.

PR #242, opened 2018-02-13
README: add e2fsprogs and xfsprogs to package list

lucaswerkmeister:


poettering:
@lucaswerkmeister our mkosi.default file is supposed to be good enough to be able to regenerate itself. 

Ideally we'd turn this into some CI test, but so far we have no CI hookup...

lucaswerkmeister:
> our mkosi.default file is supposed to be good enough to be able to regenerate itself.

Then I don’t understand some of the packages in there, like `arch-install-scripts` or `btrfs-prog`…

PR #243, opened 2018-03-02
Use path strings instead of pathlib.Path objects

filbranden:
The methods in os and shutil only accept pathlib.Path objects starting with Python 3.6, so revert those changes to keep compatibility with Python 3.5.

If use of pathlib is desirable, I'd suggest either using the native methods in pathlib.Path itself, or bumping the Python version requirement to 3.6. But given the use of pathlib was very limited, I think sticking to 3.5 for now makes sense...

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -14,7 +14,6 @@ import getpass
 import glob
 import hashlib
 import os
-import pathlib
 import platform
 import re
 import shutil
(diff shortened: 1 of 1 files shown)


lucaswerkmeister:
Ah, so that was why `mkosi` was failing on my Debian server :) yes, if at all possible let’s keep 3.5 compatibility IMO.

filbranden:
@keszybz Ping?

lucaswerkmeister:
There’s been enough time for review IMHO, so let’s merge :)

Issue #245, opened 2018-04-10
mkosi failed at installing packages inside deb-based container


andrew-shelyakov:
When i try to build deb-based container with following command:
```shell
mkosi --distribution debian --release stable --format directory --output test
```
This error occurs:
```
...
I: Base system installed successfully.
Reading package lists... Done
Building dependency tree... Done
The following additional packages will be installed:
  libdbus-1-3 libexpat1
Suggested packages:
  default-dbus-session-bus | dbus-session-bus
The following NEW packages will be installed:
  dbus libdbus-1-3 libexpat1 libpam-systemd
0 upgraded, 4 newly installed, 0 to remove and 0 not upgraded.
Need to get 675 kB of archives.
After this operation, 1759 kB of additional disk space will be used.
Get:1 http://cdn-fastly.deb.debian.org/debian stable/main amd64 libdbus-1-3 amd64 1.10.26-0+deb9u1 [194 kB]
Get:2 http://cdn-fastly.deb.debian.org/debian stable/main amd64 libexpat1 amd64 2.2.0-2+deb9u1 [83.4 kB]
Get:3 http://cdn-fastly.deb.debian.org/debian stable/main amd64 dbus amd64 1.10.26-0+deb9u1 [210 kB]
Get:4 http://cdn-fastly.deb.debian.org/debian stable/main amd64 libpam-systemd amd64 232-25+deb9u2 [188 kB]
Fetched 675 kB in 0s (738 kB/s)        
debconf: delaying package configuration, since apt-utils is not installed
dpkg: warning: 'ldconfig' not found in PATH or not executable
dpkg: warning: 'start-stop-daemon' not found in PATH or not executable
dpkg: error: 2 expected programs not found in PATH or not executable
Note: root's PATH should usually contain /usr/local/sbin, /usr/sbin and /sbin
E: Sub-process /usr/bin/dpkg returned an error code (2)
‣ Unmounting Package Cache...
‣ Unmounting Package Cache complete.
Traceback (most recent call last):
  File "/usr/bin/mkosi", line 3605, in <module>
    main()
  File "/usr/bin/mkosi", line 3595, in main
    build_stuff(args)
  File "/usr/bin/mkosi", line 3454, in build_stuff
    raw, tar, root_hash = build_image(args, workspace, run_build_script=False)
  File "/usr/bin/mkosi", line 3303, in build_image
    install_distribution(args, workspace.name, run_build_script, cached)
[...]

PR #246, opened 2018-04-10
Fix systemd/mkosi#245

andrew-shelyakov:


lucasdemarchi:
Please describe what's being fixed in the commit rather than a link to an issue.

mkosi
```diff
@@ -918,6 +918,10 @@ def run_workspace_command(args, workspace, *cmd, network=False, env={}, nspawn_p
     else:
         cmdline += ["--private-network"]
 
+    # Fix systemd/mkosi#245
```
lucasdemarchi: And no comments like this... If you are going to add a comment, tell why you need to modify the environment, not a link to an issue

lucasdemarchi:
Not needed anymore

PR #250, opened 2018-05-18
Fix file descriptor leak

### LINKED ISSUE #249: OSError: [Errno 16] Device or resource busy: 'machine-id'
Using mkosi v4-15-g9160f3e and trying to install Debian to a directory, it fails with the error below.

```
:/scratch/local$ sudo ~/src/mkosi/mkosi -d debian -t directory -o debian-sid
[…]
‣ Installing Debian complete.
‣ Assigning hostname...
‣ Assigning hostname complete.
‣ Unmounting Package Cache...
‣ Resetting machine ID...
  File "/home/pmenzel/src/mkosi/mkosi", line 3689, in <module>
    main()
  File "/home/pmenzel/src/mkosi/mkosi", line 3679, in main
    build_stuff(args)
  File "/home/pmenzel/src/mkosi/mkosi", line 3538, in build_stuff
  File "/home/pmenzel/src/mkosi/mkosi", line 3396, in build_image
    reset_machine_id(args, workspace.name, run_build_script, for_cache)
  File "/home/pmenzel/src/mkosi/mkosi", line 1613, in reset_machine_id
    os.unlink(machine_id)
OSError: [Errno 16] Device or resource busy: '/scratch/local/.mkosi-i978x_3d/root/etc/machine-id'
  File "/pkg/python-3.6.4-2/lib/python3.6/weakref.py", line 624, in _exitfunc
    f()
  File "/pkg/python-3.6.4-2/lib/python3.6/weakref.py", line 548, in __call__
  File "/pkg/python-3.6.4-2/lib/python3.6/tempfile.py", line 797, in _cleanup
    _shutil.rmtree(name)
  File "/pkg/python-3.6.4-2/lib/python3.6/shutil.py", line 480, in rmtree
  File "/pkg/python-3.6.4-2/lib/python3.6/shutil.py", line 418, in _rmtree_safe_fd
    _rmtree_safe_fd(dirfd, fullname, onerror)
  File "/pkg/python-3.6.4-2/lib/python3.6/shutil.py", line 418, in _rmtree_safe_fd
    _rmtree_safe_fd(dirfd, fullname, onerror)
  File "/pkg/python-3.6.4-2/lib/python3.6/shutil.py", line 438, in _rmtree_safe_fd
    onerror(os.unlink, fullname, sys.exc_info())
  File "/pkg/python-3.6.4-2/lib/python3.6/shutil.py", line 436, in _rmtree_safe_fd
    os.unlink(name, dir_fd=topfd)
OSError: [Errno 16] Device or resource busy: 'machine-id'
```

Accessing this file works fine though.

```
:/scratch/local$ sudo more /scratch/local/.mkosi-i978x_3d/root/etc/machine-id
dd5ea68b4f2d41a4a11d53ee71ac61d1
:/scratch/local$ findmnt /scratch/local 
TARGET         SOURCE              FSTYPE OPTIONS
/scratch/local /dev/sda2[/scratch] xfs    rw,relatime,attr2,inode64,noquota
```

lucaswerkmeister:
Fixes #249.

(The conversion of string concatenation to two writes makes this snippet more similar to the cmdline snippet below; perhaps we can later refactor it into a helper function, similar to systemd’s `write_string_file()`.)

---

@paulmenzel can you try if this fixes the issue for you?

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -837,7 +837,9 @@ def prepare_tree(args, workspace, run_build_script, cached):
     if args.bootable:
         # We need an initialized machine ID for the boot logic to work
         os.mkdir(os.path.join(workspace, "root", "etc"), 0o755)
-        open(os.path.join(workspace, "root", "etc/machine-id"), "w").write(args.machine_id + "\n")
+        with open(os.path.join(workspace, "root", "etc/machine-id"), "w") as f:
+            f.write(args.machine_id)
+            f.write("\n")
 
         os.mkdir(os.path.join(workspace, "root", "efi/EFI"), 0o700)
         os.mkdir(os.path.join(workspace, "root", "efi/EFI/BOOT"), 0o700)


keszybz:
This is the correct thing to do, in the sense that it follows python best practices, and because we use `with` just a few lines down for a similar purpose. But actually it doesn't change anything ;) The object is not stored anywhere, so it must be destroyed as soon as this line finished, and the file will get closed then. So I'd be very surprised if it had any effect on #249. Let's merge, but without the "fixes" tag.

lucaswerkmeister:
Okay – I’m not sure how memory management works in Python, so I thought the file descriptor might perhaps stay open until some unspecified point in the future where the file object is garbage collected. But according to Google CPython mainly uses reference counting, so the object would be garbage collected immediately?

keszybz:
Essentially, yes. In CPython the `File` object goes away as soon as there are no references to the object, and the file descriptor is closed when the object is destroyed. And since the object is never stored in a variable, normally the `File` object would be destroyed as soon as the `.write` function call returns. However, the `File` object can be referenced if an exception occurs, e.g. if the `write` fails and throws an exception. But in that case the whole script would abort anyway. That's why it doesn't really matter, although `with` is nicer style.

lucaswerkmeister:
Okay, thanks for the explanation!

PR #254, opened 2018-06-19
Fedora update

keszybz:


poettering:
We might want to make the same changes to the mkosi files included in the systemd tree...

PR #261, opened 2018-08-11
Add BootProtocols option

rantogno:
I've been using these changes to generate images that work on some old systems that do not support UEFI, and also on some simulation environments. Hopefully it can be useful to someone else too, so I'm starting the discussion here.

Also, @lucasdemarchi keeps poking me about submitting this.

He also pointed out that install_grub() should be inside install_boot_loader(). The main issue I see right now with that idea is that we need to properly generate the initramfs, and it's going to be different on every distro. Maybe we should just keep using kernel-install and then copy it to /boot, where grub-mkconfig can find it?

Anyway, ideas appreciated.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -151,6 +151,7 @@ GPT_ROOT_ARM           = uuid.UUID("69dad7102ce44e3cb16c21a1d49abed3")
 GPT_ROOT_ARM_64        = uuid.UUID("b921b0451df041c3af444c6f280d3fae")
 GPT_ROOT_IA64          = uuid.UUID("993d8d3df80e4225855a9daf8ed7ea97")
 GPT_ESP                = uuid.UUID("c12a7328f81f11d2ba4b00a0c93ec93b")
+GPT_BIOS               = uuid.UUID("2168614864496e6f744e656564454649")
 GPT_SWAP               = uuid.UUID("0657fd6da4ab43c484e50933c84b4f4f")
 GPT_HOME               = uuid.UUID("933ac7e12eb44f13b8440e14e2aef915")
 GPT_SRV                = uuid.UUID("3b8f842520e04f3b907f1a25a76f98e8")
@@ -159,6 +160,15 @@ GPT_ROOT_X86_64_VERITY = uuid.UUID("2c7357edebd246d9aec123d437ec2bf5")
 GPT_ROOT_ARM_VERITY    = uuid.UUID("7386cdf2203c47a9a498f2ecce45a2d6")
 GPT_ROOT_ARM_64_VERITY = uuid.UUID("df3300ced69f4c92978c9bfb0f38d820")
 GPT_ROOT_IA64_VERITY   = uuid.UUID("86ed10d5b60745bb8957d350f23d0571")
+
+# This is a non-formatted partition used to store the second stage
+# part of the bootloader because it doesn't necessarily fits the MBR
+# available space. 1MiB is more than enough for our usages and there's
+# little reason for customization since it only stores the bootloader and
+# not user-owned configuration files or kernels. See
+# https://en.wikipedia.org/wiki/BIOS_boot_partition
+# and https://www.gnu.org/software/grub/manual/grub/html_node/BIOS-installation.html
+BIOS_PARTITION_SIZE = 1024 * 1024
 
 CLONE_NEWNS = 0x00020000
 
@@ -404,7 +414,10 @@ def image_size(args: CommandLineArguments) -> int:
     if args.srv_size is not None:
         size += args.srv_size
     if args.bootable:
-        size += args.esp_size
+        if "uefi" in args.boot_protocols:
+            size += args.esp_size
+        if "bios" in args.boot_protocols:
+            size += BIOS_PARTITION_SIZE
     if args.swap_size is not None:
         size += args.swap_size
     if args.verity_size is not None:
@@ -421,14 +434,21 @@ def determine_partition_table(args: CommandLineArguments):
     pn = 1
     table = "label: gpt\n"
     run_sfdisk = False
+    args.esp_partno = None
+    args.bios_partno = None
 
     if args.bootable:
-        table += f'size={args.esp_size // 512}, type={GPT_ESP}, name="ESP System Partition"\n'
-        args.esp_partno = pn
-        pn += 1
+        if "uefi" in args.boot_protocols:
+            table += f'size={args.esp_size // 512}, type={GPT_ESP}, name="ESP System Partition"\n'
+            args.esp_partno = pn
+            pn += 1
+
+        if "bios" in args.boot_protocols:
+            table += f'size={BIOS_PARTITION_SIZE // 512}, type={GPT_BIOS}, name="BIOS Boot Partition"\n'
+            args.bios_partno = pn
+            pn += 1
+
         run_sfdisk = True
-    else:
-        args.esp_partno = None
 
     if args.swap_size is not None:
         table += f'size={args.swap_size // 512}, type={GPT_SWAP}, name="Swap Partition"\n'
@@ -903,18 +923,25 @@ def prepare_tree(args: CommandLineArguments, workspace: str, run_build_script: b
             f.write(args.machine_id)
             f.write("\n")
 
-        os.mkdir(os.path.join(workspace, "root", "efi/EFI"), 0o700)
-        os.mkdir(os.path.join(workspace, "root", "efi/EFI/BOOT"), 0o700)
-        os.mkdir(os.path.join(workspace, "root", "efi/EFI/Linux"), 0o700)
-        os.mkdir(os.path.join(workspace, "root", "efi/EFI/systemd"), 0o700)
-        os.mkdir(os.path.join(workspace, "root", "efi/loader"), 0o700)
-        os.mkdir(os.path.join(workspace, "root", "efi/loader/entries"), 0o700)
-        os.mkdir(os.path.join(workspace, "root", "efi", args.machine_id), 0o700)
-
         os.mkdir(os.path.join(workspace, "root", "boot"), 0o700)
-        os.symlink("../efi", os.path.join(workspace, "root", "boot/efi"))
-        os.symlink("efi/loader", os.path.join(workspace, "root", "boot/loader"))
-        os.symlink("efi/" + args.machine_id, os.path.join(workspace, "root", "boot", args.machine_id))
+
+        if args.esp_partno is not None:
+            os.mkdir(os.path.join(workspace, "root", "efi/EFI"), 0o700)
+            os.mkdir(os.path.join(workspace, "root", "efi/EFI/BOOT"), 0o700)
+            os.mkdir(os.path.join(workspace, "root", "efi/EFI/Linux"), 0o700)
+            os.mkdir(os.path.join(workspace, "root", "efi/EFI/systemd"), 0o700)
+            os.mkdir(os.path.join(workspace, "root", "efi/loader"), 0o700)
+            os.mkdir(os.path.join(workspace, "root", "efi/loader/entries"), 0o700)
+            os.mkdir(os.path.join(workspace, "root", "efi", args.machine_id), 0o700)
+
+            os.symlink("../efi", os.path.join(workspace, "root", "boot/efi"))
+            os.symlink("efi/loader", os.path.join(workspace, "root", "boot/loader"))
+            os.symlink("efi/" + args.machine_id, os.path.join(workspace, "root", "boot", args.machine_id))
+
+        if args.bios_partno is not None:
+            os.makedirs(os.path.join(workspace, "root", "etc/default"), exist_ok=True)
+            with open(os.path.join(workspace, "root", "etc/default/grub"), "w+") as f:
+                f.write("GRUB_CMDLINE_LINUX=\"" + args.kernel_commandline + "\"")
 
         os.mkdir(os.path.join(workspace, "root", "etc/kernel"), 0o755)
 
@@ -1005,8 +1032,9 @@ def disable_kernel_install(args: CommandLineArguments, workspace: str) -> List[s
     # signed as a single EFI executable. Since the root hash is only
     # known when the root file system is finalized we turn off any
     # kernel installation beforehand.
-
-    if not args.bootable:
+    #
+    # For BIOS mode, we don't have that option, so do not mask the units
+    if not args.bootable or args.bios_partno is not None:
         return []
 
     for d in ("etc", "etc/kernel", "etc/kernel/install.d"):
@@ -1047,6 +1075,9 @@ def make_rpm_list(args: argparse.Namespace, packages: List[str]) -> List[str]:
 
         if args.output_format == OutputFormat.gpt_btrfs:
             packages += ['btrfs-progs']
+
+        if args.bios_partno:
+            packages += ["grub2-pc"]
 
     return packages
 
@@ -1832,6 +1863,29 @@ def find_kernel_file(workspace_root, pattern):
         warn('More than one kernel file found, will use {}', kernel_file)
     return kernel_file
 
+def install_grub(args, workspace, loopdev, grub):
+    if args.bios_partno is None:
+        return
+
+    nspawn_params = [
+        "--bind-ro=/dev",
+        "--property=DeviceAllow=" + loopdev,
+        "--property=DeviceAllow=" + partition(loopdev, args.root_partno),
+    ]
+
+    run_workspace_command(
+            args, workspace, f"{grub}-install",
+            "--modules=ext2 part_gpt", "--target=i386-pc",
+            loopdev, nspawn_params=nspawn_params)
+
+    run_workspace_command(
+            args, workspace, f"{grub}-mkconfig",
+            f"--output=/boot/{grub}/grub.cfg",
+            nspawn_params=nspawn_params)
+
+def install_boot_loader_fedora(args, workspace, loopdev):
+    install_grub(args, workspace, loopdev, "grub2")
+
 def install_boot_loader_arch(args, workspace):
     patch_file(os.path.join(workspace, "root", "etc/mkinitcpio.conf"),
                lambda line: "HOOKS=\"systemd modconf block sd-encrypt filesystems keyboard fsck\"\n" if line.startswith("HOOKS=") and args.encrypt == "all" else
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -3501,6 +3500,7 @@ def build_stuff(args):
     # always the same
     args.machine_id = uuid.uuid4().hex
 
+    make_output_dir(args)
```
lucasdemarchi: We have this:

```python
    if args.cache_path is not None:
        args.cache_path = os.path.abspath(args.cache_path)
```

And handling the case it's None by creating a temporary directory which gets mounted in the final image... And it looks for a `mkosi.cache` dir when there's no option passed (either as a command line or in the configuration).

I couldn't find a case in which the cache is inside the the output dir, so I'm a little confused about what you are fixing. Could you share the steps to reproduce or backtrace ?

lucasdemarchi:
I think it's worth mentioning that the final image will still have a GPT partition table so we can continue using the discoverable partition scheme we rely on.  Since I think we should default to systemd-boot as bootloader, I suggested to only swap it with grub if there's this --legacy flag applied... Ideally we will still use the Bootloader Spec even with grub.

In my work with graphics in Linux I'm using mkosi to generate images to test new and old hardware. Some of the old hardware doesn't support UEFI, hence my desire to add a flag to allow this. @poettering let me know what you think of the idea in general. I think it doesn't conflict much with the "legacy-free motto", even if the flag itself is called `legacy` :). All in all, it should also help the case in which we are generating a image for RPI and other embedded boards (although not the case here).

lucasdemarchi:
@rantogno I added a comment for the bug fix... not sure I understand what's being fixed. As for the main part, I will let @poettering to give a thumbs up for the idea before diving into the implementation details. Thanks for sending the patch.

mkosi
rantogno: With the regular mkosi.default that is in the repo, if I just do something like:

$ sudo ./mkosi -O images

And the "images" directory doesn't exist yet, then it will fail. I can open an issue with the proper backtrace if you don't see the same failure when running that command.

lucasdemarchi:
@poettering ping

mkosi
poettering: i'd like to understand this better, tbh, which call precisely will fail?

mkosi
```diff
@@ -85,6 +85,7 @@ GPT_ROOT_ARM           = uuid.UUID("69dad7102ce44e3cb16c21a1d49abed3")
 GPT_ROOT_ARM_64        = uuid.UUID("b921b0451df041c3af444c6f280d3fae")
 GPT_ROOT_IA64          = uuid.UUID("993d8d3df80e4225855a9daf8ed7ea97")
 GPT_ESP                = uuid.UUID("c12a7328f81f11d2ba4b00a0c93ec93b")
+GPT_BIOS               = uuid.UUID("2168614864496e6f744e656564454649")
```
poettering: hmm, is that a gpt partition type uuid used elsewhere? or is it made up for this?

I'd feel more comfortable if this would at least use the logic described here: https://systemd.io/BOOT_LOADER_SPECIFICATION, i.e. use a $BOOT partition that uses the non-ESP boot id `bc13c2ff-59e6-4262-a352-b275fd6f7172`

mkosi
```diff
@@ -1683,9 +1705,15 @@ def install_boot_loader_arch(args, workspace):
...
-    run_workspace_command(args, workspace, "/usr/bin/kernel-install", "add", kernel_version, find_kernel_file(workspace_root, "/boot/vmlinuz-*"))
+    if args.esp_partno:
+        workspace_root = os.path.join(workspace, "root")
+        kernel_version = next(filter(lambda x: x[0].isdigit(), os.listdir(os.path.join(workspace_root, "lib/modules"))))
+        run_workspace_command(args, workspace, "/usr/bin/kernel-install", "add", kernel_version, find_kernel_file(workspace_root, "/boot/vmlinuz-*"))
+    else:
+        # If we install the initramfs with kernel-install, it goes into a
+        # subdirectory of /boot that won't be found by grub. So we use
+        # mkinitcpio directly instead.
+        run_workspace_command(args, workspace, "mkinitcpio", "-p", "linux")
```
poettering: isn't that something to fix in arch, not here?

poettering:
So, I really don#t like grub, but I figure this is OK to go in anyway, the patch isn't particularly intrusive after all. 

I think it would make a ton of sense to design this stuff so that a single image could be used to boot MBR and GPT, and possibly others (for example the raspberry pi protocol comes to mind). As such I think it might make sense to replace the LegacyBoot= stanza with "BootProtocols=" or so, which would take a list of "uefi", "bios" and possibly future other stuff, and would default to "uefi".

lucasdemarchi:
@poettering thanks. I have some fixes to this PR I've been collecting but didn't push.

I agree wrt using the bootloader spec. Unfortunately last time I tried, even on Fedora 29 it was not working. I need to take a deeper look on what was the issue.

Regarding grub, it's one readily available in distros that boots on bios. Would you prefer something else like syslinux?

poettering:
> @poettering thanks. I have some fixes to this PR I've been collecting but didn't push.
> 
> I agree wrt using the bootloader spec. Unfortunately last time I tried, even on Fedora 29 it was not working. I need to take a deeper look on what was the issue.
> 
> Regarding grub, it's one readily available in distros that boots on bios. Would you prefer something else like syslinux?

Well, whatever works. ideally something that supports the boot loader spec natively, so that we can share that stuff.

poettering:
(but yeah, any boot loader that doesn't require us to run a script to generate a script for the boot loader is better than grub)

mkosi
lucasdemarchi: https://en.wikipedia.org/wiki/GUID_Partition_Table#Partition_type_GUIDs

the scheme being implemented here is roughly following the description at https://wiki.archlinux.org/index.php/GRUB#GUID_Partition_Table_.28GPT.29_specific_instructions

mkosi
lucasdemarchi: I moved this commit to #277 with a better commit message

lucasdemarchi:
@poettering @rantogno  The BootProtocols switched is now implemented. I ended up redoing several parts on this in order to have it properly working. For now I enabled it on Fedora only, but it should be pretty easy to add Arch back.

Commands to test this and more details are in the commit message of the second commit.

lucasdemarchi:
What I didn't do is to convert the grub configuration to use BLS. I will take a look later, but last time I tried, even on F29, it was producing an empty boot menu.

mkosi
```diff
@@ -2711,6 +2713,9 @@ def process_setting(args: CommandLineArguments, section: str, key: str, value: A
         elif key == "Bootable":
             if args.bootable is None:
                 args.bootable = parse_boolean(value)
+        elif key == "BootProtocols":
+            list_value = value if type(value) == list else value.split()
```
keszybz: The command line option should override, not extend this.

mkosi
```diff
@@ -3113,6 +3118,16 @@ def load_args() -> CommandLineArguments:
         if args.output_format in (OutputFormat.directory, OutputFormat.subvolume, OutputFormat.tar):
             die("Directory, subvolume and tar images cannot be booted.")
 
+        args.boot_protocols = list(set(args.boot_protocols))
```
keszybz: Maybe
```python
if not {'uefi', 'bios'}.issuperset(args.boot_protocols):
     die(...)
```
?

mkosi
```diff
@@ -3113,6 +3118,16 @@ def load_args() -> CommandLineArguments:
...
+        for b in args.boot_protocols:
+            if b not in ("uefi", "bios"):
+                die("Not a valid boot protocol")
+        if len(args.boot_protocols) == 0:
```
```python
if not args.boot_protocols:
    ...
```
?

mkosi
```diff
@@ -1744,6 +1769,44 @@ def find_kernel_file(workspace_root, pattern):
...
+        "--property=DeviceAllow=" + partition(loopdev, args.root_partno),
+    ]
+
+    run_workspace_command(
+            args, workspace, "%s-install" % grub,
+            "--modules=ext2 part_gpt", "--target=i386-pc",
+            loopdev, nspawn_params=nspawn_params)
+
+#    run_workspace_command(args, workspace, "/bin/bash",
+#                          nspawn_params=nspawn_params)
```
keszybz: Left-over debugging code?

mkosi
```diff
@@ -1744,6 +1769,44 @@ def find_kernel_file(workspace_root, pattern):
...
+    run_workspace_command(
+            args, workspace, "%s-mkconfig" % grub,
```
keszybz: Please use `.format`.

keszybz:
Maybe grub+bls doesn't work because grub uses different paths?

lucasdemarchi:
> Maybe grub+bls doesn't work because grub uses different paths?

I got it working today. However I'm not sure I like it. I had to remove the symlink /boot/loader and duplicate the entries from systemd-boot to grub. That's because each of them expect the entries to be in a different partition.

mkosi
```diff
@@ -117,6 +118,8 @@ GPT_ROOT_ARM_VERITY    = uuid.UUID("7386cdf2203c47a9a498f2ecce45a2d6")
 GPT_ROOT_ARM_64_VERITY = uuid.UUID("df3300ced69f4c92978c9bfb0f38d820")
 GPT_ROOT_IA64_VERITY   = uuid.UUID("86ed10d5b60745bb8957d350f23d0571")
 
+BIOS_PARTITION_SIZE = 1024 * 1024
```
poettering: this should probably be configurable like the other sizes...

mkosi
```diff
@@ -1744,6 +1769,44 @@ def find_kernel_file(workspace_root, pattern):
...
+            loopdev, nspawn_params=nspawn_params)
+
+#    run_workspace_command(args, workspace, "/bin/bash",
+#                          nspawn_params=nspawn_params)
+    run_workspace_command(
+            args, workspace, "%s-mkconfig" % grub,
+            "--output=/boot/%s/grub.cfg" % grub,
+            nspawn_params=nspawn_params)
+
+    cmdline = ("blkid", "--match-tag=PARTUUID", "--output=value",
```
poettering: to me it sounds the wrong way round, to create the partition with a random uuid, and then query it off the image again. I think it would be nicer to do this the other way round: pick the uuid in mkosi before calling sfdisk, and then simply remember it and use it here...

poettering:
also, needs a rebase

mkosi
lucasdemarchi: Not sure about that... This is just a workaround for not having enough space to fit grub in  (the workaround for MBR partition is to leave a gap before the first partition, which is not possible for GPT). The use case is different than e.g. ESP in which you are going to put your kernels there.

lucasdemarchi:
Rebased and handled the comments. This is still not using BLS.  We can do that on top.  All my attempts to make systemd-boot and grub share the config files failed. I think the only reasonable thing to do to support BLS is:
- if it's uefi-only, use systemd-boot
- if it's bios or uefi + bios, use grub

@poettering @keszybz can you take a look on the HACK commit on top? This wasn't happening for me last week even if I tried to reproduce it in the same point of the tree this was based on.  My guess it was caused by an update of the systemd-nspawn package (I'm on F29). Interesting that /run in the workspace command is not being treated as a tmpfs.

mkosi
```diff
@@ -3240,8 +3291,8 @@ def load_args() -> CommandLineArguments:
             args.boot_protocols = ["uefi"]
         if not {"uefi", "bios"}.issuperset(args.boot_protocols):
             die("Not a valid boot protocol")
-        if "bios" in args.boot_protocols:
-            warn("bios boot not implemented yet")
+        if "bios" in args.boot_protocols and args.distribution not in (Distribution.fedora,):
+            die("bios boot not implemented yet for %s" % args.distribution)
```
keszybz: Please use an f-string here too. %s is too ugly.

mkosi
```diff
@@ -1871,6 +1871,10 @@ def install_grub(args, workspace, loopdev, grub):
             "--modules=ext2 part_gpt", "--target=i386-pc",
             loopdev, nspawn_params=nspawn_params)
 
+    # HACK: systemd-nspawn bug? Without this it fails with
+    # bind(/run/systemd/nspawn/notify) failed: Address already in use
+    unlink_try_hard(os.path.join(os.path.join(workspace, "root", "run/systemd/nspawn/notify")))
```
keszybz: Yep, I can confirm that I get the same issue without this HACK.

But I'd prefer that we merge this without this patch, and figure out what is wrong in `systemd-nspawn` and patch it there.

mkosi
```diff
@@ -1832,6 +1856,29 @@ def find_kernel_file(workspace_root, pattern):
...
+
+    run_workspace_command(
+            args, workspace, f"{grub}-install",
+            "--modules=ext2 part_gpt", "--target=i386-pc",
+            loopdev, nspawn_params=nspawn_params)
+
+    run_workspace_command(
+            args, workspace, f"{grub}-mkconfig",
+            f"--output=/boot/{grub}/grub.cfg",
+            nspawn_params=nspawn_params)
```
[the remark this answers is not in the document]
keszybz: @poettering just merge that other PR. So this will need a rebase. Sorry!

mkosi
```diff
@@ -2841,6 +2843,9 @@ def process_setting(args: CommandLineArguments, section: str, key: str, value: A
         elif key == "Bootable":
             if args.bootable is None:
                 args.bootable = parse_boolean(value)
+        elif key == "BootProtocols":
+            if not args.boot_protocols:
+                args.boot_protocols = value if type(value) == list else value.split()
```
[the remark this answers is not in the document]
lucasdemarchi: It's the same approach used for Packages, but here cmdline args overrides what's in the config file like you asked in the previous revision  (differently from packages, that extends what's there).

mkosi
```diff
@@ -2567,6 +2567,8 @@ def parse_args() -> CommandLineArguments:
     group.add_argument('-f', "--force", action='count', dest='force_count', default=0, help='Remove existing image file before operation')
     group.add_argument('-b', "--bootable", type=parse_boolean, nargs='?', const=True,
                        help='Make image bootable on EFI (only gpt_ext4, gpt_btrfs, gpt_squashfs, gpt_xfs)')
+    group.add_argument("--boot-protocols", choices=("uefi", "bios"), default=[], nargs="+",
+                       help="Boot protocols to use on a bootable image", metavar="PROTOCOLS")
```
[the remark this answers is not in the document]
lucasdemarchi: Then error out on passing --bootable but having no boot-protocol?  I'd expect people to use uefi and only used they bios-only or hybrid version in awful scenarios like I have of supporting old hardware.

mkosi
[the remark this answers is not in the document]
lucasdemarchi: this is done already, see the update.

mkosi
lucasdemarchi: I forgot about this one, will do.

mkosi
lucasdemarchi: yes, I can remove this just before merging.

mkosi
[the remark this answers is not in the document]
lucasdemarchi: I will update to use CommaDelimitedListAction, but I'm unsure about creating the dependencies here. For me it's nice to leave the boot protocols in the config file and decide if I'm creating a bootable/non-bootable image in the command line. If one option implies the other I think it will be tricky.

lucasdemarchi:
> `sudo qemu` doesn't work under wayland. So I prefer `chmod a+rw image.raw` or `chown ... image.raw` and `qemu` sans `sudo`. OTOH, `sudo qemu` is one command.

Maybe we should do it by default? we already poke SUDO_USER to generated Clear... we could automatically chown to $SUDO_USER or chmod a+rw.

lucasdemarchi:
> @poettering @keszybz can you take a look on the HACK commit on top? This wasn't happening for me last week even if I tried to reproduce it in the same point of the tree this was based on. My guess it was caused by an update of the systemd-nspawn package (I'm on F29). Interesting that /run in the workspace command is not being treated as a tmpfs.

More interesting stuff: if I systemd-nspawn bash I can see /run is mounted as tmpfs. However surrounding the call to systemd-nspawn with ls:

```bash
XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
ls: cannot access '/var/tmp/mkosi-yn3tfrm1/root/run/systemd/nspawn/notify': No such file or directory
Installing for i386-pc platform.
Installation finished. No error reported.
XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
srwxr-xr-x 1 root root 0 Nov  6 21:06 /var/tmp/mkosi-yn3tfrm1/root/run/systemd/nspawn/notify
```

So... basically the notify unix socket is surviving on the root mount point even after systemd-nspawn returns. I couldn't find a way to reproduce it outside mkosi, even executing systemd-nspawn several times with the same arguments. /me confused.

keszybz:
> Maybe we should do it by default?

I think so.

poettering:
> `sudo qemu` doesn't work under wayland.

It does, after "xhost +"...

mkosi
poettering: hu? /run in the container should be mounted from tmpfs anyway, so why would the dir in the workspace actually work? /me is very confused...

which nspawn version is that? something very recent?

mkosi
poettering: hmm, if this is a grub-specific concept, then I think we should rename this to GPT_GRUB_BIOS or so

mkosi
poettering: could you add a comment about that to the definition?

mkosi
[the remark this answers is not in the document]
poettering: i think we should default to uefi here, really. both because that's clearly the future, and because of compat

mkosi
lucasdemarchi: See more comments below... Sneaking a `ls` around the call to systemd-boot it's possible to see the file "leaks" outside. Version is the one in F29, but the commit hash in the package doesn't exist in systems repo. I hadn't this problem last week so it's probably between 239 and this one in Fedora.

keszybz: I messed up the github push. The version in F29/F30 is exactly the same as systemd-stable/v239-stable tip now. The right thing was to rebuild the package in Fedora, but it takes a while and I hoped nobody would notice ;)

keszybz:
> It does, after "xhost +"...

That's not nice, and also, why run privileged, when non-privileged works better? Chowning the image is generally superior UX, not only for qemu, but for other operations on the image.

mkosi
lucasdemarchi: @keszybz now you probably have a reason to rebuild the package. I seems like the problem was a backport to stable.

You need either this: https://gist.github.com/cd092aa30f58e5ff2ccde6ef4b725fce

Or to bring https://github.com/systemd/systemd/commit/44ed5214add to stable. The problem was created when https://github.com/systemd/systemd/commit/c192917cab7 was backported without the other commit... so `sa.un.sun_path` would be NULL and we would end up not unlinking the socket.

mkosi
lucasdemarchi: Although it was originally grub-specific, I'm not sure if it still is or if it's used by any other bootloader. Every reference to this partition I've seen refers to it as "BIOS boot partition" not "BIOS Grub partition". Examples: https://en.wikipedia.org/wiki/BIOS_boot_partition and the output from sfdisk we use for partitioning:

```
>>> Created a new GPT disklabel (GUID: 0A2B60E5-6C24-074F-B136-3FAEDEE761D6).
/home/lucas/tmp/.mkosi-37m4cqlt1: Created a new partition 1 of type 'EFI System' and of size 256 MiB.
/home/lucas/tmp/.mkosi-37m4cqlt2: Created a new partition 2 of type 'BIOS boot' and of size 1 MiB.
/home/lucas/tmp/.mkosi-37m4cqlt3: Created a new partition 3 of type 'Linux root (x86-64)' and of size 1 GiB.
```

If you still think it's better as GPT_GRUB_BIOS, I'm ok. Or I can add a comment there about this.

mkosi
lucasdemarchi: sure

poettering:
(btw, also, please rebase and repush. there's now a CI system running, but it will only work if PR is rebased on current master)

lucasdemarchi:
@keszybz @poettering  can you take a look on https://github.com/systemd/systemd-stable/pull/12 in order to move this PR forward?

mkosi
lucasdemarchi: done

mkosi
lucasdemarchi: done

mkosi
lucasdemarchi: This is solved by backporting said commit to systemd-nspawn - tested on F29 with the additional commit and works fine.

PR #263, opened 2018-09-07
Optimize size of Debian rootfs

terceiro:
On my tests, removing the documentation installed by debootstrap reduces a
minimal Debian stable rootfs from ~204 MB to ~157 MB

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1312,35 +1312,34 @@ def install_debian_or_ubuntu(args, workspace, run_build_script, mirror):
     if args.bootable:
         extra_packages += ["linux-image-amd64", "dracut"]
 
-    if extra_packages:
-        # Debian policy is to start daemons by default.
-        # The policy-rc.d script can be used choose which ones to start
-        # Let's install one that denies all daemon startups
-        # See https://people.debian.org/~hmh/invokerc.d-policyrc.d-specification.txt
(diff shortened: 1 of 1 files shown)


lucaswerkmeister:
Thanks!

mkosi
```diff
@@ -1312,35 +1312,39 @@ def install_debian_or_ubuntu(args, workspace, run_build_script, mirror):
...
+    if not args.with_docs:
+        # remove documentation installed by debootstrap
+        cmdline = ["rm", "-rf"] + doc_paths
+        run_workspace_command(args, workspace, *cmdline)
+        # Create dpkg.cfg to ingore documentation on new packages
+        dpkg_conf = os.path.join(workspace, "root/etc/dpkg/dpkg.cfg.d/01_nodoc")
+        with open(dpkg_conf, "w") as f:
+            f.writelines(["path-exclude %s/*\n" % d for d in doc_paths])
+
+    cmdline = ["/usr/bin/apt-get", "--assume-yes", "--no-install-recommends", "install"] + extra_packages
```
fsateler: Wouldn't this invoke apt-get even when extra_packages is not defined?

lucaswerkmeister: `extra_packages` is initialized with a nonempty list in L1298, so unless I’m mistaken, it will never be empty here.

PR #264, opened 2018-10-06
Fix archlinux

lucasdemarchi:
With these commits I can generate an archlinux image to test kmod again.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -742,10 +742,6 @@ def mount_image(args, workspace, loopdev, root_dev, home_dev, srv_dev, root_read
         yield
     finally:
         with complete_step('Unmounting image'):
-
-            for d in ("home", "srv", "efi", "run", "tmp"):
-                umount(os.path.join(root, d))
-
             umount(root)
 
 @complete_step("Assigning hostname")
@@ -809,7 +805,7 @@ def mount_cache(args, workspace):
 
 def umount(where):
     # Ignore failures and error messages
-    run(["umount", "-n", where], stdout=DEVNULL, stderr=DEVNULL)
+    run(["umount", "--recursive", "-n", where], stdout=DEVNULL, stderr=DEVNULL)
 
 @complete_step('Setting up basic OS tree')
 def prepare_tree(args, workspace, run_build_script, cached):
--- a/mkosi
+++ b/mkosi
@@ -1496,6 +1496,10 @@ SigLevel    = Required DatabaseOptional
     with open(os.path.join(workspace, 'root', 'etc/locale.conf'), 'w') as f:
         f.write('LANG=en_US.UTF-8\n')
 
+    # At this point, no process should be left running, kill then
+    run(["fuser", "-c", root, "--kill"])
+
+
 @complete_step('Installing openSUSE')
 def install_opensuse(args, workspace, run_build_script):
 


mkosi
```diff
@@ -1496,6 +1492,10 @@ SigLevel    = Required DatabaseOptional
     with open(os.path.join(workspace, 'root', 'etc/locale.conf'), 'w') as f:
         f.write('LANG=en_US.UTF-8\n')
 
+    # At this point, no process should be left running, kill then
+    run(["fuser", "-c", root, "--kill"])
```
fsateler: on line 1480 you are killing gpg explicitly, and your comment says dirmngr stays alive.Why is that not a bug in gpg?

lucasdemarchi:
> I'm not sure killing everything is the right solution. Anything we start we should be able to stop before completing, no?

We are stopping everything, but pacstrap may start other things even indirectly and they are not being properly stopped.  This is a last measure if anything happened to be left running, because there's no way the unmount that follows this phase will succeed  (and currently it actually silently fails).

mkosi
lucasdemarchi: Are they the same thing? I do see a "bye" as a result from this command. I thought it was another process that was left running.

keszybz: Seems to be a bug in gpg. Killing stuff like that is too ugly.

lucasdemarchi: @falconindy any pointer on the pacstrap/pacman side where the fix for this problem would be? For context, it's agout dirmngr left running after those command return.

poettering:
I am not sure I like the first patch, it feels like something pacstrap should really fix.

The second one makes a ton of sense though. Any chance you can split that one off at least so that we can merge it quickly?

lucasdemarchi:
But then I think we should also start checking by the return of umount.

Biggest problems for me are that 1) we already tried and it broke again. I guess it depends on packages you install; 2) we just try to continue assuming umount succeeded and then fail in a unrelated point. 

I considered killing everything as a sanity thing so this doesn't keep  breaking. We can't umount the for anyway.

poettering:
I'd feel more comfortable if we'd instead of killing evreything that access a mount would just kill everything we forked off. specifically, that we modify run() to allocate a new process group for forked off children and then kill that after the main child exits.

lucasdemarchi:
@poettering for all calls to `subprocess.run()` or only the `run_*`() inside `install_arch()`?

poettering:
I think doing the same for every command we call would be pretty nice.

poettering:
(i mean, rpm and dpkg might result in children anyway, if scriptlets do weird stuff and fork something off in the bg, hence i think it's actually a really good idea if we kill it all.)

lucasdemarchi:
so... I think we need to implement our own `run()` to use `subprocess.Popen` rather than `subprocess.run()`. `Popen` has a creationflags argument that accepts CREATE_NEW_PROCESS_GROUP for calling setpgrp().  For me this is the cleaner way to do it. Do you see another one?

poettering:
i am not a python guru. maybe @keszybz has an idea?

lucasdemarchi:
oh, well.. That was added in python 3.7 only :-/

keszybz:
> Changed in version 3.7: Added the `text` parameter, as a more understandable alias of `universal_newlines`. 

Yay, finally.

keszybz:
```pycon
>>> subprocess.Popen('/bin/ls', startupinfo=...)
  File "/usr/lib64/python3.7/subprocess.py", line 670, in __init__
    raise ValueError("startupinfo is only supported on Windows "
ValueError: startupinfo is only supported on Windows platforms
```
The docs say so too.

lucasdemarchi:
I implemented using the preexec_fn callable, setting the pgrpid and then later killing all process.

However it does't work. The 2 binaries left running daemonize so killing on the same pgrpid doesn't work.

keszybz:
So my vote would be to accept your original patch then. It's a bit ugly, but meh.

keszybz:
I'll merge this now since it makes arch functional again. We can always merge a more comprehensive solution in the future.

PR #265, opened 2018-10-08
Allow to run on qemu >= 2.12

lucasdemarchi:


lucasdemarchi:
@cmarcelo  this should fix it on Clear.

mkosi
```diff
@@ -148,6 +148,10 @@ def roundup512(x):
 def print_step(text):
     sys.stderr.write("‣ \033[0;1;39m" + text + "\033[0m\n")
 
+def print_running_cmd(cmdline):
+    sys.stderr.write("‣ \033[0;1;39mRunning command:\033[0m\n")
+    sys.stderr.write(" ".join(cmdline) + "\n")
```
cmarcelo: Depending on what characters are in each entry of the command line, you might need to escape them. So the output as-is might be wrong in some cases (e.g. `args.output` with spaces in it). Not sure if it matters in practice for the case at hand.

cmarcelo:
Checked and works in Clear. Thanks!

LGTM.

mkosi
lucasdemarchi: what about 
```python
" ".join(shlex.quote(x) for x in cmdline))
```
?

cmarcelo: Thanks. LGTM.

PR #267, opened 2018-10-16
Ubuntu images can now be bootable

thedrow:
Since the only thing that was blocking this feature is correcting the package name, I adjusted it.
Ubuntu packages should now be bootable.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -40,7 +40,6 @@ if sys.version_info < (3, 5):
 
 # TODO
 # - volatile images
-# - make ubuntu images bootable
 # - work on device nodes
 # - allow passing env vars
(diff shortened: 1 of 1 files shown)


thedrow:
We tested it ourselves and for some reason it still doesn't work.
I'm not sure what else is missing.

lucaswerkmeister:
Well, if you limit the commit to the package change, I guess this would still be useful to merge… (haven’t tested it myself yet)

eliasp:
Please see #268 which should fix all this...

keszybz:
#268 has some additional changes, let's continue the discussion there.

PR #268, opened 2018-10-17
Make Ubuntu GPT images bootable

eliasp:
As today the discussion about making Ubuntu GPT images bootable came up on IRC again, I remembered having fixed this problem before but just never pushed it + filed a PR.

So here it is...

This basically overrides #267 which lacks:
- the fix for the dracut/crc32 issue
- `install_boot_loader_ubuntu()`
- the handling of the `universe` repository required for installing `dracut` on Ubuntu

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1337,13 +1337,15 @@ def install_debian_or_ubuntu(args, workspace, run_build_script, mirror):
         f.write("#!/bin/sh\n")
         f.write("exit 101")
     os.chmod(policyrcd, 0o755)
-    # Work around "Failed to find module 'crc32c'" dracut issue
-    # See also:
-    # - https://github.com/antonio-petricca/buddy-linux/issues/2#issuecomment-404505527
-    # - https://bugs.launchpad.net/ubuntu/+source/dracut/+bug/1781143
+    dracut_bug_comment = [
+        '# Work around "Failed to find module \'crc32c\'" dracut issue\n',
+        '# See also:\n',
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -1712,6 +1712,12 @@ def install_boot_loader_debian(args, workspace):
     run_workspace_command(args, workspace,
                     "/usr/bin/kernel-install", "add", kernel_version, "/boot/vmlinuz-" + kernel_version)
 
+def install_boot_loader_ubuntu(args, workspace):
```
lucaswerkmeister: Is there a difference between this function and `install_boot_loader_debian`? They look identical to me (but I haven’t verified it).

mkosi
```diff
@@ -1338,6 +1338,13 @@ def install_debian_or_ubuntu(args, workspace, run_build_script, mirror):
         f.write("#!/bin/sh\n")
         f.write("exit 101")
     os.chmod(policyrcd, 0o755)
+    # Work around "Failed to find module 'crc32c'" dracut issue
+    # See also:
+    # - https://github.com/antonio-petricca/buddy-linux/issues/2#issuecomment-404505527
+    # - https://bugs.launchpad.net/ubuntu/+source/dracut/+bug/1781143
```
lucaswerkmeister: It might be nice to include these comments in the `dracut_bug_conf` file as well, but I’m not sure how important that is.

mkosi
keszybz: It is indentical. Dunno, maybe this one should just call the other one?

PR #270, opened 2018-10-22
Add static type annotations

zyga:
This patch adds a lot of type annotations, as compatible with mypy
0.620. The type checking is not perfect as many places use loosely typed
argparse.Namespace and I didn't want to introduce any changes there.

Some errors reported by mypy now show missing None checks but I also
decided not to fix them and make the review more complex. This patch
should only annotate types, nothing else.

Signed-off-by: Zygmunt Krynicki <zygmunt.krynicki@canonical.com>

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -25,6 +25,21 @@ import tempfile
 import urllib.request
 import uuid
 
+from typing import (
+    Any,
+    Dict,
+    Callable,
+    Generator,
+    IO,
+    Iterable,
+    Iterator,
+    List,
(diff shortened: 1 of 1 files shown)


poettering:
Looks superficially good to me. I am not used to these new python features. @keszybz I think you know this better, I'll leave it to you to merge it?

keszybz:
This has no effect on general usage. `mkosi` still imports fine under python3.5, 3.6, 3.7, and of course the annotations are ignored.

PR #271, opened 2018-10-24
Update Fedora defaults to 29

keszybz:


configs/mkosi.fedora
```diff
@@ -0,0 +1,24 @@
+# SPDX-License-Identifier: LGPL-2.1+
+# Let's build an image that is just good enough to build new mkosi images again
```
poettering: in systemd this dir is currently just called ".mkosi"...

I figure we should agree on a single name there, in particular as eventually I'd like to add "mkosi --all", that spiders through all such files and builds them all in one go.

poettering: i figure the dir name should have mkosi in the name. maybe "mkosi.recipes/"? or "mkosi.files/"?

keszybz: I like `mkosi.files` better. It's more generic.

keszybz: Updated.

PR #272, opened 2018-10-24
Fix possible None value in prepare_swap

zyga:
Mypy reports that partition() may return None and break the
requirements of subprocess.run():

    mkosi:525: error: List item 2 has incompatible type "Optional[str]";
    expected "Union[bytes, str, _PathLike[Any]]"

There are plenty of issues like so please treat this patch as an RFC. I
think the best way forward is to reduce the number of optional values to
the point where such checks are not needed. The generic nature of many
helpers (looking at the args and returning None if some condition
doesn't hold) amplifies this problem.

Signed-off-by: Zygmunt Krynicki <zygmunt.krynicki@canonical.com>

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -522,7 +522,9 @@ def prepare_swap(args: argparse.Namespace, loopdev: Optional[str], cached: bool)
         return
 
     with complete_step('Formatting swap partition'):
-        run(["mkswap", "-Lswap", partition(loopdev, args.swap_partno)], check=True)
+        part = partition(loopdev, args.swap_partno)
+        if part is not None:
+            run(["mkswap", "-Lswap", part], check=True)
 
 def prepare_esp(args: argparse.Namespace, loopdev: Optional[str], cached: bool) -> None:
     if loopdev is None:
--- a/mkosi
+++ b/mkosi
@@ -507,7 +507,7 @@ def attach_image_loopback(args: argparse.Namespace, raw: Optional[IO[str]]):
         with complete_step('Detaching image file'):
             run(["losetup", "--detach", loopdev], check=True)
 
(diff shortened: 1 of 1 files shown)


zyga:
On 2nd look I can simplify this further.

mkosi
```diff
@@ -507,7 +507,7 @@ def attach_image_loopback(args: argparse.Namespace, raw: Optional[IO[str]]):
         with complete_step('Detaching image file'):
             run(["losetup", "--detach", loopdev], check=True)
 
-def partition(loopdev: str, partno: Optional[int]) -> Optional[str]:
+def partition(loopdev: str, partno: int) -> str:
```
keszybz: What happens if partno is None?

zyga: At runtime, nothing new. Type annotation simply tells Mypy to report errors if `partno` is anything but an int. Since at present it is obtained from `args.swap_partno` (for example) and that has a type of `Any` it is not type checked further.

I have a follow-up patch that makes this strictly typed, that is `args.swap_partno` is no longer `Any` but `Optional[int]`. Then the places that check for `if args.swap_partno is None: return' allow Mypy to infer that the more narrow type applies and thus get the correct `int` type.

zyga: A follow up to this is in https://github.com/systemd/mkosi/pull/273 - it allows Mypy to prove that the type is correct in this fragment now.

keszybz: But right below, there is a check `if partno is None`, which is in disagreement with the declared type `partno: int`. I think the code inside the function should agree with the declared type.

zyga: That's correct, I didn't notice it, it should be removed along with the introduction of the type annotation.

keszybz:
Closing in favour of #279.

PR #275, opened 2018-10-25
Plain squashfs image output and making final images leaner

keszybz:
This provides the basis for generating squashfs images that could almost be used as snaps. The snap meta-data is still missing, and the squashfs generation options are not exactly as the snap store wants them, but I'll submit that in a later PR. I think the parts here are useful on their own and I'm tired of rebasing my patches.


mkosi
```diff
@@ -2406,11 +2419,18 @@ def setup_package_cache(args: CommandLineArguments) -> Optional[tempfile.Tempora
...
     def __call__(self, parser, namespace, values, option_string=None):
         l = getattr(namespace, self.dest)
         if l is None:
             l = []
-        l.extend(values.split(self.delimiter))
+        new = values.split(self.delimiter)
+        for x in new:
+            if self.list_choices is not None and x not in self.list_choices:
+                raise ValueError('Unknown value {!r}'.format(x))
+            l.append(x)
```
poettering: not that it matters, but it would be sexier to continue to use l.extend() if lis_choices is None... I mean, Python is not the fastest language in the world and needless loops sound needless to me ;-)

mkosi
```diff
@@ -97,6 +97,8 @@ class OutputFormat(enum.Enum):
     gpt_squashfs = enum.auto()
     gpt_xfs = enum.auto()
 
+    squashfs = enum.auto()
```
poettering: hmm, maybe call this "naked_squashfs"? to emphasize that it's really that: "naked"?

(other word instead of naked might be fine too)

poettering: also, shouldn't "is_disk()" return true for this?

(and maybe add is_gpt() or so, to cover the gpt ones?)

mkosi
```diff
@@ -3501,7 +3500,7 @@ def build_image(args: CommandLineArguments, workspace: tempfile.TemporaryDirecto
 
     tar = make_tar(args, workspace.name, run_build_script, for_cache)
 
-    return raw, tar, root_hash
+    return raw or squashfs, tar, root_hash
```
poettering: hmm, what does this resolve to? my python-fu is apparently not good enough to understand what a "string or string" resolves to?

mkosi
```diff
@@ -3633,10 +3632,7 @@ def build_stuff(args: CommandLineArguments) -> None:
     signature = calculate_signature(args, checksum)
     bmap = calculate_bmap(args, raw)
 
-    link_output(args,
-                workspace.name,
-                raw.name if raw is not None else None,
-                tar.name if tar is not None else None)
+    link_output(args, workspace, raw or tar)
```
poettering: hmm, how come this .name is dropped here?

mkosi
```diff
@@ -397,7 +397,6 @@ def disable_cow(path: str) -> None:
     run(["chattr", "+C", path], stdout=DEVNULL, stderr=DEVNULL, check=False)
 
 def determine_partition_table(args: CommandLineArguments):
```
poettering: "begging", eh? ;-)

mkosi
```diff
@@ -1054,8 +1054,27 @@ def invoke_dnf(args: CommandLineArguments, workspace: str, repositories: List[st
...
+        keep_rpm_data = keep_dnf_data or 'rpm' in packages
+
+        if not keep_dnf_data:
+            print_step('Cleaning dnf metadata...')
+            remove_glob(root + '/var/lib/dnf',
+                        root + '/var/log/dnf.*',
+                        root + '/var/log/hawkey.*',
+                        root + '/var/cache/dnf',
+                        root + '/etc/yum/repos.d',
+                        root + '/etc/dnf')
```
poettering: what's in these /etc directories? why do they exist at all if dnf is not installed?

I mean, I can totally see that dnf creates the dirs in /var while running even if it is not installed in the image itself, but what about /etc?

mkosi
```diff
@@ -1245,14 +1268,30 @@ def invoke_yum(args: CommandLineArguments, workspace: str, repositories: List[st
...
+        if not keep_yum_data:
+            print_step('Cleaning yum metadata...')
+            remove_glob(root + '/var/lib/yum',
+                        root + '/var/log/yum.*',
+                        root + '/var/cache/yum',
+                        root + '/etc/yum/repos.d',
+                        root + '/etc/yum')
+        if not keep_rpm_data:
+            print_step('Cleaning rpm metadata...')
+            remove_glob(root + '/var/lib/rpm')
```
poettering: hmm, the code a second time? this should be a function and shared between both callers

mkosi
```diff
@@ -1054,8 +1054,27 @@ def invoke_dnf(args: CommandLineArguments, workspace: str, repositories: List[st
     with mount_api_vfs(args, workspace):
         run(cmdline, check=True)
 
+    if cleanup:
```
poettering: hmm, this is quite generically named, no? maybe call this "package_manager_cleanup"?

mkosi
keszybz: The logical expressions resolve lazily to the argument that determines the outcome. I.e., A and B becomes A if A is false, and B otherwise. A or B becomes A if A is true, and B otherwise. We know that both `raw` and `squashfs` cannot be true at once, so this essentially means "pick the non-false one".

mkosi
keszybz: `.name` is used in `link_output`.

mkosi
keszybz: Hmm, I guess it's better to keep the files in /etc. They won't be used be anything, but they are very small, so it doesn't much either way. I'll trim the list to `/var` only.

mkosi
keszybz: It *was* supposed to be generic. My idea was to stuff additional cleanup steps there in the future. E.g. logs or something.

mkosi
poettering: urks, I can't say this is a particularly pretty python feature.

mkosi
keszybz: I changed the name to "plain_squshfs".

> also, shouldn't "is_disk()" return true for this?

No, the idea is that "is_disk()" returns true for whole-disk images, and a single fs image is more like "partition". I added some docstrings to clarify this.

mkosi
keszybz: Nah, it's in option parsing code, and a few ten or hundreds of microseconds don't matter here. This runs just once.

keszybz:
Updated. I think I addressed all the comments, except as noted above. I added some bugfixes for stuff that I found when working on the update, so the code is changed in various places. PTAL.

keszybz:
Rebased.

mkosi
poettering: maybe then do is_disk() → is_partitioned()?

PR #276, opened 2018-10-31
drop random seed in built images, and other fixlets

poettering:


poettering:
@lucasdemarchi by all means if you like something, go ahead and merge it!

poettering:
and thanks for the review of course!

lucasdemarchi:
ok. My usual approach when reviewing things from people with commit rights is to let them merge on their own pace rather than merging myself. Or give some time for anyone to chime in.

PR #278, opened 2018-11-01
optionally convert output to qcow2

poettering:


mkosi
```diff
@@ -2740,6 +2754,9 @@ def process_setting(args: CommandLineArguments, section: str, key: str, value: A
         elif key == "XZ":
             if args.xz is None:
                 args.xz = parse_boolean(value)
+        elif key == "QCow2":
```
keszybz: Hmmm, this fancy capitalization might be hard to remember. "qcow2" is short for "qemu copy-on-write format 2", so I think it should be all capitalized: "QCOW2".

PR #279, opened 2018-11-02
Fix typing in partition() function

keszybz:
Replacement for #272.


keszybz:
I'll just merge this because it is so simple.

lucasdemarchi:
> I'll just merge this because it is so simple.

this breaks mkosi when we don't have a home partition and we are generating a disk image:

```
    subprocess.CalledProcessError: Command '['mkfs.ext4', '-L', 'home', '-M', '/home', '/dev/loop0pNone']' returned non-zero exit status 1.
```

See #282

Issue #138, opened 2017-08-22
Cross-Architecture support in mkosi: permit creating arm images from x86-64 and suchlike


Pillar1989:
Hi
    I want  to use mkosi create armhf rootfs image. Is there any docments about this?

poettering:
Sorry, but cross-building is something mkosi can't do for you, as it is built around dnf/debootstrap and such, and they don't support that. Sorry!

poettering:
So, as it turns out I was wrong, dnf actually supports bootstrapping non-native archs, and I think it would make a ton of sense to support this in mkosi. Hence, let's reopen this, and let's make this an RFE issue.

Adding this should be fairly straightforward implementation-wise, given the hard work is already done in dnf/qemu.

poettering:
https://npmccallum.gitlab.io/post/cross-architecture-roots-with-dnf/

fsateler:
deboostrap also has support for cross-installing, but it's a bit more involved https://wiki.ubuntu.com/ARM/RootfsFromScratch/QemuDebootstrap

gyscos:
pacstrap also supports cross-architecture installations by specifying an alternative config file with `-C`.

DaanDeMeyer:
We have an --architecture switch now so I think this can be closed.

PR #281, opened 2018-11-04
Missing bits to allow snaps to be generated using mkosi

keszybz:
The patch for mksquashfs args is fairly ugly, I know. Suggestions for alternative approaches are very much welcome.

Instructions:
```console
$ git clone https://gitlab.com/keszybz/fedora29
$ cd fedora29
$ sudo SNAP_VERSION=$(date +%Y.%m.%d) SNAP_ARCH=amd64 mkosi -i -f --default mkosi.default.fedora -r 29 -o fedora29.$(date +%Y.%m.%d).amd64.snap
$ sudo SNAP_VERSION=$(date +%Y.%m.%d) SNAP_ARCH=i386 mkosi -i -f --default mkosi.default.fedora -r 29 --arch=i686 -o fedora29.$(date +%Y.%m.%d).i686.snap
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ they exist in the local directory:
   `systemd-nspawn`), with `$SRCDIR` pointing to the *source*
   tree. `$DESTDIR` points to a directory where the script should place
   any files generated it would like to end up in the *final*
-  image. Note that `make`/`automake` based build systems generally
+  image. Note that `make`/`automake`/`meson` based build systems generally
   honour `$DESTDIR`, thus making it very natural to build *source*
   trees from the build script. After the *development* image was built
   and the build script ran inside of it, it is removed again. After
@@ -205,16 +205,34 @@ they exist in the local directory:
   artifacts.
 
 * `mkosi.postinst` may be an executable script. If it exists it is
-  invoked as last step of preparing an image, from within the image
-  context. It is once called for the *development* image (if this is
-  enabled, see above) with the "build" command line parameter, right
-  before invoking the build script. It is called a second time for the
-  *final* image with the "final" command line parameter, right before
-  the image is considered complete. This script may be used to alter
-  the images without any restrictions, after all software packages and
-  built sources have been installed. Note that this script is executed
-  directly in the image context with the final root directory in
-  place, without any `$SRCDIR`/`$DESTDIR` setup.
+  invoked as the penultimate step of preparing an image, from within
+  the image context. It is once called for the *development* image (if
+  this is enabled, see above) with the "build" command line parameter,
+  right before invoking the build script. It is called a second time
+  for the *final* image with the "final" command line parameter, right
+  before the image is considered complete. This script may be used to
+  alter the images without any restrictions, after all software
+  packages and built sources have been installed. Note that this
+  script is executed directly in the image context with the final root
+  directory in place, without any `$SRCDIR`/`$DESTDIR` setup.
+
+* `mkosi.finalize` may be an executable script. If it exists it is
+  invoked as last step of preparing an image, from the host system.
+  It is once called for the *development* image (if this is enabled,
+  see above) with the "build" command line parameter, as the last step
+  before invoking the build script, after the `mkosi.postinst` script
+  is invoked.  It is called the second time with the "final" command
+  line parameter as the last step before the image is considered
+  complete. The environment variable `$BUILDROOT` points to the root
+  directory of the installation image. Additional verbs may be added
+  in the future, the script should be prepared for that. This script
+  may be used to alter the images without any restrictions, after all
+  software packages and built sources have been installed. This script
+  is more flexible than `mkosi.postinst` in two regards: it has access
+  to the host file system so it's easier to copy in additional files
+  or to modify the image based on external configuration, and the
+  script is run in the host, so it can be used even without emulation
+  even if the image has a foreign architecture.
 
 * `mkosi.mksquashfs-tool` may be an executable script. If it exists is
   is called instead of `mksquashfs`.
--- a/README.md
+++ b/README.md
@@ -216,6 +216,9 @@ they exist in the local directory:
   directly in the image context with the final root directory in
   place, without any `$SRCDIR`/`$DESTDIR` setup.
 
+* `mkosi.mksquashfs-tool` may be an executable script. If it exists is
+  is called instead of `mksquashfs`.
+
 * `mkosi.nspawn` may be an nspawn settings file. If this exists
   it will be copied into the same place as the output image
   file. This is useful since nspawn looks for settings files
--- a/mkosi
+++ b/mkosi
@@ -180,15 +180,17 @@ GPT_FOOTER_SIZE = 1024*1024
 
 GPTRootTypePair = collections.namedtuple('GPTRootTypePair', 'root verity')
 
-def gpt_root_native():
-    """The tag for the native GPT root partition
+def gpt_root_native(arch: str) -> Tuple[uuid.UUID, uuid.UUID]:
+    """The tag for the native GPT root partition for the given architecture
 
     Returns a tuple of two tags: for the root partition and for the
     matching verity partition.
     """
-    if platform.machine() == "x86_64":
+    if arch is None:
+        arch = platform.machine()
+    if arch == 'x86_64':
         return GPTRootTypePair(GPT_ROOT_X86_64, GPT_ROOT_X86_64_VERITY)
-    elif platform.machine() == "aarch64":
+    elif arch == 'aarch64':
         return GPTRootTypePair(GPT_ROOT_ARM_64, GPT_ROOT_ARM_64_VERITY)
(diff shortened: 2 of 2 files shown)


mkosi
```diff
@@ -2560,6 +2583,7 @@ def parse_args() -> CommandLineArguments:
     group.add_argument("--build-dir", help='Path to use as persistent build directory', metavar='PATH')
     group.add_argument("--build-package", action=CommaDelimitedListAction, dest='build_packages', default=[], help='Additional packages needed for build script', metavar='PACKAGE')
     group.add_argument("--postinst-script", help='Postinstall script to run inside image', metavar='PATH')
+    group.add_argument("--postinst2-script", help='Postinstall script to run outside image', metavar='PATH')
```
zyga: I don't know if there is any precedent for naming thing like that but perhaps `--postinst-outer-script` would be more direct, the `2` is a bit magic (granted the help message is clear and explicit).

mkosi
```diff
@@ -2017,10 +2017,13 @@ def make_squashfs(args, workspace, for_cache):
...
 
-    comp_args = [] if args.compress is True else ['-comp', args.compress]
+    comp_args = args.mksquashfs_extra_args if args.mksquashfs_extra_args else ['-noappend']
+    if args.compress is not True:
+        assert args.compress is not False
+        comp_args += ['-comp', args.compress]
     with complete_step('Creating squashfs file system'):
         f = tempfile.NamedTemporaryFile(dir=os.path.dirname(args.output), prefix=".mkosi-squashfs")
-        run(["mksquashfs", os.path.join(workspace, "root"), f.name, "-noappend", *comp_args],
+        run(["mksquashfs", os.path.join(workspace, "root"), f.name, *comp_args],
```
poettering: hmm, I am kinda afraid that this might end up in an inflation of similar options sooner or later for all the other tools, and hence I'd prefer if we'd find a generic solution for this so that we can cover other tools with a similar approach. 

here's one idea: maybe, let's change run() to redirect the executed command line to `./mkosi.tool-<command>` if that file exists and is executable. i.e. if you want to replace or tweak the "mksquashfs" command you'd drop a shell script in called `./mkosi.tool-mksquashfs", and it would just exec the real mksquashfs but add in some args. Such a scheme could eventually be useful for allow other, less established distros to be hooked in, i.e. if "./mkosi.tool-pkg-install" or so exists we'd use that tool instead of apt-get/dnf/...

mkosi
```diff
@@ -2529,6 +2539,7 @@ def parse_args() -> CommandLineArguments:
     group.add_argument('-r', "--release", help='Distribution release to install')
     group.add_argument('-m', "--mirror", help='Distribution mirror to use')
     group.add_argument("--repositories", action=CommaDelimitedListAction, dest='repositories', help='Repositories to use', metavar='REPOS')
+    group.add_argument('--arch', help='Override the architecture of installation')
```
poettering: let's call this "architecture" please, this is exotic enough to not benefit from abbreviation

mkosi
```diff
@@ -3745,6 +3759,10 @@ def check_root():
     if os.getuid() != 0:
         die("Must be invoked as root.")
 
+def check_native(args: CommandLineArguments) -> None:
+    if args.arch is not None and args.arch != platform.machine() and args.build_script:
+        die('Cannot (currently) override the architecture and run build commands')
```
poettering: We could eventually make this useful for https://github.com/systemd/mkosi/issues/138#issuecomment-338143991

README.md
```diff
@@ -205,16 +205,29 @@ they exist in the local directory:
...
+  for the *final* image with the "final" command line parameter, right
+  before the image is considered complete. This script may be used to
+  alter the images without any restrictions, after all software
+  packages and built sources have been installed. Note that this
+  script is executed directly in the image context with the final root
+  directory in place, without any `$SRCDIR`/`$DESTDIR` setup.
+
+* `mkosi.postinst2` may be an executable script. If it exists it is
+  invoked as last step of preparing an image, from the host system.
+  It is called as `mkosi.postinst2 final $ROOT`, where `$ROOT` is the
```
poettering: why is this not called for the "build" operation?

mkosi
keszybz: Right now, when I call dnf for archs others than amd64  and i386, installation fails. But they might work if binfmt is set up properly. I need to figure this out. If that is true, then #138 would just work.

poettering:
btw, i think another option would be to define "snap" as output type, and it would result in the right mksquashfs options to be picked. I think snap is well established enough to deserve its own first-level output mode

keszybz:
Updated as suggested:
- `mkosi.postinst2` is renamed to `mkosi.finalize`
- it is also called with "build"
- `mkosi.mksquashfs-tool` will be autodetected, but can also be overridden using configuration, similarly to other options
- `--arch` is renamed to `--architecture`

keszybz:
> another option would be to define "snap" as output type

I considered that, but I don't think we're quite there yet. This only work "base" snaps, for now, because any kind of layering is not implemented. Also, the output is not really verified, etc. I'd prefer to "start small", without declaring this any kind of official snap output, and maybe add "snap" as a format later, if it turns out that people actually want to use `mkosi` to produce snaps.

mkosi
```diff
@@ -1799,9 +1801,18 @@ def run_postinst_script(args, workspace, run_build_script, for_cache):
...
 
-        run_workspace_command(args, workspace, "/root/postinst", "build" if run_build_script else "final", network=args.with_network)
+        run_workspace_command(args, workspace, "/root/postinst", verb, network=args.with_network)
         os.unlink(os.path.join(workspace, "root", "root/postinst"))
 
+def run_finalize_script(args: CommandLineArguments, workspace: str, *, verb: str):
+    if args.finalize_script is None:
+        return
+
+    with complete_step('Running external postinstall script'):
```
poettering: should be updated: "external postinstall" → "finalize". let's not confuse users with different terminology for the same stuff

mkosi
```diff
@@ -3135,14 +3107,16 @@ def load_args() -> CommandLineArguments:
...
 
     load_defaults(args)
-    find_nspawn_settings(args)
+
+    args_find_path(args, 'nspawn_settings', "mkosi.nspawn")
+    args_find_path(args, 'build_script',    "mkosi.build")
+    args_find_path(args, 'build_sources',   ".")
+    args_find_path(args, 'build_dir',       "mkosi.builddir/")
+    args_find_path(args, 'postinst_script', "mkosi.postinst");
+    args_find_path(args, 'output_dir',      "mkosi.output/");
```
poettering: not that it matters, but is there are a reason for using `''` quotes on the second parameter, but `""` quotes on the third?

mkosi
```diff
@@ -3005,6 +3005,12 @@ def find_skeleton(args: CommandLineArguments) -> None:
     if os.path.isfile("mkosi.skeleton.tar"):
         args.skeleton_trees.append("mkosi.skeleton.tar")
 
+def args_find_path(args: CommandLineArguments, name: str, path: str) -> None:
+    if getattr(args, name) is not None:
+        return None
```
poettering: why does this return `None` here? it shoud not return anything at all, no?

mkosi
keszybz: It's written in the commit message. For visual distinction between variable names and paths.

mkosi
keszybz: Left-over from previous version of the code. I'll remove it.

keszybz:
We call `mount`, `btrfs`, `dnf`, `yum`, `chattr`, `sfdisk`, `sync`, `losetup`, `mkswap`, `mkfs.fat/ext4/btrfs/xfs`, `cryptsetup`, `umount`, `systemctl`, `systemd-nspawn`, `swupd-extract`, `debootstrap`, `pacman`, `pacman-key, `pacstrap`, `gpg-connect-agent`, `fuser`, `zypper`, `git`, `tar`, `mksquashfs`, `sfdisk`, `dd`, `veritysetup`, `sbsign`, `xz/pxz`, `qemu-img`, `qemu*`, `gpg`, `bmaptool`. I guess we could allow wrappers for all those tools, but I don't see the use case. We call `mksquashfs` just once, and in a fairly simple way, so it's easy for the user to add a few options. For many tools we either make a few different calls with completely different options, or even if we call them just once, we construct an elaborate command lines. Overriding those cases is going to be pretty hard, because the override script can either just pass through all the options, in which case it case just append some options that don't conflict, or it must "understand" all options and filter them or such. So I'd prefer to add support for overrides for specific tools as the use cases come up.

keszybz:
Updated with the two simple fixes. I left out `mkosi.*-tool` support out for now.

mkosi
Conan-Kudo: @keszybz This is documented here as a way that works (at least for me): https://wiki.mageia.org/en/Using_DNF#Setting_up_a_container_for_a_non-native_architecture

PR #282, opened 2018-11-05
Fixes for latest merges

lucasdemarchi:
Fixes I had to do for rebasing my PRs

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -3718,7 +3718,7 @@ def build_stuff(args: CommandLineArguments) -> None:
     signature = calculate_signature(args, checksum)
     bmap = calculate_bmap(args, raw)
 
-    link_output(args, workspace, raw or tar)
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -558,6 +558,9 @@ def attach_image_loopback(args: CommandLineArguments, raw: Optional[IO[str]]):
             run(["losetup", "--detach", loopdev], check=True)
 
 def partition(loopdev: str, partno: int) -> str:
```
keszybz: So please fix the type annotation too. It should be `Optional[int]` in this case.

keszybz: The annotations are not complete, and there's hundreds of warnings reported by mypy, so it's pretty hard to notice new bugs. The longer term goal is to clean this all up, but we're not there yet.

PR #283, opened 2018-11-06
Use f-strings and some minor cleanups

keszybz:


mkosi
```diff
@@ -193,7 +193,7 @@ def gpt_root_native(arch: str) -> Tuple[uuid.UUID, uuid.UUID]:
     elif arch == 'aarch64':
         return GPTRootTypePair(GPT_ROOT_ARM_64, GPT_ROOT_ARM_64_VERITY)
     else:
```
poettering: hmm, the python version check at the top of the script should be updated to 3.6 now, right?

poettering:
lgtm but misses the updated python version check

keszybz:
UPdated.

fsateler:
Setup.py is missing (which is probably where distros will look).

lucaswerkmeister:
> Setup.py is missing (which is probably where distros will look).

#285

PR #284, opened 2018-11-07
chown output files to the calling user

lucasdemarchi:
Use $SUDO_GID and $SUDO_UID to chown the final output files. This way the user doesn't end up with root-owned files that he has to chmod/chown to continue working with.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -2431,6 +2431,16 @@ def _link_output(args: CommandLineArguments, oldpath: str, newpath: str):
     os.chmod(oldpath, 0o666 & ~args.original_umask)
     os.link(oldpath, newpath)
 
+    sudo_uid = os.getenv("SUDO_UID")
+    sudo_gid = os.getenv("SUDO_GID")
+    if not (sudo_uid and sudo_gid):
+        return
+
+    sudo_user = os.getenv("SUDO_USER", default=sudo_uid)
+    with complete_step(f"Changing ownership of output file {newpath} to user {sudo_user} (acquired from sudo)",
+                       f"Successfully changed ownership of {newpath}"):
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -2422,6 +2422,10 @@ def save_cache(args: CommandLineArguments, workspace: str, raw: str, cache_path:
 def _link_output(args: CommandLineArguments, oldpath: str, newpath: str):
     os.chmod(oldpath, 0o666 & ~args.original_umask)
     os.link(oldpath, newpath)
+    sudo_uid = os.getenv("SUDO_UID")
+    sudo_gid = os.getenv("SUDO_GID")
+    if sudo_uid and sudo_gid:
+        os.chown(newpath, int(sudo_uid), int(sudo_gid))
```
poettering: this really should be enclosed in `with complete_step(…)`, so that this appears visibly in the output. It's otherwise too surprising that we do this.

Also, I'd feel more comfortable if there was a way to disable such magic...

poettering: the complete step things can be recursive. All i want is that there's some explanation of what#s happening in the log output. At least to me it's kinda surprising to honour the sudo UID, i have never seen such behaviour in any other tool. While I think it's useful it also should not be magic behind the scenes, but clearly communicated. hence I'd really like to see in the log about something brief along the lines of "Changing ownership of output file foobar to user quux (acquired from sudo)…" or so.

poettering:
(btw, also, please rebase and repush. there's now a CI system running, but it will only work if PR is rebased on current master)

lucasdemarchi:
@poettering are you ok with the "--no-chown" argument?

PR #286, opened 2018-11-11
Add support for building mkosi with semaphore CI

poettering:


poettering:
Yay, it finally works!

PR #289, opened 2018-11-14
Bind mount postinst script rather than copying

lucasdemarchi:
@keszybz  what did you mean in the comment I just removed? Just bind-mounting the postinst script to / works for me, so I don't understand the issue.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1796,17 +1796,8 @@ def run_postinst_script(args, workspace, run_build_script, for_cache):
     verb = "build" if run_build_script else "final"
 
     with complete_step('Running postinstall script'):
-
-        # We copy the postinst script into the build tree. We'd prefer
-        # mounting it into the tree, but for that we'd need a good
-        # place to mount it to. But if we create that we might as well
-        # just copy the file anyway.
-
-        shutil.copy2(args.postinst_script,
-                     os.path.join(workspace, "root", "root/postinst"))
-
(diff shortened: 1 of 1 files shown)


poettering:
But this leaves the /postinst file we mount over around in the container after it completed, no?

lucasdemarchi:
it uses systemd-nspawn to mount it. systemd-nspawn doesn't leave them mounted after return. This is equivalent to how we enable `--with-network`.

poettering:
> it uses systemd-nspawn to mount it. systemd-nspawn doesn't leave them mounted after return. This is equivalent to how we enable `--with-network`.

It doesn't leave them mounted, but it does create the file/dir to mount it over, and it doesn#t clean that up...

poettering:
The old code did an unlink() after it completed operation. The new code doesn't do that, it binds mounts it in using the fact that nspawn will create a mount point to mount over implicitly if it's missing, but then never cleans that mount point up after use...

The comment in there is supposed to explain that: creating the file to mount over first, and then mounting it over is not too much better than copying the file in the first place, since either way a file is create that needs to be unlinked afterwards...

lucasdemarchi:
humn... true, that. I was expecting systemd-nspawn to actually remove the file if it created it in order to bind-mount.... I missed that nspawn-mount.c doesn't undo the file/dir creation

PR #290, opened 2018-11-17
Add back again the ovmf firmware path from Arch Linux

jimenezrick:
I tested running `mkosi qemu` under Arch Linux and it works perfectly fine with the shipped ovmf firmware. I think we should include back again it in the search path. I don't see anything wrong with it.

I also fixed a few typos in the README.


jimenezrick:
Sounds good!

PR #291, opened 2018-11-20
set correct defaults for optional args

foxxx0:
args.with_network and args.with_docs were always initialized to `False`
thus ignoring any setting in a config file (WithNetwork/WithDocs).

This breaks the config parser (load_defaults) because it checks if the
current value of `key` is None before assigning the value from the
config file.

Signed-off-by: Thore Bödecker <me@foxxx0.de>

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -2594,7 +2594,7 @@ def parse_args() -> CommandLineArguments:
 
     group = parser.add_argument_group("Packages")
     group.add_argument('-p', "--package", action=CommaDelimitedListAction, dest='packages', default=[], help='Add an additional package to the OS image', metavar='PACKAGE')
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -85,6 +85,8 @@ class CommandLineArguments(argparse.Namespace):
 
     swap_partno = None  # type: Optional[int]
     esp_partno = None  # type: Optional[int]
+    with_network = None  # type: Optional[bool]
+    with_docs = None  # type: Optional[bool]
```
lucasdemarchi: @keszybz is this the preferred way to fix? Or should we just pass `default=None` to the option?

poettering: I think default=None is preferable.

foxxx0:
Thanks for the quick feedback, I've changed the PR according to your recommendations.

PR #292, opened 2018-11-21
Mount build source

lucasdemarchi:

PR #295, opened 2018-11-26
Refactor config options parsing

lucasdemarchi:
This is on top of #292 because I think that one is much closer to be merged. First 3 commits can be ignored.

This reworks the way we add commandline and config options to be more automatic and avoid mistakes we often do while adding an option: wrong section, override vs extend, only override if not set in command line.

In order to simplify, there's a new ConfigParser wrapper to glue together the argparse and configparser. Any argument added to argparse in the form "--this-is-an-option" adds the equivalent CamelCase option to the config parser. There are some exceptions, so we can force the name of the config option, too.

Flags are supported to handle the cases in which we are extending the values from each config file or command line.  Some of our options were not consistent in the way this is handled. All of them are handled the same way now. Example: "Packages", "ExtraTrees", "BuildPackages" were extending the list for multiple files. All the others like "SkeletonTrees" were extending only the command line argument.

There was a lack of consistency in the use of plural with regard to command line and config option. We had `--build-package`, but `BuildPackages`, just to name one. Now they are the same entry, with the only difference being the use of CamelCase for config files.

We are also not parsing the command line arguments multiple times anymore, since validation can be done without it.

There are some intended bugs: `ESPSize` is now `EspSize`, `XZ` is now `Xz`, etc. I think once this is accepted we have to go one by one and decide what to do.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -3008,7 +3008,7 @@ def parse_compression(value: str) -> Union[str, bool]:
     return parse_boolean(value)
 
 
-def parse_args() -> CommandLineArguments:
+def parse_args() -> (CommandLineArguments, ArgumentConfigParser):
     parser = ArgumentConfigParser(description='Build Legacy-Free OS Images', add_help=False)
 
     # command-line arguments only
@@ -3148,7 +3148,7 @@ def parse_args() -> CommandLineArguments:
 
     args_find_path(args, 'all_directory', "mkosi.files/")
 
-    return args
+    return args, parser
 
 
 def parse_bytes(bytes: Optional[str]) -> Optional[int]:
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -2654,80 +2699,86 @@ def parse_compression(value: str) -> Union[str, bool]:
...
-    group.add_argument("--checksum", action='store_true', help='Write SHA256SUMS file')
-    group.add_argument("--sign", action='store_true', help='Write and sign SHA256SUMS file')
-    group.add_argument("--key", help='GPG key to use for signing')
-    group.add_argument("--bmap", action='store_true', help='Write block map file (.bmap) for bmaptool usage (only gpt_ext4, gpt_btrfs)')
-    group.add_argument("--password", help='Set the root password')
-
-    group = parser.add_argument_group("Host configuration")
-    group.add_argument("--extra-search-paths", action=ColonDelimitedListAction, default=[], help="List of colon-separated paths to look for programs before looking in PATH")
+    group = config.add_section(parser.add_argument_group("Distribution"))
+    config.add_argument(group.add_argument('-d', "--distribution", choices=Distribution.__members__, help='Distribution to install'))
```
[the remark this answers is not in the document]
lucasdemarchi: What I could do is to move  group.add_argument() inside config.add_argument() and fwd kargs/kwargs, popping the keywords I'm using myself.... not sure if it's clearer, but I could give this a try.

mkosi
```diff
@@ -3051,7 +3051,7 @@ def parse_args() -> CommandLineArguments:
     group.add_argument("--password", help='Set the root password')
 
     group = parser.add_argument_group("Host configuration")
-    group.add_argument("--extra-search-paths", action=ColonDelimitedListAction, default=[],
+    group.add_argument("--extra-search-path", dest='extra_search_paths', action=ColonDelimitedListAction, default=[],
```
poettering: the internal field should probably also drop the "s"

do we care about compat here btw? if this has been added more recently than the last release I figure we don't need to care

poettering:
patch looks OK to me, but I'll leave this to @keszybz for a thorough review, as my python-fu is too limited to talk about python stylistic issues in a smart way, and this stuff looks like it's in the area of stylistic issues. @keszybz ?

mkosi
lucasdemarchi: @cmarcelo do you think we need the compat option here?

@poettering I don't think the way you did for `--kernel-commandline` works... that creates a `dest.kernel_commandline` that we don't use anywhere. Gotta check if we can have multiple options setting the same dest.

lucasdemarchi:
Ping @keszybz

mkosi
```diff
@@ -2909,6 +2909,64 @@ def setup_package_cache(args: CommandLineArguments) -> Optional[tempfile.Tempora
...
+    argparser = None
+    curr_group_title = None
+    curr_group = None
+    FLAGS = enum.Enum('FLAGS', 'Extend')
+
+    def __init__(self, *kargs, **kwargs):
+        self.argparser = argparse.ArgumentParser(*kargs, **kwargs)
+
+    def add_section(self, *kargs, **kwargs):
+        fp_name = kwargs.pop('file_parser_name', None)
```
keszybz: Please use the normal kwarg-only syntax:
```python3
def add_section(self, *args, file_parser_name=None, **kwargs):
    ...
```
The syntax with `pop` is just a work-around for python2 which we don't support anyway.

Also, I think the second args should traditionally be `args` not `kargs`.

Or is this interface somehow required by `argparse`?

mkosi
```diff
@@ -2909,6 +2909,64 @@ def setup_package_cache(args: CommandLineArguments) -> Optional[tempfile.Tempora
     return d
 
 
+class ArgumentConfigParser():
+    options = dict()
+    argparser = None
+    curr_group_title = None
+    curr_group = None
+    FLAGS = enum.Enum('FLAGS', 'Extend')
```
keszybz: I don't understand why those class variables are need. At least `argparser` is unconditionally hidden by the assignment in `__init__`.

mkosi
```diff
@@ -2909,6 +2909,64 @@ def setup_package_cache(args: CommandLineArguments) -> Optional[tempfile.Tempora
     return d
 
 
+class ArgumentConfigParser():
```
```python3
class ArgumentConfigParser:
```

mkosi
```diff
@@ -2909,6 +2909,64 @@ def setup_package_cache(args: CommandLineArguments) -> Optional[tempfile.Tempora
...
+        self.curr_group = self.argparser.add_argument_group(*kargs, **kwargs)
+        self.curr_group_title = fp_name or self.curr_group.title
+        self.options[self.curr_group_title] = []
+
+    def add_argument(self, *kargs, **kwargs):
+        # default to add an option to config file that has the same name as the argument option,
+        # except that it is converted to CamelCase as per _arg_to_option().
+        # Setting file_parser_name=None allows to add the option only to the command line
+        fp_name = kwargs.pop('file_parser_name', "")
+        fp_flags = kwargs.pop('file_parser_flags', None)
```
keszybz: Here too, those should be kwarg-only arguments.

afreof:
Hi Lucas,
The command line parser of this nice tool defenitively needs refactoring and bug fixing.

Parameters provided by mkosi.defaults or mkosi.defaults.d/* often cannot be overwritten by command line arguments or highr priority files as mentiond in the documentation. The root cause for these bugs is that default files are parsed after command line arguments are parsed. If parameters would be parsed in the order of their prioriy the code would be:
a=1
a=2
a=3
And a would be assigned to 3

The current implementation does something like:
a=1
if a is None:
  a=2
if a is None:
  a=3

Another major problem is that python's argument parser does a type conversion where the config file parser simply provides strings for certain parameters, which leads to exceptions.

After testing this patch set, I ended up trying to fix this in a different way. argparse provides a framework to implement a default file parser. Basically the approach is to convert the parameters from files to their command line argument counter part. As a last step the extended list of parameters (from files and from command line) are fed to the ArgumentParser. Based on that I'm working on an alternative PR.

[Here is my WIP](https://github.com/afreof/mkosi/commits/master-af) (Please ignore the 14 patches providing trivial fixes):
The patches on this branch solve the problems mentioned above. It's already possible to get the expected values in the summary from different priority config files and command line arguments.  It's also easy to support parameters like QCow2 which is probabely not supported by your implementation right now.

mkosi
lucasdemarchi: Just initialize the defaults. I could turn them into object members and initialize them in `__init__()`, but here it has the same effect.

lucasdemarchi:
> If parameters would be parsed in the order of their prioriy the code would be:
> a=1
> a=2
> a=3
> And a would be assigned to 3

Yeah, I agree and initially I also thought about changing the order of the parser. However, there are some problems: we have cases in which we extend the value from previous parser, cases in which we unconditionally replace and cases in which we replace only if it was not set.

> Any feedback is very welcome.

I didn't think about extending the ArgumentParser by inheriting from it though: that was a nice idea. But it's also more complex. The version here is tested on multiple scenarios and working for me with different distros. I think we should go with this versions and potentially replace with yours once it's not in a wip-state anymore and it proves to be better.

lucasdemarchi:
> The version here is tested on multiple scenarios and working for me with different distros

I take this back. I tested more and discovered problems like parsing values vs default arg. We also need to streamline the bool parsing on config file vs args...  I will take a while to get back to this and fix the issues. Feel free to submit a PR if you get to a solution that works first.

lucasdemarchi:
@afreof are you going to create a PR or should I take a look again on fixing this branch?

afreof:
Hi Lucas,
Yes, the PR is almost ready. Unfortunately solving all the details became a bit more invasive as I initially expected. To minimize the risk I added 55 unit tests just for the new parse_args function.
However, I think there are still some manual tests needed. Since I do not have any advanced use cases ready for testing mkosi, I would like to ask you for giving my improve-args-parsing branch from https://github.com/afreof/mkosi a try with your current setup? Especially the --all mode is not tested yet.

Of course, I could also split the PR into smaller junks or delete/squash the first set of trivial cleanup patches.

afreof:
Please have a look at PR [#330](https://github.com/systemd/mkosi/pull/330).

PR #300, opened 2018-12-02
Modify so that `flake8` and `mypy` are happy

LukeShu:
This is all small changes (naturally), but since there are so many of them it can be hard to review them all at once, so this is a whole bunch of small commits, for easy reviewing.

Come back with no complaints from the following (sometimes using `# NOQA` or `# type: ignore`):

 - `flake8` 3.6.0 with `flake8-isort` 2.5: This turns out to be entirely stylistic changes
 - `mypy --strict` 0.641: Mostly mucking with and adding type hints, but the last 6 commits do make actual changes.

The individual commit messages should be sufficiently detailed.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1543,8 +1543,8 @@ def install_debian_or_ubuntu(args: CommandLineArguments,
         cmdline += ["--merged-usr"]
 
     cmdline += [args.release,
-               workspace + "/root",
-               mirror]
+                workspace + "/root",
+                mirror]
 
     if args.bootable and args.output_format == OutputFormat.gpt_btrfs:
         cmdline[4] += ",btrfs-tools"
@@ -2039,9 +2039,9 @@ def install_grub(args, workspace, loopdev, grub):
     ]
 
     run_workspace_command(
-            args, workspace, f"{grub}-install",
-            "--modules=ext2 part_gpt", "--target=i386-pc",
-            loopdev, nspawn_params=nspawn_params)
+        args, workspace, f"{grub}-install",
+        "--modules=ext2 part_gpt", "--target=i386-pc",
+        loopdev, nspawn_params=nspawn_params)
 
     run_workspace_command(
             args, workspace, f"{grub}-mkconfig",
@@ -2073,7 +2073,7 @@ def install_boot_loader_debian(args, workspace):
     kernel_version = next(filter(lambda x: x[0].isdigit(), os.listdir(os.path.join(workspace, "root", "lib/modules"))))
 
     run_workspace_command(args, workspace,
-                    "/usr/bin/kernel-install", "add", kernel_version, "/boot/vmlinuz-" + kernel_version)
+                          "/usr/bin/kernel-install", "add", kernel_version, "/boot/vmlinuz-" + kernel_version)
 
 
 def install_boot_loader_ubuntu(args, workspace):
@@ -2711,7 +2711,7 @@ def _link_output(args: CommandLineArguments, oldpath: str, newpath: str):
     sudo_user = os.getenv("SUDO_USER", default=sudo_uid)
     with complete_step(f"Changing ownership of output file {newpath} to user {sudo_user} (acquired from sudo)",
                        f"Successfully changed ownership of {newpath}"):
-       os.chown(newpath, int(sudo_uid), int(sudo_gid))
+        os.chown(newpath, int(sudo_uid), int(sudo_gid))
 
 
 def link_output(args: CommandLineArguments, workspace: str, artifact: str) -> None:
@@ -4204,8 +4204,8 @@ def run_qemu(args: CommandLineArguments) -> None:
                     ['qemu', '-machine', 'accel=kvm'],
                     ['qemu-kvm']):
 
-           if cmdline[0] and shutil.which(cmdline[0]):
-               break
+        if cmdline[0] and shutil.which(cmdline[0]):
+            break
     else:
         die("Couldn't find QEMU/KVM binary")
 
--- a/mkosi
+++ b/mkosi
@@ -45,12 +45,6 @@ from typing import (
     cast,
 )
 
-try:
-    import argcomplete
-except ImportError:
-    pass
-
-
 __version__ = '4'
 
 if sys.version_info < (3, 6):
@@ -2966,8 +2960,9 @@ def parse_args() -> CommandLineArguments:
                        help='When running with sudo, disable reassignment of ownership of the generated files to the original user')  # NOQA: E501
 
     try:
+        import argcomplete  # type: ignore
         argcomplete.autocomplete(parser)
-    except NameError:
+    except ImportError:
         pass
 
     args = parser.parse_args(namespace=CommandLineArguments())
--- a/mkosi
+++ b/mkosi
@@ -123,7 +123,7 @@ class OutputFormat(enum.Enum):
             return cls[name]
         except KeyError:
             # this let's argparse generate a proper error message
-            return name
+            return name  # type: ignore
 
     def is_disk_rw(self) -> bool:
         "Output format is a disk image with a parition table and a writable filesystem"
@@ -368,7 +368,7 @@ def copy(oldpath: str, newpath: str) -> None:
             symlink_f(target, newentry)
             shutil.copystat(entry.path, newentry, follow_symlinks=False)
         else:
-            st = entry.stat(follow_symlinks=False)
+            st = entry.stat(follow_symlinks=False)  # type: ignore  # mypy 0.641 doesn't know about follow_symlinks
             if stat.S_ISREG(st.st_mode):
                 copy_file(entry.path, newentry)
             else:
@@ -2868,6 +2868,7 @@ def print_output_size(args: CommandLineArguments) -> None:
 
 
 def setup_package_cache(args: CommandLineArguments) -> Optional[tempfile.TemporaryDirectory]:
+    d: Optional[tempfile.TemporaryDirectory] = None
     with complete_step('Setting up package cache',
                        'Setting up package cache {} complete') as output:
         if args.cache_path is None:
@@ -2875,7 +2876,6 @@ def setup_package_cache(args: CommandLineArguments) -> Optional[tempfile.Tempora
             args.cache_path = d.name
         else:
             os.makedirs(args.cache_path, 0o755, exist_ok=True)
-            d = None
         output.append(args.cache_path)
 
     return d
@@ -3041,7 +3041,7 @@ def parse_args() -> CommandLineArguments:
     except ImportError:
         pass
 
-    args = parser.parse_args(namespace=CommandLineArguments())
+    args = cast(CommandLineArguments, parser.parse_args(namespace=CommandLineArguments()))
 
     if args.verb == "help":
         parser.print_help()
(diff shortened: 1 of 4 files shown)


mkosi
```diff
@@ -1898,7 +1900,12 @@ def install_opensuse(args, workspace, *, run_build_script):
             cmdlinefile.write(args.kernel_commandline + " root=/dev/gpt-auto-root\n")
 
 
-def install_distribution(args, workspace, *, cached, **kwargs):
+def install_distribution(args: CommandLineArguments,
+                         workspace: str,
+                         *,
+                         run_build_script: bool,
```
lucaswerkmeister: Is this parameter new? I don’t see it used below.

mkosi
```diff
@@ -380,7 +381,12 @@ def complete_step(text: str, text2: Optional[str] = None) -> Iterator[List[Any]]
     print_step(text2.format(*args) + '.')
 
 
-@complete_step('Detaching namespace')
+# https://github.com/python/mypy/issues/1317
+C = TypeVar('C', bound=Callable)
+completestep = cast(Callable[[str], Callable[[C], C]], complete_step)
```
lucaswerkmeister: Alternative suggestion: rename the `def` to `completestep`, and make `complete_step` the cast version of it, so all the uses of the decorator wouldn’t need to be adjusted. I feel like that would be better, but I’m not sure, so for now that’s just a suggestion :)

mkosi
```diff
@@ -2719,12 +2736,13 @@ def calculate_signature(args: CommandLineArguments, checksum: Optional[IO[Any]])
...
 
 
-def calculate_bmap(args: CommandLineArguments, raw: BinaryIO) -> Optional[TextIO]:
+def calculate_bmap(args: CommandLineArguments, raw: Optional[BinaryIO]) -> Optional[TextIO]:
     if not args.bmap:
         return None
 
     if not args.output_format.is_disk_rw():
         return None
+    assert raw is not None
```
lucaswerkmeister: I wonder if there should be a blank line before this addition. Usually, the `assert … is not None` is right after the single `if …: return` check, but here we have two such checks separated by a blank line, and the `assert` doesn’t really belong any closer to the second one than the first one, in my opinion.

mkosi
```diff
@@ -1626,12 +1626,17 @@ SigLevel    = Required DatabaseOptional
...
     if args.bootable:
-        # Patch mkinitcpio configuration so: 1) we remove autodetect and
+        # Patch mkinitcpio configuration so:
+        # 1) we remove autodetect and
         # 2) we add the modules needed for encrypt.
-        patch_file(os.path.join(workspace, "root", "etc/mkinitcpio.conf"),
-                   lambda line: "HOOKS=\"systemd modconf block sd-encrypt filesystems keyboard fsck\"\n" if line.startswith("HOOKS=") and args.encrypt == "all" else
-                                "HOOKS=\"systemd modconf block filesystems fsck\"\n"                     if line.startswith("HOOKS=") else
-                                line)
+        def jj(line: str) -> str:
```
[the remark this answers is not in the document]
LukeShu: I used `jj` because the lambdas in `set_root_password()` were already named `jj`, so I just kept that name for those, and used it in the other places too, for consistency.  If I'd made up the name "from scratch", I'd probably have named it `patchfn`

It looks like `jj` was originally introduced by @keszybz, you'd have to ask him.

mkosi
```diff
@@ -2810,15 +2810,15 @@ class ListAction(argparse.Action):
...
+        if ary is None:
+            ary = []
         new = values.split(self.delimiter)
         for x in new:
             if self.list_choices is not None and x not in self.list_choices:
                 raise ValueError(f'Unknown value {x!r}')
-            l.append(x)
-        setattr(namespace, self.dest, l)
+            ary.append(x)
+        setattr(namespace, self.dest, ary)
```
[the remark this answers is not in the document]
LukeShu: Huh. I'm not sure why `flake8` doesn't catch that.

mkosi
LukeShu: It's also used as a generator inside of `with` statements.  I'm not sure in which case it's used more, but I figured the decorator usage was easier to search/replace for.

mkosi
LukeShu: I know that `is_disk_rw()` implies `is_disk()` implies `raw is not None`.

I'm not sure what the relationship between `bmap` and `raw` is; in my understanding, it is related the the second one, and not to the first.

LukeShu:
> but that’s just a suggestion. (In the last case, the “blank line between method definitions” commit could be squashed into it as well.)

class & function definitions get 2 blank lines between them; methods only get 1

mkosi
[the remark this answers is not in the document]
LukeShu: Ah, E741 does't trigger for *argument* names.

mkosi
LukeShu: Whoops, it was previously part of `**kwargs`.  A later commit fixes this when it drops kwargs.

LukeShu:
v2:
 - Reword the E302,E305 commit message
 - Reword the E301 commit message
 - Remove a duplicate bullet point in the E251 commit message
 - Also modify line_join_list() in the E741 commit, even though E741
   doesn't care
 - Don't add the 'run_build_script' kwarg to install_distibution() in
   the "add missing type-hints" commit; only hint existing args in
   that commit.  Instead, do that in a later commit:
 - In the install_distribution()/install_{DISTRIBUTION}() type
   unification commit, switch install_distribution() from using
   `**kwargs` to listing each the kwargs explicitly.

That is almost entirely metadata/history change, not content change.
The only content change is as follows:

```patch
$ git diff to-upstream/2018-12-02-flake8+mypy-v1 to-upstream/2018-12-03-flake8+mypy-v2 
diff --git a/mkosi b/mkosi
index abeda33..0026560 100755
--- a/mkosi
+++ b/mkosi
@@ -3870,12 +3870,12 @@ def none_to_none(s: Optional[str]) -> str:
     return "none" if s is None else s
 
 
-def line_join_list(l: List[str]) -> str:
+def line_join_list(ary: List[str]) -> str:
 
-    if not l:
+    if not ary:
         return "none"
 
-    return "\n                        ".join(l)
+    return "\n                        ".join(ary)
 
 
 def print_summary(args: CommandLineArguments) -> None:
```

poettering:
Hmm, this will regress quickly again if not verified all the time. Any chance you can include a patch that updates /ci/semaphore.sh to validate this on each PR?

LukeShu:
Why does `semaphore.sh` manually add an apt repo for python3.6, and install python from there?  There's an option in Semaphore to just start with 3.6.

poettering:
> Why does `semaphore.sh` manually add an apt repo for python3.6, and install python from there? There's an option in Semaphore to just start with 3.6.

iirc then the code runs in a docker container, but we need more than that for losetup and stuff to work

zyga:
Wow, this is super nice. What is the baseline python version for mkosi? The attribute type syntax is quite new and will not parse on older python versions.

LukeShu:
@poettering That's definitely not the case on Semaphore 2.0, but apparently the systemd org is on Semaphore Classic.

@zyga  The attribute type syntax is introduced in Python 3.6, which a recent commit (e965e58a3fac2b79f54cafe1c1deba6197570c27) has already bumped the required version to (the appropriate commit message notes this).

mkosi
```diff
@@ -61,7 +61,7 @@ if sys.version_info < (3, 6):
 arg_debug = ()
 
 
-def run(cmdline: List[str], execvp: bool = False, **kwargs) -> Optional[subprocess.CompletedProcess]:
+def run(cmdline: List[str], execvp: bool = False, **kwargs) -> subprocess.CompletedProcess:
```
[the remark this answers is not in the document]
LukeShu: I didn't realize that there's a special no return type.  Good suggestion!

poettering:
needs a rebase.

regarding the contents of the patch... I am not really a guru when it comes to python, and this looks like something only a true pythonista can appreciate, hence maybe @keszybz could you have a look if this makes sense to you?

poettering:
@poettering That's definitely not the case on Semaphore 2.0, but apparently the systemd org is on Semaphore Classic, so IDK.

That sounds like you are volunteering to move the CI to semaphore 2.0? ;-)

mkosi
```diff
@@ -401,9 +441,11 @@ def btrfs_subvol_delete(path: str) -> None:
     # Delete the subvolume now that all its descendants have been deleted
     run(["btrfs", "subvol", "delete", path], stdout=DEVNULL, stderr=DEVNULL, check=True)
 
-def btrfs_subvol_make_ro(path: str, b:bool=True) -> None:
+
+def btrfs_subvol_make_ro(path: str, b: bool = True) -> None:
```
zyga: I may be wrong but last time I looked at this the assignment of default values was supposed to not use whitespace around it.

mkosi
```diff
@@ -492,14 +536,15 @@ def determine_partition_table(args: CommandLineArguments):
...
-def create_image(args: CommandLineArguments, workspace: str, for_cache: bool) -> Optional[IO[str]]:
+def create_image(args: CommandLineArguments, workspace: str, for_cache: bool) -> Optional[BinaryIO]:
     if not args.output_format.is_disk():
         return None
 
     with complete_step('Creating partition table',
                        'Created partition table as {.name}') as output:
 
-        f = tempfile.NamedTemporaryFile(dir=os.path.dirname(args.output), prefix='.mkosi-', delete=not for_cache)
+        f: BinaryIO = cast(BinaryIO, tempfile.NamedTemporaryFile(prefix='.mkosi-', delete=not for_cache,
```
zyga: Doesn't the cast tell the inference engine enough about `f` that you can drop the type annotation entirely?

mkosi
```diff
@@ -541,7 +590,8 @@ def reuse_cache_image(args: CommandLineArguments, workspace: str, run_build_scri
             return None, False
 
         with source:
-            f = tempfile.NamedTemporaryFile(dir = os.path.dirname(args.output), prefix='.mkosi-')
+            f: BinaryIO = cast(BinaryIO, tempfile.NamedTemporaryFile(prefix='.mkosi-',
```
zyga: Same question about the inference engine as above.

mkosi
lucaswerkmeister: Whitespace is supposed to be there iff the parameter has a type specified, see 554a455934c9214ffc4aa5011aec399a2967a424.

LukeShu:
> That sounds like you are volunteering to move the CI to semaphore 2.0? ;-)

@poettering Sure, if I got the appropriate credentials :)

mkosi
LukeShu: Yes.

My general mindset is "every use of `cast()` is a workaround for a bug/deficiency in mypy, and should be able to be dropped as mypy improves."  So, eventually, when mypy improves, I think the line should be:

    f: BinaryIO = tempfile.NamedTemporaryFile(prefix='.mkosi-', delete=not for_cache,

So, I included the `: BinaryIO` in it, since I think it should be there when the `cast()` is (hopefully) eventually dropped; even though it isn't necessary as long as the `cast()` is present.

LukeShu: Yes.

My mindset is "every call to `cast()` is a workaround for a bug/deficiency in mypy, and should be droppable as mypy improves."

I think the "correct" way to write the line is

    f: BinaryIO = tempfile.NamedTemporaryFile(prefix='.mkosi-', delete=not for_cache,

Since I'd like the `: BinaryIO` to be there after the `cast()` is removed, I included it now, even though it isn't necessary as long as `cast()` is there.

PR #306, opened 2018-12-14
Two trivial stylistic fixups

keszybz:


mkosi
```diff
@@ -1807,9 +1807,9 @@ SigLevel    = Required DatabaseOptional
         def jj(line: str) -> str:
             if line.startswith("HOOKS="):
                 if args.encrypt == "all":
-                    return "HOOKS=\"systemd modconf block sd-encrypt filesystems keyboard fsck\"\n"
+                    return 'HOOKS="systemd modconf block sd-encrypt filesystems keyboard fsck"\n'
                 else:
-                    return "HOOKS=\"systemd modconf block filesystems fsck\"\n"
+                    return 'HOOKS=\"systemd modconf block filesystems fsck"\n'
```
filbranden: You missed removing one backslash here.

keszybz: Thanks, fixed.

keszybz:
Well, this patch is consistent: it removes all uses of this style of quoting ;) But more seriously, I think that the insistence on C-style quoting only makes sense if one is accustomed to C. In Python it is natural and actually recommended to mix `'`, `'''`, `"`, `"""`, and `r` as is best for each specific case. (And personally, I think `'` is better than `"` because it is more light-weight.)

PR #307, opened 2018-12-14
Do not allow dd to create output file

keszybz:


mkosi
```diff
@@ -946,6 +946,10 @@ def mount_image(args: CommandLineArguments,
         yield
     finally:
         with complete_step('Unmounting image'):
+
+            for d in ("home", "srv", "efi", "run", "tmp"):
```
lucasdemarchi: not a fan of keeping this list growing. Particularly as there are plans to support arbitrary partitions taken from `mkosi.sfdisk`.

mkosi
```diff
@@ -1014,7 +1018,7 @@ def mount_cache(args: CommandLineArguments, workspace: str) -> Generator[None, N
 
 def umount(where: str) -> None:
     # Ignore failures and error messages
-    run(["umount", "--recursive", "-n", where], stdout=DEVNULL, stderr=DEVNULL)
+    run(["umount", "-n", where], stdout=DEVNULL, stderr=DEVNULL)
```
lucasdemarchi: What about instead of doing this turning this into a mountpoint in the squashfs case and avoid the problem here?
`mount --bind root root` should work and then the `umount --recursive`, too.

mkosi
```diff
@@ -2451,14 +2452,18 @@ def insert_partition(args: CommandLineArguments,
...
-        dev = None
+        device = None
 
+    path = pathlib.Path(device if device is not None else partition(loopdev, partno))
     try:
-        run(["dd", "if=" + blob.name, "of=" + (dev if dev is not None else partition(loopdev, partno))], check=True)
+        # make sure we are writing to an actual device file
+        assert path.is_block_device()
+
+        run(["dd", f'if={blob.name}', f'of={path}'], check=True)
```
poettering: why not just use `dd conv=nocreat`?

mkosi
poettering: yeah, i think this is a nice approach and makes our codepaths a bit more alike

keszybz:
Since #328 was merged, no need to keep the first patch. I changed the second patch to use conv=nocreat as suggested.

PR #308, opened 2018-12-14
add a man page

poettering:


TODO.md
```diff
@@ -0,0 +1,24 @@
...
+
+* allow passing env vars
+
+* mkosi --all (for building everything in mkosi.files/)
+
+* --architecture= is chaos: we need to define a clear vocabulary of
+  architectures that can deal with the different names of
+  architectures in debian, fedora and uname.
+
+* --boot-protocols=/BootProtocols= should be stricter and refuse
```
lucasdemarchi: isn't it doing that already? Last time I tried it was.

mkosi
```diff
@@ -690,6 +690,16 @@ def mkfs_btrfs(label: str, dev: str) -> None:
     run(["mkfs.btrfs", "-L", label, "-d", "single", "-m", "single", dev], check=True)
 
 
+def mkfs_generic(args: CommandLineArguments, label: str, mount: str, dev: str) -> None:
+
```
lucasdemarchi: annoying extra blank line that just got removed in a recent commit from other functions. So we may want to avoid adding them on new commits.

lucasdemarchi:
Some trivial comments above, otherwise LGTM. I'll leave it up to you to merge it as is and fix on top or to fix it in this PR.

Some of the fixes conflict with #295 (they were actually done there as well). It would be good to know if that's good to go after rebasing on this one so we don't keep getting new conflicts.

TODO.md
poettering: ah, true, i was confused that it was not implemented as Enum (though probably should be)

poettering:
Force pushed a new version: rebase and with all of @lucasdemarchi's issues addressed, hopfeully. PTAL!

fsateler:
What command is needed to build mkosi.1 from mkosi.md ? It'd be nice if there was a script invoking the tool with the right parameters (if any)

poettering:
Use pandoc.

yeah, there should be a hookup for that in setup.py, but I couldn't figure out how to do that, and docs for this appear pretty much non-existant? Or maybe I am just bad at google... happy to merge a patch for that...

fsateler:
I used `pandoc -f markdown -t man mkosi.md -o mkosi.1`, but the result is not very nice:

image

PR #309, opened 2018-12-15
add new "mkosi -a" switch for building all images in mkosi.files/

poettering:
This comes without documentation for now, I'll add it later as soon as #308 is fixed.


poettering:
Force pushed a new version with the commit msg fixed, ptal!

Issue #252, opened 2018-05-30
systemd-tmpfiles-clean.service removes files while an image is created


michaelolbrich:
mkosi failed for me because part of the OS tree was deleted while creating the image:
The files created in /var/tmp/mkosi-* are older than 30 days. systemd-tmpfiles-clean.service started while I created the image and deleted most of the OS tree.

It seems to me, that /var/tmp is a bad choice for the temporary directory.

poettering:
Fix waiting in #319

PR #319, opened 2019-01-18
mkosi: take a BSD lock on the workspace directory while running

poettering:
This makes use of the new logic here:

https://github.com/systemd/systemd/pull/11482

(And on older systems has no effect, but that shouldn't hurt either)

Fixes: #252

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -4242,6 +4242,11 @@ def build_stuff(args: CommandLineArguments) -> None:
     setup_package_cache(args)
     workspace = setup_workspace(args)
 
+    # Make sure tmpfiles' aging doesn't interfere with our workspace
+    # while we are working on it.
+    dir_fd = os.open(workspace.name, os.O_RDONLY|os.O_DIRECTORY|os.O_CLOEXEC)
+    fcntl.flock(dir_fd, fcntl.LOCK_EX)
+
     # If caching is requested, then make sure we have cache images around we can make use of
     if need_cache_images(args):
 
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -4307,6 +4312,7 @@ def build_stuff(args: CommandLineArguments) -> None:
     if root_hash is not None:
         print_step(f'Root hash is {root_hash}.')
 
+    del dir_fd
```
lucaswerkmeister: Shouldn’t we `os.close(fd)` instead? As far as I can tell, `os.open` returns a plain `int`, so deleting it won’t have any special effect.

mkosi
```diff
@@ -4243,6 +4243,11 @@ def build_stuff(args: CommandLineArguments) -> None:
     setup_package_cache(args)
     workspace = setup_workspace(args)
 
+    # Make sure tmpfiles' aging doesn't interfere with our workspace
+    # while we are working on it.
+    dir_fd = os.open(workspace.name, os.O_RDONLY|os.O_DIRECTORY|os.O_CLOEXEC)
+    fcntl.flock(dir_fd, fcntl.LOCK_EX)
```
lucaswerkmeister: Should this include `fcntl.LOCK_NB`? If a lock is already held on the directory for some reason, it’s probably better for us to proceed (and risk that the other lock is released before we’re done, making us vulnerable to the tmpfiles cleanup again) than block until the other lock is released (we have no idea when that would be).

poettering: well, the idea is that if tmpfiles is currently iterating through our directory, then we should wait until it is finished with that before we start using the directory. hence it should be a blocking open

mkosi
poettering: indeed

poettering:
force pushed a new version. only change: use "os.close()" instead of "del" for the fd.

PTAL!

PR #326, opened 2019-02-14
support for "xbootldr" partitions

poettering:
This implements the ability to optoinally create "xbootldr" partitions, as defined in the boot loader spec.

This goes nicely with https://github.com/systemd/systemd/pull/11701


lucasdemarchi:
LGTM but I guess it's better to wait the support be merged in systemd

poettering:
the systemd side has been merged now. Since @lucasdemarchi liked this PR otherwise, let's merge!

PR #328, opened 2019-02-19
Turn root into a mountpoint

lucasdemarchi:
By doing this we can safely rely on mount --recursive doing the right
thing.

This is the alternative to the revert commit in #307

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -917,16 +917,15 @@ def mount_image(args: CommandLineArguments,
                 home_dev: Optional[str],
                 srv_dev: Optional[str],
                 root_read_only: bool = False) -> Generator[None, None, None]:
-    if loopdev is None:
-        yield None
-        return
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -917,16 +917,16 @@ def mount_image(args: CommandLineArguments,
...
-    if loopdev is None:
-        yield None
-        return
-    assert root_dev is not None
-
     with complete_step('Mounting image'):
         root = os.path.join(workspace, "root")
 
-        if args.output_format != OutputFormat.gpt_squashfs:
+        print(root_dev)
```
poettering: leftover from debugging?

poettering:
looks good to me, except for the debug log msg that appears left by accident

mkosi
lucasdemarchi: fixed, thanks

PR #329, opened 2019-02-24
Minor improvements

afreof:
Improve some details that pylint complained about.


keszybz:
OK. Generally, I think it's acceptable to have a local variable that overrides some obscure global, but it seems to come up again and again, so I guess it's better to rename those variables.

PR #330, opened 2019-02-24
Improve args parsing

afreof:
Improve argument parsing and mkosi.default files loading as already dicussed in [#295](https://github.com/systemd/mkosi/pull/295).

Bugfix: This PR delegates mkosi.default file loading to python's ArgumentParser. This allows to override settings from unlimited number of mkosi.default.d files as well as from command line arguments. This approach also fixes some issues with type conversion for settings loaded from mkosi.default files.

Add pytest based tests for command line arguments and mkosi.default files parsing. pytest is called by ci/semaphore.sh.

Feature: It's possible to remove items from a comma separated list argument (e.g. a package).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,13 @@
 *.cache-pre-dev
 *.cache-pre-inst
+.cache
 .mypy_cache/
+.project
+.pydevproject
+.pytest_cache/
 /.mkosi-*
 /SHA256SUMS
 /SHA256SUMS.gpg
-/__pycache__
 /build
 /dist
 /image
@@ -18,4 +21,4 @@
 /mkosi.extra
 /mkosi.nspawn
 /mkosi.rootpw
-.mypy_cache/
+__pycache__
--- a/mkosi
+++ b/mkosi
@@ -3054,7 +3054,7 @@ def create_parser() -> argparse.ArgumentParser:
     group.add_argument('-d', "--distribution", choices=Distribution.__members__, help='Distribution to install')
     group.add_argument('-r', "--release", help='Distribution release to install')
     group.add_argument('-m', "--mirror", help='Distribution mirror to use')
-    group.add_argument("--repositories", action=CommaDelimitedListAction, dest='repositories',
+    group.add_argument("--repositories", action=CommaDelimitedListAction, dest='repositories', default=[],
                        help='Repositories to use', metavar='REPOS')
     group.add_argument('--architecture', help='Override the architecture of installation')
 
@@ -3101,6 +3101,7 @@ def create_parser() -> argparse.ArgumentParser:
                        help='Install documentation')
     group.add_argument('-T', "--without-tests", action=BooleanAction, dest='with_tests', default=True,
                        help='Do not run tests as part of build script, if supported')
+    group.add_argument("--with-tests", action=BooleanAction, default=True, help=argparse.SUPPRESS)  # Compatibility option
     group.add_argument("--cache", dest='cache_path', help='Package cache path', metavar='PATH')
     group.add_argument("--extra-tree", action='append', dest='extra_trees', default=[],
                        help='Copy an extra tree on top of image', metavar='PATH')
@@ -3108,7 +3109,8 @@ def create_parser() -> argparse.ArgumentParser:
                        help='Use a skeleton tree to bootstrap the image before installing anything', metavar='PATH')
     group.add_argument("--build-script", help='Build script to run inside image', metavar='PATH')
     group.add_argument("--build-sources", help='Path for sources to build', metavar='PATH')
-    group.add_argument("--build-dir", help='Path to use as persistent build directory', metavar='PATH')
+    group.add_argument("--build-dir", help=argparse.SUPPRESS, metavar='PATH')  # Compatibility option
+    group.add_argument("--build-directory", dest='build_dir', help='Path to use as persistent build directory', metavar='PATH')
     group.add_argument("--build-package", action=CommaDelimitedListAction, dest='build_packages', default=[],
                        help='Additional packages needed for build script', metavar='PACKAGE')
     group.add_argument("--postinst-script", help='Postinstall script to run inside image', metavar='PATH')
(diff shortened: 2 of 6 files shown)


afreof:
No idea why, but the order of the commits got messed up in this view. [Here, the order is correct.](https://github.com/afreof/mkosi/commits/improve-args-parsing)

bustervill:
Various true/false options that are set in mkosi.default are not respected by the master branch, but work correctly with the code in this PR.

I found 2 visible differences in behaviour:

-   When running with `-i` I get:

    ```
    mkosi: error: argument -i/--incremental: Invalid literal for bool(): 'summary'
    ```

    i.e. now I have to run it with `-i yes`

-   The `KernelCommandLine=` option in mkosi.default no longer works with multi-line entry, like the `Packages=` option does.

afreof:
Thank you very much for the feedback.
Since -i optionally consumes a parameter the parser tries to consume the next parameter as bool. Since "summary" does not start with "-" it is treated valid which turns out to -i=summary. The fix is almost ready.
I guess the solution to the second problem is almost ready, too. I've found a way to reuse Python's configparser module instead of my handwritten ini file parser.

afreof:
Both problems are fixed on the force pushed update. The new version uses python's ConfigParser again.

One note about the --kernel-command-line argument. Kernel command line arguments are now appended. If you like to overide the default kernel parameters add a !* as the first parameter to get rid of the default arguments. For details, the man page are updated as well.

poettering:
hmm, this needs a rebase.

poettering:
Ill leave review of this to @keszybz or @lucasdemarchi, as this is beyond my python-fu...

.gitignore
```diff
@@ -18,4 +18,8 @@
 /mkosi.extra
 /mkosi.nspawn
 /mkosi.rootpw
-.mypy_cache/
+.pytest_cache/
+.cache
+.project
+.pydevproject
+tags
```
lucasdemarchi: this file should remain sorted. `LANG=C sort -u` will do it.

mkosi
```diff
@@ -3032,6 +3032,101 @@ class BooleanAction(argparse.Action):
...
+        'XZ': '--xz',
+        'NSpawnSettings': '--settings',
+        'ESPSize': '--esp-size',
+        'CheckSum': '--checksum',
+        'BMap': '--bmap',
+        'Packages': '--package',
+        'ExtraTrees': '--extra-tree',
+        'SkeletonTrees': '--skeleton-tree',
+        'BuildPackages': '--build-package',
+        'PostInstallationScript': '--postinst-script',
```
lucasdemarchi: I don't like much the different names for command line options and for configuration files. It's much easier when they match. Plural vs non plural and the case sensitive parser IMO makes it more difficult to use. Can we use this moment to fix that (possibly adding compatibility options)?

mkosi
```diff
@@ -3032,6 +3032,101 @@ class BooleanAction(argparse.Action):
...
+                return __class__.SPECIAL_MKOSI_DEFAULT_PARAMS[key]
+            except KeyError:
+                return '--' + camel_to_arg(key)
+
+        # expand arguments referencing files
+        new_arg_strings = []
+        for arg_string in arg_strings:
+            # for regular arguments, just add them back into the list
+            if not arg_string or arg_string[0] not in self.fromfile_prefix_chars:
+                new_arg_strings.append(arg_string)
```
lucasdemarchi: a `continue` here would be good to reduce 1 indent

mkosi
```diff
@@ -3032,6 +3032,101 @@ class BooleanAction(argparse.Action):
...
+            # replace arguments referencing files with the file content
+            else:
+                try:
+                    with open(arg_string[1:]) as args_file:
+                        config = configparser.ConfigParser(delimiters='=')
+                        config.optionxform = str  # type: ignore
+                        config.read_file(args_file)
```
lucasdemarchi: the `with` context should be only arround this `read_file`. No need to keep the file open afterwards.

mkosi
```diff
@@ -3032,6 +3032,101 @@ class BooleanAction(argparse.Action):
...
+                                cli_arg = ini_key_to_cli_arg(key)
+
+                                # \n in value strings is forwarded. Depending on the action type, \n is considered as a delimiter or needs to be replaced by a ' '
+                                for action in self._actions:
+                                    if cli_arg in action.option_strings:
+                                        if isinstance(action, ListAction):
+                                            value = value.replace(os.linesep, action.delimiter)
+                                new_arg_strings.extend([cli_arg, value])
+                except OSError:
+                    err = sys.exc_info()[1]
```
lucasdemarchi: what is this ? why not catch the exception `as e`?

mkosi
```diff
@@ -3167,34 +3262,127 @@ def create_parser() -> argparse.ArgumentParser:
     return parser
 
 
-def parse_args() -> CommandLineArguments:
+class MkosiParseError(Exception):
```
lucasdemarchi: MkosiParseException?

lucasdemarchi:
Overall I'm happy with this. I added some comments above. I think we can put this in soon, leave it there some time to catch regressions and then cut a release.

mkosi
afreof: The Python code already does what you suggest.
When ExtraTree is written to the configuration file, mkosi tries to find the ExtraTree key in this dictionary. Since this fails, ExtraTree is converted to --extra-tree.
When ExtraTree**s** is written to the configuration file, mkosi uses the transformation from the dictionary. The purpose of this dictionary is backward compatibility.

I will add another commit that removes some "s" from the man page.

afreof:
Thank you Lucas for the review. I force pushed an update that implements all your suggestions. Changing parameter names to singular or plural form is left for another commit. See my response above.

lucasdemarchi:
LGTM. @poettering your use case is a little different than mine. Could you try this PR before merging to make sure we aren't breaking anything?

lucasdemarchi:
@afreof thanks a lot! In the end we decided it was better to create a release and then merge this, to allow time for any fix up needed before the next release.

PR #331, opened 2019-02-28
Debian: Use btrfs-progs instead of btrfs-tools

llucax:
The package was [renamed](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=780081) some time ago (around March 2016). The package btrfs-progs is [available from Stretch](https://packages.debian.org/stretch/btrfs-progs) but it's also [in Jessie backports](https://packages.debian.org/jessie-backports/btrfs-progs); in Ubuntu it was introduced [in Bionic](https://packages.ubuntu.com/bionic/btrfs-progs), so it's probably more appropriate to use the new name.

Ideally a way to select which package to use should be provided, but I'm not sure it is worth the effort.

### DIFF
--- a/ci/semaphore.sh
+++ b/ci/semaphore.sh
@@ -6,6 +6,17 @@ sudo add-apt-repository --yes ppa:jonathonf/python-3.6
 sudo apt --yes update
 sudo apt --yes install python3.6 debootstrap systemd-container squashfs-tools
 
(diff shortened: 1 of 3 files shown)


llucax:
Added a test for btrfs. It would be better to test all the images, but I guess you need to enable docker support in semaphoreci so a base docker image matching the distro to generate can be used.

keszybz:
LGTM.

PR #337, opened 2019-03-24
Fix setting of boolean parameters in mkosi.default

L3n41c:
Some Boolean parameters are currently not taken into account when specified in `mkosi.default`.
The reason is that, when parsing `mkosi.default`, in `process_setting`, `args` parameters are set only if their previous value is `None` so that parameters passed on the command line have the priority over the content of `mkosi.default`.
But, when a parameter defined in `parse_args`, has `action='store_true'`, its default value is `False`.

So, let’s put `None` default value in `parse_args` so that parameters not specified on the command line can be overridden in `process_setting`.


afreof:
May I ask you to review and test #330 before pushing this one further? It solves the issue as well.

lucasdemarchi:
Should be fixed by #330 - let me know if it didn't.

PR #338, opened 2019-05-17
Fix BootProtocols parsing

lucasdemarchi:
Mark it with the same action as other lists, otherwise its parsing from
a config file is broken.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -3208,7 +3208,7 @@ def create_parser() -> ArgumentParserMkosi:
                        help='Do not run tests as part of build script, if supported')
     group.add_argument("--with-tests", action=BooleanAction, default=True, help=argparse.SUPPRESS)  # Compatibility option
     group.add_argument("--cache", dest='cache_path', help='Package cache path', metavar='PATH')
-    group.add_argument("--extra-tree", action='append', dest='extra_trees', default=[],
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
@afreof  can you take a look?

afreof:
Hi Lucas

The problem about the CI runner seems to be related to the old ppa based Ubuntu setup. I suggest to update to Ubuntu 18.04 which is now available. I added two commits to your branch to test this quickly.
[afreof tip-fix-bootprotocols](https://github.com/afreof/mkosi/commits/tip-fix-bootprotocols). Feel free to pick them to your PR.

Since I do not have access to the semaphore configuration I cannot test my changes against mkosi's current semaphore configuration. I tested successfully with a default project created by the wizard which is probably different to the current semaphore configuration of mkosi. Please note that I never did something with semaphore before.

lucasdemarchi:
@afreof in order to update to Ubuntu 18.04 and use the yml file we would need to update to semaphoreci 2.0.


I checked yesterday and I don't see a way to use the free tier inside a organization. I do use it on my personal account, but it seems it's not free for organizations. So either we fix it in another way or we have to choose another ci.

lucasdemarchi:
Let's discuss the CI issue on #341

PR #339, opened 2019-05-17
Photon OS support

suezzelur:
Known limitations:
- The gpg key is not published on the web, so VMWARE-RPM-GPG-KEY should be present in the host 
- No uefi boot supported yet

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -169,6 +169,7 @@ class Distribution(enum.Enum):
     mageia = 6
     centos = 7
     clear = 8
+    photon = 9
 
 
 GPT_ROOT_X86           = uuid.UUID("44479540f29741b29af7d131d5f0458a")  # NOQA: E221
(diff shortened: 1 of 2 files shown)


johannbg:
Can anyone create images and access the entire Photon OS repository's ( packages, source code etc ) without having to go through some registration process or paywall like subscription of some sort?

johannbg:
is tdnf available as an package manager in the primary distributions so photon os can be installed on mkosi on arch,debian,fedora and opensuse?

poettering:
hmm, is photon OS a distro that is going to stay? it uses "tdnf", what is that? is that pcakaged anywhere else?

I am not against adding this, just wondering if the distro is big enough and has a future so that we should care?

poettering:
patch looks good otherwise, but needs rebase

if the gpg keys are not downloadable, maybe just commit them here?

poettering:
I guess we can close this now that #357 is merged.

PR #340, opened 2019-05-22
Remove keys for EOL Fedora releases

johannbg:


mkosi
```diff
@@ -200,12 +200,6 @@ BIOS_PARTITION_SIZE = 1024 * 1024
 CLONE_NEWNS = 0x00020000
 
 FEDORA_KEYS_MAP = {
-    '23': '34EC9CBA',
-    '24': '81B46521',
-    '25': 'FDB19C98',
-    '26': '64DAB85D',
-    '27': 'F5282EE4',
-    '28': '9DB62FB1',
```
keszybz: Hmmm, I would prefer to keep them, because one might want to install an old release. Not for production, but for testing. Maybe we could just mark such keys as deprecated, and emit a warning when using them?

johannbg: I fail to see the pros it serves testing an EOL distribution since what's being tested will never make it to the distribution nor will ( or should ) it be deployed so it would be interesting to read what use cases you have in mind in that regard as well as your view if mkosi upstream is the correct venue to maintain that list ( think for example if mirrors are still carrying packages for those EOL distribution, required to generate the OS image etc. ).

Cons of having this is is that it will add to already existing scalability problem, for Fedora alone you are effectively arguing that mkosi should be shipping the current and past ( 31+IOT ) keys and that's just Fedora but same argument applies to keep supporting all the other distributions mkosi is supporting ( which based on current non existing policy regarding the topic can lead to all systemd based distribution + their EOL ones ) and what you are proposing is a solution to trying sustain that model which leads to a code bloat on a massive scale right.

lucasdemarchi: Keys are maintained here because no one came up with a better way to negotiate the keys.

It's up to Fedora to decide how long they maintain the packages on the server. If Fedora maintains the packages there, I don't see much gain in removing the keys.

johannbg: Well the fact is @lucasdemarchi all Fedora releases have been removed from mirrors except the current ones and the EOL'ed 28 which is scheduled to be removed ( cant remember the exact time frame for that stuff since I'm not maintaining Fedora mirror anymore ) so the gain is we dont include stuff that no longer works. 

Now that everybody seem to be unaware of that is just highlighting the fact we need to start testing if the "supported/included" releases can actually be generated as well as nobody seems to be using those EOL releases which renders @keszybz test argument kinda mute. 

Now once we start doing tests those distribution image genereration tests can quickly grow up to be in the hundreds ( or be like 8 or something ) depending on the vendor "include" policy which we need to hash out and decide for the project.

lucasdemarchi: If Fedora packages have been removed and there's no way to generate that distro anymore, then this is the main argument, which I agree with.

keszybz: > Well the fact is @lucasdemarchi all Fedora releases have been removed from mirrors except the current ones

This doesn't seem to be true at all. I just did `mkosi -d fedora -r 24` for testing, and it works without issue, so the packages are still available in the same location. In fact I know all Fedora releases are available since the very beginning (some packages from the early ones have been misplaced, but that was by mistake, only some subset). Also, GPL requires sources to be available for a long time, and it's very likely that distributions will want to keep the binary packages around for reference too.

I'll close this for now. I users want to install old releases, we should allow them to install them reliably.

johannbg:
@keszybz, @poettering, @lucasdemarchi  now that the policy has been decided to keep supporting older downstream distribution releases and reliably as well what's the minimum requirement for those distributions ( which will save me some time before I submit pull request for those archived releases )?
 
Should I submit pull request beginning from FC1 and onwards or is there an minimal requirement ( like systemd version foo, python version foo, yum/dnf/pkg etc version foo ) which mkosi can work with? 

With regards to test cases for this does the current CI infrastructure have enough resources to support those (  We need one test case per release and release version ( maybe two if we intend to do nspawn test as well ) for Fedora alone that's atleast 7 cases just with what we have now, more when I submit pull request for EOL release that still work )

PR #346, opened 2019-06-21
add full basic repo and epel for better centos container building

sequencer:


sequencer:
CentOS might be impossible to reuse without `epel` repo, should we add `epel` repo by default?

johannbg:
Repo is a third party repo to centOS so if that's implemented. The flood gates open for those ( rpm fusion for Fedora. Non oss for suse etc ). 

The centOS community itself recommends against enabling centOS plus probably due to potential conflicts

"The CentOSPlus repository contains a group of packages, all of which are updates. You probably do not want to enable CentOSPlus as a whole, but instead want to pick the exact packages that you want to use."

sequencer:
I'm don't use `CentOSPlus` too.
But I just found it exist in the default `yum.conf.d/CentOS-Base.repo`, if not include it, the image will still have `CentOSPlus` with default built, removing which will cause the repos in `CentOS-Base.repo` has different mirror.
The only **unofficial** repo is possibly the `epel`, but many user certainly need `python36` and some important packages.

johannbg:
Including repository's that are provided by the distributions themselves makes sense. 

Including repositories that are not included by the distributions themselves does not make sense.

In those cases people that need newer python or whatever, they will just switch to different distributions that do provide those component ( rhel8/fedora/arch etc ) and or should be working with the distribution to provide the missing ones or update the current ones.

sequencer:
You might have some misunderstanding about what CentOS is, the relationship between CentOS and Fedora, and the importance of `epel` for CentOS.
1. CentOS is the community version of RHEL, commercial tools will mostly support RHEL by default to avoid the up-to-date version's bug, the codebase is maintained by RedHat company. When RHEL is updating, RedHat have to make sure all commercial codes and softwares can work correctly, and take **responsibility** to that.
2. Fedora is the proving ground for RHEL, and most of RedHat employees are using it, they use this distribution to contribute to upstream and community, and of cause it's commonly accepted by community.  
3. Because RHEL codebase is responsible for commercial use, they are generally a small mount of stale codes with security updates(but stable for huge commercial software). `epel`, which is call as **Extra Packages for Enterprise Linux**, is maintained by Fedora project and truly necessary for developers, but **won't** take any responsibility for commercial use in RHEL, so developers will need `epel` for their developing.

> Including repositories that are not included by the distributions themselves does not make sense.

You can consider that `epel` is a semi-official repository for CentOS, and it's truely trusted, maintained by a same group of people, and **really really necessary** for developers.

> In those cases people that need newer python or whatever, they will just switch to different distributions that do provide those component ( rhel8/fedora/arch etc ) and or should be working with the distribution to provide the missing ones or update the current ones. 

If you are in some specific domains, and doesn't have OPS from RedHat, you will understand what CentOS is for, and always cannot get rid of which. `mkosi` is also a clean container builder, which can provide different distributions, and make them work as a nspawn container under a modern distributions(in my case is Arch Linux). It is important to be convenient for common usecase, without judging if this method is political right.

johannbg:
I have absolutely no misunderstanding what CentOS is and epel is not "really really necessary for developers" and the correct place to advocate for it to be included, is downstream with CentOS community, you need to ask the CentOS community why the epel repository is not shipped/included by default with their distribution ( not sure if the argument that CentOS is just rebuilds of RHEL applies anymore after RH bought CentOS ).  Upstreams cant start including 3rd party repository's ( epel, corp/rpmfusion etc ) because it will lead to package collision just like happened back in the day with freshrpm's. livna, dag,dribble etc.

The fact is, mkosi does not scale very well as is, let alone if we add 200+ systemd based distribution, their unsupported releases and on top of that every possible combination of 3rd party repo's/addon out there ( combination in which we would probably have to then start testing for as well ).

If anything CentOS along with any non primary distribution ( primary distributions are Arch,Debian,Fedora and OpenSuse ) should be removed from Mkosi until it has been re-factured to deal with this and other  scalability issue it suffers from before inclutions like these ( added distributions or 3rd party repo's other distro's  ) grow out of hand.

lucasdemarchi:
@sequencer if you want to add additional repositories for your own use, I recommend using the skeleton tree feature. I implemented that for my needs so I could add copr repositories for my own test images.  It's not possible to add third-party repositories like you are doing because it will conflict with people that don't have a use for them.

lucasdemarchi:
> If anything CentOS along with any non primary distribution ( primary distributions are Arch,Debian,Fedora and OpenSuse ) should be removed from Mkosi until it has been re-factured to deal with this and other scalability issue

where is this coming from? don't just go and build a list of your preferred distributions and say the rest should be removed.

sequencer:
Hello, @lucasdemarchi, with respect to official CentOS, `epel` is a semi-official repo that maintained by fedora guys. If users wanna use `--packages` to preinstall packages like `python36`, `nginx`, this repo should be added.
As for the rest repos `CentOSPlus` And `Extras` Are both from the official CentOS group, which are the default repo for standard CentOS distribution.

sequencer:
In other way, if you think `epel` is not an appropriate repo to be added to CentOS, how about accepting an API for letting use being able to add their own repos? For example a `mkosi.repos` file to use.
In current state of `mkosi`, packages not existed in the default repo cannot be installed by default in any term of use. If `epel` cannot be PRed in this way, how about accepting PR for this API?

lucasdemarchi:
@sequencer have you seen my suggestion of using skeleton tree? It was done specifically for that purpose. In that dir you should basically have a pre-populated `/etc/yum.d` with the additional repos.

poettering:
hmm, maybe introdue a new distro "centos-epel" which is just like centos but also includes epel?

I think covering epel is probably worth it, but maybe not under the normal "centos" name. if you follow what i mean? 

"centos-epel" would use the same code paths as "centos", but just add the repos your patch currently does.

Does that sound like a compromise?

sequencer:
Alright! Thank you! I’ll modify this PR!

johannbg:
> hmm, maybe introdue a new distro "centos-epel" which is just like centos but also includes epel?
> 
> I think covering epel is probably worth it, but maybe not under the normal "centos" name. if you follow what i mean?
> 
> "centos-epel" would use the same code paths as "centos", but just add the repos your patch currently does.
> 
> Does that sound like a compromise?

This will just make matters worse and cause confusion. 

If the distribution implementation would be limited by specific set of package managers then it would make sense ( well it always makes sense ) to move the distribution specific code out of mkosi and into distribution configuration files in which adding repositories would get defined there ( along with gpg keys where applicable ) so you end up with something more in line with this, which would open up the possibility for downstream to add whatever repository they would like to ( local or third party whatever ). 

Mkosi would then just ship/support the default downstream supported versions of the the primary ones ( which we can test ) as in those packaged universally on all distributions ( Arch,Debian,Fedora,OpenSuse ) and downstream ( users,developers,distributions,vendors ) could just customize it as they see fit by adding a distribution tailored to their needs into something like /usr/lib/mkosi/distributions folder downstream distribution packaging perspective  or mkosi.distributions folder in users,developers,distributions,vendors code repository.

fedora.distribution

[Distribution]
 Distribution=Fedora
 Release=30

[Partitions]
 ESPSize=256M
 RootSize=3G
 SwapSize=1024M

[PackageManager]
  Packagemanager=dnf
  Key=CFC659B9
  Repository= ( could be distribution default )
  Repository= ( could be distribution 3rd party rpmfusion,copr in case of centos EPEL )
  Repository= ( could be local repository )
  Repository= ( could be vendor specific repository )   
  
.....

[Packages]
BuildPackages=
....
Packages=
...
   
[Output]
Bootable=yes
Format=raw_gpt
Output=/var/lib/libvirt/images/my-fedora.raw

sequencer:
cause `extras_url` and `centosplus_url` is currently a mainline `centos` repo, so it should be add to the standard `centos` release, maybe nobody use packages in those repos, so no bug report for them till now.
Now `epel` is enabled only when in distro is `centos_epel`.

sequencer:
Hello, @poettering. Any update for this?

stejoo:
Thanks for including support for EPEL, it has the potential to make my life easier. I've been trying to use the `centos_epel` "distro" to build an image recently. However was unable to get it going correctly due to yum bailing out. First of all the CentOS Plus repository did not work and EPEL failed to download as well.
So I dove into the code. I have fixed both these issues (and ran into another one for which I will create a bug report) and created a pull request just now: https://github.com/systemd/mkosi/pull/346

poettering:
@sequencer can you comment on #355 please? I don't know epel/centos really, kinda hoping for your input there, given you did the original patch?

PR #350, opened 2019-07-08
Add command to generate a manpage

fsateler:
And fix a small formatting error.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -4,7 +4,7 @@
 
 # NAME
 
-mkosi — Build Legacy-Free OS Images
+mkosi - Build Legacy-Free OS Images
 
 # SYNOPSIS
 
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,19 @@
 
 import sys
 
-from setuptools import setup
+from setuptools import setup, Command
+
+class BuildManpage(Command):
+    description = ('builds the manpage')
+    user_options = []
+
+    def initialize_options(self):
+        pass
+    def finalize_options(self):
+        pass
+
+    def run(self):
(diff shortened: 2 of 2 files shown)


mkosi.md
```diff
@@ -4,7 +4,7 @@
 
 # NAME
 
-mkosi — Build Legacy-Free OS Images
+mkosi - Build Legacy-Free OS Images
```
poettering: lexgrog? what#s that? and isn't this something to fix in lexgrog? i mean, it's 2019, utf8 support isn't precisely exotic these days...

fsateler: This is what lintian says:

```
W: manpage-has-bad-whatis-entry
N:
N:   Each manual page should start with a "NAME" section, which lists the
N:   name and a brief description of the page separated by "\-". The "NAME"
N:   section is parsed by lexgrog and used to generate a database that's
N:   queried by commands like apropos and whatis. This tag indicates that
N:   lexgrog was unable to parse the NAME section of this manual page.
N:   
N:   For manual pages that document multiple programs, functions, files, or
N:   other things, the part before "\-" should list each separated by a
N:   comma and a space. Each thing listed must not contain spaces; a man
N:   page for a two-part command like "fs listacl" must use something like
N:   "fs_listacl" in the "NAME" section so that it can be parsed by
N:   lexgrog.
N:   
N:   Refer to the lexgrog(1) manual page, the groff_man(7) manual page, and
N:   the groff_mdoc(7) manual page for details.
N:   
N:   Severity: normal, Certainty: certain
N:   
N:   Check: manpages, Type: binary
N:
```

So I guess the difference between `\-` and `[em]` in groff is significant for this...

PR #352, opened 2019-07-26
GPT_Squashfs fix

juan-riveros:
Before this change, gpt_squashfs would just fail to build on my systems (Ubuntu 18.04 LTS and Fedora 30). Tested it against gpt_ext4 and that built. 

Within determine_partition_table, `args.root_partno` should be set to `None` and checked against the OutputFormat being not gpt_squashfs.


juan-riveros:
This allowed the creation of a bootable squashfs, but during boot it failed to mount the squashfs partition.

PR #355, opened 2019-08-22
mkosi: Fix centos_epel mirror configuration

stejoo:
Fix the mirrors for the `centos_epel` _distribution_.
Related: https://github.com/systemd/mkosi/pull/346


stejoo:
In short: `centos_epel` distribution could use broken mirror configuration when falling back on the defaults. I altered the mirrorlist URL for the centosplus repository and changed the mirrorlist for EPEL to a baseurl, because that's what it should be.
Works for me now when building a non-bootable image using:
```
$ sudo mkosi --distribution centos --format raw_gpt --output centos.img --password ohmypassphrase
```

poettering:
hmm @sequencer any chance you can comment on this?

sequencer:
I’ll check this tonight. Sorry for the delay.

sequencer:
I just checked the upstream config, @stejoo is right.
These are my fault, sorry for the bug I introduced.

PR #361, opened 2019-09-10
Add the ability to set pre-hashed root passwords

### LINKED ISSUE #360: Allowing using an already hashed root password
Currently `--password=` and `mkosi.rootpw` only allow cleartext passwords. It would be nice if one could provide an already hashed password as well.

behrmann:
This PR introduces a new commandline switch `--password-is-hashed` such that the supplied root password will not be hashed, but written to `/etc/shadow` literally.

Closes #360


behrmann:
Sorry, I overlooked the tests for the configparser.

PR #363, opened 2019-09-18
Include the .git directory with the files it contains

purpleidea:
When copying files in from git, this also pulls in the .git/ directory
so that builds that use this information will have it available. For
example, builds that run a `git describe` to determine a build test
version number will now be able to query it.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -98,6 +98,7 @@ class SourceFileTransfer(enum.Enum):
     copy_all = "copy-all"
     copy_git_cached = "copy-git-cached"
     copy_git_others = "copy-git-others"
+    copy_git_more = "copy-git-more"
     mount = "mount"
 
     def __str__(self):
@@ -108,6 +109,7 @@ class SourceFileTransfer(enum.Enum):
         return {cls.copy_all: "normal file copy",
                 cls.copy_git_cached: "use git-ls-files --cached, ignoring any file that git itself ignores",
                 cls.copy_git_others: "use git-ls-files --others, ignoring any file that git itself ignores",
+                cls.copy_git_more: "use git-ls-files --cached, ignoring any file that git itself ignores, but include the .git/ directory",
                 cls.mount: "bind mount source files into the build image"}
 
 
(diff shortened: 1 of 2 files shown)


purpleidea:
In combination with https://github.com/systemd/mkosi/pull/365 this makes `mkosi` very useful for combining with fpm to build packages in the correct OS environments.

lucasdemarchi:
In the past I added `mount` option to bind mount the source tree, which then includes the .git. Is it sufficient for you?

purpleidea:
@lucasdemarchi I think the copy is better so there is no chance of writing or editing the source dir. There's already a problem with this since it runs as root. See: https://github.com/systemd/mkosi/issues/366

purpleidea:
Hi, can we merge this? Thanks!

keszybz:
I don't think this is a good idea. In many cases .git can be very large, and the whole point of using `git ls-files` to have a list of files was to avoid copying stuff. If something requires .git to be present to build correctly, then this thing is broken. Copying the whole .git dir into the build image just to get a version seems very inefficient.

purpleidea:
@keszybz I can change the patch to put it behind a flag if you want. In my performance tests it only adds about a second or two, and some projects benefit from having the git history to build. (Mine needs this for example.)

keszybz:
It obviously depends on the project. For small projects it obviously doesn't matter.

So... maybe add another setting to SourceFileTransfer `copy_git_more`?

purpleidea:
On Tue, Oct 8, 2019 at 7:17 AM Zbigniew Jędrzejewski-Szmek
<notifications@github.com> wrote:
>
> It obviously depends on the project. For small projects it obviously doesn't matter.
>
> So... maybe add another setting to SourceFileTransfer copy_git_more?

Sounds good, I'll add that as a flag and ping again when the new PR is
ready. Thanks for the review!

poettering:
i think too it would make most sense if this was a separate option.

purpleidea:
@keszybz 

> So... maybe add another setting to SourceFileTransfer copy_git_more?

@poettering 

> i think too it would make most sense if this was a separate option.

Patch updated with the requested option as requested.

Also rebased to current git master.

Thanks!

purpleidea:
\o/ Thanks!

PR #365, opened 2019-09-18
Add flag to skip (second) final image build phase

purpleidea:
This adds a new flag to cause the (second) final stage to be skipped.
This is useful for when you're more interested in the build artifacts
being built in the correct environment, and less interested in the final
image.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -3385,6 +3385,7 @@ def create_parser() -> ArgumentParserMkosi:
     group.add_argument("--build-directory", dest='build_dir', help='Path to use as persistent build directory', metavar='PATH')
     group.add_argument("--build-package", action=CommaDelimitedListAction, dest='build_packages', default=[],
                        help='Additional packages needed for build script', metavar='PACKAGE')
+    group.add_argument("--skip-final-phase", action=BooleanAction, help='Skip the (second) final image building phase.', default=False)
     group.add_argument("--postinst-script", help='Postinstall script to run inside image', metavar='PATH')
     group.add_argument("--finalize-script", help='Postinstall script to run outside image', metavar='PATH')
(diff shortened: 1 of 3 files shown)


purpleidea:
In combination with https://github.com/systemd/mkosi/pull/363 this makes `mkosi` very useful for combining with fpm to build packages in the correct OS environments.

mkosi
```diff
@@ -4422,7 +4424,10 @@ def build_stuff(args: CommandLineArguments) -> None:
     run_finalize_script(args, workspace.name, verb='final')
 
     # Run the image builder for the second (final) stage
-    raw, tar, root_hash = build_image(args, workspace, do_run_build_script=False, cleanup=True)
+    if args.skip_final_phase:
+        print_step('Skipping (second) final image build phase.')
+    else:
+        raw, tar, root_hash = build_image(args, workspace, do_run_build_script=False, cleanup=True)
```
lucasdemarchi: But you're not skipping the steps below... In your usecase you may also want to make sure type is set to directory.

purpleidea: I'm not sure I understand your comment, but I tested the patch and it seems to work as I expected. Did I miss something?

purpleidea:
Hi, can we merge this? Thanks!

mkosi
keszybz: `raw`, `tar`, `root_squash` are used unconditionally right below. If this call to `build_image` is skipped, they might not be set at all.

poettering:
also, needs rebase

purpleidea:
@keszybz 

> raw, tar, root_squash are used unconditionally right below. If this call to build_image is skipped, they might not be set at all.

Good catch, thank you. Fixed.

@poettering 

> also, needs rebase

Rebased to current git master.

Thank you!

purpleidea:
\o/ Thanks!

PR #367, opened 2019-09-20
Add env to build that specifies which "default" we're using

purpleidea:
It's not possible to determine which "mkosi.default" we're using, when
using a single "mkosi.build" script with `mkosi --default <file> build`
so this adds a variable that allows the build script to know which one
is being used.


purpleidea:
One more =D

purpleidea:
@keszybz Thank you! \o/

PR #368, opened 2019-09-27
Fix type error on args variable

behrmann:
The CommandLineArguments are in `a` instead of `args`, so the `hasattr` check will always fail.


behrmann:
I've not fully checked whether `directory` is always a key in `CommandLineArgs`. I have only checked that it's set calling mkosi with and without `--directory`. Is this assumption correct?

Issue #358, opened 2019-09-01
Installation of additional packages fails for Arch images


rgeorgiev583:
```
# mkosi -o /tmp/mkosi/arch --default arch.default build
‣ Processing default...
COMMANDS:
                  verb: build
               cmdline: 
DISTRIBUTION:
          Distribution: arch
               Release: n/a
                Mirror: https://mirrors.kernel.org/archlinux

OUTPUT:
              Hostname: pensieve
         Output Format: directory
                Output: /tmp/mkosi/arch
       Output Checksum: n/a
      Output Signature: n/a
           Output Bmap: n/a
Output nspawn Settings: n/a
           Incremental: no
             Read-only: no
        FS Compression: no
        XZ Compression: no
            Encryption: no
                Verity: no

PACKAGES:
              Packages: base base-devel xorg-server plasma plasma-meta kde-applications kde-applications-meta
         Package Cache: none
           Extra Trees: none
        Skeleton Trees: none
          Build Script: none
         Build Sources: /tmp
  Source File Transfer: none
       Build Directory: none
        Build Packages: none
    Postinstall Script: none
       Finalize Script: none
  Scripts with network: no
       nspawn Settings: none

HOST CONFIGURATION:
    Extra search paths: none
‣ Detaching namespace...
‣ Detaching namespace complete.
‣ Setting up package cache...
‣ Setting up package cache /tmp/mkosi/.mkosi-bar9xzuw complete.
‣ Setting up temporary workspace.
‣ Temporary workspace in /tmp/mkosi/.mkosi-2a8v2n64 is now set up.
‣ Mounting image...
‣ Mounting image complete.
‣ Setting up basic OS tree...
‣ Setting up basic OS tree complete.
‣ Mounting Package Cache...
‣ Mounting Package Cache complete.
‣ Installing Arch Linux...
gpg: /tmp/mkosi/.mkosi-2a8v2n64/root/etc/pacman.d/gnupg/trustdb.gpg: trustdb created
gpg: no ultimately trusted keys found
gpg: starting migration from earlier GnuPG versions
gpg: porting secret keys from '/tmp/mkosi/.mkosi-2a8v2n64/root/etc/pacman.d/gnupg/secring.gpg' to gpg-agent
gpg: migration succeeded
gpg: Generating pacman keyring master key...
gpg: key B0A9F5C0515D1F30 marked as ultimately trusted
[...]

PR #373, opened 2019-10-22
Make sure the gpg agent is terminated even if an exception occurs

juergenhoetzel:
This prevents EBUSY exceptions when TemporaryDirectory finalizer is
invoked (#358).

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1935,93 +1935,94 @@ SigLevel    = Required DatabaseOptional TrustAll
         cmdline = ["pacstrap", "-C", pacman_conf, "-dGM", root]
         run(cmdline + list(packages), check=True)
 
-    keyring = "archlinux"
-    if platform.machine() == "aarch64":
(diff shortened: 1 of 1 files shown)


keszybz:
Can we please use a context manager for this?

juergenhoetzel:
> Can we please use a context manager for this?

What kind of implementation with `@contextmanager`  do you have in mind?  Annotating `install_arch` would  mean that the implementation of `install_distribution` (using `with`) and all other `install_*`  functions (also use `contextmanager`)  would have to be changed.

behrmann:
The idea, I think, would be to make some function `cleanup_gpg` (or similarly named) that is annotated with `@contextlib.contextmanager` and wraps this try-block such that it yields after the try and does the cleanup in the finally-clause, see `mount_image` for an example.

DaanDeMeyer:
Superseded by #441

PR #375, opened 2019-10-29
Fix errors in mypy's strict mode

behrmann:
Most of these changes should be fairly uncontroversial, they only add missing type hints or remove type comments and casts that are no longer needed, when using a recent mypy.

Some additions are a bit busy work, as they silence warnings about generics (mostly return values of `CompletedSubprocess` or `TemporaryDirectory`), or are somewhat stylistic (usage of `__class__`) and some work around limitations in mypy (values in enums having type `Any` or the weird power issue in `parse_bytes`)

```mkosi:432: error: Missing type parameters for generic type "Callable"``` (the typing for the `complete_step` decorator) is left open, as this is an open mypy issue and I haven't found a way to work around it yet.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -409,7 +409,7 @@ def copy_path(oldpath: str, newpath: str) -> None:
             symlink_f(target, newentry)
             shutil.copystat(entry.path, newentry, follow_symlinks=False)
         else:
-            st = entry.stat(follow_symlinks=False)  # type: ignore  # mypy 0.641 doesn't know about follow_symlinks
+            st = entry.stat(follow_symlinks=False)
             if stat.S_ISREG(st.st_mode):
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by behrmann** [comment]
The breaking run of semaphoreci seems to be due to a too old mypy
```
+ sudo python3.6 ./mkosi --default ./mkosi.files/mkosi.ubuntu
  File "./mkosi", line 69, in <module>
    def run(cmdline: List[str], execvp: bool = False, **kwargs: Any) -> subprocess.CompletedProcess[Any]:
TypeError: 'type' object is not subscriptable
```

I'll have a look how the tests are run. It might make sense to pip install a recent mypy.

behrmann:
The error was rather through types that are generic during type-checking, but are not during runtime.

PR #377, opened 2019-10-31
Add missing packages on arch

keur:
With Bootable=no mkosi will fail on arch with the following errors.

Generating locales...
/usr/bin/locale-gen: line 33: sed: command not found
  .UTF-8/usr/bin/locale-gen: line 35: sed: command not found
.../usr/bin/locale-gen: line 38: sed: command not found

/tmp/alpm_PENqPM/.INSTALL: line 3: chmod: command not found
/tmp/alpm_PENqPM/.INSTALL: line 4: chmod: command not found
error: command failed to execute correctly

The coreutils package installs chmod, and the sed package installs sed.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1943,23 +1943,13 @@ SigLevel    = Required DatabaseOptional TrustAll
 
     run_pacman(["-Sy"])
     # determine base packages list from base group
-    c = run_pacman(["-Sqg", "base"], stdout=PIPE, universal_newlines=True)
-    packages = set(c.stdout.split())
-    packages -= {
-        "cryptsetup",
-        "device-mapper",
-        "dhcpcd",
-        "e2fsprogs",
-        "jfsutils",
-        "linux",
-        "lvm2",
-        "man-db",
-        "man-pages",
-        "mdadm",
-        "netctl",
-        "reiserfsprogs",
-        "xfsprogs",
-    }
+    c = run_pacman(["-Si", "base"], stdout=PIPE, universal_newlines=True)
+    packages = set()
+    base_re = re.search("^Depends On\s+:(.*?)(?=^Optional Deps)", c.stdout, flags=re.S | re.M)
+    if base_re:
+        packages |= set(base_re.group(1).strip().split())
+    else:
+        die("Arch base package could not be found.")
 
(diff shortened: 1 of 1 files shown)


marcosfrm:
What about https://www.archlinux.org/news/base-group-replaced-by-mandatory-base-package-manual-intervention-required/ ?

keur:
@marcosfrm Thanks! Look better now?

behrmann:
Just this morning I thought anout the `base` package change. Good to see somebody already put in the work. :)

Some comments:
1. `run_pacman(["-Si", "base"])` should probably be `-Sqi` instead.
2. `packages = set()` seems a bit unneccary, you could just do `packages = ` instead of `packages |=` in the if branch
3. The regular expression approach seems a bit fragile, if the output of pacman changes, this might break. Since the announcement is quite clear, that the `base` package must be installed to request support, I guess we should just install the base package, which is now also a lot slimmer than the base package group used to be, i.e. the whole section could just be changed to 
```python
packages = { "base" }
```
What do you think @marcosfrm?

marcosfrm:
> 3. The regular expression approach seems a bit fragile, if the output of pacman changes, this might break. Since the announcement is quite clear, that the `base` package must be installed to request support, I guess we should just install the base package, which is now also a lot slimmer than the base package group used to be, i.e. the whole section could just be changed to

I agree.

craftyguy:
> 3. The regular expression approach seems a bit fragile, if the output of pacman changes, this might break. Since the announcement is quite clear, that the `base` package must be installed to request support, I guess we should just install the base package, which is now also a lot slimmer than the base package group used to be, i.e. the whole section could just be changed to
> 
> ```python
> packages = { "base" }
> ```

I can confirm that this change to the proposed patch works.

lucasdemarchi:
> `packages = { "base" }`

Lgtm. Could any of you update this PR?

behrmann:
I've reworked the patch, since @craftyguy said it works, and opened #381, since I obviously can't edit this PR.

keur:
Closing in favor of https://github.com/systemd/mkosi/pull/381

Sorry I didn't get back to you guys. Was in the wilderness with no Internet.

Issue #384, opened 2019-11-17
Support systemd-firstboot


shibumi:
Hi,
I try to generate raw_gpt images with enabled systemd-firstboot.

Here are my files:

minimal.default
```
[Distribution]
Distribution=arch

[Output]
Format=raw_gpt
Bootable=yes
Output=minimal.img
```

minimal.postinst:
```
#!/bin/bash

cat > /etc/systemd/system/systemd-firstboot.service << EOF
[Unit]
Description=First Boot Wizard
Documentation=man:systemd-firstboot(1)
DefaultDependencies=no
Conflicts=shutdown.target
After=systemd-remount-fs.service
Before=systemd-sysusers.service sysinit.target shutdown.target
ConditionPathIsReadWrite=/etc
ConditionFirstBoot=yes

[Service]
Type=oneshot
RemainAfterExit=yes
ExecStart=/usr/bin/systemd-firstboot --prompt --setup-machine-id
StandardOutput=tty
StandardInput=tty
StandardError=tty

[Install]
WantedBy=sysinit.target
EOF

systemctl enable systemd-firstboot.service
rm /etc/{machine-id,localtime,shadow,locale.conf}
```

I build them via: 
```
sudo mkosi --default minimal/minimal.default --postinst minimal/minimal.postinst --root-size=2G --force
```

The build works so far, what is not working is booting into `systemd-firstboot`.
When the VM boots up, the system will have machine-id file and therefore it's not possible to boot into systemd-firstboot. Here is my qemu line: 

```
qemu-system-x86_64 -m 512 -smp 2 -bios /usr/share/ovmf/x64/OVMF_CODE.fd -drive format=raw,file=minimal.img
```

DaanDeMeyer:
We explicitly remove the machine-id from the final image now so this should work, let me know if it doesn't and I can reopen

shibumi:
Hi @DaanDeMeyer  I tried this and it doesn't work. The VM is not booting into firstboot mode.

behrmann:
@shibumi Fixed by #1118

Issue #376, opened 2019-10-30
Adopt /etc/kernel/entry-token once all distros have sufficiently new bootctl/kernel-install


lucasdemarchi:
I have the following problem when generating bootable Fedora images:

- A directory /efi/\<fake-machine-id\> is leftover in the image
- /etc/machine-id is wiped so a new one is created on first boot
- kernel-install fails to install new kernels when the dir /efi/\<machine-id\> doesn't exist

I use kernel-install in the final image to install additional kernels for testing. I don't know exactly where to fix this current behavior. My current approach is to always remove the fake-machine-id dir and add a new one, but there should be a better way.

Related question: could we have a toggle to avoid using a unified kernel? For my use cases having a bootloader conf entry with separate initrd and kernel image would be better.

DaanDeMeyer:
Might be interested in this PR: https://github.com/systemd/systemd/pull/16687

DaanDeMeyer:
We should be able to fix this with the new /etc/kernel/entry-token support that was added to bootctl and kernel-install. It might just take a while until we have sufficiently new systemd in all distros to be able to depend on it.

We already have an option to not use unified kernel images.

DaanDeMeyer:
We don't create an entry-token directory at all anymore, so let's close this

PR #385, opened 2019-11-17
add firstboot parameter

shibumi:
When generating cloud images or other images that should use
systemd-firstboot, we want to make sure that /etc/machine-id is not
existent. Creating an empty /etc/machine-id file is not enough, because
systemds `ConditionFirstBoot` will only check for the file, not its
content. Furthermore systemd will generate a new /etc/machine-id file if
the /etc/machine-id file is empty. This is supposed to fix https://github.com/systemd/mkosi/issues/384

Signed-off-by: Christian Rebischke <chris@nullday.de>

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -2483,7 +2483,8 @@ def reset_machine_id(args: CommandLineArguments, root: str, do_run_build_script:
             os.unlink(machine_id)
         except FileNotFoundError:
             pass
-        open(machine_id, "w+b").close()
+        if not args.firstboot:
+            open(machine_id, "w+b").close()
         dbus_machine_id = os.path.join(root, 'var/lib/dbus/machine-id')
         try:
             os.unlink(dbus_machine_id)
@@ -3661,6 +3662,7 @@ class ArgumentParserMkosi(argparse.ArgumentParser):
         'QCow2': '--qcow2',
         'OutputDirectory': '--output-dir',
         'XZ': '--xz',
+        'Firstboot': '--firstboot',
         'NSpawnSettings': '--settings',
         'ESPSize': '--esp-size',
         'CheckSum': '--checksum',
@@ -3782,6 +3784,7 @@ def create_parser() -> ArgumentParserMkosi:
                        help='Sign the resulting kernel/initrd image for UEFI SecureBoot')
     group.add_argument("--secure-boot-key", help="UEFI SecureBoot private key in PEM format", metavar='PATH')
     group.add_argument("--secure-boot-certificate", help="UEFI SecureBoot certificate in X509 format", metavar='PATH')
+    group.add_argument("--firstboot", action=BooleanAction, help="Do not add /etc/machine-id to the image")
     group.add_argument("--read-only", action=BooleanAction,
                        help='Make root volume read-only (only gpt_ext4, gpt_xfs, gpt_btrfs, subvolume, implied with gpt_squashfs and plain_squashfs)')
     group.add_argument("--encrypt", choices=("all", "data"),
@@ -4662,6 +4665,7 @@ def print_summary(args: CommandLineArguments) -> None:
     sys.stderr.write("        FS Compression: " + yes_no(args.compress) + detail + "\n")
 
     sys.stderr.write("        XZ Compression: " + yes_no(args.xz) + "\n")
+    sys.stderr.write("             FirstBoot: " + yes_no(args.firstboot) + "\n")
     if args.mksquashfs_tool:
         sys.stderr.write("       Mksquashfs tool: " + ' '.join(args.mksquashfs_tool) + "\n")
 
--- a/mkosi.md
+++ b/mkosi.md
@@ -265,6 +265,11 @@ details see the table below.
   to the kernel command line created by lower priority configuration
   files or previous --kernel-command-line command line arguments.
 
+`--firstboot`
+
+: Do not generate /etc/machine-id file. This is useful, for images that 
+  shall trigger systemd-firstboot.service.
+
 `--secure-boot`
 
 : Sign the resulting kernel/initrd image for UEFI SecureBoot
@@ -692,6 +697,7 @@ which settings file options.
 | `--bootable`, `-b`           | `[Output]`              | `Bootable=`               |
 | `--boot-protocols=`          | `[Output]`              | `BootProtocols=`          |
 | `--kernel-command-line=`     | `[Output]`              | `KernelCommandLine=`      |
+| `--firstboot`                | `[Output]`              | `FirstBoot=`              |
 | `--secure-boot`              | `[Output]`              | `SecureBoot=`             |
 | `--secure-boot-key=`         | `[Output]`              | `SecureBootKey=`          |
 | `--secure-boot-certificate=` | `[Output]`              | `SecureBootCertificate=`  |
--- a/tests/test_config_parser.py
+++ b/tests/test_config_parser.py
@@ -89,6 +89,7 @@ class MkosiConfig(object):
             'release': None,
             'repositories': [],
             'root_size': None,
+            'firstboot': False,
             'secure_boot': False,
             'secure_boot_certificate': None,
             'secure_boot_key': None,
@@ -195,6 +196,8 @@ class MkosiConfig(object):
                 self._append_list('boot_protocols', mk_config_output['BootProtocols'], job_name)
             if 'KernelCommandLine' in mk_config_output:
                 self._append_list('kernel_command_line', mk_config_output['KernelCommandLine'], job_name, ' ')
+            if 'FirstBoot' in mk_config_output:
+                self.reference_config[job_name]['first_boot'] = mk_config_output['FirstBoot']
             if 'SecureBoot' in mk_config_output:
                 self.reference_config[job_name]['secure_boot'] = mk_config_output['SecureBoot']
             if 'SecureBootKey' in mk_config_output:


shibumi:
anyone an idea, what I am missing for succeeding the checks?

poettering:
hmm, i wonder if it wouldn't be nicer to instead change systemd to consider /etc/machine-id empty a first boot case too, in light of https://github.com/systemd/systemd/issues/4511 we need to do thta anyway... that should fix your issue too, right? after all we already reset the machine ID file anyway.

shibumi:
I am not sure. Right now we have three options (and I kind of like this approach):

1.  /etc/machine-id contains an id. everything is fine, systemd will boot and use the machine-id in /etc/machine-id.
2. /etc/machine-id is empty: systemd will boot and generate an machine-id itself, even if systemd-firstboot is enabled
3. /etc/machine-id is not existent: systemd will boot and start systemd-firstboot.

behrmann:
@shibumi When adding a new commandline option you also need to add it in the tests, that's why the tests fail.

shibumi:
@behrmann can you point me to the test files? And before I further work on this, it would be cool if we could make a decision on either implementing this in mkosi or patching systemd (like @poettering has proposed). 

I am not 100% against @poettering's proposal, but I am unsure if changing such behavior wouldn't hurt some systems (just imagine systems booting in first-boot suddenly and therefore resetting system configuration). Can somebody explain this further?

behrmann:
That would be `tests/test_config_parser.py` Have a look at c8bfb0b5490d8297d5c8ad3878a6f6d2c7728c27 for an example. If you finish your work on this, please also add documention on the new commandline switch.

mkosi
```diff
@@ -3348,6 +3350,7 @@ def create_parser() -> ArgumentParserMkosi:
                        help='Sign the resulting kernel/initrd image for UEFI SecureBoot')
     group.add_argument("--secure-boot-key", help="UEFI SecureBoot private key in PEM format", metavar='PATH')
     group.add_argument("--secure-boot-certificate", help="UEFI SecureBoot certificate in X509 format", metavar='PATH')
+    group.add_argument("--firstboot", action=BooleanAction, help="removes /etc/machine-id file")
```
behrmann: This is different from your documentation. Maybe 
```suggestion
    group.add_argument("--firstboot", action=BooleanAction, help="Do not add /etc/machine-id to the image")
```

shibumi:
@behrmann thanks, I have edited it in my branch and squashed the commits. Please verify :)

behrmann:
I've tested the PR and it works. LGTM.

shibumi:
Hi,
Can we get this merged?

lucasdemarchi:
@poettering what do you think?

DaanDeMeyer:
This would be great to have. We could extend this with settings to run systemd-firstboot on the image as well when running mkosi. (Doesn't have to be in this PR, I'll add those myself)

shibumi:
@DaanDeMeyer what do you mean with your second sentence? systemd-firstboot should get activated automatically, if certain files are missing (like /etc/machine-id, /etc/vconsole.conf .. etc)

DaanDeMeyer:
There's also the command line tool that you can run to configure stuff when building the image. Check https://www.freedesktop.org/software/systemd/man/systemd-firstboot.html

mkosi.md
```diff
@@ -263,6 +263,11 @@ details see the table below.
   to the kernel command line created by lower priority configuration
   files or previous --kernel-command-line command line arguments.
 
+`--firstboot`
+
+: Do not generate /etc/machine-id file. This is useful, for images that 
+  shall trigger systemd-firstboot.service
```
DaanDeMeyer, line 269:
```suggestion
  shall trigger systemd-firstboot.service.
```

shibumi: fixed

johannbg:
Note that there is some bug in the logic ( probably in kernel-install/bootctl ) which creates the efi directory which is based on or relies on the entry in machine-id and has been reported in mkosi, here in systemd and is also affecting fedora silverblue and basically any users that install sd-boot on fedora ( probably affects all distributions except maybe Arch ) when users are trying to use sd-boot.

Based on the kernel-install repair option that I had started working on ( to workaround this issue as well as re-generate it encase the efi setup got lost or corrupted for one reason or another ) the machine-id can be picked up in 5 places which firstboot has to take into account as in "systemd.machine_id=" entry which could be found in /proc/cmdline, /etc/kernel/cmdline, /usr/lib/kernel/cmdline and an entry in /var/lib/dbus/machine-id and finally /etc/machine-id which is being discussed here. 

Presumably firstboot has to check if all these places are empty and then take action and generate a new machine_id and update all the relevant of those 5 places and in addition create a new efi directory and run kernel-install to install the kernels into to a new efi directory based recreate relevant efi conf to prevent boot/kernel updates breaking for users. 

Arguably any *new* logic that relies on existance,non-existance or empty /etc/machine-id or those 5 places cannot be added until the current code has been re-factored to deal with that.

@poettering  What's the reason for using the UUID in machined in bootctl/kernel-install as opposed to use something like the GUID/PARTUUID of the EFI partition itself as in why are the tools depending on machine-id et all?

shibumi:
Thanks @johannbg, I only tested this with Arch Linux so I had no clue, that the situation is such a big problem right now.
If it's really that complicated, we maybe should go with @poettering 's proposal to change systemd's behavior? Would this fix the issue? I doubt so, because @poettering mentioned /etc/machine-id only.

johannbg:
I do believe that all the issues relating to the EFI can be simply fixed by creating the efi directory using the /efi/$PARTUUID of the EFI System Partition itself then just have kernel-install check if the directory exists if so proceed with install, if not create the directory then proceed with the install and have bootctl install initially create the directory then just leave /etc/machine_id empty and or as an symlink to /var/lib/dbus/machine-id and merge this. Afaikt that's the win win solution but without knowing why the efi directory is created using the UUID from machined_id in the first place I cant tell if it's an acceptable solution.

johannbg:
Seems to be line 891+ in bootctl.c and 81+ in kernel-install that would need fixing.

DaanDeMeyer:
@johannbg Could you please provide the links to the relevant issues in systemd/mkosi?

@shibumi How exactly are you testing this? I checked out your branch and tried a few combinations with the --firstboot option with both nspawn booting and qemu booting but I'm not getting any systemd-firstboot prompts. Is there any other configuration I have to do to get systemd-firstboot working?

Whatever fix made in systemd will still take quite a while to be available everywhere else so I'd still like to merge something that makes this work in mkosi generated images if possible.

shibumi:
@DaanDeMeyer which distribution did you use for testing? I tried it with Arch Linux, if you use another distribution you might run into issues that @johannbg has discovered, I wasn't aware of.

Also, I have problems with QEMU + systemd-firstboot right now, in case you are doing full bootable images. I see that systemd-firstboot starts, but I can't type something in. The VM will just reject all key presses and nothing happens.

DaanDeMeyer:
@shibumi I tried arch as well but didn't get the prompt. The exact commands (executed from the mkosi repo dir):

`sudo ./mkosi --default /dev/null -d arch --password '' --firstboot -f boot`
`sudo ./mkosi --default /dev/null -d arch --password '' --firstboot --qemu-headless -f qemu`

systemd-firstboot does seem to get activated when booting with nspawn but I don't get a prompt.

shibumi:
@DaanDeMeyer This is because it will generate the machine-id on first-boot silently.
If you want to have a prompt, you need to make sure to delete various artifacts:

Use this as `mkosi.postinst`:
```sh
#!/bin/bash

# clean locale
rm /etc/locale.cconf
# clean timezone
rm /etc/localtime
# clean keymap
rm /etc/vconsole.conf
# clean hostname
rm /etc/hostname
# clean shadow file
rm /etc/shadow
```

DaanDeMeyer:
OK. I would add a little note to the --firstboot documentation mentioning that. 

I wonder if we should go further and, when --firstboot is specified, delete all config files that are not explicitly configured by the user. As an example, if --firstboot is specified but --password isn't, delete /etc/shadow so the user is prompted for a root password. Same logic would apply to the other firstboot options.

shibumi:
@DaanDeMeyer this is a good idea, but I don't know if I have time to further work on this PR in near future (finishing my master in Computer Science right now...)

behrmann:
I think we can iterate on this later. Also, maybe we could make `--firstboot` just take an optional value to denote what to clean out of the image. The coupling between different switches is difficult enough as it is already.

DaanDeMeyer:
I verified this works on Arch and not on Ubuntu at the moment.

johannbg:
> @johannbg Could you please provide the links to the relevant issues in systemd/mkosi?
@DaanDeMeyer Here are few ( there could be more ) also note any empty or non existent /etc/machined-id and or changes to bootctl and kernel-install might break dracut and potentially other initramfs generators
@poettering probably knows about bunch of other places where it might be used and could comment on, if using $UUID or $PARTUUID of the EFI System Partition instead of the machine-id is a viable option and even changed it in bootctl/kernel-install in an hour tops if that's the case. 

[https://github.com/systemd/systemd/issues/14115](url)
[https://github.com/systemd/systemd/issues/12363](url)
[https://github.com/systemd/mkosi/issues/376](url)

DaanDeMeyer:
@johannbg Thanks for the links. I just encountered the machine-id issue myself in #489 and I agree this is quite broken at the moment. This could be an awkward thing to fix though as its public API at this point.

DaanDeMeyer:

PR #389, opened 2019-12-02
Make mkosi mypy strict compliant and add mypy to testing

### LINKED ISSUE #369: Mypy warnings
Running a current mypy on mkosi gives the following warnings 
```
mkosi:968: error: Argument 1 to "partition" has incompatible type "Optional[str]"; expected "str"
mkosi:971: error: Argument 1 to "partition" has incompatible type "Optional[str]"; expected "str"
mkosi:4624: error: "Dict[str, CommandLineArguments]" has no attribute "director
```

The last line is harmless and addressed in #368. I guess it simply prevents the `--directory` switch to take effect.

I haven't yet followed through the first error, though. The `partition` function cannot deal with the `None` that `mount_image` says `loopdev` might be. The only user of `mount_image` is in `build_image` which gets the `loopdev` it passes on from `attach_image_loopback`, which will produce a `None` if `raw` is `None`. That seems to be the case if the output format is not a disk format. Right now I wonder where to best untangle this. Any ideas?

behrmann:
This PR supersedes #374 and #375 and combines them. Additionally it adds a mypy section to setup.cfg switching on a subset of the strict switches of mypy, disallowing untyped definitions, but keeping out the warnings on Generics, that can sometimes be a bit burdensome. Furthermore it adds running mypy to the tests run by semaphore.

This PR keeps the changes form #374 and removes a couple of workaround that are no longer needed with mypy 0.750 from #375.

Fixes #369

### DIFF
--- a/.github/workflows/ci-mypy.yml
+++ b/.github/workflows/ci-mypy.yml
@@ -1,0 +1,14 @@
+name: CI Unit Test
+on: [push]
+jobs:
+  unit-test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v1
+      with:
+        python-version: '3.x'
+    - name: install mypy via pip
+      run: python3 -m pip install mypy
+    - name: run mypy on mkosi with settings from setup.cfg
+      run: python3 -m mypy mkosi
--- a/.github/workflows/ci-mypy.yml
+++ b/.github/workflows/ci-mypy.yml
@@ -9,6 +9,6 @@ jobs:
       with:
         python-version: '3.x'
     - name: install mypy via pip
-      run: python3 -m pip install mypy
+      run: python3 -m pip install 'mypy==0.770'
     - name: run mypy on mkosi with settings from setup.cfg
       run: python3 -m mypy mkosi
--- a/mkosi
+++ b/mkosi
@@ -1117,10 +1117,10 @@ def mount_image(args: CommandLineArguments,
         if tmp_dev is not None:
             mount_loop(args, tmp_dev, os.path.join(root, "var/tmp"))
 
-        if args.esp_partno is not None:
+        if args.esp_partno is not None and loopdev is not None:
             mount_loop(args, partition(loopdev, args.esp_partno), os.path.join(root, "efi"))
 
-        if args.xbootldr_partno is not None:
+        if args.xbootldr_partno is not None and loopdev is not None:
             mount_loop(args, partition(loopdev, args.xbootldr_partno), os.path.join(root, "boot"))
 
         # Make sure /tmp and /run are not part of the image
--- a/mkosi
+++ b/mkosi
@@ -48,12 +48,24 @@ from typing import (
     TypeVar,
     Union,
     cast,
+    TYPE_CHECKING,
 )
 
 __version__ = '5'
 
 if sys.version_info < (3, 6):
     sys.exit("Sorry, we need at least Python 3.6.")
+
+
+# These types are only generic during type checking and not at runtime, leading
+# to a TypeError during compilation.
+# Let's be as strict as we can with the description for the usage we have.
+if TYPE_CHECKING:
+    CompletedProcess = subprocess.CompletedProcess[Any]
+    TempDir = tempfile.TemporaryDirectory[str]
+else:
+    CompletedProcess = subprocess.CompletedProcess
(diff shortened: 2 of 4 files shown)


behrmann:
Having had a look at all mypy options I decided to switch on all options that are part of strict and then some more for nicer output. I'd also like to switch on testing for unreachable code, but this is currently triggered by a photon code path, but already showed me two things that were missing in the first iteration of this PR.

@keszybz Judging form #369 you seem to be the one to ask for review. I hope this is not too forward. I'd be happy about review once your time allows.

poettering:
So I would review this, but my python-fu is too limited to say something smart about this. @keszybz any chance you can take care of this one?

behrmann:
I rebased the PR due to conflicts with the current master and since the Github actions PR has been merged, I also added one of those.

mkosi
```diff
@@ -439,8 +448,20 @@ def complete_step(text: str, text2: Optional[str] = None) -> Generator[List[Any]
     print_step(text2.format(*args) + '.')
 
 
+# mypy current still has problems with generic decorators:
 # https://github.com/python/mypy/issues/1317
-C = TypeVar('C', bound=Callable)
+# When running mypy in strict mode it considers the complate_step decorator
```
DaanDeMeyer, line 453:
```suggestion
# When running mypy in strict mode it considers the complete_step decorator
```

mkosi
```diff
@@ -439,8 +448,20 @@ def complete_step(text: str, text2: Optional[str] = None) -> Generator[List[Any]
     print_step(text2.format(*args) + '.')
 
 
+# mypy current still has problems with generic decorators:
```
DaanDeMeyer, line 451:
```suggestion
# mypy still has problems with generic decorators:
```

mkosi
```diff
@@ -56,6 +57,14 @@ if sys.version_info < (3, 6):
     sys.exit("Sorry, we need at least Python 3.6.")
 
 
+if TYPE_CHECKING:
```
DaanDeMeyer, line 60: I didn't read the other PRs yet so this might have been clarified already but a comment in the source code explaining why we do this would be great.

DaanDeMeyer:
This looks OK but probably needs a rebase with some changes for recent additions.

mkosi
behrmann: Do you mean a specific comment for this block or why typing in general?

behrmann:
@DaanDeMeyer Thanks for looking at this!

I rebased again current master. Which needed some small fixups (a now outdated return type, and making `fix_permissions_dirs` in `install_arch` a dict to make it palatable for mypy). Unfortunately I had to pin the mypy version at 0.770 for now, because on the current version (0.78*) we are being hit by python/mypy#8962. 

A remaining issue is that `warn_unreachable = True` (which I'd like to turn on as well) flags a check for photon
```python
if "uefi" in args.boot_protocols and args.distribution == Distribution.photon:
```
that is resolved by reducing the indentation level by one (taking it out of the `if "bios" in args.boot_protocols` block). I think there was a tabs vs spaces issue in the Photon PR, that might have screwed this up. I haven't opened an issue for this yet because it's a bit of a chicken and egg problem with the rest of the typing. :)

behrmann:
I had missed pinning the mypy version for semaphore as well.

mkosi
DaanDeMeyer: This block, why typing should hopefully be pretty obvious to everyone.

.github/workflows/ci-mypy.yml
```diff
@@ -0,0 +1,14 @@
...
+  unit-test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v1
+      with:
+        python-version: '3.x'
+    - name: install mypy via pip
+      run: python3 -m pip install 'mypy==0.770'
+    - name: run mypy on mkosi with settings from
```
DaanDeMeyer, line 13: Is there a word missing here?

behrmann: There was, thanks!

mkosi
behrmann: The reason is simply, that these types are only generic during type checking, but not during runtime, making CPython throw its hands up in the air during compilation and throwing a TypeError. `TYPE_CHECKING` is only true during type checking time and guards against this. I've added a comment.

DaanDeMeyer:
LGTM, thanks for working on this!

PR #392, opened 2019-12-13
add support for building "minimized" file system images

poettering:
This adds support for images that have files systems that are as short as possible. This is implemented on ext4 via resize2fs -M (which isn't perfect, the images are still larger than necessary) and on btrfs via mkfs.btrfs --shrink (which works really well).

Also, beef up CI to test images of this type and all others.

(And while we are at it also add a "strict" mode without network. It just passes --nocache to dnf, thus permitting a modicum of reproducibility, as long as the cache is fully populated)

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -4524,32 +4524,35 @@ def build_stuff(args: CommandLineArguments) -> None:
 
         # There is no point generating a pre-dev cache image if no build script is provided
         if args.build_script:
-            # Generate the cache version of the build image, and store it as "cache-pre-dev"
-            raw, tar, root_hash = build_image(args, workspace, do_run_build_script=True, for_cache=True)
-            save_cache(args,
-                       workspace.name,
-                       raw.name if raw is not None else None,
-                       args.cache_pre_dev)
+            with complete_step("Running first (development) stage to generate cached copy"):
+                # Generate the cache version of the build image, and store it as "cache-pre-dev"
+                raw, tar, root_hash = build_image(args, workspace, do_run_build_script=True, for_cache=True)
+                save_cache(args,
+                           workspace.name,
+                           raw.name if raw is not None else None,
+                           args.cache_pre_dev)
 
-            remove_artifacts(args, workspace.name, raw, tar, do_run_build_script=True)
+                remove_artifacts(args, workspace.name, raw, tar, do_run_build_script=True)
 
-        # Generate the cache version of the build image, and store it as "cache-pre-inst"
-        raw, tar, root_hash = build_image(args, workspace, do_run_build_script=False, for_cache=True)
-        if raw:
-            save_cache(args,
-                       workspace.name,
-                       raw.name,
-                       args.cache_pre_inst)
-            remove_artifacts(args, workspace.name, raw, tar, do_run_build_script=False)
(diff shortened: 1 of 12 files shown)


mkosi
```diff
@@ -97,10 +97,10 @@ class CommandLineArguments(argparse.Namespace):
     esp_partno: Optional[int] = None
     xbootldr_partno: Optional[int] = None
 
-    def generated_root(self):
+    def generated_root(self) -> bool:
         """Returns whether this configuration means we need to generate a file system from a prepared tree,
         as needed for anything squashfs and when root minimization is required."""
-        return self.minimize or self.output_format.is_squashfs()
+        return cast(bool, self.minimize) or self.output_format.is_squashfs()
```
keszybz: Shouldn't there be a line like
```python3
minimize: bool = False
```
in the declaration section instead? Having to do a cast sounds wrong.

mkosi.md
```diff
@@ -280,9 +280,18 @@ details see the table below.
...
   `gpt_squashfs` and `plain_squashfs`.
 
+`--minimize`
+
+: Attempt to make the resulting root file system as small as possible,
+  reducing free space on the file system as much as possible. Only
+  supported for `gpt_ext4` and `gpt_btrfs`. For ext4 this relies on
+  `resize2fs -M`, which reduces the free disk space but is not perfect
+  and thus generally leaves some free space in place. For btrfs the
+  results are perfect, leaving no free space.
```
keszybz: "as possible" and "perfect" repeats awkwardly. Maybe:
```
: Attempt to make the resulting root file system as small as possible by
  removing free space from the file system. Only
  supported for `gpt_ext4` and `gpt_btrfs`. For ext4 this relies on
  `resize2fs -M`, which reduces the free disk space but is not perfect
  and generally leaves some free space. For btrfs the
  results are optimal and no free space is left.
```

mkosi.md
```diff
@@ -524,7 +533,11 @@ details see the table below.
   invoked. By default, the build script runs with networking turned
   off. The `$WITH_NETWORK` environment variable is passed to the
   `mkosi.build` build script indicating whether the build is done with
-  or without this option.
+  or without this option. If specified as `--with-network=strict` the
+  package manager is instructed not to contact the network for
+  updating package data. This provides a minimal level of
+  reproducibility, as long as the package data cache is already fully
+  populated.
```
keszybz: The naming of this option seems rather strange: the default is no network, `--with-network` enables network, `--with-network=strict` does the opposite. "strict" is meaningless. Maybe call with `--with-network=never` instead?

poettering:
Ok, let's merge this, will prep follow-up PR right after

mkosi
poettering: hmm, this is something https://github.com/poettering/mkosi/pull/1 added and I just merged. My python-fu is limited. @behrmann can you elaborate?

behrmann: I'm not too happy about the cast myself and would love a cleaner solution. If I remember correctly, `self.minimize` comes directly from argparse via the `BooleanAction` and I haven't yet found the proper incantations to have mypy figure out that the value must hence be boolean. I'll have a look whether mypy can be persuaded to do that.

PR #398, opened 2020-01-18
add basic support for --repositories option for Arch Linux

nicoulaj:
The `--repositories` option was silently ignored for Arch Linux. This patch adds very basic support for this option.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1925,6 +1925,19 @@ SigLevel    = Required DatabaseOptional TrustAll
 {server}
 """)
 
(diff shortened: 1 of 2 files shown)


behrmann:
Neat :) But couldn't the name for the repository not be generated from the URL? This would avoid special-casing Arch.

nicoulaj:
Unfortunately pacman looks for `{repo url}/{repo name}.db`, so the repo name is not just a cosmetic label...

PR #400, opened 2020-02-02
mkosi: add support for rust package manager's cache

ronnychevalier:
If one uses cargo in mkosi.build or mkosi.postinst, cargo will store its registry in mkosi.cache/cargo.
It saves time by avoiding to download the registry and the packages every time.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1064,11 +1064,13 @@ def mount_cache(args: CommandLineArguments, workspace: str) -> Generator[None, N
             mount_bind(args.cache_path, os.path.join(workspace, "root", "var/cache/zypp/packages"))
         elif args.distribution == Distribution.photon:
             mount_bind(os.path.join(args.cache_path, "tdnf"), os.path.join(workspace, "root", "var/cache/tdnf"))
+
+        mount_bind(os.path.join(args.cache_path, "cargo"), os.path.join(workspace, "root", "var/cache/cargo"))
     try:
         yield
     finally:
         with complete_step('Unmounting Package Cache'):
-            for d in ("var/cache/dnf", "var/cache/yum", "var/cache/apt/archives", "var/cache/pacman/pkg", "var/cache/zypp/packages"):  # NOQA: E501
+            for d in ("var/cache/dnf", "var/cache/yum", "var/cache/apt/archives", "var/cache/pacman/pkg", "var/cache/zypp/packages", "var/cache/cargo"):  # NOQA: E501
                 umount(os.path.join(workspace, "root", d))
 
 
@@ -2239,7 +2241,10 @@ def run_postinst_script(args: CommandLineArguments, workspace: str, do_run_build
         shutil.copy2(args.postinst_script,
                      os.path.join(workspace, "root", "root/postinst"))
 
-        run_workspace_command(args, workspace, "/root/postinst", verb, network=args.with_network)
+        env = {
+            "CARGO_HOME": "/var/cache/cargo",
+        }
+        run_workspace_command(args, workspace, "/root/postinst", verb, network=args.with_network, env=env)
         os.unlink(os.path.join(workspace, "root", "root/postinst"))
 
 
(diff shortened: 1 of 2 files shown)


ronnychevalier:
I don't know if it worth it to put a setting in the `[Packages]` section to enable this or not, like `CargoCache=true` (defaults to false) or similar?

behrmann:
Not that I really have something to say about this, but this seems really specific to me. Wouldn't it be better to make the caches hookable in some sense, so that arbitrary caching directories (pip's cache for python or ccache maybe) could be added?

mkosi
```diff
@@ -4565,6 +4570,10 @@ def run_build_script(args: CommandLineArguments, workspace: str, raw: Optional[B
         else:
             cmdline.append("--private-network")
 
+        if args.cache_path:
+            cmdline.append("--bind=" + os.path.join(args.cache_path, "cargo") + ":/var/cache/cargo")
+            cmdline.append("--setenv=CARGO_HOME=/var/cache/cargo")
```
DaanDeMeyer, line 4575: Maybe just wrap the build script stuff in a `mount_cache` context? That way we don't have to add every cache in two places.

mkosi
```diff
@@ -2239,7 +2241,10 @@ def run_postinst_script(args: CommandLineArguments, workspace: str, do_run_build
         shutil.copy2(args.postinst_script,
                      os.path.join(workspace, "root", "root/postinst"))
 
-        run_workspace_command(args, workspace, "/root/postinst", verb, network=args.with_network)
+        env = {
+            "CARGO_HOME": "/var/cache/cargo",
+        }
```
DaanDeMeyer, line 2246: Why not simply mount the cache to root/.cargo? Then we don't need all the CARGO_HOME stuff.

mmorvai:
> If we're mounting the cache, we can go one step further and actually fill it in as well. I would gate all the cache stuff behind a check for a Cargo.toml file in the project source directory. If it exists, create the cache directory and run cargo fetch to fetch the dependencies. We can't do that in the build script because it doesn't run with network access.

I like the idea of doing this automatically for some package managers like cargo/pip/npm, but there are some things i am not sure about:
1. Should mkosi automatically take care of installing the package manager (and the language libs like rust/nodejs/python) as well? If so, which version? And from which source (distros packages or rustup/nvm/get-pip)?
2. Should we only fetch the packages? Or install them? In the build image, final image or both?

I tend to agree with @behrmann that this might be too specific. Instead we could add a parameter like
`--custom-cache-directories=mkosi.cargocache:/root/.cargo,mkosi.pipcache:/root/cache/.pip`
which could be bind mounted in every workspace call (mkosi.prepare,mkosi.postinst,mkosi.build) using a mount_cache context.

Additionally we could bind mount some folder read-only for mkosi.prepare to have requirements.txt/Cargo.toml/packages.json/.... As you suggested in #450, although i am not sure which directory? Just the project root? Or some special-files-for-prepare-script diectory?

This way the user could decide which version to use, and they can fetch the packages or install them or do whatever they like.

DaanDeMeyer:
> Should mkosi automatically take care of installing the package manager (and the language libs like rust/nodejs/python) as well? If so, which version? And from which source (distros packages or rustup/nvm/get-pip)?

Great question, I hadn't thought of this yet. This would also require dealing with how each supported distro handles rust support (or any other package manager)

> Should we only fetch the packages? Or install them? In the build image, final image or both?

So in terms of distro packages we fetch and install but the difference is that those packages are prebuilt. In the case of rust where they have to be compiled from source and users might want to play with the compiler options with which they compile crates, we definitely want to give that control to the users so we'd only fetch and leave the building to the build script so the user has full control. Generally, if the user has reason to influence how something is compiled, we shouldn't do it for them I think without giving access to all the necessary knobs which would put too much complexity in mkosi itself.

> Additionally we could bind mount some folder read-only for mkosi.prepare to have requirements.txt/Cargo.toml/packages.json/.... As you suggested in #450, although i am not sure which directory? Just the project root? Or some special-files-for-prepare-script diectory?

I'd just reuse SourceFileTransfer here. That seems like the easiest and safest solution since no the necessary files will always be available.

> I tend to agree with @behrmann that this might be too specific. Instead we could add a parameter like
--custom-cache-directories=mkosi.cargocache:/root/.cargo,mkosi.pipcache:/root/.cache/.pip
which could be bind mounted in every workspace call (mkosi.prepare,mkosi.postinst,mkosi.build) using a mount_cache context.

I agree we should have this regardless of whether we add built-in support for popular package managers.

In general, I think you're right, supporting rust directly is going to be a lot more work than I initially thought. Let's start with custom cache directories and making the project sources available to the prepare script, we can move from there.

DaanDeMeyer:

PR #402, opened 2020-02-20
Changes to support unified kernel+initrd in Ubuntu

fernandolobato:
I need to create an Ubuntu custom image with a read only file system (+ dm verity) and secure boot. I realized mkosi has all of this but the option for a unified kernel is disabled for Linux. I realized that the changes to get it to work are minimal, just differences in paths that can be fixed with usrmerge.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1802,7 +1802,7 @@ def install_debian_or_ubuntu(args: CommandLineArguments,
     # Debootstrap is not smart enough to deal correctly with alternative dependencies
     # Installing libpam-systemd via debootstrap results in systemd-shim being installed
     # Therefore, prefer to install via apt from inside the container
-    extra_packages = ['dbus', 'libpam-systemd']
+    extra_packages = ['dbus', 'libpam-systemd', 'binutils']
 
     # Also install extra packages via the secondary APT run, because it is smarter and
     # can deal better with any conflicts
@@ -2864,7 +2864,7 @@ def install_unified_kernel(args: CommandLineArguments,
     if do_run_build_script:
         return
 
-    if args.distribution not in (Distribution.fedora, Distribution.mageia):
+    if args.distribution not in (Distribution.fedora, Distribution.mageia, Distribution.ubuntu, Distribution.debian):
         return
 
     with complete_step("Generating combined kernel + initrd boot file"):
@@ -2889,6 +2889,7 @@ def install_unified_kernel(args: CommandLineArguments,
                       "-v",
                       "--no-hostonly",
                       "--uefi",
+                      "--uefi-stub", "/usr/lib/systemd/boot/efi/linuxx64.efi.stub",
                       "--kver", kver.name,
                       "--kernel-cmdline", cmdline]
 


mkosi
```diff
@@ -2864,19 +2864,21 @@ def install_unified_kernel(args: CommandLineArguments,
...
         cmdline = ' '.join(args.kernel_command_line)
         if root_hash is not None:
             cmdline += " roothash=" + root_hash
 
-        for kver in os.scandir(os.path.join(workspace, "root", "usr/lib/modules")):
+        for kver in os.scandir(os.path.join(workspace, "root", lib_path, "modules")):
             if not kver.is_dir():
                 continue
-
+
```
behrmann: There is spurious whitespace here

behrmann:
Is there a reason this shouldn't also work with Debian?

mkosi
fernandolobato: Fixed

fernandolobato:
> Is there a reason this shouldn't also work with Debian?


No, it actually works, just tested it. I have added Debian in the support list.

fernandolobato:
Hey @behrmann , are we good to merge this?

behrmann:
@fernandolobato Sorry, I can't merge anything. I just browse incoming PRs for fun and give comments if I think what I say might be of interest. Also, I haven't tested your PR. For merging @poettering is the one to ask.

mkosi
```diff
@@ -2864,16 +2864,18 @@ def install_unified_kernel(args: CommandLineArguments,
     if do_run_build_script:
         return
 
-    if args.distribution not in (Distribution.fedora, Distribution.mageia):
+    if args.distribution not in (Distribution.fedora, Distribution.mageia, Distribution.ubuntu, Distribution.debian):
         return
+    
+    lib_path = 'lib' if args.distribution in (Distribution.ubuntu, Distribution.debian) else 'usr/lib'
```
behrmann: On an unrelated note, is this strictly necessary? Current versions of Debian default to `merged-usr` and although Debian probably won't ever mandate this, I guess we can be opinionated here. At least on my machine I see `lib -> usr/lib` when doing `ls -l /`.

fernandolobato: For Debian, this is actually not needed since it has that sym link, on Ubuntu though it's needed.

behrmann: I quickly booted an Ubuntu 19.04 live usb system, that I had flying around, so this may only count as anecdotal evidence, but it does have the `lib -> usr/lib` symlink as well. I also googled around a bit and it seems merged-usr is the default since 19.04.

fernandolobato: Yes, this is right. I am doing this on Ubuntu 18.04 though, but would you suggest instead to create a symlink in my setup and just change this PR to enable the Ubuntu flag?

behrmann: Good question. Looking at `install_debian_or_ubuntu` debootstrap would already install the system as merged usr if it knew it. I guess you could convert the system to merged usr afterwards by installing the `usrmerge` package, that should be available in 18.04 as well (haven't checked that though). I don't really know what the best way forward is, but specialcasing Ubuntu for a two-year old LTS when the new one is around the corner doesn't seem optimal to me.

behrmann:
I gave this a quick spin and tried to generate a Debian buster image, but it failed because it needed `objdump` from the `binutils` package. Should this maybe be installed as well now?

mkosi
fernandolobato: Yeah, you are right, it's best to keep it simple. I will make changes based on this.

fernandolobato:
> objdump

I will also add this, thanks for all the feedback.

behrmann:
@fernandolobato Looks pretty good to me :)

mkosi
```diff
@@ -2889,6 +2889,7 @@ def install_unified_kernel(args: CommandLineArguments,
                       "-v",
                       "--no-hostonly",
                       "--uefi",
+                      "--uefi-stub", "/usr/lib/systemd/boot/efi/linuxx64.efi.stub",
```
behrmann: The PR is already quite minimal at this point, but might it be whittled down a bit further? 

I just looked at the the dracut man page and the default search path for `uefi_stub` is `/lib/systemd`. Since on a Debian with merged usr `/lib` is symlink to `/usr/lib` this line shouldn't be strictly necessary, right?

PR #406, opened 2020-03-11
Add Github Actions for CI work

killermoehre:
Hi,

I created a few simple workflows testing with github actions if a simple image can be build. Right now it features Debian, Ubuntu, ClearOS, Arch and CentOS (Arch and CentOS seems to be broken from the mkosi side due to changed packages/repositories).

I did this mostly to play around with github actions, but also to broaden the CI scope.
HTH


lucasdemarchi:
Looks great! Thanks!

PR #409, opened 2020-03-23
debootstrap fixes with https and upgrading

lahtima:
I needed to use https-mirror for Ubuntu and wrote some workarounds for debootstrap problems.
I also made a new flag to allow installing security upgrades directly after debootstrap if apt is configured correctly. This is mainly because Ubuntu seems to split newest updates to separate repositories instead of the main repository debootstrap uses.

I haven't tested if debian has similar problems with debootstrap and https.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1899,6 +1899,14 @@ def install_debian_or_ubuntu(args: CommandLineArguments,
         cmdline = ["/usr/bin/apt-get", "--assume-yes", "update"]
         run_workspace_command(args, workspace, network=True, env=env, *cmdline)
 
+    if args.upgrade:
+        # Assumes correct configuration in /etc/apt/sources.list.d/ at skeleton
+        # folder as debootstrap overrides /etc/apt/sources.list provided
+        cmdline = ["/usr/bin/apt-get", "--assume-yes", "update"]
+        run_workspace_command(args, workspace, network=True, env=env, *cmdline)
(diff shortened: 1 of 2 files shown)


mkosi
```diff
@@ -1785,6 +1785,13 @@ def install_debian_or_ubuntu(args: CommandLineArguments,
                "--exclude=sysv-rc,initscripts,startpar,lsb-base,insserv",
                "--components=" + ','.join(repos)]
 
+    # Atleast ubuntu doesn't have apt-transport-https anymore, except on xenial,
+    # but debootstrap scripting tries to install it.
+    # Not checking end-of-life releases
+    # Correct patch should be in debootstrap scripting
+    if args.distribution == Distribution.ubuntu and args.release not in ["xenial"]:
+        cmdline[4] += ",apt-transport-https"
+
```
lucasdemarchi: Sounds like a fix to debootstrap. Have you created a bug report?

lahtima: There has been bug report open for debootstrap since 2017 [debian bug 879755].
Now that I rechecked the status, there finally seems to be fix in debootstrap upstream for this since 2020-03-07 but this kind of workaround probably will be needed for LTS distributions for a while like current Ubuntu LTS and possibly next one, have to check if it's fixed there also and comment on the bug report already made for Ubuntu bug tracking too.

DaanDeMeyer:
The --update switch is too specific too Ubuntu/Debian and we prefer to not workaround tools in mkosi anymore so I'm going to close this PR. You can do implement the update switch with a mkosi.prepare script instead.

Issue #410, opened 2020-03-30
Ubuntu: failure at install_unified_kernel from 531f52d


alessandrogambino:
Hello,

Could be already known or work-in-progress as it is quite recent, by the way commit 531f52d4815cd4465ce3be82835ea0edbf005272 introduce some regression on Ubuntu.
Command and stacktrace in attachment. Previous commit (22dc19b4cc9f5bcfee3c45426a579755cf77bef) conclude the command successfully.

[stack.txt](https://github.com/systemd/mkosi/files/4403885/stack.txt)

Regards

fernandolobato:
For ubuntu 18.04 you will need to add usrmerge to the list of packages. Maybe this should be installed by default. I will send another PR.

fernandolobato:
The fix has been merge, closing this issue.

fernandolobato:
@alessandrogambino , The fix can be merge, you can now check.

DaanDeMeyer:
Let's close this one as a fix was merged.

PR #411, opened 2020-04-02
Add usrmerge for ubuntu previous to 18.04

fernandolobato:
This change is a follow up to this PR:

https://github.com/systemd/mkosi/pull/402  - This introduced the unified kernel EFI for debian and ubuntu distros. 

Looking at this [issue](https://github.com/systemd/mkosi/issues/410), ubuntu 18.04 and below require usrmerge for those paths to work. The PR originally special cased 18.04, but I moved away from it and forgot to add usrmerge.


behrmann:
The build failures are unrelated to this, so they shouldn't stand in the way of merging this.

PR #412, opened 2020-04-04
Fix ci-arch.yml to support zstd-compressed packages

johnp789:
This installs libzstd1-dev as an additional package, then downloads and builds libarchive, so that the resulting library supports zstd.

### DIFF
--- a/.github/workflows/ci-arch.yml
+++ b/.github/workflows/ci-arch.yml
@@ -23,7 +23,18 @@ jobs:
     - name: Update package cache
       run: sudo apt-get --assume-yes update
     - name: Install dependencies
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
Wouldn't it be possible to bootstrap a minimal arch (maybe just pacman) inside the container and then use that one to install the rest of the system? I think this would be nice so it doesn't depend much on the host installing it.

johnp789:
I suppose so, but that would take mkosi beyond being a wrapper around pacstrap.  The goal of this PR is simply to get the CI setup working for Arch, now that Arch uses zstd-compressed package files.

behrmann:
@lucasdemarchi: Do I understand you correctly, that you mean to pacstrap an Arch and use that for running mkosi in the tests?

lucasdemarchi:
I actually think this is good. Just need a squash

PR #420, opened 2020-05-14
Arch: setup locale earlier to silence pacman perl warnings

DaanDeMeyer:
On arch, installing perl via pacman warns if no locale is set. By setting up the locale a bit earlier, we avoid this warning.


DaanDeMeyer:
Nevermind, perl already gets pulled in in the initial pacstrap so we can't generate the locale beforehand.

PR #421, opened 2020-05-17
Arch: Improve perf + fix pacman warnings/errors

DaanDeMeyer:
Saves about 45s on my machine with systemd's mkosi.arch config. Also includes a general change to not install a boot loader on build script machines.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1220,7 +1220,7 @@ def prepare_tree(args: CommandLineArguments, workspace: str, do_run_build_script
     if cached:
         return
 
-    if args.bootable:
+    if not do_run_build_script and args.bootable:
         # We need an initialized machine ID for the boot logic to work
         os.mkdir(os.path.join(workspace, "root", "etc"), 0o755)
         with open(os.path.join(workspace, "root", "etc/machine-id"), "w") as f:
--- a/mkosi
+++ b/mkosi
@@ -2136,7 +2136,21 @@ Server = {repository_server}
                 else:
                     return 'HOOKS="systemd modconf block filesystems fsck"\n'
             return line
-        patch_file(os.path.join(workspace, "root", "etc/mkinitcpio.conf"), jj)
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
I've been using these changes for quite a while now without any issues. Any chance they could get a review @lucasdemarchi @keszybz? (If there are any Arch mkosi experts I should tag instead, let me know)

mkosi
```diff
@@ -2023,6 +2023,25 @@ def install_arch(args: CommandLineArguments, workspace: str, do_run_build_script
...
+    fix_permissions_dirs = [
+        ["boot", 0o755],
+        ["etc", 0o755],
+        ["etc/pacman.d", 0o755],
+        ["var", 0o755],
+        ["var/lib", 0o755],
+        ["var/cache", 0o755],
+        ["var/cache/pacman", 0o755],
+        ["var/tmp", 0o1777],
+    ]
```
keszybz, line 2038: I'd use tuples for this, but of course it doesn't matter.

mkosi
```diff
@@ -2023,6 +2023,31 @@ def install_arch(args: CommandLineArguments, workspace: str, do_run_build_script
...
+    for dir, permissions in fix_permissions_dirs:
+        path = os.path.join(root, dir)
+        if os.path.exists(path):
+            os.chmod(path, permissions)
+
+    # Disable mkinitcpio hook because we manually install the kernel image to
+    # /boot using kernel-install in install_boot_loader_arch.
+    hooks = os.path.join(root, "etc/pacman.d/hooks")
+    os.makedirs(hooks, 0o755, exist_ok=True)
+    os.symlink("/dev/null", os.path.join(hooks, "90-mkinitcpio-install.hook"))
```
lucasdemarchi, line 2049: don't you want to undo this when you are done? Otherwise the image produced is not so useful as this hook will remain disabled (see similar issue solved for fedora

DaanDeMeyer: Good point, I'll make a new PR that removes this after pacman has finished.

PR #424, opened 2020-05-22
Arch: Fix root login when using systemd-nspawn

DaanDeMeyer:


DaanDeMeyer:
Updated to simply remove the pam_securetty line from /etc/pam.d/login instead of adding the pts/n lines to /etc/securetty.

DaanDeMeyer:
@lucasdemarchi Anything else to be done to get this merged? I've been using this change for a while and its been working flawlessly.

PR #425, opened 2020-05-23
Install the boot loader after copying in the build tree.

DaanDeMeyer:
This makes sure the binaries from the build tree are bundled in the
initramfs as well (which is mostly useful for systemd).


DaanDeMeyer:
This doesn't play well with caching since the boot loader installation is cached. However, disabling boot loader installation caching makes incremental builds a lot slower. Maybe put this behind an option so that initramfs stuff can be effectively debugged without slowing down normal development?

DaanDeMeyer:
Closing this for now until I find a better solution.

PR #427, opened 2020-05-29
Clean the tab idents

potens1:
Hi,
This is only (and only that) to clean the tab indents. I know it's not high priority or anything, but sooner or later it will be problematic (either not supported for mixing both, either introducing bugs in control statements) so, let's do it.
I changed only that "on purpose" to have a very clean commit about it (no other pep8 corrections, no realignments and so on), so I guess it can be auto merged in other pull requests.
I also changed the editorconfig file so, if/when the mkosi file is renamed to mkosi.py, all editors understanding the editorconfig file will auto-use spaces for indentation.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1467,14 +1467,14 @@ def clean_package_manager_metadata(workspace: str) -> None:
     # FIXME: implement cleanup for other package managers
 
 def clean_tdnf_metadata(root: str) -> None:
-        """Removes tdnf metadata iff /bin/tdnf is not present in the image"""
-        tdnf_path = root + '/usr/bin/tdnf'
-        keep_tdnf_data = os.access(tdnf_path, os.F_OK, follow_symlinks=False)
+    """Removes tdnf metadata iff /bin/tdnf is not present in the image"""
+    tdnf_path = root + '/usr/bin/tdnf'
+    keep_tdnf_data = os.access(tdnf_path, os.F_OK, follow_symlinks=False)
 
-        if not keep_tdnf_data:
-            print_step('Cleaning tdnf metadata...')
-            remove_glob(root + '/var/log/tdnf.*',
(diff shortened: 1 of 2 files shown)


mkosi
```diff
@@ -1467,14 +1467,14 @@ def clean_package_manager_metadata(workspace: str) -> None:
     # FIXME: implement cleanup for other package managers
 
 def clean_tdnf_metadata(root: str) -> None:
-	    """Removes tdnf metadata iff /bin/tdnf is not present in the image"""
-	    tdnf_path = root + '/usr/bin/tdnf'
-	    keep_tdnf_data = os.access(tdnf_path, os.F_OK, follow_symlinks=False)
+        """Removes tdnf metadata iff /bin/tdnf is not present in the image"""
+        tdnf_path = root + '/usr/bin/tdnf'
+        keep_tdnf_data = os.access(tdnf_path, os.F_OK, follow_symlinks=False)
```
keszybz, line 1472: That's not 4 spaces, is it?

potens1: No, it  is not, the original was neither one tab, neither 4 spaces but tab and 4 spaces.
I changed tabs without realigning (on purpose), but is you prefer (and I prefer also that, by the way) I suppress all redundant idents, let me know, I'll manage it before it to be merged (I have not done it since I did not know the feeling about the indentation problem), I've done that so it was pretty clear only tabs where changed.

keszybz:
```
Error: Cannot retrieve repository metadata (repomd.xml) for repository: base. Please verify its path and try again
YumRepo Error: All mirror URLs are not using ftp, http[s] or file.
 Eg. Invalid release/repo/arch combination/
removing mirrorlist with no valid mirrors: /var/tmp/mkosi-kkrbcjlc/root/var/cache/yum/base/mirrorlist.txt
```
looks unrelated.

PR #429, opened 2020-06-01
Factor securetty disabling and apply to debian

potens1:
Following the  #424 merge, the same can be applied to debian since it does still use pam_securetty with old set of permissions.

I don't know if other distribs use it but the function call can be added to any required.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1283,6 +1283,17 @@ def patch_file(filepath: str, line_rewriter: Callable[[str], str]) -> None:
     shutil.move(temp_new_filepath, filepath)
 
 
+def disable_pam_securetty(workspace: str,
+                          file_path: str = 'root/etc/pam.d/login') -> None:
+    patch_file(os.path.join(workspace, file_path), _rm_securetty)
+
+
+def _rm_securetty(line: str) -> str:
+    if 'pam_securetty.so' in line:
+        return ''
+    return line
+
+
 def enable_networkd(workspace: str) -> None:
     run(["systemctl",
          "--root", os.path.join(workspace, "root"),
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -1283,6 +1283,17 @@ def patch_file(filepath: str, line_rewriter: Callable[[str], str]) -> None:
     shutil.move(temp_new_filepath, filepath)
 
 
+def disable_pam_securetty(workspace: str,
+                          file_path: str = 'root/etc/pam.d/login') -> None:
+    patch_file(os.path.join(workspace, file_path), _rm_securetty)
+
+
+def _rm_securetty(line: str) -> str:
```
DaanDeMeyer, line 1291: Maybe make this function local to `disable_pam_securetty`?

mkosi
```diff
@@ -1283,6 +1283,17 @@ def patch_file(filepath: str, line_rewriter: Callable[[str], str]) -> None:
     shutil.move(temp_new_filepath, filepath)
 
 
+def disable_pam_securetty(workspace: str,
+                          file_path: str = 'root/etc/pam.d/login') -> None:
```
DaanDeMeyer, line 1287: No reason to make the path configurable until we actually have a usecase where we need a different path. I'd just remove the parameter until we actually need it.

mkosi
potens1: Done

mkosi
potens1: I would answer "no reason to remove it until we are sure all distributions have to same conf file for that", but I trust you about it, so, changed also.

poettering:
I wish arch and Debian would just get their act together and remove this properly. It's 2020... Kinda sucks having to carry this compat kludge.

@DaanDeMeyer you should have all the privs now

PR #432, opened 2020-06-12
Call finalize script as described in documentation

### LINKED ISSUE #310: mkosi.finalize hookup appears borked
run_finalize_script() is currently called with verb="build" after the incremental build image is created and then a second time with verb="final" after the development image is created. This looks wrong. I am sure it should not be called during the incremental image generation at all, but instead only in the later real development/final image generation.

@keszybz could you look into this plz?

(Note that build_image() is called four times currently: 1. to generate incremental devel image, 2. to generate incremental final image. 3. to generate real devel image off incremental one, 4. to generate real final image off incremental one, and afaics the hookups are done between the wrong stages of this)

mmorvai:
Finalize script is now called as last step before build script in first
stage and as last step before image is made read-only/generated image is
made in second stage.

Finalize script is not called when creating cached/incremental images.

This should fix #310

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -2500,9 +2500,13 @@ def run_postinst_script(args: CommandLineArguments, root: str, do_run_build_scri
         os.unlink(os.path.join(root, "root/postinst"))
 
 
-def run_finalize_script(args: CommandLineArguments, root: str, *, verb: str) -> None:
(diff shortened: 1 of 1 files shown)


dellgreen:
looking forward to testing this, as I require a working finalize phase :)

DaanDeMeyer:
LGTM, I rebased this on master so it can run with all the CI fixes.

PR #413, opened 2020-04-11
Fix packages and build packages not being installed on Centos and Mageia.

liam-mcb:
Fix build packages not being installed for Centos builds.
Fix packages and build packages not being installed for Mageia builds.

PR #434, opened 2020-06-12
Implement bootable centOS 8 images.

JoseTeuttli:
This was tested on Fedora 32 x86, and the image was tested using both systemd-nspawn and qemu.

The commit history is very messy and requires a squash merge.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1953,17 +1953,26 @@ def install_centos(args: CommandLineArguments, root: str, do_run_build_script: b
 
     packages = ['centos-release']
     packages += args.packages or []
+
     if args.bootable:
-        packages += ["kernel", "systemd-udev", "binutils"]
+        packages += ["kernel", "systemd", "binutils"]
+
     if do_run_build_script:
         packages += args.build_packages or []
 
     repos = args.repositories or default_repos
+
     if args.distribution == Distribution.centos_epel:
         repos += ["epel"]
         packages += ["epel-release"]
 
-    invoke_dnf_or_yum(args, root, repos, packages, config_file)
+    if do_run_build_script:
+        packages += args.build_packages or []
+
+    invoke_dnf_or_yum(args, root,
+                      repos,
+                      packages,
+                      config_file)
 
     reenable_kernel_install(args, root, masked)
 
@@ -2583,6 +2592,8 @@ def install_grub(args: CommandLineArguments, root: str, loopdev: str, grub: str)
 def install_boot_loader_fedora(args: CommandLineArguments, root: str, loopdev: str) -> None:
     install_grub(args, root, loopdev, "grub2")
 
+def install_boot_loader_centos(args:CommandLineArguments, root: str, loopdev: str) -> None:
+    install_grub(args, root, loopdev, "grub2")
 
 def install_boot_loader_arch(args: CommandLineArguments, root: str, loopdev: str) -> None:
     if "uefi" in args.boot_protocols:
@@ -2675,6 +2686,9 @@ def install_boot_loader(args: CommandLineArguments, root: str, loopdev: Optional
 
         if args.distribution == Distribution.photon:
             install_boot_loader_photon(args, root, loopdev)
+
+        if args.distribution == Distribution.centos:
+            install_boot_loader_centos(args, root, loopdev)
 
 def install_extra_trees(args: CommandLineArguments, root: str, for_cache: bool) -> None:
     if not args.extra_trees:
@@ -3134,8 +3148,7 @@ def install_unified_kernel(args: CommandLineArguments,
     # actually invoke the boot loader on the development image.
     if do_run_build_script:
         return
-
-    if args.distribution not in (Distribution.fedora, Distribution.mageia, Distribution.ubuntu, Distribution.debian):
+    if args.distribution not in (Distribution.fedora, Distribution.mageia, Distribution.ubuntu, Distribution.debian, Distribution.centos, Distribution.centos_epel):
         return
 
     with complete_step("Generating combined kernel + initrd boot file"):


mkosi
```diff
@@ -2493,7 +2545,8 @@ def install_grub(args: CommandLineArguments, workspace: str, loopdev: str, grub:
 
 def install_boot_loader_fedora(args: CommandLineArguments, workspace: str, loopdev: str) -> None:
     install_grub(args, workspace, loopdev, "grub2")
-
+def install_boot_loader_centos(args:CommandLineArguments, workspace: str, loopdev: str) -> None:
+    install_grub(args,workspace,loopdev,"grub2")
```
poettering, line 2549: please follow coding style, i.e.  spaces after comma...

But actually, given that this is the same function as  install_boot_loader_fedora() maybe just rename that one to install_boot_loader_and_centos()

liam-mcb:
I have a similar pull request #413, sans it making CentOS bootable.

mkosi
JoseTeuttli: I think I managed to fix the messy commit history, I also made changes to try to align my changes with the coding style. I plan to make CentOS specific changes to the bootloader method so I think keeping them separate is best.

DaanDeMeyer:
Can you rebase this on master? CentOS 8 support was just added by another PR. I think this PR can be reduced to only the changes to make CentOS 8 bootable.

mkosi
```diff
@@ -2307,7 +2360,11 @@ default_image="/boot/initramfs-{kernel}.img"
 
     # Arch still uses pam_securetty which prevents root login into
     # systemd-nspawn containers. See https://bugs.archlinux.org/task/45903.
-    disable_pam_securetty(root)
```
DaanDeMeyer, line 2310: Seems like a merge went wrong here. `disable_pam_securetty` should be kept instead.

mkosi
```diff
@@ -2674,7 +2733,10 @@ def install_boot_loader(args: CommandLineArguments, root: str, loopdev: Optional
             install_boot_loader_clear(args, root, loopdev)
 
         if args.distribution == Distribution.photon:
-            install_boot_loader_photon(args, root, loopdev)
```
DaanDeMeyer, line 2677: Same here, we now pass around `root` instead of `workspace` so the above statement was the right one.

mkosi
```diff
@@ -2583,6 +2640,8 @@ def install_grub(args: CommandLineArguments, root: str, loopdev: str, grub: str)
 def install_boot_loader_fedora(args: CommandLineArguments, root: str, loopdev: str) -> None:
     install_grub(args, root, loopdev, "grub2")
 
+def install_boot_loader_centos(args:CommandLineArguments, workspace: str, loopdev: str) -> None:
```
DaanDeMeyer, line 2643: Replace `workspace` by `root` here.

mkosi
JoseTeuttli: Fixed, thanks for catching this.

mkosi
JoseTeuttli: Fixed, thanks for catching this.

mkosi
JoseTeuttli: Fixed, thanks for catching this. Do you think a rebase is necessary to reduce the changes to 1 commit? Originally when I submitted the push request it was incredibly messy, and I am unsure on the norms of PR's

DaanDeMeyer: So with systemd and mkosi, we generally go for "perfect" commits when merging PRs. The idea is that each commit stands completely on its own and works perfectly fine on its own (it builds correctly, all tests pass, no regressions, ...). That makes it easy to run git bisect later and generally keeps the git history clean.

So yes, a rebase that merges everything in one commit would be great.

JoseTeuttli: I think I did it, but would appreciate if you could look it over again to make sure I didn't mess anything up in the process. This has been a learning experience and I appreciate the patience.

DaanDeMeyer: Seems like `workspace` still needs to be updated to `root` here.

DaanDeMeyer:
All the mirror/gpg stuff for centos 8 was already added by a previous PR. Your current PR still contains similar changes which I'm pretty sure can all be removed now.

JoseTeuttli:
> All the mirror/gpg stuff for centos 8 was already added by a previous PR. Your current PR still contains similar changes which I'm pretty sure can all be removed now.

I am not entirely sure how I changed the install_centos_old method, but it should be fixed now.

mkosi
JoseTeuttli: Fixed, thanks for catching this.

mkosi
```diff
@@ -1953,17 +1942,26 @@ def install_centos(args: CommandLineArguments, root: str, do_run_build_script: b
 
     packages = ['centos-release']
     packages += args.packages or []
-    if args.bootable:
-        packages += ["kernel", "systemd-udev", "binutils"]
     if do_run_build_script:
         packages += args.build_packages or []
+    if args.release == "8":
+        if args.bootable:
+            packages += ["kernel", "systemd-udev", "binutils"]
```
DaanDeMeyer, line 1949: Why pull in `systemd-udev` here? Don't we need the entire `systemd` package here as well? And is `binutils` strictly necessary for a bootable image?

mkosi
```diff
@@ -1953,17 +1942,26 @@ def install_centos(args: CommandLineArguments, root: str, do_run_build_script: b
...
+    else:
+        if args.bootable:
+            packages += ["kernel", "systemd", "binutils"]
 
-    repos = args.repositories or default_repos
     if args.distribution == Distribution.centos_epel:
-        repos += ["epel"]
+        default_repos += ["epel"]
         packages += ["epel-release"]
```
DaanDeMeyer, line 1956: I think this can be reverted as well.

mkosi
```diff
@@ -1836,35 +1836,29 @@ def install_centos_old(args: CommandLineArguments,
...
 [updates]
 name=CentOS-{args.release} - Updates
 {updates_url}
 gpgkey={gpg_key}
-
 [extras]
 name=CentOS-{args.release} - Extras
 {extras_url}
 gpgkey={gpg_key}
-
```
DaanDeMeyer, line 1854: Let's keep the newlines here and below intact. We use the same style elsewhere as well so no point in removing it here.

mkosi
DaanDeMeyer: Actually, now that I look again, it seems the previous PR already added code for the bootable option. Can you check if that works correctly already for both centos 7 and centos 8? If so, these changes can be reverted as well

mkosi
JoseTeuttli: This was reverted, I also changed the name of the variable that the return of install_centos_ is put into too match

mkosi
JoseTeuttli: I just tested, the reason I originally separated them out because centOS 7 does not have systemd-udev, so it needs systemd, however I never implemented bootable support for centOS 7, and it currently is not working on this PR. However you can just resolve these lines into what the original PR had, and so I reverted it since it is cleaner.

mkosi
JoseTeuttli: Sounds good, changed back to how it was

JoseTeuttli:
Rebased to 1 commit

mkosi
JoseTeuttli: Nevermind during the merge I had removed the line "repos = args.repositories or default_repos" accidentally which was causing me confusion, it has been reverted to the way it was originally.

mkosi
```diff
@@ -1953,17 +1953,26 @@ def install_centos(args: CommandLineArguments, root: str, do_run_build_script: b
 
     packages = ['centos-release']
     packages += args.packages or []
+
     if args.bootable:
-        packages += ["kernel", "systemd-udev", "binutils"]
+        packages += ["kernel", "systemd", "binutils"]
```
DaanDeMeyer, line 1958: We should probably install systemd by default to align with the other distros. I know at least Arch and ubuntu install systemd by default. We want to unify this so that mkosi boot always works regardless if --bootable or not is specified (--bootable applies more to being able to boot via qemu).

Does centos-release not install systemd by default?

DaanDeMeyer:
LGTM, Thanks for working on this!

Issue #130, opened 2017-07-24
mkosi should be made compatible with a host with SELinux enabled


sourcejedi:
[RHBZ#1374427](https://bugzilla.redhat.com/show_bug.cgi?id=1374427) equally affects `mkosi`.  `mkosi` was written, in part, for testing modifications to systemd.  But this issue means systemd's HACKING instructions do not work, on a default Fedora install.

`mkosi` does not request installation of selinux support in images, and [deliberately disables it](https://github.com/systemd/mkosi/blob/40df6a3acbcdb26d40471109fa2a53c15dd113e3/mkosi#L2494).  This seems correct when using `mkosi` to develop systemd or any other software.  Any modification to systemd could fail, until the downstream distribution updates their SELinux policy to support it.

However `dnf` still tries to assign SELinux labels to the installed files.  When running the image the labels cause SELinux transitions, e.g. when running `passwd`.  We don't want this, and it causes failures.  This behaviour of `dnf --installroot` under SELinux was included in the above report.  Sadly this did not elicit any suggestion of how to use `dnf --installroot` and get _consistent_ SELinux labels, let alone disable the labelling as `mkosi` would want.

When a user creates a Fedora image using `dnf`, they can very simply work around the issue using `restorecon` or possibly `fixfiles`.  The `mkosi` tool could do something similar.  The files in the image could be relabelled using `chcon unconfined_u:object_r:unlabeled_t:s0 -R`.  `unlabeled_t` is the value shown for files which have no context xattr.  This might be desirable, to avoid unnecessary differences between images generated on systems with SELinux enabled v.s. disabled.  (Or it could use `default_t`, the label assigned when you create a new directory underneath `/`).

Assigning labels is only possible on the system which has SELinux configured.  Therefore this needs to be done conditionally, and not from inside a container.  Without SELinux (e.g inside a container):

```
-bash-4.4# chcon unconfined_u:object_r:unlabeled_t:s0 / 
chcon: invalid context: 'unconfined_u:object_r:unlabeled_t:s0': No such file or directory
```

In theory, at least the `unconfined_u:object_r:` part is policy-specific.  As is the ability for "unconfined" processes to access `unlabeled_t` files without interference.  So this is not a universal solution for any SELinux policy.  However it should make mkosi usable on popular Linux systems and the SELinux reference policy.

## Specific failure

```
$ rpm -q mkosi
mkosi-2-1.fc26.noarch
```

1. `mkosi`
[...]

PR #437, opened 2020-06-18
Add option to strip SELinux file context from tar'ed files

ValdikSS:
This option removes (or not includes) SELinux xattrs into output tar archive.
This is a hack as mkosi does not properly support SELinux contexts. See #130.

The patch adds `--tar-strip-selinux-context` flag and `TarStripSELinuxContext` configuration file option.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3293,7 +3293,10 @@ def make_tar(args: CommandLineArguments, root: str, do_run_build_script: bool, f
         if shutil.which("bsdtar") and args.distribution == Distribution.openmandriva:
             _tar_cmd = ["bsdtar", "-C", root, "-c", "-J", "--xattrs", "-f", "-", "."]
         else:
-            _tar_cmd = ["tar", "-C", root, "-c", "-J", "--xattrs", "--xattrs-include=*", "."]
+            _tar_cmd = ["tar", "-C", root, "-c", "-J", "--xattrs", "--xattrs-include=*"]
+            if args.tar_strip_selinux_context:
+                _tar_cmd.append("--xattrs-exclude=security.selinux")
+            _tar_cmd.append(".")
 
         run(_tar_cmd, env={"XZ_OPT": "-T0"}, stdout=f)
 
@@ -4120,6 +4123,7 @@ class ArgumentParserMkosi(argparse.ArgumentParser):
         "BuildPackages": "--build-package",
         "PostInstallationScript": "--postinst-script",
         "GPTFirstLBA": "--gpt-first-lba",
+        "TarStripSELinuxContext": "--tar-strip-selinux-context",
     }
 
     fromfile_prefix_chars = "@"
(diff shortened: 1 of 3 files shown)


DaanDeMeyer:
Needs rebase, docs, and an addition to the tests config parser to make tests work. Aside from that, I don't see a big problem with adding this.

ValdikSS:
`bsdtar` does not support this option. Could you give me a hint where `--tar-strip-selinux-context` and `bsdtar` incompatibility should be handled?

behrmann:
`bsdtar` is only used for OpenMandriva, so it would make sense to fall back to regular tar in that case, since even for OpenMandriva it is only used if `bsdtar` is installed (which it would be by default, I guess)

ValdikSS:
Please check the current version.
It does not check `bsdtar` incompatibility.
Unfortunately I have no experience with test systems, kindly asking you to implement it yourself or give me a hint how to do that.

DaanDeMeyer:
Let's make this easier and simply not support this for openmandriva. Just add a check to `load_args` that exits early if distro is openmandriva and this option is set.

ValdikSS:
I'm not sure I understand the following line:
https://github.com/systemd/mkosi/blob/master/mkosi/__init__.py#L3292-L3293

`args.distribution` is a distribution name which `mkosi` is building, not the host distribution on which is executes. As far as I undrestand, `tar` is run on the host machine (not inside the container which has been built). Is there any reason to prefer `bsdtar` of the host machine when building OpenMandriva, or is that a mistake?
The [initial commit](https://github.com/systemd/mkosi/commit/4e077d8fb9cd503142255cab28406239157a4b34) of OpenMandriva support also doesn't comment on this.

DaanDeMeyer:
Yes, we could improve this but the current approach works as I don't think anyone is even building openmandriva on a host that's not itself openmandriva. You can just extract `shutil.which("bsdtar") and args.distribution == Distribution.openmandriva` into a function and check against that function in `load_args`. If it returns true and the new option is set, simply call `die` and tell the user that this option is not supported when using bsdtar.

ValdikSS:
Updated, please check.

DaanDeMeyer:
Needs a rebase

mkosi/__init__.py
```diff
@@ -3293,7 +3293,10 @@ def make_tar(args: CommandLineArguments, root: str, do_run_build_script: bool, f
         if shutil.which("bsdtar") and args.distribution == Distribution.openmandriva:
             _tar_cmd = ["bsdtar", "-C", root, "-c", "-J", "--xattrs", "-f", "-", "."]
         else:
-            _tar_cmd = ["tar", "-C", root, "-c", "-J", "--xattrs", "--xattrs-include=*", "."]
+            _tar_cmd = ["tar", "-C", root, "-c", "-J", "--xattrs", "--xattrs-include=*"]
+            if args.tar_strip_selinux_context:
+                _tar_cmd.append('--xattrs-exclude=security.selinux')
+            _tar_cmd.append('.')
```
DaanDeMeyer, line 3299: Please run black to ensure consistent formatting (or manually fix this to use double quotes).

ValdikSS:
Done, done.

DaanDeMeyer:
Still needs test additions, see test_config_parser.py. Seach for the network_veth option there and do the same changes for this new option

ValdikSS:
Updated. Is that sufficient?
P.S. `--no-chown` also missing in tests.

DaanDeMeyer:
We can fix that later, as long as tests pass we're good

DaanDeMeyer:
LGTM! Thanks for the contribution. Sorry for taking so long in reviewing this.

PR #438, opened 2020-06-23
mkosi: fix `-t submodule` output

### LINKED ISSUE #396: btrfs builds fail due to already existing path
Hello

```bash
mkosi -t subvolume -o debian-test -r testing -d debian
‣ Processing default...
[...]
‣ Detaching namespace...
‣ Detaching namespace complete.
‣ Setting up package cache...
‣ Setting up package cache /var/lib/machines/.mkosi-hvtzmez6 complete.
‣ Setting up temporary workspace.
‣ Temporary workspace in /var/lib/machines/.mkosi-t0yppp_i is now set up.
‣ Running second (final) stage...
‣ Mounting image...
‣ Mounting image complete.
‣ Setting up basic OS tree...
ERROR: target path already exists: /var/lib/machines/.mkosi-t0yppp_i/root
‣ Unmounting image...
‣ Unmounting image complete.
  File "mkosi", line 4872, in <module>
    main()
  File "mkosi", line 4868, in main
    run_verb(a)
  File "mkosi", line 4843, in run_verb
    build_stuff(args)
  File "mkosi", line 4667, in build_stuff
    raw, tar, root_hash = build_image(args, workspace, do_run_build_script=False, cleanup=True)
  File "mkosi", line 4471, in build_image
    prepare_tree(args, workspace.name, do_run_build_script, cached)
  File "/usr/lib/python3.7/contextlib.py", line 74, in inner
    return func(*args, **kwds)
  File "mkosi", line 1083, in prepare_tree
    btrfs_subvol_create(os.path.join(workspace, "root"))
  File "mkosi", line 467, in btrfs_subvol_create
    run(["btrfs", "subvol", "create", path], check=True)
  File "mkosi", line 76, in run
    return subprocess.run(cmdline, **kwargs)
  File "/usr/lib/python3.7/subprocess.py", line 512, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['btrfs', 'subvol', 'create', '/var/lib/machines/.mkosi-t0yppp_i/root']' returned non-zero exit status 1.
```

I've tracked down this issue to this [commit](https://github.com/systemd/mkosi/commit/0a9928d5506fa94d866da4be64dc1bcf5f9d97d0#diff-6980c629516e4165bfaf96be8e5003cf).
@lucasdemarchi I'm not sure about the proper fix to bring. I've checked that adding back the condition fixes the issue, however if I understand properly your commit the idea is to mount something on `root` so that the later `umount` cleans out all things that may have been mounted on top this tree. Any suggestion about the proper way to fix this ?

intelfx:
By the time we get to creating the root submodule, the directory already exists
because we self-bind-mount it in mount_image(), and it is too late to convert it
because of the same.

Thus, move the root subvolume creation into its own step just prior to
mount_image().

Fixes #396.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1202,13 +1202,14 @@ def umount(where: str) -> None:
     run(["umount", "--recursive", "-n", where], stdout=DEVNULL, stderr=DEVNULL)
 
 
-@completestep('Setting up basic OS tree')
-def prepare_tree(args: CommandLineArguments, workspace: str, do_run_build_script: bool, cached: bool) -> None:
+@completestep('Setting up OS tree root')
+def prepare_tree_root(args: CommandLineArguments, workspace: str) -> None:
     if args.output_format == OutputFormat.subvolume:
         btrfs_subvol_create(os.path.join(workspace, "root"))
-    else:
-        mkdir_last(os.path.join(workspace, "root"), 0o755)
 
+
+@completestep('Setting up basic OS tree')
+def prepare_tree(args: CommandLineArguments, workspace: str, do_run_build_script: bool, cached: bool) -> None:
     if args.output_format is OutputFormat.subvolume or \
        (args.output_format is OutputFormat.gpt_btrfs and not (args.minimize or cached)):
         btrfs_subvol_create(os.path.join(workspace, "root", "home"))
@@ -4688,6 +4689,7 @@ def build_image(args: CommandLineArguments,
 
             # Mount everything together, but let's not mount the root
             # dir if we still have to generate the root image here
+            prepare_tree_root(args, workspace.name)
             with mount_image(args, workspace.name, loopdev, None if args.generated_root() else encrypted_root,
                              encrypted_home, encrypted_srv, encrypted_var, encrypted_tmp):
                 prepare_tree(args, workspace.name, do_run_build_script, cached)


mkosi
```diff
@@ -1202,13 +1202,14 @@ def umount(where: str) -> None:
...
 
 
-@completestep('Setting up basic OS tree')
-def prepare_tree(args: CommandLineArguments, workspace: str, do_run_build_script: bool, cached: bool) -> None:
+@completestep('Setting up OS tree root')
+def prepare_tree_root(args: CommandLineArguments, workspace: str) -> None:
     if args.output_format == OutputFormat.subvolume:
         btrfs_subvol_create(os.path.join(workspace, "root"))
-    else:
-        mkdir_last(os.path.join(workspace, "root"), 0o755)
```
DaanDeMeyer, line 1210: Why remove this? Haven't tested this yet but I think we'll start getting failures if workspace/root doesn't exist when not using a subvolume no?

DaanDeMeyer: Nvm, `mount_bind` makes sure the directory exists.

DaanDeMeyer:
LGTM. I checked out the branch and verified this fixes the issue.

PR #439, opened 2020-06-30
Pass root around instead of workspace

DaanDeMeyer:
Most of the functions operate on the image root instead of the
workspace so it makes sense to pass the root directory instead of
the workspace. Access to the workspace directory is done via the
workspace function that simply returns the parent directory of the
root directory.

This commit only refactors code and does not change behavior.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -623,7 +623,7 @@ def determine_partition_table(args: CommandLineArguments) -> Tuple[str, bool]:
     return table, run_sfdisk
 
 
-def create_image(args: CommandLineArguments, workspace: str, for_cache: bool) -> Optional[BinaryIO]:
+def create_image(args: CommandLineArguments, root: str, for_cache: bool) -> Optional[BinaryIO]:
     if not args.output_format.is_disk():
         return None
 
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
LGTM, but needs a rebase. Archlinux failure is unrelated:

```
error: ca-certificates-mozilla: key "3B94A80E50A477C7" is unknown
:: Import PGP key 3B94A80E50A477C7? [Y/n] 
:: File /home/runner/work/mkosi/mkosi/mkosi.output/.mkosi-_nu2f2y5/root/var/cache/pacman/pkg/ca-certificates-mozilla-3.53.1-1-x86_64.pkg.tar.zst is corrupted (invalid or corrupted package (PGP signature)).
error: dbus: key "3B94A80E50A477C7" is unknown
Do you want to delete it? [Y/n] 
:: Import PGP key 3B94A80E50A477C7? [Y/n] 
```

we probably need a way to make it always import

DaanDeMeyer:
Rebased.

As far as I can see, pacstrap adds --noconfirm so pacman should be importing the key. I'm not sure why it isn't. I'll try and find the time to look into it.

PR #440, opened 2020-06-30
Stop ignoring umount errors

DaanDeMeyer:
This was hiding a rather nasty bug in pacstrap/unshare. Silencing
the umount errors complicated finding the actual error so let's not
ignore unmounting errors.


lucasdemarchi:
LGTM

PR #441, opened 2020-06-30
Stop gpg-agent and friends even when execution is interrupted

DaanDeMeyer:
This supersedes #373.

CC: @keszybz (Added the contextmanager as requested in the original PR)

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -2159,13 +2159,21 @@ SigLevel = Optional TrustedOnly
 Server = {repository_server}
 """)
 
+    def stop_gpg_agent() -> None:
+        # Kill the gpg-agent used by pacman and pacman-key
+        run(['gpgconf', '--homedir', os.path.join(root, 'etc/pacman.d/gnupg'), '--kill', 'all'])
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -2097,13 +2097,23 @@ SigLevel = Optional TrustedOnly
...
+
     def run_pacman_key(args: List[str]) -> subprocess.CompletedProcess:
         cmdline = [
             "pacman-key",
             "--nocolor",
             "--config", pacman_conf,
         ]
-        return run(cmdline + args, check=True)
+
+        with stop_gpg_agent():
```
lucasdemarchi, line 2115: nitpick, but I think it would make more sense to be `with gpg_agent()`

DaanDeMeyer: I changed the invocations to use try finally instead. In hindsight it doesn't really seem all that beneficial to use a contextmanager here.

DaanDeMeyer:
Let's merge this as it works quite well from my own testing and CI results.

PR #442, opened 2020-06-30
Add better CTRL+C handling when running a subprocess

DaanDeMeyer:
We delay mkosi's cleanup until the subprocess has finished. Because
CTRL+C is sent to the entire process group, we don't have to forward
the signal to the subprocess.

This does break the use case where SIGINT is sent to only mkosi and not the entire mkosi process group. Fixing that would require forwarding the SIGINT signal to the subprocess so we'd have to figure out the pid of the subprocess in the signal handler somehow to send the signal. I'm not entirely sure how we'd do that though, python doesn't seem to have an equivalent to bash's $! which returns the pid of the last command executed.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -21,6 +21,7 @@ import platform
 import re
 import shlex
 import shutil
+import signal
 import stat
 import string
 import subprocess
@@ -50,6 +51,7 @@ from typing import (
     cast,
     TYPE_CHECKING,
 )
+from types import FrameType
 
 __version__ = '5'
 
@@ -95,6 +97,31 @@ def print_error(text: str) -> None:
     sys.stderr.write(f"‣ \033[31;1m{text}\033[0m\n")
 
 
+@contextlib.contextmanager
+def delay_interrupt() -> Generator[None, None, None]:
+    # CTRL+C is sent to the entire process group. We delay its handling in mkosi itself so the subprocess can
+    # exit cleanly before doing mkosi's cleanup. If we don't do this, we get device or resource is busy
+    # errors when unmounting stuff later on during cleanup. We only delay a single CTRL+C interrupt so that a
+    # user can always exit mkosi even if a subprocess hangs by pressing CTRL+C twice.
+    interrupted = False
+    def handler(signal: int, frame: FrameType) -> None:
+        nonlocal interrupted
+        if interrupted:
+            raise KeyboardInterrupt()
+        else:
+            interrupted = True
+
+    s = signal.signal(signal.SIGINT, handler)
+
+    try:
+        yield
+    finally:
+        signal.signal(signal.SIGINT, s)
+
+        if interrupted:
+            die("Interrupted")
+
+
 def run(cmdline: List[str], execvp: bool = False, **kwargs: Any) -> CompletedProcess:
     if 'run' in arg_debug:
         sys.stderr.write('+ ' + ' '.join(shlex.quote(x) for x in cmdline) + '\n')
@@ -103,7 +130,8 @@ def run(cmdline: List[str], execvp: bool = False, **kwargs: Any) -> CompletedPro
             assert not kwargs
             os.execvp(cmdline[0], cmdline)
         else:
-            return subprocess.run(cmdline, **kwargs)
+            with delay_interrupt():
+                return subprocess.run(cmdline, **kwargs)
     except FileNotFoundError as e:
         die(f"{cmdline[0]} not found in PATH.")
 
--- a/mkosi
+++ b/mkosi
@@ -78,7 +78,7 @@ MKOSI_COMMANDS = ("build", "clean", "help", "summary") + MKOSI_COMMANDS_CMDLINE
 arg_debug = ()
 
 
-class MkosiCommandException(Exception):
+class MkosiException(Exception):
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -86,7 +87,25 @@ def run(cmdline: List[str], execvp: bool = False, **kwargs: Any) -> subprocess.C
         assert not kwargs
         os.execvp(cmdline[0], cmdline)
     else:
-        return subprocess.run(cmdline, **kwargs)
+        # CTRL+C is sent to the entire process group. We delay its handling in mkosi itself so the subprocess
+        # can exit cleanly before doing mkosi's cleanup. If we don't do this, we get device or resource is
+        # busy errors when unmounting stuff later on during cleanup.
```
lucasdemarchi, line 92: humn.. but why we don't want to really kill the subprocesses and wait for completion?  It sucks if it's a long running process or even if the process gets stuck for some reason.

I think we will need to be smarter than this to really kill any process an guarantee no one is left behind. @poettering  ideas?

DaanDeMeyer: The subprocess still receives the SIGINT signal and (if it's well-behaved) should exit shortly after receiving it, this just makes mkosi wait until the subprocess exits before continuing. If we don't, we get race conditions since if the subprocess is still running, unmounting stuff later on in the mkosi cleanup will fail.

systemd does something similar by waiting (by default) for 30 seconds after sending SIGTERM before killing a process.

What I could add is an override so that if CTRL+C is pressed again while we're waiting for a subprocess to exit after a first CTRL+C, we stop waiting in mkosi and proceed with cleanup regardless of whether the subprocess is still running or not.

lucasdemarchi: > What I could add is an override so that if CTRL+C is pressed again while we're waiting for a subprocess to exit after a first CTRL+C, we stop waiting in mkosi and proceed with cleanup regardless of whether the subprocess is still running or not.

then we would need to be sure to kill the sub-processes that are still using a mountpoint. I thought we had something like that already

DaanDeMeyer: Currently, we don't. I'm not sure if anything more complex than this can be implemented without moving from the `subprocess.run` interface to `Popen` since `subprocess.run` blocks until the child process exits.

DaanDeMeyer:
Updated to only delay a single CTRL+C. If CTRL+C is pressed twice, mkosi throws KeyboardInterrupt even if the subprocess is still running. This allows users to kill mkosi even if a subprocess hangs.

Also added another commit that unifies error handling a little bit. We use a single exception type MkosiException and always raise instead of calling sys.exit() in the middle of execution so context managers always run properly to clean stuff up. We also print errors immediately instead of from exception handlers which is more intuitive compared to the opposite (which is first all cleanup messages print and then the error message.

mkosi
```diff
@@ -103,15 +104,38 @@ def run(cmdline: List[str], execvp: bool = False, **kwargs: Any) -> CompletedPro
...
             os.execvp(cmdline[0], cmdline)
         else:
-            return subprocess.run(cmdline, **kwargs)
+            # CTRL+C is sent to the entire process group. We delay its handling in mkosi itself so the
+            # subprocess can exit cleanly before doing mkosi's cleanup. If we don't do this, we get device or
+            # resource is busy errors when unmounting stuff later on during cleanup. We only delay a single
+            # CTRL+C interrupt so that a user can always exit mkosi even if a subprocess hangs by pressing
+            # CTRL+C twice.
+            interrupted = False
+            def handler(signal, frame):
```
behrmann, line 113:
```suggestion
            def handler(signal: int, frame: FrameType) -> None:
```
also needs the import
```python
from types import FrameType
```

behrmann:
I am +1 in terms of unifying on a single type of exception, but I'm a bit torn whether it wouldn't make sense at this point to do away with `die`. The call to `print_error` could be put in the except clause of the main function and all calls to die could simply be a `raise MkosiException`. I think this would be more pythonic since explicit is better than implicit.

DaanDeMeyer:
That would indeed be ideal but that leads to a little bit different behavior since then the error message will be printed after all the context manager cleanup messages (which is what happens currently). I think its a lot more intuitive when the error is printed immediately when it occurs instead since then it's easy to associate it with an operation compared to printing it after all cleanup messages.

We could rename `die` to something more descriptive such as `log_and_raise`. That should make it explicit that the function will be raising an exception.

behrmann:
> That would indeed be ideal but that leads to a little bit different behavior since then the error message will be printed after all the context manager cleanup messages (which is what happens currently). I think its a lot more intuitive when the error is printed immediately when it occurs instead since then it's easy to associate it with an operation compared to printing it after all cleanup messages.
> 
You are right, I didn't think about that and this is indeed the better behaviour.

> We could rename `die` to something more descriptive such as `log_and_raise`. That should make it explicit that the function will be raising an exception.
>
I thought that it might be a good idea to refactor all the print statements to use a logger anyway, so this could go hand in hand (although I would prefer logging and raising in place), but that's beyond the scope of this PR.

DaanDeMeyer:
Factored the interrupt handling out into a contextmanager.

DaanDeMeyer:
Let's merge as this should be an improvement over what we have now. We can always do further improvements later.

PR #446, opened 2020-07-01
Use open_close contextmanager for dir_fd

DaanDeMeyer:


behrmann:
The Github diff makes it a bit difficult to see, but it's just straight-forward addition of a level of indentation for the with block.

DaanDeMeyer:
Thanks for the review!

PR #447, opened 2020-07-02
Call pacman-key --refresh-keys in Arch CI

DaanDeMeyer:
This refreshes local keys from the upstream keyserver. This should
mitigate having to update archlinux-keyring all the time somewhat.

It might not work for new keys that aren't in the local keyring installed via archlinux-keyring though.


lucasdemarchi:
LGTM

Issue #391, opened 2019-12-05
RFE: add hook script for installing additional packages with secondary package manager before taking incremental mode snapshot


pieterlexis:
Hi,

For a project I'm working on, I would like to have an image with systemd. So mkosi comes to mind. I need to build some dependencies that are not in any distro, but these don't change often. It would be neat if the build could be done in 2 stages, in the first one I'd build the dependencies and in the second (using the image produced in the first stage) the actual software I'm working on.

Let me know if this is a feature that would make sense for mkosi and I I'll see if I can find the time to implement it.

behrmann:
I don't quite understand, the mkosi build already happens in two phases, quoting from the docs

> mkosi.build may be an executable script. If it exists the image will be built twice: the first iteration will be the development image, the second iteration will be the final image. The development image is used to build the project in the current working directory (the source tree). For that the whole directory is copied into the image, along with the mkosi.build build script. The script is then invoked inside the image (via systemd-nspawn), with $SRCDIR pointing to the source tree. $DESTDIR points to a directory where the script should place any files generated it would like to end up in the final image. Note that make/automake/meson based build systems generally honor $DESTDIR, thus making it very natural to build source trees from the build script. After the development image was built and the build script ran inside of it, it is removed again. After that the final image is built, without any source tree or build script copied in. However, this time the contents of $DESTDIR are added into the image.

poettering:
I guess we have no script hook right now that allows installing stuff with additional package managers before we take the copy in incremental mode. I mean, you could just install the additional stuff from the build script itself, but that means incremental mode doesn't work, and the build script needs to run with networking. We could add another hook that runs right after "dnf" and friends run, with full network, and only after that we take the snapshot for incremental mode. i.e. maybe call it "mkosi.prepare" or so. Would be delighted to take a patch for that.

mmorvai:
I really like the idea of a prepare script. It can be a small step towards reproducible builds if you save the cached image (in incremental mode) to your VCS.

@poettering I have created a pull request which should do exactly what you described in your comment.

DaanDeMeyer:
We have a prepare script option now.

PR #450, opened 2020-07-13
Feature/prepare script

mmorvai:
Added prepare script hook as proposed in #391.

The prepare script has always network access and is run before the image is cached for incremental mode. Just like the postinst script it is called twice, once with "build" and once with "final" as command line argument.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -2547,6 +2547,27 @@ Environment=TERM={os.getenv('TERM', 'vt220')}
 """)
 
 
+def run_prepare_script(args: CommandLineArguments, root: str, do_run_build_script: bool, cached: bool) -> None:
+    if args.prepare_script is None:
+        return
+    if cached:
+        return
+
+    verb = "build" if do_run_build_script else "final"
+
+    with complete_step('Running prepare script'):
+
+        # We copy the prepare script into the build tree. We'd prefer
+        # mounting it into the tree, but for that we'd need a good
+        # place to mount it to. But if we create that we might as well
+        # just copy the file anyway.
+
+        shutil.copy2(args.prepare_script, os.path.join(root, "root/prepare"))
+
+        run_workspace_command(args, root, "/root/prepare", verb, network=True)
+        os.unlink(os.path.join(root, "root/prepare"))
+
+
 def run_postinst_script(args: CommandLineArguments, root: str, do_run_build_script: bool, for_cache: bool) -> None:
     if args.postinst_script is None:
         return
@@ -3824,6 +3845,7 @@ def create_parser() -> ArgumentParserMkosi:
                        help='Additional packages needed for build script', metavar='PACKAGE')
     group.add_argument("--skip-final-phase", action=BooleanAction, help='Skip the (second) final image building phase.', default=False)
     group.add_argument("--postinst-script", help='Postinstall script to run inside image', metavar='PATH')
+    group.add_argument("--prepare-script", help='Prepare script to run inside the image before it is cached', metavar='PATH')
     group.add_argument("--finalize-script", help='Postinstall script to run outside image', metavar='PATH')
     group.add_argument("--source-file-transfer", type=SourceFileTransfer, choices=list(SourceFileTransfer), default=None,
                        help="Method used to copy build sources to the build image." +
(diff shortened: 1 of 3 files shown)


behrmann:
LGTM, I'll give it whirl tomorrow.

mkosi
```diff
@@ -2492,6 +2492,28 @@ def set_root_password(args: CommandLineArguments, root: str, do_run_build_script
...
+
+    with complete_step('Running prepare script'):
+
+        # We copy the prepare script into the build tree. We'd prefer
+        # mounting it into the tree, but for that we'd need a good
+        # place to mount it to. But if we create that we might as well
+        # just copy the file anyway.
+
+        shutil.copy2(args.prepare_script,
+                     os.path.join(root, "root/prepare"))
```
DaanDeMeyer, line 2511: I think the general guideline for line length in mkosi is the same as systemd now (which is 109 columns) so this can be one line.

mkosi
```diff
@@ -3756,6 +3778,7 @@ def create_parser() -> ArgumentParserMkosi:
                        help='Additional packages needed for build script', metavar='PACKAGE')
     group.add_argument("--skip-final-phase", action=BooleanAction, help='Skip the (second) final image building phase.', default=False)
     group.add_argument("--postinst-script", help='Postinstall script to run inside image', metavar='PATH')
+    group.add_argument("--prepare-script", help='Prepare script to run inside image before saving cache image', metavar='PATH')
```
DaanDeMeyer, line 3781:
```suggestion
    group.add_argument("--prepare-script", help='Prepare script to run inside the image before it is cached', metavar='PATH')
```

mkosi.md
```diff
@@ -495,6 +495,18 @@ details see the table below.
...
 
+`--prepare-script=`
+
+: Takes a path to an executable that is invoked inside the image
+  right after installing the software packages. It is
+  the last step before the image is cached (if incremental mode is 
+  enabled).
+  This script is invoked inside a `systemd-nspawn` container
+  environment, and thus does not have access to host resources.
+  If this option is not used, but an executable `mkosi.prepare` is
```
DaanDeMeyer, line 506:
```suggestion
  If this option is not used, but an executable script `mkosi.prepare` is 
```

mkosi.md
```diff
@@ -867,6 +879,20 @@ local directory:
...
   script so that you know which `mkosi.default` (if any) was passed in.
 
+* `mkosi.prepare` may be an executable script. If it exists it is
+  invoked directly after the software packages are installed,
+  from within the image context. It is once called for the *development*
+  image (if this is enabled, see above) with the "build" command line
+  parameter, right before copying the extra tree. It is called a second
+  time for the *final* image with the "final" command line parameter.
+  This script has network access and may be used to install packages
+  with an additional package manager (e.g. pip, npm, ...),
```
DaanDeMeyer, line 889:
```suggestion
  from other sources than the distro's package manager (e.g. pip, npm, ...),
```

mkosi
```diff
@@ -2492,6 +2492,28 @@ def set_root_password(args: CommandLineArguments, root: str, do_run_build_script
             patch_file(os.path.join(root, 'etc/shadow'), jj)
 
 
+def run_prepare_script(args: CommandLineArguments, root: str, do_run_build_script: bool, cached: bool) -> None:
+    if args.prepare_script is None:
+        return
+    if cached:
```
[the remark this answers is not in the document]
behrmann: I agree, but maybe we should do this treewide instead of here particularly, since this is basically just a copy of `run_postinst_script` (and one could maybe factor the common stuff out). I think there is some value in keeping it having the same overall structure as long as it is a mostly verbatim copy.

mkosi
mmorvai: Done.

mkosi.md
mmorvai: Done.

mkosi.md
mmorvai: Done.

DaanDeMeyer:
This has conflicts, can you rebase on master and squash the commits into one commit?

DaanDeMeyer:
Thanks for working on this!

DaanDeMeyer:
Actually, maybe I pulled the trigger here a little too quickly. One thing that's often the case with package managers such as pip/cargo/npm and others is that the dependencies are declared in another file e.g. requirements.txt/Cargo.toml/package.json. I guess that means we should mount the project source directory into the image when we run the prepare script so that whatever package manager that runs from the prepare script has access to the file it needs to install its dependencies. @mmorvai would you mind making a follow up PR for this one that adds that?

behrmann:
Maybe not only the project source, but also more? Something similar was the goal of #400

DaanDeMeyer:
Yeah looking at that one made me reconsider this PR. I think there's two parts to this. One part is simply needing a hook to do stuff that isn't supported in mkosi yet, like pip/cargo/whatever. The prepare script is ideal for that but needs the project source directory so it can get its information from there if necessary. There might also be a need to mount extra caches to streamline this a bit more but that can be added in a separate PR.

On the other hand, it would be nice if we could automatically do the right thing in mkosi for somewhat popular stuff. It should be trivial to detect whether the project source directory has a Cargo.toml and fetch the dependencies if it exists. Same with other package managers. We can discuss that in #400 though.

PR #451, opened 2020-07-16
Add --qemu-headless option

DaanDeMeyer:
This option streamlines launching headless qemu virtual machines from
the command line via mkosi's qemu verb. When this option is specified,
mkosi configures the image for headless connections and adds the
-nographic option when launching the vm via mkosi's qemu verb. Instead
of launching a separate window, qemu connects the host's terminal to the
vm via the serial console. This allows working with the vm from the
comfort of your own terminal instead of the linux console which is used
when qemu starts the vm in a separate window.

We make sure all terminal features still work by overriding the TERM
value for serial-getty@ttyS0 in the vm with the TERM value from the
host's terminal emulator which we get from the TERM environment variable
that mkosi is started with.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1264,7 +1264,7 @@ def prepare_tree(args: CommandLineArguments, root: str, do_run_build_script: boo
         os.mkdir(os.path.join(root, "etc/kernel"), 0o755)
 
         with open(os.path.join(root, "etc/kernel/cmdline"), "w") as cmdline:
-            cmdline.write(' '.join(args.kernel_command_line))
+            cmdline.write(' '.join(args.kernel_command_line + ["console=ttyS0"] if args.qemu_headless else []))
             cmdline.write("\n")
 
     if do_run_build_script:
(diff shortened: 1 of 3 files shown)


mkosi
```diff
@@ -2492,6 +2492,20 @@ def set_root_password(args: CommandLineArguments, root: str, do_run_build_script
...
+    """Override TERM for the serial console with the terminal type from the host."""
+
+    if do_run_build_script or for_cache or not args.qemu_headless:
+        return
+
+    override_dir = mkdir_last(os.path.join(root, "etc/systemd/system/serial-getty@ttyS0.service.d"))
+    with open(os.path.join(override_dir, "override.conf"), 'w') as f:
+        f.write(f"""
+[Service]
+Environment=TERM={os.getenv('TERM')}
```
behrmann, line 2505: This might be a silly question, but is TERM always guaranteed to be set (e.g. in CI systems)? If not a default for getenv would be a good idea.

DaanDeMeyer: Good point, we now default to vt220 which is what systemd does as well.

DaanDeMeyer:
Moved this to a separate qemu argument group since it doesn't really fit the other ones imo.

behrmann:
True, it doesn't really fit anywhere. Maybe it could be added to the (so far rather lonely) `Host` section, because one might argue that this couples the image to the host it was generated on. This might be a bit better than adding another single key section.

mkosi.md
```diff
@@ -655,6 +655,22 @@ details see the table below.
...
+  serial console in the image to the terminal type of the host (more
+  specifically, the value of the TERM environment variable passed to
+  mkosi). This makes sure that all terminal features such as colors
+  and shortcuts still work as expected when connecting to the qemu
+  vm over the serial console (for example via `-nographic`).
+
+  When used with the qemu verb, this option adds the `-nographic`
+  option to qemu's command line so qemu starts a headless vm and
+  connects to its serial console from the current terminal instead
+  of launching the vm in a separate window.
```
behrmann, line 671: A last nitpick. In the only other instance (Line 140, description of the `qemu` verb) it is written in caps as "VM".

DaanDeMeyer:
Last commit was just a doc change so let's not wait for CI.

PR #452, opened 2020-07-17
Simplify CI

DaanDeMeyer:


.github/workflows/ci.yml
```diff
@@ -9,66 +9,75 @@ on:
...
       with:
         python-version: '3.x'
-    - name: Update package cache
-      run: sudo apt-get --assume-yes update
+
     - name: Install dependencies
-      run: sudo apt-get --assume-yes install systemd-container squashfs-tools e2fsprogs xfsprogs libzstd1-dev libgpgme-dev libcurl4-openssl-dev asciidoc
-    - name: Download libarchive source
-      run: |
-        wget "https://github.com/libarchive/libarchive/releases/download/v3.4.3/libarchive-3.4.3.tar.gz"
```
lucasdemarchi, line 29: humn.. so now in ubuntu 20.04 we have zstd?

DaanDeMeyer: Yup, It was only released to github actions a month ago so I'm kinda lucky that I look at it now since all this stuff can be simplified.

The bad thing is yum doesn't exist anymore in focal repos so the centos builds are broken. I'm seeing if I can bootstrap dnf on ubuntu but it's a pain. Ridiculously complex compared to pacman imo.

DaanDeMeyer:
So building dnf on Ubuntu is a massive pain to get working and will probably be a massive pain to maintain as well. I was thinking we could maybe simply host prebuilt mkosi images via github releases for each supported distro (build with mkosi of course) and download those in CI and use them to build the changed images. Then CI becomes a simple matter of pulling down the right image and using it to build the new image. I still need to so if nspawn works as expected in actions but if that works it should be a much simpler way to get this working. Github Releases has a per file size limit of 2G but aside from that there's no limits at all.

DaanDeMeyer:
Nevermind, we don't have loop devices in containers so we wouldn't be able to run mkosi. There's a patchset that fixes this (https://lwn.net/Articles/819625/) but it's going to take ages before that becomes available.

DaanDeMeyer:
I finally managed to get dnf working. I also went ahead and streamlined the arch dependencies process as well and added fedora to the list of distros.

DaanDeMeyer:
Also added a small fix for `detect_distribution` on ubuntu.

behrmann:
@DaanDeMeyer Looks sensible and it obviously runs, so LGTM.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,228 @@
...
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  ci:
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
```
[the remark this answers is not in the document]
DaanDeMeyer: Yeah, maybe you're right, I had to cancel a lot of faulty workflows while testing this out which would have been automatic if I didn't turn this off. I'll remove it.

PR #453, opened 2020-07-17
Don't print stacktrace when a binary is not found.

DaanDeMeyer:
Trying mkosi for the first time and seeing it crash with a stacktrace because you don't have a package manager installed is not an amazing experience for new users. Let's show a simple error message instead when we can't find one of the necessary tools in the PATH.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -66,6 +66,10 @@ MKOSI_COMMANDS = ("build", "clean", "help", "summary") + MKOSI_COMMANDS_CMDLINE
 arg_debug = ()
 
 
+class MkosiCommandException(Exception):
+    """Leads to sys.exit"""
+
+
 def print_step(text: str) -> None:
     sys.stderr.write("‣ \033[0;1;39m" + text + "\033[0m\n")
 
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -82,16 +86,19 @@ def print_error(text: str) -> None:
...
-    else:
-        return subprocess.run(cmdline, **kwargs)
+    try:
+        if execvp:
+            assert not kwargs
+            os.execvp(cmdline[0], cmdline)
+        else:
+            return subprocess.run(cmdline, **kwargs)
+    except FileNotFoundError:
+        raise MkosiCommandException(f"{cmdline[0]} not found in PATH.")
```
behrmann, line 96: Unfortunately I cannot seem to make a multiline suggestion, but a nicer wrapping would be
```python
except FileNotFoundError as e:
    raise MkosiCommandException(f"{cmdline[0]} not found in PATH.") from e
```
This does not immediately affect anything, because the exception is caught in main, but I've always thought about reusing mkosi as a library and there this would lead to a more informative stack trace.

behrmann:
LGTM with a minor suggestion.

PR #454, opened 2020-07-17
add support for OpenMandriva Linux distribution

tpgxyz:
I've added support for OpenMandriva Lx distribution for current rolling release model (cooker, rolling, rock).
mkosi creates non-bootable images without problems, as i had to add support for bsdtar(libarchive) as it is default tar implementation in OpenMandriva.

http://openmandriva.org

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -221,6 +221,7 @@ class Distribution(enum.Enum):
     centos_epel = 8
     clear = 9
     photon = 10
+    openmandriva = 11
 
 
 GPT_ROOT_X86           = uuid.UUID("44479540f29741b29af7d131d5f0458a")  # NOQA: E221
@@ -1207,7 +1208,7 @@ def mount_cache(args: CommandLineArguments, root: str) -> Generator[None, None, 
 
     # We can't do this in mount_image() yet, as /var itself might have to be created as a subvolume first
     with complete_step('Mounting Package Cache'):
-        if args.distribution in (Distribution.fedora, Distribution.mageia):
+        if args.distribution in (Distribution.fedora, Distribution.mageia, Distribution.openmandriva):
             caches = [mount_bind(args.cache_path, os.path.join(root, "var/cache/dnf"))]
         elif args.distribution in (Distribution.centos, Distribution.centos_epel):
             # We mount both the YUM and the DNF cache in this case, as
@@ -1791,6 +1792,67 @@ gpgkey={gpg_key}
 
     reenable_kernel_install(args, root, masked)
 
+@completestep('Installing OpenMandriva')
+def install_openmandriva(args: CommandLineArguments, root: str, do_run_build_script: bool) -> None:
+    release = args.release.strip("'")
+    masked = disable_kernel_install(args, root)
+
+    # OpenMandriva does not (yet) have RPM GPG key on the web
+    gpg_key = '/etc/pki/rpm-gpg/RPM-GPG-KEY-OpenMandriva'
+    if os.path.exists(gpg_key):
+        gpg_key_string = f'file://{gpg_key}'
+        gpgcheck = "gpgcheck=1"
+        cmdline = ["rpm", "--import", gpg_key, "--root", root]
+        run(cmdline, check=True)
+    else:
+        gpgcheck = "gpgcheck=0"
+        gpg_key_string = ""
+
+    if release[0].isdigit():
+        release_model = "rock"
+    elif release == "cooker":
+        release_model = "cooker"
+    else:
+        release_model = release
+
+    if args.mirror:
+        baseurl = f"{args.mirror}/{release_model}/repository/x86_64/main"
+        release_url = f"baseurl={baseurl}/release/"
+        updates_url = f"baseurl={baseurl}/updates/"
+    else:
+        baseurl = f"http://mirrors.openmandriva.org/mirrors.php?platform={release_model}&arch=x86_64&repo=main"
+        release_url = f"mirrorlist={baseurl}&release=release"
+        updates_url = f"mirrorlist={baseurl}&release=updates"
+
+    config_file = os.path.join(workspace(root), "dnf.conf")
+    with open(config_file, "w") as f:
+        f.write(f"""\
+[main]
+{gpgcheck}
+
+[openmandriva]
+name=OpenMandriva {release_model} Main
+{release_url}
+gpgkey={gpg_key_string}
+
+[updates]
+name=OpenMandriva {release_model} Main Updates
+{updates_url}
+gpgkey={gpg_key_string}
+""")
+
+    packages = ["basesystem-minimal", "systemd"] # well we may use basesystem here, but that pulls lot of stuff
+    packages += args.packages or []
+    if args.bootable:
+        packages += ["kernel-release-server", "binutils", "systemd-boot", "dracut", "timezone", "systemd-cryptsetup"]
+    if do_run_build_script:
+        packages += args.build_packages or []
+    invoke_dnf(args, root,
+               args.repositories if args.repositories else ["openmandriva", "updates"],
+               packages,
+               config_file)
+
+    reenable_kernel_install(args, root, masked)
 
 def invoke_yum(args: CommandLineArguments,
                root: str,
@@ -2459,6 +2521,7 @@ def install_distribution(args: CommandLineArguments,
         Distribution.opensuse: install_opensuse,
         Distribution.clear: install_clear,
         Distribution.photon: install_photon,
+        Distribution.openmandriva: install_openmandriva,
     }
 
     install[args.distribution](args, root, do_run_build_script)
@@ -2715,6 +2778,8 @@ def install_boot_loader_clear(args: CommandLineArguments, root: str, loopdev: st
 def install_boot_loader_photon(args: CommandLineArguments, root: str, loopdev: str) -> None:
     install_grub(args, root, loopdev, "grub2")
 
+def install_boot_loader_openmandriva(args: CommandLineArguments, root: str, loopdev: str) -> None:
+    install_grub(args, root, loopdev, "grub2")
 
 def install_boot_loader(args: CommandLineArguments, root: str, loopdev: Optional[str], do_run_build_script: bool, cached: bool) -> None:
     if not args.bootable or do_run_build_script:
@@ -2755,6 +2820,9 @@ def install_boot_loader(args: CommandLineArguments, root: str, loopdev: Optional
 
         if args.distribution == Distribution.centos:
             install_boot_loader_centos(args, root, loopdev)
+
+        if args.distribution == Distribution.openmandriva:
+            install_boot_loader_openmandriva(args, root, loopdev)
 
 def install_extra_trees(args: CommandLineArguments, root: str, for_cache: bool) -> None:
     if not args.extra_trees:
@@ -2908,7 +2976,13 @@ def make_tar(args: CommandLineArguments,
 
     with complete_step('Creating archive'):
         f: BinaryIO = cast(BinaryIO, tempfile.NamedTemporaryFile(dir=os.path.dirname(args.output), prefix=".mkosi-"))
-        run(["tar", "-C", root, "-c", "-J", "--xattrs", "--xattrs-include=*", "."],
+# OpenMandriva defaults to bsdtar(libarchive) which uses POSIX argument list so let's keep a separate list
+        if shutil.which('bsdtar') and args.distribution == Distribution.openmandriva:
+           _tar_cmd = ["bsdtar", "-C", root, "-c", "-J", "--xattrs", "-f", "-", "."]
+        else:
+           _tar_cmd = ["tar", "-C", root, "-c", "-J", "--xattrs", "--xattrs-include=*", "."]
+
+        run(_tar_cmd,
             env={"XZ_OPT": "-T0"},
             stdout=f, check=True)
 
@@ -4402,6 +4476,8 @@ def load_args(args: CommandLineArguments) -> CommandLineArguments:
             args.release = "latest"
         elif args.release == Distribution.photon:
             args.release = "3.0"
+        elif args.distribution == Distribution.openmandriva:
+            args.release = "cooker"
 
     find_cache(args)
 
--- a/mkosi.files/mkosi.openmandriva
+++ b/mkosi.files/mkosi.openmandriva
@@ -1,0 +1,24 @@
+# SPDX-License-Identifier: LGPL-2.1+
+# Let's build an image that is just good enough to build new mkosi images again
+
+[Distribution]
+Distribution=openmandriva
+Release=cooker
+
+[Output]
+Format=gpt_ext4
+Bootable=no
+Output=openmandriva.raw
+
+[Packages]
+Packages=
+        basesystem
+        btrfs-progs
+        dnf
+        dosfstools
+        git-core
+        gnupg
+        squashfs-tools
+        bsdtar
+        systemd
+        systemd-cryptsetup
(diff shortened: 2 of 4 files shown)


mkosi
```diff
@@ -2852,7 +2916,13 @@ def make_tar(args: CommandLineArguments,
 
     with complete_step('Creating archive'):
         f: BinaryIO = cast(BinaryIO, tempfile.NamedTemporaryFile(dir=os.path.dirname(args.output), prefix=".mkosi-"))
-        run(["tar", "-C", root, "-c", "-J", "--xattrs", "--xattrs-include=*", "."],
+# some distributions defaults to bsdtar(libarchive) which uses POSIX argument list so let's keep a separate list
+        if shutil.which('bsdtar') is not False:
```
DaanDeMeyer, line 2920:
```suggestion
        if shutil.which('bsdtar'):
```

mkosi
```diff
@@ -1784,6 +1785,61 @@ gpgkey={gpg_key}
...
     reenable_kernel_install(args, root, masked)
 
+@completestep('Installing OpenMandriva')
+def install_openmandriva(args: CommandLineArguments, root: str, do_run_build_script: bool) -> None:
+    release = args.release.strip("'")
+    masked = disable_kernel_install(args, root)
+
+    # OpenMandriva does not (yet) have RPM GPG key on the web
+    gpg_key = '/etc/pki/rpm-gpg/RPM-GPG-KEY-OpenMandriva'
+    if os.path.exists(gpg_key):
```
DaanDeMeyer, line 1795: Should consider the scenario here where the key is not found. You can define a new exception and catch it in main alongside the two other ones and print an error.

mkosi
```diff
@@ -2852,7 +2916,13 @@ def make_tar(args: CommandLineArguments,
...
     with complete_step('Creating archive'):
         f: BinaryIO = cast(BinaryIO, tempfile.NamedTemporaryFile(dir=os.path.dirname(args.output), prefix=".mkosi-"))
-        run(["tar", "-C", root, "-c", "-J", "--xattrs", "--xattrs-include=*", "."],
+# some distributions defaults to bsdtar(libarchive) which uses POSIX argument list so let's keep a separate list
+        if shutil.which('bsdtar'):
+           _tar_cmd = ["bsdtar", "-cf", "-", "-J", "-C", root, "."]
+        else:
+           _tar_cmd = ["tar", "-C", root, "-c", "-J", "--xattrs", "--xattrs-include=*", "."]
+
+        run([*_tar_cmd],
```
DaanDeMeyer, line 2925: This can simply be `run(_tar_cmd, ...

mkosi
```diff
@@ -2852,7 +2916,13 @@ def make_tar(args: CommandLineArguments,
 
     with complete_step('Creating archive'):
         f: BinaryIO = cast(BinaryIO, tempfile.NamedTemporaryFile(dir=os.path.dirname(args.output), prefix=".mkosi-"))
-        run(["tar", "-C", root, "-c", "-J", "--xattrs", "--xattrs-include=*", "."],
+# some distributions defaults to bsdtar(libarchive) which uses POSIX argument list so let's keep a separate list
+        if shutil.which('bsdtar'):
```
DaanDeMeyer, line 2920: Maybe make this a bit less intrusive, prefer tar over bsdtar but if tar is not found, check if bsdtar exists and use it if it does.

behrmann:
The changes around bsdtar seem to break tar support on Arch, could you have a look at the stack trace?

Adding OpenMandriva to the Github CI would also be worthwhile to see if unrelated changes break it in the future.

mkosi
```diff
@@ -2852,7 +2916,13 @@ def make_tar(args: CommandLineArguments,
...
+           _tar_cmd = ["bsdtar", "-cf", "-", "-J", "-C", root, "."]
```
behrmann, line 2921: It would read nicer if the options for `_tar_cmd` would be in the same order as for GNU tar. `--xattrs` is the default when run as root, but maybe add it for completeness.

tpgxyz: libarchives's bsdtar follow POSIX standard so it has its own options order. libarchive is default tar implementation in OpenMandriva for 4 or 5 years.

tpgxyz:
> The changes around bsdtar seem to break tar support on Arch, could you have a look at the stack trace?
> 
@behrmann sounds like arch is broken. Either its chroot already have broken bsdtar or downstream symlinked gnu tar to bsdtar

mkosi
tpgxyz: @DaanDeMeyer in case of libarchive it does provide tar as a symlink to bsdtar.

There are two options.
1. Parse tar --version output for bsdtar string
2. Check of tar is not a soft/hard link to badtar

mkosi
tpgxyz: Yes I know about that, but public keys are needed for rpm package. So if this file is missing then it means things are heavily broken. I assume that scenario never happens on purpose.

mkosi
tpgxyz: Let me check it again, as I tried without unpacking the _tar_cmd listand run was throwing exceptions

mkosi
DaanDeMeyer: It's possible that we'd want to build openmandriva packages on other distros where this key might not be available. In that case, we'd crash somewhere further down with maybe a not as clear error. If we check here, we get a clear error that there's a missing key that we need which will be much more user friendly than crashing with an unhandled python exception further down.

mkosi
DaanDeMeyer: I think we can make this a distribution check for now (just if mandriva => bsdtar). We can always revisit this if we need bsdtar for something else as well.

mkosi
behrmann: So, `["bsdtar", "-C", root, "-c", "-J", "--xattrs", "-f", "-", "."]` would be an illegal invocation of bsdtar?

mkosi.files/mkosi.openmandriva
```diff
@@ -0,0 +1,24 @@
...
+        basesystem
+        btrfs-progs
+        dnf
+        dosfstools
+        git-core
+        gnupg
+        squashfs-tools
+        bsdtar
+        systemd
+        systemd-cryptysetup
```
DaanDeMeyer, line 24: Shouldn't this be systemd-cryptsetup?

DaanDeMeyer:
Also, please update the readme list of supported distributions.

behrmann:
> @behrmann sounds like arch is broken. Either its chroot already have broken bsdtar or downstream symlinked gnu tar to bsdtar
Arch's bsdtar is also just libarchive's tar (it's owned by the libarchive package) and libarchive is always pulled in by mkinitcpio. It would anyway not be a good idea to break Arch by adding OpenMandriva. I guess this error will go away if bsdtar is hidden behind a check for `Distribution.openmandriva` as @DaanDeMeyer suggested.

johannbg:
In the long run it's probably better to re-factor the code and split out distribution from mkosi into separated files which then get included since there are roughly 220 Linux distribution that use systemd ( and thus potential consumers of mkosi ) and if the idea is to start working around or "hide" the ( arguably needless ) differences in downstream and start reflecting that in the code since they have started to collide with one another ( as is happening OpenMandriva breaking Arch which is a primary distribution ), mkosi can grow quite big, quite messy, quite fast right. 
A better approach is either to just include support for the package managers and or primary distribution in mkosi itself ( Arch/Debian/Fedora/OpenSuse ) while the parent distributions and deviation ( workaround, distribution specific bits ) in the derivatives of the parent distributions can be included in their own specific distribution file which should reduce the risk of them affecting/colliding with each other.

mkosi
tpgxyz: anyways i'll add some error message

mkosi.files/mkosi.openmandriva
tpgxyz: Fixed

mkosi
tpgxyz: Fixed, thanks for the hint.

mkosi
tpgxyz: fixed, now it checks for bsdtar and Distrobution.openmandriva

mkosi
tpgxyz: Thanks, fixed

mkosi
tpgxyz: i've added die() with proper message.

mkosi
```diff
@@ -2887,7 +2954,13 @@ def make_tar(args: CommandLineArguments,
 
     with complete_step('Creating archive'):
         f: BinaryIO = cast(BinaryIO, tempfile.NamedTemporaryFile(dir=os.path.dirname(args.output), prefix=".mkosi-"))
-        run(["tar", "-C", root, "-c", "-J", "--xattrs", "--xattrs-include=*", "."],
+# OpenMandriva defaults to bsdtar(libarchive) which uses POSIX argument list so let's keep a separate list
+        if (shutil.which('bsdtar') and Distribution.openmandriva):
```
behrmann, line 2958:
```suggestion
        if shutil.which('bsdtar') and args.distribution == Distribution.openmandriva:
```
Otherwise the second half of the conditional is always truthy because the value of Distribution.openmandriva is 10. Also, the parentheses are uncommon in Python for single-line statements.

tpgxyz: Thanks for catching this.

behrmann:
Excellent! Now the whole CI is happy. :)

Could you rebase against current master and squash the commits? In going with systemd tradition I think going with perfect commits of logical units is the way to go.

behrmann:
@tpgxyz That went a bit awry. The easiest way to salvage this is to start from a new branch, cherry-pick your commits on it (ae48940a628bfb153c7add329f5b7df57c9efc6a, dafdb979a30477417271f4e4bc4b0fc698781b48, e6c1c001ccbeeebad6858baa146d843b2756151d, e9fa095723668f1c9716d4b8ec05784137466b9e, d17e5d6858d82e4797355a63c876271a82e0d08c) and afterwards rebase them with `git rebase -i` with "pick" for the topmost commit and "fixup" for the later ones to squash them and discard their commits messages. Hope this helps!

I think git rebase might stumble over the backmerge (7f02b336b3e801063ba87d4b479759f5f3fb0724). Backmerges from master to a topic branch are rarely a good idea.

tpgxyz:
@behrmann Sorry for the noise, i did cherry picks with squashing commit messages and updated my branch

DaanDeMeyer:
Almost there, just need to remove the merge commit.

```
git checkout openmandriva
git rebase -i HEAD~2
delete the line with the merge commit
save and exit
git push -f
```

tpgxyz:
> git checkout openmandriva
> git rebase -i HEAD~2

pick 0179323 add support for OpenMandriva Linux distribution
pick b45e3c2 Don't print stacktrace when a binary is not found.
pick 8eac89f mypy: enable warning for unreachable code
pick 20b8828 Make test for uefi boot protocol on Photon reachable
pick 90f1ac4 ci: enable mypy on PRs
pick 3acd555 Implemented bootable CentOS 8
pick 703b981 Add --qemu-headless option
pick 63b9bcf Fix --qemu-headless test.
pick d5476c7 Add build-script option to --debug
pick db27953 Install systemd by default on centos systems.

# Rebase e032aae..6226035 onto 63b9bcf (10 commands)

If mark second and rest rows as fixup i'll end in same situation like before.

> delete the line with the merge commit
> save and exit
> git push -f
> ```

behrmann:
@tpgxyz That commit still has more than it should have. I cleaned up [a version of your PR](https://github.com/behrmann/mkosi/tree/myopenmandriva) on my fork.

tpgxyz:
@behrmann i do not know what is happening here. I all the time got conflicts with https://github.com/systemd/mkosi/commit/3acd5555263a800fd699f4308e372784aa6947ad#diff-6980c629516e4165bfaf96be8e5003cfR2721

behrmann:
@tpgxyz Yeah, when cherry-picking the first commit of your initial series (ae48940) I got a trivial conflict with the commit you mentioned because they edit the same part of `install_boot_loader`. One can just commit the `mkosi` and `git cherry-pick --continue` over it.

behrmann:
The shortest way of generating a single commit for your patch (starting from current master) is
```
git cherry-pick ae48940
git add mkosi
git commit
git cherry-pick --continue
git cherry-pick --no-commit -n dafdb97 e6c1c00 e9fa095 d17e5d6
git commit --amend
```
but you can also take [the branch I generated this way](https://github.com/behrmann/mkosi/tree/myopenmandriva). Sorry for the inconvenience!

mkosi
```diff
@@ -1295,7 +1296,7 @@ def prepare_tree(args: CommandLineArguments, root: str, do_run_build_script: boo
         os.mkdir(os.path.join(root, "etc/kernel"), 0o755)
 
         with open(os.path.join(root, "etc/kernel/cmdline"), "w") as cmdline:
-            cmdline.write(' '.join(args.kernel_command_line))
+            cmdline.write(' '.join(args.kernel_command_line + ["console=ttyS0"] if args.qemu_headless else []))
```
behrmann, line 1299: This line still needs to go, because it would clash with the current HEAD of master.

behrmann:
Great, almost there! Somehow there is still one line that wasn't part of your initial series.

mkosi
tpgxyz: Omg

tpgxyz: should be fixed

tpgxyz:
> 
> 
> Great, almost there! Somehow there is still one line that wasn't part of your initial series.

I think i've sorted it out.

mkosi
```diff
@@ -1791,6 +1792,63 @@ gpgkey={gpg_key}
...
+def install_openmandriva(args: CommandLineArguments, root: str, do_run_build_script: bool) -> None:
+    release = args.release.strip("'")
+    masked = disable_kernel_install(args, root)
+
+    # OpenMandriva does not (yet) have RPM GPG key on the web
+    gpg_key = '/etc/pki/rpm-gpg/RPM-GPG-KEY-OpenMandriva'
+    if os.path.exists(gpg_key):
+        gpg_key = f'file://{gpg_key}'
+    else:
+        die(f'Missing RPM GPG key, please install distro-release-repo-keys.')
```
DaanDeMeyer, line 1805: This is actually not a great idea since it immediately exits the mkosi process without doing any cleanup at all. Maybe replace it with a gpgcheck=0 in the dnf config? That way we can run openmandriva on the CI even without access to the gpg key.

DaanDeMeyer:
@tpgxyz I just merged a PR that builds dnf on the CI machines so I think we could add openmandriva to the CI now as well. Could you do that? Rebase on latest master (.github/workflows/ci.yml) and put in a `gpgcheck=0` in the dnf config if the GPG key doesn't exist to get around the fact that mandriva's gpg key is not on the web yet?

behrmann:
@tpgxyz With #452 merged, you can also add OpenMandriva to the list of distros in `.github/workflows/ci.yml` (and add it to the ones pulling , than we can see right away, that it builds, because the CI runs on the version of the file in the PR, if i understand it correctly.

mkosi
tpgxyz: Fixed

tpgxyz:
> With #452 merged, you can also add OpenMandriva to the list of distros in `.github/workflows/ci.yml`

@behrmann done. Let's see if CI tests will finish with success.

behrmann:
LGTM, barring CI. @DaanDeMeyer your thoughts?

DaanDeMeyer:
Thanks for working on this! When OpenMandriva's key can be imported from the web, you can make another PR to have CI use it as well. For now, this seems to work nicely. One more possible improvement could be to print a warning when we're not verifying packages with gpg and add a command line parameter that forces gpg to be used but let's leave that for subsequent PRs since photon currently does neither as well.

PR #457, opened 2020-07-18
Exchange --password-is-hashed for --password-hashed=

behrmann:
This is more of an RFC than an outright PR. 

This change brings mkosi in line with the similar systemd-firstboot option introduced by @DaanDeMeyer in systemd/systemd#15893. The change is a bigger than `--root-password-is-hashed` was, but I think there is value in having systemd-firstboot's behaviour mirrored. 

Also, the change to `set_root_password` introduces the `nonlocal` keyword, which might be regarded as a code smell. This could also be done with `functools.partial`, but I wanted a to hear a comment first, whether this change would be regarded positively, since it does break backward compatibility (although `--root-password-is-hashed` has never been in a tagged release).

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -2503,25 +2503,29 @@ def set_root_password(args: CommandLineArguments, root: str, do_run_build_script
     if for_cache:
         return
 
-    if args.password == '':
-        with complete_step("Deleting root password"):
-            def jj(line: str) -> str:
-                if line.startswith('root:'):
-                    return ':'.join(['root', ''] + line.split(':')[2:])
-                return line
-            patch_file(os.path.join(root, 'etc/passwd'), jj)
+    if args.password_hashed == "" or args.password == "":
+        msg = "Deleting root password"
+        password = ""
+        target = "etc/passwd"
+    elif args.password_hashed:
+        msg = "Setting hashed root password"
+        password = args.password_hashed
+        target = "etc/shadow"
     elif args.password:
-        with complete_step("Setting root password"):
-            if args.password_is_hashed:
-                password = args.password
-            else:
-                password = crypt.crypt(args.password, crypt.mksalt(crypt.METHOD_SHA512))
+        msg = "Setting root password"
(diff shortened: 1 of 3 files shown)


DaanDeMeyer:
I don't mind, especially if it's not been in a tagged release but we should wait for some input from @poettering before we start breaking compat.

johannbg:
Not being in tagged releases is no excuse for breaking existing implementation for years. Is it not better to change systemd-firstboot behaviour to be in-line with this instead given that it has been newly merged...

DaanDeMeyer:
@johannbg My first implementation of this in systemd-firstboot followed the same approach as mkosi did. However, @poettering was not a fan of options that modify the behaviour of other options so the current solution was implemented instead (which I'm somewhat more partial to as well). This isn't really a massive problem since we can easily transform the mkosi arguments to whatever systemd-firstboot requires. You make a point however that since this has been in for years, breaking compat just to have both mkosi and systemd-firstboot use the same interface doesn't really seem worth it.

behrmann:
@johannbg Well, "years" is a bit strong. :) It's been in the tree since September 21st last year.

Would everybody be happy if we still accepted `--root-is-hashed`, scrapping its documentation and emitting a warning on it being used?

johannbg:
Well I personally consider "fading out period" being an acceptable compromise in most case since it gives people the time to adjust and it's probably not a good course of action if either of the components have to be adjusted to "mirror" each other everytime a change is made on either side ( thou an "rare" exception could be made in this particular case ). That said what is the is value in having systemd-firstboot's behaviour mirrored beside being a "cosmetic" one?

behrmann:
> That said what is value in having systemd-firstboot's behaviour mirrored beside being a "cosmetic" one?

I don't see it as purely cosmetic. It's a more discoverable for a (new) user and therefore a better UX.

I'll readd `--password-is-hashed` later.

DaanDeMeyer:
Let's not bother with this one, it doesn't add too much value overall.

PR #459, opened 2020-07-18
Add build-script option to --debug

DaanDeMeyer:
Drops the user into a shell if the build script fails with a non-zero
exit code. We still fail the build afterwards so the user can
immediately start again after he fixes his build script.


behrmann:
Nice ergonomic change!

PR #460, opened 2020-07-19
Install systemd by default on centos systems.

DaanDeMeyer:
--bootable is intended to install a bootloader into the image. mkosi
boot should always work which means we need systemd in the image by
default.


behrmann:
LGTM, mkosi should always use systemd, but I'm a bit surprised that it's necessary. Doesn't nothing else pull in systemd by default?

DaanDeMeyer:
The only thing installed by default is centos-release. I couldn't find exactly what that entails but trying to boot it failed saying init didn't exist which was fixed by installing systemd explicitly.

PR #462, opened 2020-07-19
Increase default root size to 3G

DaanDeMeyer:
1G isn't sufficient for a default ubuntu focal bootable image. Let's
up the size for a better new user experience.

We also remove the special casing for XFS since it should comfortably
fit under the 3G as well.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -4529,11 +4529,8 @@ def load_args(args: CommandLineArguments) -> CommandLineArguments:
     args.xbootldr_size = parse_bytes(args.xbootldr_size)
     args.swap_size = parse_bytes(args.swap_size)
 
-    if args.output_format in (OutputFormat.gpt_ext4, OutputFormat.gpt_btrfs) and args.root_size is None:
-        args.root_size = 1024*1024*1024
-
-    if args.output_format == OutputFormat.gpt_xfs and args.root_size is None:
-        args.root_size = 1300*1024*1024
+    if args.root_size is None:
+        args.root_size = 3*1024*1024*1024
 
     if args.bootable and args.esp_size is None:
         args.esp_size = 256*1024*1024


DaanDeMeyer:
Seems I don't understand how xfs works. Anyone have an idea why this doesn't work for xfs?

behrmann:
I have no idea either, in the original commit it only says 
>Root size is 1300Mib in this case because of additional metadata

Maybe @randomvariable (the original author) can comment on this?

randomvariable:
Barely remember the reason, but I'm guessing XFS had extra journaling metadata of around 300MB. If you're bumping up to 3GB anyway, then the special casing can go I think.

DaanDeMeyer:
Ah wait it was just a stupid mistake by me I forgot to let the 3G statement apply to xfs as well when I first tried to remove the special casing. Let's see if this works now.

DaanDeMeyer:
I'm a bit scared of merging this since it could cause issues for anyone using mkosi at scale since you suddenly have 3x the disk usage which adds up when building a lot of images. Maybe we should gate this behind a distribution/bootable check? (Only do this for bootable ubuntu images)

behrmann:
I tend to agree that gating this behind a distribution check makes sense for the time being. Besides that, I think a check whether the output format `is_disk()` might make more sense than the image being bootable. 

Maybe once a new release is cut, the gating could be taken away and this could be made the default.

DaanDeMeyer:
So it turns out we're actually using sparse files so what we use as the root size is only the apparent size and not the actual size on disk. So this is less impactful than I first thought. It does still impact filesystems and tools that do not support sparse files though, I'm wondering how widespread that is in practice though. I think we should be fine.

behrmann:
I don't think any of the filesystems we target as output doesn't support sparse files and I don't assume most people are writing their images to anything but those filesystems.

DaanDeMeyer:
OK, let's merge

PR #463, opened 2020-07-19
Fix --qemu-headless so it works for all bootloaders.

DaanDeMeyer:


mkosi
```diff
@@ -4567,6 +4567,9 @@ def load_args(args: CommandLineArguments) -> CommandLineArguments:
         if not args.output_format.is_disk():
             die("Sorry, can't boot non-disk images with qemu.")
 
+    if args.qemu_headless and not "console=ttyS0" in args.kernel_command_line:
```
behrmann, line 4570:
```suggestion
    if args.qemu_headless and "console=ttyS0" not in args.kernel_command_line:
```
would be more pythonic

behrmann:
LGTM, besides the stylistic quibble

PR #464, opened 2020-07-20
Implemented SELinux for CentOS 8

JoseTeuttli:
This patch implements support for SELinux for CentOS 8. This means the CentOS 8 image boots with SELinux in enforcing mode.

This includes adding a `--selinux` flag to mkosi. Installing the required SELinux packages for CentOS 8 that provide full SELinux tools. 

The largest change is adding the method create_autorelabel_file. This method is needed because when installing packages their files don't get labeled properly. This causes important functionality such as auditd to fail to start during boot because of SELinux. This is fixed by creating a ./autorelabel file which relabels all the files on first boot, then restarts. This allows the image to boot in enforcing mode starting from first boot.

This was tested on x86 Fedora 32 Build

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1997,6 +1997,9 @@ def install_centos(args: CommandLineArguments, root: str, do_run_build_script: b
         repos += ["epel"]
         packages += ["epel-release"]
 
+    if args.selinux:
+        packages += ["policycoreutils", "selinux-policy", "selinux-policy-targeted", "libselinux-utils", "setroubleshoot-server", "setools", "setools-console", "mcstrans"]
+
     if do_run_build_script:
         packages += args.build_packages or []
 
@@ -3241,6 +3244,11 @@ def install_unified_kernel(args: CommandLineArguments,
 
             run_workspace_command(args, root, *dracut)
 
+def create_autorelabel_file(args: CommandLineArguments,
+                           root: str) -> None:
+    touch = ["/usr/bin/touch",
+             "/.autorelabel"]
+    run_workspace_command(args, root, *touch)
 
 def secure_boot_sign(args: CommandLineArguments, root: str, do_run_build_script: bool, for_cache: bool) -> None:
     if do_run_build_script:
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -3241,6 +3244,11 @@ def install_unified_kernel(args: CommandLineArguments,
 
             run_workspace_command(args, root, *dracut)
 
+def create_autorelabel_file(args: CommandLineArguments,
+                           root: str) -> None:
+    touch = ["/usr/bin/touch",
+             "/.autorelabel"]
+    run_workspace_command(args, root, *touch)
```
DaanDeMeyer, line 3251: `run_workspace_command(args, root, 'touch', '/.autorelabel')`?

mkosi
```diff
@@ -4339,6 +4349,10 @@ def load_args(args: CommandLineArguments) -> CommandLineArguments:
 
     args.extra_search_paths = expand_paths(args.extra_search_paths)
 
+    if args.selinux:
+        args.kernel_command_line.remove('selinux=0')
+        args.kernel_command_line.remove('audit=0')
```
DaanDeMeyer, line 4354: Let's invert this, remove these args from the default kernel args and if selinux is not specified, add them to the kernel args.

mkosi
```diff
@@ -1997,6 +1997,9 @@ def install_centos(args: CommandLineArguments, root: str, do_run_build_script: b
         repos += ["epel"]
         packages += ["epel-release"]
 
+    if args.selinux:
+        packages += ["policycoreutils", "selinux-policy", "selinux-policy-targeted", "libselinux-utils", "setroubleshoot-server", "setools", "setools-console", "mcstrans"]
```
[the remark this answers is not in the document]
behrmann: I'm really not a fan of the overly long lines - personally I find 109 already a tad on the long side - but the `setup.cfg` lists 119 as maximum line length. 

Nevertheless, this should be broken after `"libselinux-utils"`

DaanDeMeyer:
Looking at this again, from a quick look it seems possible to do all this stuff outside of mkosi with tools already provided. Adding an option for a single distribution is something we generally want to avoid. Let's close this. Feel free to post a comment if you need more info about getting this to work with existing mkosi features or if there's a part that doesn't work.

PR #466, opened 2020-07-24
Fix default source file transfer

bcbnz:
According to the man page, `--source-file-transfer` "defaults to copy-git-cached if a git source tree is detected, otherwise copy-all". However the existing code only sets it to copy-git-cached if a .git folder exists and leaves it at None otherwise, meaning a non-Git source is never transferred.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -2916,13 +2916,11 @@ def install_build_src(args: CommandLineArguments, root: str, do_run_build_script
         if args.build_sources is not None:
             target = os.path.join(root, "root/src")
 
-            source_file_transfer = args.source_file_transfer
-            if source_file_transfer is None and (os.path.exists('.git') or os.path.exists(os.path.join(args.build_sources, '.git'))):
-                source_file_transfer = SourceFileTransfer.copy_git_cached
-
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
Humn... at one point I think this came from the args definition (https://github.com/systemd/mkosi/blob/6d10c524cab7d67ca614d33b619586427e1a3316/mkosi#L3853). Maybe we should change the default there?

bcbnz:
Would that allow the choice between copy-git-cached and copy-all based on detecting a git source tree though?

(For the record, I have no strong opinions on how it is done, as long as the manual and the code are in sync. It took me some time to figure out my build script was failing because the source was not being copied as promised. Having to manually specify the option on the CLI or in mkosi.default is not a big deal.)

behrmann:
I'm pretty sure you could generate the default via a function that checks for `.git`, but since it also depends on `args.build_sources` your way seems to be the cleanest solution.

DaanDeMeyer:
Let's move this logic to `load_args` while we're at it. That seems to be the most logical place for it to be.

DaanDeMeyer:
Rebased on master and moved the logic to `load_args` so we can move this forward since this is an important fix.

behrmann:
I guess the CI failure is unrelated

DaanDeMeyer:
Tested locally and can confirm this works, let's merge. Thanks @bcbnz!

PR #469, opened 2020-07-27
Require --merged-usr support in debootstrap.

DaanDeMeyer:
Let's assume debootstrap is recent enough to support --merged-usr. The
option was added in 2016 so it should be widely available now.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1984,16 +1984,16 @@ def install_debian_or_ubuntu(args: CommandLineArguments,
                "--variant=minbase",
                "--include=systemd-sysv",
                "--exclude=sysv-rc,initscripts,startpar,lsb-base,insserv",
+               "--merged-usr",
                "--components=" + ','.join(repos)]
 
     if args.architecture is not None:
         debarch = DEBIAN_ARCHITECTURES.get(args.architecture)
         cmdline += [f"--arch={debarch}"]
 
-    # Let's use --merged-usr and --no-check-valid-until only if debootstrap knows it
-    for arg in ["--merged-usr", "--no-check-valid-until"]:
-        if debootstrap_knows_arg(arg):
-            cmdline += [arg]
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
Also added a commit to drop Semaphore CI as it seems to still be running on Ubuntu Xenial and also seems to be surpassed in almost every way by Github Actions. Maintaining it alongside Github Actions doesn't seem like it's worth the effort.

behrmann:
I think requiring `--merged-usr` is fine and it has been the default for everything but buildds for a while, but I think the removal of semaphore could have been it's own PR. Also, it might be a good idea to keep the files from `mkosi.files` around as examples for new users.

DaanDeMeyer:
I'll move the semaphore commit to a new PR. I agree about the mkosi.files/ but right now there's about 6 different ubuntu configurations that only differ in the filesystem I think. These were I assume mostly added for the semaphore CI. We can always add a more full-featured ubuntu config in mkosi.files again later.

DaanDeMeyer:
Actually, I'll wait with removing Semaphore as GA is currently a bit spotty. (See https://github.community/t/jobs-randomly-hanging-at-starting-your-workflow-run/124295/5). If we switch to a single solution we should at least wait until it's working properly.

behrmann:
@DaanDeMeyer could you rebase this? Somehow LGTM thinks this has conflicts. My assumption for the breaking semaphore CI is that the `debootstrap` there doesn't know `--merged-usr` yet. It would be good to know what ubuntu is running there to know whether breaking this backwards compatibility is reasonable.

behrmann:
@DaanDeMeyer The rebase squashed into HEAD on master

DaanDeMeyer:
Thanks for the heads up! let me fix that

behrmann:
The oldest debootstrap I care about (on buster) speaks `--merged-usr`, so Ubuntu should be fine.

PR #470, opened 2020-07-27
Don't install bootable packages when building the build image

DaanDeMeyer:


behrmann:
Could you elaborate the reasoning behind this change? I don't quite get it.

DaanDeMeyer:
Faster build image generation. We're only using the build image to compile software. No reason to install a kernel and whatnot in there.

behrmann:
Agreed, none of these should be necessary for that and it's fair to require them if needed.

PR #473, opened 2020-07-28
Use Ubuntu Focal in all Github Actions workflows

DaanDeMeyer:
We also stop installing Python via setup-python as the version in
Focal should be more than recent enough.

### DIFF
--- a/.github/workflows/ci-mypy.yml
+++ b/.github/workflows/ci-mypy.yml
@@ -10,12 +10,9 @@ on:
 
 jobs:
   unit-test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@v2
-    - uses: actions/setup-python@v1
-      with:
-        python-version: '3.x'
     - name: install mypy via pip
       run: python3 -m pip install 'mypy==0.770'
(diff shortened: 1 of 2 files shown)


behrmann:
It might make sense to still use setup-python. At the moment mkosi targets python 3.6 and it might easily happen that something is introduced, that breaks compatibility with the 3.6 stdlib, so it might actually make more sense to tighten the CI here and target and older Python version.

A different topic is that the minimum Python version could probably be bumped to 3.7 now...

DaanDeMeyer:
Have there ever been any issues with using features from newer versions? I'm not sure if we can make an educated guess at how likely this is going to bite us. I had to drop setup-python in the main CI script because it did not play well with dnf and libdnf's build scripts so I wanted to drop the usages here as well for consistency.

behrmann:
I guess the risk is quite low. The `run` wrapper shields people from accidentally using `text=` as alternative to `universal_newlines=` and as long as people don't start using `f"{variable=}"` to print `variable=value` or use dataclasses we should be fine.

DaanDeMeyer:
OK, let's merge. in the event that anything breaks we should get complaints anyway and we can fix it then.

PR #475, opened 2020-07-28
Script improvements

DaanDeMeyer:
- Refactor prepare and postinst to use the same function.
- Check if script files exist and are executable before building images.


DaanDeMeyer:
Dropped the refactoring, didn't actually provide that much benefit.

behrmann:
I would have named it `ensure_valid_script`, but LGTM.

PR #476, opened 2020-07-29
Dnf refactoring

DaanDeMeyer:
- Introduce setup_dnf to write the dnf config file
- Remove config_file argument everywhere, just have dnf, tdnf and yum all use dnf.conf as the config file.
- Add Mageia to CI

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1545,10 +1545,9 @@ def clean_tdnf_metadata(root: str) -> None:
 def invoke_dnf(args: CommandLineArguments,
                root: str,
                repositories: List[str],
-               packages: List[str],
-               config_file: str) -> None:
+               packages: List[str]) -> None:
     repos = ["--enablerepo=" + repo for repo in repositories]
-
+    config_file = os.path.join(workspace(root), 'dnf.conf')
     packages = make_rpm_list(args, packages)
 
     cmdline = ["dnf",
@@ -1577,13 +1576,9 @@ def invoke_dnf(args: CommandLineArguments,
     with mount_api_vfs(args, root):
         run(cmdline, check=True)
 
-def invoke_tdnf(args: CommandLineArguments,
-                root: str,
-                repositories: List[str],
-                packages: List[str],
-                config_file: str) -> None:
+def invoke_tdnf(args: CommandLineArguments, root: str, repositories: List[str], packages: List[str]) -> None:
     repos = ["--enablerepo=" + repo for repo in repositories]
-
+    config_file = os.path.join(workspace(root), 'dnf.conf')
     packages = make_rpm_list(args, packages)
 
     cmdline = ["tdnf",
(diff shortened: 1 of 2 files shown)


mkosi
```diff
@@ -1600,56 +1595,57 @@ def invoke_tdnf(args: CommandLineArguments,
...
 
     packages = ["minimal"]
     if not do_run_build_script and args.bootable:
         packages += ["linux", "initramfs"]
 
-    invoke_tdnf(args, root,
-                args.repositories if args.repositories else ["photon", "photon-updates"],
-                packages,
-                config_file)
+    invoke_tdnf(args, root, args.repositories if args.repositories else ["photon", "photon-updates"], packages)
```
behrmann, line 1648:
```suggestion
    invoke_tdnf(args, root, args.repositories or ["photon", "photon-updates"], packages)
```
To make this more inline with line 1745, also it's nicer to read :)

mkosi
```diff
@@ -1600,56 +1595,57 @@ def invoke_tdnf(args: CommandLineArguments,
...
-
-    gpg_key = '/etc/pki/rpm-gpg/VMWARE-RPM-GPG-KEY'
-    gpg_key_string = f'file://{gpg_key}'
 
-    if os.path.exists(gpg_key):
-        gpgcheck = "gpgcheck=1"
-        cmdline = ["rpm", "--import", gpg_key, "--root", root]
-        run(cmdline, check=True)
+def setup_dnf(args: CommandLineArguments, root: str,
+              repos: List[Tuple[str, str, str, str, Optional[str]]] = []) -> None:
```
behrmann, line 1600: This type signature is pretty much inscrutable. :) It would be better if this were a dataclass or (since we're still on 3.6) a named tuple.

mkosi
```diff
@@ -1600,56 +1595,63 @@ def invoke_tdnf(args: CommandLineArguments,
...
-        cmdline = ["rpm", "--import", gpg_key, "--root", root]
-        run(cmdline, check=True)
+def setup_dnf(args: CommandLineArguments, root: str, repos: List[Repo] = []) -> None:
+    gpgcheck = True
 
-    else:
-        gpgcheck = "gpgcheck=0"
+    repo_file = os.path.join(workspace(root), "temp.repo")
+    with open(repo_file, 'w') as f:
+        for (id, name, url, gpgpath, gpgurl) in repos:
```
behrmann, line 1610: It might make sense not to destructure this here, since named tuples allow dot-access, but I don't particularly mind either way. Whatever you like better.

behrmann:
LGTM, but it seems Mageia is still missing the dnf dependency step.

mkosi
DaanDeMeyer: I actually had no idea about NamedTuple, neat!

behrmann: Yes, they are great, but have been mostly supplanted by dataclasses on 3.7+, since they are still tuples, i.e. they are immutable (generally a good thing) and compare with tuples of the same size (might not be wanted, if they are supposed to represent different types), but for now they are the best thing we have.

PR #478, opened 2020-07-30
Don't depend on python3 being at /usr/bin/python3

DaanDeMeyer:


behrmann:
I was a bit quick there. This is a sensible change in principle. It allows using mkosi with the python executable from a virtual environment, which is great e.g. for testing with multiple different versions of python via pyenv in different virtual environments. 

The downside is, that it will pick up the python version of a venv if one has one activated at the moment. This should not be a problem, since mkosi has no dependencies outside of the stdlib and should be compatible to everything 3.6+, but it's still a gotcha to keep in mind. Should we document that somewhere?

DaanDeMeyer:
I don't think we should. I mean, when you think about it, mkosi is kind of an alternative to a venv. I don't think there will be many situations where both will be used at the same time. (unless for testing purposes but that's internal stuff and not user-facing)

behrmann:
I don't necessarily see mkosi as an alternative to venv, but rather orthogonal. People deploying Python projects usually deploy them inside virtualenvs, just because tools like poetry or pipenv do that by default. I have a venv for mkosi, to keep a downgraded mypy version. 

I guess this both goes ways, because previously mkosi would have always ignored the venv python binary. Well, if somebody runs into this problem, they will surely find this thread.

PR #482, opened 2020-07-30
Socket activated devserver

DaanDeMeyer:
So this is still very WIP.

I got as far as being able to start the mkosi container using socket activation with ssh, but the connection is immediately terminated with the following error:

```
daan@Clevo-N65 ~/p/systemd (firstboot-prompt)> ssh ::1 -p 7774
kex_exchange_identification: read: Connection reset by peer
Connection reset by ::1 port 7774
```

So somewhere along the path something is still going wrong.

The approach is inspired by http://0pointer.de/blog/projects/socket-activated-containers.html.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -70,8 +70,8 @@ else:
 
 MKOSI_COMMANDS_CMDLINE = ("shell", "boot", "qemu")
 MKOSI_COMMANDS_NEED_BUILD = MKOSI_COMMANDS_CMDLINE
-MKOSI_COMMANDS_SUDO = ("build", "clean") + MKOSI_COMMANDS_CMDLINE
-MKOSI_COMMANDS = ("build", "clean", "help", "summary") + MKOSI_COMMANDS_CMDLINE
+MKOSI_COMMANDS_SUDO = ("build", "clean", "link") + MKOSI_COMMANDS_CMDLINE
+MKOSI_COMMANDS = ("build", "clean", "help", "summary", "link") + MKOSI_COMMANDS_CMDLINE
 
 DRACUT_SYSTEMD_EXTRAS = [
     "/usr/lib/systemd/systemd-veritysetup",
@@ -1469,6 +1469,74 @@ def enable_networkmanager(root: str) -> None:
     run(["systemctl", "--root", root, "enable", "NetworkManager"])
 
 
+def setup_devserver(args: CommandLineArguments, root: str, do_run_build_script: bool, for_cache: bool) -> None:
+    if for_cache:
+        return
+    if not args.devserver:
+        return
+
+    with complete_step("Setting up devserver"):
+        image_socket_path = os.path.join(root, "etc/systemd/system/sshd.socket")
+        image_service_path = os.path.join(root, "etc/systemd/system/sshd@.service")
+        image_stop_path = os.path.join(root, "etc/systemd/system/poweroff-when-unneeded.service")
+
+        with open(image_socket_path, "w") as f:
+            f.write(
+                dedent(
+                    f"""\
+                    [Unit]
+                    Description=SSH Socket for {args.devserver_name} devserver
+
+                    [Socket]
+                    ListenStream={args.devserver_port}
+                    Accept=yes
+
+                    [Install]
+                    WantedBy=sockets.target
+                    """
+                )
+            )
+
+        with open(image_service_path, "w") as f:
+            f.write(
+                dedent(
+                    f"""\
+                    [Unit]
+                    Description=SSH Server for {args.devserver_name} devserver
+                    Wants=sshdgenkeys.service poweroff-when-unneeded.service
+                    After=sshdgenkeys.service
+
+                    [Service]
+                    ExecStart=/usr/sbin/sshd -i
+                    StandardInput=socket
+                    """
+                )
+            )
+
+        with open(image_stop_path, "w") as f:
+            f.write(
+                dedent(
+                    f"""\
+                    [Unit]
+                    StopWhenUnneeded=true
+
+                    [Service]
+                    Type=oneshot
+                    RemainAfterExit=yes
+                    ExecStop=systemctl poweroff
+                    """
+                )
+            )
+
+        def jj(line: str) -> str:
+            if "PermitRootLogin" in line:
+                return "PermitRootLogin yes"
+            return line
+        patch_file(os.path.join(root, "etc/ssh/sshd_config"), jj)
+
+        run(["systemctl", "--root", root, "enable", "sshd.socket"])
+
+
 def run_workspace_command(args: CommandLineArguments,
                           root: str,
                           cmd: List[str],
@@ -2448,6 +2516,8 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
         packages.add("linux")
 
     if do_run_build_script:
+        if args.devserver:
+            packages.add("openssh")
         packages.update(args.build_packages)
 
     def run_pacman(packages: Set[str]) -> None:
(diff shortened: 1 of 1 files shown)


behrmann:
First of all, I think this is an useful idea.

Some thoughts on the design:
- Wouldn't it make sense for `devserver` to be a verb (like `qemu`) and then for all the devserver settings to get their own group/secion in the config?
- What is the upside on using socket activation in this case (besides the socket being available right away)? Not using socket activation could let us use the service files that the distros ship (and I am not aware of any distro shipping socket activated sshd). The images are already very minimal and shouldn't start superfluous services so startup should be fast.
- Should devserver maybe imply other settings? It could set the default output to `/var/lib/machines` to properly integrate with `machinectl` or maybe it could drop a file into `/etc/sysusers.d` for a `mkosi` user that is automatically added to `wheel` (or whatever the admin group of that distro is).

johannbg:
The problem with turning an image generation tool into something it's not ( some form of server ) is that you need to setup the host environment for running that server which includes things like you pointed out adding users to group, setting up polkit rules for that or those users etc. so basically the only thing that can be done in mkosi is to add an -r --run switch which just executes systemd-nspawn $foo to spawn/run the newly created image and the rest needs to be configured/setup inside the container or image and on the host system.  

On top of that it's generally considered a best practice to avoid running SSH in a container and recommended to use machinectl instead.

behrmann:
> On top of that it's generally considered a best practice to avoid running SSH in a container and recommended to use machinectl instead.

That is true, but as far as I imagine @DaanDeMeyer's use case, he wants to generate and image and have it running in such a way that third-party programs that speak SSH (he mentioned VSCode, for me Emacs comes to mind) to work (in that case develop) remotely on that machine. While I generally would prefer if those programs would be compatible with machinectl (actually, it's a cool idea to teach Tramp machinectl), some of them don't and only will if they are taught, but they speak SSH. So I don't see this functionality much differently than, e.g. the `qemu` verb as a nice to have feature that will make life nicer for some users while not impacting those who don't use it. 

I might be wrong, though. I don't know where @DaanDeMeyer wants to go with this.

johannbg:
Presumably you always need to setup the ssh connection on your workstation as a jumper for this to work with ssh listening to different ports or the same port using different ip addresses.

You would put something like below in ~/.ssh/config

Host remote-container-s
        User $USER-ON-REMOTE-SRV
        Hostname $IP

Host dev-01
        User $USER-ON-DEV-01
        Hostname 127.0.0.1
        Port 2222
        ProxyCommand ssh -q -W %h:%p remote-srv

Host dev-02
        User $USER-ON-DEV-02
        Hostname 127.0.0.1
        Port 2223
        ProxyCommand ssh -q -W %h:%p remote-srv

Or

Host remote-container-srv
        User $USER-ON-REMOTE-SRV
        Hostname $IP

Host dev-01
        User $USER-ON-DEV-01
        Hostname $IP
        Port 2222
        ProxyCommand ssh -q -W %h:%p remote-srv

Host dev-02
        User $USER-ON-DEV-02
        Hostname $IP
        Port 2222
        ProxyCommand ssh -q -W %h:%p remote-srv


then you could ssh directly into the container/image via ssh dev-01 and start hacking away which is presumably is the desired behaviour so basically all mkosi would need to to is to run the container/image after it gets created which brings us back to just implement a "-r --run" switch in mkosi but not turning into some kind of server.

DaanDeMeyer:
I'm still struggling with the design, maybe posting my problems will help in clarifying:

RIght now, I can use mkosi to build and test systemd in isolation which is starting to work really well. However, I still have a local build of systemd outside of mkosi so that my editor (vscode) and the accompanying language server (clangd) can find all the necessary include files, compiler flags, ... to provide rich auto completion and diagnostics in vscode when working on systemd. This defeats part of the purpose of mkosi since ideally I should not have to pollute my system with systemd build dependencies when using mkosi. To solve that, I want vscode and clangd to use the dependencies and build configuration that mkosi stores in the mkosi build image. Lucky for me, vscode supports remote development in containers (docker/podman) or over ssh. So, if I install clangd and ssh in the build image and connect vscode to it using ssh, vscode will run clangd (and any other language servers I might want to run)  in the build image and provide the same experience as developing locally but without having to install all of systemd's build dependencies (compiler, libs, include files, ...).

The socket activation is so you don't even have to manually start the devserver manually. I just connect to the systemd-devserver SSH host from within VSCode and systemd automatically starts the build image for me. It makes the entire thing a seamless experience.

One thing I haven't figured out is how this should integrate in the usual mkosi development flow. I'm leaning towards having mkosi check if a devserver is running and if one is running, simply use machinectl to execute the build script in the running build image instead and run the final image like usual outside the devserver. 

I'm not sure if we need to do all too many changes to the build image aside from adding ssh and maybe configuring it a bit. We just do root login just like we do with the final images now and bind mount the sources in. When the devserver shuts down, we do a recursive chown to make sure all newly added stuff is assigned to the correct group/user. 

@behrmann I was thinking of making `mkosi dev` boot the build image. It's probably a good idea to add a separate argument group.

Shutting down after the last ssh connection exits is also not supported by systemd I think. I made an issue for that: https://github.com/systemd/systemd/issues/16637

behrmann:
> One thing I haven't figured out is how this should integrate in the usual mkosi development flow. I'm leaning towards having mkosi check if a devserver is running and if one is running, simply use machinectl to execute the build script in the running build image instead and run the final image like usual outside the devserver.

I'm not quite sure I fully understand this
1. You use an image as a workspace holding all the build dependencies for the project you are working on.
2. You want to start the image automatically via socket-activation (when you start VSCode)
3. In that image you want to run mkosi again to build a image for testing.
4. mkosi on the host should then defer to a mkosi instance in your workspace, extricate the resulting image and run it as a sibling to your workspace image on the host

Sorry if I misparsed your intention.

DaanDeMeyer:
Almost, I just want to run mkosi outside the image again to build the final test image. But when doing that and a devserver is already running, instead of building a new build image, mkosi could just run the build script in the already running devserver via machinectl.

behrmann:
Understood. The matching is supposed to be via having machinectl check if an image with `devserver-name` already exists?

DaanDeMeyer:
Got the basic idea working, `ssh root@::1 -p 7774` starts the container and starts an ssh session in it. Now its just a matter of making it use the build image instead of the final image and making sure building gets done using `machinectl` when using the devserver. I'm also going to look into something that allows the whole build/boot process to be done in a loop. Idea is that you're developing in the devserver image and testing in the final image, whenever you reboot the final image with `systemctl reboot`, it builds a new final image with the latest changes and boots it again. This should be no problem with nspawn but might be harder with qemu.

DaanDeMeyer:
This now provides `mkosi link` which sets up the unit files on the host (socket and service). After running mkosi link, you can `ssh root@::1 -p 7774` and systemd will invoke mkosi to build the build image and run the build script. It additionally sets up the necessary SSH stuff inside the container and once its time to run the build script, it boots the image instead with execvp enabled. The ssh connection then succeeds and you get a login prompt. 

I'll probably have to add a chown to the service so that any new files created inside the devserver don't stay owned by root when the devserver exits. There's also the question of configuring the devserver options. Right now these are explicit options set by the user (name, port) but I'm toying with the idea of just taking the directory name mkosi is running in as the devserver name and running the devserver with --private-network so we can use port 22 everywhere and do `ssh <devserver-name> which automatically works because of nss mymachines. 

So for systemd you'd simply do:

```
cd systemd
mkosi link
ssh systemd
```

and end up with a prompt in a systemd devserver image. I'm also currently regenerating the build image every time you start the devserver again. With incremental enabled, this is fast enough (although I really need to switch to btrfs on my machine to make it even faster). The idea is that the devserver itself is ephemeral, maybe making it read-only is a good idea so users don't rely on any changes they make while in the devserver. 

I also added some code to allow building in the devserver when it is running but I haven't tested that yet. We use systemd-run instead of systemd-nspawn and just use all the same environment variables and chdirs from the build script systemd-nspawn invocation. Spawning the devserver itself already sets up all the necessary bind mounts so its as simple as just running the build script.

DaanDeMeyer:
While this approach works, this is just too much added complexity just to work around vscode's lack of systemd-nspawn support. I'll create an issue for their remote extensions to add systemd-nspawn support instead.

PR #484, opened 2020-07-31
ci: add openSUSE

behrmann:
I noticed that Debian has packaged zypper and that openSUSE is missing from the CI.


DaanDeMeyer:
I should look into building tdnf as well. Gotta catch em all at this point.

Issue #483, opened 2020-07-31
Split up mkosi into modules


behrmann:
Right now mkosi weighs in at 5236 lines in a single file. While it is nice to not have to think about where a function is, it is sometimes unwieldy and it will only get more so, if more distributions and or features are added. Right now it is also easy to inadvertently break another distribution while working on some unrelated functionality.

My suggestion would be to break mkosi up into multiple modules. I've started the work on a local branch to see where it could go and a possible split might look like this
```
mkosi
├── cli.py                    # taking the commandline interface
├── distributions             # distribution specific function
│   ├── arch.py
│   ├── centos.py
│   ├── clear.py
│   ├── debian.py
│   ├── fedora.py
│   ├── __init__.py
│   ├── mageia.py
│   ├── openmandriva.py
│   ├── opensuse.py
│   ├── photon.py
│   └── ubuntu.py
├── image.py                  # image creation, mounting
├── __init__.py               # functions used all over the place
├── install.py                # running install and copying files to the image
├── __main__.py               # main()
├── output.py                 # make_* and friends
└── partition.py              # partioning and filesystem creation
```

The mkosi binary would be created via a setuptools entrypoint and the work doesn't have to be done in one go. A first step can be seen in #401 where mkosi is moved to `mkosi/__init__.py` and the `main()` is moved to `mkosi/__main__.py`. The rest could be done step by step. 

The coupling is between the different parts of mkosi is very tight at the moment, mostly because everything is passed via the `args` variable. Splitting up mkosi like this (or similarly, I'm open to any sort of ordering) would be a first step to uncoupling the different parts, so that some interfaces are put into the pipeline. a `Distribution` class that handles the bootstrapping of an installation and advertises its features instead of hardcoding all possible variants. 

Related problems and opportunities to splitting:
- Unifying (mostly) on logging instead of printing, because (in my opinion) only the direct cli code should print to screen right away. Right now most stuff is printed to console via `sys.stderr.write`, although there are a few calls to print here and there. My assumption is, that this was done to flush to screen right away. 
- Adding new distributions or splitting out distributions. setuptools has a feature called namespace packages. In this way distributions could be split out and installed when needed. It would also allow for people to have their own installers without touching mkosi itself, which would make for a nicer development experience.

Possible steps:
1. #401 
2. Splitting out `cli.py` and moving to logging for everything that is not in `cli.py`
[...]

PR #485, opened 2020-07-31
Add missing whitespace and properly indent text blocks

behrmann:
This is a bit of a pet peeve of mine. Large text blocks that are not indented with the rest of the code break my flow when reading. Fortunately, there is `textwrap.dedent`.

The rest is missing empty line for the convention of having two empty lines between functions.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -30,6 +30,7 @@ import tempfile
 import urllib.request
 import uuid
 from subprocess import DEVNULL, PIPE
+from textwrap import dedent
 from typing import (
     IO,
     Any,
@@ -53,7 +54,9 @@ from typing import (
 )
 from types import FrameType
 
+
 __version__ = '5'
+
 
 if sys.version_info < (3, 6):
     sys.exit("Sorry, we need at least Python 3.6.")
@@ -1364,13 +1367,17 @@ def enable_networkd(root: str) -> None:
     os.symlink("../run/systemd/resolve/stub-resolv.conf", os.path.join(root, "etc/resolv.conf"))
 
     with open(os.path.join(root, "etc/systemd/network/all-ethernet.network"), "w") as f:
-        f.write("""\
-[Match]
-Type=ether
+        f.write(
+            dedent(
+                """\
+                [Match]
+                Type=ether
 
-[Network]
-DHCP=yes
-""")
+                [Network]
+                DHCP=yes
+                """
+            )
+        )
 
 
 def enable_networkmanager(root: str) -> None:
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -1364,13 +1367,17 @@ def enable_networkd(root: str) -> None:
     os.symlink("../run/systemd/resolve/stub-resolv.conf", os.path.join(root, "etc/resolv.conf"))
 
     with open(os.path.join(root, "etc/systemd/network/all-ethernet.network"), "w") as f:
-        f.write("""\
-[Match]
-Type=ether
+        f.write(
+            dedent(
```
DaanDeMeyer, line 1371: Nit but this seems like excessive newlining (we already have 5k lines as is :p). I would just do `f.write(dedent(f"""\` and end with `"""))`

DaanDeMeyer: Just to be clear: 

```python
f.write(dedent(f"""\
    [Unit]
"""))
```

behrmann: I could definitely do that, but in the long run I want to convince all of you to use black for auto-formatting, since it integrates nicely into CI (thus preventing tabs vs spaces issues like we had at one point) and optimizes for small diffs, making PRs nicer to read. :)

DaanDeMeyer: Maybe add a helper function then for writing and dedenting? (Can we monkey patch f.write to dedent for us? I think we're in the scenario where we always want dedent)

behrmann: I wouldn't like to monkey patch that. In the medium term I want to switch the `str` that are paths to `pathlib.Path`. I've started the work, but I've run into some problems getting this to type check and there is such an awful lot of those. This would then change
```python
foo = "somefile"
with open(os.path.join(root, foo), "w") as f:
    f.write(
        dedent(
            f"""
            content
            """
        )
    )
```
to
```python
foo = Path("somefile")
(root / foo).write_text(
    dedent(
        f"""
        content
        """
    )
)
```
Since I have my own wrapper for `Path` to wrap some Mkosi's path-related functions, I'll happily add a `.write_text_dedent()` then as well. So the discomfort should not last for too long. :)

DaanDeMeyer: OK, sounds like a good plan. With your focus on code cleanup, I was already expecting a pathlib PR at some point.

behrmann: The road there is somewhat winding and #483 is part of that plan. The argparse machinery seems to be somewhat finicky when trying to get the correct types out, which is why I want to first get a proper interface between the CLI and the rest.

PR #489, opened 2020-08-01
Add bootable builds to CI

DaanDeMeyer:
With all the distros included, we already have way too many builds so it
starts making sense to simply add more steps instead of adding more
options to the build matrix.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -4485,12 +4485,6 @@ def load_args(args: CommandLineArguments) -> CommandLineArguments:
 
         if not {"uefi", "bios"}.issuperset(args.boot_protocols):
             die("Not a valid boot protocol")
-        if "bios" in args.boot_protocols and args.distribution not in (Distribution.fedora,
-                                                                       Distribution.arch,
-                                                                       Distribution.debian,
-                                                                       Distribution.ubuntu,
-                                                                       Distribution.photon):
-            die(f"bios boot not implemented yet for {args.distribution}")
 
         if "uefi" in args.boot_protocols and args.distribution == Distribution.photon:
             die(f"uefi boot not supported for {args.distribution}")
--- a/mkosi
+++ b/mkosi
@@ -4646,6 +4646,9 @@ def load_args(args: CommandLineArguments) -> CommandLineArguments:
     if args.qemu_headless and "console=ttyS0" not in args.kernel_command_line:
         args.kernel_command_line.append("console=ttyS0")
 
+    if args.generated_root() and "bios" in args.boot_protocols:
+        die("Sorry, BIOS cannot be combined with --minimize or squashfs filesystems")
+
     return args
 
 
--- a/mkosi
+++ b/mkosi
@@ -4443,6 +4443,9 @@ def load_args(args: CommandLineArguments) -> CommandLineArguments:
     if args.distribution == Distribution.clear and args.output_format == OutputFormat.gpt_btrfs:
         die("Sorry, Clear Linux does not support btrfs")
 
+    if args.distribution == Distribution.clear and "," in args.boot_protocols:
+        die("Sorry, Clear Linux does not support hybrid BIOS/UEFI images")
+
     find_cache(args)
 
     if args.mirror is None:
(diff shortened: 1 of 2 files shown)


DaanDeMeyer:
So gpt_squashfs (or --minimize) and grub are not good friends. Grub fails with the following error on my machine and in Github Actions:

```
Installing for i386-pc platform.
grub-install: warning: disk does not exist, so falling back to partition device /dev/sda2.
grub-install: warning: disk does not exist, so falling back to partition device /dev/sda2.
grub-install: warning: disk does not exist, so falling back to partition device /dev/sda2.
grub-install: error: disk `hostdisk//dev/sda2' not found.
```

DaanDeMeyer:
My guess was correct, delaying boot loader installation until after the generated root is inserted fixes the issue. One other issue it did introduce was break bootctl install (which I added instead of manually copying the files over) which depends on a machine-id being available. However, by delaying the boot loader installation until after the generated root is inserted, the boot loader installation was also ordered after reset_machine_id which leaves the image without a machine id. This caused bootctl install to fail. I dropped the bootctl change which hopefully fixes that issue.

DaanDeMeyer:
The Clear Linux issue has to do with something missing in /dev I think The squashfs issue I have absolutely no idea. Hopefully strace will give more information as to why these are failing.

DaanDeMeyer:
I also found [this](https://linux.die.net/man/1/expect) which makes me hopeful that we can also go even further than this and actually boot in systemd-nspawn and qemu as well.

DaanDeMeyer:
So squashfs fails because grub-install tries to create /boot/grub but /boot is not a separate mountpoint and squashfs is read only so it fails to create the directory. If I modify grub to install itself to /efi (which is a separate partition), everything works as expected. 

So the crux is that there's a circular dependency between grub and the squashfs root partition. grub-install needs the partition to exist before it can install itself but the squashfs root partition cannot be made until grub is installed to it. 

I still need to test if installing grub to /efi actually makes booting itself work. I have no idea how grub finds its config when booting so this might break horribly. I also wonder if (if this works at all), we should install grub to an XBOOTLDR partition instead of the esp partition. If grub has to be installed on the root partition, I don't immediately see how we're going to make this work.

DaanDeMeyer:
Ok, I got Clear Linux working except for btrfs which isn't supported yet (but should be soon hopefully, see https://github.com/clearlinux/clr-boot-manager/pull/238). 

I added validation to fail on the bios + --minimize/gpt_squashfs combo for now. Let's wait until we get an issue so we know that someone is actually trying to use that combination before we complicate mkosi to add support for it.

DaanDeMeyer:
I added back the commit that delays the boot loader installation until after the generated root partition is inserted into the image. This fixes the Clear Linux + UEFI + gpt_squashfs combination as Clear Linux's bootloader tries to stat the root partition when installing. 

This could in theory break some finalize scripts that depend on the boot loader already being installed. The finalize script can still modify the boot partitions, the boot loader just won't be there yet. I think this is niche enough that we can merge this without causing any serious breakage.

DaanDeMeyer:
Nvm, removed again as getting this to work will require a few bigger changes and this PR is already big enough as it is. I just disabled Clear + UEFI + gpt_squashfs for now. I'll submit an actual fix in a subsequent PR.

behrmann:
Excellent work, @DaanDeMeyer. LGTM.

mkosi
```diff
@@ -4429,6 +4434,18 @@ def load_args(args: CommandLineArguments) -> CommandLineArguments:
         elif args.distribution == Distribution.openmandriva:
             args.release = "cooker"
 
+    if args.distribution in (Distribution.centos, Distribution.centos_epel) and \
+        args.release == "8" and \
+        args.output_format == OutputFormat.gpt_btrfs:
```
behrmann, line 4439: Minor stylistic note: This could be written as 
```python
if (args.distribution in (Distribution.centos, Distribution.centos_epel) and
    args.release == "8" and
    args.output_format == OutputFormat.gpt_btrfs):
```
since the parentheses around the whole condition imply the line breaks, since it's then a single statement. This is, I think more commanly used than explicit line breaks.

DaanDeMeyer:
Cleaned up the git history

behrmann:
@NicolasT Thanks for the ideas! Could you open a new issue? It's easier to track than on a closed PR.

PR #490, opened 2020-08-03
Unified kernel images improvements

DaanDeMeyer:
Doing the initial kernel installs with kernel-install and the rest with mkinitcpio is weird. Let's just use kernel-install all the time. Rest of the commits are simplification of the existing logic.

I also want to add unified kernel images support for Arch but I want to accompany that with a flag to disable unified kernel images (which I really think should be the default but that breaks backwards compat)

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -2322,7 +2322,8 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
         if args.bios_partno:
             packages.add("grub")
 
-        packages.add("mkinitcpio")
+        packages.add("dracut")
+        packages.add("binutils")
 
     def run_pacstrap(packages: Set[str]) -> None:
         cmdline = ["pacstrap", "-C", pacman_conf, "-GM", root]
@@ -2331,37 +2332,6 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
 
     # Set up system with packages from the base group
     run_pacstrap(packages)
-
-    if not do_run_build_script and args.bootable:
-        # Patch mkinitcpio configuration so:
-        # 1) we remove autodetect and
-        # 2) we add the modules needed for encrypt.
-        def jj(line: str) -> str:
-            if line.startswith("HOOKS="):
-                if args.encrypt == "all":
-                    return 'HOOKS="systemd modconf block sd-encrypt filesystems keyboard fsck"\n'
-                else:
-                    return 'HOOKS="systemd modconf block filesystems fsck"\n'
-            return line
-        patch_file(os.path.join(root, "etc/mkinitcpio.conf"), jj)
-
-        # Disable fallback images to save time. We override the preset files
-        # before installing any kernels to have the updated config available
-        # when the mkinitcpio pacman hook is run during installation.
-        presetdir = os.path.join(root, "etc/mkinitcpio.d")
-        for kernel in kernel_packages:
-            preset = os.path.join(presetdir, f"{kernel}.preset")
-            with open(preset, "w") as f:
-                f.write(
-                    dedent(
-                        f"""
-                        ALL_config="/etc/mkinitcpio.conf"
-                        ALL_kver="/boot/vmlinuz-{kernel}"
-                        PRESETS=('default')
-                        default_image="/boot/initramfs-{kernel}.img"
-                        """
-                    )
-                )
 
     # Install the user-specified packages and kernel
     packages = set(args.packages)
@@ -2372,16 +2342,14 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
         packages.update(args.build_packages)
 
     if not do_run_build_script and args.bootable:
-        # Replace mkinitcpio and depmod hooks by kernel-install hooks which additionally creates systemd-boot
-        # loader entries.
         hooks_dir = os.path.join(root, "etc/pacman.d/hooks")
         scripts_dir = os.path.join(root, "etc/pacman.d/scripts")
 
         os.makedirs(hooks_dir, 0o755, exist_ok=True)
         os.makedirs(scripts_dir, 0o755, exist_ok=True)
 
-        for hook in ["60-depmod.hook", "90-mkinitcpio-install.hook"]:
-            os.symlink("/dev/null", os.path.join(hooks_dir, hook))
+        # Disable depmod pacman hook as depmod is handled by kernel-install as well.
+        os.symlink("/dev/null", os.path.join(hooks_dir, "60-depmod.hook"))
 
         kernel_add_hook = os.path.join(hooks_dir, "90-mkosi-kernel-add.hook")
         with open(kernel_add_hook, 'w') as f:
--- a/mkosi
+++ b/mkosi
@@ -2384,10 +2384,6 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
         finally:
             stop_gpg_agent()
 
-    # Remove already installed packages
-    c = run_pacman(['-Qq'], stdout=PIPE, universal_newlines=True)
-    packages.difference_update(c.stdout.split())
-
     if packages:
         # Disable mkinitcpio hook because we manually install the kernel image
         # to /boot using kernel-install in install_boot_loader_arch.
--- a/mkosi
+++ b/mkosi
@@ -2375,7 +2375,6 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
         cmdline = [
             "pacman",
             "--noconfirm",
-            "--color", "never",
             "--config", pacman_conf,
         ]
 
--- a/mkosi
+++ b/mkosi
@@ -1525,7 +1525,7 @@ def disable_kernel_install(args: CommandLineArguments, root: str) -> None:
     # off any kernel installation beforehand.
     #
     # For BIOS mode, we don't have that option, so do not mask the units.
-    if not args.bootable or args.bios_partno is not None:
+    if not args.bootable or args.bios_partno is not None or not args.with_unified_kernel_images:
         return
 
     for d in ("etc", "etc/kernel", "etc/kernel/install.d"):
@@ -1537,7 +1537,7 @@ def disable_kernel_install(args: CommandLineArguments, root: str) -> None:
 
 
 def reenable_kernel_install(args: CommandLineArguments, root: str) -> None:
-    if not args.bootable or args.bios_partno is not None:
+    if not args.bootable or args.bios_partno is not None or not args.with_unified_kernel_images:
         return
 
     hook_path = os.path.join(root, "etc/kernel/install.d/50-mkosi-dracut-unified-kernel.install")
@@ -3298,7 +3298,7 @@ def install_unified_kernel(args: CommandLineArguments,
     # benefit that they can be signed like normal EFI binaries, and can encode everything necessary to boot a
     # specific root device, including the root hash.
 
-    if not args.bootable or args.esp_partno is None:
+    if not args.bootable or args.esp_partno is None or not args.with_unified_kernel_images:
         return
     if for_cache:
         return
@@ -3892,6 +3892,9 @@ def create_parser() -> ArgumentParserMkosi:
                        help='Make use of and generate intermediary cache images')
     group.add_argument('-M', "--minimize", action=BooleanAction,
                        help='Minimize root file system size')
+    group.add_argument("--without-unified-kernel-images", action=BooleanAction,
+                       dest='with_unified_kernel_images', default=True,
+                       help="Do not install unified kernel images")
 
     group = parser.add_argument_group("Packages")
     group.add_argument('-p', "--package", action=CommaDelimitedListAction, dest='packages', default=[],
@@ -4688,6 +4691,13 @@ def load_args(args: CommandLineArguments) -> CommandLineArguments:
     if args.generated_root() and "bios" in args.boot_protocols:
         die("Sorry, BIOS cannot be combined with --minimize or squashfs filesystems")
 
+    if args.with_unified_kernel_images and args.distribution == Distribution.clear:
+        warn("Using --without-unified-kernel-images as Clear Linux does not support unified kernel images.")
+        args.with_unified_kernel_images = False
+
+    if args.verity and not args.with_unified_kernel_images:
+        die("Sorry, --verity can only be used with unified kernel images")
+
     return args
 
 
@@ -4801,6 +4811,7 @@ def print_summary(args: CommandLineArguments) -> None:
                 sys.stderr.write(" UEFI SecureBoot Cert.: " + args.secure_boot_certificate + "\n")
 
             sys.stderr.write("        Boot Protocols: " + line_join_list(args.boot_protocols) + "\n")
+            sys.stderr.write(" Unified Kernel Images: " + yes_no(args.with_unified_kernel_images))
 
     sys.stderr.write("\nPACKAGES:\n")
     sys.stderr.write("              Packages: " + line_join_list(args.packages) + "\n")
--- a/mkosi
+++ b/mkosi
@@ -207,8 +207,8 @@ def die(message: str) -> NoReturn:
     raise MkosiException(message)
 
 
-def warn(message: str, *args: Any, **kwargs: Any) -> None:
-    sys.stderr.write('WARNING: ' + message.format(*args, **kwargs) + '\n')
+def warn(message: str) -> None:
+    sys.stderr.write(f"‣ \033[33;1mWarning: {message}\033[0m\n")
 
 
 def tmp_dir() -> str:
@@ -4326,9 +4326,14 @@ def find_cache(args: CommandLineArguments) -> None:
 def require_private_file(name: str, description: str) -> None:
     mode = os.stat(name).st_mode & 0o777
     if mode & 0o007:
-        warn("Permissions of '{}' of '{}' are too open.\n" +
-             "When creating {} files use an access mode that restricts access to the owner only.",
-             name, oct(mode), description)
+        warn(
+            dedent(
+                f"""
+                Permissions of '{name}' of '{mode:04o}' are too open.
+                When creating {description} files use an access mode that restricts access to the owner only.\
+                """
+            )
+        )
 
 
 def find_passphrase(args: CommandLineArguments) -> None:
--- a/mkosi
+++ b/mkosi
@@ -2646,7 +2646,9 @@ def set_serial_terminal(args: CommandLineArguments, root: str, do_run_build_scri
     if do_run_build_script or for_cache or not args.qemu_headless:
         return
 
-    override_dir = mkdir_last(os.path.join(root, "etc/systemd/system/serial-getty@ttyS0.service.d"))
+    override_dir = os.path.join(root, "etc/systemd/system/serial-getty@ttyS0.service.d")
+    os.makedirs(override_dir, mode=0o755, exist_ok=True)
+
     with open(os.path.join(override_dir, "override.conf"), 'w') as f:
         f.write(
             dedent(
--- a/mkosi
+++ b/mkosi
@@ -2270,77 +2270,6 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
                     )
                 )
 
-    def stop_gpg_agent() -> None:
-        # Kill the gpg-agent used by pacman and pacman-key
-        run(['gpgconf', '--homedir', os.path.join(root, 'etc/pacman.d/gnupg'), '--kill', 'all'])
-
-    def run_pacman_key(args: List[str]) -> CompletedProcess:
-        cmdline = [
-            "pacman-key",
-            "--nocolor",
-            "--config", pacman_conf,
-        ]
-
-        try:
-            return run(cmdline + args, check=True)
-        finally:
-            stop_gpg_agent()
-
-    keyring = "archlinux"
-    if platform.machine() == "aarch64":
-        keyring += "arm"
-    run_pacman_key(["--init"])
-    run_pacman_key(["--populate", keyring])
-
-    # the base metapackage is mandatory
-    packages = {"base"}
-
-    official_kernel_packages = {
-        "linux",
-        "linux-lts",
-        "linux-hardened",
-        "linux-zen",
-    }
-
-    kernel_packages = official_kernel_packages.intersection(args.packages)
-    if len(kernel_packages) > 1:
-        warn('More than one kernel will be installed: {}', ' '.join(kernel_packages))
-
-    if not do_run_build_script and args.bootable:
-        if args.output_format == OutputFormat.gpt_ext4:
-            packages.add("e2fsprogs")
-        elif args.output_format == OutputFormat.gpt_btrfs:
-            packages.add("btrfs-progs")
-        elif args.output_format == OutputFormat.gpt_xfs:
-            packages.add("xfsprogs")
-        if args.encrypt:
-            packages.add("cryptsetup")
-            packages.add("device-mapper")
-        if not kernel_packages:
-            # No user-specified kernel
-            kernel_packages.add("linux")
-        if args.bios_partno:
-            packages.add("grub")
-
-        packages.add("dracut")
-        packages.add("binutils")
-
-    def run_pacstrap(packages: Set[str]) -> None:
-        cmdline = ["pacstrap", "-C", pacman_conf, "-GM", root]
-        # pacstrap handles gpg-agent on its own
-        run(cmdline + list(packages), check=True)
-
-    # Set up system with packages from the base group
-    run_pacstrap(packages)
-
-    # Install the user-specified packages and kernel
-    packages = set(args.packages)
-    if not do_run_build_script and args.bootable:
-        packages |= kernel_packages
-
-    if do_run_build_script:
-        packages.update(args.build_packages)
-
     if not do_run_build_script and args.bootable:
         hooks_dir = os.path.join(root, "etc/pacman.d/hooks")
         scripts_dir = os.path.join(root, "etc/pacman.d/scripts")
@@ -2454,19 +2383,62 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
 
         make_executable(kernel_remove_script)
 
-    def run_pacman(packages: List[str]) -> CompletedProcess:
-        if len(packages) == 0:
-            return
-
-        cmdline = ["pacman", "--noconfirm", "--config", pacman_conf, "-S", *packages]
+    def run_pacman_key(args: List[str]) -> CompletedProcess:
+        cmdline = ["pacman-key", "--nocolor", "--config", pacman_conf]
 
         try:
-            return run(cmdline, check=True)
+            return run(cmdline + args, check=True)
         finally:
-            stop_gpg_agent()
+            # Kill the gpg-agent started by pacman-key
+            run(['gpgconf', '--homedir', os.path.join(root, 'etc/pacman.d/gnupg'), '--kill', 'all'])
 
-    with mount_api_vfs(args, root):
-        run_pacman(packages)
+    keyring = "archlinux"
+    if platform.machine() == "aarch64":
+        keyring += "arm"
+    run_pacman_key(["--init"])
+    run_pacman_key(["--populate", keyring])
+
+    packages = {"base"}
+
+    if not do_run_build_script and args.bootable:
+        if args.output_format == OutputFormat.gpt_ext4:
+            packages.add("e2fsprogs")
+        elif args.output_format == OutputFormat.gpt_btrfs:
+            packages.add("btrfs-progs")
+        elif args.output_format == OutputFormat.gpt_xfs:
+            packages.add("xfsprogs")
+        if args.encrypt:
+            packages.add("cryptsetup")
+            packages.add("device-mapper")
+        if args.bios_partno:
+            packages.add("grub")
+
+        packages.add("dracut")
+        packages.add("binutils")
+
+    packages.update(args.packages)
+
+    official_kernel_packages = {
+        "linux",
+        "linux-lts",
+        "linux-hardened",
+        "linux-zen",
+    }
+
+    has_kernel_package = official_kernel_packages.intersection(args.packages)
+    if not do_run_build_script and args.bootable and not has_kernel_package:
+        # No user-specified kernel
+        packages.add("linux")
+
+    if do_run_build_script:
+        packages.update(args.build_packages)
+
+    def run_pacstrap(packages: Set[str]) -> None:
+        cmdline = ["pacstrap", "-C", pacman_conf, "-GM", root]
+        # pacstrap handles gpg-agent on its own
+        run(cmdline + list(packages), check=True)
+
+    run_pacstrap(packages)
 
     if "networkmanager" in args.packages:
         enable_networkmanager(root)
--- a/mkosi
+++ b/mkosi
@@ -2379,6 +2379,25 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
 
         make_executable(kernel_remove_script)
 
+        if args.esp_partno is not None:
+            bootctl_update_hook = os.path.join(hooks_dir, "91-mkosi-bootctl-update-hook")
+            with open(bootctl_update_hook, 'w') as f:
+                f.write(
+                    dedent(
+                        """\
+                        [Trigger]
+                        Operation = Upgrade
+                        Type = Package
+                        Target = systemd
+
+                        [Action]
+                        Description = Updating systemd-boot...
+                        When = PostTransaction
+                        Exec = /usr/bin/bootctl update
+                        """
+                    )
+                )
+
     keyring = "archlinux"
     if platform.machine() == "aarch64":
         keyring += "arm"
--- a/mkosi
+++ b/mkosi
@@ -78,6 +78,73 @@ MKOSI_COMMANDS_NEED_BUILD = MKOSI_COMMANDS_CMDLINE
 MKOSI_COMMANDS_SUDO = ("build", "clean") + MKOSI_COMMANDS_CMDLINE
 MKOSI_COMMANDS = ("build", "clean", "help", "summary") + MKOSI_COMMANDS_CMDLINE
 
+DRACUT_SYSTEMD_EXTRAS = [
+    "/usr/lib/systemd/systemd-veritysetup",
+    "/usr/lib/systemd/system-generators/systemd-veritysetup-generator",
+    "/usr/bin/systemd-repart",
+    "/usr/lib/systemd/system/systemd-repart.service",
+    "/usr/lib/systemd/system/initrd-root-fs.target.wants/systemd-repart.service",
+]
+
+DRACUT_UNIFIED_KERNEL_INSTALL = """\
+#!/bin/bash -e
+
+COMMAND="$1"
+KERNEL_VERSION="$2"
+BOOT_DIR_ABS="$3"
+KERNEL_IMAGE="$4"
+ROOTHASH="${5:-}"
+
+# If KERNEL_INSTALL_MACHINE_ID is defined but empty, BOOT_DIR_ABS is a fake directory so let's skip creating
+# the unified kernel image.
+if [[ -z "${KERNEL_INSTALL_MACHINE_ID-unset}" ]]; then
+    exit 0
+fi
+
+# Strip machine ID and kernel version to get the boot directory.
+PREFIX=$(dirname $(dirname "$BOOT_DIR_ABS"))
+
+if [[ -n "$ROOTHASH" ]]; then
+    BOOT_BINARY="${PREFIX}/EFI/Linux/linux-${KERNEL_VERSION}-${ROOTHASH}.efi"
+else
+    BOOT_BINARY="${PREFIX}/EFI/Linux/linux-${KERNEL_VERSION}.efi"
+fi
+
+case "$COMMAND" in
+    add)
+        if [[ -f /etc/kernel/cmdline ]]; then
+            read -r -d '' BOOT_OPTIONS < /etc/kernel/cmdline || true
+        elif [[ -f /usr/lib/kernel/cmdline ]]; then
+            read -r -d '' BOOT_OPTIONS < /usr/lib/kernel/cmdline || true
+        else
+            read -r -d '' BOOT_OPTIONS < /proc/cmdline || true
+        fi
+
+        if [[ -n "$ROOTHASH" ]]; then
+            BOOT_OPTIONS="${BOOT_OPTIONS} roothash=${ROOTHASH}"
+        fi
+
+        if [[ -n "$KERNEL_IMAGE" ]]; then
+            DRACUT_KERNEL_IMAGE_OPTION="--kernel-image ${KERNEL_IMAGE}"
+        else
+            DRACUT_KERNEL_IMAGE_OPTION=""
+        fi
+
+        dracut \\
+            --verbose \\
+            --uefi \\
+            --kver "$KERNEL_VERSION" \\
+            $DRACUT_KERNEL_IMAGE_OPTION \\
+            --kernel-cmdline "$BOOT_OPTIONS" \\
+            --force \\
+            "$BOOT_BINARY"
+        ;;
+    remove)
+        rm -f -- "$BOOT_BINARY"
+        ;;
+esac
+"""
+
 
 # This global should be initialized after parsing arguments
 arg_debug = ()
@@ -1329,6 +1396,24 @@ def prepare_tree(args: CommandLineArguments, root: str, do_run_build_script: boo
             cmdline.write(' '.join(args.kernel_command_line))
             cmdline.write("\n")
 
+        dracut_dir = os.path.join(root, "etc/dracut.conf.d")
+        os.mkdir(dracut_dir, 0o755)
+
+        with open(os.path.join(dracut_dir, "30-mkosi-hostonly.conf"), "w") as f:
+            f.write("hostonly=no\n")
+
+        with open(os.path.join(dracut_dir, "30-mkosi-systemd-extras.conf"), "w") as f:
+            for extra in DRACUT_SYSTEMD_EXTRAS:
+                f.write(f"install_optional_items+=\" {extra} \"\n")
+
+        with open(os.path.join(dracut_dir, "30-mkosi-qemu.conf"), "w") as f:
+            f.write("add_dracutmodules+=\" qemu \"\n")
+
+        # These distros need uefi_stub configured explicitly for dracut to find the systemd-boot uefi stub.
+        if args.distribution in (Distribution.ubuntu, Distribution.mageia, Distribution.openmandriva):
+            with open(os.path.join(dracut_dir, "30-mkosi-uefi-stub.conf"), "w") as f:
+                f.write("uefi_stub=/usr/lib/systemd/boot/efi/linuxx64.efi.stub\n")
+
     if do_run_build_script:
         os.mkdir(os.path.join(root, "root"), 0o750)
         os.mkdir(os.path.join(root, "root/dest"), 0o755)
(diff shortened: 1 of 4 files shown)


DaanDeMeyer:
Switched arch to dracut as well. Also had a go at a kernel-install script that does the whole unified kernel shenanigans so it's actually possible to update the kernel when using a mkosi image. Not entirely sure yet how to handle the default dracut kernel-install scripts. They're currently re-enabled for other distros but I think if we're using unified kernel images we might as well force every kernel to be a unified kernel image. Next step is to add a switch that allows disabling unified kernels. The only difference we'll have then is that without that option the unified kernel install script is used and with that option the default dracut scripts are used.

mkosi
```diff
@@ -79,6 +79,70 @@ MKOSI_COMMANDS_SUDO = ("build", "clean") + MKOSI_COMMANDS_CMDLINE
...
+    # If KERNEL_INSTALL_MACHINE_ID is defined but empty, BOOT_DIR_ABS is a fake directory.
+    # So, let's skip to create initrd.
+    if ! [[ ${KERNEL_INSTALL_MACHINE_ID-x} ]]; then
+        exit 0
+    fi
+
+    if [[ -d "$BOOT_DIR_ABS" ]]; then
+        INITRD="initrd"
+    else
+        BOOT_DIR_ABS="/boot"
```
behrmann, line 100: This variable is not used after being assigned here.

mkosi
```diff
@@ -2263,76 +2323,85 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
...
+        kernel_install_hook = os.path.join(hooks_dir, "kernel-install.hook")
+        with open(kernel_install_hook, 'w') as f:
+            f.write(
+                dedent(
+                    f"""
+                    [Trigger]
+                    Operation = Install
+                    Operation = Upgrade
+                    Type = File
+                    Target = usr/lib/modules/*/vmlinuz
```
behrmann, line 2351: The hook should probably also run on other Targets as well, e.g. `boot/intel-ucode.img` or dracut being upgraded itself.

mkosi
```diff
@@ -2263,76 +2323,85 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
...
+        os.symlink("/dev/null", os.path.join(hooks_dir, "60-depmod.hook"))
 
-    # Remove already installed packages
-    c = run_pacman(['-Qq'], stdout=PIPE, universal_newlines=True)
-    packages.difference_update(c.stdout.split())
+        kernel_install_hook = os.path.join(hooks_dir, "kernel-install.hook")
+        with open(kernel_install_hook, 'w') as f:
+            f.write(
+                dedent(
+                    f"""
```
behrmann, line 2346: Missing a line continuation.

mkosi
```diff
@@ -2263,76 +2323,85 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
...
+                    [Trigger]
+                    Operation = Install
+                    Operation = Upgrade
+                    Type = File
+                    Target = usr/lib/modules/*/vmlinuz
+
+                    [Action]
+                    Description = Adding kernel and initramfs images to /boot...
+                    When = PostTransaction
+                    Exec = /bin/bash -c 'while read f; do kernel-install add $(basename $(dirname $f)) $f; done'
```
behrmann, line 2356:
```suggestion
                    Exec = /bin/bash -c 'while read -r f; do kernel-install add $(basename $(dirname $f)) $f; done'
```

mkosi
```diff
@@ -2263,76 +2323,85 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
...
+                    Depends = systemd
+                    Depends = dracut
+                    Depends = bash
+                    Depends = coreutils
```
behrmann, line 2360: The depends on systemd, bash and coreutils is not strictly necessary, as they are dependencies of the base package and Arch only supports systems running base.

mkosi
```diff
@@ -2263,76 +2323,85 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
...
+                    [Trigger]
+                    Operation = Upgrade
+                    Operation = Remove
+                    Type = File
+                    Target = usr/lib/modules/*/vmlinuz
+
+                    [Action]
+                    Description = Removing kernel and initramfs images from /boot...
+                    When = PreTransaction
+                    Exec = /bin/bash -c 'while read f; do kernel-install remove $(basename $(dirname $f)); done'
```
behrmann, line 2380:
```suggestion
                    Exec = /bin/bash -c 'while read -r f; do kernel-install remove $(basename $(dirname $f)); done'
```

mkosi
```diff
@@ -2263,76 +2323,85 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
...
+                    Depends = systemd
+                    Depends = dracut
+                    Depends = bash
+                    Depends = coreutils
```
behrmann, line 2384: Same here for the systemd, bash and coreutils dependency.

behrmann:
@DaanDeMeyer Good job! Moving arch to dracut was also on my radar :)

mkosi
behrmann: Of course, if this hook is run on more targets (e.g. if run on microcode or dracut upgrade), than this would have to moved out into an extra script anyway, because the kernel version could no longer be determined from the targets.

behrmann:
While at it, we should also install a hook like this for bootable EFI images
```
[Trigger]
Type = Package
Operation = Upgrade
Target = systemd

[Action]
Description = Updating systemd-boot
When = PostTransaction
Exec = /usr/bin/bootctl update
```

mkosi
```diff
@@ -79,6 +79,70 @@ MKOSI_COMMANDS_SUDO = ("build", "clean") + MKOSI_COMMANDS_CMDLINE
...
+    #!/bin/bash
+
+    COMMAND="$1"
+    KERNEL_VERSION="$2"
+    BOOT_DIR_ABS="$3"
+    KERNEL_IMAGE="$4"
+
+    # If KERNEL_INSTALL_MACHINE_ID is defined but empty, BOOT_DIR_ABS is a fake directory.
+    # So, let's skip to create initrd.
+    if ! [[ ${KERNEL_INSTALL_MACHINE_ID-x} ]]; then
```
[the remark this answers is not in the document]
DaanDeMeyer: I adapted this from the upstream dracut-50.install script. Not sure either.

mkosi
DaanDeMeyer: It's kinda annoying that there's this small mismatch between the kernel-install interface and the pacman hook interface. For now, this works well enough. When we add the rest, we just add separate pacman hooks that call kernel-install internally.

mkosi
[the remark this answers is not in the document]
behrmann: TIL a new scary Bash feature, `${VAR-default}` is the same as `${VAR:-default}`.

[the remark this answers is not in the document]
behrmann: Missed that while skimming the [Bash manual](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) this morning: 
> Omitting the colon results in a test only for a parameter that is unset. Put another way, if the colon is included, the operator tests for both parameter’s existence and that its value is not null; if the colon is omitted, the operator tests only for existence. 

I'll keep that footgun in mind in the future.

mkosi
```diff
@@ -2263,76 +2323,85 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
...
+        kernel_remove_hook = os.path.join(hooks_dir, "kernel-remove.hook")
+        with open(kernel_remove_hook, 'w') as f:
+            f.write(
+                dedent(
+                    f"""
+                    [Trigger]
+                    Operation = Upgrade
+                    Operation = Remove
+                    Type = File
+                    Target = usr/lib/modules/*/vmlinuz
```
behrmann, line 2375: Just to start a list on which packages we should rebuild the initrd on: linux-*, systemd, dracut, intel-ucode, amd-ucode, lvm2, device-mapper, cryptsetup. I probably forget all the RAID packages I never use, but since mkosi doesn't target that, we can probably do without them (which would also be true for LVM).

mkosi
```diff
@@ -79,6 +79,70 @@ MKOSI_COMMANDS_SUDO = ("build", "clean") + MKOSI_COMMANDS_CMDLINE
...
+    ret=0
+    case "$COMMAND" in
+        add)
+            if [[ -f /etc/kernel/cmdline ]]; then
+                read -r -d '' BOOT_OPTIONS < /etc/kernel/cmdline
+            elif [[ -f /usr/lib/kernel/cmdline ]]; then
+                read -r -d '' BOOT_OPTIONS < /usr/lib/kernel/cmdline
+            else
+                read -r -d '' BOOT_OPTIONS < /proc/cmdline
+            fi
```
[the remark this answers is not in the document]
DaanDeMeyer: This is adapted from its upstream kernel-install script so I guess it doesn't.

DaanDeMeyer:
Had to do a hack to allow reusing the dracut kernel install script in install_unified_kernel since kernel-install requires a machine-id to work correctly. However, it only reads from /etc/machine-id if it exists so if we make sure it doesn't exist and set the MACHINE_ID environment variable, kernel-install picks it up and installs the unified kernel image as required.

The only annoying thing left is that the depmod kernel-install hook spams errors in install_unified_kernel because the filesystem is read only. Maybe that should be modified in systemd to only do something on non read-only filesystems. We can also override the script ourselves until that's fixed.

DaanDeMeyer:
OK, hack is gone by just invoking the script directly instead of going via kernel-install in install_unified_kernel. Added advantage is depmod error spam is gone as well.

mkosi
DaanDeMeyer: Does it need to depend on dracut actually? It's a bit more complicated with the custom dracut hook we add but in general I don't think we need to add dracut here either since it installs its own kernel-install script.

mkosi
DaanDeMeyer: So we'd make a separate pacman hook and in the hook check whether there's paths that don't end with vmlinuz. If there are, rebuild everything, if not, loop over the given paths and only run kernel-install on those?

behrmann: I don't think a separate hook, but instead of this hook calling kernel-install calling that script instead. Apart from that, yes, loop over the input to see whether it's a kernel package or something else and on that basis rebuild the kernel image.

mkosi
```diff
@@ -2266,76 +2298,75 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
...
-                        default_image="/boot/initramfs-{kernel}.img"
-                        """
-                    )
-                )
-
     # Install the user-specified packages and kernel
     packages = set(args.packages)
-    if not do_run_build_script and args.bootable:
-        packages |= kernel_packages
+    if not do_run_build_script and args.bootable and not any(["linux" in package for package in args.packages]):
```
behrmann, line 2303:
```suggestion
    if not do_run_build_script and args.bootable and not any("linux" in package for package in args.packages):
```
to not generate the whole list unnecessarily or
```suggestion
    if not do_run_build_script and args.bootable and not any(package.startswith("linux") for package in args.packages):
```
to better check that it's a kernel package as well, because otherwise `util-linux` would already trigger the `any`.

Thinking about this, I guess that the check against `official_kernel_packages` should maybe remain, because there are some packages that are not the kernel that have linux in their name or might even start with linux (e.g. linuxconsole), which could lead to images being created without a kernel.

mkosi
behrmann: Ah, and of course this should be in the upgrade hook instead of this one.

behrmann: This should of course be in the add Installation hook instead of the Remova hook.

mkosi
```diff
@@ -79,6 +79,73 @@ MKOSI_COMMANDS_SUDO = ("build", "clean") + MKOSI_COMMANDS_CMDLINE
...
+        if [[ -n "$KERNEL_IMAGE" ]]; then
+            DRACUT_KERNEL_IMAGE_OPTION="--kernel-image ${KERNEL_IMAGE}"
+        else
+            DRACUT_KERNEL_IMAGE_OPTION=""
+        fi
+
+        dracut \\
+            --quiet \\
+            --force \\
+            --no-hostonly \\
```
behrmann, line 129: Since this script replaces the normal dracut kernel-install plugins, it might make sense to also have a replacement for the one that generates the rescue image and have the `--no-hostonly` only in that one.

mkosi
```diff
@@ -79,6 +79,73 @@ MKOSI_COMMANDS_SUDO = ("build", "clean") + MKOSI_COMMANDS_CMDLINE
...
 
 
+DRACUT_UNIFIED_KERNEL_INSTALL = """\
+#!/bin/bash -e
+
+COMMAND="$1"
+KERNEL_VERSION="$2"
+BOOT_DIR_ABS="$3"
+KERNEL_IMAGE="$4"
+ROOTHASH="${5:-}"
```
behrmann, line 89: I have only checked the `kernel-install` man page and it only lists four arguments; will the fifth be passed through?

DaanDeMeyer: The ROOTHASH is only there so we can reuse the script in `install_unified_kernel` where we invoke it directly with the extra roothash parameter. In general kernel-install operation, it won't get used although kernel-install also passes extra arguments to the scripts it invokes so in theory we could pass calculate the roothash in the pacman hook and pass it to kernel-install as well. However, roothash is for dm-verity in which case the rootfs is readonly so passing it via the hook isn't really useful. It's still useful for reusing the logic in `install_unified_kernel` though.

mkosi
DaanDeMeyer: mkosi is a tool for making disk images though. I don't think we should limit ourselves to the system we're currently running on since the image might be deployed on a different device.

mkosi
DaanDeMeyer: You're right, this is actually broken. I'll add back the official kernel packages.

DaanDeMeyer:
OK, I cleaned this up a lot (including the git history) and added the kernel-add, kernel-remove scripts which are invoked by the pacman hooks. Can you check the hooks to see if there aren't any more packages/paths/other stuff missing?

Also removed the pacman call from install_arch. We just do a single pacstrap now.

mkosi
DaanDeMeyer: What we should do though is add a hostonly option to configure this. At thesame time we should also move whatever we can out of the install script and into dracut config files.

behrmann: > mkosi is a tool for making disk images though. I don't think we should limit ourselves to the system we're currently running on since the image might be deployed on a different device.

That's true, my thought, though, was that the files stay around. If those images are used longer, it would make sense to install them in generic mode (`--no-hostonly`), but have both lying around for future kernel upgrades.

behrmann:
> Can you check the hooks to see if there aren't any more packages/paths/other stuff missing?

I'll look at it tomorrow.

mkosi
```diff
@@ -2204,42 +2249,125 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
...
+                    for i in "${!TARGETS[@]}"; do
+                        [[ ${TARGETS[$i]} != */vmlinuz ]] && ALL=1
+                        TARGETS[$i]="/${TARGETS[$i]}"
+                    done
+
+                    [[ "$ALL" ]] && TARGETS=(/usr/lib/modules/*/vmlinuz)
+
+                    for f in "${TARGETS[@]}"; do
+                        kernel-install add $(basename $(dirname $f)) $f
+                    done
```
behrmann, line 2311: Thinking about it, this works, but, as bash is, it's a bit noisy. On my personal, I have something similar to this, which is a bit more verbose, but hopefully has less line noise
```bash
#!/bin/bash -e                                                                                                   
                                                                                                                 
declare kernel_version                                                                                           
                                                                                                                 
while read -r line                                                                                               
do                                                                                                               
    if [[ "$line" =~ usr/lib/modules/([^/]+)/vmlinuz ]]                                                          
    then                                                                                                         
        kernel_version="${BASH_REMATCH[1]}"                                                                      
        break                                                                                                    
    fi                                                                                                           
done                                                                                                             
                                                                                                                 
                                                                                                                 
if [[ -n "$kernel_version" ]]                                                                                    
then                                                                                                             
    kernel-install add "$kernel_version" "/usr/lib/${kernel_version}/vmlinuz"                                    
else                                                                                                             
    for f in /usr/lib/modules/*/vmlinuz                                                                          
    do                                                                                                           
        kernel-install add $(basename $(dirname $f)) $f                                                          
    done                                                                                                         
fi       
```

mkosi
```diff
@@ -2250,92 +2378,35 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
...
+        "linux-lts",
+        "linux-hardened",
+        "linux-zen",
+    }
 
-    # Install the user-specified packages and kernel
-    packages = set(args.packages)
-    if not do_run_build_script and args.bootable:
-        packages |= kernel_packages
+    has_kernel_package = any(package in official_kernel_packages for package in args.packages)
```
behrmann, line 2396: for the usage below this can be simplified to
```suggestion
    has_kernel_package = official_kernel_packages.intersection(args.packages)
```

mkosi
```diff
@@ -4646,6 +4611,13 @@ def load_args(args: CommandLineArguments) -> CommandLineArguments:
     if args.generated_root() and "bios" in args.boot_protocols:
         die("Sorry, BIOS cannot be combined with --minimize or squashfs filesystems")
 
+    if args.with_unified_kernel_images and args.distribution == Distribution.clear:
+        warn("Enabling --without-unified-kernel-images as Clear Linux does not support unified kernel images.")
```
behrmann, line 4615:
```suggestion
        warn("Using --without-unified-kernel-images as Clear Linux does not support unified kernel images.")
```
since it's actually disabling something

mkosi
```diff
@@ -4720,7 +4692,7 @@ def print_summary(args: CommandLineArguments) -> None:
         sys.stderr.write("          Repositories: " + ','.join(args.repositories) + "\n")
     sys.stderr.write("\nOUTPUT:\n")
     if args.hostname:
-        sys.stderr.write("              Hostname: " + args.hostname + "\n")
+        sys.stderr.write("                  Hostname: " + args.hostname + "\n")
```
behrmann, line 4695: This whitespace moves the colon out of the line, doesn't it?

mkosi
```diff
@@ -4759,20 +4731,21 @@ def print_summary(args: CommandLineArguments) -> None:
...
 
             sys.stderr.write("        Boot Protocols: " + line_join_list(args.boot_protocols) + "\n")
+            sys.stderr.write(" Unified Kernel Images: " + yes_no(args.with_unified_kernel_images))
 
     sys.stderr.write("\nPACKAGES:\n")
     sys.stderr.write("              Packages: " + line_join_list(args.packages) + "\n")
 
     if args.distribution in (Distribution.fedora, Distribution.centos, Distribution.centos_epel, Distribution.mageia):
-        sys.stderr.write("    With Documentation: " + yes_no(args.with_docs) + "\n")
+        sys.stderr.write("   With Documentation: " + yes_no(args.with_docs) + "\n")
```
behrmann, line 4740: This also seems to get the colon out of line.

mkosi
```diff
@@ -4792,7 +4765,7 @@ def print_summary(args: CommandLineArguments) -> None:
         if "uefi" in args.boot_protocols:
             sys.stderr.write("                   ESP: " + format_bytes_or_disabled(args.esp_size) + "\n")
         if "bios" in args.boot_protocols:
-            sys.stderr.write("                  BIOS: " + format_bytes_or_disabled(BIOS_PARTITION_SIZE) + "\n")
+            sys.stderr.write("              BIOS: " + format_bytes_or_disabled(BIOS_PARTITION_SIZE) + "\n")
```
behrmann, line 4768: This also moves the colon out of line.

mkosi
```diff
@@ -1428,40 +1491,40 @@ def check_if_url_exists(url: str) -> bool:
...
 
 
-def disable_kernel_install(args: CommandLineArguments, root: str) -> List[str]:
-    # Let's disable the automatic kernel installation done by the
-    # kernel RPMs. After all, we want to built our own unified kernels
-    # that include the root hash in the kernel command line and can be
-    # signed as a single EFI executable. Since the root hash is only
-    # known when the root file system is finalized we turn off any
-    # kernel installation beforehand.
+def make_executable(path: str) -> None:
```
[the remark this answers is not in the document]
DaanDeMeyer: Hmm, I'm not convinced on this one. Imo this is clear enough for our purposes.

mkosi
DaanDeMeyer: Does this work when 2 kernels out of 5 (for example) get upgraded?

mkosi
DaanDeMeyer: I had no idea lists/sets could evaluate to bool. Nice!

mkosi
DaanDeMeyer: Ugh my bad I had to indent these back again after the option name changed and it seems I forgot some.

DaanDeMeyer:
Clear Linux builds now but it doesn't actually boot. I spent last evening trying different stuff in order to fix it but didn't manage to find a solution. It's broken on master as well so I say we leave it as is for now until I get to implementing the boot tests in CI.

mkosi
behrmann: No, this works under the assumption that the update will be either a kernel package or one of the others. This can be changed to
```bash
#!/bin/bash -e

declare -a kernel_version

# Check lines we are passed from the pacman hook
while read -r line
do
    if [[ "$line" =~ usr/lib/modules/([^/]+)/vmlinuz ]]
    then
        kernel_version+=("${BASH_REMATCH[1]}")
    fi
done

# If kernel_version is empty, the hook was triggered by an update, that was not
# a kernel package
if [[ "${#kernel_version[@]}" -eq 0 ]]
then
    for f in /usr/lib/modules/*/vmlinuz
    do
        kernel_version+=( $(basename $(dirname $f)) )
    done
fi
                                                                                                                                                                                                                                   
# (re)build the kernel images
for kv in "${kernel_version[@]}"
do
    kernel-install add "$kv" "/usr/lib/modules/${kv}/vmlinuz"
done
```
which I think is still more readable, but also more verbose (although it can be compressed, I like the idiosyncratic linebreaks in the syntax that are usually omitted via a semicolon)

behrmann: All of these scripts also have in common, that they would break, if somebody ever got the idea of having a space in `/usr/lib/modules/*/vmlinuz`.

mkosi
behrmann: Every class that implements `__bool__` can be used in boolean contextes. All the standard implement it such, that they are False, when they are empty and True, when they are not. That's also why the empty string is False (also the empty tuple).

behrmann:
> Clear Linux builds now but it doesn't actually boot. I spent last evening trying different stuff in order to fix it but didn't manage to find a solution. It's broken on master as well so I say we leave it as is for now until I get to implementing the boot tests in CI.

Sounds reasonable. Checking for the ability to boot is also an excellent idea, I wondered about, whether we should do it.

DaanDeMeyer:
Added one more small commit to drop the dependency on pacstrap.

mkosi
DaanDeMeyer: I'm still not sure this does exactly what we want. With this version, if dracut and a kernel are updated at the same time, only a single kernel will be rebuilt no?

Looking again at this I do see that my ALL boolean doesn't work. I pushed a new version with that addressed.

behrmann: Ah, true. Missed that edge case.
```bash
#!/bin/bash -e

declare -a kernel_version

# Check lines we are passed from the pacman hook
while read -r line
do
    if [[ "$line" =~ usr/lib/modules/([^/]+)/vmlinuz ]]
    then
        kernel_version+=( "${BASH_REMATCH[1]}" )
    else
        # If a non-matching line is passed, just rebuild all kernels
        kernel_version=()
        for f in /usr/lib/modules/*/vmlinuz
        do
            kernel_version+=( "$(basename "$(dirname "$f")")" )
        done
        break
    fi
done

# (re)build the kernel images
for kv in "${kernel_version[@]}"
do
    kernel-install add "$kv" "/usr/lib/modules/${kv}/vmlinuz"
done
```
It's unfortunate, that there is nothing cleaner than die basename dirname combo. 

While I like mine better (less sigil noise), as this point it's in the eye of the beholder. :)

behrmann:
> Added one more small commit to drop the dependency on pacstrap.

I checked pacstrap and this change makes sense.

mkosi
DaanDeMeyer: I'll honor all the effort you put in by picking your solution. I do agree that it's easier to read without all the sigil noise.

mkosi
```diff
@@ -2204,42 +2250,114 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
...
+                    [Trigger]
+                    Operation = Upgrade
+                    Operation = Remove
+                    Type = File
+                    Target = usr/lib/modules/*/vmlinuz
+
+                    [Action]
+                    Description = Removing kernel and initramfs images from /boot...
+                    When = PreTransaction
+                    Exec = /bin/bash -c 'while read -r f; do kernel-install remove $(basename $(dirname $f)); done'
```
behrmann, line 2333: Shouldn't this call `kernel_remove_script`?

mkosi
behrmann: No worries, a couple fewer edge cases for my work machine. :) (Not that it matters a lot, since it always ever has a single kernel.)

mkosi
DaanDeMeyer: Ouch! Good catch!

DaanDeMeyer:
- Switched to @behrmann kernel-add script
- Moved dracut cli args to dracut config dropins where possible
- Prefixed all new configuration files with mkosi- so users know where they come from

mkosi
```diff
@@ -4285,9 +4264,12 @@ def find_cache(args: CommandLineArguments) -> None:
 def require_private_file(name: str, description: str) -> None:
     mode = os.stat(name).st_mode & 0o777
     if mode & 0o007:
-        warn("Permissions of '{}' of '{}' are too open.\n" +
-             "When creating {} files use an access mode that restricts access to the owner only.",
-             name, oct(mode), description)
+        warn(
+            f"""\
+            Permissions of '{name}' of '{oct(mode)}' are too open.
```
behrmann, line 4269:
```suggestion
            Permissions of '{name}' of '{mode:04o}' are too open.
```
It's a change, but it will print proper four-digit permissions without the `0b` prefix.

behrmann:
>     * Moved dracut cli args to dracut config dropins where possible
>     * Prefixed all new configuration files with mkosi- so users know where they come from

Good idea. I'll have a look at the changes tomorrow, with fresher eyes.

mkosi
```diff
@@ -1329,6 +1391,18 @@ def prepare_tree(args: CommandLineArguments, root: str, do_run_build_script: boo
...
             cmdline.write("\n")
 
+        dracut_dir = os.path.join(root, "etc/dracut.conf.d")
+        os.mkdir(dracut_dir, 0o755)
+
+        with open(os.path.join(dracut_dir, "mkosi-hostonly.conf"), "w") as f:
+            f.write("hostonly=no\n")
+
+        with open(os.path.join(dracut_dir, "mkosi-systemd-extras.conf"), "w") as f:
+            f.write(f"install_optional_items+={' '.join(DRACUT_SYSTEMD_EXTRAS)}\n")
```
behrmann, line 1401:
```suggestion
            f.write(f'install_optional_items+=" {" ".join(DRACUT_SYSTEMD_EXTRAS)} "\n')
```
dracut wants spaces at the start and end of these strings, since they use them as arrays and I guess they want to be save about splitting elements. 

Alternatively this could be written as
```python
for arg in DRACUT_SYSTEMD_EXTRAS:
    f.write(f'install_optional_items+=" {arg} "\n')
```
To make it easier to read, because the line is over 270 characters long.

mkosi
```diff
@@ -2170,7 +2226,6 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
                 HookDir     = {root}/etc/pacman.d/hooks/
                 HoldPkg     = pacman glibc
                 Architecture = auto
-                UseSyslog
```
behrmann, line 2173: Shouldn't we keep this around? Without this there are no logs in the journal of what pacman does.

mkosi
```diff
@@ -2204,42 +2259,125 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
...
+        os.makedirs(scripts_dir, 0o755, exist_ok=True)
 
-        try:
-            return run(cmdline + args, check=True)
-        finally:
-            stop_gpg_agent()
+        # Disable depmod pacman hook as depmod is handled by kernel-install as well.
+        os.symlink("/dev/null", os.path.join(hooks_dir, "60-depmod.hook"))
+
+        kernel_add_hook = os.path.join(hooks_dir, "mkosi-kernel-add.hook")
```
behrmann, line 2272: This hook should also get a numerical prefix, e.g. `90-`, since this is the number of the appropriate mkinitcpio hook.

mkosi
```diff
@@ -2204,42 +2259,125 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
...
+                    do
+                        kernel-install add "$kv" "/usr/lib/modules/${kv}/vmlinuz"
+                    done
+                    """
+                )
+            )
+
+        make_executable(kernel_add_script)
+
+        kernel_remove_hook = os.path.join(hooks_dir, "mkosi-kernel-remove.hook")
```
behrmann, line 2339: This hook should also get a numerical prefix, e.g. `60-`, since this is the number of the appropriate mkinitcpio hook.

mkosi
```diff
@@ -1329,6 +1391,18 @@ def prepare_tree(args: CommandLineArguments, root: str, do_run_build_script: boo
...
+        os.mkdir(dracut_dir, 0o755)
+
+        with open(os.path.join(dracut_dir, "mkosi-hostonly.conf"), "w") as f:
+            f.write("hostonly=no\n")
+
+        with open(os.path.join(dracut_dir, "mkosi-systemd-extras.conf"), "w") as f:
+            f.write(f"install_optional_items+={' '.join(DRACUT_SYSTEMD_EXTRAS)}\n")
+
+        with open(os.path.join(dracut_dir, "mkosi-uefi-stub.conf"), "w") as f:
+            f.write("uefi_stub=/usr/lib/systemd/boot/efi/linuxx64.efi.stub\n")
```
behrmann, line 1404: I'm not sure this line is necessary, dracut defaults to `/lib/systemd/boot/efi/linuxx64.efi.stub` which should always be appropriately linked.

behrmann:
It would also still make sense to add a hook run `bootctl upgrade` on systemd upgrade for Arch.

mkosi
```diff
@@ -2204,42 +2259,125 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
...
+
+        kernel_remove_hook = os.path.join(hooks_dir, "mkosi-kernel-remove.hook")
+        with open(kernel_remove_hook, 'w') as f:
+            f.write(
+                dedent(
+                    """\
+                    [Trigger]
+                    Operation = Upgrade
+                    Operation = Remove
+                    Type = File
```
behrmann, line 2347:
```suggestion
                    Type = Path
```
File has been deprecated in favour of Path.

mkosi
```diff
@@ -1329,6 +1391,18 @@ def prepare_tree(args: CommandLineArguments, root: str, do_run_build_script: boo
             cmdline.write(' '.join(args.kernel_command_line))
             cmdline.write("\n")
 
+        dracut_dir = os.path.join(root, "etc/dracut.conf.d")
```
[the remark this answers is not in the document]
DaanDeMeyer: I'm not entirely sure we want to do this. I think pretty much all stuff in /usr comes from the package manager (or at least I think that's how systemd wants it to be). If we do it like this, users can trust that whatever's in /usr corresponds to the list of installed packages without mkosi putting extra stuff in there.

@poettering @keszybz Opinions? (They might not reply as I think they're a bit stretched thin on systemd itself in which case I think we should stick to /etc until we can get some more opinions on this)

mkosi
DaanDeMeyer: pacman logs everything at log level WARNING which I think is spammy and not very useful. It also logs under the pacman identifier which is unintuitive since when looking back later one might confuse it with pacman running on the host system. Finally, with systemd v246, stuff from the systemd binaries invoked by the pacman hooks is correctly logged to stdout by pacman which makes me think that its stdout output is logging enough.

mkosi
DaanDeMeyer: I had one distro that failed without this. I'm not sure which one it was, I'll remove the line and let CI run to figure out which distro so we can only add the file on that distro with a comment so it's clearer why we do this.

mkosi
DaanDeMeyer: Woops I fixed this in one hook but forgot to fix it in the other one.

DaanDeMeyer:
@behrmann Thanks for the extended and detailed reviews on this PR!

mkosi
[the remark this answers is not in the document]
behrmann: My take is that logging things to the journal doesn't hurt, but I'm fine either way.

behrmann:
So, Ubuntu, Openmandriva and Mageia seem to need an explicit EFI stub setting, weird. Tthe Arch directory one should be unrelated.

mkosi
DaanDeMeyer: Maybe we could do image generation with --no-hostonly and the configure --hostonly afterwards so subsequent updates are faster since by then the image is likely deployed. Implementing that would be a bit awkward though. For unified kernels we'd have to pass an extra parameter to the dracut script and for normal kernel install we'd have to write a temporary dracut config file to override hostonly.

DaanDeMeyer:
Added one more fix that adds another dracut config file to enable the dracut qemu module. Without it, booting in a GUI actually doesn't work, we never get to the login prompt. (--qemu-headless works fine though)

I think its important stuff like this works by default. We can always add an option to configure the dracut modules included in the future or existing mechanisms can be used to the same effect (although we might need to reorder operations a bit to actually support that).

DaanDeMeyer:
Using 00- prefix isn't a great idea as we still want to override distribution dracut config in /usr/lib/dracut.conf.d (I actually lost an hour because --verity was broken because a 01 distro config on Fedora was overriding hostonly). I switched to 30- which seems like a good compromise between overriding distribution defaults and allowing room for further configuration.

behrmann:
> I think its important stuff like this works by default. We can always add an option to configure the dracut modules included in the future or existing mechanisms can be used to the same effect (although we might need to reorder operations a bit to actually support that).

I think configuring this via the current mechanism to get trees into the image should be the way to target this.

> Using 00- prefix isn't a great idea as we still want to override distribution dracut config in /usr/lib/dracut.conf.d (I actually lost an hour because --verity was broken because a 01 distro config on Fedora was overriding hostonly). I switched to 30- which seems like a good compromise between overriding distribution defaults and allowing room for further configuration.

Ah, sorry about that. I didn't give that enough thought. The bane of manual ordering. 30 should hopefully be fine.

behrmann:
Let's wait for CI, but this should be good to go. The path for where we install dracut files can still be changed in the future.

DaanDeMeyer:
I might focus on more CI next since even with what we have now I still found bugs with build script and verity that weren't caught by CI.

behrmann:
Excellent idea, I started separating out the CLI stuff from the rest, based on #401. This should give some indication whether this is worthwhile, because rebasing #401 on every new commit gets old with time :)

johannbg:
@DaanDeMeyer @behrmann tell me you did not just unilaterally decided to make mkosi always use unified kernel images thus changing the behaviour of mkosi from installing kernels with a separate initrd and boot loader config with this commit?

behrmann:
@johannbg When I reviewed this, my intention was not to change any defaults, with the exception of seeing Arch switch to dracut. I'm sorry, that this seems to have inadvertently happened. 

I've combed through the diffs again and the pertinent changes are:
- I missed that Fedora and Centos always called `install_grub` unconditionally, which has changed now, but does not seem to be your issue.
- `disable_kernel_install` runs under the same circumstances as before (it gained `args.with_unified_kernel_images`, which is `True` by default, but it is paired with `not` thus not changing the truth table of the initial return in `disable_kernel_install`) and the same files are disabled.
- `reenable_kernel_install` has changed in that it does not reenable the disabled files anymore, if the system is to be booted in pure EFI mode and does not opt out of unified kernel images, thus disabling separate initrd for future kernel upgrades for the case where the system is initially prepared with a unified kernel image.

Do I understand correctly that the latter point is what is affecting you or am I missing something (yet again)? Sorry for any grief that my oversight might have incurred.

DaanDeMeyer:
Only Opensuse, Arch, OpenMandriva, Photon and Clear were not using them before. Opensuse and OpenMandriva were only added recently so I think we're allowed to change behavior for those.  Clear and Photon don't support them yet so those should have unchanged behavior so only Arch is left and I think we can afford to break Arch in this case to get more consistency across distros.

It is true that we don't enable the usual kernel-install scripts again on distros that build with unified kernel images. We did replace it by an equivalent script that builds unified kernel images instead. This behavior seems better than before as it is consistent. You start with a unified kernel image and any kernel upgrades or new kernels you install also give you unified kernel images in $BOOT. I guess if we really care about 100% backwards compatibility we can change that back but then we kinda have to go back and revert half of my other distro fix commits since they technically change behavior as well even though they fix broken stuff.

johannbg:
Right you guys have been not exactly conservative of changes that you guys have been making. mkosi is being used in testing and ranging from creating containers in infrastructures to vm's to even iso images using something like https://github.com/not-a-user/raw2iso. By working around upstream/downstream bugs quirks and decisions and changing previous behaviour ( or existing behaviour in downstream ) you just end up creating more load either on yourselves ( like "Work around debian bug #835628" which has been fixed downstream now you need to remove that code ) or on downstream maintainers which are debugging reports against downstream components and suddenly after endless back and forth the reporter mentioned he generated the image using mkosi right. 

Any quirks and workarounds should be happening in the distribution files themselves ( and configuration residing in extra which get added and or removed by the consumer ) not be hardcode ( if like for example if end users want to use sd boot not grub ) or atleast if they must reside in the code, the code needs to be split into distributions mkosi "modules" ( which actually would help mkosi scale bettes,  a distribution would get added as a new module ) and reside there. 

What's the end game here where are you guys going with the project, what's the direction that you are taking?

DaanDeMeyer:
I can definitely empathize with the not changing distros too much if possible. On the other hand, I think one of the goals of mkosi set by @poettering is to do legacy free images and that sometimes requires doing things a little differently (e.g. unified kernel images via dracut or using systemd-boot by default). I agree the debian workaround should be removed and we should prefer not adding distro specific fixes. Maybe we can drop the dracut dependency on a few distros as well when not doing unified kernel images and stick to distribution defaults instead.

In terms of direction, It started out as just fixing broken stuff primarily for Arch for me. Then I moved on to fixing broken stuff for the other distros as well and adding CI to make sure stuff keeps working. Then I realised that all the effort I did to get mkosi CI working on Github Actions could be extremely useful for Linux userspace projects that want to run mkosi in CI (including dracut which I saw you're looking into as well). For now that includes publishing a Github Action to do all the menial work of getting mkosi installed on Github Actions and extending the configuration mechanism to support detailed per distro configuration (see the latest issue). The longterm goal is to get mkosi in a state where we can use it for systemd CI (and preferably any project) by adding the github action to the systemd repository and adding a workflow that builds, installs, boots and maybe even runs the systemd integration tests in mkosi for each supported distro.

One of @behrmann's longterm goals is splitting mkosi up into something a bit more maintainable (correct me if I'm wrong)

There's also the devserver thing but that's on the backburner for now.

johannbg:
Dont get me wrong there are alot of good that you guys are doing ( like that github action stuff had seen am going to use it )  but you cant have it either or I mean the grub stuff goes directly against any future vision but there are other things like changing Debian's policy of starting daemons by default to one that denies all daemon in startups which has nothing to do with any "future" vision past or present and the Debian community is *very* conservative about making changes we are talking literally years of debate before the tiniest consensus is reached and the relevant knob is changed so Debian users running Debian based image regardless of where it comes from expect everything on the image being running. 

Last time I checked the Arch community had not decided on Dracut and it could go either way if someone picks up the maintainership of Mkinitcpio but feel free to correct me if I'm wrong and we would actually love to get Giancarlo engage directly upstream with us. 

If the intent is going for the future then make it explicit dependent on x components being available and used by X distro ( not necessary by default but for example Dracut is the only supported initrd tool not all of them encase downstream is supporting/shipping a bunch of those ). 

The unified kernel images needs to be off by default and configurable in the distro files before being turned on by default to allow for adjustments to happen that's kinda a given for such a change. 

@behrmann longterm goals are fine and probably best if he re factors the code and splits distributions into modules which isolates breakage to modules and allows mkosi to scale properly ( by my last count 220 distro's used systemd you can just image how big of a mess mkosi becomes if they all decided one day to be added to mkosi ) just try to make it one time big breakage based on re-factoring the entire code and it's clear what breaks if something breaks with that change.

Regardless of what @poettering vision initial was with the project you guys have kinda taken the lead of it now, for better or worse ( those that contribute the most are the ones that lead ) ;)

behrmann:
> One of @behrmann's longterm goals is splitting mkosi up into something a bit more maintainable (correct me if I'm wrong)

That's correct. I also find the installation logic in mkosi valuable in itself and would like to reuse it `$DAYJOB` without having to vendor mkosi into my own codebase and jump through hoops while doing that.

> [..] like changing Debian's policy of starting daemons by default to one that denies all daemon in startups [..]

This has been in the codebase since 2018, which was about when I first looked at it. 

> Last time I checked the Arch community had not decided on Dracut and it could go either way if someone picks up the maintainership of Mkinitcpio but feel free to correct me if I'm wrong and we would actually love to get Giancarlo engage directly upstream with us.

I don't know the status of this either, but dracut has at least made it to the official repositories now, something that I welcome.

> The unified kernel images needs to be off by default and configurable in the distro files before being turned on by default to allow for adjustments to happen that's kinda a given for such a change.

Wouldn't that be a breaking change? From my understanding mkosi was already defaulting to unified kernel images for some distributions in the case of a bootable image only supporting EFI without the ability to switch this off except for also supporting bios boot.

My understanding is, that mkosi is supposed to be an opinionated tool, see the `policy-rc.d` file or using dracut for Debian (don't they still default to initramfs-tools?). My intention is definitely not to wantonly break things, but I think it is reasonable to streamline all supported distributions as much as possible to a single stack, e.g. using dracut and unified kernel images wherever it is possible to do that, and changes might - inadvertently - break things. It is a bit of a downside of mkosi releases being few and far in between, but simply cloning from master without pinning the commit is generally a recipe for disaster.

For what its worth, in my long term changes, I strife to have them functionally transparent. I hope this qualifies as not having lost it. :)

grazzolini:
The status of dracut on Arch is that it's packages, it's part of initramfs provides and that's it. I have been maintaining mkinitcpio and I also have been packaging dracut. But, one of the reasons why dracut doesn't install the kernel, is that I don't like that idea of hardcoding arguments for it on a pacman hook. It doesn't have an easy way to mimic mkinitcpio's profile logic either, so I've left the dracut package without hooks for now.

I do plan however to make it on par with mkinitcpio, at least with regards with kernel copying/installation. Another thing is, the dracut code base is quite interesting, but it has a lot of rpm world specific stuff. For example, I can't run dracut tests on Arch due to it's rpm requirement (we do have rpm-tools on Arch now though, still, they are specific to Fedora and others).

Having said that, dracut works fine on Arch, I have myself used it for more than a year now. And, even though it produces slightly larger images than mkinitcpio, I think the fact it's systemd based *by default* has a lot of benefits. But it also will have the systemd issues, when they arise. I've noticed that dracut on Arch has a tendency of breaking boot more often, due to the fact that Arch moves relatively fast with systemd (compared to other distros). So, having said all this, I wouldn't recommend *right now* the switch from mkinitcpio. Also, I do foresee mkinitcpio being maintained for the foreseeable future.

DaanDeMeyer:
@grazzolini I saw https://gitlab.archlinux.org/archlinux/arch-boxes/-/issues/110 and was wondering if you'd consider switching Arch to kernel-install along with the changes to bring dracut up to the same level as mkinitcpio. If installing systemd installed kernel-install pacman hooks and the mkinitcpio and depmod pacman hooks were removed, mkosi would be able to remove a ton of code and stick very close to upstream Arch instead of deviating from it a little bit as we do now.

(I'd reply on Gitlab itself but non-devs don't have access yet)

grazzolini:
Last time this was discussed, this topic was quite controversial. I myself used kernel-install for a while, and it works. However, it can be finicky with other bootloaders and it also is somewhat complex for the job. But, since it's already present on Arch, due to systemd, shouldn't be hard to use it for mkosi. Also, regarding an Arch gitlab account, if you want to contribute something to arch-boxes, or another Arch project, I can give you an account there. It will open to the general public soon, but there's no date yet.

PR #491, opened 2020-08-05
Fedora: Fix --mirror

DaanDeMeyer:


DaanDeMeyer:
Let's merge as CI is green and the fix is pretty trivial.

PR #499, opened 2020-08-08
Manpage and entrypoint fix

### LINKED ISSUE #497: Include man page in Python package
When building the man page with
```
pandoc -t man -o mkosi.1 mkosi.md
```
the output looks rather unpleasing:
```
mkosi  ‐  Build  Legacy‐Free  OS  Images mkosi [options...] build
mkosi [options...] clean mkosi [options...]  summary  mkosi  [op‐
tions...]  shell [command line...]  mkosi [options...] boot [com‐
mand line...]  mkosi [options...] qemu mkosi is a tool for easily
building  legacy‐free OS images.  It’s a fancy wrapper around dnf
‐‐installroot, debootstrap, pacstrap and zypper that may generate
disk  images  with a number of bells and whistles.  The following
output formats are supported: Raw GPT disk image,  with  ext4  as
[...]
```

We should fix that and maybe generate the man page automatically via a commit hook, so that the man page can be included (and installed) when doing a pip install.

behrmann:
This PR does two things: It fixes the entrypoint installation for non-editable installs. Seems like I always ever did that and didn't notice that pip complains in that case (and only in that) that the entrypoint must be a callable.

I also generated the man page and include it now as a data file to be installed to `installation root/share/man/man1`. This doesn't really work for virtual envs, since that won't be in the `MANPATH`, but for the other cases it should be the most portable solution. Data files also seem to only be installed for non-editable installations.

Fixes #497


DaanDeMeyer:
Should we put the main function in __init__ as well then? We could also rewrite the entrypoint as mkosi:main then I think.

behrmann:
I'd leave it out. We need a `__main__.py` anyways to be able to call mkosi via `python -m mkosi`, which I'd like to keep, because it helps when having multiple Python versions around. Also, I'd like to keep the reusable parts in in one place and have the stuff specific for mkosi the program in `__main__.py` and in the future hopefully in `mkosi/cli.py`.

DaanDeMeyer:
One problem I just noticed is this breaks when using sudo. sudo drops PATH modifications from the user which means the mkosi in ~/.local/bin cannot be found anymore.

behrmann:
This is a terrible problem. I didn't run into it, because I run this in a venv, where it works. I think the problem is rather, that sudo changes the user and home, which screws up `sys.path`, which is circumvented in a venv. 

I guess this rules out entrypoints for now. I'll try to fixup a script that can be used in the same way, but works around this problem, but until that is ready we should probably revert 5c9faf85fd6cc3101bbed9c506bfcce612102eae 97f22b9fada69687127c8086e9ea4ee869a33c15 0e477deeabb51e99ce4f85538b079ed8b42b250d

DaanDeMeyer:
I was looking at this myself, I tried doing this if we're not root and it finds the script but not the module

```python
def elevate() -> None:
    if os.getuid() != 0:
        os.execlp("sudo", "-E", "--", *sys.argv)
```

```
‣ Processing default...
  File "/home/daan/.local/bin/mkosi", line 33, in <module>
    sys.exit(load_entry_point('mkosi', 'console_scripts', 'mkosi')())
  File "/home/daan/.local/bin/mkosi", line 22, in importlib_load_entry_point
    for entry_point in distribution(dist_name).entry_points
  File "/usr/lib/python3.8/importlib/metadata.py", line 504, in distribution
    return Distribution.from_name(distribution_name)
  File "/usr/lib/python3.8/importlib/metadata.py", line 177, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: mkosi
```

behrmann:
Yeah, apparently the egg.link is not recognised as a proper package even when the search path is right.

I have a rough solution in bash
```bash
#!/bin/bash
export PYTHONPATH=/home/${SUDO_USER}/.local/lib/python3.8/site-packages/:<path from mkosi.egg-link>
python -m mkosi $@
```

This again, would fail inside a virtualenv. I'll flesh this out tomorrow.

DaanDeMeyer:
`sudo python -m mkosi` does work. I'll put that in the readme for now while we work out a fix.

DaanDeMeyer:
We could simply put a bash script in ~/.local/bin/mkosi that calls `python -m mkosi`. That works flawlessly except that pip won't uninstall it.

behrmann:
I fear it is not that easy. If mkosi is installed with `pip install --user` calling it via `python -m mkosi` doesn't work for me.

behrmann:
I have a rough workaround like this already
```bash
export PYTHONPATH=/home/${SUDO_USER}/.local/lib/python3.8/site-packages/:<path from mkosi.egg-link>

python -m mkosi $@
```
but this will definitely break when used in a virtual environment.

DaanDeMeyer:
Ah damn that was just because I was executing from the mkosi project directory. OK, I'll revert for now.

DaanDeMeyer:
Thinking mroe about this, sudo and local installs are fundamentally incompatible because ~/.local isn't secured. If your machine is compromised and someone puts a nasty module in ~/.local, it could easily get included into a script that is executed with sudo. So It's completely logical that when running with sudo python only loads from /usr. 

The fix for this is to make sure python doesn't have to go searching which means either running from the project directory or packaging everything together in an archive (e.g. use zipapp).

Instead of hosting on pypi, we'd just release the zipapp using github releases. User's put them wherever they want and we're done. Distros can package it as well. it shouldn't be a problem for them that we're putting all the files in a zip as long as we don't depend on anything but the standard library.

behrmann:
The zip app route is a possibility and I hear it also works fine when vendoring dependencies. That being said, I don't think they had security in mind. Python will load load packages from `/root/.local` just fine, also when being privileges are escalated via sudo. It's also debatable were the compromise is. If a user has a nasty package in his home it can probably already read everything they care about and unfortunately, most users will just run `sudo pip` just fine if they want to get something to work.

DaanDeMeyer:
Sure vendoring dependencies works with zipapps but distros wouldn't be all too happy about that. I just tried locally and zipapp works great with a makefile. development is still a problem though. We could add a `make dev` that installs a file to ~/.local/bin that calls `sudo PYTHONPATH=<mkosi_dir> python -m mkosi` and `make undev` to remove that file. Then we can do the actual packaging with zipapp and we're done. No complicated build dependencies or anything, just standard python and a makefile.

PR #500, opened 2020-08-08
Add setup-mkosi Github Action

DaanDeMeyer:
Let's see how this goes. Theoretically if this works we could integrate this action into systemd's repository and start using all the free Actions CI capacity to test systemd on all kinds of distros.


DaanDeMeyer:
Seems like support for using actions like this was released yesterday. Talk about nice timing!

DaanDeMeyer:
Removed the namespacing in the step names as they will be visualized under the step node in the future. The feature is still a bit barebones but it saves us from writing Javascript to implement the action. Let's keep this internal to the repository for now until some other features are released that make this a bit more featureful.

I combined all the dependencies into a single step as well. It takes a little bit longer but shouldn't be too excessive.

DaanDeMeyer:
Changed the CI script to use ./mkosi instead of the mkosi installed by the Github Action. The action installs from Github which takes the master branch instead of the PR branch. This is good for external users but for mkosi itself we actually want the PR branch version instead.

DaanDeMeyer:
Let's merge. CI is green and we didn't do a lot more than move code from the workflow to the action.

PR #504, opened 2020-08-09
Entrypoint v2

behrmann:
As a followup to the discussions in #499, this is a draft (it needs rebasing against master and some documentation) for how we could use a wrapper script instead of an entrypoint to have mkosi a proper module and use it, too.

This works for me installed with `--user`, `--user --editable` and in a virtual environment. 

The handling of virtual environments so far is not reasonable (it should also sudo in the exec). I don't bother saving an existing `PYTHONPATH` because it would usually get clobbered by sudo anyway.


DaanDeMeyer:
It's great that you got this working but I'm not sure if this is necessarily better than zipapp

pip:
- Somewhat standardized development mode by using editable installs. However, has small issues and bugs as we've noticed.
- Relatively more complex bash script that has to play with PYTHONPATH to get sudo working from within a local environment.
- Can release on pypi so installing via pip is easy
- Can potentially depend on libraries via pip in the future although I don't think we actually want to do this
- Can import as a module? I think you need the bash script either way no and if you need the bash script importing as a module becomes impossible no? Unless you import stuff from mkosi that doesn't require root.

zipapp:
- Our own development mode by putting a 2 line script in ~/.local/bin that adds the mkosi source directory to PYTHONPATH and calls python -m mkosi. sudo is kept out of the script which meshes well with how mkosi is currently invoked. The pip approach would break scripts that call mkosi with sudo I think.
- Extra Makefile to do stuff which is standard practice for linux userspace tool but maybe not as much for python projects. I think mkosi identifies more as a linux userspace tool that happens to be written in python than as an actual python project so imo this wouldn't be a problem. Most of our users and contributers should be familiar with the concept of a makefile.
- No dependency on pip at all, we can drop setup.py as well. On Arch, this saves me 17 packages on my system actually since I don't use pip at all for any other stuff.
- No need for a dependency on an external package repository like pypi, releases are done using github release functionality
- One single executable pyz file containing all source code. Extremely simply for users to use, simply put in PATH and done
- Not installable as a module. We can still support importing as a module but only if the project source directory is in PYTHONPATH. 
- Bit harder for users to inspect an installed version of mkosi since it's a zip file and not separate python files anymore. Installed mkosi becomes more like a compiled binary.

Personally, I'm slightly in favor of using zipapp. I think especially our no dependencies except stdlib constraint makes us an ideal candidate for using it. Not having to deal with pypi makes things easier as we don't have to maintain the project on pypi and not needing pip is nice as well. I'm not a huge fan of makefiles but in this case I think its a good fit.

behrmann:
I don't think this is necessarily an either or question. We can in principle do both, release on PyPI and as a zipapp on Github.

As soon as we split out mkosi into a library (which makes sense to better structure the code and for possible reuse), we will need to find a way to actually drive it. zipapp is a solution for this, since it basically just runs `python -m mkosi` and in the library case it could be imported (when installed for the root user). For the development use case, though, we would still need a way to pass mkosi, that will in most cases live in a user's home, to the root user in such a way that it's obtained from the calling users environment.

> pip:
>
> * Relatively more complex bash script that has to play with PYTHONPATH to get sudo working from within a local environment.
>
> zipapp:
>
> * Our own development mode by putting a 2 line script in ~/.local/bin that adds the mkosi source directory to PYTHONPATH and calls python -m mkosi. sudo is kept out of the script which meshes well with how mkosi is currently invoked. The pip approach would break scripts that call mkosi with sudo I think.

This would be a very simple solution, as it's easy to hardcode the path of the checkout, but we would neeed a completely separate installation mechanism for this (e.g. the mentioned Makefile). I don't think it is the best user experience to have multiple ways of installing something, which is why I opted for a slightly more involved wrapper script, that programmatically determines where mkosi is checked out from.

> pip:
>
> * Can import as a module? I think you need the bash script either way no and if you need the bash script importing as a module becomes impossible no? Unless you import stuff from mkosi that doesn't require root.

You can import it as a module and the bash script shouldn't interfer with this in any way. In this PR `mkosi` the module would be installed in the appropriate path und would always be importable for that user that installed it, although a regular user would not be able to do much with it, but library reuse in some CI or installer would need to run as root anyway and would have therefore probably have it installed globally or locally for the root user anyway.

The script would also end up in the appropriate path for that and its point is only really to pass the user's environment through to root, when installed by a regular user.

> zipapp:
>
> * No dependency on pip at all, we can drop setup.py as well. On Arch, this saves me 17 packages on my system actually since I don't use pip at all for any other stuff.

I think via transitive dependencies it might actually be a bit more, but quite a few of these are extremely common and even the direct dependencies of `python-pip` on Arch are transitive dependencies of some of the others and either directly or transitively dependencies of a lot of other common packages. Also, they are tiny and disk space is cheap. ;)

Actually, when not using the `console_script` entrypoint, with the explicit wrapper script, we can also get completely rid of `setup.py` and still have it installable via pip :) (Note, for an `--editable` install we'd still need a `setup.py` until that has been standardised.)

> pip:
>
> * Can release on pypi so installing via pip is easy
>
> zippapp:
>
> * Extra Makefile to do stuff which is standard practice for linux userspace tool but maybe not as much for python projects. I think mkosi identifies more as a linux userspace tool that happens to be written in python than as an actual python project so imo this wouldn't be a problem. Most of our users and contributers should be familiar with the concept of a makefile.
>
> * No need for a dependency on an external package repository like pypi, releases are done using github release functionality
>
> * One single executable pyz file containing all source code. Extremely simply for users to use, simply put in PATH and done
>
> * Bit harder for users to inspect an installed version of mkosi since it's a zip file and not separate python files anymore. Installed mkosi becomes more like a compiled binary.

The reason I think maintaining the possibility to install via pip (while also being able to split mkosi into a library) is that it would make it quite a bit easier to install it for most users. Sure downloading a zipapp is just as simple, but it's less discoverable; you have to browse to the Github releases page, download and then possibly move it to your PATH, whereas with pip it is searchable and the installation is a one-liner that already puts everything in its place. I think it's easier to win over the hearts and minds of prospective new users by making the installation as seemless as possible.

DaanDeMeyer:
> As soon as we split out mkosi into a library (which makes sense to better structure the code and for possible reuse), we will need to find a way to actually drive it. zipapp is a solution for this, since it basically just runs python -m mkosi and in the library case it could be imported (when installed for the root user). For the development use case, though, we would still need a way to pass mkosi, that will in most cases live in a user's home, to the root user in such a way that it's obtained from the calling users environment.

I agree with better structuring the code but giving mkosi a public python API is a whole other can of worms that I don't think we want to open. Development case is easy if we don't do library, see my draft PR, just symlink the bash script to somewhere in the PATH if you want to run in development mkosi from anywhere. Distribution is also easy since we just release the zipapp.

> You can import it as a module and the bash script shouldn't interfer with this in any way. In this PR mkosi the module would be installed in the appropriate path und would always be importable for that user that installed it, although a regular user would not be able to do much with it, but library reuse in some CI or installer would need to run as root anyway and would have therefore probably have it installed globally or locally for the root user anyway.

As before, I don't think we want to go through the hassle of supporting library usage.

> The reason I think maintaining the possibility to install via pip (while also being able to split mkosi into a library) is that it would make it quite a bit easier to install it for most users. Sure downloading a zipapp is just as simple, but it's less discoverable; you have to browse to the Github releases page, download and then possibly move it to your PATH, whereas with pip it is searchable and the installation is a one-liner that already puts everything in its place. I think it's easier to win over the hearts and minds of prospective new users by making the installation as seemless as possible.

I disagree pip is easier, we can provide a one-liner in the readme to install it with curl as well. Whether using zipapp or pip users will have to make sure ~/.local/bin (or whatever directory they want to install to) is in the PATH. Users will always end up on the github page before searching on pip unless they install via distro package manager. 

I don't mind trying to support mkosi as a module, but only from the source directory. Distributing a module is a whole lot of extra maintenance effort without a big payoff. I think we're better off focusing on making people want to use mkosi the command line tool before we start thinking about making people want to use mkosi the library.

behrmann:
> I agree with better structuring the code but giving mkosi a public python API is a whole other can of worms that I don't think we want to open. Development case is easy if we don't do library, see my draft PR, just symlink the bash script to somewhere in the PATH if you want to run in development mkosi from anywhere. Distribution is also easy since we just release the zipapp.

For what it's worth, in Python I use library and module interchangeably, since Python basically only has public APIs. I also don't want to suggest to support library usage right out of the gate, the API would be anymore useful that it is now and people could have already just vendored the the file if they wanted to. 

I see the appeal of zipapp, but I think that supporting proper installation via pip is worthwhile as well and I don't think it needs to be an either or question.

behrmann:
I actually forgot to mention my main point. Too hot today :)

Right now, mkosi is installable via pip (pip puts the script into the appropriate directory). Changing wholly to distribution via zipapp and not providing some wrapper that pip can handle, would break that. That is what I am trying to avoid.

DaanDeMeyer:
So how would doing both work? We drop setup.py, use mkosi/mkosi.py mkosi/__init__.py mkosi/__main__.py directory layout (do we even need __init__.py really?), keep the makefile from the zipapp PR and rename the script from the zipapp PR to mkosi.sh so it doesn't conflict with the mkosi directory?

DaanDeMeyer:
If that allows pip to install mkosi without setup.py we're good right? We can keep the directory layout in such a way that it install fine via pip and use zipapp for official distribution.

behrmann:
I quickly added another commit to show what a synthesis of this PR and #505 could look like.

I'm not quite sure why `setup.py` should be dropped. It could be dropped, as it's nowadays not strictly necessary anymore (most can be declared in `setup.cfg` or a `pyproject.toml` when using a different build system), but it would be needed to allow `--editable` installs in the scheme of `/bin/mkosi`. An advantage of this method over a zipapp is also, that is allow to distribute related files, like the man page.

The `__init__.py` in `mkosi/` is needed because the file is needed for `mkosi/` to be the module `mkosi`.

DaanDeMeyer:
I still think the added complexity is not worth the benefit (the whole required sudo hackery is just ugly). For me, the problem to solve is allowing to split mkosi up into multiple files without complicating distribution. I don't see how pip is better for that than zipapp. For my use case and the use cases of distros packaging mkosi, I don't think the whole venv, module, pip stuff is all that useful.

Even the commit logs of mkosi never mention pip. The only mention I find is of setup.py being invoked manually to build an sdist. fedora currently installs the mkosi file directly, debian uses its own build system, and the aur package just calls setup.py install --root. All these use cases are perfectly served by just building and installing the zipapp. Normal installs just curl the executable zip to somewhere in their PATH.

Let's leave this open for others to comment but my vote goes to just using zipapp. If there's a huge demand for pip installation I'll reconsider but atm I don't think there is :/.

behrmann:
Actually, I think that this proposal simplifies distribution for distributions or rather, doesn't increase their complexity, since I think that the ability to something via pip is equivalent to what distributions do. For the longest time pip didn't really do all that much more than running `setup.py` (after fetching the package from somehwere), so the whole logic we currently have and the wrapper script would just work for distributions, at least for Debian and Arch, I'm not sufficiently familiar with the RPM way.

I would also imagine, that currently most installations are via pip and not distribution packages, since they tend to be thoroughly outdated. 

I agree pushing sudo into the wrapper script is non-optimal, but I chose this because it was less intrusive and more robust than assuming it would be called via sudo, but it could definitely be changed to do that.

behrmann:
I had another stab add providing a palatable wrapper script. This one does no sudo hackery, all path lookups are are simply what Python would regularly do, but for another user (to accomodate running the script with sudo). The hackiest thing is the bash condom the script is wrapped in, because when installing scripts in `--editable` mode setuptools installs a shim, if it's a Python script, to allow for that script to be editable as well. This unfortunately leads to the same path lookup issues. The bash condom is simply to make the script script in its entirety *not* valid Python.

DaanDeMeyer:
Can we keep the script in bash and use PYTHONPATH? Aside from that this looks solid.

behrmann:
I'm pretty sure that should be possible. I'll reimplement this logic in bash tomorrow, it's better than the previous version of the script.

bin/mkosi
```diff
@@ -0,0 +1,49 @@
...
+        additional_path = usersite
+
+    return additional_path
+
+    return
+
+
+if __name__ == "__main__":
+    try:
+        from mkosi.__main__ import main as mkosi_main
```
DaanDeMeyer, line 36: This os probably the most awkward thing to implement in bash but if we just prefer local installation over system installation I think we're fine since we can just always set PYTHONPATH before invoking python. That also allows both local and system versions to be installed with the local version being preferred over the system version which is great for development.

DaanDeMeyer:
And if possible, try to keep it zipappable as well, we just won't mention it in the docs for now. Then we don't lock ourselves in to one approach should we have to switch in the future.

behrmann:
That should already be the case in this version of the PR, see `generate_zipapp.sh`.

DaanDeMeyer:
Awesome!, Maybe disregard my mention of not mentioning zipapp in the docs. Let's remove the generate_zipapp.sh script and add a single line command to the installation docs on how to get a zipapp. Let's also run that command in CI to make sure it keeps working. If it becomes too much of a maintenance hassle we can remove it but I don't think that will be the case.

behrmann:
This version has the wrapper script reimplemented in bash. I do want to add some Ci tests to check all installation paths before this is merged, though.

bin/mkosi
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/bash
+
+PYTHON=$(command -v python3)
+PYVERSION=$($PYTHON -c '
```
DaanDeMeyer, line 4: `python --version | cut -d ' ' -f 2 | cut -d '.' -f 1-2`?

bin/mkosi
```diff
@@ -0,0 +1,32 @@
...
+import sys
+version = sys.version_info
+print(f"{version[0]}.{version[1]}")
+')
+OLD_PYTHONPATH="$PYTHONPATH"
```
DaanDeMeyer, line 9: Isn't it simpler to simply use a different variable name for the extra python path additions instead of saving the old pythonpath just so we can reuse PYTHONPATH as a variable name?

bin/mkosi
behrmann: True, that is a reasonable solution. Alternatively 
```bash
python --version | awk '{ split($2, a, "."); print a[1] "." a[2] }'
```
which omits starting another process, but I guess I'll go with your solution as to not depend on AWK.

bin/mkosi
behrmann: I don't think that makes much of a difference, since we'd still need to check at the end whether we need to combine the two or not, but I have no particular preference either way. I think I might have chosen this way, because it is similar to the dance `bin/activate` does in a virtualenv.

DaanDeMeyer:
I guess the unit test ci workflow is the best place to put the install logic tests. (It's not really unit testing though so I guess we should change the name of that workflow to something more generic maybe)

behrmann:
> I guess the unit test ci workflow is the best place to put the install logic and zipapp tests. (It's not really unit testing though so I guess we should change the name of that workflow to something more generic maybe)

Adding it there was also my plan, but is there a downside to having it in separate files other that Github needs to start another runner? I might make for a better view to know from the failing CI job what the issue was instead of having them all in a single CI job.

behrmann:
@DaanDeMeyer Also, I thought it might make sense to name the module `_mkosi` instead of `mkosi`. This would clearly communicate that everything in that module is internal and not a usable API; at least as clearly as Python communicates this. I haven't tested this yet, but it should not be a functional change.

DaanDeMeyer:
Can we hide the functions from __init__.py somehow? If so, we can do that. If not, tbh I wouldn't bother too much. We can put a note somewhere that there's no API but I really don't think that's necessary atm.

DaanDeMeyer:
I'd prefer to keep them in a single job so actions doesn't have to spawn an entire machine just to test the installs. Let's keep it in the unit test workflow. I don't think it matters too much if we use that for all the stuff that doesn't fit in the normal workflow. If it's in separate steps its easy enough to figure out what went wrong

behrmann:
No, Python has no ability to hide anything, which is why internal functions (or modules) are usually prefixed with an underscore. 

I'll add a note to the README then.

DaanDeMeyer:
Just to be sure, if we install mkosi to the system with the github action; when we call `python -m mkosi` it'll still search the current directory first right?

behrmann:
Yes. The [relevant part of the documentation](https://docs.python.org/3/library/sys.html#sys.path) is 
> As initialized upon program startup, the first item of this list, path[0], is the directory containing the script that was used to invoke the Python interpreter. If the script directory is not available (e.g. if the interpreter is invoked interactively or if the script is read from standard input), path[0] is the empty string, which directs Python to search modules in the current directory first. Notice that the script directory is inserted before the entries inserted as a result of PYTHONPATH.

If we call `python -m mkosi` while being in a directory, that contains the `mkosi` module, we will pick that up first, if not, it will go through the rest of `sys.path`.

behrmann:
I added a draft for the installation CI, let's see whether this works as I hope it does.

behrmann:
Since the tests I added seem to work, I reordered the commits slightly to make a more logical series.

DaanDeMeyer:
Maybe add system installation tests as well? Just to complete the set.

And I keep reading online the __init__.py isn't strictly necessary anymore. Are you sure we can't do without? It doesn't matter in the grand scheme of things but we might as well see if we can keep it simple.

behrmann:
> Maybe add system installation tests as well? Just to complete the set.

Will do.

> And I keep reading online the **init**.py isn't strictly necessary anymore. Are you sure we can't do without? It doesn't matter in the grand scheme of things but we might as well see if we can keep it simple.

I haven't heard that and the [official documentation](https://docs.python.org/3/tutorial/modules.html) still says
> The `__init__.py` files are required to make Python treat directories containing the file as packages. This prevents directories with a common name, such as string, unintentionally hiding valid modules that occur later on the module search path. In the simplest case, `__init__.py` can just be an empty file, but it can also execute initialization code for the package or set the `__all__` variable, described later.

DaanDeMeyer:
This is what's confusing me: https://docs.python.org/3/reference/import.html#namespace-packages

behrmann:
Ah! Now I understand. Namespace packages are a feature we might look at at some point in the future, once distributions are split out into submodules. 

What namespace packages do is to provide a mechanism to modularise packages, i.e. to allow packages that complement other packages in their namespace. In that case they might not need a `__init__.py`, because its already provided by the host package, although I'm not entirely sure, since I've never used them personally.

DaanDeMeyer:
Tried this out and running `sudo mkosi -d arch -f` after an editable install drops me into a python prompt.

bin/mkosi
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/bash
+
+PYTHON=$(command -v python3)
+PYVERSION=$($PYTHON | cut -d ' ' -f 2 | cut -d '.' -f 1-2)
```
DaanDeMeyer, line 4: Missing `--version`

behrmann: Thanks! I guess I didn't try it out again after changing that line.

behrmann:
I touched the wrapper script once more, I noticed that reusing `PYTHONPATH` did have an edge case, if `PYTHONPATH` would not be set, it would be doubled at the end of the script. I also added a check to prevent system installations from pulling in the sitedir of a user calling mkosi via sudo.

I am confused, though, by the failing test case for system installation. I tested system installation both from within the repo and from the outside and both work and don't throw the error that is shown in CI.

DaanDeMeyer:
Maybe side effects from the previous installations? Try uninstalling the local ones before installing the system one.

behrmann:
I'll try that, I'll also add a test installing into a virtual environment for good measure.

behrmann:
Somehow I can't seem to figure out how to test the system installation properly in the Github action. There is still some interaction that needs to be broken.

This version is rebased against the current master. I also touched the wrapper script once more, this version adds a heuristic that more often leaves figuring out the Python search path for python itself, which allows calling `mkosi` when installed into a venv, without activating the virtual environment (something the [pipx project](https://github.com/pipxproject/pipx) uses).

DaanDeMeyer:
Maybe put an strace in there to see what files are being installed?

DaanDeMeyer:
I think I encountered the same problem when I was trying out zipapp locally. From what I remember, moving everything from \_\_init\_\_.py to mkosi.py fixed the issue. You can then import stuff from mkosi.py in \_\_init\_\_.py to keep the same behaviour.

behrmann:
> I think I encountered the same problem when I was trying out zipapp locally. From what I remember, moving everything from __init__.py to mkosi.py fixed the issue. You can then import stuff from mkosi.py in __init__.py to keep the same behaviour.

I think I've wrapped my head around zipapp. There is a subtlety I did not anticipate. When pointing zipapp at a directory, it will run `__main__.py`, but it will make this directory the top-level directory in the zip file, so if this directory is also a module (contains an `__init__.py`) this will break the module, i.e. break all relative imports, which is not very ergonomic. Moving everything from `__init__.py` to a `mkosi.py` solves this, but would leave the namespace of the package as `mkosi.mkosi`, which is not very ergonomic. So the trick is to have the directory module in another directory, so that the module is not broken up by zipapp; I assume this is why you moved to the `src/` based layout in #505, which has the unfortunate downside, that mkosi cannot be run as `python3 -m mkosi` from the top-level of the repository anymore. 

I hope this now combines the best of both worlds, clear namespace, non-broken relative imports and a working zipapp. The command to generate the zipapp has become a bit more intricate, though, so maybe we do add a `generate_zipapp` script and maybe put it into a `tools/` directory together with `do_a_release.sh`

.github/workflows/ci-unit-test.yml
```diff
@@ -22,3 +22,58 @@ jobs:
...
+
+    - name: Test execution from current working directory
+      run: python3 -m mkosi -h
+
+    - name: Test execution from current working directory (sudo call)
+      run: sudo python3 -m mkosi -h
+
+    - name: Test system installation
+      run: |
+        sudo python3 -m pip install .
```
DaanDeMeyer, line 34: Does using `python3 -m pip` make a difference? I'd prefer testing with `pip3` since that's what I think most people will be using.

README.md
```diff
@@ -7,6 +7,68 @@ bells and whistles.
...
+Besides the mkosi binary, you can also call mkosi via
+```shell
+python -m mkosi
+```
+when not installed as a zipapp.
+
+Please note, that the the python module exists solely for the usage of the mkosi
+binary and is not to be considered a public API.
+
+## Python module
```
DaanDeMeyer, line 61: This is duplicated.

README.md
```diff
@@ -7,6 +7,68 @@ bells and whistles.
...
+This will install mkosi in editable mode to `~/.local/bin/mkosi`, allowing you
+to use your own changes right away.
+
+For development you optionally also need [mypy](https://github.com/python/mypy)
+and [pytest](https://github.com/pytest-dev/pytest). We check tests and typing in
+CI (see `.github/workflows`), but you can run the tests locally as well.
+
+## zipapp
+
+You can also packup mkosi as a
```
DaanDeMeyer, line 37: I'd use package instead of packup here.

README.md
```diff
@@ -7,6 +7,68 @@ bells and whistles.
 For a longer description and available features and options, see the
 [man page](mkosi.md).
 
+# Installation
+
+Installing `mkosi` is easy, as it has no runtime Python dependencies (you will
```
DaanDeMeyer, line 12: mkosi shouldn't be formatted as code here

README.md
```diff
@@ -7,6 +7,68 @@ bells and whistles.
...
+
+For development you optionally also need [mypy](https://github.com/python/mypy)
+and [pytest](https://github.com/pytest-dev/pytest). We check tests and typing in
+CI (see `.github/workflows`), but you can run the tests locally as well.
+
+## zipapp
+
+You can also packup mkosi as a
+[zipapp](https://docs.python.org/3/library/zipapp.html) that you can deploy
+anywhere in your `PATH`. This will leave an `mkosi` binary in `outdir/`
```
DaanDeMeyer, line 39: a mkosi binary instead of an?

README.md
```diff
@@ -7,6 +7,68 @@ bells and whistles.
...
+```shell
+mkdir -p builddir outdir
```
DaanDeMeyer, line 41: Yeah, we can put this in a scripts directory. Maybe put do_a_release in there as well

README.md
```diff
@@ -7,6 +7,68 @@ bells and whistles.
...
+
+## Python module
+
+Besides the mkosi binary, you can also call mkosi via
+```shell
+python -m mkosi
+```
+when not installed as a zipapp.
+
+Please note, that the the python module exists solely for the usage of the mkosi
```
DaanDeMeyer, line 69: Duplicated "the"

README.md
DaanDeMeyer: Make sure to clean up the build directory as well after generating the zipapp

bin/mkosi
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/bash
+
+PYROOT=$(dirname "$0")
```
DaanDeMeyer, line 3: Maybe add a comment here at the top to explain all the stuff we do in this script and which use cases we're trying to make work?

.github/workflows/ci-unit-test.yml
behrmann: Generally the standard for advising users goes in the direction of this version. A lot of Python people tend to have multiple Python installations and then it becomes unclear which pip relates to which python binary. Functionally doing `python3 -m pip` is exactly the same as directly calling `pip3` for the appropriate `python3` binary.

behrmann:
CI failure is unrelated.

tools/generate-zipapp.sh
```diff
@@ -0,0 +1,17 @@
...
+}
+trap cleanup EXIT
+
+mkdir -p builddir
+
+cp -r mkosi "${BUILDDIR}/"
+
+python3 -m zipapp \
+        -p "/usr/bin/env python3" \
+        -o builddir/mkosi \
```
DaanDeMeyer, line 15: I think I'm missing something, won't this delete the mkosi binary immediately after we build it with the trap?

I would copy mkosi dir to builddir, build the zipapp as mkosi.pyz (or whatever the default extension is), delete $BUILDDIR/mkosi and rename mkosi.pyz to mkosi and keep it in the builddir.

.github/workflows/ci-unit-test.yml
```diff
@@ -22,3 +22,58 @@ jobs:
...
+        testvenv/bin/python3 -m pip install .
+        testvenv/bin/mkosi -h
+
+    - name: Test venv installation (sudo call)
+      run: |
+        sudo testvenv/bin/mkosi -h
+
+    - name: Test zipapp creation
+      run: |
+        mkdir -p builddir outdir
```
DaanDeMeyer, line 76: Let's use the zipapp script here as well

PR #505, opened 2020-08-09
Zipapp

DaanDeMeyer:
@behrmann This is how zipapp would look like

The diffstat is actually small, github can't seem to handle renaming a file and creating the same file again with new contents in the same PR.

Installation would be:

`curl -Lo ~/.local/bin/mkosi --create-dirs https://github.com/systemd/mkosi/releases/5/mkosi`

This allows running mkosi from anywhere without having to install any files just like before. If one wants it in the path, simply add a symlink in ~/.local/bin `ln -s $PWD/mkosi ~/.local/bin/mkosi`.


behrmann:
Yes, and this is compatible to #504. 
```shell
python -m zipapp mkosi -p "/usr/bin/env python3" -o builddir/mkosi
```
would build a similar zipapp there as well (it needs an absolute import in `__main__.py` instead of a relative one, but that's a negligible change).

DaanDeMeyer:
We can do even better, just build the zipapp from the mkosi dev script and run the zipapp itself.

behrmann:
> We can do even better, just build the zipapp from the mkosi dev script and run the zipapp itself.

I think that's needlessly indirect. Why pack it up just to unpack it again right away?

DaanDeMeyer:
@behrmann I slimmed this down to the bare necessities. If we merge this, it should at least unblock you from splitting up mkosi into multiple files until we can get a third opinion on the entrypoint changes.

behrmann:
I think this change might be too costly. 

Moving `mkosi` to `src/mkosi.py` comes with the downside of the huge diffstat I battled initially as well, which is an unfortunate git problem. This would work out ok, when combining this with the move to a module, since `mkosi` to `mkosi/__init__.py` works fine.

Removing `setup.py` will also force distribution maintainers the redo their packaging to work with zipapps. I don't know, whether e.g. Debian's packaging policy is equipped to handle this. Also, it removes the ability to use pip to install mkosi (including a man page and whatever static files we might need add in the future).

behrmann:
Apart from that the PR is from a technical point of view fine.

DaanDeMeyer:
The diffstat is only on the Github diff. It does the same rename the entrypoint PR does. I'm actually kind of curious if distros would dislike this kind of packaging. We're doing nothing more than bundling our source files together in a single executable file. I don't see how distros could have a problem with that since we use system installed python and don't bundle any dependencies. It's basically like a compiled static executable without the dependency problems. I am curious though so maybe I'll post the question on the Arch to see what the devs there think. The pip install thing is unfortunate but local installs currently don't work either. (In fact, I just tried and `pip install --user` on the master branch and I got a massive red error more than 50 lines long). The workarounds required to get pip local installs working are just a massive neon sign to me that we shouldn't even bother when a simpler solution using zipapp exists that has zero of those problems and comes included with standard python instead of relying on an external package manager.

I sympathize with why you want pip install to work but we just fit zipapps model so perfectly that it feels wrong to maintain all the extra stuff just to make pip work.

behrmann:
> The diffstat is only on the Github diff. It does the same rename the entrypoint PR does. 

Excellent, that is not a problem then. I'm still not sold on changing to the `src/` based directory layout, though.

> I'm actually kind of curious if distros would dislike this kind of packaging. We're doing nothing more than bundling our source files together in a single executable file. I don't see how distros could have a problem with that since we use system installed python and don't bundle any dependencies. It's basically like a compiled static executable without the dependency problems. I am curious though so maybe I'll post the question on the Arch mailing list to see what the devs there think.

I agree, which is why I think this might pose a problem. 

We are perfectly suited for zipapps (with the known caveats, that if we distribute zipapps, we don't distribute the man page, the license or other static files that we can distribute via pip), since we only have stdlib runtime dependencies (with the exemption of bash completion, which I'd like to change but which would yet more files to distribute), but generally zipapp would vendor their non-stdlib dependencies. In that light I don't know whether distribution will like them on these general grounds.

> The pip install thing is unfortunate but local installs currently don't work either. (In fact, I just tried and `pip install --user` on the master branch and I got a massive red error more than 50 lines long). The workarounds required to get pip local installs working are just a massive neon sign to me that we shouldn't even bother when a simpler solution using zipapp exists that has zero of those problems and comes included with standard python instead of relying on an external package manager.

I've just tested my local checkout of the current master on my Arch with `pip install --user <path to checkout>` and on a Debian machine with `python3 -m pip install --user git+https://github.com/systemd/mkosi`, both installed fine. I cannot seem to reproduce this. In that sense removing the `setup.py` is breaking this functionality. 

Integrating with the Python ecosystem is in my opinion the a good way forward to win users, since I think the user experience is better than manually downloading a binary and placing it in your `PATH`. I don't think that the manual setup of the path is a red flag, since this is something that Python normally does on every startup of the interpreter; we just fix a very particular corner case that we experience. 

This is why I'm arguing for "Why not both?" :)

DaanDeMeyer:
I'd prefer not maintaining and supporting two solutions. Let's go with pip for now in the interest of moving forward.

PR #507, opened 2020-08-13
Interpret CLI args as nspawn args when verb == "boot"

DaanDeMeyer:
Normal command line args are ignored by nspawn when --boot is used.
Instead of doing nothing with them, let's allow users to pass nspawn
args via the command line directly when boot is used.


DaanDeMeyer:
The generated by pandoc version message in the generated man page might become annoying. I'll need to look into a way to remove that if possible.

behrmann:
> The generated by pandoc version message in the generated man page might become annoying. I'll need to look into a way to remove that if possible.

It's a comment and not visible in the man page itself (at least I cannot seem to find it), so it shouldn't be too much of a problem.

DaanDeMeyer:
It turns up in diffs, that's whats annoying about it. Anyway let's merge this.

PR #508, opened 2020-08-14
Debian/Ubuntu: Simplify and fix issues

DaanDeMeyer:
- Only install dracut when building unified kernel images
- Remove workarounds
- Enable services by default, should be disabled via skeleton tree
- Fix unified kernel images build on Debian unstable
- Don't have apt-get run dracut when installing unified kernel image
- Install initramfs-tools when not building unified kernel image
- Explicit uefi-stub for Debian as well since older releases need it

Debian non unified kernel image boots seem broken for now (systemd-boot and grub). That was the case before as well so we can fix that later I guess. We might have to start using grub for UEFI as well on some distros when not doing unified kernel images. I tried getting debian to do kernel-install but the added complexity seems too much.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -1414,7 +1414,10 @@ def prepare_tree(args: CommandLineArguments, root: str, do_run_build_script: boo
             f.write("add_dracutmodules+=\" qemu \"\n")
 
         # These distros need uefi_stub configured explicitly for dracut to find the systemd-boot uefi stub.
-        if args.distribution in (Distribution.ubuntu, Distribution.mageia, Distribution.openmandriva):
+        if args.esp_partno is not None and args.distribution in (Distribution.ubuntu,
+                                                                 Distribution.debian,
+                                                                 Distribution.mageia,
+                                                                 Distribution.openmandriva):
             with open(os.path.join(dracut_dir, "30-mkosi-uefi-stub.conf"), "w") as f:
                 f.write("uefi_stub=/usr/lib/systemd/boot/efi/linuxx64.efi.stub\n")
 
@@ -2086,18 +2089,12 @@ def install_debian_or_ubuntu(args: CommandLineArguments,
                              *,
                              do_run_build_script: bool,
                              mirror: str) -> None:
-    repos = args.repositories or ["main"]
+    repos = set(args.repositories) or {"main"}
     # Ubuntu needs the 'universe' repo to install 'dracut'
-    if args.distribution == Distribution.ubuntu and args.bootable and 'universe' not in repos:
-        repos.append('universe')
+    if args.distribution == Distribution.ubuntu and args.bootable and args.with_unified_kernel_images:
+        repos.add('universe')
 
-    cmdline = ["debootstrap",
-               "--verbose",
-               "--variant=minbase",
-               "--include=systemd-sysv",
-               "--exclude=sysv-rc,initscripts,startpar,lsb-base,insserv",
-               "--merged-usr",
-               "--components=" + ','.join(repos)]
+    cmdline = ["debootstrap", "--variant=minbase", "--merged-usr", f"--components={','.join(repos)}"]
 
     if args.architecture is not None:
         debarch = DEBIAN_ARCHITECTURES.get(args.architecture)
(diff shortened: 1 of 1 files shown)


mkosi
```diff
@@ -1413,7 +1413,10 @@ def prepare_tree(args: CommandLineArguments, root: str, do_run_build_script: boo
             f.write("add_dracutmodules+=\" qemu \"\n")
 
         # These distros need uefi_stub configured explicitly for dracut to find the systemd-boot uefi stub.
-        if args.distribution in (Distribution.ubuntu, Distribution.mageia, Distribution.openmandriva):
+        if args.distribution in (Distribution.ubuntu,
+                                 Distribution.debian,
+                                 Distribution.mageia,
+                                 Distribution.openmandriva):
```
behrmann, line 1419: Should we maybe also depend this on being an EFI system?

mkosi
```diff
@@ -2107,59 +2104,33 @@ def install_debian_or_ubuntu(args: CommandLineArguments,
...
-    # The policy-rc.d script can be used choose which ones to start
-    # Let's install one that denies all daemon startups
-    # See https://people.debian.org/~hmh/invokerc.d-policyrc.d-specification.txt
-    # Note: despite writing in /usr/sbin, this file is not shipped by the OS
-    # and instead should be managed by the admin.
-    policyrcd = os.path.join(root, "usr/sbin/policy-rc.d")
-    with open(policyrcd, "w") as f:
-        f.write("#!/bin/sh\n")
-        f.write("exit 101")
-    os.chmod(policyrcd, 0o755)
```
behrmann, line 2153: I don't think we should change that. This has been in the codebase for two years with nobody complaining and changing it now would be a backwards incompatible change, since people building Debian/Ubuntu images would need to disable all the services that are now started.

mkosi
```diff
@@ -2107,59 +2104,33 @@ def install_debian_or_ubuntu(args: CommandLineArguments,
...
-        f.write("hostonly=no")
+        extra_packages.update(args.build_packages)
 
     if not do_run_build_script and args.bootable:
-        extra_packages += ["dracut"]
+        if args.with_unified_kernel_images:
+            extra_packages.add("dracut")
+            extra_packages.add("binutils")
+        else:
+            extra_packages.add("initramfs-tools")
```
behrmann, line 2122: I don't think we should change that. Installing dracut on all bootable systems has been aronud since at least 2018. Sure, Debian doesn't default to dracut (and Debian's dracut maintainer is a man with a hatred against systemd and tends to patch bashisms out of dracut), but initramfs-tools is a legacy project and starting to support that would be akin to supporting ifconfig over networkd. Being opinionated is ok. :)

mkosi
```diff
@@ -2085,18 +2088,12 @@ def install_debian_or_ubuntu(args: CommandLineArguments,
...
-    if args.distribution == Distribution.ubuntu and args.bootable and 'universe' not in repos:
-        repos.append('universe')
+    if args.distribution == Distribution.ubuntu and args.bootable and args.with_unified_kernel_images:
+        repos.add('universe')
 
-    cmdline = ["debootstrap",
-               "--verbose",
-               "--variant=minbase",
-               "--include=systemd-sysv",
-               "--exclude=sysv-rc,initscripts,startpar,lsb-base,insserv",
```
behrmann, line 2097: I think we should keep that around. They've been excluded since 2016 (I guess since the beginning).

behrmann:
I don't think we need to make all these changes. Most of them have been around for a long while and nobody has been complaining about them. I think it is ok for mkosi to be opinionated and build a modern (at least what this camp considers so) system. If I want a system that uses the Debian defaults, I can just take anyone of the myriad of other tools that has grown in the Debian ecosystem. :) 

> Debian non unified kernel image boots seem broken for now (systemd-boot and grub). That was the case before as well so we can fix that later I guess. We might have to start using grub for UEFI as well on some distros when not doing unified kernel images. I tried getting debian to do kernel-install but the added complexity seems too much.

Yes, this is an issue. I remember open bugs on the bugtracker about that, since when gummiboot became systemd-boot, but I think this would be a worthwhile thing to take on.

mkosi
DaanDeMeyer: I checked, they aren't actually installed on debian unstable, I'll check with ubuntu as well just to be sure.

DaanDeMeyer:
Ok, let me clean this up a bit to break fewer things.

mkosi
DaanDeMeyer: Ideally we'd still figure out kernel updates both with and without unified kernel images on debian though. Although I can't really be bothered more than this PR so I guess it'll have to wait until someone is interested in that.

mkosi
DaanDeMeyer: Only lsb-base is installed by default on debian unstable and ubuntu focal which is just a shell functions library, the rest is not installed. I don't think we should exclude that in mkosi, it's harmless enough.

mkosi
```diff
@@ -2107,59 +2104,40 @@ def install_debian_or_ubuntu(args: CommandLineArguments,
...
-
-    if not do_run_build_script and args.bootable and args.output_format == OutputFormat.gpt_btrfs:
-        cmdline[4] += ",btrfs-progs"
-
     run(cmdline, check=True)
 
-    # Debootstrap is not smart enough to deal correctly with alternative dependencies
-    # Installing libpam-systemd via debootstrap results in systemd-shim being installed
-    # Therefore, prefer to install via apt from inside the container
-    extra_packages = ['dbus', 'libpam-systemd', 'binutils']
```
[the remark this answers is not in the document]
DaanDeMeyer: Good catch, I'll add these back

mkosi
behrmann: Right, the list might contain remnants from earlier, when Debian did not default to systemd. I'll have a look later what debootstrap currently graces a system with.

Sure, `lsb-base` doesn't do much in itself, but it only provides function for system V init scripts, so we can (and probably should) skip it (like we're already doing). Unfortunately, more than enough Debian packages still pull them in anyways at least transitively because maintainers cannot be bothered with writing a service file, so if they're needed, they'll be there.

DaanDeMeyer: I think this is one case where we don't gain much by deviating from the distro defaults. Let's leave it up to debian to fix this instead of doing it in mkosi. It might take up a little bit of extra space but that isn't bad enough to exclude it right?

behrmann: I guess we can add it to the roster, it won't hurt, but I don't think its necessary to any of our goals, i.e. it's an unnecessary piece of legacy code and it doesn't really interact with systemd (in the testing case) at all if nothing else pulls it back in.

Btw, I just noticed that this line of excluded has been there since the initial commit of mkosi. ;)

behrmann:
Just as a note for anyone following up this change: I checked the differences in installed packages (for current stable) with both the `--include` and `--exclude`, missing both and with just the `--include`. Having both and just having the `--include` makes no difference and having neither bootstraps a system without systemd (actually without init). Since nobody should bootstrap anything older than buster, the change in the debootstrap invocation should make no user visible difference, since systemd is installed by apt right afterwards.

Looking at `systemd-sysv`, this might actually be something we could cut, since it's a package being highlighted by Debian's `cruft` packages. :) (`systemd-sysv` installs `/sbin/shutdown`, `/sbin/runlevel` and friends)

PR #509, opened 2020-08-15
Arch: Make pacman hooks more generic

DaanDeMeyer:
It's generally good to not depend on specific packages in pacman hooks
since it doesn't take new packages into account. To get around, this we
can simply depend on the /usr/lib/kernel/install.d directory so any
package that installs a file into this directory can plug into the
pacman hook.

One annoyance we can't solve without package specific information in the
pacman hook is packages that themself support hook directories that
other packages can install to. This would require a mechanism in pacman
where packages can contribute triggers to hooks installed by other
packages.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -2300,16 +2300,14 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
                     Operation = Upgrade
                     Type = Path
                     Target = usr/lib/modules/*/vmlinuz
+                    Target = usr/lib/kernel/install.d/*
+                    Target = boot/*-ucode.img
 
                     [Trigger]
                     Operation = Install
                     Operation = Upgrade
-                    Operation = Remove
                     Type = Package
                     Target = systemd
-                    Target = dracut
-                    Target = intel-ucode
-                    Target = amd-ucode
 
                     [Action]
                     Description = Adding kernel and initramfs images to /boot...


DaanDeMeyer:
Added two extra commits that drop some of Arch's default configuration which is too opiniated to be in mkosi itself in my opinion. 

Removing these does make me wonder if adding --qemu-headless wasn't a bit too opiniated either.

behrmann:
> It's generally good to not depend on specific packages in pacman hooks since it doesn't take new packages into account. To get around, this we can simply depend on the /usr/lib/kernel/install.d directory so any package that installs a file into this directory can plug into the pacman hook.

That's a reasonable change.

> Added two extra commits that drop some of Arch's default configuration which is too opiniated to be in mkosi itself in my opinion.

I think we should keep around the networkd file. It's been hanging around since 2016, so people might be expecting it. I could see it being renamed with an mkosi prefix and moved to `/usr/lib/systemd/network` though, so that people can easily disable it.

> Removing these does make me wonder if adding --qemu-headless wasn't a bit too opiniated either.

It is opinionated, but it's a niche change only affecting people who opt in and providing value to those who do.

DaanDeMeyer:
I wonder if we should move the networkd file out of Arch though then and use it everywhere. Ideally we'd use networkd as the default on all distros but centos doesn't support it for example.

DaanDeMeyer:
Dropped both the networkd and the networkmanager change since the networkmanager change depends on the networkd change. Enabling networkd on all ethernet interfaces would give a rather bad experience to users installing networkmanager.

(I need to find a free year(s) somewhere to add full Wireless/VPN support to networkd so we can use it by default without having to worry about NetworkManager)

behrmann:
> I wonder if we should move the networkd file out of Arch though then and use it everywhere. Ideally we'd use networkd as the default on all distros but centos doesn't support it for example.

Agreed, but we should discuss this in a separate PR.

> (I need to find a free year(s) somewhere to add full Wireless/VPN support to networkd so we can use it by default without having to worry about NetworkManager)

Good luck with that can of worms. Beware of any turf wars with iwd. :)

PR #510, opened 2020-08-15
Make run() set check=True by default

DaanDeMeyer:


behrmann:
First of all, I think this is a very sensible change, because always checking the return value is just safer and it is more explicit on what the actual commandline arguments are. 

Unfortunately, I cannot annotate the line that needs annotating in the diff. On [this line](https://github.com/systemd/mkosi/pull/510/files#diff-6980c629516e4165bfaf96be8e5003cfL1502) `run` would throw a `CalledProcessError`, so we wouldn't be able to change for non-zero exit code ourselves. You need to either catch the error or set `check` to `False` there.

DaanDeMeyer:
Can you give me the line number? The link doesn't seem to work.

behrmann:
Line 1502/1503 in the old version / net version respectively.

DaanDeMeyer:
Nice catch!

mkosi
```diff
@@ -1499,11 +1500,11 @@ def run_workspace_command(args: CommandLineArguments,
     if nspawn_params:
         cmdline += nspawn_params
 
-    result = run(cmdline + ['--', *cmd])
+    result = run(cmdline + ['--', *cmd], check=False)
```
behrmann, line 1503:
```suggestion
    result = run(cmdline + ['--'] + cmd, check=False)
```

behrmann:
Did you go through all instances of `run`? I haven't yet checked whether there are any missing ones, that might need a `check = False`.

DaanDeMeyer:
Found one more, I'll have one more look through everything just to be sure.

mkosi
```diff
@@ -765,7 +765,7 @@ def create_image(args: CommandLineArguments, root: str, for_cache: bool) -> Opti
         table, run_sfdisk = determine_partition_table(args)
 
         if run_sfdisk:
-            run(["sfdisk", "--color=never", f.name], input=table.encode("utf-8"), check=True)
+            run(["sfdisk", "--color=never", f.name], input=table.encode("utf-8"))
             run(["sync"])
```
DaanDeMeyer, line 769: This one wasn't checked but maybe we should check the result of sync? I don't think we expect it to fail.

mkosi
```diff
@@ -2455,9 +2456,9 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
...
         try:
-            run(["pacman-key", *conf, "--init"], check=True)
-            run(["pacman-key", *conf, "--populate"], check=True)
-            run(["pacman", *conf, "--noconfirm", "-Sy", *packages], check=True)
+            run(["pacman-key", *conf, "--init"])
+            run(["pacman-key", *conf, "--populate"])
+            run(["pacman", *conf, "--noconfirm", "-Sy", *packages])
         finally:
             # Kill the gpg-agent started by pacman and pacman-key.
             run(['gpgconf', '--homedir', os.path.join(root, 'etc/pacman.d/gnupg'), '--kill', 'all'])
```
DaanDeMeyer, line 2464: Same here, CI works so I guess we can check the result here safely as well?

mkosi
```diff
@@ -3019,7 +3015,7 @@ def make_minimal_ext4(args: CommandLineArguments, root: str, for_cache: bool) ->
         f: BinaryIO = cast(BinaryIO, tempfile.NamedTemporaryFile(prefix=".mkosi-mkfs-ext4",
                                                                  dir=os.path.dirname(args.output)))
         f.truncate(args.root_size)
-        run(["mkfs.ext4", "-I", "256", "-L", "root", "-M", "/", "-d", root, f.name], check=True)
+        run(["mkfs.ext4", "-I", "256", "-L", "root", "-M", "/", "-d", root, f.name])
 
     with complete_step('Minimizing ext4 root file system'):
         run(["resize2fs", "-M", f.name])
```
DaanDeMeyer, line 3021: This also seems like something we should check

mkosi
behrmann: I think it's fine if the sync is checked, if it fails we shouldn't go ahead anyway.

mkosi
behrmann: If those fail, we'd need to die anyway, something the code would have already done, so this is fine.

mkosi
behrmann: If those fail, we'd need to die anyway, something the code would have already done, so this is fine. In the future this be coupled with automatic cleanup for those cases?

PR #511, opened 2020-08-15
Support distro specific subdirectories in mkosi.default.d

### LINKED ISSUE #506: RFE: Allow distro specific subdirectories in mkosi.default.d
These only get applied for that specific distro. Also allow mkosi specific files in mkosi.default.d and the distro specific subdirectories. The last one found gets applied.

This is a backwards compatible way to get advanced per distro configuration.

If required we can go further and allow per release subdirectories in each distro directory and per architecture directories in each per release directory. Let's start with just distros though.

DaanDeMeyer:
Fixes #506

Only tested superficially for now but wanted to get it out there already.

This does require a somewhat significant breaking chance in that config files in mkosi.default.d now need the .conf extension. This should have been the default when it was added but unfortunately we have to break stuff now to get this working.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -310,6 +310,9 @@ class Distribution(enum.Enum):
     photon = 10
     openmandriva = 11
 
+    def __str__(self) -> str:
+        return self.name
+
 
 GPT_ROOT_X86           = uuid.UUID("44479540f29741b29af7d131d5f0458a")  # NOQA: E221
 GPT_ROOT_X86_64        = uuid.UUID("4f68bce3e8cd4db196e7fbcaf984b709")  # NOQA: E221
@@ -4161,6 +4164,25 @@ def create_parser() -> ArgumentParserMkosi:
     return parser
 
 
+def load_distribution(args: CommandLineArguments) -> CommandLineArguments:
+    if args.distribution is not None:
+        args.distribution = Distribution[args.distribution]
+
+    if args.distribution is None or args.release is None:
+        d, r = detect_distribution()
+
+        if args.distribution is None:
+            args.distribution = d
+
+        if args.distribution == d and d != Distribution.clear and args.release is None:
+            args.release = r
+
+    if args.distribution is None:
+        die("Couldn't detect distribution.")
+
+    return args
+
+
 def parse_args(argv: Optional[List[str]]=None) -> Dict[str, CommandLineArguments]:
     """Load default values from files and parse command line arguments
 
@@ -4242,6 +4264,12 @@ def parse_args(argv: Optional[List[str]]=None) -> Dict[str, CommandLineArguments
     # Parse everything in normal mode
     else:
         args = parse_args_file_group(argv, default_path)
+
+        args = load_distribution(args)
+
+        # Parse again with any extra distribution files included.
+        args = parse_args_distribution_group(argv, default_path, str(args.distribution))
+
         args_all['default'] = args
 
     return args_all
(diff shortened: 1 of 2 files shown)


DaanDeMeyer:
Updated to keep backwards compat. We now check for a directory named after the distro we're building in mkosi.default.d. If it exists, we also load all files from that directory identically to how we load files from mkosi.default.d. This doesn't allow putting mkosi.build etc scripts in mkosi.default.d anymore since that requires breaking compat but I don't think it matters too much. By having per distro config files, those config files can easily include explicit paths to point mkosi at the correct build scripts, extra trees, etc.

DaanDeMeyer:
Did a bit of cleanup. If needed, we can extend this scheme in the future  to have per release directories in the per distro directories as well.

mkosi/__init__.py
```diff
@@ -298,17 +298,20 @@ def can_minimize(self) -> bool:
...
+    opensuse = "opensuse"
+    mageia = "mageia"
+    centos = "centos"
+    centos_epel = "centos_epel"
+    clear = "clear"
+    photon = "photon"
+    openmandriva = "openmandriva"
+
+    def __str__(self) -> str:
+        return cast(str, self.value)
```
behrmann, line 314: I don't think this renaming is necessary. Couldn't the `__str__` method could be `cast(str, self.name)` instead? This way the members don't need to repeat the keys. Easiest would be to use `enum.auto()` anyways.

behrmann:
Sorry for the late look at this. 

I support this change. I guess this could be really helpful for a number of use cases, but since this touches The config loading, which I always found somewhat gnarly, I think this code path should also be added to the tests.

DaanDeMeyer:
Did one small change to have the distro specific config files override mkosi.default and mkosi.default.d files (but not cli args). I want to add a test but the current test design is an overengineered mess imo. I need to simplify it but haven't had the time yet. The logic here is pretty simple, we read the config once, figure out the distro and use that to read any extra config files if there are any. I say we merge this, get the new release out and then we can worry about the tests.

mkosi/__init__.py
```diff
@@ -4277,6 +4305,24 @@ def parse_args_file_group(argv_post_parsed: List[str], default_path: str) -> Com
...
+
+    distribution_dir = os.path.join(defaults_dir, distribution)
+    if os.path.isdir(distribution_dir):
+        for distribution_file in sorted(os.listdir(distribution_dir)):
+            distribution_path = os.path.join(distribution_dir, distribution_file)
+            if os.path.isfile(distribution_path):
+                all_defaults_files.append(ArgumentParserMkosi.fromfile_prefix_chars + distribution_path)
+
+    # Insert the distro specific config files after the rest of the config files so they override these.
+    i = next(i for i, v in enumerate(argv_post_parsed) if not v.startswith('@'))
```
behrmann, line 4320: Should the `@` here be a literal @ or should this be
```suggestion
    i = next(i for i, v in enumerate(argv_post_parsed) if not v.startswith(ArgumentParserMkosi.fromfile_prefix_chars))
```

behrmann: But just in general, I find this line a bit hard to read. It makes perfect sense and I think it's the best performing way to write this, but whenever I see `next` I need to think about the intent for a second.

This is equivalent to (combined with the line below)
```python
for i, v in enumerate(argv_post_parsed):
    if not v.startswith('@'):
        argv_post_parsed[i:i] = all_defaults_files
        break
```
is it?

behrmann (commented):
> Did one small change to have the distro specific config files override mkosi.default and mkosi.default.d files (but not cli args).

That makes sense.

> I want to add a test but the current test design is an overengineered mess imo.

That is what I meant by gnarly. I am not too eager too touch them, but maybe that's a reason to do so in the first place.

behrmann:
The failures look unrelated, some Github internal caching problems maybe.

DaanDeMeyer:
Added one more fix for an edge case.

mkosi/__init__.py
```diff
@@ -4277,6 +4305,29 @@ def parse_args_file_group(argv_post_parsed: List[str], default_path: str) -> Com
...
+                all_defaults_files.append(ArgumentParserMkosi.fromfile_prefix_chars + distribution_path)
+
+    # Insert the distro specific config files after the rest of the config files so they override these.
+    for i, v in enumerate(argv_post_parsed):
+        if not v.startswith(ArgumentParserMkosi.fromfile_prefix_chars):
+            argv_post_parsed[i:i] = all_defaults_files
+            break
+        # Append to the end if the args only contain files and no regular args.
+        if i == len(argv_post_parsed) - 1:
+            argv_post_parsed += all_defaults_files
```
behrmann, line 4326: Although uncommon, this is *the* textbook case for `else` on the for loop
```suggestion
        # Append to the end if the args only contain files and no regular args.
    else:
        argv_post_parsed += all_defaults_files
```

DaanDeMeyer: Pretty nifty!

DaanDeMeyer:
Let's merge, we can deal with the tests later.

PR #514, opened 2020-08-15
CI: Add systemd-nspawn boot test

DaanDeMeyer:
Let's see how this goes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,11 +51,30 @@ jobs:
     - uses: actions/checkout@v2
     - uses: ./.github/actions/setup-mkosi
 
+    - name: Install pexpect
+      run: sudo apt-get install python3-pexpect
+
     - name: Build ${{ matrix.distro }}/${{ matrix.format }}
-      run: sudo ./mkosi
-            --debug run
-            --distribution ${{ matrix.distro }}
-            --format ${{ matrix.format }}
+      run: |
+        tee mkosi.default << EOF
+        [Distribution]
+        Distribution=${{ matrix.distro }}
+
+        [Output]
+        Format=${{ matrix.format }}
+
+        [Validation]
+        # Set to empty password if we ever get opensuse to work with empty root password.
+        # See https://github.com/systemd/mkosi/pull/514.
+        Password=root
+        EOF
+
+        sudo ./mkosi build
+
+    - name: Boot ${{ matrix.distro }}/${{ matrix.format }}
+      # photon boot gets stuck on systemd-networkd-wait-online. See https://github.com/systemd/mkosi/pull/514.
+      if: matrix.format != 'tar' && matrix.distro != 'photon'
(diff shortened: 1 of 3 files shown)


DaanDeMeyer:
The good thing is expect works. The bad thing is that it doesn't work without an open stdin for some reason. As expected simply adding this to CI is uncovering a host of bugs as well so I'm fixing those as they appear.

behrmann:
Excellent work, @DaanDeMeyer!

DaanDeMeyer:
Switched to pexpect, doesn't suffer from the stdin issue and we can stick to the same programming language used in mkosi itself.

DaanDeMeyer:
@Werkov Any idea what opensuse could be doing that prevents root login? On the other distros we simply remove pam_securetty.so from /etc/pam.d/login which fixes the issue but this doesn't seem to work on opensuse. I tried copying the file from /usr/etc/pam.d/login first but that doesn't seem to work either. Removing /etc/securetty altogether is also not a fix.

DaanDeMeyer:
@ssahani Any chance you could look at the Photon failures? Seems like systemd-networkd-wait-online is failing, even though no network related services should be started since the container is started with --private-networkd (no CAP_NET_ADMIN). See https://github.com/systemd/mkosi/pull/514/checks?check_run_id=990384883#step:6:75

behrmann:
> Any idea what opensuse could be doing that prevents root login without a root password set? 

Maybe as a workaround set a password and funnel it through expect for the login?

DaanDeMeyer:
> Maybe as a workaround set a password and funnel it through expect for the login?

Yeah did that for now. I'd prefer to get all this stuff fixed though. We're already at way too much ifs in the CI imo. Should be green now.

behrmann:
I don't think its too much of a problem setting a root password, is it? It's an edge case we shouldn't worry about. Either root accounts are disabled or they have a password and we can just set one.

DaanDeMeyer:
This is purely anecdotal, but my first experience with mkosi was trying to build an arch image for systemd, trying to boot it, not being able to login as root, trying to set --password '' and still not being able to login. That did annoy me a bit at first so I think its important this works as expected, especially since we explicitly advertise it works in the docs. We're in a pretty good spot right now though, only opensuse is still broken in terms of deleting the root password.

DaanDeMeyer:
Added one more doc fix so not going to wait out CI. I do think depending on the amount of pexpect tests we add that we should eventually run all these under pytest as well so we can run them locally as well if we want to. For now though, let's merge this as its simple and provides a lot of value already.

Werkov:
@DaanDeMeyer (a bit delayed answer) I have no experience with passwordless logins but I traced it down to missing `nullok` in common-auth:pam_unix (I didn't verify it in an actual image though).

DaanDeMeyer:
@Werkov I'll try that out. Thanks for the tip!

PR #515, opened 2020-08-16
Arch: Use Arch mirrorlist generator to get a list of mirrors.

### LINKED ISSUE #467: RFE: utilize (or allow utilizing) the Arch Linux mirrorlist
When bootstraping the Arch Linux image I sometimes end up with:
```
error: failed retrieving file 'libutil-linux-2.36-1-x86_64.pkg.tar.zst' from mirrors.kernel.org : Operation too slow. Less than 1 bytes/sec transferred the last 10 seconds
```

This could, probably, be avoided (to some degree) by utilizing the [Arch Linux mirrorlist](https://www.archlinux.org/mirrorlist/?country=all) as right now there's only the kernel.org mirror enabled by default. There's the `--mirror` option, but that supports just a single mirror. Also, other distros already have a fallback to their respective mirrorlists (like [Fedora](https://github.com/systemd/mkosi/blob/4ac0cd84e3cc4161d1bff3fd32ba03766a602da6/mkosi#L1708), [CentOS](https://github.com/systemd/mkosi/blob/4ac0cd84e3cc4161d1bff3fd32ba03766a602da6/mkosi#L1926), or [Mageia](https://github.com/systemd/mkosi/blob/4ac0cd84e3cc4161d1bff3fd32ba03766a602da6/mkosi#L1760)).

Right now, I could think of two ways how to do it:

1) Pick a several mirrors at random and enable them
This would make the behavior somewhat consistent with the other supported distributions, which already have a mirrorlist fallback. It could be done either statically (pick several mirrors from the mirrorlist and hardcode them), or automatically (fetch the mirrorlist before the bootstrap and generate a custom mirrorlist with randomly enabled mirrors) but that would make the Arch-related code bit more complex (not counting the new possible point-of-failure).

2) Expose a new option (like `--mirrorlist`) and let the user pass their own mirrorlist
This would be more flexible, since users could choose the closest/best mirrors to them.

DaanDeMeyer:
Fixes #467 and puts us in a good spot for when pacman 6.0 releases with
parallel download support.

@behrmann We should still add a --mirrorlist option but I think this is an improvement over the current situation. I did a few runs on my machine and I always got a solid mirror with this approach. I considered trying to get the country from somewhere on the system to filter the list a bit but I think we're guaranteed to run into annoying edge cases if we take that approach so let's start with this.

cc @mrc0mmand. I've seen a few comments from you about looking into mkosi for systemd centos CI and I definitely want to get it into a spot where it can be used for that so if there's anything else missing that we can fix it would be great if you could make issues for that. Most of the work I'm doing for Github Actions CI with mkosi should apply to centos CI as well but there might be some specifics that need special attention.

### DIFF
--- a/mkosi
+++ b/mkosi
@@ -2195,10 +2195,23 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
     if args.release is not None:
         sys.stderr.write("Distribution release specification is not supported for Arch Linux, ignoring.\n")
 
-    if platform.machine() == "aarch64":
-        server = f"Server = {args.mirror}/$arch/$repo"
+    if args.mirror:
+        if platform.machine() == "aarch64":
+            server = f"Server = {args.mirror}/$arch/$repo"
+        else:
+            server = f"Server = {args.mirror}/$repo/os/$arch"
     else:
-        server = f"Server = {args.mirror}/$repo/os/$arch"
+        # Instead of harcoding a single mirror, we retrieve a list of mirrors from Arch's mirrorlist
+        # generator ordered by mirror score. This usually results in a solid mirror and also ensures that we
+        # have fallback mirrors available if necessary. Finally, the mirrors will be more likely to be up to
+        # date and we won't end up with a stable release that hardcodes a broken mirror.
+        mirrorlist = os.path.join(workspace(root), "mirrorlist")
+        with urllib.request.urlopen("https://www.archlinux.org/mirrorlist/?country=all&protocol=https&ip_version=4&use_mirror_status=on") as r:
+            with open(mirrorlist, 'w') as f:
+                mirrors = r.readlines()
+                uncommented = [line.decode('utf-8')[1:] for line in mirrors]
+                f.writelines(uncommented)
+                server = f"Include = {mirrorlist}"
 
     # Create base layout for pacman and pacman-key
     os.makedirs(os.path.join(root, "var/lib/pacman"), 0o755, exist_ok=True)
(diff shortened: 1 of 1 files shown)


mrc0mmand:
> cc @mrc0mmand. I've seen a few comments from you about looking into mkosi for systemd centos CI and I definitely want to get it into a spot where it can be used for that so if there's anything else missing that we can fix it would be great if you could make issues for that. Most of the work I'm doing for Github Actions CI with mkosi should apply to centos CI as well but there might be some specifics that need special attention.

Using `mkosi` for the CentOS CI stuff would be indeed a preferable path (given the current state of the `vagrant-libvirt` plugin), unfortunately I stumbled upon one, particularly big blocker - `mkosi` generates only EFI-based images. That wouldn't be an issue, but the `edk2-ovmf` package on RHEL/CentOS 8 is shipped without the non-SecureBoot/SMM UEFI image, thus preventing us to boot any image created by `mkosi`. I guess SB with `mkosi` would be an option, but I have no idea how much of a hassle it would be (I have zero experience with SB in general).

I tried to work around this using the Fedora `edk2-ovmf` package, which worked, but I'm not sure how stable/sustainable this actually is. I'll need to probe this a bit further to see if there's anything I could do to make it work, as it would make the image preparation significantly easier.

behrmann:
@DaanDeMeyer LGTM, this is definitely an improvement, but maybe put a comment before the whole mirror list block to explain the rationale. So that when touching this again later, it's more clear why this mechanism was chosen for Arch.

DaanDeMeyer:
Let's merge this, I added a doc string to clarify the approach. I'll make a separate issue for discussing secure boot.

PR #521, opened 2020-08-20
Give prepare scripts access to SRCDIR.

DaanDeMeyer:
There are often project specific files that contain the dependencies
necessary to be installed. Make sure the prepare script has access to
these files so the logic doesn't have to be repeated in the prepare
script itself.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -913,10 +913,12 @@ local directory:
   This script has network access and may be used to install packages
   from other sources than the distro's package manager (e.g. pip, npm, ...),
   after all software packages are installed but before the image is
-  cached (if incremental mode is enabled).
-  Note that this script is executed directly in the image context with
-  the final root directory in place, without any `$SRCDIR`/`$DESTDIR`
-  setup.
+  cached (if incremental mode is enabled). This script is executed within
+  `$SRCDIR`. In contrast to a general purpose installation, it is safe to
+  install packages to the system (`pip install`, `npm install -g`) instead
+  of in `$SRCDIR` itself because the build image is only used for a single
+  project and can easily be thrown away and rebuilt so there's no risk of
+  conflicting dependencies and no risk of polluting the host system.
 
 * `mkosi.postinst` may be an executable script. If it exists it is
   invoked as the penultimate step of preparing an image, from within
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2738,6 +2738,23 @@ def set_serial_terminal(args: CommandLineArguments, root: str, do_run_build_scri
         )
 
 
+def nspawn_params_for_build_sources(args: CommandLineArguments, sft: SourceFileTransfer) -> List[str]:
+    params = []
+
+    if args.build_sources is not None:
+        params.append("--setenv=SRCDIR=/root/src")
+        params.append("--chdir=/root/src")
+        if sft == SourceFileTransfer.mount:
+            params.append("--bind=" + args.build_sources + ":/root/src")
+
+        if args.read_only:
+            params.append("--overlay=+/root/src::/root/src")
+    else:
+        params.append("--chdir=/root")
+
+    return params
+
+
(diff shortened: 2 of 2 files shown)


DaanDeMeyer:
@behrmann Have you managed to take a look at this? I'd prefer merging this before we go for a release.

behrmann:
Not yet, I'll have a look tomorrow.

mkosi.md
```diff
@@ -913,10 +913,12 @@ local directory:
   This script has network access and may be used to install packages
   from other sources than the distro's package manager (e.g. pip, npm, ...),
   after all software packages are installed but before the image is
-  cached (if incremental mode is enabled).
-  Note that this script is executed directly in the image context with
-  the final root directory in place, without any `$SRCDIR`/`$DESTDIR`
-  setup.
+  cached (if incremental mode is enabled). This script is executed within
+  `$SRCDIR`. Contrary to usual behavior, you should install packages
+  to the system (`sudo pip install`, `npm install -g`) instead of in `$SRCDIR`
```
behrmann, line 918: I think formulating this as a should is a bit strong, how about "In contrast to a general purpose installation, it is safe to install packages to the system (`sudo pip install`, `npm install -g`) instead of in `$SRCDIR`..."

mkosi/__init__.py
```diff
@@ -2966,16 +2968,14 @@ def copy_git_files(src: str, dest: str, *, source_file_transfer: SourceFileTrans
 
 
 def install_build_src(args: CommandLineArguments, root: str, do_run_build_script: bool, for_cache: bool) -> None:
-    if not do_run_build_script:
+    if (not do_run_build_script or args.build_script is None) and not args.prepare_script:
```
behrmann, line 2971: This conditional is a bit complicated. We bail out of `do_run_build_script` is false or we don't have a `build_script` unless we do have a `prepare_script`, right? Although I'm not a fan of breaking this up in general, maybe this is clearer
```python
if args.prepare_script is None:
    if not do_run_build_script or args.build_script is None:
        return
```
or maybe just
```suggestion
    if not args.prepare_script is None and (not do_run_build_script or args.build_script is None):
```

behrmann (commented):
I haven't yet used the prepare scripts scripts, since my use case doesn't really need it. So let me try to rephrase this, just so I don't misunderstand the point here.

The idea is to install dependencies with language specific package managers, which generally will need some specification of what the dependencies are from the `$SRCDIR`, before the build script runs, so that everything is properly cached., right?

I think that's a sensible goal and might be the best way, without supporting language specific package managers in mkosi directly or forcing people to specify their dependencies in two places. I do wonder, though, whether some current use case might be broken, by `/root/src` now existing when running `mkosi.prepare`, when it didn't before, or whether it could hurt us down the line when `mkosi.prepare` has different working directories depending on `args.build_sources`.

DaanDeMeyer:
> The idea is to install dependencies with language specific package managers, which generally will need some specification of what the dependencies are from the $SRCDIR, before the build script runs, so that everything is properly cached., right?

Correct.

> I think that's a sensible goal and might be the best way, without supporting language specific package managers in mkosi directly or forcing people to specify their dependencies in two places. I do wonder, though, whether some current use case might be broken, by /root/src now existing when running mkosi.prepare, when it didn't before, or whether it could hurt us down the line when mkosi.prepare has different working directories depending on args.build_sources.

This was added less than a month ago and isn't in a stable release so I don't think we have to worry about breakage here. I think we're fine with the different working directories? Its mostly about the necessary files being available if needed. If there are no build_sources, /root/src will be empty anyway so there's not much difference between running in /root or /root/src.

mkosi/__init__.py
DaanDeMeyer: Actually, we need to worry about the final image here as well. We can't just go copying in a ton of source files there without removing them.

DaanDeMeyer: Pushed a new version that just mounts the source directory in if it exists. I think mounting is fine in this case. We could go further and add an overlay so that packages have to be installed somewhere out of $SRCDIR but I'm not sure if we want this.

PR #523, opened 2020-08-21
Various CentOS 7 fixes and some features/changes

foxxx0:
This PR is the result of quite intensive debugging and fixing CentOS 7 compatibility.

It consists of:
 - optionally override GPT `first-lba` header
 - disable `metadata_csum` ext4 feature for CentOS 7 images (not supported by their old `e2fsprogs` version)
 - enable `64bit` ext4 feature on supported architectures
 - switch to `grub2-efi` for efi-bootable CentOS 7 images (their `bootctl`/`systemd-boot` is broken / too old)
 - run the postinstall script with nspawn blockdev access for bootable images

I have tried to group the changes into individual logical commits as best as possible, so reviewing should probably done commit-by-commit.

If you want, I can also split this into multiple PRs, just say so.

Any feedback is welcome and I'm happy to adjust it to meet your requirements.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -237,6 +237,14 @@ details see the table below.
   run, when specified once the incremental cache files are deleted
   too, and when specified twice the package cache is also removed.
 
+`--gpt-first-lba`
+
+: Override the first usable LBA (Logical Block Address) within the
+  GPT header. This defaults to `2048` which is actually the desired value.
+  However, some tools, e.g. the `prl_disk_tool` utility from the
+  Parallels virtualization suite require this to be set to `34`, otherwise
+  they might fail to resize the disk image and/or partitions inside it.
+
 `--bootable`, `-b`
 
 : Generate a bootable image. By default this will generate an image
@@ -708,6 +716,7 @@ which settings file options.
 | `--force`, `-f`                   | `[Output]`              | `Force=`                      |
 | `--bootable`, `-b`                | `[Output]`              | `Bootable=`                   |
 | `--boot-protocols=`               | `[Output]`              | `BootProtocols=`              |
+| `--gpt-first-lba=`                | `[Output]`              | `GPTFirstLBA=`                |
 | `--kernel-command-line=`          | `[Output]`              | `KernelCommandLine=`          |
 | `--secure-boot`                   | `[Output]`              | `SecureBoot=`                 |
 | `--secure-boot-key=`              | `[Output]`              | `SecureBootKey=`              |
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -675,6 +675,8 @@ def disable_cow(path: str) -> None:
 def determine_partition_table(args: CommandLineArguments) -> Tuple[str, bool]:
     pn = 1
     table = "label: gpt\n"
+    if args.gpt_first_lba is not None:
+        table += f"first-lba: {args.gpt_first_lba:d}\n"
     run_sfdisk = False
     args.esp_partno = None
     args.bios_partno = None
@@ -3223,6 +3225,8 @@ def insert_partition(args: CommandLineArguments,
     print_step(f'Inserting partition of {format_bytes(blob_size)}...')
 
     table = "label: gpt\n"
+    if args.gpt_first_lba is not None:
+        table += f"first-lba: {args.gpt_first_lba:d}\n"
 
     for t in old_table:
         table += t + "\n"
@@ -3833,6 +3837,7 @@ class ArgumentParserMkosi(argparse.ArgumentParser):
         'SkeletonTrees': '--skeleton-tree',
         'BuildPackages': '--build-package',
         'PostInstallationScript': '--postinst-script',
+        'GPTFirstLBA': '--gpt-first-lba',
     }
 
     fromfile_prefix_chars = '@'
@@ -3970,6 +3975,7 @@ def create_parser() -> ArgumentParserMkosi:
                        help="Do not install unified kernel images")
     group.add_argument("--with-unified-kernel-images", action=BooleanAction, default=True,
                        help=argparse.SUPPRESS)
+    group.add_argument("--gpt-first-lba", type=int, help='Set the first LBA within GPT Header', metavar='FIRSTLBA')
 
     group = parser.add_argument_group("Packages")
     group.add_argument('-p', "--package", action=CommaDelimitedListAction, dest='packages', default=[],
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2100,8 +2100,13 @@ def install_centos(args: CommandLineArguments, root: str, do_run_build_script: b
     packages += args.packages or []
 
     if not do_run_build_script and args.bootable:
-        packages += ["kernel", "systemd-udev", "dracut", "binutils"]
+        packages += ["kernel", "dracut", "binutils"]
         configure_dracut(args, root)
+        if old:
+            packages += ["grub2-efi", "grub2-tools", "grub2-efi-x64-modules", "shim-x64", "efibootmgr", "efivar-libs"]
+        else:
+            # this does not exist on CentOS 7
+            packages += ["systemd-udev"]
 
     if do_run_build_script:
         packages += args.build_packages or []
@@ -2843,10 +2848,7 @@ def run_finalize_script(args: CommandLineArguments, root: str, do_run_build_scri
         run([args.finalize_script, verb], env=env)
 
 
-def install_grub(args: CommandLineArguments, root: str, loopdev: str, grub: str) -> None:
-    if args.bios_partno is None:
-        return
-
+def write_grub_config(args: CommandLineArguments, root: str) -> None:
     kernel_cmd_line = ' '.join(args.kernel_command_line)
     grub_cmdline = f'GRUB_CMDLINE_LINUX="{kernel_cmd_line}"\n'
     os.makedirs(os.path.join(root, "etc/default"), exist_ok=True, mode=0o755)
@@ -2870,6 +2872,13 @@ def install_grub(args: CommandLineArguments, root: str, loopdev: str, grub: str)
         if args.qemu_headless:
             with open(grub_config, 'a') as f:
                 f.write("GRUB_SERIAL_COMMAND=\"serial --unit=0 --speed 115200\"\n")
+
+
+def install_grub(args: CommandLineArguments, root: str, loopdev: str, grub: str) -> None:
+    if args.bios_partno is None:
+        return
+
+    write_grub_config(args, root)
 
     # /dev/disk and /dev/block are required so grub can access the root partition UUID/PARTUUID and add it to
     # the kernel command line. If we don't do this, it adds root=/dev/loop* which breaks the boot later on
@@ -2911,6 +2920,37 @@ def install_boot_loader_clear(args: CommandLineArguments, root: str, loopdev: st
     cmdline = ["/usr/bin/clr-boot-manager", "update", "-i"]
     run_workspace_command(args, root, cmdline, nspawn_params=nspawn_params)
 
+def install_boot_loader_centos_old_efi(args: CommandLineArguments, root: str, loopdev: str) -> None:
+    nspawn_params = [
+        f"--bind-ro={loopdev}",
+        f"--bind-ro=/dev/block",
+        f"--bind-ro=/dev/disk",
+        f"--property=DeviceAllow={loopdev}",
+    ]
+    for partno in (args.esp_partno, args.bios_partno, args.root_partno, args.xbootldr_partno):
+        if partno is not None:
+            p = partition(loopdev, partno)
+            nspawn_params += [f"--bind-ro={p}", f"--property=DeviceAllow={p}"]
+
+    # prepare EFI directory on ESP
+    os.makedirs(os.path.join(root, 'efi/EFI/centos'), exist_ok=True)
+
+    # patch existing or create minimal GRUB_CMDLINE config
+    write_grub_config(args, root)
+
+    # generate grub2 efi boot config
+    cmdline = ['/sbin/grub2-mkconfig', "-o", "/efi/EFI/centos/grub.cfg"]
+    run_workspace_command(args, root, cmdline, nspawn_params=nspawn_params)
+
+    # if /sys/firmware/efi is not present within systemd-nspawn the grub2-mkconfig makes false assumptions, let's fix this
+    def _fix_grub(line: str) -> str:
+        if 'linux16' in line:
+            return line.replace('linux16', 'linuxefi')
+        elif 'initrd16' in line:
+            return line.replace('initrd16', 'initrdefi')
+        return line
+    patch_file(os.path.join(root, 'efi/EFI/centos/grub.cfg'), _fix_grub)
+
 
 def install_boot_loader(args: CommandLineArguments, root: str, loopdev: Optional[str], do_run_build_script: bool, cached: bool) -> None:
     if not args.bootable or do_run_build_script:
@@ -2921,8 +2961,14 @@ def install_boot_loader(args: CommandLineArguments, root: str, loopdev: Optional
         return
 
     with complete_step("Installing boot loader"):
-        if args.esp_partno and args.distribution != Distribution.clear:
-            run_workspace_command(args, root, ["bootctl", "install"])
+        if args.esp_partno:
+            if args.distribution == Distribution.clear:
+                pass
+            elif (args.distribution in (Distribution.centos, Distribution.centos_epel) and
+                  is_older_than_centos8(args.release)):
+                install_boot_loader_centos_old_efi(args, root, loopdev)
+            else:
+                run_workspace_command(args, root, ["bootctl", "install"])
 
         if args.bios_partno and args.distribution != Distribution.clear:
             grub = "grub" if args.distribution in (Distribution.ubuntu, Distribution.debian, Distribution.arch) else "grub2"
@@ -4588,10 +4634,13 @@ def load_args(args: CommandLineArguments) -> CommandLineArguments:
         elif args.distribution == Distribution.openmandriva:
             args.release = "cooker"
 
-    if (args.distribution in (Distribution.centos, Distribution.centos_epel) and
-        args.release == "8" and
-        args.output_format == OutputFormat.gpt_btrfs):
-        die("Sorry, CentOS 8 does not support btrfs")
+    if args.distribution in (Distribution.centos, Distribution.centos_epel):
+        epel_release = int(args.release.split('.')[0])
+        if epel_release <= 8 and args.output_format == OutputFormat.gpt_btrfs:
+            die(f"Sorry, CentOS {epel_release} does not support btrfs")
+        if epel_release <= 7 and args.bootable and 'uefi' in args.boot_protocols and args.with_unified_kernel_images:
+            die(f"Sorry, CentOS {epel_release} does not support unified kernel images. "
+                "You must use --without-unified-kernel-images.")
 
     # Remove once https://github.com/clearlinux/clr-boot-manager/pull/238 is merged and available.
     if args.distribution == Distribution.clear and args.output_format == OutputFormat.gpt_btrfs:
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -898,25 +898,39 @@ def prepare_xbootldr(args: CommandLineArguments, loopdev: Optional[str], cached:
         run(["mkfs.fat", "-nXBOOTLDR", "-F32", partition(loopdev, args.xbootldr_partno)])
 
 
-def mkfs_ext4(label: str, mount: str, dev: str) -> None:
-    run(["mkfs.ext4", "-I", "256", "-L", label, "-M", mount, dev])
+def mkfs_ext4_cmd(label: str, mount: str) -> List[str]:
+    return ["mkfs.ext4", "-I", "256", "-L", label, "-M", mount]
 
 
-def mkfs_xfs(label: str, dev: str) -> None:
-    run(["mkfs.xfs", "-n", "ftype=1", "-L", label, dev])
+def mkfs_xfs_cmd(label: str) -> List[str]:
+    return ["mkfs.xfs", "-n", "ftype=1", "-L", label]
 
 
-def mkfs_btrfs(label: str, dev: str) -> None:
-    run(["mkfs.btrfs", "-L", label, "-d", "single", "-m", "single", dev])
+def mkfs_btrfs_cmd(label: str) -> List[str]:
+    return ["mkfs.btrfs", "-L", label, "-d", "single", "-m", "single"]
 
 
 def mkfs_generic(args: CommandLineArguments, label: str, mount: str, dev: str) -> None:
+    cmdline = []
+
     if args.output_format == OutputFormat.gpt_btrfs:
-        mkfs_btrfs(label, dev)
+        cmdline = mkfs_btrfs_cmd(label)
     elif args.output_format == OutputFormat.gpt_xfs:
-        mkfs_xfs(label, dev)
+        cmdline = mkfs_xfs_cmd(label)
     else:
-        mkfs_ext4(label, mount, dev)
+        cmdline = mkfs_ext4_cmd(label, mount)
+
+    if args.output_format == OutputFormat.gpt_ext4:
+        if (args.distribution in (Distribution.centos, Distribution.centos_epel) and
+            is_older_than_centos8(args.release)):
+            # e2fsprogs in centos7 is too old and doesn't support this feature
+            cmdline += ["-O", "^metadata_csum"]
+
+        if args.architecture in ("x86_64", "aarch64"):
+            # enable 64bit filesystem feature on supported architectures
+            cmdline += ["-O", "64bit"]
+
+    run(cmdline + [dev])
 
 
 def luks_format(dev: str, passphrase: Dict[str, str]) -> None:
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2814,7 +2814,7 @@ def run_prepare_script(args: CommandLineArguments, root: str, do_run_build_scrip
         os.unlink(os.path.join(root, "root/prepare"))
 
 
-def run_postinst_script(args: CommandLineArguments, root: str, do_run_build_script: bool, for_cache: bool) -> None:
+def run_postinst_script(args: CommandLineArguments, root: str, loopdev: Optional[str], do_run_build_script: bool, for_cache: bool) -> None:
     if args.postinst_script is None:
         return
     if for_cache:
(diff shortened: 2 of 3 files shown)


mkosi
```diff
@@ -891,25 +891,49 @@ def prepare_xbootldr(args: CommandLineArguments, loopdev: Optional[str], cached:
         run(["mkfs.fat", "-nXBOOTLDR", "-F32", partition(loopdev, args.xbootldr_partno)])
 
 
-def mkfs_ext4(label: str, mount: str, dev: str) -> None:
-    run(["mkfs.ext4", "-I", "256", "-L", label, "-M", mount, dev])
+def mkfs_ext4(label: str, mount: str, dev: str, cmdline_additional: List[str]) -> None:
```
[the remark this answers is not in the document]
foxxx0: fixed

mkosi
```diff
@@ -2806,6 +2815,30 @@ def install_boot_loader_clear(args: CommandLineArguments, root: str, loopdev: st
...
+    nspawn_params = [
+        f"--bind-ro={loopdev}",
+        f"--bind-ro=/dev/block",
+        f"--bind-ro=/dev/disk",
+        f"--property=DeviceAllow={loopdev}",
+    ]
+    for partno in (args.esp_partno, args.bios_partno, args.root_partno, args.xbootldr_partno):
+        if partno is not None:
+            p = partition(loopdev, partno)
+            nspawn_params += [f"--bind-ro={p}", f"--property=DeviceAllow={p}"]
```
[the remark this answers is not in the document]
foxxx0: fixed

mkosi
```diff
@@ -2806,6 +2815,30 @@ def install_boot_loader_clear(args: CommandLineArguments, root: str, loopdev: st
...
+
+    # patch existing or create minimal GRUB_CMDLINE config
+    write_grub_config(args, root)
+
+    # generate grub2 efi boot config
+    cmdline = ['/sbin/grub2-mkconfig', "-o", "/efi/EFI/centos/grub.cfg"]
+    run_workspace_command(args, root, cmdline, nspawn_params=nspawn_params)
+
+    # if /sys/firmware/efi is not present within systemd-nspawn the grub2-mkconfig makes false assumptions, let's fix this
+    run_workspace_command(args, root, ["sed", "-i", "s/linux16 /linuxefi /g;s/initrd16 /initrdefi /g;s/ ro / /g", "/efi/EFI/centos/grub.cfg"], nspawn_params=nspawn_params)
```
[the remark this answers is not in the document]
foxxx0: fixed

mkosi
```diff
@@ -2745,7 +2746,23 @@ def run_postinst_script(args: CommandLineArguments, root: str, do_run_build_scri
...
+            nspawn_params += [
+                f"--bind-ro={loopdev}",
+                f"--bind-ro=/dev/block",
+                f"--bind-ro=/dev/disk",
+                f"--property=DeviceAllow={loopdev}"
+            ]
+            for partno in (args.esp_partno, args.bios_partno, args.root_partno, args.xbootldr_partno):
+                if partno is not None:
+                    p = partition(loopdev, partno)
+                    nspawn_params += [f"--bind-ro={p}", f"--property=DeviceAllow={p}"]
```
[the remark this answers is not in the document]
foxxx0: fixed

mkosi
```diff
@@ -889,25 +891,49 @@ def prepare_xbootldr(args: CommandLineArguments, loopdev: Optional[str], cached:
         run(["mkfs.fat", "-nXBOOTLDR", "-F32", partition(loopdev, args.xbootldr_partno)])
 
 
-def mkfs_ext4(label: str, mount: str, dev: str) -> None:
-    run(["mkfs.ext4", "-I", "256", "-L", label, "-M", mount, dev])
+def mkfs_ext4(label: str, mount: str, dev: str, cmdline_additional: List[str]) -> None:
+    cmdline = ["mkfs.ext4", "-I", "256", "-L", label, "-M", mount]
+    if cmdline_additional is not None:
```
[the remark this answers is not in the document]
foxxx0: fixed

mkosi
```diff
@@ -889,25 +891,49 @@ def prepare_xbootldr(args: CommandLineArguments, loopdev: Optional[str], cached:
...
+        cmdline += cmdline_additional
+    cmdline += [dev]
+    run(cmdline)
 
 
-def mkfs_btrfs(label: str, dev: str) -> None:
-    run(["mkfs.btrfs", "-L", label, "-d", "single", "-m", "single", dev])
+def mkfs_btrfs(label: str, dev: str, cmdline_additional: List[str]) -> None:
+    cmdline = ["mkfs.btrfs", "-L", label, "-d", "single", "-m", "single"]
+    if cmdline_additional is not None:
```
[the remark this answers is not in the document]
foxxx0: fixed

mkosi
```diff
@@ -889,25 +891,49 @@ def prepare_xbootldr(args: CommandLineArguments, loopdev: Optional[str], cached:
...
+        cmdline += cmdline_additional
+    cmdline += [dev]
+    run(cmdline)
 
 
-def mkfs_xfs(label: str, dev: str) -> None:
-    run(["mkfs.xfs", "-n", "ftype=1", "-L", label, dev])
+def mkfs_xfs(label: str, dev: str, cmdline_additional: List[str]) -> None:
+    cmdline = ["mkfs.xfs", "-n", "ftype=1", "-L", label]
+    if cmdline_additional is not None:
```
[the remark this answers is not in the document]
foxxx0: fixed

mkosi
```diff
@@ -889,25 +891,49 @@ def prepare_xbootldr(args: CommandLineArguments, loopdev: Optional[str], cached:
...
+        cmdline += cmdline_additional
+    cmdline += [dev]
+    run(cmdline)
 
 
 def mkfs_generic(args: CommandLineArguments, label: str, mount: str, dev: str) -> None:
+    cmdline_additional = []
+
+    if args.output_format == OutputFormat.gpt_ext4:
+        if (args.distribution in [Distribution.centos, Distribution.centos_epel] and
```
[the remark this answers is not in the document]
foxxx0: fixed

mkosi
```diff
@@ -889,25 +891,49 @@ def prepare_xbootldr(args: CommandLineArguments, loopdev: Optional[str], cached:
...
+            is_centos_old(args.release.split('.')[0])):
```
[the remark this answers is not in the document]
foxxx0: fixed

mkosi
```diff
@@ -1983,90 +2009,97 @@ def invoke_dnf_or_yum(args: CommandLineArguments,
...
     except ValueError:
-        new = True
+        return False
 
-    if new:
-        default_repos = install_centos_new(args, root, epel_release)
+@completestep('Installing CentOS')
+def install_centos(args: CommandLineArguments, root: str, do_run_build_script: bool) -> None:
+    release_major = args.release.split('.')[0]
+    old = is_centos_old(release_major)
```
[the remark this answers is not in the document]
foxxx0: fixed? please check again if I understood your suggestion correctly.

mkosi
```diff
@@ -2695,12 +2728,13 @@ def run_prepare_script(args: CommandLineArguments, root: str, do_run_build_scrip
...
 
 
-def run_postinst_script(args: CommandLineArguments, root: str, do_run_build_script: bool, for_cache: bool) -> None:
+def run_postinst_script(args: CommandLineArguments, root: str, loopdev: Optional[str], do_run_build_script: bool, for_cache: bool) -> None:
     if args.postinst_script is None:
         return
     if for_cache:
         return
 
+
```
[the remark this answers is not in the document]
foxxx0: fixed

mkosi
```diff
@@ -2712,7 +2746,23 @@ def run_postinst_script(args: CommandLineArguments, root: str, do_run_build_scri
 
         shutil.copy2(args.postinst_script, os.path.join(root, "root/postinst"))
 
-        run_workspace_command(args, root, ["/root/postinst", verb], network=args.with_network)
+        nspawn_params = []
+        # in order to have full blockdev access, i.e. for making grub2 bootloader changes
+        # we need to have these bind mounts for a proper chroot setup
+        if args.bootable:
+            assert loopdev is not None
```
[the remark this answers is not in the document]
foxxx0: fixed

mkosi.md
```diff
@@ -237,6 +237,13 @@ details see the table below.
   run, when specified once the incremental cache files are deleted
   too, and when specified twice the package cache is also removed.
 
+`--gpt-first-lba`
+
+: Override the first usable LBA within the GPT header. This defaults
+  to 2048 which is actually the desired value. Should you have the
+  need the override this, you can do this here and e.g. go as low
+  as `34`. (Optional)
```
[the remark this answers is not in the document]
foxxx0: done

foxxx0:
Thanks for the detailed review with very constructive feedback.
I have dropped the `epel_release` -> `release_major` rename commit and adjusted the other commits and tried to implement/fix all suggestions.

> > I've never tried to use unified kernel images with grub.
> 
> I don't think it actually supports them. I don't think we need grub-efi to work with unified kernel images either. We just fail early and tell users to use `--without-unified-kernel-images`.

For this I have followed the early call to `die()`/`warn()` within `load_args()` for Clear Linux, i.e. for the `CentOS7-bootable-uefi` config it will print a warning and set `args.with_unified_kernel_images` to `False`. I hope this is what you had in mind.

Please give it another look and let me know if there are still changes required.

Also please advise if you want to have the generic "Bootloader Choice" (grub/systemd-boot) feature implemented within this PR or if that can happen seperatly once this PR is merged.

foxxx0:
rebased onto current master

mkosi/__init__.py
```diff
@@ -4434,10 +4514,13 @@ def load_args(args: CommandLineArguments) -> CommandLineArguments:
...
-        args.release == "8" and
-        args.output_format == OutputFormat.gpt_btrfs):
-        die("Sorry, CentOS 8 does not support btrfs")
+    if args.distribution in (Distribution.centos, Distribution.centos_epel):
+        epel_release = int(args.release.split('.')[0])
+        if epel_release <= 8 and args.output_format == OutputFormat.gpt_btrfs:
+            die(f"Sorry, CentOS {epel_release} does not support btrfs")
+        if epel_release <= 7 and args.bootable and 'uefi' in args.boot_protocols and args.with_unified_kernel_images:
+            warn(f"Using --without-unified-kernel-images as CentOS {epel_release} does not support unified kernel images.")
+            args.with_unified_kernel_images = False
```
DaanDeMeyer, line 4523: I'd prefer simply failing early if `--without-unified-kernel-images` is not specified for either the bootloader option or centos 7 (depending on if you're up for adding the bootloader option).

I know I added the Clear Linux automatic fallback but I already removed that (or its in one of the open PRs) that. It adds to much magic and I'd prefer users specifically acknowledging and being okay with centos 7 not doing unified kernel images rather than automatically assume they will be ok with it just because centos 7 happens to not support them.

mkosi/__init__.py
```diff
@@ -2755,23 +2814,35 @@ def jj(line: str) -> str:
...
+    nspawn_params = nspawn_params_for_blockdev_access(args, loopdev)
 
     cmdline = ["/usr/bin/clr-boot-manager", "update", "-i"]
     run_workspace_command(args, root, cmdline, nspawn_params=nspawn_params)
 
+def install_boot_loader_centos_old_efi(args: CommandLineArguments, root: str, loopdev: str) -> None:
+    nspawn_params = nspawn_params_for_blockdev_access(args, loopdev)
+
+    # prepare EFI directory on ESP
+    os.makedirs(os.path.join(root, 'efi/EFI/centos'), exist_ok=True)
```
DaanDeMeyer, line 2828: Do we need this directory specifically? What will grub use as the default?

mkosi/__init__.py
```diff
@@ -2755,23 +2814,35 @@ def jj(line: str) -> str:
...
+
+    # patch existing or create minimal GRUB_CMDLINE config
+    write_grub_config(args, root)
+
+    # generate grub2 efi boot config
+    cmdline = ['/sbin/grub2-mkconfig', "-o", "/efi/EFI/centos/grub.cfg"]
```
DaanDeMeyer, line 2834: How does grub decide between BIOS and UEFI? Does it just look at the directory it's being installed in?

DaanDeMeyer (changes requested):
Looks great! A few more comments

mkosi/__init__.py
```diff
@@ -2041,26 +2067,37 @@ def install_centos_new(args: CommandLineArguments, root: str, epel_release: str)
...
     packages += args.packages or []
 
     if not do_run_build_script and args.bootable:
-        packages += ["kernel", "systemd-udev", "binutils"]
+        packages += ["kernel", "binutils"]
+        if old:
+            packages += ["grub2-efi", "grub2-tools", "grub2-efi-x64-modules", "shim-x64", "efibootmgr", "efivar-libs"]
+        else:
+            # this does not exist on CentOS 7
+            packages += ["systemd-udev"]
```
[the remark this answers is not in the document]
foxxx0: AFAIK the udev files are directly packaged within the `systemd` package and not split out, will check tomorrow.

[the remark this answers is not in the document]
DaanDeMeyer: Nvm, if udev is packaged it should be fine. Try to make sure you can run the built image with `mkosi qemu`. If that works then all is good.

mkosi/__init__.py
foxxx0: As stated in the comment, it actually does a check for a directory existence of `/sys/firmware/efi`, which fails at least on my build host due to the underlying OS not being booted via UEFI. This in turn causes `grub2-mkconfig` to fall back into `16bit` mode, generating the `linux16 ...` and `initrd16 ...` bootconfig lines.
That is the reason why we need to patch it afterwards.
I've searched a bit through the various manpages and `--help` outputs for the `grub2` utilities but haven't found anything that would allow explicitly forcing an install method while ignoring/overriding the internal checks. IIRC they are in `/etc/grub.d/10-linux`.

mkosi/__init__.py
```diff
@@ -2782,8 +2853,13 @@ def install_boot_loader(args: CommandLineArguments, root: str, loopdev: Optional
         return
 
     with complete_step("Installing boot loader"):
-        if args.esp_partno and args.distribution != Distribution.clear:
-            run_workspace_command(args, root, ["bootctl", "install"])
+        if args.esp_partno:
+            if args.distribution == Distribution.clear:
+                pass
+            elif args.distribution in (Distribution.centos, Distribution.centos_epel) and is_older_than_centos8(args.release):
+                install_boot_loader_centos_old_efi(args, root, loopdev)
```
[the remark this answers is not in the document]
foxxx0: Not sure when I'll have spare time to implement this. For now I'd prefer to get the CentOS 7 fixes merged and maybe open an issue tracking that bootloader choice feature.
I might give it a try sometime in the future since I'll be maintaining our mkosi build pipeline, but no promises.

mkosi/__init__.py
foxxx0: > Looks great! A few more comments

Mh, I'm unable to comment on the `mkosi.md` remarks. Indeed I seem to have forgotten to extend that, only reworded the commit message. Gonna fix that tomorrow.

mkosi/__init__.py
foxxx0: Fine by me, I'd actually prefer that too.

mkosi/__init__.py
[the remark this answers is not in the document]
DaanDeMeyer: Could you add the option and only implement it for centos 7? Early failure is fine on all the other distros. The idea behind that is that I want to keep the default behaviour for as much features as possible consistent across distros. Right now, uefi means systemd-boot for everything (except clear but that's a special one). I want to keep it that way ideally. Adding the bootloader option allows us to fail early for centos 7 if --bootloader grub is not specified so users have to add --bootloader grub explicitly for it to work.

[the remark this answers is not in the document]
DaanDeMeyer: Also, I'm really interested in what you're using mkosi for. Can you shed any light on that?

mkosi/__init__.py
DaanDeMeyer: Seems you replied to the wrong comment here (or Github did something weird)

mkosi/__init__.py
```diff
@@ -883,25 +885,49 @@ def prepare_xbootldr(args: CommandLineArguments, loopdev: Optional[str], cached:
         run(["mkfs.fat", "-nXBOOTLDR", "-F32", partition(loopdev, args.xbootldr_partno)])
 
 
-def mkfs_ext4(label: str, mount: str, dev: str) -> None:
-    run(["mkfs.ext4", "-I", "256", "-L", label, "-M", mount, dev])
+def mkfs_ext4(label: str, mount: str, dev: str, args: Optional[List[str]]) -> None:
```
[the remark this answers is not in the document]
behrmann: > Maybe do `args: List[str] = []` instead. 

This is something that should generally be avoided, because Python initialises default arguments only once, i.e. if the list were mutated inside the function, these would persist function invocations. Only primitive types make good defaults. The idiomatic way to circumvent this is assigning `None` as the default and pushing initialisation into the function body.

DaanDeMeyer:
@foxxx0 Any idea when you'll be able to take another look at this? Ideally we'd get this in before the next release

foxxx0:
@DaanDeMeyer alright so I gave it a quick rework to rebase it onto the current master, which was a bit trickier than expected since you've been quite busy. :stuck_out_tongue: 
I've seem to have resolved the conflicts, but it sure does need another proper review to check if I derped.

I have changed the warning to an early `die()` for centos7 bootable builds in case `--without-unified-kernel-images` was not specified. (along with an instructional message)

I have **not** reworked the `mkfs` part **yet**.

Also I won't be able to implement the bootloader selection stuff anytime soon.
I'll try to get the `mkfs` part done tomorrow but for the other things we either create an issue to track it (feel free to assign me to it) or we keep this PR open until I find some time.
I would highly suggest avoiding the latter, since rebasing will just get trickier and tricker and I feel like the current state should already provide viable fixes to the project best suited for inclusion before the next release.

mkosi.md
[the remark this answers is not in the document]
behrmann: This is still missing the example use case and what LBA means (I just looked it up again...)

mkosi/__init__.py
```diff
@@ -896,25 +898,49 @@ def prepare_xbootldr(args: CommandLineArguments, loopdev: Optional[str], cached:
...
 def mkfs_generic(args: CommandLineArguments, label: str, mount: str, dev: str) -> None:
+    mkfs_args = []
+
+    if args.output_format == OutputFormat.gpt_ext4:
+        if (args.distribution in (Distribution.centos, Distribution.centos_epel) and
+            is_older_than_centos8(args.release)):
+            # e2fsprogs in centos7 is too old and doesn't support this feature
+            mkfs_args += ["-O", "^metadata_csum"]
+
+        if args.architecture in ["x86_64", "aarch64"]:
```
behrmann, line 934:
```suggestion
        if args.architecture in ("x86_64", "aarch64"):
```

behrmann: Also, isn't this set automatically when the image is built on a 64bit system?

mkosi/__init__.py
```diff
@@ -2059,27 +2085,38 @@ def install_centos_new(args: CommandLineArguments, root: str, epel_release: str)
...
-
-    # Centos Repositories were changed between 7 and 8
+def is_older_than_centos8(release: str) -> bool:
+    # CentOS 7 contains some very old versions of certain libraries
+    # which require workarounds in different places.
+    # Additionally the repositories have been changed between 7 and 8
+    epel_release = release.split('.')[0]
     try:
-        new = (int(epel_release) > 7)
+        return (int(epel_release) <= 7)
```
behrmann, line 2094:
```suggestion
        return int(epel_release) <= 7
```

mkosi/__init__.py
```diff
@@ -2899,8 +2962,13 @@ def install_boot_loader(args: CommandLineArguments, root: str, loopdev: Optional
         return
 
     with complete_step("Installing boot loader"):
-        if args.esp_partno and args.distribution != Distribution.clear:
-            run_workspace_command(args, root, ["bootctl", "install"])
+        if args.esp_partno:
+            if args.distribution == Distribution.clear:
+                pass
+            elif args.distribution in (Distribution.centos, Distribution.centos_epel) and is_older_than_centos8(args.release):
```
behrmann, line 2968: This line was broken above and should be broken here as well.

mkosi/__init__.py
```diff
@@ -3970,6 +4041,7 @@ def create_parser() -> ArgumentParserMkosi:
                        help="Do not install unified kernel images")
     group.add_argument("--with-unified-kernel-images", action=BooleanAction, default=True,
                        help=argparse.SUPPRESS)
+    group.add_argument("--gpt-first-lba", type=int, help='Set the first LBA within GPT Header', metavar='GPTFIRSTLBA')
```
behrmann, line 4044:
```suggestion
    group.add_argument("--gpt-first-lba", type=int, help='Set the first LBA within GPT Header', metavar='FIRSTLBA')
```
This is a bit shorted and more closely matches the fdisk option

mkosi/__init__.py
```diff
@@ -4562,10 +4634,13 @@ def load_args(args: CommandLineArguments) -> CommandLineArguments:
...
-        args.release == "8" and
-        args.output_format == OutputFormat.gpt_btrfs):
-        die("Sorry, CentOS 8 does not support btrfs")
+    if args.distribution in (Distribution.centos, Distribution.centos_epel):
+        epel_release = int(args.release.split('.')[0])
+        if epel_release <= 8 and args.output_format == OutputFormat.gpt_btrfs:
+            die(f"Sorry, CentOS {epel_release} does not support btrfs")
+        if epel_release <= 7 and args.bootable and 'uefi' in args.boot_protocols and args.with_unified_kernel_images:
+            die(f"Sorry, CentOS {epel_release} does not support unified kernel images." \
+                " You must use --without-unified-kernel-images.")
```
behrmann, line 4643:
```suggestion
            die(f"Sorry, CentOS {epel_release} does not support unified kernel images. "
                "You must use --without-unified-kernel-images.")
```
The strings will be automatically concatenated.

behrmann (commented):
Quickly looked, but haven't checked the logic yet.

mkosi/__init__.py
```diff
@@ -2872,23 +2923,35 @@ def jj(line: str) -> str:
...
+    nspawn_params = nspawn_params_for_blockdev_access(args, loopdev)
+
+    # prepare EFI directory on ESP
+    os.makedirs(os.path.join(root, 'efi/EFI/centos'), exist_ok=True)
+
+    # patch existing or create minimal GRUB_CMDLINE config
+    write_grub_config(args, root)
+
+    # generate grub2 efi boot config
+    cmdline = ['/sbin/grub2-mkconfig', "-o", "/efi/EFI/centos/grub.cfg"]
```
DaanDeMeyer, line 2943: Don't we need to actually install grub as well? I might be missing something obvious but we seem to be missing a grub-install call no?

foxxx0: The necessary EFI binaries are copied to the ESP upon package installation. So all that's left is to give them a config file to instruct where and how to boot from.

When dealing with BIOS boot, you need to write some magic bytes to the MBR and for GPT >2TB you need to actually have that 1MiB gap within the first 2TB of the disk space so that grub can embed the required code into there because it won't fit into the MBR for this scenario.

I will test and validate that again later today, but so far it has worked on CentOS 7 + EFI + grub2 without a call to `grub2-install`.

mkosi/__init__.py
foxxx0: Depending on the locally used `e2fsprogs` release it might, but not necessarily (defaults have changed over time). For 64bit target architectures and reproducibility purposes I think it's best to explicityly enable it.

I've committed the suggested change.

mkosi/__init__.py
foxxx0: Please check if I understood your remark correctly, I was a bit confused by the wording. :confused:

mkosi/__init__.py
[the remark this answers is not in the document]
foxxx0: Reworked the `mkfs_<filesystem>()` methods to `mkfs_<filesystem>_cmd()` as suggested, please have a look.

mkosi.md
```diff
@@ -237,6 +237,16 @@ details see the table below.
...
 
+`--gpt-first-lba`
+
+: Override the first usable LBA (Logical Block Address) within the
+  GPT header. This defaults to `2048` which is actually the desired value.
+  However, some tools, e.g. the `prl_disk_tool` utility from the
+  Parallels virtualization suite require this to be set to `34`, otherwise
+  they might fail to resize the disk image and/or partitions inside it.
+  Should you have the need the override this, you can do this here and
+  go as low as `34`. (Optional)
```
behrmann, line 248: Just a nitpick, but I think this is more in line with the tone of the rest of the docs.
```suggestion
: Override the first usable LBA (Logical Block Address) within the
  GPT header. This defaults to `2048` which is the desired value,
  however, some tools, e.g. the `prl_disk_tool` utility from the
  Parallels virtualization suite require this to be set to `34`, otherwise
  they might fail to resize the disk image and/or partitions inside it.
```
The last two lines are not needed after the excellent example.

mkosi/__init__.py
```diff
@@ -896,25 +898,42 @@ def prepare_xbootldr(args: CommandLineArguments, loopdev: Optional[str], cached:
...
 
-def mkfs_btrfs(label: str, dev: str) -> None:
-    run(["mkfs.btrfs", "-L", label, "-d", "single", "-m", "single", dev])
+def mkfs_btrfs_cmd(label: str) -> List[str]:
+    return ["mkfs.btrfs", "-L", label, "-d", "single", "-m", "single"]
 
 
 def mkfs_generic(args: CommandLineArguments, label: str, mount: str, dev: str) -> None:
+    cmdline = []
+    mkfs_args = []
```
behrmann, line 915: This logic can be straightened out a bit. If you move the block were `cmdline` is set via `mkfs_*_cmd` before the one handling `mkfs_args` you can add directly to `cmdline` instead of having `mkfs_args` extra.

mkosi/__init__.py
```diff
@@ -896,25 +898,42 @@ def prepare_xbootldr(args: CommandLineArguments, loopdev: Optional[str], cached:
...
     elif args.output_format == OutputFormat.gpt_xfs:
-        mkfs_xfs(label, dev)
+        cmdline = mkfs_xfs_cmd(label)
     else:
-        mkfs_ext4(label, mount, dev)
+        cmdline = mkfs_ext4_cmd(label, mount)
+
+    cmdline += mkfs_args
+    cmdline += [dev]
+    run(cmdline)
```
behrmann, line 936:
```suggestion
    run(cmdline + mkfs_args + [dev])
```
or even just `run(cmdline + [dev])` if you take the above suggestion.

mkosi/__init__.py
```diff
@@ -2899,8 +2955,14 @@ def install_boot_loader(args: CommandLineArguments, root: str, loopdev: Optional
         return
 
     with complete_step("Installing boot loader"):
-        if args.esp_partno and args.distribution != Distribution.clear:
-            run_workspace_command(args, root, ["bootctl", "install"])
+        if args.esp_partno:
+            if args.distribution == Distribution.clear:
+                pass
+            elif args.distribution in (Distribution.centos, Distribution.centos_epel) and \
+                    is_older_than_centos8(args.release):
```
behrmann, line 2962: Almost, sorry if my previous wording was unclear. I was referring to line 918.
```suggestion
            elif (args.distribution in (Distribution.centos, Distribution.centos_epel) and 
                  is_older_than_centos8(args.release)):
```
From what I see parentheses are generally preferred in Python over line breaks, sine they explicitly group the predicate.

behrmann (commented):
Nice work!

mkosi/__init__.py
foxxx0: Great suggestions, thanks. Why didn't I think of that...

mkosi/__init__.py
foxxx0: Ah, that makes more sense indeed.

foxxx0:
Thanks @behrmann for the prompt review!
I've addressed your remarks and force-pushed.

behrmann (approved):
From my side this is good to go, if the tests pass.

DaanDeMeyer:
I would have liked to add least add the bootloader argument so we don't have to change behavior later to make things consistent but I'm a bit swamped right now so let's add this as is so we can make a new release.

PR #525, opened 2020-08-22
Cache unified kernel images and secure boot when not doing verity

DaanDeMeyer:
There's no need to re-generate these every time if we're not embedding
the root hash.


DaanDeMeyer:
Let's merge this. One caveat is that this does kinda break testing systemd initramfs changes when using cached images since the initramfs won't include the latest systemd changes (since it's cached now) but that doesn't weight against the cumulated time savings gained by doing this.

PR #528, opened 2020-08-26
Make sure we apply the btrfs stale devices fix everywhere

DaanDeMeyer:
Currently, failing build scripts can also trigger the
btrfs stale devices failure. Let's apply the fix more
generally so we always catch the error.

I'm still not sure whether this is an nspawn or a btrfs bug. It's somehow related to the default loop devices, one systemd-nspawn exhausts all of those the error dissapears. There's definitely a real fix hiding somewhere, but for now this will have to do.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -625,6 +625,19 @@ def btrfs_subvol_make_ro(path: str, b: bool = True) -> None:
     run(["btrfs", "property", "set", path, "ro", "true" if b else "false"])
 
 
+@contextlib.contextmanager
+def btrfs_forget_stale_devices() -> Generator[None, None, None]:
+    # When using cached images (-i), mounting btrfs images would sometimes fail
+    # with EEXIST. This is likely because a stale device is leftover somewhere
+    # from the previous run. To fix this, we make sure to always clean up stale
+    # btrfs devices after unmounting the image.
+    try:
+        yield
+    finally:
+        if shutil.which("btrfs"):
+            run(["btrfs", "device", "scan", "-u"])
+
+
 def image_size(args: CommandLineArguments) -> int:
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -5175,7 +5182,8 @@ def build_stuff(args: CommandLineArguments) -> None:
 
     # Make sure tmpfiles' aging doesn't interfere with our workspace
     # while we are working on it.
-    with open_close(workspace.name, os.O_RDONLY|os.O_DIRECTORY|os.O_CLOEXEC) as dir_fd:
+    with open_close(workspace.name, os.O_RDONLY|os.O_DIRECTORY|os.O_CLOEXEC) as dir_fd,\
```
behrmann, line 5185: I do agree, that the line is a bit long, but I'd put it on one line to omit the line continuation. It's a bit awkward to read otherwise.

mkosi/__init__.py
```diff
@@ -625,6 +625,19 @@ def btrfs_subvol_make_ro(path: str, b: bool = True) -> None:
...
+@contextlib.contextmanager
+def btrfs_forget_stale_devices(args: CommandLineArguments) -> Generator[None, None, None]:
+    # When using cached images (-i), mounting btrfs images would sometimes fail
+    # with EEXIST. This is likely because a stale device is leftover somewhere
+    # from the previous run. To fix this, we make sure to always clean up stale
+    # btrfs devices after unmounting the image.
+    try:
+        yield
+    finally:
+        if shutil.which("btrfs"):
```
behrmann, line 637: If we don't ask for `args.output_format == OutputFormat.gpt_btrfs` here as well, we can omit the argument in the signature, no point in passing it, if it's not used.

DaanDeMeyer: Nice catch, forgot to remove that.

PR #530, opened 2020-09-06
Refactor argument handling

DaanDeMeyer:
@behrmann Rough (and very WIP, haven't even tested yet) idea of how we could do argument parsing with dataclasses. 

TLDR; Store all argparse arguments as dataclass field metadata and retrieve them later to build the actual parser. We get fully typed arguments without having to update both the CommandLineArguments class and the Parser definition when adding a new option.

I also thought we could start formatting new files with black while we're at it.

Let me know what you think on a high level. Reviewing in detail isn't really useful yet as this will likely change quite a bit before its ready for a thorough review.

### DIFF
--- a/mkosi/args.py
+++ b/mkosi/args.py
@@ -1,0 +1,520 @@
+from dataclasses import dataclass, field, fields, Field, MISSING
+from typing import (
+    Any,
+    List,
+    TypeVar,
+    Optional,
+    overload,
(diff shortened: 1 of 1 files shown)


behrmann:
First impression is good, but I'll need to play around with it a bit. The comment delimited section in the args object hint at points where to split it up, though.

PR #540, opened 2020-12-01
setup-mkosi: Update dependencies to latest versions

DaanDeMeyer:


action.yaml
```diff
@@ -197,13 +197,13 @@ runs:
...
-      LIBREPO_VERSION: "1.12.0"
-      LIBMODULEMD_VERSION: "2.9.4"
-      LIBSOLV_VERSION: "0.7.14"
-      LIBDNF_VERSION: "0.48.0"
-      DNF_VERSION: "4.2.23"
+      LIBREPO_VERSION: "1.12.1"
+      LIBMODULEMD_VERSION: "2.10.0"
+      LIBSOLV_VERSION: "0.7.16"
+      LIBDNF_VERSION: "0.55.0"
+      DNF_VERSION: "4.2.6"
```
behrmann, line 206: Maybe it's because I haven't had coffee yet, but isn't 4.2.6 smaller than 4.2.23?

DaanDeMeyer: Woops, fixed. Didn't have coffee either. A few of the libs stopped using Github releases and switched to tags only so I always end up picking the version in the releases page while I should be looking at the tags.

DaanDeMeyer:
tdnf latest version is broken and doesn't compile so leaving that one as is.

behrmann (approved):

PR #541, opened 2020-12-01
Handle rw and ro kernel command line parameters.

drewmoseley:
Is this the right approach or is there a better way to handle this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ You can also package mkosi as a
 [zipapp](https://docs.python.org/3/library/zipapp.html) that you can deploy
 anywhere in your `PATH`. Running this will leave a `mkosi` binary in `builddir/`
 ```shell
-tools/generate_zipapp.sh
+tools/generate-zipapp.sh
 ```
 
 ## Python module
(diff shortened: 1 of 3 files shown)


behrmann:
This would certainly be a way. Could you describe (and maybe open) an issue it fixes? It would still need fixing the tests for the CLI, though.

Another way would be to set bit 60 of the partitions in accordance with the [discoverable partition spec](https://systemd.io/DISCOVERABLE_PARTITIONS/), which might be preferable, but about which I don't have strong opinions.

DaanDeMeyer:
> Another way would be to set bit 60 of the partitions in accordance with the discoverable partition spec, which might be preferable, but about which I don't have strong opinions.

We already do this I think.

I agree some more context on the issue we're fixing here would be nice. I don't see a big harm in adding this though but I'm not familiar with the details of ro/rw.

drewmoseley:
OK.  I'll take a step back and see if I can generate a minimal reproducible issue. It will likely be a few days before I can get to it.

behrmann:
> We already do this I think.
> 

Ah, true, that's what the `read_only` argument in `insert_partition` is for. It's only used for squash images and verity partitions, though and not further exposed.

drewmoseley:
I created an[ issue here.](https://github.com/systemd/mkosi/issues/543)

Based on the comment above about squashfs and verity I guess this is not surprising.

drewmoseley:
This has been adapted per the [discussion here](https://github.com/systemd/mkosi/issues/543). I also added a typo fix to the README.

I'll look into the CI failure next.

DaanDeMeyer:
This looks a lot better but I would add back 'rw' to the kernel command line if args.read_only is not set. That makes sure we don't change behavior for the majority of use cases.

DaanDeMeyer:
LGTM, thanks for the contribution!

PR #542, opened 2020-12-01
Rpm from source

DaanDeMeyer:
Possible fix for fedora issue in https://github.com/systemd/systemd/pull/17787

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,14 +18,9 @@ jobs:
           - arch
           - centos
           - centos_epel
-          - clear
           - debian
           - ubuntu
           - fedora
-          - openmandriva
-          - mageia
-          - opensuse
-          - photon
         format:
           - directory
           - tar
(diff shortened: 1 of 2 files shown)


behrmann:
This still misses the lua headers if `--with-lua` is necessary.

behrmann:
RPM says in their INSTALL file that having the lua scripting support is recommended, so installing liblua5.2-dev seems sensible.

DaanDeMeyer:
You're right, the CI was failing due to missing lua stuff in rpm. The errors are very similar to the ones I'm trying to fix in systemd so I'm pretty confident this will fix that issue as well.

behrmann:
The opensuse tar build is an interesting false positive, as it doesn't seem to build anything.

DaanDeMeyer:
I think we need to build libzypp and zypper from source now as well

DaanDeMeyer:
Nvm, decided to drop opensuse, mageia, openmandriva, clear and photon support from CI. The overhead is just too big while we don't even know if anyone is using these with mkosi. Let's stick to the popular distros and save ourselves the hassle.

Will split everything up into proper commits tomorrow.

behrmann:
Agreed, although the opensuse stuff might be worthwhile, since it's advertised on the packaging (the README talks about being mkosi being a fancy wrapper around zypper as well) and opensuse is somewhat popular in university/research settings and finance, if I remember correctly.

DaanDeMeyer:
I can't be bothered getting libzypp/zypper to build from source on Ubuntu. I've never worked with CMake builds as bad as those (I have to install nginx to build a package manager). If someone is interested in using mkosi's opensuse support and wants to make sure CI is running for mkosi, they should step up and help us out. If I have to deal any more with package manager build issues I'll lose my sanity and no work will get done at all on mkosi.

DaanDeMeyer:
Split into two commits.

behrmann (approved):

behrmann:
LGTM

DaanDeMeyer:
I'm going to do one more release with this fix in as well. Bit of release spam doesn't hurt anyone I guess.

PR #544, opened 2020-12-05
Arch: Add --needed option when installing packages with pacman

DaanDeMeyer:
Using --needed prevents packages that are already installed from
being reinstalled.


behrmann:
TIL, I didn't know `--needed` yet. :) The path to nspawn needs to be fixed, though.

DaanDeMeyer:
Lol woops, I'm debugging systemd at the same time, let me remove that.

behrmann (approved):

PR #545, opened 2020-12-06
Add --network-veth option

DaanDeMeyer:
Sets up a virtual ethernet link between the host and the container/vm.
For boot, we just pass --network-veth to nspawn. For qemu, we set up
a TAP nic with the correct ifname so that it is picked up by
systemd-networkd on the host.


behrmann:
Haven't tested this, but LGTM.

PR #546, opened 2020-12-06
Add --autologin option

DaanDeMeyer:
This is useful when using mkosi for development as systemd does since
it removes a manual step from the testing workflow.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2754,6 +2754,45 @@ def set_root_password(args: CommandLineArguments, root: str, do_run_build_script
             patch_file(os.path.join(root, 'etc/shadow'), jj)
 
 
+def set_autologin(args: CommandLineArguments, root: str, do_run_build_script: bool, for_cache: bool) -> None:
+    if do_run_build_script or for_cache or not args.autologin:
+        return
+
+    override_dir = os.path.join(root, "etc/systemd/system/console-getty.service.d")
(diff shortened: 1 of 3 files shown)


mkosi/__init__.py
```diff
@@ -2751,6 +2751,45 @@ def jj(line: str) -> str:
...
+        return
+
+    override_dir = os.path.join(root, "etc/systemd/system/console-getty.service.d")
+    os.makedirs(override_dir, mode=0o755, exist_ok=True)
+
+    override_file = os.path.join(override_dir, "autologin.conf")
+    with open(override_file, 'w') as f:
+        f.write(
+            dedent(
+                f"""
```
behrmann, line 2765: I don't think this needs to be an f-string, but it could be made a raw string for less escaping further down.

mkosi/__init__.py
```diff
@@ -2751,6 +2751,45 @@ def jj(line: str) -> str:
...
+    os.chmod(override_file, 0o644)
+
+    override_dir = os.path.join(root, "etc/systemd/system/serial-getty@ttyS0.service.d")
+    os.makedirs(override_dir, mode=0o755, exist_ok=True)
+
+    override_file = os.path.join(override_dir, "autologin.conf")
+    with open(override_file, 'w') as f:
+        f.write(
+            dedent(
+                f"""
```
behrmann, line 2782: Same as above.

behrmann (approved):

PR #548, opened 2020-12-06
Set COLUMNS and LINES to current terminal values when using --qemu-headless

DaanDeMeyer:
By default, systemd uses 80x24 dimensions for terminal size when logging
in over the serial console. Just like we set TERM already when using
--qemu-headless, set COLUMNS and LINES as well so the serial terminal
can use the full size of the screen.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2760,12 +2760,16 @@ def set_serial_terminal(args: CommandLineArguments, root: str, do_run_build_scri
     override_dir = os.path.join(root, "etc/systemd/system/serial-getty@ttyS0.service.d")
     os.makedirs(override_dir, mode=0o755, exist_ok=True)
 
+    columns, lines = shutil.get_terminal_size(fallback=(80, 24))
+
     with open(os.path.join(override_dir, "override.conf"), 'w') as f:
         f.write(
             dedent(
                 f"""
                 [Service]
                 Environment=TERM={os.getenv('TERM', 'vt220')}
+                Environment=COLUMNS={columns}
+                Environment=LINES={lines}
                 """
             )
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -2760,12 +2760,17 @@ def set_serial_terminal(args: CommandLineArguments, root: str, do_run_build_scri
     override_dir = os.path.join(root, "etc/systemd/system/serial-getty@ttyS0.service.d")
     os.makedirs(override_dir, mode=0o755, exist_ok=True)
 
+    columns = run(['tput', 'cols'], capture_output=True, text=True).stdout.strip()
+    lines = run(['tput', 'lines'], capture_output=True, text=True).stdout.strip()
+
```
behrmann, line 2765: We don't need to shell out
```suggestion
    columns, lines = shutil.get_terminal_size(fallback=(80, 24))

```

mkosi/__init__.py
```diff
@@ -2760,12 +2760,17 @@ def set_serial_terminal(args: CommandLineArguments, root: str, do_run_build_scri
...
     with open(os.path.join(override_dir, "override.conf"), 'w') as f:
         f.write(
             dedent(
                 f"""
                 [Service]
                 Environment=TERM={os.getenv('TERM', 'vt220')}
+                Environment=COLUMNS={colums}
```
behrmann, line 2772:
```suggestion
                Environment=COLUMNS={columns}
```

mkosi/__init__.py
DaanDeMeyer: Nice!

behrmann (approved):

PR #551, opened 2020-12-06
Replace "-boot c" QEMU option with "-nodefaults -serial mon:stdio"

DaanDeMeyer:
This doesn't actually change anything. Boot order when using
OVMF can only be changed by editing OVMF_VARS.fd.


DaanDeMeyer:
Updated to use -nodefaults instead. Only do this for --qemu-headless for now to avoid too much impact on regular qemu usage.

DaanDeMeyer:
Merging this one since CI is green and I'm pretty sure I'm (still) the only one using --qemu-headless. If it turns out -nodefaults removed some other necessary devices we can always add them back as well.

PR #552, opened 2020-12-06
Add --hostonly-initrd option

DaanDeMeyer:
Enabling --hostonly speeds up QEMU boots by roughly 2 seconds on
my machine.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -365,6 +365,12 @@ details see the table below.
 : If specified, mkosi does not build unified kernel images and instead installs kernels with a separate
   initrd and boot loader config to the efi or bootloader partition.
 
+`--hostonly-initrd`
+
+: If specified, mkosi will run the tool to create the initrd such that a non-generic initrd is created that
+  will only be able to run on the system mkosi is run on. Currently mkosi uses dracut for all supported
+  distributions except Clear Linux and this option translates to enabling dracut's hostonly option.
+
 `--no-chown`
 
 : By default, if `mkosi` is run inside a `sudo` environment all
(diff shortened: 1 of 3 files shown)


behrmann:
I'm sympathetic to exposing this, as I think it's one of the more important dracut switches, that people might want to set. On the other hand, could it bite us if some years down the road something other than dracut comes around and this could also be done with `--extra-tree`. Without having thought too much about it, I'm +0 on this.

DaanDeMeyer:
I think the concept of hostonly isn't dracut specific though. We could remove the dracut specifics from the documentation and use it as the option to tell mkosi to produce images that will only run on the current host machine.

behrmann:
Definitely, it's not a feature that would be super-specific to dracut and should be something that most tools for generating initrd's should support.

mkosi.md
```diff
@@ -365,6 +365,11 @@ details see the table below.
 : If specified, mkosi does not build unified kernel images and instead installs kernels with a separate
   initrd and boot loader config to the efi or bootloader partition.
 
+`--hostonly`
```
behrmann, line 368:
```suggestion
`--hostonly-initrd`
```
This would make it clearer, without reading the docs, what this is for. It would need renaming of `args.hostonly` in either the whole PR or at least the argparse bits and docs, though.

mkosi.md
```diff
@@ -365,6 +365,11 @@ details see the table below.
...
+
+: If specified, The dracut hostonly option is enabled which tells dracut to only include drivers for the
+  current system in the initramfs. This can speed up boot times when using mkosi's qemu verb.
```
behrmann, line 371:
```suggestion
: If specified, mkosi will run the tool to create the initrd such that a non-generic
  initrd is created that will only be able to run on the system mkosi is run on.
  Currently mkosi uses dracut for all supported distributions and this option translates
  to enabling dracut's hostonly option.
```

DaanDeMeyer:
Addressed comments and CI is green so merging.

PR #553, opened 2020-12-06
Only print summary when summary verb is used

DaanDeMeyer:
Printing the entire summary every time we're building the image
is unnecessarily verbose. Let's only print the summary when the
user asks for it.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -5550,7 +5550,7 @@ def run_verb(args: CommandLineArguments) -> None:
     if args.verb == "build":
         check_output(args)
 
-    if args.verb == "summary" or needs_build(args):
+    if args.verb == "summary" or (needs_build(args) and need_cache_images(args)):
         print_summary(args)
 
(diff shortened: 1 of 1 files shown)


behrmann:
I agree the output is extremely verbose, but this is a rather drastic change, isn't it? Currently the summary is only shown on the initial build and even though it's not needed in most cases, it's nice - or at least doesn't hurt - to have the output in CI to see that the input is actually what's being done without having to rerun the CI.

DaanDeMeyer:
What if I change this to not print the summary when re-using cached images?

behrmann:
I think that would be a good change.

DaanDeMeyer:
Now we don't print the summary if we're reusing cache images. That way, we keep the summary intact for anyone running an initial build but for the code -> compile -> test loop (during which it'd be insane not to use cache images), we remove the extra verbosity from the summary.

behrmann (approved):

PR #554, opened 2020-12-06
qemu: Add if=virtio option.

DaanDeMeyer:
Speeds up boot from 12s to 6s.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -5472,7 +5472,7 @@ def run_qemu(args: CommandLineArguments) -> None:
         cmdline += ["-drive", f"if=pflash,format=raw,readonly,file={firmware}"]
 
     cmdline += [
-        "-drive", f"format={'qcow2' if args.qcow2 else 'raw'},file={args.output}",
+        "-drive", f"format={'qcow2' if args.qcow2 else 'raw'},file={args.output},if=virtio",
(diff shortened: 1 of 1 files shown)


behrmann:
I guess it's worth to try this out. Starting qemu directly might not affect too many people and if it makes problems it can always be hidden behind a switch.

DaanDeMeyer:
I certainly hope no one's using `mkosi qemu` to run production QEMU VMs. This does require a few drivers on the guest image to be installed though so I'm going to do a quick check for ubuntu, debian, fedora and centos to make sure those drivers are available there.

DaanDeMeyer:
These all boot fine for me so I'm guessing we're good. I think the dracut qemu module pulls in the necessary drivers.

behrmann:
Indeed. For later reference qemu pulls in the following modules at the moment via `90qemu`
- ata_piix ata_generic 
- pata_acpi cdrom 
- sr_mod ahci 
- virtio_blk 
- virtio virtio_ring 
- virtio_pci 
- virtio_scsi virtio_console 
- virtio_rng 
- spapr-vscsi 
- qemu_fw_cfg
and 
- virtio_net 
- e1000 
- 8139cp 
- pcnet32 
- e100 
- ne2k_pci
via `90qemu-net`. 

I think we should be on the safe side with this already, but if somethings missing we can add it if bug reports come rolling in.

PR #555, opened 2020-12-06
qemu: Add -cpu host option

DaanDeMeyer:
We don't need compatibility for our use case so let's squeeze out
all the performance we can.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -5470,6 +5470,9 @@ def run_qemu(args: CommandLineArguments) -> None:
 
     cmdline += ["-smp", "2", "-m", "1024"]
 
+    if has_kvm:
+        cmdline += ["-cpu", "host"]
+
     if "uefi" in args.boot_protocols:
         cmdline += ["-drive", f"if=pflash,format=raw,readonly,file={firmware}"]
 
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -5426,14 +5426,16 @@ def run_shell(args: CommandLineArguments) -> None:
...
     ARCH_BINARIES = {'x86_64': 'qemu-system-x86_64',
                      'i386': 'qemu-system-i386'}
     arch_binary = ARCH_BINARIES.get(platform.machine(), None)
+    accel = "kvm" if has_kvm else "tcg"
     if arch_binary is not None:
-        cmdlines += [[arch_binary, '-machine', 'accel=kvm:tcg']]
+        cmdlines += [[arch_binary, '-machine', f'accel={accel}']]
     cmdlines += [
-        ['qemu', '-machine', 'accel=kvm:tcg'],
+        ['qemu', '-machine', f'accel={accel}'],
```
behrmann, line 5438: Is this strictly necessary? The man page says that `-machine accel=kvm:tcg` uses the latter accelerators only if the former fails to initialise.

DaanDeMeyer: It avoids a warning from qemu. I figured since I was adding the has_kvm thing anyway for -cpu host I might as well use it for the accel as well. If the check we use isn't entirely right in some case, qemu will break regardless because -cpu host can't be used with anything but kvm so we don't gain anything by relying on the fallback logic of qemu here.

behrmann: That makes sense.

behrmann (approved):

PR #557, opened 2020-12-07
Arch: Switch back to using a single call to pacman

DaanDeMeyer:
We shouldn't be working around dependency issues in Arch. I'm not
getting any failures using this anymore right now but if any turn
up, we should file bug reports for the relevant Arch packages to
add the necessary dependencies. Now that base is a metapackage and
not a group, the burden to add it as a dependency should be very
minimal.


behrmann:
Agreed.

The boot failures seem unrelated to *this*, but qemu is timing out for Centos and Fedora.

DaanDeMeyer:
Hmm, there seems to have been an update specific to BIOS there that breaks things. Will investigate later today.

PR #560, opened 2020-12-09
Add a vga device for --qemu-headless when using BIOS with Fedora/CentOS

### LINKED ISSUE #559: CI failures for Fedora and CentOS
I can reproduce this locally with Fedora 33:

```
SeaBIOS (version ArchLinux 1.14.0-1)
Booting from Hard Disk...
...
```

When starting QEMU, it gets stuck in a boot loop and prints this message every time. I'm super confused as to why it's using version ArchLinux on Fedora. Maybe that has something to do with the failure.

DaanDeMeyer:
Fixed #559


behrmann:
Maybe it would make sense to add this to all distros instead of doing this specifically for Fedora and CentOS, if the other's work with this that is. This would help to keep the distro specifics as small as possible.

DaanDeMeyer:
I'll see if it has any overhead. If not, we can do it for all distros.

DaanDeMeyer:
Actually, It's just BIOS, I don't really care too much about performance details there. Let's just do it for all distros.

PR #564, opened 2020-12-11
Allow build script to read from stdin when running in a script

DaanDeMeyer:
When running language servers like clangd via a mkosi build-script,
we need clangd to be able to read from stdin to interact with the
editor. By default, systemd-nspawn doesn't do this for security
issues but in this case we trust the build script so we can specify
--pipe to enable the build script to read from stdin.


behrmann:
This is a nicely self-contained change. :) Great this finally fulfills your dev server ideas!

PR #565, opened 2020-12-11
Add --include-directory option

DaanDeMeyer:
--include-directory makes the header files installed in the container
available on the host. This is useful when running a language server
such as clangd in the mkosi container. If we define a path mapping
in clangd from /usr/include to the include directory, we can jump
to all the header files of external dependencies in the include
directory right from the editor.

Without this, we can't actually use jump to declaration for any third-party
library headers whose packages are not installed on the host system. By
making the include files from the container available on the host and defining
a path mapping, clangd will jump right into the specified --include-directory
when using go to declaration on any symbol belonging to a third-party library
we're using.

Updated clangd config that makes this work:

```
#!/usr/bin/env sh
exec clangd \
        --compile-commands-dir=/root/build \
        --path-mappings=/home/daan/projects/systemd=/root/src,/home/daan/projects/systemd/mkosi.builddir=/root/build,/home/daan/projects/systemd/mkosi.includedir=/usr/include \
        --header-insertion=never
```

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -494,6 +494,17 @@ details see the table below.
   but a directory `mkosi.builddir/` exists in the local directory it
   is automatically used for this purpose (also see below).
 
+`--include-directory`
+
+: Takes a path of a directory to use as the include directory. This
+  directory is mounted at /usr/include when building the build image
+  and when running the build script. This means all include files
+  installed to /usr/include will be stored in this directory. This is
+  useful to make include files available on the host system for use by
+  language servers to provide code completion. If this option is not
+  specified, but a directory `mkosi.includedir/` exists in the local
+  directory, it is automatically used for this purpose (also see below).
+
 `--build-package=`
 
 : Similar to `--package=`, but configures packages to install only in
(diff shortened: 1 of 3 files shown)


behrmann:
So the idea of this PR is to have the include directory be back channel for the image to propagate build artefacts (in this case header files) back to the build host?

If that is correct this could be a more generally useful feature, than this somewhat narrow usage, and could be named appropriately.

DaanDeMeyer:
What would be other use cases? I think the source, build and /usr/include directories already cover many of the possible use cases. Is there really a need to export any other information out of the build image?

Maybe there's opportunity for more use cases when using `mkosi boot` but we already allow passing custom nspawn arguments to boot so users can just use --bind directly there.

behrmann:
Another use case might be to extract the binary itself, but maybe I'm misjudging the use case.

DaanDeMeyer:
This is already possible via the --build-directory option. What might be possible is that there's SDK packages from vendors that don't behave well and store their header files outside of /usr/include.

In the future, we could extend this option to also support ':' separated paths that contain the directory in the host and the directory in the container to mount to. But let's leave that for the future, since I don't think this option will be used extensively overall and we shouldn't support complicated use cases that we're not sure anyone will even use.

behrmann:
Right, I had forgotten about `--build-directory`, then this is complementary to this, makes sense.

mkosi.md
```diff
@@ -494,6 +494,17 @@ details see the table below.
...
   is automatically used for this purpose (also see below).
 
+`--include-directory`
+
+: Takes a path of a directory to use as the include directory. This
+  directory is mounted at /usr/include when building the build image
+  and when running the build script. This means all include files
+  installed to /usr/include will be stored in this directory. This is
+  useful to make include files available on the host system for use by
+  language servers to provide intellisense. If this option is not
```
behrmann, line 504:
```suggestion
  language servers to provide code completion. If this option is not
```

PR #566, opened 2020-12-11
Redirect non-redirected subprocesses output to stderr except build-script

DaanDeMeyer:
When running a language server via mkosi, communication with the editor
is done via mkosi's stdin/stdout. If the client reads output from mkosi's
subprocesses instead of LSP json messages from stdout, it won't recognize
the output and crash. To make the language server integration work when the
build image needs to be built from scratch, we redirect all non-redirected
output from subprocesses to stderr except for the build-script. This way,
we can get progress output from mkosi right in the editor without breaking
the language server integration.

The split was already quite arbitrary in the first place. All of mkosi's messages
go to stderr and most subprocess output went to stdout. Instead, let's define
the split between stdout/stderr a bit more so we can take advantage of it.
Only the build script gets to write to stdout which simplifies language server
usage with mkosi.

(This should be the last language server specific change)

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -181,17 +181,30 @@ def delay_interrupt() -> Generator[None, None, None]:
         if interrupted:
             die("Interrupted")
 
+# Borrowed from https://github.com/python/typeshed/blob/3d14016085aed8bcf0cf67e9e5a70790ce1ad8ea/stdlib/3/subprocess.pyi#L24
+_FILE = Union[None, int, IO[Any]]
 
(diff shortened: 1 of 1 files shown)


behrmann:
I like this change. It's a rather big one and I hope nobody relied on the current behaviour in any way, but I think this 
> The split was already quite arbitrary in the first place. All of mkosi's messages go to stderr and most subprocess output went to stdout. Instead, let's define the split between stdout/stderr a bit more so we can take advantage of it. Only the build script gets to write to stdout which simplifies language server usage with mkosi.

says it perfectly. Now the output has clear semantics and if need be we can probably accommodate changes to it, now that there's a defined baseline.

PR #567, opened 2020-12-11
Action + CI improvements

DaanDeMeyer:
I'm also going to have us not install the master branch version anymore from the action once the v9 tag is there. Since the action is now being used in systemd, we should probably not have it pull in the latest code from master.


behrmann:
Sounds good to me.

PR #569, opened 2020-12-11
Use unsafe-io with dpkg and restore compatiblity with nspawn v241 and lower

bluca:
Couple of fixes added while using mkosi on my Debian stable machine, to build Debian/Ubuntu images.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2206,6 +2206,11 @@ def install_debian_or_ubuntu(args: CommandLineArguments,
         with open(dpkg_conf, "w") as f:
             f.writelines(f'path-exclude {d}/*\n' for d in doc_paths)
 
+    # Either the image builds or it fails and we restart, we don't need safety fsyncs when bootstrapping
+    dpkg_conf = os.path.join(root, "etc/dpkg/dpkg.cfg.d/unsafe_io")
+    with open(dpkg_conf, "w") as f:
+        f.write('force-unsafe-io\n')
+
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -2206,6 +2206,11 @@ def install_debian_or_ubuntu(args: CommandLineArguments,
         with open(dpkg_conf, "w") as f:
             f.writelines(f'path-exclude {d}/*\n' for d in doc_paths)
 
+    # Either the image builds or it fails and we restart, we don't need safety fsyncs when bootstrapping
+    dpkg_conf = os.path.join(root, "etc/dpkg/dpkg.cfg.d/unsafe_io")
+    with open(dpkg_conf, "w") as f:
+        f.write('force-unsafe-io\n')
```
DaanDeMeyer, line 2212: This is exactly the kind of stuff I was hoping you'd contribute! I had no idea we could actually make dpkg faster!

DaanDeMeyer:
LGTM

mkosi/__init__.py
DaanDeMeyer: Wait, maybe I pulled the trigger a bit too quickly here. Should we disable unsafe_io again after bootstrapping? I assume we'd want it enabled if a user ever does regular updates in the image after building it?

bluca: Sure, can do that. I was mainly considering the use case of testing/CI. Will send a PR later.

PR #571, opened 2020-12-12
CI: Install mkosi explicitly

DaanDeMeyer:
Since we'll soon be installing the latest stable version with the
action, let's modify CI already to do an explicit install that
installs mkosi with the latest changes from the PR.


DaanDeMeyer:
Failure is unrelated.

PR #573, opened 2020-12-12
Add --install-directory option

DaanDeMeyer:
While binaries can already be retrieved using the --build-directory
option. Sometimes it's a lot easier to have access to the installed
contents instead. --install-directory allows accessing the contents
of /root/dest from the build container outside of the build image.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3133,7 +3133,8 @@ def install_build_src(args: CommandLineArguments, root: str, do_run_build_script
                                                 os.path.basename(args.output_dir)+"/" if args.output_dir else "mkosi.output/",  # NOQA: E501
                                                 os.path.basename(args.cache_path)+"/" if args.cache_path else "mkosi.cache/",  # NOQA: E501
                                                 os.path.basename(args.build_dir)+"/" if args.build_dir else "mkosi.builddir/",  # NOQA: E501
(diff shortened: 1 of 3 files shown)


mkosi/__init__.py
```diff
@@ -3126,7 +3126,8 @@ def install_build_src(args: CommandLineArguments, root: str, do_run_build_script
                                                 os.path.basename(args.output_dir)+"/" if args.output_dir else "mkosi.output/",  # NOQA: E501
                                                 os.path.basename(args.cache_path)+"/" if args.cache_path else "mkosi.cache/",  # NOQA: E501
                                                 os.path.basename(args.build_dir)+"/" if args.build_dir else "mkosi.builddir/",  # NOQA: E501
-                                                os.path.basename(args.include_dir)+"/" if args.include_dir else "mkosi.includedir/")  # NOQA: E501
+                                                os.path.basename(args.include_dir)+"/" if args.include_dir else "mkosi.includedir/",  # NOQA: E501
+                                                os.path.basename(args.install_dir)+"/" if args.install_dir else "mkosi.installdir/")
```
behrmann, line 3130:
```suggestion
                                                os.path.basename(args.install_dir)+"/" if args.install_dir else "mkosi.installdir/")  # NOQA: E501
```

behrmann:
I guess this should also be added to the summary.

PR #574, opened 2020-12-12
Empty include directory before installing to it

DaanDeMeyer:
Pacman (and probably other package managers as well) expect
/usr/include to be empty when we're bootstrapping so let's make
sure we empty the directory before installing to it.


DaanDeMeyer:
Added a second commit to fix permission errors as well.

PR #576, opened 2020-12-14
Explicitly specify dracut filesystem when using --hostonly-initrd

DaanDeMeyer:
It seems dracut includes the host's filesystem when building a
hostonly image and not the image's filesystem. This causes issues
when these two don't match. Include the correct one explicitly
as a workaround.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -296,6 +296,16 @@ class OutputFormat(enum.Enum):
         "The output format can be 'minimized'"
         return self in (OutputFormat.gpt_ext4, OutputFormat.gpt_btrfs)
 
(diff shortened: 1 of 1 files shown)


behrmann:
This is indeed a very good fix, but I think the `filesystem` function might be better places as a method `needed_kernel_module` directly on the `OutputFormat` enum.

DaanDeMeyer:
Can I ask you to take a look at #566 and #511 as well? I want to get another release out so we can adopt some of the new options in systemd.

behrmann:
Yes, definitely, I've had them open for a while now, but haven't gotten around to saying anything substantial yet.

PR #578, opened 2020-12-14
Run finalize script with OUTPUTDIR environment variable set

### LINKED ISSUE #575: Output dir variable in the finalize script
In my case i need to copy some artifacts, like ssh key, from image to output dir. I did not find any way to do this expect the hardcode mkosi.output in the finalize script. Environment variable like $OUTPUTDIR will be very useful to this case

DaanDeMeyer:
Fixes #575.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2896,6 +2896,10 @@ def run_postinst_script(args: CommandLineArguments, root: str, loopdev: Optional
         os.unlink(os.path.join(root, "root/postinst"))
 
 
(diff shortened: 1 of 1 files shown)


behrmann (approved):

behrmann:
I do agree, that this is a nice little change. I do hope though, that we don't clobber somebody's `OUTPUTDIR` variable in a finalize script, since that's a common name (at least one that crops up in my shell scripts). Maybe in the long run, we could slowly move to offering them under a `MKOSI_` prefix.

PR #581, opened 2020-12-17
Some message printing improvements, flake8 warnings

whot:
No real functional changes here, just a bit of polish for the output and less flake8 complaints.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1561,7 +1561,7 @@ def run_workspace_command(args: CommandLineArguments,
     if result.returncode != 0:
         if 'workspace-command' in arg_debug:
             run(cmdline, check=False)
-        die(f"Workspace command {' '.join(cmd)} returned non-zero exit code {result.returncode}.")
+        die(f"Workspace command `{' '.join(cmd)}` returned non-zero exit code {result.returncode}.")
 
 
 def check_if_url_exists(url: str) -> bool:
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -53,6 +53,8 @@ from typing import (
 )
 from types import FrameType
 
+from .printer import MkosiPrinter
+
 
 __version__ = '8'
 
@@ -148,13 +150,9 @@ class MkosiException(Exception):
     """Leads to sys.exit"""
 
 
-def print_step(text: str) -> None:
-    sys.stderr.write("‣ \033[0;1;39m" + text + "\033[0m\n")
-
-
 def print_running_cmd(cmdline: Iterable[str]) -> None:
-    sys.stderr.write("‣ \033[0;1;39mRunning command:\033[0m\n")
-    sys.stderr.write(" ".join(shlex.quote(x) for x in cmdline) + "\n")
+    MkosiPrinter.print_step("Running command:")
+    MkosiPrinter.print_step(" ".join(shlex.quote(x) for x in cmdline) + "\n")
 
 
 @contextlib.contextmanager
@@ -194,7 +192,7 @@ def run(cmdline: List[str],
         stderr: _FILE = None,
         **kwargs: Any) -> CompletedProcess:
     if 'run' in arg_debug:
-        sys.stderr.write('+ ' + ' '.join(shlex.quote(x) for x in cmdline) + '\n')
+        MkosiPrinter.info('+ ' + ' '.join(shlex.quote(x) for x in cmdline))
 
     if not stdout and not stderr:
         # Unless explicit redirection is done, print all subprocess output on stderr since we do so as well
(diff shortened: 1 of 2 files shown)


mkosi/__init__.py
```diff
@@ -592,7 +607,7 @@ def complete_step(text: str, text2: Optional[str] = None) -> Generator[List[Any]
                 Callable[[CommandLineArguments, str, bool], None],
                 Callable[[CommandLineArguments, str, bool, bool], None],
             ]
-)
+            )
```
behrmann, line 610: Not that this is at the moment an issue, but this is the difference between black and flake8 and at some point I would like to see black as consistent formatting in this codebase and I'd rather switch this warning off than accomodate flake8 here.

mkosi/__init__.py
```diff
@@ -148,13 +148,24 @@ class MkosiException(Exception):
     """Leads to sys.exit"""
 
 
+class ColorPrinter():
+    isatty = sys.stderr.isatty()
+    prefix = '‣ '
+    bold = '\033[0;1;39m' if isatty else ''
+    reset = '\033[0m' if isatty else ''
```
behrmann, line 155: This is an an excellent change, but I think it could go a bit further. I think it would make sense to subsume all calls to `sys.stderr.write` (mostly in the summary) in a `MkosiPrinter` class and maybe also get rid of the calls to `print()` here and there, so that everything goes through the same machinery.

behrmann (commented):
Thank you for the contribution!

DaanDeMeyer:
I'd prefer if we enforced this in CI but most of these changes look good so let's merge.

@behrmann I'd merge a PR reformatting all the code with black and adding it to the CI so we don't regress. If we do that, we should also add a file that can be passed to the --ignore-revs-file git blame option containing that commit so we can still have an effective git blame.

mkosi/__init__.py
DaanDeMeyer: I tend to agree here, if we do this, let's do it properly.

behrmann:
> @behrmann I'd merge a PR reformatting all the code with black and adding it to the CI so we don't regress. If we do that, we should also add a file that can be passed to the --ignore-revs-file git blame option containing that commit so we can still have an effective git blame.

I'll get to that tomorrow.

mkosi/__init__.py
whot: any thoughts on going through the python logging facilities? That'd be an easy option to control formatting and have a distinct stream of outputs.

behrmann: Definitely open to it. I think I still have a branch lying around where I had a go at it, but I was unhappy with the log levels and messing around with handlers to remove the terminal control sequences when not needed. I think moving everything to a single printer would be a sensible step to move to logging later.

whot:
I've updated this now with an `MkosiPrinter` class, replacing all the `sys.stderr.write` calls in the file. This is almost ready for replacing with python logging in the future, the structure is quite similar after all.

A heads-up though: I'll be afk until the new year from tonight onwards, so feel free to ignore this until then if it's not merge-ready.

mkosi/__init__.py
whot: fixed in the latest revision with the `MkosiPrinter` class, we can swap that out for a python logger in the future.

mkosi/__init__.py
```diff
@@ -148,13 +148,36 @@ class MkosiException(Exception):
     """Leads to sys.exit"""
 
 
-def print_step(text: str) -> None:
-    sys.stderr.write("‣ \033[0;1;39m" + text + "\033[0m\n")
+class MkosiPrinter():
```
DaanDeMeyer, line 151: If we're doing refactoring anyway, let's take advantage of all the work @behrmann did to allow splitting mkosi up into multiple files and move MkosiPrinter into its own file.

mkosi/__init__.py
```diff
@@ -148,13 +148,36 @@ class MkosiException(Exception):
...
+    isatty = out_file.isatty()
+
+    bold = "\033[0;1;39m" if isatty else ""
+    red = "\033[31;1m" if isatty else ""
+    reset = "\033[0m" if isatty else ""
+
+    prefix = "‣ "
+
+    @classmethod
+    def _p(cls, text: str) -> None:
```
DaanDeMeyer, line 162: Saving letters doesn't matter here, better to call it _print

DaanDeMeyer:
Overall this looks pretty good!

mkosi/__init__.py
whot: done

mkosi/__init__.py
whot: moved to `printer.py`

DaanDeMeyer (approved):
LGTM but let's wait for @behrmann's approval as well.

behrmann (approved):
LGTM

PR #582, opened 2020-12-17
Allow disabling unified kernel images on BIOS for all distros

### LINKED ISSUE #580: Unable to create bootable (bios) Debian image
I'm unable to create any qemu-bootable Debian image using `BootProtocols=bios`. 

`mkosi.default` file looks like this:
```ini
[Distribution]
Distribution=debian
Release=testing

[Output]
Format=gpt_ext4
Bootable=yes
BootProtocols=bios

[Partitions]
RootSize=2G
```
Same bug appears for Debian stable/unstable  and the recent Ubuntus (tested the last two, I think). In all cases, the issue is that there is no initrd file in `/boot` which then unsurprisingly fails during boot. The same `mkosi.default` file with Fedora 33 as distribution works just fine.

Host system is Fedora 33, reproducible with:
```
$ sudo mkosi --force
$ qemu-system-x86_64 -machine accel=kvm \
                       -smp 2 -m 1024 \
                       -drive format=raw,file=image.raw \
                       -device virtio-net-pci,netdev=net0 \
                       -netdev user,id=net0,hostfwd=tcp::5555-:22 \
                       -display sdl \
                       -serial file:console.out
```

`git bisect` says this started with 46ceca8ffc8a2edc0539e01eadc9d99357dd247 which (handwavy guess) could explain why the initrd never appears, maybe the partition isn't mounted correctly?

DaanDeMeyer:
Fixes #580.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,10 +139,7 @@ jobs:
       if: matrix.format != 'directory' &&
           matrix.format != 'gpt_squashfs' &&
           matrix.format != 'plain_squashfs' &&
-          matrix.format != 'tar' && (
-          matrix.distro == 'arch' ||
-          matrix.distro == 'centos' ||
-          matrix.distro == 'fedora')
+          matrix.format != 'tar'
       run: |
         tee mkosi.default << EOF
         [Distribution]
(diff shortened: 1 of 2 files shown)


DaanDeMeyer:
We're green so let's merge.

PR #583, opened 2020-12-18
Add incremental mode support for tar,directory and plain_squashfs

mmorvai:
Add incremental mode support for tar,directory and plain_squashfs by caching the root tree and reusing it. Everything was already in the code just needed to "activate" it.

The only major change is, that plain_squashfs is handled similarly to tar instead of being treated like "generated images" (minimized and gpt_squashfs images).

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -234,7 +234,7 @@ class CommandLineArguments(argparse.Namespace):
         """Returns whether this configuration means we need to generate a file system from a prepared tree
 
         This is needed for anything squashfs and when root minimization is required."""
-        return self.minimize or self.output_format.is_squashfs()
+        return self.minimize or self.output_format == OutputFormat.gpt_squashfs
 
 
 class SourceFileTransfer(enum.Enum):
@@ -3222,10 +3222,12 @@ def make_tar(args: CommandLineArguments,
     return f
 
 
-def make_squashfs(args: CommandLineArguments, root: str, for_cache: bool) -> Optional[BinaryIO]:
+def make_squashfs(args: CommandLineArguments, root: str, do_run_build_script: bool, for_cache: bool) -> Optional[BinaryIO]:
     if not args.output_format.is_squashfs():
         return None
     if for_cache:
+        return None
+    if do_run_build_script and args.output_format == OutputFormat.plain_squashfs:
         return None
 
     command = args.mksquashfs_tool[0] if args.mksquashfs_tool else 'mksquashfs'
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -5229,18 +5240,18 @@ def build_image(args: CommandLineArguments,
                 *,
                 do_run_build_script: bool,
                 for_cache: bool = False,
-                cleanup: bool = False) -> Tuple[Optional[BinaryIO], Optional[BinaryIO], Optional[str]]:
+                cleanup: bool = False) -> Tuple[Optional[BinaryIO], Optional[BinaryIO],  Optional[BinaryIO], Optional[str]]:
```
behrmann, line 5243:
```suggestion
                cleanup: bool = False) -> Tuple[Optional[BinaryIO], Optional[BinaryIO], Optional[BinaryIO], Optional[str]]:
```

behrmann (commented):
Thanks for this contribution! I haven't thought through the logic yet, just scrolled through the PR once.

DaanDeMeyer:
We should also add incremental mode to CI to make sure this works in most scenarios. Could you add a commit that adds the --incremental switch to every invocation of mkosi in .github/workflows/ci.yml?

mkosi/__init__.py
```diff
@@ -3717,13 +3722,19 @@ def calculate_bmap(args: CommandLineArguments, raw: Optional[BinaryIO]) -> Optio
 
 
 def save_cache(args: CommandLineArguments, root: str, raw: Optional[str], cache_path: Optional[str]) -> None:
-    if cache_path is None or raw is None:
+    if cache_path is None:
+        return
+
+    # bail out if cache already exists
+    if os.path.exists(cache_path):
```
behrmann, line 3729:
```suggestion
    if cache_path is None or os.path.exists(cache_path):
        return
```
I think this is clearer than two separate conditionals.

mkosi/__init__.py
```diff
@@ -3717,13 +3722,19 @@ def calculate_bmap(args: CommandLineArguments, raw: Optional[BinaryIO]) -> Optio
...
         return
 
     with complete_step('Installing cache copy ',
                        'Successfully installed cache copy ' + cache_path):
 
         if args.output_format.is_disk_rw():
+            if raw is None:
+                return
```
behrmann, line 3737: I'm not quite convinced that having the `if raw is None` check here communicates the intention very will, because if `raw` is `None` nothing will be done, but the messages will be printed. Maybe you could move this before the `with` statement as `if args.output_format.is_disk_rw() and raw is None`

mkosi/__init__.py
behrmann: This will be fixed by the rebase and running black.

behrmann (commented):
I think this makes sense, although I tend to get a knot in my brain when looking at the output pipeline. I agree this should be added to CI so be on the safe side, that it works and keeps working.

DaanDeMeyer:
Closing this for now, feel free to reopen once you have more time to work on this.

PR #584, opened 2020-12-18
Use black for code formatting

behrmann:
This is a big one and [black](https://github.com/psf/black) is not PEP8-compliant (not like we were before with the line length of 119), but it's being used more and more from what I see. The nice thing is, that it has almost *no* configuration. There is no discussion about style. It comes in any colour you want, as long as it's black. :)

The three commits in this PR: configure black, run black on all relevant files, and add it to our CI.

### DIFF
--- a/.github/workflows/ci-unit-test.yml
+++ b/.github/workflows/ci-unit-test.yml
@@ -15,7 +15,10 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Install
-      run: python3 -m pip install pytest mypy
+      run: python3 -m pip install pytest mypy black
+
+    - name: Check formatting
+      run: python3 -m black --check mkosi/ tests/
 
     - name: Type Checking
       run: python3 -m mypy mkosi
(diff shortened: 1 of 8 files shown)


behrmann:
Ah, the mere existence of pyproject.toml breaks the editable user install.

behrmann:
The latest commit is a bit on the nose. I wanted to see, whether this works as a workaround. I guess I'll need to look into how to get us into the glorious pyproject.toml future.

DaanDeMeyer:
There should be a --no-pep-517 flag we can pass to tell pip not to use pyproject.toml. We can just add that to the instructions for the editable install no? Pep 517 doesn't support editable installs it seems.

DaanDeMeyer:
Should we use [this](https://github.com/psf/black#github-actions) for CI?

behrmann:
> There should be a --no-pep-517 flag we can pass to tell pip not to use pyproject.toml. We can just add that to the instructions for the editable install no? Pep 517 doesn't support editable installs it seems.

Yes, there is `--no-use-pep517`, but I had hoped there was something else than having people need to type that when using editable install, but it seems that's it for now. I added it to the README, hopefully nobody will stumble too hard over this. I'm always hesitant to change the installation workflow.

> Should we use [this](https://github.com/psf/black#github-actions) for CI?

I saw this as well, but I don't think it's really worth it to pull this in and have spin up another test runner, if we can add two lines to ours to do roughly the same.

DaanDeMeyer:
> Yes, there is --no-use-pep517, but I had hoped there was something else than having people need to type that when using editable install, but it seems that's it for now. I added it to the README, hopefully nobody will stumble too hard over this. I'm always hesitant to change the installation workflow.

This is fine, ti's a one time setup anyway. I've executed it once and never looked at it again. (Somewhat related, thank you for pushing so much for pip when I was pushing for zipapp. the editable install turns out to be extremely useful in practice, I can't live without it anymore).

behrmann:
> [..] (Somewhat related, thank you for pushing so much for pip when I was pushing for zipapp. the editable install turns out to be extremely useful in practice, I can't live without it anymore).

Glad this worked out. :)

PR #585, opened 2020-12-19
Add file to pass to git blame --ignore-revs-file containing the black commit

DaanDeMeyer:


behrmann:
TIL, that's a blame option I didn't know existed.

Maybe we should mention `git config blame.ignoreRevsFile` (maybe with with the `--local` switch) in the README as well?

PR #586, opened 2020-12-19
Add --ephemeral option

DaanDeMeyer:
Identical to systemd-nspawn's --ephemeral option (and that's what how
we implement this for mkosi boot) but also implemented for qemu.

Also includes  a small commit to clean up `run_shell`.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -734,6 +734,13 @@ details see the table below.
   in systemd-nspawn's man page:
   https://www.freedesktop.org/software/systemd/man/systemd-nspawn.html#-n
 
+`--ephemeral`
+: When used with the shell, boot or qemu verbs, this option runs the specified
+  verb on a temporary snapshot of the output image that is removed immediately
+  when the container terminates. Taking the temporary snapshot is more efficient
+  on file systems that support subvolume snapshots or 'reflinks' natively ("btrfs"
+  or new "xfs") than on more traditional file systems that do not ("ext4").
+
 ## Command Line Parameters and their Settings File Counterparts
 
 Most command line parameters may also be placed in an `mkosi.default`
@@ -802,6 +809,7 @@ which settings file options.
 | `--extra-search-paths=`           | `[Host]`                | `ExtraSearchPaths=`           |
 | `--qemu-headless`                 | `[Host]`                | `QemuHeadless=`               |
 | `--network-veth`                  | `[Host]`                | `NetworkVeth=`                |
+| `--ephemeral`                     | `[Host]`                | `Ephemeral=`                  |
 
 Command line options that take no argument are not suffixed with a `=`
 in their long version in the table above. In the `mkosi.default` file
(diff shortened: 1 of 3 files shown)


behrmann:
Would maybe a [qemu-snapshot](https://wiki.qemu.org/Documentation/CreateSnapshot#Create_a_snapshot) be an alternative way to implement this?

DaanDeMeyer:
It might be but the documentation is extremely sparse. I think we're better off being consistent here and doing the same as nspawn does internally.

mkosi.md
```diff
@@ -734,6 +734,11 @@ details see the table below.
   in systemd-nspawn's man page:
   https://www.freedesktop.org/software/systemd/man/systemd-nspawn.html#-n
 
+`--ephemeral`
+: When used with the shell, boot or qemu verbs, this option runs the specified
+  verb on a temporary snapshot of the output image that is removed immediately
+  when the container terminates.
```
behrmann, line 740: Maybe add here, that it uses the `--ephemeral` switch for nspawn and a copy of the image for qemu

mkosi/__init__.py
```diff
@@ -5947,11 +5956,19 @@ def run_qemu(args: CommandLineArguments) -> None:
...
+            fname = args.output
+
+        cmdline += ["-drive", f"format={'qcow2' if args.qcow2 else 'raw'},file={fname},if=virtio"]
+        cmdline += args.cmdline
 
-    print_running_cmd(cmdline)
+        print_running_cmd(cmdline)
 
-    run(cmdline, execvp=True)
+        run(cmdline, execvp=True)
```
behrmann, line 5971: I haven't tested this, but does this clean up properly? From the top of my head I don't know whether the ExitStack can be / will be worked through if we exec here.

DaanDeMeyer: Shit, I didn't consider that. Maybe we should get rid of the execvp?

DaanDeMeyer: With a very quick test, this doesn't seem to cause any major problems.

behrmann: That could be done, I guess. Although I don't know whether there was any other intent behind execing here, besides "mkosi is done and doesn't need to stay around".

DaanDeMeyer: It makes signal handling somewhat more awkward as there's two processes running now. I'll need to verify this doesn't break CTRL+C when using the qemu gui.

DaanDeMeyer: This works fine, we already handle the two process scenario in run() correctly. I think we can go ahead and change this. I'll do the same change for mkosi boot

mkosi.md
DaanDeMeyer: I wouldn't go into specifics, I can copy paste some more info from the nspawn man page though

DaanDeMeyer: Let's not add too many implementation details. I extended with some more info from nspawn's man page that informs the user that the copy is more efficient on filesystems that support reflink.

PR #590, opened 2020-12-21
Change --source-file-transfer default to copy-git-others

DaanDeMeyer:
I've gotten bitten multiple times by the fact that by default,
untracked files in git are not copied into the build image. Let's
fix this by changing the --source-file-transfer default to
copy-git-others. This adds any untracked files to the build image
(but still excludes ignored files).

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -4400,7 +4400,7 @@ def create_parser() -> ArgumentParserMkosi:
         default=None,
         help="Method used to copy build sources to the build image."
         + "; ".join([f"'{k}': {v}" for k, v in SourceFileTransfer.doc().items()])
-        + " (default: copy-git-cached if in a git repository, otherwise copy-all)",
+        + " (default: copy-git-others if in a git repository, otherwise copy-all)",
     )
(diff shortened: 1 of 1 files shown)


behrmann:
I guess this makes sense, but I guess it mostly changes to people being used to the current default being bitten instead. :) 

For the next release we definitely need to add a changelog, so that people have a chance to notice this. I guess over / after Christmas I'll compile one.

DaanDeMeyer:
> I guess this makes sense, but I guess it mostly changes to people being used to the current default being bitten instead. :)

In the worst case, there's users that have non-trivial amounts of untracked files in their repo which will notice slowdowns when upgrading to the release that includes this change because all those files are now copied into the build image. To me, this seems like it would be a rare scenario though and it's much more likely any large files that stick around would have been added to the .gitignore already.

On the other hand, it's extremely unintuitive to add a new file to some project, write code, run mkosi and then find out the build fails because mkosi didn't copy the new file into the build image. This seems to me like a much more common occurrence than the above scenario which is why I think changing the default might be a good idea.

PR #591, opened 2020-12-21
Add --source-file-transfer-final option

DaanDeMeyer:
When debugging in the final image using a debugger, the debugging
experience is greatly improved by having the source files available.
--source-file-transfer-final does exactly what --source-file-transfer
does but applies to the final image instead of the build image. While
supporting 'mount' might be possible, we don't do so yet because
the use case is unclear and configuring QEMU to pass through
folders is non-trivial.


behrmann (approved):

DaanDeMeyer:
Updated the summary as well (I keep forgetting that)

PR #592, opened 2020-12-21
Make --autologin work even when a root password is set

DaanDeMeyer:
Currently, --autologin still asks for a password if the root
password is not explicitly deleted. By adding some PAM short circuit
rules for the relevant TTYs, we avoid PAM asking for the root password
when a user logs in on either pts/0 (nspawn) or ttyS0 (QEMU with
--qemu-headless).

Also includes a commit that enables autologin for QEMU boots without
--qemu-headless.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -383,7 +383,7 @@ details see the table below.
 : If running on a SELinux-enabled system (Fedora, CentOS), files inside
   the container are tagged with SELinux context extended attributes
   (`xattrs`), which may interfere with host SELinux rules in building
-  or further container import stages.  
+  or further container import stages.
   This option strips SELinux context attributes from the resulting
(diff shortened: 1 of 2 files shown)


mkosi.md
```diff
@@ -674,7 +674,8 @@ details see the table below.
 
 `--autologin`
 
-: Enable autologin for the `root` user.
+: Enable autologin for the `root` user on pts/0 (nspawn), tty1 (QEMU) and
```
behrmann, line 677: Maybe "enable autologin by patching `/etc/pam.d/login`" to give users a pointer where to look if they want / need to change something about this and can decide whether this is the appropriate option for them.

behrmann (commented):
LGTM

PR #594, opened 2020-12-22
Add --ssh option and ssh verb

### LINKED ISSUE #593: Streamline setting up SSH access to QEMU vm
While --qemu-headless is a perfect solution to work interactively with a vm launched from a user's terminal, the serial connection can't be "multiplexed" to allow usage by other tools. More specifically, it's currently not possible to give VSCode access to a shell in the virtual machine. With SSH access to a remote host, VSCode makes remote debugging possible without having to install any tools on the local host. See https://code.visualstudio.com/docs/cpp/pipe-transport.

The most obvious way to give other tools access to shells in the VM is to setup SSH in the final image. Currently, this requires installing and enabling SSH in the final image and either modifying sshd_config to allow root login with a password or setup an ssh key and copy it to the final image. Logging in with a password is annoying as it requires entering the password every time so the second solution seems more attractive.

This seems like something we could automate in mkosi though. We could add an --ssh option that takes care of installing ssh, enabling sshd and generating an SSH key on the host which is copied to the final image. The password for the SSH key would be the root password provided to mkosi.

This would allow users to specify the --ssh option, boot the QEMU vm and immediately have SSH access to it while only having to enter the root password once if ssh-agent is setup correctly.

DaanDeMeyer:
Fixes #593.
    
Aside from the --ssh option, we also add the "ssh" verb that runs
ssh with all the necessary options.

Also includes a commit to use sets everywhere for handling packages
and a commit to not print a stacktrace if boot/qemu return a  non-zero
exit code.

I might try to extend CI to test this to make sure this works across all
major supported distros. For now, both `boot` and `qemu` have been
tested with Arch.

Basic idea is you build with --ssh and can then use `mkosi ssh` from
the project's root directory to get SSH access to the running container/vm
started with `mkosi boot` or `mkosi qemu`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,6 @@ jobs:
       matrix:
         distro:
           - arch
-          - centos
           - centos_epel
           - debian
           - ubuntu
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,11 +48,9 @@ jobs:
 
     - name: Write shared mkosi config
       run: |
-        sudo systemctl start systemd-networkd
-
         mkdir -p mkosi.default.d
 
-        tee mkosi.default.d/mkosi.default << EOF
+        tee mkosi.default.d/mkosi.default <<- EOF
         [Distribution]
         Distribution=${{ matrix.distro }}
 
@@ -80,7 +78,7 @@ jobs:
           matrix.format != 'plain_squashfs' &&
           matrix.format != 'tar'
       run: |
-        tee mkosi.default << EOF
+        tee mkosi.default <<- EOF
         [Output]
         Bootable=yes
         BootProtocols=uefi
@@ -107,7 +105,7 @@ jobs:
           matrix.distro == 'centos' ||
           matrix.distro == 'fedora')
       run: |
-        tee mkosi.default << EOF
+        tee mkosi.default <<- EOF
         [Output]
         Bootable=yes
         BootProtocols=uefi
@@ -136,7 +134,7 @@ jobs:
           matrix.format != 'plain_squashfs' &&
           matrix.format != 'tar'
       run: |
-        tee mkosi.default << EOF
+        tee mkosi.default <<- EOF
         [Output]
         Bootable=yes
         BootProtocols=bios
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,9 +47,13 @@ jobs:
     - name: Install
       run: sudo python3 -m pip install .
 
-    - name: Build ${{ matrix.distro }}/${{ matrix.format }}
+    - name: Write shared mkosi config
       run: |
-        tee mkosi.default << EOF
+        sudo systemctl start systemd-networkd
+
+        mkdir -p mkosi.default.d
+
+        tee mkosi.default.d/mkosi.default << EOF
         [Distribution]
         Distribution=${{ matrix.distro }}
 
@@ -58,8 +62,14 @@ jobs:
 
         [Validation]
         Password=
+
+        [Host]
+        Ssh=yes
+        NetworkVeth=yes
         EOF
 
+    - name: Build ${{ matrix.distro }}/${{ matrix.format }}
+      run: |
         sudo python3 -m mkosi build
 
     - name: Boot ${{ matrix.distro }}/${{ matrix.format }}
@@ -72,16 +82,9 @@ jobs:
           matrix.format != 'tar'
       run: |
         tee mkosi.default << EOF
-        [Distribution]
-        Distribution=${{ matrix.distro }}
-
         [Output]
-        Format=${{ matrix.format }}
         Bootable=yes
         BootProtocols=uefi
-
-        [Validation]
-        Password=
 
         [Host]
         QemuHeadless=yes
@@ -106,19 +109,12 @@ jobs:
           matrix.distro == 'fedora')
       run: |
         tee mkosi.default << EOF
-        [Distribution]
-        Distribution=${{ matrix.distro }}
-
         [Output]
-        Format=${{ matrix.format }}
         Bootable=yes
         BootProtocols=uefi
         WithUnifiedKernelImages=no
         # Boot with systemd.volatile=overlay so squashfs images can write to the filesystem
         KernelCommandLine=systemd.volatile=overlay
-
-        [Validation]
-        Password=
 
         [Host]
         QemuHeadless=yes
@@ -142,17 +138,10 @@ jobs:
           matrix.format != 'tar'
       run: |
         tee mkosi.default << EOF
-        [Distribution]
-        Distribution=${{ matrix.distro }}
-
         [Output]
-        Format=${{ matrix.format }}
         Bootable=yes
         BootProtocols=bios
         WithUnifiedKernelImages=no
-
-        [Validation]
-        Password=
 
         [Host]
         QemuHeadless=yes
--- a/mkosi.md
+++ b/mkosi.md
@@ -142,6 +142,14 @@ The following command line verbs are known:
   built with `--bootable` (see below). This command must be executed
   as `root` unless the image already exists and `-f` is not specified.
 
+`ssh`
+
+: When the image is built with the `--ssh` option, this command connects
+  to a booted (`boot`, `qemu` verbs) container/VM via SSH. Make sure to
+  run `mkosi ssh` with the same config as `mkosi build` was run with so
+  that it has the necessary information available to connect to the running
+  container/VM via SSH.
+
 `help`
 
 : This verb is equivalent to the `--help` switch documented below: it
@@ -758,6 +766,13 @@ details see the table below.
   on file systems that support subvolume snapshots or 'reflinks' natively ("btrfs"
   or new "xfs") than on more traditional file systems that do not ("ext4").
 
+`--ssh`
+: If specified, installs and enables sshd in the final image and generates
+  a SSH keypair and adds the public key to root's authorized keys in the final
+  image. The private key is stored in mkosi's output directory. When building
+  with this  option and running the image using `mkosi boot` or `mkosi qemu`,
+  the `mkosi ssh` command can be used to connect to the container/VM via SSH.
+
 ## Command Line Parameters and their Settings File Counterparts
 
 Most command line parameters may also be placed in an `mkosi.default`
@@ -830,6 +845,7 @@ which settings file options.
 | `--qemu-headless`                 | `[Host]`                | `QemuHeadless=`               |
 | `--network-veth`                  | `[Host]`                | `NetworkVeth=`                |
 | `--ephemeral`                     | `[Host]`                | `Ephemeral=`                  |
+| `--ssh`                           | `[Host]`                | `Ssh=`                        |
 
 Command line options that take no argument are not suffixed with a `=`
 in their long version in the table above. In the `mkosi.default` file
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -6056,6 +6056,49 @@ def virt_name(args: CommandLineArguments) -> str:
     return cast(str, name[:13])
 
 
+def has_networkd_vm_vt() -> bool:
+    for path in ["/usr/lib/systemd/network", "/lib/systemd/network", "/etc/systemd/network"]:
+        if os.path.exists(os.path.join(path, "80-vm-vt.network")):
+            return True
+
+    return False
+
+
+def ensure_networkd(args: CommandLineArguments) -> None:
+    networkd_is_running = run(["systemctl", "is-active", "--quiet", "systemd-networkd"], check=False).returncode == 0
+    if not networkd_is_running:
+        die("--network-veth requires systemd-networkd to be running (`systemctl enable --now systemd-networkd`)")
+
+    if args.verb == "qemu" and not has_networkd_vm_vt():
+        die(
+            r"""
+            mkosi didn't find 80-vm-vt.network. This is one of systemd's built-in systemd-networkd config
+            files which configures vt-* interfaces. mkosi needs this file in order for --network-veth to work
+            properly for QEMU virtual machines. The file likely cannot be found because the systemd version
+            on the host is too old (< 246) and it isn't included yet.
+
+            As a workaround until the file is shipped by the systemd package of your distro, add a user
+            network file /etc/systemd/network/80-vm-vt.network with the following contents:
+
+            ```
+            [Match]
+            Name=vt-*
+            Driver=tun
+
+            [Network]
+            # Default to using a /28 prefix, giving up to 13 addresses per VM.
+            Address=0.0.0.0/28
+            LinkLocalAddressing=yes
+            DHCPServer=yes
+            IPMasquerade=yes
+            LLDP=yes
+            EmitLLDP=customer-bridge
+            IPv6PrefixDelegation=yes
+            ```
+            """
+        )
+
+
 def run_shell(args: CommandLineArguments) -> None:
     if args.output_format in (OutputFormat.directory, OutputFormat.subvolume):
         target = f"--directory={args.output}"
@@ -6078,6 +6121,7 @@ def run_shell(args: CommandLineArguments) -> None:
         cmdline.append("--volatile=overlay")
 
     if args.network_veth:
+        ensure_networkd(args)
         cmdline.append("--network-veth")
 
     if args.ephemeral:
@@ -6165,6 +6209,7 @@ def run_qemu(args: CommandLineArguments) -> None:
         cmdline += ["-vga", "virtio"]
 
     if args.network_veth:
+        ensure_networkd(args)
         # Use vt- prefix so we can take advantage of systemd-networkd's builtin network file for VMs.
         ifname = f"vt-{virt_name(args)}"
         # vt-<image-name> is the ifname on the host and is automatically picked up by systemd-networkd which
@@ -6220,7 +6265,7 @@ def find_address(args: CommandLineArguments) -> Tuple[str, str]:
             if dst.startswith("fe80"):
                 return dev, dst
 
-        time.sleep(0.2)
+        time.sleep(0.4)
 
     die("Container/VM address not found")
 
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1515,6 +1515,10 @@ def prepare_tree(args: CommandLineArguments, root: str, do_run_build_script: boo
         if args.build_dir is not None:
             os.mkdir(os.path.join(root, "root/build"), 0o755)
 
+    if args.network_veth and not do_run_build_script:
+        os.mkdir(os.path.join(root, "etc/systemd"), 0o755)
+        os.mkdir(os.path.join(root, "etc/systemd/network"), 0o755)
+
 
 def patch_file(filepath: str, line_rewriter: Callable[[str], str]) -> None:
     temp_new_filepath = filepath + ".tmp.new"
@@ -1987,6 +1991,8 @@ def install_fedora(args: CommandLineArguments, root: str, do_run_build_script: b
         configure_dracut(args, root)
     if do_run_build_script:
         packages.update(args.build_packages)
+    if not do_run_build_script and args.network_veth:
+        packages.add("systemd-networkd")
     invoke_dnf(args, root, args.repositories or ["fedora", "updates"], packages, do_run_build_script)
 
     with open(os.path.join(root, "etc/locale.conf"), "w") as f:
@@ -2240,6 +2246,9 @@ def install_centos(args: CommandLineArguments, root: str, do_run_build_script: b
 
     if do_run_build_script:
         packages.update(args.build_packages)
+
+    if not do_run_build_script and args.distribution == Distribution.centos_epel and args.network_veth:
+        packages.add("systemd-networkd")
 
     invoke_dnf_or_yum(args, root, repos, packages, do_run_build_script)
 
@@ -5680,6 +5689,37 @@ def setup_ssh(args: CommandLineArguments, root: str, do_run_build_script: bool, 
     return f
 
 
+def setup_network_veth(args: CommandLineArguments, root: str, do_run_build_script: bool, for_cache: bool) -> None:
+    if do_run_build_script or for_cache or not args.network_veth:
+        return
+
+    network_file = os.path.join(root, "etc/systemd/network/80-mkosi-network-veth.network")
+    with open(network_file, "w") as f:
+        # Adapted from https://github.com/systemd/systemd/blob/v247/network/80-container-host0.network
+        f.write(
+            dedent(
+                """\
+                [Match]
+                Virtualization=!container
+                Type=ether
+
+                [Network]
+                DHCP=yes
+                LinkLocalAddressing=yes
+                LLDP=yes
+                EmitLLDP=customer-bridge
+
+                [DHCP]
+                UseTimezone=yes
+                """
+            )
+        )
+
+    os.chmod(network_file, 0o644)
+
+    run(["systemctl", "--root", root, "enable", "systemd-networkd"])
+
+
 def build_image(
     args: CommandLineArguments, root: str, *, do_run_build_script: bool, for_cache: bool = False, cleanup: bool = False
 ) -> Tuple[Optional[BinaryIO], Optional[BinaryIO], Optional[str], Optional[TextIO]]:
@@ -5759,6 +5799,7 @@ def build_image(
                     set_serial_terminal(args, root, do_run_build_script, for_cache)
                     set_autologin(args, root, do_run_build_script, for_cache)
                     sshkey = setup_ssh(args, root, do_run_build_script, for_cache)
+                    setup_network_veth(args, root, do_run_build_script, for_cache)
                     run_postinst_script(args, root, loopdev, do_run_build_script, for_cache)
 
                 if cleanup:
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -13,6 +13,7 @@ import errno
 import fcntl
 import getpass
 import glob
+import time
 import hashlib
 import os
 import platform
@@ -25,6 +26,7 @@ import string
 import subprocess
 import sys
 import tempfile
+import json
 import urllib.request
 import urllib.parse
 import uuid
@@ -70,9 +72,9 @@ else:
     TempDir = tempfile.TemporaryDirectory
 
 
-MKOSI_COMMANDS_CMDLINE = ("shell", "boot", "qemu")
-MKOSI_COMMANDS_NEED_BUILD = MKOSI_COMMANDS_CMDLINE
-MKOSI_COMMANDS_SUDO = ("build", "clean") + MKOSI_COMMANDS_CMDLINE
+MKOSI_COMMANDS_CMDLINE = ("shell", "boot", "qemu", "ssh")
+MKOSI_COMMANDS_NEED_BUILD = ("shell", "boot", "qemu")
+MKOSI_COMMANDS_SUDO = ("build", "clean", "shell", "boot", "qemu")
 MKOSI_COMMANDS = ("build", "clean", "help", "summary") + MKOSI_COMMANDS_CMDLINE
 
 DRACUT_SYSTEMD_EXTRAS = [
@@ -1497,8 +1499,13 @@ def prepare_tree(args: CommandLineArguments, root: str, do_run_build_script: boo
             cmdline.write(" ".join(args.kernel_command_line))
             cmdline.write("\n")
 
-    if do_run_build_script:
+    if do_run_build_script or args.ssh:
         os.mkdir(os.path.join(root, "root"), 0o750)
+
+    if args.ssh and not do_run_build_script:
+        os.mkdir(os.path.join(root, "root/.ssh"), 0o700)
+
+    if do_run_build_script:
         os.mkdir(os.path.join(root, "root/dest"), 0o755)
 
         if args.include_dir is not None:
@@ -1640,7 +1647,7 @@ def reenable_kernel_install(args: CommandLineArguments, root: str) -> None:
     make_executable(hook_path)
 
 
-def make_rpm_list(args: argparse.Namespace, packages: Set[str]) -> Set[str]:
+def make_rpm_list(args: argparse.Namespace, packages: Set[str], do_run_build_script: bool) -> Set[str]:
     packages = packages.copy()
 
     if args.bootable:
@@ -1662,6 +1669,9 @@ def make_rpm_list(args: argparse.Namespace, packages: Set[str]) -> Set[str]:
                 packages.add("grub2")
             else:
                 packages.add("grub2-pc")
+
+    if not do_run_build_script and args.ssh:
+        packages.add("openssh-server")
 
     return packages
 
@@ -1729,10 +1739,12 @@ def clean_tdnf_metadata(root: str) -> None:
         remove_glob(root + "/var/log/tdnf.*", root + "/var/cache/tdnf")
 
 
-def invoke_dnf(args: CommandLineArguments, root: str, repositories: List[str], packages: Set[str]) -> None:
+def invoke_dnf(
+    args: CommandLineArguments, root: str, repositories: List[str], packages: Set[str], do_run_build_script: bool
+) -> None:
     repos = ["--enablerepo=" + repo for repo in repositories]
     config_file = os.path.join(workspace(root), "dnf.conf")
-    packages = make_rpm_list(args, packages)
+    packages = make_rpm_list(args, packages, do_run_build_script)
 
     cmdline = [
         "dnf",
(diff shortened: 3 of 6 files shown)


DaanDeMeyer:
I also added a requirement to set a root password when using this option so users aren't carelessly setting up root SSH to vms and containers with full root access without a password on the private key. I'm not entirely sure how much we should consider the security aspect here but with ssh we might need to be a bit more careful than previously as this provides a way to access an nspawn container via ssh without needing root privileges. I'm not super familiar with the security aspects but I've read multiple times that it's rather easy to break out of a container which makes me cautious about making it too easy to access a container without root privileges.

behrmann:
> I also added a requirement to set a root password when using this option so users aren't carelessly setting up root SSH to vms and containers with full root access without a password on the private key. I'm not entirely sure how much we should consider the security aspect here but with ssh we might need to be a bit more careful than previously as this provides a way to access an nspawn container via ssh without needing root privileges. I'm not super familiar with the security aspects but I've read multiple times that it's rather easy to break out of a container which makes me cautious about making it too easy to access a container without root privileges.

The main problem comes from root in the container being the same as root outside of it and then getting access to some root-owned resource outside of it. This is something that user namespaces can solve, but I'm not sure there available on every distro by default (looking at Debian here) and for the longest time they had their own security issues, which might still be a problem with ancient kernels (again looking at Debian).

I guess the safest might be to always add a snippet to `/etc/ssh/sshd_config` disabling root access via password (`PermitRootLogin prohibt-password`) or generally password access (`PasswordAuthentication no`), since these changes would install a key anyways. One could also disable all access in sshd_config generally and enable it conditionally via `Match Address <host IP>` for the build host or the network mask of the network nspawn sets up.

DaanDeMeyer:
Prohibiting root login based on password is the default (at least on Arch). While match address would be a nice addition, I was actually worried about attacks coming from the same host where some malicious software executed by the user finds out the IP of the running mkosi container and tries to ssh into it. Without a password on the ssh key, that login would succeed.

The again, with ssh-agent in the mix, it would succeed even if there's a password on the SSH key if the user logged in himself recently enough. I don't know if our threat model should include a compromised build host user account.

behrmann:
>  I don't know if our threat model should include a compromised build host user account.

I don't think it can. If a host is compromised, it's compromised and there's little that can be done at that point. Assuming a host is compromised an attacker could just start throwing their own stuff into the image, during the build, while everything is mounted. I don't think that's something we should worry about.

DaanDeMeyer:
> I don't think it can. If a host is compromised, it's compromised and there's little that can be done at that point. Assuming a host is compromised an attacker could just start throwing their own stuff into the image, during the build, while everything is mounted. I don't think that's something we should worry about.

Good point, maybe we don't need to enforce a password on the SSH key then. As long as we set 0600 permissions on it for the current user, we should be mostly good.

bluca:
> The main problem comes from root in the container being the same as root outside of it and then getting access to some root-owned resource outside of it. This is something that user namespaces can solve, but I'm not sure there available on every distro by default (looking at Debian here) and for the longest time they had their own security issues, which might still be a problem with ancient kernels (again looking at Debian).

Note that user namespaces have been available in Debian, but disable by default, for at least a couple releases including the current stable. From next year's release, 11/Bullseye, they will be enabled by default.

To enable them if disabled by default there's a sysctl knob:

```
sysctl -w kernel.unprivileged_userns_clone=1
```

DaanDeMeyer:
nspawn has an option -U for conditionally using user namespaces if they are enabled. I wonder if we should use that in `mkosi boot`. I know that systemd-nspawn's template service file already uses it. I can't find any mention of it in systemd's release notes so I assume it's been available for quite a while already.

bluca:
Reducing or even eliminating requiring privileges for building/running images would be great. I experimented a bit with partfs ( https://github.com/braincorp/partfs ) as a userspace-substitute of loop devices, which pretty much the main stumbling block when building images - it works most of the times, but unfortunately corrupted reads happened more than once, and I have way too little knowledge of the filesystem layer (and zero of FUSE) to try and fix it.

DaanDeMeyer:
Maybe we can use guestfs (https://libguestfs.org/guestfish.1.html)? That would still leave the problem that nspawn doesn't run without root yet. However, assuming that works, we'd still need to find a way to run nspawn without requiring root privilegees. This would likely require a bit of work in nspawn itself.

.github/workflows/ci.yml
```diff
@@ -41,25 +41,36 @@ jobs:
...
     - name: Install
       run: sudo python3 -m pip install .
 
-    - name: Build ${{ matrix.distro }}/${{ matrix.format }}
+    - name: Write shared mkosi config
       run: |
-        tee mkosi.default << EOF
+        mkdir -p mkosi.default.d
+
+        sudo systemctl enable --now systemd-networkd
```
bluca, line 54: Why is it necessary to enable it on the build host?

behrmann: I guess this is for the `resolvectl query` in the `tests/pexpect/boot.py`.

DaanDeMeyer: `mkosi ssh` works by sshing into the hostname we assigned to the container. Of course, that hostname has to be resolved first to the IP address of the container/vm. To make that work, the container/vm needs to have an IP address. This is why we enable `NetworkVeth` so that we have a Veth/Tap link to the container.

A Veth/Tap link isn't sufficient though to give the container/vm an IP address, we also need a DHCP server to hand out an IP. This is why we enable systemd-networkd. networkd has a default network file that matches TAP devices on the host with a specific prefix and starts a DHCP server on them to hand out IP addresses (see https://github.com/systemd/systemd/blob/master/network/80-vm-vt.network). mkosi makes sure the TAP device is named so that systemd-networkd automatically starts a DHCP server on the host side of the TAP device. 

Once the container/vm has an IP address, we also need that IP address to be associated with the container/vm's hostname. When using systemd-nspawn, this happens automatically as systemd-nspawn registers the container along with its hostname and the ifindex of its veth link with systemd-machined which uses this information to allow resolving the hostname to the link local address of the container via nss-mymachines.

Theoretically, this should be possible with QEMU as well, but QEMU doesn't integrate with systemd-machined yet. I actually started on this but dropped it since I found a quicker solution. That is, once systemd-resolved is running in the container, it advertises the hostname of the container/vm via LLMNR by default. If systemd-resolved is running on the host, it picks this up and the hostname of the container/vm can be resolved to its IP address.

Does that make sense? networkd runs the DHCP server, resolved does LLMNR to make sure we can resolve the hostname of the container/vm on the host.

bluca: That makes sense, although requiring changes on the host is a bit too much IMHO. I am not going to override the default the distro sets for my machine, that just spells trouble. On Debian we have network-manager for desktops and ifupdown for headless installations.

What about the qemu config for usermode emulated networking I suggested here:

https://github.com/systemd/mkosi/issues/593#issuecomment-750867664

With that, you don't need to do anything on the host. In fact I've used that for my lightweight VMs (ie: no libvirt) for ages now.

DaanDeMeyer: networkd should be non-intrusive by default. In other works, it won't try to manage any network interfaces except those it is explicitly told to manage and a few network interfaces that follow some specific naming convention. Of course, if the other network manager you've got running tries to manage all interfaces that appear on the host, there will be trouble as both networkd and the other network manager will be trying to manage the same network interface.

Although really, we shouldn't care too much about what's running on the host, as long as in the end, the hostname of the container/vm is resolvable. We might not even need networkd on the host if LLMNR advertises link local addresses. We just need whatever network manager on the host to pick up LLMNR and make it available via NSS. If LLMNR doesn't work, we can also configure resolved in the vm/container to do MulticastDNS as well.

I can also get back to integrating QEMU with systemd-machined but that does require nss-mymachines to be configured in nssswitch on the host.

Do you know if Debian makes hostnames received via LLMNR/mDNS available via NSS? I know resolved does it but I'm not sure if what Debian uses does it as well.

DaanDeMeyer: Reworked to use port forwarding as that doesn't require as much of the host. I started with the hostname solution because hostnames are easier to remember than port names but since we have `mkosi ssh` regardless, it's not a big problem anymore to use hostnames instead of ports since `mkosi ssh` hides that for us.

bluca: > As for the user mode solution, I'd prefer solutions that work the same across containers/vms and play well with NetworkVeth. 

Up to you of course, but personally I wouldn't mind if the implementation uses different methods. After all, qemu and nspawn are different things. As long as the user interface in mkosi is the same. I don't see an issue there.

I quickly tried your branch and it looks nice! A couple of things:

- for Debian we still need to enable networkd/resolved in the image, and drop in a .network file that matches the interface
- when I tried mkosi ssh, it could connect but then it asked for a passphrase for the ssh key:
```
$ sudo bin/mkosi --default mkosi.files/mkosi.debian ssh
[sudo] password for luca: 
Enter passphrase for key '/home/luca/git/mkosi/mkosi.output/id_rsa':
```

What's the passphrase set to?

behrmann: > What's the passphrase set to?

`args.password` or `""` if that's empty

bluca: Ah right, thanks - that works!

DaanDeMeyer: I'm running blank on a good way to support this for nspawn. Using hostnames requires the host to support LLMNR, Using port forwarding doesn't work with nspawn. Using machinectl shell requires users to enter their password (which wouldn't be required or would be a different password than the on required for SSH with qemu). Fixing nspawn will take ages to become available in all distros. 

Maybe we can hack in our own iptables rule to make port forwarding with nspawn work with localhost?

DaanDeMeyer:
So now we use localhost for qemu and the default route ip for nspawn which seems to work until nspawn can do localhost port forwarding.

DaanDeMeyer:
That doesn't work either on CI it seems. I'll need to look into an iptables rule to fix nspawn's behavior then.

mkosi/__init__.py
```diff
@@ -5588,24 +5658,86 @@ def make_build_dir(args: CommandLineArguments) -> None:
...
 
 
+def setup_ssh(args: CommandLineArguments, root: str, do_run_build_script: bool, for_cache: bool) -> Optional[TextIO]:
+    if do_run_build_script or for_cache or not args.ssh:
+        return None
+
+    if args.distribution in [Distribution.ubuntu, Distribution.debian]:
+        unit = "ssh"
+    else:
+        unit = "sshd"
```
behrmann, line 5668: I'm not sure, this is necessary, at least on the Debian systems I run `sshd.service` is a valid alias to `ssh.service`

.github/workflows/ci.yml
behrmann: I'm not sure it makes sense to bend over backwards. I think it's fine to document under what circumstances this will work and how these can be achieved, i.e. requiring networkd. That should be less surprising than setting iptable rules.

mkosi/__init__.py
DaanDeMeyer: I'll try removing this again but I'm pretty sure I got crashes somewhere without this.

bluca: sshd.service is an alias of ssh.service - there's also ssh.socket for socket-based activation

```
$ systemctl cat ssh.service
# /lib/systemd/system/ssh.service
[Unit]
Description=OpenBSD Secure Shell server
Documentation=man:sshd(8) man:sshd_config(5)
After=network.target auditd.service
ConditionPathExists=!/etc/ssh/sshd_not_to_be_run

[Service]
EnvironmentFile=-/etc/default/ssh
ExecStartPre=/usr/sbin/sshd -t
ExecStart=/usr/sbin/sshd -D $SSHD_OPTS
ExecReload=/usr/sbin/sshd -t
ExecReload=/bin/kill -HUP $MAINPID
KillMode=process
Restart=on-failure
RestartPreventExitStatus=255
Type=notify
RuntimeDirectory=sshd
RuntimeDirectoryMode=0755

[Install]
WantedBy=multi-user.target
Alias=sshd.service
$ systemctl cat ssh.socket
# /lib/systemd/system/ssh.socket
[Unit]
Description=OpenBSD Secure Shell server socket
Before=ssh.service
Conflicts=ssh.service
ConditionPathExists=!/etc/ssh/sshd_not_to_be_run

[Socket]
ListenStream=22
Accept=yes

[Install]
WantedBy=sockets.target
```

mkosi/__init__.py
```diff
@@ -5588,24 +5664,86 @@ def make_build_dir(args: CommandLineArguments) -> None:
...
+
+                [DHCP]
+                UseTimezone=yes
+                """
+            )
+        )
+
+    os.chmod(network_file, 0o644)
+
+    run(["systemctl", "--root", root, "enable", "systemd-networkd"])
```
bluca, line 5724: at least on Debian systems, systemd-resolved also need to be enabled as it's not by default

DaanDeMeyer: We don't need resolved anymore to get --ssh working though with the current approach.

We could enable resolved by default when using --network-veth but let's put that in a separate PR.

DaanDeMeyer:
Alright, one massive hack later, I seem to have gotten this working properly by circumventing the --port option of nspawn and just doing the port forwarding in mkosi itself. I'll do an nspawn PR as well so that in a few years we can hopefully remove this hack from mkosi.

I did have to drop centos from CI as you can only get systemd-networkd on centos when using EPEL I'm fine with just having epel in CI as it should sufficiently test centos as well I think.

mkosi/__init__.py
DaanDeMeyer: I get the following error: `Failed to enable unit, refusing to operate on linked unit file sshd.service`

mkosi/__init__.py
bluca: Yeah I didn't mean it for ssh, but in general for networking. But yes it can be done separately

mkosi/__init__.py
bluca: Yeah, I get the same. I guess you can't enable/disable aliases, only actual units.

mkosi/__init__.py
```diff
@@ -5588,24 +5665,86 @@ def make_build_dir(args: CommandLineArguments) -> None:
...
+    if do_run_build_script or for_cache or not args.network_veth:
+        return
+
+    network_file = os.path.join(root, "etc/systemd/network/80-mkosi-network-veth.network")
+    with open(network_file, "w") as f:
+        f.write(
+            dedent(
+                """\
+                [Match]
+                Virtualization=qemu
```
bluca, line 5708: This setting breaks networkd for me in the Debian build - unless I remove it, ens2 is ignored.

```
root@localhost:~# networkctl status ens2
● 2: ens2                                                           
                      
                     Link File: /usr/lib/systemd/network/99-defau
lt.link                  
                  Network File: n/a                              
                         
                          Type: ether                            
                         
                         State: off (unmanaged)                  
                         
             Alternative Names: enp0s2                           
                         
                          Path: pci-0000:00:02.0                 
                         
                        Driver: e1000                            
                         
                        Vendor: Intel Corporation                
                         
                         Model: 82540EM Gigabit Ethernet Controll
er (QEMU Virtual Machine)
                    HW Address: 52:54:00:12:34:56                
                         
                           MTU: 1500 (min: 46, max: 16110)       
                         
                         QDisc: noop                             
                         
  IPv6 Address Generation Mode: eui64                            
                         
          Queue Length (Tx/Rx): 1/1                              
                         
              Auto negotiation: yes                              
                         
                         Speed: 1Gbps                            
                         
                        Duplex: full                             
                         
                          Port: tp
```

bluca: Ah, it needs to be Virtualization=kvm

```
root@localhost:~# systemd-detect-virt
kvm
```

bluca:
When booting the Debian image with nspawn, ssh fails to started as port 22 is already in use:

```
root@image:~# networkctl status
WARNING: systemd-networkd is not running, output will be incomplete.

●   State: n/a                                                      
  Address: 192.168.1.91 on eth0                                     
           2a01:4b00:f419:6f00:e2db:6a88:4676:d01b on eth0          
           fe80::ed12:aacb:9b6d:5756 on eth0                        
  Gateway: 192.168.1.1 (ASUSTek COMPUTER INC.) on eth0              
           fe80::3a2c:4aff:feca:8278 (ASUSTek COMPUTER INC.) on eth0
root@image:~# systemctl status ssh
● ssh.service - OpenBSD Secure Shell server
     Loaded: loaded (/lib/systemd/system/ssh.service; enabled; vendor preset: enabled)
     Active: failed (Result: exit-code) since Fri 2020-12-25 17:25:30 GMT; 16s ago
       Docs: man:sshd(8)
             man:sshd_config(5)
    Process: 27 ExecStartPre=/usr/sbin/sshd -t (code=exited, status=0/SUCCESS)
    Process: 31 ExecStart=/usr/sbin/sshd -D $SSHD_OPTS (code=exited, status=255/EXCEPTION)
   Main PID: 31 (code=exited, status=255/EXCEPTION)

Dec 25 17:25:30 image systemd[1]: Starting OpenBSD Secure Shell server...
Dec 25 17:25:30 image sshd[31]: error: Bind to port 22 on 0.0.0.0 failed: Address already in u
se.
Dec 25 17:25:30 image sshd[31]: error: Bind to port 22 on :: failed: Address already in use.

Dec 25 17:25:30 image sshd[31]: fatal: Cannot bind any address.
Dec 25 17:25:30 image systemd[1]: ssh.service: Main process exited, code=exited, status=255/EX
CEPTION
Dec 25 17:25:30 image systemd[1]: ssh.service: Failed with result 'exit-code'.
Dec 25 17:25:30 image systemd[1]: Failed to start OpenBSD Secure Shell server.
```

bluca:
NVM my bad, had forgot to pass network-veth to the boot command

bluca:
> NVM my bad, had forgot to pass network-veth to the boot command

Hang on, when booting with --network-veth I get no address?

```
$ sudo bin/mkosi --default mkosi.files/mkosi.debian --ssh=yes --network-veth=yes boot

<...>

root@image:~# networkctl status
● State: no-carrier

Dec 25 17:33:49 image systemd-networkd[22]: Enumeration completed
Dec 25 17:33:49 image systemd-networkd[22]: host0: Link UP
Dec 25 17:33:49 image systemd[1]: Started Network Service.
```

```
$ sudo bin/mkosi --default mkosi.files/mkosi.debian --ssh=yes --network-veth=yes ssh
‣ Error: No address registered for the container with systemd-machined. Use --network-veth to make sure the container has an IP address.
```

DaanDeMeyer:
Can you give the contents of mkosi.debian? I guess you added that one yourself since it's not in the repo.

bluca:
```
[Distribution]
Distribution=debian
Release=unstable

[Output]
Bootable=yes

[Host]
QemuHeadless=yes
```

build command is

```
$ sudo bin/mkosi --default mkosi.files/mkosi.debian --autologin --qemu-headless --force --password=test --hostonly-initrd=yes --ssh=yes --network-veth=yes build
```

DaanDeMeyer:
I cannot reproduce on my machine. I have a hunch the lack of networkd on the host might be the cause. Maybe there's nothing bringing up the host side of the veth link?

I wonder how it's working on CI then though.

bluca:
> I cannot reproduce on my machine. I have a hunch the lack of networkd on the host might be the cause. Maybe there's nothing bringing up the host side of the veth link?
> 
> I wonder how it's working on CI then though.

Strange - but no need to block the feature for this. But the CI seems unhappy, any idea what that error is?

bluca (approved):

DaanDeMeyer:
I was so deep in the iptables weeds yesterday that I didn't think that if we can find the necessary info to setup port forwarding, then we don't even need port forwarding, we can just ssh to the right address directly. I removed all the iptables code and just kept the code to figure out the ip address from machinectl.

mkosi.md
```diff
@@ -142,6 +142,14 @@ The following command line verbs are known:
   built with `--bootable` (see below). This command must be executed
   as `root` unless the image already exists and `-f` is not specified.
 
+`ssh`
+
+: When the image is built with the `--ssh` option, this command connects
+  to a booted (`boot`, `qemu` verbs) container/vm via SSH. Make sure to
```
behrmann, line 148:
```suggestion
  to a booted (`boot`, `qemu` verbs) container/VM via SSH. Make sure to
```

mkosi.md
```diff
@@ -142,6 +142,14 @@ The following command line verbs are known:
...
+  run `mkosi ssh` with the same config as `mkosi build` was run with so
+  that it has the necessary information available to connect to the running
+  container/vm via SSH.
```
behrmann, line 151:
```suggestion
  container/VM via SSH.
```

mkosi.md
```diff
@@ -758,6 +766,17 @@ details see the table below.
   on file systems that support subvolume snapshots or 'reflinks' natively ("btrfs"
   or new "xfs") than on more traditional file systems that do not ("ext4").
 
+`--ssh`
+: If specified, installs and enables sshd in the final image. Also generates
+  a public/private SSH keypair and adds the public key to the SSH authorized keys
```
behrmann, line 771:
```suggestion
: If specified, installs and enables sshd in the final image and generates
  a SSH keypair and adds the public key to root's authorized keys
```

mkosi.md
```diff
@@ -758,6 +766,17 @@ details see the table below.
...
+  in the final image. The private key is stored in mkosi's output directory. When
+  building with this option and running the image using `mkosi boot` or `mkosi qemu`,
+  the `mkosi ssh` command can be used to connect to the container/vm via SSH.
+
+`--ssh-port`
+: Sets the port from the host that is forwarded to port 22 in the container/vm when
```
behrmann, line 777:
```suggestion
: Sets the port from the host that is forwarded to port 22 in the container/VM when
```

mkosi/__init__.py
```diff
@@ -3968,6 +4011,15 @@ def link_output_bmap(args: CommandLineArguments, bmap: Optional[str]) -> None:
         _link_output(args, bmap, args.output_bmap)
 
 
+def link_output_sshkey(args: CommandLineArguments, sshkey: Optional[str]) -> None:
+    if sshkey is None:
+        return
+
+    with complete_step("Linking private ssh key file", "Successfully linked " + args.output_sshkey):
```
behrmann, line 4018: Why not an f-string here?

mkosi/__init__.py
```diff
@@ -4522,6 +4581,16 @@ def create_parser() -> ArgumentParserMkosi:
         help="If specified, the container/VM is run with a temporary snapshot of the output image that is "
         "removed immediately when the container/VM terminates",
     )
+    group.add_argument(
+        "--ssh", action=BooleanAction, help="Setup SSH access from the host to the final image via `mkosi ssh`"
```
behrmann, line 4585: I'm not entirely sure, but I stumble over setup as a verb
```suggestion
        "--ssh", action=BooleanAction, help="Set up SSH access from the host to the final image via `mkosi ssh`"
```

mkosi/__init__.py
```diff
@@ -5588,24 +5665,87 @@ def make_build_dir(args: CommandLineArguments) -> None:
...
+                [Match]
+                Virtualization=|qemu
+                Virtualization=|kvm
+                Type=ether
+
+                [Network]
+                DHCP=yes
+                LinkLocalAddressing=yes
+                LLDP=yes
+                EmitLLDP=customer-bridge
```
behrmann, line 5716: I don't think we need to enable LLDP for the SSH functionality.

mkosi/__init__.py
```diff
@@ -5588,24 +5665,87 @@ def make_build_dir(args: CommandLineArguments) -> None:
...
+
+                [DHCP]
+                UseTimezone=yes
```
behrmann, line 5719: I think I would also leave this to users for configuration.

behrmann (commented):
I haven't yet looked through the whole PR, as Christmas stretches me somewhat thin, but here a couple of comments. The man page will also need regeneration.

mkosi/__init__.py
DaanDeMeyer: This is copied from networkd's file for containers. I'll add a comment. I'd prefer to use the same file for containers and vms.

bluca:
Mmmh something regressed with latest changes - I get 2 interfaces in qemu since -nic is specified twice, but they are both unmanaged:

```
root@localhost:~# networkctl status ens2
● 2: ens2                                                           
                      
                     Link File: /usr/lib/systemd/network/99-defau
lt.link                  
                  Network File: n/a                              
                         
                          Type: ether                            
                         
                         State: off (unmanaged)                  
                         
             Alternative Names: enp0s2                           
                         
                          Path: pci-0000:00:02.0                 
                         
                        Driver: e1000                            
                         
                        Vendor: Intel Corporation                
                         
                         Model: 82540EM Gigabit Ethernet Controll
er (QEMU Virtual Machine)
                    HW Address: 52:54:00:12:34:56                
                         
                           MTU: 1500 (min: 46, max: 16110)       
                         
                         QDisc: noop                             
                         
  IPv6 Address Generation Mode: eui64                            
                         
          Queue Length (Tx/Rx): 1/1                              
                         
              Auto negotiation: yes                              
                         
                         Speed: 1Gbps                            
                         
                        Duplex: full                             
                         
                          Port: tp                               
                         
root@localhost:~# networkctl status ens3
● 3: ens3                                                           
                      
                     Link File: /usr/lib/systemd/network/99-defau
lt.link                  
                  Network File: n/a                              
                         
                          Type: ether                            
                         
                         State: off (unmanaged)                  
                         
             Alternative Names: enp0s3                           
                         
                          Path: pci-0000:00:03.0                 
                         
                        Driver: e1000                            
                         
                        Vendor: Intel Corporation                
                         
                         Model: 82540EM Gigabit Ethernet Controll
er (QEMU Virtual Machine)
                    HW Address: 52:54:00:12:34:57                
                         
                           MTU: 1500 (min: 46, max: 16110)       
                         
                         QDisc: noop                             
                         
  IPv6 Address Generation Mode: eui64                            
                         
          Queue Length (Tx/Rx): 1/1                              
                         
              Auto negotiation: yes                              
                         
                         Speed: 1Gbps                            
                         
                        Duplex: full                             
                         
                          Port: tp
```

Maybe networkd doesn't support the ```|foo``` notation for Virtualization? I have to remove it, and just use ```kvm```, and then it works

DaanDeMeyer:
I switched to "!container", that one should be supported in networkd.

bluca:
> I switched to "!container", that one should be supported in networkd.

It works. Ship it!

DaanDeMeyer:
A little bit more investigation:

- If networkd is not running, the host side of the veth link when booting with nspawn does not get configured. I figure this is what you're running into @bluca. So there's always going to have to be something on the host side to at least bring up the veth link interface. I think we shouldn't enforce anything here but let's at least advertise networkd in the docs.
- We can use IPv6 neighbor discovery to figure out the IP of the container/VM so we don't need to rely on machinectl or QEMU hostfwd. We can just get the IP address via `ip neighbor` and use that as the address for `SSH`. This is nice because it gets us down to one network interface in the QEMU vm and we don't have to parse `machinectl` output which can always change in the future. The problem is that triggering it with `ping` takes multiple seconds the first time so we'd have a little bit of a delay when connecting with ssh for the first time. I opened https://github.com/systemd/systemd/pull/18085 to fix this but of course it will take ages to be deployed everywhere.

DaanDeMeyer:
I implemented the neighbor discovery solution and it turns out it's plenty fast even without the networkd solution.

behrmann:
@DaanDeMeyer would you mind splitting the first two commits out into a separate PR? I think those two can be merged right away and it would make reviewing the rest a tad simpler.

mkosi/__init__.py
```diff
@@ -5584,24 +5655,87 @@ def make_build_dir(args: CommandLineArguments) -> None:
...
+    f: TextIO = cast(
+        TextIO,
+        tempfile.NamedTemporaryFile(
+            mode="w+", prefix=".mkosi-", encoding="utf-8", dir=os.path.dirname(args.output_sshkey)
+        ),
+    )
+
+    with complete_step("Generating SSH keypair"):
+        # Write a 'y' to confirm to overwrite the file.
+        run(["ssh-keygen", "-f", f.name, "-N", args.password or ""], input=f"y\n", text=True)
```
behrmann, line 5678: Why exactly is the input needed here? Just running `ssh-keygen -f ./testkey -N foobar` I'm not asked for input. 

Also, I think it would make sense to add a `-C mkosi` (or similar) so that the comment for the host key isn't `builduser@buildhost`. 

Another minor nit, this leaves open which algorithm is used and for backwards compat it still defaults to RSA. In the spirit of modern defaults, we could choose `-t ed25519` here, which has been an option for years already, so should be supported everywhere.

mkosi/__init__.py
```diff
@@ -5584,24 +5655,87 @@ def make_build_dir(args: CommandLineArguments) -> None:
...
     raw, cached = reuse_cache_image(args, root, do_run_build_script, for_cache)
     if for_cache and cached:
         # Found existing cache image, exiting build_image
-        return None, None, None
+        return None, None, None, None
 
     if not cached:
         raw = create_image(args, root, for_cache)
 
+    sshkey = None
```
behrmann, line 5737: This shouldn't be necessary, since `setup_ssh` will set it to `None`.

DaanDeMeyer:
So I added logic to enable the interface if it wasn't already up in `run_ssh` but that turned out to be super unreliable compared to having networkd manage the interface for some reason. So I was planning on just requiring users to use networkd, except the vm-vt networkd file is only shipped since systemd v246 so it isn't even in Focal.

Would it kill us to write it ourselves to /etc/systemd/network in the host? We could put it behind a prompt asking for explicit permission to write to /etc and only do it if it isn't available yet at /usr/lib/systemd/network/80-vm-vt.network

mkosi/__init__.py
```diff
@@ -6052,6 +6052,50 @@ def virt_name(args: CommandLineArguments) -> str:
     return cast(str, name[:13])
 
 
+def has_networkd_vm_vt() -> bool:
+    networkd_search_paths = run(["systemd-path", "systemd-search-network"], stdout=PIPE, text=True).stdout
+    for path in networkd_search_paths.split(":"):
```
behrmann, line 6057: This may be a stupid nit, but
```suggestion
    networkd_search_paths = run(["systemd-path", "systemd-search-network"], stdout=PIPE, text=True).stdout.split(":")
    for path in networkd_search_paths:
```

mkosi/__init__.py
```diff
@@ -6052,6 +6052,50 @@ def virt_name(args: CommandLineArguments) -> str:
...
+        if os.path.exists(os.path.join(path, "80-vm-vt.network")):
+            return True
+
+    return False
+
+
+def ensure_networkd(args: CommandLineArguments) -> None:
+    networkd_is_running = run(["systemctl", "is-active", "--quiet", "systemd-networkd"], check=False).returncode == 0
+    if not networkd_is_running:
+        die("--network-veth requires systemd-networkd to be running (systemctl enable --now systemd-networkd)")
```
behrmann, line 6067: To bring it in line with other messages
```suggestion
        die("--network-veth requires systemd-networkd to be running (`systemctl enable --now systemd-networkd`)")
```

mkosi/__init__.py
```diff
@@ -6052,6 +6052,50 @@ def virt_name(args: CommandLineArguments) -> str:
...
+    if args.verb == "qemu" and not has_networkd_vm_vt():
+        die(
+            r"""
+            mkosi didn't find 80-vm-vt.network. This is one of systemd's built-in systemd-networkd config
+            files which configures vt-* interfaces. mkosi needs this file in order for --network-veth to work
+            properly for QEMU virtual machines. The file likely cannot be found because the systemd version
+            on the host is too old (< 246) and it isn't included yet.
+
+            As a workaround until the file is shipped by the systemd package of your distro, add a user
+            network file /etc/systemd/network/80-vm-vt.network with the following contents:
```
behrmann, line 6078: This is an excellently helpful error message! :+1:

DaanDeMeyer:
I swear I spend half my time developing software working around outdated software. Life would be so much easier if everyone used Arch :/.

behrmann:
> I swear I spend half my time developing software working around outdated software. Life would be so much easier if everyone used Arch :/.

I sysadmin Debian for a living, half my time is spent working around outdated software and the other half is working around Debianisms :)

DaanDeMeyer:
It's finally green! If I knew the amount of time I'd spend on this before I started I never would have done it.

The 15s sleep in the boot test is kinda annoying though. There's probably a way to make all of this work with netlink so we don't have to do polling but let's leave that for another PR.

mkosi/__init__.py
DaanDeMeyer: We need to overwrite the named temporary file we create just before running ssh-keygen. I did it this way so that we can still take advantage of the named temporary file being automatically removed if it goes out of scope so that we don't leave anything to clean up if we fail somwhere further down the line.

mkosi/__init__.py
```diff
@@ -6056,6 +6056,49 @@ def virt_name(args: CommandLineArguments) -> str:
     return cast(str, name[:13])
 
 
+def has_networkd_vm_vt() -> bool:
+    for path in ["/usr/lib/systemd/network", "/etc/systemd/network"]:
```
bluca, line 6060: On non-usr-merged distros this is in /lib, so both should be checked

DaanDeMeyer:
Moved the vm-vt CI logic to the Action so we don't have to replicate the same logic in systemd's mkosi CI.

mkosi/__init__.py
behrmann: Not that those are really common anymore. debootstrap defaults to merged usr and if my recollection is correct only buildds are currently not installed with merged usr. The crowd opting for non-merged usr might not be the one using mkosi. ;)

bluca: True, but there's also the case of old systems that were never reinstalled and just upgraded from before the default changed, and didn't manually switch. Like my desktop :-)

DaanDeMeyer: We already so some similar stuff when building unified kernel images so let's include the same logic here as well

DaanDeMeyer:
I think this is in a good state now, let's merge and improve further in separate PRs if needed.

PR #598, opened 2020-12-26
Use sets for packages + suppress shell/boot/qemu stacktraces

DaanDeMeyer:
Split out of #594

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1640,28 +1640,28 @@ def reenable_kernel_install(args: CommandLineArguments, root: str) -> None:
     make_executable(hook_path)
 
 
-def make_rpm_list(args: argparse.Namespace, packages: List[str]) -> List[str]:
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -1640,28 +1640,28 @@ def reenable_kernel_install(args: CommandLineArguments, root: str) -> None:
     make_executable(hook_path)
 
 
-def make_rpm_list(args: argparse.Namespace, packages: List[str]) -> List[str]:
-    packages = list(packages)  # make a copy
+def make_rpm_list(args: argparse.Namespace, packages: Set[str]) -> Set[str]:
+    packages = set(packages)  # make a copy
```
behrmann, line 1644:
```suggestion
    packages = packages.copy()
```
I think this would be more explicit.

behrmann (approved):

PR #606, opened 2020-12-27
Arch: Remove network configuration

DaanDeMeyer:
Let's pull Arch in line with the other distros and not do any network
configuration. If needed, we can add this back, but let's do it behind
an option that we implement for all distros so we at least have uniformity
between distros.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,9 @@
 
 ### Highlighted Changes
 
+- Remove Arch Linux network automatic configuration to bring the different
+  distros more in line with each other. To add it back, add a postinstall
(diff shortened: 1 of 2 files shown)


DaanDeMeyer:
(Treat this as an RFC)

I think we should go ahead and remove this bit even if it breaks backwards compat. If we add options for network configuration, we should do a comprehensive job that works for all major distros and that is configurable so that the functionality can be turned off as well.

keszybz:
Yeah, I think we should avoid this kind of complexity. Especially creating a custom network file feels like overreach.

PR #607, opened 2020-12-30
Remove -e switch from calls to sed in release

DaanDeMeyer:


DaanDeMeyer:
I always have to remove the files when doing the release and they don't really provide anything useful since we have version control anyway so let's remove them.

PR #613, opened 2021-01-06
Remove "quiet" from default kernel cmdline

DaanDeMeyer:
We seem to be running into boot issues semi-regularly in CI because
of the usage of systemd.volatile=overlay. These issues are hard to
debug because we silence boot output by default. Let's enable boot
output by default so we have an easier time debugging these issues
when they occur. Also, mkosi boot shows the same output on
container startup anyway so this brings qemu more in line with mkosi
boot as well.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,7 @@
 - Change the --default option to not affect the search location of
   mkosi.default.d anymore. mkosi now always searches for mkosi.default.d
   in the working directory.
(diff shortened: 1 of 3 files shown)


DaanDeMeyer:
Added back rhgb since it only has effect on Fedora/CentOS so let's leave it to a user of those distros to tell us if it's causing issues.

PR #614, opened 2021-01-06
Cache ssh and network veth setup steps

DaanDeMeyer:
These produce output each time they're run so cache the results to
reduce the output clutter when doing an incremental build.


DaanDeMeyer:
Going to go ahead and merge this one as its pretty basic and CI is green except for the kernel overlayfs regression on Arch.

PR #615, opened 2021-01-06
Add --root-size-build option

### LINKED ISSUE #611: Allow different development and final image sizes
Some of my images require building some large packages during the development phase, but only a small subset of those build artefacts are copied to the final image. Making `RootSize` large enough to accommodate the build means the final image is much larger than its contents require. Could another option, maybe something like `FinalRootSize`, which defaults to `RootSize` if not given be added to allow different image sizes for the two steps?

DaanDeMeyer:
Allows independently setting the build image root partition size.
This is useful when the build image is much larger than the final
image due to the extra tools that need to be installed.

Fixed #611.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -665,11 +665,13 @@ def btrfs_forget_stale_devices() -> Generator[None, None, None]:
             run(["btrfs", "device", "scan", "-u"])
 
 
-def image_size(args: CommandLineArguments) -> int:
+def image_size(args: CommandLineArguments, do_run_build_script: bool) -> int:
     size = GPT_HEADER_SIZE + GPT_FOOTER_SIZE
 
-    if args.root_size is not None:
+    if not do_run_build_script and args.root_size is not None:
         size += args.root_size
+    if do_run_build_script and args.root_size_build is not None:
+        size += args.root_size_build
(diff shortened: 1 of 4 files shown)


mkosi/__init__.py
```diff
@@ -3560,7 +3570,7 @@ def insert_generated_root(
     assert image is not None
 
     with complete_step("Inserting generated root partition"):
-        args.root_size = insert_partition(
+        insert_partition(
             args,
```
DaanDeMeyer, line 3574: From what I can see, the assigned value here is never read again so I just removed it. Would be great to have a second look here though, I might have missed something.

behrmann: I don't think it is used afterwards, but I see no reason not to keep it around, since I don't think it's a maintenance burden. Maybe an explanatory comment would be better.

behrmann (commented):
Haven't tested it, but it looks sensible. The CI failures look unrelated, but it needs a rebase against master because of the changes in the news file anyways, so we'll see whether they persist.

DaanDeMeyer:
@bcbnz Could you test this PR out and see if it solves your issue?

DaanDeMeyer:
Let's close this one as there's been no response

PR #618, opened 2021-01-09
Reduce output noise 

DaanDeMeyer:
Some commits that reduce the noise outputted by mkosi during runtime. Changes either remove messages for trivial steps or change steps to only print their step message if they're actually executed.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -584,9 +584,8 @@ def complete_step(text: str, text2: Optional[str] = None) -> Generator[List[Any]
     MkosiPrinter.print_step(text + "...")
     args: List[Any] = []
     yield args
-    if text2 is None:
-        text2 = text + " complete"
-    MkosiPrinter.print_step(text2.format(*args) + ".")
+    if text2 is not None:
+        MkosiPrinter.print_step(text2.format(*args) + ".")
 
 
 @complete_step("Detaching namespace")
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2777,7 +2777,7 @@ def install_opensuse(args: CommandLineArguments, root: str, do_run_build_script:
         patch_file(os.path.join(root, "etc/pam.d/common-auth"), jj)
 
 
-def install_distribution(args: CommandLineArguments, root: str, *, do_run_build_script: bool, cached: bool) -> None:
+def install_distribution(args: CommandLineArguments, root: str, do_run_build_script: bool, cached: bool) -> None:
     if cached:
         return
 
@@ -5809,7 +5809,7 @@ def build_image(
 
                 cached_tree = reuse_cache_tree(args, root, do_run_build_script, for_cache, cached)
                 install_skeleton_trees(args, root, for_cache)
-                install_distribution(args, root, do_run_build_script=do_run_build_script, cached=cached_tree)
+                install_distribution(args, root, do_run_build_script, cached_tree)
                 install_etc_hostname(args, root, cached_tree)
                 install_boot_loader(args, root, loopdev, do_run_build_script, cached_tree)
                 run_prepare_script(args, root, do_run_build_script, cached_tree)
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1318,7 +1318,10 @@ def mount_image(
             umount(root)
 
 
-def install_etc_hostname(args: CommandLineArguments, root: str) -> None:
+def install_etc_hostname(args: CommandLineArguments, root: str, cached: bool) -> None:
+    if cached:
+        return
+
     etc_hostname = os.path.join(root, "etc/hostname")
 
     # Always unlink first, so that we don't get in trouble due to a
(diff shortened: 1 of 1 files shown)


behrmann:
I'm not sure about `completestep` in general. The idea of contextmanagers is ressource allocation and cleanup, which `completestep` doesn't really do. I think the additional indentation for printing a status message doesn't always increase readability. Any thoughts?

DaanDeMeyer:
> I'm not sure about completestep in general. The idea of contextmanagers is ressource allocation and cleanup, which completestep doesn't really do. I think the additional indentation for printing a status message doesn't always increase readability. Any thoughts?

It has its uses, making sure we don't forget to print a message when the step is complete is valuable I think.

I removed the commit that removes trivial messages as people might find those valuable.

mkosi/__init__.py
```diff
@@ -1654,40 +1658,69 @@ def make_rpm_list(args: argparse.Namespace, packages: Set[str], do_run_build_scr
 
 
 def clean_dnf_metadata(root: str) -> None:
-    """Removes dnf metadata iff /bin/dnf is not present in the image
+    """Removes dnf metadata if /bin/dnf is not present in the image
```
behrmann, line 1661: I don't think that's a typo, iff is usually short for "if and only if"

behrmann (commented):
I skimmed through, but this is a big one and I will have to look at the separate commits later, to wrap my head around what all the changes are. I'm not entirely sold and all the extra indentation, as I find it a bit harder to read (when not highlighting indentation), but I guess that might just be my preference.

DaanDeMeyer:
> I skimmed through, but this is a big one and I will have to look at the separate commits later, to wrap my head around what all the changes are. I'm not entirely sold and all the extra indentation, as I find it a bit harder to read (when not highlighting indentation), but I guess that might just be my preference.

Yeah, all the indentation is not ideal. Maybe we should get rid of `complete_step` for most steps since we don't really care about the completed messages most of the time. But let's leave that for another PR.

DaanDeMeyer:
Added one more commit that removes the generic "complete" messages from the output as the end of the previous step is implied by the start message of the next step.

DaanDeMeyer:
@behrmann Have you had a chance to take a look at this one? The diff is big but the changes themselves are pretty basic.

behrmann:
The past week has been pretty busy, but I'll make time this evening.

mkosi/__init__.py
```diff
@@ -1642,34 +1645,63 @@ def clean_dnf_metadata(root: str) -> None:
...
+    ]
+    dnf_path = os.path.join(root, "bin/dnf")
     keep_dnf_data = os.access(dnf_path, os.F_OK, follow_symlinks=False)
 
-    if not keep_dnf_data:
-        MkosiPrinter.print_step("Cleaning dnf metadata...")
-        remove_glob(
-            root + "/var/lib/dnf", root + "/var/log/dnf.*", root + "/var/log/hawkey.*", root + "/var/cache/dnf"
-        )
+    if not keep_dnf_data or all(not os.path.exists(path) for path in dnf_metadata_paths):
```
behrmann, line 1657: The check here is only to bail out and not print the "Cleaning dnf metadata" message, if there is none, right?

It's maybe an unnecessarily picky, but since this repeats in the other functions, i'd factor it out and name it something like `none_of_exists(path: Iterable) -> bool`

mkosi/__init__.py
```diff
@@ -2754,7 +2776,7 @@ def jj(line: str) -> str:
         patch_file(os.path.join(root, "etc/pam.d/common-auth"), jj)
 
 
-def install_distribution(args: CommandLineArguments, root: str, *, do_run_build_script: bool, cached: bool) -> None:
```
behrmann, line 2757: Is there a reason to relax this? The argument lists are often awfully long and I do hate counting, so I do prefer when keyword arguments are forced to be used with their names.

behrmann (commented):
Haven't run it, but looked through it and I don't see anything alarming. 

I stumbled a bit over the mount change, because I couldn't think of reason, from the top of my head. Doesn't this mean, that, when it's necessary to mount, that we mount and unmount twice instead of once? Is this a worthwhile trade-off?

DaanDeMeyer:
> I stumbled a bit over the mount change, because I couldn't think of reason, from the top of my head. Doesn't this mean, that, when it's necessary to mount, that we mount and unmount twice instead of once? Is this a worthwhile trade-off?

If for any reason both the unified kernel image and secure boot steps don't need to run, we currently mount the image for no reason. By passing a lambda to the secure boot and unified kernel image functions, we only mount the image if one of these steps is actually executed. This provides a minor speedup in the incremental build case and removes noise from the output that comes with mounting the image.

mkosi/__init__.py
DaanDeMeyer: Consistency, all the other functions work the same way and are called without keyword arguments in `build_image`. You have a point that using keyword arguments might be better but let's make sure we do it for all the callsites then instead of for only one.

mkosi/__init__.py
DaanDeMeyer: Let's leave this for a subsequent cleanup if it becomes a problem.

mkosi/__init__.py
behrmann: I see, I guess I would have left it in, to change this for the others, but it's fine either way, since there are no external users.

PR #619, opened 2021-01-09
Run ssh-keygen -A during build to speedup initial boot

DaanDeMeyer:
On Arch, when using --ssh, initial boots are bottlenecked by
sshdgenkeys.service which runs ssh-keygen -A. Let's run it during
the build phase to speed up initial boot times.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -5655,6 +5655,10 @@ def setup_ssh(args: CommandLineArguments, root: str, do_run_build_script: bool, 
 
     os.chmod(os.path.join(root, "root/.ssh/authorized_keys"), 0o600)
 
+    with complete_step("Generating SSH host keys"):
+        # Generate SSH host keys already to speed up the initial boot.
+        run_workspace_command(args, root, ["/usr/bin/ssh-keygen", "-A"])
+
     return f
 
 


mkosi/__init__.py
```diff
@@ -5655,6 +5655,10 @@ def setup_ssh(args: CommandLineArguments, root: str, do_run_build_script: bool,
 
     os.chmod(os.path.join(root, "root/.ssh/authorized_keys"), 0o600)
 
+    with complete_step("Generating SSH host keys"):
+        # Generate SSH host keys already to speed up the initial boot.
+        run_workspace_command(args, root, ["/usr/bin/ssh-keygen", "-A"])
```
poettering, line 5660: hmm, wha? then all instances of the image will come up with the same key?

DaanDeMeyer: The `--ssh` option is primarily intended to allows tools easy access to run stuff in a test qemu VM so this didn't seem like a huge problem since we don't assume there's going to be more than one instance of the image running at any given time.

Although I guess if someone decided to use the --ssh option for images that are later distributed and have multiple instances running concurrently this would be a problem.

I guess this opens a can of worms we might not want to deal with, especially since we only save +- 0.5s anyway.

PR #621, opened 2021-01-12
Drop completestep mypy workaround

DaanDeMeyer:
Seems to not be necessary anymore on latest mypy.


behrmann:
Guess I would have just renamed `complete_step` to `completestep` ;) But I'm happy this workaround is gone.

PR #628, opened 2021-01-14
Fix ssh cache

DaanDeMeyer:


DaanDeMeyer:
Caching a step with an output file on the host actually turns out to be non-trivial. Opted to not cache the --ssh setup step instead in #629.

PR #632, opened 2021-01-17
Fixed autologin for centos

MarkWalters-pw:
When installing centos_epel from my centos VPS it still asks for a password even though --autologin was used.  No problem when installing centos under fedora.  Adding auth sufficient pam_rootok.so fixed it.


DaanDeMeyer:
This seems a bit drastic. Can you maybe show the debugging logs for a failed login? After booting a systemd-nspawn container with `mkosi boot`, you can use `machinectl shell` to get a root shell in the image. If you run `journalctl -e` in the machinectl shell after trying to login from the `mkosi boot` login shell, there should be a few lines of pam output showing where the authentication fails. If we have that debugging output, maybe we can do a more precise fix instead of always allowing every root login from anywhere.

Issue #623, opened 2021-01-14
RFE: new verb for generate secureboot keypair


poettering:
I think it would make a ton of sense to have a new verb "mkosi genkey" or so, which generates a suitable keypair and self signed certificate for the secureboot stuff and places it in the right files (i.e. `mkosi.secure-boot.crt` + `mkosi.secureboot.key`), if they don#t exist yet. The openssl commands for that are a bit obscure, hence wrapping that nicely would make sense. The CNs and stuff should be encoded in some new mkosi.defaults settings I figure. At least initially I wouldn't make this too configurable, i.e. just pick some good crypto params and stick to them. If people want to fine-tune the crypto params they can always generate the keys themselves.

`mkosi` could either shell out to openssl for this, or (probably better) use the appropriate Python-native API for generating them?

(it might make to add some extra checking btw, that ensures that `mkosi.secure-boot.key` is not readable by anyone else, and maybe even warn if it is noticed that it is managed in git, which is probably a bad idea given its secret nature)

behrmann:
Good idea. I don't think Python has openssl bindings. That is the purview of [pyopenssl](https://github.com/pyca/pyopenssl) or the [cryptography](https://github.com/pyca/cryptography) package, that pyopenssl seems to recommend for everything non-TLS-related. In going with the current "no runtime dependencies" theme, it might be easiest to stick to shelling out, though.

DaanDeMeyer:
So from what I understand, with the genkey PR merged, we generate the Key Exchange Key and corresponding certificate. We then sign the image with the private part of the key and now need to somehow enroll the public part of the key into secure boot. 

I'm trying to get systemd-boot to do this for us automatically by implementing the approach suggested [here](https://github.com/systemd/mkosi/issues/516#issuecomment-674926756). I went through the UEFI spec and one thing that stands out is that there's both the Key Exchange Key and the Platform Key. The KEK signs individual images, the Platform Key signs KEKs so they can be added to the secure boot signature database.

As far as I can understand, if we want to have zero-configuration setup of secure boot, we'll need a few more extra things in mkosi and systemd-boot:

1. (mkosi) Copy the public part of the generated KEK and the certificate (by genkey) into some predefined location in ESP partition so they can be found by systemd-boot
2. (mkosi) Generate a Platform Key and copy the public part of the key into some predefined location in the ESP partition and use the private part to sign the public part of the KEK before copying it into the ESP as well (so it can be enrolled into the secure boot signature database by systemd-boot)

[...]

Issue #626, opened 2021-01-14
RFE: add simple version concept


poettering:
In light of #625 I'd like to generate images with versioning. For me, versioning means primarily three things:

1. The output files should contain the version somewhere in their name. i.e. instead of generating foo.raw we should generate foo-4.76.5.raw in case version 4.76.5 of the image is built, and that for the whole shebang of generated files (except for SHA256SUMS of course, since that name needs to be fixed)
2. The images themselves should incorporate the version in the /usr/lib/os-release file, and possibly other files. This probably means that mkosi.build should probably know the current version so that it can generate/patch these files.
3. The unified kernel images, need to integrate the number too. sd-boot after all will use the version to order its entries in the menu, and pick the newest.

Here's my implementation idea:

1. Introduce `mkosi.version` that contains the version to generate
2. If this file exists, mkosi reads it and passes it as $VERSION into `mkosi.build`. If the file doesn't exist, everything happens as it always did. (the `mkosi.build` scripts would then patch /usr/lib/os-release with this info)
3. In the Output= stanza we should probably introduce some super basic variable expansion. Initially maybe just say that `%v` when used in Output= would be replaced by the version from `mkosi.version`. Thus if you specify `Output=foobar-%v.raw` you get what is needed.
4. Maybe have a new verb "mkosi bump" or so, which reads mkosi.build, identifies the last numeric component of it, increments it by one, writes it again. It should just do this trivial operation I guess, i.e. not bother with major/minor semantics or so. If people want more complex version bumps they can edit trivially hand.
5. Maybe add an option to mkosi.defaults called `AutoBump=yes`. If set after each *successful* build the equivalent of `mkosi bump` would be run.

With all that in place one could run `mkosi build` and it would generate a new version that would be immediately deployable, including for updates.

(Hmm, maybe changes to the unified kernel generation are probably not needed at all, as long as it picks up any updated /usr/lib/os-release file, which it already does)

DaanDeMeyer:
> If this file exists, mkosi reads it and passes it as $VERSION into mkosi.build. If the file doesn't exist, everything happens as it always did. (the mkosi.build scripts would then patch /usr/lib/os-release with this info)

I don't see why this should be done by mkosi.build. Can't mkosi just do this itself? or is the os-release format distro specific? The rest looks good to me.

poettering:
> I don't see why this should be done by mkosi.build. Can't mkosi just do this itself? or is the os-release format distro specific? The rest looks good to me.

[...]

PR #633, opened 2021-01-18
add command genkey

behrmann:
First draft to address #623.

This leaves out any checking whether the keys are in git or whether they or have wrong permissions, since openssl created them with the right ones.

I'm not entirely sure what a good choice for the CN or the expiration length of the keys should be (ten years might be a tad long), hence this is a draft. I haven't yet generated my own secure boot keys, so I'm also a bit hazy the exact incantations, i.e. if longer keys than 2048 bit are allowed (I remember reading somewhere that they are not) and that to put them into EFI variables, the cert needs to be converted from PEM into DER, so maybe we should do this as well.

Happy about comments.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -4258,7 +4258,7 @@ class ArgumentParserMkosi(argparse.ArgumentParser):
                 continue
             # replace arguments referencing files with the file content
             try:
-                config = configparser.ConfigParser(delimiters="=")
+                config = configparser.RawConfigParser(delimiters="=")
                 config.optionxform = str  # type: ignore
                 with open(arg_string[1:]) as args_file:
                     config.read_file(args_file)
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -8,6 +8,7 @@ import copy
 import crypt
 import ctypes
 import ctypes.util
+import datetime
 import enum
 import errno
 import fcntl
@@ -76,7 +77,7 @@ else:
 MKOSI_COMMANDS_CMDLINE = ("build", "shell", "boot", "qemu", "ssh")
 MKOSI_COMMANDS_NEED_BUILD = ("shell", "boot", "qemu")
 MKOSI_COMMANDS_SUDO = ("build", "clean", "shell", "boot", "qemu")
-MKOSI_COMMANDS = ("build", "clean", "help", "summary") + MKOSI_COMMANDS_CMDLINE
+MKOSI_COMMANDS = ("build", "clean", "help", "summary", "genkey") + MKOSI_COMMANDS_CMDLINE
 
 DRACUT_SYSTEMD_EXTRAS = [
     "/usr/lib/systemd/systemd-veritysetup",
@@ -4258,6 +4259,9 @@ class ArgumentParserMkosi(argparse.ArgumentParser):
                 continue
             # replace arguments referencing files with the file content
             try:
+                # This used to use configparser.ConfigParser before, but
+                # ConfigParser's interpolation clashes with systemd style
+                # specifier, e.g. %u for user, since both use % as a sigil.
                 config = configparser.RawConfigParser(delimiters="=")
                 config.optionxform = str  # type: ignore
                 with open(arg_string[1:]) as args_file:
@@ -4367,6 +4371,18 @@ def create_parser() -> ArgumentParserMkosi:
     )
     group.add_argument("--secure-boot-key", help="UEFI SecureBoot private key in PEM format", metavar="PATH")
     group.add_argument("--secure-boot-certificate", help="UEFI SecureBoot certificate in X509 format", metavar="PATH")
+    group.add_argument(
+        "--secure-boot-valid-days",
+        help="Number of days UEFI SecureBoot keys should be valid when generating keys",
+        metavar="DAYS",
+        default="730",
+    )
+    group.add_argument(
+        "--secure-boot-common-name",
+        help="Template for the UEFI SecureBoot CN when generating keys",
+        metavar="CN",
+        default="mkosi of %u",
+    )
     group.add_argument(
         "--read-only",
         action=BooleanAction,
@@ -6334,6 +6350,57 @@ def run_ssh(args: CommandLineArguments) -> None:
         )
 
 
+def generate_secure_boot_key(args: CommandLineArguments) -> NoReturn:
+    """Generate secure boot keys using openssl"""
+    args.secure_boot_key = args.secure_boot_key or "./mkosi.secure-boot.key"
+    args.secure_boot_certificate = args.secure_boot_certificate or "./mkosi.secure-boot.crt"
+
+    keylength = 2048
+    expiration_date = datetime.date.today() + datetime.timedelta(int(args.secure_boot_valid_days))
+    cn = expand_specifier(args.secure_boot_common_name)
+
+    for f in (args.secure_boot_key, args.secure_boot_certificate):
+        if os.path.exists(f) and not args.force:
+            die(
+                dedent(
+                    f"""\
+                    {f} already exists.
+                    If you are sure you want to generate new secure boot keys
+                    remove {args.secure_boot_key} and {args.secure_boot_certificate} first.
+                    """
+                )
+            )
+
+    MkosiPrinter.print_step(f"Generating secure boot keys rsa:{keylength} for CN `{cn}`.")
+    MkosiPrinter.info(
+        dedent(
+            f"""
+            The keys will expire in {args.secure_boot_valid_days} days ({expiration_date:%A %d. %B %Y}).
+            Remember to roll them over to new ones before then.
+            """
+        )
+    )
+
+    cmd = [
+        "openssl",
+        "req",
+        "-new",
+        "-x509",
+        "-newkey",
+        f"rsa:{keylength}",
+        "-keyout",
+        args.secure_boot_key,
+        "-out",
+        args.secure_boot_certificate,
+        "-days",
+        str(args.secure_boot_valid_days),
+        "-subj",
+        f"/CN={cn}/",
+    ]
+
+    os.execvp(cmd[0], cmd)
+
+
 def expand_paths(paths: List[str]) -> List[str]:
     if not paths:
         return []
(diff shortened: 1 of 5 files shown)


mkosi/__init__.py
```diff
@@ -6273,6 +6273,37 @@ def run_ssh(args: CommandLineArguments) -> None:
...
+            "-new",
+            "-x509",
+            "-newkey",
+            "rsa:2048",
+            "-keyout",
+            args.secure_boot_key or "./mkosi.secure-boot.key",
+            "-out",
+            args.secure_boot_certificate or "./mkosi.secure-boot.crt",
+            "-days",
+            "3650",
```
poettering, line 6300: i'd probably do 2y. i.e. 1y sounds like a good time to roll them over, and let's add a bit of space around that, and doubling sounds like a safe choice for that.

mkosi/__init__.py
```diff
@@ -6273,6 +6273,37 @@ def run_ssh(args: CommandLineArguments) -> None:
...
+            "-subj",
+            "/CN=mkosi/",
```
poettering, line 6302: I'd probably include the local user name. i.e. maybe generate `mkosi of lennart` for a user lennart  or something like that.

I think the validity and the CN are something that might be worth making configurable in new `mkosi.default` keys.

poettering (changes requested):
looks pretty good to me. I think 2K RSA sounds like a good default. Until someone asks for something else I don't think we need to make that configurable. i.e. i'd hardcode this for now until someone complains.

PEM should be the right choice for encoding the key material. If some tools need DER, then we this should be converted on-the-fly, but never be stored in DER persistently ever.

poettering:
maybe while generating them tell user they need to rollover in the configured time, and the CN that was chosen.

mkosi/__init__.py
behrmann: Changed the time to two years and added a note to the user, when the key expires.

mkosi/__init__.py
behrmann: Added a note about the CN, which defaults to `mkosi of user`, where user goes through `SUDO_USER`, `USER` and defaults to `unknown user`.

I've so far decided not to add any configurability.

behrmann:
@DaanDeMeyer Currently our CI always seems to fail for arch `gpt_ext4` and `gpt_xfs` due to `systemd-tmpfiles-setup.service` hanging. Any ideas?

DaanDeMeyer:
https://github.com/systemd/systemd/issues/18140

mkosi/__init__.py
```diff
@@ -6285,6 +6286,51 @@ def run_ssh(args: CommandLineArguments) -> None:
...
+    user = os.getenv("SUDO_USER") or os.getenv("USER") or "unknown user"
+    cn = f"mkosi of {user}"
+
+    for f in (args.secure_boot_key, args.secure_boot_certificate):
+        if os.path.exists(f):
+            die(
+                f"{f} already exists.\n"
+                "If you are sure you want to generate new secure boot keys\n"
+                f"remove {args.secure_boot_key} and {args.secure_boot_certificate} first."
+            )
```
poettering, line 6306: this check should probably be skipped if --force is specified (in which case the key files should just be replaced)

poettering:
So, would still love to see the cn and the days configureable in the mkosi file, and the --force hookup. Otherwise looks excellent to me

behrmann:
This version makes the validity of the keys and the CN configurable. It also adds a very rudimentary `expand_specifier` function, that's not yet hooked up to the config parsing mechanism, so that CN can be `mkosi of %u` and expand to the user. This comes with the downside of having to disable the interpolation that comes with `configparser.ConfigParser` by switching to `configparser.RawConfigParser` instead. This is a backwards-incompatible change and might break somebody who used this configparser feature. It was never advertised, but there's the off chance that somebody did. I hope nobody did, at least I always forget, that it exists.

I also added an unrelated commit on top to satisfy mypy, since the new version released yesterday fixed a long-standing bug that made a cast unnecessary.

behrmann:
I'm not entirely happy with the new secure boot related options in the `[Output]` section, but I guess it's better to keep them close to the other secure boot related options, than start a section solely for them.

mkosi/__init__.py
```diff
@@ -6285,6 +6298,49 @@ def run_ssh(args: CommandLineArguments) -> None:
...
+    keylength = 2048
+    expiration_date = datetime.date.today() + datetime.timedelta(int(args.secure_boot_valid_days))
+    cn = expand_specifier(args.secure_boot_common_name)
+
+    for f in (args.secure_boot_key, args.secure_boot_certificate):
+        if os.path.exists(f) and not args.force:
+            die(
+                f"{f} already exists.\n"
+                "If you are sure you want to generate new secure boot keys\n"
+                f"remove {args.secure_boot_key} and {args.secure_boot_certificate} first."
```
DaanDeMeyer, line 6364: I'd make this a multiline string and use dedent

mkosi/__init__.py
```diff
@@ -6285,6 +6298,49 @@ def run_ssh(args: CommandLineArguments) -> None:
...
+            )
+
+    MkosiPrinter.print_step(f"Generating secure boot keys rsa:{keylength} for CN `{cn}`.")
+    MkosiPrinter.info(
+        f"The keys will expire in {args.secure_boot_valid_days} days ({expiration_date:%A %d. %B %Y}). "
+        "Remember to roll them over to new ones before."
```
DaanDeMeyer, line 6370: before => before then?

mkosi/__init__.py
```diff
@@ -4214,7 +4215,7 @@ def ini_key_to_cli_arg(key: str) -> str:
                 continue
             # replace arguments referencing files with the file content
             try:
-                config = configparser.ConfigParser(delimiters="=")
+                config = configparser.RawConfigParser(delimiters="=")
```
DaanDeMeyer, line 4262: What's the motivation behind this change?

mkosi/__init__.py
```diff
@@ -6285,6 +6298,49 @@ def run_ssh(args: CommandLineArguments) -> None:
         )
 
 
+def generate_secure_boot_key(args: CommandLineArguments) -> NoReturn:
+    """Generate secure boot keys using openssl"""
+    args.secure_boot_key = args.secure_boot_key or "./mkosi.secure-boot.key"
+    args.secure_boot_certificate = args.secure_boot_certificate or "./mkosi.secure-boot.crt"
```
DaanDeMeyer, line 6353: Do these end up in the output directory if it is defined?

mkosi/__init__.py
behrmann: Will do, but as long as it's only a few lines, I tend to prefer implicit concatenation instead of the additional level of indentation.

mkosi/__init__.py
behrmann: The problem is that `ConfigParser` comes with its own interpolation facilities, that unfortunately clash with the format of systemd specifiers, because they too use %, e.g. `%(key)` expands to the value that key has if it was defined before in the config file. I also made a note of this in the commit message changing this.

Since @poettering noted in #626 that it would need some basic variable expansion, I went for `%u` to expand to user, so that specifier knowledge carries over from systemd. 

Unfortunately this change is backwards-incompatible, but I also don't think this is a terribly good feature to have for the mkosi config files as it was. That being said, I hope nobody uses this at the moment.

mkosi/__init__.py
behrmann: Good question. I forgot to test that. The question is: should they? I think they should end up such, that mkosi can pick them up right away without setting `--secure-boot-{key,certificate}`.

mkosi/__init__.py
DaanDeMeyer: I doubt it, LGTM

mkosi/__init__.py
DaanDeMeyer: I think they should, we do it for all the other stuff so I don't see a big reason to deviate from that here.

behrmann: Ok, I'll have a look at it tomorrow.

mkosi/__init__.py
```diff
@@ -4258,7 +4259,10 @@ def ini_key_to_cli_arg(key: str) -> str:
                 continue
             # replace arguments referencing files with the file content
             try:
-                config = configparser.ConfigParser(delimiters="=")
+                # This used to us configparser.ConfigParser before, but
```
DaanDeMeyer, line 4262: us => use?

mkosi/__init__.py
```diff
@@ -6334,6 +6350,57 @@ def run_ssh(args: CommandLineArguments) -> None:
...
+        )
+    )
+
+    cmd = [
+        "openssl",
+        "req",
+        "-new",
+        "-x509",
+        "-newkey",
+        f"rsa:2048",
```
DaanDeMeyer, line 6390: Need to use `keylength` here?

DaanDeMeyer (commented):
One more question, how does this integrate with the existing secureboot stuff? Will the keys generated here automatically be used to sign the image if building with mkosi afterwards?

mkosi/__init__.py
behrmann: Thanks! That's a holdover from an earlier revision.

mkosi/__init__.py
behrmann: So currently this would ignore `args.output_dir`, but to answer the question below, there they wouldn't be picked up automatically, since `find_secure_boot` looks relative to `args.directory`.

I'm still not sure whether supporting whether moving this relative to the output directory makes sense. `--secure-boot-key` and `--secure-boot-certificate` take paths that are not necessarily relative to the output directory. If somebody would set `--output-dir` and a relative path for either of the secure boot options the result could be confusing. 

I'm not sure what the best option is, but I think it would be great if they would be just be picked up for `--secure-boot`, which should usually be the case, if people just run the command at the root of their config. What do you think?

DaanDeMeyer: We can do it this way, thinking further, these keys are not something you'll be regenerating again and again anyway so it kinda makes sense that we just output them in the working directory instead of the output dir.

PR #636, opened 2021-01-21
RFC UEFI boot: make sure efivarfs loaded in initrd

### LINKED ISSUE #562: Debian unstable with HostonlyInitrd enabled fails to boot in QEMU
Error: `[FAILED] Failed to start Switch Root.`

Suspicious logs:

```
Dec 10 18:44:52 localhost systemd-gpt-auto-generator[213]: EFI loader partition unknown, exiting.
Dec 10 18:44:52 localhost systemd-gpt-auto-generator[213]: (The boot loader did not set EFI variable LoaderDevicePartUUID.)
```

Config to reproduce: 

```
[Distribution]
Distribution=debian
Release=unstable

[Output]
Bootable=yes
HostonlyInitrd=yes

[Host]
QemuHeadless=yes
```

Werkov:
The efivarfs is needed in order to GPT root partition discovery work.
Without efivarfs initrd won't be able to switch to the real root.

(Make module inclusion distro specific in case it's included by other
reasons.) -- **this is why the patch is RFC**. I guess #562 it can be problem on any distro when the build host is not UEFI-booted itself (i.e. `efivarfs` won't be loaded with hostonly=1). So I'd be inclined to drop the distro-guard, OTOH, I don't know if there are any distros that have `efivarfs` built in in the kernel (which would break initrd creation).

Fixes #562

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1416,6 +1416,10 @@ def configure_dracut(args: CommandLineArguments, root: str) -> None:
     ):
         with open(os.path.join(dracut_dir, "30-mkosi-uefi-stub.conf"), "w") as f:
             f.write("uefi_stub=/usr/lib/systemd/boot/efi/linuxx64.efi.stub\n")
+
+    # efivarfs must be present in order to GPT root discovery work
+    if args.esp_partno is not None:
+        with open(os.path.join(dracut_dir, "30-mkosi-efivarfs.conf"), "w") as f:
             f.write('add_drivers+=" efivarfs "\n')
 
 


DaanDeMeyer:
LGTM but I'm confused, does this fix anything for you? All it does is limit the distros that we add the efivarfs kernel module for right?

mkosi/__init__.py
```diff
@@ -1416,6 +1416,13 @@ def configure_dracut(args: CommandLineArguments, root: str) -> None:
     ):
         with open(os.path.join(dracut_dir, "30-mkosi-uefi-stub.conf"), "w") as f:
             f.write("uefi_stub=/usr/lib/systemd/boot/efi/linuxx64.efi.stub\n")
+
+    # efivarfs must be present in order to GPT root discovery work
+    if args.esp_partno is not None and args.distribution in (
+        Distribution.debian,
```
behrmann, line 1422: I'm always wary of guards for Debian, but not Ubuntu. Is it correct, that it's not needed for Ubuntu?

DaanDeMeyer: I'm 99% sure the original bug I found with booting only occurred with Debian and not Ubuntu.

Werkov:
The fix -- my development machine/build host is openSUSE TW with _legacy_ boot. As a consequence, the built image doesn't have efivarfs and fails before switch root -- I fixed that.

This issue dominantly boils down to the _building host_, not the particular target distro. The dracut invocation hostonly=1 takes information from the running kernel and that is shared into container (whatever distro it is).

The original fix of #562 was not IMO well designed -- it piggy-backed onto linux-uefi-stub and despite reporting issue with Debian only, it changed the behavior for {ubuntu,mageia,openmadriva} as well.

As described in the initial comment, my only concern is failure in the case if `efivarfs` was builtin module. But I see now that mkosi CI should clarify that -- I'll push the unguarded variant. (FTR, the current arch, gpt_{ext4,xfs} CI failures are timeouts of systemd-tmpfiles-setup.service.)

DaanDeMeyer:
LGTM, I got confused because Github didn't show the extra context. The change makes sense now.

PR #637, opened 2021-01-21
Drop networkd not running to a warning instead of a fatal error

DaanDeMeyer:
Without networkd, the veth link won't come up properly on the host
but that doesn't prevent mkosi boot or mkosi qemu from working so
let's drop those messages to warnings instead of fatal errors.

cc: @Werkov @bluca


bluca (approved):
lgtm - thanks!

behrmann (approved):

PR #638, opened 2021-01-22
Enabling console autologin

MarkWalters-pw:
When installing centos_epel from my centos VPS it still asks for a password even though `--autologin` was used. No problem when installing centos under fedora. Apparently it is using console instead of tty1 or pty/0.  Adding `auth sufficient pam_succeed_if.so tty = console` fixed it.


DaanDeMeyer:
LGTM

PR #640, opened 2021-01-23
Add support for passing arguments to the build command

DaanDeMeyer:
Aside from environment variables, it's also useful to be able to
pass arguments to the build script. For example, in systemd we can
use this to pass the target to build to the build script. Build
scripts determine how the passed arguments are interpreted, mkosi
just passes them on.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -103,7 +103,8 @@ The following command line verbs are known:
 : This builds the image, based on the settings passed in on the
   command line or read from a `mkosi.default` file, see below. This
   verb is the default if no verb is explicitly specified. This command
-  must be executed as `root`.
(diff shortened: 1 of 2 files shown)


DaanDeMeyer:
Added another commit to remove the cast that was made redundant by the new mypy version.

behrmann:
> Added another commit to remove the cast that was made redundant by the new mypy version.

Looks like we have the same commit on two PRs (here and in #633) :) But this PR is short enough and I'll drop it later from mine. 

LGTM, the only downside I see is that it doesn't make the build script self-contained, so that it could rely on the call to mkosi itself, but this isn't anything that couldn't have been done with the script itself, so that should be fine.

behrmann:
This does have more than the currently expected CI failures, though.

PR #641, opened 2021-01-23
Do a recursive chown of the sources if they are mounted into the build image

DaanDeMeyer:
Avoids files owned by root on the host. We exclude all the mkosi to
keep things snappy and to avoid nasty surprises with permissions in
the final image.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -20,6 +20,7 @@ import platform
 import shlex
 import shutil
 import signal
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
Unfortunately, this isn't particularly fast. It works fine for systemd's repo but I also have a mkosi build of every KDE project and the slowdown becomes very noticeable there.

PR #646, opened 2021-02-01
Keep commented locales in /etc/locale.gen

L3n41c:
`/etc/locale.gen` contains the list of supported locales.
In order to enable one, it needs to be uncommented.

The issue with removing all the commented locales and that this PR addresses is the fact that some tools might rely on the commented locales to know what are the valid ones.

For ex., here is the [Ansible module `locale_gen`](https://docs.ansible.com/ansible/latest/collections/community/general/locale_gen_module.html#ansible-collections-community-general-locale-gen-module) [error](https://github.com/ansible-collections/community.general/blob/2.0.0/plugins/modules/system/locale_gen.py#L66) that this PR fixes:
```
TASK [post_install : Enable locale] ******************************************************************************************************************************************************************************************************************************************
fatal: [archlinux]: FAILED! => {"changed": false, "msg": "The locale you've entered is not available on your system."}
```

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2678,8 +2678,20 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
     with mount_api_vfs(args, root):
         run_pacman(packages)
 
-    with open(os.path.join(root, "etc/locale.gen"), "w") as f:
-        f.write("en_US.UTF-8 UTF-8\n")
+    # If /etc/locale.gen exists, uncomment the desired locale and leave the rest of the file untouched.
+    # If it doesn’t exist, just write the desired locale in it.
+    try:
+
+        def _enable_locale(line: str) -> str:
+            if line.startswith("#en_US.UTF-8"):
+                return line.replace("#", "")
+            return line
+
+        patch_file(os.path.join(root, "etc/locale.gen"), _enable_locale)
+
+    except FileNotFoundError:
+        with open(os.path.join(root, "etc/locale.gen"), "x") as f:
+            f.write("en_US.UTF-8 UTF-8\n")
 
     run_workspace_command(args, root, ["/usr/bin/locale-gen"])
 


L3n41c:
Hi @behrmann,

Thanks for the review !

I’ve just rebased on `master`.

Indeed, it seems that `mkosi` sets the locale only for Arch.
On Arch, the `/etc/locale.gen` file shipped in the `glibc` package contains this in the header:
```
#  A list of supported locales is included in this file.
#  Uncomment the ones you need.
```
That’s why I didn’t challenge the Ansible behavior.

According to the following comment in Ansible:
https://github.com/ansible-collections/community.general/blob/2.0.0/plugins/modules/system/locale_gen.py#L63-L67
Ubuntu would be the only distribution where `/etc/locale.gen` isn’t the source of truth for the list of supported locale.

behrmann (approved):

behrmann:
> That’s why I didn’t challenge the Ansible behavior.

I understand that they want to check whether a locale exists, even though I don't think that this is entirely helpful, since it makes the comments in the file syntactically relevant, which they definitely shouldn't be.

mkosi/__init__.py
```diff
@@ -2678,8 +2678,12 @@ def run_pacman(packages: Set[str]) -> None:
     with mount_api_vfs(args, root):
         run_pacman(packages)
 
-    with open(os.path.join(root, "etc/locale.gen"), "w") as f:
-        f.write("en_US.UTF-8 UTF-8\n")
+    def _enable_locale(line: str) -> str:
+        if line.startswith("#en_US.UTF-8"):
+            return line.replace("#", "")
+        return line
+
```
DaanDeMeyer, line 2685: This assumes that /etc/locale.gen exists and that the locale will always be commented in /etc/locale.gen. While I don't think that's going to change anytime soon, let's make this a bit more future proof by only patching /etc/locale.gen if it and appending the locale to the file if it doesn't exist as a commented locale.

L3n41c: @DaanDeMeyer Thanks for the suggestion !
According to your proposition, I’ve just updated the PR to make `mkosi` uncomment the locale in `/etc/locale.gen` if the file exists and to create this file and write the locale in it if it doesn’t exist.

DaanDeMeyer:
LGTM, thanks for the contribution!

PR #647, opened 2021-02-02
Fix bootable gpt_btrfs compression

johnp789:
For bootable gpt_btrfs format output with compression enabled, `mkosi` would crash trying to mount the EFI partition with a compression option.  This change skips adding any compression options if we are mounting the EFI partition.


behrmann:
Thanks for the catch in line 1247! 

Generally, you should also rebase this change against current main and/or drop the spurious merge commits.

johnp789:
> Generally, you should also rebase this change against current main and/or drop the spurious merge commits.

Since I cannot find a way to fix this pull request, I will prepare a new one after learning how to create a proper one.

PR #649, opened 2021-02-03
qemu: Use q35 machine

DaanDeMeyer:
This is necessary for booting with secure boot enabled in QEMU.
Switching to q35 somehow changed the boot order causing us to drop
into the EFI shell at boot. To fix this, I switched from virtio-blk
to virtio-scsci-pci (which is supposed to be the future anyway)
which allows us to set bootindex to override the boot order.


DaanDeMeyer:
Closing as this one is included in the secure boot PR

PR #650, opened 2021-02-03
qemu: Add secure boot support

DaanDeMeyer:
So I'm trying to get this working so I can use it to test secure boot enrollement in sd-boot which I'm working on atm. Just using OVMF_CODE.secboot.fd causes qemu to enter a busy loop on startup. Adding `-global ICH9-LPC.disable_s3=1` causes the following error:

```
daan@archlinux ~/p/systemd (sd-boot-no-include)> sudo mkosi -i qemu
‣ Running command:
‣ qemu-system-x86_64 -machine type=q35,smm=on,accel=kvm -smp 2 -m 1024 -cpu host -drive if=pflash,format=raw,readonly,file=/usr/share/ovmf/x64/OVMF_CODE.secboot.fd -global driver=cfi.pflash01,property=secure,value=on -global ICH9-LPC.disable_s3=1 -object rng-random,filename=/dev/urandom,id=rng0 -device virtio-rng-pci,rng=rng0,id=rng-device0 -nographic -nodefaults -serial mon:stdio -nic tap,script=no,downscript=no,ifname=vt-image,model=virtio-net-pci -drive if=none,id=hd,file=/home/daan/projects/systemd/mkosi.output/image.raw,format=raw -device virtio-scsi-pci,id=scsi -device scsi-hd,drive=hd,bootindex=1

!!!! X64 Exception Type - 06(#UD - Invalid Opcode)  CPU Apic ID - 00000000 !!!!
RIP  - 0000000000030000, CS  - 0000000000000038, RFLAGS - 0000000000010246
RAX  - 0000000000000000, RCX - 0000000000000000, RDX - 0000000000000000
RBX  - FFFFFFFFFFFFFFFF, RSP - 000000003EFB3E38, RBP - 000000003EFB3EA8
RSI  - 000000003E8ED798, RDI - 000000003EFCA378
R8   - 0000000000000000, R9  - 0000000003041001, R10 - 000000000000003A
R11  - 000000003E8E9988, R12 - 000000003E8E9018, R13 - 0000000000000001
R14  - 000000008000F880, R15 - 000000008000F844
DS   - 0000000000000030, ES  - 0000000000000030, FS  - 0000000000000030
GS   - 0000000000000030, SS  - 0000000000000030
CR0  - 0000000080010033, CR2 - 0000000000000000, CR3 - 000000003EC01000
CR4  - 0000000000000668, CR8 - 0000000000000000
DR0  - 0000000000000000, DR1 - 0000000000000000, DR2 - 0000000000000000
DR3  - 0000000000000000, DR6 - 00000000FFFF0FF0, DR7 - 0000000000000400
GDTR - 000000003E9EEA98 0000000000000047, LDTR - 0000000000000000
IDTR - 000000003E5BB018 0000000000000FFF,   TR - 0000000000000000
FXSAVE_STATE - 000000003EFB3A90
!!!! Can't find image information. !!!!
```

I wonder if this is a bug in ovmf itself. Can anyone give this a test run to see if the problem is specific to my machine? I guess it could also be due to the fact that the UEFI bootloader can't find a signed image since secure boot should be enabled but this seems like a super weird error to raise if that's the case.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -810,6 +810,13 @@ def copy_image_temporary(src: str, dir: str) -> BinaryIO:
         return f
 
 
+def copy_file_temporary(src: str, dir: str) -> BinaryIO:
+    with open(src, "rb") as source:
+        f = cast(BinaryIO, tempfile.NamedTemporaryFile(prefix=".mkosi-", dir=dir))
+        copy_file_object(source, f)
+        return f
+
+
 def reuse_cache_image(
     args: CommandLineArguments, root: str, do_run_build_script: bool, for_cache: bool
 ) -> Tuple[Optional[BinaryIO], bool]:
@@ -6189,10 +6196,33 @@ def find_qemu_binary() -> str:
     die("Couldn't find QEMU/KVM binary")
 
 
-def find_qemu_firmware() -> str:
+def find_qemu_firmware() -> Tuple[str, bool]:
     # UEFI firmware blobs are found in a variety of locations,
     # depending on distribution and package.
     FIRMWARE_LOCATIONS = []
+
+    if platform.machine() == "x86_64":
+        FIRMWARE_LOCATIONS.append("/usr/share/ovmf/x64/OVMF_CODE.secboot.fd")
+    elif platform.machine() == "i386":
+        FIRMWARE_LOCATIONS.append("/usr/share/edk2/ovmf-ia32/OVMF_CODE.secboot.fd")
+
+    FIRMWARE_LOCATIONS.append("/usr/share/edk2/ovmf/OVMF_CODE.secboot.fd")
+    FIRMWARE_LOCATIONS.append("/usr/share/qemu/OVMF_CODE.secboot.fd")
+    FIRMWARE_LOCATIONS.append("/usr/share/ovmf/OVMF.secboot.fd")
+
+    for firmware in FIRMWARE_LOCATIONS:
+        if os.path.exists(firmware):
+            return firmware, True
+
+    warn(
+        """\
+        Couldn't find OVMF firmware blob with secure boot support,
+        falling back to OVMF firmware blobs without secure boot support.
+        """
+    )
+
+    FIRMWARE_LOCATIONS = []
+
     # First, we look in paths that contain the architecture –
     # if they exist, they’re almost certainly correct.
     if platform.machine() == "x86_64":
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
Managed to get things working by pointing qemu explicitly at the OVMF_VARS.fd file. Currently, we point to the vars file directly in /usr and add the readonly flag to make sure we don't modify it. If it turns out this prevents UEFI variable writes in the bootloader from succeeding, we'll need to make a copy of the file somewhere else so we can remove the readonly flag.

@mrc0mmand With these changes, I can boot mkosi images using the secure boot ovmf blobs without the actual image being enrolled in secure boot. This PR might remove the blocker from using mkosi for CentOS CI because CentOS 8 only has OVMF_CODE.secboot.fd.

mkosi/__init__.py
```diff
@@ -6174,28 +6174,48 @@ def run_shell(args: CommandLineArguments) -> None:
         run(cmdline, stdout=sys.stdout, stderr=sys.stderr)
 
 
-def run_qemu(args: CommandLineArguments) -> None:
-    # Look for the right qemu command line to use
-    has_kvm = os.path.exists("/dev/kvm")
-    cmdlines: List[List[str]] = []
+def find_qemu_binary() -> str:
     ARCH_BINARIES = {"x86_64": "qemu-system-x86_64", "i386": "qemu-system-i386"}
     arch_binary = ARCH_BINARIES.get(platform.machine(), None)
```
behrmann, line 6179: Not your change, but the `None` here is redundant, since `dict.get` will always return `None` on a missing key.

mrc0mmand:
> Managed to get things working by pointing qemu explicitly at the OVMF_VARS.fd file. Currently, we point to the vars file directly in /usr and add the readonly flag to make sure we don't modify it. If it turns out this prevents UEFI variable writes in the bootloader from succeeding, we'll need to make a copy of the file somewhere else so we can remove the readonly flag.
> 
> Since it seems to be perfectly possible to boot non-secure-boot enrolled images with the secure boot OVMF blob, we use the secure boot firmware by default if it is available.
> 
> @mrc0mmand With these changes, I can boot mkosi images using the secure boot ovmf blobs without the actual image being enrolled in secure boot. This PR might remove the blocker from using mkosi for CentOS CI because CentOS 8 only has OVMF_CODE.secboot.fd.

Excellent, thanks a lot for working on this!

DaanDeMeyer:
Updated to not use readonly for the vars file but make a temporary copy in /var/tmp instead that is removed when the VM exits. If needed we can add an option later to persist the vars file between VM runs.

Also, nothing here is really image specific, it's just about booting qemu with the right options when using the OVMF_CODE.secboot.FD binary blob.

bluca (approved):
Nice! LGTM

bluca:
@DaanDeMeyer looks like this is causing a Debian image to fail to boot in qemu, it times out in the initrd and falls to emergency console:

```
[ TIME ] Timed out waiting for device /dev/gpt-auto-root
```

rdsosreport: https://paste.debian.net/1198880/

It's specifically the switch from virtio-blk to virtio-scsi-pci, changing that fixes the problem

bluca:
Host is debian stable, with qemu 5.2 from backports, image is built from the systemd repo

poettering:
maybe the right kernel module for this is simply not added to the dracut initrd?

bluca:
Added an `f.write('add_drivers+=" virtio_scsi "\n')` to the dracut conf file, but still same issue

PR #651, opened 2021-02-09
Add support for writing authenticated ESLs to /efi/loader/secure-boot

DaanDeMeyer:
In preparation for systemd adding secure-boot enrollment support to
sd-boot, let's have mkosi write authenticated EFI signature lists to
/efi/loader/secure-boot which sd-boot will read to enroll the host
into secure-boot.

cc @poettering 

Not tested as I still need to finish the sd-boot integration so let's wait
until that's finished before merging this.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1603,6 +1603,9 @@ def prepare_tree(args: CommandLineArguments, root: str, do_run_build_script: boo
                 os.mkdir(os.path.join(root, "efi/EFI/systemd"), 0o700)
                 os.mkdir(os.path.join(root, "efi/loader"), 0o700)
 
+                if args.secure_boot:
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
Switched to efitools binaries instead of sbsigntools binaries because the ones from sbsigntools work on DER encoded certificates instead of PEM encoded certificates. If we don't want the efitools dependency, I guess we can add some logic to decode to DER before invoking the sbsigntools binaries.

keszybz:
What's the status here?

DaanDeMeyer:
I've been a bit distracted by my job for a while so I haven't been able to bring up the motivation to finish this. :(.

poettering:
I'd still *love* for this to become reality.

DaanDeMeyer:
Adding blocked since we need support in systemd-boot before this can actually be useful, will also close it for now to avoid cluttering up the PRs page

PR #653, opened 2021-02-14
Make CommandLineArguments fully typed

DaanDeMeyer:
This commit extends CommandLineArguments with all necessary fields
from the argument parser, removes the inheritance from argparse.Namespace
and fixes all resulting typing errors.

We try to stick to typing only changes as much as possible to reduce the
chance of breaking something (although there are a few non-typing changes
where doing so made things easier).

Because we use a dataclass now for the CommandLineArguments class, we
up the required python version to 3.7.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: LGPL-2.1+
 
+import dataclasses
 import argparse
 import collections
 import configparser
@@ -49,7 +50,6 @@ from typing import (
     Set,
     TextIO,
     Tuple,
-    TypeVar,
     Union,
     cast,
     TYPE_CHECKING,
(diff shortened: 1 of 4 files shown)


mkosi/__init__.py
```diff
@@ -5577,7 +5738,7 @@ def print_summary(args: CommandLineArguments) -> None:
 
     MkosiPrinter.info("                 Read-only: " + yes_no(args.read_only))
     detail = " ({})".format(args.compress) if args.compress and not isinstance(args.compress, bool) else ""
-    MkosiPrinter.info("            FS Compression: " + yes_no(args.compress) + detail)
+    MkosiPrinter.info("            FS Compression: " + yes_no(cast(bool, args.compress)) + detail)
```
behrmann, line 5741: Why the cast here, if the signature is changed above to `Optional[bool]`?

behrmann (commented):
Excellent ground work!

mkosi/__init__.py
DaanDeMeyer: Not sure what you mean, `compress` was changed to `Union[str, bool, None]`.

behrmann: Sorry, my bad. Guess I wasn't fully awake yet, when I looked over this.

DaanDeMeyer:
@behrmann Any more comments on this? If not, I'll merge

behrmann:
> @behrmann Any more comments on this? If not, I'll merge

Nope, I think this is good to go and everything else can be changed later.

PR #654, opened 2021-02-16
openSUSE: Fix autologin setup

Werkov:
The distro PAM config resides under /usr/etc and customizations are
supposed to be under /etc. Use the distro file as a template for the
autologin customization.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2803,6 +2803,10 @@ def install_opensuse(args: CommandLineArguments, root: str, do_run_build_script:
 
         patch_file(os.path.join(root, "etc/pam.d/common-auth"), jj)
 
+    if args.autologin:
+        # copy now, patch later (in set_autologin())
+        shutil.copy2(os.path.join(root, "usr/etc/pam.d/login"), os.path.join(root, "etc/pam.d/login"))
+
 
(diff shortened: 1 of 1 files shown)


behrmann:
Looks sensible to me. CI's acting up again or maybe still acting up from when Github had problems a couple of days ago, nothing looks really related.

DaanDeMeyer:
Rerunning the jobs just to be sure. Change looks good to me.

DaanDeMeyer:
Arch still broken unfortunately. This is a bit of a pain, not sure how we're going to fix it.

mrc0mmand:
> Rerunning the jobs just to be sure. Running Arch in nspawn also broke recently because of a glibc update. That's hopely fixed now.

Unfortunately, the Arch job is FUBAR. According to Arch it's the other distros fault they're not up to date, so until glibc/libseccomp on Ubuntu Focal (in this case) gets patched, Arch containers won't work.

In systemd I had to pin the Arch repos to a pre-glibc-2.33-3 snapshot (https://github.com/systemd/systemd/commit/67c972c610de61789d1b6560d28340f70d4b1196), but this is not a long-term fix, since it pins *all* packages, so we lose one of the main benefits of Arch which is bleeding-edge packages.

See: https://bugs.archlinux.org/task/69563

PR #596, opened 2020-12-23
Debian: fix HostonlyInitrd and autologin

bluca:

PR #658, opened 2021-02-20
Arch: Fix autologin on nspawn --boot and qemu

labichn:
This PR fixes auto-login for Arch guests when run with `systemd-nspawn --boot` or by qemu with the same change as in #596.

For an image `mkosi --bootable --distribution=arch --autologin`:

Before:
- `systemd-nspawn -i image.raw` boots to root bash prompt (as always)
- `systemd-nspawn -bi image.raw` boots to a password prompt
- `qemu-system-x86_64 -m 512 -bios /usr/share/ovmf/x64/OVMF.fd -drive format=raw,file=image.raw` boots to password prompt

After: all three boot to root bash prompt.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2918,8 +2918,8 @@ def set_autologin(args: CommandLineArguments, root: str, do_run_build_script: bo
         return
 
     with complete_step("Setting up autologin"):
-        # On Debian, PAM wants the full path to the console device or it will refuse access
-        device_prefix = "/dev/" if args.distribution is Distribution.debian else ""
+        # On Arch, Debian, PAM wants the full path to the console device or it will refuse access
+        device_prefix = "/dev/" if args.distribution in [Distribution.arch, Distribution.debian] else ""
 
         override_dir = os.path.join(root, "etc/systemd/system/console-getty.service.d")
         os.makedirs(override_dir, mode=0o755, exist_ok=True)


DaanDeMeyer:
Can you give more information about your environment? (mkosi version, host system, anything else that might be important)

I'm running Arch and autologin works as expected for me (nspawn and qemu) with Arch guests so there's likely something in the host environment that's causing this I guess.

labichn:
I'm running Arch as well. mkosi was b7ff7cf1f1daebf36dd5c4f4cd741c44c3d328d1 when I branched (just missing today's merges). Glad to provide any other details you think are relevant.
```
[nick@zing mkosi]$ uname -a
Linux zing 5.10.13-arch1-1 #1 SMP PREEMPT Wed, 03 Feb 2021 23:44:07 +0000 x86_64 GNU/Linux
[nick@zing mkosi]$ systemctl --version
systemd 247 (247.3-1-arch)
+PAM +AUDIT -SELINUX -IMA -APPARMOR +SMACK -SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ +LZ4 +ZSTD +SECCOMP +BLKID +ELFUTILS +KMOD +IDN2 -IDN +PCRE2 default-hierarchy=hybrid
```

DaanDeMeyer:
I just tested and this works on my machine as well. I think we can just merge this without looking into the details too much.

PR #659, opened 2021-02-22
Check if btrfs is available before trying to unlink a subvolume

mrc0mmand:
otherwise the output is full of pointless errors on certain systems
(like CentOS 8, which doesn't support btrfs):

```
# mkosi --force --debug run --qemu-headless=true build
‣ Removing output files...
+ btrfs subvol show /home/vagrant/mkosi/mkosi.output/fedora.raw
‣ Error: btrfs not found in PATH.
...
‣ Configuring serial tty (ttyS0)...
‣ Cleaning dnf metadata......
+ btrfs subvol show /var/tmp/mkosi-polqexiq/root/var/log/dnf.log
‣ Error: btrfs not found in PATH.
+ btrfs subvol show /var/tmp/mkosi-polqexiq/root/var/cache/dnf
‣ Error: btrfs not found in PATH.
+ btrfs subvol show /var/tmp/mkosi-polqexiq/root/var/log/dnf.librepo.log
‣ Error: btrfs not found in PATH.
+ btrfs subvol show /var/tmp/mkosi-polqexiq/root/var/log/hawkey.log
‣ Error: btrfs not found in PATH.
+ btrfs subvol show /var/tmp/mkosi-polqexiq/root/var/lib/dnf
‣ Error: btrfs not found in PATH.
+ btrfs subvol show /var/tmp/mkosi-polqexiq/root/var/log/dnf.rpm.log
‣ Error: btrfs not found in PATH.
‣ Cleaning rpm metadata......
+ btrfs subvol show /var/tmp/mkosi-polqexiq/root/var/lib/rpm
‣ Error: btrfs not found in PATH.
‣ Resetting machine ID...
...
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,20 @@ jobs:
         NetworkVeth=yes
         EOF
 
+    # Ubuntu's systemd-nspawn doesn't support faccessat2() syscall, which is
+    # required, since current Arch's glibc implements faccessat() via faccessat2().
+    - name: Update systemd-nspawn
+      if: ${{ matrix.distro == 'arch' }}
+      run: |
+        echo "deb-src http://archive.ubuntu.com/ubuntu/ $(lsb_release -cs) main restricted universe multiverse" | sudo tee -a /etc/apt/sources.list
+        sudo apt update
+        sudo apt build-dep systemd
+        git clone https://github.com/systemd/systemd --depth=1 && cd systemd
+        meson build
+        ninja -C build
+        sudo ln -svf $PWD/build/systemd-nspawn `which systemd-nspawn`
+        systemd-nspawn --version
+
     - name: Build ${{ matrix.distro }}/${{ matrix.format }}
       run: |
         sudo python3 -m mkosi build
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -5074,10 +5074,11 @@ def unlink_try_hard(path: Optional[str]) -> None:
     except:  # NOQA: E722
         pass
 
-    try:
-        btrfs_subvol_delete(path)
-    except:  # NOQA: E722
-        pass
+    if shutil.which("btrfs"):
+        try:
+            btrfs_subvol_delete(path)
+        except:  # NOQA: E722
+            pass
 
     try:
         shutil.rmtree(path)


mrc0mmand:
@DaanDeMeyer the second commit unbreaks the Arch part of the CI by pinning repositories to the last snapshot before the dreaded glibc update. I did a similar thing in systemd (systemd/systemd@67c972c610de61789d1b6560d28340f70d4b1196) so we get rid of the annoying fails and test at least something. Downside is that we lose the main advantage of Arch - bleeding edge software - but for now it's the only solution (known to me) to keep the Arch CI in GH Actions alive.

behrmann:
LGTM.

@mrc0mmand Thanks for the pin, although whacking this one problem seems to have made some others pop up. Also, can you say anything about [Allan MacRae's comment](https://bugs.archlinux.org/task/69563#comment196734), that `glibc-2.33-4`, which has been out of testing for a while now, fixing this? Apparently not, it seems, otherwise this problem would have gone away already, no?

mrc0mmand:
> LGTM.
> 
> @mrc0mmand Thanks for the pin, although whacking this one problem seems to have made some others pop up. Also, can you say anything about [Allan MacRae's comment](https://bugs.archlinux.org/task/69563#comment196734), that `glibc-2.33-4`, which has been out of testing for a while now, fixing this? Apparently not, it seems, otherwise this problem would have gone away already, no?

Unfortunately, it doesn't seem to change anything. I played around with it in GH Actions, including the nspawn's seccomp options, but to no avail. I'll still periodically check it, so we can get rid of the workaround (here as well as in systemd), but for the time being I guess we're stuck with it.

DaanDeMeyer:
Unfortunately, by pinning glibc, we also get an old kernel which means systemd-tmpfiles breaks on overlayfs because of https://github.com/systemd/systemd/issues/18140. I'll look into the plain_squashfs failure later.

behrmann:
> Unfortunately, it doesn't seem to change anything. I played around with it in GH Actions, including the nspawn's seccomp options, but to no avail. I'll still periodically check it, so we can get rid of the workaround (here as well as in systemd), but for the time being I guess we're stuck with it.

The bug's closed and marked as fixed, but considering this, might it make sense to ask for the bug to be reopened? I don't feel qualified to do that, since I've only witnessed the fallout, but you have investigated this further so you are in a good position to do that.

mrc0mmand:
> Unfortunately, by pinning glibc, we also get an old kernel which means systemd-tmpfiles breaks on overlayfs because of [systemd/systemd#18140](https://github.com/systemd/systemd/issues/18140). I'll look into the plain_squashfs failure later.

Ah, darn, that's unfortunate...

mrc0mmand:
@behrmann @DaanDeMeyer soo, I did some more testing and it looks like the latest Arch's glibc in combination with newer `systemd-nspawn` (which supports `faccessat2()`) seems to work. I guess this workaround is a much better variant to the previous repo pinning, since we can still *enjoy* all the benefits of bleeding-edge packages.

behrmann:
@mrc0mmand This is excellent news. Good job! :)

DaanDeMeyer:
Yes, this is amazing work! Great to see CI all green again.

PR #660, opened 2021-02-25
Allow overriding # of CPUs and amount of RAM for qemu guests

mrc0mmand:
This commit introduces two options - `--qemu-smp` and `--qemu-mem` - which
can be used to override the default number of CPUs and amount of RAM for
guests started via the `qemu` verb.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -772,6 +772,16 @@ details see the table below.
   connects to its serial console from the current terminal instead
   of launching the VM in a separate window.
 
+`--qemu-smp=`
+
+: When used with the qemu verb, this options sets the qemu's `-smp`
+  argument which controls the number of guest's CPUs. Defaults to `2`.
+
+`--qemu-mem=`
+
+: When used with the qemu verb, this options sets the qemu's `-m`
+  argument which controls the amount of guest's RAM. Defaults to `1G`.
+
 `--network-veth`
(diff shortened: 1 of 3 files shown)


behrmann:
LGTM, but needs reformatting with black and the new options need to be added to the tests.

behrmann:
I somewhat rue, that I argued against having a qemu section in the config, because at the time it would have had only a single option.

behrmann (approved):

mrc0mmand:
Thanks for the quick review!



> I somewhat rue, that I argued against having a qemu section in the config, because at the time it would have had only a single option.

Hopefully, the list of qemu options won't grow much further, since the rest of the stuff _should_ be configurable via the positional `cmdline` argument.

DaanDeMeyer:
I think the sections are a bit of a lost cause regardless since they're full of what I would consider unrelated options so this doesn't matter too much.

DaanDeMeyer:
LGTM @mrc0mmand Are you playing around with mkosi for CentOS CI by any chance?

mrc0mmand:
> LGTM @mrc0mmand Are you playing around with mkosi for CentOS CI by any chance?

Yes! However, it goes slowly so far, due to other tasks. Nevertheless, the secure boot stuff you've recently implemented works like a charm, so even with the SB-only ovmf the VMs generated by mkosi finally work on CentOS 8.

PR #661, opened 2021-02-26
OpenMandriva: support more architectures

tpgxyz:
This allow to build image on other than x86_64 architectures for OpenMandriva

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2158,6 +2158,7 @@ def install_mageia(args: CommandLineArguments, root: str, do_run_build_script: b
 @complete_step("Installing OpenMandriva")
 def install_openmandriva(args: CommandLineArguments, root: str, do_run_build_script: bool) -> None:
     release = args.release.strip("'")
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -2191,6 +2192,8 @@ def install_openmandriva(args: CommandLineArguments, root: str, do_run_build_scr
     if not do_run_build_script and args.bootable:
         packages |= {"kernel-release-server", "binutils", "systemd-boot", "dracut", "timezone", "systemd-cryptsetup"}
         configure_dracut(args, root)
+    if args.network_veth:
+        packages |= {"systemd-networkd"}
```
behrmann, line 2196: Could you split this into a separate commit or at least update the commit message, since this is unrelated to the architectures?

behrmann (commented):
LGTM

mkosi/__init__.py
tpgxyz: Well i forgot to mention in commit about this

PR #667, opened 2021-03-02
mkosi: explicitly flush SHA256SUMS file

poettering:
Apparently we have to flush things explicitly after writing it,
otherwise things remain buffered by python IO and the SHA256SUMS file
will remain empty on disk.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -4037,6 +4037,8 @@ def calculate_sha256sum(
             assert args.output_nspawn_settings is not None
             hash_file(f, nspawn_settings, os.path.basename(args.output_nspawn_settings))
 
+        f.flush()
+
     return f
 
 


DaanDeMeyer (approved):

behrmann:
Huh, the file should be flushed on close. Do we leak the file here?

poettering:
We copy it in before we close it I think. which is probably OK – as long as the data really hit the disk by then

DaanDeMeyer:
I wonder if we're not hitting the same issue with other files if it's a problem here. Then again, I don't mind squashing these as we discover them. If it's a more general problem, let's fix it properly when it occurs

behrmann:
> I wonder if we're not hitting the same issue with other files if it's a problem here. Then again, I don't mind squashing these as we discover them. If it's a more general problem, let's fix it properly when it occurs

Looking at the code we should definitely hit this at least with bmap as well.

poettering:
i gave this a quick look, but i think we always call external tools on those other files, and don't write them ourselves in text mode. But dunno, I don't know Python that well i guess..

behrmann:
I think ssh should be fine, since there `ssh-keygen` handles the file, we only pass a name, but for bmap it's stdout being redirected through python's subprocess module to the `TextIO` variable, that might hang in buffering. I don't think it would hurt to flush there as well, for more info I'd need to look into the implementation.

DaanDeMeyer:
CI failure is unrelated. Let's do any other fixes in a separate PR.

Issue #674, opened 2021-03-10
RFE: Include `systemd-volatile-root` (generator and service) in `initrd`


NicolasT:
Currently `/usr/lib/systemd/systemd-volatile-root` and `/usr/lib/systemd/system/systemd-volatile-root.service` are not included in the generated `initramfs`. As a result, booting a readonly `gpt_*` image (e.g., `gpt_ext4` created with `--readonly`, or a `gpt_squashfs` image) with `systemd.volatile=yes` added to the kernel command-line fails.

Adding these two files to `DRACUT_SYSYTEMD_EXTRAS` fixes this, at least for CentOS 8 images.

DaanDeMeyer:
This is fine, feel free to make a PR

PR #675, opened 2021-03-11
Include `systemd.volatile` infrastructure in initrd

NicolasT:
For the `systemd.volatile=...` feature to work, the
`systemd-volatile-root` generator and `systemd-volatile-root.service`
service are required to be present. Booting a read-only image generated
by `mkosi` (e.g., `gpt_ext4` built with the `--read-only` flag, or a
`gpt_squashfs` image) with a volatile but mutable overlay is useful, so
we need to provide these files in the `initrd` Dracut generates.

Hence, adding both files to `DRACUT_SYSTEMD_EXTRAS`.

Fixes: #674
See: https://github.com/systemd/mkosi/issues/674


behrmann:
LGTM

Issue #634, opened 2021-01-19
RFE: support building images with only /usr/ hierarchy, but no root dir


poettering:
I'd like to see a mode supported where mkosi prepares things as it usually does, but instead of putting the whole rootfs into the image actually just puts the /usr part of it in the image, and dropping everything else.

usecase: in systemd we nowadays support setting up /etc and /var and everything else to the minimum necessary if they are not populated. Thus, an image like this could either:

1. mount a tmpfs as root, so that every boot comes up with a freshly initialized /etc and /var (kernel cmdline option `root=tmpfs usr=/dev/by-…`
2. on first boot, from the initrd generate a root partition, for example via a simple `systemd-repart` drop-in, so that that a basic root fs is freshly initialized on first boot, but persists from that point on. (users could then do factory reset, by removing the root partition again, which `systemd-repart` already implements, too)

(`root=tmpfs` is not yet supported by systemd, but i do intend to change that soon)

poettering:
A bit more background on my usecase for tall of this: i eventually want to come to a state where I can use `mkosi` to build images that consist of just:

1. an ESP
2. a read-only /usr/ file system partition
3. a verity partition for /usr/

And then at first boot I want to use `systemd-repart` to add in:
1. A second version of the /usr/ partition, so that A/B updates can work
2. Similar, a second version of the verity partition for /usr.
3. An encrypted root partition, with keys in TPM
4. A /home partition unencrypted, but with dm-integrity based on an HMAC, with keys in TPM (so that the file system is effectively read-only outside of the local system. This partition should then contain `systemd-homed` encrypted home volumes. This scheme avoids double encryption (as `systemd-homed` encrypts anyway), and people can sitll extract their encrypted volumes easily if the system breaks.
5. An encrypted swap partition, with keys in TPM

DaanDeMeyer:
Any thoughts on how to handle non-system users in this scheme? Where do the systemd-homed volumes come from? I really want to setup a little Linux home router based on the scheme you described here eventually where every update comes from running mkosi which is then deployed via casync. I assume we'd add an option to mkosi to initialize regular users as well (optionally with systemd-homed encryption)?

poettering:
[...]

Issue #625, opened 2021-01-14
RFE: way to also output generated partitions + kernels separately, too


poettering:
Right now mkosi builds a combined disk image containing multiple partitions + the signed kernels.

I am currently toying with some very simple tools for upadating images sanely after they have been booted. Idea is to support an A/B scheme, i.e. something along these lines:

1. Build minimal mkosi images, consisting of just one root partition + one verity partition + one EFI partition
2. At first boot, use `systemd-repart` to add in another (initially empty) root partition of the same size, plus another verity partition plus encrypted/TPM'ed/… /var and /home partitions.
3. At update download just the new root partition + verity partition + signed kernel, stream them into the empty partitions and the ESP + reboot, done.

What's missing for this is that mkosi spits out not only the combined partitioned image, but also the root partition + verity partition as separate files, as well as the signed kernel it puts into the ESP.

With that in place fresh installs would use the full disk image, but updates would download the three files separately, and put them in the right places, and reboot.

(Implementation would probably easy to implement (I mean it just means to dd/cp out the raw partitions/kernels into new files). Question though is whether we should maybe swap the whole order of partitioning around in mkosi. i.e. instead of setting up the partition table first, and mount the file systems directly from it, operate the other way: generate all file systems as separate files, and as last step combine them all in a GPT partition table.)

behrmann:
As a short note: This scheme can currently not be implemented for Debian (and I guess Ubuntu), since Debian doesn't build `systemd-repartd` and likely won't ship it before the [bookworm](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=976959) release, maybe ~2023.

DaanDeMeyer:
A separate kernel would also allow us to provide a qemu fast path using the `-kernel` option. That could shave another 2 seconds of qemu boot times.

PR #676, opened 2021-03-11
mkosi: add UsrOnly= setting for generating images with only a /usr/ p…

poettering:
…artition

This adds a new boolean option that changes our image layout slightly:
instead of include the whole root fs in the image, we just pack up the
/usr/ subtree.

This has two usecases:

1. Truly stateless systems that come up pristine on every single boot,
   where /etc and /var are populated by tmpfiles/sysusers and related
   calls

2. Systems that are shipped without root fs, but where systemd-repart
   adds one in, with a locally generated encryption key and so on. (this
   doesn't work fully yet, because of some initrd issues, will look at
   this later.)

A companion PR on the systemd side is: https://github.com/systemd/systemd/pull/18958

Fixes: #634

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -394,6 +394,7 @@ class CommandLineArguments:
     srv_size: Optional[int]
     var_size: Optional[int]
     tmp_size: Optional[int]
+    usr_only: bool
     checksum: bool
     sign: bool
     key: Optional[str]
@@ -453,6 +454,11 @@ GPT_ROOT_X86_64        = uuid.UUID("4f68bce3e8cd4db196e7fbcaf984b709")  # NOQA: 
 GPT_ROOT_ARM           = uuid.UUID("69dad7102ce44e3cb16c21a1d49abed3")  # NOQA: E221
 GPT_ROOT_ARM_64        = uuid.UUID("b921b0451df041c3af444c6f280d3fae")  # NOQA: E221
 GPT_ROOT_IA64          = uuid.UUID("993d8d3df80e4225855a9daf8ed7ea97")  # NOQA: E221
+GPT_USR_X86            = uuid.UUID("75250d768cc6458ebd66bd47cc81a812")  # NOQA: E221
+GPT_USR_X86_64         = uuid.UUID("8484680c952148c69c11b0720656f69e")  # NOQA: E221
+GPT_USR_ARM            = uuid.UUID("7d0359a302b34f0a865c654403e70625")  # NOQA: E221
+GPT_USR_ARM_64         = uuid.UUID("b0e01050ee5f4390949a9101b17104e9")  # NOQA: E221
+GPT_USR_IA64           = uuid.UUID("4301d2a64e3b4b2abb949e0b2c4225ea")  # NOQA: E221
 GPT_ESP                = uuid.UUID("c12a7328f81f11d2ba4b00a0c93ec93b")  # NOQA: E221
 GPT_BIOS               = uuid.UUID("2168614864496e6f744e656564454649")  # NOQA: E221
 GPT_SWAP               = uuid.UUID("0657fd6da4ab43c484e50933c84b4f4f")  # NOQA: E221
@@ -464,6 +470,11 @@ GPT_ROOT_X86_64_VERITY = uuid.UUID("2c7357edebd246d9aec123d437ec2bf5")  # NOQA: 
 GPT_ROOT_ARM_VERITY    = uuid.UUID("7386cdf2203c47a9a498f2ecce45a2d6")  # NOQA: E221
 GPT_ROOT_ARM_64_VERITY = uuid.UUID("df3300ced69f4c92978c9bfb0f38d820")  # NOQA: E221
 GPT_ROOT_IA64_VERITY   = uuid.UUID("86ed10d5b60745bb8957d350f23d0571")  # NOQA: E221
+GPT_USR_X86_VERITY     = uuid.UUID("8f461b0d14ee4e819aa9049b6fb97abd")  # NOQA: E221
+GPT_USR_X86_64_VERITY  = uuid.UUID("77ff5f63e7b64633acf41565b864c0e6")  # NOQA: E221
+GPT_USR_ARM_VERITY     = uuid.UUID("c215d7517bcd4649be906627490a4c05")  # NOQA: E221
+GPT_USR_ARM_64_VERITY  = uuid.UUID("6e11a4e7fbca4dedb9e9e1a512bb664e")  # NOQA: E221
+GPT_USR_IA64_VERITY    = uuid.UUID("6a491e033be745458e3883320e0ea880")  # NOQA: E221
 GPT_TMP                = uuid.UUID("7ec6f5573bc54acab29316ef5df639d1")  # NOQA: E221
 GPT_VAR                = uuid.UUID("4d21b016b53445c2a9fb5c16e091fd2d")  # NOQA: E221
 # fmt: on
@@ -515,7 +526,7 @@ class GPTRootTypePair(NamedTuple):
     verity: uuid.UUID
 
 
-def gpt_root_native(arch: Optional[str]) -> GPTRootTypePair:
+def gpt_root_native(arch: Optional[str], usr_only: bool = False) -> GPTRootTypePair:
     """The tag for the native GPT root partition for the given architecture
 
     Returns a tuple of two tags: for the root partition and for the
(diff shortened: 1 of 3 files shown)


behrmann:
Looks good on first, look. I expect the unit tests to fail, because the new option needs to be added to the tests.

mkosi/__init__.py
```diff
@@ -4757,6 +4790,9 @@ def create_parser() -> ArgumentParserMkosi:
     group.add_argument(
         "--tmp-size", help="Set size of /var/tmp partition (only gpt_ext4, gpt_xfs, gpt_squashfs)", metavar="BYTES"
     )
+    group.add_argument(
+        "--usr-only", action=BooleanAction, help="Generate only a /usr/ partition, but not root partition"
```
DaanDeMeyer, line 4794: "Generate a /usr/ partition instead of a root partition" ?

mkosi/__init__.py
```diff
@@ -3585,26 +3607,37 @@ def make_minimal_btrfs(args: CommandLineArguments, root: str, for_cache: bool) -
...
     return f
 
 
 def make_generated_root(args: CommandLineArguments, root: str, for_cache: bool) -> Optional[BinaryIO]:
 
+    if not is_generated_root(args):
+        return None
+
+    label = "usr" if args.usr_only else "root"
+    patched_root = (root + "/usr") if args.usr_only else root
```
DaanDeMeyer, line 3633: Let's use a f-string here: `f"{root}/usr}"`

mkosi/__init__.py
```diff
@@ -3585,26 +3607,37 @@ def make_minimal_btrfs(args: CommandLineArguments, root: str, for_cache: bool) -
...
         f.truncate(args.root_size)
 
-        command = ["mkfs.btrfs", "-L", "root", "-d", "single", "-m", "single", "--shrink", "--rootdir", root, f.name]
-        try:
-            run(command)
-        except subprocess.CalledProcessError:
-            # The --shrink option was added in btrfs-tools 4.14.1, before that it was the default behaviour.
-            # If the above fails, let's see if things work if we drop it
-            command.remove("--shrink")
+        command = ["mkfs.btrfs", "-L", label, "-d", "single", "-m", "single", "--rootdir", root, f.name]
```
DaanDeMeyer, line 3610: Nit: we tend to name these variables `cmdline`.

mkosi/__init__.py
```diff
@@ -3585,26 +3607,37 @@ def make_minimal_btrfs(args: CommandLineArguments, root: str, for_cache: bool) -
...
+
+        if args.minimal:
+            command.append("--shrink")
+            try:
+                run(command)
+            except subprocess.CalledProcessError:
+                # The --shrink option was added in btrfs-tools 4.14.1, before that it was the default behaviour.
+                # If the above fails, let's see if things work if we drop it
+                command.remove("--shrink")
+                run(command)
```
behrmann, line 3620: Instead of adding and then removing it again, this might be clearer
```suggestion
            try:
                run(command + ["--shrink"])
            except subprocess.CalledProcessError:
                # The --shrink option was added in btrfs-tools 4.14.1, before that it was the default behaviour.
                # If the above fails, let's see if things work without it
                run(command)
```

behrmann (commented):
Only tangentially related

mkosi/__init__.py
```diff
@@ -3585,26 +3607,37 @@ def make_minimal_btrfs(args: CommandLineArguments, root: str, for_cache: bool) -
...
-        except subprocess.CalledProcessError:
-            # The --shrink option was added in btrfs-tools 4.14.1, before that it was the default behaviour.
-            # If the above fails, let's see if things work if we drop it
-            command.remove("--shrink")
+        command = ["mkfs.btrfs", "-L", label, "-d", "single", "-m", "single", "--rootdir", root, f.name]
+
+        if args.minimal:
+            command.append("--shrink")
+            try:
+                run(command)
```
DaanDeMeyer, line 3615: This seems like it would clutter mkosi's output when --shrink is not supported. Would it be OK if we just grepped the --help output of mkfs.btrfs for --shrink and use that to determine whether we use it or not?

mkosi/__init__.py
```diff
@@ -3512,13 +3535,15 @@ def make_tar(args: CommandLineArguments, root: str, do_run_build_script: bool, f
     if for_cache:
         return None
 
+    tar_root_dir = (root + "/usr") if args.usr_only else root
```
DaanDeMeyer, line 3538: Ditto for using a f-string (see below)

mkosi/__init__.py
```diff
@@ -1324,7 +1347,7 @@ def prepare_root(args: CommandLineArguments, dev: Optional[str], cached: bool) -
     if cached:
         return
 
-    with complete_step("Formatting root partition"):
+    with complete_step("Formatting {} partition".format("usr" if args.usr_only else "root")):
```
DaanDeMeyer, line 1350: f-string here as well (see below)

DaanDeMeyer (commented):
LGTM, just a few nits.

If I understand this correctly, one way of using this is that on the initial OS install, systemd-repart would create the root partition and the /usr partition and initialize it from the partition generated by mkosi. How would updates to the /usr partition work? Would we "download" those to some temp location in the rootfs and then run systemd-repart again which would overwrite the existing /usr partition with the new one (or the B partition when we're using an A/B scheme)?

(I also need to get to writing some small framework so we can easily add more integration tests aside from the few existing ones. Once all the pieces of this larger work are in place, it'd be super nice to have some kind of test in mkosi (or systemd I guess) verifying it all works end-to-end)

poettering:
> If I understand this correctly, one way of using this is that on the initial OS install, systemd-repart would create the root partition and the /usr partition and initialize it from the partition generated by mkosi. How would updates to the /usr partition work? Would we "download" those to some temp location in the rootfs and then run systemd-repart again which would overwrite the existing /usr partition with the new one (or the B partition when we're using an A/B scheme)?

So the model I am primarily going for is like this:

1. with mkosi build a minimal GPT image that consists of an ESP + a squashfs /usr + a verity for that /usr. dd that onto the disk you want to use.
2. On first boot, in the initrd, add an empty second pair of partitions for /usr + verity for it; then add an encrypted root fs on the remaining space, and maybe other partitions such as home or swap (all enrolled to TPM if available) — i.e. run systemd-repart
3. on first update find an empty pair of /usr + verity for it, download the new partitions into that; then download efi unified kernel and drop that into ESP. reboot
4. at boot sd-boot will now automatically pick the kernel with the newer version string and it's embedded command line ensures it can only boot the root partition/verity that matches the same version, configured via hash
5. on next update we do the same as on the first update, but because now both sets of partitions  are filled we first empty the oldest (i.e. the one with the lowest version string), and also delete any matching kernel from the ESP. Then we do the exact same as on the first update.

In other words: the initial image ships a partition table like this:

1. ESP (includes unified kernel v0.1)
2. squashfs, immutable /usr fs (with OS v0.1)
3. verity for /usr (with the right hashes for the OS v0.1)

Then on first boot this is extended to:

1. ESP (includes unified kernel v0.1)
2. /usr fs (with OS v0.1)
3. verity for /usr (with the right hashes for the OS v0.1)
4. /usr fs nr. 2 (empty for now)
5. verity for /usr nr .2 (empty for now)
6. luksified root fs (with tpm and all bells and whistles)

Then on first update this becomes:

1. ESP (includes unified kernels v0.1 + v0.2)
2. /usr fs (with OS v0.1)
3. verity for /usr (with the right hashes for the OS v0.1)
4. /usr fs nr. 2 (with OS v0.2)
5. verity for /usr nr .2 (with the right hashes for OS v.0.2)
6. luksified root fs (with tpm and all bells and whistles)

On the second update this becomes:

1. ESP (includes unified kernels v0.2 + v0.3)
2. /usr fs (with OS v0.3)
3. verity for /usr (with the right hashes for the OS v0.3)
4. /usr fs nr. 2 (with OS v0.2)
5. verity for /usr nr .2 (with the right hashes for OS v.0.2)
6. luksified root fs (with tpm and all bells and whistles)

And so on.

On factory reset we simply reformat then root fs, leave the other 5 partitions the way they are.

And a live-installer could boot up exactly like that and use repart to set up another disk, in which case we set things up exactly like in the first step, and stream the data for that from the currently booted /usr and reboot.

This is the primary model I am focussing on, but this is supposed to be generic of course, there are other models I also care for a lot, like one where we ship a full immutable, versioned rootfs and then only generate a luksified/writable /var on first boot.

most of the components of what i describe above already exist in one form of another. It's mostly jut filling the gaps. my work on the updater component is WIP on this branch btw: https://github.com/poettering/systemd/commits/sysupdate – it's pretty complete already, but currently supports downloads only via plain http, not yet casync.

> (I also need to get to writing some small framework so we can easily add more integration tests aside from the few existing ones. Once all the pieces of this larger work are in place, it'd be super nice to have some kind of test in mkosi (or systemd I guess) verifying it all works end-to-end)

Oh, yeah, tests would be good!

mkosi/__init__.py
behrmann: I'm not sure that's worth it here, even on Debian stretch btrfs-tools 4.20 is available via backports. The only distro for which this might be a problem (I haven't checked) could be CentOS 7, so this should be a rather rare occurrence.

mkosi/__init__.py
```diff
@@ -3585,26 +3607,37 @@ def make_minimal_btrfs(args: CommandLineArguments, root: str, for_cache: bool) -
...
-        command = ["mkfs.btrfs", "-L", "root", "-d", "single", "-m", "single", "--shrink", "--rootdir", root, f.name]
-        try:
-            run(command)
-        except subprocess.CalledProcessError:
-            # The --shrink option was added in btrfs-tools 4.14.1, before that it was the default behaviour.
-            # If the above fails, let's see if things work if we drop it
-            command.remove("--shrink")
+        command = ["mkfs.btrfs", "-L", label, "-d", "single", "-m", "single", "--rootdir", root, f.name]
+
+        if args.minimal:
```
behrmann, line 3612: This was flagged by mypy
```suggestion
        if args.minimize:
```

behrmann (changes requested):

DaanDeMeyer:
Thanks for the detailed explanation! I'm Looking forward to having this available (in v249 maybe?)

mkosi/__init__.py
poettering: I remember I added this back because it was missing on the CI. 

If the CI is updated we could certainly drop the compat here, but maybe we should do this independently of this, this --shrink stuff is pre-existing code after all

poettering:
Force pushed a new version. Everything should be addressed but the --shrink thing, see comment above

poettering:
Updated again, incorporated @behrmann's fixes (thanks so much!) and did some minor additional fixes on the documentation. Let's see if this passes now.

poettering:
> Thanks for the detailed explanation! I'm Looking forward to having this available (in v249 maybe?)

I am trying for v249, but let's see... I need to address #625 first (and maybe #626) to actually have something to update with.

poettering:
Yay! It passed!

PR #679, opened 2021-03-17
Allow overriding the SSH key pair

mrc0mmand:
Allow using a pre-defined SSH key pair instead of generating a new one.
This should be useful, for example, in CI environments, where the images
are built and used at different times, and using the same SSH key pair
across multiple images is more convenient.

---

I'm not completely sure about correctness of the `unlink`-related shenanigans, so please feel free to kick me in the right direction if needed.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -812,6 +812,12 @@ details see the table below.
   with this  option and running the image using `mkosi boot` or `mkosi qemu`,
   the `mkosi ssh` command can be used to connect to the container/VM via SSH.
 
+`--ssh-key=`
+: If specified, use the given private key when connecting to the guest machine
+  via `mkosi ssh`. This requires the public key counterpart to be present at
+  the same location, suffixed with `.pub` (as done by `ssh-keygen`). If this
+  option is not present `mkosi` generates a new key pair automatically.
+
 ## Command Line Parameters and their Settings File Counterparts
 
 Most command line parameters may also be placed in an `mkosi.default`
@@ -891,6 +897,7 @@ which settings file options.
 | `--network-veth`                  | `[Host]`                | `NetworkVeth=`                |
 | `--ephemeral`                     | `[Host]`                | `Ephemeral=`                  |
 | `--ssh`                           | `[Host]`                | `Ssh=`                        |
+| `--ssh-key=`                      | `[Host]`                | `SshKey=`                     |
 
 Command line options that take no argument are not suffixed with a `=`
 in their long version in the table above. In the `mkosi.default` file
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -405,6 +405,7 @@ class CommandLineArguments:
     network_veth: bool
     ephemeral: bool
     ssh: bool
+    ssh_key: Optional[str]
     directory: Optional[str]
     default_path: Optional[str]
     all: bool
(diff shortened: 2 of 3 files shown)


behrmann:
Looks like a pretty sensible addition. :)

Couldn't this be whittled down a bit, though? Correct me if I'm wrong, but doesn't `ssh-keygen` always generate key pairs as `foo` and `foo.pub`? So in principle we could do with a single switch instead of one for the private and one for the public key. Am I missing something?

mrc0mmand:
> Couldn't this be whittled down a bit, though? Correct me if I'm wrong, but doesn't `ssh-keygen` always generate key pairs as `foo` and `foo.pub`? So in principle we could do with a single switch instead of one for the private and one for the public key. Am I missing something?

It indeed does and I though about it, but I wasn't sure if setting only the private key name wouldn't be too limiting (for my use case it's not, though). If there aren't any objections, I'd be happy to drop the public key part and make the code a bit simpler.

behrmann:
I think it would make sense to make the logic a bit simpler, then. :) @DaanDeMeyer, thoughts?

DaanDeMeyer:
Yeah, sounds good, let's keep it to a single option.

mkosi.md
```diff
@@ -812,6 +812,12 @@ details see the table below.
   with this  option and running the image using `mkosi boot` or `mkosi qemu`,
   the `mkosi ssh` command can be used to connect to the container/VM via SSH.
 
+`--ssh-private-key=`
```
behrmann, line 815: I think this could safely be called `--ssh-key=` now.

mkosi/__init__.py
```diff
@@ -5755,7 +5762,12 @@ def print_summary(args: CommandLineArguments) -> None:
         "    Output nspawn Settings: "
         + none_to_na(args.output_nspawn_settings if args.nspawn_settings is not None else None)
     )  # NOQA: E501
-    MkosiPrinter.info("            Output SSH key: " + none_to_na(args.output_sshkey if args.ssh else None))
+    if args.ssh_private_key:
+        MkosiPrinter.info("           Private SSH key: " + args.ssh_private_key)
+        MkosiPrinter.info(f"            Public SSH key: {args.ssh_private_key}.pub")
+    else:
+        MkosiPrinter.info("            Output SSH key: " + none_to_na(args.output_sshkey if args.ssh else None))
+
```
behrmann, line 5770: I think we can fold this into one
```suggestion
        MkosiPrinter.info("                   SSH key: " + none_to_na((args.ssh_private_key or args.output_sshkey) if args.ssh else None))

```
`or` is short-circuiting, so it will take the first truthty (i.e. non-`None`, non-empty) string

mkosi/__init__.py
```diff
@@ -6601,11 +6619,12 @@ def find_address(args: CommandLineArguments) -> Tuple[str, str]:
 
 
 def run_ssh(args: CommandLineArguments) -> None:
-    assert args.output_sshkey is not None
+    ssh_key = args.ssh_private_key if args.ssh_private_key else args.output_sshkey
```
behrmann, line 6622:
```suggestion
    ssh_key = args.ssh_private_key or args.output_sshkey
```
This is also compatible with the following assert, since if both or `None`, the return will be `None`.

mkosi.md
mrc0mmand: Quite indeed!

mkosi/__init__.py
mrc0mmand: This is definitely better, thanks.

mkosi/__init__.py
mrc0mmand: Perfect, thanks, that looks a way better.

behrmann:
Haven't run it, but LGTM.

PR #680, opened 2021-03-18
mkosi: don't make efivarfs.ko a hard dep

poettering:
Some distros build this as kmod, others build it in. Hence only package
it up if it exists, i.e. pull it in weakly only.

(Quite frankly, distros that don't build it in are stupid, they should
fix that, it's needed almost always and is tiny.)

Follow-up for 9244e1dcdc056a7e9c457222a858a852a7965078

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1597,7 +1597,9 @@ def configure_dracut(args: CommandLineArguments, root: str) -> None:
     # efivarfs must be present in order to GPT root discovery work
     if args.esp_partno is not None:
         with open(os.path.join(dracut_dir, "30-mkosi-efivarfs.conf"), "w") as f:
-            f.write('add_drivers+=" efivarfs "\n')
+            f.write(
+                '[[ $(modinfo -k "$kernel" -F filename efivarfs 2>/dev/null) == /* ]] && add_drivers+=" efivarfs "\n'
+            )
 
 
 def prepare_tree_root(args: CommandLineArguments, root: str) -> None:


poettering:
(and yeah dracut docs suck, but drivers+= is the graceful alternative to add-drivers+=)

poettering:
(this makes bootable image work again on fedora)

bluca:
Whops, sorry, wasn't aware of this better keyword

bluca (approved):
lgtm

DaanDeMeyer:
CI doesn't like it! I agree we need to fix this though, It causes an annoying dracut warning when building Arch images as well

poettering:
hmm, wah? i don't get it though? why would that fail?

poettering:
my i misunderstood dracut on this?

```
       drivers+=" <kernel modules> "
           Specify a space-separated list of kernel modules to exclusively include in the initramfs. The kernel modules have to be specified without the ".ko" suffix.

       add_drivers+=" <kernel modules> "
           Specify a space-separated list of kernel modules to add to the initramfs. The kernel modules have to be specified without the ".ko" suffix.
```

That's the documentation of these two switches. I have no idea what either means compared to the other? What does "exclusively include" even mean?

DaanDeMeyer:
I suspect we're removing a bunch of "default" drivers by using drivers instead of add_drivers.

Maybe we can just conditionally write the file only on debian/ubuntu instead of relying on dracut?

poettering:
@haraldh any chance you can enlighten us on this? we are looking for a way to include efivarfs.ko in dracut on systems where it exists, but not where it doesn't. How do we do that?

behrmann:
I mean, this might be silly, but isn't the dracut config just a bash snippet? So couldn't we just do this?
```bash
[ -f /usr/lib/modules/$KERNELVERSION/kernel/drivers/firmware/efi/efivars.ko ] && add_drivers+=" efivarfs "
```

bluca:
Also if it's just for Debian/Ubuntu support, maybe check explicitly for those before adding the snippet could be a cleaner solution

haraldh:
NAK... `add_drivers` was fine.

First we need a [fix](https://github.com/dracutdevs/dracut/pull/1158) for dracut to not error on builtin modules.
Then we need something like `add_optional_drivers` like `install_optional_items`, so, that no error message is displayed, if the kernel module was not shipped.

haraldh:
> I mean, this might be silly, but isn't the dracut config just a bash snippet? So couldn't we just do this?
> 
> ```shell
> [ -f /usr/lib/modules/$KERNELVERSION/kernel/drivers/firmware/efi/efivars.ko ] && add_drivers+=" efivarfs "
> ```

This is the easy way workaround for now.

haraldh:
> I mean, this might be silly, but isn't the dracut config just a bash snippet? So couldn't we just do this?
> 
> ```shell
> [ -f /usr/lib/modules/$KERNELVERSION/kernel/drivers/firmware/efi/efivars.ko ] && add_drivers+=" efivarfs "
> ```

This might be better:
```
[[ $(modinfo -k "$KERNELVERSION" -F filename efivarfs 2>/dev/null) == /* ]] && add_drivers+=" efivarfs "
```

haraldh:
> Then we need something like `add_optional_drivers` like `install_optional_items`, so, that no error message is displayed, if the kernel module was not shipped.

https://github.com/dracutdevs/dracut/issues/1159

poettering:
@haraldh thanks a lot for reporting back!

hmm, what's the difference between drivers= and add-drivers= though? Any chance you can explain (best would be in a man page update?) i.e. i am puzzled what "exclusively including" from the docs is supposed to mean?

haraldh:
As written in the man page `drivers=` specifies the exact kernel module list.. nothing will be added by dracut modules.
`add_drivers=` adds those drivers to the list of kernel modules specified by the used dracut modules.

poettering:
Force pushed new version with the suggested line. Thanks for the suggestion!

haraldh:
Oh.. hmm, do you even use export `KERNELVERSION` somewhere?

mkosi/__init__.py
```diff
@@ -1589,7 +1589,7 @@ def configure_dracut(args: CommandLineArguments, root: str) -> None:
     # efivarfs must be present in order to GPT root discovery work
     if args.esp_partno is not None:
         with open(os.path.join(dracut_dir, "30-mkosi-efivarfs.conf"), "w") as f:
-            f.write('add_drivers+=" efivarfs "\n')
+            f.write('[[ $(modinfo -k "$KERNELVERSION" -F filename efivarfs 2>/dev/null) == /* ]] && drivers+=" efivarfs "\n')
```
haraldh, line 1592: Might want to use the dracut `$kernel` variable here. Yeah, I know... lower case and the name is not clear.

```suggestion
            f.write('[[ $(modinfo -k "$kernel" -F filename efivarfs 2>/dev/null) == /* ]] && drivers+=" efivarfs "\n')
```

haraldh (changes requested):

poettering:
> Oh.. hmm, do you even export `KERNELVERSION` somewhere?

Hmm, now I am weirded out why this didn't fail for me when I tested it. /me is puzzled

poettering:
> > Oh.. hmm, do you even export `KERNELVERSION` somewhere?
> 
> Hmm, now I am weirded out why this didn't fail for me when I tested it. /me is puzzled

Oh, right of course, I test this on Fedora, where the module is built in, hence if using hte wrong path of coure nothing fails... Makes sense.

poettering:
Force pushed a new version, let's see if that passes CI! @haraldh thanks so much for all the help!

poettering:
Sniff. That didn#t work

mkosi/__init__.py
```diff
@@ -1597,7 +1597,7 @@ def configure_dracut(args: CommandLineArguments, root: str) -> None:
     # efivarfs must be present in order to GPT root discovery work
     if args.esp_partno is not None:
         with open(os.path.join(dracut_dir, "30-mkosi-efivarfs.conf"), "w") as f:
-            f.write('add_drivers+=" efivarfs "\n')
+            f.write('[[ $(modinfo -k "$kernel" -F filename efivarfs 2>/dev/null) == /* ]] && drivers+=" efivarfs "\n')
```
mrc0mmand, line 1600: Shouldn't this remain `add_drivers+=" efivarfs "` instead of just `drivers+=" efivarfs "`? Currently this drops all other drivers, which would explain why Debian CIs fail. Otherwise this fix works, the `efivarfs`-related error is not present on systems where the `efivarfs` module is not a separate thing.

haraldh: yes... `add_drivers`

mrc0mmand:
Wooo, everything is green!

PR #682, opened 2021-03-18
mkosi: fix bootable + verity /usr/ builds

poettering:
This fixes two issues with UsrOnly=yes builds that are bootable and have
verity:

1. use the right partition type uuid for the verity partition
2. fix the umount path when we invoke dracut in a /usr-only build, since
   "umount --recursive" insists on having a mount point at the specified
   path, which we previously didn't necessarily. Let's add a bind mount
   on itself there to address this.

A follow-up for: #676

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1421,7 +1421,10 @@ def mount_loop(args: CommandLineArguments, dev: str, where: str, read_only: bool
     run(cmd)
 
 
-def mount_bind(what: str, where: str) -> str:
+def mount_bind(what: str, where: Optional[str] = None) -> str:
+    if where is None:
+        where = what
+
     os.makedirs(what, 0o755, True)
     os.makedirs(where, 0o755, True)
     run(["mount", "--bind", what, where])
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -1446,7 +1449,12 @@ def mount_image(
     with complete_step("Mounting image"):
 
         if root_dev is not None:
-            mount_loop(args, root_dev, root, root_read_only)
+            if args.usr_only:
+                # In UsrOnly mode let's have a bind mount at the top so that umount --recursive works nicely later
+                mount_bind(root)
+                mount_loop(args, root_dev, os.path.join(root, "usr"), root_read_only)
+            else:
+                mount_loop(args, root_dev, root, root_read_only)
```
DaanDeMeyer, line 1457: Not sure I follow here, for usr_only, is_generated_root() returns True so mount_image() will always be called with root_dev set to None right which means we get the bnid mount from the else case of the parent if statement.

poettering: There are two calls of mount_image(). The second one is where this matters: where we already built almost evertyhing and now need to mount it in order to be able to call dracut to generate the right kernel for it. In that scond call root_dev will be non-None, and that's where this matters

poettering: (the second one is the one with the lambda stuff)

PR #683, opened 2021-03-18
optionally write out rootfs/verity/kernel as split out files in GPT images + a simple versioning logic

poettering:
Fix for #625 + #626

(The two first commits I also posted as #680  and #682, because they can be merged independently. I included them here because otherwise I cannot build test images with UsrOnly= + SplitArtifacts= + Bootable= on Fedora, i.e. my main focus at the moment I develop this for)

(And yes, the versioning and the UsrOnly= stuff don't really have to do too much with each other, but splitting this up is nasty because of the non-trivial rebases, so I figured I'll post it as one)


mkosi/__init__.py
```diff
@@ -6733,6 +6932,32 @@ def generate_secure_boot_key(args: CommandLineArguments) -> NoReturn:
...
+            m = int(v[-1])
+        except:
+            m = None
+
+        if m is None:
+            new_version = args.image_version + ".2"
+            print(
+                f"Last component of current version is not a decimal integer, appending '.2', bumping '{args.image_version}' → '{new_version}'."
+            )
+        else:
```
behrmann, line 6954: Just because I noticed this, while scrolling through:
```suggestion
        try:
            m = int(v[-1])
        except ValueError:
            new_version = args.image_version + ".2"
            print(
                f"Last component of current version is not a decimal integer, appending '.2', bumping '{args.image_version}' → '{new_version}'."
            )
        else:
```
The `else` can be subsumed into the try-except. Python has `else` clauses on basically everything. For try-except it's "run this code iff no exception was raised"

poettering:
Nice! All CI green!

poettering:
I split out the versioning stuff from this PR into #692, in the hope to get that merged/reviewed earlier.

poettering:
Split out more stuff now: #693 + #694 + #695 + #696

poettering:
Now that most has been merged, everything left unmerged in this PR is also in the split-out PR #693. Let' hence close this PR, and continue in #693.

Issue #684, opened 2021-03-22
Ensure synchronization between `mkosi qemu` and `mkosi ssh`


mrc0mmand:
Hello!

In my recent struggles I came across an issue I can't wrap my head around. I'd like to use `mkosi` in a scripted environment (CentOS CI), so for this purpose I run `mkosi qemu` via `systemd-run` (since that ensures `mkosi` doesn't leave the `qemu` process behind when terminated via `SIGTERM`) and then run a couple of arbitrary commands via `mkosi ssh`:

```bash
MKOSI_UNIT="mkosi-build-image-$RANDOM.service"
systemd-run --unit "$MKOSI_UNIT" -p WorkingDirectory="$PWD" -- mkosi "${MKOSI_ARGS[@]}" qemu
systemctl --no-pager -l status "$MKOSI_UNIT"

mkosi "${MKOSI_ARGS[@]}" ssh -o BatchMode=yes "echo 'hello world'"
```

This, however, poses a problem - `mkosi ssh` will fail with `Error: Container/VM interface not found`, since `systemd-run` returns immediately without waiting for `mkosi` to configure the veth interface and spawn the `qemu` process.

Fine, let's "fix" this (I apologize beforehand for the following code):

```bash
MKOSI_UNIT="mkosi-build-image-$RANDOM.service"
systemd-run --unit "$MKOSI_UNIT" \
            -p WorkingDirectory="$PWD" \
            -p ExecStartPost="/bin/bash -c \"while ! /bin/systemd-cgls --unit $MKOSI_UNIT | grep -E '[0-9]+ qemu-kvm '; do sleep 1; done\"" \
            -- mkosi "${MKOSI_ARGS[@]}" qemu
systemctl --no-pager -l status "$MKOSI_UNIT"

mkosi "${MKOSI_ARGS[@]}" ssh -o BatchMode=yes "echo 'hello world'"
```

With this *work of art* `systemd-run` waits until the `qemu-kvm` process pops up in the service's cgroup and then returns. But that's still not enough, since the networking may not be up, causing `mkosi ssh` to fail with `Error: Container/VM address not found`.

Originally I thought about using `mkosi -o ConnectionAttempts=60 ssh` but that doesn't apply in this case either, since the error is thrown by `mkosi`. My current *solution* is to call `mkosi ssh` in a loop for a minute and then eventually give up, but that's not very pretty.

[...]

PR #685, opened 2021-03-23
Attempt to retry the SSH connection if the veth/SSH server is not up yet

mrc0mmand:
This commit introduces a new option `--ssh-timeout=` to allow setting
a timeout for which `mkosi` will attempt to retry the SSH connection in
case the link not properly set up yet (e.g. when running `mkosi qemu` and
`mkosi ssh` in a script, the veth may not be configured yet, causing the
ssh connection to fail).

Fixes: #684

---

My attempt at fixing #684. It behaved as expected when I tested it, but I still have a feeling it could be written much nicer. PTAL!

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -407,6 +407,7 @@ class CommandLineArguments:
     ephemeral: bool
     ssh: bool
     ssh_key: Optional[str]
+    ssh_timeout: int
     directory: Optional[str]
     default_path: Optional[str]
     all: bool
@@ -4842,6 +4843,13 @@ def create_parser() -> ArgumentParserMkosi:
         metavar="PATH",
         help="Use the specified private key when using `mkosi ssh` (requires a corresponding public key)",
     )
+    group.add_argument(
+        "--ssh-timeout",
+        metavar="SECONDS",
+        type=int,
+        default=0,
+        help="Wait up to SECONDS seconds for the SSH connection to be available when using `mkosi ssh`",
+    )
 
     group = parser.add_argument_group("Additional Configuration")
     group.add_argument("-C", "--directory", help="Change to specified directory before doing anything", metavar="PATH")
@@ -5693,6 +5701,9 @@ def load_args(args: argparse.Namespace) -> CommandLineArguments:
 
     if args.ssh and not args.network_veth:
         die("--ssh cannot be used without --network-veth")
+
+    if args.ssh_timeout < 0:
+        die("--ssh-timeout must be >= 0")
 
     args.original_umask = os.umask(0o000)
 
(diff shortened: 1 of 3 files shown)


behrmann:
Interesting, I do think, that the lgtm-bot is wrong here.

mrc0mmand:
> Interesting, I do think, that the lgtm-bot is wrong here.

He's right, the condition was indeed unreachable, since the `timeout` might get below zero only just after the `except` section.

behrmann:
> > Interesting, I do think, that the lgtm-bot is wrong here.
> 
> He's right, the condition was indeed unreachable, since the `timeout` might get below zero only just after the `except` section.

Ah, I thought that was intentional, but on right, this doesn't make sense.

mkosi/__init__.py
```diff
@@ -4842,6 +4843,13 @@ def create_parser() -> ArgumentParserMkosi:
         metavar="PATH",
         help="Use the specified private key when using `mkosi ssh` (requires a corresponding public key)",
     )
+    group.add_argument(
+        "--ssh-timeout",
+        metavar="SECONDS",
+        type=int,
+        default=0,
+        help="Wait up to SECONDS seconds for the SSH connection to be available when using `mkosi ssh`",
```
behrmann, line 4851: This should also be added to the docs. 

Should we also document, that this is mainly to wait for the devices to come online, since there is the 20s loop to ping neighbours?

mkosi/__init__.py
```diff
@@ -6623,33 +6634,48 @@ def interface_exists(dev: str) -> bool:
 
 def find_address(args: CommandLineArguments) -> Tuple[str, str]:
     name = virt_name(args)
+    last_exception = None
```
behrmann, line 6637: This line is not strictly necessary here, isn't it? It will be reset right at the start of the loop.

mkosi/__init__.py
```diff
@@ -6623,33 +6634,48 @@ def interface_exists(dev: str) -> bool:
...
-            dst = cast(str, neighbor["dst"])
-            if dst.startswith("fe80"):
-                return dev, dst
+                time.sleep(0.4)
+        except MkosiException as e:
+            last_exception = str(e)
 
-        time.sleep(0.4)
+        time.sleep(1)
+        timeout = timeout - (time.time() - stime)
```
behrmann, line 6676: I think this is slightly easier to read
```suggestion
        timeout -= time.time() - stime
```

mkosi/__init__.py
mrc0mmand: I knew I forgot something, thanks. I added a documentation snipped mentioning the use case.

mkosi/__init__.py
mrc0mmand: Ah, quite indeed!

mkosi/__init__.py
mrc0mmand: It sure is, thanks.

mrc0mmand:
Thanks a lot for the review!

mkosi/__init__.py
```diff
@@ -6623,33 +6634,47 @@ def interface_exists(dev: str) -> bool:
...
+                    if dst.startswith("fe80"):
+                        return dev, dst
 
-        for neighbor in neighbors:
-            dst = cast(str, neighbor["dst"])
-            if dst.startswith("fe80"):
-                return dev, dst
+                time.sleep(0.4)
+        except MkosiException as e:
+            last_exception = str(e)
```
DaanDeMeyer, line 6672: I think we might not need the last_exception variable if you check if the timeout has expired in the except clause here and reraise the exception if the timeout has expired.

This also means that we'll always try to run through the loop once, even if the timeout is 0 (which I think is a good thing)

mrc0mmand: That actually was the original approach (sans some minor details), but I didn't want to make the timeout calculation twice. But I guess it would look nicer than the current `last_exception` approach. I'll prep a fix shortly.

mrc0mmand: Ah, simple `raise` would hide the actual error message, so we have to use either `die()` or `warn(); raise`, which is not as pretty.

mkosi/__init__.py
```diff
@@ -6623,31 +6634,45 @@ def interface_exists(dev: str) -> bool:
...
+                    if dst.startswith("fe80"):
+                        return dev, dst
 
-        for neighbor in neighbors:
-            dst = cast(str, neighbor["dst"])
-            if dst.startswith("fe80"):
-                return dev, dst
+                time.sleep(0.4)
+        except MkosiException as e:
+            if timeout - (time.time() - stime) < 0:
```
behrmann, line 6671: I think this might be clearer
```suggestion
            if (time.time() - stime) > timeout:
```
The parentheses are optional.

mrc0mmand: Indeed, thanks!

DaanDeMeyer:
LGTM, hopefully this makes the ssh stuff a bit more robust.

mrc0mmand:
Thank y'all for the feedback, much appreciated!

PR #690, opened 2021-04-10
[distro] add support for gentoo

257:


behrmann:
Thanks for the contribution @257!

The unit test failure is a bit weird, I guess `isort` doesn't like the import in the function. I'll have a look at that. 

Besides that this should definitely be added to the docs and I think that when we add new distros, they should add accompanying boot tests, otherwise we will be blind for future breakage.

mkosi/__init__.py
```diff
@@ -2241,6 +2242,56 @@ def install_openmandriva(args: CommandLineArguments, root: str, do_run_build_scr
...
+
+    from _emerge.actions import load_emerge_config
+    from _emerge.actions import action_info
+    from _emerge.actions import action_sync
+    from _emerge.actions import action_build
+
+    portdir = "/var/db/repos/gentoo"
+    pkgdir = "/var/cache/binpkgs"
+    os.makedirs(portdir, exist_ok=1)
+    os.makedirs(pkgdir, exist_ok=1)
```
behrmann, line 2260: Please use booleans, this is not C :)
```suggestion
    os.makedirs(portdir, exist_ok=True)
    os.makedirs(pkgdir, exist_ok=True)
```

mkosi/__init__.py
```diff
@@ -2241,6 +2242,56 @@ def install_openmandriva(args: CommandLineArguments, root: str, do_run_build_scr
...
+
+    # TODO: add hooks for setting profile in args.gentoo_profile
+    os.symlink(
+        os.path.join(portdir, "profiles/default/linux/amd64/17.1/no-multilib"),
```
behrmann, line 2264: What is the `17.1` here? That number seems somewhat magical to me.

mkosi/__init__.py
```diff
@@ -2241,6 +2242,56 @@ def install_openmandriva(args: CommandLineArguments, root: str, do_run_build_scr
...
+def install_gentoo(args: CommandLineArguments, root: str, do_run_build_script: bool) -> None:
+    try:
+        import portage
+    except ImportError:
+        die("you need portage module for Gentoo: https://gitweb.gentoo.org/proj/portage.git")
+
+    from _emerge.actions import load_emerge_config
+    from _emerge.actions import action_info
+    from _emerge.actions import action_sync
+    from _emerge.actions import action_build
```
behrmann, line 2255: The `portage` import doesn't seem to be needed? If you don't, leave it out and move the imports from `_emerge` into the try block.

Also, move all the `_emerge` imports into a single statement.

behrmann (commented):
Here some comments.

The TODOs should be addressed, because they are not that helpful for anybody who doesn't do any Gentoo :)

mkosi/__init__.py
257: noted

mkosi/__init__.py
257: ```17.1``` is just the latest profile version; there is 17.0 but that's about it. gentoo folks only support 17.1 btw.

but more important bit there is the ARCH (amd64) i think. this would matter for IoT et al.

in general profile in something left to users to choose; see here for more:
 https://wiki.gentoo.org/wiki/Profile_(Portage) 
i hesitated exposing this via, say, ```---gentoo-profile```/```GentooProfile```; it's internal to gentoo and i don't know if there is an equivalence for it in other distros or if if there is one whether ```mkosi``` supports such thing.

i will put together something for now.

257:
> Thanks for the contribution @257!
> 
> The unit test failure is a bit weird, I guess `isort` doesn't like the import in the function. I'll have a look at that.

thanks.
i also just install ```isort``` will give it a spin shortly.

> 
> Besides that this should definitely be added to the docs and I think that when we add new distros, they should add accompanying boot tests, otherwise we will be blind for future breakage.

will look into documenting.

boot is a bigger task; kernel is hand-compiled in gentoo; or that's how i have done it personally... there is an automated way, genkernel, and there is also a binary kernel. i have run into trouble trying those since they conflict with ```systemd```'s handling of kernel installation. i though maybe boot should be the next stage once the primary hooks are in. in any events, i will start working on boot first chance i have.

mkosi/__init__.py
behrmann: Sorry for replying late. I don't think there is a need to add distro specific flags. We've managed without so far. If the profiles are something like the release versions for other profiles, than it would make sense to reuse those flags.

257:
On Mon, Apr 12, 2021 at 10:25:37AM -0700, Jörg Behrmann wrote:
> Sorry for replying late. I don't think there is a need to add distro specific flags. We've managed without so far. If the profiles are something like the release versions for other profiles, than it would make sense to reuse those flags.

profiles must be set or ```emerge``` is no go.

i don't think i understand what you mean by *release version of other profiles*.

from what i understand, *profiles*, are more like:

*stage/os/architecture/release-version/flavour*

they can mostly be narrowed down to sane default options, except for flavour.
(flavour is how i call it)

* stage: most probably ```default``` for 99.9% of users
* os: i guess only linux (gentoo supports bsd/darwin/plan9?)
* architecture: i can read it out of ```---architecture``` and only accept
amd64, ppc, arm, riscv.
* release-version: same as stage, almost invariably the lastes for users
* flavour: this is where it gets interesting, there is a plethora of choices;
systemd, musl, no-multilib, hardened, x32... i don't see any defaults here.

i'm not aware of other distros having the same notion like flavour so i don't
know if there would be a general flag to hide flavour behind it reuse it with
other distros.

combination of architecture+flavour is what makes this useful for IoT btw.

257:
> > Thanks for the contribution @257!
> > The unit test failure is a bit weird, I guess `isort` doesn't like the import in the function. I'll have a look at that.
> 
> thanks.
> i also just install `isort` will give it a spin shortly.
> 
> > Besides that this should definitely be added to the docs and I think that when we add new distros, they should add accompanying boot tests, otherwise we will be blind for future breakage.
> 
> will look into documenting.
> 
> boot is a bigger task; kernel is hand-compiled in gentoo; or that's how i have done it personally... there is an automated way, genkernel, and there is also a binary kernel. i have run into trouble trying those since they conflict with `systemd`'s handling of kernel installation. i though maybe boot should be the next stage once the primary hooks are in. in any events, i will start working on boot first chance i have.


profiles must be set or ```emerge``` is no go.

i don't think i understand what you mean by *release version of other profiles*.

from what i understand, *profiles*, are more like:

*stage/os/architecture/release-version/flavour*

they can mostly be narrowed down to sane default options, except for flavour.
(flavour is how i call it)

* stage: most probably ```default``` for 99.9% of users
* os: i guess only linux (gentoo supports bsd/darwin/plan9?)
* architecture: i can read it out of ```---architecture``` and only accept
amd64, ppc, arm, riscv.
* release-version: same as stage, almost invariably the lastes for users
* flavour: this is where it gets interesting, there is a plethora of choices;
systemd, musl, no-multilib, hardened, x32... i don't see any defaults here.

i'm not aware of other distros having the same notion like flavour so i don't
know if there would be a general flag to hide flavour behind it reuse it with
other distros.

combination of architecture+flavour is what makes this useful for IoT btw.

behrmann:
> i don't think i understand what you mean by _release version of other profiles_.
> 
> from what i understand, _profiles_, are more like:
> 
> _stage/os/architecture/release-version/flavour_

What I meant is that this sounds like something the `--release` option can be used for. 

>     * flavour: this is where it gets interesting, there is a plethora of choices;
>       systemd, musl, no-multilib, hardened, x32... i don't see any defaults here.
> 
> i'm not aware of other distros having the same notion like flavour so i don't
> know if there would be a general flag to hide flavour behind it reuse it with
> other distros.

I don't know gentoo well enough to really weigh this, but I think mkosi is about sane, modern, mostly uniform defaults, so let that be the guide. mkosi doesn't built non-systemd images, so that should be the flavour of choice. systemd depends on glibc, so musl is out. x32 is mostly dead, if gentoo is multilib by default, than keep that. I have no opinions on extra hardening, but I'd try to make it as vanilla as possible.

mkosi/__init__.py
```diff
@@ -2241,6 +2242,95 @@ def install_openmandriva(args: CommandLineArguments, root: str, do_run_build_scr
...
+def install_gentoo(args: CommandLineArguments, root: str, do_run_build_script: bool) -> None:
+    try:
+        from _emerge.actions import action_build, action_sync, load_emerge_config
+
+        if "build-script" in args.debug:
+            from _emerge.actions import action_info
+    except ImportError:
+        die("You need portage module for Gentoo: https://gitweb.gentoo.org/proj/portage.git")
+
+    import multiprocessing
```
DaanDeMeyer, line 2255: Move this one to the top of the file with the rest of the imports

mkosi/__init__.py
```diff
@@ -2241,6 +2242,95 @@ def install_openmandriva(args: CommandLineArguments, root: str, do_run_build_scr
     disable_pam_securetty(root)
 
 
+@complete_step("Installing Gentoo")
+def install_gentoo(args: CommandLineArguments, root: str, do_run_build_script: bool) -> None:
+    try:
+        from _emerge.actions import action_build, action_sync, load_emerge_config
```
DaanDeMeyer, line 2248: Is this a python API to emerge? Current practice in mkosi is to not depend on any python library dependencies. Can you call emerge via the command line instead?

DaanDeMeyer (commented):
Looks pretty good, left some comments.

If you can make this boot in CI (nspawn/qemu) that would be even better.

mkosi/__init__.py
behrmann: I think we could maybe loosen this as long as this is in distro-specific code, when its confined to a function, and significantly shortens our work. When people build emerge it should be little overhead to also build the python bindings.

behrmann:
> If you can make this boot in CI (nspawn/qemu) that would be even better.

I think that's maybe something we should require when including new distros, otherwise we won't know what breaks down the line. What do you think?

mkosi/__init__.py
257: shell version:
https://github.com/systemd/mkosi/pull/691

mkosi/__init__.py
257: done

DaanDeMeyer:
I would very strongly suggest it but if gentoo has to compile for a few hours before it can run in CI, that's not exactly going to be something we can add to CI. Of course, if it's not in CI, we make no guarantees and it could break at any moment (although of course we won't actively try to break it)

behrmann:
That's a good point, but I looking at the [gentoo wiki](https://wiki.gentoo.org/wiki/Binary_package_guide), they do seem to have support for pre-compiled binaries. I don't know whether they run official mirrors for those, but if so that could be used for the CI case. Maybe @257 can shed some light on this.

mkosi/__init__.py
```diff
@@ -2241,6 +2242,100 @@ def install_openmandriva(args: CommandLineArguments, root: str, do_run_build_scr
     disable_pam_securetty(root)
 
 
+@complete_step("Installing Gentoo")
+def install_gentoo(args: CommandLineArguments, root: str, do_run_build_script: bool) -> None:
+    try:
+        import multiprocessing
+
+        from _emerge.actions import (
```
poettering, line 2250: interesting, it would be the first time we do this, i.e. use the packaging libs directly. But I must say, I kinda sympathize with this. And we probably should do the same for other OSes where we can, i.e. use the dnf python modules on fedora for example.

Why that? I think in the dnf case at least this opens up new possibilities for us, i.e. we could add support for "pinning" package version: i.e. think that you first build your OS like you always do, but we output RPM package version data to some file while doing so. Next time when we rebuild we read that file and make sure to pick the precise same versions of the RPMs. And if you drop the file you basically update to current RPMs. Something like that would hopefully bring us a bit closer to reproducible builds in the long run.

poettering:
> I don't know gentoo well enough to really weigh this, but I think mkosi is about sane, modern, mostly uniform defaults, so let that be the guide. mkosi doesn't built non-systemd images, so that should be the flavour of choice. systemd depends on glibc, so musl is out. x32 is mostly dead, if gentoo is multilib by default, than keep that. I have no opinions on extra hardening, but I'd try to make it as vanilla as possible.

(I mostly agree with this, just one note: we don't strictly depend on glibc, we just say that glibc defines the Linux API, and as long as musl conforms to thta we are fine with that. If there's a gnu specific API that has a POSIX counterpart we usually prefer the POSIX one to make things easer, but we do not carry compat kludges for other libcs in our tree, and we do use useful APIs of glibc that are outside of POSIX aplenty. Thus, if musl picks up the GNU APIs that we use all should be ok with musl too. The GNU APIs we use are not crazy stuff btw, but pretty sensible APIs. Thus as long as musl is not a religious POSIX-or-bust thing then musl should work ok too with systemd, even though we personally don't run it)

mkosi/__init__.py
poettering: as commented below I'd be fine with using python APIs of packaging scripts, see below for my reasoning.

poettering:
(I'd probably consider these "flavours" of gentoo [i.e. glibc vs. musl, …] as distinct distros in mkosi. i mean, they are a close family, but that's centos and fedora too in a way I'd claim, maybe not quite as close. hence, i'd start with the flavour that actually makes sense, and and if people really want others then add them as separate distros. But yeah, stuff like x32 is on its deathbed i doubt we should add that)

poettering:
(and one of those days we should probably start splitting up the humungous script in more digestable files. my python-fu is not good enough to do know how this done the most pythonic way though)

behrmann:
> (and one of those days we should probably start splitting up the humungous script in more digestable files. my python-fu is not good enough to do know how this done the most pythonic way though)

I've experimented with this a bit a while ago [on a branch](https://github.com/systemd/mkosi/compare/main...behrmann:split), but stopped because of urgent things at `$DAYJOB`. I guess I can pick it up again. My idea was to bundle all installation specific functions into a class, with the common functions in the default implementation. Distros would inherit from that default implementation and override the methods if they have more specific needs. The most difficult bit about all of this is picking apart what should stay in `args` and what can move to the distros proper, which is why there is an ugly escape hatch with a circular reference.

257:
On Thu, Apr 15, 2021 at 12:32:32AM -0700, Jörg Behrmann wrote:
> That's a good point, but I looking at the [gentoo
> wiki](https://wiki.gentoo.org/wiki/Binary_package_guide), they do seem to have
> support for pre-compiled binaries. I don't know whether they run official
> mirrors for those, but if so that could be used for the CI case. Maybe @257
> can shed some light on this.

no official mirrors that i'm aware of. consecutive runs however can and will
take advantage of binary packages, à la ```binpkg```, so long as CI doesn't
clean up the artefacts (packaged binaries) in ```pkgdir``` and ```${USE}```[1]
flags are not changed and there is no update to the package version itself.

[1]```${USER}``` are basically flags passed to ```autotools``` and friends.

257:
On Thu, Apr 15, 2021 at 01:34:08AM -0700, Lennart Poettering wrote:
> interesting, it would be the first time we do this, i.e. use the packaging
> libs directly. But I must say, I kinda sympathize with this. And we probably
> should do the same for other OSes where we can, i.e. use the dnf python
> modules on fedora for example.

@poettering, does that mean i should drop the shell version?  https://github.com/systemd/mkosi/pull/691

> 
> Why that? I think in the dnf case at least this opens up new possibilities for
> us, i.e. we could add support for "pinning" package version: i.e. think that
> you first build your OS like you always do, but we output RPM package version
> data to some file while doing so. Next time when we rebuild we read that file
> and make sure to pick the precise same versions of the RPMs. And if you drop
> the file you basically update to current RPMs. Something like that would
> hopefully bring us a bit closer to reproducible builds in the long run.

gentoo also can do "pinning" via ```/etc/portage/package.mask```:
https://wiki.gentoo.org/wiki//etc/portage/package.mask
and ```/etc/portage/package.accept_keywords```:
https://wiki.gentoo.org/wiki//etc/portage/package.accept_keywords

257:
On Thu, Apr 15, 2021 at 01:44:39AM -0700, Lennart Poettering wrote:
> (I'd probably consider these "flavours" of gentoo [i.e. glibc vs. musl, …] as
> distinct distros in mkosi. i mean, they are a close family, but that's centos
> and fedora too in a way I'd claim, maybe not quite as close. hence, i'd start
> with the flavour that actually makes sense, and and if people really want
> others then add them as separate distros. But yeah, stuff like x32 is on its
> deathbed i doubt we should add that)

the patch, in its latest iteration, simply uses no flavour but allows user
familiar with gentoo profiles to pick one if need be.
no flavour seems to me to be the lowest common denominator and also what gentoo
seems to considers default, sort of; it comes first in their profile listing.

on my system:
```
***@***.***:~ eselect profile list
Available profile symlink targets:
  [1]   default/linux/amd64/17.1 (stable)
  [2]   default/linux/amd64/17.1/selinux (stable)
  [3]   default/linux/amd64/17.1/hardened (stable)
  [4]   default/linux/amd64/17.1/hardened/selinux (stable)
  [5]   default/linux/amd64/17.1/desktop (stable)
  [6]   default/linux/amd64/17.1/desktop/gnome (stable)
  [7]   default/linux/amd64/17.1/desktop/gnome/systemd (stable)
  [8]   default/linux/amd64/17.1/desktop/plasma (stable)
  [9]   default/linux/amd64/17.1/desktop/plasma/systemd (stable)
  [10]  default/linux/amd64/17.1/desktop/systemd (stable)
  [11]  default/linux/amd64/17.1/developer (stable)
  [12]  default/linux/amd64/17.1/no-multilib (stable)
  [13]  default/linux/amd64/17.1/no-multilib/hardened (stable)
  [14]  default/linux/amd64/17.1/no-multilib/hardened/selinux (stable)
  [15]  default/linux/amd64/17.1/systemd (stable) *
  [16]  default/linux/amd64/17.0 (dev)
  [17]  default/linux/amd64/17.0/selinux (dev)
  [18]  default/linux/amd64/17.0/hardened (dev)
  [19]  default/linux/amd64/17.0/hardened/selinux (dev)
  [20]  default/linux/amd64/17.0/desktop (dev)
  [21]  default/linux/amd64/17.0/desktop/gnome (dev)
  [22]  default/linux/amd64/17.0/desktop/gnome/systemd (dev)
  [23]  default/linux/amd64/17.0/desktop/plasma (dev)
  [24]  default/linux/amd64/17.0/desktop/plasma/systemd (dev)
  [25]  default/linux/amd64/17.0/developer (dev)
  [26]  default/linux/amd64/17.0/no-multilib (dev)
  [27]  default/linux/amd64/17.0/no-multilib/hardened (dev)
  [28]  default/linux/amd64/17.0/no-multilib/hardened/selinux (dev)
  [29]  default/linux/amd64/17.0/systemd (dev)
  [30]  default/linux/amd64/17.0/x32 (dev)
  [31]  default/linux/amd64/17.0/musl (exp)
  [32]  default/linux/amd64/17.0/musl/hardened (exp)
  [33]  default/linux/amd64/17.0/musl/hardened/selinux (exp)
  [34]  default/linux/amd64/17.0/uclibc (exp)
  [35]  default/linux/amd64/17.0/uclibc/hardened (exp)
```

mkosi/__init__.py
257: missed this one. great, i will continue with python API then.

257:
> That's a good point, but I looking at the [gentoo wiki](https://wiki.gentoo.org/wiki/Binary_package_guide), they do seem to have support for pre-compiled binaries. I don't know whether they run official mirrors for those, but if so that could be used for the CI case. Maybe @257 can shed some light on this.

no official mirrors that i'm aware of. consecutive runs however can and will
take advantage of binary packages, à la ```binpkg```, so long as CI doesn't
clean up the artefacts (packaged binaries) in ```pkgdir``` and ```${USE}```[1]
flags are not changed and there is no update to the package version itself.

[1]```${USER}``` are basically flags passed to ```autotools``` and friends.

mkosi/__init__.py
DaanDeMeyer: Sounds good, we can just depend on these being installed alongside the corresponding package manager.

@257 are the Python files part of the default emerge install?

mkosi/__init__.py
```diff
@@ -2241,6 +2242,100 @@ def install_openmandriva(args: CommandLineArguments, root: str, do_run_build_scr
     disable_pam_securetty(root)
 
 
+@complete_step("Installing Gentoo")
+def install_gentoo(args: CommandLineArguments, root: str, do_run_build_script: bool) -> None:
+    try:
+        import multiprocessing
```
DaanDeMeyer, line 2248: I meant move to the top of the entire file where all the other imports are located. The emerge imports can stay here as they only need to be available when gentoo is built.

mkosi/__init__.py
```diff
@@ -2241,6 +2242,100 @@ def install_openmandriva(args: CommandLineArguments, root: str, do_run_build_scr
...
+        "x86_64": "amd64",
+        "aarch64": "arm64",
+    }
+
+    if args.architecture:
+        gentoo_arch = GENTOO_ARCHITECTURES.get(args.architecture, "amd64")
+
+    # this should let users hack this flag by passing *flavour* for their image
+    # as in --release=17.1/no-multilib
+    GENTOO_LATEST = "17.1"
```
DaanDeMeyer, line 2296: Can you set this in `load_args`? That's where all the other distro release defaults are located.

DaanDeMeyer:
Can you also try to add this to CI? Take a look at the .github/workflows/ci.yml script.

I'm guessing emerge won't have a corresponding ubuntu package so you'll also have to modify action.yaml to build it from source if that's the case.

257:
> Can you also try to add this to CI? Take a look at the .github/workflows/ci.yml script.
> 
> I'm guessing emerge won't have a corresponding ubuntu package so you'll also have to modify action.yaml to build it from source if that's the case.

working on it. automated kernel installation seems to be a hit and miss.

257:
On Fri, Apr 16, 2021 at 10:04:40AM -0700, Daan De Meyer wrote:
> @257 are the Python files part of the default emerge install?

yes

mkosi/__init__.py
257: done

257:
> Looks pretty good, left some comments.
> 
> If you can make this boot in CI (nspawn/qemu) that would be even better.

it does now, *but* (arch, gpt_{ext4,btrfs}) fail and i have no idea why: https://github.com/systemd/mkosi/actions/runs/843083906

and it's random, they did succeed on my runs:
https://github.com/257/mkosi/actions/runs/843000400

257:
> Looks pretty good, left some comments.
> 
> If you can make this boot in CI (nspawn/qemu) that would be even better.

Done!: https://github.com/systemd/mkosi/actions/runs/845053559

PR #695, opened 2021-04-15
mkosi: insert rootfs in read-only mode if that's requested, not only …

poettering:
…if squashfs


poettering:
Split out of #683

behrmann:
LGTM.

poettering:
all green and @behrmann liked it. Let's merge!

PR #696, opened 2021-04-15
mkosi: add new initrd-usr-fs.target to initrd

poettering:
This is a recent addition in systemd, required to make cases work where
the initrd creates a root fs on first boot and images are initially
shipped without one.

(See https://github.com/systemd/systemd/pull/19234)

While we are at it, let's order the list of extra initrd files
alphabetically.


poettering:
Also split out of #683

behrmann:
LGTM, shouldn't change anything in the worst case, since the modules are included optionally.

poettering:
All green and liked by @behrmann. Merging!

PR #697, opened 2021-04-15
Replace "legacy-free" with "bespoke" in descriptions

keszybz:
Since the introduction of grub2 support, legacy-free is not really true.

  bespoke
    adj 1: (of clothing) custom-made [syn: {bespoke}, {bespoken},
           {made-to-order}, {tailored}, {tailor-made}]

    Any product that is special in some way, individually created for
    a specific user or system, as opposed to generic or off-the-shelf.

In the test, say "customized" to avoid repeating the same word.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
-# mkosi - Create legacy-free OS images
+# mkosi — Build Bespoke OS Images
 
-A fancy wrapper around `dnf --installroot`, `debootstrap`,
-`pacstrap` and `zypper` that may generate disk images with a number of
+A fancy wrapper around `dnf --installroot`, `debootstrap`, `pacstrap`
+and `zypper` that generates customized disk images with a number of
 bells and whistles.
 
 For a longer description and available features and options, see the
--- a/man/mkosi.1
+++ b/man/mkosi.1
@@ -5,7 +5,7 @@
 .hy
 .SH NAME
 .PP
-mkosi - Build Legacy-Free OS Images
+mkosi - Build Bespoke OS Images
 .SH SYNOPSIS
 .PP
 \f[C]mkosi [options\&...] build\f[R]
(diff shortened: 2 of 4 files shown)


behrmann:
Sounds good to me, although I do still like the legacy-free tagline even though it's technically not correct anymore. One could call it *opinionated*, since there's still a tight focus on what the images should look like, but I guess the connotations are worse than the ones for *bespoke*.

poettering:
I like bespoke better than legacy-free, since it's a more "positive" outlook. Defining yourself by what you are not is always a bit icky to me.

poettering:
lgtm

README.md
```diff
@@ -1,7 +1,7 @@
-# mkosi - Create legacy-free OS images
+# mkosi - Create bespoke OS images
```
poettering, line 1: maybe change this "Build Bespoke OS Image"? i.e. use uppercase since we use it almost everywhere else, and "build" instead of "create" since it's really about building and is what we use at most places already

man/mkosi.1
```diff
@@ -5,7 +5,7 @@
 .hy
 .SH NAME
 .PP
-mkosi - Build Legacy-Free OS Images
+mkosi - Build bespoke OS Images
```
poettering, line 8: uppercase?

setup.py
```diff
@@ -19,7 +19,7 @@ def run(self):
 setup(
     name="mkosi",
     version="9",
-    description="Create legacy-free OS images",
+    description="Create bespoke OS images",
```
poettering, line 22: uppercase? "build"?

keszybz:
Updated along the suggested lines.

poettering:
lgtm

DaanDeMeyer:
I'd probably go for "Custom" since I don't think most people will know the meaning of Bespoke without looking it up but I don't mind too much, let's merge.

keszybz:
> I'd probably go for "Custom" since I don't think most people will know the meaning of Bespoke without looking it up but I don't mind too much, let's merge.

"Custom" is very generic... "Build custom images" is not even googlable. But "build bespoke images" might be. That's why I used "custom[ized]" in the text, but "bespoke" in the tag line.

PR #698, opened 2021-04-15
tree-wide: change tagline to "Build Bespoke OS Images"

poettering:
@keszybz suggested we should drop the "legacy-free" tag, and he's right:
since we added support for BIOS supports we aren't really legacy-free
anymore.

Let's use a nice word less focussed on computer history instead: 🔥 bespoke 🔥


poettering:
Closing this one in favour of #697

PR #702, opened 2021-04-16
always use same tagline "Build Bespoke OS Images"

poettering:
Follow-up for b365720596d6ec6c54b3c5fe10d793e5596eeda6: two cases were
missed.


keszybz:
I have no idea how I missed those.

PR #704, opened 2021-04-16
mkosi: UsrOnly= btrfs fixups

poettering:
I put this together als alternative to f0386b00cfd2c897cbbaad50b0e717e7221676bc which got merged by now. 

Compared to what was merged it drops two explicit usr_only checks (since redundant, as 
is_generated_root() checks this internally anyway.

It also fixes the root partition situation. The idea of UsrOnly= was to not have a root partition, so that repart can create it at boot. The merged patch undid that, which defeated the whole point.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1669,7 +1669,7 @@ def configure_dracut(args: CommandLineArguments, root: str) -> None:
 
 
 def prepare_tree_root(args: CommandLineArguments, root: str) -> None:
-    if args.output_format == OutputFormat.subvolume:
+    if args.output_format == OutputFormat.subvolume and not is_generated_root(args):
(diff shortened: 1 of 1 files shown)


poettering:
> It also fixes the root partition situation. The idea of UsrOnly= was to not have a root partition, so that repart can create it at boot. The merged patch undid that, which defeated the whole point.

I mean, @257 apparently had a different idea what UsrOnly= was about? i.e. that both the root fs and /usr exist, but the root fs is empty initially? it might be worth supporting that, but it should be a new, seperate thing. i.e. UsrOnly=yes|no|strip. And if yes/no the behaviour would be as this commit restores, but if UsrOnly=strip would be set we'd delete everyting from the rootfs but /usr.

PR #706, opened 2021-04-16
mkosi: make workspace dir configurable and discover mkosi.workspace/

poettering:
So far we determined the workspace directory from $TMPDIR and /var/tmp
if that wasn't set. In some cases this is less than ideal though, in
particular if one wants to build a btrfs image but /var/tmp is not
btrfs. In that case it makes sense to always use a local workspace dir.
This adds support for this:

WorkspaceDirectory= + --workspace-dir= may now be used to configure it
explicitly.

if "mkosi.workspace/" exists, we use it automatically.

Otherwise we use TMPDIR and /var/tmp as fallback.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -264,6 +264,20 @@ details see the table below.
   such a directory does not exist, all output artifacts are placed
   adjacent to the output image file.
 
+`--workspace-dir=`
+
+: Path to a directory where to store data required temporarily while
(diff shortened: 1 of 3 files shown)


behrmann:
The new switch still needs to be added to the tests, that test will fail.

behrmann:
LGTM otherwise.

Issue #678, opened 2021-03-17
RFE: trim file systems after we are done, right before unmounting them


poettering:
This will punch holes into the backing loopback file systems, taking the air out. This is not too interesting in itself I think (though doesn't hurt), but it has the benefit that deleted files will be replaced by zeros, which should improve compression later on.

It's just a matter of calling `fstrim` on the mounted fs right before unmounting the final image

PR #707, opened 2021-04-16
mkosi: trim file system after finishing all writes

poettering:
This is not really about actually generated sparse images, but about the
side effect that unused blocks will appear as zero, and thus the images
will be able to compress better.

Fixes: #678

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3151,6 +3151,21 @@ def set_root_password(args: CommandLineArguments, root: str, do_run_build_script
             patch_file(os.path.join(root, "etc/shadow"), jj)
 
 
+def invoke_fstrim(args: CommandLineArguments, root: str, do_run_build_script: bool, for_cache: bool) -> None:
+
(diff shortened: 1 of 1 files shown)


behrmann:
LGTM.

mkosi/__init__.py
```diff
@@ -6417,6 +6432,7 @@ def build_image(
                 reset_machine_id(args, root, do_run_build_script, for_cache)
                 reset_random_seed(args, root)
                 run_finalize_script(args, root, do_run_build_script, for_cache)
+                invoke_fstrim(args, root, do_run_build_script, for_cache)
```
DaanDeMeyer, line 6435: Is this correct, you check if the image is cached in the function but you pass for_cache instead of cached as the argument. 

So currently, this only runs when building a cached image. Is that intended? If so, let's name the function argument for_cache instead of cached

poettering: fixed

PR #708, opened 2021-04-16
mkosi: fix extraction of partitions

poettering:
Add missing ","


behrmann (approved):
A nicely subtle bug when Python decides to be implicit instead of explicit...

PR #710, opened 2021-04-29
Fix: comma-delimited-list in docs examples for {,Build}Packages=

elloe:
`--package` and `Packages=` are processed by `CommaDelimitedListAction()` but examples for `{,Build}Packages=` show space separated lists (and do not show indented-line form although that is generally described when talking about `.ini` file format).

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -523,7 +523,7 @@ details see the table below.
 : Install the specified distribution packages (i.e. RPM, DEB, …) in
   the image. Takes a comma separated list of packages. This option may
   be used multiple times in which case the specified package list is
-  combined. Packaged specified this way will be installed both in the
+  combined. Packages specified this way will be installed both in the
   development and the final image (see below). Use `--build-package=`
   (see below) to specify packages that shall only be used for the
   image generated in the build image, but that shall not appear in the
@@ -1430,8 +1430,8 @@ Format=gpt_btrfs
 Bootable=yes
 
 [Packages]
-Packages=openssh-clients httpd
-BuildPackages=make gcc libcurl-devel
+Packages=openssh-clients,httpd
+BuildPackages=make,gcc,libcurl-devel
 EOF
 # cat > mkosi.build <<EOF
 #!/bin/sh


man/mkosi.1
```diff
@@ -416,7 +416,7 @@ the image.
 Takes a comma separated list of packages.
 This option may be used multiple times in which case the specified
 package list is combined.
-Packaged specified this way will be installed both in the development
+Packages specified this way will be installed both in the development
```
behrmann, line 419: This needs to fixed in `mkosi.md` instead of here, since this file is generated from it.

behrmann (changes requested):

behrmann:
@DaanDeMeyer: The message about 
> First-time contributors need a maintainer to approve running workflows.

is new. Not that this PR really needs everything to run, but this could be a bit cumbersome down the road.

mrc0mmand:
> @DaanDeMeyer: The message about
> 
> > First-time contributors need a maintainer to approve running workflows.
> 
> is new. Not that this PR really needs everything to run, but this could be a bit cumbersome down the road.

That's a new GH feature to combat crypto miners (https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/#new-features-to-help-protect-maintainers) - the approval is required only for the first contribution of each contributor, so hopefully it won't be *that* disruptive.

behrmann:
Damn crypto miners, ruining it for everybody. :man_facepalming:

DaanDeMeyer:
Thanks for the fix!

poettering:
Hmm, we accept the list in both ways though, no? i.e. whitespace *and* comma separated, no? might be worth fixing the docs for? /me is confused

behrmann:
I haven't thought about this deeply, but I think [this line](https://github.com/systemd/mkosi/blob/63bb21bab5cd3ecc7f306d2dffc5c82ea20cbf04/mkosi/__init__.py#L4720) in `ArgumentParserMkosi._read_args_from_files` only makes newlines into whitespace that is accepted for these comma-separated lists.

elloe:
We were using the documented examples with "Packages=a b c" and found, for Debian/Ubuntu builds, that the nspawn session to install these extra packages is doing "apt-get install ... 'a b c' " - note the single quotes, and that causes an error which, from memory, read something like: "Unable to find package: a b c" since the single quotes on the command-line causes 'a b c' to be a single argument, not a space separated list.

PR #715, opened 2021-04-30
RFC: Move distributions into their own submodules

behrmann:
This is a RFC and should **not** be merged. 

This is a terribly big and ugly PR, although for the most part it is only moving stuff around, and I'm only opening it up to see what breaks in CI and as a RFC whether this is a direction that should be taken. I haven't addressed the CLI tests yet and right now I can't even get it to typecheck my lazy dictionary (more on that later), although the rest typechecked just fine before I added that.

What is new? This adds a class `DistributionInstaller` that tries to bundle everything needed to install a certain distribution. Every supported distribution inherits from this (or its parent distro) to specify how it is to be installed. This mostly get's rid of the `Distribution` enum, but a few holdovers in `mkosi.__init__` remain, changed for subclass checks. This is ugly and needs to go, so I'm happy to hear ideas on how to do that. Also, currently `CommandLineArguments` and `DistributionInstaller` reference each other in a cycle, which too is supremely ugly and something I'd like to fix, but was a needed step to pick the two apart and I think I can now grep for all the things that could move over from `CommandLineArguments` to `DistributionInstaller` .

To break circular imports `CommandLineArguments` and `DistributionInstaller` live in a new module `backend.py`. All they functions they need are moved there as well for the same reason. To forestall further circular imports the main mkosi module 
accesses the distribution installers via a lazy object, look for `SUPPORTED_DISTRIBUTIONS` in `backend.py`.

If this is something we want to do and once I squash all the CI failures, I could try to pick this apart into a more digestable series of commits.


DaanDeMeyer:
We definitely need to split things up but this seems to be doing a bit too much all in one go. Can we do this a bit more gradual? Let's start pulling out the low-level basic library stuff into separate files (which we then re-import back into the main file). Once that's done, re-evaluate the next thing in the main file that can be moved out. Keep applying this process until all the stuff the distros depend on is moved out into separate files, and then start moving each of the distro functions into separate files.

I'd definitely separate the moving to different files from any other changes like the introduction of `DistributionInstaller`. Doing it gradual like this makes it a bit more tenable for me to review as well.

behrmann:
I never intended this to be merged as is. :) Haven't managed to have another go at this, since all the DNF distros are broken currently. Once I fix that and I know this can work, I'll start over and split out the groundwork, but I think afterwards it's still going to be a mostly uncomfortably big diff.

poettering:
I'd *love* if the huge file would be split up. The only reason I never did this myself, is that I am not a Python-guru enough myself to know what the right approach for splitting up projects actually is....

behrmann:
I'll close this now, since this thing is not rebaseable, but I split out the first step into a separate PR since I had a couple of minutes downtime and we can go from there.

257:
> I'd _love_ if the huge file would be split up. The only reason I never did this myself, is that I am not a Python-guru enough myself to know what the right approach for splitting up projects actually is....

my _un-educated_ 2-cents -- since i didn't get to read @behrmann's work -- is that chopping up by package manager would coagulate distros more organically, by the way of family resemblance if you like. ```yml``` (```dnf``` too?) already have their own invoke function as does ```portage```, that is once the gentoo support is merged.

i'm null when it comes to pythonic-foo so i leave that to the masters (and go hiding but listening).

Issue #718, opened 2021-05-14
Support foreign architectures (e.g. aarch64 on amd64)


elloe:
Issue ##138 added the `--architecture` option, but it is limited to architectures compatible with the host (e.g. `x86` on `x86_64`). That issue originally asked for `armhf` support.

We're working on a project that harnesses `systemd-nspawn` and `mkosi` creating base template images for `amd64` and `arm64` Debian architectures. Containers will run on native host architectures but whilst we're building and refining the images it is much faster working on powerful `amd64` hosts when building and testing the `arm64 (aarch64)` images.

With the support of `qemu-user-static` and its `qemu-aarch64-static` we can successfully build and run Debian `chroot`s using `debootstrap --arch=arm64 --foreign ...` followed by `chroot target/ /debootstrap/debootstrap --second-stage`, or more conveniently by simply calling the wrapper `qemu-debootstrap --arch=arm64 ...`

We're investigating adding support for this into `mkosi` locally but consider it would be a great addition to mkosi's capabilities and so are seeking feedback on any wider considerations we should take into account before submitting a PR.

Most obviously would be supporting it across distro's and there we'd need input and help from `Fedora` and `dnf` experts since we're a Debian/Ubuntu site.

elloe:
This appears to work perfectly. We're using Ubuntu 20.04 `amd64` hosts and creating Debian `bullseye (testing)` containers as `btrfs subvolume`s.

There's an unrelated issue apparently in `systemd-nspawn` where `machinectl stop <name>` doesn't cause a shutdown and the container ends up in `emergency` mode.

poettering:
> Do we need to detect and confirm the presence of `qemu-user-static` and `qemu-debootstrap` on the host and bail out reporting it missing before allowing installation to proceed?

right now, if some tool we need is missing we'll just fail invoking it, which isn't ideal, but also not a loss, since this is developer-focussed stuff and the error yous see pretty discriptive.

Also, if this works for Debian only initially, that's fine. If others are interested for other distros we can extend this later. Hence please submit the change as PR

DaanDeMeyer:
This should work as is these days if the version of systemd-nspawn is recent enough (v250 or later) so let's close this. If there are any remaining issues, please open a new issue.

PR #719, opened 2021-05-14
Support foreign architecture containers

elloe:
Initial support for building foreign architecture containers as
subvolumes or tar. This patch enables building, for example,
aarch64 on amd64 for Debian/Ubuntu.

Not (yet) tested for building bootable images or other distributions.

Initial work for Issue #718

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -532,6 +532,12 @@ FEDORA_KEYS_MAP = {
 GPT_HEADER_SIZE = 1024 * 1024
 GPT_FOOTER_SIZE = 1024 * 1024
 
+# for alternate architectures we need to know which are compatible with host
+# and which are 'foreign' (e.g. aarch64 on x86_64 is foreign)
+ARCHITECTURES_COMPAT = {
+    "x86" : "x86_64",
+    "armhfp" : "aarch64",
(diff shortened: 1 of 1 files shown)


behrmann:
That's a cool change! Thanks for working on this.

Not withstanding the trivial change needed (formatting with black), the other CI failures need to be fixed (they're extremely unlikely to be on your side, though). The other things missing would be throwing an error in `load_args` if using this outside of Debian/Ubuntu is attempted and some docs would need to be added.

mkosi/__init__.py
```diff
@@ -532,6 +532,12 @@ class CommandLineArguments:
 GPT_HEADER_SIZE = 1024 * 1024
 GPT_FOOTER_SIZE = 1024 * 1024
 
+# for alternate architectures we need to know which are compatible with host
+# and which are 'foreign' (e.g. aarch64 on x86_64 is foreign)
+ARCHITECTURES_COMPAT = {
```
poettering, line 537: not a fan of needless abbreviations. Why not name this COMPATIBLE_ARCHITECTURES or so?

DaanDeMeyer:
Closing due to lack of response, we can open again later

PR #720, opened 2021-05-21
Some refactorings and bugfixes

keszybz:


mkosi/__init__.py
```diff
@@ -168,6 +170,28 @@ class MkosiException(Exception):
...
+
+@dictify
+def read_os_release():
+    try:
+        f = open('/etc/os-release')
+    except FileNotFoundError:
+        f = open('usr/lib/os-release')
+
+    for line_number, line in enumerate(f):
+        m = re.match(r'([A-Z][A-Z_0-9]+)=(.*?)\s*$', line)
```
behrmann, line 186: This regex is a bit too strict. When I have a comment or empty line it will print that this line is bad, although the `os-release` man page says that's allowed.

mkosi/__init__.py
```diff
@@ -168,6 +170,28 @@ class MkosiException(Exception):
...
+    def wrapper(*args, **kwargs):
+        return dict(f(*args, **kwargs))
+    return functools.update_wrapper(wrapper, f)
+
+@dictify
+def read_os_release():
+    try:
+        f = open('/etc/os-release')
+    except FileNotFoundError:
+        f = open('usr/lib/os-release')
```
behrmann, line 183: I don't really see that as a grave problem, but this is less robust than before, since it doesn't catch a `FileNotFoundError` if both files don't exist. Previously the defaults would have been used.

behrmann (commented):
LGTM. Nothing too controversial. The failing tests are missing black formatting and later on (I guess) missing type annotations for `dictify` and `read_os_release`.

I think it would be easier to read without `dictify` and `read_os_release` directly returning a dict instead of yielding, but I guess that's a matter of taste.

mkosi/__init__.py
keszybz: Of course. I forgot about comments because I mostly tested on real os-release files, but it's still a stupid omission on my part.

mkosi/__init__.py
keszybz: That's somewhat on purpose: os-release is supposed to always exists (and it's going to be there on any system released in the last 8 years...)
So I though that this function should return an error. The error is still caught in detect_distribution(), so effectively you still get `None, None` as the answer.

keszybz:
> I think it would be easier to read without `dictify` and `read_os_release` directly returning a dict instead of yielding, but I guess that's a matter of taste.

I made the decorator function because I hope it'll be useful later in other contexts. I think the pattern of not having to worry at about dictionary creation and `return` is nice. With one user it's a bit silly, but with two users, there should be less lines of code overall.

keszybz:
> The failing tests are missing black formatting and later on (I guess) missing type annotations for `dictify` and `read_os_release`.

Yeah, I couldn't figure out how to do the annotations for a decorator. I even have the docs for `typing` open, but it's a bit of a high bar.

behrmann:
> Yeah, I couldn't figure out how to do the annotations for a decorator. I even have the docs for `typing` open, but it's a bit of a high bar.

I struggled with that as well, when I read the code yesterday :)

This might be a tad to confining, but it does the trick
```python
def dictify(
        f: Callable[..., Generator[Tuple[str, str], None, None]]
) -> Callable[..., Dict[str, str]]:
    pass

def read_os_release() -> Generator[Tuple[str, str], None, None]:
    pass
```

keszybz:
Yeah, it uses concrete types. It would be nice to express this using generics…

But even with concrete types:
```py
def dictify(
        f: Callable[..., Generator[Tuple[str, str], None, None]]
) -> Callable[..., Dict[str, str]]:
    def wrapper(*args, **kwargs) -> Dict[str, str]:
        return dict(f(*args, **kwargs))
    return functools.update_wrapper(wrapper, f)
```
gives me:
```
mkosi/__init__.py:176:5: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]
        def wrapper(*args, **kwargs) -> Dict[str, str]:
        ^
mkosi/__init__.py:716:9: error: Cannot infer type argument 1 of "copyfileobj"  [misc]
            shutil.copyfileobj(oldobject, newobject)
            ^
Found 2 errors in 1 file (checked 1 source file)
```

behrmann:
Sorry, forgot the the inner function above
```python
def dictify(
        f: Callable[..., Generator[Tuple[str, str], None, None]]
) -> Callable[..., Dict[str, str]]:
    def wrapper(*args: Any, **kwargs: Any) -> Dict[str, str]:
        return dict(f(*args, **kwargs))
    return functools.update_wrapper(wrapper, f)
```

I'm running 
```
$ mypy --version
mypy 0.812
$ mypy mkosi
Success: no issues found in 3 source files
```

keszybz:
Thanks, that works.

I added some more commits for other minor issues. I hope that's ok.

keszybz:
BTW, I'm still getting the error with `Cannot infer type argument 1 of "copyfileobj"`, but I have the same mypy version:
python3-mypy-0.812-1.fc34.noarch
python3-3.9.5-2.fc34.x86_64

man/mkosi.1
```diff
@@ -276,7 +276,7 @@ or the command line arguments passes \[lq]console=ttyS0 rw\[rq] to the
 kernel in any case.
 Just adding \[lq]console=ttyS0 rw\[rq] would append these two arguments
 to the kernel command line created by lower priority configuration files
-or previous \[en]kernel-command-line command line arguments.
```
behrmann, line 279: That file is autogenerated from `mkosi.md`. The fix is rather to put the bare switches in the text in backquotes.

mkosi/__init__.py
```diff
@@ -168,6 +170,36 @@ class MkosiException(Exception):
...
+@dictify
+def read_os_release() -> Generator[Tuple[str, str], None, None]:
+    try:
+        filename = '/etc/os-release'
+        f = open(filename)
+    except FileNotFoundError:
+        filename = '/usr/lib/os-release'
+        f = open(filename)
+
+    for line_number, line in enumerate(f, start=1):
```
behrmann, line 189: Nice, TIL! I never even considered `enumerate` might have a `start=` keyword.

mkosi/__init__.py
```diff
@@ -286,13 +318,12 @@ def __str__(self) -> str:
         return self.name
 
     @classmethod
-    def from_string(cls, name: str) -> "OutputFormat":
+    def Format(cls, name: str) -> "OutputFormat":
```
behrmann, line 321: The uppercase method name is a bit unusual, since uppercase things are usually classes. How about `format_from_string` or `get_format` or just keeping it `from_string`?

mkosi/__init__.py
```diff
@@ -4666,6 +4696,15 @@ def __init__(self, *kargs: Any, **kwargs: Any) -> None:
...
         super().__init__(*kargs, **kwargs)
 
+    @staticmethod
+    def _camel_to_arg(camel: str) -> str:
+        s1 = re.sub("(.)([A-Z][a-z]+)", r"\1-\2", camel)
+        return re.sub("([a-z0-9])([A-Z])", r"\1-\2", s1).lower()
+
+    @classmethod
+    def _ini_key_to_cli_arg(cls, key: str) -> str:
+        return cls.SPECIAL_MKOSI_DEFAULT_PARAMS.get(key, '') or ('--' + cls._camel_to_arg(key))
```
behrmann, line 4706:
```suggestion
        return cls.SPECIAL_MKOSI_DEFAULT_PARAMS.get(key) or ('--' + cls._camel_to_arg(key))
```
Setting the default of `get` to an empty string should be unnecessary, since `None` is falsey and will trigger the other branch of the `or`

behrmann (commented):
> BTW, I'm still getting the error with Cannot infer type argument 1 of "copyfileobj", but I have the same mypy version:
python3-mypy-0.812-1.fc34.noarch
python3-3.9.5-2.fc34.x86_64

That's curious, since I don't even think you are touching that line. Can you run `black` on your PR? Then we can see whether the issue is general (when it gets to the mypy check) or lies elsewhere.

man/mkosi.1
keszybz: And I was wondering why anyone would bother writing  a page in raw roff. I'll replace that patch with a patch for the markdown file and regenerate the man page.

mkosi/__init__.py
keszybz: Please see the commit message: "Format" is the same as the option in the config file, so we get a nice message:
```
    mkosi: error: argument -t/--format: invalid Format value: 'cpioa'
```

"invalid from_string/format_from_string/get_format" are not nearly as nice...

mkosi/__init__.py
keszybz: So it is. I was convinced that that'd raise an exception. Duh.

keszybz: `mypy` was unhappy with `.get(key, None)`, so I changed that to `.get(key, '')`. But `.get(key)` is obviously nicer, and `mypy` likes it too.

mkosi/__init__.py
behrmann: Ah, I didn't get that. I see the point. `format` in lowercase could make this look, like it had a (stringlike) format method, but otherwise I don't think it would clash with anything.

keszybz:
> That's curious, since I don't even think you are touching that line. Can you run `black` on your PR? Then we can see whether the issue is general (when it gets to the mypy check) or lies elsewhere.

I find the style grating. In particular the double-quoted strings are not nice and not being able to deviate from the style where it makes sense is imho a net loss. But OK.

keszybz:
I dropped the commit `mkosi: fix inverted condition for rpm metadata removal` from the pull request. I found another issue in that area, and I'll make a separate pull request.

behrmann:
> I find the style grating. In particular the double-quoted strings are not nice and not being able to deviate from the style where it makes sense is imho a net loss. But OK.

I can see that, it takes some getting used to, but I've grown quite fond of it, mostly because it makes diffs smaller. If necessary one can switch it off and we do use that already for the partitition UUIDs for discoverable partitions spec. :)

mkosi/__init__.py
```diff
@@ -318,13 +318,12 @@ def __str__(self) -> str:
...
     @classmethod
-    def from_string(cls, name: str) -> "OutputFormat":
+    def Format(cls, name: str) -> "OutputFormat":
         """A convenience method to be used with argparse"""
         try:
             return cls[name]
         except KeyError:
-            # this let's argparse generate a proper error message
-            return name  # type: ignore
+            raise ValueError
```
DaanDeMeyer, line 326: Shouldn't this be `raise ValueError()` so we raise an instance of `ValueError` instead of the type itself?

mkosi/__init__.py
DaanDeMeyer: I'd prefer we stick to the lowercase naming style here. The message is nice but it wouldn't be consistent across all options so if we do this I'd prefer it to be all or nothing.

mkosi/__init__.py
```diff
@@ -168,6 +170,36 @@ class MkosiException(Exception):
     """Leads to sys.exit"""
 
 
+def dictify(f: Callable[..., Generator[Tuple[str, str], None, None]]) -> Callable[..., Dict[str, str]]:
+    def wrapper(*args: Any, **kwargs: Any) -> Dict[str, str]:
+        return dict(f(*args, **kwargs))
+
+    return functools.update_wrapper(wrapper, f)
```
DaanDeMeyer, line 177: I believe the preferred way of doing this is to use `functools.wraps` (https://docs.python.org/3/library/functools.html#functools.wraps)

mkosi/__init__.py
```diff
@@ -168,6 +170,36 @@ class MkosiException(Exception):
...
+
+
+@dictify
```
DaanDeMeyer, line 180: Maybe we can just use a `dict` in `read_os_release` to store the results and return that? I think that ends up being a simpler solution overall.

mkosi/__init__.py
keszybz: That's the same. There's special support for `raise <type>`.

mkosi/__init__.py
keszybz: It's equivalent. `functools.wraps` is a decorator and is used when the wrapped function is longer. For a short one like this, `update_wrapper` is fine.

mkosi/__init__.py
keszybz: I prefer to keep the *interesting* part, i.e. `read_os_release()` as simple and clean as possible. `dictify` is something that you don't need to look at. As mentioned before, I hope to be able to reuse it in other places later.

keszybz:
> I'd prefer we stick to the lowercase naming style here. The message is nice but it wouldn't be consistent across all options so if we do this I'd prefer it to be all or nothing.

It turns out it was the wrong solution all the time ;]. `raise argparse.ArgumentTypeError` does the job.

mkosi/__init__.py
behrmann: I think one can argue, whether moving the dictionary handling outside makes this simpler ;)
```python
def read_os_release() -> Dict[str, str]:
    osrelease = {}
    try:
        filename = "/etc/os-release"
        f = open(filename)
    except FileNotFoundError:
        filename = "/usr/lib/os-release"
        f = open(filename)

    for line_number, line in enumerate(f, start=1):
        line = line.rstrip()
        if not line or line.startswith("#"):
            continue
        m = re.match(r"([A-Z][A-Z_0-9]+)=(.*)", line)
        if m:
            name, val = m.groups()
            if val and val[0] in "\"'":
                val = ast.literal_eval(val)
            osrelease[name] = val
        else:
            print(f"{filename}:{line_number}: bad line {line!r}", file=sys.stderr)
    
    return osrelease
```
But I get you point and if it finds other users I don't see an issue with it.

PR #722, opened 2021-05-23
mkosi: fix inverted conditions for package metadata removal

keszybz:
The same (wrong) pattern was used in a few places. And I think 'not any(…)' is
more natural in this can than 'all(not …)'.


behrmann (approved):
LGTM

PR #723, opened 2021-05-24
mkosi: allow inline comments (but only with "#")

keszybz:
For example, it's super useful to comment on a list of packages:

Packages=
        glibc-minimal-langpack    # avoid pulling in other glibc langpacks
        less                      # this makes 'systemctl' much nicer to use ;)

;-comments are not allowed, because the colon doesn't stand out
enough. I was considering also adding comment_prefixes=('#',) to
disallow ;-comments everywhere, but that'd be a compatibility break
which I don't see enough justification for.


behrmann:
That looks like a good addition.

Issue #677, opened 2021-03-17
RFE: AlpineLinux support


Edu4rdSHL:
Hello, would be awesome to have Alpine Linux support in `mkosi`, they're well-known for being lightweight and other things.

After taking a look, they've an [alpine-make-rootfs](https://github.com/alpinelinux/alpine-make-rootfs) script that is able to automatically package the rootfs in several formats, for example to get a minimal TAR container ready to import with `machinectl` we just need to use:

```
┌─[✗]─[ SecHackLabs /tmp/tmp.L1y954VIV3/apk-tools-2.10.4 ]
└──╼ $ sudo ./alpine-make-rootfs alpine.tar.gz

> Installing system
fetch https://nl.alpinelinux.org/alpine/latest-stable/main/x86_64/APKINDEX.tar.gz
fetch https://nl.alpinelinux.org/alpine/latest-stable/community/x86_64/APKINDEX.tar.gz
(1/6) Installing musl (1.2.2-r0)
(2/6) Installing busybox (1.32.1-r3)
Executing busybox-1.32.1-r3.post-install
(3/6) Installing alpine-baselayout (3.2.0-r8)
Executing alpine-baselayout-3.2.0-r8.pre-install
var/cache/misc/alpine-baselayout-3.2.0-r8.pre-install: line 3: can't create /dev/null: Permission denied
Executing alpine-baselayout-3.2.0-r8.post-install
(4/6) Installing busybox-suid (1.32.1-r3)
(5/6) Installing scanelf (1.2.8-r0)
(6/6) Installing musl-utils (1.2.2-r0)
Executing busybox-1.32.1-r3.trigger
OK: 2 MiB in 6 packages
tar: Ignoring unknown extended header keyword 'APK-TOOLS.checksum.SHA1'
tar: Ignoring unknown extended header keyword 'APK-TOOLS.checksum.SHA1'
tar: Ignoring unknown extended header keyword 'APK-TOOLS.checksum.SHA1'

> Cleaning-up rootfs

> Creating rootfs archive
-rw-r--r-- 1 edu4rdshl wheel 970748 Mar 16 22:10 alpine.tar.gz
┌─[✗]─[ SecHackLabs /tmp/tmp.L1y954VIV3/apk-tools-2.10.4 ]
└──╼ $ sudo machinectl import-tar ./alpine.tar.gz alpinelinux
Enqueued transfer job 1. Press C-c to continue download in background.
[...]

PR #724, opened 2021-05-24
mkosi: stop automatically inserting packages

keszybz:
systemd, systemd-udev and some other packages were always automatically
inserted since the first version of mkosi. But I think this is the
wrong approach: packages to install should mostly be specified in the
configuration file, and not in mkosi sources.

Looking at some concrete examples:
- dracut: dracut is only necessary if using an initrd (and of
  course only if the initrd image is generated using dracut, and only
  if it is generated locally). But even for bootable images, a kernel
  with no initramfs might be enough. Or when using kvm, a direct
  kernel boot with an externally-provided kernel+initrd, etc.
  So pulling in dracut and dependencies is often not necessary.
- kernel-modules: despite the name, actually basic modules are in
  kernel-core. And kernel-modules is non-core modules. (And yes,
  kernel-modules-extra is extra-non-core-modules, and
  kernel-modules-internal is something else yet again).
- kernel-core: as with dracut, having a kernel image is not necessary
  in many scenarios.
- systemd: one big example where it's not useful is systemd itself.
  When using mkosi to build a systemd test image from sources, installing
  the big system image forces all deps to be installed, which might
  not be desired. Dropping systemd from the installation list makes
  mkosi useful for non-systemd stuff (either containers or non-systemd
  inits).
- grub* / systemd-boot: people might want to use grub even if sd-boot
  is available, etc.

To make this change more palatable, I kept various packages in
conditional mode: e.g. systemd-udev is probably necessary *if* installing
systemd and making a bootable image. This way we make more things
work automatically out box (the user only needs to specify 'systemd',
not the subpackages), but allow the unnecessary deps to be cut down.

It's possible that even those conditional deps are too much and should
be removed from here.

Also, I don't know non-rpm packaing with enough detail, so I didn't
touch the non-rpm parts, though they probably should get similar treatment.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -518,6 +518,27 @@ details see the table below.
   cached images, combined `-i` with `-ff`, which ensures the cached
   images are removed first, and then re-created.
 
+`--base-packages`
+
+: If true, automatically install packages to ensure basic
+  functionality, as appropriate for the given image type. For example,
+  `systemd` is always included, `systemd-udev` and `dracut` if the
+  image is bootable, and so on.
+
+: If false, only packages specified with `--package`/`Packages` will
+  be installed.
+
+: If `conditional`, the list of packages to install will be extended
+  with boolean dependencies
+  (c.f. https://rpm.org/user_doc/boolean_dependencies.html), to
+  install specific packages when *other* packages are in the list. For
+  example, `systemd-udev` may be automatically included if the image
+  is bootable and `systemd` is installed. With this, various "base"
+  packages still need to be specified if they should be included, but
+  the corresponding "extension" packages will be added automatically
+  when appropriate. This feature depends on support in the package
+  manager, so it is not implemented for all distributions.
+
 `--package=`, `-p`
 
 : Install the specified distribution packages (i.e. RPM, DEB, …) in
@@ -528,7 +549,8 @@ details see the table below.
   (see below) to specify packages that shall only be used for the
   image generated in the build image, but that shall not appear in the
   final image.
-  To remove a package e.g. added by a mkosi.default configuration file
+
+: To remove a package e.g. added by a mkosi.default configuration file
   prepend the package name with a ! letter. For example -p "!apache2"
   would remove the apache2 package. To replace the apache2 package by
   the httpd package just add -p "!apache2,httpd" to the command line
@@ -961,6 +983,7 @@ which settings file options.
 | `--hostonly-initrd`               | `[Output]`              | `HostonlyInitrd=`             |
 | `--usr-only`                      | `[Output]`              | `UsrOnly=`                    |
 | `--split-artifacts`               | `[Output]`              | `SplitArtifacts=`             |
+| `--base-packages=`                | `[Packages]`            | `BasePackages=`               |
 | `--package=`                      | `[Packages]`            | `Packages=`                   |
 | `--with-docs`                     | `[Packages]`            | `WithDocs=`                   |
 | `--without-tests`, `-T`           | `[Packages]`            | `WithTests=`                  |
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2689,6 +2689,16 @@ def install_ubuntu(args: CommandLineArguments, root: str, do_run_build_script: b
     install_debian_or_ubuntu(args, root, do_run_build_script=do_run_build_script)
 
 
+def run_pacman(root: str, pacman_conf: str, packages: Set[str]) -> None:
+    try:
+        run(["pacman-key", "--config", pacman_conf, "--init"])
+        run(["pacman-key", "--config", pacman_conf, "--populate"])
+        run(["pacman", "--config", pacman_conf, "--noconfirm", "-Sy", *sort_packages(packages)])
+    finally:
+        # Kill the gpg-agent started by pacman and pacman-key.
+        run(["gpgconf", "--homedir", os.path.join(root, "etc/pacman.d/gnupg"), "--kill", "all"])
+
+
 @complete_step("Installing Arch Linux")
 def install_arch(args: CommandLineArguments, root: str, do_run_build_script: bool) -> None:
     if args.release is not None:
@@ -2996,19 +3006,8 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
     if not do_run_build_script and args.ssh:
         add_packages(args, packages, "openssh")
 
-    def run_pacman(packages: Set[str]) -> None:
-        conf = ["--config", pacman_conf]
-
-        try:
-            run(["pacman-key", *conf, "--init"])
-            run(["pacman-key", *conf, "--populate"])
-            run(["pacman", *conf, "--noconfirm", "-Sy", *sort_packages(packages)])
-        finally:
-            # Kill the gpg-agent started by pacman and pacman-key.
-            run(["gpgconf", "--homedir", os.path.join(root, "etc/pacman.d/gnupg"), "--kill", "all"])
-
     with mount_api_vfs(args, root):
-        run_pacman(packages)
+        run_pacman(root, pacman_conf, packages)
 
     # If /etc/locale.gen exists, uncomment the desired locale and leave the rest of the file untouched.
     # If it doesn’t exist, just write the desired locale in it.
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1919,6 +1919,14 @@ def add_packages(
             packages.add(f"({name} if {conditional})" if conditional else name)
 
 
+def sort_packages(packages: Set[str]) -> List[str]:
+    """Sorts packages: normal first, paths second, conditional third"""
+
+    m = {"(": 2, "/": 1}
+    sort = lambda name: (m.get(name[0], 0), name)
+    return sorted(packages, key=sort)
+
+
 def make_rpm_list(args: CommandLineArguments, packages: Set[str], do_run_build_script: bool) -> Set[str]:
     packages = packages.copy()
 
@@ -2057,7 +2065,7 @@ def invoke_dnf(
     if not args.with_docs:
         cmdline += ["--nodocs"]
 
-    cmdline += ["install", *packages]
+    cmdline += ["install", *sort_packages(packages)]
 
     with mount_api_vfs(args, root):
         run(cmdline)
@@ -2088,7 +2096,7 @@ def invoke_tdnf(
     if not gpgcheck:
         cmdline.append("--nogpgcheck")
 
-    cmdline += ["install", *packages]
+    cmdline += ["install", *sort_packages(packages)]
 
     with mount_api_vfs(args, root):
         run(cmdline)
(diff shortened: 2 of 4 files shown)


behrmann:
This is an unexpected change :)

I do see the possible container usage and supporting non-systemd inits would make some people happy (looking at #677). I do see the possible problem, though, that this could invalidate previous config files on update to a newer mkosi, though. 

Maybe we can have our cake and eat it too, and do something like package profiles, e.g. if we were to do something like in #715, the base class of a distribution installer would have a minimal set of packages and others could pile on top of that.

keszybz:
Yep, it's a change in direction… Full disclosure: I'm working on generating initrds using mkosi. Those will still be based on systemd, so for me the intent is not to remove systemd, but to minimize the installation image. And to do that I want to be able to hand-select the few packages that are neccessary, and prune everything else.

> the base class of a distribution installer would have a minimal set of packages and others could pile on top of that.

I think that even a "minimal set" is not necessary. With Fedora, depending on the use case, there is no fixed set that is always present. For my initrd case, I install `systemd`, `systemd-udev`, `less`, `util-linux`, `coreutils`, `glibc`. But for a container I might use `systemd-standalone-sysusers` + `fedora-release` + `glibc` + something-rust-or-go-based-with-no-further-deps. So only `glibc` would be common, but it's much better to pull it in through dependencies.

behrmann:
> Full disclosure: I'm working on generating initrds using mkosi.

That's sounds like a cool project!

> I think that even a "minimal set" is not necessary.

Well, the set might turn well out to be empty. :) The point I was trying to make, was rather that I think there is also a value in having current configs to keep working the way they currently do on the one hand and to be able to spin up images for other distros (if one has the package manager toolchain...) without having to think about which packages need to be included for e.g. a basic booting system. I definitely agree, that there is value in a smaller package footprint to allow for more use cases. 

About the CI failures: I assume the packages included in those CI steps need to be amended.

keszybz:
OK. So should this be handled though some config option, BasePackages=`yes|no|conditional` ?

FWIW, I think the CI case is not representative. Right now the CI checks that we can boot in various distros with very minimal configuration. For this use case it *is* useful to not have to specify any bootloader or kernel-module packages, and just get everything pulled in on all distros automatically. But I doubt that this is what most people use `mkosi` for. At least in my case, I pretty much always use the same distro, and only sometimes different versions of the same distro. And in this case, I know what packages I need and I'm completely fine with having no default package set.

DaanDeMeyer:
The idea behind the current setup is to have the first user experience "just work". You generate an image and boot it and there's no caveats like "you need to add this package to make it work".

We also try to keep backwards compatibility somewhat intact so changing the defaults might cause issues in that regard.

I do sympathize with your use case though. A `BasePackages` switch that allows switching off any default "extra" packages (including extra ones that may be added in the future) sounds good.

johannbg:
@keszybz what's your end game here? 
Will the dracut support be removed from mkosi or can people still opt in to use it instead of your initrd generation integrated directly into mkosi?

DaanDeMeyer:
> First, in general, I think mkosi is the wrong place to do complex integration of third-party components. So I would like to see less dracut-specific code. Right now mkosi has code to interface between dracut and kernel-install, both third party components. In particular, providing custom kernel-install plugin here to do unified kernel images seems wrong. Mkosi images are certainly not the only place where unified kernel images make sense, so this should be either moved to systemd upstream alongside other kernel-install plugins, or it maybe packaged separately. Similarly, mkosi has code to neuter some functionality in the kernel package to prevent it from generating a kernel by temporarily hiding some kernel-install files. The same issue would be faced on real systems, so it seems the kernel package should be adjusted to handle this better. I would like to see those parts dropped from mkosi after appropriate adjustments have been made in other places.

I wholeheartedly agree with this, if we can drop all this stuff from mkosi that would be all the better. Unfortunately the problem we tend to have to deal with is that we do kind of support centos and debian where it takes ages for new systemd versions to become available so even if we make fixes in systemd, we'd have to keep the mkosi cruft intact for a long time to not break older distros.

johannbg:
@keszybz mkosi by it's nature is bound to do complex integration of third party components otherwise all those package management solutions would never have been integrated and mkosi would just use it's own package management system right which brings us to your PR should cover changes for your tool in *all* supported distribution not just "Fedora".

Now the benefit of having mkosi stick to third party designs is that it does not have to worry about maintaining equal solutions in it's code base ( of course that benefit quickly disappears if the intent is support more than one such solutions, like alternative init system, alternative boot loaders, alternative init generation tools etc ) which in turn the code base can be kept small. 

The fact is regardless what the init generation tool is called, it's bound to evolve in the same direction as dracut has for it's decade+ of existence, to support all the deviations on the core base os layer between distributions and to make matters even more complex there are certain upstream that have become systemd only ( most recent that I'm aware of being lvm ) it's getting increasing more difficult for distribution to support multiple alternative init system and for other upstream components ( such as dracut ) to support distributions that do.

johannbg:
> > 
> 
> I wholeheartedly agree with this, if we can drop all this stuff from mkosi that would be all the better. Unfortunately the problem we tend to have to deal with is that we do kind of support centos and debian where it takes ages for new systemd versions to become available so even if we make fixes in systemd, we'd have to keep the mkosi cruft intact for a long time to not break older distros.

The utopia that @keszybz want's to achieve cannot be achieved unless existing distribution support is entirely dropped and mkosi just focuses on supporting it's `own` distribution, in which it has full control of all the components that make up that distro it's creating and which releases those component that make up that distribution are otherwise mkosi is bound to whatever supported downstream distributions are doing.

keszybz:
New and improved: BasePackages= is added, see the commit message for a lengthy description. Backwards-compat is now preserved, so this should be easier to merge ;)

First two commits are bugfixes which we might want to merge separately if the main patch needs more discussion.

The last two commits are cosmetic fixes.

mkosi/__init__.py
```diff
@@ -1902,31 +1903,54 @@ def reenable_kernel_install(args: CommandLineArguments, root: str) -> None:
     make_executable(hook_path)
 
 
+def add_packages(
+    args: CommandLineArguments, packages: Set[str], *names: str, conditional: Optional[str] = None
```
behrmann, line 1907:
```suggestion
    args: CommandLineArguments, packages: Set[str], *names: str, *, conditional: Optional[str] = None
```
to force the usage of the keyword. I think this makes functions with varargs a bit less confusing.

mkosi/__init__.py
```diff
@@ -1902,31 +1903,54 @@ def reenable_kernel_install(args: CommandLineArguments, root: str) -> None:
...
+
+    if args.base_packages or (conditional or args.base_packages is True):
+        for name in names:
+            packages.add(f"({name} if {conditional})" if conditional else name)
+
+
+def sort_packages(packages: Set[str]) -> List[str]:
+    """Sorts packages: normal first, paths second, conditional third"""
+
+    m = {"(": 2, "/": 1}
```
behrmann, line 1924: I think this only works for absolute paths, could this be an issue?

mkosi/__init__.py
```diff
@@ -1902,31 +1903,54 @@ def reenable_kernel_install(args: CommandLineArguments, root: str) -> None:
...
+    args: CommandLineArguments, packages: Set[str], *names: str, conditional: Optional[str] = None
+) -> None:
+    """Add packages in @names to @packages, if enabled by --base-packages.
+
+    If @conditional is specifed, rpm-specific syntax for boolean
+    dependencies will be used to include @names on if @conditional is.
+    """
+    assert args.base_packages is True or args.base_packages is False or args.base_packages == "conditional"
+
+    if args.base_packages or (conditional or args.base_packages is True):
```
behrmann, line 1916: This line seems weird to me. If `base_packages` is `True` or `"conditional"`, the first branch will trigger. If it's `False` the `is True` in the other branch can't trigger anymore.

behrmann (commented):
Haven't fully thought all the implications, but a first quick review.

mkosi/__init__.py
keszybz: At least for virtual deps, the path must be absolute, and this was the case I was thinking about.

I guess you can also specify an rpm by path, to install a local rpm instead of downloading it. Then it would be either sorted last (if an absolute path is given), and somewhere in the middle alphabetically if the path starts with letters. I think it's OK to ignore this corner case. The sort would still be stable.

mkosi/__init__.py
keszybz: I'm pretty sure that keyword usage is already forced by `*names`. There is no way to specify `conditional` without a keyword. The syntax with `, *, ` was added to support the case where there is no `*args` parameter.

mkosi/__init__.py
keszybz: Yep, will fix.

mkosi/__init__.py
behrmann: Of course, right, without the keyword it would always be gobbled up by `names`.

mkosi/__init__.py
keszybz: Thanks for catching this.  I reworked this as:
```python
if args.base_packages is True or (args.base_packages == 'conditional' and conditional):
    ...
```
I think this is better.

behrmann (commented):
Haven't yet looked where the built failures might come from, but the rest looks like a rather straightforward move to `add_packages`

mkosi.md
```diff
@@ -518,6 +518,27 @@ details see the table below.
   cached images, combined `-i` with `-ff`, which ensures the cached
   images are removed first, and then re-created.
 
+`--base-packages`
+
+: If true, automatically install packages to ensure basic
+  functionality, as appropriate for the given image type. For example,
+  `systemd` is always included, `systemd-udev` if the image is
```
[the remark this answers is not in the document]
keszybz: I added `dracut`. I think listing everything would be too complicated, since the package *names* are slightly different for each distro.

keszybz:
> Haven't yet looked where the built failures might come from, but the rest looks like a rather straightforward move to `add_packages`

I found a stupid editing mistake: some packages were dropped from the installation list. It should pass now.

keszybz:
arch/gpt_ext4: `Failed to dissect image '/home/runner/work/mkosi/mkosi/mkosi.output/image.raw': Connection timed out`. Not related.

DaanDeMeyer:
Looks super clean.

Still hoping someone can figure out the cause of the connection timeout :(

keszybz:
@mrc0mmand See above… Didn't you work on this issue over in systemd?

mrc0mmand:
> @mrc0mmand See above… Didn't you work on this issue over in systemd?

Yeah, it's https://github.com/systemd/systemd/issues/19442 but I still have no idea what's going on :(

mkosi.md
```diff
@@ -518,6 +518,27 @@ details see the table below.
   cached images, combined `-i` with `-ff`, which ensures the cached
   images are removed first, and then re-created.
 
+`--base-packages`
```
poettering, line 521: should have a "=" at the end, no? i.e. this takes a parameter...

mkosi.md
```diff
@@ -518,6 +518,27 @@ details see the table below.
...
+
+: If true, automatically install packages to ensure basic
+  functionality, as appropriate for the given image type. For example,
+  `systemd` is always included, `systemd-udev` and `dracut` if the
+  image is bootable, and so on.
+
+: If false, only packages specified with `--package`/`Packages` will
```
poettering, line 528: I'd suffix --package= and Packages= with a "=" here, too

PR #465, opened 2020-07-21
add support for Zstandard data compression algorithm

tpgxyz:
This PR adds support for Zstandard [1].
Tested with OpenMandriva:
* image creation works with --xz and --zstd options
* gpt_btrfs or gpt_squashfs file systems are compressed with ZSTD if `--compression zstd` is passed as mkosi command option

Yes i am aware of this https://github.com/systemd/systemd/issues/16222, but maybe adding support for Zstandard to mkosi will trigger some focus on adding corresponding support for machinectl/importd

[1] https://en.wikipedia.org/wiki/Zstandard

PR #728, opened 2021-06-04
cpio and zstd

keszybz:
This adds support for building cpio.zstd archives. See https://github.com/keszybz/mkosi-initrd for instructions how to create a functional pure-systemd initrd from Fedora packages.

I'm marking it as RFC. The cpio part is fine I think. The *implementation* of compression with zstd is also OK. But the user interface needs some design: right now we have `--xz` option, based on the assumption that `xz` is the only compression one would want to use. I think this isn't really true any more: `zstd` achieves very similar compression at a fraction of the compression/decompression times. For images that are widely distributed `xz` is still probably the best choice, but `zstd` is better for local work and anything where speed matters.

So I think we should deprecate the `--xz` option, and have `--inner-compression` (internally in the fs or archive where supported) and `--outer-compression` (for doing compression of the full output, like `--xz` right now), and `--compression` that would pick `--inner-compression` or `--outer-compression` as appropriate. Most of the time users would specify `--compression=yes` or `--compression=zstd|xz|gz|…` and `mkosi` would apply it in the appropriate place.

If you think that's the way to go, I will rework the pull request along those lines.

### DIFF
--- a/man/mkosi.1
+++ b/man/mkosi.1
@@ -47,7 +47,10 @@ btrfs subvolume, with separate subvolumes for \f[C]/var\f[R],
 \f[C]/home\f[R], \f[C]/srv\f[R], \f[C]/var/tmp\f[R]
 (\f[I]subvolume\f[R])
 .IP \[bu] 2
-Tarball (\f[I]tar\f[R])
+Tar archive (\f[I]tar\f[R])
+.IP \[bu] 2
+CPIO archive (\f[I]cpio\f[R]) in the format appropriate for a kernel
+initrd
 .PP
 When a \f[I]GPT\f[R] disk image is created, the following additional
 options are available:
@@ -527,7 +530,8 @@ removed first, and then re-created.
 If true, automatically install packages to ensure basic functionality,
 as appropriate for the given image type.
 For example, \f[C]systemd\f[R] is always included,
-\f[C]systemd-udev\f[R] if the image is bootable, and so on.
+\f[C]systemd-udev\f[R] and \f[C]dracut\f[R] if the image is bootable,
+and so on.
 If false, only packages specified with
 \f[C]--package\f[R]/\f[C]Packages\f[R] will be installed.
 If \f[C]conditional\f[R], the list of packages to install will be
--- a/man/mkosi.1
+++ b/man/mkosi.1
@@ -238,7 +238,8 @@ placed.
 If neither this option nor \f[C]--output-dir=\f[R] is used (see below),
 the image is generated under the name \f[C]image\f[R], but its name
 suffixed with an appropriate file suffix (e.g.\ \f[C]image.raw.xz\f[R]
-in case \f[C]gpt_ext4\f[R] is used in combination with \f[C]--xz\f[R]).
+in case \f[C]gpt_ext4\f[R] is used in combination with \f[C]xz\f[R]
+compression).
 If the \f[C]--image-id=\f[R] option is configured it is used instead of
 \f[C]image\f[R] in the default output name.
 If an image version is specified (via
@@ -392,32 +393,38 @@ against off-line modification, the verification data is placed in an
 additional GPT partition.
 Implies \f[C]--read-only\f[R].
 .TP
-\f[B]\f[CB]--compress=\f[B]\f[R]
-Compress the generated file systems.
-Only applies to \f[C]gpt_btrfs\f[R], \f[C]subvolume\f[R],
-\f[C]gpt_squashfs\f[R], \f[C]plain_squashfs\f[R].
+\f[B]\f[CB]--compress-fs=\f[B]\f[R]
+Enable or disable internal compression in the file system.
+Only applies to output formats with squashfs or btrfs.
 Takes one of \f[C]zlib\f[R], \f[C]lzo\f[R], \f[C]zstd\f[R],
 \f[C]lz4\f[R], \f[C]xz\f[R] or a boolean value as argument.
 If the latter is used compression is enabled/disabled and the default
 algorithm is used.
 In case of the \f[C]squashfs\f[R] output formats compression is implied,
-however this option may be used to select the algorithm.
+but this option may be used to select the algorithm.
+.TP
+\f[B]\f[CB]--compress-output=\f[B]\f[R]
+Configure compression for the resulting image or archive.
+The argument can be either a boolean or a compression algorithm
+(\f[C]xz\f[R], \f[C]zstd\f[R]).
+\f[C]xz\f[R] compression is used by default.
+Note that when applied to block device image types this means the image
+cannot be started directly but needs to be decompressed first.
+This also means that the \f[C]shell\f[R], \f[C]boot\f[R], \f[C]qemu\f[R]
+verbs are not available when this option is used.
+Implied for \f[C]tar\f[R] and \f[C]cpio\f[R].
+.TP
+\f[B]\f[CB]--compress=\f[B]\f[R]
+Enable compression.
+Using this option is equivalent to either \f[C]--compress-fs=\f[R] or
+\f[C]--compress-output=\f[R]; the appropriate type of compression is
+selected automatically.
 .TP
 \f[B]\f[CB]--mksquashfs=\f[B]\f[R]
 Set the path to the \f[C]mksquashfs\f[R] executable to use.
 This is useful in case the parameters for the tool shall be augmented,
 as the tool may be replaced by a script invoking it with the right
 parameters, this way.
-.TP
-\f[B]\f[CB]--xz\f[B]\f[R]
-Compress the resulting image with \f[C]xz\f[R].
-This only applies to \f[C]gpt_ext4\f[R], \f[C]gpt_xfs\f[R],
-\f[C]gpt_btrfs\f[R], \f[C]gpt_squashfs\f[R] and is implied for
-\f[C]tar\f[R].
-Note that when applied to the block device image types this means the
-image cannot be started directly but needs to be decompressed first.
-This also means that the \f[C]shell\f[R], \f[C]boot\f[R], \f[C]qemu\f[R]
-verbs are not available when this option is used.
 .TP
 \f[B]\f[CB]--qcow2\f[B]\f[R]
 Encode the resulting image as QEMU QCOW2 image.
@@ -1149,18 +1156,25 @@ T}@T{
 \f[C]Compress=\f[R]
 T}
 T{
+\f[C]--compress-fs=\f[R]
+T}@T{
+\f[C][Output]\f[R]
+T}@T{
+\f[C]CompressFs=\f[R]
+T}
+T{
+\f[C]--compress-output=\f[R]
+T}@T{
+\f[C][Output]\f[R]
+T}@T{
+\f[C]CompressOutput=\f[R]
+T}
+T{
 \f[C]--mksquashfs=\f[R]
 T}@T{
 \f[C][Output]\f[R]
 T}@T{
 \f[C]Mksquashfs=\f[R]
-T}
-T{
-\f[C]--xz\f[R]
-T}@T{
-\f[C][Output]\f[R]
-T}@T{
-\f[C]XZ=\f[R]
 T}
 T{
 \f[C]--qcow2\f[R]
@@ -2010,7 +2024,7 @@ file, and install \f[I]SSH\f[R] into it:
 .IP
 .nf
 \f[C]
-# mkosi -d fedora -t gpt_squashfs --checksum --xz --package=openssh-clients
+# mkosi -d fedora -t gpt_squashfs --checksum --compress --package=openssh-clients
 \f[R]
 .fi
 .PP
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -78,6 +78,7 @@ from .backend import (
     patch_file,
     run,
     run_workspace_command,
+    spawn,
     tmp_dir,
     var_tmp,
     warn,
@@ -3308,6 +3309,10 @@ def make_read_only(args: CommandLineArguments, root: str, for_cache: bool, b: bo
         btrfs_subvol_make_ro(root, b)
 
 
+def xz_binary() -> str:
+    return "pxz" if shutil.which("pxz") else "xz"
+
+
 def make_tar(args: CommandLineArguments, root: str, do_run_build_script: bool, for_cache: bool) -> Optional[BinaryIO]:
     if do_run_build_script:
         return None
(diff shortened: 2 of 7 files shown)


behrmann:
Very cool!

We already hat a PR about adding `--zstd` in addition to `--xz`, but it stalled.

keszybz:
That was #465. Yeah, I don't think another `--zstd` switch is the way to go. The "best" compression algorithm changes surprisingly often.

poettering:
not a fan of the "inner" and "outer" compression. I'd be precise here, and call one "fs" compression an the other "archive" compression. Naming is hard...

mkosi/__init__.py
```diff
@@ -5998,12 +5998,13 @@ def load_args(args: argparse.Namespace) -> CommandLineArguments:
...
+        opname = "acquire shell" if args.verb == "shell" else "boot"
         if args.output_format == OutputFormat.tar:
-            die("Sorry, can't acquire shell in or boot a tar archive.")
+            die(f"Sorry, can't {opname} with a tar archive.")
         if args.xz:
-            die("Sorry, can't acquire shell in or boot an XZ compressed image.")
+            die("Sorry, can't {opname} with a compressed image.")
         if args.qcow2:
-            die("Sorry, can't acquire shell in or boot a qcow2 image.")
+            die("Sorry, can't {opname} using a qcow2 image.")
```
poettering, line 6085: split out this one into a separate PR and merge right away?

mkosi/__init__.py
```diff
@@ -3721,20 +3722,63 @@ def make_tar(args: CommandLineArguments, root: str, do_run_build_script: bool, f
     if for_cache:
         return None
 
-    tar_root_dir = f"{root}/usr" if args.usr_only else root
+    root_dir = f"{root}/usr" if args.usr_only else root
```
poettering, line 3729: Maybe rename the function make_tar() → make_archive(). The log message already talks about "archive" anyway...

mkosi/__init__.py
```diff
@@ -3721,20 +3722,63 @@ def make_tar(args: CommandLineArguments, root: str, do_run_build_script: bool, f
...
+        with delay_interrupt():
+            files = file_list(root_dir)
+
+            with subprocess.Popen(
+                ["cpio", "-o", "--reproducible", "--null", "-H", "newc", "--quiet", "-D", root_dir],
+                stdin=subprocess.PIPE,
+                stdout=subprocess.PIPE,
+            ) as cpio:
+                with subprocess.Popen(
+                    ["pxz", "--check=crc32", "--lzma2=dict=1MiB", "-T0"], stdin=cpio.stdout, stdout=f
```
poettering: share the infra from xz_binary() here to use pxz only if installed, and use plain xz otherwise

mkosi/__init__.py
```diff
@@ -2304,6 +2304,10 @@ def install_fedora(args: CommandLineArguments, root: str, do_run_build_script: b
     with open(os.path.join(root, "etc/locale.conf"), "w") as f:
         f.write("LANG=C.UTF-8\n")
 
+    # FIXME: should this be conditionalized on args.with_docs like in install_debian_or_ubuntu()?
+    #        But we set LANG=C.UTF-8 anyway.
+    shutil.rmtree(os.path.join(root, "usr/share/locale"))
```
poettering, line 2309: isn't there an official way to get rid of this in fedora?

mkosi/__init__.py
```diff
@@ -3756,6 +3756,25 @@ def file_list(root: str) -> Generator[str, None, None]:
                 queue.append(entry.path)
 
 
+def cpio_compressor_command(option: Union[None, str, bool]) -> List[str]:
```
poettering, line 3759: shouldn't we share this with the tar codepaths? i.e. if we invoke the compressor manually for the cpio case there shouldn't be any reason why we couldn't do this for the tar case the exact same way?

mkosi/__init__.py
keszybz: That chunk is replaced later on with something that does the check…

mkosi/__init__.py
keszybz: Not directly. `rpm` has `--excludepaths`, but I'm not sure if there's a way to tell dnf to pass that.

mkosi/__init__.py
keszybz: Yeah, it would be nice to share this. I didn't do that so far because that'll require a bigger restructuring of how `tar` is invoked.

mkosi/__init__.py
keszybz: I changed tar→archive in various places to make the code more similar in `make_tar` and `make_cpio`. But this function should still be `make_tar`, because it creates a specific type of archive, and there are other functions for other types.

mkosi/__init__.py
keszybz: It's interleaved with subsequent changes… I had to make a few attempts to have sentence structure which works well with templating of all the options.

keszybz:
> not a fan of the "inner" and "outer" compression. I'd be precise here, and call one "fs" compression an the other "archive" compression. Naming is hard...

So... let's say we add EROFS support. (I think this could be a nicer alternative to squashfs images.)
EROFS has inner compression (lz4 only?). It might be useful to add outer compression with `xz` or `zstd` to shave off some bytes for distribution. It doesn't fit to call the outer compression "archive" in this case, because there's not archive involved.

mkosi/__init__.py
keszybz: https://bugzilla.redhat.com/show_bug.cgi?id=1967970

keszybz:
I think this is mostly ready. The last commit is the compression part, significantly reworked along the suggested lines.

keszybz:
```
Run python3 -m mypy mkosi
mkosi/__init__.py:3871:21: error: Item "None" of "Optional[IO[Any]]" has no
attribute "write"  [union-attr]
                        cpio.stdin.write(file.encode("utf8") + b"\0")
                        ^
mkosi/__init__.py:3872:17: error: Item "None" of "Optional[IO[Any]]" has no
attribute "close"  [union-attr]
                    cpio.stdin.close()
                    ^
Found 2 errors in 1 file (checked 3 source files)
Error: Process completed with exit code 1.
```
I don't get this here… Seems bogus.

behrmann:
> ```
> Run python3 -m mypy mkosi
> mkosi/__init__.py:3871:21: error: Item "None" of "Optional[IO[Any]]" has no
> attribute "write"  [union-attr]
>                         cpio.stdin.write(file.encode("utf8") + b"\0")
>                         ^
> mkosi/__init__.py:3872:17: error: Item "None" of "Optional[IO[Any]]" has no
> attribute "close"  [union-attr]
>                     cpio.stdin.close()
>                     ^
> Found 2 errors in 1 file (checked 3 source files)
> Error: Process completed with exit code 1.
> ```
> 
> I don't get this here… Seems bogus.

This is semi-bogus, I think. I'm not sure what the idea is, maybe they think, that the subprocess could have closed stdin? Adding a `assert cpio.stdin is not None` below resolves the issue.

behrmann:
Ah, I only just saw, that this is a new `popen` function. Yes, I guess is that mypy can't properly determine the type of `stdin` in the Popen and throws this error, because it could be `None`.

keszybz:
But I'm passing `stdin=subprocess.PIPE`, so we know that it is not None. Even if the file was closed, the object is not replaced by None, just a call to `.write()` would return an error. I will add the `assert` as suggested if that is necessary to resolve the warning, but I don't think mypy is being smart here. (I'll wait to see if there are any other comments though. I don't think it's worth to restart the CI just for that.)

behrmann:
That's true and it's clearly a bug. Would you care to open on against [mypy](https://github.com/python/mypy/issues)?

You can also add an appropriate type ignore comment to mollify mypy.

keszybz:
https://github.com/python/mypy/issues/10583

keszybz:
Now with a workaround for mypy.

I was experimenting with redirecting stdin from /dev/null for subprocesses. It doesn't help with the mypy error, but I think it's good to do that anyway, so I left that as a separate commit.

behrmann (commented):
Only did a quick readthrough and haven't tested this yet. I like the new option names a lot better. 

I think it would make sense to also add some test for the cpio output, so that we don't break it by chance at some future point.

poettering:
> So... let's say we add EROFS support. (I think this could be a nicer alternative to squashfs images.)
> EROFS has inner compression (lz4 only?). It might be useful to add outer compression with `xz` or `zstd` to shave off some bytes for distribution. It doesn't fit to call the outer compression "archive" in this case, because there's not archive involved.

The gzip manpage uses the word "archive" when talking about anything *.gz.  The "xz" manpaged uses it generically too to some point, but less so. (zstd man page doesn't mention the word at all). Hence I think it would not be too far off to say that a disk image that is compressed with one of the three compressors at the end, without a tar envelope is still an archive.

mkosi/__init__.py
```diff
@@ -3725,20 +3764,61 @@ def make_tar(args: CommandLineArguments, root: str, do_run_build_script: bool, f
...
         return None
 
-    tar_root_dir = f"{root}/usr" if args.usr_only else root
+    root_dir = f"{root}/usr" if args.usr_only else root
 
     with complete_step("Creating archive"):
         f: BinaryIO = cast(BinaryIO, tempfile.NamedTemporaryFile(dir=os.path.dirname(args.output), prefix=".mkosi-"))
         # OpenMandriva defaults to bsdtar(libarchive) which uses POSIX argument list so let's keep a separate list
         if shutil.which("bsdtar") and args.distribution == Distribution.openmandriva:
-            _tar_cmd = ["bsdtar", "-C", tar_root_dir, "-c", "-J", "--xattrs", "-f", "-", "."]
```
poettering, line 3730: (btw, i think to keep things systematic and well-defined on all distros we should always use "gtar" here first, and fall back to "tar", but never use bsdtar. In particular as bsdtar has no selinux control iirc)

but that's something to fix separately.

man/mkosi.1
```diff
@@ -392,33 +393,39 @@ against off-line modification, the verification data is placed in an
...
+Only applies to output formats with squashfs or btrfs.
 Takes one of \f[C]zlib\f[R], \f[C]lzo\f[R], \f[C]zstd\f[R],
 \f[C]lz4\f[R], \f[C]xz\f[R] or a boolean value as argument.
 If the latter is used compression is enabled/disabled and the default
 algorithm is used.
 In case of the \f[C]squashfs\f[R] output formats compression is implied,
-however this option may be used to select the algorithm.
+but this option may be used to select the algorithm.
+.TP
+\f[B]\f[CB]--compress-output=\f[B]\f[R]
```
poettering, line 406: "output" sounds good to me too, maybe indeed better than "archive"

poettering:
looks all good to me, but still no the python guru who could say overly smart things about the pythonicness of the code

mkosi/__init__.py
poettering: hmm, maybe we should add a new generic RemoveFiles= stanza that works on all distros, and can be used here? It would just list paths (with wildcards) to remove as very last step of the build or so?

mkosi/__init__.py
```diff
@@ -251,13 +260,46 @@ def run(
...
+
+    cm = do_delay_interrupt if delay_interrupt else do_noop
+    try:
+        with cm():
+            return subprocess.run(cmdline, check=check, stdin=stdin, stdout=stdout, stderr=stderr, **kwargs)
+    except FileNotFoundError:
+        die(f"{cmdline[0]} not found in PATH.")
+
+
+def popen(
```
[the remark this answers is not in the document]
keszybz: I did it because there's a `run()` function, which invokes `subprocess.run()`. So a new function which invokes `subprocess.Popen()` is called `popen()`. But I don't think the `Popen()` name is very good, so I'm happy to rename this. Maybe `spawn` ?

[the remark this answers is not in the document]
behrmann: I think I like spawn.

mkosi/__init__.py
poettering: And then maybe explicit provide an example in the docs, that it's a good idea to list /usr/share/locale in this setting

mkosi/__init__.py
```diff
@@ -3713,6 +3797,23 @@ def make_read_only(args: CommandLineArguments, root: str, for_cache: bool, b: bo
...
+def xz_binary() -> str:
+    return "pxz" if shutil.which("pxz") else "xz"
+
+
+def compressor_command(option: Union[str, bool]) -> List[str]:
+    """Returns a command suitable for compressing archives."""
+
+    if option == "xz":
+        return [xz_binary(), "--check=crc32", "--lzma2=dict=1MiB", "-T0"]
+    elif option == "zstd" or option is True:
```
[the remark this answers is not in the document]
keszybz: Indeed, this was a left-over from previous times.

mkosi/__init__.py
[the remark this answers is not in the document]
keszybz: I rewrote the two functions a bit: now there's spawn() which does all the work, and run() which is a thin wrapper. This way it's not necessary to duplicate the logging and argument mangling logic.

keszybz:
Updated with the requested changes.
I didn't add a test though: I'll work in that next.

mkosi/__init__.py
keszybz: `RemoveFiles=` sounds good. But let's do that separately, this one is already getting rather complex.

mkosi/__init__.py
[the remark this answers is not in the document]
keszybz: I rewrote the wrapper again. I had trouble getting all the stdin/stdout/input/output/stderr/text cases right.

mkosi/__init__.py
```diff
@@ -236,28 +238,71 @@ def handler(signal: int, frame: FrameType) -> None:
...
+        # output.
         stdout = sys.stderr
 
+    cm = do_delay_interrupt if delay_interrupt else do_noop
     try:
-        with delay_interrupt():
-            return subprocess.run(cmdline, check=check, stdout=stdout, stderr=stderr, **kwargs)
+        with cm():
+            return subprocess.run(
+                cmdline, check=check, input=input, stdin=stdin, stdout=stdout, stderr=stderr, **kwargs
```
behrmann, line 304: This can potentially byte a user, since `input` and `stdin` cannot be non-`None` at the same time. It might make more sense to standardise on either in our wrapper.

mkosi/__init__.py
```diff
@@ -3721,20 +3825,69 @@ def make_tar(args: CommandLineArguments, root: str, do_run_build_script: bool, f
...
-                _tar_cmd.append("--xattrs-exclude=security.selinux")
-            _tar_cmd.append(".")
+                cmd.append("--xattrs-exclude=security.selinux")
+            cmd.append(".")
 
-        run(_tar_cmd, env={"XZ_OPT": "-T0"}, stdout=f)
+        compressor = compressor_command(should_compress_output(args))
+
+        with spawn(cmd, stdout=subprocess.PIPE) as tar:
+            run(compressor, stdin=tar.stdout, stdout=f, delay_interrupt=False)
```
behrmann, line 3844: Is there an advantage not to use tar's `--use-compress-program` as `f"--use-compress-program='{compressor}'"`? The short option is different between GNU tar and BSD tar, but the long option has the same name.

mkosi/__init__.py
keszybz: There's a check above, and I think that should be enough. Standardizing on one of them is not easy, because they have completely different types and semantics...

mkosi/__init__.py
keszybz: Indeed. I'll rework it like that.

mkosi/__init__.py
behrmann: True, the thought came from the above check being for both not being `None`, but in the end this will be caught in `subprocess.run` itself, so that's fine.

keszybz:
Updated with two changes:
- `tar --use-compress-program=` is used instead of building a pipeline internally
- I dropped the commit to redirect stdin from `/dev/null`. I didn't investigate the details, but I think this was causing the CI failure. I'll submit that separately once dust settles.

keszybz:
CI passes, yay. Let's give it more to chew.

keszybz:
Rebased for the backend.py split.

behrmann:
> Rebased for the backend.py split.

Thanks! I almost feared I would have to pick out the things in #729 again :)

I'll have a look at this PR later today.

mkosi/__init__.py
```diff
@@ -5308,8 +5405,10 @@ def xescape(s: str) -> str:
     return ret
 
 
-def build_auxiliary_output_path(path: str, suffix: str, xz: Optional[bool] = False) -> str:
-    return strip_suffixes(path) + suffix + (".xz" if xz else "")
+def build_auxiliary_output_path(args: argparse.Namespace, suffix: str, can_compress: bool = False) -> str:
+    output = strip_suffixes(args.output)
+    compress = should_compress_output(args) if can_compress else False
+    return output + suffix + (f".{compress}" if compress else "")
```
behrmann, line 5411: Just a stylistic quibble, feel free to ignore it
```suggestion
    return f"{output}{suffix}.{compress or ''}"
```

behrmann (approved):
LGTM. Haven't run it myself, but there's little I could do that the tests don't already do. 

@DaanDeMeyer Can you have a look at this so that another pair of eyes looked this over?

mkosi/__init__.py
keszybz: No, that makes the dot unconditional, but it should be added only if the compression suffix is added.

behrmann: Right, chalk it up to post-vaccine grogginess :)

keszybz:
@DaanDeMeyer any comments?

DaanDeMeyer:
No time to give this a thorough look! But the changes look good and given @behrmann gave this a thorough look, I think we can merge.

(Given that I'm swamped with other things at the moment, feel free to merge things yourself @keszybz @poettering if I take too long)

PR #729, opened 2021-06-04
Split precursor

behrmann:
This PR splits out functions that will be reused when separating out distributions. This might look arbitrary, but it is just retracing the steps from #715.

Nothing changes functionally, the functions are just moved over into the (badly named) `mkosi/backend.py`. I'd be happy if somebody could come up with a better name.


keszybz:
The one Ci failure is a timeout, unrelated to this PR.

I think this is a good step.

PR #730, opened 2021-06-05
mkosi: do not repeat option lists in help output

keszybz:
```diff
--- /tmp/out1	2021-06-05 17:01:24.327828067 +0200
+++ /tmp/out2	2021-06-05 17:01:32.367897483 +0200
@@ -59,21 +59,19 @@
   --version             show program's version number and exit

 Distribution:
-  -d {fedora,debian,ubuntu,arch,opensuse,mageia,centos,centos_epel,clear,photon,openmandriva}, --distribution {fedora,debian,ubuntu,arch,opensuse,mageia,centos,centos_epel,clear,photon,openmandriva}
+  -d, --distribution {fedora,debian,ubuntu,arch,opensuse,mageia,centos,centos_epel,clear,photon,openmandriva}
                         Distribution to install
-  -r RELEASE, --release RELEASE
+  -r, --release RELEASE
                         Distribution release to install
-  -m MIRROR, --mirror MIRROR
-                        Distribution mirror to use
+  -m, --mirror MIRROR   Distribution mirror to use
   --repositories REPOS  Repositories to use
   --architecture ARCHITECTURE
                         Override the architecture of installation

 Output:
-  -t {directory,subvolume,tar,cpio,gpt_ext4,gpt_xfs,gpt_btrfs,gpt_squashfs,plain_squashfs}, --format {directory,subvolume,tar,cpio,gpt_ext4,gpt_xfs,gpt_btrfs,gpt_squashfs,plain_squashfs}
+  -t, --format {directory,subvolume,tar,cpio,gpt_ext4,gpt_xfs,gpt_btrfs,gpt_squashfs,plain_squashfs}
                         Output Format
-  -o PATH, --output PATH
-                        Output image path
+  -o, --output PATH     Output image path
   --output-split-root PATH
                         Output root or /usr/ partition image path (if --split-
                         artifacts is used)
```
… and so on.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -4696,6 +4696,15 @@ class WithNetworkAction(BooleanAction):
         super().__call__(parser, namespace, values, option_string)
 
 
+class CustomHelpFormatter(argparse.HelpFormatter):
+    def _format_action_invocation(self, action):
+        if not action.option_strings or action.nargs == 0:
+            return super()._format_action_invocation(action)
+        default = self._get_default_metavar_for_optional(action)
+        args_string = self._format_args(action, default)
+        return ", ".join(action.option_strings) + " " + args_string
+
+
 class ArgumentParserMkosi(argparse.ArgumentParser):
     """ArgumentParser with support for mkosi.defaults file(s)
 
@@ -4735,8 +4744,12 @@ class ArgumentParserMkosi(argparse.ArgumentParser):
         self._ini_file_list_mode = False
 
         # Add config files to be parsed
-        kwargs["fromfile_prefix_chars"] = ArgumentParserMkosi.fromfile_prefix_chars
-        super().__init__(*kargs, **kwargs)
+        super().__init__(
+            *kargs,
+            **kwargs,
+            fromfile_prefix_chars=ArgumentParserMkosi.fromfile_prefix_chars,
+            formatter_class=CustomHelpFormatter,
+        )
 
     @staticmethod
     def _camel_to_arg(camel: str) -> str:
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -4696,6 +4696,15 @@ def __call__(
         super().__call__(parser, namespace, values, option_string)
 
 
+class CustomHelpFormatter(argparse.HelpFormatter):
+    def _format_action_invocation(self, action):
```
behrmann, line 4700:
```suggestion
    def _format_action_invocation(self, action: argparse.Action) -> str:
```

mkosi/__init__.py
```diff
@@ -4735,8 +4744,12 @@ def __init__(self, *kargs: Any, **kwargs: Any) -> None:
...
 
         # Add config files to be parsed
-        kwargs["fromfile_prefix_chars"] = ArgumentParserMkosi.fromfile_prefix_chars
-        super().__init__(*kargs, **kwargs)
+        super().__init__(
+            *kargs,
+            **kwargs,
+            fromfile_prefix_chars=ArgumentParserMkosi.fromfile_prefix_chars,
+            formatter_class=CustomHelpFormatter,
+        )
```
behrmann, line 4752: mypy errors here because it can't generally prove, that kwargs will never have those keys, which would lead to an error at runtime. This doesn't have the same "problem" (not that practically would be an issue) and maybe conveys the the intention a bit better.
```suggestion
        kwargs["fromfile_prefix_chars"] = ArgumentParserMkosi.fromfile_prefix_chars
        kwargs["formatter_class"] = CustomHelpFormatter

        super().__init__(*kargs, **kwargs)
```

behrmann (commented):
A very welcome change!

PR #731, opened 2021-06-08
Fix short-form 'readonly' deprecation

eankeen:
Running `mkosi` on my machine

```sh
➤ sudo mkosi qemu
‣ Running command:
‣ qemu-system-x86_64 -machine accel=kvm -smp 2 -m 1024 -cpu host -drive if=pflash,format=raw,readonly,file=/usr/share/ovmf/x64/OVMF_CODE.fd -object rng-random,filename=/dev/urandom,id=rng0 -device virtio-rng-pci,rng=rng0,id=rng-device0 -vga virtio -drive format=raw,file=/storage/edwin/Docs/Programming/repos/dots-bootstrap/tests/data/mkosi/image.raw,if=virtio

qemu-system-x86_64: -drive if=pflash,format=raw,readonly,file=/usr/share/ovmf/x64/OVMF_CODE.fd: warning: short-form boolean option 'readonly' deprecated
Please use readonly=on instead
```

This appends `=on` to remove the deprecation warning

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -7014,7 +7014,7 @@ def run_qemu(args: CommandLineArguments) -> None:
         cmdline += ["-nic", f"tap,script=no,downscript=no,ifname={ifname},model=virtio-net-pci"]
 
     if "uefi" in args.boot_protocols:
-        cmdline += ["-drive", f"if=pflash,format=raw,readonly,file={firmware}"]
+        cmdline += ["-drive", f"if=pflash,format=raw,readonly=on,file={firmware}"]
 
     with contextlib.ExitStack() as stack:
         if fw_supports_sb:


behrmann:
Thank you for the contribution! This looks uncritical, but do you know when this was deprecated and whether older versions of qemu support `readonly=on`? I've only quickly glanced the Qemu manpage on Arch (qemu 6.0) and Debian buster (qemu 3.1 and qemu 5.2 from buster-backports) and the latter doesn't mention `readonly=on` in its man page. Since qemu 5.2 is rather new (and will also be the default version in Debian bullseye which isn't even released yet) it would be problematic to change this if it isn't supported on older versions.

eankeen:
Ahh, good point, I did some digging and these shortform options have been deprecated since [version 6.0](https://qemu-project.gitlab.io/qemu/system/deprecated.html#short-form-boolean-options-since-6-0) and introduced in commit [ccd3b3b8112b670fdccf8a392b8419b173ffccb4](https://gitlab.com/qemu-project/qemu/-/commit/ccd3b3b8112b670fdccf8a392b8419b173ffccb4). The test that covers the behavior the deprecation recommends looks to be [here](https://gitlab.com/qemu-project/qemu/-/blob/ccd3b3b8112b670fdccf8a392b8419b173ffccb4/tests/test-qemu-opts.c#L518) and was introduced [4 years prior](https://gitlab.com/qemu-project/qemu/-/blob/694baf57ae103954e2337d1fa22d1dcdfa8e2f5a/tests/test-qemu-opts.c#L527). It looks like this test was added right before the [v2.9.0-rc0 release](https://gitlab.com/qemu-project/qemu/-/tree/d84f714eafedd8bb9d4aaec8b76417bef8e3535e/)

According to Repology, Fedora 26 uses version 2.9.1, Debian Oldstable uses 2.8 (Stable uses 3.1), and Ubuntu 16.04 uses 2.5 (Ubuntu 18.04 uses 2.11). Is Oldstable and 16.04 supported? If not, then in my view this is a mergable change, but if you intend to support versions of QEMU that are more than 4 years old, then there is still a slight chance that it could break - I suppose then I would have to check the behavior manually

keszybz:
Oldstable has python-3.5.3, and we require >= 3.7, so it's already unsupported.
I'm not sure about 16.04: https://packages.ubuntu.com/search?keywords=python3 doesn't list it for 16.04 at all.
So I think it's fine to merge this.

DaanDeMeyer:
CI failure seems unrelated.

Issue #536, opened 2020-10-17
Support providiing extra repositories from config files


markwellis:
I can no longer use extra repos with fedora, I get an error they don't exist.

Looks like it's because `install_fedora` calls `setup_dnf` here https://github.com/systemd/mkosi/blob/master/mkosi/__init__.py#L1881-L1884 but only with 2 predefined repos.

When `invoke_dnf` is ran here https://github.com/systemd/mkosi/blob/master/mkosi/__init__.py#L1893 with the extra repos, they're not in `temp.root` as created by setup_dnf.

DaanDeMeyer:
PRs appreciated! Unfortunately, --repositories is a bit of a mess with different meanings per distro.

DaanDeMeyer:
Could you provide some more information on how you were passing these extra repos to mkosi? An example to reproduce would be great!

markwellis:
mkosi.default
```
[Distribution]
Distribution=fedora
Release=33

Repositories=
     fedora
     updates
     rpmfusion-nonfree
     rpmfusion-nonfree-updates

[Output]
Format=directory
Output=out

[Packages]
Packages=
    unrar
```

DaanDeMeyer:
I can't seem to figure out how this worked before. Is there an associated url for these repositories?

markwellis:
If I remember correctly, they came from the host. Because that's fedora, and the repos were inherited and the ones listed here were not removed

DaanDeMeyer:
So what behavior do you want here? The idea is to make mkosi as independent from the host it's running on as we can. To configure these repositories, I'd need a url to configure in dnf's repo file (and preferably a url for the gpg key as well). Without those, I'm not sure how to support this.

markwellis:
Makes sense. I'd be happy to add those to my config. I just after a way of adding repos via the config file so that I can install packages from there. If that means it needs a directory full or .repo files, or a bunch of config with names, urls, pgp keys etc, that's also fine.

DaanDeMeyer:
Allowing custom repo files sounds like it would be a good addition. No cycles to work on that atm unfortunately but will be happy to review a PR.

PR #732, opened 2021-06-10
Added option to skip use of hardcoded default distro repositories

ray-sit:
An alternative solution to https://github.com/systemd/mkosi/pull/662 that should also resolve https://github.com/systemd/mkosi/issues/536

Adding this option reverts mkosi functionality to how it was working for fedora/centos/rhel previously.

Allowing it to use the repo files that exist on the system, instead of locking the script into using the hardcoded repo file only containing the default repos for each distro.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -335,6 +335,15 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
   Linux, additional repositories must be passed in the form `<name>::<url>`
   (e.g. `myrepo::https://myrepo.net`).
 
+`UseSystemRepositories=`, `--use-system-repositories`
+
+: This option is only applicable for dnf-based distributions:
+  *CentOS*, *Fedora Linux*, *Mageia*, *Photon*, and *OpenMandriva*.
+  Allows use of the system's existing dnf repositories.
+  By default, a hardcoded set of default dnf repositories is generated and used.
+  Use `--repositories=` to identify a custom set of repositories to be enabled
+  and used for the build.
+
 `Architecture=`, `--architecture=`
 
 : The architecture to build the image for. Note that this currently
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1925,13 +1925,18 @@ def setup_dnf(args: CommandLineArguments, root: Path, repos: Sequence[Repo] = ()
                 )
             )
 
+    if args.use_system_repositories:
+        default_repos  = ""
+    else:
+        default_repos  = f"{'repodir' if args.distribution == Distribution.photon else 'reposdir'}={workspace(root)}"
+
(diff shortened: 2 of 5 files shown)


ray-sit:
Sorry after talking about this to some of my colleagues seems like I mistook an internal patch as legacy code, nevertheless this issue still exists and will be resolved by this change.

mkosi/__init__.py
```diff
@@ -1792,13 +1792,14 @@ def setup_dnf(args: CommandLineArguments, root: str, repos: List[Repo] = []) ->
             )
 
     config_file = os.path.join(workspace(root), "dnf.conf")
+    default_repos = f"{'repodir' if args.distribution == Distribution.photon else 'reposdir'}={workspace(root)}"
```
keszybz, line 1795: Maybe write this as
```python
if args.use_system_repositories:
   default_repos = ''
else:
   default_repos = f"{'repodir' if args.distribution == Distribution.photon else 'reposdir'}={workspace(root)}\n"
```
This will be longer but clearer, I think.

keszybz (changes requested):
Sorry for the delay. This looks reasonable, but needs a rebase now. Please also adjust the commit message with the stuff that is currently only present in your comment on the pull request.

mkosi/__init__.py
ray-sit: Applied suggestion.

Rebased and added the additional details into the commit message.

mkosi.md
```diff
@@ -335,6 +335,14 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
   Linux, additional repositories must be passed in the form `<name>::<url>`
   (e.g. `myrepo::https://myrepo.net`).
 
+`UseSystemRepositories=`, `--use-system-repositories`
+
+: Allows use of the system's existing software package repositories.
+  By default a hardcoded set of default repositories for each distribution
```
behrmann, line 341: Currently this is an dnf-only thing, right? If so, it would be good to note that here.

ray-sit: Yes, good point.

I've taken a stab at updating this to identify that it is dnf-only.

keszybz:
```
[  224.783474] watchdog: BUG: soft lockup - CPU#1 stuck for 186s! [systemd-udevd:405]
```
Man, you broke the cloud!

keszybz:
OK, I think this can go in. With the default, there should be no change, so this pretty safe.

poettering:
hmm, what is "system" supposed to mean here? the host? if so, i figure this should be called "host".

"system" can mean anything and nothing really. But "host" clearly means that it's not the container we build, but the host we build it on, and that's what is meant here, no?

poettering:
Filed #808 to keep track of the request to rename this option

Issue #733, opened 2021-06-10
Dracut fails on missing modules.dep


ray-sit:
There was a change from `v6` that removed a check that a kernel module directory contains a dependency list (`modules.dep`). This causes dracut to fail in some cases where a module lib is added but depmod isn't successfully executed to populate a `modules.dep` file.

The expected behavior should be the same as it was before, check if the modules dependency file exists, and if it doesn't then simply skip that directory.

Line in question: https://github.com/systemd/mkosi/blob/0d7de1dee05d4f6d581254c144c95bbe7a761307/mkosi/__init__.py#L4096

ray-sit:
Sorry seems like I mistook an internal patch as legacy code, nevertheless this issue still exists.

poettering:
This suggests /usr/lib/kernel/install.d/50-depmod.install doesn't work for you?

which distro?

ray-sit:
On CentOS 8.

The cause is trying install a precompiled kernel module via RPM, where the RPM installation creates its own directory (due to a mismatch in kernel version specified in the RPM and the system itself) in `/usr/lib/...` to install its files there.

There is a line in `/usr/lib/kernel/install.d/50-depmod.install` that seems to delete `modules.{...}` after running `depmod -a`? I am not honestly not sure what the script is trying to accomplish here, maybe this is an issue or incompatibility with mksoi?
https://github.com/systemd/systemd/blob/07eabc2bebf6aae5c02ba15afc80235e916131cb/src/kernel-install/50-depmod.install#L39

I can confirm that skipping the directories without `modules.dep` results in not trying to run `dracut` in a directory without the required `modules.dep`, and produces an image with the precompiled module installed as intended.

PR #734, opened 2021-06-10
Added in check for modules.dep prior to running dracut install

ray-sit:
Revert change that happened from v5 to v6 that removes the check for `modules.dep` and decreases the reliability of the tool.

Resolves https://github.com/systemd/mkosi/issues/733


keszybz:
Eh, let's merge this, it shouldn't hurt. As discussed in #733, it solves an issue on centos.

PR #735, opened 2021-06-11
[DON'T MERGE] typing: use ParamSpec for @dictify

keszybz:
This only works with Python 3.10, see
https://www.python.org/dev/peps/pep-0612/.
Unfortunately mypy still doesn't support this, so it throws
an error. But based on the docs, this should "just work".


DaanDeMeyer:
Let's close this, we can resurrect once we move to python 3.10

PR #739, opened 2021-06-14
documented execution flow

### LINKED ISSUE #737: RFC: document execution flow
When starting with mkosi it is hard to distinguish/order the actions that mkosi takes to construct an image,
especially since information on each step are separated over multiple locations in the man page
A simple flowchart would help to locate these steps and see in which context they are executed.

nopeslide:
fix #737 
It is hard to understand the mkosi execution flow,
since relevant information is scattered throughout the man page.
Therefore I added a simplified execution flow graph as ascii art to the man page.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -179,6 +179,113 @@ The following command line verbs are known:
 : This verb is equivalent to the `--help` switch documented below: it
   shows a brief usage explanation.
 
+## Execution flow
+
+Execution flow for `mkosi build`. Columns represent the execution context.
+Default values/calls are shown in parentheses.
+When building with `--incremental` mkosi creates a cache of the distribution
+installation for both images if not already existing and replaces the
+distribution installation in consecutive runs with data from the cached one.
+
+```
+            HOST            .          BUILD          .          FINAL
+                            .          IMAGE          .          IMAGE
+                            .                         .
+   start                    .                         .
+     |                      .                         .
+     v                      .                         .
+build script?  -------exists----->     copy           .
+     |                      .      skeleton trees     .
+     |                      .     (mkosi.skeleton/)   .
+    none                    .            |            .
+     |                      .            v            .
+     v                      .         install         .
+    skip                    .       distribution,     .
+ build image                .       packages and      .
+     |                      .      build packages,    .
+     |                      .           run           .
+     |                      .      prepare script     .
+     |                      .   (mkosi.prepare build) .
+     |                      .     or if incremental   .
+     |                      .  use cached build image .
+     |                      .            |            .
+     |                      .            v            .
+     |                      .          copy           .
+     |                      .       extra trees       .
+     |                      .      (mkosi.extra/)     .
+     |                      .            |            .
+     |                      .            v            .
+     |                      .          copy           .
+     |                      .      build sources      .
+     |                      .          (./)           .
+     |                      .            |            .
+     |                      .            v            .
+     |                      .           run           .
+     |                      .    postinstall script   .
+     |                      .  (mkosi.postinst build) .
+     |                      .            |            .
+     |         .-------------------------'            .
+     |         |            .                         .
+     |         v            .                         .
+     |        run           .                         .
+     |   finalize script    .                         .
+     |(mkosi.finalize build).                         .
+     |         |            .                         .
+     |         '-------------------------.            .
+     |                      .            |            .
+     |                      .            v            .
+     |                      .           run           .
+     |                      .       build script      .
+     |                      .      (mkosi.build)      .
+     |                      .            |            .
+     '-----------------------------------+------------------------.
+                            .                         .           |
+                            .                         .           v
+                            .                         .          copy
+                            .                         .     skeleton trees
+                            .                         .    (mkosi.skeleton/)
+                            .                         .           |
+                            .                         .           v
+                            .                         .        install
+                            .                         .      distribution
+                            .                         .      and packages,
+                            .                         .          run
+                            .                         .     prepare script
+                            .                         .  (mkosi.prepare final)
+                            .                         .    or if incremental
+                            .                         . use cached final image
+                            .                         .           |
+                            .                         .           v
+                            .                         .         copy
+                            .                         .      extra trees
+                            .                         .     (mkosi.extra/)
+                            .                         .           |
+                            .                         .           v
+                            .                         .         copy
+                            .                         .     build results
+                            .                         .           |
+                            .                         .           v
+                            .                         .          run
+                            .                         .    postinstall script
+                            .                         .  (mkosi.postinst final)
+                            .                         .           |
+               .--------------------------------------------------'
+               |            .                         .
+               v            .                         .
+              run           .                         .
+        finalize script     .                         .
+    (mkosi.finalize final)  .                         .
+               |            .                         .
+     .---------'            .                         .
+     |                      .                         .
+     v                      .                         .
+    end                     .                         .
+                            .                         .
+            HOST            .          BUILD          .          FINAL
+                            .          IMAGE          .          IMAGE
+                            .                         .
+```
+
 ## Command Line Parameters
 
 The following command line parameters are understood. Note that many


keszybz:
Thanks, this looks pretty nice.
The only issue is where in the man page this new section should go.
Right now it is between "Command line parameters" and "Command line parameters and their settings file counterparts", which seems like the worst possible place. Can you move after "Commands" instead?

(Why there: In general, an overview chart like this should be as early as possible, because the man page should start with general concepts and the "drill down" into details. But this chart only applies to `build` verb. So we can't show the chart before introducing `build`, which is in "Commands".)

keszybz:
Oh, and maybe also add a sentence like:
This is the workflow when `build` is invoked.

mkosi.md
```diff
@@ -937,6 +937,110 @@ details see the table below.
...
+      |                 .         packages         .
+      |                 .            |             .
+      |                 .            v             .
+      |                 .         install          .
+      |                 .      build packages      .
+      |                 .            |             .
+      |                 .            v             .
+      |                 .           run            .
+      |                 .      prepare script      .
+      |                 .         (build)          .
```
keszybz, line 967: Maybe mention `mkosi.prepare` here? E.g. change
```diff
- (build)
+ (`mkosi.prepare build`)
```

mkosi.md
```diff
@@ -937,6 +937,110 @@ details see the table below.
...
+      |                 .       extra trees        .
+      |                 .            |             .
+      |                 .            v             .
+      |                 .          copy            .
+      |                 .      build sources       .
+      |                 .            |             .
+      |                 .            v             .
+      |                 .           run            .
+      |                 .     postinstall script   .
+      |                 .         (build)          .
```
keszybz, line 980: `(mkosi.postinstall build)` ?

mkosi.md
```diff
@@ -937,6 +937,110 @@ details see the table below.
...
+      |        v        .                          .
+      |       run       .                          .
+      | finalize script .                          .
+      |     (build)     .                          .
+      |        |        .                          .
+      |        '---------------------.             .
+      |                 .            |             .
+      |                 .            v             .
+      |                 .           run            .
+      |                 .       build script       .
```
keszybz, line 993: ...

mkosi.md
```diff
@@ -937,6 +937,110 @@ details see the table below.
...
+                        .                          .       distribution
+                        .                          .            |
+                        .                          .            v
+                        .                          .         install
+                        .                          .         packages
+                        .                          .            |
+                        .                          .            v
+                        .                          .           run
+                        .                          .      prepare script
+                        .                          .         (final)
```
keszybz, line 1012: ...

mkosi.md
```diff
@@ -937,6 +937,110 @@ details see the table below.
...
+                        .                          .       extra trees
+                        .                          .            |
+                        .                          .            v
+                        .                          .          copy
+                        .                          .   build destination
+                        .                          .            |
+                        .                          .            v
+                        .                          .           run
+                        .                          .     postinstall script
+                        .                          .         (final)
```
keszybz, line 1025: ...

mkosi.md
```diff
@@ -937,6 +937,110 @@ details see the table below.
...
+                        .                          .            |
+               .------------------------------------------------'
+               |        .                          .
+               v        .                          .
+              run       .                          .
+        finalize script .                          .
+            (final)     .                          .
```
keszybz, line 1032: ...

keszybz (commented):
I think it'd be nice to add the (default) script names, to make it easy to figure out which file one wants to edit to change some phase of the build.

behrmann:
Just ran `pandoc -t man -s -o man/mkosi.1 mkosi.md` on it to generate the man page and it also comes out nicely there. Adding a short paragraph before the flowchart does improve the placement within the section, though.

behrmann:
This looks pretty great. I think you could still add the default path names for `mkosi.extra` and `mkosi.skeleton` and squash the commits, but apart from that, this is really nice!

bcbnz:
I think it would be useful to add the points at which the cached images are stored/loaded for incremental builds (after the prepare script is run). Maybe it should also be noted (afterwards, it might be confusing to try and add it in the chart) that the first run of an incremental build constructs both cached images first and then continues the rest of the build.

nopeslide:
> I think it would be useful to add the points at which the cached images are stored/loaded for incremental builds (after the prepare script is run). Maybe it should also be noted (afterwards, it might be confusing to try and add it in the chart) that the first run of an incremental build constructs both cached images first and then continues the rest of the build.

I left the incremental build out by intention, because it would be quite confusing because it breaks the "logical" flow.

nopeslide:
> This looks pretty great. I think you could still add the default path names for `mkosi.extra` and `mkosi.skeleton` and squash the commits, but apart from that, this is really nice!

good point. I added the default paths

bcbnz:
> I left the incremental build out intentionally.
> the result would be quite confusing because it breaks the "logical" flow.

OK. Could it be mentioned in the text before or after the diagram? I feel it is an important part of the execution flow and should be part of this section.

Nice diagram by the way, I am finding it quite useful as a source for documenting one of our repositories which builds images.

nopeslide:
> OK. Could it be mentioned in the text before or after the diagram? I feel it is an important part of the execution flow and should be part of this section.

Is this alright? I tried to solve it textual, so it doesn't complicate the diagram itself

bcbnz:
I couldn't get a review on the commit to work properly, but how about

> When building with \`--incremental\` mkosi creates a cache of both images
> after running the prepare script on the first run, and starts from this point in
> subsequent runs.

instead? It is a bit more explicit about when it creates the cache and how it uses it.

nopeslide:
> I couldn't get a review on the commit to work properly, but how about
> 
> > When building with `--incremental` mkosi creates a cache of both images
> > after running the prepare script on the first run, and starts from this point in
> > subsequent runs.
> 
> instead? It is a bit more explicit about when it creates the cache and how it uses it.

I skimmed the code and it seems mkosi skips every hook when building the cached image and only installs the distribution related stuff.
When using the cached image it copies data from the cached image and skips the distribution installation.
Did I misunderstand sth?

nopeslide:
I commented out the lines that shouldn't be triggered in each situation by simply looking if `cached_tree` or `for_cache` is used as parameter

when building the cache:
```python
def build_image(...)
    ...
    cached_tree = reuse_cache_tree(args, root, do_run_build_script, for_cache, cached)
    install_skeleton_trees(args, root)
    install_distribution(args, root, do_run_build_script, cached_tree)
    install_etc_hostname(args, root, cached_tree)
    install_boot_loader(args, root, loopdev, do_run_build_script, cached_tree)
    run_prepare_script(args, root, do_run_build_script, cached_tree)
    #install_extra_trees(args, root, for_cache)
    #install_build_src(args, root, do_run_build_script, for_cache)
    #install_build_dest(args, root, do_run_build_script, for_cache)
    set_root_password(args, root, do_run_build_script, cached_tree)
    set_serial_terminal(args, root, do_run_build_script, cached_tree)
    set_autologin(args, root, do_run_build_script, cached_tree)
    sshkey = setup_ssh(args, root, do_run_build_script, for_cache, cached_tree)
    setup_network_veth(args, root, do_run_build_script, cached_tree)
    #run_postinst_script(args, root, loopdev, do_run_build_script, for_cache)

    if cleanup:
        clean_package_manager_metadata(root)
        remove_files(args, root)
    #reset_machine_id(args, root, do_run_build_script, for_cache)
    reset_random_seed(args, root)
    #run_finalize_script(args, root, do_run_build_script, for_cache)
    #invoke_fstrim(args, root, do_run_build_script, for_cache)
    #make_read_only(args, root, for_cache)
    ...
```
when using the cache
```python
def build_image(...)
    ...
    cached_tree = reuse_cache_tree(args, root, do_run_build_script, for_cache, cached)
    install_skeleton_trees(args, root)
    #install_distribution(args, root, do_run_build_script, cached_tree)
    #install_etc_hostname(args, root, cached_tree)
    #install_boot_loader(args, root, loopdev, do_run_build_script, cached_tree)
    #run_prepare_script(args, root, do_run_build_script, cached_tree)
    install_extra_trees(args, root, for_cache)
    install_build_src(args, root, do_run_build_script, for_cache)
    install_build_dest(args, root, do_run_build_script, for_cache)
    #set_root_password(args, root, do_run_build_script, cached_tree)
    #set_serial_terminal(args, root, do_run_build_script, cached_tree)
    set_autologin(args, root, do_run_build_script, cached_tree)
    #sshkey = setup_ssh(args, root, do_run_build_script, for_cache, cached_tree)
    #setup_network_veth(args, root, do_run_build_script, cached_tree)
    run_postinst_script(args, root, loopdev, do_run_build_script, for_cache)

    if cleanup:
        clean_package_manager_metadata(root)
        remove_files(args, root)
    reset_machine_id(args, root, do_run_build_script, for_cache)
    reset_random_seed(args, root)
    run_finalize_script(args, root, do_run_build_script, for_cache)
    invoke_fstrim(args, root, do_run_build_script, for_cache)
    make_read_only(args, root, for_cache)
    ...
```

bcbnz:
It certainly runs the prepare script -- I just booted one of the cached images I have here and the changes made by prepare exist in the cached image. Also, the current manpage says 

> --prepare-script=
>           Takes  a  path  to an executable that is invoked inside the image right after installing the software packages.  It is
>            the last step before the image is cached (if incremental mode is enabled).

This is using raw image outputs, somebody more familiar with the code may be able to comment if there are differences with other image types.

bcbnz:
And the skeleton files are in the cached image as well.

nopeslide:
> And the skeleton files are in the cached image as well.

they are copied regardless if ached or not

nopeslide:
> > --prepare-script=
> > Takes  a  path  to an executable that is invoked inside the image right after installing the software packages.  It is
> > the last step before the image is cached (if incremental mode is enabled).

good catch. I merged the steps

bcbnz:
> > And the skeleton files are in the cached image as well.
> 
> they are copied regardless if cached or not

You're correct, I hadn't noticed. Output from one of my builds confirms that.

> ‣ Copied cached image as ...
> ‣ Attaching image file...
> ‣ Attached image file as /dev/loop0.
> ‣ Mounting image...
> ‣ Copying in skeleton file trees...
> ‣ Copying in extra file trees...
> ‣ Copying in build script...
> ‣ Copying in sources...
> ‣ Mounting Package Cache...
> ‣ Running postinstall script...

I would actually argue this is a bug -- if the package installation or the prepare script (which is not run again after loading the cache) changes any of the files from the skeleton tree (e.g., if the skeleton tree was an initial configuration for something which is then updated) then re-copying would overwrite those changes. I guess that should probably be the topic of a different discussion though.

nopeslide:
@bcbnz more thoughts? would squash otherwise

keszybz (commented):
Looks great, thanks.

PR #740, opened 2021-06-15
mkosi: speed up "mkosi -i -f -B"

poettering:
When building images with image IDs and image versions it's a good idea
to reuse the pre-pared cached images from the earlier versions (I mean,
that's why cacheing/incremental operation exists). So far the cached
image names was derived 1:1 from the output file name, but if that
includs the image version it will be different for each build and hence
never suitable as cache.

Hence, if the image ID/version logic is used, let's change the naming
regime for the cache file, and let's derive it direcly from image name,
not taking the version into account.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -725,11 +725,24 @@ def determine_partition_table(args: CommandLineArguments) -> Tuple[str, bool]:
     return table, run_sfdisk
 
 
+def exec_sfdisk(args: CommandLineArguments, f: BinaryIO) -> None:
+
+    table, run_sfdisk = determine_partition_table(args)
+
+    if run_sfdisk:
+        run(["sfdisk", "--color=never", f.name], input=table.encode("utf-8"))
+        run(["sync"])
+
+    args.ran_sfdisk = run_sfdisk
+
+
 def create_image(args: CommandLineArguments, for_cache: bool) -> Optional[BinaryIO]:
     if not args.output_format.is_disk():
         return None
 
-    with complete_step("Creating partition table", "Created partition table as {.name}") as output:
+    with complete_step(
+        "Creating image with partition table", "Created image with partition table as {.name}"
+    ) as output:
 
         f: BinaryIO = cast(
             BinaryIO,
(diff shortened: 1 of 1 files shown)


behrmann:
mypy needs the change to `.github/workflows/ci-unit-test.yml` from #728 to not break.

poettering:
hmm, quite possibly this might not acually the best idea to merge.

the thing is that when operating like this the partition labels/uuids and fs uuid are always identical (since they come from the original cached entry). Which is extremely confusing to the linux kernel (btrfs really hates multiple file system with the same uuid).

this probably needs more love: it's fine to reuse the file systems I guess, but we should update at least:

1. The partition uuids
2. The partition labels (in particular if image_version is set, so that they match the current version being built)
3. The disk uuid
4. The file system uuid 
5. The file system label

the three uuids should probably be hashed out of the image id + image version in a stable way, to provide some form of reproducible builds at least superficially.

poettering:
I added another commit now that addresses the UUID issue by randomizing them whenever we reuse a cached build. Should be good for review/merge now.

mkosi/__init__.py
```diff
@@ -739,15 +752,61 @@ def create_image(args: CommandLineArguments, root: str, for_cache: bool) -> Opti
...
+    # btrfs since it prohibits mounting multiple file systems at the
+    # same time that carry the same UUID.
+    #
+    # FIXME: One of those days we should generate the UUIDs as hashes
+    # of the used configuration, so that they remain stable as the
+    # configuration is identical.
+
+    with complete_step(f"Refreshing file system {dev}."):
+        if args.output_format == OutputFormat.gpt_btrfs:
+            run(["btrfstune", "-m", dev])
```
behrmann, line 801: The `-m` key is only available starting with kernel 5.0 and the current Debian stable is still on 4.19, when not using backports-kernels. I'm not too concerned about this, since Debian is missing repartd and all the other nice things, that would probably be needed, to actually make use of this.

An alternative would be to use `-M` and Python's `uuid` module.

poettering: Changed it to use -M. PTAL

mkosi/__init__.py
```diff
@@ -739,15 +752,61 @@ def create_image(args: CommandLineArguments, root: str, for_cache: bool) -> Opti
...
-            run(["sfdisk", "--color=never", f.name], input=table.encode("utf-8"))
-            run(["sync"])
+    return f
 
-        args.ran_sfdisk = run_sfdisk
 
-    return f
+def refresh_partition_table(args: CommandLineArguments, f: BinaryIO) -> None:
+    if not args.output_format.is_disk():
+        return None
```
behrmann, line 762: Let's try to make for a more uniform style here. 
```suggestion
        return
```
Whether `None` is returned explicitly or implicitly is a bit all over the place, but the implicit returns outnumber the explicit returns 2:1, so I guess that's the way to go for new code.

mkosi/backend.py
```diff
@@ -457,6 +457,7 @@ def run(
     cmdline: List[str],
     check: bool = True,
     delay_interrupt: bool = True,
+    stdin: _FILE = None,
```
behrmann, line 460: The two changes in this file are not strictly necessary, since `stdin` will be just picked up via `kwargs` and mypy also doesn't complain. 

I also remember @keszybz wanted to have another go at stdin handling in this function anyway.

behrmann (commented):
LGTM. Haven't run it yet, but I think the logic is sound. 

Some minor stylistic comment.

mkosi/__init__.py
poettering: ah, i didn't know that "return" and "return None" was equivalent. Interesting.

What's the general coding style suggestions (as implemented in black and so on) about this? Is one clearly preferable to the other? What if the return type of the function is `Optional[int]`, i.e. wouldn't it make sense in such a case to do an explicit `return None`, because it's not a void function but actually one that sometimes returns something?

mkosi/backend.py
poettering: ah, interesting, makes sense

poettering:
force pushed a new version, fixing the issues raised. ptal!

mkosi/__init__.py
behrmann: All functions in Python return a value, so there isn't really the concept of a void function. Not having a return is equivalent to having a bare return, both return `None`.

What's more pythonic is, I think, harder to answer. Sure, Python generally says "Explicit is better than implicit", but grepping the stdlib, there too implicitly returning `None` outweighs explicit returns 2:1. Since type hints are still a rather young thing in the whole scheme of things Python, I don't think there's a specific culture for that yet either. I think it's mostly a thing of personal preference and at least black hasn't started regulate this yet and I only remarked on this, because there was one function with an explicit `None`-return and one with implicit ones. :)

I think implicitly returning `None` makes sense at points where the code bails out and the value of `None` is not of interest, i.e. in something close to a void function, whereas explicitly returning `None` is useful when we are interested in the actual value down the road.

mkosi/__init__.py
```diff
@@ -725,11 +725,24 @@ def determine_partition_table(args: CommandLineArguments) -> Tuple[str, bool]:
...
 
-def create_image(args: CommandLineArguments, root: str, for_cache: bool) -> Optional[BinaryIO]:
+def exec_sfdisk(args: CommandLineArguments, f: BinaryIO) -> None:
+
+    table, run_sfdisk = determine_partition_table(args)
+
+    if run_sfdisk:
+        run(["sfdisk", "--color=never", f.name], input=table.encode("utf-8"))
+        run(["sync"])
+
```
behrmann, line 735: Comparing with `insert_partition` should a call to `partx --update` be added here or in `refresh_partitions` added as well to reread labels from the partition table?

behrmann:
Besides that, still haven't run it, but LGTM. CI failures are unrelated.

mkosi/__init__.py
poettering: So, the "partx --update" is just synchronously synthesizes kernel partition block devices from userspace. But we don't need that here. there partition devices already exist, we just refresh the uuids and labels here, but the block layer doesn't care about that.

The partx call is a bit icky btw, we shouldn't need that, udev asynchronously retiggers a partition rescan anyway. There's no nice way to wait for that though... partx register partitions from userspace, and does so synchronously, which is a bit messy, since it duplicates the partition parser in userspace. The reason it helps here is because it operates synchronously...

poettering:
Force pushed a new version, only change is a rebase

PR #741, opened 2021-06-16
mkosi: don't use bsdtar on mandriva

poettering:
Let's simplify things and use GNU tar everywhere, to expose somewhat
systematic behaviour on all distros.

This mirrors what systemd's importd daemon already does: force GNU tar
use even on distros where "tar" resolves to bsdtar.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3326,6 +3326,17 @@ def compressor_command(option: Union[str, bool]) -> List[str]:
         die(f"Unknown compression {option}")
 
 
+def tar_binary() -> str:
+    # Some distros (Mandriva) install BSD tar as "tar", hence prefer
+    # "gtar" if it exists, which should be GNU tar wherever it exists.
+    # We are interested in exposing same behaviour everywhere hence
+    # it's preferable to use the same implementation of tar
+    # everywhere. In particular given the limited/different SELinux
+    # support in BSD tar and the different command line syntax
+    # compared to GNU tar.
+    return "gtar" if shutil.which("gtar") else "tar"
+
+
 def make_tar(args: CommandLineArguments, root: str, do_run_build_script: bool, for_cache: bool) -> Optional[BinaryIO]:
     if do_run_build_script:
         return None
@@ -3338,13 +3349,10 @@ def make_tar(args: CommandLineArguments, root: str, do_run_build_script: bool, f
 
     with complete_step("Creating archive"):
         f: BinaryIO = cast(BinaryIO, tempfile.NamedTemporaryFile(dir=os.path.dirname(args.output), prefix=".mkosi-"))
-        # OpenMandriva defaults to bsdtar(libarchive) which uses POSIX argument list so let's keep a separate list
-        if shutil.which("bsdtar") and args.distribution == Distribution.openmandriva:
-            cmd = ["bsdtar", "-C", root_dir, "-c", "--xattrs", "-f", "-"]
-        else:
-            cmd = ["tar", "-C", root_dir, "-c", "--xattrs", "--xattrs-include=*"]
-            if args.tar_strip_selinux_context:
-                cmd += ["--xattrs-exclude=security.selinux"]
+
+        cmd = [tar_binary(), "-C", root_dir, "-c", "--xattrs", "--xattrs-include=*"]
+        if args.tar_strip_selinux_context:
+            cmd += ["--xattrs-exclude=security.selinux"]
 
         compress = should_compress_output(args)
         if compress:


mkosi/__init__.py
```diff
@@ -3338,13 +3338,18 @@ def make_tar(args: CommandLineArguments, root: str, do_run_build_script: bool, f
...
+
+        # Some distros (Mandriva) install BSD tar as "tar", hence
+        # prefer "gtar" if it exists, which should be GNU tar wherever
+        # it exists.  We are interested in exposing same behaviour
+        # everywhere hence it's preferable to use the same
+        # implementation of tar everywhere. In particular given the
+        # limited/different SELinux support in BSD tar compared to GNU
+        # tar.
+
+        cmd = [("gtar" if shutil.which("gtar") else "tar"), "-C", root_dir, "-c", "--xattrs", "--xattrs-include=*"]
```
keszybz, line 3350: Maybe add `gtar_binary()`, like `xz_binary()`? It's nice to abstract this.

tpgxyz:
Hi, gtar is not default tar implementation here, and that why i've added support for libarchive's tar.
As you may have noticed gtar command options are not compatible with bsdtar i.e.` --xattrs-include=` and i doubt that libarchive will ever add support for that.

Bespoke turns into ready-made then.

keszybz:
> gtar command options are not compatible with bsdtar i.e.` --xattrs-include=`

Oh, right. So this will not work at all. So I think this should at least prefer `gtar` if available, because of the better feature set.

tpgxyz:
speaking of feature set, then libarchive is like systemd, it merges many tools into one piece to reduce fragmentation. Sounds familiar, right ? :)

poettering:
Well, this patch wants to streamline behaviour, and just stick to GNU tar everywhere, so that we expose the same behaviour and don't have to work around all the quirks alternative implementations require. We should make a choice, and not work around the needless differences between the distros, in particular as the Selinux stuff in bsd tar doesn't really exist apparently

To my knowledge mandriva makes GNU tar available as "gtar", no? hence this change that prefers to use "gtar" if that exists.

keszybz:
Yeah, but if `gtar` is not present, and we call `bsdtar`, it'll choke on `--xattrs-include=`, iiuc.

poettering:
well, consider it a missing dep. we also fail if "sfdisk" is not installed. I am mostly concerned that things work if all deps are fulfilled.

keszybz:
I haven't used mandriva/mandrake in two decades, so please correct me if I got this wrong, but my interpretation is that both `bsdtar` and `gtar` provide the name `tar`. Thus if `gtar` is not available, `bsdtar` will be the implementation that will appear as `tar`. If we then call that with `--xattrs-include=`, it will fail.

tpgxyz:
By default `bsdtar` is installed and yes it does provide tar on rpm and filesystem levels. So if you call `tar` (symlinked to bsdtar)  with not supported options it will fail.

I see four options here:
1. I'll carry bsdtar support as downstream patch
2. I'll add `Requires: gtar` and replace tar with gtar on downstream spec
3. You check for `gtar` first and as a fallback of that check use `tar` + fig. 1
4. You reject that PR :)

poettering:
> I haven't used mandriva/mandrake in two decades, so please correct me if I got this wrong, but my interpretation is that both `bsdtar` and `gtar` provide the name `tar`. Thus if `gtar` is not available, `bsdtar` will be the implementation that will appear as `tar`. If we then call that with `--xattrs-include=`, it will fail.

Sure, then install "tar" and it magically works. We will fail if deps are missing. I'd consider "gnu tar" a dep. On Fedora, on Suse, and on Mandriva too. As long as it just fails I think we are good.

Quite frankly, the problem Mandriva creates for itself there are so utterly pointless. If you make your tool available as "tar" then you better implement the proper switches that callers expect. If you don't then the burden for that should be on you, not everyone else. Or in other words. I'd expect them to patch the missing command line switches into their version of tar and not expect everyone else to special case their distro.

I think trying "gtar" first, and falling back to "tar" is a pretty Ok thing, since even on fedora gnu tar can be called as gtar. So the "gtar" name is more than just a one-distro-does-everything-different-thing. 

Let's also not forget that the current code in git main regarding this is actually just plain broken. It changes the command line depending on the distro picked for the image we build, not on the distro we are running on. We do support cross-distro build however, and if anyone ever tries to build a debian image on mandriva things will just fail badly: we'll put together a gnu tar command line because we are building debian, but since we are running on mandriva the command line won't work.

Seriously, let's not needlessly let our code be made complicate because of Mandriva's poor choices on this...

poettering:
Force pushed a new version, that adds `tar_binary()` as suggested

behrmann:
I think the brokenness for cross building is a good point. If @tpgxyz can just add `Requires: gtar` to Mandriva's mkosi packaging that sounds like the best solution.

tpgxyz:
Yes i'll do that as i proposed this as one of options.

behrmann (approved):

PR #742, opened 2021-06-16
mkosi: call dracut/unified kernel image on the final image, not the c…

poettering:
…ached image

We need to invoke dracut on the final image, not on the cached image.
Currently, it's done the other way round, in light of the fact that
dracut is the slowest part of the build. However that doesn't really
work, since the generated unified kernel image depends on the final
build.

The generated image after all will:

1. when verity is enabled receives the root hash which depends on the
   final fs image to be complete

2. when the image id/version logic is enabled the unified kernel image
   name will change with every build

3. we want to base the initrd on the resources installed in the OS
   during the build. Besides configuration this prominently includes a
   freshly build systemd. After all, one of the major usecases for mkosi
   was to be useful to build test images for systemd development. There
   we really want to ensure the systemd we just built is included in the
   initrd, and not an old version that happened to be around when the
   cache image was first created.

This reverts major parts of e4277c77618f47be6a5126a709f64ad48492807b

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3746,9 +3746,17 @@ def install_unified_kernel(
 
     if not args.bootable or args.esp_partno is None or not args.with_unified_kernel_images:
         return
-    if for_cache and args.verity:
-        return
-    if cached and not args.verity:
+
+    # Don't run dracut if this is for the cache. The unified kernel
+    # typically includes the image ID, roothash and other data that
+    # differs between cached version and final result. Moreover, we
+    # want that the initrd for the image actually takes the changes we
+    # make to the image into account (e.g. when we build a systemd
+    # test image with this we want that the systemd we just built is
+    # in the initrd, and not one from the cache. Hence even though
+    # dracut is slow we invoke it only during the last final build,
+    # never for the cached builds.
+    if for_cache:
         return
 
     # Don't bother running dracut if this is a development build. Strictly speaking it would probably be a


DaanDeMeyer:
We shouldn't forget that there's a ton of use cases where mkosi can be used for development that don't depend on the initrd being rebuilt on every build. e.g. I used mkosi to test some resolved changes I was making to systemd and I didn't run into issues by caused caching the dracut output. Same goes for when I worked on a patch for pacman or some KDE patches where I used mkosi for development.

I'd much prefer to add an option RebuildInitrd or so that takes no, yes and auto as arguments. The default would be auto which does what we do now: check if any options are used that require dracut to run in the final phase and if so, run dracut in the final phase. Otherwise, run it during the cache phase.

This keeps the default fast for development use cases that don't need fiddling with the initrd.

keszybz:
@poettering?

FWIW, I think we should merge this, because in general slow is better than buggy.

DaanDeMeyer:
I'm OK with this but I'd be happy to have an option like I mentioned. As long as I have a knob to turn it off and build the initrd as part of caching as happens now I'd be happy.

poettering:
I'd be happy to add an option that adds the behaviour @DaanDeMeyer wants, but I think we should default to "correct-even-if-slow" and then add "faster-but-only-safe-if-you-know-what-you-do" after that.

So I'd also merge this now, but would be happy to merge a PR laer than makes this optional

DaanDeMeyer:
Sounds good to me, will do a PR to add that option. Let's merge.

DaanDeMeyer:
I was looking at adding the option but found a better solution, we already have an option to disable usage of unified kernel images which implicitly causes the initrd to get cached as it's built as part of installing the kernel so I can just disable usage of unified kernel images to get the faster builds.

PR #744, opened 2021-06-18
Add --remove-files=/RemoveFiles=

keszybz:
This allows a post-install cleanup to be done. For example,
for initrd images:

  RemoveFiles=
        # we don't need this after the binary catalogs have been built
        /usr/lib/systemd/catalog
        /etc/udev/hwdb.d
        /usr/lib/udev/hwdb.d

As suggested in https://github.com/systemd/mkosi/pull/728#discussion_r646523736

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -617,6 +617,11 @@ details see the table below.
   instead of a directory a `tar` file may be used too, and
   `mkosi.skeleton.tar` is automatically used.
 
+`--remove-files=`
+
+: Takes a comma-separated list of globs. Files in the image matching
+  the globs will be purged at the end.
+
 `--build-script=`
 
 : Takes a path to an executable that is used as build script for this
@@ -1000,6 +1005,7 @@ which settings file options.
 | `--cache=`                        | `[Packages]`            | `Cache=`                      |
 | `--extra-tree=`                   | `[Packages]`            | `ExtraTrees=`                 |
 | `--skeleton-tree=`                | `[Packages]`            | `SkeletonTrees=`              |
+| `--remove-files=`                 | `[Packages]`            | `RemoveFiles=`                |
 | `--build-script=`                 | `[Packages]`            | `BuildScript=`                |
 | `--build-environment=`            | `[Packages]`            | `BuildEnvironment=`           |
 | `--build-sources=`                | `[Packages]`            | `BuildSources=`               |
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -4277,7 +4277,7 @@ class ListAction(argparse.Action):
             if self.delimiter == "," and values.startswith("[") and values.endswith("]"):
                 values = values[1:-1]
 
-            # Make sure delimiters between quotes are ignored by using the csv module.
+            # Make sure delimiters between quotes are ignored.
             # Inspired by https://stackoverflow.com/a/2787979.
             values = [x.strip() for x in re.split(f"""{self.delimiter}(?=(?:[^'"]|'[^']*'|"[^"]*")*$)""", values) if x]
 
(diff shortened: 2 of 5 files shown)


mkosi.md
```diff
@@ -617,6 +617,11 @@ details see the table below.
   instead of a directory a `tar` file may be used too, and
   `mkosi.skeleton.tar` is automatically used.
 
+`--remove-files=`
+
+: Takes a path glob. Files in the image matching the glob will be
```
behrmann, line 622:
```suggestion
: Takes a comma-separated list of globs. Files in the image matching the globs will be
```

mkosi/__init__.py
```diff
@@ -4498,6 +4507,15 @@ def parse_base_packages(value: str) -> Union[str, bool]:
     return parse_boolean(value)
 
 
+def parse_remove_files(value: str) -> List[str]:
+    """Normalize paths as relative to / to ensure we don't go outside of our root."""
+
+    # os.path.normpath() doesn't normalize exactly '//' (prolly
+    # inspired by Windows paths), even though it normalizes '///',
+    # so we can't just prepend '/' unconditionally.
```
behrmann, line 4515: This sent me down a little rabbit hole (after looking at the implementation of `os.path.normpath`) and TIL. :) This is actually [POSIX fault](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_13)
> If a pathname begins with two successive <slash> characters, the first component following the leading <slash> characters may be interpreted in an implementation-defined manner, although more than two leading <slash> characters shall be treated as a single <slash> character.

I haven't grepped through the kernel code, but quickly browsing the Linux Programming Interface I only find the known rule that multiple slashes are collapsed into one, implying that initial slashes are not different from later ones and this behaviour `normpath` being for compatibility with other weird unices (apparently cygwin treats them differently). 

Another quick check shows me that at least `os.unlink` doesn't care about `//foo` vs `/foo` (as it should, since the underlying libc and syscall probably won't), so I guess you *can* just unconditionally prepend a slash.

keszybz: Thanks for the correction. I'll adjust the comment to blame POSIX instead of MS.



> Another quick check shows me that at least `os.unlink` doesn't care about `//foo` vs `/foo` (as it should, since the underlying libc and syscall probably won't), so I guess you _can_ just unconditionally prepend a slash.

The syscalls don't care, but `os.path.join()` does. And also `pathlib.Path.__div__()` does, which I was very disappointed to see. It makes it very annoying to use either of the two APIs, because you have to constantly check if the path does not start with a slash or do other workarounds.

Duh, but now I see that my code is borked, because it'll mishandle a user-specified path starting with '//'. I'll rework this.

keszybz:
Adjusted.

mkosi/__init__.py
behrmann: As a side note for the problem of reparenting a path with pathlib `Path`s, I use this as an idiom `newroot / oldroot.relative_to(path)`.

mkosi/__init__.py
```diff
@@ -4255,28 +4268,25 @@ def __call__(
...
-            values = values[1:-1]
+        if isinstance(values, str):
+            # Support list syntax for comma separated lists as well
+            if self.delimiter == "," and values.startswith("[") and values.endswith("]"):
+                values = values[1:-1]
 
-        # Make sure delimiters between quotes are ignored by using the csv module.
-        # Inspired by https://stackoverflow.com/a/2787979.
-        new = re.split(f"""{self.delimiter}(?=(?:[^'"]|'[^']*'|"[^"]*")*$)""", values)
+            # Make sure delimiters between quotes are ignored by using the csv module.
```
behrmann, line 4280: I think this comment is outdated, since we don't use the `csv` module anywhere.

behrmann (commented):
LGTM

mkosi/__init__.py
keszybz: That looked promising… but both paths must be absolute, or both must be relative. So it's not useful for the case here. I think the approach with `.lstrip()` is OK. Maybe not very elegant, but it's simple and should be very robust.

Issue #725, opened 2021-05-27
Debian UEFI image boot in qemu fails at /dev/gpt-auto-root and goes to emergency console


bluca:
Since the switch to virtio_scsi from virtio_blk in https://github.com/systemd/mkosi/pull/650 I can no longer boot a Debian image built from the systemd repository in qemu:

```
[ TIME ] Timed out waiting for device /dev/gpt-auto-root.
[DEPEND] Dependency failed for Root Partition.
[DEPEND] Dependency failed for Initrd Root File System.
[DEPEND] Dependency failed for Relo…figuration from the Real Root.
[DEPEND] Dependency failed for Initrd Root Device.
[DEPEND] Dependency failed for File…m Check on /dev/gpt-auto-root.
```

rdsosreport: https://paste.debian.net/1198880/

Reverting the driver change fixes the problem. Adding the driver manually to dracut doesn't seem to help:

```
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1684,6 +1684,7 @@ def configure_dracut(args: CommandLineArguments, root: str) -> None:
     if args.hostonly_initrd:
         with open(os.path.join(dracut_dir, "30-mkosi-filesystem.conf"), "w") as f:
             f.write(f'filesystems+=" {(args.output_format.needed_kernel_module())} "\n')
+            f.write('add_drivers+=" virtio_scsi "\n')
 
     # These distros need uefi_stub configured explicitly for dracut to find the systemd-boot uefi stub.
     if args.esp_partno is not None and args.distribution in (
```

PR #745, opened 2021-06-18
qemu: revert back to virtio from virtio-scsi-pci on Debian

bluca:
qemu fails to boot with virtio-scsi-pci. Reverting back to virtio
does not show issues with the EFI shell, everything boots correctly.

Fixes: https://github.com/systemd/mkosi/issues/725

@DaanDeMeyer could you please double check if you still see the EFI shell problem with the boot order? It works fine for me on Debian 11. But booting with virtio-scsi fails as described in the issue.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -6874,14 +6874,21 @@ def run_qemu(args: CommandLineArguments) -> None:
         else:
             fname = args.output
 
-        cmdline += [
-            "-drive",
-            f"if=none,id=hd,file={fname},format={'qcow2' if args.qcow2 else 'raw'}",
-            "-device",
-            "virtio-scsi-pci,id=scsi",
-            "-device",
-            "scsi-hd,drive=hd,bootindex=1",
-        ]
+        # Debian images fail to boot with virtio-scsi, see: https://github.com/systemd/mkosi/issues/725
+        if args.distribution == Distribution.debian:
+            cmdline += [
+                "-drive",
+                f"if=virtio,id=hd,file={fname},format={'qcow2' if args.qcow2 else 'raw'}",
+            ]
+        else:
+            cmdline += [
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
I'll try this out to see if we still get the boot order issue.

bluca:
@DaanDeMeyer did you have time to check this out?

keszybz:
Once this is merged or otherwise resolved, I think we should make a release.

bluca:
Updated to only fall back to virtio for Debian images, this should avoid regressions for other distros, while fixing the issue with Debian images. PTAL!

behrmann:
LGTM, but the Debian version has an additional `id=hd` compared to the original version
```
cmdline += ["-drive", f"format={'qcow2' if args.qcow2 else 'raw'},file={fname},if=virtio"]
```
I've only quickly skimmed the qemu manpage and don't see that option listed for `-drive` or `-blockdev`, but most everything else and generic object seem to take this option, so I guess it's fine and I guess you've tried it, so it's probably good.

bluca:
> LGTM, but the Debian version has an additional `id=hd` compared to the original version
> 
> ```
> cmdline += ["-drive", f"format={'qcow2' if args.qcow2 else 'raw'},file={fname},if=virtio"]
> ```
> 
> I've only quickly skimmed the qemu manpage and don't see that option listed for `-drive` or `-blockdev`, but most everything else and generic object seem to take this option, so I guess it's fine and I guess you've tried it, so it's probably good.

Yeah, it boots fine, probably just not used when there's no -drive + -device combination

DaanDeMeyer:
@bluca I can't seem to reproduce the bootindex problem anymore. I think the current solution is fine though as supposedly virtio-scsi-pci is the new hotness so I guess it makes sense to use it where we can.

PR #746, opened 2021-06-18
Debian: fallback to user/host network in qemu/nspawn with no networkd, manually enable resolved

bluca:


mkosi/__init__.py
```diff
@@ -6750,7 +6750,9 @@ def has_networkd_vm_vt() -> bool:
     return False
 
 
-def ensure_networkd(args: CommandLineArguments) -> None:
+def ensure_networkd(args: CommandLineArguments) -> int:
```
DaanDeMeyer, line 6759: Can't we just return a boolean?

mkosi/__init__.py
```diff
@@ -6948,15 +6954,19 @@ def run_qemu(args: CommandLineArguments) -> None:
...
-        # vt-<image-name> is the ifname on the host and is automatically picked up by systemd-networkd which
-        # starts a DHCP server on that interface. This gives IP connectivity to the VM. By default, QEMU
-        # itself tries to bring up the vt network interface which conflicts with systemd-networkd which is
-        # trying to do the same. By specifiying script=no and downscript=no, We tell QEMU to not touch vt
-        # after it is created.
-        cmdline += ["-nic", f"tap,script=no,downscript=no,ifname={ifname},model=virtio-net-pci"]
+        networkd_configured = ensure_networkd(args) == 0
+        if not networkd_configured:
+            # Fall back to usermode networking if the host doesn't have networkd (eg: Debian)
+            cmdline += ["-nic", f"user,model=virtio-net-pci"]
```
DaanDeMeyer, line 6966: I always find automatic fallback awkward, since a user might explicitly want networkd to be used or have no networking at all otherwise. should we make this an explicit option?

DaanDeMeyer (changes requested):

mkosi/__init__.py
bluca: There's a big red warning when networkd is enabled, that's should be enough? In most cases you don't want to change the default network management configuration of your distribution, as chances are it's not supported and things will break. Having things "just work" would be nice. The nice thing is that the fallback will _always_ work, regardless of system settings or anything else, and doesn't require any root  privileges, as it's entirely self-contained in qemu. So it's pretty safe to use.

mkosi/__init__.py
bluca: ok, updated

mkosi/__init__.py
bluca:
```
‣ Warning: 
            --network-veth requires systemd-networkd to be running to initialize the host interface of the
            veth link (`systemctl enable --now systemd-networkd`)")
```

DaanDeMeyer:
@bluca I'm blanking on why we enable resolved for debian when we don't do so for any of the other distros. Any objection against removing that again?

bluca:
Others enable it by default, Debian doesn't, so there's no DNS otherwise

DaanDeMeyer:
> Others enable it by default, Debian doesn't, so there's no DNS otherwise

Others == Fedora? Arch definitely doesn't, and I don't think centos does either. I'd prefer to leave this stuff up to the distro instead of doing this in mkosi

bluca:
Debian is not going to change those defaults anytime soon

DaanDeMeyer:
That's true, but neither are Arch or centos and we don't special case those either so I'd like to bring Debian in line with those distros.

bluca:
But that would mean networking is broken by default and everyone would have to use the same postinst to fix it... What problem is this causing exactly?

DaanDeMeyer:
The problem is that we're not being consistent across distros. We should either enable resolved everywhere, or don't touch it anywhere. Right now, we don't touch it anywhere except Debian.

Sure DNS will be disabled by default but we've never documented that the images we produce will have networking enabled and working. I don't want to force a specific network configuration onto users. Users should not have to disable resolved and enable their dns daemon of choice, they should just have to enable their DNS daemon of choice to have things work.

I don't mind gating this behind the --netdev option since that already implies a bunch of custom networkd configuration on our end so we can add the resolved configuration behind that as well. But by default, I'd like us to not touch DNS daemons anywhere.

bluca:
Ubuntu, Fedora (maybe SUSE too?) already use resolved by default, which is good. Enable it by default in Arch and Centos too then? I don't think it makes any sense to build images that are not usable out of the box because networking is broken, and I don't think it makes sense to build images that don't use systemd by default - after all this is part of the systemd project, to showcase and make use of all the systemd components. There are already many other tools to build Debian images using the crufty 30 years old networking config shell scripts that are the default there, and if someone wants to use those in an mkosi-built image, they certainly can via postprocessing

behrmann:
I think I'm with @bluca on this. This will be something that people will need to reimplement over and over if they want to have a useful Debian image. An alternative I could see to pushing this into all distros would be to maybe add a new `mkosi template` (or `mkosi new-project` or...) verb that templates a directory with mkosi config files. There we could also drop scripts most everybody will need, but somebody might want to opt out of, without proliferating our already stately collection of config arguments.

DaanDeMeyer:
Does everyone actually use resolved on Debian then? I thought it had its own thing.

Even if we do this, shouldn't we use a preset for this so that it can be easily overriden?

bluca:
Yeah Debian doesn't use networkd nor resolved by default - it uses the old crufty ifupdown scripts&friends. That's why this is needed. There's also no preset usage in Debian, not sure if any infrastructure is needed for it - if it just works, then sure, that sounds fine.

PR #747, opened 2021-06-24
Add option to resolve symlinks when copying build sources

bcbnz:
Currently, the `copy-all` method of copying build sources copies any symlinks into the image. This pull request adds a `--source-file-resolve-symlinks` option (and a corresponding `-final` option for copying into the final image) which corresponds to setting `symlinks=False` when calling `shutil.copytree`. If sets, this means the symlink is resolved and the contents of the file are copied to the image rather than the symlink itself. The option defaults to off so the behaviour is backwards-compatible.

For my use case, I have some utility scripts I want to reuse when generating different images; this change allows me to have them in one location and symlink them when needed, and then call them from the main build script.

### DIFF
--- a/man/mkosi.1
+++ b/man/mkosi.1
@@ -769,6 +769,18 @@ Takes the same values as \f[C]--source-file-transfer\f[R] except
 \f[C]mount\f[R].
 By default, sources are not copied into the final image.
 .TP
+\f[B]\f[CB]--source-resolve-symlinks\f[B]\f[R]
+If given, any symbolic links in the source file tree are resolved and
+the file contents are copied to the build image.
+If not given, they are left as symbolic links.
+This only applies if \f[C]--source-file-transfer\f[R] is
+\f[C]copy-all\f[R].
+Defaults to leaving them as symbolic links.
+.TP
+\f[B]\f[CB]--source-resolve-symlinks-final\f[B]\f[R]
+Same as \f[C]--source-resolve-symlinks\f[R] but for the final image
+instead of the build image.
+.TP
 \f[B]\f[CB]--with-network\f[B]\f[R]
 Enables network connectivity while the build script
 \f[C]mkosi.build\f[R] is invoked.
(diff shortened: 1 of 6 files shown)


man/mkosi.1
```diff
@@ -769,6 +769,16 @@ Takes the same values as \f[C]--source-file-transfer\f[R] except
 \f[C]mount\f[R].
 By default, sources are not copied into the final image.
 .TP
+\f[B]\f[CB]--source-file-resolve-symlinks\f[B]\f[R]
```
behrmann, line 772: I think we really need to make a note that, this file is autogenerated from `mkosi.md`, so that people don't go through the pain of editing raw roff :)

mkosi/__init__.py
```diff
@@ -3259,10 +3259,13 @@ def install_build_src(args: CommandLineArguments, root: str, do_run_build_script
             copy_file(args.build_script, os.path.join(root_home(args, root), os.path.basename(args.build_script)))
 
     sft: Optional[SourceFileTransfer] = None
+    resolve_symlinks: bool = None
```
behrmann, line 3262:
```suggestion
    resolve_symlinks: bool = False
```
The less `None`s the better.

behrmann (commented):
Thank you for your contribution!

man/mkosi.1
bcbnz: Damn! pandoc adds a note at the top which I totally missed. Though I would wonder if an automatically generated file should be tracked in Git... anyhow, I have added the changes to mkosi.md

mkosi/__init__.py
bcbnz: Done.

bcbnz:
Force-pushed a squashed commit to resolve the above comments. I also renamed the options from `--source-file-resolve-symlinks[-final]` to `--source-resolve-symlinks[-final]` to make them (slightly) shorter since on reflection I figured the `file` was redundant.

Issue #750, opened 2021-07-01
[archlinux] Pacman hook incorrectly named


aw1cks:
Pacman hooks must end with `.hook` as per [Arch wiki](https://wiki.archlinux.org/title/pacman#Hooks)

However mkosi's hook to update bootctl when the kernel updates ends in `-hook` instead of `.hook` (likely a typo)
This means the hook doesn't get fired and kernel upgrades fail - see [this line](https://github.com/systemd/mkosi/blob/main/mkosi/__init__.py#L2580)

aw1cks:
Above PR should fix this issue :thumbsup:

PR #751, opened 2021-07-01
Correct pacman hook name

aw1cks:
Fixes systemd/mkosi#750

Pacman hooks must end with `.hook` as per [Arch wiki](https://wiki.archlinux.org/title/pacman#Hooks)

However mkosi's hook to update bootctl when the kernel updates ends in `-hook` instead of `.hook` (likely a typo)
This means the hook doesn't get fired and kernel upgrades fail - see [this line](https://github.com/systemd/mkosi/blob/main/mkosi/__init__.py#L2580)

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2577,7 +2577,7 @@ def install_arch(args: CommandLineArguments, root: str, do_run_build_script: boo
         make_executable(kernel_remove_script)
 
         if args.esp_partno is not None:
-            bootctl_update_hook = os.path.join(hooks_dir, "91-mkosi-bootctl-update-hook")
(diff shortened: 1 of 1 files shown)


keszybz:
Please don't create issues and then submit a pull request to immediately fix the issue… Just open the pull request.

Please move the descripition of the patch into the patch message.

aw1cks:
> Please don't create issues and then submit a pull request to immediately fix the issue… Just open the pull request.
> 
> Please move the descripition of the patch into the patch message (from the issue).

Apologies, updated now...
N.B. there is no contribution guide for this repo, just some feedback as a first time contributor, if you have such rules perhaps they could be made clearer. Cheers

aw1cks:
> I don't see the update? Please update the _commit message_.

Commit message updated, kindly review again.

PR #760, opened 2021-07-30
Various cleanups without functional changes

keszybz:


mkosi/__init__.py
```diff
@@ -6739,49 +6743,48 @@ def find_qemu_binary() -> str:
...
-        if os.path.exists(firmware):
+        if firmware and os.path.exists(firmware):
             return firmware, True
 
     warn(
-        """\
-        Couldn't find OVMF firmware blob with secure boot support,
-        falling back to OVMF firmware blobs without secure boot support.
-        """
+        "Couldn't find OVMF firmware blob with secure boot support, "
```
behrmann, line 6763: This removes a linebreak from the message.

mkosi/__init__.py
```diff
@@ -6739,49 +6743,48 @@ def find_qemu_binary() -> str:
...
+        # After that, we try some generic paths and hope that if they exist,
+        # they’ll correspond to the current architecture, thanks to the package manager.
+        "/usr/share/edk2/ovmf/OVMF_CODE.fd",
+        "/usr/share/qemu/OVMF_CODE.fd",
+        "/usr/share/ovmf/OVMF.fd",
+    ]
 
     for firmware in FIRMWARE_LOCATIONS:
-        if os.path.exists(firmware):
+        if firmware and os.path.exists(firmware):
```
behrmann, line 6787: the check for truthiness of `firmware` is not necessary here, since if the `.get` falls back to an empty list. I'd actually prefer that way for the first `FIRMWARE_LOCATIONS`, too.

I think that's also what mypy was complaining about (what the last commit addresses), since the `.get` could produce a `None` and that cannot be expanded with `*`.

mkosi/__init__.py
keszybz: Yeah, on purpose — it's generally better to have error messages on one line. If on a terminal, it'll be wrapped automatically.

mkosi/__init__.py
```diff
@@ -2438,6 +2438,23 @@ def run_pacman(root: str, pacman_conf: str, packages: Set[str]) -> None:
         run(["gpgconf", "--homedir", os.path.join(root, "etc/pacman.d/gnupg"), "--kill", "all"])
 
 
+def enable_locale(args: CommandLineArguments, root: str) -> None:
```
poettering, line 2441: enable_locale() sounds so generic. But this is specific to Debian and a few other distros. Maybe call this function patch_locale_gen() or so, since that's what it does?

mkosi/__init__.py
```diff
@@ -2438,6 +2438,23 @@ def run_pacman(root: str, pacman_conf: str, packages: Set[str]) -> None:
...
+    # If /etc/locale.gen exists, uncomment the desired locale and leave the rest of the file untouched.
+    # If it doesn’t exist, just write the desired locale in it.
+    try:
+
+        def _patch_line(line: str) -> str:
+            if line.startswith("en_US.UTF-8"):
```
poettering, line 2447: you lost the # here?

mkosi/__init__.py
```diff
@@ -6743,49 +6743,47 @@ def find_qemu_binary() -> str:
...
-    FIRMWARE_LOCATIONS.append("/usr/share/edk2/ovmf/OVMF_CODE.secboot.fd")
-    FIRMWARE_LOCATIONS.append("/usr/share/qemu/OVMF_CODE.secboot.fd")
-    FIRMWARE_LOCATIONS.append("/usr/share/ovmf/OVMF.secboot.fd")
+    FIRMWARE_LOCATIONS = [
+        # UEFI firmware blobs are found in a variety of locations,
+        # depending on distribution and package.
+        {
+            "x86_64": "/usr/share/ovmf/x64/OVMF_CODE.secboot.fd",
+            "i386": "/usr/share/edk2/ovmf-ia32/OVMF_CODE.secboot.fd",
+        }.get(platform.machine()),
```
poettering, line 6752: what happens on platforms other than x8664/i386 with this expression? Doesn't this resolve to None then? Is that desirable?

mkosi/__init__.py
keszybz: Indeed, I messed this up and I thought mypy is to blame. It's fixed now.

keszybz:
Updated to address all the issues.

poettering:
lgtm

behrmann:
LGTM.

PR #761, opened 2021-07-30
fedora: use F34 by default

keszybz:
F32 is no more… Also switch to btrfs, to match
https://fedoraproject.org/wiki/Changes/BtrfsByDefault.

It would be nice to enable compression too to match
https://fedoraproject.org/wiki/Changes/BtrfsTransparentCompression,
but that's more involved because mount options have to be set.

### DIFF
--- a/mkosi.files/mkosi.fedora
+++ b/mkosi.files/mkosi.fedora
@@ -3,7 +3,7 @@
 
 [Distribution]
 Distribution=fedora
-Release=32
+Release=34
 
 [Output]
 Format=gpt_ext4


poettering:
what i don't like about the btrfs default is that it's uuid handling is so messy. you can't really operate with multiple copies of the resulting image since the kernel notices the uuid conflict: there'd be two mounted fs with the same uuid. It's really stupid. ext4 doesn't have restrictions like that. and the errors are annoying to deal with.

it kind made me step away with toying around with btrfs images. It's fine for the final product, but while testing stuff it just is annoying, in particular as the uuid checking is asynchronous in the btrfs kernel code, so even if you unmount everything you might still run into conflicts.

behrmann:
LGTM, CI failure is unrelated.

keszybz:
I dropped the fs change. Yeah, if ext4 works better, let's use that for now.

poettering:
lgtm

Issue #664, opened 2021-03-01
photon is completely broken


ssahani:
```
`commit 0f167a2205070096e182222a84ddc954a6c23997
Author: Daan De Meyer <daan.j.demeyer@gmail.com>
Date:   Wed Dec 2 19:57:48 2020 +0000

```

@DaanDeMeyer your commit broke photon

/cc @smaliakkal

behrmann:
That commit only removed CI for Photon. Could you describe what the problem is, on which version of mkosi it appears, how you call mkosi (i.e. provide a minimal example to reproduce this issue) and a version where this last worked for you? If you could bisect the specific commit that broke you, that would be tremendously helpful, but the other infos would be a great start.

ssahani:
See this a8a14b8bfe39d91f933062667877bfc0de686541.  Have you tested even before you make. changes ?

DaanDeMeyer:
Unfortunately, half the features didn't even work when I tried (for a very long time) to add Photon to the CI. I even pinged you for help to get Photon working in CI but you never responded so we didn't add it to CI so we can't guarantee it keeps working. If you do the work of maintaining the photon implementation and put in the effort to add it to the CI (full feature set including QEMU and nspawn boot), I'll happily make sure it stays working.

But you have to understand that we just can't make progress if every contributor has to manually test every niche distro supported by mkosi without getting any support from the person that originally added it.

TLDR; Make PRs to fix all the issues with Photon and add Photon to mkosi CI and we'll make sure it doesn't break again.

behrmann:
Back then there was no CI for Photon and CI for Photon was only added in f06b436899ea4502634e868e64ab53ee5407db08 (by @DaanDeMeyer), but removed again in 0f167a2205070096e182222a84ddc954a6c23997, since it was too much of a maintenance burden and generated images already didn't boot (if I remember correctly). Neither is the fault of @DaanDeMeyer. 

If you provide a reproducer, we can still have a look what the issue is.

ssahani:
> Back then there was no CI for Photon and CI for Photon was only added in [f06b436](https://github.com/systemd/mkosi/commit/f06b436899ea4502634e868e64ab53ee5407db08) (by @DaanDeMeyer), but removed again in [0f167a2](https://github.com/systemd/mkosi/commit/0f167a2205070096e182222a84ddc954a6c23997), since it was too much of a maintenance burden and generated images already didn't boot (if I remember correctly). Neither is the fault of @DaanDeMeyer.
[...]

PR #762, opened 2021-08-03
photon: switch mirror to packages.vmware.com

Finkregh:
following <https://kb.vmware.com/s/article/82322> the repos have been switched to packages.vmware.com


behrmann:
LGTM, but I fear photon might still be broken beyond that, since I haven't gotten around to having a look what in the photon code path broke, see #664

Finkregh:
thanks, i just saw that the resulting image does not boot :(

behrmann:
@Finkregh Sorry about that, Photon isn't covered by CI and got broken by mistake and so far nobody has found the time to fix this. If you don't use any of the features added to mkosi since Photon was added, you can just go back to an older tag and rebase this fix to that. Sorry about the inconvenience.

PR #764, opened 2021-08-05
Various cleanups and refactorings

keszybz:


bluca (approved):
nice cleanups - lgtm!

mkosi/__init__.py
```diff
@@ -4190,96 +4210,69 @@ def link_output(args: CommandLineArguments, root: str, artifact: Optional[Binary
...
-
-    suffix = roothash_suffix(args.usr_only)
-    with complete_step(f"Linking {suffix} file", "Successfully linked " + args.output_root_hash_file):
-        _link_output(args, root_hash_file, args.output_root_hash_file)
-
+def link_output_nspawn_settings(args: CommandLineArguments, path: Optional[SomeIO]) -> None:
+    if path:
+        assert args.output_nspawn_settings
+        with complete_step("Linking nspawn settings file…", f"Successfully linked {args.output_nspawn_settings}"):
+            _link_output(args, path.name, args.output_nspawn_settings)
```
behrmann, line 4217: I rather prefer checking the bailout condition first and returning instead of adding more indentation, but I guess that's just a matter of preference.

mkosi/__init__.py
```diff
@@ -6240,28 +6227,37 @@ def setup_network_veth(args: CommandLineArguments, root: str, do_run_build_scrip
         run(["systemctl", "--root", root, "enable", "systemd-networkd"])
 
 
+class BuildOutput(NamedTuple):
```
behrmann, line 6230: I think we should go here for a dataclass as well.

mkosi/__init__.py
```diff
@@ -6240,28 +6227,37 @@ def setup_network_veth(args: CommandLineArguments, root: str, do_run_build_scrip
...
+    sshkey: Optional[TextIO]
+    split_root: Optional[BinaryIO]
+    split_verity: Optional[BinaryIO]
+    split_kernel: Optional[BinaryIO]
+
+    def raw_name(self) -> Optional[str]:
+        return self.raw.name if self.raw is not None else None
+
+    @classmethod
+    def Empty(cls) -> "BuildOutput":
```
behrmann, line 6243:
```suggestion
    def empty(cls) -> "BuildOutput":
```

keszybz: I assumed that the convention is to use class-like naming for "alternate constructors". But looking online, there seems to be no such rule ;)

mkosi/__init__.py
behrmann: True, this repetition has irked me before, too. Maybe some templating?
```python
def _link_output_template(
    args: CommandLineArguments,
    thing: str,
    pretty_name: Optional[str] = None,
): -> None:
    def _link_output_file(args: CommandLineargument, path: Optiona[str]):
        if path is None:
            return None

        output = getattr(args, f"output_{thing}")
        assert output is not None

        with complete_step(f"Linking {pretty_name or thing} file", f"Successfully linked {output}"):
            _link_output(args, path, output)

    return _link_output_file

link_output_checksum = _link_output_template("checksum", "SHA256SUMS")
```
(Haven't tested this, just an idea)

mkosi/__init__.py
```diff
@@ -7016,7 +6982,7 @@ def generate_secure_boot_key(args: CommandLineArguments) -> NoReturn:
                 )
             )
 
-    MkosiPrinter.print_step(f"Generating secure boot keys rsa:{keylength} for CN `{cn}`.")
+    MkosiPrinter.print_step(f"Generating secure boot keys rsa:{keylength} for CN {cn!r}.")
```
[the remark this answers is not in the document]
keszybz: Yeah, the comment was supposed to be half joking ;) I think it's better not to use backticks because they are special in the shell and make select-paste harder.

mkosi/__init__.py
keszybz: Those function are similar… but not identical. So the template would be quite complex. I don't think this is worth it.

mkosi/__init__.py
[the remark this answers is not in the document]
behrmann: I'm not disagreeing :) What I actually wanted to say is, that we should then remove them in the other cases as well.

mkosi/__init__.py
behrmann: Yeah, you're right. The ones for roothash and ssh keys are different. Adding hooks back, front and centre to make this work would just be unwieldy.

mkosi/__init__.py
```diff
@@ -1171,16 +1165,35 @@ def luks_setup_tmp(args: CommandLineArguments, loopdev: str, do_run_build_script
         return None
     assert args.passphrase is not None
 
-    with complete_step("Opening LUKS temporary data partition"):
+    with complete_step("Opening LUKS temporary data partition…"):
         return luks_open(partition(loopdev, args.tmp_partno), args.passphrase)
 
 
+class LuksSetupOutput(NamedTuple):
```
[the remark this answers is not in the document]
keszybz: For BuildOutput, dataclass works just fine. But here it is problematic: I want to iterate over the fields, and with a dataclass I'd need to provide `__iter__()` and `__getitem__()`. I added that, but with the typing annotations it makes the class twice as long… So I'll reverted those changes. NamedTuple isn't very nice to print or work with, but for the limited use case here it seems to fit nicely.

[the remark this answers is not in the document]
behrmann: That's a good reason, thanks for the comment in the commit message!

mkosi/__init__.py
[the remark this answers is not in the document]
keszybz: Oh! I pushed two more commits to remove the rest of the backticks.

keszybz:
I'll merge this myself, without waiting for further reviews, because I want to push a follow-up pull request, and there'd be too many conflicts.

mkosi/backend.py
```diff
@@ -561,20 +561,23 @@ class MkosiPrinter:
...
 
+    level = 0
+
     @classmethod
     def _print(cls, text: str) -> None:
         cls.out_file.write(text)
 
     @classmethod
     def print_step(cls, text: str) -> None:
+        prefix = cls.prefix + ' ' * cls.level
```
poettering: might be pretty to draw unicode `╰` character here, to present this is a tree structure?

PR #765, opened 2021-08-07
Better --environment setting

keszybz:


mkosi/__init__.py
```diff
@@ -3180,7 +3182,8 @@ def run_finalize_script(args: CommandLineArguments, root: str, do_run_build_scri
     verb = "build" if do_run_build_script else "final"
 
     with complete_step("Running finalize script…"):
-        env = collections.ChainMap({"BUILDROOT": root, "OUTPUTDIR": output_dir(args)}, os.environ)
+        env = {v.partition("=")[0]: v.partition("=")[1] for v in args.environment}
+        env = collections.ChainMap(dict(BUILDROOT=root, OUTPUTDIR=output_dir(args)), env, os.environ)
```
DaanDeMeyer, line 3186: I'm not entirely sure if using the `dict` builder is common practice compared but I don't mind too much either.

DaanDeMeyer (approved):
LGTM, I'll leave this open for a little longer so others can take a look as well

mkosi/__init__.py
keszybz: It seemed a bit more readable to me in this case, but if the other form is preferred, I can change it back.

BTW, I think I forgot to mention this in the commit message: the construct with `v.partition()` is not very elegant, but mypy was unhappy with other forms I tried. In particular it doesn't seem to understand `.split(…, maxsplit=1)` which would be more natural to use here.

behrmann: A maybe somewhat dirty trick (because it's lying to the type checker), would be this:
```python
env = dict(cast(Tuple[str, str], v.split("=", maxsplit=1)) for v in args.environment)
```
It's obviously not true, but the dict constructor accepts any iterable of length two, but I don't think this information can be encoded in the type system (at the moment). This avoids the partition construction.

behrmann (commented):
LGTM, but maybe add a note to the changelog, that `--build-environment` is deprecated in favour of `--environment`.

keszybz:
Updated with the two suggestions: `cast()` and a NEWS entry.

behrmann (approved):
LGTM

NEWS.md
```diff
@@ -1,5 +1,11 @@
 # mkosi Changelog
 
+## v11 (unreleased)
+
+- The `--build-environemnt` option was renamed to `--environment` and
```
poettering, line 5: typo. environemnt → environment

poettering: please suffix with "="

poettering:
(we should pick up environment vars from some file `mkosi.environment` or so btw)

PR #766, opened 2021-08-08
Fix pyright errors + Add pyright to CI

DaanDeMeyer:


mkosi/__init__.py
```diff
@@ -1197,6 +1197,7 @@ def luks_setup_all(
         return
     assert loopdev is not None
 
+    root = home = srv = var = tmp = None
```
keszybz, line 1200: Hmm, this is a fix in the sense that e.g. `root` could be undefined in the `finally` block. But I don't think it's the correct fix, because the code is "off-by-one", the `finally` blocks are attached one level too high. It should be something like this:
```python
root = luks_setup_root(args, loopdev, do_run_build_script)
try:
     …
finally:
     luks_close(root, "Closing LUKS root partition")
```

But I think a better solution would be to define a context manager and do this properly… We should be able to get rid of this crazy nesting then.

mkosi/__init__.py
```diff
@@ -5697,7 +5698,7 @@ def load_args(args: argparse.Namespace) -> CommandLineArguments:
         args.cache_pre_dev = None
         args.cache_pre_inst = None
 
-    args.output = os.path.abspath(args.output)
+    args.output = os.path.abspath(str(args.output))
```
keszybz, line 5701: This doesn't look correct. `args.output` is a string already, no? And if it *wasn't* a string, this conversion would only hide the issue.

mkosi/__init__.py
DaanDeMeyer: Turned `luks_open()` into a context manager. Should be a lot cleaner now.

DaanDeMeyer:
mypy should be using the same python version for type checking as the python interpreter it's running under but for some reason that's not happening here. Specifying it explicitly with --python-version allows us to use Python 3.7 features and we don't have to install types-dataclasses anymore.

mkosi/__init__.py
DaanDeMeyer: This seemingly fixed itself in one of the newer pyright releases it seems. Removed the `str` again

behrmann:
LGTM, I really like the rework of the luks stuff as a context manager!

mkosi/__init__.py
```diff
@@ -1100,73 +1079,97 @@ def luks_format_tmp(args: CommandLineArguments, loopdev: str, do_run_build_scrip
...
+            run(["cryptsetup", "close", path])
+
+
 def luks_setup_root(
     args: CommandLineArguments, loopdev: str, do_run_build_script: bool, inserting_generated_root: bool = False
-) -> Optional[str]:
+) -> ContextManager[Optional[str]]:
     if args.encrypt != "all":
-        return None
+        return contextlib.nullcontext()
```
keszybz, line 1107: Oh, I didn't know about `nullcontext`.

mkosi/__init__.py
```diff
@@ -1197,34 +1200,20 @@ def luks_setup_all(
...
-        finally:
-            luks_close(home, "Closing LUKS home partition")
-    finally:
-        luks_close(root, "Closing LUKS root partition")
+    with contextlib.ExitStack() as stack:
+        root = stack.enter_context(luks_setup_root(args, loopdev, do_run_build_script))
+        home = stack.enter_context(luks_setup_home(args, loopdev, do_run_build_script))
+        srv = stack.enter_context(luks_setup_srv(args, loopdev, do_run_build_script))
+        var = stack.enter_context(luks_setup_var(args, loopdev, do_run_build_script))
+        tmp = stack.enter_context(luks_setup_tmp(args, loopdev, do_run_build_script))
```
keszybz, line 1208: We should be able to do this as
```py
with (luks_setup_root(args, loopdev, do_run_build_script) as root,
         luks_setup_home(args, loopdev, do_run_build_script) as home,
         luks_setup_srv(args, loopdev, do_run_build_script) as srv,
         luks_setup_var(args, loopdev, do_run_build_script) as var,
         luks_setup_tmp(args, loopdev, do_run_build_script) as tmp):
             yield …
```

DaanDeMeyer: Only from python 3.10 onwards.

behrmann: So only three more years, when we target Debian stable. :)

Issue #700, opened 2021-04-16
Add support for pinning package versions


poettering:
I think it would make sense to optionally support "pinning" package versions, i.e. write the used RPM/DEB package versions to some file while doing an initial build, and if that file exists, use exactly these versions to build the package. The file could then be commited to git, ensuring that images are somehwat stabe in their package selection.

This should bring us a step closer towards reasonably reproducible builds.

I figure this means, at least for the Fedora case, to call directly into the python dnf module to retrieve these package lists.

poettering:
Also see #748

keszybz:
FWIW, just putting specific package versions in the manifest already works, to some extent. You can specify a full package nevra in `Packages=`, and dnf will honour that. I also use this to install locally built packages: when a path with slashes is given, dnf will install that package from disk.

We could retrieve the package lists, even post-factum by introspecting the rpm database or parsing the logs.

But the hard part is retrieving the specific versions of packages from somewhere. Packages in the distro stop being available from the mirrors after after a new version appears. It is possible to retrieve them with a bit of an effort (e.g. in Fedora's case, directly from koji). So pinning of a specific version would work if there is a local cache, or for a while as long as the distro doesn't update, but then it would break.

Foxboron:
Does RH/Fedora have a package archive? For both Arch Linux and Debian all released packages are uploaded and stored on the archive. Recreating any ISO from the past couple of years should be possible by just fetching them and passing them as arguments to `pacstrap`.

https://archive.archlinux.org/

How this is implemented in the current Arch Linux repro tooling; https://github.com/archlinux/archlinux-repro/blob/master/repro.in#L490

behrmann:
No Fedora expert, but https://archives.fedoraproject.org/pub/archive/fedora/linux/releases/ looks similar to the Debian archives. No idea though, whether this can be directly hooked into `dnf` (or `yum` for the more ancient releases there).

DaanDeMeyer:
I'm not sure if this should be implemented by mkosi tbh. Users can either use --cacheonly to never hit the mirrors again after the initial download, which effectively pins the package versions. The other option is to setup your own mirror with the versions of the package that you want and then configure mkosi to use that mirror. That seems like it would scale much better than pinning individual package versions.

Closing this for now but we can reopen if needed.

pbjhelmert:
[...]

PR #768, opened 2021-08-10
mkosi: record package list and changelogs in files

### LINKED ISSUE #748: Idea: add .packages file similar to what kiwi-ng generates
Kiwi-ng has this nice property that it generates sidecar files that provide metadata about the generated image, see https://osinside.github.io/kiwi/image_types_and_results.html#image-results.

In particular I think the following are interesting:
- `<image-name>.<arch>-<version>.packages`: a sorted list of the packages that are included into the image. In fact this is normalized dump of the package manager database. It follows the following cvs format where each line is represented by: `<name>|<epoch>|<version>|<release>|<arch>|<disturl>|<license>`.
  The format is somewhat clunky, but it works. It could be useful to generate the same for compatibility. Or we could generate a json file.
- `*.changelog`: a concatenated changelog of rpms in the image.

Such metadata makes it easier to figure out "what changed" between image builds. In particular, when packages are pulled through transitive deps it can be very helpful.

keszybz:
The output is something like this

```
{
  "packages": [
    {
      "name": "acl",
      "version": "2.3.1",
      "release": "2.fc35",
      "changelog": "* Wed Jul 21 2021 Fedora Release Engineering <releng@fedoraproject.org> - 2.3.1-2\n..."
    },
    {
      "name": "alternatives",
      "epoch": "3",
      "version": "1.19",
      "release": "1.fc35",
      "changelog": "* Fri Jul 23 2021 Jan Macku <jamacku@redhat.com> - 1.19-1\n..."
    },
    ...
 ]
}
```

Changelogs are identical for all packages built from a given srpm,
so the changelog field is only populated for the first of those.

Fixes #748.

Maybe it would make sense to also do a text format so it's nicer to diff?

Docs and some config options are missing… I'll add later when there's consensus on functionality.

### DIFF
--- a/man/mkosi.1
+++ b/man/mkosi.1
@@ -341,6 +341,13 @@ system), \f[C]gpt_btrfs\f[R] (similar, but with an btrfs file system),
 \f[C]plain_squashfs\f[R] (a plain squashfs file system without a
 partition table).
 .TP
+\f[B]\f[CB]--manifest-format=\f[B]\f[R]
+The manifest format type or types to generate.
+A comma-delimited list consisting of \f[C]json\f[R] (the standard JSON
+output format that describes the packages installed),
+\f[C]changelog\f[R] (a human-readable text format designed for diffing).
+Defaults to \f[C]json\f[R].
+.TP
 \f[B]\f[CB]--output=\f[B]\f[R], \f[B]\f[CB]-o\f[B]\f[R]
 Path for the output image file to generate.
 Takes a relative or absolute path where the generated image will be
@@ -753,8 +760,8 @@ the local directory it is automatically used for this purpose (also see
 below).
 .TP
 \f[B]\f[CB]--environment=\f[B]\f[R]
-Adds environment variables to the environment that the build script is
-executed with.
+Adds variables to the environment that the
+build/prepare/postinstall/finalize scripts are executed with.
 Takes a space-separated list of variable assignments or just variable
 names.
 In the latter case, the values of those variables will be passed through
@@ -1151,6 +1158,13 @@ T}@T{
 \f[C][Output]\f[R]
 T}@T{
 \f[C]Format=\f[R]
+T}
+T{
+\f[C]--manifest-format=\f[R]
+T}@T{
+\f[C][Output]\f[R]
+T}@T{
+\f[C]ManifestFormat=\f[R]
 T}
 T{
 \f[C]--output=\f[R], \f[C]-o\f[R]
--- a/mkosi.md
+++ b/mkosi.md
@@ -342,6 +342,13 @@ details see the table below.
   squashfs file system), `plain_squashfs` (a plain squashfs file
   system without a partition table).
 
+`--manifest-format=`
+
+: The manifest format type or types to generate. A comma-delimited
+  list consisting of `json` (the standard JSON output format that
+  describes the packages installed), `changelog` (a human-readable
+  text format designed for diffing). Defaults to `json`.
+
 `--output=`, `-o`
 
 : Path for the output image file to generate. Takes a relative or
@@ -1087,6 +1094,7 @@ which settings file options.
 | `--mirror=`, `-m`                 | `[Distribution]`        | `Mirror=`                     |
 | `--architecture=`                 | `[Distribution]`        | `Architecture=`               |
 | `--format=`, `-t`                 | `[Output]`              | `Format=`                     |
+| `--manifest-format=`              | `[Output]`              | `ManifestFormat=`             |
 | `--output=`, `-o`                 | `[Output]`              | `Output=`                     |
 | `--output-split-root=`            | `[Output]`              | `OutputSplitRoot=`            |
 | `--output-split-verity=`          | `[Output]`              | `OutputSplitVerity=`          |
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -87,6 +87,7 @@ from .backend import (
     workspace,
     write_grub_config,
 )
+from .manifest import Manifest
 
 complete_step = MkosiPrinter.complete_step
 
@@ -4292,6 +4293,23 @@ def dir_size(path: str) -> int:
     return dir_sum
 
 
+def save_manifest(args: CommandLineArguments, manifest: Manifest) -> None:
+    if manifest.has_data():
+        with complete_step(f"Saving manifest {args.output}.manifest"):
+            f: TextIO = cast(
+                TextIO,
+                tempfile.NamedTemporaryFile(
+                    mode="w+",
+                    encoding="utf-8",
+                    prefix=".mkosi-",
+                    dir=os.path.dirname(args.output),
+                ),
+            )
+            with f:
+                manifest.write_json(f)
+                _link_output(args, f.name, f"{args.output}.manifest")
+
+
 def print_output_size(args: CommandLineArguments) -> None:
     if args.output_format in (OutputFormat.directory, OutputFormat.subvolume):
         MkosiPrinter.print_step("Resulting image size is " + format_bytes(dir_size(args.output)) + ".")
@@ -5286,6 +5304,7 @@ def unlink_output(args: CommandLineArguments) -> None:
     if not args.skip_final_phase:
         with complete_step("Removing output files…"):
             unlink_try_hard(args.output)
+            unlink_try_hard(f"{args.output}.manifest")
 
             if args.checksum:
                 unlink_try_hard(args.output_checksum)
@@ -6260,7 +6279,13 @@ class BuildOutput:
 
 
 def build_image(
-    args: CommandLineArguments, root: str, *, do_run_build_script: bool, for_cache: bool = False, cleanup: bool = False
+    args: CommandLineArguments,
+    root: str,
+    *,
+    manifest: Optional[Manifest] = None,
+    do_run_build_script: bool,
+    for_cache: bool = False,
+    cleanup: bool = False,
 ) -> BuildOutput:
     # If there's no build script set, there's no point in executing
     # the build script iteration. Let's quit early.
@@ -6334,6 +6359,10 @@ def build_image(
                 sshkey = setup_ssh(args, root, do_run_build_script, for_cache, cached_tree)
                 setup_network_veth(args, root, do_run_build_script, cached_tree)
                 run_postinst_script(args, root, loopdev, do_run_build_script, for_cache)
+
+                if manifest:
+                    with complete_step("Recording packages in manifest…"):
+                        manifest.record_packages(root)
 
                 if cleanup:
                     clean_package_manager_metadata(root)
@@ -6509,12 +6538,13 @@ def remove_artifacts(
             unlink_try_hard(root_home(args, root))
 
 
-def build_stuff(args: CommandLineArguments) -> None:
+def build_stuff(args: CommandLineArguments) -> Manifest:
     make_output_dir(args)
     setup_package_cache(args)
     workspace = setup_workspace(args)
 
     image = BuildOutput.empty()
+    manifest = Manifest(args)
 
     # Make sure tmpfiles' aging doesn't interfere with our workspace
     # while we are working on it.
@@ -6553,7 +6583,7 @@ def build_stuff(args: CommandLineArguments) -> None:
         # Run the image builder for the second (final) stage
         if not args.skip_final_phase:
             with complete_step("Running second (final) stage…"):
-                image = build_image(args, root, do_run_build_script=False, cleanup=True)
+                image = build_image(args, root, manifest=manifest, do_run_build_script=False, cleanup=True)
         else:
             MkosiPrinter.print_step("Skipping (second) final image build phase.")
 
@@ -6584,6 +6614,8 @@ def build_stuff(args: CommandLineArguments) -> None:
 
         if image.root_hash is not None:
             MkosiPrinter.print_step(f"Root hash is {image.root_hash}.")
+
+        return manifest
 
 
 def check_root() -> None:
@@ -7120,11 +7152,14 @@ def run_verb(raw: argparse.Namespace) -> None:
         check_root()
         check_native(args)
         init_namespace(args)
-        build_stuff(args)
-        print_output_size(args)
+        manifest = build_stuff(args)
 
         if args.auto_bump:
             bump_image_version(args)
+
+        save_manifest(args, manifest)
+
+        print_output_size(args)
 
     if args.verb in ("shell", "boot"):
         run_shell(args)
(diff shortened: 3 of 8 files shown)


bluca:
any chance we can use the `buildinfo` format? https://reproducible-builds.org/docs/recording/

keszybz:
> any chance we can use the `buildinfo` format? https://reproducible-builds.org/docs/recording/

Which one? ;)

> This product is called the ‘buildinfo’, but its exact format and the way it is distributed differs across ecosystems.

I think we can have a multi-value `--manifest=` option: `yes` or `json` would produce a full-featured json file, `debian-buildinfo` could produce the debian flavour of control format, `kiwi-packages` could produce a kiwi-ng style .packages file.

poettering:
Hmm, the manifest should record somewhere that we talk about rpm here. And I'd really try to make this generic enough so that one day .deb or any other package format can be reasonably covered by this too, even if this is not done initially.

Why not use the RPM filenames as they are? i.e. why split up the json? the filenames already contain everything we need in compressed form, no? and ultimately it's kinda what we want to write down here: an image is made up from packages, hence the package names are what matters.

What's the point of the changelog? If there's not strong usecase I'd suppress that, it's auxiliary info you can get with the rest of the info.

poettering:
hmm, regarding workflow, I think we should read "mkosi.manifest" if it exists, and pass the package versions recorded in it to dnf/yum for install. 

And then we should have "mkosi genmanifest" or so, which will unlink the current manifest, and then initiate a build, but only goes as far as determining the package versions to install, and then places them in "mkosi.manifest".

(I think implicitly generating the manifest in the output dir is also good, i think doing that is a good idea anyway, independently if people use mkosi.manifest or not).

With this, people could then do "mkosi genmanifest", and that will "seal" the package versions for them, and all rebuilds will use the same set, until they hit "mkosi genmanifest" again at which point the set will be updated to the current distro versions. And "mkosi.manifest" they'd commit to their git, so that they can do "git diff" and such easily.

keszybz:
> Hmm, the manifest should record somewhere that we talk about rpm here. 

You're right. And we already did the work to figure this out before: https://systemd.io/COREDUMP_PACKAGE_METADATA/#well-known-keys. I'll just reuse the subset that makes sense here. This way we should support deb and others without any additional effort.

> Why not use the RPM filenames as they are? i.e. why split up the json?

Actually this is not generated from package names: it's generated from the rpm db, which has package name and version as separate fields. We could *recreate* the package name, but I don't think we should. The problem is that package managers have different rules how to split the package file name into the name and version. For example, with rpm, evra separator is something like `-(?=[^a-zA-Z_])`, which is not trivial to implement e.g. in shell. Deb will have some different set of rules again. So this is much easier to process if the name alone is a separate field. That's what we do in the ELF metadata format too.

> What's the point of the changelog?

As described in the original issue, it allows users to do a diff between two manifests and see what changed in the packages. But yeah, this shouldn't be part of the changelog. I'll move it to a separate (optional) file.

> I think implicitly generating the manifest in the output dir is also good

Yeah, it should be useful for a few different things, even just after-the-fact checking of correctness.

> hmm, regarding workflow, I think we should read "mkosi.manifest" if it exists, and pass the package versions recorded in it to dnf/yum for install.

As mentioned in #700, this is rather hard to implement in a generic fashion, because packages disappear from mirrors fairly quickly and at least for Fedora, an alternate download mechanism would need to be implemented. But if the manifest is there, it certainly should be possible to implement if somebody wants in the future.

poettering:
what does dnf/yum actually do if user requests installation of a specific RPM in a specific version and the upstream repos don#t have it anymore, but the local download cache has? does it allow it because it already downloaded it? or does it refuse it because repo listings claim it doesn't exist anymore?

As long as people carefully keep their download caches around packages vanishing from download mirrors probably shouldn't matter

bluca:
That would push responsibility for handling mirrors down to the users, which doesn't sound great. For debian we got snapshot.debian.org which archives every version of everything - does fedora have something similar?

keszybz:
> what does dnf/yum actually do if user requests installation of a specific RPM in a specific version and the upstream repos don#t have it anymore, but the local download cache has? does it allow it because it already downloaded it? or does it refuse it because repo listings claim it doesn't exist anymore?

It allows it.

> As long as people carefully keep their download caches around packages vanishing from download mirrors probably shouldn't matter

Yeah, but this means that the manifest is only good to rebuild stuff that was built locally, and if no cache cleanup has been done. I actually do something like this by calling `dnf -C`. It will not download new package lists, so if you just redo an installation, it'll install the same packages as before from the cache. It's very useful as a hack to speed things up when working on mkosi, but for a user it's of limited usefulness… Why should the user repeatedly rebuild the image with the same package list?

> For debian we got snapshot.debian.org which archives every version of everything - does fedora have something similar?

Yes, it's the same in Fedora. You can download every build artifact that was ever distributed. This is also motivated by license compliance: the easiest way to satisfy GPL* is to always have the source package available somewhere. The difference is that the "current" packages are distributed through a mirror network, and automatically handled through `dnf`. "Old" packages are downloaded from a single server that can be quite slow. (It varies by location… For a few years routing to Poland was abysmal, and I'd get download speeds in the 64-128 kb/s range, so I could wait a few hours for packages when building stuff. And people wouldn't be happy if too much load was created. As long as it's just a few mkosi developers, this wouldn't matter, but as a general solution it's problematic.)  I don't want to make this sound too complicated — it's certainly doable, but not as obvious as the installations we are doing now.

poettering:
> > As long as people carefully keep their download caches around packages vanishing from download mirrors probably shouldn't matter
> 
> Yeah, but this means that the manifest is only good to rebuild stuff that was built locally, and if no cache cleanup has been done. 

So if people commit "mkosi.cache/" into their git repo (or that large file git thing) then they get reasonable reproducibility.

> I actually do something like this by calling `dnf -C`. It will not download new package lists, so if you just redo an installation, it'll install the same packages as before from the cache. It's very useful as a hack to speed things up when working on mkosi, but for a user it's of limited usefulness… Why should the user repeatedly rebuild the image with the same package list?

Because they need to fix a bug in their own sources, and then rebuild the image with their own sources updated but everything else remaining the same. I mean, this is what *I* do every day: I hack on systemd, then issue "mkosi boot -i -f" to test if it works now. To make things reasonably debuggable I want to make sure the set of RPMs remains the same between each of these builds. 

Or to check if things are really reproducible. I mean, the whole reproducibility project only exists to do exactly this: rebuild stuff and see if it matches the earlier results.

poettering:
> That would push responsibility for handling mirrors down to the users, which doesn't sound great. For debian we got snapshot.debian.org which archives every version of everything - does fedora have something similar?

Well, i would consider "mkosi.cache/" as something to commit to git then. So yes, this would be basically a local mirror, but one managed via git concepts. (by git i mean either git or git lfs or so)

keszybz:
I think this is ready for review now. The output formats will probably require some tweaking, but I'd prefer to get the basic code in place, and then work on those details laters.

mkosi/manifest.py
```diff
@@ -0,0 +1,130 @@
...
+                    stderr=DEVNULL,
+                    universal_newlines=True,
+                )
+                changelog = c.stdout.strip()
+                srpm = SourcePackageManifest(srpm, changelog)
+                self.source_packages[srpm.name] = srpm
+
+            package = PackageManifest("rpm", name, evra, size)
+            self.packages.append(package)
+            srpm.add(package)
```
behrmann, line 103: I don't have the numbers to back this up, but Python exceptions are built around the assumption, that triggering them is rare, because the stack unwinding is expensive, but this looks like it will be more common, so a check `if srpm in self.source_packages` might be a bit faster. Since I don't have a RPM system lying around I can't readily test the difference.

mkosi/__init__.py
```diff
@@ -40,32 +40,34 @@
 from textwrap import dedent
 from types import FrameType
 from typing import (
-    IO,
-    TYPE_CHECKING,
```
behrmann, line 44: That's what isort is complaining about.

mkosi/__init__.py
```diff
@@ -4292,6 +4301,41 @@ def dir_size(path: str) -> int:
     return dir_sum
 
 
+def save_manifest(args: CommandLineArguments, manifest: Manifest) -> None:
+    if manifest.has_data():
+        relpath = path_relative_to_cwd(args.output)
+
+        if ManifestFormat.json in args.manifest_format:
+            with complete_step(f"Saving manifest {relpath}.manifest"):
+                f: TextIO = cast(
```
behrmann, line 4310: With the cast the annotation should not be necessary (they are also unnecessary in two other places). I also just checked, that the cast itself doesn't seem to be necessary necessary here. (At least it type checks for me)

mkosi/__init__.py
```diff
@@ -4318,6 +4362,12 @@ def setup_package_cache(args: CommandLineArguments) -> Optional[TempDir]:
     return d
 
 
+def remove_duplicates(items: List[T]) -> List[T]:
+    "Return list with any repeats removed"
```
behrmann, line 4366:
```suggestion
    "Return list with any repetitions removed"
```
or 
```suggestion
    "Return list with any duplicates removed"
```

mkosi/backend.py
```diff
@@ -13,16 +14,17 @@
 import uuid
 from types import FrameType
 from typing import (
-    IO,
-    TYPE_CHECKING,
```
behrmann, line 17: isort also complains about this

mkosi/backend.py
```diff
@@ -52,18 +54,54 @@ class MkosiException(Exception):
...
+    centos_epel = PackageType.rpm
+    clear = PackageType.bundle
+    photon = PackageType.rpm
+    openmandriva = PackageType.rpm
+
+    def __new__(cls, package_type: PackageType) -> "Distribution":
+        entry = object.__new__(cls)
+        entry._value_ = len(cls.__members__)
+        entry.package_type = package_type
+        return cast("Distribution", entry)
```
behrmann, line 104: That is an interesting construction. Took me a while to wrap my head around it. Why not a property on the enum looking up the type in a dictionary? Maybe a tad slower and the list of distros is doubled, but a bit less magical. :)

behrmann (commented):
Haven't fully wrapped my head around everything, but here a first quick review.

mkosi/manifest.py
keszybz: Indeed the version with `if` is faster: the try..except block costs a bit, and dict lookup is O(1) (and I think the most recent lookup is also cached specially, so it is a very small O(1)). I'll convert that.

mkosi/backend.py
behrmann: No, that's understandable, but maybe add a link to the [specific doc section](https://docs.python.org/3.9/library/enum.html#when-to-use-new-vs-init) as a comment.

Maybe an additional idea: I think this would also be less magical, if you'd pass the index in explicitly instead of of calculating it dynamically via `len(cls.__members__)`, i.e. 
```python
class Distribution(enum.Enum):
    fedora = (0, PackageType.rpm)
    debian = (1, PackageType.deb)
```
While this is repetitive, I mostly stumbled over things getting the same value, because multiple distros do share packaging types. This would also be in line with the example from the docs.

mkosi/__init__.py
keszybz:
```

In [3]: d = {str(x):len(str(x)) for x in range(1000_000)}                                         

In [4]: %%timeit 
   ...: for i in range(0, 1000_000, 9): 
   ...:   try: 
   ...:     l = d[i] 
   ...:   except KeyError: 
   ...:     l = 0 
   ...:                                                                                           
41.2 ms ± 692 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [5]: %%timeit 
   ...: for i in range(0, 1000_000, 9): 
   ...:   if i in d: 
   ...:     l = d[i] 
   ...:   else: 
   ...:     l = 0 
   ...:                                                                                           
16.9 ms ± 172 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [6]: %%timeit 
   ...: for i in range(1000_000, 2000_000, 9): 
   ...:   if i in d: 
   ...:     l = d[i] 
   ...:   else: 
   ...:     l = 0 
   ...:                                                                                           
16.9 ms ± 196 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [7]: %%timeit 
   ...: for i in range(1000_000, 2000_000, 9): 
   ...:   try: 
   ...:     l = d[i] 
   ...:   except KeyError: 
   ...:     l = 0 
   ...:                                                                                           
40.9 ms ± 729 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

mkosi/__init__.py
keszybz: Oh, I see. It thinks that `IO` needs to go first because it's all-capitals, but `BinaryIO` can go in the usual place, because it's clearly completely different than `IO`. Oh, well. In effect, this order seems semi-random. I don't like automated tools that override logic.

Anyway, I dropped the commit and applied isort to keep the order that it likes in individual commits.

behrmann: I think the idea is: 
1. `CONSTANTS`
2. `Classes`
3. `functions`
and alphabetically within.

keszybz:
Updated with the change to `class Distribution` that was discussed above.

mkosi/manifest.py
```diff
@@ -0,0 +1,74 @@
...
+    type: str
+    name: str
+    version: str
+    size: int
+
+    def as_dict(self) -> Dict[str, str]:
+        return {
+            "type": self.type,
+            "name": self.name,
+            "version": self.version,
```
poettering, line 29: soo, in an earlier version i asked why not save this as the original package name. You said that it's hard to parse and hence useful to have split out. But my next question would be: why not save it in both the full standards RPM file name *and* the split out fields?

Having the full name around might be useful for cache management and such, as we then know what file to look for on disk. Having this in both ways looks like something incredibly useful to me?

man/mkosi.1
```diff
@@ -341,6 +341,13 @@ system), \f[C]gpt_btrfs\f[R] (similar, but with an btrfs file system),
 \f[C]plain_squashfs\f[R] (a plain squashfs file system without a
 partition table).
 .TP
+\f[B]\f[CB]--manifest-format=\f[B]\f[R]
+The manifest format type or types to generate.
+A comma-delimited list consisting of \f[C]json\f[R] (the standard json
```
poettering, line 346: the second "json" should be uppercased, no?

mkosi.md
```diff
@@ -342,6 +342,13 @@ details see the table below.
   squashfs file system), `plain_squashfs` (a plain squashfs file
   system without a partition table).
 
+`--manifest-format=`
+
+: The manifest format type or types to generate. A comma-delimited
+  list consisting of `json` (the standard json output format that
```
poettering, line 348: as above

poettering (commented):
lgtm, but eventually my limited python-fu comes to its end and I don't follow anymore with all the pythonesque magic.

PR #772, opened 2021-08-20
Don't let dracut fiddle with the kernel commandline in hostonly mode

DaanDeMeyer:
If hostonly_cmdline is enabled, details from the host system (aside from
needed kernel modules) leak into the initrd which we want to avoid. For
example, this caused the seclabel mount option to leak into the mount options
of my mkosi generated BTRFS mount options rendering the image unbootable in QEMU.


behrmann:
I am only vaguely familiar with this, I thought `hostonly_cmdline` was necessary for the unified efi binary boot magic. Granted, I do use it with an additional `install_items+=` to install a cmdline to `/etc/cmdline.d` in the initrd dracut generates. Could this be an alternative?

DaanDeMeyer:
Honestly, I'm not too familiar with this either but I'm pretty sure we don't want stuff from host disks/mounts influencing the kernel commandline for the generated image. Putting fewer drivers in there is OK for hostonly but that should be the extent of the host's effect on the generated image I think.

the unified kernel stuff already reads the necessary cmdline from /etc/kernel/cmdline, so I don't think dracut is doing anything extra there.

behrmann:
True, I just noticed that my usage of `hostonly_cmdline` is only a vestige in my personal config as well, but grepping for `hostonly_cmdline` shows this snippet to be the most pertinent:
```bash 
if [[ $uefi == yes ]]; then
    if [[ $kernel_cmdline ]]; then
        echo -n "$kernel_cmdline" > "$uefi_outdir/cmdline.txt"
    elif [[ $hostonly_cmdline == yes ]] && [ -d "$initdir/etc/cmdline.d" ]; then
        for conf in "$initdir"/etc/cmdline.d/*.conf; do
            [ -e "$conf" ] || continue
            printf "%s " "$(< "$conf")" >> "$uefi_outdir/cmdline.txt"
        done
    fi

    if [[ $kernel_cmdline ]] || [[ $hostonly_cmdline == yes && -d "$initdir/etc/cmdline.d" ]]; then
        echo -ne "\x00" >> "$uefi_outdir/cmdline.txt"
        dinfo "Using UEFI kernel cmdline:"
        dinfo "$(tr -d '\000' < "$uefi_outdir/cmdline.txt")"
        uefi_cmdline="${uefi_outdir}/cmdline.txt"
    else
        unset uefi_cmdline
    fi
# snip
fi
```

Am I misunderstanding this, if I think that it doesn't leak the host's cmdline into the initrd? It seems to me to only operate on conf files in `/etc/cmdline.d` in the initrd itself.

DaanDeMeyer:
I can verify again on monday, but a quick grep through dracut's source code shows no mention of the "seclabel" mount option. A grep through mkosi doesn't turn up any mention of it either and neither does systemd's source code. On the other hand, my fedora machine has a bunch of host mounts with the seclabel option. That makes me pretty sure it's leaking from the host into dracut.

DaanDeMeyer:
@behrmann So I looked a little further, and you're right that dracut only operates on conf files in /etc/cmdline.d in the initrd itself. However, dracut also puts config files in there. The one we're interested in is 95root-dev.conf. This file contributes the `seclabel` option (on my machine) that causes the issue. If we go digging in dracut source code, we find modules.d/95rootfs-block/module-setup.sh which is a script that, if hostonly_cmdline is set, figures out the mount options for the root device and adds them to 95root-dev.conf

I've had similar problems with grub in the past where it was also fiddling with the root device from inside the mkosi container.

Turning of hostonly_cmdline prevents dracut from writing 95root-dev.conf which fixes the issue. But in reading the source code I found out that there's also a hostonly_default_device option which if disabled stops dracut from fiddling with devices. This also fixes the issue so I opted for that option instead since it's less broadly impacting than hostonly_cmdline.

behrmann:
Looks good. If it doesn't have any other ramifications and all partitions are still found on boot, which should work thanks to the generator reading the partition UUIDs, this looks like a good solution.

PR #777, opened 2021-08-28
Replace contextlib.ExitStack by direct cm invocations

keszybz:
I think it's nicer to keep things simple. Unfortunately black
turns the first chunk into a complete mess, so it needs to be disabled.
With python 3.9+ we will be able to drop the line continuations.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1198,14 +1198,15 @@ def luks_setup_all(
     if not args.output_format.is_disk():
         yield LuksSetupOutput.empty()
         return
+
     assert loopdev is not None
 
-    with contextlib.ExitStack() as stack:
-        root = stack.enter_context(luks_setup_root(args, loopdev, do_run_build_script))
-        home = stack.enter_context(luks_setup_home(args, loopdev, do_run_build_script))
-        srv = stack.enter_context(luks_setup_srv(args, loopdev, do_run_build_script))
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -3676,15 +3678,14 @@ def insert_partition(
...
+    if args.root_partno == partno:
+        luks_format_root(args, loopdev, False, False, True)
+        cm = luks_setup_root(args, loopdev, False, True)
+    else:
+        cm = contextlib.nullcontext(partition(loopdev, partno))
 
-        path = dev if dev is not None else partition(loopdev, partno)
-        run(["dd", f"if={blob.name}", f"of={path}", "conv=nocreat,sparse"])
+    with cm as dev:
+        run(["dd", f"if={blob.name}", f"of={dev}", "conv=nocreat,sparse"])
```
DaanDeMeyer, line 3688: This doesn't handle the case where `dev` is `None`

DaanDeMeyer (changes requested):

mkosi/__init__.py
keszybz: Ooops, indeed. Should be OK now.

keszybz:
I restarted the CI here — it failed in arch/gpt_*, and I'm not sure if it was related to this PR or an unrelated transient issue.

DaanDeMeyer:
Yeah there's something that got released recently causing timeouts in the Arch boot tests, but the current failures aren't related to this PR so LGTM

PR #778, opened 2021-09-01
Update some Package Versions to latest

pheiduck:
missing dnf and rpm


DaanDeMeyer:
pacman is now packaged as an xz archive instead of gz so that should be changed in the url to avoid 404 errors when trying to build pacman

pheiduck:
> pacman is now packaged as an xz archive instead of gz so that should be changed in the url to avoid 404 errors when trying to build pacman

Yup, you right, forgot to change it. Will fix it

pheiduck:
@DaanDeMeyer rework is done on my side.

pheiduck:
`meson.build:111:8: ERROR: Problem encountered: A Python3 binding requires a GObject introspection. `

Do not have a solution for this

DaanDeMeyer:
Try installing python3-gi

pheiduck:
Close this pr because I can not fix the problem

Issue #769, opened 2021-08-12
Refactor everything to make file-based configuration primary, cmdline args secondary


keszybz:
In the man page, and in the code, the cmdline options are the most exposed. But that seems backwards: the user can hardly put 50 options on the commandline. I think we should invert the code and the docs to make the config file options the "primary key".

In the man page:
```
Command Line Parameters
--foo
... 
Command Line Parameters and their Settings File Counterparts
--foo Output Foo=
...
```
becomes
```
Settings
[Output] section
Foo=
--foo
...
```

Also, I would like to write the whole configuration, i.e. the `args` variable, to the manifest as json (for #768). But I think it'd be better to rearrange the code first. Not everything should be saved to the manifest, and I think the data structure that we have should also provide a way to annotate what should be stored there. And then there should be an automatic way to populate `args` from a saved manifest. In the proposed man page structure above we'd add another line that `Output.Foo` becomes `["output"]["foo"]` in the manifest.

This will also make the addition of new things easier: there will be one place to modify in the docs, and one place to modify in the code.

I'll work on this, but I want to gather some feedback before I start.

behrmann:
I like the idea, but I'm not sure that files should trump the CLI. In my view files should provide the the baseline and the CLI should be the escape hatch to quickly explore "Is this the option I need?" without having to touch a file, but maybe I'm not fully understanding your reasoning. From my point of view, changing the man page to focus on the file based config would be the central thing. 

Refactoring `args` is something I've already noodled over quite a bit with the splitting out the distributions project, since it's weaved through everything. It would be great if it could be split up a bit into logical chunks, e.g. maybe everything concerned with the (partition) layout of the image going in one bucket and so on.

behrmann:
Ah, then I misunderstood.

DaanDeMeyer:
> Not everything should be saved to the manifest, and I think the data structure that we have should also provide a way to annotate what should be stored there.

Might be worth looking into https://docs.python.org/3/library/dataclasses.html#dataclasses.field to annotate which fields should be stored in JSON and which shouldn't.

But yeah, making the config the source of truth sounds like a good way to go. Like @behrmann said, I use CLI options more when trying stuff out rather than as a primary configuration mechanism.

DaanDeMeyer:
@keszybz With your documentation rework and the recently merged configuration rework, I think this is more or less fixed now? We don't yet support writing the config to the manifest but I think that should be a different issue.

keszybz:
[...]

PR #780, opened 2021-09-02
Use config keys as primary in documentation, rename [Packages] to [Content]

keszybz:
This implements the first part of #769. While working on this, I found a bunch of small discrepancies, and subsequent commits fix a bunch. See individual commit messages for details.

(I had to merge #768, since this PR builds on it.)


mkosi.md
```diff
@@ -380,61 +392,66 @@ details see the table below.
...
+  cache files are removed, too, before the re-build is initiated. If a
+  package cache is used (also see the "Files" section below),
+  specifying this option thrice will ensure the package cache is
+  removed too, before the re-build is initiated. For the `clean`
+  operation this option has a slightly different effect: by default
+  the verb will only remove build artifacts from a previous run, when
+  specified once the incremental cache files are deleted too, and when
+  specified twice the package cache is also removed.
+
+  <!--  FIXME: allow `Force=<n>` -->
```
DaanDeMeyer, line 429: I'm wondering whether we should do Force in config files at all. This seems like one of those options that's better off as a CLI only option. Same with --incremental.

General rant: We should get rid of the argparse hack and generate the config parser and argparse object from the `CommandLineArguments` dataclass we have now (which we should rename to just `Args` or something similar.

mkosi.md
```diff
@@ -797,183 +804,258 @@ details see the table below.
...
+`Autologin=`, `--autologin`
 
-: Enable autologin for the `root` user on pts/0 (nspawn), tty1 (QEMU) and
-  ttyS0 (QEMU with `--qemu-headless`) by patching `/etc/pam.d/login`.
+: Enable autologin for the `root` user on `/dev/pts/0` (nspawn),
+  `/dev/tty1` (QEMU) and `/dev/ttyS0` (QEMU with `QemuHeadless=yes`)
+  by patching `/etc/pam.d/login`.
 
-`--extra-search-paths=`
+<!-- FIXME: why are all those settings in [Validation] ?! -->
```
DaanDeMeyer: The groupings don't really make a lot of sense for a lot of settings. AFAIK we don't even enforce them. I wonder if we should drop them alltogether.

DaanDeMeyer:
DIdn't do a super detailed look through every change but LGTM.

mkosi.md
keszybz: I think they are useful to make it easier for humans to read the config files… I'd keep them, at least for now.

keszybz:
The error from CI is pyright complaining about str↔Path assignments. I have a patch ready to rework most of that code, so I don't think it makes sense to fix the issue here. (It's not related to this PR, but preexisting.)

OK, since this is mostly moving documentation text around and @DaanDeMeyer gave it a thumbs up, let's merge.

mkosi/__init__.py
```diff
@@ -6142,6 +6144,9 @@ def print_summary(args: CommandLineArguments) -> None:
     if args.build_script:
         MkosiPrinter.info("                 Run tests: " + yes_no(args.with_tests))
 
+    MkosiPrinter.info("                  Password: " + ("default" if args.password is None else "set"))
```
poettering, line 6147: maybe actually change left hand side to "Root Password"?

mkosi.md
poettering: > I'm wondering whether we should do Force in config files at all. This seems like one of those options that's better off as a CLI only option. Same with --incremental.

I agree, Force in the config file is weird. Maybe at least drop from the docs, and keep it in the code for compat?

PR #781, opened 2021-09-03
Fix --debug=run

keszybz:


DaanDeMeyer:
LGTM

keszybz:
I added another type annotation to make mypy happy. For some reason it wasn't complaining locally on my machine…

keszybz:
pyright is unhappy, but it's because of the str↔Path issues, unrelated to this PR.

PR #783, opened 2021-09-03
Drop black from ci, add a check that tabs are not used, do pyflakes check, adjust indentation

keszybz:


keszybz:
fedora/plain_squashfs died with "HTTP request sent, awaiting response... 502 Bad Gateway".
arch/gpt_btrfs died with the usual timeout.
Both seem unrelated.

behrmann:
LGTM (mostly, I guess we can argue about style for a long time ;) ), but the unit test CI didn't run because of a startup failure
> The workflow is not valid. .github/workflows/ci-unit-test.yml: (Line: 26, Col: 12, Idx: 476) - (Line: 26, Col: 13, Idx: 477): While parsing a tag, did not find expected tag URI.

keszybz:
It seems to not like `! git grep`. I wrapped in in `sh -c '…'` now.

keszybz:
`sh -c '! git grep "\t" "*.py"'` does not work as expected… I'm trying now with another layer of escaping added.

keszybz:
Now I'm really confused. This is executed by some non-bash shell that gets quoting wrong. How is that even possible?

mrc0mmand:
> Now I'm really confused. This is executed by some non-bash shell that gets quoting wrong. How is that even possible?
> 
> (And it doesn't seem to be dash, because dash works correctly here…)

This is actually `(git) grep`'s fault as it can't handle `\t`, since `\t` is not defined in POSIX. You'd have to use something like:

```
$ git grep $'\t' '*.py'
mkosi/manifest.py:              self.packages.append(package)
```

but that's not portable (since `$''` is a bashism), or tell `(git) grep` to use PCRE:
```
$ git grep -P '\t' mkosi/manifest.py
mkosi/manifest.py:              self.packages.append(package)
```

So I guess using an external python script here is a sensible choice.

keszybz:
Oh, OK. I have `grep.patternType = perl` defined in `~/.gitconfig`. So it worked for me only because of that. Thank you for the explanation. Hmm, I think it's better to use PCREs. I'll revert the patch to do `sh -c '…'` again.

DaanDeMeyer:
LGTM, we should modify the unit test CI to allow individual steps to fail without blocking the rest of the steps from executing though. I'll do a PR for that

PR #784, opened 2021-09-05
fedora: add keys until F36

keszybz:
I also added older keys: we might want to build images with
older Fedora, and it's better to have the key hashes than
not to have them. The list now includes all keys listed by
fedora-gpg-keys-34-2.noarch.


bluca (approved):
lgtm

poettering:
(maybe one day we should switch to longer hashes, the abbreviated ones are problematic security wise)

keszybz:
https://pagure.io/fedora-web/websites/issue/196

PR #786, opened 2021-09-07
Override various autodetection features

keszybz:
Again, this is a grab-bag of loosely related commits. But splitting this into multiple PRs would generate conflicts, so I don't think it's worth it. Apart from the refactorings, this implements two new "features": `--finalize-script=` makes mkosi skip the finalize script (and similarly for the build script and others), and `--clean-package-metadata=yes|no` disables autodetection and does the cleanup or not as specified.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1687,7 +1687,7 @@ def clean_dnf_metadata(root: Path, always: bool) -> None:
     keeping the dnf metadata, since it's not usable from within the
     image anyway.
     """
-    dnf_metadata_paths = [
+    paths = [
         root / "var/lib/dnf",
         *root.glob("var/log/dnf.*"),
         *root.glob("var/log/hawkey.*"),
@@ -1696,16 +1696,17 @@ def clean_dnf_metadata(root: Path, always: bool) -> None:
 
     cond = always or os.access(root / "bin/dnf", os.F_OK, follow_symlinks=False)
 
-    if  not cond or not any(os.path.exists(path) for path in dnf_metadata_paths):
+    if not cond or not any(path.exists() for path in paths):
         return
 
     with complete_step("Cleaning dnf metadata…"):
-        remove_glob(*dnf_metadata_paths)
+        for path in paths:
+            unlink_try_hard(path)
 
 
 def clean_yum_metadata(root: Path, always: bool) -> None:
     """Remove yum metadata if /bin/yum is not present in the image"""
-    yum_metadata_paths = [
+    paths = [
         root / "var/lib/yum",
         *root.glob("var/log/yum.*"),
         root / "var/cache/yum",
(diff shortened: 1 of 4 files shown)


mkosi/__init__.py
```diff
@@ -5650,13 +5688,27 @@ def build_auxiliary_output_path(args: argparse.Namespace, suffix: str, can_compr
...
+def script_path(value: Optional[str]) -> Optional[Path]:
+    if value is None:
+        return None
+    if value == '':
+        return DISABLED
+    return Path(value)
+
+
+def normalize_script(path: Optional[Path]) -> Optional[Path]:
+    if not path or path is DISABLED:
```
behrmann, line 5702:
```suggestion
    if not path or path == DISABLED:
```
Even though the string might get interred and this could work in that case (don't really know), I generally get
```
>>> a = Path('DISABLED')
>>> b = Path('DISABLED')
>>> a is b
False
```

behrmann (commented):
LGTM, with on minor thing

mkosi/__init__.py
```diff
@@ -5680,13 +5685,27 @@ def build_auxiliary_output_path(args: argparse.Namespace, suffix: str, can_compr
     return output.with_name(f"{output.name}{suffix}{compression or ''}")
 
 
-def check_valid_script(path: Path) -> None:
+DISABLED = Path('DISABLED')  # a placeholder value to suppress autodetection
```
poettering, line 5691: hmmm, but this is a theoretically valid path, no? I mean, why not just use `/dev/null` here? I think this makes conceptionally a lot more sense, and is also a special, valid path here.

poettering:
lgtm otherwise, but I am not the best to comment on the pythonicness on the code

mkosi/__init__.py
keszybz: Internment of strings is not relevant here: we are comparing Path objects. And I used `is` on purpose: `DISABLED` is created once in the module, and we always use this particular object in the code. If we were to recreate the object in a new place, `is` would not work, and `==` would be required. But I don't think we want to allow this: the original object should always be used. In fact, if I specify `--finalize-script=DISABLED`, this will be handled like a normal path, which I think is good.

(Let's say that I want to reuse this logic in `backend.py`:
```
from mkosi import DISABLED
x: Path = DISABLED
```
but not
```
x: Path = Path('DISABLED')
```)

(Overall, I'd prefer to create a real singleton object for this. But Python doesn't really make this easy. A different option would be to create a non-Path object, e.g. `DISABLED = object()` and use that as the placeholder. But this would be a typing violation, and I don't think it's worth it to adjust the code to allow a different type. In practice, the thing with Path('DISABLED') should just work without any issues except theoretical concerns.)

File: mkosi/__init__.py (hunk 2 above)
Comment (keszybz) [reply mkosi/__init__.py:5691]: See the discussion above. `--script=DISABLED` would actually work (and run `./DISABLED`).

File: mkosi/__init__.py (hunk 1 above)
Comment (behrmann) [reply mkosi/__init__.py:5702]: Ah, I see, you're right. No problem then.

File: mkosi/__init__.py (hunk 2 above)
Comment (DaanDeMeyer) [reply mkosi/__init__.py:5691]: Let's add this as a comment with the constant maybe?

Comment (keszybz) [reply mkosi/__init__.py]: >  I mean, why not just use `/dev/null`

We *could*, but it doesn't seem better. It might trick the reader into thinking that we actually read `/dev/null` and then do nothing because it's empty.

> Let's add this as a comment with the constant maybe?

Added now.

**Comment by DaanDeMeyer** [comment]
LGTM

File: mkosi/__init__.py [hunk 3]
```diff
@@ -1693,71 +1693,79 @@ def clean_dnf_metadata(root: Path) -> None:
...
+    """Remove yum metadata if /bin/yum is not present in the image"""
     yum_metadata_paths = [
         root / "var/lib/yum",
         *root.glob("var/log/yum.*"),
         root / "var/cache/yum",
     ]
-    keep_yum_data = os.access(root / "bin/yum", os.F_OK, follow_symlinks=False)
 
-    if keep_yum_data or not any(os.path.exists(path) for path in yum_metadata_paths):
+    cond = always or os.access(root / "bin/yum", os.F_OK, follow_symlinks=False)
```
bluca, line 1715: @keszybz this changed the default to always remove metadata if the binary is present - was that intentional?

PR #787, opened 2021-09-07
fedora: store full key identifiers internally

keszybz:
Fedora website currently uses short identifiers, and we can't do anything
about this until they fix it. But let's store the full hash internally.

Inpired by https://github.com/systemd/mkosi/pull/784#issuecomment-913819662.


poettering:
excellent! thanks!

lgtm

behrmann:
LGTM.

PR #788, opened 2021-09-07
Bump advertised minimum Python version and grab bag of typing

behrmann:
So far the documentation and mypy config target Python 3.6, but our setup.py already specified 3.7.

My mypy just complained about `contextlib.nullcontext` not existing and lo and behold, it was only added in 3.7. Since even Debian oldstable has Python 3.7 and we've already set it in `setup.py` I think bumping the minimum we advertise makes the most sense.


setup.cfg
```diff
@@ -4,7 +4,7 @@ max-line-length = 119
 multi_line_output = 3
 include_trailing_comma = True
 [mypy]
-python_version = 3.6
+python_version = 3.7
```
DaanDeMeyer, line 7: Ahhhhhh, this is why mypy in CI was being so annoying about all the 3.7 stuff.

Can you remove the --python-version cli argument from mypy in CI? I'm pretty sure it won't be necessary anymore with this change.

behrmann: And I just wondered why it was working in CI and not on my machine :D

behrmann: I think we could move almost everything to `pyproject.toml`, now that we already have it, but flake8 doesn't seem to support it, so we won't be able to move to a single config file just yet.

behrmann:
Just added a couple of commits, since I found the failing pyright test annoying and at least on my machine these are gone now, although only by disabling the check for general type issues, but I think those are spurious.

mkosi/__init__.py
```diff
@@ -3458,8 +3458,8 @@ def make_tar(args: CommandLineArguments, root: Path, do_run_build_script: bool,
 
 def find_files(root: Path) -> Generator[str, None, None]:
     """Generate a list of all filepaths relative to @root"""
-    root = str(root).rstrip("/") + "/"  # make sure the path ends in exactly one '/'
-    prefix_len = len(root)
+    root: str = str(root).rstrip("/") + "/"  # make sure the path ends in exactly one '/'
```
DaanDeMeyer, line 3461: We should just use a different variable name here

mkosi/__init__.py
```diff
@@ -5527,7 +5527,7 @@ def find_skeleton(args: argparse.Namespace) -> None:
 def args_find_path(args: argparse.Namespace, name: str, path: str, *, as_list: bool = False) -> None:
     if getattr(args, name) is not None:
         return
-    path = Path(path).absolute()
+    path: Path = Path(path).absolute()
```
DaanDeMeyer, line 5530: Ditto

pyproject.toml
```diff
@@ -6,3 +6,7 @@ target-version = ['py37']
 profile = "black"
 multi_line_output = 3
 py_version = "37"
+
+[tool.pyright]
+pythonVersion = "3.7"
+reportGeneralTypeIssues = false
```
DaanDeMeyer, line 12: I'm pretty sure we should be able to fix these. I'd be very surprised if these were spurious.

DaanDeMeyer (changes requested):

pyproject.toml
behrmann: While the ones about shadowing variables are easy to fix (although I don't really see the necessity), the last one about 
```
  /home/mkosi/mkosi/backend.py:107:15 - error: Cannot assign member "package_type" for type "Distribution"
  Member "package_type" is unknown (reportGeneralTypeIssues)
 ```
really stumps me. I think that's a point where Python is just too dynamic for static analysis.

Comment (DaanDeMeyer) [reply pyproject.toml:12]: OK, I actually missed that, it's kinda understandable that typing doesn't work there. Can we add `package_type: PackageType` to the enum declaration to fix this?

Comment (behrmann) [reply pyproject.toml:12]: I don't really know how. I mean, we're doing that right now in the canonical way. I'm not sure I'd call this a bug in pyright, but it's a missing feature and something that they should specialcase. For now I'd say there's two ways to get rid of the error: Switch of `reportGeneralTypeIssues` for pyright or adding a `# type: ignore` which blinds both pyright and mypy to this line. I'd lean to the former.

File: mkosi/__init__.py (hunk 3 above)
Comment (behrmann) [reply mkosi/__init__.py]: Renamed it `abspath`, for now. I think the proper solution would be to do what you started in #790 and add constants like `MKOSI_EXTRA = Path('mkosi.extra')` and so on for all special files, then this function could easily just take a `Path`, too.

File: pyproject.toml (hunk 4 above)
Comment (DaanDeMeyer) [reply pyproject.toml:12]: I mean something like this:

```
class Distribution(enum.Enum):
    package_type: PackageType
    fedora = 0, PackageType.rpm
    ....
```

mkosi/__init__.py
behrmann: Rewrote the function to return a `Path`, which then implies getting calling `os.fspath` on the results. In my tests the resulting lists are the same (except for the type), so this shouldn't break @keszybz's cpio archives.

pyproject.toml
behrmann: Ah, that does work!

behrmann:
Only fixed the commit message on the commit adding the pyright config, since it was now outdated.

keszybz:
LGTM.

behrmann:
Ok, somehow the rewrite of `parse_args_file_group` breaks the tests even with #792 applied. Locally I've already dropped the commit and exchanged it for something smaller. Once #792 is merged, I'll rebase against that and push that.

mkosi/__init__.py
DaanDeMeyer: Still using the same variable name here though, can we fix that as well? Changing an existing variable's type just feels messy to me

behrmann: Are we looking at the same version? In the version here `root` isn't touched anymore, no?

behrmann:
Changed the title so Github hopefully stops running the tests needlessly. Will change it back after rebasing against main once #792 is merged.

behrmann:
Since the pertinent tests don't run in CI in #792, since they are after pyright, I've merged #792 locally and added them here. If CI runs through without problems here, we'll see that #792 is good as well.

keszybz:
arch/gpt_btrfs timed out, but that seems unrelated. So #792 should be good to merge.

DaanDeMeyer:
This one can be rebased now

keszybz:
All 41 checks successful, this must be some black magic!

PR #790, opened 2021-09-07
Make sure we append Paths to the lists of extra/skeleton trees

DaanDeMeyer:


keszybz:
LGTM.

behrmann:
LGTM. As mentioned in #788 I think we could take this further and define something like `MKOSI_EXTRA = Path('mkosi.extra')` for all special files. I think this would help in switching over the rest of the paths to `Path` objects.

PR #791, opened 2021-09-07
Smaller gpt header and footer

keszybz:


keszybz:
Note that the commits should be viewed with `-b`, because the function is indented, which makes for a scary-looking diff, but the actual change is much smaller.

keszybz:
Fixed one missing type annotation.
In the meantime, the suspected bug in util-linux turned out to be already solved. I split out that part to a separate commit and adjusted comments.

DaanDeMeyer:
This one can be rebased after @behrmann's PR is merged and unit test CI should be green again then

mkosi/__init__.py
```diff
@@ -3726,7 +3726,7 @@ def insert_partition(
 
     run(["sfdisk", "--color=never", loopdev], input='\n'.join(table).encode("utf-8"))
     run(["sync"])
-    run(["partx", "--update", loopdev])
+    run(["partprobe", loopdev])
```
poettering, line 3729: hmm, this adds a dep onto "parted" which we so far didn't have.

Maybe use "blockdev --rereadpt"?

keszybz: Oh, indeed. For some reason I thought it's from util-linux too. I'll change this.

mkosi/__init__.py
```diff
@@ -384,8 +380,8 @@ def format_bytes(num_bytes: int) -> str:
     return f"{num_bytes}B"
 
 
-def roundup512(x: int) -> int:
-    return (x + 511) & ~511
+def roundup(x: int, step: int) -> int:
+    return (x + step-1) & ~(step-1)
```
poettering, line 391: hmm, `& ~(step-1)` only works if `step` is an exponent of 2, but the very generic name doesn't make this clear. maybe write as:

```py
return ((x + step - 1) // step) * step 
```

which has the big benefit that it works for any values of `step`

mkosi/__init__.py
```diff
@@ -3675,11 +3699,12 @@ def insert_partition(
         old_table = PartitionTable.read(loopdev)
     else:
         # No partition table yet? Then let's fake one...
-        old_table = PartitionTable.empty()
+        old_table = PartitionTable.empty(args.gpt_first_lba)
 
-    blob_size = roundup512(os.stat(blob.name).st_size)
+    blob_size = roundup(os.stat(blob.name).st_size, 512)
```
poettering, line 3714: we should probably start rounding up partition sizes, and start offsets to multiples of 4K too. it's where things are really going with today's HDDs and it makes a ton of sense to align disk sector size and memory page sizes. and the 3.5K this might waste here and there are probably nothing to be concerned about, regardless if we are talking about tiny initrd extensions or anything else.

mkosi/__init__.py
```diff
@@ -197,6 +197,13 @@
 V = TypeVar("V")
 
 
+def print_between_underlines(s: str) -> None:
```
poettering, line 200: and "underline" for me would be `_`, but you use `-` here...

mkosi.md
```diff
@@ -1126,6 +1126,12 @@ Those settings cannot be configured in the configuration files.
   rebuilding of these cached images, combine `-i` with `-ff` to ensure
   cached images are first removed and then re-created.
 
+`--debug=`
+
+: Enable additional debugging output. Takes a comma-separated list of
+  arguments specifying the area of interest. Pass any invalid value
+  (e.g. empty) to list currently accepted values.
```
poettering, line 1133: enclose `empty` in backticks?

keszybz: It was intended to mean really empty, i.e. `--debug=`. I want to rework it to allow `--debug=help` later on, so the exact phrasing here will change here anyway.

mkosi/__init__.py
keszybz: Yep. I also changed partition alignment to 4k, so rounding to 512 will not make any difference.

keszybz:
I changed `roundup()` as suggested, and renamed `print_between_underlines()` to `print_between_lines()`.
The alignment changes I completely agree with, but I think it's better to do it in a separate patch later, this pull request is already changing too many things.

keszybz:
```
--2021-09-09 13:56:01--  https://github.com/rpm-software-management/libcomps/archive/libcomps-0.1.17.tar.gz
Resolving github.com (github.com)... 192.30.255.112
Connecting to github.com (github.com)|192.30.255.112|:443... connected.
HTTP request sent, awaiting response... 302 Found
Location: https://codeload.github.com/rpm-software-management/libcomps/tar.gz/libcomps-0.1.17 [following]
--2021-09-09 13:56:01--  https://codeload.github.com/rpm-software-management/libcomps/tar.gz/libcomps-0.1.17
Resolving codeload.github.com (codeload.github.com)... 192.30.255.120
Connecting to codeload.github.com (codeload.github.com)|192.30.255.120|:443... connected.
HTTP request sent, awaiting response... 404 Not Found
2021-09-09 13:56:01 ERROR 404: Not Found.
```
Hmm.

DaanDeMeyer:
This is #793 breaking CI. libcomps archive name changes when moving to 0.1.17. #803 contains the correct fix along with the version upgrade.

keszybz:
But #793 passed CI… Anyway, I'll just wait for #803.

keszybz:
CI passes with one (hopefully unrelated) timeout.

mkosi/__init__.py
```diff
@@ -3645,46 +3645,56 @@ def make_generated_root(args: CommandLineArguments, root: Path, for_cache: bool)
...
+@dataclasses.dataclass
+class PartitionTable:
+    partitions: List[str]
+    last_partition_sector: int
 
-    c = run(["sfdisk", "--dump", loopdev],
-            stdout=PIPE,
-            universal_newlines=True)
+    @classmethod
+    def read(cls, loopdev: Path) -> 'PartitionTable':
```
DaanDeMeyer, line 3663: Add `from __future__ import annotations` at the very top of the file and you can remove the string quotation marks from `PartitionTable` here

mkosi/__init__.py
```diff
@@ -3645,46 +3645,56 @@ def make_generated_root(args: CommandLineArguments, root: Path, for_cache: bool)
...
+            if start is not None and size is not None:
+                end = start + size
+                if end > last_sector:
+                    last_sector = end
 
-    return table, last_sector * 512
+        return cls(table, last_sector * 512)
+
+    @classmethod
+    def empty(cls) -> 'PartitionTable':
```
DaanDeMeyer: Ditto

mkosi/backend.py
```diff
@@ -519,6 +520,29 @@ def run(
...
+    for attempt in range(attempts):
+        try:
+            return run(cmdline, check, delay_interrupt, stdout, stderr, **kwargs)
+        except subprocess.CalledProcessError:
+            if attempt == attempts - 1:
+                raise
+            time.sleep(delay)
+            delay = min(delay * 2 + 0.01, 1)
+
+    assert False  # make mypy happy
```
DaanDeMeyer, line 543: What about the following instead:

```
for attempt in range(attempts):
    try:
        return run(cmdline, check, delay_interrupt, stdout, stderr, **kwargs)
    except subprocess.CalledProcessError as e:
        time.sleep(delay)
        delay = min(delay * 2 + 0.01, 1)

raise e
```

behrmann: `e` doesn't leak out of the except scope:
```
In [1]: try:
   ...:     1/0
   ...: except ZeroDivisionError as e:
   ...:     print("oops")
   ...: 
oops

In [2]: e
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-3-094e3afb2fe8> in <module>
----> 1 e

NameError: name 'e' is not defined
```

DaanDeMeyer: Damn I was hoping it would, nvm then

keszybz: It could be saved, e.g. with `ee = e`, but the bigger problem is that this sleep one extra time. And since the backoff is exponential, we essentially sleep the longest at the end.

mkosi/__init__.py
keszybz: Thanks, I didn't know this is possible. I did that, and also pushed one more commit to use the same trick in other places.

keszybz:
arch/gpt_ext4 timed out, looks unrelated.

mkosi/__init__.py
poettering: you did? don't see it?

keszybz: I changed the partition alignment, not the size of the blob…

keszybz:
PTAL, pretty please ;)

DaanDeMeyer:
LGTM

PR #792, opened 2021-09-07
Fix tests

keszybz:


keszybz:
Mea culpa. @DaanDeMeyer you probably want to look at this one before the others.

behrmann:
What a wonderful coincidence. :) I was just typing those very same fixes because I wondered how my changes in #788 broke the tests until I noticed they are broken on `main`.

behrmann (approved):

keszybz:
One more obvious fixlet.

DaanDeMeyer:
CI fails but that's because of pyright, let's merge

PR #793, opened 2021-09-07
Update Dependencies skipping Update to Pacman 6.x because we had erro…

pheiduck:
…rs (Build failes)


keszybz:
My biased opinion is that this approach isn't useful. I can understand specifying a version constrains, occasionally and temporarily, as a one-sided range, to work around some specific bug. But pinning package versions and installing and compiling base system software just seems wrong. `action.yaml` currently redoes packaging for four or five different system packages. Can't we find a reasonable distro that provides the required packages and just use that? (Also why do we install by hand a file that is provided by systemd?)

Rant aside, @Werkov, @DaanDeMeyer you probably want to review this.

behrmann:
I'm not sure I know a distro that provides everything. Debian provides apt (obviously), dnf (since bullseye), yum (until buster) and zypper, but not Arch's pacman and for the ones it provides I have no idea how current they are. Arch might have everything on the AUR, but that also means building everything by again. I don't really see an out to building at least *some* of the stuff ourselves. What would be great if the build artifacts could be just reused.

keszybz:
Fedora stable has `rpm-4.16.1.3-1.fc34.x86_64`, `dnf-4.8.0-1.fc34.noarch`, `pacman-6.0.0-1.fc34.x86_64`, `apt-0:2.3.8-1.fc34.x86_64`, and of course the latest systemd. `mypy` might be bit behind in the stable versions, but that's easy to install with pip if necessary. Do we need something else?

Werkov:
I'm not also super-excited about the repeated builds for every CI action run. I thought the limitation was the GH actions environment that's based on Ubuntu (20.04?) and therefore rebuilt newer versions are unfortunately needed. (Or can we get out of this by a GH action deployed inside a CI container image?(?)). 

I'm only familiar with opensuse case and there was the requirement rpm >= 4.16.1 which is alas not fulfilled in Ubuntu 20.04 (but I'm [testing](https://github.com/systemd/mkosi/pull/800) whether it matters).

DaanDeMeyer:
To give some context, I went with this approach as using containers didn't work because they don't have access to loop devices and VMs weren't really an option as they weren't KVM accelerated last time I checked so they'd be too slow. Unfortunately we can only use Ubuntu with Github Actions so I had to find a way to make it work which ended up with me building everything by hand.

The current approach where we put everything in the yaml  is a massive PITA to debug though so we should probably extract all this logic into a shell script that we can run in a Ubuntu container which would make this whole setup a lot easier to debug. In fact, we can just call that shell script from a mkosi.build file and that should give us what we need. I'll try and set that up tomorrow.

Not sure if we could get some self hosted Fedora machines to use with Github Actions, that would make everything a lot easier as well as Fedora has everything we need packaged.

behrmann:
If were stuck with Ubuntu as a base, maybe a silly idea: Couldn't we build the the dependencies in a separate repository and publish the results as debs as releases? If one doesn't do proper lintian-clean Debian packages, they're actually quite easy to generate. This way they'd be built once and the CI could just download the artifacts and installs them.

DaanDeMeyer:
I think that adds complexity we don't really want to manage. Luckily, Github released support for `uses` in composite actions (which is what we use) last month. This should allow us to use the caching action to cache the build results directory. This saves us from having to rebuild all packages every time.

keszybz:
+1 for making this easier to run locally.

I don't think we can get Fedora machines in any reasonable manner. I don't think there's even an official Fedora image in Azure. We have been trying to fix this from the Fedora side for years, but it seems impossible.

keszybz:
```
Resolving github.com (github.com)... 140.82.113.3
Connecting to github.com (github.com)|140.82.113.3|:443... connected.
HTTP request sent, awaiting response... 302 Found
Location: https://codeload.github.com/rpm-software-management/rpm/tar.gz/rpm-4.16.1.3-release [following]
--2021-09-08 13:26:50--  https://codeload.github.com/rpm-software-management/rpm/tar.gz/rpm-4.16.1.3-release
Resolving codeload.github.com (codeload.github.com)... 140.82.112.10
Connecting to codeload.github.com (codeload.github.com)|140.82.112.10|:443... connected.
HTTP request sent, awaiting response... 404 Not Found
```

Werkov:
Hm, so the rpm bump is wrong, `rpm-4.16.1.3` != `rpm-4.16.1.3-release` in https://github.com/systemd/mkosi/blob/main/action.yaml#L70 and https://github.com/rpm-software-management/rpm/releases.

DaanDeMeyer:
You can checkout https://github.com/systemd/mkosi/pull/803 to iterate locally on the dependency version changes instead of having to wait for GA all the time.

keszybz:
The tests all pass, let's merge this.

PR #796, opened 2021-09-08
mkosi: dont encrypt genkey private keys

poettering:
Let's turn off "DES" encryption of X509 private keys we generate. It's
not really a useful feature, we aren't really an interactive tool (and
thus asking for a password is weird). And it's not secure at all.


DaanDeMeyer:
Lgtm

behrmann:
Ah, missed that. LGTM.

PR #798, opened 2021-09-08
mkosi: add support for PKCS7 signed images (WIP)

poettering:
This adds support for PKCS7 signed Verity disk images, compatible with the kernel's logic for that.


mkosi/__init__.py
```diff
@@ -32,6 +33,7 @@
 import urllib.parse
 import urllib.request
 import uuid
+from OpenSSL import crypto
```
behrmann, line 36: This means that we add the python openssl bindings as a first non-dev dependency. We can either do that, or move the import into the verity using functions and bail out if it's not there. 

The latter would preserve being (mostly) stdlib only, whereas the former could open the option of using other external libraries. In the former case some version constraints should be added to `setup.py`.

I'm not opining either way for or against the usage of external libraries, but I think this might warrant wider discussion.

mkosi/__init__.py
```diff
@@ -3797,13 +3821,117 @@ def insert_verity(
...
+
+        with open(args.secure_boot_certificate) as f:
+            certificate = crypto.load_certificate(crypto.FILETYPE_PEM, f.read())
+
+        # Note that the library returns the SHA256 digest in an
+        # uppecase format with : as byte separators. Let's convert to
+        # classic lowercase series of un-separated hex digits
+        fingerprint = certificate.digest("sha256").decode("ascii").replace(":", "").lower()
+
+        bio_in = crypto._new_mem_buf(root_hash.encode("utf-8"))
```
behrmann, line 3860: Using the internal functions of an external library does introduce the possibility of breakage, since they probably don't guarantee API stability between versions.

mkosi/__init__.py
```diff
@@ -3797,13 +3821,117 @@ def insert_verity(
...
+
+    with complete_step("Signing verity root hash…"):
+
+        PKCS7_NOCERTS  = 0x2    # Don't include signature certificates in result
+        PKCS7_DETACHED = 0x40   # Don't include data to sign in result
+        PKCS7_BINARY   = 0x80   # Don't mangle newlines for MIME canonical format
+        PKCS7_NOATTR   = 0x100  # Don't include signature time, … in result
+
+        with open(args.secure_boot_key) as f:
+            key = crypto.load_privatekey(crypto.FILETYPE_PEM, f.read())
```
behrmann, line 3850: With `args.secure_boot_key` being `Path` object this can more succinctly be
```suggestion
        key = crypto.load_privatekey(crypto.FILETYPE_PEM, args.secure_boot_key.read_text())
```

mkosi/__init__.py
```diff
@@ -3797,13 +3821,117 @@ def insert_verity(
...
+    if do_run_build_script or args.verity != "signed":
+        return None, None, None
+    if for_cache:
+        return None, None, None
+
+    assert root_hash is not None
+
+    with complete_step("Signing verity root hash…"):
+
+        PKCS7_NOCERTS  = 0x2    # Don't include signature certificates in result
```
bluca, line 3844: I need to double check, but this could be problematic when using cert chains (ie: the kernel keyring has a root cert, signing is done with a leaf)

mkosi/__init__.py
bluca: moving the import down where it is used sounds like a good solution to me

mkosi/__init__.py
```diff
@@ -640,8 +658,8 @@ def root_partition_name(
     # If no image id is specified we just return a descriptive string
     # for the partition.
     prefix = "System Resources" if usr_only else "Root"
-    if verity:
-        return prefix + " Verity"
+    if suffix is not None:
+        return prefix + " " + suffix
     return prefix + " Partition"
```
keszybz, line 663: `return prefix + ' ' + (suffix if suffix is not None else 'Partition')` ?

mkosi/__init__.py
keszybz: I think it'd be nicer to something like 
```
try: from OpenSSL import crypto
except ImportError: crypto=None
```
and only error out after parsing args below if crypto would be used.

keszybz (commented):
Looks very reasonable.

poettering:
systemd side for picking up these embedded signatures is in #20691

mkosi/__init__.py
poettering: Kernel example is with this flag set, and the verification code as far as I read it in the kernel doesn't care.

bluca: it will be fine for self-signed yeah, but does it still work if you sign with a leaf and have only the root in /etc/verity.d?

poettering: no, but the code i wrote does not follow certificate chains right now. given that secureboot in uefi afair also disables all certificate chains logic i am not sure we should enable it. 

or to say this differently: let's start with the cert chains stuff off, we can still enable this later on any day.

poettering:
this is no longer wip. Should be good for review/merge. PTAL!

mkosi/__init__.py
bluca: yeah that's fine, this is more of a reminder to myself to go and check what happens

mkosi/__init__.py
```diff
@@ -3607,13 +3634,116 @@ def insert_verity(
...
+    args: CommandLineArguments, root_hash: Optional[str], do_run_build_script: bool, for_cache: bool
+) -> Tuple[Optional[BinaryIO], Optional[bytes], Optional[str]]:
+
+    if do_run_build_script or args.verity != "signed":
+        return None, None, None
+    if for_cache:
+        return None, None, None
+
+    assert root_hash is not None
+    assert crypto is not None
```
bluca, line 3660: shouldn't this return and print an error, instead of asserting?

poettering: well, so far if any of our deps are missing (as in: binaries we want to call) we just fail with an uncaught exception, so my idea was here to expose the same behaviour for the openssl lib missing, after all this will trigger an uncaught exception here too.

keszybz: Please just return an error. That's much more user-friendly.

mkosi/__init__.py
```diff
@@ -3607,13 +3634,116 @@ def insert_verity(
...
+    if verity_sig is None:
+        return
+    if for_cache:
+        return
+    assert loopdev is not None
+    assert raw is not None
+    assert root_hash is not None
+    assert fingerprint is not None
+    assert args.verity_sig_partno is not None
+
```
keszybz, line 3729: That seems strange. Why not remove Optional from the types instead?

mkosi/__init__.py
```diff
@@ -3852,6 +3982,25 @@ def write_root_hash_file(args: CommandLineArguments, root_hash: Optional[str]) -
...
+        return None
+
+    assert args.output_root_hash_p7s_file is not None
+
+    suffix = roothash_p7s_suffix(args.usr_only)
+    with complete_step(f"Writing {suffix} file…"):
+        f: BinaryIO = cast(
+            BinaryIO,
+            tempfile.NamedTemporaryFile(
+                mode="w+b", prefix=".mkosi", dir=os.path.dirname(args.output_root_hash_p7s_file)
```
keszybz, line 3996: `args.output_root_hash_p7s_file.parent`

mkosi/__init__.py
```diff
@@ -3918,12 +4069,18 @@ def calculate_sha256sum(
         if root_hash_file is not None:
             assert args.output_root_hash_file is not None
             hash_file(f, root_hash_file, os.path.basename(args.output_root_hash_file))
+        if root_hash_p7s_file is not None:
+            assert args.output_root_hash_p7s_file is not None
+            hash_file(f, root_hash_p7s_file, os.path.basename(args.output_root_hash_p7s_file))
```
keszybz, line 4074: args.output_root_hash_p7s_file.name

mkosi/__init__.py
```diff
@@ -3918,12 +4069,18 @@ def calculate_sha256sum(
...
         if split_root is not None:
             assert args.output_split_root is not None
             hash_file(f, split_root, os.path.basename(args.output_split_root))
         if split_verity is not None:
             assert args.output_split_verity is not None
             hash_file(f, split_verity, os.path.basename(args.output_split_verity))
+        if split_verity_sig is not None:
+            assert args.output_split_verity_sig is not None
+            hash_file(f, split_verity_sig, os.path.basename(args.output_split_verity_sig))
```
keszybz, line 4083: ditto

mkosi.md
```diff
@@ -528,10 +528,24 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
 
 `Verity=`, `--verity`
 
-: Add an "Verity" integrity partition to the image. If enabled, the
-  root partition is protected with `dm-verity` against off-line
+: Add an "Verity" integrity partition to the image. Takes a boolean or
+  the special value `signed`, and defaults to disabled. If enabled,
+  the root partition (or `/usr/` partition, inc ase `UsrOnly=` is
```
keszybz, line 533: typo

mkosi.md
```diff
@@ -528,10 +528,24 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
...
+  enabled) is protected with `dm-verity` against off-line
```
keszybz, line 534: "offline"

mkosi.md
```diff
@@ -528,10 +528,24 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
...
   modification, the verification data is placed in an additional GPT
-  partition. Implies `ReadOnly=yes`.
+  partition. Implies `ReadOnly=yes`. If this is enabled the Verity
+  root hash is written to an output file with `.roothash` or
+  `.usrhash` suffix. If set to `signed` Verity is also enabled, but
```
keszybz, line 538: I'd put a comma after "enabled" and "`signed`"…

mkosi.md
```diff
@@ -528,10 +528,24 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
...
+  the resulting root hash is then also signed (in PKCS#7 format) with
+  the signature key configured with `SecureBootKey=` (or in other
```
keszybz, line 540: The part staring with "(or in other words" should be a new sentence.

mkosi.md
```diff
@@ -528,10 +528,24 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
...
+  words: the SecureBoot key pair is then both used to sign the kernel,
+  if that's enabled, and the root/`/usr/` file system). This signature
+  is then stored in an additiona output file with the `.roothash.p7s`
```
keszybz, line 543: typo

mkosi/__init__.py
behrmann: I think we might have this in a few other places as well and it is a potential footgun. python as a `-O` option (and `PYTHONOPTIMIZE` environment variable) that throws away asserts. I know that at least Arch builds its Python packages with this by default.

behrmann: On a related note, I think it's still better to move the import into this function and die with a nice error message on `ImportError`, since mypy doesn't like a potential module object being `None`. 

There's probably a nicer solution for this, but since this function is the only user of this module - at least for now -  I think this is a clean solution.

mkosi/__init__.py
```diff
@@ -3607,13 +3634,116 @@ def insert_verity(
...
+
+    with complete_step("Signing verity root hash…"):
+
+        key = crypto.load_privatekey(crypto.FILETYPE_PEM, args.secure_boot_key.read_text())
+        certificate = crypto.load_certificate(crypto.FILETYPE_PEM, args.secure_boot_certificate.read_text())
+
+        # Note that the library returns the SHA256 digest in an
+        # uppecase format with : as byte separators. Let's convert to
+        # classic lowercase series of unseparated hex digits
+        fingerprint = certificate.digest("sha256").decode("ascii").replace(":", "").lower()
```
behrmann, line 3670:
```suggestion
        fingerprint = certificate.digest(b"sha256").decode("ascii").replace(":", "").lower()
```

mkosi/__init__.py
```diff
@@ -3607,13 +3634,116 @@ def insert_verity(
...
+        return None, None, None
+    if for_cache:
+        return None, None, None
+
+    assert root_hash is not None
+    assert crypto is not None
+
+    with complete_step("Signing verity root hash…"):
+
+        key = crypto.load_privatekey(crypto.FILETYPE_PEM, args.secure_boot_key.read_text())
```
behrmann, line 3664:
```suggestion
        key = crypto.load_privatekey(crypto.FILETYPE_PEM, args.secure_boot_key.read_bytes())
```

behrmann (commented):
@poettering could you add `types-openssl-python` to `ci-unit-test.yml`, so that
```yaml
    - name: Install
      run: |
        python3 -m pip install pytest mypy types-openssl-python isort pyflakes
        npm install -g pyright
```

I've only quickly ran mypy with this and will probably not manage to have a further look at this before next week, but here are the easy to fix typing changes (I have never used the openssl API, though, so beware, I don't *know* whether the typing information is correct). Beyond that, mypy mostly stumbles over all the typing for internal functions that it doesn't have. For now the easiest solution would be to smother all of this with type ignore hints, but we can probably add some typing later and maybe upstream that to typeshed.

keszybz:
This needs another round…

mkosi/__init__.py
behrmann: Ok, this one is spurious.

mkosi/__init__.py
poettering: @keszybz What do you mean "return an error"? you mean call `die()` with an explanatory message?

mkosi/__init__.py
poettering: I don't follow? remove it where precisely? from the function parameter list?

That's because we want that insert_verity_sig() may be called when verity is off, and is a NOP then. i.e. the first line in this function checks for this case indirectly.

I mean, all our functions that encapsulate image building steps are put together that way: they can be called in any config, and just become nops when what they are supposed to do doesn't apply because config turned something off or something not on.

And in that case all the function args that carry the actual data are typically None, hence we must allow Optional[]. But then in the case where we actaully need to do something we have to assert that they are all set (because otherwise mypy will whine).

I mean, you could move all the conditionalizations outside of these functions, but that would mean build_image() would become super hard to follow, due to all the externalized conditionalizations. Moreover, the precise conditionalizations are often not trivial hence would suck if not encapsulated in the functions.

Or maybe I don't get at all, what you are trying to say?

behrmann:
I've opened poettering/mkosi#4, since this was a bit easier than doing this via the review interface. I think exchanging [pyopenssl](https://github.com/pyca/pyopenssl) for [cryptography](https://github.com/pyca/cryptography) is were the ecosystem is going and what upstream itself recommends. The backend is, as far as I understand, openssl, too, but the API seems a lot nicer and it is already fully typed. 

With that the typing checks run through, but for some reason the unit tests are still borked.

poettering:
I force pushed a new version, still using the OpenSSL thing, but with all issues supposedly fixed.

Using the OpenSSL module indeed seems a bit awkward, if there's now something more pythonic, that's great. I was wondering about that when I picked OpenSSL — which I picked simply for the reason it has the most hits on google.

(the "genkey" code should be ported over too btw)

poettering:
not sure i grok the import ordering thing? is it that isort doesn't like imports inside of functions?

behrmann:
> not sure i grok the import ordering thing? is it that isort doesn't like imports inside of functions?

No, that is `isort` pedantically disliking the empty line (line 40)

behrmann:
> I force pushed a new version, still using the OpenSSL thing, but with all issues supposedly fixed.
> 

There will be a few others. mypy will still stumble over the internals of `OpenSSL` (which could be silenced, with poettering/mkosi#4 being the alternative), but there's also errors in the CLI unit tests I haven't yet looked at (and which are also present in poettering/mkosi#4).

mkosi/__init__.py
```diff
@@ -3717,35 +3711,41 @@ def make_verity_sig(
...
     assert root_hash is not None
 
-    try:
-        from OpenSSL import crypto
-    except ImportError:
-        die("Python OpenSSL package is not installed.");
+    from cryptography import x509
+    from cryptography.hazmat.primitives import hashes
+    from cryptography.hazmat.primitives import serialization
+    from cryptography.hazmat.primitives.serialization import pkcs7
```
behrmann, line 3717: This is what `isort` will want to have, I didn't check that before
```suggestion
    from cryptography.hazmat.primitives import hashes, serialization
    from cryptography.hazmat.primitives.serialization import pkcs7
    from cryptography import x509
```

mkosi/__init__.py
keszybz: Oh, OK, I didn't look at the `is None` check above. So it's OK as is.

poettering:
Force pushed a new version now, incorporating @behrmann's work on python-cryptography. Alas, it doesn't work anymore. Something is broken with inserting the verity-sig partition after we put it together. I have the strong suspicion that #791 is at fault for that.

keszybz:
>  I have the strong suspicion that #791 is at fault for that.

I'll take a look tomorrow.

keszybz:
Sorry for the delay. I made a huge refactoring of the sfdisk code, but it still needs some debugging. I'm not working today, but hopefully it'll be ready on Monday.

keszybz:
I force-pushed here, I hope that's OK.

The code is refactored a lot, but I think it's a lot clearer now. It should be easier to add further manipulations to the partition table later on.

.github/workflows/ci-unit-test.yml
```diff
@@ -16,7 +16,7 @@ jobs:
 
     - name: Install
       run: |
-        python3 -m pip install pytest mypy isort pyflakes
+        python3 -m pip install pytest mypy types-openssl-python isort pyflakes
```
behrmann, line 19: With cryptography this is no longer necessary
```suggestion
        python3 -m pip install pytest mypy isort pyflakes
```

mkosi.md
```diff
@@ -528,10 +528,24 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
...
+  the root partition (or `/usr/` partition, in case `UsrOnly=` is
+  enabled) is protected with `dm-verity` against offline modification,
+  the verification data is placed in an additional GPT
+  partition. Implies `ReadOnly=yes`. If this is enabled, the Verity
+  root hash is written to an output file with `.roothash` or
+  `.usrhash` suffix. If set to `signed`, Verity is also enabled, but
+  the resulting root hash is then also signed (in PKCS#7 format) with
+  the signature key configured with `SecureBootKey=`. Or in other
+  words: the SecureBoot key pair is then both used to sign the kernel,
+  if that's enabled, and the root/`/usr/` file system. This signature
```
behrmann, line 542: Bikeshedding and a formatting fix
```suggestion
  words: the SecureBoot key pair is then used to both sign the kernel,
  if that is enabled, and the `root`/`/usr/` file system. This signature
```

mkosi/__init__.py
```diff
@@ -36,6 +36,7 @@
 import urllib.parse
 import urllib.request
 import uuid
+
```
behrmann, line 39: I think isort is complaining about this empty line.

behrmann (commented):
A first trivial look. I'll do the rest after getting some sleep.

mkosi/__init__.py
```diff
@@ -21,7 +21,6 @@
 import http.server
```
poettering, line 22: commit msg has a typo, but doesn't matter

mkosi/__init__.py
```diff
@@ -602,115 +568,46 @@ def root_partition_name(
...
+             'var', args.var_size, GPT_VAR, "Variable Data Partition", False),
+            (no_btrfs and args.tmp_size is not None,
+             'tmp', args.tmp_size, GPT_TMP, "Temporary Data Partition", False),
+            (not is_generated_root(args),
+             'root', args.root_size,
+             gpt_root_native(args.architecture, args.usr_only).root,
+             root_partition_name(args),
+             args.read_only)):
+
+        if condition and size is not None:
```
poettering, line 599: this means you check the size agaist None twice each time? via the list above and here again?

I presume the first field of the list is resolved at the moment the list is put together? i.e. greedely? by moving this into lambda funcs we could resolve it lazily? would that be nicer?

mkosi/__init__.py
```diff
@@ -721,9 +618,11 @@ def create_image(args: CommandLineArguments, for_cache: bool) -> Optional[Binary
         )
         output.append(f)
         disable_cow(f.name)
-        f.truncate(image_size(args))
+        disk_size = args.partition_table.disk_size()
+        f.truncate(disk_size)
 
-        exec_sfdisk(args, f)
+        if args.partition_table.partitions:
```
poettering, line 625: maybe move this check into run_sfdisk()?

mkosi/backend.py
```diff
@@ -270,23 +280,86 @@ def read(cls, loopdev: Path) -> PartitionTable:
...
         return (1 + pea_sectors) * self.sector_size
 
+    def disk_size(self) -> int:
+        return roundup(self.last_partition_offset() + self.footer_size(), self.grain)
+
+    def add(self,
+            label: str,
+            size: int,
+            type_uuid: uuid.UUID,
+            name: str,
```
poettering, line 312: hmm, so name and label? and the partition label is actually called 'name' confusing, no?

maybe rename "label" → "id", and "name" → "gpt_label"? or so?

mkosi/backend.py
```diff
@@ -214,6 +223,7 @@ class PartitionTable:
     first_lba: Optional[int]
 
     grain: int = 4096
+    _labelled: Dict[str, Partition] = dataclasses.field(default_factory=dict)
```
poettering, line 227: wouldn't it be nicer to have an `Enum` for the partition types we care about instead of a string? that way python and the various tools should be better at finding typos with these ids?

mkosi/backend.py
```diff
@@ -270,23 +280,86 @@ def read(cls, loopdev: Path) -> PartitionTable:
...
+               f'attrs={"GUID:60" if read_only else ""}',
+               f'name="{name}"',
+               f'uuid={part_uuid}' if part_uuid is not None else None]
+
+        self.partitions += [', '.join(filter(None, new))]
+        self.last_partition_sector = self.last_partition_offset() // self.sector_size + n_sectors
+
+        part = Partition(size, len(self.partitions))
+
+        if label is not None:
```
poettering, line 331: why make this optional? shouldn't this be a requirement?

if this was required then we could just store the entries in one data structure instead of two

mkosi/backend.py
```diff
@@ -210,6 +210,7 @@ class ManifestFormat(Parseable, enum.Enum):
 class Partition:
     size: int
     number: int
+    name: str
```
poettering, line 213: this is a descriptive text we only use for log messages, no? should be called `description` or `comment` or so maybe?

poettering:
hmm, the reworked version doesn't work at all for me unfortunately.

```
‣ Running second (final) stage…
‣  Basing off cached image usr-test.cache-pre-inst
‣  Copied cached image as /home/lennart/projects/mkosi/usr-test/.mkosi-66vkrqqt
  File "/usr/lib64/python3.9/runpy.py", line 197, in _run_module_as_main
  File "/usr/lib64/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/lennart/projects/mkosi/mkosi/__main__.py", line 32, in <module>
    main()
  File "/home/lennart/projects/mkosi/mkosi/__main__.py", line 26, in main
    run_verb(a)
  File "/home/lennart/projects/mkosi/mkosi/__init__.py", line 7390, in run_verb
    manifest = build_stuff(args)
  File "/home/lennart/projects/mkosi/mkosi/__init__.py", line 6820, in build_stuff
    image = build_image(args, root, manifest=manifest, do_run_build_script=False, cleanup=True)
  File "/home/lennart/projects/mkosi/mkosi/__init__.py", line 6523, in build_image
    refresh_partition_table(args, raw)
  File "/home/lennart/projects/mkosi/mkosi/__init__.py", line 635, in refresh_partition_table
    assert args.partition_table is not None
AssertionError
```
(when starting from an earlier cached image)

and:

```
‣  Setting up basic OS tree…
‣  (Unmounting image)
‣  (Detaching image file)
  File "/usr/lib64/python3.9/runpy.py", line 197, in _run_module_as_main
  File "/usr/lib64/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/lennart/projects/mkosi/mkosi/__main__.py", line 32, in <module>
    main()
  File "/home/lennart/projects/mkosi/mkosi/__main__.py", line 26, in main
    run_verb(a)
  File "/home/lennart/projects/mkosi/mkosi/__init__.py", line 7390, in run_verb
    manifest = build_stuff(args)
  File "/home/lennart/projects/mkosi/mkosi/__init__.py", line 6805, in build_stuff
    image = build_image(args, root, do_run_build_script=False, for_cache=True)
  File "/home/lennart/projects/mkosi/mkosi/__init__.py", line 6559, in build_image
    prepare_tree(args, root, do_run_build_script, cached)
  File "/home/lennart/projects/mkosi/mkosi/__init__.py", line 1418, in prepare_tree
    root.joinpath("efi/EFI").mkdir(mode=0o700)
  File "/usr/lib64/python3.9/pathlib.py", line 1323, in mkdir
    self._accessor.mkdir(self, mode)
FileNotFoundError: [Errno 2] No such file or directory: '/home/lennart/projects/mkosi/usr-test/mkosi.workspace/id34owbh/root/efi/EFI'
```

when starting fresh.

keszybz:
> hmm, the reworked version doesn't work at all for me unfortunately.

Please say what config and commands you are using…

poettering:
Here's the mkosi.default: https://paste.centos.org/view/faabd5d3

There's also an "mkosi.extra/" dir with some files, an "mkosi.version"  file with some version string, and a 
mkosi.secure-boot.{crt|key} pair

keszybz:
Oh, a typo indeed. I put `'efi'` in one place, and `'esp'` in another. I'll convert the code to use an enum, as suggested.

keszybz:
This should work now.

I still get a failure from SplitArtifacts=yes, but it is because there are no kernels under `/efi/EFI/Linux`. I assume that is not caused by this PR.

keszybz:
Oh, I see that I missed some comments. I'll push another version.

poettering:
Hmm, if I run this on the systemd main repo then it still doesn#t work: (i.e. `mkosi -i -f -f boot`)

```
‣ Running second (final) stage to generate cached copy…
  File "/usr/lib64/python3.9/runpy.py", line 197, in _run_module_as_main
  File "/usr/lib64/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/lennart/projects/mkosi/mkosi/__main__.py", line 32, in <module>
    main()
  File "/home/lennart/projects/mkosi/mkosi/__main__.py", line 26, in main
    run_verb(a)
  File "/home/lennart/projects/mkosi/mkosi/__init__.py", line 7204, in run_verb
    manifest = build_stuff(args)
  File "/home/lennart/projects/mkosi/mkosi/__init__.py", line 6619, in build_stuff
    image = build_image(args, root, do_run_build_script=False, for_cache=True)
  File "/home/lennart/projects/mkosi/mkosi/__init__.py", line 6339, in build_image
    raw = create_image(args, for_cache)
  File "/home/lennart/projects/mkosi/mkosi/__init__.py", line 608, in create_image
    assert args.partition_table is None
AssertionError
```

mkosi/__init__.py
keszybz: Applied, except for "`root`" → it shouldn't be quoted, because it's a name, not a path.

mkosi.md
keszybz: Applied, except for "root" → it shouldn't be quoted, because it's a name, not a path.

mkosi/__init__.py
keszybz: I think this would make the code more complicated for no gain. We go through the whole list anyway, so evaluating a few trivial conditional eagerly is probably faster then going lazy evaluation later on.

mkosi/__init__.py
keszybz: I thought about this, but decided to use the current form. In principle you can execute sfdisk even without any partitions, and it's a completely valid and meaningful operation. We don't want to do this here, because we only interact with the disk through partitions, so we only want to format in that case.

keszybz:
Hmm, it now builds, but doesn't boot. I need to investigate more.

keszybz:
I dropped a `not` in one of the conditionals… Let's see if it passes now.

keszybz:
(As discussed in the chat, we'll deal with the gpt header corruption later.)

PR #799, opened 2021-09-08
manifest: do not gather changelogs if they will not be used

keszybz:
I wanted to keep the data collection and serialization parts completely
separate, and I thought that gathering of the changelogs is insignificant.
But with a lot of packages, it can add up. So let's do it only if it'll
be useful later on.

A direct motivation is https://bugzilla.redhat.com/show_bug.cgi?id=2002038:
rpm gets confused by some package names and cannot show the changelog.

### DIFF
--- a/mkosi/manifest.py
+++ b/mkosi/manifest.py
@@ -7,7 +7,7 @@ from subprocess import DEVNULL, PIPE
 from textwrap import dedent
 from typing import IO, Any, Dict, List, Optional, cast
 
-from .backend import CommandLineArguments, PackageType, run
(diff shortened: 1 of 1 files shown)


keszybz:
arch/gpt_xfs couldn't find the boot entry, looks unrelated.

behrmann:
LGTM.

keszybz:
arch/gpt_xfs: `‣ Error: Container/VM address not found`
So I think it's an "unrelated failure". At least it passed the image-building phase where it was failing before.

PR #800, opened 2021-09-08
Gh action cleanup

Werkov:
Don't merge, just testing whether GH action environment can be stripped down.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,7 @@ jobs:
           - arch
           - centos_epel
           - debian
+          - opensuse
           - ubuntu
           - fedora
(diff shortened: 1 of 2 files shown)


keszybz:
Should this be closed or fixed?

DaanDeMeyer:
I think closed, if I remember correctly, I started building rpm from source because fedora packages were starting to use newer features that weren't in available old rpm.

Werkov:
I can be closed. (I tested whether openSUSE can be built using the Ubuntu's rpm package to reduce the time rebuilding it everytime. But that wouldn't work universally (e.g. the Fedora) and the proper solution seems to use GH action artifact caching.)

PR #802, opened 2021-09-08
WIP: support for creating sysexts

keszybz:
Please ignore the commits before `Merge branch 'smaller-gpt-areas` — they are included in #791.

The general idea is that you first generate a "sysext base" with something like:
```console
$ mkosi -o base.raw --format=gpt_squashfs
$ mkosi -o base.d --format=directory --clean-package-metadata=0
$ mkosi --build-env=SYSEXT=sshd -o sshd_sysext.raw --package='!*,openssh-server' --format=gpt_squashfs --sysext-base=mkosi.output/base.d
```
The first to builds are necessary because the "real" image should have package metadata removed, but to be usable as a base for the overlay, it needs the metadata to be there, so that the manager installs only the specified packages.

The `mkosi.finalize` script is used to insert appropriate extension-release.d file.

This obviously needs more work, but I think it's nice that the amount of new code needed for this is so tiny.

Questions:
- is the approach with `.overlayfs-workdir` and bind mounts appropriate? It seems to work, but is quite complex.
- should `mkosi` take care of the extension-release files, or should that be up to `mkosi.finalize`
- should this be "sysext-base" or "overlay-base" or something else? It could be used for non-sysext stuff too.

Right now the package list that is generated includes also packages from the base system. This is because we just do `rpm -qa`. We could query `sudo dnf history info`, but the output from this is a wall of text, which would need to be parsed. I'll ask if there's a nicer way to do this.


poettering:
Hmm, the requirement to have two different versions of the image around reminds me a bit of the two stages we need to build regular images, i.e. one image with devel headers and everything and the other without any of that. maybe this is the lesson to use here? i.e. base the sysext build image on the devel image of the base, and the final sysext packaging on the final image of the base? If so this would mean we'd have to teach mkosi to optionally leave the devel image around for this purpose. Right now we only keep this around if caching/incremental mode is enabled, but i guess we could leave the last version of the devel image "officially" around, too, if requested?

Or hmm, taking one step back, what about a quite different approach: instead of teaching mkosi the sysext/overlay concept as such, what about instead teaching it a "substraction image" — which would be an image we "substract" from the final image as last step, where "substracting" means: we'll delete any file in our newly built image that exists with the same contents (and inode metadata, but ignoring timestamps) already in the "substraction image".

The benefit of the latter is that building extensions and the original image is a bit more decoupled: the extension build process would after all be pretty much independent from the main image build process, only in this very last step we need a single file from the original build process: the image to use for substraction.

This would also mean we could build extensions without relying on overlayfs? which might be good for CI?

(Or thinking further: instead of operating with "subtraction images", we could just output a file manifest during build as one more resource to generate. Then, use that as operand for figuring out what to "subtract". i.e. we could probably with very little code generate a BSD mtree compatible output as `<image-name>.mtree`, and then people could copy that file (or symlink) into their sysext build env as `mkosi.substract` or so, and we'd use it as request to delete everything listed there)

keszybz:
> Hmm, the requirement to have two different versions of the image around reminds me a bit of the two stages we need to build regular images, i.e. one image with devel headers and everything and the other without any of that. maybe this is the lesson to use here? i.e. base the sysext build image on the devel image of the base, and the final sysext packaging on the final image of the base?

Yeah, that's something to explore definitely. I'm not entirely sure how all the cached images work, but I'll look into it.

> Or hmm, taking one step back, what about a quite different approach: instead of teaching mkosi the sysext/overlay concept as such, what about instead teaching it a "substraction image" — which would be an image we "substract" from the final image as last step, where "substracting" means: we'll delete any file in our newly built image that exists with the same contents (and inode metadata, but ignoring timestamps) already in the "substraction image".

The nice thing about the overlay approach is that it's very quick — in the few cases I tried so far, all the deps tend to be in the base image, so the overlay layer is just one or two packages. If instead we installed everything and then went for "subtraction", the whole process would be significantly slower… The workflow that I have in mind is that I'll have the config for the base initrd image, and then a set of ~10 sysexts with various package sets. So keeping the original image around in two copies is not a problem. Once you have more than 2 sysexts to build, it will start saving time and space. Dunno, I'm not too worried about the req for overlayfs.

keszybz:
Rebased to get rid of the stuff that was already merged.

mkosi/__init__.py
```diff
@@ -829,24 +829,22 @@ def reuse_cache_image(
 
 
 @contextlib.contextmanager
-def attach_image_loopback(
-    args: CommandLineArguments, raw: Optional[BinaryIO]
-) -> Generator[Optional[Path], None, None]:
-    if raw is None:
+def attach_image_loopback(image: Optional[BinaryIO]) -> Generator[Optional[Path], None, None]:
```
poettering, line 832: btw, just wondering, does python support giving names to return type annotations? i mean, input params have names, but output params do not?

mkosi/__init__.py
```diff
@@ -847,6 +847,21 @@ def attach_image_loopback(image: Optional[BinaryIO]) -> Generator[Optional[Path]
...
+
+    with complete_step(f"Using {sysext_base} as sysext base image"):
+        if sysext_base.is_dir():
+            yield sysext_base
+        else:
+            with sysext_base.open('rb') as f, \
+                 attach_image_loopback(f) as loopdev:
+
+                yield loopdev
+
```
poettering, line 864: hmm, shouldn't this happen read-only?

mkosi/__init__.py
```diff
@@ -1321,6 +1321,20 @@ def mount_overlay(
...
+    """Delete any char(0,0) device nodes underneath @root
+
+    Overlayfs uses such files to mark "whiteouts" (files present in
+    the lower layers, but removed in the upper one).
+    """
+
+    with complete_step("Removing overlay whiteout files…"):
+        for entry in cast(Generator[os.DirEntry[str], None, None], scandir_recursive(root)):
+            st = entry.stat(follow_symlinks=False)
+            if stat.S_ISCHR(st.st_mode) and st.st_rdev == 0:
```
poettering, line 1334: maybe encapsulate this in a helper func stat_is_whiteout()?

mkosi/__init__.py
```diff
@@ -1331,6 +1376,17 @@ def mount_image(
         yield
     finally:
         with complete_step("Unmounting image"):
+            if realroot is not None:
+                umount(root)
```
poettering, line 1394: you unmount root twice now... the next invocation of this is just a few lines below

mkosi/__init__.py
```diff
@@ -1331,6 +1376,17 @@ def mount_image(
...
+                umount(realroot)
+
+                workdir.cleanup()
+
+                # Let's now move the contents of realroot into root
+                for entry in os.scandir(realroot):
+                    os.rename(realroot / entry.name, root / entry.name)
+                realroot.rmdir()
```
poettering, line 1402: i am pretty sure this rename() loop should be under a new `with complete_step()` since it's slow and that deserves a spearate explanation given this is not really "unmounting" anymore

poettering: maybe move the body of this into a new helper function: move_dir_contents() or so?

poettering:
lgtm, just some small comments, see above.

poettering:
hmm, so i am pretty sure when creating the sysexts we should mount the "base" image read-only. That is hard though and requires extra work:

https://github.com/systemd/systemd/commit/b620bf332f575ba9b8e4cd60c93446a0c35c23e8

i.e. we probably need to use `norecovery` for mounting here.

mkosi/__init__.py
keszybz: I'm not sure if I understand the question. There's always just one return param, that can be None or a tuple… It does't have a name. But it is possible to define an alias for a type.

mkosi/__init__.py
keszybz: Yes, but I also mount it twice: the first mount is the overlayfs mount, and then I do a bind mount to adjust the location one level up…

poettering: ah, makes sense. Can you add a comment though?

mkosi/__init__.py
poettering: wel, i mean here we can see that the first input param of the function is the image. But looking at what it returns we have no such names, we really have no clue what this could be. I figured it would be kinda nice if input and output params would be a bit symmetric here: that you could name both.

mkosi/__init__.py
```diff
@@ -1387,6 +1401,8 @@ def mount_image(
                     os.rename(realroot / entry.name, root / entry.name)
                 realroot.rmdir()
 
+                delete_whiteout_files(root)
```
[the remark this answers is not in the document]
keszybz: I think it's reasonable to do this here. Those files are very specific to overlayfs, and since we're cleaning up after overlayfs here, it seems like the appropriate place to remove those too.

keszybz:
Updated. I implemented the suggested changes, except for making the mount read-only.
But to work properly, this needs changes for partition handling in #798.

poettering:
i don#t understand that breakage, but if it's safe to ignore, then looks good to merge to me. but of course this means CI will remain broken then?

behrmann:
Haven't yet had the time to look at the new developments here. What I assume pyright is complaining about in the case of `scandir_recursive` is that it (rightfully I think) doesn't know what to do with the type variable `T` and that it doesn't see a way to force the return type of `scandir_recursive` to be a `Path` for `find_files`. I think this should resolve this
```python
def scandir_recursive(
        root: Path,
        filter: Callable[[os.DirEntry[str]], Any] = lambda x: x,
) -> Generator[Path, None, None]:
    """Recursively walk the tree starting at @root, yield entries where @predicate returns not None"""
    queue: Deque[Union[str, Path]] = collections.deque([root])

    while queue:
        for entry in os.scandir(queue.pop()):
            pred = filter(entry)
            if pred is not None:
                yield Path(entry.path)
            if entry.is_dir(follow_symlinks=False):
                queue.append(entry.path)
```

This is somewhat stricter than what the docstring says, since it returns `Path`s and not the entry objects, but the current version also returns the predicate and not the entries. Was this intended?

For the `workdir` error I have to look at this more closely, this might be bogus.

keszybz:
Oh, right. Now I remember:
I though this should be written as
```py
filter: Callable[[os.DirEntry[str]], T]] = lambda x: x,
```
but mypy doesn't like this, or
```py
filter: Callable[[os.DirEntry[str]], T]] = lambda x: cast(T, x),
```
so in the end I gave up and replaced the T with Any.

But I think it's better to just make this
```py
filter: Optional[Callable[[os.DirEntry[str]], T]]] = None,
```

keszybz:
I'll rebase this on top of #798 when that's merged.

keszybz:
Updated to fix the type checks.

keszybz:
centos_epel/tar — network timeout
alma_epel/gpt_ext4,gpt_squashfs — network timeout

keszybz:
Updated: more docstrings and a description of the new option in the man page.
I also removed the dot from temporary directory names, as discussed above.

behrmann:
I like the changes and the CI failures seem unrelated to this (404 for Alma, CentOS and Rocky packages, timeouts during boot for ubuntu xfs and arch btrfs).

PR #803, opened 2021-09-08
Extract package manager build logic into separate script

DaanDeMeyer:
To make the action easier to debug, let's extract the logic for
building dnf and pacman into a script that we can run outside of
github actions. We also include an accompanying mkosi config that
can be used to test the script locally. Building the mkosi image
builds and installs all the package managers into the final image.

The final image can then be booted in QEMU, mkosi can be cloned and
tested using the installed package managers.

Usage:

- cd action/
- Repeat until build succeeds:
  - Modify setup-github-actions.sh
  - mkosi -i -f
- mkosi qemu
- systemctl start systemd-networkd systemd-resolved
- ln -sf /run/systemd/resolve/stub-resolv.conf /etc/resolv.conf
- git clone https://github.com/systemd/mkosi
- cd mkosi
- ./setup.py install
- mkosi -d \<distro\>

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -4,200 +4,16 @@ description: Install mkosi and all its dependencies
 runs:
   using: composite
   steps:
-  - name: apt-get update
-    shell: bash
-    run: sudo apt-get update
-
-  - name: apt-get install
-    shell: bash
-    run: sudo apt-get --assume-yes --no-install-recommends install
-          asciidoc
-          autopoint
-          check
-          debootstrap
-          docbook-xsl
-          e2fsprogs
-          gettext
-          libarchive-dev
-          libcap-dev
-          libcppunit-dev
-          libcurl4-openssl-dev
-          libdb-dev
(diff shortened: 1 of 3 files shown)


DaanDeMeyer:
Seems to be working so marking this ready for review.

behrmann:
Am I missing any differences from copying the build instructions into the shell script? 

Besides that there are some minor formatting nits (double spaces before line continuations and sometimes spaces between `]` and `;` in conditions), but I see nothing problematic.

DaanDeMeyer:
Fixed the formatting nits

I made everything idempotent so we don't rebuild everything when we're caching stuff using mkosi or GA in the future. Aside from that there's a few smaller changes but nothing major.

DaanDeMeyer:
Added one more change to update package versions to the latest available. Was supposed to be a separate commit but I messed up the rebase and it'll be a massive pain to split the commit so keeping it like this.

PR #804, opened 2021-09-09
use importlib for static resources

behrmann:
Now that we are on python 3.7, we can finally use [`importlib.resources`](https://docs.python.org/3.7/library/importlib.html#module-importlib.resources) and move static content into a submodule, which gives us proper editor support for those files, while making the main `__init__.py` a couple hundred lines shorter.

This could probably be combined with loading formatable strings from there and running `.format` on them later, but I haven't tried that yet, because I wanted to check first, whether we want to do this.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -123,6 +123,17 @@ DRACUT_SYSTEMD_EXTRAS = [
 ]
 
 
+def write_resource(
+        where: Path, resource: str, key: str, *, executable: bool = False, mode: Optional[int] = None
+) -> None:
+    text = importlib.resources.read_text(resource, key)
+    where.write_text(text)
+    if mode is not None:
+        where.chmod(mode)
+    elif executable:
+        make_executable(where)
+
+
 T = TypeVar("T")
 V = TypeVar("V")
 
@@ -1554,9 +1565,12 @@ def reenable_kernel_install(args: CommandLineArguments, root: Path) -> None:
     if not args.bootable or args.bios_partno is not None or not args.with_unified_kernel_images:
         return
 
-    hook_path = root / "etc/kernel/install.d/50-mkosi-dracut-unified-kernel.install"
-    hook_path.write_text(importlib.resources.read_text("mkosi.resources", "dracut_unified_kernel_install.sh"))
-    make_executable(hook_path)
+    write_resource(
+        root / "etc/kernel/install.d/50-mkosi-dracut-unified-kernel.install",
+        "mkosi.resources",
+        "dracut_unified_kernel_install.sh",
+        executable=True,
+    )
 
 
 def add_packages(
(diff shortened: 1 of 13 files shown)


DaanDeMeyer:
I had no idea this was possible. Neat!

mkosi/__init__.py
```diff
@@ -2621,161 +2552,35 @@ def install_arch(args: CommandLineArguments, root: Path, do_run_build_script: bo
...
-
-                [Action]
-                Description = Adding kernel and initramfs images to /boot...
-                When = PostTransaction
-                Exec = /etc/pacman.d/scripts/mkosi-kernel-add
-                NeedsTargets
-                """
-            )
+            importlib.resources.read_text("mkosi.resources.arch", "60_kernel_remove.hook")
         )
```
keszybz, line 2556: I think it'd make sense to add a helper for this:
```py
def write_resource(where: Path, resource: str, key: str, bool: executabe = False) -> None:
    text = importlib.resources.read_text(resource, key)
    where.write_text(text)
    if executable: …

write_resource(hooks_dir / '90-mkosi-kernel-add.hook', 'mkosi.resources.arch', '60_kernel_remove.hook')
```

```

**Comment by behrmann** [comment]
> EDIT: Actually, how does this work? does this mean we need to install all the resources alongside the python files? I don't see anything integrating this with setup.py on first sight so I'm curious how this would work.

Good question, I'm still in the process of figuring this out. :) So far I've only used this feature together with zipapps.

**Review by keszybz — COMMENTED** [review]
This is great.

File: mkosi/__init__.py (hunk 1 above)
Comment (behrmann) [reply mkosi/__init__.py:2556]: Will change it accordingly.

Comment (behrmann) [reply mkosi/__init__.py]: Added a `mode` switch as well.

**Comment by behrmann** [comment]
Ok, now it's hooked up to `setup.py` and the files are installed alongside accordingly.

While at it I removed the executable bit from two hook files in Arch, because I don't think they need to be executable and switched some `os.makedirs` to the pathlib calls.

**Comment by behrmann** [comment]
Just changed the explicit list of packages in `setup.py` for setuptools' `find_packages` to be futureproof.

File: mkosi/__init__.py [hunk 2]
```diff
@@ -2551,37 +2565,32 @@ def install_arch(args: CommandLineArguments, root: Path, do_run_build_script: bo
...
-        kernel_remove_script = scripts_dir / "mkosi-kernel-remove"
-        kernel_remove_script.write_text(importlib.resources.read_text("mkosi.resources.arch", "kernel_remove.sh"))
-        make_executable(kernel_remove_script)
+        write_resource(hooks_dir / "60-mkosi-kernel-remove.hook", "mkosi.resources.arch", "60_kernel_remove.hook")
+        write_resource(
+            scripts_dir / "mkosi-kernel-remove",
+            "mkosi.resources.arch",
+            "kernel_remove.sh",
+            executable=True,
+        )
```
keszybz: This really should be formatted as 
```py
       write_resource(hooks_dir / "60-mkosi-kernel-remove.hook", "mkosi.resources.arch", "60_kernel_remove.hook")
       write_resource(scripts_dir / "mkosi-kernel-remove", "mkosi.resources.arch", "kernel_remove.sh",
                                       executable=True)
```
Black makes it seem like two completely diferent things. OTOH, the PEP-8-style approach makes it easy to do a spot differences and similarities.

Obviously the same applies below: for no good reason, half of the calls to write_resource are compact, and half are spread.

keszybz (commented):
LGTM. I dislike the formatting, but I understand that this matches your preference, so I won't argue.

mkosi/__init__.py
behrmann: I'm not quite sold, that the PEP-8 style makes it easier to spot differences, but I'm generally not fond of overly long lines and prefer the more vertical style, as I find it more declarative. But this is something we can argue for a long time, now it's in the style my Emacs produces without interference :)

DaanDeMeyer:
All these install tests you wrote a long time ago will come in handy now. If everything passes this is good to merge.

poettering:
Amazeballs!

PR #806, opened 2021-09-10
Install extra trees after installing build results

### LINKED ISSUE #789: RFC: Switch order of mkosi.extra/ and copying build results?
In the case of systemd, copying the build results into the final image also copies a bunch of configuration files into the final image. We currently can't override these by using mkosi.extra/. I'm unsure if there's any big advantage to the current order so I propose we switch the order so we can override installed files from the build if needed.

DaanDeMeyer:
Fixes #789


behrmann:
I think the flow chart in the man page needs to be adjusted, no? "copy build results" moves before "copy extra trees".

DaanDeMeyer:
Updated manual with the necessary changes. @behrmann Heads up we added you to mkosi-maintainers so you can merge PRs now as well.

behrmann:
Yeah, got the mail. Thanks for added responsibility :)

poettering:
makes sense to me. In particular as we have the "skeleton" trees for the early stuff already.

lgtm

PR #815, opened 2021-09-18
Add minimal support for foreign packages in debian/ubuntu

mrogez:
With these changes, it is now possible to install packages coming from
other repository by declaring them in the Package setting, instead of
installing them during the postinst phase.

The trick is to populate the skeleton directory with the required GPG keys
and the corresponding repo to to sources.list.d.

For example with telegraf from influxdb:

```
mkdir -p mkosi.skeleton/usr/local/share/keyrings
curl https://repos.influxdata.com/influxdb.key | gpg --dearmor > mkosi.skeleton/usr/local/share/keyrings/influxdb-archive-keyring.gpg

mkdir -p mkosi.skeleton/etc/apt/sources.list.d
echo "deb [arch=amd64 signed-by=/usr/local/share/keyrings/influxdb-archive-keyring.gpg] https://repos.influxdata.com/debian buster stable" > mkosi.skeleton/etc/apt/sources.list.d/influxdb.list

cat > mkosi.default << EOF
[Distribution]
Distribution=debian
Release=buster

[Packages]
Packages=telegraf
EOF

sudo mkosi
```

The patch adds the installation of ca-certificates from debootstrap
stage, so that https URL are usable in extra sources.list.
Also, an apt-get update is needed before installing packages from
Packages setting, so that apt index takes into account extra repository
from sources.list.

These steps are done unconditionally because of simplicity.
If one wanted, we could include ca-certificates only if an extra
repo in the skeleton directory requires an https repo url.
And the apt-get update is only required if there are any extra repo in
skeleton directory.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2413,6 +2413,9 @@ def install_debian_or_ubuntu(args: CommandLineArguments, root: Path, *, do_run_b
         f"--components={','.join(repos)}",
     ]
 
+    # ca-certificates is needed for extra repos using https
+    cmdline += ["--include=ca-certificates"]
+
     if args.architecture is not None:
         debarch = DEBIAN_ARCHITECTURES.get(args.architecture)
         cmdline += [f"--arch={debarch}"]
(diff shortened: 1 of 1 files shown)


bluca:
Yeah make the extra calls dependent on there being external repos please, so that we don't slow every build down unnecessarily

keszybz:
@mrogez ^

DaanDeMeyer:
Closing for now due to lack of response. We can reopen again later.

PR #816, opened 2021-09-20
alma: add support for alma linux 8 with and without EPEL

### LINKED ISSUE #814: RFE: Add Alma Linux Support
Hot off the heels of adding Rocky Linux (#811), it should be straight forward to add support for Alma Linux.

I am happy to work on this one also

https://almalinux.org/

dellgreen:
- Closes #814

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
           - ubuntu
           - fedora
           - rocky_epel
+          - alma_epel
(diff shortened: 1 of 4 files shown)


behrmann:
LGTM.

DaanDeMeyer:
LGTM but we'll need to rework CI somehow so we run less jobs per distro as these new additions will strain the already limited CI capacity even more.

PR #818, opened 2021-09-23
dracut: make sure images with IMAGE_VERSION but without IMAGE_ID can …

poettering:
…boot

If an image version is set but no image ID mkosi's table currently
places generic labels in the partition labels, instead of the image ids
(because we have none...). This means we cannot refrence the root
partition via root=PARTLABEL=… on the kernel cmdline. Hence do not do
that.

This change ensures we don't try to use IMAGE_ID-based root=PARTLABEL=
kernel cmdline swtches without IMAGE_ID being set.

(Note that the main reason IMAGE_ID/IMAGE_VERSION exists is to allow
versioned setups, i.e. where multiple versions of the same thing exist.
In such a case it's important to reference the right rootfs version that
matches the whole setup we are building here. But if IMAGE_ID isn't set
then this multi-version logic is not desired and we can assume that only
a single version of the OS is in the partition table, and thus rely on
gpt-auto-generators automatic root file system discovery)

### DIFF
--- a/mkosi/resources/dracut_unified_kernel_install.sh
+++ b/mkosi/resources/dracut_unified_kernel_install.sh
@@ -15,7 +15,13 @@ fi
 PREFIX=$(dirname $(dirname "$BOOT_DIR_ABS"))
 
 # Pick a default prefix name for the unified kernel binary
-if [[ -z "$IMAGE_ID" ]] ; then
+if [[ -n "$IMAGE_ID" ]] ; then
+    if [[ -n "$IMAGE_VERSION" ]]; then
+        PARTLABEL="${IMAGE_ID}_${IMAGE_VERSION}"
+    else
+        PARTLABEL="${IMAGE_ID}"
+    fi
+else
     IMAGE_ID=linux
 fi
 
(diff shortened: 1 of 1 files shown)


behrmann:
LGTM.

poettering:
CI failures appear unrelated. Let's merge

mkosi/resources/dracut_unified_kernel_install.sh
```diff
@@ -15,7 +15,13 @@ fi
...
 # Pick a default prefix name for the unified kernel binary
-if [[ -z "$IMAGE_ID" ]] ; then
+if [[ -n "$IMAGE_ID" ]] ; then
+    if [[ -n "$IMAGE_VERSION" ]]; then
+        PARTLABEL="${IMAGE_ID}_${IMAGE_VERSION}"
+    else
+        PARTLABEL="${IMAGE_ID}"
+    fi
+else
     IMAGE_ID=linux
```
bluca, line 25: mmh `PARTLABEL` is not set in this branch, no? shouldn't this be changed to `PARTLABEL=linux` ?

poettering: no. the idea is if the PARTLABEL env var is set we use it to rference the root fs. But if we have no IMAGE_ID then the partition label is not supposed to be an "id" but just some string that nice for humans to read, and apparently people of no intention to make the image or partitions identifiable with it. Thus in that case we want to rely on systemd-gpt-auto-generator's logic of automatically finding the roots for you, by looking for a any suitable partition on the disk with the ESP on it.

bluca: But then PARTLABEL should be set as an empty string at least, no? Otherwise it's used while undefined, which is not great and would make shellcheck angry:

```
BOOT_OPTIONS="${BOOT_OPTIONS} root=PARTLABEL=${PARTLABEL}"
```

bluca: actually that's not true, shellcheck is happy with -n $undefined, so ignore me

PR #819, opened 2021-09-23
Add shellcheck for embedded shell scripts

behrmann:
Since @bluca just mentioned shellcheck in #818 I thought it might be a good time to add this. This is not yet all-encompassing, but it should get all the scripts that are factored out into the resource module.


bluca (approved):
Oh very nice, LGTM

keszybz:
Failure is unrelated.

Issue #823, opened 2021-10-06
Broken --autologin with Fedora 34


stellarhopper:
Until Fedora 33 (for the guest/root image), autologin worked fine. Upgrading to Fedora 34, this seems to be broken:

```
Fedora 34 (Thirty Four)
Kernel 5.15.0-rc2+ on an x86_64 (ttyS0)

fedora login: root (automatic login)

Password: login: timed out after 60 seconds
```

The getty service seems to be installed as before. There were some differences in systemctl status:

F33:
```
# systemctl status getty@tty1.service
● getty@tty1.service - Getty on tty1
     Loaded: loaded (/usr/lib/systemd/system/getty@.service; enabled; vendor pr>
    Drop-In: /etc/systemd/system/getty@tty1.service.d
             └─autologin.conf
     Active: active (running) since Tue 2021-09-07 17:15:25 MDT; 4 weeks 0 days>
       Docs: man:agetty(8)
             man:systemd-getty-generator(8)
             http://0pointer.de/blog/projects/serial-console.html
   Main PID: 720 (login)
     Memory: 1.3M
        CPU: 51ms
     CGroup: /system.slice/system-getty.slice/getty@tty1.service
             ‣ 720 login -- root

Sep 07 17:15:25 fedora systemd[1]: Started Getty on tty1.
Sep 07 17:15:26 fedora login[720]: pam_succeed_if(login:auth): requirement "tty>
Sep 07 17:15:26 fedora login[720]: pam_succeed_if(login:auth): requirement "tty>
Sep 07 17:15:27 fedora login[720]: pam_unix(login:session): session opened for >
Sep 07 17:15:27 fedora login[720]: ROOT LOGIN ON tty1
lines 1-19/19 (END)
```

F34:
```
# systemctl status getty@tty1.service 
● getty@tty1.service - Getty on tty1
     Loaded: loaded (/usr/lib/systemd/system/getty@.service; enabled; vendor pr>
    Drop-In: /etc/systemd/system/getty@tty1.service.d
             └─autologin.conf
     Active: active (running) since Wed 2021-10-06 00:06:19 MDT; 24s ago
       Docs: man:agetty(8)
             man:systemd-getty-generator(8)
             http://0pointer.de/blog/projects/serial-console.html
   Main PID: 770 (login)
     Memory: 1.1M
        CPU: 18ms
     CGroup: /system.slice/system-getty.slice/getty@tty1.service
             └─770 /bin/login -p --

[...]

PR #824, opened 2021-10-06
mkosi: use '/dev/' prefix for autologin setup in Fedora

stellarhopper:
Starting with Fedora 34, the PAM login config requires full dev paths
for the autologin setup. Add Fedora to the list of distros that require
this with a gate on releases >= 34 only.

This addresses #823


behrmann:
Not unreasonable, but if this depends on the version of PAM, do you know which version that is, because this looks like it's going to be a game of whack-a-mole. Might make more sense if we could gate this behind a version check for the PAM package or at least add a comment for the next person to stumble over this problem.

keszybz:
And what happens if the prefix is used on older releases? Could we just write it unconditionally everywhere?

stellarhopper:
> Not unreasonable, but if this depends on the version of PAM, do you know which version that is, because this looks like it's going to be a game of whack-a-mole. Might make more sense if we could gate this behind a version check for the PAM package or at least add a comment for the next person to stumble over this problem.

Yeah I was thinking along those lines. On Fedora, for 33->34, pam went from 1.4.0 to 1.5.1.

My python isn't nearly sharp enough though to plumb in that check :)

behrmann:
Good point, but it might just be a downstream patch, Debian tends to have quite a few. :)

poettering:
lgtm. 

(Somehow the CI got lost?)

stellarhopper:
@keszybz Yes good point - I pushed an update which composes the in-memory buffer, and then writes it out in one go. Thanks for the feedback.

PR #825, opened 2021-10-06
README: add repology "versions" badge

keszybz:


keszybz:
Through a very efficient and scientific process of pushing different formatting changes into this PR, I determined that the layout looks best if the badge is on the right and starts in parallel to the "Installation" section. When it was at the very top, it was mis-aligned wrt. to the first header, and when it started below "Installation", the "Installation" header was weirdly extended over the badge.

behrmann:
Neat, but is there a slimmed down version of that? Not sure we need to track all the old Fedora versions that are no longer supported or Devuan that will probably throw this package out as fast as possible as soon as anybody tells them this project might have a whiff of systemd in it :)

keszybz:
Please make your voice heard in https://github.com/repology/repology-webapp/issues/117.

keszybz:
Oh, Debuan. I wouldn't worry about it — I don't think it matters too much if they keep version 4 in their repo or not.

behrmann:
Will do, otherwise LGTM.

bluca:
@mbiebl do you need some help with https://tracker.debian.org/pkg/mkosi ? It's falling a bit behind

PR #826, opened 2021-10-06
Add Binds option

DaanDeMeyer:
Mounts the given files/directories into any nspawn containers spawned
by mkosi


DaanDeMeyer:
> The option is called Bind and BindReadOnly in .nspawn, and I think the same convention should be used here.

I did that initially but `Binds` seemed a bit awkward as the config file option so I changed it to `Mounts`. I don't really mind so let me change it back to `Binds`.

PR #829, opened 2021-10-06
mkosi: do not set umask(0)

### LINKED ISSUE #699: Some mkosi created files left world writable in the image
It seems like mkosi sets the python process umask to 000 causing all files created with python to be world writable by default.

For example `/etc/machine-id` is unlinked and recreated with open in python code at the end of image creation but as permissions aren't changed after that, it is left as 0666 in the final image and systemd generation for new machine-id during boot won't change file permissions.

There are others like this but I haven't checked all options/distros for these as this is probably easier to check/fix just by finding all places where files are created in mkosi.

All of these can of course be hotfixed in mkosi.finalize script with find but it would be nicer if mkosi set proper permissions when creating files.

The hotfix for reference is `find "$BUILDROOT" -perm -o+w -type f -print -exec chmod o-w "{}" \;`

keszybz:
This will have very bad results unless one controls *all* places where
a file can be created. This is possible in systemd and C code, but is
unlikely to work well here, where we have lots of high-level code and
call helpers which can also create files.

Also, it is much better to create files with convervative permissions,
i.e. at least honouring the umask, and only adjust them to more permissive
is controlled cases, than to create things world-writable and try to
remember to modify the permissions to something reasonable after the
fact.

Fixes #699.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -4167,12 +4167,22 @@ def save_cache(args: CommandLineArguments, root: Path, raw: Optional[str], cache
             shutil.move(cast(str, root), cache_path)  # typing bug, .move() accepts Path
 
 
-def _link_output(args: CommandLineArguments, oldpath: PathString, newpath: PathString) -> None:
+def _link_output(
+        args: CommandLineArguments,
+        oldpath: PathString,
+        newpath: PathString,
+        mode: int = 0o666,
+) -> None:
+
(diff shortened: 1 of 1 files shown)


behrmann:
The arch gpt images look somewhat misbehaving. Timeouts for ext4 and xfs and the ramdisk not being found for btrfs. I think they should be unrelated to this though.

PR #830, opened 2021-10-08
mkosi: fix dd sparse mess

poettering:


poettering:
See https://github.com/systemd/mkosi/pull/802#discussion_r724887643 for an explanation

PR #837, opened 2021-10-13
nspawn: Copy RLIMIT_CORE and RLIMIT_NOFILE in non-booted nspawn containers

daandemeyer:
Avoid surprises by copying open files and coredump limits from the user
running mkosi. Most noteably, this makes sure core dumps in non-booted
mkosi containers actually end up on the host as previously the coredump
size limit was zero in non-booted mkosi nspawn containers which led to
no coredumps being generated at all on the host of processes that dumped
core in the build containers (e.g. tests that raise SIGABRT).

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -78,6 +78,7 @@ from .backend import (
     die,
     install_grub,
     nspawn_params_for_blockdev_access,
+    nspawn_rlimit_params,
     patch_file,
     path_relative_to_cwd,
     run,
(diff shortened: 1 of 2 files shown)


poettering:
hmm, why muck with RLIMIT_NOFILE? That sounds unnecessary, and wrong?

(And we should add `--rlimit=RLIMIT_COREDUMP=inherit` to nspawn, so that you can express this natively)

poettering:
the soft limit is at 1K. But that really is a must, because select() can't deal with more and if you raise the soft limit beyond 1K then this means that fds might be created you cannot use in APIs that internally use select(). Thus, the soft limit should never be raised because otherwise you might silently trigger unexpected failure codepaths.

The hardlimit should be 4K right now in nspawn. Which is OK'ish. But yeah, we bumped the hard rlimit system-wide to 512K some time ago in systemd, so nspawn should probably do the same.

But i'd argue this is something for nspawn to figure out, not mkosi. 

So I think we should revert the part about RLIMIT_NOFILE. The RLIMIT_CORE one makes more sense I guess.

PR #840, opened 2021-10-15
Clean up repo management and add RemovePackages= option

keszybz:


poettering:
i.e. isn't "dnf autoremove" smart enough to do this for us? i.e. couldn't we just call tat immediately after installing the initial image?

PR #844, opened 2021-10-19
typing: use Iterator instead of Generator

keszybz:
Inspired by https://sethmlarson.dev/blog/2021-10-18/tests-arent-enough-case-study-after-adding-types-to-urllib3:
> Don’t expose Generators unless you want Generator functionality
>
> Generators have additional behaviors over iterables so if the API isn’t meant
> to be used like a generator then it’s best to keep this fact a secret and
> annotate with Iterable[X] instead of Generator[X, None, None].

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -48,8 +48,8 @@ from typing import (
     ContextManager,
     Deque,
     Dict,
-    Generator,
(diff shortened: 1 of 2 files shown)


keszybz:
The failures are in timouts in booting, should be unrelated.

PR #847, opened 2021-10-21
bin/mkosi: support running from the source checkout

keszybz:
It is super useful for development if we can use the helper and not worry about setting PYTHONPATH manually.

### DIFF
--- a/bin/mkosi
+++ b/bin/mkosi
@@ -9,16 +9,18 @@
 # home directory via "python3 -m pip install --user"
 #
 # We support installation via:
-# sudo python3 -m pip install <mkosi>
-# python3 -m pip install --user <mkosi>
-# python3 -m pip install --user --editable <mkosi>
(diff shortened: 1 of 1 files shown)


behrmann:
Maybe that's a silly question, but why not an editable install?

keszybz:
> Maybe that's a silly question, but why not an editable install?

Editable-install is per-user. Since mkosi is not useful as non-root, it seems silly to me to create an installation for the user, and then have root reach into the user config to figure out what code to execute. It's more straightforward and easy to execute directly from the checkout directory.

Also editable-install is per checkout directory. I'm a heavy user of git worktrees, and I want to be able to call mkosi from multiple checkout directories, e.g. `sudo ~/src/mkosi/bin/mkosi …` and `sudo ~/src/mkosi-work1/bin/mkosi …`. Editable installs would not help here.

PR #848, opened 2021-10-27
Set new TTYRows and TTYColumns properties when overriding getty units

DaanDeMeyer:
Sometimes, it's not sufficient to set the LINES and COLUMNS environment
variables to configure the serial terminal size. To properly configure
the serial terminal size when the env variables are not sufficient, we
need to configure the tty size in the kernel as well. To accomplish this,
two new properties TTYRows and TTYCols were added to systemd. Let's set
these properties when we override the getty units so systemd properly
configures the kernel tty size for us when systemd's version is recent
enough.

Additionally, this commit updates the getty unit resources to include
the recent systemd change to how the TTY is passed to agetty. Without
this change, the tty size configured with TTYRows and TTYCols is reset
when systemd invokes agetty. See systemd/systemd#21171
for more information.

### DIFF
--- a/mkosi/resources/getty_autologin.conf
+++ b/mkosi/resources/getty_autologin.conf
@@ -1,3 +1,4 @@
 [Service]
 ExecStart=
-ExecStart=-/sbin/agetty -o '-p -- \\u' --autologin root --noclear %I $TERM
+ExecStart=-/sbin/agetty -o '-p -- \\u' --autologin root --noclear - $TERM
+StandardInput=tty
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3119,6 +3119,8 @@ def set_serial_terminal(args: CommandLineArguments, root: Path, do_run_build_scr
                           Environment=TERM={os.getenv('TERM', 'vt220')}
                           Environment=COLUMNS={columns}
                           Environment=LINES={lines}
+                          TTYColumns={columns}
+                          TTYRows={lines}
                           """)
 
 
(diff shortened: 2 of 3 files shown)


DaanDeMeyer:
Updated to use the new properties that will be added by https://github.com/systemd/systemd/pull/21171. This means the fix will only work on distros that have the corresponding util-linux fix (https://github.com/karelzak/util-linux/pull/1485) available but this seems better than merging what is essentially a hack.

@bluca Would that work for you? I'm not sure what you're running in mkosi but I'm guessing it's Debian testing so the next util-linux update should be available there relatively soonish as well no?

bluca:
Yeah that's perfectly fine, thanks for fixing this!

DaanDeMeyer:
PRs are merged but let's wait until the util-linux fix is included in a new release before merging this since it's not really useful without the util-linux fix

bluca:
> PRs are merged but let's wait until the util-linux fix is included in a new release before merging this since it's not really useful without the util-linux fix

It won't hurt either, right? If not, I'd just merge it, given all other parts are merged too in their respective upstreams

behrmann:
I agree with @bluca. Better merge it now before we forget about it, since it won't hurt anybody and this way it's useful as soon as stuff lands downstream (also, we don't have to decide when it useful enough to merge).

DaanDeMeyer:
Arch failures are unrelated,

mkosi/resources/getty_autologin.conf
```diff
@@ -1,3 +1,4 @@
 [Service]
 ExecStart=
-ExecStart=-/sbin/agetty -o '-p -- \\u' --autologin root --noclear %I $TERM
+ExecStart=-/sbin/agetty -o '-p -- \\u' --autologin root --noclear - $TERM
+StandardInput=tty
```
poettering, line 4: StandardOutput=tty would be good here too, exactly like over in systemd

PR #849, opened 2021-10-31
Call flush() in a few more places

DaanDeMeyer:
Since we never explicitly close these files, sometimes the stuff
we write to them doesn't seem to get flushed which causes the
files to be empty in the output directory. Add calls to flush()
to make sure the files aren't empty when mkosi finishes.


behrmann:
LGTM.

keszybz:
Yeah, this looks like something we should always do.

PR #850, opened 2021-11-02
ssh: add agent-socket and port parameters

bluca:


mkosi.md
```diff
@@ -1143,6 +1143,13 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
   the same location, suffixed with `.pub` (as done by `ssh-keygen`). If this
   option is not present, `mkosi` generates a new key pair automatically.
 
+`SshAgentSocket=`, `--ssh-agent-socket=`
+
+: If specified, use the given socket to connect to the ssh agent when building an
+  image and when conncting via `mkosi ssh` instead of hard-coding a key. The
```
DaanDeMeyer, line 1149: conncting => connecting

mkosi/__init__.py
```diff
@@ -6610,6 +6616,11 @@ def setup_ssh(
     if args.ssh_key:
         f = open(args.ssh_key, mode="r", encoding="utf-8")
         copy_file(f"{args.ssh_key}.pub", authorized_keys)
+    elif args.ssh_agent_socket is not None:
+        env = {"SSH_AUTH_SOCK": args.ssh_agent_socket}
+        result = run(["ssh-add", "-L"], env=env, check=True, text=True, stdout=subprocess.PIPE)
```
DaanDeMeyer, line 6644: `check=True` is the default if I remember correctly so the `check=True` param can be removed

DaanDeMeyer (commented):
Can we make this work as a boolean? Maybe automatically derive the ssh-agent socket path from `SUDO_USER` or something like that when the option is enabled?

mkosi/__init__.py
bluca: I'm not sure, I used it because it's used in many places (search for check=True)

DaanDeMeyer: Yeah, just check the definition of run() in backend.py, it's the default. I'll do a commit to fix the other callsites

bluca: ok, removed

bluca:
> Can we make this work as a boolean? Maybe automatically derive the ssh-agent socket path from `SUDO_USER` or something like that when the option is enabled?

It varies depending on the ssh agent being used - for example I use gpg, not ssh, as my ssh keys as derived from the gpg auth subkey stored on my yubikey. The login environment exports it via `$SSH_AUTH_SOCK`, but you lose it once you sudo I think

DaanDeMeyer:
> It varies depending on the ssh agent being used - for example I use gpg, not ssh, as my ssh keys as derived from the gpg auth subkey stored on my yubikey. The login environment exports it via $SSH_AUTH_SOCK, but you lose it once you sudo I think

It's possible with `sudo -E` I think. Eventually I'd really like if we could run the build step of mkosi without requiring root at all. Hopefully we can add an nspawn --rootless option similar to podman someday.

Maybe split this up into two options? "--use-ssh-agent" to use the ssh agent instead of generating the SSH key, and "--ssh-agent-socket" to specify the path. If no path is specified, check if the `SSH_AUTH_SOCK` env variable is set and if it is, get the path from there. That way this can be used both with regular sudo (but you have to specify the --ssh-agent-socket option) and sudo -E.

mkosi/__init__.py
```diff
@@ -7299,7 +7333,10 @@ def run_qemu(args: CommandLineArguments) -> None:
     if args.network_veth:
         if not ensure_networkd(args):
             # Fall back to usermode networking if the host doesn't have networkd (eg: Debian)
-            cmdline += ["-nic", "user,model=virtio-net-pci"]
+            fwd = ""
+            if args.ssh_port is not None:
+                fwd = f",hostfwd=tcp::{args.ssh_port}-:{args.ssh_port}"
```
behrmann, line 7338:
```suggestion
            fwd = f",hostfwd=tcp::{args.ssh_port}-:{args.ssh_port}"  if args.ssh_port is not None or ""
```
This should be within our linelength limit.

mkosi/__init__.py
```diff
@@ -7391,7 +7431,7 @@ def find_address(args: CommandLineArguments) -> Tuple[str, str]:
                 for neighbor in neighbors:
                     dst = cast(str, neighbor["dst"])
                     if dst.startswith("fe80"):
-                        return dev, dst
+                        return "%" + dev, dst
```
behrmann, line 7434:
```suggestion
                        return f"%{dev}", dst
```

bluca:
> > It varies depending on the ssh agent being used - for example I use gpg, not ssh, as my ssh keys as derived from the gpg auth subkey stored on my yubikey. The login environment exports it via $SSH_AUTH_SOCK, but you lose it once you sudo I think
> 
> It's possible with `sudo -E` I think. Eventually I'd really like if we could run the build step of mkosi without requiring root at all. Hopefully we can add an nspawn --rootless option similar to podman someday.

Unfortunately until the kernel gives us namespaces for loop devices and device mapper, that can't work, because we can't mount the images. It could work for the `directory` type output.

> Maybe split this up into two options? "--ssh-use-agent" to use the ssh agent instead of generating the SSH key, and "--ssh-agent-socket" to specify the path. If no path is specified, check if the `SSH_AUTH_SOCK` env variable is set and if it is, get the path from there. That way this can be used both with regular sudo (but you have to specify the --ssh-agent-socket option) and sudo -E.

Yes indeed, sudo -E works - I've changed the option to be a bool/path union. If `true` is passed then it will try to parse the env var automatically, which works if one uses `sudo -E`. If it's a path, then it's interpreted as a path to the socket to use. Seems nicer than having two separate options.

mkosi.md
```diff
@@ -1143,6 +1143,14 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
   the same location, suffixed with `.pub` (as done by `ssh-keygen`). If this
   option is not present, `mkosi` generates a new key pair automatically.
 
+`SshAgent=`, `--ssh-agent=`
+
+: If specified as a path, use the given socket to connect to the ssh agent when
+  building an image and when connecting via `mkosi ssh` instead of hard-coding
+  a key. If specified as `true`, `$SSH_AUTH_SOCK` will be parsed instead (hint:
+  use `sudo` with `-E`). The available public keys will be installed at build
+  time from the output of `ssh-add -L`.
```
keszybz, line 1152: That last sentence is awkward. Maybe "The keys listed by `ssh-add -L` will be installed as authorized_keys in the built image." ?

mkosi/__init__.py
```diff
@@ -4848,6 +4848,23 @@ def parse_remove_files(value: str) -> List[str]:
...
+        if not parse_boolean(value):
+            return False
+    except ValueError:
+        return Path(value)
+
+    sock = os.getenv("SSH_AUTH_SOCK")
+    if sock is None:
+        die(
+            "--ssh-agent=true but cannot parse $SSH_AUTH_SOCK (consider running 'sudo' with '-E')"
+        )  # NOQA: E501
```
keszybz, line 4864: Can we please write this as `die("message")` ? This black misformatting is madness.

mkosi/__init__.py
```diff
@@ -6613,6 +6638,11 @@ def setup_ssh(
     if args.ssh_key:
         f = open(args.ssh_key, mode="r", encoding="utf-8")
         copy_file(f"{args.ssh_key}.pub", authorized_keys)
+    elif args.ssh_agent is not False:
+        env = {"SSH_AUTH_SOCK": args.ssh_agent}
+        result = run(["ssh-add", "-L"], env=env, text=True, stdout=subprocess.PIPE)
```
keszybz, line 6669: `check=True` ?

mkosi/__init__.py
```diff
@@ -6613,6 +6638,11 @@ def setup_ssh(
...
+        with open(authorized_keys, "w") as f:
+            f.write(result.stdout)
```
keszybz, line 6671: `authorized_keys.write_text(…)` ?

mkosi/__init__.py
```diff
@@ -7408,36 +7438,36 @@ def find_address(args: CommandLineArguments) -> Tuple[str, str]:
...
-    assert ssh_key is not None
+    cmd = [
+            "ssh",
+            # Silence known hosts file errors/warnings.
+            "-o",
+            "UserKnownHostsFile=/dev/null",
+            "-o",
+            "StrictHostKeyChecking=no",
+            "-o",
+            "LogLevel ERROR",
```
keszybz, line 7479: It's be more readable if each `'-o', arg` pair was on the same line…

mkosi/__init__.py
keszybz: Oh, I see it's the default. NVM.

mkosi/__init__.py
bluca: sure, just cargo-culted it over

mkosi/__init__.py
```diff
@@ -6610,6 +6662,11 @@ def setup_ssh(
     if args.ssh_key:
         f = open(args.ssh_key, mode="r", encoding="utf-8")
         copy_file(f"{args.ssh_key}.pub", authorized_keys)
+    elif args.ssh_agent is not False:
+        env = {"SSH_AUTH_SOCK": args.ssh_agent}
+        result = run(["ssh-add", "-L"], env=env, text=True, stdout=subprocess.PIPE)
+        authorized_keys.write_text(result.stdout)
+        f = cast(TextIO, result.stdout)
```
bluca, line 6669: not sure this is legal/sensible...

mkosi/__init__.py
```diff
@@ -4848,6 +4848,21 @@ def parse_remove_files(value: str) -> List[str]:
...
+    """Can be True/False or a path to a socket."""
+
+    try:
+        if not parse_boolean(value):
+            return False
+    except ValueError:
+        return Path(value)
+
+    sock = os.getenv("SSH_AUTH_SOCK")
+    if sock is None:
```
behrmann, line 4861: Is it sensible if `SSH_AUTH_SOCK` exists and is empty? If not this might be better served as
```suggestion
    if not sock:
```

mkosi/__init__.py
```diff
@@ -6295,12 +6324,18 @@ def load_args(args: argparse.Namespace) -> CommandLineArguments:
...
+    if args.ssh and not args.base_packages:
+        die("--ssh cannot be installed without --base-packages")
+
+    if args.ssh and not args.network_veth and args.ssh_port is None:
+        die("--ssh cannot be used without --network-veth or --ssh-port")
 
     if args.ssh_timeout < 0:
         die("--ssh-timeout must be >= 0")
 
+    if args.ssh_port is not None and args.ssh_port <= 0:
```
behrmann, line 6336: Seeing this `not None` check, would it maybe make sense to give `ssh_port` a default of 22, so that we always know it has a sensible value?

bluca: ok, done

mkosi/__init__.py
DaanDeMeyer: Yeah this seems a bit iffy, can't you just set `f` to `None`, we return an `Optional` so that should just work I think?

mkosi.md
```diff
@@ -1143,6 +1143,14 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
   the same location, suffixed with `.pub` (as done by `ssh-keygen`). If this
   option is not present, `mkosi` generates a new key pair automatically.
 
+`SshAgent=`, `--ssh-agent=`
+
+: If specified as a path, use the given socket to connect to the ssh agent when
+  building an image and when connecting via `mkosi ssh` instead of hard-coding
+  a key. If specified as `true`, `$SSH_AUTH_SOCK` will be parsed instead (hint:
+  use `sudo` with `-E`). The keys listed by `ssh-add -L` will be installed as
+  authorized_keys in the built image.
```
keszybz, line 1152: "authorized keys" (no underscore), or "`authorized_keys`"

Oh, so this does two things: first it uses public keys to initialize `authorized_keys`, and then it uses the agent when connecting over ssh. I think the description should make it clearer.

mkosi/__init__.py
```diff
@@ -4848,6 +4848,21 @@ def parse_remove_files(value: str) -> List[str]:
...
+
+    try:
+        if not parse_boolean(value):
+            return False
+    except ValueError:
+        return Path(value)
+
+    sock = os.getenv("SSH_AUTH_SOCK")
+    if not sock:
+        die("--ssh-agent=true but cannot parse $SSH_AUTH_SOCK (consider running 'sudo' with '-E')")
```
keszybz, line 4862: "$SSH_AUTH_SOCK is not set"

mkosi/__init__.py
```diff
@@ -5368,6 +5383,14 @@ def create_parser() -> ArgumentParserMkosi:
         default=0,
         help="Wait up to SECONDS seconds for the SSH connection to be available when using 'mkosi ssh'",
     )
+    group.add_argument(
+        "--ssh-agent",
+        type=parse_ssh_agent,
+        default=False,
+        metavar="PATH",
+        help="If True, 'mkosi ssh' will use $SSH_AUTH_SOCK (hint: use sudo -E) to talk to the ssh agent socket "
+        "to fetch and use the private key. Can also be used to manually pass the path to the ssh agent socket.",
```
keszybz, line 5392: So this description is misleading. I primarily uses the agent to fetch the public key list…

mkosi/__init__.py
bluca: ok, done

mkosi.md
bluca: ok, updated

mkosi/__init__.py
bluca: nope:

```
mkosi/__init__.py:6669:13: error: Incompatible types in assignment (expression
has type "None", variable has type "TextIO")  [assignment]
            f = None
```

mkosi/__init__.py
```diff
@@ -5368,6 +5383,16 @@ def create_parser() -> ArgumentParserMkosi:
...
     )
+    group.add_argument(
+        "--ssh-agent",
+        type=parse_ssh_agent,
+        default=False,
+        metavar="PATH",
+        help="If True, 'mkosi build' will use $SSH_AUTH_SOCK (hint: use sudo -E) to talk to the ssh "
+        "agent to fetch the available public keys when building an image. `mkosi ssh` will pass the "
+        "variable through to ssh when logging in. This option can also be used to manually specify a "
+        "path to the ssh agent socket.",
```
keszybz, line 5394: Hmm, we generally keep the texts in `--help` fairly short. This is supposed to be reminder, not an explanation…
Why not just say "Path to the ssh agent socket, or true to use $SSH_AUTH_SOCK" ?

This is not markdown, so single quotes should be used, not backticks.

mkosi/__init__.py
```diff
@@ -7408,36 +7438,33 @@ def find_address(args: CommandLineArguments) -> Tuple[str, str]:
...
 
 def run_ssh(args: CommandLineArguments) -> None:
-    ssh_key = args.ssh_key or args.output_sshkey
-    assert ssh_key is not None
+    cmd = [
+            "ssh",
+            # Silence known hosts file errors/warnings.
+            "-o", "UserKnownHostsFile=/dev/null",
+            "-o", "StrictHostKeyChecking=no",
+            "-o", "LogLevel ERROR",
```
keszybz, line 7480: Not that it matters greatly, but maybe use "=" here too to match the other two lines?

mkosi/__init__.py
```diff
@@ -6323,12 +6330,15 @@ def load_args(args: argparse.Namespace) -> CommandLineArguments:
     if args.ssh and not args.base_packages:
         die("--ssh cannot be installed without --base-packages")
 
-    if args.ssh and not args.network_veth:
-        die("--ssh cannot be used without --network-veth")
+    if args.ssh and not args.network_veth and args.ssh_port == 22:
+        die("--ssh cannot be used without --network-veth or a custom --ssh-port")
```
keszybz, line 6334: This assumes you have ssh running locally… But if it's not, that should be OK, no?

mkosi.md
```diff
@@ -1152,6 +1152,11 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
   authorized keys in the built image. The `ssh` invocation done by `mkosi ssh`
   will inherit `$SSH_AUTH_SOCK` for authentication purposes.
 
+`SshPort=`, `--ssh-port=`
+
+: If specified, when `NetworkVeth=` is disabled, use this port to connect via
+  `mkosi ssh` on localhost.
```
keszybz, line 1158: It's a bit unclear if this is the originating port or the destination port. Also, is this actually conditional on NetworkVeth? Should this just say "In the image, sshd will be configured to listen on this port. 'mkosi ssh' will connect to this port." ?

bluca: ok, updated

mkosi/__init__.py
DaanDeMeyer: The error is because `f` is declared as `TextIO`  a little earlier in the function. Just change that to `Optional[TextIO]` and it should work

bluca: done - one day I will understand all this typing stuff. One day...

mkosi/__init__.py
```diff
@@ -6295,6 +6295,9 @@ def load_args(args: argparse.Namespace) -> CommandLineArguments:
     if args.skip_final_phase and args.verb != "build":
         die("--skip-final-phase can only be used when building an image using 'mkosi build'")
 
+    if args.ssh and not args.base_packages:
+        die("--ssh cannot be installed without --base-packages")
```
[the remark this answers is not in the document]
bluca: The issue is that it results in a very confusing error - but I'll drop it for now, it's not really related anyway, just something that I stumbled upon while testing this

mkosi/__init__.py
```diff
@@ -2581,7 +2581,7 @@ def install_debian_or_ubuntu(args: CommandLineArguments, root: Path, *, do_run_b
         # Debian still has pam_securetty module enabled, disable it in the base image.
         disable_pam_securetty(root)
 
-    if args.distribution == Distribution.debian and args.base_image is None:
+    if args.distribution == Distribution.debian and args.base_image is None and args.base_packages is not False:
```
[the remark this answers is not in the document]
bluca: I have updated this to check whether 'systemd' is in the package set instead, as this should be conclusive (note that there's no separate resolved package in debian/ubuntu)

mkosi/__init__.py
```diff
@@ -4848,6 +4848,21 @@ def parse_remove_files(value: str) -> List[str]:
     return ["/" + os.path.normpath(p).lstrip("/") for p in value.split(",") if p]
 
 
+def parse_ssh_agent(value: str) -> Union[Path, bool]:
+    """Can be True/False or a path to a socket."""
```
behrmann, line 4852: This isn't quite correct, no? The return value of this function will be either `False` or the path `SSH_AUTH_SOCK` points, too?

mkosi/__init__.py
```diff
@@ -4848,6 +4848,21 @@ def parse_remove_files(value: str) -> List[str]:
     return ["/" + os.path.normpath(p).lstrip("/") for p in value.split(",") if p]
 
 
+def parse_ssh_agent(value: str) -> Union[Path, bool]:
```
behrmann, line 4851: In this case the return annotation might make more sense as `Optional[Path]` instead of `Union[Path, bool]`. 
```suggestion
def parse_ssh_agent(value: str) -> Optional[Path]:
```

mkosi/__init__.py
```diff
@@ -5368,6 +5383,20 @@ def create_parser() -> ArgumentParserMkosi:
         default=0,
         help="Wait up to SECONDS seconds for the SSH connection to be available when using 'mkosi ssh'",
     )
+    group.add_argument(
+        "--ssh-agent",
+        type=parse_ssh_agent,
+        default=False,
```
behrmann, line 5389: The default should then be `None`
```suggestion
        default=None,
```

mkosi/__init__.py
```diff
@@ -6158,7 +6187,7 @@ def load_args(args: argparse.Namespace) -> CommandLineArguments:
         args.output_nspawn_settings = build_auxiliary_output_path(args, ".nspawn")
 
     # We want this set even if --ssh is not specified so we can find the SSH key when verb == "ssh".
-    if args.ssh_key is None:
+    if args.ssh_key is None and args.ssh_agent is False:
```
behrmann, line 6190: also
```suggestion
    if args.ssh_key is None and args.ssh_agent is None:
```

mkosi/backend.py
```diff
@@ -440,7 +440,9 @@ class CommandLineArguments:
     ephemeral: bool
     ssh: bool
     ssh_key: Optional[Path]
+    ssh_agent: Union[bool, Path]
```
behrmann, line 443:
```suggestion
    ssh_agent: Optional[Path]
```

tests/test_config_parser.py
```diff
@@ -134,6 +134,8 @@ def add_reference_config(self, job_name=DEFAULT_JOB_NAME):
             "ssh": False,
             "ssh_key": None,
             "ssh_timeout": 0,
+            "ssh_agent": False,
```
behrmann, line 137:
```suggestion
            "ssh_agent": None,
```

mkosi/__init__.py
```diff
@@ -6434,8 +6463,10 @@ def print_summary(args: CommandLineArguments) -> None:
         f"    Output nspawn Settings: {none_to_na(args.output_nspawn_settings if args.nspawn_settings is not None else None)}"
     )
     MkosiPrinter.info(
-        f"                   SSH key: {none_to_na((args.ssh_key or args.output_sshkey) if args.ssh else None)}"
+        f"                   SSH key: {none_to_na((args.ssh_key or args.output_sshkey or args.ssh_agent) if args.ssh else None)}"
```
behrmann, line 6466: This is actually the reason why I'm saying this, since a value of `False` wouldn't really mesh with `none_to_na`, but a value of `None` would.

behrmann (commented):
Finally had some time to properly read this. Looks pretty good to me, just one minor thought.

behrmann:
@bluca all of the above comments hinge on the initial assumption, please ignore the rest if it's wrong.

bluca:
> @bluca all of the above comments hinge on the initial assumption, please ignore the rest if it's wrong.

ok, changed False -> None

mkosi/__init__.py
```diff
@@ -4848,6 +4848,28 @@ def parse_remove_files(value: str) -> List[str]:
     return ["/" + os.path.normpath(p).lstrip("/") for p in value.split(",") if p]
 
 
+def parse_ssh_agent(value: Optional[str]) -> Optional[Path]:
+    """Will return None or a path to a socket."""
+
+    if value is None:
+            return None
```
behrmann, line 4855: The indentation is off here.

mkosi/__init__.py
```diff
@@ -4848,6 +4848,28 @@ def parse_remove_files(value: str) -> List[str]:
...
+    except ValueError:
+        if not os.path.exists(value):
+            die(f"Agent socket specified by --ssh-agent={value} cannot be found")
+        return Path(value)
+
+    sock = os.getenv("SSH_AUTH_SOCK")
+    if not sock:
+        die("--ssh-agent=true but $SSH_AUTH_SOCK is not set (consider running 'sudo' with '-E')")
+    if not os.path.exists(sock):
```
behrmann, line 4868: If you make `sock` a `Path` earlier, you can use `Path.is_socket` here.

mkosi/__init__.py
```diff
@@ -6592,7 +6630,22 @@ def setup_ssh(
...
+                s.write(
+                    dedent(
+                        f"""\
+                        [Socket]
+                        ListenStream=
+                        ListenStream={args.ssh_port}
+                        """
+                    )
+                )
+            os.chmod(override_dir / "port.conf", 0o644)
```
behrmann, line 6648: With #853 landed you can use `add_dropin_config` here.

behrmann (commented):
One minor issue and two comments, but apart from the indentation thing LGTM.

mkosi/__init__.py
bluca: ok, rebased and updated

mkosi/__init__.py
```diff
@@ -6590,7 +6628,15 @@ def setup_ssh(
...
-        unit = "ssh"
+        unit = "ssh.socket"
+
+        if args.ssh_port != 22:
+            add_dropin_config(root, unit, "port",
+                      f"""\
+                      [Socket]
+                      ListenStream=
+                      ListenStream={args.ssh_port}
+                      """)
```
behrmann, line 6639:
```suggestion
            add_dropin_config(root, unit, "port",
                              f"""\
                              [Socket]
                              ListenStream=
                              ListenStream={args.ssh_port}
                              """) 
```
I think this identation will make @keszybz happier :) (me, too for that matter)

behrmann (approved):
Barring CI, LGTM.

mkosi/__init__.py
bluca: updated

behrmann: Your speed always amazes me. Such roundtrip times make working a real pleasure. :)

bluca: Fastest `git push` in the West!

keszybz (approved):

PR #851, opened 2021-11-02
Remove check=True argument from calls to run()

DaanDeMeyer:
check=True is the default for our version of run() so let's not pass
check=True explicitly anywhere whenever we call run().


behrmann:
LGTM.

PR #852, opened 2021-11-04
manifest: record the architecture field separately

bluca:
We recently split this out in https://systemd.io/COREDUMP_PACKAGE_METADATA/
and the package-notes tooling, so split it out here too.


keszybz:
LGMT.

PR #856, opened 2021-11-12
Set loglevel to something sane when --qemu-headless is set

DaanDeMeyer:
Avoids spamming the console with kernel messages when using
--qemu-headless.


DaanDeMeyer:
cc @bluca

bluca (approved):
nice, thank you! was meaning to look into this as well

PR #857, opened 2021-11-23
Import updated GPT partition UUID list from systemd

DaanDeMeyer:
Source: https://github.com/systemd/systemd/blob/244cab4a651d4a5b5d9f0275ef59b8273418ee51/src/shared/gpt.h

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -197,38 +197,132 @@ def print_running_cmd(cmdline: Iterable[str]) -> None:
     MkosiPrinter.print_step(" ".join(shlex.quote(x) for x in cmdline) + "\n")
 
 
-GPT_ROOT_X86               = uuid.UUID("44479540f29741b29af7d131d5f0458a")  # NOQA: E221
-GPT_ROOT_X86_64            = uuid.UUID("4f68bce3e8cd4db196e7fbcaf984b709")  # NOQA: E221
-GPT_ROOT_ARM               = uuid.UUID("69dad7102ce44e3cb16c21a1d49abed3")  # NOQA: E221
-GPT_ROOT_ARM_64            = uuid.UUID("b921b0451df041c3af444c6f280d3fae")  # NOQA: E221
-GPT_USR_X86                = uuid.UUID("75250d768cc6458ebd66bd47cc81a812")  # NOQA: E221
-GPT_USR_X86_64             = uuid.UUID("8484680c952148c69c11b0720656f69e")  # NOQA: E221
-GPT_USR_ARM                = uuid.UUID("7d0359a302b34f0a865c654403e70625")  # NOQA: E221
-GPT_USR_ARM_64             = uuid.UUID("b0e01050ee5f4390949a9101b17104e9")  # NOQA: E221
-GPT_ESP                    = uuid.UUID("c12a7328f81f11d2ba4b00a0c93ec93b")  # NOQA: E221
-GPT_BIOS                   = uuid.UUID("2168614864496e6f744e656564454649")  # NOQA: E221
-GPT_SWAP                   = uuid.UUID("0657fd6da4ab43c484e50933c84b4f4f")  # NOQA: E221
-GPT_HOME                   = uuid.UUID("933ac7e12eb44f13b8440e14e2aef915")  # NOQA: E221
-GPT_SRV                    = uuid.UUID("3b8f842520e04f3b907f1a25a76f98e8")  # NOQA: E221
-GPT_XBOOTLDR               = uuid.UUID("bc13c2ff59e64262a352b275fd6f7172")  # NOQA: E221
-GPT_ROOT_X86_VERITY        = uuid.UUID("d13c5d3bb5d1422ab29f9454fdc89d76")  # NOQA: E221
-GPT_ROOT_X86_64_VERITY     = uuid.UUID("2c7357edebd246d9aec123d437ec2bf5")  # NOQA: E221
-GPT_ROOT_ARM_VERITY        = uuid.UUID("7386cdf2203c47a9a498f2ecce45a2d6")  # NOQA: E221
-GPT_ROOT_ARM_64_VERITY     = uuid.UUID("df3300ced69f4c92978c9bfb0f38d820")  # NOQA: E221
-GPT_USR_X86_VERITY         = uuid.UUID("8f461b0d14ee4e819aa9049b6fb97abd")  # NOQA: E221
-GPT_USR_X86_64_VERITY      = uuid.UUID("77ff5f63e7b64633acf41565b864c0e6")  # NOQA: E221
-GPT_USR_ARM_VERITY         = uuid.UUID("c215d7517bcd4649be906627490a4c05")  # NOQA: E221
-GPT_USR_ARM_64_VERITY      = uuid.UUID("6e11a4e7fbca4dedb9e9e1a512bb664e")  # NOQA: E221
-GPT_ROOT_X86_VERITY_SIG    = uuid.UUID("5996fc05109c48de808b23fa0830b676")  # NOQA: E221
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
I invented short identifiers for the new architectures because I couldn't find a list of short identifiers online anywhere. If someone knows where I can find such a list, please point me to it so we don't have to invent our own identifiers.

behrmann:
Don't those short names come from the kernel tree?

mkosi/__init__.py
```diff
@@ -197,38 +197,132 @@ def print_running_cmd(cmdline: Iterable[str]) -> None:
...
-GPT_ROOT_X86_64_VERITY_SIG = uuid.UUID("41092b059fc84523994f2def0408b176")  # NOQA: E221
-GPT_ROOT_ARM_VERITY_SIG    = uuid.UUID("42b0455feb11491d98d356145ba9d037")  # NOQA: E221
-GPT_ROOT_ARM_64_VERITY_SIG = uuid.UUID("6db69de629f44758a7a5962190f00ce3")  # NOQA: E221
-GPT_USR_X86_VERITY_SIG     = uuid.UUID("974a71c0de4143c3be5d5c5ccd1ad2c0")  # NOQA: E221
-GPT_USR_X86_64_VERITY_SIG  = uuid.UUID("e7bb33fb06cf4e818273e543b413e2e2")  # NOQA: E221
-GPT_USR_ARM_VERITY_SIG     = uuid.UUID("d7ff812f37d14902a810d76ba57b975a")  # NOQA: E221
-GPT_USR_ARM_64_VERITY_SIG  = uuid.UUID("c23ce4ff44bd4b00b2d4b41b3419e02a")  # NOQA: E221
-GPT_TMP                    = uuid.UUID("7ec6f5573bc54acab29316ef5df639d1")  # NOQA: E221
-GPT_VAR                    = uuid.UUID("4d21b016b53445c2a9fb5c16e091fd2d")  # NOQA: E221
+GPT_ROOT_ALPHA                  = uuid.UUID("6523f8ae3eb14e2aa05a18b695ae656f")  # NOQA: E221
```
keszybz, line 200: Do we still need the NOQA stuff?

mkosi/__init__.py
```diff
@@ -311,25 +405,77 @@ def gpt_root_native(arch: Optional[str], usr_only: bool = False) -> GPTRootTypeT
...
+        elif arch == "sparc":
+            return GPTRootTypeTriplet(GPT_USR_ARC, GPT_USR_ARC_VERITY, GPT_USR_ARC_VERITY_SIG)
+        elif arch == "armv7l":
+            return GPTRootTypeTriplet(GPT_USR_ARM, GPT_USR_ARM_VERITY, GPT_USR_ARM_VERITY_SIG)
+        elif arch == "aarch64":
+            return GPTRootTypeTriplet(GPT_USR_ARM64, GPT_USR_ARM64_VERITY, GPT_USR_ARM64_VERITY_SIG)
+        elif arch == "ia64":
+            return GPTRootTypeTriplet(GPT_USR_IA64, GPT_USR_IA64_VERITY, GPT_USR_IA64_VERITY_SIG)
+        elif arch == "mips":
+            return GPTRootTypeTriplet(GPT_USR_MIPS_LE, GPT_USR_MIPS_LE_VERITY, GPT_USR_MIPS_LE_VERITY_SIG)
```
keszybz, line 419: Hmm, I think this isn't good. The arch is usually called mipsel, to distinguish it from the big-endian variant. Please use the same identifiers that systemd uses.

(Unfortunately the identifiers in systemd are not the standard ones. For example, "mips-le" instead of the usual "mipsel". I think we should at some point change them, but we need to do it in a backwards-compatible fashion, so it's not so easy.)

keszybz (changes requested):

mkosi/__init__.py
DaanDeMeyer: I think we do, it flags multiple spaces before an operator which we still have

mkosi/__init__.py
DaanDeMeyer: Didn't know we had `uname_architecture()` right in the systemd tree. Updated to use the names from there

poettering: > (Unfortunately the identifiers in systemd are not the standard ones. For example, "mips-le" instead of the usual "mipsel". I think we should at some point change them, but we need to do it in a backwards-compatible fashion, so it's not so easy.)

if there was a "standard" that'd be great. Unfortunately noone can agree on anything. i.e. amd64 is what debian calls x86-64 and so on...

go calls "mipsle" what debian calls "mipsel" which we call "mips-le"... (see https://github.com/golang/go/commit/ad366fdbe40eea7f4fbdc75e8a27304c696793f1)

keszybz: Yeah, but Debian has been calling it mipsel already before go was a thing… So I don't see a reason to not follow the same naming in 2016.

poettering: I think this was done to be systematic, i.e. whenever there are two endianess flavours of some arch we used the _LE and _BE suffixes, to distinguish them.

mkosi/__init__.py
```diff
@@ -311,25 +405,77 @@ def gpt_root_native(arch: Optional[str], usr_only: bool = False) -> GPTRootTypeT
         arch = platform.machine()
 
     if usr_only:
-        if arch in ("i386", "i486", "i586", "i686"):
+        if arch == "alpha":
+            return GPTRootTypeTriplet(GPT_USR_ALPHA, GPT_USR_ALPHA_VERITY, GPT_USR_ALPHA_VERITY_SIG)
+        elif arch == "arc":
+            return GPTRootTypeTriplet(GPT_USR_ARC, GPT_USR_ARC_VERITY, GPT_USR_ARC_VERITY_SIG)
+        elif arch == "armv7l":
```
poettering, line 412: there are about ~500 other names to match here... see architecture.c over in systemd

DaanDeMeyer:
Not that I know of, I'm fine with releasing a new version.

keszybz:
https://github.com/systemd/mkosi/pull/858

mkosi/__init__.py
keszybz: Maybe `arch.startswith("armv")` ?
Or `re.match(r'armv[4-8]', arch)` ?

keszybz (changes requested):

keszybz:
See https://github.com/systemd/systemd/pull/21485 also.

mkosi/__init__.py
DaanDeMeyer: Went with `startswith` so we don't have to modify if they start inventing new numbers

DaanDeMeyer:
Also added loongarch64

behrmann:
LGTM.

keszybz:
```
--- Couldn't find '=sys-apps/baselayout-2.7' to unmerge.
‣   Fetching http://distfiles.gentoo.org/releases/amd64/autobuilds/20211121T170545Z/stage3-amd64-nomultilib-systemd-20211121T170545Z.tar.xz
‣   Extracting stage3-amd64-nomultilib-systemd-20211121T170545Z.tar.xz to /var/cache/distfiles/20211121T170545Z/stage3-amd64-nomultilib-systemd-20211121T170545Z.tar.xz.tmp
‣   Copying /var/cache/distfiles/20211121T170545Z/stage3-amd64-nomultilib-systemd-20211121T170545Z.tar.xz.tmp to /var/tmp/mkosi-gfyd7aii/root
‣   Invoking emerge(1) inside stage3
‣ Error: Workspace command /usr/bin/emerge =sys-apps/baselayout-2.7 --buildpkg=y --usepkg=y --keep-going=y --jobs=2 --load-average=1 --nospinner --quiet-build --quiet --unmerge returned non-zero exit code 1.
‣  (Unmounting Package Cache)
‣  (Unmounting image)
```
@257

257:
https://github.com/systemd/mkosi/pull/859

PR #858, opened 2021-11-23
NEWS: add entries for v11

keszybz:


NEWS.md
```diff
@@ -2,9 +2,43 @@
...
+  `changelog`, a longer human-readable file that shows package descriptions and
+  changelogs will be generated. This latter format should be considered
+  experimental and likely to change in later versions.
+- A new `RemovePackages=` option can be used to uninstall packages after the
+  build and finalize scripts have been done. This is useful for the case where
+  packages are required by the build scripts, or pulled in as dependencies
+  for scriptlets of other packages, but are not necessary in the final image.
+- A new `BaseImage=` option can be used to build "system extensions" a.k.a.
+  "sysexts" — partial images which are mounted on top of an existing system
+  to provide additional files under `/usr/`.
```
poettering, line 18: maybe add link to https://www.freedesktop.org/software/systemd/man/systemd-sysext.html

NEWS.md
```diff
@@ -2,9 +2,43 @@
...
+- A new `CleanPackageMetadata=` option can be used to force or disable the removal
+  of package manager files.
```
poettering, line 20: maybe mention what the default is

poettering:
lgtm.

keszybz:
Adjusted.

poettering:
One more note:

```
* The Verity= setting supports a new value "signed". If set the Verity data will be signed and the result be inserted as additional partition in the image. See https://systemd.io/DISCOVERABLE_PARTITIONS for details about signed disk images. This information is used by systemd-nspawn, systemd-dissect, systemd-sysext, systemd-portabled and systemd's RootImage= setting (among others) to cryptographically validate the image file systems before use.
* ```

File: NEWS.md [hunk 3]
```diff
@@ -2,9 +2,53 @@
...
+  significantly faster).
+- The `--debug=` option gained a new value `disk` to show information about disk
+  sized and partition allocations.
+- Some sections and settings have been renamed for clarity: [Packages] is now
+  [Content], `Password=`, `PasswordIsHashed=`, and `Autologin=` are now in
+  [Content]. The old names are still supported, but not documented.
+- When `--prepare-script=`/`--build-script=`/`--finalize-script=` is used with
+  an empty argument, the corresponding script will not be called.
+- Python 3.7 is the minimal supported version.
+- The Python `cryptography` module is used instead of `OpenSSL`.
```
behrmann, line 50: These could be marked as a note to packagers.

behrmann: Also, I don't think `OpenSSL` (as in the Python module of that name) needs to be mentioned, as it was only used during the development of the signed Verity PR, I'm not sure it was ever on the main branch.

We might also call this dependency optional for now, since it is only imported when signed Verity support is used.

keszybz: Updated.

behrmann:
LGTM

PR #859, opened 2021-11-24
gentoo: unmerge baselayout early on.

257:
Signed-off-by: Paymon MARANDI <darwinskernel@gmail.com>

### DIFF
--- a/mkosi/gentoo.py
+++ b/mkosi/gentoo.py
@@ -261,7 +261,7 @@ class Gentoo:
         self.set_useflags()
         self.mkosi_conf()
         self.baselayout(args, root)
-        self.fetch_fix_stage3(root)
+        self.fetch_fix_stage3(args, root)
         self.update_stage3(args, root)
         self.depclean(args, root)
 
@@ -269,7 +269,7 @@ class Gentoo:
                           root: Path) -> None:
         self.invoke_emerge(args, root, inside_stage3=False, actions=["--sync"])
 
-    def fetch_fix_stage3(self, root: Path) -> None:
+    def fetch_fix_stage3(self, args: CommandLineArguments, root: Path) -> None:
         """usrmerge tracker bug: https://bugs.gentoo.org/690294"""
 
         # e.g.:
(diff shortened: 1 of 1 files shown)


257:
hopefully we didn't miss v11, did we? i see the tag, i'm wondering if we could move the tag after this lands.

behrmann:
Well, no idea what the change is, but it does fix CI. The CI failures are unrelated.

DaanDeMeyer:
Can we get a little more detail in the commit message on what the problem was and how this fixes it?

257:
> Can we get a little more detail in the commit message on what the problem was and how this fixes it?

updated the commit msg.

behrmann:
@257 That commit message is quite helpful, but not knowing upstream Gentoo, can we expect more breakage once the mentioned bugs are fixed?

257:
yes and no. bulk of what we do is fixing the usrmerge, lots of code will be dropped once that's fixed upstream. so is baselayout. i don't think upstream would be keen on supporting all systemd's features though so that part of the code is very likely to stay around for a long time. 

also gentoo-releng, perhaps after [this](https://bugs.gentoo.org/792081), is working on standardizing the stage3 naming and signing the releases etc, even support for `machinectl`... those will effect us, mostly the fetch part of fetch_update_stage3(), but towards a more stable situation.

257:
[gentoo just merged app-admin/mkosi](https://github.com/gentoo/gentoo/commit/90bcb1fdec53af57cc4bfeb67975c633b7f9cf4e) but `v11` points to 6575c8236f1ee20bfeec8e9a35b36afbd79c64b1 which doesn't include this fix. could we move tag to this (5c77e786a673ed1a10e4008c19048fe0383100e7) please?

behrmann:
I guess that's a question for @keszybz, since he tagged v11. I'm not sure moving that tag around is the nicest solution, maybe tagging v12 right away would be better.

keszybz:
Yeah, we'd never "move" a tag that was made public. I'll just tag v12.

DaanDeMeyer:
> Yeah, we'd never "move" a tag that was made public. I'll just tag v12.

Slightly unfortunate since we're still waiting for https://github.com/systemd/mkosi/pull/862 to fix opensuse as well. I guess we'll release v13 when we merge that?

behrmann:
I guess that would make the most sense.

keszybz:
Oh, I was under the impression that #862 is mostly for the CI.
But anyway, it shouldn't be a problem to release again once that's fixed.

PR #860, opened 2021-11-24
Fix some Lintian warnings

bluca:


behrmann:
The typos look good to me. The executable bits don't hurt, but they're also somewhat unnecessary, I think, since the executable bit is set by `write_resource`. (It might be, that that's unnecessary, but then again it's nice to explicitly see this instead of having to look at the permissions.)

bluca:
> The typos look good to me. The executable bits don't hurt, but they're also somewhat unnecessary, I think, since the executable bit is set by `write_resource`. (It might be, that that's unnecessary, but then again it's nice to explicitly see this instead of having to look at the permissions.)

actually lintian complains because the installed scripts don't have +x either https://salsa.debian.org/systemd-team/mkosi/-/jobs/2214828

behrmann:
Ah, I think I didn't word this clearly enough. These scripts don't *need* to be executable at that location, because they are not meant to be used on that system (the host system), since they are just resources to be installed into the images. When mkosi puts them into the image, the executable bit will be set, because those files are templated into the image by `write_resource` with `executable=True`.

That being said, I don't think it's a problem if those scripts had the executable bit set on the host system, but the warning is spurious in my opinion, but maybe I'm also misunderstanding your point.

bluca:
That makes sense - lintian can't tell the difference, it sees .sh with shebang and without +x so it complains, it doesn't know it's used later. But as you said it's very cheap to fix

behrmann:
LGTM, if nobody else has objections to the executable bit being set. The Gentoo CI failures are unrelated.

PR #865, opened 2021-12-06
fedora: Fedora >= 35 don't use short key IDs anymore

DaanDeMeyer:


behrmann:
LGTM. While we're at it, we should maybe also bump the Fedora default version to 35, so we notice such things in CI and maybe also add the Fedora 37 key `ACB5EE4E831C74BB7C168D27F55AD3FB5323552A`

keszybz:
CI / ci (fedora, gpt_squashfs): Resolving sources.archlinux.org (sources.archlinux.org)... 49.12.124.107, 2a01:4f8:242:5614::2

PR #868, opened 2021-12-07
Add support for CentOS Stream releases

davide125:
Extend the existing `centos` and `centos_epel` distribution to support CentOS Stream releases (`8-stream` and `9-stream` currently).

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -6200,7 +6200,7 @@ def load_args(args: argparse.Namespace) -> MkosiArgs:
         if args.distribution == Distribution.fedora:
             args.release = "35"
         elif args.distribution in (Distribution.centos, Distribution.centos_epel):
-            args.release = "8"
+            args.release = "8-stream"
         elif args.distribution in (Distribution.rocky, Distribution.rocky_epel):
             args.release = "8"
         elif args.distribution in (Distribution.alma, Distribution.alma_epel):
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2376,7 +2376,7 @@ def install_openmandriva(args: MkosiArgs, root: Path, do_run_build_script: bool)
 
 
 def install_centos_repos_old(args: MkosiArgs, root: Path, epel_release: int) -> None:
-    # Repos for CentOS 7 and earlier
+    # Repos for CentOS Linux 7 and earlier
 
     gpgpath = Path(f"/etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-{args.release}")
     gpgurl = f"https://www.centos.org/keys/RPM-GPG-KEY-CentOS-{args.release}"
@@ -2409,7 +2409,7 @@ def install_centos_repos_old(args: MkosiArgs, root: Path, epel_release: int) -> 
 
 
 def install_centos_repos_new(args: MkosiArgs, root: Path, epel_release: int) -> None:
-    # Repos for CentOS 8 and later
+    # Repos for CentOS Linux 8 and CentOS Stream 8
 
     gpgpath = Path("/etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial")
     gpgurl = "https://www.centos.org/keys/RPM-GPG-KEY-CentOS-Official"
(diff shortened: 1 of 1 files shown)


behrmann:
Thanks for the contribution, @davide125! 

Could you run mypy on this to check it? It seems you have an undefined variable

davide125:
Thanks! Looks like it's passing now.

mkosi/__init__.py
```diff
@@ -2508,33 +2540,43 @@ def install_alma_repos(args: CommandLineArguments, root: Path, epel_release: int
...
+    epel_release, is_stream = parse_epel_release(args.release)
 
-    if old:
+    if epel_release <= 7:
         install_centos_repos_old(args, root, epel_release)
-    else:
+    elif epel_release <= 8:
         install_centos_repos_new(args, root, epel_release)
+    else:
+        install_centos_stream_repos(args, root, epel_release)
```
behrmann, line 2572: Two things here:
1. `is_older_than_centos8` is no longer used. Either use it or remove it
2. I'm not sure I fully understand the logic here, but that might be my fault, because there's now so many CentOS versions in flight, I can't keep up anymore. :) Isn't there now CentOS 8 Stream and CentOS 9 Stream? I'm not quite sure I understand how this meshes with EPEL, but wouldn't I end up in the middle branch when trying to install CentOS 8 Stream? I don't think I see `is_stream` being used.

davide125: `is_older_than_centos8` is used in a bunch of other functions, that's why I left it in place for now; happy to codemod it out though if you prefer

`epel_release` is misnamed; while it may have originally been used to denote the EPEL version, it looks like its actual meaning (even before my PR) was to be the integer version of the RHEL-like distro that's being processed. It probably should be renamed (maybe to `el_release`?), but I didn't want to do it here as it would have made the PR harder to review; again, happy to codemod it if you prefer

About what's going on with CentOS, here's the short version. There's two distros, CentOS Linux and CentOS Stream. CentOS Linux is what used to be called just CentOS, and is a straight rebuilt of RHEL; its final release (CentOS Linux 8) goes EOL on Dec 31; it is served from mirror.centos.org via mirrorlist. CentOS Stream is a new distribution that tracks the next minor release of RHEL and allows community contributions (as opposed to just being a rebuild); there's currently CentOS Stream 8 (served from mirror.centos.org via mirrorlist) and CentOS Stream 9 (served via mirror.stream.centos.org via the Fedora mirrormanager infra). So, to recap:
- `install_centos_repos_old` is used for CentOS Linux <= 7
- `install_centos_repos_new` is used for CentOS Linux 8 and CentOS Stream 8
- `install_centos_stream_repos` is used for CentOS Stream 9 (and there is no CentOS Linux 9)

EPEL just tracks the major version of the corresponding CentOS/RHEL release, so e.g. EPEL 8 will work with CentOS Linux 8, CentOS Stream 8 and RHEL 8 (with a tiny caveat that's not relevant here).

DaanDeMeyer:
I think we should go ahead and change the default CentOS release to CentOS Stream 9 as well. We should probably drop CentOS 7 support as well since it's EOL but let's leave that for another PR.

Maybe we should go further and just support CentOS Stream since CentOS 8 Linux will also be EOL in December anyway. That should allow us to simplify this logic a bit I think.

mkosi/__init__.py
DaanDeMeyer: Current logic looks fine except the is_stream variable doesn't seem used so I'd change parse_epel_release to not return a Tuple and remove the is_stream variable

davide125:
CI failures are because systemd-networkd is missing for 9-stream at the moment, it will be provided by systemd-extras once https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2021-f2a4e24f9e clears bodhi.

DaanDeMeyer:
Let's land this after systemd-extras becomes available so we can do a full CI run

behrmann:
Restarted the CI since it looks like https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2021-f2a4e24f9e has hit stable a couple of hours ago.

behrmann:
> I think we should go ahead and change the default CentOS release to CentOS Stream 9 as well. We should probably drop CentOS 7 support as well since it's EOL but let's leave that for another PR.

Thinking about it, *is* CentOS 7 actually EOL? I know CentOS 8 will be in a couple of days, but I haven't found any official notice for the end of CentOS 7.

davide125:
> Thinking about it, is CentOS 7 actually EOL? I know CentOS 8 will be in a couple of days, but I haven't found any official notice for the end of CentOS 7.

CentOS Linux 8 goes EOL on Dec 31 2021. CentOS Linux 7 goes EOL on Jun 2024.

davide125:
Ugh, looks like systemd in c9s has dropped sd-boot and bootctl, which is breaking the bootable images. Checking now if we can get that added to systemd-extras as well.

davide125:
Talked to @robert-scheck, we'll get this added to systemd-extras.

davide125:
And getting this added to systemd-extras is in turn blocked by https://bugzilla.redhat.com/show_bug.cgi?id=2031426 which is getting sorted out upstream in https://github.com/systemd/systemd/pull/21759.

robert-scheck:
As https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2021-02d864bd69 landed about 2 months ago in EPEL 8, I wonder if you already had a chance to test the systemd-boot package?

davide125:
@robert-scheck yup, the plan is to merge this one with 8-stream as the default and then make another PR to flip to 9-stream. This way we'll get both to run on the CI and ensure things are fine. AFAICT your packaging of systemd-boot works fine, at least in mkosi, but I have not tested it extensively beyond that. It looks like CentOS Stream 9 will be rebasing systemd to 250, which includes my fix, so once that happens you should be able to drop the manpages hack in your specfile.

PR #869, opened 2021-12-08
typing: Fix typing error when calling add_signer()

DaanDeMeyer:
The add_signer() method of PKCS7SignatureBuilder only supports a
subset of the key types returned by load_pem_private_key() so let's
make sure the loaded key is of one of the supported types to fix
the error.


behrmann (approved):
LGTM.

mkosi/__init__.py
```diff
@@ -3951,12 +3951,16 @@ def make_verity_sig(
     from cryptography import x509
     from cryptography.hazmat.primitives import hashes, serialization
     from cryptography.hazmat.primitives.serialization import pkcs7
+    from cryptography.hazmat.primitives.asymmetric import ec, rsa
```
behrmann, line 3954: isort wants this line one line up, I think, since `asymmetric` comes before `serialization`.

behrmann:
Merging this, since there is no need to wait for the last Gentoo build, because it's not affected by the change. All the rest is green.

PR #870, opened 2021-12-08
RFC: Remove argparse tests + pytest

DaanDeMeyer:
I've never seen the argparse tests catch errors and I find them extremely hard to understand. They're also written in a different style than what we tend to write in mkosi and systemd. Unless anyone thinks these tests are useful, I'd go for removing them to reduce the overhead in maintenance and for contributors when adding new options since for every contributor there's a dance to figure out how the tests work.

The second commit removes the usage of pytest since the remaining tests are basic enough that we should be fine with using the standard library testing framework.


behrmann:
I agree in principle, they are terribly hard to understand and they've probably never caught a bug and we have to go through the motions of adding some lines whenever we add a new CLI option. That being said, I think they are more meant as a spec for the huge homegrown argument parsing framework we have. At some point we should have a look whether this can be streamlined and for that they should be kept around, I think (even though they would probably need to be touched in the same move).

README.md
```diff
@@ -32,11 +32,9 @@ python3 -m pip install --user --no-use-pep517 --editable .
 This will install mkosi in editable mode to `~/.local/bin/mkosi`, allowing you
 to use your own changes right away.
 
-For development you also need [mypy](https://github.com/python/mypy), for type
-checking, [pytest](https://github.com/pytest-dev/pytest), to run tests, and
-[black](https://github.com/psf/black), for code formatting. We check tests,
-typing and code formatting in CI (see `.github/workflows`), but you can run the
-tests locally as well.
+For development you also need [mypy](https://github.com/python/mypy) for type
+checking, for code formatting. We check tests, typing and code formatting in CI
```
dnicolodi, line 36: The link to black went missing here.

DaanDeMeyer: We don't enforce it anymore so I removed it as well

DaanDeMeyer:
> I agree in principle, they are terribly hard to understand and they've probably never caught a bug and we have to go through the motions of adding some lines whenever we add a new CLI option. That being said, I think they are more meant as a spec for the huge homegrown argument parsing framework we have. At some point we should have a look whether this can be streamlined and for that they should be kept around, I think (even though they would probably need to be touched in the same move).

If this was a large amount of code, I'd agree that doing it the proper way, (rewrite arg parsing without touching tests, verify all tests still pass, rewrite tests) would be better. However, I have an intern starting in early January who'll be working on improving mkosi testing and adding more tests in general and I think having them work with the existing tests is going to lead to a lot of confusion (hell I don't even understand those tests myself). I'm thinking i'll be much more productive to start from scratch instead of trying to incrementally move from the current stuff to something else.

I don't really mind keeping these tests intact for now but I would advocate against making these tests work with any changes we make to the argument parsing as I don't think the value add will be worth the extra work.

behrmann:
That's great that somebody will be adding more tests! :) This also changes the situation somewhat. Let's just see what the changes will be and we can still decide once the old tests get in the way. I'm looking forward to seeing more tests.

DaanDeMeyer:
Let's close for now then

PR #871, opened 2021-12-08
Various improvements

DaanDeMeyer:


mkosi/__init__.py
```diff
@@ -65,7 +65,7 @@
 
 from .backend import (
     ARG_DEBUG,
-    CommandLineArguments,
+    MkosiArgs,
```
behrmann, line 68: isort complains about this not being above `MkosiException`

mkosi/manifest.py
```diff
@@ -9,7 +9,7 @@
 from typing import IO, Any, Dict, List, Optional, cast
 
 from .backend import (
-    CommandLineArguments,
+    MkosiArgs,
```
behrmann, line 12: same here

behrmann (approved):
LGTM. 

The case of `args.verity_size` and `args.verity_sig_size` is a bit curious, they got lost in the refactor towards `PartitionTable`, but it was never really used in the first place, because the code paths using them would have only done so, when they're not `None` and they always only were that and there was no knob to change that.

behrmann:
I'm waiting for somebody else to have a look at this. This all looks pretty straight forward, but it's a lot, so I'd like somebody else to give it a thumbs up as well.

DaanDeMeyer:
Let's merge, Aside from the rename of CommandLineArguments to MkosiArgs, the changes are pretty small.

PR #872, opened 2021-12-08
CI: Remove ext4, xfs and gpt_squashfs jobs

DaanDeMeyer:
We're running too many CI jobs, let's disable some that are similar
to others to reduce the strain on the CI.


behrmann:
LGTM.

keszybz:
Looks reasonable. We can always add some back if it turns out we're not testing enough of the code.

PR #877, opened 2021-12-23
Docs improvements

DaanDeMeyer:


mkosi.md
```diff
@@ -1612,23 +1612,21 @@ variables:
 Create and run a raw *GPT* image with *ext4*, as `image.raw`:
 
 ```bash
-# mkosi
-# systemd-nspawn -b -i image.raw
+# mkosi -b -i boot
```
bluca, line 1615: For docs and scripts I think it would be better and clearer to use the `--long-form` for switches, otherwise if you are new it's very opaque what this does

behrmann: I agree with @bluca, also it might be nice to keep the other stuff around to show what's happening under the hood (without having to look at the code).

bluca (approved):

Issue #861, opened 2021-11-25
openSUSE can't be built with v11


evverx:
Trying to bump the mkosi action from v10 to v11 in https://github.com/systemd/systemd/pull/21505, the systemd CI discovered that openSUSE can't be built:
```
2021-11-24T15:41:28.8020117Z ‣  (Detaching /dev/loop4)
2021-11-24T15:41:28.8020511Z Traceback (most recent call last):
2021-11-24T15:41:28.8021002Z   File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
2021-11-24T15:41:28.8021699Z     return _run_code(code, main_globals, None,
2021-11-24T15:41:28.8022373Z   File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
2021-11-24T15:41:28.8022844Z     exec(code, run_globals)
2021-11-24T15:41:28.8024160Z   File "/usr/local/lib/python3.8/dist-packages/mkosi/__main__.py", line 32, in <module>
2021-11-24T15:41:28.8024941Z     main()
2021-11-24T15:41:28.8025982Z   File "/usr/local/lib/python3.8/dist-packages/mkosi/__main__.py", line 26, in main
2021-11-24T15:41:28.8026496Z     run_verb(a)
2021-11-24T15:41:28.8027522Z   File "/usr/local/lib/python3.8/dist-packages/mkosi/__init__.py", line 7809, in run_verb
2021-11-24T15:41:28.8028389Z     manifest = build_stuff(args)
2021-11-24T15:41:28.8029310Z   File "/usr/local/lib/python3.8/dist-packages/mkosi/__init__.py", line 7219, in build_stuff
2021-11-24T15:41:28.8029910Z     image = build_image(args, root, do_run_build_script=True)
2021-11-24T15:41:28.8030654Z   File "/usr/local/lib/python3.8/dist-packages/mkosi/__init__.py", line 6965, in build_image
2021-11-24T15:41:28.8031297Z     install_distribution(args, root, do_run_build_script, cached_tree)
2021-11-24T15:41:28.8032128Z   File "/usr/local/lib/python3.8/dist-packages/mkosi/__init__.py", line 3111, in install_distribution
2021-11-24T15:41:28.8033948Z     install[args.distribution](args, root, do_run_build_script)
2021-11-24T15:41:28.8034735Z   File "/usr/lib/python3.8/contextlib.py", line 75, in inner
[...]

PR #878, opened 2022-01-03
action: Fix libsolv in combination with rpm 4.17.0

DaanDeMeyer:
This applies the patch from https://github.com/systemd/mkosi/issues/861#issuecomment-1003624115
until the issue is fixed upstream and the fix is released.


behrmann (approved):

DaanDeMeyer:
This doesn't apply to gentoo so let's merge.

Issue #879, opened 2022-01-03
Building a Fedora image fails on CentOS Stream 8


DaanDeMeyer:
/etc/os-release:

```
NAME="CentOS Stream"
VERSION="8"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="8"
PLATFORM_ID="platform:el8"
PRETTY_NAME="CentOS Stream 8"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:8"
HOME_URL="https://centos.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux 8"
REDHAT_SUPPORT_PRODUCT_VERSION="CentOS Stream"
```

```
git clone git@github.com:systemd/mkosi
sudo mkosi/bin/mkosi -d fedora
```

First failure:

```
~/mkosi (main)> sudo bin/mkosi -d fedora
‣ Detaching namespace
‣ Setting up temporary workspace.
‣ Temporary workspace set up in /var/tmp/mkosi-8qgm5g66
‣ Running second (final) stage…
‣  Creating image with partition table…
label: gpt
grain: 4096
first-lba: 40
size=6291456, type=4f68bce3-e8cd-4db1-96e7-fbcaf984b709, attrs=, name="Root Partition"
Disk /home/daandemeyer/mkosi/mkosi.output/.mkosi-cfyl3mx0: 3 GiB, 3221266432 bytes, 6291536 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes

>>> Script header accepted.
>>> line 2: unsupported command

New situation:
Leaving.

Traceback (most recent call last):
  File "/usr/local/fbcode/platform009/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/fbcode/platform009/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/daandemeyer/mkosi/mkosi/__main__.py", line 32, in <module>
    main()
  File "/home/daandemeyer/mkosi/mkosi/__main__.py", line 26, in main
    run_verb(a)
  File "/home/daandemeyer/mkosi/mkosi/__init__.py", line 7812, in run_verb
[...]

PR #880, opened 2022-01-03
Support older versions of sfdisk without "grain" support

DaanDeMeyer:
If "grain" is not available, let's fall back to the default grain
size of 1 MiB. To check if "grain" is available, let's run sfdisk
against /dev/full and try to configure the "grain" which will fail
if "grain" is not supported.

### DIFF
--- a/mkosi/backend.py
+++ b/mkosi/backend.py
@@ -6,6 +6,7 @@ import argparse
 import contextlib
 import dataclasses
 import enum
+import functools
 import math
 import os
 import resource
@@ -260,6 +261,18 @@ class Partition:
         return ', '.join(filter(None, desc))
 
 
+@functools.lru_cache(maxsize=None)
+def sfdisk_grain_is_supported() -> bool:
+    cmd: List[PathString] = ["sfdisk", "--no-reread", "--no-act", "--quiet", "/dev/full"]
+
+    try:
+        run(cmd, text=True, input='\n'.join(["label: gpt", "grain: 4096", "quit"]), stderr=subprocess.DEVNULL)
+    except subprocess.CalledProcessError:
+        return False
+
+    return True
+
+
 @dataclasses.dataclass
 class PartitionTable:
     partitions: Dict[PartitionIdentifier, Partition] = dataclasses.field(default_factory=dict)
@@ -268,6 +281,10 @@ class PartitionTable:
     first_lba: Optional[int] = None
 
     grain: int = 4096
+
+    def __post_init__(self) -> None:
+        if not sfdisk_grain_is_supported():
+            self.grain = 1024 ** 2 # Use sfdisk default grain size of 1 MiB.
 
     def first_partition_offset(self, max_partitions: int = 128) -> int:
         if self.first_lba is not None:
@@ -323,7 +340,7 @@ class PartitionTable:
 
     def sfdisk_spec(self) -> str:
         table = ["label: gpt",
-                 f"grain: {self.grain}",
+                 f"grain: {self.grain}" if sfdisk_grain_is_supported() else '\n',
                  f"first-lba: {self.first_partition_offset() // self.sector_size}",
                  *(p.sfdisk_spec() for p in self.partitions.values())]
         return '\n'.join(table)


DaanDeMeyer:
Fixes #879

mkosi/backend.py
```diff
@@ -323,7 +338,7 @@ def partition_path(self, ident: PartitionIdentifier, loopdev: Path) -> Optional[
 
     def sfdisk_spec(self) -> str:
         table = ["label: gpt",
-                 f"grain: {self.grain}",
+                 f"grain: {self.grain}" if sfdisk_grain_is_supported() else '\n',
```
behrmann, line 341: Having two calls to `sfdisk_grain_is_supported` is a bit unfortunate, but I guess it's still a better option than the alternative of checking whether the grain is `1024**2`, what do you think?

mkosi/backend.py
```diff
@@ -269,6 +280,10 @@ class PartitionTable:
 
     grain: int = 4096
 
+    def __post_init__(self) -> None:
+        if not sfdisk_grain_is_supported():
+            self.grain = 1048576 # Use sfdisk default grain size of 1 MiB.
```
behrmann, line 285: Maybe a matter of preference in terms of readability
```suggestion
            self.grain = 1024 ** 2 # Use sfdisk default grain size of 1 MiB.
```
I think Python would constant fold this.

mkosi/backend.py
```diff
@@ -260,6 +260,17 @@ def sfdisk_spec(self) -> str:
         return ', '.join(filter(None, desc))
 
 
+def sfdisk_grain_is_supported() -> bool:
```
behrmann, line 263: Since sfdisk is already coupled to `PartitionTable` by name, this could live as a `@staticmethod` (and with a `_` in front of its name) within `PartitionTable`, since there probably won't be other users. It would also serve as a reminder once we want to remove this.

behrmann (commented):
LGTM. Some minor nibs.

behrmann:
@keszybz I don't remember the reason, what is `grain` for here exactly? The man page "strongly recommends" to use the default.

keszybz:
Is this enough to support the fdisk version in centos 8? Based on the comment in #879 I assumed that there's a further problem with size calculation.

keszybz:
> @keszybz I don't remember the reason, what is `grain` for here exactly? The man page "strongly recommends" to use the default.

grain was used because w/o grain we get a noticeable bump is image size for small images. E.g. a small sysext image could be ~1MB, and with the default grain of 1MB, and two signature partitions, there'd be ~3 MB of overhead.

DaanDeMeyer:
> Is this enough to support the fdisk version in centos 8? Based on the comment in #879 I assumed that there's a further problem with size calculation.

Size calculation works if we make sure to default grain to 1 MiB if grain isn't supported which we do in this commit.

mkosi/backend.py
```diff
@@ -260,6 +260,17 @@ def sfdisk_spec(self) -> str:
...
+    cmd: List[PathString] = ["sfdisk", "--no-reread", "--no-act", "--quiet", "/dev/full"]
+
+    try:
+        run(cmd, input='\n'.join(["label: gpt", "grain: 4096", "quit"]).encode("utf-8"), stderr=subprocess.DEVNULL)
```
keszybz, line 267: Please just use `text=True` instead of `.encode()`.

keszybz:
Looks reasonable.

mkosi/backend.py
mrc0mmand: Maybe caching `sfdisk_grain_is_supported`'s result would help here? Basically having a "static" (whatever that translates to python) tristate variable, which would initialize itself on the first run of the function and would return its result on all subsequent calls without the need to call `sfdisk`.

behrmann: This could be done most simply by adding another attribute `_sfdisk_has_grain: bool = False` to the dataclass and setting it in `__post_init__`.

DaanDeMeyer: I went with `functools.lru_cache()` on `sfdisk_grain_is_supported()` to memoize the result.

mkosi/backend.py
DaanDeMeyer: I don't think this matters too much, sfdisk doesn't really have any tight coupling to PartitionTable so let's keep the function outside as it is.

PR #882, opened 2022-01-12
Use unittest instead of pytest for new tests

DaanDeMeyer:
Hacking on mkosi on older distributions becomes a lot easier when you
don't have any external library dependencies. You can simply get an
updated version of Python without the need to maintain external libraries
for that Python version.

This benefit seems substantial enough to me to justify using unittest
over pytest as it's part of the standard library. Arguably the biggest
loss is that we can't use "assert" anymore but have to use the assert
methods of unittest but this shouldn't be a huge deal.

I didn't touch the config tests since the plan is to remove them once
we add some simpler argparsing tests in the near future.

We can still use pytest to run all the tests as it supports running
unittest based tests as well.

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1,24 +1,25 @@
 # SPDX-License-Identifier: LGPL-2.1+
 
 import os
+import unittest
 
 from mkosi.backend import PackageType, Distribution, set_umask
 
+class BackendTests(unittest.TestCase):
+    def test_distribution(self):
+        self.assertEqual(Distribution.fedora.package_type, PackageType.rpm)
+        self.assertIs(Distribution.fedora, Distribution.fedora)
+        self.assertIsNot(Distribution.fedora,  Distribution.debian)
+        self.assertEqual(str(Distribution.photon), "photon")
 
-def test_distribution():
-    assert Distribution.fedora.package_type == PackageType.rpm
-    assert Distribution.fedora is Distribution.fedora
-    assert Distribution.fedora is not Distribution.debian
-    assert str(Distribution.photon) == "photon"
 
+    def test_set_umask(self):
+        with set_umask(0o767):
+            tmp1 = os.umask(0o777)
+            with set_umask(0o757):
+                tmp2 = os.umask(0o727)
+            tmp3 = os.umask(0o727)
 
-def test_set_umask():
-    with set_umask(0o767):
-        tmp1 = os.umask(0o777)
-        with set_umask(0o757):
-            tmp2 = os.umask(0o727)
-        tmp3 = os.umask(0o727)
-
-    assert tmp1 == 0o767
-    assert tmp2 == 0o757
-    assert tmp3 == 0o777
+        self.assertEqual(tmp1, 0o767)
+        self.assertEqual(tmp2, 0o757)
+        self.assertEqual(tmp3, 0o777)


behrmann:
LGTM, but just to question the train of thought: I'd guess pytest is packaged as a distro package everywhere and we don't use but the basic features (which is why I don't really see an issue to use unittest), so even old versions should be fine. Is using the distro version of pytest an option on those machines?

DaanDeMeyer:
> LGTM, but just to question the train of thought: I'd guess pytest is packaged as a distro package everywhere and we don't use but the basic features (which is why I don't really see an issue to use unittest), so even old versions should be fine. Is using the distro version of pytest an option on those machines?

The problem is that we can't use the distro's python because it's too old. It's not too hard to get a newer python but it's painful to add packages to it.

keszybz:
I don't think this is a good idea. pytest is so much better that it's worth the additional pain of installing it separately.

unittest has the pointless javaesque syntax (and it's not just syntax overhead, but the methods are attached to wrong objects, i.e. it's a result of the forced "objectification" in java, and it just doesn't make any sense).

I think we need to rework the tests that we have: right now they don't really test much. And if we are to make the tests smarter, we really don't want to use an inferior testing framework.

DaanDeMeyer:
> unittest has the pointless javaesque syntax (and it's not just syntax overhead, but the methods are attached to wrong objects, i.e. it's a result of the forced "objectification" in java, and it just doesn't make any sense).

I'm not a fan of it either, but I looked into both unittest and pytest for integration testing and found unittest a lot more flexible than pytest. With unittest, we can write a subclass of unittest.Testcase that can build, boot and shut down an image. We can configure subclasses of this helper class using https://www.python.org/dev/peps/pep-0487/ which gives us a lot of flexibility when writing tests (e.g. optionally reboot after every test, use ephemeral images, use nspawn or qemu, ...).

With pytest, we'd have to use fixtures but it doesn't even seem possible to pass arguments to fixtures in a nice way so there seems to be no way to configure a fixture differently for a test. Running into limitations for stuff like this that should be simple does not make me confident about using pytest for more complicated tests.

Anecdotally, @company 90% of python tests are written unittest style and there don't seem to be too many issues with the approach.

The big advantage of pytest fixtures is that they compose better than inheritance, but we don't really need that for a project of mkosi's size.. I don't see us having more than one test helper class that every integration test will inherit from. Aside from that, I don't think it provides any major advantages over unittest which makes me reluctant to keep using it.

> unittest has the pointless javaesque syntax (and it's not just syntax overhead, but the methods are attached to wrong objects, i.e. it's a result of the forced "objectification" in java, and it just doesn't make any sense).

This is slightly vague, unittest seems pretty simple to me, you have an object, every method prefixed with test_ is a test. If you need setup/teardown, you have setUp(), tearDown() which are invoked before each test and setUpClass() and tearDownClass() which are invoked once per class. Sure there's a bit of boilerplate and it doesn't compose but I don't find it too hard to understand.

DaanDeMeyer:
To unblock @gsegatti's PRs I'd like to reach a decision on this PR this week if possible. If no more big issues with the approach are raised, I'll go ahead and merge this one on Sunday.

cc @keszybz in case you haven't seen my reply above yet. Just trying to make some progress on this PR.

DaanDeMeyer:
That sounds reasonable. Let's delay the decision until we have an example integration test ready.

DaanDeMeyer:
After thinking this over again and considering all the feedback, let's stick to pytest for the unit tests. I'm still a bit hesitant about whether it'll do the job for integration tests but even then, there's nothing preventing us from mixing unittest test cases inbetween the pytest ones. The pytest runner will discover and run both styles of unit tests seamlessly.

PR #883, opened 2022-01-13
Test init backend

gsegatti:
Adding unit tests to methods of backend.py and __init__.py. I'm currently an intern @ Meta under @DaanDeMeyer's guidance.

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1,8 +1,10 @@
 # SPDX-License-Identifier: LGPL-2.1+
 
 import os
+from pathlib import Path
 
-from mkosi.backend import PackageType, Distribution, set_umask
+import mkosi.backend as backend
+from mkosi.backend import Distribution, PackageType, set_umask
 
 
 def test_distribution():
@@ -22,3 +24,89 @@ def test_set_umask():
     assert tmp1 == 0o767
     assert tmp2 == 0o757
     assert tmp3 == 0o777
+
+
+def test_workspace():
+    assert backend.workspace(Path("/home/folder/mkosi/mkosi")) == Path("/home/folder/mkosi")
+    assert backend.workspace(Path("/home/../home/folder/mkosi/mkosi")) == Path("/home/../home/folder/mkosi")
+    assert backend.workspace(Path("/")) == Path("/")
+    assert backend.workspace(Path()) == Path()
+
+
+def test_footer_size():
+    table = backend.PartitionTable()
+    assert table.footer_size() == 16896
+    assert table.footer_size(max_partitions=64) == 8704
+    assert table.footer_size(max_partitions=1) == 1024
+    assert table.footer_size(max_partitions=0) == 512
+
+def test_first_partition_offset():
+    table = backend.PartitionTable()
+    table.grain = 4096
+
+    # Grain = 4096, first_lba = None.
+    # 20480 = (512 bytes sector size + 16896 bytes footer) rounded up to multiple of 4096 grain.
+    assert table.first_partition_offset() == 20480
+
+    # Grain = 4096, first_lba not None.
+    # 32768 = first_lba of 64 * 512 bytes of sector_size.
+    table.first_lba = 64
+    assert table.first_partition_offset() == 32768
+
+    # 0 since first_lba = 0 will be multiplied by some value of sector_size.
+    table.first_lba = 0
+    assert table.first_partition_offset() == 0
+
+    # Grain = 1024 ** 2, first_lba = None.
+    # 1048576 = (512 bytes sector size + 16896 bytes footer) rounded up to multiple of 1024 ** 2 grain.
+    table.grain = 1024 ** 2
+    table.first_lba = None
+    assert table.first_partition_offset() == 1048576
+
+    # Grain = 1024 ** 2, first_lba not None.
+    # 65536 = first_lba of 128 * 512 bytes of sector_size.
+    table.first_lba = 128
+    assert table.first_partition_offset() == 65536
+
+    # Grain = 1024 ** 2, first_lba not None.
+    # 131072 = first_lba of 256 * 512 bytes of sector_size.
+    table.first_lba = 256
+    assert table.first_partition_offset() == 131072
+
+
+def test_last_partition_offset():
+    table = backend.PartitionTable()
+    table.grain = 4096
+
+    table.last_partition_sector = 32
+    assert table.last_partition_offset() == 16384
+
+    table.last_partition_sector = 16
+    assert table.last_partition_offset() == 8192
+
+    table.last_partition_sector = 1
+    assert table.last_partition_offset() == 4096
+
+    table.last_partition_sector = 0
+    table.first_lba = 64
+    assert table.last_partition_offset() == 32768
+
+    table.first_lba = 0
+    assert table.last_partition_offset() == 0
+
+
+def test_disk_size():
+    table = backend.PartitionTable()
+    table.grain = 4096
+    table.last_partition_sector = 0
+    table.first_lba = 64
+    # When disk_size() cascade upwards all the way to first_partition_offset, if clause.
+    assert table.disk_size() == 53248
+
+    # When disk_size() cascade upwards all the way to first_partition_offset, else clause.
+    table.first_lba = None
+    assert table.disk_size() == 40960
+
+    # When disk_size() cascade upwards to last_partition_offset, if clause.
+    table.last_partition_sector = 32
+    assert table.disk_size() == 36864
--- a/tests/test_init.py
+++ b/tests/test_init.py
@@ -1,0 +1,55 @@
+# SPDX-License-Identifier: LGPL-2.1+
+
+import filecmp
+from pathlib import Path
+import pytest
+
+import mkosi
+
+
+def test_fedora_release_cmp():
+    assert mkosi.fedora_release_cmp("rawhide", "rawhide") == 0
+    assert mkosi.fedora_release_cmp("32", "32") == 0
+    assert mkosi.fedora_release_cmp("33", "32") > 0
+    assert mkosi.fedora_release_cmp("30", "31") < 0
+    assert mkosi.fedora_release_cmp("-1", "-2") > 0
+    assert mkosi.fedora_release_cmp("1", "-2") > 0
+    with pytest.raises(ValueError):
+        mkosi.fedora_release_cmp("literal", "rawhide")
+
+
+def test_strip_suffixes():
+    assert mkosi.strip_suffixes(Path("home/test.zstd")) == Path("home/test")
+    assert mkosi.strip_suffixes(Path("home/test.xz")) == Path("home/test")
+    assert mkosi.strip_suffixes(Path("home/test.raw")) == Path("home/test")
+    assert mkosi.strip_suffixes(Path("home/test.tar")) == Path("home/test")
+    assert mkosi.strip_suffixes(Path("home/test.cpio")) == Path("home/test")
+    assert mkosi.strip_suffixes(Path("home/test.qcow2")) == Path("home/test")
+    assert mkosi.strip_suffixes(Path("home.xz/test.xz")) == Path("home.xz/test")
+    assert mkosi.strip_suffixes(Path("home.xz/test")) == Path("home.xz/test")
+    assert mkosi.strip_suffixes(Path("home.xz/test.txt")) == Path("home.xz/test.txt")
+
+def test_copy_file(tmpdir):
+        dir_path = Path(tmpdir)
+        file_1 = Path(dir_path) / "file_1.txt"
+        file_2 = Path(dir_path) / "file_2.txt"
+        file_1.touch()
+        file_2.touch()
+
+        # Copying two empty files.
+        mkosi.copy_file(file_1, file_2)
+        assert filecmp.cmp(file_1, file_2)
+
+        # Copying content from one file.
+        file_1.write_text("Testing copying content from this file to file_2.")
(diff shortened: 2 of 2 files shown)


behrmann:
Welcome! :)

I see this is using the unittest style, I think that's something that still needs to be deiced in #882.

As a comment for later: The commits need squashing.

tests/test_init_and_backend.py
```diff
@@ -0,0 +1,69 @@
...
+        self.assertEqual(Path("home/test"), init.strip_suffixes(Path("home/test.tar")))
+        self.assertEqual(Path("home/test"), init.strip_suffixes(Path("home/test.cpio")))
+        self.assertEqual(Path("home/test"), init.strip_suffixes(Path("home/test.qcow2")))
+        self.assertEqual(Path("home.xz/test"), init.strip_suffixes(Path("home.xz/test.xz")))
+        self.assertEqual(Path("home.xz/test"), init.strip_suffixes(Path("home.xz/test")))
+
+
+    def test_copy_file(self):
+        temp1 = tempfile.NamedTemporaryFile(delete=True)
+        temp2 = tempfile.NamedTemporaryFile(delete=True)
```
behrmann, line 48: I think this should better be used in a `with` block. `delete=True` can be left out, since it is the default.

tests/test_init_and_backend.py
```diff
@@ -0,0 +1,69 @@
+import os
+from os import error
+import unittest
+from pathlib import Path
+import sys
+import tempfile
+import shutil
+import filecmp
```
behrmann, line 8: You will probably need `isort` here.

tests/test_init_and_backend.py
```diff
@@ -0,0 +1,69 @@
...
+
+sys.path.insert(0, os.getcwd() +'/../')
+import mkosi.backend as backend
+import mkosi.__init__ as init
```
behrmann, line 12: importing just `mkosi` import from `__init__`.

tests/test_init_and_backend.py
```diff
@@ -0,0 +1,69 @@
+import os
+from os import error
+import unittest
+from pathlib import Path
+import sys
+import tempfile
+import shutil
+import filecmp
+
+sys.path.insert(0, os.getcwd() +'/../')
```
behrmann, line 10: Are the path shenanigans strictly necessary? When you run the test module from the top-level of the repo `mkosi` should be in the module search path.

Manually changing the search path tends to be brittle.

DaanDeMeyer (changes requested):
Great start! I left some comments. The first two tests look great but I think we should use a slightly different approach in the third test.

tests/test_init_and_backend.py
```diff
@@ -0,0 +1,69 @@
+import os
```
bluca, line 1: SPDX license comment

DaanDeMeyer: To add a little more context, we require every file to start with an appropriate license comment for legal purposes. Instead of putting the entire license in each file, we put the SPDX license identifier in the file instead which can be used to look up the full license on https://spdx.org/licenses/.

You can simply copy the license comment from another file to this file as well to fix this.

tests/test_init_and_backend.py
```diff
@@ -0,0 +1,71 @@
...
+        self.assertEqual(mkosi.strip_suffixes(Path("home.xz/test.xz")), Path("home.xz/test"))
+        self.assertEqual(mkosi.strip_suffixes(Path("home.xz/test")), Path("home.xz/test"))
+        self.assertEqual(mkosi.strip_suffixes(Path("home.xz/test.txt")), Path("home.xz/test.txt"))
+
+    def test_copy_file(self):
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            dir_path = Path(tmp_dir)
+
+            file_1 = Path(dir_path.joinpath("file_1.txt")).touch()
+            file_2 = Path(dir_path.joinpath("file_2.txt")).touch()
```
DaanDeMeyer, line 46: There's quite a bit of duplication in this test. How about the following:

```
file_1 = Path(dir_path) / "file_1.txt"
file_2 = Path(dir_path) / "file_2.txt"

file_1.touch()
file_2.touch()

# Use file_1 and file_2 directory in the rest of the test
```

tests/test_init_and_backend.py
```diff
@@ -0,0 +1,70 @@
...
+
+            # Copying two empty files
+            mkosi.copy_file(dir_path.joinpath("file_1.txt"), dir_path.joinpath("file_2.txt"))
+            self.assertTrue(filecmp.cmp(dir_path.joinpath("file_1.txt"), dir_path.joinpath("file_2.txt"), shallow=True))
+
+            # Copying content from one file
+            with open(dir_path.joinpath("file_1.txt"), 'w') as file_1:
+                file_1.write("Testing copying content from this file to file_2.")
```
DaanDeMeyer, line 54: With the above change, this then becomes:

```
file_1.write_text("Testing copying content from this file to file_2.")
```

tests/test_init_and_backend.py
```diff
@@ -0,0 +1,70 @@
...
+            mkosi.copy_file(dir_path.joinpath("file_1.txt"), dir_path.joinpath("file_2.txt"))
+            self.assertTrue(filecmp.cmp(dir_path.joinpath("file_1.txt"), dir_path.joinpath("file_2.txt"), shallow=True))
+
+            # Giving a non existing path/file
+            with self.assertRaises(error):
+                mkosi.copy_file('nullFilePath', dir_path.joinpath("file_1.txt"))
+
+            # Copying when there's already content in both files
+            with open(dir_path.joinpath("file_2.txt"), 'w') as file_2:
+                file_2.write("Testing copying content from file_1 to file_2, with previous data.")
```
DaanDeMeyer, line 64: Same comment for these two lines

tests/test_init_and_backend.py
```diff
@@ -0,0 +1,70 @@
...
+    def test_copy_file(self):
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            dir_path = Path(tmp_dir)
+
+            file_1 = Path(dir_path.joinpath("file_1.txt")).touch()
+            file_2 = Path(dir_path.joinpath("file_2.txt")).touch()
+
+            # Copying two empty files
+            mkosi.copy_file(dir_path.joinpath("file_1.txt"), dir_path.joinpath("file_2.txt"))
+            self.assertTrue(filecmp.cmp(dir_path.joinpath("file_1.txt"), dir_path.joinpath("file_2.txt"), shallow=True))
```
DaanDeMeyer, line 50: Why pass `shallow=True` here and in the other assertions? If we do that, the actual contents of the files aren't checked. We want to make sure the contents of the files are the same no?

tests/test_init_and_backend.py
```diff
@@ -0,0 +1,70 @@
...
+
+            # Copying content from one file
+            with open(dir_path.joinpath("file_1.txt"), 'w') as file_1:
+                file_1.write("Testing copying content from this file to file_2.")
+            mkosi.copy_file(dir_path.joinpath("file_1.txt"), dir_path.joinpath("file_2.txt"))
+            self.assertTrue(filecmp.cmp(dir_path.joinpath("file_1.txt"), dir_path.joinpath("file_2.txt"), shallow=True))
+
+            # Giving a non existing path/file
+            with self.assertRaises(error):
```
DaanDeMeyer, line 59: I think error here can be replaced with `OSError`.

tests/test_init_and_backend.py
```diff
@@ -0,0 +1,70 @@
...
+
+            # Copying when there's already content in both files
+            with open(dir_path.joinpath("file_2.txt"), 'w') as file_2:
+                file_2.write("Testing copying content from file_1 to file_2, with previous data.")
+            mkosi.copy_file(dir_path.joinpath("file_1.txt"), dir_path.joinpath("file_2.txt"))
+            self.assertTrue(filecmp.cmp(dir_path.joinpath("file_1.txt"), dir_path.joinpath("file_2.txt"), shallow=True))
+
+
+if __name__ == '__main__':
+    unittest.main()
```
DaanDeMeyer, line 70: Since we always test with `python -m unittest`, the main here can be removed

DaanDeMeyer (changes requested):
We're almost there! Just a few more changes for the copy_file() test.

tests/test_init_and_backend.py
```diff
@@ -0,0 +1,70 @@
...
+            self.assertTrue(filecmp.cmp(file_1, file_2))
+
+            # Copying content from one file
+            file_1.write_text("Testing copying content from this file to file_2.")
+            mkosi.copy_file(file_1, file_2)
+            self.assertTrue(filecmp.cmp(file_1, file_2))
+
+            # Giving a non existing path/file
+            with self.assertRaises(error):
+                mkosi.copy_file('nullFilePath', file_1)
```
DaanDeMeyer, line 61: Use double quotes here for the string

tests/test_init_and_backend.py
gsegatti: Turns out it is ValueError!

DaanDeMeyer: Then I'd suggest changing it to ValueError (maybe you forgot to push?)

gsegatti:
I've tried to get the code to be matching to everyone's initial feedback. 
Furthermore, I've added unit tests for some methods of the PartitionTable class. 
This can be found in the following methods: test_first_partition_offset(), test_last_partition_offset, test_footer_size() and test_disk_size().

tests/test_init_and_backend.py
```diff
@@ -0,0 +1,131 @@
...
+        self.assertEqual(table.last_partition_offset(), 0)
+
+    def test_footer_size(self):
+        table = backend.PartitionTable()
+        self.assertEqual(table.footer_size(), 16896)
+        self.assertEqual(table.footer_size(max_partitions=64), 8704)
+        self.assertEqual(table.footer_size(max_partitions=1), 1024)
+        self.assertEqual(table.footer_size(max_partitions=0), 512)
+
+
```
DaanDeMeyer, line 116: Redundant whitespace here

tests/test_init_and_backend.py
```diff
@@ -0,0 +1,131 @@
...
+        table.last_partition_sector = 1
+        self.assertEqual(table.last_partition_offset(), 4096)
+
+        table.last_partition_sector = 0
+        table.first_lba = 64
+        self.assertEqual(table.last_partition_offset(), 32768)
+        table.first_lba = 0
+        self.assertEqual(table.last_partition_offset(), 0)
+
+    def test_footer_size(self):
```
behrmann, line 109: Since the `first_partition_offset` can depend on `footer_size` I'd move this test before the one for the former.

tests/test_init_and_backend.py
```diff
@@ -0,0 +1,139 @@
...
+    assert mkosi.strip_suffixes(Path("home/test.raw")) == Path("home/test")
+    assert mkosi.strip_suffixes(Path("home/test.tar")) == Path("home/test")
+    assert mkosi.strip_suffixes(Path("home/test.cpio")) == Path("home/test")
+    assert mkosi.strip_suffixes(Path("home/test.qcow2")) == Path("home/test")
+    assert mkosi.strip_suffixes(Path("home.xz/test.xz")) == Path("home.xz/test")
+    assert mkosi.strip_suffixes(Path("home.xz/test")) == Path("home.xz/test")
+    assert mkosi.strip_suffixes(Path("home.xz/test.txt")) == Path("home.xz/test.txt")
+
+def test_copy_file():
+    with tempfile.TemporaryDirectory() as tmp_dir:
```
DaanDeMeyer, line 40: With pytest, we can simplify this by using the `tmpdir` fixture (see https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmpdir-fixture). We don't need to create the temporary directory ourselves. Instead, pytest can do it for us.

tests/test_init_and_backend.py
```diff
@@ -0,0 +1,139 @@
...
+        dir_path = Path(tmp_dir)
+        file_1 = Path(dir_path) / "file_1.txt"
+        file_2 = Path(dir_path) / "file_2.txt"
+        file_1.touch()
+        file_2.touch()
+
+        # Copying two empty files.
+        mkosi.copy_file(file_1, file_2)
+        assert (filecmp.cmp(file_1, file_2)) == True
```
DaanDeMeyer, line 49: Asserts like these are a tautology: `assert (filecmp.cmp(file_1, file_2)) == True`

You can write this as `assert filecmp.cmp(file_1, file_2)`

DaanDeMeyer (changes requested):
Almost good!

Now that we're using pytest, let's move the tests covering functions from backend.py into test_backend.py. The remaining tests can go in a new test_init.py file.

tests/test_init_and_backend.py
```diff
@@ -0,0 +1,139 @@
...
+import mkosi.backend as backend
+import pytest
+
+
+def test_workspace():
+    assert backend.workspace(Path("/home/folder/mkosi/mkosi")) == Path("/home/folder/mkosi")
+    assert backend.workspace(Path("/home/../home/folder/mkosi/mkosi")) == Path("/home/../home/folder/mkosi")
+    assert backend.workspace(Path("/")) == Path("/")
+    assert backend.workspace(Path()) == Path()
+
```
behrmann, line 17: Small style nit: Here and between all top-level functions and classes two empty lines.

DaanDeMeyer:
LGTM!

PR #886, opened 2022-01-18
build: check for BUILD_ID/VERSION_ID before adding systemd-boot workaround

bluca:
Instead of hard-coding the debian release, which breaks booting on
testing/bookworm, check os-release to see if BUILD_ID or VERSION_ID
are actually there.


evverx:
I pointed the systemd CI to @bluca's branch in https://github.com/systemd/systemd/pull/22153. The results should be available at https://github.com/systemd/systemd/actions/runs/1710428574

behrmann (approved):
LGTM.

PR #887, opened 2022-01-18
Adding --machine-id option.

gsegatti:
Allows users to provide their own Machine ID via CLI or configuration file.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -653,6 +653,12 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
   in building or further container import stages.  This option strips
   SELinux context attributes from the resulting tar archive.
 
+`MachineID=`, `--machine-id`
+
+: Set the machine's ID to the specified value. If unused, a random ID will
+be used while building the image and the final image will be shipped without
+a machine ID.
+
 ### [Content] Section
 
 `BasePackages=`, `--base-packages`
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3150,12 +3150,13 @@ def reset_machine_id(args: MkosiArgs, root: Path, do_run_build_script: bool, for
         return
 
     with complete_step("Resetting machine ID"):
-        machine_id = root / "etc/machine-id"
-        try:
-            machine_id.unlink()
-        except FileNotFoundError:
-            pass
-        machine_id.touch()
+        if not args.machine_id_is_fixed:
+            machine_id = root / "etc/machine-id"
+            try:
+                machine_id.unlink()
+            except FileNotFoundError:
+                pass
+            machine_id.touch()
 
         dbus_machine_id = root / "var/lib/dbus/machine-id"
         try:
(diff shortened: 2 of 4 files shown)


mkosi.md
```diff
@@ -653,6 +653,11 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
   in building or further container import stages.  This option strips
   SELinux context attributes from the resulting tar archive.
 
+`MachineID=`, `--machine-id`
+
+: Set the machine's ID to the specified value. If unused, a random ID will
+be generated using uuid4().
```
DaanDeMeyer, line 659: No need to detail implementation details here. Just mention that a random machine ID will be used while building the image and that the final image will ship without a machine ID.

mkosi/__init__.py
```diff
@@ -6505,7 +6513,15 @@ def load_args(args: argparse.Namespace) -> MkosiArgs:
...
     # always the same
-    args.machine_id = uuid.uuid4().hex
+    if args.machine_id is None:
+        args.machine_id = uuid.uuid4().hex
+        args.machine_id_is_fixed = False
+    else:
+        args.machine_id_is_fixed = True
+        try:
+            uuid.UUID(hex=args.machine_id)
+        except:
```
behrmann, line 6523:
```suggestion
        except ValueError:
```
Please always be explicit about what you are catching. :)

mkosi/__init__.py
```diff
@@ -5600,6 +5607,7 @@ def create_parser() -> ArgumentParserMkosi:
         help="Turn on debugging output",
         choices=("run", "build-script", "workspace-command", "disk"),
     )
+
```
behrmann, line 5610: Unrelated whitespace change.

mkosi/__init__.py
DaanDeMeyer, line 6523: This should be `except Exception`

behrmann (commented):
LGTM.

mkosi/__init__.py
behrmann: Ah, you were a bit quicker than me. :) I think something more specific than `Exception` would better, though.

gsegatti: You're absolutely right, thanks for the suggestion!

mkosi.md
```diff
@@ -653,6 +653,12 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
   in building or further container import stages.  This option strips
   SELinux context attributes from the resulting tar archive.
 
+`MachineID=`, `--machine-id`
+
+: Set the machine's ID to the specified value. If unused, a random ID will
+be used while building the image. The final image will be shipped without
```
DaanDeMeyer, line 659:
```suggestion
be used while building the image and the final image will be shipped without
```

anitazha (commented):
Did you mean to squash the two commits? Usually we don't have separate commits for "historical" changes (e.g. minor updates to a former commit in the same PR)

PR #889, opened 2022-01-20
Initial framework for testing argument parsing

gsegatti:


tests/test_parse_load_args.py
```diff
@@ -0,0 +1,68 @@
...
+import unittest
+import uuid
+from typing import List, Optional
+
+import mkosi
+import mkosi.backend as backend
+from mkosi.backend import MkosiArgs, MkosiException, Distribution
+
+
+def load_args_wrapper(argv: Optional[List[str]] = None) -> MkosiArgs:
```
DaanDeMeyer, line 11: This is a very long name for a local function. Since we use this a lot and only in this file, we can get away with just calling it parse() I think.

tests/test_parse_load_args.py
```diff
@@ -0,0 +1,68 @@
...
+    return mkosi.load_args(mkosi.parse_args(argv)["default"])
+
+
+class TestMkosiMethods(unittest.TestCase):
+
+    def test_parse_load_verb(self):
+        self.assertIsInstance(load_args_wrapper([]), MkosiArgs)
```
DaanDeMeyer, line 18: This one is superfluous and can be removed. We already have typing to verify load_args_wrapper returns MkosiArgs.

tests/test_parse_load_args.py
```diff
@@ -0,0 +1,68 @@
...
+        with self.assertRaises(SystemExit):
+            load_args_wrapper(["help"])
+        self.assertEqual(load_args_wrapper(["genkey"]).verb, "genkey")
+        self.assertEqual(load_args_wrapper(["bump"]).verb, "bump")
+        self.assertEqual(load_args_wrapper(["serve"]).verb, "serve")
+        self.assertEqual(load_args_wrapper(["build"]).verb, "build")
+        self.assertEqual(load_args_wrapper(["shell"]).verb, "shell")
+        self.assertEqual(load_args_wrapper(["boot"]).verb, "boot")
+        self.assertEqual(load_args_wrapper(["qemu"]).verb, "qemu")
+        #self.assertEqual(load_args_wrapper(["ssh"]), self.test_MkosiArgs)
```
DaanDeMeyer, line 30: Leftover comment.

tests/test_parse_load_args.py
```diff
@@ -0,0 +1,68 @@
...
+        self.assertEqual(load_args_wrapper(["-d", "mageia"]).distribution, Distribution.mageia)
+        self.assertEqual(load_args_wrapper(["-d", "centos"]).distribution, Distribution.centos)
+        self.assertEqual(load_args_wrapper(["-d", "centos_epel"]).distribution, Distribution.centos_epel)
+        self.assertEqual(load_args_wrapper(["-d", "clear"]).distribution, Distribution.clear)
+        self.assertEqual(load_args_wrapper(["-d", "photon"]).distribution, Distribution.photon)
+        self.assertEqual(load_args_wrapper(["-d", "openmandriva"]).distribution, Distribution.openmandriva)
+        self.assertEqual(load_args_wrapper(["-d", "rocky"]).distribution, Distribution.rocky)
+        self.assertEqual(load_args_wrapper(["-d", "rocky_epel"]).distribution, Distribution.rocky_epel)
+        self.assertEqual(load_args_wrapper(["-d", "alma"]).distribution, Distribution.alma)
+        self.assertEqual(load_args_wrapper(["-d", "alma_epel"]).distribution, Distribution.alma_epel)
```
DaanDeMeyer, line 49: This isn't robust to new distribution being added. Instead, we should loop over the members of the `Distribution` enum so that any new additions are tested as well.

```
for d in Distribution:
    self.assertEqual(load_args_wrapper(["-d", d.name]).distribution, d)
```

tests/test_parse_load_args.py
```diff
@@ -0,0 +1,68 @@
...
+        with self.assertRaises((argparse.ArgumentError, SystemExit)):
+            load_args_wrapper(["-d", "invalidDistro"])
+        with self.assertRaises((argparse.ArgumentError, SystemExit)):
+            load_args_wrapper(["-d"])
+
+    def test_machine_id(self):
+        id = uuid.uuid4().hex
+        self.assertEqual(load_args_wrapper(["--machine-id", id]).machine_id, id)
```
DaanDeMeyer, line 57: Need to test that machine_id_is_fixed is set correctly here as well

DaanDeMeyer (changes requested):

tests/test_parse_load_args.py
```diff
@@ -0,0 +1,116 @@
...
+        with self.assertRaises((argparse.ArgumentError, SystemExit)):
+            parse(["-d", "invalidDistro"])
+        with self.assertRaises((argparse.ArgumentError, SystemExit)):
+            parse(["-d"])
+
+        for dist in Distribution:
+
+            with change_dir(self.tmp_dir.name):
+                with open(str(dist) + '.txt', "w") as file:
+                    file.write("[Distribution]\nDistribution=" + str(dist))
```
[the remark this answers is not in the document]
DaanDeMeyer: Also use Path(<path>).write_text() instead of opening the file and writing to it. Same goes for the rest of the file.

DaanDeMeyer (changes requested):

tests/test_parse_load_args.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def parse(argv: Optional[List[str]] = None) -> MkosiArgs:
+    return mkosi.load_args(mkosi.parse_args(argv)["default"])
+
+@contextmanager
+def change_dir(folder: str):
+    old_dir = getcwd()
+    try:
+        chdir(folder)
```
keszybz, line 21: Didn't we already have the same context manager somewhere?

The `chdir` needs to be before `try`.

tests/test_parse_load_args.py
```diff
@@ -0,0 +1,116 @@
...
+        chdir(old_dir)
+
+
+class TestMkosiMethods(unittest.TestCase):
+
+    def setUp(self):
+        self.tmp_dir = tempfile.TemporaryDirectory()
+
+    def tearDown(self) -> None:
+        self.tmp_dir.cleanup()
```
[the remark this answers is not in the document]
keszybz: Hmm, where did we end up with the pytest vs. unittest decision? pytest has a fixture for a temporary directory, so this whole setup would go away…

DaanDeMeyer:
@gsegatti Can you rebase this on the latest main branch so the gentoo check removal commit is included? That should make the unit test CI pass as it's currently failing with the gentoo portage error.

tests/test_parse_load_args.py
```diff
@@ -0,0 +1,142 @@
...
+                    with open(config, 'w') as file:
+                        file.write(f"[Distribution]\nDistribution={dist}\n")
+                        file.write(f"Release={release}\n\n")
+                        file.write("[Output]\nFormat=gpt_btrfs")
+                    with self.assertRaises(MkosiException):
+                        parse([])
+
+
+        with cd_temp_dir():
+            config = (getcwd()) / Path("mkosi.default")
```
DaanDeMeyer, line 116: getcwd() can be removed here and elsewhere I think. All paths that don't start with a '/' are relative paths which are resolved relative to the current working directory.

tests/test_parse_load_args.py
```diff
@@ -0,0 +1,142 @@
...
+            for dist in (Distribution.centos, Distribution.centos_epel):
+                for release in range (2, 8):
+                    with open(config, 'w') as file:
+                        file.write(f"[Distribution]\nDistribution={Distribution.centos}\n")
+                        file.write(f"Release={release}\n\n")
+                        file.write("Bootable=yes")
```
DaanDeMeyer, line 122: Let's use Path().write_text() here. You can look in \_\_init\_\_.py for examples on how to use write_text() to write multiline files.

tests/test_parse_load_args.py
```diff
@@ -0,0 +1,142 @@
...
+from mkosi.backend import Distribution, MkosiArgs, MkosiException
+
+
+def parse(argv: Optional[List[str]] = None) -> MkosiArgs:
+    return mkosi.load_args(mkosi.parse_args(argv)["default"])
+
+@contextmanager
+def cd_temp_dir():
+    old_dir = getcwd()
+    tmp_dir = tempfile.TemporaryDirectory()
```
DaanDeMeyer, line 20: We should use TemporaryDirectory() as a context manager here using with()

tests/test_parse_load_args.py
```diff
@@ -0,0 +1,160 @@
...
+
+    with tmp_dir:
+        chdir(tmp_dir.name)
+        try:
+            yield
+        finally:
+            chdir(old_dir)
+        tmp_dir.cleanup()
```
DaanDeMeyer, line 29: When using with, you don't have to call `cleanup` anymore so this can be removed

tests/test_parse_load_args.py
```diff
@@ -0,0 +1,160 @@
...
+
+def parse(argv: Optional[List[str]] = None) -> MkosiArgs:
+    return mkosi.load_args(mkosi.parse_args(argv)["default"])
+
+@contextmanager
+def cd_temp_dir():
+    old_dir = getcwd()
+    tmp_dir = tempfile.TemporaryDirectory()
+
+    with tmp_dir:
```
DaanDeMeyer, line 23: Let's write this as `with tempfile.TemporaryDirectory() as tmp_dir:`

tests/test_parse_load_args.py
```diff
@@ -0,0 +1,160 @@
...
+    assert parse(["build"]).verb == "build"
+    assert parse(["shell"]).verb == "shell"
+    assert parse(["boot"]).verb == "boot"
+    assert parse(["qemu"]).verb == "qemu"
+    with pytest.raises(SystemExit):
+        parse(["invalid"])
+
+def test_os_distribution():
+    for dist in Distribution:
+        assert parse(["-d", dist.name]).distribution, dist
```
DaanDeMeyer, line 48: Missing "==" here

tests/test_parse_load_args.py
```diff
@@ -0,0 +1,160 @@
...
+        parse(["--machine-id", "notValidKey"])
+    with pytest.raises((argparse.ArgumentError, SystemExit)):
+        parse(["--machine-id"])
+
+    with cd_temp_dir():
+        config = Path("mkosi.default")
+        config.write_text(f"[Output]\nMachineID={id}")
+        load_args = parse([])
+        assert load_args.machine_id == id
+        assert load_args.machine_id_is_fixed == True
```
DaanDeMeyer, line 78: No need for "== True" here

tests/test_parse_load_args.py
```diff
@@ -0,0 +1,160 @@
...
+                    textwrap.dedent(
+                        f"""
+                        [Distribution]
+                        Distribution={Distribution.centos}
+                        Release={release}
+                        Bootable=yes
+                        """
+                    )
+                )
+                with pytest.raises(MkosiException):
```
DaanDeMeyer, line 137: Need a message match here

tests/test_parse_load_args.py
```diff
@@ -0,0 +1,160 @@
...
+                        f"""
+                        [Distribution]
+                        Distribution={dist}
+                        Release={release}
+                        [Output]
+                        Format=gpt_btrfs
+                        """
+                    )
+                )
+                with pytest.raises(MkosiException):
```
DaanDeMeyer, line 120: Need a message match here

DaanDeMeyer (changes requested):

tests/test_parse_load_args.py
gsegatti: finally did it! my bad.

tests/test_parse_load_args.py
```diff
@@ -0,0 +1,160 @@
...
+                        f"""
+                        [Distribution]
+                        Distribution={dist}
+                        Release={release}
+                        [Output]
+                        Format=gpt_btrfs
+                        """
+                    )
+                )
+                with pytest.raises(MkosiException, match=".CentOS.*btrfs"):
```
DaanDeMeyer, line 120: The match here is wrong causing the CI to fail. I'm assuming you meant ".*CentOS.*btrfs"

DaanDeMeyer (changes requested):

tests/test_parse_load_args.py
gsegatti: It was not actually the *, but rather the error message being completely different from what I thought it was.
Not sure how I let this one slip.

> AssertionError: Regex pattern '.CentOS.*btrfs' does not match 'Sorry, CentOS 2 does not support unified kernel images. You must use --without-unified-kernel-images.'

Will fix it to:
image

gsegatti: Thinking about it, its best to add the * in case the order of the words are ever changed.

DaanDeMeyer:
LGTM!

Issue #709, opened 2021-04-18
Drop support for non-generated build modes


poettering:
mkosi currently has to different ways to build things:

1. regular builds — where we first set up a disk image, create an fs in it, mount it, and install linux into
2. "generated" builds — where we set up the OS in the workdir first, and finally pack it up into a disk images

Originally all builds were regular builds, but eventually more and more "generated" build schemes where added, to cover for tarballs, squashfs, "minimized build", /usr-only builds...

I'd like to propose that we drop support for regular builds entirely, and just build everything the same way: prepartion in some dir in the workdir, and then packing things up via a tool that builds a disk image from it (which can be a simple "cp" if the used image format has no tool for packing things up)

Why? To bring the combinatory explosion down a bit, and simplify things. More systematic behaviour is a good thing.

What would we lose this way? I kinda liked the fact that the file system features we provide while building the image ultimately matched exactly what we'd generate, which is great, in particular for btrfs images and such. Things might become a bit slower in some cases, as we'd always need a final copy operation.

incremental builds wozuld have to changed a bit. i.e. incremental build snapshots would have to be taken on an fs tree level, we could no longer do this on the loopback file level.

THoughts?

behrmann:
I like the idea, when I first tried to factor the distros out into their own modules, I stumbled over disentangling that from the outputs, so I think this might go hand in hand.

DaanDeMeyer:
For me, it would depend on how much slower incremental becomes. I quite like that I can build and boot into a qemu vm with the latest systemd changes in 12s or so when using btrfs with cached images and an incremental systemd build on my crappy laptop. Of course, like you say, the combinatorial explosion we have now isn't ideal. It's next to impossible to make sure everything works correctly.

What about doing the opposite? Building the generated stuff in a disk image as well instead of a separate workdir and just packing the disk image into a tar, squashfs, w.e. at the end?

poettering:
> What about doing the opposite? Building the generated stuff in a disk image as well instead of a separate workdir and just packing the disk image contents into a tar, squashfs, w.e. at the end?

Hmm, interesting idea. Also has the nice effect of putting a natural quota on things, which I think is a good idea.

Maybe this is indeed the way to go. i.e. always start out with a loopback image, consider it our "storage pool" that might or might not end up being the image in the end.

So what about this approach: instead of starting it out with fully partitioned block device which we then only use partly and then later on patch, do it like this:

[...]

PR #892, opened 2022-01-24
Add option to make `systemd-nspawn` keep unit scope

### LINKED ISSUE #248: Allow building inside a container
It would be nice if mkosi could be launched inside a container to build an image.

rezib:
Hello mkosi maintainers,

I propose this PR so mkosi could run inside a systemd-nspawn container with nested systemd-nspawn in the build steps.

I realized that `--register=no` is already hard-coded but I wasn't able to figure out why. In this situation, I though you could consider `--keep-unit` to be hard-coded as well. What do you thing about this?

Note this partially fixes #248 for systemd-nspawn containers, as far as I tested.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -1087,6 +1087,13 @@ a machine ID.
 : When used with the `qemu` verb, this options sets `qemu`'s `-m`
   argument which controls the amount of guest's RAM. Defaults to `1G`.
 
+`NspawnKeepUnit=`, `--nspawn-keep-unit`
+
+: When used, this option instructs underlying calls of systemd-nspawn to
+  use the current unit scope, instead of creating a dedicated transcient
+  scope unit for the containers. This option should be used when mkosi is
+  run by a service unit.
+
 `Netdev=`, `--netdev`
 
 : When used with the boot or qemu verbs, this option creates a virtual
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -5590,6 +5590,11 @@ def create_parser() -> ArgumentParserMkosi:
     group.add_argument("--qemu-smp", help="Configure guest's SMP settings", metavar="SMP", default="2")
     group.add_argument("--qemu-mem", help="Configure guest's RAM size", metavar="MEM", default="1G")
     group.add_argument(
+        "--nspawn-keep-unit",
+        action=BooleanAction,
+        help="If specified, underlying systemd-nspawn containers use the ressources of the current unit."
+    )
+    group.add_argument(
         "--network-veth",
         dest="netdev",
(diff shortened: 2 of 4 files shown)


DaanDeMeyer:
I assume you're using the directory, btrfs subvolume or tar output? Only those output formats have a chance of working in a nested container as anything requiring loop devices will fail since the kernel does not virtualize loop devices as described in #248.

Since in the future we might want to do all image building using loop devices (see #709), I don't think we want to explicitly support building in a nested container.

Of course, as soon as the kernel adds support for virtualized loop devices, we'll be happy to support running mkosi in containers, but we need kernel changes for that to happen.

rezib:
> I assume you're using the directory, btrfs subvolume or tar output?

Yes indeed, I only tested with tar and directory outputs.

> Since in the future we might want to do all image building using loop devices (see #709), I don't think we want to explicitly support building in a nested container.

OK, it sounds perfectly legit.

Does `--keep-unit` hurt though? This setting seems also fair when `mkosi` is run by a system service, as I would expect it to use the resources allocated to that particular service, doesn't it?

DaanDeMeyer:
> Does --keep-unit hurt though? This setting seems also fair when mkosi is run by a system service, as I would expect it to use the resources allocated to that particular service, doesn't it?

But it shouldn't be added when mkosi is executed from the CLI. We could add an option that adds --keep-unit to nspawn calls I guess.

man/mkosi.1
```diff
@@ -1109,6 +1109,12 @@ When used with the \f[C]qemu\f[R] verb, this options sets
 guest\[cq]s RAM.
 Defaults to \f[C]1G\f[R].
 .TP
+\f[B]\f[CB]NspawnKeepUnit=\f[B]\f[R], \f[B]\f[CB]--nspawn-keep-unit\f[B]\f[R]
+When used, this option instructs underlying calls of systemd-nspawn to
+use the current unit scope, instead of creating a dedicated transcient
+scope unit for the containers. This option should be used when mkosi is
+run by a service unit.
+.TP
```
behrmann, line 1117: You don't need to write raw roff manually, edit `mkosi.md`, the man page is generated from it.

rezib:
Hi @DaanDeMeyer, I just force-pushed a new commit that adds `--nspawn-keep-unit` in the _Host configuration_ group for this purpose. Does it look OK?

man/mkosi.1
rezib: Thank you @behrmann, this is fixed in latest push.

behrmann:
Looks reasonable to me, but damn, do we have too many run functions doing mostly the same ever so slightly differently.

PR #893, opened 2022-01-24
Removal of Gentoo check at parse_args().

gsegatti:
The Gentoo check at parse_args(), while allowing the system to fail early, makes it harder to unit test such function since it will fail unless Gentoo modules are installed. By removing this, testings for argument parsing are executable on any distro regardless of the installation of Gentoo's modules.

Since this check is also done at __init__ in _gentoo.py_, we're able to remove this without compromising Mkosi's capacity of functioning properly.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -5711,10 +5711,6 @@ def parse_args(argv: Optional[List[str]] = None) -> Dict[str, argparse.Namespace
             # Parse again with any extra distribution files included.
             args = parse_args_file_group(argv, os.fspath(default_path), args.distribution)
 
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
Let's add a bit more context to the commit message. While the gentoo check is useful as it allows us to fail early if the required modules are not available, it makes unit testing harder as the parse_args() method won't execute unless the required gentoo modules are installed. Removing the check from parse_args() allows the argparse tests to run on any distro instead of requiring the gentoo modules to be installed when we're testing anything gentoo specific.

DaanDeMeyer (approved):

PR #894, opened 2022-01-26
Machine class for integration testing

gsegatti:
The goal is to eventually have an interface that facilitates integration testing. This is a class that aims to provide an abstraction for building/booting/running images generated by Mkosi via a python interface. 

This is set to be done by creating a context manager:
- Mkosi parameters can be sent via class constructor or config files.
- The enter method builds and boots (via qemu, shell, boot verbs) the image.
- This is done by using `parse_args()`, `load_args` and parts of `run_verb()`.
- Instead of calling `run()` from _backend.py_, we initialise a process by giving to pexpect the command line generated by `run_shell()` or `run_qemu`.
- Both functions above were slightly refactored so that the `cmdline` can be returned to this class.
- Commands can be sent via run() method and its results are displayed on screen.
- SIGTERM is sent whenever exiting the context manager.

This is a work in progress, currently systemd-run is used for running commands on images built with "boot", while ssh is used for "qemu".

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -64,6 +64,7 @@ from typing import (
 )
 
 from .backend import (
+    _FILE,
     ARG_DEBUG,
     Distribution,
     ManifestFormat,
@@ -7409,7 +7410,7 @@ def ensure_networkd(args: MkosiArgs) -> bool:
     return True
 
 
-def run_shell(args: MkosiArgs) -> None:
+def run_shell_cmdline(args: MkosiArgs) -> List[str]:
     if args.output_format in (OutputFormat.directory, OutputFormat.subvolume):
         target = f"--directory={args.output}"
     else:
@@ -7448,8 +7449,12 @@ def run_shell(args: MkosiArgs) -> None:
             cmdline += ["--"]
         cmdline += args.cmdline
 
+    return cmdline
+
+
+def run_shell(args: MkosiArgs) -> None:
     with suppress_stacktrace():
-        run(cmdline, stdout=sys.stdout, stderr=sys.stderr)
+        run(run_shell_cmdline(args), stdout=sys.stdout, stderr=sys.stderr)
 
 
 def find_qemu_binary() -> str:
@@ -7534,7 +7539,8 @@ def find_ovmf_vars() -> Path:
     die("Couldn't find OVMF UEFI variables file.")
 
 
-def run_qemu(args: MkosiArgs) -> None:
+@contextlib.contextmanager
+def run_qemu_cmdline(args: MkosiArgs) -> Iterator[List[str]]:
     has_kvm = os.path.exists("/dev/kvm")
     accel = "kvm" if has_kvm else "tcg"
 
@@ -7622,7 +7628,11 @@ def run_qemu(args: MkosiArgs) -> None:
         cmdline += args.cmdline
 
         print_running_cmd(cmdline)
+        yield cmdline
 
+
+def run_qemu(args: MkosiArgs) -> None:
+    with run_qemu_cmdline(args) as cmdline:
         with suppress_stacktrace():
             run(cmdline, stdout=sys.stdout, stderr=sys.stderr)
 
@@ -7679,7 +7689,16 @@ def find_address(args: MkosiArgs) -> Tuple[str, str]:
     die("Container/VM address not found")
 
 
-def run_ssh(args: MkosiArgs) -> None:
+def run_command_image(args: MkosiArgs, commands: Sequence[str], timeout: int, check: bool, stdout: _FILE = sys.stdout, stderr: _FILE = sys.stderr) -> CompletedProcess:
+    if args.verb == "qemu":
+        return run_ssh(args, commands, check, stdout, stderr, timeout)
+    else:
+        cmdline = ["systemd-run", "--quiet", "--wait", "--pipe", "-M", virt_name(args), "/usr/bin/env", *commands]
+        with suppress_stacktrace():
+            return run(cmdline, check=check, stdout=stdout, stderr=stderr, text=True, timeout=timeout)
+
+
+def run_ssh_cmdline(args: MkosiArgs, commands: Optional[Sequence[str]] = None) -> Sequence[str]:
     cmd = [
             "ssh",
             # Silence known hosts file errors/warnings.
@@ -7706,10 +7725,22 @@ def run_ssh(args: MkosiArgs) -> None:
         cmd += ["-p", f"{args.ssh_port}"]
 
     dev, address = find_address(args)
-    cmd += [f"root@{address}{dev}", *args.cmdline]
+    cmd += [f"root@{address}{dev}"]
+    cmd += commands or args.cmdline
 
+    return cmd
+
+
+def run_ssh(
+    args: MkosiArgs,
+    commands: Optional[Sequence[str]] = None,
+    check: bool = True,
+    stdout: _FILE = sys.stdout,
+    stderr: _FILE = sys.stderr,
+    timeout: Optional[int] = None,
+) -> CompletedProcess:
     with suppress_stacktrace():
-        run(cmd, stdout=sys.stdout, stderr=sys.stderr)
+        return run(run_ssh_cmdline(args, commands), check=check, stdout=stdout, stderr=stderr, text=True, timeout=timeout)
 
 
 def run_serve(args: MkosiArgs) -> None:
--- a/mkosi/machine.py
+++ b/mkosi/machine.py
@@ -1,0 +1,130 @@
+# SPDX-License-Identifier: LGPL-2.1+
+
+from __future__ import annotations
+
+import contextlib
+import signal
+import subprocess
+from textwrap import dedent
+from typing import Any, Optional, Sequence
+
+import pexpect  # type: ignore
+
+from . import (
+    MKOSI_COMMANDS_SUDO,
+    CompletedProcess,
+    build_stuff,
+    check_native,
+    check_output,
+    check_root,
+    init_namespace,
+    load_args,
+    needs_build,
+    parse_args,
+    prepend_to_environ_path,
+    run_command_image,
+    run_qemu_cmdline,
+    run_shell_cmdline,
+    unlink_output,
+)
+from .backend import MkosiArgs, die
+
+
+class Machine:
+    def __init__(self, args: Optional[Sequence[str]] = None, debug: bool = False) -> None:
+        # Remains None until image is built and booted, then receives pexpect process
+        self._serial: Optional[pexpect.spawn] = None
+        self.exit_code: int = -1
+        self.debug = debug
+        self.stack = contextlib.ExitStack()
+        self.args: MkosiArgs
+
+        # We make sure to add the arguments in the machine class itself, rather than typing this for every testing function.
+        tmp = parse_args(args)["default"]
+        tmp.force = 1
+        tmp.autologin = True
+        if tmp.verb == "qemu":
+            tmp.bootable = True
+            tmp.qemu_headless = True
+            tmp.hostonly_initrd = True
+            tmp.network_veth = True
+            tmp.ssh = True
+        elif tmp.verb == "boot":
+            pass
+        else:
+            die("No valid verb was entered.")
+
+        self.args = load_args(tmp)
+
+    @property
+    def serial(self) -> pexpect.spawn:
+        if self._serial is None:
+            raise ValueError(
+                        dedent(
+                            """\
+                            Trying to access serial console before machine boot or after machine shutdown.
+                            In order to boot the machine properly, use it as a context manager.
+                            Then, a Mkosi image will be booted in the __enter__ method.
+                            """
+                        )
+                    )
+        return self._serial
+
+    def ensure_booted(self) -> None:
+        # Try to access the serial console which will raise an exception if the machine is not currently booted.
+        assert self._serial is not None
+
+    def __enter__(self) -> Machine:
+        if self.args.verb in MKOSI_COMMANDS_SUDO:
+            check_root()
+            unlink_output(self.args)
+
+        if self.args.verb == "build":
+            check_output(self.args)
+
+        if needs_build(self.args):
+            check_root()
+            check_native(self.args)
+            init_namespace(self.args)
+            build_stuff(self.args)
+
+        with contextlib.ExitStack() as stack:
+            prepend_to_environ_path(self.args.extra_search_paths)
+
+            if self.args.verb in ("shell", "boot"):
+                cmdline = run_shell_cmdline(self.args)
+            elif self.args.verb == "qemu":
+                # We must keep the temporary file opened at run_qemu_cmdline accessible, hence the context stack.
+                cmdline = stack.enter_context(run_qemu_cmdline(self.args))
+            else:
+                die("No valid verb was entered.")
+
+            cmd = " ".join(str(x) for x in cmdline)
+
+            # Here we have something equivalent to the command lines used on spawn() and run() from backend.py
+            # We use pexpect to boot an image that we will be able to interact with in the future
+            # Then we tell the process to look for the # sign, which indicates the CLI for that image is active
+            # Once we've build/boot an image the CLI will prompt "root@image ~]# "
+            # Then, when pexpects finds the "#" it means we're ready to interact with the process
+            self._serial = pexpect.spawnu(cmd, logfile=None, timeout=240)
+            self._serial.expect("#")
+            self.stack = stack.pop_all()
+
+        return self
+
+    def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -> CompletedProcess:
+        self.ensure_booted()
+
+        process = run_command_image(self.args, commands, timeout, check, subprocess.PIPE, subprocess.PIPE)
+        if self.debug:
+            print(f"Stdout:\n {process.stdout}")
+            print(f"Stderr:\n {process.stderr}")
+
+        return process
+
+    def __exit__(self, *args: Any, **kwargs: Any) -> None:
+        if self._serial:
+            self._serial.kill(signal.SIGTERM)
+            self.exit_code = self._serial.wait()
+            self._serial = None
+        self.stack.__exit__(*args, **kwargs)
--- a/tests/test_machine.py
+++ b/tests/test_machine.py
@@ -1,0 +1,69 @@
+# SPDX-License-Identifier: LGPL-2.1+
+
+import os
+from subprocess import TimeoutExpired
+
+import pytest
+
+import mkosi.machine as machine
+from mkosi.backend import MkosiException
+
+
+pytestmark = [
+    pytest.mark.integration,
+    pytest.mark.parametrize("verb", ["boot", "qemu"]),
+    pytest.mark.skipif(os.getuid() != 0, reason="Must be invoked as root.")
+]
+
+
+def test_simple_run(verb: str) -> None:
+    with machine.Machine([verb]) as m:
+        p = m.run(["echo", "This is a test."])
+        assert "This is a test." == p.stdout.strip("\n")
+
+    assert m.exit_code == 0
+
+
+def test_wrong_command(verb: str) -> None:
+    # First tests with argument check = True from mkosi.backend.run(), therefore we see if an exception is raised
+    with machine.Machine([verb]) as m:
+        with pytest.raises(MkosiException):
+            m.run(["NonExisting", "Command"])
+        with pytest.raises(MkosiException):
+            m.run(["ls", "NullDirectory"])
+
+    assert m.exit_code == 0
+
+    # Second group of tests with check = False to see if stderr and returncode have the expected values
+    with machine.Machine([verb]) as m:
+        result = m.run(["NonExisting", "Command"], check=False)
+        assert result.returncode in (203, 127)
+
+        result = m.run(["ls", "-"], check=False)
+        assert result.returncode == 2
+        assert "No such file or directory" in result.stderr
+
+    assert m.exit_code == 0
+
+
+def test_infinite_command(verb: str) -> None:
+    with machine.Machine([verb]) as m:
+        with pytest.raises(TimeoutExpired):
+            m.run(["tail", "-f", "/dev/null"], 2)
+
+    assert m.exit_code == 0
+
+
+def test_before_boot(verb: str) -> None:
+    m = machine.Machine([verb])
+    with pytest.raises(AssertionError):
+        m.run(["ls"])
+
+
+def test_after_shutdown(verb: str) -> None:
+    with machine.Machine([verb]) as m:
+        pass
+
+    with pytest.raises(AssertionError):
+        m.run(["ls"])
+    assert m.exit_code == 0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,15 +103,29 @@ jobs:
       group: ${{ github.workflow }}-${{ github.ref }}
       cancel-in-progress: true
 
+    strategy:
+      matrix:
+        distro:
+        - fedora
+
     steps:
     - uses: actions/checkout@v2
     - uses: ./
+
+    - name: Configure test distribution
+      run: |
+        mkdir -p mkosi.default.d
+
+        tee mkosi.default.d/test-distribution.conf <<- EOF
+        [Distribution]
+        Distribution=${{ matrix.distro }}
+        EOF
 
     - name: Install dependencies
       run: sudo apt-get update && sudo apt-get install --no-install-recommends python3-pexpect python3-pytest
 
     - name: Run integration tests
-      run: sudo python3 -m pytest tests
+      run: sudo python3 -m pytest -m integration tests
 
   integration-test:
     runs-on: ubuntu-20.04


machine.py
```diff
@@ -0,0 +1,102 @@
...
+        parsed_args = parsed_args["default"]
+        self.args = mkosi.load_args(parsed_args)
+
+        mkosi.prepend_to_environ_path(self.args.extra_search_paths)
+
+        if self.args.verb == "genkey":
+            mkosi.generate_secure_boot_key(self.args)
+
+        if self.args.verb == "bump":
+            mkosi.bump_image_version(self.args)
```
DaanDeMeyer, line 32: We don't need these in integration tests so these lines can be removed

machine.py
```diff
@@ -0,0 +1,102 @@
...
+
+        if self.args.verb in mkosi.MKOSI_COMMANDS_SUDO:
+            mkosi.check_root()
+            mkosi.unlink_output(self.args)
+
+        if self.args.verb == "build":
+            mkosi.check_output(self.args)
+
+        #if args.verb == "summary":
+            #print_summary(args)
```
DaanDeMeyer, line 42: Idem, don't need this

machine.py
```diff
@@ -0,0 +1,102 @@
...
+from mkosi.backend import Distribution, MkosiArgs, MkosiException, ARG_DEBUG, tmp_dir
+
+class Machine():
+    def __init__(self):
+        # Remains None until image is built and booted
+        self.serial = None
+        self.args = None
+
+    def __enter__(self):
+        parsed_args = mkosi.parse_args(sys.argv[1:])
```
DaanDeMeyer, line 22: Instead of using sys.argv, let's only allow the user to pass arguments via the constructor.

mkosi/__init__.py
```diff
@@ -7370,7 +7370,7 @@ def ensure_networkd(args: MkosiArgs) -> bool:
     return True
 
 
-def run_shell(args: MkosiArgs) -> None:
+def pre_run_shell(args: MkosiArgs) -> MkosiArgs:
```
DaanDeMeyer, line 7373: Instead of "pre_run_shell()", I'd name this function "run_shell_cmdline()" which is less generic and more clearly indicates that the function returns the cmdline for run_shell()

Also, wrong return type here, should be Sequence[str]

mkosi/__init__.py
```diff
@@ -7494,96 +7498,102 @@ def find_ovmf_vars() -> Path:
 
     die("Couldn't find OVMF UEFI variables file.")
 
+@contextlib.contextmanager
+def pre_run_qemu(args: MkosiArgs) -> [str]:
+    try:
```
DaanDeMeyer, line 7503: The try, finally here isn't necessary so let's remove it.

mkosi/__init__.py
```diff
@@ -7494,96 +7498,102 @@ def find_ovmf_vars() -> Path:
 
     die("Couldn't find OVMF UEFI variables file.")
 
+@contextlib.contextmanager
+def pre_run_qemu(args: MkosiArgs) -> [str]:
```
DaanDeMeyer, line 7502: Similar to above, let's name this context manager "run_qemu_cmdline()"

The type here is wrong (or maybe this is newer syntax? In that case, we need to support old python so we can't use newer typing syntax yet), It should be `Iterator[Sequence[str]]`

machine.py
```diff
@@ -0,0 +1,102 @@
...
+
+            mkosi.save_manifest(self.args, manifest)
+
+            mkosi.print_output_size(self.args)
+
+        if self.args.verb in ("shell", "boot"):
+            self.serial = pexpect.spawnu(" ".join(mkosi.pre_run_shell(self.args)), logfile=sys.stdout, timeout=240)
+
+        if self.args.verb == "qemu":
+            with mkosi.pre_run_qemu(self.args) as cmdline:
```
DaanDeMeyer, line 61: Hmm, so one thing I didn't consider here is that pexpect will continue immediately after starting the process which means we'll exit the context manager and the temporary file allocated by pre_run_qemu() will be removed.

The solution here is store a context stack as part of the Machine class that will keep the context manager alive for as long as needed. I'll explain this in more detail in the chat we'll have today.

DaanDeMeyer (commented):
Added a few suggestions. The initial WIP looks good though! Looking forward to reviewing the final result

mkosi/machine.py
```diff
@@ -0,0 +1,97 @@
...
+
+import pexpect  # type: ignore
+
+import mkosi
+
+from .backend import die
+
+
+class Machine():
+    def __init__(self, args: Optional[List[str]] = None) -> None:
```
DaanDeMeyer, line 18: Let's make args `Sequence[str]` instead of `List[str]`. `Sequence` is a read-only iterable and since we're not gonna modify it we should specify so instead of fixing the type to be `List`.

tests/test_machine.py
```diff
@@ -0,0 +1,21 @@
...
+
+@pytest.mark.parametrize("boot", [["-d", "fedora", "-f", "--autologin", "--bootable", "--qemu-headless", "--hostonly-initrd", "qemu"], \
+                                    ["-d", "fedora", "-f", "--autologin", "boot"], \
+                                    ["-d", "fedora", "-f", "--autologin", "shell"]])
+
+
+@pytest.mark.skipif(os.getuid() != 0, reason="Must be invoked as root.")
+def test_run(boot):
+    m = machine.Machine(boot)
+    with m:
```
DaanDeMeyer, line 17: Python doesn't have block scoped variables, so you can actually write this as:

```
with machine.Machine() as m:
```

and m will still be valid after the with block is finished

tests/test_machine.py
```diff
@@ -0,0 +1,21 @@
...
+
+import os
+
+import mkosi.machine as machine
+import pytest
+
+
+@pytest.mark.parametrize("boot", [["-d", "fedora", "-f", "--autologin", "--bootable", "--qemu-headless", "--hostonly-initrd", "qemu"], \
+                                    ["-d", "fedora", "-f", "--autologin", "boot"], \
+                                    ["-d", "fedora", "-f", "--autologin", "shell"]])
```
DaanDeMeyer, line 11: This is a bit redundant, let's just put the verb in the parametrize and then add the other arguments in the test itself based on the verb used.

DaanDeMeyer:
Needs a rebase now that #897 is merged.

mkosi/machine.py
DaanDeMeyer: This should work now that #898 landed

mkosi/machine.py
```diff
@@ -0,0 +1,97 @@
...
+        if self.verb in ("boot", "qemu"):
+            self.run("systemctl exit")
+        if self.verb == "shell":
+            self.run("exit")
+
+        self.serial.expect(pexpect.EOF)
+
+
+    def __exit__(self, *args: Any, **kwargs: Any) -> None:
+        self.poweroff()
```
DaanDeMeyer, line 93: Since we already send sigterm, there's no need to call this method here.

mkosi/machine.py
```diff
@@ -0,0 +1,97 @@
...
+        self.serial.kill(signal.SIGTERM)
+        self.stack.__exit__(*args, **kwargs)
```
DaanDeMeyer, line 95: Calling this immediately means we don't wait for the image to exit completely before we delete any temporary files managed by the context stack. This should be called after `self.serial.wait()`

mkosi/machine.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class Machine():
+    def __init__(self, args: Optional[Sequence[str]] = None) -> None:
+        # Remains None until image is built and booted, then receives pexpect process
+        self.serial: pexpect.spawnu
+        self.stack = contextlib.ExitStack()
+        self.args = args
+        self.exit_code: int
+        self.verb: str
```
DaanDeMeyer, line 24: We should always assign default values for all fields in a class. Otherwise, trying to access these fields before they are set will throw a non-useful exception. For serial and verb, with the current design, we'll have to use `None`.

mkosi/machine.py
```diff
@@ -0,0 +1,97 @@
...
+                manifest = mkosi.build_stuff(loaded_args)
+
+                if loaded_args.auto_bump:
+                    mkosi.bump_image_version(loaded_args)
+
+                mkosi.save_manifest(loaded_args, manifest)
+
+                mkosi.print_output_size(loaded_args)
+
+            cmdline = None
```
DaanDeMeyer, line 57: python variables are function scoped, not block scoped so this initialization is not needed.

mkosi/machine.py
```diff
@@ -0,0 +1,97 @@
...
+            if loaded_args.verb in ("shell", "boot"):
+                cmdline = mkosi.run_shell_cmdline(loaded_args)
+            elif loaded_args.verb == "qemu":
+                # We must keep the temporary file opened at run_qemu_cmdline accessible, hence the context stack.
+                cmdline = stack.enter_context(mkosi.run_qemu_cmdline(loaded_args))
+            else:
+                die("No valid verb was entered.")
```
DaanDeMeyer, line 64: Let's replace "entered" with "provided"

mkosi/machine.py
```diff
@@ -0,0 +1,97 @@
...
+        self.serial.sendline(cmdline)
+
+
+    def poweroff(self) -> None:
+        if self.verb in ("boot", "qemu"):
+            self.run("systemctl exit")
+        if self.verb == "shell":
+            self.run("exit")
+
+        self.serial.expect(pexpect.EOF)
```
DaanDeMeyer, line 89: I'd remove these methods altogether. The serial console isn't reliable so trying to implement functionality on top of it isn't going to be very reliable either. Users using the serial console can easily end up running commands that make the serial console inaccessible which means we won't be able to use it to shut down the image.

Sending SIGTERM like we already do in __exit__ is sufficient to shut down the image. If users want to run commands in images via the serial console, let's just have them do so directly via the pexpect API.

tests/test_machine.py
```diff
@@ -0,0 +1,25 @@
...
+
+import os
+
+import mkosi.machine as machine
+import pytest
+
+
+@pytest.mark.parametrize("verb", ["qemu", "boot", "shell"])
+
+
```
DaanDeMeyer, line 11: Unnecessary newlines here

DaanDeMeyer (changes requested):

mkosi/machine.py
gsegatti: that's really neat, thanks!

mkosi/machine.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def machine_run(self, command: List[str]) -> None:
+        self.serial.expect("#")
+
+        if self.args.verb == "boot":
+            cmdline = ['sudo', 'systemd-run']
+            cmdline += ["--quiet", "--wait", "--pipe"]
+            cmdline += ["-M", virt_name(self.args)]
+        else:
+            cmdline = ['sudo', 'venv/bin/python', '-m', "mkosi", "ssh"]
```
DaanDeMeyer, line 107: This won't work for a multitude of reasons. There's no reason to call mkosi as a subprocess when we can simply import the relevant function and run it directly. This should be replaced with a call to run_ssh(). run_ssh() itself might have to be modified a little so that we can capture its output.

The new run_ssh() signature should look something like this:

```
def run_ssh(args: MkosiArgs, cmd: Optional[Sequence[str]] = None, stdout: _FILE = None, stderr: _FILE = None) -> CompletedProcess:
```

When using it from the Machine class, we'd call it like this:

```
result  = run_ssh(self.args, command, stdout=PIPE, stderr=PIPE)
```

And then we can access `result.stdout` to get the stdout output of the command and `result.stderr` to get the stderr output of the command.

mkosi/machine.py
```diff
@@ -0,0 +1,129 @@
...
+
+            cmdline = [str(x) for x in cmdline]
+            cmdline = " ".join(cmdline)
+
+            self.serial = cmdline
+            self.stack = stack.pop_all()
+
+        return self
+
+    def machine_run(self, command: List[str]) -> None:
```
DaanDeMeyer, line 99: We need to return the output of the command from `machine_run()`

mkosi/machine.py
```diff
@@ -0,0 +1,129 @@
...
+        self.serial.expect("#")
+
+        if self.args.verb == "boot":
+            cmdline = ['sudo', 'systemd-run']
```
DaanDeMeyer, line 103: No need for sudo here, we're already running as the root user.

mkosi/machine.py
```diff
@@ -0,0 +1,129 @@
...
+            cmdline = [str(x) for x in cmdline]
+            cmdline = " ".join(cmdline)
+
+            self.serial = cmdline
+            self.stack = stack.pop_all()
+
+        return self
+
+    def machine_run(self, command: List[str]) -> None:
+        self.serial.expect("#")
```
DaanDeMeyer, line 100: This shouldn't be in machine_run(). We only need to wait once for the machine to boot and we should do that in `__enter__`.

mkosi/machine.py
```diff
@@ -0,0 +1,129 @@
...
+            elif self.args.verb == "qemu":
+                # We must keep the temporary file opened at run_qemu_cmdline accessible, hence the context stack.
+                cmdline = stack.enter_context(run_qemu_cmdline(self.args))
+            else:
+                die("No valid verb was entered.")
+
+            cmdline = [str(x) for x in cmdline]
+            cmdline = " ".join(cmdline)
+
+            self.serial = cmdline
```
DaanDeMeyer, line 94: I'd replace this with 

```
self._serial = pexpect.spawnu(cmdline, logfile=None, timeout=240)
```

and remove the property setter for `serial`. It's unintuitive to assign a list to `serial` when the final value will be a pexpect process.

mkosi/machine.py
```diff
@@ -0,0 +1,129 @@
...
+    unlink_output,
+    virt_name,
+)
+from .backend import die, run
+
+
+class Machine:
+    def __init__(self, args: Optional[Sequence[str]] = None) -> None:
+        # Remains None until image is built and booted, then receives pexpect process
+        self._serial: Optional[pexpect.spawnu] = None
```
DaanDeMeyer, line 40: Type here is wrong I think, the function is called `spawnu` but the object returned by `spawnu` is `pexpect.spawn`. If you fix this you might also be able to remove the `type: ignore` above.

mkosi/machine.py
```diff
@@ -0,0 +1,129 @@
...
+        self.stack = contextlib.ExitStack()
+        self.args = load_args(parse_args(args)["default"])
+        self.verb = self.args.verb
+        self.exit_code: int
```
DaanDeMeyer, line 44: TThis still needs a default value

mkosi/machine.py
```diff
@@ -0,0 +1,129 @@
...
+            cmdline = ['sudo', 'venv/bin/python', '-m', "mkosi", "ssh"]
+
+        cmdline += command
+        with suppress_stacktrace():
+            run(cmdline, stdout=sys.stdout, stderr=sys.stderr)
+
+
+    def __exit__(self, *args: Any, **kwargs: Any) -> None:
+        self.serial.wait()
+        self.serial.kill(signal.SIGTERM)
```
DaanDeMeyer, line 116: I don't think this can work. You first wait for the pexpect process to exit, and only after it has exited you send the SIGTERM signal? This will probably result in self.serial.wait() blocking until it times out which is not what we want. We should first send the signal and then wait for the pexpect process to exit using the wait() method.

DaanDeMeyer (changes requested):
I'm confused why you removed `run_command_image()` altogether and moved all the logic to the Machine class. This means we won't be able to reuse the logic we implement here to add a `mkosi run` verb in the future.

mkosi/machine.py
gsegatti: The only reason I had the setter was to stop mypy from saying `error: Item "None" of "Optional[Any]" has no attribute "kill"`. But agree, definitely didn't look good.

mkosi/backend.py
```diff
@@ -103,6 +103,9 @@ def from_string(cls: Any, name: str) -> Any:
     def parse_list(cls: Any, string: str) -> List[Any]:
         return [cls.from_string(p) for p in string.split(",") if p]
 
+class SystemdRunMethod(enum.Enum):
+    machine = "-M"
+    ssh = "-H"
```
DaanDeMeyer, line 108: This can be removed

mkosi/__init__.py
```diff
@@ -7663,10 +7683,14 @@ def run_ssh(args: MkosiArgs) -> None:
         cmd += ["-p", f"{args.ssh_port}"]
 
     dev, address = find_address(args)
-    cmd += [f"root@{address}{dev}", *args.cmdline]
+    cmd += [f"root@{address}{dev}"]
+    if commands:
+        cmd += commands
+    else:
+        cmd += [*args.cmdline]
```
DaanDeMeyer, line 7690: You can write this as `cmd += commands or args.cmdline` I think

mkosi/__init__.py
```diff
@@ -6525,6 +6527,7 @@ def load_args(args: argparse.Namespace) -> MkosiArgs:
     if args.base_image is not None:
         args.base_packages = False
 
+    #print(MkosiArgs(**vars(args)))
```
DaanDeMeyer, line 6530: This should be removed as well.

mkosi/__init__.py
```diff
@@ -7635,8 +7645,18 @@ def find_address(args: MkosiArgs) -> Tuple[str, str]:
 
     die("Container/VM address not found")
 
+def run_command_image(args: MkosiArgs, commands: List[str], stdout: _FILE = sys.stdout, stderr: _FILE = sys.stderr) -> CompletedProcess:
+    if args.verb == "qemu":
+        return run_ssh(args, commands, stdout, stderr)
+    else:
+        cmdline = ['systemd-run', "--quiet", "--wait", "--pipe"]
+        cmdline += ["-M", virt_name(args)]
+        cmdline += commands
```
DaanDeMeyer, line 7654: You can merge all this into the same line

mkosi/machine.py
```diff
@@ -0,0 +1,104 @@
...
+                check_native(self.args)
+                init_namespace(self.args)
+                manifest = build_stuff(self.args)
+
+                if self.args.auto_bump:
+                    bump_image_version(self.args)
+
+                save_manifest(self.args, manifest)
+
+                print_output_size(self.args)
```
DaanDeMeyer, line 76: We don't need this stuff.

mkosi/machine.py
```diff
@@ -0,0 +1,104 @@
...
+                unlink_output(self.args)
+
+            if self.args.verb == "build":
+                check_output(self.args)
+
+            if needs_build(self.args):
+                check_root()
+                check_native(self.args)
+                init_namespace(self.args)
+                manifest = build_stuff(self.args)
```
DaanDeMeyer, line 69: All this can be moved outside of the `with` statement

mkosi/machine.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def machine_run(self, commands: List[str], stdout: _FILE = sys.stdout, stderr: _FILE = sys.stderr) -> CompletedProcess:
+        return run_command_image(self.args, commands, PIPE, PIPE)
+
+
+    def __exit__(self, *args: Any, **kwargs: Any) -> None:
+        self._serial.kill(signal.SIGTERM)
+        self._serial.wait()
+        self.stack.__exit__(*args, **kwargs)
+        self.exit_code = self._serial.wait()
```
DaanDeMeyer, line 104: We call wait() twice now which is redundant. You can simply remove the first call to wait() with the second call to wait()

mkosi/machine.py
```diff
@@ -0,0 +1,104 @@
...
+            cmdline = " ".join(cmdline)
+
+            self._serial = pexpect.spawnu(cmdline, logfile=None, timeout=240)
+            self._serial.expect("#")
+            self.stack = stack.pop_all()
+
+        return self
+
+
+    def machine_run(self, commands: List[str], stdout: _FILE = sys.stdout, stderr: _FILE = sys.stderr) -> CompletedProcess:
```
DaanDeMeyer, line 96: This can be named `run()` since it'll be used a lot so we want to keep it short

mkosi/machine.py
```diff
@@ -0,0 +1,104 @@
...
+        return run_command_image(self.args, commands, PIPE, PIPE)
```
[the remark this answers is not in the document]
gsegatti: Yeah, I didn't realize I wasn't using the parameters lol. But what if someone wants to direct run_command_image()'s output to the Pipe or Console, how would they do it without the arguments from machine_run()?

Unless maybe we remove machine_run() completely and start calling run_command_image() directly?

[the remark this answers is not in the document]
gsegatti: I'm also using the stdout/stderr arguments in the test_machine.py file in order to assert that the run function is running correctly, since the pexpect.expect() function wasn't able to find the output of subprocess.run(). So. now, I redirect output to PIPE and then assert it with .stdout.

mkosi/__init__.py
```diff
@@ -7663,10 +7679,12 @@ def run_ssh(args: MkosiArgs) -> None:
         cmd += ["-p", f"{args.ssh_port}"]
 
     dev, address = find_address(args)
-    cmd += [f"root@{address}{dev}", *args.cmdline]
+    cmd += [f"root@{address}{dev}"]
+
+    cmd += commands or [*args.cmdline]
```
DaanDeMeyer, line 7684: Shouldn't `cmd += commands or args.cmdline` work here as well? I'm curious why we need `[*args.cmdline]`

mkosi/__init__.py
```diff
@@ -7635,8 +7643,16 @@ def find_address(args: MkosiArgs) -> Tuple[str, str]:
 
     die("Container/VM address not found")
 
+def run_command_image(args: MkosiArgs, commands: List[str], stdout: _FILE = sys.stdout, stderr: _FILE = sys.stderr) -> CompletedProcess:
```
DaanDeMeyer, line 7646: commands can be `Sequence[str]`, no need for `List[str]` here.

mkosi/machine.py
```diff
@@ -0,0 +1,97 @@
...
+    def __init__(self, args: Optional[Sequence[str]] = None) -> None:
+        # Remains None until image is built and booted, then receives pexpect process
+        self._serial: Optional[pexpect.spawn] = None
+        self.stack = contextlib.ExitStack()
+        self.args = load_args(parse_args(args)["default"])
+        self.exit_code: int = 1
+
+
+    @property
+    def serial(self) -> pexpect.spawnu:
```
DaanDeMeyer, line 48: This should return `pexpect.spawn`

mkosi/machine.py
```diff
@@ -0,0 +1,97 @@
...
+            cmdline = " ".join(cmdline)
+
+            self._serial = pexpect.spawnu(cmdline, logfile=None, timeout=240)
+            self._serial.expect("#")
+            self.stack = stack.pop_all()
+
+        return self
+
+
+    def run(self, commands: List[str], stdout: _FILE = sys.stdout, stderr: _FILE = sys.stderr) -> CompletedProcess:
```
DaanDeMeyer, line 90: No need for `stdout` and `stderr` parameters here. We're always going to want to use `PIPE` here.

mkosi/__init__.py
```diff
@@ -7635,8 +7643,16 @@ def find_address(args: MkosiArgs) -> Tuple[str, str]:
...
+    if args.verb == "qemu":
+        return run_ssh(args, commands, stdout, stderr)
+    else:
+        cmdline = ['systemd-run', "--quiet", "--wait", "--pipe", "-M", virt_name(args)] + commands
+        with suppress_stacktrace():
+            return run(cmdline, stdout=stdout, stderr=stderr)
```
DaanDeMeyer, line 7652: I think you can add `text=True` here and in `run_ssh()` so that in the test you don't need to decode the output to UTF-8.

mkosi/machine.py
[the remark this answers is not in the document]
DaanDeMeyer: I agree that it's useful to have commands log to stdout/stderr for debugging, but if we implement that by passing sys.stdout and sys.stderr to run_command_image(), we'll break the corresponding assertions since the stdout and stderr fields of the CompletedProcess will be empty.

It would also be quite painful for users to have to change the stdout/stderr arguments for every call to machine_run() when debugging. Instead, we could have a `debug` field as part of `Machine` that can be toggled on/off by the user. When enabled, after `run_command_image` finishes, we print the stdout and stderr fields of CompletedProcess.

mkosi/machine.py
DaanDeMeyer: Ok, there's nothing we can do about the mypy error so keeping the type: ignore is fine

tests/test_machine.py
```diff
@@ -0,0 +1,25 @@
...
+def test_run(verb: str) -> None:
+    boot = ["-d", "fedora", "-f", "--autologin"]
+
+    if verb == "qemu":
+        boot += ["--bootable", "--qemu-headless", "--hostonly-initrd", "--network-veth", "--ssh", verb]
+    if verb == ("boot"):
+        boot += [verb]
+
+    with machine.Machine(boot) as m:
+        p = m.run(["echo", "This is a test."])
```
DaanDeMeyer, line 22: There's room for a lot more testing here:

- What happens if we run a command that fails?
- What happens if we run the same command a bunch of times in a loop?
- What if we run a command that never exits?
- What happens if we try to run a command before the machine has booted or after it has shut down?

tests/test_machine.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+@pytest.mark.parametrize("verb", ["qemu", "boot"])
+@pytest.mark.skipif(os.getuid() != 0, reason="Must be invoked as root.")
+@pytest.mark.integration
+def test_run(verb: str) -> None:
+    boot = ["-d", "fedora", "-f", "--autologin"]
+
+    if verb == "qemu":
+        boot += ["--bootable", "--qemu-headless", "--hostonly-initrd", "--network-veth", "--ssh", verb]
```
DaanDeMeyer, line 17: Having to add these arguments for every test would be rather repetitive and error prone no? The Machine class should make sure to add the arguments it needs itself.

tests/test_machine.py
```diff
@@ -0,0 +1,25 @@
...
+    if verb == ("boot"):
```
DaanDeMeyer, line 18: All these string comparisons against the verb are getting rather brittle. A good refactoring would be to introduce an enum to represent the verb, similar to how we use an enum to represent the current distribution.

mkosi/machine.py
```diff
@@ -0,0 +1,102 @@
...
+            self.stack = stack.pop_all()
+
+        return self
+
+
+    def run(self, commands: Sequence[str]) -> CompletedProcess:
+        process = run_command_image(self.args, commands, subprocess.PIPE, subprocess.PIPE)
+        if self.debug:
+            print("Stdout:\n" + process.stdout)
+            print("Stderr:\n" + process.stderr)
```
DaanDeMeyer, line 93: Use format strings here instead of string concatenation.

mkosi/machine.py
```diff
@@ -0,0 +1,102 @@
...
+
+        return process
+
+
+    def __exit__(self, *args: Any, **kwargs: Any) -> None:
+        if self._serial is not None:
+            self._serial.kill(signal.SIGTERM)
+            self.exit_code = self._serial.wait()
```
DaanDeMeyer, line 101: We should assign `None` again to `self._serial` after we kill the process so it can't be used after the machine is shut down.

mkosi/machine.py
```diff
@@ -0,0 +1,102 @@
...
+                # We must keep the temporary file opened at run_qemu_cmdline accessible, hence the context stack.
+                cmdline = stack.enter_context(run_qemu_cmdline(self.args))
+            else:
+                die("No valid verb was entered.")
+
+            cmdline = [str(x) for x in cmdline]
+            cmdline = " ".join(cmdline)
+
+            self._serial = pexpect.spawnu(cmdline, logfile=None, timeout=240)
+            self._serial.expect("#")
```
DaanDeMeyer, line 83: This could use a comment explaining what it does

DaanDeMeyer (changes requested):
I figure we can make the Machine class easier to use than it currently is. Also, try to add a few more tests to figure out whether writing multiple tests using the Machine class is a good experience.

gsegatti:
Fixed most of Daan's last comments, CI will probably fail due to isort/mypy stuff. Finish it tomorrow.

DaanDeMeyer (changes requested):

tests/test_machine.py
```diff
@@ -0,0 +1,52 @@
...
+
+import mkosi.machine as machine
+from mkosi.backend import MkosiException
+
+
+@pytest.mark.skipif(os.getuid() != 0, reason="Must be invoked as root.")
+@pytest.mark.parametrize("verb", ["boot", "qemu"])
+@pytest.mark.integration
+
+class TestVerbs:
```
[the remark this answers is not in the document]
gsegatti: Bc then I didn't need to repeat pytest.mark.parametrize and mark.integration for every function. Will remove the class then

mkosi/machine.py
```diff
@@ -0,0 +1,125 @@
...
+    def __init__(self, args: Optional[Sequence[str]] = None, debug: bool = False, test: bool = False) -> None:
+        # Remains None until image is built and booted, then receives pexpect process
+        self._serial: Optional[pexpect.spawn] = None
+        self.exit_code: int = -1
+        self.debug = debug
+        self.stack = contextlib.ExitStack()
+        self.args: MkosiArgs
+
+        # We make sure to add the arguments in the machine class itself, rather than typing this for every testing function.
+        if test:
```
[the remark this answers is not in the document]
gsegatti: Not really, this just replace the hard coded logid from here: https://github.com/systemd/mkosi/pull/894#discussion_r803608343

But maybe this should be removed altogether, if someone needs to type parameters multiple times, just use _parametrize_ from pytest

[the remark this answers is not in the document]
DaanDeMeyer: I think we're misunderstanding each other here.  Adding the necessary arguments depending on what the verb is in the constructor is a good thing. What I don't understand is why we gate this behind the `test` boolean argument. Right now, in all your tests, you specify `test=True`. What I don't understand is: in which scenario would users specify `test=False`?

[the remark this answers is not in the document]
gsegatti: That is gated behind `test` because I understood we would only autocomplete arguments to generate a fixed set of images (i.e the ones used in testing currently). So for every other scenario outside of the ones existing on test_machine.py all arguments would actually be specified through `args`.

tests/test_machine.py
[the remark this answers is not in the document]
DaanDeMeyer: Ahah, OK that makes sense. I had a look at the documentation and you can also define the `pytestmark` global variable which will do the same for all tests in this module. Since this is something we currently want for all tests, I'd just use that for now.

mkosi/machine.py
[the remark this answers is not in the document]
DaanDeMeyer: But doing so would make the Machine class quite error prone to use no? Everyone that tried to use it for the first time might forget to add the --ssh option for example and suddenly Machine.run() won't work correctly because there's no SSH inside the virtual machine.

So I'd argue that we definitely by default want to add these arguments.

In the future, there could be special use cases for very specific tests where we don't want to add these arguments, but when that happens we should add an option to disable the arguments from being added, instead of adding an option that has to be enabled to add the extra arguments. This way we'd do the right thing by default but still give users an escape hatch if they need it.

Let's wait with adding that option until we have a concrete use case though. For now I'd go ahead and just remove the `test` argument and always add the extra arguments.

mkosi/machine.py
```diff
@@ -0,0 +1,132 @@
...
+                            Then, a Mkosi image will be booted in the __enter__ method.
+                            """
+                        )
+                    )
+
+        return self._serial
+
+
+    def __enter__(self) -> Machine:
+        print("enter")
```
gsegatti, line 76: Please disregard this, will take it out on the next time I push something to not waste ci on this alone

mkosi/machine.py
```diff
@@ -0,0 +1,132 @@
...
+            pass
+        else:
+            die("No valid verb was entered.")
+
+        self.args = load_args(tmp)
+
+    @property
+    def serial(self) -> pexpect.spawn:
+        if self._serial is None:
+            raise   ValueError(
```
DaanDeMeyer, line 62: The whitespace here still isn't quite right, need just a single space after raise

mkosi/machine.py
```diff
@@ -0,0 +1,132 @@
...
+            init_namespace(self.args)
+            build_stuff(self.args)
+
+        with contextlib.ExitStack() as stack:
+            prepend_to_environ_path(self.args.extra_search_paths)
+
+            if self.args.verb in ("shell", "boot"):
+                cmdline = run_shell_cmdline(self.args)
+            elif self.args.verb == "qemu":
+                # We must keep the tmporary file opened at run_qemu_cmdline accessible, hence the context stack.
```
DaanDeMeyer, line 96: tmporary => temporary

mkosi/machine.py
```diff
@@ -0,0 +1,132 @@
...
+            self._serial = pexpect.spawnu(cmdline, logfile=None, timeout=240)
+            self._serial.expect("#")
+            self.stack = stack.pop_all()
+
+        return self
+
+
+    def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -> CompletedProcess:
+        # This property accessor throws an exception if serial is not defined
+        self.serial
```
DaanDeMeyer, line 117: What's the reasoning behind putting this here? I guess you want to check if the machine is booted? In that case, I would define a new method ensure_booted() and put this logic in there as it's quite unintuitive to read now.

tests/test_machine.py
```diff
@@ -0,0 +1,53 @@
...
+
+    assert m.exit_code == 0
+
+
+def test_wrong_command(verb: str) -> None:
+    with machine.Machine([verb]) as m:
+        with pytest.raises(MkosiException):
+            m.run(["NonExisting", "Command"], check=True)
+        with pytest.raises(MkosiException):
+            m.run(["ls", "NullDirectory"], check=True)
```
DaanDeMeyer, line 28: We should also test with `check=False`. And `check=True` is the default so no need to repeat it here.

mkosi/machine.py
gsegatti: I got it from here: https://github.com/systemd/mkosi/pull/894#discussion_r798688082, I reckon it made sense at the time. But, yeah, a method would be much better now

DaanDeMeyer:
@behrmann This is almost good to go for me, do you want to go over it as well before merging?

behrmann:
Will do, I haven't followed the changes over time, but I should have time later today to give this a look.

mkosi/__init__.py
```diff
@@ -7533,8 +7538,8 @@ def find_ovmf_vars() -> Path:
 
     die("Couldn't find OVMF UEFI variables file.")
 
-
```
behrmann, line 7536: This empty line is missing

mkosi/machine.py
```diff
@@ -0,0 +1,132 @@
...
+            die("No valid verb was entered.")
+
+        self.args = load_args(tmp)
+
+    @property
+    def serial(self) -> pexpect.spawn:
+        if self._serial is None:
+            raise   ValueError(
+                        dedent(
+                            """
```
behrmann, line 64: Is the additional newline here wanted? If not 
```suggestion
                            """\
```

mkosi/machine.py
```diff
@@ -0,0 +1,132 @@
...
+                            Trying to access serial console before machine boot or after machine shutdown.
+                            In order to boot the machine properly, use it as a context manager.
+                            Then, a Mkosi image will be booted in the __enter__ method.
+                            """
+                        )
+                    )
+
+        return self._serial
+
+
```
behrmann, line 74: This empty line is too much

mkosi/__init__.py
```diff
@@ -7533,8 +7538,8 @@ def find_ovmf_vars() -> Path:
...
-def run_qemu(args: MkosiArgs) -> None:
+@contextlib.contextmanager
+def run_qemu_cmdline(args: MkosiArgs) -> Iterator[Sequence[str]]:
```
behrmann, line 7542: Same here
```suggestion
def run_qemu_cmdline(args: MkosiArgs) -> Iterator[List[str]]:
```
this has the added benefit (if I might call it that), that later on at line 101 in `machine.py`, we get complaints, because both strings and lists are sequences, so mypy doesn't notice the type of the variable changing from a list to a string.

mkosi/machine.py
```diff
@@ -0,0 +1,132 @@
...
+                die("No valid verb was entered.")
+
+            cmdline = " ".join([str(x) for x in cmdline])
+
+            # Here we have something equivalent to the command lines used on spawn() and run() from backend.py.
+            # We use pexpect to boot an image that we will be able to interact with in the future.
+            # Then we tell the process to look for the # sign, which indicates the CLI for that image is active.
+            # Once we've build/boot an image the CLI will prompt "root@image ~]# "
+            # Then, when pexpects finds the "#" it means we're ready to interact with the process.
+            self._serial = pexpect.spawnu(cmdline, logfile=None, timeout=240)
```
behrmann, line 108: This is what I described just above. I changed the name from `cmdline` to `cmd`, since we disallow redefinition (which sometimes is a bit onerous, I'd be open to changing that).
```suggestion
            cmd = " ".join(str(x) for x in cmdline)

            # Here we have something equivalent to the command lines used on spawn() and run() from backend.py.
            # We use pexpect to boot an image that we will be able to interact with in the future.
            # Then we tell the process to look for the # sign, which indicates the CLI for that image is active.
            # Once we've build/boot an image the CLI will prompt "root@image ~]# "
            # Then, when pexpects finds the "#" it means we're ready to interact with the process.
            self._serial = pexpect.spawnu(cmd, logfile=None, timeout=240)
```
As an added comment: You don't need to create the intermediate list (first line of this hunk)

behrmann (commented):
Just had a quick look during lunch to paint some bikesheds and haven't followed the logic yet. 

Is there a reason the permissions for `mkosi/backend.py` are changed? That's unfortunately not something I can annotate.

mkosi/machine.py
gsegatti: Oh, will keep the redefinition rule in my head, probably a bad habit I carry. Removed the list as well!

gsegatti:
> Is there a reason the permissions for `mkosi/backend.py` are changed? That's unfortunately not something I can annotate.

It was I guess a bug within Meta's vscode that made me lose write access permission to all files, so I changed the permissions with chmod to see if it would go away, then I forgot to revert it back for backend.py. I changed it back, thanks for pointing it out!

mkosi/machine.py
```diff
@@ -0,0 +1,131 @@
...
+        process = run_command_image(self.args, commands, timeout, check, subprocess.PIPE, subprocess.PIPE)
+        if self.debug:
+            print(f"Stdout:\n {process.stdout}")
+            print(f"Stderr:\n {process.stderr}")
+
+        return process
+
+
+    def __exit__(self, *args: Any, **kwargs: Any) -> None:
+        if self._serial:
```
gsegatti, line 125: I'm a bit confused as to what I should do here. If I try:
`if ensure_booted:`
Mypy will point it out:
> mkosi/machine.py:127:13: error: Item "None" of "Optional[Any]" has no attribute "kill"  [union-attr]
                ^
> mkosi/machine.py:128:30: error: Item "None" of "Optional[Any]" has no attribute "wait"  [union-attr]

Did this only to stop it from failing the CI, but not sure if its good enough

mkosi/machine.py
```diff
@@ -0,0 +1,131 @@
...
+            raise ValueError(
+                        dedent(
+                            """\
+                            Trying to access serial console before machine boot or after machine shutdown.
+                            In order to boot the machine properly, use it as a context manager.
+                            Then, a Mkosi image will be booted in the __enter__ method.
+                            """
+                        )
+                    )
+        return self._serial
```
DaanDeMeyer, line 71: My suggestion was to add an additional method ensure_booted() that we could call from just `run()`, not to replace the serial property getter. The serial property was fine, we should keep it. On top of the serial property, we should add a method ensure_booted() that looks as follows:

```
def ensure_booted() -> None:
    # Try to access the serial console which will raise an exception if the machine is not currently booted.
    self.serial
```

And we'd call this from the run() method. The method is only there to clearly indicate to someone reading the code that we're accessing the self.serial property to verify that the machine is booted, which is not immediately obvious

gsegatti: ooooooooh, I see. Will add it back

mkosi/machine.py
gsegatti: [This](https://github.com/systemd/mkosi/pull/894#discussion_r807028122) would solve it then

mkosi/machine.py
```diff
@@ -0,0 +1,133 @@
...
+    @property
+    def serial(self) -> pexpect.spawn:
+        if self._serial is None:
+            raise ValueError(
+                            dedent("""Trying to access serial console before machine boot or after machine shutdown.
+                            In order to boot the machine properly, use it as a context manager.
+                            Then, a Mkosi image will be booted in the __enter__ method.
+                            """
+                                )
+                            )
```
DaanDeMeyer, line 69: Indentation is off again

mkosi/machine.py
DaanDeMeyer: What you currently have is fine.

mkosi/machine.py
gsegatti: oh, my bad. Will wait to push it if there's a next batch of comments to address.

mkosi/machine.py
```diff
@@ -0,0 +1,133 @@
...
+            tmp.network_veth = True
+            tmp.ssh = True
+        elif tmp.verb == "boot":
+            pass
+        else:
+            die("No valid verb was entered.")
+
+        self.args = load_args(tmp)
+
+
```
DaanDeMeyer, line 59: Need to be consistent with the whitespace between class methods as well. 2 newlines between functions in a file. 1 newline between methods in a class. I know its weird but it's how we do it everywhere so we should stick to it.

DaanDeMeyer:
LGTM aside from the remaining style nits. The python-integration-test CI is also failing That one needs to be green before we can merge this.

mkosi/machine.py
gsegatti: Oh, I really thought it was 2, since functions are the ones appearing the most. Will write that down so I don't forget.

behrmann: This is just [PEP 8](https://www.python.org/dev/peps/pep-0008/#blank-lines) :)

DaanDeMeyer:
> Whenever the CI calls sleep within Bash the following error appears:

This is a side effect of running things with systemd-run. A systemd unit always expects an absolute path to an executable. But if it works on the devserver maybe they changed it recently. You can fix it by prepending commands with /usr/bin/env.

behrmann:
> This is a side effect of running things with systemd-run. A systemd unit always expects an absolute path to an executable. But if it works on the devserver maybe they changed it recently. 

This surprised me recently as well. It was changed in [systemd 239](https://github.com/systemd/systemd/blob/main/NEWS#L5338)

gsegatti:
Also adding CI configuration to run Fedora instances since Ubuntu in causing SSH to fail with code 255.

mkosi/machine.py
```diff
@@ -0,0 +1,130 @@
...
+                            In order to boot the machine properly, use it as a context manager.
+                            Then, a Mkosi image will be booted in the __enter__ method.
+                            """
+                        )
+                    )
+        return self._serial
+
+    def ensure_booted(self) -> None:
+        # Try to access the serial console which will raise an exception if the machine is not currently booted.
+        self.serial
```
behrmann, line 75: Shouldn't this be?
```suggestion
        self._serial
```
Stylistically I do find this a bit hacky, maybe 
```suggestion
        assert self._serial is not None
```

tests/test_machine.py
```diff
@@ -0,0 +1,65 @@
...
+import os
+from subprocess import TimeoutExpired
+
+import pytest
+
+import mkosi.machine as machine
+from mkosi.backend import MkosiException
+
+
+pytestmark = [pytest.mark.integration, pytest.mark.parametrize("verb", ["boot", "qemu"]), pytest.mark.skipif(os.getuid() != 0, reason="Must be invoked as root.")]
```
behrmann, line 12: For better leggibility
```suggestion
pytestmark = [
    pytest.mark.integration, 
    pytest.mark.parametrize("verb", ["boot", "qemu"]), 
    pytest.mark.skipif(os.getuid() != 0, reason="Must be invoked as root.")
]
```

mkosi/__init__.py
```diff
@@ -7622,7 +7628,10 @@ def run_qemu(args: MkosiArgs) -> None:
         cmdline += args.cmdline
 
         print_running_cmd(cmdline)
+        yield cmdline
```
behrmann, line 7632: There's still a blank line missing here for PEP 8 ;)

mkosi/machine.py
gsegatti: That's actually way better, thanks

DaanDeMeyer:
LGTM, we can always do more follow up in subsequent PRs.

DaanDeMeyer (approved):
LGTM, we can always do more follow up in subsequent PRs.

PR #895, opened 2022-02-01
ci: Only run CI workflow if the unit tests workflow succeeded

daandemeyer:


behrmann:
Great change! Let's see whether it works as advertised. :)

behrmann:
Looks like the rest is missing?

DaanDeMeyer:
Trying some different syntax, from what I can understand from the docs, this should work

DaanDeMeyer:
That didn't work either, trying a different approach.

DaanDeMeyer:
Seems it works with this approach

image

DaanDeMeyer:
Let's merge since this works

PR #896, opened 2022-02-01
Update Unit Test CI for Pytest

gsegatti:
Adding sudo to the "Unit Tests" with Pytest portion of the CI, in order to allow using Mkosi's commands with root privilege when testing.

### DIFF
--- a/tests/test_parse_load_args.py
+++ b/tests/test_parse_load_args.py
@@ -1,0 +1,160 @@
+# SPDX-License-Identifier: LGPL-2.1+
+
+import argparse
+import tempfile
+import textwrap
+import uuid
+from contextlib import contextmanager
+from os import chdir, getcwd
+from pathlib import Path
+from typing import List, Optional
+
+import mkosi
+import pytest
+from mkosi.backend import Distribution, MkosiArgs, MkosiException
+
+
+def parse(argv: Optional[List[str]] = None) -> MkosiArgs:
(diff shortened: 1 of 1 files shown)


behrmann:
Haven't yet looked at the WIP stuff, but is there a reason to use `sudo` for unit tests? I think it much preferable, if the unit tests could be run without privileges.

gsegatti:
> Hi Behrmann, let me know if this makes sense:
Currently, I'm building an abstraction that allows building/booting and running commands inside a Mkosi generated image from a python class. In order to build/boot an image we will necessarily go through the check_root() function inside run_verb on __init__.py.
Therefore, without sudo die() will be invoked. If I simply remove the check_root() function, execution will still be stopped due to privileges in init_namespace() -> unshare().

If there's a smarter way of doing this, please, let me know.

gsegatti:
> I see. I guess that is #894? I'll have a look at it tomorrow. If it's #894 we're talking about it would be great if you could write a short description there to see where this is supposed to go.

Yes! I'll write it in the meantime.

behrmann:
Just read the rationale in #894 and for that it will be necessary to run those tests with elevated privileges, that much is true. Mark the integration tests (or all actually using mkosi or parts that n) as skip if EUID isn't 0 and I think that's alright.

gsegatti:
Good suggestion, I've added a skipif() to the pytest file.

PR #898, opened 2022-02-03
Change parse_args() arg type from List to Sequence

DaanDeMeyer:
We don't modify the passed args so we can use Sequence as the type
instead of List.


behrmann (approved):

DaanDeMeyer:
unit tests succeeded, no need to wait for the rest.

behrmann:
> unit tests succeeded, no need to wait for the rest.

was writing very same thing just this instance, you were a bit faster. :)

PR #899, opened 2022-02-03
ci: Add python integration tests job

DaanDeMeyer:
To run the actual integration tests, we need to have the mkosi
action and dependencies installed. Instead of running the integration
tests as part of the unit tests job, let's add a new job for these
integration tests where we install the necessary dependencies (including
mkosi itself) and then run all tests marked with "integration" in pytest.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
 
     - name: Install
       run: |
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
Eventually we'll need to expand this to include multiple distros and stuff but let's start simple with a single job that runs the configured tests.

DaanDeMeyer:
Going to land this one to unblock #894

PR #902, opened 2022-02-06
Add `--update` and `--deep` to invoke_emerge call for pkgs_boot

m0gg:
When generating bootable images for gentoo, the emerge call to install
pkgs_boot fails for `sys-kernel/installkernel-systemd-boot`:

```
[…]
‣   Invoking emerge(1) inside stage3
!!! /etc/portage/binrepos.conf is missing (or PORTAGE_BINHOST is unset), but use is requested.

 * IMPORTANT: 9 news items need reading for repository 'gentoo'.
 * Use eselect news read to view new items.

!!! /etc/portage/binrepos.conf is missing (or PORTAGE_BINHOST is unset), but use is requested.
[binary  N    ] app-arch/cpio-2.12-r1-1
[binary  N    ] sys-firmware/edk2-ovmf-202105-r2-1
[binary  N    ] dev-libs/elfutils-0.186-1
[binary  N    ] sys-kernel/installkernel-systemd-boot-1-1
[binary  N    ] sys-kernel/dracut-055-r4-2
[binary  N    ] virtual/libelf-3-1
[binary  N    ] sys-kernel/gentoo-kernel-bin-5.15.19-1
[binary  N    ] virtual/dist-kernel-5.15.19-1
[blocks B     ] sys-kernel/installkernel-gentoo ("sys-kernel/installkernel-gentoo" is soft blocking sys-kernel/installkernel-systemd-boot-1)
[blocks B     ] sys-kernel/installkernel-systemd-boot ("sys-kernel/installkernel-systemd-boot" is soft blocking sys-kernel/installkernel-gentoo-3)

 * Error: The above package list contains packages which cannot be
 * installed at the same time on the same system.

  (sys-kernel/installkernel-gentoo-3-1:0/0::gentoo, installed) pulled in by
    sys-kernel/installkernel-gentoo required by (sys-kernel/gentoo-kernel-bin-5.15.19-1:5.15.19/5.15.19::gentoo, binary scheduled for merge) USE="initramfs -test"

  (sys-kernel/installkernel-systemd-boot-1-1:0/0::gentoo, binary scheduled for merge) pulled inby
    sys-kernel/installkernel-systemd-boot

‣ Error: Workspace command /usr/bin/emerge sys-kernel/installkernel-systemd-boot sys-kernel/gentoo-kernel-bin sys-firmware/edk2-ovmf --buildpkg=y --usepkg=y --keep-going=y --jobs=16 --load-average=15 --nospinner --quiet-build --quiet returned non-zero exit code 1.
‣  (Unmounting Package Cache)
‣  (Unmounting image)
‣  (Detaching /dev/loop0)
```

Adding `--update` and `--deep` allows to unmerge
`sys-kernel/installkernel-gentoo` and use `sys-kernel/installkernel-systemd-boot` instead.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3122,7 +3122,11 @@ def install_gentoo(
         # Please don't move, needs to be called before installing dracut
         # dracut is part of gentoo_pkgs_boot
         configure_dracut(args, packages={"dracut"}, root=root)
-        gentoo.invoke_emerge(args, root, pkgs=gentoo.pkgs_boot)
+        # Installing sys-kernel/installkernel-systemd-boot fails spectacularly
+        # due to beeing soft blocked by sys-kernel/installkernel-gentoo (which
+        # is pulled in by debianutils) unless --update and --deep is also
+        # specified
+        gentoo.invoke_emerge(args, root, pkgs=gentoo.pkgs_boot, opts=["--update", "--deep"])
 
     if args.packages:
         gentoo.invoke_emerge(args, root, pkgs=args.packages)


DaanDeMeyer:
cc @257

mkosi/__init__.py
```diff
@@ -3122,7 +3122,11 @@ def install_gentoo(
         # Please don't move, needs to be called before installing dracut
         # dracut is part of gentoo_pkgs_boot
         configure_dracut(args, packages={"dracut"}, root=root)
-        gentoo.invoke_emerge(args, root, pkgs=gentoo.pkgs_boot)
+        # Installing sys-kernel/installkernel-systemd-boot fails spectacularly
```
behrmann, line 3125: "spectacularly" is a bit harsh, don't you think? ;)

mkosi/__init__.py
```diff
@@ -3122,7 +3122,11 @@ def install_gentoo(
...
+        # due to beeing soft blocked by sys-kernel/installkernel-gentoo (which
+        # is pulled in by debianutils) unless --update and --deep is also
```
behrmann, line 3127: From the PR description I understand that there's two competing `sys-kernel/install-foo`. I don't really speak Gentoo, so I'm not sure what "unmerge" means, but I guess something like "uninstall" or "removing from the transaction" (please correct me if I'm wrong). Great if `--update` and `--deep` fix this! Could you maybe make the comment slightly longer why that is or link to the appropriate manpage, so that us non-Gentoo people have an easier time to follow the reasoning?

behrmann (commented):
Thanks for the contribution! Great to see more pickup in Gentoo.

257:
thanks for the patch.

first impression i have:

- why `update_stage3()` call, which indeed calls `emerge` with `--update` `--deep` flags, see `EMERGE_UPDATE_OPTS`) didn't pick this up? do we apply our `USE` flags after calling `update_stage3()`? our `USE` flags did use to trigger unmerge of `sys-kernel/installkernel-gentoo`.

- what do you think of unmerging `sys-kernel/installkernel-gentoo` inside `fetch_fix_stage3()`? [we do unmerge `sys-apps/baselayout` there](https://github.com/systemd/mkosi/blob/07863c38d07607bcb7252e8de7884dcde5ddf87b/mkosi/gentoo.py#L326), i.e. we take out `baselayout` before even copying stage3 tree to the `root` of the image.

m0gg:
> m0gg closed this

Wow... I'm sorry... Big thumbs on a small screen...

> our `USE` flags did use to trigger unmerge of `sys-kernel/installkernel-gentoo`.

Are you sure about that? Every ebuild with a dep on either `-gentoo` or `-systemd-boot` (except for each other) i was able to find, defined its dep as:
```
|| ( sys-kernel/installkernel-gentoo sys-kernel/installkernel-systemd-boot )
```
While this smells like a case for a new virtual, i don't think we're currently able to automagicaly bring portage to emerge '-systemd-boot' over '-gentoo' (which is baked into the stage3) by using anything other than `package.mask` or explicitly un-/merging one of them.

Since passing `--update` and `--deep` should not issue actual "updates" (`[U]`) of any selected packages or deps in this case, this should not be safe.

I tried your unmerge-in-`fetch_fix_stage3()` suggestion, but the following depclean fails if you take away debianutils' `PDEPEND`. It works as intended if it's unmerged after the deplcean though.  
`--unmerging` is sometimes the shortest (or only) solution, but almost never without side-effects (at least from my experience). So i'd stick with explicitly merging instead of explicitly unmerging.

mkosi/__init__.py
m0gg: Eeh... I may have been a little annoyed because building a gentoo osi takes some time and it always barfed on me in the last steps 😅

mkosi/__init__.py
m0gg: Will do.

257:
> > m0gg closed this
> 
> Wow... I'm sorry... Big thumbs on a small screen...
> 
> > our `USE` flags did use to trigger unmerge of `sys-kernel/installkernel-gentoo`.
> 
> Are you sure about that?

yep, otherwise it wouldn't pass the tests.

> Every ebuild with a dep on either `-gentoo` or `-systemd-boot` (except for each other) i was able to find, defined its dep as:
> 
> ```
> || ( sys-kernel/installkernel-gentoo sys-kernel/installkernel-systemd-boot )
> ```

i see. well something might have changed upstream in one of those ebuilds. your solution should make us immune to those changes in the future, or at least that's the hope.
 
> Since passing `--update` and `--deep` should not issue actual "updates" (`[U]`) of any selected packages or deps in this case, this should not be safe.

it should be safe, i concur :)

> I tried your unmerge-in-`fetch_fix_stage3()` suggestion, but the following depclean fails if you take away debianutils' `PDEPEND`.

wait, what? `debianutils`? where do you see that?

> It works as intended if it's unmerged after the deplcean though. `--unmerging` is sometimes the shortest (or only) solution, but almost never without side-effects (at least from my experience). So i'd stick with explicitly merging instead of explicitly unmerging.

that's a good point.

on last thing, what do thing of passing `opts=gentoo.EMERGE_UPDATE_OPTS` instead? there are extra flags there but they should be safe.

@DaanDeMeyer this looks tip-top and good-to-go to me, with or without that^.

257:
> While this smells like a case for a new virtual, i don't think we're currently able to automagicaly bring portage to emerge `-systemd-boot` over `-gentoo` (which is baked into the stage3) by using anything other than `package.mask` or explicitly un-/merging one of them.

i will give a shout out to @mgorny about this, hopefully he will help us out here.

m0gg:
> yep, otherwise it wouldn't pass the tests.

I tried to find older CI results but i'm not familiar with github actions, I'm more a gitlab-guy. If you say so, I guess `sys-kernel/installkernel-gentoo` has not always been in the stage3.

> > this should not be safe.
>
> it should be safe, i concur :)

Oops. I guess you got what i really wanted to express :)

> wait, what? debianutils? where do you see that?

`sys-apps/debianutils` is included in the stage3. Pulled in by (at least) [`app-misc/ca-certificates`](https://github.com/gentoo/gentoo/blob/7756de4318427c9b5810f62730b1bef9f87ea9e4/app-misc/ca-certificates/ca-certificates-20211016.3.72.ebuild#L60).
My first thought was, that maybe `sys-apps/debianutils` is the [culprit for pulling in `sys-kernel/installkernel-gentoo`](https://github.com/gentoo/gentoo/blob/7756de4318427c9b5810f62730b1bef9f87ea9e4/sys-apps/debianutils/debianutils-5.5.ebuild#L15), but apparently it's been this way for quite some time ([PV 4.11.2](https://github.com/gentoo/gentoo/blob/eb4e911aacdeed47cab0f6993c32ae63f5f35b89/sys-apps/debianutils/debianutils-4.11.2.ebuild#L15) - > 10 months ago)

> on last thing, what do thing of passing opts=gentoo.EMERGE_UPDATE_OPTS instead?

Yep. This should keep maintenance lower aswell.

> i will give a shout out to @mgorny about this, hopefully he will help us out here.

If it's @mgorny, i'm positive about it ;)

257:
> > on last thing, what do thing of passing opts=gentoo.EMERGE_UPDATE_OPTS instead?
> 
> Yep. This should keep maintenance lower aswell.

my thoughts exactly.
@mgorny (via #gentoo-kernel) also suggests adding `--complete-graph` which is already in `EMERGE_UPDATE_OPTS`.

btw, reading `man 1 emerge` on `--complete-graph` makes me think `--deep` is superfluous (in `EMERGE_UPDATE_OPTS`), any thoughts?

257:
> btw, reading `man 1 emerge` on `--complete-graph` makes me think `--deep` is superfluous (in `EMERGE_UPDATE_OPTS`), any thoughts?

correction: we pass `--complete-graph-if-new-use=y`, hmm!

257:
@mgorny thinks "_99% a bug in `portage`_" and is considering a virtual for it, as a work around i guess.

@DaanDeMeyer could we hold this off for now?

257:
[emerge refuses to replace sys-kernel/installkernel-gentoo with sys-kernel/installkernel-systemd-boot if sys-kernel/gentoo-kernel-bin is also on cmdline](https://bugs.gentoo.org/show_bug.cgi?id=833014)

mgorny:
The "easy workaround" is to run a separate emerge to install installkernel-systemd-boot first — as long as Portage doesn't get other packages into the depgraph, it seems to be able to figure out what to do.

The virtual doesn't seem to change anything. Even if it's installed before and I'm running emerge without `--deep` (so theoretically it shouldn't even go beyond the virtual since it's installed already), it fails on the blockers (being the deps of virtual now).

One way I was able to resolve this was adding a virtual with explicit USE flags, and then using `--deep --changed-use`. This works because USE flags make it clear that only the desired solution is possible. However, this may be PITA for users who will now to have to adjust the USE flags or otherwise emerge will quickly replace their desired installkernel with whatever default we decided on. So I don't think this is the way to go.

257:
[we have a bisect](https://bugs.gentoo.org/833014#c2)

257:
@DaanDeMeyer fix for portage won't be coming anytime soon apparently. @m0gg's solution is *defensive* and i think we should go ahead with it.

PR #903, opened 2022-02-11
qemu: Add sanity check for bootable

alexgartrell:
KVM throws a pretty confusing "internal error" if you try to boot an
arbitrary drive, so it's better to just alert the user ahead of time.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -7535,6 +7535,9 @@ def find_ovmf_vars() -> Path:
 
 
 def run_qemu(args: MkosiArgs) -> None:
+    if not args.bootable:
+        warn("This image may not be not bootable. Set Bootable=yes in [Output]")
+
     has_kvm = os.path.exists("/dev/kvm")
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
So technically this breaks backwards compat because when specifying options using the CLI, you only have to specify --bootable when building the image and not anymore when running the image using qemu.

I know @keszybz mentioned he wanted to add the config used to build the image to the generated JSON manifest for the image. Once we have that in place, we could instead program this check by checking if the Bootable option is set in the manifest.

But since that will take a while to manifest, it would be good to at least improve the messaging until we can get the proper solution in place. So the only thing I'd change is to downgrade this to a warning (replace die() with warn()) so we don't accidentally end up breaking any users.

DaanDeMeyer:
We merged #904 so I'm going to close this one but thanks for bringing up the issue. I've definitely been bitten a few times before trying to boot an image in qemu built without the --bootable option so this was definitely something that we needed to improve upon.

alexgartrell:
No stress. Thanks for the fix!

PR #904, opened 2022-02-11
Error when trying to build an image without --bootable and run it with qemu

DaanDeMeyer:
We can only check this when we're building the image, since when we're not
building the image it's not required to pass the --bootable option in order
to be able to use the qemu verb to run it in qemu.

This also fixes needs_build() to take the args.force option into account.
This fixes usage of needs_build() before unlink_output() is called.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -6488,6 +6488,9 @@ def load_args(args: argparse.Namespace) -> MkosiArgs:
         if not args.output_format.is_disk():
             die("Sorry, can't boot non-disk images with qemu.")
 
+    if needs_build(args) and args.verb == "qemu" and not args.bootable:
+        die("Images built without the --bootable option cannot be booted using qemu")
+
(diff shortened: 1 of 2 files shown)


DaanDeMeyer:
Replaces #903

behrmann (approved):
Looks sensible to me.

mkosi/__init__.py
```diff
@@ -6488,6 +6488,9 @@ def load_args(args: argparse.Namespace) -> MkosiArgs:
         if not args.output_format.is_disk():
             die("Sorry, can't boot non-disk images with qemu.")
 
+    if needs_build(args) and args.verb == "qemu" and not args.bootable:
```
behrmann, line 6491: This is the change breaking the unit tests

behrmann: The failing line in `test_parse_load_args.py` can be fixed like this
```python
    assert parse(["--bootable", "qemu"]).verb == "qemu"
```

PR #905, opened 2022-02-14
Fix needs_build() logic

daandemeyer:
We should only check the -f option if the verb is one of the verbs
that needs a build. With the current check, we start rebuilding an
image even if the verb is "summary".


behrmann:
ah, the vagaries of how strong how strong which logic operator binds.

behrmann (approved):

PR #906, opened 2022-02-15
Install a more specific match when --network-veth is used

daandemeyer:
When doing more complicated network stuff in a QEMU VM, the network
file we install starts matching stuff it shouldn't match. Let's narrow
the match down by driver so we don't accidentally match against stuff
we don't want to match against.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,7 +3,7 @@
 ## v13
 
 - The networkd config file installed by mkosi when the --network-veth option is
-  used now only matches against network interfaces using the virtio_net driver.
+  used (/etc/systemd/network/80-mkosi-network-veth.network in the image) now only matches against network interfaces using the virtio_net driver.
   Please make sure you weren't relying on this file to configure any network
   interfaces other than the tun/tap virtio-net interface created by mkosi when
   booting the image in QEMU with the --network-veth option. If you were relying


DaanDeMeyer:
@behrmann Can you take a quick look at this one as well? I verified that with both TCG and KVM the driver is reported as virtio_net. The driver is also reported as virtio_net when we used qemu user mode networking instead of TUN/TAP.

behrmann:
If you verified that it works in those scenarios, it's probably ok, but I don't really have a good feel for the matrix of possibilities here, so I can't really say whether there might be problems lurking somewhere.

One possibility of working around this, could be to install a separate network file for qemu, and first match `Virtualization=qemu` explicitly and only there do the `Driver=virtio_net` match and leave the original as is. That being said, I'm not a networking expert, so that might be unnecessarily safety-minded. 

Although I've never seen him here, maybe @yuwata has an opinion on this.

mkosi/__init__.py
```diff
@@ -6923,6 +6923,7 @@ def setup_network_veth(args: MkosiArgs, root: Path, do_run_build_script: bool, c
                     [Match]
                     Virtualization=!container
                     Type=ether
+                    Driver=virtio_net
```
yuwata, line 6926: If this is really for veth, then this should be "Driver=veth", no?
When I create veth, then networkd detects its driver as "veth".
```
systemd-networkd[1335053]: veth99: Saved new link: ifindex=240, iftype=ETHER(1), kind=veth
systemd-networkd[1335053]: veth99: Saved permanent hardware address: 00:00:00:00:00:00
systemd-networkd[1335053]: veth99: Found driver: veth
systemd-networkd[1335053]: veth99: Saved original MTU 1500 (min: 68, max: 65535)
systemd-networkd[1335053]: veth99: Saved hardware address: 8e:e0:63:d0:aa:70
systemd-networkd[1335053]: veth99: Saved IPv6 link-local address generation mode: eui64
systemd-networkd[1335053]: veth99: Flags change: +MULTICAST +BROADCAST
```
If this is correct, then maybe adjust the function name 'setup_network_veth'. Or please add a comment about this.

DaanDeMeyer:
We try to mimick https://github.com/systemd/systemd/blob/v247/network/80-container-host0.network from systemd here. That config file has it easy since it can just match directly against the interface name which is guaranteed to be host0 when --network-veth is used with nspawn. 

When using qemu, we don't have any direct control over the interface name in the guest so we can't match against the name. We do have control over the mac address so what we could do is assign a fixed mac address to the tun/tap guest interface and match against that. This would mean that each instance of the image would be booted with the same mac address which doesn't seem ideal either because mac addresses are supposed to be unique.

I'm not sure what else we can match against except for the driver used.

yuwata:
I do not say using `Driver=` is bad. I am just confused with the inconsistency of the driver name and the function name.

DaanDeMeyer:
Yeah the naming is rather confusing. I duplicated the naming from nspawn when I added this option which also uses --network-veth but obviously it doesn't apply to qemu VMs using tun/tap at all. I'll do a PR to rename things to something more sensible when I can.

DaanDeMeyer:
> One possibility of working around this, could be to install a separate network file for qemu, and first match Virtualization=qemu explicitly and only there do the Driver=virtio_net match and leave the original as is. That being said, I'm not a networking expert, so that might be unnecessarily safety-minded.

I guess technically this change could break users that depend on this file to configure their network interface. The original intent of this config file was to just configure the tun/tap network interface that we add when running under qemu and the --network-veth option is used. But I added the original file with a too wide match so it might be used for more than just configuring the qemu tun/tap interface now.

Should we consider this public API? The documentation of the --network-veth option doesn't mention this file at all. IMO we should be allowed to modify it at will as long as we keep having a network connection when booting with nspawn/qemu and using the --network-veth option.

behrmann:
I guess as long as we have network people won't care about how this is achieved. I agree that how our networkd config does it, is somewhat of an implementation detail. Let's document the change for the next release and if it breaks people who were for some reason relying on on the how, we can still find another solution.

DaanDeMeyer:
I added a note to NEWS about the potential breakage

NEWS.md
```diff
@@ -1,5 +1,16 @@
 # mkosi Changelog
 
+## v13
+
+- The networkd config file installed by mkosi when the --network-veth option is
+  used now only matches against network interfaces using the virtio_net driver.
```
behrmann, line 6:
```suggestion
- The networkd config file installed by mkosi when the --network-veth option is
  used (/etc/systemd/network/80-mkosi-network-veth.network in the image) now only matches against network interfaces using the virtio_net driver.
```

NEWS.md
```diff
@@ -1,5 +1,16 @@
...
+  Please make sure you weren't relying on this file to configure any network
+  interfaces other than the tun/tap virtio-net interface created by mkosi when
+  booting the image in QEMU with the --network-veth option. If you were relying
+  on this config file to configure other interfaces, you'll have to re-create it
+  with the correct match to make sure networkd will keep configuring your
```
behrmann, line 11:
```suggestion
  with the correct match and a lower initial number in the filename to make sure networkd will keep configuring your
```

behrmann (commented):
Having answered a few questions on this lately, some ideas to hopefully make this more helpful to users.

DaanDeMeyer:
Let's merge this one, I'll address the naming issue in a second PR.

PR #907, opened 2022-02-16
QEMU fixes

DaanDeMeyer:
A few fixes to address the stuff @mrc0mmand figured out in https://github.com/systemd/systemd/pull/22417#issuecomment-1033660354

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -5574,7 +5574,7 @@ def create_parser() -> ArgumentParserMkosi:
         "--extra-search-paths", dest="extra_search_paths", action=ColonDelimitedListAction, help=argparse.SUPPRESS
     )  # Compatibility option
     group.add_argument("--qemu-headless", action=BooleanAction, help="Configure image for qemu's -nographic mode")
-    group.add_argument("--qemu-smp", help="Configure guest's SMP settings", metavar="SMP", default="2")
+    group.add_argument("--qemu-smp", help="Configure guest's SMP settings", metavar="SMP", default="1")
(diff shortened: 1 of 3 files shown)


mkosi/__init__.py
```diff
@@ -5574,7 +5574,7 @@ def create_parser() -> ArgumentParserMkosi:
         "--extra-search-paths", dest="extra_search_paths", action=ColonDelimitedListAction, help=argparse.SUPPRESS
     )  # Compatibility option
     group.add_argument("--qemu-headless", action=BooleanAction, help="Configure image for qemu's -nographic mode")
-    group.add_argument("--qemu-smp", help="Configure guest's SMP settings", metavar="SMP", default="2")
+    group.add_argument("--qemu-smp", help="Configure guest's SMP settings", metavar="SMP", default="1")
```
behrmann, line 5577: Needs to be adjusted in `test_config_parser.py:129` as well

behrmann:
Disregarding the test failure (due to the missing change in `test_config_parser.py`), LGTM

DaanDeMeyer:
So centos_epel doesn't hang but it still fails on trying to connect with SSH. However, changing these options seems to break Arch/Fedora instead. I can't reproduce the failures locally though so I'm guessing there's a bug in qemu 4.2 somewhere which is what focal provides.

The next Ubuntu LTS release is coming out in April so we might want to simply keep the status quo for now and revisit this once we can move to 22.04 in GA since we'll have access to a more recent version of qemu which hopefully won't have the bug we're seeing here anymore.

DaanDeMeyer:
Merging into #974

PR #909, opened 2022-02-18
Rename --network-veth to --netdev

DaanDeMeyer:
--network-veth implies usage of veth virtual ethernet links which we
use for containers but not for VMs. Let's rename the option to --netdev
instead which encapsulates both veth links and tun/tap links which we
use for QEMU VMs.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,11 +2,11 @@
 
 ## v13
 
-- The networkd config file installed by mkosi when the --network-veth option is
-  used (/etc/systemd/network/80-mkosi-network-veth.network in the image) now only matches against network interfaces using the virtio_net driver.
+- The networkd config file installed by mkosi when the --netdev option is
+  used (/etc/systemd/network/80-mkosi-netdev.network in the image) now only matches against network interfaces using the virtio_net driver.
   Please make sure you weren't relying on this file to configure any network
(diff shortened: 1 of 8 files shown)


NEWS.md
```diff
@@ -2,11 +2,11 @@
 
 ## v13
 
-- The networkd config file installed by mkosi when the --network-veth option is
-  used (/etc/systemd/network/80-mkosi-network-veth.network in the image) now only matches against network interfaces using the virtio_net driver.
+- The networkd config file installed by mkosi when the --netdev option is
+  used (/etc/systemd/network/80-mkosi-netdev.network in the image) now only matches against network interfaces using the virtio_net driver.
```
behrmann, line 6:
```suggestion
- The `--network-veth` option has been renamed `--netdev`, since it implies usage of a virtual ethernet device, which is not what is done when booting images with qemu where a TUN/TAP device is used.
- The networkd config file installed by mkosi when the `--netdev`  option (what used to be `--network-veth`) is
  used (formerly `/etc/systemd/network/80-mkosi-network-veth.network` in the image) now only matches against network interfaces using the virtio_net driver.
```

behrmann: Users reading this might be confused by `--netdev` which they haven't seen before.

Issue #891, opened 2022-01-21
Ubuntu 20.04 build issues


brpatel12:
Hi all - couple of questions. I'm trying to build a base Ubuntu image before attempting to migrate a larger build process to entirely generate images using mkosi. I assume I'm making a mistake somewhere and am hoping for some assistance.
- Ubuntu 20.04 LTS
- NVIDIA GPU

# Configs
build produces an image with kernel `5.4.0-26-generic`

## Version
```bash
brpatel@bergie:~/mkosi$ mkosi --version
mkosi 12
```

## mkosi.default
```bash
[Distribution]
Distribution=ubuntu
Release=focal
Repositories=focal,main,restricted,universe
Architecture=x86_64

[Output]
Format=gpt_ext4
ManifestFormat=json
OutputDirectory=mkosi.output
ImageId=test
ImageVersion=1
SplitArtifacts=no
Bootable=yes
BootProtocols=uefi
SecureBoot=no
ReadOnly=no
Minimize=no

[Content]
WithNetwork=yes
Cache=mkosi.cache
BasePackages=yes

Packages=
  sudo
  vim
  iputils-ping
  less
  lsof

  # syslog
  #rsyslog
  inetutils-syslogd

CleanPackageMetadata=auto
Password=test
ExtraTree=mkosi.extra
PostInstallationScript=mkosi_postinst.sh

[Partitions]
RootSize=5G
ESPSize=1G
```
## mkosi.extra
```bash
brpatel@bergie:~/mkosi$ tree mkosi.extra/
mkosi.extra/
└── etc
    └── apt
        └── sources.list
```
## /etc/apt/sources.list
```bash
deb http://us.archive.ubuntu.com/ubuntu/ focal main restricted universe multiverse
deb http://us.archive.ubuntu.com/ubuntu/ focal-updates main restricted universe multiverse
deb http://us.archive.ubuntu.com/ubuntu/ focal-backports main restricted universe multiverse
deb http://security.ubuntu.com/ubuntu focal-security main restricted universe multiverse
```
## mkosi_postinst.sh
```bash
#!/bin/bash
set -euo pipefail

apt-get -y update
apt-get -y upgrade
apt-get install -y --no-install-recommends nvidia-driver-470 
apt-get -y update
```
# Issues
when I boot with qemu or the target hardware I run into the same issues - appreciate any suggestions

## systemd services
[...]

PR #911, opened 2022-02-21
Only copystat() on newly created directories in copy_path()

DaanDeMeyer:
Fixes unexpected scenarios where we modify the permissions of / when
using mkosi.extra/. See https://github.com/systemd/systemd/pull/22569#issuecomment-1045992142
for more information.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -628,7 +628,7 @@ def symlink_f(target: str, path: Path) -> None:
         path.symlink_to(target)
 
 
-def copy_path(oldpath: PathString, newpath: Path) -> None:
+def copy_path(oldpath: PathString, newpath: Path, *, copystat: bool = True) -> None:
     try:
         newpath.mkdir(exist_ok=True)
     except FileExistsError:
@@ -651,7 +651,9 @@ def copy_path(oldpath: PathString, newpath: Path) -> None:
             else:
                 print("Ignoring", entry.path)
                 continue
-    shutil.copystat(oldpath, newpath, follow_symlinks=True)
+
+    if copystat:
+        shutil.copystat(oldpath, newpath, follow_symlinks=True)
 
 
 @complete_step("Detaching namespace")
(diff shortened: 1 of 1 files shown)


behrmann:
I guess that's the same people also already stumbled over in #891.

This should fix the issue, but I think down the line we might run into different ones with this change. Right now `copy_path` clobbers the permissions on `/`, but wouldn't this change keep the permissions of something even below `/` the same, if it already exists? I'm not sure, but I can imagine people using this to change permissions via `mkosi.skeleton` or `mkosi.extra_trees`.

What about the alternative of changing the signature of `copy_path` like this:
```diff
-def copy_path(oldpath: PathString, newpath: Path) -> None:
+def copy_path(oldpath: PathString, newpath: Path, *, copystat: bool = True) -> None:
```
such that we conditionalise the call to `shutil.copystat` at the end on this. This way we keep the current behaviour, but could switch it off for the root of extra trees (namely then in `install_extra_trees` and `install_skeleton_tree`).

DaanDeMeyer:
I wonder if it makes sense at all though to set permissions on existing directories by copying them from mkosi.extra/ or similar directories. For most use cases, I suspect it won't be the user's intention to change that stuff and they're only re-creating the required directory structure to make sure their file ends up in the right location in the image.

It seems safer and more readable to have users explicitly configure permissions where needed instead of implicitly relying on permissions from directories like in an extra tree which are copied over.

Yeah this might break backwards compat if anyone is explicitly relying on this, but it seems like a small breakage overall.

Of course this assumes this is the correct approach at all. cc @keszybz who added the original implementation of copy_path().

keszybz:
Sorry for the late response.

I think we want to adjust permissions. In particular, *not* copying modes deep in the tree e.g. somewhere in `reuse_cache_tree()` seems wrongs. The approach proposed by @behrmann above looks much better.

DaanDeMeyer:
Updated to use @behrmann's approach. PTAL

behrmann (approved):
LGTM.

Something broke the Debian `gpt_ext4` CI (not this, I already saw it failing somewhere else today)
```
 ‣  Installing boot loader…
execv(bootctl) failed: No such file or directory
‣ Error: Workspace command bootctl install returned non-zero exit code 1.
```

DaanDeMeyer:
> Something broke the Debian gpt_ext4 CI (not this, I already saw it failing somewhere else today)

systemd-boot just became a split package in Debian, I'm working on a fix.

behrmann:
Merging, since CI is happy (except for the one case already fixed on `main`)

PR #913, opened 2022-02-22
Adding an Enum class for verb comparison.

gsegatti:
As suggested here: [894](https://github.com/systemd/mkosi/pull/894#discussion_r803610296).
The idea is: we parse CLI-entered verbs into the Enum class to allow comparisons to happen more easily.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -77,6 +77,7 @@ from .backend import (
     PartitionIdentifier,
     PartitionTable,
     SourceFileTransfer,
+    Verb,
     die,
     install_grub,
     nspawn_params_for_blockdev_access,
@@ -115,10 +116,9 @@ else:
 SomeIO = Union[BinaryIO, TextIO]
 PathString = Union[Path, str]
 
-MKOSI_COMMANDS_CMDLINE = ("build", "shell", "boot", "qemu", "ssh")
-MKOSI_COMMANDS_NEED_BUILD = ("shell", "boot", "qemu", "serve")
-MKOSI_COMMANDS_SUDO = ("build", "clean", "shell", "boot", "qemu", "serve")
-MKOSI_COMMANDS = ("build", "clean", "help", "summary", "genkey", "bump", "serve") + MKOSI_COMMANDS_CMDLINE
+MKOSI_COMMANDS_NEED_BUILD = (Verb.shell, Verb.boot, Verb.qemu, Verb.serve)
+MKOSI_COMMANDS_SUDO = (Verb.build, Verb.clean, Verb.shell, Verb.boot, Verb.qemu, Verb.serve)
+MKOSI_COMMANDS_CMDLINE = (Verb.build, Verb.shell, Verb.boot, Verb.qemu, Verb.ssh)
 
 DRACUT_SYSTEMD_EXTRAS = [
     "/usr/bin/systemd-repart",
@@ -5083,9 +5083,9 @@ def create_parser() -> ArgumentParserMkosi:
     parser = ArgumentParserMkosi(prog="mkosi", description="Build Bespoke OS Images", add_help=False)
 
     group = parser.add_argument_group("Commands")
-    group.add_argument("verb", choices=MKOSI_COMMANDS, default="build", help="Operation to execute")
+    group.add_argument("verb", type=Verb, choices=list(Verb), default=Verb.build, help="Operation to execute")
     group.add_argument(
-        "cmdline", nargs=argparse.REMAINDER, help="The command line to use for " + str(MKOSI_COMMANDS_CMDLINE)[1:-1]
+        "cmdline", nargs=argparse.REMAINDER, help="The command line to use for " + str([verb.name for verb in MKOSI_COMMANDS_CMDLINE])[1:-1]
     )
     group.add_argument("-h", "--help", action="help", help="Show this help")
     group.add_argument("--version", action="version", version="%(prog)s " + __version__)
@@ -5703,9 +5703,9 @@ def parse_args(argv: Optional[Sequence[str]] = None) -> Dict[str, argparse.Names
     # values of positional arguments. Make sure the verb command gets explicitly passed.
     # Insert a -- before the positional verb argument otherwise it might be considered as an argument of
     # a parameter with nargs='?'. For example mkosi -i summary would be treated as -i=summary.
-    for verb in MKOSI_COMMANDS:
+    for verb in Verb:
         try:
-            v_i = argv.index(verb)
+            v_i = argv.index(verb.name)
         except ValueError:
             continue
 
@@ -5718,7 +5718,7 @@ def parse_args(argv: Optional[Sequence[str]] = None) -> Dict[str, argparse.Names
     # First run of command line arguments parsing to get the directory of mkosi.default file and the verb argument.
     args_pre_parsed, _ = parser.parse_known_args(argv)
 
-    if args_pre_parsed.verb == "help":
+    if args_pre_parsed.verb == Verb.help:
         parser.print_help()
         sys.exit(0)
 
@@ -5902,7 +5902,7 @@ def empty_directory(path: Path) -> None:
 
 
 def unlink_output(args: MkosiArgs) -> None:
-    if not args.force and args.verb != "clean":
+    if not args.force and args.verb != Verb.clean:
         return
 
     if not args.skip_final_phase:
@@ -5942,7 +5942,7 @@ def unlink_output(args: MkosiArgs) -> None:
     # remove the downloaded package cache if the user specified one
     # additional "--force".
 
-    if args.verb == "clean":
+    if args.verb == Verb.clean:
         remove_build_cache = args.force > 0
         remove_package_cache = args.force > 1
     else:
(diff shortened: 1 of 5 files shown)


mkosi/backend.py
```diff
@@ -225,17 +243,17 @@ class ManifestFormat(Parseable, enum.Enum):
...
-    xbootldr   = 'xbootldr'
-    root       = 'root'
-    swap       = 'swap'
-    home       = 'home'
-    srv        = 'srv'
-    var        = 'var'
-    tmp        = 'tmp'
-    verity     = 'verity'
-    verity_sig = 'verity-sig'
+    esp        = "esp"
```
gsegatti, line 246: Just bumped into this usage of single quotes and thought I'd quickly fix it myself

mkosi/backend.py
```diff
@@ -225,17 +243,17 @@ class ManifestFormat(Parseable, enum.Enum):
...
+    bios       = "bios"
+    xbootldr   = "xbootldr"
+    root       = "root"
+    swap       = "swap"
+    home       = "home"
+    srv        = "srv"
+    var        = "var"
+    tmp        = "tmp"
+    verity     = "verity"
+    verity_sig = "verity-sig"
```
gsegatti, line 256: Bumped into this usage of single quotes and as this has been a topic in quite a few of my PR's, I thought it couldn't hurt fixing it quickly

mkosi/backend.py
```diff
@@ -112,6 +112,24 @@ class PackageType(enum.Enum):
...
+    clean   = "clean"
+    summary = "summary"
+    shell   = "shell"
+    boot    = "boot"
+    qemu    = "qemu"
+    ssh     = "ssh"
+    serve   = "serve"
+    bump    = "bump"
+    help    = "help"
+    genkey  = "genkey"
```
gsegatti, line 126: I used tabbing to separate the verbs and '=' since this was done as well into other Enum classes with strings, such as `PartitionIdentifier` and `ManifestFormat`

gsegatti:
Forgot a few testing files behind, will edit it as well

mkosi/__init__.py
```diff
@@ -7877,28 +7878,28 @@ def expand_specifier(s: str) -> str:
...
-    if args.verb == "genkey":
+    if args.verb == Verb.genkey:
         generate_secure_boot_key(args)
 
-    if args.verb == "bump":
+    if args.verb == Verb.bump:
         bump_image_version(args)
 
-    if args.verb in MKOSI_COMMANDS_SUDO:
+    if args.verb.name in MKOSI_COMMANDS_SUDO:
```
DaanDeMeyer, line 7895: Quick hint that you should go the other way there, don't check the string name of the enum is in the list. Instead, change the list to contain enum values instead of strings

gsegatti: Hm, that makes sense, if we're getting rid of string comparisons, checking if the string name of the enum is in the list is contradicting exactly what we're trying to do.

mkosi/__init__.py
```diff
@@ -116,8 +117,8 @@
 PathString = Union[Path, str]
 
 MKOSI_COMMANDS_CMDLINE = ("build", "shell", "boot", "qemu", "ssh")
-MKOSI_COMMANDS_NEED_BUILD = ("shell", "boot", "qemu", "serve")
-MKOSI_COMMANDS_SUDO = ("build", "clean", "shell", "boot", "qemu", "serve")
+MKOSI_COMMANDS_NEED_BUILD = (Verb.shell, Verb.boot, Verb.qemu, Verb.serve)
+MKOSI_COMMANDS_SUDO = (Verb.build, Verb.clean, Verb.shell, Verb.boot, Verb.qemu, Verb.serve)
 MKOSI_COMMANDS = ("build", "clean", "help", "summary", "genkey", "bump", "serve") + MKOSI_COMMANDS_CMDLINE
```
gsegatti, line 122: Will change the order so that `MKOSI_COMMANDS_CMDLINE` and `MKOSI_COMMANDS` are not separate, since it would look more homogeneous

gsegatti, line 122: Will change the order so that MKOSI_COMMANDS_CMDLINE and MKOSI_COMMANDS are not separate, since it would look more homogeneous

mkosi/__init__.py
```diff
@@ -115,9 +116,9 @@
 SomeIO = Union[BinaryIO, TextIO]
 PathString = Union[Path, str]
 
+MKOSI_COMMANDS_NEED_BUILD = (Verb.shell, Verb.boot, Verb.qemu, Verb.serve)
+MKOSI_COMMANDS_SUDO = (Verb.build, Verb.clean, Verb.shell, Verb.boot, Verb.qemu, Verb.serve)
 MKOSI_COMMANDS_CMDLINE = ("build", "shell", "boot", "qemu", "ssh")
```
DaanDeMeyer, line 121: Why not change this list to contain enum values instead of strings? I found a `args.verb not in MKOSI_COMMANDS_CMDLINE` check which will break unless this list is changed to contain enum values.

mkosi/__init__.py
```diff
@@ -5083,7 +5084,7 @@ def create_parser() -> ArgumentParserMkosi:
     parser = ArgumentParserMkosi(prog="mkosi", description="Build Bespoke OS Images", add_help=False)
 
     group = parser.add_argument_group("Commands")
-    group.add_argument("verb", choices=MKOSI_COMMANDS, default="build", help="Operation to execute")
+    group.add_argument("verb", type=Verb, choices=list(Verb), default="build", help="Operation to execute")
```
DaanDeMeyer, line 5087: We can change the default to `Verb.build` here.

mkosi/__init__.py
```diff
@@ -115,9 +116,9 @@
 SomeIO = Union[BinaryIO, TextIO]
 PathString = Union[Path, str]
 
+MKOSI_COMMANDS_NEED_BUILD = (Verb.shell, Verb.boot, Verb.qemu, Verb.serve)
+MKOSI_COMMANDS_SUDO = (Verb.build, Verb.clean, Verb.shell, Verb.boot, Verb.qemu, Verb.serve)
 MKOSI_COMMANDS_CMDLINE = ("build", "shell", "boot", "qemu", "ssh")
-MKOSI_COMMANDS_NEED_BUILD = ("shell", "boot", "qemu", "serve")
-MKOSI_COMMANDS_SUDO = ("build", "clean", "shell", "boot", "qemu", "serve")
 MKOSI_COMMANDS = ("build", "clean", "help", "summary", "genkey", "bump", "serve") + MKOSI_COMMANDS_CMDLINE
```
DaanDeMeyer, line 122: `MKOSI_COMMANDS` is only used once in parse_args(), you can change that usage to `for verb in Verb:` and get rid of `MKOSI_COMMANDS`.

mkosi/__init__.py
```diff
@@ -6643,7 +6644,7 @@ def line_join_list(array: Sequence[PathString]) -> str:
 def print_summary(args: MkosiArgs) -> None:
     # FIXME: normal print
     MkosiPrinter.info("COMMANDS:")
-    MkosiPrinter.info("                      verb: " + args.verb)
+    MkosiPrinter.info("                      verb: " + args.verb.name)
```
behrmann, line 6647: The reason you have to use `+` here, is because `__str__` is defined for `Verb`, but `__add__` is not.
```suggestion
    MkosiPrinter.info(f"                      verb: {args.verb})
```
should call `__str__` instead of `__add__` and we don't have to reach into the enum to pull out name or value.

DaanDeMeyer (changes requested):
You forgot one usage of "build" in load_args() as well.

mkosi/__init__.py
gsegatti: Good catch

mkosi/__init__.py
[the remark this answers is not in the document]
gsegatti: I actually let that slip through, thx for pointing it out!

mkosi/__init__.py
```diff
@@ -5703,9 +5703,9 @@ def parse_args(argv: Optional[Sequence[str]] = None) -> Dict[str, argparse.Names
     # values of positional arguments. Make sure the verb command gets explicitly passed.
     # Insert a -- before the positional verb argument otherwise it might be considered as an argument of
     # a parameter with nargs='?'. For example mkosi -i summary would be treated as -i=summary.
-    for verb in MKOSI_COMMANDS:
+    for verb in Verb:
         try:
-            v_i = argv.index(verb)
+            v_i = argv.index(verb.name)
```
gsegatti, line 5708: Could also use `f"{verb}"`, but this seems more readable? Unless we're insisting on using format strings for this. Let me know which one and why, please

DaanDeMeyer (approved):
LGTM

mkosi/__init__.py
```diff
@@ -5083,9 +5083,9 @@ def create_parser() -> ArgumentParserMkosi:
     parser = ArgumentParserMkosi(prog="mkosi", description="Build Bespoke OS Images", add_help=False)
 
     group = parser.add_argument_group("Commands")
-    group.add_argument("verb", choices=MKOSI_COMMANDS, default="build", help="Operation to execute")
+    group.add_argument("verb", type=Verb, choices=list(Verb), default=Verb.build, help="Operation to execute")
     group.add_argument(
-        "cmdline", nargs=argparse.REMAINDER, help="The command line to use for " + str(MKOSI_COMMANDS_CMDLINE)[1:-1]
+        "cmdline", nargs=argparse.REMAINDER, help="The command line to use for " + str([verb.name for verb in MKOSI_COMMANDS_CMDLINE])[1:-1]
```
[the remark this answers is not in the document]
DaanDeMeyer: Doesn't look great but it was already using the same approach before so we don't need to fix it in this PR.

Issue #914, opened 2022-02-22
mkosi misdetects kvm availability


vt-alt:
We had [accidental boot failures](https://github.com/lkrg-org/lkrg/issues/132#issuecomment-1048134856) on GA for `mkosi qemu` [like this](https://github.com/lkrg-org/lkrg/runs/5259672421?check_suite_focus=true):
```
‣ Running command:
‣ qemu-system-x86_64 -machine type=q35,accel=kvm,smm=off -smp 2 -m 1G -object rng-random,filename=/dev/urandom,id=rng0 -device virtio-rng-pci,rng=rng0,id=rng-device0 -cpu host -nographic -nodefaults -serial mon:stdio -vga virtio -drive if=none,id=hd,file=/home/runner/work/lkrg/lkrg/image.raw,format=raw -device virtio-scsi-pci,id=scsi -device scsi-hd,drive=hd,bootindex=1 -no-reboot

7lSeaBIOS (version 1.13.0-1ubuntu1.1)
Booting from Hard Disk...
[    0.0KVM: entry failed, hardware error 0xffffffff
...
Error: Process completed with exit code 1.
```
This is because mkosi tries to boot with KVM but [GA does not have KVM support](https://github.com/actions/virtual-environments/issues?q=kvm). In turn this is because mkosi misdetects availability of KVM just by presence of `/dev/kvm` like this:
```
mkosi (main)$ git grep -e has_kvm -e accel
mkosi/__init__.py:    has_kvm = os.path.exists("/dev/kvm")
mkosi/__init__.py:    accel = "kvm" if has_kvm else "tcg"
mkosi/__init__.py:        f"type=q35,accel={accel},smm={'on' if fw_supports_sb else 'off'}",
mkosi/__init__.py:    if has_kvm:
```
I suggest better detection mechanism — open `/dev/kvm` for reading and if it fails there is no KVM support.

behrmann:
Thanks for the detailed bug report!

Will this be as simple as this
```python
try:
    with open("/dev/kvm"):
        has_kvm = True
except:
    has_kvm = False
```
or is something else necessary?

vt-alt:
Looks good!

PR #915, opened 2022-02-23
Check /dev/kvm more thoroughly.

behrmann:
Some CI runners, among them apparently GA, present non-working KVM devices. Let's try to open it, to see whether we can.

Fixes: #914

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -7546,9 +7546,21 @@ def find_ovmf_vars() -> Path:
     die("Couldn't find OVMF UEFI variables file.")
 
 
+def qemu_check_kvm_support() -> bool:
+    kvm = Path("/dev/kvm")
+    if not kvm.is_char_device():
+        return False
+    # some CI runners may present a non-working KVM device
+    try:
+        with kvm.open("r+b"):
+            return True
+    except OSError:
+        return False
+
+
 @contextlib.contextmanager
 def run_qemu_cmdline(args: MkosiArgs) -> Iterator[List[str]]:
-    has_kvm = os.path.exists("/dev/kvm")
+    has_kvm = qemu_check_kvm_support()
     accel = "kvm" if has_kvm else "tcg"
 
     firmware, fw_supports_sb = find_qemu_firmware()


behrmann:
@vt-alt Can you have a look whether this fixes your issue?

mkosi/__init__.py
```diff
@@ -7546,9 +7546,21 @@ def find_ovmf_vars() -> Path:
     die("Couldn't find OVMF UEFI variables file.")
 
 
+def qemu_check_kvm_support() -> bool:
+    kvm = Path("/dev/kvm")
+    if not kvm.exists():
```
behrmann, line 7551: Not sure, but I guess `/dev/kvm` is always a character device, if so we could make this more precise. 
```suggestion
    if not kvm.is_char_device():
```
I don't know whether this would already obviate the attempted solution below

behrmann:
@davide125 I remember you being the CentOS expert, can you help me understand this CI failure?
> Error: Transaction test error:
  file /usr/lib/systemd/system/systemd-network-generator.service conflicts between attempted installs of systemd-networkd-249.7-2.el9.x86_64 and systemd-250-3.el9.x86_64
  file /usr/lib/systemd/systemd-network-generator conflicts between attempted installs of systemd-networkd-249.7-2.el9.x86_64 and systemd-250-3.el9.x86_64

mkosi/__init__.py
DaanDeMeyer: Probably a good idea to add it but I don't think it hurts to still do the below check.

mkosi/__init__.py
```diff
@@ -7546,9 +7546,21 @@ def find_ovmf_vars() -> Path:
...
+        return False
+    # some CI runners may present a non-working KVM device
+    try:
+        with kvm.open():
```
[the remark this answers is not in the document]
behrmann: One question, should this be
```suggestion
        with kvm.open("r+"):
```
or
```suggestion
        with kvm.open("r+b"):
```
though?

davide125:
systemd-networkd in CentOS comes from the systemd-extras EPEL package (https://src.fedoraproject.org/rpms/systemd-extras). That package is still at 249, while systemd proper was just rebased to 250 last week, which is why you're seeing that failure. cc @robert-scheck

behrmann:
Thanks for the explanation, @davide125! Is @robert-scheck the right person to ping if we notice this?

davide125:
I'd recommend filing an issue on https://bugzilla.redhat.com/buglist.cgi?bug_status=NEW&bug_status=ASSIGNED&classification=Fedora&component=systemd-extras&product=Fedora&product=Fedora%20EPEL

In this case, looks like we're already covered by https://bugzilla.redhat.com/show_bug.cgi?id=2056276

robert-scheck:
In general it's fine to ping me here or file a RHBZ, in this specific case a RHBZ already exists - and I need to rebase systemd-extras to v250 first, which is on my to-do list for the end of the week and/or the weekend.

mkosi/__init__.py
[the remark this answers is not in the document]
vt-alt: Read-write open seems to be better because it also implicitly checks for (required) write permissions.

vt-alt:
> @vt-alt Can you have a look whether this fixes your issue?

Since unusable `/dev/kvm` on GA is intermittent and hard to debug, to test this properly on GA would be overly complicated.

So I tested locally — `rmmod'`ing `kvm-intel` and creating KVM device with `mknod /dev/kvm c 10 232` — mkosi worked good by not trying to use KVM.
```
root@ubuntu-jammy:~/lkrg# rmmod kvm-intel
root@ubuntu-jammy:~/lkrg# mknod /dev/kvm c 10 232
root@ubuntu-jammy:~/lkrg# chmod a+rw /dev/kvm
root@ubuntu-jammy:~/lkrg# ls -l /dev/kvm
crw-rw-rw- 1 root root 10, 232 Feb 23 20:30 /dev/kvm
root@ubuntu-jammy:~/lkrg# mkosi qemu ...
‣ Warning: Couldn't find OVMF firmware blob with secure boot support,
falling back to OVMF firmware blobs without secure boot support.
‣ Running command:
‣ qemu-system-x86_64 -machine type=q35,accel=tcg,smm=off -smp 2 -m 1G ...
```

vt-alt:
Well, I'm reported that GA failures are not always at qemu start, but sometimes boot is proceeding several milliseconds before reaching `KVM: entry failed, hardware error 0xffffffff` there is example (old runs w/o this patch):
https://github.com/lkrg-org/lkrg/runs/5251322870?check_suite_focus=true
https://github.com/lkrg-org/lkrg/runs/5251323060?check_suite_focus=true

This is obscure! But this does not invalidate this PR improvement. Since non-working (created by legacy boot scripts) or write protected `/dev/kvm` is common case too.

vt-alt:
JFYI I created GA issue https://github.com/actions/virtual-environments/issues/5128

DaanDeMeyer:
Restarted the CI jobs since the CentOS Stream 9 networkd issue should be fixed now.

DaanDeMeyer:
Hmm, seems the change hasn't been fully deployed yet.

robert-scheck:
[FEDORA-EPEL-2022-10eee90ae6](https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2022-10eee90ae6) hasn't been pushed to any repository, yet.

mkosi/__init__.py
```diff
@@ -7546,9 +7546,21 @@ def find_ovmf_vars() -> Path:
     die("Couldn't find OVMF UEFI variables file.")
 
 
+def qemu_check_kvm_support() -> bool:
+    kvm = Path("/dev/kvm")
+    if not kvm.is_char_device():
+        return False
+    # some CI runners may present a non-working KVM device
+    try:
+        with kvm.open("r+b"):
```
DaanDeMeyer, line 7555: The `b` flag actually doesn't have an effect on Linux, so we can omit it. (see https://unix.stackexchange.com/questions/429969/why-opening-file-for-binary-access-is-meaningless-on-unix-nor-the-for-bot)

behrmann: It does make a difference in Python, though.
```
In [1]: from pathlib import Path

In [2]: bar = Path("bar")

In [3]: with bar.open("r+") as f:
   ...:     foo = f.read()

In [4]: foo
Out[4]: 'foo\n'

In [5]: with bar.open("r+b") as f:
   ...:     foo = f.read()

In [6]: foo
Out[6]: b'foo\n'
```
When opened with `b` Python will read bytes instead of strings, which won't attempt any decoding.

We don't really want to read anything here, but I'd gather that in the end this is more defensive.

DaanDeMeyer: Ahh, in that case we can keep the flat yeah since we don't care about python's text processing

DaanDeMeyer (approved):
LGTM

PR #916, opened 2022-02-24
Add PowerTools repo for CentOS Stream 8

DaanDeMeyer:
Enables installing development related packages (e.g. meson).


DaanDeMeyer:
centos tests failing is expected because of https://github.com/systemd/mkosi/pull/915#issuecomment-1049035316

bluca (approved):

DaanDeMeyer:
Doesn't affect gentoo so merging

davide125:
Ah, good catch! We should also add CRB (the equivalent of PowerTools) for CentOS Stream 9.

DaanDeMeyer:
I checked and it seems you added that one already

davide125:
Yup, I just realized: https://github.com/systemd/mkosi/blob/main/mkosi/__init__.py#L2466 so we're good there.

PR #917, opened 2022-02-25
ci: Remove nspawn Arch workaround

DaanDeMeyer:


DaanDeMeyer:
Let's see if this has magically been fixed by now

DaanDeMeyer:
All the arch tests passed so it seems this has been fixed, let's merge. Will remove the workaround in the systemd repo as well.

behrmann:
Few things feel as good as a kludge removed.

PR #919, opened 2022-02-28
Proof of concept for integration testing

gsegatti:
This PR has the goal of presenting a proof of concept implementation of what could be Mkosi's integration testing scheme.

Not so long ago [Machine](https://github.com/systemd/mkosi/pull/894) class was merged into main. That class provided us an abstraction to build/boot a Mkosi image and then run commands in that particular image from within a [python script ](https://github.com/systemd/mkosi/blob/main/tests/test_machine.py). This allowed setting testing functions that would build/boot run commands and exit the a Mkosi image within its scope.

Now, we've extended the usage of a single Mkosi build. That is, we are able to keep an image configuration running between multiple test cases. Therefore, the time once spent waiting during testing, while an image was built is dramatically decreased, since we don't need to rebuilt it at the beginning of every function anymore (we do reboot the image so that previous test runs don't interfere, but the time cost is marginal).

The way this is being leveraged into testing is through the Unittest library, the main reason being: we want to allow
as much possible scenarios of testing with Mkosi's structure, in the most manageable way (will get to why Pytest didn't provide that in a second).
Currently, we have a `MkosiMachineTest` class, which gives us a few things:

- `setUpClass()`, grants every subclass a running Mkosi machine with the specified configurations from "args".
- `setUp()` boots an Image at each testing function.
- `tearDown`, kill that image so the next test has a clean image.

Then, for each test class inheriting`MkosiMachineTest`, we'll initialise it with the arguments that its Mkosi image should have in order to run tests. Then, each class contains tests for which one specific configuration would be required.

Where Pytest was less than ideal in order to accomplish this (as far as I could find), is when we'd have the need to generate multiple Mkosi images with different configurations.
We want each fixture to return an instance of the Machine class, which will contain a running Mkosi image. [Pytest's fixtures are unable to receive parameters](https://github.com/pytest-dev/pytest/issues/8109), which means we would have to replicate fixtures multiple times with the hard coded parameters needed. This would also cause testing functions to become huge, if we desired to reuse one instance of the Machine class multiple times. 

So, if we wanted 2 Mkosi images, with different configurations, we'd have:

```
@pytest.fixture
def fixture_1():
    return machine[args]

def test_1(machine):
     machine.run([...])

@pytest.fixture
def fixture_2():
    return other_machine[args]

def test_2(other_machine):
     machine.run([...])
```

And, as we wanted to reutilize the configurations from `fixture_1`, the testing function would start to look like this:

```
@pytest.fixture
def fixture_1():
    return machine

def test_1(machine):
    machine.run([...])
    assert (...)
    machine.run([...])
    assert(...)
        .
        .
        .
```

With Unittest the scalability of tests becomes easier to manage, since, as long as there's a class with the configuration you need, you can simply add another testing method to it.
We also still use Pytest as the tests runner.

Please, let me know what you think about this, any suggestions or doubts and so on.

### DIFF
--- a/mkosi/machine.py
+++ b/mkosi/machine.py
@@ -3,8 +3,10 @@
 from __future__ import annotations
 
 import contextlib
+import os
 import signal
 import subprocess
+import unittest
 from textwrap import dedent
 from typing import Any, Optional, Sequence
 
@@ -21,6 +23,7 @@ from . import (
     load_args,
     needs_build,
     parse_args,
+    parse_boolean,
     prepend_to_environ_path,
     run_command_image,
     run_qemu_cmdline,
@@ -31,16 +34,30 @@ from .backend import MkosiArgs, Verb, die
 
 
 class Machine:
-    def __init__(self, args: Optional[Sequence[str]] = None, debug: bool = False) -> None:
-        # Remains None until image is built and booted, then receives pexpect process
+    def __init__(self, args: Sequence[str] = [], debug: bool = False) -> None:
+        # Remains None until image is built and booted, then receives pexpect process.
         self._serial: Optional[pexpect.spawn] = None
         self.exit_code: int = -1
         self.debug = debug
         self.stack = contextlib.ExitStack()
         self.args: MkosiArgs
 
-        # We make sure to add the arguments in the machine class itself, rather than typing this for every testing function.
         tmp = parse_args(args)["default"]
+
+        # By default, Mkosi makes Verb = build.
+        # But we want to know if a test class passed args without a Verb.
+        # If so, we'd like to assign default values or the environment's variable value.
+        if tmp.verb == Verb.build and Verb.build.name not in args:
+            verb = os.getenv("MKOSI_TEST_DEFAULT_VERB")
+            # This way, if environment variable is not set, we assign nspawn by default to test classes with no Verb.
+            if verb in (Verb.boot.name, None):
+                tmp.verb = Verb.boot
+            elif verb == Verb.qemu.name:
+                tmp.verb = Verb.qemu
+            else:
+                die("No valid verb was entered.")
+
+        # Add the arguments in the machine class itself, rather than typing this for every testing function.
         tmp.force = 1
         tmp.autologin = True
         if tmp.verb == Verb.qemu:
@@ -70,11 +87,11 @@ class Machine:
                     )
         return self._serial
 
-    def ensure_booted(self) -> None:
+    def _ensure_booted(self) -> None:
         # Try to access the serial console which will raise an exception if the machine is not currently booted.
         assert self._serial is not None
 
-    def __enter__(self) -> Machine:
+    def build(self) -> None:
         if self.args.verb in MKOSI_COMMANDS_SUDO:
             check_root()
             unlink_output(self.args)
@@ -88,10 +105,17 @@ class Machine:
             init_namespace(self.args)
             build_stuff(self.args)
 
+    def __enter__(self) -> Machine:
+        self.build()
+        self.boot()
+
+        return self
+
+    def boot(self) -> None:
         with contextlib.ExitStack() as stack:
             prepend_to_environ_path(self.args.extra_search_paths)
 
-            if self.args.verb in (Verb.shell, Verb.boot):
+            if self.args.verb == Verb.boot:
                 cmdline = run_shell_cmdline(self.args)
             elif self.args.verb == Verb.qemu:
                 # We must keep the temporary file opened at run_qemu_cmdline accessible, hence the context stack.
@@ -101,19 +125,17 @@ class Machine:
 
             cmd = " ".join(str(x) for x in cmdline)
 
-            # Here we have something equivalent to the command lines used on spawn() and run() from backend.py
-            # We use pexpect to boot an image that we will be able to interact with in the future
-            # Then we tell the process to look for the # sign, which indicates the CLI for that image is active
-            # Once we've build/boot an image the CLI will prompt "root@image ~]# "
-            # Then, when pexpects finds the '#' it means we're ready to interact with the process
+            # Here we have something equivalent to the command lines used on spawn() and run() from backend.py.
+            # We use pexpect to boot an image that we will be able to interact with in the future.
+            # Then we tell the process to look for the # sign, which indicates the CLI for that image is active.
+            # Once we've build/boot an image the CLI will prompt "root@image ~]# ".
+            # Then, when pexpects finds the '#' it means we're ready to interact with the process.
             self._serial = pexpect.spawnu(cmd, logfile=None, timeout=240)
             self._serial.expect("#")
             self.stack = stack.pop_all()
 
-        return self
-
     def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -> CompletedProcess:
-        self.ensure_booted()
+        self._ensure_booted()
 
         process = run_command_image(self.args, commands, timeout, check, subprocess.PIPE, subprocess.PIPE)
         if self.debug:
@@ -122,9 +144,41 @@ class Machine:
 
         return process
 
+    def kill(self) -> None:
+        self.__exit__(None, None, None)
+
     def __exit__(self, *args: Any, **kwargs: Any) -> None:
         if self._serial:
             self._serial.kill(signal.SIGTERM)
             self.exit_code = self._serial.wait()
             self._serial = None
         self.stack.__exit__(*args, **kwargs)
+
+
+class MkosiMachineTest(unittest.TestCase):
+    args: Sequence[str]
+    machine: Machine
+
+    def __init_subclass__(cls, args: Sequence[str] = []) -> None:
+        cls.args = args
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        cls.machine = Machine(cls.args)
+
+        verb = cls.machine.args.verb
+        no_nspawn = parse_boolean(os.getenv("MKOSI_TEST_NO_NSPAWN", "0"))
+        no_qemu = parse_boolean(os.getenv("MKOSI_TEST_NO_QEMU", "0"))
+
+        if no_nspawn and verb == Verb.boot:
+            raise unittest.SkipTest("Nspawn test skipped due to environment variable.")
+        if no_qemu and verb == Verb.qemu:
+            raise unittest.SkipTest("Qemu test skipped due to environment variable.")
+
+        cls.machine.build()
+
+    def setUp(self) -> None:
+        self.machine.boot()
+
+    def tearDown(self) -> None:
+        self.machine.kill()
--- a/tests/test_machine.py
+++ b/tests/test_machine.py
@@ -5,64 +5,50 @@ from subprocess import TimeoutExpired
 
 import pytest
 
-import mkosi.machine as machine
 from mkosi.backend import MkosiException
+from mkosi.machine import Machine, MkosiMachineTest
 
 pytestmark = [
     pytest.mark.integration,
-    pytest.mark.parametrize("verb", ["boot", "qemu"]),
     pytest.mark.skipif(os.getuid() != 0, reason="Must be invoked as root.")
 ]
 
+class MkosiMachineTestCase(MkosiMachineTest):
+    def test_simple_run(self) -> None:
+        process = self.machine.run(["echo", "This is a test."])
+        assert process.stdout.strip("\n") == "This is a test."
 
-def test_simple_run(verb: str) -> None:
-    with machine.Machine([verb]) as m:
-        p = m.run(["echo", "This is a test."])
-        assert "This is a test." == p.stdout.strip("\n")
-
-    assert m.exit_code == 0
-
-
-def test_wrong_command(verb: str) -> None:
-    # First tests with argument check = True from mkosi.backend.run(), therefore we see if an exception is raised
-    with machine.Machine([verb]) as m:
+    def test_wrong_command(self) -> None:
+        # Check = True from mkosi.backend.run(), therefore we see if an exception is raised
         with pytest.raises(MkosiException):
-            m.run(["NonExisting", "Command"])
+            self.machine.run(["NonExisting", "Command"])
         with pytest.raises(MkosiException):
-            m.run(["ls", "NullDirectory"])
+            self.machine.run(["ls", "NullDirectory"])
 
-    assert m.exit_code == 0
-
-    # Second group of tests with check = False to see if stderr and returncode have the expected values
-    with machine.Machine([verb]) as m:
-        result = m.run(["NonExisting", "Command"], check=False)
+        # Check = False to see if stderr and returncode have the expected values
+        result = self.machine.run(["NonExisting", "Command"], check=False)
         assert result.returncode in (203, 127)
 
-        result = m.run(["ls", "-"], check=False)
+        result = self.machine.run(["ls", "-"], check=False)
         assert result.returncode == 2
         assert "No such file or directory" in result.stderr
 
-    assert m.exit_code == 0
-
-
-def test_infinite_command(verb: str) -> None:
-    with machine.Machine([verb]) as m:
+    def test_infinite_command(self) -> None:
         with pytest.raises(TimeoutExpired):
-            m.run(["tail", "-f", "/dev/null"], 2)
-
-    assert m.exit_code == 0
+            self.machine.run(["tail", "-f", "/dev/null"], 2)
 
 
-def test_before_boot(verb: str) -> None:
-    m = machine.Machine([verb])
+def test_before_boot() -> None:
+    m = Machine()
     with pytest.raises(AssertionError):
         m.run(["ls"])
 
 
-def test_after_shutdown(verb: str) -> None:
-    with machine.Machine([verb]) as m:
+def test_after_shutdown() -> None:
+    with Machine() as m:
         pass
 
     with pytest.raises(AssertionError):
         m.run(["ls"])
+
     assert m.exit_code == 0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,7 +125,9 @@ jobs:
       run: sudo apt-get update && sudo apt-get install --no-install-recommends python3-pexpect python3-pytest
 
     - name: Run integration tests
-      run: sudo python3 -m pytest -m integration tests
+      run: |
+        MKOSI_TEST_DEFAULT_VERB=qemu sudo python3 -m pytest -m integration tests
+        MKOSI_TEST_DEFAULT_VERB=boot sudo python3 -m pytest -m integration tests
 
   integration-test:
     runs-on: ubuntu-20.04
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -5986,12 +5986,12 @@ def unlink_output(args: MkosiArgs) -> None:
 
 
 def parse_boolean(s: str) -> bool:
-    "Parse 1/true/yes as true and 0/false/no as false"
+    "Parse 1/true/yes/y/t/on as true and 0/false/no/n/f/off/None as false"
     s_l = s.lower()
-    if s_l in {"1", "true", "yes"}:
+    if s_l in {"1", "true", "yes", "y", "t", "on"}:
         return True
 
-    if s_l in {"0", "false", "no"}:
+    if s_l in {"0", "false", "no", "n", "f", "off"}:
         return False
 
     raise ValueError(f"Invalid literal for bool(): {s!r}")


tests/test_machine.py
```diff
@@ -1,69 +1,68 @@
...
-        with pytest.raises(TimeoutExpired):
-            m.run(["tail", "-f", "/dev/null"], 2)
+        # Check = False to see if stderr and returncode have the expected values
+        result = self.machine.run(["NonExisting", "Command"], check=False)
+        self.assertTrue(result.returncode in (203, 127))
 
-    assert m.exit_code == 0
+        result = self.machine.run(["ls", "-"], check=False)
+        self.assertEqual(result.returncode, 2)
+        self.assertTrue("No such file or directory" in result.stderr)
```
DaanDeMeyer, line 51: We'll still be using pytest as the test runner so we can just use `assert` here instead of the unittest specific methods

mkosi/machine.py
```diff
@@ -62,19 +63,19 @@ def serial(self) -> pexpect.spawn:
...
                     )
         return self._serial
 
-    def ensure_booted(self) -> None:
+    def __ensure_booted(self) -> None:
         # Try to access the serial console which will raise an exception if the machine is not currently booted.
         assert self._serial is not None
 
-    def __enter__(self) -> Machine:
+    def __build_image(self) -> None:
```
DaanDeMeyer, line 78: This doesn't really need to be a separate method, we can just keep this logic in __enter__

mkosi/machine.py
```diff
@@ -88,6 +89,7 @@ def __enter__(self) -> Machine:
             init_namespace(self.args)
             build_stuff(self.args)
 
+    def __boot_image(self) -> None:
```
DaanDeMeyer, line 92: Private methods in python should generally be prefixed with just a single underscore, not two.

mkosi/machine.py
```diff
@@ -122,9 +127,16 @@ def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -
...
             self._serial.kill(signal.SIGTERM)
             self.exit_code = self._serial.wait()
             self._serial = None
         self.stack.__exit__(*args, **kwargs)
+
+    def __exit__(self, *args: Any, **kwargs: Any) -> None:
+        self.__kill_image(*args, **kwargs)
+
+    def reboot_image(self) -> None:
+        self.__exit__(*exc_info())
```
DaanDeMeyer, line 141: This should call _kill_image instead of __exit__.

mkosi/machine.py
```diff
@@ -122,9 +127,16 @@ def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -
 
         return process
 
-    def __exit__(self, *args: Any, **kwargs: Any) -> None:
+    def __kill_image(self, *args: Any, **kwargs: Any) -> None:
```
DaanDeMeyer, line 130: You can name this _kill, no need for the image suffix. Since this is a method of the machine class, it's pretty obvious that the _kill method means that we're killing the machine. Same applies to the _boot_image method.

tests/test_machine.py
```diff
@@ -1,69 +1,68 @@
...
+    machine: Machine
 
-def test_wrong_command(verb: str) -> None:
-    # First tests with argument check = True from mkosi.backend.run(), therefore we see if an exception is raised
-    with machine.Machine([verb]) as m:
-        with pytest.raises(MkosiException):
-            m.run(["NonExisting", "Command"])
-        with pytest.raises(MkosiException):
-            m.run(["ls", "NullDirectory"])
+    def __init_subclass__(cls, /, args: str, reboot: bool = True) -> None:
```
DaanDeMeyer, line 16: Why is `args` a string here? Shouldn't it be a list of strings?

tests/test_machine.py
```diff
@@ -1,69 +1,68 @@
...
+        cls.machine = Machine([cls.args])
+        cls.machine.__enter__()
 
-    # Second group of tests with check = False to see if stderr and returncode have the expected values
-    with machine.Machine([verb]) as m:
-        result = m.run(["NonExisting", "Command"], check=False)
-        assert result.returncode in (203, 127)
+    @classmethod
+    def tearDownClass(cls) -> None:
+        cls.machine.__exit__(*exc_info())
```
DaanDeMeyer, line 27: exc_info() will always return (None, None, None) here since teardownClass() isn't called inside the exception handler block by unittest.

The slightly hacky way to deal with this is to override the internal _callTestMethod() method of unittest.TestCase so we can save the exception result:

Example:

```
def _callTestMethod(self, method: str) -> None:
    try:
        super()._callTestMethod(method)
    except Exception as e:
        self.exc_info = exc_info()
        raise e
```

And then you can pass `self.exc_info()` to `__exit__()` in `tearDownClass()`.

tests/test_machine.py
```diff
@@ -1,69 +1,68 @@
...
 
+    def test_infinite_command(self) -> None:
+        with self.assertRaises(TimeoutExpired):
+            self.machine.run(["tail", "-f", "/dev/null"], 2)
 
-def test_before_boot(verb: str) -> None:
-    m = machine.Machine([verb])
-    with pytest.raises(AssertionError):
-        m.run(["ls"])
+    def test_before_boot(self) -> None:
```
DaanDeMeyer, line 57: So this method and `test_after_shutdown()` are slightly special in that what they do is kinda outside the scope of what's supported by MkosiMachineTest. As such, we can keep these two outside of MkosiMachineTest.

DaanDeMeyer (changes requested):

mkosi/machine.py
gsegatti: I assume the same for `_reboot_image` then?

tests/test_machine.py
gsegatti: This was actually a typo.

mkosi/machine.py
```diff
@@ -122,9 +125,16 @@ def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -
...
         if self._serial:
             self._serial.kill(signal.SIGTERM)
             self.exit_code = self._serial.wait()
             self._serial = None
         self.stack.__exit__(*args, **kwargs)
+
+    def __exit__(self, *args: Any, **kwargs: Any) -> None:
+        self._kill(*args, **kwargs)
+
+    def reboot_image(self, *args: Any, **kwargs: Any) -> None:
```
DaanDeMeyer, line 138: This one can be renamed to just reboot() as well

gsegatti: oh, good, asked that above as well

mkosi/machine.py
```diff
@@ -101,19 +89,34 @@ def __enter__(self) -> Machine:
...
+            check_output(self.args)
+
+        if needs_build(self.args):
+            check_root()
+            check_native(self.args)
+            init_namespace(self.args)
+            build_stuff(self.args)
+        self._boot()
+
         return self
```
DaanDeMeyer, line 116: This has unnecessarily been moved downwards. You should always try to avoid unnecessary changes in PRs to keep them small.

gsegatti: Hmm, that's fair. Didn't really think of that until now, will keep it in mind in the future!

tests/test_machine.py
```diff
@@ -1,69 +1,82 @@
...
 
-def test_after_shutdown(verb: str) -> None:
-    with machine.Machine([verb]) as m:
+def test_after_shutdown() -> None:
+    with Machine(["boot"]) as machine:
         pass
 
     with pytest.raises(AssertionError):
-        m.run(["ls"])
-    assert m.exit_code == 0
```
DaanDeMeyer, line 69: Same here, a bunch of unnecessary changes. Let's keep these the same as much as possible aside from the verb change

tests/test_machine.py
```diff
@@ -1,69 +1,82 @@
...
+    exc_info: Tuple[Optional[Type[BaseException]], Optional[BaseException], Optional[TracebackType]]
+
+    def _callTestMethod(self, method: str) -> None:
+        try:
+            super()._callTestMethod(method)
+        except Exception as e:
+            self.exc_info = exc_info()
+            raise e
+
+    def __init_subclass__(cls, args: List[str], reboot: bool = True) -> None:
```
DaanDeMeyer, line 27: Like we discussed, let's not make `reboot` configurable. We always want to reboot so that tests cannot accidentally depend on the result of previous tests which makes adding concurrency in the future much harder.

DaanDeMeyer: Also, `args` can be `Sequence` instead of `List` I think

tests/test_machine.py
```diff
@@ -1,69 +1,82 @@
...
-    assert m.exit_code == 0
-
-
-def test_wrong_command(verb: str) -> None:
-    # First tests with argument check = True from mkosi.backend.run(), therefore we see if an exception is raised
-    with machine.Machine([verb]) as m:
+from mkosi.machine import Machine
+
+
+class MkosiMachineTest(unittest.TestCase):
```
DaanDeMeyer, line 14: We should not put this class in test_machine as we want it to be reusable across multiple test files. You can put this class next to the `Machine` class in mkosi/machine.py

mkosi/machine.py
```diff
@@ -122,9 +127,44 @@ def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -
...
+    def reboot(self, *args: Any, **kwargs: Any) -> None:
+        self._kill(*args, **kwargs)
+        self._boot()
+
+
+class MkosiMachineTest(unittest.TestCase):
+    args: Sequence[str]
+    machine: Machine
+    exc_info: Tuple[Optional[Type[BaseException]], Optional[BaseException], Optional[TracebackType]] \
+              = (None, None, None)
```
gsegatti, line 149: I've used the same formatting as `__init__.py`, like line 954 or 7107 and so on.

mkosi/machine.py
```diff
@@ -122,9 +127,44 @@ def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -
...
             self._serial.kill(signal.SIGTERM)
             self.exit_code = self._serial.wait()
             self._serial = None
         self.stack.__exit__(*args, **kwargs)
+
+    def __exit__(self, *args: Any, **kwargs: Any) -> None:
+        self._kill(*args, **kwargs)
+
+    def reboot(self, *args: Any, **kwargs: Any) -> None:
+        self._kill(*args, **kwargs)
```
DaanDeMeyer, line 141: This signature for reboot() is quite confusing. It's hard to see from the signature what arguments we are supposed to pass to it.

I'd go for a different solution where if reboot() is called manually, let's not pass any exception information to `_kill`. So remove the arguments from reboot() and have it call `_kill` like so: `self._kill(None, None, None)`

tests/test_machine.py
```diff
@@ -5,64 +5,51 @@
...
-    pytest.mark.parametrize("verb", ["boot", "qemu"]),
     pytest.mark.skipif(os.getuid() != 0, reason="Must be invoked as root.")
 ]
 
 
-def test_simple_run(verb: str) -> None:
-    with machine.Machine([verb]) as m:
-        p = m.run(["echo", "This is a test."])
-        assert "This is a test." == p.stdout.strip("\n")
+class MkosiMachineTester(MkosiMachineTest, args=["boot"]):
```
DaanDeMeyer, line 17: `MkosiMachineTester` is a weird name for a class. Maybe go for `MkosiMachineTestCase` which matches `unittest.TestCase` more closely?

DaanDeMeyer (commented):
Looking good, left a few more comments

mkosi/machine.py
gsegatti: This may have broken the exit of qemu, will try again later to confirm it

gsegatti:
This is just an early version with the `MKOSI_TEST_DEFAULT_VERB` variable included.

mkosi/machine.py
```diff
@@ -32,15 +36,27 @@
...
+        # We could have the Verb passed by an environment variable, rather than only args.
+        if tmp.verb not in (Verb.qemu, Verb.boot):
+            default_verb = os.environ.get("MKOSI_TEST_DEFAULT_VERB")
+            # This way, if environment var is not set, we assign nspawn by default to test cases with no Verb
+            if default_verb in (Verb.boot.name, None):
+                tmp.verb = Verb.boot
+            elif default_verb == Verb.qemu.name:
+                tmp.verb = Verb.qemu
+            else:
+                die("No valid verb was entered.")
```
gsegatti, line 57: The problem with this is: if someone gives us a verb like `shell`, `help`, we'll override that value with `boot` or `qemu`, instead of calling `die`.
So I should change to something like `if tmp.verb == None`

mkosi/machine.py
```diff
@@ -32,15 +36,27 @@
...
+
+        # Add the arguments in the machine class itself, rather than typing this for every testing function.
```
gsegatti, line 59: The problem here is: if `MKOSI_TEST_DEFAULT_VERB == None` and someone passes a Verb not in `(Verb.qemu, Verb.boot)`, like "shell", we'll override that value with Verb.boot

gsegatti: I've found a way around it will push it

mkosi/machine.py
```diff
@@ -122,9 +142,46 @@ def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -
...
             self.exit_code = self._serial.wait()
             self._serial = None
         self.stack.__exit__(*args, **kwargs)
+
+    def __exit__(self) -> None:
+        self._kill(*exc_info())
+
+    def reboot(self) -> None:
+        self._kill(*exc_info())
+        self._boot()
```
gsegatti, line 157: I'm aware of [this comment](https://github.com/systemd/mkosi/pull/919#discussion_r815975681). But this looks cleaner since we don't override anything with the `MkosiMachineTest` class. 

Plus the signature for both `exit` and `reboot` looks simpler. 
Unless there's a specific reason to do it that way? Then, please, let me know,

gsegatti: Well, this apparently breaks the Machine when used as a context manager.
```
def test_after_shutdown() -> None:
        with Machine(["boot"]) as m:
>           pass
E           TypeError: __exit__() takes 1 positional argument but 4 were given
```

gsegatti: This works for all scenarios:
`def __exit__(self, *args: Any, **kwargs: Any) -> None:
        self._kill(*exc_info())

    def reboot(self) -> None:
        self._kill(*exc_info())
        self._boot()`

mkosi/machine.py
```diff
@@ -122,9 +142,46 @@ def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -
...
+
+    def __init_subclass__(cls, args: Sequence[str] = []) -> None:
+        cls.args = args
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        cls.machine = Machine(cls.args)
+
+        mkosi_tests_no_qemu = os.environ.get("MKOSI_TESTS_NO_QEMU")
+        mkosi_tests_no_nspawn = os.environ.get("MKOSI_TESTS_NO_NSPAWN")
```
gsegatti, line 174: Maybe add a check to make sure the values are (1,0) and if anything else, call `die()`?

mkosi/machine.py
```diff
@@ -31,16 +35,30 @@
...
         self.stack = contextlib.ExitStack()
         self.args: MkosiArgs
 
-        # We make sure to add the arguments in the machine class itself, rather than typing this for every testing function.
         tmp = parse_args(args)["default"]
+
+        # By default, Mkosi makes Verb = build.
+        # But we want to know if a test class passed args without a Verb
+        # If so, we'd like to check the values of the environment variable.
+        if Verb.build.name not in args:
```
gsegatti, line 51: This is actually:
`if tmp.verb == Verb.build and Verb.build.name not in args:`

mkosi/machine.py
DaanDeMeyer: Why not just do:

```
def __exit__(self, *args: Any, **kwargs: Any) -> None:
    self._kill(*args, **kwargs)
```

?

exc_info() should only be used as a last resort since it will only give you the exception that's being processed at the moment. If __exit__ is called after the exception handling block ends, exc_info() will return (None, None, None) which isn't all that useful if the actual exception information is being passed as arguments to `__exit__()`.

See https://docs.python.org/3/library/sys.html#sys.exc_info for more information

mkosi/machine.py
DaanDeMeyer: Systemd uses the following logic to parse booleans from environment variables. Let's allow the same strings to be used for mkosi's environment variables.

```

int parse_boolean(const char *v) {
        if (!v)
                return -EINVAL;

        if (STRCASE_IN_SET(v,
                           "1",
                           "yes",
                           "y",
                           "true",
                           "t",
                           "on"))
                return 1;

        if (STRCASE_IN_SET(v,
                           "0",
                           "no",
                           "n",
                           "false",
                           "f",
                           "off"))
                return 0;

        return -EINVAL;
}
```

mkosi/machine.py
```diff
@@ -31,16 +35,30 @@
 
 
 class Machine:
-    def __init__(self, args: Optional[Sequence[str]] = None, debug: bool = False) -> None:
-        # Remains None until image is built and booted, then receives pexpect process
+    def __init__(self, args: Sequence[str], debug: bool = False) -> None:
```
gsegatti, line 38: Mypy, in line 51, does not allows me to look for the `Verb.build` in `args` if it's type is `Optional[Sequence[str]] = None`.
```

mkosi/machine.py:51:39: error: Unsupported right operand type for in ("Optional[Sequence[str]]")  [operator]
            if tmp.verb == Verb.build and Verb.build.name not in args:
```

mkosi/machine.py
```diff
@@ -122,9 +142,71 @@ def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -
...
+    def __init_subclass__(cls, args: Sequence[str] = []) -> None:
+        cls.args = args
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        cls.machine = Machine(cls.args)
+
+        machine_verb = cls.machine.args.verb
+        no_nspawn = cls._parse_boolean("MKOSI_TESTS_NO_NSPAWN")
+        no_qemu = cls._parse_boolean("MKOSI_TESTS_NO_QEMU")
```
gsegatti, line 198: The way this is setup, both environment variables need values in order to execute tests.
Maybe if we have no assigned value we should just raise a warning message? As the number of tests pile up it would be important to make sure such variables are defined though.

mkosi/machine.py
DaanDeMeyer: Adding `= []` to the constructor should be fine here, since `Sequence` means the variable is read-only and can't be modified. Just be careful that you can never use mutable objects as the default value as they will not be reset after the method finishes.

mkosi/machine.py
```diff
@@ -122,9 +142,71 @@ def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -
...
+        self._kill(*args, **kwargs)
+
+    def reboot(self) -> None:
+        self._kill(None, None, None)
+        self._boot()
+
+
+class MkosiMachineTest(unittest.TestCase):
+    mkosi_tests_no_qemu: bool
+    mkosi_tests_no_nspawn: bool
```
DaanDeMeyer, line 162: These are unused and can be removed

mkosi/machine.py
```diff
@@ -122,9 +142,71 @@ def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -
...
+        value = os.getenv(environment_var)
+        if not value:
+            die(f"No value assigned to {environment_var}.")
+
+        if value in ("1", "yes", "y", "true", "t", "on"):
+            return True
+        if value in ("0", "no", "n", "false", "f", "off"):
+            return False
+
+        die(f"No valid value assigned to {environment_var}.")
```
DaanDeMeyer, line 187: Turns out we already have a parse_boolean function in `__init__.py. Let's extend that function with the same values as from the systemd function and reuse it here instead of adding our own function.

General feedback on this function is that reading the environment variable inside the _parse_boolean method means we can't use it for anything else than environment variables. Instead, the right interface is the same as the one from systemd, where `parse_boolean()` takes a string and as a result can be used to parse booleans from environment variable values but also from any other string.

tests/test_machine.py
```diff
@@ -5,64 +5,50 @@
...
-            m.run(["tail", "-f", "/dev/null"], 2)
-
-    assert m.exit_code == 0
+            self.machine.run(["tail", "-f", "/dev/null"], 2)
 
 
-def test_before_boot(verb: str) -> None:
-    m = machine.Machine([verb])
+def test_before_boot() -> None:
+    m = Machine(["boot"])
```
DaanDeMeyer, line 42: We can remove the explicit verb passing here and below now that we assign a default value in Machine's constructor

mkosi/machine.py
```diff
@@ -122,9 +142,70 @@ def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -
...
+            return False
+
+        s_l = s.lower()
+        if s_l in {"1", "true", "yes", "y", "t", "on"}:
+            return True
+
+        if s_l in {"0", "false", "no", "n", "f", "off"}:
+            return False
+
+        raise ValueError(f"Invalid literal for bool(): {s!r}")
```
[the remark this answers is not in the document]
gsegatti: I just grossly misunderstood what you said, my bad.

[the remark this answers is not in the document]
gsegatti: The thing is parse_boolean requires a value, while `os.getenv()` returns an `Optional[sequence]`.
If we pass an optional arg, then we can't use the `s_l = s.lower()`, since we cant lower() None.
So this would probably break this function as to how it works right now.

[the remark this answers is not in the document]
gsegatti: I just grossly misunderstood what you said. The downside here is that this kinda changes how `parse_boolean`() works by making it return False if the string is empty.
This would raise a valueError originally.

We could check if the strings are different than None before calling the function, but we would still need to change the argument type and use the `if s`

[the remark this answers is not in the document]
gsegatti: We could do this:
`if not os.getenv("MKOSI_TESTS_NO_NSPAWN"):
            no_nspawn = False
        else:
            no_nspawn = parse_boolean(os.getenv("MKOSI_TESTS_NO_NSPAWN"))

        if not os.getenv("MKOSI_TESTS_NO_NSPAWN"):
            no_qemu = False
        else:
            no_qemu = parse_boolean(os.getenv("MKOSI_TESTS_NO_QEMU"))`

And then parse_boolean would look like this:
`def parse_boolean(s: Optional[str]) -> bool:
    "Parse 1/true/yes/y/t/on as true and 0/false/no/n/f/off/None as false"
    if s:
        s_l = s.lower()

    if s_l in {"1", "true", "yes", "y", "t", "on"}:
        return True

    if s_l in {"0", "false", "no", "n", "f", "off"}:
        return False

    raise ValueError(f"Invalid literal for bool(): {s!r}")
`

So its behavior remains the same.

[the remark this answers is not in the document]
DaanDeMeyer: You can configure the default value that `getenv` should return instead of `None`

So what you could do is `no_nspawn = parse_boolean(os.getenv("MKOSI_TESTS_NO_NSPAWN", "0"))`

mkosi/machine.py
```diff
@@ -31,16 +36,30 @@
...
         self.args: MkosiArgs
 
-        # We make sure to add the arguments in the machine class itself, rather than typing this for every testing function.
         tmp = parse_args(args)["default"]
+
+        # By default, Mkosi makes Verb = build.
+        # But we want to know if a test class passed args without a Verb.
+        # If so, we'd like to assign default values or the environment's variable value.
+        if tmp.verb == Verb.build and Verb.build.name not in args:
+            default_verb = os.getenv("MKOSI_TEST_DEFAULT_VERB")
```
DaanDeMeyer, line 53: Nit: for local variables, there's no need to bother with long variable names, the code is perfectly readable if this variable is just named `verb`, no need to go for a long name such as `default_verb`

mkosi/machine.py
```diff
@@ -122,9 +143,52 @@ def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -
...
+            super()._callTestMethod(method) # type: ignore
+        except Exception as e:
+            self.exc_info = exc_info()
+            raise e
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        cls.machine = Machine(cls.args)
+
+        machine_verb = cls.machine.args.verb
```
DaanDeMeyer, line 181: Same here, really long name for a local variable. You can just call this `verb` as well and everyone will still be able to read the code.

mkosi/machine.py
```diff
@@ -122,9 +143,52 @@ def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -
...
+        no_nspawn = parse_boolean(os.getenv("MKOSI_TESTS_NO_NSPAWN", "0"))
+        no_qemu = parse_boolean(os.getenv("MKOSI_TESTS_NO_QEMU", "0"))
```
DaanDeMeyer, line 183: Let's follow the systemd naming and name these `MKOSI_TEST_NO_NSPAWN` and `MKOSI_TEST_NO_QEMU`. Otherwise it gets confusing since we have `MKOSI_TEST_DEFAULT_VERB` where test is not plural and the other two where test is plural. Better to use the non-plural form everywhere.

mkosi/machine.py
```diff
@@ -122,9 +143,52 @@ def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -
...
+            raise unittest.SkipTest("Qemu test skipped due to environment variable.")
+
+        cls.machine.__enter__()
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        cls.machine.__exit__(*cls.exc_info)
+
+    def tearDown(self) -> None:
+        self.machine.reboot()
```
DaanDeMeyer, line 197: Hmm, so one realization I just had here is that calling reboot() in tearDown() means we'll unnecessarily boot the machine after the last test.

So maybe calling reboot() here was the wrong call from me, my bad. What we can do instead is make boot() and kill() public methods of the Machine class, and call kill() in tearDown() and boot() in setUp(). That's equivalent to rebooting but will avoid booting the machine unnecessarily after the last test finishes.

mkosi/machine.py
```diff
@@ -122,9 +143,55 @@ def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -
...
             self.exit_code = self._serial.wait()
             self._serial = None
         self.stack.__exit__(*args, **kwargs)
+
+    def __exit__(self, *args: Any, **kwargs: Any) -> None:
+        self._kill(*args, **kwargs)
+
+    def reboot(self) -> None:
+        self._kill(None, None, None)
+        self._boot()
```
DaanDeMeyer, line 158: Given the comment below that reboot() doesn't really fit for unittest since we end up unnecessarily booting the machine, let's remove the reboot() method here as well. While it works, I think we should leave the opportunity open to add a reboot() method in the future that does not throw away all the state of the image (in other words, it would keep the temporary copy of the image intact when --ephemeral is used and also keeps the copy of efivars intact when qemu is used, which we don't do today). Note that you don't need to implement this reboot() method now, we can add the new version of it when we need it.

mkosi/machine.py
```diff
@@ -122,9 +143,55 @@ def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -
 
         return process
 
-    def __exit__(self, *args: Any, **kwargs: Any) -> None:
+    def _kill(self, *args: Any, **kwargs: Any) -> None:
```
DaanDeMeyer, line 146: When you make kill() a public method like I suggested below, I'd also remove the `args` and `kwargs` arguments to make the method easier to use. We can switch things around and have kill() call `__exit__()` instead of `__exit__()` calling kill(). kill() can then call `__exit__()` with arguments `(None, None, None)` and we can then remove `args` and `kwargs` from kill()

DaanDeMeyer (changes requested):
Looking good, but I noticed one small flaw in the design which means we can't use reboot() but have to use separate kill() and boot() methods instead, so that's one thing that still needs to be cleaned up.

mkosi/machine.py
```diff
@@ -122,9 +143,55 @@ def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -
...
+        machine_verb = cls.machine.args.verb
+        no_nspawn = parse_boolean(os.getenv("MKOSI_TESTS_NO_NSPAWN", "0"))
+        no_qemu = parse_boolean(os.getenv("MKOSI_TESTS_NO_QEMU", "0"))
+
+        if no_nspawn and machine_verb == Verb.boot:
+            raise unittest.SkipTest("Nspawn test skipped due to environment variable.")
+        if no_qemu and machine_verb == Verb.qemu:
+            raise unittest.SkipTest("Qemu test skipped due to environment variable.")
+
+        cls.machine.__enter__()
```
DaanDeMeyer, line 190: Given that we'll start booting the machine in setUp(), we should avoid trying to boot the machine here as well.

I'd fix this by adding a public method build() that builds the machine and calling it here instead of `__enter__()`.

mkosi/machine.py
```diff
@@ -122,9 +143,55 @@ def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -
...
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        cls.machine.__exit__(*cls.exc_info)
```
DaanDeMeyer, line 194: Because we won't be using Machine() as a context manager anymore after we modify setUpClass() to use the new build public method, we shouldn't call `__exit__()` anymore either in `tearDownClass()`, which means we can just remove the `tearDownClass()` altogether since all the cleanup will already be done by calling the kill() method in `tearDown()`.

DaanDeMeyer (changes requested):

mkosi/machine.py
gsegatti: I'm unable to remove `args` and `kwargs` when running QEMU:

```
>               _fix_exception_context(new_exc_details[1], exc_details[1])
E               IndexError: tuple index out of range
```

Couldn't find a way to overcome this

DaanDeMeyer: What I'm proposing is the following:

```
def kill(self):
    self.__exit__(None, None, None)

def __exit__(self, *args, **kwargs):
    if self._serial:
        self._serial = None
    self.stack.__exit__(*args, **kwargs)
```

mkosi/machine.py
```diff
@@ -122,9 +146,50 @@ def run(self, commands: Sequence[str], timeout: int = 900, check: bool = True) -
 
         return process
 
+    def kill(self, *args: Any, **kwargs: Any) -> None:
+        self.__exit__(*args, **kwargs)
```
DaanDeMeyer, line 150: So I'd go ahead and remove the arguments from kill() just like we removed them from reboot() and pass None, None, None to `__exit__()` instead. We don't really use the exception information anyway so no need to complicate the signature of kill() so we can pass along exception information we don't use.

This would also mean we can get rid of `exec_info` in `MkosiMachineTest` and we wouldn't need to override `_callTestMethod` anymore either, which would be a nice simplification.

mkosi/machine.py
gsegatti: oh, I understood ((None, None, None)), this works

DaanDeMeyer (approved):
LGTM

PR #921, opened 2022-03-01
Running mypy in tests/

gsegatti:
Apparently `tests/` was not being reviewed by Mypy. Added the folder to the respective line on the CI and tried to correct most of the stuff pointed by Mypy.

Will leave a few comments because some of the errors are not necessarily crystal clear, a few things are still open issues so it seems.

Please feel free to leave suggestions on better approaches.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,44 +1,44 @@
 # SPDX-License-Identifier: LGPL-2.1+
 
+from typing import Any, Dict, List, Set
+
 from tests.test_config_parser import MkosiConfig
 
 
 class DictDiffer:
-    def __init__(self, expected_dict, current_dict):
+    def __init__(self, expected_dict: Dict[str, Any], current_dict: Dict[str, Any]) -> None:
         self.current_dict = current_dict
         self.expected_dict = expected_dict
         self.set_current, self.set_past = set(current_dict.keys()), set(expected_dict.keys())
         self.intersect = self.set_current.intersection(self.set_past)
 
     @property
-    def unexpected(self):
+    def unexpected(self) -> List[str]:
         return [f"{k}={self.current_dict[k]}" for k in self.set_current - self.intersect]
 
     @property
-    def missing(self):
+    def missing(self) -> List[str]:
         return [str(k) for k in self.set_past - self.intersect]
 
     @property
-    def invalid(self):
+    def invalid(self) -> List[str]:
         inva = {o for o in self.intersect if self.expected_dict[o] != self.current_dict[o]}
         return [f"{k}={self.current_dict[k]} (exp: {self.expected_dict[k]})" for k in inva]
 
     @property
-    def valid(self):
+    def valid(self) -> Set[str]:
         return {o for o in self.intersect if self.expected_dict[o] == self.current_dict[o]}
 
 
-def pytest_assertrepr_compare(op, left, right):
-    if not isinstance(left, MkosiConfig):
-        return
+def pytest_assertrepr_compare(op: str, left: MkosiConfig, right: Dict[str, Any]) -> List[str]:
     if not isinstance(right, dict):
-        return
+        return []
     for r in right.values():
         if not isinstance(vars(r), dict):
             return ["Invalid datatype"]
     if op == "==":
 
-        def compare_job_args(job, l_a, r_a):
+        def compare_job_args(job: str, l_a: Dict[str, Any], r_a: Dict[str, Any]) -> None:
             ddiff = DictDiffer(l_a, r_a)
             ret.append(f'Comparing parsed configuration {job} against expected configuration:')
             ret.append("unexpected:")
@@ -63,3 +63,4 @@ def pytest_assertrepr_compare(op, left, right):
             if not left_job in verified_keys:
                 ret.append(f'Missing job: {left_job}')
         return ret
+    return []
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -7,14 +7,14 @@ import mkosi.backend as backend
 from mkosi.backend import Distribution, PackageType, set_umask
 
 
-def test_distribution():
+def test_distribution() -> None:
     assert Distribution.fedora.package_type == PackageType.rpm
     assert Distribution.fedora is Distribution.fedora
-    assert Distribution.fedora is not Distribution.debian
+    assert Distribution.fedora.package_type is not Distribution.debian.package_type
     assert str(Distribution.photon) == "photon"
 
 
-def test_set_umask():
+def test_set_umask() -> None:
     with set_umask(0o767):
         tmp1 = os.umask(0o777)
         with set_umask(0o757):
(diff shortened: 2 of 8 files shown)


tests/test_config_parser.py
```diff
@@ -5,12 +5,15 @@
...
+from typing import TYPE_CHECKING, Any, Dict, Generator, List, Mapping, Optional, Type
 
 import pytest
 
 import mkosi
-from mkosi.backend import Verb
+from mkosi.backend import MkosiException, OutputFormat, SourceFileTransfer, Verb
+
+if TYPE_CHECKING:
+    __class__ : Type[Any]
```
gsegatti, line 16: Added because of this: https://github.com/python/mypy/issues/4177

tests/test_config_parser.py
```diff
@@ -185,14 +190,14 @@ def _append_list(self, ref_entry, new_args, job_name=DEFAULT_JOB_NAME, separator
...
-    def write_ini(dname: str, fname: str, config: dict, prio=1000) -> None:
+    def write_ini(dname: str, fname: str, config: Dict[str, Any], prio: int = 1000) -> None:
         """Write mkosi.default(.d/*) files"""
         if not os.path.exists(dname):
             os.makedirs(dname)
         if prio < 1000:
             fname = f"{prio:03d}_{fname}"
         config_parser = configparser.RawConfigParser()
-        config_parser.optionxform = lambda optionstr: str(optionstr)
+        config_parser.optionxform = lambda optionstr: str(optionstr) # type: ignore
```
gsegatti, line 200: The discussed solutions seemed to be `# type ignore` and `setattr`

tests/test_config_parser.py
```diff
@@ -185,14 +190,14 @@ def _append_list(self, ref_entry, new_args, job_name=DEFAULT_JOB_NAME, separator
                 self.reference_config[job_name][ref_entry].append(arg)
 
     @staticmethod
-    def write_ini(dname: str, fname: str, config: dict, prio=1000) -> None:
+    def write_ini(dname: str, fname: str, config: Dict[str, Any], prio: int = 1000) -> None:
```
gsegatti, line 193: At some point we pass a `Dict[str, object] `, but at other it becomes `Dict[str, Dict[str, str]]`. Went with this approach, let me know if there's a better way.

tests/test_config_parser.py
```diff
@@ -205,7 +210,7 @@ def write_ini(dname: str, fname: str, config: dict, prio=1000) -> None:
         with open(os.path.join(dname, fname), "w") as f_ini:
             config_parser.write(f_ini)
 
-    def _update_ref_from_file(self, mk_config: dict, job_name=DEFAULT_JOB_NAME) -> None:
+    def _update_ref_from_file(self, mk_config: Dict[str, Any], job_name: str = DEFAULT_JOB_NAME) -> None:
```
gsegatti, line 213: Same as line [193](https://github.com/systemd/mkosi/pull/921/files#r816692193).

tests/test_config_parser.py
```diff
@@ -166,7 +171,7 @@ def __eq__(self, other: Mapping[str, Any]) -> bool:
                 is_eq = False
         return is_eq
 
-    def _append_list(self, ref_entry, new_args, job_name=DEFAULT_JOB_NAME, separator=","):
+    def _append_list(self, ref_entry: str, new_args: Any, job_name: str = DEFAULT_JOB_NAME, separator: str = ",") -> None:
```
gsegatti, line 174: `new_args` is either a `List[str]` or `str`. Went with `Any` like the [docs said](https://mypy.readthedocs.io/en/stable/common_issues.html#redefinitions-with-incompatible-types)

tests/test_config_parser.py
```diff
@@ -150,7 +155,7 @@ def add_reference_config(self, job_name=DEFAULT_JOB_NAME):
             "workspace_dir": None,
         }
 
-    def __eq__(self, other: Mapping[str, Any]) -> bool:
+    def __eq__(self, other: Mapping[str, Any]) -> bool: # type: ignore
```
gsegatti, line 158: This one was tricky. I tried doing it like [here](https://stackoverflow.com/questions/54801832/mypy-eq-incompatible-with-supertype-object):

```
def __eq__(self, other: object) -> bool:
    if not isinstance(other, Mapping[str, Any]):
        return NotImplemented
        [...]
```

But then Mypy says:
`error: Parameterized generics cannot be used with class or instance checks  [misc]`

tests/test_config_parser.py
```diff
@@ -468,7 +473,7 @@ class MkosiConfigDistro(MkosiConfigOne):
     - --distribution
     """
 
-    def __init__(self, subdir_name=None, alldir_name=None):
+    def __init__(self, subdir_name: str = "", alldir_name: str = "") -> None:
```
gsegatti, line 476: `alldir_name` is never used so maybe this can be removed.

tests/conftest.py
```diff
@@ -1,44 +1,44 @@
...
 
     @property
-    def valid(self):
+    def valid(self) -> Set[str]:
         return {o for o in self.intersect if self.expected_dict[o] == self.current_dict[o]}
 
 
-def pytest_assertrepr_compare(op, left, right):
-    if not isinstance(left, MkosiConfig):
-        return
```
gsegatti, line 33: Mypy said this `return` was unreachable after I added the Typing for the arguments. Not really sure though

tests/test_backend.py
```diff
@@ -7,14 +7,14 @@
 from mkosi.backend import Distribution, PackageType, set_umask
 
 
-def test_distribution():
+def test_distribution() -> None:
     assert Distribution.fedora.package_type == PackageType.rpm
     assert Distribution.fedora is Distribution.fedora
-    assert Distribution.fedora is not Distribution.debian
+    assert Distribution.fedora.package_type is not Distribution.debian.package_type
```
gsegatti, line 13: [Mypy dislikes Enum comparisons](https://github.com/python/mypy/issues/9005). Since the testing is quite trivial I just compared different package types to assure it is actually different.

tests/test_config_parser.py
DaanDeMeyer: Ignoring is fine here, but add a comment linking to that discussion so we know why we added the ignore in the future.

tests/test_config_parser.py
DaanDeMeyer: When something can be one of two types, you can use `Union[str, List[str]]` to limit the type to only those two types

tests/test_parse_load_args.py
```diff
@@ -44,13 +44,13 @@ def test_parse_load_verb():
     with pytest.raises(SystemExit):
         parse(["invalid"])
 
-def test_os_distribution():
+def test_os_distribution() -> None:
     for dist in Distribution:
         assert parse(["-d", dist.name]).distribution == dist
 
-    with pytest.raises((argparse.ArgumentError, SystemExit)):
+    with pytest.raises(tuple((argparse.ArgumentError, SystemExit))):
```
DaanDeMeyer, line 51: What error did you get here that mad you add the `tuple` keyword?

DaanDeMeyer (approved):
I don't wanna waste too much time on those overengineered config parser tests in test_confg_parser and conftest so I don't mind a few Any's in there to make it pass type checking.

I have one comment on the use of the tuple keyword aside from that it looks fine

gsegatti:
This is what I got without the tuple:

`error: Argument 1 to "raises" has incompatible type "Tuple[Type[ArgumentError], Type[SystemExit]]"; expected
"Union[Type[<nothing>], Tuple[Type[<nothing>], ...]]"  [arg-type]
        with pytest.raises((argparse.ArgumentError, SystemExit)):
                           ^
`

PR #923, opened 2022-03-04
Remove dracut unified kernel image kernel-install script

DaanDeMeyer:
Let's simplify and remove the script and instead call dracut directly.
In many of the scenarios we support now, the script is useless after
we've built the image since it requires extra environment variables to
be set that are only set when building the image and not when the script
is executed by kernel-install after building the image.

Instead of trying to make unified kernel images work as a kernel script.
Let's assume that users that want a unified kernel image are building
"golden" images and will update by distributing a newly built version
of the image instead of trying to update using conventional means.


behrmann:
I'm using this script on my desktop - well, a slightly outdated (without `IMAGE_ID`) stripped down version (without the `ROOTHASH`/`USRHASH`/`PARTLABEL` block and the block checking for `IMAGE_VERSION` stripped down to the else path).

I'm not opposed to this change. I think it jives with the whole not building initrds locally, but centrally building them, and you are of course right that those environment variables are missing and of course the hashes will change in a system that updates, since it's not immutable, so the whole thing is a bit suspect, and of course this whole thing only ever worked for distributions where one can use `kernel-install` (thus disqualifying the whole Debian family). That being said, it does work for mutable systems, which with this change would need to supply their own `kernel-install` script, so a more minimal solution might be to just delete the script at the end, when we are in one of the `IMAGE_ID`/verity situations.

DaanDeMeyer:
> That being said, it does work for mutable systems, which with this change would need to supply their own kernel-install script.

I figure keeping this out of mkosi and having users provide their own script is probably preferable in this case. I generally want to try to reduce the number of use cases we support a bit rather than add more unless they add a lot of value. This stuff is the kind of thing that is super iffy, doesn't work on half the distros and only is the very specific case where you have a mutable system.

mkosi-initrd will also arrive eventually and I figure when that happens we'll want to stop using dracut altogether so this seems like a good move in that direction as well since we hide the internals of UKI generation from the generated image more.

DaanDeMeyer:
First commit inspired me to look at the initrd mess again and I came up with some more stuff. Not done yet so marking as draft.

Turns out, generating unified kernel images ain't all that difficult, and we really don't need dracut to do it for us. Dropping down a level and calling objcopy directly means we can generate a unified kernel image from any initrd, it doesn't have to be generated by dracut. This in turn means we don't need to use dracut anymore on debian like systems and we can switch back to mkinitcpio on Arch as well. Same applies for other distros. The only thing we need to know is the location of the kernel image and the generated initrd and that's sufficient to generate the unified kernel image.

The end result of this is that we'd end the dracut monopoly and just let every distro use its preferred initrd generator (at least until mkosi-initrd comes along). Using dracut everywhere requires a ton of customizations on distros where it's not the default and means extra work for users that now have to learn something that's custom to mkosi images compared to the distro they're used to.

DaanDeMeyer:
Uploaded a slightly less radical version, we stick to dracut because it generates initrds that use systemd by default, which isn't the case for the other generators. Instead of calling dracut to generate the unified kernel image, we call objcopy directly. This allows us to reuse the initrd generated in the install step which is cached which means incremental builds of bootable images become a lot faster.

I also added calls to kernel-install for opensuse, debian and ubuntu in the install step to make sure --without-unified-kernel-images works on these distros as well.

Conan-Kudo:
> end the dracut monopoly

Please don't use phrasing like this, it's insulting to other projects. Of all projects, systemd should be keenly aware of how much of a problem this kind of language causes.

Also, as one of the upstream dracut developers, I'm not terribly happy with the attitude being taken here about dracut. 

> at least until mkosi-initrd comes along

And statements like this are hypocritical. So please stay away from such dangerous value judgments. We're all trying to work together to make good FOSS systems.

DaanDeMeyer:
> Also, as one of the upstream dracut developers, I'm not terribly happy with the attitude being taken here about dracut.

Apologies, I meant no harm. The main goal here is to make us a little less reliant on dracut and to speed up the overall mkosi image generation process. I'd be proposing this patch regardless of the initrd generator we used.

behrmann:
I like this version better than the first one, but I'm still a bit unsure about it. I definitely see the value of doing the objcopy to assemble the EFI file ourselves, but it does introduce a point where we might do things differently from dracut (or mkinitcpio if we were to do a version where distros use their preferred initrd generator, since mkinitcpio introduced generating EFI binaries a while ago). I'll have a closer look at this once I'm back from vacation next week.

@Conan-Kudo: Sorry for not noting that the wording was problematic. I know that Daan didn't mean anything by it, since the "monopoly" in mkosi is something he introduced by standardising all distros to use dracut (since it's available everywhere and at the time it looked like Arch would switch soon enough to dracut, since mkinitcpio was unmaintained at the time, so it might have been the default everywhere except for Debian anyway). I for one still think that it is worthwhile to mainly support dracut, since I'm greatly in favour of standardised stacks. It is also clear to me, that, while the idea of producing initrds directly from packages greatly appeals to me and is something I want to see happening, dracut will for the foreseeable future do a lot of things better -- at least out of the box, since it has amassed specialised solutions for most every conceivable boot scenario -- and be more broadly supported across distros (I mean, dracut is still not the default everywhere although it should have been years ago). I think, which initrd ends up being, would be the one thing where we would benefit most from pluggability.

Conan-Kudo:
For what it's worth, I would love to see dracut used in all distros by default, but that requires some direct advocacy that I don't think I could do alone. :)

mkosi/__init__.py
```diff
@@ -4170,45 +4162,59 @@ def install_unified_kernel(
...
-                    continue
+        for kver, kimg in gen_kernel_images(args, root):
+            if args.image_id:
+                image_id = args.image_id
+                if args.image_version:
+                    partlabel = f"{args.image_id}_{args.image_version}"
+                else:
+                    partlabel = f"{args.image_id}"
+            else:
+                image_id = "linux"
```
behrmann, line 4176: That fallback is maybe a bit generic. How about `mkosi` or `mkosi-$DISTRO`? (I know it's lifted from the current script, but I was thinking about the docu for `IMAGE_ID` in the yet unreleased version of the `/etc/os-release` man page)

mkosi/__init__.py
```diff
@@ -4170,45 +4162,59 @@ def install_unified_kernel(
...
+            elif partlabel:
+                boot_options = f"{boot_options} root=PARTLABEL={partlabel}"
+
+            osrelease = root / "usr/lib/os-release"
+            # Write to something in $BOOT because everything else is read-only.
+            cmdline = root / prefix / "cmdline"
+            cmdline.write_text(boot_options)
+            initrd = root / prefix / args.machine_id / kver / "initrd"
+
+            cmd: Sequence[PathString] = [
```
behrmann, line 4205: Could you add a todo comment here or maybe a cast? This is a mypy bug. It correctly identifies the type of `root` and `boot_binary` (and without them also `cmd`), but leaving either in result in `cmd` being wrongly identified it as `list[object]`. A cast might be preferable, since it will be flagged automatically, when in the future the cast is identified as unnecessary, but it will be a false positive in some CI pipeline on some random PR in the future.

behrmann (commented):
@DaanDeMeyer sorry for taking so long with this, I was hit by the "coming back from vacation whammie" last week. Hadn't had a time to run it, but LGTM, since the behaviour for the cmdline looks right and the integration tests pass, which is good. :)

I'm still somewhat hesitant, because this will run all kernel install snippets, which didn't happen before, but since the boot tests pass, this doesn't seem to have any negative repercussions. 

This should also definitely get a news snippet, so that people know what's up and what to change to keep the current behaviour, if they want that. In that sense it might also make sense to still ship the old kernel install script, that would currently be deleted, in the next release, so that people have an easier time copying it into their tree, but that would mostly be a quality of life improvement. 

Another pair of eyes on this would be good, though, so cc @keszybz.

DaanDeMeyer:
> This should also definitely get a news snippet, so that people know what's up and what to change to keep the current behaviour, if they want that. In that sense it might also make sense to still ship the old kernel install script, that would currently be deleted, in the next release, so that people have an easier time copying it into their tree, but that would mostly be a quality of life improvement.

Do you mean we keep it in the repo or we keep it in the final image? If the first, we can just link to the history in the NEWS note I think.

mkosi/__init__.py
DaanDeMeyer: I think mypy is actually correct here and our typing for run() is wrong. The typing for run() should be `Sequence[object]` since it accepts anything that we can call `str()` on.

For proper typing we should probably turn that from `Sequence[object]` to `Sequence[str]`. It'll be kinda verbose to call `str()` on every path but it'll be safer overall since the type checker can pass when we pass something that isn't a string.

behrmann:
> Do you mean we keep it in the repo or we keep it in the final image? If the first, we can just link to the history in the NEWS note I think.

I meant the former and such a link would be fine as well. The idea of keeping it in the repo for a release would make it easier to copy the file out, but it's mostly an irrelevant point as long we document the change and leave a pointer where to get the file.

DaanDeMeyer:
Added one more commit to fix the mypy typing error

mkosi/__init__.py
DaanDeMeyer: Regardless, let's fix this in a separate PR since we do this in more than one place

DaanDeMeyer:
Let's merge this one to unblock #926

PR #924, opened 2022-03-04
debian: Switch default release from unstable to testing

DaanDeMeyer:
Unstable is a bit too bleeding edge. Let's switch to testing which
is roughly 5 days behind unstable so we don't get affected by
temporary issues that make it to unstable.


bluca (approved):

DaanDeMeyer:
Debian CIs succeeded so let's merge

PR #925, opened 2022-03-07
Adding shell verb to Machine

gsegatti:


mkosi/__init__.py
```diff
@@ -7458,12 +7462,13 @@ def run_shell_cmdline(args: MkosiArgs) -> List[str]:
...
 
-    if args.cmdline:
+    if args.cmdline or commands:
         # If the verb is 'shell', args.cmdline contains the command to run.
         # Otherwise, the verb is 'boot', and we assume args.cmdline contains nspawn arguments.
         if args.verb == Verb.shell:
             cmdline += ["--"]
-        cmdline += args.cmdline
+        cmdline += commands or args.cmdline
+
```
gsegatti, line 7471: Unnecessary extra line, will remove it.

mkosi/machine.py
```diff
@@ -152,7 +157,9 @@ def __exit__(self, *args: Any, **kwargs: Any) -> None:
             self._serial.kill(signal.SIGTERM)
             self.exit_code = self._serial.wait()
             self._serial = None
-        self.stack.__exit__(*args, **kwargs)
+
+        if self.stack:
+            self.stack.__exit__(*args, **kwargs)
```
gsegatti, line 162: Using "shell" we'll call `_kill()` multiple times. Since `self.stack` is only initialised once, I thought it was better to preemptively check if it exists.

mkosi/machine.py
```diff
@@ -66,7 +68,7 @@ def __init__(self, args: Sequence[str] = [], debug: bool = False) -> None:
             tmp.hostonly_initrd = True
             tmp.netdev = True
             tmp.ssh = True
-        elif tmp.verb == Verb.boot:
+        elif tmp.verb in (Verb.shell, Verb.boot):
             pass
         else:
             die("No valid verb was entered.")
```
DaanDeMeyer, line 74: This can be simplified to 

```
elif tmp.verb not in (Verb.shell, Verb.boot):
    die(...)
```

mkosi/machine.py
DaanDeMeyer: `self.stack` is not optional so I'm not sure what this if is checking. Calling `__exit__()` on the exit stack will empty the exit stack so calling it more than once isn't an issue.

tests/test_machine.py
```diff
@@ -27,23 +27,23 @@ def test_wrong_command(self) -> None:
...
-        assert result.returncode == 2
+        assert result.returncode in (1, 2)
         assert "No such file or directory" in result.stderr
 
     def test_infinite_command(self) -> None:
         with pytest.raises(TimeoutExpired):
             self.machine.run(["tail", "-f", "/dev/null"], 2)
 
-
+@pytest.mark.skipif(os.getenv("MKOSI_TEST_DEFAULT_VERB") == "shell", reason="Shell never boots the machine.")
```
DaanDeMeyer, line 40: I think it's more robust to check the actual verb of the machine instead of checking the environment variable. You can check the verb inside the test and call pytest.skip() to skip the test if it is shell.

mkosi/__init__.py
```diff
@@ -7426,14 +7426,18 @@ def ensure_networkd(args: MkosiArgs) -> bool:
...
     if args.output_format in (OutputFormat.directory, OutputFormat.subvolume):
         target = f"--directory={args.output}"
     else:
         target = f"--image={args.output}"
 
     cmdline = ["systemd-nspawn", target]
 
+    # Redirecting output correctly when testing and stopping racing conditions to Machine instance.
+    if commands:
+        cmdline += ["--pipe", "--register=no", "--keep-unit"]
```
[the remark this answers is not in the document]
DaanDeMeyer: Yes, but we also need to consider that mkosi itself can be called from a script that isn't connected to a terminal, in which case we should also specify the --pipe option. 

If there's a script out there that runs mkosi using the subprocess module, currently, the --pipe option won't be added even though it should. We can fix this by adding the `pipe` option and initializing it with `not sys.stdout.isatty()`  when run_shell_cmdline() is called from run_verb().

tests/test_machine.py
```diff
@@ -27,23 +27,23 @@ def test_wrong_command(self) -> None:
 
         # Check = False to see if stderr and returncode have the expected values
         result = self.machine.run(["NonExisting", "Command"], check=False)
-        assert result.returncode in (203, 127)
+        assert result.returncode in (1, 127, 203)
 
         result = self.machine.run(["ls", "-"], check=False)
-        assert result.returncode == 2
+        assert result.returncode in (1, 2)
```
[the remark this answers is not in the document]
gsegatti: yep

mkosi/__init__.py
```diff
@@ -7426,13 +7426,18 @@ def ensure_networkd(args: MkosiArgs) -> bool:
...
     if args.output_format in (OutputFormat.directory, OutputFormat.subvolume):
         target = f"--directory={args.output}"
     else:
         target = f"--image={args.output}"
 
-    cmdline = ["systemd-nspawn", target]
+    cmdline = ["systemd-nspawn", target, "--register=no", "--keep-unit"]
+
+    # Redirecting output correctly when not running directly from the terminal.
+    console_arg = f"--console={'interactive' if not pipe else 'pipe'}"
```
DaanDeMeyer, line 7438: Should be '--pipe' instead of 'pipe'

DaanDeMeyer (changes requested):

mkosi/machine.py
```diff
@@ -178,6 +186,10 @@ def setUpClass(cls) -> None:
         cls.machine.build()
 
     def setUp(self) -> None:
+        # First 5 characters are always "test_", also replacing underscores which makes name invalid.
+        # Necessary for shell otherwise racing conditions to the disk image will happen.
+        if self.machine.args.verb == Verb.shell:
+            self.machine.args.hostname = self._testMethodName[:9].replace("_", "")
```
DaanDeMeyer, line 192: Maybe we can do this always, no need to do it for "shell" only

mkosi/__init__.py
```diff
@@ -7720,11 +7724,13 @@ def find_address(args: MkosiArgs) -> Tuple[str, str]:
...
         return run_ssh(args, commands, check, stdout, stderr, timeout)
-    else:
+    elif args.verb == Verb.boot:
         cmdline = ["systemd-run", "--quiet", "--wait", "--pipe", "-M", virt_name(args), "/usr/bin/env", *commands]
         with suppress_stacktrace():
             return run(cmdline, check=check, stdout=stdout, stderr=stderr, text=True, timeout=timeout)
-
+    else:
+        with suppress_stacktrace():
+            return run(run_shell_cmdline(args, True, commands), check=check, stdout=stdout, stderr=stderr, text=True, timeout=timeout)
```
DaanDeMeyer, line 7733: Just passing True is a bit unreadable, let's add the keyword here as well to explain things `pipe=True`

mkosi/machine.py
```diff
@@ -66,8 +69,8 @@ def __init__(self, args: Sequence[str] = [], debug: bool = False) -> None:
             tmp.hostonly_initrd = True
             tmp.netdev = True
             tmp.ssh = True
-        elif tmp.verb == Verb.boot:
-            pass
+        elif tmp.verb in (Verb.shell, Verb.boot):
+            tmp.ephemeral = True
```
DaanDeMeyer, line 73: Why not enable --epehemeral for "qemu"?

mkosi/machine.py
```diff
@@ -155,6 +160,12 @@ def __exit__(self, *args: Any, **kwargs: Any) -> None:
         self.stack.__exit__(*args, **kwargs)
 
 
+pytestmark = [
+    pytest.mark.integration,
+    pytest.mark.skipif(os.getuid() != 0, reason="Must be invoked as root.")
+]
+
+
```
DaanDeMeyer, line 168: We shouldn't put pytest specifics in the machine.py file. We should instead make sure we do the same pytest check in MkosiMachineTest as well so that we skip tests when uid != 0 even if pytest isn't installed.

mkosi/__init__.py
```diff
@@ -877,7 +877,6 @@ def copy_image_temporary(src: Path, dir: Path) -> BinaryIO:
 
         disable_cow(f.name)
         copy_file_object(source, f)
-
```
DaanDeMeyer, line 880: Spurious newline removed

DaanDeMeyer:
Needs a rebase and removal of suppress_stacktrace() from run_command_image()

DaanDeMeyer:
Needs a rebase and removal of suppress_stacktrace() from run_command_image()

mkosi/machine.py
```diff
@@ -129,9 +133,9 @@ def boot(self) -> None:
             # We use pexpect to boot an image that we will be able to interact with in the future.
             # Then we tell the process to look for the # sign, which indicates the CLI for that image is active.
             # Once we've build/boot an image the CLI will prompt "root@image ~]# ".
-            # Then, when pexpects finds the '#' it means we're ready to interact with the process.
+            # Then, when pexpects finds the ']#' it means we're ready to interact with the process.
             self._serial = pexpect.spawnu(cmd, logfile=None, timeout=240)
-            self._serial.expect("#")
+            self._serial.expect("]#")
```
DaanDeMeyer, line 138: This is the shell I get when I boot an opensuse image in qemu:

```
localhost:~ # 
```

mkosi/machine.py
```diff
@@ -178,6 +184,9 @@ def setUpClass(cls) -> None:
         cls.machine.build()
 
     def setUp(self) -> None:
+        # Replacing underscores which makes name invalid.
+        # Necessary for shell otherwise racing conditions to the disk image will happen.
+        self.machine.args.hostname = self._testMethodName.replace("_", "-")
```
DaanDeMeyer, line 189: _testMethodName is an internal method, why not use the `id()' method here? https://docs.python.org/3/library/unittest.html#unittest.TestCase.id

mkosi/machine.py
gsegatti: Yeah, I just wanted to see if it'd pass the current integration tests in the CI.

mkosi/machine.py
gsegatti: Just because it'd look better than` self.id().split(".")[3]` to get the name. But this is a bad practice, right? so maybe I should do it this way.

DaanDeMeyer (approved):

PR #926, opened 2022-03-07
Qemu linux boot

daandemeyer:
This allows using qemu's direct linux boot to boot images directly without involving a bootloader. This boots faster compared to going through UEFI but more importantly, it allows booting images of architectures that don't support UEFI  in qemu.

I wonder if we'd want to add "linux" as an option for "--boot-protocols" instead of gating all the build time changes behind the --qemu-linux-boot option. If we'd do that, we could also forgo adding boot loader partitions and all the related stuff if only "linux" is enabled in "--boot-protocols".

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -456,10 +456,14 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
 
 : Pick one or more boot protocols to support when generating a
   bootable image, as enabled with `Bootable=`. Takes a comma-separated
-  list of `uefi` or `bios`. May be specified more than once in which
+  list of `uefi`, `bios`, or `linux`. May be specified more than once in which
   case the specified lists are merged. If `uefi` is specified the
   `sd-boot` UEFI boot loader is used, if `bios` is specified the GNU
-  Grub boot loader is used. Use "!\*" to remove all previously added
+  Grub boot loader is used. If `linux` is specified, the kernel image, initrd
+  and kernel cmdline are extracted from the image and stored in the output
+  directory. When running the `qemu` verb and setting the `--qemu-boot` option
+  to `linux`, qemu will be instructed to do a direct Linux kernel boot using
+  the previously extracted files. Use "!\*" to remove all previously added
   protocols or "!protocol" to remove one protocol.
 
 `KernelCommandLine=`, `--kernel-command-line=`
@@ -1128,6 +1132,13 @@ a machine ID.
   use the current unit scope, instead of creating a dedicated transcient
   scope unit for the containers. This option should be used when mkosi is
   run by a service unit.
+
+`QemuBoot=`, `--qemu-boot=`
+
+: When used with the `qemu` verb, this option sets the boot protocol to be used
+  by qemu. Can be set to either `uefi`, `bios`, or `linux`. Note that a boot
+  procotol needs to be included in `BootProtocols=` when building the image for
+  it to be usable with this option.
 
 `Netdev=`, `--netdev`
 
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1820,8 +1820,10 @@ def prepare_tree(args: MkosiArgs, root: Path, do_run_build_script: bool, cached:
                 root.joinpath("boot").mkdir(mode=0o700)
                 # Make sure kernel-install actually runs when needed by creating the machine-id subdirectory
                 # under /boot. For "bios" on Debian/Ubuntu, it's required for grub to pick up the generated
-                # initrd.
-                if args.distribution in (Distribution.debian, Distribution.ubuntu) and "bios" in args.boot_protocols:
+                # initrd. For "linux", we need kernel-install to run so we can extract the generated initrd
+                # from /boot later.
+                if ((args.distribution in (Distribution.debian, Distribution.ubuntu) and "bios" in args.boot_protocols) or
+                    ("linux" in args.boot_protocols and "uefi" not in args.boot_protocols)):
                     root.joinpath("boot", args.machine_id).mkdir(mode=0o700)
 
             if args.get_partition(PartitionIdentifier.esp):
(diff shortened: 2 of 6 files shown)


lucasdemarchi:
@DaanDeMeyer  I like the idea of adding it to --boot-protocols. It would make much sense for testing kernel changes.

DaanDeMeyer:
Updated to add "linux" to --boot-protocols. We still allocate the ESP partition since kernel-install doesn't generate initrds otherwise. Renamed --qemu-linux-boot to --qemu-boot and it now takes one of the boot protocols from --boot-protocols. It's used to determine the boot protocol we instruct qemu to use. So you can build an image with all three boot protocols enabled and boot each individually with qemu.

bluca:
I like this, i have local hacks for the same purpose

behrmann:
I too have use for this, since I sometimes need to extract the initrd and kernel. The name "linux" for this is a bit generic (I didn't have an immediate idea what it meant), how about something like "external" or "extracted"?

Since I see some overlap, should this gone in after #923?

DaanDeMeyer:
This should go in after #923 indeed. I'm stealing the name from qemu itself, which calls this "Direct Linux Boot" (https://qemu-project.gitlab.io/qemu/system/linuxboot.html).  I don't think "external" or "extracted" would be all that obvious either.

behrmann:
Ok, if this is a mostly standard term, then nevermind.

DaanDeMeyer:
Rebased but this also depends on #1007 to work properly in all scenarios.

mkosi/__init__.py
```diff
@@ -4350,17 +4347,24 @@ def install_unified_kernel(
...
-                "--uefi",
-                "--kver", kver,
-                "--kernel-image", Path("/") / kimg,
-                "--kernel-cmdline", boot_options,
-                "--force",
+                "objcopy",
+                "--add-section", f".osrel={osrelease}",   "--change-section-vma", ".osrel=0x20000",
+                "--add-section", f".cmdline={cmdline}",   "--change-section-vma", ".cmdline=0x30000",
+                "--add-section", f".linux={root / kimg}", "--change-section-vma", ".linux=0x2000000",
+                "--add-section", f".initrd={initrd}",     "--change-section-vma", ".initrd=0x3000000",
```
keszybz, line 4362: FWIW, dracut varied this by arch: 0x3000000 by default, but 0x4000000 on arm64. Was there some discussion of this value?

mkosi/__init__.py
```diff
@@ -7275,6 +7279,21 @@ def setup_netdev(args: MkosiArgs, root: Path, do_run_build_script: bool, cached:
...
+
+    with complete_step("Generating initramfs images…"):
+        # Running kernel-install on Debian/Ubuntu doesn't regenerate the initramfs. Instead, we can trigger
+        # regeneration of the initramfs via "dpkg-reconfigure dracut". kernel-install can then be called to put
+        # the generated initrds in the right place.
+        if args.distribution in (Distribution.debian, Distribution.ubuntu):
+            run_workspace_command(args, root, ["dpkg-reconfigure", "dracut"])
+
+        for kver, kimg in gen_kernel_images(args, root):
+            run_workspace_command(args, root, ["kernel-install", "add", kver, Path("/") / kimg])
```
keszybz, line 7392: > `Path("/") / kimg`

That's ugly, and all the callers do this fixup. Please just make the path absolute in `gen_kernel_images()`.

mkosi.md
```diff
@@ -1129,6 +1133,13 @@ a machine ID.
   scope unit for the containers. This option should be used when mkosi is
   run by a service unit.
 
+`QemuBoot=`, `--qemu-boot=`
+
+: When used with the `qemu` verb, this option sets the boot protocol to be used
+  by qemu. Can be set to either `uefi`, `bios` or `linux`. Note that a boot
```
keszybz, line 1139: Same?

keszybz (commented):
Looks reasonable. I doubt we'd get this entirely correct on the first try, so I'd just merge this and apply fixes later if we find something is missing.

mkosi/__init__.py
DaanDeMeyer: Frankly, I don't immediately see a cleaner way to do this since we need to support referring to the kernel image in the context of the image rootfs and in the context of the host rootfs. There's 3 options:

1. Return the path relative to the image rootfs. This is what we do now. It means we have to prefix with the "anchor" whenever trying to use the path, either `/` or `root`.
2. Return the absolute path in the image rootfs. This means we only have to prefix the path when using it in the context of the host rootfs. However, joining 2 absolute paths in python is not exactly intuitive

```
>>> from pathlib import Path
>>> a = Path("/usr")
>>> b = Path("/abc")
>>> a / b
PosixPath('/abc')
```

3. Return the absolute path in the host rootfs. This means we have to transform the path when trying to use it in the image rootfs context. This comes down to something like the following:

```
Path("/") / kimg.relative_to(root)
```

Both 2 and 3 don't seem super intuitive to me, so it seems better to have `gen_kernel_images()` return a path relative to the image rootfs and either prepend `/` or `root` as needed depending on the context.

mkosi/__init__.py
DaanDeMeyer: Hmm, I just copied the original values from the systemd-stub manpage. I figure we can keep using the same ones for now. If we ever run into an issue building aarch64 UEFI images we can update the defaults.

mkosi/__init__.py
keszybz: OK.

keszybz (approved):

mkosi/__init__.py
```diff
@@ -4458,6 +4460,70 @@ def extract_unified_kernel(
...
+        kimgabs = None
+        initrd = None
+
+        for kver, kimg in gen_kernel_images(args, root):
+            kimgabs = root / kimg
+            initrd = root / prefix / args.machine_id / kver / "initrd"
+
+        if kimgabs is None:
+            die("No kernel image found, can't extract.")
+        assert initrd is not None
```
behrmann, line 4485:
```suggestion
        if kimgabs is None:
            die("No kernel image found, can't extract.")
        assert initrd is not None
```
That's a bit weird, no? For `kimgabs` we print a nice error and for `initrd` being `None` only an assertion error that would need to be figured out?

mkosi/__init__.py
```diff
@@ -4458,6 +4460,70 @@ def extract_unified_kernel(
     return f
 
 
+def extract_kernel_image_initrd(
+    args: MkosiArgs,
+    root: Path,
+    do_run_build_script: bool,
+    for_cache: bool,
+    mount: Callable[[], ContextManager[None]],
+) -> Tuple[Optional[BinaryIO], Optional[BinaryIO]]:
```
behrmann, line 4469: Could this maybe be made more precise like this?
```suggestion
) -> Union[Tuple[BinaryIO], BinaryIO], Tuple[None, None]]:
```

This looks more ugly now, but in the future (3.10+) this will would actually be nicely readable
```python
tuple[BinaryIO, BinaryIO] | tuple[None, None]
```
(while also possibly being more precise)

behrmann (commented):
One serving of LGTM with a few nits on the side. :)

PR #927, opened 2022-03-07
Centralize usage of suppress_stacktrace() and propagate returncode

DaanDeMeyer:
Let's centralize the usage of suppress_stacktrace() to run_verb()
and make sure we properly propagate the returncode of the command
that failed.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -7366,7 +7366,7 @@ def suppress_stacktrace() -> Iterator[None]:
         yield
     except subprocess.CalledProcessError as e:
         # MkosiException is silenced in main() so it doesn't print a stacktrace.
-        raise MkosiException(e)
+        raise MkosiException() from e
 
 
 def virt_name(args: MkosiArgs) -> str:
@@ -7469,8 +7469,7 @@ def run_shell_cmdline(args: MkosiArgs) -> List[str]:
 
 
 def run_shell(args: MkosiArgs) -> None:
-    with suppress_stacktrace():
-        run(run_shell_cmdline(args), stdout=sys.stdout, stderr=sys.stderr)
+    run(run_shell_cmdline(args), stdout=sys.stdout, stderr=sys.stderr)
 
 
 def find_qemu_binary() -> str:
@@ -7661,8 +7660,7 @@ def run_qemu_cmdline(args: MkosiArgs) -> Iterator[List[str]]:
 
 def run_qemu(args: MkosiArgs) -> None:
     with run_qemu_cmdline(args) as cmdline:
-        with suppress_stacktrace():
-            run(cmdline, stdout=sys.stdout, stderr=sys.stderr)
+        run(cmdline, stdout=sys.stdout, stderr=sys.stderr)
 
 
 def interface_exists(dev: str) -> bool:
@@ -7722,8 +7720,7 @@ def run_command_image(args: MkosiArgs, commands: Sequence[str], timeout: int, ch
         return run_ssh(args, commands, check, stdout, stderr, timeout)
     else:
         cmdline = ["systemd-run", "--quiet", "--wait", "--pipe", "-M", virt_name(args), "/usr/bin/env", *commands]
-        with suppress_stacktrace():
-            return run(cmdline, check=check, stdout=stdout, stderr=stderr, text=True, timeout=timeout)
+        return run(cmdline, check=check, stdout=stdout, stderr=stderr, text=True, timeout=timeout)
 
 
 def run_ssh_cmdline(args: MkosiArgs, commands: Optional[Sequence[str]] = None) -> Sequence[str]:
@@ -7767,8 +7764,7 @@ def run_ssh(
     stderr: _FILE = sys.stderr,
     timeout: Optional[int] = None,
 ) -> CompletedProcess:
-    with suppress_stacktrace():
-        return run(run_ssh_cmdline(args, commands), check=check, stdout=stdout, stderr=stderr, text=True, timeout=timeout)
+    return run(run_ssh_cmdline(args, commands), check=check, stdout=stdout, stderr=stderr, text=True, timeout=timeout)
 
 
 def run_serve(args: MkosiArgs) -> None:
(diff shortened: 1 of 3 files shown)


mkosi/backend.py
```diff
@@ -76,6 +76,10 @@ def roundup(x: int, step: int) -> int:
 class MkosiException(Exception):
     """Leads to sys.exit"""
 
+    def __init__(self, *args: Any, returncode: int = 1, **kwargs: Any) -> None:
```
behrmann, line 79:
```suggestion
    def __init__(self, *, returncode: int = 1) -> None:
```
I think keeping variable and keyword arguments apart is something we should enforce in our signatures (unfortunately we our minimum Python requirement doesn't yet support `/`). This would leave open adding any variable (or keyword) args later on andI think it's generally more readable (the constructor below can be called without arguments).

mkosi/__main__.py
```diff
@@ -24,8 +24,8 @@ def main() -> None:
                     run_verb(a)
             else:
                 run_verb(a)
-    except MkosiException:
-        sys.exit(1)
+    except MkosiException as e:
+        sys.exit(e.returncode)
```
behrmann, line 28: Not sure how valuable that information is, since we can't really expect from where this comes and this could be any number. Maybe this could print the failing cmdline (or binary as well).

behrmann (commented):
Looks sensible to me.

mkosi/__main__.py
DaanDeMeyer: The behavior I'm going for here is that when running "shell", "boot" or "qemu", mkosi behaves as an invisible wrapper around the underlying command we're actually running (systemd-nspawn or qemu). I don't think we should show any extra output by default either (unless we have to build the image of course).

Anyway, for "shell", propagating the return code seems useful, since you can then use the "shell" verb in scripting to run commands and check the returncode for additional failure information from the command you tried to run since nspawn propagates that information in that scenario.

mkosi/backend.py
DaanDeMeyer: Yeah, but I thought it'd still be useful to use MkosiException as a general purpose exception as well. If we don't pass any arguments to the constructor we lose that.

behrmann: Ah, right, then I'd go for 
```suggestion
    def __init__(self, *args: Any, *, returncode: int = 1, **kwargs: Any) -> None:
```
while meaning the not yet available
```suggestion
    def __init__(self, *args: Any, /, *, returncode: int = 1, **kwargs: Any) -> None:
```

mkosi/__main__.py
behrmann: Ah, I see, from that point of view this makes a lot of sense. How about putting the additional metainformation in the exception and only showing it when we actually do a build (or rather, as a first iteration, here only checking if the command is `build`)?

mkosi/backend.py
DaanDeMeyer: I get a pyright error telling me the keyword only argument separator isn't allowed after `*args` when trying to apply this suggestion.

mkosi/__main__.py
DaanDeMeyer: I'm not sure I understand what you're trying to say here. I'd start with just propagating exit codes since silencing the output will take a bit more thinking since there's quite a few useful warnings in the qemu path. Maybe we should just make sure we don't print any output on the "shell" path (which we already do) and leave the qemu output for what it is.

behrmann: Sorry for being unclear. What I meant was adding adding a `cmdline` or `args` attribute to `MkosiException` as well and in main, additionally to propagating the returncode, also printing the cmdline if we are not doing a build. Since that might be too cumbersome to decide right now (since some commands imply build), I was thinking something along the lines
```python
except MkosiException as e:
    if args.verb == "build"
        print(e.cmdline)
    sys.exit(e.returncode)
```

That being said, I just remembered, that maybe we don't need to decorate `MkosiException` with more attributes, that are already in the original exception. I'll add another comment at the appropriate point in the code.

mkosi/__init__.py
```diff
@@ -7366,7 +7366,7 @@ def suppress_stacktrace() -> Iterator[None]:
         yield
     except subprocess.CalledProcessError as e:
         # MkosiException is silenced in main() so it doesn't print a stacktrace.
-        raise MkosiException(e)
+        raise MkosiException(e, returncode=e.returncode)
```
behrmann, line 7369: If we chain the exceptions like this
```suggestion
        raise MkosiException(e) from e
```
we can do something like this in main:
```python
except MkosiException as e:
    cause = getattr(e, "__cause__", None)
    if cause:
        if isinstance(cause, CalledProcessError):
            if args.verb == "build":
                print(cause.args)
            sys.exit(cause.returncode)
    sys.exit(1)
```

This way we would reuse the contextual information we already have from the initial exception and MkosiException could be a thin wrapper around wherever we might use it.

mkosi/__main__.py
behrmann: Ah, now I get it. I had forgotten this wasn't in the build path (just looking on this on the go, so looking at the greater picture is a bit difficult). Nevertheless, grabbing the information from the initial exception as described below might still be worthwhile, your call. :)

mkosi/__init__.py
DaanDeMeyer: I'm still not 100% convinced. We tend to print errors when they happen and before all the finalizing messages get printed. This would change that behavior. I figure if we want to start printing cmdlines, we should do so right before we raise the exception and print all the finalizer messages instead of after all the finalizer messages. For now, let's merge this as is since this is a slightly different issue than the one we're trying to fix here.

257:
binary packages are out of date i imagine.
i will either update my repo (the one mkosi uses now) or point mkosi to a new experimental binary-hosting gentoo has now [[`read more here`]](https://dilfridge.blogspot.com/2021/09/experimental-binary-gentoo-package.html)

257:
updated my binpkgs for now...

257:
@DaanDeMeyer could you check if the build times are dropped? max i'm getting for gentoo is 15m for BIOS build.
https://github.com/systemd/mkosi/runs/5469386125?check_suite_focus=true

DaanDeMeyer:
I'll let you know if I notice them taking a long time again. Thanks for taking a look!

257:
> I'll let you know if I notice them taking a long time again. Thanks for taking a look!

np@all

PR #929, opened 2022-03-08
gentoo: add homed flag to pambase

257:
[`upstream bug`](https://bugs.gentoo.org/832167):
sys-apps/systemd[homed] should depend on sys-auth/pambase[homed]

### DIFF
--- a/mkosi/gentoo.py
+++ b/mkosi/gentoo.py
@@ -433,7 +433,12 @@ class Gentoo:
 
                 # sys-apps/systemd -pkcs11
                 # sys-apps/systemd importd lzma
+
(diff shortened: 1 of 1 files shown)


behrmann (approved):
LGTM if you say that is the proper solution to have homed support in Gentoo, you are the resident Gentoo person. The build failures are unrelated, since they only occured in distros that are not touched by this.

257:
`homed` support has always been there, and why not, it's part of `systemd` :)
however, `pambase` stopped building with `homed` support. see the upstream bug i referenced in comments for more info.

PR #930, opened 2022-03-10
Merging CI tests with MkosiMachineTest class

gsegatti:


.github/workflows/ci.yml
```diff
@@ -96,40 +96,6 @@ jobs:
         sudo apt-get update && sudo apt-get install --no-install-recommends shellcheck
         bash -c 'set globstar; shellcheck mkosi/**/*.sh'
 
-  python-integration-test:
```
gsegatti, line 99: Actually we may still need these to test changes to the Machine class?

mkosi/machine.py
```diff
@@ -171,7 +171,10 @@ def setUpClass(cls) -> None:
         if os.getuid() != 0:
             raise unittest.SkipTest("Must be invoked as root.")
 
-        cls.machine = Machine(cls.args)
+        try:
+            cls.machine = Machine(cls.args)
+        except MkosiException as exception:
+            raise unittest.SkipTest(str(exception))
```
gsegatti, line 177: I removed the `if`'s from ci.yml and passed the invalid combinations of distribution and format to Mkosi anyways. __init__.py raises an exception if such combination is invalid, then, we skip.

mkosi/backend.py
```diff
@@ -828,9 +832,9 @@ def install_grub(args: MkosiArgs, root: Path, loopdev: Path, grub: str) -> None:
     run_workspace_command(args, root, cmdline, nspawn_params=nspawn_params)
 
 
-def die(message: str) -> NoReturn:
+def die(message: str, exception: Type[MkosiException] = MkosiException) -> NoReturn:
```
gsegatti, line 835: According to [PEP 484](https://www.python.org/dev/peps/pep-0484/#type-definition-syntax) "Expressions whose type is a subtype of a specific argument type are also accepted for that argument."

But, [this may break compatibility with python 2](https://stackoverflow.com/questions/4015417/why-do-python-classes-inherit-object/45062077#45062077).

gsegatti:
I'm unsure if the `die() `from `make_verity_sig` is supposed to raise an `InvalidArgs` exception

.github/workflows/ci.yml
```diff
@@ -236,103 +195,51 @@ jobs:
...
-
-    - name: Build ${{ matrix.distro }}/${{ matrix.format }} UEFI UKI
-      if: matrix.format != 'directory' &&
-          matrix.format != 'plain_squashfs' &&
-          matrix.format != 'tar' &&
-          matrix.format != 'cpio'
+    - name: Build/Boot ${{ matrix.distro }}/${{ matrix.format }}
+      run: MKOSI_TEST_DEFAULT_VERB=boot sudo python3 -m pytest -m integration tests
+
+    - name: Write shared mkosi config UEFI UKI
```
DaanDeMeyer, line 201: There's no real benefit to having this as a separate step I think. This will never fail so it just adds noise to have it as a separate step. You can merge it with the corresponding build/boot step

.github/workflows/ci.yml
```diff
@@ -236,103 +195,51 @@ jobs:
...
+
+    - name: Write shared mkosi config UEFI
       run: |
         tee mkosi.default <<- EOF
+        [Distribution]
+        Distribution=${{ matrix.distro }}
+
         [Output]
-        Bootable=yes
+        Format=${{ matrix.format }}
```
DaanDeMeyer, line 222: Yikes, nice find, turns out all these steps were running under the default format instead of the specified one

mkosi/__init__.py
```diff
@@ -66,6 +66,7 @@
 from .backend import (
     _FILE,
     ARG_DEBUG,
+    ArgsException,
```
DaanDeMeyer, line 69: I'd name this `MkosiNotSupportedException` to make it very clear we only use this for unsupported combinations of options.

mkosi/__init__.py
```diff
@@ -3670,7 +3671,7 @@ def compressor_command(option: Union[str, bool]) -> List[str]:
     elif option is False:
         return ["cat"]
     else:
-        die(f"Unknown compression {option}")
+        die(f"Unknown compression {option}", ArgsException)
```
DaanDeMeyer, line 3674: This should stay MkosiException. We only want to use ConfigException for invalid combinations of valid options. This is just wrong values being passed

mkosi/__init__.py
```diff
@@ -5689,7 +5690,7 @@ def load_distribution(args: argparse.Namespace) -> argparse.Namespace:
             args.release = r
 
     if args.distribution is None:
-        die("Couldn't detect distribution.")
+        die("Couldn't detect distribution.", ArgsException)
```
DaanDeMeyer, line 5693: Ditto. Nothing in parse_args() should throw MkosiNotSupportedException. All errors here should be fatal and not lead to the test being skipped.

mkosi/__init__.py
```diff
@@ -6205,7 +6206,7 @@ def load_args(args: argparse.Namespace) -> MkosiArgs:
     args.extra_search_paths = expand_paths(args.extra_search_paths)
 
     if args.cmdline and args.verb not in MKOSI_COMMANDS_CMDLINE:
-        die("Additional parameters only accepted for " + str([verb.name for verb in MKOSI_COMMANDS_CMDLINE])[1:-1] + " invocations.")
+        die("Additional parameters only accepted for " + str([verb.name for verb in MKOSI_COMMANDS_CMDLINE])[1:-1] + " invocations.", ArgsException)
```
DaanDeMeyer, line 6209: Ditto, we should not skip tests if we provide an invalid config

mkosi/__init__.py
```diff
@@ -6257,40 +6258,40 @@ def load_args(args: argparse.Namespace) -> MkosiArgs:
                 args.boot_protocols = ["bios"]
 
         if not {"uefi", "bios"}.issuperset(args.boot_protocols):
-            die("Not a valid boot protocol")
+            die("Not a valid boot protocol", ArgsException)
```
DaanDeMeyer, line 6261: Ditto, we shouldnt't skip if the boot protocol is invalid.

mkosi/__init__.py
```diff
@@ -6480,32 +6481,32 @@ def load_args(args: argparse.Namespace) -> MkosiArgs:
     if args.secure_boot or args.verity == "signed":
         if args.secure_boot_key is None:
             die(
-                "UEFI SecureBoot or signed Verity enabled, but couldn't find private key. (Consider placing it in mkosi.secure-boot.key?)"
+                "UEFI SecureBoot or signed Verity enabled, but couldn't find private key. (Consider placing it in mkosi.secure-boot.key?)", ArgsException
```
DaanDeMeyer, line 6484: Ditto, this is an error, and should be fatal and not skipped.

mkosi/__init__.py
```diff
@@ -6480,32 +6481,32 @@ def load_args(args: argparse.Namespace) -> MkosiArgs:
...
             )  # NOQA: E501
 
         if args.secure_boot_certificate is None:
             die(
-                "UEFI SecureBoot or signed Verity enabled, but couldn't find certificate. (Consider placing it in mkosi.secure-boot.crt?)"
+                "UEFI SecureBoot or signed Verity enabled, but couldn't find certificate. (Consider placing it in mkosi.secure-boot.crt?)", ArgsException
```
DaanDeMeyer, line 6489: Same here

DaanDeMeyer (changes requested):
Didn't annotate all of them but we should rename ArgsException to something more specific and only raise it when unsupported combinations of options are provided, not for every error that's possible in parse_args()/load_args().

With the current approach, passing an invalid value for --compression leads to the test being skipped. We don't want the test to be skipped in that case, we want it to fail. There's tons of similar examples that could happen with the current approach. Please limit the usage of the new exception to only errors where we're using invalid combinations of options.

mkosi/__init__.py
```diff
@@ -6552,7 +6553,7 @@ def load_args(args: argparse.Namespace) -> MkosiArgs:
         die("Sorry, --source-file-transfer-final=mount is not supported")
```
DaanDeMeyer, line 6553: Missed one here

mkosi/backend.py
DaanDeMeyer: I'm kinda surprised that works but OK

mkosi/__init__.py
gsegatti: I really thought this one wasn't the case, thanks

mkosi/machine.py
```diff
@@ -171,7 +171,10 @@ def setUpClass(cls) -> None:
         if os.getuid() != 0:
             raise unittest.SkipTest("Must be invoked as root.")
 
-        cls.machine = Machine(cls.args)
+        try:
+            cls.machine = Machine(cls.args)
```
DaanDeMeyer, line 175: So this approach means we won't build any of the images that aren't bootable anymore. In the CI we, first used the "build" verb and then used the "boot" or "qemu" verbs to boot the image in a second step.

What we can do to fix this is to move all the "not supported" checks involving verbs out of load_args() and into their respective run_xxx() functions. This means we'll still build the image but will raise MkosiNotSupported only when trying to boot the image.

DaanDeMeyer (changes requested):

mkosi/machine.py
```diff
@@ -64,16 +68,36 @@ def __init__(self, args: Sequence[str] = [], debug: bool = False) -> None:
...
+        # Now we can use the verb passed by the constructor/environment variable.
+        # Exceptions about combinations of options not being bootable won't be raised.
+        self.args.verb = v
+
+        # Some checks don't take the Verb into account, but rather the attributes below.
+        if v == Verb.qemu:
+            self.parsed_copy.bootable = True
+            self.parsed_copy.qemu_headless = True
+            tmp.bootable = True
+            tmp.qemu_headless = True
```
gsegatti, line 99: Just temporarily omitting the verb is not enough to built the image without raising exceptions in load_args() sometimes.

tests/conftest.py
```diff
@@ -32,7 +32,7 @@ def valid(self) -> Set[str]:
 
 def pytest_assertrepr_compare(op: str, left: MkosiConfig, right: Dict[str, Any]) -> List[str]:
     if not isinstance(right, dict):
-        return []
+        return [] # type: ignore
```
gsegatti, line 35: mypy suddenly started complaining this is unreachable

tests/test_machine.py
```diff
@@ -39,16 +39,18 @@ def test_infinite_command(self) -> None:
...
+        m = Machine()
+        if m.args.verb == Verb.shell:
+            pytest.skip("Shell never boots the machine.")
+        with pytest.raises(AssertionError):
+            m.run(["ls"])
 
 
 def test_after_shutdown() -> None:
-    with Machine() as m:
+    with pytest_skip_not_supported(), Machine() as m:
```
DaanDeMeyer, line 51: I'd put `Machine() as m` in a nested with block so it's more clear that the first context manager also applies to it.

DaanDeMeyer (changes requested):

mkosi/machine.py
```diff
@@ -63,8 +63,8 @@ def __init__(self, args: Sequence[str] = [], debug: bool = False) -> None:
         tmp.force = 1
         tmp.autologin = True
         tmp.ephemeral = True
+        tmp.bootable = True
```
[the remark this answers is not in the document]
gsegatti: yep, tested it today as well.

mkosi/machine.py
```diff
@@ -41,36 +42,45 @@ def __init__(self, args: Sequence[str] = [], debug: bool = False) -> None:
...
+        self.config.autologin = True
+        self.config.ephemeral = True
+        if self.config.verb == Verb.qemu:
+            self.config.bootable = True
+            self.config.qemu_headless = True
+            self.config.hostonly_initrd = True
+            self.config.netdev = True
+            self.config.ssh = True
+        elif self.config.verb not in (Verb.shell, Verb.boot):
+            die("No valid verb was entered.")
```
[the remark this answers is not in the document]
gsegatti: huh, I'm not sure where this code came from bc I'm not using `self.config`. Anyway I'll address this on the second PR where we separate build and boot, since for simply merging the CI only one call of `load_args()` is needed.

mkosi/machine.py
```diff
@@ -168,21 +178,19 @@ def __init_subclass__(cls, args: Sequence[str] = []) -> None:
...
+            no_qemu = parse_boolean(os.getenv("MKOSI_TEST_NO_QEMU", "0"))
 
-        if no_nspawn and verb == Verb.boot:
-            raise unittest.SkipTest("Nspawn test skipped due to environment variable.")
-        if no_qemu and verb == Verb.qemu:
-            raise unittest.SkipTest("Qemu test skipped due to environment variable.")
+            if no_nspawn and verb == Verb.boot:
+                raise unittest.SkipTest("Nspawn test skipped due to environment variable.")
+            if no_qemu and verb == Verb.qemu:
+                raise unittest.SkipTest("Qemu test skipped due to environment variable.")
```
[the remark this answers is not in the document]
gsegatti: For now we only need it at the __init__ method, since this PR only merges the CI tests.
Once we allow machines that can't be booted to be built there will be more usage of the `skip_not_supported` context manager.

mkosi/machine.py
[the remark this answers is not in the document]
DaanDeMeyer: Euhhh my bad this was a comment from a long time ago, feel free to ignore

DaanDeMeyer:
Commit message is out-of-date and needs to be updated

PR #932, opened 2022-03-24
Various fixes

DaanDeMeyer:


behrmann (approved):
LGTM.

.github/workflows/ci.yml
```diff
@@ -118,22 +118,8 @@ jobs:
           - directory
           - tar
           - cpio
-          - gpt_btrfs
+          - gpt_xfs
```
behrmann, line 121: Wouldn't `gpt_ext4` make as much sense there then? It might be a bit faster, too.

.github/workflows/ci.yml
```diff
@@ -198,8 +198,6 @@ jobs:
         [Output]
         BootProtocols=uefi
         WithUnifiedKernelImages=no
-        # Boot with systemd.volatile=overlay so squashfs images can write to the filesystem
-        KernelCommandLine=systemd.volatile=overlay
```
behrmann, line 216: Here, since I can't comment on the commit message: You mean `gpt_squashfs`, right? Since we still test plain squashfs.

DaanDeMeyer:
For some reason, the SSH connection is reset by the peer on Ubuntu: `kex_exchange_identification: read: Connection reset by peer`

behrmann:
@DaanDeMeyer the three CI errors look unrelated (CentOS doesn't like x86_64-v2, Rocky has a server error and Ubuntu boots to prompt to fail afterwards). This still LGTM, should I expect further changes or is this good to go?

DaanDeMeyer:
Well I figure we still need to do something about centos and ubuntu. centos is probably fixable once we get newer qemu when ubuntu 22.04 comes to github actions. ubuntu I'm kinda stuck on since I can't reproduce locally.

behrmann:
The Ubuntu one persists when CI is rerun? I would have expected it to be a random fluke. Centos we could mark xfail in pytest for the time being and watch for it become an xpass.

DaanDeMeyer:
So it seems to be the amount of SSH connections. The 4th SSH connection is guaranteed to fail with "connection reset by peer".

DaanDeMeyer:
I can spam commands to ssh as much as I want locally to a qemu booted ubuntu image though. I'm at a loss where the difference could be coming from.

DaanDeMeyer:
Hmm, adding a few sleeps also seems to fix the issue.

DaanDeMeyer:
Alright, this should finally make CI fully green. We had to add `RuntimeDirectoryPreserve=yes` to the ssh service unit on Ubuntu to make things work.

I did a bunch of other stuff as well while debugging this. Apologies for making it one huge PR let me know if I should split it up.

mkosi/__init__.py
```diff
@@ -6932,6 +6932,12 @@ def setup_ssh(
                               ListenStream=
                               ListenStream={args.ssh_port}
                               """)
+
+        add_dropin_config(root, "ssh@.service", "runtime-directory-preserve",
```
behrmann, line 6944: Great work debugging this! 

I'm a bit surprised seeing the template unit here, is that because of the `ssh.socket` unit we use? Socket activation for SSH is a bit of a niche thing (Arch stopped defaulting to that because of some potential DoS issue) and on a sufficiently new OpenSSH (8.4 in Debian bullseye seems to be enough) this could also be done with a config file in `/etc/ssh/sshd_config.d`, which are pulled in by the snippet at the top of
```
# /etc/ssh/sshd_config
Include /etc/ssh/sshd_config.d/*.conf
```
shipped by default in Debian.

behrmann (commented):
Great work! A few comments

mkosi/__init__.py
```diff
@@ -5734,14 +5734,11 @@ def load_distribution(args: argparse.Namespace) -> argparse.Namespace:
         args.distribution = Distribution[args.distribution]
 
     if args.distribution is None or args.release is None:
-        d, r = detect_distribution()
+        d, _ = detect_distribution()
 
         if args.distribution is None:
             args.distribution = d
 
-        if args.distribution == d and d != Distribution.clear and args.release is None:
```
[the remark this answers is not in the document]
DaanDeMeyer: This was also one of my attempts at fixing the ubuntu thing which turned out not to fix it. I can drop this and do it in a separate PR for more discussion

mkosi/__init__.py
```diff
@@ -2794,7 +2794,15 @@ def install_debian_or_ubuntu(args: MkosiArgs, root: Path, *, do_run_build_script
             if "VERSION_ID" not in os_release and "BUILD_ID" not in os_release:
                 f.write(f"BUILD_ID=mkosi-{args.release}\n")
 
-    invoke_apt(args, do_run_build_script, root, "install", extra_packages)
+    if args.distribution == Distribution.ubuntu:
+        updates = f"deb http://archive.ubuntu.com/ubuntu {args.release}-updates {' '.join(repos)}\n"
+    else:
+        updates = f"deb http://deb.debian.org/debian {args.release}-updates {' '.join(repos)}\n"
```
bluca, line 2800: note that updates != security (although the latter gets synced to the former eventually)

this is the security repo on bullseye (note format changed from this release):

```
deb https://deb.debian.org/debian-security bullseye-security main
```

bluca: (it's good to enable updates ofc)

mkosi/__init__.py
bluca: I enabled this - it seemed much nicer to use socket activation

behrmann: Ah, now I remember! :)

Looking at the Arch wiki, [this is the bug report](https://bugs.archlinux.org/task/62248) about the DoS issue. Apparently upstream used to ship the socket and template service units, but stopped with OpenSSH 8.0 due to the trigger limit. 

I very much like socket activation, but maybe that's something to go with the upstream default, which would also bring Debian in line with the rest. This is nothing, though, that this PR needs to do.

mkosi/__init__.py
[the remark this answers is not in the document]
behrmann: I think that would be best.

action/setup-github-actions.sh
```diff
@@ -100,12 +105,15 @@ fi
 
 meson install -C pacman-$PACMAN_VERSION-build
 
-if [ ! -f archlinux-keyring-$ARCHLINUX_KEYRING_VERSION.tar.gz ]; then
-    wget https://sources.archlinux.org/other/archlinux-keyring/archlinux-keyring-$ARCHLINUX_KEYRING_VERSION.tar.gz
-    tar xf archlinux-keyring-$ARCHLINUX_KEYRING_VERSION.tar.gz
+cargo install sequoia-sq --version $SEQUOIA_SQ_VERSION --features 'crypto-nettle compression-bzip2 autocrypt'
```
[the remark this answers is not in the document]
DaanDeMeyer: You can check out the gitlab repo where we get the sources. They've been working on something new and recently started using it as a replacement for the old thing.

mkosi/__init__.py
DaanDeMeyer: I figure we can add security as well; is there harm in adding it by default?

mkosi/__init__.py
DaanDeMeyer: @bluca You introduced usage of the socket activation for SSH when you added the ssh port option. I'm wondering if we shouldn't stick to the default usage of sshd instead. I also realized that it doesn't seem like we support usage of --ssh-port unless the socket unit on Debian/Ubuntu is used.

mkosi/__init__.py
bluca: we should, yes - with two caveats, one the format of the sources line changed in bullseye as mentioned, and two testing/unstable don't have a security repository, so apt update will fail.

So, before bullseye:

`deb http://security.debian.org/debian-security/ buster/updates main`

Bullseye and newer:

`deb https://security.debian.org/debian-security bullseye-security main`

And testing/unstable don't have it (and they also don't have the updates repository)

mkosi/__init__.py
bluca: As long as the port can be changed easily, that's fine - it seemed much more robust to use the socket unit, since support for drop-ins sshd config snippets is relatively new and not guaranteed to work on older ubuntu/debian releases, while the socket always works

mkosi/__init__.py
```diff
@@ -2817,6 +2817,9 @@ def install_debian_or_ubuntu(args: MkosiArgs, root: Path, *, do_run_build_script
         for kver, kimg in gen_kernel_images(args, root):
             run_workspace_command(args, root, ["kernel-install", "add", kver, Path("/") / kimg])
 
+    root.joinpath("etc/locale.conf").write_text("LANG=C.UTF-8\n")
```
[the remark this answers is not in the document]
DaanDeMeyer: Before the 2.35 release, C.UTF-8 has always been backported by distros (except Arch) so I'm pretty sure we can rely on it being there on Debian.

Only on Arch it has to be generated (which reminds me that I should fix that since we don't use C.UTF-8 on Arch atm).

action/setup-github-actions.sh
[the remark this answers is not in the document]
behrmann: I had seen keyringctl, but I had assumed it would be written in bash, as usual. Neat stuff!

mkosi/__init__.py
DaanDeMeyer: Added security repos to ubuntu/debian as well

mkosi/__init__.py
```diff
@@ -2792,7 +2794,25 @@ def install_debian_or_ubuntu(args: MkosiArgs, root: Path, *, do_run_build_script
...
+        if args.distribution == Distribution.ubuntu:
+            updates = f"deb http://archive.ubuntu.com/ubuntu {args.release}-updates {' '.join(repos)}"
+        else:
+            updates = f"deb http://deb.debian.org/debian {args.release}-updates {' '.join(repos)}"
+
+        root.joinpath(f"etc/apt/sources.list.d/{args.release}-updates.list").write_text(f"{updates}\n")
+
+        if args.distribution == Distribution.ubuntu:
+            security = f"deb http://archive.ubuntu.com/ubuntu {args.release}-security {' '.join(repos)}"
+        elif args.release in ("jessie", "stretch", "buster"):
```
behrmann, line 2807:
```suggestion
        elif args.release in ("stretch", "buster"):
```
Jessie is EOL

behrmann (commented):
LGTM, just one little thing.

mkosi/__init__.py
behrmann: For that matter, stretch will be fully out of long term support, too, in two months and I guess the finally, fully EOL.

bluca (approved):

mkosi/__init__.py
bluca: I'd leave stretch in, it will be used for a while yet, and it doesn't cost much as the workaround is needed for buster anyway

PR #933, opened 2022-03-24
Building "unbootable" images

gsegatti:
- #930 migrated the CI tests from using solely github actions to using the Machine class.
- Previously, the logic to see whether an image was bootable or not not was within ci.yml (please check systemd#930).
- Since migrating such logic to the Machine class, building and booting became a "single" step.
- Which meant, images that couldn't be booted were not even built. Since users may need the built image, this PR separates such processes.
- This is done by delaying the addition of some arguments (which are always added to make sure the Machine goes live).

### DIFF
--- a/mkosi/machine.py
+++ b/mkosi/machine.py
@@ -3,10 +3,12 @@
 from __future__ import annotations
 
 import contextlib
+import copy
 import os
 import signal
 import subprocess
 import unittest
+from argparse import Namespace
 from textwrap import dedent
 from typing import Any, Iterator, Optional, Sequence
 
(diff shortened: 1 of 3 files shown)


mkosi/machine.py
```diff
@@ -113,6 +126,14 @@ def __enter__(self) -> Machine:
         return self
 
     def boot(self) -> None:
+        # We restore the configurations for booting the image, now that it's built.
+        # So, if the image can't be booted, it'll actually fail during this step.
+        self.parsed.bootable = True
```
gsegatti, line 131: Have to place this inside the if below, when #932 lands.

PR #935, opened 2022-03-25
ci: Remove shellcheck step

DaanDeMeyer:
We don't have any scripts to check anymore.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,11 +91,6 @@ jobs:
         ./tools/generate-zipapp.sh
         ./builddir/mkosi -h
 
-    - name: Test shell scripts
-      run: |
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
Not sure how #923 got past CI since it should have failed on the shellcheck step.

DaanDeMeyer:
unit-test workflow passed so let's merge to fix CI

behrmann:
> We don't have any scripts to check anymore.

Not quite true, there's still `mkosi/resources/arch/kernel_{add,remove}.sh`, which are now unused, so those might be removed as well.

DaanDeMeyer:
So the glob wasn't working then? I'm confused

DaanDeMeyer:
https://github.com/systemd/mkosi/pull/936 fixes it

PR #938, opened 2022-03-26
Add missing environments to prepare und postinst scripts

behrmann:
Fixes: #937

PR #939, opened 2022-03-26
Make MkosiArgs.environment a dictionary

behrmann:
This is a possible followup to #938 (and contains its commit) to make `MkosiArgs.environment` a dictionary of environment variable names and their values instead of the list of strings it currently is.

I guess it would be nicer to do the handling as an argparse action, but having it, like it is now, in `load_args` is shorter.


DaanDeMeyer (approved):
CI failures are unrelated

Issue #934, opened 2022-03-24
bootstrapping a CentOS 7 results in unusable YUM and no rpm db


brancomat:
Hi, I might be missing something macroscopic, but I'm not able to create a CentOS 7 image with a working YUM/rpm database.

How to reproduce:
```
# mkosi -d centos -r 7 -o centos7.raw -p yum build
# systemd-nspawn -i centos7.raw
Spawning container centos7 on /moncic-ci/tmp/centos7.raw.
Press ^] three times within 1s to kill container.
-bash-4.2# rpm -qa
-bash-4.2# yum update
Loaded plugins: fastestmirror
Loading mirror speeds from cached hostfile


 One of the configured repositories failed (Unknown),
 and yum doesn't have enough cached data to continue. At this point the only
 safe thing yum can do is fail. There are a few ways to work "fix" this:

     1. Contact the upstream for the repository and get them to fix the problem.

     2. Reconfigure the baseurl/etc. for the repository, to point to a working
        upstream. This is most often useful if you are using a newer
        distribution release than is supported by the repository (and the
        packages for the previous distribution release still work).

     3. Run the command with the repository temporarily disabled
            yum --disablerepo=<repoid> ...

     4. Disable the repository permanently, so yum won't use it by default. Yum
        will then just ignore the repository until you permanently enable it
        again or use --enablerepo for temporary usage:

            yum-config-manager --disable <repoid>
        or
            subscription-manager repos --disable=<repoid>

     5. Configure the failing repository to be skipped, if it is unavailable.
        Note that yum will try to contact the repo. when it runs most commands,
        so will have to try and fail each time (and thus. yum will be be much
        slower). If it is a very temporary problem though, this is often a nice
        compromise:

            yum-config-manager --save --setopt=<repoid>.skip_if_unavailable=true

Cannot find a valid baseurl for repo: base/$releasever/x86_64
```

Tested on a Fedora 34. A rocky 8 image works fine but needs a `rpmdb --rebuilddb` before using `dnf` or `yum`.

spanezz:
#644 might be somewhat related

spanezz:
Likewise these two issues in Debian's BTS:

* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1008169
* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1004863

coling:
[...]

PR #940, opened 2022-03-29
Fix bootstrapping RPM distro on Debian

bluca:
The Debian rpm/dnf packages store the db in the home directory, so
the bootstrapped image has a broken rpm database.
Move it to the right place if it happens.

See: https://bugs.debian.org/1004863

Fixes https://github.com/systemd/mkosi/issues/934

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2012,6 +2012,22 @@ def invoke_dnf(
     with mount_api_vfs(args, root):
         run(cmdline)
 
+    distribution, release = detect_distribution()
+    if distribution not in (Distribution.debian, Distribution.ubuntu):
+        return
+
+    # On Debian, rpm/dnf ship with a patch to store the rpmdb under ~/
+    # so it needs to be copied back in the right location, otherwise
+    # the rpmdb will be broken. See: https://bugs.debian.org/1004863
+    rpmdb_home = root / "root/.rpmdb"
+    if rpmdb_home.exists():
+        # Take into account the new location in F36
+        rpmdb = root / "usr/lib/sysimage/rpm"
+        if not rpmdb.exists():
+            rpmdb = root / "var/lib/rpm"
+        unlink_try_hard(rpmdb)
+        shutil.move(cast(str, rpmdb_home), rpmdb)
+
 
 def install_packages_dnf(
     args: MkosiArgs,


mkosi/__init__.py
```diff
@@ -2012,6 +2012,13 @@ def invoke_dnf(
     with mount_api_vfs(args, root):
         run(cmdline)
 
+    # On Debian, rpm/dnf ship with a patch to store the rpmdb under ~/
+    # so it needs to be copied back in the right location, otherwise
+    # the rpmdb will be broken. See: https://bugs.debian.org/1004863
+    if root.joinpath("root/.rpmdb").exists():
+        unlink_try_hard(root / "var/lib/rpm")
+        shutil.move(cast(str, root / "root/.rpmdb"), root / "var/lib/rpm")
```
keszybz, line 2020: I think this hack should be conditionalized on being on Debian. Other people may have `/root/.rpmdb` for other reasons, and this shouldn't matter.

The database actually lives under `/usr/lib/sysimage/rpm/` now. (In opensuse since a while, in Fedora since ~now, https://fedoraproject.org/wiki/Changes/RelocateRPMToUsr). I have no idea how to figure out the right location automatically.

bluca: I can add more conditionals - but is it reasonable to have /root/.rpmdb in a freshly built image for reasons other than this debian thing?

bluca: added the conditionals, and checking if the new location exists

bluca:
> rpmd --rebuilddb

```
[root@image ~]# rpm -qa > /tmp/pre
[root@image ~]# rpmdb --rebuilddb
[root@image ~]# rpm -qa > /tmp/post
[root@image ~]# md5sum /tmp/pre /tmp/post
3565b2e3fedae8897bf8098ab76ab390  /tmp/pre
3565b2e3fedae8897bf8098ab76ab390  /tmp/post
```

I think it was mentioned in the bug simply as sa way to try and repair the database

davide125:
A better approach here could be to set the `_dbpath` macro for rpm (on Fedora 35, it defaults to `%{_var}/lib/rpm`). You could do that by passing `-D` to the rpm invocations, or by setting it in one of the macros files.

bluca:
> A better approach here could be to set the `_dbpath` macro for rpm (on Fedora 35, it defaults to `%{_var}/lib/rpm`). You could do that by passing `-D` to the rpm invocations, or by setting it in one of the macros files.

But every user of the image would have to do the same, no? Or do you mean at image build time?

davide125:
Yeah, I meant at image build time. The generated image would be fine, as it's only the `rpm` shipped on Debian that has this problem, not the one shipped in the target distributions.

bluca:
Is there an env var for that? Otherwise we'd need to change the system config files of the _host_ and that is not good. rpm is not called directly, it's called through yum/dnf.

keszybz:
I don't think this can be configured at the dnf level: at least the docs for dnf don't suggest any way.

Let's merge this. The workaround could be somewhat fragile, but it should work. Ideally, the stupid patch in Debian will be removed and then we can drop this again.

PR #943, opened 2022-03-30
Add --repository-directory option

DaanDeMeyer:
The repository directory can contain extra repositories to be
used when installing packages. If mkosi.reposdir/ exists, it's
used as the repository directory unless it's explicitly
specified.

Only supported for rpm based distros and Arch for now. We don't
support Ubuntu/Debian atm because we can't point apt to the
extra repositories directory because apt runs inside the image
compared to dnf and pacman which run outside of the image.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -314,13 +314,21 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
 
 `UseHostRepositories=`, `--use-host-repositories`
 
-: This option is only applicable for dnf-based distributions:
+: This option is only applicable for RPM-based distributions:
   *CentOS*, *Fedora Linux*, *Mageia*, *Photon*, *Rocky Linux*, *Alma Linux*
   and *OpenMandriva*.
-  Allows use of the host's existing dnf repositories.
-  By default, a hardcoded set of default dnf repositories is generated and used.
+  Allows use of the host's existing RPM repositories.
+  By default, a hardcoded set of default RPM repositories is generated and used.
   Use `--repositories=` to identify a custom set of repositories to be enabled
   and used for the build.
+
+`RepositoryDirectory`, `--repository-directory`
+
+: This option can (for now) only be used with RPM-based istributions and Arch
+  Linux. It identifies a directory containing extra repository definitions that
+  will be used when installing packages. The files are passed directly to the
+  corresponding package manager and should be written in the format expected by
+  the package manager of the image's distro.
 
 `Architecture=`, `--architecture=`
 
@@ -1507,6 +1515,10 @@ local directory:
   shall be built from the same working directory, as otherwise the
   build result of a preceding run might be copied into a build image
   as part of the source tree (see above).
+
+* The **`mkosi.reposdir/`** directory, if it exists, is automatically
+  used as the repository directory for extra repository files. See
+  the `RepositoryDirectory` option for more information.
 
 All these files are optional.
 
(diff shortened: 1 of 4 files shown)


behrmann:
For Debian we could bind mount the dir into the image and temporarily add a file to `/etc/apt/sources.list.d` that we clean up at the end.

DaanDeMeyer:
It gets more awkward when file paths such as gpg keys are specified in those files though, since you'd have to make sure all of those are available in the image at the correct location as well

DaanDeMeyer:
Pushed another commit that adds Ubuntu/Debian support. I'm not huge a fan though,

mkosi/backend.py
```diff
@@ -168,6 +168,19 @@ def __new__(cls, number: int, package_type: PackageType) -> Distribution:
...
+        Distribution.fedora,
+        Distribution.mageia,
+        Distribution.centos,
+        Distribution.centos_epel,
+        Distribution.openmandriva,
+        Distribution.rocky,
+        Distribution.rocky_epel,
+        Distribution.alma,
+        Distribution.alma_epel
+    ]
```
behrmann, line 182:
```suggestion
    return d in (
        Distribution.fedora,
        Distribution.mageia,
        Distribution.centos,
        Distribution.centos_epel,
        Distribution.openmandriva,
        Distribution.rocky,
        Distribution.rocky_epel,
        Distribution.alma,
        Distribution.alma_epel
    )
```
not that it really matters, but tuples are faster.

mkosi/__init__.py
```diff
@@ -2765,7 +2767,22 @@ def install_debian_or_ubuntu(args: MkosiArgs, root: Path, *, do_run_build_script
             if "VERSION_ID" not in os_release and "BUILD_ID" not in os_release:
                 f.write(f"BUILD_ID=mkosi-{args.release}\n")
 
-    invoke_apt(args, do_run_build_script, root, "install", extra_packages)
+    if args.repos_dir:
+        for path in glob.glob(f"{args.repos_dir}/*"):
+            filename = Path(path).name
+            if root.joinpath(f"etc/apt/sources.list.d/{filename}").exists():
+                die(f"Repository file {filename} already exists in image")
+            shutil.copy(path, root / "etc/apt/sources.list.d")
```
behrmann, line 2775: This won't really work like this, we'd need to mount the repos (or copy their conents) from `args.repos_dir` somewhere in the image, e.g. `/var/lib/mkosi` and add files in `/etc/sources.list.d` of the form 
```
deb file:/var/lib/mkosi/foo {args.release} main contrib non-free
deb-src file:/var/lib/mkosi/foo {args.release} main contrib non-free
```
for some repo `foo/` below `args.repodir`.

behrmann (commented):
I know what you mean. I didn't mean to pressure you into adding this right now. I think it's fine to add just the first commit of the series now and add this later.

Otherwise LGTM

DaanDeMeyer:
Dropped the 2nd commit for now.

mkosi.md
```diff
@@ -314,14 +314,22 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
...
-  Allows use of the host's existing dnf repositories.
-  By default, a hardcoded set of default dnf repositories is generated and used.
+  Allows use of the host's existing RPM repositories.
+  By default, a hardcoded set of default RPM repositories is generated and used.
   Use `--repositories=` to identify a custom set of repositories to be enabled
   and used for the build.
 
+`RepositoryDirectory`, `--repository-directory`
+
+: This option can only be used with RPM-based istributions and Arch Linux. It
```
behrmann, line 327:
```suggestion
: This option can at the moment only be used with RPM-based istributions and Arch Linux. It
```

PR #945, opened 2022-04-04
centos: Write stream dnf variable file when a stream release is used

DaanDeMeyer:
CentOS Stream SIGs use the $stream variable in their repo files to
identify the current CentOS Stream version. Let's make sure we write
this file so that SIG repo files can be copied into mkosi.reposdir
as is without requiring modifications to make them work.

Example SIG repo file using $stream:
https://git.centos.org/rpms/centos-release-hyperscale/blob/c8s-sig-hyperscale/f/SOURCES/CentOS-Stream-Hyperscale.repo


behrmann (approved):
LGTM.

PR #951, opened 2022-04-20
do not symlink outside of image root

behrmann:
Not sure about this, but I saw the failure in systemd/systemd#23134 and I think the symlink might try reaching out of the image.


DaanDeMeyer:
I just double checked, it doesn't point outside of the image. I kinda like the relative symlink more though, so let's merge this regardless.

DaanDeMeyer:
Needs a rebase

PR #952, opened 2022-04-20
mkosi: Force /etc/default/locale to be a symlink to locale.conf

DaanDeMeyer:
If the locales package is installed, /etc/default/locale will exist
in the filesystem already. Let's make sure we handle that properly.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2836,6 +2836,10 @@ def install_debian_or_ubuntu(args: MkosiArgs, root: Path, *, do_run_build_script
             run_workspace_command(args, root, ["kernel-install", "add", kver, Path("/") / kimg])
 
     root.joinpath("etc/locale.conf").write_text("LANG=C.UTF-8\n")
+    try:
+        root.joinpath("etc/default/locale").unlink()
+    except FileNotFoundError:
+        pass
     root.joinpath("etc/default/locale").symlink_to("/etc/locale.conf")
 
 


behrmann:
LGTM.

Only one more minor version bump to Python 3.8 and we can use `missing_ok=True` on the `unlink` method. :)

mkosi/__init__.py
```diff
@@ -2836,6 +2836,10 @@ def install_debian_or_ubuntu(args: MkosiArgs, root: Path, *, do_run_build_script
             run_workspace_command(args, root, ["kernel-install", "add", kver, Path("/") / kimg])
 
     root.joinpath("etc/locale.conf").write_text("LANG=C.UTF-8\n")
+    try:
```
bluca, line 2839: wouldn't it be better to check if it exists, and avoid overwriting it? So that the distro configuration is what gets applied to the image?

DaanDeMeyer: At least on Ubuntu Focal, /etc/default/locale is populated as follows:

```
root@image:~# cat /etc/default/locale 
#  File generated by update-locale
```

which doesn't seem exactly useful

behrmann: Isn't `/etc/default/locale` rewritten all the time when one does `dpkg-reconfigure locales`? Does debconf have any problems when the file is a symlink? 

It's been a while I last checked the patches for localectl in Debian, but I vaguely remember those using `/etc/default/locale` and `/etc/default/keyboard` in addition to what localectl uses by upstream. 

If the symlink doesn't pose a problem, I think it would be preferable to have a single file instead of two and to keep Debianisms to what's absolutely necessary.

behrmann:
@bluca What do you say? Is this good to go?

bluca (approved):

PR #953, opened 2022-04-21
Apply args.environment to all run() and run_workspace_command() calls

DaanDeMeyer:
Let's make sure every command we run in or outside of image we're
building inherits the environment variables specified in the Environment
option.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3380,8 +3380,6 @@ def nspawn_params_for_build_sources(args: MkosiArgs, sft: SourceFileTransfer) ->
     else:
         params += ["--chdir=/root"]
 
-    params += [f"--setenv={env}={value}" for env, value in args.environment.items()]
-
     return params
 
 
@@ -3441,8 +3439,7 @@ def run_postinst_script(
 
         run_workspace_command(args, root, ["/root/postinst", verb],
                               network=(args.with_network is True),
-                              nspawn_params=nspawn_params,
-                              env=args.environment)
+                              nspawn_params=nspawn_params)
         root_home(args, root).joinpath("postinst").unlink()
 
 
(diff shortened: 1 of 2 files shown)


behrmann:
Sorry, somehow I forgot about this one. I thought it had been merged already.

Only had a quick look. Is my understanding correct, that you populate `os.environ` from `args.env`, so that `run` always sees everything from `args.env` (and `env` passed as parameter at the call site) as well? With the rest being some cleanup so that environments are always set from `args.env` and some `env` parameter? If so, then I think that should be fine. I don't think this should provide any footguns that weren't there before and it makes the code cleaner.

behrmann:
I haven't thought too deeply about this, though. It might warrant a news entry.

DaanDeMeyer:
> Only had a quick look. Is my understanding correct, that you populate os.environ from args.env, so that run always sees everything from args.env (and env passed as parameter at the call site) as well? With the rest being some cleanup so that environments are always set from args.env and some env parameter? If so, then I think that should be fine. I don't think this should provide any footguns that weren't there before and it makes the code cleaner.

Yes, I'll add a comment to explain that when we copy the stuff to os.environ. I'll also add a NEWS entry and I'll extend the documentation for the option.

keszybz:
> Only had a quick look. Is my understanding correct, that you populate `os.environ` from `args.env`, so that `run` always sees everything from `args.env` (and `env` passed as parameter at the call site) as well? With the rest being some cleanup so that environments are always set from `args.env` and some `env` parameter? If so, then I think that should be fine. I don't think this should provide any footguns that weren't there before and it makes the code cleaner.

IMO this should be part of the commit message.

DaanDeMeyer:
> I don't think this is a change for the better, because it removes a separation that is generally clean and useful. What is the justification for the change?

I was thinking about how mkosi-initrd would set `KERNEL_INSTALL_BYPASS=1`, which wouldn't work at the moment because the environment setting isn't applied to the environment when we execute the distro's package manager. So I figured we should apply the environment to those run() calls so that we can influence them by setting environment variables. I took that reasoning a bit further and figured every command we run might benefit from this which led to the current PR.

It might be the wrong approach though because it allows for quite a bit of interference with how we execute commands. It might be better to set KERNEL_INSTALL_BYPASS conditionally in mkosi itself (it'll only work on distros that have recent kernel-install though)

DaanDeMeyer:
Sounds good let's close this

PR #955, opened 2022-04-22
Ensure we use C.UTF-8 as the default locale on all distributions

DaanDeMeyer:


behrmann:
needs a rebase due to the relative symlink change, besides that LGTM.

mkosi/__init__.py
```diff
@@ -2835,7 +2848,8 @@ def install_debian_or_ubuntu(args: MkosiArgs, root: Path, *, do_run_build_script
         for kver, kimg in gen_kernel_images(args, root):
             run_workspace_command(args, root, ["kernel-install", "add", kver, Path("/") / kimg])
 
-    root.joinpath("etc/locale.conf").write_text("LANG=C.UTF-8\n")
+    # Debian/Ubuntu use a different path to store the locale so let's make sure that path is a symlink to
+    # etc/locale.conf.
```
behrmann, line 2852: Could at this point the symlink be dangling? Some software might not like it it has no `/etc/default/locale` at all.

PR #956, opened 2022-04-22
load skeletons after running debootstrap for debian and ubuntu

behrmann:
Right now `mkosi.skeleton` cannot be used for dpkg-based distributions, since debootstrap will not work on a non-empty target. 

This adds a parameter to `install_skeleton_trees` to hack around this for Debian and Ubuntu, so that the call before `install_distribution` is skipped and we only add skeletons before invoking apt again after doing the initial debootstrap.

/cc @bluca


DaanDeMeyer:
Seems we still have a race condition with the SSH stuff.  We're getting "connection refused" in a few of the CIs now.

behrmann:
Indeed, although this shouldn't have anything to do with this PR, since the CI failures are in Arch and Alma, but this only touches Debian and Ubuntu directly.

PR #958, opened 2022-04-26
Tweaks to argparse setup to make --help output nicer

keszybz:


mkosi/__init__.py
```diff
@@ -5001,6 +5006,20 @@ def _format_action_invocation(self, action: argparse.Action) -> str:
         args_string = self._format_args(action, default)
         return ", ".join(action.option_strings) + " " + args_string
 
+    def _split_lines(self, text: str, width: int) -> List[str]:
```
behrmann, line 5009: Am I missing something or is this function unused?

mkosi/__init__.py
```diff
@@ -5184,10 +5205,15 @@ def create_parser() -> ArgumentParserMkosi:
...
-        "--repositories", action=CommaDelimitedListAction, default=[], help="Repositories to use", metavar="REPOS"
+        '--repositories',
+        metavar='REPOS',
+        action=CommaDelimitedListAction,
+        default=[],
+        help='Repositories to use',
     )
     group.add_argument(
-        "--use-host-repositories",
+        '--use-host-repositories',
```
behrmann, line 5215: Just as a stand in for the others: Is this really necessary? That's an awful lot of churn.

mkosi/__init__.py
```diff
@@ -174,6 +176,9 @@ def wrapper(*args: Any, **kwargs: Any) -> Dict[T, V]:
     return functools.update_wrapper(wrapper, f)
 
 
+def format_list(seq: Iterator[str]) -> str:
```
behrmann, line 179: I appreciate the little saver, but the name is a bit generic.

mkosi/__init__.py
keszybz: In general, single quotes are the standard for Python. I'm adding new lines with single quotes, and it'd be awkward to have a lines with double quotes and single quotes in a single invocation.

mkosi/__init__.py
behrmann: I just wanted to ask how that was picked up, but then I finally saw the `self` and realised it wasn't a freestanding function, but a method. Nevermind then.

mkosi/__init__.py
behrmann: I wouldn't go as far as saying they are standard. 
> In Python, single-quoted strings and double-quoted strings are the same. This PEP does not make a recommendation for this. Pick a rule and stick to it. When a string contains single or double quote characters, however, use the other one to avoid backslashes in the string. It improves readability.

With the large number of projects adopting black, I would say that double quotes are becoming ever more prevalent, but I'm not going to start an argument about this, since I know we disagree on this. :) That being said, I agree that having different quotes on lines directly following each other would be weird, but the diffstat would be smaller if they added lines had the same quotes as the ones around them.

keszybz: Yeah, but then I'd have to add lines with double quotes, and I'd prefer not to do that ;)

behrmann: I know. [But look how much cleaner the diffstat could be](https://github.com/behrmann/mkosi/commit/45e561cd6dbaa6466fe5da0c1bf0be8a63d30d3c) ;)

DaanDeMeyer: Just throwing out my preference for double quotes. On the other hand, I don't care at all as long as we primarily use one over the other.

behrmann:
The goofing around quotes not withstanding, this is a good change, that makes the help output *much* more usable. Two points for further discussion, though:
1. The options that are now nicely indented are still in single quotes. I think this would be more readable still, if we could get rid of them. 
2. We are properly bound by CLI backwards compat to accept something like `--bootable yes` to `--bootable`, but it's a bit weird as CLIs go, no? Is this something we should advertise, i.e. should we show a metavar at all? Also, should we maybe use something like `BooleanOptionalAction` to accept `--no-bootable` in addition (I thought we already did, but I haven't found it in the code when I was just having a quick look for it)? I think that would be a lot more conventional CLI.

DaanDeMeyer:
@keszybz ping we'd like to get this into the next release if possible

mkosi/__init__.py
keszybz: I think the priority of single quotes over double quotes was fairly well established as pythonic. Old manuals generally said "use single unless you have a reason to use double". Python code was written in a minimalistic way, much more than nowadays. Hence it was natural to prefer a character which uses 50% of black.

```pycon
>>> ["abc", "123", "\"", "'"]
['abc', '123', '"', "'"]
```
The interpreter implements the same rule ;)

Nevertheless, I [un]changed the quotes to not add any single-quotes. The diff is much smaller ;)

mkosi/__init__.py
keszybz: I renamed this to `list_to_string` and added a comment. I you still think it's cryptic, I'm fine with dropping the patch.

keszybz:
> 2\. We are properly bound by CLI backwards compat to accept something like `--bootable yes` to `--bootable`, but it's a bit weird as CLIs go, no? Is this something we should advertise, i.e. should we show a metavar at all? Also, should we maybe use something like `BooleanOptionalAction` to accept `--no-bootable` in addition (I thought we already did, but I haven't found it in the code when I was just having a quick look for it)? I think that would be a lot more conventional CLI.

What about only allowing the argument to be specified if "=" is used:
I.e. those would be all good: `--bootable=no`, `--bootable=yes`, `--bootable`.
But `--bootable no` would be interpreted as an option and a positional argument.
This way we wouldn't have ambiguity whether the argument is part of the option or not.
?

>  `--no-bootable`

I don't see much appeal. It's exactly the same number of letters as `--bootable=no`. I would prefer no add yet-another way to specify options. It also doesn't "mesh" with .ini files, where we need to have `key=value`. But it's just a weak preference — it people find that interface appealing, I wouldn't oppose it.

behrmann (approved):
LGTM. I'd ignore the warning from lgtm.com about the import style, I have no qualms about importing often-used things as `from import` and taking the rest via the proper namespace.

DaanDeMeyer:
> What about only allowing the argument to be specified if "=" is used:
I.e. those would be all good: --bootable=no, --bootable=yes, --bootable.
But --bootable no would be interpreted as an option and a positional argument.
This way we wouldn't have ambiguity whether the argument is part of the option or not.

This seems reasonable but I think we can do it in another PR. 

> I don't see much appeal. It's exactly the same number of letters as --bootable=no. I would prefer no add yet-another way to specify options. It also doesn't "mesh" with .ini files, where we need to have key=value. But it's just a weak preference — it people find that interface appealing, I wouldn't oppose it.

Agreed.

Let's wait for CI and merge this.

PR #959, opened 2022-05-05
Small improvements

DaanDeMeyer:


mkosi/__init__.py
```diff
@@ -7469,11 +7470,6 @@ def check_root() -> None:
         die("Must be invoked as root.")
 
 
-def check_native(args: MkosiArgs) -> None:
-    if args.architecture is not None and args.architecture != platform.machine() and args.build_script:
-        die("Cannot (currently) override the architecture and run build commands")
```
behrmann, line 7474: It will be a really long time before this lands in some distros and I think most people using mkosi in earnest will use the main branch (or at least a recent tag, although thanks to @bluca mkosi is really current in Debian now), so I think it would be safer to either guard this behind a distribution check or start checking the systemd version we are running on.

mkosi/backend.py
```diff
@@ -627,6 +627,10 @@ def nspawn_rlimit_params() -> Sequence[str]:
     ]
 
 
+def nspawn_executable() -> Path:
```
behrmann, line 630: This is a silly nit, feel free to ignore it, but the fallback is not really a path, but a string that will be looked up in `PATH` to get a path. I think `str` as a return type would express this better.

mkosi/__init__.py
bluca: yeah checking the nspawn version seems prudent

DaanDeMeyer: Added nspawn version check

behrmann (approved):

mkosi/backend.py
```diff
@@ -627,6 +627,14 @@ def nspawn_rlimit_params() -> Sequence[str]:
     ]
 
 
+def nspawn_executable() -> str:
+    return os.getenv("MKOSI_NSPAWN_EXECUTABLE", "systemd-nspawn")
```
bluca, line 631: should this env var be documented somewhere?

DaanDeMeyer: Probably not, it's mostly for me anyway

bluca: if it's included, we should document it - others might have the same use case (I might from time to time, and am very likely to forget)

DaanDeMeyer: Added a snippet to the docs

behrmann: 'Guess it would make sense to add this as a separate section in the man page. I always forget about the `TMPDIR` env variable, as that is only mentioned once somewhere inline.

bluca (approved):

PR #961, opened 2022-05-10
Update default fedora version to 36

DaanDeMeyer:


behrmann:
Fails at `pam_add_autologin` because of missing `/etc/pam.d/login`?

mkosi/__init__.py
```diff
@@ -3337,9 +3337,14 @@ def invoke_fstrim(args: MkosiArgs, root: Path, do_run_build_script: bool, for_ca
...
 def pam_add_autologin(root: Path, ttys: List[str]) -> None:
-    with open(root / "etc/pam.d/login", "r+") as f:
-        original = f.read()
-        f.seek(0)
+    login = root / "etc/pam.d/login"
+    if login.exists():
+        with open(login) as f:
+            original = f.read()
+    else:
+        original = ""
```
behrmann, line 3345: Just a bit shorter
```suggestion
    original = login.read_text() if login.exists() else ""
```

behrmann (commented):
TIL the failure mode that open with `r+` doesn't create the file, if it doesn't exist...

DaanDeMeyer:
Something definitely changed, autologin isn't working anymore on Fedora 36

DaanDeMeyer:
So it seems on Fedora 36 we only pull in util-linux-core and not the entirety of util-linux anymore. util-linux-core doesn't contain /bin/login which causes agetty to fail.

behrmann:
That definitely explains why the PAM file is missing.

DaanDeMeyer:
I guess for now we can explicitly install util-linux to fix the issue

DaanDeMeyer:
Reported upstream: https://bugzilla.redhat.com/show_bug.cgi?id=2084114

DaanDeMeyer:
CI failure is unrelated

Issue #962, opened 2022-05-10
ca-certificate needed when using debian security sources


nickbroon:
Trying to create a debian bullseye base image

```
[Distribution]
Distribution=debian
Release=bullseye
```

and on running `mkosi` I see the following:

```
I: Unpacking the base system...
I: Base system installed successfully.
Err:1 https://security.debian.org/debian-security bullseye-security InRelease
  Certificate verification failed: The certificate is NOT trusted. The certificate issuer is unknown.  Could not handshake: Error in the certificate verification. [IP: 151.101.194.132 443]
Hit:2 http://deb.debian.org/debian bullseye InRelease
Get:3 http://deb.debian.org/debian bullseye-updates InRelease [39.4 kB]
Get:4 http://deb.debian.org/debian bullseye/main Translation-en [6241 kB]
Get:5 http://deb.debian.org/debian bullseye-updates/main amd64 Packages [2596 B]
Get:6 http://deb.debian.org/debian bullseye-updates/main Translation-en [2343 B]
Fetched 6286 kB in 1s (4418 kB/s)
Reading package lists... Done
W: https://security.debian.org/debian-security/dists/bullseye-security/InRelease: No system certificates available. Try installing ca-certificates.
W: Failed to fetch https://security.debian.org/debian-security/dists/bullseye-security/InRelease  Certificate verification failed: The certificate is NOT trusted. The certificate issuer is unknown.  Could not handshake: Error in the certificate verification. [IP: 151.101.194.132 443]
W: Some index files failed to download. They have been ignored, or old ones used instead.
Reading package lists... Done
```

I suspect this because 'mkosi' is [invoking](https://github.com/systemd/mkosi/blob/6d01362b7d15eb66217d2d4636b9dc49a9e9857c/mkosi/__init__.py#L2744) 'debootstrap' with `--variant-minbase` which will not include ca-certificates, but is also [setting up debian security sources](https://github.com/systemd/mkosi/blob/main/mkosi/__init__.py#L2837) that need these certificates to perform verification.

Should `--include=ca-certificates` be added to the debootstrap arguments?

[...]

PR #963, opened 2022-05-11
debian: include ca-certificates for bootstrap packages

behrmann:
`apt` throws warnings because it cannot verify the certificates for the security repositories we included recently. We could add this to `extra_packages`, but then `ca-certificates` is missing when we call `apt update` for the first time, so add it to the debootsrap call.

@nickbroon Can you test whether this addresses your issue?

Fixes: #962


nickbroon:
Yes, I've tested this change and can confirm it addresses the issue. Thanks :-)

Issue #323, opened 2019-01-24
arm64 support broken


janwillies:
trying to execute `mkosi -d fedora -t directory -o fedora-osi` on arm64 results in:

```
DISTRIBUTION:                                                                              
          Distribution: fedora                                                                                                                                                                                                              
               Release: 29                                                                                                            
                                                                                            
OUTPUT:                                                                                                    
         Output Format: directory                                                                          
                Output: /root/fedora-osi                                                                              
       Output Checksum: n/a                                                                                                   
      Output Signature: n/a                                                                                     
           Output Bmap: n/a                                                                                       
Output nspawn Settings: n/a                                                                                       
           Incremental: no                                                                                                      
            Encryption: no                                                                                             
                Verity: no                                                                                         
                                                                                                                            
PACKAGES:                                                                                                          
              Packages: none                                                                                          
    With Documentation: no                                                                                 
         Package Cache: none                                                                               
           Extra Trees: none                                                                                        
        Skeleton Trees: none                                                           
          Build Script: none                                                                               
         Build Sources: /root                                                                                                                            
       Build Directory: none                                                                                                
        Build Packages: none                                                                
     Post Inst. Script: none                                                                                                     
  Scripts with network: no                                                             
       nspawn Settings: none                                                                       
                                                                                                    
HOST CONFIGURATION:                                                                                                       
    Extra search paths: none                                                                
‣ Detaching namespace...                                                                   
‣ Detaching namespace complete.                                               
‣ Setting up package cache...                                                         
‣ Setting up package cache /root/.mkosi-oa8_mne6 complete.                             
‣ Setting up temporary workspace.                                                
‣ Temporary workspace in /root/.mkosi-zo6vcdec is now set up.                    
‣ Setting up basic OS tree...                                                                
‣ Setting up basic OS tree complete.                                                
‣ Mounting Package Cache...                                                            
‣ Mounting Package Cache complete.                                           
‣ Installing Fedora...                                                         
‣ Mounting API VFS...                                                                      
‣ Mounting API VFS complete.                                              
Failed to set locale, defaulting to C                                            
Repository updates is listed more than once in the configuration                             
Repository fedora is listed more than once in the configuration           
Fedora 29 - base                                                                    7.3 MB/s |  62 MB     00:08
Fedora 29 - updates                        [   ===                                ] ---  B/s |   0  B     --:-Fedora 29 - updates                                                                 2.3 MB/s |  20 MB     00:08                                           
[...]

PR #970, opened 2022-05-14
Allow rhel-likes to build on more than x86_64

NeilHanlon:
* also add Rocky 9 key name change

Relates-to: #323


DaanDeMeyer:
Do we need to set basearch explicitly when calling dnf or does that happen automatically?

NeilHanlon:
It should happen automatically with --forcearch being specified as it's an automatic DNF variable.

Using this patch I was able to successfully build x86_64 and aarch64 images on their respective host architectures

DaanDeMeyer:
There's some CI failures for rocky:

`Status code: 404 for [https://download.rockylinux.org/pub/rocky/RPM-GPG-KEY-{keyname}](https://download.rockylinux.org/pub/rocky/RPM-GPG-KEY-%7Bkeyname%7D) (IP: 146.75.94.132)`

mstg (approved):

PR #972, opened 2022-05-16
Replace references to pacstrap with pacman

DaanDeMeyer:
We call pacman directly these days so let's remove all references
to pacstrap.


behrmann (approved):
No need to burn CPU cycles on CI for this.

PR #973, opened 2022-05-16
mkosi: Always use the embedded default version when no release is specified

DaanDeMeyer:
Let's not have the host system determine the image distribution release.
Instead, let's always default to the default release embedded within mkosi.
This gives more consistent results when building images for a single distro
regardless of the host distribution.


behrmann:
Indeed I did!

LGTM. I'm +0 on this one. I see your reasoning and I think it makes sense. I also think that the current setup makes sense, although I like this one slightly better. I don't have any strong feelings either way. It does put the onus on us to keep the default sensibly updated. :)

PR #974, opened 2022-05-16
ci: Update to Ubuntu 22.04 LTS

DaanDeMeyer:
Note: The LTS image is still in beta (https://github.com/actions/virtual-environments/issues/5490).

In Jammy, we have recent versions of zypper and dnf packaged so we
don't have to build them from source anymore. Also, sq is packaged
so we don't have to build sq from source anymore either.

The setup-github-actions.sh script is renamed to setup-pacman.sh
and it's reduced to only install the dependencies necessary to
build pacman and archlinux-keyring.

All other dependencies are moved to action.yaml and the action
mkosi.default script.


action/setup-pacman.sh
```diff
@@ -0,0 +1,48 @@
...
+        -Dldconfig=/usr/bin/ldconfig \
+        pacman-$PACMAN_VERSION-build \
+        pacman-$PACMAN_VERSION
+fi
+
+meson install -C pacman-$PACMAN_VERSION-build
+
+if [ ! -f /usr/bin/python ]; then
+    ln -s /usr/bin/python3 /usr/bin/python
+fi
```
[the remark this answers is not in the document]
behrmann: Ah, my bad then. I only looked at Debian, which still ships it, but apparently still referencing python 3.9.

DaanDeMeyer:
Final lines of output:

```
NEEDRESTART-VER: 3.5
NEEDRESTART-KCUR: 5.15.0-1005-azure
NEEDRESTART-KEXP: 5.15.0-1005-azure
NEEDRESTART-KSTA: 1
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
sudo: a password is required
```

I guess a bug in the new image?

DaanDeMeyer:
Nvm just a confusing error message from sudo

action/setup-pacman.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
```
bluca, line 1: SPDX line?

DaanDeMeyer: Let's fix in a separate PR no? Half the files in mkosi are missing SPDX lines.

bluca: Fair enough, didn't notice that - just saw this new file without it

bluca (approved):

DaanDeMeyer:
So it seems with TCG + secure-boot qemu gets stuck on Jammy

DaanDeMeyer:
Should be good to go hopefully now. Fedora qemu tests still fail on SSH connections being refused but that's not specific to this PR. Arch has failures when trying to unmount because something keeps the mount busy but that's also not specific to this PR.

DaanDeMeyer:
cc @behrmann can you give this a look?

behrmann (commented):
LGTM. One minor question, probably since this has been flying around for so long now.

action.yaml
```diff
@@ -43,6 +64,20 @@ runs:
...
 
+  # Try to eliminate "Failed to dissect image: Connection timed out" errors from nspawn by compiling
+  # systemd-nspawn from v251 from source.
+  - name: Update systemd-nspawn
+    shell: bash
+    run: |
+      echo "deb-src http://archive.ubuntu.com/ubuntu/ $(lsb_release -cs) main restricted universe multiverse" | sudo tee -a /etc/apt/sources.list
+      sudo apt-get update
+      sudo apt-get build-dep systemd
+      git clone https://github.com/systemd/systemd --branch v251-rc3 --depth=1
```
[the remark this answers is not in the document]
keszybz: v251.2 even? Or maybe `https://github.com/systemd/systemd-stable --branch v251-stable`?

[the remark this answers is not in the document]
DaanDeMeyer: Oops, let me update this

DaanDeMeyer:
Addressed final comment, setting the green label

action.yaml
```diff
@@ -56,6 +56,20 @@ runs:
...
+  # systemd-nspawn from v251 from source.
+  - name: Update systemd-nspawn
+    shell: bash
+    run: |
+      echo "deb-src http://archive.ubuntu.com/ubuntu/ $(lsb_release -cs) main restricted universe multiverse" | sudo tee -a /etc/apt/sources.list
+      sudo apt-get update
+      sudo apt-get build-dep systemd
+      git clone https://github.com/systemd/systemd-stable --branch v251.2 --depth=1
+      meson systemd-stable/build systemd-stable
+      ninja -C systemd-stable/build
```
bluca, line 77: we should add a meson build target for nspawn, if it's not there already, to speed this up a bit and avoid to rebuild the world?

DaanDeMeyer: Good point, this is available, let me post a PR to use it.

PR #975, opened 2022-05-17
machine: Translate \r\n to \n in logfile

daandemeyer:
Output lines from pexpect sent to the logfile will always end with
"\r\n" (side-effect of working with pseudo-TTYs) . On Github Actions,
this results in blank lines in the test output. Let's add a simple
adapter that translates "\r\n" back to "\n" before actually writing
to the logfile.


DaanDeMeyer:
CI failures unrelated, I checked and this is working as expected, should reduce CI log noise a bit

behrmann (approved):

PR #978, opened 2022-05-17
manifest: don't assume the size field to be valid on deb systems

behrmann:
@dellgreen Can you check whether this fixes your issue?

Fixes: 977

### DIFF
--- a/mkosi/manifest.py
+++ b/mkosi/manifest.py
@@ -149,8 +149,8 @@ class Manifest:
         for package in packages:
             name, source, version, arch, size, installtime = package.split("\t")
 
-            # dpkg records the size in KBs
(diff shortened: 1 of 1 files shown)


dellgreen:
testing the change now...

dellgreen:
@behrmann , yep that all builds fine now thank you. :+1:

DaanDeMeyer (approved):

dellgreen:
I don't believe the Fedora CI failure is related to this fix.

behrmann:
It's not. We currently don't have tests exercising the manifest (even so, having a test would not have found this bug earlier, as regular Debian packages all seem to have this field)

PR #981, opened 2022-05-18
qemu: Only enable uefi related options when we want to boot using uefi

DaanDeMeyer:


mkosi/__init__.py
```diff
@@ -7715,12 +7715,20 @@ def qemu_check_kvm_support() -> bool:
 def run_qemu_cmdline(args: MkosiArgs) -> Iterator[List[str]]:
     accel = "kvm" if args.qemu_kvm else "tcg"
 
+    if "uefi" in args.boot_protocols:
+        mode = "uefi"
+    elif "bios" in args.boot_protocols:
+        mode = "bios"
+    else:
+        mode = "uefi"
```
behrmann, line 7723:
```suggestion
    mode = "bios" if args.boot_protocols == ["bios"] else "uefi"
```
We ensure early on that `args.boot_protocols` can at most be `["bios", "uefi"]` (in some order).

behrmann (commented):
LGTM, if this fixes CI.

mkosi/__init__.py
DaanDeMeyer: Yeah but when we add the linux boot we'll need to split it up again so I would keep it as is

behrmann:
Still has SSH connection errors on Fedora and Gentoo `gpt_ext4`, but at least it boots to prompt. Gonna merge this so that the other `gpt_ext4` CI failures are fixed.

PR #982, opened 2022-05-18
Fixes

DaanDeMeyer:


DaanDeMeyer:
Split out from #926

mkosi/machine.py
```diff
@@ -211,7 +212,11 @@ def setUp(self) -> None:
         # Necessary for shell otherwise racing conditions to the disk image will happen.
         test_name = self.id().split(".")[3]
         self.machine.args.hostname = test_name.replace("_", "-")
-        self.machine.boot()
+
+        try:
+            self.machine.boot()
+        except pexpect.EOF:
+            self.fail(f"Failed to boot machine with command \"{self.machine.serial.args}\"")
```
behrmann, line 219:
```suggestion
            self.fail(f'Failed to boot machine with command "{self.machine.serial.args}"')
```
Less quoting is a good reason not to use double quotes :)

Would this maybe be more complete?
```suggestion
            self.fail(f'Failed to boot machine with command "{self.machine.serial.cmd} {self.machine.serial.args}"')
```

DaanDeMeyer: `args` includes `cmd`. Bit weird but it is what it is.

PR #985, opened 2022-05-20
machine: Hide subprocess stacktrace when a command fails during build

DaanDeMeyer:
Similar to #983 but when building the image instead of booting it.


DaanDeMeyer:
cc @behrmann

behrmann (approved):

behrmann:
Sorry, had missed this one because I confused it with #986 when looking at the titles.

PR #986, opened 2022-05-20
Fix losetup race condition with initializing partition devices

DaanDeMeyer:
This fixes the same issue we've seen in the systemd repo where
using PARTSCAN introduces a race condition with trying to use
the partition device since the kernel initializes the partition
devices asynchronously. To avoid the issue, we initialize partition
devices manually using the BLKPG ioctl(). We also avoid the same
problem on detaching loop devices by removing partition devices
explicitly using the BLKPG ioctl().

See https://github.com/systemd/systemd/pull/22992,
https://github.com/systemd/systemd/pull/23427,
https://github.com/systemd/systemd/issues/23174 and
https://github.com/systemd/systemd/issues/17469 for more context.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -540,13 +540,17 @@ _IOC_WRITE = 1  # NOQA: E221
 _IOC_READ  = 2  # NOQA: E221
 
 
-def _IOC(dir_rw: int, type_drv: int, nr: int, argtype: str) -> int:
-    size = {"int": 4, "size_t": 8}[argtype]
+def _IOC(dir_rw: int, type_drv: int, nr: int, argtype: Optional[str] = None) -> int:
+    size = {"int": 4, "size_t": 8}[argtype] if argtype else 0
     return dir_rw << _IOC_DIRSHIFT | type_drv << _IOC_TYPESHIFT | nr << _IOC_NRSHIFT | size << _IOC_SIZESHIFT
 
 
 def _IOW(type_drv: int, nr: int, size: str) -> int:
     return _IOC(_IOC_WRITE, type_drv, nr, size)
+
+
+def _IO(type_drv: int, nr: int) -> int:
+    return _IOC(_IOC_NONE, type_drv, nr)
 
 
 FICLONE = _IOW(0x94, 9, "int")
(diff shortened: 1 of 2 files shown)


DaanDeMeyer:
This should fix the errors we see [here](https://github.com/systemd/mkosi/runs/6507847869?check_suite_focus=true) when upgrading CI to jammy

mkosi/__init__.py
```diff
@@ -540,15 +540,19 @@ def format_bytes(num_bytes: int) -> str:
 _IOC_READ  = 2  # NOQA: E221
 
 
-def _IOC(dir_rw: int, type_drv: int, nr: int, argtype: str) -> int:
-    size = {"int": 4, "size_t": 8}[argtype]
+def _IOC(dir_rw: int, type_drv: int, nr: int, argtype: Optional[str] = None) -> int:
+    size = {"int": 4, "size_t": 8}[argtype] if argtype else 0
```
behrmann, line 544: Not that it matters, but one can lean into this being a dict :)
```suggestion
    size = {"int": 4, "size_t": 8}.get(argtype, 0)
```

behrmann (commented):
Seeing these PR in the systemd repo I already wondered when this would see accompanying action here. :)

LGTM, but this is lower-level than I am used to, so this is probably something sobedy else should have a look at

mkosi/__init__.py
DaanDeMeyer: Unfortunately, mypy is not happy because argtype is optional

behrmann: Looks like it's python/mypy#9430. A workaround here would be to not change the type to `Optional[str]`, but keep it `str` and make the default the empty string. As I said, though, it doesn't really matter. I just saw the `[argtype]` on the dict literal and thought it cute it lean more heavily into this being a dict. :)

DaanDeMeyer: Let's keep it like this for now I'd say

DaanDeMeyer:
First CI run in https://github.com/systemd/mkosi/pull/974 with this fix applied didn't have any loop device/partition related failures

DaanDeMeyer:
Unfortunately saw the same error even with this fix in place: https://github.com/systemd/mkosi/runs/6543132063?check_suite_focus=true

keszybz:
The code looks good to me. Dunno, if it doesn't solve the issue completely, at least it shouldn't make it worse.

PR #987, opened 2022-05-22
Fix failure of sfdisk grain detection on Ubuntu

jackwasey:
Removing the '--no-act' flag from the call to sfdisk on /dev/full
allows the command to return correctly. Without removing it, Ubuntu
sfdisk 2.37.2 (Jammy 'util-linux') tries to fsync and fails, there-
fore mkosi incorrectly decides that grain is not available, resulting in
failures with later (esp. incremental) disk image manipulations.

### DIFF
--- a/mkosi/backend.py
+++ b/mkosi/backend.py
@@ -302,7 +302,7 @@ class Partition:
 
 @functools.lru_cache(maxsize=None)
 def sfdisk_grain_is_supported() -> bool:
-    cmd: List[PathString] = ["sfdisk", "--no-reread", "--no-act", "--quiet", "/dev/full"]
+    cmd: List[PathString] = ["sfdisk", "--no-reread", "--quiet", "/dev/full"]
 
     try:
         run(cmd, text=True, input='\n'.join(["label: gpt", "grain: 4096", "quit"]), stderr=subprocess.DEVNULL)


DaanDeMeyer:
Isn't that a bug in sfdisk on Ubuntu? The --no-act flag is explicitly documented as instructing sfdisk to not write to the device node  but removing it causes it to go from writing to the device to not writing to the device?

jackwasey:
I provisionally agree it is a logic flaw in fdisk, in which case, this constitutes a work-around. Have not had time to look into, or report that bug, hence provisional, and this is all at the margin of my knowledge.

behrmann:
@jackwasey It would be great if you could file a bug and link it here. As to the change: I think this would be ok, conditionalised on being used for Ubuntu only.

keszybz:
Hmm, I tried this, and the command doesn't work at all on F36:
```console
$ echo -e 'label: gpt\ngrain: 4096\nquit' | "sfdisk" --no-reread /dev/full     
sfdisk: cannot open /dev/full: Invalid argument
$ echo -e 'label: gpt\ngrain: 4096\nquit' | "sfdisk" --no-reread  --no-act /dev/full
sfdisk: cannot open /dev/full: Invalid argument
```

With an empty temporary file:
```console
$ truncate -s 256M /tmp/img1.img
$ echo -e 'label: gpt\ngrain: 4096\nquit' | "sfdisk" --no-reread /tmp/img1.img; echo $?
...
1
$ echo -e 'label: gpt\ngrain: 4096\nquit' | "sfdisk" --no-reread --no-act /tmp/img1.img; echo $?
...
0
```

Overall, this check seems busted.

DaanDeMeyer:
If it doesn't work let's drop the check (https://github.com/systemd/mkosi/pull/1016). I added this to unblock my intern but the internship is already over and my I'm forcing my next intern to get a Fedora laptop instead of dealing with centos 8 for mkosi development so I don't need centos stream 8 support anymore.

Issue #990, opened 2022-05-26
Combining ImageId= and Encrypt=all fails in initrd on mounting /sysroot


nlacombe42:
### Error message
i am getting an error when trying to boot an arch image created by mkosi that has the root partition encrypted with luks:
![Screenshot_20220526_181227](https://user-images.githubusercontent.com/2486370/170588695-ca18bfcd-87fd-4160-b39e-4f5ea546e1c8.png)

### mkosi project files
you can see the mkosi project files here: https://github.com/nlacombe42/mkosi-bug

here is the content of `mkosi.default` : 
```
[Distribution]
Distribution=arch
Architecture=x86_64

[Output]
Format=gpt_btrfs
Bootable=true
BootProtocols=uefi
Encrypt=all
Hostname=testhost
ImageId=test-img

[Packages]
Packages=
    base
    linux
    util-linux
    systemd
    bash

[Partitions]
RootSize=3G
ESPSize=256M
```

mkosi build logs: https://0x0.st/oBzR.log

### Host system info
i ran the build inside [this arch vm image](https://gitlab.archlinux.org/archlinux/arch-boxes/-/jobs/57805/artifacts/raw/output/Arch-Linux-x86_64-basic-20220526.57805.qcow2) using qemu and [these steps](https://github.com/nlacombe42/mkosi-bug/blob/main/howto-build-in-qemu.txt).

```
$ uname -a
Linux archlinux 5.17.9-arch1-1 #1 SMP PREEMPT Wed, 18 May 2022 17:30:11 +0000 x86_64 GNU/Linux
$ cat /etc/os-release
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://bugs.archlinux.org/"
LOGO=archlinux-logo
$ pacman -Qs 'mkosi|systemd'
local/mkosi 12-1
    Build Legacy-Free OS Images
local/systemd 251.1-1
    system and service manager
local/systemd-libs 251.1-1
    systemd client libraries
local/systemd-sysvcompat 251.1-1
    sysvinit compat for systemd
```

### Steps to reproduce
1. execute the following on your host system to get an arch vm image:
```
#!/usr/bin/env bash

[...]

PR #992, opened 2022-05-29
fix mkosi luks with image id bug

nlacombe42:
ok, so i am a bit suspisious of that change, but removing `partlabel` from `install_unified_kernel()` seems to fix bug #990,
and all tests (including integration tests) passes (on my local computer).

since all tests passes, i am venturing to create this pr.

keep in mind that i am not familiar with this codebase, and my python knowledge is very limited,
so please triple check everything.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -4277,13 +4277,8 @@ def install_unified_kernel(
         for kver, kimg in gen_kernel_images(args, root):
             if args.image_id:
                 image_id = args.image_id
-                if args.image_version:
-                    partlabel = f"{args.image_id}_{args.image_version}"
-                else:
-                    partlabel = f"{args.image_id}"
             else:
                 image_id = f"mkosi-{args.distribution}"
-                partlabel = None
 
             if args.image_version:
                 boot_binary = Path(prefix) / f"EFI/Linux/{image_id}_{args.image_version}.efi"
@@ -4302,8 +4297,6 @@ def install_unified_kernel(
             if root_hash:
                 option = "usrhash" if args.usr_only else "roothash"
                 boot_options = f"{boot_options} {option}={root_hash}"
-            elif partlabel:
-                boot_options = f"{boot_options} root=PARTLABEL={partlabel}"
 
             cmd: Sequence[PathString] = [
                 "dracut",
--- a/mkosi/machine.py
+++ b/mkosi/machine.py
@@ -46,12 +46,13 @@ class LogfileAdapter:
 
 
 class Machine:
-    def __init__(self, args: Sequence[str] = []) -> None:
+    def __init__(self, args: Sequence[str] = [], luks_password: str = None) -> None:
         # Remains None until image is built and booted, then receives pexpect process.
         self._serial: Optional[pexpect.spawn] = None
         self.exit_code: int = -1
         self.stack = contextlib.ExitStack()
         self.args: MkosiArgs
+        self.luks_password: str = luks_password
 
         tmp = parse_args(args)["default"]
 
@@ -147,6 +148,11 @@ class Machine:
             # Then, when pexpects finds the '#' it means we're ready to interact with the process.
             self._serial = pexpect.spawnu(command=cmd[0], args=cmd[1:], logfile=LogfileAdapter(sys.stdout),
                                           timeout=240)
+
+            if self.luks_password is not None:
+                self._serial.expect("Please enter passphrase for disk")
+                self._serial.write(self.luks_password + "\n")
+
             self._serial.expect("#")
             self.stack = stack.pop_all()
 
@@ -187,14 +193,16 @@ def skip_not_supported() -> Iterator[None]:
 class MkosiMachineTest(unittest.TestCase):
     args: Sequence[str]
     machine: Machine
+    luks_password: str
 
-    def __init_subclass__(cls, args: Sequence[str] = []) -> None:
+    def __init_subclass__(cls, args: Sequence[str] = [], luks_password: str = None) -> None:
         cls.args = args
+        cls.luks_password = luks_password
 
     @classmethod
     def setUpClass(cls) -> None:
         with skip_not_supported():
-            cls.machine = Machine(cls.args)
+            cls.machine = Machine(cls.args, cls.luks_password)
 
         verb = cls.machine.args.verb
         no_nspawn = parse_boolean(os.getenv("MKOSI_TEST_NO_NSPAWN", "0"))
--- a/mkosi/machine.py
+++ b/mkosi/machine.py
@@ -46,13 +46,13 @@ class LogfileAdapter:
 
 
 class Machine:
-    def __init__(self, args: Sequence[str] = [], luks_password: str = None) -> None:
+    def __init__(self, args: Sequence[str] = [], luks_password: Union[str, None] = None) -> None:
         # Remains None until image is built and booted, then receives pexpect process.
         self._serial: Optional[pexpect.spawn] = None
         self.exit_code: int = -1
         self.stack = contextlib.ExitStack()
         self.args: MkosiArgs
-        self.luks_password: str = luks_password
+        self.luks_password: Union[str, None] = luks_password
 
         tmp = parse_args(args)["default"]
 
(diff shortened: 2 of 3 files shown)


behrmann (changes requested):

behrmann:
@DaanDeMeyer I'll leave checking the integration tests to you, since you are the expert there.

mkosi/__init__.py
```diff
@@ -4277,13 +4277,8 @@ def install_unified_kernel(
         for kver, kimg in gen_kernel_images(args, root):
             if args.image_id:
                 image_id = args.image_id
-                if args.image_version:
-                    partlabel = f"{args.image_id}_{args.image_version}"
-                else:
-                    partlabel = f"{args.image_id}"
```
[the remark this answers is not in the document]
nlacombe42: first i want to say that i didn't know there was other tests than the ones ran by pytest (the integration tests in the ci), so i will see what makes those fail and try to fix that.
i assume that the ci tests that failed probably failed because they broke "building different image versions".


i made that change because:
when building images with no modifications to the production code of mkosi,
when building images that have luks root partition and an explicitely set image id, the `root=PARTLABEL=IMAGE_ID` kernel parameter is sent to dracut, and booting the image fails;
when building images that have luks root partition and **no** explicitely set image id, the `root` kernel parameter is **not** sent to dracut, and booting the image works.

when the image fails to boot, the error is that it tried to mount the partition with the label provided in the `root` kernel parameter without decrypting it first.

when the image boots fine, no `root` kernel parameter is sent to dracut, and i could not (and still can't) see any place where information is passed to the kernel or initrd to specify which partition to boot.
so i assume that dracut, or whatever process is ran inside the initrd must autodectect what partition to boot and correctly ask for luks password (and decrypt the partition) when it sees a luks partition.

i tried to quickly read some documentation on dracut and systemd, but i could not see how it seems to be able to find the right partition (and to ask password and decrypt in the case of luks) when there is no kernel paremeter passed to specify which partition to boot.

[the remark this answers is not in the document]
behrmann: No need to apologise! Thanks for the analysis of the problem.

I'm unsure that the failing tests have nothing to do with your current change here, the comment about building different image versions is regarding that this would break building system extension images (see `man systemd-systext`).

Without any kernel parameters the discoverable partition scheme should (and apparently does) pick up the right partitions to mount, but in apparently we have some ill-defined mixture here. Will have to look more closely at this.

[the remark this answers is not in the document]
nlacombe42: by "discoverable partition scheme" do you mean [this](https://systemd.io/DISCOVERABLE_PARTITIONS/) ? that's relly cool i didn't know about this.

i read the [systemd-sysext man page](https://man.archlinux.org/man/systemd-sysext.8) but i still don't understand how not passing the root kernel parameter could break that. it looks like to me that it's either finding the extension via discoverable partition, in which case it does not need the kernel parameter, or it reads the file from already mounted fs, in which case you's have to already have mounted the fs in another way, which i am assuming would be fstab or discoverable partitions, or means custom change made by a user that is not made by mkosi.
maybe you could explain to me more how you think my change would break this.

also, now that i know about discoverable partitions, if mkosi always generate an image with discoverable partitions, then i don't see any need for the root kernel prameter.
and the `partlabel` variable which i have removed was **only** (afaict) used to set the root kernel parameter.

tests/test_machine_luks.py
```diff
@@ -0,0 +1,40 @@
+# SPDX-License-Identifier: LGPL-2.1+
+
+import os
+import pytest
+
+from mkosi.backend import Distribution, Verb
+from mkosi.machine import MkosiMachineTest
+
+pytestmark = [pytest.mark.integration, pytest.mark.skipif(os.getuid() != 0, reason="Must be invoked as root.")]
```
DaanDeMeyer, line 9: No need for this

mkosi/machine.py
```diff
@@ -147,6 +148,11 @@ def boot(self) -> None:
             # Then, when pexpects finds the '#' it means we're ready to interact with the process.
             self._serial = pexpect.spawnu(command=cmd[0], args=cmd[1:], logfile=LogfileAdapter(sys.stdout),
                                           timeout=240)
+
+            if self.luks_password is not None:
+                self._serial.expect("Please enter passphrase for disk")
+                self._serial.write(self.luks_password + "\n")
+
```
DaanDeMeyer, line 155: Why do we have both a password file and this code to pass the password in via the terminal? Shouldn't we just use the password file and drop this?

tests/test_machine_luks.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class MkosiLuksWithImageIdMachineTestCase(MkosiMachineTest):
+    luks_password_file_path: str = "./mkosi.passphrase"
```
DaanDeMeyer, line 13: `luks_password_file: Path = Path("mkosi.passphrase"`)

tests/test_machine_luks.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        luks_password = "luks"
+
+        with open(cls.luks_password_file_path, "w", encoding="utf-8") as luks_password_file:
+            luks_password_file.write(luks_password)
```
DaanDeMeyer, line 20: `luks_password_file.write_text("luks")`

tests/test_machine_luks.py
```diff
@@ -0,0 +1,40 @@
...
+
+        command = "-d arch --encrypt all --image-id test-img qemu".split(" ")
+
+        super().__init_subclass__(command, luks_password)
```
DaanDeMeyer, line 24: This should be done as

```
class MkosiLuksWithImageIdMachineTestCase(MkosiMachineTest, args=[...]):
    ....
```

tests/test_machine_luks.py
```diff
@@ -0,0 +1,40 @@
...
+        if (
+            cls.machine.args.distribution == Distribution.centos_epel
+            and cls.machine.args.verb == Verb.qemu
+            and not cls.machine.args.qemu_kvm
+        ):
+            pytest.xfail("QEMU's CPU does not support the CentOS EPEL image arch when running without KVM")
+
+    def test_simple_run(self) -> None:
+        process = self.machine.run(["echo", "This is a test."], capture_output=True)
+        assert process.stdout.strip("\n") == "This is a test."
```
DaanDeMeyer, line 36: This can be removed, we just need to check if the machine boots, no need to check if running commands works, we already do that in the basic tests.

tests/test_machine_luks.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def tearDown(self) -> None:
+        super().tearDown()
+        os.remove(self.luks_password_file_path)
```
DaanDeMeyer, line 40: `luks_password_file.unlink()`

DaanDeMeyer (changes requested):
Only reviewed the integration test.

mkosi/__init__.py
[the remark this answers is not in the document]
behrmann: @nlacombe42 Sorry for the late reply, I was travelling last week. 

We can't do without the partlabel because we document that it will be set when `ImageID` is set and it's documented to be used in the discoverable partition scheme as being used for version comparisons if multiple images with the same image ID are being passed. The relevant PR introducing this is #754.

Also thinking about the discoverable partition scheme was a bit of a red herring here (sorry), since setting `root=` on the kernel cmdline disables the generator.

nlacombe42:
before making any further code changes on this pr, i will wait to see if there is at least a concensus on #990 about if it should be allowed to use `ImageId=` in conjunction with `Encrypt=all`, see https://github.com/systemd/mkosi/issues/990#issuecomment-1151380280

mkosi/__init__.py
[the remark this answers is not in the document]
DaanDeMeyer: Euhh I'm slightly confused here as well on why we configure PARTLABEL on the cmdline explicitly if we're using image IDs. Don't we always want to use systemd-gpt-generator to figure out the partition to mount? It should automatically use the partition label in its version comparison logic to figure out the right partition no?

cc @poettering

DaanDeMeyer:
Closing due to inactivity, we can reopen again if needed

Issue #950, opened 2022-04-17
bmap file is being created after compression


gsalvatella:
Using `--bmap` and `--compress-output=xz` flags together, I was expecting the `.bmap` file to be created on the raw image and the compression to take place thereafter. Instead I get the following output:
```
‣ Compressing output file /home/user/images/.mkosi-v92rp6c8...
‣ Creating BMAP file...
bmaptool: WARNING: all 330.8 MiB are mapped, no holes in '/home/user/images/.mkosi-jy908gzc'
bmaptool: WARNING: was the image handled incorrectly and holes were expanded?
```
When using `bmaptool copy` I then get an error
```
bmaptool: ERROR: checksum mismatch for blocks range 0-84687: calculated 4c4cb38c90709a58fb4696e2c87663bc2644009100da275a6f492017e9524b8f, should be 7a22c8bf1bf56316b5adfdbabbcf4b5f62a9ab6544b1065bf0ad7b8c1b27e28a (image file image.raw.xz)
```
Is there a reason for this? Should I be creating the raw image first without compression and compress it manually afterwards?

```shell
mkosi --version
mkosi 10
```

gsalvatella:
The bmap file must be created before compression since bmap needs to know where the "holes" of the image are. After compression the holes are gone and bmap complains with the "no holes found" warning.

>  Does this work if you don't have mkosi compress the output image?

Yes, the `--bmap` flag works fine. The compression step should simply come thereafter.

behrmann:
> Be aware if you are running on a ZFS partition, you need to set a ZFS property, instructions are in the docs.

Just out of morbid curiosity about the world of out of tree modules: what would the proper incantations be?

dellgreen:
> > Be aware if you are running on a ZFS partition, you need to set a ZFS property, instructions are in the docs.
> 
> Just out of morbid curiosity about the world of out of tree modules: what would the proper incantations be?

I added the incantations here:

https://github.com/intel/bmap-tools#zfs-file-system

:)

behrmann:
Closing thanks to #994

PR #994, opened 2022-06-01
mkosi: create bmap before compression

gsalvatella:
Creation of bmap files needs to take place before any compression
happens, since bmaptool has to know where the "holes" of the image lie.
Compression removes the holes, preventing bmap from recreating the
original raw image. Move the bmap calculation step before the
compression.

Use case: compressed bmap images are very convenient for deploying OS images to remote repositories. These can be flashed directly to e.g. embedded devices or USB installers using `bmaptool copy https://remote/file /dev/sdX`, which streamlines download, decompression and flashing of the image in one go.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -7448,6 +7448,7 @@ def build_stuff(args: MkosiArgs) -> Manifest:
             MkosiPrinter.print_step("Skipping (second) final image build phase.")
 
         raw = qcow2_output(args, image.raw)
+        bmap = calculate_bmap(args, raw)
         raw = compress_output(args, raw)
         split_root = compress_output(args, image.split_root, ".usr" if args.usr_only else ".root")
         split_verity = compress_output(args, image.split_verity, ".verity")
@@ -7469,7 +7470,6 @@ def build_stuff(args: MkosiArgs) -> Manifest:
             settings,
         )
         signature = calculate_signature(args, checksum)
-        bmap = calculate_bmap(args, raw)
 
         link_output(args, root, raw or image.archive)
         link_output_root_hash_file(args, root_hash_file)


behrmann:
Does this fix #950 for you?

behrmann:
The pyright failure is unrelated to your change, probably some pyright change needs us to also install `cryptography` in addition to `types-cryptography` (in `.github/workflows/ci.yml`)

behrmann:
> I'm not sure what's gone wrong with the unit test. Seems like `sock = Path(value)` fails on type checking? You should probably check with `if value:`... Actually you have this on top, sorry.

It's not your bug and looking at the line I actually think this is wrong. Will prep a patch.

behrmann:
@gsalvatella Could you rebase against `main`, please?

behrmann:
> Nice, your patch has worked on the previous error at least...

No worries, some of the tests are flaky and the Gento directory test seems to have an issue with the mirror.

behrmann (approved):

behrmann:
@dellgreen Since you seem to be the expert on bmap. Could you have a look whether this now works for you as expected, too?

dellgreen:
> @dellgreen Since you seem to be the expert on bmap. Could you have a look whether this now works for you as expected, too?

@behrmann I wouldn't go that far, but thanks for the vote of confidence. ;)

But yes that all works now as expected

```
# bmaptool copy --bmap foo.bmap foo.raw.xz /dev/loop50
bmaptool: info: block map format version 2.0
bmaptool: info: 26640 blocks of size 131072 (3.3 GiB), mapped 17247 blocks (2.1 GiB or 64.7%)
bmaptool: info: copying image 'foo.raw.xz' to block device '/dev/loop50' using bmap file 'foo.bmap'
bmaptool: info: 100% copied
bmaptool: info: synchronizing '/dev/loop50'
bmaptool: info: copying time: 43.7s, copying speed 49.4 MiB/sec
```

```
Disk foo.raw: 9.78 GiB, 10485760000 bytes, 20480000 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: gpt
Disk identifier: 880DEA96-2CCD-5C47-8D7C-95D5CE997E71

Device       Start     End Sectors  Size Type
foo.raw1   2048  526335  524288  256M EFI System
foo.raw2 526336 6817791 6291456    3G Linux root (x86-64)
```

gsalvatella:
Thank you. I also confirm it's working for me. The compression process now takes place after bmap...

```
‣  Detaching /dev/loop4
‣ Creating BMAP file…
‣ Compressing output file /home/gmask/repos/ansible/lava/images/.mkosi-ldyeqb9n…
‣ Linking image file…
```

PR #996, opened 2022-06-01
ssh: make parse_ssh_agent only handle strings

behrmann:
pyright complains (wrongly I think) about value being None when passed to Path to create the socket variable. Let's work around this by eliminating Nones as values.


behrmann:
As a note for later readers: This does introduce a slight behavioural change, though. Before passing an empty string as argument would have fallen through to the end and raised an error since the check whether the current working directory (the default when `Path` is given an empty string) whereas now it bails out early.

PR #1001, opened 2022-06-09
arch: Drop locale-gen logic

DaanDeMeyer:
C.UTF-8 is included by default in the latest glibc package update.


behrmann:
That means `C.UTF-8` is available on sufficiently new glibc even when the locale has not been built? If so, LGTM. Could you add the glibc version in the commit message, so we remember to drop similar workaround for other when they get sufficiently new glibc?

behrmann:
Right. Quickly grepping for `C.UTF-8` still showed me hits, but that was `locale.conf`, nevermind then.

behrmann (approved):

PR #1002, opened 2022-06-09
arch: Use cloudflare mirror on CI

DaanDeMeyer:
Should be a little more reliable than picking a random mirror from
the mirrorlist.


behrmann:
Either the cloudflare mirror changed right while the tests were running or this is flakier than the current random choice

DaanDeMeyer:
Uhhh yeah it's not making a good first impression, let me retry.

DaanDeMeyer:
Nvm this just seems worse

Issue #752, opened 2021-07-02
Do unified kernel image generation (i.e. objcopy invocation) without dracut


poettering:
I think the objcopy line for combining the kernel image with the initrd and the kernel cmdline is probably something we should do on the host instead of inside the build environment.

Right now, we rely on dracut invoking objcopy for us. However, I think that is less than ideal:

1. It means we strictly have to use dracut for initrds, but I think we should add support for other options (right, @keszybz ;-))
2. objcopy needs to be installed in the final image, so that dracut can install it, objcopy means binutils which is pretty clearly a development package, and it kinda sucks that it's kind the only tool we need to install in the final image too, just because of this. If mkosi instead of dracut combines the initrd + kernel via objcopy this means we can do this host side, and thus require no deps inside the container for it.
3. this would allow us to move building of initrds from the 2nd phase of build to the 1st phase in more cases. Currently it was done in phase 2 in particular when verity was used, because in that case the top hash of the verity image must be included in the kernel cmdline embedded into the image and that is only known at the end of the 2nd phase. But given that this hash actually is only needed for the objcopy invocation and not already for the initrd generation, we can do building of the initrds in phase 1. This also means we no longer would have to install dracut at all in the 2nd phase of the boot.
4. It might make sense to support upstream distro pre-generated initrds one day, i.e. where the cpio is already complete, and we then just need to combine it with a kernel cmdline and sign it afterwards. in this mode we'd not need dracut at all anymore, and the objcopy step we do ourselves would suffice entirely.

brpatel12:
Hi @poettering - until this get's added to mkosi, is there a suggested workaround to use objcopy directly (as opposed to dracut)?

behrmann:
You could probably do that in a build or finalise script - dracut would still be run - but you could do your own thing on top.

keszybz:
Fixed by 683268367374e7c1c659af38b575ab0ad0c1088b.

PR #1007, opened 2022-06-13
Rework initramfs generation (again)

DaanDeMeyer:
See second commit message for a detailed description.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1814,6 +1814,11 @@ def prepare_tree(args: MkosiArgs, root: Path, do_run_build_script: bool, cached:
             else:
                 # If this is not enabled, let's create an empty directory on /boot
                 root.joinpath("boot").mkdir(mode=0o700)
+                # Make sure kernel-install actually runs when needed by creating the machine-id subdirectory
+                # under /boot. For "bios" on Debian/Ubuntu, it's required for grub to pick up the generated
+                # initrd.
+                if args.distribution in (Distribution.debian, Distribution.ubuntu) and "bios" in args.boot_protocols:
+                    root.joinpath("boot", args.machine_id).mkdir(mode=0o700)
 
             if args.get_partition(PartitionIdentifier.esp):
                 root.joinpath("efi/EFI").mkdir(mode=0o700)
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -4266,7 +4258,7 @@ def extract_partition(
 def gen_kernel_images(args: MkosiArgs, root: Path) -> Iterator[Tuple[str, Path]]:
     # Apparently openmandriva hasn't yet completed its usrmerge so we use lib here instead of usr/lib.
     for kver in root.joinpath("lib/modules").iterdir():
-        if not (kver.is_dir() and os.path.isfile(os.path.join(kver, "modules.dep"))): # type: ignore
```
DaanDeMeyer, line 4269: I had to drop the modules.dep check here since we haven't run depmod yet when running kernel-install here since we disable running kernel-install as part of package installation. If this is still a problem, we'll need to figure out another way to distinguish valid directories from nonvalid directories

DaanDeMeyer:
Fixes #752

mkosi/__init__.py
```diff
@@ -4350,17 +4347,24 @@ def install_unified_kernel(
...
-                "--uefi",
-                "--kver", kver,
-                "--kernel-image", Path("/") / kimg,
-                "--kernel-cmdline", boot_options,
-                "--force",
+                "objcopy",
+                "--add-section", f".osrel={osrelease}",   "--change-section-vma", ".osrel=0x20000",
+                "--add-section", f".cmdline={cmdline}",   "--change-section-vma", ".cmdline=0x30000",
+                "--add-section", f".linux={root / kimg}", "--change-section-vma", ".linux=0x2000000",
+                "--add-section", f".initrd={initrd}",     "--change-section-vma", ".initrd=0x3000000",
```
keszybz, line 4360: I'll repeat the comment from the other PR, so this doesn't get lost:
dracut varied this by arch: 0x3000000 by default, but 0x4000000 on arm64. Was there some discussion of this value?

keszybz (commented):
LGTM. I'm not sure about the offset (see comment), but we can adjust this later if appropriate.

mkosi/__init__.py
DaanDeMeyer: I think I just stole this from dracut when I initially wrote this, but I might have gotten it wrong, will double check again.

PR #1008, opened 2022-06-14
Log umount blockers when umount fails

DaanDeMeyer:
We've been experiencing sporadic CI failures because umount returns
EBUSY. Similar to https://github.com/systemd/systemd/pull/23348,
let's try to find out which processes are blocking umount() and
log about them so we can try to figure out why these processes are
still running and keeping files open when we try to run umount.


mkosi/backend.py
```diff
@@ -950,3 +952,44 @@ def complete_step(cls, text: str, text2: Optional[str] = None) -> Iterator[List[
...
+
+            if path_startswith_any(p, ("/dev", "/sys", "/proc")):
+                continue
+
+            blockers += [(comm, p)]
+
+    if blockers:
+        MkosiPrinter.warn(f"Unmounting '{mount}' is blocked by:")
+        for comm, p in blockers:
+            MkosiPrinter.warn(f"- {comm} ({p})")
```
behrmann, line 995:
```suggestion
            MkosiPrinter.warn(f"- {comm} ({p})")
```
and the `MkosiPrinter.warn(f"Unmounting '{mount}' is blocked by:")` can be moved to any point before the beginning of the loop, since in we are in this function we know we are being blocked by *something* already.

DaanDeMeyer: Not entirely. umount can fail for a number of reasons, and unfortunately we can't detect whether it's EBUSY or something else so it could potentially be that there's a different error and that there aren't any processes blocking the unmount. We could grep the error message but that's kinda nasty, so this seemed like the better option.

behrmann: True, there could be other reasons for umount failure, didn't think about that. TIL, only has an opaque 1 return code on error. How helpful.

I still think printing the blocking files right away, maybe as `{comm} blocking on {p}` or `{comm} holding {p} open` would be simpler than collecting them all first, but at this point that's a question of stylistic preference.

behrmann:
LGTM.

mkosi/backend.py
```diff
@@ -950,3 +952,37 @@ def complete_step(cls, text: str, text2: Optional[str] = None) -> Iterator[List[
 
         if text2 is not None:
             cls.print_step(text2.format(*args))
+
```
behrmann, line 955: Just saw that, one empty line missing here. ;)

DaanDeMeyer:
Hmm, it doesn't seem to work: https://github.com/systemd/mkosi/runs/6900102329?check_suite_focus=true

DaanDeMeyer:
Updated to not reinvent the wheel and simply use lsof

DaanDeMeyer:
Ugh that seems to show every path instead of just the ones in the directory: https://github.com/systemd/mkosi/runs/6902361283?check_suite_focus=true.

DaanDeMeyer:
Let's try fuser

DaanDeMeyer:
I'm going to close this one for now since arch seems surprisingly stable now, I can resurrect it if we run into the same problem again

PR #1009, opened 2022-06-15
arch: Use gpgdir from host system

DaanDeMeyer:
Instead of setting up the keyring in the image, let's reuse the
keyring from the host. If users want to use pacman in the image,
they just have to run pacman-key themselves in a postinst script
or such.

This speeds up building of images and hopefully also gets rid of
our CI issues with Arch where there's something keeping files open
in the root mount (which I expect is gpg-agent).


DaanDeMeyer:
This will require an initialized keyring on the host, so aside from installing archlinux-keyring (which was already required), users will also have to run `pacman-key --init` and `pacman-key --populate archlinux` once. On Arch (and hopefully on Fedora if https://src.fedoraproject.org/rpms/archlinux-keyring/pull-request/1 is accepted), the keyring is automatically kept up-to-date on updates of archlinux-keyring if it is initialized, so I don't think it's too much effort for users.

behrmann:
Looks good. No failures from the Arch CI. Can you add a news entry for this, though, so as not to surprise anybody?

PR #1011, opened 2022-06-16
Communicate the associated dir in the ESP via /etc/kernel/entry-token

### LINKED ISSUE #686: EFI entries in the built image have a different machine ID
Hello,

When building an image and running some tests (using the latest mkosi master) I noticed that the entries under `/efi` are generated for a different machine ID than is present in the final image booted via `mkosi qemu`:

`mkosi.default`
```ini
[Distribution]
Distribution=arch

[Host]
NetworkVeth=yes
Ssh=yes

[Output]
Bootable=yes
Format=gpt_ext4
WithoutUnifiedKernelImages=yes

[Partitions]
RootSize=32G
```

```
# mkosi build
...
(10/11) Adding kernel and initramfs images to /boot...
dracut: Executing: /usr/bin/dracut -f /efi/492174d5cb10476084e2d69df3cf4568/5.11.8-arch1-1/initrd 5.11.8-arch1-1
...
dracut: *** Creating image file '/efi/492174d5cb10476084e2d69df3cf4568/5.11.8-arch1-1/initrd' ***
dracut: *** Creating initramfs image file '/efi/492174d5cb10476084e2d69df3cf4568/5.11.8-arch1-1/initrd' done ***
...
‣ Successfully linked /root/test/image.raw.
‣ Linking private ssh key file...
‣ Successfully linked /root/test/id_rsa.
‣ Resulting image size is 32.3G, consumes 1.0G.

# mkosi qemu &
# mkosi ssh "bootctl list"
Boot Loader Entries:
        title: Arch Linux (default)
           id: 492174d5cb10476084e2d69df3cf4568-5.11.8-arch1-1.conf
       source: /efi/loader/entries/492174d5cb10476084e2d69df3cf4568-5.11.8-arch1-1.conf
      version: 5.11.8-arch1-1
   machine-id: 492174d5cb10476084e2d69df3cf4568
        linux: /492174d5cb10476084e2d69df3cf4568/5.11.8-arch1-1/linux
       initrd: /492174d5cb10476084e2d69df3cf4568/5.11.8-arch1-1/initrd
      options: rhgb selinux=0 audit=0 rw

        title: Reboot Into Firmware Interface
           id: auto-reboot-to-firmware-setup
       source: /sys/firmware/efi/efivars/LoaderEntries-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f
# mkosi ssh "cat /etc/machine-id"
2a7b0c0b016f4fdea7a6a4ecc63a28ce
```

This causes issues with scripts which expect the EFI configuration under a subdirectory matching the current `machine-id` (like the systemd's integration test suite).

DaanDeMeyer:
/etc/kernel/entry-token is the new way introduced in systemd v251 to
identify the directory that kernel-install and bootctl should install
their stuff to. If it exists, bootctl and kernel-install will read it
and use the directory inside to install things to.

Currently, in mkosi, we generate a random machine ID during the build
and use that as the directory under the ESP to install things to. Until
all distros we support get support for /etc/kernel/entry-token, we're
limited to using the machine ID as the directory to install stuff under
in the ESP.

Since the machine ID used during the build is scrubbed from the image,
users don't a way to figure out the directory in the ESP associated
with the rootfs after the build is finished. To fix this, let's write
the machine ID to /etc/kernel/entry-token before it is scrubbed so
that users can read the file to figure out which directory in the ESP
that they should look under to find stuff associated with the
corresponding rootfs.

Fixes #686

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1835,6 +1835,7 @@ def prepare_tree(args: MkosiArgs, root: Path, do_run_build_script: bool, cached:
             root.joinpath("etc/kernel").mkdir(mode=0o755)
 
             root.joinpath("etc/kernel/cmdline").write_text(" ".join(args.kernel_command_line) + "\n")
+            root.joinpath("etc/kernel/entry-token").write_text(f"{args.machine_id}\n")
 
         if do_run_build_script or args.ssh:
             root_home(args, root).mkdir(mode=0o750)


mkosi/__init__.py
```diff
@@ -1836,6 +1836,7 @@ def prepare_tree(args: MkosiArgs, root: Path, do_run_build_script: bool, cached:
             root.joinpath("etc/kernel").mkdir(mode=0o755)
 
             root.joinpath("etc/kernel/cmdline").write_text(" ".join(args.kernel_command_line) + "\n")
+            root.joinpath("etc/kernel/entry-token").write_text(f"{args.machine_id}\n")
```
behrmann, line 1839: I think I understand your reasoning for using the machine id here, but wouldn't the image id not also be a sensible choice, so that in the end this would be sensible to be configurable?

DaanDeMeyer: Yes eventually it should be configurable. I just figured we could wait with adding that until this is supported somewhat widely

PR #1015, opened 2022-06-19
Add veritysetup.target to initrd

### LINKED ISSUE #1014: systemd-veritysetup@root.service not started
I've created an Arch image with these options: `mkosi --distribution=arch --bootable --verity --split-artifacts --qemu-headless --autologin qemu`.
However /dev/mapper/root isn't created: `[   91.134659] archlinux systemd[1]: Timed out waiting for device /dev/mapper/root.` and the emergency shell starts. If I run `# systemctl start systemd-veritysetup@root.service` the root mapper is created and I can start `multi-user.target` afterward.

Should this be fixed in mkosi? What is the right way to make sure `systemd-veritysetup@root.service` is started? 

[Log output](https://gist.github.com/pyfisch/a358037a1dbe072f5039610199e88e39). There is a `blockdev@dev-mapper-root.target` which isn't found.

Edit: I am running mkosi from the main branch, because #1003 is fixed there.

pyfisch:
The target is needed so that usr and root partitions
can be verified and mounted early during boot.

Closes #1014


DaanDeMeyer:
Lgtm

PR #1016, opened 2022-06-20
Drop sfdisk grain support check

DaanDeMeyer:
This check doesn't actually work on Fedora so let's drop it until
we can find a better check.

### DIFF
--- a/mkosi/backend.py
+++ b/mkosi/backend.py
@@ -6,7 +6,6 @@ import argparse
 import contextlib
 import dataclasses
 import enum
-import functools
 import math
 import os
 import resource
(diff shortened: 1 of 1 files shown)


keszybz:
LGTM.

keszybz:
```
mkosi/backend.py:379:44 undefined name 'sfdisk_grain_is_supported'
```

nick123pig:
@DaanDeMeyer Was there any more details about how it didn't work in fedora? I am trying to get this going on rhel8, they have the old sfdisk. Or do you want to enforce a grain-supporting `sfdisk`?

PR #1017, opened 2022-06-20
debian/ubuntu: Handle scenario where systemd-boot is a split package

DaanDeMeyer:


mkosi/__init__.py
```diff
@@ -2932,6 +2932,12 @@ def install_debian_or_ubuntu(args: MkosiArgs, root: Path, *, do_run_build_script
     install_skeleton_trees(args, root, False, late=True)
 
     invoke_apt(args, do_run_build_script, root, "update", [])
+
+    if args.bootable and not do_run_build_script and args.get_partition(PartitionIdentifier.esp):
```
bluca, line 2936: it should also have `and args.base_image is None` in the check I think?

DaanDeMeyer: I think that's handled by `add_packages()`?

bluca: ok

bluca (approved):

mkosi/__init__.py
behrmann: Well, `add_packages` uses `args.base_packages`, which is set to `False` when `args.base_image is None`, so it's a bit more general, but it should work out the same.

behrmann (approved):

PR #1019, opened 2022-06-20
action: Build only systemd-nspawn

DaanDeMeyer:
Speed things up by only building nspawn instead of the entirety of
systemd.


bluca (approved):

keszybz:
LGTM too.

PR #1025, opened 2022-06-24
action: Pin systemd-container version

DaanDeMeyer:
A faulty metadata update in Github Actions azure mirror for Ubuntu
is causing issues with unmet systemd-container versions. Let's pin
the systemd-container version as a workaround until the issue is
fixed.

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -27,7 +27,12 @@ runs:
         debootstrap \
         zypper \
         dnf \
-        systemd-container \
+        systemd=249.11-0ubuntu3.3 \
+        systemd-sysv=249.11-0ubuntu3.3 \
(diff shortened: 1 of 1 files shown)


behrmann (approved):

DaanDeMeyer:
Updated to pin the systemd-container version instead since it also ships machined which we also need

behrmann:
We do? I had the idea we might need it, but `git grep machinectl` only shows documentation files containing `machinectl`.

DaanDeMeyer:
> We do? I had the idea we might need it, but git grep machinectl only shows documentation files containing machinectl.

nspawn registers stuff with machined which is required to use systemd-run on the container for the integration tests.

PR #1027, opened 2022-06-27
Ignore ENOTTY in copy_file_object() and copy_fd()

DaanDeMeyer:
This is another possible error that can be returned when FICLONE is
not supported.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -583,7 +583,7 @@ def copy_fd(oldfd: int, newfd: int) -> None:
     try:
         _reflink(oldfd, newfd)
     except OSError as e:
-        if e.errno not in {errno.EXDEV, errno.EOPNOTSUPP}:
+        if e.errno not in {errno.EXDEV, errno.EOPNOTSUPP, errno.ENOTTY}:
             raise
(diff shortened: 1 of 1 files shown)


bluca (approved):

behrmann (approved):

mkosi/__init__.py
```diff
@@ -583,7 +583,7 @@ def copy_fd(oldfd: int, newfd: int) -> None:
     try:
         _reflink(oldfd, newfd)
     except OSError as e:
-        if e.errno not in {errno.EXDEV, errno.EOPNOTSUPP}:
+        if e.errno not in {errno.EXDEV, errno.EOPNOTSUPP, errno.ENOTTY}:
```
behrmann, line 586: As a general note (doesn't need to be changed): While in general sets are preferable for these checks, due to O(1) access, for such small sets tuples beat them due to memory locality (at least they did, when I last benchmarked this)

bluca (approved):
tested on a ppc64el machine, can confirm it now works - ship it!

behrmann:
Not much sense on wasting CI cycles then. We have no ppc in CI and the syntax checks out. Merging.

PR #1031, opened 2022-07-05
Make sure we extend the parent env instead of overriding it in run()

DaanDeMeyer:
This makes sure any environment variables set in the parent env
don't suddenly stop working when we set our own (e.g. proxy env
variables).

### DIFF
--- a/mkosi/backend.py
+++ b/mkosi/backend.py
@@ -752,6 +752,7 @@ def run(
     delay_interrupt: bool = True,
     stdout: _FILE = None,
     stderr: _FILE = None,
+    env: Mapping[str, Any] = {},
     **kwargs: Any,
 ) -> CompletedProcess:
     cmdline = [str(x) for x in cmdline]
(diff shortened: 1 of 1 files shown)


mkosi/backend.py
```diff
@@ -752,6 +752,7 @@ def run(
     delay_interrupt: bool = True,
     stdout: _FILE = None,
     stderr: _FILE = None,
+    env: Mapping[str, Any] = {},
```
behrmann, line 755: In principle this is not problematic, since this dict is not changed in the body of the function, but it is a potential waiting footgun, since (at least until late-bound arguments land) this is only initialised to the empty dict once and will be the same dict whenever `run` is invoked.

behrmann (commented):
LGTM, but has a potential for sharp edge.

mkosi/backend.py
DaanDeMeyer: Yeah, but I figure the typing gives sufficient protection against these type of footguns. I do agree that if the type was Dict, we wouldn't want to do this

PR #1032, opened 2022-07-05
Add --cache-initrd option

daandemeyer:
Building initrds is slow, let's add an option that allows building
the initrd as part of a cached image so that's it's only built once
and reused in any final images.


keszybz:
Looks reasonable.

PR #1033, opened 2022-07-06
Added support for Rolling Rhino Remix (https://rollingrhino.org)

mcassaniti:
Initial support for Rolling Rhino.

Note:

  * I may have misunderstood when to check `do_run_build_script`
  * Once https://github.com/rollingrhinoremix/rhino-convert/pull/2 is merged then the update process should be just a little bit smoother
  * The basic process is to install the latest Ubuntu development release (kinetic at this time) and run the conversion script provided upstream
  * I'm not associated with the Rolling Rhino project
  * If it speeds things up, I'm happy for changes to be done on my branch directly

### DIFF
--- a/mkosi/resources/convert_to_rolling_rhino.sh
+++ b/mkosi/resources/convert_to_rolling_rhino.sh
@@ -1,0 +1,65 @@
+#!/bin/bash
+
+set -e
+
+cd $(dirname $0)
+[ -d rhino-convert ] && rm -r rhino-convert
+[ -d convert ] && rm -r convert
+
+#TODO: Uncomment when the below PR is merged:
+#
+#   https://github.com/rollingrhinoremix/rhino-convert/pull/2
+#
+#git clone https://github.com/rollingrhinoremix/rhino-convert.git
+
+#TODO: Remove these lines when the above PR is merged
+git clone https://github.com/mcassaniti/rhino-convert.git
+cd rhino-convert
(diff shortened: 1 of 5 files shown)


mkosi/resources/convert_to_rolling_rhino.sh
```diff
@@ -0,0 +1,65 @@
...
+#
+#   https://github.com/rollingrhinoremix/rhino-convert/pull/2
+#
+#git clone https://github.com/rollingrhinoremix/rhino-convert.git
+
+#TODO: Remove these lines when the above PR is merged
+git clone https://github.com/mcassaniti/rhino-convert.git
+cd rhino-convert
+git checkout origin/convert-vs-creation-directories-v1 > /dev/null 2>&1
+cd ..
```
behrmann, line 19:
```suggestion
{
    cd rhino-convert
    git checkout origin/convert-vs-creation-directories-v1 &>/dev/null
}
```

behrmann: subshells are better for quickly going somewhere and popping back. Our shell style has no spaces between redirection operator and filename.

DaanDeMeyer:
This doesn't seem like something we should support upstream in mkosi itself. What you can do is use the shell script you're using here via --prepare-script to convert a regular ubuntu distro to rolling rhino

mcassaniti:
OK, I'll take that on board. Thanks for taking a look.

PR #1037, opened 2022-07-07
Drop Clear Linux support

DaanDeMeyer:
Due to lack of a maintainer, and deviating from established practices
almost universally (custom bootloader, custom package manager, ...),
it's not productive to keep support for Clear Linux in mkosi. We have
no idea if it works at all, no idea if something we do is going to
break it, and haven't received any feedback from Clear Linux users
for multiple years. Hence let's drop support for Clear Linux from
mkosi to reduce maintenance costs.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -297,10 +297,10 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
 `Distribution=`, `--distribution=`, `-d`
 
 : The distribution to install in the image. Takes one of the following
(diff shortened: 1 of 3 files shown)


poettering:
sounds ok to me. i think intel deemphesized clear linux a bit, so i guess there's no reason for us to keep this bitrotting, if it has no future.

behrmann (approved):

behrmann:
The removal should also get a changelog entry, but this can be done after the fact. Merging.

PR #1038, opened 2022-07-07
arch: Drop pacman kernel install hooks

DaanDeMeyer:
We shouldn't special case updates Arch if we don't do so on other
distros either. We don't guarantee availability of a package
manager in generated images, and the only reason Arch images ship
with pacman is because it's part of the base group. This might
change in the future and as such, there's no guarantee that Arch
images will always have pacman. Hence, let's remove the pacman hooks
we install for Arch to bring it more in line with the other distros.


behrmann:
LGTM. Well, I don't particularly think the removed stuff was a maintenance bother and generated images will not by default be upgradable I'm ok with that, if we document the removal of the hooks and and the addition of the symlinks for grub, so that people can work around that.

DaanDeMeyer:
Added a note to NEWS about the removal of the hooks. I don't think we need to mention the symlinks. That's just us conforming to Arch's interface for these things.

behrmann:
True, I'd mention the anyway, since people would need to remember to update those symlinks if they ever update the kernel in the image.

DaanDeMeyer:
I've added a note. Btw, did you look at https://github.com/systemd/mkosi/issues/1042 at all? That would allow us to drop a lot of bios code, including this hack.

DaanDeMeyer:
@behrmann Is this one good to go now?

behrmann:
I restarted the alma ext4 job, because the error looked weird. I can't reall have anything to do with the changes here, but I wanted to see whether it failed reproducibly, but it doesn't.

PR #1043, opened 2022-07-09
Fedora fixes

DaanDeMeyer:


mkosi/__init__.py
```diff
@@ -2116,8 +2121,7 @@ def invoke_dnf(
     with mount_api_vfs(args, root):
         run(cmdline, env=dict(KERNEL_INSTALL_BYPASS="1"))
 
-    distribution, release = detect_distribution()
-    if distribution not in (Distribution.debian, Distribution.ubuntu):
+    if args.distribution not in (Distribution.debian, Distribution.ubuntu):
```
behrmann, line 2124: `args.distribution` is the target distribution, but the idea was to return here if the host distribution is Debian (because of some changes in the path of the RPM db that has not yet/cannot happen in Debian)

behrmann (approved):

PR #1048, opened 2022-07-13
Drop Photon support

DaanDeMeyer:
Booting a systemd-nspawn container with a Photon image produced by
mkosi without --netdev waits for 2 minutes until "wait for network
to come online" times out. After logging in there's 3 failed services
in systemctl status. This is without trying to use any of mkosi's
more advanced features.

Bootable images are not supported at all on Photon. Given this and
the broken container support, it's clear that Photon support isn't
really usable at the moment and probably hasn't been for quite a
while (see #664).

Given that Photon has been broken for multiple releases and that
aside form #664, no one has bothered to report concrete issues or
make an attempt to fix the issues aside from a mirror update, this
gives us a pretty clear indication that no one is using or trying
to use mkosi to build Photon images.

Given that none of the existing maintainers are familiar with Photon
(and are likely not interested in doing the effort needed to support
it), let's drop the Photon support from mkosi.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 - Support for Clear Linux was dropped. See https://github.com/systemd/mkosi/pull/1037
 for more information.
+- Support for Photon was dropped. See https://github.com/systemd/mkosi/pull/1048
+for more information.
 - The Arch kernel/bootloader pacman hooks were removed. For anyone that still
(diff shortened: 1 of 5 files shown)


mrc0mmand (approved):
LGTM. The Photon support has been in shambles for a long time without nobody stepping up to fix it, so dropping it makes sense. It can be easily re-added in the future if there's interest.

PR #1051, opened 2022-07-13
Clover experiments

DaanDeMeyer:
This uses Clover to emulate UEFI on BIOS and to chainload into systemd-boot. It works, but with a few caveats:

- Using qemu virtio disks does not work (qemu hangs on boot)
- Using the qemu serial console without graphical interface does not work (qemu hangs on boot)
- Using -cpu max causes a hardware exception

We can generally work around these but it makes me wonder how well this would work on actual BIOS hardware though.

This is just me experimenting, the code is in no state to be merged.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3440,6 +3440,7 @@ def install_boot_loader(
     if not args.bootable or do_run_build_script:
         return
     assert loopdev is not None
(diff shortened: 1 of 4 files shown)


DaanDeMeyer:
Closing for now due to lack of support from upstream

PR #1052, opened 2022-07-15
Pass kernel args to systemd when booting in nspawn

DaanDeMeyer:
When booting systemd in systemd-nspawn, additional arguments can
be passed to systemd via nspawn's command line. The arguments get
interpreted by systemd in the container as if they were kernel
command line args.

To take advantage of this, let's pass any configured kernel command
line arguments to systemd running in nspawn as well. For example,
this allows using systemd.setenv to set environment variables as
early as possible for both qemu VMs and nspawn containers.


mrc0mmand (approved):
LGTM, but needs rebase now

PR #1053, opened 2022-07-15
Only pass arguments to build script when verb is "build"

DaanDeMeyer:
When we're building the image because it's required for another verb,
any passed arguments are most likely intended for the target verb,
and not for "build", so don't add them in that case.


bluca (approved):
Maybe add a comment to clarify it, as it might not be immediately obvious

PR #1055, opened 2022-07-15
Move debian/ubuntu locale symlink logic to install function

DaanDeMeyer:
Let's get rid of another distribution check by simply creating the
symlink in the install function.


mkosi/__init__.py
```diff
@@ -2357,6 +2348,14 @@ def install_fedora(args: MkosiArgs, root: Path, do_run_build_script: bool) -> No
     #        But we set LANG=C.UTF-8 anyway.
     shutil.rmtree(root / "usr/share/locale", ignore_errors=True)
 
+    # Debian/Ubuntu use a different path to store the locale so let's make sure that path is a symlink to
```
bluca, line 2351: I'm a bit confused, this is ubuntu specific but it's being moved inside the `install_fedora` function?

DaanDeMeyer: Euhhhh yikes

bluca (approved):

PR #1056, opened 2022-07-15
Remove dracut uefi_stub configuration

DaanDeMeyer:
We don't use dracut's --uefi option anymore so we don't need this
config anymore either.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1750,16 +1750,6 @@ def configure_dracut(args: MkosiArgs, packages: Set[str], root: Path) -> None:
         )
 
     if args.get_partition(PartitionIdentifier.esp):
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -3175,9 +3166,6 @@ def install_gentoo(
         gentoo.invoke_emerge(args, root, pkgs=gentoo.pkgs_fs)
 
     if not do_run_build_script and args.bootable:
-        # Please don't move, needs to be called before installing dracut
-        # dracut is part of gentoo_pkgs_boot
```
bluca, line 3189: what about this comment? no longer valid?

DaanDeMeyer: Well I mean gentoo still boots fine, so I assume there's no problem. We run kernel-install manually later which runs dracut so all should be good

bluca (approved):

PR #1061, opened 2022-07-16
Copy systemd manager conf to initrd if set

DaanDeMeyer:


behrmann:
LGTM. I always assumed dracut would pick that up by itself? Any idea why it doesn't do that?

DaanDeMeyer:
No idea tbh, probably to avoid unintended changes to the initrd when a user modifies systemd's config on the host?

PR #1063, opened 2022-07-17
Handle directories when copying untracked git files

DaanDeMeyer:


behrmann:
LGTM. 

Looking at this maybe a change for the future: While I do think it is a sensible assumption that all locales nowadays should be UTF-8, we don't really need to do any decoding (and risk decoding errors). Paths can also just straight up take bytes. This would make this a bit easier to read, too.

PR #1064, opened 2022-07-18
Fixed running prepare and post-install scripts when UsrOnly is in use

mcassaniti:


DaanDeMeyer (changes requested):

mkosi/backend.py
```diff
@@ -35,6 +35,7 @@
     Union,
     cast,
 )
+from . import root_home
```
DaanDeMeyer, line 38: Please move root_home from __init__.py to backend.py as well. We don't want imports from the main file into backend.py

DaanDeMeyer (changes requested):

DaanDeMeyer:
Please squash all commits together into a single commit

mcassaniti:
I've only ever done a squash commit when merging so I created another PR. I'm sorry if it's not the way it's usually done.

behrmann:
@mcassaniti No worries. It's quite easy. You squash the commits (like you already did) and then you force (I recommend `--force-with-lease`) push the branch behind this PR (`usr-only-with-scripts`), e.g.
```
# do your squash
git rebase -i
git push --force-with-lease origin usr-only-with-scripts
```

If you did your squashing on a different branch, you can push *that* to your upstream branch instead, this could all in all look something like this
```
# new version
git checkout -b usr-only-with-scripts-v2 usr-only-with-scripts
# squashing
git rebase -i
# push the new version to this PR
git push --force-with-lease origin usr-only-with-scripts-v2:usr-only-with-scripts
```

PR #1066, opened 2022-07-18
Added support for generating an XFS UsrOnly root

mcassaniti:
When the options `UsrOnly = yes` and `Format = gpt_xfs` are combined, `mkosi` throws an exception because the FS image is never created or populated. There is no error beforehand that this combination of options is invalid.

The only issue I can see with the code I've added is that XFS doesn't seem to support a minimize option similar to ext4 after the FS is populated.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3654,6 +3654,39 @@ def generate_btrfs(args: MkosiArgs, root: Path, label: str, for_cache: bool) -> 
     return f
 
 
+def generate_xfs(args: MkosiArgs, root: Path, label: str, for_cache: bool) -> Optional[BinaryIO]:
+    if args.output_format != OutputFormat.gpt_xfs:
+        return None
+    if for_cache:
+        return None
+
+    @contextlib.contextmanager
+    def get_loopdev(f: BinaryIO) -> Iterator[Path]:
+        c = run(["losetup", "--find", "--show", f.name], stdout=PIPE, text=True)
+        l = Path(c.stdout.strip())
+        try:
+            yield l
+        finally:
+            run(["losetup", "--detach", l])
+
+    with complete_step("Creating xfs root file system…"):
+        f: BinaryIO = cast(
+            BinaryIO,
+            tempfile.NamedTemporaryFile(prefix=".mkosi-mkfs-xfs", dir=os.path.dirname(args.output))
+        )
+
+        f.truncate(args.root_size)
+        run(mkfs_xfs_cmd(label) + [f.name])
+
+        xfs_dir = workspace(root) / "xfs"
+        xfs_dir.mkdir()
+        with get_loopdev(f) as loopdev:
+            with mount_loop(args, loopdev, xfs_dir) as mp:
+                copy_path(root, mp)
+
+    return f
+
+
 def make_generated_root(args: MkosiArgs, root: Path, for_cache: bool) -> Optional[BinaryIO]:
 
     if not is_generated_root(args):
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -3898,6 +3898,38 @@ def generate_btrfs(args: MkosiArgs, root: Path, label: str, for_cache: bool) ->
...
+
+        f.truncate(args.root_size)
+        run(mkfs_xfs_cmd(label) + [f.name])
+
+        with tempfile.TemporaryDirectory(prefix=".mkosi-xfs-mnt", dir=os.path.dirname(args.output)) as td:
+            with get_loopdev(f) as loopdev:
+                mount_point = Path(td)
+                mount_loop(args, loopdev, mount_point)
+                copy_path(root, mount_point)
+                umount(mount_point)
```
DaanDeMeyer, line 3928: Please use a XFS "protofile" instead, see the manpage: https://man7.org/linux/man-pages/man8/mkfs.xfs.8.html

DaanDeMeyer (changes requested):

mkosi/__init__.py
mcassaniti: If I have this right I'd need to write a generator for a protofile based on the root directory and use this to load the file system contents. Is this what you're asking for? It seems the contents of such a file would certainly be quite big.

DaanDeMeyer: Nvm, It seems I misunderstood the option. It doesn't do what I thought it did

mcassaniti: So what's next?

DaanDeMeyer: I guess we'll have to do it like this? Ideally mkfs.xfs would support this via an option just like mkfs.btrfs and mkfs.ext4. I would like #1070 to land first with the changes to mount_loop()

DaanDeMeyer: Please rebase and update the PR to use the new `mount_loop()` context manager

mcassaniti:
Anything else holding this one up?

mkosi/__init__.py
```diff
@@ -3654,6 +3654,39 @@ def generate_btrfs(args: MkosiArgs, root: Path, label: str, for_cache: bool) ->
...
+    with complete_step("Creating xfs root file system…"):
+        f: BinaryIO = cast(
+            BinaryIO,
+            tempfile.NamedTemporaryFile(prefix=".mkosi-mkfs-xfs", dir=os.path.dirname(args.output))
+        )
+
+        f.truncate(args.root_size)
+        run(mkfs_xfs_cmd(label) + [f.name])
+
+        with tempfile.TemporaryDirectory(prefix=".mkosi-xfs-mnt", dir=os.path.dirname(args.output)) as td:
```
DaanDeMeyer, line 3681: We can just mount to `xfs` under `workspace(root)`, no need to use a separate temporary directory

DaanDeMeyer:
I wrote these comments ages ago but forgot to submit, apologies

mkosi/__init__.py
mcassaniti: > We can just mount to `xfs` under `workspace(root)`, ...

I don't understand. Where am I supposed to mount the loopback device to copy the file system contents into?

DaanDeMeyer: You can use `workspace(root) / "xfs"` as the directory to mount to instead of a temporary directory. Of course you'll have to create it first.

mkosi/__init__.py
```diff
@@ -3654,6 +3654,39 @@ def generate_btrfs(args: MkosiArgs, root: Path, label: str, for_cache: bool) ->
...
+            with get_loopdev(f) as loopdev:
+                mount_point = Path(td)
+                with mount_loop(args, loopdev, mount_point):
+                    copy_path(root, mount_point)
```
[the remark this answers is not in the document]
mcassaniti: Fixed

mkosi/__init__.py
mcassaniti: Fixed

DaanDeMeyer (approved):

mcassaniti:
Thanks for all the effort @DaanDeMeyer

PR #1069, opened 2022-07-18
Remove mkosi.files/ directory

DaanDeMeyer:
These configs are just bitrotting and we don't use them for anything.
Let's drop the directory.


keszybz:
Dropping stuff is usually good ;)

PR #1070, opened 2022-07-18
Refactor mount functions to use context managers

DaanDeMeyer:
Let's have a single mount() function that's a context manager and
build everything else on top of that. To handle the non-fixed amounts
of context managers, we use ExitStack().

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1458,30 +1458,41 @@ def delete_whiteout_files(path: Path) -> None:
                 os.unlink(entry.path)
 
 
-def do_mount(
+@contextlib.contextmanager
+def mount(
         what: PathString,
         where: Path,
+        operation: Optional[str] = None,
         options: Sequence[str] = (),
         type: Optional[str] = None,
         read_only: bool = False,
-) -> None:
+) -> Iterator[Path]:
     os.makedirs(where, 0o755, True)
 
     if read_only:
         options = ["ro", *options]
 
-    cmd: List[PathString] = ["mount", "-n", what, where]
+    cmd: List[PathString] = ["mount", "--no-mtab"]
+
+    if operation:
+        cmd += [operation]
+
+    cmd += [what, where]
 
     if type:
-        cmd += ["-t", type]
+        cmd += ["--types", type]
 
     if options:
-        cmd += ["-o", ",".join(options)]
+        cmd += ["--options", ",".join(options)]
 
-    run(cmd)
+    try:
+        run(cmd)
+        yield where
+    finally:
+        run(["umount", "--no-mtab", "--recursive", where])
 
 
-def mount_loop(args: MkosiArgs, dev: Path, where: Path, read_only: bool = False) -> None:
+def mount_loop(args: MkosiArgs, dev: Path, where: Path, read_only: bool = False) -> ContextManager[Path]:
     options = []
     if not args.output_format.is_squashfs():
         options += ["discard"]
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -1449,30 +1449,41 @@ def delete_whiteout_files(path: Path) -> None:
...
 
     if options:
         cmd += ["-o", ",".join(options)]
 
-    run(cmd)
+    try:
+        run(cmd)
+        yield where
+    finally:
+        run(["umount", "-n", where])
```
behrmann, line 1483:
```suggestion
        run(["umount", "--no-mtab", where])
```
since it's nicer to follow and easier to look up when one doesn't have the umount options in one's head.

The idea here is to not use `--recursive`, since when exiting the stack all the exiting umounts will be popped of the stack, right? A `--recursive` wouldn't hurt, though, and might clean up things even if an umount fails (since I think the `__exit__`s from the context manager will be called regardless even on an exception)

behrmann (commented):
Only had a cursory look, but looks very nice.

mkosi/__init__.py
DaanDeMeyer: I figure since it works without --recursive that's also nice though, since it lets us know when we're forgetting to unmount something.

DaanDeMeyer: Added it back, probably safer to keep it there

mkosi/__init__.py
```diff
@@ -1521,103 +1531,79 @@ def mount_overlay(
...
-    umount(realroot)
-
-    workdir.cleanup()
-
-    # Let's now move the contents of realroot into root
-    for entry in os.scandir(realroot):
-        os.rename(realroot / entry.name, root / entry.name)
-    realroot.rmdir()
+    try:
+        with workdir, mount("overlay", realroot, options=options, type="overlay", read_only=read_only), mount_bind(realroot, root):
```
keszybz, line 1535: I think it's one of the cases where judicious use of line continuations would be more readable…

mkosi/__init__.py
```diff
@@ -1521,103 +1531,79 @@ def mount_overlay(
...
     base_image: Optional[Path],  # the path to the mounted base image root
     loopdev: Optional[Path],
     image: LuksSetupOutput,
     root_read_only: bool = False,
 ) -> Iterator[None]:
-    with complete_step("Mounting image…"):
-
-        realroot: Optional[Path] = None
-        workdir: Optional[TempDir] = None
+    with complete_step("Mounting image…", "Unmounting image…"), contextlib.ExitStack() as stack:
```
keszybz, line 1558: Here and below too…

keszybz (commented):
Looks reasonable. I remember that getting the cleanup for overlayfs was tough, but if the new approach works, that's all that matters.

Some formatting suggestions…

behrmann (approved):
LGTM. I find the exit stack a bit harder to read, but all in all I think this is an improvement. 

The linebreaks @keszybz mentioned would be nice, though.

mkosi/__init__.py
DaanDeMeyer: I moved the call to `mount` out of the `with` statement into a variable to shorten the line

mkosi/__init__.py
DaanDeMeyer: Adding line breaks to a `with` statements means using `\` at the end though? We can't yet use parentheses for multi line with because that got added after python 3.7 which is our minimum version atm

DaanDeMeyer:
Fixed one case but I'm not sure about the others

DaanDeMeyer:
Let's go ahead and merge this, I'll do a follow up PR to fix the formatting for the other cases if it's still needed.

PR #1071, opened 2022-07-19
Always define args.architecture and add args.host_architecture

behrmann:
I've now stumbled over `args.architecture` being `None` multiple times and since we have `args.architecture or platform.machine()` in a few places, I think it makes sense to always define `args.architecture` and default it to `platform.machine()`

There's a few places where we check whether `args.architecture` is different from the host's architecture I added `args.host_architecture` that's not plumbed through to the CLI and which always has the value of `platform.machine()`

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2102,7 +2102,7 @@ def invoke_dnf(
     if args.with_network == "never":
         cmdline += ["-C"]
 
-    if args.architecture is not None:
+    if not args.architecture_is_native():
         cmdline += [f"--forcearch={args.architecture}"]
 
     if not args.with_docs:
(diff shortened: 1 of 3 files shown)


mkosi/__init__.py
```diff
@@ -2102,7 +2102,7 @@ def invoke_dnf(
     if args.with_network == "never":
         cmdline += ["-C"]
 
-    if args.architecture is not None:
+    if args.architecture != args.host_architecture:
```
behrmann, line 2105: While I know that calling `debootstrap` with `--arch` is not problematic, I was unsure about dnf's `--forcearch` and went for this check here.

It would be great if the dnf experts could tell me whether this makes sense.

mkosi/backend.py
```diff
@@ -415,7 +415,8 @@ class MkosiArgs:
     repositories: List[str]
     use_host_repositories: bool
     repos_dir: Optional[str]
-    architecture: Optional[str]
+    architecture: str
+    host_architecture: str
```
DaanDeMeyer, line 419: I don't think we should store this in `host_architecture`. We can add a more clearly named function to get the current host architecture and just use that everywhere.

DaanDeMeyer: That'll do the trick

mkosi/__init__.py
DaanDeMeyer: It's probably fine if CI passes

PR #1072, opened 2022-07-19
Use distro~release subdirectories under mkosi.output

DaanDeMeyer:
Currently, when using incremental mode, building for a different
release or distribution means throwing away the cached images for
the previous distribution or release used unless each distro/release
combo is configured with an explicit output directory. Let's try to
be smarter here, by using the same logic as used for the cache path.
We create distro~release subdirectory under mkosi.output/ and use that
as the output directory. This makes sure cached images stay intact
even if we build for a different distribution.

This will end up using slightly more disk space when building for many
different distros when using mkosi.output/, but this should be a good
tradeoff to make regardless. If looking to regain disk space, a user
simply has to remove the output subdirectory for the distros he's not
interested in keeping.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,10 @@ for more information.
 wants to use them, they can be found [here](https://github.com/systemd/mkosi/tree/v13/mkosi/resources/arch).
 When building a bios image, /boot/vmlinuz-kver and /boot/initramfs-kver.img are
 now symlinks to the actual files as installed by kernel-install.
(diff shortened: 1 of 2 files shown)


behrmann:
Seems very sensible to me. Only thing missing is a changelog entry.

DaanDeMeyer:
Added the same change for mkosi.builddir/ as well.

bluca (approved):

PR #1073, opened 2022-07-21
Fixed regression when using UsrOnly during initial install

mcassaniti:
Several distributions call `run_workspace_command()` during installation and this now requires the root directory that gets bind mounted in to exist. The created directory will be removed anyway since it exists under a temporary workspace.

As an example, the documentation removal when installing a Debian derivative will fail without this change.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1841,7 +1841,7 @@ def prepare_tree(args: MkosiArgs, root: Path, do_run_build_script: bool, cached:
             root.joinpath("etc/kernel/entry-token").write_text(f"{args.machine_id}\n")
             root.joinpath("etc/kernel/install.conf").write_text("layout=bls\n")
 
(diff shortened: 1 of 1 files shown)


DaanDeMeyer (changes requested):
This should be changed in prepare_os_tree to create root_home() there if usr_only is set

DaanDeMeyer:
Can you provide a stacktrace of the failure you're seeing as well?

mcassaniti:
Traceback/error from final image phase:
```
I: Base system installed successfully.
Failed to stat /home/ubuntu/mkosi/mkosi.workspace/evxnh6m0/home-root: No such file or directory
‣ Error: Workspace command /usr/bin/apt-get --assume-yes update returned non-zero exit code 1.
‣  (Unmounting Package Cache)
‣  (Unmounting image)
‣  (Detaching /dev/loop10)
```

PR #1080, opened 2022-07-23
Allow running qemu as an unprivileged user

### LINKED ISSUE #1076: Allow running qemu as an unprivileged user
The man page suggests that `mkosi qemu` can be run as an unprivileged user under some circumstances:

> This command must be executed as root unless the image already exists and -f is not specified.

but in practice mkosi's [sudo check](https://github.com/systemd/mkosi/blob/e2681d2adbaabe1065f7cff307368d0ff0d69bb5/mkosi/__init__.py#L126) prevents this.

On my machine (Arch Linux with Wayland/Sway), qemu fails to start if invoked through sudo (with the message `Authorization required, but no authorization protocol specified`), but after removing `Verb.qemu` from the sudo check in a local checkout of the source code I was able to build an image (as root) and run it with `mkosi qemu` as my normal user.  There are clearly limitations in running qemu as an unprivileged user (setting up netdev requires root, for example), but it seems to be working well enough that I'd prefer if the qemu verb were removed from the sudo check.

I only discovered mkosi a few days ago and had never delved very deeply into qemu's workings beforehand, so it's quite possible there are potential dangers with this setup I'm not aware of.  If so, I can understand the desire to limit the tool to known-working configurations.  But given the indication from the docs, I wanted to see if this was something the maintainers would be willing to enable.

jre21:
When testing on my system (Wayland/Sway on Arch Linux), I was able to successfully boot an image as an unprivileged user when running `mkosi qemu` with default arguments.

I'd appreciate a sanity check on my changes to the `unlink_output()` handling.  On reading through the code, it seemed to me like the conditionals before and after entering the function were trying to express "do we need a rebuild?" and thus a better fit for `needs_build()` than for `MKOSI_COMMANDS_SUDO`.  I think the new check is equivalent to old checks except that, if the output files don't exist when we start a build, we attempt to unlink them anyway (which seems to be harmless).

Running `mkosi qemu` with `Netdev=true` as an unprivileged user results in qemu being unable to set up the network tunnel and erroring out, and I'm a bit ambivalent about how to deal with that.  Possible approaches:

1. Fall back to usermode networking (I'm not currently sure whether this works, since I haven't set up a SPICE client yet).
2. Let qemu error out and trust the user to figure out what's going on (somewhat unfriendly, especially to less experienced sysadmins).
3. Trigger `check_root()` when `args.netdev` is set.

I also haven't exhaustively checked the other qemu flags mkosi can set to see if any require root privileges, but none of them seem especially likely to.  Obviously users could get into trouble with custom `QemuArgs`, but I think that at that point it's reasonable to expect them to do their own debugging legwork.

This resolves #1076.

### DIFF
--- a/tests/test_config_parser.py
+++ b/tests/test_config_parser.py
@@ -600,7 +600,7 @@ class MkosiConfigManyParams(MkosiConfigOne):
             "Host": {
                 "ExtraSearchPaths": "search/here:search/there",
                 "QemuHeadless": True,
-                "QemuArgs": "-vga none -device virtio-vga-gl",
+                "QemuArgs": "-device virtio-vga-gl -vga none",
                 "Netdev": True,
             },
         }
@@ -734,7 +734,7 @@ class MkosiConfigManyParams(MkosiConfigOne):
             "Host": {
                 "ExtraSearchPaths": "search/debi",
                 "QemuHeadless": True,
-                "QemuArgs": "-nic user,model=virtio-net-pci",
+                "QemuArgs": "-device virtio-vga-gl,xres=1920,yres=1080 -display sdl,gl=on",
                 "Netdev": True,
             },
         }
--- a/tests/test_config_parser.py
+++ b/tests/test_config_parser.py
@@ -174,7 +174,7 @@ class MkosiConfig:
                 is_eq = False
         return is_eq
 
-    def _append_list(self, ref_entry: str, new_args: Any, job_name: str = DEFAULT_JOB_NAME, separator: str = ",") -> None:
+    def _append_list(self, ref_entry: str, new_args: Any, job_name: str = DEFAULT_JOB_NAME, separator: str = ",", with_duplicates: bool = False) -> None:
         """Helper function handling comma separated list as supported by mkosi"""
         args_list = []
         if isinstance(new_args, str):
@@ -189,7 +189,7 @@ class MkosiConfig:
             if isinstance(arg, str) and arg.startswith("!"):
                 if arg[1:] in self.reference_config[job_name][ref_entry]:
                     self.reference_config[job_name][ref_entry].remove(arg[1:])
-            elif arg not in self.reference_config[job_name][ref_entry]:
+            elif with_duplicates or arg not in self.reference_config[job_name][ref_entry]:
                 self.reference_config[job_name][ref_entry].append(arg)
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


jre21:
I've confirmed that user-mode networking is usable when running QEMU with SPICE enabled,  I've updated the pull request to use it instead of TAP when not running as root.

jre21:
In the process of putting together this pull request, I ran into two issues with how QEMU arguments are parsed.

1.The QEMU command line may contain repeated tokens (e.g., when attaching multiple devices to a VM), and we need to not remove those duplicates.  I've already included a fix for this.
2. Because QEMU arguments start with a single hyphen, argparse can confuse the qemu-args argument for a series of short options if the first character matches one of the recognized short options.  Any fix will require fiddling with argparse internals more than we already are, but I'll check tomorrow to see if I can find a relatively simple way to get it to parse correctly.

The upshot is that QEMU's argument parsing is _extremely_ nonstandard, and I've chosen to name the new action I'm writing to cover it QemuAction instead of a more generic name to reflect the fact that this is enough of a special case that I don't expect it to be reusable in any future command line arguments.

Since the scope of this pull request is growing beyond the problem I originally set out to solve, let me know if you'd prefer that I move the parser changes into a separate PR.

jre21:
The fix for the second parse error turned out to be cleaner than I anticipated, so I managed to avoid piling more magic onto the argparse action for qemu.  This change is now ready for review, assuming the integration tests pass.

tests/test_config_parser.py
```diff
@@ -378,6 +378,8 @@ def _update_ref_from_file(self, mk_config: Dict[str, Any], job_name: str = DEFAU
                 self._append_list("extra_search_paths", mk_config_host["ExtraSearchPaths"], job_name, ":")
             if "QemuHeadless" in mk_config_host:
                 self.reference_config[job_name]["qemu_headless"] = mk_config_host["QemuHeadless"]
+            if "QemuArgs" in mk_config_host:
+                self._append_list("qemu_args", mk_config_host["QemuArgs"], job_name, " ", True)
```
DaanDeMeyer, line 382: Please don't just pass `True`, use named arguments so we know what we're setting to `True`

jre21:
I addressed the feedback comment.  Could you take another look at the latest version?

cryptoknight:
Was just about to submit an issue for the QemuArgs deduplication issue and start trying to solve. Thanks for the fixes!

DaanDeMeyer (approved):

PR #1081, opened 2022-07-25
Fix debian releases w/o updates and security fixes

rezib:
Actually Debian testing release has security updates available, metadata are available at these URL:

http://ftp.debian.org/debian/dists/testing-updates/
http://security.debian.org/debian-security/dists/testing-security/

However, sid (the other name of unstable) does not have additional security updates repositories, as fixes land in sid main repository directly.

This PR fixes the set of Debian releases for which the security and updates repositories are not deployed.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2860,7 +2860,7 @@ def install_debian_or_ubuntu(args: MkosiArgs, root: Path, *, do_run_build_script
             if "VERSION_ID" not in os_release and "BUILD_ID" not in os_release:
                 f.write(f"BUILD_ID=mkosi-{args.release}\n")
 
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
cc @bluca

DaanDeMeyer:
cc @257 for failing Gentoo CI

bluca (approved):
I mean there are repos, they are just never used, they'll always be empty... but I guess it doesn't hurt, and one day they might be used

257:
> cc @257 for failing Gentoo CI

i have a [PR](https://github.com/gentoo/baselayout/pull/2) for upstream, can Gentoo fails be ignored for a while?

PR #1083, opened 2022-07-26
arch: debug

DaanDeMeyer:


jre21:
I've run into issues similar to this in the past (Arch Linux can't find a PGP key it needs to import).  I've usually worked around it by [changing pacman's configured key server](https://wiki.archlinux.org/title/Pacman/Package_signing#Change_keyserver).

jre21:
Another possibly-relevant oddity: mkosi appears to be setting the pacman keyring to a directory [outside the image root](https://github.com/systemd/mkosi/blob/fde6e2e264193600e105ab383775c6a51e7473f8/mkosi/__init__.py#L2981).

DaanDeMeyer:
> Another possibly-relevant oddity: mkosi appears to be setting the pacman keyring to a directory [outside the image root](https://github.com/systemd/mkosi/blob/fde6e2e264193600e105ab383775c6a51e7473f8/mkosi/__init__.py#L2981).

That's intentional. It saves a lot of time rebuilding the keyring on every build.

DaanDeMeyer:
This is fixed by #1084 so closing

PR #1084, opened 2022-07-26
action: bump archlinux-keyring to 20220713

mrc0mmand:
See: https://bugs.archlinux.org/task/75427

### DIFF
--- a/action/setup-pacman.sh
+++ b/action/setup-pacman.sh
@@ -2,7 +2,7 @@
 set -e
 
 PACMAN_VERSION="6.0.1"
-ARCHLINUX_KEYRING_VERSION="20220424"
+ARCHLINUX_KEYRING_VERSION="20220713"
 
 apt-get --assume-yes --no-install-recommends install \
         gcc \


DaanDeMeyer:
Thanks! this is a lot simpler than what I'd have tried

DaanDeMeyer (approved):

mrc0mmand:
Oh well, I though this was way too simple - now pacman doesn't segfault, but still fails to import the signature:

```
 error: libcap: key "991F6E3F0765CF6295888586139B09DA5BF0D338" is unknown
:: Import PGP key 991F6E3F0765CF6295888586139B09DA5BF0D338? [Y/n] error: key "991F6E3F0765CF6295888586139B09DA5BF0D338" could not be imported

:: File /home/runner/work/mkosi/mkosi/mkosi.output/.mkosi-j2iy9772/root/var/cache/pacman/pkg/libcap-2.65-1-x86_64.pkg.tar.zst is corrupted (invalid or corrupted package (PGP signature)).
Do you want to delete it? [Y/n] error: failed to commit transaction (invalid or corrupted package)
```

*sigh*

mrc0mmand:
But I have another idea, gimme a sec.

DaanDeMeyer:
Do we need both or could bumping the keyring package version be sufficient?

mrc0mmand:
> Do we need both or could bumping the keyring package version be sufficient?

I just tried it without the first commit locally and it seems to be working (it didn't for me when I tried it for the first time, but that's probably on me), so let me drop the workaround and see how it's going to fare.

mrc0mmand:
Oh well, apparently it was indeed an issue on my side, so bumping the keyring version is enough.

DaanDeMeyer:
Cool, let's do that for now.

It'd be kinda nice if pacman/archlinux-keyring got packaged in ubuntu

DaanDeMeyer:
No need to wait for the rest

mrc0mmand:
> Cool, let's do that for now.
> 
> It'd be kinda nice if pacman/archlinux-keyring got packaged in ubuntu

Yeah, that would make things definitely much easier.

PR #1085, opened 2022-07-26
action: Install dependencies in action repo checkout

DaanDeMeyer:
Since we symlink systemd-nspawn from the build directory to /usr/bin,
if the directory that nspawn was built in is cleaned up (e.g. via
git clean), systemd-nspawn is gone as well. To make it harder to
accidentally delete systemd-nspawn, let's clone and build all the
dependencies we compile from source in the repo checkout of the
action instead of in the repo of the project that's using the mkosi
action.

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -21,6 +21,7 @@ runs:
 
   - name: Dependencies
     shell: bash
+    working-directory: ${{ github.action_path }}
     run: |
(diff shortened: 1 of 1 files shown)


mrc0mmand (approved):
LGTM, thanks! The CI fail looks unrelated.

PR #1086, opened 2022-07-27
Skip setting kernel root cmdline parameter for UsrOnly without verity

mcassaniti:
When an image is built as a /usr only image and does not use verity, the kernel command line should (and already does) contain the parameter `mount.usr`. It should not contain a root parameter since `systemd-repart` and `systemd-gpt-auto-generator` will take care of generating and mounting the root partition.

Without this change the root partition label used on the kernel command line will be invalid.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -4077,7 +4077,8 @@ def install_unified_kernel(
                 option = "usrhash" if args.usr_only else "roothash"
                 boot_options = f"{boot_options} {option}={root_hash}"
             elif partlabel:
-                boot_options = f"{boot_options} root=PARTLABEL={partlabel}"
+                option = "mount.usr" if args.usr_only else "root"
+                boot_options = f"{boot_options} {option}=PARTLABEL={partlabel}"
 
             osrelease = root / "usr/lib/os-release"
             cmdline = workspace(root) / "cmdline"


mcassaniti:
`extract_kernel_cmdline()` might need some work too.

mkosi/__init__.py
```diff
@@ -4279,6 +4279,9 @@ def install_unified_kernel(
             if root_hash:
                 option = "usrhash" if args.usr_only else "roothash"
                 boot_options = f"{boot_options} {option}={root_hash}"
+            elif args.usr_only and not args.verity:
+                # Leave systemd-gpt-auto-generator to mount the root partition
+                pass
```
DaanDeMeyer, line 4284: How does this accomplish what you want? `root=` is only set if the image has an image ID. If you don't have an image ID, `root=` won't be set

mcassaniti: Sorry, I do have an image ID set, but you want to use it as `mount.usr=/dev/disk/by-partlabel/<ImageID>_<ImageVersion>` (as is set currently) without also setting `root=PARTLABEL=<ImageID>_<ImageVersion>` for this particular case.

DaanDeMeyer: What if we just set `mount.usr` here instead of `root` if `usr_only == True`? Does that also address the issue?

mcassaniti: `mount.usr` is already set over [here](https://github.com/systemd/mkosi/blob/f9ebc563877583d9d49d8d566ea0a65c33f01b7d/mkosi/__init__.py#L6864) so by the time it comes round to this bit of code it's only `root=` that is being set or skipped. The grand total is a kernel command line with `mount.usr` and no `root` parameter.

DaanDeMeyer: Yes, you'll end up mount.usr set 2x but the last occurrence will override so if we set `mount.usr` here instead of `root`, things should still work.

The entire calculation of kernel command line arguments needs to be reworked but the most "correct" fix atm that doesn't touch too much code is to set `mount.usr` here if `args.usr_only == True`

mcassaniti: OK, I've added `mount.usr` instead of skipping.

mkosi/__init__.py
```diff
@@ -4280,7 +4280,10 @@ def install_unified_kernel(
                 option = "usrhash" if args.usr_only else "roothash"
                 boot_options = f"{boot_options} {option}={root_hash}"
             elif partlabel:
-                boot_options = f"{boot_options} root=PARTLABEL={partlabel}"
+                if args.usr_only:
+                    boot_options = f"{boot_options} mount.usr=/dev/disk/by-partlabel/{partlabel}"
+                else:
+                    boot_options = f"{boot_options} root=PARTLABEL={partlabel}"
```
DaanDeMeyer, line 4286: `mount.usr` takes the same options as `root`, hence please use `mount.usr=PARTLABEL` here as well, so the following:

```
option = "mount.usr" if args.usr_only else "root"
boot_options = f"{boot_options} {option}=PARTLABEL={partlabel}
```

mcassaniti: Fixed

PR #1090, opened 2022-07-27
Only check whether a format can be booted if verb == qemu

DaanDeMeyer:
We can still build these images without any problems, we just can't
boot them, so let's still allow building the images, just not booting
them.


behrmann (approved):
CI failure for Arch is unrelated

PR #1094, opened 2022-07-28
Ignore EBUSY on partition removal

DaanDeMeyer:
This can happen just after inserting a partition, probably because
udev starts operating on a device which makes us unable to remove
the partition, let's ignore EBUSY as a workaround until we get a
proper fix

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -984,7 +984,13 @@ def ioctl_partition_add(fd: int, nr: int, start: int, size: int) -> None:
 def ioctl_partition_remove(fd: int, nr: int) -> None:
     bp = blkpg_partition(pno=nr)
     ba = blkpg_ioctl_arg(op=BLKPG_DEL_PARTITION, data=ctypes.addressof(bp), datalen=ctypes.sizeof(bp))
-    fcntl.ioctl(fd, BLKPG, ba)
+    try:
+        fcntl.ioctl(fd, BLKPG, ba)
+    except OSError as e:
+        if e.errno != errno.EBUSY:
+            raise
+        else:
+            warn("Got EBUSY from the kernel while trying to remove partition, ignoring")
 
 
 @contextlib.contextmanager


mkosi/__init__.py
```diff
@@ -984,7 +984,13 @@ def ioctl_partition_add(fd: int, nr: int, start: int, size: int) -> None:
...
     bp = blkpg_partition(pno=nr)
     ba = blkpg_ioctl_arg(op=BLKPG_DEL_PARTITION, data=ctypes.addressof(bp), datalen=ctypes.sizeof(bp))
-    fcntl.ioctl(fd, BLKPG, ba)
+    try:
+        fcntl.ioctl(fd, BLKPG, ba)
+    except OSError as e:
+        if e.errno != errno.EBUSY:
+            raise
+        else:
+            warn("Got EBUSY from the kernel while trying to remove partition, ignoring")
```
behrmann, line 993: Should we maybe retry after a short wait time, i.e. wrap this try in a loop?

DaanDeMeyer: That's more hacks than I'm willing to do, let's see if this fixes the issue first

behrmann: Agreed. My thought was this would give the operation a chance to succeed. This way we just drop it, no?

DaanDeMeyer: It's best effort with this patch, which I figure is OK. We really need to start using repart and systemd-dissect as a proper fix

behrmann: systemd-dissect is already available in Debian, repart will finally be in bookworm (to be released next year)

DaanDeMeyer: But repart isn't in jammy, so we're stuck for the foreseeable future

behrmann: We could build it ourselves like we do for nspawn, couldn't we?

DaanDeMeyer: We're going to try to get it into Jammy first, and maybe try shipping it in a PPA if that doesn't work

PR #1096, opened 2022-07-28
action: Download archlinux-keyring from Michel Salim's kernel-utils PPA

DaanDeMeyer:
Pending availability in Ubuntu itself, let's download the archlinux-keyring
from the kernel-utils PPA maintained by Michel Salim. This means we can
update the Archlinux keyring without having to commit to mkosi itself
which means that users of the action will pick up the keyring updates
automatically.


DaanDeMeyer:
Still waiting for the pacman package to be uploaded to the PPA, so this doesn't work just yet

bluca (approved):

Issue #1092, opened 2022-07-27
Set squashFS minimum partition size to RootSize


mcassaniti:
When building with SquashFS (and potentially other FS types) the partition size inserted is the size of the file system blob when the partition needs to be generated (i.e.: `is_generated_root()` is true). For SquashFS this may be much smaller than the value specified for RootSize. In general this isn't an issue but it becomes a problem when we have the following layout:

  1. An EFI partition
  1. A SquashFS root or /usr partition
  1. A verity partition for the previous partition

If the SquashFS partition is ever updated (like A/B updates) and the new partition doesn't fit within the allocated space then there's an issue. `sysetmd-repart` will not assist here to resize the SquashFS partition either because that would require moving data around.

I mainly wanted feedback before I start a PR on this one. It seems adjusting the minimum size when inserting the partition based on the partition type and file system would be sufficient.

mcassaniti:
I decided to have a go anyway.

PR #1097, opened 2022-07-28
Make inserted root or /usr partition at least as big as RootSize

mcassaniti:
When using SquashFS in particular, the size of the partition inserted will always be the size of the SquashFS image. This change allows the partition size to be set which allows for future growth.

See #1092 for more details.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3637,6 +3637,9 @@ def insert_partition(
 
     luks_extra = 16 * 1024 * 1024 if state.config.encrypt == "all" else 0
     blob_size = os.stat(blob.name).st_size
+    if ident == PartitionIdentifier.root and not state.config.minimize:
+        # Make root partition at least as big as the specified size
+        blob_size = max(blob_size, state.config.root_size)
     part = state.partition_table.add(ident, blob_size + luks_extra, type_uuid, description, part_uuid)
 
     disk_size = state.partition_table.disk_size()
@@ -6429,7 +6432,9 @@ def load_args(args: argparse.Namespace) -> MkosiConfig:
 
     if args.output_format.is_squashfs():
         args.read_only = True
-        args.root_size = None
+        if args.root_size is None:
+            # Size will be automatic
+            args.minimize = True
         if args.compress is False:
             die("Cannot disable compression with squashfs", MkosiNotSupportedException)
         if args.compress is None:


mcassaniti:
@DaanDeMeyer Might as well look at this one too.

mkosi/__init__.py
```diff
@@ -6514,7 +6517,9 @@ def load_args(args: argparse.Namespace) -> MkosiArgs:
 
     if args.output_format.is_squashfs():
         args.read_only = True
-        args.root_size = None
+        if args.root_size is None:
+            # Size will be automatic
+            args.minimize = True
```
keszybz, line 6522: Hmm, this seems iffy. There can be multiple partitions, and we might not specify the root size because we're fine with whatever gets assigned automatically, but we might now want overall minimization to be enabled.

mkosi/__init__.py
```diff
@@ -3717,6 +3717,9 @@ def insert_partition(
 
     luks_extra = 16 * 1024 * 1024 if args.encrypt == "all" else 0
     blob_size = os.stat(blob.name).st_size
+    if ident == PartitionIdentifier.root and not args.minimize:
+        # Make root partition at least as big as the specified size
+        blob_size = max(blob_size, args.root_size)
```
keszybz, line 3722: Hmm, I think a bigger fix should be applied to this. Non only root but other partitions also may use squashfs. We should apply the same logic to them.

I know the docs for `--minimize` say "root file system", but I think that's a left-over. If we're build a usr-only system, we'd want to minimize it too.

I think we need to first figure out what the desired user-facing logic is…

mkosi/__init__.py
mcassaniti: I think I understand the concern here. I've written the code to overload the use of `args.minimize` and it might have other effects. In the current code when I check where it is used though, there isn't any other locations where I can see this causing issues.

Currently the code will use the minimum partition size for SquashFS. For example, if the generated SquashFS image is only 800MB then the partition size will be 800MB **always** even if the user:

  * Leaves the root size at the default (3GB)
  * Sets an explicit root size that is larger

Note that enabling `UsrOnly` means using the same root size rules as above. For non-root (and non `/usr`) partitions, the size of partitions is set explicitly so these proposed changes have no effect on them. In other words, setting a home size will ensure that the home partition is that size always.

Can you let me know if I'm missing something or what you propose is a better option?

mkosi/__init__.py
mcassaniti: If `UsrOnly` is set then this logic will apply to that partition type too. Based on the `PartitionIdentifier` Enum class we have the following partition types:

  * `root` or `usr` depending on the `UsrOnly` option. The code however treats them both as `PartitionIdentifier.root`.
  * `verity` and `verity-sig` partition types which are always automatically sized. It would be ideal to also have size controls for these too but that is another PR for another day.
  * An `esp` partition type which has a default size and an ability to set an explicit size
  * Other partition types under `PartitionIdentifier` which each have an explicit control for specifying their individual sizes

I would suggest the following logic as below.

| Partition Type | SquashFS | Minimize | Outcome | Matches Current Behaviour |
|--------------------|---------------|--------------|--------------|--------------------------------------|
| `root` or `usr` |  False        | False       | Use explicit root size with fallback to default | True |
| `root` or `usr` |  False        | True        | Use the minimized size for the partition even if the root size was set | True |
| `root` or `usr` |  True        | False        | Use explicit root size with fallback to minimized size | False |
| `root` or `usr` |  True        | True          | Use the minimized size for the partition even if the root size was set | True |
| `esp` | N/A | N/A | Use explicit size with fallback to default | True |
| `verity` and `verity-sig` | N/A | N/A | Use minimized size | True |
| All others | N/A | N/A | Use explicit size | True |

mcassaniti: @keszybz Did you get a chance to look over the feedback I provided?

mcassaniti:
Can you take a look @behrmann, @keszybz or @DaanDeMeyer ?

mcassaniti:
Still trying to get a review on this one to get it merged.

keszybz (approved):
I expect that we'll need to refactor what `--minimize` does in the future… But right now it indeed seems to only affect root partition, so this patch should do the right thing.

PR #1100, opened 2022-08-02
Add support for mkosi.conf and mkosi.conf.d configuration files/dirs

DaanDeMeyer:
"default" is a rather unintuitive file extension for a config file.
Let's prefer the more widespread "conf" file extension instead. We'll
now look for mkosi.conf and mkosi.conf.d in addition to the already
supported mkosi.default and mkosi.default.d.

We also rename the --default option to --config.

All mentions of mkosi.default, mkosi.default.d and --default in the
docs have been replaced with their conf counterpart.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -165,9 +165,9 @@ jobs:
 
     - name: Build/Boot ${{ matrix.distro }}/${{ matrix.format }}
       run: |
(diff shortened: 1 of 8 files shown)


keszybz (approved):
Yeah, that's something that we should've done a long time ago.

I went over the diff, and it looks reasonable. I guess it's likely that some spots might have been missed, but we can always fix those later.

PR #1104, opened 2022-08-04
Add compat with older nspawn that doesn't know --rlimit

DaanDeMeyer:


Edu4rdSHL (approved):

bluca (approved):

behrmann (approved):
LGTM. This blocks on systemd v239, so at least Ubuntu 18.04 is a blocker here, so in principle this could be dropped next April. It might be helpful to note in a comment or the commit which distros make this necessary.

keszybz:
LGTM too, but I'll leave it open for the commit text update.

Issue #1105, opened 2022-08-05
Recording deb manifest fails on Ubuntu 18.04


DaanDeMeyer:
```
‣  Recording packages in manifest…
‣  (Unmounting image)
‣  (Detaching /dev/loop36)
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/rphibel/src/mkosi/mkosi/__main__.py", line 36, in <module>
    main()
  File "/home/rphibel/src/mkosi/mkosi/__main__.py", line 27, in main
    run_verb(a)
  File "/home/rphibel/src/mkosi/mkosi/__init__.py", line 8138, in run_verb
    manifest = build_stuff(args)
  File "/home/rphibel/src/mkosi/mkosi/__init__.py", line 7490, in build_stuff
    image = build_image(args, root, manifest=manifest, do_run_build_script=False, cleanup=True)
  File "/home/rphibel/src/mkosi/mkosi/__init__.py", line 7239, in build_image
    manifest.record_packages(root)
  File "/home/rphibel/src/mkosi/mkosi/manifest.py", line 83, in record_packages
    self.record_deb_packages(root)
  File "/home/rphibel/src/mkosi/mkosi/manifest.py", line 152, in record_deb_packages
    installtime = datetime.fromtimestamp(int(installtime))
ValueError: invalid literal for int() with base 10: ''
```

DaanDeMeyer:
cc @bluca

bluca:
interesting, I'll have a look when I got a moment

PR #1106, opened 2022-08-05
Fix manifest creation when running on Bionic

bluca:
Very old dpkg doesn't have the db-fsys:Last-Modified field
so it errors out. Skip creating manifests for sysext when
building on such old distros. Only affects Bionic right now.

Fixes https://github.com/systemd/mkosi/issues/1105


keszybz (approved):
Yeah, setting it to 0 should be good enough as a work-around.

Issue #1107, opened 2022-08-06
Signing systemd-boot EFI binaries under /usr


mcassaniti:
All files ending with `.efi` or `.EFI` under the ESP partition get signed for secure boot if using the secure boot option. This is indeed a good thing. If something like `bootctl add` is run then the file `/usr/lib/systemd/boot/efi/systemd-bootx64.efi` can be copied back into the ESP. This copy is unsigned however and will break secure boot.

Note that copying sytemd-boot to the ESP can sometimes happen without an explicit user request, and even if the user explicitly does request it that should not cause secure boot to fail.

My suggestion would be to sign EFI files under `/usr/lib/systemd/boot/efi` before the contents of `root` or `/usr` become read-only. Currently files under the ESP partition are signed after the point where no more changes to `root` or `/usr` should occur. I'd like to collect feedback before submitting a PR.

poettering:
Note that this is not as easy as it sounds. In Verity mode we want to encode the root hash in the kernel command line included in the unified kernel image. And since the Verity root hash covers the /usr tree, you thus cannot embed the kernel in its final form in /usr/, as you have a cycle problem then...

mcassaniti:
@DaanDeMeyer Should I take that as a sign to start on a PR?

behrmann:
That's just just a label for sorting issues, but just start a PR and then we'll see what comes out. It's always easier to talk about concrete stuff. :) 

To topic: If you're only after signing `systemd-boot`. That sounds pretty self-contained and doesn't have the the cycle problem.

poettering:
> @poettering You are correct about the cyclic problem, but (unless I'm wrong) the file I'm looking to make sure is signed is not the kernel EFI image. It should be the `systemd-boot` loader that will load such a kernel image.

Oh, you are right. Ignore me, signing sd-boot in /usr/ should be OK.

poettering:
> I'm not sure if this is something mkosi should do. If I understand correctly, `bootctl` should take care of signing the binaries from /usr when `bootctl add` is executed

Hmm, given sbsign and so on exist already, is bootctl really the right place?

Givning the signing keys for the secureboot stuff area actually maintained by mkosi right now, it probably does make sense that mkosi also signs the sd-boot EFI binary with the same key it would sign the unified kernels with.

mcassaniti:
> I'm not sure if this is something mkosi should do. If I understand correctly, `bootctl` should take care of signing the binaries from /usr when `bootctl add` is executed

[...]

PR #1109, opened 2022-08-08
Sign systemd-boot EFI binaries under /usr

mcassaniti:
If secure boot signing is enabled with UEFI then `systemd-boot` binaries which are at `/usr/lib/systemd/boot/efi` should be signed with the same signing key that is used for EFI binaries within the ESP. In comparison to signed ESP binaries, the '.signed' extension will be kept.

Without this change, any tool that copies these binaries into the ESP will (likely) be copying an unsigned copy of `systemd-boot`. The service `systemd-boot-update.service` is just one example of a service that will regularly attempt to update `systemd-boot` within the ESP.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -4041,9 +4041,11 @@ def install_unified_kernel(
 def secure_boot_sign(
     config: MkosiConfig,
     state: MkosiState,
+    directory: Path,
     for_cache: bool,
     cached: bool,
     mount: Callable[[], ContextManager[None]],
+    replace: bool = False,
 ) -> None:
     if state.do_run_build_script:
         return
@@ -4057,28 +4059,27 @@ def secure_boot_sign(
         return
 
     with mount():
-        for path, _, filenames in os.walk(state.root / "efi"):
-            for i in filenames:
-                if not i.endswith(".efi") and not i.endswith(".EFI"):
-                    continue
+        for f in itertools.chain(directory.glob('*.efi'), directory.glob('*.EFI')):
+            if os.path.exists(f"{f}.signed"):
+                MkosiPrinter.info(f"Not overwriting existing signed EFI binary {f}.signed")
+                continue
 
-                with complete_step(f"Signing EFI binary {i} in ESP…"):
-                    p = os.path.join(path, i)
+            with complete_step(f"Signing EFI binary {f}…"):
+                run(
+                    [
+                        "sbsign",
+                        "--key",
+                        config.secure_boot_key,
+                        "--cert",
+                        config.secure_boot_certificate,
+                        "--output",
+                        f"{f}.signed",
+                        f,
+                    ],
+                )
 
-                    run(
-                        [
-                            "sbsign",
-                            "--key",
-                            config.secure_boot_key,
-                            "--cert",
-                            config.secure_boot_certificate,
-                            "--output",
-                            p + ".signed",
-                            p,
-                        ],
-                    )
-
-                    os.rename(p + ".signed", p)
+                if replace:
+                    os.rename(f"{f}.signed", f)
 
 
 def extract_unified_kernel(
@@ -7215,6 +7216,9 @@ def build_image(
                 sshkey = setup_ssh(config, state, for_cache, cached_tree)
                 setup_netdev(config, state, cached_tree)
                 run_postinst_script(config, state, loopdev, for_cache)
+                # Sign systemd-boot / sd-boot EFI binaries
+                secure_boot_sign(config, state, state.root / 'usr/lib/systemd/boot/efi',
+                    for_cache, cached, mount=contextlib.nullcontext)
 
                 if cleanup:
                     remove_packages(config, state.root)
@@ -7267,7 +7271,9 @@ def build_image(
                                         root_read_only=True)
 
             install_unified_kernel(config, state, root_hash, for_cache, cached, mount)
-            secure_boot_sign(config, state, for_cache, cached, mount)
+            # Sign EFI binaries under these directories within the ESP
+            for esp_dir in ['efi/EFI/BOOT', 'efi/EFI/systemd', 'efi/EFI/Linux']:
+                secure_boot_sign(config, state, state.root / esp_dir, for_cache, cached, mount, replace=True)
             split_kernel = (
                 extract_unified_kernel(config, state, for_cache, mount)
                 if config.split_artifacts


mcassaniti:
Fixes #1107.

mkosi/__init__.py
```diff
@@ -4113,6 +4113,7 @@ def secure_boot_sign(
     for_cache: bool,
     cached: bool,
     mount: Callable[[], ContextManager[None]],
+    esp: bool,
```
behrmann, line 4116: I think a more general name like `replace` or `overwrite` might be more suitable.

mkosi/__init__.py
```diff
@@ -4126,14 +4127,18 @@ def secure_boot_sign(
         return
 
     with mount():
-        for path, _, filenames in os.walk(root / "efi"):
+        base_path = root / "efi" if esp else root / "usr/lib/systemd/boot/efi"
+        for path, _, filenames in os.walk(base_path):
```
behrmann, line 4131: Since `root` comes in as a `Path` nowadays, this could be a `.glob("*.efi")` (which unfortunately would need to be chained with `.glob("*.EFI")`) .

mkosi/__init__.py
```diff
@@ -4126,14 +4127,18 @@ def secure_boot_sign(
...
             for i in filenames:
+                p = os.path.join(path, i)
                 if not i.endswith(".efi") and not i.endswith(".EFI"):
                     continue
+                if os.path.exists(f"{p}.signed"):
+                    # Do not overwrite existing signed EFI binaries
```
behrmann, line 4137: Writing a warning to the user here might be nice, since then there's a clear answer to "why didn't it sign X with my keys?"

mkosi/__init__.py
```diff
@@ -4126,14 +4127,18 @@ def secure_boot_sign(
         return
 
     with mount():
-        for path, _, filenames in os.walk(root / "efi"):
+        base_path = root / "efi" if esp else root / "usr/lib/systemd/boot/efi"
```
behrmann, line 4130: Should we maybe always the stuff in `/usr/lib/systemd/boot/efi`? In that case it would maybe also make sense to lift which path is being searched for things to sign out of this function and make this path an argument (in lieu of `root`).

mkosi/__init__.py
DaanDeMeyer: I wouldn't want, just an informational statement is sufficient here since it's not actually a problem if distros ship presigned binaries

mkosi/__init__.py
mcassaniti: I can't say I fully understand. Can you give me a short code snippet as an example?

mkosi/__init__.py
mcassaniti: Fixed.

mkosi/__init__.py
mcassaniti: Fixed.

mkosi/__init__.py
mcassaniti: Fixed.

mkosi/__init__.py
behrmann:
```diff
-       for path, _, filenames in os.walk(base_path):
+       for path in itertools.chain(base_path.glob("*.efi"), base_path.glob("*.EFI")):
```

mkosi/__init__.py
```diff
@@ -7260,6 +7260,9 @@ def build_image(
                 sshkey = setup_ssh(args, root, do_run_build_script, for_cache, cached_tree)
                 setup_netdev(args, root, do_run_build_script, cached_tree)
                 run_postinst_script(args, root, loopdev, do_run_build_script, for_cache)
+                # Sign systemd-boot / sd-boot EFI binaries
+                secure_boot_sign(args, root + 'usr/lib/systemd/boot/efi', do_run_build_script,
```
behrmann, line 7264:
```suggestion
                secure_boot_sign(args, root / 'usr/lib/systemd/boot/efi', do_run_build_script,
```
and below.

mcassaniti: Whoops. Fixed.

behrmann (approved):

mcassaniti:
Any EFI binaries within the ESP should be signed including anything in a sub-directory of `/efi`. This was there to begin with but I missed it along the way. It seems that `Path.glob()` doesn't cover sub-directories. I've tested the changes already to confirm the extracted unified kernel which lives in the ESP was signed.

mkosi/__init__.py
```diff
@@ -4126,14 +4128,18 @@ def secure_boot_sign(
...
-                    continue
+        if recursive:
+            paths = [Path(item[0]) for item in os.walk(efi_path)]
+        else:
+            paths = [efi_path]
 
-                with complete_step(f"Signing EFI binary {i} in ESP…"):
-                    p = os.path.join(path, i)
+        for path in paths:
+            for f in itertools.chain(path.glob('*.efi'), path.glob('*.EFI')):
```
behrmann, line 4137: You can do recursive globs with what in bash parlance is called `globstar`, i.e. `**/*.efi`.

behrmann (changes requested):

behrmann:
@mcassaniti good catch, that this is recursive!

mkosi/__init__.py
mcassaniti: Should this function always operate recursively or should it still be optional? Based on that I'll update the code.

behrmann: Since the original is always recursive, this should be, too.

mkosi/__init__.py
```diff
@@ -4126,28 +4127,27 @@ def secure_boot_sign(
         return
 
     with mount():
-        for path, _, filenames in os.walk(root / "efi"):
-            for i in filenames:
-                if not i.endswith(".efi") and not i.endswith(".EFI"):
-                    continue
+        for f in itertools.chain(efi_path.glob('**/*.efi'), efi_path.glob('**/*.EFI')):
+            if os.path.exists(f"{f}.signed"):
```
behrmann, line 4131:
```suggestion
            if f.with_suffix(".efi.signed").exists()
```

mkosi/__init__.py
```diff
@@ -4126,28 +4127,27 @@ def secure_boot_sign(
...
+                        args.secure_boot_certificate,
+                        "--output",
+                        f"{f}.signed",
+                        f,
+                    ],
+                )
 
-                    os.rename(p + ".signed", p)
+                if replace:
+                    os.rename(f"{f}.signed", f)
```
behrmann, line 4150:
```suggestion
                    f.rename(f.with_suffix(".efi.signed"))
```

mkosi/__init__.py
mcassaniti: If you insist I'll change it, but this needs to be `if f.with_suffix(".efi.signed").exists() or if f.with_suffix(".EFI.signed").exists()`. I'm not seeing the benefit beyond casting a path to a string.

mkosi/__init__.py
mcassaniti: Does this need to preserve the case (efi vs EFI)? See previous comment if it should.

mkosi/__init__.py
behrmann: Nevermind then, I forgot about the blasted case insensitivity of FAT32 again. `Path`s really need an `add_suffix` method.

mkosi/__init__.py
behrmann: I don't think it does, but see above.

behrmann (approved):

mcassaniti:
Thanks for all the hard efforts reviewing these.

mcassaniti:
@DaanDeMeyer Would you mind also checking and (hopefully) merging?

mkosi/__init__.py
```diff
@@ -4108,11 +4108,12 @@ def install_unified_kernel(
 
 def secure_boot_sign(
     args: MkosiArgs,
-    root: Path,
+    efi_path: Path,
```
DaanDeMeyer, line 4111: Just name this `directory`

mkosi/__init__.py
```diff
@@ -4126,28 +4127,27 @@ def secure_boot_sign(
         return
 
     with mount():
-        for path, _, filenames in os.walk(root / "efi"):
-            for i in filenames:
-                if not i.endswith(".efi") and not i.endswith(".EFI"):
-                    continue
+        for f in itertools.chain(efi_path.glob('**/*.efi'), efi_path.glob('**/*.EFI')):
```
DaanDeMeyer, line 4130: I'd prefer if we do not glob recursive everything here. I'd change this to just glob "*.efi" and call secure_boot_sign() multiple times if you need to sign in multiple directories

DaanDeMeyer (changes requested):

mkosi/__init__.py

DaanDeMeyer (approved):

DaanDeMeyer:
Needs a rebase

DaanDeMeyer:
Sorry for the churn, but we need another rebase, (we have a few large refactorings going on)

DaanDeMeyer:
> If secure boot signing is enabled with UEFI then systemd-boot binaries which are at /usr/lib/systemd/boot/efi should be signed with the same signing key that is used for EFI binaries within the ESP. In comparison to signed ESP binaries, the '.signed' extension will be kept.

This doesn't really say why the binaries in /usr need to be signed as well.. Can you clarify why the binaries need to be signed in /usr? It should be sufficient if they're signed in the ESP no? It'd be great if you could extend the commit message with a more detailed explanation

mcassaniti:
> > If secure boot signing is enabled with UEFI then systemd-boot binaries which are at /usr/lib/systemd/boot/efi should be signed with the same signing key that is used for EFI binaries within the ESP. In comparison to signed ESP binaries, the '.signed' extension will be kept.
> 
> This doesn't really say why the binaries in /usr need to be signed as well.. Can you clarify why the binaries need to be signed in /usr? It should be sufficient if they're signed in the ESP no? It'd be great if you could extend the commit message with a more detailed explanation

I've updated the commit message and the PR description to cover this. It (hopefully) is also covered in #1107 which prompted this PR. I'm happy to provide something more detailed if required.

DaanDeMeyer (approved):

bluca:
What happens if the binaries are already signed? Normally they would be, as they are produced by a different package and just used as-is

mcassaniti:
> What happens if the binaries are already signed? Normally they would be, as they are produced by a different package and just used as-is

Good question. We get an additional signature on the EFI binary.

```
mkosi-test$ sbsign --key mkosi.secure-boot.key --cert mkosi.secure-boot.crt --output systemd-bootx64.efi.double-signed systemd-bootx64.efi.signed 
warning: data remaining[108576 vs 123400]: gaps between PE/COFF sections?
Image was already signed; adding additional signature
mkosi-test$ echo $?
0
mkosi-test$ sbverify --list systemd-bootx64.efi.double-signed 
warning: data remaining[110136 vs 124960]: gaps between PE/COFF sections?
signature 1
image signature issuers:
 - /CN=mkosi of mcassaniti
image signature certificates:
 - subject: /CN=mkosi of mcassaniti
   issuer:  /CN=mkosi of mcassaniti
signature 2
image signature issuers:
 - /CN=mkosi of michael
image signature certificates:
 - subject: /CN=mkosi of michael
   issuer:  /CN=mkosi of michael
```

DaanDeMeyer:
Can we check for that and skip signing if the image is already signed?

mcassaniti:
We could, but what is the harm in having multiple signatures? If we're relying on using the hash of the EFI binaries then it's highly likely the user will already need to enrol those. If instead it's just signature validation then either signature will validate.

If  an upstream distributor signed the EFI binary for `systemd-boot` and did **not** leave it with a `.signed` extension then `mkosi` will add a signature alongisde the existing signature. Should the user choose not to trust the certificate chain for the upstream distributor and only enrol the certificate chain `mkosi` is using then everything will still work fine.

There's already logic to skip over EFI binaries ending with `.signed` in the directories that are searched. A vendor **should** make sure their signed `systemd-boot` EFI binary ends with `.signed`.

PR #1110, opened 2022-08-08
Support ListAction with type=Path

DaanDeMeyer:
We don't yet properly process multiple values of type Path such
as in --extra-tree. This commit updates ListAction to support such
cases.


DaanDeMeyer:
Merging early since this isn't covered by the integration tests

PR #1111, opened 2022-08-09
Always use a distro~release subdirectory for output, cache, builddir

DaanDeMeyer:
Let's be consistent and always use a distro~release subdirectory for
these directories, instead of only using these when the correct directory
exists in the directory mkosi is invoked from.


behrmann (approved):
Makes sense. Should be mentioned in the manpage and the news, though.

DaanDeMeyer:
Added a blurb to news, let's keep it out of the manpage though I'd say, seems a bit too detailed to mention there, more of an implementation detail.

behrmann:
Agreed for the cache and builddir, but on the output directory it is userfacing. I mean, it's not like this will confuse users, but it is something they'll see.

DaanDeMeyer:
Only docs updates, so let's not wait for CI

PR #1114, opened 2022-08-11
manifest: Add --manifest to toggling manifest generation

Foxboron:
Some Linux distros might not have the tools available to generate the
manifests. This flag allows us to turn it off in cases where it's
needed.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -5119,6 +5119,7 @@ def create_parser() -> ArgumentParserMkosi:
         type=cast(Callable[[str], ManifestFormat], ManifestFormat.parse_list),
         help="Manifest Format",
     )
+    group.add_argument('--manifest', default=True, action=argparse.BooleanOptionalAction),
     group.add_argument(
         "-o", "--output",
         help="Output image path",
@@ -7445,7 +7446,9 @@ def build_stuff(args: MkosiArgs) -> Manifest:
     workspace = setup_workspace(args)
 
     image = BuildOutput.empty()
-    manifest = Manifest(args)
+    manifest = None
+    if args.manifest:
+        manifest = Manifest(args)
 
     # Make sure tmpfiles' aging doesn't interfere with our workspace
     # while we are working on it.
@@ -8137,7 +8140,8 @@ def run_verb(raw: argparse.Namespace) -> None:
         if args.auto_bump:
             bump_image_version(args)
 
-        save_manifest(args, manifest)
+        if args.manifest:
+            save_manifest(args, manifest)
 
         print_output_size(args)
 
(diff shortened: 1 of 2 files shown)


Foxboron:
>Is the use-case for this rpm or apt-get being absent in the image (which I guess might happen if people remove them in the postinst script)?

The tools are run on the host system. So `dpkg-query` needs to be on my Arch machine for this to work.

>Anyway, I think totally disabling the manifest output should also be possible via setting --manifest-format= explicitly empty. (because the fallback value of json output is triggered for the passed argument being None instead of an empty string). This is admittedly not the nicest UX, but I think adding a line to the docs might be easier than another switch (of which we have quite a lot already).

Ooh, that does make things easier. But admittedly a bit hard to figure out :)

Foxboron:
Actually, this doesn't work. The code forces a format on us if the manifest-format flag is set empty. So there should be some way to disable the manifest generation if needed.

https://github.com/systemd/mkosi/blob/main/mkosi/__init__.py#L6488-L6489

behrmann:
> The tools are run on the host system. So `dpkg-query` needs to be on my Arch machine for this to work.

Ah, right, that makes sense. Although, `dpkg` is in `community`. ;)
 
> Actually, this doesn't work. The code forces a format on us if the manifest-format flag is set empty. So there should be some way to disable the manifest generation if needed.

You are right, although the reason is slightly different. The mentioned lines are only exercised when `--manifest-format=` is not passed at all, since then argparse will fall back to `None`. On an empty value this is parsed to an empty list, because of the `type=` option in argparse there. 

The problem is, that `Manifest.record_packages` tries to do its work because it looks at what the target distribution is and not the host and `record_deb_packages` and `record_rpm_packages` use `run` for the initial call to `rpm` and `dpkg-query`.

I'm not sure whether either could be missing in the image under normal circumstances. You could give it a try change the `dpkg-query` call to use `run_workspace_command` instead of `run` there.

DaanDeMeyer:
We do try to run common tools on the host machine, because it makes building for different architectures a lot easier and faster, especially if the operations are heavy.

Foxboron:
@DaanDeMeyer But the calling of `rpm` and `dpkg-query` should be fine to run in the container? no?

DaanDeMeyer:
I'd keep running `rpm` out of the host system, it's required to be installed anyway to build rpm based images. dpkg-query can be run inside the container because apt has to be executed inside the container as well because it doesn't support a proper --root option. We only need debootstrap for debian/ubuntu based images and we want to keep it that way so moving dpkg-query inside the container makes sense.

Foxboron:
Superceded by https://github.com/systemd/mkosi/pull/1122

We probably don't need '--no-manifest' or an equivalent with this.

Issue #687, opened 2021-03-25
Optionally, right before completing set all mtimes to zero


poettering:
Our images are unfortunately currently mostly not reproducible, since the package manager has a lot of freedom of writing stuff to disk, and so have the file systems used as backend. But we can at least slightly improve things, so that the file manifests at least become reproducible a bit more: let's optionally reset the mtimes of everything in the root partition (not in the var/home/srv/… partitions) to zero right before completing things.

DaanDeMeyer:
If we ever get things fully reproducible we maybe do double builds in CI to make sure it stays that way. I'm kinda sceptical we'll get there soon though.

Foxboron:
Is there a list of directories which should not be zeroed?

poettering:
> Is there a list of directories which should not be zeroed?

A simple approach might be to just zero the mtime of everything outside of /proc and /sys, and then wait for complaints where that broke anything.

malt3:
Me and @katexochen would like to take a stab at this.
How would you handle this with regards to `SOURCE_DATE_EPOCH`?

Possible implementations:

- Just use `Environment=SOURCE_DATE_EPOCH=0`. This works well with other tools and doesn't need any more flags / parameters.
- Have `NormalizeMtimes=0` as a setting and allow arbitrary timestamps to be chosen).
- Have `ZeroMtimes` and some sensible fallback for cases where `SOURCE_DATE_EPOCH` is set.

@behrmann @DaanDeMeyer

DaanDeMeyer:
See the discussion in https://github.com/systemd/mkosi/pull/1695

malt3:
To inline this, and ensure this goes into the right direction:

> If $SOURCE_DATE_EPOCH is set, always use it.
If $SOURCE_DATE_EPOCH is not set and --reproducible is set, set mtimes to zero
If $SOURCE_DATE_EPOCH is not set and --reproducible is not set, don't touch mtimes

Adding a `--reproducible` flag and settings `SOURCE_DATE_EPOCH` if wasn't set by the user can be follow ups to that.

DaanDeMeyer:
Sounds good, for now, let's just use `SOURCE_DATE_EPOCH` whenever it is set.

keszybz:
> Adding a `--reproducible` flag and setting `SOURCE_DATE_EPOCH` if it wasn't set by the user can be follow ups to the initial PR for normalizing mtimes.

In the rpm world, we have the following:
[...]

PR #1115, opened 2022-08-11
mkosi: Add --reproduce to produce reproducible images

Foxboron:
- [ ] Should we support `SOURCE_DATE_EPOCH`?
- [ ] Missing documentation
- [ ] Missing scripts for other essential distributions
- [ ] Needs a list of directories we should not null the mtime of


Foxboron:
So this is pretty much a first step. This can create two idential images if we build them twice in a short timeframe.

The next goal would be to take the `manifest.json` and recreate the OS images. I have a couple of ideas here already so I think it might be doable atleast for Debian and Arch Linux. I'm more uncertain of Fedora/RPM as I'm less familiar with the ecosystem, but I'm sure we can figure out something.

mkosi/__init__.py
```diff
@@ -3260,6 +3260,26 @@ def run_finalize_script(args: MkosiArgs, root: Path, do_run_build_script: bool,
...
+
+    with complete_step("Attempting to make build reproducible…"):
+        env = collections.ChainMap(dict(BUILDROOT=str(root), OUTPUTDIR=str(output_dir(args))),
+                                   args.environment,
+                                   os.environ)
+
+        file = f"reproducible.{args.distribution}.sh"
+        if importlib.resources.is_resource("mkosi.resources", file):
+            with importlib.resources.path("mkosi.resources", file) as p:
+                run(["bash", p, verb], env=env)
```
behrmann, line 3274: More in line with the rest of the functions would be to match `args.distribution` and call some `make_reproducible_$DISTRO`. This could be something like
```python
def make_reproducible_debian(args, root):
    for p in ("var/log/dpkg.log", "var/log/bootstrap.log", "var/cache/apt/pkgcache.bin", "var/log/apt/history.log", "var/log/apt/term.log", "var/log/alternatives.log", "var/cache/ldconfig/aux-cache", "var/log/apt/eipp.log.xz", "var/lib/dbus/machine-id"):
        (root / p).unlink(missing_ok=True)
```
(with the caveat, that we are still on Python 3.7 so `missing_ok=` doesn't yet exist, so a try-except block would be needed).

In the case of keeping the extra script: `run` passes keyword arguments it doesn't use through, so instead calling bash yourself you could pass in `shell=True`.

mkosi/__init__.py
```diff
@@ -3260,6 +3260,26 @@ def run_finalize_script(args: MkosiArgs, root: Path, do_run_build_script: bool,
...
+        else:
+            warn("Reproducible images is not support on {arge.distribution}")
+
+    # TODO: Support SOURCE_DATE_EPOCH
+    reset_timestamps = ["find", root, "-mindepth", "1", "-execdir", "touch", "-hcd", "@0", "{}", "+"]
```
behrmann, line 3279:
```suggestion
    reset_timestamps = ["find", root, "-mindepth", "1", "-execdir", "touch", "--no-dereference", "--no-create", "--date=@0", "{}", "+"]
```
should be more readable long term.

mkosi/__init__.py
```diff
@@ -5337,6 +5357,13 @@ def create_parser() -> ArgumentParserMkosi:
         action=BooleanAction,
         help="Generate split out root/verity/kernel images, too",
     )
+    group.add_argument(
+        "--reproducible",
+        metavar="BOOL",
+        action=BooleanAction,
+        default=False,
```
behrmann, line 5364: Just to open that discussion: Is there an argument why the default should *not* be reproducible?

mkosi/resources/reproducible.debian.sh
```diff
@@ -0,0 +1,11 @@
...
+
+rm -f ${BUILDROOT}/var/log/dpkg.log
+rm -f ${BUILDROOT}/var/log/bootstrap.log
+rm -f ${BUILDROOT}/var/cache/apt/pkgcache.bin
+rm -f ${BUILDROOT}/var/log/apt/history.log
+rm -f ${BUILDROOT}/var/log/apt/term.log
+rm -f ${BUILDROOT}/var/log/alternatives.log
+rm -f ${BUILDROOT}/var/cache/ldconfig/aux-cache
+rm -f ${BUILDROOT}/var/log/apt/eipp.log.xz
+rm -f ${BUILDROOT}/var/lib/dbus/machine-id
```
behrmann, line 11: `/var/lib/dbus/machine-id` is handled in `reset_machine_id` already/too so I think this might need some care.

mkosi/resources/reproducible.debian.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash -x
```
behrmann, line 1: I've preferred this style a long time as well, but I think this has the downside of potentially spamming output with stuff from `/etc/profile.d`.
```suggestion
#!/bin/bash
set -x
```

mkosi/__init__.py
Foxboron: I don't think we want by default zero out the mtime. The alternative would be to define `SOURCE_DATE_EPOCH` to the current build time, and set all files to this time. Then include the timestamp in the manifest.

mkosi/__init__.py
Foxboron: We could do something like that. But currently I think the bash stuff is easier to dry-run and experiment with. But I might be alone with that opinion :)

mkosi/__init__.py
behrmann: I think using `SOURCE_DATE_EPOCH` and
1. saving it to the manifest, and
2. passing it through to build/postinst/finalise script 

would make sense, but let's see what other's think about it.

Foxboron: Hmm, then we need to think a little bit about the UX here.

* Should we demand a `manifest.json` when we use `--reproducible`? This is a chicken and egg problem.
* Error or warning if we pass `--reproducible` without setting `SOURCE_DATE_EPOCH` in the env?
  * `--reproducible` might not actually make a reproducible image by default. This could be confusing.
* Zeroing `mtime` could be a separate option.

We might need a new option section for reproducability then as well.

mkosi/resources/reproducible.debian.sh
Foxboron: I was suppose to remove `-x`. But there is no output from this (strangely enough) so I forgot about it.

Foxboron:
For reproducing fedora installs I have submitted an issue at authselect as the files are not deterministic.

https://github.com/authselect/authselect/issues/319

DaanDeMeyer:
If we do SOURCE_DATE_EPOCH, we should also set it whenever invoking a postinst script, build script, prepare script, ... . Especially for the build script it's important since build systems pick it up automatically.

mkosi/__init__.py
```diff
@@ -1489,7 +1489,9 @@ def mount(
         run(cmd)
         yield where
     finally:
+        timestamp = os.path.getmtime(where)
         run(["umount", "--no-mtab", "--recursive", where])
+        os.utime(where, (timestamp, timestamp))
```
poettering, line 1494: what's the rationale here?

mkosi/__init__.py
poettering: Can we do this without shelling out? sounds simple enough to do that within python, even if it might be a tad bit slower

poettering:
Hmm, regarding the handling of those log files in /var/, shouldn't we just say people should build their images without /var/? I mean, we support that already...

I'd rather avoid having to maintain file lists, and tmpfiles.d/ stuff in the distros really should be able to set up /var/ anyway already if it is empty.

mkosi/__init__.py
Foxboron: This is because of the find hack previously.

We mount `tmpfs` mounts on a few mountpoints in the Arch installation, and the mtime/atime modifications done previously does doesn't propegate.

mkosi/__init__.py
Foxboron: Sure, can give that a try.

poettering:
Or to say this differently: if there's the choice between:

1. Either, empty /var/ entirely, and then possibly having to ship a couple of tmpfiles.d/ snippets with mkosi (thought they really should end up in the distros instead, sooner or later)

2. Or, maintain list of specific files to remove in /var/ as part of the build process.

Then I am strongly in favour of the first option.

mkosi/__init__.py
poettering: But I think if reproducible stuff is requested we probably should create/update the mount point inodes with the right mtime already *before* mounting stuff over.

Foxboron:
We could also issue a warning if we find `/var` mounted when using the `--reproducible` switch?

The issue here (i think) is balancing this so we don't end up having reproducible images becomming a long convoluted checklist.

I need to think a bit on which option is better. I'm not completley sure how the tmpfiles.d solution would look like I guess?

poettering:
btw, I think it would make sense to do the mtime-to-zero stuff via an option independent of reproducibility, it sounds good even just to save some disk space, by keeping deltas smaller.

i.e. a fix for #687 definitely makes sense as part of the goal to deliver reproducible builds, but also independently of that.

mkosi/__init__.py
Foxboron: I can try see if we can fit it in a better location while writing this into python.

Foxboron:
I can carve it out of this PR? We could also say that `--zero-mtime` is implied with `--reproducible`

poettering:
> I need to think a bit on which option is better. I'm not completley sure how the tmpfiles.d solution would look like I guess?

Well, I am trying to push for an "hermetic /usr/" OS. i.e. where the OS resources are monopolized and hermetic in /usr/, and if you then combine such a /usr/ with an otherwise empty root dir, then the snippets in /usr/lib/tmpfiles.d/ are enough to set up just enough in /etc + /var and elsewhere to get a working system.

At this point, the concept of "hermetic /usr/" has not been adopted officially by distros yet (though, surprisingly, it works quite well nonetheless, at least on Fedora), i.e packages tend not to ship enough rules in /usr/lib/tmpfiles.d/ to be able to reproduce a working setup in /etc/ and /var/ to just work. But I hope we can sooner or later convince distros to just ship that in their .deb/.rpm. And in the meantime we could just ship some subset of what's necessary with mkosi.

Distros have adopted tmpfiles.d/ to set up the /run/ hierarchies well enough to just work. Doing the same for /etc/ + /var/ should be a smaller step actually, given the acceptance of tmpfiles.d/ as a concept already happened now.

poettering:
> I can carve it out of this PR? We could also say that `--zero-mtime` is implied with `--reproducible`

Yes, that would be ideal.

poettering:
> At this point, the concept of "hermetic /usr/" has not been adopted officially by distros yet (though, surprisingly, it works quite well nonetheless, at least on Fedora), i.e packages tend not to ship enough rules yet in /usr/lib/tmpfiles.d/ to be able to reproduce a working setup in /etc/ and /var/ to just work. But I hope we can sooner or later convince distros to just ship that in their .deb/.rpm. And in the meantime we could just ship some subset of what's necessary with mkosi.

To extend on that: I think the rules necessary to boot up with an empty /var/ are much fewer and simpler than those you'll need to boot up with an empty /etc/. So I'd always focus on that first. And it should be 95% of what you need for reproducible images, I guess.

mkosi/__init__.py
keszybz: I'm in agreement with @behrmann. In fact, I came here to make pretty much the exactly same comment. Making things reproducible by default is IMHO nicer and simpler. I.e. there should be no config option. If `SOURCE_DATE_EPOCH` is set, we use it. If not, we set it to current time. In both cases, we save the value to the manifest, so that the build can be reproduced later. Also, I think the variable should be passed to all scripts by default. I don't see any downside to doing that. They might use it or ignore it, but that's no worse than _status quo_.

mkosi/__init__.py
keszybz: I think the whole approach of resetting the time is somewhat iffy. Why would we need to reset the time? For stuff that is installed from packages, I think we'd want to keep the installed timestamps. OTOH, it is the build tools that should honour the variable. In particular, I'm surprised that `meson` doesn't do this natively. E.g. wouldn't it be better if `meson install` touched the time on all files it installs? I would open a pull request against meson to add support for this.

I see though how a blanket reset of the time can be useful. But maybe this should be opt-in, and not actually advertised as "reproducible builds", but just as a low-level feature: `ResetFileTimes=` that'd accept a set of path globs, similarly to `RemoveFiles=` and other settings. This would be useful as a work-around for tools that don't support `SOURCE_DATE_EPOCH`.

mkosi/__init__.py
```diff
@@ -5337,6 +5357,13 @@ def create_parser() -> ArgumentParserMkosi:
         action=BooleanAction,
         help="Generate split out root/verity/kernel images, too",
     )
+    group.add_argument(
+        "--reproducible",
+        metavar="BOOL",
+        action=BooleanAction,
+        default=False,
+        help="Attempt to generate reproducible images"
```
keszybz, line 5367: Comma  missing.

mkosi/__init__.py
```diff
@@ -2042,7 +2042,7 @@ def invoke_dnf(
     cmdline += [command, *sort_packages(packages)]
 
     with mount_api_vfs(args, root):
-        run(cmdline, env=dict(KERNEL_INSTALL_BYPASS="1"))
+        run(cmdline, env=dict(SOURCE_DATE_EPOCH="0", KERNEL_INSTALL_BYPASS="1"))
```
keszybz, line 2045: I don't understand. Why would we want to pass "0" ?

keszybz (changes requested):

mkosi/__init__.py
Foxboron: Patching this into `meson` doesn't solve a lot considering most projects doesn't even use it. We do need to touch the time since we are aiming for reproducible builds and the timestamps are part of the archive. Doing this in one swoop when people do request a repoducible builds is a small thing imo and doesn't need a lot more.

Having this on the side, as well as including it in the `--reproduce` switch, is fine imo. But not necessarily something that belongs in this PR.

mkosi/__init__.py
Foxboron: It's a draft, this can be changed to CURRENT_TIME or the manifest timestamp. `0` is just a simple default to get things working.

Foxboron:
I haven't forgottent his. Just doing some other stuff at work and haven't had time to pick up on this :)

DaanDeMeyer:
Let's close this one for now since it's extremely old and mkosi has practically been rewritten in the meantime. We can resurrect the discussion in a new PR once ready.

PR #1116, opened 2022-08-12
Set default geo mirror for Arch Linux

Foxboron:

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -6444,6 +6444,8 @@ def load_args(args: argparse.Namespace) -> MkosiArgs:
                 args.mirror = "http://ports.ubuntu.com/"
         elif args.distribution == Distribution.arch and platform.machine() == "aarch64":
             args.mirror = "http://mirror.archlinuxarm.org"
+        elif args.distribution == Distribution.arch and platform.machine() == "x86_64":
+            args.mirror = "https://geo.mirror.pkgbuild.com"
         elif args.distribution == Distribution.opensuse:
             args.mirror = "http://download.opensuse.org"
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
With this, can we drop the logic to select a random mirror from the mirrorlist in install_arch()? That doesn't really work all that well since we often get a less than perfect mirror.

Foxboron:
Oh, I missed there was logic for that. I'll rm that.

Foxboron:
@DaanDeMeyer Should be good. The only issue would be archlinux32 I *think*, but the code would not have worked there anyway.

mkosi/__init__.py
```diff
@@ -2710,19 +2710,8 @@ def install_arch(args: MkosiArgs, root: Path, do_run_build_script: bool) -> None
...
-        with urllib.request.urlopen(
-            "https://www.archlinux.org/mirrorlist/?country=all&protocol=https&ip_version=4&use_mirror_status=on"
-        ) as r:
-            mirrors = r.readlines()
-            uncommented = [line.decode("utf-8")[1:] for line in mirrors]
-            with mirrorlist.open("w") as f:
-                f.writelines(uncommented)
-            server = f"Include = {mirrorlist}"
+        # This should not happen, but for good measure.
+        raise ValueError("No repository mirror has been selected.")
```
DaanDeMeyer, line 2714: We use the `die()` function for this.

DaanDeMeyer (changes requested):

DaanDeMeyer (approved):

behrmann:
Merging, since all the Arch tests have passed and this doesn't touch anything else.

PR #1117, opened 2022-08-15
Don't load passwords unless a build is needed

jre21:
This change allows the user to set password files to be readable only by root for added security.  Additionally, if encryption is requested and no passphrase file exists, the user will not be prompted to enter a passphrase unless the mkosi invocation results in an image build.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -6122,7 +6122,7 @@ def require_private_file(name: str, description: str) -> None:
 
 
 def find_passphrase(args: argparse.Namespace) -> None:
-    if args.encrypt is None:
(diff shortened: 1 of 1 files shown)


behrmann:
Seems sensible to me.

jre21:
@DaanDeMeyer Could you take a look at this?  I'm somewhat confused by the integration test failures.  They seem to be the same failure mode as recent main branch bulids, but I haven't been able to reproduce that behavior locally (Arch Linux host).

behrmann:
The CI failures are not you're fault, we have them all over at the moment. I think @DaanDeMeyer is working on them.

DaanDeMeyer (approved):

PR #1118, opened 2022-08-15
Write uninitialized to /etc/machine-id when resetting

DaanDeMeyer:
This makes sure the firstboot logic is triggered when the image is
booted for the first time.


mkosi/__init__.py
```diff
@@ -3025,7 +3025,7 @@ def reset_machine_id(args: MkosiArgs, root: Path, do_run_build_script: bool, for
                 machine_id.unlink()
             except FileNotFoundError:
                 pass
-            machine_id.touch()
+            machine_id.write_text("uninitialized\n")
```
poettering, line 3028: ouch.

PR #1119, opened 2022-08-15
Support manifest files for pacman

Foxboron:
This implements manifest files for pacman with some awk-ward code.

~~Also implements support for `shell` in `run` as the code would always split up the string.~~

### DIFF
--- a/mkosi/manifest.py
+++ b/mkosi/manifest.py
@@ -6,7 +6,7 @@ from datetime import datetime
 from pathlib import Path
 from subprocess import DEVNULL, PIPE
 from textwrap import dedent
-from typing import IO, Any, Dict, List, Optional, cast
+from typing import IO, Any, Dict, List, Optional, Tuple, cast
 
 from .backend import (
     Distribution,
@@ -65,6 +65,23 @@ class SourcePackageManifest:
         return t
 
 
+def parse_pkg_desc(f: Path) -> Tuple[str, str, str, str]:
+    name = version = base = arch = ""
+    with f.open() as desc:
+        for line in desc:
+            line = line.strip()
+            if line == "%NAME%":
+                name = next(desc).strip()
+            elif line == "%VERSION%":
+                version = next(desc).strip()
+            elif line == "%BASE%":
+                base = next(desc).strip()
+            elif line == "%ARCH%":
+                arch = next(desc).strip()
+                break
+    return name, version, base, arch
+
+
 @dataclasses.dataclass
 class Manifest:
     args: MkosiArgs
@@ -81,6 +98,8 @@ class Manifest:
             self.record_rpm_packages(root)
         if cast(Any, self.args.distribution).package_type == PackageType.deb:
             self.record_deb_packages(root)
+        if cast(Any, self.args.distribution).package_type == PackageType.pkg:
+            self.record_pkg_packages(root)
         # TODO: add implementations for other package managers
 
     def record_rpm_packages(self, root: Path) -> None:
@@ -195,6 +214,20 @@ class Manifest:
                 source_package = SourcePackageManifest(source, result.stdout.strip())
                 self.source_packages[source] = source_package
 
+            source_package.add(package)
+
+    def record_pkg_packages(self, root: Path) -> None:
+        packages = sorted(root.joinpath("var/lib/pacman/local").glob("*/desc"))
+
+        for desc in packages:
+            name, version, source, arch = parse_pkg_desc(desc)
+            package = PackageManifest("pkg", name, version, arch, 0)
+            self.packages.append(package)
+
+            source_package = self.source_packages.get(source)
+            if source_package is None:
+                source_package = SourcePackageManifest(source, None)
+                self.source_packages[source] = source_package
             source_package.add(package)
 
     def has_data(self) -> bool:


behrmann (commented):
Nice! With this all major distros support the manifest. :)

mkosi/backend.py
```diff
@@ -790,13 +790,16 @@ def spawn(
...
     delay_interrupt: bool = True,
     stdout: _FILE = None,
     stderr: _FILE = None,
     env: Mapping[str, Any] = {},
     **kwargs: Any,
 ) -> CompletedProcess:
-    cmdline = [str(x) for x in cmdline]
+
+    if not shell:
+        cmdline = [str(x) for x in cmdline]
```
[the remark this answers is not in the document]
Foxboron: Yes, and with this we should not hit the case where we pass a list with `shell=True`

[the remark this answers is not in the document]
behrmann: Ah, sorry, I was misunderstanding this, since I'm mostly only used to passing lists to `subproces.run` and that's what usually (always?) happening in the rest of the codebase.

How about?
```suggestion
    cmdline = [str(x) for x in cmdline]
    if shell and isinstance(cmdline, list):
        cmdline = " ".join(t for t in cmdline))
```

Normally I'd throw in a `shlex.quote` for good measure, too, but this would preclude the use with globs that you want.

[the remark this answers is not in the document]
Foxboron: `isinstance(cmdline, list)` would always be true. I don't see much use in joining the list again when we get the string itself passed as a parameter.

mkosi/manifest.py
```diff
@@ -197,6 +199,27 @@ def record_deb_packages(self, root: Path) -> None:
 
             source_package.add(package)
 
+    def record_pkg_packages(self, root: Path) -> None:
+        c = run(f'awk \'/%NAME%/ {{getline; printf("%s ", $1)}}; /%VERSION%/ {{getline;printf("%s ", $1)}}; /%BASE%/ {{getline;printf("%s ", $1)}}; /%ARCH%/ {{getline; printf("%s\\n", $1)}}\' {root}/var/lib/pacman/local/*/desc',
+            stdout=PIPE,
+            text=True,
+            shell=True
+        )
```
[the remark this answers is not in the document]
Foxboron: Yep, this makes sense. The `awk` stuff isn't great and pacman is hopefully getting support for `-Q --print-format` in the future.

mkosi/backend.py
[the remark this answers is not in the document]
behrmann: Sorry, I didn't properly explain my thinking. 

I quickly grepped through the codebase and didn't find  an instance of `run` not being called with a list as an argument. I might have missed something, but the vast majority of call sites are like this. This can't be really encoded on the type level, though, since the `cmdline` argument for `run` is `Sequence[str]` (it's actually not, but I want to simplify this), which was chosen because sequences are immutable by default and it's nice to be general about what one accepts (also it's hard to just change this to list, because lists are invariante and sequenes are covariant so mypy complains about this). 

Unfortunately this is a bit of a footgun, because `str` is also a `Sequence[str]`, so this doesn't ensure the assumption of something that's a bag of words coming in.

Long story short: I was always thinking about this with a list going in, and in that case the join is needed to make `shell=True` work. 

You're right about the `isinstance` there being pointless, I tacked that on without thinking, because I wanted to make the case of a string being passed in work, but in any case keeping a list as input for `run` makes the most sense overall.

Foxboron:
I went with the code suggested by @behrmann and dropped the `shell` argument for the sake of keeping it simple :p

Do you want co-author credit on the commit @behrmann?

mkosi/manifest.py
```diff
@@ -197,6 +199,31 @@ def record_deb_packages(self, root: Path) -> None:
 
             source_package.add(package)
 
+    def record_pkg_packages(self, root: Path) -> None:
+        package_descriptions = [str(p) for p in Path("/var/lib/pacman/local").glob("*/desc")]
```
DaanDeMeyer, line 203: Shouldn't we be looking inside `root` here? This will operate on the host system no?

Foxboron: Argh, I didn't tripple check the path. You are correct!

behrmann: Sorry, my bad, copied that from my local test.

behrmann:
> Do you want co-author credit on the commit @behrmann?

I don't need the laurels, but it might be nice to future people for blameability. You decide.

mkosi/manifest.py
```diff
@@ -197,6 +199,31 @@ def record_deb_packages(self, root: Path) -> None:
 
             source_package.add(package)
 
+    def record_pkg_packages(self, root: Path) -> None:
+        package_descriptions = [str(p) for p in Path(f"{root}/var/lib/pacman/local").glob("*/desc")]
```
behrmann, line 203:
```suggestion
        package_descriptions = [str(p) for p in (root / "var/lib/pacman/local").glob("*/desc")]
```
or
```suggestion
        package_descriptions = [str(p) for p in root.joinpath("var/lib/pacman/local").glob("*/desc")]
```
since root already comes in as `Path`.

Foxboron: The slashes are nicer

Foxboron:
>I don't need the laurels, but it might be nice to future people for blameability. You decide.

I'm lazy and would need to google how to do this again :p So can drop it then.

mkosi/manifest.py
```diff
@@ -197,6 +199,31 @@ def record_deb_packages(self, root: Path) -> None:
...
+        package_descriptions = [str(p) for p in (root / "/var/lib/pacman/local").glob("*/desc")]
+        awkprog = """\
+BEGIN { RS="\\n\\n"; FS="\\n"; }
+/%NAME%/ { name=$2 };
+/%VERSION%/ { version=$2 };
+/%BASE%/ { base=$2 };
+/%ARCH%/ { arch=$2 };
+ENDFILE { print name, version, base, arch }
+"""
+        c = run(["awk", awkprog] + package_descriptions, stdout=PIPE, text=True)
```
poettering, line 212: hmpf, you shell out to awk for a simple parser like that? Let's maybe not do that... The awk expression above sounds simple enough to just implement in python.

Foxboron: Fixed :)

Foxboron:
Last iteration removes the awk-ward code :)

mkosi/manifest.py
```diff
@@ -197,6 +214,21 @@ def record_deb_packages(self, root: Path) -> None:
 
             source_package.add(package)
 
+    def record_pkg_packages(self, root: Path) -> None:
+        package_descriptions = [str(p) for p in (root / "var/lib/pacman/local").glob("*/desc")]
+        packages = sorted(package_descriptions)
```
behrmann, line 219:
```suggestion
        packages = sorted(root.joinpath("var/lib/pacman/local").glob("*/desc"))
```
should get rid of the mypy error (`package` is a str instead of the expected `Path`)

Foxboron: Clearly I do need to install mypy so I don't need to rely on the CI :/

mkosi/manifest.py
```diff
@@ -64,6 +64,21 @@ def report(self) -> str:
...
+            if line == "%NAME%":
+                ret.append(lines[n+1])
+            if line == "%VERSION%":
+                ret.append(lines[n+1])
+            if line == "%BASE%":
+                ret.append(lines[n+1])
+            if line == "%ARCH%":
+                ret.append(lines[n+1])
+                break
+    return ret
```
keszybz, line 81:
```
def parse_pkg_desc(f: Path) -> (str,):
     with f.open() as desc:
        for line in desc:
            match line.strip():
               case '%NAME%':
                  name = next(desc).strip()
               case '%VERSION%':
                  version = next(desc).strip()
               case '%BASE%':
                  base = next(desc).strip()
               case '%ARCH%':
                  arch = next(desc).strip()
                  return name, version, base, arch
```

Untested, but I think that's better.

keszybz: I don't know how reliable those files are. This will "crash" if some field is not found.

DaanDeMeyer: Using `match` isn't an option until we bump the minimum python version unfortunately

Foxboron: I checked if I could use match actually :) but currently `mkosi` is only up to python3.7

Foxboron: @keszybz The files should *always* have these fields present, and in *exactly* this order. There is no spec/manpage, and we can make it more flexible. But this is very well handled in `makepkg`

behrmann: We'd need to bump the Python version quite a bit. ;) We're on 3.7+ and `match` has only been available for less than a year with 3.10. I'm looking forward to using it but maybe we should do a more conservative bump before that. 

But I also prefer the pre 3.10 version of the code (among other reasons because the returned type has a more concrete shape).
```python
def parse_pkg_desc(f: Path) -> Tuple[str, str, str, str]:
    with open(f) as desc:
        for line in desc:
            line = line.strip()
            if line == "%NAME%":
                name = next(desc).strip()
            elif line == "%VERSION%":
                version = next(desc).strip()
            elif line == "%BASE%":
                base = next(desc).strip()
            elif line == "%ARCH%":
                arch = next(desc).strip()
                break
    return name, version, base, break
```

mkosi/manifest.py
```diff
@@ -197,6 +199,31 @@ def record_deb_packages(self, root: Path) -> None:
 
             source_package.add(package)
 
+    def record_pkg_packages(self, root: Path) -> None:
+        package_descriptions = [str(p) for p in (root / "/var/lib/pacman/local").glob("*/desc")]
```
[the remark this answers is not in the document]
behrmann: Yes, with the `/` before `var` this ignores the initial `root`.

mkosi/manifest.py
Foxboron: Ahh, I need to get my Python-foo up to snuff. That is much better.

mkosi/manifest.py
behrmann: It takes some getting used to, but as soon as a Python codebase gets a sufficient size it *does* help to keep ones sanity. :)

mkosi/manifest.py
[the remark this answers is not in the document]
Foxboron: Yes, but note this comment is done on an outdated revision :)

I did notice this error myself while testing.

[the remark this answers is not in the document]
keszybz: Yeah, I noticed I had some outdated comment when I was "publishing" the review. Github shows e.g. "2 comments", but if the comment is on a part of code that has been since changed, it doesn't show the comment anywhere. In the past it used to show all unpublished comments on the "comments" pane, but now it doesn't. So in the end I just pressed the green button without and forgot about this…

mkosi/manifest.py
```diff
@@ -64,6 +64,20 @@ def report(self) -> str:
             t += f"""\nChangelog:\n{self.changelog}\n"""
         return t
 
+def parse_pkg_desc(f: Path) -> Tuple[str, str, str, str]:
+    with open(f) as desc:
```
keszybz, line 68: I think `f.open()` is nicer, but OK.

keszybz:
```
pyright 1.1.258
/home/runner/work/mkosi/mkosi/mkosi/manifest.py
  /home/runner/work/mkosi/mkosi/mkosi/manifest.py:80:12 - error: "name" is possibly unbound (reportUnboundVariable)
  /home/runner/work/mkosi/mkosi/mkosi/manifest.py:80:18 - error: "version" is possibly unbound (reportUnboundVariable)
  /home/runner/work/mkosi/mkosi/mkosi/manifest.py:80:27 - error: "base" is possibly unbound (reportUnboundVariable)
  /home/runner/work/mkosi/mkosi/mkosi/manifest.py:80:33 - error: "arch" is possibly unbound (reportUnboundVariable)
```

Meh, pyright strikes again. I guess this error needs to be silenced.

mkosi/manifest.py
```diff
@@ -64,6 +64,20 @@ def report(self) -> str:
             t += f"""\nChangelog:\n{self.changelog}\n"""
         return t
 
+def parse_pkg_desc(f: Path) -> Tuple[str, str, str, str]:
+    with f.open() as desc:
```
behrmann, line 68:
```suggestion

def parse_pkg_desc(f: Path) -> Tuple[str, str, str, str]:
    name = version = base = arch = ""
    with f.open() as desc:
```
Fixes the error pyright is complaining about and makes the code not break in case of a broken `desc` file.

mkosi/manifest.py
```diff
@@ -64,6 +64,20 @@ def report(self) -> str:
...
+            if line == "%NAME%":
+                name = next(desc).strip()
+            elif line == "%VERSION%":
+                version = next(desc).strip()
+            elif line == "%BASE%":
+                base = next(desc).strip()
+            elif line == "%ARCH%":
+                arch = next(desc).strip()
+                break
+    return name, version, base, arch
```
behrmann, line 80:
```suggestion
    return name, version, base, arch

```
while looking at what pyright is complaining about, I noticed missing newlines in between the functions. We follow a two newlines between functions style, like PEP8.

behrmann (approved):

DaanDeMeyer (approved):

behrmann:
CI failure is unrelated. Merging.

PR #1123, opened 2022-08-17
Use chunked reads when inserting partition

mcassaniti:
When inserting a large partition that exceeds the available system memory, a `MemoryError` exception will occur during the partition insert. This is regardless of both the input and output files being opened in buffered I/O mode.

I did try setting a fixed size for the buffers during open without chunking the read but this made no difference.

If you want to test an alternate solution, read your entire disk out to `/dev/null` and if the memory usage of python doesn't keep going up then success. I know I've almost always had to re-write code changes, so I'm hoping this might help before someone makes a suggestion.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3713,7 +3713,10 @@ def insert_partition(
             # Let's discard the partition block device first, to ensure the GPT partition table footer that
             # likely is stored in it is flushed out. After all we want to write with dd's sparse option.
             run(["blkdiscard", path])
-            path.write_bytes(blob.read())
+
+            # Without this the entire blob will be read into memory which could exceed system memory
+            with open(path, mode='wb') as path_fp:
+                os.sendfile(path_fp.fileno(), blob.fileno(), offset=0, count=blob_size)
 
     return part
 


mkosi/__init__.py
```diff
@@ -3661,6 +3662,16 @@ def make_generated_root(args: MkosiArgs, root: Path, for_cache: bool) -> Optiona
     return None
 
 
+def read_chunked(blob: BinaryIO) -> bytes:
```
behrmann, line 3665:
```suggestion
def read_chunked(blob: BinaryIO) -> Iterator[bytes]:
```

mkosi/__init__.py
```diff
@@ -3661,6 +3662,16 @@ def make_generated_root(args: MkosiArgs, root: Path, for_cache: bool) -> Optiona
...
+    '''Read a file or binary blob in chunks'''
+
+    while(True):
+        data = blob.read(DEFAULT_BUFFER_SIZE)
+        if not data:
+            break
+        yield data
```
behrmann, line 3672: To somewhat shorten the loop (and make moving to the walrus operator later more mechanical)
```suggestion
    data = blob.read(DEFAULT_BUFFER_SIZE)
    while data
        yield data
        data = blob.read(DEFAULT_BUFFER_SIZE)
```

behrmann (commented):
LGTM, some stylistic nit and some typing appeasement.

DaanDeMeyer:
Did you try `os.sendfile()`? I'd try that first, should also be a lot more efficient if it works

behrmann:
> Did you try `os.sendfile()`? I'd try that first, should also be a lot more efficient if it works

Nice! TIL there was a sendfile wrapper in `os`.

mcassaniti:
I'll need some help with the type checking issues.

mkosi/__init__.py
mcassaniti: I've migrated to `os.sendfile()` so this should now be redundant. If that doesn't work I'll look back to your suggested changes.

mkosi/__init__.py
mcassaniti: That makes more sense. It's no longer required though.

DaanDeMeyer (approved):

PR #1126, opened 2022-08-18
Add --use-mirror-verbatim=[bool] option

bluca:
Allows to configure a single plain mirror, without the os/updates/media
repositories for RPM distros. Necessary to be able to perform fully
offline builds when the build system provides the required packages
configured in a local directory.

I'm adding mkosi support to https://openbuildservice.org/ where builds are done offline with a single, pre-prepared flat repository

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -315,6 +315,13 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
 : The mirror to use for downloading the distribution packages. Expects
   a mirror URL as argument.
 
+`UseMirrorVerbatim=`, `--use-mirror-verbatim=`
+
+: When `--mirror` is used it will be used as a plain and direct mirror instead
+  of using it as a prefix for the full set of repositories normally supported
+  by distributions like Fedora and CentOS. Useful for fully offline builds with
+  a single repository. Only supported for RPM-based distributions for now.
+
 `Repositories=`, `--repositories=`
 
 : Additional package repositories to use during installation. Expects
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2154,7 +2154,10 @@ def parse_fedora_release(release: str) -> Tuple[str, str]:
 def install_fedora(args: MkosiArgs, root: Path, do_run_build_script: bool) -> None:
     release, releasever = parse_fedora_release(args.release)
 
-    if args.mirror:
+    if args.use_mirror_verbatim and args.mirror:
+        release_url = f"baseurl={args.mirror}"
+        updates_url = None
+    elif args.mirror:
         baseurl = urllib.parse.urljoin(args.mirror, f"releases/{release}/Everything/$basearch/os/")
         media = urllib.parse.urljoin(baseurl.replace("$basearch", args.architecture), "media.repo")
         if not url_exists(media):
@@ -2168,6 +2171,10 @@ def install_fedora(args: MkosiArgs, root: Path, do_run_build_script: bool) -> No
             "metalink=https://mirrors.fedoraproject.org/metalink?"
             f"repo=updates-released-f{release}&arch=$basearch"
         )
+    if release == 'rawhide':
+        # On rawhide, the "updates" repo is the same as the "fedora" repo.
+        # In other versions, the "fedora" repo is frozen at release, and "updates" provides any new packages.
+        updates_url = None
 
     if releasever in FEDORA_KEYS_MAP:
         key = FEDORA_KEYS_MAP[releasever]
@@ -2184,9 +2191,7 @@ def install_fedora(args: MkosiArgs, root: Path, do_run_build_script: bool) -> No
     gpgurl = urllib.parse.urljoin("https://getfedora.org/static/", gpgid)
 
     repos = [Repo("fedora", release_url, gpgpath, gpgurl)]
-    if release != 'rawhide':
-        # On rawhide, the "updates" repo is the same as the "fedora" repo.
-        # In other versions, the "fedora" repo is frozen at release, and "updates" provides any new packages.
+    if updates_url is not None:
         repos += [Repo("updates", updates_url, gpgpath, gpgurl)]
 
     setup_dnf(args, root, repos)
@@ -2213,7 +2218,10 @@ def install_fedora(args: MkosiArgs, root: Path, do_run_build_script: bool) -> No
 
 @complete_step("Installing Mageia…")
 def install_mageia(args: MkosiArgs, root: Path, do_run_build_script: bool) -> None:
-    if args.mirror:
+    if args.use_mirror_verbatim and args.mirror:
+        release_url = f"baseurl={args.mirror}"
+        updates_url = None
+    elif args.mirror:
         baseurl = f"{args.mirror}/distrib/{args.release}/x86_64/media/core/"
         release_url = f"baseurl={baseurl}/release/"
         updates_url = f"baseurl={baseurl}/updates/"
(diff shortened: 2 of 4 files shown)


mkosi/__init__.py
```diff
@@ -2704,7 +2739,9 @@ def install_arch(args: MkosiArgs, root: Path, do_run_build_script: bool) -> None
     if args.release is not None:
         MkosiPrinter.info("Distribution release specification is not supported for Arch Linux, ignoring.")
 
-    if args.mirror:
+    if args.plain_mirror and args.mirror:
+        server = f"Server = {args.mirror}"
```
behrmann, line 2743: I haven't run this (and I'm not quite sure how to test this properly), but I'm a bit unsure about the semantics here. This omits the `$repo/os/$arch` trailer on the `Server` directive so that pacman would look for `core`, `extra` and `community` in the very same location, i.e. not interpolate the names of the repos (and first architecture), so that all three of them would need to be same repo. This necessitates some custom merged repo setup instead of just some mirrors, because we install packages from both `core` and `extra` (mainly dracut), no?

mkosi.md
```diff
@@ -315,6 +315,13 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
 : The mirror to use for downloading the distribution packages. Expects
   a mirror URL as argument.
 
+`PlainMirror=`, `--plain-mirror=`
+
+: When `--mirror` is used it will use it as a plain and direct mirror instead
+  of using it as a prefix for the full set of repositories normally supported
+  by distributions like Fedora and CentOS. Useful for fully offline builds with
+  a single repository.
```
behrmann, line 323: Could you add a line for which distributions this currently is supported?

behrmann (commented):
Neat, OBS support is a great thing! No objections to the design, although I'd colour the bike shed a bit differently (`--use-plain-mirror` or `--use-mirror-verbatim`).

mkosi/__init__.py
bluca: I have not tested this on arch yet, only on rpm ones

bluca:
I've managed to get the first successful Fedora image build! Plain squashfs format because those nasty partition setup race conditions are a massive pain, but it's a start https://github.com/bluca/obs-build/tree/mkosi

mkosi/__init__.py
behrmann: Might be safer to drop this for a first go then, if nobody else with more knowledge of pacman repo management innards interjects (@Foxboron maybe?).

behrmann:
Right now this doesn't do anything for Debian. Should it maybe exclude the updates and security repos there, too?

mkosi/__init__.py
Foxboron: It should just work™️. But I find this weird and I'm not sure how OBS deals with Arch repositories? Is there any docs on the offline builds somewher?

It's partially a bikeshed, but this is effectively just an online package cache with the database files, not a "plain mirror". A "flat mirror" would a better name I think. But I'll leave that up for the maintainers :)

bluca:
> Right now this doesn't do anything for Debian. Should it maybe exclude the updates and security repos there, too?

I am having problems with debootstrap, so haven't gotten that far with debian builds yet. Will likely do the same, yes

DaanDeMeyer:
--use-mirror-verbatim seems like a good name, we want the option to use complicated words because this doesn't seem like it'll be necessary all that often.

mkosi/__init__.py
bluca: I have not yet implemented it, but it will do the same thing that kiwi does, calling repo-add on the directory where the packages are downloaded in a separate step by the resolver:

https://github.com/bluca/obs-build/blob/mkosi/build-recipe-kiwi#L812

I'll remove the arch part for now

bluca:
> --use-mirror-verbatim seems like a good name, we want the option to use complicated words because this doesn't seem like it'll be necessary all that often.

updated

mkosi/__init__.py
Foxboron: Hmm, okay. But this won't give you `core`, `extra`, `community`. It would give you one repo defined to be whatever the name of the database is. If you want to make 3 repos in the same directory you'd need to specify each of the packages manually instead of doing a glob.

This implies that the `pacman.conf` as distributed by `mkosi` is not going to be compatible with the flat repository you are currently building there.

So yeah, I think removing the pacman part of this is a good idea :)

mkosi/__init__.py
```diff
@@ -6806,7 +6852,7 @@ def print_summary(args: MkosiArgs) -> None:
     MkosiPrinter.info("                   Release: " + none_to_na(args.release))
     MkosiPrinter.info("              Architecture: " + args.architecture)
     if args.mirror is not None:
-        MkosiPrinter.info("                    Mirror: " + args.mirror)
+        MkosiPrinter.info(f"            {'(plain)' if args.use_mirror_verbatim else '       '} Mirror: " + args.mirror)
```
DaanDeMeyer, line 6855: I'd update plain to verbatim here as well

mkosi/__init__.py
```diff
@@ -2847,7 +2882,10 @@ def install_opensuse(args: MkosiArgs, root: Path, do_run_build_script: bool) ->
 
     # If the release looks like a timestamp, it's Tumbleweed. 13.x is legacy (14.x won't ever appear). For
     # anything else, let's default to Leap.
-    if release.isdigit() or release == "tumbleweed":
+    if args.use_mirror_verbatim and args.mirror:
+        release_url = f"{args.mirror}"
```
DaanDeMeyer, line 2886: Why use the format string here? wouldn't just assigning to args.mirror work?

mkosi/__init__.py
```diff
@@ -2408,7 +2437,12 @@ def install_centos_stream_repos(args: MkosiArgs, root: Path, epel_release: int)
 
     release = f"{epel_release}-stream"
 
-    if args.mirror:
+    if args.use_mirror_verbatim and args.mirror:
+        appstream_url = f"baseurl={args.mirror}"
+        baseos_url = None
+        crb_url = None
+        epel_url = None
```
DaanDeMeyer, line 2444: Nit here and everywhere else: you can shorten this as follows:

```
baseos_url = crb_url = epel_url = None
```

mkosi.md
```diff
@@ -315,6 +315,13 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
 : The mirror to use for downloading the distribution packages. Expects
   a mirror URL as argument.
 
+`UseMirrorVerbatim=`, `--use-mirror-verbatim=`
+
+: When `--mirror` is used it will use it as a plain and direct mirror instead
```
DaanDeMeyer, line 320: "it will use it" => "it will be used as"

DaanDeMeyer (changes requested):
LGTM, Few nits

mkosi/__init__.py
bluca: Yes, that is how it works, all the actual repositories (in all their multiple forms) are configured on the source project and fetched by the resolver in the normal way, when the dependency resolution step for a given build is performed. This on the other hand is the isolated build step - the scheduler gets all the packages that were downloaded for a given build job, and makes them available in a local directory to the build worker root, which is booted in kvm with a custom init that just installs them all and calls the build hooks.
So at the build stage you don't have multiple repositories or multiple definitions or anything like that, because it's moot - it's completely offline with no network access whatsoever by design.

DaanDeMeyer (approved):

PR #1127, opened 2022-08-18
name squashfs images as .raw too

bluca:
This is expected by systemd when used with RootImage= and friends


behrmann:
LGTM, maybe add a line to the news, too, so that users of squash image are not surprised by the names of their outputs changing.

bluca:
> LGTM, maybe add a line to the news, too, so that users of squash image are not surprised by the names of their outputs changing.

Yeah this was likely an oversight, we expect all binary images to be .raw elsewhere in the stack, will add a note

bluca:
Added line to NEWS

DaanDeMeyer (approved):

PR #1130, opened 2022-08-18
Replace blockdev --reread-pt by manually adding partitions

DaanDeMeyer:
Instead of relying on blockdev, let's use our own infra to make
sure all partitions have appeared after we modify the partition
table. This should hopefully reduce the number of race conditions
involved with modifying the partition table and loop devices.

### DIFF
--- a/mkosi/syscall.py
+++ b/mkosi/syscall.py
@@ -1,0 +1,85 @@
+# SPDX-License-Identifier: LGPL-2.1+
+
+import ctypes
+import errno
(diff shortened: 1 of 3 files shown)


mkosi/syscall.py
```diff
@@ -0,0 +1,84 @@
+import ctypes
```
bluca, line 1: SPDX line

bluca (approved):
Fixes my issue, nice!

PR #1132, opened 2022-08-19
Change --use-mirror-verbatim to --local-mirror, support Debian/Arch/SUSE

bluca:
Also skip signature checks on both, as the repository is locally
built on-the-fly and not signed.


mkosi/__init__.py
```diff
@@ -2699,17 +2704,17 @@ def install_arch(args: MkosiArgs, root: Path, do_run_build_script: bool) -> None
...
 
-                [extra]
-                {server}
+                {"[extra]" if not args.use_mirror_verbatim else ""}
+                {f"{server}" if not args.use_mirror_verbatim else ""}
 
-                [community]
-                {server}
+                {"[community]" if not args.use_mirror_verbatim else ""}
+                {f"{server}" if not args.use_mirror_verbatim else ""}
```
behrmann, line 2717: I think this would be clearer if split into two different versions of `pacman.conf`, one for the normal case and one for `--use-mirror-verbatim`.

behrmann (commented):
LGTM, one suggestion, though.

mkosi/__init__.py
bluca: rather than two versions, I've split the write in two - there's already a separate write later with optional elements

mkosi/__init__.py
```diff
@@ -2699,22 +2710,30 @@ def install_arch(args: MkosiArgs, root: Path, do_run_build_script: bool) -> None
                 Architecture = auto
                 Color
                 CheckSpace
-                SigLevel = Required DatabaseOptional TrustAll
+                SigLevel = {sig_level}
                 ParallelDownloads = 5
 
+                {f"Include = {args.repos_dir}/*" if args.repos_dir else ""}
```
behrmann, line 2716: Moving the include up here changes the precedence of repositories defined in there to be higher than core, I'm not sure that's what we want or at least it's an unrelated change.

bluca: ah got it - I didn't know the order mattered, fixed

behrmann:
LGTM in principle. I'm a bit hesitant about mixing disabling signature checking into the option, but I see that it makes sense for the use case you target. I just wonder whether we should have named the option something like `--offline-build` or similar instead, because the name intuitively implies "no name mangling", but now the semantics is that plus disable signature checking.

bluca:
> LGTM in principle. I'm a bit hesitant about mixing disabling signature checking into the option, but I see that it makes sense for the use case you target. I just wonder whether we should have named the option something like `--offline-build` or similar instead, because the name intuitively implies "no name mangling", but now the semantics is that plus disable signature checking.

I don't mind renaming it like that, will send an update

bluca:
> > LGTM in principle. I'm a bit hesitant about mixing disabling signature checking into the option, but I see that it makes sense for the use case you target. I just wonder whether we should have named the option something like `--offline-build` or similar instead, because the name intuitively implies "no name mangling", but now the semantics is that plus disable signature checking.
> 
> I don't mind renaming it like that, will send an update

Done. Also added another fix for the hard-coded debian kernel architecture.

bluca:
> > > LGTM in principle. I'm a bit hesitant about mixing disabling signature checking into the option, but I see that it makes sense for the use case you target. I just wonder whether we should have named the option something like `--offline-build` or similar instead, because the name intuitively implies "no name mangling", but now the semantics is that plus disable signature checking.
> > 
> > 
> > I don't mind renaming it like that, will send an update
> 
> Done. Also added another fix for the hard-coded debian kernel architecture.

actually I'll drop that for now, as it's slightly more complicated to properly fix it

DaanDeMeyer:
> LGTM in principle. I'm a bit hesitant about mixing disabling signature checking into the option, but I see that it makes sense for the use case you target. I just wonder whether we should have named the option something like --offline-build or similar instead, because the name intuitively implies "no name mangling", but now the semantics is that plus disable signature checking.

Wouldn't a better fix be to just add another option to disable signature checking? That seems like a generic option that we should have regardless. Sorry for the churn on this

bluca:
> > LGTM in principle. I'm a bit hesitant about mixing disabling signature checking into the option, but I see that it makes sense for the use case you target. I just wonder whether we should have named the option something like --offline-build or similar instead, because the name intuitively implies "no name mangling", but now the semantics is that plus disable signature checking.
> 
> Wouldn't a better fix be to just add another option to disable signature checking? That seems like a generic option that we should have regardless. Sorry for the churn on this

Ok, added a separate parameter, ptal

mkosi/__init__.py
```diff
@@ -2493,8 +2493,34 @@ def invoke_apt(
         INITRD="No",
     )
 
-    run_workspace_command(args, root, cmdline, network=True, env=env)
+    # Ensure apt inside the image can see the local mirror outside of it
+    if args.offline_build and args.mirror and args.mirror.startswith("file:"):
+        nspawn_params = [f"--bind-ro={args.mirror[5:]}"]
+    else:
+        nspawn_params = None
```
behrmann, line 2500: Wouldn't bind mounting the repo on `args.mirror` starting with `file:` make sense generally irrespective of `args.offline`?

behrmann (commented):
LGTM, with the signature checking split out, I think the commit renaming `--use-mirror-verbatim` can be dropped, one quesiton though:

mkosi/__init__.py
bluca: yeah I guess so

bluca:
> LGTM, with the signature checking split out, I think the commit renaming `--use-mirror-verbatim` can be dropped, one quesiton though:

Ok, dropped, last time though :-)

bluca:
@DaanDeMeyer does this look good now?

DaanDeMeyer:
Not at the laptop atm will review tomorrow

Foxboron:
I'm a bit sceptical just having all packages as `[core]`, but probably the only half-decent solution on the pacman side of things :/

Hoping we land the `CacheServer` patches soon which would make this easier to implement.
https://bugs.archlinux.org/task/23407

mkosi/__init__.py
```diff
@@ -3222,7 +3284,7 @@ def run_postinst_script(
             nspawn_params += nspawn_params_for_blockdev_access(args, loopdev)
 
         run_workspace_command(args, root, ["/root/postinst", verb],
-                              network=(args.with_network is True),
+                              network=(args.with_network is True and args.use_mirror_verbatim is False),
```
behrmann, line 3287:
```suggestion
                              network=(args.with_network is True and not args.use_mirror_verbatim),
```
The `args.with_network is True` check is because it can also have the string value `never`

mkosi/__init__.py
```diff
@@ -7361,7 +7431,7 @@ def run_build_script(args: MkosiArgs, root: Path, raw: Optional[BinaryIO]) -> No
 
         target = f"--directory={root}" if raw is None else f"--image={raw.name}"
 
-        with_network = 1 if args.with_network is True else 0
+        with_network = 1 if args.with_network is True and args.use_mirror_verbatim is False else 0
```
behrmann, line 7434:
```suggestion
        with_network = 1 if args.with_network is True and not args.use_mirror_verbatim else 0
```

mkosi/__init__.py
```diff
@@ -7408,7 +7478,7 @@ def run_build_script(args: MkosiArgs, root: Path, raw: Optional[BinaryIO]) -> No
         if args.include_dir is not None:
             cmdline += [f"--bind={args.include_dir}:/usr/include"]
 
-        if args.with_network is True:
+        if args.with_network is True and args.use_mirror_verbatim is False:
```
behrmann, line 7481:
```suggestion
        if args.with_network is True and not args.use_mirror_verbatim:
```

behrmann (commented):
LGTM, some minor remaining nits

bluca:
> I'm a bit sceptical just having all packages as `[core]`, but probably the only half-decent solution on the pacman side of things :/
> 
> Hoping we land the `CacheServer` patches soon which would make this easier to implement. https://bugs.archlinux.org/task/23407

This is for a local only repo built on the fly and then destroyed. Only a directory with a dump of all the packages is available. No network activity is permitted.

mkosi.md
```diff
@@ -320,7 +320,12 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
 : When `--mirror` is used it will be used as a plain and direct mirror instead
   of using it as a prefix for the full set of repositories normally supported
   by distributions like Fedora and CentOS. Useful for fully offline builds with
-  a single repository. Only supported for RPM-based distributions for now.
+  a single repository. Supported on deb/rpm/arch based distributions.
+
+`RepositoryKeyCheck=`, `--repository-key-check=`
```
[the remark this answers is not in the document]
bluca: It applies to suse too

bluca:
Changed yet again: I realized that in some distros (apt, zypper) the mirror is also configured inside the image, which obviously is broken with the local one as it won't exist anymore when the image is downloaded and used. So I renamed the option to `--local-mirror` and made it a full argument rather than a boolean. `--mirror` is still used to configure the final image, while `--local-mirror` is only used at build time and then removed.

mkosi.md
```diff
@@ -315,12 +315,20 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
 : The mirror to use for downloading the distribution packages. Expects
   a mirror URL as argument.
 
-`UseMirrorVerbatim=`, `--use-mirror-verbatim=`
+`LocalMirror=`, `--local-mirror=`
 
-: When `--mirror` is used it will be used as a plain and direct mirror instead
+: The mirror will be used as a local, plain and direct mirror instead
   of using it as a prefix for the full set of repositories normally supported
   by distributions like Fedora and CentOS. Useful for fully offline builds with
```
behrmann, line 322: How about 
```suggestion
  by distributions. Useful for fully offline builds with
```
since two sentences later we say that it is supported for all major package formats.

mkosi/__init__.py
```diff
@@ -2525,8 +2553,11 @@ def install_debian_or_ubuntu(args: MkosiArgs, root: Path, *, do_run_build_script
         if debootstrap_knows_arg("--no-check-valid-until"):
             cmdline += ["--no-check-valid-until"]
 
-        assert args.mirror is not None
-        cmdline += [args.release, root, args.mirror]
+        if not args.repository_key_check:
+            cmdline += ["--no-check-gpg"]
+
+        assert args.mirror is not None or args.local_mirror is not None
+        cmdline += [args.release, root, typing.cast(Union[Path, str], args.local_mirror if args.local_mirror else args.mirror)]
```
behrmann, line 2560:
```suggestion
        cmdline += [args.release, root, cast(PathString, args.local_mirror or args.mirror)]
```
`cast` is already imported from `typing` and the path string union is used so often that we have this type alias and `a if a else b` can be written as `a or b`.

mkosi/__init__.py
```diff
@@ -2493,8 +2494,35 @@ def invoke_apt(
         INITRD="No",
     )
 
-    run_workspace_command(args, root, cmdline, network=True, env=env)
+    # Ensure apt inside the image can see the local mirror outside of it
+    if args.local_mirror:
+        mirror : Optional[str] = args.local_mirror
```
behrmann, line 2499: This cannot be `Optional[str]`, because the check in the line before will ensure `args.local_mirror` is not `None` and not the empty string, so the line conditional can be condensed to
```python
mirror = args.local_mirror or args.mirror
```

mkosi/__init__.py
```diff
@@ -2493,8 +2494,35 @@ def invoke_apt(
...
+    else:
+        mirror = args.mirror
+    if mirror and mirror.startswith("file:"):
```
behrmann, line 2502: With the assumptions from the above comment this can be just 
```suggestion
    if mirror.startswith("file:"):
```

Would it also make sense to ensure in `load_args` that on Debian of the argument of `--local-mirror` start with `file:`?

mkosi/__init__.py
```diff
@@ -2493,8 +2494,35 @@ def invoke_apt(
...
+        nspawn_params = [f"--bind-ro={mirror[5:]}"]
+    else:
+        nspawn_params = None
+
+    run_workspace_command(args, root, cmdline, network=(not args.local_mirror), env=env, nspawn_params=nspawn_params)
```
behrmann, line 2507: I think the `not args.local_mirror` has the right semantics, but I think this is a bit clearer, since `args.local_mirror=""` doesn't make sense anyway. 
```suggestion
    run_workspace_command(args, root, cmdline, network=(args.local_mirror is None), env=env, nspawn_params=nspawn_params)
```

mkosi/__init__.py
bluca: Back in MY day, if a variable walked like a duck and quacked like a duck, then it was a duck! None of this typing malarky!

mkosi/__init__.py
bluca: > Would it also make sense to ensure in load_args that on Debian of the argument of --local-mirror start with file:?

I don't think so, technically a local mirror might be on a different machine on the same local network, with the same principle applying that the local one is used only for the build, but not configured in the image

mkosi/__init__.py
bluca: I've done this, but now mypy fails with:

```
mkosi/__init__.py:2498:8: error: Item "None" of "Optional[str]" has no
attribute "startswith"  [union-attr]
        if mirror.startswith("file:"):
           ^
mkosi/__init__.py:2499:26: error: Value of type "Optional[str]" is not
indexable  [index]
            nspawn_params = [f"--bind-ro={mirror[[5](https://github.com/systemd/mkosi/runs/7970986764?check_suite_focus=true#step:7:6):]}"]
```

behrmann: Ah, my bad I misremembered `args.mirror` as `str`, but it is also `Optional[Str]` so the check needs to stay
```python
if mirror and mirror.startswith("file:")
```

mkosi/__init__.py
```diff
@@ -2493,8 +2493,34 @@ def invoke_apt(
         INITRD="No",
     )
 
-    run_workspace_command(args, root, cmdline, network=True, env=env)
+    # Ensure apt inside the image can see the local mirror outside of it
+    if args.mirror and args.mirror.startswith("file:"):
```
DaanDeMeyer: I'd really prefer a new --bind option to bind mount directories from the host into the container when building instead of doing it this way.

(Or someone finally switches us to mmdebstrap so we don't have to run apt inside the container anymore)

mkosi/__init__.py
```diff
@@ -2625,6 +2640,16 @@ def install_debian_or_ubuntu(args: MkosiArgs, root: Path, *, do_run_build_script
...
 
+    # Now clean up and add the real repositories, so that the image is ready
+    if args.use_mirror_verbatim:
+        if args.distribution == Distribution.ubuntu:
+            main_repo = f"deb http://archive.ubuntu.com/ubuntu {args.release} {' '.join(repos)}\n"
+        else:
+            main_repo = f"deb http://deb.debian.org/debian {args.release} {' '.join(repos)}\n"
+        root.joinpath("etc/apt/sources.list").write_text(main_repo)
+        root.joinpath("etc/apt/sources.list.d/mirror.list").unlink()
+        add_apt_auxiliary_repos(args, root, repos)
```
DaanDeMeyer, line 2654: Technically the mirrors are only intended for building the image, not necessarily to make sure installing packages afterwards works correctly, it just happens to work that way for apt because we have to run it from inside the container unlike the other package managers. But I guess there's no reason to regress here unnecessarily so we can keep this.

behrmann (approved):
LGTM. Maybe `--repository-key-check` could be extended to work for the RH family as well, but that doesn't need to happen in this PR.

mkosi/__init__.py
```diff
@@ -2680,7 +2705,9 @@ def install_arch(args: MkosiArgs, root: Path, do_run_build_script: bool) -> None
     if args.release is not None:
         MkosiPrinter.info("Distribution release specification is not supported for Arch Linux, ignoring.")
 
-    if args.mirror:
+    if args.use_mirror_verbatim and args.mirror:
```
DaanDeMeyer: Shouldn't this be moved inside the original `if args.mirror` conditional?

mkosi/__init__.py
```diff
@@ -3214,7 +3257,7 @@ def run_postinst_script(
             nspawn_params += nspawn_params_for_blockdev_access(args, loopdev)
 
         run_workspace_command(args, root, ["/root/postinst", verb],
-                              network=(args.with_network is True),
+                              network=(args.with_network is True and not args.use_mirror_verbatim),
```
DaanDeMeyer: I wouldn't couple `--use-mirror-verbatim` to the network option. It's not because someone is using a local mirror that they will not want to access the network in the post install script.

mkosi/__init__.py
```diff
@@ -7353,7 +7396,7 @@ def run_build_script(args: MkosiArgs, root: Path, raw: Optional[BinaryIO]) -> No
 
         target = f"--directory={root}" if raw is None else f"--image={raw.name}"
 
-        with_network = 1 if args.with_network is True else 0
+        with_network = 1 if args.with_network is True and not args.use_mirror_verbatim else 0
```
DaanDeMeyer: Ditto

mkosi/__init__.py
```diff
@@ -7400,7 +7443,7 @@ def run_build_script(args: MkosiArgs, root: Path, raw: Optional[BinaryIO]) -> No
         if args.include_dir is not None:
             cmdline += [f"--bind={args.include_dir}:/usr/include"]
 
-        if args.with_network is True:
+        if args.with_network is True and not args.use_mirror_verbatim:
```
DaanDeMeyer: Ditto

mkosi/__init__.py
bluca: But then you need to pass different options depending on the type of mirror? It seems much better to me to do it this way, as it's simpler for users - it's trivial to detect local vs remote (file:) so doing the right thing automatically is nice and easy

bluca: and also this would be a specific option for debian/ubuntu builds, which is not nice either (I need to use the same arguments for every build in OBS)

behrmann: I agree with @bluca here. A general bind option would be neat, but I think that's a bit beyond this PR. Because either `--local-mirror` would still be needed for what it does here or the whole rigamarole would need to be implemented as some mix of build and postinst scripts, which would be cumbersome for those users.

Also, is GH acting up for me or is the review for an older version of the PR?

mkosi/__init__.py
bluca: this is rearranged again by a later commit, so no longer applies

mkosi/__init__.py
bluca: There was some side-effect when using with-network=no, let me check again

bluca: ah yes, `-C` is passed to `dnf` which breaks it, maybe I can just add the workaround for that

bluca: ok, I've changed this to just skip `-C` with a TODO to figure it out later, and remove the rest.

bluca:
> LGTM. Maybe `--repository-key-check` could be extended to work for the RH family as well, but that doesn't need to happen in this PR.

Yeah `-C` to `dnf` breaks the build, so I'd leave it out for now, I don't need it for my use case

bluca:
fixed merged conflict from latest main update

mkosi/__init__.py
```diff
@@ -2425,6 +2426,35 @@ def debootstrap_knows_arg(arg: str) -> bool:
...
+    mirror = args.local_mirror or args.mirror
+    if mirror and mirror.startswith("file:"):
+        # Strip leading/trailing '/' as Path() does not behave well when concatenating
+        mirror_dir = mirror[5:].strip("/")
+    else:
+        yield
+        return
+
+    with complete_step("Mounting apt local mirror…", "Unmounting apt local mirror…"), contextlib.ExitStack() as stack:
+        os.makedirs(root / mirror_dir, mode=0o755, exist_ok=True)
```
DaanDeMeyer, line 2452: `mount_bind()` will automatically create any directories that don't exist yet so this can be dropped

mkosi/__init__.py
```diff
@@ -2425,6 +2426,35 @@ def debootstrap_knows_arg(arg: str) -> bool:
...
+@contextlib.contextmanager
+def mount_apt_local_mirror(args: MkosiArgs, root: Path) -> Iterator[None]:
+    # Ensure apt inside the image can see the local mirror outside of it
+    mirror = args.local_mirror or args.mirror
+    if mirror and mirror.startswith("file:"):
+        # Strip leading/trailing '/' as Path() does not behave well when concatenating
+        mirror_dir = mirror[5:].strip("/")
+    else:
+        yield
+        return
```
DaanDeMeyer, line 2449: Can we do an early return here instead?

```
if not mirror or not mirror.startswith("file:"):
    yield
    return
```

mkosi/__init__.py
```diff
@@ -2447,10 +2477,27 @@ def invoke_apt(
         INITRD="No",
     )
 
-    with mount_api_vfs(args, root):
+    with mount_api_vfs(args, root), mount_apt_local_mirror(args, root), mount_apt_override_config(root):
```
DaanDeMeyer, line 2480: Do we need to mount the local mirror everytime we invoke apt or is it sufficient to do it for the "install" operation?

mkosi/__init__.py
```diff
@@ -2480,8 +2527,11 @@ def install_debian_or_ubuntu(args: MkosiArgs, root: Path, *, do_run_build_script
         if debootstrap_knows_arg("--no-check-valid-until"):
             cmdline += ["--no-check-valid-until"]
 
-        assert args.mirror is not None
-        cmdline += [args.release, root, args.mirror]
+        if not args.repository_key_check:
+            cmdline += ["--no-check-gpg"]
+
+        assert args.mirror is not None or args.local_mirror is not None
+        cmdline += [args.release, root, cast(PathString, args.local_mirror or args.mirror)]
```
DaanDeMeyer, line 2534:
```
mirror = args.local_mirror or args.mirror
assert mirror is not None
...
```

mkosi/__init__.py
```diff
@@ -2815,13 +2894,16 @@ def install_opensuse(args: MkosiArgs, root: Path, do_run_build_script: bool) ->
...
+        "--gpg-auto-import-keys" if args.repository_key_check else "--no-gpg-checks",
         "install",
         "-y",
         "--no-recommends",
         "--download-in-advance",
         *sort_packages(packages),
     ]
+    if args.local_mirror:
+        cmdline.insert(1, "--reposd-dir")
+        cmdline.insert(2, workspace(root) / "zypper-repos.d")
```
DaanDeMeyer, line 2906: You can inline this into the definition:

```
*(["--reposd-dir", workspace(root) / "zypper-repos.d"] if args.local_mirror else [])
```

DaanDeMeyer (changes requested):
Can we drop the last commit and merge my other PR for that instead?

mkosi/__init__.py
```diff
@@ -2425,6 +2426,23 @@ def debootstrap_knows_arg(arg: str) -> bool:
...
+def mount_apt_local_mirror(args: MkosiArgs, root: Path) -> Iterator[None]:
+    # Ensure apt inside the image can see the local mirror outside of it
+    mirror = args.local_mirror or args.mirror
+    if mirror and mirror.startswith("file:"):
+        # Strip leading/trailing '/' as Path() does not behave well when concatenating
+        mirror_dir = mirror[5:].strip("/")
+    else:
+        return None
+
+    with complete_step("Mounting apt local mirror…", "Unmounting apt local mirror…"), contextlib.ExitStack() as stack:
```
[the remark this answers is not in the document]
bluca: What do you mean precisely?

mkosi/__init__.py
```diff
@@ -2425,6 +2426,23 @@ def debootstrap_knows_arg(arg: str) -> bool:
     return bytes("invalid option", "UTF-8") not in run(["debootstrap", arg], stdout=PIPE, check=False).stdout
 
 
+@contextlib.contextmanager
+def mount_apt_local_mirror(args: MkosiArgs, root: Path) -> Iterator[None]:
+    # Ensure apt inside the image can see the local mirror outside of it
+    mirror = args.local_mirror or args.mirror
```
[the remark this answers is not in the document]
bluca: it was a previous review comment iirc - it won't hurt anyway

mkosi/__init__.py
```diff
@@ -2448,8 +2466,29 @@ def invoke_apt(
     )
 
     with mount_api_vfs(args, root):
-        return run(cmdline, env=env, text=True, **kwargs)
+        with mount_apt_local_mirror(args, root):
```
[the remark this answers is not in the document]
bluca: it is not nested? old revision?

mkosi/__init__.py
bluca: yes, the repository metadata is also in the mirror, so update/search need it too

mkosi/__init__.py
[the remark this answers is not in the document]
DaanDeMeyer:
```
contextlib.ExitStack() as stack
```

The above can be dropped. It's only needed when you don't know the number of context managers up front.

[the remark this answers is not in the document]
bluca: got it, removed

mkosi/__init__.py
```diff
@@ -2480,8 +2530,12 @@ def install_debian_or_ubuntu(args: MkosiArgs, root: Path, *, do_run_build_script
...
             cmdline += ["--no-check-valid-until"]
 
-        assert args.mirror is not None
-        cmdline += [args.release, root, args.mirror]
+        if not args.repository_key_check:
+            cmdline += ["--no-check-gpg"]
+
+        mirror = args.local_mirror or args.mirror
+        assert mirror is not None
+        cmdline += [args.release, root, cast(PathString, mirror)]
```
DaanDeMeyer, line 2538: Pretty sure we don't need the cast anymore now

mkosi/__init__.py
```diff
@@ -2586,10 +2638,13 @@ def install_debian_or_ubuntu(args: MkosiArgs, root: Path, *, do_run_build_script
         # Debian still has pam_securetty module enabled, disable it in the base image.
         disable_pam_securetty(root)
 
-    if args.distribution == Distribution.debian and "systemd" in extra_packages:
+    if (args.distribution == Distribution.debian and "systemd" in extra_packages and
+            ("systemd-resolved" not in extra_packages)):
```
DaanDeMeyer, line 2642: Why would we not want to fixup resolv.conf if systemd-resolved is in the package list?

DaanDeMeyer (changes requested):
Two more things

mkosi/__init__.py
bluca: because it is done automatically by the package (it's the reason I added a separate package for it, to takeover automatically)

DaanDeMeyer (approved):

PR #1134, opened 2022-08-19
Include image ID and version in environment variables

mcassaniti:
This change will allow the image ID and image version arguments to be referenced in scripts (prepare, build, postinst, finalize). This is useful for substituting or adding into relevant files.

For example:

```shell
$ cat mkosi.postinst
#!/bin/sh

...

echo "IMAGE_ID=$IMAGE_ID"           >> /etc/os-release
echo "IMAGE_VERSION=$IMAGE_VERSION" >> /etc/os-release

# These files were added from mkosi.extra
for f in $(find /usr/lib/sysupdate.d/ -name '*.conf') ; do
    sed --in-place s/'IMAGE_ID'/"$IMAGE_ID"/ $f
done
```

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3165,7 +3165,7 @@ def run_prepare_script(state: MkosiState, cached: bool) -> None:
 
         nspawn_params = nspawn_params_for_build_sources(state.config, SourceFileTransfer.mount)
         run_workspace_command(state, ["/root/prepare", verb],
-                              network=True, nspawn_params=nspawn_params, env=state.config.environment)
+                              network=True, nspawn_params=nspawn_params, env=state.environment)
 
         srcdir = root_home(state) / "src"
         if srcdir.exists():
@@ -3192,7 +3192,7 @@ def run_postinst_script(state: MkosiState) -> None:
         shutil.copy2(state.config.postinst_script, root_home(state) / "postinst")
 
         run_workspace_command(state, ["/root/postinst", verb],
-                              network=(state.config.with_network is True), env=state.config.environment)
+                              network=(state.config.with_network is True), env=state.environment)
         root_home(state).joinpath("postinst").unlink()
 
 
@@ -3210,8 +3210,7 @@ def run_finalize_script(state: MkosiState) -> None:
 
     with complete_step("Running finalize script…"):
         run([state.config.finalize_script, verb],
-            env={**state.config.environment, "BUILDROOT": str(state.root), "OUTPUTDIR": str(output_dir(state.config))})
-
+            env={**state.environment, "BUILDROOT": str(state.root), "OUTPUTDIR": str(output_dir(state.config))})
 
 
 def install_boot_loader(
@@ -6495,6 +6494,11 @@ def load_args(args: argparse.Namespace) -> MkosiConfig:
     else:
         args.environment = {}
 
+    if args.image_id is not None:
+        args.environment['IMAGE_ID'] = args.image_id
+    if args.image_version is not None:
+        args.environment['IMAGE_VERSION'] = args.image_version
+
     if args.cache_path is not None:
         args.cache_path = args.cache_path.absolute()
 
@@ -7265,12 +7269,6 @@ def run_build_script(state: MkosiState, raw: Optional[BinaryIO]) -> None:
                 f"--setenv=MKOSI_DEFAULT={state.config.config_path}"
             ]
 
-        if state.config.image_version is not None:
-            cmdline += [f"--setenv=IMAGE_VERSION={state.config.image_version}"]
-
-        if state.config.image_id is not None:
-            cmdline += [f"--setenv=IMAGE_ID={state.config.image_id}"]
-
         cmdline += nspawn_params_for_build_sources(state.config, state.config.source_file_transfer)
 
         if state.config.build_dir is not None:
@@ -7292,7 +7290,7 @@ def run_build_script(state: MkosiState, raw: Optional[BinaryIO]) -> None:
         if state.config.nspawn_keep_unit:
             cmdline += ["--keep-unit"]
 
-        cmdline += [f"--setenv={env}={value}" for env, value in state.config.environment.items()]
+        cmdline += [f"--setenv={env}={value}" for env, value in state.environment.items()]
 
         cmdline += [f"/root/{state.config.build_script.name}"]
 
@@ -7373,6 +7371,7 @@ def build_stuff(config: MkosiConfig) -> Manifest:
             workspace=Path(workspace.name),
             cache=cache,
             do_run_build_script=False,
+            environment=config.environment,
             machine_id=config.machine_id or uuid.uuid4().hex,
             for_cache=False,
         )
--- a/mkosi/backend.py
+++ b/mkosi/backend.py
@@ -506,6 +506,7 @@ class MkosiConfig:
     skeleton_trees: List[Path]
     clean_package_metadata: Union[bool, str]
     remove_files: List[Path]
+    # Environment should not be used directly. Use MkosiState environment instead.
     environment: Dict[str, str]
     build_sources: Optional[Path]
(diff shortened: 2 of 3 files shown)


mkosi/backend.py
```diff
@@ -694,6 +694,12 @@ def run_workspace_command(
     if "workspace-command" in ARG_DEBUG:
         nspawn += ["--setenv=SYSTEMD_LOG_LEVEL=debug"]
 
+    if args.image_id:
+        nspawn += [f"--setenv={args.image_id}"]
+
+    if args.image_version:
+        nspawn += [f"--setenv={args.image_version}"]
```
behrmann, line 701: Aren't the environment variable names missing here?
```suggestion
    if args.image_id:
        nspawn += [f"--setenv=IMAGE_ID={args.image_id}"]

    if args.image_version:
        nspawn += [f"--setenv=IMAGE_VERSION={args.image_version}"]
```

mcassaniti: 🤦‍♂️ Whoops. Fixed now.

behrmann (approved):

mcassaniti:
@behrmann Who does the merging once this is approved?

behrmann:
> @behrmann Who does the merging once this is approved?

Anybody in the maintainer's group, as a rule of thumb those are people with the *member* pill next to their name. My usual personal policy is to merge stuff that two people have looked at, i.e. approved or said their LGTM to.

behrmann:
Not sure the errors qemu spews are connected to this, since at least for Debian I think that nothing should pick up the `IMAGE_ID` (the `kernel-install` in the systemd package is too old, and it only reacts to `IMAGE_ID` in the environment anyway only when `/etc/kernel/entry-token` doesn't exist, which it does for us).

Thinking about this again, it might be cleaner though, to just pass in these variables as the `env` parameter in all the `run_foo_script` functions.

mkosi/__init__.py
```diff
@@ -3203,11 +3204,26 @@ def nspawn_params_for_build_sources(args: MkosiArgs, sft: SourceFileTransfer) ->
     else:
         params += ["--chdir=/root"]
 
-    params += [f"--setenv={env}={value}" for env, value in args.environment.items()]
+    params += [
+        f"--setenv={env}={value}"
+        for env, value in extend_run_env(args, args.environment).items()
+    ]
```
behrmann, line 3210: Looking at the call sites of `nspaw_params_for_build_sources`, I see that the variables are already set for build scripts, so those lines should be removed then.

mcassaniti: When I look the call sites are [here](https://github.com/systemd/mkosi/blob/59aa8283d5ce564d5ea7c04103168e4cf35eb552/mkosi/__init__.py#L3179) in `run_prepare_script()` and [here](https://github.com/systemd/mkosi/blob/59aa8283d5ce564d5ea7c04103168e4cf35eb552/mkosi/__init__.py#L7394) in `run_build_script()`. It's the second call site that sets these variables but not the first. I'll remove [these lines](https://github.com/systemd/mkosi/blob/59aa8283d5ce564d5ea7c04103168e4cf35eb552/mkosi/__init__.py#L7388-#L7392) from `run_build_script()` to compensate.

mcassaniti:
I've done a bit of a cleanup as well for `run_build_script()` because there was redundancy between setting the environment variables there and also in `nspawn_params_for_build_sources()`.

mkosi/__init__.py
```diff
@@ -3203,11 +3205,28 @@ def nspawn_params_for_build_sources(args: MkosiArgs, sft: SourceFileTransfer) ->
...
 
     return params
 
 
+def extend_run_env(args: MkosiArgs, env: Mapping[str, str]) -> Mapping[str, str]:
```
behrmann, line 3216:
```suggestion
def extend_run_env(args: MkosiArgs, env: Mapping[str, str]) -> Dict[str, str]:
```
since it's clear what we return here. I think that will also make the cast further below unnecessary.

mcassaniti: Fixed. It seems you were right about the additional cast as well.

behrmann (approved):
LGTM

mcassaniti:
@bluca Rebase done.

mcassaniti:
@bluca, @behrmann, @DaanDeMeyer who wants to merge this one?

DaanDeMeyer (changes requested):
Now that MkosiState is a thing, we should do this a little differently.

Similar to MkosiConfig, we should have a field `environment` in MkosiState, that is initialized from `config.environment` + plus any extras we want to add on top, like `IMAGE_ID` and `IMAGE_VERSION`

Then, we should pass `state.environment` instead of `config.environment` everywhere.

mcassaniti:
> Now that MkosiState is a thing, we should do this a little differently.
> 
> Similar to MkosiConfig, we should have a field `environment` in MkosiState, that is initialized from `config.environment` + plus any extras we want to add on top, like `IMAGE_ID` and `IMAGE_VERSION`
> 
> Then, we should pass `state.environment` instead of `config.environment` everywhere.

We found that adding `IMAGE_ID` and `IMAGE_VERSION` to the environment all the time may have ran into issues (somehow) with `debian_gpt4` CI tests under qemu. See [this](https://github.com/systemd/mkosi/pull/1134#issuecomment-1220531516) previous comment requesting only changing specific call sites.

DaanDeMeyer:
That should be fixed after https://github.com/systemd/mkosi/pull/1170

mcassaniti:
I'll wait until #1170 is merged before I update the PR. I'd rather have the CI tests pass. I'm also testing my patches with my own builds and I'd like to confirm they work before I push.

mcassaniti:
Take a look now and see what you think.

mkosi/backend.py
```diff
@@ -608,6 +610,13 @@ class MkosiState:
     def root(self) -> Path:
         return self.workspace / "root"
 
+    def init_env(self) -> None:
+        self.environment = dict(self.config.environment)
+        if self.config.image_id is not None:
+            self.environment['IMAGE_ID'] = self.config.image_id
+        if self.config.image_version is not None:
+            self.environment['IMAGE_VERSION'] = self.config.image_version
```
DaanDeMeyer, line 618: No need for this method, we can just pass the environment to the MkosiConfig constructor

DaanDeMeyer (changes requested):

mkosi/backend.py
mcassaniti: I believe this is now resolved.

mkosi/__init__.py
```diff
@@ -6495,6 +6494,11 @@ def load_args(args: argparse.Namespace) -> MkosiConfig:
     else:
         args.environment = {}
 
+    if args.image_id is not None:
+        args.environment['IMAGE_ID'] = args.image_id
+    if args.image_version is not None:
+        args.environment['IMAGE_VERSION'] = args.image_version
```
DaanDeMeyer, line 6500: These should not be set in `args.environment`, since then they'll end up in `config.environment`. We only want user config in `config.environment`, not the extra stuff we determine ourselves, so this can't be added to `args.environment`, you need to make sure it only exists in `state.environment` and not in `config.environment`

DaanDeMeyer (changes requested):

mkosi/__init__.py
mcassaniti: I'm not sure how you would like this resolved. I could essentially turn that `init_env()` method I wrote before into something that wraps `config.environment` during `MkosiState` instantiation, but I'd like some clarification. By this I mean:

```
state = MkosiState(
    ...
    environment = adjust_env(config),
    ...
)
```

mcassaniti: I could also use getters and setters, but it's getting a bit ugly.

DaanDeMeyer:
Fixed by #1178

PR #1135, opened 2022-08-19
Fixed compression suffix

mcassaniti:
If for example zstd compression is enabled then the output file is `artifact.rawzstd` instead of `artifact.raw.zstd`.


bluca (approved):

behrmann:
Took me a bit of waffling whether this is correct, but I think it is and I wonder how that never came up before.

PR #1137, opened 2022-08-19
Revert "Drop call to blkdiscard"

DaanDeMeyer:
This reverts commit bb41549f4b2db20fedb988a1b382ff73adbfbc14.


DaanDeMeyer:
Actually, this is only triggerd in the insert_generated_partition() codepath, which shouldn't be triggered by the qemu CI, so I'm not sure if it's this after all

bluca:
Can failure to run it be made non-fatal if it's reintroduced please?

DaanDeMeyer:
This and the other one are just to figure out what's making CI fail

PR #1139, opened 2022-08-19
Always use a pseudo-tty when boot verb is used

DaanDeMeyer:
A container where we boot systemd is not useful without /dev/console
in it. If --pipe is used, /dev/console is not available. To make sure
/dev/console is always available when we're booting the container,
only apply the --pipe logic when booting in shell mode.

Fixes missing console output when booting a container image in a
non-interactive shell script.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -7621,11 +7621,6 @@ def run_shell_cmdline(args: MkosiArgs, pipe: bool = False, commands: Optional[Se
 
     cmdline = [nspawn_executable(), "--quiet", target]
 
-    # Redirecting output correctly when not running directly from the terminal.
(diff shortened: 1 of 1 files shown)


bluca (approved):

mkosi/__init__.py
```diff
@@ -7638,6 +7633,11 @@ def run_shell_cmdline(args: MkosiArgs, pipe: bool = False, commands: Optional[Se
     else:
         cmdline += nspawn_rlimit_params()
 
+        # Redirecting output correctly when not running directly from the terminal.
+        console_arg = f"--console={'interactive' if not pipe else 'pipe'}"
```
behrmann, line 7637: Doesn't this make us fall back to `--console=read-only` in the `boot` case when run in CI?

PR #1142, opened 2022-08-21
Remove all existing partition devices before running sfdisk

DaanDeMeyer:
Hopefully helps to avoid race conditions in loop device handling.


bluca:
I'll test this shortly in obs too which suffers from these issues badly

bluca:
> I'll test this shortly in obs too which suffers from these issues badly

I've seen no regression with this, ship it!

PR #1143, opened 2022-08-21
Make Debian kernel package code more prose-like

behrmann:
Since I wasn't early enough out of bed to get in on the action in #1140, here a little followup to. Feel free to ignore, but I think it's a bit easier to read through.


bluca (approved):

bluca:
```
            elif args.distribution == Distribution.debian:
>               for package in extra_packages:
E               RuntimeError: Set changed size during iteration
```

behrmann:
Sorry for the noise, silly mistake when copying into Emacs, after first typing that in a comment field here, but thinking about this, this is a bit simpler still.

bluca (approved):

PR #1144, opened 2022-08-21
manifest: include metadata about the image itself

bluca:
It would be useful to record more metadata than just the list of packages. eg:

```
{
  "image": {
    "distribution": "opensuse",
    "release": "tumbleweed",
    "architecture": "x86_64",
    "format": "plain_squashfs",
    "version": "1"
  },
  "packages": [
    {
      "type": "rpm",
      "name": "aaa_base",
      "version": "84.87+g",
      "architecture": "x86_64"
    },
<...>
```

### DIFF
--- a/mkosi/manifest.py
+++ b/mkosi/manifest.py
@@ -234,7 +234,22 @@ class Manifest:
         return len(self.packages) > 0
 
     def as_dict(self) -> Dict[str, Any]:
+        config = {
+            "name": self.args.image_id or "image",
+            "distribution": self.args.distribution.name,
+            "architecture": self.args.architecture,
+        }
+        if self.args.image_version is not None:
+            config["version"] = self.args.image_version
+        if self.args.release is not None:
+            config["release"] = self.args.release
+
         return {
+            # Bump this when incompatible changes are made to the manifest format.
+            "manifest_version": 1,
+            # Describe the image itself.
+            "config": config,
+            # Describe the image content in terms of packages.
             "packages": [package.as_dict() for package in self.packages],
         }
 


mkosi/manifest.py
```diff
@@ -234,7 +234,21 @@ def has_data(self) -> bool:
...
+        image = {
+            "distribution": self.args.distribution.name,
+            "release": self.args.release,
+            "architecture": self.args.architecture,
+            "format": self.args.output_format.name,
+        }
+        if self.args.image_id is not None:
+            image["name"] = self.args.image_id
+        else:
+            image["name"] = "image"
```
behrmann, line 246:
```suggestion
        image["name"] = self.args.image_id or "image"
```
(this does has a semantic difference, though, `name` cannot be set to the empty string)

mkosi/manifest.py
```diff
@@ -234,7 +234,21 @@ def has_data(self) -> bool:
...
+        if self.args.image_version is not None:
+            image["version"] = self.args.image_version
```
behrmann, line 248: Could we maybe also set this to some fallback value? This would make it easier for users, since they can assume keys are just there.

mkosi/manifest.py
```diff
@@ -234,7 +234,21 @@ def has_data(self) -> bool:
...
+
         return {
+            "image": image,
```
behrmann, line 251: While at it, we could maybe add a `_manifest_version` key, since we'll probably touch this in the future and if we ever get external users this would make their life easier.

Such a key could either be the same as the mkosi release version or start at 1 now and be counted up on every release that touches the format. I'd lean towards the latter.

behrmann (commented):
/cc @keszybz @Foxboron

Foxboron:
I'd move the version value to the top-level? I'd also include the compression flag.

With these additions I could start implementing a `reproduce` subcommand to `mkosi` actually.

behrmann:
> I'd move the version value to the top-level?

The image version or the manifest version I'd like to add (if the latter, that's only mentioned on the line itself because of GH's UI.)

Foxboron:
I meant the manfiest version :)

mkosi/manifest.py
bluca: Checking for a key existence when parsing json is easy. If there's no version then we don't use it anywhere in the image, so we shouldn't make one up for the manifest. I intend to use the manifest content to make decisions about the image itself.

bluca:
Added the manifest_version field

bluca:
> I eventually want to serialize the full config used to build the image in the manifest. Can we rename the "image" field to "config" instead? Otherwise we'd end up duplicating information beneath "image" and "config"

Sure, done

Foxboron:
Could we also get the compression flag as mentioned :)?

bluca:
> Could we also get the compression flag as mentioned :)?

That is more complex because of the interleaving between the FS compression and outer compression, so I'd rather leave that for when Daan will do the full config serialization, as it will likely have to change at that point

Foxboron:
Ah okay, that is fine :)

DaanDeMeyer (approved):
Just a reminder that this won't be stable until it gets in a release, (and maybe not after that either), so I reserve the right to change the format as needed. (Except for the packages field, since that's been available in stable releases for a while)

bluca:
> Just a reminder that this won't be stable until it gets in a release, (and maybe not after that either), so I reserve the right to change the format as needed. (Except for the packages field, since that's been available in stable releases for a while)

No problem, just CC me in the PR if you do, so that I can change anything if needed

PR #1147, opened 2022-08-23
arch: Modify SigLevel to pacman default

DaanDeMeyer:
Let's use the same default as pacman uses in Arch


behrmann (approved):
Sensible change, but should maybe be mentioned in the changelog for people relying that, that install some homebrew repos (although I'd imagine this to be a very niche problem).

DaanDeMeyer:
Only NEWS update, let's not wait for CI

PR #1148, opened 2022-08-24
Split MkosiArgs into MkosiConfig and MkosiState

rphibel:
This split will enable serializing the configuration used to build the
image

### DIFF
--- a/mkosi/backend.py
+++ b/mkosi/backend.py
@@ -450,7 +450,7 @@ class PartitionTable:
 
 
 @dataclasses.dataclass
-class MkosiArgs:
+class MkosiConfig:
     """Type-hinted storage for command line arguments."""
 
     verb: Verb
@@ -570,9 +570,7 @@ class MkosiArgs:
     # systemd-nspawn specific options
     nspawn_keep_unit: bool
 
-    # Some extra stuff that's stored in MkosiArgs for convenience but isn't populated by arguments
     machine_id_is_fixed: bool
-    original_umask: int
     passphrase: Optional[Dict[str, str]]
 
     output_checksum: Optional[Path] = None
@@ -585,6 +583,17 @@ class MkosiArgs:
     output_split_verity: Optional[Path] = None
     output_split_verity_sig: Optional[Path] = None
     output_split_kernel: Optional[Path] = None
+
+    def architecture_is_native(self) -> bool:
+        return self.architecture == platform.machine()
+
+
+@dataclasses.dataclass
+class MkosiState:
+    """State related properties."""
+
+    original_umask: int
+
     cache_pre_inst: Optional[Path] = None
     cache_pre_dev: Optional[Path] = None
     output_signature: Optional[Path] = None
@@ -597,11 +606,8 @@ class MkosiArgs:
             return None
         return self.partition_table.partitions.get(ident)
 
-    def architecture_is_native(self) -> bool:
-        return self.architecture == platform.machine()
 
-
-def should_compress_fs(args: Union[argparse.Namespace, MkosiArgs]) -> Union[bool, str]:
+def should_compress_fs(config: Union[argparse.Namespace, MkosiConfig]) -> Union[bool, str]:
     """True for the default compression, a string, or False.
 
     When explicitly configured with --compress-fs=, just return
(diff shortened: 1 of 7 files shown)


behrmann:
I have not seen *do not review* before. :D

rphibel:
> I have not seen _do not review_ before. :D

😁

behrmann:
Even though this should not be reviewed, I could not but have a quick look, since I've been waiting for this to drop and I might have a slightly different idea, since this is quite a bit churn and I could imagine us juggling whether to pass an args, conf or state in the future. 

How about instead of making a sort of matrioshka doll of dataclasses, doing this one with a bit of subclassing? This could look like this
```python
from dataclasses import dataclass, fields

@dataclass
class A:
    a: int = 1
    b: int = 2

@dataclass
class B:
    c: int = 3

@dataclass
class C(A, B):
    def extract_parent(self, name):
        # the first entry of the resolution order is the class itself, the last is object
        for cls in self.__class__.mro()[1:-1]:
            if cls.__name__ == name:
                d = {field.name: getattr(self, field.name) for field in fields(cls)}
                return cls(**d)
# > A()
# A(a=1, b=2)
# > B()
# B(c=3)
# > c = C()
# > c
# C(c=3, a=1, b=2)
# > c.extract_parent("A")
# A(a=1, b=2)
# > c.extract_parent("B")
# B(c=3)
```

This obviously has some edge cases, e.g. when A and B define the same fields, but this shouldn't be an issue for our use case, and it only works for dataclasses, since it uses some dataclass-specific stuff, but we have those. 

I haven't explored this further, but this came to my mind, when I had a quick look at your work, since it could the diff small, keep the (admitted) ergonomics of being able to grab everything out of the `MkosiArgs` bag, but would allow to slice and dice it for serializing and such.

Just an idea, might be totally harebrained, but I found it neat and wanted to share it.

rphibel:
> Even though this should not be reviewed, I could not but have a quick look, since I've been waiting for this to drop and I might have a slightly different idea, since this is quite a bit churn and I could imagine us juggling whether to pass an args, conf or state in the future.
> 
> How about instead of making a sort of matrioshka doll of dataclasses, doing this one with a bit of subclassing? This could look like this
> 
> ```python
> from dataclasses import dataclass, fields
> 
> @dataclass
> class A:
>     a: int = 1
>     b: int = 2
> 
> @dataclass
> class B:
>     c: int = 3
> 
> @dataclass
> class C(A, B):
>     def extract_parent(self, name):
>         # the first entry of the resolution order is the class itself, the last is object
>         for cls in self.__class__.mro()[1:-1]:
>             if cls.__name__ == name:
>                 d = {field.name: getattr(self, field.name) for field in fields(cls)}
>                 return cls(**d)
> # > A()
> # A(a=1, b=2)
> # > B()
> # B(c=3)
> # > c = C()
> # > c
> # C(c=3, a=1, b=2)
> # > c.extract_parent("A")
> # A(a=1, b=2)
> # > c.extract_parent("B")
> # B(c=3)
> ```
> 
> This obviously has some edge cases, e.g. when A and B define the same fields, but this shouldn't be an issue for our use case, and it only works for dataclasses, since it uses some dataclass-specific stuff, but we have those.
> 
> I haven't explored this further, but this came to my mind, when I had a quick look at your work, since it could the diff small, keep the (admitted) ergonomics of being able to grab everything out of the `MkosiArgs` bag, but would allow to slice and dice it for serializing and such.
> 
> Just an idea, might be totally harebrained, but I found it neat and wanted to share it.

Thanks for the suggestion! It seems like a nice way to be able to isolate config arguments while keeping the convenience of accessing everything from the `MkosiArgs` object. I will have a look.

DaanDeMeyer:
> How about instead of making a sort of matrioshka doll of dataclasses, doing this one with a bit of subclassing? This could look like this

I don't think we're at a point yet where we need something like this. We'll have a maximum of 2 structs that need to be passed around instead of the current single struct. That doesn't seem overly excessive to me. Sure we'll have some churn in this PR but afterwards it should be smooth sailing. So I'd like to keep things simple.

behrmann:
I'd argue that adding one small method on `MkosiArgs` is a bit simpler, than churning through everything once. ;) It would also make it easier to split the whole thing up some more, without additional churn.

Nevertheless, I'll review whatever this becomes.

mkosi/backend.py
```diff
@@ -571,19 +571,29 @@ class MkosiArgs:
...
+    """State related properties."""
+
+    original_umask: int
+
     output_checksum: Optional[Path] = None
     output_nspawn_settings: Optional[Path] = None
     output_sshkey: Optional[Path] = None
     output_root_hash_file: Optional[Path] = None
     output_root_hash_p7s_file: Optional[Path] = None
     output_bmap: Optional[Path] = None
```
DaanDeMeyer, line 596: So these don't directly map to an option but they are derived from an option, so I would consider them config as well.

mkosi/backend.py
```diff
@@ -585,6 +583,17 @@ class MkosiArgs:
...
+
+@dataclasses.dataclass
+class MkosiState:
+    """State related properties."""
+
+    original_umask: int
+
     cache_pre_inst: Optional[Path] = None
     cache_pre_dev: Optional[Path] = None
     output_signature: Optional[Path] = None
```
DaanDeMeyer, line 599: This one should also move to `MkosiConfig`

DaanDeMeyer (approved):

mkosi/backend.py
DaanDeMeyer: But can be done in a separate PR

PR #1149, opened 2022-08-24
tar/cpio: fix compression options handling

bluca:
Let users build uncompressed tarballs. Also take into account the
usual algorithm to determine the compression type for cpio.
Defaults are unchanged.


behrmann (approved):
LGTM, should be mentioned in the news, since I think this might have been asked for before.

bluca:
> LGTM, should be mentioned in the news, since I think this might have been asked for before.

added

behrmann (approved):
LGTM, merging.

Issue #1113, opened 2022-08-11
rpm compat symlink can result in circular links


olopf:
##	Description

Building a Fedora 36 image using mkosi 13 on Fedora 35 results in circular symlinks between `/var/lib/rpm` and `/usr/lib/sysimage/rpm`.

##	Steps to Reproduce

On Fedora 35:

```
mkosi --distribution=fedora --release=36 --format=directory --output=rootfs --package=dnf

‣ Detaching namespace
‣ Setting up package cache…
‣ Setting up package cache ~/sandpit/.mkosi-fzlcqnin complete
‣ Setting up temporary workspace.
‣ Temporary workspace set up in ~/sandpit/.mkosi-ry8jf8u9
‣ Running second (final) stage…
‣  Mounting image…
‣  Setting up basic OS tree…
‣  Mounting Package Cache
‣  Installing Fedora Linux…
‣   Mounting API VFS
Fedora 36 - base ...
...
Complete!
‣   Unmounting API VFS
‣  Unmounting Package Cache
‣  Creating compat symlink /var/lib/rpm → /usr/lib/sysimage/rpm
‣  Recording packages in manifest…
error: Unable to open sqlite database ~/sandpit/.mkosi-ry8jf8u9/root/var/lib/rpm/rpmdb.sqlite: unable to open database file
error: cannot open Packages index using sqlite - Operation not permitted (1)
error: cannot open Packages database in ~/sandpit/.mkosi-ry8jf8u9/root/var/lib/rpm
‣  (Unmounting image)
Traceback (most recent call last):
  File "/usr/lib64/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib64/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/usr/lib/python3.10/site-packages/mkosi/__main__.py", line 36, in <module>
    main()
  File "/usr/lib/python3.10/site-packages/mkosi/__main__.py", line 27, in main
    run_verb(a)
  File "/usr/lib/python3.10/site-packages/mkosi/__init__.py", line 8413, in run_verb
    manifest = build_stuff(args)
  File "/usr/lib/python3.10/site-packages/mkosi/__init__.py", line 7776, in build_stuff
    image = build_image(args, root, manifest=manifest, do_run_build_script=False, cleanup=True)
  File "/usr/lib/python3.10/site-packages/mkosi/__init__.py", line 7525, in build_image
[...]

PR #1150, opened 2022-08-26
fix: Move rpmdb content before replacing with symlink

olopf:
Patch for Issue #1113


labichn:
I also see issue #1113, and this patch fixes the issue. Can this be merged, @keszybz?

labichn:
> I tested this on both F37 (to build F36 and F37) and on F35 (to build F36).

8a73905 also works for me on an Arch host building F36 and F37, thanks @keszybz!

PR #1151, opened 2022-08-26
Implement support for generated files ownership management.

qdeslandes:
Implement support for generated files ownership management: files and folders created by `mkosi` will be owned by the calling user (`SUDO_UID` or `PKEXEC_UID`, if defined) instead of `root`.

This PR is composed of 2 commits.

### `chown` mkosi-generated directories

Use `chown` to change the owner of directories created by `mkosi`, unless `--no-chown` is used. Directories owner will be set to `SUDO_UID` or `PKEXEC_UID` if defined, current UID otherwise.

### Use nspawn's rootidmap option for bind mount

Directories bind mounted with `systemd-nspawn` will now use `rootidmap` option if `systemd-nspawn` v252 or later is used. This will ensure the owner of the files within the mounted directory inside the container will be the same as the owner of the directory on the backing filesystem.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -46,6 +46,12 @@
   `image.root.roothash`, `image.root.roothash.p7s` (same for `usr` variants).
 - `mkosi` will again default to the same OS release as the host system when the
   host system uses the same distribution as the image that's being built.
+- By default, `mkosi` will now change the owner of newly created directories to
+  `SUDO_UID` or `PKEXEC_UID` if defined, unless `--no-chown` is used.
+- If `systemd-nspawn` v252 or newer is used, bind-mounted directories with
+  `systemd-nspawn` will use the new `rootidmap` option so files and directories
+  created from within the container will be owned by the actual directory owner
+  on the host.
 
 ## v13
 
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3154,7 +3154,8 @@ def nspawn_params_for_build_sources(config: MkosiConfig, sft: SourceFileTransfer
         params += ["--setenv=SRCDIR=/root/src",
                    "--chdir=/root/src"]
         if sft == SourceFileTransfer.mount:
-            params += [f"--bind={config.build_sources}:/root/src"]
+            idmap_opt = ":rootidmap" if nspawn_version() >= 252 else ""
+            params += [f"--bind={config.build_sources}:/root/src{idmap_opt}"]
 
         if config.read_only:
             params += ["--overlay=+/root/src::/root/src"]
@@ -7370,6 +7371,8 @@ def run_build_script(state: MkosiState, raw: Optional[BinaryIO]) -> None:
     if state.config.build_script is None:
         return
 
+    idmap_opt = ":rootidmap" if nspawn_version() >= 252 else ""
+
     with complete_step("Running build script…"):
         os.makedirs(install_dir(state), mode=0o755, exist_ok=True)
 
@@ -7385,8 +7388,8 @@ def run_build_script(state: MkosiState, raw: Optional[BinaryIO]) -> None:
             "--as-pid2",
             "--link-journal=no",
             "--register=no",
-            f"--bind={install_dir(state)}:/root/dest",
-            f"--bind={state.var_tmp()}:/var/tmp",
+            f"--bind={install_dir(state)}:/root/dest{idmap_opt}",
+            f"--bind={state.var_tmp()}:/var/tmp{idmap_opt}",
             f"--setenv=WITH_DOCS={one_zero(state.config.with_docs)}",
             f"--setenv=WITH_TESTS={one_zero(state.config.with_tests)}",
             f"--setenv=WITH_NETWORK={with_network}",
@@ -7409,10 +7412,10 @@ def run_build_script(state: MkosiState, raw: Optional[BinaryIO]) -> None:
 
         if state.config.build_dir is not None:
             cmdline += ["--setenv=BUILDDIR=/root/build",
-                        f"--bind={state.config.build_dir}:/root/build"]
+                        f"--bind={state.config.build_dir}:/root/build{idmap_opt}"]
 
         if state.config.include_dir is not None:
-            cmdline += [f"--bind={state.config.include_dir}:/usr/include"]
+            cmdline += [f"--bind={state.config.include_dir}:/usr/include{idmap_opt}"]
 
         if state.config.with_network is True:
             # If we're using the host network namespace, use the same resolver
@@ -7421,7 +7424,7 @@ def run_build_script(state: MkosiState, raw: Optional[BinaryIO]) -> None:
             cmdline += ["--private-network"]
 
         if state.config.usr_only:
-            cmdline += [f"--bind={root_home(state)}:/root"]
+            cmdline += [f"--bind={root_home(state)}:/root{idmap_opt}"]
 
         if state.config.nspawn_keep_unit:
             cmdline += ["--keep-unit"]
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -79,10 +79,12 @@ from .backend import (
     PartitionTable,
     SourceFileTransfer,
     Verb,
+    chown_to_running_user,
     die,
     is_centos_variant,
     is_epel_variant,
     is_rpm_distribution,
+    mkdirp_chown_current_user,
     nspawn_knows_arg,
     nspawn_rlimit_params,
     nspawn_version,
@@ -4449,6 +4451,9 @@ def save_cache(state: MkosiState, raw: Optional[str], cache_path: Optional[Path]
             unlink_try_hard(cache_path)
             shutil.move(cast(str, state.root), cache_path)  # typing bug, .move() accepts Path
 
+    if not state.config.no_chown:
+        chown_to_running_user(cache_path)
+
 
 def _link_output(
         config: MkosiConfig,
@@ -4467,19 +4472,8 @@ def _link_output(
     if config.no_chown:
         return
 
-    sudo_uid = os.getenv("SUDO_UID")
-    sudo_gid = os.getenv("SUDO_GID")
-    if not (sudo_uid and sudo_gid):
-        return
-
     relpath = path_relative_to_cwd(newpath)
-
-    sudo_user = os.getenv("SUDO_USER", default=sudo_uid)
-    with complete_step(
-        f"Changing ownership of output file {relpath} to user {sudo_user} (acquired from sudo)…",
-        f"Changed ownership of {relpath}",
-    ):
-        os.chown(newpath, int(sudo_uid), int(sudo_gid))
+    chown_to_running_user(relpath)
 
 
 def link_output(state: MkosiState, artifact: Optional[BinaryIO]) -> None:
@@ -4681,8 +4675,7 @@ def setup_package_cache(config: MkosiConfig, workspace: Path) -> Path:
         cache = workspace / "cache"
     else:
         cache = config.cache_path
-
-    os.makedirs(cache, 0o755, exist_ok=True)
+        mkdirp_chown_current_user(cache, skip_chown=config.no_chown, mode=0o755)
 
     return cache
 
@@ -6252,7 +6245,7 @@ def find_output(args: argparse.Namespace) -> None:
     else:
         return
 
-    args.output_dir.mkdir(parents=True, exist_ok=True)
+    mkdirp_chown_current_user(args.output_dir, skip_chown=args.no_chown)
 
 
 def find_builddir(args: argparse.Namespace) -> None:
@@ -6265,7 +6258,7 @@ def find_builddir(args: argparse.Namespace) -> None:
     else:
         return
 
-    args.build_dir.mkdir(parents=True, exist_ok=True)
+    mkdirp_chown_current_user(args.build_dir, skip_chown=args.no_chown)
 
 
 def find_cache(args: argparse.Namespace) -> None:
@@ -6278,7 +6271,7 @@ def find_cache(args: argparse.Namespace) -> None:
     else:
         return
 
-    args.cache_path.mkdir(parents=True, exist_ok=True)
+    mkdirp_chown_current_user(args.cache_path, skip_chown=args.no_chown)
 
 
 def require_private_file(name: str, description: str) -> None:
@@ -7049,7 +7042,7 @@ def make_output_dir(config: MkosiConfig) -> None:
     if config.output_dir is None:
         return
 
-    config.output_dir.mkdir(mode=0o755, exist_ok=True)
+    mkdirp_chown_current_user(config.output_dir, skip_chown=config.no_chown, mode=0o755)
 
 
 def make_build_dir(config: MkosiConfig) -> None:
@@ -7057,7 +7050,7 @@ def make_build_dir(config: MkosiConfig) -> None:
     if config.build_dir is None:
         return
 
-    config.build_dir.mkdir(mode=0o755, exist_ok=True)
+    mkdirp_chown_current_user(config.build_dir, skip_chown=config.no_chown, mode=0o755)
 
 
 def configure_ssh(state: MkosiState, cached: bool) -> Optional[TextIO]:
--- a/mkosi/backend.py
+++ b/mkosi/backend.py
@@ -10,6 +10,7 @@ import errno
 import math
 import os
 import platform
+import pwd
 import resource
 import shlex
 import shutil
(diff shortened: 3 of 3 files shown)


behrmann:
I've only had a quick look for now, but some thoughts:
- Maybe I missed it, but I don't see a reasoning for making the mount executable configurable. Could you maybe explain that a bit? Since only the executable would be switched, it would still need to adhere to the CLI of regular `mount`, no?
- I'm not opposed to chowning the more persistent directories - this is something that people have asked for in the past -, but I'd rather not stuff the user running mkosi into the args. Stuff like `int(os.getenv("SUDO_UID") or os.getenv("PKEXEC_UID") or '0')` could just be a function or possibly include the chown call into some `chown_to_running_user`.
- The idmapping option for nspawn should check whether the nspawn version is recent enough, because most users will probably not run the most recent systemd releases.

qdeslandes:
> I've only had a quick look for now, but some thoughts:
> 
>     * Maybe I missed it, but I don't see a reasoning for making the mount executable configurable. Could you maybe explain that a bit? Since only the executable would be switched, it would still need to adhere to the CLI of regular `mount`, no?

Making mount executable configurable was useful to me during development, I thought I might as well include it in the PR. nspawn executable is configurable already (`MKOSI_NSPAWN_EXECUTABLE`), I assume the benefits are identical. 

>     * I'm not opposed to chowning the more persistent directories - this is something that people have asked for in the past -, but I'd rather not stuff the user running mkosi into the args. Stuff like `int(os.getenv("SUDO_UID") or os.getenv("PKEXEC_UID") or '0')` could just be a function or possibly include the chown call into some `chown_to_running_user`.

What you suggest would be to only be aware of the running user when `chown` is actually called, instead of storing it within `MkosiArgs`, is that what you mean?

>     * The idmapping option for nspawn should check whether the nspawn version is recent enough, because most users will probably not run the most recent systemd releases.

I agree, although `rootidmap` is not yet in any nspawn release, so I don't know yet which version number to check for. Ideally, we should check for `mount` version too, as not all versions support `X-mount.idmap`.

behrmann:
> Making mount executable configurable was useful to me during development, I thought I might as well include it in the PR. nspawn executable is configurable already (`MKOSI_NSPAWN_EXECUTABLE`), I assume the benefits are identical.

Ah, that was to have an escape hatch when working on nspawn itself. Yeah, sure. I'm ±0 on this change.

> What you suggest would be to only be aware of the running user when `chown` is actually called, instead of storing it within `MkosiArgs`, is that what you mean?

Yes, mkosi is either run by root or those environment variables are in the env during the whole of mkosi's run.

> I agree, although `rootidmap` is not yet in any nspawn release, so I don't know yet which version number to check for. Ideally, we should check for `mount` version too, as not all versions support `X-mount.idmap`.

I see, then this being merged is contingent on that being merged first and ending in a release.

qdeslandes:
> > What you suggest would be to only be aware of the running user when `chown` is actually called, instead of storing it within `MkosiArgs`, is that what you mean?
> 
> Yes, mkosi is either run by root or those environment variables are in the env during the whole of mkosi's run.

Sounds good to me, 

> > I agree, although `rootidmap` is not yet in any nspawn release, so I don't know yet which version number to check for. Ideally, we should check for `mount` version too, as not all versions support `X-mount.idmap`.
> 
> I see, then this being merged is contingent on that being merged first and ending in a release.

This PR is composed of 4 commits, the 3rd one requires https://github.com/systemd/systemd/pull/24467 and the 4th one requires https://github.com/util-linux/util-linux/pull/1661. The 1st and 2nd one can be merged.

qdeslandes:
Rebased on main branch, fixed the merge conflict and added a check in [Use rootidmap with nspawn's bind mounts](https://github.com/systemd/mkosi/pull/1151/commits/59d774c5af1c03ee2f0c15171ab0e6e1ab6fb16d) to ensure we use systemd-nspawn 252 or later.

Last commit ([Use idmapping for package cache bind mount](https://github.com/systemd/mkosi/pull/1151/commits/7b1dd0bfef241d342006f07d303482f813b1b6ad)) still needs to wait for `X-mount.idmap` support to be merged.

mkosi/backend.py
```diff
@@ -951,3 +957,22 @@ def complete_step(cls, text: str, text2: Optional[str] = None) -> Iterator[List[
 
         if text2 is not None:
             cls.print_step(text2.format(*args))
+
+
+def chown_to_running_user(paths: Union[PathString, [PathString]]):
```
DaanDeMeyer, line 962: Our python version is still locked to 3.7, so this should use `Sequence[PathString]` instead of `[PathString]`

mkosi/backend.py
```diff
@@ -951,3 +957,22 @@ def complete_step(cls, text: str, text2: Optional[str] = None) -> Iterator[List[
...
+    uid = int(os.getenv("SUDO_UID") or os.getenv("PKEXEC_UID") or '0')
+    user = pwd.getpwuid(uid).pw_name
+    gid = pwd.getpwuid(uid).pw_gid
+
+    # Root already owns generated files/directories, skipping.
+    if uid == 0:
+        return
```
DaanDeMeyer, line 969: I'd go for extra safety and still do the chowning even if it's root, you never know what the actual owner might be

mkosi/__init__.py
```diff
@@ -6900,6 +6918,8 @@ def make_output_dir(config: MkosiConfig) -> None:
         return
 
     config.output_dir.mkdir(mode=0o755, exist_ok=True)
+    if config.map_permissions:
+        chown_to_running_user(config.output_dir)
```
DaanDeMeyer, line 6922: Pretty sure we can do this unconditionally (here and elsewhere) regardless of whether the option is set or not no? There wouldn't be too much benefit when --map-permissions is not used but it doesn't hurt either right?

DaanDeMeyer (changes requested):
A few nits, but looks great overall!

Can we separate the mount stuff into another PR? That way we can merge the nspawn stuff already

mkosi/__init__.py
```diff
@@ -1453,13 +1455,19 @@ def mount_loop(config: MkosiConfig, dev: Path, where: Path, read_only: bool = Fa
     return mount(dev, where, options=options, read_only=read_only)
 
 
-def mount_bind(what: Path, where: Optional[Path] = None) -> ContextManager[Path]:
+def mount_bind(what: Path, where: Optional[Path] = None, with_idmapping: Bool = False) -> ContextManager[Path]:
```
DaanDeMeyer, line 1458: `Bool` needs to be lowercase

DaanDeMeyer:
Also, the MOUNT_EXECUTABLE stuff can be dropped, #1190 allows you to use --extra-search-paths instead to override the mount executable.

poettering:
instead of the mount binary stuff I think we should people towards using ExtraSearchPaths= instead. Or maybe add mkosi.path/ or so as special dir people can link files or dirs into that shall be preferred over the system ones.

mkosi/__init__.py
```diff
@@ -6108,6 +6110,8 @@ def find_output(args: argparse.Namespace) -> None:
         return
 
     args.output_dir.mkdir(parents=True, exist_ok=True)
+    if args.map_permissions:
+        chown_to_running_user([args.output_dir.parent, args.output_dir])
```
poettering, line 6120: shouldn't we chown this only when we created it? and if it already exists, leave it alone?

mkosi/__init__.py
```diff
@@ -6121,6 +6125,8 @@ def find_builddir(args: argparse.Namespace) -> None:
         return
 
     args.build_dir.mkdir(parents=True, exist_ok=True)
+    if args.map_permissions:
+        chown_to_running_user([args.build_dir.parent, args.build_dir])
```
poettering, line 6135: similar here

mkosi/__init__.py
```diff
@@ -6134,6 +6140,8 @@ def find_cache(args: argparse.Namespace) -> None:
         return
 
     args.cache_path.mkdir(parents=True, exist_ok=True)
+    if args.map_permissions:
+        chown_to_running_user([args.cache_path.parent, args.cache_path])
```
poettering, line 6150: and here, and everywhere else.

mkosi/backend.py
```diff
@@ -955,3 +957,22 @@ def complete_step(cls, text: str, text2: Optional[str] = None) -> Iterator[List[
...
+
+    if type(paths) is not list: paths = [ paths ]
+
+    for path in paths:
+        with MkosiPrinter.complete_step(
+            f"Changing ownership of output file {path} to user {user}…",
+            f"Changed ownership of {path}",
+        ):
+            os.chown(path, uid, gid)
```
poettering, line 978: i don't grok the relationship to --no-chown?

it appears to me that there should be only one option: either we do the chowning to the sudo/pkexec user or we do not. other than that we should probably always use the new `rootidmap` setting if nspawn supports it for everything we bind mount from the build tree.

to summarize:
1. if chowning is on, and nspawn supports rootidmap: always chown dirs we create, and use rootidmap when passing them to the container
2. otherwise let dirs we created owned by root, and do not use rootidmap
3. if chowning is on, chown output files

That would mean one option less, and we should retain compat with status quo ante mkosi trees.

mkosi/__init__.py
```diff
@@ -4379,19 +4383,8 @@ def _link_output(
...
-
     relpath = path_relative_to_cwd(newpath)
-
-    sudo_user = os.getenv("SUDO_USER", default=sudo_uid)
-    with complete_step(
-        f"Changing ownership of output file {relpath} to user {sudo_user} (acquired from sudo)…",
-        f"Changed ownership of {relpath}",
-    ):
-        os.chown(newpath, int(sudo_uid), int(sudo_gid))
+    chown_to_running_user(relpath)
```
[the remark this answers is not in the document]
qdeslandes: It's the same thing here: `_link_output()` will always `chown` to `SUDO_UID` (or `PKEXEC_UID`).

mkosi/backend.py
qdeslandes: The relation with `--no-chown` is the following:
- If `--no-chown` is used: `--map-permissions` can't be used, this is checked in `load_args()`).
- If `--no-chown` is **not** used and `--map-permission` is **not** used: no change from previous behaviour.
- If `--no-chown` is **not** used and `--map-permissions` is used: created directories are `chown`ed to `SUDO_UID` or `PKEXEC_UID`, and `rootidmap` is used. This way, this would preserve the existing behaviour.

Following your and Daan's comments:
- Remove `--map-permissions` and used `rootidmap` by default if nspawn supports it. This would be OK whatever the owner of the directory is, the worse thing is it will remap `root` to `root` if the owner of the `--bind` mounted directory is `root`.
- By default, `chown` all newly created build directories, except if `--no-chown` is used.

qdeslandes: Updated the PR's details with the latest changes and explanations.

mkosi/backend.py
```diff
@@ -948,3 +949,25 @@ def complete_step(cls, text: str, text2: Optional[str] = None) -> Iterator[List[
...
+    gid = pwd.getpwuid(uid).pw_gid
+
+    with MkosiPrinter.complete_step(
+        f"Changing ownership of output file {path} to user {user}…",
+        f"Changed ownership of {path}",
+    ):
+        os.chown(path, uid, gid)
+
+
+def mkdir_chown_current_user(path: Path, skip_chown: bool = False, mode: int = 0o777) -> None:
```
behrmann, line 966: The default 
```suggestion
def mkdir_chown_current_user(path: Path, skip_chown: bool = False, mode: int = 0o511) -> None:
```
The mode `0o777` is neither what was commonly used previously (`0o755`) nor the default of `Path.mkdir` (`0o511`). To not move to widely from pathlib I'd recommend that, so one doesn't have to think too much about semantics.

mkosi/backend.py
```diff
@@ -948,3 +949,25 @@ def complete_step(cls, text: str, text2: Optional[str] = None) -> Iterator[List[
...
+    if os.path.isdir(path):
+        return
+
+    path.mkdir(mode=mode)
```
behrmann, line 970: This basically makes `exists_ok=True` the one and only behaviour. I'd prefer adding the same argument (with `True` as default) and plumbing it through.

mkosi/backend.py
```diff
@@ -948,3 +949,25 @@ def complete_step(cls, text: str, text2: Optional[str] = None) -> Iterator[List[
...
+
+    if not skip_chown:
+        chown_to_running_user(path)
```
behrmann, line 973: This gives the function a very weird semantic, because it runs counter to the name. How about just naming the function `mkdir`, plumbing through all the options of `Path.mkdir` and giving it additional keyword arguments `owner=None` and `group=None` that will be passed to a `chown` at the end?

mkosi/backend.py
labichn, line 970: This function assumes that there's at most one missing parent directory: `sudo mkosi --build-dir=a/b/c` fails with a `FileNotFoundError`. Note that `parents=True` won't work, since any created parents will be owned by root. See [here](https://github.com/systemd/mkosi/compare/main...labichn:mkosi:mkdirp-chown) for one way to fix that (feel free to copy anything useful).

mkosi/backend.py
qdeslandes: [`pathlib.mkdir`'s default is `0o777`](https://docs.python.org/3.9/library/pathlib.html#pathlib.Path.mkdir)?

mkosi/backend.py
qdeslandes: Lennart (I think?) suggested scoping `SUDO_UID` and `PKEXEC_UID` into a specific function, in this case it would be `chown_to_running_user()`, so having `owner=` and `group=` argument would go against that. I used @labichn's code to create and chown missing directories recursively and added `if skip_chown: continue` to skip the call to `chown_to_running_user()`. This makes the code flow a bit better.

mkosi/backend.py
behrmann: My bad... `0o777` is 511 and I only had a quick look at the help output in my shell and didn't think twice.

mkosi/__init__.py
```diff
@@ -6251,7 +6244,8 @@ def find_output(args: argparse.Namespace) -> None:
     else:
         return
 
-    args.output_dir.mkdir(parents=True, exist_ok=True)
+    mkdirp_chown_current_user(args.output_dir.parent, args.no_chown)
+    mkdirp_chown_current_user(args.output_dir, args.no_chown)
```
DaanDeMeyer, line 6254: Why do we need both here if the function already takes care of creating the parent directories?

mkosi/backend.py
```diff
@@ -948,3 +949,32 @@ def complete_step(cls, text: str, text2: Optional[str] = None) -> Iterator[List[
...
+    gid = pwd.getpwuid(uid).pw_gid
+
+    with MkosiPrinter.complete_step(
+        f"Changing ownership of output file {path} to user {user}…",
+        f"Changed ownership of {path}",
+    ):
+        os.chown(path, uid, gid)
+
+
+def mkdirp_chown_current_user(path: PathString, skip_chown: bool = False, mode: int = 0o777, exist_ok: bool = True) -> None:
```
behrmann, line 966:
```suggestion
def mkdirp_chown_current_user(path: PathString, *, skip_chown: bool = False, mode: int = 0o777, exist_ok: bool = True) -> None:
```

mkosi/__init__.py
```diff
@@ -6264,7 +6258,8 @@ def find_builddir(args: argparse.Namespace) -> None:
     else:
         return
 
-    args.build_dir.mkdir(parents=True, exist_ok=True)
+    mkdirp_chown_current_user(args.build_dir.parent, args.no_chown)
+    mkdirp_chown_current_user(args.build_dir, args.no_chown)
```
DaanDeMeyer, line 6268: ditto

mkosi/__init__.py
```diff
@@ -6277,7 +6272,8 @@ def find_cache(args: argparse.Namespace) -> None:
     else:
         return
 
-    args.cache_path.mkdir(parents=True, exist_ok=True)
+    mkdirp_chown_current_user(args.cache_path.parent, args.no_chown)
+    mkdirp_chown_current_user(args.cache_path, args.no_chown)
```
DaanDeMeyer, line 6282: ditto

DaanDeMeyer (commented):
`nspawn_params_for_build_sources()` needs to be modified as well to add `rootidmap`.

Also please move the last commit to a new PR so that we can delay merging it until the util-linux option gets merged.

mkosi/backend.py
behrmann: I can live with the structure of the code, but the name just makes what the function does non-obvious, because the name doesn't really imply that the `chown` would be optional, but I'm a bit hard-pressed to come up with a better name at the moment.

NEWS.md
```diff
@@ -46,6 +46,12 @@
   `image.root.roothash`, `image.root.roothash.p7s` (same for `usr` variants).
 - `mkosi` will again default to the same OS release as the host system when the
   host system uses the same distribution as the image that's being built.
+- By default, `mkosi` will now change the owner of newly created directories to
+  `SUDO_UID` or `PKEXEC_UID` if defined, unless `--no-chown` is used.
+- If `systemd-nspawn` v252 or newer is used, bind-mounted directories with
+  `systemd-nspawn` will use the new `rootidmap` option so files and directories
+  created from within the virtual machine (or the container) will be owned by
```
DaanDeMeyer, line 53: Just `container` here, we only do this for systemd-nspawn, for for qemu.

DaanDeMeyer (changes requested):

PR #1152, opened 2022-08-27
Drop CentOS 7 and BIOS/grub support

DaanDeMeyer:
See the individual commit messages. The goal is to align on EFI systems and systemd-boot as the only supported bootloaders for mkosi generated bootable images as discussed in the last biweekly.

cc @bluca @poettering @keszybz

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,8 +8,6 @@ for more information.
 for more information.
 - The Arch kernel/bootloader pacman hooks were removed. For anyone that still
 wants to use them, they can be found [here](https://github.com/systemd/mkosi/tree/v13/mkosi/resources/arch).
-When building a bios image, /boot/vmlinuz-kver and /boot/initramfs-kver.img are
-now symlinks to the actual files as installed by kernel-install.
 - mkosi now creates distro~release subdirectories inside the build, cache and output
 directories for each distro~release combination that is built. This allows building
 for multiple distros without throwing away the results of a previous distro build every
@@ -23,6 +21,9 @@ been removed from the docs but are still supported for backwards compatibility.
 which is `SigLevel = Required DatabaseOptional`. If this results in keyring errors,
 you need to update the keyring by running `pacman-key --populate archlinux`.
 - Support for CentOS 7 was dropped.
+- Support for BIOS/grub was dropped. To allow users to configure grub themselves, the
+new `--bios-size` option can be used to add a BIOS boot partition of the specified size
+on which grub can be installed.
 
 ## v13
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -22,6 +22,7 @@ been removed from the docs but are still supported for backwards compatibility.
 - Pacman's `SigLevel` option was changed to use the same default value as used on Arch
 which is `SigLevel = Required DatabaseOptional`. If this results in keyring errors,
 you need to update the keyring by running `pacman-key --populate archlinux`.
+- Support for CentOS 7 was dropped.
 
 ## v13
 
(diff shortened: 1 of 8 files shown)


behrmann:
Dropping CentOS 7 doesn't worry me, but I'm not quite sold on dropping Bios support. Servers without EFI support have been sold in the not too distant past - at least I still have to handle such hardware often enough.

Mind you, I'm not necessarily against dropping grub, but for those people stuck with having to deal with Bios-only hardware it would be nice to keep around some bare bones support, e.g only creating the `/boot/` partition, but leaving the user to do the rest via scripts.

DaanDeMeyer:
> Mind you, I'm not necessarily against dropping grub, but for those people stuck with having to deal with Bios-only hardware it would be nice to keep around some bare bones support, e.g only creating the /boot/ partition, but leaving the user to do the rest via scripts.

Makes sense, I'll add a `bios-size` option that configures a bios boot partition if set.

bluca (approved):

keszybz:
```
=================================== FAILURES ===================================
______________________________ test_centos_brtfs _______________________________

    def test_centos_brtfs() -> None:
        with cd_temp_dir():
            config = Path("mkosi.conf")
            for dist in (Distribution.centos, Distribution.centos_epel):
                for release in range (2, 9):
                    config.write_text(
                        textwrap.dedent(
                            f"""
                            [Distribution]
                            Distribution={dist}
                            Release={release}
                            [Output]
                            Format=gpt_btrfs
                            """
                        )
                    )
                    with pytest.raises(MkosiException, match=".CentOS.*btrfs"):
                        parse([])
    
        with cd_temp_dir():
            config = Path("mkosi.conf")
            for dist in (Distribution.centos, Distribution.centos_epel):
                for release in range (2, 8):
                    config.write_text(
                        textwrap.dedent(
                            f"""
                            [Distribution]
                            Distribution={Distribution.centos}
                            Release={release}
                            Bootable=yes
                            """
                        )
                    )
>                   with pytest.raises(MkosiException, match=".CentOS.*unified.*kernel"):
E                   Failed: DID NOT RAISE <class 'mkosi.backend.MkosiException'>

tests/test_parse_load_args.py:138: Failed
```

LGTM otherwise.

DaanDeMeyer:
Went a bit further and dropped the --boot-protocols option as well. We just unconditionally extract kernel, initrd, and cmdline from the image now so that they can be used with qemu direct linux boot if needed. I kept the parser for boot protocols intact (but hidden, with a deprecation warning) to avoid breaking users that aren't using bios.

NEWS.md
```diff
@@ -22,6 +20,10 @@ been removed from the docs but are still supported for backwards compatibility.
 - Pacman's `SigLevel` option was changed to use the same default value as used on Arch
 which is `SigLevel = Required DatabaseOptional`. If this results in keyring errors,
 you need to update the keyring by running `pacman-key --populate archlinux`.
+- Support for CentOS 7 was dropped.
+- Support for BIOS/grub was dropped. To allow users to configure grub themselves, the
```
behrmann, line 24:
```suggestion
- Support for BIOS/grub was dropped, because EFI hardware is widely available and legacy BIOS systems do not support the feature set to fully verify a boot chain from firmware to userland and it has become bothersome to maintain for little use. To generate BIOS images you can use any version of mkosi up to mkosi 13 or the 
```

NEWS.md
```diff
@@ -22,6 +20,10 @@ been removed from the docs but are still supported for backwards compatibility.
...
+new `--bios-size` option can be used to add a BIOS boot partition of the specified size
+on which grub can be installed.
```
behrmann, line 26:
```suggestion
on which grub (or any other bootloader) can be installed with the help of mkosi's script support.
```

behrmann: Maybe the unconditional copying out of the kernel and initrd should be mentioned as well, since for most users this will be new output.

behrmann (approved):

behrmann:
Ah, a tad too late because of lunch :D Will make this a PR.

PR #1159, opened 2022-08-29
nspawn: use --link-journal=no

bluca:
The host might not have a journal.

On a minimal qemu build image on OBS, building SUSE Tumbleweed,
linking the journal fails with:

  Failed to retrieve machine ID: No medium found

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -7307,6 +7307,7 @@ def run_build_script(args: MkosiArgs, root: Path, raw: Optional[BinaryIO]) -> No
             f"--uuid={args.machine_id}",
             f"--machine=mkosi-{uuid.uuid4().hex}",
             "--as-pid2",
+            "--link-journal=no",
             "--register=no",
             f"--bind={install_dir(args, root)}:/root/dest",
             f"--bind={var_tmp(root)}:/var/tmp",
--- a/mkosi/backend.py
+++ b/mkosi/backend.py
@@ -686,6 +686,7 @@ def run_workspace_command(
         "--uuid=" + args.machine_id,
         "--machine=mkosi-" + uuid.uuid4().hex,
         "--as-pid2",
+        "--link-journal=no",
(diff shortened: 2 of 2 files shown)


behrmann:
I accept that the problem is in nspawn's `setup_journal`, since that exits pretty early on when the journals are not supposed to be linked, but looking at the nspawn docs, not having a (persistent) journal should be a noop?

The error you mention is from nspawn's
```c
r = sd_id128_get_machine(&this_id);
if (r < 0)
        return log_error_errno(r, "Failed to retrieve machine ID: %m");
```
so the problem is (I assume) the build machine not having a machine ID?

bluca:
> I accept that the problem is in nspawn's `setup_journal`, since that exits pretty early on when the journals are not supposed to be linked, but looking at the nspawn docs, not having a (persistent) journal should be a noop?
> 
> The error you mention is from nspawn's
> 
> ```c
> r = sd_id128_get_machine(&this_id);
> if (r < 0)
>         return log_error_errno(r, "Failed to retrieve machine ID: %m");
> ```
> 
> so the problem is (I assume) the build machine not having a machine ID?

Could be - I have little control over the build machine, in the production system

behrmann (approved):
Just wanted to make sure the right culprit is blamed. Since we obviously can fall back to not linking the journal this would maybe warrant upstreaming into nspawn: log an error, but continue.

DaanDeMeyer (changes requested):
We call nspawn in a few other places as well, such as when running the build script, we should disable journal linking there as well. I'd leave run_shell_cmdline() untouched though

bluca:
> We call nspawn in a few other places as well, such as when running the build script, we should disable journal linking there as well. I'd leave run_shell_cmdline() untouched though

Added to the build script run too. The only other one I can find is the `shell` one, which I have not changed as requested.

DaanDeMeyer (approved):

PR #1170, opened 2022-09-05
Don't pass environment when running kernel-install

DaanDeMeyer:
Initially introduced in 32d09033e94f36993bede7016948e0702a4cb185,
but the reason wasn't recorded. Generally, we don't want to pass
just any environment variables to kernel-install, so let's not pass
the entire environment when running it.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3159,8 +3159,6 @@ def nspawn_params_for_build_sources(config: MkosiConfig, sft: SourceFileTransfer
     else:
         params += ["--chdir=/root"]
 
-    params += [f"--setenv={env}={value}" for env, value in config.environment.items()]
-
     return params
 
 
@@ -3182,7 +3180,8 @@ def run_prepare_script(config: MkosiConfig, state: MkosiState, cached: bool) -> 
         shutil.copy2(config.prepare_script, root_home(config, state.root) / "prepare")
 
         nspawn_params = nspawn_params_for_build_sources(config, SourceFileTransfer.mount)
-        run_workspace_command(config, state.root, ["/root/prepare", verb], network=True, nspawn_params=nspawn_params)
+        run_workspace_command(config, state.root, ["/root/prepare", verb],
+                              network=True, nspawn_params=nspawn_params, env=config.environment)
 
         srcdir = root_home(config, state.root) / "src"
         if srcdir.exists():
@@ -3211,8 +3210,7 @@ def run_postinst_script(
         shutil.copy2(config.postinst_script, root_home(config, state.root) / "postinst")
 
         run_workspace_command(config, state.root, ["/root/postinst", verb],
-                              network=(config.with_network is True),
-                              env=config.environment)
+                              network=(config.with_network is True), env=config.environment)
         root_home(config, state.root).joinpath("postinst").unlink()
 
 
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
Added one more commit to normalize the environment used across script runs

mkosi/__init__.py
```diff
@@ -3229,10 +3227,8 @@ def run_finalize_script(config: MkosiConfig, state: MkosiState, for_cache: bool)
     verb = "build" if state.do_run_build_script else "final"
 
     with complete_step("Running finalize script…"):
-        env = collections.ChainMap(dict(BUILDROOT=str(state.root), OUTPUTDIR=str(output_dir(config))),
-                                   config.environment,
-                                   os.environ)
-        run([config.finalize_script, verb], env=env)
+        run([config.finalize_script, verb],
+            env=dict(BUILDROOT=str(state.root), OUTPUTDIR=str(output_dir(config), **config.environment)))
```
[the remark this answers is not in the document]
DaanDeMeyer: Nice catch!

mkosi/__init__.py
```diff
@@ -3229,10 +3227,8 @@ def run_finalize_script(config: MkosiConfig, state: MkosiState, for_cache: bool)
     verb = "build" if state.do_run_build_script else "final"
 
     with complete_step("Running finalize script…"):
-        env = collections.ChainMap(dict(BUILDROOT=str(state.root), OUTPUTDIR=str(output_dir(config))),
-                                   config.environment,
-                                   os.environ)
-        run([config.finalize_script, verb], env=env)
+        run([config.finalize_script, verb],
+            env=dict(BUILDROOT=str(state.root), OUTPUTDIR=str(output_dir(config)), **config.environment))
```
behrmann, line 3231: Thinking some more about this, I think this version is slightly better.
```suggestion
            env={**config.environment), "BUILDROOT": str(state.root), "OUTPUTDIR": str(output_dir(config))} 
```

This couldn't happen before (because of the `ChainMap`), but now if there's already a `BUILDROOT` or `OUTPUTDIR` in the `config.environment` this would throw a `TypeError` because the same keyword argument is passed multiple times. 

Doing it in dict literal is slightly more efficient (because `dict` could be overwritten by a custom function, while the literal can't, so less lookups) and it doesn't have this problem, because the last value for a key wins, but then we need to unpack the `config.environment` first, so that our values win.

mkosi/__init__.py
```diff
@@ -3229,10 +3227,8 @@ def run_finalize_script(config: MkosiConfig, state: MkosiState, for_cache: bool)
     verb = "build" if state.do_run_build_script else "final"
 
     with complete_step("Running finalize script…"):
-        env = collections.ChainMap(dict(BUILDROOT=str(state.root), OUTPUTDIR=str(output_dir(config))),
-                                   config.environment,
-                                   os.environ)
-        run([config.finalize_script, verb], env=env)
+        run([config.finalize_script, verb],
+            env={**config.environment, "BUILDROOT": str(state.root), "OUTPUTDIR": str(output_dir(config))})
```
mcassaniti, line 3231: I just noticed that `os.environ` is missing from here. Is that intentional?

DaanDeMeyer: Yes it's added automatically by run()

DaanDeMeyer:
This one should be good to go as it already got reviewed multiple times and all comments have been addressed

PR #1173, opened 2022-09-05
[gentoo] preper for usrmerge

257:
upstream has added merged-usr profile: https://bugs.gentoo.org/843818#c2
this makes that profile default for us. I expect a stage3 for merged-usr
sometime soon which should result in signicant code-drop at our end.

Signed-off-by: Paymon MARANDI <darwinskernel@gmail.com>

### DIFF
--- a/mkosi/distributions/gentoo.py
+++ b/mkosi/distributions/gentoo.py
@@ -13,7 +13,6 @@ from typing import Dict, Generator, List, Sequence
 
 from mkosi.backend import (
     ARG_DEBUG,
-    MkosiConfig,
     MkosiException,
     MkosiPrinter,
     MkosiState,
@@ -44,23 +43,16 @@ def flock_path(path: Path) -> Generator[int, None, None]:
 
 class Gentoo:
     arch_profile: Path
-    baselayout_use: Path
-    DEFAULT_NSPAWN_PARAMS: List[str]
-    emerge_default_opts: List[str]
     arch: str
+    custom_profile_path: Path
+    DEFAULT_NSPAWN_PARAMS: List[str]
+    ebuild_sh_env_dir: Path
+    emerge_default_opts: List[str]
     emerge_vars: Dict[str, str]
-    # sys-boot and sys-kernel mainly for boot
-    pkgs_boot: List[str]
-    # @system set (https://wiki.gentoo.org/wiki/System_set_(Portage))
-    pkgs_sys: List[str]
-    # filesystem packages (dosfstools, btrfs, squashfs, etc)
-    pkgs_fs: List[str]
-    UNINSTALL_IGNORE: List[str]
-    root: Path
     portage_cfg_dir: Path
     profile_path: Path
-    custom_profile_path: Path
-    ebuild_sh_env_dir: Path
+    root: Path
+    pkgs: Dict[str, List[str]] = {}
     dracut_atom = "sys-kernel/dracut"
 
     EMERGE_UPDATE_OPTS = [
@@ -73,19 +65,10 @@ class Gentoo:
         "--complete-graph-if-new-use=y",
     ]
 
-    UNINSTALL_IGNORE = ["/bin", "/sbin", "/lib", "/lib64"]
-
     portage_use_flags = [
-        "systemd",  # 'systemd' is a dependancy
         "initramfs",
-        "git",  # 'git' for sync-type=git
-        "symlink",  # 'symlink' for kernel
-        "sdl",
-        "-filecaps",
-        "-savedconfig",
-        "-split-bin",
-        "-split-sbin",
-        "-split-usr",
+        "git",  # for sync-type=git
+        "symlink",  # for kernel
     ]
 
     # TODO: portage_features.add("ccache"), this shall expedite the builds
@@ -96,7 +79,7 @@ class Gentoo:
         "-usersync",
         "-usersandbox",
         "-sandbox",
-        "-pid-sandbox",  # -pid-sandbox is required for cross-compile scenarios
+        "-pid-sandbox",  # for cross-compile scenarios
         "-network-sandbox",
         "parallel-install",
         "buildpkg",
@@ -140,7 +123,9 @@ class Gentoo:
 
         sudo python setup.py install
 
-        sudo ln -s --relative /var/db/repos/gentoo/profiles/default/linux/amd64/17.1/no-multilib/systemd /etc/portage/make.profile
+        sudo ln -s --relative \
+            /var/db/repos/gentoo/profiles/default/linux/amd64/17.1/no-multilib/systemd/merged-usr \
+            /etc/portage/make.profile
         """
         try:
             from portage.const import (  # type: ignore
@@ -159,19 +144,21 @@ class Gentoo:
                     ebuild_sh_env_dir=EBUILD_SH_ENV_DIR,
                     portage_cfg_dir=USER_CONFIG_PATH)
 
-    def __init__(
-        self,
-        state: MkosiState,
-    ) -> None:
-
+    @complete_step("Installing Gentoo…")
+    def __init__(self, state: MkosiState) -> None:
+        # TOCLEANUP: legacy namig, to be cleaned up
+        self.state = state
+        self.config = self.state.config
+        self.root = self.state.root
         ret = self.try_import_portage()
 
         from portage.package.ebuild.config import config as portage_cfg  # type: ignore
 
-        self.portage_cfg = portage_cfg(config_root=str(state.root), target_root=str(state.root),
-                                  sysroot=str(state.root), eprefix=None)
+        self.portage_cfg = portage_cfg(config_root=str(state.root),
+                                       target_root=str(state.root),
+                                       sysroot=str(state.root), eprefix=None)
 
-        PORTAGE_MISCONFIGURED_MSG = "You have portage(5) installed but it's probably missing defaults, bailing out"
+        PORTAGE_MISCONFIGURED_MSG = "Missing defaults for portage, bailing out"
         # we check for PORTDIR, but we could check for any other one
         if self.portage_cfg['PORTDIR'] is None:
             die(PORTAGE_MISCONFIGURED_MSG)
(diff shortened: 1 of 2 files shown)


DaanDeMeyer:
What's the status here?

257:
i will rework it this weekend.

257:
> What's the status here?

actually it passes the test. alert are outdated/fixed. let me know if you see otherwise.

DaanDeMeyer:
Any idea when we'll be getting the merged-usr stage3?

257:
soon i hope. i'm hoping that gentoo-releng's uses `mkosi` for at least generating the initial stage3s.
pinging @floppym

257:
> Any idea when we'll be getting the merged-usr stage3?

_by the EOY_, supposidly.

DaanDeMeyer:
There's still a CI failure in the gentoo gpt ext4 CI run, so I don't think we can merge this just yet

257:
i will have a look over the weekend. yes, please hold off of merging this.

257:
> There's still a CI failure in the gentoo gpt ext4 CI run, so I don't think we can merge this just yet

possibly fails on `sys-kernel/gentoo-kernel-bin`...

builds here, stage3 images upstream are *autobuilt*. i even thought of freezing the build that works (today it's `stage3-amd64-nomultilib-systemd-20220911T170535Z.tar.xz` and `=sys-kernel/gentoo-kernel-bin-5.15.69`) but it's pointless because we don't have a guarantee from upstream that image would even be there at the time of next `mkosi` release. 

my understanding is that there is an effort upstream to stanardise the way these stage3 release work. 

i understand this breaks `mkosi`'s promise of 100% reproducibility but as you you can options are limited.

what do you think?

257:
> There's still a CI failure in the gentoo gpt ext4 CI run, so I don't think we can merge this just yet

hmm, actually [image boots](https://github.com/systemd/mkosi/actions/runs/3070063742/jobs/5098459142#step:9:995) [we fail on ssh](https://github.com/systemd/mkosi/actions/runs/3070063742/jobs/5098459142#step:9:1027) odd!

257:
seems to be a bug with portage and binary packages, fixed upstream. how do i force the actions to run again without pushing to the branch?

DaanDeMeyer:
I restarted the failed job

DaanDeMeyer:
@257 Can you take another look, something is borked with ssh

257:
> @257 Can you take another look, something is borked with ssh

i did while back i couldn't find a way to run the test locally. any pointers?

257:
> @257 Can you take another look, something is borked with ssh

ssh is fixed but now squash tests fail! don't merge; i will have another look.

257:
> > @257 Can you take another look, something is borked with ssh
> 
> ssh is fixed but now squash tests fail! don't merge; i will have another look.



> > @257 Can you take another look, something is borked with ssh
> 
> ssh is fixed but now squash tests fail! don't merge; i will have another look.

seems to be related to https://github.com/systemd/systemd/commit/a8040b6d0ad43e5cb7168fd3b739c8c68aaa61b1
@keszybz could you have a look? it only happens on plain_squash, systemd-251 both on github and my test environment.

257:
https://lore.kernel.org/lkml/20191130181548.GA28459@gentoo-tp.home/

behrmann:
@257 can you rebase against main? `gentoo.py` has moved to `distributions/gentoo.py`. The rest should be pretty much intact (except for the interface talking to the rest of mkosi having moved into a subclass of `DistributionInstaller`).

257:
[disabled plain_squashfs supported for the time being](https://github.com/systemd/mkosi/actions/runs/3575374955/jobs/6011854813#step:7:51)

mkosi/distributions/gentoo.py
```diff
@@ -563,20 +441,4 @@ def kernel_image(name: str, architecture: str) -> Path:
...
-            # pkgs_boot. Using Gentoo.EMERGE_UPDATE_OPTS for opts allows the
-            # package manager to uninstall blockers.
-            gentoo.invoke_emerge(state, pkgs=gentoo.pkgs_boot, opts=Gentoo.EMERGE_UPDATE_OPTS)
-
-        if state.config.packages:
-            gentoo.invoke_emerge(state, pkgs=state.config.packages)
-
-        if state.do_run_build_script:
-            gentoo.invoke_emerge(state, pkgs=state.config.build_packages)
+        Gentoo(state)
```
behrmann, line 444: It looks like you're doing all the work in the `__init__` method of this class. At this point it doesn't need to be a class anymore. You could just move all of it into this method, no?

257: one at the time :) moving things around like that is yet another big change.

257: is this necessary, `fedora` seems to be doing the same thing, albeit not inside a class.

DaanDeMeyer:
@257 Can you rebase? plain_squashfs just got removed by #1276 so we don't need to disable it anymore

257:
np, working on it

257:
@DaanDeMeyer which version of `systemd-repart` comes with `--defer-partitions`? i'm hitting this here:
```
systemd-repart: unrecognized option '--defer-partitions'
```
```
systemd 252 (252)
+PAM -AUDIT -SELINUX -APPARMOR +IMA +SMACK +SECCOMP +GCRYPT -GNUTLS +OPENSSL +ACL +BLKID +CURL +ELFUTILS -FIDO2 +IDN2 -IDN -IPTC +KMOD +LIBCRYPTSETUP +LIBFDISK +PCRE2 +PWQUALITY -P11KIT +QRENCODE +TPM2 -BZIP2 +LZ4 -XZ -ZLIB +ZSTD -BPF_FRAMEWORK +XKBCOMMON +UTMP +SYSVINIT default-hierarchy=unified
```

behrmann:
@257 see #1290

257:
> np, working on it

ready with possible exception of ssh. please review.

257:
> ready with possible exception of ssh. please review.
ssh is also fixed. probably a side-effect of usrmerge.

257:
not sure why centos is failing

DaanDeMeyer:
> not sure why centos is failing

Unrelated to this PR

mkosi/distributions/gentoo.py
```diff
@@ -196,67 +184,53 @@ def __init__(
...
+                                 f"{state.config.release}/no-multilib/systemd/merged-usr")
+        self.pkgs['sys'] = ["@world"]
+
+        self.pkgs['fs'] = ["sys-fs/dosfstools"]
+        if state.config.output_format == OutputFormat.subvolume:
+            self.pkgs['fs'] += ["sys-fs/btrfs-progs"]
+        elif state.config.output_format == OutputFormat.disk:
+            self.pkgs['fs'] += ["sys-fs/btrfs-progs"]
+            self.pkgs['fs'] += ["sys-fs/xfsprogs"]
+            self.pkgs['fs'] += ["sys-fs/e2fsprogs"]
```
DaanDeMeyer, line 206: We should drop this, we don't install any fs tools by default for other distros anymore either

mkosi/distributions/gentoo.py
```diff
@@ -449,16 +341,39 @@ def mkosi_conf(self) -> None:
...
+
+    def finishup(self) -> None:
+        self.invoke_emerge(pkgs=self.pkgs['fs'])
+
+        if not self.state.do_run_build_script:
+            if self.state.config.ssh:
+                MkosiPrinter.print_step("enabling sshd")
+                cmd = ["/usr/bin/systemctl", "enable", "sshd"]
+                run_workspace_command(self.state, cmd, nspawn_params=self.DEFAULT_NSPAWN_PARAMS,
+                                      check=True)
```
DaanDeMeyer, line 367: This should be done with host systemctl instead of image systemctl, so use `run()` instead of `run_workspace_command()`

DaanDeMeyer: Also, why do this here and not in the shared code we have for this?

mkosi/distributions/gentoo.py
```diff
@@ -481,83 +396,30 @@ def invoke_emerge(
...
-        # that point.
-        self.baselayout_use.unlink()
-
-    def depclean(self, state: MkosiState) -> None:
-        self.invoke_emerge(state, actions=["--depclean"])
+            MkosiPrinter.print_step("Invoking emerge(1) inside stage3"
+                                    f"{self.root}")
+            run_workspace_command(self.state, cmd, network=True, env=self.emerge_vars,
+                                  nspawn_params=self.DEFAULT_NSPAWN_PARAMS,
+                                  check=check)
```
DaanDeMeyer, line 403: Why does this have to run inside the image? We're trying to move all the package manager stuff to run outside of the image instead of inside the image

DaanDeMeyer (changes requested):
Looks good, a few comments though

mkosi/distributions/gentoo.py
257: even `btrfs-progs` and `dosfstools`?

DaanDeMeyer: Yup, because with systemd-repart, we don't know what filesystems will be in the final image, so we don't add anything by default, and leave it up to the user.

mkosi/distributions/gentoo.py
257: we ran into number of problems using host's `portage` in the past. only initial `sync_portage_tree()` uses host's `portage` and i'm tempted to even remove that.

also we still do make some changes related to `systemd` and we also install kernel related packages for bootable cases, see `pkgs['boot']` especially under `update_stage3()`.
also any extra packages asked by user also needs this, see `merge_user_pkgs()`.

mkosi/distributions/gentoo.py
257: we do if output is subvolume! unless i'm misunderstood what that means.

257: removing `dostfstools`

257: removed `btrfs-progs` too but i would be interested to know more about the case  `output == subvolume`.

mkosi/distributions/gentoo.py
257: don't remember why. i had a bit in `mkosi/__init__.py` related to this which got removed after switching to usrmerge. removed it to see what broke in the past, although it seems to me that CI doesn't have a test for ssh?

mkosi/distributions/gentoo.py
DaanDeMeyer: In case of `output == subvolume`, root directory is indeed a subvolume but that doesn't mean that the user won't pack up the subvolume inside an ext4 filesystem or something else later, just because the root directory is a subvolume does not mean that the user will want to have btrfs-progs available inside the root directory

mkosi/distributions/gentoo.py
DaanDeMeyer: I would prefer if we use host's portage if possible. We do the same for all other distributions. But we can leave that for another PR since this one is already been open for a long time.

mkosi/distributions/gentoo.py
257: ok! but now i would like to meet that user, lol.
well actually i just don't know a use case like that but i get that we need to stay as generic as possible.

DaanDeMeyer (approved):

mkosi/distributions/gentoo.py
257: i started out using host's `portage` and its python interface in face, which turned out to be horrible. even @poettering mentioned that we probably do the same for `dnf`...

yeah changing to host's would be a big big change with lots of test. although with usermerge in place we're dropping lots of code and complexity. i will have a go at it once day job gets a bit lighter.

mkosi/distributions/gentoo.py
```diff
@@ -268,9 +233,8 @@ def fetch_fix_stage3(self, state: MkosiState, root: Path) -> None:
         # and more... so we can gladly escape all this hideousness!
         ###########################################################
         with urllib.request.urlopen(stage3tsf_path_url) as r:
-            args_profile = "nomultilib"
-            # 20210711T170538Z/stage3-amd64-nomultilib-systemd-20210711T170538Z.tar.xz 214470580
-            regexp = f"^[0-9TZ]+/stage3-{self.arch}-{args_profile}-systemd-[0-9TZ]+[.]tar[.]xz"
+            # e.g.: stage3-amd64-nomultilib-systemd-mergedusr-20221127T170156Z.tar
+            regexp = f"^[0-9TZ]+[/]stage3[-]{self.arch}[-]nomultilib[-]systemd[-]mergedusr[-[0-9TZ]+[.]tar[.]xz"
```
behrmann, line 237: The parentheses are not balanced (why escape the dashes anyway?) and the regex doesn't match the example above, since the initial `[0-9TZ]+/` is and the `.xz` at the end are missing, how about
```suggestion
            # e.g.: 20221127T170156Z/stage3-amd64-nomultilib-systemd-mergedusr-20221127T170156Z.tar.xz
            regexp = f"^[0-9]+T[0-9]+Z/stage3-{arch}-nomultilib-systemd-mergedusr-[0-9]+T[0-9]+Z\.tar\.xz"
```

257: do we need the `[0-9TZ]+/` at the beginning and the `.xz` at the end?

behrmann: You tell me :) My point is mainly that the example and the regexp should actually match and document what is expected or needed. Also, the groups for dashes and dots are still a bit weird and the square brackets don't match (the one after `mergedusr` never closes).

257: you're right. fixed with your suggestion except for `\.`, i get `invalid escape sequence` here for some reason. python version? editor stuff?

mkosi/distributions/gentoo.py
257: seems fine, closing this.

mkosi/distributions/gentoo.py
257: keeping this unresolved for future ref.

mkosi/distributions/gentoo.py
behrmann: That's with your editor then, I assume, I ran this directly in the REPL (3.9 and 3.10)

257: changed to `\.`

PR #1178, opened 2022-09-09
Add environment to MkosiState 

pyfisch:
Add a environment attribute to MkosiState and initialize it from MkosiConfig. This change allows for the addition of derived variables like `IMAGE_ID` and `IMAGE_VERSION` to the environment passed to scripts.

Based on the PR #1134 by @mcassaniti and includes the requested changes by @DaanDeMeyer.
It uses dataclass [Post-init processing](https://docs.python.org/3/library/dataclasses.html#post-init-processing) to avoid ugly getters and setters.

### DIFF
--- a/mkosi/backend.py
+++ b/mkosi/backend.py
@@ -506,6 +506,7 @@ class MkosiConfig:
     skeleton_trees: List[Path]
     clean_package_metadata: Union[bool, str]
     remove_files: List[Path]
+    # Environment should not be used directly. Use MkosiState environment instead.
     environment: Dict[str, str]
     build_sources: Optional[Path]
     build_dir: Optional[Path]
@@ -598,6 +599,7 @@ class MkosiState:
     do_run_build_script: bool
     machine_id: str
     for_cache: bool
+    environment: Dict[str, str]
 
     cache_pre_inst: Optional[Path] = None
     cache_pre_dev: Optional[Path] = None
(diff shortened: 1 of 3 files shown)


mkosi/machine.py
```diff
@@ -117,6 +117,10 @@ def build(self) -> None:
         if needs_build(self.config):
             check_root()
             check_native(self.config)
+
+            # Useful if testing within Docker
+            if parse_boolean(os.getenv("MKOSI_TEST_NO_NAMESPACE", "0")):
+                raise unittest.SkipTest("Build test skipped due to environment variable.")
```
DaanDeMeyer, line 123: Let's drop this and do it in a separate PR if you really want it included

mkosi/backend.py
```diff
@@ -506,6 +506,7 @@ class MkosiConfig:
     skeleton_trees: List[Path]
     clean_package_metadata: Union[bool, str]
     remove_files: List[Path]
+    # Environment should not be used directly. Use MkosiState environment instead.
```
DaanDeMeyer, line 509: Can we drop this comment and replace it with a general comment above the config class itself to always use the respective field from MkosiState if it exists?

DaanDeMeyer (changes requested):

mkosi/machine.py
mcassaniti: See #1179

DaanDeMeyer:
Note that we want perfect PRs in mkosi, so next time please use git rebase to fix the existing commits instead of adding new ones on top. I'll squash this PR this time but in the future please use git rebase.

PR #1179, opened 2022-09-09
Skip tests requiring namespaces when using env MKOSI_TEST_NO_NAMESPACE=1

mcassaniti:
This change will skip any unit tests that require Linux namespace support if the environment variable `MKOSI_TEST_NO_NAMESPACE=1`. This is primarily useful for wrapping unit testing inside a Docker container.

### DIFF
--- a/mkosi/machine.py
+++ b/mkosi/machine.py
@@ -117,6 +117,10 @@ class Machine:
         if needs_build(self.config):
             check_root()
             check_native(self.config)
+
+            # Useful if testing within Docker which will not allow more namespaces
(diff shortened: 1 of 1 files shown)


mcassaniti:
@DaanDeMeyer, @behrmann can either of you review?

DaanDeMeyer:
I don't think we should do this in mkosi. We already mark the tests with the "integration" marker so you can easily skip them with pytest -m "not integration"

mcassaniti:
After all this time I've finally confirmed that this works for me (i.e.: `pytest -m 'not integration'`).

PR #1183, opened 2022-09-10
Install package manager by default

DaanDeMeyer:
Currently, we have package managers installed by default in Arch/Ubuntu but not in Fedora/CentOS and variants. Let's make this more consistent by making a package manager available by default in all distributions.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2120,7 +2120,7 @@ def install_fedora(state: MkosiState) -> None:
     setup_dnf(state, repos)
 
     packages = {*state.config.packages}
-    add_packages(state.config, packages, "systemd", "util-linux")
+    add_packages(state.config, packages, "systemd", "util-linux", "dnf")
 
     if not state.do_run_build_script and state.config.bootable:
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -2654,16 +2654,15 @@ def install_arch(state: MkosiState) -> None:
     if state.config.release is not None:
         MkosiPrinter.info("Distribution release specification is not supported for Arch Linux, ignoring.")
 
+    assert state.config.mirror
+
     if state.config.local_mirror:
         server = f"Server = {state.config.local_mirror}"
-    elif state.config.mirror:
+    else:
         if platform.machine() == "aarch64":
```
behrmann, line 2662: Since this is the mirror for the image we're building, shouldn't this be `config.arch`?

behrmann (commented):
LGTM, totally unrelated question, though

mkosi/__init__.py
DaanDeMeyer: Yeah, good point

Issue #779, opened 2021-09-01
proposal: let's drop black enforcement


keszybz:
Autoformatting is **good idea** in general, but in real life it is doesn't work too well.

First, black makes some very peculiar formatting choices, e.g. the use on double-quoted string even when there's no reason for this, and the rust-like extensive splitting to multiple lines with opening and closing parenthesis on multiple lines. This is different than the generally accepted traditional python style.

Second, in mkosi there is a lot of "table-like" code, e.g. various lists of identifiers and repeated definitions. Black doesn't (and probably cannot) understand that there's a repeatable structure to the code and tries to squish everything together to save space.

We *could* sprinkle `#fmt=off`  / `#fmt=on` in more places, but I don't like this approach. If there's a tool that gets in your way, but doesn't really offer any benefits, it's time to drop the tool. If some of us want to use black while editing, that's completely fine, but only use it for newly added or significantly rewritten chunks and where it makes sense.

Some particularly egregious examples:
![Screenshot from 2021-08-28 11-23-36](https://user-images.githubusercontent.com/349618/131650596-c2505027-c6e2-45e3-88ec-5dc6161b8aae.png)
![Screenshot from 2021-08-28 11-32-01](https://user-images.githubusercontent.com/349618/131650709-fc0c4680-5fac-40ff-aa28-f472fbb18527.png)
![Screenshot from 2021-09-01 11-55-48](https://user-images.githubusercontent.com/349618/131651658-5fc1f023-8119-413d-a704-6895288dfa49.png)
![Screenshot from 2021-09-01 11-56-50](https://user-images.githubusercontent.com/349618/131651780-2f7467bb-1d1b-4d8b-b9f7-db7ae2df9b84.png)

DaanDeMeyer:
Sure, we can remove the enforcement, I don't mind too much either way. @behrmann will disagree I think though.

keszybz:
I'm now preparing a patch that replaces os.path.join() with pathlib.Path(). This obviously touches a lot of code in a fairly trivial way, and I'm noticing how much the code is convoluted by the automatic formatting.

behrmann:
> Sure, we can remove the enforcement, I don't mind too much either way. Though I think @behrmann will disagree.

Not opposed per se :) 

[...]

PR #1184, opened 2022-09-11
No longer mention black and fix indentation

pyfisch:
Mkosi uses 4 spaces for indentation, fix 2 functions with 8 spaces. Since #779 formatting with black wasn't enforced but still mentioned as required in README.


behrmann (approved):
LGTM, although I'd probably keep the section in pyproject.toml around, if we ever were to go back to using it. It can be added back just as easily, though.

PR #1189, opened 2022-09-18
Split3

behrmann:
This is the third attempt to split out the distributions into their own seperate submodules. Since subclassing this via `MkosiArgs` wasn't liked, I added this to the Distribution enum. The meat and potatoes of this approach is in the first five commits, although I added a few tweaks later on, as I actually got to moving a few distros over (Arch, Debian and Ubuntu so far).

Along the way, to get things to import probably I had to cut out some more stuff out of `__init__`, a few things went into `backend` (which I still haven't renamed, I'm sorry :) ), but I've tried to thematically carve out a few into `mounts` (a few `mount_foo` functions) and `install` (another terrible name) for `install_foo` and friends. So far I've only moved what explicitly needed moving to appease mypy, though.

I haven't gotten around to actually testing this yet, because of a meager internet connection, but it type checks and I think I squashed circular imports.

Let's hope third time's the charm. :)


behrmann:
Fine by me, although this is just one namespace removed from the previous attempt. ;) 

I'll rework it later. 'guess I will try to move `MkosiState`, `MkosiConfig` and the installer out into their own `basetypes` (or so) module to solve the cyclical dependencies.

PR #1190, opened 2022-09-21
Allow single files to be used with --extra-search-paths

DaanDeMeyer:
Sometimes we just want to use one file from a directory of files
with --extra-search-paths. Let's make sure we support this by putting
a symlink to the file in a temporary directory and adding that directory
to PATH.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -86,7 +86,6 @@ from .backend import (
     is_centos_variant,
     is_epel_variant,
     is_rpm_distribution,
-    nspawn_executable,
     nspawn_knows_arg,
     nspawn_rlimit_params,
     nspawn_version,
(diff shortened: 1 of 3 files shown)


mkosi/__init__.py
```diff
@@ -8002,20 +8003,31 @@ def expand_paths(paths: List[str]) -> List[str]:
...
             # Skip path if it uses a variable not defined.
             pass
     return expanded
 
 
-def prepend_to_environ_path(paths: List[Path]) -> None:
+@contextlib.contextmanager
+def prepend_to_environ_path(paths: Sequence[Path]) -> Iterator[None]:
     if not paths:
         return
```
behrmann, line 8015: The `with` statement in `run_verb` doesn't like the early exit here. You can put the whole body of the with statement in this function except for the yield behind this conditional, then it should work fine.

DaanDeMeyer: I just added an extra `yield` before, that also works

behrmann (approved):

PR #1192, opened 2022-09-21
run swtpm in the background

behrmann:


DaanDeMeyer:
We have this wonderful service management daemon called systemd to manage processes running in the background, we should use it no? I'd suggest we use `systemd-run` here.

Issue #1191, opened 2022-09-21
"mkosi qemu" should automatically start swtpm and expose it


poettering:
It's 2022, and VMs really should come with TPMs just working.

hence, I think we should unconditionally run swtpm in the background when "mkosi qemu" is invoked.

This means forking off swtpm before invoking qemu, with a command line such as:

```
swtpm socket --tpm2 --tpmstate dir="$TPM_STATE" --ctrl type=unixio,path="$TPM_STATE/sock"
```

(with some temporary dir in $TPM_STATE)

And then invoke qemu with `-chardev socket,id=chrtpm,path=$TPM_STATE/sock -tpmdev emulator,id=tpm0,chardev=chrtpm -device tpm-tis,tpmdev=tpm0` added to the command line.

After qemu exits we can remove $TPM_STATE again.

behrmann:
Haven't tested this, but hacked up a quick way of what this might look like in #1192. Is that the direction you imagine?

poettering:
Ah, eh, after posting this, I also started working on this, see #1193...

That one definitely works, I have tested the systemd TPM stuff with it.

behrmann:
No worries. :) Having something that's already tested to work is worth more than a sketch.

PR #1193, opened 2022-09-21
mkosi: invoke qemu with swtpm, it's 2022!

poettering:
If we find swtpm around we'll use it, otherwise not.

Fixes: #1191

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -7693,7 +7693,7 @@ def qemu_check_kvm_support() -> bool:
 
 
 @contextlib.contextmanager
-def run_qemu_cmdline(config: MkosiConfig) -> Iterator[List[str]]:
+def run_qemu_cmdline(config: MkosiConfig, swtpm_socket: Optional[Path] = None) -> Iterator[List[str]]:
     accel = "kvm" if config.qemu_kvm else "tcg"
 
     firmware, fw_supports_sb = find_qemu_firmware(config)
@@ -7753,6 +7753,13 @@ def run_qemu_cmdline(config: MkosiConfig) -> Iterator[List[str]]:
             "-append", build_auxiliary_output_path(config, ".cmdline").read_text().strip(),
         ]
 
+    if swtpm_socket is not None:
+        cmdline += [
+            "-chardev", f"socket,id=chrtpm,path={swtpm_socket}",
+            "-tpmdev", "emulator,id=tpm0,chardev=chrtpm",
+            "-device", "tpm-tis,tpmdev=tpm0",
+            ]
+
     with contextlib.ExitStack() as stack:
         if config.qemu_boot == "uefi" and fw_supports_sb:
             ovmf_vars = stack.enter_context(copy_file_temporary(src=find_ovmf_vars(config), dir=tmp_dir()))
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
Feels rather ironic, but like I mentioned in the other PR, shouldn't we use systemd-run to spawn swtmp?

poettering:
Dunno, not sure. Build systems might not have PID 1 around.

if this was deployment stuff, yes, I'd wrap it in systemd-run. But this is a fancy build script. I think for that it's OK if we just fork of two things, and wait for them afterwards, synchronously, to die, so that we can remove its resources afterwards.

DaanDeMeyer:
> Dunno, not sure. Build systems might not have PID 1 around.

I'm pretty sure we're allowed to say mkosi isn't intended for systems that aren't running systemd as pid1. And I don't think qemu is going to run inside a container either, so I feel like we can rely on systemd-run being available in this scenario.

mkosi/__init__.py
```diff
@@ -7794,9 +7801,36 @@ def run_qemu_cmdline(config: MkosiConfig) -> Iterator[List[str]]:
...
+               "--tpm2",
+               "--tpmstate", f"dir={swtpm_state}",
+               "--ctrl", f"type=unixio,path={swtpm_sock}",
+         ]
+
+        swtpm_proc = spawn(cmd)
+
+        yield swtpm_sock
+
+        swtpm_proc.wait()
```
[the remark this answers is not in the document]
poettering: terminates on its own once it sees qemu disconnect. given it writes stuff to disk, I think we should just let it finish on its own.

poettering:
> I'm pretty sure we're allowed to say mkosi isn't intended for systems that aren't running systemd as pid1. And I don't think qemu is going to run inside a container either, so I feel like we can rely on systemd-run being available in this scenario.

But it doesn't really get us much. we don#t invoke any of the other tools through systemd-run, so why would we spawn this one like that?

behrmann:
Daan and I were talking earlier about maybe wrapping qemu in systemd-run as well and using that to bind the runtimes of both together, but since `swtemp` just terminates on its own once qemu disconnects, I'm fine with this.

poettering:
fixed the mypy thing. PTAL.

mkosi/__init__.py
```diff
@@ -7693,7 +7693,7 @@ def qemu_check_kvm_support() -> bool:
 
 
 @contextlib.contextmanager
-def run_qemu_cmdline(config: MkosiConfig) -> Iterator[List[str]]:
+def run_qemu_cmdline(config: MkosiConfig, swtpm_socket: Optional[Path]) -> Iterator[List[str]]:
```
DaanDeMeyer, line 7696: Need to add a default of `None` here

mkosi/__init__.py
```diff
@@ -7794,9 +7801,36 @@ def run_qemu_cmdline(config: MkosiConfig) -> Iterator[List[str]]:
...
+        yield swtpm_sock
+
+        swtpm_proc.wait()
+
+
 def run_qemu(config: MkosiConfig) -> None:
-    with run_qemu_cmdline(config) as cmdline:
-        run(cmdline, stdout=sys.stdout, stderr=sys.stderr)
+
+    with start_swtpm() as swtpm_socket:
```
DaanDeMeyer, line 7831: I would move this to within `run_qemu_cmdline()`, we probably need to rename it at some point since it does a bit more than that, but it's still the best fit for this stuff

DaanDeMeyer (changes requested):

mkosi/__init__.py
```diff
@@ -7753,6 +7753,13 @@ def run_qemu_cmdline(config: MkosiConfig) -> Iterator[List[str]]:
...
         ]
 
+    if swtpm_socket is not None:
+        cmdline += [
+            "-chardev", f"socket,id=chrtpm,path={swtpm_socket}",
+            "-tpmdev", "emulator,id=tpm0,chardev=chrtpm",
+            "-device", "tpm-tis,tpmdev=tpm0",
+            ]
+
     with contextlib.ExitStack() as stack:
```
DaanDeMeyer, line 7763: You can call it in this block as follows:

```
socket = stack.enter_context(start_swtpm())
```

and then add it to the cmdline right after that

mkosi/__init__.py
poettering: i am not sure we should bother with swtpm where we don't need it though? I'd just expose it where we need it, or where we expose a qemu machine to the user.

quite frankly, i don't understand why run_qemu_cmdline() is implemented as one of those context objects? i am by no means a python guru but i thought the point of context objects was that they can nicely auto-destroy stuff. but the cmdline?

mkosi/__init__.py
poettering: i don't grok what you are saying? i guess my python-fu is too limited...

poettering:
mypy should be fixed now.

DaanDeMeyer:
I'll do a follow up PR

DaanDeMeyer (approved):

PR #1194, opened 2022-09-21
mkosi: add new "pcrphase" tool to dracut initrds

poettering:
And lets sort the list of dracut additions again.

This is a companion to https://github.com/systemd/systemd/pull/24771, and should probably be merged only after that got merged.

(If it's merged before it doesn't have any ill effects, but it's pretty useless...)

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -127,18 +127,25 @@ MKOSI_COMMANDS_SUDO = (Verb.build, Verb.clean, Verb.shell, Verb.boot)
 MKOSI_COMMANDS_CMDLINE = (Verb.build, Verb.shell, Verb.boot, Verb.qemu, Verb.ssh)
 
 DRACUT_SYSTEMD_EXTRAS = [
+    "/usr/bin/systemd-ask-password",
     "/usr/bin/systemd-repart",
+    "/usr/bin/systemd-tty-ask-password-agent",
     "/usr/lib/systemd/system-generators/systemd-veritysetup-generator",
     "/usr/lib/systemd/system/initrd-root-fs.target.wants/systemd-repart.service",
     "/usr/lib/systemd/system/initrd-usr-fs.target",
+    "/usr/lib/systemd/system/initrd.target.wants/systemd-pcrphase-initrd.service",
     "/usr/lib/systemd/system/sysinit.target.wants/veritysetup.target",
+    "/usr/lib/systemd/system/systemd-pcrphase-initrd.service",
     "/usr/lib/systemd/system/systemd-repart.service",
     "/usr/lib/systemd/system/systemd-volatile-root.service",
     "/usr/lib/systemd/system/veritysetup.target",
+    "/usr/lib/systemd/systemd-pcrphase",
     "/usr/lib/systemd/systemd-veritysetup",
     "/usr/lib/systemd/systemd-volatile-root",
-    "/usr/bin/systemd-ask-password",
-    "/usr/bin/systemd-tty-ask-password-agent"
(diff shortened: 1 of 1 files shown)


poettering:
The systemd PR was merged btw, I guess this is ready to go in now!

behrmann (approved):

mkosi/__init__.py
```diff
@@ -127,18 +127,25 @@
...
     "/usr/lib/systemd/system/veritysetup.target",
+    "/usr/lib/systemd/systemd-pcrphase",
     "/usr/lib/systemd/systemd-veritysetup",
     "/usr/lib/systemd/systemd-volatile-root",
-    "/usr/bin/systemd-ask-password",
-    "/usr/bin/systemd-tty-ask-password-agent"
+    "/usr/lib64/libtss2-esys.so.0",
+    "/usr/lib64/libtss2-mu.so.0",
+    "/usr/lib64/libtss2-rc.so.0",
+    "/usr/lib64/libtss2-tcti-device.so.0",
```
behrmann, line 148: This doesn't yet generalize to Debian and Ubuntu, though, because of their multiarch setup. This is not a stopper, since these are all optional items, so we can just add the needed paths there later without any special casing.

poettering: i figure we have to hardcode for each distro/arch combo. grrrr, that sucks. But I think we have no better way....

behrmann: I don't think it's all that dire. `/usr/lib64/` catches Arch and Fedora (and I guess the RHEL family), the same so files under the prefix `/usr/lib/x86_64-linux-gnu/` should catch Debian and Ubuntu. That should make up most already. So adding another four lines here should be enough, since we include the members of this list via `install_optional_item` in dracut. 

Generally it would probably be better, though, if this were a dracut module.

johannbg: > I don't think it's all that dire. `/usr/lib64/` catches Arch and Fedora (and I guess the RHEL family), the same so files under the prefix `/usr/lib/x86_64-linux-gnu/` should catch Debian and Ubuntu. That should make up most already. So adding another four lines here should be enough, since we include the members of this list via `install_optional_item` in dracut.
> 
> Generally it would probably be better, though, if this were a dracut module.

I just noticed this in the upstream systemd src and submitted a module for it however we also have a module for the systemd-repart ( 01systemd-repart ), systemd-veritysetup ( 01systemd-veritysetup ) and systemd-ask-password stuff ( 01systemd-ask-password ) you have in the DRACUT_SYSTEMD_EXTRAS section but for whatever reason are modules that you dont seem to be using.

behrmann: Haven't looked at dracut modules recently, I'd happy about PRs to reduce our homegrown lists!

johannbg: So I think have figured out why you might not be using those modules and that is because systemd module ( along with tpm2-tss ) depend on systemd-udevd ( which is the udev module closer to systemd upstream ) but the base module still depends on the old udev-rules module not the systemd-udevd module, which triggers a failure in dracut. 

If you change the udev-rules to systemd-udevd in the depends() section of the base module thinks should just work.

I need get of my lazy ass and fix that upstream.

keszybz:
Overall, I think that this a bad pattern. We have some list of files in the systemd project, in dracut modules for systemd, here, and in systemd packages for the distro. But meh, if we have the list here, we should / need to keep updating it.

Also, if some paths are missing for Debian, I guess it's fine to update it later.

Let's merge this, since systemd-252 is about to come out. We should release mkosi too.

PR #1195, opened 2022-09-21
Soothe some CodeQL warnings

behrmann:
I just went through the open CodeQL warnings and most of them are stylistic stuff I don't particularly agree with. I applied some of the things that were not outrageous (we did import `DEVNULL` and `PIPE` from `subprocess`, but sometimes used it via the `subprocess` namespace).

- https://github.com/systemd/mkosi/security/code-scanning/3: I haven't followed through all call graphs, but the problem here is that the argument to `die` might contain *part* of a cmdline that also contained a *path* to file containing sensitive material
- Overly permissive file permissions ([1](https://github.com/systemd/mkosi/security/code-scanning/6), [2](https://github.com/systemd/mkosi/security/code-scanning/5), [3](https://github.com/systemd/mkosi/security/code-scanning/4)): I think those are fine.
- https://github.com/systemd/mkosi/security/code-scanning/16: True, but enums can actually point to the same value, so I'd leave that there.
- https://github.com/systemd/mkosi/security/code-scanning/38: Here it can't follow the `nonlocal` variable
- https://github.com/systemd/mkosi/security/code-scanning/37: The file is written for `no_docs` cases, so that should be fine.
- File is not always closed ([1](https://github.com/systemd/mkosi/security/code-scanning/34), [2](https://github.com/systemd/mkosi/security/code-scanning/33)): I think it doesn't understand the `closefd` option of `open`
- I disagree with all the empty except warnings and the implicit return warnings and warnings about overwriting superclass attributes.

This one might have merit: https://github.com/systemd/mkosi/security/code-scanning/35 The file is never closed explicitly by us.

### DIFF
--- a/tests/test_config_parser.py
+++ b/tests/test_config_parser.py
@@ -204,7 +204,7 @@ class MkosiConfig:
             fname = f"{prio:03d}_{fname}"
         config_parser = configparser.RawConfigParser()
         # This is still an open issue on: https://github.com/python/mypy/issues/2427
-        config_parser.optionxform = lambda optionstr: str(optionstr) # type: ignore
+        config_parser.optionxform = str # type: ignore
 
         # Replace lists in dict before calling config_parser write file
         config_all_normalized = copy.deepcopy(config)
@@ -812,7 +812,7 @@ class MkosiConfigIniLists1(MkosiConfigOne):
         ini_lines = [
             "[Content]",
             "Packages=[ vim,!vi",
-            "  ca-certificates, bzip ]" "",
+            "  ca-certificates, bzip ]  ",
             "[Output]",
             "KernelCommandLine=console=ttyS1",
             "  driver.feature=1",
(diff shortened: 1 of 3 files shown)


DaanDeMeyer (approved):

behrmann:
I'll go through the warnings in the dashboard tomorrow and dismiss them (you can give reasons for the dismissal)

tests/test_config_parser.py
```diff
@@ -812,7 +812,7 @@ def prepare_mkosi_conf_d_2(self, directory: str) -> None:
         ini_lines = [
             "[Content]",
             "Packages=[ vim,!vi",
-            "  ca-certificates, bzip ]" "",
+            "  ca-certificates, bzip ]  ",
```
keszybz, line 815: Hmm, this adds a trailing whitespace, no?

behrmann: Yes, indeed! I forgot I wanted to put that into a separate commit. Before, this did an implicit string concatenation, adding an empty string to the line, so a noop that's already done when compiling the byte code. I though this would be more realistic as a test case if the line had some trailing whitespace and the tests pass either way. This should have been a separate commit, though, explaining the reasoning, sorry.

Issue #1180, opened 2022-09-09
RFE: Support building images with signed PCR policies


poettering:
It would be great to support building OS images with signed PCR policies in mkosi.

With https://github.com/systemd/systemd/pull/24351 and https://github.com/systemd/systemd/pull/24458 there's now support in systemd for TPM2 based disk encryption policies that bind the unsealed keys to the booted kernel image. And if the boot image locks the root fs via verity then this transiently means that you can build OS images now that can have encrypted file systems that can be unlocked in unattended scenarios via TPM2 in a way that strictly requires that OS image (or any other signed by the same key pair) to unlock it.

I think this is quite a major step ahead: secure encryption by default without requiring interactivity. This I am sure should become the default for servers and iot devices sooner or later, and probably even for desktops (as this means we can boot up without interactivity and have encrypted storage already accessible *before* login, so that only the actual act of logging in will require interactivity).

Implementing this would probably require change things so that dracut doesn't do the objcopy to merge the seperate parts into a unified kernel image anymore. Instead we should do that in mkosi ourselves, and do a simple `systemd-measure sign` call before, with pretty much the same parameters.

The linked PRs have man pages with examples for this.

I will prep more docs soon, but the man pages should be enough already to get things started.

I'll eventually hack this up myself for mkosi, but would love if somebody interest of course would beat me to it ;-)

poettering:
(moving the objcopy call out of dracut and into mkosi is probably also nice to support alternative initrd generators, like the arch one or the debian one?)

DaanDeMeyer:
We already do the objcopy out of dracut these days btw

DaanDeMeyer:
> (moving the objcopy call out of dracut and into mkosi is probably also nice to support alternative initrd generators, like the arch one or the debian one?)

I tried, but these don't run systemd in the initrd by default, and we definitely want that, so we're still using dracut for now.

behrmann:
On Fri, Sep 09, 2022 at 12:33:29PM -0700, Daan De Meyer wrote:
> I tried, but these don't run systemd in the initrd by default, and we definitely want that, so we're still using dracut for now.
> 

At least mkinitcpio can be made to, not sure about initramfs-tools.

PR #1197, opened 2022-09-21
mkosi: pre-calculate expected PCR 11 values, and sign them, and inclu…

poettering:
…de in unified kernel image

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3975,13 +3975,61 @@ def install_unified_kernel(
             cmdline = state.workspace / "cmdline"
             cmdline.write_text(boot_options)
             initrd = state.root / boot_directory(state, kver) / "initrd"
+            pcrsig = None
+            pcrpkey = None
 
-            cmd: Sequence[PathString] = [
+            # If a SecureBoot key is configured, and we have the
+            # systemd-measure binary around, then also include a
+            # signature of expected PCR 11 values in the kernel image
+            if state.config.secure_boot:
+                if shutil.which('systemd-measure'):
+                    with complete_step("Generating PCR 11 signature…"):
+                        from cryptography import x509
+                        from cryptography.hazmat.primitives import serialization
+
+                        # Extract the public key from the SecureBoot certificate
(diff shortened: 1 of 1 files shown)


poettering:
Fixes #1180

mkosi/__init__.py
```diff
@@ -3975,20 +3975,72 @@ def install_unified_kernel(
...
+                "--add-section", f".osrel={osrelease}",         "--change-section-vma",   ".osrel=0x20000",
+                "--add-section", f".cmdline={cmdline}",         "--change-section-vma", ".cmdline=0x30000",
+                "--add-section", f".linux={state.root / kimg}", "--change-section-vma",   ".linux=0x2000000",
+                "--add-section", f".initrd={initrd}",           "--change-section-vma",  ".initrd=0x3000000",
+            ]
+
+            if pcrsig is not None:
+                cmd += [
+                    "--add-section", f".pcrsig={pcrsig}",       "--change-section-vma",  ".pcrsig=0x80000",
+                    "--add-section", f".pcrpkey={pcrpkey}",     "--change-section-vma", ".pcrpkey=0x100000",
```
DaanDeMeyer, line 4029: Not aligned correctly?

DaanDeMeyer (approved):

mkosi/__init__.py
```diff
@@ -3975,20 +3975,72 @@ def install_unified_kernel(
...
+            pcrpkey = None
+
+            # If a SecureBoot key is configured, and we have the
+            # systemd-measure binary around, then also include a
+            # signature of expected PCR 11 values in the kernel image
+            if state.config.secure_boot:
+                if shutil.which('systemd-measure'):
+                    with complete_step("Generating PCR 11 signature…"):
+                        from cryptography import x509
+                        from cryptography.hazmat.primitives import serialization
```
behrmann, line 3988: Until now `cryptography` was only imported in `make_verity_sig`, so not using verity one could forego the cryptography dependency. Since this is now in the main path we should decide whether we skip this and warn or whether we say that `cryptography` is now an mkosi dependency proper, since people usually won't get do decide whether `systemd-measure` is installed or not.

PR #1198, opened 2022-09-21
swtpm followups

DaanDeMeyer:


DaanDeMeyer:
I wonder if swtpm should support socket activation with `Accept=yes`. The annoying thing would be configuring the CLI options to use for each instantiation, it only really works if you know the set of CLI options up front.

Maybe we should ship and link such a unit in mkosi.

mkosi/__init__.py
```diff
@@ -7693,7 +7693,33 @@ def qemu_check_kvm_support() -> bool:
...
+               "--tpmstate", f"dir={swtpm_state}",
+               "--ctrl", f"type=unixio,path={swtpm_sock}",
+         ]
+
+        swtpm_proc = spawn(cmd)
+
+        try:
+            yield swtpm_sock
+        finally:
+            swtpm_proc.wait()
```
poettering, line 7718: for someone who is not a python guru, why pack this in try/finally?

DaanDeMeyer: An exception can be thrown through the `yield`, which will immediately bubble up without executing the rest of the block, so if an exception was thrown, we'd never wait for the swtpm process to exit.

mkosi/__init__.py
```diff
@@ -7794,43 +7813,24 @@ def run_qemu_cmdline(config: MkosiConfig, swtpm_socket: Optional[Path] = None) -
                 "scsi-hd,drive=hd,bootindex=1",
             ]
 
+        swtpm_socket = stack.enter_context(start_swtpm())
```
poettering, line 7816: hmm, for someone who is not a python guru what precisely does this do? this is a fake "with" expression, right?  but how does destruction here work? when is control returned to start_swtpm() to destroy the tempory dir?

DaanDeMeyer: So this block starts with the following:

```
with contextlib.Exitstack() as stack:
```

That creates an exit stack, which you can read up on [here](https://docs.python.org/3/library/contextlib.html). To summarize, whenever `stack.enter_context()` is called, the `__enter__` method of the context manager is called, and whatever that returns is propagated as the return value of `enter_context()`. When using `contextlib.contextmanager`, this means everything up to the first `yield` is executed and whatever is passed to `yield` becomes the return value of `enter_context()`.

Cleanup happens when the `__exit__()` method of the exit stack itself is called, which happens at the end of the `with` block that created the exit stack. It will call the `__exit__()` methods of all the context manager that were pushed onto the stack in reverse order. So when the `with` block ends, control is handed back to `start_swtpm()` as it's `__exit__()` method is called along with those of the other context managers that we pushed onto the exit stack.

behrmann (approved):

DaanDeMeyer:
I played around a bit and these seem to do the trick:

```
# /etc/systemd/system/swtpm.socket
[Unit]
Before=sockets.target

[Socket]
ListenStream=/run/swtpm.socket
Accept=yes
SocketMode=0666

# /etc/systemd/system/swtpm@.service
[Unit]
CollectMode=inactive-or-failed

[Service]
Type=exec
RuntimeDirectory=swtpm-%i
StandardInput=socket
StandardOutput=journal
ExecStart=swtpm socket --tpm2 --tpmstate dir=${RUNTIME_DIRECTORY} --ctrl type=unixio,clientfd=0
```

You just pass /run/swtpm.socket to qemu and it all works beautifully. I wonder if this is a better approach than starting swtpm ourselves.

mkosi/__init__.py
poettering: ah, thanks, got it! Learning something new every day!

poettering:
I wonder if we can convince swtpm to just ship a pair of unit files like that upstream?

DaanDeMeyer:
See https://github.com/stefanberger/swtpm/issues/753, which has made me very confused on how swtpm actually works

mkosi/__init__.py
keszybz: In other words, `ExitStack` allows items in the context manager chain (stack?) to be applied in separate steps, which also means that some of them can be skipped.

keszybz (commented):
LGTM too.

PR #1199, opened 2022-09-22
Fix undefined PIPE issue

DaanDeMeyer:


poettering:
hmm, why did this work for me?

behrmann:
That's because of the order of the PRs being merged.

behrmann:
Merging because this doesn't need to burn more CI.

PR #1200, opened 2022-09-22
Guard cryptography imports

behrmann:
This guards the cryptography imports - and adds a nicer error in the verity case - so that we are still nominally dependency-free beyond the stdlib.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3757,10 +3757,13 @@ def make_verity_sig(
 
     assert root_hash is not None
 
-    from cryptography import x509
-    from cryptography.hazmat.primitives import hashes, serialization
-    from cryptography.hazmat.primitives.asymmetric import ec, rsa
-    from cryptography.hazmat.primitives.serialization import pkcs7
+    try:
+        from cryptography import x509
+        from cryptography.hazmat.primitives import hashes, serialization
+        from cryptography.hazmat.primitives.asymmetric import ec, rsa
+        from cryptography.hazmat.primitives.serialization import pkcs7
+    except ImportError:
+        die("Verity support needs the cryptography module. Please install it.")
 
     with complete_step("Signing verity root hash…"):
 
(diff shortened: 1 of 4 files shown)


mkosi/__init__.py
```diff
@@ -3987,38 +3990,42 @@ def install_unified_kernel(
...
+                            ]
+
+                            c = run(cmd_measure, stdout=subprocess.PIPE)
+
+                            pcrsig = state.workspace / "pcrsig.json"
+                            pcrsig.write_bytes(c.stdout)
+                    else:
+                        MkosiPrinter.info("Couldn't find systemd-measure binary, not embedding PCR signature in unified kernel image.")
+                except ImportError:
+                    MkosiPrinter.info("Couldn't import the cryptography Python module, not embedding PCR signature in unified kernel image.")
```
DaanDeMeyer, line 4028: This isn't ideal, maybe we should gate the measure stuff behind a --measure option, so that we can fail hard when any of the dependencies are not available

behrmann: Fine by me, I just wanted to get the crash for the case of `systemd-measure` existing, but `cryptography` not being installed out of the way.

mkosi/__init__.py
```diff
@@ -3989,43 +3989,42 @@ def install_unified_kernel(
...
-                            pcrsig = state.workspace / "pcrsig.json"
-                            pcrsig.write_bytes(c.stdout)
-                    else:
-                        MkosiPrinter.info("Couldn't find systemd-measure binary, not embedding PCR signature in unified kernel image.")
                 except ImportError:
-                    MkosiPrinter.info("Couldn't import the cryptography Python module, not embedding PCR signature in unified kernel image.")
+                    die("Couldn't import the cryptography Python module. This is needed for the --measure option")
+
+                if not shutil.which('systemd-measure'):
+                    die("Couldn't find systemd-measure binary, not embedding PCR signature in unified kernel image.")
```
DaanDeMeyer, line 4000: This message should be updated to say --measure requires the systemd-measure binary or so

DaanDeMeyer (changes requested):

mkosi/__init__.py
behrmann: done

behrmann:
Ok, rewrote this so that it's an option instead of skipping it if `cryptography` is not there.

@poettering; This changes the semantics to make this explicitly opt in. If you're not fine with that, we could also do this during argument parsing and add an `auto` for this, so that it's on by default if both `cryptography` and `systemd-measure` are there and only fail hard in the case of people choosing `--measure=true` but missing either.

DaanDeMeyer (approved):
LGTM, but let's wait for @poettering to take a look as well

behrmann:
I'll prep a PR to make this try-state and rename this over the weekend.

PR #1203, opened 2022-09-25
ubuntu: Check both native machine arch and command line arch

marex:
Test both native machine architecture and command line --architecture switch when selecting the ports.ubuntu.com . The former only works in case mkosi runs on aarch64 already, the later is necessary in case the mkosi is generated foreign architecture container (e.g. mkosi on amd64 generates arm64 container).

Fix the mirror URL which was incorrect and was missing the ubuntu-ports suffix.

Fix security.ubuntu.com URL as well, which has to point to ubuntu-ports for aarch64 machines.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2464,7 +2464,10 @@ def add_apt_auxiliary_repos(state: MkosiState, repos: Set[str]) -> None:
 
     # Security updates repos are never mirrored
     if state.config.distribution == Distribution.ubuntu:
-        security = f"deb http://security.ubuntu.com/ubuntu/ {state.config.release}-security {' '.join(repos)}"
+        if state.config.architecture == "x86" or state.config.architecture == "x86_64":
+            security = f"deb http://security.ubuntu.com/ubuntu/ {state.config.release}-security {' '.join(repos)}"
+        else:
+            security = f"deb http://ports.ubuntu.com/ {state.config.release}-security {' '.join(repos)}"
     elif state.config.release in ("stretch", "buster"):
         security = f"deb http://security.debian.org/debian-security/ {state.config.release}/updates main"
     else:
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -6412,8 +6415,8 @@ def load_args(args: argparse.Namespace) -> MkosiConfig:
             args.mirror = "http://deb.debian.org/debian"
         elif args.distribution == Distribution.ubuntu:
             args.mirror = "http://archive.ubuntu.com/ubuntu"
-            if platform.machine() == "aarch64":
-                args.mirror = "http://ports.ubuntu.com/"
+            if platform.machine() == "aarch64" or args.architecture == "aarch64":
+                args.mirror = "http://ports.ubuntu.com/ubuntu-ports"
```
DaanDeMeyer, line 6419: Same here, please remove the `platform.machine()` check and keep only the `args.architecture` check

DaanDeMeyer (changes requested):

behrmann:
Also, doesn't have Debian (and hence I'd guess Ubuntu) this whole multiarch thing, where it can transparently put qemu in between, so you can run non-native binaries?

DaanDeMeyer (approved):

mkosi/__init__.py
```diff
@@ -2464,7 +2464,10 @@ def add_apt_auxiliary_repos(state: MkosiState, repos: Set[str]) -> None:
 
     # Security updates repos are never mirrored
     if state.config.distribution == Distribution.ubuntu:
-        security = f"deb http://security.ubuntu.com/ubuntu/ {state.config.release}-security {' '.join(repos)}"
+        if state.config.architecture == "aarch64":
```
bluca, line 2467: ports is used for other architectures I think? armhf, ppc64el, riscv64, s390x http://ports.ubuntu.com/dists/jammy/

DaanDeMeyer: @marex Can you add these architectures as well here?

bluca (approved):

Issue #1201, opened 2022-09-23
Support libarchive cpio


tpgxyz:
Hi,

please support libarchive cpio(bsdcpio), which does not have `--reproducible`

https://lists.gnu.org/archive/html/bug-cpio/2014-11/msg00000.html

https://github.com/systemd/mkosi/blob/2c45d0effb1871750a2e9f897510d2745cb6d6b9/mkosi/__init__.py#L3489

behrmann:
I'd review a PR for this, but I'm not entirely sold, that this is a good idea. Why isn't using the GNU cpio an option or why can't BSD cpio implement these options? Making this configurable would make the output unreproducible and I don't think this is were everybody's headed nowadays. Granted, unreproducible output would be preferable to no output at all, if GNU cpio really is not an option, but some background would be helpful.

tpgxyz:
@behrmann First there are some benefits of having libarchive as replacement for GNU archiving tools.
Well my attached thread for GNU ML explains couple of things. Second by default libarchive cpio does provide reproducible archives

poettering:
gah, which distros install /usr/bin/cpio as the bsd version? This is a mess to support. what's the point? different for the sake of doing things differently?

do these distros make the gnu version available as gcpio or so? if so, I'd just change mkosi to search for "gcpio" first, and use "cpio" only as fallback. I'd rather have systematic behaviour  so that we use the same tools on all distros, then to add messy quirks for the bad choices of specific distros...

behrmann:
> @behrmann First there are some benefits of having libarchive as replacement for GNU archiving tools. Well my attached thread for GNU ML explains couple of things. Second by default libarchive cpio does provide reproducible archives

It would have helped to get a link to the message that actually said something about libarchive's version, instead of me having to read through the whole thread only to find the info the the very last message :)

> gah, which distros install /usr/bin/cpio as the bsd version? This is a mess to support. what's the point? different for the sake of doing things differently?
> 
> do these distros make the gnu version available as gcpio or so? if so, I'd just change mkosi to search for "gcpio" first, and use "cpio" only as fallback. I'd rather have systematic behaviour so that we use the same tools on all distros, then to add messy quirks for the bad choices of specific distros...

I'd assume Mandriva does this, since this already prompted the default to `gtar` wit `tar_binary()`. I guess doing the same for `cpio` would be the consistent solution.

tpgxyz:
[...]

PR #1205, opened 2022-09-26
add support for bsdcpio

tpgxyz:
This should improve the situation https://github.com/systemd/mkosi/issues/1201

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3470,6 +3470,12 @@ def find_files(root: Path) -> Iterator[Path]:
                                  lambda entry: Path(entry.path).relative_to(root))
 
 
+def cpio_binary() -> str:
+    # Some distros (Mandriva) install BSD cpio as "cpio", hence prefer
+    # GNU "cpio" if it exists, which should be GNU cpio wherever it exists.
+    return "bsdcpio" if shutil.which("bsdcpio") else "cpio"
+
+
 def make_cpio(state: MkosiState) -> Optional[BinaryIO]:
     if state.do_run_build_script:
         return None
@@ -3485,9 +3491,11 @@ def make_cpio(state: MkosiState) -> Optional[BinaryIO]:
 
         compressor = compressor_command(should_compress_output(state.config))
         files = find_files(root_dir)
-        cmd: List[PathString] = [
-            "cpio", "-o", "--reproducible", "--null", "-H", "newc", "--quiet", "-D", root_dir
-        ]
+        cmd: List[PathString] = [cpio_binary(), "-o", "--null", "-H", "newc", "--quiet"]
+        if cpio_binary() = "cpio":
+            cmd += ["--reproducible", "-D"]
+
+        cmd += [root_dir]
 
         with spawn(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE) as cpio:
             #  https://github.com/python/mypy/issues/10583


poettering:
I really would prefer going the other way: always pick the same tool.

mkosi/__init__.py
```diff
@@ -3469,6 +3469,13 @@ def find_files(root: Path) -> Iterator[Path]:
     yield from scandir_recursive(root,
                                  lambda entry: Path(entry.path).relative_to(root))
 
+def cpio_binary() -> str:
+    # Some distros (Mandriva) install BSD cpio as "cpio", hence prefer
+    # GNU "cpio" if it exists, which should be GNU cpio wherever it exists.
+    # We are interested in exposing same behaviour everywhere hence
+    # it's preferable to use the same implementation of cpio
+    # everywhere.
```
behrmann, line 3477: That's not what the code does, though, since it prefers `bsdcpio` if it exists (and on my Arch machines it exists).

mkosi/__init__.py
```diff
@@ -3469,6 +3469,13 @@ def find_files(root: Path) -> Iterator[Path]:
...
+    return "bsdcpio" if shutil.which("bsdcpio") else "cpio"
```
behrmann, line 3479: this should be two empty lines.

mkosi/__init__.py
```diff
@@ -3485,9 +3492,13 @@ def make_cpio(state: MkosiState) -> Optional[BinaryIO]:
 
         compressor = compressor_command(should_compress_output(state.config))
         files = find_files(root_dir)
-        cmd: List[PathString] = [
-            "cpio", "-o", "--reproducible", "--null", "-H", "newc", "--quiet", "-D", root_dir
-        ]
+        cmd: List[PathString] = [cpio_binary(), "-o", "--null", "-H", "newc", "--quiet"]
```
behrmann, line 3495: Would this be reproducible, though? From the discussion I got the impression, that libarchive's cpio is reproducible, since it uses pax, but `-H newc` will use `nwc` instead of `pax`.

mkosi/__init__.py
```diff
@@ -3485,9 +3492,13 @@ def make_cpio(state: MkosiState) -> Optional[BinaryIO]:
...
+        if cpio_binary() == "bsdcpio":
+            cmd += ["-d"]
```
behrmann, line 3497: GNU cpio has the same option. Is there a reason to add it here? Maybe the same would apply for GNU cpio as well.

behrmann (commented):
This should consistently choose GNU cpio if GNU cpio exists on the system, so deciding whether to use the BSD cpio syntax should probably only happen if the result from `cpio_binary` is `cpio` instead of `gcpio` and if some function `cpio_is_bsd_cpio` tells us that we indeed need have a BSD cpio, e.g. via checking the `--version` output.

mkosi/__init__.py
tpgxyz: Well originally it was "-D" in this code, while man bsdcpio says "-d". I checked man gnutar and it says "-d".

mkosi/__init__.py
tpgxyz: Yes, thanks for spotting this

mkosi/__init__.py
tpgxyz: removed

mkosi/__init__.py
behrmann: Well, now I'm confused. `-d` and `-D` do different things in GNU cpio. `-D` changes the working directory and `-d` "creates leading directories where needed". BSD cpio seems to support only the latter. 

My questions was that if creating needed directories was needed for BSD cpio, might it also be needed or beneficial in the GNU case. Might your issue #1206 stem from BSD cpio not supporting changing the directory?

mkosi/__init__.py
behrmann: Sorry, maybe I was unclear. Between function definitions should be two empty lines, so two before and after this function.

mkosi/__init__.py
behrmann: Could you maybe solve this in Mandriva? Because as the code currently is this would prefer BSD cpio over GNU cpio, but we want to always use GNU cpio when it is available.

poettering:
again, given bsdcpio doesn't even work (→ #1206), why not just be systematic and stick to GNU tools for everything. i really see no point in special casing these artifical problems that mandriva creates for itself there. There's really no point in what they are doing, and I doubt we have to cater for their bad choices. Let#s do what we do for tar: use gcpio if it exists, and fallback to plain cpio only if it does not.

tpgxyz:
> Let#s do what we do for tar: use gcpio if it exists, and fallback to plain cpio only if it does not.
Contrary to tar, a gcpio symlink does not exist in cpio like gtar in tar

mkosi/__init__.py
poettering: under which name does the gnu cpio RPM install the binary then under mandriva? it must pick something

keszybz: The comment still doesn't match the code in the latest version…

DaanDeMeyer:
Let's close this until there's an update

Issue #1207, opened 2022-09-27
mkosi genkey fails


baaa-sh:
I just discovered mkosi genkey... excellent.

...only it doesn't work:

```
[baaash@delta ~]$ sudo mkosi genkey
[sudo] password for baaash:
‣ Generating secure boot keys rsa:2048 for CN 'mkosi of baaash'.

The keys will expire in 730 days (Thursday 26. September 2024).
Remember to roll them over to new ones before then.

Traceback (most recent call last):
  File "/usr/lib64/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib64/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/usr/lib/python3.10/site-packages/mkosi/__main__.py", line 36, in <module>
    main()
  File "/usr/lib/python3.10/site-packages/mkosi/__main__.py", line 27, in main
    run_verb(a)
  File "/usr/lib/python3.10/site-packages/mkosi/__init__.py", line 8390, in run_verb
    generate_secure_boot_key(args)
  File "/usr/lib/python3.10/site-packages/mkosi/__init__.py", line 8317, in generate_secure_boot_key
    os.execvp(cmd[0], cmd)
  File "/usr/lib64/python3.10/os.py", line 574, in execvp
    _execvpe(file, args)
  File "/usr/lib64/python3.10/os.py", line 616, in _execvpe
    raise last_exc
  File "/usr/lib64/python3.10/os.py", line 607, in _execvpe
    exec_func(fullname, *argrest)
FileNotFoundError: [Errno 2] No such file or directory

```
It's looking for something there... what?  that isn't exactly clear to me.

looking at the function statements where it's choking:

```
def execvpe(file, args, env):
    """execvpe(file, args, env)

    Execute the executable file (which is searched for along $PATH)
    with argument list args and environment env, replacing the
    current process.
    args may be a list or tuple of strings. """
    _execvpe(file, args, env)

__all__.extend(["execl","execle","execlp","execlpe","execvp","execvpe"])

def _execvpe(file, args, env=None):
    if env is not None:
        exec_func = execve
        argrest = (args, env)
    else:
        exec_func = execv
        argrest = (args,)
        env = environ

    if path.dirname(file):
        exec_func(file, *argrest)
        return
    saved_exc = None
    path_list = get_exec_path(env)
    if name != 'nt':
        file = fsencode(file)
[...]

PR #1208, opened 2022-09-30
Use run instead of execvp in genkey command

behrmann:
As discussed in #1207.

As a driveby this also changes `run` to use `os.fspath` instead of `str` to stringify its arguments.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -53,7 +53,6 @@ from typing import (
     Iterator,
     List,
     NamedTuple,
-    NoReturn,
     Optional,
     Sequence,
     Set,
@@ -8002,7 +8001,7 @@ def run_serve(config: MkosiConfig) -> None:
         httpd.serve_forever()
 
 
-def generate_secure_boot_key(config: MkosiConfig) -> NoReturn:
+def generate_secure_boot_key(config: MkosiConfig) -> None:
     """Generate secure boot keys using openssl"""
 
     keylength = 2048
@@ -8031,7 +8030,7 @@ def generate_secure_boot_key(config: MkosiConfig) -> NoReturn:
         )
     )
 
-    cmd: List[str] = [
+    cmd: List[PathString] = [
         "openssl",
         "req",
         "-new",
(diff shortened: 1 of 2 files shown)


mkosi/__init__.py
```diff
@@ -8031,25 +8031,25 @@ def generate_secure_boot_key(config: MkosiConfig) -> NoReturn:
...
         "-days",
         str(config.secure_boot_valid_days),
         "-subj",
         f"/CN={cn}/",
         "-nodes",
     ]
-
-    os.execvp(cmd[0], cmd)
+    proc = run(cmd)
+    sys.exit(proc.returncode)
```
DaanDeMeyer, line 8052: Can we drop this? run() will throw an exception if the command fails anyway

mkosi/backend.py
```diff
@@ -833,7 +833,7 @@ def run(
     env: Mapping[str, Any] = {},
     **kwargs: Any,
 ) -> CompletedProcess:
-    cmdline = [str(x) for x in cmdline]
+    cmdline = [os.fspath(x) for x in cmdline]
```
DaanDeMeyer, line 836: Is this correct? I thought we supported passing literally anything to run() and it would be stringified

DaanDeMeyer (changes requested):

mkosi/backend.py
behrmann: Well, we do `Sequence[PathString]` for the command, which is `Sequence[Union[str, Path]]`, so I think this should be correct.

mkosi/__init__.py
DaanDeMeyer: OK, but then we should handle that in run_verb(), and not in this function. `sys.exit()` is just bad because it skips all cleanup that may still need to happen

behrmann: `sys.exit` only raises `SystemExit`. It's not an abort like `os._exit`. Looking for example at the documentation of [atexit](https://docs.python.org/3/library/atexit.html) those are handled fine by a regular exit, i.e. `sys.exit`. Also the documentation to [`SystemExit`](https://docs.python.org/3/library/exceptions.html#SystemExit) says
> A call to [sys.exit()](https://docs.python.org/3/library/sys.html#sys.exit) is translated into an exception so that clean-up handlers ([finally](https://docs.python.org/3/reference/compound_stmts.html#finally) clauses of [try](https://docs.python.org/3/reference/compound_stmts.html#try) statements) can be executed, and so that a debugger can execute a script without running the risk of losing control.

DaanDeMeyer: I still don't think a nested function should make the decision to exit the program. Let's say we want to be able to generate keys and build an image at the same time, we'd have to remove this code again.

We can just do `return generate_secure_boot_key()` in `run_verb()` and it'll work exactly the same without encoding the exit information in `generate_secure_boot_key()`

PR #1213, opened 2022-10-03
Rename --measure option and add a special value auto

behrmann:
As discussed in #1200 here another follow-up. 

It's a bit more unwieldy than I'd liked. Why that is, can be found in the message of the second commit. This would probably work just a fine without the additional argparse.action, but I didn't test this so far, because I didn't want to stray too much from what we had so far.


poettering:
lgtm

DaanDeMeyer:
I hate argparse

PR #1216, opened 2022-10-04
Refactor mount_cache for better readability

behrmann:


DaanDeMeyer:
Can't set label but good to merge

PR #1217, opened 2022-10-04
Split v4

behrmann:
And another round :)

This is basically the same series as #1189, but rebased against current main, (hopefully) more bisectable (making sure at least the basic CI tests pass in between) and, as disccused in #1189, attached to `MkosiState` instead of the `Distribution` enum.


mkosi/backend.py
```diff
@@ -37,12 +38,16 @@
...
+from .distributions import DistributionInstaller
 from .syscall import (
     blkpg_add_partition,
     blkpg_del_partition,
     block_reread_partition_table,
 )
 
+
+
+
```
DaanDeMeyer: Spurious newlines

DaanDeMeyer (changes requested):
Looks great, definitely think this is the way we should go, some comments though.

mkosi/backend.py
```diff
@@ -634,6 +639,24 @@ def get_partition(self, ident: PartitionIdentifier) -> Optional[Partition]:
...
+        except AttributeError:
+            try:
+                distro = str(self.config.distribution)
+                mod = importlib.import_module(f"mkosi.distributions.{distro}")
+                installer = getattr(mod, f"{distro.capitalize()}Installer")
+                instance = installer() if issubclass(installer, DistributionInstaller) else None
+            except (ImportError, AttributeError):
+                instance = None
+            setattr(self, "_installer", instance)
+            return cast(Optional["DistributionInstaller"], instance)
```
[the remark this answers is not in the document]
behrmann: Yes, this *is* unnecessarily complex, but it's only scaffolding. I wanted to port some more distros this weekend and once all of them are ported this can go away pretty much wholesale.

mkosi/backend.py
behrmann: Ah, thought I had gotten rid of them already, they're a rebase artifact. I'll remove them.

mkosi/distributions/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+from pathlib import Path
+from typing import TYPE_CHECKING, List
+
+if TYPE_CHECKING:
+    from ..backend import MkosiState
+
+
+class DistributionInstaller:
+    @classmethod
+    def install(cls, state: "MkosiState") -> None:
```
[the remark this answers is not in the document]
behrmann: Probably, but not sure... I think this import had some other side effects, will look into it.

mkosi/distributions/arch.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+class ArchInstaller(DistributionInstaller):
+    @classmethod
+    def cache_path(cls) -> List[str]:
+        return ["var/cache/pacman/pkg"]
+
+    @classmethod
+    def install(cls, state: "MkosiState") -> None:
+        return install_arch(state)
```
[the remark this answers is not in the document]
behrmann: My thinking was, that `DistributionInstaller` would never have state beyond state that generally applies to that specific distribution - there's already examples in the Debian version - and state specific to our current invocation would be in `MkosiState`. 

I'd prefer to keep it like this for now and we can change it later, since changing it from a class to an instance method will always be easier.

mkosi/distributions/arch.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+def invoke_pacman(root: Path, pacman_conf: Path, packages: Set[str]) -> None:
+    run(["pacman", "--config", pacman_conf, "--noconfirm", "-Sy", *sort_packages(packages)], env=dict(KERNEL_INSTALL_BYPASS="1"))
+
+
+@complete_step("Installing Arch Linux…")
+def install_arch(state: MkosiState) -> None:
```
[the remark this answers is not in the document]
behrmann: Will do. I mostly did it in two different ways for Arch and Debian to showcase how it can work.

mkosi/distributions/__init__.py
```diff
@@ -8,6 +8,8 @@
 
 
 class DistributionInstaller:
+    needs_skeletons_after_bootstrap = False
```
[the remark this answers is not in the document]
behrmann: In principle I'm fine with making everything a method, but it might be a bit overkill.

mkosi/distributions/ubuntu.py
```diff
@@ -0,0 +1,34 @@
+# SPDX-License-Identifier: LGPL-2.1+
+
+from typing import Set
+
+from ..backend import MkosiState, add_packages
+from .debian import DebianInstaller
+
+
+class UbuntuInstaller(DebianInstaller):
```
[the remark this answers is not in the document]
behrmann: Generally not a big fan of it either, but to put it in some words I heard recently: I think it can be done tastefully. :)

Without multiple inheritance this is a one way street of children overriding parents and I look at this more like drop in configuration and pushing this out to some other functions that we're then overwriting just reinvents inheritance again, so I'd like to keep it simple and stay with what's the language points one to.

behrmann:
I'll rework the PR for the changes over the weekend, but of a post vaccination haze at the moment.

mkosi/distributions/ubuntu.py
```diff
@@ -0,0 +1,34 @@
...
+        state.root.joinpath(f"etc/apt/sources.list.d/{state.config.release}-updates.list").write_text(f"{updates}\n")
+
+        # Security updates repos are never mirrored
+        security = f"deb http://security.ubuntu.com/ubuntu/ {state.config.release}-security {' '.join(repos)}"
+
+        state.root.joinpath(f"etc/apt/sources.list.d/{state.config.release}-security.list").write_text(f"{security}\n")
+
+    @classmethod
+    def fixup_resolved(cls, state: MkosiState, extra_packages: Set[str]) -> None:
+        pass
```
DaanDeMeyer, line 34: If we just have a shared function between debian/ubuntu, this would simply be something that's not part of that and is only done by debian after calling the shared function

mkosi/distributions/ubuntu.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @classmethod
+    def add_default_kernel_package(cls, state: MkosiState, extra_packages: Set[str]) -> None:
+        # use the global metapckage linux-generic if the user didn't pick one
+        if ("linux-generic" not in extra_packages and
+            not any(package.startswith("linux-image") for package in extra_packages)):
+            add_packages(state.config, extra_packages, "linux-generic")
+
+    @classmethod
+    def add_apt_auxiliary_repos(cls, state: MkosiState, repos: Set[str]) -> None:
```
DaanDeMeyer, line 20: We already have the `Repo` concept for RPM distros, we can easily do the same here and pass those repo definitions to a shared install function.

mkosi/distributions/ubuntu.py
```diff
@@ -0,0 +1,34 @@
...
+
+from ..backend import MkosiState, add_packages
+from .debian import DebianInstaller
+
+
+class UbuntuInstaller(DebianInstaller):
+    repositories_for_boot = {"universe"}
+
+    @classmethod
+    def add_default_kernel_package(cls, state: MkosiState, extra_packages: Set[str]) -> None:
```
DaanDeMeyer, line 13: We can easily do this calculation before calling a shared install function and pass the requested default kernel package as an argument.

mkosi/distributions/ubuntu.py
DaanDeMeyer: In fact, I don't see why we can't do this outside of the shared install function. If no local mirror is specified, write the auxiliary repos before calling the shared install function. If one is specified, write the shared repos after calling the shared install function.

mkosi/distributions/ubuntu.py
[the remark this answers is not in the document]
DaanDeMeyer: I'm not convinced. I think all the things we're overriding can easily be done locally in the ubuntu class before or after calling a shared install function. See my other comments

mkosi/distributions/__init__.py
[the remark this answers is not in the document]
keszybz: Yeah, I think `from __future__ import annotations` is the way to go.

mkosi/backend.py
```diff
@@ -141,33 +141,34 @@ def __str__(self) -> str:
...
+    opensuse = "opensuse", PackageType.rpm
+    mageia = "mageia", PackageType.rpm
+    centos = "centos", PackageType.rpm
+    centos_epel = "centos_epel", PackageType.rpm
+    openmandriva = "openmandriva", PackageType.rpm
+    rocky = "rocky", PackageType.rpm
+    rocky_epel = "rocky_epel", PackageType.rpm
+    alma = "alma", PackageType.rpm
+    alma_epel = "alma_epel", PackageType.rpm
+    gentoo = "gentoo", PackageType.ebuild
```
keszybz, line 165: Hmm, would it be possible to align the names at the first `"`? Without that, it's really hard to scan the list. And also the second item column too…

mkosi/distributions/__init__.py
```diff
@@ -0,0 +1,21 @@
+# SPDX-License-Identifier: LGPL-2.1+
+
+from pathlib import Path
+from typing import TYPE_CHECKING, List
+
+if TYPE_CHECKING:
+    from ..backend import MkosiState
```
keszybz, line 7: Hmm, I think the general consensus nowadays is to not use relative imports without a good reason. `from mkosi.backend` is much clearer…

mkosi/__init__.py
```diff
@@ -1611,7 +1611,9 @@ def mount_api_vfs(root: Path) -> Iterator[None]:
 
 @contextlib.contextmanager
 def mount_cache(state: MkosiState) -> Iterator[None]:
-    if state.config.distribution in (Distribution.fedora, Distribution.mageia, Distribution.openmandriva):
+    if state.installer is not None:
```
keszybz, line 1377: Just `state.installer` is good enough. This is a normal object which will never be false, so checking for None is just unnecessary pedantry.

(Also below.)

mkosi/__init__.py
```diff
@@ -1611,7 +1611,9 @@ def mount_api_vfs(root: Path) -> Iterator[None]:
...
+        cache_paths = state.installer.cache_path()
+    elif state.config.distribution in (Distribution.fedora, Distribution.mageia, Distribution.openmandriva):
```
keszybz, line 1379: () → {}  ?

mkosi/backend.py
```diff
@@ -972,3 +975,56 @@ def complete_step(cls, text: str, text2: Optional[str] = None) -> Iterator[List[
...
+    """Recursively walk the tree starting at @root, optionally apply filter, yield non-none values"""
+    queue: Deque[Union[str, Path]] = collections.deque([root])
+
+    while queue:
+        for entry in os.scandir(queue.pop()):
+            pred = filter(entry) if filter is not None else entry
+            if pred is not None:
+                yield cast(T, pred)
+            if entry.is_dir(follow_symlinks=False):
+                queue.append(entry.path)
```
keszybz, line 1030: I think it'd be nicer to create some `utils.py` or `helpers.py` file, for all those standalone generic python functions that are very generic. `backend.py` has become another dumping ground, and we should instead move things into smaller files divided by function. We discussed this before and I think this idea was generally acceptable.

mkosi/install.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+def copy_fd(oldfd: int, newfd: int) -> None:
+    try:
+        reflink(oldfd, newfd)
+    except OSError as e:
+        if e.errno not in {errno.EXDEV, errno.EOPNOTSUPP, errno.ENOTTY}:
+            raise
+        # While mypy handles this correctly, Pyright doesn't yet.
+        shutil.copyfileobj(open(oldfd, "rb", closefd=False), cast(Any, open(newfd, "wb", closefd=False)))
```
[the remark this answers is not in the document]
keszybz: Would it be possible to kill this warning altogether for files opened in read mode?

mkosi/distributions/__init__.py
```diff
@@ -19,3 +19,7 @@ def kernel_image(name: str) -> Path:
     @classmethod
     def cache_path(cls) -> List[str]:
         raise NotImplementedError
+
+    @classmethod
+    def remove_packages(cls, state: "MkosiState", remove: List[str]) -> None:
+        pass
```
keszybz, line 27: Shouldn't this throw an exception if `remove` is not empty? That way RemovePackages= will be refused if not implemented for a distro.

mkosi/distributions/__init__.py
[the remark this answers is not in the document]
behrmann: I've had a look at this, and I don't think so. That's a PEP that's been punted down the line for the past two Python releases and this week's release says [that it might not be the future](https://docs.python.org/3.11/whatsnew/3.11.html#pep-563-may-not-be-the-future), so we shouldn't depend on something that might be different than what might be the default in the future.

mkosi/distributions/ubuntu.py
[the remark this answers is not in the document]
keszybz: I'm not convinced either. But maybe let's get this merged in this approximate form, and then maybe refactor later. How things are organized within the debian and ubuntu modules shouldn't impact the rest of the code, and we shouldn't get bogged down by this part.

keszybz (changes requested):

DaanDeMeyer:
I'd prefer delaying the merge until all distributions are moved to the new model. This is just refactoring anyway and doesn't introduce any new features, so there's no need to merge it before the release

keszybz:
So… what's the status here? Can we agree on some game plan and start moving forward?

I think we shouldn't block this on being "perfect", because that'll just take forever. Mkosi simply has now too much code to convert everything in one PR. Thus, I think we shouldn't require code for all distributions to be converted at once. But I'm fine with other decisions. But let's decide *something*.

behrmann:
`$DAYJOB` was quite busy the past few weeks, but it's finally on my todo list for today to fix the conflicts (and address the comments) that have accrued in the intervening weeks.

behrmann:
That's the rebase, addressing the comments will have to wait for tomorow.

mkosi/distributions/__init__.py
behrmann: Changed that to a `NotImplementedError`.

mkosi/distributions/arch.py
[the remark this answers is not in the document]
behrmann: I decided against this for now, since this PR is already unwieldy enough and I think it makes it easier to compare without the additional change in indentation.

mkosi/backend.py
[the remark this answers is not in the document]
behrmann: It's bit less complex now, since the extra sentinel is gone, now we just dynamically load the the installer from the module. This is necessary, because otherwise there would be an import loop, since the distribution modules always load at least `MkosiState` from `backend.py`.

mkosi/backend.py
behrmann: Agreed, which is why I moved a few things to new files already, but I think dissolving this dump is best left to followup work, since this PR is too big by half already.

behrmann:
I've finally moved all distros into submodules and I think `DistributionInstaller` gives a pretty mnimal interaction surface to the rest of the code now. Everything that's not in `DistributionInstaller` is now marked as "private" with an initial underscore. Inheritance is kept to a minimum, and all inheritance only affects distro internal stuff, which are always only private functions. 

I specifically did not inline all code, because - at least for me -  it makes it easier to compare the code.

There's still a lot of room for improvement and there's distro specific stuff scattered in many places, which I want to remove, but as @keszybz said, this should be a start.

DaanDeMeyer:
I'll take a look when I can but I can already say one of us is going to have a lot of fun resolving merge conflicts

behrmann:
Please, I've rebased this one and its predecessors enough already ;)

mkosi/backend.py
DaanDeMeyer: These are still here?

mkosi/distributions/__init__.py
DaanDeMeyer: This isn't fixed?

mkosi/distributions/__init__.py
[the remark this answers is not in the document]
DaanDeMeyer: Sure, it might not be the future, but whatever solution they come up with will likely not be all that different in usage, so I say we use `from __future__ import annotations` for now and deal with that problem later

DaanDeMeyer (changes requested):
Seems you resolved a bunch of conversations but the issues are not fixed? Did you perhaps forget to push some changes?

mkosi/distributions/__init__.py
behrmann: I'm not sure I follow. The diff is marked as outdated. I fixed that in the commit "Make all imports absolute" and in the version I see here ([b8580b7](https://github.com/systemd/mkosi/pull/1217/commits/b8580b79fc15031cee82efb4e8963aed52e2cb36)), which is the same as I have locally, the import is `mkosi.backend`.

mkosi/backend.py
behrmann: On my side this diff too is marked as outdated and I don't see these newlines in [b8580b7](https://github.com/systemd/mkosi/pull/1217/commits/b8580b79fc15031cee82efb4e8963aed52e2cb36) anymore.

mkosi/distributions/__init__.py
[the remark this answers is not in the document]
behrmann: Ah, since there was no reaction anymore I thought this one has runs it's course, sorry.

I disagree. I don't think we should `from __future__ import annotation`. This feature is opt and still very much in flux  or under consideration upstream and has fallen so much out of favour that they've added the above-linked note in the release notes after punting on accepting the PEP for two releases. Opting into this, I think we would rely on an unstable feature for, what I think is, mostly aesthetic gain. I think it's saver to stay with the language defaults and wait this out.

mkosi/distributions/__init__.py
DaanDeMeyer: Ahh, now it makes sense. In systemd contribution guidelines we always things to be fixed in the commit that introduces them instead of in a later commit, can you do that here as well? Makes things easier to review commit by commit

behrmann: Fair enough, I can rebase that once more, but since this was a thing used all over the codebase, i.e. existing code, I thought it made more sense to keep the relative commits around everywhere and have one point where they are all switched over.

mkosi/distributions/__init__.py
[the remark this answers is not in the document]
DaanDeMeyer: I'm fine with discussing that in a separate PR, but we already use `from __future__ import annotations` in our other code. So I'd still prefer that in this PR, we stick to that and then in another PR you can remove all uses of this and we can discuss it there.

But we shouldn't do this switch in the middle of another large PR, let's keep this one focused on the split

DaanDeMeyer:
Wayy too much implementation inheritance if you ask me, but I want to make progress, so I'm merging this, I'll do a follow up PR at some point to address the stuff I disagree with

mkosi/distributions/__init__.py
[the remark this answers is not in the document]
behrmann: Agreed, opened #1280 to discuss dropping the rest, which I had actually missed. The diff is rather small.

PR #1222, opened 2022-10-07
Insert partition without using sendfile dues to transfer limits

### LINKED ISSUE #1221: Inserted partition is too small
When a partition is inserted into a GPT image, the inserted partition **contents** does not match the contents of the inserted partition. The partition size is big enough, but the data copied in is too small. This can be triggered when either:

  * The `UsrOnly` option is set
  * The `Minimize` option is set
  * SquashFS is used

I've tracked this down to the use of `os.sendfile()` which is only transferring 2147479552 bytes (2GB - 1020KB). I then found [this bug](https://github.com/python/cpython/issues/79360), especially [this](https://github.com/python/cpython/issues/79360#issuecomment-1093804149) reply.

Test code below.

```python
from os import sendfile

f_in = open('testfile', 'r+b')
f_out = open('/dev/null', 'wb')
for size_gb in range(8):
    _ = f_in.seek(0)
    size_bytes = (size_gb+1) * 2 ** 30
    print(f'Truncated to {f_in.truncate(size_bytes)} bytes')
    size_bytes = sendfile(f_out.fileno(), f_in.fileno(), offset=0, count=size_bytes)
    print(f'Transferred  {size_bytes} bytes')
```

Test output.
```
Truncated to 1073741824 bytes
Transferred  1073741824 bytes
Truncated to 2147483648 bytes
Transferred  2147479552 bytes
Truncated to 3221225472 bytes
Transferred  2147479552 bytes
Truncated to 4294967296 bytes
Transferred  2147479552 bytes
Truncated to 5368709120 bytes
Transferred  2147479552 bytes
Truncated to 6442450944 bytes
Transferred  2147479552 bytes
Truncated to 7516192768 bytes
Transferred  2147479552 bytes
Truncated to 8589934592 bytes
Transferred  2147479552 bytes
```

This explains why I was getting issues with the verity partition data.

mcassaniti:
Fixes #1221 using `shutil.copyfileobj()`

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3665,7 +3665,15 @@ def insert_partition(
 
             # Without this the entire blob will be read into memory which could exceed system memory
             with open(path, mode='wb') as path_fp:
-                os.sendfile(path_fp.fileno(), blob.fileno(), offset=0, count=blob_size)
+                # Chunk size for 32/64-bit systems
+                # Chunking required because sendfile under Linux has a maximum copy size
+                chunksize = 2 ** 30 if sys.maxsize < 2 ** 32 else 0x7ffff000
+                offset = 0
+                while True:
+                    sent = os.sendfile(path_fp.fileno(), blob.fileno(), offset=offset, count=chunksize)
+                    if not sent:
+                        break
+                    offset += sent
 
     return part
 


behrmann:
Great find!

I think this is an ok change to merge, although it could maybe have a bit worse performance. `shutil.copyfileobj` is pretty simple code:
```python
def copyfileobj(fsrc, fdst, length=0):
    """copy data from file-like object fsrc to file-like object fdst"""
    if not length:
        length = COPY_BUFSIZE
    # Localize variable access to minimize overhead.
    fsrc_read = fsrc.read
    fdst_write = fdst.write
    while True:
        buf = fsrc_read(length)
        if not buf:
            break
        fdst_write(buf)
```
which is apparently frozen in time so as not to generate any backwards incompatibilities. There's `shutil.copyfile`, which under the hood tries to use `sendfile` if possible, but it only takes paths. Not wanting to use internal helper functions, I guess we could just do the chunking ourselves.

The CI failure is unrelated.

DaanDeMeyer:
What about calling sendfile in a loop? Shouldn't that be possible?

behrmann:
Yes, with an appropriate chunk size and counting up the starting point for sendfile and breaking if nothings send anymore (the return from sendfile is 0). That's what the fastcopy implementation in cpython does.

mcassaniti:
I've just pushed an update. To prove it actually works I've got the following test code below.

```python
from os import sendfile

f_in = open('test.in', 'x+b')
f_out = open('test.out', 'wb')
size = 8 * 2 ** 30
f_in.truncate(size)
for offset in range(0, size, 0x7ffff000):
    print(sendfile(f_out.fileno(), f_in.fileno(), offset=offset, count=0x7ffff000))
f_out.tell() == size


```

Output
```
2147479552
2147479552
2147479552
2147479552
16384
True
```

mkosi/__init__.py
```diff
@@ -3665,7 +3665,9 @@ def insert_partition(
 
             # Without this the entire blob will be read into memory which could exceed system memory
             with open(path, mode='wb') as path_fp:
-                os.sendfile(path_fp.fileno(), blob.fileno(), offset=0, count=blob_size)
+                # Chunking required because sendfile under Linux has a maximum copy size
+                for offset in range(0, blob_size, 0x7ffff000):
+                    os.sendfile(path_fp.fileno(), blob.fileno(), offset=offset, count=0x7ffff000)
```
behrmann, line 3670: Looking at the cpython implementation this seems to have overflow issues on 32bit architectures, which is why they choose either the file size as the block size or `2**30`, whichever is smaller.

mcassaniti: Is it the offset or the count that overflows under 32-bit? Either way I'm open to suggestions if you send me through an example.

behrmann: It's the [count](https://github.com/python/cpython/blob/main/Lib/shutil.py#L135)

behrmann: Exactly. I think going with a default block size of `0x7ffff000` is fine, since sendfile won't do more than that anyway, it only precludes automatically picking up a larger block size if sendfile ever starts doing that. I guess that's why they go with a larger block size in the first place. It only means you need to to the while loop and look at sendfile's return.

I think we should do the cap to `2**30` for 32 bit systems, too.

mcassaniti: Why not make our lives easier knowing that we can't send more than `0x7ffff000` bytes regardless of 32 or 64 bit architecture and just cap at `2**30`? The difference is about 1MiB (i.e.: a 1MiB inefficiency).

behrmann: Isn't the difference a factor of two and about 1GiB?

mcassaniti: You're right, it is about 1GiB. Whoops on my part. Looking harder I'm not convinced we need to set the count to `2**30` on 32-bit systems. It seems it just needs to be below `2**31- 1` for 32-bit.

From https://bugs.python.org/issue38319 which was mentioned in `_fastcopy_sendfile()`:

*I understand that the error comes from the 4th sendfile() parameter: "count". The C code (of the Linux implementation) uses the "n" format for Py_ssize_t: [Python/getargs.c](https://github.com/python/cpython/blob/master/Python/getargs.c) calls PyLong_AsSsize_t().*

*On a 64-bit system, it's less likely to reach Py_ssize_t maximum value (max = `2**63-1`), but it's easy to reach on a 32-bit system (max = `2**31-1`).*

mcassaniti: @behrmann Did you get a change to review further?

behrmann: Sorry forgot to answer. I guess we should err on the same size and mirror cpython here, e.g.
```python
chunksize = 0x7ffff000 if is_64bit() else 2**30
offset = 0
while True:
    sent = os.sendfile(path_fp.fileno(), blob.fileno(), offset=offset, count=chunksize)
    if not sent:
        break
    offset += sent
```
for some not yet existing function `is_64bit`

mkosi/__init__.py
```diff
@@ -3665,7 +3665,15 @@ def insert_partition(
 
             # Without this the entire blob will be read into memory which could exceed system memory
             with open(path, mode='wb') as path_fp:
-                os.sendfile(path_fp.fileno(), blob.fileno(), offset=0, count=blob_size)
+                # Chunk size for 32/64-bit systems
+                # Chunking required because sendfile under Linux has a maximum copy size
+                chunksize = 2 ** 30 if sys.maxsize < 2 ** 32 else 0x7ffff000
```
DaanDeMeyer, line 3670: Shouldn't this be `<=`?

mcassaniti: I copied this from the shutil code [here](https://github.com/python/cpython/blob/main/Lib/shutil.py#L138), but according to the official documentation it should really be `< 2 ** 31` since `sys.maxsize = 2 ** 31 - 1` for 32-bit systems (see [here](https://docs.python.org/3/library/sys.html#sys.maxsize)).

DaanDeMeyer (approved):

mcassaniti:
@DaanDeMeyer  Seems like were now good for a merge.

PR #1225, opened 2022-10-09
initrd can be versioned in Debian

bluca:
In Bookworm the initrd is installed with the same version suffix as the directory, for reasons. Check for both.


behrmann:
Generally LGTM, but would it maybe make more sense to *prefer* the versioned initrd if it exists? My thinking is that less information / less specific versions make me think more of fallback.

bluca:
> Generally LGTM, but would it maybe make more sense to _prefer_ the versioned initrd if it exists? My thinking is that less information / less specific versions make me think more of fallback.

sure, updated

behrmann (approved):

PR #1226, opened 2022-10-11
Copy also .git file for submodules

fwiesel:
To properly get the commit history of a submodule, we need the .git file pointing to the submodule
in the top-level git directory


mkosi/__init__.py
```diff
@@ -3303,6 +3303,12 @@ def copy_git_files(src: Path, dest: Path, *, source_file_transfer: SourceFileTra
         files |= {sm / x.decode("utf-8") for x in c.stdout.rstrip(b"\0").split(b"\0")}
         files -= submodules
 
+        # Add the .git submodule file well.
+        if source_file_transfer == SourceFileTransfer.copy_git_more:
+            fp = os.path.join(str(src / sm), ".git")
+            fr = fp[len(str(src))+1 :]
+            files.add(fr)
```
behrmann, line 3310: The first line of this block would be better as
```python
fp = src / sm / ".git"
```
but I gather you just want
```suggestion
            files.add(os.path.join(sm, ".git"))
```
or
```suggestion
            files.add(f"{sm}/.git")
```
no?

behrmann (changes requested):

mkosi/__init__.py
fwiesel: Ah, indeed. I went with the first suggestion.

mkosi/__init__.py
```diff
@@ -3303,6 +3303,10 @@ def copy_git_files(src: Path, dest: Path, *, source_file_transfer: SourceFileTra
         files |= {sm / x.decode("utf-8") for x in c.stdout.rstrip(b"\0").split(b"\0")}
         files -= submodules
 
+        # Add the .git submodule file well.
```
behrmann, line 3306:
```suggestion
        # Add the .git of the submodule as well.
```

mkosi/__init__.py
```diff
@@ -3303,6 +3303,10 @@ def copy_git_files(src: Path, dest: Path, *, source_file_transfer: SourceFileTra
...
+        if source_file_transfer == SourceFileTransfer.copy_git_more:
+            files.add(os.path.join(sm, ".git"))
```
behrmann, line 3308: Have you tested that this works for you? I just had a minute to look further up in the function and saw the dance we do to include the files in the `.git/` of the parent; so just to make sure, have you tested that the contents are included here?

fwiesel: Well, the dance above is necessary because it includes all the files in the `.git/` directory of the checkout, while I have a submodule, and that is just a single file pointing to a directory within the `.git/` directory already copied before.

I was trying to do a `pip install <path-to-submodule>` and that was breaking because it couldn't get the version from git.
With this change, it works.

mkosi/__init__.py
behrmann: Ah, right, submodule, nevermind then.

PR #1227, opened 2022-10-14
Don't cache bootloader installation

DaanDeMeyer:
bootctl might configure itself differently depending on what's on the rest of the system. If we run it as part of the cached image build, we're doing the installation with files missing that might influence the end result, so let's not do the installation as part of the cached image build.

We pass --quiet to avoid extra noise when doing incremental builds.


behrmann:
LGTM, but did you forget the `--quiet` or am I missing something?

DaanDeMeyer:
Ah no it didn't work, it's very new in bootctl

PR #1231, opened 2022-10-19
Adding custom retry amount to Machine.run()

gsegatti:
I've recently noted the retry logic for `run()` within `machine.py` has a fixed value of 30 retries and a 1 second sleep between each iteration.

I'd like to propose the possibility of letting the user set this amount via the `Machine` constructor. This way an user can await for a lesser time span between trying to run a `Machine` instance.

Please let me know if this would be something that makes sense as my last PR was quite a few months ago.

### DIFF
--- a/mkosi/machine.py
+++ b/mkosi/machine.py
@@ -48,9 +48,10 @@ class LogfileAdapter:
 
 
 class Machine:
-    def __init__(self, args: Sequence[str] = []) -> None:
+    def __init__(self, args: Sequence[str] = [], *, num_retries: int = 30) -> None:
         # Remains None until image is built and booted, then receives pexpect process.
         self._serial: Optional[pexpect.spawn] = None
+        self.num_retries = num_retries
         self.exit_code: int = -1
         self.stack = contextlib.ExitStack()
         self.config: MkosiConfig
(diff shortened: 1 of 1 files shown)


mkosi/machine.py
```diff
@@ -48,9 +48,10 @@ def flush(self) -> None:
 
 
 class Machine:
-    def __init__(self, args: Sequence[str] = []) -> None:
+    def __init__(self, args: Sequence[str] = [], retry_amount: int = 30) -> None:
```
behrmann, line 51:
```suggestion
    def __init__(self, args: Sequence[str] = [], *, retry_amount: int = 30) -> None:
```
I'd prefer to make keyword arguments with defaults to only be callable via keyword

behrmann (commented):
LGTM, with one nit. Would be nice to plumb this through to pytest in the machine test.

mkosi/machine.py
DaanDeMeyer: Let's rename it to `num_retries` as well

gsegatti: Yep, via keyword seems better. Regarding your other comment, I didn't quite understand what "plumb through" the `num_retries` to pytest would be, so I tried to implement what I thought correct.

Let me know if I got it wrong.

behrmann: Sorry for being unclear: What I meant is to be able to run pytest such, that i can set `num_retries` on the CLI.

DaanDeMeyer (approved):

PR #1236, opened 2022-10-23
mkosi: add dracut extra modules

johannbg:
Adds the ability to include additional dracut modules to mkosi

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -138,6 +138,7 @@ DRACUT_SYSTEMD_EXTRAS = [
     "/usr/bin/systemd-tty-ask-password-agent"
 ]
 
+DRACUT_EXTRA_MODULES: List[str] = []
 
 def write_resource(
         where: Path, resource: str, key: str, *, executable: bool = False, mode: Optional[int] = None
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -138,6 +138,7 @@
     "/usr/bin/systemd-tty-ask-password-agent"
 ]
 
+DRACUT_EXTRA_MODULES = [] # type: List[str]
```
behrmann, line 141:
```suggestion
DRACUT_EXTRA_MODULES: List[str] = [] 
```

behrmann (commented):
I guess it's not a problem for dracut if there's overlap between files installed via modules and `install_optional_items` (assuming that the module is purely additive in nature)? i.e. we can keep `DRACUT_SYSTEMD_EXTRAS` around and add modules that do the same thing to `DRACUT_EXTRA_MODULES` and we're good with both older and recent dracut releases?

mkosi/__init__.py
behrmann: Looking at the previous version of this PR, you did `list[str]` instead of `List[str]`, we don't do that yet, since we're still lower-bounded to Python 3.7 (I need to push my 3.9 branch), so that's why that didn't work. `List[str]` as annotation works.

> If you want to support older dracut releases as well it should be fine but we dont support downstream playing russian release roulette with components. Using new dracut with old systemd etc.

Wanting and having to live with what's generally available are two different things. :)

behrmann (commented):
Two questions:
- How does dracut handle it if a module is not available?
- Could you fill out the `DRACUT_EXTRA_MODULES` section? Little point in adding scaffolding if it's not being used.

PR #1237, opened 2022-10-24
dnf: Disable versionlock plugin

ignatenkobrain:
The host's versionlock plugin / setup will exclude stuff that one tries to get into the image… Most of the other tools (mock, kiwi) are doing it this way.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1957,6 +1957,7 @@ def invoke_dnf(state: MkosiState, command: str, packages: Iterable[str]) -> None
         "--allowerasing",
         f"--releasever={release}",
         f"--installroot={state.root}",
+        "--disableplugin=versionlock",
         "--setopt=keepcache=1",
         "--setopt=install_weak_deps=0",
     ]


DaanDeMeyer:
We should probably just use `--noplugins` to disable all plugins.

ignatenkobrain:
I'd say there are some plugins that are actually useful (e.g. the cow stuff from Facebook folks)

DaanDeMeyer:
Sure but users should explicitly enable those. The idea is to minimize the effect of host specific configuration on image building. We can add a separate setting to enable specific plugins once someone is interested.

behrmann (approved):

ignatenkobrain:
So do you want me to update it to use --noplugins instead?

keszybz:
I think it's better to disable all plugins. After all, we generate a new config file, so it also makes sense to disable plugins. Plugins can be of two kinds: those that provide some additional command that does nothing unless invoked, things like download/repoclosure/repodiff/repograph/reposync/repomanage/copr/debuginfo-install/config-manager/builddep/changelog/show-leaves/showvars/system-upgrade, and plugins that influence behaviour, like cow/versionlock/subscription-manager/tracer. There are more of the first kind, and they shouldn't cause any problem, but they are also not used, since we don't invoke them. OTOH, the second kind could cause problems and/or unexpected behaviour. I think we want to disable all plugins by default, and then maybe add a mechanism to enable selected ones if people find some use. Thus, please update to `--noplugins`.

DaanDeMeyer:
Replaced by #1242

PR #1240, opened 2022-10-26
Use the detected distribution version

keszybz:
I used mkosi without any explicit distro/version parameters and was a bit surprised that it built for Fedora 36 by default after successfully detecting that it's running on Fedora 37. I don't think this makes much sense. Let's use the same distro+version combination as what is detected. This is more likely to be what users expect, and also avoids the awkward issue that when we hardcode some default version, this version is likely to be "wrong" (not the latest) until both the distro and we make a release.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -44,6 +44,8 @@
 - Split dm-verity artifacts default names have been changed to match what
   `systemd` and other tools expect: `image.root.raw`, `image.root.verity`,
   `image.root.roothash`, `image.root.roothash.p7s` (same for `usr` variants).
+- `mkosi` will again default to the same OS release as the host system when the
+  host system uses the same distribution as the image that's being built.
 
 ## v13
 
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -5864,6 +5864,9 @@ def load_distribution(args: argparse.Namespace) -> argparse.Namespace:
         if args.distribution is None:
             args.distribution = d
 
+        if args.distribution == d and args.release is None:
+            args.release = r
+
     if args.distribution is None:
         die("Couldn't detect distribution.")
 


DaanDeMeyer:
I removed this because it means the image distro version that gets built can depend on the host that is being built from. If you build on the same distro, the version will suddenly be different than if you're building from another distro. I don't think we should discard the consistency that the version that gets built is always the same.

keszybz:
If we're using autodetection, the image that gets built already depends on the host. But we're making a somewhat arbitrary choice to detect the host distro, but for the version, use something unrelated. I think that this results in very surprising behaviour, it certainly "broke" mkosi-initrd. (Things still boot with a mismatched distro version, but that's not what is wanted.)

To look at it from another side, with some constant config that doesn't specify the distro version, we have three moving parts: host distro version, default version in mkosi, and latest distro version. On non-rolling traditional distros, users expect changes in behaviour when upgrading between distro versions. Noticeable changes in behaviour are unexpected and discouraged in updates to stable distro release, including even stuff like graphical design and UI changes that don't really "matter", see e.g. https://docs.fedoraproject.org/en-US/fesco/Updates_Policy/#stable-releases. This means that users generally learn to expect that if they upgrade the distro version, they may get a new version of gcc with new warnings, or a different look of the graphical interface, etc. But when doing updates within a given distro version, things should slowly improve, but not "change". For various fast-moving projects, like mkosi, it makes a lot of sense to stretch this policy a bit, and release new versions of mkosi in stable distro releases, because it needs to adjust to changes in other distributions and in general we fix many bugs over time. In a stable release of a distro, we would generally expect the default distro version to change in mkosi when the host distro version changes, similarly to `dnf` and various other tools. If it is embedded in `mkosi`, then we could never upgrade `mkosi` in a stable update. (Also, tying the default distro version in mkosi to the third moving part, i.e. the latest released version of the distro would be unexpected because it'd create changes in behaviour dependent on completely external state. The exception to this would if there's an explicit opt-in, e.g. `--version=latest`.)

Current behaviour also leads to strange behaviours like building an *older* version of a distro than the host.

I'll update my commit to say that it is a revert.

keszybz:
I restored previous logic and updated the commit description.

DaanDeMeyer (approved):

Issue #1238, opened 2022-10-25
ssh file not found authorized_keys


williamcroberts:
This was discovered on Version 13, but appears to be fixed on master. I just wanted some archive in case others have issues, but this ticket could likely just be closed.

When attempting to copy the public key to the remote images authorized_keys file, the parent path does not exist and the following    exception is thrown[1]:
```
    ‣  Generating SSH key pair…
    ‣  (Unmounting image)
    ‣  (Detaching /dev/loop17)
    Traceback (most recent call last):
      File "/home/wcrobert/workspace/mkosi/mkosi/__init__.py", line 7357, in setup_ssh
        copy_file(f"{f.name}.pub", authorized_keys)
      File "/home/wcrobert/workspace/mkosi/mkosi/__init__.py", line 614, in copy_file
        with open_close(newpath, os.O_WRONLY | os.O_CREAT | os.O_EXCL, st.st_mode) as newfd:
      File "/usr/lib/python3.8/contextlib.py", line 113, in __enter__
        return next(self.gen)
      File "/home/wcrobert/workspace/mkosi/mkosi/__init__.py", line 571, in open_close
        fd = os.open(path, flags | os.O_CLOEXEC, mode)
    FileNotFoundError: [Errno 2] No such file or directory: '/var/tmp/mkosi-_0v_4emp/root/root/.ssh/authorized_keys'
   ``` 
The path only existed up to "/var/tmp/mkosi-_0v_4emp/root", thus a mkdir
with -p semantics is required to create the full path, add that in.
    
[1] Note line numbers are off due to having some print's scattered
through the code for debugging.

This was my patch for v13:
```diff
diff --git a/mkosi/__init__.py b/mkosi/__init__.py
index 4419975267f7..213d77cc2134 100644
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -7099,6 +7099,7 @@ def configure_ssh(state: MkosiState, cached: bool) -> Optional[TextIO]:
                 stdout=subprocess.DEVNULL,
             )
 
+        os.makedirs(os.path.dirname(authorized_keys), exist_ok=True)
         copy_file(f"{f.name}.pub", authorized_keys)
         os.remove(f"{f.name}.pub")
```

williamcroberts:
@behrmann added PR and change you suggested. Seems fine locally :-p

PR #1241, opened 2022-10-26
ssh: fix copy_file for authorized_keys

williamcroberts:
When attempting to copy the public key to the remote images authorized_keys file, the parent path does not exist and the following exception is thrown[1]:
‣  Generating SSH key pair…
‣  (Unmounting image)
‣  (Detaching /dev/loop17)
  File "/home/wcrobert/workspace/mkosi/mkosi/__init__.py", line 7357, in setup_ssh
    copy_file(f"{f.name}.pub", authorized_keys)
  File "/home/wcrobert/workspace/mkosi/mkosi/__init__.py", line 614, in copy_file
    with open_close(newpath, os.O_WRONLY | os.O_CREAT | os.O_EXCL, st.st_mode) as newfd:
  File "/usr/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/home/wcrobert/workspace/mkosi/mkosi/__init__.py", line 571, in open_close
    fd = os.open(path, flags | os.O_CLOEXEC, mode)
FileNotFoundError: [Errno 2] No such file or directory: '/var/tmp/mkosi-_0v_4emp/root/root/.ssh/authorized_keys'

The path only existed up to "/var/tmp/mkosi-_0v_4emp/root", thus a mkdir with -p semantics is required to create the full path, add that in.

[1] Note line numbers are off due to having some print's scattered through the code for debugging.

Fixes: #1238

Signed-off-by: William Roberts <william.c.roberts@intel.com>

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -7099,6 +7099,7 @@ def configure_ssh(state: MkosiState, cached: bool) -> Optional[TextIO]:
                 stdout=subprocess.DEVNULL,
             )
 
+        authorized_keys.parent.mkdir(parents=True, exist_ok=True)
         copy_file(f"{f.name}.pub", authorized_keys)
         os.remove(f"{f.name}.pub")
 


behrmann:
LGTM, although thinking about it, maybe this shouldn't even have `parents=True`, since if `/root` doesn't exist something's fishy anyway, so we only need to create a single directory and don't need `mkdir -p` semantics.

williamcroberts:
> LGTM, although thinking about it, maybe this shouldn't even have `parents=True`, since if `/root` doesn't exist something's fishy anyway, so we only need to create a single directory and don't need `mkdir -p` semantics.

You need 2 directories, hence `-p` semantics. Mine was missing `root/.ssh/` of  `/root/root/.ssh/`. I tried to get that wording into the commit message as, "The path only existed up to "/var/tmp/mkosi-_0v_4emp/root" and the stack trace shows, "/var/tmp/mkosi-_0v_4emp/root/root/.ssh/authorized_keys"

PR #1246, opened 2022-10-27
Print summary to stdout, use colour

keszybz:


DaanDeMeyer (approved):

mkosi/__init__.py
```diff
@@ -6791,24 +6786,62 @@ def cache_image_path(config: MkosiConfig, is_final_image: bool) -> Optional[Path
...
+            config.output_checksum if config.checksum else None,
+            config.output_signature if config.sign else None,
+            config.output_bmap if config.bmap else None,
+            config.output_nspawn_settings if config.nspawn_settings is not None else None,
+            config.output_root_hash_file if config.verity else None,
+            config.output_sshkey if config.ssh else None,
+            config.output_split_root if config.split_artifacts else None,
+            config.output_split_verity if config.split_artifacts else None,
+            config.output_split_verity_sig if config.split_artifacts else None,
+            config.output_split_kernel if config.split_artifacts else None,
```
behrmann, line 6843: Is there a reason to add indentation here? It's a lot of churn for at most a small gain in readability.

mkosi/__init__.py
```diff
@@ -6791,24 +6786,62 @@ def cache_image_path(config: MkosiConfig, is_final_image: bool) -> Optional[Path
...
+        raise OSError(errno.ENOENT, 'Not a normal file')
+    if not os.access(path, os.X_OK):
+        raise OSError(errno.ENOENT, 'Not executable')
+    return None
+
+
+def check_inputs(config: MkosiConfig) -> None:
+    try:
+        for tree in (config.skeleton_trees,
+                     config.extra_trees):
```
behrmann, line 6814:
```suggestion
        for tree in (config.skeleton_trees, config.extra_trees):
```
since the line is very short already

behrmann (commented):
Mostly LGTM. The first commit in the series makes a lot of churn and apart from that we did get all prints in mkosi to go through this class (except for some newer cases where we did drop the ball a bit), but it makes sense to print to stdout here and there is little gain to add basically a normal print to `MkosiPrinter` and use that. That's just the same thing with additional steps.

Some style nits, since we always disagree there. :)

behrmann:
CI failure is unrelated

mkosi/__init__.py
keszybz: Hmm, this is on separate lines on purpose, to match the style of following loop. They both have lists of config options to go over, and I think it's nice to have the same structure.

mkosi/__init__.py
keszybz: I dropped this chunk now.

keszybz:
All CI failures are network-related. I dropped that chunk that @behrmann didn't like, so I think this can be merged.

PR #1247, opened 2022-10-28
Allow --compress=no with squashfs images

reubeno:
Running `mkosi --format=plain_squashfs --compress=no` yields this error:

```
‣ Error: Cannot disable compression with squashfs
```

The argument parsing logic blocks this case even though `generate_squashfs` handles the case correctly and `--compress-fs=no` would have worked too. Testing with this change, the above command produces an uncompressed squashfs image:

```
Found a valid SQUASHFS 4:0 superblock on uncompressed.img.
Creation or last append time Thu Oct 27 17:08:06 2022
Filesystem size 128206861 bytes (125202.01 Kbytes / 122.27 Mbytes)
Compression gzip
Block size 131072
Filesystem is exportable via NFS
Inodes are uncompressed
Data is uncompressed
Uids/Gids (Id table) are uncompressed
Fragments are uncompressed
Always-use-fragments option is not specified
Xattrs are uncompressed
Duplicates are removed
Number of fragments 308
Number of inodes 4018
Number of ids 9
Number of xattr ids 0
```

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -6580,8 +6580,6 @@ def load_args(args: argparse.Namespace) -> MkosiConfig:
         if args.root_size is None:
             # Size will be automatic
             args.minimize = True
-        if args.compress is False:
-            die("Cannot disable compression with squashfs", MkosiNotSupportedException)
         if args.compress is None:
             args.compress = True
 
--- a/tests/test_parse_load_args.py
+++ b/tests/test_parse_load_args.py
@@ -135,6 +135,5 @@ def test_shell_boot() -> None:
 def test_compression() -> None:
     assert parse(["--format", "gpt_squashfs"]).compress
 
-    with pytest.raises(MkosiException, match=".*compression.*squashfs"):
-        parse(["--format", "gpt_squashfs", "--compress", "False"])
+    assert not parse(["--format", "gpt_squashfs", "--compress", "False"]).compress
 


behrmann (approved):
No idea why this was disabled in the first place. Maybe @keszybz knows, because he added the message (although this forcing compression predates his last changes).

reubeno:
I looked back through git history and found the commit that introduced the block/error: https://github.com/systemd/mkosi/commit/dc2e6b3b58ebc0bcb8b43b117e2de11087bf9b03 . Looking at the version of the code then, the `mksquashfs` logic couldn't yet handle disabling compression so the block would have made sense then. And it looks like the change that added support for building uncompressed squashfs images (https://github.com/systemd/mkosi/commit/18d428f309126fef1aa1621bc8dd02a00f502789) doesn't look like it adjusted the argument parsing; I can see how that wouldn't have been caught, since `--compress-fs=no` *does* work correctly without this PR.

As a side question -- I see a few checks failed, seemingly unrelated to my change. Looks like difficulty retrieving packages from an Internet source for an integration test? While I would generally prefer not to resort to it, is there a way to try re-running checks for potentially transient issues in those integration tests?

(And thanks for the approval! This is my first contribution to mkosi, and I appreciate any and all feedback :))

behrmann:
Ah, yes, that clears it up. I only looked at the blame for the code around the check.

I gather the CI errors are just transient issues. This happen once in a while for every distros. I can rerun the failed jobs if you like, but it's all stuff that is not touched by this PR, so I don't really see added value in burning more CPU time on them.

reubeno:
No, that's okay -- you don't need to re-run the failed distro checks. Just wasn't sure if it was strictly required to get all checks passing. Thanks!

reubeno:
Just checking in -- is there anything else I should be doing to ready this change for further review/merging? I understand y'all are probably busy, I just wanted to make sure you're not waiting on something from me here.

PR #1248, opened 2022-10-28
Make `load_args` sideeffect free

behrmann:
It is unnecessary for the find_foo functions in load_args to generate and chown directories, since for output and build directory this is already done by the make_foo_dir functions in build_stuff. At a similar function for the cache directory.

Having the find_foo functions make the directories has the downside of leading to permission errors, when the tests are run with fakeroot as done by many distro buildsystems.

/cc @qdeslandes

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -6256,8 +6256,6 @@ def find_output(args: argparse.Namespace) -> None:
     else:
         return
 
-    mkdirp_chown_current_user(args.output_dir, skip_chown=args.no_chown)
-
 
 def find_builddir(args: argparse.Namespace) -> None:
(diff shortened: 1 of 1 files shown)


DaanDeMeyer (approved):
LGTM

DaanDeMeyer:
Maybe we should also remove the fallback to os.getuid() in chown_current_user(). We only really want to go from root to non root when running as root, not in yhe the opposite direction.

dvzrv:
Thanks! That appears to fix the test suite for me in 14.

behrmann:
I just gave this a quick go and all directories generated in `mkosi.cache` and `mkosi.output` have my user, but the cached packages are still owned by root, but that's the same before this change. I guess we should chown the cached packages as well?

PR #1250, opened 2022-10-29
Add more Debian architecture mappings

bluca:


mkosi/__init__.py
```diff
@@ -411,25 +411,39 @@ def fedora_release_cmp(a: str, b: str) -> int:
...
+# uname -m -> dpkg --print-architecture
 DEBIAN_ARCHITECTURES = {
-    "x86_64": "amd64",
-    "x86": "i386",
     "aarch64": "arm64",
     "armhfp": "armhf",
+    "armv7l": "armhf",
+    "ia64": "ia64",
+    "mips64": "mipsel",
+    "m68k": "m68k",
```
behrmann, line 421: I only had a quick look, but m68k is not in den DPS, right? If it's not in there (yet?), maybe we should restrict ourselves to architectures that are.

behrmann (commented):
LGTM, two questions, about DPS coverage:
1. Is there a working loongarch port in Debian? If so, we should probably add that, too
2. The other about m68k is inline.

The rest should be covered by DPS, if I'm not missing anything.

mkosi/__init__.py
bluca: we can add it - but there's not that many architectures, and they rarely change, so I'd rather just have them all and be done with it, since it's just two tables so doesn't take that much effort

bluca:
> LGTM, two questions, about DPS coverage:
> 
> 1. Is there a working loongarch port in Debian? If so, we should probably add that, too
> 2. The other about m68k is inline.
> 
> The rest should be covered by DPS, if I'm not missing anything.

Work in progress, but no dpkg name yet, so can't add it https://wiki.debian.org/Ports/loongarch64
Same for `arc`, we know the dpkg name but not the kernel one

PR #1256, opened 2022-11-02
Add device mapper kernel modules to initrd when required

reubeno:
If dm-verity or dm-crypt is used, then ask `dracut` to confirm that the appropriate kernel modules are included in the initrd: `dm-mod`, `dm-crypt`, and/or `dm-verity`.

As one case, this resolves issues using `mkosi` to build an image with `--verity=yes` when the included kernel doesn't include `dm-mod` as builtin.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1694,6 +1694,20 @@ def configure_dracut(state: MkosiState, cached: bool) -> None:
             '[[ $(modinfo -k "$kernel" -F filename efivarfs 2>/dev/null) == /* ]] && add_drivers+=" efivarfs "\n'
         )
 
+    # Modules may be required for device mapper targets
+    dm_modules = []
+    if state.config.encrypt:
+        dm_modules.append("dm-crypt")
+    if state.config.verity:
+        dm_modules.append("dm-verity")
+    if dm_modules:
+        dm_modules.append("dm-mod")
+
+        conf_content = 'add_dracutmodules+=" dm "\n'
+        for dm_module in dm_modules:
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -1694,6 +1694,20 @@ def configure_dracut(state: MkosiState, cached: bool) -> None:
...
+    if state.config.encrypt:
+        dm_modules.append("dm-crypt")
+    if state.config.verity:
+        dm_modules.append("dm-verity")
+    if dm_modules:
+        dm_modules.append("dm-mod")
+
+        conf_content = 'add_dracutmodules+=" dm "\n'
+        for dm_module in dm_modules:
+            conf_content += f'[[ $(modinfo -k "$kernel" -F filename {dm_module} 2>/dev/null) == /* ]] && add_drivers+=" {dm_module} "\n'
```
behrmann, line 1708: Isn't that something dracut does under the hood?

behrmann: To rephrase this: apparently not, otherwise you wouldn't have added this. Nevertheless, is there a reason the dracut `dm` module doesn't pull in these things as dependencies? I don't really think we want to reimplement that, since we could only do that badly and it would be brittle.

reubeno:
Those are great points. I dug in a bit more to figure out what was going on -- in my case, the initrd was getting created in "host only" mode without my realizing it. That, in turn, caused `dracut` to disregard the (dracut) dm module's directives to add (kernel) dm modules.

Taking a step back, it's possible there's something that's worth doing here -- to ensure that DM kernel components are available when generating an initrd even in hostonly mode -- but I'd agree that this change as it stands isn't the right one to make. I'm going to close the PR and reevaluate.

(Thanks for the feedback!)

PR #1259, opened 2022-11-03
Flush writes to temporary files after calling copy_file_object

### LINKED ISSUE #1258: QEMU sometimes sees incomplete image when --ephemeral is used
When using `--ephemeral` with the `qemu` verb, I've sometimes seen images not boot correctly: the symptom is UEFI dropping me out to a shell and not finding a valid loader to execute.

Digging into this, it looks like the problem is that, when `copy_file_object` takes the code path where it calls `shutil.copyfileobj`, there's a risk that the writes to the destination aren't flushed and visible to external processes (namely, `qemu`). Adding in a call to `f.flush()` resolves this problem.

I'll post a PR shortly with a proposed fix.

reubeno:
This resolves an issue where qemu sees an incomplete image when --ephemeral is used. I've seen failures as a result of the missing flush in `copy_image_temporary`, and upon seeing similar logic in `copy_file_temporary` have defensively added a flush call there as well.

Alternate options for the solution might include flushing in `copy_file_object` after invoking `shutil.copyfileobj`, or instead letting the caller to `copy_file_object` decide if it's appropriate for it to explicitly flush before closing the destination object. I'd be interested in input as to whether one of these alternative options may be preferable.

Fixes #1258

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -616,6 +616,7 @@ def copy_file_object(oldobject: BinaryIO, newobject: BinaryIO) -> None:
         if e.errno not in {errno.EXDEV, errno.EOPNOTSUPP, errno.ENOTTY}:
             raise
         shutil.copyfileobj(oldobject, newobject)
+        newobject.flush()
 
 
 def copy_file(oldpath: PathString, newpath: PathString) -> None:


behrmann:
Why not just flush the `newobject` in `copy_file_object`? 

The changed places are the only users, but if we ever were to reuse `copy_file_object` we might run into the problem again.

behrmann (approved):

Issue #1028, opened 2022-06-28
Use of BaseImage fails with mount of mkosi-real-root


chymian:
I'm trying to build  a multilayer Image for use in set of PortableServcies and on SBCs

A) creating a minimal Image:
```
[Distribution]
Distribution=debian
Release=bullseye
Mirror=http://mirror:3142/debian

[Output]
Format=gpt_squashfs
BootProtocols=uefi
Bootable=yes
Verity=true
ReadOnly=true
CompressFs=zstd
ImageId=eb8-oci-minimal
NoChown=true

[Content]
BasePackages=true
Packages=
    busybox-static
    btrfs-progs

PasswordIsHashed=false
Autologin=true

[Partition]
RootSize=1G

[Validation]
Checksum=true
```

B) The Base-Image, which is the next layer, basically  the same definition, but with some extra packages, files and the `BaseImage=`
```
[Distribution]
Distribution=debian
Release=bullseye
Mirror=http://mirror:3142/debian

[Output]
Format=gpt_squashfs
BootProtocols=uefi
Bootable=yes
Verity=true
ReadOnly=true

ImageId=eb8-oci-base
NoChown=true

[Content]
BasePackages=true
Packages=
        iputils-ping
        wget

PasswordIsHashed=true
Autologin=true
WithNetwork=true

[Partition]
BaseImage=../eb8-oci-minimal/mkosi.output/eb8-oci-minimal_0.3.raw
RootSize=1G
[Validation]
Checksum=true
```

calling mkosi on that configuration throws that error:
```
…snip…
The partition table has been altered.                                                                               
‣  Created image with partition table as /home/rsg/tc.images/images/oci-images/eb8-oci-base/mkosi.output/.mkosi-och4
ivqt                                                                                                                
‣  Using ../eb8-oci-minimal/mkosi.output/eb8-oci-minimal_0.3.raw as the base image
‣   Attaching ../eb8-oci-minimal/mkosi.output/eb8-oci-minimal_0.3.raw as loopback…
‣   Attached /dev/loop1                                                                                             
‣   Attaching /home/rsg/tc.images/images/oci-images/eb8-oci-base/mkosi.output/.mkosi-och4ivqt as loopback…
‣   Attached /dev/loop2
‣   Formatting ESP partition                              
mkfs.fat 4.2 (2021-01-31)                                 
‣   Mounting image…                                       
[...]

PR #1260, opened 2022-11-04
Fix failure using BaseImage with a filesystem image

reubeno:
When `BaseImage` / `--base-image` are pointed at a root filesystem image, the image gets attached to a loopback device, but the filesystem is not mounted and subsequent attempts to create an overlay atop the loop device fail. This also adjusts logic in `attach_image_loopback` to cope with filesystem (i.e., non-disk) images that will have no partition table.

_Doesn't quite address the specific case in #1028 (where the base image is a GPT disk image as opposed to a filesystem image of the root partition), but related. If there is desire to support that case in the future, then mkosi probably can't safely assume that the partition structure of the base image is the same as that of the output image._

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -994,8 +994,6 @@ def attach_image_loopback(image: Optional[BinaryIO], table: Optional[PartitionTa
         yield None
         return
 
-    assert table
-
     with get_loopdev(image) as loopdev, flock(loopdev):
         # losetup --partscan instructs the kernel to scan the partition table and add separate partition
         # devices for each of the partitions it finds. However, this operation is asynchronous which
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
Sorry for the late response, I'm busy getting systemd-repart to a point where we can use it in mkosi, at which point a lot of this logic will change drastically, so I'm not super inclined to spend too much effort in what's currently used in mkosi. Are you fine with keeping this fix in a fork for now?

reubeno:
Thanks for taking a look at this. I appreciate the ongoing work to push disk/partitioning work to systemd-repart, so if you think the right call is to delay on this, then I'll hold onto this in my fork.

With that said, I'd expect that there will still need to be some logic added to mount the rootfs of an input base image (separate from the existing logic that handles the input base image just being a dir)... unless you're thinking that would be pushed onto the caller as a requirement before invoking `mkosi`?

DaanDeMeyer:
@reubeno We've just merged #1276 which significantly changes how this works, can you verify that this is still a problem?

DaanDeMeyer:
Closing for now, @reubeno let me know if this is still a problem on the latest commit. I haven't been testing --base-image really since I'm still planning bigger changes for it but in the current version I can take a look at it if it's not working

PR #1262, opened 2022-11-07
Make sure parse_args_file_group() looks in the right directory

DaanDeMeyer:
We should look for dropins in the same directory as where we look for the main config file.


qdeslandes:
Tested: `mkosi` is now able to run `mkosi -C $PATH ssh` if the guest distribution version is different from the host.

behrmann (approved):

Issue #1263, opened 2022-11-07
KERNEL_INSTALL_BYPASS=1 issue


leifliddy:
I'm attempting to build these Fedora 37 configs (on a Fedora 36 system)
https://github.com/leifliddy/asahi-fedora-usb/blob/main/mkosi.default
https://github.com/leifliddy/asahi-fedora-builder/blob/main/mkosi.default

And am running into this issue when running `mkosi`
```
  Running scriptlet: kernel-core-6.1.0-0.rc3.asahi1.1.fc37.aarch64                                            335/335 
kernel-install: Skipping execution because KERNEL_INSTALL_BYPASS=1
```
Setting `KERNEL_INSTALL_BYPASS` to `0` in  `mkosi/__init__.py` solves the issue  
```
run(cmdline, env=dict(KERNEL_INSTALL_BYPASS="0"))
```
I'm wondering why that wasn't occurring when I was building out the Fedora 36 config (on a Fedora 36 system).   
Also, what is the proper solution or workaround for this?

Update: 
This issue also happens when attempting to build a Fedora 37 config on a Fedora 37 system.

behrmann:
`KERNEL_INSTALL_BYPASS` is a variable to skip `kernel-install` if it's run somewhere, where you can't change that it's being invoked, but where you can change the environment. Apparently the `kernel-core` started setting it, although I have no idea why. 

I guess the right way to counteract this would be to set `KERNEL_INSTALL_BYPASS=0` in mkosi, but first I'd like to understand why it was set in the first place. @keszybz do you have any idea?

DaanDeMeyer:
We set this because we run kernel-install manually later anyway so to avoid building the initrd twice we disable kernel-install execution as part of running dnf.

leifliddy:
It's not the log message that's the issue, it's that kernel-install never seems to be run and I'm left with a boot partition that looks like this:
```[root@fedora boot]# ll
total 16
drwx------. 1 root root 14 Nov  8 12:47 efi
drwx------. 1 root root 24 Nov  8 12:46 grub2
drwxr-xr-x. 1 root root 14 Nov  8 12:46 loader
lrwxrwxrwx. 1 root root 57 Nov  8 12:46 symvers-6.1.0-0.rc3.asahi1.1.fc37.aarch64.gz -> /lib/modules/6.1.0-0.rc3.asahi1.1.fc37.aarch64/symvers.gz
```
What is meant by ` we run kernel-install manual later anyway`?   
[...]

PR #1264, opened 2022-11-08
Pass KERNEL_INSTALL_BYPASS through to the places where it is used

behrmann:
The idea of using KERNEL_INSTALL_BYPASS is to not run kernel-install twice unnecessarily, since we run it later anyway, but a user reported a regression of kernel-install not being run at all, although it apparently was in earlier. Passing the environment variable through will at most run kernel-install twice, this might be a workaround.

`invoke_pacman` is inlined into `install_arch` because that's the only place where it is used and it has a different signature than `invoke_dnf` and `invoke_apt` which breaks symmetry and makes it an unnecessary unicorn.

@leifliddy Does this solve your problem (when you add `KERNEL_INSTALL_BYPASS=0` to the environment in mkosi.default?)

Fixes: #1263

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2017,7 +2017,7 @@ def invoke_dnf(state: MkosiState, command: str, packages: Iterable[str]) -> None
     cmdline += [command, *sort_packages(packages)]
 
     with mount_api_vfs(state.root):
-        run(cmdline, env=dict(KERNEL_INSTALL_BYPASS="1"))
+        run(cmdline, env={"KERNEL_INSTALL_BYPASS": state.environment.get("KERNEL_INSTALL_BYPASS", "1")})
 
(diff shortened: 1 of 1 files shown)


leifliddy:
Yup, that worked. Thanks!

DaanDeMeyer (approved):

leifliddy:
@behrmann My project is like the unoffical Fedora asahi distro. 
The official Fedora release version is being working on here: 
https://pagure.io/fedora-asahi/mkosi/tree/main
And they're using a bootable gpt_btrfs image and I'm interested to see how that turns out. The existing installer doesn't currently support that but I bet they're working on it. 

My project is based off how the official Asahi (Arch) image gets built 
https://github.com/AsahiLinux/asahi-alarm-builder
I only did that so it would be compatible with the existing installer.

PR #1267, opened 2022-11-14
Fix passing extra options to systemd-nspawn

DaanDeMeyer:
systemd-nspawn ignores all options after the first argument so when we're passing kernel command line arguments, we need to make sure we pass the nspawn options first, otherwise they'll get ignored.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -6030,18 +6030,18 @@ def parse_args_file_group(
 
     d = config_path.parent
 
-    for dropin_dir in (d / "mkosi.conf.d", d / "mkosi.default.d"):
+    dirs = [Path("mkosi.conf.d"), Path("mkosi.default.d")]
+    if not d.samefile(Path.cwd()):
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -7812,15 +7812,13 @@ def run_shell_setup(config: MkosiConfig, pipe: bool = False, commands: Optional[
         cmdline += [f"--bind={config.build_sources}:/root/src", "--chdir=/root/src"]
 
     if config.verb == Verb.boot:
```
behrmann, line 7820: Shouldn't this be
```suggestion
    if config.verb == Verb.boot and (commands or config.cmdline):
```
since otherwise `commands or config.cmdline` might be `None`?

DaanDeMeyer: `config.cmdline` is typed to never be `None`, so we'll just add no arguments in that case

behrmann (approved):

behrmann:
CI failure in Arch CPIO is unrelated.

Issue #1265, opened 2022-11-14
Consider adding a fallback for nspawn's rootidmap


jre21:
After updating to systemd 252, my mkosi builds started failing with errors like

```
Failed to map ids for bind mount /home/jre2718/src/mkosi-images/mkosi.workspace/glfx9oa4/root/root/src: Operation not permitted
```

I eventually tracked this down to systemd-nspawn's `rootidmap` flag, added in #1151 (mkosi runs correctly if I unconditionally set `idmap_opt = ""` in the two places where it's defined).  I think there's something a bit odd with my system (more on that below), but it would be nice to do one of the following:

1.  Test whether the source directory's containing filesystem supports idmapped mounts and if so, drop the `rootidmap` flag.
2.  Add a configuration option to disable the idmap behavior.

If the first option is doable, that would obviously give the cleaner user experience, although I'm not sure what a clean test would look like.

As to why my setup might be weird: my home directory is a luks/btrfs volume managed by systemd-homed, with ext4 as the underlying root filesystem.  With some experimentation, I found that I could mount a directory on the root filesystem with idmap, but anything under my homedir fails.  I also tried manually replicating the homed setup by creating a scratch file on my root filesystem, opening it as a luks volume, and creating a fresh btrfs filesystem in the new volume.  This new btrfs filesystem was mountable with idmap.  So either something in homed itself is interfering with idmap, or my home directory's filesystem is in an old enough version of the format to be incompatible.

To be honest, I hadn't encountered idmap before today, so I'm not sure what else to check while debugging, but at the very least it seems like checking the systemd-nspawn version isn't enough to guarantee that it's safe to use idmap in bind mounts.

behrmann:
Since `idmap_opt` is set for you I gather you are running systemd 252. Can you also tell us what kernel and mkosi versions you are running on what distro?

jre21:
One additional data point here: if I mount my home directory manually (losetup + cryptsetup + mount) instead of through systemd-homed, then I can bind-mount it into a container with rootidmap, so the failure case is specifically an interaction of homed and idmap.

qdeslandes:
As you've been able to run `systemd-nspawn` with `rootidmap`, I assume your system supports it. 

Regarding fixes in case `rootidmap` wouldn't work on any given system:
- Detecting whether it's supported is not easy, see [`testsuite-13.sh`](https://github.com/systemd/systemd/blob/main/test/units/testsuite-13.sh#L77).
[...]

PR #1270, opened 2022-11-16
Allow for UID mapping to be disabled

qdeslandes:
By default, mkosi would bind mount directories with nspawn's rootidmap option to provide consistent files ownership. If the system doesn't support ID mapping, mkosi would fail.

Add --idmap option, default to True, to allow users to disabled UID mapping (--idmap=no).

Workaround for #1265.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 - Rename `--no-chown` to `--chown` and set it to default to `True`, preserving
   current behaviour.
+- Add `--idmap` option to run `--systemd-nspawn` with ID mapping support. Defaults
+  to `True`. `--idmap=no` can be used to prevent usage of ID mapping.
 
 ## v14
 
--- a/man/mkosi.1
+++ b/man/mkosi.1
@@ -711,6 +711,12 @@ archive.
 Set the machine\[cq]s ID to the specified value.
 If unused, a random ID will be used while building the image and the
 final image will be shipped without a machine ID.
+.TP
+\f[B]\f[CB]Idmap=\f[B]\f[R], \f[B]\f[CB]--idmap\f[B]\f[R]
+By default, \f[C]mkosi\f[R] will run \f[C]systemd-nspawn\f[R] with
+ID mapping enabled. This ensures newly-created directories and files are
+owned by the user calling \f[C]sudo\f[R] or \f[C]pkexec\f[R] instead of
+\f[C]root\f[R]. This behaviour can be disabled with \f[C]--idmap=no\f[R].
 .SS [Content] Section
 .TP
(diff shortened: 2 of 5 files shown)


behrmann:
LGTM, although this might be a bit difficult to discover, since users will only see nspawn failing.  We could probably also do some autodetection to try to figure this out ourselves, but this is error prone, since we don't yet know where the problem actually is (homed? kernel?). 

My point is: some way to steer users to use this would be nice.

On a semi-related not, not knowing where the bug is, it might be resolved at some point and we will have to keep the option around, which seems a bit cumbersome. I prefer having a superfluous option (that we can suppress and make a noop later) to having this be broken for users.

behrmann:
> > On a semi-related not, not knowing where the bug is, it might be resolved at some point and we will have to keep the option around, which seems a bit cumbersome. I prefer having a superfluous option (that we can suppress and make a noop later) to having this be broken for users.
> 
> This option should be kept IMO, as some systems disable ID mapping.

Ok, that's a good point (although it would be fair for nspawn to have fallback behaviour, I guess it can't so that we can detect idmapped mounts :) ).

behrmann:
> > LGTM, although this might be a bit difficult to discover, since users will only see nspawn failing. We could probably also do some autodetection to try to figure this out ourselves, but this is error prone, since we don't yet know where the problem actually is (homed? kernel?).
> > My point is: some way to steer users to use this would be nice.
> 
> The way ID mapping support is detected in `systemd` integration test is through failure: if `systemd-nspawn` fails, it parses the logs and checks whether ID mapping is involved. It's not great, and I haven't found any deterministic approach yet.

Maybe that's what we should do, too. If nspawn fails, we could check for `Failed to map ids` in the stderr output and give a more helpful error, that steers people towards giving `--idmap=no` a try. 

Maybe at some point we could add an ASCII clippy to out error messages :)

qdeslandes:
> > > LGTM, although this might be a bit difficult to discover, since users will only see nspawn failing. We could probably also do some autodetection to try to figure this out ourselves, but this is error prone, since we don't yet know where the problem actually is (homed? kernel?).
> > > My point is: some way to steer users to use this would be nice.
> > 
> > 
> > The way ID mapping support is detected in `systemd` integration test is through failure: if `systemd-nspawn` fails, it parses the logs and checks whether ID mapping is involved. It's not great, and I haven't found any deterministic approach yet.
> 
> Maybe that's what we should do, too. If nspawn fails, we could check for `Failed to map ids` in the stderr output and give a more helpful error, that steers people towards giving `--idmap=no` a try.
> 
> Maybe at some point we could add an ASCII clippy to out error messages :)

I've added a man page entry for `--idmap` option. I'm not keen to add code to look for this specific idmap-related error message in `systemd-nspawn`'s output to be honest. I think it's error prone, and the command's output is printed anyway.

behrmann:
@qdeslandes This needs a rebase now.

PR #1271, opened 2022-11-16
Rename --no-chown option to --chown

qdeslandes:
Rename `--no-chown` to `--chown` and defaults to true. This will make it easier to reason with the code behaviour when this option is used.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3033,7 +3033,7 @@ def save_cache(state: MkosiState, raw: Optional[str], cache_path: Optional[Path]
             unlink_try_hard(cache_path)
             shutil.move(cast(str, state.root), cache_path)  # typing bug, .move() accepts Path
 
-    if not state.config.no_chown:
+    if state.config.chown:
         chown_to_running_user(cache_path)
 
 
@@ -3051,11 +3051,9 @@ def _link_output(
 
     os.link(oldpath, newpath)
 
-    if config.no_chown:
-        return
-
-    relpath = path_relative_to_cwd(newpath)
-    chown_to_running_user(relpath)
+    if config.chown:
+        relpath = path_relative_to_cwd(newpath)
+        chown_to_running_user(relpath)
 
 
 def link_output(state: MkosiState, artifact: Optional[BinaryIO]) -> None:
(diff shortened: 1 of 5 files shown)


mkosi/__init__.py
```diff
@@ -5372,10 +5370,11 @@ def create_parser() -> ArgumentParserMkosi:
     group.add_argument("--image-version", help="Set version for image")
     group.add_argument("--image-id", help="Set ID for image")
     group.add_argument(
-        "--no-chown",
+        "--chown",
```
behrmann, line 5373: I'm fine with this change in principal and we should have done this like this from the start. Could you add `--no-chown`, though as a suppressed option that goes into the same dest? `BooleanAction` does a lot (`--not-chown`, `--withouth-chown`), but I think with this change passing `--no-chown` would be an error.

mkosi/__init__.py
```diff
@@ -6293,6 +6292,7 @@ def find_output(args: argparse.Namespace) -> None:
     else:
         return
 
+    mkdirp_chown_current_user(args.output_dir, chown=args.chown)
```
behrmann, line 6295: Are these (and the other additions like it) new or am I missing something? If they are new, could you put them in a separate commit?

mkosi/__init__.py
DaanDeMeyer: I tend to agree but on the other hand, we're going to be breaking compat all over the place anyway in this release, so maybe we should take the time and get rid of all the deprecated stuff and do some extra breaking changes without caring for backwards compat?

mkosi/__init__.py
qdeslandes: My bad, commits mixup.

DaanDeMeyer:
Missing documentation update,

could you also add an entry to NEWS?

behrmann:
Apart from missing docs and news, LGTM.

keszybz:
Hmm, this breaks backwards-compat without a particularly good reason. I don't think we should do this. Having the general mindset that you don't do that unless necessary just makes things easier in the long run.

I think it's completely fine to invert the internal boolean. But the commandline option should just remain in place.
I see two possibilities: 
1. add a new positive option and make the old one suppressed in help
2. bump required python version and use BooleanOptionalAction that was added in 3.9 (https://docs.python.org/3/library/argparse.html?highlight=booleanoptionalaction#action).

behrmann:
I raised the same point initially, but once @DaanDeMeyer posts his big push to use repart internally, we'll be losing a lot of command line arguments anyway - at least many will lose their raison d'être and keeping them around like some Potempkin village will probably be too much of a burden in the long run, so some form a clean break might, for once, be warranted.

keszybz:
Ah, OK. I didn't see that comment because it was hidden as resolved. Makes sense, I withdraw my comment.

behrmann (approved):

PR #1272, opened 2022-11-18
Drop HostonlyInitrd

DaanDeMeyer:
Adding this option was a mistake (mea culpa), we should limit host impact on image builds as much as possible so let's drop this option that makes the initrd generated by dracut host specific. This will slow down initrd generation but once we switch to generating initrds with mkosi we'll have a proper fix for that.


DaanDeMeyer:
I'll just include this in the repart PR

PR #1275, opened 2022-11-22
Change default suffix for verity signature partition

### LINKED ISSUE #1274: Attempts to write to *.roothash.p7s or *.usrhash.p7s twice
Running a build with both `Verity=signed` and `SplitArtifacts=yes` causes the build to fail at the last step with the following error:
```
FileExistsError: [Errno 17] File exists: '/home/user/src/build/mkosi.output/arch~rolling/.mkosi-rae4hbks' -> '/home/user/src/build/mkosi.output/arch~rolling/archetype_2022.11.13-1.root.roothash.p7s'
```
This appears to be because both `link_output_split_verity_sig` and `link_output_root_hash_p7s_file` attempt to write to the same path, namely `*.roothash.p7s` (or `*.usrhash.p7s` when using `UsrOnly`). It is not clear to me whether both of these functions are meant to write the same contents, but if they do one solution could be to disable one of those functions when both `Verity=signed` and `SplitArtifacts=yes`.

jheretic:
Currently the filename suffix used for the `roothash` (or `usrhash`) signature file and the filename suffix used for the partition image of the verity signature partition conflict. Consequently, it is not currently possible to successfully complete a build with both the `SplitArtifacts=yes` and `Verity=signed` options due to a file conflict.

This PR changes the output file suffix of the verity partition image to `*.verity-sig` instead `*.roothash.p7s`, which on review I believe of available options most closely maps the expected naming scheme for partitions and files.

Fixes #1274

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -6662,7 +6662,8 @@ def load_args(args: argparse.Namespace) -> MkosiConfig:
         if args.verity:
             args.output_split_verity = build_auxiliary_output_path(args, f"{root_or_usr(args)}.verity", True)
             if args.verity == "signed":
-                args.output_split_verity_sig = build_auxiliary_output_path(args, f"{roothash_suffix(args)}.p7s", True)
+                args.output_split_verity_sig = build_auxiliary_output_path(args,
+                                                                           f"{root_or_usr(args)}.verity-sig", True)
         if args.bootable:
             args.output_split_kernel = build_auxiliary_output_path(args, ".efi", True)
 


bluca:
sorry no can do, this breaks backward compatibility. It's also wrong, the canonical suffix for pkcs7 is .p7s

keszybz:
We can't really have two outputs with the same name. We'll be breaking backwards compat massively with the next release anyway. So now is a good time to do the rename. 

As to the specific suffix, I don't have an opinion. I think that something with `.p7s` would be nicer. "-sig" could mean many different formats. `.verity.p7s` ?

bluca:
The files that are input to systemd (eg: RootImage= and friends) cannot be changed, as the expectations are baked in and used in production. pid1 expects `foo.raw` to be accompanied by `foo.verity`, `foo.roothash` and `foo.roothash.p7s`

DaanDeMeyer:
Now that we've migrated to systemd-repart, mkosi is not responsible anymore for determining the split verity sig suffix, so let's close this. You can configure this with `SplitName=` in repart partition definition files now.

PR #1276, opened 2022-11-22
Migrate disk image building to systemd-repart

DaanDeMeyer:
Let's switch to systemd-repart to build disk images. systemd-repart
allows building disk images declaratively. repart will assemble a
disk image according to a set of partition definition files, which
it will search for in a set of predefined locations or in the locations
provided via the --definitions option. Partition definition files are
defined in the usual systemd ini style and define a single partition
each. Many aspects of the partition can be configured, including its
size, type, and how it should be populated.

Of note should be the CopyFiles= setting that allows providing a set
of files/directories from which the partition should be populated. The
CopyFiles= setting takes a list of path pairs, where the first path in
a pair defines the source location, and the second path the destination
location in the partition. If combined with repart's --root or --image
options, the source path is resolved relative to the given root directory.

Along with switching to systemd-repart, we also rework how we build
images. Previously, we first provisioned the disk image with all its
partitions, then mounted the disk image and finally populated it with
contents. Now, we provision to a regular directory on the host filesystem
first, followed by creating the disk image with systemd-repart. The partition
definition files can be supplied via the --repart-directory option or
in the mkosi.repart/ directory. If not provided, a default configuration
consisting of an automatically sized root partition and an ESP partition of
256M is used.

repart's --root switch is used so that the partition definition files
don't need to encode the full path to the root directory of the image
we're building. For example, to copy the root directory to a root
partition, it's sufficient to specify "CopyFiles=/:/" in the definition
file.

If we're building non-bootable images, any ESP partition definitions
are excluded. To allow building UKIs with an embedded roothash cmdline
parameter, we run repart twice, first to populate all partitions except
the ESP/XBOOTLDR partitions, and again to populate the ESP/XBOOTLDR
partitions including any UKIs with embedded roothash cmdline parameter
(which we determined when running repart the first time).

Because we don't know up-front anymore where the ESP partition will be
mounted, all boot loader files are installed to /boot. So to populate
an ESP partition, you'd use "CopyFiles=/boot:/" in the partition
definition file of the ESP partition.

Also, since we don't know up-front anymore which filesystem we'll be
building for, we stop installing filesystem related packages by default.
Users will be required to add the necessary filesystem packages themselves.
Similarly, we also stop installing cryptsetup and device-mapper by default
since we don't know upfront anymore whether partitions will be verity
protected or encrypted.

By switching to systemd-repart, we also set ourselves up for building
images without needing root privileges or loop devices. systemd-repart
is fully capable (from v253 onwards) of building disk images without
needing root privileges or loop devices. After we switch to
systemd-repart, we'll only need root privileges to be able to run
systemd-nspawn (which should not be necessary in the future anymore).

This PR also removes all the option related to disk image building that
can now be specified in repart definition files instead. This includes:

- Format=gpt_xxx options are replaced with a single "disk" options.
Filesystem can now be specified with repart's Format= option
- Format=plain_squashfs (Can be reproduced by a single repart squashfs
root partition combined with SplitArtifacts=yes)
- Verity= (Replaced by repart's Verity= options)
- Encrypt= (Replaced by repart's Encrypt= option)
- RootSize=, HomeSize=, VarSize=, TmpSize=, ESPSize=, SwapSize=, SrvSize=
(Replaced by repart's size options)
- UsrOnly= (replaced with `CopyFiles=/:/usr` in a usr partition definition)
- OutputSplitRoot=, OutputSplitVerity=, (Replaced by repart's SplitName= option)
- OutputSplitKernel= (UKI is now always written to its own output file)
- GPTFirstLBA (Removed, no equivalent in repart)
- ReadOnly= (Replaced by repart's ReadOnly= option per partition)
- Minimize= (Replaced by repart's Minimize= option per partition)
- CompressFs= (No equivalent in repart, can be replicated by replacing mkfs.<fs>
in $PATH with a script that adds the necessary command line option)
- CompressOutput= is renamed to Compress=
- MkSquashfs= (Can be replaced with a script in $PATH that invokes
the correct binary)

We also remove the WithoutUnifiedKernelImages= switch as building unified
kernel images is trivial and fast these days.


action.yaml
```diff
@@ -29,20 +29,23 @@ runs:
...
       sudo apt-get update
       sudo apt-get build-dep systemd
-      git clone https://github.com/systemd/systemd-stable --branch v251.2 --depth=1
-      meson systemd-stable/build systemd-stable
-      ninja -C systemd-stable/build systemd-nspawn
-      sudo ln -svf $PWD/systemd-stable/build/systemd-nspawn $(which systemd-nspawn)
+      sudo apt-get install libfdisk-dev
+      git clone https://github.com/systemd/systemd --depth=1
+      meson systemd/build systemd -Drepart=true -Defi=true
+      ninja -C systemd/build
```
bluca, line 42: just build the individual targets needed? repart, nspawn, bootctl, should save some time

DaanDeMeyer: I was thinking a bit of the future where we'll start using more executables as well, systemd-dissect, systemd-firstboot, maybe tmpfilies ... . It seemed more robust to just build everything

bluca: Given it runs on every build of every config of every PR, even saving a few minutes will add up, no? Should be easy to see when something new is needed

DaanDeMeyer:
This also gets rid of the pytest integration tests we added a while ago. Given that we'll be moving to integrating more with build systems in the future, building and booting the image as part of the same integration test seems like the wrong way to go. Instead we should just build the images in one step in a build system such as meson and then have the integration test(s) depend on it. For now, I switched us to the same approach used in systemd which does the trick as well.

DaanDeMeyer:
I also switched us from using temporary files and passing around `Optional[BinaryIO]` everywhere to a staging directory that all functions that produce output write into and look into for dependent files. This was necessary because systemd-repart produces one or more output files which we can't account for with the current model and because it simplifies everything quite a bit as well.

bluca:
I think it would be good to add what you wrote in the commit message about the old option in a new section in the manpage, called "migration guide" or so?

action.yaml
```diff
@@ -24,25 +24,31 @@ runs:
...
+      meson systemd/build systemd -Drepart=true -Defi=true
+      ninja -C systemd/build systemd-nspawn systemd-dissect systemd-repart bootctl
+      sudo ln -svf $PWD/systemd/build/systemd-nspawn /usr/bin/systemd-nspawn
+      sudo ln -svf $PWD/systemd/build/systemd-dissect /usr/bin/systemd-dissect
+      sudo ln -svf $PWD/systemd/build/systemd-repart /usr/bin/systemd-repart
+      sudo ln -svf $PWD/systemd/build/bootctl /usr/bin/bootctl
       systemd-nspawn --version
+      systemd-dissect --version
+      systemd-repart --version
+      bootctl --version
```
behrmann, line 51: Maybe we could add something like this to `tools/` so that people can trivially build the stuff they need.

mkosi/__init__.py
```diff
@@ -2645,534 +1138,197 @@ def install_unified_kernel(
...
 
-def link_output_root_hash_file(config: MkosiConfig, root_hash_file: Optional[SomeIO]) -> None:
-    if root_hash_file:
-        assert config.output_root_hash_file
-        suffix = roothash_suffix(config)
-        with complete_step(f"Linking {suffix} file…", f"Linked {path_relative_to_cwd(config.output_root_hash_file)}"):
-            _link_output(config, root_hash_file.name, config.output_root_hash_file)
 
+def hash_file(of: TextIO, path: Path) -> None:
+    bs = 16 * 1024 ** 2
```
behrmann, line 1260: Tiny of a style nit for readability
```suggestion
    bs = 16 * 1024**2
```

mkosi.extra/usr/lib/systemd/mkosi-check-and-shutdown.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -eux
+# SPDX-License-Identifier: LGPL-2.1-or-later
```
behrmann, line 2:
```suggestion
#!/bin/bash
# SPDX-License-Identifier: LGPL-2.1-or-later
set -ex
```
this way we don't get noise from bash config files being read. Dropped the `-u`, since there are no variables, but it also doesn't hurt if it's there.

behrmann (commented):
Haven't had time to give this a spin, but tests pass and looking over the code this looks sensible, mostly dropping stuff and moving smaller bits around. It's probably easier to just pick up any pieces we dropped after merging this.

Besides the really, really minor nits, I agree with @bluca that some sort of migration guide (and a NEWS entry), also documenting what version of systemd we target, is something we should add, because people will have to be aware what they need to change compared to before.

.gitignore
```diff
@@ -12,7 +12,6 @@
 /dist
 /mkosi.build
 /mkosi.egg-info
-/mkosi.extra
```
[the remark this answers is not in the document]
DaanDeMeyer: This seems a bit too restrictive, we should have the freedom to reorder things inside the repository. If users are running from the repository, they'll have to accept that that might involve their images changing when they pull down a new version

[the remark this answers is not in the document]
behrmann: Agreed, we should have that freedom, but these files are just for testing purposes, no? My thinking was, that this would be better communicated by not having them in the tree proper.

keszybz:
> * Format=plain_squashfs (Can be reproduced by a single repart squashfs
>   root partition combined with SplitArtifacts=yes)

I find this problematic. I expect that it'll make building of initrds more expensive. But I think it's not a blocker… In the worst case we can always add some support to build such images directly again.

> * CompressFs= (No equivalent in repart, can be replicated by replacing mkfs.
>   in $PATH with a script that adds the necessary command line option)

I don't think the approach of requiring the user to create fake wrappers to pass options is acceptable in the long run. But it shouldn't be too hard to add support for passing compression options again.

> CompressOutput= is renamed to Compress=

Can we please not do that. There's bound to be many different compression settings. No need to recreate the old issue that was once solved.

.github/mkosi.extra/usr/lib/systemd/system/mkosi-check-and-shutdown.service
```diff
@@ -0,0 +1,13 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+[Unit]
+Description=Check if any service failed and then shutdown the machine
```
keszybz, line 3: "shut down"

.github/mkosi.extra/usr/lib/systemd/system/mkosi-check-and-shutdown.service
```diff
@@ -0,0 +1,13 @@
...
+After=multi-user.target network-online.target
+Requires=multi-user.target
+OnFailure=poweroff.target
+OnFailureJobMode=replace-irreversibly
+
+[Service]
+Type=oneshot
+ExecStartPre=-rm -f /failed-services /testok
```
keszybz, line 11: `-` is not useful, I think. If the removal fails, we probably want to abort the test.

.github/workflows/ci.yml
```diff
@@ -173,6 +172,18 @@ jobs:
 
         [Output]
         Format=${{ matrix.format }}
+        Bootable=yes
+        KernelCommandLine=systemd.unit=mkosi-check-and-shutdown.service
+                          !quiet
```
keszybz, line 177: I didn't know this syntax.

mkosi/__init__.py
```diff
@@ -3888,32 +1972,8 @@ def create_parser() -> ArgumentParserMkosi:
...
-    )
-    group.add_argument(
-        "--compress-fs",
-        type=parse_compression,
-        nargs="?",
-        metavar="ALG",
-        help="Enable in-filesystem compression (gpt_btrfs, subvolume, gpt_squashfs, plain_squashfs)",
-    )
-    group.add_argument(
-        "--compress-output",
```
keszybz, line 3916: As mentioned above, I don't think this rename should be done.

mkosi/__init__.py
```diff
@@ -5779,188 +3559,138 @@ def run_kernel_install(state: MkosiState, cached: bool) -> None:
...
 
-    # Partition contents
-    split_root: Optional[BinaryIO]
-    split_verity: Optional[BinaryIO]
-    split_verity_sig: Optional[BinaryIO]
-    split_kernel: Optional[BinaryIO]
-    split_kernel_image: Optional[BinaryIO]
-    split_initrd: Optional[BinaryIO]
-    split_kernel_cmdline: Optional[TextIO]
+    cache = cache_tree_path(state.config, is_final_image=False) if state.do_run_build_script else cache_tree_path(state.config, is_final_image=True)
```
keszybz, line 3566:
```py
cache = cache_tree_path(state.config, is_final_image=not state.do_run_build_script)
```
?

keszybz (changes requested):
I think we should merge this asap.

DaanDeMeyer:
> I don't think the approach of requiring the user to create fake wrappers to pass options is acceptable in the long run. But it shouldn't be too hard to add support for passing compression options again.

It's not a mkosi problem anymore after we merge this. It'll be a repart problem. If we want to do this slightly nicer we'll need to have repart read extra mkfs options from environment variables like we do in homed and such.

> I find this problematic. I expect that it'll make building of initrds more expensive. But I think it's not a blocker… In the worst case we can always add some support to build such images directly again.

cpio building hasn't changed, we only remove the plain squashfs output format because the complexity to keep it is bigger than then the added benefit of supporting it. cpio and tar output formats are still supported as before. (Unless you're saying that you want squashfs initrds?)

> Can we please not do that. There's bound to be many different compression settings. No need to recreate the old issue that was once solved.

Reverted this, we drop CompressFs and keep CompressOutput and remove Compress.

DaanDeMeyer:
CI failures is unrelated, shall we merge this and fix any remaining issues in follow up PRs?

.gitignore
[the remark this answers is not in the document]
behrmann: With the files moved this isn't necessary anymore, no?

behrmann (approved):
Sounds good to me, one last thing I noticed.

DaanDeMeyer:
Only changed .gitignore, let's merge

PR #1277, opened 2022-11-22
Automatically detect if the password is hashed

keszybz:


DaanDeMeyer:
My plan was to switch us to systemd-firstboot after the move to repart, which would mean the --password-is-hashed option would be delegated to systemd-firstboot as well, maybe implement this in systemd-firstboot instead?

keszybz:
OK. But let's leave this open for now as a reminder…

mkosi/__init__.py
```diff
@@ -3080,6 +3080,28 @@ def reset_random_seed(root: Path) -> None:
     with complete_step("Removing random seed"):
         random_seed.unlink()
 
+HASHED_PASSWORD_RE = r'''
```
behrmann, line 3083:
```suggestion
HASHED_PASSWORD_RE = re.compile(r'''
```
This way the `re.VERBOSE` flag can live directly on the same object as the regex that is written in its style and setting the flag doesn't need to be remembered at the site of use.

behrmann (commented):
I guess we're not going to use it, but since it's here anyway :)

DaanDeMeyer:
Should be superseded by #1543

DaanDeMeyer:
So instead of automatically detecting whether the root password is hashed, you can now prefix the password with "hashed:" to indicate that it is hashed, I figure that should be a sufficiently nice user experience without having to add extra complexity to mkosi to figure out whether the root password is hashed or not. Hence, let's close this PR.

PR #1280, opened 2022-11-29
Remove annotations import from __future__

behrmann:
PEP 563 is a PEP that stringifies type annotations. The feature was made available as opt in. It is still very much in flux or under consideration upstream and has fallen so much out of favour that the Python steering council added [the note](https://docs.python.org/3.11/whatsnew/3.11.html#pep-563-may-not-be-the-future)

> PEP 563 Postponed Evaluation of Annotations (the from __future__ import annotations future statement) that was originally planned for release in Python 3.10 has been put on hold indefinitely. See this message from the Steering Council for more information.

in the release notes of CPython 3.11 after punting on accepting the PEP for two releases. The import does a lot more, but we only use the type stringification. There is little gain from using this, but with one of the main features of the import indefinitely on hold, it seems prudent not to use it, since removing it doesn't cost as much.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1,7 +1,5 @@
 # SPDX-License-Identifier: LGPL-2.1+
 
-from __future__ import annotations
-
 import argparse
 import configparser
 import contextlib
--- a/mkosi/backend.py
+++ b/mkosi/backend.py
@@ -1,7 +1,5 @@
 # SPDX-License-Identifier: LGPL-2.1+
 
-from __future__ import annotations
-
 import argparse
 import ast
 import collections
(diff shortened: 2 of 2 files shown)


keszybz:
Hmm, but do we actually care? I think it's unlikely that they'll implement a breaking change that breaks this.

behrmann:
It's actually a good question whether they would break this... I don't know how widely used this is and it is an underscored namespace, although probably the most-widely used one. Be that as it may, my preference would be towards getting rid of this deviation from the mainline, since the diff is pretty small. This PR is not yet rebased against #1276 - after that the change is smaller still, if I remember correctly.

I'm pretty sure that in the longer term *something* will happen (maybe together with lazy imports of modules, which also bubbles up every few months), but I'd rather wait for PEPs regarding this to be accepted.

behrmann:
With the move to repart this problem mosly solved itself. Starting with 3.11 the alternative for our usecase is probably the [`Self` type](https://docs.python.org/3.11/whatsnew/3.11.html#whatsnew311-pep673).

DaanDeMeyer (approved):

PR #1281, opened 2022-11-29
Bump Python to 3.9

behrmann:
With 3.7 we are currently at the lower end of supported Python versions. The current Debian stable, that will become oldstable next year, runs 3.9, so it's maybe time for us to get the goodness like `Path.readlink`, `missing_ok=` on `Path`s, `str.removesuffix` and `str.removeprefix` as well as much fewer typing imports.

This is based on #1280, so the first commit can be ignored. 

@DaanDeMeyer Since this PR, too, might generate a lot of merge conflicts for you: The last commit can be mostly ignored (i.e. if it makes it easier dropped and postponed), it was generated with [pyupgrade](https://github.com/asottile/pyupgrade) and the remaining imports were found by pyflakes (which we already run in CI).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
       group: ${{ github.workflow }}-${{ github.ref }}
       cancel-in-progress: true
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
     - name: Install
       run: |
@@ -121,7 +121,7 @@ jobs:
           - disk
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - uses: ./
 
     - name: Install dependencies
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -513,12 +513,7 @@ def remove_files(state: MkosiState) -> None:
 
 def parse_epel_release(release: str) -> int:
     fields = release.split(".")
-    if fields[0].endswith("-stream"):
-        epel_release = fields[0].split("-")[0]
-    else:
-        epel_release = fields[0]
-
-    return int(epel_release)
+    return int(fields[0].removesuffix("-stream"))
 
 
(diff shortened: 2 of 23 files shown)


behrmann:
Added the "dont merge" because this contains #1280, which should be discussed there.

DaanDeMeyer:
Yeah this one can wait until after repart, fixing merge conflicts for the other PR was painful enough already

behrmann:
I feel your pain :) But no worries, this one is extremely easy to recreate, so no hurry.

keszybz:
> ukify requires python 3.11 I think? So we can bump even more

Does it? At least 3.10 seems to accept the syntax without issues.

behrmann:
Yeah, 3.10 should be sufficient and I wanted to ask about that too, but I haven't found the time to rebase this yet, but should tomorrow.

Bumping to 3.11 seems too early to me, since 3.11 hasn't landed in Arch and Debian yet. I'm not too optimistic that 3.10 will make it into Debian bookworm; they're trying, though. It's one thing we ask people to build systemd, but that's quite easy, but asking people to manage a separate Python toolchain might be asking too much.

Bumping to 3.10 would be a possibility and we basically already require it because of ukify, but I'm a bit hesitant. ukify is pretty easy to backport to 3.9, but once we move mkosi proper to 3.10 we're back to requiring people to build a separate cpython on distros that don't offer it (Arch and Debian stable).

DaanDeMeyer:
Hmm, if bookworm doesn't get 3.10 we'd probably do good to downgrade ukify's required version to 3.9

behrmann:
Bookworm *will* get 3.10, but it's - at least to me - unclear whether it will get 3.11.

keszybz:
Yeah, that was my plan when writing ukify: I wanted to start with the modern syntax and other frills because it's nicer this way, but I have no qualms about adding workarounds for older versions. Supporting 3.9 would require getting rid of `match`, which I think works quite well in that code, but support for older distros is more important.

DaanDeMeyer (approved):

PR #1298, opened 2023-01-09
Add --initrd option

rphibel:
This change adds support for a `--initrd` option to provide a list of initrd files that are used in the generated image instead of the initrd generated by `dracut`. For the kernel modules, an initrd is generated from the directory `/usr/lib/modules` and is added to the list of initrds passed to `ukify`.

### DIFF
--- a/.github/initrd/alma.conf
+++ b/.github/initrd/alma.conf
@@ -1,0 +1,13 @@
+[Distribution]
+Distribution=alma
+
+[Output]
+ImageId=initrd
+Format=cpio
+ManifestFormat=
+CompressOutput=zstd
+
+[Content]
+Packages=
+        systemd
+        systemd-udev
--- a/.github/initrd/arch.conf
+++ b/.github/initrd/arch.conf
@@ -1,0 +1,12 @@
+[Distribution]
+Distribution=arch
+
+[Output]
+ImageId=initrd
+Format=cpio
+ManifestFormat=
+CompressOutput=zstd
+
+[Content]
+Packages=
+        systemd
--- a/.github/initrd/centos.conf
+++ b/.github/initrd/centos.conf
@@ -1,0 +1,13 @@
+[Distribution]
+Distribution=centos
+
+[Output]
+ImageId=initrd
+Format=cpio
+ManifestFormat=
+CompressOutput=zstd
+
+[Content]
+Packages=
+        systemd
+        systemd-udev
--- a/.github/initrd/gentoo.conf
+++ b/.github/initrd/gentoo.conf
@@ -1,0 +1,12 @@
+[Distribution]
+Distribution=ubuntu
+
+[Output]
+ImageId=initrd
+Format=cpio
+ManifestFormat=
+CompressOutput=zstd
+
+[Content]
+Packages=
+        systemd
--- a/docs/initrd.md
+++ b/docs/initrd.md
@@ -1,0 +1,28 @@
+# Building a custom initrd and using it in a mkosi image
+
+To build an image with a mkosi-built initrd:
+1. Build an initrd image. You can for example build a **compressed** `cpio`
+archive with mkosi containing the packages `systemd` and `udev`. Here is a
+configuration file to build such an image for `Fedora`:
+```
+[Distribution]
+Distribution=fedora
+
+[Output]
+ImageId=initrd
+Format=cpio
+ManifestFormat=
+CompressOutput=zstd
+
+[Content]
+Packages=
+        systemd
+        systemd-udev
+```
+2. Invoke `mkosi` passing the initrd image via the `--initrd` option:
+```bash
+mkosi --initrd=<path-to-initrd-image> ...
+```
+This will build an image using the provided initrd image.
+mkosi will add the kernel modules found in the distribution image to this initrd.
+
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -66,6 +66,7 @@ from mkosi.backend import (
     nspawn_version,
     patch_file,
     path_relative_to_cwd,
+    read_os_release,
     run,
     run_workspace_command,
     scandir_recursive,
@@ -160,6 +161,13 @@ EFI_ARCHITECTURES = {
     "aarch64": "aa64",
     "armhfp": "arm",
     "riscv64:": "riscv64",
+}
+
+
+# Magic numbers for compression formats
+MAGIC_NUMBERS = {
+    "zstd": b"\x28\xB5\x2F\xFD",
+    "xz": b"\xFD\x37\x7A\x58\x5A\x00",
 }
 
 
@@ -881,6 +889,12 @@ def compressor_command(option: Union[str, bool], src: Path) -> list[PathString]:
         die(f"Unknown compression {option}")
 
 
+def uncompress_command(option: str, src: Path) -> list[PathString]:
+    """Returns a command suitable for decompressing archives to stdout."""
+
+    return [option, "-d", "-c", src]
+
+
 def tar_binary() -> str:
     # Some distros (Mandriva) install BSD tar as "tar", hence prefer
     # "gtar" if it exists, which should be GNU tar wherever it exists.
@@ -910,24 +924,26 @@ def make_tar(state: MkosiState) -> None:
         run(cmd)
 
 
-def find_files(root: Path) -> Iterator[Path]:
-    """Generate a list of all filepaths relative to @root"""
-    yield from scandir_recursive(root,
+def find_files(dir: Path, root: Path) -> Iterator[Path]:
+    """Generate a list of all filepaths in directory @dir relative to @root"""
+    yield from scandir_recursive(dir,
                                  lambda entry: Path(entry.path).relative_to(root))
 
 
-def make_cpio(state: MkosiState) -> None:
+def should_make_cpio(state: MkosiState) -> bool:
     if state.do_run_build_script:
-        return
+        return False
     if state.config.output_format != OutputFormat.cpio:
-        return
+        return False
     if state.for_cache:
-        return
+        return False
+    return True
 
-    with complete_step("Creating archive…"), open(state.staging / state.config.output.name, "wb") as f:
-        files = find_files(state.root)
+
+def make_cpio(root: Path, files: Iterator[Path], output: Path) -> None:
+    with complete_step("Creating archive…"), open(output, "wb") as f:
         cmd: list[PathString] = [
-            "cpio", "-o", "--reproducible", "--null", "-H", "newc", "--quiet", "-D", state.root
+            "cpio", "-o", "--reproducible", "--null", "-H", "newc", "--quiet", "-D", root
         ]
 
         with spawn(cmd, stdin=subprocess.PIPE, stdout=f) as cpio:
@@ -939,6 +955,23 @@ def make_cpio(state: MkosiState) -> None:
             cpio.stdin.close()
         if cpio.wait() != 0:
             die("Failed to create archive")
+
+
+def update_cpio(root: Path, files: Iterator[Path], output: Path) -> None:
+    with complete_step("Updating archive…"):
+        cmd: list[PathString] = [
+            "cpio", "-o", "-A", "-F", output, "--reproducible", "--null", "-H", "newc", "--quiet", "-D", root
+        ]
+
+        with spawn(cmd, stdin=subprocess.PIPE, stdout=subprocess.DEVNULL) as cpio:
+            #  https://github.com/python/mypy/issues/10583
+            assert cpio.stdin is not None
+
+            for file in files:
+                cpio.stdin.write(os.fspath(file).encode("utf8") + b"\0")
+            cpio.stdin.close()
+        if cpio.wait() != 0:
+            die("Failed to update archive")
 
 
 def make_directory(state: MkosiState) -> None:
@@ -962,13 +995,41 @@ def gen_kernel_images(state: MkosiState) -> Iterator[tuple[str, Path]]:
         yield kver.name, kimg
 
 
-def initrd_path(state: MkosiState, kver: str) -> Path:
+def initrd_paths(state: MkosiState, kver: str) -> list[Path]:
+    if state.config.initrds:
+        kmods_initrd = kernel_modules_initrd_path(state, kver)
+        assert kmods_initrd
+        return [*state.config.initrds, kmods_initrd]
+
     # initrd file is versioned in Debian Bookworm
     initrd = state.root / boot_directory(state, kver) / f"initrd.img-{kver}"
     if not initrd.exists():
         initrd = state.root / boot_directory(state, kver) / "initrd"
 
-    return initrd
+    return [initrd]
+
+
+def kernel_modules_initrd_path(state: MkosiState, kver: str) -> Optional[Path]:
+    if not state.config.initrds:
+        return None
+
+    compression = kernel_modules_initrd_compression(state.config)
+    if compression == "xz":
+        ext = "xz"
+    elif compression == "zstd":
+        ext = "zst"
+    else:
+        die(f"Unknown compression {compression}")
+
+    cpio_path = kernel_modules_initrd_cpio_path(state, kver)
+    assert cpio_path
+    return cpio_path.parent / f"{cpio_path.name}.{ext}"
+
+
+def kernel_modules_initrd_cpio_path(state: MkosiState, kver: str) -> Optional[Path]:
+    if not state.config.initrds:
+        return None
+    return state.workspace / f"initrd.{kver}.kmods"
 
 
 def install_unified_kernel(state: MkosiState, label: Optional[str], root_hash: Optional[str], usr_only: bool) -> None:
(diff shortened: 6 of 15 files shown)


mkosi/__init__.py
```diff
@@ -1020,13 +1022,28 @@ def gen_kernel_images(state: MkosiState) -> Iterator[Tuple[str, Path]]:
...
         initrd = state.root / boot_directory(state, kver) / "initrd"
 
-    return initrd
+    return [initrd]
+
+
+def kmods_initrd_path(state: MkosiState, kver: str) -> Optional[Path]:
+    if not state.config.initrd:
+        return None
+    return state.var_tmp() / f"initrd.{kver}.kmods.zst"
```
behrmann, line 1040: The zst extension seems a bit too specific, which distro did you test this with?

mkosi/__init__.py
```diff
@@ -968,22 +968,24 @@ def make_tar(state: MkosiState) -> None:
         run(cmd)
 
 
-def find_files(root: Path) -> Iterator[Path]:
+def find_files(dir: Path, root: Path) -> Iterator[Path]:
     """Generate a list of all filepaths relative to @root"""
```
behrmann, line 972: The docstring is wrong with this change.

mkosi/__init__.py
```diff
@@ -2636,6 +2677,11 @@ def unlink_output(config: MkosiConfig) -> None:
             unlink_try_hard(config.output_split_kernel_image)
             unlink_try_hard(config.output_split_initrd)
             unlink_try_hard(config.output_split_cmdline)
+            unlink_try_hard(config.output_split_kmods_initrd)
+            i = 0
+            while config.output_split_initrds(i).exists():
+                unlink_try_hard(config.output_split_initrds(i))
+                i += 1
```
behrmann, line 2684: Jugling indices is ugly and checking whether something exists before removing it is inherently racy. The problem is, that there's no way get all possible initrds without iterating over them until one doesn't exist (what if one in the middle is missing for some reason?), because then one could just `unlink_try_hard` each of them, since that's a noop if the file doesn't exist (anymore).

mkosi/__init__.py
```diff
@@ -3104,6 +3150,10 @@ def load_args(args: argparse.Namespace) -> MkosiConfig:
     if args.qemu_kvm and not qemu_check_kvm_support():
         die("Sorry, the host machine does not support KVM acceleration.")
 
+    if args.initrd:
+        for i in range(len(args.initrd)):
+            args.initrd[i] = args.initrd[i].absolute()
```
behrmann, line 3155:
```suggestion
        args.initrd  = [p.absolute() for p in args.initrd]
```

mkosi/__init__.py
```diff
@@ -1187,18 +1207,31 @@ def extract_kernel_image_initrd(state: MkosiState) -> None:
...
+    assert initrds
 
     copy_file(kimgabs, state.staging / state.config.output_split_kernel_image.name)
-    copy_file(initrd, state.staging / state.config.output_split_initrd.name)
+
+    if len(initrds) == 1:
+        copy_file(initrds[0], state.staging / state.config.output_split_initrd.name)
+    else:
+        for i, initrd in enumerate(initrds):
+            copy_file(initrd, state.staging / state.config.output_split_initrds(i).name)
```
behrmann, line 1228: I'm not sure I understand why there's a need to split this and have both `output_split_initrd` and `output_split_initrds`, which might get confusing.

behrmann (commented):
I'm not sure I get the full logic of the changes. A bit more explanation in the commit message might be helpful.

mkosi/__init__.py
rphibel: I actually control this extension. The "kmods" `initrd` is generated by `mkosi` for the kernel modules in new function [`generate_kmods_initrd`](https://github.com/systemd/mkosi/blob/47531782f9278eaf67d758148f44045827efc748/mkosi/__init__.py#L3547). This function generates a `cpio` archive from the `/usr/lib/modules` directory then compresses it with `zstd`.

mkosi/__init__.py
rphibel: well spotted!

mkosi/__init__.py
rphibel: Maybe I could use [`remove_glob`](https://github.com/systemd/mkosi/blob/aaaed2a5e409efbb1ddf3b87be14a92c08b08cc3/mkosi/__init__.py#L2591) here.

mkosi/__init__.py
rphibel: I did this to avoid a breaking change if people rely on the existing filename. I keep the existing name (without index) if there is only one initrd; otherwise I append an index.

rphibel:
> I'm not sure I get the full logic of the changes. A bit more explanation in the commit message might be helpful.

Thanks for the review.  I updated the commit message

mkosi/__init__.py
DaanDeMeyer: We should handle this a bit differently. Instead of outputting multiple initrds, we should output the combined initrd section that's generated by ukify. You can do that by using llvm-objcopy or objcopy with the --dump-section command.

`objcopy --dump-section .initrd=<output_split_initrd`. We should prefer using llvm-objcopy if it is available and fallback to objcopy otherwise.

DaanDeMeyer (changes requested):
Looks good but some comments

DaanDeMeyer:
I'm actually going to proposa a `--output-split-initrd` option for ukify so we don't have to call objcopy ourselves.

mkosi/__init__.py
rphibel: I have the following error when I use `llvm-copy` to extract the initrd:
```
llvm-objcopy: error: option is not supported for COFF
```

DaanDeMeyer: This actually shouldn't be necessary anymore after we merge https://github.com/systemd/mkosi/pull/1301

mkosi/__init__.py
```diff
@@ -3494,7 +3540,20 @@ def boot_directory(state: MkosiState, kver: str) -> Path:
     return Path("boot") / state.machine_id / kver
 
 
+def generate_kmods_initrd(state: MkosiState) -> None:
+    with complete_step("Generating initramfs images for kernel modules…"):
+        for kver, _ in gen_kernel_images(state):
+            cpio_path = kmods_initrd_cpio_path(state, kver)
+            assert cpio_path
+            make_cpio(state, find_files(state.root / "usr/lib/modules", state.root), str(cpio_path))
+            run(compressor_command("zstd", cpio_path))
```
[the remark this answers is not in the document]
rphibel: I don't think that I can do the same thing as in `compress_output()`. It seems to rely on `config.compress_output` setting. Here, I need to compress the cpio even if `compress_output` is not specified (if I don't compress the kernel module initrd, there is an error when I try to boot the image).

DaanDeMeyer:
If you rebase, output_split_initrd() will be removed and you can remove all changes related to that

mkosi/__init__.py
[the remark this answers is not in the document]
DaanDeMeyer: You can use something like `should_compress_output() or "zstd"` to default to zstd if nothing else is configured

docs/initrd.md
```diff
@@ -0,0 +1,10 @@
+To build an image with a mkosi-built initrd:
+1. Follow the steps described [here](https://github.com/systemd/mkosi-initrd/blob/main/docs/fedora.md)
```
DaanDeMeyer, line 2: Can we inline this in this document and make it non-fedora specific?

I would just replace this with the most minimal config we can find that produces a working initrd.

docs/initrd.md
```diff
@@ -0,0 +1,10 @@
+To build an image with a mkosi-built initrd:
```
DaanDeMeyer, line 1: Let's add a title here: `# Building a custom initrd and using it in a mkosi image`

.github/initrd/alma.conf
```diff
@@ -0,0 +1,14 @@
...
+[Output]
+ImageId=initrd
+Format=cpio
+ManifestFormat=
+CompressOutput=zstd
+
+[Content]
+BasePackages=conditional
+Packages=
+        systemd                   # sine qua non
```
DaanDeMeyer, line 13: Let's drop the superfluous comment

.github/initrd/alma.conf
```diff
@@ -0,0 +1,14 @@
+[Distribution]
+Distribution=alma
```
DaanDeMeyer, line 2: Alma, Rocky and CentOS are pretty much the same, so let's only provide a config file for centos and not alma/rocky

.github/initrd/arch.conf
```diff
@@ -0,0 +1,14 @@
...
+ImageId=initrd
+Format=cpio
+ManifestFormat=
+CompressOutput=zstd
+
+[Content]
+BasePackages=conditional
+Packages=
+        systemd                   # sine qua non
+        udev
```
DaanDeMeyer, line 14: udev is an alias for systemd on arch so this can be dropped

.github/initrd/arch.conf
```diff
@@ -0,0 +1,14 @@
...
+[Output]
+ImageId=initrd
+Format=cpio
+ManifestFormat=
+CompressOutput=zstd
+
+[Content]
+BasePackages=conditional
+Packages=
+        systemd                   # sine qua non
```
DaanDeMeyer, line 13: Let's drop the superfluous sine qua non comment everywhere

.github/initrd/centos.conf
```diff
@@ -0,0 +1,14 @@
...
+Distribution=centos
+
+[Output]
+ImageId=initrd
+Format=cpio
+ManifestFormat=
+CompressOutput=zstd
+
+[Content]
+BasePackages=conditional
```
DaanDeMeyer, line 11: We don't need `BasePackages=conditional` here, we'll only install a minimal set of packages by default even if this is enabled so we can just drop it

.github/initrd/gentoo.conf
```diff
@@ -0,0 +1,14 @@
...
+ImageId=initrd
+Format=cpio
+ManifestFormat=
+CompressOutput=zstd
+
+[Content]
+BasePackages=conditional
+Packages=
+        systemd                   # sine qua non
+        udev
```
DaanDeMeyer, line 14: udev is an alias to systemd on gentoo as well, no need to install it explicitly

DaanDeMeyer (changes requested):
We don't need the `mkosi.finalize` at all, that's doing the same what we're doing in mkosi itself in this PR by creating the kernel modules initrd, so we should just drop the `mkosi.finalize` part completely.

mkosi/__init__.py
```diff
@@ -3398,7 +3453,141 @@ def boot_directory(state: MkosiState, kver: str) -> Path:
...
+    for line_number, line in enumerate(f, start=1):
+        if not line.strip() or line.startswith('#'):
+            continue
+        if m := re.match(r'([A-Z][A-Z_0-9]+)=(.*)', line):
+            name, val = m.groups()
+            if val and val[0] in '"\'':
+                val = ast.literal_eval(val)
+            yield name, val
+        else:
+            print(f'Bad line {line_number}: {line}', file=sys.stderr)
```
DaanDeMeyer, line 3505: This function already exists in mkosi.backend

mkosi/__init__.py
```diff
@@ -3398,7 +3453,141 @@ def boot_directory(state: MkosiState, kver: str) -> Path:
...
+def make_init_symlink(root: Path) -> Path:
+    init = init_symlink_path(root)
+    init.unlink(missing_ok=True)
+    init.symlink_to('usr/lib/systemd/systemd')
+    return init
+
+
+def write_initrd_release(root: Path) -> Path:
+    initrd_release = initrd_release_path(root)
+    initrd_release.unlink(missing_ok=True)
```
DaanDeMeyer, line 3546: We should not remove existing initrd-release files. Let's only add one if none exists yet

mkosi/__init__.py
```diff
@@ -3398,7 +3453,141 @@ def boot_directory(state: MkosiState, kver: str) -> Path:
...
+    return root / 'init'
+
+
+def initrd_release_path(root: Path) -> Path:
+    return root / 'etc/initrd-release'
+
+
+def make_init_symlink(root: Path) -> Path:
+    init = init_symlink_path(root)
+    init.unlink(missing_ok=True)
```
DaanDeMeyer, line 3539: Let's only add this if it doesn't exist yet

mkosi/__init__.py
```diff
@@ -3408,6 +3597,10 @@ def run_kernel_install(state: MkosiState, cached: bool) -> None:
     if state.config.cache_initrd and cached:
         return
 
+    if state.config.initrds:
+        generate_kernel_modules_initrd(state)
+        return
```
DaanDeMeyer, line 3602: This should go under the `complete_step()` just below this

mkosi/__init__.py
```diff
@@ -3398,7 +3453,141 @@ def boot_directory(state: MkosiState, kver: str) -> Path:
...
+
+    return initrd_release
+
+
+def generate_init_and_initrd_release_files(root: Path) -> Iterator[Path]:
+    initrd_release = write_initrd_release(root)
+    yield initrd_release.relative_to(root)
+
+    init = make_init_symlink(root)
+    yield init.relative_to(root)
```
DaanDeMeyer, line 3559: This function only has a single caller so let's inline it in `find_files_for_kernel_modules_initrd()`

mkosi/__init__.py
```diff
@@ -3398,7 +3453,141 @@ def boot_directory(state: MkosiState, kver: str) -> Path:
...
+
+
+def update_suffixed(items: dict[str, str], name: str, fallback: str) -> None:
+    value = items.get(name, fallback)
+    if 'mkosi-initrd' in value:
+        return
+    items[name] = value + ' (mkosi-initrd)'
+
+
+def make_initrd_release(root: Path, out: TextIO) -> None:
```
DaanDeMeyer, line 3515: Let's inline this function into `write_initrd_release()` since it's the only caller of this function

mkosi/__init__.py
```diff
@@ -3398,7 +3453,141 @@ def boot_directory(state: MkosiState, kver: str) -> Path:
...
+    init = make_init_symlink(root)
+    yield init.relative_to(root)
+
+
+def cleanup_init_and_initrd_release_files(root: Path) -> None:
+    initrd_release = initrd_release_path(root)
+    initrd_release.unlink()
+
+    init = init_symlink_path(root)
+    init.unlink()
```
DaanDeMeyer, line 3567: You should write a function as follows:

```
@contextlib.contextmanager
def unlink(path: Path) -> Iterator[Path]:
    try:
        yield path
    finally:
        path.unlink(missing_ok=True)
```

You can then use that to replace this function in `generate_kernel_modules_initrd()` as follows:

```
with unlink(root / "init") as init, with unlink(root / "etc/initrd-release") as release:
```

mkosi/__init__.py
```diff
@@ -962,13 +979,39 @@ def gen_kernel_images(state: MkosiState) -> Iterator[tuple[str, Path]]:
...
+    else:
+        die(f"Unknown compression {compression}")
+
+    return state.var_tmp() / f"initrd.{kver}.kmods.{ext}"
+
+
+def kernel_modules_initrd_cpio_path(state: MkosiState, kver: str) -> Optional[Path]:
+    if not state.config.initrds:
+        return None
+    return state.var_tmp() / f"initrd.{kver}.kmods"
```
DaanDeMeyer, line 1014: This should go into `state.workspace`, `state.var_tmp()` is for files generated by tools running in the context of the image

mkosi/__init__.py
DaanDeMeyer, line 3559: Actually, we should do this differently. We shouldn't write these files to the final root directory. Instead, we should create a separate initrd root directory.

So you should create `state.workspace / "initrd"`, and then write the /init symlink and etc/initrd-release files in there. You can use the `cpio -A` switch to append files to an existing archive, so you can run cpio on `state.workspace / "initrd"` to append the files to the kernel modules initrd.

mkosi/__init__.py
```diff
@@ -3398,7 +3476,133 @@ def boot_directory(state: MkosiState, kver: str) -> Path:
...
+                if file not in kernel_modules_in_initrds:
+                    path = initrd_writers[file](state, kver)
+                    initrd_files.append(path.relative_to(initrd_dir_path(state, kver)))
+                else:
+                    MkosiPrinter.info(f"{file} found in {kernel_modules_in_initrds[file]}: skipping")
+            if initrd_files:
+                if cpio_path.exists():
+                    update_cpio(initrd_dir_path(state, kver), iter(initrd_files), cpio_path)
+                else:
+                    make_cpio(initrd_dir_path(state, kver), iter(initrd_files), cpio_path)
```
DaanDeMeyer, line 3598: So after thinking this over a little more, I think we should do this differently. Instead of adding these files in the kernel modules initrd, we should add them when we build the initial initrd cpio.

So instead of adding these here, when the output format is a cpio, we should add the symlink and /etc/initrd-release to the image before packing it up as a cpio.

DaanDeMeyer (changes requested):

mkosi/__init__.py
rphibel: > when we build the initial initrd cpio.

You mean the initrd that is provided by the user via the `--initrd` option?

DaanDeMeyer: Exactly, that initrd should have the init symlink and /etc/initrd-release, and we can easily add those when generating the initrd.

DaanDeMeyer:
Closing in favor of #1350

PR #1299, opened 2023-01-09
Add --credential option to set systemd credentials

DaanDeMeyer:


behrmann:
Missing test aside, LGTM.

mkosi/__init__.py
```diff
@@ -4146,6 +4166,9 @@ def run_qemu(config: MkosiConfig) -> None:
             "-append", config.output_split_cmdline.read_text().strip(),
         ]
 
+    for k, v in config.credentials.items():
+        cmdline += ["-smbios", f"type=11,value=io.systemd.credential:{k}={v}"]
```
bluca, line 4170: In some cases escaping is needed, in that case `io.systemd.credential.binary` can be used with the payload encoded in base64, can that be supported too?

Also, but can be done later, perhaps if the cred is passed to the `build` verb it could be included in the ESP instead

DaanDeMeyer: I figure both of those can be added later without any issues

PR #1300, opened 2023-01-10
setup.py: Use find_namespace_packages()

DaanDeMeyer:
This makes sure the repart/ directory in resources is included as well.


behrmann:
This might be too big a gun, since namespaces packages are a somewhat different beast entirely, if I remember correctly (e.g. if we wanted to factor out distros into separately installable packages). Not sure, but I think adding `mkosi/resources/repart/__init__.py` should have the files picked up as package data instead (via the existing `package_data` and `include_package_data` directives).

PR #1301, opened 2023-01-10
Remove --qemu-boot option

DaanDeMeyer:
Given that we're standardizing more and more on UEFI and unified kernel images, let's drop support for qemu direct linux boot, to drop yet another seldomly exercised codepath.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -206,16 +206,8 @@ jobs:
 
     - name: Boot ${{ matrix.distro }}/${{ matrix.format }} UEFI
       if: matrix.format == 'disk'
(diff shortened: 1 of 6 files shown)


DaanDeMeyer:
If users still want the individual initrd or linux kernel image it's trivial to get access to them with `objcopy --dump-section`

keszybz:
I'm always a bit apprehensive when dropping "features", but since you added this yourself, I'm pretty sure that you're in a very good position to judge when it's not useful anymore.

Centos CI fails with
```
Error: Transaction test error:
  file /usr/bin/bootctl conflicts between attempted installs of systemd-boot-250.3-1.el9.x86_64 and systemd-udev-252-2.el9.x86_64
```
which I thought we had resolved already, but apparently not.

PR #1302, opened 2023-01-10
Default to the distro preferred filesystem instead of ext4

DaanDeMeyer:


poettering:
uh btrfs is problematic. because it does the uuid-based superblock recognition in kernel, and that means you'll get EBUSY all the time if MS_LAZY unmounts are used things die abnormally. it's a mess.

i wish btrfs had some special mount flag to disable that behaviour, but until it has I think we should never default onto btrfs for anything.

DaanDeMeyer:
Can't we at least give this a try and see if we run into issues in mkosi or systemd CI? Sure btrfs might have issues but we can't really get around the fact that it's the default filesystem for Fedora and Opensuse

DaanDeMeyer:
I think the issue you mentioned will also be fixed by https://lore.kernel.org/linux-btrfs/837cdf72-727d-b46b-f5ab-acd16be7006a@oracle.com/t/. I'm working with the btrfs developers to get that merged.

poettering:
oh, that patch looks awesome. if that gets in this would alleviate my issues.

i mean, i am mostly concerned of the bug reports we'll get upstream in systemd, because people will deploy stuff like this and then are upset if they hit this. given this is a race condition it might not be visible in the CI and if accessed with our tools of choice, but once dropped into the real-world i am sur ethis will pop up all the time (because we get those reports in systemd upstream every now and then already)

DaanDeMeyer:
Interestingly enough, I can only trigger mount failures after booting the image in nspawn. Without booting, I can keep rebuilding the image without running into mount() failing with EINVAL.

DaanDeMeyer:
With https://github.com/systemd/systemd/pull/26022 merged, we should be OK to use btrfs by default for Fedora and Opensuse.

PR #1303, opened 2023-01-10
centos: Stop installing systemd-boot on CentOS Stream 9

DaanDeMeyer:
With CentOS Stream 9 moving to systemd v252, systemd-boot has been
added back to the systemd-udev package, so we don't need to install
it from EPEL anymore.


DaanDeMeyer:
This fixes CI so let's merge. If there's any feedback I'll address it in a follow up PR

PR #1305, opened 2023-01-13
Assorted changes

DaanDeMeyer:


behrmann:
Looks mostly good to me (disregarding the CI failure), but I'm not entirely sold on the `--epel` option. I mean, I'm fine with not having the EPEL variants being their own distros, but the commandline switch is very specific to the distros of that family. 

How about a `--distro-specific-option KEY:VALUE` switch, that works like our environment variable switch? This would be generally usable and punt all that from core to distro implementations.

DaanDeMeyer:
> Looks mostly good to me (disregarding the CI failure), but I'm not entirely sold on the --epel option. I mean, I'm fine with not having the EPEL variants being their own distros, but the commandline switch is very specific to the distros of that family.

Turns out we can just write the epel repository definition disabled by default and use `--repositories` to enable it. That way we don't need a separate option.

mkosi.md
```diff
@@ -320,23 +320,13 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
...
-: This option is only applicable for RPM-based distributions:
-  *CentOS*, *Fedora Linux*, *Mageia*, *Rocky Linux*, *Alma Linux*
-  and *OpenMandriva*.
-  Allows use of the host's existing RPM repositories.
-  By default, a hardcoded set of default RPM repositories is generated and used.
-  Use `--repositories=` to identify a custom set of repositories to be enabled
-  and used for the build.
-
-`RepositoryDirectory`, `--repository-directory`
+`RepositoryDirectories`, `--repository-directory`
```
behrmann, line 323: We already have a few cases, where the command line option is singular and the config file option is plural, but do we need to add more? Since this is a comma-delimited list that goes in as a single argument, I'd opt for
```suggestion
`RepositoryDirectories`, `--repository-directories`
```

behrmann (commented):
Nice cleanup. LGTM, except for the divergent naming of the option.

mkosi.md
DaanDeMeyer: Yeah, but when on the CLI, you're able to use this option as `--repository-directory` with a single argument. I think for now it's nice to be consistent with the other options in this regard.

PR #1307, opened 2023-01-16
Fix man page to mention apt requirement on host

DaanDeMeyer:


behrmann:
LGTM. No need to waste CI cycles, it's docs only.

PR #1308, opened 2023-01-16
Centos fixes

DaanDeMeyer:


behrmann:
LGTM, CI is only complaining about the no longer needed imports of `MkosiConfig` and the f-strings that don't need to be f-strings anymore.

PR #1310, opened 2023-01-17
centos: Remove "-stream" from releases

DaanDeMeyer:
We don't support non centos stream releases anymore, so let's remove the "-stream" suffix from centos releases.


behrmann (approved):
LGTM, but this should get a NEWS entry.

PR #1313, opened 2023-01-22
Enable unprivileged image builds

DaanDeMeyer:
To enable this, when doing a build, we unshare a user namespace
with it's own private set of uids/gids obtained using newuidmap
and newgidmap. We also map the current user to the last UID/GID
in the UID/GID range from /etc/subuid and /etc/subgid. Together
with unsharing the mount namespace, this allows us to do
unprivileged bind and overlay mounts.

Next, we replace all usages of systemd-nspawn during the image build
with bubblewrap. systemd-nspawn cannot run as an unprivileged user
yet so we use bubblewrap which can. bubblewrap can also be used to
setup a chroot environment with API VFS filesystems so we make use
of that to setup chroot environments and remove all our homegrown
logic for it. This allows us to significantly reduce the amount of
mounts we do in mkosi itself.

To further reduce the amount of mounts, we modify the invocations
of all package managers to specify the cache directory via the
relevant option instead of mounting the cache directory into the
chroot. For apt, to accomplish this, we switch from using
DPkg::Chroot-Directory to setting the "--root" option for each
invocation of dpkg so that dpkg can access files outside of the
chroot.

Finally, we remove some options which become obsolete with this
commit, --idmap, --chown and --nspawn-keep-unit.

We also remove --source-file-transfer, --source-file-transfer-final
and the corresponding symlink options. Instead, we default to mounting
source files into the build tree. In the future, we'll add virtiofsd
support to allow accessing source files in qemu VMs.

We also move stuff around and create a few new files to store
helpers to avoid circular imports. There's also a little bit of
refactoring and cleanup all around.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -534,13 +534,6 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
   image root, so any `CopyFiles=` source paths in partition definition files will
   be relative to the image root directory.
 
-`NoChown=`, `--no-chown`
-
-: By default, if `mkosi` is run inside a `sudo` environment all
-  generated artifacts have their UNIX user/group ownership changed to
-  the user which invoked `sudo`. With this option this may be turned
-  off and all generated files are owned by `root`.
-
 `TarStripSELinuxContext=`, `--tar-strip-selinux-context`
 
 : If running on a SELinux-enabled system (Fedora Linux, CentOS, Rocky Linux,
@@ -704,11 +697,8 @@ a machine ID.
 
 `BuildSources=`, `--build-sources=`
 
-: Takes a path to a source tree to copy into the development image, if
-  the build script is used. This only applies if a build script is
-  used, and defaults to the local directory. Use `SourceFileTransfer=`
-  to configure how the files are transferred from the host to the
-  container image.
+: Takes a path to a source tree to mount into the development image, if
+  the build script is used.
 
 `BuildDirectory=`, `--build-dir=`
 
@@ -792,33 +782,32 @@ a machine ID.
 : Takes a path to an executable that is used as build script for this
   image. If this option is used the build process will be two-phased
   instead of single-phased. The specified script is copied onto the
-  development image and executed inside an `systemd-nspawn` container
-  environment. If this option is not used, but the `mkosi.build` file
-  found in the local directory it is automatically used for this
-  purpose (also see the "Files" section below). Specify an empty value
-  to disable automatic detection.
+  development image and executed inside a namespaced chroot environment.
+  If this option is not used, but the `mkosi.build` file found in the
+  local directory it is automatically used for this purpose (also see
+  the "Files" section below). Specify an empty value to disable
+  automatic detection.
 
 `PrepareScript=`, `--prepare-script=`
 
 : Takes a path to an executable that is invoked inside the image right
   after installing the software packages. It is the last step before
   the image is cached (if incremental mode is enabled).  This script
-  is invoked inside a `systemd-nspawn` container environment, and thus
-  does not have access to host resources.  If this option is not used,
-  but an executable script `mkosi.prepare` is found in the local
-  directory, it is automatically used for this purpose. Specify an
-  empty value to disable automatic detection.
+  is invoked inside a namespaced chroot environment, and thus does not
+  have access to host resources.  If this option is not used, but an
+  executable script `mkosi.prepare` is found in the local directory, it
+  is automatically used for this purpose. Specify an empty value to
+  disable automatic detection.
 
 `PostInstallationScript=`, `--postinst-script=`
 
 : Takes a path to an executable that is invoked inside the final image
   right after copying in the build artifacts generated in the first
-  phase of the build. This script is invoked inside a `systemd-nspawn`
-  container environment, and thus does not have access to host
-  resources. If this option is not used, but an executable
-  `mkosi.postinst` is found in the local directory, it is
-  automatically used for this purpose. Specify an empty value to
-  disable automatic detection.
+  phase of the build. This script is invoked inside a namespaced chroot
+  environment, and thus does not have access to host resources. If this
+  option is not used, but an executable `mkosi.postinst` is found in the
+  local directory, it is automatically used for this purpose. Specify an
+  empty value to disable automatic detection.
 
 `FinalizeScript=`, `--finalize-script=`
 
@@ -831,38 +820,6 @@ a machine ID.
   executable `mkosi.finalize` is found in the local directory, it is
   automatically used for this purpose. Specify an empty value to
   disable automatic detection.
-
-`SourceFileTransfer=`, `--source-file-transfer=`
-
-: Configures how the source file tree (as configured with
-  `BuildSources=`) is transferred into the container image during the
-  first phase of the build. Takes one of `copy-all` (to copy all files
-  from the source tree), `copy-git-cached` (to copy only those files
-  `git ls-files --cached` lists), `copy-git-others` (to copy only
-  those files `git ls-files --others` lists), `mount` to bind mount
-  the source tree directly. Defaults to `copy-git-cached` if a `git`
-  source tree is detected, otherwise `copy-all`. When you specify
-  `copy-git-more`, it is the same as `copy-git-cached`, except it also
-  includes the `.git/` directory.
-
-`SourceFileTransferFinal=`, `--source-file-transfer-final=`
-
-: Same as `SourceFileTransfer=`, but for the final image instead of
-  the build image. Takes the same values as `SourceFileFransfer=`
-  except `mount`. By default, sources are not copied into the final
-  image.
-
-`SourceResolveSymlinks=`, `--source-resolve-symlinks`
-
-: If given, any symbolic links in the source file tree are resolved and the
-  file contents are copied to the build image. If not given, they are left as
-  symbolic links. This only applies if `SourceFileTransfer=` is `copy-all`.
-  Defaults to leaving them as symbolic links.
-
-`SourceResolveSymlinksFinal=`, `--source-resolve-symlinks-final`
-
-: Same as `SourceResolveSymlinks=`, but for the final image instead of
-  the build image.
 
 `WithNetwork=`, `--with-network`
 
@@ -966,13 +923,6 @@ a machine ID.
 
 : Space-delimited list of additional arguments to pass when invoking
   qemu.
-
-`NspawnKeepUnit=`, `--nspawn-keep-unit`
-
-: When used, this option instructs underlying calls of systemd-nspawn to
-  use the current unit scope, instead of creating a dedicated transcient
-  scope unit for the containers. This option should be used when mkosi is
-  run by a service unit.
 
 `Netdev=`, `--netdev`
 
@@ -1227,14 +1177,13 @@ local directory:
   image. The *development* image is used to build the project in the
   current working directory (the *source* tree). For that the whole
   directory is copied into the image, along with the `mkosi.build`
-  script. The script is then invoked inside the image (via
-  `systemd-nspawn`), with `$SRCDIR` pointing to the *source*
-  tree. `$DESTDIR` points to a directory where the script should place
-  any files generated it would like to end up in the *final*
-  image. Note that `make`/`automake`/`meson` based build systems
-  generally honor `$DESTDIR`, thus making it very natural to build
-  *source* trees from the build script. After the *development* image
-  was built and the build script ran inside of it, it is removed
+  script. The script is then invoked inside the image, with `$SRCDIR`
+  pointing to the *source* tree. `$DESTDIR` points to a directory where
+  the script should place any files generated it would like to end up
+  in the *final* image. Note that `make`/`automake`/`meson` based build
+  systems generally honor `$DESTDIR`, thus making it very natural to
+  build *source* trees from the build script. After the *development*
+  image was built and the build script ran inside of it, it is removed
   again. After that the *final* image is built, without any *source*
   tree or build script copied in. However, this time the contents of
   `$DESTDIR` are added into the image.
@@ -1574,7 +1523,7 @@ When not using distribution packages make sure to install the
 necessary dependencies. For example, on *Fedora Linux* you need:
 
 ```bash
-dnf install btrfs-progs apt debootstrap dosfstools mtools edk2-ovmf e2fsprogs squashfs-tools gnupg python3 tar xfsprogs xz zypper sbsigntools
+dnf install bubblewrap btrfs-progs apt debootstrap dosfstools mtools edk2-ovmf e2fsprogs squashfs-tools gnupg python3 tar xfsprogs xz zypper sbsigntools
 ```
 
 On Debian/Ubuntu it might be necessary to install the `ubuntu-keyring`,
@@ -1583,7 +1532,7 @@ in addition to `apt` and `debootstrap`, depending on what kind of distribution i
 you want to build. `debootstrap` on Debian only pulls in the Debian keyring
 on its own, and the version on Ubuntu only the one from Ubuntu.
 
-Note that the minimum required Python version is 3.7.
+Note that the minimum required Python version is 3.9.
 
 # REFERENCES
 * [Primary mkosi git repository on GitHub](https://github.com/systemd/mkosi/)
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -4,8 +4,6 @@ import argparse
 import configparser
 import contextlib
 import crypt
-import ctypes
-import ctypes.util
 import dataclasses
 import datetime
 import errno
@@ -18,6 +16,7 @@ import math
 import os
 import platform
 import re
+import resource
 import shlex
 import shutil
 import string
@@ -29,50 +28,26 @@ import uuid
 from collections.abc import Iterable, Iterator, Sequence
 from pathlib import Path
 from textwrap import dedent, wrap
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    BinaryIO,
-    Callable,
-    NoReturn,
-    Optional,
-    TextIO,
-    TypeVar,
-    Union,
-    cast,
-)
+from typing import Any, Callable, NoReturn, Optional, TextIO, TypeVar, Union, cast
 
 from mkosi.backend import (
-    ARG_DEBUG,
     Distribution,
     ManifestFormat,
     MkosiConfig,
-    MkosiException,
-    MkosiNotSupportedException,
-    MkosiPrinter,
     MkosiState,
     OutputFormat,
-    SourceFileTransfer,
     Verb,
-    chown_to_running_user,
+    current_user_uid_gid,
     detect_distribution,
-    die,
+    flatten,
+    format_rlimit,
     is_centos_variant,
     is_rpm_distribution,
-    mkdirp_chown_current_user,
-    nspawn_knows_arg,
-    nspawn_rlimit_params,
-    nspawn_version,
     patch_file,
     path_relative_to_cwd,
-    run,
-    run_workspace_command,
-    scandir_recursive,
     set_umask,
     should_compress_output,
-    spawn,
     tmp_dir,
-    warn,
 )
 from mkosi.install import (
     add_dropin_config,
@@ -81,9 +56,26 @@ from mkosi.install import (
     flock,
     install_skeleton_trees,
 )
+from mkosi.log import (
+    ARG_DEBUG,
+    MkosiException,
+    MkosiNotSupportedException,
+    MkosiPrinter,
+    die,
+    warn,
+)
 from mkosi.manifest import Manifest
-from mkosi.mounts import dissect_and_mount, mount_bind, mount_overlay, mount_tmpfs
+from mkosi.mounts import dissect_and_mount, mount_bind, mount_overlay, scandir_recursive
 from mkosi.remove import unlink_try_hard
+from mkosi.run import (
+    become_root,
+    fork_and_wait,
+    init_mount_namespace,
+    run,
+    run_workspace_command,
+    spawn,
+)
+from mkosi.types import PathString, TempDir
 
 complete_step = MkosiPrinter.complete_step
 color_error = MkosiPrinter.color_error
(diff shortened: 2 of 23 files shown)


DaanDeMeyer:
@257 Do you have a clue about the gentoo failures?

257:
maybe this:
https://github.com/systemd/mkosi/pull/1313/files#diff-e89c10e3e1a25022797c92a8ba44fda97927bbb6a59220a48084d3bf70378de8L174

`CAP_SYS_ADMIN,CAP_MKNOD`

mkosi/__init__.py
```diff
@@ -4354,6 +4276,40 @@ def needs_build(config: Union[argparse.Namespace, MkosiConfig]) -> bool:
...
+    subgid, subgidcount = read_subrange(Path("/etc/subgid"))
+
+    event = multiprocessing.Event()
+    pid = os.getpid()
+
+    child = os.fork()
+    if child == 0:
+        event.wait()
+        run(["newuidmap", str(pid), "0", str(os.getuid()), "1", "1", str(subuid), str(int(subuidcount) - 1)])
+        run(["newgidmap", str(pid), "0", str(os.getuid()), "1", "1", str(subgid), str(int(subgidcount) - 1)])
```
behrmann, line 4305:
```suggestion
        run(["newuidmap", f"{pid}, "0", f"{os.getuid()}", "1", "1", f"{subuid}", f"{subuidcount - 1}"])
        run(["newgidmap", f"{pid}", "0", f"{os.getuid()}", "1", "1", f"{subgid}", f"{subgidcount - 1}"])
```
Feels like less obstrusive than the `str` calls, also `subuidcount` and and `subgidcount` are already ints coming out of `read_subrange`.

Also this code might make more sense in `run.py`.

behrmann (commented):
Looks really nice!

257:
@DaanDeMeyer did that help?

DaanDeMeyer:
@257 I've been busy fixing some other issues, haven't had a time to check whether it works yet. I don't think it will though, since we're running as root in the user namespace, so we should already have all capabilities.

DaanDeMeyer:
So --format directory images have issues booting because systemd-nspawn does not map the directory owner to root in the container. However, even if it did, there's still the issue that there might be files owned by other users in the directory tree, which are mapped to UIDs outside the valid UID range due to us using user namespacing with `newuidmap` and `newgidmap`.  The only way we'll get the directory tree booting properly would be to use the exact same idmapping used when building the image. I first figured we could force chown all deviating stuff to the current user and rely on tmpfiles to fix things on first boot but then we'll still end up with files owned by random UIDs on the host afterwards.

DaanDeMeyer:
@257 Can you rework the gentoo implementation to store logs and such in `state.workspace` instead of in the root filesystem? Doing that won't work anymore when running as an unprivileged user.

257:
will have a look

DaanDeMeyer:
Reworked to use bubblewrap instead of unshare. unshare doesn't set up the API VFS properly and it turns out doing that ourselves isn't trivial. bubblewrap does what unshare does and has option that allow us to easily configure a API VFS for a chroot which simplifies the code a lot.

Also reworked run_build_script() to use run_workspace_command() which was long overdue. All package managers now use a new function run_with_apivfs() which sets up a chroot using bubblewrap and then invokes the passed command in it.

mkosi/__init__.py
```diff
@@ -3581,75 +3498,49 @@ def run_build_script(state: MkosiState) -> None:
     if state.config.build_script is None:
         return
 
-    idmap_opt = ":rootidmap" if nspawn_id_map_supported() and state.config.idmap else ""
+    os.makedirs(install_dir(state), mode=0o755, exist_ok=True)
```
behrmann, line 3501: Any reason against
```suggestion
    install_dir(state).mkdir(mode=0o755, exist_ok=True)
```

mkosi/distributions/__init__.py
```diff
@@ -29,3 +32,8 @@ def remove_packages(cls, state: "MkosiState", remove: list[str]) -> None:
     @classmethod
     def filesystem(cls) -> str:
         raise NotImplementedError
+
+
+def cache_params(state: "MkosiState") -> list[PathString]:
```
behrmann, line 37: This is quite specific to using bubblewrap, no? I guess that would be better places in `run.py`.

mkosi/__init__.py
behrmann: Not addressed.

DaanDeMeyer: To be fair, I disagree here, I think  using `str()` is much nicer than using f strings here

DaanDeMeyer: Addressed the other thing

behrmann: > To be fair, I disagree here, I think using `str()` is much nicer than using f strings here

Totally fair, this actually was more about the `int` calls.

DaanDeMeyer:
Another rework, I've given up on trying to make all files on the host owned by the user running mkosi, it just doesn't work when using directory output. Instead, the current user is mapped to the same uid in the user namespace, and then we do the same as we do when running as root, we just chown the necessary stuff when needed. Also, to avoid chowning more than necessary, when possible we run processes as the current user where applicable.

With this rework, when using --format=directory, we end up with a directory tree on the host completely owned by the first UID in the private UID range from /etc/subuid. This directory tree is usable with `systemd-nspawn --private-users` so it can be booted in nspawn, unlike the previous approach where that wasn't possible.

mkosi/__init__.py
```diff
@@ -202,7 +179,8 @@
             while str(p).startswith(str(config.build_sources)):
                 p = p.parent
 
-            d = tempfile.TemporaryDirectory(dir=p, prefix=f"mkosi.{config.build_sources.name}.tmp")
+            d = tempfile.TemporaryDirectory(dir=p, prefix=f".mkosi.{config.build_sources.name}.tmp")
+        os.chmod(d.name, 0o755)
```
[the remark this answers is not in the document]
behrmann: This should get a comment why the permisson needs to be like this.

mkosi/__init__.py
```diff
@@ -608,22 +562,8 @@ def configure_serial_terminal(state: MkosiState, cached: bool) -> None:
...
-def nspawn_params_for_build_sources(config: MkosiConfig, sft: SourceFileTransfer) -> list[str]:
-    params = ["--setenv=SRCDIR=/root/src",
-              "--chdir=/root/src"]
-    if sft == SourceFileTransfer.mount:
-        idmap_opt = ":rootidmap" if nspawn_id_map_supported() and config.idmap else ""
-        params += [f"--bind={config.build_sources}:/root/src{idmap_opt}"]
-
-    return params
+def script_env_params(state: MkosiState) -> list[str]:
+    return flatten(("--setenv", k, v) for k, v in state.environment.items())
```
behrmann, line 566: This should be moved to `run.py`, so `__init__.py` gets even emptier.

mkosi/__init__.py
```diff
@@ -3799,13 +3536,26 @@ def ensure_networkd(config: MkosiConfig) -> bool:
...
+def nspawn_knows_arg(arg: str) -> bool:
+    # Specify some extra incompatible options so nspawn doesn't try to boot a container in the current
+    # directory if it has a compatible layout.
+    return "unrecognized option" not in run(["systemd-nspawn", arg,
+                                            "--directory", "/dev/null", "--image", "/dev/null"],
+                                            stdout=subprocess.DEVNULL, stderr=subprocess.PIPE, check=False,
+                                            text=True).stderr
+
+
 def run_shell(config: MkosiConfig) -> None:
```
behrmann, line 3548: Any reason not to move `run_shell` into `run.py` as well?

mkosi/__init__.py
```diff
@@ -4334,9 +4079,71 @@ def needs_build(config: Union[argparse.Namespace, MkosiConfig]) -> bool:
...
+
+        if name == pwd.getpwuid(os.getuid()).pw_name:
+            break
+    else:
+        die(f"No mapping found for current user in {path}")
+
+    return int(start), int(count)
+
+
+def become_root() -> None:
```
behrmann, line 4094: I think this would also be a good candidate for `run.py`.

behrmann (commented):
Looks really good! Some nits about placement of functions. This cuts so much stuff out of `__init__.py` so maybe the new stuff can be added elsewhere as well.

Gentoo still has permission errors.

DaanDeMeyer:
> Gentoo still has permission errors.

Yeah Gentoo is going to need a big rework. We currently use system directories all over the place. This should all be changed to use the workspace instead. I also think we should move back to using `emerge` as a command line tool instead of a python import so we can run it using bubblewrap with `run_with_apivfs()`.

I really don't have the time or expertise to do this Gentoo rework so I hope @257 can take care of that. In the meantime I'd suggest we disable Gentoo in CI so we can land this.

mkosi/__init__.py
DaanDeMeyer: Depends on `MkosiConfig` and we can't import `backend` into `run`. Doing this requires splitting things up more, so I'm not doing it for now

DaanDeMeyer: PR is big already, let's leave moving the other run() functions in for another PR

mkosi/__init__.py
DaanDeMeyer: Moved to backend, I think it fits better there

DaanDeMeyer:
One more change. Instead of mapping the current user to the same uid in the user namespace, we map it to the last uid/gid in the subuid/subgid ranges from /etc/subuid and /etc/subgid. This should be safer and pretty much ensure we'll never be using a uid/gid that's actually used in the image itself as well.

mkosi/__init__.py
```diff
@@ -1163,15 +983,18 @@ def calculate_bmap(state: MkosiState) -> None:
         run(cmdline)
 
 
+def acl_allow_remove(root: Path, uid: int) -> None:
+    run(["setfacl", "-R", "-m", f"user:{uid}:rwx", root,
```
behrmann, line 987: By default a recursive setfacl will follow symlink arguments, but not symlinks it encounters later. If you can't think of a good reason against it, I'd rather throw in a `--physical` here just to be on the safe side.

Also, we should probably be lenient here, if this fails and just emit a warning, that people will have to clean up their stuff with elevated privileges, since I'm sure somebody will try this on an FS that can't do ACLs or only has them as opt-in (i.e. zfs).

behrmann:
LGTM, cleared the CI warnings. This is a pretty big one, though, so somebody else should look at it, too.

mkosi.md
```diff
@@ -792,32 +782,31 @@ a machine ID.
 : Takes a path to an executable that is used as build script for this
   image. If this option is used the build process will be two-phased
   instead of single-phased. The specified script is copied onto the
-  development image and executed inside an `systemd-nspawn` container
-  environment. If this option is not used, but the `mkosi.build` file
-  found in the local directory it is automatically used for this
-  purpose (also see the "Files" section below). Specify an empty value
-  to disable automatic detection.
+  development image and executed inside a chroot environment. If this
```
behrmann, line 785: I think we should be explicit here (and in the rest of the docs, where it says `chroot`), that we're using bubblewrap, just so that people know that the assurances are stronger than with a plain chroot.

mkosi/run.py
```diff
@@ -0,0 +1,273 @@
...
+    if libc.unshare(ctypes.c_int(flags)) != 0:
+        e = ctypes.get_errno()
+        raise OSError(e, os.strerror(e))
+
+
+def read_subrange(path: Path) -> tuple[int, int]:
+    for line in path.read_text().splitlines():
+        name, start, count = line.split(":")
+
+        if name == pwd.getpwuid(os.getuid()).pw_name:
```
behrmann, line 38: That's maybe an esoteric worry, but if the file contains UIDs that are not mapped on the host (where we run `pwd.getpwuid`) this will throw a `KeyError`.

mkosi/run.py
```diff
@@ -0,0 +1,273 @@
...
+        event.wait()
+
+        # We map the private UID range configured in /etc/subuid and /etc/subgid into the container using
+        # newuidmap and newgidmap. On top of that, we also make sure to map in the user running mkosi so that
+        # we can run still chown stuff to that user or run stuff as that user which will make sure any
+        # generated files are owned by that user.
+
+        newuidmap = [
+            "newuidmap", pid,
+            0, subuid, subuidcount - 100,
```
behrmann, line 67: What if `subuidcount` is smaller than 100? If we disallow that, this should be caught earlier.

mkosi.md
DaanDeMeyer: The thing is, bubblewrap on its own doesn't say much, since its a pretty generic tool

mkosi/run.py
DaanDeMeyer: I added enforcing that the range is of size 65536

mkosi.md
behrmann: That's true, although this might be a tad too vague, how about
```suggestion
  development image and executed inside a namespaced chroot environment. If this
```

It's not really ideal, since it doesn't say anything about what namespaces are being used, but reading it, will hopefully give people the idea that it's more than a simple chroot.

bluca:
> For apt, to accomplish this, we switch from using
DPkg::Chroot-Directory to setting the "--root" option for each
invocation of dpkg so that dpkg can access files outside of the
chroot.

Uhh - very few maintainer scripts actually support `$DPKG_ROOT` - are you sure this is working as intended and it's not running postinst/postrm scripts on the host filesystem?

DaanDeMeyer:
> > For apt, to accomplish this, we switch from using
> DPkg::Chroot-Directory to setting the "--root" option for each
> invocation of dpkg so that dpkg can access files outside of the
> chroot.
> 
> Uhh - very few maintainer scripts actually support `$DPKG_ROOT` - are you sure this is working as intended and it's not running postinst/postrm scripts on the host filesystem?

The option is documented as dpkg only chrooting for the install and for running the scripts. The entire thing is also running unprivileged, if any scripts were running in the host filesystem, I'd be running into permission errors (which I did when playing with --force-script-chrootless but not without it)

mkosi/backend.py
```diff
@@ -874,17 +543,12 @@ def sort_packages(packages: Iterable[str]) -> list[str]:
...
-                yield cast(T, pred)
-            if entry.is_dir(follow_symlinks=False):
-                queue.append(entry.path)
+def flatten(lists: Iterable[Iterable[T]]) -> list[T]:
+    """Flatten a sequence of sequences into a single list."""
+    return list(itertools.chain.from_iterable(lists))
+
+
+def current_user_uid_gid() -> tuple[int, int]:
+    uid = int(os.getenv("SUDO_UID") or os.getenv("PKEXEC_UID") or str(os.getuid()))
```
behrmann, line 552:
```suggestion
    uid = int(os.getenv("SUDO_UID") or os.getenv("PKEXEC_UID") or os.getuid())
```
`int()` is idempotent for ints, so not need to roundtrip this via `str()`

mkosi/run.py
```diff
@@ -0,0 +1,313 @@
...
+
+
+def init_namespace() -> None:
+    unshare(CLONE_NEWNS)
+    run(["mount", "--make-rslave", "/"])
+
+
+def foreground() -> None:
+    # If we're connected to a terminal, put the the child process in a new process group and make that
+    # the foreground process group so that only the child process receives SIGINT.
```
behrmann, line 116:
```suggestion
def foreground() -> None:
    """
    If we're connected to a terminal, put the the child process in a new process group and make that
    # the foreground process group so that only the child process receives SIGINT.
    """
```
Talking about the child process here is a bit confusing.

mkosi/run.py
```diff
@@ -0,0 +1,313 @@
...
+    # By default, we're root in the user namespace because if we were our current user by default, we
+    # wouldn't be able to chown stuff to be owned by root while the reverse is possible.
+    os.setresuid(0, 0, 0)
+    os.setresgid(0, 0, 0)
+    os.setgroups([0])
+
+    return SUBRANGE - 100, SUBRANGE - 100
+
+
+def init_namespace() -> None:
```
behrmann, line 109:
```suggestion
def init_mount_namespace() -> None:
    """Set up a mount namespace and one-way recursive mount propagation from the host"""
```

mkosi/run.py
```diff
@@ -0,0 +1,313 @@
...
+    else:
+        die(f"No mapping found for {user or uid} in {path}")
+
+    if int(count) != SUBRANGE:
+        die(f"subuid/subgid range length must be {SUBRANGE}, got {count} for {user or uid} from line '{line}'")
+
+    return int(start)
+
+
+def become_root() -> tuple[int, int]:
```
behrmann, line 59:
```suggestion
def become_root() -> tuple[int, int]:
    """Set up a new user namespace mapping using /etc/subuid and /etc/subgid.
    
    The current user will be mapped to root and 65436 will be mapped to the UID/GID of the invoking user. 
    The other IDs will bemapped through.
    
    The function returns the UID-GID pair of the invoking user in the namespace (65436, 65436).
    """
```
Shame docstrings can't be f-strings, Would be nice if the docs could be self-updating.

mkosi/run.py
```diff
@@ -0,0 +1,313 @@
...
+
+    for line in path.read_text().splitlines():
+        name, start, count = line.split(":")
+
+        if name == uid or name == user:
+            break
+    else:
+        die(f"No mapping found for {user or uid} in {path}")
+
+    if int(count) != SUBRANGE:
```
behrmann, line 53: That seems a bit strong. Maybe
```suggestion
    if int(count) < SUBRANGE:
```
so that it's at least this wide and we only handle the window of length `SUBRANGE`?

mkosi/run.py
```diff
@@ -0,0 +1,313 @@
...
+        os.setpgrp()
+        old = signal.signal(signal.SIGTTOU, signal.SIG_IGN)
+        os.tcsetpgrp(0, os.getpgrp())
+        signal.signal(signal.SIGTTOU, old)
+
+
+# Stores the exception from a subprocess along with its traceback. We have to do this explicitly because the
+# original traceback object cannot be pickled. When stringified, produces the subprocess stacktrace plus the
+# exception message.
+class RemoteException(Exception):
```
behrmann, line 127:
```suggestion
class RemoteException(Exception):
    """
    Stores the exception from a subprocess along with its traceback. We have to do this explicitly because the
    original traceback object cannot be pickled. When stringified, produces the subprocess stacktrace plus the
    exception message.
    """
```

mkosi/run.py
```diff
@@ -0,0 +1,313 @@
...
+    def __init__(self, e: BaseException, tb: traceback.StackSummary):
+        self.exception = e
+        self.tb = tb
+
+    def __str__(self) -> str:
+        return f"Traceback (most recent call last):\n{''.join(self.tb.format()).strip()}\n{type(self.exception).__name__}: {self.exception}"
+
+
+# Attach to sys.excepthook to automically format exceptions with a RemoteException attached correctly.
+def excepthook(exctype: Type[BaseException], exc: BaseException, tb: Optional[TracebackType]) -> None:
```
behrmann, line 137:
```suggestion
def excepthook(exctype: Type[BaseException], exc: BaseException, tb: Optional[TracebackType]) -> None:
    """Attach to sys.excepthook to automically format exceptions with a RemoteException attached correctly."""
```

mkosi/run.py
```diff
@@ -0,0 +1,313 @@
...
+    if isinstance(exc.__cause__, RemoteException):
+        print(exc.__cause__, file=sys.stderr)
+    else:
+        sys.__excepthook__(exctype, exc, tb)
+
+
+def fork_and_wait(target: Callable[[], T]) -> T:
```
behrmann, line 144:
```suggestion
def fork_and_wait(target: Callable[[], T]) -> T:
    """Run target function in a child process and collect its backtrace if there is one."""
```

mkosi/__init__.py
```diff
@@ -4356,16 +4116,18 @@ def run_verb(raw: argparse.Namespace) -> None:
...
 
-        if config.verb == Verb.summary:
-            print_summary(config)
-
         if needs_build(config):
-            check_native(config)
-            init_namespace()
-            build_stuff(config)
+            def target() -> None:
+                uid, gid = become_root() if os.getuid() != 0 else current_user_uid_gid()
```
behrmann, line 4123:
```suggestion
                # Get the user UID/GID either on the host or in user ns running the build
                uid, gid = become_root() if os.getuid() != 0 else current_user_uid_gid()
```

behrmann (commented):
This is some really great work with some gnarly low-level Python.

The suggestions are mostly cosmetic stuff, to move comments to docstrings and add some more comments and docstrings to make this code easier to follow for the people who haven't written it. :)

mkosi/run.py
```diff
@@ -0,0 +1,326 @@
...
+
+    for line in path.read_text().splitlines():
+        name, start, count = line.split(":")
+
+        if name == uid or name == user:
+            break
+    else:
+        die(f"No mapping found for {user or uid} in {path}")
+
+    if int(count) <= SUBRANGE:
```
behrmann, line 53: Now if the range is exactly 2¹⁶ this will break.

bluca:
> > > For apt, to accomplish this, we switch from using
> > > DPkg::Chroot-Directory to setting the "--root" option for each
> > > invocation of dpkg so that dpkg can access files outside of the
> > > chroot.
> > 
> > 
> > Uhh - very few maintainer scripts actually support `$DPKG_ROOT` - are you sure this is working as intended and it's not running postinst/postrm scripts on the host filesystem?
> 
> The option is documented as dpkg only chrooting for the install and for running the scripts. The entire thing is also running unprivileged, if any scripts were running in the host filesystem, I'd be running into permission errors (which I did when playing with --force-script-chrootless but not without it)

The problem is that all scripts must be updated to the use $DPKG_ROOT env var, and most are not, because the scope for that is to allow cross-bootstrapping new architectures, so only a select few packages in the essential set have been updated.
So, I'd recommend to add a check that euid != 0 before doing that, because if it is ran as root it might do some serious damage to the host. Yes, this sucks.

DaanDeMeyer:
> The problem is that all scripts must be updated to the use $DPKG_ROOT env var, and most are not, because the scope for that is to allow cross-bootstrapping new architectures, so only a select few packages in the essential set have been updated.
So, I'd recommend to add a check that euid != 0 before doing that, because if it is ran as root it might do some serious damage to the host. Yes, this sucks.

But the thing is that with this option, dpkg still chroots() before invoking a maintainer script. So they don't need to support `DPKG_ROOT`. With the apt option `DPkg::Chroot-Directory`, apt does the chroot before invoking dpkg, so dpkg can't access outside the chroot at all. When using `--root`, dpkg will chroot itself, and only when needed.

bluca:
> > The problem is that all scripts must be updated to the use $DPKG_ROOT env var, and most are not, because the scope for that is to allow cross-bootstrapping new architectures, so only a select few packages in the essential set have been updated.
> > So, I'd recommend to add a check that euid != 0 before doing that, because if it is ran as root it might do some serious damage to the host. Yes, this sucks.
> 
> But the thing is that with this option, dpkg still chroots() before invoking a maintainer script. So they don't need to support `DPKG_ROOT`. With the apt option `DPkg::Chroot-Directory`, apt does the chroot before invoking dpkg, so dpkg can't access outside the chroot at all. When using `--root`, dpkg will chroot itself, and only when needed.

ok, I guess I'm getting confused as they are named in the same way, but if you double-checked that chroot is always used then it's fine

257:
@DaanDeMeyer you fixed gentoo too!? it works for me here, i'm updating `mkosi.ebuild`. thanks.

DaanDeMeyer:
@257 I didn't fix Gentoo unfortunately. Can you verify you can build Gentoo images without root from main?

257:
it builds here, i only had to install `bubblewrap`. are tests disabled on main? maybe we should test with CI and see what breaks and i will pick it up from there. but here, on a gentoo machine, build finished fine. didn't test for qemu though.

DaanDeMeyer:
Can you make a PR to enable gentoo again, that should surface some failures.

257:
will do in a minute, qemu works here too btw.

PR #1317, opened 2023-01-24
centos: Use ext4 as the default filesystem instead of xfs

DaanDeMeyer:


DaanDeMeyer:
Relevant CIs passed, merging

PR #1318, opened 2023-01-25
repart: check all created partitions to determine disk layout

malt3:
Check all partitions created by sd-repart to determine disk layout.

PR was remodeled. Original issue was resolved by https://github.com/systemd/systemd/pull/26199
Original description below

---

The extraction of the `roothash` or `usrhash` currently depends on the order of partitions (naming of the repart config files).

## Expected behavior:

Kernel commandline should include `roothash=...` to be picked up by [systemd-veritysetup-generator](https://www.freedesktop.org/software/systemd/man/systemd-veritysetup-generator.html).

## Actual behavior:

Kernel commandline includes `root=LABEL=...` instead.

## Explanation

This current logic in `invoke_repart` leads to empty hashes if the `root-<ARCH>` partition comes before the `root-<ARCH>-verity` partition (or if the `usr-<ARCH>` partition comes before the `usr-<ARCH>-verity` partition).
The logic should check if the partition type ends with `-verity` (corresponding to the repart config option `Verity=hash`) when extracting the `roothash`.

## How to reproduce

Config files to reproduce the issue:

`mkosi.repart/00-esp.conf`:

```
[Partition]
Type=esp
Format=vfat
CopyFiles=/boot:/
SizeMinBytes=256M
SizeMaxBytes=512M
```

`mkosi.repart/10-root.conf`:

```
[Partition]
Type=root
Format=squashfs
Verity=data
VerityMatchKey=root
CopyFiles=/
Minimize=guess
```

`mkosi.repart/20-root-verity.conf`:

```
[Partition]
Type=root-verity
Verity=hash
VerityMatchKey=root
SizeMinBytes=64M
SizeMaxBytes=64M
```

JSON output of sd-repart:

```json
[
    {
        "type": "esp",
        "label": "esp",
        "uuid": "da584991-3b7c-4868-a2f6-f82b79dc70db",
        "file": "/home/malte/proj/mkosi.repart/00-resp.conf",
        "node": "/home/malte/proj/mkosi.image.tmps5gepw44/staging/image.raw1",
        "offset": 1048576,
        "old_size": 0,
        "raw_size": 268435456,
        "old_padding": 0,
        "raw_padding": 0,
        "activity": "create",
        "roothash": null
    },
    {
        "type": "root-x86-64",
        "label": "root-x86-64",
        "uuid": "39c082ea-bb08-b549-b946-f618d4cc962b",
        "file": "/home/malte/proj/mkosi.repart/10-root.conf",
        "node": "/home/malte/proj/mkosi.image.tmps5gepw44/staging/image.raw2",
        "offset": 269484032,
        "old_size": 0,
        "raw_size": 375660544,
        "old_padding": 0,
        "raw_padding": 0,
        "activity": "create",
        "roothash": null
    },
    {
        "type": "root-x86-64-verity",
        "label": "root-x86-64-verity",
        "uuid": "7091afec-c827-638b-28bf-87c4f16ef738",
        "file": "/home/malte/proj/mkosi.repart/20-root-verity.conf",
        "node": "/home/malte/proj/mkosi.image.tmps5gepw44/staging/image.raw3",
        "offset": 645144576,
        "old_size": 0,
        "raw_size": 67108864,
        "old_padding": 0,
        "raw_padding": 0,
        "activity": "create",
        "roothash": "39c082eabb08b549b946f618d4cc962b7091afecc827638b28bf87c4f16ef738"
    }
]
```

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3497,12 +3497,30 @@ def invoke_repart(
 
     output = json.loads(run(cmdline, stdout=subprocess.PIPE).stdout)
 
+    usr_label = None
+    root_label = None
+    usr_hash = None
+    root_hash = None
+    has_usr = False
+    has_root = False
     for p in output:
-        if p["type"].startswith("usr") or p["type"].startswith("root"):
-            usr_only = p["type"].startswith("usr")
-            return (p["label"], p.get("roothash"), usr_only)
+        is_data = not p["type"].endswith("-verity")
+        usr = p["type"].startswith("usr") and is_data
+        root = p["type"].startswith("root") and is_data
+        has_usr |= usr
+        has_root |= root
+        if usr:
+            usr_label = p["label"]
+            usr_hash = p.get("roothash")
+        elif root:
+            root_label = p["label"]
+            root_hash = p.get("roothash")
 
-    return (None, None, False)
+    usr_only = has_usr and not has_root
+    label = usr_label if usr_only else root_label
+    root_hash = usr_hash if usr_only else root_hash
+
+    return (label, root_hash, usr_only)
 
 
 def build_image(state: MkosiState, *, manifest: Optional[Manifest] = None) -> None:


behrmann:
Thanks for the detailed report! systemd/systemd#26199, which is a reaction to this PR, should make this easier, because this solution - while definitely better - still has some edge cases. Since we currently require an upstream systemd for the main branch of mkosi, we should wait for this.

malt3:
Maybe I can rework this PR to result in better detection of `usr_only` then?
The old code will set `usr_only` to true if a `usr` partition is found first.
More correct would be:

- If a root partition exists, `usr_only` should be false
- If a usr partition exists and no root partition, `usr_only` should be true
- Otherwise `usr_only` should be false

behrmann:
Yes, that would be very welcome.

DaanDeMeyer:
@malt3 I opened https://github.com/systemd/mkosi/pull/1323 to simplify this a bit, can you check if that works for you?

PR #1319, opened 2023-01-25
Rework copying

DaanDeMeyer:
Let's just delegate to `cp` and get rid of a ton of cruft and speed things up at the same time.


behrmann:
LGTM, only nit I'd prefer long options for subprocesses instead of shortopts, since I find them easier to understand when skimming the code.

DaanDeMeyer:
Opensuse CI failure seems unrelated, I can build fine locally

PR #1321, opened 2023-01-26
Remove PathString in favour of CommandLine types

behrmann:
This is a followup to #1320, but can wait until after #1313, if it makes rebasing harder.

Except for two remaining uses, where using the explicit `Union[Path, str]`, which we will at some point be able to exchange for `Path | str`, is no bother and doesn't warrant keeping around an additional type alias, the remaining users for `PathString` are either `list[PathString]`, `Sequence[PathString]` or `Iterable[PathString]`.

The latter can be exchanged for `Sequence[PathString]`, leaving us with one mutable and one immutable version, that can be nicely aliased as `CommandLine` and `CommandLineReadOnly`, making clearer what is being handled and expected.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -73,7 +73,7 @@ from mkosi.run import (
     run_workspace_command,
     spawn,
 )
-from mkosi.types import PathString
+from mkosi.types import CommandLine, MutableCommandLine
 
 complete_step = MkosiPrinter.complete_step
 color_error = MkosiPrinter.color_error
@@ -120,7 +120,7 @@ def list_to_string(seq: Iterator[str]) -> str:
     return str(list(seq))[1:-1]
 
 
-def print_running_cmd(cmdline: Iterable[PathString]) -> None:
+def print_running_cmd(cmdline: CommandLine) -> None:
     MkosiPrinter.print_step("Running command:")
     MkosiPrinter.print_step(" ".join(shlex.quote(str(x)) for x in cmdline) + "\n")
 
@@ -500,7 +500,7 @@ def configure_serial_terminal(state: MkosiState, cached: bool) -> None:
                           """)
 
 
-def cache_params(state: MkosiState, root: Path) -> list[PathString]:
+def cache_params(state: MkosiState, root: Path) -> MutableCommandLine:
     return flatten(("--bind", state.cache, root / p) for p in state.installer.cache_path())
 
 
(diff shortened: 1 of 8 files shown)


mkosi/types.py
```diff
@@ -19,4 +19,6 @@
 _FILE = Union[None, int, IO[Any]]
 PathString = Union[Path, str]
 
-CommandArgument = Union[str, Path]
+CommandArgument = Union[str, Path, int]
+CommandLine = list[CommandArgument]
+CommandLineReadOnly = Sequence[CommandArgument]
```
DaanDeMeyer, line 23: I'd prefer not to introduce these two, I feel they obfuscate what the type really is without adding much value

mkosi/run.py
```diff
@@ -230,6 +230,9 @@ def run(
             TERM=os.getenv("TERM", "vt220"),
         ) | env
 
+    if kwargs.get("shell"):
+        cmd = [shlex.quote(s) for s in cmd]
```
DaanDeMeyer, line 234: Do we actually use `shell` anywhere? If not I wouldn't bother adding this until we do

DaanDeMeyer (changes requested):

mkosi/run.py
behrmann: We don't, but I stumbled over the `shlex.join` for printing the cmdline in one of the run functions, which is a bit performative, but then I thought that this is a potential footgun, since we pass `kwargs` through.

mkosi/types.py
behrmann: My thinking was, that this actually makes this clearer (and shorter), because it clearly communicates what functions return and take, e.g. something is not just a list of strings or a list of paths or an atom of some command line, but this is explicitly a command line being returned and this is a command line being used without changing it, without having to be too deeply into Python typing and knowing the differences between `list`, `Iterable` and `Sequence`.

mkosi/run.py
DaanDeMeyer: This wouldn't even work right, since we have to join the command to a string when shell is used. I'd just drop this.

mkosi/__init__.py
```diff
@@ -73,7 +73,7 @@
     run_workspace_command,
     spawn,
 )
-from mkosi.types import PathString
+from mkosi.types import CommandLine, CommandLineReadOnly
```
DaanDeMeyer, line 76: Let's switch this around, CommandLine should be read-only by default and we should add MutableCommandLine for when we want to be able to modify it.

DaanDeMeyer (changes requested):

DaanDeMeyer:
Closing until you can get back to this

PR #1323, opened 2023-01-27
Rework roothash handling

DaanDeMeyer:
First, we clean up the leftover remnants of our qemu direct linux boot support from install_unified_kernel(). Then we simplify passing around data from invoke_repart() to install_unified_kernel() by only passing around a single argument, the roothash/usrhash argument. Finally, now that repart has "roothash" set on every verity sibling partition, we can just read "roothash" off of any of them. We also change the logic so that if we find a roothash, it always has precedence over a usrhash.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -940,7 +940,7 @@ def initrd_path(state: MkosiState, kver: str) -> Path:
     return initrd
 
 
-def install_unified_kernel(state: MkosiState, label: Optional[str], root_hash: Optional[str], usr_only: bool) -> None:
+def install_unified_kernel(state: MkosiState, roothash: Optional[str]) -> None:
     # Iterates through all kernel versions included in the image and generates a combined
     # kernel+initrd+cmdline+osrelease EFI file from it and places it in the /EFI/Linux directory of the ESP.
     # sd-boot iterates through them and shows them in the menu. These "unified" single-file images have the
@@ -983,8 +983,9 @@ def install_unified_kernel(state: MkosiState, label: Optional[str], root_hash: O
 
             if state.config.image_version:
                 boot_binary = state.root / prefix / f"EFI/Linux/{image_id}_{state.config.image_version}{boot_count}.efi"
-            elif root_hash:
-                boot_binary = state.root / prefix / f"EFI/Linux/{image_id}-{kver}-{root_hash}{boot_count}.efi"
+            elif roothash:
+                _, _, h = roothash.partition("=")
+                boot_binary = state.root / prefix / f"EFI/Linux/{image_id}-{kver}-{h}{boot_count}.efi"
             else:
                 boot_binary = state.root / prefix / f"EFI/Linux/{image_id}-{kver}{boot_count}.efi"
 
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -3466,12 +3456,19 @@ def invoke_repart(
...
 
+    roothash = usrhash = None
     for p in output:
-        if p["type"].startswith("usr") or p["type"].startswith("root"):
-            usr_only = p["type"].startswith("usr")
-            return (p["label"], p.get("roothash"), usr_only)
+        if (h := p.get("roothash")) is None:
+            continue
+
+        assert p["type"].startswith("usr") or p["type"].startswith("root")
```
behrmann, line 3464: We use `asserts` in many places, but I'd like to get rid of them. They can be compiled out and their error is a bit nondescript. 
```suggestion
        if no (p["type"].startswith("usr") or p["type"].startswith("root")):
            raise MkosiException(f"Unknown verity partition type {p['type']}")
```

mkosi/__init__.py
```diff
@@ -3466,12 +3456,19 @@ def invoke_repart(
...
+
+        if p["type"].startswith("usr"):
+            usrhash = usrhash or h
+        else:
+            roothash = roothash or h
```
behrmann, line 3469: So, the first one of either we find always wins, e.g. having multiple `/usr/` in a single image is nothing we need to think about? Either way a comment would be helpful for the future.

mkosi/__init__.py
```diff
@@ -983,8 +983,9 @@ def install_unified_kernel(state: MkosiState, label: Optional[str], root_hash: O
 
             if state.config.image_version:
                 boot_binary = state.root / prefix / f"EFI/Linux/{image_id}_{state.config.image_version}{boot_count}.efi"
-            elif root_hash:
-                boot_binary = state.root / prefix / f"EFI/Linux/{image_id}-{kver}-{root_hash}{boot_count}.efi"
+            elif roothash:
+                h = roothash[roothash.find('=') + 1:]
```
behrmann, line 987:
```suggestion
                _, _, h = roothash.partition("=")
```

mkosi/__init__.py
```diff
@@ -3466,12 +3456,21 @@ def invoke_repart(
...
+            die(f"Found roothash property on unexpected partition type {p['type']}")
+
+        # When there's multiple verity enabled root or usr partitions, the first one wins.
+        if p["type"].startswith("usr"):
+            usrhash = usrhash or h
+        else:
+            roothash = roothash or h
 
-    return (None, None, False)
+    return f"roothash={roothash}" if roothash else f"usrhash={usrhash}"
```
behrmann, line 3473: If we find neither, this should probably not return `usrhash=None` or am I missing something?
```python
            # having both root and usr verity is not supported, always prefer root
            if p["type"].startswith("root"):
                return f"roothash={h}"
            # always prefer the first verity partition we find
            usrhash = usrhash or h
            
    return f"usrhash={usrhash}" if usrhash or None

DaanDeMeyer: Oops

behrmann:
I of course prefer my above version, but this should work now.

DaanDeMeyer:
Let's merge this already, we can iterate more in a follow up PR

PR #1324, opened 2023-01-27
centos: Refactor epel handling

DaanDeMeyer:
- Use URLs from CentOS itself and metalinks everywhere.
- Add epel-testing, disabled by default
- Move everything into a common function


DaanDeMeyer:
Merging early because I need it for systemd

PR #1325, opened 2023-01-29
centos: Install epel-release in both build and final images

DaanDeMeyer:
This allows mkosi.prepare scripts to install extra dependencies from epel as well.


DaanDeMeyer:
Trivial so merging immediately

PR #1327, opened 2023-01-31
Create secure boot auto enrollment files

medhefgo:


medhefgo:
This should make it way easier to test secure boot things and especially would allow giving the auto-enrollment in systemd some ci coverage.

Though, I am not too sure if this should be in mkosi or done as part of the systemd ci… (I assume mkosi is supposed to be for vm/container deployment only, otherwise forced enrollment isn't the greatest of ideas)

mkosi/__init__.py
```diff
@@ -1064,6 +1064,49 @@ def secure_boot_sign(state: MkosiState, directory: Path, replace: bool = False)
                 os.rename(f"{f}.signed", f)
 
 
+def secure_boot_create_auto_enroll(state: MkosiState) -> None:
```
DaanDeMeyer, line 1067: Let's call this `secure_boot_configure_auto_enroll()`

DaanDeMeyer (changes requested):
We have logic to detect whether we're in a hypervisor now in systemd-boot, can't we modify the secure boot enrollment to automatically enroll keys in auto/ when running in a hypervisor?

DaanDeMeyer:
Also, any chance we can use `sbsigntools` for this? I tried that a long time ago in https://github.com/systemd/mkosi/pull/651 but it didn't work but I'm hoping the situation's changed. Otherwise, can we migrate the existing sbsign stuff to efitools? Preferably we wouldn't have to depend on both

medhefgo:
> We have logic to detect whether we're in a hypervisor now in systemd-boot, can't we modify the secure boot enrollment to automatically enroll keys in auto/ when running in a hypervisor?

It currently defaults to manual, while force would always auto-enroll (with a timeout on non-vms). The idea was to eventually add a "auto-if-safe" mode, but for some reason that was never realized. I guess now there is a incentive for it once this is in mkosi…

> Also, any chance we can use `sbsigntools` for this?

Turns out there is. I knew about sbsiglist, but for some reason it didn't occur to me that sbvarsign could be used to create the signed signature lists. And why sbsiglist wants a DER certificate is beyond me. Also, sbvarsign only creates correct results if passing the attributes because otherwise it defaults to append mode, which gets rejected by ovmf for some reason.

mkosi/__init__.py
```diff
@@ -1064,6 +1064,65 @@ def secure_boot_sign(state: MkosiState, directory: Path, replace: bool = False)
...
+        run(
+            [
+                "openssl",
+                "x509",
+                "-outform",
+                "DER",
+                "-in",
+                state.config.secure_boot_certificate,
+                "-out",
+                keys_dir / "mkosi.der",
```
DaanDeMeyer, line 1091: Do we need the `der` in the final image? If not, you can put it in `state.workspace` instead so it'll get auto cleaned up

medhefgo: Nope. Same with the esl. Good to know I can put them there.

behrmann (approved):

PR #1328, opened 2023-01-31
stuff

DaanDeMeyer:


behrmann:
I don't quite share the sentiment of the commit message to drop the config parser tests, that they've never found anything useful - but it is exceedingly rare, that much is true. I'd keep them around, but if that stands in the way of dropping `--all` and `--all-directory` then I'm fine with that. Rest LGTM.

PR #1329, opened 2023-02-02
Fix --checksum true

manuelm:
File should be opened for reading


behrmann (approved):

behrmann:
Thanks for the contribution and picking that up, I missed that reviewing the repart PR.

PR #1331, opened 2023-02-08
Ensure apivfs directories are not included in the final image

DaanDeMeyer:


mkosi/__init__.py
```diff
@@ -3399,6 +3399,17 @@ def run_preset_all(state: MkosiState) -> None:
         run(["systemctl", "--root", state.root, "preset-all"])
 
 
+def remove_apivfs_dirs(state: MkosiState):
```
behrmann, line 3402:
```suggestion
def remove_apivfs_dirs(state: MkosiState) -> None:
```
to make mypy happy.

mkosi/__init__.py
```diff
@@ -3399,6 +3399,17 @@ def run_preset_all(state: MkosiState) -> None:
...
+    for d in ("run", "dev", "tmp", "proc", "sys"):
```
behrmann, line 3403: Should `/tmp` be necesarily included here, i.e. I'm not sure I would include it among API VFS directories.

poettering:
hmm? why would you remove these? you can't mount anything on them if you remove them, so immutable images will break.

PR #1332, opened 2023-02-09
Drop --include-dir

DaanDeMeyer:
Let's get rid of one more source of mounts. --include-dir can be replaced by using --incremental and reading includes from the cached build image tree. In the future, once build images become regular images, users can read includes by simply using --directory output and reading includes from the regular image output tree.


behrmann:
I'm fine with the change, but the upgrade instructions could be slightly less terse. :)

DaanDeMeyer:
I mean, even I haven't used this in ages, I'm pretty sure no one else is either.

PR #1333, opened 2023-02-09
Simplify kernel image and initrd handling

DaanDeMeyer:
- Let's stop writing files in /etc in favor of passing the
information via other ways
- Let's stop defaulting to "bls" layout which is intended
for type 1 images, we only use UKIs so we don't need the
"bls" layout
- kernel-install now defaults to the "other" layout, which
means it won't create the entry directory in /boot anymore.
We update the initrd find logic to take this into account
- Remove --machine-id as it was only really there for testing
the config parsing which we now deal with by not storing the
machine ID at all

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -249,27 +249,6 @@ def prepare_tree_root(state: MkosiState) -> None:
             btrfs_subvol_create(state.root)
 
 
-def prepare_tree(state: MkosiState, cached: bool) -> None:
-    if cached:
-        # Reuse machine-id from cached image.
-        state.machine_id = uuid.UUID(state.root.joinpath("etc/machine-id").read_text().strip()).hex
-        # Always update kernel command line.
-        if not state.do_run_build_script and state.config.bootable:
-            state.root.joinpath("etc/kernel/cmdline").write_text(" ".join(state.config.kernel_command_line) + "\n")
-        return
-
-    with complete_step("Setting up basic OS tree…"):
-        state.root.mkdir(mode=0o755, exist_ok=True)
-        # We need an initialized machine ID for the build & boot logic to work
-        state.root.joinpath("etc").mkdir(mode=0o755, exist_ok=True)
-        state.root.joinpath("etc/machine-id").write_text(f"{state.machine_id}\n")
-
-        state.root.joinpath("etc/kernel").mkdir(mode=0o755, exist_ok=True)
-        state.root.joinpath("etc/kernel/cmdline").write_text(" ".join(state.config.kernel_command_line) + "\n")
-        state.root.joinpath("etc/kernel/entry-token").write_text(f"{state.machine_id}\n")
-        state.root.joinpath("etc/kernel/install.conf").write_text("layout=bls\n")
-
-
 def clean_paths(
         root: Path,
         globs: Sequence[str],
@@ -429,18 +408,9 @@ def reset_machine_id(state: MkosiState) -> None:
         return
 
     with complete_step("Resetting machine ID"):
-        if not state.config.machine_id:
-            machine_id = state.root / "etc/machine-id"
-            machine_id.unlink(missing_ok=True)
-            machine_id.write_text("uninitialized\n")
-
-        dbus_machine_id = state.root / "var/lib/dbus/machine-id"
-        try:
-            dbus_machine_id.unlink()
-        except FileNotFoundError:
-            pass
-        else:
-            dbus_machine_id.symlink_to("../../../etc/machine-id")
+        machine_id = state.root / "etc/machine-id"
+        machine_id.unlink(missing_ok=True)
+        machine_id.write_text("uninitialized\n")
 
 
 def reset_random_seed(root: Path) -> None:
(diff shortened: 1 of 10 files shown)


mkosi/__init__.py
```diff
@@ -429,18 +408,9 @@ def reset_machine_id(state: MkosiState) -> None:
...
-
-        dbus_machine_id = state.root / "var/lib/dbus/machine-id"
-        try:
-            dbus_machine_id.unlink()
-        except FileNotFoundError:
-            pass
-        else:
-            dbus_machine_id.symlink_to("../../../etc/machine-id")
+        machine_id = state.root / "etc/machine-id"
+        machine_id.unlink(missing_ok=True)
```
bluca, line 412: if the rootfs is read-only, this should create an empty file instead - so that it can be used as a mount point for the ephemeral (or command-line driven) machine-id

mkosi/__init__.py
```diff
@@ -249,27 +249,6 @@ def prepare_tree_root(state: MkosiState) -> None:
...
-    with complete_step("Setting up basic OS tree…"):
-        state.root.mkdir(mode=0o755, exist_ok=True)
-        # We need an initialized machine ID for the build & boot logic to work
-        state.root.joinpath("etc").mkdir(mode=0o755, exist_ok=True)
-        state.root.joinpath("etc/machine-id").write_text(f"{state.machine_id}\n")
-
-        state.root.joinpath("etc/kernel").mkdir(mode=0o755, exist_ok=True)
-        state.root.joinpath("etc/kernel/cmdline").write_text(" ".join(state.config.kernel_command_line) + "\n")
-        state.root.joinpath("etc/kernel/entry-token").write_text(f"{state.machine_id}\n")
-        state.root.joinpath("etc/kernel/install.conf").write_text("layout=bls\n")
```
behrmann, line 270: not that it matters terribly, since we don't guarantee that the images can be updated, but since we generate UKIs wouldn't the proper way to specify `layout=uki`. Probably moot, though, since with v254 this will probably be autodetected.

mkosi/__init__.py
```diff
@@ -610,7 +580,7 @@ def install_boot_loader(state: MkosiState) -> None:
         return
 
     with complete_step("Installing boot loader…"):
-        run(["bootctl", "install", "--root", state.root], env={"SYSTEMD_ESP_PATH": "/boot", **os.environ})
+        run(["bootctl", "install", "--root", state.root], env={"SYSTEMD_ESP_PATH": "/boot"})
```
behrmann, line 583: Probably also moot, since we don't guarantee upgradability, but this is where Debian loses that, since dpkg and its maintainer throw their hands into the air. I'm fine with continuing to have ESP on `/boot`, but BLS prefers `/efi`, so we could do that.

mkosi/backend.py
```diff
@@ -420,7 +418,9 @@ def __post_init__(self) -> None:
 
     @property
     def root(self) -> Path:
-        return self.workspace / "root"
+        p = self.workspace / "root"
+        p.mkdir(exist_ok=True, mode=0o755)
```
behrmann, line 422: I'm not entirely happy with a property having a side effect, just because that's not what one usually expects from attribute lookup. Maybe this could become a method or grow a new method `ensure_root_exists` or similar, that is called once somewhere at the beginning.

behrmann (commented):
LGTM, two thoughts and a nit.

mkosi/__init__.py
DaanDeMeyer: Ideally we could set the layout via the CLI of bootctl and kernel-install, but we can't. I want to write the least amount of files possible in the image itself, so I'd prefer to not configure the layout actually

mkosi/__init__.py
bluca: yeah this will break badly, if /boot is fat32 stuff will just start exploding left and right

DaanDeMeyer: I'm confused, what's the exact problem here? We install into /boot because we have to install somewhere, but afterwards this directory is just packed up into the EFI partition which systemd can mount wherever it wants when booting.

bluca: as long as this doesn't mean that /boot ends up as fat32 it's fine - we should have the /efi mountpoint in the image for read-only cases, is that there? any reason not to just install to /efi directly?

mkosi/__init__.py
behrmann: For `kernel-install` we could do it via environment variable (`KERNEL_INSTALL_LAYOUT=uki`), but yes, for `bootctl` nothing like this exists to my knowledge.

mkosi/__init__.py
DaanDeMeyer: > as long as this doesn't mean that /boot ends up as fat32 it's fine - we should have the /efi mountpoint in the image for read-only cases, is that there? any reason not to just install to /efi directly?

The dracut kernel-install script and other things hardcode /boot as the output directory.

So we just need to make sure that the EFI partition ends up being mounted at /efi in the booted image?

mkosi/__init__.py
DaanDeMeyer: I'm pretty sure that doesn't work, kernel-install only supports reading the layout from the config file

behrmann: Yep, looking through the code, you're right.

mkosi/__init__.py
johannbg: > The dracut kernel-install script and other things hardcode /boot as the output directory.
Just an late FYI 
The reason why dracut ( and presumably other things ) have /boot hardcoded is because GRUB 2 bootloader expects the initrd to be installed in /boot which may or may not affect your change here.

PR #1334, opened 2023-02-09
Fix bind mounting resolv.conf when it's a symlink

DaanDeMeyer:
bwrap doesn't support mounting over symlinks, so let's take that into account.


DaanDeMeyer:
Merging early to unblock systemd mkosi upgrade

PR #1338, opened 2023-02-09
Fix cache directory

DaanDeMeyer:
Make sure to use state.cache where applicable. Also get rid of setup_package_cache() since it comes down to a one liner


behrmann (approved):
Closes #1337

PR #1343, opened 2023-02-10
Move to `pyproject.toml` instead of `setup.py`

behrmann:


DaanDeMeyer:
So we have README as a symlink to README.md because of some setuptools thing, can you check if that's still needed with this approach?

Same thing for MANIFEST.in? Is there a way to include the LICENSE file without that?

behrmann:
Changed accordingly. `setup.cfg` seems to still be needed because the pyflakes people seem to be vehemently opposed to supporting `pyproject.toml`. Keeping the file around seems more sensible than any third-party hacks.

pyproject.toml
```diff
@@ -1,5 +1,40 @@
...
+    "types-cryptography",
+    "isort",
+    "pyflakes",
+]
+
+[project.scripts]
+mkosi = "mkosi.__main__:main"
+
+[tool.setuptools]
+packages = ["mkosi"]
```
behrmann, line 30: Haven't checked this yet, but this seems to be mostly needed because we don't keep the `mkosi` module in ` src/` directory, which seems to be what all the cool kids do nowadays.

DaanDeMeyer: Can we rename it to src?

pyproject.toml
```diff
@@ -1,5 +1,40 @@
...
+requires = ["setuptools", "setuptools-scm"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "mkosi"
+authors = [
+    {name = "mkosi contributors", email = "systemd-devel@lists.freedesktop.org"},
+]
+version = "14"
+description = "Build Bespoke IS Images"
```
DaanDeMeyer, line 11: ?

pyproject.toml
behrmann: I think that's more of a [move](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/), i.e. `mkosi/ -> src/mkosi/`. I don't think that would be a problem and I don't have strong preferences here.

behrmann:
Open question:
- Should we move to ` src`-layout, i.e. move `mkosi/ -> src/mkosi/`. This does make it impossible to import mkosi from the repository root.
- Check the minimum setuptools version for the build backend. Probably not a problem on Arch and Fedora, but something to check whether we have it on Debian and other stable distros. Probably not a problem for users, but something for packagers
- Without the dance `bin/mkosi` did, a user-installed mkosi will not have the right paths set up. Is this a problem beyond the `shell` and `boot` verbs or is it enough to throw some `sudo` in the command line there, conditioned on not already being root?

mkosi/__main__.py
```diff
@@ -25,6 +25,7 @@ def propagate_failed_return() -> Iterator[None]:
 
 @propagate_failed_return()
 def main() -> None:
+    sys.excepthook = excepthook
```
keszybz, line 28: Justification?

pyproject.toml
keszybz: What is the point of the move? This makes development a bit more inconvenient, so if there's a reason to do it, it should be provided.

keszybz (commented):
Wouldn't it be better to keep `setup.py` at least for a little while?

pyproject.toml
behrmann: I'm lukewarm on changing this. It looks like it's more and more the default for Python projects nowadays and tools seem to target this as the default. The reason seems to be because people actually want to avoid being able to import something by accident. I don't really see this as a problem and am fine with keeping this as is, I only wanted to note here why this looks the way it does and what could be done about it.

mkosi/__main__.py
behrmann: This PR removes the `bin/mkosi` script in favour of using the entrypoint
```
[project.scripts]
mkosi = "mkosi.__main__:main"
```
in `pyproject.toml`, because the rootless builds make the dance we did in `bin/mkosi` if not unnecessary at least less so (see the open questions). The same would be possible with `setup.py`.

The entrypoint only takes a single function, so I moved the setting of the excepthook here, since otherwise it would only be called if the module were invoked.

behrmann:
> Wouldn't it be better to keep setup.py at least for a little while?

`setup.py` can be kept around longer, but pip's documentation already warns that
> The interface documented here is retained currently solely for legacy purposes, until the migration to pyproject.toml-based builds can be completed.

Granted, I, too, think, that this is still somewhat far off, but if everything we use is is already supported in some way by pyproject.toml (previously editable installs were missing), it might be worth to rip the bandaid off and be done with it. If you see good reasons against doing so, that might be a good reason to keep a `setup.py` around for longer.

DaanDeMeyer:
With this change, how do I run mkosi from the repo directly without bin/mkosi?

behrmann:
Directly from the repo you can
```
sudo python3 -m mkosi --your commandline --arguments here
```
but if we want to start processes that need elevated privileges then it might make sense to go back into execing into them or wrap starting them via sudo (or some other means).

DaanDeMeyer:
And what if I'm located in another repo? Let's say I'm working on systemd and I want to run mkosi from its sources directly on the config of the systemd repo when I'm located in the systemd repo?

DaanDeMeyer:
Closing until this is updated

PR #1345, opened 2023-02-13
Allow RPM to return total file sizes larger than 4GB

Nirusu:
When an RPM has a total content size over 4GB, the --qf parameter needs to use LONGSIZE instead of SIZE to display the total package content file size, likely to keep compatibility with code expecting 32 bit friendly values.

Otherwise, if a package is larger than 4GB, RPM returns (none). Since this is later given as input to Python's int(), it will throw an exception due to (none) not being a number.

For more information, see:
https://rpm.org/devel_doc/large_files.html

The line where the exception is thrown is:
https://github.com/systemd/mkosi/blob/af87828065c8eb65e4c089f6d07a6e231a3dafc0/mkosi/manifest.py#L130

Not sure if it makes sense to also add:
```python
if size != "(none)":
  size = int(size)
else:
  size = 0
```
here to prevent further crashes, similar as the code for dpkg does. However, judging from the comment, the size field is optional for dpkg, whereas it seems to be mandatory for RPM? So not sure if does make sense to `0` this, if the 4GB case now seemed to be one of the few triggers to actually return `(none)` and thus catch a bug?

### DIFF
--- a/mkosi/manifest.py
+++ b/mkosi/manifest.py
@@ -106,7 +106,7 @@ class Manifest:
 
         c = run(
             ["rpm", f"--root={root}", f"--dbpath={dbpath}", "-qa", "--qf",
-             r"%{NEVRA}\t%{SOURCERPM}\t%{NAME}\t%{ARCH}\t%{SIZE}\t%{INSTALLTIME}\n"],
+             r"%{NEVRA}\t%{SOURCERPM}\t%{NAME}\t%{ARCH}\t%{LONGSIZE}\t%{INSTALLTIME}\n"],
             stdout=PIPE,
             text=True,
         )


behrmann:
LGTM, but I'm not an RPM expert. Is there any downside to using the 64bit version, i.e. is it for unfathomable reason nut supported somehwere or is the only problem that one needs to know to always use `LONGFOO` instead of `FOO`?

Nirusu:
I think the only downside is that any downstream / parsing code needs to be handle large numbers (aka: be able to handle 64 bit integers).

Here it's fine because it's Python 3 and it doesn't limit the size of integers. But let's say any code parsing the manifest should be able to handle 64 bit integers. Though I believe most JSON parsers do.

PR #1346, opened 2023-02-14
Relabel image when selinux policy is installed

DaanDeMeyer:
If an selinux policy is installed in the image, let's relabel it automatically.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3135,6 +3135,25 @@ def run_preset_all(state: MkosiState) -> None:
         run(["systemctl", "--root", state.root, "preset-all"])
 
 
+def run_selinux_relabel(state: MkosiState) -> None:
+    selinux = state.root / "etc/selinux/config"
+    if not selinux.exists():
+        return
+
+    policy = run(["sh", "-c", f". {selinux} && echo $SELINUXTYPE"], text=True, stdout=subprocess.PIPE).stdout.strip()
+    if not policy:
+        return
+
+    fc = Path('/etc/selinux') / policy / 'contexts/files/file_contexts'
+
+    # We want to be able to relabel the underlying APIVFS mountpoints, so mount root non-recursive to a
+    # temporary location so that the underlying mountpoints become visible.
+    cmd = f"mkdir /tmp/relabel && mount --bind / /tmp/relabel && exec setfiles -m -r /tmp/relabel -F {fc} /tmp/relabel || exit $?"
+
+    with complete_step(f"Relabeling files using {policy} policy"):
+        run_workspace_command(state, ["sh", "-c", cmd])
+
+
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -3135,6 +3135,25 @@ def run_preset_all(state: MkosiState) -> None:
         run(["systemctl", "--root", state.root, "preset-all"])
 
 
+def run_selinux_relabel(state: MkosiState) -> None:
+    selinux = state.root / "etc/selinux/config"
+    if not selinux.exists():
+        return
+
+    policy = run(["sh", "-c", f". {selinux} && echo $SELINUXTYPE"], text=True, stdout=subprocess.PIPE).stdout.strip()
```
behrmann, line 3143: Wouldn't that be a good use of `shell=True`?

mkosi/__init__.py
```diff
@@ -3135,6 +3135,25 @@ def run_preset_all(state: MkosiState) -> None:
...
+    if not policy:
+        return
+
+    fc = Path('/etc/selinux') / policy / 'contexts/files/file_contexts'
+
+    # We want to be able to relabel the underlying APIVFS mountpoints, so mount root non-recursive to a
+    # temporary location so that the underlying mountpoints become visible.
+    cmd = f"mkdir /tmp/relabel && mount --bind / /tmp/relabel && exec setfiles -m -r /tmp/relabel -F {fc} /tmp/relabel || exit $?"
```
behrmann, line 3151:
```suggestion
    cmd = f"set -e; mkdir /tmp/relabel; mount --bind / /tmp/relabel; exec setfiles -m -r /tmp/relabel -F {fc} /tmp/relabel"
```
maybe easier as a oneliner to reason about, without having to think about associativity. (Only tested with bash invoked as `/bin/sh`.)

behrmann (commented):
LGTM, two nits.

mkosi/__init__.py
DaanDeMeyer: Maybe but I wouldn't sweat it with just one usecase

mkosi/__init__.py
DaanDeMeyer: I prefer the above actually

DaanDeMeyer:
Note that this doesn't work on its own, we also need https://github.com/systemd/systemd/pull/26410

mkosi/__init__.py
behrmann: I just did another quick check and they are semantically different:
```
$ /bin/sh -c "set -e; foo() { return 19 ; }; echo foo; foo; echo bar"; echo $?
foo
19
$ /bin/sh -c "foo() { return 19 ; }; echo foo && foo && echo bar || exit $?"; echo $?
foo
0
```
So it does depend a bit on what you want.

PR #1347, opened 2023-02-14
Fix need_cache_trees()

DaanDeMeyer:


DaanDeMeyer:
Let's merge since these are trivial fixes/improvements

PR #1349, opened 2023-02-14
Centos fixes

DaanDeMeyer:


mkosi/backend.py
```diff
@@ -244,24 +244,24 @@ class MkosiConfig:
     """
 
     verb: Verb
-    cmdline: list[str]
+    cmdline: Sequence[str]
```
behrmann, line 247: This change is not a good idea, because it gives us less type safety instead of more.  The one thing is, that we actually do fill this with lists, which have a different variance than `Sequence`  and nothing can guarantee that they are not changed (i.e. even if the dataclass were frozen, it would be the list changing instead of the `MkosiConfig` instance). The other problem is that `str` is an instnace of `Sequence[str]`, so with this change bare strings would properly typecheck wherever we don't expect it.

DaanDeMeyer: Dropped

mkosi/__init__.py
```diff
@@ -770,21 +770,33 @@ def install_unified_kernel(state: MkosiState, roothash: Optional[str]) -> None:
                 boot_binary = state.root / prefix / f"EFI/Linux/{image_id}-{kver}{boot_count}.efi"
 
             if state.root.joinpath("etc/kernel/cmdline").exists():
-                boot_options = state.root.joinpath("etc/kernel/cmdline").read_text().strip()
+                cmdline = state.root.joinpath("etc/kernel/cmdline").read_text().strip()
```
behrmann, line 773: Maybe life would be easier this way:
```suggestion
                cmdline = [state.root.joinpath("etc/kernel/cmdline").read_bytes().strip()]
```
and then adding to the list, e.g.
```
cmdline += b' '.join(arg.encode() for arg in state.installer.kernel_command_line(state))
```
followed by
```
state.workspace.joinpath("cmdline").write_bytes(b" ".join(cmdline) + b"\x00")
```

Alternatively sans the handling the stuff as bytes (that's maybe just my innate mistrust of adding a null-byte to a string).

behrmann:
Rest looks sensible to me.

mkosi/__init__.py
DaanDeMeyer: Not sure that's more readable than the current version. Preferably we'd store everything as a list but we can't just split the contents from the cmdline files so we have to treat it as a string.

behrmann: I think we can have it both ways and not think about splitting in between. From the `etc/kernel/cmdline` you'd get something like
```
cmdline = ["foo bar baz"]
```
and then add multiple `frob` and at the end, so that
```
cmdline == ["foo bar baz", "frob1", "frob2"]
" ".join(cmdline) == "foo bar baz frob1 frob2"
```
This would work regardless of this being strings or bytes. (Reading bytes form `etc/kernel/cmdline` would have the tiny advantage of leaving opaque what's in there and being 100% sure about the handling of null-byte, but I'm ±0 on that myself.)

PR #1350, opened 2023-02-15
Add --initrd option

daandemeyer:


docs/initrd.md
```diff
@@ -0,0 +1,21 @@
+# Building a custom initrd and using it in a mkosi image
+
+To build an image with a mkosi-built initrd:
```
behrmann, line 3:
```suggestion
Building an image with a mkosi-built initrd is a two step process, because you will build two images - the initrd and your distribution image.
```

docs/initrd.md
```diff
@@ -0,0 +1,21 @@
...
+1. Build an initrd image. You can for example build a **compressed** `cpio`
+archive with mkosi for any distribution containing the packages `systemd` and `udev`:
```
behrmann, line 5: Maybe a bit more assertive
```suggestion
1. Build an initrd image using the `cpio` output format with the same target distribtution as you want to use for your distribution image. mkosi compresses the `cpio` output format by default.
```
I'm not sure the notes on udev and systemd are necessary, since we already pull them in everywhere, no?

docs/initrd.md
```diff
@@ -0,0 +1,21 @@
...
+```
+[Output]
+Format=cpio
+
+[Content]
+Packages=systemd
+         udev
+         kmod
+```
+2. Invoke `mkosi` passing the initrd image via the `--initrd` option:
```
behrmann, line 15:
```suggestion
2. Invoke `mkosi` passing the initrd image via the `--initrd` option or add the `Initrd=` option to your mkosi config when building your distribution image.
```

mkosi.md
```diff
@@ -543,6 +543,12 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
   in building or further container import stages.  This option strips
   SELinux context attributes from the resulting tar archive.
 
+`Initrd=`, `--initrd`
```
behrmann, line 546: Should this maybe be `InitRd`? Also, should this support a directory like `mkosi.initrd`?

behrmann (commented):
Looks great. Some thoughts about the docs, to hopefully make it a bit clearer to users how to use this.

docs/initrd.md
DaanDeMeyer: We don't pull in udev by default unless it's a bootable image

behrmann: Ah, brain fart on my side, I thought systemd would generally pull in udev, but of course, it doesn't. 

But then maybe we should just pull udev in automatically when building initrds, because I guess we'll only want to support initrds with udev and systemd as init?

mkosi.md
DaanDeMeyer: We already have CacheInitrd=, so let's use the same naming here

PR #1351, opened 2023-02-16
Updates for openSUSE

fbuihuu:


mkosi/distributions/opensuse.py
```diff
@@ -28,6 +28,43 @@ def initrd_path(kver: str) -> Path:
...
+    if caching is not None:
+        cmdline += ["-k" if caching else "-K"]
+
+    run(cmdline + [repo])
+
+
+def zypper_install(state, packages):
+    if not state.config.with_docs:
+        # zypper has no option for excluding the docs...
+        state.root.joinpath("etc/zypp/zypp.conf").write_text("rpm.install.excludedocs = yes\n")
```
DaanDeMeyer: Can we tell zypper to look for its configuration elsewhere? If so, I'd prefer if we could write it under `state.workspace` to avoid touching the image as little as possible.

mkosi/distributions/opensuse.py
```diff
@@ -32,46 +32,65 @@ def initrd_path(kver: str) -> Path:
...
+    # files from the host environment. If we want to use our custom versions for
+    # the rootfs, we're left with two ways to specify them, depending on whether
+    # we need to customize a setting defined in zypp.conf or in zypper.conf. If
+    # it's in zypp.conf then the environment variable 'ZYPP_CONF' must be used
+    # (!) otherwise a custom zypper.conf can be specified with the global option
+    # '--config'.
 
+    zypp_conf = state.root.joinpath("etc/zypp/zypp.conf")
+    zypp_conf_content  = "[main]\n"
+    zypp_conf_content += "solver.onlyRequires = true\n"
```
DaanDeMeyer, line 62: Python has multiline strings, the arch.py file has a good example

mkosi/distributions/opensuse.py
```diff
@@ -32,46 +32,65 @@ def initrd_path(kver: str) -> Path:
         return Path("boot") / f"initrd-{kver}"
 
 
-def zypper_addrepo(state, url, name, caching=False):
-    run(["zypper", "--root", state.root, "addrepo", "-c", "-k" if caching else "-K", url, name])
-
+def run_zypper(state, global_opts, verb, verb_opts, *args, with_apivfs=False):
```
DaanDeMeyer, line 41: We use `invoke_<package-manager>` everywhere else so this should be named similar

DaanDeMeyer: with_apivfs is not used? Can't we just always use `run_apivfs` to keep things simple?

mkosi/backend.py
```diff
@@ -183,14 +183,7 @@ def detect_distribution() -> tuple[Optional[Distribution], Optional[str]]:
 
 
 def is_rpm_distribution(d: Distribution) -> bool:
-    return d in (
```
DaanDeMeyer, line 186: This should probably just be renamed to `is_dnf_distribution()` since that's what it is used for

DaanDeMeyer (changes requested):
This needs typing annotations added everywhere as well

mkosi/distributions/opensuse.py
fbuihuu: I think in this case it's better to make the configuration of zypper used to create the rootfs persistent so subsequent calls to zypper behave consistently. It would be odd for example if the docs was excluded initially but suddenly installed when the image is updated.

DaanDeMeyer: Current policy for mkosi is to not configure the package manager in the image, but just use the distribution provided config there and leave it to the user to configure it in the image itself. We already do this for the other distros, so we should do the same for opensuse.

We tried to support updating images in the past but it turns out its a huge can of worms so we're moving more towards the "golden image" usecase where we don't expect images to be updated in the traditional way after we've produced. If users want an image to behave in that way, it's possible but they'll have to do the configuration required for that themselves.

mkosi/distributions/opensuse.py
fbuihuu: It is used, `git grep with_apivfs` should tell you where it's needed... or maybe I'm missing your point.

DaanDeMeyer: I meant inside of `run_zypper()`, it's not used inside the function unless github is showing me the wrong diff

fbuihuu: ouch, yeah you're right ;)

mkosi/distributions/opensuse.py
fbuihuu: Even if not used later, the conf file can be still used to document the way the image was setup though. 

BTW my patch isn't changing the location of the configuration file, it was done this way previously.

If you think that needs to be changed, we can do that later in a different PR.

fbuihuu:
@DaanDeMeyer thanks for your review. I pushed a new version that will hopefully address most of your points.

DaanDeMeyer:
You can use isort to fix the sorting

fbuihuu:
> You can use isort to fix the sorting

Done.

mkosi/distributions/opensuse.py
```diff
@@ -21,47 +21,131 @@ def filesystem(cls) -> str:
...
+
     @staticmethod
     def initrd_path(kver: str) -> Path:
         return Path("boot") / f"initrd-{kver}"
 
 
+def invoke_zypper(state: MkosiState,
+                  global_opts: set[str],
+                  verb: str,
+                  verb_opts: set[str],
```
behrmann, line 44:
```suggestion
                  global_opts: list[str],
                  verb: str,
                  verb_opts: list[str],
```

mkosi/distributions/opensuse.py
```diff
@@ -21,47 +21,131 @@ def filesystem(cls) -> str:
...
+
+def zypper_addrepo(state: "MkosiState", url: str, name: str, caching: bool = False) -> None:
+    invoke_zypper(state, [], "addrepo", ["-c", "-k" if caching else "-K"], url, name)
+
+
+def zypper_removerepo(state: "MkosiState", repo: str) -> None:
+    invoke_zypper(state, [], "removerepo", [], repo)
+
+
+def zypper_modifyrepo(state: "MkosiState", repo: str, caching: bool = None) -> None:
```
behrmann, line 94:
```suggestion
def zypper_modifyrepo(state: "MkosiState", repo: str, caching: bool = False) -> None:
```

DaanDeMeyer:
`mypy mkosi/` to debug the type errors

mkosi/distributions/opensuse.py
```diff
@@ -21,47 +21,131 @@ def filesystem(cls) -> str:
...
+    invoke_zypper(state, [], "modifyrepo", ["-k" if caching else "-K"], repo)
+
+
+def zypper_install(state: "MkosiState", packages: set[str]) -> None:
+    global_opts = [
+        "--cache-dir", state.cache,
```
behrmann, line 100:
```suggestion
        "--cache-dir", os.fspath(state.cache),
```
is the most minimal fix (needs `import os` at the top, too), although all the zypper stuff should probably not take `list[str]`, but `list[PathString]` where appropriate.

mkosi/distributions/opensuse.py
```diff
@@ -21,47 +21,131 @@ def filesystem(cls) -> str:
...
+
+    verb_opts = [
+        "-y",
+        "--download-in-advance",
+    ]
+
+    invoke_zypper(state, global_opts, "install", verb_opts, *packages, with_apivfs=True)
+
+
+def zypper_remove(state: "MkosiState", packages: set[str]) -> None:
```
behrmann, line 112:
```suggestion
def zypper_remove(state: "MkosiState", packages: list[str]) -> None:
```

mkosi/distributions/opensuse.py
```diff
@@ -21,47 +21,131 @@ def filesystem(cls) -> str:
...
+                  global_opts: set[str],
+                  verb: str,
+                  verb_opts: set[str],
+                  *args: Any,
+                  with_apivfs: bool = False) -> None:
+    zypp_conf = state.root.joinpath("etc/zypp/zypp.conf")
+
+    cmdline = ["zypper", "--root", state.root, *global_opts, verb, *verb_opts, *args]
+
+    env=dict(ZYPP_CONF=zypp_conf)
```
behrmann, line 51:
```suggestion
    cmdline = ["zypper", "--root", state.root, *global_opts, verb, *verb_opts, *args]
    env={"ZYPP_CONF": state.root.joinpath("etc/zypp/zypp.conf")}

```
since `zypp_conf` is not used anywhere else.

mkosi/distributions/opensuse.py
```diff
@@ -21,47 +21,131 @@ def filesystem(cls) -> str:
...
+    # For some reason zypper has no command line option to exclude the docs,
+    # this can only be configured via zypp.conf.
+    def yes_no(b: bool) -> str:
+        return "yes" if b else "no"
+
+    zypp_conf_content = f"""\
+    [main]
+    solver.onlyRequires = yes
+    rpm.install.excludedocs = {yes_no(state.config.with_docs)}
+    """
```
behrmann, line 81: If you don't want this indented you will need `textwrap,dedent`
```suggestion
    zypp_conf_content = dedent(f"""\
    [main]
    solver.onlyRequires = yes
    rpm.install.excludedocs = {"yes" if state.config.with_docs else "no"}
    """)
```
You can also directly inline this into the `zypp_conf.write_text` call and it's not necessary to define the `yes_no` function.

behrmann (commented):
Another two stylistic nits

mkosi/distributions/opensuse.py
fbuihuu: Nope, if caching is not specified we shouldn't try to change the caching setting for the repo. What you suggest would  disable caching if `caching` was not specified.

behrmann: This doesn't make sense. First of all, the typing as it stands is wrong. `None` is not a `bool`, so mypy will complain about this. You would need to type this as `Optional[bool]` if you want `None` to be a valid value. But using `False` here has no measurable difference, since the truthiness of `None` is false and using it in a bool context (as done below) will use it in this way, e.g.
```
>>> caching = None
>>> ["-k" if caching else "-K"]
['-K']
>>> caching = False
>>> ["-k" if caching else "-K"]
['-K']
```

bluca:
@fbuihuu any chance we can get some help for building mkosi extensions on OBS? I am stuck here: https://github.com/openSUSE/obs-build/pull/872

mkosi/distributions/opensuse.py
fbuihuu: My initial intention was to anticipate the fact that `zypper_modifyrepo()` should be able to handle multiple repository properties, not just one as it's the case now with `caching`.

Hence if for example the helper has support for setting/unsetting properties `A` and `B`, calling `zypper_modifyrepo(state, repo, B=False)` should never try to change `A` property. To make this possible we need to make distinction between `A` set to a value and `A` not passed at all.

But given the fact that only one property is supported at the moment, making `caching` optional doesn't really make sense. So for now I changed the code to make `caching` parameter mandatory.

fbuihuu:
@behrmann @DaanDeMeyer thanks a lot for the feedback and tips, new version force pushed.

fbuihuu:
> @fbuihuu any chance we can get some help for building mkosi extensions on OBS? I am stuck here: [openSUSE/obs-build#872](https://github.com/openSUSE/obs-build/pull/872)

I'll try to find someone who can help.

mkosi/distributions/opensuse.py
behrmann: Ok, but the code for this would need to look differently, because the above 
```python
["-k" if caching else "-K"]
```
will always evaluate to either `-k` or `-K` whatever is passed in (as long as it support the `__bool__` protocol). For something agnostic the code would need to look something like this
```python
def zypper_modifyrepo(state: "MkosiState", repo: str, caching: Optiona[bool] = None) -> None:
    zypper_args = []
    if caching is not None:
        zypper_args = ["-k" if caching else "-K"]
    invoke_zypper(state, [], "modifyrepo", *zypper_args, repo)
```

mkosi/distributions/opensuse.py
```diff
@@ -21,47 +23,128 @@ def filesystem(cls) -> str:
...
+    def yes_no(b: bool) -> str:
+        return "yes" if b else "no"
+
+    zypp_conf.write_text(
+        dedent(
+             f"""\
+             [main]
+             solver.onlyRequires = yes
+             rpm.install.excludedocs = {yes_no(not state.config.with_docs)}
+             """
```
behrmann, line 83:
```suggestion
    zypp_conf.write_text(
        dedent(
             f"""\
             [main]
             solver.onlyRequires = yes
             rpm.install.excludedocs = {"no" if state.config.with_docs else "yes"}
```
can we inline this to avoid the double negative?

mkosi/distributions/opensuse.py
fbuihuu: Correct, that was my initial implementation but somehow screwed it up  by the commit that introduced `invoke_zypper()`.

But as I said, I finally make `caching` parameter mandatory for now.

mkosi/distributions/opensuse.py
```diff
@@ -21,47 +23,128 @@ def filesystem(cls) -> str:
 
     @classmethod
     def install(cls, state: "MkosiState") -> None:
```
behrmann, line 25:
```suggestion
    def install(cls, state: MkosiState) -> None:
```
This is actually a copy paste error introduced earlier, so the string quoting `"MkosiState"` is unnecessary in the rest.

mkosi/distributions/opensuse.py
```diff
@@ -21,47 +23,128 @@ def filesystem(cls) -> str:
...
     @staticmethod
     def initrd_path(kver: str) -> Path:
         return Path("boot") / f"initrd-{kver}"
 
 
+def invoke_zypper(state: MkosiState,
+                  global_opts: list[str],
+                  verb: str,
+                  verb_opts: list[str],
+                  *args: Any,
```
behrmann, line 47: They can't be `Any`, no? Since these will be a command line they can only be `str` or `Path`, i.e. `PathString`. This also needs a corresponding annotation below
```python
cmdline: list[PathString] = ["zypper", "--root", state.root, *global_opts, verb, *verb_opts, *args]
```

mkosi/distributions/opensuse.py
```diff
@@ -21,47 +23,128 @@ def filesystem(cls) -> str:
...
+             [main]
+             solver.onlyRequires = yes
+             rpm.install.excludedocs = {yes_no(not state.config.with_docs)}
+             """
+        )
+    )
+
+
+def zypper_addrepo(state: "MkosiState", url: str, name: str, caching: bool = False) -> None:
+    invoke_zypper(state, [], "addrepo", ["-c", "-k" if caching else "-K"], url, name)
```
behrmann, line 89: nit: The long options would be nice (here and elsewhere), because it's easier to follow when not using zypper.

bluca:
> > @fbuihuu any chance we can get some help for building mkosi extensions on OBS? I am stuck here: [openSUSE/obs-build#872](https://github.com/openSUSE/obs-build/pull/872)
> 
> I'll try to find someone who can help.

Thank you! I'm also always hanging out in the obs IRC channel, in case that helps and something needs to be discussed

mkosi/distributions/opensuse.py
```diff
@@ -20,48 +22,126 @@ def filesystem(cls) -> str:
...
+        if state.config.base_image:
+            # We assume that the base image has been properly initialized and it
+            # contains all the metadata we need to install the additional
+            # packages.
+            return zypper_install(state, {*state.config.packages})
+
         return install_opensuse(state)
 
+    @classmethod
+    def remove_packages(cls, state: "MkosiState", packages: list[str]) -> None:
```
behrmann, line 35:
```suggestion
    def remove_packages(cls, state: MkosiState, packages: list[str]) -> None:
```
Missed on.

behrmann (approved):
LGTM, one remaining nit.

DaanDeMeyer (approved):

bluca (approved):

PR #1355, opened 2023-02-21
Always run subprocesses with constrained environment

DaanDeMeyer:
Also set LANG=C.UTF-8 if not set explicitly.

### DIFF
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -193,7 +193,7 @@ def run(
     check: bool = True,
     stdout: _FILE = None,
     stderr: _FILE = None,
-    env: Optional[Mapping[str, Any]] = None,
+    env: Mapping[str, Any] = {},
     **kwargs: Any,
 ) -> CompletedProcess:
     cmdline = [os.fspath(x) for x in cmdline]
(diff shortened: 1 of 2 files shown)


mkosi/run.py
```diff
@@ -207,13 +207,12 @@ def run(
...
-    else:
-        env = dict(
-            PATH=os.environ["PATH"],
-            TERM=os.getenv("TERM", "vt220"),
-        ) | env
+    env = dict(
+        PATH=os.environ["PATH"],
+        TERM=os.getenv("TERM", "vt220"),
+        LANG=os.getenv("LANG", "C.UTF-8"),
+        *env,
```
behrmann, line 214:
```suggestion
        **env,
```

DaanDeMeyer:
Updated to always use `C.UTF-8` since that's the only we can assume to always be available

mkosi/run.py
```diff
@@ -193,7 +193,7 @@ def run(
     check: bool = True,
     stdout: _FILE = None,
     stderr: _FILE = None,
-    env: Optional[Mapping[str, Any]] = None,
+    env: Mapping[str, Any] = {},
```
behrmann, line 196: This should probably not be changed, because that's a footgun. Function arguments are only ever initialised once, so subsequent calls to this function might get previous invocations' `env`. I think the problem is sidestepped in this case, because `env` is overwritten wholesale and the initial `env` not changed, but this is too brittle.

Also, since we pass `env` to a subprocess, can't we whittle this down to `Mapping[str, str]`?

DaanDeMeyer: We use a read-only type so this isn't a problem

PR #1356, opened 2023-02-21
Tighten typing of env variable passed to run functions

behrmann:


mkosi/distributions/debian.py
```diff
@@ -292,12 +292,12 @@ def invoke_apt(
...
-        DEBIAN_FRONTEND="noninteractive",
-        DEBCONF_INTERACTIVE_SEEN="true",
-        INITRD="No",
-    )
+    env: dict[str, PathString] = {
+        "APT_CONFIG": config_file,
+        "DEBIAN_FRONTEND": "noninteractive",
+        "DEBCONF_INTERACTIVE_SEEN": "true",
+        "INITRD": "No",
+    }
```
DaanDeMeyer, line 300: Why chance this? I prefer using `dict()` over `{}`

behrmann: I can leave the `dict` call if you want, but generally using the literal for initialisation is preferable, since it allows for the little compile time optimisation that Python does. Using the literal it can become a dict (or e.g. in the case of sets a set or a frozen set) at compile time, whereas the dict call will only happen at runtime (which also means that there is a call happening). This is because Python will have to resolve the constructor, since it might have been overridden.

DaanDeMeyer: There's a lot more pressing stuff to fix if we care about performance, let's leave the dict() as it is

DaanDeMeyer (approved):

mkosi/distributions/debian.py
behrmann: Sure, no problem. I do find literals much easier on the eyes, than functions calls, but let's leave that for another day. :)

PR #1359, opened 2023-02-22
Gather all boot loader related logic in install_boot_loader()

DaanDeMeyer:
We drop the logic to sign unified kernel images as ukify takes care of that for us.


behrmann:
The idea of letting ukify do its thing is sensible, but I'm not sure we should cram it all into `install_bootloader` (I find the smaller functions easier to reason about), apart from that LGTM.

PR #1363, opened 2023-02-24
Drop unused chdir() context manager

DaanDeMeyer:


DaanDeMeyer:
Added some extra stuff and all the trivial stuff from #1298 as well

mkosi/__init__.py
```diff
@@ -726,35 +715,37 @@ def make_tar(state: MkosiState) -> None:
...
 
-        with spawn(cmd, stdin=subprocess.PIPE, stdout=f) as cpio:
+        with spawn(cmd, stdin=subprocess.PIPE, text=True) as cpio:
             #  https://github.com/python/mypy/issues/10583
             assert cpio.stdin is not None
 
             for file in files:
-                cpio.stdin.write(os.fspath(file).encode("utf8") + b"\0")
+                cpio.stdin.write(os.fspath(file))
+                cpio.stdin.write("\0")
```
behrmann, line 747: This change seems not necessary. Why decode and reencode the bytes when we're just passing them through?

mkosi/__init__.py
```diff
@@ -821,12 +802,12 @@ def install_unified_kernel(state: MkosiState, roothash: Optional[str]) -> None:
                 boot_count = f'+{state.root.joinpath("etc/kernel/tries").read_text().strip()}'
 
             if state.config.image_version:
-                boot_binary = state.root / prefix / f"EFI/Linux/{image_id}_{state.config.image_version}{boot_count}.efi"
+                boot_binary = state.root / f"boot/EFI/Linux/{image_id}_{state.config.image_version}{boot_count}.efi"
```
behrmann, line 805: Should be maybe go for `efi` as the prefix here? It shouldn't change anything in the UKI case and people who want to do an XBOOTLDR partition would have a mountpoint for it.

behrmann (commented):
Overall LGTM, one nit and one thought.

mkosi/__init__.py
DaanDeMeyer: Not until we drop dracut as it writes to /boot. Maybe in a separate PR

mkosi/__init__.py
behrmann: Ah, right... Forgot about that.

behrmann: That comment was actually meant for the other thread.

behrmann (approved):

PR #1365, opened 2023-02-26
gentoo: reenable CI tests

257:
some cosmetic linewraps


257:
@DaanDeMeyer dunno why debian fails, gentoo is fine though.

mkosi/distributions/gentoo.py
```diff
@@ -164,13 +164,15 @@ def __init__(self, state: MkosiState) -> None:
...
+            self.emerge_default_opts += ["--verbose", "--quiet=n",
+                                         "--quiet-fail=n"]
         else:
             self.emerge_default_opts += ["--quiet-build", "--quiet"]
 
         self.arch, _ = ARCHITECTURES[state.config.architecture or "x86_64"]
         self.arch_profile = Path(f"default/linux/{self.arch}/"
-                                 f"{state.config.release}/no-multilib/systemd/merged-usr")
+                                 f"{state.config.release}/"
+                                 "no-multilib/systemd/merged-usr")
```
behrmann, line 175: Any need to change these?

mkosi/distributions/gentoo.py
```diff
@@ -301,7 +286,9 @@ def mkosi_conf(self) -> None:
         # we use this so we don't need to touch upstream files.
         # we also use this for documenting build environment.
         emerge_vars_str = ""
-        emerge_vars_str += "\n".join(f'{k}="${{{k}}} {v}"' for k, v in self.emerge_vars.items())
+        emerge_vars_str += "\n".join(
+            f'{k}="${{{k}}} {v}"' for k, v in self.emerge_vars.items()
+        )
```
behrmann, line 291: This is also just a formatting change.

behrmann (commented):
Looks reasonable. Nit: some seemingly unnecessary formatting changes.

mkosi/distributions/gentoo.py
257: > This is also just a formatting change.

E501 line too long (size > 79 characters)

mkosi/distributions/gentoo.py
257: > Any need to change these?

E501 line too long (size > 79 characters)

behrmann: What did you run to get that warning and how did you run it? We set 
```
[flake8]
max-line-length = 119
```
in `setup.cfg`

mkosi/backend.py
```diff
@@ -492,15 +492,18 @@ def safe_tar_extract(tar: tarfile.TarFile, path: Path=Path("."), *, numeric_owne
...
     """
     path = path.resolve()
+    members = []
     for member in tar.getmembers():
         target = path / member.name
         try:
             # a.relative_to(b) throws a ValueError if a is not a subpath of b
-            target.resolve().relative_to(path)
+            if not (member.ischr() or member.isblk()):
+                target.resolve().relative_to(path)
```
DaanDeMeyer, line 501: Please move the accompanying comment down here as well

mkosi/backend.py
```diff
@@ -492,15 +492,18 @@ def safe_tar_extract(tar: tarfile.TarFile, path: Path=Path("."), *, numeric_owne
     See https://github.com/advisories/GHSA-gw9q-c7gh-j9vm
     """
     path = path.resolve()
+    members = []
     for member in tar.getmembers():
         target = path / member.name
         try:
             # a.relative_to(b) throws a ValueError if a is not a subpath of b
-            target.resolve().relative_to(path)
+            if not (member.ischr() or member.isblk()):
```
DaanDeMeyer, line 500: Please document this change and put it in a separate commit

mkosi/distributions/gentoo.py
DaanDeMeyer: No but it makes much more sense for this to be one line instead of split over multiple lines so let's keep it that way. Maybe figure out how to configure the line length for pyright/pylsp in nvim?

mkosi/distributions/gentoo.py
```diff
@@ -224,8 +222,9 @@ def fetch_fix_stage3(self) -> None:
             self.portage_cfg["GENTOO_MIRRORS"],
             f"releases/{self.arch}/autobuilds/{stage3_tar}",
         )
-        stage3_tar_path = self.portage_cfg["DISTDIR"] / stage3_tar
-        stage3_tmp_extract = stage3_tar_path.with_name(stage3_tar.name + ".tmp")
+        stage3_tar_path = self.state.cache / stage3_tar
+        stage3_tmp_extract = stage3_tar_path.with_name(stage3_tar.name +
```
DaanDeMeyer, line 226: Ditto, please keep this  a single line

mkosi/distributions/gentoo.py
```diff
@@ -238,7 +237,8 @@ def fetch_fix_stage3(self) -> None:
                 with tarfile.open(stage3_tar_path) as tfd:
                     MkosiPrinter.print_step(f"Extracting {stage3_tar.name} to "
                                             f"{stage3_tmp_extract}")
-                    safe_tar_extract(tfd, stage3_tmp_extract, numeric_owner=True)
+                    safe_tar_extract(tfd, stage3_tmp_extract,
```
DaanDeMeyer, line 240: Ditto, single line

mkosi/distributions/gentoo.py
DaanDeMeyer: Ditto, keep this a single line

mkosi/distributions/gentoo.py
```diff
@@ -313,8 +300,13 @@ def mkosi_conf(self) -> None:
             )
         )
 
+    def get_snapshot_of_portage_tree(self) -> None:
+        cmd = ["/usr/bin/emerge-webrsync"]
```
DaanDeMeyer, line 304: No need for a separate variable here

mkosi/distributions/gentoo.py
```diff
@@ -313,8 +300,13 @@ def mkosi_conf(self) -> None:
...
+        run_workspace_command(self.state, cmd, network=True)
+
     def update_stage3(self) -> None:
-        self.invoke_emerge(opts=self.EMERGE_UPDATE_OPTS, pkgs=self.pkgs['boot'])
+        self.invoke_emerge(opts=self.EMERGE_UPDATE_OPTS,
```
DaanDeMeyer, line 308: Single line

mkosi/distributions/gentoo.py
```diff
@@ -329,50 +321,19 @@ def merge_user_pkgs(self) -> None:
...
-                    f"--root={self.root.resolve()}",
-                    f"--sysroot={self.root.resolve()}",
-                ]
-
-            MkosiPrinter.print_step(f"Invoking emerge(1) pkgs={pkgs} "
-                                    f"actions={actions} outside stage3")
-            emerge_main([*pkgs, *opts, *actions] + PREFIX_OPTS + self.emerge_default_opts)
-        else:
-            cmd = ["/usr/bin/emerge", *pkgs, *self.emerge_default_opts, *opts, *actions]
+        cmd = ["/usr/bin/emerge", *pkgs, *self.emerge_default_opts, *opts,
```
DaanDeMeyer, line 330: For all other distros, we run the package manager on the host outside of the image, not inside it. I would very much want us to do the same for gentoo. If I read the emerge man page, it seems to support the necessary options (--root and such) so can you modify the gentoo implementation to use emerge installed on the host instead of invoking it inside the image?

DaanDeMeyer (changes requested):

mkosi/backend.py
257: this is the result of dropping root privileges. should it be included in the commit?

mkosi/distributions/gentoo.py
257: that would be an almost complete re-write. lets have this merged for now and i will start working on what you're suggesting.

257: > what you're suggesting.
although i'm not sure why that's a better way of doing it. i was actually going to remove `portage` requirement on the host and have the whole thing self-contained.

mkosi/backend.py
DaanDeMeyer: This is a self-contained change on its own, where we don't unpack character or device nodes from an archive. So it makes sense on its own and should be a separate commit.

mkosi/distributions/gentoo.py
DaanDeMeyer: We want to be able to build images without the tools that were used to build them included. Especially for package managers with heavy dependencies, such as emerge and dnf, which pull in python, we want to be able to use those to build images that don't include dnf or emerge so we can keep those images small. While we could try to uninstall the package managers after building the images, it'll be a constant battle to ensure we remove all files associated with the package manager. The other solution is to never install the package manager in the image in the first place, which we can do by running it on the host instead of inside the image.

mkosi/distributions/gentoo.py
```diff
@@ -164,13 +161,13 @@ def __init__(self, state: MkosiState) -> None:
             "--nospinner",
         ]
         if "build-script" in ARG_DEBUG:
-            self.emerge_default_opts += ["--verbose", "--quiet=n", "--quiet-fail=n"]
+            self.emerge_default_opts += ["--verbose", "--quiet=n",
```
DaanDeMeyer, line 164: Please make this a single line again

mkosi/distributions/gentoo.py
```diff
@@ -329,50 +309,24 @@ def merge_user_pkgs(self) -> None:
         if self.state.config.packages:
             self.invoke_emerge(pkgs=self.state.config.packages)
 
+    def cleanup(self) -> None:
```
DaanDeMeyer, line 312: Please do this in clean_package_manager_metadata() instead of here

mkosi/distributions/gentoo.py
```diff
@@ -329,50 +309,24 @@ def merge_user_pkgs(self) -> None:
...
-            MkosiPrinter.print_step(f"Invoking emerge(1) pkgs={pkgs} "
-                                    f"actions={actions} outside stage3")
-            emerge_main([*pkgs, *opts, *actions] + PREFIX_OPTS + self.emerge_default_opts)
-        else:
-            cmd = ["/usr/bin/emerge", *pkgs, *self.emerge_default_opts, *opts, *actions]
-
-            MkosiPrinter.print_step("Invoking emerge(1) inside stage3"
-                                    f"{self.root}")
+        cmd = ["/usr/bin/emerge", *pkgs, *self.emerge_default_opts, *opts,
+               *actions]
```
DaanDeMeyer, line 324: This can be a single line as well

DaanDeMeyer (changes requested):

mkosi/distributions/gentoo.py
257: regarding the image size, i am not happy with the result either, and there are lots that can be done there. ATM many, but not all, of those options can be exercised using `mkosi.skeleton/etc/portage/*`. we could expose those but that would possibly require distro-specific flags. i wonder what you guys think?

as for not including the package manager, not an option ATM i'm afraid. upstream ships `portage` unconditionally with their stage3 and i guess sticking with [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) we want to adhere to that exercise, i would imagine. any thoughts?

there is also an argument to be made from user's point of view. does the user want to have the package manager at its disposal or not. there are many use cases, including my owns, where the answer is Yes. i can think of few cases where that answer is No... it's trade off, as always, so i guess it's better to allow the use to choose.

removing `portage` incidentally doesn't buy us much, low hanging fruits are somewhere else. build dependencies is one, which can be easily done with tweaking the `portage`, tool-chain is another one but do we really want that?
again all of that, IMHO, should be seen from user's point of view; i can't think of many cases where a developer (the main target-user of `mkosi`) wants a final image on the first run, for me at least it takes couple of try to arrive at what i would consider a *golden-master* and even then, at least with Gentoo, things move quickly enough that sometime i even need to start from scratch.

mkosi/distributions/gentoo.py
257: not the same thing, image includes `portage`, those are simply sources, binary pkgs, etc. should i remove it for now? we can revisit this when we deal with image size.

DaanDeMeyer: Sounds good, let's remove this for now

mkosi/distributions/gentoo.py
DaanDeMeyer: The thing is, we want to make mkosi a proper tool for building "golden" non-development images as well, as this is one of the main target use cases aside from the development use case. And to build those images, we should build minimal images by default, without build dependencies, without toolchains, and without heavyweight package managers. Users that want those to be available can then simply add them to the list of packages to be installed.

257: thinking more about this more i have some thoughts to share.

i don't see a way to use host system's `portage` for building images at all and here is why:

allowing users to install their desired packages is an essential part of `mkosi`, i understand. also upstream `toolchain`, the ones shipped within `stage3` root-tree, is a hard dependency for building a consistent os image in case of Gentoo, being a source-based distro.
 
as for image size and what we can shed off:
inclusion of `toolchains` is an upstream-thing, i'm not aware of a smaller system root-tree upstream than `stage3` in fact. note that there is no desktop related stuff (kde/gnome what have you) included in `stage3` so it is, to the best of my knowledge, _the_ smallest we can get from upstream.
yes, there are `stage1` and `stage2` but those are beyond the scope of `mkosi`.

so considering those getting into the business of removing the `toolchains` wouldn't make sense IMHO.
take what we did before `usrmerge` as an example. we used to go to great length to modify the upstream root-tree to make sure `usronly` et al work. that added lots of logic that could be avoided if upstream already provided a root-tree contained under `/usr` which they do now and that reduces code and complexity at our end.

if, one day, upstream provides a root-tree without `toolchains`, considering that they do have a semi-official binary repo now, we can look into using that root-tree.

257: ... also note that users can ask portage to not to installed build dependencies of their packages by using `mkosi.skeleton/etc/portage` and/or `mkosi.extra/etc/poratge`. we could document this i guess.

DaanDeMeyer: I get your point about the toolchains. How is that related to portage though? Can't we build images using portage from the host while still using the toolchains from the image? I understand that probably wouldn't meaningfully impact the image size, but I would still like it that way for consistency with the other distributions.

257: sounds like [cross-compiling](https://wiki.gentoo.org/wiki/Cross_build_environment), possible i guess but complicated.

mkosi/distributions/gentoo.py
```diff
@@ -329,50 +307,18 @@ def merge_user_pkgs(self) -> None:
...
-
-            MkosiPrinter.print_step(f"Invoking emerge(1) pkgs={pkgs} "
-                                    f"actions={actions} outside stage3")
-            emerge_main([*pkgs, *opts, *actions] + PREFIX_OPTS + self.emerge_default_opts)
-        else:
-            cmd = ["/usr/bin/emerge", *pkgs, *self.emerge_default_opts, *opts, *actions]
-
-            MkosiPrinter.print_step("Invoking emerge(1) inside stage3"
-                                    f"{self.root}")
+        cmd = ["/usr/bin/emerge", *pkgs, *self.emerge_default_opts, *opts, *actions]
```
DaanDeMeyer, line 316: Just use "emerge" so it gets looked up in PATH

mkosi/distributions/gentoo.py
```diff
@@ -329,50 +307,18 @@ def merge_user_pkgs(self) -> None:
...
+    def _dbg(self) -> None:
         """this is for dropping into shell to see what's wrong"""
-
-        bwrap = [
-            "--bind", self.portage_cfg['PORTDIR'], self.portage_cfg['PORTDIR'],
-            "--bind", self.portage_cfg['DISTDIR'], self.portage_cfg['DISTDIR'],
-            "--bind", self.portage_cfg['PKGDIR'],  self.portage_cfg['PKGDIR'],
-        ]
-        run_workspace_command(self.state, ["sh"], network=True, bwrap_params=bwrap)
+        run_workspace_command(self.state, ["sh"], network=True)
```
DaanDeMeyer, line 321: Can we drop this? We already have --debug run and if needed you can always just add this locally since its a one liner

DaanDeMeyer (changes requested):
Looks good, just two more small changes

mkosi/distributions/gentoo.py
257: done

257:
i noticed CI is failing, let me look into that.

DaanDeMeyer (approved):

DaanDeMeyer:
Centos CI is expected to fail

DaanDeMeyer:
I saw that you disabled the prebuilt initrd test for gentoo. This is fine for now, but the plan is to drop dracut support in the (maybe near) future in favor of prebuilt initrds built by mkosi. I'm guessing that the current gentoo images as generated by mkosi are wayyy too big to be used as initrds. So ideally you could find some way to get minimal gentoo images working so we can use those as initrds instead of using dracut.

257:
no, it was failing because there is no `udev` package on `systemd` profiles.
what is the average image size, say on fedora?

DaanDeMeyer:
> no, it was failing because there is no udev package on systemd profiles.

Any chance we could get an alias or something? It's rather convenient that `systemd`, `kmod` and `udev` exist on all popular distros.

> what is the average image size, say on fedora?

An uncompressed Fedora initrd generated by mkosi with systemd and udev is 114M

PR #1366, opened 2023-02-27
debian: create /efi mountpoint

bluca:
Ensure ESP is mounted on /efi rather than /boot so that dpkg doesn't attempt to install files on vFAT

### DIFF
--- a/mkosi/distributions/debian.py
+++ b/mkosi/distributions/debian.py
@@ -156,6 +156,8 @@ class DebianInstaller(DistributionInstaller):
         invoke_apt(state, "get", "update", ["--assume-yes"])
 
         if state.config.bootable and not state.do_run_build_script:
+            # Ensure /efi exists so that the ESP is mounted there, and we never run dpkg -i on vfat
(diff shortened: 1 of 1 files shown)


behrmann (approved):

DaanDeMeyer:
I'm not sure if this is sufficient on its own. We only pack up the /boot folder into the ESP, not the /efi folder. What we can do is create /efi unconditionally, and then pack up /efi instead of /boot. We'll still need to look for initrds in /boot though, because dracut puts them there.

bluca:
> I'm not sure if this is sufficient on its own. We only pack up the /boot folder into the ESP, not the /efi folder. What we can do is create /efi unconditionally, and then pack up /efi instead of /boot. We'll still need to look for initrds in /boot though, because dracut puts them there.

Seems to work fine for me? /efi is empty, nothing is installed there, as the bootctl et al calls are configured here to install in /boot. But that's fine, it's the path used at build time. If /efi is available, the gpt auto gen will prefer it.

bluca:
unlikely, as grub is still the default

PR #1369, opened 2023-03-03
Path fixes

DaanDeMeyer:


mkosi/run.py
```diff
@@ -319,9 +321,11 @@ def run_workspace_command(
         container="mkosi",
         SYSTEMD_OFFLINE=str(int(network)),
         HOME="/",
+        # Make sure the default PATH of the distro shell is used.
+        PATH="",
     ) | env | state.environment
 
-    template = "chmod 1777 /tmp /var/tmp /dev/shm && exec {} || exit $?"
+    template = "[ -f /etc/profile ] && source /etc/profile; chmod 1777 /tmp /var/tmp /dev/shm && exec {} || exit $?"
```
behrmann, line 328: Could this have weird edge cases? At least bash when invoked as non-interactive, non-login `sh` will not read any startup files. I see the appeal of trying to use the default shell the distro sets, but maybe for this just setting a sensible default ourselves, that includes `sbin`, is enough.

DaanDeMeyer: Agreed, changed approach so we just append /usr./bin and /usr/sbin to PATH as fallback values

PR #1370, opened 2023-03-06
Packages related refactors

DaanDeMeyer:
- Use Sequence as the type for passing package lists around
- Add install_packages() method to DistributionInstaller
- Remove install_packages_rpm() and use invoke_dnf() directly


DaanDeMeyer:
This was mostly already reviewed in #1358 so let's merge.

PR #1376, opened 2023-03-07
centos: Install systemd-boot again on CentOS Stream 9

DaanDeMeyer:
systemd-boot was removed from the systemd package again so let's start installing it explicitly again on CentOS Stream 9.


DaanDeMeyer:
This should break horribly until systemd-extras in EPEL 9 is updated to systemd v252

behrmann (approved):

PR #1383, opened 2023-03-12
fedora: Don't unconditionally remove /usr/share/locale

DaanDeMeyer:


DaanDeMeyer:
Added one more commit to expand user paths even when not running as root

behrmann (approved):

PR #1385, opened 2023-03-12
Extend credentials support

DaanDeMeyer:
We now also read credentials from mkosi.credentials/. If a file in mkosi.credentials/ is executable, we run the file and use its output as the credential.


mkosi.md
```diff
@@ -1220,6 +1225,13 @@ local directory:
   used as the repository directory for extra repository files. See
   the `RepositoryDirectories` option for more information.
 
+* The **`mkosi.credentials/`** directory, if it exists, is used as a
+  source of extra credentials as configured with `--credential`. For
+  each file in the directory, the filename becomes the credential
+  name and its contents become the credential value unless the file is
+  executable, in which case mkosi will execute the file and the command
+  its stdout is used as the credential's value.
```
behrmann, line 1233:
```suggestion
* The **`mkosi.credentials/`** directory is used as a
  source of extra credentials similar to the `Credentials=` option. For
  each file in the directory, the filename will be used as the credential
  name and the file contents become the credential value, or, if the file is
  executable, mkosi will execute the file and the command's
  output to stdout will be used as the credential value. Output to stderr will be ignored. 
  Credentials configured with `Credentials=` take precedence over files in `mkosi.credentials`.
```
Dropped the "if it exists" since it's a bit superfluous. Changed the option to the config file option, since I think's that more in line with the rest. Shortened the rest and added a sentence about order and stderr.

behrmann (approved):

PR #1389, opened 2023-03-14
Fix documentation for WorkspaceDirectory

### LINKED ISSUE #1388: Behavior of WorkspaceDirectory is not as documented
The documentation for `WorkspaceDirectory` says

>WorkspaceDirectory=, --workspace-dir=
>
>: Path to a directory where to store data required temporarily while building the image. This directory should have enough space >to store the full OS image, though in most modes the actually used disk space is smaller. If not specified, and mkosi.workspace/ >exists in the local directory, it is used for this purpose. Otherwise, a subdirectory in the output directory is used.

If I read that correctly, if I don't set this option, then a subdirectory of `OutputDirectory` should be used (if set). However, [in the code](https://github.com/systemd/mkosi/blob/main/mkosi/__init__.py#L3298#L3299) this behavior is not reflected and the current directory is used instead.

Cornelicorn:
Closes #1388

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -292,8 +292,8 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
   building the image. This directory should have enough space to store
   the full OS image, though in most modes the actually used disk space
   is smaller. If not specified, and `mkosi.workspace/` exists in the
-  local directory, it is used for this purpose. Otherwise, a
-  subdirectory in the output directory is used.
+  local directory, it is used for this purpose. Otherwise, hidden
+  subdirectories of the current working directory are used.
 
 : The data in this directory is removed automatically after each
   build. It's safe to manually remove the contents of this directory


mkosi/__init__.py
```diff
@@ -2854,12 +2854,12 @@ def print_summary(config: MkosiConfig) -> None:
     print("                    Netdev:", yes_no(config.netdev))
 
 
-def make_output_dir(state: MkosiState) -> None:
+def make_output_dir(config: MkosiConfig, uid: int, gid: int) -> None:
```
behrmann, line 2857: Stylistic nit, I'd prefer this to mirror the signature of `build_stuff`.

mkosi/__init__.py
```diff
@@ -3296,7 +3296,8 @@ def need_cache_tree(state: MkosiState) -> bool:
 
 
 def build_stuff(uid: int, gid: int, config: MkosiConfig) -> None:
-    workspace = tempfile.TemporaryDirectory(dir=config.workspace_dir or Path.cwd(), prefix=".mkosi.tmp")
+    make_output_dir(config, uid, gid)
+    workspace = tempfile.TemporaryDirectory(dir=config.workspace_dir or config.output_dir or Path.cwd(), prefix=".mkosi.tmp")
```
behrmann, line 3300: Maybe the fallback to the current working directory should also be explicitly mentioned in the docs.

Cornelicorn: The current working directory should be the output directory if it is not specified so it is there implicitely, but I can add a comment along the lines of 
> ..., which is the current working directory if not specified otherwise with the config option `OutputDirectory=` or the commandline flags `--output-dir=` or  `-O`...

DaanDeMeyer:
Shouldn't we just update the documentation instead? Do you actually have a usecase for having the workspace in the output directory?

mkosi/__init__.py
behrmann: Yes, it's implicitly clear, but explicit is better than implicit, since it requires less mental load on the reader to make the connection.

Cornelicorn:
I also don't see a necessity to do this and `WorkspaceDirectory` should be able to do everything. This would just be a convenience factor when the config is stored on a tmpfs where the image wouldn't fit. In this case you would have to set both `OutputDirectory` and `WorkspaceDirectory`.

DaanDeMeyer:
> I also don't see a necessity to do this and WorkspaceDirectory should be able to do everything. This would just be a convenience factor when the config is stored on a tmpfs where the image wouldn't fit. In this case you would have to set both OutputDirectory and WorkspaceDirectory.

Can you update the documentation then instead? The PR looks good but keeping the current behavior is simpler in terms of logic so I'd prefer to keep that.

DaanDeMeyer (approved):

PR #1396, opened 2023-03-17
fedora: Default to the minimal authselect profile if authselect is installed

DaanDeMeyer:


poettering:
really new authselect has a "home" feature one can enable. we should enable that.

DaanDeMeyer:
Let's wait with doing that until it's more widespread. We can test it out in the systemd repo by enabling it in the postinstall script.

behrmann:
Is `authselect` something like the Debian PAM config snippets>

DaanDeMeyer:
No idea, I don't know how Debian handles this

poettering:
> Let's wait with doing that until it's more widespread. We can test it out in the systemd repo by enabling it in the postinstall script.

the snippet should have zero effect unless homed is also used and has an account. the fact that the thing is hidden behind a flag is pretty pointless if you ask me, more a political choice than a technical one, given the stuff is as non-intrusive as it could be.

DaanDeMeyer:
We use home if it's there and minimal otherwise

poettering:
btw, you might just write `/etc/authselect/authselect.conf` directly, and set it to the raw contents:

```
minimal
with-homed
```

and do that ahead of time before invoking dnf. so that the system never comes up with crap config we then later turn off again.

PR #1397, opened 2023-03-17
centos: Add missing base features for ext4 on c8s

DaanDeMeyer:
Accidentally only added the ext4 specific features and none of the base features, let's make sure we add the base features as well, especially sparse_super is important to make repart's minimize guessing logic work properly.


DaanDeMeyer:
Trivial bugfix so merging

PR #1398, opened 2023-03-17
Add source target support to extra and skeleton trees

DaanDeMeyer:


mkosi.md
```diff
@@ -544,12 +544,15 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
...
-  purpose (also see the "Files" section below).
+: Takes a colon separated pair of paths. The first path refers to a
+  directory to copy into the OS tree before invoking the package
+  manager. The second path refers to the target directory inside the
+  image. If the second path is not provided, the directory is copied
+  on top of the root directory of the image. Use this to insert files
+  and directories into the OS tree before the package manager installs
+  any packages. If this option is not used, but the `mkosi.skeleton/`
+  directory is found in the local directory it is automatically used
+  for this purpose (also see the "Files" section below).
```
behrmann, line 555: Let's maybe be super clear about `mkosi.skeleton/` doesn't support a target and uses the root directory
```suggestion
  for this purpose with the root directory as target (also see the "Files" section below).
```
same below for `ExtraTree=`

mkosi/__init__.py
```diff
@@ -1434,6 +1440,12 @@ def parse_ssh_agent(value: str) -> Optional[Path]:
        mkosi --version
 """.format(b=MkosiPrinter.bold, e=MkosiPrinter.reset)
 
+
+def parse_source_target_paths(value: str) -> tuple[Path, Optional[Path]]:
+    split = value.split(':', maxsplit=1)
+    return (Path(split[0]), Path(split[1]) if len(split) > 1 else None)
```
behrmann, line 1446: Maybe a bit less sigil noise
```suggestion
    src, _, target = value.partition(':')
    return Path(src), Path(target) if target else None
```

mkosi/__init__.py
```diff
@@ -2726,6 +2745,21 @@ def line_join_list(
...
+    if not array:
+        return "none"
+
+    items = []
+    for source, target in array:
+        item = str(source)
+        if target:
+            item = f"{source}:{target}"
+
+        items += [item]
```
behrmann, line 2758:
```suggestion
    items = [f"{source}:{target}" if target else f"{source}" for source, target in array]
```

mkosi/install.py
```diff
@@ -82,10 +82,15 @@ def install_skeleton_trees(state: MkosiState, cached: bool, *, late: bool=False)
...
+                t = state.root / str(target).lstrip("/")
+
+            t.mkdir(mode=0o755, parents=True, exist_ok=True)
+            if source.is_dir():
+                copy_path(source, t, preserve_owner=False)
             else:
                 # unpack_archive() groks Paths, but mypy doesn't know this.
                 # Pretend that tree is a str.
-                shutil.unpack_archive(tree, state.root)
+                shutil.unpack_archive(source, t)
```
behrmann, line 96: We've got the exact same logic here and in `__init__.py`for the extra tree, this should probably live in its own function in this module.

mkosi/__init__.py
```diff
@@ -659,13 +659,19 @@ def install_extra_trees(state: MkosiState) -> None:
         return
 
     with complete_step("Copying in extra file trees…"):
-        for tree in state.config.extra_trees:
-            if tree.is_dir():
-                copy_path(tree, state.root, preserve_owner=False)
+        for source, target in state.config.extra_trees:
+            t = state.root
+            if target:
+                t = state.root / str(target).lstrip("/")
```
behrmann, line 665:
```suggestion
                t = state.root / target.relative_to("/")
```
This has the benefit of not round-tripping it through a string and also breaking for relative paths, thus making it harder to escape to path above `state.root`, well, at least I think of that as a benefit.

behrmann (commented):
Functionality looks good, a few nits.

mkosi/install.py
DaanDeMeyer: Lets leave this for another PR

behrmann (approved):

PR #1399, opened 2023-03-17
Drop --ssh-key and --ssh-agent options

DaanDeMeyer:
These don't work properly anymore with unprivileged mkosi and these days, we have a much better way of provisioning the public key, add a script "ssh.authorized_keys.root" to mkosi.credentials/ with the following contents:

```
#!/bin/sh
ssh-add -L
```


bluca (approved):

behrmann:
LGTM and happy to drop this, especially since @bluca agrees, but a lot of CI seems to choke on vsock.

PR #1407, opened 2023-03-23
Only parse configuration files suffixed with ".conf" in dropin directories

### LINKED ISSUE #1406: Only parse configuration files suffixed with ".conf" in dropin directories
I've encountered crash when run `mkosi` command. The problem was it tried to parse `.*.swp` file in `conf.d` directory. I opened config on server, vim created its file, I got python traceback.

Maybe filter out hidden files here?
https://github.com/systemd/mkosi/blob/beee070079e057b83cc8cae791e7b140f6f4b005/mkosi/__init__.py#L2108

nikandfor:
Fixes #1406


nikandfor:
Do we want to hidden files were parsed? Now they are parsed.

DaanDeMeyer:
Yes, we don't seem to ignore hidden files in systemd config file parsing either, so we don't want to do so in mkosi either.

PR #1408, opened 2023-03-24
overlayfs: avoid userxattr if kernel version is < 5.11

bluca:
userxattr was added as a mount option for overlayfs only in kernel 5.11, so avoid it on older versions (like Debian stable):

‣   Running build script…
mount: /home/bluca/git/systemd/.mkosi.tmpzt2q0dyp/root: wrong fs type, bad option, bad superblock on overlay, missing codepage or helper program, or other error.
umount: /home/bluca/git/systemd/.mkosi.tmpzt2q0dyp/root: not mounted

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2590,6 +2590,11 @@ def load_args(args: argparse.Namespace) -> MkosiConfig:
             if not p.is_file():
                 die(f"Initrd {p} is not a file")
 
+    # For unprivileged builds we need the userxattr OverlayFS mount option, which is only available in Linux v5.11 and later.
+    with prepend_to_environ_path(args.extra_search_paths):
+        if (args.build_script is not None or args.base_image is not None) and GenericVersion(platform.release()) < GenericVersion("5.11") and os.geteuid() != 0:
+            die("This unprivileged build configuration requires at least Linux v5.11")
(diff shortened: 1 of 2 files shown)


DaanDeMeyer:
Does unprivileged mkosi in the systemd repo work without userxattr? I remember running into issues without adding this option. If that's the case, we should fail earlier when loading arguments if we're not running as root, kernel <= 5.11 and a build script is provided or --base-image is used and then only set userxattr here if we're not running as root

bluca:
> Does unprivileged mkosi in the systemd repo work without userxattr? I remember running into issues without adding this option. If that's the case, we should fail earlier when loading arguments if we're not running as root, kernel <= 5.11 and a build script is provided or --base-image is used and then only set userxattr here if we're not running as root

Added early check and bail

DaanDeMeyer (approved):

bluca:
I wish we could have a sane way of checking for mount options availability, rather than having to rely on the version, but I don't think there's any way

DaanDeMeyer:
> I wish we could have a sane way of checking for mount options availability, rather than having to rely on the version, but I don't think there's any way

You can do it using the new mount API, but a bit awkward to do that in python, so kernel version is fine

PR #1410, opened 2023-03-27
Add --acl option

DaanDeMeyer:
Let's make the usage of ACLs opt-in since they leak into the final image if the directory output is used.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1008,7 +1008,10 @@ def calculate_bmap(state: MkosiState) -> None:
         run(cmdline)
 
 
-def acl_toggle_remove(root: Path, uid: int, *, allow: bool) -> None:
+def acl_toggle_remove(config: MkosiConfig, root: Path, uid: int, *, allow: bool) -> None:
+    if not config.acl:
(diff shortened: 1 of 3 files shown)


mkosi/__init__.py
```diff
@@ -3345,6 +3354,8 @@ def build_stuff(uid: int, gid: int, config: MkosiConfig) -> None:
             if p.name.startswith(state.config.output.name):
                 compress_output(state.config, p, uid=state.uid, gid=state.gid)
 
+    print_output_size(config)
```
behrmann, line 3357: That's a bit of a non-topical addition here?

behrmann (approved):
LGTM, though I could equally well see this be opt-out instead of opt-in, but I guess stemming the ACL leaking in will be something more people want more than the removal.

mkosi/__init__.py
DaanDeMeyer: It's required to make this work actually, we don't have permission to read certain directories outside of the user namespace, so calculating the size of the output only works inside the image if we're outputting a directory.

PR #1412, opened 2023-03-27
Strip the stub before passing it to ukify

daandemeyer:
Older stubs need some preprocessing to get accepted by ukify.

See https://github.com/systemd/systemd/pull/27009#issuecomment-1485743500

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -868,11 +868,19 @@ def install_unified_kernel(state: MkosiState, roothash: Optional[str]) -> None:
             # nul terminators in argv so let's communicate the cmdline via a file instead.
             state.workspace.joinpath("cmdline").write_text(f"{' '.join(cmdline).strip()}\x00")
 
+            run([
+                "strip",
+                "-s",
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
@medhefgo Running a strip seems to make us hit the assertion again. Any suggestions on how to invoke strip or objcopy here to fix the stub?

medhefgo:
I don't understand why. It works for me locally. Maybe too old binutils? You could try llvm-strip.

DaanDeMeyer:
llvm-strip does correctly ensure aligned values for PointerToRawData.

keszybz:
I think `mkosi` is the wrong place to do this. Especially main code in mkosi. We shouldn't get into the business of fixing kernels here.

keszybz:
Yeah, sloppy wording on my part. It is even more so with the stub, which is a component under our control.

DaanDeMeyer:
See https://github.com/systemd/systemd/pull/27035 which fixes the issue in ukify

DaanDeMeyer:
Let's close this one as we can fix it in ukify

PR #1415, opened 2023-03-29
Drop --netdev

DaanDeMeyer:
Let's not configure systemd-networkd in mkosi, but leave this up to users instead. As our SSH support now works on top of vsock, we don't need this option anymore to make the SSH support work.

We add network devices by default now when starting a container or VM.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -49,7 +49,6 @@ from mkosi.backend import (
     detect_distribution,
     flatten,
     format_rlimit,
-    is_centos_variant,
     is_dnf_distribution,
     patch_file,
     set_umask,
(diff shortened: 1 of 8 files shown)


mkosi/__init__.py
```diff
@@ -2570,9 +2557,6 @@ def load_args(args: argparse.Namespace) -> MkosiConfig:
     if args.repo_dirs:
         args.repo_dirs = [p.absolute() for p in args.repo_dirs]
 
-    if args.netdev and is_centos_variant(args.distribution) and "epel" not in args.repositories:
```
behrmann, line 2573: I think that's the last use of `is_centos_variant`, so that can be removed.

behrmann (approved):

PR #1416, opened 2023-03-29
Always use mkosi.extra/ and mkosi.skeleton/ if they exist

DaanDeMeyer:
These should be additive instead of only being used when no explicit extra/skeleton trees are configured.


behrmann:
Should also get a news entry for the unlikely case that somebody overwrites `mkosi.skeleton` just by adding some config line to a different dir.

PR #1417, opened 2023-03-30
qemu: Use virtconsole serial console instead of regular serial console

DaanDeMeyer:
With --qemu-headless, qemu stops working when latest systemd is not
available because smbios provided kernel cmdline arguments are ignored.
This includes "console=ttyS0" meaning we don't get any output on the
serial console when booting in qemu.

As a workaround, let's use qemu's virtconsole instead. This means the
serial console will be available in the VM as /dev/hvc0, on which
systemd will automatically spawn a serial getty. This means we'll
eventually get a login prompt, even if latest systemd is not used.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -672,7 +672,7 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
 `Autologin=`, `--autologin`
 
 : Enable autologin for the `root` user on `/dev/pts/0` (nspawn),
-  `/dev/tty1` and `/dev/ttyS0` by patching `/etc/pam.d/login`.
+  `/dev/tty1` and `/dev/hvc0` by patching `/etc/pam.d/login`.
 
 `BuildScript=`, `--build-script=`
 
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -16,7 +16,6 @@ import os
 import platform
 import re
 import resource
-import shlex
 import shutil
 import string
 import subprocess
@@ -125,11 +124,6 @@ def list_to_string(seq: Iterator[str]) -> str:
     ['a', "b", 11] → "'a', 'b', 11"
     """
     return str(list(seq))[1:-1]
-
-
-def print_running_cmd(cmdline: Iterable[PathString]) -> None:
-    MkosiPrinter.print_step("Running command:")
-    MkosiPrinter.print_step(" ".join(shlex.quote(str(x)) for x in cmdline) + "\n")
 
 
 # EFI has its own conventions too
@@ -3493,7 +3487,6 @@ def qemu_check_kvm_support() -> bool:
 def start_swtpm() -> Iterator[Optional[Path]]:
 
     if not shutil.which("swtpm"):
-        MkosiPrinter.info("Couldn't find swtpm binary, not invoking qemu with TPM2 device.")
         yield None
         return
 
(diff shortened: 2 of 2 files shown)


DaanDeMeyer:
@medhefgo With this change, I stop getting the sd-stub console output. Any idea what might be causing that?

medhefgo:
Afaik EDK2 does not do virtio-serial.

DaanDeMeyer:
I found a few references to serial over mmio in the edk2 source code, maybe I have to modify the approach to use that instead of PCI? Or is the transport not the issue here?

medhefgo:
You're asking the wrong person.

You can create a serial device and connect it to the virtconsole with `-serial chardev:console`, but that doesn't seem to react to any input on the console.

DaanDeMeyer:
Maybe @kraxel can help out here?

DaanDeMeyer:
I looked past the most obvious solution, just add a virtio serial device and a regular one as well just for use by the bootloader.

DaanDeMeyer:
I added the regular serial console as well. Being able to provide input to the serial console isn't really that important. We just need to make sure we get the bootloader output.

DaanDeMeyer:
To do input you might need to use ctrl-a + c to switch between backends to get to the regular serial console first.

kraxel:
> Maybe @kraxel can help out here?

Yes, there is no virtio-serial driver in edk2, so there will be no bootloader / stub output ...
Or was the question something else?

DaanDeMeyer:
> Yes, there is no virtio-serial driver in edk2, so there will be no bootloader / stub output ...
> Or was the question something else?

I was also wondering if there's a better alternative to what I've come up with here (adding both a regular serial console for the bootloader and a virtio-serial virtconsole for linux).

kraxel:
> I was also wondering if there's a better alternative to what I've come up with here (adding both a regular serial console for the bootloader and a virtio-serial virtconsole for linux).

I don't think so given this is for compatibility with already released, older versions.

DaanDeMeyer:
@kraxel Unrelated, but do you know if I can configure edk2 to not clear the terminal window when booting up in a terminal with qemu? Currently, when building an image with mkosi and then immediately booting it with qemu, I lose the last lines of mkosi output in my terminal because edk2 clears the visible terminal window on startup. If possible, I'd like to disable that behavior so it just writes output without clearing the terminal window first.

DaanDeMeyer:
> Not sure why they all are there, it's not like edk2 uses any advanced terminal features so for the most part it doesn't make much of a difference which one you pick. But the driver expects they all understand basic ansi/xterm-style escape sequences. There isn't something simliar to TERM=dumb.

That's a shame. Does edk2 have to clear the screen by default when connected to a serial console? Can I force a text mode only or so where it wouldn't clear the screen?

> Also unrelated: You might want play with https://github.com/kraxel/edk2/commits/devel/virtio-serial

Looks great! That'll allow us to drop the regular serial console workaround in a few years when this is widely available.

kraxel:
> > Not sure why they all are there, it's not like edk2 uses any advanced terminal features so for the most part it doesn't make much of a difference which one you pick. But the driver expects they all understand basic ansi/xterm-style escape sequences. There isn't something simliar to TERM=dumb.
> 
> That's a shame. Does edk2 have to clear the screen by default when connected to a serial console? Can I force a text mode only or so where it wouldn't clear the screen?

Console setup clears the screen somewhere and I don't think a making that configurable is going to be accepted.
Adding a 'dumb' terminal type for logfile-style usage (and don't send any escape sequences to it) might work.

PR #1420, opened 2023-04-03
Rework configuration parsing

DaanDeMeyer:


mkosi/config.py
```diff
@@ -0,0 +1,275 @@
...
+        Distribution.gentoo: "17.1",
+    }.get(d, "rolling")
+
+    setattr(namespace, dest, d)
+    setattr(namespace, "release", r)
+
+
+ConfigParseCallback = Callable[[str, str, Optional[str], argparse.Namespace], None]
+ConfigMatchCallback = Callable[[str, str, str, argparse.Namespace], bool]
+
```
behrmann, line 113: Missing line 
```suggestion
ConfigMatchCallback = Callable[[str, str, str, argparse.Namespace], bool]

```

mkosi/config.py
```diff
@@ -0,0 +1,275 @@
...
+@dataclasses.dataclass(frozen=True)
+class ConfigSetting:
+    dest: str
+    section: str
+    parse: ConfigParseCallback = config_parse_string
+    match: Optional[ConfigMatchCallback] = None
+    name: str = ""
+    default: Any = None
+    paths: tuple[str, ...] = tuple()
```
behrmann, line 122:
```suggestion
    paths: tuple[str, ...] = field(default_factory=tuple)
```

mkosi/config.py
```diff
@@ -0,0 +1,275 @@
...
+        die("Target path must be absolute")
+    return Path(src), Path(target) if target else None
+
+
+def config_parse_string(name: str, dest: str, value: Optional[str], namespace: argparse.Namespace) -> None:
+    setattr(namespace, dest, value)
+
+
+def config_match_string(name: str, dest: str, value: str, namespace: argparse.Namespace) -> bool:
+    return bool(value == getattr(namespace, dest))
```
behrmann, line 38: The bool call here is a bit weird. I think this is rather an artifact of mypy not being able to look into `namespace`. Let's make this a type-check time thing instead of a runtime thing
```suggestion
    return cast(bool, value == getattr(namespace, dest))
```

mkosi/config.py
```diff
@@ -0,0 +1,275 @@
...
+
+    config_parse_path(name, dest, value, namespace)
+
+
+def config_parse_boolean(name: str, dest: str, value: Optional[str], namespace: argparse.Namespace) -> None:
+    setattr(namespace, dest, parse_boolean(value) if value is not None else False)
+
+
+def config_match_boolean(name: str, dest: str, value: str, namespace: argparse.Namespace) -> bool:
+    return bool(getattr(namespace, dest) == parse_boolean(value))
```
behrmann, line 63: for symmetry with the above one
```suggestion
    return cast(bool, parse_boolean(value) == getattr(namespace, dest))
```

mkosi/config.py
```diff
@@ -0,0 +1,275 @@
...
+    name: str = ""
+    default: Any = None
+    paths: tuple[str, ...] = tuple()
+
+    def __post_init__(self) -> None:
+        if not self.name:
+            object.__setattr__(self, 'name', ''.join(x.capitalize() for x in self.dest.split('_') if x))
+
+
+class ConfigParser:
```
behrmann, line 129: Can we name this a bit differently so as to make it clearly different from `configparser.ConfigParser`? Maybe just
```suggestion
class MkosiConfigParser:
```

mkosi/__init__.py
```diff
@@ -568,6 +570,9 @@ def install_boot_loader(state: MkosiState) -> None:
         return
 
     if state.config.secure_boot:
+        assert state.config.secure_boot_key
+        assert state.config.secure_boot_certificate
```
behrmann, line 574: This should probably be checked earlier (and less often). Up until now this would blow up in `load_args`, so if we have to assert this now later, this seems like a small regression. I wonder if mypy could be placated to accept that keys exist if there were a dynamic validation check that checks that some dependent keys exist. Probably too difficult with the simple namespaces.

behrmann (commented):
Had a first go through this.

I'm not quite sure that dropping the bmap option is really necessary, but I guess there's also not that many users.

I do think, though, that we should keep the skeleton trees. Even without the possibility to seed package manager config, populating the tree before the package managers runs, is something that can be really helpful to prevent maintainer scripts looking for existing files from wreaking havoc, that can be hard to clean up in later scripts.

DaanDeMeyer:
> I do think, though, that we should keep the skeleton trees. Even without the possibility to seed package manager config, populating the tree before the package managers runs, is something that can be really helpful to prevent maintainer scripts looking for existing files from wreaking havoc, that can be hard to clean up in later scripts.

Do you have some concrete examples? If a package can't bootstrap itself without certain files being available, I'd consider that a bug in the package to be fixed, not something we should accomodate

mkosi/config.py
DaanDeMeyer: tuple is immutable, so there's no point in using default_factory

mkosi/__init__.py
DaanDeMeyer: We still check this in load_args()

bluca:
yeah preseeding is quite useful

DaanDeMeyer:
Can I get an example of how preseeding would work with mkosi --skeleton-tree? I did some googling but didn't immediately find anything useful

DaanDeMeyer:
Regardless, dropped the commit that removes --skeleton-tree for now so we can make progress on the rest.

behrmann:
> Can I get an example of how preseeding would work with mkosi --skeleton-tree? I did some googling but didn't immediately find anything useful

Debconf stores it's stuff in `/var/cache/debconf` and since we don't set anything otherwise I *assume* dpkg will dpkg will resolve it relative to `--root`. 

Generally, can't we make skeleton trees compatible with calling the package manager from the outside by dumping the files from the skeleton tree into the image as well as into `state.workspace` - maybe if we put the stuff the currently just lives in the workspace one hierarchy level deeper to get some separation, like e.g. this line from `invoke_apt`
```diff
-Dir::State "{state.workspace / "apt"}";
+Dir::State "{state.workspace / "package_manager" / "apt"}";
```

DaanDeMeyer:
> Generally, can't we make skeleton trees compatible with calling the package manager from the outside by dumping the files from the skeleton tree into the image as well as into state.workspace - maybe if we put the stuff the currently just lives in the workspace one hierarchy level deeper to get some separation, like e.g. this line from invoke_apt

Perhaps, but that's a discussion for another PR, let's focus on getting the config rework in first

NEWS.md
```diff
@@ -58,8 +58,10 @@
   runtime. Use the new `--qemu-gui` option to start qemu in its graphical interface.
 - Removed `--netdev`. Can be replaced by manually installing systemd-networkd, putting a network file in the
   image and enabling systemd-networkd.
-- If `mkosi.extra/` or `mkosi.skeleton/` exist, they are now always used instead of only when no explicit
-  extra/skeleton trees are defined.
+- If `mkosi.extra/` exists, it is now always used instead of only when no explicit extra trees are defined.
+- Removed `--qcow2` option in favor of supporting only raw disk images as the disk image output format.
+- Removed `--bmap` option as it can be trivially added manually by utilizing a finalize script.
```
behrmann, line 63: That's maybe very nitpicky, but the commit message dropping `--bmap` too only mentions `--qcow2`, adding `--bmap` there might be nice when looking for it in the log.

mkosi/run.py
```diff
@@ -11,7 +11,7 @@
 import traceback
 from pathlib import Path
 from types import TracebackType
-from typing import Any, Callable, Iterable, Mapping, Optional, Sequence, Type, TypeVar
```
behrmann, line 14: That ended up in the wrong commit during a rebase. It should be in *Print cmdline in run() before doing fspath() conversion* but ended up in the one afterwards. Not much of an issue, but could trip up bisecting.

mkosi/config.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+def config_make_list_parser(delimiter: str, parse: Callable[[str], Any] = str) -> ConfigParseCallback:
+    def config_parse_list(dest: str, value: Optional[str], namespace: argparse.Namespace) -> None:
+        if not value:
+            setattr(namespace, dest, [])
+            return
+
+        value = value.replace(os.linesep, delimiter)
+        values = [v for v in value.split(delimiter) if v]
```
behrmann, line 244: While `os.linesep` is the correct thing, we don't support `mkosi` on anything other than running on Linux and can hardcode `\n` to make this easier to follow.

Alternatively, can this could also be
```suggestion
        values = [v for line in value.splitlines() for v in line.split(delimiter) if v]
```

behrmann (commented):
Some nits otherwise LGTM.

PR #1426, opened 2023-04-07
make sure mkfs randomises its UUID

257:


DaanDeMeyer:
We should already be using randomized uuids since we remove the machine id before running repart. Are you running into cases where this is not happening?

257:
yes, almost 90% failure here. but i realise we don't hit this with CI.

257:
> We should already be using randomized uuids since we remove the machine id before running repart.

hmm, let me dig a bit deeper, i might be a local problem at my end.

DaanDeMeyer:
If you see "No machine ID set, using randomized partition UUIDs." in the mkosi outputs, it should be using randomized UUIDs, if it isn't, that's a bug in systemd-repart

keszybz:
@257 did you figure this one out?

257:
> @257 did you figure this one out?

yes

PR #1428, opened 2023-04-07
Simplify autologin

DaanDeMeyer:
agetty actually has a -f option to skip pam authentication, so let's use that instead of all the pam hacks to get autologin.

### DIFF
--- a/mkosi/resources/serial_getty_autologin.conf
+++ b/mkosi/resources/serial_getty_autologin.conf
@@ -1,5 +1,5 @@
 [Service]
 ExecStart=
-ExecStart=-/sbin/agetty -o '-p -- \\u' --autologin root --keep-baud 115200,57600,38400,9600 - $TERM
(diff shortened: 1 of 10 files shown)


mkosi/resources/serial_getty_autologin.conf
```diff
@@ -1,5 +1,5 @@
 [Service]
 ExecStart=
-ExecStart=-/sbin/agetty -o '-p -- \\u' --autologin root --keep-baud 115200,57600,38400,9600 - $TERM
+ExecStart=-/sbin/agetty -o '-f -p -- \\u' --autologin root --noclear --keep-baud 115200,57600,38400,9600 - $TERM
```
bluca, line 3: `--noclear` is new here, was that intentional? If so please mention in the commit message

bluca (approved):

mkosi/resources/serial_getty_autologin.conf
DaanDeMeyer: Dropped until I figure out why all the other getty units have it and serial-getty doesn't.

PR #1429, opened 2023-04-11
Add --die-with-parent when invoking bubblewrap

DaanDeMeyer:
rpm blocks most signals when doing a transaction. dnf doesn't block anything. The end result is that when we do ctrl+c when dnf is running the rpm transaction, dnf gets interrupted and exits, which means mkosi exits and tries to remove the workspace directory on which rpm is still operating, causing all kinds of nastiness.

Because we don't care about transaction safety since we're operating on a chroot, let's add --die-with-parent when running bubblewrap. This makes bubblewrap ensure cleanup of all child processes underneath it, and because rpm can't ignore SIGKILL, it also cleans up rpm properly, fixing the issue.

### DIFF
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -265,6 +265,7 @@ def run_with_apivfs(
         "--dev", state.root / "dev",
         "--ro-bind", "/sys", state.root / "sys",
         "--bind", state.var_tmp, state.root / "var/tmp",
+        "--die-with-parent",
         *bwrap_params,
         "sh", "-c",
     ]
@@ -302,6 +303,7 @@ def run_workspace_command(
         "--proc", "/proc",
         "--ro-bind", "/sys", "/sys",
         "--bind", state.var_tmp, "/var/tmp",
+        "--die-with-parent",
         *bwrap_params,
     ]
 


poettering:
hmm? relying on newuidmap is hardly without privileges?

Either everything in the image is owned by exactly by one user/group now (i.e. by root:root), or you depend on newuidmap, but then it is privileged.

I am really not a fan of such games. newuidmap is an abomination, not a scheme to support.

poettering:
i mean, if you are fine with newuidmap being suid, and being called with mkosi why the entire excercise? you could as well just mark nspawn as suid root? and claim "unprivileged!" the same way!

DaanDeMeyer:
> i mean, if you are fine with newuidmap being suid, and being called with mkosi why the entire excercise? you could as well just mark nspawn as suid root? and claim "unprivileged!" the same way!

newuidmap is shipped suid by distros though, nspawn isn't.

I will happily adopt the userdb stuff to do user namespaces once it's widely available (assuming you intend to support the mkosi usecase), but not supporting unprivileged mkosi until we have that seems like overkill.

DaanDeMeyer:
I mean, either we don't use newuidmap and need to run as root, or we use newuidmap and we're privileged because newuidmap is setuid. Regardless of whether mkosi uses newuidmap or not, it's available on all distros these days and marked setuid. If an attacker wants to try to use it to gain privileges, they can always do that, regardless of whether mkosi uses it or not.

If it's available regardless and the security is the same (which it is because an attacker can just invoke newuidmap directly, no need to go via mkosi), we might as well use it so that users don't have to enter their password all the time to use mkosi.

PR #1432, opened 2023-04-13
Package installation streamlining

DaanDeMeyer:


mkosi/distributions/centos.py
```diff
@@ -89,32 +94,6 @@ def install(cls, state: MkosiState) -> None:
...
-        # move it back to /var/lib/rpm on CentOS.
-        move_rpm_db(state.root)
-
-        # Centos Stream 8 and below can't write to the sqlite db backend used by
-        # default in newer RPM releases so let's rebuild the DB to use the old bdb
-        # backend instead. Because newer RPM releases have dropped support for the
-        # bdb backend completely, we check if rpm is installed and use
-        # run_workspace_command() to rebuild the rpm db.
-        if release <= 8 and state.root.joinpath("usr/bin/rpm").exists():
-            cmdline = ["rpm", "--rebuilddb", "--define", "_db_backend bdb"]
```
behrmann, line 113: Considering how often mucking around with this has bitten us, this should definitely be mentioned in the NEWS.

mkosi/distributions/fedora.py
```diff
@@ -50,59 +86,6 @@ def parse_fedora_release(release: str) -> tuple[str, str]:
...
-
-    # Fedora defaults to sssd authselect profile, let's override it with the minimal profile if it exists and
-    # extend it with the with-homed feature if we can find it.
-    if state.root.joinpath("usr/share/authselect/default/minimal").exists():
-        run_workspace_command(state, ["authselect", "select", "minimal"])
-
-        features = run_workspace_command(state, ["authselect", "list-features", "minimal"],
-                                         stdout=subprocess.PIPE).stdout.split()
-        if "with-homed" in features:
-            run_workspace_command(state, ["authselect", "enable-feature", "with-homed"])
```
behrmann, line 103: I guess dropping this is fine, you only added it a month ago, but I can also see some appeal in leaving it in, because cleaning up the installation of SSSD in a postinst seems slightly more icky to me than starting of with a cleaner, minimal situation. 

Is there maybe a way to cleanly do this in with a skeleton?

mkosi/run.py
```diff
@@ -266,6 +266,19 @@ def run_with_apivfs(
...
+
+    # If passwd or a related file exists in the root directory, bind mount it over the host files while we
+    # run the command, to make sure that the command we run uses user/group information from the root instead
+    # of from the host. If the file doesn't exist yet, mount over /dev/null instead.
+    for f in ("passwd", "group", "shadow", "gshadow"):
+        p = state.root / "etc" / f
+        if p.exists():
+            cmdline += ["--bind", p, f"/etc/{f}"]
+        else:
+            cmdline += ["--bind", "/dev/null", f"/etc/{f}"]
```
behrmann, line 279: This seems very sensible, but isn't there the potential for breakage here? At least in the Debian case I know that postinst scripts regularly look whether users exist and create them if they don't. I'm sure RPM postinst scriplets do the same. Assuming that the writes to e.g. `/etc/passwd` now go to `/dev/null`, won't the created user be missing in the image?

DaanDeMeyer: If RPM or dpkg are writing to host /etc/passwd when installing in a chroot, something is seriously wrong. This fix is more about making sure that when installing files from an RPM or deb, the package managers use users/groups from the chroot instead of from the host. What could happen now is that instead of using the wrong uid for a user, the user simply won't be found anymore if we mount /dev/null over /etc/passwd. But I don't expect that to happen given the 2 stage package install process we use.

mkosi/distributions/fedora.py
DaanDeMeyer: It's very hard to do this correctly when you consider base images and such. If we do this in install(), it won't always apply since authselect might not always be installed immediately. If we do this in install_packages(), we might be overwriting the postinst script from a base image. This is why I think it's better to leave this to the user to configure.

Also, the debian preset issue we ran into in systemd shows that mkosi mucking around with distro defaults can be very confusing for users. People expect their distro to behave the same as usual regardless of whether it's installed by mkosi or not, which shows that we shouldn't modify defaults unless it's really important.

DaanDeMeyer:
Only changed NEWS, let's merge

PR #1434, opened 2023-04-13
Drop --hostname and default locale configuration

DaanDeMeyer:
These are trivial to configure manually and with credentials, so let's drop the options for them.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -161,6 +161,9 @@
   image. Note that this option is currently only supported for `pacman` and
   `dnf`-based distros.
 - Option `--skeleton-tree` is now supported on Debian-based distros.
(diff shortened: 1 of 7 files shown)


behrmann (approved):
I think this change is ok, considering firstboot can do this now and the existence of `mkosi.credentials`. Pretty sure this is going to break on the still current Debian stable, though.

bluca:
what's required to make this work?

DaanDeMeyer:
> what's required to make this work?

On stable? No clue, on testing it seems to work without issues going by CI

bluca (approved):
debian stable also builds and boots from this branch, so it's probably good enough

PR #1435, opened 2023-04-14
Rework opensuse module

DaanDeMeyer:
Instead of doing everything via zypper, let's just write the repo
definition files ourselves directly. Also, to mimick what we do
with dnf, let's write the repo files outside of the root instead of
inside the root and leave installing repos inside the root up to
the user.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1141,7 +1141,7 @@ def load_args(args: argparse.Namespace) -> MkosiConfig:
             else:
                 args.mirror = "https://geo.mirror.pkgbuild.com"
         elif args.distribution == Distribution.opensuse:
-            args.mirror = "http://download.opensuse.org"
+            args.mirror = "https://download.opensuse.org"
         elif args.distribution == Distribution.rocky:
             args.mirror = None
         elif args.distribution == Distribution.alma:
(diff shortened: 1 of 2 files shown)


behrmann:
LGTM, let's see what CI says. 

@fbuihuu Any thoughts on this? Streamlining this with the other RPM distros would be tremendously helpful.

DaanDeMeyer:
Merging to make progress, we'll address any feedback in a follow up PR

fbuihuu:
> We can't work around every distro peculiarity in mkosi. If suse forbids the empty password then that's something we'll have to accept. If you want to allow the empty root password, please take it up with the opensuse maintainers or do it via a postinstall script.

Maybe we could at least refuse to generate images with an empty password or emit a warning on distros where it's no allowed otherwise I guess many users will lost time since it's mkosi default behavior.

> Doesn't Suse ship a package that provides the repos in the image like fedora, centos and such do (fedora-repos, ...)? 

I'm not aware of such package. If it doesn't exist then probably no one needed it until mkosi introduced such a need.

> Yes this is a deliberate change. mkosi can be used for all kinds of use cases so we decided to drop most of our defaults. If a kernel and systemd-boot are installed, we'll set up a bootable image. We added documentation to docs/bootable.md with example configs to generate bootable images.

I guess that's not really user friendly as now users need to figure out the name of the packages needed to create useful (bootable) images. Is docs/bootable.md accessible by non developer users ?

PR #1437, opened 2023-04-14
Build systemctl in the github action

DaanDeMeyer:
To make sure systemctl understands to new ignore directive in preset.


DaanDeMeyer:
Trivial so merging immediately

PR #1441, opened 2023-04-15
fedora: use dnf5 if available

keszybz:
Fedora plans to switch to dnf5 for F40. For now, the new binary is called 'dnf5', and the old 'dnf'. Let's try to call 'dnf5' if available. It is generally faster. E.g.
'mkosi build' in the mkosi directory after all packages have been downloaded:
   950 ms with dnf5
  1350 ms with dnf

Dnf5 also doesn't download file metadata by default, so it'll be faster if downloads are necessary too.

Let's try the new shiny to save some time on installations.

For some reason, dnf5 doesn't like it if command options are before the command. Old dnf is fine with either order, so let's just move the command earlier.

### DIFF
--- a/mkosi/distributions/fedora.py
+++ b/mkosi/distributions/fedora.py
@@ -159,9 +159,10 @@ def invoke_dnf(state: MkosiState, command: str, packages: Iterable[str], env: Ma
     state.workspace.joinpath("vars").mkdir(exist_ok=True)
 
     cmdline = [
-        'dnf' if shutil.which('dnf') else 'yum',
+        shutil.which('dnf5') or shutil.which('dnf') or 'yum',
         "-y",
         f"--config={state.workspace.joinpath('dnf.conf')}",
+        command,
         "--best",
         "--allowerasing",
         f"--releasever={release}",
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
Let's drop the yup fallback while we're at it.

mkosi/distributions/fedora.py
```diff
@@ -159,9 +159,10 @@ def invoke_dnf(state: MkosiState, command: str, packages: Iterable[str], env: Ma
     state.workspace.joinpath("vars").mkdir(exist_ok=True)
 
     cmdline = [
-        'dnf' if shutil.which('dnf') else 'yum',
+        'dnf5' if shutil.which('dnf5') else ('dnf' if shutil.which('dnf') else 'yum'),
```
behrmann, line 162: If we call `shutil.which` we can directly use the result, since it's `None` if what we're looking for doesn't exist, which makes this a bit shorter.
```suggestion
        shutil.which("dnf5") or shutil.which("dnf"') or "yum"
```

behrmann (approved):
LGTM, though I agree that the fallback to `yum` seems a bit like a blast from the past nowadays.

mkosi/distributions/fedora.py
keszybz: Updated like that.

keszybz:
> Let's drop the yum fallback while we're at it.

I tried to figure out if `dnf` command would be available in all cases, and I'm not sure. E.g. on RHEL7 `yum` still seems to be the primary name. So I'd prefer to leave it for now, unless somebody who knows this stuff can confirm that we don't need the fallback.

DaanDeMeyer:
bubblewrap on EPEL 7 is too old for mkosi to work properly, but in the large picture this is irrelevant, let's merge once CI is green so we can get the dnf5 speedups.

DaanDeMeyer (approved):

PR #1442, opened 2023-04-15
Get rid of debootstrap dependency

DaanDeMeyer:
Let's get rid of our dependency on debootstrap by replicating its
core functionality ourselves. To make sure the necessary tools for
maintainer scripts to run are available in the chroot, we have to
extract the essential debs manually first before installing all the
essential debs.

Additionally, we use the keyring from the host now instead of the one
from the chroot, similar to how we do it for other distros.

This also allows us to get rid of our skeleton tree hack we added for
debootstrap.


mkosi/distributions/debian.py
```diff
@@ -158,6 +169,8 @@ def setup_apt(state: MkosiState, repos: Sequence[str]) -> None:
     config = state.workspace / "apt/apt.conf"
     debarch = DEBIAN_ARCHITECTURES[state.config.architecture]
 
+    # Before usrmerge is installed, Debian and Ubuntu systems are still split-usr to we make sure to add /bin
```
bluca, line 172: this is not true, debootstrap sets up the new root already in the right format

mkosi/distributions/debian.py
```diff
@@ -52,32 +50,55 @@ def repositories(state: MkosiState) -> list[str]:
...
+        # impossible to use apt directly to bootstrap a Debian chroot since dpkg will try to run a maintainer
+        # script which depends on some basic tool to be available in the chroot from a deb which hasn't been
+        # unpacked yet, causing the script to fail. To avoid these issues, we have to extract all the
+        # essential debs first, and only then run the maintainer scripts for them.
+
+        # First, download the essential debs. Add usrmerge explicitly because it's not marked Essential on
+        # Ubuntu.
+        setup_apt(state, cls.repositories(state))
+        invoke_apt(state, "update")
+        invoke_apt(state, "install", ["--download-only", "?essential", "usrmerge"])
```
bluca, line 64: Don't install usrmerge, that does a live conversion which is complex and has tons of dependencies. Instead, install usr-is-merged if it is available

bluca:
I'm not sure I like the idea of not using debootstrap, it does complicated crap that I'm pretty sure we don't want to have to reinvent and maintain here forever, what's the advantage?

DaanDeMeyer:
> I'm not sure I like the idea of not using debootstrap, it does complicated crap that I'm pretty sure we don't want to have to reinvent and maintain here forever, what's the advantage?

Much smaller images by default, since we have much larger control over what's installed. Also, we get to use apt's dependency resolution and caching for everything, we don't have to configure both tools to do the right thing, can use more than one mirror which debootstrap doesn't support, it's faster, ... .

The core limitation with debootstrap is that it's built around apt not being available on the host system, which means it has to (usually poorly) re-implement apt. Luckily, we're very much OK with running apt on the host system, and because of that, we're mostly hindered by debootstrap poorly re-implementing apt. We can get much more reliable behavior by just using apt instead of debootstrap.

The only things we have to take care of is setting up merged usr and extracting the essential debs so maintainer scripts can run properly. After that, we delegate everything to apt.

bluca:
I tried the branch, something is getting lost along the way as there are no apt repos configured anymore

mkosi/distributions/debian.py
```diff
@@ -28,63 +26,97 @@ def kernel_image(name: str, architecture: str) -> Path:
...
+        # essential debs first, and only then run the maintainer scripts for them.
+
+        # First, we set up merged usr.
+        mergedusrextra = {
+            "mipsel"  : ["lib32", "lib64"],
+            "ppc64"   : ["lib32", "lib64"],
+            "ppc64el" : ["lib64"],
+            "s390x"   : ["lib32"],
+            "i386"    : ["lib64", "libx32"],
+            "amd64"   : ["lib32", "lib64", "libx32"],
```
bluca, line 67: There's a bunch more architectures that need to be taken care of, even just restricting to EFI stuff there's riscv and loongarch that should really be handled

DaanDeMeyer: I copied these from the debootstrap function that sets up merged usr. It doesn't have any special handling for risc and loongarch, see https://salsa.debian.org/installer-team/debootstrap/-/blob/master/functions#L1369

bluca: loongarch is there?

```
	    loongarch64*)
			link_dir="lib32 lib64" ;;

```

DaanDeMeyer:
> I tried the branch, something is getting lost along the way as there are no apt repos configured anymore

So I still really think we shouldn't configure repos by default. We don't do it for any other distro, and to keep things consistent we shouldn't do it for Debian/Ubuntu either.

mkosi is an image builder, and I'm sure we should focus on image based workflows, including image based updates, for which configured package manager repositories in the image aren't needed.

With these changes, we don't even install apt by default anymore in the produced Ubuntu and Debian images (which finally makes them consistent with the other distros where we don't install a package manager either), so putting repositories in the image just doesn't make sense anymore if you ask me.

bluca:
IIRC the others can initialize themselves, apt cannot. It's not good UX, and it's not how other image builders work so the result is surprising. Maybe add them only if `apt` is passed to the package list? That way for small images without apt there's no such config file

DaanDeMeyer:
> IIRC the others can initialize themselves, apt cannot. It's not good UX, and it's not how other image builders work so the result is surprising. Maybe add them only if apt is passed to the package list? That way for small images without apt there's no such config file

Can we please make an effort to get a package in Debian that installs the repositories then? e.g. a package debian-release that writes to /etc/apt/sources.list.d/xxx.list what we write in mkosi. We can do it in mkosi for now but we should have a plan to get rid of it.

DaanDeMeyer:
Reworked to add the repositories if apt was installed.

DaanDeMeyer:
Added some NEWS updates

bluca:
> > IIRC the others can initialize themselves, apt cannot. It's not good UX, and it's not how other image builders work so the result is surprising. Maybe add them only if apt is passed to the package list? That way for small images without apt there's no such config file
> 
> Can we please make an effort to get a package in Debian that installs the repositories then? e.g. a package debian-release that writes to /etc/apt/sources.list.d/xxx.list what we write in mkosi. We can do it in mkosi for now but we should have a plan to get rid of it.

I don't think that's possible, as everything is uploaded to unstable, then migrates to testing, and eventually is promoted to stable, same identical binary package, so you cannot have a fixed configuration for one suite

mkosi/distributions/debian.py
```diff
@@ -28,67 +26,101 @@ def kernel_image(name: str, architecture: str) -> Path:
...
+
+        updates = f"deb {state.config.mirror} {state.config.release}-updates {repos}"
+
+        # Security updates repos are never mirrored
+        if state.config.release in ("stretch", "buster"):
+            security = f"deb http://security.debian.org/debian-security/ {state.config.release}/updates {repos}"
         else:
-            # Add a single local offline repository, and then remove it after apt has ran
-            state.root.joinpath("etc/apt/sources.list.d/mirror.list").write_text(f"deb [trusted=yes] {state.config.local_mirror} {state.config.release} main\n")
+            security = f"deb https://security.debian.org/debian-security {state.config.release}-security {repos}"
```
bluca, line 47:
```
root@image:~# apt update
Err:1 https://security.debian.org/debian-security stable-security InRelease
  Certificate verification failed: The certificate is NOT trusted. The certificate issuer is unknown.  Could not handshake: Error in the certificate verification. [IP: 2a04:4e42:600::644 443]
Get:2 http://deb.debian.org/debian stable InRelease [116 kB]
Get:3 http://deb.debian.org/debian stable-updates InRelease [44.1 kB]
Get:4 http://deb.debian.org/debian stable/main amd64 Packages [8183 kB]
Get:5 http://deb.debian.org/debian stable/main Translation-en [6240 kB]
Get:6 http://deb.debian.org/debian stable-updates/main amd64 Packages [14.6 kB]
Get:7 http://deb.debian.org/debian stable-updates/main Translation-en [7929 B]
Fetched 14.6 MB in 2s (7310 kB/s)                               
Reading package lists... Done
Building dependency tree... Done
All packages are up to date.
W: https://security.debian.org/debian-security/dists/stable-security/InRelease: No system certificates available. Try installing ca-certificates.
W: Failed to fetch https://security.debian.org/debian-security/dists/stable-security/InRelease  Certificate verification failed: The certificate is NOT trusted. The certificate issuer is unknown.  Could not handshake: Error in the certificate verification. [IP: 2a04:4e42:600::644 443]
W: Some index files failed to download. They have been ignored, or old ones used instead.
```

Given https is used here, it also needs to add ca-certificates to the package list, otherwise it will not work

DaanDeMeyer: Shouldn't we use the cert store from the host? I wonder if we're missing some apt config

DaanDeMeyer: I just switched to http since we use http for the default debian and ubuntu mirrors as well.

mkosi/__init__.py
```diff
@@ -537,6 +532,25 @@ def install_boot_loader(state: MkosiState) -> None:
...
+            t = state.root
+            if target:
+                t = state.root / target.relative_to("/")
+
+            t.mkdir(mode=0o755, parents=True, exist_ok=True)
+            if source.is_dir():
+                copy_path(source, t, preserve_owner=False)
+            else:
+                # unpack_archive() groks Paths, but mypy doesn't know this.
+                # Pretend that tree is a str.
```
keszybz, line 550: That comment was left behind in 0e8b39a1546f12a99c79ed1dc8cfc72a427d6173. I doesn't make sense now.

mkosi/distributions/debian.py
```diff
@@ -52,36 +50,73 @@ def repositories(state: MkosiState, local: bool = True) -> list[str]:
...
+            "ppc64el"     : ["lib64"],
+            "s390x"       : ["lib32"],
+            "sparc"       : ["lib64"],
+            "sparc64"     : ["lib32", "lib64"],
+            "x32"         : ["lib32", "lib64", "libx32"],
+        }
+
+        state.root.joinpath("usr").mkdir(mode=0o755)
+
+        for d in ["bin", "sbin", "lib"] + mergedusrextra.get(DEBIAN_ARCHITECTURES[state.config.architecture], []):
```
keszybz, line 80: It'd be nice to split the initialization out and keep the loop simple:
```python
subdirs = ["bin", "sbin", "lib",
           *mergedusrextra.get(DEBIAN_ARCHITECTURES[state.config.architecture], [])]
for d in subdirs:
   ...
```

keszybz (commented):
This is a bit … involved, but not too horrible.

keszybz:
> > > IIRC the others can initialize themselves, apt cannot. It's not good UX, and it's not how other image builders work so the result is surprising. Maybe add them only if apt is passed to the package list? That way for small images without apt there's no such config file
> > 
> > 
> > Can we please make an effort to get a package in Debian that installs the repositories then? e.g. a package debian-release that writes to /etc/apt/sources.list.d/xxx.list what we write in mkosi. We can do it in mkosi for now but we should have a plan to get rid of it.
> 
> I don't think that's possible, as everything is uploaded to unstable, then migrates to testing, and eventually is promoted to stable, same identical binary package, so you cannot have a fixed configuration for one suite

This shouldn't be too hard. Have some concept of "distro version" and insert it into a template?
Not being able to install *config* as a static package seems so 90's.

bluca:
> > > > IIRC the others can initialize themselves, apt cannot. It's not good UX, and it's not how other image builders work so the result is surprising. Maybe add them only if apt is passed to the package list? That way for small images without apt there's no such config file
> > > 
> > > 
> > > Can we please make an effort to get a package in Debian that installs the repositories then? e.g. a package debian-release that writes to /etc/apt/sources.list.d/xxx.list what we write in mkosi. We can do it in mkosi for now but we should have a plan to get rid of it.
> > 
> > 
> > I don't think that's possible, as everything is uploaded to unstable, then migrates to testing, and eventually is promoted to stable, same identical binary package, so you cannot have a fixed configuration for one suite
> 
> This shouldn't be too hard. Have some concept of "distro version" and insert it into a template? Not being able to install _config_ as a static package seems so 90's.

Nope, the only official way to differentiate between testing and unstable _is_ the apt repositories configuration. Without it, you cannot tell which is which, hence you cannot install the right one.

keszybz:
`W: No sandbox user '_apt' on the system, can not drop privileges`
— is this something to worry about? Maybe we should provide the user to get rid of the warning even if it isn't strictly necessary.

mkosi/distributions/debian.py
```diff
@@ -52,36 +50,73 @@ def repositories(state: MkosiState, local: bool = True) -> list[str]:
...
-        # Let's use --no-check-valid-until only if debootstrap knows it
-        if debootstrap_knows_arg("--no-check-valid-until"):
-            cmdline += ["--no-check-valid-until"]
+            invoke_apt(state, "install", [
+                "-oDebug::pkgDpkgPm=1",
+                f"-oAPT::Keep-Fds::={f.fileno()}",
+                f"-oDPkg::Tools::options::'cat >&$fd'::InfoFD={f.fileno()}",
+                f"-oDpkg::Pre-Install-Pkgs::=cat >&{f.fileno()}",
+                "?essential", "?name(usr-is-merged)",
+            ], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
```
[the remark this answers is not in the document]
DaanDeMeyer: Yeah, this is rather ugly, it works great though.

DaanDeMeyer:
> W: No sandbox user '_apt' on the system, can not drop privileges
> — is this something to worry about? Maybe we should provide the user to get rid of the warning even if it isn't strictly necessary.

The thing is, apt wants this user on the host system, so this would be something to fix in the packages providing apt. Instead, I modified the apt config to use `root` as the sandbox user. If you care about this, you can run mkosi unprivileged and in that case, root is already sandboxed as we'll be using a user namespace and it should be fine to use as the sandbox user. If you're running mkosi as root, it's full of holes anyway so using root as the sandbox user shouldn't be an issue either.

keszybz:
> > > > > IIRC the others can initialize themselves, apt cannot. It's not good UX, and it's not how other image builders work so the result is surprising. Maybe add them only if apt is passed to the package list? That way for small images without apt there's no such config file
> > > > 
> > > > 
> > > > Can we please make an effort to get a package in Debian that installs the repositories then? e.g. a package debian-release that writes to /etc/apt/sources.list.d/xxx.list what we write in mkosi. We can do it in mkosi for now but we should have a plan to get rid of it.
> > > 
> > > 
> > > I don't think that's possible, as everything is uploaded to unstable, then migrates to testing, and eventually is promoted to stable, same identical binary package, so you cannot have a fixed configuration for one suite
> > 
> > 
> > This shouldn't be too hard. Have some concept of "distro version" and insert it into a template? Not being able to install _config_ as a static package seems so 90's.
> 
> Nope, the only official way to differentiate between testing and unstable _is_ the apt repositories configuration. Without it, you cannot tell which is which, hence you cannot install the right one.

Yeah, this is exactly the problem to fix. Add `VERSION_ID=`/`VERSION_CODENAME=` fields to `os-release`, read the file when installing packages, and substitute `{version}` in `apt.sources` as appropriate.

keszybz (approved):
LGTM.

bluca:
> > > > > > IIRC the others can initialize themselves, apt cannot. It's not good UX, and it's not how other image builders work so the result is surprising. Maybe add them only if apt is passed to the package list? That way for small images without apt there's no such config file
> > > > > 
> > > > > 
> > > > > Can we please make an effort to get a package in Debian that installs the repositories then? e.g. a package debian-release that writes to /etc/apt/sources.list.d/xxx.list what we write in mkosi. We can do it in mkosi for now but we should have a plan to get rid of it.
> > > > 
> > > > 
> > > > I don't think that's possible, as everything is uploaded to unstable, then migrates to testing, and eventually is promoted to stable, same identical binary package, so you cannot have a fixed configuration for one suite
> > > 
> > > 
> > > This shouldn't be too hard. Have some concept of "distro version" and insert it into a template? Not being able to install _config_ as a static package seems so 90's.
> > 
> > 
> > Nope, the only official way to differentiate between testing and unstable _is_ the apt repositories configuration. Without it, you cannot tell which is which, hence you cannot install the right one.
> 
> Yeah, this is exactly the problem to fix. Add `VERSION_ID=`/`VERSION_CODENAME=` fields to `os-release`, read the file when installing packages, and substitute `{version}` in `apt.sources` as appropriate.

Nope, the VERSION_ID is `testing/unstable` intentionally https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1021663

bluca (approved):

mkosi/distributions/debian.py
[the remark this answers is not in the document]
behrmann: > Wow.

I second this.

PR #1445, opened 2023-04-17
Use kernel-core instead of kernel in CI for some distros

DaanDeMeyer:
kernel-core is smaller and has fewer dependencies.


DaanDeMeyer:
@keszybz approved the exact same PR in systemd, so I'm assuming it's good to go here as well.

PR #1446, opened 2023-04-17
Do not fail when writing root password and /etc/shadow is missing

keszybz:


DaanDeMeyer:
We really need to switch to systemd-firstboot instead of doing this directly in mkosi.

The only thing we need to figure out is how to allow passing config to systemd-firstboot without having to add options for each firstboot setting to mkosi.

I was thinking of having some concept of "build credentials", which we would write to a directory and then set `CREDENTIALS_DIRECTORY` for all the tools we invoke during the build. Then users could simply set `BuildCredential=passwd.plaintext-password.root` and firstboot would pick that up automatically.

keszybz:
> We really need to switch to systemd-firstboot instead of doing this directly in mkosi.

That would certainly allow removing some special codepaths from mkosi… OTOH, it is a fairly complex solution. I hope we can merge this to make #1442 not crash stupidly.

I added another small fixlet.

mkosi/__init__.py
```diff
@@ -362,7 +362,11 @@ def set_root_pw(line: str) -> str:
                     return ":".join(["root", password] + line.split(":")[2:])
                 return line
 
-            patch_file(state.root / "etc/shadow", set_root_pw)
+            shadow = state.root / "etc/shadow"
+            try:
+                patch_file(shadow, set_root_pw)
+            except FileNotFoundError:
+                shadow.write_text(f"root:{password}:0:0:99999:7:::")
```
behrmann, line 369: Bit late, but `Path.write_text` won't append a trailing newline, so the generated file will be missing that.

keszybz:
→ https://github.com/systemd/mkosi/pull/1447

PR #1448, opened 2023-04-17
Generate the summary in a multiline f-string

behrmann:
This also makes all outut that was conditional on some other settings unconditional, since it was a bit mixed what we always showed and for which we defaulted to some fallback string anyway.

with_docs was conditioned on which distributions supported it, but distribution support for other features has been mixed as well and accompanying settings have been shown regardless.

@keszybz This was what I was thinking of earlier. I think this is a bit more readable and consistent, what do you think?

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -9,7 +9,6 @@ import datetime
 import errno
 import hashlib
 import http.server
-import io
 import itertools
 import json
 import os
@@ -1338,6 +1337,10 @@ def none_to_none(s: Optional[T]) -> Union[T, str]:
     return "none" if s is None else s
 
 
+def none_to_default(s: Optional[T]) -> Union[T, str]:
+    return "default" if s is None else s
+
+
 def path_or_none(
         path: Optional[Path],
         checker: Optional[Callable[[Optional[Path]], None]] = None,
@@ -1349,6 +1352,7 @@ def path_or_none(
         return f'{color_error(path)} ({e.strerror})'
     else:
         return path
+
 
 def line_join_list(
         array: Sequence[PathString],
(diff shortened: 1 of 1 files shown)


DaanDeMeyer (approved):
This looks great but lets wait for @keszybz opinion

keszybz (commented):
The code is very readable. I didn't expect it to look so nice.

But I think we would still want to keep at least some of the conditionalizations. E.g. the whole VALIDATION section is be shown only if the setting make sense for the given output type. For settings like RemoveFiles and RemovePackages, it is not a big issue if they are shown, but they are almost always empty, so it is a nice optimization to show them it not empty. I think it is valuable to keep the output short: it is much easier to grok the output if it it fits on one screen and as many of the lines that are shown are useful as possible.

behrmann:
I made the validation section conditional again, since that was the most tenuous one in terms of showing it by default. Maybe it's just me, but when looking at logs I often parse them by shape and look for lines relative for other lines, so switching lines in the middle on and off makes this a bit harder and a more uniform look makes this easier I think, so maybe this is a compromise. Also, what we show by default and what we don't did strike me as a bit haphazard at times, but for an entire section, even more so at the very end, I think this is very understandable choice. PTAL.

mkosi/__init__.py
```diff
@@ -1363,128 +1367,84 @@ def line_join_source_target_list(array: Sequence[tuple[Path, Optional[Path]]]) -
...
+        Postinstall Script: {path_or_none(config.postinst_script, check_script_input)}
+            Prepare Script: {path_or_none(config.prepare_script, check_script_input)}
+           Finalize Script: {path_or_none(config.finalize_script, check_script_input)}
+        Script Environment: {line_join_list(env)}
+      Scripts with network: {yes_no(config.with_network)}
+           nspawn Settings: {none_to_none(config.nspawn_settings)}
+                  Password: {("(default)" if config.password is None else "(set)")}
+                 Autologin: {yes_no(config.autologin)}
+
+{bold("VALIDATION")}:
```
keszybz, line 1433: VALIDATION was supposed to be conditional, but now it's HOST CONFIGURATION?

keszybz:
I tested this on some repos that I have, and I still get "none" in way too many places. But that was already the case before so a large extent… So I think we can merge this now, and maybe add conditionalization back in later on.

PR #1450, opened 2023-04-17
opensuse: Install repositories into image if zypper is installed

DaanDeMeyer:
Like Debian, OpenSUSE does not provide packages to install the repos into the image so let's adopt the same fix and install the repositories in the image if zypper is installed into the image.

### DIFF
--- a/mkosi/distributions/debian.py
+++ b/mkosi/distributions/debian.py
@@ -131,6 +131,8 @@ class DebianInstaller(DistributionInstaller):
         policyrcd.write_text("#!/bin/sh\nexit 101\n")
         policyrcd.chmod(0o755)
 
+        before = state.root.joinpath("usr/bin/apt").exists()
+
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
Note that if `dnf` is being used in the image, you can use `rpm-repos-openSUSE-Tumbleweed` for openSUSE Tumbleweed, and `rpm-repos-openSUSE-Leap` for Leap.

For `zypper`, Tumbleweed and Leap have the `openSUSE-repos` package.

DaanDeMeyer:
> For zypper, Tumbleweed and Leap have the openSUSE-repos package.

Ahah I had no clue this was a thing, thanks for the link! @fbuihuu this should resolve the issue about the repos not being available in the image so is it OK if I close this PR?

DaanDeMeyer:
Let's close due to lack of response

PR #1452, opened 2023-04-17
Add back --bootable option

DaanDeMeyer:
There are valid scenarios where one might want to install a kernel without wanting a bootable image, so let's add back the --bootable= option as a feature option. It defaults to "auto" which is the current behavior, but can also be explicitly enabled or disabled. If enabled, we'll fail if we can't produce a bootable image. If disabled, we won't generate a bootable image even if all the necessary components are available.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -456,11 +456,17 @@ def run_finalize_script(state: MkosiState) -> None:
 
 
 def install_boot_loader(state: MkosiState) -> None:
-    if state.for_cache:
+    if state.for_cache or state.config.bootable is False:
+        return
+
+    if state.config.output_format == OutputFormat.cpio and state.config.bootable is None:
         return
 
     directory = state.root / "usr/lib/systemd/boot/efi"
-    if not directory.exists():
+    if not directory.exists() or not any(directory.iterdir()):
+        if state.config.bootable is True:
+            die("A bootable image was requested but systemd-boot was not found at "
+                f"{directory.relative_to(state.root)}")
         return
 
     if state.config.secure_boot:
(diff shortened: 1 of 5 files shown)


DaanDeMeyer:
Tested by @Leif Liddy from the Asahi Fedora matrix channel

mkosi/__init__.py
```diff
@@ -1713,8 +1727,14 @@ def invoke_repart(state: MkosiState, skip: Sequence[str] = [], split: bool = Fal
...
             bootdir = state.root.joinpath("boot/EFI/BOOT")
 
-            if bootdir.exists() and any(bootdir.iterdir()) and any(gen_kernel_images(state)):
-                # If we have at least one kernel images and a bootloader, let's generate an ESP partition.
+            # If Bootable=auto and we have at least one UKI and a bootloader, let's generate an ESP partition.
+            add = (state.config.bootable is True or
+                  (state.config.bootable is None and
+                   bootdir.exists() and
+                   any(bootdir.iterdir()) and
+                   any(gen_kernel_images(state))))
```
behrmann, line 1735: The `is True` everywhere is a bit unpythonic and I think the code would look better without, but it would leave the `is False` to differentiate that from `None` as the sentinel value for auto. How about adding a tristate enum
```python
class BoolOrAuto(enum.Enum):
     TRUE = True
     FALSE = False
     AUTO = "auto"

     def __bool__(self):
         return bool(self.value)

     def is_auto(self):
         return self.value == "auto"
```
which would make this
```suggestion
            add = (state.config.bootable or 
                   (state.config.bootable.is_auto() and
                    bootdir.exists() and
                    any(bootdir.iterdir()) and
                    any(gen_kernel_images(state))))
```

behrmann (approved):
Functionalitywise LGTM, I think the something else than `None` as a token value for auto would make this more readable.

mkosi/__init__.py
DaanDeMeyer: I agree but I'll fix that in a follow up PR

PR #1454, opened 2023-04-18
debian: Add debugging for oDebug::pkgDpkgPm=1 command

DaanDeMeyer:
Let's make sure we get an error message should this command fail to run.


DaanDeMeyer:
It should be fine to do this by default, this command is pretty fast

behrmann (approved):

PR #1458, opened 2023-04-18
debian: Simplify cmdline to get the list of packages to install

DaanDeMeyer:
Let's not bother with file descriptor inheritance and just write the list of packages to a temporary file.


DaanDeMeyer:
This fixes the CI error in https://github.com/systemd/systemd/pull/27329 so let's merge.

PR #1459, opened 2023-04-18
Don't make output an absolute path while parsing

DaanDeMeyer:
Fixes a regression from 2a6aa720f59e4f66ba63d311f5fa79c3a3ac0a61.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -203,7 +203,7 @@ def make_path_parser(required: bool) -> Callable[[str], Path]:
     return parse_path
 
 
-def config_make_path_parser(required: bool) -> ConfigParseCallback:
+def config_make_path_parser(required: bool, absolute: bool = True) -> ConfigParseCallback:
     def config_parse_path(dest: str, value: Optional[str], namespace: argparse.Namespace) -> Optional[Path]:
         if dest in namespace:
             return getattr(namespace, dest) # type: ignore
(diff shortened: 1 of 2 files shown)


DaanDeMeyer:
@keszybz Can you verify this fixes your issue?

mkosi/config.py
```diff
@@ -203,15 +203,15 @@ def parse_path(value: str) -> Path:
...
+def config_make_path_parser(required: bool, absolute: bool = True) -> ConfigParseCallback:
     def config_parse_path(dest: str, value: Optional[str], namespace: argparse.Namespace) -> Optional[Path]:
         if dest in namespace:
             return getattr(namespace, dest) # type: ignore
 
         if value and required and not Path(value).exists():
             die(f"{value} does not exist")
 
-        return Path(value).absolute() if value else None
+        return Path(value).absolute() if value and absolute else Path(value) if value else None
```
behrmann, line 214: More vertical, but I think easier to parse.
```suggestion
        if value:
            return Path(value).absolute() if absolute else Path(value)
        return None
```

behrmann (approved):

keszybz:
Yes, this seems to work fine.

DaanDeMeyer:
> Maybe it'd be better to revert https://github.com/systemd/mkosi/commit/2a6aa720f59e4f66ba63d311f5fa79c3a3ac0a61 and instead make the paths passed to bwrap absolute? (Or maybe even fix bwrap?) Absolute paths are not very nice for users, they tend to be long and the prefix is not interesting.

We do the same in systemd though, so I figured I'd follow suite here.

Is this error caused by this PR? This usually indicates systemd-boot isn't installed in the image

keszybz:
Ah, OK. It expects the stub in the image. Please disregard my comment then.

DaanDeMeyer:
@keszybz I opened https://github.com/systemd/mkosi/pull/1465 to show a better error if the stub is not installed.

PR #1461, opened 2023-04-19
Stop copying the cached build overlay

DaanDeMeyer:
The build overlay is not modified during the final build, it's only used to run the build script which cannot touch the build image. Let's enforce this even more by mounting the build overlay read-only when running the script.

With this assurance, we can stop copying the cached build overlay every time and just use it directly since we're certain it cannot be modified anyway.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -290,6 +290,13 @@ def install_build_packages(state: MkosiState, cached: bool) -> None:
     with mount_build_overlay(state):
         state.installer.install_packages(state, state.config.build_packages)
 
+        # Create a few necessary mount points inside the build overlay for later.
+        state.root.joinpath("work").mkdir(mode=0o755)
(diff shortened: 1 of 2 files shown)


mkosi/__init__.py
```diff
@@ -1645,8 +1652,8 @@ def reuse_cache_tree(state: MkosiState) -> bool:
         copy_path(final, state.root)
         acl_toggle_remove(state.config, state.root, state.uid, allow=False)
         if state.config.build_script:
-            copy_path(build, state.build_overlay)
-            acl_toggle_remove(state.config, state.build_overlay, state.uid, allow=False)
+            state.build_overlay.rmdir()
+            state.build_overlay.symlink_to(build, target_is_directory=True)
```
behrmann, line 1656: `target_is_directory=` is a Windows thing, it's ignored on Linux systems.

PR #1462, opened 2023-04-19
Allow booting cpio images in qemu

DaanDeMeyer:
Let's add back some form of direct linux boot by allowing to boot
cpio images. To make this work, either the user has to pass -kernel
or a kernel has to be installed inside the cpio, which we'll copy
out and use as the -kernel argument.

### DIFF
--- a/mkosi/__main__.py
+++ b/mkosi/__main__.py
@@ -38,7 +38,7 @@ def propagate_failed_return() -> Iterator[None]:
         if ARG_DEBUG:
             raise e
 
-        MkosiPrinter.error("Error: mkosi failed because of an exception, rerun mkosi with --debug run to get more information")
+        MkosiPrinter.error(f"Error: {str(e)}, rerun mkosi with --debug run to get more information")
         sys.exit(1)
 
 
--- a/mkosi/backend.py
+++ b/mkosi/backend.py
@@ -89,6 +89,16 @@ class Distribution(enum.Enum):
         return self.name
 
 
+class Compression(str, enum.Enum):
+    none = "none"
+    zst = "zst"
+    xz = "xz"
+    bz2 = "bz2"
+    gz = "gz"
+    lz4 = "lz4"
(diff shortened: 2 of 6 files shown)


mkosi/backend.py
```diff
@@ -89,6 +89,16 @@ def __str__(self) -> str:
...
 
 
+class Compression(str, enum.Enum):
+    none = "none"
+    zst = "zst"
+    xz = "xz"
+    bz2 = "bz2"
+    gz = "gz"
+    lz4 = "lz4"
+    lzma = "lzma"
```
behrmann, line 99: I don't think this being a str subclass is used anywhere
```suggestion
class Compression(enum.Enum):
    none = None
    zst = "zst"
    xz = "xz"
    bz2 = "bz2"
    gz = "gz"
    lz4 = "lz4"
    lzma = "lzma"
    
    def __str__(self):
        return str(self.value).lower()
        
    def __bool__(self):
        return bool(self.value)
```
will produce the same stringifcation, but you can call `Compression(None)` directly, which might be helpful for the argument parser and the bool would make it intuitively usable, because `Compression.none` would be falsey and the others truthy.

behrmann: Sure, no problem.

mkosi/__main__.py
```diff
@@ -38,7 +38,7 @@ def propagate_failed_return() -> Iterator[None]:
         if ARG_DEBUG:
             raise e
 
-        MkosiPrinter.error("Error: mkosi failed because of an exception, rerun mkosi with --debug run to get more information")
+        MkosiPrinter.error(f"Error: {str(e)}, rerun mkosi with --debug run to get more information")
```
keszybz, line 41: Why `str(e)` ?

DaanDeMeyer: Maybe that's unnecessary actually, I'll do a follow up PR

PR #1464, opened 2023-04-19
Update default Fedora release to 38

DaanDeMeyer:


behrmann:
All of Fedora has built, no need to wait.

PR #1465, opened 2023-04-19
Fail with a clear error if sd-stub is not installed in the image

DaanDeMeyer:


mkosi/__init__.py
```diff
@@ -766,11 +766,15 @@ def install_unified_kernel(state: MkosiState, roothash: Optional[str]) -> None:
             # nul terminators in argv so let's communicate the cmdline via a file instead.
             state.workspace.joinpath("cmdline").write_text(f"{' '.join(cmdline).strip()}\x00")
 
+            stub = state.root / f"lib/systemd/boot/efi/linux{EFI_ARCHITECTURES[state.config.architecture]}.efi.stub"
+            if not stub.exists():
+                die(f"sd-stub not found at {stub.relative_to(state.root)}")
```
behrmann, line 771: On error the message would be, e.g.
> sd-stub not found at lib/systemd/boot/efi/linux-x64.efi.stub

how about
```suggestion
                die(f"sd-stub not found at /{stub.relative_to(state.root) in the image}")
```

behrmann (approved):

PR #1468, opened 2023-04-20
Split up --base-image into --base-tree and --overlay

DaanDeMeyer:
--base-tree indicates a base image. It differs from --skeleton-tree
in that use of --base-tree indicates that we've already installed a
distribution and should only install extra packages, not install the
distribution from scratch.

If --overlay is not specified, we just copy all the specified base trees
to the root directory before doing anything else (even before copying the
skeleton trees) and after that we operate as usual.

If --overlay is specified, instead of copying all the base trees to
the root directory, we set up an overlayfs mount with all the base
trees as lowerdirs. After that we operate as usual. The effect is that
all our usual steps will operate on a full view of the image, but the
output will only contain the additions we made on top of the specified
base trees.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -509,12 +509,6 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
   in building or further container import stages.  This option strips
   SELinux context attributes from the resulting tar archive.
 
-`Initrd=`, `--initrd`
-
-: Use user-provided initrd(s). Takes a comma separated list of paths to initrd
-  files. This option may be used multiple times in which case the initrd lists
-  are combined.
-
 ### [Content] Section
 
 `Packages=`, `--package=`, `-p`
@@ -773,6 +767,12 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
   environment when the image is run. If this setting is not used but
   an `mkosi.nspawn` file found in the local directory it is
   automatically used for this purpose.
+
+`Initrd=`, `--initrd`
+
+: Use user-provided initrd(s). Takes a comma separated list of paths to initrd
+  files. This option may be used multiple times in which case the initrd lists
+  are combined.
 
 `BaseImage=`, `--base-image=`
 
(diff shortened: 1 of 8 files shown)


mkosi.md
```diff
@@ -500,6 +500,19 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
   image root, so any `CopyFiles=` source paths in partition definition files will
   be relative to the image root directory.
 
+`Overlay=`, `--overlay`
+
+: When used together with `BaseTrees=`, the output will consist only out of of
```
bluca, line 505: "of of"

mkosi.md
```diff
@@ -500,6 +500,19 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
...
+  changes to the specified base trees. Each base tree is attached as a lower
+  layer in an overlayfs structure, and the output becomes the upper layer,
+  initially empty. Thus files that are not modified compared to the base trees
+  will not be present in the final output.
+
+: This option may be used to create systemd "system extensions" or
+  portable services. See
+  https://systemd.io/PORTABLE_SERVICES/#extension-images for more
```
bluca, line 513: better to link https://uapi-group.org/specifications/specs/extension_image/

mkosi.md
```diff
@@ -570,6 +583,22 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
   way is mounted into both the development and the final image while
   the package manager is running.
 
+`BaseTrees=`, `--base-tree=`
+
+: Takes a colon separated pair of directories to use as base images. When
```
bluca, line 588: Only directories? Not images?

DaanDeMeyer: I'd love for privileged and unprivileged mkosi to support the same set of features and disk image support here would break that. I figure you can always pack up your disk image as a tar as well and build your extension images using that as the base tree. Do you absolutely need to support building extensions against disk images?

bluca: It's a bit of a pain to do, why not support both, and simply refuse if getuid != 0 if an image is passed?

DaanDeMeyer: Added back disk image support.

bluca (approved):

behrmann (approved):
Don't the road we should probably add a `mkosi.base` to automatically be used as base tree or image

keszybz (commented):
Late review: LGTM. I didn't test it with initrds, but I don't see anything that'd make it now work.

PR #1469, opened 2023-04-20
Rework exception handling yet again

keszybz:
e25d746f9d7668edc5134cde4ec381c8b2da0136 removed printing of the exception backtrace, except if --debug is used. (Later
afb709356fed78a1b6994e72d651e92f354be333 added back printing of the exception message, and b89db00359ad28e90a1923f3b483b63e7f47829a changed it to print the class name, but the general approach of not printing the backtrace remained.) This is not very useful.

My motivating use case was an error introduced in one of my patches. A ValueError was thrown by subprocess.run(). This would be trivial to fix except that without the backtrace I had no idea what was going on.

We shouldn't print an exception if the failure is "expected". When we call die(), then the code was written to expect a failure at this point and we print an error message and we don't need to print the backtrace. So far good. But if a different exception type is thrown, then most likely we made a programmatic error and called something with wrong arguments, or the environment unexpectedly prevents us from doing something that shouldn't fail. In that case, just the exception (class or string or whatever) is not enough to diagnose the issue, because exception names and messages are generic, and without a backtrace won't know where exactly the failure happened. Telling the user to rerun with --debug is not a good solution, because they might not be able to recreate the failure.

The code is reworked to use MkosiError for our own exceptions and we only print the error message (unless --debug is used). Other exceptions print the backtrace, on the assumption that this is a failure that should be reported to developers and fixed. (Or if a developer is using mkosi, they fix it themselves ;)). If it turns out that mkosi fails in a way that prints a backtrace in normal use, then we can always add special handling for that case. This way, we won't get backtraces in normal use, but we will get them if there's something to fix in the code.

'raise MkosiError from e' is used so that we get chained backtraces.

### DIFF
--- a/mkosi/__main__.py
+++ b/mkosi/__main__.py
@@ -10,7 +10,7 @@ from collections.abc import Iterator
 from mkosi import load_args, run_verb
 from mkosi.config import MkosiConfigParser
 from mkosi.log import ARG_DEBUG, MkosiError, MkosiPrinter, die
-from mkosi.run import excepthook
+from mkosi.run import excepthook, RemoteException
 
 
 @contextlib.contextmanager
@@ -39,9 +39,15 @@ def propagate_failed_return() -> Iterator[None]:
             raise e
         sys.exit(1)
     except Exception as e:
-        if not ARG_DEBUG:
+        remote_mkosi_error = isinstance(e, RemoteException) and isinstance(e.exception, MkosiError)
+
+        if not remote_mkosi_error and not ARG_DEBUG:
             MkosiPrinter.info(f"Hint: mkosi failed because of an internal exception {e.__class__.__name__}, "
                               "rerun mkosi with --debug to get more information")
+
+        if remote_mkosi_error:
+            sys.exit(1)
+
         raise e
 
 
--- a/mkosi/__main__.py
+++ b/mkosi/__main__.py
@@ -9,7 +9,7 @@ from collections.abc import Iterator
 
 from mkosi import load_args, run_verb
 from mkosi.config import MkosiConfigParser
-from mkosi.log import ARG_DEBUG, MkosiPrinter, die
+from mkosi.log import ARG_DEBUG, MkosiError, MkosiPrinter, die
 from mkosi.run import excepthook
 
 
@@ -34,14 +34,15 @@ def propagate_failed_return() -> Iterator[None]:
 
         # We always log when subprocess.CalledProcessError is raised, so we don't log again here.
         sys.exit(e.returncode)
-    except Exception as e:
+    except MkosiError as e:
         if ARG_DEBUG:
             raise e
-        elif not isinstance(e, RuntimeError):
-            # RuntimeError is used to wrap generic errors, and the message that was printed should be enough.
+        sys.exit(1)
+    except Exception as e:
+        if not ARG_DEBUG:
             MkosiPrinter.info(f"Hint: mkosi failed because of an internal exception {e.__class__.__name__}, "
                               "rerun mkosi with --debug to get more information")
-        sys.exit(1)
+        raise e
 
 
 @propagate_failed_return()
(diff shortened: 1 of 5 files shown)


behrmann:
The test failure is because pytest is expecting `RuntimeError` but is getting `MkosiError` in `tests/test_parse_load_args.py` and is getting `ValueError` but getting `MkosiError` in tests/test_backend.py`.

DaanDeMeyer:
The issue with this approach is that we don't want to change the type of the exception we throw. If run() fails, we use die() to log about it, but we still want to raise subprocess.CalledProcessError, not MkosiError().

Instead of changing the exception type, we should attach some metadata to the exception to indicate that we've already logged about it, without changing the exception type, but this is not trivial because the exception has to traverse a process boundary (see fork_and_wait()) and I'm not sure if arbitrary fields on the exception will be serialized across that boundary.

The pragmatic solution is to stop using die() in run() and friends, and simply use MkosiPrinter.error() there to log and then reraise the original exception. For the rest, it doesn't matter too much.

behrmann:
> The issue with this approach is that we don't want to change the type of the exception we throw. If run() fails, we use die() to log about it, but we still want to raise subprocess.CalledProcessError, not MkosiError().

Why actually? With the `raise MkosiException from initial_exception` the initial error is still in the stack trace and available to work with via the `__cause__` attribute on the raised `MkosiException`.

DaanDeMeyer:
> Why actually? With the raise MkosiException from initial_exception the initial error is still in the stack trace and available to work with via the __cause__ attribute on the raised MkosiException.

Yeah but I can't `except <original_exception>` anymore.

behrmann:
That's true, but I think it could be used as the metadata you wanted, since we could `catch MkosiException`, handle it and in the worst case reraise `Mkosi.__cause__`. This way expected and unexpected errors could be discerned.

keszybz:
Python3.11 has a new "notes" feature:
https://docs.python.org/3/library/exceptions.html#BaseException.add_note
https://docs.python.org/3/tutorial/errors.html#enriching-exceptions-with-notes
That could be useful for us, but we can't use that yet.

Updated. I think the tests should pass now.

keszybz:
> The issue with this approach is that we don't want to change the type of the exception we throw. If run() fails, we use die() to log about it, but we still want to raise subprocess.CalledProcessError, not MkosiError().

Please explain. I thought that if we raise an exception via `die()`, we just expect to terminate the program. So the exception type doesn't matter until the very end where we do some handling in `__main__.py`.

DaanDeMeyer:
> Please explain. I thought that if we raise an exception via die(), we just expect to terminate the program. So the exception type > doesn't matter until the very end where we do some handling in __main__.py.

That's more or less the idea, but it doesn't always apply e.g. for stuff like run().

Maybe we should throw the whole thing around again and do the logging in propagate_failed_return() instead of in die()? I originally changed it to logging when we call die() because we had tons of complete_step() stuff that would log after the error was raised, making it hard to figure out exactly where the error happened. But we could change complete_step() to not log when the step is complete (that happens implicitly by just logging the next step that we're executing) and if we do that I figure we could do the logging in propagate_failed_return() and have more terse logging for both subprocess.CalledProcessError and MkosiError()

mkosi/__main__.py
```diff
@@ -9,8 +9,8 @@
 
 from mkosi import load_args, run_verb
 from mkosi.config import MkosiConfigParser
-from mkosi.log import ARG_DEBUG, MkosiPrinter, die
-from mkosi.run import excepthook
+from mkosi.log import ARG_DEBUG, MkosiError, MkosiPrinter, die
+from mkosi.run import excepthook, RemoteException
```
behrmann, line 13:
```suggestion
from mkosi.run import RemoteException, excepthook
```
This is what isort is complaining about in CI.

DaanDeMeyer:
I've posted an alternative solution in #1491.

keszybz:
I think the second commit might still be useful in some form. Let's wait until https://github.com/systemd/mkosi/pull/1491 is merged.

keszybz:
OK, this seems not needed.

PR #1470, opened 2023-04-20
Use a console script entry point

gdiscry:
The use of a custom script to launch mkosi has several shortcomings:

1. It is tied to setuptools and has no equivalent in the standard project metadata (as in pyproject.toml).
2. It translates `$PKEXEC_UID` into various `$SUDO_`* variables. That translation is lost if mkosi is launched directly as a module with `pkexec python3 -m mkosi`.
3. It has to muck around with `$PYTHONPATH` to launch mkosi directly from its source code.
4. And it also has to find the correct interpreter and maybe muck around with `$PYTHONPATH` to launch mkosi from its installation, which may be on the system, in the user's sitedir or in a virtualenv, as a standard or an editable installation.

Point 2 is not needed anymore now that `mkosi.backend.current_user()` is used everywhere (see #1472).

Point 1 and 4 are better supported by using a `"console_scripts"` entry point when installed. However, it is not anymore possible to run a user-installed mkosi with sudo/pkexec.

Point 3 is also better supported by making the project directory executable as a script thanks to a `__main__.py` script next to the package, without any need to muck around with `$PYTHONPATH`. That way, mkosi can be launched directly from its repository with:

- `[sudo|pkexec] python3 -m mkosi` inside the project directory
- `[sudo|pkexec] python3 <PROJECT_PATH>` from anywhere
- `[sudo|pkexec] <PROJECT_PATH>/bin/mkosi` from anywhere as the script is kept for compatibility, but only for launching mkosi from its repository as it is not installed anymore.

---

This is sort of a prerequisite/spin-off of #1343 focusing only on replacing the `bin/mkosi` script with an entry point to `mkosi.__main__:main` and fixing some issues along the way.

Unlike #1343, the `bin/mkosi` script is kept for compatibility reasons, but only for running directly from the repo as it is not installed anymore. It is also possible to run mkosi (with/without sudo/pkexec) from the repo with:

- `python3 <PROJECT_PATH>` from anywhere (which is used by `bin/mkosi`)
- `python3 -m mkosi` from inside the repo
- `PYTHONPATH=<PROJECT_PATH> python3 -m mkosi` from anywhere

However, like #1343, we lose the ability to `pip install --user [--editable]` mkosi and run it with sudo or pkexec, but this is an inherent limitation of Python tools using an entry point. It is possible to run a user-installed mkosi with sudo by manually setting up a virtualenv or by using pipx.

It's currently a draft because it depends on #1471 and #1472, which were split from this PR.


gdiscry:
Unit tests failed when trying to run a user-installed mkosi with sudo, because the user site directory is obviously not in `sys.path` anymore when Python is run by root.

A workaround seems impossible as the failure occurs before any of the code is run, so I've removed those tests.

__main__.py
behrmann, line 1: I don't think this is a good idea. Having a dunder file in the project root is seems smelly to me and is something I fear we could stumble over down the line. It is definitely something users do not expect.

bin/mkosi
behrmann, line 1: With a console script I think this shim should be removed. There's so many ways to install Python projects, one can set `PYTHONPATH` and `mkosi` can be executed as a module. This is shim is something I'd like to get rid of, as seen in #1343.

behrmann (commented):
Just looked at the console script part for now. I think this would be easier to review as two PRs, since most of the changes look orthogonal to the console script stuff, as is the case in #1343, where it's only included because in that context it was (seemed?) impossible to ship this extra script.

> A workaround seems impossible as the failure occurs before any of the code is run, so I've removed those tests.

which tests need removal when changing this can be seen in #1343

gdiscry:
> Just looked at the console script part for now. I think this would be easier to review as two PRs, since most of the changes look orthogonal to the console script stuff, as is the case in #1343, where it's only included because in that context it was (seemed?) impossible to ship this extra script.

Yeah, this could be split in several PRs and progressively merged, but (nearly) each commit depends on the previous. But I get your point that while the last commit depends on the previous ones and is quite small, it's also the most disruptive.

bin/mkosi
gdiscry: The shim is not installed anymore and is kept only for compatibility. I'd also like to remove it, but I wanted to show that it's still possible to have an executable for running mkosi from its repo.

__main__.py
gdiscry: It also feels icky to me, but it's also the least complex way to run mkosi from its repo without being in it.

It's also possible to use something like this in `mkosi/__main__.py`:

```python
if not __package__:
    # Run as a script
    sys.path[0] = os.path.dirname(sys.path[0])

from mkosi import ...
```

and run mkosi from its repo with `python3 <PROJECT_PATH>/mkosi`.

I'd rather not include this file, there are indeed other (more complicated) ways to run mkosi from its repo but the use case seemed important in the conversation around #1343.

behrmann: The reason for this shim was keep several workflows possible in the context of mkosi needing to run as root. With the ability to use mkosi as a regular user, this has become far less pressing. With the options
1. running `mkosi` directly from a checkout via `python -m mkosi`
2. installing (maybe in a venv or with pipx) either as regular user or the root user or systemdwide and using the console script or `python -m mkosi`
3. doing an editable install (something that blocked using `pyproject.toml` for a while)

I think there's enough options to go around to not add more workarounds.

gdiscry:
Personally, I would drop `bin/mkosi` and `__main__.py` (but not `mkosi/__main__.py`).

@DaanDeMeyer You asked about running mkosi with sudo from its sources while being located elsewhere in https://github.com/systemd/mkosi/issues/1343#issuecomment-1456218235, are you ok with `sudo PYTHONPATH=<MKOSI_SOURCES> python3 -m mkosi`?

gdiscry:
Well I pulled the trigger and completely removed `bin/mkosi`and dropped `__main__.py` 😉.

DaanDeMeyer:
I want to be able to put a symlink to mkosi in /usr/local/bin and have it work when I run `sudo mkosi`. If that works its fine.

behrmann:
> I want to be able to put a symlink to mkosi in /usr/local/bin and have it work when I run sudo mkosi. If that works its fine.

Not like this, but I really don't think we should carry this kludge for this single use case. There are two options around this:
1. Install mkosi as editable install globally (icky)
2. Drop this shim into `/usr/local/bin`
```bash
#!/usr/bin/env bash
export PYTHONPATH=~yourusername/path/to/mkosi/repo
python3 -B -m mkosi
```

DaanDeMeyer:
> > I want to be able to put a symlink to mkosi in /usr/local/bin and have it work when I run sudo mkosi. If that works its fine.
> 
> Not like this, but I really don't think we should carry this kludge for this single use case. There are two options around this:
> 
> 1. Install mkosi as editable install globally (icky)
> 2. Drop this shim into `/usr/local/bin`
> 
> ```shell
> #!/usr/bin/env bash
> export PYTHONPATH=~yourusername/path/to/mkosi/repo
> python3 -B -m mkosi
> ```

That's good enough for me

gdiscry:
The needed PR have been merged!

DaanDeMeyer:
I think we should still keep the bin/mkosi shim, but in a simpler form that users can just symlink into their PATH.

behrmann (approved):
I don't see much added value in the shim, otherwise LGTM.

PR #1472, opened 2023-04-21
Path expansion

gdiscry:
When the `extra_search_paths` configuration was introduced, it would perform path expansions with the environment variables (using $) and the home directories (starting with ~). Home directories would not be expanded when mkosi was run with sudo or pkexec. Instead, `$SUDO_HOME` could be used but only when run with sudo or pkexec.

However, those expansions were not applied to other paths in the configuration and were broken with the rewrite of the configuration loading, as the parser would check if the paths exist before expanding.

Those expansions are now handled directly in the parser and are performed before validation. Environment variables expansion is enabled for all paths and home directory expansion is enabled for all paths on
the host.

Furthermore, home directory expansion is always available, with `~` expanding to the user's home directory when sudo or pkexec are used (and not `/root`), replacing the need for `$SUDO_HOME`.

Split from #1470.

### DIFF
--- a/mkosi/backend.py
+++ b/mkosi/backend.py
@@ -471,10 +471,49 @@ def flatten(lists: Iterable[Iterable[T]]) -> list[T]:
     return list(itertools.chain.from_iterable(lists))
 
 
-def current_user_uid_gid() -> tuple[int, int]:
-    uid = int(os.getenv("SUDO_UID") or os.getenv("PKEXEC_UID") or os.getuid())
-    gid = pwd.getpwuid(uid).pw_gid
-    return uid, gid
+@dataclasses.dataclass
+class InvokingUser:
+    _pw: Optional[pwd.struct_passwd] = None
+
+    @classmethod
+    def for_uid(cls, uid: int) -> "InvokingUser":
+        return cls(pwd.getpwuid(uid))
+
+    @property
+    def uid(self) -> int:
+        if self._pw is not None:
+            return self._pw.pw_uid
+        return os.getuid()
+
+    @property
+    def gid(self) -> int:
+        if self._pw is not None:
+            return self._pw.pw_gid
+        return os.getgid()
+
+    @property
+    def name(self) -> str:
+        if self._pw is not None:
+            return self._pw.pw_name
+        return os.getlogin()
+
+    @property
+    def home(self) -> Path:
+        if self._pw is not None:
+            return Path(self._pw.pw_dir)
+        return Path.home()
+
+    def is_running_user(self) -> bool:
+        if self._pw is not None:
+            return self._pw.pw_uid == os.getuid()
+        return True
+
+
+def current_user() -> InvokingUser:
+    uid = os.getenv("SUDO_UID") or os.getenv("PKEXEC_UID")
+    if uid:
+        return InvokingUser.for_uid(int(uid))
+    return InvokingUser()
 
 
 @contextlib.contextmanager
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -3,6 +3,7 @@ import configparser
 import dataclasses
 import enum
 import fnmatch
+import functools
 import os
 import platform
 import sys
@@ -44,13 +45,28 @@ def parse_boolean(s: str) -> bool:
     die(f"Invalid boolean literal: {s!r}")
 
 
+def parse_path(value: str, *, required: bool, absolute: bool = True) -> Path:
+    path = Path(value)
+
+    if required and not path.exists():
+        die(f"{value} does not exist")
+
+    if absolute:
+        path = path.absolute()
+
+    return path
+
+
 def parse_source_target_paths(value: str) -> tuple[Path, Optional[Path]]:
     src, _, target = value.partition(':')
-    if not Path(src).exists():
-        die(f"{src} does not exist")
-    if target and not Path(target).is_absolute():
-        die("Target path must be absolute")
-    return Path(src).absolute(), Path(target) if target else None
+    src_path = parse_path(src, required=True)
+    if target:
+        target_path = parse_path(target, required=False, absolute=False)
+        if not target_path.is_absolute():
+            die("Target path must be absolute")
+    else:
+        target_path = None
+    return src_path, target_path
 
 
 def config_parse_string(dest: str, value: Optional[str], namespace: argparse.Namespace) -> Optional[str]:
@@ -69,12 +85,12 @@ def config_parse_script(dest: str, value: Optional[str], namespace: argparse.Nam
         return getattr(namespace, dest) # type: ignore
 
     if value:
-        if not Path(value).exists():
-            die(f"{value} does not exist")
-        if not os.access(value, os.X_OK):
+        path = parse_path(value, required=True)
+        if not os.access(path, os.X_OK):
             die(f"{value} is not executable")
+        return path
 
-    return Path(value).absolute() if value else None
+    return None
 
 
 def config_parse_boolean(dest: str, value: Optional[str], namespace: argparse.Namespace) -> bool:
@@ -197,26 +213,25 @@ def config_make_list_parser(delimiter: str, parse: Callable[[str], Any] = str) -
     return config_parse_list
 
 
-def make_path_parser(required: bool) -> Callable[[str], Path]:
-    def parse_path(value: str) -> Path:
-        if required and not Path(value).exists():
-            die(f"{value} does not exist")
-
-        return Path(value).absolute()
-
-    return parse_path
+def make_path_parser(*, required: bool, absolute: bool = True) -> Callable[[str], Path]:
+    return functools.partial(
+        parse_path,
+        required=required,
+        absolute=absolute,
+    )
 
 
-def config_make_path_parser(required: bool, absolute: bool = True) -> ConfigParseCallback:
+def config_make_path_parser(*, required: bool, absolute: bool = True) -> ConfigParseCallback:
     def config_parse_path(dest: str, value: Optional[str], namespace: argparse.Namespace) -> Optional[Path]:
         if dest in namespace:
             return getattr(namespace, dest) # type: ignore
 
-        if value and required and not Path(value).exists():
-            die(f"{value} does not exist")
-
         if value:
-            return Path(value).absolute() if absolute else Path(value)
+            return parse_path(
+                value,
+                required=required,
+                absolute=absolute,
+            )
 
         return None
 
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -4,7 +4,7 @@ import dataclasses
 import enum
 import fnmatch
 import functools
-import os
+import os.path
 import platform
 import sys
 import textwrap
@@ -19,6 +19,7 @@ from mkosi.backend import (
     OutputFormat,
     Verb,
     chdir,
+    current_user,
     detect_distribution,
     flatten,
 )
@@ -45,8 +46,18 @@ def parse_boolean(s: str) -> bool:
     die(f"Invalid boolean literal: {s!r}")
 
 
-def parse_path(value: str, *, required: bool, absolute: bool = True) -> Path:
+def parse_path(value: str, *, required: bool, absolute: bool = True, expanduser: bool = True, expandvars: bool = True) -> Path:
+    if expandvars:
+        value = os.path.expandvars(value)
+
     path = Path(value)
+
+    if expanduser:
+        user = current_user()
+        if path.is_relative_to("~") and not user.is_running_user():
+            path = user.home / path.relative_to("~")
+        else:
+            path = path.expanduser()
 
     if required and not path.exists():
         die(f"{value} does not exist")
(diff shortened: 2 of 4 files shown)


mkosi/backend.py
```diff
@@ -471,10 +471,9 @@ def flatten(lists: Iterable[Iterable[T]]) -> list[T]:
     return list(itertools.chain.from_iterable(lists))
 
 
-def current_user_uid_gid() -> tuple[int, int]:
+def current_user() -> pwd.struct_passwd:
```
behrmann, line 474: I'm not sure we should expose the raw `struct_passwd`. In most places we only use one or two things, but it litters the code with the struct attribute names, somewhat obscuring what is happening.

mkosi/config.py
```diff
@@ -44,13 +47,43 @@ def parse_boolean(s: str) -> bool:
...
+        path = Path(string.Template(value).substitute(os.environ))
+    else:
+        path = Path(value)
+
+    if expanduser:
+        user = current_user()
+        if (
+            not path.is_absolute()
+            and path.parts
+            and path.parts[0] == "~"
```
behrmann, line 61:
```suggestion
            path.is_relative_to("~")
```

mkosi/run.py
```diff
@@ -66,6 +66,9 @@ def become_root() -> tuple[int, int]:
 
     The function returns the UID-GID pair of the invoking user in the namespace (65436, 65436).
     """
+    if os.getuid() == 0:
+        # Already root
```
DaanDeMeyer, line 70: This is pretty obvious, let's drop the comment

mkosi/config.py
```diff
@@ -44,13 +47,43 @@ def parse_boolean(s: str) -> bool:
...
+            and os.getuid() != user.pw_uid
+        ):
+            path = Path(user.pw_dir, *path.parts[1:])
```
behrmann, line 64:
```suggestion
            path = Path(user.pw_dir) / path.relative_to("~")
```
though I'd find the expanding `Path("~username")` and calling `Path.expanduser()` unconditionally clearer still.

mkosi/config.py
```diff
@@ -44,13 +47,43 @@ def parse_boolean(s: str) -> bool:
     die(f"Invalid boolean literal: {s!r}")
 
 
+def parse_path(value: str, required: bool, absolute: bool = True, expanduser: bool = True, expandvars: bool = True) -> Path:
```
behrmann, line 50: Not a fan of so many flags, at least 
```suggestion
def parse_path(value: str, required: bool, *, absolute: bool = True, expanduser: bool = True, expandvars: bool = True) -> Path:
```
or better
```suggestion
def parse_path(value: str, *, required: bool = True/False, absolute: bool = True, expanduser: bool = True, expandvars: bool = True) -> Path:
```

gdiscry: No need for default values after `, *,` in which case the parameter must be specified and must be named.

mkosi/backend.py
DaanDeMeyer: I think this is fine, it's more descriptive in whats being returned

mkosi/config.py
```diff
@@ -44,13 +47,38 @@ def parse_boolean(s: str) -> bool:
     die(f"Invalid boolean literal: {s!r}")
 
 
+def parse_path(value: str, *, required: bool, absolute: bool = True, expanduser: bool = True, expandvars: bool = True) -> Path:
+    if expandvars:
+        path = Path(string.Template(value).substitute(os.environ))
```
gdiscry, line 52: This can raise `ValueError` (if a $ is incorrectly used) or `KeyError` (if an environment variable is no set).

Should I encapsulate the exception? Use `.expandvars()` which would leave those intact ?

mkosi/backend.py
gdiscry: I'm not a fan of the `pw_` prefix. However, most of the struct attributes are used (`pw_uid`, `pw_gid`, `pw_name`, `pw_dir`).

mkosi/config.py
gdiscry: I used your suggestion.

However I'm not sure I agree with `Path.expanduser()` being clearer, we are clearly doing something special (and we already have access to the home directory).

mkosi/config.py
behrmann: Yes, sure, but having one mandatory keyword-only argument and three optional ones makes for a somewhat less uniform API, choosing a default (probably false, maybe all of them should be false by default) makes it more uniform because it handles all the keyword-only arguments on the same footing.

gdiscry: The default `True` value for the other parameters makes sense because there are only a few callers using `False`, so they are clearly the exception.

When I first introduced the `expanduser` and `expandvars` parameters, I used `False` as default. The result was a lot of parameters when configuring `search_extra_paths` and an inconsistency between that option and all the others for no obvious reason. Then I switched to `True` and it was clearer that `expandvars` can always be enabled and `expanduser` must be disabled for the targets of `parse_source_target_paths`.

However, I don't feel like there is a clear default for `required`. Some use `True`, some use `False`, and I find it useful that it's explicit when reading `MkosiConfigParser.SETTINGS`.

mkosi/backend.py
behrmann: > I think this is fine, it's more descriptive in whats being returned

It's more descriptive here, but I think the surrounding code where it calls becomes a bit harder to read.

The only use sites of `pw_dir` below would become
```python
        if path.is_relative_to("~") and os.getuid() != invoking_user_uid():
            path = Path(f"~{invoking_user_name}") / path.relative_to("~")
        path = path.expanduser()
```
for some 
```python
def invoking_user_uid() -> int:
    return int(os.getenv("SUDO_UID") or os.getenv("PKEXEC_UID") or os.getuid())

def invoking_user_uid_gid() -> tuple[int, int]:
    uid = invoking_user_uid()
    gid = int(os.getenv("SUDO_GID") or pwd.getpwuid(uid).pw_gid)
    return uid, gid

def invoking_user_name() -> str:
    return os.getenv("SUDO_USER") or os.getenv("USER") or os.getlogin()
```

Maybe more generally:
```python
class InvokingUser:
    @staticmethod
    def uid() -> int:
         return int(os.getenv("SUDO_UID") or os.getenv("PKEXEC_UID") or os.getuid())
    @classmethod
    def gid(cls) -> int:
        return int(os.getenv("SUDO_GID") or pwd.getpwuid(cls.uid()).pw_gid)
    @staticmethod
    def name() -> str:
        return os.getenv("SUDO_USER") or os.getenv("USER") or os.getlogin()
    @classmethod
    def home(cls) -> Path:
        if p := (os.getenv("SUDO_HOME") or os.getenv("HOME")):
            return Path(p)
        return Path(f"~{cls.name()}").expanduser()
    @classmethod
    def is_running_user(cls) -> bool:
         return cls.uid == os.getuid()
```

mkosi/config.py
DaanDeMeyer: Probably `expandvars()`

DaanDeMeyer (approved):

mkosi/backend.py
gdiscry: I like the approach with a class, but I'm less a fan of going through the other `SUDO_`* environment variables, because they do not work when pkexec is used (name would be root and home would be /root). Or to be correct, they will not work once `bin/mkosi` is dropped.

mkosi/backend.py
```diff
@@ -471,10 +471,49 @@ def flatten(lists: Iterable[Iterable[T]]) -> list[T]:
...
+    def gid(self) -> int:
+        if self._pw is not None:
+            return self._pw.pw_gid
+        return os.getgid()
+
+    @property
+    def name(self) -> str:
+        if self._pw is not None:
+            return self._pw.pw_name
+        return os.getlogin()
```
gdiscry, line 498: Or should we use `getpass.getuser()`?

gdiscry:
I made it so that the `pwd` entry is only used with sudo/pkexec. Otherwise, the higher-level functions are used.

mkosi/backend.py
```diff
@@ -471,10 +471,49 @@ def flatten(lists: Iterable[Iterable[T]]) -> list[T]:
...
+
+    @property
+    def home(self) -> Path:
+        if self._pw is not None:
+            return Path(self._pw.pw_dir)
+        return Path("~").expanduser()
```
behrmann, line 504:
```suggestion
        return Path().home()
```

mkosi/backend.py
```diff
@@ -471,10 +471,49 @@ def flatten(lists: Iterable[Iterable[T]]) -> list[T]:
     return list(itertools.chain.from_iterable(lists))
 
 
-def current_user_uid_gid() -> tuple[int, int]:
-    uid = int(os.getenv("SUDO_UID") or os.getenv("PKEXEC_UID") or os.getuid())
-    gid = pwd.getpwuid(uid).pw_gid
-    return uid, gid
+@dataclasses.dataclass
+class InvokingUser:
+    _pw: Optional[pwd.struct_passwd] = None
```
behrmann, line 476: I don't think this should be cached
```suggestion
    @staticmethod
    def _uid_from_env() -> Optional[int]
        if uid := (os.getenv("SUDO_UID") or os.getenv("PKEXEC_UID")):
            return int(uid)
        return None
```
this would also preclude this form being a dataclass and the properties would need to be methods, since class properties or static properties don't exist without hackery.

The rest would then become 
```python
@classmethod
def uid(cls) -> int:
    return _uid_from_env() or os.getuid()

Comment (gdiscry) [reply mkosi/backend.py:476]: It's not cached, each call to current_user() will return a new instance of `InvokingUser`. `InvokingUser._pw` is an instance attribute, not a class attribute (as it's a dataclass).

File: mkosi/backend.py (hunk 8 above)
Comment (gdiscry) [reply mkosi/backend.py:504]: Nice catch 😅.

File: mkosi/backend.py (hunk 9 above)
Comment (behrmann) [reply mkosi/backend.py:476]: Well, it is cached in the instance. Of course, if you call the (alternative) constructor again, you will get a different instance. 

What I am proposing is just using a class as a namespace for a few associated functions instead of going through the whole dataclass machinery every time. I'll do a followup.

Comment (gdiscry) [reply mkosi/backend.py:476]: > Well, it is cached in the instance. Of course, if you call the (alternative) constructor again, you will get a different instance.

But that's kind of the point. Once we have a uid (validated as existing in passwd), the other attributes are available and coherent with that uid. If I have some logic based on one attribute (the uid most of the time), the other attributes are guaranteed to be for the same uid.

Furthermore, apart from using the `SUDO_`* environment variables (but see https://github.com/systemd/mkosi/pull/1472#discussion_r1173669571), getting those attributes from the uid always goes through `pwd.getpwuid` anyway (directly or indirectly).

And dropping the alternative constructor can easily be done by directly using a default factory for `_pw`. If you want, I can create a new PR for that.

> What I am proposing is just using a class as a namespace for a few associated functions instead of going through the whole dataclass machinery every time. I'll do a followup.

But there is no dataclass machinery. An advantage of dataclasses is that it just generates the boilerplate code for the missing dunder methods when the class is decorated.

Anyway, I'm interested in your followup.

**Comment by DaanDeMeyer** [comment]
@gdiscry This is causing a CI failure in systemd:

```
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.10/dist-packages/mkosi/__main__.py", line 60, in <module>
    main()
  File "/usr/lib/python3.10/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/usr/local/lib/python3.10/dist-packages/mkosi/__main__.py", line [5](https://github.com/systemd/systemd/actions/runs/4780001667/jobs/8497443434?pr=27368#step:5:6)[6](https://github.com/systemd/systemd/actions/runs/4780001667/jobs/8497443434?pr=27368#step:5:7), in main
    run_verb(load_args(args))
  File "/usr/local/lib/python3.10/dist-packages/mkosi/__init__.py", line 240[7](https://github.com/systemd/systemd/actions/runs/4780001667/jobs/8497443434?pr=27368#step:5:8), in run_verb
    return generate_secure_boot_key(config)
  File "/usr/local/lib/python3.10/dist-packages/mkosi/__init__.py", line 2317, in generate_secure_boot_key
    cn = expand_specifier(config.secure_boot_common_name)
  File "/usr/local/lib/python3.10/dist-packages/mkosi/__init__.py", line 23[9](https://github.com/systemd/systemd/actions/runs/4780001667/jobs/8497443434?pr=27368#step:5:10)7, in expand_specifier
    return s.replace("%u", current_user().name)
  File "/usr/local/lib/python3.[10](https://github.com/systemd/systemd/actions/runs/4780001667/jobs/8497443434?pr=27368#step:5:11)/dist-packages/mkosi/backend.py", line 473, in name
    return os.getlogin()
OSError: [Errno 6] No such device or address
```

gdiscry:
#1492 should fix that.

PR #1475, opened 2023-04-21
Add MakeInitrd=

DaanDeMeyer:
Instead of automatically making cpio's initrds, let's put this behind a MakeInitrd= option. This allows booting directly into cpios with systemd when they're not configured as an initramfs.


DaanDeMeyer:
This one is trivial so I'm going to go ahead and merge it early

PR #1476, opened 2023-04-21
Get rid of MkosiPrinter

DaanDeMeyer:
It's unnecessarily verbose for no reason. Let's mimick systemd and just have functions to log prefixed with log_(). For the styling, we add a namespace class Style.


DaanDeMeyer:
Nvm, have a better approach

PR #1477, opened 2023-04-21
Use python logging module for logging

DaanDeMeyer:
Let's get rid of our home grown solution and use python's logging module instead.


DaanDeMeyer:
Merging this one early to make some follow up work easier.

PR #1479, opened 2023-04-21
Fix installation of essential packages on Debian stretch

gdiscry:
There is a bug in Debian stretch where `libuuid1` (which is essential) unnecessarily depends on `passwd`. When configuring the essential packages and their dependencies, `passwd` is configured before `base-passwd` and the installation fails.

debootstrap did not have this issue because it explicitly orders the installation of some essential packages, including `base-passwd`.

By explicitly installing `base-passwd` first, the rest of the essential packages can then be configured without failure.

There is no hope for a fix from Debian as the package was fixed in 2017 but only released for buster. I also could not find a fix that would work by putting some files in mkosi.skeleton and/or setting some environment variables.

### DIFF
--- a/mkosi/distributions/debian.py
+++ b/mkosi/distributions/debian.py
@@ -101,6 +101,12 @@ class DebianInstaller(DistributionInstaller):
                 run(["dpkg-deb", "--fsys-tarfile", deb], stdout=f)
                 run(["tar", "-C", state.root, "--keep-directory-symlink", "--extract", "--file", f.name])
 
+        # There is a bug in Debian stretch where libuuid1 (which is essential) unecessarily depends on passwd,
+        # which breaks the installation as passwd is then configured before base-passwd
+
+        if state.config.release == "stretch":
+            cls.install_packages(state, ["base-passwd"])
+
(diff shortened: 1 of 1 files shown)


behrmann:
Why is this an issue? stretch went out of support three years ago and even extended support last year.

gdiscry:
It's still in ELTS by Freexian 😇.

But more seriously, because it's (sadly) useful for me and I could not find another way to fix it…

There are a few other mentions of stretch in the code. It's still publicly visible as oldoldstable and distributed normally until the release of bookworm.

The fix is clearly marked so that it can be removed when support for stretch is dropped from mkosi.

behrmann:
> There are a few other mentions of stretch in the code.

Before merging this PR there was one other mention of stretch. At this point we should not be adding *more* support for stretch. Upstream Debian discontinued security support and just because commercial support is available by some company - something the release of bookworm some time this summer will change nothing about - is no reason to keep up the appearance that stretch is a viable platform.

DaanDeMeyer:
Meh if its trivial stuff like this, we can do it. If the fixes are more involved, that's another story

PR #1480, opened 2023-04-22
Display "Skeleton Trees" in the summary again

gdiscry:
The skeleton trees were lost from the summary with e1bbc39.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1448,6 +1448,7 @@ def print_summary(config: MkosiConfig) -> None:
                   Packages: {line_join_list(config.packages)}
         With Documentation: {yes_no(config.with_docs)}
              Package Cache: {none_to_none(config.cache_dir)}
+            Skeleton Trees: {line_join_source_target_list(config.skeleton_trees)}
                Extra Trees: {line_join_source_target_list(config.extra_trees)}
     Clean Package Metadata: {yes_no_auto(config.clean_package_metadata)}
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -1449,6 +1449,7 @@ def print_summary(config: MkosiConfig) -> None:
         With Documentation: {yes_no(config.with_docs)}
              Package Cache: {none_to_none(config.cache_dir)}
                Extra Trees: {line_join_source_target_list(config.extra_trees)}
+            Skeleton Trees: {line_join_source_target_list(config.skeleton_trees)}
```
behrmann, line 1452: Somehow I missed that this was removed. While that's the order it was printed in before, mentioning skeleton trees before extra trees would make a bit more sense, since they are applied first.
```suggestion
            Skeleton Trees: {line_join_source_target_list(config.skeleton_trees)}
               Extra Trees: {line_join_source_target_list(config.extra_trees)}
```

gdiscry: Done.

Initially, I also wanted to check the order of the summary, but gave up and just put back skeleton trees where it was.

DaanDeMeyer:
Please align the line properly with the rest of the summary

gdiscry:
I don't understand, it's aligned. I've re-checked that there is not a mixed of tabs and spaces, but no.

DaanDeMeyer:
Must be something with GitHub mobile then

PR #1481, opened 2023-04-22
Enable apt configuration with skeleton files

### LINKED ISSUE #1419: Make `--install-recommends` configurable
The distribution scripts default hard-coded to not installing recommended dependencies.

While this is a sane default, it should be configurable with an option, probably separately for the build and final images.

gdiscry:
The configuration passed to apt with $APT_CONFIG no longer ignores files from `mkosi.skeleton/etc/apt`, making it possible to configure apt by:

- adding new repositories in `sources.list.d`
- tweaking priorities in `preferences.d` (e.g. for backports)
- configuring apt itself in `apt.conf` and `apt.conf.d` (e.g. `APT::Install-Recommends "true";`)

Essential packages are installed with `--no-install-recommends` to avoid pulling too many packages during that step.

Should fix #1419.

### DIFF
--- a/mkosi/distributions/debian.py
+++ b/mkosi/distributions/debian.py
@@ -54,7 +54,8 @@ class DebianInstaller(DistributionInstaller):
         # impossible to use apt directly to bootstrap a Debian chroot since dpkg will try to run a maintainer
         # script which depends on some basic tool to be available in the chroot from a deb which hasn't been
         # unpacked yet, causing the script to fail. To avoid these issues, we have to extract all the
-        # essential debs first, and only then run the maintainer scripts for them.
+        # essential debs first, and only then run the maintainer scripts for them. Each call to apt-get requires
+        # --no-install-recommends as 'APT::Install-Recommends "false"' could be overrided by a skeleton file.
 
         # First, we set up merged usr.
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
We shouldn't look for distribution config files in the image root itself. Rather we should support specifying a directory with additional package manager configuration. We already have --repository-directory for specifying extra repositories, maybe we should generalize this as a directory for putting extra package manager configuration. If the package manager does not support multiple configuration directories, we can just copy all the stuff from those directories into one directory in the workspace and use that. It's a bit long winded but we could rename --repository-directory to --package-manager-directory and allow dropping all config in there. I think a single directory should be fine since all package manager seem to use distinct file extensions for different kinds of configs so we can always do post processing per package manager to split those files up into more directories based on the file extension if needed.

gdiscry:
I see your point. My current use of mkosi with Debian is to create the initial image for the system, but keeping everything in place as if it was installed normally. But using the configuration from the image is more of an issue for image-based system where the metadata of the distribution is not wanted anymore.

A generalization of `--repo-dir` to more globally configure the package managers seems the way to go, but I have no experience with DNF.

As a first step, I've opened #1490 to at least add `--repo-dir` support to Debian. It will probably inform how a generalized approach would look like.

Issue #1483, opened 2023-04-22
[Match] multiple Distribution= with OR


iam-TJ:
Whilst re-writing several project configurations to work with the `[Match]` semantics and having fragments for both Debian and Ubuntu I realised `Packages=`  and `BuildPackages=` has to be duplicated in both despite being the same. That made me wish I could do:
```
[Match]
Distribution=debian
Distribution=ubuntu
```
or
```
[Match]
Distribution=debian ubuntu
```
and a logical OR would be applied to the multiple names.

DaanDeMeyer:
Makes perfect sense. I'd go for something like this:

```
[Match]
Distribution=debian|ubuntu
```

Will happily accept a PR for this. should be trivial to implement.

iam-TJ:
I already started experimenting but thought using existing list semantics (space separated) rather than introducing an operator would be preferred.

DaanDeMeyer:
We should add `config_make_list_matcher()` which mimicks `config_make_list_parser()` but splits the value based on the configured delimiter, and passes each value to the provided match function and if any match, it returns True, otherwise False.

iam-TJ:
That makes sense.

PR #1485, opened 2023-04-22
Match: allow multiple distributions

iam-TJ:
For distributions like Debian and Ubuntu that share common infrastructure this allows sharing common settings such as Packages= and BuildPackages=

Fixes: Issue #1483

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -192,6 +192,20 @@ def config_make_enum_matcher(type: Type[enum.Enum]) -> ConfigMatchCallback:
 
     return config_match_enum
 
+def config_make_list_matcher(type: Type[enum.Enum], delimiter: str, parse: Callable[[str], Any] = bool) -> ConfigMatchCallback:
+    def config_match_list(dest: str, value: str, namespace: argparse.Namespace) -> bool:
+        values = value.replace(delimiter, "\n").split("\n")
+
+        for v in values:
+            if not v:
+                continue
+
+            if cast(bool, make_enum_parser(type)(v) == getattr(namespace, dest)):
+                return True
+
+        return False
+
+    return config_match_list
 
 def config_make_list_parser(delimiter: str, unescape: bool = False, parse: Callable[[str], Any] = str) -> ConfigParseCallback:
     ignore: set[str] = set()
@@ -358,7 +372,7 @@ class MkosiConfigParser:
             dest="distribution",
             section="Distribution",
             parse=config_make_enum_parser(Distribution),
-            match=config_make_enum_matcher(Distribution),
+            match=config_make_list_matcher(Distribution, delimiter="|"),
             default=detect_distribution()[0],
         ),
         MkosiConfigSetting(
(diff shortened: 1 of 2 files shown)


iam-TJ:
The one thing I haven't been able to figure out is how to add/modify the tests.

DaanDeMeyer:
No need to bother with tests, I still need to write tests for all the config parsing.

But we're moving pretty fast right now and things are still rapidly changing, so too many unit tests would slow us down. Manual tests and the integration tests we have in CI are fine for now.

mkosi/config.py
```diff
@@ -192,6 +192,22 @@ def config_match_enum(dest: str, value: str, namespace: argparse.Namespace) -> b
 
     return config_match_enum
 
+def config_make_list_matcher(type: Type[enum.Enum], delimiter: str) -> ConfigMatchCallback:
```
DaanDeMeyer, line 195: Similar to `config_make_list_parser()`, this should take `match` function that does the actual match for a single element in the list.

mkosi/config.py
```diff
@@ -192,6 +192,22 @@ def config_match_enum(dest: str, value: str, namespace: argparse.Namespace) -> b
...
+    def config_match_list(dest: str, value: str, namespace: argparse.Namespace) -> bool:
+        matched = False
```
DaanDeMeyer, line 197: No need for this variable, just return immediately if the match function returns true once

mkosi/config.py
iam-TJ: I cannot fathom how these Callable things work, even after reading the Python docs, and the prototype for `ConfigParseCallback`  has a different prototype to `ConfigMatchCallback` so I'm grasping at straws. Since the code works as-is, and with the following possible revision, I'm getting no help from the interpreter either. Give me a strongly typed compiled language any day - help!
```
# prototype: ConfigMatchCallback = Callable[[str, str, argparse.Namespace], bool]
def config_make_list_matcher(type: Type[enum.Enum], delimiter: str, parse: Callable[[str], Any] = bool) -> ConfigMatchCallback:
    def config_match_list(dest: str, value: str, namespace: argparse.Namespace) -> bool:
```

man/mkosi.1
```diff
@@ -259,6 +259,8 @@ either \[lq]1\[rq], \[lq]yes\[rq], or \[lq]true\[rq] to enable, or
 .TP
 \f[V]Distribution=\f[R]
 Matches against the configured distribution.
+Matches any one of multiple distributions. Separate multiple distribution
+names with \[lq]|\[rq] e.g: \[lq]debian|ubuntu\[rq].
```
behrmann, line 263: This file is generated from `mkosi.md` in the project root, we don't force anybody to write raw troff. :) Please make the changes there, so they don't get lost when the file is regenerated.

mkosi/config.py
behrmann: `Callable` is just how to encode a function in the type system. The first argument to `Callable` is the list of types of the arguments (only positional arguments are supported), the second is the return type. 

The interpreter is of little help here, but you can run a type checker, mainly `mypy`, but also `pyright` and others on it. It's not much different from a statically compiled language then.

DaanDeMeyer:
I took the liberty to implement it myself in #1487 because we need it in systemd so closing this one

PR #1486, opened 2023-04-22
debian: Add TODO to drop /var/lib/dpkg/status logic when possible

DaanDeMeyer:
It's created automatically by apt since apt 2.5.4 so add a note that
we can drop it once that version is widely available. Also disable
dpkg locking since it causes a apt failure due to not being able to
access the lock file in /var/lib/dpkg. Since we're building in a
chroot, we don't care about locking anyway.


DaanDeMeyer:
Merging since this one is trivial

PR #1487, opened 2023-04-23
Support matching against any of a list of distributions

DaanDeMeyer:
Replaces #1485


bluca (approved):

gdiscry:
Should we add the same logic to Release=?

DaanDeMeyer:
> Should we add the same logic to Release=?

Sure, would happily review a PR for that.

PR #1489, opened 2023-04-23
Disallow `<SOURCE>:<TARGET>` with an empty target

gdiscry:
The configurations of the form `<SOURCE>[:<TARGET>]` require an absolute target if given. However, `<SOURCE>:` would be ambiguously interpreted as `<SOURCE>` so it is now disallowed.

For reference, `systemd-nspawn --bind` uses a similar form and refuses to take an empty target.


DaanDeMeyer (changes requested):
Does `parse_path()` refuse an empty path? I don't think it does

gdiscry:
No it doesn't, but an empty path is relative and it will fail the check.

PR #1490, opened 2023-04-23
Enable --repo-dir for Debian and Ubuntu

gdiscry:
Additional repositories can now be configured for Debian and Ubuntu.

Because APT can only be configured with a single path for `Dir::Etc::sourceparts`, the content of the directories listed in `repo_dirs` are copied in a `apt/sources.list.d` directory managed by mkosi outside the image root. Therefore, only one file will be kept if several files have the same name.

Like the main `sources.list` generated by mkosi, the additional repositories are copied in the image if APT is installed.

### DIFF
--- a/mkosi/distributions/debian.py
+++ b/mkosi/distributions/debian.py
@@ -126,14 +126,9 @@ class DebianInstaller(DistributionInstaller):
         setup_apt(state, cls.repositories(state))
         invoke_apt(state, "update", apivfs=False)
         invoke_apt(state, "install", packages, apivfs=apivfs)
+        install_apt_sources(state, cls.repositories(state, local=False))
 
         policyrcd.unlink()
-
-        sources = state.root / "etc/apt/sources.list"
-        if not sources.exists() and state.root.joinpath("usr/bin/apt").exists():
-            with sources.open("w") as f:
-                for repo in cls.repositories(state, local=False):
-                    f.write(f"{repo}\n")
 
     @classmethod
     def remove_packages(cls, state: MkosiState, packages: Sequence[str]) -> None:
@@ -163,6 +158,7 @@ def setup_apt(state: MkosiState, repos: Sequence[str]) -> None:
     state.workspace.joinpath("apt").mkdir(exist_ok=True)
     state.workspace.joinpath("apt/apt.conf.d").mkdir(exist_ok=True)
     state.workspace.joinpath("apt/preferences.d").mkdir(exist_ok=True)
+    state.workspace.joinpath("apt/sources.list.d").mkdir(exist_ok=True)
     state.workspace.joinpath("apt/log").mkdir(exist_ok=True)
 
     # TODO: Drop once apt 2.5.4 is widely available.
@@ -206,6 +202,13 @@ def setup_apt(state: MkosiState, repos: Sequence[str]) -> None:
         for repo in repos:
             f.write(f"{repo}\n")
 
+    for repo_dir in state.config.repo_dirs:
+        for src in repo_dir.iterdir():
+            if not src.is_file():
+                continue
+            if src.suffix in (".list", ".sources"):
+                shutil.copyfile(src, state.workspace.joinpath("apt/sources.list.d", src.name))
+
 
 def invoke_apt(
     state: MkosiState,
@@ -222,3 +225,21 @@ def invoke_apt(
     )
 
     return bwrap(["apt-get", operation, *extra], apivfs=state.root if apivfs else None, env=env | state.environment)
+
+
+def install_apt_sources(state: MkosiState, repos: Sequence[str]) -> None:
+    if not state.root.joinpath("usr/bin/apt").exists():
+        return
+
+    sources = state.root / "etc/apt/sources.list"
+    if not sources.exists():
+        with sources.open("w") as f:
+            for repo in repos:
+                f.write(f"{repo}\n")
+
+    # Already contains a merged tree of repo_dirs after setup_apt
+    for src in state.workspace.joinpath("apt/sources.list.d").iterdir():
+        dst = state.root.joinpath("etc/apt/sources.list.d", src.name)
+        if dst.exists():
+            continue
+        shutil.copyfile(src, dst)
--- a/mkosi/install.py
+++ b/mkosi/install.py
@@ -47,11 +47,17 @@ def flock(path: Path) -> Iterator[Path]:
         os.close(fd)
 
 
-def copy_path(src: Path, dst: Path, preserve_owner: bool = True) -> None:
+def copy_path(
+    src: Path,
+    dst: Path,
+    *,
+    dereference: bool = False,
+    preserve_owner: bool = True,
+) -> None:
     run([
         "cp",
         "--recursive",
-        "--no-dereference",
+        f"--{'' if dereference else 'no-'}dereference",
         f"--preserve=mode,timestamps,links,xattr{',ownership' if preserve_owner else ''}",
         "--no-target-directory",
         "--reflink=auto",
--- a/mkosi.md
+++ b/mkosi.md
@@ -287,11 +287,12 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
 
 `RepositoryDirectories`, `--repo-dir=`
 
-: This option can (for now) only be used with RPM-based distributions and Arch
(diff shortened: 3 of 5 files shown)


mkosi/distributions/debian.py
```diff
@@ -130,10 +130,13 @@ def install_packages(cls, state: MkosiState, packages: Sequence[str]) -> None:
         policyrcd.unlink()
 
         sources = state.root / "etc/apt/sources.list"
+        sources_dir = state.root / "etc/apt/sources.list.d"
         if not sources.exists() and state.root.joinpath("usr/bin/apt").exists():
             with sources.open("w") as f:
                 for repo in cls.repositories(state, local=False):
                     f.write(f"{repo}\n")
+            for repo_dir in state.config.repo_dirs:
+                shutil.copytree(repo_dir, sources_dir, dirs_exist_ok=True)
```
DaanDeMeyer, line 139: We shouldn't overwrite existing files here, this should be purely additive. I'd extend copy_path() with `overwrite` and `dereference` booleans and use that here (with overwrite disabled and dereference enabled) instead of `copytree()`

DaanDeMeyer (changes requested):
Looks good, except for one thing

mkosi/install.py
```diff
@@ -47,11 +47,19 @@ def flock(path: Path) -> Iterator[Path]:
...
+    dereference: bool = False,
+    overwrite: bool = True,
+    preserve_owner: bool = True,
+) -> None:
     run([
         "cp",
         "--recursive",
-        "--no-dereference",
+        f"--{'' if dereference else 'no-'}dereference",
+        f"--update={'all' if overwrite else 'none'}",
```
DaanDeMeyer, line 62: We should use --no-clobber here instead of --update

gdiscry: So just to make sure, you want it to fail if `overwrite=False` and a file already exists?

gdiscry:
```console
$ touch foo
$ touch bar
$ cp --no-clobber foo bar
cp: not replacing 'bar'
$ echo $?
1
```

DaanDeMeyer: This seems like a huge mess, let's do it in python instead, but keep it simple, just copy files, no need for a full recursive copy

mkosi/distributions/debian.py
```diff
@@ -129,11 +130,25 @@ def install_packages(cls, state: MkosiState, packages: Sequence[str]) -> None:
...
+
+            # Already contains a merged tree of repo_dirs after setup_apt
+            workspace_sources_dir = state.workspace / "apt/sources.list.d"
+            if workspace_sources_dir.exists():
+                sources_dir = state.root / "etc/apt/sources.list.d"
+                sources_dir.mkdir(mode=0o755, parents=True, exist_ok=True)
+                for src in workspace_sources_dir.iterdir():
+                    if not src.is_file():
+                        continue
+                    dst = sources_dir / src.relative_to(workspace_sources_dir)
```
DaanDeMeyer, line 148: You can use `sources_dir / src.name` here no?

DaanDeMeyer (changes requested):
Let's put this into its own function install_apt_mirrors() which doesn't do anything if apt isn't installed.

mkosi/distributions/debian.py
```diff
@@ -221,3 +220,26 @@ def invoke_apt(
...
+        return
+
+    sources = state.root / "etc/apt/sources.list"
+    if not sources.exists():
+        with sources.open("w") as f:
+            for repo in repos:
+                f.write(f"{repo}\n")
+
+    # Already contains a merged tree of repo_dirs after setup_apt
+    workspace_sources_dir = state.workspace / "apt/sources.list.d"
```
DaanDeMeyer, line 236: Let's create this directory unconditionally in the workspace so we don't have to check if it exists here. Then we can get rid of the workspace_sources_dir and sources_dir variables since both will only have one use.

DaanDeMeyer (changes requested):
One more thing and its good to go

DaanDeMeyer (approved):

gdiscry:
Thanks for the feedback. Next step: --package-manager-directory 😅

DaanDeMeyer:
> Thanks for the feedback. Next step: --package-manager-directory sweat_smile

Actually, taking that into consideration, maybe only copy files with the .list suffix from the repo directories? Since we'll need that kind of logic for --package-manager-directory for sure

DaanDeMeyer (approved):

behrmann:
Bit of a late review: LGTM, but besides `sources.list.d` this should also cover `trusted.gpg.d`, since most external repos will needs to reference a separate gpg key.

mcassaniti:
If I have this correct, you'd be suggesting that any directories under `--package-manager-directory` would magically appear for apt to use? Do we really need to cherry-pick which files apt sees?

Remember too that a source list could have an absolute path specified. See the example below. This means mounting over the top of `/etc/apt` when calling `invoke_apt()`. I'm (extremely) happy to be proved wrong.

```
Types: deb
URIs: https://download.virtualbox.org/virtualbox/debian
Suites: jammy
Components: contrib
Signed-By: /etc/apt/keyrings/virtualbox.gpg
Arch: amd64
```

DaanDeMeyer:
> When looking at the content of /etc/apt, I noticed that dispatching by extension already is problematic because apt.conf.d and auth.conf.d both contain *.conf files.

Yeah perhaps we should go with separate options. It quickly becomes a mess though. e.g. apt uses a keyring directory but the rpm based distros use absolute paths to refer to keys and the archlinux keyring and opensuse are other beasts entirely. It's not going to be trivial to figure out an abstraction that works for all distros.

> Remember too that a source list could have an absolute path specified. See the example below. This means mounting over the top of /etc/apt when calling invoke_apt(). I'm (extremely) happy to be proved wrong.

If the source list specifies an absolute path you have to make sure the key is installed at that path on the host. That's more or less the same as us expecting the debian keys to be in /usr/share/keyrings.

mcassaniti:
> If the source list specifies an absolute path you have to make sure the key is installed at that path on the host. That's more or less the same as us expecting the debian keys to be in /usr/share/keyrings.

Maybe not. Apt is running in a separate mount namespace that can have whatever mounts are required. If `--package-manager-directory` is specifying what to bring in, then setting up something to mount at `/etc/apt` and going from there is reasonable. For example, a temporary directory could be populated with the basic contents that `setup_apt()` does now and then mounted at `/etc/apt` as required. Anything in addition could be copied into the temporary directory.

I do see that we can populate `/etc/apt` correctly without needing to modify the host, and we certainly can combine `/usr/share/keyrings` too if required. See #1520 as an example. How we populate those things is, well, the art of software development.



> What we could do with --package-manager-directory is treat it as an /etc directory, that would allow us to keep the directory structure intact. So for dnf we'd look in the directory for ...

I like this idea. If we're coping things around then it might become even easier (i.e.: just populate a base directory with what is necessary, copy over the additional contents then bind mount).

DaanDeMeyer:
> Maybe not. Apt is running in a separate mount namespace that can have whatever mounts are required. If --package-manager-directory is specifying what to bring in, then setting up something to mount at /etc/apt and going from there is reasonable. For example, a temporary directory could be populated with the basic contents that setup_apt() does now and then mounted at /etc/apt as required. Anything in addition could be copied into the temporary directory.

What I'm saying is that we shouldn't be working around broken repo files. The repo file shouldn't be using an absolute path that completely ignores where we tell apt to look for its configuration and keyrings. If it does, that's something to be fixed in the repo file, not something for us to work around.

DaanDeMeyer:
We should go about extra keys for Debian as follows:

In the directory specified by --package-manager-directory, you can have a directory trusted.d where you can put extra keys. Internally, we'll combine those keys with the keys from /usr/share/keyrings and store them in `state.workspace / "apt/trusted.gpg.d` and tell apt to look for gpg keys there. Any mirrors should be declared such that they will look for keys in the trustedparts directory specified in apt config.

mcassaniti:
I see your point about wanting to use relative key paths, but you can't specify relative paths in those files.

from `man 5 sources.list`:
>Signed-By (signed-by) is an option to .... It is specified as a list of **absolute paths** to keyring files.

The other option is to use inline signature blocks (also in the man page). I'm going to give that one a go instead considering I'm doing the keyring conversion during my builds.


Nothing is ever as easy as we'd want it to be.

PR #1491, opened 2023-04-23
Tweak exception handling

DaanDeMeyer:
- Remove optional exception argument of die() in favor of just using
  logging.error() and re-raising the exception
- Instead of raising RuntimeError in die(), just call sys.exit()
  instead which raises SystemExit.
- Show stacktrace of every exception that isn't SystemExit,
  KeyboardInterrupt or subprocess.CalledProcessError.


keszybz:
Looks reasonable, but I'd want to do a few builds with this version to see how it works out in practice.

keszybz:
isort is unhappy.

keszybz:
One downside that I see is that it makes it harder to import mkosi as a module. But mkosi isn't really designed for this and it doesn't work well anyway, so that's not a huge loss.

I suspect we would want to suppress some more tracebacks, e.g. for PermissionError. But we don't currently do that and this PR does not change anything.

So… let's merge this. If it turns out to not work for some reason, we can always revisit. Clearly it's hard to strike the right balance here.

keszybz (approved):

PR #1495, opened 2023-04-24
Various refactorings and improvements

DaanDeMeyer:


behrmann:
The CentOS family fails, otherwise looks sensible all around to me. 

> backend.py is a horrible name, let's use util.py instead.

Hey, I always asked whether anybody had a better name ;)

DaanDeMeyer:
All the centos/fedora builds passed, let's merge

PR #1498, opened 2023-04-24
Make --debug a boolean

DaanDeMeyer:
Let's drop the support for multiple string arguments since we only have "run" anyway and make --boolean a simple yes/no option.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -931,9 +931,12 @@ Those settings cannot be configured in the configuration files.
 
 `--debug=`
 
-: Enable additional debugging output. Takes a comma-separated list of
-  arguments specifying the area of interest. Pass any invalid value
-  (e.g. empty) to list currently accepted values.
+: Enable additional debugging output.
+
+`--debug-shell=`
+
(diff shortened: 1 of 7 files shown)


mkosi.md
```diff
@@ -931,9 +931,9 @@ Those settings cannot be configured in the configuration files.
 
 `--debug=`
 
-: Enable additional debugging output. Takes a comma-separated list of
-  arguments specifying the area of interest. Pass any invalid value
-  (e.g. empty) to list currently accepted values.
+: Enable additional debugging output. Aditionally, when executing a command
+  in the image fails. mkosi will start an interactive shell in the image
```
keszybz, line 935: There's a stray dot after "fails", instead of a comma.

I think this "interactive shell" thing should be a separate option. It is blocking and unexpected. We give a hint to users to use `--debug`, but starting an interactive shell seems like too big of a consequence.

 I like the idea in general though.

DaanDeMeyer: Added --debug-shell for the interactive shell stuff

behrmann (approved):

PR #1500, opened 2023-04-24
Assorted followups

behrmann:


gdiscry (commented):
The API has a better feel to it, I like it.

However this loses a concept I liked after your feedback on my PR and this PR, which is:

1. If `$SUDO_UID` or `$PKEXEC_UID` are defined, we try to use the `$SUDO_`* envvars and fallback to `pwd`.
2. If not, we use the functions from the standard library as they correctly handle the envvars and fall back to `pwd` internally.

mkosi/config.py
```diff
@@ -56,18 +56,21 @@ def parse_boolean(s: str) -> bool:
...
 
     if expanduser:
-        user = current_user()
-        if path.is_relative_to("~") and not user.is_running_user():
-            path = user.home / path.relative_to("~")
-        else:
-            path = path.expanduser()
+        if path.is_relative_to("~") and not InvokingUser.is_running_user():
+            path = InvokingUser.home() / path.relative_to("~")
+        path = path.expanduser()
```
gdiscry, line 73: `InvokingUser.is_running_user()` and `InvokingUser.home()` are only used here, so should we instead do:

```suggestion
        path = InvokingUser.expanduser(path)
```

behrmann:
Addressed the comments, but I left the stuff around the home and `.is_running_user`, since I think that keeps the code where it is used clearer.

mkosi/util.py
```diff
@@ -221,49 +226,36 @@ def flatten(lists: Iterable[Iterable[T]]) -> list[T]:
...
+        return int(cls._uid_from_env() or os.getuid())
 
-    @property
-    def home(self) -> Path:
-        if self._pw is not None:
-            return Path(self._pw.pw_dir)
-        return Path.home()
+    @classmethod
+    def uid_gid(cls) -> tuple[int, int]:
+        if uid := cls._uid_from_env:
```
gdiscry, line 240:
```suggestion
        if uid := cls._uid_from_env():
```

mkosi/util.py
```diff
@@ -221,49 +226,36 @@ def flatten(lists: Iterable[Iterable[T]]) -> list[T]:
...
-        return True
+    @classmethod
+    def name(cls) -> str:
+        if uid := cls._uid_from_env():
+            return pwd.getpwuid(uid).pw_name
+        return getpass.getuser()
 
+    @classmethod
+    def home(cls) -> Path:
+        home = os.getenv("SUDO_HOME") or os.getenv("HOME") or pwd.getpwuid(cls.uid()).pw_dir
```
gdiscry, line 253: This is wrong, it returns `Path("/root")` with sudo and pkexec. `$SUDO_HOME` is not a real environment variable set by sudo, but a fake one that was set by `--extra-search-path` when it was first introduced and now replaced by expanding `~`.

mkosi/util.py
```diff
@@ -221,49 +226,36 @@ def flatten(lists: Iterable[Iterable[T]]) -> list[T]:
...
 
-    @property
-    def home(self) -> Path:
-        if self._pw is not None:
-            return Path(self._pw.pw_dir)
-        return Path.home()
+    @classmethod
+    def uid_gid(cls) -> tuple[int, int]:
+        if uid := cls._uid_from_env:
+            gid = int(os.getenv("SUDO_GID") or pwd.getpwuid(uid).pw_gid)
```
gdiscry, line 241:
```suggestion
            gid = int(os.getenv("SUDO_GID") or pwd.getpwuid(int(uid)).pw_gid)
```

mkosi/util.py
```diff
@@ -221,49 +226,36 @@ def flatten(lists: Iterable[Iterable[T]]) -> list[T]:
...
+        return os.getuid(), os.getgid()
 
-    def is_running_user(self) -> bool:
-        if self._pw is not None:
-            return self._pw.pw_uid == os.getuid()
-        return True
+    @classmethod
+    def name(cls) -> str:
+        if uid := cls._uid_from_env():
+            return pwd.getpwuid(uid).pw_name
```
gdiscry, line 248:
```suggestion
            return pwd.getpwuid(int(uid)).pw_name
```

gdiscry (commented):
There are a bunch of issues.

When running with sudo:
```
Python 3.10.10 (main, Mar  5 2023, 22:26:53) [GCC 12.2.1 20230201] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from mkosi.util import InvokingUser
>>> InvokingUser.uid_gid()
(<function InvokingUser._uid_from_env at 0x7f3a14eb2950>, 1000)
>>> InvokingUser.name()
  File "/home/georges/Projects/mkosi/mkosi/util.py", line 248, in name
    return pwd.getpwuid(uid).pw_name
TypeError: uid should be integer, not str
>>> InvokingUser.home()
PosixPath('/root')
```

When running with pkexec:
```
Python 3.10.10 (main, Mar  5 2023, 22:26:53) [GCC 12.2.1 20230201] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from mkosi.util import InvokingUser
>>> InvokingUser.uid_gid()
  File "/home/georges/Projects/mkosi/mkosi/util.py", line 241, in uid_gid
    gid = int(os.getenv("SUDO_GID") or pwd.getpwuid(uid).pw_gid)
TypeError: uid should be integer, not function
>>> InvokingUser.name()
  File "/home/georges/Projects/mkosi/mkosi/util.py", line 248, in name
    return pwd.getpwuid(uid).pw_name
TypeError: uid should be integer, not str
>>> InvokingUser.home()
PosixPath('/root')
```

When running as the current user:
```
Python 3.10.10 (main, Mar  5 2023, 22:26:53) [GCC 12.2.1 20230201] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from mkosi.util import InvokingUser
>>> InvokingUser.uid_gid()
  File "/home/georges/Projects/mkosi/mkosi/util.py", line 241, in uid_gid
    gid = int(os.getenv("SUDO_GID") or pwd.getpwuid(uid).pw_gid)
TypeError: uid should be integer, not function
```

behrmann:
Dropped a followup in #1512

PR #1501, opened 2023-04-24
Various fixes

DaanDeMeyer:


DaanDeMeyer:
Merging to unblock systemd and since aside from the revert these are all pretty trivial anyway

PR #1506, opened 2023-04-25
Add preset support

DaanDeMeyer:
Presets can be defined in mkosi.presets/. A preset is just like a
regular config file/directory, except that mkosi can build multiple
presets sequentially.

If mkosi.presets/ exists, for each preset mkosi will read the global
configuration, followed by the individual preset configuration. It
will then build each of the presets in alpha-numerical order. Later
presets can use outputs of earlier presets, specifically using the
BaseTrees= and Initrds= options.

While this has many use cases, one promising use case is to allow
building an initrd and a final image that uses that initrd within a
single invocation of mkosi.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2028,7 +2028,7 @@ def generate_secure_boot_key(args: MkosiArgs) -> None:
     cn = expand_specifier(args.secure_boot_common_name)
 
     for f in ("mkosi.secure-boot.key", "mkosi.secure-boot.crt"):
-        if f and not args.force:
+        if Path(f).exists() and not args.force:
             die(f"{f} already exists",
                 hint=("To generate new secure boot keys, "
                       "first remove mkosi.secure-boot.key and mkosi.secure-boot.crt"))
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -576,10 +576,16 @@ def install_build_dest(state: MkosiState) -> None:
         copy_path(install_dir(state), state.root, preserve_owner=False)
 
 
+def gzip_binary() -> str:
+    return "pigz" if shutil.which("pigz") else "gzip"
+
+
 def compressor_command(compression: Compression, src: Path) -> list[PathString]:
     """Returns a command suitable for compressing archives."""
 
-    if compression == Compression.xz:
+    if compression == Compression.gz:
+        return [gzip_binary(), "--fast", src]
+    elif compression == Compression.xz:
         return ["xz", "--check=crc32", "--fast", "-T0", src]
     elif compression == Compression.zst:
         return ["zstd", "-q", "-T0", "--rm", src]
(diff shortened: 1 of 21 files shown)


mkosi/config.py
```diff
@@ -1697,23 +1710,51 @@ def parse(self, argv: Optional[Sequence[str]] = None) -> tuple[MkosiArgs, MkosiC
...
-            if s.default_factory:
-                default = s.default_factory(namespace)
-            elif s.default is None:
-                default = s.parse(s.dest, None, namespace)
-            else:
-                default = s.default
+                with chdir(p if p.is_dir() else Path.cwd()):
+                    self.parse_config(p if p.is_file() else Path("."), cp)
+
+                self.parse_config(Path("."), cp)
```
behrmann, line 1726: I think this section could do with a few comments. The idea is that we pop into the profile dir, parse the profile config and then pop back and parse our global config there?

mkosi/config.py
```diff
@@ -1697,23 +1710,51 @@ def parse(self, argv: Optional[Sequence[str]] = None) -> tuple[MkosiArgs, MkosiC
...
+
+                name = p.name.lstrip(string.digits + "-").removesuffix(".conf")
```
behrmann, line 1728: This turns up a few times, let's factor it out
```python
def profilename_from_filename(name: str):= -> str:
    return name.lstrip(string.digits + "-").removesuffix(".conf")
```

mkosi/__init__.py
```diff
@@ -1130,75 +1130,77 @@ def print_summary(args: MkosiArgs, config: MkosiConfig) -> None:
...
+
+    {bold("DISTRIBUTION")}:
+                Distribution: {bold(config.distribution.name)}
+                    Release: {bold(none_to_na(config.release))}
+                Architecture: {config.architecture}
+                        Mirror: {none_to_default(config.mirror)}
+        Local Mirror (build): {none_to_none(config.local_mirror)}
+    Repo Signature/Key check: {yes_no(config.repository_key_check)}
+                Repositories: {",".join(config.repositories)}
+                    Initrds: {",".join(os.fspath(p) for p in config.initrds)}
```
behrmann, line 1147: The alignment is off here.

behrmann (commented):
Really looking forward to this. :)

mkosi/config.py
DaanDeMeyer: This was accidentally duplicated, dropped

mkosi/__init__.py
```diff
@@ -1574,6 +1567,14 @@ def run_build_script(state: MkosiState) -> None:
     if state.config.install_dir:
         empty_directory(state.config.install_dir)
 
+    # Create a few necessary mount points inside the build overlay.
```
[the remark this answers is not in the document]
DaanDeMeyer: With that change, we don't create the mount points anymore if there are no build packages (which is wrong, we still need to create them), so I moved the logic here to make sure the mountpoints are always created.

bluca (approved):
There's some static analysis warnings to fix or dismiss

PR #1511, opened 2023-04-25
Drop shell workaround in bwrap() and run_workspace_command()

DaanDeMeyer:
Let's use --perms and --chmod to fix the permissions of /tmp, /var/tmp and /dev/shm instead of our hacky shell workaround. Let's also drop all usage of shlex.join() in run() since it doesn't really by us anything.

This required bubblewrap 0.5.0.

### DIFF
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -9,7 +9,6 @@ import multiprocessing
 import os
 import pwd
 import queue
-import shlex
 import shutil
 import signal
 import subprocess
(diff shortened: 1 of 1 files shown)


bluca:
Can we postpone this for a month? Debian 12 releases in June

DaanDeMeyer:
Gentoo failure is unrelated, let's merge this now that Bookworm has been released. CentOS Stream 9 still ships bubblewrap 0.4.0 but it's trivial to build it from source so that shouldn't be a huge issue.

PR #1512, opened 2023-04-25
Unbreak Invokinguser

behrmann:


DaanDeMeyer (approved):

mkosi/util.py
gdiscry, line 1: I can't find how to leave a comment on a unmodified line…

To be consistent with `InvokingUser.uid_gid()`, I feel like `InvokingUser.name()` should use:

```python
if (uid := cls._uid_from_env()) is not None:
    return os.getenv("SUDO_USER") or pwd.getpwuid(uid).pw_name
```

That way, we use the same condition and we are consistent with our use of the `SUDO_*` environment variables.

mkosi/util.py
```diff
@@ -250,8 +251,7 @@ def name(cls) -> str:
 
     @classmethod
     def home(cls) -> Path:
-        home = os.getenv("SUDO_HOME") or os.getenv("HOME") or pwd.getpwuid(cls.uid()).pw_dir
-        return Path(home)
+        return Path(f"~{cls.name()}").expanduser()
```
gdiscry, line 254: When sudo/pkexec are not used, I feel like `Path.home()` (which is just `Path("~").expanduser()`) should be used (taking `$HOME` into account and not going through `pwd.getpwnam()`).

Maybe `name()` and `home()` could share something like:

```python
    @classmethod
    def _name_from_env(cls) -> str:
        if (uid := cls._uid_from_env()) is not None:
            return os.getenv("SUDO_USER") or pwd.getpwuid(uid).pw_name
        return ""

    @classmethod
    def name(cls) -> str:
        return cls._name_from_env() or getpass.getuser()

    @classmethod
    def home(cls) -> Path:
        return Path(f"~{cls._name_from_env()}").expanduser()
```

mkosi/util.py
```diff
@@ -228,17 +228,18 @@ def flatten(lists: Iterable[Iterable[T]]) -> list[T]:
...
     def uid(cls) -> int:
-        return int(cls._uid_from_env() or os.getuid())
+        return cls._uid_from_env() or os.getuid()
 
     @classmethod
     def uid_gid(cls) -> tuple[int, int]:
-        if uid := cls._uid_from_env:
-            gid = int(os.getenv("SUDO_GID") or pwd.getpwuid(uid).pw_gid)
+        if (uid := cls._uid_from_env()) is not None:
+            gid = int(os.getenv("SUDO_GID", pwd.getpwuid(uid).pw_gid))
```
gdiscry, line 242: Using `int(os.getenv("SUDO_GID") or pwd.getpwuid(uid).pw_gid)` meant that `pwd.getpwuid()`is not called when `$SUDO_GID` is set.

mkosi/util.py
```diff
@@ -228,17 +228,18 @@ def flatten(lists: Iterable[Iterable[T]]) -> list[T]:
 
 class InvokingUser:
     @staticmethod
-    def _uid_from_env() -> Optional[str]:
-        return os.getenv("SUDO_UID") or os.getenv("PKEXEC_UID")
+    def _uid_from_env() -> Optional[int]:
+        uid = os.getenv("SUDO_UID") or os.getenv("PKEXEC_UID")
+        return int(uid) if uid is not None else None
```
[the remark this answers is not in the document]
behrmann: No, let's be explicit here and reduce possible confusion for readers between `None`, `0` and `"0"`.

mkosi/util.py
behrmann: The relevant catch here is that `uid` could be `None`. Curious that mypy didn't flag this, since it gave me grief in the other function.

mkosi/util.py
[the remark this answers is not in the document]
gdiscry: My bad, I thought `uid` it would be `""` and that int("") would fail… 🤦

mkosi/util.py
gdiscry: Yes, my comment was for both lines

behrmann: There's a followup in #1514, PTAL.

PR #1513, opened 2023-04-25
More man page adjustments

keszybz:


mkosi.md
```diff
@@ -1360,15 +1358,15 @@ Format=disk
...
 #!/bin/sh
 cd $SRCDIR
 ./autogen.sh
 ./configure --prefix=/usr
 make -j `nproc`
 make install
 EOF
-# chmod +x mkosi.build
+$ chmod +x mkosi.build
 # mkosi --incremental boot
```
behrmann, line 1370: That should also work as normal user, no?

behrmann (approved):

DaanDeMeyer:
The last commit isn't entirely correct, we actually boot with 2 serial consoles, one regular one for the bootloader and a virtio serial console for systemd. But maybe we should go back to a single one and hardcode console=ttyS0 in the kernel arguments.

mkosi.md
keszybz: I thought that this uses systemd-nspawn which requires root?

DaanDeMeyer: Exactly this still needs root

keszybz:
> The last commit isn't entirely correct, we actually boot with 2 serial consoles, one regular one for the bootloader and a virtio serial console for systemd. But maybe we should go back to a single one and hardcode console=ttyS0 in the kernel arguments.

I wanted to express that it all appears on the caller's terminal. (What happens inside of the VM is a separate story). Do you have a suggestion how to phrase this better without going into too much detail?

mkosi.md
behrmann: My bad, when reading this I mixed it with the line above and read `build` instead of `boot`.

behrmann:
The best solution would be a footnote. A lie that's easier to understand with an accompanying deeper explanation hiding the internals that can be skipped or changed when they change, would maybe be helpful.

DaanDeMeyer:
@keszybz #1506 now contains a commit that drops the virtio-serial console in favor of only a regular serial console. The virtio-serial console is too flaky since the virtio_console module is not built in on every distro and edk2 does not support virtio-serial yet either so it makes debugging early boot stuff too complicated. Once that commit is merged this can go in as is.

PR #1516, opened 2023-04-26
Fix signing of checksums, especially for non-root user

mcassaniti:
Signing is failing when using the below configuration snippet.

```ini
[Validation]
Checksum   = yes
Sign       = yes
Key        = ...
```

This is occurring for two reasons:

  * The option `--default-key` needs to be specified before the file to sign
  * The keyring path is incorrect when running as a non-root user

I'm happy to split these patches into two PRs. I'm also happy to take any feedback since I'm munging environment variables and there's likely a nicer way to get this done.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -848,7 +848,21 @@ def calculate_signature(state: MkosiState) -> None:
             state.staging / state.config.output_checksum.name,
         ]
 
-        run(cmdline)
+        run(
+            cmdline,
+            # Do not output warnings about keyring permissions
+            stderr=subprocess.DEVNULL,
+            env={
+                # Set the path of the keyring to use based on the environment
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -848,7 +848,21 @@ def calculate_signature(state: MkosiState) -> None:
...
+            cmdline,
+            # Do not output warnings about keyring permissions
+            stderr=subprocess.DEVNULL,
+            env={
+                # Set the path of the keyring to use based on the environment
+                # if possible and fallback to the default path. Without this the
+                # keyring for the root user will instead be used which will fail
+                # for a non-root build.
+                'GNUPGHOME': os.environ.get(
+                    'GNUPGUHOME',
```
DaanDeMeyer, line 861: Typo here?

behrmann (approved):

PR #1518, opened 2023-04-26
Store file hash and signature in output directory

mcassaniti:
Every other output file is stored within the output directory from what I could see, so the hash and signature should be stored there too.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -608,11 +608,11 @@ class MkosiConfig:
 
     @property
     def output_checksum(self) -> Path:
-        return Path("SHA256SUMS")
+        return build_auxiliary_output_path(self, ".SHA256SUMS")
 
     @property
     def output_signature(self) -> Path:
-        return Path("SHA256SUMS.gpg")
+        return build_auxiliary_output_path(self, ".SHA256SUMS.gpg")
 
     @property
     def output_sshkey(self) -> Path:


DaanDeMeyer:
Can you rebase? CI was broken and was fixed

mcassaniti:
The one failing check is a transient issue where a package failed to be fetched.

mkosi/config.py
```diff
@@ -608,11 +608,11 @@ def output_nspawn_settings(self) -> Path:
 
     @property
     def output_checksum(self) -> Path:
-        return Path("SHA256SUMS")
+        return with_output_dir(self, "SHA256SUMS")
```
DaanDeMeyer, line 611: How about we use `build_auxiliary_output_path()` for these as well? Otherwise we're going to run into issues when using the same output directory for multiple images

DaanDeMeyer (changes requested):

mkosi/config.py
mcassaniti: Happy to do that, but that would then generate files like `MyImage_123.SHA256SUMS` or something close to it. If that is OK then let's do it. I'll need to adjust the check for `SHA256SUMS` too.

DaanDeMeyer: Yeah I think its fine to generate files like that. Let's go with this approach, it keeps things consistent with the naming of the other output files

DaanDeMeyer (approved):

PR #1520, opened 2023-04-27
Support additional keyrings on Debian derivatives

mcassaniti:
If `mkosi` is run with additional repositories (i.e.: `mkosi.reposdir`), the specified repositories should include a path to their own keyring used for verification of the repository. Normally the keyrings are stored under `/etc/apt/keyrings` although `/usr/share/keyrings` could also be extended.

If the directory `/etc/apt/keyrings` is found then it will be bind-mounted into place for `apt-get` to use. The directory would normally be populated with `mkosi.skeleton` or similar.

If the directory `/usr/share/keyrings` is found then the contents of the host directory `/usr/share/keyrings` and the skeleton directory are combined. This prevents issues during bootstrap where required keys are not within the skeleton directory.

### DIFF
--- a/mkosi/distributions/debian.py
+++ b/mkosi/distributions/debian.py
@@ -3,10 +3,12 @@
 import shutil
 import tempfile
 from collections.abc import Sequence
+from contextlib import ExitStack
 from pathlib import Path
 from textwrap import dedent
 
 from mkosi.distributions import DistributionInstaller
+from mkosi.mounts import mount
 from mkosi.run import bwrap, run
 from mkosi.state import MkosiState
 from mkosi.types import CompletedProcess, PathString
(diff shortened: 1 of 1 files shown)


mkosi/distributions/debian.py
```diff
@@ -2,11 +2,13 @@
 
 import shutil
 import tempfile
-from collections.abc import Sequence
+from collections.abc import Iterator, Sequence
```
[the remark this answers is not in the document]
mcassaniti: Fixed.

mcassaniti:
Not sure how to fix the unit test issue. Considering the results of `invoke_apt()` are never used, can the function instead return `None` and we move on?

mcassaniti:
Thanks. I'll add to the conversations as appropriate and see what I can bring to the table.

mkosi/distributions/debian.py
```diff
@@ -224,7 +227,27 @@ def invoke_apt(
...
+            shutil.copytree("/usr/share/keyrings", keydir, dirs_exist_ok=True)
+
+            # Copy the additonal keyrings
+            shutil.copytree(extra_usr_keyrings, keydir, dirs_exist_ok=True)
+
+            # Create a bind mount using the combined key directory
+            # OverlayFS is not used because it is unknown which partitions data resides on
+            stack.enter_context(mount(keydir, Path("/usr/share/keyrings"), options=['bind']))
+
+        return bwrap(["apt-get", operation, *extra], apivfs=state.root if apivfs else None, env=env | state.environment)
```
DaanDeMeyer, line 250: No need to muck around with bind mounts, we have full control over where apt searches for the keyring via its config parameters `Dir::etc::trusted` and `Dir::etc::trustedparts`.

mcassaniti: Maybe not. Take a look (here)[https://github.com/systemd/mkosi/pull/1490#issuecomment-1524984015]. I've added the details merely for clarity here. This PR is likely to get closed.

mcassaniti:
After looking at the man page for sources.list and based on my own use case, I'd highly recommend using the deb822 format and inlining repository signing keys.

Issue #1522, opened 2023-04-28
Support setting FS creation options


mcassaniti:
I don't know if this should be raised here or a change made to `systemd-repart`. I'd like to set specific file system creation options and I know at present these can be set with `SYSTEMD_REPART_MKFS_OPTIONS_<FSTYPE>` environment variables. The easiest solution I can see is to extend the environment that `systemd-repart` is fed by also checking the running environment (i.e.: look for keys starting with `SYSTEMD_REPART_MKFS_OPTIONS_<FSTYPE>` in `os.environ`).

While we're at it, I'd also like to set the sector size that `systemd-repart` uses through an environment variable or another configuration option. The size of a verity partition varies wildly based on a 512 byte or 4096 byte sector size.

DaanDeMeyer:
Sounds good, but don't check `os.environ`, look in `state.environment`.

For the sector size, it's one of those options that i'd like to be able to pass around without needing explicit support in mkosi. I've been floating around the idea of allowing to specify "build" credentials which we'd put in a directory and set `CREDENTIALS_DIRECTORY` so that systemd tooling can pick them up from there. Still need to implement that though and discuss with the others whether that's the way to go

mcassaniti:
It would be wise to raise an RFE for `CREDENTIALS_DIRECTORY` and at least document what you're thinking of. I'll close this one off.

PR #1527, opened 2023-05-01
repart: Support mkfs options from configured environment

mcassaniti:
Covers #1522.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1478,6 +1478,10 @@ def invoke_repart(state: MkosiState, skip: Sequence[str] = [], split: bool = Fal
     for fs, options in state.installer.filesystem_options(state).items():
         env[f"SYSTEMD_REPART_MKFS_OPTIONS_{fs.upper()}"] = " ".join(options)
 
+    for option, value in state.environment.items():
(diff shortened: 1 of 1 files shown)


mcassaniti:
I cannot fathom how I've managed to break the type checking.

mkosi/__init__.py
```diff
@@ -1478,6 +1478,10 @@ def invoke_repart(state: MkosiState, skip: Sequence[str] = [], split: bool = Fal
     for fs, options in state.installer.filesystem_options(state).items():
         env[f"SYSTEMD_REPART_MKFS_OPTIONS_{fs.upper()}"] = " ".join(options)
 
+    for option, value in state.environment.items():
+        if option[0:28] == "SYSTEMD_REPART_MKFS_OPTIONS_":
```
DaanDeMeyer, line 1482: Use startswith() here.

DaanDeMeyer (changes requested):

mkosi/__init__.py
mcassaniti: Now I finally see why I broke the type checker when using `startswith()`.

PR #1528, opened 2023-05-01
Mount package cache for all scripts

mcassaniti:
A script may need to add additional packages at any time, even during post-install. As a result it is worth the effort to mount the package cache during every execution of a user provided script.

As an example, here is what an adjusted systemd build would look like under Ubuntu:

mkosi.build:
```shell
# Install build dependencies for systemd
mkdir /var/tmp/systemd && cd /var/tmp/systemd
pull-lp-source systemd $(lsb_release -cs)
cd $(find -mindepth 1 -maxdepth 1 -type d)
/usr/lib/pbuilder/pbuilder-satisfydepends-apt --control ../*.dsc

# Enable sysupdate support
sed -i s/"sysupdate=false"/"sysupdate=true"/ debian/rules

# Bump version
dch --increment "Enabled additional options"

# Build source
dpkg-buildpackage --build=source --unsigned-source --unsigned-changes --no-check-builddeps

# Build binary
dpkg-buildpackage --build=binary --unsigned-source --unsigned-changes

# Prune and copy packages
rm ../systemd-standalone*.deb
mv ../*.deb "$DESTDIR/"
```

mkosi.postinst:
```shell
# There are no additional dependencies installed, but just pretend there are.
apt-get install --yes -o Dpkg::Progress-Fancy="0" /*.deb
```

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -393,6 +393,9 @@ def run_prepare_script(state: MkosiState, build: bool) -> None:
         "--chdir", "/root/src",
     ]
 
+    if state.installer.cache_path():
+        bwrap.extend(["--bind", state.cache, cast(Path, state.installer.cache_path())])
+
     def clean() -> None:
         srcdir = state.root / "root/src"
         if srcdir.exists():
@@ -430,6 +433,9 @@ def run_postinst_script(state: MkosiState) -> None:
         bwrap: list[PathString] = [
             "--bind", state.config.postinst_script, "/root/postinst",
         ]
+
+    if state.installer.cache_path():
+        bwrap.extend(["--bind", state.cache, cast(Path, state.installer.cache_path())])
 
         run_workspace_command(state.root, ["/root/postinst", "final"], bwrap_params=bwrap,
                               network=state.config.with_network, env=state.environment)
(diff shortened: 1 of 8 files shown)


DaanDeMeyer:
I would much prefer that we always use the host package manager to install packages. We could instead mount a `PACKAGEDIR` directory when we run the prepare script and we'll install any file we find in that directory after we run the prepare script. If the file is empty, we'll treat it as a package to install from the repositories.

Then you can build your modified deb in the prepare script, put it in `PACKAGEDIR` and it'll be installed by mkosi afterwards.

mcassaniti:
So how would we get build dependencies installed using `apt-get build-dep` or using `/usr/lib/pbuilder/pbuilder-satisfydepends-apt`?

I do like the idea of having a `PACKAGEDIR` that can install arbitrary package archives. That covers a second use case I have.

DaanDeMeyer:
I would add a separate option for installing build dependencies of packages (BuildDependencies= or so). It seems to be supported by apt, dnf, asp (arch) and zypper so we'd just need to implement it for all distros.

mcassaniti:
I've added support for `PACKAGEDIR` under #1531. This covers one of the two issues. There's a bit of a problem adding packages for the build script. These could either be invoked as something like `apt-get -y build-dep systemd` or `apt-get -y build-dep systemd_v253.dsc`. The second is a file path. I'm not familiar with how yum, dnf and zypper handle these sorts of things (yet), but I wanted to make sure I covered both cases.

DaanDeMeyer:
I think we can close this one as we've that we're going to pursue a different solution

PR #1529, opened 2023-05-01
Ensure output is compatible with systemd-sysupdate

mcassaniti:
When using `systemd-sysupdate` it is wise to use split artifacts and to compress artifacts. These changes ensure that split artifacts are compressed and the file paths are stored in the checksum.

Please note that `systemd-sysupdate` expects a checksum name of `SHA256SUMS` and uses the contained file names to determine which files to download. The compression suffix is also used to determine how to decompress files.

Additional CI checks have been added to make sure no regressions in output format occur.

These changes attempt to closely match what was and was not compressed before the move to `systemd-repart`.

The manual page has also been regenerated.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -178,6 +178,19 @@ jobs:
         Format=${{ matrix.format }}
         EOF
 
+    # Make sure systemd-repart and split artifacts generates a usable output for systemd-sysupdate
+    - name: Configure ${{ matrix.distro }}/${{ matrix.format }} repart
+      if: matrix.format == 'disk'
+      run: |
+        tee mkosi.conf.d/01-repart.conf <<- EOF
+        [Output]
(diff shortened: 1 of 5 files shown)


bluca:
needs a rebase

keszybz (commented):
Looks reasonable and the CI passes. The CI doesn't always cover all aspects of output naming, but let's merge this and test it out in the wild.

DaanDeMeyer:
I really don't think this is the way to go, we should look into having systemd-sysupdate look for files like we write them.

mcassaniti:
> I really don't think this is the way to go, we should look into having systemd-sysupdate look for files like we write them.

I'm more than happy to make changes after we've got a rough design ready to go. Do take a look at the man page for sysupdate.d. The design personally makes a lot of sense. If there's holes in it or a better way then I'm all for reworking this.

Issue #1224, opened 2022-10-09
Incompatible image name with `machinectl`


netzego:
The generated image name when used with `mkosi.version` is incompatible to `machinectl`. `mkosi` uses an `_` (underscore) as delimiter which is not allowed in valid hostnames. If one copies an image `image_0.1.raw` to `/var/lib/machines` and attempt to start it via `machinectl start image_0.1` an error occurs: `Invalid machine name image_0.1.`.

Starting the image via `systemd-nspawn` works, and `machinectl list` shows that nspawn simply discard the invalid character. So the name of that machine/image would be `image0.1`.

When using `mkosi boot` the hostname (`/etc/hostname`) of that machine is used by `machinectl` -- which is fine, imo.

Any how, i consider this as a naming inconvenience but not a *bug*. So feel free to *just* notice and close this issue. Thanks for this neat tool!

keszybz:
Hmm, shouldn't `machined` and `nspawn` ignore `_*` in image names? It seems a bit of a bummer to have this incompatibility. @poettering, opinions?

DaanDeMeyer:
Fixed by #1739

Issue #1525, opened 2023-04-29
subvol: output missing!


iam-TJ:
After an apparently successful build the output is missing. For the config I have:
```
$ ls -latr
total 36
drwxrwx--- 3 tj   adm 4096 2021-04-27 10:46 mkosi.extra
lrwxrwxrwx 1 tj   adm   16 2021-08-10 11:12 mkosi.cache -> /var/cache/mkosi
drwxr-xr-x 3 tj   tj  4096 2023-04-22 14:39 ..
lrwxrwxrwx 1 tj   tj    12 2023-04-22 14:39 mkosi.default.d -> mkosi.conf.d
-rw-rw---- 1 tj   adm   62 2023-04-24 13:50 mkosi.nspawn
drwxrwx--- 3 tj   adm 4096 2023-04-24 14:03 mkosi.skeleton
-rwxrwx--- 1 tj   adm  224 2023-04-24 14:32 mkosi.postinst
drwxr-xr-x 2 tj   tj  4096 2023-04-24 16:25 match
lrwxrwxrwx 1 tj   tj    17 2023-04-25 22:07 mkosi.output -> /var/lib/machines
drwxr-xr-x 2 tj   tj  4096 2023-04-29 11:04 mkosi.conf.d
-rwxr-xr-x 1 tj   tj   281 2023-04-29 13:45 mkosi.finalize
drwxr-xr-x 7 tj   tj  4096 2023-04-29 13:45 .
drwxrwxrwx 1 root adm    0 2023-04-29 14:13 mkosi.workspace
```
Summary:
```
DISTRIBUTION
              Distribution: debian
                   Release: bullseye
              Architecture: x86_64
                    Mirror: http://deb.debian.org/debian
      Local Mirror (build): none
  Repo Signature/Key check: yes
              Repositories: 
                   Initrds: 

OUTPUT:
                  Image ID: debian-bullseye-amd64
             Image Version: 0.2
             Output Format: subvolume
          Manifest Formats: json
          Output Directory: /home/tj/iam/mkosi/debian-amd64-base/mkosi.output
       Workspace Directory: /home/tj/iam/mkosi/debian-amd64-base/mkosi.workspace
                    Output: /home/tj/iam/mkosi/debian-amd64-base/mkosi.output/debian-bullseye-amd64_0.2
           Output Checksum: n/a
          Output Signature: n/a
[...]

PR #1530, opened 2023-05-01
fix: prefix must be valid machine name

iam-TJ:
Due to strict rules checks by machinectl using validate_hostname() underscores are invalid in machine names.

Noted in issue #1525.

Reported documentation error in
https://github.com/systemd/systemd/issues/27481

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -1964,6 +1964,11 @@ def load_config(args: argparse.Namespace) -> MkosiConfig:
             output = prefix
         args.output = Path(output)
 
+    if "_" in str(args.output) and args.output_format == OutputFormat.subvolume:
+        old = str(args.output)
+        args.output = args.output.with_name(args.output.name.replace("_", "-"))
+        logging.warning(f"subvolume (systemd-machined) names cannot contain underscore(_); converting to hypen (-). {old} -> {args.output} ")
+
     if args.output_dir is not None:
         if "/" not in str(args.output):
             args.output = args.output_dir / args.output


iam-TJ:
This also resolves issue #1224

mkosi/config.py
```diff
@@ -1964,6 +1964,11 @@ def load_config(args: argparse.Namespace) -> MkosiConfig:
             output = prefix
         args.output = Path(output)
 
+    if "_" in str(args.output) and args.output_format in [OutputFormat.subvolume]:
```
behrmann, line 1967: Since it's only one and will most likely stay like that.
```suggestion
    if "_" in str(args.output) and args.output_format == OutputFormat.subvolume:
```

mkosi/config.py
```diff
@@ -1964,6 +1964,11 @@ def load_config(args: argparse.Namespace) -> MkosiConfig:
...
+        old = str(args.output)
+        args.output = old.replace("_", "-")
```
behrmann, line 1969: This would leave `args.output` as string and a few lines below we call `args.output.absolute()`, which would then fail. You maybe want something similar to
```suggestion
        args.output = args.output.with_name(args.output.name.replace("_", "-"))
```

mkosi/config.py
iam-TJ: Thanks for the review. The use of a list here is that, originally, I developed this fix after/on top of another PR of mine  (#1535)- fixing subvolume creation which also reinstates read-only subvolume support by adding an additional `OutputFormat.subvolume_ro`   and initially posted this PR with that included in this list. When I realised I removed the `subvolume_ro` but left the list structure in the expectation that `subvolume_ro` may well exist soon! I can change this though, and move this list construct to the PR that adds `subvolume_ro`

mkosi/config.py
iam-TJ: Ooops! still getting up to speed with the intricacies of modern Python.

mkosi/config.py
```diff
@@ -1964,6 +1964,11 @@ def load_config(args: argparse.Namespace) -> MkosiConfig:
             output = prefix
         args.output = Path(output)
 
+    if "_" in str(args.output) and args.output_format == OutputFormat.subvolume:
+        old = str(args.output)
+        args.output = args.output.with_name(args.output.name.replace("_", "-"))
+        logging.warning(f"subvolume (systemd-machined) names cannot contain underscore(_); converting to hypen (-). {old} -> {args.output} ")
```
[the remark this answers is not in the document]
iam-TJ: I was confused by your first comment since I see:
```
‣ subvolume (systemd-machined) names cannot contain underscore(_); converting to hypen (-). debian_bullseye_amd64_0.2 -> debian-bullseye-amd64-0.2 
```
but then realised you are referring to `output` containing a path component, whereas in my scenario it is created from `ImageId` and `ImageVersion`. So, would need to do this only on the last component in the path.

Typo for "hyphen" - my brain and fingers have different ideas on spelling!

I'd rather it doesn't fail since the same configuration and naming may be used with different `output_format` s and a fail would involve having to change the config just for the `subvolume` build. E.g. in scenarios where builds use each format in turn, in a loop.

Also note that currently when `image_version` is added to `prefix` it inserts an underscore and I was trying to avoid changing that as well (originally I was replacing that underscore with "-v").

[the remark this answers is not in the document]
keszybz: OK. Then change the warning to a info or debug message.

DaanDeMeyer:
Replaced by https://github.com/systemd/mkosi/pull/1739 so let's close this one

Issue #1534, opened 2023-05-02
btrfs subvol: ReadOnly no longer possible due to repart


iam-TJ:
Since I cannot reopen #1484 I'm starting again!

Advice was to use the finalize script to mark the subvol read-only. Now I have subvol builds working again it turns out not to be possible since at the point the finalize script runs it can alter the content of workspace/root and the output subvol has not yet been created.

I added some reports to the finalize script to be clear on what it sees (`mkosi.worksapce` and `mkosi.output` are both bind-mounts of `/var/lib/machines`):
```
‣    Recording packages in manifest…
‣ + dpkg-query --admindir=/home/tj/iam/mkosi/debian-amd64-base/mkosi.workspace/.mkosi.tmps3e9zaip/root/var/lib/dpkg --show --showformat '${Package}\t${source:Package}\t${Version}\t${Architecture}\t${Installed-Size}\t${db-fsys:Last-Modified}\n'
‣    Resetting machine ID
‣    Running finalize script…
‣ + /home/tj/iam/mkosi/debian-amd64-base/mkosi.finalize
+ echo mkosi.finalize: $@='' BUILDROOT='/home/tj/iam/mkosi/debian-amd64-base/mkosi.workspace/.mkosi.tmps3e9zaip/root' OUTPUTDIR='/home/tj/iam/mkosi/debian-amd64-base/mkosi.output'
mkosi.finalize: $@='' BUILDROOT='/home/tj/iam/mkosi/debian-amd64-base/mkosi.workspace/.mkosi.tmps3e9zaip/root' OUTPUTDIR='/home/tj/iam/mkosi/debian-amd64-base/mkosi.output'
+ final
+ echo final: BUILDROOT=/home/tj/iam/mkosi/debian-amd64-base/mkosi.workspace/.mkosi.tmps3e9zaip/root OUTPUTDIR=/home/tj/iam/mkosi/debian-amd64-base/mkosi.output
final: BUILDROOT=/home/tj/iam/mkosi/debian-amd64-base/mkosi.workspace/.mkosi.tmps3e9zaip/root OUTPUTDIR=/home/tj/iam/mkosi/debian-amd64-base/mkosi.output
+ ls -latr /home/tj/iam/mkosi/debian-amd64-base/mkosi.output
total 116
lrwxrwxrwx  1 root root        31 Oct 10  2021 linux-builder -> /mnt/machines_old/linux-builder
[...]

PR #1535, opened 2023-05-02
subvolume: create and take advantage of copy-on-write

iam-TJ:
Ensure a subvolume is created.

When the output subvol and workspace/root are in the same BTRFS take advantage of copy-on-write and snapshots by making root a subvol and taking a snapshot of it for the output before deleting workspace.

Closes: https://github.com/systemd/mkosi/issues/1525
Closes: https://github.com/systemd/mkosi/issues/1534

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -331,7 +331,8 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
 
 : The image format type to generate. One of `directory` (for generating OS
   images inside a local directory), `subvolume` (similar, but as a btrfs
-  subvolume), `tar` (similar, but a tarball of the image is generated), `cpio`
+  subvolume), `subvolume_ro` (btrfs read-only subvolume),
+  `tar` (similar, but a tarball of the image is generated), `cpio`
   (similar, but a cpio archive is generated), `disk` (a block device image
   with a GPT partition table).
 
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -94,9 +94,32 @@ def btrfs_subvol_create(path: Path, mode: int = 0o755) -> None:
         run(["btrfs", "subvol", "create", path])
 
 
+def btrfs_subvol_snapshot(source: Path, destination: Path, mode: int = 0o755) -> None:
+    with set_umask(~mode & 0o7777):
+        run(["btrfs", "subvol", "snapshot", source, destination])
+
+
 @contextlib.contextmanager
 def mount_image(state: MkosiState) -> Iterator[None]:
     with complete_step("Mounting image…", "Unmounting image…"), contextlib.ExitStack() as stack:
+
+        @stack.callback
+        def subvolume_create() -> None:
+            if state.output_format == OutputFormat.subvolume:
+                if state.btrfs_snapshot:
+                    logging.debug("maybe_btrfs_snapshot() will take subvolume snapshot")
+                    if not state.root.exists():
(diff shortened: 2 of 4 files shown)


mkosi.md
```diff
@@ -331,7 +331,8 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
 
 : The image format type to generate. One of `directory` (for generating OS
   images inside a local directory), `subvolume` (similar, but as a btrfs
-  subvolume), `tar` (similar, but a tarball of the image is generated), `cpio`
+  subvolume), `subvolume_ro` (btrfs read-only subvolume),
```
behrmann, line 334: What's the reasoning behind adding `subvolume_ro`? Couldn't this be exchanged for just doing `subvolume` and have people who need that set the read-only flag in a finalise script? I'm asking because the extra tracking of two different output formats seems like added complexity that might be good to avoid.

DaanDeMeyer (changes requested):
This is not the way to go, we should just always try to snapshot and if it fails fall back to a regular copy.

In fact, I don't think we should have a subvolume output format. We should just automatically create them if we're on btrfs. I would modify copy_path() to try to snapshot the source if the to the destination if the destination does not exist yet. If the destination already exists, I would try to btrfs send receive the source into the destination. If that fails because the source is not a subvolume, then we do a regular copy.

iam-TJ:
Currently `copy-path()` isn't called at all when using `subvolume` which is part of the problem with recent changes having broken subvolume completely as reported in #1525

Extract from a build log using:
```
$ git l -n 4
38e6134 2023-05-03 11:54:48 +0100 N Tj fix: prefix must be valid machine name
ca8a943 2023-05-02 10:22:48 +0200 N Tj distro: debian: allow RemovePackages
3008c7e 2023-05-01 13:16:54 +0200 E Daan De Meyer Merge pull request #1513 from keszybz/more-man
be290f2 2023-05-01 13:15:33 +0200 E Daan De Meyer Merge pull request #1506 from DaanDeMeyer/profiles
```
shows no call to `copy_path()` when building a subvolume:
```
...
$ grep copy_path /tmp/mkosi/mkosi.36.log 
‣ copy_path('/home/tj/iam/mkosi/debian-amd64-base/mkosi.skeleton', '/home/tj/iam/mkosi/debian-amd64-base/mkosi.workspace/.mkosi.tmpzzhk_1bd/root, 'False', 'False')
‣ copy_path('/home/tj/iam/mkosi/debian-amd64-base/mkosi.extra', '/home/tj/iam/mkosi/debian-amd64-base/mkosi.workspace/.mkosi.tmpzzhk_1bd/root, 'False', 'False')
‣ copy_path('/home/tj/iam/mkosi/debian-amd64-base/mkosi.nspawn', '/home/tj/iam/mkosi/debian-amd64-base/mkosi.workspace/.mkosi.tmpzzhk_1bd/staging/debian-bullseye-amd64-0.2.nspawn, 'False', 'True')
```

iam-TJ:
On 04/05/2023 09:24, Daan De Meyer wrote:
> Yes, so we shouldn't have a subvolume output format, we should have a --subvolume option, which makes us use subvolumes for certain directories:

That's a major change so I'll leave that to you.

I've been trying to re-instate broken functionality in order to continue 
working with existing project definitions.

DaanDeMeyer:
Let's close this, I have an implementation of --subvolume locally already, will publish it after #1538 is merged.

PR #1536, opened 2023-05-03
opensuse: Use dnf instead of zypper

DaanDeMeyer:
Let's reduce the number of package managers we use by using dnf to
build opensuse images.

The only caveat is that extra GPG keys are listed in repomd.xml for
the opensuse repos, so we have to download that file and parse it to
figure out the extra keys that we have to write to the repo configs.

### DIFF
--- a/mkosi/distributions/fedora.py
+++ b/mkosi/distributions/fedora.py
@@ -52,13 +52,12 @@ class FedoraInstaller(DistributionInstaller):
             # In other versions, the "fedora" repo is frozen at release, and "updates" provides any new packages.
             updates_url = None
 
-        gpgpath = Path(f"/etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-{releasever}-{state.config.architecture}")
         # See: https://fedoraproject.org/security/
         gpgurl = "https://fedoraproject.org/fedora.gpg"
 
-        repos = [Repo("fedora", release_url, gpgpath, gpgurl)]
+        repos = [Repo("fedora", release_url, gpgurl)]
         if updates_url is not None:
-            repos += [Repo("updates", updates_url, gpgpath, gpgurl)]
+            repos += [Repo("updates", updates_url, gpgurl)]
 
         setup_dnf(state, repos)
         invoke_dnf(state, "install", packages, apivfs=apivfs)
@@ -90,7 +89,6 @@ def url_exists(url: str) -> bool:
 class Repo(NamedTuple):
     id: str
     url: str
-    gpgpath: Path
     gpgurl: str
     enabled: bool = True
 
@@ -98,19 +96,13 @@ class Repo(NamedTuple):
 def setup_dnf(state: MkosiState, repos: Sequence[Repo] = ()) -> None:
     with state.workspace.joinpath("dnf.conf").open("w") as f:
         for repo in repos:
-
-            if repo.gpgpath.exists():
-                gpgkey = f"file://{repo.gpgpath}"
-            else:
-                gpgkey = repo.gpgurl
-
             f.write(
                 dedent(
                     f"""\
                     [{repo.id}]
                     name={repo.id}
                     {repo.url}
-                    gpgkey={gpgkey}
+                    gpgkey={repo.gpgurl}
                     gpgcheck=1
                     enabled={int(repo.enabled)}
                     """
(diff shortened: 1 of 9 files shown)


DaanDeMeyer:
> I'm not sure I like this, I don't even know if there's a working and reliable dnf implementation when doing native builds. And if there is and it doesn't work, it's going to be unsupported and tricky to fix

I've been in contact with someone from opensuse with regards to some question about the gpg keys, and he mentioned that there's quite a few people using dnf with opensuse. 

It even has a wiki page and everything: https://en.opensuse.org/SDB:DNF

mkosi/distributions/fedora.py
```diff
@@ -90,39 +89,33 @@ def url_exists(url: str) -> bool:
...
                 )
             )
 
+            ident = 0
+            for url in repo.gpgurls:
+                f.write(ident * " ")
+                f.write(f"{url}\n")
+                ident = len("gpgkey=")
+
+            f.write("\n")
```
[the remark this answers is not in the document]
DaanDeMeyer: Did it slightly different

DaanDeMeyer:
> But is it fully supported? I mean, can I rely on it when building an opensuse image on opensuse on OBS?

I wouldn't see why not? dnf is packaged in opensuse, and it's all just rpm under the hood.

DaanDeMeyer:
What can happen is that they decide to do one of these weird hacks like putting extra gpgkeys in repomd.xml and we'd have to add some extra logic to make that work with dnf likely. But given that it's literally the only thing we had to do to make dnf work I'd say the risk of that is pretty low.

mkosi/distributions/fedora.py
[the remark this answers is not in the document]
behrmann: That's also a good solution, though a comment will probably be clearer then recalculating the indentation every time
```suggestion
                # indent subquent keys to len("gpgkey=")
                f.write("gpgkey=" if not i else "       ")
```

mkosi/distributions/fedora.py
```diff
@@ -145,6 +145,11 @@ def invoke_dnf(
         "--noplugins",
     ]
 
+    # Make sure we download filelists so all dependencies can be resolved.
+    # See https://bugzilla.redhat.com/show_bug.cgi?id=2180842
```
keszybz, line 149: That bug seems quite confused. Did you encounter some issue with mkosi or is this just a precaution?

mkosi/distributions/opensuse.py
```diff
@@ -1,13 +1,13 @@
 # SPDX-License-Identifier: LGPL-2.1+
 
+import urllib.request
+import xml.etree.ElementTree as ET
```
keszybz, line 4: `import xml.etree.ElementTree as ElementTree` ? It is used exactly once, so this is just obfuscation.

mkosi/distributions/opensuse.py
```diff
@@ -34,74 +34,38 @@ def install_packages(cls, state: MkosiState, packages: Sequence[str], apivfs: bo
...
+def fetch_gpgurls(repourl: str) -> list[str]:
+    gpgurls = [f"{repourl}/repodata/repomd.xml.key"]
+
+    with urllib.request.urlopen(f"{repourl}/repodata/repomd.xml") as f:
+        xml = f.read().decode()
+        root = ET.fromstring(xml)
+
+        tags = root.find("{http://linux.duke.edu/metadata/repo}tags")
+        if not tags:
+            die("repomd.xml missing tags element")
```
keszybz, line 64: `<tags>`?

mkosi/distributions/fedora.py
keszybz: Yeah, I think this patch can be dropped. After the latest updates, it's just a bunch of fringe packages. `dnf5` is not even installed on older releases, so we don't have to worry so much about it.

DaanDeMeyer: @keszybz Does that include all the centos packages? I still need to build centos 8 with mkosi.

keszybz: Ah, OK. I didn't think about the case where you use `dnf5` from the host to install an older distro. So let's merge this, and then maybe we can skip it again on Fedora >= 38. But that can be done later.

PR #1538, opened 2023-05-04
Make sure a stable path to the output path exists

DaanDeMeyer:
Rework output related logic
- Let's store the output as a "base" name without any suffixes from
which we construct all other output paths.
- Let's not allow --output to be specified as a path anymore and
always put it in the configured output directory.
- Let's return all output paths as strings instead of paths.
- Let's always create a symlink from the "base" name to the full
output path so it can be referred to regardless of the output
format, compression or image version that's used.
- Let's make sure we compress split partitions if requested

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -738,6 +738,7 @@ def install_unified_kernel(state: MkosiState, roothash: Optional[str]) -> None:
                 "build",
             ])
 
+            unlink_output(args, presets[0])
             build_stuff(state.uid, state.gid, args, presets[0])
 
             initrds = [presets[0].output_compressed]
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -122,12 +122,6 @@ def mount_image(state: MkosiState) -> Iterator[None]:
         yield
 
 
-def prepare_tree_root(state: MkosiState) -> None:
-    if state.config.output_format == OutputFormat.subvolume:
-        with complete_step("Setting up OS tree root…"):
-            btrfs_subvol_create(state.root)
-
-
 def clean_paths(
         root: Path,
         globs: Sequence[str],
@@ -858,7 +852,7 @@ def hash_file(of: TextIO, path: Path) -> None:
 
 
 def calculate_sha256sum(state: MkosiState) -> None:
-    if state.config.output_format in (OutputFormat.directory, OutputFormat.subvolume):
+    if state.config.output_format == OutputFormat.directory:
         return None
 
     if not state.config.checksum:
@@ -949,7 +943,7 @@ def print_output_size(config: MkosiConfig) -> None:
     if not config.output_compressed.exists():
         return
 
-    if config.output_format in (OutputFormat.directory, OutputFormat.subvolume):
+    if config.output_format == OutputFormat.directory:
         log_step("Resulting image size is " + format_bytes(dir_size(config.output)) + ".")
     else:
         st = os.stat(config.output_compressed)
@@ -1736,7 +1730,7 @@ def acl_toggle_boot(config: MkosiConfig) -> Iterator[None]:
 def run_shell(args: MkosiArgs, config: MkosiConfig) -> None:
     cmdline: list[PathString] = ["systemd-nspawn", "--quiet"]
 
-    if config.output_format in (OutputFormat.directory, OutputFormat.subvolume):
+    if config.output_format == OutputFormat.directory:
         cmdline += ["--directory", config.output]
 
         owner = os.stat(config.output).st_uid
@@ -2129,7 +2123,6 @@ def run_verb(args: MkosiArgs, presets: Sequence[MkosiConfig]) -> None:
 
     if args.verb == Verb.qemu and last.output_format in (
         OutputFormat.directory,
-        OutputFormat.subvolume,
         OutputFormat.tar,
     ):
         die(f"{last.output_format} images cannot be booted in qemu.")
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -835,10 +835,6 @@ def compress_output(config: MkosiConfig, src: Path, uid: int, gid: int) -> None:
             run(["fallocate", "--dig-holes", src], user=uid, group=gid)
     else:
         with complete_step(f"Compressing output file {src}…"):
-            compressed_path = Path(f"{src}.{config.compress_output.value}")
-            if compressed_path.exists():
-                compressed_path.unlink()
-
             run(compressor_command(config.compress_output, src), user=uid, group=gid)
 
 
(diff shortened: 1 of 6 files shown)


bluca:
`FileExistsError: [Errno 17] File exists: '/home/runner/work/mkosi/mkosi/.github/image'`

keszybz:
`.exists()` returns False for dangling symlinks.

DaanDeMeyer:
@mcassaniti Aside from the SHA256SUMS thing, can you check if this works for your needs? I would make a PR for systemd-sysupdate to have it look for `f"{url}.SHA256SUMS" aside from f"{url}/SHA256SUMS" so that it also works with the way we output those files.

mcassaniti:
There's a lot in this PR and I personally would've split it up a bit. I struggled to separate out everything mentally.

  > Let's return all output paths as strings instead of paths.

Can I suggest the use of a new type for file names, something like `FileName = typing.NewType("FileName", str)`? This will assist with type checking. Now maybe it just isn't that important and we can live with strings instead, but if we want to take advantage of type checking then lets do it.

I'd really recommend putting a check somewhere to make sure the outputs don't regress. That was why I added the additional CI checks.

Instead of doing a full PR review I tried to build using the configuration below and got build failures. I can't even build from main at present, so I applied my changes to before `7f820b7 Do prebuilt initrds by default` to get the results below. I've raised #1540 to note the initrd build problem.

If the file names in `SHA256SUMS` matches what I've provided when you've done a build with the provided configuration, and the file names in `mkosi.output` match what is in `SHA256SUMS` then perfect.

```shell
$ cat mkosi.output/SHA256SUMS 
543f67185c70971b8571690cad92fe6cc64abff4f5f82158107652a3de2b9491 *Rescue_0.raw.zst
f7c5c7da1a472a176c77d8f768efdf088c5ce323a9a8ac3159022cf9dc446b0e *Rescue_0.efi.zst
b3ae271e206db1abf0f578a775db5c221fdc3aa37dc1fd660adae8bae71ec428 *Rescue_0.vmlinuz
d31939958eedde582ead4b18d7aeaca8dbbbe099d711e1d3b26bd3aca5bfa62d *Rescue_0.manifest.zst
480011aae1cb4bb9531bd1c53572a23fa46ac76ef9b6ad0da6b29ab07cd9e35f *Rescue_0.esp.raw.zst
1fa79e97a76ebeb81109b3967e8cb24c591b79894b3c452246580b76af68a85d *Rescue_0.root-x86-64.raw.zst
8cd89d801499363825daaf912a26c19c4379be917ba41ef41a2ae5c54a4caf75 *Rescue_0.root-x86-64-verity.raw.zst
```

# Configuration

```shell
$ cat mkosi.conf
[Distribution]
Distribution = ubuntu
Release      = lunar
Repositories = main,universe

[Output]
ImageId        = Rescue
ImageVersion   = 0
Format         = disk
SplitArtifacts = yes
SecureBoot     = yes

CompressOutput = zst

# Make root volatile
KernelCommandLine = systemd.volatile=overlay

[Validation]
Checksum  = yes

[Content]
Autologin = yes
Packages  = ubuntu-minimal,dracut,linux-image-generic,systemd-boot-efi

$ cat mkosi.repart/00-ESP.conf 
[Partition]
Type         = esp
Label        = ESP
Format       = vfat

SizeMinBytes = 128M
SizeMaxBytes = 128M

CopyFiles    = /boot/EFI:/

$ cat mkosi.repart/10-os.conf 
[Partition]
Type           = root
Label          = rescue
Format         = squashfs

Minimize       = best
CopyFiles      = /
ExcludeFiles   = /boot/EFI

VerityMatchKey = os
Verity         = data

$ cat mkosi.repart/11-os-verity.conf 
[Partition]
Type           = root-verity
Label          = rescue

SizeMinBytes   = 50M
SizeMaxBytes   = 50M

VerityMatchKey = os
Verity         = hash
```

DaanDeMeyer:
@mcassaniti This PR now generates the following SHA256SUMS:

```
➜  mkosi git:(subvolume) cat mkosi.output/Rescue_0.SHA256SUMS
b3ae271e206db1abf0f578a775db5c221fdc3aa37dc1fd660adae8bae71ec428 *Rescue_0.vmlinuz
39176cceb07925328a18e8d8ca5c50a6d072f15e865ede7c68ef742cb3993860 *Rescue_0.efi
0d9f259fab8bda97ffee6087c7febff51d2f4a0ae03ec7ad23d638bcb0163e63 *Rescue_0.esp.raw.zst
87ccff0d7076ccd2706fa12d5f021c2c528e7a887f3b0c7f1a24103c1103d277 *Rescue_0.root-x86-64.raw.zst
84970747b55bf73236076a52efc0986c2a200a69bd74574d2522b9e9b99a3135 *Rescue_0.root-x86-64-verity.raw.zst
dcd2d3dc566b7c9657ad09e1ba28cb486499aa963a89763a5dda1be3ba22a101 *Rescue_0.raw.zst
```

We don't compress vmlinuz nor the UKI because they're composed out of already compressed parts so it doesn't make sense to compress them again.

Also, you'll need to do a PR for sysupdate so it can pick up the SHA256SUMS at <image-name>.SHA256SUMS. Should be trivial to add to systemd.

mkosi/__init__.py
```diff
@@ -910,7 +918,7 @@ def save_cache(state: MkosiState) -> None:
             shutil.move(state.build_overlay, build)
 
 
-def dir_size(path: PathString) -> int:
+def dir_size(path: Path) -> int:
```
behrmann, line 921: Let's accept `os.DirEntry` here as well, saves the unnecessary cast further down and since this is a recursive function, the call is paid multiple times.

behrmann (commented):
A few nits. I think the change is fine, but an ok from @mcassaniti would be nice, since they seem to be the one battling this issue the most, currently. Before the next release we should really write some test for this.

mkosi/config.py
```diff
@@ -1801,7 +1824,6 @@ def strip_suffixes(path: Path) -> Path:
     while path.suffix in {
```
mcassaniti, line 1824: Taking a look at the outputs in this PR suggests that `.zst` is a valid suffix.

DaanDeMeyer:
Yes we should have more integration tests but not by dropping them all directly in the yaml workflow file, that's not going to scale at all.

mcassaniti (changes requested):

DaanDeMeyer:
Given that SHA256SUMS matches what @mcassaniti requests aside from the things I already mentioned, I think we can go ahead and merge this once CI is done (since I have some more stuff to put up that depends on this)

mcassaniti:
> I would make a PR for systemd-sysupdate to have it look for `f"{url}.SHA256SUMS"` aside from `f"{url}/SHA256SUMS"` so that it also works with the way we output those files.

I don't understand. Can you give me an example URL or two?

> This PR now generates the following SHA256SUMS:
> 
> ```
> ➜  mkosi git:(subvolume) cat mkosi.output/Rescue_0.SHA256SUMS
> b3ae271e206db1abf0f578a775db5c221fdc3aa37dc1fd660adae8bae71ec428 *Rescue_0.vmlinuz
> 39176cceb07925328a18e8d8ca5c50a6d072f15e865ede7c68ef742cb3993860 *Rescue_0.efi
> 0d9f259fab8bda97ffee6087c7febff51d2f4a0ae03ec7ad23d638bcb0163e63 *Rescue_0.esp.raw.zst
> 87ccff0d7076ccd2706fa12d5f021c2c528e7a887f3b0c7f1a24103c1103d277 *Rescue_0.root-x86-64.raw.zst
> 84970747b55bf73236076a52efc0986c2a200a69bd74574d2522b9e9b99a3135 *Rescue_0.root-x86-64-verity.raw.zst
> dcd2d3dc566b7c9657ad09e1ba28cb486499aa963a89763a5dda1be3ba22a101 *Rescue_0.raw.zst
> ```
> 
Brilliant. Exactly what we're all after.
> We don't compress vmlinuz nor the UKI because they're composed out of already compressed parts so it doesn't make sense to compress them again.
> 
Fair enough.

DaanDeMeyer:
> I don't understand. Can you give me an example URL or two?

If you give sysupdate a url such as `https://myimages/my-custom-image`, it should look for ` `https://myimages/my-custom-image/SHA256SUMS`, which it already does, but also for ` `https://myimages/my-custom-image.SHA256SUMS`, which is what I'm suggesting here.

mcassaniti:
I'll ask, but it will be strictly for `${IMAGE_ID}.SHA256SUMS`. It's the only thing that could be substituted in there that is going to be consistent.

mcassaniti:
I've raised https://github.com/systemd/systemd/issues/27537.

PR #1542, opened 2023-05-05
Kernel modules initrd improvements

DaanDeMeyer:
- Let's compress the kernel modules initrd on Debian to at least reduce the disk usage required for the initrd. Not required for other distros since those compress their kernel modules already individually.
- Add option --kernel-modules-initrd to enable/disable usage of the kernel modules initrd. Can be used to disable it if the main initrd already contains the necessary kernel modules
- Add options --kernel-modules-initrd-include/exclude to allow including/excluding the initrds to put in the kernel modules initrd by regex patterns.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -11,6 +11,7 @@ import itertools
 import json
 import logging
 import os
+import re
 import resource
 import shutil
 import subprocess
@@ -55,6 +56,7 @@ from mkosi.util import (
     Verb,
     flatten,
     format_rlimit,
+    is_apt_distribution,
     patch_file,
     prepend_to_environ_path,
     tmp_dir,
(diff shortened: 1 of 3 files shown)


DaanDeMeyer:
Running modinfo once per module was kinda slow but luckily we can pass any number of modules to it and it'll operate on them all so now we run modinfo once in total instead of once per module.

bluca (approved):

mkosi/__init__.py
```diff
@@ -645,19 +647,152 @@ def gen_kernel_images(state: MkosiState) -> Iterator[tuple[str, Path]]:
...
+    while len(todo) > 0:
+        m = todo.pop()
+        if not m or m in mods:
+            continue
+
+        mods.add(m)
+        todo += moddep.get(m, [])
+        firmware.update(firmwaredep.get(m, []))
+
+    return [nametofile[m] for m in mods], list(firmware)
```
behrmann, line 721: Not sure if this is relevant for reproducibility, but unlike dicts, sets are not sorted in insertion order so the `list(firmware)` will have arbitrary order. If that's relevant this should be `sorted(firmware)` instead or a dict used instead of a set by just using the dict keys as a sorted keys with some dummy `None` values. Dito for mods.

PR #1543, opened 2023-05-05
Adopt systemd-firstboot

DaanDeMeyer:
We keep it simple and just adopt the same settings that firstboot
has and pass them through directly.

This commit also reworks how we reset the machine ID, random seed
and a few other files.

This is on top of #1542

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -591,28 +591,60 @@ def gen_kernel_images(state: MkosiState) -> Iterator[tuple[str, Path]]:
         yield kver.name, kimg
 
 
+def filter_kernel_modules(root: Path, kver: str, include: Sequence[str], exclude: Sequence[str]) -> list[Path]:
+    modulesd = Path("usr/lib/modules") / kver
+    modules = set(m.relative_to(root) for m in (root / modulesd).glob("**/*.ko*"))
+
+    for pattern in exclude:
+        regex = re.compile(pattern)
+        remove = set()
+        for m in modules:
+            rel = os.fspath(m.relative_to(modulesd / "kernel"))
+            if regex.search(rel):
+                logging.debug(f"Excluding module {rel}")
+                remove.add(m)
+
+        modules -= remove
+
+    if include:
+        keep: set[Path] = set()
+        for pattern in include:
+            regex = re.compile(pattern)
+            for m in modules:
+                rel = os.fspath(m.relative_to(modulesd / "kernel"))
+                if regex.search(rel):
+                    logging.debug(f"Including module {rel}")
+                    keep.add(m)
+
+        modules = keep
+
+    return sorted(modules)
+
+
 def module_path_to_name(path: Path) -> str:
     return path.name.partition(".")[0]
 
 
-def resolve_module_dependencies(root: Path, kver: str, modules: Sequence[str]) -> tuple[list[Path], list[Path]]:
+def resolve_module_dependencies(root: Path, kver: str, modules: Sequence[str]) -> tuple[set[Path], set[Path]]:
     """
     Returns a tuple of lists containing the paths to the module and firmware dependencies of the given list
-    of module names (including the given module paths themselves).
+    of module names (including the given module paths themselves). The paths are returned relative to the
+    given root directory.
     """
-    allmodules = root.joinpath("usr/lib/modules").joinpath(kver).joinpath("kernel").glob("**/*.ko*")
-    nametofile = {module_path_to_name(m): m for m in allmodules}
+    modulesd = Path("usr/lib/modules") / kver
+    builtin = set(module_path_to_name(Path(m)) for m in (root / modulesd / "modules.builtin").read_text().splitlines())
+    allmodules = set((root / modulesd / "kernel").glob("**/*.ko*"))
+    nametofile = {module_path_to_name(m): m.relative_to(root) for m in allmodules}
 
     # We could run modinfo once for each module but that's slow. Luckily we can pass multiple modules to
     # modinfo and it'll process them all in a single go. We get the modinfo for all modules to build two maps
     # that map the path of the module to its module dependencies and its firmware dependencies respectively.
-    info = run(["modinfo", "--basedir", root, "--set-version", kver, "--null", *nametofile.keys()],
+    info = run(["modinfo", "--basedir", root, "--set-version", kver, "--null", *nametofile.keys(), *builtin],
                text=True, stdout=subprocess.PIPE).stdout
 
     moddep = {}
     firmwaredep = {}
 
-    name = ""
     depends = []
     firmware = []
     for line in info.split("\0"):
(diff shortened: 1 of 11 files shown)


bluca (approved):

bluca:
needs rebase

mkosi/__init__.py
```diff
@@ -1245,7 +1293,7 @@ def configure_ssh(state: MkosiState) -> None:
     if not state.config.ssh:
         return
 
-    state.root.joinpath("etc/systemd/system/ssh.socket").write_text(
+    state.root.joinpath("usr/lib/systemd/system/ssh.socket").write_text(
```
behrmann, line 1296: I think this change is in the right direction. I've been saying in the past, that we should put these into `/usr/lib/` instead of `/etc/`. That being said, we are directly clobbering stuff from the package manager here. There's a point in between: `/usr/local/lib`. The docs mention it for stuff the administrator installs (compared to what they create in `/etc/`), but in principle we could encroach there as the image builder and not step on the toes of package managers.

bluca: The image comes from the same vendor as the packages, I don't think usr/local is appropriate

behrmann: I see your your point, but I think it depends a bit on the definition of vendor, whether one sees the distro or the image builder as vendor.  It's probably not much of an issue here, since these paths only come up in the context of `configure_ssh` and these images probably won't be shipped anywhere. We don't go out of our way for images to be upgradable, but generally I think if we can easily make that use case more seamless, by leaving all the package manager installed files alone (e.g. in Debian `/usr/lib/systemd/system/ssh.socket` is shipped by `openssh-server`), that's a good thing. 

I just generally prefer not to touch the distro stuff, if I can drop stuff in a higher level, but `/usr/local/`'s semantics are a bit unclear.

DaanDeMeyer: Yeah /usr/local doesn't seem entirely appropriate for this stuff either.

I don't think we should make a big deal out of discussing this now. As soon as I can define dropins/units via credentials, I'm removing the --ssh option and this stuff completely, same for the autologin stuff.

DaanDeMeyer:
Added a bunch of other stuff on top, PTAL

mkosi/__init__.py
```diff
@@ -2174,9 +2196,9 @@ def generate_secure_boot_key(args: MkosiArgs) -> None:
         "-new",
         "-x509",
         "-newkey", f"rsa:{keylength}",
-        "-keyout", "mkosi.secure-boot.key",
-        "-out", "mkosi.secure-boot.crt",
-        "-days", str(args.secure_boot_valid_days),
+        "-keyout", "mkosi.key",
+        "-out", "mkosi.crt",
+        "-days", str(args.valid_days),
         "-subj", f"/CN={cn}/",
```
behrmann, line 2202: Not sure whether we *should* recycle keys for other than secureboot stuff, but I guess it's fine until somebody complains. I would prefer a less generic name than `args.valid_days`, though, e.g. `args.keypar_valid_days`. Same for the common name.

mkosi/config.py
```diff
@@ -1007,11 +1007,30 @@ class MkosiConfigParser:
             dest="kernel_modules_initrd_include",
             section="Content",
             parse=config_make_list_parser(delimiter=","),
+            default=[
```
bluca, line 1010: I'm not sure we should exclude stuff here? Isn't this better done in the configuration in the systemd repo?

mkosi/install.py
```diff
@@ -61,5 +61,6 @@ def copy_path(
         f"--preserve=mode,timestamps,links,xattr{',ownership' if preserve_owner else ''}",
         "--no-target-directory",
         "--reflink=auto",
+        "--sparse=always",
```
behrmann, line 64: Out of curiosity, why is this preferable to the default `auto` behaviour?

mkosi/__init__.py
```diff
@@ -2287,6 +2309,10 @@ def target() -> None:
     if build and args.auto_bump:
         bump_image_version()
 
+    # Give disk images some space to play around with if we're booting one.
+    if args.verb in (Verb.shell, Verb.boot, Verb.qemu) and last.output_format == OutputFormat.disk:
+        run(["truncate", "-s", "8G", last.output_dir / last.output])
```
behrmann, line 2314:
```suggestion
        run(["truncate", "--size", "8G", last.output_dir / last.output])
```
This does make the minimum size of the image file 8G when we build it in the context of either of these verbs, right? That should probably be mentioned in the news lest we forget.

mkosi/__init__.py
```diff
@@ -647,28 +591,57 @@ def gen_kernel_images(state: MkosiState) -> Iterator[tuple[str, Path]]:
...
 
+def filter_kernel_modules(root: Path, kver: str, include: Sequence[str], exclude: Sequence[str]) -> list[Path]:
+    modulesd = Path("usr/lib/modules") / kver
+    modules = set(m.relative_to(root) for m in (root / modulesd).glob("**/*.ko*"))
+
+    keep = set()
+    for pattern in include:
+        regex = re.compile(pattern)
+        for m in modules:
+            rel = str(m.relative_to(modulesd / "kernel"))
```
behrmann, line 602:
```suggestion
            rel = os.fspath(m.relative_to(modulesd / "kernel"))
```

mkosi/__init__.py
```diff
@@ -647,28 +591,57 @@ def gen_kernel_images(state: MkosiState) -> Iterator[tuple[str, Path]]:
...
+            if regex.search(rel):
+                logging.debug(f"Including module {rel}")
+                keep.add(m)
+
+    for pattern in exclude:
+        regex = re.compile(pattern)
+        remove = set()
+        for m in modules:
+            rel = str(m.relative_to(modulesd / "kernel"))
```
behrmann, line 611:
```suggestion
            rel = os.fspath(m.relative_to(modulesd / "kernel"))
```

mkosi/__init__.py
```diff
@@ -677,113 +650,82 @@ def resolve_module_dependencies(root: Path, kver: str, modules: Sequence[str]) -
...
-                    continue
-
-                if d not in nametofile:
-                    logging.warning(f"{d} is a dependency of {name} but is not installed, ignoring ")
-                    continue
-
-                depends.append(d)
+            depends += [d for d in value.strip().split(",") if d]
 
         if key == "firmware":
```
behrmann, line 655:
```suggestion
        elif key == "firmware":
```

mkosi/__init__.py
```diff
@@ -677,113 +650,82 @@ def resolve_module_dependencies(root: Path, kver: str, modules: Sequence[str]) -
...
-
-        if key == "filename":
-            if name:
-                moddep[name] = depends
-                firmwaredep[name] = firmware
+            firmware += [f.relative_to(root) for f in root.joinpath("usr/lib/firmware").glob(f"{value.strip()}*")]
 
-                depends = []
-                firmware = []
+        if key == "name":
```
behrmann, line 658:
```suggestion
        elif key == "name":
```

mkosi/__init__.py
```diff
@@ -677,113 +650,82 @@ def resolve_module_dependencies(root: Path, kver: str, modules: Sequence[str]) -
...
-    firmwaredep[name] = firmware
+            depends = []
+            firmware = []
 
-    todo = [*modules]
+    todo = [*builtin, *modules]
     mods = set()
     firmware = set()
 
     while len(todo) > 0:
```
behrmann, line 671:
```suggestion
    while todo:
```
Since the empty list is `False`, this is a bit more pythonic (also more C, I guess :) )

mkosi/config.py
DaanDeMeyer: Given that it's trivial to override this by just assigning to the setting, I think its better to exclude some stuff here. Having initrds of slightly more reasonable size seems like a better default than generating one that includes absolutely everything.

bluca: For generic, default use case this is a surprising behaviour - we cannot possibly know what somebody might need in the initrd - eg: accessibility tools that need audio for example. For our own use case, like the config used for CI jobs here, and for images we build in the systemd repo, we should totally exclude these.

DaanDeMeyer: Dropped default exclude list for now and bumped ESP size to 512MB instead of 256MB to accomodate

mkosi/install.py
DaanDeMeyer: This was a leftover from something else, dropped the commit

PR #1555, opened 2023-05-11
Propagate systemd exit status from the VM

DaanDeMeyer:
Let's make use of the new vmm.notify_socket credential to fetch
systemd's exit status from the VM and propagate it as our own exit
status, just like already happens automatically for containers with
systemd-nspawn.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: LGPL-2.1+
 
+import asyncio
 import base64
 import contextlib
 import datetime
@@ -13,6 +14,7 @@ import os
 import re
 import resource
 import shutil
+import socket
 import subprocess
 import sys
 import tempfile
@@ -36,7 +38,14 @@ from mkosi.manifest import Manifest
 from mkosi.mounts import dissect_and_mount, mount_overlay, scandir_recursive
 from mkosi.pager import page
 from mkosi.remove import unlink_try_hard
-from mkosi.run import become_root, fork_and_wait, run, run_workspace_command, spawn
+from mkosi.run import (
+    MkosiAsyncioThread,
+    become_root,
+    fork_and_wait,
+    run,
+    run_workspace_command,
+    spawn,
+)
 from mkosi.state import MkosiState
 from mkosi.types import PathString
 from mkosi.util import (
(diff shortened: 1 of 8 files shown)


mkosi/__init__.py
```diff
@@ -2024,6 +2033,40 @@ def start_swtpm() -> Iterator[Optional[Path]]:
...
 
 
+@contextlib.contextmanager
+def vsock_notify_handler() -> Iterator[tuple[str, dict[str, str]]]:
+    """
+    This yields a vsock address and a dict that will be filled in with the notifications from the VM. The
+    dict should only be accessed after the context manager has been finalized.
+    """
+    with socket.socket(socket.AF_VSOCK, socket.SOCK_SEQPACKET) as vsock:
+        vsock.bind((socket.VMADDR_CID_ANY, -1))
```
bluca, line 2043: I think you'll want to ensure a low port here, as the client uses that to check if it should trust the host

mkosi/__init__.py
```diff
@@ -2127,11 +2172,17 @@ def run_qemu(args: MkosiArgs, config: MkosiConfig) -> None:
...
+        addr, notifications = stack.enter_context(vsock_notify_handler())
+        cmdline += ["-smbios", f"type=11,value=io.systemd.credential:vmm.notify_socket={addr}"]
+
         cmdline += config.qemu_args
         cmdline += args.cmdline
 
         run(cmdline, stdin=sys.stdin, stdout=sys.stdout, env=os.environ, log=False)
 
+    if "EXIT_STATUS" in notifications:
+        sys.exit(int(notifications["EXIT_STATUS"]))
```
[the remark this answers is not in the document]
DaanDeMeyer: Nothing, you just get the qemu exit status

mkosi/__init__.py
bluca: nvm that's the other way around

bluca (approved):

mkosi/run.py
```diff
@@ -402,3 +406,44 @@ def run_workspace_command(
             if tmp.is_symlink():
                 resolve.unlink(missing_ok=True)
                 shutil.move(tmp, resolve)
+
+
+class AsyncioThread(threading.Thread):
```
behrmann, line 411: Could we name this `MkosiAsyncioThread` or `MkosiAsyncThread` so that there is no confusion if something like this ever pops up in the stdlib? The name would fit pretty well in the `asyncio` module.

mkosi/__init__.py
behrmann, line 2184: Right now `run_qemu` (and `run_ssh` and `run_serve`) fall through to the end of `run_verb`. We removed the parts that `exec`ed so that exits where handled in a single place and while this already pretty much the end, returning the exit code and propagating returning it via `run_verb` and then doing the `sys.exit` in `__main__` would be nicer, since the function can then also be called in isolation without having to catch the `SystemExit`.

DaanDeMeyer: I'm not a fan of returning the exit code. I agree that calling sys.exit() is not very nice though, so I changed it to raise subprocess.CalledProcessError() instead, so it follows exactly the same codepath as when another command fails.

behrmann (approved):

bluca:
`qemu-system-x86_64: -device vhost-vsock-pci,guest-cid=3436578145: vhost-vsock: failed to open vhost device: Permission denied`

PR #1556, opened 2023-05-11
Use /efi as the ESP path when building the image

DaanDeMeyer:
We used to use /boot because kernel-install would hardcode it but since we don't use kernel-install anymore, let's use /efi instead. This way, if anyone wants to install grub, they can do so in /boot and package /boot up as an XBOOTLDR partition.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -399,14 +399,14 @@ def install_boot_loader(state: MkosiState) -> None:
                      f])
 
     with complete_step("Installing boot loader…"):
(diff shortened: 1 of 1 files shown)


behrmann (approved):
I don't quite follow the explanation, because until the C-rewrite `kernel-install` would prefer `/efi` to `/boot` (it now prefers `/boot` if it's `XBOOTLDR`), but this is an all around good change, since e.g. for Debian this would also nicely sidestep `/boot` being filled with all sorts of gunk by dpkg.

DaanDeMeyer:
I meant that kernel-install plugins would hardcode stuff in /boot, my bad.

PR #1560, opened 2023-05-12
Check for btrfs before invoking run()

DaanDeMeyer:
check= does not handle missing binaries so let's check explicitly before running btrfs.


behrmann (approved):
LGTM in general. Not quite sure about erroring out if `bootctl` is not found in case anybody is trying to use mkosi with grub, which would force them to have `bootctl` around if they don't need it, but then again, I don't know whether those people exist.

PR #1570, opened 2023-05-20
Gentoo

257:


mkosi/config.py
```diff
@@ -2228,7 +2229,11 @@ def load_config(args: argparse.Namespace) -> MkosiConfig:
     if args.qemu_vsock == ConfigFeature.enabled and not qemu_check_vsock_support(log=False):
         die("Sorry, the host machine does not support vsock")
 
-    if args.repositories and not (is_dnf_distribution(args.distribution) or is_apt_distribution(args.distribution)):
+    if args.repositories and not (
+        is_dnf_distribution(args.distribution) or
+        is_apt_distribution(args.distribution) or
+        is_portage_distribution(args.distribution)
+    ):
         die("Sorry, the --repositories option is only supported on DNF/Debian based distributions")
```
behrmann, line 2237: The message should be changed accordingly.

mkosi/distributions/gentoo.py
```diff
@@ -27,41 +30,86 @@
...
+    bwrap_params: list[PathString] = [],
     pkgs: Sequence[str] = (),
     actions: Sequence[str] = (),
     opts: Sequence[str] = (),
+    env: dict[str, str] = {},
 ) -> None:
     jobs = os.cpu_count() or 1
+    if sysroot is not None:
+        target_root_mntp = "/tmp/mkosi-root"
+        bwrap_params += ["--bind", state.root, target_root_mntp]
```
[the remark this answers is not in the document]
behrmann: Please don't change this in place, but exchange it for a new list.

mkosi/distributions/gentoo.py
```diff
@@ -27,41 +29,86 @@
...
         "--nospinner",
+        *([f"--root={root}"] if root else []),
     ]
     if ARG_DEBUG.get():
         emerge_default_opts += ["--verbose", "--quiet=n", "--quiet-fail=n"]
     else:
         emerge_default_opts += ["--quiet-build", "--quiet"]
     cmd = ["emerge", *pkgs, *emerge_default_opts, *opts, *actions]
-    run_workspace_command(state.root, cmd, network=True, env=state.environment)
+    bwrap_params += [
```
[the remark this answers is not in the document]
behrmann: Same here.

mkosi/distributions/gentoo.py
```diff
@@ -27,41 +30,86 @@
...
+    root: Optional[Path] = None,
+    bwrap_params: list[PathString] = [],
     pkgs: Sequence[str] = (),
     actions: Sequence[str] = (),
     opts: Sequence[str] = (),
+    env: dict[str, str] = {},
 ) -> None:
     jobs = os.cpu_count() or 1
+    if sysroot is not None:
+        target_root_mntp = "/tmp/mkosi-root"
```
behrmann, line 43: This should maybe be relative to the workspace directory.

behrmann (changes requested):

mkosi/distributions/gentoo.py
257: not sure, could you elaborate please?

257:
@DaanDeMeyer could [this test fail](https://github.com/systemd/mkosi/actions/runs/5069757948/jobs/9104611556#step:9:2652) be related to [running repart without root privileges](https://github.com/systemd/systemd/pull/24944)

DaanDeMeyer:
> @DaanDeMeyer could [this test fail](https://github.com/systemd/mkosi/actions/runs/5069757948/jobs/9104611556#step:9:2652) be related to https://github.com/systemd/systemd/pull/24944

That's just that the ESP partition is too small, the gentoo initrd is too large ;)

257:
> > @DaanDeMeyer could [this test fail](https://github.com/systemd/mkosi/actions/runs/5069757948/jobs/9104611556#step:9:2652) be related to [systemd/systemd#24944](https://github.com/systemd/systemd/pull/24944)
> 
> That's just that the ESP partition is too small, the gentoo initrd is too large ;)

uh, what's the limit?

257:
build successed as root though!

257:
> > > @DaanDeMeyer could [this test fail](https://github.com/systemd/mkosi/actions/runs/5069757948/jobs/9104611556#step:9:2652) be related to [systemd/systemd#24944](https://github.com/systemd/systemd/pull/24944)
> > 
> > 
> > That's just that the ESP partition is too small, the gentoo initrd is too large ;)
> 
> uh, what's the limit?

[are we hard-coding that to 512M?](https://github.com/systemd/mkosi/blob/main/mkosi/__init__.py#L1593) if yes then i guess my question would be why? amd64 can boot much bigger images (i believe up to 4G).

257:
should be fixed now

mkosi/distributions/gentoo.py
257: @behrmann could we mark this as resolved?

behrmann: Sorry, missed your question. My point was, why is this path hardcoded and not relative to `state.workspace`?

257: it's just mount-point inside "/tmp" so it disappears after each `bwrap` call. is it worth making it relative to `state.workspace`?
also, any example where we do something like this somewhere else?

behrmann: I don't think we should scatter random mount points outside of the directories we directly control. At least don't hardcode it to `/tmp/`, but use `mkosi.util.tmp_dir()` and some function from the `tempfile` module so as not to make the path predictable.

257: > I don't think we should scatter random mount points outside of the directories we directly control. At least don't hardcode it to `/tmp/`, but use `mkosi.util.tmp_dir()` and some function from the `tempfile` module so as not to make the path predictable.

how is that *outside of the directories we directly control*?

257: that's a mount-point *inside* `/tmp`, bind-mounted as `tmpfs` again inside root tree of what we get from upstream (stage3) which we save under `cache_dir`. even stage3 is not tainted by this, so i'm not sure what you mean by mount point outside of the ddirectories we control.

marking as resolved, unresolve if need be.

257: that is NOT host's `/tmp/`

behrmann: Applogies, I confused that, but maybe that's an indication to put a comment there to make the context for calling it this way clear.

mkosi/distributions/gentoo.py
```diff
@@ -16,45 +17,103 @@
...
+            "sys-kernel/gentoo-kernel-bin",
+        ],
+        system=["@system"]
+    )
+    portage_consts: GentooConstants = GentooConstants(
+        gentoo_mirrors="http://distfiles.gentoo.org",
+        portdir=Path("/var/db/repos/gentoo"),
+        binrepos_conf_file=Path("etc/portage/binrepos.conf"),
+        ebuild_sh_env_dir=Path("etc/portage/env"),
+        user_config_path=Path("etc/portage")
```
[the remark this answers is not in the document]
behrmann: I'm sorry if that was unclear: What is the benefit of defining a dataclass and encapsulating these constants in it, if those are just four values that will likely never change? I don't think it make it improves readability to have `portage_consts.user_config_path`, when it could be `USER_CONFIG_PATH` for some `USER_CONFIG_PATH=Path("etc/portage")` at the top of the file.

DaanDeMeyer (changes requested):
Can we please get rid of the `Gentoo` class? I think that's the first step into getting the gentoo integration into a better place. There generally should be no need to store variables in a class. There probably shouldn't be more than just be the `GentooInstaller` implementation, a `setup_emerge()` function and a `invoke_emerge()` function to implement Gentoo support. 

Really, just look at how the other distros are handled and copy that as much as possible, the gentoo module shouldn't be a special snowflake, please just write it as similar as possible to how the other distributions are implemented.

DaanDeMeyer (changes requested):
Can we please get rid of the `Gentoo` class? I think that's the first step into getting the gentoo integration into a better place. There generally should be no need to store variables in a class. There probably shouldn't be more than just be the `GentooInstaller` implementation, a `setup_emerge()` function and a `invoke_emerge()` function to implement Gentoo support. 

Really, just look at how the other distros are handled and copy that as much as possible, the gentoo module shouldn't be a special snowflake, please just write it as similar as possible to how the other distributions are implemented.

257:
should be fixed.

DaanDeMeyer:
When I say, remove the class, I don't mean that all methods should become inner functions, please only keep functions where they are reasonable (have clear inputs and outputs).

DaanDeMeyer:
> functions there are really for keeping tasks named and organised. i could just flatten everything in one function.

Please flatten except for stuff like setup_emerge() and invoke_emerge() (look at the other distros). You can sprinkle a few comments around GentooInstaller.install() to delimit the various steps if needed.

mkosi/distributions/gentoo.py
```diff
@@ -82,117 +118,40 @@ class Gentoo:
...
-        self.fetch_fix_stage3()
-        self.set_useflags()
-        self.mkosi_conf()
-        self.get_snapshot_of_portage_tree()
-        self.update_stage3()
-        self.depclean()
+    @classmethod
+    def install(cls, state: MkosiState) -> None:
+        config = state.config
+        root = state.root
```
DaanDeMeyer, line 143: Please drop these and just use `state.config` and `state.root` directly.

mkosi/distributions/gentoo.py
```diff
@@ -16,56 +17,91 @@
...
+        ],
+        boot=[
+            "sys-kernel/gentoo-kernel-bin",
+        ],
+        system=["@system"]
+    )
+    gentoo_mirrors = "http://distfiles.gentoo.org"
+    BINREPOS_CONF_FILE = Path("etc/portage/binrepos.conf")
+    EBUILD_SH_ENV_DIR = Path("etc/portage/env")
+    USER_CONFIG_PATH = Path("etc/portage")
```
DaanDeMeyer, line 101: Let's drop all these and just inline them where they're used

mkosi/distributions/gentoo.py
```diff
@@ -16,56 +17,91 @@
...
 
 
-class Gentoo:
-    arch_profile: Path
+@dataclass
+class GentooAtoms:
+    stage2: list[str]
+    bare_minimal: list[str]
+    boot: list[str]
+    system: list[str]
```
DaanDeMeyer, line 74: Let's drop this class and just inline its uses

mkosi/distributions/gentoo.py
```diff
@@ -16,56 +17,91 @@
 from mkosi.remove import unlink_try_hard
 from mkosi.run import run, run_workspace_command
 from mkosi.state import MkosiState
+from mkosi.types import PathString
+
+thread_counts = (os.cpu_count() or 1) * 2
```
DaanDeMeyer, line 22: Please move this variable to invoke_emerge(). Also, why multiply by 2?

mkosi/distributions/gentoo.py
```diff
@@ -16,56 +17,91 @@
...
 
 
 def invoke_emerge(
     state: MkosiState,
+    sysroot: Optional[Path] = None,
+    root: Optional[Path] = None,
```
DaanDeMeyer, line 28: This argument does not seem to be used?

mkosi/distributions/gentoo.py
```diff
@@ -16,56 +17,91 @@
...
     pkgs: Sequence[str] = (),
     actions: Sequence[str] = (),
     opts: Sequence[str] = (),
+    env: dict[str, str] = {},
 ) -> None:
-    jobs = os.cpu_count() or 1
+    bwrap: list[PathString] = []
+    if sysroot is not None:
+        # this is inside our ns for stage, i.e. this is NOT host's /tmp
+        target_root_mntp = "/tmp/mkosi-root"
```
DaanDeMeyer, line 38: This could use a lot more information on why this is necessary. I'm confused why we need both `root` and `sysroot`?

mkosi/distributions/gentoo.py
```diff
@@ -211,43 +170,50 @@ def fetch_fix_stage3(self) -> None:
...
             stage3_tar_path.parent.mkdir(parents=True, exist_ok=True)
             urllib.request.urlretrieve(stage3_url_path, stage3_tar_path)
 
-        stage3_tmp_extract.mkdir(parents=True, exist_ok=True)
+        cls.stage3_cache.mkdir(parents=True, exist_ok=True)
 
-        with flock(stage3_tmp_extract):
-            if not stage3_tmp_extract.joinpath(".cache_isclean").exists():
-                log_step(f"Extracting {stage3_tar.name} to {stage3_tmp_extract}")
+        with flock(cls.stage3_cache):
```
DaanDeMeyer, line 197: This shouldn't be necessary. If this is required, we should lock the cache directory itself for all distros, not just gentoo

mkosi/distributions/gentoo.py
```diff
@@ -295,41 +270,62 @@ def mkosi_conf(self) -> None:
...
+        opts += ["--noreplace", "--root-deps=rdeps"]
+        with complete_step("Merging bare minimal atoms"):
+            invoke_emerge(state, sysroot=cls.stage3_cache,
+                          opts=opts+["--exclude", "sys-devel/*"],
+                          pkgs=cls.pkgs.bare_minimal, env=emerge_vars)
+        with complete_step("Merging atoms required for boot"):
+            invoke_emerge(state, sysroot=cls.stage3_cache, opts=opts,
+                          pkgs=cls.pkgs.boot)
+        if state.config.make_initrd:
+            return
```
DaanDeMeyer, line 320: Please use newlines a bit more judiciously here

mkosi/distributions/gentoo.py
```diff
@@ -82,117 +118,40 @@ class Gentoo:
...
+        config = state.config
+        root = state.root
+
+        arch = state.installer.architecture(state.config.architecture)
 
-    def fetch_fix_stage3(self) -> None:
         """usrmerge tracker bug: https://bugs.gentoo.org/690294"""
 
+        if config.mirror:
+            cls.gentoo_mirrors = config.mirror
```
DaanDeMeyer, line 150: Let's drop this and just use `config.mirror` directly.

DaanDeMeyer (changes requested):
Looking a lot better already,

mkosi/distributions/gentoo.py
257: `root` is where we're installing, `sysroot` is our toolchain (portage, compiler, linker etc).
See here for more: [emerge(1)](https://dev.gentoo.org/~zmedico/portage/doc/man/emerge.1.html).

mkosi/distributions/gentoo.py
257: what if `config.mirror` is `None`? `MkosiConfig` is `frozen`.

mkosi/distributions/gentoo.py
257: defensive really... removed.

mkosi/distributions/gentoo.py
DaanDeMeyer: You can set a default mirror in config_default_mirror() in config.py

257: done!

mkosi/distributions/gentoo.py
257: 2 is for hypethreading.

mkosi/distributions/gentoo.py
```diff
@@ -295,41 +220,60 @@ def mkosi_conf(self) -> None:
...
+        opts = [
+            "--with-bdeps=n",
+            "--complete-graph-if-new-use=y",
+            "--verbose-conflicts",
+            "--changed-use",
+            "--newuse",
+        ]
+        env = {}
+        env.update(emerge_vars)
+        env.update({"USE": f"{emerge_vars['USE']} build"})
```
DaanDeMeyer, line 256: These can be merged into a single initialization, no need to use `update()`

mkosi/distributions/gentoo.py
```diff
@@ -295,41 +220,60 @@ def mkosi_conf(self) -> None:
...
+        with complete_step("merging stage2"):
+            invoke_emerge(state, sysroot=cls.stage3_cache, opts=opts+["--emptytree", "--nodeps"],
+                          pkgs=["sys-apps/baselayout", "sys-apps/util-linux"], env=env)
+        opts += ["--noreplace", "--root-deps=rdeps"]
+        with complete_step("Merging bare minimal atoms"):
+            invoke_emerge(state, sysroot=cls.stage3_cache, opts=opts+["--exclude", "sys-devel/*"],
+                          pkgs=["app-shells/bash", "sys-apps/systemd"], env=emerge_vars)
+        with complete_step("Merging atoms required for boot"):
+            invoke_emerge(state, sysroot=cls.stage3_cache, opts=opts,
+                          pkgs=["sys-kernel/gentoo-kernel-bin"])
```
DaanDeMeyer, line 266: Please add a newline after each with block

mkosi/distributions/gentoo.py
```diff
@@ -295,41 +220,60 @@ def mkosi_conf(self) -> None:
...
+        if state.config.make_initrd:
+            return
+
+        invoke_emerge(state, sysroot=cls.stage3_cache, opts=opts,
+                      pkgs=["@system"])
```
DaanDeMeyer, line 271: We should not install `@system` by default. For all other distributions, we only install the package providing the base filesystem layout and the package providing /etc/passwd. We should do the same for gentoo.

mkosi/distributions/gentoo.py
```diff
@@ -1,198 +1,99 @@
...
 from mkosi.log import ARG_DEBUG, complete_step, die, log_step
 from mkosi.remove import unlink_try_hard
 from mkosi.run import run, run_workspace_command
 from mkosi.state import MkosiState
+from mkosi.types import PathString
 
 
 def invoke_emerge(
     state: MkosiState,
+    sysroot: Optional[Path] = None,
```
DaanDeMeyer, line 24: I don't see `invoke_emerge()` being called anywhere with `sysroot` set to `None` so this doesn't need to be optional?

mkosi/distributions/gentoo.py
```diff
@@ -1,198 +1,99 @@
...
+    bwrap_params: list[PathString] = [],
```
DaanDeMeyer, line 25: This argument seems to never be used? Please remove it if that's the case

mkosi/distributions/gentoo.py
```diff
@@ -1,198 +1,99 @@
...
-        "parallel-install",
-        "getbinpkg",
-        "-candy",
-    ]
+class GentooInstaller(DistributionInstaller):
+    stage3_cache: Path
+
+    @classmethod
+    def filesystem(cls) -> str:
+        return "btrfs"
```
DaanDeMeyer, line 71: Please add some justification to the commit message for this change? What makes btrfs a better default filesystem than ext4 for Gentoo?

mkosi/distributions/gentoo.py
```diff
@@ -211,43 +112,58 @@ def fetch_fix_stage3(self) -> None:
             else:
                 die("profile names changed upstream?")
 
+        stage3_tar_path = state.cache_dir / stage3_tar
+        cls.stage3_cache = stage3_tar_path.with_name(stage3_tar.name).with_suffix(".tmp")
+
+        user_config_path = cls.stage3_cache / USER_CONFIG_PATH
```
DaanDeMeyer, line 118: This should probably take the user config from `state.pkgmngr / "etc/portage"`

mkosi/distributions/gentoo.py
```diff
@@ -211,43 +112,58 @@ def fetch_fix_stage3(self) -> None:
...
+
+        emerge_vars = {
```
DaanDeMeyer, line 120: Can we move this closer to where its first used?

mkosi/distributions/gentoo.py
```diff
@@ -211,43 +112,58 @@ def fetch_fix_stage3(self) -> None:
...
+                    "-usersandbox",
+                    "-sandbox",
+                    "-pid-sandbox",  # for cross-compile scenarios
+                    "-network-sandbox",
+                    "parallel-install",
+                    "getbinpkg",
+                    "-candy",
+                    "noman",
+                    "nodoc",
+                    "noinfo",
```
DaanDeMeyer, line 135: We shouldn't set these by default right? Only if `config.with_docs` is `True`?

mkosi/distributions/gentoo.py
```diff
@@ -211,43 +112,58 @@ def fetch_fix_stage3(self) -> None:
...
+        emerge_vars = {
+            "FEATURES": " ".join([
+                    # -user* are required for access to USER_CONFIG_PATH
+                    "-userfetch",
+                    "-userpriv",
+                    "-usersync",
+                    "-usersandbox",
+                    "-sandbox",
+                    "-pid-sandbox",  # for cross-compile scenarios
+                    "-network-sandbox",
```
DaanDeMeyer, line 129: Do we need all these if `bwrap` is already providing the sandbox? Seems like doing double work for no benefit

mkosi/distributions/gentoo.py
```diff
@@ -211,43 +112,58 @@ def fetch_fix_stage3(self) -> None:
...
-                stage3_tmp_extract.joinpath(".cache_isclean").touch()
-
-        log_step(f"Copying {stage3_tmp_extract} to {self.root}")
-        copy_path(stage3_tmp_extract, self.root)
-
-    def set_useflags(self) -> None:
-        package_use = self.portage_cfg_dir / "package.use"
+        cls.stage3_cache.mkdir(parents=True, exist_ok=True)
+
+        log_step(f"Extracting {stage3_tar.name} to {cls.stage3_cache}")
```
DaanDeMeyer, line 153: Please use `complete_step()` here instead of `log_step()`

mkosi/distributions/gentoo.py
```diff
@@ -271,21 +187,30 @@ def set_useflags(self) -> None:
...
         )
+        if state.config.make_initrd:
+            package_use.joinpath("minimal").write_text(
+                dedent(
+                    """\
+                    # MKOSI
+                    */* minimal
+                    """
+                )
+            )
```
DaanDeMeyer, line 198: We should do the `minimal` stuff by default, not conditional on the value of `config.make_initrd`

DaanDeMeyer (changes requested):
Another round

mkosi/distributions/gentoo.py
257: `-sandbox' reads dont-do-sandbox, hyphene '-' prefix is for _minus_.

mkosi/distributions/gentoo.py
257: we wanted smaller image no? :)

mkosi/distributions/gentoo.py
```diff
@@ -211,43 +112,58 @@ def fetch_fix_stage3(self) -> None:
...
-            self.portage_cfg["GENTOO_MIRRORS"],
-            f"releases/{self.arch}/autobuilds/{stage3_tar}",
+            state.config.mirror, f"releases/{arch}/autobuilds/{stage3_tar}",
         )
-        stage3_tar_path = self.state.cache_dir / stage3_tar
-        stage3_tmp_extract = stage3_tar_path.with_name(stage3_tar.name + ".tmp")
         if not stage3_tar_path.is_file():
             log_step(f"Fetching {stage3_url_path}")
             stage3_tar_path.parent.mkdir(parents=True, exist_ok=True)
             urllib.request.urlretrieve(stage3_url_path, stage3_tar_path)
```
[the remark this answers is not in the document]
257: there were promisses upstream (2 years ago now) for changing the way they would release stage3. i will have to double check, but lets not make the PR bigger than what it needs to be. it's in my TODO list.

mkosi/distributions/gentoo.py
257: right, that's what the `flock` and `cache_isclean` were for. i can put them back.

mkosi/distributions/gentoo.py
```diff
@@ -211,43 +112,58 @@ def fetch_fix_stage3(self) -> None:
...
+
+        log_step(f"Extracting {stage3_tar.name} to {cls.stage3_cache}")
+        run([
+            "tar",
+            "--numeric-owner",
+            "-C", cls.stage3_cache,
+            "--extract",
+            "--file", stage3_tar_path,
+            "--exclude", "./dev",
+        ])
```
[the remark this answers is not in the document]
257: right, that's what `flock` and `cache_isclean` were for, AFAIR. i can put them back.

mkosi/distributions/gentoo.py
DaanDeMeyer: This was not addressed?

mkosi/distributions/gentoo.py
257: i will update the commit msg with i just said above. let me know.

mkosi/distributions/gentoo.py
```diff
@@ -211,43 +112,58 @@ def fetch_fix_stage3(self) -> None:
...
+        unlink_try_hard(cls.stage3_cache.joinpath("dev"))
+        unlink_try_hard(cls.stage3_cache.joinpath("proc"))
+        unlink_try_hard(cls.stage3_cache.joinpath("sys"))
```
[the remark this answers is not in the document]
257: > we now exclude those in `tar` extract.

seems to break the compiler, reverting for now.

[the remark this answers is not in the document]
257: updated!

mkosi/distributions/gentoo.py
257: removed the `Optional`.

mkosi/distributions/gentoo.py
257: done!

mkosi/distributions/gentoo.py
257: they're conditional now.

mkosi/distributions/gentoo.py
```diff
@@ -1,198 +1,96 @@
...
 def invoke_emerge(
     state: MkosiState,
+    sysroot: Optional[Path] = None,
     pkgs: Sequence[str] = (),
     actions: Sequence[str] = (),
     opts: Sequence[str] = (),
+    env: dict[str, str] = {},
 ) -> None:
-    jobs = os.cpu_count() or 1
+    thread_counts = (os.cpu_count() or 1) * 2  # * 2 for hyperthreading
```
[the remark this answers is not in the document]
257: we don't support building on ARM do we? at least not for Gentoo yet (i would flip to `llvm` happily once all pieces are there upstream, and one day we may even have cross-compilation support).

this also effects `emerge`'s own `load-average` which greatly speeds up things in binary installations, i.e. when a bin repo is used or in successive builds. i would keep it.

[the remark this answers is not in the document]
behrmann: Please read the comment again. This is not about building on ARM, but the assertion that adding a factor of two somehow accounts for hyperthreading is mistaken.

Take for example an `12th Gen Intel(R) Core(TM) i9-12900`. `os.cpu_count()` will return 24, thus setting `thread_counts` to 48 in this case. The actual number of CPU cores is 16, though, and of those only 8 even support hyperthreading (hence the 24). At least for numerical workloads, limiting oneself to the 8 fast cores will actually yield optimal performance. The mileage for compilation workloads will surely be different, but the right number will most certainly be ≤`os.cpu_count()` not twice the number.

That being said, the choice of maximum load average as one larger than `thread_counts` is weird as well.

mkosi/distributions/gentoo.py
```diff
@@ -211,43 +109,39 @@ def fetch_fix_stage3(self) -> None:
             else:
                 die("profile names changed upstream?")
 
+        stage3_tar_path = state.cache_dir / stage3_tar
+        cls.stage3_cache = stage3_tar_path.with_name(stage3_tar.name).with_suffix(".tmp")
```
DaanDeMeyer, line 113: Instead of adding the ".tmp" suffix, we should use the tar name without any extensions. Also, we should extract to a fixed location regardless of the tar name and get rid of `cls.stage3_cache` and hardcode the location instead.

mkosi/distributions/gentoo.py
DaanDeMeyer: Not addressed, also you can just call this `config` or so

mkosi/distributions/gentoo.py
```diff
@@ -211,43 +109,39 @@ def fetch_fix_stage3(self) -> None:
...
+                "-C", cls.stage3_cache,
+                "--extract",
+                "--file", stage3_tar_path,
+                "--exclude", "./dev",
+            ])
 
-                stage3_tmp_extract.joinpath(".cache_isclean").touch()
+            unlink_try_hard(cls.stage3_cache.joinpath("dev"))
+            unlink_try_hard(cls.stage3_cache.joinpath("proc"))
+            unlink_try_hard(cls.stage3_cache.joinpath("sys"))
```
DaanDeMeyer, line 142: We're still doing this? I thought you commented elsewhere this was removed and replaced by excluding in `tar`?

DaanDeMeyer (changes requested):

mkosi/distributions/gentoo.py
```diff
@@ -295,41 +214,64 @@ def mkosi_conf(self) -> None:
...
 
-    def get_snapshot_of_portage_tree(self) -> None:
-        run_workspace_command(self.state.root, ["/usr/bin/emerge-webrsync"], network=True,
-                              env=self.state.environment)
-
-    def update_stage3(self) -> None:
-        invoke_emerge(self.state, opts=self.EMERGE_UPDATE_OPTS, pkgs=self.pkgs['boot'])
-        invoke_emerge(self.state, opts=["--config"], pkgs=["sys-kernel/gentoo-kernel-bin"])
-        invoke_emerge(self.state, opts=self.EMERGE_UPDATE_OPTS, pkgs=self.pkgs['sys'])
+        repos_cfg = cls.stage3_cache / BINREPOS_CONF_FILE
```
DaanDeMeyer, line 217: This is only used once so you can inline the variable

mkosi/distributions/gentoo.py
```diff
@@ -295,41 +214,64 @@ def mkosi_conf(self) -> None:
...
+
+        opts = [
+            "--with-bdeps=n",
+            "--complete-graph-if-new-use=y",
+            "--verbose-conflicts",
+            "--changed-use",
+            "--newuse",
+            "--root-deps=rdeps",
+        ]
+        with complete_step("merging stage2"):
```
DaanDeMeyer, line 249: "merging" should be capitalized

mkosi/distributions/gentoo.py
```diff
@@ -295,41 +214,64 @@ def mkosi_conf(self) -> None:
...
 
-    def depclean(self) -> None:
-        invoke_emerge(self.state, actions=["--depclean"])
+        opts += ["--noreplace"]
 
+        with complete_step("Merging bare minimal atoms"):
+            invoke_emerge(state, sysroot=cls.stage3_cache,
+                          opts=opts+["--exclude", "sys-devel/*"],
+                          pkgs=["app-shells/bash", "sys-apps/systemd"],
+                          env=emerge_vars)
```
DaanDeMeyer, line 261: Can we drop this? Ideally the base system without any configuration has just the base directory layout and /etc/passwd + related files, nothing else

mkosi/distributions/gentoo.py
DaanDeMeyer: Yes please remove it. By default we try to build the most minimal image possible, even if it's not useful because it's much easier to add missing stuff than it is to remove one of mkosi's defaults that a user doesn't need.

Think of use cases like building sysexts or portable services where systemd is not necessary in the image.

mkosi/distributions/gentoo.py
```diff
@@ -295,41 +214,64 @@ def mkosi_conf(self) -> None:
...
+            for repo in state.config.repositories:
+                f.write(
+                    dedent(
+                        f"""\
+                        # MKOSI
+                        [binhost]
+                        sync-uri = {repo}
+                        """
+                    )
+                )
```
DaanDeMeyer, line 228: `--repositories` should only be used to enable repositories that are disabled by default, not to add new ones. For that, we now have `--package-manager-tree` so this should be reworked to pick up `etc/portage/binrepos.conf` from `state.pkgmngr` if it exists there.

mkosi/distributions/gentoo.py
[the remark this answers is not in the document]
257: [it actually does!](https://github.com/gentoo/portage/blob/master/lib/portage/util/cpuinfo.py#L31). lets drop this.

mkosi/distributions/gentoo.py
```diff
@@ -11,188 +10,87 @@
...
-        "--changed-use",
-        "--newuse",
-        "--deep",
-        "--with-bdeps=y",
-        "--complete-graph-if-new-use=y",
-        "--verbose-conflicts",
+    bwrap += [
+        "--bind", state.cache_dir / "binpkgs", "/var/cache/binpkgs",
+        "--bind", state.cache_dir / "distfiles", "/var/cache/distfiles",
+        "--bind", state.cache_dir / "repos", "/var/db/repos",
```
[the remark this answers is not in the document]
257: yes, `emerge` looks up `DISTDIR` in `env` for `distfiles` for instance. but why would we want to set them to non-default locations?

this is to preserve the `binpkgs`, `repos` and source codes (`distfiles`) in consecutive calls, since they're expensive to get/build.

would you explain a bit more please?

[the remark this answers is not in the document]
DaanDeMeyer: Yes my bad I misunderstood that this is actually required

DaanDeMeyer:
A more general question, if I understand correctly, gentoo has experimental support for binary packages. Would it make sense to only support Gentoo with binary packages in mkosi? And if a package is not available in the binary package repositories, instruct users to either have it added to an existing repo or build their own and specify it in binrepos?

That should allow us to get rid of the sysroot usage and use emerge directly from the host I think as we wouldn't depend on a compiler and such anymore since every package would be a binary package. Does this make sense or is Gentoo not ready yet to be used like this?

257:
> A more general question, if I understand correctly, gentoo has experimental support for binary packages. Would it make sense to only support Gentoo with binary packages in mkosi? And if a package is not available in the binary package repositories, instruct users to either have it added to an existing repo or build their own and specify it in binrepos?
> 
> That should allow us to get rid of the sysroot usage and use emerge directly from the host I think as we wouldn't depend on a compiler and such anymore since every package would be a binary package. Does this make sense or is Gentoo not ready yet to be used like this?

1. `portage` from non-Gentoo hosts is not even properly supported upstream.
2. Gentoo is a source-based distro, i don't thing it will ever be "ready" for such a usage.

mkosi/distributions/gentoo.py
257: dropped the `.tmp` suffix.

for `cls.stage3_cache`:
`invoke_emerge()`, called from `install_package()` needs to pass `sysroot`, which points to `cls.stage3_cache`. say we hardcode `cls.stage_cache`, a `Path`, to some directory name so that said `sysroot` ends something like `state.cache_dir.joinpath(stage3_dir)`, where `stage3_dir` is set to `"stage3"`, a `str`. doesn't this buy us anything?

mkosi/distributions/gentoo.py
257: excluding `sys` breaks compiler! i suspect permission issue.

mkosi/distributions/gentoo.py
```diff
@@ -16,183 +14,92 @@
 from mkosi.remove import unlink_try_hard
 from mkosi.run import run, run_workspace_command
 from mkosi.state import MkosiState
+from mkosi.types import PathString
 
 
 def invoke_emerge(
     state: MkosiState,
+    sysroot: Path,
     pkgs: Sequence[str] = (),
```
DaanDeMeyer, line 23: Please write the variable name out in full

mkosi/distributions/gentoo.py
```diff
@@ -16,183 +14,92 @@
...
     actions: Sequence[str] = (),
     opts: Sequence[str] = (),
```
DaanDeMeyer, line 25: Same here, please use "options"

mkosi/distributions/gentoo.py
DaanDeMeyer: Yeah sure we can keep the cache_isclean but let's drop the flock()

257: dropped the `flock`

mkosi/distributions/gentoo.py
```diff
@@ -211,125 +118,142 @@ def fetch_fix_stage3(self) -> None:
...
+        cls.stage3_cache = state.cache_dir.joinpath("stage3")
+        if not stage3_tar_path.is_file():
+            if cls.stage3_cache.exists():
+                log_step("New stage3 is available, removing the old one")
+                unlink_try_hard(cls.stage3_cache)
+            with complete_step(f"Fetching {stage3_url_path}"):
+                stage3_tar_path.parent.mkdir(parents=True, exist_ok=True)
+                urllib.request.urlretrieve(stage3_url_path, stage3_tar_path)
+        cls.stage3_cache.mkdir(parents=True, exist_ok=True)
+        if not cls.stage3_cache.joinpath(".cache_isclean").exists():
```
DaanDeMeyer, line 136: Do we really need this file? We can just check if the extracted directory exists no?

mkosi/distributions/gentoo.py
```diff
@@ -211,125 +118,142 @@ def fetch_fix_stage3(self) -> None:
...
-                sys-apps/systemd homed cryptsetup -pkcs11
-                # See: https://bugs.gentoo.org/832167
-                sys-auth/pambase homed
-
-                # MKOSI: usronly
-                sys-apps/systemd repart
-                # MKOSI: make sure we're init (no openrc)
-                sys-apps/systemd sysv-utils
-                """
+            for d in ("binpkgs", "distfiles", "repos"):
```
DaanDeMeyer, line 148: Everything from this point onwards doesn't need to be cached as it will be more or less instant to execute.

mkosi/distributions/gentoo.py
```diff
@@ -211,125 +118,142 @@ def fetch_fix_stage3(self) -> None:
...
+                    sys-apps/systemd homed cryptsetup -pkcs11
+                    # See: https://bugs.gentoo.org/832167
+                    sys-auth/pambase homed
+
+                    # MKOSI: usronly
+                    sys-apps/systemd repart
+                    # MKOSI: make sure we're init (no openrc)
+                    sys-apps/systemd sysv-utils
+                    """
+                )
```
DaanDeMeyer, line 174: I don't think we should hardcode this into mkosi. systemd might not even be installed into the image, so we shouldn't be configuring it. This should be configured by the distribution.

mkosi/distributions/gentoo.py
```diff
@@ -211,125 +118,142 @@ def fetch_fix_stage3(self) -> None:
...
+            emerge_vars_str = ""
+            emerge_vars_str += "\n".join(f'{k}="${{{k}}} {v}"' for k, v in emerge_vars.items())
+            ebuild_sh_env_dir.joinpath("mkosi.conf").write_text(
+                dedent(
+                    f"""\
+                    # MKOSI: these were used during image creation...
+                    # and some more! see under package.*/
+                    {emerge_vars_str}
+                    """
+                )
```
DaanDeMeyer, line 195: The emerge vars make sense when emerge is invoked by mkosi but I don't think they make sense to use in the image as well. Again, this should probably be left to the distro defaults or configured by the user but not configured by mkosi

DaanDeMeyer (changes requested):

mkosi/distributions/gentoo.py
257: i have related question. `install_package_manager_trees()` is called before `install_distribution()` in `build_image()`.

look at these two lines in `gentoo.py`:
```
copy_path(config, root_portage_cfg)
copy_path(config, cls.stage3_cache / "etc/portage")
```

am i replicating what `install_package_manager_trees()` does there?

mkosi/distributions/gentoo.py
257: my idea was to document what we used for building the image. i can drop the documentation dance but we needed those for `emerge` to function properly in a sandboxed enviroment. this might have changed by now. that bit of code is old.

257: ... let me know if we want the documentaion bit to be dropped, please.

mkosi/distributions/gentoo.py
257: `gnuefi` is a must or else `mkosi` fails.
other stuff are from before `usr-merge` era. let make an issue out of this so i remember to come back to it once we have landed this.

257: https://github.com/systemd/mkosi/issues/1640

mkosi/distributions/gentoo.py
257: we make that directory unconditionally since `tar` changes into, `-C`, before extracting.

mkosi/distributions/gentoo.py
```diff
@@ -211,125 +118,143 @@ def fetch_fix_stage3(self) -> None:
...
-        invoke_emerge(self.state, opts=self.EMERGE_UPDATE_OPTS, pkgs=self.pkgs['sys'])
+        opts += ["--noreplace"]
 
-    def depclean(self) -> None:
-        invoke_emerge(self.state, actions=["--depclean"])
+        with complete_step("Merging bare minimal atoms"):
+            invoke_emerge(state, sysroot=cls.stage3_cache,
+                          opts=opts+["--exclude", "sys-devel/*"],
+                          pkgs=["sys-apps/systemd"],
+                          env=emerge_vars)
```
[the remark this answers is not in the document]
257: without systemd we `die` like this:
```
  File "./src/github/257/mkosi/mkosi/__init__.py", line 1725, in build_image
    install_boot_loader(state)
  File "./src/github/257/mkosi/mkosi/__init__.py", line 424, in install_boot_loader
    die("A bootable image was requested but systemd-boot was not found at "
  File "./src/github/257/mkosi/mkosi/log.py", line 30, in die
    sys.exit(1)
```

mkosi/distributions/gentoo.py
```diff
@@ -16,183 +14,92 @@
...
     actions: Sequence[str] = (),
     opts: Sequence[str] = (),
+    env: dict[str, str] = {},
 ) -> None:
-    jobs = os.cpu_count() or 1
-    emerge_default_opts = [
+    # This is the mount-point inside our sysroot where we mount root
+    target_root_mntp = "/tmp/mkosi-root"
+    bwrap: list[PathString] = ["--bind", state.root, target_root_mntp]
+    root = Path(target_root_mntp)
```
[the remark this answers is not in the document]
257: not sure, please see the updated version.

mkosi/distributions/gentoo.py
```diff
@@ -16,183 +14,92 @@
...
     pkgs: Sequence[str] = (),
     actions: Sequence[str] = (),
     opts: Sequence[str] = (),
+    env: dict[str, str] = {},
 ) -> None:
-    jobs = os.cpu_count() or 1
-    emerge_default_opts = [
+    # This is the mount-point inside our sysroot where we mount root
+    target_root_mntp = "/tmp/mkosi-root"
+    bwrap: list[PathString] = ["--bind", state.root, target_root_mntp]
```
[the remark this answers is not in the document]
257: tests will fail:
`mkosi/distributions/gentoo.py:54:54: error: Argument "bwrap_params" to "run_workspace_command" has incompatible type "List[object]"; expected "Sequence[Union[Path, str]]"  [arg-type]`

mkosi/distributions/gentoo.py
DaanDeMeyer: Yeah let's drop it, we don't do it for other distros either

mkosi/distributions/gentoo.py
DaanDeMeyer: Check if it is non-empty then

mkosi/distributions/gentoo.py
```diff
@@ -10,189 +8,98 @@
...
+        *pkgs,
         "--buildpkg=y",
         "--usepkg=y",
         "--keep-going=y",
-        f"--jobs={jobs}",
-        f"--load-average={jobs+1}",
+        "--jobs",
+        "--load-average",
         "--nospinner",
+        *([f"--root={root}"] if root != sysroot else []),
```
DaanDeMeyer, line 47: When will: `root == sysroot` ever be true?

mkosi/distributions/gentoo.py
```diff
@@ -10,189 +8,98 @@
...
-    jobs = os.cpu_count() or 1
-    emerge_default_opts = [
+    # This is the mount-point inside our sysroot where we mount root
+    bwrap: list[PathString] = [
+        "--bind", state.root, "/tmp/mkosi-root",
+        "--bind", state.cache_dir / "binpkgs", "/var/cache/binpkgs",
+        "--bind", state.cache_dir / "distfiles", "/var/cache/distfiles",
+        "--bind", state.cache_dir / "repos", "/var/db/repos",
+    ]
+    root = Path("/tmp/mkosi-root")
```
DaanDeMeyer, line 36: I'm pretty sure we can drop this variable as well

mkosi/distributions/gentoo.py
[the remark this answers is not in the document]
DaanDeMeyer: Yes you need to add a gentoo config file in `.github/mkosi.conf.d/20-gentoo.conf` to build a bootable gentoo image.

mkosi/distributions/gentoo.py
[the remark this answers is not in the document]
257: https://github.com/systemd/mkosi/issues/1641

mkosi/distributions/gentoo.py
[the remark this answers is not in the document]
DaanDeMeyer: Well you need to add a `Packages=` setting that installs systemd and a kernel and such

mkosi/distributions/gentoo.py
257: you're right, in this form it will never be.

mkosi/distributions/gentoo.py
```diff
@@ -10,189 +8,97 @@
...
+                    "getbinpkg",
+                    "-candy",
+                    *(["noman", "nodoc", "noinfo"] if state.config.with_docs else []),
+            ]),
+            "USE": "initramfs minimal symlink"
+        }
+        arch = state.installer.architecture(state.config.architecture)
 
-    def fetch_fix_stage3(self) -> None:
         """usrmerge tracker bug: https://bugs.gentoo.org/690294"""
```
behrmann, line 95: This should be a comment now, since right now it's an unused string in the middle of a function.

mkosi/distributions/gentoo.py
```diff
@@ -211,45 +117,42 @@ def fetch_fix_stage3(self) -> None:
...
+            state.cache_dir.joinpath(d).mkdir(exist_ok=True)
 
-    def set_useflags(self) -> None:
-        package_use = self.portage_cfg_dir / "package.use"
-        package_use.mkdir(exist_ok=True)
+        config = state.pkgmngr / "etc/portage"
 
+        package_use = config / "package.use"
+        package_use.mkdir(parents=True, exist_ok=True)
         package_use.joinpath("systemd").write_text(
```
behrmann, line 156: This and some other instances further below clobber the files in `state.pkgmngr`. The idea is for users to be able to pass in their own files. Is there a way to make this work for Gentoo? In the other distros we try hard to not clobber things and instead pass arguments on the cmdline to overwrite config files.

257: > This and some other instances further below clobber the files in state.pkgmngr.

that's the effect of gentoo being bare minimum whereas other distros, it seems to me, come with a canonical base config which is probably the side effect of being debian/fedora offshoots basically and following the _upstream_ to the letter.
 
> The idea is for users to be able to pass in their own files.

i thought and used, `skeleton/extra` for that in the past.

> Is there a way to make this work for Gentoo? In the other distros we try hard to not clobber things and instead pass arguments on the cmdline to overwrite config files.

i see. please see my question to @DaanDeMeyer above, regarding `install_package_manager_trees()`. that happens before things are handed over to the `installer()` instance. i revisit these as in  https://github.com/systemd/mkosi/issues/1640

mkosi/distributions/gentoo.py
257: good catch, removed.

mkosi/distributions/gentoo.py
```diff
@@ -10,189 +8,97 @@
...
-        "-usersync",
-        "-usersandbox",
-        "-sandbox",
-        "-pid-sandbox",  # for cross-compile scenarios
-        "-network-sandbox",
-        "parallel-install",
-        "getbinpkg",
-        "-candy",
-    ]
+    # This is the mount-point inside our sysroot where we mount root
```
DaanDeMeyer, line 29: Outdated comment, should be moved as well

mkosi/distributions/gentoo.py
```diff
@@ -10,189 +8,97 @@
...
 
 def invoke_emerge(
     state: MkosiState,
-    pkgs: Sequence[str] = (),
+    sysroot: Path,
+    packages: Sequence[str] = (),
     actions: Sequence[str] = (),
-    opts: Sequence[str] = (),
+    options: Sequence[str] = (),
+    env: dict[str, str] = {},
```
DaanDeMeyer, line 27: Please use Mapping as the type for this

mkosi/distributions/gentoo.py
```diff
@@ -247,8 +248,8 @@ def install(cls, state: MkosiState) -> None:
         if state.config.make_initrd:
             return
 
-        invoke_emerge(state, sysroot=cls.stage3_cache, opts=opts,
-                      pkgs=["sys-kernel/gentoo-kernel-bin"])
+        if state.config.bootable == ConfigFeature.enabled and not state.staging.joinpath(state.config.output_split_uki).exists():
+            invoke_emerge(state, sysroot=cls.stage3_cache, opts=opts, pkgs=["sys-kernel/gentoo-kernel-bin"])
```
DaanDeMeyer: We do not install a kernel even if bootable is specified, again, please look at the other distributions, none of them do this, gentoo shouldn't do it either.

mkosi/distributions/gentoo.py
```diff
@@ -241,16 +241,14 @@ def install(cls, state: MkosiState) -> None:
         with complete_step("Merging bare minimal atoms"):
             invoke_emerge(state, sysroot=cls.stage3_cache,
                           opts=opts+["--exclude", "sys-devel/*"],
-                          pkgs=["app-shells/bash", "sys-apps/systemd"],
+                          pkgs=["sys-apps/systemd"],
```
DaanDeMeyer: We should also drop systemd by default, just like the other distributions

mkosi/distributions/gentoo.py
```diff
@@ -105,39 +119,41 @@ def install(cls, state: MkosiState) -> None:
...
 
         stage3_url_path = urllib.parse.urljoin(
             state.config.mirror, f"releases/{arch}/autobuilds/{stage3_tar}",
         )
-        if not stage3_tar_path.is_file():
-            log_step(f"Fetching {stage3_url_path}")
-            stage3_tar_path.parent.mkdir(parents=True, exist_ok=True)
-            urllib.request.urlretrieve(stage3_url_path, stage3_tar_path)
 
+        cls.stage3_cache = state.cache_dir.joinpath("stage3")
```
DaanDeMeyer, line 126: Because we hardcode it now, there's no need for cls.stage3_cache anymore, please drop the class variable

mkosi/distributions/gentoo.py
```diff
@@ -105,39 +119,41 @@ def install(cls, state: MkosiState) -> None:
...
-                "-C", cls.stage3_cache,
-                "--extract",
-                "--file", stage3_tar_path,
-                "--exclude", "./dev",
-            ])
-
-            unlink_try_hard(cls.stage3_cache.joinpath("dev"))
-            unlink_try_hard(cls.stage3_cache.joinpath("proc"))
-            unlink_try_hard(cls.stage3_cache.joinpath("sys"))
+        if not cls.stage3_cache.joinpath(".cache_isclean").exists():
```
DaanDeMeyer: Please add more newlines

DaanDeMeyer (changes requested):
One more round but this is becoming completely unreviewable. Please get CI passing and we'll merge this and I'll fix all the issues in a follow up PR

mkosi/distributions/gentoo.py
257: that would violate POLA, but sure.

mkosi/distributions/gentoo.py
257: ok!

257:
give me a bit to fix/rework repo and pkgmngr stuff.

mkosi/distributions/gentoo.py
257: so on the POLA violation:

we're really asking the user to go down the trial-and-error to find out that if he wants a bootable image he needs to do 2 things:
1. add `gnuefi` to `pkgmngr`
2. add `systemd` to `Packages=`. (actually `sys-apps/systemd` see below)

do we really want this? i for one can't think of a clean workaround for this. for instance, we check in `invoke_emerge()` to to see if users wants a bootable image and `systemd` was asked to be merged in `packages`. this is messy and error prone, for one `systemd` could mean `dev-ml/systemd`.

257: so we enforce `gnuefi` in `invoke_emerge()` and that should fix it. now let me check the CI failures.

mkosi/distributions/gentoo.py
257: see the latest:
1. we append(?) to our defaults whatever user puts in `pkgmngr`
2. we honor `--repositories` in `invoke_emerge()` but we don't document it; i.e. clobber the `pkgmngr` or our `stage3_cache`.

mkosi/distributions/gentoo.py
257: > i see. please see my question to @DaanDeMeyer above, regarding `install_package_manager_trees()`. that happens before things are handed over to the `installer()` instance. i revisit these as in #1640

https://github.com/systemd/mkosi/pull/1570#discussion_r1238413322

257: > This and some other instances further below clobber the files in `state.pkgmngr`.

see this commit: `gentoo: don't clobber pkgmngr, rework repositories`

mkosi/distributions/gentoo.py
257: should be fixed although there might be a more *pythonic* way to do this.

257:
you can look now :)

257:
@DaanDeMeyer tests pass here with systemd from git HEAD. could you have a look?

257:
> @DaanDeMeyer tests pass here with systemd from git HEAD. could you have a look?

spoke too realy, it still fails here too; never mind.

257:
[SUCCESS!](https://github.com/systemd/mkosi/actions/runs/5381155152)

lastest should be fine to merge.

257:
no idea why opensuse tests fail.

DaanDeMeyer:
> no idea why opensuse tests fail.

Not related

DaanDeMeyer (approved):

PR #1572, opened 2023-05-22
Use systemd-repart to grow disk image

DaanDeMeyer:
Using truncate isn't entirely right, let's do this properly by letting systemd-repart handle it.


DaanDeMeyer:
> Out of curiousity, what's the issue with truncate?

The partition table needs to be resized as well

PR #1573, opened 2023-05-22
Refactor image growing

DaanDeMeyer:


mkosi/__init__.py
```diff
@@ -1903,6 +1903,9 @@ def run_shell(args: MkosiArgs, config: MkosiConfig) -> None:
         cmdline += ["--"]
         cmdline += args.cmdline
 
+    if config.output_format == OutputFormat.disk:
+        grow_image(config.output_dir / config.output, size="8G")
```
behrmann, line 1907: Could that maybe bite us UX-wise? Booting the image with nspawn via mkosi will then balloon it to at least 8GB. Granted, the alternative could be that it doesn't work at all,

behrmann (approved):
LGTM, one question

DaanDeMeyer:
Yeah, but let's address that later

PR #1576, opened 2023-05-24
Allow negative matches

behrmann:


mkosi.md
```diff
@@ -287,30 +287,34 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
...
 
-: Matches against the configured distribution. Multiple distributions may
-  be specified, separated by spaces. If multiple distributions are specified,
-  the condition is satisfied if the current distribution equals any of the
-  specified distributions.
+: Matches against the configured distribution. Multiple distributions may be
+  specified, separated by spaces. A single distribution match can be negated by
+  preceding the value with `!`. If multiple distributions are specified, the
+  condition is satisfied if the current distribution equals any of the specified
+  distributions.
```
DaanDeMeyer, line 294: Does this make sense? If I do `Distribution=centos !fedora` the `!fedora` adds zero value, there's only value if all the individual distros are prefixed with`!` but in that case, a single `!` at the front of the condition that negates the full result will also do the trick. In other words, instead of allowing negation of individual list items, should we instead just support negating the full result of the condition?

mkosi.md
```diff
@@ -287,30 +287,34 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
...
-  is used and no distribution has been explicitly configured yet, the
-  host distribution and release are used. Multiple releases may be specified,
-  separated by spaces. If multiple releases are specified, the condition is
-  satisfied if the current release equals any of the specified releases.
+: Matches against the configured distribution release. If this condition is used
+  and no distribution has been explicitly configured yet, the host distribution
+  and release are used. A single release match can be negated by preceding the
+  value with `!`. Multiple releases may be specified, separated by spaces. If
+  multiple releases are specified, the condition is satisfied if the current
+  release equals any of the specified releases.
```
DaanDeMeyer, line 303: Does it make sense to support negation here since we already have the comparison operators? Technically you can already negate here by just using the reverse comparison operator right?

mkosi.md
```diff
@@ -287,30 +287,34 @@ a boolean argument: either "1", "yes", or "true" to enable, or "0",
...
+  condition is read from. A match can be negated by preceding the path with a
+  `!`.
 
 `ImageId=`
 
 : Matches against the configured image ID, supporting globs. If this condition
   is used and no image ID has been explicitly configured yet, this condition
-  fails. Multiple image IDs may be specified, separated by spaces. If multiple
+  fails.  A single image ID match can be negated by preceding the value with
+  `!`. Multiple image IDs may be specified, separated by spaces. If multiple
```
DaanDeMeyer, line 317: I think we can also get away here by allow negation of the full result but not individual items?

DaanDeMeyer (changes requested):
We should consider whether to allow only negating the full match result instead of individual items

mkosi.md
behrmann: The rich comparisons are for image versions. For releases we can't really do version compare, since for Debian these will be codenames.

mkosi.md
behrmann: I wondered about this myself and went for negating single items since that was easiest to implement, since it can be done at each step and we don't have to somehow track whether we're at the end and negate then. 

I agree this can have unforeseen edge cases and some are covered in the tests, but I figured it would cover most usage, since people will most often match a single thing or want to exclude a single thing, e.g. "Include this package for everything but Fedora (with everything probably being some RHEL derivative)" or "Include this package for this old release, but not in newer ones."

mkosi.md
DaanDeMeyer: Ah lol my bad misread this, but then the same argument I made for distributions applies here

mkosi.md
behrmann: In other words, I expect usage to look like this: You're a Fedora/RHEL shop, than the common match will look like either of these
```
Distribution=centos
Distribution=!fedora
```
which is why I think that usually negating a single thing is already negating the whole thing. 

Also, if only the whole thing can be negated we just switch the situation around and to solve that we'd need parentheses (which in a sense we already have in the sense of adding another file).

Cornelicorn: I think negating the whole thing would fit better.
If I want to exclude for example centos and fedora for some reason, this would not work with single negation, since
`!fedora || !centos = !(fedora && centos)`, so the condition is always true.
If the whole thing is negated, this would be `!(fedora || centos) = !fedora && !centos` and this exclusion would work.

DaanDeMeyer:
So what's the final take on this? It seems we still do individual negations instead of just negating the entire result of the match?

behrmann:
What's implemented here allows both. You can negate individual things or the whole thing if you put it in parentheses. I'm not in favour of just negating the whole thing for
```
Foo=!bar baz
```
since normally this would match `Foo` being `bar || baz` and people would have to know basic logic and add implicit parentheses to make this `!(bar || baz)` equalling `!bar && !baz`. Making the parentheses explicit should be a better UX.

behrmann:
Superseded by #1613

Issue #1569, opened 2023-05-20
btrfs subvolumes (UseSubvolumes=true) still not working


iam-TJ:
Trying the latest code it is still not possible to get a subvolume output; it is just a directory copy:
```
$ ls -idlatr mkosi.output/{.,debian}* mkosi.cache mkosi.workspace
 440832 drwxrwx---  1 root root       168 2021-04-27 10:46 mkosi.output/debian_bullseye_amd64_0.2
 457953 -rw-rw----  1 root root        62 2023-04-24 13:50 mkosi.output/debian_bullseye_amd64_0.2.nspawn
 440538 drwxr-xr-x  1 root root        70 2023-05-20 13:23 mkosi.output/debian~bullseye
 457954 -rw-r--r--  1 root root     17806 2023-05-20 13:23 mkosi.output/debian_bullseye_amd64_0.2.manifest
 457955 lrwxrwxrwx  1 root root        25 2023-05-20 13:23 mkosi.output/debian_bullseye_amd64 -> debian_bullseye_amd64_0.2
    256 drwxr-x---+ 1 root machines   484 2023-05-20 13:23 mkosi.workspace
    256 drwxr-x---+ 1 root machines   484 2023-05-20 13:23 mkosi.output/.
    256 drwxr-x---+ 1 root machines   484 2023-05-20 13:23 mkosi.cache
```
Each of `mkosi.cache` , `mkosi.workspace` and `mkosi.output` are bind mounts to `/var/lib/machines/` which is the root of a BTRFS, as can be seen by the inode numbers above (256).
```
$ mount | grep mkosi
/dev/mapper/SUNNY-Machines on /home/tj/iam/mkosi/debian-amd64-base/mkosi.workspace type btrfs (rw,relatime,space_cache,subvolid=5,subvol=/)
/dev/mapper/SUNNY-Machines on /home/tj/iam/mkosi/debian-amd64-base/mkosi.output type btrfs (rw,relatime,space_cache,subvolid=5,subvol=/)
/dev/mapper/SUNNY-Machines on /home/tj/iam/mkosi/debian-amd64-base/mkosi.cache type btrfs (rw,relatime,space_cache,subvolid=5,subvol=/)
```
I instrumented the code with a function tracer and there is only a single call related to BTRFS when the workspace is being created:
```
$ grep '^. TRACE:' /tmp/mkosi/mkosi.log  | grep -B 2 btrfs
[...]

PR #1580, opened 2023-05-25
btrfs: fix subvolume snapshots

iam-TJ:
Refactor and add option to fall-back to a move not just a copy_path. Always try to snapshot from workspace to the output directory but fall-back to using move.

Fixes: #1569

--

I explored extending `shutil.move()` with a copy function but that is only called if a rename cannot be done, and as a rename can be done, the copy function never gets called and therefore no way to use that to hook in the btrfs snapshot operation.

Since snapshot only applies to directories it is straight-forward to simply always call `btrfs_maybe_snapshot_subvolume()` and let its fall-back operation handle the not-a-subvol case.

Existing fall-back only does `copy_path()` so I extend the function parameters with `move: bool = False` so the caller can specify whether the fall-back will do `shutil.move()` or `copy_path()`.

I've also refactored to avoid duplicate code for handling the non-subvolume and fall-back-on-failure and lose the `result` value dance.

### DIFF
--- a/mkosi/btrfs.py
+++ b/mkosi/btrfs.py
@@ -23,26 +23,22 @@ def btrfs_maybe_make_subvolume(config: MkosiConfig, path: Path, mode: int) -> No
         path.mkdir(mode)
 
 
-def btrfs_maybe_snapshot_subvolume(config: MkosiConfig, src: Path, dst: Path) -> None:
-    subvolume = (config.use_subvolumes == ConfigFeature.enabled or
-                 config.use_subvolumes == ConfigFeature.auto and shutil.which("btrfs") is not None)
-
+def btrfs_maybe_snapshot_subvolume(config: MkosiConfig, src: Path, dst: Path, move: bool = False) -> None:
     if config.use_subvolumes == ConfigFeature.enabled and not shutil.which("btrfs"):
         die("Subvolumes requested but the btrfs command was not found")
 
-    # Subvolumes always have inode 256 so we can use that to check if a directory is a subvolume.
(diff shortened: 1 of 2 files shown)


mkosi/btrfs.py
```diff
@@ -23,26 +23,22 @@ def btrfs_maybe_make_subvolume(config: MkosiConfig, path: Path, mode: int) -> No
...
 
-    # btrfs can't snapshot to an existing directory so make sure the destination does not exist.
-    if dst.exists():
-        dst.rmdir()
+    # Subvolumes always have inode 256
+    if subvolume and src.stat().st_ino == 256:
+        if dst.exists():
+            dst.rmdir()
+        if run(["btrfs", "subvolume", "snapshot", src, dst],
+           check=config.use_subvolumes == ConfigFeature.enabled).returncode == 0:
```
behrmann, line 38:
```suggestion
        if run(["btrfs", "subvolume", "snapshot", src, dst],
               check=config.use_subvolumes == ConfigFeature.enabled).returncode == 0:
```

behrmann (commented):
Could you add some comments or a longer discussion to the commit message, similar to what you describe here. This way the next person touching the code can have a bit more context n why it looks the way it does.

keszybz:
Needs a rebase now.

DaanDeMeyer:
I still need to take a look at this, pretty sure we can find a simpler way. Please don't merge this just yet.

DaanDeMeyer:
Let's close this in favor of #1692

PR #1582, opened 2023-05-25
Fix cache directory logic

DaanDeMeyer:
Let's make sure we don't create nested per distribution, release directories if --cache-dir already points to such a directory and let's make sure both the regular user and the newuidmap root user can look into the workspace to fix the cache logic if no cache directory is set.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -495,8 +495,7 @@ def install_build_dest(state: MkosiState) -> None:
         return
 
     with complete_step("Copying in build tree…"):
-        # The build is executed as a regular user, so we don't want to copy ownership in this scenario.
-        copy_path(install_dir(state), state.root, preserve_owner=False)
+        copy_path(state.install_dir, state.root)
 
 
 def gzip_binary() -> str:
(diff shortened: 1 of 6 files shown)


behrmann:
I was thinking about this just yesterday. Right now the docs don't mention that these directories are namespaced, but mention it for the output directory, where we did away with that. I was thinking, that we should maybe do away with this namespacing in the case that the directory was explicitly set and only do this type of namespacing in the default case or when working in `mkosi.foo` directories. I think this would make this easier to use, and users would not have to arrange the names of their directories to match this scheme; we'd just say "when you set directories explicitly, you're on your own".

DaanDeMeyer:
> I was thinking about this just yesterday. Right now the docs don't mention that these directories are namespaced, but mention it for the output directory, where we did away with that. I was thinking, that we should maybe do away with this namespacing in the case that the directory was explicitly set and only do this type of namespacing in the default case or when working in mkosi.foo directories. I think this would make this easier to use, and users would not have to arrange the names of their directories to match this scheme; we'd just say "when you set directories explicitly, you're on your own".

Yeah I agree, but let's do that in a separate PR.

DaanDeMeyer:
Reworked this a bit to clean up the directory stuff a lot more in general.

behrmann (approved):

PR #1583, opened 2023-05-26
Improve basename symlinks

Cornelicorn:
Fixes the symlink having the wrong uid:gid (root inside the uid namespace)
Always creates the symlink so it points to the newest version of the image
The link is created in the staging area first to make the change atomic


Cornelicorn:
This poses the problem that the initrd is generated earlier than the actual image, so there is a time where `image-initrd` points to the newest version, which does not exist yet for `image.raw`

Cornelicorn:
As requested by @behrmann, I split the commit into two, one for the uid:gid fix and one for overwriting the symlink to the current version.

DaanDeMeyer (approved):

PR #1584, opened 2023-05-26
Support apt pinning

Cornelicorn:


behrmann:
While this is useful, it's too specific. Adding a directory to pass config to package managers to generally has been discussed, but hasn't been implemeneted yet, though this could be a starting point.

Cornelicorn:
The problem is that apt would still need extra care, since it needs a separate `preferences.d` besides `apt.conf.d`
Except the config passing directly takes `/etc/apt` instead `/etc/apt/apt.conf.d`, which of course would be possible.

Cornelicorn:
For Arch, `RepositoryDirectories` essentially already provides the possibility to set arbitrary config options, so this can just be duplicated and renamed.

For apt based distros, it would be possible to install every directory into `/etc/apt` before or after templating mkosi's "native" configuration there.

For rpm based distros, I guess just appending all the contents of a config file directory to the configuration file would be possible.

Don't have any knowledge, how this could be done with mageiea or openmandriva, but these two also already are missing some other options, e.g. `RepositoryDirectories`.


I saw there are a few issues for discussing this, but each restricted to a single requested option, pretty much all solvable by providing config overrides. Maybe this should be moved into a new issue, where the wanted implementation can be discussed?

DaanDeMeyer:
I opened https://github.com/systemd/mkosi/issues/1603 for discussing this, let's close this PR as I'm pretty sure its not the way to go

PR #1585, opened 2023-05-26
debian: Add --force-architecture to dpkg options

DaanDeMeyer:
This fixes builds for foreign architectures

### DIFF
--- a/mkosi/distributions/debian.py
+++ b/mkosi/distributions/debian.py
@@ -170,6 +170,7 @@ def setup_apt(state: MkosiState, repos: Sequence[str]) -> None:
         dedent(
             f"""\
             APT::Architecture "{debarch}";
+            APT::Architectures "{debarch}";
             APT::Immediate-Configure "off";
             APT::Install-Recommends "false";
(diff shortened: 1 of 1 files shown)


behrmann:
Could we gate the `--force-architecture` and `--force-depends` to only builds where the target architecture is not the same as the host architecture?

bluca (approved):
fixes my issue, thanks

DaanDeMeyer:
> Could we gate the --force-architecture and --force-depends to only builds where the target architecture is not the same as the host architecture?

I think these are mostly harmless? --force-depends only does something if dpkg can't figure it out on its own and I think we prefer having dpkg try something in that case rather than hard failing

behrmann:
> I think these are mostly harmless? --force-depends only does something if dpkg can't figure it out on its own and I think we prefer having dpkg try something in that case rather than hard failing

I'm not entirely sure that downgrading all dependency problems to warnings is desirable and I'm also not entirely sure about ignoring all architecture constraints, but we can run with it and see whether anybody reports a bug.

PR #1587, opened 2023-05-26
bootctl: use '--all-architectures' for cross-builds

bluca:
Do not restrict to the native architecture, if there's an EFI binary install it. Even if there are multiple ones it won't cause issues, as they are all suffixed by the architecture.


DaanDeMeyer:
Unfortunately a duplicate

PR #1590, opened 2023-05-26
repart: explicitly use arch in root repart conf file

bluca:
'root' will resolve to the native architecture, but when --architecture is used we need to specify a foreign one, so always specify it. Requires commit 08a2bb7b82 in systemd

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -9,6 +9,7 @@ import itertools
 import json
 import logging
 import os
+import platform
 import re
 import resource
 import shutil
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
So I was thinking about this. And I think we should add an --architecture switch to repart to modify the arch used by the arch independent partition types. Then we set that from mkosi and fix the issue not just for the mkosi generated config files but also for custom ones

bluca:
sure, but until that happens, can we go for this so that cross-builds can work?

DaanDeMeyer:
Can we only do this when the host architecture does not match the target architecture? Otherwise everyone is going to have to pull in the very latest repart because on x86 this will only work properly with the alias for x86_64 you added.

bluca:
sure, done

DaanDeMeyer:
https://github.com/systemd/mkosi/pull/1592 takes care of this and more

DaanDeMeyer:
Let's close this in favor of https://github.com/systemd/mkosi/pull/1592

PR #1591, opened 2023-05-27
Make sure we handle any exceptions thrown in MkosiAsyncioThread()

DaanDeMeyer:
Any unhandled exceptions from run() are by default just printed to stderr. Let's make sure that these exceptions cause mkosi itself to fail by catching them and re-raising them when the thread is joined.


DaanDeMeyer:
Merging early to debug stuff in systemd

PR #1592, opened 2023-05-29
Rework architecture handling

DaanDeMeyer:
Let's lock architectures down by making it an enum instead of a free form string. We also introduce a bunch of mapping functions to map the Architecture enum to qemu, distribution arches, efi arches and the discoverable partitions spec arches.

### DIFF
--- a/mkosi/architecture.py
+++ b/mkosi/architecture.py
@@ -1,0 +1,131 @@
+# SPDX-License-Identifier: LGPL-2.1+
+
+import enum
+import platform
+
+from mkosi.log import die
+
+
+class Architecture(enum.Enum):
+    alpha       = "alpha"
+    arc         = "arc"
+    arm         = "arm"
+    arm64       = "arm64"
+    ia64        = "ia64"
+    loongarch64 = "loongarch64"
+    mips_le     = "mips-le"
+    mips64_le   = "mips64-le"
+    parisc      = "parisc"
+    ppc         = "ppc"
+    ppc64       = "ppc64"
+    ppc64_le    = "ppc64-le"
+    riscv32     = "riscv32"
+    riscv64     = "riscv64"
+    s390        = "s390"
+    s390x       = "s390x"
+    tilegx      = "tilegx"
+    x86         = "x86"
+    x86_64      = "x86-64"
+
+    def __str__(self) -> str:
+        return self.value
+
+    @staticmethod
+    def from_uname(s: str) -> "Architecture":
+        a = {
+            "aarch64"     : Architecture.arm64,
+            "aarch64_be"  : Architecture.arm64,
+            "armv8l"      : Architecture.arm,
+            "armv8b"      : Architecture.arm,
+            "armv7ml"     : Architecture.arm,
+            "armv7mb"     : Architecture.arm,
+            "armv7l"      : Architecture.arm,
+            "armv7b"      : Architecture.arm,
+            "armv6l"      : Architecture.arm,
+            "armv6b"      : Architecture.arm,
+            "armv5tl"     : Architecture.arm,
+            "armv5tel"    : Architecture.arm,
+            "armv5tejl"   : Architecture.arm,
+            "armv5tejb"   : Architecture.arm,
+            "armv5teb"    : Architecture.arm,
+            "armv5tb"     : Architecture.arm,
+            "armv4tl"     : Architecture.arm,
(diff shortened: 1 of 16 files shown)


mkosi/architecture.py
```diff
@@ -0,0 +1,187 @@
+import enum
```
bluca, line 1: license

mkosi/distributions/opensuse.py
```diff
@@ -51,6 +52,17 @@ def install_packages(cls, state: MkosiState, packages: Sequence[str], apivfs: bo
     def remove_packages(cls, state: MkosiState, packages: Sequence[str]) -> None:
         invoke_dnf(state, "remove", packages)
 
+    @staticmethod
+    def architecture(arch: Architecture) -> str:
+        a = {
+            Architecture.x86_64 : "x86_64",
```
bluca, line 58: suse supports the same as fedora, give or take

DaanDeMeyer: Yes but it'll require different mirrors or something like that which I'll leave for suse folks to figure out.

bluca: No, it's all on the same download server

mkosi/architecture.py
```diff
@@ -0,0 +1,189 @@
...
+        a = {
+            Architecture.x86_64  : "x64",
+            Architecture.x86     : "ia32",
+            Architecture.arm64   : "aa64",
+            Architecture.arm     : "arm",
+            Architecture.riscv64 : "riscv64",
+        }.get(self)
+
+        if not a:
+            die(f"Architecture {self} does not support by UEFI")
```
medhefgo, line 146: That's some interesting english.

mkosi/distributions/opensuse.py
DaanDeMeyer: http://download.opensuse.org/tumbleweed/repo/oss/

I don't see anything other than x86_64 here. The ports are there but they will require more changes so I'd like to leave that for another PR.

mkosi/architecture.py
```diff
@@ -0,0 +1,133 @@
...
+    s390x       = "s390x"
+    tilegx      = "tilegx"
+    x86         = "x86"
+    x86_64      = "x86-64"
+
+    def __str__(self) -> str:
+        return self.value
+
+    @staticmethod
+    def from_uname(s: str) -> "Architecture":
```
behrmann, line 34:
```suggestion
    @classmethod
    def from_uname(cls, s: str) -> "Architecture":
        if not (a := getattr(cls, s)):
            die(f"Architecture {a} is not supported")
        return a
```
would be a possibility if the unames were entered as aliases above, e.g.
```python

class Architecture(enum.Enum):
    alpha       = "alpha"
    arc         = "arc"
    arm         = "arm"
    arm64       = "arm64"
    ia64        = "ia64"
    loongarch64 = "loongarch64"
    mips_le     = "mips-le"
    mips64_le   = "mips64-le"
    parisc      = "parisc"
    ppc         = "ppc"
    ppc64       = "ppc64"
    ppc64_le    = "ppc64-le"
    riscv32     = "riscv32"
    riscv64     = "riscv64"
    s390        = "s390"
    s390x       = "s390x"
    tilegx      = "tilegx"
    x86         = "x86"
    x86_64      = "x86-64"

    # aliases
    aarch64 = "arm64"
    aarch64_be = "arm64"
```

They are not iterated over and also don't appear in most tab completions, but `Architecture.aarch64` would resolve to `Architecture.arm64` right away. Unfortunately it doesn't come with a string constructor out of the box, but it would remove a bit of the repetition.

mkosi/architecture.py
```diff
@@ -0,0 +1,133 @@
...
+        if not a:
+            die(f"Architecture {self} not supported by QEMU")
+
+        return a
+
+    def is_native(self) -> bool:
+        return self == native_architecture()
+
+
+def native_architecture() -> Architecture:
```
behrmann, line 132: As a alternative constructor I think that should live in `Architecture`.
```suggestion
@ classmethod
def native_architecture(cls) -> "Architecture":
    return cls.from_uname(platform.machine())
```

behrmann (commented):
Looks sensible to me.

DaanDeMeyer:
Added one more trivial commit to not use dnf5 when building for foreign architectures since it doesn't support that yet.

PR #1593, opened 2023-05-30
Use repart's new --architecture option

DaanDeMeyer:
This makes sure that --architecture works with user specified repart configs as well instead of just the ones that we generate ourselves.


DaanDeMeyer:
Depends on https://github.com/systemd/systemd/pull/27844

DaanDeMeyer:
The systemd PR was merged so let's merge this one as well

PR #1595, opened 2023-05-30
README.md: Update instructions how to run mkosi

Werkov:
This has changed after commit 04d5c24 ("Use an entry point instead of a script"). Update the readme to avoid confusion of users running into various Python packaging error messages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,11 +25,10 @@ python3 -m pip install --user git+https://github.com/systemd/mkosi.git
 If you want to hack on mkosi do
 ```shell
 # clone either this repository or your fork of it
-git clone https://github.com/systemd/mkosi/
-cd mkosi
-python3 -m pip install --user --no-use-pep517 --editable .
+git clone https://github.com/systemd/mkosi/ <PROJECT_PATH>
+[sudo|pkexec] <PROJECT_PATH>/bin/mkosi
 ```
-This will install mkosi in editable mode to `~/.local/bin/mkosi`, allowing you
+This will run mkosi as present in the module source, allowing you
 to use your own changes right away.
 
 For development you also need [mypy](https://github.com/python/mypy), for type


README.md
```diff
@@ -25,11 +25,10 @@ python3 -m pip install --user git+https://github.com/systemd/mkosi.git
 If you want to hack on mkosi do
 ```shell
 # clone either this repository or your fork of it
-git clone https://github.com/systemd/mkosi/
-cd mkosi
-python3 -m pip install --user --no-use-pep517 --editable .
+git clone https://github.com/systemd/mkosi/ <PROJECT_PATH>
+[sudo|pkexec] <PROJECT_PATH>/bin/mkosi
```
behrmann, line 29: mkosi is able to work unprivileged now, so privilege escalation is not necessary. Also, the shim is just that and installing with pip (and running via `python -m mkosi` should work fine. Have you experienced any issue with this?

Werkov: `mkosi boot` (and perhaps `shell` too) must still be invoked as root (that's what `mkosi` said.)

Note the `sudo` is optional in the instructions, the command is taken verbatim from the commit 04d5c247e22f1c02f5311677c342982ac8c7a792.

When I ran the original `sudo ~/.local/bin/mkosi`  I ran into error messages with:
> importlib.metadata.PackageNotFoundError: No package metadata was found for mkosi

(Yeah, unpriviliged invocation may work but it didn't cut it for my `boot` usage.)

Werkov: Also, calling a script from cloned repository instead of "polluting" home with a user-installed module seems a nicer way for quick trials.

behrmann: > Also, calling a script from cloned repository instead of "polluting" home with a user-installed module executable seems a nicer way for quick trials.

Let's agree to disagree here. :) I for one prefer a proper installation than calling shims in random paths.

>  importlib.metadata.PackageNotFoundError: No package metadata was found for mkosi

That's because mkosi the library is not in the `sys.path` of `root`. The shim works around this by setting `PYTHONPATH` to the parent of where the script is (assuming it stays in the repo), but you can just set it like that yourself. You can `sudo python -m mkosi` in the repo yourself or install it into virtual environment instead of your user environment.

Werkov: Think of this use case:
- a random person wants to try/use latest version of mkosi
- they clone the repo, check README that tells them to 
```
python3 -m pip install --user --no-use-pep517 --editable .
```
- they run it arbitrary work directory, "boot" command
- command instructs them to run as root
- they invoke with `sudo`
- and it's broken

That is not very smooth user experience.

Do you suggest the following in the README
```
PYTHONPATH=<PROJECT PATH> sudo --preserve-env=PYTHONPATH ~/.local/bin/mkosi  -i -T boot
```
?

What was the point of 04d5c247e22f1c02f5311677c342982ac8c7a792 if it's fruits can't be leveraged? (Cc: @gdiscry)

behrmann: I agree that the README needs to be updated, but I disagree that the shim is the primary entry point that should be advertised.

If you just clone the repo, enter it and run `python -m mkosi` this will work with and without sudo. If you 
```
python -m venv testvenv
testvenv/bin/python -m pip install git+https://github.com/systemd/mkosi
```
you can `testvenv/bin/mkosi` with and without sudo.

But the user experience is a bit rocky in that regard anyway, since mkosi will need pretty recent systemd components, that require building them from source on most distros. :)

keszybz: I think this patch DTRT. Calling `~/projects/mkosi/bin/mkosi`, with or without `sudo`, is something that can be memorized and makes sense. OTOH, some insane `pip` command with four enigmatic options is something that can be only copied from the docs, on every use. If it worked, then OK, we could discuss this despite the ugliness, but it doesn't actually work, because you need to set `$PYTHONPATH`. And I agree that having the module "installed" (which doesn't really mean installed, but just symlinked), where the symlink is very hard to discover and clean up, is just not good.

I don't think we should recommend using `pip` installations. Either the user does a clone and calls our helper, and everything is contained within one directory and works without any issues, or they install a distro package and all files are accounted for and can be cleaned up. Let's not encourage other hacky solutions that are neither this nor that.

README.md
```diff
@@ -25,11 +25,10 @@ python3 -m pip install --user git+https://github.com/systemd/mkosi.git
...
 ```
-This will install mkosi in editable mode to `~/.local/bin/mkosi`, allowing you
+This will run mkosi as present in the module source, allowing you
```
keszybz, line 31: "module source" is python jargon. Maybe just say "This will run mkosi from this checkout, …" ?

README.md
behrmann: > OTOH, some insane pip command with four enigmatic options is something that can be only copied from the docs, on every use. If it worked, then OK, we could discuss this despite the ugliness, but it doesn't actually work, because you need to set $PYTHONPATH. 

I'm not sure I get this? 
```
python -m venv testvenv
testvenv/bin/python -m pip install git+https://github.com/systemd/mkosi
```
might be a bit verbose, sorry, I'm got into the habit of never calling pip directly, but it's equivalent to
```
python -m venv testvenv
testvenv/bin/pip install git+https://github.com/systemd/mkosi
```
which I hardly think is insane.

It also doesn't need messing around with `$PYTHONPATH`, since python, when called with a prefix of a virtual environment will set its path to the corresponding directory in the virtual environment. 

What will not work is a plain `pip install git+https://github.com/systemd/mkosi`, for the above mentioned reason, which would necessitate messing around with `$PYTHONPATH`, but that's not what I said.

keszybz: Right, but what's the *point*? Does this behave better than the trivial call to `~/something/bin/mkosi`?

behrmann: Another alternative, shorter still:
```
pipx install git+https://github.com/systemd/mkosi
```
which does the above explicit virtual environment under the hood and also works with sudo. 

For hacking around with mkosi the shim is totally servicable, the point I'm trying to make is that advertising the shim as the canonical way to run mkosi is reinventing the wheel and more manual than what maybe more regular users might expect.

keszybz: OK. I looked at the README, and it lists all the messy ways to install via pip, but doesn't actually say that the distro packages are the primary way to install. I think we should recommend the distro method as the primary way. And then we have the case where you want to develop or hack on the project, and then `git clone && bin/mkosi` is the way to go. All the other ways can be described in an appendix somewhere.

behrmann: > Right, but what's the point? Does this behave better than the trivial call to ~/something/bin/mkosi?

Well, it's nothing that we have to maintain and users have to think about that's different from how other projects work.

Also, if we were to add the cryptography library as an (optional) dependency in our `setup.py` (or `pyproject.toml` if we were to make that switch finally), something like 
```
pipx install git+https://github.com/systemd/mkosi
```
would work out of the box and produce a fully self-contained mkosi, that can be just as easily gotten rid of. This should also be closer to something distros package. Last time I looked debhelper mostly created the packaging scaffolding from the regular Python packaging inputs and in Arch the packages are mostly a repackaging of the wheel that's generated, if my understanding is correct.

DaanDeMeyer:
Let's close this as the other PR was merged

Issue #236, opened 2018-02-12
[Meta] declare a policy about adding new distributions


r-darwish:
Before people start creating issues asking to support their favorite distribution, I think that mkosi should declare its policy regarding new distributions support.

The policy should state in which terms (if any) you will be willing to support a new distributions.

poettering:
yeah, it's a difficult question where to draw the line. And I figure when adding this we should also open the door for smaller distros to hook things up to mkosi too, by adding some form of drop-in logic, where additional distros could be added by dropping in some minimal file that does the minimal core bits to hook things up...

Issue #101, opened 2017-06-29
Provide a way to provide build scripts for unsupported distros


madsl:
It would be awesome if mkosi could be made distribution agnostic! What if there were a distro option called "custom"? So that when invoking mkosi with an option like "--distribution=custom", you have to add some mandatory options like e.g. --image-devel-script and --image-final-script which both points to scripts that does the caching and package managing manually?

I just stumbled upon this project, so I do not yet have a full overview how many steps that would need to have a custom script to make this work. I just thought I should throw this suggestion out to get any input before trying to create a patch :)

luntik2012:
also, please, provide a way to remove system from the internet

fsateler:
Why not simply add support for your distro to mkosi?

madsl:
Lennart has already added most of the distributions already that has a straight-forward predefined way to install a base system, but not all distributions are like that. The use case might be doing something custom with yocto or buildroot, maybe OpenWRT, or even Gentoo. To code support for all these different ways of putting together a distro seems futile, so a better solution might be to provide a easy way to hook in scripts so you can define these targets yourself...

poettering:
so what about this: if a distribution is selected that we do not know about right now, we look for a file `mkosi.distribution-<name>` in the local directory, and if it exists and is executable we invoke that for installing the distribution.

That said, I think add support for better known distros such as yocto or Gentoo directly would make a ton of sense, hence I'd suggest to submit patches for them directly.

madsl:
The mkosi.distribution-<name> solutions sounds perfect!

I agree that it would be best to have proper distro support in mkosi too, but I guess that'll we have a lot of bikeshedding before us if mkosi should be made aware of all the sets of config files needed to build images for source based distros...

I think the best way is to begin with the mkosi.distribution-<name>-solution anyway, the scripts needed to add support that way could be a base for further support work! I'll make sure to share my scripts in that regard at least :)

dteleguin:
Hi Lennart @poettering, are you interested in Mageia Linux support?

We are nearing Mageia 6 release now, with one of its highlights being native DNF support out of the box. Hence, the process shouldn't be much different for Fedora. However, we have systemd 230. Will mkosi operation still be possible, maybe in some limited manner?

If positive, I can contribute a patch for mkosi + Mageia RPM package.

lucaswerkmeister:
[...]

PR #1596, opened 2023-05-30
Opensuse: use zypper if present, fallback to dnf if not

Werkov:
The reasoning in #1536 perhaps considers only cross-distro builds.
zypper is just fine when building openSUSE image on openSUSE host and yet another package manager needn't be installed on the host. (I'm not even sure dnf would run various transaction hooks/scriptlets equally as zypper.)

The commit dc8560f049b220f3b67ce201acf5e75081ed1c8d breaks mkosi image builds on my openSUSE host where I don't deem dnf necessary.

This is a revert that enables oS builds on oS again (mkosi from `main` (8e9f84d), systemd image from its `main` (.600bf76c17).)

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -46,7 +46,8 @@ runs:
         squashfs-tools \
         swtpm \
         systemd-container \
-        xfsprogs
+        xfsprogs \
+        zypper
 
       sudo pacman-key --init
       sudo pacman-key --populate archlinux
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -254,7 +254,7 @@ def config_default_mirror(namespace: argparse.Namespace) -> Optional[str]:
         else:
             return "https://geo.mirror.pkgbuild.com"
     elif d == Distribution.opensuse:
-        return "https://download.opensuse.org"
+        return "http://download.opensuse.org"
     elif d == Distribution.fedora and r == "eln":
         return "https://odcs.fedoraproject.org/composes/production/latest-Fedora-ELN/compose"
     elif d == Distribution.gentoo:
--- a/mkosi/distributions/opensuse.py
+++ b/mkosi/distributions/opensuse.py
@@ -1,11 +1,14 @@
 # SPDX-License-Identifier: LGPL-2.1+
 
+import shutil
+import urllib.request
+import xml.etree.ElementTree as ElementTree
 from collections.abc import Sequence
 from textwrap import dedent
 
 from mkosi.architecture import Architecture
 from mkosi.distributions import DistributionInstaller
-from mkosi.distributions.fedora import fixup_rpmdb_location
+from mkosi.distributions.fedora import Repo, fixup_rpmdb_location, invoke_dnf, setup_dnf
 from mkosi.log import die
 from mkosi.run import bwrap
 from mkosi.state import MkosiState
@@ -42,16 +45,27 @@ class OpensuseInstaller(DistributionInstaller):
             release_url = f"{state.config.mirror}/distribution/leap/{release}/repo/oss/"
             updates_url = f"{state.config.mirror}/update/leap/{release}/oss/"
 
-        repos = [("repo-oss", release_url)]
-        if updates_url is not None:
-            repos += [("repo-update", updates_url)]
+        if shutil.which("zypper") is not None:
+            repo_list = [("repo-oss", release_url)]
+            if updates_url is not None:
+                repo_list += [("repo-update", updates_url)]
 
-        setup_zypper(state, repos)
-        invoke_zypper(state, "install", ["-y", "--download-in-advance", "--no-recommends"], packages, apivfs=apivfs)
+            setup_zypper(state, repo_list)
+            invoke_zypper(state, "install", ["-y", "--download-in-advance", "--no-recommends"], packages, apivfs=apivfs)
+        else:
+            repos = [Repo("repo-oss", f"baseurl={release_url}", fetch_gpgurls(release_url))]
+            if updates_url is not None:
+                repos += [Repo("repo-update", f"baseurl={updates_url}", fetch_gpgurls(updates_url))]
+
+            setup_dnf(state, repos)
+            invoke_dnf(state, "install", packages, apivfs=apivfs)
 
     @classmethod
     def remove_packages(cls, state: MkosiState, packages: Sequence[str]) -> None:
-        invoke_zypper(state, "remove", ["-y", "--clean-deps"], packages)
+        if shutil.which("zypper") is not None:
+            invoke_zypper(state, "remove", ["-y", "--clean-deps"], packages)
+        else:
+            invoke_dnf(state, "remove", packages)
 
     @staticmethod
     def architecture(arch: Architecture) -> str:
@@ -63,6 +77,25 @@ class OpensuseInstaller(DistributionInstaller):
             die(f"Architecture {a} is not supported by OpenSUSE")
 
         return a
+
+
+def fetch_gpgurls(repourl: str) -> list[str]:
+    gpgurls = [f"{repourl}/repodata/repomd.xml.key"]
+
+    with urllib.request.urlopen(f"{repourl}/repodata/repomd.xml") as f:
+        xml = f.read().decode()
+        root = ElementTree.fromstring(xml)
+
+        tags = root.find("{http://linux.duke.edu/metadata/repo}tags")
+        if not tags:
+            die("repomd.xml missing <tags> element")
+
+        for child in tags.iter("{http://linux.duke.edu/metadata/repo}content"):
+            if child.text and child.text.startswith("gpg-pubkey"):
+                gpgkey = child.text.partition("?")[0]
+                gpgurls += [f"{repourl}{gpgkey}"]
+
+    return gpgurls
 
 
 def setup_zypper(state: MkosiState, repos: Sequence[tuple[str, str]] = ()) -> None:
--- a/mkosi/distributions/opensuse.py
+++ b/mkosi/distributions/opensuse.py
@@ -1,14 +1,15 @@
 # SPDX-License-Identifier: LGPL-2.1+
 
-import urllib.request
-import xml.etree.ElementTree as ElementTree
 from collections.abc import Sequence
+from textwrap import dedent
 
 from mkosi.architecture import Architecture
 from mkosi.distributions import DistributionInstaller
-from mkosi.distributions.fedora import Repo, invoke_dnf, setup_dnf
+from mkosi.distributions.fedora import fixup_rpmdb_location
 from mkosi.log import die
+from mkosi.run import bwrap
 from mkosi.state import MkosiState
+from mkosi.types import PathString
 
 
 class OpensuseInstaller(DistributionInstaller):
@@ -35,22 +36,22 @@ class OpensuseInstaller(DistributionInstaller):
             release_url = f"{state.config.mirror}/tumbleweed/repo/oss/"
             updates_url = f"{state.config.mirror}/update/tumbleweed/"
         elif release in ("current", "stable"):
-            release_url = f"{state.config.mirror}/distribution/openSUSE-{release}/repo/oss/"
+            release_url = f"{state.config.mirror}/distribution/openSUSE-stable/repo/oss/"
             updates_url = f"{state.config.mirror}/update/openSUSE-{release}/"
         else:
             release_url = f"{state.config.mirror}/distribution/leap/{release}/repo/oss/"
             updates_url = f"{state.config.mirror}/update/leap/{release}/oss/"
 
-        repos = [Repo("repo-oss", f"baseurl={release_url}", fetch_gpgurls(release_url))]
+        repos = [("repo-oss", release_url)]
         if updates_url is not None:
-            repos += [Repo("repo-update", f"baseurl={updates_url}", fetch_gpgurls(updates_url))]
+            repos += [("repo-update", updates_url)]
 
-        setup_dnf(state, repos)
-        invoke_dnf(state, "install", packages, apivfs=apivfs)
+        setup_zypper(state, repos)
+        invoke_zypper(state, "install", ["-y", "--download-in-advance", "--no-recommends"], packages, apivfs=apivfs)
 
     @classmethod
     def remove_packages(cls, state: MkosiState, packages: Sequence[str]) -> None:
-        invoke_dnf(state, "remove", packages)
+        invoke_zypper(state, "remove", ["-y", "--clean-deps"], packages)
 
     @staticmethod
     def architecture(arch: Architecture) -> str:
@@ -64,20 +65,56 @@ class OpensuseInstaller(DistributionInstaller):
         return a
 
 
-def fetch_gpgurls(repourl: str) -> list[str]:
-    gpgurls = [f"{repourl}/repodata/repomd.xml.key"]
+def setup_zypper(state: MkosiState, repos: Sequence[tuple[str, str]] = ()) -> None:
+    with state.workspace.joinpath("zypp.conf").open("w") as f:
+        f.write(
+            dedent(
+                f"""\
+                [main]
+                rpm.install.excludedocs = {"no" if state.config.with_docs else "yes"}
+                """
+            )
+        )
 
-    with urllib.request.urlopen(f"{repourl}/repodata/repomd.xml") as f:
-        xml = f.read().decode()
-        root = ElementTree.fromstring(xml)
+    state.workspace.joinpath("zypp.repos.d").mkdir(exist_ok=True)
 
-        tags = root.find("{http://linux.duke.edu/metadata/repo}tags")
-        if not tags:
-            die("repomd.xml missing <tags> element")
+    with state.workspace.joinpath("zypp.repos.d/mkosi.repo").open("w") as f:
+        for id, url in repos:
+            f.write(
+                dedent(
+                    f"""\
+                    [{id}]
+                    name={id}
+                    baseurl={url}
+                    autorefresh=0
+                    enabled=1
+                    keeppackages=1
+                    """
+                )
+            )
 
-        for child in tags.iter("{http://linux.duke.edu/metadata/repo}content"):
-            if child.text and child.text.startswith("gpg-pubkey"):
-                gpgkey = child.text.partition("?")[0]
-                gpgurls += [f"{repourl}{gpgkey}"]
 
-    return gpgurls
+def invoke_zypper(
+    state: MkosiState,
+    verb: str,
+    options: Sequence[str],
+    packages: Sequence[str],
+    apivfs: bool = True
+) -> None:
+    cmdline: list[PathString] = [
+        "zypper",
+        "--root", state.root,
+        f"--cache-dir={state.cache_dir}",
+        f"--reposd-dir={state.workspace / 'zypp.repos.d'}",
+        "--gpg-auto-import-keys" if state.config.repository_key_check else "--no-gpg-checks",
+        "--non-interactive",
+        verb,
+        *options,
+        *packages,
+    ]
+
+    env = dict(ZYPP_CONF=str(state.workspace / "zypp.conf"), KERNEL_INSTALL_BYPASS="1") | state.environment
+
+    bwrap(cmdline, apivfs=state.root if apivfs else None, env=env)
+
+    fixup_rpmdb_location(state.root)


Werkov:
Having both tools sounds like a good middle ground. I'm converting this PR to draft before I get down to it.

> I'm pretty sure everyone that doesn't run opensuse themselves would much rather use dnf rather than zypper.

I'd say people building cross-distro images would prefer image-distro's pkg manager shipped with their buildhost-distro. But this is all subjective. 

> zypper is extremely unreliable in CI, and is prone to fail with HTTP/2 protocol errors. 

I remember there were some troubles with the hosting infrastructure. Do you suggest it is a bug on the client side? Do you have any more pointers? (Or a bug report :-p)

keszybz:
> zypper is just fine when building openSUSE image on openSUSE host and yet another package manager needn't be installed on the host.

Hmm, I think this is a pretty weak justification. `mkosi` requires and will always require various tools to be installed. And the dnf dependency shouldn't be a big problem. We need rpm anyway, and also python for mkosi, so it's just a matter of a few packages that are available in the distro.

Adding back support for zypper as an option is not very attractive. It makes the code more complicated, and this additonal code path will not be tested in CI: after all, the initial impetus for the change came from the CI being unreliable.

bluca:
Are you really sure those errors were really because of zypper? Unlike other distros, OBS publishes new repos all the time, so I imagine the likelihood of hitting the repo at exactly the wrong time when doing large installations continuosly becomes non-negligible. Debian updates repos twice a day or so, so it's really much more difficult to catch the server-side in an inconsistent state (but it can happen)

keszybz:
> Are you really sure those errors were really because of zypper?

It was a bug in http transport code.

DaanDeMeyer:
> It's all self-contained in the suse-specific class file as far as I can see from this PR? It doesn't seem that complicated at least to me

The main thing is that any new features that involve package managers have to be implemented for both zypper and dnf. There's been users asking for more package manager related features and if we can implement them for dnf and take advantage of them for OpenSUSE as well, that's a win

Werkov:
Having code for all distros in one mkosi repo is a liability, as you hint it restricts mkosi features development because it must be added/exempted for each of them. I think the distro specific code should be shipped and maintained within distros and mkosi would provide API. (I don't mean reinventing dracut.) I'm digressing -- #101 and #236 seems like a better (yet staler) place to discuss this.

So, do you have reliable confirmation that the CI issues are gone since the zypper->dnf switch? (I.e. it was a client-side bug. Still, I'd be glad for any more details that could be passed to zypper folks.)

DaanDeMeyer:
> So, do you have reliable confirmation that the CI issues are gone since the zypper->dnf switch? (I.e. it was a client-side bug. Still, I'd be glad for any more details that could be passed to zypper folks.)

I haven't seen a single opensuse CI failure due to opensuse mirror issues since the switch so I'm very confident its fixed with dnf

There's an upstream issue for the problem we were seeing: https://github.com/openSUSE/zypper/issues/457

mrc0mmand:
> > So, do you have reliable confirmation that the CI issues are gone since the zypper->dnf switch? (I.e. it was a client-side bug. Still, I'd be glad for any more details that could be passed to zypper folks.)
> 
> I haven't seen a single opensuse CI failure due to opensuse mirror issues since the switch so I'm very confident its fixed with dnf

To be fair, it happened with dnf as well:
 - https://github.com/systemd/systemd/actions/runs/5138767017/jobs/9248440256

But it definitely seems to happen _way_ less than with zypper, i.e.:
 - https://github.com/systemd/systemd/actions/runs/4946441216/jobs/8844516461
 - https://github.com/systemd/systemd/actions/runs/4945848416/jobs/8843132503
 - https://github.com/systemd/systemd/actions/runs/4942833313/jobs/8836727418
 - https://github.com/systemd/systemd/actions/runs/4939146392/jobs/8829630438
 - https://github.com/systemd/systemd/actions/runs/4938237923/jobs/8827726098
 - https://github.com/systemd/systemd/actions/runs/4929311604/jobs/8808766890
 - https://github.com/systemd/systemd/actions/runs/4926195198/jobs/8801405877

In other words - I found only a single fail with dnf in a ~3 week window, whereas the 8 fails above happened during a 3-day window.

keszybz:
I think we should close this. If we merged this now, we'd need to either disable tests on opensuse or deal with CI flaky to an unacceptable level. Both options are worse than status quo. Once https://github.com/openSUSE/zypper/issues/457 is fixed, we can consider reverting back to zypper, but even then I think this would require some discussion.

bluca:
isn't this saying there's a fix? https://github.com/openSUSE/zypper/issues/457#issuecomment-1587630590

bluca:
Also, I think we should let distro maintainers decide what tool is used for their distro. If I told you you had to use apt + alien to build Fedora images, I don't imagine that would make you too happy to work on this project, no?

Vogtinator:
> isn't this saying there's a fix? [openSUSE/zypper#457 (comment)](https://github.com/openSUSE/zypper/issues/457#issuecomment-1587630590)

For some instances of `curl error 92`, yes. The list above also has `curl error 16` and a 500 internal server error thrown in.

Werkov:
So this must wait until libzypp 17.31.13 reaches Ubuntu 22.04 (the version used in Github CI), if less frequent failures due to downloads are desired.

@bluca thanks!

bluca:
> > isn't this saying there's a fix? [openSUSE/zypper#457 (comment)](https://github.com/openSUSE/zypper/issues/457#issuecomment-1587630590)
> 
> For some instances of `curl error 92`, yes. The list above also has `curl error 16` and a 500 internal server error thrown in.

Are there other tickets/issue tracking the other errors?

bluca:
> So this must wait until libzypp 17.31.13 reaches Ubuntu 22.04 (the version used in Github CI), if less frequent failures due to downloads are desired.

Wholesale upgrades are very unlikely, however if you can pin point the exact patch or patches needed, I can arrange for backports?

Vogtinator:
> So this must wait until libzypp 17.31.13 reaches Ubuntu 22.04 (the version used in Github CI), if less frequent failures due to downloads are desired.

This particular issue also disappears if the `/etc/products.d/baseproduct` file contains the proper `<register><target>openSUSE-Tumbleweed-x86_64</target>...`. I'm not sure what it uses in the mkosi context though.

I doubt that this particular fix addresses most of the failures observed in the CI though, it's probably more networking or download infra issues. (In my experience, GH actions frequently have network issues, also on Ubuntu. Downloads and uploads fail randomly.)

bluca:
> > So this must wait until libzypp 17.31.13 reaches Ubuntu 22.04 (the version used in Github CI), if less frequent failures due to downloads are desired.
> 
> This particular issue also disappears if the `/etc/products.d/baseproduct` file contains the proper `<register><target>openSUSE-Tumbleweed-x86_64</target>...`. I'm not sure what it uses in the mkosi context though.

Is that in the guest image being built, or on the host?

Werkov:
> I doubt that this particular fix addresses most of the failures observed in the CI though, it's probably more networking or download infra issues. 

I see that openSUSE QA is hit similarly and related reports are:
- https://github.com/openSUSE/zypper/issues/399
- https://github.com/openSUSE/zypper/issues/420

and frequent workaround are retries.

> Wholesale upgrades are very unlikely, however if you can pin point the exact patch or patches needed, I can arrange for backports?

So the relevant libzypp updates are:

- https://github.com/openSUSE/libzypp/pull/461 -- handle curl error 92 specifically
- https://github.com/openSUSE/libzypp/pull/433 -- retry downloads on failure

Those would not handle the errors 500 or repomd.xml updates, hopefully, should made download more robust anyway.

bluca:
> > I doubt that this particular fix addresses most of the failures observed in the CI though, it's probably more networking or download infra issues.
> 
> I see that openSUSE QA is hit similarly and related reports are:
> 
> * [Curl error 16 (CURLE_HTTP2) openSUSE/zypper#399](https://github.com/openSUSE/zypper/issues/399)
> * [Automatic retry with max retries and timeout in both interactive and non-interactive mode (to work around mirror problems) openSUSE/zypper#420](https://github.com/openSUSE/zypper/issues/420)
> 
> and frequent workaround are retries.

Is there a TL;DR: of things that can be done in the CI integration to solve the issue? Or is it still pending?

> > Wholesale upgrades are very unlikely, however if you can pin point the exact patch or patches needed, I can arrange for backports?
> 
> So the relevant libzypp updates are:
> 
> * [curl: Trim user agent string (bsc#1212187) openSUSE/libzypp#461](https://github.com/openSUSE/libzypp/pull/461) -- handle curl error 92 specifically
> * [Download: retry internally on certain errors (like e.g. 92-CURLE_HTTP… openSUSE/libzypp#433](https://github.com/openSUSE/libzypp/pull/433) -- retry downloads on failure
> 
> Those would not handle the errors 500 or repomd.xml updates, hopefully, should made download more robust anyway.

Thanks I'll try and get that SRU'ed

Werkov:
> Is there a TL;DR: of things that can be done in the CI integration to solve the issue?

It seems people wrapped their invocation into retry loops (even before zypper [had it](https://github.com/openSUSE/zypper/issues/399#issuecomment-1401793349)).

Also trying to get more info about the server side...

Vogtinator:
> Is there a TL;DR: of things that can be done in the CI integration to solve the issue?

As a workaround it might work to use the repos over plain `http://`, as that effectively disables HTTP/2. The GPG signatures should be checked anyway.

bluca:
> > Is there a TL;DR: of things that can be done in the CI integration to solve the issue?
> 
> As a workaround it might work to use the repos over plain `http://`, as that effectively disables HTTP/2. The GPG signatures should be checked anyway.

ah that's not a bad idea, it's all throw-away CI stuff anyway  - @Werkov can you implement that in this PR? Then we can start it over and over to see if it holds

Werkov:
[LOL](https://github.com/systemd/mkosi/pull/71#issuecomment-286887858), HTTP for the third time.

(I keeping the Draft mode before I try rebase out.)

DaanDeMeyer:
Given all these issues, I'm even more inclined to just use dnf as that actually seems to work

bluca:
> [LOL](https://github.com/systemd/mkosi/pull/71#issuecomment-286887858), HTTP for the third time.
> 
> (I keep the Draft mode before I try rebase out.)

Is this the same old issue, or is it a different one?

```
Retrieving package crypto-policies-20230420.3d08ae7-1.4.noarch (1/116),  62.4 KiB (104.5 KiB unpacked)
Media source 'http://download.opensuse.org/tumbleweed/repo/oss/' does not contain the desired medium
History:
 - Location 'http://download.opensuse.org/tumbleweed/repo/oss/media.1/media' is temporarily unaccessible.
 ```

**Comment by bluca** [comment]
> Given all these issues, I'm even more inclined to just use dnf as that actually seems to work

If Suse users wanted to use dnf, they'd use Fedora or CentOS. This makes mkosi useless for them. It's fine until we work out CI issues, but it needs to be fixed before the next release

**Comment by DaanDeMeyer** [comment]
> If Suse users wanted to use dnf, they'd use Fedora or CentOS. This makes mkosi useless for them. It's fine until we work out CI issues, but it needs to be fixed before the next release

It's nothing more than a frontend for using rpm. I would hope distro users look at more than just the frontend for their actual package manager when deciding on what to use. You get pretty much the same image regardless of whether you use dnf or zypper. You can even install zypper and use it in the image even if it was built with dnf. I really don't see the issue with us using dnf in mkosi, it's nothing more than an implementation detail that requires you to install 1 extra package + deps which is already packaged in suse and elsewhere.

**Comment by bluca** [comment]
> > If Suse users wanted to use dnf, they'd use Fedora or CentOS. This makes mkosi useless for them. It's fine until we work out CI issues, but it needs to be fixed before the next release
> 
> It's nothing more than a frontend for using rpm. I would hope distro users look at more than just the frontend for their actual package manager when deciding on what to use. You get pretty much the same image regardless of whether you use dnf or zypper. You can even install zypper and use it in the image even if it was built with dnf. I really don't see the issue with us using dnf in mkosi, it's nothing more than an implementation detail that requires you to install 1 extra package + deps which is already packaged in suse and elsewhere.

All the support, documentation, tutorial, experience etc is different, so it does make a huge difference, it's not a hidden detail. As already said, imagine I came in and told you, for Fedora images we'll start using alien and apt, no more dnf. That wouldn't be so nice, no?

**Comment by Werkov** [comment]
> Given all these issues, I'm even more inclined to just use dnf as that actually seems to work

IIUC, you're giving the highest weight to the Github CI scenario (where download errors can be easily spotted and in hopeful future reduced (as discussed above)). Or are there other issues besides those?

Nevertheless, host pkg mgr == guest pkg mgr is presumably the most common practical usage combo.

It tests unnatural combo -- while unifying pkg managers used by mkosi in general is expected to reduce maintenance, this specificity can as well raise it higher again (not sure how to quantize either of the sides).

zypper vs dnf is not symmetric when the host system uses one of them (as you wrote mkosi requires more deps).

**Comment by keszybz** [comment]
> Also, I think we should let distro maintainers decide what tool is used for their distro. If I told you you had to use apt + alien to build Fedora images, I don't imagine that would make you too happy to work on this project, no?

Maybe that's just me, but I wouldn't particularly mind. If the other tool worked better, then I wouldn't be opposed. This particular example is weak, because packages converted with alien to rpm generally don't work at all. (It's been a few years since I last tried though.)

I wouldn't treat what the distro says as gospel. We switched to `dnf5` while it is still in fairly early phases in Fedora, and not supported at all in RHEL or CentOS. Using your argument, RHEL folks would be expected to complain that we're using `dnf5` for install their distro, which isn't even available for them.

More widely, `mkosi` is about installing distros in quite a different way then they're usually installed. E.g. we're not calling Anaconda or using kickstart or doing any of the other things that Fedora always does. We're also not installing grub2, or using the expected layout, etc. So I think the argument to let "distro maintainers decide" carries only so much weight.

**Comment by DaanDeMeyer** [comment]
> DNF doesn't seem able to build opensuse images anymore as of yesterday, it keeps failing in all PRs: https://github.com/systemd/systemd/actions/runs/5377624758/jobs/9756336340?pr=28159

Not dnf related, with this PR applied mkosi fails to build opensuse images as well

**Comment by bluca** [comment]
@Werkov one for you then, could you please check

**Comment by bluca** [comment]
> > DNF doesn't seem able to build opensuse images anymore as of yesterday, it keeps failing in all PRs: https://github.com/systemd/systemd/actions/runs/5377624758/jobs/9756336340?pr=28159
> 
> Not dnf related, with this PR applied mkosi fails to build opensuse images as well

Uhm I just checked and this PR does fix the issue. There is some ordering/bootstrapping issue that zypper can handle and dnf cannot. Hence I've un-drafted this and I'll merge it tomorrow morning to fix the build issue that has been happening for the past week.

**Comment by DaanDeMeyer** [comment]
> Uhm I just checked and this PR does fix the issue. There is some ordering/bootstrapping issue that zypper can handle and dnf cannot. Hence I've un-drafted this and I'll merge it tomorrow morning to fix the build issue that has been happening for the past week.

It most definitely does fail with zypper on my machine. Let's not be hasty and merge this PR to supposedly fix what might be an unrelated issue. A fix for this was recently merged in opensuse and this could be as simple as a mirror sync issue.

```
The following package failed to install:
  system-group-hardware
‣ "zypper --root /home/daandemeyer/projects/mkosi/.mkosi.tmpxfg1sb9e/root '--cache-dir=/home/daandemeyer/projects/mkosi/mkosi.cache/opensuse~tumbleweed' --reposd-dir=/home/daandemeyer/projects/mkosi/.mkosi.tmpxfg1sb9e/zypp.repos.d --gpg-auto-import-keys --non-interactive install -y --download-in-advance --no-recommends rpm kernel-kvmsmall udev systemd" returned non-zero exit code 8.
```

DaanDeMeyer:
@Werkov Can you verify if the fix for the issue should be deployed everywhere already? I still get failures building locally with dnf, dnf5 and zypper.

Vogtinator:
> @Werkov Can you verify if the fix for the issue should be deployed everywhere already? I still get failures building locally with dnf, dnf5 and zypper.

The fix is in TW snapshot 20230629+, hopefully published later today.

bluca:
> > Uhm I just checked and this PR does fix the issue. There is some ordering/bootstrapping issue that zypper can handle and dnf cannot. Hence I've un-drafted this and I'll merge it tomorrow morning to fix the build issue that has been happening for the past week.
> 
> It most definitely does fail with zypper on my machine. Let's not be hasty and merge this PR to supposedly fix what might be an unrelated issue. A fix for this was recently merged in opensuse and this could be as simple as a mirror sync issue.
> 
> ```
> The following package failed to install:
>   system-group-hardware
> ‣ "zypper --root /home/daandemeyer/projects/mkosi/.mkosi.tmpxfg1sb9e/root '--cache-dir=/home/daandemeyer/projects/mkosi/mkosi.cache/opensuse~tumbleweed' --reposd-dir=/home/daandemeyer/projects/mkosi/.mkosi.tmpxfg1sb9e/zypp.repos.d --gpg-auto-import-keys --non-interactive install -y --download-in-advance --no-recommends rpm kernel-kvmsmall udev systemd" returned non-zero exit code 8.
> ```

You might have some stale caches or something, it works for me starting from a clean build (and reliably fail with dnf) and also you can see it in the CI here which is working fine but failing in other PRs

DaanDeMeyer:
> You might have some stale caches or something, it works for me starting from a clean build (and reliably fail with dnf) and also you can see it in the CI here which is working fine but failing in other PRs

Sure, but that might be something specific to CI, where we run whatever frankenstein version of rpm is shipped by Ubuntu. I cleared all caches and it's still failing. Let's wait until the TW snapshot is out as mentioned by @Vogtinator, since until that's out it's actually expected for this to fail since it's a circular dependency in the packaging.

bluca:
> > You might have some stale caches or something, it works for me starting from a clean build (and reliably fail with dnf) and also you can see it in the CI here which is working fine but failing in other PRs
> 
> Sure, but that might be something specific to CI, where we run whatever frankenstein version of rpm is shipped by Ubuntu. I cleared all caches and it's still failing. Let's wait until the TW snapshot is out as mentioned by @Vogtinator, since until that's out it's actually expected for this to fail since it's a circular dependency in the packaging.

It's not specific to the CI, it's clearly working everywhere with zypper and failing everywhere with dnf. And it's not really surprising, since it's shoehorning a foreign tool that obviously the distro builders have no interest in using or maintaining, so this ends up being a combination that is tested by exactly nobody by us. It is all extremely silly and pointless. The original purpose of using a non-distro tool to build the distro was for a supposed CI fix, but actually it's been causing the CI to be broken for two weeks.

DaanDeMeyer:
If it's a dnf only issue, why is there a zypper log attachement in  https://bugzilla.suse.com/show_bug.cgi?id=1212674#c5 that shows the issue? I get that you want us to not use dnf for opensuse but using this to push it when it's blatantly a circular dependency in packaging and not a dnf specific bug feels kinda disingenuous to me.

And it's not working everywhere with zypper, it fails with zypper on a standard Fedora 38 install. It also fails with dnf on a standard Fedora 38 install. It doesn't fail with dnf5, but I can still see the same pre-install script errors when using dnf5.

DaanDeMeyer:
Also, as mentioned earlier, I'm fine with merging this PR if we support both zypper and dnf. If zypper is installed we can use it, if it's not installed we should use dnf.

DaanDeMeyer:
> The stated goal of using the foreign tool was to fix build issues, but now it is quite clearly causing them, so it needs to be reverted.

And it is still doing that, dnf5 is the only package manager with which I can currently build Opensuse images on Fedora. 

Given that I can reproduce the issue with zypper on Fedora, I'm pretty sure we should stop assigning blame to just the package manager as the issue is clearly more nuanced than that. I'm also very confident that once the new Tumbleweed snapshot with the packaging fix is out, both zypper and dnf will work fine again on both Ubuntu and Fedora. That will keep CI happy for now, which gives @Werkov time to rework this PR so we can support both zypper and dnf, with preference being given to zypper, after which both sides are happy.

bluca:
> > The stated goal of using the foreign tool was to fix build issues, but now it is quite clearly causing them, so it needs to be reverted.
> 
> And it is still doing that, dnf5 is the only package manager with which I can currently build Opensuse images on Fedora.
> 
> Given that I can reproduce the issue with zypper on Fedora, I'm pretty sure we should stop assigning blame to just the package manager as the issue is clearly more nuanced than that. I'm also very confident that once the new Tumbleweed snapshot with the packaging fix is out, both zypper and dnf will work fine again on both Ubuntu and Fedora. That will keep CI happy for now, which gives @Werkov time to rework this PR so we can support both zypper and dnf, with preference being given to zypper, after which both sides are happy.

Why can't that be done as a later follow-up? Current status: all CIs, all builds everywhere are broken. After this is merged, CIs are fixed, and builds everywhere but Fedora are fixed. Seems like an improvement to me?

DaanDeMeyer:
Because I can guarantee you that no one is going to add back dnf support for opensuse if we merge this as is now. The opensuse tumbleweed snapshot should be tagged soon, and having opensuse broken a little longer isn't going to hurt anyone.

bluca:
Ok, I've implemented the fallback to dnf now, so everyone should be happy and we can move on

bluca:
Also just noticed that the DNF implementation is broken w.r.t. offline builds, given it needs to fetch some keys from a remote URL, so it cannot work offline. As a fallback it's fine, but without zypper support as the first alternative for me it's a release blocker.

DaanDeMeyer:
I figured it'd be faster if I opened https://github.com/systemd/mkosi/pull/1651 instead of reviewing this one which makes zypper follow the latest changes in where to pick up config files and such

bluca:
Fixed by https://github.com/systemd/mkosi/pull/1651

PR #1607, opened 2023-06-02
Only raise exception if notify exit status is nonzero.

DaanDeMeyer:


DaanDeMeyer:
Trivial fix so merging

mkosi/qemu.py
```diff
@@ -305,5 +305,5 @@ def run_qemu(args: MkosiArgs, config: MkosiConfig) -> None:
 
         run(cmdline, stdin=sys.stdin, stdout=sys.stdout, env=os.environ, log=False)
 
-    if "EXIT_STATUS" in notifications:
-        raise subprocess.CalledProcessError(int(notifications["EXIT_STATUS"]), cmdline)
+    if "EXIT_STATUS" in notifications and (status := int(notifications["EXIT_STATUS"])) != 0:
```
behrmann, line 308:
```suggestion
    if status := int(notifications.get("EXIT_STATUS", 0)):
```
A bit shorter. The explicit `!=` is not necessary, because a numerical value different from zero will already trigger the if.

PR #1608, opened 2023-06-02
Fixes

DaanDeMeyer:


behrmann:
LGTM, only pyflakes is unhappy about an unused dependency.

PR #1613, opened 2023-06-06
Make matches work more like systemd conditions

DaanDeMeyer:
Let's make matches behave like systemd conditions. We drop support for list matches. Instead, we add support for match negation and trigger matches. A match is a trigger match if it's prefixed with the pipe symbol (|). A match is satisfied if all its regular matches and one of its trigger matches are satisfied.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -272,73 +272,57 @@ prefixed with `!`, if any later assignment of that setting tries to add
 the same value, that value is ignored. Values prefixed with `!` can be
 globs to ignore more than one value.
 
-To conditionally include configuration files, the `[Match]` section can
-be used. A configuration file is only included if all the conditions in the
-`[Match]` block are satisfied. If a condition in `[Match]` depends on a
-setting and the setting has not been explicitly configured when the condition
-is evaluated, the setting is assigned its default value.
+To conditionally include configuration files, the `[Match]` section can be used. Matches can use a pipe
+symbol ("|") after the equals sign ("…=|…"), which causes the match to become a triggering match. The config
+file will be included if the logical AND of all non-triggering matches and the logical OR of all triggering
+matches is satisfied. To negate the result of a match, prefix the argument with an exclamation mark. If an
+argument is prefixed with the pipe symbol and an exclamation mark, the pipe symbol must be passed first, and
+the exclamation second.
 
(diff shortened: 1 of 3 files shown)


mkosi.md
```diff
@@ -272,73 +272,59 @@ prefixed with `!`, if any later assignment of that setting tries to add
...
-Command line options that take no argument are shown without "=" in
-their long version. In the config files, they should be specified with
-a boolean argument: either "1", "yes", or "true" to enable, or "0",
-"no", "false" to disable.
+To conditionally include configuration files, the `[Match]` section can be used. If multiple matches are
+specified, the config will be included if all of them apply (i.e. a logical AND is applied). Matches can use
+a pipe symbol ("|") after the equals sign ("…=|…"), which causes the match to become a triggering match. If
+at least one triggering match is defined for a config file, then the config file will be included if at least
+one of the triggering matches of the config file applies and all of the regular (i.e. non-triggering) matches
+apply. To negate the result of a match, prefix the argument with an exclamation mark. If you prefix an
```
behrmann, line 280: Let's drive this home harder by writing something along the lines that "a `[Match]` secion is included if the logical AND of all non-triggering AND the logical OR of all triggering matches is satisfied".

mkosi.md
```diff
@@ -272,73 +272,59 @@ prefixed with `!`, if any later assignment of that setting tries to add
...
-  according to the UAPI group version format specification. If no operator is
-  prepended, the equality operator is assumed by default If this condition is
-  used and no image Version has be explicitly configured yet, this condition
-  fails. Multiple image version constraints can be specified as a
-  space-separated list. If multiple image version constraints are specified, all
-  must be satisfied for the match to succeed.
+: Matches against the configured image version. Image versions can be prepended by the operators `==`, `!=`,
+  `>=`, `<=`, `<`, `>` for rich version comparisons according to the UAPI group version format specification.
+  If no operator is prepended, the equality operator is assumed by default If this condition is used and no
+  image Version has be explicitly configured yet, this condition fails.
```
behrmann, line 314:
```suggestion
  If no operator is prepended, the equality operator is assumed by default. If this condition is used and no
  image version has been explicitly configured yet, this condition fails.
```

Issue #1603, opened 2023-06-01
Make package manager configuration more generic


DaanDeMeyer:
There's stuff like package pinning and tons of other configuration that's very package manager specific and wouldn't make sense for mkosi to implement for each package manager. Instead, we can allow users to specify and "etc" directory for the package manager which we treat as /etc for the package manager specifically. Users can put whatever package manager config files in there that they want and we'll make sure those files and directories are taken into account by the package manager.

behrmann:
As discussed previously this overlaps skeletons quite strongly (though not fully, since e.g. postinst scripts in Debian may react to files already in a root), so a solution I had in mind would be to copy the skeletons both into the image and into a separate dir in the workspace and resolve all config relative to that root.

DaanDeMeyer:
The big difference is that skeleton trees are copied into the image, and this directory would not be.

behrmann:
That's true, but is that necessary, though? I see two scenarios
1. You wanted an image where the package manager can run after you deploy it, then you will probably want to use the same config (repos, allowed keys, package pins) for the package manager in the image as when generating it
2. You don't want to run the package manager in the image, but then you don't really care whether these files will be copied into the image, because you either don't care they are there or you clean them up already.

gdiscry:
Like @behrmann, I only see those two scenarios.

Personally, I mainly build Debian images matching scenario 1 (but I am interested in building Arch images matching scenario 2). With the previous releases of mkosi (14 and before), I used `mkosi.skeleton` to add the backports repository, tweak the package selection in `/etc/apt/preferences.d` and seed a debconf configuration in `/var/cache/debconf`. I also disabled CleanPackageMetadata.

With the coming version, I understand the shift towards running the package manager directly on the host. I was glad to see debootstrap go as it does not support a skeleton filesystem, which is a pain with debconf. However, moving the configuration outside the image meant that configuring APT does not work anymore, hence the PRs I sent some time ago to start fixing that, starting with adding custom repositories.

So here are my 2 cents:

* I get that configuring the package manager is complicated because mkosi provides parts of the configuration and the user might want to tweak the rest, but `mkosi.skeleton` fits that role as long as mkosi can inject/override what it needs.
* There might also be some differences between the configuration used when building the image and the configuration inside the final image (like the use of a local mirror for custom repositories), which can be covered by `mkosi.extra`.
[...]

PR #1614, opened 2023-06-07
Seed package manager config

behrmann:
This is an idea how to address #1603. As argued there, I think skeletons are a sensible solution for this, since there is little point in having skeletons except to preseed config for the package manager, therefore I also don't think it's necessary to have a separate config for this, because either people will not want to update the image (and not care about having the same package manager config in the image as to generate it or they will already clean it up in some postinst) or they will want the same config in the image as to generate it, if they want to be able to update the image.

What this PR does, is to move the package manager config to generate the image into a separate subdirectory below the workspace and moves the files in there into their places where they should be by default, i.e. make it a normal filesystem tree. 

It still needs a few tweaks so that the package manager config can be extended with config from the skeletons (e.g. reading trusted keys for Debian), but it should allow us to remove some config of our own by punting it down to the skeletons (e.g. `repo_dirs`, which is not yet fully removed because I haven't thought about how to best change the pacman config for Arch).

For now this is a discussion piece, whether this is the way to go.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -26,8 +26,7 @@
     kernel images is trivial and fast these days.
 - Support for --qemu-boot was dropped
 - Support for --use-host-repositories was dropped, use --repository-directory instead
-- `RepositoryDirectory` was renamed to `RepositoryDirectories` and now takes a comma-separated
-  list of directories to look for extra repository files.
+- `RepositoryDirectory` was removed, use `PackageManagerTrees=` or `SkeletonTrees=` instead.
 - `--repositories` is now only usable on Debian/RPM based distros and can only be used to enable additional
   repositories. Specifically, it cannot be used on Arch Linux anymore to add new repositories.
 - The `_epel` distributions were removed. Use `--repositories=epel` instead to enable
--- a/mkosi.md
+++ b/mkosi.md
@@ -699,6 +699,19 @@ they should be specified with a boolean argument: either "1", "yes", or "true" t
   file may be provided too. `mkosi.skeleton.tar` will be automatically
   used if found in the local directory.
 
+`PackageManagerTrees=`, `--package-manager-tree=`
+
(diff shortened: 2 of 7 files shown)


DaanDeMeyer:
I would still prefer to decouple copying the package manager configuration files into the image from this. What about using the directories from --package-manager-tree (or w.e. it is called) if it is provided and falling back to the skeleton trees otherwise? And then if the relevant directory/tree exists in the package manager directory or skeleton tree, we use that and if it doesn't, we fall back to our builtin default?

behrmann:
> --package-manager-tree  

Not sure which option, you mean. The closest match would we `RepositoryDirectories=`, but that has a much narrower scope and I think it's actually something we can remove with this approach.

> I would still prefer to decouple copying the package manager configuration files into the image from this.

I guess we need to answer the question what skeleton trees are for. What I describe above are my use cases, so maybe I'm a bit myopic when it comes to this, but my thinking is that skeleton trees are there to preseed the tree before the package manager acts on it. In my use cases there two reasons to do this:
- configure the package manager (which stopped working when the package manager invocations started to happen from the outside and had some edge cases before then, too).
- preempt postinst scriptlets from packages from doing things (if that's supported in the first place)

This PR should reenable the first use case.

In both cases I want to have those files from the skeleton tree in the final image as well, i.e. if we had another option for this, it would in my use case have the same value as the skeleton tree, but if those files shouldn't end up in the final, they can be cleaned up in finalize. Alternatively we can still add a new option after the fact and default it to same value as the skeleton tree and give people an opt out of this behaviour once we see it's necessary.

behrmann:
Okay... While removing the `--repo-dir` option I found a valid use case: the initrd shouldn't contain the skeletons, but use the config. I'll rework it as mentioned above.

gdiscry:
I posted my general opinion in #1603, but FYI I like what you are doing here.

PR #1615, opened 2023-06-07
Fixes

DaanDeMeyer:


behrmann:
LGTM. On first glance I think the tar ambiguity should be resolved such that  we let tar cd to the target before unpacking. This could even work with the behaviour from this PR, if the cd fails, the target should be a copy.

PR #1618, opened 2023-06-07
Seed package manager config

behrmann:
Same as #1614, which wouldn't update anymore on pushes to the branch and couldn't be reopened, because it was force pushed.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -523,6 +523,24 @@ def install_skeleton_trees(state: MkosiState) -> None:
                 shutil.unpack_archive(source, t)
 
 
+def install_package_manager_trees(state: MkosiState) -> None:
+    if not state.config.package_manager_trees:
+        return
+
+    with complete_step("Copying in package maneger file trees…"):
+        for source, target in state.config.package_manager_trees:
+            t = state.workspace / "pkgmngr"
+            if target:
+                t = state.workspace / "pkgmngr" / target.relative_to("/")
+
+            t.parent.mkdir(mode=0o755, parents=True, exist_ok=True)
+
+            if source.is_dir() or target:
+                copy_path(source, t, preserve_owner=False)
+            else:
+                shutil.unpack_archive(source, t)
+
+
 def install_extra_trees(state: MkosiState) -> None:
     if not state.config.extra_trees:
         return
@@ -815,6 +833,7 @@ def install_unified_kernel(state: MkosiState, roothash: Optional[str]) -> None:
                 *(["--mirror", state.config.mirror] if state.config.mirror else []),
                 "--repository-key-check", yes_no(state.config.repository_key_check),
                 "--repositories", ",".join(state.config.repositories),
+                "--package-manager-tree", ",".join(format_source_target(s, t) for s, t in state.config.package_manager_trees),
                 *(["--compress-output", str(state.config.compress_output)] if state.config.compress_output else []),
                 "--with-network", yes_no(state.config.with_network),
                 "--cache-only", yes_no(state.config.cache_only),
@@ -1246,11 +1265,15 @@ def line_join_list(
     return "\n                                ".join(items)
 
 
+def format_source_target(source: Path, target: Optional[Path]) -> str:
+    return f"{source}:{target}" if target else f"{source}"
+
+
 def line_join_source_target_list(array: Sequence[tuple[Path, Optional[Path]]]) -> str:
     if not array:
         return "none"
 
-    items = [f"{source}:{target}" if target else f"{source}" for source, target in array]
+    items = [format_source_target(source, target) for source, target in array]
     return "\n                                ".join(items)
 
 
@@ -1299,6 +1322,7 @@ def summary(args: MkosiArgs, config: MkosiConfig) -> str:
                       Packages: {line_join_list(config.packages)}
             With Documentation: {yes_no(config.with_docs)}
                 Skeleton Trees: {line_join_source_target_list(config.skeleton_trees)}
+         Package Manager Trees: {line_join_source_target_list(config.package_manager_trees)}
                    Extra Trees: {line_join_source_target_list(config.extra_trees)}
         Clean Package Metadata: {yes_no_auto(config.clean_package_metadata)}
                   Remove Files: {line_join_list(config.remove_files)}
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -815,7 +815,6 @@ def install_unified_kernel(state: MkosiState, roothash: Optional[str]) -> None:
                 *(["--mirror", state.config.mirror] if state.config.mirror else []),
                 "--repository-key-check", yes_no(state.config.repository_key_check),
                 "--repositories", ",".join(state.config.repositories),
-                "--repo-dir", ",".join(str(p) for p in state.config.repo_dirs),
                 *(["--compress-output", str(state.config.compress_output)] if state.config.compress_output else []),
                 "--with-network", yes_no(state.config.with_network),
                 "--cache-only", yes_no(state.config.cache_only),
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -572,7 +572,6 @@ class MkosiConfig:
     local_mirror: Optional[str]
     repository_key_check: bool
     repositories: list[str]
-    repo_dirs: list[Path]
     repart_dirs: list[Path]
     overlay: bool
     architecture: Architecture
@@ -776,13 +775,6 @@ class MkosiConfigParser:
             dest="repositories",
             section="Distribution",
             parse=config_make_list_parser(delimiter=","),
-        ),
-        MkosiConfigSetting(
-            dest="repo_dirs",
-            name="RepositoryDirectories",
-            section="Distribution",
-            parse=config_make_list_parser(delimiter=",", parse=make_path_parser()),
-            paths=("mkosi.reposdir",),
         ),
         MkosiConfigSetting(
             dest="output_format",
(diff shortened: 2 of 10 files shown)


mkosi/distributions/debian.py
```diff
@@ -159,19 +159,20 @@ def architecture(arch: Architecture) -> str:
 
 
 def setup_apt(state: MkosiState, repos: Sequence[str]) -> None:
-    state.workspace.joinpath("apt").mkdir(exist_ok=True)
-    state.workspace.joinpath("apt/apt.conf.d").mkdir(exist_ok=True)
-    state.workspace.joinpath("apt/preferences.d").mkdir(exist_ok=True)
-    state.workspace.joinpath("apt/sources.list.d").mkdir(exist_ok=True)
-    state.workspace.joinpath("apt/log").mkdir(exist_ok=True)
+    state.workspace.joinpath("pkgmngr/etc/apt").mkdir(exist_ok=True, parents=True)
```
DaanDeMeyer, line 162: Let's add a helper method to `MkosiState` to access this directory like we have a helper for the staging directory..

mkosi/config.py
```diff
@@ -657,6 +658,12 @@ class MkosiConfig:
 
     preset: Optional[str]
 
+    def __post_init__(self) -> None:
+        if not self.package_manager_trees:
+            # this ugly construction is necessary to circumvent the freeze, we only care about
+            # being frozen after post init
+            object.__setattr__(self, "package_manager_trees", self.skeleton_trees)
```
DaanDeMeyer, line 664: Please do this using MkosiConfigSetting.default_factory

DaanDeMeyer (changes requested):
We should go further and only write files in pkgmngr if they don't already exists, even for files such as dnf.conf, so that users can override all of this if they want to.

behrmann:
> We should go further and only write files in pkgmngr if they don't already exists, even for files such as dnf.conf, so that users can override all of this if they want to.

Addressed the comments, but I'd like to leave this for a later follow up. This will need a bit more plumbing in the form of templating, since those configs will probably need to know the value of e.g. the workspace.

mkosi/state.py
```diff
@@ -61,6 +62,10 @@ def workdir(self) -> Path:
     def staging(self) -> Path:
         return self.workspace / "staging"
 
+    @property
+    def pkgmngr(self) -> Path:
+        return self.workspace / "pkgmngr"
+
```
DaanDeMeyer, line 68: You introduced this in the wrong commit, this should be done as the first commit

DaanDeMeyer (changes requested):

DaanDeMeyer:
> Addressed the comments, but I'd like to leave this for a later follow up. This will need a bit more plumbing in the form of templating, since those configs will probably need to know the value of e.g. the workspace.

So at least dnf and apt support setting options via the command line. What if we move all the options that users aren't allowed to change to be set via the command line, and this would include all the options that use the workspace (and the architecture and such)? pacman also supports overriding stuff via the CLI it seems

mkosi/__init__.py
```diff
@@ -815,6 +833,7 @@ def install_unified_kernel(state: MkosiState, roothash: Optional[str]) -> None:
                 *(["--mirror", state.config.mirror] if state.config.mirror else []),
                 "--repository-key-check", yes_no(state.config.repository_key_check),
                 "--repositories", ",".join(state.config.repositories),
+                *flatten([("--package-manager-tree", format_source_target(s, t)) for s, t in state.config.package_manager_trees]),
```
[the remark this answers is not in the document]
behrmann: This would be more similar to the rest, but I think
```python
*(["--package-manager-tree", ",".join(format_source_target(s, t) for s, t in state.config.package_manager_trees] if state.config.package_manager_trees else [])
```
is less readable than the `flatten`.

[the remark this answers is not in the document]
DaanDeMeyer: You should be able to get away with 

```python
"--package-manager-tree", ",".join(...)
```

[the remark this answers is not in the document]
behrmann: Exchanged it for that.

mkosi/distributions/arch.py
```diff
@@ -82,8 +86,16 @@ def install_packages(cls, state: MkosiState, packages: Sequence[str], apivfs: bo
                     )
                 )
 
-            for d in state.config.repo_dirs:
-                f.write(f"Include = {d}/*\n")
+            confsnippetsdir = state.pkgmngr / "etc/pacman.d/conf/"
+            if any(confsnippetsdir.glob("*")):
```
DaanDeMeyer, line 90: iterdir() instead of glob()?

mkosi/distributions/debian.py
```diff
@@ -245,10 +246,3 @@ def install_apt_sources(state: MkosiState, repos: Sequence[str]) -> None:
         with sources.open("w") as f:
             for repo in repos:
                 f.write(f"{repo}\n")
-
-    # Already contains a merged tree of repo_dirs after setup_apt
-    for src in state.workspace.joinpath("apt/sources.list.d").iterdir():
-        dst = state.root.joinpath("etc/apt/sources.list.d", src.name)
-        if dst.exists():
-            continue
-        shutil.copyfile(src, dst)
```
DaanDeMeyer, line 254: Can we just drop this? This is writing to the root, which we don't do anywhere else?

DaanDeMeyer (changes requested):
Almost there. We can either do the CLI stuff in this PR or another one

DaanDeMeyer (changes requested):
Almost there. We can either do the CLI stuff in this PR or another one

mkosi/distributions/arch.py
behrmann: Sure, made it a bit more robust, too.

mkosi/distributions/debian.py
behrmann: Thought about this already and wasn't entirely sure, but it shouldn't be a problem and the file can be easily passed in via some tree. Added it as an extra commit in case it needs reverting.

mkosi/distributions/debian.py
```diff
@@ -124,7 +124,6 @@ def install_packages(cls, state: MkosiState, packages: Sequence[str], apivfs: bo
         setup_apt(state, cls.repositories(state))
         invoke_apt(state, "update", apivfs=False)
         invoke_apt(state, "install", packages, apivfs=apivfs)
-        install_apt_sources(state, cls.repositories(state, local=False))
```
DaanDeMeyer, line 127: The thing is, if we drop this, the user has to write the file himself, which is kind of a bad user experience unlike the other distros where you can just install a package, so I'd keep this intact for now

DaanDeMeyer (changes requested):

mkosi/distributions/debian.py
behrmann: Dropped the commit again.

mkosi/distributions/debian.py
```diff
@@ -216,7 +217,7 @@ def setup_apt(state: MkosiState, repos: Sequence[str]) -> None:
             if not src.is_file():
                 continue
             if src.suffix in (".list", ".sources"):
-                shutil.copyfile(src, state.workspace.joinpath("apt/sources.list.d", src.name))
+                shutil.copyfile(src, state.workspace.joinpath("pkgmngr/etc/apt/sources.list.d", src.name))
```
DaanDeMeyer: `state.pkgmngr`

mkosi/distributions/fedora.py
```diff
@@ -172,10 +174,9 @@ def invoke_dnf(
         "--setopt=keepcache=1",
         "--setopt=install_weak_deps=0",
         f"--setopt=cachedir={state.cache_dir}",
-        f"--setopt=reposdir={' '.join(str(p) for p in state.config.repo_dirs)}",
```
DaanDeMeyer, line 175: We drop this but never set it again to the path in the pkgmngr tree?

DaanDeMeyer (changes requested):
Found a few more things

mkosi/distributions/arch.py
```diff
@@ -84,6 +84,18 @@ def install_packages(cls, state: MkosiState, packages: Sequence[str], apivfs: bo
...
+            # confsnippetsdir does not exist.
+            if any(f.is_file() for f in confsnippetsdir.glob("*")):
+                f.write(
+                    dedent(
+                        f"""\
+
+                        Include = {state.pkgmngr}/etc/pacman.d/conf/*
+                        """
+                    )
+                )
```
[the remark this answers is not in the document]
behrmann: That doesn't work, pacman complains about no such file existing.

mkosi/distributions/fedora.py
behrmann: That must have gotten lost during a rebase.

mkosi/distributions/debian.py
```diff
@@ -247,7 +248,7 @@ def install_apt_sources(state: MkosiState, repos: Sequence[str]) -> None:
                 f.write(f"{repo}\n")
 
     # Already contains a merged tree of repo_dirs after setup_apt
-    for src in state.workspace.joinpath("apt/sources.list.d").iterdir():
+    for src in state.workspace.joinpath("pkgmngr/etc/apt/sources.list.d").iterdir():
```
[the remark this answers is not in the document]
behrmann: I don't see it and GitHub showed the hunk as outdated from the start.

mkosi/distributions/debian.py
behrmann: same

mkosi/distributions/debian.py
```diff
@@ -207,7 +208,7 @@ def setup_apt(state: MkosiState, repos: Sequence[str]) -> None:
         )
     )
 
-    with state.workspace.joinpath("apt/sources.list").open("w") as f:
+    with state.workspace.joinpath("pkgmngr/etc/apt/sources.list").open("w") as f:
```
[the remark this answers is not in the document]
behrmann: same

DaanDeMeyer (approved):

PR #1620, opened 2023-06-08
Prefer configuration via CLI over configuration via config file

DaanDeMeyer:
This makes it easier to debug things with --debug as all the options used will be logged instead of having to go and figure out what's in the config file.


behrmann:
Looks sensible to me, but having the refactors in Arch in separate commits would be nice.

PR #1621, opened 2023-06-08
debian: fix default keyring

behrmann:
The key file RELEASE-archive-keyring does not exist, but debian-archive-keyring.gpg contains all currently used keys.


DaanDeMeyer:
What about ubuntu though? Maybe I meant to uze distro here originally?

DaanDeMeyer (approved):

PR #1624, opened 2023-06-11
Set Dir::Etc from config file for apt

### LINKED ISSUE #1623: Mkosi also runs dpkg hooks from the host system
Hi,

I noticed a regression from https://github.com/systemd/mkosi/commit/a7d5ee6ed08c4fb356b19b683ab759fb816b440c.

On the host where I use mkosi, I use a dpkg `Post-Invoke` command to run bootctl update. That command needs root privileges of course.

With a7d5ee6ed08c4fb356b19b683ab759fb816b440c this hook is also executed when building an image with mkosi. As an unprivileged user this fails. As a privileged user, I probably don't want it to be run.
I isolated it to this commit, the commit before (3b57cb5b4fd9c0c3ffa1fb17c536343f76ddb554) works fine, a7d5ee6ed08c4fb356b19b683ab759fb816b440c itself, not.

As an easy way to try this out, put this into `/etc/apt/apt.conf.d/90test`
```perl
DPkg::Post-Invoke {
    "/usr/bin/false";
};
```
and then run `mkosi`.

I also noticed that two options, I set inside `/etc/apt/apt.conf.d`, namely 
```
Dpkg::Options {
   "--force-confdef";
   "--force-confold";
}
```
are not included anymore in the dpkg commandline.

With the working version, the dpkg invocation is
```
/usr/bin/dpkg --root=/var/lib/nfsrootserver/nfsroot/.mkosi.tmpwdncnjva/root --log=/var/lib/nfsrootserver/nfsroot/.mkosi.tmpwdncnjva/pkgmngr/var/log/apt/dpkg.log --force-unsafe-io --force-architecture --force-depends --force-confdef --force-confold --root=/var/lib/nfsrootserver/nfsroot/.mkosi.tmpwdncnjva/root --log=/var/lib/nfsrootserver/nfsroot/.mkosi.tmpwdncnjva/pkgmngr/var/log/apt/dpkg.log --force-unsafe-io --force-architecture --force-depends --status-fd 20 --configure --pending
```
(The double specifications of `--root`, `--log`, etc. are really there and not some copying error)

After the change to cmdline options it is
```
/usr/bin/dpkg --root=/var/lib/nfsrootserver/nfsroot/.mkosi.tmp4pirwdy1/root --log=/var/lib/nfsrootserver/nfsroot/.mkosi.tmp4pirwdy1/pkgmngr/var/log/apt/dpkg.log --force-unsafe-io --force-architecture --force-depends --status-fd 15 --configure --pending
```

Cornelicorn:
Specifying it as an argument to apt-get doesn't work. from apt.conf(5):

Dir::Etc contains the location of configuration files, sourcelist gives the location of the sourcelist and main is the default configuration file (setting has no effect, unless it is done from the config file specified by the "APT_CONFIG" environment variable).

Closes #1623.

### DIFF
--- a/mkosi/distributions/debian.py
+++ b/mkosi/distributions/debian.py
@@ -177,8 +177,9 @@ def setup_apt(state: MkosiState, repos: Sequence[str]) -> None:
     # Anything that users can override with dropins is written into the config file.
     config.write_text(
         dedent(
-            """\
+            f"""\
             APT::Install-Recommends "false";
+            Dir::Etc "{state.pkgmngr / "etc/apt"}";
             """
         )
     )
@@ -223,7 +224,6 @@ def invoke_apt(
         "-o", f"Dir::Cache={state.cache_dir}",
         "-o", f"Dir::State={state.pkgmngr / 'var/lib/apt'}",
         "-o", f"Dir::State::status={state.root / 'var/lib/dpkg/status'}",
-        "-o", f"Dir::Etc={state.pkgmngr / 'etc/apt'}",
         "-o", f"Dir::Etc::trusted={trustedkeys}",
         "-o", f"Dir::Etc::trustedparts={trustedkeys_dir}",
         "-o", f"Dir::Log={state.pkgmngr / 'var/log/apt'}",


DaanDeMeyer:
Can you give https://github.com/systemd/mkosi/pull/1625 a try? That should also solve the issue

behrmann:
Superseded by #1625

PR #1625, opened 2023-06-11
Fixes

DaanDeMeyer:


Cornelicorn:
At least for me c40db8572768a57d562338afcbf45ac1290fab9b doesn't do it.
Judging from the [source](https://salsa.debian.org/apt-team/apt/-/blob/main/apt-pkg/init.cc#L227), I think it should be possible to pass apt a config with the env variable `APT_CONFIG` and in that file only set `Dir::Etc`, [which should load `apt.conf` from that tree](https://salsa.debian.org/apt-team/apt/-/blob/main/apt-pkg/init.cc#L246), so there would be no risk that this file is getting overwritten, if `APT_CONFIG` is just written to somewhere outside of `state.pkgmngr`.

DaanDeMeyer:
> At least for me https://github.com/systemd/mkosi/commit/c40db8572768a57d562338afcbf45ac1290fab9b doesn't do it.
Judging from the [source](https://salsa.debian.org/apt-team/apt/-/blob/main/apt-pkg/init.cc#L227), I think it should be possible to pass apt a config with the env variable APT_CONFIG and in that file only set Dir::Etc, [which should load apt.conf from that tree](https://salsa.debian.org/apt-team/apt/-/blob/main/apt-pkg/init.cc#L246), so there would be no risk that this file is getting overwritten, if APT_CONFIG is just written to somewhere outside of state.pkgmngr.

Implemented exactly this, please give it another try.

Cornelicorn:
> > At least for me [c40db85](https://github.com/systemd/mkosi/commit/c40db8572768a57d562338afcbf45ac1290fab9b) doesn't do it.
> > Judging from the [source](https://salsa.debian.org/apt-team/apt/-/blob/main/apt-pkg/init.cc#L227), I think it should be possible to pass apt a config with the env variable APT_CONFIG and in that file only set Dir::Etc, [which should load apt.conf from that tree](https://salsa.debian.org/apt-team/apt/-/blob/main/apt-pkg/init.cc#L246), so there would be no risk that this file is getting overwritten, if APT_CONFIG is just written to somewhere outside of state.pkgmngr.
> 
> Implemented exactly this, please give it another try.

Yes, now it works.

mkosi/distributions/debian.py
```diff
@@ -172,18 +172,32 @@ def setup_apt(state: MkosiState, repos: Sequence[str]) -> None:
     state.root.joinpath("var/lib/dpkg").mkdir(mode=0o755, exist_ok=True)
     state.root.joinpath("var/lib/dpkg/status").touch()
 
-    config = state.pkgmngr / "etc/apt/apt.conf"
+    # We have a special apt.conf outside of pkgmngr dir that only configures "Dir" that we pass to
+    # APT_CONFIG to tell apt it should read config files in pkgmngr instead of in its usual locations. This
+    # is required because apt parses CLI configuration options after parsing its configuration files and as
+    # such we can't use CLI options to tell apt where to look for configuration files.
+    config = state.workspace / "apt.conf"
+    if not config.exists():
```
Cornelicorn, line 180: Should this file not always be overwritten?

DaanDeMeyer: Why? If it exists, it'll have the content we want. Its wasteful to write the same thing over and over again. It's in the workspace so the user can't control the content.

Cornelicorn: True, nvm

behrmann (approved):

PR #1627, opened 2023-06-12
Ensure we return the same exit code in debug mode

DaanDeMeyer:
When running in debug mode, we shouldn't return a different exit code compared to when we run outside of debug mode.


behrmann:
No real objections from me, but could you add an example or comment to the commit message when the mentioned issue is triggered?

DaanDeMeyer:
Added an example to the commit message, only updated commit message and CI was green so merging

PR #1629, opened 2023-06-13
Remove several repository config options

behrmann:
This PR removes the options `LocalMirror=`, `RepositoryKeycheck=` and `Repositories=`. This PR is a draft, because I'm not sure about all changes, so see this as a discussion piece.

`LocalMirror=` and `RepositoryKeycheck=` are removed since they were workarounds to configure repositories, which can now be done via `PackageManagerTrees=` (which can use paths outside the image) and `SkeletonTrees=` in a natural way by passing a tree with config for the package manager, which should mostly allow to configure it without us needing to add any further options to mkosi.

`Repositories=` is removed because the semantics are different between distros. In Debian it sets the components in `sources.list` whereas for Fedora it adds config switches to enable repos. This is a slightly regressive change, since for Debian this now means that if people want more than `main`, they will have to add a `sources.list`. For DNF-based distros  I assume this is a transparent change, since the written repo files already have an `enabled=1` line by default. 

/cc @bluca since your OBS usage might be most in need of change if this PR would go in.

### DIFF
--- a/mkosi/distributions/debian.py
+++ b/mkosi/distributions/debian.py
@@ -24,11 +24,8 @@ class DebianInstaller(DistributionInstaller):
         return Path(f"boot/vmlinuz-{name}")
 
     @staticmethod
-    def repositories(state: MkosiState, local: bool = True) -> list[str]:
+    def repositories(state: MkosiState) -> list[str]:
         repos = ' '.join(("main", *state.config.repositories))
-
-        if state.config.local_mirror and local:
-            return [f"deb [trusted=yes] {state.config.local_mirror} {state.config.release} {repos}"]
 
         main = f"deb {state.config.mirror} {state.config.release} {repos}"
 
@@ -124,7 +121,7 @@ class DebianInstaller(DistributionInstaller):
         setup_apt(state, cls.repositories(state))
         invoke_apt(state, "update", apivfs=False)
         invoke_apt(state, "install", packages, apivfs=apivfs)
-        install_apt_sources(state, cls.repositories(state, local=False))
+        install_apt_sources(state, cls.repositories(state))
 
         policyrcd.unlink()
 
--- a/mkosi.md
+++ b/mkosi.md
@@ -337,16 +337,6 @@ they should be specified with a boolean argument: either "1", "yes", or "true" t
 : The mirror to use for downloading the distribution packages. Expects
   a mirror URL as argument.
 
-`LocalMirror=`, `--local-mirror=`
-
-: The mirror will be used as a local, plain and direct mirror instead
-  of using it as a prefix for the full set of repositories normally supported
-  by distributions. Useful for fully offline builds with a single repository.
-  Supported on deb/rpm/arch based distributions. Overrides `--mirror=` but only
-  for the local mkosi build, it will not be configured inside the final image,
-  `--mirror=` (or the default repository) will be configured inside the final
-  image instead.
-
 `RepositoryKeyCheck=`, `--repository-key-check=`
 
 : Controls signature/key checks when using repositories, enabled by default.
--- a/mkosi.md
+++ b/mkosi.md
@@ -337,12 +337,6 @@ they should be specified with a boolean argument: either "1", "yes", or "true" t
 : The mirror to use for downloading the distribution packages. Expects
   a mirror URL as argument.
 
-`RepositoryKeyCheck=`, `--repository-key-check=`
-
-: Controls signature/key checks when using repositories, enabled by default.
-  Useful to disable checks when combined with `--local-mirror=` and using only
-  a repository from a local filesystem. Not used for DNF-based distros yet.
-
 `Repositories=`, `--repositories=`
 
 : Additional package repositories to use during installation. Expects
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -832,7 +832,6 @@ def install_unified_kernel(state: MkosiState, roothash: Optional[str]) -> None:
                 "--release", state.config.release,
                 "--architecture", str(state.config.architecture),
                 *(["--mirror", state.config.mirror] if state.config.mirror else []),
-                "--repository-key-check", yes_no(state.config.repository_key_check),
                 "--repositories", ",".join(state.config.repositories),
                 "--package-manager-tree", ",".join(format_source_target(s, t) for s, t in state.config.package_manager_trees),
                 *(["--compress-output", str(state.config.compress_output)] if state.config.compress_output else []),
@@ -1299,7 +1298,6 @@ def summary(args: MkosiArgs, config: MkosiConfig) -> str:
                        Release: {bold(none_to_na(config.release))}
                   Architecture: {config.architecture}
                         Mirror: {none_to_default(config.mirror)}
-      Repo Signature/Key check: {yes_no(config.repository_key_check)}
                   Repositories: {",".join(config.repositories)}
 
     {bold("OUTPUT")}:
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1299,7 +1299,6 @@ def summary(args: MkosiArgs, config: MkosiConfig) -> str:
                        Release: {bold(none_to_na(config.release))}
                   Architecture: {config.architecture}
                         Mirror: {none_to_default(config.mirror)}
-          Local Mirror (build): {none_to_none(config.local_mirror)}
       Repo Signature/Key check: {yes_no(config.repository_key_check)}
                   Repositories: {",".join(config.repositories)}
 
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -585,7 +585,6 @@ class MkosiConfig:
     distribution: Distribution
     release: str
     mirror: Optional[str]
-    repository_key_check: bool
     repositories: list[str]
     repart_dirs: list[Path]
     overlay: bool
@@ -783,15 +782,6 @@ class MkosiConfigParser:
             section="Distribution",
             default_factory=config_default_mirror,
             help="Distribution mirror to use",
-        ),
-        MkosiConfigSetting(
-            dest="repository_key_check",
-            metavar="BOOL",
-            nargs="?",
-            section="Distribution",
-            default=True,
-            parse=config_parse_boolean,
-            help="Controls signature and key checks on repositories",
         ),
         MkosiConfigSetting(
             dest="repositories",
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -585,7 +585,6 @@ class MkosiConfig:
     distribution: Distribution
     release: str
     mirror: Optional[str]
-    local_mirror: Optional[str]
     repository_key_check: bool
     repositories: list[str]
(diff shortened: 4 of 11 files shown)


bluca:
> `LocalMirror=` and `RepositoryKeycheck=` are removed since they were workarounds to configure repositories, which can now be done via `PackageManagerTrees=` (which can use paths outside the image) and `SkeletonTrees=` in a natural way by passing a tree with config for the package manager, which should mostly allow to configure it without us needing to add any further options to mkosi.

Does this mean I need to know in advance how package managers work for that key skip option? Or is it something that mkosi ships?

DaanDeMeyer:
I'm not sure about removing Repositories=. It's quite useful for the EPEL stuff where we can ship the repos in mkosi but have them disabled by default. Is it conceptually really that different from the debian components? They both seem to act like repos to me. We could even implement it for arch to allow people to disable extra (although I think that would break horribly if there's still stuff from core that depends on stuff from extra)

behrmann:
> I'm not sure about removing Repositories=. It's quite useful for the EPEL stuff where we can ship the repos in mkosi but have them disabled by default. Is it conceptually really that different from the debian components? They both seem to act like repos to me.

I actually worked in that direction earlier but decided against it. I'm not quite sure what the design goals there were, but repos and components seem somewhat orthogonal to me - you get a matrix of repo component combinations. I remember people wanting to disable repos, but right now that would only be possible by overwriting `sources.list`.  It's probably a bit more common to want to include more components, so using `Repositories=` for that is not totally unreasonable, but I overlooked the EPEL use case, so I'll just drop that commit. 

> (although I think that would break horribly if there's still stuff from core that depends on stuff from extra)

I guess that's the case. I added back the `extra` repo to debug a build failure for Arch and it went away.

behrmann:
>> LocalMirror= and RepositoryKeycheck= are removed since they were workarounds to configure repositories, which can now be done via PackageManagerTrees= (which can use paths outside the image) and SkeletonTrees= in a natural way by passing a tree with config for the package manager, which should mostly allow to configure it without us needing to add any further options to mkosi.
> 
> Does this mean I need to know in advance how package managers work for that key skip option? Or is it something that mkosi ships?

Yes, this would mean that some knowledge about the used package manager would be necessary. Instead of setting the config options, you would drop a `pacman.conf`
```
[options]
SigLevel = Never

[core]
Server = file:///path/to/your/repo
```
in a directory and reference it as `PackageManagerTree=` or do the same thing with a `dnf.conf` or `apt.conf.d` snippet and a `sources.list`.

We recently added `PackageManagerTree=` to cater to all package manager related things that were not already covered, but I wanted to talk about the ones addressed here, because they have overlap with that and I assume we wouldn't add at least some them today, if we had had the other option earlier, and since we are already breaking compat up and down the line and promised to not do that anymore after the next release - now would be the time to remove them, hence this discussion piece.

I agree that it's slightly more cumbersome to not have mkosi do this and it requires more knowledge of the used package manager, but these are somewhat niche things to do anyway and they would also be ignored when `PackageManagerTrees=` are used for any other reason (if the package manager can't do drop in config).

bluca:
> > > LocalMirror= and RepositoryKeycheck= are removed since they were workarounds to configure repositories, which can now be done via PackageManagerTrees= (which can use paths outside the image) and SkeletonTrees= in a natural way by passing a tree with config for the package manager, which should mostly allow to configure it without us needing to add any further options to mkosi.
> > 
> > 
> > Does this mean I need to know in advance how package managers work for that key skip option? Or is it something that mkosi ships?
> 
> Yes, this would mean that some knowledge about the used package manager would be necessary. Instead of setting the config options, you would drop a `pacman.conf`
> 
> ```
> [options]
> SigLevel = Never
> 
> [core]
> Server = file:///path/to/your/repo
> ```
> 
> in a directory and reference it as `PackageManagerTree=` or do the same thing with a `dnf.conf` or `apt.conf.d` snippet and a `sources.list`.
> 
> We recently added `PackageManagerTree=` to cater to all package manager related things that were not already covered, but I wanted to talk about the ones addressed here, because they have overlap with that and I assume we wouldn't add at least some them today, if we had had the other option earlier, and since we are already breaking compat up and down the line and promised to not do that anymore after the next release - now would be the time to remove them, hence this discussion piece.
> 
> I agree that it's slightly more cumbersome to not have mkosi do this and it requires more knowledge of the used package manager, but these are somewhat niche things to do anyway and they would also be ignored when `PackageManagerTrees=` are used for any other reason (if the package manager can't do drop in config).

It's a bit of a problem because the build runner is the same regardless of what distro is built or what version, and it is _slow_ to update - we are talking deployments a couple of times a year, so having to constantly update the glue script to adapt to changing distributions and package managers is not realistic

behrmann:
> It's a bit of a problem because the build runner is the same regardless of what distro is built or what version, and it is slow to update - we are talking deployments a couple of times a year, so having to constantly update the glue script to adapt to changing distributions and package managers is not realistic

I see, but shouldn't this still be doable? For one, I don't think we'll be seeing a lot of churn in terms of distros or package managers. In terms of distros we already have the majority of distros. I'd assume mkosi's distro coverage could cater at least 90% of all Linux usage and most of the rest would be totally out of scope anyway. In terms of package managers: most package manager options seem rather stable to me. The odd one out, I don't know too much about is DNF were I see changing names with version 5, but to me at least it seemed to only affect the CLI options. So in all, I'd wager that once written these config files would be rather stable - maybe more stable than what we've done in the past.

On the other hand I think that you I imaged that this change would be mostly a movement from mkosi options, i.e. mkosi writing config files, to external config files that could still live in the same tree. For example this might look like this: You make a package manager tree for each distro, with the needed config files to include a single repo and not check its keys. Then you would include those trees with one config file snippet per distro that has a match section on that distro. This way the config for the runner would be the static and it would still only be parametrized by the distro you are building. Does that make sense?

bluca:
> > It's a bit of a problem because the build runner is the same regardless of what distro is built or what version, and it is slow to update - we are talking deployments a couple of times a year, so having to constantly update the glue script to adapt to changing distributions and package managers is not realistic
> 
> I see, but shouldn't this still be doable? For one, I don't think we'll be seeing a lot of churn in terms of distros or package managers. In terms of distros we already have the majority of distros. I'd assume mkosi's distro coverage could cater at least 90% of all Linux usage and most of the rest would be totally out of scope anyway. In terms of package managers: most package manager options seem rather stable to me. The odd one out, I don't know too much about is DNF were I see changing names with version 5, but to me at least it seemed to only affect the CLI options. So in all, I'd wager that once written these config files would be rather stable - maybe more stable than what we've done in the past.
> 
> On the other hand I think that you I imaged that this change would be mostly a movement from mkosi options, i.e. mkosi writing config files, to external config files that could still live in the same tree. For example this might look like this: You make a package manager tree for each distro, with the needed config files to include a single repo and not check its keys. Then you would include those trees with one config file snippet per distro that has a match section on that distro. This way the config for the runner would be the static and it would still only be parametrized by the distro you are building. Does that make sense?

That would necessarily have to live in the upstream mkosi tree. If you want to ship all such configurations pre-prepared for all distributions, that's fine.

Having an OBS-specific package just for mkosi is not viable.

behrmann:
I think I'm not understanding how the OBS setup looks like. What does the input to mkosi there look like?

bluca:
> I think I'm not understanding how the OBS setup looks like. What does the input to mkosi there look like?

There's the user-customized part (how the image will look like) and there's the build system specific parts (disable key checks, use local mirrors, etc)

behrmann:
Is there a repo to see what that concretely looks like? My question is how the build system specific parts are passed in / where they come from, because what I described above would be, at least in my mind, exchanging one build system specific part for another, and I'd like to understand how to make that fit, if it's possible.

bluca:
(this is not updated to the breaking changes since last version)

https://build.opensuse.org/package/view_file/home:bluca:mkosi/debian-image/mkosi.debian?expand=1

that's the config file provided by the user

https://github.com/openSUSE/obs-build/blob/master/build-recipe-mkosi

and that's the glue script that creates the local mirrors and calls mkosi

behrmann:
> (this is not updated to the breaking changes since last version)
> 
> https://build.opensuse.org/package/view_file/home:bluca:mkosi/debian-image/mkosi.debian?expand=1
> 
> that's the config file provided by the user
> 
> https://github.com/openSUSE/obs-build/blob/master/build-recipe-mkosi
> 
> and that's the glue script that creates the local mirrors and calls mkosi

Thanks! This is really helpful!

What I would be proposing would be to add a single directory tree in `recipe_setup_mkosi` that contains an `sources.list`, a `pacman.conf` and a `dnf.conf`, with the single repository configuration and the directive to not check keys and pass `--package-manager-tree` to that directory instead of `--repository-key-check=no --local-mirror file:///.build.binaries/` in `recipe_build_mkosi`. Since the package managers don't interfere with one another and only one will be used by mkosi during image build, reading stuff from the package manager tree, and those files don't end up in the image, this wouldn't even need any fancy matching. Would that be possible?

bluca:
You mean adding those configuration in the mkosi package? As long as that's stored here and picked up automatically that's fine, but I'm not sure if that's better than having command line switches

behrmann:
I mean having them be part of the OBS recipe, since that recipe needs to be touched in case of incompatible changes in mkosi anyway. At least that's my understanding from
> (this is not updated to the breaking changes since last version)

i.e. when updating the recipe for the next mkosi release the three files for apt, dnf and pacman would need to be added. They would then stay constant until these package managers would change their config syntax (which I don't expect to happen all that much) in which case the recipe might need touching up anyway, because a new mkosi release would be needed.

bluca:
That means having to change OBS's code for every package manager used, with per-package manager configuration, just because of internal mkosi quirks. I don't think that's feasible.

DaanDeMeyer:
> That means having to change OBS's code for every package manager used, with per-package manager configuration, just because of internal mkosi quirks. I don't think that's feasible.

It's not like new package managers pop up left and right you know. Instead of specifying --local-mirror and --repository-key-check, you'd now write the config files. We can add documentation to mkosi on how to do it  but we wouldn't do it in mkosi itself as it requires two options for something that's trivial to do now outside of mkosi.

mkosi/distributions/debian.py
```diff
@@ -124,7 +112,6 @@ def install_packages(cls, state: MkosiState, packages: Sequence[str], apivfs: bo
         setup_apt(state, cls.repositories(state))
         invoke_apt(state, "update", apivfs=False)
         invoke_apt(state, "install", packages, apivfs=apivfs)
-        install_apt_sources(state, cls.repositories(state, local=False))
```
DaanDeMeyer, line 127: So I still don't think we've made this redundant since there's no easy way for debian users to get the default mkosi settings into the image. We can still discuss this, but let's do that in a separate PR

behrmann:
Rebased and factored out the unrelated Debian stuff into #1648.

DaanDeMeyer:
Let's keep these options. `--repository-key-check` is useful on its own. `--local-mirror` is a bit up in the air but we might as well keep it, it's not that painful to support it.

PR #1630, opened 2023-06-14
Run repart before booting image in systemd-nspawn

DaanDeMeyer:


mkosi/qemu.py
```diff
@@ -191,8 +194,20 @@ async def notify() -> None:
         logging.debug(f"Received {num_messages} notify messages totalling {format_bytes(num_bytes)} bytes")
 
 
+@contextlib.contextmanager
+def copy_ephemeral(config: MkosiConfig, src: Path) -> Iterator[Path]:
+    src = src.resolve()
+    tmp = Path(src.parent / f"{src.name}-{uuid.uuid4().hex}")
```
behrmann, line 200: This doesn't use `tempfile` because you only need a temporary name instead of a directory for the btrfs snapshot? If so, a comment would probably make sense.

behrmann (approved):
LGTM, just one question

behrmann (approved):

DaanDeMeyer:
No need to wait for full CI since only a comment was modified

PR #1632, opened 2023-06-16
Support Ukify config

MariusSchiffer:
Checks for existence of /etc/kernel/uki.conf and gives this to ukify.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -1451,7 +1451,11 @@ the default kernel commandline embedded in the kernel
 and any previous `-append` specifications.
 
 `mkosi` builds a Unified Kernel Image (UKI).
-It is also copied into the output directory and may be booted directly:
+Further customization, e.g. a splash image, can be applied using a configuration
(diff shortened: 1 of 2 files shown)


behrmann:
I don't think we should add this specific option. There's sure going to be more of them, but if there's a config file in a standard location (that then can contain e.g. the splash option) we should include `--config /path/to/config/file`.

MariusSchiffer:
@behrmann Would that be read again on incremental builds? I'm a bit confused on what `state.pkgmngr` does.

MariusSchiffer:
Also I just noticed: Ukify is of course not run in the image itself. Without a specified working directory, it's kinda hard to specify paths in the UKI config file. Can I add a `cwd`?

MariusSchiffer:
Okay, works wonderfully. I added a little bit of documentation. Example section seemed to fit the best.

behrmann:
Thanks for the contribution, @MariusSchiffer!

PR #1633, opened 2023-06-19
Add kernel version to UKI name for versioned images

### LINKED ISSUE #1631: Support multiple kernel versions with image version
When setting an image version, the same name is used for the UKI filename for multiple kernel versions.

See here: https://github.com/systemd/mkosi/blob/5616f28505a00dfe58793fadfa81919446024596/mkosi/__init__.py#L880-L881

MariusSchiffer:
Fixes #1631

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -878,7 +878,7 @@ def install_unified_kernel(state: MkosiState, roothash: Optional[str]) -> None:
                 boot_count = f'+{state.root.joinpath("etc/kernel/tries").read_text().strip()}'
 
             if state.config.image_version:
-                boot_binary = state.root / f"efi/EFI/Linux/{image_id}_{state.config.image_version}{boot_count}.efi"
+                boot_binary = state.root / f"efi/EFI/Linux/{image_id}_{state.config.image_version}-{kver}{boot_count}.efi"
             elif roothash:
                 _, _, h = roothash.partition("=")
                 boot_binary = state.root / f"efi/EFI/Linux/{image_id}-{kver}-{h}{boot_count}.efi"


DaanDeMeyer (approved):

mkosi/__init__.py
```diff
@@ -878,7 +878,7 @@ def install_unified_kernel(state: MkosiState, roothash: Optional[str]) -> None:
                 boot_count = f'+{state.root.joinpath("etc/kernel/tries").read_text().strip()}'
 
             if state.config.image_version:
-                boot_binary = state.root / f"efi/EFI/Linux/{image_id}_{state.config.image_version}{boot_count}.efi"
+                boot_binary = state.root / f"efi/EFI/Linux/{image_id}_{state.config.image_version}-{kver}{boot_count}.efi"
```
behrmann, line 881: Looking at this, have we been doing this wrong for a long time? Shouldn't this be
```suggestion
                boot_binary = state.root / f"efi/EFI/Linux/{image_id}_{state.config.image_version}-{kver}+{boot_count}.efi"
```
> systemd-boot implements a simple boot counting mechanism on top of the Boot Loader Specification[1], for automatic and unattended fallback to older kernel versions/boot loader entries when a specific entry continuously fails. Any boot loader entry file and unified kernel image file that contains a "+" followed by one or two numbers (if two they need to be separated by a "-"), before the .conf or .efi suffix is subject to boot counting:

MariusSchiffer: I think the `+` is already included in the `boot_count` variable.

behrmann: Ah, right, nevermind. Just saw it in the diff in the mail

PR #1634, opened 2023-06-20
Remove --install-directory= option

DaanDeMeyer:
We don't benefit from the caching anymore since we started emptying the directory completely on reuse as otherwise old leftover files might get installed. Without the caching, the option does not have a ton of use anymore, so let's remove it.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -83,6 +83,10 @@
   the time.
 - Disk images are now resized to 8G when booted to give some disk space to play around with in the booted
   image.
+- Removed `--install-directory=` option. This was originally added for caching the installation results, but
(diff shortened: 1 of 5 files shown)


NEWS.md
```diff
@@ -83,6 +83,10 @@
   the time.
 - Disk images are now resized to 8G when booted to give some disk space to play around with in the booted
   image.
+- Removed `--install-directory=` option. This was originally added for caching the installationr results, but
```
behrmann, line 86:
```suggestion
- Removed `--install-directory=` option. This was originally added for caching the installation results, but
```

behrmann (approved):

DaanDeMeyer:
Just a doc fix, so let's merge

PR #1635, opened 2023-06-20
Add support for building arbitrary output artifacts

DaanDeMeyer:


behrmann:
LGTM, the only thing I'd like to bikeshed is the name `none` for the output type, since it's both a bit confusing ("no output?") and can be confused with Python's `None`. How about something like `input` or `pipeline` or `intermediate`?

DaanDeMeyer:
I wanted to use "build" but it conflicts with one of the verb names which causes issues. Maybe "artifact"?

DaanDeMeyer:
Tbf "none" kinda fits snice we're not rpoducing an image

behrmann:
> Tbf "none" kinda fits snice we're not rpoducing an image

It does make some sense. I would prefer `artifact`, though, or `intermediate-artifact`. Or maybe `temporary-artifact` or `temporary` or any of the words we use to describe these things: ephemeral, volatile, runtime.

DaanDeMeyer:
"temporary" would be confusing, it's not temporary, it will never materialize in the output directory at all. That's why "none" seems like a better fit, it describes what appears in the output directory, which is nothing since the image is not put in the output directory.

behrmann:
I think I'd go with `intermediate` or `pipeline`, or if we want to stress the not-being-there-ness, maybe `no-file` or `no-output`, but I can live with `none`. What bugs me is just the closeness to `None`.

keszybz:
Code-wise, this is very simple and clear.



> I think I'd go with `intermediate` or `pipeline`, or if we want to stress the not-being-there-ness, maybe `no-file` or `no-output`, but I can live with `none`. What bugs me is just the closeness to `None`.

I don't think there's any chance of confusion. One is a string in user config, the other is a keyword in the program code. This is not perl with implicit conversions. I very much prefer 'none' over the proposed alternatives.

PR #1636, opened 2023-06-20
Generate argument parser from config setting list

keszybz:


keszybz:
CI / integration-test (ubuntu, directory) failed on some apt network error. I'll restart it.

mkosi/config.py
```diff
@@ -742,465 +751,681 @@ class MkosiConfigParser:
...
         ),
         MkosiConfigSetting(
             dest="root_password",
+            metavar="PASSWORD",
             section="Content",
             parse=config_parse_root_password,
             paths=("mkosi.rootpw",),
             path_read_text=True,
             path_secret=True,
+            help="Set the root's password",
```
behrmann, line 1213: the combination of article and possessive marker looks weird, I think either of these alternatives would work
```suggestion
            help="Set the root password",
```
or
```suggestion
            help="Set root's password",
```
or
```suggestion
            help="Set the root account's password",
```
or
```suggestion
            help="Set the root user's password",
```

My preference would be the first one.

mkosi/config.py
```diff
@@ -742,465 +751,681 @@ class MkosiConfigParser:
...
         ),
         MkosiConfigSetting(
             dest="root_shell",
+            metavar="SHELL",
             section="Content",
             parse=config_parse_string,
+            help="Set the root's shell",
```
behrmann, line 1220: Same, and also in the commit message.

mkosi/config.py
```diff
@@ -1422,545 +1647,27 @@ def create_argument_parser(self) -> argparse.ArgumentParser:
...
-        )
-        group.add_argument(
-            "--sign",
-            help="Write and sign SHA256SUMS file",
-            metavar="BOOL",
-            nargs="?",
-            action=action,
-        )
-        group.add_argument("--key", help="GPG key to use for signing", action=action)
+            group.add_argument(    # type: ignore
```
behrmann, line 1660: If you still have the warning in some log adding that to the commit message would be nice.

behrmann (commented):
Excellent change, some minor nits.

behrmann:
The CI failure is unrelated.

DaanDeMeyer (approved):

mkosi/config.py
keszybz: Hmm, the first one is how we'd usually write it. But "set zbyszek password" looks very strange, and really "root" is just the account name here. What about "set password for root"? I think that avoids the issue nicely.

keszybz:
Updated with "change the password|shell for root" (split out into a separate commit since with a whole paragraph of musings for the gramatically inclined), and the amended message in the last commit.

PR #1639, opened 2023-06-21
Run systemd-hwdb when building an image

DaanDeMeyer:


mkosi/__init__.py
```diff
@@ -1478,6 +1478,11 @@ def run_preset(state: MkosiState) -> None:
         run(["systemctl", "--root", state.root, "preset-all"])
 
 
+def run_hwdb(state: MkosiState) -> None:
+    with complete_step("Generating hardware database"):
+        run(["systemd-hwdb", "--root", state.root, "--usr", "--strict", "update"])
```
behrmann, line 1483: Is any `systemd-hwdb` fine or should we advise to build a recent one like we do for other tools?

behrmann (commented):
LGTM, one question

mkosi/__init__.py
keszybz: Any should be fine. I don't think it changed in living memory, except for support for `query --root`. But `update --root` worked always.

keszybz: `--strict` is from 2018, and `--usr` since the first version.

PR #1642, opened 2023-06-23
fedora: fix RemovePackages with dnf5

keszybz:


DaanDeMeyer (approved):

behrmann:
I reran CI because most of them were flukes, but on opensuse
```
Failed:
  dbus-1-common-1.14.8-1.2.noarch  system-group-hardware-20170617-24.13.noarch 
  systemd-253.5-3.1.x86_64    
```
consistently fail to install

DaanDeMeyer:
@Werkov Seems there's a dependency issue in some of the opensuse packages, dbus is getting installed before sysusers-shadow is installed.

Werkov:
When I check the downloaded package manually, it has the `pre` dependency:
```
rpm -qp -v --requires /tmp/dbus-1-common-1.14.8-1.2.noarch.rpm | grep pre
...
pre: sysuser-shadow >= 3.1
```
The order of RPM installation in dnf seems to ignore it while [zypper](https://github.com/systemd/mkosi/actions/runs/5378294902/jobs/9757977998?pr=1596#step:9:1028) installation seems to honor it (but I can't tell whether it's accidental luck).

DaanDeMeyer:
> The order of RPM installation in dnf seems to ignore it while [zypper](https://github.com/systemd/mkosi/actions/runs/5378294902/jobs/9757977998?pr=1596#step:9:1028) installation seems to honor it (but I can't tell whether it's accidental luck).

I get the same error locally on Fedora with your PR applied that uses zypper instead of dnf

Werkov:
This is likely a recent dependency cycle in [openSUSE](https://bugzilla.suse.com/show_bug.cgi?id=1212674#c5) -- once there's a cycle, there's no single transaction order and zypper may have more luck picking the right one than dnf. But principally, the cause are broken deps within distro -- i.e. not a bug in this PR.

behrmann:
Let's merge this, then, since this is orthogonal to the zypper issue.

PR #1646, opened 2023-06-27
Gentoo fixes

DaanDeMeyer:
- Use curl to download stage3 tarball so we get a progress bar
- Do not exclude dev, proc and sys directories when extracting tarball
  (only exclude their contents)
- Copy pkgmngr/ directory into stage3/ directory wholesale instead of
  individual files
- Various coding style fixes
- Stop using Repositories= to specify binary package repositories as it
  is not its intended purpose. Instead, pass configured environment
  variables to emerge so users can set PORTAGE_BINHOST instead.

### DIFF
--- a/mkosi/distributions/gentoo.py
+++ b/mkosi/distributions/gentoo.py
@@ -10,7 +10,7 @@ from typing import Mapping
 from mkosi.architecture import Architecture
 from mkosi.distributions import DistributionInstaller
 from mkosi.install import copy_path
-from mkosi.log import ARG_DEBUG, complete_step, die, log_step
+from mkosi.log import ARG_DEBUG, complete_step, die
 from mkosi.remove import unlink_try_hard
 from mkosi.run import run, run_workspace_command
 from mkosi.state import MkosiState
@@ -20,17 +20,16 @@ from mkosi.types import PathString
 def invoke_emerge(
     state: MkosiState,
     packages: Sequence[str] = (),
-    actions: Sequence[str] = (),
     options: Sequence[str] = (),
     env: Mapping[str, str] = {},
 ) -> None:
-    print(f"{' '.join(state.config.repositories)}")
     run_workspace_command(
-        state.cache_dir.joinpath("stage3"),
+        state.cache_dir / "stage3",
         cmd=[
             "emerge",
             *packages,
             "--update",
+            "--deep",
             "--buildpkg=y",
             "--usepkg=y",
             "--keep-going=y",
@@ -43,10 +42,10 @@ def invoke_emerge(
             "--verbose-conflicts",
             "--changed-use",
             "--newuse",
-            f"--root={Path('/tmp/mkosi-root')}",
+            "--root=/tmp/mkosi-root",
+            "--binpkg-respect-use",
             *(["--verbose", "--quiet=n", "--quiet-fail=n"] if ARG_DEBUG.get() else ["--quiet-build", "--quiet"]),
             *options,
-            *actions,
         ],
         bwrap_params=[
             "--bind", state.root, "/tmp/mkosi-root",
@@ -55,20 +54,18 @@ def invoke_emerge(
             "--bind", state.cache_dir / "repos", "/var/db/repos",
         ],
         network=True,
-        env={
-            'PORTAGE_BINHOST': ' '.join(state.config.repositories),
-            'FEATURES': ' '.join([
+        env=dict(
+            FEATURES=" ".join([
                 "getbinpkg",
                 "-candy",
-                'parallel-install',
-                *(['noman', 'nodoc', 'noinfo'] if state.config.with_docs else []),
+                "parallel-install",
+                *(["noman", "nodoc", "noinfo"] if state.config.with_docs else []),
             ]),
-            # gnuefi: for systemd
+            # boot: for systemd
             # minimal: because we like minimals
             # initramfs, symlink for kernel
-            'USE': 'gnuefi initramfs minimal symlink',
-            **env,
-        },
+            USE="boot initramfs minimal symlink",
+        ) | env | state.environment,
     )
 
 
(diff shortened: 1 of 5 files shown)


DaanDeMeyer:
cc @257

mkosi/distributions/gentoo.py
```diff
@@ -104,78 +101,49 @@ def install(cls, state: MkosiState) -> None:
...
-        run_workspace_command(stage3_cache, ["/usr/bin/emerge-webrsync"],
-                              bwrap_params=bwrap_params, network=True)
-
-        with complete_step("Layingout basic filesystem"):
-            invoke_emerge(state, options=["--emptytree"],
-                          packages=["sys-apps/baselayout"],
-                          env={'USE': 'build'})
+            (state.cache_dir / d).mkdir(exist_ok=True)
+
+        copy_path(state.pkgmngr, stage3, preserve_owner=False)
```
257, line 137: image needs these configs, most importantly `make.profile`. we probably want `copy_path(config, root_portage_cfg)` back.

DaanDeMeyer: If I look inside /etc/portage in the stage3 directory in mkosi.cache, make.profile is a dangling symlink. The only actual file in /etc/portage is make.conf with the following contents:

```
# These settings were set by the catalyst build script that automatically
# built this stage.
# Please consult /usr/share/portage/config/make.conf.example for a more
# detailed example.
COMMON_FLAGS="-O2 -pipe"
CFLAGS="${COMMON_FLAGS}"
CXXFLAGS="${COMMON_FLAGS}"
FCFLAGS="${COMMON_FLAGS}"
FFLAGS="${COMMON_FLAGS}"

# NOTE: This stage was built with the bindist Use flag enabled

# This sets the language of build output to English.
# Please keep this setting intact when reporting bugs.
LC_MESSAGES=C.utf8
```

I don't think users get much value from us copying this into the image. Anyone that wants an updateable gentoo image will install portage in the image and provide custom configuration files in mkosi.skeleton/ which will be included in the image so there should be no need for us to copy these to the image.

mkosi/distributions/gentoo.py
```diff
@@ -35,40 +33,39 @@ def invoke_emerge(
...
             *(["--verbose", "--quiet=n", "--quiet-fail=n"] if ARG_DEBUG.get() else ["--quiet-build", "--quiet"]),
             *options,
-            *actions,
         ],
         bwrap_params=[
             "--bind", state.root, "/tmp/mkosi-root",
             "--bind", state.cache_dir / "binpkgs", "/var/cache/binpkgs",
             "--bind", state.cache_dir / "distfiles", "/var/cache/distfiles",
-            "--bind", state.cache_dir / "repos", "/var/db/repos",
+            "--bind", state.cache_dir / "repos", "/var/db/repos/gentoo",
```
257, line 54: this would disable [overlays](https://overlays.gentoo.org/). i would keep the mnt-point at `/var/db/repo`

257:
> * Stop using Repositories= to specify binary package repositories as it
>   is not its intended purpose. Instead, pass configured environment
>   variables to emerge so users can set PORTAGE_BINHOST instead.

what is the intended purpose of `Repositories=`? could `overlays` be passed here?

mkosi/distributions/gentoo.py
```diff
@@ -35,40 +33,39 @@ def invoke_emerge(
             "--usepkg=y",
             "--keep-going=y",
             "--jobs",
-            "--load-average",
+            # Make sure the system stays somewhat usable by putting the max load average at 90%.
+            f"--load-average={0.9 * len(os.sched_getaffinity(0))}",
```
257, line 37: how would this interact with [`portage`'s logic for `--jobs` and `--load-average`](https://github.com/gentoo/portage/blob/master/lib/portage/util/cpuinfo.py)?

mkosi/distributions/gentoo.py
DaanDeMeyer: > then they will have a drift no? they have to keep a copy of the config in mkosi.skeleton and also in pkgmng.
also i suppose you mean mkosi.extra because then we're mixing stuff in skeleton and pkgmng, it will be a hell to maintain for users.

By default, unless --package-manager-tree is specified explicitly, mkosi.skeleton/ is used. So there won't be any drift as the skeleton tree and the package manager tree will be the exact same tree, so the same configs that are used to build the image will be copied into the image as well with no drift.

DaanDeMeyer:
> what is the intended purpose of Repositories=? could overlays be passed here?

`Repositories=` is used to enable repositories that are disabled by default. The primary example is the EPEL and EPEL-NEXT repositories in CentOS. It is **not** intended to add new repositories, for that we have --package-manager-tree or `PORTAGE_BINHOST` in case of gentoo

mkosi/distributions/gentoo.py
```diff
@@ -104,78 +101,49 @@ def install(cls, state: MkosiState) -> None:
...
 
-        stage3_tar_path = state.cache_dir / stage3_tar
+        stage3_url = urllib.parse.urljoin(state.config.mirror, f"releases/{arch}/autobuilds/{stage3_latest}")
+        stage3_tar = state.cache_dir / "stage3.tar"
+        stage3 = state.cache_dir / "stage3"
 
-        stage3_url_path = urllib.parse.urljoin(
-            state.config.mirror, f"releases/{arch}/autobuilds/{stage3_tar}",
-        )
+        if not stage3_tar.exists():
```
257, line 113: how does this detect that a new stage3 released upstream?

mkosi/distributions/gentoo.py
DaanDeMeyer: Not sure what you mean here? This makes sure gentoo doesn't use more than 90% of the allocated CPUs to keep the system somewhat responsive even during a build

257: not sure, afk, but I think there is that logic baked into portage. this effects -j and -l in MAKEOPTS too not just emerge. I would leave it up to portage to handle this.
in my personal experience not passing anything here has only increased the load with 0 noticable effect on system's responsiveness.

DaanDeMeyer: But the documentation for --load-average mentions that providing an empty value removes any previous assignment, so by setting an empty value we pretty much unset the load average.

257: see the link I posted above to the code itself. and I wouldn't stop there, load_average is spread all over the code. documentation is most probably misleading. leaving it empty, from what I remember, tells portage to go parallel and if there is no value, find out number of cores etc via get_cpu_count() which actually calls getaffinity().

DaanDeMeyer: Dropped these changes

DaanDeMeyer:
I pulled 2 changes from https://github.com/systemd/mkosi/pull/1650 into this one

DaanDeMeyer:
CI failure is unrelated. @257 Can you bless this? (I'd prefer not to wait hours for CI again)

257:
it's good to go

257:
looking at the logs though, did you add systemd-9999? I guess it doesn't matter. I will fix it whatever it is in the  clang PR; _Tuesday_.

PR #1648, opened 2023-06-27
Remove stretch support

behrmann:


mkosi/distributions/debian.py
```diff
@@ -124,7 +115,6 @@ def install_packages(cls, state: MkosiState, packages: Sequence[str], apivfs: bo
         setup_apt(state, cls.repositories(state))
         invoke_apt(state, "update", apivfs=False)
         invoke_apt(state, "install", packages, apivfs=apivfs)
-        install_apt_sources(state, cls.repositories(state, local=False))
```
bluca, line 127: Please leave this, it is useful to have, and if someone wants to build a really minimal one they can remove the package manager and then it won't be added

behrmann: Dropped the commit from this PR and spin it out into a separate one after testing this again. I remember it not doing anything in the first place and parity between what we do for different distros would be nice, but this definitely needs more discussion.

DaanDeMeyer:
cc @gdiscry are you OK with maintaining the stretch support in a fork?

gdiscry:
> cc @gdiscry are you OK with maintaining the stretch support in a fork?

Yes, no problem, I can maintain a fork for my needs. And the fix is "documented" in the project history.

With a bit of luck, I won't need it for too long.

bluca (approved):

PR #1649, opened 2023-06-30
Remove setup.py and use setuptools build backend in pyproject.toml

behrmann:
Also remove setup.cfg, since we don't use flake8 and everything else can be handled by pyproject.toml

This is a rebase of #1343.

I removed the user and system installation test, since with PEP668 most systems will need `--break-system-packages` for that anyway and that's not something to advertise. Either people use their distro package or they'll install it in a venv or use the shim.


DaanDeMeyer (approved):

behrmann:
That's a cute error.  `setup.py` and `pyproject.toml` have nothing to do with resolving paths in the current working directory.

PR #1650, opened 2023-06-30
gentoo: swith to llvm (clang)

257:
This will pave the way for actual arch support and cross-compiling

### DIFF
--- a/mkosi/distributions/gentoo.py
+++ b/mkosi/distributions/gentoo.py
@@ -104,7 +104,7 @@ class GentooInstaller(DistributionInstaller):
 
         with urllib.request.urlopen(stage3tsf_path_url) as r:
             # e.g.: 20230108T161708Z/stage3-amd64-nomultilib-systemd-mergedusr-20230108T161708Z.tar.xz
-            regexp = rf"^[0-9]+T[0-9]+Z/stage3-{arch}-nomultilib-systemd-mergedusr-[0-9]+T[0-9]+Z\.tar\.xz"
+            regexp = rf"^[0-9]+T[0-9]+Z/stage3-{build_arch}-llvm-systemd-mergedusr-[0-9]+T[0-9]+Z\.tar\.xz"
             all_lines = r.readlines()
             for line in all_lines:
                 if (m := re.match(regexp, line.decode("utf-8"))):


mkosi/distributions/gentoo.py
```diff
@@ -147,6 +148,11 @@ def install(cls, state: MkosiState) -> None:
                     "--exclude", "./dev",
                     "--exclude", "./proc",
                 ])
+            # gnuefi on clang will trigger this: https://bugs.gentoo.org/887047
+            # REMOVE when https://github.com/systemd/systemd/pull/26641 is released
+            with (config / "package.accept_keywords/9999").open("a") as f:
+                f.write("=sys-apps/systemd-9999 **")
```
behrmann, line 154: Probably missing a newline?

257:
forgot to comment out `Repository=` in ci to actually test the builds with clang (binaries under my repo are built with clang though).

behrmann (commented):
How common is it to build Gentoo with LLVM instead of GCC and why can't GCC not do cross-compilation?

257:
> How common is it to build Gentoo with LLVM instead of GCC

no idea. there is no stats on that.

> and why can't GCC not do cross-compilation?

gcc:
https://wiki.gentoo.org/wiki/Cross_build_environment
https://wiki.gentoo.org/wiki/Embedded_Handbook/General/Creating_a_cross-compiler

clang, none of the above.

DaanDeMeyer:
I think you're going to have to update the binpkgs repo first, this is way too slow for CI now

257:
out dated gcc, updating...

257:
> How common is it to build Gentoo with LLVM instead of GCC and why can't GCC not do cross-compilation?

no so bad read on this: [Cross compiling made easy, using Clang and LLVM](https://mcilloni.ovh/2021/02/09/cxx-cross-clang/)

DaanDeMeyer:
Can you paste the error from source build run into somewhere regarding /sys? I'd prefer to fix the cause instead of merging workarounds if possible

257:
> Can you paste the error from source build run into somewhere regarding /sys? I'd prefer to fix the cause instead of merging workarounds if possible

https://github.com/systemd/mkosi/pull/1665

257:
> Can you get me the contents of config.log? That should tell us what specifically is failing. Also, is this with or without LLVM?


off your `gentoo-bwrap`: https://gist.github.com/257/6a174cd8fe2d29ab78d085963e2741d0

DaanDeMeyer:
Let's close this until glibc supports compilation with LLVM

DaanDeMeyer:
I can't reopen this one, can you recreate?

PR #1651, opened 2023-07-02
opensuse: Add back support for zypper

DaanDeMeyer:
Let's add back support for zypper to build opensuse images. If both zypper and dnf are installed, let's prefer using zypper.

### DIFF
--- a/mkosi/distributions/opensuse.py
+++ b/mkosi/distributions/opensuse.py
@@ -1,13 +1,16 @@
 # SPDX-License-Identifier: LGPL-2.1+
 
+import shutil
+import textwrap
 import urllib.request
 import xml.etree.ElementTree as ElementTree
 from collections.abc import Sequence
 
 from mkosi.architecture import Architecture
 from mkosi.distributions import DistributionInstaller
-from mkosi.distributions.fedora import Repo, invoke_dnf, setup_dnf
+from mkosi.distributions.fedora import Repo, fixup_rpmdb_location, invoke_dnf, setup_dnf
 from mkosi.log import die
+from mkosi.run import bwrap
 from mkosi.state import MkosiState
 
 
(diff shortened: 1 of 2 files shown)


mkosi/distributions/opensuse.py
```diff
@@ -45,12 +48,19 @@ def install_packages(cls, state: MkosiState, packages: Sequence[str], apivfs: bo
         if updates_url is not None:
             repos += [Repo("repo-update", f"baseurl={updates_url}", fetch_gpgurls(updates_url))]
```
bluca, line 49: this "fetch_gpgurls" needs to be also under the fallback path

bluca:
There's a few missing things from https://github.com/systemd/mkosi/pull/1596/files

- install zypper in CI
- avoid HTTPS
- stable repo

DaanDeMeyer:
- CI seems to be fine with https from the results in this PR, so I suggest we keep the https url for now.
- Using {release} instead of stable is something I would keep instead of hardcoding "stable". That was an improvement in the commit removing zypper that's unrelated to removing zypper itself.

mkosi/distributions/opensuse.py
DaanDeMeyer: Does it need to be? I figure it doesn't hurt to do this for opensuse as well. If any new keys pop up, zypper will download them as needed.

bluca:
> Using {release} instead of stable is something I would keep instead of hardcoding "stable". That was an improvement in the commit removing zypper that's unrelated to removing zypper itself.

The problem is that the branch also matches on 'current', no?

mkosi/distributions/opensuse.py
bluca: This breaks offline builds, and it is unnecessary for zypper, so please move it under the fallback branch

DaanDeMeyer:
> The problem is that the branch also matches on 'current', no?

I see both stable and current here: https://download.opensuse.org/distribution/

bluca:
ah no they are both aliases to the same, never mind

bluca (approved):

PR #1654, opened 2023-07-03
Drop support for rocky, alma, mageia and openmandriva

DaanDeMeyer:
These are basically unmaintained upstream and can be implemented using --package-manager-tree, so let's remove them from the repo.


bluca:
This makes it extremely annoying to use those distributions with mkosi, as complicated files have to be created and there's no default that 'just works'. Having sensible defaults is important and user-friendly.

DaanDeMeyer:
> This makes it extremely annoying to use those distributions with mkosi, as complicated files have to be created and there's no default that 'just works'. Having sensible defaults is important and user-friendly.

Sure, and whoever is interested in maintaining support for those distros can maintain those files in a separate repository. But given that in the last 2 years no one has stepped up to maintain these distros and I'm already way too occupied maintaining the other ones, we should drop them instead of pretending they're supported. You can still have a "just works" setup, you just have to get the necessary package manager files from somewhere else.

bluca:
What maintenance is required? What's missing in these?

mkosi/config.py
```diff
@@ -1897,9 +1893,6 @@ def load_config(args: argparse.Namespace) -> MkosiConfig:
     if args.cmdline and args.verb not in MKOSI_COMMANDS_CMDLINE:
         die(f"Parameters after verb are only accepted for {' '.join(verb.name for verb in MKOSI_COMMANDS_CMDLINE)}.")
 
-    if shutil.which("bsdtar") and args.distribution == Distribution.openmandriva and args.tar_strip_selinux_context:
```
bluca, line 1900: how can this be done from an external repository/config?

mkosi/util.py
```diff
@@ -155,14 +148,7 @@ def detect_distribution() -> tuple[Optional[Distribution], Optional[str]]:
 
 
 def is_dnf_distribution(d: Distribution) -> bool:
-    return d in (
```
bluca, line 158: and this too

DaanDeMeyer:
Alright given someone's shown interest in keeping these in mkosi, let's not drop them just yet

PR #1655, opened 2023-07-03
Use stat tool to check if we're on a btrfs subsystem

DaanDeMeyer:
Currently, we let the btrfs tool log errors when we're not on a btrfs filesystem. Let's avoid unnecessary errors by checking if we're on a btrfs filesystem before invoking the btrfs tool.


behrmann (approved):

bluca (approved):
thanks, warning is gone

PR #1659, opened 2023-07-03
Add backward-compat stubs for infrastructure options

bluca:
These options are used by the infrastructure service, rather than image-specific configuration, so add stubs to avoid breaking compatibility. This lets OBS use the new mkosi without breaking compatibility and requiring a new OBS release to be published.


bluca:
Once we have a release out, and it becomes available in stable distributions, then I can do the change in OBS, which is published once a year or so

DaanDeMeyer:
Is it OK if I leave this open until you figure out OBS completely? I expect more than this might be necessary

bluca:
sure

DaanDeMeyer (approved):

PR #1660, opened 2023-07-03
Make sure PATH is correct when using --tools-tree

DaanDeMeyer:


mkosi/run.py
```diff
@@ -366,6 +366,13 @@ def bwrap(
     **kwargs: Any,
 ) -> CompletedProcess:
     with bwrap_cmd(root=root, apivfs=apivfs) as bwrap:
+        if root:
+            # If a root is specified, we should ignore any local modifications made to PATH as any of those
```
behrmann, line 370: Something similar to this comment should also be in the docs, e.g. 
> When using `ToolsTree=` the `PATH` variable will be set to `/usr/bin/` and `/usr/sbin/` in that tree.

mkosi/run.py
```diff
@@ -366,6 +366,13 @@ def bwrap(
...
+            # tools might not work anymore when /usr is replaced wholesale. We also make sure that both
```
behrmann, line 371: Is the sentence starting with "We also make sure" missing a part, i.e. what it is we're making sure?

PR #1662, opened 2023-07-04
Protect common system directories when running with bubblewrap

DaanDeMeyer:
Let's make sure that even when we run as root the tools we run can't brick the system by making core system directories read-only bind mounts when running tools with bubblewrap.

As an example of the stuff this protects again, let's say a package installs an absolute symlink to /usr and in the finalize script a user tries to remove everything under this symlink, they'll end up erasing /usr from their host system. By making /usr read-only, we prevent this from happening.

### DIFF
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -306,10 +306,12 @@ def bwrap_cmd(
         "--dev-bind", "/", "/",
         "--chdir", Path.cwd(),
         "--die-with-parent",
+        "--ro-bind", (root or Path("/")) / "usr", "/usr",
     ]
 
-    if root:
-        cmdline += ["--bind", root / "usr", "/usr"]
+    for d in ("/etc", "/opt", "/srv", "/boot", "/efi"):
+        if Path(d).exists():
+            cmdline += ["--ro-bind", d, d]
 
     if apivfs:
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
(I once erased half of my /usr/share directory while hacking on mkosi, this is an attempt to prevent that from happening again)

mkosi/run.py
```diff
@@ -306,10 +306,12 @@ def bwrap_cmd(
         "--dev-bind", "/", "/",
         "--chdir", Path.cwd(),
         "--die-with-parent",
+        "--ro-bind", (root or Path("/")) / "usr", "/usr",
     ]
 
-    if root:
-        cmdline += ["--bind", root / "usr", "/usr"]
+    for p in ("/etc", "/opt", "/srv", "/root", "/boot", "/efi"):
```
bluca, line 312: conceivably on a build vm/chroot someone might be executing from under /root, so maybe first check that the working dir is not under it?

behrmann: On a similar note: With a check for the current working directory, this could also include `/home`? 

Protecting at least the unnecessary (for us) bits of `/var` might also be a good idea.

bluca (approved):

PR #1666, opened 2023-07-05
Make BuildSources more flexible

DaanDeMeyer:
Currently, when you want to build multiple projects together, you
have to make the other projects subdirectories of the one with the
mkosi config, which is inconvenient. To make this more flexible, let's
allow specifying multiple source trees and where to mount them under
/work/src so that multiple projects can be mounted in a fixed location
regardless of their location on the host.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -725,8 +725,10 @@ they should be specified with a boolean argument: either "1", "yes", or "true" t
 
 `BuildSources=`, `--build-sources=`
 
-: Takes a path to a source tree to mount into the development image, if
-  the build script is used.
+: Takes a list of colon-separated pairs of paths to source trees and where to mount them in the development
+  image, if the build script is used. Every target path is prefixed with `/work/src` and all build sources
+  are sorted lexicographically by mount target before mounting so that top level paths are mounted first. By
+  default, the current working directory is mounted to `/work/src`.
 
 `BuildPackages=`, `--build-package=`
 
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1977,7 +1977,6 @@ def run_shell(args: MkosiArgs, config: MkosiConfig) -> None:
         ]
 
     cmdline += ["--machine", config.output]
-    cmdline += [f"--bind={config.build_sources}:/root/src", "--chdir=/root/src"]
 
     for k, v in config.credentials.items():
         cmdline += [f"--set-credential={k}:{v}"]
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -240,7 +240,7 @@ def install_distribution(state: MkosiState) -> None:
 
 
 def install_build_packages(state: MkosiState) -> None:
-    if state.config.build_script is None or not state.config.build_packages:
+    if not need_build_packages(state.config):
         return
 
     with complete_step(f"Installing build packages for {str(state.config.distribution).capitalize()}"), mount_build_overlay(state):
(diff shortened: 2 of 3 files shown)


mkosi.md
```diff
@@ -725,8 +725,10 @@ they should be specified with a boolean argument: either "1", "yes", or "true" t
 
 `BuildSources=`, `--build-sources=`
 
-: Takes a path to a source tree to mount into the development image, if
-  the build script is used.
+: Takes a list of colon-separated pairs of paths to source trees and where to mount them in the development
+  image, if the build script is used. Every target path is prefixed with /work/src and all build sources are
```
behrmann, line 729:
```suggestion
  image, if the build script is used. Every target path is prefixed with `/work/src` and all build sources are
```

mkosi.md
```diff
@@ -725,8 +725,10 @@ they should be specified with a boolean argument: either "1", "yes", or "true" t
...
+  sorted before mounting so that top level paths are mounted first. By default, the current working directory
+  is mounted to /work/src.
```
behrmann, line 731:
```suggestion
  is mounted to `/work/src`.
```

mkosi.md
```diff
@@ -725,8 +725,10 @@ they should be specified with a boolean argument: either "1", "yes", or "true" t
 
 `BuildSources=`, `--build-sources=`
 
-: Takes a path to a source tree to mount into the development image, if
-  the build script is used.
+: Takes a list of colon-separated pairs of paths to source trees and where to mount them in the development
+  image, if the build script is used. Every target path is prefixed with /work/src and all build sources are
+  sorted before mounting so that top level paths are mounted first. By default, the current working directory
```
behrmann, line 730:
```suggestion
  sorted lexicographically by mount target before mounting so that top level paths are mounted first. By default, the current working directory
```

mkosi/__init__.py
```diff
@@ -294,45 +294,50 @@ def mount_build_overlay(state: MkosiState, read_only: bool = False) -> ContextMa
...
     if build:
         with complete_step("Running prepare script in build overlay…"), mount_build_overlay(state):
             run_workspace_command(
                 state.root,
-                ["/root/prepare", "build"],
+                ["/work/prepare", "build"],
                 network=True,
                 bwrap_params=bwrap,
-                env=dict(SRCDIR="/root/src") | state.environment,
+                env=dict(SRCDIR="/work/src") | state.environment,
```
behrmann, line 328: Shouldn't that actually be the other way 'round? This way `SRCDIR` in `state.environment` can clobber our `SRCDIR`.

behrmann (commented):
LGTM, a few nits and a question.

mkosi/__init__.py
DaanDeMeyer: Yeah, but we do this in a few places, let's address this in another PR

DaanDeMeyer:
Only documentation changes, so let's merge

septatrix:
How do I specify multiple sources - with multiple `BuildSources=` statements?

Also: This is missing in the NEWS file (and the release notes). Please, please, make them more comprehensive. Today I was updating a project from v14 to v15 and practically all problems I faced were due to undocumented changes between versions! This is a really great tool which fits my use-case perfectly but small things like these make it less enjoyable to use.

Maybe a PR template with a checkbox for "Has the change been documented in the NEWS.md file" would help. Not only for external contributors but also for the maintainers...

PR #1668, opened 2023-07-07
Add BuildSources mount points

MariusSchiffer:
Only `/work/src` was created before, so any deeper mounts like `../testprogram:/testprogram` failed as bwrap couldn't create the directory for the mountpoint in the read-only overlay.


DaanDeMeyer (approved):

behrmann:
CI failure is unrelated, merging.

PR #1669, opened 2023-07-07
Reimplement version comparison in pure python

behrmann:
This reimplements `GenericVersion` in pure Python, since shelling out to systemd-analyze is not nicely compatible with tool trees, since we need it at config parsing time already, but will have the tool tree only afterwards.

This series first adds (almost) all the tests for `systemd-analyze compare-versions` I could find and then implements the algorithm described in the [version format specification](https://uapi-group.org/specifications/specs/version_format_specification/). 

It did have to deviate from the spec in two points, though, which I think might be bugs in the spec and/or `systemd-analyze`, which I'll be happy to fix if they are.

The first thing I ran into were the fixed examples from https://github.com/uapi-group/specifications/pull/62. Looking at the code for [strverscmp_improved](https://github.com/systemd/systemd/blob/main/src/fundamental/string-util-fundamental.c#L128) the check from step 2 in the spec comes only after step 3, which looks to me like step 2 in the spec should be dropped.

The other thing is that I am confused by
```
$ systemd-analyze compare-versions ^ .; echo $?
^ < .
12
```
(with exit status 12 meaning the left side is smaller) since the spec says
> 6. If the remaining part of one of strings starts with ^: if the other remaining part does not start with ^, the string with ^ compares higher. Otherwise, both caret characters are skipped.

and the check for dot only comes afterwards. This looks like `^` works the same as `-` or `.`, which makes sense, since there is no difference between them in the code
```C
/* Handle '-', which separates version and release, e.g 123.4-3.1.fc33.x86_64 */
if (*a == '-' || *b == '-') {
        /* The string prefixed with '-' is older (e.g., 123-9 vs 123.1-1) */
        r = CMP(*a != '-', *b != '-');
        if (r != 0)
                return r;

        a++;
        b++;
}

/* Handle '^'. Used for patched release. */
if (*a == '^' || *b == '^') {
        r = CMP(*a != '^', *b != '^');
        if (r != 0)
                return r;

        a++;
        b++;
}

/* Handle '.'. Used for point releases. */
if (*a == '.' || *b == '.') {
        r = CMP(*a != '.', *b != '.');
        if (r != 0)
                return r;

        a++;
        b++;
}
```
So should the caret one actually `return -r`?

/cc @keszybz

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -19,6 +19,7 @@ import subprocess
 import sys
 import textwrap
 from collections.abc import Sequence
+from itertools import takewhile
 from pathlib import Path
 from typing import Any, Callable, Optional, Type, Union, cast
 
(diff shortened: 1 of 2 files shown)


mkosi/config.py
```diff
@@ -1809,44 +1810,153 @@ def parse(self, argv: Optional[Sequence[str]] = None) -> tuple[MkosiArgs, tuple[
 
 
 class GenericVersion:
+    _EQUAL = 0
+    _RIGHT_SMALLER = 11
+    _LEFT_SMALLER = 12
```
DaanDeMeyer, line 1815: Why 11 and 12? Shouldn't we follow C here and do 0, -1 and 1?

mkosi/config.py
```diff
@@ -1809,44 +1810,153 @@ def parse(self, argv: Optional[Sequence[str]] = None) -> tuple[MkosiArgs, tuple[
...
+
     def __init__(self, version: str):
         self._version = version
 
+    @classmethod
+    def compare_versions(cls, v1: str, v2: str) -> int:
+        """Implements comparison according to UAPI Group Version Format Specification"""
+        def rstrip_invalid_version_chars(s: str) -> str:
+            valid_version_chars ={*string.ascii_letters, *string.digits, "~", "-", "^", "."}
```
DaanDeMeyer, line 1824: Missing space here

mkosi/config.py
```diff
@@ -1809,44 +1810,153 @@ def parse(self, argv: Optional[Sequence[str]] = None) -> tuple[MkosiArgs, tuple[
...
+            return "".join(takewhile(lambda c: c.isalpha(), s))
+
+        while True:
+            #breakpoint()
+            # Any characters which are outside of the set of listed above (a-z, A-Z, 0-9, -, ., ~,
+            # ^) are skipped in both strings. In particular, this means that non-ASCII characters
+            # that are Unicode digits or letters are skipped too.
+            v1 = rstrip_invalid_version_chars(v1)
+            v2 = rstrip_invalid_version_chars(v2)
+            # If the remaining part of one of strings starts with ~: if other remaining part does
```
DaanDeMeyer, line 1852: Is "~:" intended here?

DaanDeMeyer (approved):
LGTM aside from the confusion you already mentioned, there seems to be a few colons that should be commas instead?

mkosi/config.py
behrmann: Here because 
> In the first mode — when OP is not specified — it will compare the two version strings and print either "VERSION1 < VERSION2", or "VERSION1 == VERSION2", or "VERSION1 > VERSION2" as appropriate.
>
> The exit status is 0 if the versions are equal, 11 if the version of the right is smaller, and 12 if the version of the left is smaller. (This matches the convention used by rpmdev-vercmp.)

and I wanted to make the `compare_versions` method behave like `systemd-analyze compare-versions` with only two arguments.

mkosi/config.py
behrmann: Yeah, makes more sense, just copied the text from the spec verbatim and the formatting got lost.

mkosi/config.py
DaanDeMeyer: Can we add a comment mentioning that?

tests/test_config.py
```diff
@@ -0,0 +1,60 @@
...
+    assert GenericVersion("1_") < GenericVersion("1.2")
+    assert GenericVersion("1_2_3") > GenericVersion("1.3.3")
+    assert GenericVersion("1+") == GenericVersion("1")
+    assert GenericVersion("+1") == GenericVersion("1")
+    assert GenericVersion("1+") < GenericVersion("1.2")
+    assert GenericVersion("1+2+3") > GenericVersion("1.3.3")
+
+
+def test_generic_version_strverscmp_improved_doc() -> None:
+    """Example from the doc string of strverscmp_improved in systmd/src/fundamental/string-util-fundamental.c"""
```
DaanDeMeyer, line 46: Typo here systmd

mkosi/config.py
behrmann: Added a comment.

behrmann:
I also inlined the function for step 2, which was reused for step 4, since I dropped step 2, see the mention in the original post.

behrmann:
Just checked and `rpmdev-vercmp ^ .` gives the opposite result of `systemd-analyze compare-versions ^ .`, so I think that's a bug. I'll open a companion to this PR against systemd, let's keep this open until then.

mkosi/config.py
DaanDeMeyer: So after thinking a bit more, I think we shouldn't bother with 11 and 12 unless we need compat with rpmdev-vercmp somewhere, which I don't think we need. So I'd just use -1, 1 and 0 here instead.

PR #1675, opened 2023-07-13
Beef up --kernel-command-line-extra= defaults a little bit

DaanDeMeyer:
Let's provide a better experience out of the box when booting an image:
- Always blacklist vmw_vmci so it can't interfere with vsock
- Always pull in network-online.target
- Configure systemd-network-generator for enp0s1 and host0 if not
  configured explicitly
- Use a default loglevel where only warnings or higher are logged to
  the serial console by the kernel
- Set SYSTEMD_SULOGIN_FORCE=1 unless configured explicitly so users
  get a shell in the initrd if something goes wrong, even if no root
  password is configured

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -227,7 +227,7 @@ def config_default_release(namespace: argparse.Namespace) -> str:
         Distribution.alma: "9",
         Distribution.mageia: "cauldron",
         Distribution.debian: "testing",
-        Distribution.ubuntu: "jammy",
+        Distribution.ubuntu: "lunar",
         Distribution.opensuse: "tumbleweed",
         Distribution.openmandriva: "cooker",
         Distribution.gentoo: "17.1",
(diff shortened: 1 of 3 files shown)


mkosi/config.py
```diff
@@ -981,7 +981,7 @@ class MkosiConfigParser:
             metavar="OPTIONS",
             section="Content",
             parse=config_make_list_parser(delimiter=" "),
-            default=["console=ttyS0"],
+            default=["console=ttyS0", "loglevel=4"],
```
bluca, line 984: should this be documented in the manpage too?

DaanDeMeyer:
Changed to modify --kernel-command-line-extra= so we only modify the extra kernel cmdline stuff added when booting the VM with qemu/nspawn (which is less invasive).

Also added a commit to visualize these extra kernel cmdline arguments in the summary to make them more visible. I think that's sufficient and we don't have to document them per se.

bluca (approved):

behrmann (approved):

PR #1676, opened 2023-07-14
Fix logdir handling for dnf

DaanDeMeyer:
logdir= is always taken relative to the installroot. Let's leave it at it's default value and remove any log files from the install root after running dnf.


behrmann:
looks ok to me, although in general a more helpful approach could be to move such logs in an external directory (if it exists), since this can make debugging a bit easier.

DaanDeMeyer:
Yeah, but this will need a bit more holistic approach, since they should persist after mkosi exits to be useful. Currently you just do `run(["bash"], stdin=sys.stdin, cwd=state.workspace)` before the logs are deleted and you can read them.

PR #1677, opened 2023-07-16
Rework chrooting and add scripts support to bwrap()

DaanDeMeyer:


mkosi/__init__.py
```diff
@@ -310,31 +311,31 @@ def run_prepare_script(state: MkosiState, build: bool) -> None:
...
             shutil.rmtree(state.root / "work")
     else:
         with complete_step("Running prepare script…"):
-            run_workspace_command(
-                state.root,
-                ["/work/prepare", "final"],
-                network=True,
-                bwrap_params=bwrap,
+            bwrap(
+                ["chroot", "/work/prepare", "final"],
```
behrmann, line 335: A renaming to `/work/prepare` ended up in the commit "Rename extra argument of invoke_apt() to packages"

mkosi/run.py
```diff
@@ -359,101 +403,91 @@ def bwrap_cmd(
...
-            logging.error(f"\"{' '.join(str(s) for s in cmd)}\" returned non-zero exit code {e.returncode}.")
-            if ARG_DEBUG_SHELL.get():
-                run([*cmdline, "sh"], stdin=sys.stdin, check=False, env=env, log=False)
-            raise e
-        finally:
-            if tmp.is_symlink():
-                resolve.unlink(missing_ok=True)
-                shutil.move(tmp, resolve)
+def which(program: str, tools: Optional[Path]) -> Optional[Path]:
+    result = bwrap(["sh", "-c", f"command -v {program}"], tools=tools, check=False, stdout=subprocess.PIPE)
```
behrmann, line 486: Instead of shelling out for each which, why not call `shutil.which` with the `path` argument? This would be a bit more lightweight than starting the whole bwrap machinery and a shell every time.

mkosi/run.py
```diff
@@ -359,101 +403,91 @@ def bwrap_cmd(
...
-
     if network:
-        # Bubblewrap does not mount over symlinks and /etc/resolv.conf might be a symlink. Deal with this by
-        # temporarily moving the file somewhere else.
-        if resolve.is_symlink():
-            shutil.move(resolve, tmp)
+        resolve = Path("etc/resolv.conf")
+        if (root / resolve).is_symlink():
+            # bubblewrap cannot mount over symlinks so handle that manually here.
+            resolve = (root / resolve).readlink()
```
behrmann, line 475: The idea here is that since overmounting is not possible, we just exchange the target? If so, the comment could be a bit clearer.

mkosi/run.py
DaanDeMeyer: Because we want our implementation of `which()` to run in the same environment the command will be executed in. I think it's valuable to match the environment as much as possible here.

We could use `which` instead of the shell but that's another tool that needs to be installed in the tools tree whether the shell is already needed for other things.

behrmann: I generally agree that we want to match the execution environment as closely as possible, but the only thing `shutil.which` does is go through the files in a list of directories and return the first matching the given name that is both readable and executable. I assume `command` does little else, so I'm not sure there is added value to set up another execution environment when we can make the same decision for the same paths without.

mkosi/__init__.py
```diff
@@ -310,31 +311,31 @@ def run_prepare_script(state: MkosiState, build: bool) -> None:
...
-            run_workspace_command(
-                state.root,
-                ["/work/prepare", "build"],
-                network=True,
-                bwrap_params=bwrap,
+            bwrap(
+                ["chroot", "/work/prepare", "build"],
+                tools=state.config.tools_tree,
+                apivfs=state.root,
+                scripts=dict(chroot=chroot_cmd(state.root, options=options, network=True)),
```
behrmann, line 328: I know you like `dict` better, because the keys don't need to be stringified, but hear and in the similar examples below, I really think it lessens readability, because it's "just another function call"
```suggestion
                scripts={"chroot": chroot_cmd(state.root, options=options, network=True)},
```

behrmann (approved):
LGTM. One nit, but that can be fixed later. CI failures look unrelated, but apparently we're not emerging systemd-boot on gentoo.

PR #1679, opened 2023-07-17
Make docs available via the help verb.

behrmann:
This PR makes the docs available via `args.cmdline` on the help verb.

This allows, especially when running different versions of mkosi, to always have the right documentation on hand no matter how that version of mkosi was installed.


behrmann:
I dropped the pandoc stuff from `setup.py`, since it's hooked into running `setup.py`, i.e. it runs during install, which is not ideal, since installing then fails if pandoc is not available. I'm not sure hooking it into `setup.py` is the best idea, since the Python world doesn't really care all that much about data outside of `lib/pythonX.Y`, so there's no real mechanism to install to those places (which I think is a good thing). Also, I'm not entirely sure this could be carried over to `pyproject.toml`.  So, instead of shoehorning this in here, generating the man page and putting it in the right place for packaging is something that should happen explicitly during distro packaging. I'll rework this a bit to hopefully make this a bit easier.

DaanDeMeyer (changes requested):
I think we should just add a documentation verb instead of making it a subcommand of the help verb. Also quite a few typos in the news segment.

mkosi/config.py
```diff
@@ -595,6 +596,24 @@ def __call__(
...
+        option_string: Optional[str] = None
+    ) -> None:
+        with importlib.resources.path("mkosi.resources", "mkosi.1") as man:
+            if man.exists():
+                run(["man", man])
+                parser.exit()
+
+        md = importlib.resources.read_text("mkosi.resources", "mkosi.md")
+        page(md, namespace.pager)
+        parser.exit()
```
DaanDeMeyer, line 614: I would change this to only show the man page and error out if the man page is not available. Or even better, generate it on demand from mkosi.md if pandoc is installed and error out if pandoc is not installed. But I don't think we should show the raw markdown file here tbh.

behrmann:
> I think we should just add a documentation verb instead of making it a subcommand of the help verb. Also quite a few typos in the news segment.

Sure, I'll move it over.

mkosi/config.py
behrmann: I thought about falling back to pandoc if no man page is available and had the code in there in one version, but I'm not sure we should hardcode that here, since it would make pandoc a recommends or similar in most distros.

keszybz: That should be fine. If the distro puts the man page in the appropriate place, then the pandoc dependency is not needed.

mkosi/__init__.py
```diff
@@ -1637,6 +1639,37 @@ def bump_image_version(uid: int = -1, gid: int = -1) -> None:
...
+                    return run(["man", "--local-file", "-"], input=pandoc.stdout)
+            elif form == DocFormat.markdown:
+                md = importlib.resources.read_text("mkosi.resources", "mkosi.md")
+                return page(md, args.pager)
+            elif form == DocFormat.system:
+                return run(["man", "mkosi"])
+        except (FileNotFoundError, subprocess.CalledProcessError) as e:
+            if formats:
+                continue
+            raise e
```
DaanDeMeyer, line 1670: `if not formats: raise e`?

DaanDeMeyer (changes requested):

mkosi/__init__.py
```diff
@@ -1637,6 +1639,37 @@ def bump_image_version(uid: int = -1, gid: int = -1) -> None:
...
+                    logging.warning("pandoc is not available")
+                with importlib.resources.path("mkosi.resources", "mkosi.md") as md:
+                    pandoc = run(["pandoc", "-t", "man", "-s", md], stdout=subprocess.PIPE)
+                    return run(["man", "--local-file", "-"], input=pandoc.stdout)
+            elif form == DocFormat.markdown:
+                md = importlib.resources.read_text("mkosi.resources", "mkosi.md")
+                return page(md, args.pager)
+            elif form == DocFormat.system:
+                return run(["man", "mkosi"])
+        except (FileNotFoundError, subprocess.CalledProcessError) as e:
```
DaanDeMeyer, line 1667: I would translate `FileNotFoundError` into a `die()` if we're going to raise so we don't show an ugly exception stacktrace.

mkosi/__init__.py
```diff
@@ -1637,6 +1639,37 @@ def bump_image_version(uid: int = -1, gid: int = -1) -> None:
...
+def show_docs(args: MkosiArgs) -> None:
+    if args.doc_format == DocFormat.auto:
+        formats = [DocFormat.system, DocFormat.markdown, DocFormat.pandoc, DocFormat.man]
+    else:
+        formats = [args.doc_format]
+
+    while formats:
+        form = formats.pop()
+        try:
+            if form == DocFormat.man:
```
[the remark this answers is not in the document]
behrmann: They are, in stack order ;) I'll make it a `.pop(0` and switch the order above.

mkosi/resources/mkosi.md
```diff
@@ -0,0 +1,1543 @@
...
+
+`genkey`
+
+: Generate a pair of SecureBoot keys for usage with the
+  `SecureBootKey=`/`--secure-boot-key=` and
+  `SecureBootCertificate=`/`--secure-boot-certificate=` options.
+
+`documentation`
+
+: Show mkosi's documentation.
```
DaanDeMeyer, line 127: Can you repeat that blurb from NEWS here as well? About generating the man page with pandoc and putting it in resources?

PR #1681, opened 2023-07-17
More tools fixes

DaanDeMeyer:


mkosi/manifest.py
```diff
@@ -101,13 +101,13 @@ def record_rpm_packages(self, root: Path) -> None:
         # On Debian, rpm/dnf ship with a patch to store the rpmdb under ~/ so rpm
         # has to be told to use the location the rpmdb was moved to.
         # Otherwise the rpmdb will appear empty. See: https://bugs.debian.org/1004863
-        dbpath = "/usr/lib/sysimage/rpm"
+        dbpath = "usr/lib/sysimage/rpm"
         if not (root / dbpath).exists():
```
behrmann, line 105: This was wrong before, because otherwise `dbpath` completely clobbers `root`, making the current expression equivalent to `not dbpath.exists()`. This probably should have been
```suggestion
        if not (root / dbpath.relative_to("/")).exists():
```
I think that might be the shorter version of what is achieved in this commit, no?

behrmann (approved):
LGTM otherwise.

PR #1685, opened 2023-07-19
gentoo: stop passing USE environment to emerge

257:
also make sure we stay with systemd-9999 (version built from git) in order keep up with changes in systemd

### DIFF
--- a/.github/mkosi.pkgmngr/gentoo/etc/portage/package.use/systemd
+++ b/.github/mkosi.pkgmngr/gentoo/etc/portage/package.use/systemd
@@ -1,0 +1,1 @@
+sys-apps/systemd boot
--- a/.github/mkosi.conf.d/20-gentoo.conf
+++ b/.github/mkosi.conf.d/20-gentoo.conf
@@ -3,8 +3,10 @@ Distribution=gentoo
 
 [Content]
 Environment=PORTAGE_BINHOST=https://raw.githubusercontent.com/257/binpkgs/master
-Packages=sys-kernel/gentoo-kernel-bin
-         sys-apps/systemd
+# needed for system-9999
+PackageManagerTrees=mkosi.pkgmngr/gentoo
+Packages=sys-kernel/gentoo-kernel-bin[initramfs]
+         =sys-apps/systemd-9999[boot]
          # Failed to execute /usr/lib/systemd/system-environment-generators/10-gentoo-path: No such file or directory
          # Failed to execute /usr/lib/systemd/system-generators/gentoo-local-generator: No such file or directory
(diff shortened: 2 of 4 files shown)


.github/mkosi.pkgmngr/gentoo/etc/portage/binrepos.conf
```diff
@@ -0,0 +1,2 @@
+[binhost]
+sync-uri = https://raw.githubusercontent.com/257/binpkgs/master
```
DaanDeMeyer, line 2: Why bother? We already set `PORTAGE_BINHOST` anyway

257: dropped, never mind.

.github/mkosi.pkgmngr/gentoo/etc/portage/package.use/systemd
```diff
@@ -0,0 +1 @@
+sys-apps/systemd boot
```
DaanDeMeyer, line 1: Can you get rid of the `boot` flag upstream when systemd v254 releases? Given that gnu-efi isn't necessary anymore from v254 onwards I'm not sure if putting sd-boot behind a flag still provides much value

257: i'm not the maintainer but i see it's used in number of places [here](https://github.com/gentoo/gentoo/blob/master/sys-apps/systemd/systemd-9999.ebuild):
```
...
boot? ( $(python_gen_any_dep 'dev-python/pyelftools[${PYTHON_USEDEP}]') )
...
$(meson_native_use_bool boot bootloader)
...
```

DaanDeMeyer: Yeah unless gentoo really cares about the extra python build dependencies, I would get rid of `boot` in the systemd package and just build the boot stuff by default.

257:
@DaanDeMeyer can you verify if the fail is related to this commit?

DaanDeMeyer:
> @DaanDeMeyer can you verify if the fail is related to this commit?

How old is the systemd we're running? The failure we're seeing should be fixed in git main of upstream.

257:
got HEAD

DaanDeMeyer:
> git HEAD

And we're rebuilding that every time and not using a binary package or so that might be older?

257:
yes, good catch.

behrmann:
CI failure for gentoo disk looks unrelated and is only when powering down the machine. Merging.

PR #1686, opened 2023-07-20
Stop tmpfiles from creating subvolumes

DaanDeMeyer:
Generally when building an image there's no point in creating nested
subvolumes in these directories so let's disable that.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1499,16 +1499,28 @@ def run_depmod(state: MkosiState) -> None:
 
 
 def run_sysusers(state: MkosiState) -> None:
+    if not shutil.which("systemd-sysusers"):
+        logging.info("systemd-sysusers is not installed, not generating system users")
+        return
+
     with complete_step("Generating system users"):
         run(["systemd-sysusers", "--root", state.root])
 
 
 def run_preset(state: MkosiState) -> None:
+    if not shutil.which("systemctl"):
+        logging.info("systemctl is not installed, not applying presets")
+        return
+
     with complete_step("Applying presets…"):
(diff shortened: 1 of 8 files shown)


mkosi/__init__.py
```diff
@@ -1567,19 +1579,15 @@ def run_selinux_relabel(state: MkosiState) -> None:
...
-    fc = Path('/etc/selinux') / policy / 'contexts/files/file_contexts'
+    if not shutil.which("setfiles"):
+        logging.info("setfiles is not installed, not relabeling files")
+        return
 
-    # We want to be able to relabel the underlying APIVFS mountpoints, so mount root non-recursive to a
-    # temporary location so that the underlying mountpoints become visible.
-    cmd = f"mkdir /tmp/relabel && mount --bind / /tmp/relabel && exec setfiles -m -r /tmp/relabel -F {fc} /tmp/relabel || exit $?"
+    fc = state.root / "etc/selinux" / policy / "contexts/files/file_contexts"
+    bin = next((state.root / "etc/selinux" / policy / "policy").iterdir())
```
behrmann, line 1587: I'm not an SELinux expert, but the binary policy is just the first random file in that directory? That seems like a weird design choice. Don't they have any known in advance filenames?

If there's a reasonable possibility that the directory might be empty, it would make sense to catch that error since `next` will then throw a `StopIteration`. 

Nit: `binpol` might be a better name, I only learnt that's a binary policy when I looked up the `setfiles` options (how I do love programs without descriptive switches)

behrmann (commented):
LGTM, just a few SELinux questions.

PR #1688, opened 2023-07-20
Simplify become_root()

DaanDeMeyer:
- Instead of forking, use spawn() to start the newuidmap, newgidmap processes early
- Instead of multiprocessing.Event(), use flock to handle the necessary locking

### DIFF
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -4,8 +4,8 @@ import asyncio
 import asyncio.tasks
 import ctypes
 import ctypes.util
+import fcntl
 import logging
-import multiprocessing
 import os
 import pwd
 import queue
@@ -22,7 +22,7 @@ from typing import Any, Awaitable, Mapping, Optional, Sequence, Tuple, Type, Typ
 
 from mkosi.log import ARG_DEBUG, ARG_DEBUG_SHELL, die
 from mkosi.types import _FILE, CompletedProcess, PathString, Popen
-from mkosi.util import InvokingUser, make_executable
+from mkosi.util import InvokingUser, flock, make_executable
 
 CLONE_NEWNS = 0x00020000
 CLONE_NEWUSER = 0x10000000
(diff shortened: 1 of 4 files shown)


mkosi/run.py
```diff
@@ -79,42 +79,41 @@ def become_root() -> tuple[int, int]:
...
+    newuidmap = [
+        "flock", "--exclusive", "--no-fork", "/etc/subuid", "newuidmap", pid,
+        0, subuid, SUBRANGE - 100,
+        SUBRANGE - 100, os.getuid(), 1,
+        SUBRANGE - 100 + 1, subuid + SUBRANGE - 100 + 1, 99
+    ]
 
-        os._exit(0)
+    newgidmap = [
+        "flock", "--exclusive", "--no-fork", "/etc/subuid", "newgidmap", pid,
```
behrmann, line 97: Shouldn't that be?
```suggestion
        "flock", "--exclusive", "--no-fork", "/etc/subgid", "newgidmap", pid,
```
Probably not, in light of the comment further down.

mkosi/run.py
```diff
@@ -79,42 +79,41 @@ def become_root() -> tuple[int, int]:
...
+    newuidmap = [str(x) for x in newuidmap]
+    newgidmap = [str(x) for x in newgidmap]
+
+    # newuidmap and newgidmap have to run from outside the user namespace to be able to assign a uid mapping
+    # to the process in the user namespace. The mapping can only be assigned after the user namespace has
+    # been unshared. To make this work, we first lock /etc/subuid, then spawn the newu!dmap and newgidmap
+    # processes, which we execute using flock so they don't execute before they can get a lock on /etc/subuid,
+    # then we unshare the user namespace and finally we unlock /etc/subuid, which allows the newuidmap and
+    # newgidmap processes to execute. we then wait for the processes to finish before continuing.
+    with flock(Path("/etc/subuid")) as fd, spawn(newuidmap) as uidmap, spawn(newgidmap) as gidmap:
```
behrmann, line 112: I'm a bit surprised this works syntactically, since `spawn` isn't a context manager, but `Popen` is, so that's why it works. TIL.

mkosi/run.py
```diff
@@ -79,42 +79,41 @@ def become_root() -> tuple[int, int]:
...
 
-    unshare(CLONE_NEWUSER)
-    event.set()
-    os.waitpid(child, 0)
+    newuidmap = [str(x) for x in newuidmap]
+    newgidmap = [str(x) for x in newgidmap]
+
+    # newuidmap and newgidmap have to run from outside the user namespace to be able to assign a uid mapping
+    # to the process in the user namespace. The mapping can only be assigned after the user namespace has
+    # been unshared. To make this work, we first lock /etc/subuid, then spawn the newu!dmap and newgidmap
```
behrmann, line 108:
```suggestion
    # been unshared. To make this work, we first lock /etc/subuid, then spawn the newuidmap and newgidmap
```

mkosi/run.py
behrmann: On a different note, why can we get away with just locking `/etc/subuid`? Why don't we have to lock `/etc/subgid`, too? Is there maybe an implicit assumption, that they will assign the same range?

DaanDeMeyer: It doesn't matter what we lock, as long as both the main process and the child processes lock the same thing. The only thing we're trying to achieve with the locking here is that newuidmap and newgidmap run "after" the main process has unshared the user namespace. Given we've verified in this function that /etc/subuid exists, it's as good a file as any to use for the locking.

behrmann (approved):

PR #1689, opened 2023-07-20
gentoo: swith to llvm (clang)

257:
This will pave the way for actual arch support and cross-compiling

### DIFF
--- a/mkosi/distributions/gentoo.py
+++ b/mkosi/distributions/gentoo.py
@@ -99,7 +99,7 @@ class GentooInstaller(DistributionInstaller):
 
         with urllib.request.urlopen(stage3tsf_path_url) as r:
             # e.g.: 20230108T161708Z/stage3-amd64-nomultilib-systemd-mergedusr-20230108T161708Z.tar.xz
-            regexp = rf"^[0-9]+T[0-9]+Z/stage3-{arch}-nomultilib-systemd-mergedusr-[0-9]+T[0-9]+Z\.tar\.xz"
+            regexp = rf"^[0-9]+T[0-9]+Z/stage3-{arch}-llvm-systemd-mergedusr-[0-9]+T[0-9]+Z\.tar\.xz"
             all_lines = r.readlines()
             for line in all_lines:
                 if (m := re.match(regexp, line.decode("utf-8"))):


mkosi/distributions/gentoo.py
```diff
@@ -99,7 +99,7 @@ def install(cls, state: MkosiState) -> None:
 
         with urllib.request.urlopen(stage3tsf_path_url) as r:
             # e.g.: 20230108T161708Z/stage3-amd64-nomultilib-systemd-mergedusr-20230108T161708Z.tar.xz
-            regexp = rf"^[0-9]+T[0-9]+Z/stage3-{arch}-nomultilib-systemd-mergedusr-[0-9]+T[0-9]+Z\.tar\.xz"
+            regexp = rf"^[0-9]+T[0-9]+Z/stage3-{build_arch}-llvm-systemd-mergedusr-[0-9]+T[0-9]+Z\.tar\.xz"
```
behrmann, line 102: `build_arch` doesn't seem to be defined.

257: fixed

DaanDeMeyer:
@257 This seems to have an effect of kernel image size? At least suddenly the EFI partition isn't big enough anymore to fit everything.

257:
looking into it.

DaanDeMeyer:
I've also noticed that setting `FEATURES` doesn't seem to actually work. When running an ebuild none of our features from the environment variable are actually used.

257:
i saw that on #gentoo-portage. something odd at your end, maybe?

DaanDeMeyer:
@257 Does gentoo do kernel module compression by default? If not we should probably configure the use flags for that by default, otherwise the initrd is just going to be huge.

257:
```
root@asus:~ equery uses gentoo-kernel-bin
[ Legend : U - final flag setting for installation]
[        : I - package is installed with flag     ]
[ Colors : set, unset                             ]
 * Found these USE flags for sys-kernel/gentoo-kernel-bin-6.4.4:
 U I
 + + initramfs : Build initramfs along with the kernel.
 - - test      : Enable dependencies and/or preparations necessary to run tests (usually controlled by FEATURES=test but can be toggled independently)
 -
```

257:
what's the state of building custom kernels in `mkosi`? could be relevant here.

257:
> what's the state of building custom kernels in `mkosi`? could be relevant here.

all we need is for the image to boot with `qemu`, more elaborate configs cab be provided too.

so a minimal `.config` with `CONFIG_MODULES=n` and we would have the smallest kernel in all distros. no need for modules and kernel's compression can be whatever we want, `CONFIG_*_[DE]COMPRESS=` stuff.

basically
- mount `/usr/src` from `cache` or `workspace` because we don't need the actual souce
- `emerge gentoo-sources`
- build the kernel

obviously it adds to the total build time though.

DaanDeMeyer:
This doesn't seem to cause any more regressions

PR #1692, opened 2023-07-21
More stuff

DaanDeMeyer:


mkosi/distributions/debian.py
```diff
@@ -119,6 +115,13 @@ def install_packages(cls, state: MkosiState, packages: Sequence[str], apivfs: bo
 
         policyrcd.unlink()
 
+        for d in state.root.glob("boot/vmlinuz-*"):
+            kver = d.name.removeprefix("vmlinuz-")
+            vmlinuz = state.root / "usr/lib/modules" / kver / "vmlinuz"
+            if not vmlinuz.exists():
+                shutil.move(d, vmlinuz)
```
behrmann, line 122: Should this maybe be a copy instead of the move so as not to break stuff that depends on the original location? A symlink might be a solution too, if `/boot/` supports it.

mkosi/__init__.py
```diff
@@ -2119,7 +2103,7 @@ def bump_image_version(uid: Optional[int] = None, gid: Optional[int] = None) ->
         logging.info(f"Increasing last component of version by one, bumping '{version}' → '{new_version}'.")
 
     Path("mkosi.version").write_text(f"{new_version}\n")
-    if uid and gid:
+    if uid is not None and gid is not None:
```
behrmann, line 2106: I wondered whether this had an edge case where `uid` or `gid` could be non-`None` and `os.chown` would be called with one `None` value, which would throw an error, but this is caught by an assert at the beginning of the function (why is that a requirement anyway?). Why don't we make the default in the signature `-1` instead of `None` and type this as an `int`. Passing `-1` will make the chown a noop and we don't have to have this check anymore.

mkosi/util.py
```diff
@@ -321,3 +267,14 @@ def flock(path: Path) -> Iterator[int]:
         yield fd
     finally:
         os.close(fd)
+
+
+@contextlib.contextmanager
+def withenv(env: Mapping[str, Any]) -> Iterator[None]:
```
behrmann, line 273: Bikeshed: `with withenv` sounds a bit weird, how about `tempenv`?

mkosi/tree.py
```diff
@@ -61,4 +76,21 @@ def btrfs_maybe_snapshot_subvolume(config: MkosiConfig, src: Path, dst: Path) ->
...
+        dst = dst / src.name
+
+    try:
+        src.rename(dst)
+        return
+    except OSError as e:
+        if e.errno != errno.EXDEV:
+            raise e
+
+    copy_tree(config, src, dst)
```
behrmann, line 96:
```suggestion
        if e.errno != errno.EXDEV:
            raise e

        copy_tree(config, src, dst)
```
to make it visually a bit clearer, that this will only be hit in the exception case. That being said, the semantic is a bit weird, since in the except case the source will still be there, no?

behrmann (commented):
LGTM, otherwise, especially like the `MkosiState` as a context manager.

mkosi/util.py
DaanDeMeyer: Renamed to `scopedenv()`

mkosi/tree.py
DaanDeMeyer: Oops I added an `rm -rf` to remove the tree when we copy it.

PR #1697, opened 2023-07-22
gentoo: Set FEATURES by appending to /etc/portage/make.conf

DaanDeMeyer:
Setting via the environment variable doesn't work in all cases, so let's append to /etc/portage/make.conf instead. This allows us to get rid of the custom devtpms for portage as it doesn't try to chown tty's anymore now that userpriv is actually disabled.

### DIFF
--- a/mkosi/distributions/gentoo.py
+++ b/mkosi/distributions/gentoo.py
@@ -19,7 +19,6 @@ from mkosi.types import PathString
 def invoke_emerge(state: MkosiState, packages: Sequence[str] = (), apivfs: bool = True) -> None:
     bwrap(
         cmd=[
-            "sh", "-c", "chmod 1777 /dev/shm && exec $0 \"$@\" || exit $?",
             "emerge",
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
cc @257

mkosi/distributions/gentoo.py
```diff
@@ -142,6 +126,25 @@ def install(cls, state: MkosiState) -> None:
...
+            "-userfetch",
+            "-userpriv",
+            "-usersandbox",
+            "-usersync",
+            "-ebuild-locks",
+            "parallel-install",
+            *(["noman", "nodoc", "noinfo"] if state.config.with_docs else []),
+        ])
+
+        # Setting FEATURES via the environment variable does seem to apply to ebuilds in portage, so we
```
257, line 143: *does NOT

DaanDeMeyer:
Only a comment fix so let's merge

PR #1698, opened 2023-07-23
Gentoo fixes

DaanDeMeyer:
- Drop --update, we never want to update packages, only install new ones
- Drop --changed-use and --new-use, these apply to updating an existing system, we're always installing a new one
- Drop --deep and --complete-graph-if-new-use, these don't seem to have any noticeable effect
- Move getbinpkg from FEATURE to emerge CLI option --getbinpkg
- Do not disable spinner and candy option
- Explicitly disable all sandboxing features


DaanDeMeyer:
cc @257

257:
looks good to me, let wait for the CI.

DaanDeMeyer:
CI is the usual failure so let's merge

PR #1699, opened 2023-07-24
gentoo: Only support binary package repositories

DaanDeMeyer:
Supporting only binary packages will allow us to greatly simplify
our gentoo implementation:

- Won't need to make a distinction between sysroot and root anymore
- No more need for a stage 3 tarball once we get portage packaged in
fedora and other distributions
- No more messing around with profiles and use flags

The main change is that users are required to configure a binary package
repository to be able to build gentoo images. They have to either configure
a binary package repository mirror using --mirror or use a local binary
package directory by setting "Environment=PKGDIR=...".

Also contains the following changes:

- Make sure emerge shows the dep tree before installing
- Drop --quiet flags from emerge to get more info by default
- Set PORTAGE_USERNAME=root and PORTAGE_GRPNAME=root since we're in
a user namespace anyway so it doesn't matter to run as root

### DIFF
--- a/.github/mkosi.conf.d/20-gentoo.conf
+++ b/.github/mkosi.conf.d/20-gentoo.conf
@@ -1,15 +1,12 @@
 [Match]
 Distribution=gentoo
 
+[Distribution]
+Mirror=https://raw.githubusercontent.com/257/binpkgs/master
+
 [Content]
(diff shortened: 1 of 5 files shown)


DaanDeMeyer:
Closing until some distribution manages to package portage

PR #1701, opened 2023-07-26
Run scripts on the host by default

DaanDeMeyer:
Scripts currently run chrooted in the image. This is not great because
it means the tool you want to run from a script has to be installed in
the image, even if it has --root support to run from outside the image.

Specifically, to install extra packages from a script, you currently
have to install the package manager inside the image itself. Even then,
it might be completely different than the package manager on the host
(both in version and in build options), possibly leading to all kinds
of weird issues.

To allow users to install extra packages from scripts, we now default
to running scripts on the host. Additionally, to make life easy for
users, we provide a set of scripts in the PATH for the package managers
we support that call them with the necessary options to install packages
in the same way mkosi installs packages into the root directory. This
means all that users have to do is "dnf install xxx" from the script to
install a new package into the root directory.

Because some tools do not have a --root option and to provide an easy
migration for users that depend on scripts running in the image, we
also put a script "chroot" in the PATH which uses bubblewrap to chroot
into the image as we did before. Users can keep their old scripts working
by simply adding the following to the top of their script:

```
if [ ! "$container" = "mkosi" ]; then
    exec chroot $SCRIPT "$@"
fi
```

When running scripts on the host, no APIVFS directories are mounted into
the image. When using the "chroot" and package manager scripts, the APIVFS
directories are automatically mounted before executing the corresponding
command. The apivfs_cmd() function is introduced to make implementing this
easier.

Additionally, we now always consider the current working directory a
BuildSources= directory. BuildSources= is now used to declare additional
source directories. The current working directory can always be overmounted
with another directory by simply specifying a source directory in BuildSources=
without a target directory.

To allow scripts running on the host to find the image, we set the
BUILDROOT variable for all scripts.


DaanDeMeyer:
Still need to update the documentation.

mkosi/tree.py
```diff
@@ -79,6 +79,10 @@ def copy_tree(config: MkosiConfig, src: Path, dst: Path, *, preserve_owner: bool
         run(copy)
 
 
+def rmtree(path: Path) -> None:
+    run(["rm", "-rf", path])
```
behrmann, line 83: It's rarely an issue and shouldn't be one here in most cases, but let's make sure that `rm` doesn't parse a path as an option. 
```suggestion
    run(["rm", "-rf", "--", path])
```

Maybe something we should do in other places as well. 

The function does give me a bit of pause. I really hope we never nuke somebody's home by accident.

behrmann (commented):
I think putting `dnf.py`, `pacman.py`, `apt.py` and `zypper.py` into a submodule would bring a bit more order into this., would also make clear that these should present the same interface.

DaanDeMeyer:
> I think putting dnf.py, pacman.py, apt.py and zypper.py into a submodule would bring a bit more order into this., would also make clear that these should present the same interface.

The interface is very loose, also, I'm not sure what that submodule should be named. package_manager seems way too long.

keszybz:
> `exec chroot $SCRIPT "$@"`

Hmm, doesn't this suffer from the problem we're complaining about a lot in our code, i.e. that API filesystems are not mounted? Shouldn't we recommend something different that works correctly?

DaanDeMeyer:
> Hmm, doesn't this suffer from the problem we're complaining about a lot in our code, i.e. that API filesystems are not mounted? Shouldn't we recommend something different that works correctly?

We run scripts in an environment where API filesystems are properly mounted. Also, the script is called `chroot` but it doesn't actually call `chroot`, it calls a shell script that execs `bwrap` to do all the necessary setup. We should rename it to `mkosi-chroot` to better separate it from normal `chroot`.

keszybz:
> We run scripts in an environment where API filesystems are properly mounted. Also, the script is called `chroot` but it doesn't actually call `chroot`, it calls a shell script that execs `bwrap` to do all the necessary setup. We should rename it to `mkosi-chroot` to better separate it from normal `chroot`.

Oh, OK. I'm still reading the patches, so I'll comment more later. But I think calling it `chroot` and putting in `$PATH` could be problematic if something tries to call the real `chroot` later on. Maybe something "short and sweet" like `_chroot` ?

mkosi/__init__.py
```diff
@@ -1051,24 +1051,56 @@ def empty_directory(path: Path) -> None:
...
 
 
+def getxattr(path: Path, name: str) -> Optional[str]:
+    result = run(["getfattr", "--name", f"user.{name}", "--only-values", "--absolute-names", path],
+                 stdout=subprocess.PIPE, stderr=subprocess.DEVNULL, check=False)
+    return result.stdout.strip() if result.returncode == 0 else None
+
+
+def setfattr(path: Path, name: str, value: str) -> None:
+    run(["setfattr", "--name", f"user.{name}", "--value", value, path])
```
keszybz, line 1061:
```python
os.setxattr(path, f"user.{name}", value)
```
?

keszybz (commented):
Looks reasonable.

mkosi/distributions/arch.py
```diff
@@ -1,17 +1,12 @@
...
 from collections.abc import Sequence
-from pathlib import Path
-from textwrap import dedent
 
 from mkosi.architecture import Architecture
-from mkosi.config import ConfigFeature
 from mkosi.distributions import DistributionInstaller
 from mkosi.log import die
-from mkosi.run import bwrap
+from mkosi.pacman import invoke_pacman, setup_pacman
```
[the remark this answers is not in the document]
DaanDeMeyer: It's not only used in mkosi/distributions though? In a later commit we start using it directly in the main file. So I'd prefer to keep it out of mkosi/distributions

mkosi/__init__.py
```diff
@@ -86,114 +79,28 @@ def mount_image(state: MkosiState) -> Iterator[None]:
...
-
 def clean_package_manager_metadata(state: MkosiState) -> None:
-    """Remove package manager metadata
+    """
+    Remove package manager metadata
 
-    Try them all regardless of the distro: metadata is only removed if the
-    package manager is present in the image.
+    Try them all regardless of the distro: metadata is only removed if the package manager is present in the
+    image.
```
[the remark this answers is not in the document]
DaanDeMeyer: Having a different column width for text is just annoying. We don't do that in systemd either.

behrmann:
> The interface is very loose, also, I'm not sure what that submodule should be named. package_manager seems way too long.

Maybe going with other uses we already have `pkgmngr` or just `packages`?

DaanDeMeyer:
Still need to write docs but I started with a commit to fix large parts of the current documentation, which was horribly out of date.

DaanDeMeyer:
> Maybe going with other uses we already have pkgmngr or just packages?

I went for "installer" as these are tools that install packages.

keszybz:
```
‣  Building initrd
‣ 'x86-64' is not a valid Architecture
  File "/usr/local/lib/python3.10/dist-packages/mkosi/config.py", line 322, in parse_enum
    return type[value]
  File "/usr/lib/python3.10/enum.py", line 440, in __getitem__
    return cls._member_map_[name]
KeyError: 'x86-64'
```

CI is unhappy.

keszybz (changes requested):
Oops, I made some comments but then didn't submit the review. I hate when that happens.

Anyway, needs a rebase now.

mkosi.md
```diff
@@ -10,7 +10,7 @@ mkosi — Build Bespoke OS Images
 
 `mkosi [options…] summary`
 
-`mkosi [options…] build [script parameters…]`
+`mkosi [options…] build`
```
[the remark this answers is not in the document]
DaanDeMeyer: Yes we removed it again. I don't think it was the right way to go in the first place

mkosi/run.py
```diff
@@ -381,7 +367,10 @@ def chroot_cmd(root: Path, *, options: Sequence[PathString] = (), network: bool
     else:
         cmdline += ["--unshare-net"]
 
-    return cmdline
+    # No exec here because we need to clean up the /work directory afterwards.
+    cmdline += ["sh", "-c", f"$0 \"$@\" && rm -rf {root / 'work'} || exit $?"]
```
behrmann, line 371: Isn't the `|| exit $?` superfluous here, i.e. wouldn't
```suggestion
    cmdline += ["sh", "-c", f"$0 \"$@\" && rm -rf {root / 'work'}"]
```
do the same thing (as long as there is no command afterwards)? If the command before the `&&` exits nonzero this will trigger the `||` and exit with the exit code code the command before the `&&`, but without the `||`, the right side of `&&` will not be triggered either and the compound command will have the return value of left hand side (i.e. the command before `&&`), no?

mkosi/run.py
```diff
@@ -343,19 +299,49 @@ def bwrap(
...
+            cmdline += ["--bind", p, f"/etc/{f}"]
+        else:
+            cmdline += ["--bind", "/dev/null", f"/etc/{f}"]
+
+    chmod = f"chmod 1777 {root / 'tmp'} {root / 'var/tmp'} {root / 'dev/shm'}"
+    # Make sure anything running in the root directory thinks it's in a container. $container can't always be
+    # accessed so we write /run/host/container-manager as well which is always accessible.
+    container = f"mkdir {root}/run/host && echo mkosi > {root}/run/host/container-manager"
+
+    cmdline += ["sh", "-c", f"{chmod} && {container} && exec $0 \"$@\" || exit $?"]
```
behrmann, line 339: same here

mkosi/run.py
```diff
@@ -381,7 +367,10 @@ def chroot_cmd(root: Path, *, options: Sequence[PathString] = (), network: bool
     else:
         cmdline += ["--unshare-net"]
 
-    return cmdline
+    # No exec here because we need to clean up the /work directory afterwards.
+    cmdline += ["sh", "-c", f"$0 \"$@\" && rm -rf {root / 'work'} || exit $?"]
+
+    return apivfs_cmd(root) + cmdline
```
behrmann, line 373: So this pastes together something like `bwrap [options] sh -c "stuff" sh -c "stuff"`? I'm surprised that works without anything to combine the two.

mkosi/run.py
```diff
@@ -343,19 +299,49 @@ def bwrap(
...
+        p = root / "etc" / f
+        if p.exists():
+            cmdline += ["--bind", p, f"/etc/{f}"]
+        else:
+            cmdline += ["--bind", "/dev/null", f"/etc/{f}"]
+
+    chmod = f"chmod 1777 {root / 'tmp'} {root / 'var/tmp'} {root / 'dev/shm'}"
+    # Make sure anything running in the root directory thinks it's in a container. $container can't always be
+    # accessed so we write /run/host/container-manager as well which is always accessible.
+    container = f"mkdir {root}/run/host && echo mkosi > {root}/run/host/container-manager"
```
behrmann, line 337: Nit
```suggestion
    container = f"mkdir {root}/run/host && echo mkosi >{root}/run/host/container-manager"
```
for the same style as the other shell stuff in systemd

mkosi/run.py
DaanDeMeyer: The latter one just becomes an argument to the earlier one. Works great, allows you to do arbitrary setup without losing the interface where you just string command lines together.

DaanDeMeyer: From the bash man page:

```
       -c        If the -c option is present, then commands are read from the first non-option argument command_string.  If there are arguments after the command_string, the first argument is assigned to $0 and any remaining argu‐
                 ments are assigned to the positional parameters.  The assignment to $0 sets the name of the shell, which is used in warning and error messages.
```

NEWS.md
```diff
@@ -87,6 +87,15 @@
   this doesn't work properly as it might result in leftover files in the install directory from a previous
   installation, so we have to empty the directory before reusing it, invalidating the caching, so the option
   was removed.
+- Build scripts are now executed on the host. See the `SCRIPTS` section
+  in the manual for more information. To run a script in the image, put
+  the following snippet at the top of the script:
+
+  ```sh
+  if [ ! "$container" = "mkosi" ]; then
+    exec chroot $SCRIPT "$@"
```
behrmann, line 96: I'm with @keszybz on calling this something else, e.g. as he proposed `_chroot` or maybe `mkosi-chroot` to be clear this is us.

mkosi.md
```diff
@@ -1082,6 +1056,123 @@ images. Any distribution that packages `zypper` may be used to build
...
+  scripts and tools executed inside the chroot work properly. It also
+  propagates `/etc/resolv.conf` from the host into the chroot if
+  requested so that DNS resolution works inside the chroot. After the
+  chroot command exits, various mount points are cleaned up.
+
+  To execute the entire script inside the image, put the following
+  snippet at the start of the script:
+
+  ```sh
+  if [ ! "$container" = "mkosi" ]; then
```
behrmann, line 1164:
```suggestion
  if [ "$container" != "mkosi" ]; then
```
Let's not make POSIX shell even worse than it is :)

NEWS.md
```diff
@@ -87,6 +87,15 @@
   this doesn't work properly as it might result in leftover files in the install directory from a previous
   installation, so we have to empty the directory before reusing it, invalidating the caching, so the option
   was removed.
+- Build scripts are now executed on the host. See the `SCRIPTS` section
+  in the manual for more information. To run a script in the image, put
+  the following snippet at the top of the script:
+
+  ```sh
+  if [ ! "$container" = "mkosi" ]; then
```
behrmann, line 95:
```suggestion
  if [ "$container" != "mkosi" ]; then
```
same here

mkosi.md
```diff
@@ -1082,6 +1056,123 @@ images. Any distribution that packages `zypper` may be used to build
 
 Currently, *Fedora Linux* packages all relevant tools as of Fedora 28.
 
+# SCRIPTS
```
behrmann, line 1059: I think it would be helpful to move the execution flow section down here, so people can easily compare it with the scripts.

mkosi/run.py
behrmann: Makes sense, neat! A little comment explaining this would be good, though, since this does rely on what `apivfs_cmd` is to be of the form `sh -c "$0 \"$@\""` then.

mkosi/installer/apt.py
```diff
@@ -103,6 +103,5 @@ def invoke_apt(
     packages: Sequence[str] = (),
     apivfs: bool = True,
 ) -> None:
-    bwrap(apt_cmd(state, command) + [operation, *packages],
-          apivfs=state.root if apivfs else None,
-          env=state.config.environment)
+    cmd = apivfs_cmd(state.root) if apivfs else []
+    bwrap(cmd + apt_cmd(state, command) + [operation, *packages], env=state.config.environment)
```
behrmann, line 107: We're sorting package names for dnf and pacman, but not for apt and zypper, I think we should too.

mkosi/installer/zypper.py
```diff
@@ -67,8 +67,7 @@ def invoke_zypper(
     options: Sequence[str] = (),
     apivfs: bool = True,
 ) -> None:
-    bwrap(zypper_cmd(state) + [verb, *packages, *options],
-          apivfs=state.root if apivfs else None,
-          env=state.config.environment)
+    cmd = apivfs_cmd(state.root) if apivfs else []
+    bwrap(cmd + zypper_cmd(state) + [verb, *packages, *options], env=state.config.environment)
```
behrmann, line 71: see above.

mkosi/distributions/gentoo.py
```diff
@@ -34,10 +40,8 @@ def invoke_emerge(state: MkosiState, packages: Sequence[str] = (), apivfs: bool
             f"--root={state.root}",
             *packages,
```
behrmann, line 41: Here then too, probably.

mkosi.md
```diff
@@ -1320,6 +1340,11 @@ BuildPackages=make,gcc,libcurl-devel
 EOF
 $ cat >mkosi.build <<EOF
 #!/bin/sh
+
+if [ ! "$container" = "mkosi" ]; then
```
behrmann, line 1344:
```suggestion
if [ "$container" != "mkosi" ]; then
```
here, too.

DaanDeMeyer:
Renamed script to mkosi-chroot to distinguish it from regular chroot

behrmann (approved):

bluca (approved):

NEWS.md
```diff
@@ -87,6 +87,15 @@
   this doesn't work properly as it might result in leftover files in the install directory from a previous
   installation, so we have to empty the directory before reusing it, invalidating the caching, so the option
   was removed.
+- Build scripts are now executed on the host. See the `SCRIPTS` section
```
bluca, line 90: This description should be beefed up, saying that existing scripts need to be evaluated to ensure what they do is safe, and saying that as a safeguard the filesystem is mostly r/o. Yes it's in the manpage too, but given the potential for damage, it's better to be repetitive I think

DaanDeMeyer:
Updated snippet, just docs and CI already passed so merging

DaanDeMeyer:
> Hmm, does this make sense? Wouldnt it make more sense to simply add a second set of scripts so that one set is run outside and one inside? i.e. mkosi.build and mkosi.build.outside?

> Just breaking everything for little gain and forcing churn on any sounds worse to me.

Everything is already broken since the last release anyway, this change is pretty tame in comparison to all the rest we've broken, so I don't think it matters too much. And the migration path is trivial, you just need to add 3 lines to your script to have it running in the image again. I prefer doing things as right as we can now so we don't have to break compat like this again later on.

PR #1716, opened 2023-07-31
Fix filelists on fedora

DaanDeMeyer:
We should only download these on older Fedora's, not newer ones.


bluca (approved):

behrmann:
`filelists` is only used together with `dnf5` and the [linked bug in the comment](https://bugzilla.redhat.com/show_bug.cgi?id=2180842) is only assigned but not resolved. Is dnf5 with older Fedora a relevant thing and is this not needed anymore with dnf5 on Fedora 38 or newer? If that's the case the comment should also get a todo to remove this, once these Fedora versions go EOL.

PR #1717, opened 2023-07-31
apt: Make invoke_apt() and apt_cmd() slightly more generic

DaanDeMeyer:
Let's allow passing in the apt command we want to execute.

Split out of #1701


DaanDeMeyer:
Absolutely trivial, so let's merge

PR #1718, opened 2023-07-31
Add source and debuginfo repositories

DaanDeMeyer:
Split out of #1701


bluca:
Debian and Ubuntu also have separate debug symbols repositories, eg: `deb https://deb.debian.org/debian-debug/ bookworm-debug main`

mkosi/distributions/debian.py
```diff
@@ -20,22 +20,38 @@ def filesystem(cls) -> str:
...
-            return [f"deb [trusted=yes] {state.config.local_mirror} {state.config.release} {repos}"]
+            return [f"deb [trusted=yes] {state.config.local_mirror} {state.config.release} {components}"]
+
+        repos = [
+            f"{archive} {state.config.mirror} {state.config.release} {components}"
+            for archive in archives
+        ]
 
-        main = f"deb {state.config.mirror} {state.config.release} {repos}"
+        # debian debug repositories are separate mirrors so we have to muck around with the URL to make this
```
bluca, line 36: It can't be that easy, can it now :-) the debug repositories are typically not mirrored either, so you can treat them like the security one

DaanDeMeyer: Ah lovely that makes things easier

bluca: That's the 'local' case, which is different

DaanDeMeyer: That's not local right? local mirror is when it's on the same machine.

I found the following in debian docs though:

> Is a mirror the right choice? If you primarily want to support users at your ISP/facility, then maybe a caching proxy such as apt-cacher-ng, squid, or varnish might be the better choice.

So let's stick with this for now.

bluca: > That's not local right? local mirror is when it's on the same machine.
> 
> I found the following in debian docs though:
> 
> > Is a mirror the right choice? If you primarily want to support users at your ISP/facility, then maybe a caching proxy such as apt-cacher-ng, squid, or varnish might be the better choice.
> 
> So let's stick with this for now.

No, this doesn't work - the debug archive is not mirrored, so it would just result in a bunch of timeouts and failed apt update runs

bluca: ah sorry, misunderstood - yes this is the right thing, I think -debug is not mirrored intentionally, because it takes tons of space and it's disabled by default, and nothing gets installed from it unless specifically configured and selected

bluca (approved):

keszybz:
The non-Debian parts LGTM.

PR #1719, opened 2023-07-31
Distribution cleanups

DaanDeMeyer:
Split out of #1701


keszybz:
I reviewed this before, so I did just a quick scan now. Still LGTM.

PR #1720, opened 2023-08-01
Use repart definition files from image if none are provided explicitly

DaanDeMeyer:


behrmann (approved):
And there goes the 254 lower bound :)

bluca:
> > And there goes the 254 lower bound :)
> 
> Just a little. If you don't ship repart definition files in the image itself it all still works with v254

This should be documented clearly

DaanDeMeyer:
Closing as this doesn't work properly at all atm and needs a lot more work

PR #1723, opened 2023-08-02
Stuff

DaanDeMeyer:


DaanDeMeyer:
Most of this was already reviewed as part of #1701 so merging

mkosi/util.py
```diff
@@ -197,3 +198,18 @@ def scopedenv(env: Mapping[str, Any]) -> Iterator[None]:
     finally:
         os.environ = old
         tempfile.tempdir = None
+
+
+class StrEnum(enum.Enum):
```
behrmann, line 203: Why not backport the whole `StrEnum` verbatim from 3.11? There's not a much of a difference, but it would make sure that our `StrEnum` is the `StrEnum` people expect, when they see the name.

PR #1724, opened 2023-08-02
Make sure outputs are always prefixed with the output name

DaanDeMeyer:
Our output unlinking logic currently works based off the name of the output. Everything that is prefixed with the output name is removed. We need to fix that properly, but for now, let's make sure that all outputs that we generate are prefixed with the name of the output.


bluca (approved):

DaanDeMeyer:
CI failures unrelated

PR #1726, opened 2023-08-03
Fix shellcheck warning in mkosi-chroot examples

DaanDeMeyer:


behrmann (approved):
Can we also bump the indentation to four spaces? That would be consistent with the stuff shipped by systemd.

PR #1727, opened 2023-08-03
Fix variable referenced but not set shellcheck warning in examples

DaanDeMeyer:


behrmann:
Not quite sure about this change. I think shellcheck complains because the variable is lower case and doesn't know where it comes from, since it only assumes for all caps variables that they come from the environment. This is maybe rather something for a shellcheck disable directive and maybe also something to document where this variable comes from.

DaanDeMeyer:
I'm not sure, I think its good style to assume that $container might not be set. In fact, it won't be set when initially running the script, so acknowledging that and falling back to a default empty value seems right to me

behrmann:
That's true generally for variables under the control of the script, but for stuff coming from the outside a script, one can't really expect whether something is set or not - that's why shellcheck makes this difference on capitalisation. As long as people don't set `set -u` (something people tend to cargo cult nowadays), the unset variable will produce an empty string in the test, so adding a default just adds more noise without more signal. Ideally `$container` would be `$CONTAINER`.

PR #1731, opened 2023-08-04
opensuse: Install distribution-release by default

DaanDeMeyer:
This is the package providing /usr/lib/os-release so let's make sure it's installed by default.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -209,6 +209,7 @@ def run_prepare_script(state: MkosiState, build: bool) -> None:
     env = dict(
         SCRIPT="/work/prepare",
         SRCDIR=str(Path.cwd()),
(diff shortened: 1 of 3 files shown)


behrmann (approved):

mkosi/__init__.py
```diff
@@ -1792,9 +1804,13 @@ def build_image(args: MkosiArgs, config: MkosiConfig) -> None:
 
             install_build_dest(state)
             install_extra_trees(state)
-            install_boot_loader(state)
-            configure_ssh(state)
             run_postinst_script(state)
+
+            configure_autologin(state)
+            configure_initrd(state)
+            configure_ssh(state)
```
behrmann, line 1811: These should probably also be mentioned in the execution flow.

behrmann (approved):

PR #1732, opened 2023-08-04
Don't log exceptions from ssh/nspawn/qemu in --debug mode

DaanDeMeyer:
While we generally want full stacktraces in debug mode from every exception, let's make an exception for ssh/systemd-nspawn/qemu, since these exceptions are expected.


behrmann:
It's only these three points, but wouldn't it make be better to integrate into `run`?

DaanDeMeyer:
Reworked to something entirely different

bluca (approved):

PR #1735, opened 2023-08-05
Move summary() to config.py

DaanDeMeyer:
Also removes check_script_input() and its usages since we already check scripts during config parsing.


DaanDeMeyer:
This just moves code around with minor changes so merging

PR #1736, opened 2023-08-05
Rework tar archive handling

DaanDeMeyer:
- Instead of relying on shutil.unpack_archive(), let's always use tar
- Introduce archive_tree() and extract_tree() to abstract tar archives
- Make sure tar always uses the user/group information from the root dir
- Enable all features
- Make sure tar doesn't overwrite directory permissions

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -103,6 +103,9 @@
       exec mkosi-chroot "$SCRIPT" "$@"
   fi
   ```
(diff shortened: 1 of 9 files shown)


behrmann (approved):
LGTM, I'm not sure though whether the option should stay as a rump (I'm leaning towards not to).

DaanDeMeyer:
This is one of the super niche ones, so I don't think we have to keep it, will add a note to news though

PR #1738, opened 2023-08-05
tree: Introduce install_tree()

DaanDeMeyer:


mkosi/tree.py
```diff
@@ -160,7 +158,8 @@ def install_tree(config: MkosiConfig, src: Path, dst: Path, target: Optional[Pat
     if target:
         t = dst / target.relative_to("/")
 
-    t.parent.mkdir(mode=0o755, parents=True, exist_ok=True)
+    with umask(0o755):
+        t.parent.mkdir(parents=True, exist_ok=True)
```
behrmann, line 162: Going from mode `755` to umask `755` doesn't seem right? With a default umask of `022` a call 
```python
with umask(0o755):
    Path("bar/baz").mkdir(parents=True, exist_ok=True)
```
throws an error for me, because `baz` can't be created and `bar` has permissoins
```
d----w--w- 1 user group 0 Aug  5 21:07 bar
```

behrmann (commented):
The `install_tree` commit seems fine, but I'm not sure about the umask.

mkosi/tree.py
DaanDeMeyer: Right I forgot that umask takes a mask and not the mode, so we have to take the inverse of the mode. I fixed that, and made the mode argument keyword only to clearly indicate that we're passing a mode to the umask function and not a mask.

DaanDeMeyer: Reworked again to just always use the inverse operator when passing modes to umask(), so it's clear that we're converting them to a mask.

behrmann: Mhm, maybe adding a
```python
@contextlib.contextmanager()
def default_mode(mode: int):
    with umask(~mode):
        yield
```
would make this more readable.

PR #1741, opened 2023-08-05
Get rid of scandir_recursive() and find_files()

DaanDeMeyer:
- Use rglob() from pathlib instead of scandir_recursive()
- Use absolute paths when dealing with kernel modules instead of relative paths. Only make the paths relative in make_cpio()

### DIFF
--- a/mkosi/mounts.py
+++ b/mkosi/mounts.py
@@ -1,6 +1,5 @@
 # SPDX-License-Identifier: LGPL-2.1+
 
-import collections
 import contextlib
 import os
 import platform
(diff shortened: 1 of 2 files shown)


mkosi/mounts.py
```diff
@@ -46,9 +29,9 @@ def delete_whiteout_files(path: Path) -> None:
     """
 
     with complete_step("Removing overlay whiteout files…"):
-        for entry in cast(Iterator[os.DirEntry[str]], scandir_recursive(path)):
-            if stat_is_whiteout(entry.stat(follow_symlinks=False)):
-                os.unlink(entry.path)
+        for entry in path.rglob("*"):
+            if stat_is_whiteout(os.stat(entry, follow_symlinks=False)):
```
behrmann, line 33: Let's add a todo to do `entry.stat()` once we have Python 3.10.

behrmann (commented):
LGTM, the removal of `scandir_recursive` could degrade performance a bit, let's see if anybody complains.

PR #1743, opened 2023-08-06
Get rid of final uses of joinpath()

DaanDeMeyer:
We prefer using parentheses instead of joinpath() these days. This commit gets rid of the remaining places where we're still using joinpath().


DaanDeMeyer:
This one is trivial refactoring so merging

PR #1745, opened 2023-08-06
Split out more modules

DaanDeMeyer:


behrmann:
LGTM, I think, if CI's happy it's probably fine. The kernel module commit would be have been easier to review if the move and rework would have been separate.

PR #1746, opened 2023-08-06
Various cleanups

DaanDeMeyer:


behrmann (approved):
LGTM, though I think the module for `GenericVersion` would be better as `versioncomp` instead of `version`, so as not to be confused from where the version of mkosi is read.

bluca (approved):

PR #1747, opened 2023-08-06
Rework check_inputs()

DaanDeMeyer:
Let's just check for existence with exists() instead of relying on exceptions. Also improve error checking and make sure we check package manager trees as well.


behrmann (approved):
Though the change of the input has a better error message now on missing files, permissions errors will probably be a bit worse off.

PR #1748, opened 2023-08-07
qemu: do not enable kvm when booting on non-native arch

bluca:
kvm is only supported on the native architecture, so skip it when booting a non-native one (eg: aarch64 on x86_64)

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -207,7 +207,7 @@ def copy_ephemeral(config: MkosiConfig, src: Path) -> Iterator[Path]:
 
 def run_qemu(args: MkosiArgs, config: MkosiConfig) -> None:
     accel = "tcg"
-    if config.qemu_kvm == ConfigFeature.enabled or (config.qemu_kvm == ConfigFeature.auto and qemu_check_kvm_support()):
+    if config.qemu_kvm == ConfigFeature.enabled or (config.qemu_kvm == ConfigFeature.auto and config.architecture.is_native() and qemu_check_kvm_support()):
         accel = "kvm"
 
     firmware, fw_supports_sb = find_qemu_firmware(config)


DaanDeMeyer (changes requested):
I think we have an is_native() method on the architecture enum. Can you just use that?

bluca:
It doesn't seem to work? Doesn't return true on x86 for x86

bluca:
wait, something is going wrong somewhere else, hang on

bluca:
@DaanDeMeyer `qemu_check_kvm_support()` doesn't see /dev/kvm as r+b anymore? any idea? This is native-for-native, ie, no architecture option as confirmed by the summary, and I'm running mkosi with sudo `sudo bin/mkosi -d debian qemu`

bluca:
ie with this debug stuff:

```
diff --git a/mkosi/qemu.py b/mkosi/qemu.py
index eafe610..c78dcac 100644
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -207,6 +207,10 @@ def copy_ephemeral(config: MkosiConfig, src: Path) -> Iterator[Path]:
 
 def run_qemu(args: MkosiArgs, config: MkosiConfig) -> None:
     accel = "tcg"
+    print(config.architecture)
+    print(config.architecture.is_native())
+    print(qemu_check_kvm_support())
+    sys.exit(1)
     if config.qemu_kvm == ConfigFeature.enabled or (config.qemu_kvm == ConfigFeature.auto and config.architecture.is_native() and qemu_check_kvm_support()):
         accel = "kvm"
 
diff --git a/mkosi/util.py b/mkosi/util.py
index c1d8f66..102903b 100644
--- a/mkosi/util.py
+++ b/mkosi/util.py
@@ -123,13 +123,16 @@ def chdir(directory: Path) -> Iterator[None]:
 
 def qemu_check_kvm_support() -> bool:
     kvm = Path("/dev/kvm")
+    print(kvm.is_char_device())
     if not kvm.is_char_device():
         return False
     # some CI runners may present a non-working KVM device
     try:
         with kvm.open("r+b"):
+            print("yes")
             return True
     except OSError:
+        print("no")
         return False
 
 
```

I get:

```
$ sudo bin/mkosi -d debian qemu
x86-64
True
True
no
False
$ ls -l /dev/kvm 
crw-rw----+ 1 root kvm 10, 232 Aug  7 10:51 /dev/kvm
$ groups
luca adm cdrom floppy sudo dip video plugdev fuse scanner bluetooth netdev kvm sambashare libvirt wireshark steam
```

DaanDeMeyer:
Try with https://github.com/systemd/mkosi/pull/1750

DaanDeMeyer (approved):

PR #1764, opened 2023-08-08
Mark gentoo as experimental and unsupported

DaanDeMeyer:
Gentoo issues won't be actively looked at by the core maintainers and there's no guarantee that gentoo PRs will get reviewed or not (or they might just be merged without review). Gentoo specific hacks outside of gentoo.py will generally not be accepted.

- We also drop all gentoo specific modifications outside of gentoo.py
- We also stop running CI for gentoo

Conditions for dropping the experimental status of gentoo:

- No more stage 3 tarball required to build a gentoo image. Either we only do binary packages or gentoo images can be built from directly from an immutable host system or tools tree with portage and compiler tools installed
- Proper division of systemd components into subpackages so that we can build a gentoo initrd with systemd installed but without pulling in ukify and python
- An official and up-to-date binary package repository that allows gentoo CI to complete in a similar amount of time as the other distributions


bluca (approved):

DaanDeMeyer:
cc @257. I expect you won't agree with this resolution but gentoo is just too difficult to deal with for us to officially support, I'll merge whatever PRs you make for gentoo.py but we won't do CI anymore and we won't accept any gentoo specific modifications outside of gentoo.py. And we won't make any guarantees that we keep gentoo working when we make changes.

mkosi.md
```diff
@@ -1010,7 +1010,9 @@ distributions:
 
 * *Alma Linux*
 
-* *Gentoo*
+* *Gentoo* (**Gentoo support is experimental and unsupported. We make
```
behrmann, line 1013: "Gentoo support is [..] unsupported" sounds a bit weird.

behrmann (approved):

PR #1765, opened 2023-08-08
Run finalize script after selinux relabel

DaanDeMeyer:
Let's make the finalize script true to its name and run it as the final operation before we start packaging things up.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -1081,8 +1081,8 @@ Then, for each preset, we execute the following steps:
 * Run `systemd-firstboot`
 * Run `systemd-hwdb`
 * Remove packages and files (`RemovePackages=`, `RemoveFiles=`)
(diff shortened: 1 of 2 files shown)


keszybz:
I guess this could cause problems when the finalize scripts mucks up the selinux context and people rely on the relabel to fix things up. Is there some particular reason to change this?

DaanDeMeyer:
I'm changing this now because we're going to do a release and we can't change it again later. You're always going to run into the stuff you describe. The finalize script might add more presets but we've already run systemctl preset-all, the finalize script might add more sysusers files but we've already executed sysusers, etc etc. I'm fine with saying that you have to make sure the labels are correct yourself if you modify stuff in the finalize script when selinux is enabled

DaanDeMeyer:
It's clearly marked in the execution flow which I think is fine

PR #1766, opened 2023-08-08
Introduce preset dependencies

daandemeyer:
Instead of building presets in alphanumerical order, let's introduce
a proper concept of dependencies. Dependencies are configured with
the new "Dependencies=" option in the new "[Preset]" section. All
presets configured with "Dependencies=" are built before the current
preset.

We drop the alphanumerical preset in favor of explicit dependencies.

### DIFF
--- a/mkosi.md
+++ b/mkosi.md
@@ -190,9 +190,10 @@ Those settings cannot be configured in the configuration files.
 
 `--preset=`
 
-: If specified, only build the given preset. Can be specified multiple
-  times to build multiple presets. If not specified, all presets are
-  built. See the `Presets` section for more information.
+: If specified, only build the given presets. Can be specified multiple
+  times to build multiple presets. All the given presets and their
+  dependencies are built. If not specified, all presets are built. See
+  the `Presets` section for more information.
 
 ## Supported output formats
 
@@ -299,6 +300,15 @@ they should be specified with a boolean argument: either "1", "yes", or "true" t
 | `ImageId=`      | yes   | no               | match fails             |
 | `ImageVersion=` | no    | yes              | match fails             |
 | `Bootable=`     | no    | no               | match auto feature      |
+
+### [Preset] Section
+
+`Dependencies=`, `--dependency=`
+
+: The presets that this preset depends on specified as a comma-separated
+  list. All presets configured in this option will be built before this
+  preset and will be pulled in as dependencies of this preset when
+  `--preset` is used.
 
 ### [Distribution] Section
 
@@ -1337,22 +1347,18 @@ directories containing mkosi configuration files or regular files with
 the `.conf` extension.
 
 When presets are found in `mkosi.presets/`, mkosi will build the
-configured presets (or all of them if none were explicitly configured
-using `--preset=`) in alphanumerical order. To enforce a certain build
-order, preset names can be numerically prefixed (e.g. `00-initrd.conf`).
-The numerical prefix will be removed from the preset name during parsing,
-along with the `.conf` suffix (`00-initrd.conf` becomes `initrd`). The
-preset name is used for display purposes and also as the default output
-name if none is explicitly configured.
+configured preset and its dependencies (or all of them if no presets
+were explicitly configured using `--preset=`). To add dependencies
+between presets, the `Dependencies=` setting can be used.
 
 When presets are defined, mkosi will first read the global configuration
 (configuration outside of the `mkosi.presets/` directory), followed by
 the preset specific configuration. This means that global configuration
 takes precedence over preset specific configuration.
 
-Later presets can refer to outputs of earlier presets. Specifically, for
-the following options, mkosi will only check whether the inputs exist
-just before building the preset:
+Presets can refer to outputs of presets they depend on. Specifically,
+for the following options, mkosi will only check whether the inputs
+exist just before building the preset:
 
 - `BaseTrees=`
 - `PackageManagerTrees=`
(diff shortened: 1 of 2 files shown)


mkosi.md
```diff
@@ -300,6 +301,14 @@ they should be specified with a boolean argument: either "1", "yes", or "true" t
 | `ImageVersion=` | no    | yes              | match fails             |
 | `Bootable=`     | no    | no               | match auto feature      |
 
+### [Preset] Section
+
+`Depends=`, `--depends=`
+
+: The presets that this preset depends on. All presets configured in
```
behrmann, line 308:
```suggestion
: The presets that this preset depends on as a comma-seperated list. All presets configured in
```

mkosi/config.py
```diff
@@ -1765,6 +1773,36 @@ def backward_compat_stubs(self, namespace: argparse.Namespace) -> None:
             delattr(namespace, "cache")
             print("Warning: --cache is no longer supported")
 
+    def resolve_deps(self, args: MkosiArgs, presets: Sequence[MkosiConfig]) -> list[MkosiConfig]:
+        graph = {p.preset: p.depends for p in presets}
+
+        if args.presets:
+            for p in args.presets:
+                if p not in graph:
+                    die(f"No preset found with name {p}")
```
behrmann, line 1782:
```suggestion
            if any(p not in graph for p in args.presets):
                die(f"No preset found with name {p}")
```

mkosi/config.py
```diff
@@ -1851,7 +1887,10 @@ def parse(self, argv: Optional[Sequence[str]] = None) -> tuple[MkosiArgs, tuple[
         # infrastructure scripts rather than image-specific configuration.
         self.backward_compat_stubs(namespace)
 
-        return args, tuple(load_config(ns) for ns in presets)
+        presets = list(load_config(ns) for ns in presets)
```
behrmann, line 1890:
```suggestion
        presets = [load_config(ns) for ns in presets]
```
Just more readily identifiable as a list.

mkosi.md
```diff
@@ -300,6 +301,14 @@ they should be specified with a boolean argument: either "1", "yes", or "true" t
 | `ImageVersion=` | no    | yes              | match fails             |
 | `Bootable=`     | no    | no               | match auto feature      |
 
+### [Preset] Section
+
+`Depends=`, `--depends=`
```
keszybz, line 306: Wouldn't it be better to call this After? "Depends" doesn't sound good because "on" is missing, i.e we say "a depends on b", not "a depends b".

mkosi/config.py
```diff
@@ -1765,6 +1773,36 @@ def backward_compat_stubs(self, namespace: argparse.Namespace) -> None:
...
+            deps = set()
+            queue = [*args.presets]
+
+            while queue:
+                preset = queue.pop()
+                if preset in deps:
+                    continue
+
+                deps.add(preset)
+                queue += [*graph[preset]]
```
behrmann, line 1793: Maybe a bit pedantic, but it's a stack, not a queue. `list.pop()` will pop the last element (on the right side) and this will add new elements on the right. 

Also, with a few less sigils this is
```suggestion
                queue.extend(graph[preset])
```
but generally this would probably be better with a `collections.deque` for `queue`, which could do a `popleft` above then.

mkosi/config.py
```diff
@@ -1765,6 +1773,36 @@ def backward_compat_stubs(self, namespace: argparse.Namespace) -> None:
...
+
+            presets = [p for p in presets if p.preset in deps]
+
+        graph = {p.preset: p.depends for p in presets}
+
+        try:
+            order = list(graphlib.TopologicalSorter(graph).static_order())
```
keszybz, line 1800: `graphlib`, nice.

mkosi/config.py
keszybz, line 1890: `[load_config(ns) for ns in presets]`

mkosi.md
keszybz: Or "Dependencies". But I think "After" is nicer.

mkosi/config.py
```diff
@@ -1765,6 +1773,36 @@ def backward_compat_stubs(self, namespace: argparse.Namespace) -> None:
...
+                preset = queue.pop()
+                if preset in deps:
+                    continue
+
+                deps.add(preset)
+                queue += [*graph[preset]]
+
+            presets = [p for p in presets if p.preset in deps]
+
+        graph = {p.preset: p.depends for p in presets}
```
behrmann, line 1797: So, this rebuilds `graph` with only the members of `presets` in `deps`, right? This should be equivalent to removing all nodes from `graph` that are not in `deps`, i.e.
```suggestion
        graph = {node: graph[node] for node in graph.keys() & deps}
```

This doesn't preserve the order of the keys in `graph`, but this doesn't matter, because the order will be changed by the topological sort right atferwards anyway.

DaanDeMeyer: Yes but we also need these removed from presets for the sort we do later

DaanDeMeyer: This isn't sufficient, we need to filter the preset list as well so we can sort it properly later.

mkosi.md
DaanDeMeyer: Went for `Dependencies=`

mkosi/config.py
behrmann: It works with the walrus operator, that I keep forgetting to add, though
```python
if any((miss := p) not in graph for p in arg.presets):
    die(f"No preset found with name {miss}")
```

mkosi/config.py
behrmann: Ah, true. Maybe it could be more succinct by using the presets themselves as keys in the graph instead of their names, but I think it's fine as is.

behrmann (approved):

PR #1769, opened 2023-08-09
Make sure we output signed kernel instead of unsigned kernel

### LINKED ISSUE #1589: Uki and `.vmlinuz` signatures differ
Following https://github.com/systemd/mkosi/commit/85d8342e974e5537529ea959b4e405fa330467b9, the signatures between the `.vmlinuz` output and the kernel image inside the UKI differ.
The kernel inside the UKI is signed by the mkosi secureboot keys, the `.vmlinuz` output is not.

Since only UEFI hosts are supported with systemd-boot, I don't see why the split out kernel is needed after all, but I guess it should then work the same way as the UKI.

DaanDeMeyer:
Fixes #1589

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -773,6 +773,20 @@ def install_unified_kernel(state: MkosiState, roothash: Optional[str]) -> None:
             if not (state.staging / state.config.output_split_uki).exists():
                 shutil.copy(boot_binary, state.staging / state.config.output_split_uki)
 
+                # ukify will have signed the kernel image as well. Let's make sure we put the signed kernel
+                # image in the output directory instead of the unsigned one by reading it from the UKI.
+
+                import pefile  # type: ignore
+                pe = pefile.PE(boot_binary, fast_load=True)
+                linux = {s.Name.decode().strip("\0"): s for s in pe.sections}[".linux"]
+                run(["dd",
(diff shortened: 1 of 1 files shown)


bluca (approved):

mkosi/__init__.py
```diff
@@ -773,6 +773,20 @@ def install_unified_kernel(state: MkosiState, roothash: Optional[str]) -> None:
             if not (state.staging / state.config.output_split_uki).exists():
                 shutil.copy(boot_binary, state.staging / state.config.output_split_uki)
 
+                # ukify will have signed the kernel image as well. Let's make sure we put the signed kernel
+                # image in the output directory instead of the unsigned one by reading it from the UKI.
+
+                import pefile  # type: ignore
+                pe = pefile.PE(boot_binary, fast_load=True)
+                linux = {s.Name.decode().strip("\0"): s for s in pe.sections}[".linux"]
+                run(["dd",
```
medhefgo, line 782: Why shell out to dd if you can get the data with `pe.get_data()` and let python write out the bytes?

keszybz: Yes, please, let's not shell out unless really useful.

keszybz: This was done in 9ffcdac128c66935aa5d5a98633fa7498bce92d1.

PR #1773, opened 2023-08-09
Replace dd with pefile's get_data() method

DaanDeMeyer:


mkosi/__init__.py
```diff
@@ -779,13 +779,7 @@ def install_unified_kernel(state: MkosiState, roothash: Optional[str]) -> None:
...
                 pe = pefile.PE(boot_binary, fast_load=True)
                 linux = {s.Name.decode().strip("\0"): s for s in pe.sections}[".linux"]
-                run(["dd",
-                     f"if={boot_binary}",
-                     f"of={state.staging / state.config.output_split_kernel}",
-                     f"skip={linux.PointerToRawData}",
-                     # Get the actual size using Misc_VirtualSize instead of the aligned size from SizeOfRawData.
-                     f"count={linux.Misc_VirtualSize}",
-                     "iflag=skip_bytes,count_bytes"])
+                (state.root / state.config.output_split_kernel).write_bytes(linux.get_data(length=linux.Misc_VirtualSize))
```
medhefgo, line 782: This should do:

```suggestion
                (state.root / state.config.output_split_kernel).write_bytes(linux.get_data(ignore_padding=True))
```

DaanDeMeyer: I had that first, it's too new, ignore_padding() is not available in Jammy's version of pefile

behrmann:
LGTM, but the `pefile` dependency slipped through in the move to `pyproject.toml`. That should be added as a dependency there, with the jammy version as lower bound.

PR #1779, opened 2023-08-10
Make sure the private tmpfs in bwrap() doesn't hide workspace

DaanDeMeyer:
The workspace might be in /tmp so we need to make sure we mount it
first before we mount the workspace so that the workspace is on top
of the private tmpfs.


bluca (approved):
works, ship it

bluca (approved):

PR #1781, opened 2023-08-10
Add script to prepare man pages, retitle release commits

keszybz:


DaanDeMeyer (approved):

mkosi/resources/.gitignore
```diff
@@ -0,0 +1 @@
+/mkosi.1
```
DaanDeMeyer, line 1: Let's just add this to the global gitignore? We don't want this in the repo regardless of the directory its in

DaanDeMeyer (changes requested):

DaanDeMeyer (approved):

PR #1784, opened 2023-08-10
Make sure we use python-pefile from the tools tree

DaanDeMeyer:


mkosi/__init__.py
```diff
@@ -778,11 +778,22 @@ def install_unified_kernel(state: MkosiState, roothash: Optional[str]) -> None:
...
+                    f"""\
+                    import pefile
+                    from pathlib import Path
+                    pe = pefile.PE("{boot_binary}", fast_load=True)
+                    linux = {{s.Name.decode().strip("\\0"): s for s in pe.sections}}[".linux"]
+                    (Path("{state.root}") / "{state.config.output_split_kernel}").write_bytes(linux.get_data(length=linux.Misc_VirtualSize))
+                    """
+                ).replace("\n", ";")
+
+                run(["python3", "-c", pefile])
```
behrmann, line 796: Maybe marginally less ugly, but I think you could not make this this a oneliner, but pass the script via stdin
```suggestion
                )

                run(["python3"], input=pefile)
```

behrmann (commented):
A tad ugly, but alas. In this case, we can remove `pefile` from `pyproject.toml` again?

DaanDeMeyer:
I would keep it in pyproject.toml. It's still needed if you're not using a tools tree. Although if we ever get to this point, we know pefile will be available since ukify needs it as well

behrmann (approved):

PR #1790, opened 2023-08-13
Various cleanups

DaanDeMeyer:


behrmann (approved):
LGTM, though I'm not enthusiastic about all the import changes, e.g. `datetime.datetime`, but that only has one or two usages, so I don't care all that much.

PR #1796, opened 2023-08-15
Reword systemd version requirement a bit

DaanDeMeyer:
We should use wording that still makes sense when reading documentation from the distro packaged version of mkosi as well.


behrmann:
Didn't we already get a dependency on a post-v254 feature in repart for bootable images?

DaanDeMeyer:
No that PR was closed

behrmann (approved):

PR #1797, opened 2023-08-15
Add back BIOS support using grub

DaanDeMeyer:
Add back BIOS support using grub
Let's add back support for booting on BIOS using grub. This comes
with the following limitations:

- grub does not support UKIs on BIOS, so we set up the individual
components instead
- grub cannot search partitions by PARTUUID, so we're forced to have
it find the ESP by partition number instead.

We opt to generate grub.cfg ourselves instead of relying on grub-mkconfig.
grub-mkconfig is basically like kernel-install but for grub configuration,
it has a ton of distro specific cruft that we want to ignore, so we simply
don't use it and generate the grub configuration ourselves.

To allow for unprivileged installation of grub, we can't use grub-install
as it insists on opening the root device and probing its filesystem, which
isn't possible unprivileged. Instead, we run grub-mkimage and grub-bios-setup
ourselves, and manually copy the required files to the ESP.

We use the ESP to store the kernels, initrds and grub config. In the event
that grub adds support for UKIs on BIOS in the future, we can simply modify
the generated grub configuration to use our generated UKIs instead.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -143,6 +143,11 @@ def install_distribution(state: MkosiState) -> None:
             with umask(~0o500):
                 (state.root / "efi").mkdir(exist_ok=True)
 
+            # Some distributions install EFI binaries directly to /boot/efi. Let's redirect them to /efi
+            # instead.
+            rmtree(state.root / "boot/efi")
+            (state.root / "boot/efi").symlink_to("../efi")
+
             if state.config.packages:
                 state.config.distribution.install_packages(state, state.config.packages)
 
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1374,14 +1374,10 @@ def make_image(state: MkosiState, skip: Sequence[str] = [], split: bool = False)
         definitions = state.workspace / "repart-definitions"
         if not definitions.exists():
             definitions.mkdir()
-            bootdir = state.root / "usr/lib/systemd/boot/efi"
+            bootloader = state.root / f"efi/EFI/BOOT/BOOT{state.config.architecture.to_efi()}.EFI"
 
-            # If Bootable=auto and we have at least one UKI and a bootloader, let's generate an ESP partition.
             add = (state.config.bootable == ConfigFeature.enabled or
-                  (state.config.bootable == ConfigFeature.auto and
-                   bootdir.exists() and
-                   any(bootdir.iterdir()) and
-                   any(gen_kernel_images(state))))
+                  (state.config.bootable == ConfigFeature.auto and bootloader.exists()))
 
             if add:
                 (definitions / "00-esp.conf").write_text(
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -471,13 +471,13 @@ def install_systemd_boot(state: MkosiState) -> None:
 
     if not shutil.which("bootctl"):
         if state.config.bootable == ConfigFeature.enabled:
-            die("A bootable image was requested but bootctl was not found")
+            die("An EFI bootable image with systemd-boot was requested but bootctl was not found")
         return
 
     directory = state.root / "usr/lib/systemd/boot/efi"
     if not directory.exists() or not any(directory.iterdir()):
         if state.config.bootable == ConfigFeature.enabled:
-            die("A bootable image was requested but systemd-boot was not found at "
+            die("A EFI bootable image with systemd-boot was requested but systemd-boot was not found at "
                 f"{directory.relative_to(state.root)}")
         return
 
@@ -725,7 +725,7 @@ def install_unified_kernel(state: MkosiState, partitions: Sequence[Partition]) -
     if state.config.bootable == ConfigFeature.disabled:
         return
 
-    if state.config.bootloader not in (Bootloader.systemd_boot, Bootloader.uki):
+    if state.config.bootloader == Bootloader.none:
         return
 
     for kver, kimg in gen_kernel_images(state):
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -581,6 +581,22 @@ def find_grub_prefix(state: MkosiState) -> Optional[str]:
     return "grub2" if "grub2" in os.fspath(path) else "grub"
 
 
+def want_grub_efi(state: MkosiState) -> bool:
+    if state.config.bootable == ConfigFeature.disabled:
+        return False
+
+    if state.config.bootloader != Bootloader.grub:
+        return False
+
+    if not any((state.root / "efi").rglob("grub*.efi")):
+        if state.config.bootable == ConfigFeature.enabled:
+            die("A bootable EFI image with grub was requested but grub for EFI is not installed in /efi")
+
+        return False
+
+    return True
+
+
 def want_grub_bios(state: MkosiState, partitions: Sequence[Partition] = ()) -> bool:
     if state.config.bootable == ConfigFeature.disabled:
         return False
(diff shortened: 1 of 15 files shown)


mkosi/__init__.py
```diff
@@ -1372,22 +1529,39 @@ def make_image(state: MkosiState, skip: Sequence[str] = [], split: bool = False)
...
                     textwrap.dedent(
                         """\
                         [Partition]
                         Type=esp
                         Format=vfat
                         CopyFiles=/efi:/
-                        SizeMinBytes=512M
-                        SizeMaxBytes=512M
+                        SizeMinBytes=1G
+                        SizeMaxBytes=1G
```
behrmann, line 1547: Bumping this for all cases seems a bit extreme?

mkosi/resources/mkosi.md
```diff
@@ -790,6 +790,18 @@ they should be specified with a boolean argument: either "1", "yes", or "true" t
   (the one with the highest version) and stored in
   `EFI/BOOT/BOOTX64.EFI`.
 
+`BiosBootloader=`, `--bios-bootloader`
+
+: Takes one of `none`, `auto` and `grub`. Defaults to `auto`. If set to
+  `none`, no BIOS bootloader will be installed. If set to `auto`, if
+  grub is installed and a BIOS boot partition is configured, grub will
```
behrmann, line 797: This should mention `21686148-6449-6e6f-744e-656564454649` (and maybe also mention that 1MB is enough) so that people know what this means, maybe we should also call this a _GRUB partition_ instead of a BIOS boot partition, so that it's really clear to users, that this is a GRUB thing and doesn't have anything to do with `/boot`.

mkosi/__init__.py
```diff
@@ -143,6 +144,11 @@ def install_distribution(state: MkosiState) -> None:
             with umask(~0o500):
                 (state.root / "efi").mkdir(exist_ok=True)
 
+            # Some distributions install EFI binaries directly to /boot/efi. Let's redirect them to /efi
+            # instead.
+            rmtree(state.root / "boot/efi")
+            (state.root / "boot/efi").symlink_to("../efi")
```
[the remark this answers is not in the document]
DaanDeMeyer: Well this is a can of worms.

So currently, when an ESP is added, systemd-repart will exclude the contents of /boot and /efi when populating the root partition. Also, when populating a VFAT partition, systemd-repart will skip any symlinks it finds. So if a user adds an XBOOTLDR, it will contain just the contents of /boot without the /boot/efi symlink. This will go horribly wrong when trying to update stuff via the package manager, but then again, that's already the case as it is

mkosi/__init__.py
```diff
@@ -543,6 +549,157 @@ def install_systemd_boot(state: MkosiState) -> None:
...
+        return
+
+    root = finalize_roothash(partitions)
+    if not root:
+        root = next((f"root=PARTUUID={p.uuid}" for p in partitions if p.type.startswith("root")), None)
+    if not root:
+        root = next((f"mount.usr=PARTUUID={p.uuid}" for p in partitions if p.type.startswith("usr")), None)
+
+    assert root
+    assert (install := find_grub_install(state))
```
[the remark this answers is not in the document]
DaanDeMeyer: Yes but we're not relying on the runtime behavior of assert here, I'm solely using it to refine the type because we know find_grub_install() will not return `None` here. So for that use it should be fine?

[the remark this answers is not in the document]
behrmann: I see. It definitely is more succinct, but it is a bit of a foot gun, because a stray `PYTHONOPTIMIZE=1` in the environment could remove the assert and then `install` would be unbound. It's more verbose, but
```python
install = find_grub_install(state)
if install is not None:
    die("Oh noes!")
```
will do the same type narrowing.

Unexpectedly, at least for me, since I thought the scoping rules would be different, even this seems to work:
```python
if (install := find_grub_install(state)) is not None:
    die("Oh noes!")
```

I just checked this for a smaller example, though:
```python
#!/usr/bin/python3
from typing import Optional

def foo() -> Optional[int]:
    return 5


def main():
    if (bar := foo()) is None:
        print("eh")
    print(bar)


if __name__ == "__main__":
    main()
```
which prints 5.

Anyway, in this case avoiding the foot gun is only one more line.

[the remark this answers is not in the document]
DaanDeMeyer: So I just moved the assert to the next line, that way the variable will always be bound. the assert is just there to tell the type checker that the type is not None. If it is, we'll just get an exception later and it'll be a bug for us to fix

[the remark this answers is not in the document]
behrmann: Yeah, that's fine.

mkosi/__init__.py
```diff
@@ -1371,23 +1557,42 @@ def make_image(state: MkosiState, skip: Sequence[str] = [], split: bool = False)
...
+            bios = (state.config.bootable != ConfigFeature.disabled and
+                    want_grub_bios(state) and
+                    any(gen_kernel_images(state)))
+
+            if bios:
+                (definitions / "05-bios.conf").write_text(
+                    textwrap.dedent(
+                        """\
+                        [Partition]
+                        Type=2168614864496E6F744E656564454649
```
behrmann, line 1570: Since they live in the same file, couldn't this be
```suggestion
                        Type={Partition.GRUB_BOOT_PARTITION_UUID}
```

behrmann (commented):
LGTM otherwise

mkosi/__init__.py
```diff
@@ -543,6 +551,184 @@ def install_systemd_boot(state: MkosiState) -> None:
...
+                        unprivileged containers to solve this problem.
+                        """
+                    )
+                )
+        if want:
+            if not control.exists():
+                logging.warning("Cannot install grub because /dev/loop-control is missing, skipping")
+            else:
+                logging.warning("Cannot install grub because mkosi is running in a user namespace, skipping")
+        return False
```
behrmann, line 620: I think this can be reordered like this to save one level of logic
```suggestion
    if not control.exists():
        if state.config.bootable == ConfigFeature.enabled:
            die("A BIOS bootable image with grub was requested but grub cannot be installed because /dev/loop-control is not available")
        elif want:
            logging.warning("Cannot install grub because /dev/loop-control is missing, skipping")
        return False
    elif control.stat().st_uid != os.getuid():
        if state.config.bootable == ConfigFeature.enabled:
            die(
                textwrap.dedent(
                    """\
                    A BIOS bootable image with grub was requested but grub cannot be installed because
                    mkosi is running in a user namespace. Run mkosi with root privileges outside of any
                    unprivileged containers to solve this problem.
                    """
                )
            )
        elif want:
            logging.warning("Cannot install grub because mkosi is running in a user namespace, skipping")
        return False
```

mkosi/__init__.py
```diff
@@ -543,6 +551,234 @@ def install_systemd_boot(state: MkosiState) -> None:
...
+    with umask(~0o700):
+        config.parent.mkdir(exist_ok=True)
+
+    # For some unknown reason, if we don't set the timeout to zero, grub never leaves its menu, so we default
+    # to a zero timeout, but only if the config file hasn't been provided by the user.
+    if not config.exists():
+        with umask(~0o600), config.open("w") as f:
+            f.write("set timeout=0\n")
+
+    # EFI grub shipped by distributions reads its configuration from /efi/<distribution>/grub.cfg so let's
```
behrmann, line 660:
```suggestion
    # EFI grub shipped by distributions reads its configuration from $ESP/EFI/<distribution>/grub.cfg so let's
```
would be more in line with the definition of `efi` below

mkosi/__init__.py
```diff
@@ -543,6 +551,234 @@ def install_systemd_boot(state: MkosiState) -> None:
...
+    # put a shim there to redirect to the actual configuration file.
+    efi = state.root / "efi/EFI" / state.config.distribution.name / "grub.cfg"
+    with umask(~0o700):
+        efi.parent.mkdir(parents=True, exist_ok=True)
+
+    efi.write_text(f"configfile /{prefix}/grub.cfg\n")
```
behrmann, line 666: This is a bit confusing, I assume that the root here actually refers to the ESP for GRUB, right?

DaanDeMeyer: Indeed

bluca:
I can sort-of see the reason for bios boot, but what's the reason to support grub on efi?

DaanDeMeyer:
The whole shim thing unfortunately, until they sign systemd-boot it makes sense to support grub so we can build images that don't need a users custom key to be enrolled in the firmware

behrmann:
Another reason I see is that it provides a path to walk away from GRUB. A lot of places still do GRUB and have been for decades getting them on the mkosi train will probably need some changes to peoples' workflows. Reducing the needed changes and allowing them to switch to sd-boot after doing other initial steps might be appreciated.

behrmann (approved):
The mountinfo thing is a really clever hack.

PR #1799, opened 2023-08-16
config: Load mkosi.version from presets

mcassaniti:
This change makes sure the version is loaded When the file mkosi.version is present in a preset directory. The order of preference from least to most preferred is:

* mkosi.version in the base directory
* configuration in the base directory
* configuration in the base drop-in directory
* mkosi.version in the preset directory
* configuration in the preset directory
* configuration in the preset drop-in directory

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -1602,6 +1602,10 @@ class MkosiConfigParser:
 
         parser.remove_section("Match")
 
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
See https://github.com/systemd/mkosi/pull/1800 which should be a better solution

PR #1803, opened 2023-08-16
Add Bootloader option

DaanDeMeyer:
For some use cases, we don't want systemd-boot but to boot directly from a UKI, so let's support that by bringing back the --bootloader option.


mkosi/resources/mkosi.md
```diff
@@ -780,6 +780,15 @@ they should be specified with a boolean argument: either "1", "yes", or "true" t
...
   will be added to the image if the disk output format is used.
 
+`Bootloader=`, `--bootloader=`
+
+: Takes one of `systemd-boot` or `uki`. If set to `systemd-boot`,
+  systemd-boot will be installed and for each installed kernel, a UKI
+  will be generated and stored in `EFI/Linux` in the ESP partition. If
+  set to `uki`, systemd-boot will not be installed and a single UKI will
+  be generated for the latest installed kernel (the one with the highest
+  version) and stored in `EFI/BOOT/BOOTX64.EFI`.
```
behrmann, line 790:
```suggestion
  version) and stored in `EFI/BOOT/BOOTX64.EFI`. Defaults to `systemd-boot`.
```

behrmann (approved):

PR #1808, opened 2023-08-21
prepare for more cache modes

mcassaniti:
This change doesn't really add much more than some documentation updates and a caching enum type, but it does pave the way for more caching options to be made available. The intent is to get consensus on how other caching modes will be handled, document the current mode with a view to extending and make sure we can easily extend caching modes.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,10 @@
 
 - `mkosi.version` is now picked up from preset and dropin directories as
   well following the usual config precedence logic
+- The option `--incremental` is now deprecated. It will be removed in a
+  future version. Use `--cache-mode=prepare` instead.
+- A new option `--cache-mode` is now available which supports more caching
+  between build runs. See the documentation for further details.
 
 ## v15.1
 
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -24,7 +24,7 @@ from typing import Any, Callable, Optional, Type, Union, cast
 
 from mkosi.architecture import Architecture
 from mkosi.distributions import Distribution, detect_distribution
-from mkosi.log import ARG_DEBUG, ARG_DEBUG_SHELL, Style, die
+from mkosi.log import ARG_DEBUG, ARG_DEBUG_SHELL, Style, die, warn
 from mkosi.pager import page
 from mkosi.run import run
 from mkosi.types import PathString
@@ -118,6 +118,11 @@ class DocFormat(StrEnum):
 class Bootloader(StrEnum):
     uki          = enum.auto()
     systemd_boot = enum.auto()
+
+
+class CacheMode(StrEnum):
+    none    = enum.auto()
+    prepare = enum.auto()
 
 
 def parse_boolean(s: str) -> bool:
@@ -661,6 +666,7 @@ class MkosiConfig:
     output_dir: Path
     workspace_dir: Path
     cache_dir: Optional[Path]
+    cache_mode: CacheMode
     build_dir: Optional[Path]
     image_id: Optional[str]
     image_version: Optional[str]
@@ -727,7 +733,6 @@ class MkosiConfig:
     sign: bool
     key: Optional[str]
 
-    incremental: bool
     nspawn_settings: Optional[Path]
     extra_search_paths: list[Path]
     ephemeral: bool
@@ -818,6 +823,7 @@ class MkosiConfig:
             "packages": self.packages,
             "build_packages": self.build_packages,
             "repositories": self.repositories,
+            "cache_mode": self.cache_mode.name,
         }
 
         if self.prepare_script:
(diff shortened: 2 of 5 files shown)


DaanDeMeyer:
So let's not introduce unnecessary churn by renaming --incremental. Instead, we should just add config_parse_incremental() that first tries to parse the given value as one of the values from the enum. And if that doesn't work, fall back to parse_boolean() and map yes/no to one of the values from the enum. And then you can make config.incremental the enum type instead of bool.

mcassaniti:
I was having more of a think about this and perhaps it's heading in the wrong direction. If I look at the current execution flow as documented [here](https://github.com/systemd/mkosi/blob/main/mkosi/resources/mkosi.md#execution-flow) then what additional caching might get done? I can only foresee caching the entire execution flow, in which case you're better off capturing the output into a DDI and using it as a base tree.

If instead there should be more cache (in)validation modes then let's add that instead and leave `--incremental` alone.

DaanDeMeyer:
We're not doing additional caching with this mode, we're changing what we'll do when the cache is invalidated. This wouldn't be a new mode of cache invalidation, just a change in behavior of what we do when that happens.

mcassaniti:
OK, are we better off defining what to do when the cache is invalidated, or to define when the cache is invalidated which might be never?

mcassaniti:
> OK, are we better off defining what to do when the cache is invalidated, or to define when the cache is invalidated which might be never?

@DaanDeMeyer what do you think of the above? Should we have `--incremental=` or `CacheInvalidation=` as options? Which is clearer from a UX perspective?

behrmann:
I'm with Daan here, let's not remove `--incremental`. We've had enough compat break and I think it's a less scary name than cache invalidation.

mcassaniti:
I'm suggesting we actually have both and not touch `--incremental` at all.

DaanDeMeyer:
We can keep `--incremental=` as a boolean option to enable/disable incremental builds, and then add a new `IncrementalMode=` option or so to specify what to do when the cache is invalidated.

NEWS.md
```diff
@@ -4,6 +4,10 @@
 
 - `mkosi.version` is now picked up from preset and dropin directories as
   well following the usual config precedence logic
+- The option `--incremental` is now deprecated. It will be removed in a
+  future version. Use `--cache-mode=prepare` instead.
+- A new option `--cache-mode` is now available which supports more caching
+  between build runs. See the documentation for further details.
```
keszybz, line 10: Those two items should be in reverse order.

mkosi/config.py
```diff
@@ -1417,7 +1431,7 @@ class MkosiConfigParser:
             nargs="?",
             section="Host",
             parse=config_parse_boolean,
-            help="Make use of and generate intermediary cache images",
+            help="Deprecated. Alias for --cache-mode=prepare",
```
keszybz, line 1434: I think we generally remove deprecated stuff from help output with SUPPRESS.

mkosi/resources/mkosi.md
```diff
@@ -941,16 +953,8 @@ they should be specified with a boolean argument: either "1", "yes", or "true" t
...
-  prepare script has executed but before the `mkosi.build` script is
-  invoked (or anything that happens after it). On subsequent invocations
-  of `mkosi` with the `-i` switch this cached image may be used to skip
-  the OS package installation, thus drastically speeding up repetitive
-  build times. Note that while there is some rudimentary cache
-  invalidation, it is definitely not perfect. In order to force
-  rebuilding of the cached image, combine `-i` with `-ff` to ensure the
-  cached image is first removed and then re-created.
+: Deprecated. Alias for `CacheMode=prepare`. See the section on caching for
+  more details.
```
keszybz, line 957: Also here, just drop the descripition.

mkosi/resources/mkosi.md
```diff
@@ -1351,39 +1355,27 @@ project.
...
+* `none`: No caching is enabled. This is the default.
+
+* `prepare`: The execution flow is cached after packages are
+  installed and the prepare script has executed. This is true for both the
+  build overlay and the final image.
+
+  If the cache is present then the execution flow is skipped up to this point
+  and the cached contents directly used. This is useful for speeding up
+  the build process by skipping OS package installation. If the prepare script,
+  package list, build package list or configured repositories change then the
```
keszybz, line 1374: oxford comma please.

mkosi/resources/mkosi.md
```diff
@@ -1351,39 +1355,27 @@ project.
...
+  cache is not used and will be rebuilt.
+
+Caches can be removed with the `-ff` option. The cache invalidation is not
```
keszybz, line 1377: "The cache invalidation" → "Cache invalidation".

keszybz (changes requested):
I didn't follow the whole discussion, so I don't really have an opinion on whether `--incremental` should be dropped and how the high-level picture should look like. My comments are about the details of the patch only.

DaanDeMeyer:
Let's close this until you can find time to work on it again

Issue #1806, opened 2023-08-18
Discussion: Use overlay mounts instead of copying


mcassaniti:
There is a lot of copying going on in mkosi and a fair bit of it is fixed data that won't change. Rather than copy files around, why don't we use overlay mounts much more often. This would mean at least the following would become a part of a lower directory rather than a copy:

* Base Trees
* Skeleton Trees
* Any directory caches (see `resue_cache()`)
* Extra Trees

If this a bad idea then let's document the reason(s) why here. The reasons are:

* Makes supporting other caching options such as #1749 easier
* Reduces disk I/O and therefore build times
* Should make differencing between images much easier even though we don't do that yet

If there's no major objection I'll clobber together a PR. If we just want to see where it leads to I can get a PR together for that as well.

The challenges are:

* Any of the trees options might be a TAR archive which needs extracting first
* This needs to play nice with overlay, but it's possible to fit in reasonably well
* It might make sense to make this an option rather than enabled for good since OverlayFS comes with some limitations

DaanDeMeyer:
I'll take a look at any PRs but this problem is already solved by using a filesystem on the host that supports subvolumes or reflinks such as btrfs or xfs.

mcassaniti:
Reflinks are great if the FS supports them and if everything is on the same file system. I should add to the docs that it's highly recommended to use XFS or btrfs to run mkosi on top of.

DaanDeMeyer:
> Reflinks are great if the FS supports them and if everything is on the same file system. I should add to the docs that it's highly recommended to use XFS or btrfs to run mkosi on top of.

We try to not leave the filesystem that mkosi is invoked from to make reflinks work as much as possible. A documentation update mentioning that COW filesystems are recommended would be much appreciated

mcassaniti:
It's a bit of an aside, but how might `systemd-repart` be run to also make sure it uses the same file system?

DaanDeMeyer:
Not possible unfortunately without  mounting the filesystem which means root privileges. If we'd be OK with root I'd suggest to use sysupdate for this but since it requires root I don't think we want to support it.

mcassaniti:
Actually I think after taking a look that setting `TMPDIR` as the workspace [here](https://github.com/systemd/mkosi/blob/19bb5e274d9a9c23891905c4bcbb8f68955a701d/mkosi/__init__.py#L1447) actually resolves the whole issue since `systemd-repart` __should__ use that temporary directory.

[...]

PR #1809, opened 2023-08-21
Add support for using OverlayFS instead of copying

mcassaniti:
This PR is reasonably sized and makes use of overlay mounts instead of copying. This list of base directories is stored as part of `MkosiState`. The function `mount_image()` is no longer needed because we enable the `--overlay-as-copy` option whenever `--overlay` is set to prevent overlay stacking issues.

Unfortunately the `copy_tree()` function now needs access to the state which has meant a bit of code churn.

I also found that `mount_overlay()` was not properly setting up multiple lower directories, so if nothing else that change should be applied.

I successfully tested the following combinations with an `mkosi.extra` directory and `BaseTrees=` set:

* A normal build
* Cache mode of a normal build
* A build with `--overlay-as-copy`
* Cache mode of a build with `--overlay-as-copy`

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -21,7 +21,7 @@ from collections.abc import Iterator, Sequence
 from pathlib import Path
 from typing import Any, ContextManager, Mapping, Optional, TextIO, Union
 
-from mkosi.archive import extract_tar, make_cpio, make_tar
+from mkosi.archive import make_cpio, make_tar
 from mkosi.config import (
     Bootloader,
     Compression,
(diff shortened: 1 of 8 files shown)


mcassaniti:
Fixes #1806.

DaanDeMeyer:
Can we get some results here as well? How much time does this actually save? Regardless, this is too much added complexity as is if you ask me compared to asking users to use a COW filesystem.

mcassaniti:
On it's own it's probably not heaps but there's some context missing. I'm looking to add a feature that creates a diff from the full execution flow against the final image cache. This means another cache option (which we're working on) and it would be more ideal to do this with overlay support. The resulting output is still a full image, but the diff is also saved (something like `mkosi.output/diff-dir`).

Why? A desktop image build is around 15-20GB in size, meaning the cache will have that much data in it. Every time you update the image you need to decompress and extract the cache again, build the image, compress the image and then compress the cache again. The image is built on ephemeral cloud compute.

Instead of having the cache stored as a directory I'm looking to store it read-only as a bunch of SquashFS images that are overlaid together. At the end of each build you only need to store the diff in a new SquashFS image. Now we have a cache that only needs to be downloaded (quick on cloud compute), mounted and the next update to the cache uploaded at the end. It's only the output image that takes up most of the time compressing.

I'm sure there's other uses for the diff output too, just not something that systemd-import and systemd-sysupdate supports today.

As far as results go, my own tooling uses something similar and spends most of the time compressing the output image.

DaanDeMeyer:
Can't we make this work using the existing BaseTrees= and Overlay= options? Those would basically already do what you want no? You would just need to write some logic on top of mkosi to provision the BaseTrees= from your squashfs images (or you use DDIs instead of raw squashfs and it'll work OOTB).  Then to generate a full image you build with Overlay= disabled and to generate the diff you build with Overlay= enabled.

mcassaniti:
I haven't yet tested this, but I'm sure you are correct that between `BaseTrees=` and `Overlay=` that these things can be done, and likely without requiring the cache. All those base trees however will need copying.

DaanDeMeyer:
They don't need copying, check mount_base_trees(), we don't copy any base tree that's a directory there

mcassaniti:
> They don't need copying, check mount_base_trees(), we don't copy any base tree that's a directory there

That's only with overlay and I'll need both disabled and enabled. I was almost excited.

DaanDeMeyer:
Right but without overlay you have to copy regardless. How else are you going to get the full image? We now copy each individual base tree at the beginning, with your solution we would have to copy the entire image at end. Either way, if `Overlay=` is not used, you're going to have to copy regardless.

DaanDeMeyer:
@mcassaniti Did you manage to figure out if BaseTrees= + Overlay would solve your issue?

mcassaniti:
Not yet and I need to confirm how much copying there is and if it can be avoided. That might mean repart changes but let me verify first.

DaanDeMeyer:
Let's close this for now

PR #1817, opened 2023-08-22
Document our default systemd-repart definition files

DaanDeMeyer:


mkosi/resources/mkosi.md
```diff
@@ -1384,6 +1384,42 @@ local directory:
...
+  CopyFiles=/efi:/
+  SizeMinBytes=512M
+  SizeMaxBytes=512M
+  ```
+
+  `05-bios.conf` (if we're building a BIOS bootable image):
+
+  ```
+  [Partition]
+  Type=21686148-6449-6e6f-744e-656564454649
```
septatrix, line 1408: I think it would be a good idea to put a comment here that this is needed for GRUB to find it

septatrix (commented):
I think this is looking good.

After finding these files myself a part of my confusion came from the fact that as soon as one partition file was provided, one would also have to provide e.g. the EFI partition file (github showed me the old standalone files at first, only afterwards did I find the logic). I think this also clears that up sufficiently.

mkosi/resources/mkosi.md
behrmann: It is noted in the section on `BiosBootloader=`, though reminding it here might be helpful.

mkosi/resources/mkosi.md
```diff
@@ -1384,6 +1384,42 @@ local directory:
   output to stdout will be used as the credential value. Output to stderr will be ignored.
   Credentials configured with `Credentials=` take precedence over files in `mkosi.credentials`.
 
+* The **`mkosi.repart/`** directory is used as the source for
+  systemd-repart partition definition files which are passed to
+  systemd-repart when building a disk image. If it does not exist and
+  the `RepartDirectories=` setting is not configured, mkosi will default
+  to the following partition definition files:
```
behrmann, line 1391: Maybe also highlight the reverse, that if these these options are used, people will have to add configuration for these themselves.

PR #1819, opened 2023-08-22
Make sure we remove /var/lib/dbus/machine-id from the image

### LINKED ISSUE #1818: Images include dbus machine-id
In v15 it was changed such that images no longer ship a `/etc/machine-id`. While this is great in theory, the images still contain a dbus machine-id which is used by systemd to populate the normal machine-id if set.
Adding `RemoveFiles=/var/lib/dbus/machine-id` to the config fixed this and machines get proper random ids on bootup. It would be great if this file could also be excluded from the final images by default.

DaanDeMeyer:
Fixes #1818

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -157,7 +157,10 @@ def install_distribution(state: MkosiState) -> None:
             if state.config.packages:
                 state.config.distribution.install_packages(state, state.config.packages)
 
-    for f in ("var/lib/systemd/random-seed", "var/lib/systemd/credential.secret", "etc/machine-info"):
+    for f in ("var/lib/systemd/random-seed",
+              "var/lib/systemd/credential.secret",
+              "etc/machine-info",
+              "var/lib/dbus/machine-id"):
         # Using missing_ok=True still causes an OSError if the mount is read-only even if the
         # file doesn't exist so do an explicit exists() check first.
         if (state.root / f).exists():


septatrix:
This seems be an issue with mostly Debian (and likely Ubuntu). For Debian, dbus-uuidgen is called in a post-install script, however unlike Fedora, Debian has it configured to not create a symlink but instead copy the file itself. [The documentation](https://wiki.debian.org/MachineId#machine_id_and_cloned_systems.2C_generating_a_new_machine_id) states the following:

> If you are doing cloning, stateless systems or similar activities, and you know you will have a valid `/etc/machine-id` (you either use systemd or have taken other steps to have one), then you can make `/var/lib/dbus/machine-id` a symlink to `/etc/machine-id` (dbus comes with a systemd-tmpfiles file to do this). This is not done by default in Debian, or by `dbus-uuidgen --ensure`, for historical reasons; maybe it should be, but to be confident that it was a correct change I'd have to think about the ways in which it might go wrong on non-systemd systems (with either a non-systemd init like sysvinit, or no init at all like minimal chroots).

They even mention that is is mostly for historical reasons (whereas mkosi does not concern itself with such baggage) and suggest to instead symlink the file (as it is done on e.g. Fedora). Therefore I am not sure if we want to apply this to all distributions or even for debian itself for that matter and instead create the symlink manually or leave this completely to the developer (with maybe a disclaimer somewhere in the docs - though I have no clue where that would fit).

behrmann (approved):

DaanDeMeyer:
Since this is located in /var, this should be set up with a tmpfiles snippet (which the docs you link even suggest). So it shouldn't be problem to just remove the file. On Fedora, systemd-tmpfiles will just create it on first boot.

PR #1825, opened 2023-08-24
config: Add support for default options

### LINKED ISSUE #1761: Support overriding default values for options
For each supported option, we should introduce a variant that allows modifying the default value for that option if none is set explicitly. For example, for `Architecture=` we'd have `DefaultArchitecture=` and idem for other options. This would allow configuring global defaults for presets that can be overridden individually per preset if needed. This would allow configuring the default architecture for all images and sysexts built from a set of presets, except for the tools tree image which would use `Architecture=native` to make sure it always uses the host architecture.

mcassaniti:
Support allowing a user to specify an option to override the chosen default. For example, the user can set `DefaultDistribution=` or `--default-distribution` and the configuration will automatically resolve `state.config.distribution`. The user can also still specify `Distribution=` in a preset or configuration file and override the default.

This change covers setting a default if the configuration parser automatically sets a default. This means that `DefaultPackages=` is not a valid option since there is no default for `Packages=` set by the configuration parser.

This should fix #1761.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@
   parsing. Settings parsed later will now override earlier values and
   the `!` exclusion logic for lists was removed. Assigning the empty
   string to a list can be used to clear previously assigned values.
+- Configuration files now accept a `DefaultSetting=value` that can override
+  the default value specified in the documentation.
 
 ## v15.1
 
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -1535,7 +1535,22 @@ class MkosiConfigParser:
         self.settings_lookup = {s.name: s for s in self.SETTINGS}
         self.match_lookup = {m.name: m for m in self.MATCHES}
 
-    def parse_config(self, path: Path, namespace: argparse.Namespace) -> bool:
+    def add_defaults(self, args: argparse.Namespace, def_args: argparse.Namespace) -> argparse.Namespace:
+        '''Return a namespace with default values for all settings'''
+
+        ns = copy.deepcopy(def_args)
+        for s in self.SETTINGS:
+            if not s.dest in vars(ns):
+                if s.default_factory:
+                    setattr(ns, s.dest, s.default_factory(args))
+                elif s.default is None:
+                    setattr(ns, s.dest, s.parse(None, None))
+                else:
+                    setattr(ns, s.dest, s.default)
+
+        return ns
+
+    def parse_config(self, path: Path, namespace: argparse.Namespace, default_ns: argparse.Namespace) -> bool:
         extras = path.is_dir()
 
         if path.is_dir():
@@ -1575,20 +1590,8 @@ class MkosiConfigParser:
                         if not (match := s.match):
                             die(f"{k} cannot be used in [Match]")
 
-                        # If we encounter a setting in [Match] that has not been explicitly configured yet,
-                        # we assign the default value first so that we can [Match] on default values for
-                        # settings.
-                        if s.dest not in namespace:
-                            if s.default_factory:
-                                default = s.default_factory(namespace)
-                            elif s.default is None:
-                                default = s.parse(None, None)
-                            else:
-                                default = s.default
-
-                            setattr(namespace, s.dest, default)
-
-                        result = match(v, getattr(namespace, s.dest))
+                        match_v = getattr(namespace, s.dest, getattr(default_ns, s.dest, None))
+                        result = match(v, match_v)
 
                     elif (m := self.match_lookup.get(k)):
                         result = m.match(v)
@@ -1609,10 +1612,12 @@ class MkosiConfigParser:
 
         for section in parser.sections():
             for k, v in parser.items(section):
-                if not (s := self.settings_lookup.get(k)):
+                ns = default_ns if k.startswith("Default") else namespace
+
+                if not (s := self.settings_lookup.get(k.removeprefix("Default"))):
                     die(f"Unknown setting {k}")
 
-                setattr(namespace, s.dest, s.parse(v, getattr(namespace, s.dest, None)))
+                setattr(ns, s.dest, s.parse(v, getattr(ns, s.dest, None)))
 
         if extras:
             # Dropin configuration has priority over any default paths.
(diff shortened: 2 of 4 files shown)


mcassaniti:
To whoever reviews this, if it feels like `__getattribute__()` is being used as a hack then you'd be right. It does have the nice benefit that code doesn't need to do `state.config.XXX == YYY or state.config.default_XXX == YYY` for example.

Some default options are automatically set as part of configuration parsing so I can confirm everything works as expected without needing further unit tests. I also tried to find where defaults were being explicitly configured and handle those too.

Most of the PR is code churn for types.

If you wish I can extend this further so that the root `mkosi.conf` can have things like `DefaultPackages=` set and allow a preset to override the setting, or potentially even support `OverridePackages=`.

DaanDeMeyer:
So I've been thinking about this and I figure instead of adding default options we might be better off with adding support for overriding options, as long as we make it explicit, it should be fine and also solves the same issues. The advantage of overriding over defaults is that users can keep using the same options instead of having to set --default-xxx all the time.

So I would suggest adding a way to override instead. This should also be super simple to implement. If an option starts with `Override`, delete the option from the namespace if it is in there already before parsing the given value.

mcassaniti:
I raised #1829 for override support. If that fits we'll close this one off and probably #1761.

DaanDeMeyer:
Let's close this one as I'm confident #1829 is the way to go

mcassaniti:
Is it too late to re-open this PR or do I need to submit another one after your comments on #1830?

mcassaniti:
Better do a proper review. There's no documentation for instance and if you don't want to set `--default-XXX` on the CLI then then better be changed now too.

DaanDeMeyer (changes requested):
This approach seems rather complicated. Why not keep an extra `argparse.Namespace` around containing the default values for settings. You initialize that namespace with the defaults configured in the settings themselves. And then when a setting starts with `Default`, you pass in the defaults namespace to the parser function instead of the regular one.

Finally at the end of parsing, if a setting is not in the regular namespace, you initialize its value with the value from the defaults namespace.

I also don't think we need to add default variants to the CLI options, we should only support them in the config files for now.

NEWS.md
```diff
@@ -8,6 +8,10 @@
   parsing. Settings parsed later will now override earlier values and
   the `!` exclusion logic for lists was removed. Assigning the empty
   string to a list can be used to clear previously assigned values.
+- Configuration files now accept a `DefaultSetting=value` that will be
+  used if the `Setting=` is not set elsewhere. This allows for such things
+  as `DefaultRelease=stable` in the main configuration and `Release=unstable`
+  to be used in a preset.
```
DaanDeMeyer, line 14: This is a bad example, now that the first assignment doesn't take precedence anymore, you can just do `Release=unstable` in the preset file and it will override the main configuration. Where this is useful is to override mkosi's default value so that a value can be specified explicitly in the CLI but if it's ,not set in the CLI the correct default value is used.

DaanDeMeyer: Yeah that sounds good

mkosi/config.py
```diff
@@ -1575,20 +1590,8 @@ def parse_config(self, path: Path, namespace: argparse.Namespace) -> bool:
...
-                                default = s.default_factory(namespace)
-                            elif s.default is None:
-                                default = s.parse(None, None)
-                            else:
-                                default = s.default
-
-                            setattr(namespace, s.dest, default)
-
-                        result = match(v, getattr(namespace, s.dest))
+                        match_v = getattr(namespace, s.dest, getattr(default_ns, s.dest, None))
```
DaanDeMeyer, line 1593: We should assign the default value to the main namespace here as we did before

mkosi/config.py
```diff
@@ -1535,7 +1535,22 @@ def __init__(self) -> None:
...
+                if s.default_factory:
+                    setattr(ns, s.dest, s.default_factory(args))
+                elif s.default is None:
+                    setattr(ns, s.dest, s.parse(None, None))
+                else:
+                    setattr(ns, s.dest, s.default)
+
+        return ns
+
+    def parse_config(self, path: Path, namespace: argparse.Namespace, default_ns: argparse.Namespace) -> bool:
```
DaanDeMeyer, line 1553: Please just name it `defaults` instead of `detault_ns`

mkosi/config.py
```diff
@@ -1575,20 +1590,8 @@ def parse_config(self, path: Path, namespace: argparse.Namespace) -> bool:
...
-                        # If we encounter a setting in [Match] that has not been explicitly configured yet,
-                        # we assign the default value first so that we can [Match] on default values for
-                        # settings.
-                        if s.dest not in namespace:
-                            if s.default_factory:
-                                default = s.default_factory(namespace)
-                            elif s.default is None:
-                                default = s.parse(None, None)
-                            else:
-                                default = s.default
```
DaanDeMeyer, line 1587: Basically you should add another if condition here (that takes precedence over all these) which uses the value from the defaults namespace if it is in the defaults namespace.

mkosi/config.py
```diff
@@ -1897,41 +1905,28 @@ def parse(self, argv: Optional[Sequence[str]] = None) -> tuple[MkosiArgs, tuple[
...
-            for s in self.SETTINGS:
-                if s.dest in ns:
-                    continue
-
-                if s.default_factory:
-                    default = s.default_factory(ns)
-                elif s.default is None:
-                    default = s.parse(None, None)
-                else:
-                    default = s.default
```
DaanDeMeyer, line 1926: Same here

DaanDeMeyer (changes requested):
See comments, I think we can simplify this a lot more

DaanDeMeyer:
I need this myself so I put up my own version in https://github.com/systemd/mkosi/pull/1836, would still love to see a test suite for this though

mcassaniti:
I'll try to get to the test suite next week. It will highlight just how hard last value wins + computed defaults is.

mcassaniti:
Tests in #1840. Closing.

PR #1827, opened 2023-08-24
Manifest fixes

### LINKED ISSUE #1823: ManifestFormat should default to empty
By default `ManifestFormat` is set to `json`; when using `--format=directory` from the command line, this produces an extra file, which the user probably doesn't want if they're just trying to make a directory tree. One can pass `--manifest-format=''` to avoid that but it's not the easiest to discover, so I'd recommend defaulting this to the empty string instead so that no manifest is generated unless requested.

DaanDeMeyer:
Fixes #1823

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1270,7 +1270,10 @@ def dir_size(path: Union[Path, os.DirEntry[str]]) -> int:
     return dir_sum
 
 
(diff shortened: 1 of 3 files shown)


behrmann (approved):
The fixes to skip creation make sense. I'm not quite sure about disabling the manifest by default, instead it should be more useful, but I guess this is still an ok intermediate step.

PR #1828, opened 2023-08-24
Reword output docs a bit

DaanDeMeyer:


mkosi/resources/mkosi.md
```diff
@@ -442,15 +442,12 @@ they should be specified with a boolean argument: either "1", "yes", or "true" t
...
-  configured it is used instead of `image` in the default output name.
-  If an image version is specified via `ImageVersion=`, it is included
-  in the default name, e.g. a specified image version of `7.8` might
-  result in an image file name of `image_7.8.raw.xz`.
+: Name to use for the generated output image file or directory. All
+  outputs will be prefixed with the given name. Defaults to `image` or,
+  if `ImageId=` is specified, it is used as the default output name,
+  optionally suffixed with the version set with `ImageVersion=`. Note
+  that this option does not allow configuring the output directory, use
+  `OutputDirectory=` for that.
```
behrmann, line 450: This is definitely clearer than the old version, but I don't think we should lose the examples from the old version. How about adding that one in a second paragraph?

behrmann (approved):

PR #1830, opened 2023-08-24
Drop "first assignment wins" logic

DaanDeMeyer:
From experience in the systemd repository's usage of presets, we've
learned that we want to have fixed values for certain settings for
presets that cannot be overridden by either the CLI options or by
global config. Good examples are that the Format= of a base image
should always be "directory" and the Format= of an initrd image
should always be "cpio" or that Bootable= should always be "no" for
both the base image and the initrd image and their list of packages
should not be affected by any packages specified on the CLI.

The issue is that with "first assignment wins" logic, we need to add
an explicit "override" mechanism which almost all settings in these
presets would then use to make sure they can't be changed by CLI
options. This seems rather backwards, and is a good indication that
any settings configured in config should not be overridden by settings
set on the CLI.

Even disregarding usage of presets, any existing mkosi config will almost
certainly not be written to expect arbitrary changes in the config due
to options set on the CLI.

Also, it's currently not entirely trivial to set default values for
presets from the global config, because any values set in the global
config cannot be overridden anymore by presets. By not doing "first
assignment wins" logic, this becomes trivial as the global config can
simply set a default value and it can be overridden by presets.

Of course by removing "first assignment wins" logic, we do introduce
the issue again that "first assignment wins" solves in the first place,
which is that it becomes possible to assign a value to a setting, match
on that value and then change the setting later. We acknowledge this by
documenting it in the manual. Also, in some cases, this is exactly what
you want. For example, if you want to use a Fedora rawhide tools tree to
build CentOS 8 images, you have to first match on distribution == centos
and then set Distribution=fedora afterwards for the tools tree preset, so
this actually makes perfect sense in some cases.

While this is technically a compat break, it will only be noticed by users
doing advanced stuff with mkosi, which AFAIK does not exist yet outside of
the systemd repo. In fact even the systemd repo was not broken by this change,
so we should be OK with making it, given the large benefits we get out of it.

This commit also simplifies the interfaces of the parser and matching callbacks
to not take the namespace as an argument anymore, but to simply take the existing
value as an argument instead.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,10 @@
 
 - `mkosi.version` is now picked up from preset and dropin directories as
   well following the usual config precedence logic
+- The "first assignment wins" logic was dropped from configuration
+  parsing. Settings parsed later will now override earlier values and
+  the `!` exclusion logic for lists was removed. Assigning the empty
+  string to a list can be used to clear previously assigned values.
 
 ## v15.1
 
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -41,8 +41,8 @@ from mkosi.versioncomp import GenericVersion
 
 __version__ = "15.1"
 
-ConfigParseCallback = Callable[[str, Optional[str], argparse.Namespace], Any]
-ConfigMatchCallback = Callable[[str, str, argparse.Namespace], bool]
+ConfigParseCallback = Callable[[Optional[str], Optional[Any]], Any]
+ConfigMatchCallback = Callable[[str, Optional[Any]], bool]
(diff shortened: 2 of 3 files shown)


DaanDeMeyer:
@mcassaniti With this PR your PR adding the `Default` variants will be the way to go instead of adding `Override` variants. The `Default` variants would then be used to change the default for a preset without taking away the ability to set a specific setting on the CLI.

behrmann (approved):
I think this makes sense and I agree with the reasoning in the commit message, though that this is a compat break should go into the news as well, maybe together with a hint which sections in the man were touched by this.

PR #1834, opened 2023-08-24
propagate SOURCE_DATE_EPOCH when calling systemd-repart

malt3:
With this change, setting `Environment=SOURCE_DATE_EPOCH=0` in the mkosi config, mkosi will propagate the variable to systemd-repart.
systemd-repart calls mksquashfs and other mkfs tools.
Some mkfs tools already respect SOURCE_DATE_EPOCH, resulting in predictable timestamps in squashfs partitions (and maybe others).

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1721,6 +1721,8 @@ def make_image(state: MkosiState, skip: Sequence[str] = [], split: bool = False)
     for option, value in state.config.environment.items():
         if option.startswith("SYSTEMD_REPART_MKFS_OPTIONS_"):
             env[option] = value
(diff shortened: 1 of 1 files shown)


behrmann (approved):
Makes sense, though I do think we should propagate `SOURCE_DATE_EPOCH` to every `run` invocation.

malt3:
> Makes sense, though I do think we should propagate `SOURCE_DATE_EPOCH` to every `run` invocation.

Sounds like a good idea. I'll do a follow up PR after understanding how run works.

DaanDeMeyer (approved):

PR #1836, opened 2023-08-25
Support configuring default values for settings

DaanDeMeyer:
This allows configuring the default value for a setting. This is useful to set a custom default value for a setting while still allowing an explicit value from the CLI to take precedence.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@
   parsing. Settings parsed later will now override earlier values and
   the `!` operator for lists will now remove values already in the list
   instead of preventing specific values from being added.
+- Add support for configuring custom default values for settings by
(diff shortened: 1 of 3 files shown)


mcassaniti:
Default factories need to process on both namespaces. Why? take a look at `default_config_mirror()` and consider `@Distribution=` + `Release=` being in play.

DaanDeMeyer:
@mcassaniti Good point, please check the latest version, I introduced a default_factory_depends field to make sure we resolve all the defaults of settings some setting depends on before resolving the default of that setting itself.

mcassaniti:
I thought of the problems, just not the test suite yet. Recursively resolving each default is something I didn't consider and certainly makes the whole thing much easier.

PR #1837, opened 2023-08-25
add config setting seed to set systemd-repart `--seed`

malt3:
Seed takes a UUID as argument or the special value **random**.
Overrides the seed that [`systemd-repart(8)`](https://www.freedesktop.org/software/systemd/man/systemd-repart.service.html) uses when building a disk image. This is useful to achieve reproducible builds, where deterministic UUIDs and other partition metadata should be derived on each build.

In particular, this makes GPT UUIDs deterministic as well as the dm-verity UUID and salt.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1633,6 +1633,7 @@ def make_image(state: MkosiState, skip: Sequence[str] = [], split: bool = False)
         "--no-pager",
         "--offline=yes",
         "--root", state.root,
+        "--seed", str(state.config.seed) if state.config.seed else "random",
         state.staging / state.config.output_with_format,
     ]
 
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -19,6 +19,7 @@ import shutil
 import subprocess
 import sys
 import textwrap
+import uuid
 from collections.abc import Iterable, Sequence
 from pathlib import Path
 from typing import Any, Callable, Optional, Type, Union, cast
@@ -247,6 +248,16 @@ def config_parse_compression(value: Optional[str], old: Optional[Compression]) -
         return Compression[value]
     except KeyError:
         return Compression.zst if parse_boolean(value) else Compression.none
+
+def config_parse_seed(value: Optional[str], old: Optional[str]) -> Optional[uuid.UUID]:
+    if not value or value == "random":
+        return None
+
+    try:
+        return uuid.UUID(value)
+    except ValueError:
+        die(f"{value} is not a valid UUID")
+
 
 
 def config_default_release(namespace: argparse.Namespace) -> str:
@@ -635,6 +646,7 @@ class MkosiConfig:
     sector_size: Optional[str]
     overlay: bool
     use_subvolumes: ConfigFeature
+    seed: Optional[uuid.UUID]
 
     packages: list[str]
     build_packages: list[str]
@@ -993,6 +1005,13 @@ class MkosiConfigParser:
             section="Output",
             parse=config_parse_feature,
             help="Use btrfs subvolumes for faster directory operations where possible",
+        ),
+         MkosiConfigSetting(
+            dest="seed",
+            metavar="UUID",
+            section="Output",
+            parse=config_parse_seed,
+            help="Set the seed for systemd-repart",
         ),
 
         MkosiConfigSetting(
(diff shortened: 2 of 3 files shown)


mkosi/config.py
```diff
@@ -994,6 +995,12 @@ class MkosiConfigParser:
             parse=config_parse_feature,
             help="Use btrfs subvolumes for faster directory operations where possible",
         ),
+         MkosiConfigSetting(
+            dest="seed",
```
DaanDeMeyer, line 999: Please also set metavar

mkosi/config.py
```diff
@@ -994,6 +995,12 @@ class MkosiConfigParser:
...
+            section="Output",
+            parse=config_parse_string,
```
DaanDeMeyer, line 1001: Can you add config_parse_seed() for this and do a basic check on whether the value is a proper UUID?

DaanDeMeyer (changes requested):

mkosi/config.py
malt3: Would this be a proper metavar? `UUID|random`

DaanDeMeyer: I would just use UUID

mkosi/config.py
```diff
@@ -248,6 +249,18 @@ def config_parse_compression(value: Optional[str], old: Optional[Compression]) -
     except KeyError:
         return Compression.zst if parse_boolean(value) else Compression.none
 
+def config_parse_seed(value: Optional[str], old: Optional[str]) -> Optional[str]:
+    if not value:
+        return None
+
+    if value == "random":
+        return value
```
DaanDeMeyer, line 257: If we calculate a random UUID ourselves, we can make the type of `seed` in `MkosiConfig of `UUID` instead of having to store a string.

DaanDeMeyer (changes requested):

mkosi/config.py
malt3: I think the  meaning of no `--seed` flag is different from `random` and different from a specific uuid:

> If not specified it is attempted to read the machine ID from the host (or more precisely, the root directory configured via --root=) and use it as seed instead, falling back to a randomized seed otherwise. Use --seed=random to force a randomized seed. Explicitly specifying the seed may be used to generated strictly reproducible partition tables.

 So i think we either need a type that can hold any valid UUID and `random` or should just use `Optional[str]`.

DaanDeMeyer: Right but we never want /etc/machine-id to be used in our case since there is none anyway. So always specifiying `--seed` and just making it random if not specified seems like the way to go.

malt3: Should we even allow `random` as a user chosen value in the mkosi config?
I feel like users of repart could be tempted to call mkosi with `--seed random` so this should still be allowed and could be internally tracked as `None`.

DaanDeMeyer: That sounds good to me, but make sure to translate `None` back to "random" in the summary then

DaanDeMeyer: And also pass `--randomize` to systemd-repart if `seed` is `None`

mkosi/__init__.py
```diff
@@ -1653,6 +1653,11 @@ def make_image(state: MkosiState, skip: Sequence[str] = [], split: bool = False)
     if state.config.sector_size:
         cmdline += ["--sector-size", state.config.sector_size]
 
+    if state.config.seed:
+        cmdline += ["--seed", str(state.config.seed)]
+    else:
+        cmdline += ["--seed", "random"]
```
DaanDeMeyer, line 1659: We can just make this a one liner `["--seed", str(state.config.seed) if state.config.seed else "random"]`

mkosi/config.py
```diff
@@ -2133,6 +2152,9 @@ def yes_no_auto(f: ConfigFeature) -> str:
 def none_to_na(s: Optional[object]) -> str:
     return "n/a" if s is None else str(s)
 
+def none_to_random(s: Optional[object]) -> str:
+    return "random" if s is None else str(s)
+
```
DaanDeMeyer, line 2157: Please leave two newlines between functions

DaanDeMeyer (changes requested):
Two more small changes and it's good to go

mkosi/__init__.py
DaanDeMeyer: And then you can also make it part of the `cmdline` declaration.

DaanDeMeyer (approved):

PR #1839, opened 2023-08-25
normalize mtime

katexochen:
First part of the implementation of #687 

If set, the time stamp from SOURCE_DATE_EPOCH is used to normalize mtime of files. We also need to pass the environment trough when mkosi is invoking itself.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -934,6 +934,8 @@ def build_initrd(state: MkosiState) -> Path:
         "--make-initrd", "yes",
         "--bootable", "no",
         "--manifest-format", "",
+        *(["--source-date-epoch", str(state.config.source_date_epoch)]
+                if state.config.source_date_epoch is not None else []),
         *(["--locale", state.config.locale] if state.config.locale else []),
         *(["--locale-messages", state.config.locale_messages] if state.config.locale_messages else []),
         *(["--keymap", state.config.keymap] if state.config.keymap else []),
@@ -1813,10 +1815,13 @@ def build_image(args: MkosiArgs, config: MkosiConfig) -> None:
             run_selinux_relabel(state)
             run_finalize_script(state)
 
+        normalize_mtime(state.root, state.config.source_date_epoch)
         partitions = make_image(state, skip=("esp", "xbootldr"))
         install_unified_kernel(state, partitions)
         prepare_grub_efi(state)
         prepare_grub_bios(state, partitions)
+        normalize_mtime(state.root, state.config.source_date_epoch, directory=Path("boot"))
+        normalize_mtime(state.root, state.config.source_date_epoch, directory=Path("efi"))
         partitions = make_image(state)
         install_grub_bios(state, partitions)
         make_image(state, split=True)
@@ -2255,3 +2260,14 @@ def run_verb(args: MkosiArgs, presets: Sequence[MkosiConfig]) -> None:
 
             if args.verb == Verb.serve:
                 run_serve(last)
+
+
+def normalize_mtime(root: Path, mtime: Optional[int], directory: Optional[Path] = None) -> None:
+    directory = directory or Path("")
+    if mtime is None:
+        return
+
+    with complete_step(f"Normalizing modification times of /{directory}"):
+        os.utime(root / directory, (mtime, mtime), follow_symlinks=False)
+        for p in (root / directory).rglob("*"):
+            os.utime(p, (mtime, mtime), follow_symlinks=False)
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -249,6 +249,7 @@ def config_parse_compression(value: Optional[str], old: Optional[Compression]) -
     except KeyError:
         return Compression.zst if parse_boolean(value) else Compression.none
 
+
 def config_parse_seed(value: Optional[str], old: Optional[str]) -> Optional[uuid.UUID]:
     if not value or value == "random":
         return None
@@ -258,6 +259,18 @@ def config_parse_seed(value: Optional[str], old: Optional[str]) -> Optional[uuid
     except ValueError:
         die(f"{value} is not a valid UUID")
 
+
+def config_parse_source_date_epoch(value: Optional[str], old: Optional[int]) -> Optional[int]:
+    if value is None:
+        return None
+
+    try:
+        timestamp = int(value)
+    except ValueError:
+        raise ValueError(f"{value} is not a valid timestamp")
+    if timestamp < 0:
+        raise ValueError(f"{value} is negative")
+    return timestamp
 
 
 def config_default_compression(namespace: argparse.Namespace) -> Compression:
@@ -308,6 +321,13 @@ def config_default_mirror(namespace: argparse.Namespace) -> Optional[str]:
         return "https://distfiles.gentoo.org"
 
     return None
+
+
+def config_default_source_date_epoch(namespace: argparse.Namespace) -> Optional[int]:
+    for env in namespace.environment:
+        if env.startswith("SOURCE_DATE_EPOCH="):
+            return config_parse_source_date_epoch(env.removeprefix("SOURCE_DATE_EPOCH="), None)
+    return config_parse_source_date_epoch(os.environ.get("SOURCE_DATE_EPOCH"), None)
 
 
 def make_enum_parser(type: Type[enum.Enum]) -> Callable[[str], enum.Enum]:
@@ -649,6 +669,7 @@ class MkosiConfig:
     remove_packages: list[str]
     remove_files: list[str]
     clean_package_metadata: ConfigFeature
+    source_date_epoch: Optional[int]
 
     prepare_script: Optional[Path]
     build_script: Optional[Path]
@@ -1000,7 +1021,7 @@ class MkosiConfigParser:
             parse=config_parse_feature,
             help="Use btrfs subvolumes for faster directory operations where possible",
         ),
-         MkosiConfigSetting(
+        MkosiConfigSetting(
             dest="seed",
             metavar="UUID",
             section="Output",
@@ -1092,6 +1113,15 @@ class MkosiConfigParser:
             section="Content",
             parse=config_parse_feature,
             help="Remove package manager database and other files",
+        ),
+        MkosiConfigSetting(
+            dest="source_date_epoch",
+            metavar="TIMESTAMP",
+            section="Content",
+            parse=config_parse_source_date_epoch,
+            default_factory=config_default_source_date_epoch,
+            default_factory_depends=("environment",),
+            help="Set the $SOURCE_DATE_EPOCH timestamp",
         ),
         MkosiConfigSetting(
             dest="prepare_script",
@@ -2055,6 +2085,8 @@ def load_environment(args: argparse.Namespace) -> dict[str, str]:
         env["IMAGE_ID"] = args.image_id
     if args.image_version is not None:
         env["IMAGE_VERSION"] = args.image_version
+    if args.source_date_epoch is not None:
+        env["SOURCE_DATE_EPOCH"] = str(args.source_date_epoch)
     if (proxy := os.environ.get("http_proxy")):
         env["http_proxy"] = proxy
     if (proxy := os.environ.get("https_proxy")):
@@ -2236,6 +2268,7 @@ def summary(args: MkosiArgs, config: MkosiConfig) -> str:
                Remove Packages: {line_join_list(config.remove_packages)}
                   Remove Files: {line_join_list(config.remove_files)}
 Clean Package Manager Metadata: {yes_no_auto(config.clean_package_metadata)}
+             Source Date Epoch: {none_to_none(config.source_date_epoch)}
 
                 Prepare Script: {none_to_none(config.prepare_script)}
                   Build Script: {none_to_none(config.build_script)}
--- a/mkosi/resources/mkosi.md
+++ b/mkosi/resources/mkosi.md
@@ -577,6 +577,16 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
   builds, where deterministic UUIDs and other partition metadata should be
   derived on each build.
 
+`SourceDateEpoch=`, `--source-date-epoch=`
+
+: Takes a timestamp as argument. Resets file modification times of all files to
+  this timestamp. The variable is also propagated to systemd-repart and
+  scripts executed by mkosi. If not set explicitly, `SOURCE_DATE_EPOCH` from
+  `--environment` and from the host environment are tried in that order.
+  This is useful to make builds reproducible. See
+  [SOURCE_DATE_EPOCH](https://reproducible-builds.org/specs/source-date-epoch/)
+  for more information.
+
 ### [Content] Section
 
 `Packages=`, `--package=`, `-p`
@@ -1275,6 +1285,12 @@ Scripts executed by mkosi receive the following environment variables:
   without or with networking is being executed (`WithNetwork=no`).
   The build script should avoid any network communication in case
   `$WITH_NETWORK` is `0`.
+
+* `$SOURCE_DATE_EPOCH` is defined if requested (`SourceDateEpoch=TIMESTAMP`,
+  `Environment=SOURCE_DATE_EPOCH=TIMESTAMP` or the host environment variable
+  `$SOURCE_DATE_EPOCH`). This is useful to make builds reproducible. See
+  [SOURCE_DATE_EPOCH](https://reproducible-builds.org/specs/source-date-epoch/)
+  for more information.
 
 Additionally, when a script is executed, a few scripts are made
 available via `$PATH` to simplify common usecases.


mkosi/archive.py
```diff
@@ -102,3 +104,21 @@ def make_cpio(src: Path, dst: Path, files: Optional[Iterable[Path]] = None) -> N
...
         options=finalize_passwd_mounts(dst),
     )
+
+
+def normalize_mtime(root: Path, source_date_epoch: Optional[str]) -> None:
+    if source_date_epoch is None:
+        return
+
+    try:
+        mtime_epoch = int(source_date_epoch)
```
behrmann, line 114: This is probably good enough, but considering the nanosecond resolution used below, wouldn't a float be valid as well? Does this make sense? Are there people using subsecond resolution for this?

mkosi/archive.py
```diff
@@ -102,3 +104,21 @@ def make_cpio(src: Path, dst: Path, files: Optional[Iterable[Path]] = None) -> N
...
+    except ValueError:
+        die(f"Invalid SOURCE_DATE_EPOCH value: {source_date_epoch}")
+    if mtime_epoch < 0:
+        die(f"Invalid SOURCE_DATE_EPOCH value: {source_date_epoch}")
+
+    mtime_epoch_ns = mtime_epoch * 10**9
+
+    with complete_step("Normalizing mtime"):
+        for p in root.rglob("*"):
+            os.utime(p, (mtime_epoch_ns, mtime_epoch_ns), follow_symlinks=False)
```
behrmann, line 124: The signature of `os.utime` is
```python
def os.utime(path, times=None, *, [ns, ]dir_fd=None, follow_symlinks=True)
```
so `(mtime_epoch_ns, mtime_epoch_ns)` here is `time` instead of `ns` and the docs say
> If times is not None, it must be a 2-tuple of the form (atime, mtime) where each member is an int or float expressing seconds.

so what you probably want is
```suggestion
            os.utime(p, ns=(mtime_epoch_ns, mtime_epoch_ns), follow_symlinks=False)
```

mkosi/__init__.py
```diff
@@ -934,6 +934,7 @@ def build_initrd(state: MkosiState) -> Path:
         "--make-initrd", "yes",
         "--bootable", "no",
         "--manifest-format", "",
+        *(f"--environment={k}={v}" for k, v in state.config.environment.items()),
```
behrmann, line 937: I think we should only plumb `SOURCE_DATE_EPOCH` through, instead of the whole environment.

mkosi/archive.py
malt3: After looking at this again, I'm pretty sure SOURCE_DATE_EPOCH is meant to be an integer:
> The value MUST be an ASCII representation of an integer with no fractional component, identical to the output format of date +%s.


https://reproducible-builds.org/specs/source-date-epoch/

mkosi/archive.py
katexochen: I guess we misunderstood the function signature there. I've updated to only use seconds.

mkosi/archive.py
```diff
@@ -102,3 +104,19 @@ def make_cpio(src: Path, dst: Path, files: Optional[Iterable[Path]] = None) -> N
...
+def normalize_mtime(root: Path, source_date_epoch: Optional[str]) -> None:
+    if source_date_epoch is None:
+        return
+
+    try:
+        mtime_epoch = int(source_date_epoch)
+    except ValueError:
+        die(f"Invalid SOURCE_DATE_EPOCH value: {source_date_epoch}")
+    if mtime_epoch < 0:
+        die(f"Invalid SOURCE_DATE_EPOCH value: {source_date_epoch}")
```
behrmann, line 118: Let's make the error message a bit more concrete
```suggestion
        die(f"SOURCE_DATE_EPOCH={source_date_epoch} is not a valid integer")
    if mtime_epoch < 0:
        die(f"SOURCE_DATE_EPOCH={source_date_epoch} is negative")
```

mkosi/__init__.py
```diff
@@ -934,6 +934,7 @@ def build_initrd(state: MkosiState) -> Path:
         "--make-initrd", "yes",
         "--bootable", "no",
         "--manifest-format", "",
+        *([f"--environment=SOURCE_DATE_EPOCH={state.config.environment['SOURCE_DATE_EPOCH']}"] if "SOURCE_DATE_EPOCH" in state.config.environment else []),
```
keszybz, line 937: A linebreak before `if` would be nice.

mkosi/archive.py
```diff
@@ -102,3 +104,19 @@ def make_cpio(src: Path, dst: Path, files: Optional[Iterable[Path]] = None) -> N
...
+        return
+
+    try:
+        mtime_epoch = int(source_date_epoch)
+    except ValueError:
+        die(f"SOURCE_DATE_EPOCH={source_date_epoch} is not a valid integer")
+    if mtime_epoch < 0:
+        die(f"SOURCE_DATE_EPOCH={source_date_epoch} is negative")
+
+    with complete_step("Normalizing mtime"):
```
keszybz, line 120: Maybe "Setting mtime to $SOURCE_DATE_EPOCH"? When I saw "normalizing mtime", I had no idea what exactly this means until I read the code.

keszybz (approved):
LGTM.

keszybz:
Oh, and please put the text that is now the description of the pull request in the commit message. Signed-off-by is not defined by mkosi, so that can be dropped.

DaanDeMeyer (changes requested):
We should normalize the mtime before calling make_tar() or make_cpio() instead of doing it inside those functions

malt3:
> We should normalize the mtime before calling make_tar() or make_cpio() instead of doing it inside those functions

The reason we did it this way is that there is a high probability that the code becomes incorrect over time otherwise.

We noticed that we need to normalize mtimes multiple times during a build since cpios for example are created during the build (and need normalized mtimes), after which files in the sysroot are modified (requiring us to normalize mtimes again).
If we place this code in `make_tar`, `make_cpio` and `make_image`, there is no chance for file modifications between normalizing mtimes and creating the archive / partitions.

mkosi/__init__.py
```diff
@@ -1624,6 +1627,8 @@ def make_image(state: MkosiState, skip: Sequence[str] = [], split: bool = False)
     if not state.config.output_format == OutputFormat.disk:
         return []
 
+    normalize_mtime(state.root, state.config.environment.get("SOURCE_DATE_EPOCH"))
```
DaanDeMeyer, line 1630: We should call this before we call make_image(), and then fix the mtime's of any UKIs we install later individually.

mkosi/__init__.py
```diff
@@ -1821,6 +1826,7 @@ def build_image(args: MkosiArgs, config: MkosiConfig) -> None:
         install_grub_bios(state, partitions)
         make_image(state, split=True)
 
+        normalize_mtime(state.root,  state.config.environment.get("SOURCE_DATE_EPOCH"))
```
DaanDeMeyer, line 1829: This one shouldn't be needed if we call normalize_mtime() before we call make_image()

mkosi/archive.py
```diff
@@ -102,3 +102,19 @@ def make_cpio(src: Path, dst: Path, files: Optional[Iterable[Path]] = None) -> N
         # Make sure tar uses user/group information from the root directory instead of the host.
         options=finalize_passwd_mounts(dst),
     )
+
+
+def normalize_mtime(root: Path, source_date_epoch: Optional[str]) -> None:
```
DaanDeMeyer, line 107: I don't think this function fits here. We should do the environment variable parsing in the MkosiState constructor and then expose a property on MkosiState with the parsed value. And then just have a function normalize_mtime() in the `__init__` module that takes the  parsed mtime as an input.

DaanDeMeyer (changes requested):

mkosi/__init__.py
malt3: We tried to find all instances where files below state.root are modified and reset mtimes accordingly.

mkosi/state.py
```diff
@@ -1,14 +1,17 @@
...
 
 from mkosi.config import MkosiArgs, MkosiConfig
+from mkosi.log import die
 from mkosi.tree import make_tree
 from mkosi.util import umask
 
 
 class MkosiState:
     """State related properties."""
+    source_date_epoch: Optional[int] = None
```
behrmann, line 14: Why make this a class variable instead of an instance variable?

mkosi/state.py
```diff
@@ -21,6 +24,15 @@ def __init__(self, args: MkosiArgs, config: MkosiConfig, workspace: Path) -> Non
...
         self.install_dir.mkdir(exist_ok=True)
         self.cache_dir.mkdir(parents=True, exist_ok=True)
+        source_date_epoch = self.config.environment.get("SOURCE_DATE_EPOCH")
+        if source_date_epoch is not None:
+            try:
+                mtime_epoch = int(source_date_epoch)
+            except ValueError:
+                die(f"SOURCE_DATE_EPOCH={source_date_epoch} is not a valid integer")
+            if mtime_epoch < 0:
+                die(f"SOURCE_DATE_EPOCH={source_date_epoch} is negative")
```
behrmann, line 34: Could we maybe just (re)raise a `ValueError` in both cases? We `die` all over the place, but in a constructor it's a bit icky.

mkosi/__init__.py
```diff
@@ -2255,3 +2270,24 @@ def run_verb(args: MkosiArgs, presets: Sequence[MkosiConfig]) -> None:
...
+    with complete_step("Setting mtime to SOURCE_DATE_EPOCH"):
+        os.utime(root, (source_date_epoch, source_date_epoch), follow_symlinks=False)
+        for p in root.rglob("*"):
+            os.utime(p, (source_date_epoch, source_date_epoch), follow_symlinks=False)
+
+
+def normalize_mtime_path(root: Path, target: Path, source_date_epoch: Optional[int]) -> None:
+    if source_date_epoch is None:
+        return
+    assert root in target.parents
```
behrmann, line 2287: `asserts` are problematic in Python, since they are easy to lose (e.g. setting the environment variable `PYTHONOPTIMIZE=1`). They're ok for type refinements, but to catch an error, make it
```suggestion
    if  root not in target.parents:
        raise WhichEverExceptionYouThinkSensible("some helpful message")
```
and put empty lines around it for readability.

mkosi/state.py
katexochen: just unsure how to use type hints properly :sweat_smile:

mkosi/__init__.py
```diff
@@ -934,6 +944,8 @@ def build_initrd(state: MkosiState) -> Path:
         "--make-initrd", "yes",
         "--bootable", "no",
         "--manifest-format", "",
+        *([f"--environment=SOURCE_DATE_EPOCH={state.config.environment['SOURCE_DATE_EPOCH']}"]
+            if "SOURCE_DATE_EPOCH" in state.config.environment else []),
```
keszybz, line 948: The indentation is off by one here.

mkosi/__init__.py
```diff
@@ -2255,3 +2270,27 @@ def run_verb(args: MkosiArgs, presets: Sequence[MkosiConfig]) -> None:
...
+def normalize_mtime(root: Path, source_date_epoch: Optional[int]) -> None:
+    if source_date_epoch is None:
+        return
+    with complete_step("Setting mtime to SOURCE_DATE_EPOCH"):
+        os.utime(root, (source_date_epoch, source_date_epoch), follow_symlinks=False)
+        for p in root.rglob("*"):
+            os.utime(p, (source_date_epoch, source_date_epoch), follow_symlinks=False)
+
+
+def normalize_mtime_path(root: Path, target: Path, source_date_epoch: Optional[int]) -> None:
```
keszybz, line 2284: Please merge those two functions like this:
```python
def normalize_mtime(source_date_epoch: Optional[int], root: Path, target: Optional[Path] = None):
    ...
```

You include `root` in the message when root is reset, but don't include `target` in the message. I think it should be the other way around. You should only print the path relative to `root`. When resetting the whole root, don't print the path, it's not interesting.

mkosi/state.py
```diff
@@ -9,7 +10,6 @@
 
 class MkosiState:
     """State related properties."""
-
```
keszybz, line 12: This line was good.

mkosi/state.py
```diff
@@ -21,6 +21,16 @@ def __init__(self, args: MkosiArgs, config: MkosiConfig, workspace: Path) -> Non
         self.pkgmngr.mkdir()
         self.install_dir.mkdir(exist_ok=True)
         self.cache_dir.mkdir(parents=True, exist_ok=True)
+        source_date_epoch = self.config.environment.get("SOURCE_DATE_EPOCH")
```
keszybz, line 24: Please add an empty line here.

mkosi/state.py
```diff
@@ -21,6 +21,16 @@ def __init__(self, args: MkosiArgs, config: MkosiConfig, workspace: Path) -> Non
...
+        self.source_date_epoch: Optional[int] = None
```
keszybz, line 25: It's nice to just set `self.source_date_epoch` at the end. It's not nice to set it to one value first and to a different one later.

mkosi/state.py
```diff
@@ -21,6 +21,16 @@ def __init__(self, args: MkosiArgs, config: MkosiConfig, workspace: Path) -> Non
...
+        if source_date_epoch is not None:
```
keszybz, line 26: Drop `is not None` here. An empty setting of SOURCE_DATE_EPOCH= should be treated the same as not setting it at all.

keszybz (changes requested):

mkosi/state.py
malt3: So we can check `len(source_date_epoch) > 0` instead?

keszybz: You *could*, but there is no reason to. The check `if source_date_epoch` is exactly what we need. It is equivalent to `if source_date_epoch is not None and source_date_epoch != ''`.

mkosi/__init__.py
```diff
@@ -661,6 +661,7 @@ def prepare_grub_config(state: MkosiState) -> Optional[Path]:
     if not config.exists():
         with umask(~0o600), config.open("w") as f:
             f.write("set timeout=0\n")
+            normalize_mtime( state.source_date_epoch, state.root, config)
```
behrmann, line 664:
```suggestion
            normalize_mtime(state.source_date_epoch, state.root, config)
```

mkosi/__init__.py
```diff
@@ -934,6 +944,8 @@ def build_initrd(state: MkosiState) -> Path:
         "--make-initrd", "yes",
         "--bootable", "no",
         "--manifest-format", "",
+        *([f"--environment=SOURCE_DATE_EPOCH={state.config.environment['SOURCE_DATE_EPOCH']}"]
+        if "SOURCE_DATE_EPOCH" in state.config.environment else []),
```
behrmann, line 948: I think @keszybz meant the off by one in the other direction.

mkosi/__init__.py
```diff
@@ -2255,3 +2270,27 @@ def run_verb(args: MkosiArgs, presets: Sequence[MkosiConfig]) -> None:
...
+
+def normalize_mtime(source_date_epoch: Optional[int], root: Path, target: Optional[Path] = None) -> None:
+    if source_date_epoch is None:
+        return
+
+    # If no target given, normalize the mtime recursively for the entire tree, starting at root.
+    if not target:
+        with complete_step("Setting mtime to SOURCE_DATE_EPOCH"):
+            os.utime(root, (source_date_epoch, source_date_epoch), follow_symlinks=False)
+            for p in root.rglob("*"):
```
behrmann, line 2283: The comment says recursively, but the single `*` will get you down at most one directory level, do you mean
```suggestion
            for p in root.rglob("**/*"):
```

mkosi/__init__.py
```diff
@@ -2255,3 +2270,27 @@ def run_verb(args: MkosiArgs, presets: Sequence[MkosiConfig]) -> None:
 
             if args.verb == Verb.serve:
                 run_serve(last)
+
+
+def normalize_mtime(source_date_epoch: Optional[int], root: Path, target: Optional[Path] = None) -> None:
+    if source_date_epoch is None:
+        return
+
+    # If no target given, normalize the mtime recursively for the entire tree, starting at root.
```
behrmann, line 2279:
```suggestion
    # If no is target given, normalize the mtime recursively for the entire tree, starting at root.
```

mkosi/__init__.py
katexochen: This is using [`rglob`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.rglob), not glob, so it works as described.

DaanDeMeyer (changes requested):
Hmm this seems to granular, we should just normalize the mtimes of /boot and /efi again before we call make_image() again instead of doing it for each invididual file

mkosi/__init__.py
```diff
@@ -2255,3 +2260,13 @@ def run_verb(args: MkosiArgs, presets: Sequence[MkosiConfig]) -> None:
 
             if args.verb == Verb.serve:
                 run_serve(last)
+
+
+def normalize_mtime(source_date_epoch: Optional[int], root: Path) -> None:
```
DaanDeMeyer, line 2265: Please pass `root` as the first argument for consistency with our other functions. Also pass in the directory in `root` to operate on as a separate argument. 

Also please rename the `source_date_epoch` argument to `mtime`

mkosi/__init__.py
```diff
@@ -2255,3 +2260,13 @@ def run_verb(args: MkosiArgs, presets: Sequence[MkosiConfig]) -> None:
...
+    if source_date_epoch is None:
+        return
+
+    with complete_step("Setting mtime to SOURCE_DATE_EPOCH"):
```
DaanDeMeyer, line 2269: What about "Normalizing modification times of {directory}"

DaanDeMeyer (changes requested):

mkosi/__init__.py
malt3: Agreed but with a relative path from root to directory

DaanDeMeyer: Exactly

DaanDeMeyer (changes requested):
Ah and maybe also modify load_environment() to pick up SOURCE_DATE_EPOCH from the environment mkosi is invoked from

DaanDeMeyer (changes requested):
Sorry for the many rounds, but I realized we should make this a separate setting so it doesn't have to be configured explicitly via the environment. So what I would suggest is the following:

1. Introduce a new setting `SourceDateEpoch`, it's parse function should be more or less identical to the logic you have in the `MkosiState` constructor now.
2. Add a `default_factory` function for `SourceDateEpoch` that initializes it from the configured environment (`Environment=`) or the host environment (in that order).
3. Instead of passing `state.source_date_epoch` to `normalize_mtime()`, pass `state.config.source_date_epoch` to `normalize_mtime()`.

mkosi/config.py
```diff
@@ -1000,13 +1022,23 @@ class MkosiConfigParser:
...
             help="Set the seed for systemd-repart",
         ),
+        MkosiConfigSetting(
+            dest="source_date_epoch",
+            long="--source-date-epoch",
+            metavar="TIMESTAMP",
+            name="SourceDateEpoch",
+            section="Output",
+            parse=config_parse_source_date_epoch,
+            default_factory=config_default_source_date_epoch,
```
DaanDeMeyer, line 1039: This should have `default_factory_depends=("environment",)`

mkosi/config.py
```diff
@@ -1000,13 +1022,23 @@ class MkosiConfigParser:
...
             section="Output",
             parse=config_parse_seed,
             help="Set the seed for systemd-repart",
         ),
+        MkosiConfigSetting(
+            dest="source_date_epoch",
+            long="--source-date-epoch",
+            metavar="TIMESTAMP",
+            name="SourceDateEpoch",
+            section="Output",
```
DaanDeMeyer, line 1037: I think `Content` is more fitting here.

mkosi/config.py
```diff
@@ -310,6 +323,14 @@ def config_default_mirror(namespace: argparse.Namespace) -> Optional[str]:
     return None
 
 
+def config_default_source_date_epoch(namespace: argparse.Namespace) -> Optional[int]:
+    config_env = getattr(namespace, "environment", [])
```
DaanDeMeyer, line 327: This variable can just be inlined, also no need for the default value if you add default_factory_depends, since that will ensure the variable is set.

mkosi/__init__.py
```diff
@@ -2255,3 +2260,13 @@ def run_verb(args: MkosiArgs, presets: Sequence[MkosiConfig]) -> None:
 
             if args.verb == Verb.serve:
                 run_serve(last)
+
+
+def normalize_mtime(root: Path, directory: Path, mtime: Optional[int]) -> None:
+    if mtime is None:
+        return
+
+    with complete_step(f"Normalizing modification times of {directory}"):
```
DaanDeMeyer, line 2269: Please add "/" here before the directory

mkosi/__init__.py
```diff
@@ -2255,3 +2260,13 @@ def run_verb(args: MkosiArgs, presets: Sequence[MkosiConfig]) -> None:
 
             if args.verb == Verb.serve:
                 run_serve(last)
+
+
+def normalize_mtime(root: Path, directory: Path, mtime: Optional[int]) -> None:
```
DaanDeMeyer, line 2265: Please make `directory` optional and default to `/` if it is `None`

DaanDeMeyer (changes requested):
Almost there

mkosi/__init__.py
malt3: I think this requires reordering the args again. Should we put mtime first?

DaanDeMeyer: I would always pass it by keyword

mkosi/__init__.py
```diff
@@ -934,6 +934,8 @@ def build_initrd(state: MkosiState) -> Path:
         "--make-initrd", "yes",
         "--bootable", "no",
         "--manifest-format", "",
+        *([f"--source-date-epoch={state.config.source_date_epoch}"]
```
DaanDeMeyer, line 937: Please use two separate arguments like we do for all the other options

mkosi/__init__.py
```diff
@@ -2255,3 +2260,14 @@ def run_verb(args: MkosiArgs, presets: Sequence[MkosiConfig]) -> None:
 
             if args.verb == Verb.serve:
                 run_serve(last)
+
+
+def normalize_mtime(root: Path, mtime: Optional[int], **kwargs: Any) -> None:
+    directory = kwargs.get("directory", Path(""))
+    if mtime is None:
+        return
```
DaanDeMeyer, line 2268: This should just be:

```
normalize_mtime(root: Path, mtime: Optional[int], directory: Optional[Path] = None) -> None:
    directory = directory or Path("")
```

mkosi/config.py
```diff
@@ -1093,6 +1114,17 @@ class MkosiConfigParser:
             parse=config_parse_feature,
             help="Remove package manager database and other files",
         ),
+        MkosiConfigSetting(
+            dest="source_date_epoch",
+            long="--source-date-epoch",
```
DaanDeMeyer, line 1119: This can be dropped, it is inferred automatically

mkosi/config.py
```diff
@@ -1093,6 +1114,17 @@ class MkosiConfigParser:
...
+            metavar="TIMESTAMP",
+            name="SourceDateEpoch",
```
DaanDeMeyer, line 1121: Same here, this is inferred automatically.

mkosi/config.py
```diff
@@ -1093,6 +1114,17 @@ class MkosiConfigParser:
...
+            section="Content",
+            parse=config_parse_source_date_epoch,
+            default_factory=config_default_source_date_epoch,
+            default_factory_depends=("environment",),
+            help="Set the SOURCE_DATE_EPOCH timestamp",
```
DaanDeMeyer, line 1126: "Set the $SOURCE_DATE_EPOCH timestamp"

mkosi/__init__.py
malt3: Thanks for the explanation

mkosi/__init__.py
```diff
@@ -934,6 +934,8 @@ def build_initrd(state: MkosiState) -> Path:
         "--make-initrd", "yes",
         "--bootable", "no",
         "--manifest-format", "",
+        *(["--source-date-epoch", state.config.source_date_epoch]
+                if state.config.source_date_epoch else []),
```
DaanDeMeyer, line 938: We should probably do `is not None` here since otherwise we won't pass on a value of zero when building the initrd.

mkosi/config.py
```diff
@@ -259,6 +260,18 @@ def config_parse_seed(value: Optional[str], old: Optional[str]) -> Optional[uuid
         die(f"{value} is not a valid UUID")
 
 
+def config_parse_source_date_epoch(value: Optional[str], old: Optional[int]) -> Optional[int]:
+    if not value:
+        return None
```
DaanDeMeyer, line 265: Same here, this should be `if value is None`

DaanDeMeyer (changes requested):
Found one more thing

DaanDeMeyer (approved):

malt3:
Stop. Do not merge

DaanDeMeyer:
Thanks for contributing!

PR #1840, opened 2023-08-28
config: Test last assignment wins with defaults

mcassaniti:
@DaanDeMeyer Please use this to test #1836. It's no good relying on the CI side of things yet because it will fail until #1836 lands. If the tests are incorrect then either update the PR directly as required or merge the tests into #1836.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1,6 +1,73 @@
 # SPDX-License-Identifier: LGPL-2.1+
 
-from mkosi.config import Compression
+import contextlib
+import tempfile
+from argparse import Namespace
+from os import chdir
+from pathlib import Path
+from typing import Iterator
+
+import pytest
+
+from mkosi.architecture import Architecture
+from mkosi.config import Compression, MkosiConfig, MkosiConfigParser, OutputFormat
+
+CONF_DIR = Path(__file__).parent.absolute() / "test-config"
+
+
+def parse_paths(paths: list[Path]) -> MkosiConfig:
+    """Process these paths stacked together"""
+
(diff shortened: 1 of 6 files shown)


mcassaniti:
@DaanDeMeyer I've successfully ran the test suite against #1836 and it is passing.

DaanDeMeyer (changes requested):
Can we just write the configuration files from the test itself in a temporary directory? I think that makes all of this a more flexible

mcassaniti:
> Can we just write the configuration files from the test itself in a temporary directory? I think that makes all of this a more flexible

It wouldn't be worth it. We're running most of the tests against a static set of files, then dynamically generating and changing configuration for the last two tests. At best a copy of the static configuration could be done to a temporary directory, but the writes of drop-in files will then likely be in order.

tests/test_config.py
```diff
@@ -1,6 +1,72 @@
...
+
+    configs = list()
+    for i in range(len(paths)):
+        configs.append(parse_paths(paths[:i+1]))
+
+    return configs
+
+
+@contextlib.contextmanager
+def cd_new_dir(path: Path) -> Iterator[None]:
```
behrmann, line 50: Add a todo, this can be dropped once we are on Python 3.11

mcassaniti:
I don't know what is going on with the CentOS disk test but I highly doubt adding a test suite will affect it.

behrmann:
Removing the CI fails label because the CentOS failure is unrelated.

behrmann:
Let's merge this and fix them later, since they have some benefit.

PR #1845, opened 2023-08-29
Rework ini file parsing

DaanDeMeyer:
Python's configparser module is rather inadequate for our purposes.
Specifically, it's built around every setting only being specified once
in a config file and even though we got it to kind of parse more than
one of the same setting with our own custom dict type, there's a lot of
limitations involved. Specifically, assigning the empty value is
impossible with the current approach. To avoid all these issues, let's
introduce our own ini parser.

This also splits off the matching logic into a new method match_config()
and moves setting the debugging level just after we've parsed CLI options
so we get debug messages from the parser as well.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -1649,6 +1649,15 @@ class MkosiConfigParser:
 
         parser.remove_section("Match")
 
+        for section in parser.sections():
+            for k, v in parser.items(section):
+                ns = defaults if k.startswith("@") else namespace
+
+                if not (s := self.settings_lookup_by_name.get(k.removeprefix("@"))):
+                    die(f"Unknown setting {k}")
+
+                setattr(ns, s.dest, s.parse(v, getattr(ns, s.dest, None)))
+
         if extras:
             for s in self.SETTINGS:
                 ns = defaults if s.path_default else namespace
@@ -1666,20 +1675,11 @@ class MkosiConfigParser:
                         setattr(ns, s.dest,
                                 s.parse(p.read_text() if s.path_read_text else f, getattr(ns, s.dest, None)))
 
-        for section in parser.sections():
-            for k, v in parser.items(section):
-                ns = defaults if k.startswith("@") else namespace
-
-                if not (s := self.settings_lookup_by_name.get(k.removeprefix("@"))):
-                    die(f"Unknown setting {k}")
-
-                setattr(ns, s.dest, s.parse(v, getattr(ns, s.dest, None)))
-
-        if extras and (path.parent / "mkosi.conf.d").exists():
-            for p in sorted((path.parent / "mkosi.conf.d").iterdir()):
-                if p.is_dir() or p.suffix == ".conf":
-                    with chdir(p if p.is_dir() else Path.cwd()):
-                        self.parse_config(p if p.is_file() else Path("."), namespace, defaults)
+            if (path.parent / "mkosi.conf.d").exists():
+                for p in sorted((path.parent / "mkosi.conf.d").iterdir()):
+                    if p.is_dir() or p.suffix == ".conf":
+                        with chdir(p if p.is_dir() else Path.cwd()):
+                            self.parse_config(p if p.is_file() else Path("."), namespace, defaults)
 
         return True
 
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -2,7 +2,6 @@
 
 import argparse
 import base64
-import configparser
 import copy
 import dataclasses
 import enum
@@ -22,7 +21,7 @@ import textwrap
 import uuid
 from collections.abc import Iterable, Sequence
 from pathlib import Path
-from typing import Any, Callable, Optional, Type, Union, cast
+from typing import Any, Callable, Iterator, Optional, Type, Union, cast
 
 from mkosi.architecture import Architecture
 from mkosi.distributions import Distribution, detect_distribution
@@ -226,7 +225,7 @@ def config_parse_boolean(value: Optional[str], old: Optional[bool]) -> bool:
 
 
 def parse_feature(value: Optional[str]) -> ConfigFeature:
-    if not value or value == ConfigFeature.auto.name:
+    if value is None or value == ConfigFeature.auto.name:
         return ConfigFeature.auto
 
     return ConfigFeature.enabled if parse_boolean(value) else ConfigFeature.disabled
@@ -241,7 +240,7 @@ def config_match_feature(match: Optional[str], value: Optional[ConfigFeature]) -
 
 
 def config_parse_compression(value: Optional[str], old: Optional[Compression]) -> Optional[Compression]:
-    if not value:
+    if value is None:
         return None
 
     try:
@@ -251,7 +250,7 @@ def config_parse_compression(value: Optional[str], old: Optional[Compression]) -
 
 
 def config_parse_seed(value: Optional[str], old: Optional[str]) -> Optional[uuid.UUID]:
-    if not value or value == "random":
+    if value is None or value == "random":
         return None
 
     try:
@@ -362,7 +361,7 @@ def config_make_list_parser(delimiter: str,
         new = old.copy() if old else []
 
         # Empty strings reset the list.
-        if not value:
+        if value is None:
             return []
 
         if unescape:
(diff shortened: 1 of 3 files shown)


mkosi/config.py
```diff
@@ -795,6 +794,57 @@ def cache_manifest(self) -> dict[str, Any]:
         return manifest
 
 
+def parse_ini(path: Path, sections: Sequence[str] = ()) -> Iterator[tuple[str, str, str]]:
+    section = setting = value = None
+
+    for l in textwrap.dedent(path.read_text()).splitlines():
+        comment = l.find("#")
```
behrmann, line 801: This allows inline comments, which configparser does not. This brings with it the can of worms of maybe needing to escape them. 

Inline comments would be nice, but maybe for now:
```python
if l.lstrip().startswith("#"):
    continue
```

Also, configparser does allow `;` as comment syntax. Don't know if anybody ever uses it, but not having is technically a compat break.

mkosi/config.py
```diff
@@ -795,6 +794,57 @@ def cache_manifest(self) -> dict[str, Any]:
...
+        # So the line is not indented, that means we either found a new section or a new setting. Either way,
+        # let's yield the previous setting and its value before parsing the new section/setting.
+        if section and setting and value is not None:
+            yield section, setting, value
+            setting = value = None
+
+        l = l.strip()
+
+        if l[0] == '[':
+            if l[-1] != ']':
```
behrmann, line 823:
```suggestion
        if l[0] == '[' and l[-1] != ']':
```

mkosi/config.py
```diff
@@ -795,6 +794,57 @@ def cache_manifest(self) -> dict[str, Any]:
...
+                die(f"{l} is not a valid section")
+
+            section = l[1:-1]
+            continue
+
+        if not section:
+            die(f"Setting {l} is located outside of section")
+
+        if sections and section not in sections:
```
behrmann, line 832: How about calling `sections` `only_sections` instead, right now this is a bit of a tongue twister to read.

mkosi/config.py
```diff
@@ -795,6 +794,57 @@ def cache_manifest(self) -> dict[str, Any]:
...
+        if delimiter == 0:
+            die(f"Missing setting name before '=' in {l}")
+
+        setting = l[0:delimiter].strip()
+        value = l[delimiter+1:].strip()
+
+    # Make sure we yield any final setting and its value.
+    if section and setting and value is not None:
+        yield section, setting, value
+
```
behrmann, line 847: Missing empty line

mkosi/config.py
DaanDeMeyer: Huh? This wouldn't be the same right?

mkosi/config.py
DaanDeMeyer: configparser seems to support inline comments just fine, since we're using that in quite a few places in systemd's, mkosi config

behrmann: My bad, did only a quick sanity check yesterday and
```python
c = configparser.ConfigParser()
c.read_string("""\
[DEFAULT]
foo=bar # comment
""")
c["DEFAULT"]["foo"]
```
will return the comment as part of the value, but that's because the default for `ConfigParser` is `inline_comment_prefixes=None`.

mkosi/config.py
```diff
@@ -795,6 +794,61 @@ def cache_manifest(self) -> dict[str, Any]:
         return manifest
 
 
+def parse_ini(path: Path, only_sections: Sequence[str] = ()) -> Iterator[tuple[str, str, str]]:
+    section: Optional[str] = None
+    setting: Optional[str] = None
+    value: Optional[str] = None
+
+    for l in textwrap.dedent(path.read_text()).splitlines():
+        for c in ("#", ";"):
```
behrmann, line 803: While checking the value for inline comments, I noticed
```python
configparser.ConfigParser(
            delimiters="=",
            comment_prefixes="#",
            inline_comment_prefixes="#",
            empty_lines_in_values=True,
            interpolation=None,
            strict=False,
            dict_type=ConfigParserMultipleValues,
        )
```
so for us `;` wasn't a comment prefix anymore. That being said, let's maybe add the above definition as a comment at the top of the function, so that it's clear what it's roughly supposed to mimic.

mkosi/config.py
```diff
@@ -795,6 +794,61 @@ def cache_manifest(self) -> dict[str, Any]:
...
+            continue
+
+        delimiter = l.find("=")
+        if delimiter < 0:
+            die(f"Setting {l} must be followed by '='")
+        if delimiter == 0:
+            die(f"Missing setting name before '=' in {l}")
+
+        setting = l[0:delimiter].strip()
+        value = l[delimiter+1:].strip()
```
behrmann, line 845: It's not shorter, but it eschews having to think about indices
```suggestion
        setting, delimiter, value = l.partition("=")
        if not delimiter:
            die(f"Setting {setting} must be followed by '='")
        if not setting:
            die(f"Missing setting name before '=' in {l}")

        setting = setting.strip()
        value = value.strip()
```

mkosi/config.py
```diff
@@ -795,6 +794,61 @@ def cache_manifest(self) -> dict[str, Any]:
...
+            yield section, setting, value
+            setting = value = None
+
+        l = l.strip()
+
+        if l[0] == '[':
+            if l[-1] != ']':
+                die(f"{l} is not a valid section")
+
+            section = l[1:-1]
```
behrmann, line 829:
```suggestion
            section = l[1:-1].strip()
```
otherwise all whitespace (but non-empty) section names are possible. With the strip the below error message would become wrong, though.

mkosi/config.py
```diff
@@ -1555,69 +1617,66 @@ def __init__(self) -> None:
...
 
-    def parse_config(self, path: Path, namespace: argparse.Namespace, defaults: argparse.Namespace) -> bool:
-        extras = path.is_dir()
+    def match_config(self, path: Path, namespace: argparse.Namespace, defaults: argparse.Namespace) -> bool:
+        triggered = None
 
-        if path.is_dir():
-            path = path / "mkosi.conf"
+        if not path.exists():
+            return True
```
behrmann, line 1624: This part of the logic could use a comment, because with the `not` at the place of use in `parse_config` this is a whole lot of inversions, that I find a bit confusing

mkosi/config.py
```diff
@@ -1555,69 +1617,66 @@ def __init__(self) -> None:
...
+                result = not result
+            if not trigger and not result:
+                return False
+            if trigger:
+                triggered = bool(triggered) or result
 
-                    if (s := self.settings_lookup_by_name.get(k)):
-                        if not (match := s.match):
-                            die(f"{k} cannot be used in [Match]")
+        return triggered is not False
```
behrmann, line 1658: Sine `triggered` with the default of `None` is falsey, the `bool` call can be omitted, since `triggered` is either `None`, `False` or `True`  and `bool(None)` will evaluate the same as `None` and for the bools themselves the calls are idempotent.

That being said, since all the the `triggered = triggered or result` is doing is oring in `result` and `triggered` will stay true if `result` is true at least once, this could be shortened to a
```python
if result:
    return True
```
and then `triggered` could be omitted as well and the return at the end could become a plain `return True`. If I didn't mangle anything up while playing around with it, this passes the tests on my end. All in all:
```python
            if result:
                return True

        return True
```
and `triggered = None` removed at the top.

mkosi/config.py
```diff
@@ -1555,69 +1617,66 @@ def __init__(self) -> None:
...
+            v = v.removeprefix("!")
 
-        parser.optionxform = lambda optionstr: optionstr # type: ignore
+            if not v:
+                die("Match value cannot be empty")
 
-        if path.exists():
-            parser.read(path)
+            if (s := self.settings_lookup_by_name.get(k)):
+                if not (match := s.match):
```
behrmann, line 1636: here I'd actually not use the walrus operator, since the use of match is a bit down with something in between and I had to think to remember that they are the same.
```suggestion
                match = s.match
                if not match:
```

mkosi/config.py
DaanDeMeyer: I don't think this works, we need to check if all non-triggering conditions match, we can't leave the loop early.

Issue #1848, opened 2023-08-30
`mkosi ssh` does not work on Fedora host


septatrix:
When invoking the ssh command on a Fedora 38 host I get the following error message (with `--debug`):

```
‣ + ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -o ProxyCommand=socat - VSOCK-CONNECT:3436578145:%p root@mkosi
Bad owner or permissions on /etc/ssh/ssh_config.d/50-redhat.conf
```

Running the shown ssh command from the terminal itself works without problems.

---

PS: It would also be great if the debug log would quote command parts if they contain spaces. In this case that would be: `ssh ... -o "ProxyCommand=socat - VSOCK-CONNECT:3436578145:%p" ...`

PR #1849, opened 2023-08-31
Use shlex.join in more places

behrmann:
Addresses part of #1848.

### DIFF
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -163,10 +163,10 @@ def run(
     cwd: Optional[Path] = None,
     log: bool = True,
 ) -> CompletedProcess:
-    if ARG_DEBUG.get():
-        logging.info(f"+ {' '.join(str(s) for s in cmdline)}")
-
     cmdline = [os.fspath(x) for x in cmdline]
+
+    if ARG_DEBUG.get():
+        logging.info(f"+ {shlex.join(cmdline)}")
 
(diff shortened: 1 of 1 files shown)


DaanDeMeyer (changes requested):
shlex.join() is overkill for this, it does really weird stuff when parentheses are used in package names and such. We should probably just have our own function that just puts quotes around arguments with whitespace in them

behrmann:
Just looked at the code for `shlex.quote` and it is only
```python
def quote(s):
    """Return a shell-escaped version of the string *s*."""
    if not s:
        return "''"
    if _find_unsafe(s) is None:
        return s

    # use single quotes, and put single quotes into double quotes
    # the string $'b is then quoted as '$'"'"'b'
    return "'" + s.replace("'", "'\"'\"'") + "'"
```

and I'm not sure what you mean by the parens stuff:
```
>>> shlex.join(["foo", "baz bar", "(qux)", "log", "{qwerty}", "rhubarb[rainbow]"])
"foo 'baz bar' '(qux)' log '{qwerty}' 'rhubarb[rainbow]'"
```

DaanDeMeyer:
I vaguely remember the systemd dnf command lines looking extremely weird when we used shlex.join() but OK, let's try this again

DaanDeMeyer (approved):

PR #1850, opened 2023-08-31
Ignore ssh configuration files when running ssh

### LINKED ISSUE #1848: `mkosi ssh` does not work on Fedora host

DaanDeMeyer:
When in a user namespace, the ssh config files from /etc/ will be owned by nobody which makes ssh fail. Let's ignore these configuration files when running ssh to avoid such failures.

Fixes #1848

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -336,6 +336,7 @@ def run_qemu(args: MkosiArgs, config: MkosiConfig) -> None:
 def run_ssh(args: MkosiArgs, config: MkosiConfig) -> None:
     cmd = [
         "ssh",
+        "-F", "none",
         # Silence known hosts file errors/warnings.
         "-o", "UserKnownHostsFile=/dev/null",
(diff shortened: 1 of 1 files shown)


septatrix (approved):
I just applied this to my local 15.1 installation and it works. Thanks

behrmann (approved):

PR #1854, opened 2023-09-01
Replace ! operator for lists with empty string assignment

DaanDeMeyer:
Let's behave more like systemd and replace the `!` operator for
removing values from lists with assigning the empty string to remove
all values instead.

Also includes a bunch of commits cleaning up the unit tests


behrmann (approved):
Big diff. I've looked at it with and without whitespace and I think it should be ok.

PR #1856, opened 2023-09-02
ci: enable jammy-proposed-updates to test new libzypp

bluca:
I got some patches backported to jammy to try and fix https issues, enable proposed-updates so that it can be tested before moving to the release stage

As mentioned on https://bugs.launchpad.net/ubuntu/+source/libzypp/+bug/2025400 I need to show that the patched libzypp works fine, so let's enable proposed-updates on jammy for a while

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,6 +107,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
+    - name: Enable proposed-updates
+      run: |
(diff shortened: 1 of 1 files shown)


DaanDeMeyer (approved):

DaanDeMeyer:
You'll also need to do a patch to have zypper actually use https

bluca:
> You'll also need to do a patch to have zypper actually use https

Yep - plan is to leave it like this for a couple of days to ensure no regressions on http, then enable https later

Issue #1760, opened 2023-08-08
Support separate download verb to prepopulate caches


DaanDeMeyer:
We should support a download verb that makes sure all necessary packages and software are downloaded without building the image. This download verb would likely be implemented by having it prepopulate the mkosi cache which is then automatically picked up when we do the actual image build. Optionally, the download verb should take a mkosi manifest and download the specific package versions listed in there. Ideally the download verb also integrates with prepare scripts to allow those to download whatever they need to as well.

gsegatti:
Curious as to what the implementation would potentially look like, in its simplest form. Would it be basically to run the [`if not cached`](https://github.com/gsegatti/mkosi/blob/00a16c33f8862bc77115d247de02014b707ecc87/mkosi/__init__.py#L1389-L1396) portion of `build_image()`, for whatever distribution specified?
 
 Or Am I way out of my depth here?

keszybz:
I guess that for `dnf`, we'd essentially want to do a normal transaction, but break it off after packages have been downloaded. I don't think `dnf` has an option to do that. There is a `system-upgrade download` plugin that does that for dnf4. Maybe we should talk with the dnf folks about how to implement this best.

DaanDeMeyer:
Let's close this in favor of #1112. I'm not sure anymore that a `download` verb is the way to go

PR #1857, opened 2023-09-02
[WIP] Adding new Download verb

gsegatti:
This is an initial attempt to solve issue #1760.

The idea here is to provide the minimal set of changes which would permit to populate our cache with a new verb. This way we can (in the near future) implement taking mkosi manifestos and so on as explained on the issue.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2116,6 +2116,10 @@ def show_docs(args: MkosiArgs) -> None:
                 raise e
 
 
+def download_packages(args: MkosiArgs, config: MkosiConfig) -> None:
+    pass # TODO: implement this.
+
+
 def expand_specifier(s: str) -> str:
     return s.replace("%u", InvokingUser.name())
 
@@ -2152,6 +2156,11 @@ def run_verb(args: MkosiArgs, presets: Sequence[MkosiConfig]) -> None:
 
     if args.verb == Verb.documentation:
         return show_docs(args)
+
+    if args.verb == Verb.download:
+        for config in presets:
+            download_packages(args, config)
+        return
 
     if args.verb == Verb.genkey:
         return generate_key_cert_pair(args)
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2117,7 +2117,13 @@ def show_docs(args: MkosiArgs) -> None:
 
 
 def download_packages(args: MkosiArgs, config: MkosiConfig) -> None:
-    pass # TODO: implement this.
+    # TODO: See the best way of implementing this.
+    # For now I'll leave it like this for testing purposes.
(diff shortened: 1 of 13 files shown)


mkosi/__init__.py
```diff
@@ -2170,6 +2173,14 @@ def run_verb(args: MkosiArgs, presets: Sequence[MkosiConfig]) -> None:
 
     last = presets[-1]
 
+    if args.verb == Verb.download:
+        workspace = tempfile.TemporaryDirectory(dir=last.workspace_dir, prefix=".mkosi-tmp")
```
gsegatti, line 2177: I'm necessarily using `last` here, for the sake of simplicity while testing. But from my understanding we can receive a list (or `Sequence`) of `MkosiConfig`.

Each config can carry it's own distribution. So perhaps we should, for each `preset`, download build cache of its distribution?

mkosi/__init__.py
```diff
@@ -2170,6 +2173,14 @@ def run_verb(args: MkosiArgs, presets: Sequence[MkosiConfig]) -> None:
...
+        with workspace, scopedenv({"TMPDIR" : workspace.name}):
```
gsegatti, line 2178: Also this `with` statement is duplicated here and in `build_images()`. Maybe we should opt for a function with the context manager decorator for this (assuming this is even the correct way to go)?

DaanDeMeyer:
Each package manager should have a way to just download packages without installing them. That's what we should use to implement this functionality. Also this should be done for each preset. So loop over each preset and download the specified packages and build packages. And then later we can extend that by passing in the packages from the manifest to download specific versions of packages.

gsegatti:
Hey Daan!
> Also this should be done for each preset. So loop over each preset and download the specified packages and build packages. 

Appreciate you clarifying this, I thought so.

> Each package manager should have a way to just download packages without installing them.

I see, I tracked this down to the `DistributionInstaller` class for each distribution. Do you reckon we should implement a new Class Method `download_packages()` or should adapt one of the existing methods? I think a new one would be nice for readability/cleanliness, but let me know what you guys think.

mkosi/__init__.py
gsegatti: Marking this as resolved since Daan clarified we should loop over all presets.

DaanDeMeyer:
A new method seems like the way to go I think.

DaanDeMeyer:
We should probably have manifest support from the start in this because I'm not entirely sure whether the way we currently do things would work with the distribution provided archives to download older packages from. For Arch Linux we might need to do things quite differently than how we usually do things.

gsegatti:
It's been sometime since I last touched this, I was in the middle of moving to a different city. If what I've done so far isn't what we hoped towards implementing the `Download` verb, maybe it is best to close this PR for now?

DaanDeMeyer:
Frankly I'm not entirely sure myself, it might be different depending on the distribution. Let's close this one for now, but feel free to open a new PR if you want

PR #1858, opened 2023-09-03
Allow using a default tools tree

DaanDeMeyer:
Instead of requiring users to always set up a tools tree preset, let's
allow specifying "default" as the tools tree to have mkosi build a default
tools tree itself. This default tools tree includes all the software that
might be necessary to build an image (excluding software that might be
required by various user scripts).

For distributions that do not have a rolling release variant, we use the
closest matching distribution. e.g. for CentOS, we use Fedora and for
Ubuntu we use Debian.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2020,7 +2020,7 @@ def run_serve(config: MkosiConfig) -> None:
         os.chdir(config.output_dir)
 
     with http.server.HTTPServer(("", port), http.server.SimpleHTTPRequestHandler) as httpd:
-        print(f"Serving HTTP on port {port}: http://localhost:{port}/")
+        logging.info(f"Serving HTTP on port {port}: http://localhost:{port}/")
         httpd.serve_forever()
 
 
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1075,8 +1075,9 @@ def install_unified_kernel(state: MkosiState, partitions: Sequence[Partition]) -
                 "--efi-arch", state.config.architecture.to_efi(),
             ]
 
-            for p in state.config.extra_search_paths:
-                cmd += ["--tools", p]
+            if not state.config.tools_tree:
+                for p in state.config.extra_search_paths:
+                    cmd += ["--tools", p]
 
             uki_config = state.pkgmngr / "etc/kernel/uki.conf"
             if uki_config.exists():
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -49,7 +49,7 @@ from mkosi.qemu import copy_ephemeral, run_qemu, run_ssh
 from mkosi.run import become_root, bwrap, chroot_cmd, init_mount_namespace, run
 from mkosi.state import MkosiState
 from mkosi.tree import copy_tree, install_tree, move_tree, rmtree
-from mkosi.types import PathString
+from mkosi.types import _FILE, CompletedProcess, PathString
 from mkosi.util import (
     InvokingUser,
     flatten,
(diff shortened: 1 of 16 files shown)


mkosi/distributions/__init__.py
```diff
@@ -50,6 +50,18 @@ def architecture(arch: Architecture) -> str:
     def package_type(cls) -> PackageType:
         raise NotImplementedError()
 
+    @classmethod
+    def default_release(cls) -> str:
+        raise NotImplementedError()
```
behrmann, line 55: This should be in the previous commit.

mkosi/config.py
```diff
@@ -290,18 +290,7 @@ def config_default_release(namespace: argparse.Namespace) -> str:
...
-        Distribution.rocky: "9",
-        Distribution.alma: "9",
-        Distribution.mageia: "cauldron",
-        Distribution.debian: "testing",
-        Distribution.ubuntu: "lunar",
-        Distribution.opensuse: "tumbleweed",
-        Distribution.openmandriva: "cooker",
-        Distribution.gentoo: "17.1",
-    }.get(namespace.distribution, "rolling")
+    return cast(str, namespace.distribution.default_release())
```
behrmann, line 293: The cast should be in the previous commit, too.

mkosi/distributions/__init__.py
```diff
@@ -101,6 +113,15 @@ def architecture(self, arch: Architecture) -> str:
     def package_type(self) -> PackageType:
         return self.installer().package_type()
 
+    def default_release(self) -> str:
+        return self.installer().default_release()
```
behrmann, line 117: And this, too.

mkosi/__init__.py
```diff
@@ -946,8 +948,7 @@ def build_initrd(state: MkosiState) -> Path:
     ]
 
     with complete_step("Building initrd"):
-        args, presets = parse_config(cmdline)
-        config = presets[0]
+        args, [config] = parse_config(cmdline)
```
behrmann, line 951: I'm a bit surprised this type checks, since this would throw a `ValueError` if `presets` had more than one element. Throwing away everything but the first element would be
```suggestion
        args, [config, *_] = parse_config(cmdline)
```
but apparently mypy can prove that there will be only one element?

mkosi/config.py
```diff
@@ -488,6 +477,16 @@ def config_parse_root_password(value: Optional[str], old: Optional[tuple[str, bo
     return (value, hashed)
 
 
+def config_parse_tools_tree(value: Optional[str], old: Union[Path, None, Literal["default"]]) -> Union[Path, None, Literal["default"]]:
```
behrmann, line 480: `Union[Path, None, Literal["default"]]` is a bit of a mouthful, how about defining a type alias `OptionalOrDefaultPath` (or similar)?

behrmann (commented):
This is an excellent change! What do you think about adding an option to pull in systemd from main on top of the default tools tree? This would make mkosi pretty much self-contained and could be used in the GH action as well.

mkosi/__init__.py
DaanDeMeyer: I guess so, I wouldn't complicate things if it type checks, since we know there will only be a single config

mkosi/config.py
DaanDeMeyer: New revision drops the `Union` stuff altogether

DaanDeMeyer:
> This is an excellent change! What do you think about adding an option to pull in systemd from main on top of the default tools tree? This would make mkosi pretty much self-contained and could be used in the GH action as well.

All of the default tools tree distributions except opensuse tumbleweed have v254 already which is sufficient. I wouldn't start building systemd from source for the default tools trees just yet.

behrmann:
> > This is an excellent change! What do you think about adding an option to pull in systemd from main on top of the default tools tree? This would make mkosi pretty much self-contained and could be used in the GH action as well.
> 
> All of the default tools tree distributions except opensuse tumbleweed have v254 already which is sufficient. I wouldn't start building systemd from source for the default tools trees just yet.

For Debian this implies building at least a trixie (i.e. testing) tool tree, though, right?

behrmann (approved):

PR #1874, opened 2023-09-06
Install ca-certificates in tools tree and make sure it's used

DaanDeMeyer:
Too old ca-certificates on the host can cause package manager failures when using a tools tree, so let's install ca-certificates into the tools tree and make sure it is used.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -43,7 +43,7 @@ from mkosi.installer import clean_package_manager_metadata, package_manager_scri
 from mkosi.kmod import gen_required_kernel_modules, process_kernel_modules
 from mkosi.log import ARG_DEBUG, complete_step, die, log_step
 from mkosi.manifest import Manifest
-from mkosi.mounts import mount_overlay, mount_passwd, mount_usr
+from mkosi.mounts import mount, mount_overlay, mount_passwd, mount_usr
 from mkosi.pager import page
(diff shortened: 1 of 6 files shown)


mkosi/__init__.py
```diff
@@ -943,6 +943,7 @@ def build_initrd(state: MkosiState) -> Path:
         *(["--timezone", state.config.timezone] if state.config.timezone else []),
         *(["--hostname", state.config.hostname] if state.config.hostname else []),
         *(["--root-password", rootpwopt] if rootpwopt else []),
+        *([f"--environment={k}=\"{v}\"" for k, v in state.config.environment.items()]),
```
behrmann, line 946: nit
```suggestion
        *([f"--environment={k}='{v}'" for k, v in state.config.environment.items()]),
```
Not needing escaping is the one reason for single quotes ;)

mkosi/__init__.py
```diff
@@ -2176,6 +2177,7 @@ def finalize_tools(args: MkosiArgs, presets: Sequence[MkosiConfig]) -> Sequence[
             "--bootable", "no",
             "--manifest-format", "",
             *(["--source-date-epoch", str(p.source_date_epoch)] if p.source_date_epoch is not None else []),
+            *([f"--environment={k}=\"{v}\"" for k, v in p.environment.items()]),
```
behrmann, line 2180: Same.

behrmann (approved):

PR #1875, opened 2023-09-07
If python3 version is too old, use python3.9 if it exists

DaanDeMeyer:


bin/mkosi
```diff
@@ -3,4 +3,15 @@
 set -e
 PYTHONPATH="$(dirname "$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")")"
 export PYTHONPATH
+
+if [ -z "$MKOSI_INTERPRETER" ]; then
+    # Note the check seems to be inversed here because the if branch is executed when the exit status is 0
```
behrmann, line 8:
```suggestion
    # Note the check seems to be inverted here because the if branch is executed when the exit status is 0
```

bin/mkosi
```diff
@@ -3,4 +3,15 @@
...
+    # which is equal to `False` in python.
+    if python3 -c "import sys; sys.exit(sys.version_info < (3, 9))"; then
+        MKOSI_INTERPRETER=python3
+    elif command -v python3.9 >/dev/null; then
+        MKOSI_INTERPRETER=python3.9
+    fi
+fi
+
 exec ${MKOSI_INTERPRETER:-python3} -B -m mkosi "$@"
```
behrmann, line 17: The first branch will always set `MKOSI_INTERPRETER` to `python3` if it is sufficiently new. If it isn't an 3.9 exists, it uses that, but this leaves the case where `python3` is not sufficently new and 3.9 doesn't exist, then this line would fall back to `python3` and mkosi would crash. Let's remove the fallback for the variable
```suggestion
exec $MKOSI_INTERPRETER -B -m mkosi "$@"
```
and add an `else` branch with an explicit error for that case.

The fallback `${MKOSI_INTERPRETER:-python3}` is not needed anymore because this now sets `MKOSI_INTERPRETER` if it was zero to begin with and if it weren't the fallback wouldn't trigger either.

bin/mkosi
```diff
@@ -3,4 +3,15 @@
 set -e
 PYTHONPATH="$(dirname "$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")")"
 export PYTHONPATH
+
+if [ -z "$MKOSI_INTERPRETER" ]; then
+    # Note the check seems to be inversed here because the if branch is executed when the exit status is 0
+    # which is equal to `False` in python.
```
keszybz, line 9: Please don't use backticks. This is not markdown.

keszybz: Also still "inversed".

keszybz:
LGTM.

PR #1881, opened 2023-09-07
Support building unified kernel images

DaanDeMeyer:
All the necessary pieces for this already exists, so let's add this as an additional output format.

### DIFF
--- a/.github/mkosi.conf.d/10-common.conf
+++ b/.github/mkosi.conf.d/10-common.conf
@@ -6,7 +6,6 @@ KernelCommandLine=console=ttyS0
                   systemd.default_standard_output=journal+console
 
 [Content]
-Bootable=yes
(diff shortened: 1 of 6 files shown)


behrmann (approved):
LGTM, but should this maybe be covered by CI boot tests as well?

poettering:
btw, I think this should be communicated differently. because mkosi has been building ukis since a longer time: as part of building regular disk images. 

The novelty here really is that the UKI is the pimary output and an OS image is embedded into it.

To the casual beholder this will not be clear I figure (I am guessing, because it wasn#t clear to me personally at first). So maybe in the mkosi.md markdown say: "…`uki` (a unified kernel image with the OS image embedded in the `.initrd` section)…"

poettering:
→ #1884

PR #1896, opened 2023-09-12
Add ruff config to `pyprojects.toml` and address its warnings

behrmann:
This also add the pyupgrade and pycodestyle rules in addition and fixes their warnings. These are mostly long line warnings and are fixed by reintroducing the 119 character line limit that we had in the past, but which got lost when we moved away from black, but existed before that as well.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -19,7 +19,7 @@ import textwrap
 import uuid
 from collections.abc import Iterator, Sequence
 from pathlib import Path
-from typing import ContextManager, Optional, TextIO, Union
+from typing import Optional, TextIO, Union
 
 from mkosi.archive import extract_tar, make_cpio, make_tar
 from mkosi.config import (
@@ -151,7 +151,9 @@ def install_build_packages(state: MkosiState) -> None:
     if not need_build_packages(state.config):
         return
 
-    with complete_step(f"Installing build packages for {str(state.config.distribution).capitalize()}"), mount_build_overlay(state):
+    # TODO: move to parenthesised context managers once on 3.10
+    pd = str(state.config.distribution).capitalize()
+    with complete_step(f"Installing build packages for {pd}"), mount_build_overlay(state):
         state.config.distribution.install_packages(state, state.config.build_packages)
 
 
@@ -196,7 +198,7 @@ def mount_cache_overlay(state: MkosiState) -> Iterator[None]:
         yield
 
 
-def mount_build_overlay(state: MkosiState, read_only: bool = False) -> ContextManager[Path]:
+def mount_build_overlay(state: MkosiState, read_only: bool = False) -> contextlib.AbstractContextManager[Path]:
     d = state.workspace / "build-overlay"
     if not d.is_symlink():
         with umask(~0o755):
@@ -559,7 +561,7 @@ def find_grub_bios_directory(state: MkosiState) -> Optional[Path]:
 def find_grub_binary(state: MkosiState, binary: str) -> Optional[Path]:
     path = ":".join(os.fspath(p) for p in [state.root / "usr/bin", state.root / "usr/sbin"])
 
-    assert "grub" in binary and not "grub2" in binary
+    assert "grub" in binary and "grub2" not in binary
 
     path = shutil.which(binary, path=path) or shutil.which(binary.replace("grub", "grub2"), path=path)
     if not path:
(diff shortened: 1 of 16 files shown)


mkosi/__init__.py
```diff
@@ -702,6 +704,15 @@ def prepare_grub_bios(state: MkosiState, partitions: Sequence[Partition]) -> Non
...
     initrd = Path(shutil.copy2(initrd, dst / "initrd"))
 
+    grubsnippet = textwrap.dedent(
+        """\
+        menuentry "{distribution}-{kver}" {{
+        linux /{kimg} {root} {cmdline}
+        initrd /{initrd} /{kmods}
+        }}
+        """
+    )
```
DaanDeMeyer, line 714: I'd prefer to avoid using the formatting stuff, can't we just introduce some extra variables here instead and use those to satisfy the line limit?

mkosi/config.py
```diff
@@ -2308,9 +2322,10 @@ def line_join_source_target_list(array: Sequence[tuple[Path, Optional[Path]]]) -
 
 
 def summary(args: MkosiArgs, config: MkosiConfig) -> str:
-    b = Style.bold
-    e = Style.reset
-    bold: Callable[..., str] = lambda s: f"{b}{s}{e}"
+    def bold(s: Any) -> str:
+        b = Style.bold
+        e = Style.reset
```
DaanDeMeyer, line 2327: Is there any point in still having these at all? just inline them?

pyproject.toml
```diff
@@ -68,3 +68,8 @@ strict_equality = true
 [[tool.mypy.overrides]]
 module = ["argcomplete"]
 ignore_missing_imports = true
+
+[tool.ruff]
+target-version = "py39"
+line-length = 119
```
DaanDeMeyer, line 74: I think we want 109 which is what systemd uses as well

DaanDeMeyer (changes requested):

mkosi/__init__.py
behrmann: I'll have a look

mkosi/config.py
behrmann: Will change it like that.

pyproject.toml
behrmann: I'm open to that. The first iteration of this actually had 109, but I changed it to 119, since that was the previous max for black and flake8 before that. It's just many more instances and some are a bit awkward.

DaanDeMeyer: Right let's keep this for now then, we can always lower it later

mkosi/__init__.py
behrmann: Changed it like that. In this case I prefer calling format explicitly, since it doesn't leak all the new variable names into the surrounding scope. I think it should be fine if tests pass, since the reused variable names should be the last point of use in each case.

mkosi/kmod.py
```diff
@@ -11,7 +11,9 @@
 from mkosi.run import bwrap, chroot_cmd
 
 
-def filter_kernel_modules(root: Path, kver: str, include: Sequence[str], exclude: Sequence[str]) -> list[Path]:
+def filter_kernel_modules(
+    root: Path, kver: str, include: Sequence[str], exclude: Sequence[str]
```
DaanDeMeyer, line 15: Single line per argument here and elsewhere?

pyproject.toml
behrmann: Dropped the commit again.

mkosi/kmod.py
behrmann: I'd prefer to keep it this way, since black only moves to single line per argument if the line length goes over the column limit. In the hope that we can one day go back to an autoformatter, I'd like to not introduce additional churn here.

DaanDeMeyer: This is inconsistent with how we do it already though, I'd prefer us to use single argument per line everywhere as soon as we have to start wrapping things

behrmann: Not necessary anymore with the limit at 119.

behrmann:
Split the commit up into three, one adding the ruff config, one addressing the warnings and one adding it to CI. I have no idea why the same exact changes that checked fine with this version of ruff in CI and do so locally have an issue in GA, but if push comes to shove adding it to CI can be postponed, since running ruff locally will address everything isort and pyflakes complain about in CI (and then some) while being easier to install.

DaanDeMeyer (approved):

PR #1900, opened 2023-09-15
Help and readme tweaks

keszybz:


README.md
```diff
@@ -83,7 +83,7 @@ binary and is not to be considered a public API.
 * [Primary mkosi git repository on GitHub](https://github.com/systemd/mkosi/)
 * [mkosi — A Tool for Generating OS Images](http://0pointer.net/blog/mkosi-a-tool-for-generating-os-images.html) introductory blog post by Lennart Poettering (2017)
 * [The mkosi OS generation tool](https://lwn.net/Articles/726655/) story on LWN (2017)
-* [mkosi: Building Bespoke Operating System Images](https://media.ccc.de/v/all-systems-go-2023-190-mkosi-building-bespoke-operating-system-images) talk at All Systems Go! 2023
+* [systemd-repart: Building Discoverable Disk Images](https://media.ccc.de/v/all-systems-go-2023-191-systemd-repart-building-discoverable-disk-images) and [mkosi: Building Bespoke Operating System Images](https://media.ccc.de/v/all-systems-go-2023-190-mkosi-building-bespoke-operating-system-images) talks at All Systems Go! 2023
```
behrmann, line 86: I'd switch the order and go with the mkosi talk first. The repart talk is a good addition in case people want to change anything around the partitioning.

behrmann (commented):
LGTM all in all, but I think the talks would be better the other way 'round.

mkosi/config.py
```diff
@@ -456,7 +456,7 @@ def config_parse_filename(value: Optional[str], old: Optional[str]) -> Optional[
         die(". and .. are not valid filenames")
 
     if "/" in value:
-        die(f"{value} is not a valid filename")
+        die(f"{value!r} is not a valid filename. (Output= requires a filename with no path components, relative to output directory.)")
```
DaanDeMeyer, line 459: We should probably start passing in the setting name (either CLI switch or config file setting) to the parser functions. Then again though, we'd likely also want the file we're parsing and the line so maybe it's better to start throwing more recognizable exceptions which we catch in parse_config() and then show a proper error message that includes more information.

PR #1901, opened 2023-09-15
Precompute long options

gdiscry:
Like the `name` attribute used for the INI files, precompute the `long` attribute used by argparse when it is derived from a setting's `dest`.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -504,7 +504,7 @@ class MkosiConfigSetting:
 
     # settings for argparse
     short: Optional[str] = None
-    long: Optional[str] = None
+    long: str = ""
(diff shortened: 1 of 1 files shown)


behrmann:
This is missing a bit of context. The idea is to do this because every ini option has a long option on the CLI?

gdiscry:
Yes sorry for the lack of context. Indeed, `create_argument_parser` assigns a long option to each setting and used to derive it from `dest` if one wasn't explicitly defined.

I stumbled upon this while I was working on the script lists, but it didn't matter for that PR in the end. So, it's not (currently) an issue and more for uniformity with how `name` is handled.

behrmann (approved):

PR #1902, opened 2023-09-15
Enforce sections when parsing INI files

gdiscry:
Except for the `[Match]` section, a setting could be set in any of the valid sections because the actual section was ignored.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -1976,11 +1976,14 @@ def parse_config(argv: Sequence[str] = ()) -> tuple[MkosiArgs, tuple[MkosiConfig
         if path.exists():
             logging.debug(f"Including configuration file {Path.cwd() / path}")
 
-            for _, k, v in parse_ini(path, only_sections=["Distribution", "Output", "Content", "Validation", "Host", "Preset"]):
+            for section, k, v in parse_ini(path, only_sections=["Distribution", "Output", "Content", "Validation", "Host", "Preset"]):
                 ns = defaults if k.startswith("@") else namespace
 
                 if not (s := settings_lookup_by_name.get(k.removeprefix("@"))):
                     die(f"Unknown setting {k}")
+
+                if section != s.section:
+                    logging.warning(f"Setting {k} should be configured in [{s.section}], not [{section}].")
 
                 setattr(ns, s.dest, s.parse(v, getattr(ns, s.dest, None)))
 
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -2,6 +2,7 @@
 
 import argparse
 import itertools
+import logging
 import operator
 from pathlib import Path
 from typing import Optional
@@ -148,6 +149,8 @@ def test_parse_config(tmp_path: Path) -> None:
         """\
         [Content]
         Packages=
+
+        [Output]
         ImageId=
         """
     )
@@ -362,6 +365,8 @@ def test_match_imageid(tmp_path: Path, image1: str, image2: str) -> None:
             f"""\
             [Distribution]
             Distribution=fedora
+
+            [Output]
             ImageId={image1}
             """
         )
@@ -442,6 +447,8 @@ def test_match_imageversion(tmp_path: Path, op: str, version: str) -> None:
             """\
             [Distribution]
             Distribution=fedora
+
+            [Output]
             ImageId=testimage
             ImageVersion=123
             """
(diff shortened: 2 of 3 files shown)


mkosi/config.py
```diff
@@ -1928,10 +1937,8 @@ def match_config(path: Path, namespace: argparse.Namespace, defaults: argparse.N
             if not v:
                 die("Match value cannot be empty")
 
-            if (s := settings_lookup_by_name.get(k)):
-                if not s.match:
-                    die(f"{k} cannot be used in [Match]")
-
+            if (s := settings_lookup_by_name.get((section, k))):
+                assert s.match
```
behrmann, line 1941: Why downgrade this to an `assert`? `assert`s can be removed from Python code easily, just by setting an environment variable. Preferably `assert`s shouldn't be used outside of tests in most cases.

behrmann (changes requested):
This is a good change, thanks!

Could you add an explicit test to `test_config` that catches an explicit fail when something is on the wrong section?

mkosi/config.py
gdiscry: That's the point actually, the `assert` is only there to let the type checker know that `s.match` is not `None`.

We know it's not `None` because we are only parsing the `[Match]` section here, so `section == "Match"`. All the keys in `settings_lookup_by_name` with a `"Match"` section are there because the setting has a `match` function (see `iter_sections()`).

DaanDeMeyer:
So we're trying to be a bit more conscious about backwards compat. Can we downgrade this to warnings and then do a few releases with warnings before turning it into a hard error?

gdiscry:
I'll do that. Technically, fixing a configuration so that the settings are in the right section doesn't break the compatibility with an older version, but I agree that a warning is more user-friendly.

Even the tests and CI didn't use the correct sections.

mkosi/config.py
gdiscry: I restored the old code when downgrading to warnings. And even if we turn it back in a hard error, the fuzzy lookup might still be useful for the message.

gdiscry:
I downgraded the error to a warning with a hint about what to fix and how to fix it.

I also added a test parametrized over a few scenarios, such as:

- no warning
- multiple warnings
- warning even if the setting is also set in the valid section or on the command-line.

mkosi/config.py
```diff
@@ -1976,12 +1976,15 @@ def parse_config( path: Path, namespace: argparse.Namespace, defaults: argparse.
...
 
-            for _, k, v in parse_ini(path, only_sections=["Distribution", "Output", "Content", "Validation", "Host", "Preset"]):
+            for section, k, v in parse_ini(path, only_sections=["Distribution", "Output", "Content", "Validation", "Host", "Preset"]):
                 ns = defaults if k.startswith("@") else namespace
 
                 if not (s := settings_lookup_by_name.get(k.removeprefix("@"))):
                     die(f"Unknown setting {k}")
 
+                if section != s.section:
+                    logging.warning(f"Setting {k} should be configure in [{s.section}], not [{section}].")
```
DaanDeMeyer, line 1986: "should be configured"

tests/test_config.py
```diff
@@ -440,7 +445,7 @@ def test_match_imageversion(tmp_path: Path, op: str, version: str) -> None:
         parent = Path("mkosi.conf")
         parent.write_text(
             """\
-            [Distribution]
+            [Output]
```
DaanDeMeyer, line 448: This is wrong no? Distribution should be under [Distribution]?

tests/test_config.py
```diff
@@ -523,3 +528,39 @@ def test_package_manager_tree(tmp_path: Path, skel: Optional[Path], pkgmngr: Opt
...
+    tmp_path: Path,
+    caplog: pytest.LogCaptureFixture,
+    sections: list[str],
+    args: list[str],
+    warning_count: int,
+) -> None:
+    with chdir(tmp_path):
+        # Create a config with ImageId in the wrong section,
+        # and sometimes in the correct section
+        config = "\n".join(
```
DaanDeMeyer, line 554: I would just inline the variable since there's only one use of it

tests/test_config.py
```diff
@@ -523,3 +528,39 @@ def test_package_manager_tree(tmp_path: Path, skel: Optional[Path], pkgmngr: Opt
...
+            [{section}]
+            ImageId=testimage
+            """
+            for section in sections
+        )
+        Path("mkosi.conf").write_text(config)
+
+        with caplog.at_level(logging.WARNING):
+            # Parse the config, with --image-id sometimes given on the command line
+            parse_config(args)
```
DaanDeMeyer, line 565: Please add a newline after the with block

tests/test_config.py
```diff
@@ -523,3 +528,39 @@ def test_package_manager_tree(tmp_path: Path, skel: Optional[Path], pkgmngr: Opt
...
+        assert warning_count == len(caplog.records)
```
DaanDeMeyer, line 566: nit: I think `assert len(caplog.records) == warning_count` is easier to read

DaanDeMeyer (changes requested):

tests/test_config.py
gdiscry: Yes it should. I checked my reflog and `Distribution=fedora` was introduced in 1ad1a8b so I missed it when rebasing.

mkosi/config.py
DaanDeMeyer: This wasn't fixed?

gdiscry: It was, then I ran the tests with `die()` instead of `logging.warning()` to catch any misconfiguration and reverted too far before committing 🤦‍♂️.

DaanDeMeyer (approved):

PR #1903, opened 2023-09-15
Convert script settings to lists

gdiscry:
With `mkosi.conf.d`, `mkosi.presets` and/or `[Match]`, it's become really easy to split the configuration of an image in logical parts. It might be done to isolate some settings that go together or to share the same configuration across several images.

Each script is now managed as a list of executables so that it's also possible to split them if desired.

I have a commit that restores backward compatibility for `{Prepare,Build,PostInstallation,Finalize}Script=` (no `s`) but it's based on #1902.


gdiscry:
@DaanDeMeyer Should I add backward-compatible hidden settings (without `s`) that are just aliases? Should they emit a deprecation warning?

DaanDeMeyer:
> Should I add backward-compatible hidden settings (without `s`) that are just aliases? Should they emit a deprecation warning?

Yeah I would do it like this. Probably with a deprecation warning.

gdiscry:
A remaining point where  I'd like someone else's opinion is the execution order for the scripts.

When calling `mkosi --postinst-script baz --postinst-script quuz` in the following folder

```
.
├── baz*
├── mkosi.conf.d/
│   ├── 10-foo/
│   │   └── mkosi.postinst*
│   └── 20-bar/
│       └── mkosi.postinst*
├── mkosi.postinst*
└── quuz*
```

the scripts are executed in the following order:

1. `baz`
2. `quuz`
3. `mkosi.postinst`
4. `mkosi.conf.d/10-foo/mkosi.postinst`
5. `mkosi.conf.d/20-bar/mkosi.postinst`

However, I think the following order is worth considering:

1. `mkosi.conf.d/10-foo/mkosi.postinst`
2. `mkosi.conf.d/20-bar/mkosi.postinst`
3. `mkosi.postinst`
4. `baz`
5. `quuz`

behrmann:
Why not?

1. mkosi.postinst
1. mkosi.conf.d/10-foo/mkosi.postinst
2. mkosi.conf.d/20-bar/mkosi.postinst
4. baz
5. quuz

This way it is global file -> override files -> runtime overrides.

But while I'm happily painting the bike shed: Why `mkosi.conf.d/10-foo/mkosi.postinst` instead of e.g. `mkosi.postinst.d/10-foo`?

4nd3r:
@gdiscry Thank you for your efforts. This is precisely what has been missing from my happiness.

And FWIW, I also support `global file -> override files -> runtime overrides` exec order.

mkosi/__init__.py
```diff
@@ -266,32 +266,29 @@ def run_prepare_script(state: MkosiState, build: bool) -> None:
...
-    if build:
-        with complete_step("Running prepare script in build overlay…"), mount_build_overlay(state):
-            bwrap(
-                [state.config.prepare_script, "build"],
-                network=True,
-                readonly=True,
-                options=finalize_mounts(state.config),
-                scripts={"mkosi-chroot": chroot} | package_manager_scripts(state),
-                env=env | state.config.environment,
-                stdin=sys.stdin,
```
DaanDeMeyer, line 289: These seem to be gone?

mkosi/__init__.py
```diff
@@ -266,32 +266,29 @@ def run_prepare_script(state: MkosiState, build: bool) -> None:
...
+    with contextlib.ExitStack() as stack:
+        if build:
+            stack.enter_context(complete_step("Running prepare scripts in build overlay…"))
```
DaanDeMeyer, line 271: I would log one step per script executed and include the script name in the step message. This makes it easier to figure out for users from which script some output is coming

mkosi/__init__.py
```diff
@@ -1638,7 +1638,7 @@ def run_selinux_relabel(state: MkosiState) -> None:
 
 
 def need_build_packages(config: MkosiConfig) -> bool:
-    return config.build_script is not None and len(config.build_packages) > 0
+    return bool(config.build_scripts) and len(config.build_packages) > 0
```
DaanDeMeyer, line 1641: Either use `len()` for both of these or `bool()` for both of these, but using both doesn't seem right

DaanDeMeyer (changes requested):
Looks great, just a few minor things

mkosi/__init__.py
gdiscry: They are not gone. There was only one call to `bwrap` for the script, with a few tweaks on the parameters depending on if `build` is set or not and the call to `mount_build_overlay`.

I extracted the variations of `if build: … else: …` in a setup phase and only kept a single call to bwrap for each script.

mkosi/__init__.py
gdiscry: I didn't look at this too closely when going through all the renames . I agree, it's doesn't seem right 😅.

mkosi/__init__.py
DaanDeMeyer: Right but according to github's UI bwrap is now being called as follows:

```
bwrap(
                [script, arg],
                network=True,
                readonly=True,
                options=finalize_mounts(state.config),
	@@ -301,8 +298,8 @@ def run_prepare_script(state: MkosiState, build: bool) -> None:
            )
```

Which is missing the `env` and `stdin` arguments that were there before

gdiscry: They re not missing, they are just collapsed in the diff and they appear if you expand the `@@`.

DaanDeMeyer (approved):

mkosi/resources/mkosi.md
```diff
@@ -1280,11 +1284,11 @@ are mounted into the current working directory before running the script
 and `$SRCDIR` is set to point to the current working directory. The
 following scripts are supported:
 
-* If **`mkosi.prepare`** (`PrepareScript=`) exists, it is first called
+* If any **`mkosi.prepare`** (`PrepareScripts=`) exist, they are first called
```
behrmann, line 1287: Documentation here should probably be clearer on how to use this. In a given directory there can only be a single file `mkosi.prepare`. So as a user I would wonder does this mean I can only have a single prepare script per preset or per `mkosi.conf.d` or is `mkosi.prepare` a directory and scripts will be picked up there?

gdiscry: I see how this can be confusing with regards to file vs directory. I could revert to the singular and add a paragraph explaining how multiple scripts can be called at each point with `mkosi.conf.d`, `mkosi.preset` and/or the settings in `mkosi.conf`.

Is the following weird when talking about a single script but mentioning the setting as a plural?

```suggestion
* If **`mkosi.prepare`** (`PrepareScripts=`) exists, it is first called
```

behrmann: Losing the "any" improves readability. I'd prefer some `mkosi.prepare.d` support (well, for the others to), but than can wait for later.

behrmann (approved):

PR #1907, opened 2023-09-19
Update NEWS

DaanDeMeyer:


NEWS.md
```diff
@@ -1,5 +1,20 @@
 # mkosi Changelog
 
+## v17
+
+- Added ToolsTreePackages= to add extra packages to the default tools
```
behrmann, line 5:
```suggestion
- Added `ToolsTreePackages=` to add extra packages to the default tools
```

NEWS.md
```diff
@@ -1,5 +1,20 @@
...
+  tree.
+- Removed dnf from Debian default tools tree packages as it was removed
+  from Debian testing
+- Added SystemdVersion= match to match on the host's systemd version
```
behrmann, line 9:
```suggestion
- Added `SystemdVersion=` match to match on the host's systemd version
```

NEWS.md
```diff
@@ -1,5 +1,20 @@
...
+- Added Format= match to match on the configured output format
```
behrmann, line 10:
```suggestion
- Added `Format=` match to match on the configured output format`
```

NEWS.md
```diff
@@ -1,5 +1,20 @@
...
+- Presets= can now be configured in global configuration files to select
```
behrmann, line 11:
```suggestion
- `Presets=` can now be configured in global configuration files to select
```

NEWS.md
```diff
@@ -1,5 +1,20 @@
...
+  which presets to build
+- UKIs can now be booted using direct linux boot.
+- We don't try to make images UEFI bootable anymore on architectures
+  that do not support UEFI
+- Fixed --help to show all options again
```
behrmann, line 16:
```suggestion
- Fixed `--help` to show all options again
```

behrmann (commented):
LGTM

NEWS.md
```diff
@@ -1,5 +1,20 @@
 # mkosi Changelog
 
+## v17
+
+- Added ToolsTreePackages= to add extra packages to the default tools
+  tree.
+- Removed dnf from Debian default tools tree packages as it was removed
```
bluca, line 7: uh, what? hang on a sec and let me see what the heck is going on there

behrmann: It just didn't migrate to testing because of some regression, but this did make trixie tools trees fail to build.

bluca: worse than that, it was removed, because it fails to build and nobody fixed it, let me throw something at it

behrmann: But not point really to wait with the release for dnf to be fixed in Debian. dnf missing was the reason to add `ToolsTreePackages=` to make it easy to add it (or anything else for that matter) back while keeping the comfort of just writing `default` for the tools tree setting instead of writing a new preset file.

behrmann (approved):

NEWS.md
DaanDeMeyer: If it's easy to fix I'd prefer to add back dnf to the default list of packages to be fair

bluca: I've uploaded a workaround, but it will take a week or so to reach testing

DaanDeMeyer: Let's add back dnf then, it means debian default tool trees will be broken for a week, but that shouldn't be a huge issue.

bluca (approved):

PR #1911, opened 2023-09-20
Add support for RHEL UBI

keszybz:
The repository naming matches what is visible in the standard ubi container image (registry.access.redhat.com/ubi8/ubi).

This is the first time that I'm doing anything more complicated wit UBIs, but I hope this is a sensible beginning.

In Distribution.installer(), try..except is replaced by assert. If the installer class for a distro is missing, then it's a programming error and not somthing we should suppress.

### DIFF
--- a/mkosi/distributions/centos.py
+++ b/mkosi/distributions/centos.py
@@ -33,7 +33,7 @@ def join_mirror(config: MkosiConfig, link: str) -> str:
     return urllib.parse.urljoin(config.mirror, link)
 
 
-class CentosInstaller(DistributionInstaller):
+class Installer(DistributionInstaller):
     @classmethod
     def filesystem(cls) -> str:
         return "xfs"
--- a/mkosi/distributions/centos.py
+++ b/mkosi/distributions/centos.py
@@ -36,6 +36,10 @@ def join_mirror(config: MkosiConfig, link: str) -> str:
 
 class Installer(DistributionInstaller):
     @classmethod
+    def pretty_name(cls) -> str:
+        return "CentOS"
+
+    @classmethod
     def filesystem(cls) -> str:
         return "xfs"
 
--- a/mkosi/distributions/centos.py
+++ b/mkosi/distributions/centos.py
@@ -93,7 +93,7 @@ class Installer(DistributionInstaller):
         release = int(state.config.release)
 
         if release <= 7:
-            die("CentOS 7 or earlier variants are not supported")
+            die(f"{cls.pretty_name()} 7 or earlier variants are not supported")
 
         setup_dnf(state, cls.repositories(state.config, release))
         (state.pkgmngr / "etc/dnf/vars/stream").write_text(f"{state.config.release}-stream\n")
@@ -103,8 +103,8 @@ class Installer(DistributionInstaller):
         # Make sure glibc-minimal-langpack is installed instead of glibc-all-langpacks.
         cls.install_packages(state, ["filesystem", "glibc-minimal-langpack"], apivfs=False)
 
-        # On Fedora, the default rpmdb has moved to /usr/lib/sysimage/rpm so if that's the case we need to
-        # move it back to /var/lib/rpm on CentOS.
+        # On Fedora, the default rpmdb has moved to /usr/lib/sysimage/rpm so if that's the case we
+        # need to move it back to /var/lib/rpm on CentOS.
         move_rpm_db(state.root)
 
     @classmethod
@@ -115,8 +115,8 @@ class Installer(DistributionInstaller):
     def remove_packages(cls, state: MkosiState, packages: Sequence[str]) -> None:
         invoke_dnf(state, "remove", packages)
 
-    @staticmethod
-    def architecture(arch: Architecture) -> str:
+    @classmethod
+    def architecture(cls, arch: Architecture) -> str:
         a = {
             Architecture.x86_64   : "x86_64",
             Architecture.ppc64_le : "ppc64le",
@@ -125,7 +125,7 @@ class Installer(DistributionInstaller):
         }.get(arch)
 
         if not a:
-            die(f"Architecture {a} is not supported by CentOS")
+            die(f"Architecture {a} is not supported by {cls.pretty_name()}")
 
         return a
 
--- a/mkosi/distributions/centos.py
+++ b/mkosi/distributions/centos.py
@@ -5,6 +5,7 @@ import shutil
 import urllib.parse
 from collections.abc import Sequence
 from pathlib import Path
+from typing import Iterable
 
 from mkosi.architecture import Architecture
 from mkosi.config import MkosiConfig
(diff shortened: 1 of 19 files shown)


mkosi/distributions/rhel_ubi.py
```diff
@@ -0,0 +1,153 @@
...
+            "xz",
+            "zstd",
+        ]
+
+    @classmethod
+    def setup(cls, state: MkosiState) -> None:
+        release = int(state.config.release)
+
+        if release <= 7:
+            die("UBI-7 or earlier variants are not supported")
```
DaanDeMeyer, line 73: "Rhel UBI-7"?

mkosi/distributions/rhel_ubi.py
```diff
@@ -0,0 +1,153 @@
...
+from mkosi.log import die
+from mkosi.state import MkosiState
+
+
+def join_mirror(config: MkosiConfig, link: str) -> str:
+    mirror = config.mirror or "https://cdn-ubi.redhat.com/content/public/ubi/dist/"
+    return urllib.parse.urljoin(mirror, link)
+
+
+class RhelUbiInstaller(DistributionInstaller):
```
DaanDeMeyer, line 20: Ideally this would inherit from `CentOSInstaller`, but I guess that's awkward because the repository names are not the same.

I think we can just override the `repositories()` method in this class as well and then the rest of the logic would be shared. Not exactly the cleanest but sharing the logic would be nice because we don't have to fix things in two places then.

mkosi/resources/mkosi.md
```diff
@@ -1188,13 +1190,12 @@ distributions:
...
-Linux*, *Fedora Linux*, *OpenSUSE*, *Mageia* or *OpenMandriva* images.
-Any distro that packages `pacman` may be used to build *Arch Linux*
-images. Any distribution that packages `zypper` may be used to build
-*openSUSE* images.
+available.
+Specifically,
+any distribution that packages `apt` may be used to build *Debian* or *Ubuntu* images.
+Any distribution that packages `dnf` may be used to build images for any of the rpm-based distributions.
+Any distro that packages `pacman` may be used to build *Arch Linux* images.
+Any distribution that packages `zypper` may be used to build *openSUSE* images.
```
DaanDeMeyer, line 1198: Can we make this a list if we're going to split it across multiple lines?

DaanDeMeyer (changes requested):
Can we add this to CI? Just need a dropin in .github/mkosi.conf.d and an addition to ci.yml. Of course that's assuming this has all the packages needed to build a bootable image.

mkosi/distributions/rhel_ubi.py
```diff
@@ -0,0 +1,153 @@
...
+                yield Repo(
+                    f"ubi-{v}-{repo}",
+                    f"baseurl={join_mirror(config, f'ubi{v}/{v}/$basearch/{repo}/os')}",
+                    cls.gpgurls(),
+                    enabled=False,
+                )
+
+    @classmethod
+    def repositories(cls, config: MkosiConfig, release: int) -> list[Repo]:
+        repos = [
```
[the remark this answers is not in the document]
keszybz: I opted to use a generator because then we can deal with the conditional `codeready-builder` part nicely. We just have an extra `yield`, no need to concatenate lists.

mkosi/distributions/rhel_ubi.py
```diff
@@ -0,0 +1,153 @@
...
+    @staticmethod
+    def gpgurls() -> tuple[str, ...]:
+        return (
+            "https://access.redhat.com/security/data/fd431d51.txt",
+        )
+
+    @classmethod
+    def _repository_variants(cls, config: MkosiConfig, repo: str) -> list[Repo]:
+        if config.local_mirror:
+            yield Repo(repo, f"baseurl={config.local_mirror}", cls.gpgurls())
```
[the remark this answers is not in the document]
keszybz: I don't see why this wouldn't work. I changed it to use a `listify` decorator so that the generator becomes an internal implementation detail.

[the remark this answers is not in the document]
keszybz: I take that back. The wrapper just creates unneeded noise. This is a private function, it can be a generator.

mkosi/distributions/rhel_ubi.py
keszybz: OK, I reworked it like that. It is quite a bit shorter. Maybe in the future, if they diverge too much, it'd be better to remove the inheritance. But at least for now, it seems better.

mkosi/distributions/rhel_ubi.py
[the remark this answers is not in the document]
keszybz: I added another commit to do the same from centos.

keszybz:
> Can we add this to CI? Just need a dropin in .github/mkosi.conf.d and an addition to ci.yml. Of course that's assuming this has all the packages needed to build a bootable image.

There is no kernel. The idea is that this is intended for container images, possibly with systemd. I'll try to add the right invocation.

mkosi/distributions/centos.py
```diff
@@ -132,148 +134,142 @@ def gpgurls() -> tuple[str, ...]:
...
                         cls.gpgurls(),
                         enabled=False,
-                    ),
-                ]
+                    )
 
     @classmethod
-    def repositories(cls, config: MkosiConfig, release: int) -> list[Repo]:
+    @listify
+    def repositories(cls, config: MkosiConfig, release: int) -> Generator[Repo, None, None]:
```
DaanDeMeyer, line 254: Ditto

mkosi/distributions/centos.py
```diff
@@ -132,148 +134,142 @@ def gpgurls() -> tuple[str, ...]:
...
+            yield from cls.repository_variants(config, "PowerTools")
 
-        return repos + cls.epel_repositories(config) + cls.sig_repositories(config)
+        yield from cls.epel_repositories(config)
+        yield from cls.sig_repositories(config)
 
     @classmethod
-    def epel_repositories(cls, config: MkosiConfig) -> list[Repo]:
+    @listify
+    def epel_repositories(cls, config: MkosiConfig) -> Generator[Repo, None, None]:
```
DaanDeMeyer, line 272: Ditto

mkosi/distributions/centos.py
```diff
@@ -132,148 +134,142 @@ def gpgurls() -> tuple[str, ...]:
         )
 
     @classmethod
-    def repository_variants(cls, config: MkosiConfig, repo: str) -> list[Repo]:
+    @listify
```
DaanDeMeyer, line 141: Do we even need this everywhere? Why not have all these return generators and pass an `Iterable` to `setup_dnf()`? Or just convert the entire thing to a list just before we invoke `setup_dnf()`. Adding the listify seems like noise primarily

mkosi/distributions/rhel_ubi.py
```diff
@@ -0,0 +1,60 @@
...
+
+class Installer(centos.Installer):
+    @staticmethod
+    def gpgurls() -> tuple[str, ...]:
+        return (
+            "https://access.redhat.com/security/data/fd431d51.txt",
+        )
+
+    @classmethod
+    def _repository_variants(cls, config: MkosiConfig, repo: str) -> Generator[Repo, None, None]:
```
DaanDeMeyer, line 24: Ditto. Also let's just name this `repository_variants()` so its named the same as the CentOS method so it gets overridden

tests/test_config.py
```diff
@@ -194,7 +194,7 @@ def test_parse_load_verb(tmp_path: Path) -> None:
 def test_os_distribution(tmp_path: Path) -> None:
     with chdir(tmp_path):
         for dist in Distribution:
-            _, [config] = parse_config(["-d", dist.name])
+            _, [config] = parse_config(["-d", dist.name.replace('_', '-')])
```
DaanDeMeyer, line 197: You can just use `dist.value` here

mkosi/distributions/centos.py
```diff
@@ -36,6 +36,10 @@ def join_mirror(config: MkosiConfig, link: str) -> str:
 
 
 class Installer(DistributionInstaller):
+    @classmethod
+    def pretty_name(cls) -> str:
+        return "CentOS"
```
DaanDeMeyer, line 41: You added this but only implemented it for CentOS? Let's add implementations for the other distributions as well and expose it via `Distribution` so we can use it in more places.

mkosi/distributions/rhel_ubi.py
```diff
@@ -0,0 +1,60 @@
...
+from typing import Generator
+
+from mkosi.config import MkosiConfig
+from mkosi.distributions import centos
+from mkosi.installer.dnf import Repo
+
+
+def join_mirror(config: MkosiConfig, link: str) -> str:
+    assert config.mirror is not None
+    return urllib.parse.urljoin(config.mirror, link)
```
DaanDeMeyer, line 13: This is identical to the same function in centos.py, so let's just use that one instead.

DaanDeMeyer (changes requested):

mkosi/distributions/centos.py
keszybz: I added this because you wrote:
> But I would prefer we keep it simple and just return a list of repos like we do everywhere else

I'll rework it to be a generator.

mkosi/distributions/centos.py
```diff
@@ -132,148 +134,142 @@ def gpgurls() -> tuple[str, ...]:
         )
 
     @classmethod
-    def repository_variants(cls, config: MkosiConfig, repo: str) -> list[Repo]:
+    @listify
+    def repository_variants(cls, config: MkosiConfig, repo: str) -> Generator[Repo, None, None]:
```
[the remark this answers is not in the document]
keszybz: `Iterable` works better because then the functions that return a list also qualify.

PR #1912, opened 2023-09-20
Add virtiofsd to default tools tree packages

DaanDeMeyer:


mkosi/distributions/debian.py
```diff
@@ -65,6 +65,7 @@ def tools_tree_packages(cls) -> list[str]:
             "systemd-container",
             "systemd",
             "tar",
+            "virtiofsd",
```
bluca, line 68: This is only available in unstable/testing - and on the latter I wouldn't rely too much, given it's rust...

DaanDeMeyer: Dropped from Debian for now

bluca (approved):

PR #1913, opened 2023-09-20
Support booting directory images in qemu

DaanDeMeyer:
Using virtiofsd, we can boot straight into a virtiofs instance of
a directory image. This does require the virtiofsd instance to run
as (fake) root so we can't switch to the user running mkosi anymore
when running unprivleged but that shouldn't be a problem.

This also only works with kernels that have the virtiofs driver
builtin which I don't think is the case in any major distros yet.

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -14,6 +14,7 @@ import tempfile
 import uuid
 from collections.abc import Iterator
 from pathlib import Path
+from typing import Optional
 
 from mkosi.architecture import Architecture
 from mkosi.config import (
@@ -28,12 +29,7 @@ from mkosi.partition import finalize_root, find_partitions
 from mkosi.run import MkosiAsyncioThread, run, spawn
 from mkosi.tree import copy_tree, rmtree
 from mkosi.types import PathString
-from mkosi.util import (
-    InvokingUser,
-    format_bytes,
-    qemu_check_kvm_support,
-    qemu_check_vsock_support,
-)
+from mkosi.util import format_bytes, qemu_check_kvm_support, qemu_check_vsock_support
 
 
 def machine_cid(config: MkosiConfig) -> int:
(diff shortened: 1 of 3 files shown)


mkosi/qemu.py
```diff
@@ -293,8 +294,8 @@ def run_qemu(args: MkosiArgs, config: MkosiConfig) -> None:
     ovmf, ovmf_supports_sb = find_ovmf_firmware(config) if firmware == QemuFirmware.uefi else (None, False)
 
     shm = []
-    if config.runtime_trees:
-        shm = ["-object", "memory-backend-memfd,id=mem,size=2G,share=on"]
+    if config.runtime_trees or config.output_format == OutputFormat.directory:
+        shm = ["-object", f"memory-backend-memfd,id=mem,size={config.qemu_mem},share=on"]
```
behrmann, line 298: What's the reason for not doing this in all cases? Maybe add a comment.

mkosi/qemu.py
```diff
@@ -161,14 +157,14 @@ def start_swtpm() -> Iterator[Path]:
...
+def start_virtiofsd(directory: Path, uid: Optional[int] = None, gid: Optional[int] = None) -> Iterator[Path]:
     with tempfile.TemporaryDirectory() as state:
         # Make sure virtiofsd is allowed to create its socket in this temporary directory.
-        os.chown(state, uid, gid)
+        os.chown(state, uid if uid is not None else os.getuid(), gid if gid is not None else os.getgid())
 
-        sock = Path(state) / Path("sock")
+        # Make sure we can use the socket name as a unique identifier for the fs as well but make sure it's not too
+        # long as virtiofs tag names are limited to 36 bytes.
+        sock = Path(state) / f"sock-{uuid.uuid4().hex[:30]}"
```
behrmann, line 167:
```suggestion
        sock = Path(state) / f"sock-{uuid.uuid4().hex}"[:35]
```
less mental arithmetic

mkosi/qemu.py
```diff
@@ -406,16 +407,19 @@ def run_qemu(args: MkosiArgs, config: MkosiConfig) -> None:
...
             elif "-kernel" not in args.cmdline:
-                kernel = config.output_dir / config.output_split_kernel
+                if firmware == QemuFirmware.uefi:
+                    kernel = config.output_dir / config.output_split_uki
+                else:
+                    kernel = config.output_dir / config.output_split_kernel
                 if not kernel.exists():
                     die(
-                        "No kernel found, please install a kernel in the image "
+                        f"kernel or UKI not found at {kernel}, please install a kernel in the image "
```
behrmann, line 422:
```suggestion
                        f"Kernel or UKI not found at {kernel}, please install a kernel in the image "
```
or
```suggestion
                        f"Neither kernel nor UKI not found at {kernel}, please install a kernel in the image "
```

behrmann (commented):
LGTM, a few nits only.

The commit "Support booting directory images in qemu" has a typo in the message: unprivleged - > unprivileged

Also `OutputFormat` could maybe grow a few more methods acting as predicates, e.g. `supports_virtiofsd_boot` and such.

PR #1914, opened 2023-09-21
Make `MkosiArgs` and `MkosiConfig` dumpable to and loadble from JSON

behrmann:
This should be a stepping stone for reproducible builds.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -792,7 +792,7 @@ class MkosiConfig:
     qemu_cdrom: bool
     qemu_firmware: QemuFirmware
     qemu_kernel: Optional[Path]
-    qemu_args: Sequence[str]
+    qemu_args: list[str]
 
     preset: Optional[str]
 
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -683,10 +683,12 @@ class MkosiArgs:
         })
 
     def to_json(self, *, indent: Optional[int] = 4, sort_keys: bool = True) -> str:
+        """Dump MKosiArgs as JSON string."""
         return json.dumps(dataclasses.asdict(self), cls=MkosiJsonEncoder, indent=indent, sort_keys=sort_keys)
 
     @classmethod
-    def from_json(cls, s: Union[str, dict[str, Any], SupportsRead[str], SupportsRead[bytes]]) -> "MkosiArgs":
+    def _load_json(cls, s: Union[str, dict[str, Any], SupportsRead[str], SupportsRead[bytes]]) -> dict[str, Any]:
+        """Load JSON and transform it into a dictionary suitable compatible with instantiating a MkosiArgs object."""
         if isinstance(s, str):
             j = json.loads(s)
         elif isinstance(s, dict):
@@ -697,8 +699,19 @@ class MkosiArgs:
             raise ValueError(f"{cls.__name__} can only be constructed from JSON from strings, dictionaries and files.")
 
         transformer = json_type_transformer(cls)
-        tj = {k: transformer(k, v) for k, v in j.items()}
-        return cls(**tj)
+        return {k: transformer(k, v) for k, v in j.items()}
+
+    @classmethod
+    def from_json(cls, s: Union[str, dict[str, Any], SupportsRead[str], SupportsRead[bytes]]) -> "MkosiArgs":
+        """Instantiate a MkosiArgs object from a full JSON dump."""
+        j = cls._load_json(s)
+        return cls(**j)
+
+    @classmethod
+    def from_partial_json(cls, s: Union[str, dict[str, Any], SupportsRead[str], SupportsRead[bytes]]) -> "MkosiArgs":
+        """Return a new MkosiArgs with defaults overwritten by the attributes from passed in JSON."""
+        j = cls._load_json(s)
+        return dataclasses.replace(cls.default(), **j)
 
 
 @dataclasses.dataclass(frozen=True)
@@ -922,10 +935,12 @@ class MkosiConfig:
         }
 
     def to_json(self, *, indent: Optional[int] = 4, sort_keys: bool = True) -> str:
+        """Dump MKosiConfig as JSON string."""
         return json.dumps(dataclasses.asdict(self), cls=MkosiJsonEncoder, indent=indent, sort_keys=sort_keys)
 
     @classmethod
-    def from_json(cls, s: Union[str, dict[str, Any], SupportsRead[str], SupportsRead[bytes]]) -> "MkosiConfig":
+    def _load_json(cls, s: Union[str, dict[str, Any], SupportsRead[str], SupportsRead[bytes]]) -> dict[str, Any]:
+        """Load JSON and transform it into a dictionary suitable compatible with instantiating a MkosiConfig object."""
         if isinstance(s, str):
             j = json.loads(s)
         elif isinstance(s, dict):
@@ -936,8 +951,19 @@ class MkosiConfig:
             raise ValueError(f"{cls.__name__} can only be constructed from JSON from strings, dictionaries and files.")
 
         transformer = json_type_transformer(cls)
-        tj = {k: transformer(k, v) for k, v in j.items()}
-        return cls(**tj)
+        return {k: transformer(k, v) for k, v in j.items()}
+
+    @classmethod
+    def from_json(cls, s: Union[str, dict[str, Any], SupportsRead[str], SupportsRead[bytes]]) -> "MkosiConfig":
+        """Instantiate a MkosiConfig object from a full JSON dump."""
+        j = cls._load_json(s)
+        return cls(**j)
+
+    @classmethod
+    def from_partial_json(cls, s: Union[str, dict[str, Any], SupportsRead[str], SupportsRead[bytes]]) -> "MkosiConfig":
+        """Return a new MkosiConfig with defaults overwritten by the attributes from passed in JSON."""
+        j = cls._load_json(s)
+        return dataclasses.replace(cls.default(), **j)
 
 
 def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple[str, str, str]]:
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -679,8 +679,8 @@ class MkosiConfig:
     """
 
     include: tuple[str, ...]
-    presets: tuple[str]
-    dependencies: tuple[str]
+    presets: tuple[str, ...]
+    dependencies: tuple[str, ...]
 
     distribution: Distribution
     release: str
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -19,6 +19,7 @@ import platform
 import shlex
 import shutil
 import subprocess
+import tempfile
 import textwrap
 import uuid
 from collections.abc import Collection, Iterable, Iterator, Sequence
@@ -663,6 +664,18 @@ class MkosiArgs:
     doc_format: DocFormat
 
     @classmethod
+    def default(cls) -> "MkosiArgs":
+        """Alternative constructor to generate an all-default MkosiArgs.
+
+        This prevents MkosiArgs being generated with defaults values implicitly.
+        """
+        with tempfile.TemporaryDirectory() as tempdir:
+            with chdir(tempdir):
+                args, _ = parse_config([])
+
+        return args
+
+    @classmethod
     def from_namespace(cls, ns: argparse.Namespace) -> "MkosiArgs":
         return cls(**{
             k: v for k, v in vars(ns).items()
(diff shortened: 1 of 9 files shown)


mkosi/util.py
```diff
@@ -224,6 +224,9 @@ def _generate_next_value_(name: str, start: int, count: int, last_values: Sequen
     def values(cls) -> list[str]:
         return list(map(str, cls))
 
+    def to_json_repr(self) -> str:
+        return cast(str, self.value)
```
DaanDeMeyer, line 228: Bit confused why we need this? Can't we just check for isinstance of StrEnum in the json encoder? I think I'd prefer that

DaanDeMeyer (changes requested):
So we should probably ignore fields from the json that we don't know and use default values for fields that aren't set in the json, so that we have some freedom in adding/removing stuff

behrmann:
> So we should probably ignore fields from the json that we don't know and use default values for fields that aren't set in the json, so that we have some freedom in adding/removing stuff

So that's a bit the question of where to go with this. I don't think the goal should be to just accept any random JSON, i.e. accept hand-crafted config as JSON. Our config system and the CLI should remain the primary source of configuration.

That leaves only one generator for this JSON and that's the serialiser, which means that this should already be a fully-formed instance of the object with defaults applied. Therefore I don't think this needs its own application of defaults, since this might make recreating the exact config for reproducible builds more difficult.

Regarding missing keys: I think the only real chance for reproduction is using the version of mkosi that initially serialised the config, so staying with the above assumption, that the only source for the JSON would be the serialiser, missing keys should not happen. 

To make this workable this will need a version field somewhere. It could be added for the serialisation of `MkosiConfig` and `MkosiArgs`, but I left it out on purpose for now, since I think that's rather something to think about when putting this together in some way to actually make a reusable config. For what it's worth, I think the version field should live in the top-most JSON object that would contain (among the manifest and probably other things) the args and the config.

On top of that an upgrade tool, that consumes JSON config and applies new defaults to go from version $n$ to $n+1$, would be possible, but I think we should be strict in what we accept and not wade into the mine field of edge cases of changing things implicitly.

DaanDeMeyer:
Is reproducible builds even a use case for this? That would mean we'd have to serialize every tree used in the build process in this somehow as well, otherwise you won't be able to reproduce the build. I think this isn't actually all that useful for reproducible builds. For those you're better off just parsing the original configuration with all the trees etc included.

I was seeing this more as an alternative output for the summary verb so that users can get a machine readable presentation of the configuration. And potentially as an input where the config is provided as a JSON object instead of via individual CLI options (we could do the varlink thing Lennart suggested so many times at ASG then).

mkosi/config.py
```diff
@@ -651,7 +651,7 @@ class MkosiArgs:
     verb: Verb
     cmdline: list[str]
     force: int
-    directory: Optional[Path]
+    directory: Path
```
DaanDeMeyer, line 655: This is actually a bug, it's very much intentional that the directory can be set to something that makes sure we don't parse any configuration files. So we should fix this so that the directory can be set to the empty string again

mkosi/config.py
```diff
@@ -680,6 +680,10 @@ def from_namespace(cls, ns: argparse.Namespace) -> "MkosiArgs":
             if k in inspect.signature(cls).parameters
         })
 
+    def merge(self, other: "MkosiArgs") -> "MkosiArgs":
+        """Return a new MkosiArgs with attributes from other overwriting the ones from this instance."""
+        return dataclasses.replace(self, **dataclasses.asdict(other))
```
DaanDeMeyer, line 685: How does this work? Won't this just replace the entire dataclass since `MkosiArgs` and `MkosiConfig` must always have all fields?

mkosi/config.py
```diff
@@ -2514,3 +2529,94 @@ def default(self, obj: Any) -> Any:
         elif isinstance(obj, uuid.UUID):
             return str(obj)
         return json.JSONEncoder.default(self, obj)
+
+
+E = TypeVar("E", bound=StrEnum)
+
+
+def json_type_transformer(refcls: Union[type[MkosiArgs], type[MkosiConfig]]) -> Callable[[str, Any], Any]:
```
DaanDeMeyer, line 2583: So I wonder if we should make this a bit more robust and user friendly. If we're going to allow users to pass us json objects as config, it'd be nice if we could give proper errors if we can't parse something instead of a stacktrace.

DaanDeMeyer (changes requested):
Looks great! A few comments

mkosi/config.py
DaanDeMeyer: So I checked and if I use `--directory ""` then `args.directory` will be `None` so this is working as expected. I would just drop this commit

behrmann: Dropped.

behrmann: > So I checked and if I use --directory "" then args.directory will be None so this is working as expected. I would just drop this commit

That's what I didn't expect, because `Path("")` will be the current working directory.

mkosi/config.py
behrmann: It can't in this way. That was a brain fart. Factored out the JSON to dict conversion and used it here, this way a partial object represented as a dict can be loaded and overwrite `self`, which I would usually assume to be empty.

mkosi/config.py
behrmann: I think that should happen at the point of use, which doesn't exist yet, and not in the type transformer or the constructors. I really don't want to `die` in a constructor (yes, there's `die`s all over the place in `parse_config`) as that makes debugging in a REPL, something I do regularly, harder than it needs to be. :) 

I added a try block, though to reraise a single error type from the ones that I think should cover the ones that might appear here. 

I honestly hope, that we get to 3.11 some day and this can be made much prettier with match statements.

mkosi/config.py
```diff
@@ -1975,13 +1981,13 @@ def parse_new_includes(
         namespace: argparse.Namespace,
         defaults: argparse.Namespace,
     ) -> Iterator[None]:
-        l = len(getattr(namespace, "include", []))
+        current_includes = len(getattr(namespace, "include", []))
```
DaanDeMeyer, line 2068: This is a variable containing the length of the current includes so the variable name doesn't really seem right

mkosi/config.py
```diff
@@ -2620,3 +2635,94 @@ def default(self, obj: Any) -> Any:
...
+        Verb: enum_transformer,
+        DocFormat: enum_transformer,
+    }
+
+    def json_transformer(key: str, val: Any) -> Any:
+        fieldtype: Optional[dataclasses.Field[Any]] = fields_by_name.get(key)
+        # It is unlikely that the type of a field will be None only, so let's not bother with a different sentinel
+        # value
+        if fieldtype is None:
+            ValueError(f"{refcls} has no field {key}")
```
DaanDeMeyer: Missing `raise`?

mkosi/config.py
```diff
@@ -2620,3 +2635,94 @@ def default(self, obj: Any) -> Any:
...
+        return uuid.UUID(optuuid) if optuuid is not None else None
+
+    def root_password_transformer(
+        rootpw: Optional[list[Union[str, bool]]], fieldtype: type[Optional[tuple[str, bool]]]
+    ) -> Optional[tuple[str, bool]]:
+        if rootpw is None:
+            return None
+        return (cast(str, rootpw[0]), cast(bool, rootpw[1]))
+
+    def tree_transformer(
```
DaanDeMeyer, line 2728: I would call this source_target_transformer since not all trees support a target

mkosi/config.py
```diff
@@ -681,11 +681,15 @@ def from_namespace(cls, ns: argparse.Namespace) -> "MkosiArgs":
             if k in inspect.signature(cls).parameters
         })
 
+    def merge(self, other: "MkosiArgs") -> "MkosiArgs":
+        """Return a new MkosiArgs with attributes from other overwriting the ones from this instance."""
+        return dataclasses.replace(self, **dataclasses.asdict(other))
```
DaanDeMeyer, line 686: Do we still need this? this will just replace the entire dataclass.

mkosi/config.py
```diff
@@ -696,8 +700,17 @@ def from_json(cls, s: Union[str, dict[str, Any], SupportsRead[str], SupportsRead
...
-        tj = {k: transformer(k, v) for k, v in j.items()}
-        return cls(**tj)
+        return {k: transformer(k, v) for k, v in j.items()}
+
+    @classmethod
+    def from_json(cls, s: Union[str, dict[str, Any], SupportsRead[str], SupportsRead[bytes]]) -> "MkosiArgs":
+        j = cls._load_json(s)
+        return cls(**j)
+
+    def merge_json(self, s: Union[str, dict[str, Any], SupportsRead[str], SupportsRead[bytes]]) -> "MkosiArgs":
```
DaanDeMeyer, line 710: Do we need this method? What about having `from_json()` construct a default instance and then merging the json fields into that instance and returning that? If we have that I don't think we'd need a separate merge_json() method

mkosi/config.py
```diff
@@ -2778,7 +2778,10 @@ def json_transformer(key: str, val: Any) -> Any:
         # It is unlikely that the type of a field will be None only, so let's not bother with a different sentinel
         # value
         if fieldtype is None:
-            ValueError(f"{refcls} has no field {key}")
+            try:
+                ValueError(f"{refcls} has no field {key}")
+            except (ValueError, IndexError, AssertionError) as e:
+                raise ValueError(f"Unable to parse {val:r} for attribute {key:r} for {refcls.__name__}") from e
```
DaanDeMeyer, line 2784: Not sure I understand what's going on here?

DaanDeMeyer (changes requested):

mkosi/config.py
behrmann: Changed the name. PTAL.

mkosi/config.py
behrmann: No, copy paste error

mkosi/config.py
behrmann: Copy paste error from above. The idea is to generate a single known error when a transformer fails..

mkosi/config.py
behrmann: Changed.

mkosi/config.py
behrmann: They serve two different purposes and I added docstrings to clarify that.

`from_json` will need a full JSON dump, since the constructor for `MKosiArgs`/`MkosiConfig` doesn't have defaults. The `empty` constructor will generate an empty `MkosiArgs`/`MkosiConfig` with default values. I think this is safer than adding defaults to the regular constructor since defaults will not be used implicitly. 

To accept partial JSON, which we will need later, the way would be `MkosiConfig.empty().merge_json(some_partial_json_string)`. It will probably make sense to roll `.empty().merge_json()`  both into `load_partial_json` method.

DaanDeMeyer: Yes but when will we ever want to parse full JSON is what I'm wondering? I think we should just always allow partial JSON, to account for new properties that get added or maybe stuff that gets removed.

behrmann: I still think that parsing full JSON has a use for reproducible builds and when one explicitly doesn't want defaults applied implicitly. Right now this is a bit moot, because we still have to explore how this will be used. :)

behrmann: Added the `from_partial_json` method in this rebase.

mkosi/config.py
```diff
@@ -715,6 +715,11 @@ def merge_json(self, s: Union[str, dict[str, Any], SupportsRead[str], SupportsRe
         j = self._load_json(s)
         return dataclasses.replace(self, **j)
 
+    @classmethod
+    def from_partialjson(cls, s: Union[str, dict[str, Any], SupportsRead[str], SupportsRead[bytes]]) -> "MkosiArgs":
```
DaanDeMeyer, line 719: Missing delimiter

mkosi/config.py
```diff
@@ -2811,7 +2811,10 @@ def json_transformer(key: str, val: Any) -> Any:
         assert fieldtype is not None
         transformer = cast(Optional[Callable[[str, type], Any]], transformers.get(fieldtype.type))
         if transformer is not None:
-            return transformer(val, fieldtype.type)
+            try:
+                return transformer(val, fieldtype.type)
+            except (ValueError, IndexError, AssertionError) as e:
+                raise ValueError(f"Unable to parse {val:r} for attribute {key:r} for {refcls.__name__}") from e
```
DaanDeMeyer, line 2817: We should transform this to a `die()` somewhere in the chain.

DaanDeMeyer (changes requested):
Looking great,

DaanDeMeyer (approved):

PR #1916, opened 2023-09-21
Use raw strings for autologin dropins

DaanDeMeyer:
We use escapes in these so let's make sure we use raw strings.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -182,7 +182,7 @@ def configure_autologin(state: MkosiState) -> None:
                 """\
                 [Service]
                 ExecStart=
-                ExecStart=-/sbin/agetty -o '-f -p -- \\u' --autologin root --noclear --keep-baud console 115200,38400,9600 $TERM
+                ExecStart=-/sbin/agetty -o '-f -p -- \\\\u' --autologin root --noclear --keep-baud console 115200,38400,9600 $TERM
                 StandardInput=tty
                 StandardOutput=tty
                 """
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -179,7 +179,7 @@ def configure_autologin(state: MkosiState) -> None:
             dropin.parent.mkdir(parents=True, exist_ok=True)
         with umask(~0o644):
             dropin.write_text(
-                """\
+                r"""\
                 [Service]
                 ExecStart=
                 ExecStart=-/sbin/agetty -o '-f -p -- \\u' --autologin root --noclear --keep-baud console 115200,38400,9600 $TERM
```
behrmann, line 185: Making the change to a raw string, will write an additional slash, i.e. the diff will be
```diff
--- old	2023-09-21 17:32:43.329891671 +0200
+++ new	2023-09-21 17:32:45.843252993 +0200
@@ -1,5 +1,6 @@
+\
 [Service]
 ExecStart=
-ExecStart=-/sbin/agetty -o '-f -p -- \u' --autologin root --noclear --keep-baud console 115200,38400,9600 $TERM
+ExecStart=-/sbin/agetty -o '-f -p -- \\u' --autologin root --noclear --keep-baud console 115200,38400,9600 $TERM
 StandardInput=tty
 StandardOutput=tty
```

Is that intended?

behrmann: Also note the additional slash at the beginning

DaanDeMeyer: Right but we need the double slashes, the previous commit that moved these into the file actually introduced a bug. The first slash is a bug will fix

Issue #1918, opened 2023-09-24
git-describe fails due to dubious ownership


medhefgo:
From a systemd mkosi build:
```
[13/2287] Generating version.h with a custom command
fatal: detected dubious ownership in repository at '/work/src'
To add an exception for this directory, call:

	git config --global --add safe.directory /work/src
```
In this case it causes the build to fall back on "254-" as the version tag.

behrmann:
I'm not quite sure this is our bug. A few releases ago (I don't exactly remember when) git started to no longer run commands in repos owned by other users, since this might be a vector to for malicious code for people that automatically run things when entering git repositories.

I assume we trigger this because the repo is owned by the root user in the user namespace, which has a different GID from the user running mkosi. 

I'm not sure whether we can fix this bug in all generality, i.e. inject a local config at certain points, since people might clone repos during scripts. Probably setting the config globally and have it work for `/work/src`, might be the best option.

DaanDeMeyer:
Let's close this one since `MKOSI_UID` and `MKOSI_GID` are now made available to run commands the user invoking mkosi which makes it possible to avoid issues like this by running the command as the user invoking mkosi

PR #1919, opened 2023-09-25
Set MKOSI_UID and MKOSI_GID when running scripts

DaanDeMeyer:
These point to the UID/GID of the user running mkosi and can be used
to run commands as that user in scripts by using setpriv.

### DIFF
--- a/mkosi/resources/mkosi.md
+++ b/mkosi/resources/mkosi.md
@@ -1270,12 +1270,13 @@ Then, for each preset, we execute the following steps:
 To allow for image customization that cannot be implemented using
 mkosi's builtin features, mkosi supports running scripts at various
 points during the image build process that can customize the image as
-needed. Scripts are executed on the host system with a customized
-environment to simplify modifying the image. For each script, the
-configured build sources (`BuildSources=`) are mounted into the current
-working directory before running the script and `$SRCDIR` is set to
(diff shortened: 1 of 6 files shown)


DaanDeMeyer:
Fixes #1918 

This will allow running `meson setup` and `ninja -C build` in the systemd build script as the user running mkosi which should avoid the git ownership issue

mkosi/resources/mkosi.md
```diff
@@ -1363,26 +1364,33 @@ Scripts executed by mkosi receive the following environment variables:
   [SOURCE_DATE_EPOCH](https://reproducible-builds.org/specs/source-date-epoch/)
   for more information.
 
+* `$MKOSI_UID` is the uid of the user that invoked mkosi, potentially
+  translated to a uid in the user namespace that mkosi is running in.
+
+* `$MKOSI_GID` is the gid of the user that invoked mkosi, potentially
+  translated to a gid in the user namespace that mkosi is running in.
```
behrmann, line 1371: This should note that it can be used to run things as the owning user, to e.g. avoid git complaining.

behrmann (approved):
LGTM, but the docs should give an example, otherwise people will miss this building block.

behrmann (approved):

PR #1922, opened 2023-09-25
Use setpgid if available

DaanDeMeyer:
The next release of util-linux will include setpgid which has a
--foreground option that does exactly the same as our foreground()
function. This new utility allows us to make the processes executed
by bubblewrap the foreground process, which makes sure that SIGINT
isn't intercepted by bwrap.


behrmann:
Apart from CI failures LGTM.

PR #1928, opened 2023-09-26
Write IMAGE_ID and IMAGE_VERSION to /usr/lib/os-release

behrmann:
Totally untested, but jotted this down so I don't forget

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -176,6 +176,41 @@ def remove_packages(state: MkosiState) -> None:
             die(f"Removing packages is not supported for {state.config.distribution}")
 
 
+def configure_os_release(state: MkosiState) -> None:
+    """Write IMAGE_ID and IMAGE_VERSION to /usr/lib/os-release in the image."""
+    if not state.config.image_id and not state.config.image_version:
+        return
+
+    for candidate in ["usr/lib/os-release", "etc/os-release", "usr/lib/initrd-release", "etc/initrd-release"]:
+        osrelease = state.root / candidate
+        # at this point we know we will either change or add to the file
+        newosrelease = osrelease.with_suffix(".new")
+
+        if not osrelease.is_file() or osrelease.is_symlink():
+            continue
+
+        image_id_written = image_version_written = False
+        with osrelease.open("r") as old, newosrelease.open("w") as new:
+            # fix existing values
+            for line in old.readlines():
+                if state.config.image_id and line.startswith("IMAGE_ID="):
+                    new.write(f'IMAGE_ID="{state.config.image_id}"\n')
+                    image_id_written = True
+                elif state.config.image_version and line.startswith("IMAGE_VERSION="):
+                    new.write(f'IMAGE_VERSION="{state.config.image_version}"\n')
+                    image_version_written = True
+                else:
+                    new.write(line)
+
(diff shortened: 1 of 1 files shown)


behrmann:
Gave it a quick spin and seems to work

mkosi/__init__.py
```diff
@@ -176,6 +176,23 @@ def remove_packages(state: MkosiState) -> None:
...
+    osrelease = state.root / "usr/lib/os-release"
+
+    if not osrelease.is_file():
+        return
+
+    with osrelease.open("a") as osrel:
+        if state.config.image_id:
+            osrel.write(f'IMAGE_ID="{state.config.image_id}"\n')
+        if state.config.image_version:
+            osrel.write(f'IMAGE_VERSION="{state.config.image_version}"\n')
```
DaanDeMeyer, line 193: I don't think this actually happens in practice, but we should probably still take care to replace existing fields if they exist instead of just appending new one. We have this `read_os_release()` helper that might be useful here.

Also, I figure we should probably just fix both os-release in /usr and in /etc if they exist.

DaanDeMeyer (changes requested):

mkosi/__init__.py
behrmann: Opted for not using that, since we don't need all the parsing, just replacing or adding lines. Also the function no loops over `{usr/lib,etc}/{os,initrd}-release`

mkosi/__init__.py
```diff
@@ -176,6 +176,41 @@ def remove_packages(state: MkosiState) -> None:
...
+    """Write IMAGE_ID and IMAGE_VERSION to /usr/lib/os-release in the image."""
+    if not state.config.image_id and not state.config.image_version:
+        return
+
+    for candidate in ["usr/lib/os-release", "etc/os-release", "usr/lib/initrd-release", "etc/initrd-release"]:
+        osrelease = state.root / candidate
+        # at this point we know we will either change or add to the file
+        newosrelease = osrelease.with_suffix(".new")
+
+        if not osrelease.is_file():
```
DaanDeMeyer, line 189: We should probably skip symlinks here

DaanDeMeyer (approved):
LGTM just one nit

mkosi/__init__.py
behrmann: Right, only checked *dangling* symlinks there.

mkosi/__init__.py
```diff
@@ -1896,6 +1931,7 @@ def build_image(args: MkosiArgs, config: MkosiConfig, name: str, uid: int, gid:
 
             install_build_dest(state)
             install_extra_trees(state)
+            configure_os_release(state)
```
DaanDeMeyer, line 1934: Can we do this a bit later actually together with all the other `configure()` functions?

DaanDeMeyer (changes requested):

mkosi/__init__.py
behrmann: I went for the `configure_` name because I couldn't come up with anything, but I wanted to allow the postinst script to be able to see the full os release file already, but I'm fine with having it afterwards, since the environment variables will expose the data, too.

DaanDeMeyer (approved):

PR #1931, opened 2023-09-28
Fix setpgid logic

DaanDeMeyer:
Make sure we append the setpgid command after all the options. Let's also use -- to avoid any potential confusion about whether the following arguments are options or not.


DaanDeMeyer:
Trivial fix so merging

PR #1932, opened 2023-09-28
Add "none" distribution

snickl:
So here's a first stab.
I had to make the efi symlink conditional because my testing skeleton tree is for an arm system without EFI.
Also I just took the architecture list from Fedora for now, is there a way to say "all of them / don't care"?

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -39,6 +39,7 @@ from mkosi.config import (
     parse_config,
     summary,
 )
+from mkosi.distributions import Distribution
 from mkosi.installer import clean_package_manager_metadata, package_manager_scripts
 from mkosi.kmod import gen_required_kernel_modules, process_kernel_modules
 from mkosi.log import ARG_DEBUG, complete_step, die, log_step
@@ -138,6 +139,7 @@ def install_distribution(state: MkosiState) -> None:
             # Some distributions install EFI binaries directly to /boot/efi. Let's redirect them to /efi
             # instead.
             rmtree(state.root / "boot/efi")
+            (state.root / "boot").mkdir(exist_ok=True)
             (state.root / "boot/efi").symlink_to("../efi")
 
             if state.config.packages:
@@ -174,6 +176,14 @@ def remove_packages(state: MkosiState) -> None:
             state.config.distribution.remove_packages(state, state.config.remove_packages)
         except NotImplementedError:
             die(f"Removing packages is not supported for {state.config.distribution}")
+
+
+def check_root_populated(state: MkosiState) -> None:
+    """Check that the root was populated by looking for a os-release file."""
+    osrelease = state.root / "usr/lib/os-release"
+    if not osrelease.exists():
+        die(f"{osrelease} not found.",
+            hint="The root must be populated by the distribution, or from base trees, skeleton trees, and prepare scripts.")
 
 
 def configure_os_release(state: MkosiState) -> None:
@@ -968,6 +978,9 @@ def build_initrd(state: MkosiState) -> Path:
     symlink = state.workspace / "initrd"
     if symlink.exists():
         return symlink.resolve()
+
+    if state.config.distribution == Distribution.none:
+        die(f"Building a default initrd is not supported with distribution {state.config.distribution}")
 
     # Default values are assigned via the parser so we go via the argument parser to construct
     # the config for the initrd.
@@ -1945,6 +1958,7 @@ def build_image(args: MkosiArgs, config: MkosiConfig, name: str, uid: int, gid: 
                 save_cache(state)
                 reuse_cache(state)
 
+            check_root_populated(state)
             run_build_scripts(state)
 
             if state.config.output_format == OutputFormat.none:
--- a/mkosi/distributions/__init__.py
+++ b/mkosi/distributions/__init__.py
@@ -72,6 +72,7 @@ class DistributionInstaller:
 
 
 class Distribution(StrEnum):
+    none         = enum.auto()
     fedora       = enum.auto()
     debian       = enum.auto()
     ubuntu       = enum.auto()
--- a/mkosi/distributions/none.py
+++ b/mkosi/distributions/none.py
@@ -1,0 +1,27 @@
+# SPDX-License-Identifier: LGPL-2.1+
+
+from collections.abc import Sequence
+
+from mkosi.distributions import DistributionInstaller
+from mkosi.log import die
+from mkosi.state import MkosiState
+
+
+class Installer(DistributionInstaller):
+    @classmethod
+    def setup(cls, state: MkosiState) -> None:
+        pass
+
+    @classmethod
+    def install(cls, state: MkosiState) -> None:
+        pass
+
+    @classmethod
+    def install_packages(cls, state: MkosiState, packages: Sequence[str]) -> None:
+        if packages:
+            die("Installing packages is not supported with distribution 'none'")
+
+    @classmethod
+    def remove_packages(cls, state: MkosiState, packages: Sequence[str]) -> None:
+        if packages:
+            die("Removing packages is not supported with distribution 'none'")
--- a/mkosi/resources/mkosi.md
+++ b/mkosi/resources/mkosi.md
@@ -395,8 +395,8 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 
 : The distribution to install in the image. Takes one of the following
   arguments: `fedora`, `debian`, `ubuntu`, `arch`, `opensuse`, `mageia`,
-  `centos`, `rhel-ubi`, `openmandriva`, `rocky`, `alma`. If not
-  specified, defaults to the distribution of the host.
+  `centos`, `rhel-ubi`, `openmandriva`, `rocky`, `alma`, `none`.
+  If not specified, defaults to the distribution of the host.
 
 `Release=`, `--release=`, `-r`
 
@@ -1283,6 +1283,8 @@ distributions:
   guarantee that it will work at all and the core maintainers will
   generally not fix gentoo specific issues**)
 
+* *None* (**Requires the user to provide a pre-built rootfs**)
+
 In theory, any distribution may be used on the host for building images
 containing any other distribution, as long as the necessary tools are
 available.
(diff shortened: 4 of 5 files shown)


mkosi/distributions/none.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class NoneInstaller(DistributionInstaller):
+    @classmethod
+    def filesystem(cls) -> str:
+        return "ext4"
+
+    @classmethod
+    def package_type(cls) -> PackageType:
+        return PackageType.rpm
```
behrmann, line 19: That seems an odd choice? Maybe we should have an associated package type `none`, too.

DaanDeMeyer: Let's add `none` to `PackageType` and use it here

mkosi/distributions/none.py
```diff
@@ -0,0 +1,66 @@
...
+            Architecture.ppc64_le  : "ppc64le",
+            Architecture.riscv64   : "riscv64",
+            Architecture.s390x     : "s390x",
+            Architecture.x86_64    : "x86_64",
+        }.get(arch)
+
+        if not a:
+            die(f"Architecture {a} is not supported")
+
+        return a
```
behrmann, line 66: I see that it's taken from Fedora, by without making further any reason not do just do
```suggestion
        return str(arch)
```

DaanDeMeyer:
You can rebase on top of https://github.com/systemd/mkosi/pull/1933 once it's merged and it will make this a bit easier, you'll have to implement fewer methods.

So this should also come with a few extra checks. e.g. after running the prepare scripts we should check that `state.root` was actually populated, which I figure we can check by making sure a /usr/lib/os-release file exists.

mkosi/resources/mkosi.md
```diff
@@ -375,7 +375,7 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 
 : The distribution to install in the image. Takes one of the following
   arguments: `fedora`, `debian`, `ubuntu`, `arch`, `opensuse`, `mageia`,
-  `centos`, `openmandriva`, `rocky`, `alma`. If not
+  `centos`, `openmandriva`, `rocky`, `alma`, `none`. If not
```
DaanDeMeyer, line 378: We should document here that if the distribution is `none` that it's up to the user to populate the rootfs via a combination of base trees, skeleton trees and prepare scripts.

mkosi/distributions/none.py
```diff
@@ -0,0 +1,50 @@
...
+    def setup(cls, state: MkosiState) -> None:
+        pass
+
+    @classmethod
+    def install(cls, state: MkosiState) -> None:
+        pass
+
+    @classmethod
+    def install_packages(cls, state: MkosiState, packages: Sequence[str], apivfs: bool = True) -> None:
+        pass
```
DaanDeMeyer, line 42: We should fail explicitly here if `packages` is not empty

mkosi/distributions/none.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @classmethod
+    def remove_packages(cls, state: MkosiState, packages: Sequence[str]) -> None:
+        pass
```
DaanDeMeyer, line 46: Same here, we should fail if `packages` is not empty

mkosi/distributions/__init__.py
```diff
@@ -81,6 +83,7 @@ def is_centos_variant(self) -> bool:
 
     def is_dnf_distribution(self) -> bool:
         return self in (
+            Distribution.none,
```
DaanDeMeyer, line 86: This should be dropped

DaanDeMeyer (changes requested):

mkosi/distributions/none.py
snickl: Just an oversight. I also added "none", but without testing on my side so far.
#1933 also adds this, so should sort out itself.

snickl: Done

DaanDeMeyer:
We should also add a check in `build_initrd()` that fails if the distribution is `none` since we can't build a default initrd if that's the case.

DaanDeMeyer:
#1933 is merged now so you can rebase on main

DaanDeMeyer:
Please rebase on the latest changes from the main branch and remove the implementations of methods that have default implementation on the main branch

mkosi/__init__.py
```diff
@@ -138,7 +138,8 @@ def install_distribution(state: MkosiState) -> None:
             # Some distributions install EFI binaries directly to /boot/efi. Let's redirect them to /efi
             # instead.
             rmtree(state.root / "boot/efi")
-            (state.root / "boot/efi").symlink_to("../efi")
+            if (state.root / "boot/efi").exists():
```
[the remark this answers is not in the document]
DaanDeMeyer: This still needs to be fixed

DaanDeMeyer (changes requested):
Still missing the requested changes to build_initrd() and the check to see if $BUILDROOT was populated after running the prepare scripts. See previous comments

mkosi/__init__.py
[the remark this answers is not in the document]
snickl: I've added the mkdir, but I don't really get it

mkosi/resources/mkosi.md
snickl: Done, please check the phrasing

snickl:
> We should also add a check in `build_initrd()` that fails if the distribution is `none` since we can't build a default initrd if that's the case.

Done but not sure how to trigger this code path.

Why is this not possible? Just because of the --package cmdline elements?

snickl:
> So this should also come with a few extra checks. e.g. after running the prepare scripts we should check that `state.root` was actually populated, which I figure we can check by making sure a /usr/lib/os-release file exists.

Where would I put this? Before run_build_scripts() in build_image()?

mkosi/__init__.py
```diff
@@ -969,6 +971,9 @@ def build_initrd(state: MkosiState) -> Path:
     if symlink.exists():
         return symlink.resolve()
 
+    if state.config.distribution == Distribution.none:
+        die(f"Building an initrd is not supported with distribution {state.config.distribution}")
```
DaanDeMeyer, line 975: "Building a default initrd ..."

mkosi/distributions/none.py
```diff
@@ -0,0 +1,27 @@
...
+    @classmethod
+    def setup(cls, state: MkosiState) -> None:
+        pass
+
+    @classmethod
+    def install(cls, state: MkosiState) -> None:
+        pass
+
+    @classmethod
+    def install_packages(cls, state: MkosiState, packages: Sequence[str], apivfs: bool = True) -> None:
```
DaanDeMeyer, line 20: The `apivfs` argument can be dropped

DaanDeMeyer (changes requested):
Almost good to go

> Where would I put this? Before run_build_scripts() in build_image()?

Never mind, we'll merge without that check, I'll add it myself in a follow up.

> Why is this not possible? Just because of the --package cmdline elements?

Yes, our default initrd should just consist out of systemd, udev and kmod, but there's no way for us to request just those packages to be installed when there's no distribution, so the initrd has to be provided explicitly by the user.

mkosi/__init__.py
```diff
@@ -176,6 +178,18 @@ def remove_packages(state: MkosiState) -> None:
             die(f"Removing packages is not supported for {state.config.distribution}")
 
 
+def check_none_root_populated(state: MkosiState) -> None:
+    """If no distribution was installed, check that root was populated by looking for os-release file."""
+    if state.config.distribution != Distribution.none:
+        return
```
DaanDeMeyer, line 184: This shouldn't be specific to the `none` distribution, we should do this for every distribution.

mkosi/__init__.py
```diff
@@ -176,6 +178,18 @@ def remove_packages(state: MkosiState) -> None:
...
+
+    osrelease = state.root / "usr/lib/os-release"
+    if osrelease.is_file():
+        return
+
+    die("Root must be populated from skeleton tree or by other means for distribution 'none'")
```
keszybz, line 190: I think it'd be better to say:
```python
die(f"{osrelease} is not a file.",
    hint=""Root must be populated from skeleton tree or by other means for distribution 'none'")
```

keszybz (commented):
This seems reasonable.

mkosi/__init__.py
```diff
@@ -176,6 +178,14 @@ def remove_packages(state: MkosiState) -> None:
             die(f"Removing packages is not supported for {state.config.distribution}")
 
 
+def check_root_populated(state: MkosiState) -> None:
+    """Check that the root was populated by looking for a os-release file."""
+    osrelease = state.root / "usr/lib/os-release"
+    if not osrelease.is_file():
```
DaanDeMeyer, line 184: I would just use `exists()` here

mkosi/__init__.py
```diff
@@ -176,6 +178,14 @@ def remove_packages(state: MkosiState) -> None:
...
+        die(f"{osrelease} is not a file.",
```
DaanDeMeyer, line 185: "is not a file" => "not found"

mkosi/resources/mkosi.md
```diff
@@ -395,8 +395,11 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 
 : The distribution to install in the image. Takes one of the following
   arguments: `fedora`, `debian`, `ubuntu`, `arch`, `opensuse`, `mageia`,
-  `centos`, `rhel-ubi`, `openmandriva`, `rocky`, `alma`. If not
-  specified, defaults to the distribution of the host.
+  `centos`, `rhel-ubi`, `openmandriva`, `rocky`, `alma`, `none`.
```
DaanDeMeyer, line 398: Can you update the `Supported Distributions` section as well?

mkosi/__init__.py
snickl: Thanks, combined with the request to make it distribution-agnostic

DaanDeMeyer (changes requested):
Can you also make `config_default_distribution()` return `Distribution.none` if no distribution can be found?

mkosi/__init__.py
snickl: Done

mkosi/distributions/none.py
snickl: Done

mkosi/__init__.py
snickl: Done

keszybz:
Alsa unmark it as draft.

mkosi/__init__.py
snickl: Done

mkosi/resources/mkosi.md
snickl: Haven't seen that section before. I removed the explanation from --distribution down to that section and expanded it a little bit.

mkosi/__init__.py
snickl: Done

keszybz:
LGTM.

keszybz (approved):

PR #1935, opened 2023-09-28
Add Architecture match

DaanDeMeyer:


behrmann (approved):

behrmann:
Tests und type checkers already ran, the rest doesn't cover this, no point in wasting CI minutes.

PR #1936, opened 2023-09-28
Check early whether 'ukify' is available

keszybz:
It is an optional component. The user would be unhappy when we fail at the end of a lengthy installation process. So let's check early if we would need it and refuse to continue.
(When config.bootable == ConfigFeature.auto, we cannot to do the check before the installation is performed, so let's opt to be strict and do the check.)

Inspired by https://bugzilla.redhat.com/show_bug.cgi?id=2240598.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1126,6 +1126,28 @@ def build_uki(
     run(cmd)
 
 
+def want_uki(config: MkosiConfig) -> bool:
+    # Do we want to build an UKI according to config?
+    # Note that this returns True also in the case where autodetection might later
+    # cause the UKI not to be installed after the file system has been populated.
+
+    if config.bootable == ConfigFeature.disabled:
+        return False
+
+    if config.bootloader == Bootloader.none:
+        return False
+
+    if (config.output_format in (OutputFormat.cpio, OutputFormat.uki) and
+        config.bootable == ConfigFeature.auto):
+        return False
+
+    if (config.architecture.to_efi() is None and
+        config.bootable == ConfigFeature.auto):
+        return False
+
+    return True
+
+
 def install_uki(state: MkosiState, partitions: Sequence[Partition]) -> None:
     # Iterates through all kernel versions included in the image and generates a combined
     # kernel+initrd+cmdline+osrelease EFI file from it and places it in the /EFI/Linux directory of the ESP.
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -1450,6 +1461,10 @@ def check_inputs(config: MkosiConfig) -> None:
             if not p.is_file():
                 die(f"Initrd {p} is not a file")
 
+    if (want_uki(config) and
+        not (shutil.which("ukify") or Path("/usr/lib/systemd/ukify").exists())):
+        die(f"'ukify' not available")
```
DaanDeMeyer, line 1466: Don't need the f string here, it can be a normal one

DaanDeMeyer, line 1466: We should do this check in `load_config()` in config.py

DaanDeMeyer: Also I'd say "ukify not found"

DaanDeMeyer: And add a hint to tell users to set Bootable=no if they don't want a bootable image

DaanDeMeyer (changes requested):

keszybz:
Updated with all the suggested changes.

mkosi/config.py
```diff
@@ -2325,7 +2346,12 @@ def load_config(args: argparse.Namespace) -> MkosiConfig:
     ):
         die("This unprivileged build configuration requires at least Linux v5.11")
 
-    return MkosiConfig.from_namespace(args)
+    config = MkosiConfig.from_namespace(args)
+    if config.want_uki() and not (shutil.which("ukify") or Path("/usr/lib/systemd/ukify").exists()):
+        die("'ukify' not found.",
+            hint="Bootable=no can be used to create a non-bootable image.")
```
DaanDeMeyer, line 2352: So I forgot about the tools tree, ukify can come from the tools tree but we only mount that later. So either we have to delay this check until any tools tree is mounted or we only have to do the check if there is no tools tree define here.

keszybz: I moved the check back to `check_inputs()`. This is simple and should be fairly robust. I don't want to implement  some complicated approach where the check is delayed.

keszybz: Oh, OK, now I grok what you mean. Updated again.

mkosi/__init__.py
```diff
@@ -1126,28 +1126,39 @@ def build_uki(
...
+
+    if config.bootloader == Bootloader.none:
+        return False
+
+    if (config.output_format in (OutputFormat.cpio, OutputFormat.uki) and
+        config.bootable == ConfigFeature.auto):
+        return False
+
+    if (config.architecture.to_efi() is None and
+        config.bootable == ConfigFeature.auto):
```
behrmann, line 1145: Nit: both are still in the line length limit
```suggestion
    if config.output_format in (OutputFormat.cpio, OutputFormat.uki) and config.bootable == ConfigFeature.auto:
        return False

    if config.architecture.to_efi() is None and config.bootable == ConfigFeature.auto:
```

Other nit: `config.output_format in (OutputFormat.cpio, OutputFormat.uki)` comes up a few more times and on `Architecture`  we have a few helper predicates that say what about this it is, we are interested in. Something like that would be nice on `OutputFormat`, too.

mkosi/__init__.py
```diff
@@ -1462,6 +1473,12 @@ def check_outputs(config: MkosiConfig) -> None:
             die(f"Output path {f} exists already. (Consider invocation with --force.)")
 
 
+def check_tools(config: MkosiConfig) -> None:
+    if want_uki(config) and not (shutil.which("ukify") or Path("/usr/lib/systemd/ukify").exists()):
+        die("'ukify' not found.",
+            hint="Bootable=no can be used to create a non-bootable image.")
```
behrmann, line 1479: Nit: line length
```suggestion
    if want_uki(config) and not (shutil.which("ukify") or Path("/usr/lib/systemd/ukify").exists()):
        die("'ukify' not found.", hint="Bootable=no can be used to create a non-bootable image.")
```

behrmann (approved):
LGTM!

DaanDeMeyer (approved):
The line length stuff isn't too important, so let's merge

PR #1938, opened 2023-09-28
tooltrees: add curl to default packages

behrmann:
This makes it easy to download a file from a prepare script when using a default tool tree.

I went with `curl` instead of `curl-minimal` for Centos since `curl-minimal` is only available in Centos 9.

### DIFF
--- a/mkosi/distributions/debian.py
+++ b/mkosi/distributions/debian.py
@@ -42,6 +42,7 @@ class DebianInstaller(DistributionInstaller):
             "ca-certificates",
             "coreutils",
             "cpio",
+            "curl",
             "debian-archive-keyring",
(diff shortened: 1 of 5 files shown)


mkosi/distributions/debian.py
```diff
@@ -42,6 +42,7 @@ def tools_tree_packages(cls) -> list[str]:
             "ca-certificates",
             "coreutils",
             "cpio",
+            "curl"
```
DaanDeMeyer, line 45: Missing comma

behrmann: The vagaries of implicit string concatenation.

DaanDeMeyer (approved):

PR #1941, opened 2023-09-30
Add Include= setting

DaanDeMeyer:
This setting allows including extra configuration from user specified directories or files. The extra configuration is parsed immediately.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -18,9 +18,9 @@ import shutil
 import subprocess
 import textwrap
 import uuid
-from collections.abc import Iterable, Iterator, Sequence
+from collections.abc import Collection, Iterable, Iterator, Sequence
 from pathlib import Path
-from typing import Any, Callable, Optional, Union, cast
+from typing import Any, Callable, Collection, Optional, Union, cast
 
 from mkosi.architecture import Architecture
 from mkosi.distributions import Distribution, detect_distribution
@@ -870,7 +870,7 @@ class MkosiConfig:
         }
 
 
-def parse_ini(path: Path, only_sections: Sequence[str] = ()) -> Iterator[tuple[str, str, str]]:
+def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple[str, str, str]]:
     """
     We have our own parser instead of using configparser as the latter does not support specifying the same
     setting multiple times in the same configuration file.
@@ -2039,9 +2039,7 @@ def parse_config(argv: Sequence[str] = ()) -> tuple[MkosiArgs, tuple[MkosiConfig
         if path.exists():
             logging.debug(f"Including configuration file {Path.cwd() / path}")
 
-            for section, k, v in parse_ini(
-                path, only_sections=["Distribution", "Output", "Content", "Validation", "Host", "Preset"]
-            ):
+            for section, k, v in parse_ini(path, only_sections={s.section for s in SETTINGS}):
                 name = k.removeprefix("@")
                 ns = namespace if k == name else defaults
 
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -2027,7 +2027,7 @@ def parse_config(argv: Sequence[str] = ()) -> tuple[MkosiArgs, tuple[MkosiConfig
 
         return triggered is not False
 
-    def parse_config( path: Path, namespace: argparse.Namespace, defaults: argparse.Namespace) -> bool:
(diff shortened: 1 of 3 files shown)


tests/test_config.py
```diff
@@ -172,6 +180,36 @@ def test_parse_config(tmp_path: Path) -> None:
...
+        assert config.bootable == ConfigFeature.auto
+        assert config.split_artifacts == False
+
+        # Passing the directory should include both the main config file and the dropin.
+        _, [config] = parse_config(["--include", os.fspath(tmp_path / "abc")])
+        assert config.bootable == ConfigFeature.enabled
+        assert config.split_artifacts == True
+
+        # Passing the main config file should not include the dropin.
+        _, [config] = parse_config(["--include", os.fspath(tmp_path / "abc/mkosi.conf")])
```
behrmann, line 209: I'm not quite sure about the ergonomics of this, i.e. it seems easy to reason wrongly about this, since I always expect dropins to be parsed. How about this: When a directory is passed, we'll look for `mkosi.conf` by defeault. If a file is passed, we take it, but in either case all dropins are always honoured, i.e.

- `--include abc` -> looks for `abc/mkosi.conf` and all `abc/mkosi.conf.d/*.conf`
- `--include abc/qux.foo` looks for `abc/qux.foo` and all `abc/qux.conf.d/*conf`

mkosi/config.py
```diff
@@ -1948,6 +1924,45 @@ def parse_config(argv: Sequence[str] = ()) -> tuple[MkosiArgs, tuple[MkosiConfig
...
+        namespace: argparse.Namespace,
+        defaults: argparse.Namespace,
+    ) -> Iterator[None]:
+        l = len(getattr(namespace, "include", []))
+
+        try:
+            yield
+        finally:
+            for i in range(l, len(getattr(namespace, "include", []))):
+                parse_config(getattr(namespace, "include")[i], namespace, defaults)
```
behrmann, line 1938: Why not 
```suggestion
            for include in getattr(namespace, "include", []):
                parse_config(include, namespace, defaults)
```

mkosi/config.py
```diff
@@ -1948,6 +1924,45 @@ def parse_config(argv: Sequence[str] = ()) -> tuple[MkosiArgs, tuple[MkosiConfig
     settings_lookup_by_dest = {s.dest: s for s in SETTINGS}
     match_lookup = {m.name: m for m in MATCHES}
 
+    @contextlib.contextmanager
+    def parse_new_includes(
+        namespace: argparse.Namespace,
+        defaults: argparse.Namespace,
+    ) -> Iterator[None]:
+        l = len(getattr(namespace, "include", []))
```
behrmann, line 1932: Maybe I'm not seeing it, but is this used in this function?

tests/test_config.py
DaanDeMeyer: So this would come down to either parsing the directory of any given files or refusing files and always requiring users to pass directories.

Currently we only parse from mkosi.conf.d/ and I think it makes sense to keep that as it is instead of starting to parse arbitrary dropins to keep things simple. If users want dropins, they have to put their file in a directory and include that directory. I don't think that's a very limiting restriction.

mkosi/config.py
```diff
@@ -1948,6 +1924,45 @@ def parse_config(argv: Sequence[str] = ()) -> tuple[MkosiArgs, tuple[MkosiConfig
...
+
+        try:
+            yield
+        finally:
+            for p in getattr(namespace, "include", [])[l:]:
```
behrmann, line 1937: Ah, now I finally see it. But I don't think I understand it. For any list `a` the expression `a[len(a):]` will be the empty list, no?

DaanDeMeyer: The list could have been modified after yielding

behrmann: True, but then put some comments around this, so that the intent is clear to only parse the includes here that existed when this function was called.

tests/test_config.py
behrmann: It is not very limiting, but I was thinking about our spanking new config file spec and the lin
> All configuration paths must support drop-ins.

which we should aspire to.

DaanDeMeyer: Well, we do support drop-ins, by making the file a directory, putting the file as mkosi.conf in the directory and putting the dropins in mkosi.conf.d/.

DaanDeMeyer: We can probably still do this in some form, but it'd be in addition to what we already have, so I'd like to leave that for another PR

behrmann (approved):

PR #1942, opened 2023-09-30
Add InitrdPackages=

DaanDeMeyer:
We have ToolsTreePackages= already, so let's add a similar setting
for the default initrd to allow the default initrd to be used in
more scenarios.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2389,8 +2389,6 @@ def run_verb(args: MkosiArgs, presets: Sequence[MkosiConfig]) -> None:
     if args.verb == Verb.bump:
         return bump_image_version()
 
-    presets = finalize_tools(args, presets)
-
     if args.verb == Verb.summary:
         text = ""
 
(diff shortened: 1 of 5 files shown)


behrmann:
LGTM overall, but the removal of the default tools tree in the summary is a negative in my opinion. It makes it very easy for regular users to look at what the default looks like and base custom tool trees off of that without having to consult the source. I agree, that it would be more symmetric to the default initrd not to show it, but being more consistent in this regard comes at a slight usability cost. How about adding a switch to suppress printing of, lets call them for lack of a term now, automatic presets? This could down the line be used to show (or not) the automatic initrd as well.

DaanDeMeyer:
I think we can cover the tools tree by documenting which packages are included in the default tools tree. But showing the full preset for it in the summary is in my opinion not useful as none of the summary actually matters except the list of packages. So I would focus on documenting the packages and leaving the default summary out.

behrmann:
That 's reasonable, so let's add the packages to the docs, maybe as a table, so that it doesn't take up too much vertical space.

DaanDeMeyer:
Added the table and added one more commit to add apt and debian-keyring to CentOS tools trees as both of those are shipped in EPEL.

PR #1943, opened 2023-10-01
Don't reuse unprivileged cache when building as root

DaanDeMeyer:
When using --incremental and first building an image unprivileged, followed by building it as root, we can't reuse the cache as the ownership will be wrong. So let's make sure we don't reuse the cache when that's the case.


mkosi/__init__.py
```diff
@@ -1753,6 +1753,10 @@ def reuse_cache(state: MkosiState) -> bool:
     else:
         return False
 
+    for p in (final, build):
+        if p.exists() and p.stat().st_uid != 0:
```
behrmann, line 1757: Wouldn't that stop reusing the cache whenever it is not owned by root, i.e. also when one only ever does unprivileged builds?

DaanDeMeyer: No because we're in the user namespace so it will appear as if owned by root.

behrmann: Right, sufficiently far down in the stack. A comment for posteriority would probably make sense.

behrmann (approved):

DaanDeMeyer:
Just added a comment so let's merge

PR #1944, opened 2023-10-01
Add RuntimeSize= setting

DaanDeMeyer:
Currently we unconditionally grow disk images to 8G before booting
them in systemd-nspawn or qemu. Let's do better here by making the
size configurable and not growing the disk images by default.

We also move format_bytes() to config.py as most other formatting
functions are located there.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -773,6 +773,7 @@ class MkosiConfig:
     tools_tree_release: Optional[str]
     tools_tree_packages: list[str]
     runtime_trees: list[tuple[Path, Optional[Path]]]
+    runtime_size: Optional[int]
 
     # QEMU-specific options
     qemu_gui: bool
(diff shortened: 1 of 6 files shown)


mkosi/config.py
```diff
@@ -498,6 +499,33 @@ def match_systemd_version(value: str) -> bool:
...
 
 
+def config_parse_bytes(value: Optional[str], old: Optional[int] = None) -> Optional[int]:
+    if not value:
+        return None
+
+    if value.endswith("G"):
+        factor = 1024 ** 3
+    elif value.endswith("M"):
+        factor = 1024 ** 2
```
behrmann, line 509: nit: Even black now respects this being formatted as
```suggestion
        factor = 1024**3
    elif value.endswith("M"):
        factor = 1024**2
```

mkosi/config.py
```diff
@@ -498,6 +499,33 @@ def match_systemd_version(value: str) -> bool:
...
+        factor = 1
+
+    if factor > 1:
+        value = value[:-1]
+
+    result = math.ceil(float(value) * factor)
+    if result <= 0:
+        die("Size out of range")
+
+    rem = result % 512
```
behrmann, line 522: I vaguely remember a discussion recently to keep partitions aligned to 4K, so should we also round here to the next 4K block?

behrmann (approved):
LGTM, one question and a nit.

PR #1948, opened 2023-10-03
Always set the ttyS0 credentials

DaanDeMeyer:
Even when booting with GUI, we might still have a serial terminal, so let's always set the serial terminal specific size credentials as these don't affect the GUI terminals at all.


DaanDeMeyer:
Pretty trivial, so let's just merge this one

PR #1949, opened 2023-10-03
Use local GPG keys from distribution-gpg-keys if available

DaanDeMeyer:
Let's prefer using local keys from the distribution-gpg-keys package if available.


keszybz (approved):
LGTM.

One suggestion would be make `config` an attribute of `Distribution` (set at creation time), and then we wouldn't need to pass the argument around in some many places.

Issue #1664, opened 2023-07-04
Add specifier support when parsing configuration files


MariusSchiffer:
I'm currently using a setup similar to the presets in the main systemd repository. I would like to output to a directory specified on the command line, as in my case the repository directory should not be written to.

The relative paths used for `BaseTrees` would not work then.  
Example:  
Preset base with `OutputDirectory=../../mkosi.output` (or no directory set)
Preset initrd with `BaseTrees=../../mkosi.output/base`

Call mkosi with `mkosi --output-directory=/tmp/mkosi.output`

A workaround for me here is calling the preset manually with the `--base-tree=/tmp/mkosi.output/base`, which gets unwieldy with several presets.

A possible solution I see is adding a PathsRelativeTo similar to what repart has, or calling `.format(output_directory=)` on these paths so one can use `BaseTrees={output}/base`

DaanDeMeyer:
We should probably add specifier support. I'd also like to more easily allow specifying paths relative to the top level directory mkosi was invoked from.

PR #1952, opened 2023-10-04
Add support for specifiers

DaanDeMeyer:
Let's allow referring to the value of various settings using
specifiers.

Fixes https://github.com/systemd/mkosi/issues/1664

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -551,6 +551,7 @@ class MkosiConfigSetting:
     path_read_text: bool = False
     path_secret: bool = False
     path_default: bool = True
+    specifier: str = ""
 
     # settings for argparse
     short: Optional[str] = None
@@ -1054,6 +1055,7 @@ SETTINGS = (
         dest="distribution",
         short="-d",
         section="Distribution",
+        specifier="d",
         parse=config_make_enum_parser(Distribution),
         match=config_make_enum_matcher(Distribution),
         default_factory=config_default_distribution,
@@ -1064,6 +1066,7 @@ SETTINGS = (
         dest="release",
         short="-r",
         section="Distribution",
+        specifier="r",
         parse=config_parse_string,
         match=config_make_string_matcher(),
         default_factory=config_default_release,
@@ -1073,6 +1076,7 @@ SETTINGS = (
     MkosiConfigSetting(
         dest="architecture",
         section="Distribution",
+        specifier="a",
         parse=config_make_enum_parser(Architecture),
         match=config_make_enum_matcher(Architecture),
         default=Architecture.native(),
@@ -1123,6 +1127,7 @@ SETTINGS = (
         metavar="FORMAT",
         name="Format",
         section="Output",
+        specifier="t",
         parse=config_make_enum_parser(OutputFormat),
         match=config_make_enum_matcher(OutputFormat),
         default=OutputFormat.disk,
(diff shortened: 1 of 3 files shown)


mkosi/config.py
```diff
@@ -2068,6 +2076,37 @@ def parse_config(argv: Sequence[str] = ()) -> tuple[MkosiArgs, tuple[MkosiConfig
     # Compare inodes instead of paths so we can't get tricked by bind mounts and such.
     parsed_includes: set[tuple[int, int]] = set()
 
+    def expand_specifiers(text: str, namespace: argparse.Namespace, defaults: argparse.Namespace) -> str:
+        percent = False
+        result = ""
```
behrmann, line 2081: This will have quadratic runtime, since a new string will be allocated every time a new character is added. if you make this a list an `"".join(result)` at the end it will still bump the list around in memory a bit when a new character is appended, but the allocations will average out and the runtime should be closer to linear in the length of `text`.

mkosi/resources/mkosi.md
```diff
@@ -1254,6 +1254,21 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
+understood:
+
+| Setting            | Specifier |
+|--------------------|-----------|
+| `Distribution=`    | `%d`      |
+| `Release=`         | `%r`      |
+| `Architecture=`    | `%a`      |
+| `Format=`          | `%t`      |
+| `Output=`          | `%o`      |
+| `OutputDirectory=` | `%O`      |
```
behrmann, line 1270:
```suggestion
| `OutputDirectory=` | `%O`      |
| Literal `%`        | `%%`      |
```
so that people know how to do that. It's also a bit of a backward compat break we should mention in the news, but I don't think people commonly have percent signs in their paths. At least I hope so. :)

mkosi/config.py
```diff
@@ -2195,6 +2235,22 @@ def parse_config(path: Path, namespace: argparse.Namespace, defaults: argparse.N
         if not match_config(path, namespace, defaults):
             return False
 
+        if extras:
+            for s in SETTINGS:
```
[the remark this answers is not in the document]
DaanDeMeyer: Yeah but I prefer using `s` as the variable name for the for loop everywhere, in which case this is necessary, so i'd prefer to keep this

mkosi/resources/mkosi.md
```diff
@@ -1254,6 +1254,22 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
   Additionally, the suffixes `K`, `M` and `G` can be used to specify a
   size in kilobytes, megabytes and gigabytes respectively.
 
+## Specifiers
+
+The current value of various settings can be accessed when parsing
+configuration files by using specifiers. To write a literal `%`
+character in a configuration file without treating it as a specified,
```
behrmann, line 1261: Typo:
```suggestion
character in a configuration file without treating it as a specifier,
```

mkosi/config.py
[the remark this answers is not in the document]
behrmann: That's ok, we're not losing anything by not doing this, we're just not gaining anything (more type safety). Instead of the "Make mypy happy" comment please add a todo, though, add a todo to solve this with a type guard once we're on a sufficiently new Python version.

behrmann (approved):
LGTM, one typo and a comment nit (see above).

PR #1953, opened 2023-10-04
json: move JSON keys to PascalCase

behrmann:
Generally, the keys are changed to whatever the format in the config files is, which is a nice side effect.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -684,7 +684,7 @@ class MkosiArgs:
         })
 
     def to_json(self, *, indent: Optional[int] = 4, sort_keys: bool = True) -> str:
-        """Dump MKosiArgs as JSON string."""
+        """Dump MkosiArgs as JSON string."""
         def key_transformer(k: str) -> str:
             return "".join(p.capitalize() for p in k.split("_"))
 
@@ -943,7 +943,7 @@ class MkosiConfig:
         }
 
     def to_json(self, *, indent: Optional[int] = 4, sort_keys: bool = True) -> str:
-        """Dump MKosiConfig as JSON string."""
+        """Dump MkosiConfig as JSON string."""
         def key_transformer(k: str) -> str:
             if (s := SETTINGS_LOOKUP_BY_DEST.get(k)) is not None:
                 return s.name
@@ -2823,10 +2823,8 @@ def json_type_transformer(refcls: Union[type[MkosiArgs], type[MkosiConfig]]) -> 
         # It is unlikely that the type of a field will be None only, so let's not bother with a different sentinel
         # value
         if fieldtype is None:
-            ValueError(f"{refcls} has no field {key}")
+            raise ValueError(f"{refcls} has no field {key}")
 
-        # TODO: exchange for TypeGuard once on 3.10
-        assert fieldtype is not None
         transformer = cast(Optional[Callable[[str, type], Any]], transformers.get(fieldtype.type))
         if transformer is not None:
             try:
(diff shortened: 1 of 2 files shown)


mkosi/config.py
```diff
@@ -684,7 +685,13 @@ def from_namespace(cls, ns: argparse.Namespace) -> "MkosiArgs":
 
     def to_json(self, *, indent: Optional[int] = 4, sort_keys: bool = True) -> str:
         """Dump MKosiArgs as JSON string."""
-        return json.dumps(dataclasses.asdict(self), cls=MkosiJsonEncoder, indent=indent, sort_keys=sort_keys)
+        def key_transformer(k: str) -> str:
+            if (s := SETTINGS_LOOKUP_BY_DEST.get(k)) is not None:
+                return s.name
+            return "".join(p.capitalize() for p in k.split("_"))
```
DaanDeMeyer, line 691: Does this make sense for `MkosiArgs`? It won't have anything in the settings

DaanDeMeyer (changes requested):

mkosi/config.py
behrmann: Without the tests fail because of `AutoBump`

DaanDeMeyer: I'm confused, we don't have an `AutoBump` setting in `SETTINGS` right?

behrmann: Ah, misunderstood you and thought you meant the last line of the hunk. I thought the `genkey_*` keys were in `SETTINGS`, but that was me confusing the long block of parser settings with the `SETTINGS` one...

mkosi/config.py
```diff
@@ -936,7 +946,13 @@ def cache_manifest(self) -> dict[str, Any]:
 
     def to_json(self, *, indent: Optional[int] = 4, sort_keys: bool = True) -> str:
         """Dump MKosiConfig as JSON string."""
-        return json.dumps(dataclasses.asdict(self), cls=MkosiJsonEncoder, indent=indent, sort_keys=sort_keys)
+        def key_transformer(k: str) -> str:
+            if (s := SETTINGS_LOOKUP_BY_DEST.get(k)) is not None:
+                return s.name
+            return "".join(p.capitalize() for p in k.split("_"))
```
DaanDeMeyer, line 952: Do we need these fallbacks at all for `MkosiConfig`? I think everything in there has an associated setting no?

mkosi/config.py
```diff
@@ -950,8 +966,13 @@ def _load_json(cls, s: Union[str, dict[str, Any], SupportsRead[str], SupportsRea
         else:
             raise ValueError(f"{cls.__name__} can only be constructed from JSON from strings, dictionaries and files.")
 
-        transformer = json_type_transformer(cls)
-        return {k: transformer(k, v) for k, v in j.items()}
+        value_transformer = json_type_transformer(cls)
+        def key_transformer(k: str) -> str:
+            if (s := SETTINGS_LOOKUP_BY_NAME.get(k)) is not None:
+                return s.dest
+            return "_".join(part.lower() for part in FALLBACK_NAME_TO_DEST_SPLITTER.split(k))
```
DaanDeMeyer, line 973: Ditto, do we need the fallback for `MkosiConfig`?

DaanDeMeyer (changes requested):

mkosi/config.py
behrmann: Yes, `preset` is an attribute of `MkosiConfig`, but there is no accompanying `MkosiConfigSetting`.

mkosi/config.py
behrmann: Same, `preset` is the culprit.

DaanDeMeyer (approved):

PR #1954, opened 2023-10-04
Add ImageId= and ImageVersion= specifiers

DaanDeMeyer:
Also rework the specifier tests a bit.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -611,19 +611,33 @@ def test_specifiers(tmp_path: Path) -> None:
         Architecture=arm64
 
         [Output]
+        ImageId=my-image-id
+        ImageVersion=1.2.3
         OutputDirectory=abcde
         Output=test
 
-        [Distribution]
-        ImageId=%d~%r~%a
-
-        [Output]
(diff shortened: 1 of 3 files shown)


tests/test_config.py
```diff
@@ -611,19 +611,33 @@ def test_specifiers(tmp_path: Path) -> None:
...
-        ImageId=%d~%r~%a
-
-        [Output]
-        CacheDirectory=%O/%o
+        [Content]
+        Environment=DISTRIBUTION=%d
+                    RELEASE=%r
+                    ARCHITECTURE=%a
+                    IMAGE_ID=%i
+                    IMAGE_VERSION=%v
```
behrmann, line 624: That doesn't test all that much, since we set these environment variables to the values in `load_environment` anyway, no?

behrmann (approved):
LGTM functionality-wise, but maybe the test should set slightly differently named variables.

tests/test_config.py
DaanDeMeyer: oops didn't notice that, let me set make the pascal case

PR #1956, opened 2023-10-04
Add a policy on adding new distributions

### LINKED ISSUE #236: [Meta] declare a policy about adding new distributions

DaanDeMeyer:
Fixes #236

### DIFF
--- a/docs/distribution-policy.md
+++ b/docs/distribution-policy.md
@@ -1,0 +1,32 @@
+# Adding new distributions
+
+Merging support for a new distribution in mkosi depends on a few
+factors. Not all of these are required but depending on how many of
+these requirements are satisfied, the chances of us merging support for
+your distribution will improve:
+
+1. Is the distribution somewhat popular? mkosi's goal is not to support
+   every distribution under the sun, the distribution should have a
+   substantial amount of users.
+2. Does the distribution differentiate itself somehow from the
+   distributions that are already supported? We're generally not
+   interested in supporting distributions that only consist of minimal
+   configuration changes to another distribution.
+3. Is there a long-term maintainer for the distribution in mkosi? When
+   proposing support for a new distribution, we expect you to be the
+   maintainer for the distribution and to respond when pinged for
+   support on distribution specific issues.
+4. Does the distribution use a custom package manager or one of the
+   already supported ones (apt, dnf, pacman, zypper)? Supporting new
+   package managers in mkosi is generally a lot of work. We can support
+   new ones if needed for a new distribution, but we will insist on the
+   package manager having a somewhat sane design, with official support
+   for building in a chroot and running unprivileged in a user namespace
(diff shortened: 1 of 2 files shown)


docs/distribution-policy.md
```diff
@@ -0,0 +1,32 @@
...
+already supported ones (apt, dnf, pacman, zypper)? Supporting new
+package managers in mkosi is generally a lot of work. We can support new
+ones if needed for a new distribution, but we will insist on the package
+manager having a somewhat sane design, with official support for
+building in a chroot and running unprivileged in a user namespace being
+the bare minimum features we expect from any new package manager.
+
+If none or only a few of these requirements are satisfied for the new
+distribution you want to propose, it's unlikely that we'll be willing to
+merge support for the new distribution. However, you can still use mkosi
```
behrmann, line 29: I think it would be nicer to formulate this the other way around
```suggestion
Only if all or most of these requirements are satisfied for the new
distribution you want to propose, is it likely that we'll be willing to
merge support for the new distribution. However, you can still use mkosi
```
I think it sounds more positive.

docs/distribution-policy.md
```diff
@@ -0,0 +1,32 @@
...
+proposing support for a new distribution, we expect you to be the
+maintainer for the distribution and to respond when pinged for support
+on distribution specific issues.
+4. Does the distribution use a custom package manager or one of the
+already supported ones (apt, dnf, pacman, zypper)? Supporting new
+package managers in mkosi is generally a lot of work. We can support new
+ones if needed for a new distribution, but we will insist on the package
+manager having a somewhat sane design, with official support for
+building in a chroot and running unprivileged in a user namespace being
+the bare minimum features we expect from any new package manager.
```
behrmann, line 25: Nit: 
```suggestion
1. Is the distribution somewhat popular? mkosi's goal is not to support
   every distribution under the sun, the distribution should have a
   substantial amount of users.
2. Does the distribution differentiate itself somehow from the
   distributions that are already supported? We're generally not interested
   in supporting distributions that only consist of minimal configuration
   changes to another distribution.
3. Is there a long-term maintainer for the distribution in mkosi? When
   proposing support for a new distribution, we expect you to be the
   maintainer for the distribution and to respond when pinged for support
   on distribution specific issues.
4. Does the distribution use a custom package manager or one of the
   already supported ones (apt, dnf, pacman, zypper)? Supporting new
   package managers in mkosi is generally a lot of work. We can support new
   ones if needed for a new distribution, but we will insist on the package
   manager having a somewhat sane design, with official support for
   building in a chroot and running unprivileged in a user namespace being
   the bare minimum features we expect from any new package manager.
```

PR #1957, opened 2023-10-04
Mention OutputDirectory= when parsing Output= or --output fails

### LINKED ISSUE #1824: --output behaves inconsistently with directories
Running something like `mkosi --distribution fedora --release 38 --architecture x86-64 --format directory --output x64` works just fine, but using `--output /tmp/x64` (or any other fully qualified path) instead fails with `‣ /tmp/x64 is not a valid filename`.

DaanDeMeyer:
Fixes #1824

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -468,18 +468,14 @@ def config_make_path_parser(*,
     return config_parse_path
 
 
-def config_parse_filename(value: Optional[str], old: Optional[str]) -> Optional[str]:
+def config_parse_output(value: Optional[str], old: Optional[str]) -> Optional[str]:
     if not value:
         return None
 
-    if value == "." or value == "..":
-        die(". and .. are not valid filenames")
-
-    if "/" in value:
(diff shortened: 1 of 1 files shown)


mkosi/config.py
```diff
@@ -468,18 +468,13 @@ def config_parse_path(value: Optional[str], old: Optional[Path]) -> Optional[Pat
...
-        die(". and .. are not valid filenames")
-
-    if "/" in value:
-        die(
-            f"{value!r} is not a valid filename. "
-            "(Output= requires a filename with no path components, relative to output directory.)"
-        )
+    if value == "." or value == ".." or "/" in value:
+        die(f"{value!r} is not a valid filename.",
+            hint="Output= or --output= requires a filename with no path components, use OutputDirectory= or --output-dir= to configure the output directory.")
```
behrmann, line 477: Could you break that line? It ends at column 158.

behrmann (approved):

PR #1958, opened 2023-10-05
editorconfig: Add global defaults for indent_style and indent_size

DaanDeMeyer:


behrmann (approved):
It's not like Emacs cares about this file. ;)

PR #1969, opened 2023-10-11
Run scripts with ".chroot" extension inside the image

DaanDeMeyer:
Our current approach to running scripts inside the image is only really applicable to shell scripts. Let's make it easier to run scripts written in arbitrary languages inside the image by running scripts with the ".chroot" extension inside the image.


bluca:
I am getting live-build PTSD-style flashbacks

bluca:
> > I am getting live-build PTSD-style flashbacks
> 
> Something something history is doomed to repeat itself?

first time as tragedy, and second time as, er what was that again

behrmann (approved):

PR #1973, opened 2023-10-12
Run qemu as the invoking user again

DaanDeMeyer:
This commit also reworks InvokingUser to calculate all its members on module import (when we haven't yet unshared the user namespace). become_root() is also changed to modify the InvokingUser object instead of returning the new uid, gid. Finally, we stop passing around uid, gid everywhere and just use the InvokingUser object directly as a singleton.

We also stop dropping privileges in mkosi itself. Instead, we prefer running ssh, qemu and the embedded web server unprivileged. This allows us to get rid of the logic to not unmount the last tools tree as we will now always still have enough privileges to do so.

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -14,7 +14,6 @@ import tempfile
 import uuid
 from collections.abc import Iterator
 from pathlib import Path
-from typing import Optional
 
 from mkosi.architecture import Architecture
 from mkosi.config import (
@@ -30,7 +29,7 @@ from mkosi.partition import finalize_root, find_partitions
 from mkosi.run import MkosiAsyncioThread, run, spawn
 from mkosi.tree import copy_tree, rmtree
 from mkosi.types import PathString
-from mkosi.util import qemu_check_kvm_support, qemu_check_vsock_support
+from mkosi.util import InvokingUser, qemu_check_kvm_support, qemu_check_vsock_support
 
 
 def machine_cid(config: MkosiConfig) -> int:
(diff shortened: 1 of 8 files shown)


mkosi/util.py
```diff
@@ -76,33 +76,15 @@ def flatten(lists: Iterable[Iterable[T]]) -> list[T]:
...
-        return pwd.getpwuid(cls.uid()).pw_name
-
-    @classmethod
-    def home(cls) -> Path:
-        return Path(f"~{cls.name()}").expanduser()
+    uid = int(os.getenv("SUDO_UID") or os.getenv("PKEXEC_UID") or os.getuid())
+    gid = int(os.getenv("SUDO_GID") or os.getgid())
+    name = pwd.getpwuid(uid).pw_name
+    home = Path(f"~{name}").expanduser()
+    invoked_as_root = uid == 0
```
behrmann, line 83:
```suggestion
    invoked_as_root = (uid == 0)
```
so it's easier to glance this is not
```python
    invoked_as_root = uid = 0
```

mkosi/qemu.py
```diff
@@ -508,4 +541,12 @@ def run_ssh(args: MkosiArgs, config: MkosiConfig) -> None:
...
 
-    run(cmd, stdin=sys.stdin, stdout=sys.stdout, env=os.environ, log=False)
+    run(
+        cmd,
+        user=InvokingUser.uid,
+        group=InvokingUser.gid,
+        stdin=sys.stdin,
+        stdout=sys.stdout,
+        env=os.environ,
+        log=False
```
behrmann, line 551:
```suggestion
        log=False,
```

behrmann (commented):
I'm not sure I quite follow. If then `kvm` group problem is Debian-specific, maybe it can also be solve there. @bluca?

Apart from that looks sensible to me, a few nits.

DaanDeMeyer:
> I'm not sure I quite follow. If then kvm group problem is Debian-specific, maybe it can also be solve there. @bluca?

Distro policy on Debian is to only allow access to /dev/kvm to users in the kvm group. Had something to do with various CVEs and such so I don't think they'll be willing to change it just for us. So it's better to work around it on our end in this case by just running qemu as the user that invoked mkosi.

behrmann (approved):

mkosi/run.py
```diff
@@ -125,7 +126,8 @@ def become_root() -> tuple[int, int]:
     os.setresgid(0, 0, 0)
     os.setgroups([0])
 
-    return SUBRANGE - 100, SUBRANGE - 100
+    InvokingUser.uid = SUBRANGE - 100
+    InvokingUser.gid = SUBRANGE - 100
```
behrmann, line 130: Looking at this again could we maybe add a
```python
INVOKINGUSER = InvokingUser
```
to `util` and import the functionality via this? Otherwise there is no indication, that this is some global state that will be manipulated here, i.e. for a casual reader it's unclear that the once initialised `InvokingUser` will change at all.

DaanDeMeyer: I will fix this in a follow up

PR #1976, opened 2023-10-13
Add WithRecommends=

DaanDeMeyer:
While this can already be configured using dropins. The concept of recommended packages seems widespread enough that we can provide an option to enable/disable it via the configuration file.

### DIFF
--- a/mkosi/resources/mkosi.md
+++ b/mkosi/resources/mkosi.md
@@ -694,6 +694,14 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
   `mkosi.build` scripts require to operate. Note that packages listed
   here will be absent in the final image.
 
+`WithRecommends=`, `--with-recommends=`
+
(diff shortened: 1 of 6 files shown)


mkosi/resources/mkosi.md
```diff
@@ -694,6 +694,11 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
   `mkosi.build` scripts require to operate. Note that packages listed
   here will be absent in the final image.
 
+`WithRecommends=`, `--with-recommends=`
+
+: Configures whether to install recommended packages or not. By default,
+  recommended packages are not installed.
```
behrmann, line 700:
```suggestion
: Configures whether to install recommended packages or weak dependencies, 
  depending on how they are called by the used package manager, or not. 
  By default, recommended packages are not installed. This is only used for package
  managers that support the concept, currently apt, dnf and zypper.
```

behrmann (approved):

PR #1978, opened 2023-10-13
Open qemu device nodes before unsharing user namespace

DaanDeMeyer:
Where possible, we should open the qemu device nodes before we unshare
the user namespace as this might not be possible anymore after unsharing
the user namespace because we might lose access to the kvm group.

Currently this is only possible for /dev/vhost-vsock. I've opened
https://gitlab.com/qemu-project/qemu/-/issues/1936 to hopefully make
it work for /dev/kvm as well.


keszybz:
LGTM.

PR #1979, opened 2023-10-13
Make sure shadow-utils is installed in tools trees

DaanDeMeyer:
This provides newuidmap/newgidmap which are used by virtiofsd, so let's make sure these are available.


keszybz:
LGTM.

PR #1980, opened 2023-10-13
Add a FAQ entry about KVM on Debian/Ubuntu

DaanDeMeyer:


mkosi/resources/mkosi.md
```diff
@@ -1925,6 +1925,21 @@ to build.
...
+
+While other distributions are OK with allowing access to `/dev/kvm`, on
+Debian/Ubuntu this is only allowed for users in the `kvm` group. Because
+mkosi unshares a user namespace when running unprivileged, it loses
+access to the `kvm` group and by the time we start `qemu` we don't have
+access to `/dev/kvm` anymore. As a workaround, you can change the
+permissions of the device nodes to `0666` which is sufficient to make
+KVM work unprivileged. To persist these settings across reboots, copy
+`/usr/lib/tmpfiles.d/static-nodes-permissions.conf` to
+`/etc/tmpfiles.d/static-nodes-permissions.conf` and change the mode of
```
behrmann, line 1940: Instead of overwriting the file wholesale and then possibly missing upgrades, shouldn't we advocate using a file that's lexicographically lower in order that only has a line for `/dev/kvm`?

mkosi/resources/mkosi.md
```diff
@@ -1925,6 +1925,21 @@ to build.
 
 Note that the minimum required Python version is 3.9.
 
+# Frequently Asked Questions (FAQ)
+
+- Why does `mkosi qemu` with KVM not work on Debian/Ubuntu?
+
+While other distributions are OK with allowing access to `/dev/kvm`, on
+Debian/Ubuntu this is only allowed for users in the `kvm` group. Because
+mkosi unshares a user namespace when running unprivileged, it loses
```
keszybz, line 1934: …, even if the calling user was in the `kvm` group, when mkosi unshares the user namespace to run unprivileged, …

mkosi/resources/mkosi.md
keszybz: Yeah, I think that's better.

keszybz (changes requested):

mkosi/resources/mkosi.md
keszybz: Though I think tmpfiles will warn… But meh, pick your poison.

behrmann: I just gave it a try and it doesn't warn (at least not on v252 on Debian bookworm), but `10devkvm.conf` lost against `static-nodes-permissions.conf` while `zzdevkvm.conf` won. This isn't what I would expect from
> If multiple files specify the same path, the entry in the file with the lexicographically earliest name will be applied

keszybz: Oh, OK. This is because `z` lines are additive and can happen multiple times and effectively the last one wins. So it is just necessary to provide an drop-in that sorts later.

The docs actually describe this correctly:
> All configuration files are sorted by their filename in lexicographic order, regardless of which of the     directories they reside in.
> If multiple operations shall be applied on the same file (such as ACL, xattr, file attribute adjustments), these are always done in the same fixed order.

Or maybe not, what does "same fixed order" mean?

behrmann: Not having looked at the code yet, on first reading this my assumption was that the `z`-line from the earlier file would win and that the comment about multiple operations would apply only if there were multiple lines affecting the same file, e.g. a `z`, a `t` and an `a`, then they would happen in some fixed order, e.g. always `a -> t -> z` or some other permutation.

keszybz: Yeah, I guess that's what it means. So we don't really document what happens when there are multiple `z` lines… It seems that they are applied in the order of listing.

behrmann: It's not only `z` lines, but `a` lines, too. I wonder, is this simply a bug? It's not really easy to reason about when some lines like `f` are lexicographically first line wins and others like `z` and `a` are the opposite.

DaanDeMeyer: It kinda sucks that there's no proper way to have something sort earlier than the file though because it doesn't have a numeric prefix. Given this stuff is hardly ever changed, I'd prefer just overriding the file

behrmann: I think as things stand I agree, though moving the three lines
```
z /dev/kvm          {{DEV_KVM_MODE}} - kvm -
z /dev/vhost-net    {{DEV_KVM_MODE}} - kvm -
z /dev/vhost-vsock  {{DEV_KVM_MODE}} - kvm -
``` 
out of `static-nodes-permissions.conf.in` and making them separately overridable might make sense.

DaanDeMeyer: Problem is that that's technically a compat break

DaanDeMeyer:
Let's merge, will happily review follow up PRs to improve things

PR #1981, opened 2023-10-13
Do not try to copy vmlinuz for non-bootable images

keszybz:
When building an initrd, copy_vmlinuz() would fail because there is no vmlinuz file to copy.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1353,6 +1353,9 @@ def copy_nspawn_settings(state: MkosiState) -> None:
 
 
 def copy_vmlinuz(state: MkosiState) -> None:
+    if state.config.bootable == ConfigFeature.disabled:
+        return
+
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -1353,6 +1353,9 @@ def copy_nspawn_settings(state: MkosiState) -> None:
 
 
 def copy_vmlinuz(state: MkosiState) -> None:
+    if state.config.bootable == ConfigFeature.disabled:
+        return False
```
behrmann, line 1357:
```suggestion
        return
```

behrmann (commented):
mypy complains, otherwise LGTM.

mkosi/__init__.py
keszybz: Indeed.

behrmann (approved):

PR #1983, opened 2023-10-17
Rework qemu device support checks

DaanDeMeyer:
Let's implement an available() method on the QemuDeviceNode enum and move the checks from load_config() to run_qemu() so they don't impede showing the summary or other verbs.

Let's also prefer using the file descriptor as a check whether the feature is available in run_qemu() instead of calling the available() method, as by the time we get to run_qemu() the available() method might return a different result.

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -30,16 +30,37 @@ from mkosi.partition import finalize_root, find_partitions
 from mkosi.run import MkosiAsyncioThread, run, spawn
 from mkosi.tree import copy_tree, rmtree
 from mkosi.types import PathString
-from mkosi.util import (
-    INVOKING_USER,
-    StrEnum,
-    qemu_check_kvm_support,
-    qemu_check_vsock_support,
-)
+from mkosi.util import INVOKING_USER, StrEnum
 
 
 class QemuDeviceNode(StrEnum):
+    kvm = enum.auto()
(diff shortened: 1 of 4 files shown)


mkosi/qemu.py
```diff
@@ -30,17 +30,38 @@
...
     vhost_vsock = enum.auto()
 
+    def device(self) -> Path:
+        return Path("/dev") / str(self)
+
+    def description(self) -> str:
+        return {
+            QemuDeviceNode.kvm: "KVM acceleration",
+            QemuDeviceNode.vhost_vsock: "a VSock device",
+        }[self]
```
behrmann, line 47: If there's no other users for the value, then what `description` does here is in theory the job of `__str__`, to make a nice user-visible string.

I don't think there's any users of the automatic value of `vhost-vsock` for the `vhost_vsock` member, so if you changed the values to `KVM acceleration` and `a VSock device` all the calls to `description` in the f-strings would be obselete.

behrmann (approved):
LGTM.

mkosi/qemu.py
DaanDeMeyer: Hmm, this really is a description and not really a stringification to me, so reusing __str__() for that doesn't feel right. Also we do use the stringification already in the available() method.

PR #1984, opened 2023-10-18
Reword readme

DaanDeMeyer:
Let's reduce the amount of detail we put in the readme and list the installation methods in order of how easy-to-use they are. Let's also only recommend using the distribution's package if it's sufficiently recent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,70 +13,67 @@ For a longer description and available features and options, see the
 
 # Installation
 
-You can install mkosi from your distribution with its package manager or the
-development version from git. The distribution packages are the preferred way to
-install mkosi.
+You can install mkosi from your distribution using its package manager
+or install the development version from git. If you install mkosi using
+your distribution's package manager, make sure it installs at least
+mkosi v16 or newer (Use `mkosi --version` to check). If your
+distribution only packages an older version of mkosi, it is recommended
+to install mkosi using one of the alternative installation methods
+listed below instead.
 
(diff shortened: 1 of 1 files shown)


behrmann:
I'm not sure about making the shim the primary advertised thing. It's a hack for us and any other way will be more known to people having prior Python exposure. I think we should encourage venv workflows for people not using distribution packages, e.g. one common workflow for me is having a per project venv with a `requirements.txt` where I pin the specific hash of mkosi used there.

DaanDeMeyer:
> I'm not sure about making the shim the primary advertised thing. It's a hack for us and any other way will be more known to people having prior Python exposure. I think we should encourage venv workflows for people not using distribution packages, e.g. one common workflow for me is having a per project venv with a requirements.txt where I pin the specific hash of mkosi used there.

Look at it this way, if someone actually cares about the venv stuff, they'll manage to figure out how to do so with a little extra reading and experimentation. On the other hand, for users that don't care about venv, we shouldn't force them to use it. If they run into trouble due to not using a venv, they'll arrive themselves at the conclusion that they should use it. But until they do, trying to get them to use venvs from the beginning is just going to lead to frustration.

When someone's looking to try out mkosi, we want to make it dead simple for them to do so, and if the distribution does not package a recent version, doing a git clone and running straight from the repository is by far the simplest way to do that. No need to be familiar with python packaging tools (pip, pipx, venv), just need a git clone and running a script.

behrmann (approved):

PR #1986, opened 2023-10-18
Move CI configuration to top level

DaanDeMeyer:
Let's make it easier to run the CI stuff locally by moving it to the top level directory.


.gitignore
```diff
@@ -13,7 +13,6 @@
 /dist
 /mkosi.build
 /mkosi.egg-info
-/mkosi.extra
```
behrmann, line 16: I think it might be nicer for contributors if we kept this and instead allowed the other files instead, i.e.
```
!mkosi.extra/usr/lib/systemd/mkosi-check-and-shutdown.sh
!mkosi.extra/usr/lib/systemd/mkosi-check-and-shutdown.sh
```

behrmann (commented):
In principle I'm fine with moving this up, but it does make running straight from the checkout a bit more difficult. `10-common.conf` will always be included and probably one of the distro ones with match. Should we maybe guard all of this behind an include?

DaanDeMeyer:
> In principle I'm fine with moving this up, but it does make running straight from the checkout a bit more difficult. 10-common.conf will always be included and probably one of the distro ones with match. Should we maybe guard all of this behind an include?

I don't think we need to support this use case. If you're running straight from the checkout you've voided warranty if you ask me and you should be fine with whatever we add/remove in the repository

behrmann:
Ok, let's see how that goes, but it should be noted in the news.

.gitignore
```diff
@@ -14,6 +14,8 @@
 /mkosi.build
 /mkosi.egg-info
 /mkosi.extra
+!mkosi.extra/usr/lib/systemd/mkosi-check-and-shutdown.sh
+!mkosi.extra/usr/lib/systemd/mkosi-check-and-shutdown.sh
```
behrmann, line 18: Ups, I forgot the `system/` in the second line.

.gitignore
```diff
@@ -14,6 +14,8 @@
 /mkosi.build
 /mkosi.egg-info
 /mkosi.extra
+!mkosi.extra/usr/lib/systemd/system/mkosi-check-and-shutdown.sh
+!mkosi.extra/usr/lib/systemd/system/mkosi-check-and-shutdown.sh
```
behrmann, line 18:
```suggestion
!mkosi.extra/usr/lib/systemd/mkosi-check-and-shutdown.sh
!mkosi.extra/usr/lib/systemd/system/mkosi-check-and-shutdown.service
```

PR #1987, opened 2023-10-18
Fix spawn()

DaanDeMeyer:
The finally statement needs to run after the process finishes, so we need to make spawn a context manager.

At the same time, let's make it configurable whether the spawn() subprocess runs in the foreground or not and let's not do it by default.

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -193,13 +193,17 @@ def start_swtpm() -> Iterator[Path]:
 
             cmdline += ["--ctrl", f"type=unixio,fd={sock.fileno()}"]
 
-            proc = spawn(cmdline, user=INVOKING_USER.uid, group=INVOKING_USER.gid, pass_fds=(sock.fileno(),))
-
-            try:
-                yield path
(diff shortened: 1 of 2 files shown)


behrmann (approved):

mkosi/qemu.py
```diff
@@ -193,13 +193,17 @@ def start_swtpm() -> Iterator[Path]:
...
-            try:
-                yield path
-            finally:
-                proc.terminate()
-                proc.wait()
+            with spawn(\
+                cmdline,\
+                user=INVOKING_USER.uid,\
+                group=INVOKING_USER.gid,\
+                pass_fds=(sock.fileno(),)\
```
behrmann, line 200: The line continuations in here are actually not needed (I tested this as far back as 3.5), since they are *in* the function call - where Python needs the line continuations (at least on our lower bound) is when combining multiple context managers.

DaanDeMeyer: Ahh I wasn't sure, will address all of these in a separate PR

PR #1989, opened 2023-10-18
Beef up bundled images

DaanDeMeyer:
Let's beef up the bundled images with all the packages from the tools
trees where applicable so that we can run easily run mkosi from these
images if needed for testing on different distributions.


behrmann:
Adding all these packages to what we build by default in CI will increase our CI times, right?

DaanDeMeyer:
Yes but as you can see it also improves coverage

action.yaml
```diff
@@ -87,6 +87,31 @@ runs:
         $BINARY --version
       done
 
+  # Make sure we have mkfs.xfs that can handle spaces in protofiles.
+  # TODO: Drop when we move to the next Ubuntu LTS.
+  - name: Update xfsprogs
+    shell: bash
+    working-directory: ${{ github.action_path }}
+    run: |
+        sudo apt-get install --assume-yes --no-install-recommends \
```
behrmann, line 96: I'm very much in favour of indenting four spaces, but if you want to indent only two, this is indented too far.

PR #1992, opened 2023-10-19
Pass through debug-shell arg to initrd and tooltree invocations

Cornelicorn:


DaanDeMeyer:
Why is this needed? It's already a global, so the nested build will be able to access it as needed?

Cornelicorn:
Overlooked something while hacking at multiple locations, yeah, this is not necessary, sorry.

PR #1996, opened 2023-10-19
Add git script to set safe.directory=* whenever git is executed

DaanDeMeyer:
Instead of requiring every git command to be executed as the user invoking mkosi, let's add a git script that appends "-c safe.directory=*" to disable the git safe directory check. This also reworks the script machinery to support scripts both inside and outside of the chroot since the git script needs to be available in both.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -10,15 +10,16 @@ import json
 import logging
 import os
 import resource
+import shlex
 import shutil
 import subprocess
 import sys
 import tempfile
 import textwrap
 import uuid
-from collections.abc import Iterator, Sequence
+from collections.abc import Iterator, Mapping, Sequence
 from pathlib import Path
-from typing import Optional, TextIO, Union
+from typing import ContextManager, Optional, TextIO, Union
 
 from mkosi.archive import extract_tar, make_cpio, make_tar
 from mkosi.config import (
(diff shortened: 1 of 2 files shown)


mkosi/__init__.py
```diff
@@ -325,6 +327,39 @@ def script_maybe_chroot(script: Path, mountpoint: str) -> list[str]:
...
+    with tempfile.TemporaryDirectory(prefix="mkosi-scripts") as d:
+        for name, script in scripts.items():
+            # Make sure we don't end up in a recursive loop when we name a script after the binary it execs
+            # by removing the scripts directory from the PATH when we execute a script.
+            (Path(d) / name).write_text(
+                textwrap.dedent(
+                    f"""\
+                    #!/bin/sh
+                    DIR="$(cd "$(dirname "$0")" && pwd)"
+                    PATH="$(echo $PATH | tr ':' '\\n' | grep -v $DIR | tr '\\n' ':')"
```
behrmann, line 341:
```suggestion
                    PATH="$(echo "$PATH" | tr ':' '\\n' | grep -v "$DIR" | tr '\\n' ':')"
```
Let's be safe here, those are user directories, they might well have spaces or other shenanigans in them.

behrmann (commented):
LGTM, just one nit

PR #2001, opened 2023-10-20
Add --json option

daandemeyer:
Let's allow using --json with summary to get a JSON dump of the
configuration.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -31,6 +31,7 @@ from mkosi.config import (
     ManifestFormat,
     MkosiArgs,
     MkosiConfig,
+    MkosiJsonEncoder,
     OutputFormat,
     SecureBootSignTool,
     Verb,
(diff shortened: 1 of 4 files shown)


mkosi/__init__.py
```diff
@@ -2525,10 +2525,10 @@ def run_verb(args: MkosiArgs, presets: Sequence[MkosiConfig]) -> None:
         return bump_image_version()
 
     if args.verb == Verb.summary:
-        text = ""
-
-        for config in presets:
-            text += f"{summary(args, config)}\n"
+        if args.json:
+            text = "\n".join(p.to_json() for p in presets)
```
behrmann, line 2529: I think I'd rather see something like this (though quick and hacky)
```suggestion
            payload = ",".join(p.to_json() for p in presets)
            text = f'{"Presets": [{payload}]}'
```
instead of newline separated JSON objects. This way we don't box ourselves into a corner right away and keep what we output here extensible.

mkosi/__init__.py
```diff
@@ -2525,10 +2526,10 @@ def run_verb(args: MkosiArgs, presets: Sequence[MkosiConfig]) -> None:
         return bump_image_version()
 
     if args.verb == Verb.summary:
-        text = ""
-
-        for config in presets:
-            text += f"{summary(args, config)}\n"
+        if args.json:
+            text = json.dumps([p.to_dict() for p in presets], cls=MkosiJsonEncoder, indent=4, sort_keys=True)
```
behrmann, line 2530:
```suggestion
            text = json.dumps({"Presets": [p.to_dict() for p in presets]}, cls=MkosiJsonEncoder, indent=4, sort_keys=True)
```
I think I'd prefer this, because then we can extend the summary with other stuff later, e.g. when we want to add back the `MkosiArgs` or add version information or something like that.

behrmann (approved):

PR #2003, opened 2023-10-20
Stop using subdirectories of cache and build dirs automatically

DaanDeMeyer:
In some cases, for example mkosi-initrd running as a kernel-install script, we want to reuse the system package cache. Currently this is impossible as we unconditionally create a subdirectory beneath the provided cache directory. Let's stop doing that, as users can now explicitly configure this behavior themselves by specifying the cache directory or build directory as follows:

```
CacheDirectory=mkosi.cache/%d~%r~%a
BuildDirectory=mkosi.builddir/%d~%r~%a
```

Additionally, make sure the default tools tree only reuses the same cache as the preset it's used for when the distribution, release and architecture are the same as the preset's.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -21,6 +21,7 @@ from collections.abc import Iterator, Mapping, Sequence
 from pathlib import Path
 from typing import Optional, TextIO, Union
 
+from mkosi.architecture import Architecture
 from mkosi.archive import extract_tar, make_cpio, make_tar
 from mkosi.config import (
     BiosBootloader,
@@ -1087,7 +1088,7 @@ def build_initrd(state: MkosiState) -> Path:
         "--cache-only", str(state.config.cache_only),
         "--output-dir", str(state.workspace / "initrd"),
         *(["--workspace-dir", str(state.config.workspace_dir)] if state.config.workspace_dir else []),
-        "--cache-dir", str(state.cache_dir.parent),
+        "--cache-dir", str(state.cache_dir),
         *(["--local-mirror", str(state.config.local_mirror)] if state.config.local_mirror else []),
         "--incremental", str(state.config.incremental),
         "--acl", str(state.config.acl),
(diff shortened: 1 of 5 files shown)


mkosi/__init__.py
```diff
@@ -2439,6 +2440,12 @@ def finalize_tools(args: MkosiArgs, presets: Sequence[MkosiConfig]) -> Sequence[
         release = p.tools_tree_release or distribution.default_release()
         mirror = p.tools_tree_mirror or (p.mirror if p.mirror and p.distribution == distribution else None)
 
+        if p.cache_dir:
+            same = p.distribution == distribution and p.release == release and p.architecture == Architecture.native()
+            cache = p.cache_dir if same else p.cache_dir / "tools"
+        else:
+            cache = None
```
behrmann, line 2447: It's a bit more vertical space but I think it takes a bit less mental gymnastics by not taking the predicate inside out
```suggestion
        if p.cache_dir:
            if p.distribution == distribution and p.release == release and p.architecture == Architecture.native():
                cache = p.cache_dir
            else:
                cache = p.cache_dir / "tools"
        else:
            cache = None
```

behrmann (approved):
I'm pretty sure we will make somebody's life worse with this change, but I agree that the greater control this gives is a net positive.

keszybz:
I did the trivial rebase to make this easier to test.

DaanDeMeyer:
This is also required to make %O specifier actually useful so let's merge.

keszybz:
@DaanDeMeyer I tried to figure out what this entails in practice, but I'm not sure of all the effects. For the cache directory, dnf uses subdirectories that are specific to distribution, version, and some other details. So IIUC, this will work fine: most of the time, different subdirectories will be used, without any conflict, and if we happen to refer to the same repository version from different "images", this repository will be shared. But for build directory, I'm think this might cause problems. In particular, imagine a naive build script which does `meson setup $BUILDDIR $SOURCEDIR && ninja -C $BUILDDIR` or even `meson setup $BUILDDIR/build $SOURCEDIR && ninja -C $BUILDDIR/build`. When this output directory is shared between different builds, things are likely to go very wrong. But even if this is handled "correctly", I don't see much benefit from sharing. In other words, the occasional possible sharing of build artifacts seems to be greatly outweighed by the changes of an incomplete rebuild.

Also, I don't think that "users can now explicitly configure this behavior themselves by specifying the cache directory or build directory" is a strong argument. So far the cache and build directories were designed to DTRT without any explicit configuration. I'm pretty sure we should keep it that way, and not require users to figure out cache invalidation details on their own.

In the commit message you only provide justification for the cache directory change. That part sounds OK. The change for the build directory seems at best iffy. Please explain how the new scheme is supposed to work.

keszybz:
Oh, I see this was merged. Github was still showing me the old state. I now think this change is not good.

DaanDeMeyer:
Well I was thinking that we'll never know for sure in mkosi what will be applicable to a certain build system. E.g. If you're building 100% static binaries independent of the distribution, you might be OK with sharing the build directory between distributions.

So I think the problem is mainly the defaults. We should probably default to what we did before. 

Maybe we need CacheSubdirectory and BuildSubDirectory options that configure what subdirectory is used.

PR #2004, opened 2023-10-21
Pass file descriptor to /dev/kvm to qemu on newer versions

DaanDeMeyer:
Newer versions of qemu support passing the kvm device to use as a file descriptor. Let's make use of this if the qemu version is recent enough.

At the same time, also do some general cleanup of the KVM logic.

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -4,6 +4,7 @@ import asyncio
 import base64
 import contextlib
 import enum
+import errno
 import hashlib
 import logging
 import os
@@ -31,6 +32,9 @@ from mkosi.run import MkosiAsyncioThread, run, spawn
 from mkosi.tree import copy_tree, rmtree
 from mkosi.types import PathString
 from mkosi.util import INVOKING_USER, StrEnum
+from mkosi.versioncomp import GenericVersion
+
+QEMU_KVM_DEVICE_VERSION = GenericVersion("8.3")
 
 
 class QemuDeviceNode(StrEnum):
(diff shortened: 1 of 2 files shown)


DaanDeMeyer:
This should not be merged until https://lists.gnu.org/archive/html/qemu-devel/2023-10/msg07358.html is merged into qemu.

mkosi/qemu.py
```diff
@@ -359,21 +378,18 @@ def run_qemu(args: MkosiArgs, config: MkosiConfig, qemu_device_fds: Mapping[Qemu
...
         die("RuntimeTrees= cannot be used when booting in BIOS firmware")
 
-    if config.qemu_kvm == ConfigFeature.enabled and not QemuDeviceNode.kvm.available():
+    if config.qemu_kvm == ConfigFeature.enabled and not config.architecture.is_native():
+        die(f"KVM acceleration requested but {config.architecture} does not match the native host architecture")
+
+    have_kvm = (config.qemu_kvm != ConfigFeature.disabled and
+                (qemu_version(config) < QEMU_KVM_DEVICE_VERSION and QemuDeviceNode.kvm.available()) or
+                (qemu_version(config) >= QEMU_KVM_DEVICE_VERSION and QemuDeviceNode.kvm in qemu_device_fds))
+    if (config.qemu_kvm == ConfigFeature.enabled and not have_kvm):
```
behrmann, line 387: Checking `config.qemu_kvm` being enabled and being unequal from disabled seems a bit weird - it is tristate, but once it's enabled it can't be anything else anymore
```suggestion
    have_kvm = (
        (qemu_version(config) < QEMU_KVM_DEVICE_VERSION and QemuDeviceNode.kvm.available()) or
        (qemu_version(config) >= QEMU_KVM_DEVICE_VERSION and QemuDeviceNode.kvm in qemu_device_fds)
    )
    if (config.qemu_kvm == ConfigFeature.enabled and not have_kvm):
```

mkosi/qemu.py
```diff
@@ -444,6 +460,16 @@ def run_qemu(args: MkosiArgs, config: MkosiConfig, qemu_device_fds: Mapping[Qemu
         *shm,
     ]
 
+    if have_kvm and config.architecture.is_native():
```
behrmann, line 463: Ah, `have_kvm` is reused here. I think I would still prefer to move the `config.qemu_kvm != ConfigFeature.disabled` check here to make this more local and so as to not have to remember that this contains a check whether this is enabled as well and not just a feature support.

mkosi/qemu.py
DaanDeMeyer: But this means we'll try to open /dev/kvm even if the feature is disabled which does not seem desirable?

DaanDeMeyer: Although this shouldn't really matter, let me rework it like you suggested

PR #2005, opened 2023-10-22
Enable WithDocs= by default

DaanDeMeyer:
Let's include documentation by default so as to not confuse users.


behrmann:
I'm ±0 on this, as either default is sensible. With docs is more for the distro building use case and without docs is good for the sysexty/containery use cases. Either is fine with me, if we stick with it.

PR #2007, opened 2023-10-22
Add support for profiles

DaanDeMeyer:
A profile is a set of configuration options that represents a known
variant of the given image. Its primary purpose is to allow grouping
known useful derivatives of the current image under a recognizable
identifier.

The difference with presets is that presets represent the individual
images that might need to be built to complete the final image, whereas
profiles change how the individual presets are built. Specifically, only
ever one profile can be selected, while it's perfectly valid to build
many presets.

We parse the specified profile after mkosi.conf but before
mkosi.conf.d to allow configuring the profile to build in mkosi.conf.


mkosi/__init__.py
```diff
@@ -2549,17 +2549,17 @@ def run_verb(args: MkosiArgs, presets: Sequence[MkosiConfig]) -> None:
         if last.compress_output:
             die(f"Sorry, can't {opname} a compressed image.")
 
-    for config in presets:
+    for i in images:
```
behrmann, line 2552: Here and everywhere else: `i` is very short and at least to me will always just be some integer. :) How about 
```suggestion
    for img in images:
```
`img` is also much easier to grep for then `i`.

mkosi/resources/mkosi.md
```diff
@@ -362,6 +367,13 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 
 ### [Config] Section
 
+`Profile=`, `--profile=`
+
+: Select the given profile. A profile is a configuration file or
+  directory in the `mkosi.profiles/` directory. When selected, this
+  configuration file or directory is included after parsing the
+  `mkosi.conf` file.
```
behrmann, line 375:
```suggestion
  `mkosi.conf` file, but before any `mkosi.conf.d/*.conf` drop in configuration.
```

mkosi/__init__.py
DaanDeMeyer: Switched back to `config` instead. `image.image` is a bit weird to read.

PR #2012, opened 2023-10-26
Move PackageManagerTrees= to the Distribution section

DaanDeMeyer:
This makes more sense than the Content section, since these trees don't actually end up in the image.


behrmann (approved):

behrmann:
the vsock permission denied errors are unrelated to this, but still a problem

PR #2013, opened 2023-10-27
Don't use configured default value when empty string is assigned

DaanDeMeyer:
Currently, if a setting is assigned the empty string on the CLI, any default value configured in the config file is still used. Let's change this and not use any configured default value when the empty string is assigned so that default values configured in config files can be overridden from the CLI.


keszybz (approved):
LGTM.

PR #2014, opened 2023-10-27
Allow overriding dnf selection with $MKOSI_DNF

keszybz:
dnf5 is nice, but it has some issues, for example installation of rhel9 currently fails with some error about signature. Using dnf instead work around the problem. Allow the user to override the selection using an environment variable.

(We have the tools directory, but it doesn't help here because we want to use 'dnf' instead of 'dnf5' and not different version of 'dnf5'.)

### DIFF
--- a/mkosi/installer/dnf.py
+++ b/mkosi/installer/dnf.py
@@ -34,7 +34,10 @@ def find_rpm_gpgkey(state: MkosiState, key: str, url: str) -> str:
 
 
 def dnf_executable(state: MkosiState) -> str:
-    return shutil.which("dnf5") or shutil.which("dnf") or "yum"
+    # Allow the user to override autodetection with an environment variable
+    dnf = state.config.environment.get("MKOSI_DNF")
+
+    return dnf or shutil.which("dnf5") or shutil.which("dnf") or "yum"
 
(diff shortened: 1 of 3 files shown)


mkosi/installer/dnf.py
```diff
@@ -34,8 +34,13 @@ def find_rpm_gpgkey(state: MkosiState, key: str, url: str) -> str:
 
 
 def dnf_executable(state: MkosiState) -> str:
+    # Allow the user to override autodetection with an environment variable
+    dnf = os.getenv("MKOSI_DNF")
```
DaanDeMeyer, line 38: Let's fetch this from `state.config.environment` instead, and propagate `MKOSI_DNF` into `state.config.environment` in `load_environment()`. That allows `MKOSI_DNF` to also be configured using `Environment=`.

DaanDeMeyer (changes requested):

mkosi/installer/dnf.py
keszybz: Updated like this.

behrmann (approved):

DaanDeMeyer (approved):

PR #2018, opened 2023-10-28
Add BuildSources= match

DaanDeMeyer:
One pattern I've started using a lot when I have to build multiple projects from source in a build script is to check if a source tree has been mounted at some location using BuildSources= and to only build the project if that's the case.

The problem with this is that this only allows me to skip the build, it doesn't allow me to skip installing the necessary build and runtime packages for that particular project.

Let's add a BuildSources= match so that everything related to the project can be skipped if the project is not configured to be mounted at some location using BuildSources=.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -390,7 +390,7 @@ def run_prepare_scripts(state: MkosiState, build: bool) -> None:
             step_msg = "Running prepare script {}…"
             arg = "final"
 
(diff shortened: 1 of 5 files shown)


DaanDeMeyer:
This also reworks the trees stuff a bit to be simpler and fixes the documentation for the trees settings.

behrmann (approved):

PR #2019, opened 2023-10-28
Extend README instructions

DaanDeMeyer:
Let's make sure we have complete copy-paste shell commands for
all the different installation methods. Let's also avoid confusion
and always use python3 in the example commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,12 @@ when using the `bin/mkosi` shim to configure the python interpreter used
 to execute mkosi. The shim can be symlinked to e.g. `/usr/local/bin` to
 make it accessible from the `PATH`.
 
+```shell
+git clone https://github.com/systemd/mkosi
+ln -s $PWD/mkosi/bin/mkosi /usr/local/bin/mkosi
+mkosi --version
+```
+
 ## Python installation methods
 
 mkosi can also be installed straight from the git repository url using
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,22 +29,32 @@ when using the `bin/mkosi` shim to configure the python interpreter used
...
 virtual environment and the mkosi binary, equivalent to
+
 ```shell
-python -m venv mkosivenv
-mkosivenv/bin/pip install git+https://github.com/systemd/mkosi.git
+python3 -m venv mkosivenv
+mkosivenv/bin/pip3 install git+https://github.com/systemd/mkosi.git
 # the mkosi binary is installed to mkosivenv/bin/mkosi
+source mkosivenv
+mkosi --version
```
behrmann, line 57: Do we have to be so specific? Python 2 is EOL, there is no python 2 anymore. ;)

I'd definitely scrap the `3` on the `pip3` though because in a venv all are always only symlinks within the venv to `python` and `pip`. (Apart from that I generally prefer `python -m pip` to `pip`, because it clearly links the used pip to the used python.)

The sourcing can also be scrapped if the mkosi in the venv is called via `mkosivenv/bin/mkosi`.

DaanDeMeyer: Yes there is no python 2 anymore, but on Ubuntu Jammy python itself doesn't exist, you have to use python3 so it makes sense to use the thing that'll work everywhere.

PR #2022, opened 2023-10-28
Don't mount our own passwd when building images

DaanDeMeyer:
It's much more important that we use the image's /etc/passwd instead of our own. Both inside and outside the image so that UIDs/GIDs get resolved correctly.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -448,7 +448,6 @@ def run_build_scripts(state: MkosiState) -> None:
 
     with (
         mount_build_overlay(state),\
-        mount_passwd(state.root),\
         mount_volatile_overlay(state),\
         finalize_chroot_scripts(state) as cd\
     ):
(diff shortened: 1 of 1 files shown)


behrmann:
How does this square with the reasons we've had this so far? Should we maybe synthesize a passwd of our own, mount it over the one in the image and then remove what we added initially and write the rest back to the image?

DaanDeMeyer:
> How does this square with the reasons we've had this so far? Should we maybe synthesize a passwd of our own, mount it over the one in the image and then remove what we added initially and write the rest back to the image?

This sounds messy. The reasons we've had for mounting our own was primarily to make tools like `whoami` work that try to find the user running the command in /etc/passwd. I think those are much less important than being able to chown something in a script and making sure that does the right thing (uses UIDs/GIDs from the image). We could synthesize our own but that seems a bit excessive.

behrmann:
I agree that chowning is more important, but the reason to get `whoami` working was being able to use the kernel makefile, so will that still work?

DaanDeMeyer:
Just found out we can use `KBUILD_BUILD_USER` to avoid the kernel build infra calling `whoami` so I'll use that instead.

behrmann:
Excellent!

PR #2023, opened 2023-10-29
mkosi-chroot: just run shell if no arguments are specified

keszybz:
chroot(1) is documented to run "$SHELL -i" as the default command. Let's do something similar and call "sh -i". When a user is using '--debug-shell' and one of the scripts fail, the mkosi-chroot can be used to chroot into the image. I think this is what users expect, based on the name.

(I don't think using $SHELL makes sense. It could either be set to 'sh', or to something from the outside, which might not even be installed in the chroot. We call 'sh' ourselves, so we know it must be there, so let's just call that uncoditionally.)


mkosi/__init__.py
```diff
@@ -341,7 +341,12 @@ def finalize_scripts(scripts: Mapping[str, Sequence[PathString]] = {}) -> Iterat
                     DIR="$(cd "$(dirname "$0")" && pwd)"
                     PATH="$(echo "$PATH" | tr ':' '\\n' | grep -v "$DIR" | tr '\\n' ':')"
                     export PATH
-                    exec {shlex.join(str(s) for s in script)} "$@"
+                    if [ $# -gt 0 ]; then
+                        args="$@"
+                    else
+                        args=("sh" "-i")
+                    fi
+                    exec {shlex.join(str(s) for s in script)} "$args"
```
behrmann, line 349: POSIX shell doesn't have arrays, i.e. it's a bashism or am I misremembering that? At least in dash, which will be `/bin/sh` on Debian, 
```shell
args=("sh" "-i")
```
gives me a syntax array.

I think one could
```suggestion
                        args='sh - i'
                    fi
                    exec {shlex.join(str(s) for s in script)} $args
```
but this will lead to to globbing/expansion on the exec line, I think.

keszybz: I replaced this by just copying the `exec` line into both branches. It's maybe not the most sophisticated solution, but it should suffice ;)

behrmann: That's good enough. :)

mkosi/__init__.py
```diff
@@ -352,9 +362,12 @@ def finalize_scripts(scripts: Mapping[str, Sequence[PathString]] = {}) -> Iterat
         yield Path(d)
 
 
-def finalize_host_scripts(state: MkosiState, chroot: Sequence[PathString]) -> contextlib.AbstractContextManager[Path]:
+def finalize_host_scripts(
+        state: MkosiState,
+        helpers: Mapping[str, Sequence[PathString]],
```
behrmann, line 367: indented too far
```suggestion
    state: MkosiState,
    helpers: Mapping[str, Sequence[PathString]],
```

You might want `(setq 'python-indent-def-block-scale 1)`—solved that issue for me.

mkosi/__init__.py
```diff
@@ -447,34 +463,37 @@ def run_build_scripts(state: MkosiState) -> None:
         )
 
     with (
-        mount_build_overlay(state),\
-        mount_volatile_overlay(state),\
-        finalize_chroot_scripts(state) as cd\
+        mount_build_overlay(state),
+        mount_volatile_overlay(state),
+        finalize_chroot_scripts(state) as cd,
```
behrmann, line 468: I was hesitant about this, but it seems to work on the python 3.9 I tested on a Debian bullseye machine, so I think we should try this and see if anybody complains. We should add a warning about this in the news, though.

mkosi/__init__.py
```diff
@@ -352,9 +362,12 @@ def finalize_scripts(scripts: Mapping[str, Sequence[PathString]] = {}) -> Iterat
...
 
 
-def finalize_host_scripts(state: MkosiState, chroot: Sequence[PathString]) -> contextlib.AbstractContextManager[Path]:
+def finalize_host_scripts(
+        state: MkosiState,
+        helpers: Mapping[str, Sequence[PathString]],
+) -> contextlib.AbstractContextManager[Path]:
     git = {"git": ("git", "-c", "safe.directory=*")} if find_binary("git") else {}
-    return finalize_scripts(git | {"mkosi-chroot": chroot} | package_manager_scripts(state))
+    return finalize_scripts(git | helpers | package_manager_scripts(state))
```
behrmann, line 370: pyright complains about the type here. There are two workaround:
1. typing `helpers` in the signature as `dict` instead of `Mapping`
2. adding a cast around helpers on this line

Though we strive for generic things here and `Mapping` does imply this being a readonly in addition, I think practically beats purity here, though a TODO comment for the future to look at this again would make sense.

mkosi/__init__.py
keszybz: Yeah. I was surprised, but 3.8 fails and 3.9 works, so I think this should be fine.

behrmann (approved):
LGTM all in all.


I'm ±0 on removing the suggestion to run the whole script under `mkosi-chroot`. Maybe we could keep a note around for people how to do that, for those who do.

I agree with the commit message that a better name than `mkosi-su-caller` would be nice. My contribution to the bikeshed: `mkosi-as-builduser

keszybz:
> I'm ±0 on removing the suggestion to run the whole script under `mkosi-chroot`. Maybe we could keep a note around for people how to do that, for those who do.

You *can* do it, but what would the purpose be? I don't see any good reason to not just use the native support.

keszybz:
> I agree with the commit message that a better name than `mkosi-su-caller` would be nice. My contribution to the bikeshed: `mkosi-as-builduser`

This is going in the right direction, but "builduser" would generally be the user that we are building as, i.e. almost the opposite of what this helper does. I renamed the command to `mkosi-as-caller`.

mkosi/resources/mkosi.md
```diff
@@ -1638,6 +1638,23 @@ available via `$PATH` to simplify common usecases.
...
+  back to the original user that called mkosi. This is useful when
+  we want to invoke build steps which will write to $BUILDDIR and
+  we want to have the files owned by the calling user.
+
+  For example, a complete `mkosi.build` script might be the following:
+
+  ```sh
+  set -ex
+
+  rm -rf "$BUILDDIR/build"
```
DaanDeMeyer, line 1652: Should this be in here? This would mean there'd be no incremental builds

mkosi/resources/mkosi.md
```diff
@@ -1638,6 +1638,23 @@ available via `$PATH` to simplify common usecases.
   the host system. This means that from a script, you can do e.g.
   `dnf install vim` to install vim into the image.
 
+* `mkosi-as-caller`: This script uses `setpriv` to switch from
+  the user `root` in the user namespace used for various build steps
+  back to the original user that called mkosi. This is useful when
+  we want to invoke build steps which will write to $BUILDDIR and
```
DaanDeMeyer, line 1644: Put BUILDDIR in backticks?

DaanDeMeyer (changes requested):

DaanDeMeyer:
We can fix these later

PR #2025, opened 2023-10-30
ssh: add RuntimeDirectory=sshd

bluca:
[  209.582912] sshd[888]: fatal: Missing privilege separation directory: /run/sshd


DaanDeMeyer:
Did something change in Debian? This used to work fine.

bluca:
> Did something change in Debian? This used to work fine.

Dunno, noticed after pulling from main

bluca:
also the unit is entirely custom

PR #2026, opened 2023-10-30
spawn: inherit parent $PATH by default

malt3:
Since eb2aaad7c2309a4f10af93c2b24ea660e433d1cf, `env` is an empty dict if not set explicitly.
On NixOS, this means `$PATH` will be unset (and there is no fallback like `/usr/bin`).

### DIFF
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -247,6 +247,13 @@ def spawn(
         # output.
         stdout = sys.stderr
 
+    env = {
+        "PATH": os.environ["PATH"],
+        "TERM": os.getenv("TERM", "vt220"),
+        "LANG": "C.UTF-8",
+        **env,
+    }
+
     try:
         with subprocess.Popen(
             cmdline,


keszybz:
Nah, `spawn` should be changed to handle this correctly instead. Spawning a subprocess with empty env is not something that we'd ever want to do (except maybe for testing), so we shouldn't require every caller to add this boilerplate.

Also, having a mutable object as a default argument is ugly… Here it doesn't cause an issue (I expect), because the object is not modified by `subprocess.Popen`, but it's still bad style.

So instead, just change the function to have `None` as the default. This should DTRT.

DaanDeMeyer:
I would have spawn() mimick run()

mkosi/qemu.py
```diff
@@ -608,7 +608,6 @@ def run_qemu(args: MkosiArgs, config: MkosiConfig, qemu_device_fds: Mapping[Qemu
             stdin=sys.stdin,
             stdout=sys.stdout,
             pass_fds=qemu_device_fds.values(),
-            env=os.environ,
```
malt3, line 611: @DaanDeMeyer I'm not sure if this had other useful side effect (besides inheriting the parent `$PATH`). Is this safe to drop?

DaanDeMeyer: The idea was that while environment variables shouldn't affect the image build by default, they should be in effect when we boot the image. So I would keep this intact.

DaanDeMeyer (changes requested):
See comment

DaanDeMeyer (approved):

PR #2027, opened 2023-10-30
Restore subdirectory use for $BUILDDIR

keszybz:
Sharing of a directory for the (package) cache seems fine. The package manager shall use subdirectories and/or unique file names, so there should be no accidental conflict, and using the same cache directory allows the cache to be reused when possible. OTOH, sharing of the directory for the build directory is unlikely to be useful (since at least the configuration will usually be different, so the build would generally mean rebuilding anything anyway), but also the build system might not do cache invalidation strongly enough, leading to inadvertetent sharing of build artifacts. Let's avoid the issue completely but not sharing the build directory between different distributions releases or architectures.

This partially reverts b17810b4b571d0acbb307a6f7cd5c6456a45e4d3.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -36,13 +36,12 @@
 - We now automatically configure the qemu firmware, kernel cmdline and
   initrd based on what type of kernel is passed by the user via
   `-kernel` or `QemuKernel=`
-- We don't create subdirectories beneath the configured cache or build
-  directories anymore. To get back the previous behavior, configure the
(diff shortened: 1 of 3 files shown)


DaanDeMeyer:
It'd be nice to double check apt's behavior here as well. @bluca Any clue if sharing the apt cache directory between distros, releases and architectures is fine?

bluca:
If it's just the archives/ subdir then it should be fine, as files will have different names

PR #2030, opened 2023-10-31
New ctl verbs

sam-leonard-ct:
Adds verbs for both Journalctl and coredumpctl.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2292,6 +2292,35 @@ def run_shell(args: MkosiArgs, config: MkosiConfig) -> None:
         run(cmdline, stdin=sys.stdin, stdout=sys.stdout, env=os.environ, log=False)
 
 
+def run_systemd_tool(tool: str, args: MkosiArgs, config: MkosiConfig) -> None:
+    if config.output_format not in (OutputFormat.disk, OutputFormat.directory):
+        die(f"{config.output_format} images cannot be inspected with {tool}")
+
+    if (tool_path := find_binary(tool)) is None:
+        die(f"Failed to find {tool}")
+
+    image_arg_name = "root" if config.output_format == OutputFormat.directory else "image"
+    run(
+        [
+            tool_path,
+            f"--{image_arg_name}={config.output_dir_or_cwd() / config.output}",
+            *args.cmdline
+        ],
+        stdin=sys.stdin,
+        stdout=sys.stdout,
+        env=os.environ,
+        log=False
+    )
+
+
+def run_journalctl(args: MkosiArgs, config: MkosiConfig) -> None:
+    run_systemd_tool("journalctl", args, config)
+
+
+def run_coredumpctl(args: MkosiArgs, config: MkosiConfig) -> None:
+    run_systemd_tool("coredumpctl", args, config)
+
+
 def run_serve(config: MkosiConfig) -> None:
     """Serve the output directory via a tiny HTTP server"""
 
@@ -2559,6 +2588,13 @@ def run_verb(args: MkosiArgs, images: Sequence[MkosiConfig]) -> None:
         if last.compress_output:
             die(f"Sorry, can't {opname} a compressed image.")
 
+    if (
+        args.verb in (Verb.journalctl, Verb.coredumpctl)
+        and last.output_format == OutputFormat.disk
+        and os.getuid() != 0
+    ):
+        die(f"Must be root to run the {args.verb} command")
+
     for config in images:
         if args.verb == Verb.build and not args.force:
             check_outputs(config)
@@ -2660,3 +2696,9 @@ def run_verb(args: MkosiArgs, images: Sequence[MkosiConfig]) -> None:
 
             if args.verb == Verb.serve:
                 run_serve(last)
+
+            if args.verb == Verb.journalctl:
+                run_journalctl(args, last)
+
+            if args.verb == Verb.coredumpctl:
+                run_coredumpctl(args, last)
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -56,12 +56,14 @@ class Verb(StrEnum):
     help          = enum.auto()
     genkey        = enum.auto()
     documentation = enum.auto()
+    journalctl    = enum.auto()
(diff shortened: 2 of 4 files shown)


DaanDeMeyer (changes requested):
We should use --root if we're doing a directory image and we should show an error if the output is neither disk nor directory

sam-leonard-ct:
I've pushed a fix, I haven't squashed it into the last one but am obviously very happy to do so if you would prefer.

mkosi/config.py
```diff
@@ -56,15 +56,20 @@ class Verb(StrEnum):
...
 
     def needs_build(self) -> bool:
-        return self in (Verb.build, Verb.shell, Verb.boot, Verb.qemu, Verb.serve)
+        return self in (Verb.build, Verb.shell, Verb.boot, Verb.qemu,
+                        Verb.serve, Verb.journalctl, Verb.coredumpctl)
 
     def needs_root(self) -> bool:
-        return self in (Verb.shell, Verb.boot)
+        return self in (Verb.shell, Verb.boot, Verb.journalctl,
+                        Verb.coredumpctl)
```
behrmann, line 72: No need to break the lines here, all of them are short enough to fit on a single line within our current line limit.

behrmann (approved):
I think it would make sense to squash the whole series into a single commit.

mkosi/config.py
sam-leonard-ct: Okay I will fix these and squash it, this was just where my editor broke it and I assumed it had found some config file somewhere to tell it to do that 🤷

sam-leonard-ct: This is fixed now and the commits have been squashed.

mkosi/__init__.py
```diff
@@ -2292,6 +2292,31 @@ def run_shell(args: MkosiArgs, config: MkosiConfig) -> None:
...
+    if (tool_path := find_binary(tool)) is None:
+        die(f"Failed to find {tool}")
+
+    image_arg_name = "root" if config.output_format == OutputFormat.directory else "image"
+    fname = config.output_dir_or_cwd() / config.output
+    cmdline: list[PathString] = [
+        tool_path,
+        f"--{image_arg_name}={fname}",
+        *args.cmdline
+    ]
```
DaanDeMeyer, line 2308: This doesn't need its own variable, please inline it in the run() call

mkosi/__init__.py
```diff
@@ -2292,6 +2292,31 @@ def run_shell(args: MkosiArgs, config: MkosiConfig) -> None:
...
 
+def run_ctl_with_name(tool: str, args: MkosiArgs, config: MkosiConfig) -> None:
+    if config.output_format not in (OutputFormat.disk, OutputFormat.directory):
+        die(f"{config.output_format} images cannot be inspected with {tool}")
+
+    if (tool_path := find_binary(tool)) is None:
+        die(f"Failed to find {tool}")
+
+    image_arg_name = "root" if config.output_format == OutputFormat.directory else "image"
+    fname = config.output_dir_or_cwd() / config.output
```
DaanDeMeyer, line 2303: This variable doesn't really have much use so I'd inline it

mkosi/__init__.py
```diff
@@ -2660,3 +2685,9 @@ def run_verb(args: MkosiArgs, images: Sequence[MkosiConfig]) -> None:
 
             if args.verb == Verb.serve:
                 run_serve(last)
+
+            if args.verb == Verb.journalctl:
+                return run_journalctl(args, last)
```
DaanDeMeyer, line 2690: Let's drop the return here and below since we don't do it for run_serve() either

mkosi/config.py
```diff
@@ -56,15 +56,17 @@ class Verb(StrEnum):
...
-        return self in (Verb.build, Verb.shell, Verb.boot, Verb.qemu, Verb.ssh)
+        return self in (Verb.build, Verb.shell, Verb.boot, Verb.qemu, Verb.ssh, Verb.journalctl, Verb.coredumpctl)
 
     def needs_build(self) -> bool:
-        return self in (Verb.build, Verb.shell, Verb.boot, Verb.qemu, Verb.serve)
+        return self in (Verb.build, Verb.shell, Verb.boot, Verb.qemu, Verb.serve, Verb.journalctl, Verb.coredumpctl)
 
     def needs_root(self) -> bool:
-        return self in (Verb.shell, Verb.boot)
+        return self in (Verb.shell, Verb.boot, Verb.journalctl, Verb.coredumpctl)
```
DaanDeMeyer, line 69: Do we need root if we're invoking this on a directory image?

mkosi/__init__.py
```diff
@@ -2292,6 +2292,31 @@ def run_shell(args: MkosiArgs, config: MkosiConfig) -> None:
         run(cmdline, stdin=sys.stdin, stdout=sys.stdout, env=os.environ, log=False)
 
 
+def run_ctl_with_name(tool: str, args: MkosiArgs, config: MkosiConfig) -> None:
```
DaanDeMeyer, line 2295: I'd name this run_systemd_tool()

DaanDeMeyer (changes requested):

mkosi/__init__.py
behrmann: It's funny. I was something similar yesterday, but when testing whether the type annotation could be omitted mypy wasn't smart enough to not make this `list[Object]`, but it works when inlining. Weird edge case.

sam-leonard-ct: behrmann I also had that exact issue yesterday 😂 and when I had some slightly more complicated logic for the cmdline that was all that fixed it. I will inline it now :)

mkosi/config.py
sam-leonard-ct: I'm not sure, I'm just going for lunch and will test in an hour or so.

sam-leonard-ct: Nope it doesn't, I have added a check as early as possible for it here: https://github.com/CodethinkLabs/mkosi/blob/bb63b8d74e4379f526fe0093de9370ffed270dc1/mkosi/__init__.py#L2591

PR #2033, opened 2023-11-01
qemu: allow checking device availability with actually used  flags

behrmann:
On an Arch machine I hadn't used mkosi with in a while I just stumbled over a permission denied although `/dev/vhost-vsock` was world writable. The `os.access` with flags `os.R_OK|os.W_OK` worked, but the `os.open` with `os.O_RDWR|os.O_CLOEXEC|os.O_NONBLOCK` did not. Though this just papers over the issue, let's check the access with the flags used for the open call to not just crash.

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -46,17 +46,21 @@ class QemuDeviceNode(StrEnum):
             QemuDeviceNode.vhost_vsock: "a VSock device",
         }[self]
 
-    def available(self, log: bool = False) -> bool:
+    def available(self, *, log: bool = False, flags: int = (os.R_OK | os.W_OK)) -> bool:
         if not os.access(self.device(), os.F_OK):
             if log:
                 logging.warning(f"{self.device()} not found. Not adding {self.description()} to the virtual machine.")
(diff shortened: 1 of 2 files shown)


behrmann:
It works after a reboot, so the reason most probably was a kernel module that couldn't be loaded anymore. So while this was broken elsewhere it's still preferable not to crash.

mkosi/qemu.py
```diff
@@ -46,17 +46,21 @@ def description(self) -> str:
...
+    def available(self, *, log: bool = False, flags: int = (os.R_OK | os.W_OK)) -> bool:
         if not os.access(self.device(), os.F_OK):
             if log:
                 logging.warning(f"{self.device()} not found. Not adding {self.description()} to the virtual machine.")
             return False
 
-        if not os.access(self.device(), os.R_OK|os.W_OK):
+        try:
+            fd = os.open(self.device(), flags)
+            os.close(fd)
```
DaanDeMeyer, line 57: Just merge these?

DaanDeMeyer (approved):

PR #2038, opened 2023-11-02
config: right strip newlines from contents read via path_read_text

behrmann:
This is an attempted fix for #2031. I'm not super happy about possible fallout from this, because of the silent stripping of newlines, but let's see whether this works. @NekkoDroid could you have a look whether this solves your issue?

PR #2039, opened 2023-11-03
Skip symlink when prefixing everything in staging directory

DaanDeMeyer:
The symlink shouldn't be prefixed so make sure we skip it.


NekkoDroid:
Just tested it: the repro I posted in #2038 is able to build successfully on the branch.

PR #2040, opened 2023-11-05
Add kernel-install plugin from mkosi-initrd

DaanDeMeyer:
Let's support kernel-install natively in mkosi. This commit moves
the kernel-install plugin from mkosi-initrd into mkosi itself with
a few adaptations:

- We look for initrd configuration in /usr/lib/mkosi-initrd and
  /etc/mkosi-initrd using "--include"
- We always include the host's kernel modules using --extra-tree
- We skip the mkosi-initrd plugin unless initrd_generator is
  explicitly set to "mkosi-initrd". This allows the mkosi package
  to be installed without also using it as the initrd generator if
  initrd_generator is not configured explicitly.

This commit also extends the default fedora image to build the mkosi
rpm if the spec sources are mounted at rpm/ using BuildSources=. We
also configure the default image to use mkosi-initrd as the initrd
generator and ukify as the uki generator, so that running kernel-install
in the booted image will build a UKI using mkosi-initrd and ukify.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,19 @@ when not installed as a zipapp.
 Please note, that the python module exists solely for the usage of the
 mkosi binary and is not to be considered a public API.
 
+## kernel-install plugin
+
+mkosi can also be used as a kernel-install plugin to build initrds. To
+enable this feature, install `kernel-install/50-mkosi-initrd.install`
+into `/usr/lib/kernel/install.d` and install `mkosi-initrd/mkosi.conf`
+into `/usr/lib/mkosi-initrd`. Extra distro configuration for the initrd
+can be configured using drop-ins in `/usr/lib/mkosi-initrd`. Users can
+add their custom configuration in `/etc/mkosi-initrd`.
+
+Once installed, the mkosi plugin can be enabled by writing
+`initrd_generator=mkosi-initrd` to `/usr/lib/kernel/install.conf` or to
+`/etc/kernel/install.conf`.
+
 # Hacking on mkosi
 
 To hack on mkosi itself you will also need
--- a/kernel-install/50-mkosi-initrd.install
+++ b/kernel-install/50-mkosi-initrd.install
@@ -1,0 +1,44 @@
+#!/usr/bin/bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
+set -e
+export LANG=C.UTF-8
+
+COMMAND="$1"
+KERNEL_VERSION="$2"
+BOOT_DIR_ABS="$3"
+KERNEL_IMAGE="$4"
+INITRD_OPTIONS_SHIFT=4
+
+# Skip this plugin if we're using a different generator.
+[ "${KERNEL_INSTALL_INITRD_GENERATOR}" != "mkosi-initrd" ] && exit 0
+
+: "${KERNEL_INSTALL_STAGING_AREA:?}"
+
+case "$COMMAND" in
+    add)
+        # If the initrd was provided on the kernel command line, we shouldn't generate our own.
+        [ "$#" -gt "$INITRD_OPTIONS_SHIFT" ] && exit 0
+
+        (
+            [ "$KERNEL_INSTALL_VERBOSE" -gt 0 ] && set -x
+
+            mkosi --directory "" \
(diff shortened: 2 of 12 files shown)


docs/building-rpms-from-source.md
```diff
@@ -128,12 +128,12 @@ cd "$SRCDIR/mkosi"
...
     --build-in-place \
-    "$([ "$WITH_TESTS" = "0" ] && echo --nocheck)" \
+    $([ "$WITH_TESTS" = "0" ] && echo --nocheck) \
     --define "_topdir $SRCDIR/mkosi" \
     --define "_sourcedir $SRCDIR/mkosi/rpm" \
     --define "_rpmdir $OUTPUTDIR" \
-    "$([ -n "$BUILDDIR" ] && echo --define)" \
-    "$([ -n "$BUILDDIR" ] && echo "_vpath_builddir $BUILDDIR")" \
+    $([ -n "$BUILDDIR" ] && echo --define) \
+    $([ -n "$BUILDDIR" ] && echo "_vpath_builddir $BUILDDIR") \
```
behrmann, line 136:
```suggestion
    ${BUILDDIR:+--define} \
    ${BUILDDIR:+"_vpath_builddir $BUILDDIR"} \
```
or maybe
```suggestion
    ${BUILDDIR:+--define "_vpath_builddir $BUILDDIR"} \
```

kernel-install/50-mkosi-initrd.install
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
+set -e
+export LANG=C
```
keszybz, line 4: LANG=C.UTF-8 would be better, but if this is just importing the file from `mkosi-initrd`, then it's probably better to do that in a separate step for clarity.

keszybz (commented):
Looks reasonable, but I'd like to test this before merging.

DaanDeMeyer:
I tested in the fedora image we include in mkosi but happy to keep this open if you want to give it a more thorough test.

README.md
```diff
@@ -78,6 +78,19 @@ when not installed as a zipapp.
...
+enable this feature, install `kernel-install/50-mkosi-initrd.install`
+into `/usr/lib/kernel/install.d` and install `mkosi-initrd/mkosi.conf`
+into `/usr/lib/mkosi-initrd`. Extra distro configuration for the initrd
+can be configured using drop-ins in `/usr/lib/mkosi-initrd`. Users can
+add their custom configuration in `/etc/mkosi-initrd`.
+
+Once installed, the mkosi plugin can be enabled by writing
+`initrd_generator=mkosi-initrd` to `/usr/lib/kernel/install.conf` or to
+`/etc/kernel/install.conf`.
+
```
behrmann, line 93: This should also note, that this so far has only been tested for Fedora?

mkosi-initrd/mkosi.conf.d/10-arch.conf
```diff
@@ -0,0 +1,17 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+[Match]
+Distribution=arch
+
+[Content]
+Packages=
+        gzip # For compressed keymap unpacking by loadkeys
```
behrmann, line 8:
```suggestion
        gzip # For compressed keymap unpacking by loadkeys
        tpm2-tss
        libfido2
```

mkosi-initrd/mkosi.conf.d/10-debian-ubuntu.conf
```diff
@@ -0,0 +1,10 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+[Match]
+Distribution=|debian
+Distribution=|ubuntu
+
+[Content]
+Packages=
+        kmod     # Not pulled in as a dependency on Debian/Ubuntu
+        dmsetup  # Not pulled in as a dependency on Debian/Ubuntu
```
behrmann, line 10:
```suggestion
        dmsetup  # Not pulled in as a dependency on Debian/Ubuntu
        libfido2-1
        libtss2-tctildr0
```

behrmann (approved):
LGTM, I left the package names of what *I think* are the equivalents for the packages that were so far only in the fedora config.

keszybz:
Seems to work as expected in Fedora.

PR #2041, opened 2023-11-06

DaanDeMeyer:
Split out of #2040

### DIFF
--- a/docs/building-rpms-from-source.md
+++ b/docs/building-rpms-from-source.md
@@ -152,8 +152,5 @@ post-installation script:
 #!/bin/sh
 set -e
 
-dnf --cacheonly install "$OUTPUTDIR"/*mkosi*.rpm
+rpm --install "$OUTPUTDIR"/*mkosi*.rpm
 ```
-
-We make sure `dnf` never tries to sync the repositories when installing
-these local rpms by specifying `--cacheonly`.
--- a/docs/building-rpms-from-source.md
+++ b/docs/building-rpms-from-source.md
@@ -62,9 +62,9 @@ mkosi-chroot \
     "$DEPS" \
     --define "_topdir $CHROOT_SRCDIR/mkosi" \
     --define "_sourcedir $CHROOT_SRCDIR/mkosi/rpm" \
-    "$CHROOT_SRCDIR/mkosi/rpm/mkosi.spec" \
-    | grep -E -v "mkosi" \
-    | xargs -d '\n' dnf install --best
+    "$CHROOT_SRCDIR/mkosi/rpm/mkosi.spec" |
+        grep -E -v "mkosi" |
+        xargs -d '\n' dnf install --best
 
 if [ "$1" = "build" ]; then
     until mkosi-chroot \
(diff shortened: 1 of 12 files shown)


mkosi/config.py
```diff
@@ -844,8 +857,17 @@ class MkosiConfig:
...
-    def workspace_dir_or_cwd(self) -> Path:
-        return self.workspace_dir or Path.cwd()
+    def workspace_dir_or_default(self) -> Path:
+        if self.workspace_dir:
+            return self.workspace_dir
+
+        if os.getenv("XDG_CACHE_HOME") and Path(os.environ["XDG_CACHE_HOME"]).exists():
+            return Path(os.environ["XDG_CACHE_HOME"])
+
+        if os.getenv("HOME") and "/home" in os.environ["HOME"] and (Path(os.environ["HOME"]) / ".cache").exists():
```
behrmann, line 867: If mypy doesn't force us to (haven't checked) I think staying with `os.getenv` is a bit more uniform and nicer
```suggestion
        if os.getenv("XDG_CACHE_HOME") and Path(os.getenv("XDG_CACHE_HOME")).exists():
            return Path(os.getenv("XDG_CACHE_HOME"))

        if os.getenv("HOME") and "/home" in os.getenv("HOME") and (Path.home() / ".cache").exists():
```

I'm not sure about the `"/home" in os.getenv("HOME")` check. While generally discouraged, I've seen orgs where home directories where not under `/home` for weird legacy reasons, also, I use mkosi with system users who have their homes under `/var/lib`. I tend to configure cache directories explicitly there, but I'm not sure we should force that.

behrmann (commented):
I really like the ephemeral source directory change, it should be noted in the news, though.

docs/building-rpms-from-source.md
```diff
@@ -62,9 +62,9 @@ mkosi-chroot \
     "$DEPS" \
     --define "_topdir $CHROOT_SRCDIR/mkosi" \
     --define "_sourcedir $CHROOT_SRCDIR/mkosi/rpm" \
-    "$CHROOT_SRCDIR/mkosi/rpm/mkosi.spec" \
-    | grep -E -v "mkosi" \
-    | xargs -d '\n' dnf install --best
+    "$CHROOT_SRCDIR/mkosi/rpm/mkosi.spec" |
+    grep -E -v "mkosi" |
+    xargs -d '\n' dnf install --best
```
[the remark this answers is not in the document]
behrmann: It makes sense in that commands are visually distinct from the arguments.

mkosi/config.py
behrmann: Sounds reasonable enough, but a comment with that reasoning would probably make sense as well as a comment in the man page, that the fallback to `~/.cache` will only be used for non-system users, even if `$HOME` is set for a system user, since this might be a potential footgun in the future with `SetLoginEnvironment=`.

mkosi/config.py
```diff
@@ -2900,17 +2921,18 @@ def root_password_transformer(
             return None
         return (cast(str, rootpw[0]), cast(bool, rootpw[1]))
 
-    def source_target_transformer(
-        trees: list[list[Optional[str]]], fieldtype: type[list[tuple[Path, Path]]]
-    ) -> list[tuple[Path, Path]]:
+    def tree_transformer(trees: list[dict[str, Any]], fieldtype: type[ConfigTree]) -> list[ConfigTree]:
```
behrmann, line 2924:
```suggestion
    def config_tree_transformer(trees: list[dict[str, Any]], fieldtype: type[ConfigTree]) -> list[ConfigTree]:
```
to stay closer to the name

behrmann (approved):

mkosi/config.py
```diff
@@ -844,8 +848,26 @@ class MkosiConfig:
...
+        if (
+            (home := os.getenv("HOME")) and
+            "/home" in home and
+            self.cache_dir and
+            self.cache_dir.is_relative_to(home) and
+            (Path(home) / ".cache").exists()
+        ):
+            return Path(home) / ".cache"
+
+        return Path("/var/tmp")
```
marc-hb, line 870: I'm curious why this last fallback went (back?) to `/var/tmp` and not `/tmp`? The latter is often `tmpfs`, the former never.

Afraid to use too much RAM maybe?

More context in:
- https://github.com/pmem/run_qemu/pull/116

DaanDeMeyer: @marc-hb Indeed, we write the entire image to the workspace directory, which can potentially be multiple gigabytes, we don't want to put all of that in RAM.

PR #2043, opened 2023-11-06
Fix broken KernelModuleIncludes implementation

jnohlgard:
The new implementation behaves as described in the documentation, i.e.
the inclusion patterns will override any exclusions. The old
implementation was broken for the include case, if anything matched an
exclusion pattern it would not get added to the image regardless of any
inclusion patterns.


DaanDeMeyer:
@jnohlgard https://github.com/systemd/mkosi/pull/2045 should also fix the issue, can you give that a try? If that works for you, you can drop the second commit from the PR.

DaanDeMeyer:
I dropped the second commit from the PR but kept the first one

DaanDeMeyer (approved):

jnohlgard:
Thank you @DaanDeMeyer. Your alternative fix looks good, but I did not have time to try it yet.
Thanks for taking care of the changes to this PR 👍

PR #2046, opened 2023-11-06
Add BuildSourcesEphemeral=

DaanDeMeyer:
Unfortunately there are use cases where it's useful to persist changes to the source directory. A prime example is projects with Makefile that do not provide a properly functioning `make install`. The only way to use those with mkosi build scripts is to do an in-tree build in the build script and then mounting the source directory into the VM or container when booting it. For this to work we need changes to the source directory to be persisted. To support this use case, let's gate the ephemeral source directories behind an option that's disabled by default (for backwards compatibitity reasons).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -38,9 +38,10 @@
   `-kernel` or `QemuKernel=`
 - We don't create subdirectories beneath the configured cache directory
   anymore.
-- Source directories are now made ephemeral when running scripts. This
-  means any changes made to source directories while running scripts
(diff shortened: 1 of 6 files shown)


behrmann (approved):
LGTM, I would have probably made the check to not run `make_image` into an early return in `make_image`, but either is fine.

DaanDeMeyer:
Yeah I thought the same first but I feel its weird to have the function check both an argument and a config option for the same thing.

PR #2049, opened 2023-11-07
Add BuildSourcesEphemeral= to the rpms from source doc.

DaanDeMeyer:
rpmbuild with --build-in-place is a top offender when it comes to writing files to the source directory, so let's suggest enabling BuildSourcesEphemeral= in the building rpms from source doc.


behrmann:
Merging, since it is a doc-only change.

PR #2050, opened 2023-11-07
Add QemuDrives= option

DaanDeMeyer:
This option allows specifying extra qemu drives to pass to qemu. mkosi will create a file (optionally in the given directory) of the given size and pass it to qemu via -drive, optionally with some extra options.

Doing this in mkosi allows mkosi to automatically manage the lifetime of the file backing the drive. We can create it as needed and remove it when we exit.


mkosi/config.py
```diff
@@ -527,6 +539,29 @@ def config_parse_profile(value: Optional[str], old: Optional[int] = None) -> Opt
     return value
 
 
+def parse_drive(value: str) -> ConfigDrive:
+    parts = value.split(":")
```
behrmann, line 543: can qemu options relevant to drives contain colons? If so 
```suggestion
    parts = value.split(":", maxsplit=3)
```

tests/test_json.py
```diff
@@ -186,6 +187,14 @@ def test_config() -> None:
             "Profile": "profile",
             "QemuArgs": [],
             "QemuCdrom": false,
+            "QemuDrives": [
+                {
+                    "directory": "/foo/bar",
+                    "id": "abc",
+                    "options": "abc,qed",
+                    "size": 200
+                }
```
behrmann, line 196: Let's add the directory null case for completeness
```suggestion
                {
                    "directory": "/foo/bar",
                    "id": "abc",
                    "options": "abc,qed",
                    "size": 200
                },
                {
                    "directory": null,
                    "id": "def",
                    "options": "",
                    "size": 123456
                }
```

behrmann (approved):
LGTM, a question and a nit. Also the name `ConfigDrive` is a bit weird, as I think of a drive for config. How about `DriveConfig` or `QemuDrive`?

PR #2051, opened 2023-11-08
Use package manager specific subdirectories in the cache directory

DaanDeMeyer:
We want to be able to set --cache-dir=/var/cache to automatically reuse the system cache for each installed package manager, so let's make sure we use the right subdirectories for each package manager to make that work properly.


behrmann:
LGTM, but shouldn't this maybe be a noop if somebody alreadt set, e.g. `--cache-dir=/var/cache/pacman/pkg`?

behrmann:
Thinking about it, no probably not.

behrmann (approved):

PR #2053, opened 2023-11-08
building rpms from source doc improvements

DaanDeMeyer:
- Use env --chdir= instead of spawning a shell
- Document that scripts are executed in the same directory that mkosi is running in and get rid of all uses of $SRCDIR and $CHROOT_SRCDIR in favor of relative paths. We have to use $PWD when actually building the rpm as some macros expect the topdir to be an absolute path.


behrmann (approved):
Nice! Much more readable.

PR #2055, opened 2023-11-08
don't fail on non-controlling ttys

katexochen:
I run into this issue where the tty is non-controlling and this call failed.

### DIFF
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -6,6 +6,7 @@ import contextlib
 import ctypes
 import ctypes.util
 import enum
+import errno
 import fcntl
 import logging
 import os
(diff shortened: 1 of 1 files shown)


DaanDeMeyer (commented):
This leaves the signal blocked. We need to unblock the signal even if we fail.

katexochen:
Sorry, I'm really not good with the tty stuff. How do I unblock the signal? Should I skip the return and continue with `signal.signal(signal.SIGTTOU, old)`?

DaanDeMeyer:
Yes

DaanDeMeyer:
Please put the raise on a newline like we do everywhere else

DaanDeMeyer (approved):

PR #2056, opened 2023-11-08
Make initrd work with device mapper

keszybz:


mkosi-initrd/mkosi.finalize
```diff
@@ -0,0 +1,40 @@
...
+    return pathlib.Path(os.getenv('BUILDROOT'))
+
+def mask_units(root):
+    units = [
+        # Dracut installs a rule which sets OPTIONS+="db_persist" on all dm devices [1]
+        # (According to codesearch.debian.net, it is the only user of db_persist.)
+        # Without this, all dm units end up with SYSTEMD_READY=0 and systemd thinks the
+        # device units are missing. The system boots fine, but an attempt to call
+        # daemon-reexec or daemon-reload ends with anything that can be stopped or unmounted
+        # being purged.
```
DaanDeMeyer, line 18: This is confusing to me. Are you saying that if the `db_persist` option is set, the system boots fine but daemon-reexec or daemon-reload end up purging all the lvm stuff?

DaanDeMeyer (commented):
I don't mind shipping this temporarily, but I'd like to at least track upstream either in lvm or systemd an issue that allows us to remove this again to make this work OOTB. I don't know whether that means not pulling in `initrd-udevadm-cleanup-db.service` by default in systemd or moving the udev rules from dracut to the lvm2 package.

poettering:
Uh, I'd be careful with that. Right now, unless you throw LVM into the mix (which is a mistake, but whatever), and you transition into the host OS then because there is no db it also means all devices are considered absent by systemd. Only once you do "udevadm trigger" they start to appear.

If you keep the whole db, then things change drastically: suddenly all devices are always there, and they always are marked initialized via a db entry.

Moreover, traditionally the ruleset of the initrd and of the host was quite different. Not just smaller, but also different, because again LVM is bad and they wanted it that way. (Yes, they do ship a different set of rules in the initrd than on the host). By not clearing the db you inherited all that special casing for the initrd into the host.

So I'd be very careful with this. This doesn't sound like something we should do. I'd always start fresh here, it's the safer choice.

Let LVM deal with it's fallout, and convince them to just ship a rule for db_persist that only applies to the initrd, i.e. that carries `…, TEST="/etc/initrd-release", …` and be done with it.

poettering:
Also note that `systemd-udev-trigger.service` fires "add" events (not "change" events).

it's just a really bad idea to fire "add" if the db entries already exist for all devices.

Hence, sorry, but it's not that easy, this PR seems like a hack to me. A hack that might be fine if focussed on LVM, but just doing this universally sounds bad to me. I see no benefit in that.

keszybz:
A correction: I wrote about LVM, because this is what the patch was initially focused on, but this applies to DM in general. Also plain LUKS decryption. On my laptop with just GPT+LUKS+BTRFS, the machine doesn't boot without this patch.

> Are you saying that if the db_persist option is set, the system boots fine but daemon-reexec or daemon-reload end up purging all the lvm stuff? 

It's the opposite ;). `db_persist` is an opt-out of the device cleanup and it is used by LVM/DM to survive the switch-root. Without this, the conditions are set, and a subsequent dameon-reload/daemon-reexec triggers the issue. AFAIK, it's only set by some dracut rules and nothing else.

Ideally, we would just drop `initrd-udevadm-cleanup-db.service` from systemd. The whole idea with yanking out devices from a running system and expecting that everything will continue to run seems backwards.

I think it's too early to do this in systemd yet: `mkosi-initrd` (either the original form or the new built-in version) are fairly new. Once we've confirmed that it works for most systems, then we can figure out how to push changes into systemd upstream without causing regressions for other initrd types.

> Moreover, traditionally the ruleset of the initrd and of the host was quite different. 

Right. But "traditionally" is not relevant here: we're talking about building an initrd from the packaged software. One of the primary goals is to get rid of the little special dances and magic incantations that were done by dracut. And fundamentally, I think we should just drop this idea that "initrd is special" and everything must be done with a separate custom set of rules. New mode: the rules in the initrd are the same, just a smaller set of packages is installed. 

> Also note that `systemd-udev-trigger.service` fires "add" events (not "change" events).

This is something worth keeping in mind. I don't see immediate problems: it seems that for most things "add" is idempotent. I think we should write rules so that they are idempotent, otherwise things are going to be very fragile.

keszybz:
> I disagree, the initrd is a different system. We should be careful with what we pass over, and it should be minimized. The fact that LVM leaks so much context sounds backwards to me, and is what should be fixed.

~~LVM~~ device mapper.

OK, so do you have an actual proposal how to support systems with real storage? Did you test the initrds produced with mkosi on real hardware and did they work for you?

DaanDeMeyer:
Wouldn't the way forward here be to copy the `db_persist` rules from dracut while also following up with the relevant upstream to have them ship the `db_persist` rules themselves instead of maintaining them in dracut?

poettering:
> OK, so do you have an actual proposal how to support systems with real storage? Did you test the initrds produced with mkosi on real hardware and did they work for you?

Yeah, as @DaanDeMeyer said, I think the dracut rule should just go upstream into LVM. they ship a ton of rules anyway, they can add that one too

poettering:
> Let LVM deal with it's fallout, and convince them to just ship a rule for db_persist that only applies to the initrd, i.e. that carries `…, TEST="/etc/initrd-release", …` and be done with it.

Actually, the `TEST=` thing is unnecessary, the db_persist flag doesn't hurt on the main system, hence let's simplify things

keszybz:
I really don't understand why you think that making the `db_persist` behaviour opt-in is the right approach. It seems that it is needed in all cases where something depends on the device existing. I think that in cases of some plain mounts it is not needed because we have code that "pins" the device if the kernel reports the mount. So we don't fail if `/dev/sda1` goes away, but that is only because we actually pretend it still exists. Of course for most devices (ttys, serial ttys, network devices, etc.) we don't have anything depending on them, so nothing happens if they go away during early boot. But it's possible that somebody will add a dependency later, and then we'd also want to set `db_persist`. So overall, we either have devices for which we know `db_persist` is required, or devices for which currently it's not needed but might be in the future. Dropping `initrd-udevadm-cleanup-db.service` effectively makes `db_presist` apply to all devices and it seems the right thing to do.

DaanDeMeyer:
Regardless of whether that's the right thing to do, it's something that should be changed and discussed in systemd itself I think. I'd much prefer that initially we stick to doing this the same as dracut and the other initramfs generators do it and then we switch over automatically once this is dropped in systemd (if that turns out to be the right approach).

DaanDeMeyer:
Given that #2108 was merged, let's close this one. I'm not saying we shouldn't do this, but it's probably something to change in systemd rather than override in mkosi-initrd.

PR #2058, opened 2023-11-09
Make default workspace directory logic more robust

DaanDeMeyer:
When running with sudo, $HOME won't be set, so let's rework the logic to not rely on that.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -893,17 +893,18 @@ class MkosiConfig:
         if (cache := os.getenv("XDG_CACHE_HOME")) and Path(cache).exists():
             return Path(cache)
 
-        # If there's a cache directory and we're running from /home, we want to use a workspace directory in /home as
-        # well as /home might be on a separate partition or subvolume which means that to take advantage of reflinks
-        # and such, the workspace directory has to be on the same partition/subvolume.
-        if (
-            (home := os.getenv("HOME")) and
-            "/home" in home and
-            self.cache_dir and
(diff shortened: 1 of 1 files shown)


mkosi/config.py
```diff
@@ -893,17 +893,18 @@ def workspace_dir_or_default(self) -> Path:
...
-            "/home" in home and
-            self.cache_dir and
-            self.cache_dir.is_relative_to(home) and
-            (Path(home) / ".cache").exists()
-        ):
-            return Path(home) / ".cache"
+        # If we're running from /home and there's a cache or output directory in /home, we want to use a workspace
+        # directory in /home as well as /home might be on a separate partition or subvolume which means that to take
+        # advantage of reflinks and such, the workspace directory has to be on the same partition/subvolume.
+        if (Path.cwd().is_relative_to("/home") and len(Path.cwd().parents) >= 3):
```
behrmann, line 899: No need for the parentheses.
```suggestion
        if Path.cwd().is_relative_to("/home") and len(Path.cwd().parents) >= 3:
```

mkosi/config.py
```diff
@@ -893,17 +893,18 @@ def workspace_dir_or_default(self) -> Path:
...
+            home = Path.cwd().parents[-3]
```
behrmann, line 900: The check with the with the length is a bit obscure, how about
```suggestion
        if Path.cwd().is_relative_to(Path.home()):
            home = Path.home()
```

PR #2061, opened 2023-11-10
mkosi: add "burn" verb 🔥🔥🔥

poettering:
This simply dd's the resulting image to a block device

i.e. run "mkosi -f burn /dev/sda"

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -23,6 +23,7 @@ from typing import Optional, TextIO, Union
 
 from mkosi.architecture import Architecture
 from mkosi.archive import extract_tar, make_cpio, make_tar
+from mkosi.burn import run_burn
 from mkosi.config import (
     BiosBootloader,
     Bootloader,
@@ -2868,3 +2869,6 @@ def run_verb(args: MkosiArgs, images: Sequence[MkosiConfig]) -> None:
 
             if args.verb == Verb.coredumpctl:
                 run_coredumpctl(args, last)
+
+            if args.verb == Verb.burn:
+                run_burn(args, last)
--- a/mkosi/burn.py
+++ b/mkosi/burn.py
@@ -1,0 +1,40 @@
+# SPDX-License-Identifier: LGPL-2.1+
+
+import os
+import sys
+
+from mkosi.config import MkosiArgs, MkosiConfig, OutputFormat
+from mkosi.log import complete_step, die
+from mkosi.run import run
+
+
+def run_burn(args: MkosiArgs, config: MkosiConfig) -> None:
+    if config.output_format not in (OutputFormat.disk, OutputFormat.esp):
+        die(f"{config.output_format} images cannot be burned to disk")
+
+    fname = config.output_dir_or_cwd() / config.output
+
+    if len(args.cmdline) != 1:
+        die("Expected device argument.");
+
+    device = args.cmdline[0]
+
+    cmd = [
+        "systemd-repart",
+        "--no-pager",
+        "--pretty=no",
+        "--offline=yes",
+        "--empty=force",
+        "--dry-run=no",
+        f"--copy-from={fname}",
+        device,
+    ]
+
+    with complete_step("Burning 🔥🔥🔥 to medium…", "Burnt. 🔥🔥🔥"):
+        run(
+            cmd,
+            stdin=sys.stdin,
+            stdout=sys.stdout,
+            env=os.environ,
+            log=False,
+        )
--- a/mkosi/resources/mkosi.md
+++ b/mkosi/resources/mkosi.md
@@ -28,6 +28,8 @@ mkosi — Build Bespoke OS Images
 
 `mkosi [options…] serve`
 
+`mkosi [options…] burn <device>`
+
 `mkosi [options…] bump`
 
 `mkosi [options…] genkey`
@@ -124,6 +126,13 @@ The following command line verbs are known:
   it. This command is useful for testing network based acquisition of
   OS images, for example via `machinectl pull-raw …` and `machinectl
   pull-tar …`.
+
+`burn <device>`
+
+: This builds the image if it is not built yet, and then writes it to the
+  specified block device. The partition contents are written as-is, but the GPT
+  partition table is corrected to match sector and disk size of the specified
+  medium.
 
 `bump`
 
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -58,15 +58,16 @@ class Verb(StrEnum):
     documentation = enum.auto()
     journalctl    = enum.auto()
     coredumpctl   = enum.auto()
+    burn          = enum.auto()
 
     def supports_cmdline(self) -> bool:
-        return self in (Verb.build, Verb.shell, Verb.boot, Verb.qemu, Verb.ssh, Verb.journalctl, Verb.coredumpctl)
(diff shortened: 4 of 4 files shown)


mkosi/__init__.py
```diff
@@ -50,6 +50,7 @@
 from mkosi.pager import page
 from mkosi.partition import Partition, finalize_root, finalize_roothash
 from mkosi.qemu import QemuDeviceNode, copy_ephemeral, run_qemu, run_ssh
+from mkosi.burn import run_burn
```
behrmann, line 53: CI complains because this should be between `mkosi.architecture` and `mkosi.config` in the imports.

mkosi/burn.py
```diff
@@ -0,0 +1,44 @@
...
+from mkosi.config import (
+    MkosiArgs,
+    MkosiConfig,
+    OutputFormat,
+)
+from mkosi.log import die
+from mkosi.run import run
+
+def run_burn(args: MkosiArgs, config: MkosiConfig) -> None:
+
```
behrmann, line 15: Please two empty lines between top-level definitions and none after the signature.
```suggestion

def run_burn(args: MkosiArgs, config: MkosiConfig) -> None:
```

mkosi/burn.py
```diff
@@ -0,0 +1,44 @@
...
+    if config.output_format not in (
+        OutputFormat.disk,
+        OutputFormat.esp,
+    ):
```
behrmann, line 19: That's short enough to fit on a line within our line length limit.
```suggestion
    if config.output_format not in (OutputFormat.disk, OutputFormat.esp):
```

mkosi/burn.py
```diff
@@ -0,0 +1,44 @@
...
+
+    device = args.cmdline[0]
+
+    cmd = [
+        "dd",
+        f"if={fname}",
+        f"of={device}",
+        "bs=4096",
+        "status=progress",
+        "conv=nocreat,notrunc,fsync"
```
behrmann, line 35: A trailing comma would be nice
```suggestion
        "conv=nocreat,notrunc,fsync",
```

mkosi/burn.py
```diff
@@ -0,0 +1,44 @@
...
+
+    fname = config.output_dir_or_cwd() / config.output
+
+    if (len(args.cmdline) != 1):
+            die("Expected device argument.");
+
+    device = args.cmdline[0]
+
+    cmd = [
+        "dd",
```
behrmann, line 30: While I generally also `dd` things onto devices, I've been told that this is mostly just muscle-memory and a simple `cp` would just work as well and have `cp` figure out the block size and stuff. I have no strong opinion on this, but do you have any thoughts?

mkosi/burn.py
```diff
@@ -0,0 +1,44 @@
...
+        OutputFormat.esp,
+    ):
+        die(f"{config.output_format} images cannot be dd'ed to disk")
+
+    fname = config.output_dir_or_cwd() / config.output
+
+    if (len(args.cmdline) != 1):
+            die("Expected device argument.");
+
+    device = args.cmdline[0]
```
behrmann, line 27: Should we maybe do some sanity check before passing this on? e.g. wrapping this in a ``Path()` and then doing a `device.is_block_device()` check?

behrmann (commented):
Looking good! I'm surprised there are no fire emojis in any output :)

Cornelicorn:
Is there any reason for using `dd` or `cp` instead of `systemd-repart --image image /dev/sda` which would also apply repart definitions in the image directly?

DaanDeMeyer:
@Cornelicorn `--image=` has a few issues, especially when there's verity included, since we'll refuse to mount the image unless the verity cert is enrolled in the correct location on the host machine. `--copy-from=` is what should be used here.

poettering:
> Looking good! I'm surprised there are no fire emojis in any output :)

Shame on me! I don't know what to say. This is on me. I made a horrible mistake. I will fix this!

mkosi/burn.py
poettering: changed to repart

mkosi/burn.py
poettering: a regular file is fine too.

repart does its own checking really, no need to replicate this here.

DaanDeMeyer (changes requested):
So looks great except I think we should replace dd with systemd-repart --copy-from=. And needs documentation.

poettering:
Force pushed new version. PTAL.

poettering:
Hmm, the CI claims i didn't sort the imports? What the fuck does it want from me?

```
import os
import sys

from mkosi.config import (
    MkosiArgs,
    MkosiConfig,
    OutputFormat,
)
from mkosi.log import complete_step, die
from mkosi.run import run
```

what's not sorted about that?

mkosi/burn.py
```diff
@@ -0,0 +1,44 @@
+# SPDX-License-Identifier: LGPL-2.1+
+
+import os
+import sys
+
+from mkosi.config import (
+    MkosiArgs,
+    MkosiConfig,
+    OutputFormat,
+)
```
behrmann, line 10: The complaint's unclear, because it's sorted, but I just run isort on it and it basically just complains about line lenght
```suggestion
from mkosi.config import MkosiArgs, MkosiConfig, OutputFormat
```

mkosi/burn.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def run_burn(args: MkosiArgs, config: MkosiConfig) -> None:
+    if config.output_format not in (OutputFormat.disk, OutputFormat.esp):
+        die(f"{config.output_format} images cannot be dd'ed to disk")
+
+    fname = config.output_dir_or_cwd() / config.output
+
+    if (len(args.cmdline) != 1):
+            die("Expected device argument.");
```
behrmann, line 22: No unneeded parentheses around conditions and the indentation is a bit too far (8 instead of 4).
```suggestion
    if len(args.cmdline) != 1:
        die("Expected device argument.");
```

mkosi/burn.py
```diff
@@ -0,0 +1,44 @@
...
+    MkosiConfig,
+    OutputFormat,
+)
+from mkosi.log import complete_step, die
+from mkosi.run import run
+
+
+def run_burn(args: MkosiArgs, config: MkosiConfig) -> None:
+    if config.output_format not in (OutputFormat.disk, OutputFormat.esp):
+        die(f"{config.output_format} images cannot be dd'ed to disk")
```
Cornelicorn, line 17: Since dd isn't used anymore, maybe
```suggestion
        die(f"{config.output_format} images cannot be burned to disk")
```

poettering:
Pushed another version. Let's see if CI passes this time

mkosi/resources/mkosi.md
```diff
@@ -125,6 +127,12 @@ The following command line verbs are known:
   OS images, for example via `machinectl pull-raw …` and `machinectl
   pull-tar …`.
 
+`burn <device>`
+
+: This builds the image if it is not built yet, and then writes it to the
+  specified block device. The partition contents is written as-is, but the GPT
```
behrmann, line 133: Either 
```suggestion
  specified block device. The partition content is written as-is, but the GPT
```
or
```suggestion
  specified block device. The partition contents are written as-is, but the GPT
```
My preference is on the latter.

behrmann (approved):

mkosi/resources/mkosi.md
poettering: changed as requested

poettering:
yay all passed! merging!

PR #2062, opened 2023-11-10
Only mount over /etc/resolv.conf if network access is enabled

DaanDeMeyer:
This allows users to mess with /etc/resolv.conf in postinst scripts without having to unmount it first.


keszybz (approved):
LGTM.

PR #2063, opened 2023-11-10
Debug log when firmware is missing

DaanDeMeyer:
Sometimes missing firmware can cause issues in the initrd so let's at least debug log if we can't find some firmware.


keszybz (approved):
LGTM.

PR #2064, opened 2023-11-10
Drop compression suffix for directory output

keszybz:
The automatic logic would produce a name like "initrd_6.6.0.zst", even though there is no compression.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -116,6 +116,11 @@ class OutputFormat(StrEnum):
             OutputFormat.uki:  ".efi",
         }.get(self, "")
 
+    def use_outer_compression(self) -> bool:
+        return self in (OutputFormat.tar,
+                        OutputFormat.cpio,
+                        OutputFormat.disk)
+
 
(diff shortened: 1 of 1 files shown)


mkosi/config.py
```diff
@@ -116,6 +116,11 @@ def extension(self) -> str:
             OutputFormat.uki:  ".efi",
         }.get(self, "")
 
+    def can_compress(self):
```
DaanDeMeyer, line 119: This name isn't entirely right. UKI and ESP can compress, they're just not compressed at the outer level. Instead the inner contents are compressed. Not sure what a better name would be. `is_self_decompressing()`?

DaanDeMeyer: For directory, of course the name `can_compress()` does make sense. So maybe we need two separate methods.

keszybz: I'm not sure if "directory" is "self decompressing". I changed the name to `use_outer_compression`.

DaanDeMeyer (approved):

Issue #2068, opened 2023-11-14
Autologin doesn't support serial-getty@ttyAMA0.service


jgouly:
arm64 kernels/hw (usually?) use the PL011 driver, which exposes serial as `ttyAMA0`.

However `Autologin=true` doesn't create a drop-in for `serial-getty@ttyAMA0.service` like it does for `serial-getty@ttyS0.service`.

PR #2073, opened 2023-11-16
Ttyama0

behrmann:
@jgouly Can you check whether this addresses your issue?

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -286,6 +286,21 @@ def configure_autologin(state: MkosiState) -> None:
                 """  # noqa: E501
             )
 
+        if state.config.architecture == Architecture.arm64:
+            dropin = state.root / "usr/lib/systemd/system/serial-getty@ttyAMA0.service.d/autologin.conf"
+            with umask(~0o755):
+                dropin.parent.mkdir(parents=True, exist_ok=True)
+            with umask(~0o644):
+                dropin.write_text(
+                    """\
+                    [Service]
+                    ExecStart=
+                    ExecStart=-/sbin/agetty -o '-f -p -- \\\\u' --autologin root --keep-baud 115200,57600,38400,9600 - $TERM
+                    StandardInput=tty
+                    StandardOutput=tty
+                    """  # noqa: E501
+                )
+
 
 @contextlib.contextmanager
 def mount_cache_overlay(state: MkosiState) -> Iterator[None]:
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -61,10 +61,28 @@ class Verb(StrEnum):
     burn          = enum.auto()
 
     def supports_cmdline(self) -> bool:
-        return self in (Verb.build, Verb.shell, Verb.boot, Verb.qemu, Verb.ssh, Verb.journalctl, Verb.coredumpctl, Verb.burn)
+        return self in (
+            Verb.build,
+            Verb.shell,
+            Verb.boot,
+            Verb.qemu,
+            Verb.ssh,
+            Verb.journalctl,
+            Verb.coredumpctl,
+            Verb.burn,
+        )
 
     def needs_build(self) -> bool:
-        return self in (Verb.build, Verb.shell, Verb.boot, Verb.qemu, Verb.serve, Verb.journalctl, Verb.coredumpctl, Verb.burn)
+        return self in (
+            Verb.build,
+            Verb.shell,
+            Verb.boot,
+            Verb.qemu,
+            Verb.serve,
+            Verb.journalctl,
+            Verb.coredumpctl,
+            Verb.burn,
+        )
 
     def needs_root(self) -> bool:
         return self in (Verb.shell, Verb.boot, Verb.burn)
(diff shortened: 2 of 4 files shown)


mkosi/__init__.py
```diff
@@ -286,6 +286,21 @@ def configure_autologin(state: MkosiState) -> None:
                 """  # noqa: E501
             )
 
+        if state.config.architecture == Architecture.arm64:
+            dropin = state.root / "usr/lib/systemd/system/serial-getty@ttyAMA0.service.d/autologin.conf"
+            with umask(~0o755):
+                dropin.parent.mkdir(parents=True, exist_ok=True)
+            with umask(~0o644):
+                dropin.write_text(
```
keszybz, line 294: I think this also needs a `dedent`. Oh, so does the code above.

Those ~~two~~ three copies need to be unified.

mkosi/config.py
```diff
@@ -2575,8 +2579,14 @@ def finalize_defaults(namespace: argparse.Namespace, defaults: argparse.Namespac
...
-        die(f"{setting.long}={getattr(cli_ns, s)} was specified on the command line but is not allowed to be configured by any images.",
-            hint="Prefix the setting with '@' in the image configuration file to allow overriding it from the command line.")
+        die(
+            f"{setting.long}={getattr(cli_ns, s)} was specified on the command line "
+            "but is not allowed to be configured by any images.",
+            hint=(
+                "Prefix the setting with '@' in the image configuration file "
+                "to allow overriding it from the command line."
+            )
+        )
```
keszybz, line 2589: No, please don't break messages like this arbitrarily. This makes grepping and udnerstanding what is going on harder.

mkosi/config.py
```diff
@@ -61,10 +61,14 @@ class Verb(StrEnum):
...
-        return self in (Verb.build, Verb.shell, Verb.boot, Verb.qemu, Verb.ssh, Verb.journalctl, Verb.coredumpctl, Verb.burn)
+        return self in (
+            Verb.build, Verb.shell, Verb.boot, Verb.qemu, Verb.ssh, Verb.journalctl, Verb.coredumpctl, Verb.burn
+        )
 
     def needs_build(self) -> bool:
-        return self in (Verb.build, Verb.shell, Verb.boot, Verb.qemu, Verb.serve, Verb.journalctl, Verb.coredumpctl, Verb.burn)
+        return self in (
+            Verb.build, Verb.shell, Verb.boot, Verb.qemu, Verb.serve, Verb.journalctl, Verb.coredumpctl, Verb.burn
+        )
```
keszybz, line 71: With one-per-line this would be more readable…

keszybz:
I forgot to add: please add a short description to the patch. The text in #2068 is all that is needed.

mkosi/config.py
behrmann: Yeah, I'm not entirely happy about this break, but to me those were already the closest semantic units. Without those breaks these lines are just way too long, though. 

I just pushed a differently terrible version. PTAL.

mkosi/__init__.py
behrmann: > I think this also needs a dedent. Oh, so does the code above.

I can add that. I think I mentioned it, when the code got refactored the last time, but it was declined because the unit file doesn't care about initial whitespace

> Those ~~two~~ three copies need to be unified.

I think it's only two, because the topmost has three `--keep-baud` settings while the bottom two have four. I have really no clue how finicky real-world ttys are in this regard and only copied the bottom one. For this some feedback from  @jgouly would be needed, since I have no hardware to test this presently.

mkosi/config.py
```diff
@@ -2573,10 +2577,13 @@ def finalize_defaults(namespace: argparse.Namespace, defaults: argparse.Namespac
...
-
-        die(f"{setting.long}={getattr(cli_ns, s)} was specified on the command line but is not allowed to be configured by any images.",
-            hint="Prefix the setting with '@' in the image configuration file to allow overriding it from the command line.")
+        setting = SETTINGS_LOOKUP_BY_DEST[s].long
+        a = getattr(cli_ns, s)
+        die(
+            f"{setting}={a} was specified on the command line but is not allowed to be configured by any images.",
+            hint=
+            "Prefix the setting with '@' in the image configuration file to allow overriding it from the command line."
+        )
```
keszybz, line 2586: Sorry, but that's just wrong. It just makes the code harder to read. If the tool is broken, then the solution is to not use the tool.

keszybz (changes requested):

mkosi/config.py
behrmann: Well, wrong on which level? It is syntactically valid... :) The question is not a tool per se, but this line being too long with any tool.

But let's split the difference and just silence the warning here.

mkosi/config.py
behrmann: Playing around with `ruff format` I saw that even with a 119 column limit it would format this with one item per line, so I switched to that.

jgouly:
I didn't try the actual pull request, but I had a similar local modification and that works.

keszybz (approved):

Issue #2072, opened 2023-11-16
CompressOutput with zstd not working


kugel-:
Hello,

the man page tells

```
       CompressOutput=, --compress-output=
              Configure compression for the resulting image or archive.  The argument can be either a boolean or a compression algorithm (xz, zstd).  xz compression is used by default.  Note that when applied to block device  image
              types this means the image cannot be started directly but needs to be decompressed first.  This also means that the shell, boot, qemu verbs are not available when this option is used.  Implied for tar and cpio.
```

but with this in mkosi.conf the tool fails
```
[Output]
Format=disk
Output=vpnsrv.img
CompressOutput=zstd
```

Error: `‣ Invalid boolean literal: 'zstd'`

System Information
```
kugel@thomas-nb:vpnsrv$ mkosi --version
mkosi 18
kugel@thomas-nb:vpnsrv$ cat /etc/os-release 
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://bugs.archlinux.org/"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo
```

PR #2074, opened 2023-11-16
Rename Compression.zst to Compression.zstd

behrmann:
But also leave a fallback Compression.zst with value "zstd" as a fallback in case anybody used that. To test this we also change the construction of Compression instances in test_config.py to use __get_item__ instead of __new__, since that actually supports construction via the value and is also the one actually used in config.py

Fixes: #2072

### DIFF
--- a/mkosi/resources/mkosi.md
+++ b/mkosi/resources/mkosi.md
@@ -607,13 +607,14 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 
 `CompressOutput=`, `--compress-output=`
 
-: Configure compression for the resulting image or archive. The
-  argument can be either a boolean or a compression algorithm (`xz`,
-  `zstd`). `xz` compression is used by default. Note that when applied
-  to block device image types this means the image cannot be started
-  directly but needs to be decompressed first. This also means that
-  the `shell`, `boot`, `qemu` verbs are not available when this option
-  is used. Implied for `tar` and `cpio`.
+: Configure compression for the resulting image or archive. The argument can be
+  either a boolean or a compression algorithm (`xz`, `zstd`). `zstd`
+  compression is used by default, except CentOS and derivatives up to version
+  8, which default to `xz`. Note that when applied to block device image types,
+  compression means the image cannot be started directly but needs to be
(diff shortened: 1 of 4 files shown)


behrmann:
The reason for the rename is that our man pages always speaks of `zstd` but never of `zst`.

keszybz:
`No provider of 'filesystem' found.` — I think we may need to switch opensuse to dnf5 again ;)

mkosi/resources/mkosi.md
```diff
@@ -607,13 +607,14 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
-  `zstd`). `xz` compression is used by default. Note that when applied
-  to block device image types this means the image cannot be started
-  directly but needs to be decompressed first. This also means that
-  the `shell`, `boot`, `qemu` verbs are not available when this option
-  is used. Implied for `tar` and `cpio`.
+: Configure compression for the resulting image or archive. The argument can be
+  either a boolean or a compression algorithm (`xz`, `zstd`). `zstd`
+  compression is used by default, except CentOs and derivatives up to version
+  8, which default to `xz`.  Note that when applied to block device image types
+  this means the image cannot be started directly but needs to be decompressed
```
keszybz, line 614: "CentOS"

"This" is ambiguous here. It sounds like it refers to the compression type on CentOS and elsewhere, which was the subject of the preceding sentence. Maybe change "this" to "compression"?

Oxford comma before "and" below?

keszybz (changes requested):

bluca:
> `No provider of 'filesystem' found.` — I think we may need to switch opensuse to dnf5 again ;)

Wouldn't make any difference, the issue is that both fedora and suse this week switched to zstd from gzip for the repodata, and neither dnf nor zypper could handle it. Enable jammy-proposed in the CI and it will work again.

mkosi/resources/mkosi.md
behrmann: Changed accordingly, I think. PTAL.

septatrix:
I love it when I stumble upon an issue and go to report it, only to see that someone else was quicker and already resolved it :D (Though sadly there have now been several instances where the docs are not in sync with the actual code :( )

PR #2078, opened 2023-11-18
kmod: remove doubled root prefix

behrmann:
@winterhuman Can you check whether this solves the issue?


behrmann:
Doesn't make a difference for the issue, since if `modulesd` is absolute it clobbers `root /` anyway.

Winterhuman:
Yeah, it didn't change anything unfortunately

Issue #2081, opened 2023-11-19
unit tests fail during package builds due to uid checks


bluca:
```
==================================== ERRORS ====================================
____________________ ERROR collecting tests/test_config.py _____________________
/usr/lib/python3/dist-packages/_pytest/runner.py:341: in from_call
    result: Optional[TResult] = func()
/usr/lib/python3/dist-packages/_pytest/runner.py:372: in <lambda>
    call = CallInfo.from_call(lambda: list(collector.collect()), "collect")
/usr/lib/python3/dist-packages/_pytest/python.py:531: in collect
    self._inject_setup_module_fixture()
/usr/lib/python3/dist-packages/_pytest/python.py:545: in _inject_setup_module_fixture
    self.obj, ("setUpModule", "setup_module")
/usr/lib/python3/dist-packages/_pytest/python.py:310: in obj
    self._obj = obj = self._getobj()
/usr/lib/python3/dist-packages/_pytest/python.py:528: in _getobj
    return self._importtestmodule()
/usr/lib/python3/dist-packages/_pytest/python.py:617: in _importtestmodule
    mod = import_path(self.path, mode=importmode, root=self.config.rootpath)
/usr/lib/python3/dist-packages/_pytest/pathlib.py:567: in import_path
    importlib.import_module(module_name)
/usr/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1204: in _gcd_import
    ???
<frozen importlib._bootstrap>:1176: in _find_and_load
    ???
<frozen importlib._bootstrap>:1147: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:690: in _load_unlocked
    ???
/usr/lib/python3/dist-packages/_pytest/assertion/rewrite.py:186: in exec_module
    exec(co, module.__dict__)
tests/test_config.py:13: in <module>
    from mkosi.architecture import Architecture
mkosi/__init__.py:24: in <module>
    from mkosi.architecture import Architecture
mkosi/architecture.py:8: in <module>
    from mkosi.util import StrEnum
mkosi/util.py:77: in <module>
    class INVOKING_USER:
mkosi/util.py:80: in INVOKING_USER
    name = pwd.getpwuid(uid).pw_name
E   KeyError: 'getpwuid(): uid not found: 104'
_____________________ ERROR collecting tests/test_json.py ______________________
[...]

PR #2084, opened 2023-11-21
util: add a fallback username and home for INVOKING_USER

behrmann:
At build time, e.g. during reproducible builds, usernames are not guaranteed to be resolvable. Have a fallback so that our tests can be run in the scenario.

Fixes: #2081

### DIFF
--- a/mkosi/__main__.py
+++ b/mkosi/__main__.py
@@ -13,6 +13,7 @@ from mkosi import run_verb
 from mkosi.config import parse_config
 from mkosi.log import ARG_DEBUG, log_setup
 from mkosi.run import ensure_exc_info, run
+from mkosi.util import INVOKING_USER
 
 
 @contextlib.contextmanager
@@ -43,6 +44,8 @@ def propagate_failed_return() -> Iterator[None]:
 @propagate_failed_return()
 def main() -> None:
     log_setup()
+    # Ensure that the name and home of the user we are running as are resolved as early as possible.
+    INVOKING_USER.init()
     args, images = parse_config(sys.argv[1:])
 
     if args.debug:
--- a/mkosi/util.py
+++ b/mkosi/util.py
@@ -77,13 +77,27 @@ def flatten(lists: Iterable[Iterable[T]]) -> list[T]:
 class INVOKING_USER:
     uid = int(os.getenv("SUDO_UID") or os.getenv("PKEXEC_UID") or os.getuid())
     gid = int(os.getenv("SUDO_GID") or os.getgid())
-    name = pwd.getpwuid(uid).pw_name
-    home = Path(f"~{name}").expanduser()
     invoked_as_root = (uid == 0)
+
+    @classmethod
+    def init(cls) -> None:
+        name = cls.name()
+        home = cls.home()
+        logging.debug(f"Running as user '{name}' ({cls.uid}:{cls.gid}) with home {home}.")
 
     @classmethod
     def is_running_user(cls) -> bool:
         return cls.uid == os.getuid()
+
+    @classmethod
+    @functools.lru_cache(maxsize=1)
+    def name(cls) -> str:
+        return pwd.getpwuid(cls.uid).pw_name
+
+    @classmethod
+    @functools.lru_cache(maxsize=1)
+    def home(cls) -> Path:
+        return Path(f"~{cls.name()}").expanduser()
 
 
 @contextlib.contextmanager
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2604,7 +2604,7 @@ def show_docs(args: MkosiArgs) -> None:
 
 
 def expand_specifier(s: str) -> str:
-    return s.replace("%u", INVOKING_USER.name)
+    return s.replace("%u", INVOKING_USER.name())
 
 
 def needs_build(args: MkosiArgs, config: MkosiConfig) -> bool:
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -219,7 +219,7 @@ def parse_path(value: str,
 
     if expanduser:
         if path.is_relative_to("~") and not INVOKING_USER.is_running_user():
-            path = INVOKING_USER.home / path.relative_to("~")
+            path = INVOKING_USER.home() / path.relative_to("~")
         path = path.expanduser()
 
     if required and not path.exists():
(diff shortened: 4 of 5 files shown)


behrmann:
@bluca Can you have a look whether this works for you? The fallback home will most probably not be writable and the username will definitely not be resolvable back to a UID, but they will be stable.

bluca (approved):
I have no simple and easy way to test this, as it doesn't reproduce on the CI, so let's just go for it and see after the next release

mkosi/util.py
```diff
@@ -77,8 +77,13 @@ def flatten(lists: Iterable[Iterable[T]]) -> list[T]:
...
     gid = int(os.getenv("SUDO_GID") or os.getgid())
-    name = pwd.getpwuid(uid).pw_name
-    home = Path(f"~{name}").expanduser()
+    try:
+        name = pwd.getpwuid(uid).pw_name
+        home = Path(f"~{name}").expanduser()
+    except KeyError:
+        # names for UIDs might not be always resolvable, e.g. during reproducible builds
+        name = "fallback"
+        home = Path("/")
```
septatrix, line 86: How about adding more fallbacks before we have to resort to something hardcoded?
In the linked build log `SUDO_USER=jenkins` is shown to be available. In my local shell also `USER=`, `USERNAME=` and `LOGNAME` are available and seem like reasonable variables to check. Also checking `HOME=` would make sense. Home and logname are also part of POSIX.

Apart from that I think that this should show a warning at least. My worry is people with weird combinations running mkosi and not knowing that they triggered this fallback (e.g. some weird combination of systemd-homed or LDAP (which will not show up in the password database) and privilege escalations like doas which do not variables containing the original user.

behrmann: The `KeyError` will be raised if the numeric UID cannot be resolved to a username. As long as people have some form of working NSS that should work. That not being available should hopefully be a very niche problem, otherwise people will already have a problem. Adding `SUDO_USER`, `USER`, `USERNAME` and `LOGNAME`, would introduce edge cases where we might not be entirely sure anymore what we're getting, so I would prefer to have a clear failure mode. If this is triggered too easily, then we should revisit this.

Does this make sense?

septatrix: > Adding `SUDO_USER`, `USER`, `USERNAME` and `LOGNAME`, would introduce edge cases where we might not be entirely sure anymore what we're getting...

`SUDO_UID` and `PKEXEC_UID` are already used an I fail to see how they are different than `SUDO_USER` or `LOGNAME`. Also `PKEXEC_UID` has no gid equivalent so inconsistent states are already possible.

> ...so I would prefer to have a clear failure mode. If this is triggered too easily, then we should revisit this.

This is why I would definitely emit a warning whenever the fallback is triggered.

bluca: Note that this is only a problem for the unit tests, so having a fallback that is only enabled when being run from a unit test would be perfectly fine for me

behrmann: I find the idea appealing to limit this just to tests, but since this explodes during import time this can't be mocked out easily. Is setting an environment variable for the test an option? In that case the try-except-block could just be a check whether the magic variable is set. This would make the whole behaviour opt in and sidestep all questions of people triggering this by accident.

bluca: Yeah sure, that's not an issue

behrmann: Reworked like that. This also has the added benefit of making this behaviour easily testable.

mkosi/util.py
```diff
@@ -77,8 +77,13 @@ def flatten(lists: Iterable[Iterable[T]]) -> list[T]:
 class INVOKING_USER:
     uid = int(os.getenv("SUDO_UID") or os.getenv("PKEXEC_UID") or os.getuid())
     gid = int(os.getenv("SUDO_GID") or os.getgid())
-    name = pwd.getpwuid(uid).pw_name
-    home = Path(f"~{name}").expanduser()
+    # names for UIDs might not be always resolvable, e.g. during reproducible builds
```
keszybz, line 80: Hmm, I have to say that the comment doesn't make sense to me. Why would it matter whether a build is reproducible or not matter for name resolution? Do you maybe mean that a repeat build that does not set up the environment properly hits this case?

Isn't https://github.com/systemd/mkosi/issues/2081 really just a problem with how the environment is set up? The uid we are running under should have a name. If it doesn't many things will complain. If somebody is trying to repeat a build, they must take care of /etc/passwd and /etc/group too.

behrmann: This is probably too short, yes, but I'd like to keep the context in there. How about:
> names for UIDs might not be always resolvable, e.g. during build time without proper NSS, which is used during reproducible builds so as not to leak the username of the build user

keszybz:
I think that this is just papering over a bug in environment setup somewhere.
I see two possibilities:
- just treat it as an error. I think that's the most reasonable choice. We don't expect to be called by a user without a name.
- do something like shells do, and provide a numerical user name.

Various tools take the first option, for example sudo refuses to be called by an unknown user.

When I log in as a user who has been deleted, bash sets a numeric value for $USER, and does not set $USERNAME and $HOME. The shell needs to be more permissive to allow recovery.

The question is whether we want to be like the shell… If we want to, then I think we should use a numeric uid for the user name. But we also have support for loading paths relative to `INVOKING_USER.home`. So this will necesarrily break. So I think we would need to keep `INVOKING_USER.home==None` and raise an error if we are asked to expand such a path.

keszybz:
OK, but if we are fine with running as a user that doesn't exist, then we should treat this systematically and not load things from unexpected paths. I.e. I don't think we should look for `/.config/` and stuff.

bluca:
I don't mind either way - this is only needed for unit tests, which need to be runnable in weird and extremely limited environments. For real actual runs you can error out when the env is not as expected, that's fine. But unit tests shouldn't require very specific and limited environments that can only be expected for user runs or dedicated setups like that.

bluca (approved):

behrmann:
I beefed up the comment as written in https://github.com/systemd/mkosi/pull/2084#discussion_r1405970619. I don't think we want to support running in such broken build environments. The way this currently is, this is opt in and will probably only work for running tests, which I think is fine. For that scenario is has a somewhat identifiable username and both the username and home always come out the same. Trying to build something like this will surely break, but I think that's ok and not something we should support unless somebody asks for that and there's a compelling reason.

keszybz:
OK, please merge it you think it's the right way to go.

I think that the way that the build environment is set up is misguided… Instead of ripping out parts of the environment and hoping that somehow things will still work, the environment should be standarized. The Debian way is creating work for people doing the builds, and, as we see here, for project maintainers. But meh, as you say, this is opt-in.

DaanDeMeyer (changes requested):
I don't think this is the way to go. We should use None instead of "fallback" and conditionalize logic depending on whether the name is available or not. Probably same for the user's home dir.

mkosi/__main__.py
```diff
@@ -43,6 +44,9 @@ def propagate_failed_return() -> Iterator[None]:
 @propagate_failed_return()
 def main() -> None:
     log_setup()
+    # Ensure that the name and home of the user we are running as are resolved as early as possible.
+    INVOKING_USER.name()
+    INVOKING_USER.home()
```
DaanDeMeyer, line 49: Can we put these in an init() method or so and call that instead?

behrmann: Like this?

DaanDeMeyer (approved):

DaanDeMeyer (approved):

Issue #2086, opened 2023-11-22
Unknown setting SignExpectedPCR


NekkoDroid:
When running mkosi with the following config (and also a different config, but this is the minimal I was able to boil it down to) failes due to `Unknown setting SignExpectedPCR`:

```
[Distribution]
Distribution=arch

[Content]
Packages=
	linux
	systemd

[Validation]
SignExpectedPCR=true
```

I have tried this with mkosi 19 (Debian trixie in WSL) and also the latest git version (Arch system), both of them failed (running systemd 254.5 and 254.6 respectivly).

NekkoDroid:
This seems to be solved by using `SignExpectedPcr` (case sensitive).

This seems to be documented wrongly then since in https://github.com/systemd/mkosi/blob/main/mkosi/resources/mkosi.md?plain=1#L1144 this is documented with a capital PCR.

PR #2087, opened 2023-11-22
Fixed capitalization of `SignExpectedPcr`

NekkoDroid:
Fixes: #2086

### DIFF
--- a/mkosi/resources/mkosi.md
+++ b/mkosi/resources/mkosi.md
@@ -1141,7 +1141,7 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 : Path to the X.509 file containing the certificate for signing the verity signature, if a verity signature
   partition is added with systemd-repart.
 
-`SignExpectedPCR=`, `--sign-expected-pcr`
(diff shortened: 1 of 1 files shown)


behrmann:
That's one way of fixing that, but I think it might be better to
```diff
     MkosiConfigSetting(
         dest="sign_expected_pcr",
         metavar="FEATURE",
+        name="SignExpectedPCR",
         section="Validation",
         parse=config_parse_feature,
     )
```
which would leave the user-facing side untouched. Can you give that a test and see whether that change would work for you?

NekkoDroid:
This does work.

The reason I decided to change the documentation was mostly incase someone actually already figure this out and decided to not report it, to not break their config.

But if this is the preferred way I can gladly undo the commit and change the name instead.

NekkoDroid:
I just did a quick search on github and I looks like systemd itself (and some forks) uses the lower case variants: https://github.com/systemd/systemd/blob/main/mkosi.presets/system/mkosi.conf#L48

behrmann:
Then let's stay with that.

behrmann (approved):

PR #2093, opened 2023-11-26
Add ability to create passwordless root account

### LINKED ISSUE #2091: Allow setting `RootPassword` to empty string
Due to some checks for truthiness instead of explicitly checking against None it is not possible to assign the root password. This means that it is not possible to create a passwordless, unlocked root account. While I do not endorse open root accounts there are certainly use cases for them and we should not stand in the way of creating them. The attached patch adds the ability to add such accounts using `hashed:` (without an actual hash following it). Using `RootPassword=hashed:` instead of `RootPassword=` allows to preserve the intended behaviour that assigning the empty value resets previous assignments (and supporting the latter would also have complicated the configparser a decent bit.

```patch
From 0b46aca23e67c41618aee931fa6910b32a42111e Mon Sep 17 00:00:00 2001
From: Septatrix <24257556+Septatrix@users.noreply.github.com>
Date: Sat, 25 Nov 2023 23:14:08 +0100
Subject: [PATCH] Add ability to create passwordless root account

To create a passwordless root account one may specify `hashed:`
without passing an actual hash.

---
 mkosi/__init__.py        | 7 ++++---
 mkosi/resources/mkosi.md | 3 ++-
 2 files changed, 6 insertions(+), 4 deletions(-)

diff --git a/mkosi/__init__.py b/mkosi/__init__.py
index f0fe850..fe8e8de 100644
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1117,8 +1117,8 @@ def build_initrd(state: MkosiState) -> Path:
     # Default values are assigned via the parser so we go via the argument parser to construct
     # the config for the initrd.
 
-    password, hashed = state.config.root_password or (None, False)
-    if password:
+    if state.config.root_password:
+        password, hashed = state.config.root_password
         rootpwopt = f"hashed:{password}" if hashed else password
     else:
         rootpwopt = None
@@ -1845,7 +1845,8 @@ def run_firstboot(state: MkosiState) -> None:
     creds = []
 
     for option, cred, value in settings:
-        if not value:
+        # Check for None as password might be the empty string
+        if value is None:
             continue
 
         options += [option, value]
diff --git a/mkosi/resources/mkosi.md b/mkosi/resources/mkosi.md
index fcb043a..ce05b94 100644
--- a/mkosi/resources/mkosi.md
+++ b/mkosi/resources/mkosi.md
@@ -1073,7 +1073,8 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 : Set the system root password. If this option is not used, but a `mkosi.rootpw` file is found in the local
   directory, the password is automatically read from it. If the password starts with `hashed:`, it is treated
   as an already hashed root password. The root password is also stored in `/usr/lib/credstore` under the
-  appropriate systemd credential so that it applies even if only `/usr` is shipped in the image.
+  appropriate systemd credential so that it applies even if only `/usr` is shipped in the image. The create
+  an unlocked account without any password use `hashed:` without a hash.
 
 `Autologin=`, `--autologin`
 
-- 
2.42.0
 ```

### PR DESCRIPTION
**Pull request by septatrix**
This PR adds the ability to use a passwordless root accounts using `hashed:` (without an actual hash following it). Using `RootPassword=hashed:` instead of `RootPassword=` allows to preserve the intended behaviour that assigning the empty value resets previous assignments (and supporting the latter would also have complicated the configparser a decent bit.

Fixes #2091

### DIFF
--- a/mkosi/resources/mkosi.md
+++ b/mkosi/resources/mkosi.md
@@ -1073,7 +1073,7 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 : Set the system root password. If this option is not used, but a `mkosi.rootpw` file is found in the local
   directory, the password is automatically read from it. If the password starts with `hashed:`, it is treated
   as an already hashed root password. The root password is also stored in `/usr/lib/credstore` under the
-  appropriate systemd credential so that it applies even if only `/usr` is shipped in the image. The create
+  appropriate systemd credential so that it applies even if only `/usr` is shipped in the image. To create
   an unlocked account without any password use `hashed:` without a hash.
 
 `Autologin=`, `--autologin`
--- a/mkosi/resources/mkosi.md
+++ b/mkosi/resources/mkosi.md
@@ -1073,7 +1073,8 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 : Set the system root password. If this option is not used, but a `mkosi.rootpw` file is found in the local
   directory, the password is automatically read from it. If the password starts with `hashed:`, it is treated
   as an already hashed root password. The root password is also stored in `/usr/lib/credstore` under the
-  appropriate systemd credential so that it applies even if only `/usr` is shipped in the image.
+  appropriate systemd credential so that it applies even if only `/usr` is shipped in the image. The create
+  an unlocked account without any password use `hashed:` without a hash.
 
 `Autologin=`, `--autologin`
 
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1117,8 +1117,8 @@ def build_initrd(state: MkosiState) -> Path:
     # Default values are assigned via the parser so we go via the argument parser to construct
     # the config for the initrd.
 
-    password, hashed = state.config.root_password or (None, False)
-    if password:
+    if state.config.root_password:
+        password, hashed = state.config.root_password
         rootpwopt = f"hashed:{password}" if hashed else password
     else:
         rootpwopt = None
@@ -1845,7 +1845,8 @@ def run_firstboot(state: MkosiState) -> None:
     creds = []
 
     for option, cred, value in settings:
-        if not value:
+        # Check for None as password might be the empty string
+        if value is None:
             continue
 
         options += [option, value]

### CODE REVIEW

**Comment by septatrix** [comment]
I just noticed this would also eliminate the need for the current hack in systemd which is to set the credential manually instead of going via `RootPassword=`: https://github.com/systemd/systemd/blob/54f3b3d2a4aed7867f2c380549f7d66773387754/mkosi.conf.d/10-systemd.conf#L22

File: mkosi/resources/mkosi.md [hunk 1]
```diff
@@ -1073,7 +1073,8 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 : Set the system root password. If this option is not used, but a `mkosi.rootpw` file is found in the local
   directory, the password is automatically read from it. If the password starts with `hashed:`, it is treated
   as an already hashed root password. The root password is also stored in `/usr/lib/credstore` under the
-  appropriate systemd credential so that it applies even if only `/usr` is shipped in the image.
+  appropriate systemd credential so that it applies even if only `/usr` is shipped in the image. The create
```
keszybz, line 1076: "The create" → "To create"

keszybz:
I think this makes sense.

keszybz (changes requested):

mkosi/resources/mkosi.md
septatrix:
```suggestion
  appropriate systemd credential so that it applies even if only `/usr` is shipped in the image. To create
```

keszybz:
You are fixing an error introduced in the first commit. This shouldn't be a separate commit, but the first one should be amended. This is mostly a note for the future, we can squash it in the future.

Please don't set "need review" unless there's a good reason. I'm subscribed to all messages for this repo. (And the same applies to all other repo owners… There is no need to do that.)

keszybz:
> You are fixing an error introduced in the first commit. This shouldn't be a separate commit, but the first one should be amended. This is mostly a note for the future, we can squash it in this case.
> 
> Please don't set "need review" unless there's a good reason. I'm subscribed to all messages for this repo. (And the same applies to all other repo owners… There is no need to do that.)

septatrix:
> You are fixing an error introduced in the first commit. This shouldn't be a separate commit, but the first one should be amended. This is mostly a note for the future, we can squash it in the future.

I personally try to avoid force pushes whenever possible as I seem to remember that previously GitHub had some issues when handling/showing reviews from earlier commits after a force push - maybe this has improved since then. Especially when it is only a single commit I try to leave this choice to the maintainers which as you mentioned can squash upon merging.

Issue #2092, opened 2023-11-25
`KernelModulesInclude` doesn't handle nested dependencies


Winterhuman:
Host OS: Arch Linux
`uname`: Linux 6.6.2-arch1-1 #1 SMP PREEMPT_DYNAMIC x86_64 GNU/Linux
`mkosi`: v19 (git main)

# Description

When doing `KernelModulesInclude=drivers/nvme/host/nvme.ko.zst`, the `nvme-core` module gets pulled in correctly, however, `nvme-core`'s dependency, `nvme-common`, is missed out.

# Debug log

[debug.txt](https://github.com/systemd/mkosi/files/13465427/debug.txt)

# Module list from config

```
	arch/x86/crypto/aesni-intel.ko.zst
	arch/x86/crypto/crc32c-intel.ko.zst
	arch/x86/crypto/crc32-pclmul.ko.zst
	arch/x86/crypto/crct10dif-pclmul.ko.zst
	arch/x86/crypto/ghash-clmulni-intel.ko.zst
	arch/x86/crypto/polyval-clmulni.ko.zst
	arch/x86/crypto/sha256-ssse3.ko.zst
	arch/x86/crypto/sha512-ssse3.ko.zst
	arch/x86/events/rapl.ko.zst
	crypto/algif_hash.ko.zst
	crypto/algif_skcipher.ko.zst
	crypto/blake2b_generic.ko.zst
	crypto/cmac.ko.zst
	crypto/crc32c_generic.ko.zst
	crypto/crypto_user.ko.zst
	drivers/block/loop.ko.zst
	drivers/cpufreq/acpi-cpufreq.ko.zst
	drivers/edac/edac_mce_amd.ko.zst
	drivers/gpio/gpio-amdpt.ko.zst
	drivers/hid/usbhid/usbhid.ko.zst
	drivers/hwmon/k10temp.ko.zst
	drivers/i2c/busses/i2c-piix4.ko.zst
	drivers/i2c/i2c-dev.ko.zst
	drivers/input/joydev.ko.zst
	drivers/input/mousedev.ko.zst
	drivers/md/dm-mod.ko.zst
	drivers/net/ethernet/realtek/r8169.ko.zst
	drivers/net/phy/realtek.ko.zst
	drivers/nvme/host/nvme.ko.zst
#	drivers/nvme/common/nvme-common.ko.zst	# This is the missing module.
	drivers/platform/x86/wmi-bmof.ko.zst
[...]

PR #2094, opened 2023-11-26
kmod: normalise module names

behrmann:
The file names of modules use dashes, e.g.

    /usr/lib/modules/6.6.2-arch1-1/kernel/drivers/nvme/common/nvme-common.ko.zst

but the dependency names modinfo gives us use underscores, e.g.

    name=nvme_common

Fixes: #2092


behrmann:
@winterhuman Can you see whether this fixes your issue?

Winterhuman (changes requested):
"thefore" -> "therefore"

Winterhuman:
Nope, `nvme-common` still isn't being pulled in. I think the actual problem is that `modinfo` only prints one-level deep dependencies, so any dependencies found have to be ran through as well, like a loop where it breaks when the last two lists of modules are the same meaning all modules have been found.

Winterhuman:
Yep, it worked!

PR #2095, opened 2023-11-27
autologin: use a helper method to reduce boilerplate

keszybz:
I kept the generated option strings exactly the same. In some cases we put the baud before the device name or placeholder, and in other cases in the opposite order, but maybe there's some reason for this.

The unit file is changed to use just 'aggety' instead of the full path, and the leading whitespace is dropped. Both things should have no effect, but make the resulting file nicer to read.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -239,67 +239,39 @@ def configure_os_release(state: MkosiState) -> None:
         newosrelease.rename(osrelease)
 
 
+def configure_autologin_service(state: MkosiState, service: str, extra: str) -> None:
+    dropin = state.root / f"usr/lib/systemd/system/{service}.d/autologin.conf"
+    with umask(~0o755):
+        dropin.parent.mkdir(parents=True, exist_ok=True)
+    with umask(~0o644):
+        dropin.write_text(
+            textwrap.dedent(
+                f"""\
+                [Service]
+                ExecStart=
+                ExecStart=-agetty -o '-f -p -- \\\\u' --autologin root {extra} $TERM
+                StandardInput=tty
+                StandardOutput=tty
+                """
+            )
+        )
+
+
 def configure_autologin(state: MkosiState) -> None:
     if not state.config.autologin:
         return
 
     with complete_step("Setting up autologin…"):
-        dropin = state.root / "usr/lib/systemd/system/console-getty.service.d/autologin.conf"
-        with umask(~0o755):
-            dropin.parent.mkdir(parents=True, exist_ok=True)
-        with umask(~0o644):
-            dropin.write_text(
-                """\
-                [Service]
-                ExecStart=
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -239,67 +239,39 @@ def configure_os_release(state: MkosiState) -> None:
...
+    if not state.config.autologin:
+        return
+
+    with complete_step("Setting up autologin…"):
+        configure_autologin_service(state, "console-getty.service",
+                                    "--noclear --keep-baud console 115200,38400,9600")
+        configure_autologin_service(state, "getty@tty1.service",
+                                    "--noclear -")
+        configure_autologin_service(state, "serial-getty@ttyS0.service",
+                                    "--keep-baud 115200,57600,38400,9600 -")
```
behrmann, line 270: Even the longest line is withing the 119 char limit:
```suggestion
        configure_autologin_service(state, "console-getty.service", "--noclear --keep-baud console 115200,38400,9600")
        configure_autologin_service(state, "getty@tty1.service", "--noclear -")
        configure_autologin_service(state, "serial-getty@ttyS0.service", "--keep-baud 115200,57600,38400,9600 -")
```

mkosi/__init__.py
```diff
@@ -239,67 +239,39 @@ def configure_os_release(state: MkosiState) -> None:
...
-                    """\
-                    [Service]
-                    ExecStart=
-                    ExecStart=-/sbin/agetty -o '-f -p -- \\\\u' --autologin root --keep-baud 115200,57600,38400,9600 - $TERM
-                    StandardInput=tty
-                    StandardOutput=tty
-                    """  # noqa: E501
-                )
+            configure_autologin_service(state, "serial-getty@ttyAMA0.service",
+                                        "--keep-baud 115200,57600,38400,9600 -")
```
behrmann, line 274: same

mkosi/__init__.py
```diff
@@ -239,67 +239,39 @@ def configure_os_release(state: MkosiState) -> None:
...
+        dropin.write_text(
+            textwrap.dedent(
+                f"""\
                 [Service]
                 ExecStart=
-                ExecStart=-/sbin/agetty -o '-f -p -- \\\\u' --autologin root --noclear --keep-baud console 115200,38400,9600 $TERM
+                ExecStart=-agetty -o '-f -p -- \\\\u' --autologin root {extra} $TERM
                 StandardInput=tty
                 StandardOutput=tty
                 """  # noqa: E501
```
behrmann, line 255: That `noqa` shouldn't be necessary any longer, since the `EcecStart` is shorter now.

behrmann (approved):
LGTM, a few nits.

mkosi/__init__.py
keszybz: I dropped `noqa`.

DaanDeMeyer (approved):

PR #2107, opened 2023-11-28
Speed up empty_directory()

DaanDeMeyer:
Let's invoke rm once instead of N times.


behrmann (approved):

septatrix:
I am not sure how well optimized `rm` is (or how "slow" Python is for that matter) but you could eliminate spawning a subprocess completely by using [`shutil.rmtree`](https://docs.python.org/3/library/shutil.html#shutil.rmtree)

DaanDeMeyer:
Python is slow so we prefer subprocceses for stuff that potentially takes a long time.

PR #2108, opened 2023-11-29
mkosi-initrd: Copy md and dm udev rules from dracut

DaanDeMeyer:
Until these are upstreamed into their corresponding projects, let's carry the rules in mkosi-initrd.


DaanDeMeyer:
I'm not sure how to properly attribute these? Do I need to put a GPL v2 notice in there pointing back to dracut somehow?

DaanDeMeyer:
Replaces #2056

bluca:
> I'm not sure how to properly attribute these? Do I need to put a GPL v2 notice in there pointing back to dracut somehow?

For clarity an SPDX line first and a link to the original second won't hurt

behrmann (approved):

keszybz (commented):
FTR, I think that the first of those two files should be replaced by dropping `initrd-udevadm-cleanup-db.service` service, and the second should be moved to `mdadm`.

PR #2110, opened 2023-11-29
lock the correct file for newgidmap

sam-leonard-ct:
Previously /etc/subuid was locked for both newuidmap and newgidmap, this is incorrect as newgidmap edits /etc/subgid.
This change fixes that.


DaanDeMeyer:
It's intended that both lock /etc/subuid. It doesn't matter which file we lock, as long as we try to lock the same file.

PR #2117, opened 2023-12-01
Move integration tests into python

DaanDeMeyer:
Instead of vendor locking ourselves to Github Actions, let's move
the integration tests into python so we can run them locally and
on other CI systems.

We opt to use unittest style test cases so that we can have a
configurable base class that can be used for various integration
tests. Unfortunately, I haven't found a nice and type safe way to
make pytest fixtures configurable so we opt for unittest instead.

Note that while we use the subTest() feature of unittest, pytest
still considers test_boot() a single test because it doesn't support
this particular feature of unittest. Ideally we switch our test runner
to something else in the future which does support the subTest() feature.

We always run steps that can run unprivileged without privileges even
if we're running as root so that tests can be run locally with root
privileges without ending up with a bunch of files owned by root
afterwards.


behrmann:
I'm not a fan of unittest and will have a look at porting this to pytest next week, but apart from that this looks sensible and LGTM for now.

behrmann (approved):

tests/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+from mkosi.types import CompletedProcess
+from mkosi.util import INVOKING_USER
+
+
+class Image:
+    def __init__(self, options: Sequence[str] = []) -> None:
+        self.options = options
+
+        if d := os.getenv("MKOSI_TEST_DISTRIBUTION"):
+            self.distribution = Distribution[d]
```
behrmann, line 21: How about passing the distribution and release as arguments here, this would allow to mark them, e.g. `pytest.mark.arch` and `pytest.mark.fedora` and then the tests can be parametrized over this and called via the marks.

DaanDeMeyer: I'm not sure I understand. Maybe do a follow up PR after this one is merged?

behrmann: It's fine it can wait.

PR #2122, opened 2023-12-05
docs: Add doc on generating system extensions with mkosi

daandemeyer:


bluca:
Would be good to also link the spec https://uapi-group.org/specifications/specs/extension_image/

bluca (approved):

docs/sysext.md
```diff
@@ -0,0 +1,129 @@
...
+
+We'll also want to mark our extension as a system extension. To do that,
+write the following to
+`mkosi.images/btrfs/mkosi.extra/usr/lib/extension-release.d/extension-release.btrfs`:
+
+```conf
+ID=<distribution>
+VERSION_ID=<distribution-version>
+ARCHITECTURE=<architecture>
+SYSEXT_SCOPE=initrd
```
behrmann, line 66: This is a bit specialised and it makes sense for the example of btrfs, but a short paragraph on this at the beginning would probably be helpful for casual readers, what the example here is, otherwise I think this might be cargo culted.

behrmann (commented):
Great example! One thought, though...

behrmann (approved):

docs/sysext.md
esposem: I think any sysext would need at least ID, VERSION_ID and SYSEXT_SCOPE=initrd if it has to be an initrd extension. ARCHITECTURE can be dropped, and maybe SYSEXT_SCOPE=initrd should be documented (or at least a link to a doc should be there) to show the reader the options available there

docs/sysext.md
```diff
@@ -0,0 +1,131 @@
...
+
+```conf
+[Partition]
+Type=root-verity-sig
+Verity=signature
+VerityMatchKey=root
+```
+
+Of course we can't sign anything without a key, so let's generate one
+with `mkosi genkey`. Note that this key will need to be loaded into your
```
esposem, line 106: Please also add that an existing key that was previously added into MOK/keyring can be added by creating mkosi.crt (certificate) and mkosi.key (private key).

DaanDeMeyer: This has already been added in a follow up PR

PR #2123, opened 2023-12-06
Execute user presets after system presets

Winterhuman:
Adds a `systemctl --root "$ROOT" --global preset-all` line to the **Applying presets...** build step.


DaanDeMeyer (approved):

behrmann (approved):

septatrix:
Happy to see this merged. I actually stumbled upon this a few weeks ago but am using /usr-only images so I filed this as a bug in systemd instead: https://github.com/systemd/systemd/issues/30206

PR #2125, opened 2023-12-06
mkosi-initrd: Apply cryptsetup creds to all instances of systemd-cryp…

daandemeyer:
…tsetup@.service

This follows the suggestion made in https://github.com/systemd/systemd/pull/30309.


behrmann:
> This follows the suggestion made in https://github.com/systemd/systemd/pull/30309.

i.e. following the comment about not being more restrictive than the password caching we already do? I think that comment should go into the commit message, so that it can stand alone. Otherwise LGTM.

PR #2128, opened 2023-12-06
Add sysext, confext and portable support

DaanDeMeyer:
This uses the new --make-ddi= option in systemd-repart added in v255 to add explicit support for sysexts, confexts and portable services.

On top of using the new repart option to generate the extension images, we also write the extension-release file in case of sysexts and confexts and make sure we skip a bunch of our automatic features when building extension images or enabling the Overlay= option as in these cases many of our automatic features are undesireable.

### DIFF
--- a/docs/sysext.md
+++ b/docs/sysext.md
@@ -47,7 +47,7 @@ top of it by writing the following to `mkosi.images/btrfs/mkosi.conf`:
 Dependencies=base
 
 [Output]
-Format=disk
+Format=sysext
 Overlay=yes
 
 [Content]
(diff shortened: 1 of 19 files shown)


docs/sysext.md
```diff
@@ -4,6 +4,13 @@
 images may – dynamically at runtime — extend the base system with an
 overlay containing additional files.
 
+Note that to build system extensions with mkosi, you'll need
+systemd-repart v255 or newer installed on the host system or in the
```
behrmann, line 8: I guess that line is outdated now.

behrmann (approved):
LGTM, one outdated line, though.

DaanDeMeyer:
Just a doc change

PR #2130, opened 2023-12-06
Disable xino feature for overlayfs mounts

DaanDeMeyer:
Doesn't seem to do anything too useful and causes warnings so let's drop it.


DaanDeMeyer:
This doesn't seem to get rid of the warning so closing

PR #2131, opened 2023-12-07
Run image builds in a fork again

DaanDeMeyer:
This solves two problems:
- When not using a tools tree, we can run qemu outside of the user namespace which means that we don't need to pass fds to /dev/kvm and /dev/vhost-vsock to keep things working unprivileged
- The vmspawn verb we're about to introduce will not be able to run properly inside a user namespace, so we need to make sure we're not inside a user namespace after the image build.

Compared to our original implementation of this way back with exception propagation, this time we opt to do things differently by doing all exception handling and logging inside the fork to avoid having to propagate exceptions. This makes the overall implementation a lot simpler.

We can also run the other verbs outside of the user namespace as long as we're not using a tools tree. Because we want to keep support for using a tools tree with all verbs, we keep support for running them inside a user namespace as well. Because we already use INVOKING_USER everywhere, this actually turns out to require very little changes. We only need to make sure when starting virtiofsd that we unshare the user namespace ourselves if we're not uid mapping.


mkosi/run.py
```diff
@@ -161,6 +161,79 @@ def ensure_exc_info() -> tuple[type[BaseException], BaseException, TracebackType
...
+        # the logging for those failures so we don't log stacktraces for those either.
+        if ARG_DEBUG.get() and e.cmd and e.cmd[0] not in ("self", "qemu", "ssh", "systemd-nspawn"):
+            sys.excepthook(*ensure_exc_info())
+
+        # We always log when subprocess.CalledProcessError is raised, so we don't log again here.
+        exit(e.returncode)
+    except BaseException:
+        sys.excepthook(*ensure_exc_info())
+        exit(1)
+    else:
```
behrmann, line 192: Is there a deeper reason this is an `else` instead of ` finally`, i.e. doesn't it make sense to flush the streams in the error case as well?

behrmann (approved):
I think this is ok. I've read it a few times now, mostly the first commit as that's the meat, and as far as I see it it makes sense. It is definitely looks easier to grok than when we had this the last time, but I wouldn't be surprised if I missed some subtlety here.

mkosi/run.py
DaanDeMeyer: So I did it like this because if we call `os._exit()` the finally won't actually run and I didn't really like adding the flushing to every exception block. But we should really always flush, so I reworked this now to always flush.

PR #2134, opened 2023-12-07
Move off of the deprecated importlib.resources API

behrmann:
Python 3.9 brought a new importlib.resources API and deprecated the old one. This introduces a small shim to use the part of the APi that guarantees a Path object, thus making our usage of this just pathlike.


DaanDeMeyer:
I'm pretty sure we can remove the `__init__.py` in the mkosi/resources/repart and mkosi/resources/repart/definitions directories as well, can you add that to this PR?

PR #2136, opened 2023-12-07
qemu: Become root when copying for Ephemeral=

DaanDeMeyer:
To copy the directory, we need to be root (either real or in the user namespace), so let's do just that.


behrmann (approved):

DaanDeMeyer:
Arch CI failure is unrelated

PR #2137, opened 2023-12-08
Add support for installing shim

DaanDeMeyer:
Unfortunately shim is a necessary evil that we have to support. We
add a new option that allows choosing either a signed version, an
unsigned version or none at all (the default).

We also stop redirecting /boot/efi to /efi so that /efi is our pristine
directory for populating the ESP whereas /boot is unclaimed wasteland
free for package managers to write all kinds of stuff to.


mkosi/resources/mkosi.md
```diff
@@ -1021,6 +1021,20 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 : Even if no EFI bootloader is installed, we still need an ESP for BIOS
   boot as that's where we store the kernel, initrd and grub modules.
 
+`ShimBootloader=`, `--shim-bootloader=`
+
+: Takes one of `none`, `unsigned` or `signed`. Defaults to `none`. If
+  set to `none`, shim will not be installed. If set to `unsigned`, mkosi
```
keszybz, line 1027: Oxford comma please.

I think that "installed" should be clarified here. Because there's package installation and the installation from the installed package to the ESP.

mkosi/resources/mkosi.md
```diff
@@ -1021,6 +1021,20 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
+  will search for an unsigned shim EFI binary and install it. If
+  `SecureBoot=` is enabled, mkosi will sign the unsigned shim EFI binary
+  before installing it. If set to `signed`, mkosi will search for a
+  signed EFI binary and install it. If `SecureBoot=` is enabled, mkosi
```
keszybz, line 1031: "If" → "Even if".

mkosi/__init__.py
```diff
@@ -722,24 +717,57 @@ def pesign_prepare(state: MkosiState) -> None:
...
+        return Path(f"efi/EFI/BOOT/grub{arch}.EFI")
+
+
+def sign_efi_binary(state: MkosiState, input: Path, output: Path) -> None:
+    assert state.config.secure_boot_key
+    assert state.config.secure_boot_certificate
+
+    if (state.config.secure_boot_sign_tool == SecureBootSignTool.sbsign or
+        state.config.secure_boot_sign_tool == SecureBootSignTool.auto and
+        shutil.which("sbsign") is not None):
```
behrmann, line 741:
```suggestion
    if (
        state.config.secure_boot_sign_tool == SecureBootSignTool.sbsign or
        state.config.secure_boot_sign_tool == SecureBootSignTool.auto and
        shutil.which("sbsign") is not None
    ):
```

mkosi/__init__.py
```diff
@@ -823,6 +833,62 @@ def install_systemd_boot(state: MkosiState) -> None:
...
+    dst = efi_boot_binary(state)
+    with umask(~0o700):
+        (state.root / dst).parent.mkdir(parents=True, exist_ok=True)
+
+    if state.config.shim_bootloader == ShimBootloader.signed:
+        for pattern in signed:
+            for p in state.root.glob(pattern):
+                log_step(f"Installing presigned shim EFI binary from /{p.relative_to(state.root)} to /{dst}")
+                shutil.copy2(p, state.root / dst)
+                return
```
behrmann, line 872: Is there any possibility that a single glob could match multiple multiple files? Does it make sense to copy only the first one in that case and if so, should this maybe get some sort of version sort?

mkosi/__init__.py
```diff
@@ -823,6 +833,62 @@ def install_systemd_boot(state: MkosiState) -> None:
...
+    if state.config.shim_bootloader == ShimBootloader.none:
+        return
+
+    if not any(gen_kernel_images(state)) and state.config.bootable == ConfigFeature.auto:
+        return
+
+    signed = [
+        "usr/lib/shim/shim*.efi.signed.latest",
+        "usr/lib/shim/shim*.efi.signed",
+        "boot/efi/EFI/*/shim[a-z][0-9]*.efi",
```
behrmann, line 849: I'm not quite sure what this is supposed to match, a comment with examples would probably be helpful saying which cases occur where, since this seems to mix glob and regex semantic, e.g. this would match a file `shima9%%.efi`, because the star doesn't affect the character ranges before it. Is that intended?

DaanDeMeyer: Fixed in the latest push

mkosi/__init__.py
DaanDeMeyer: None of the distribution packages I've looked at ship names with versions in them so I wouldn't bother for now

behrmann (approved):
LGTM, but since this includes shim and mok manager, isn't there also this fallback thing at least on some distros that should maybe be included there?

DaanDeMeyer:
> LGTM, but since this includes shim and mok manager, isn't there also this fallback thing at least on some distros that should maybe be included there?

This would also require figuring out BOOT.CSV as that is parsed by fallback.efi. Let's figure that out later.

PR #2142, opened 2023-12-08
tests: Make more robust on distros with recent systemd

DaanDeMeyer:
Let's make use of the fact that we can communicate the exit status from VMs on recent versions of systemd. Even when it fails to run qemu will often exit with exit status 0 so let's make our successful exit status 123 and check for that instead of 0.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: LGPL-2.1+
 
 import os
+import subprocess
 import sys
 import tempfile
 from collections.abc import Iterator, Sequence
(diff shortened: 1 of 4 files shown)


behrmann (approved):

tests/__init__.py
```diff
@@ -83,18 +95,36 @@ def build(self, options: Sequence[str] = (), args: Sequence[str] = ()) -> Comple
...
             "qemu",
             [*options, "--debug"],
             args,
             stdin=sys.stdin if sys.stdin.isatty() else None,
             user=INVOKING_USER.uid,
             group=INVOKING_USER.gid,
+            check=False,
         )
 
+        rc = 0 if self.distribution == Distribution.ubuntu or self.distribution.is_centos_variant() else 123
```
behrmann, line 121: Thinking some more about this, is this the right way 'round? Qemu runs on the host and not in the image, no?

DaanDeMeyer: Yes? This is about systemd versions, not qemu

PR #2145, opened 2023-12-09
Pass uname to ukify

DaanDeMeyer:
Let's avoid triggering ukify's autodetect logic by passing in the
kernel version ourselves.


DaanDeMeyer:
cc @josefbacik this reverts your PR in favor of passing the kernel version to ukify explicitly from mkosi so ukify never tries to call readelf in the first place.

keszybz (approved):
LGTM.

behrmann (approved):

Issue #1863, opened 2023-09-04
Using distro-provided secure boot shim


septatrix:
Most (if not all) distributions supported by mkosi are capable of using a signed UEFI shim to enable secure boot usage. Is it possible to configure mkosi such that this shim is invoked?

This would make using this tools more useful for providing images to customers without requiring them to disable secure boot or manually enroll signing keys. With shim it would boot up even with secure boot enabled and then ask the user to trust the new keys whenever the signing key changes.

septatrix:
If found a script in the Constellation repo which seems to [put shim before systemd-boot in the boot process](https://github.com/edgelesssys/constellation/blob/main/image/secure-boot/signed-shim.sh). Maybe this can be adjusted and incorporated into mkosi? I have not yet tried it though – maybe someone more knowledgeable in the generation of bootable images can say if this should work.

behrmann:
I know that @DaanDeMeyer has done some work in that direction, this is why grub support was added back in #1797. Something like the thing linked above might already be achievable with a script.

malt3:
I wrote the script in Constellation. The basic mechanism is this:

- The shim is placed in `/efi/EFI/BOOT/BOOTX64.EFI`
- `mmx64.efi` is placed under `/efi/EFI/BOOT/mmx64.efi`
- The next stage (`systemd-bootx64.efi`, grub or a UKI) is placed under `efi/EFI/BOOT/grubx64.efi` (This is where shim expects the next stage)

The implementation is pretty bad. It would be a lot cleaner to place files under `/efi` before repart runs.

septatrix:
One aspect which came to my mind right now is that for these cases it would be good to have control over the UEFI variables. I.e. somehow tell mkosi to not load `/usr/share/OVMF/OVMF_VARS.fd` (it cannot be removed or overwritten with additional Qemu args) such that a custom file could be provided, or even better allow it to directly load `/usr/share/OVMF/OVMF_VARS.secboot.fd` if found which contains Microsoft's keys preloaded

septatrix:
Steps which seem to be working:

1. Modified local mkosi installation to use the `OVMF_VARS.secboot.fd`
2. Generate local keypair using `mkosi genkey`
3. Convert the certificate to DER format `openssl x509 -in mkosi.crt -outform DER -out mkosi.crt.der`
4. Add `mkosi.finalize` script which puts everything in place
5. Generate image and boot it with qemu `mkosi qemu`
[...]

PR #2152, opened 2023-12-10
Add QemuFirmwareVariables=

DaanDeMeyer:
This allows configuring the path to the qemu firmware variables to use. This allows users to configure their own variables using https://pypi.org/project/virt-firmware/ before passing it to mkosi.

### DIFF
--- a/mkosi/resources/mkosi.md
+++ b/mkosi/resources/mkosi.md
@@ -1290,13 +1290,25 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 
 `QemuFirmware=`, `--qemu-firmware=`
 
-: When used with the `qemu` verb, this option which firmware to use.
-  Takes one of `uefi`, `bios`, `linux`, or `auto`. Defaults to `auto`.
-  When set to `uefi`, the OVMF firmware is used. When set to `bios`, the
-  default SeaBIOS firmware is used. When set to `linux`, direct kernel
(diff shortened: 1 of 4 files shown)


behrmann (approved):
LGTM, but the comment about what tooling to use in the commit message might be a bit too well-hidden :) An example in the docs would be nice.

mkosi/resources/mkosi.md
```diff
@@ -1290,13 +1290,25 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
+`QemuFirmwareVariables=`, `--qemu-firmware-variables=`
+
+: When used with the `qemu` verb, this option specifies the path to the
+  the firmware variables file to use. Currently, this option is only
+  taken into account when the `uefi` firmware is used. If not specified,
+  mkosi will search for the default variables file and use that instead.
+
+: `virt-fw-vars` from the
+  [virt-firmware](https://gitlab.com/kraxel/virt-firmware) project can
+  be used to customize OVMF variable files.
```
septatrix, line 1311: Maybe also mention the `/usr/share/OVMF/OVMF_VARS.secboot.fd` file including the default Microsoft secure boot keys which would resolve the comment in <https://github.com/systemd/mkosi/issues/1863#issuecomment-1780055634>?

DaanDeMeyer: @septatrix Can you make a PR to add this? Happy to review that.

PR #2154, opened 2023-12-10
Use kernel-install entry token if available

DaanDeMeyer:
Let's try and look for the kernel-install entry token if
kernel-install v255 or newer is available. This makes us more
compatible with package manager upgrades as we'll use the same
directory that kernel-install will look for when invoked by a
package manager.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,17 @@
 # mkosi Changelog
 
+## v20
+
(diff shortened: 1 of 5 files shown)


behrmann (commented):
Mostly LGTM. Not sure about the last one, because I have the vague deja vu of you just changing this the other way 'round. It definitely needs a NEWS entry.

One question, though...

behrmann (approved):

PR #2158, opened 2023-12-11
Add RepartOffline= option

DaanDeMeyer:
Instead of auto-detecting all cases where --offline=no has to be used with systemd-repart, let's allow configuring it via an option so that if we discover any new cases, users can easily disable offline mode themselves.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,10 @@
 
 ## v20
 
+- We don't automatically set `--offline=no` anymore when we detect the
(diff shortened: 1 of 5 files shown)


behrmann (approved):
LGTM, but the removed check to set `--offline` in case of `Subvolumes=` should be mentioned in the NEWS and maybe a name like `RepartOffline=` or `OfflineRepart=` would be clearer to users, than the somewhat generic `Offline=`, especially in the summary.

PR #2162, opened 2023-12-11
Look for vmlinuz.efi as well

DaanDeMeyer:
On aarch64, vmlinuz.efi is installed instead of vmlinuz so let's make sure we look for both.


behrmann (commented):
LGTM, though as I understand what you are saying above, it's `vmlinuz.efi` on arm64 and `vmlinuz` otherwise. In that case I think it would be cleaner to make the name of `vmlinuz` a method on `Architecture` instead of checking both all the time.

DaanDeMeyer:
It's for more than just arm so I'd prefer to play it safe and do it in general.

PR #2163, opened 2023-12-12
make_cpio: sort files used as cpio input

malt3:
Pathlib's glob functions return files in the order used by the filesystem. This may differ between implementations and configuration (file system, locale). For better reproducibility, the file list should be sorted.

### DIFF
--- a/mkosi/archive.py
+++ b/mkosi/archive.py
@@ -89,6 +89,7 @@ def extract_tar(src: Path, dst: Path, log: bool = True) -> None:
 def make_cpio(src: Path, dst: Path, files: Optional[Iterable[Path]] = None) -> None:
     if not files:
         files = src.rglob("*")
+    files = sorted(files)
 
     log_step(f"Creating cpio archive {dst}…")
     bwrap(


mkosi/archive.py
```diff
@@ -89,6 +89,8 @@ def extract_tar(src: Path, dst: Path, log: bool = True) -> None:
 def make_cpio(src: Path, dst: Path, files: Optional[Iterable[Path]] = None) -> None:
     if not files:
         files = src.rglob("*")
+    files = list(files)
+    files.sort()
```
behrmann, line 93:
```suggestion
        files = sorted(src.rglob("*"))
```

malt3: That if the caller passed `files` as an unsorted iterable?

behrmann: Then 
```suggestion
        files = src.rglob("*")
    files = sorted(files)
```

behrmann (approved):

DaanDeMeyer (approved):

PR #2164, opened 2023-12-12
mkosi-initrd: Rewrite kernel-install plugin in python and support UKIs

DaanDeMeyer:
There's really no point for the kernel-install plugin to be in bash,
so let's rewrite it in python. While we're at it, let's also support
running as a UKI generator.


mkosi/resources/mkosi-initrd/mkosi.conf
```diff
@@ -1,8 +1,7 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
 [Output]
-ImageId=initrd
```
behrmann, line 4: Why not `@ImageId=initrd` instead?

kernel-install/50-mkosi.install
```diff
@@ -0,0 +1,136 @@
...
+    verbose: bool
+
+
+def we_are_wanted(context: Context) -> bool:
+    return context.uki_generator == "mkosi" or context.initrd_generator in ("mkosi", "mkosi-initrd")
+
+
+def mandatory_variable(name: str) -> str:
+    try:
+        return str(os.environ[name])
```
behrmann, line 37: `os.envget` can return `None`, but `os.environ` is str-only, no? mypy doesn't complain about this for me.
```suggestion
        return os.environ[name]
```

kernel-install/50-mkosi.install
```diff
@@ -0,0 +1,136 @@
...
+                   version=f"mkosi {__version__}")
+
+    context = Context(
+        **vars(p.parse_args()),
+        staging_area=Path(mandatory_variable("KERNEL_INSTALL_STAGING_AREA")),
+        layout=mandatory_variable("KERNEL_INSTALL_LAYOUT"),
+        image_type=mandatory_variable("KERNEL_INSTALL_IMAGE_TYPE"),
+        initrd_generator=os.getenv("KERNEL_INSTALL_INITRD_GENERATOR"),
+        uki_generator=os.getenv("KERNEL_INSTALL_UKI_GENERATOR"),
+        verbose=int(os.getenv("KERNEL_INSTALL_VERBOSE", "0")) > 0,
```
behrmann, line 84: Nit: 0 doesn't need to be a string here and I think it looks nicer
```suggestion
        verbose=int(os.getenv("KERNEL_INSTALL_VERBOSE", 0)) > 0,
```

behrmann (commented):
Looks reasonable to me, a few nits

mkosi/resources/mkosi-initrd/mkosi.conf
DaanDeMeyer: Because that doesn't seem like that useful an image ID. It doesn't really identify anything

kernel-install/50-mkosi.install
```diff
@@ -0,0 +1,136 @@
...
+from mkosi.log import die, log_setup
+from mkosi.run import run, uncaught_exception_handler
+from mkosi.types import PathString
+
+
+class Context(NamedTuple):
+    command: str
+    kernel_version: str
+    entry_dir: Path
+    kernel_image: Path
```
behrmann, line 21: No idea why mypy doesn't see this, but since the argparse default is `None` and `entry_dir` and `kernel_image` are `?`, in most cases `None` will be passed into the `Context` constructor further down. 
```suggestion
    entry_dir: Optional[Path]
    kernel_image: Optional[Path]
```

behrmann (approved):

kernel-install/50-mkosi.install
DaanDeMeyer: I made these required instead as they're documented to always be passed by kernel-install

PR #2166, opened 2023-12-12
Make kernel image lookup logic more robust

DaanDeMeyer:


mkosi/__init__.py
```diff
@@ -1268,13 +1268,17 @@ def gen_kernel_images(state: MkosiState) -> Iterator[tuple[str, Path]]:
...
+        # do weird stuff. But let's make sure we're not returning UKIs as the
+        # UKI on Fedora is named vmlinuz-virt.efi.
+        for kimg in kver.glob("vmlinuz*"):
+            if KernelType.identify(kimg) != KernelType.uki:
                 break
         else:
             continue
 
-        yield kver.name, Path("usr/lib/modules") / kver.name / f
+        yield kver.name, kimg
```
behrmann, line 1281: How about
```python
        for kimg in kver.glob("vmlinuz*"):
            if KernelType.identify(kimg) != KernelType.uki:
                yield kver.name, kimg
                break
```

PR #2171, opened 2023-12-13
kernel-install: Use host's package manager configuration and repos

DaanDeMeyer:
Let's make sure we use the host's package manager configuration and
repositories in the kernel-install plugin. The initrd we produce
should be as compatible with the host as we can make it and making
sure we use the same packages that the host uses is a good step in
achieving that.


behrmann:
I'm not entirely sure about the the first commit in the series, since I think the config in proper config files should probably be the source of truth, but I don't mind too much. LGTM otherwise.

DaanDeMeyer:
> I'm not entirely sure about the the first commit in the series, since I think the config in proper config files should probably be the source of truth, but I don't mind too much. LGTM otherwise.

The problem is that the config files from the host will probably not disable installation of recommended packages, which we absolutely need disabled when building the initrd, so hence we override with our own option.

PR #2173, opened 2023-12-13
Default to ttyAMA0 when building/booting arm images

DaanDeMeyer:
On ARM, the first serial console is generally called ttyAMA0, so let's configure that as the console instead of ttyS0.


behrmann (approved):
LGTM, but the name `tty` for the method on `Architecture` seems very generic. Something more descriptive like `default_serial_tty` or such would be nice.

PR #2177, opened 2023-12-14
Stop using the tools tree for the ssh verb

DaanDeMeyer:
This allows us to run ssh out of the user namespace which means we can get rid of the passwd hack to make ssh work. ssh is widespread enough that we can require users to install it on the host machine instead of using the tools tree.


behrmann (approved):
LGTM. Just to be sure, is the second commit to counter the CI failure I see once in a while?

DaanDeMeyer:
The second commit is because on our internal systems there is something that's blocking us from running `umount /usr` even when running as root in a private mount namespace with slave mounts. Figured I'd solve that error by simply not unmounting.

PR #2178, opened 2023-12-14
Fix install_tree()

DaanDeMeyer:
Let's make sure that all the skeleton, extra and package manager trees we get have absolute targets. That allows us to stop using with_prefix() when installing these trees, which means we pass target=None instead of target="/" which makes install_tree do the right thing.


malt3 (approved):
This fixes the use of tar files in `--extra-tree` (and other trees).

behrmann (approved):

PR #2181, opened 2023-12-15
Add support for loading environment files

### LINKED ISSUE #738: Support environment file(s)
Currently mkosi supports specification of environment variables via cli and inside the `mkosi.default` file.
Both variants are tedious to use when variables are generated outside of mkosi.
I propose a new option `EnvironmentFile` (`mkosi.env` as default) which takes multiple file paths that are parsed & translated into nspawn env options when building the container and set inside the `mkosi.finalize` script.
The file format that is used to describe the variables inside these files could be a "plain" shell-like format like systemd uses, which can be generated outside of mkosi.

hellozee:
Fixes #738

Was poking around the source, felt like a simple fix. Hopefully I haven't missed something 😅


mkosi/config.py
```diff
@@ -1677,6 +1678,18 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
         parse=config_make_list_parser(delimiter=" ", unescape=True),
         help="Set an environment variable when running scripts",
     ),
+    MkosiConfigSetting(
+        dest="environment_files",
+        long="--environment-files",
```
DaanDeMeyer, line 1683: Please make this `--env-file` i.e. we use singular instead of plural for CLI options and something make them a bit shorter

mkosi/config.py
```diff
@@ -1677,6 +1678,18 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
...
+        metavar="PATH",
+        name="EnvironmentFiles",
```
DaanDeMeyer, line 1685: Please drop this since it's the default value for `name`

mkosi/config.py
```diff
@@ -1677,6 +1678,18 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
...
+        section="Content",
+        parse=config_make_list_parser(delimiter=",", parse=make_path_parser()),
+        paths=("mkosi.env",),
+        path_default=False,
+        help="Enviroment files to set when running scripts",
+        compat_names=("EnvironmentFile",),
```
DaanDeMeyer, line 1691: Please drop this as well, it's a new option, there are no compatibility names for this option.

mkosi/config.py
```diff
@@ -2879,7 +2892,16 @@ def load_environment(args: argparse.Namespace) -> dict[str, str]:
...
 
-    for s in args.environment:
+    environment_values = []
+    for env_file in args.environment_files:
+        with open(env_file) as f:
+            environment_values.extend(f.read().strip().split("\n"))
+    if len(environment_values) != 0:
+        environment_values.extend(args.environment)
+    else:
+        environment_values = args.environment
```
DaanDeMeyer, line 2902: Maybe something like:

```
env = args.environment + [f.read_text().splitlines() for f in args.environment_files]
...
```

?

mkosi/resources/mkosi.md
```diff
@@ -976,6 +976,14 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
   listed variables will be set. If the same variable is set twice, the
   later setting overrides the earlier one.
 
+`EnvironmentFiles=`, `--environment-files=`
+
+: Takes a comma-separated list of paths to files that contain enviroment
+  variable definitions to be added to the scripting environment. Uses
+  `mkosi.env` if it is found in the local directory. The variables are
+  first read from `mkosi.env` if it exists, then from the given list of
+  files and then from the `--environment` flag.
```
DaanDeMeyer, line 985: "--environment flag" => "Environment= setting"

mkosi/resources/mkosi.md
```diff
@@ -976,6 +976,14 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
   listed variables will be set. If the same variable is set twice, the
   later setting overrides the earlier one.
 
+`EnvironmentFiles=`, `--environment-files=`
```
DaanDeMeyer, line 979: --env-file

mkosi/config.py
```diff
@@ -2879,7 +2892,16 @@ def load_environment(args: argparse.Namespace) -> dict[str, str]:
...
+
+    for s in environment_values:
         key, sep, value = s.partition("=")
         value = value if sep else os.getenv(key, "")
         env[key] = value
```
DaanDeMeyer, line 2907: Now that we support env files we should call `strip()` on both keys and values here to allow users to put whitespace in their env files on lines if they'd like.

tests/test_config.py
```diff
@@ -746,10 +759,14 @@ def test_specifiers(tmp_path: Path) -> None:
             "ImageVersion": "1.2.3",
             "OutputDirectory": str(Path.cwd() / "abcde"),
             "Output": "test",
+            "TestValue": "100"
         }
 
         assert {k: v for k, v in config.environment.items() if k in expected} == expected
 
+        expected_env_files = [Path.cwd() / "mkosi.env", Path.cwd() / "other.env"]
+        assert config.environment_files == expected_env_files
```
DaanDeMeyer, line 768: Please make a new test `test_environment()` for this instead of putting it in `test_specifiers()`.

DaanDeMeyer (changes requested):
Looks great! A few comments

hellozee:
damn thats a fast review

tests/test_json.py
```diff
@@ -308,6 +309,7 @@ def test_config() -> None:
         dependencies = ("dep1",),
         distribution = Distribution.fedora,
         environment = {},
+        environment_files=[],
```
DaanDeMeyer, line 312: Please put whitespace between '=' to be consistent with the rest of the file

tests/test_config.py
```diff
@@ -751,5 +751,56 @@ def test_specifiers(tmp_path: Path) -> None:
         assert {k: v for k, v in config.environment.items() if k in expected} == expected
 
 
+
```
DaanDeMeyer, line 754: Spurious whitespace

tests/test_config.py
```diff
@@ -751,5 +751,56 @@ def test_specifiers(tmp_path: Path) -> None:
...
 def test_deterministic() -> None:
     assert MkosiConfig.default() == MkosiConfig.default()
+
```
DaanDeMeyer, line 757: We always put two newlines between functions

tests/test_config.py
```diff
@@ -751,5 +751,56 @@ def test_specifiers(tmp_path: Path) -> None:
...
+        [Distribution]
+        Distribution=ubuntu
+        Release=lunar
+        Architecture=arm64
+
+        [Output]
+        ImageId=my-image-id
+        ImageVersion=1.2.3
+        OutputDirectory=abcde
+        Output=test
```
DaanDeMeyer, line 772: Please drop this as it's irrelevant for this test

mkosi/config.py
```diff
@@ -2879,9 +2890,13 @@ def load_environment(args: argparse.Namespace) -> dict[str, str]:
     if dnf := os.getenv("MKOSI_DNF"):
         env["MKOSI_DNF"] = dnf
 
-    for s in args.environment:
+    environment_values : list[str] = sum([f.read_text().strip().splitlines() for f in args.environment_files], [])
```
DaanDeMeyer, line 2893: Why `sum()`? That shouldn't be necessary. Also please just name this `env` instead of `environment_values`. We prefer shorter variable names for local variables.

mkosi/config.py
```diff
@@ -2879,9 +2890,13 @@ def load_environment(args: argparse.Namespace) -> dict[str, str]:
...
+    environment_values.extend(args.environment)
+    for s in environment_values:
         key, sep, value = s.partition("=")
+        key, value = key.strip(), value.strip()
         value = value if sep else os.getenv(key, "")
+        print(key, value)
```
DaanDeMeyer, line 2899: Leftover `print`

mkosi/config.py
DaanDeMeyer: There should also be no need to specify the type explicitly here, so please drop that.

mkosi/config.py
```diff
@@ -2879,9 +2890,13 @@ def load_environment(args: argparse.Namespace) -> dict[str, str]:
     if dnf := os.getenv("MKOSI_DNF"):
         env["MKOSI_DNF"] = dnf
 
-    for s in args.environment:
+    environment_values : list[str] = sum([f.read_text().strip().splitlines() for f in args.environment_files], [])
+    environment_values.extend(args.environment)
+    for s in environment_values:
```
DaanDeMeyer, line 2895: `for s in env + args.environment:`

DaanDeMeyer (changes requested):

mkosi/config.py
```diff
@@ -2879,8 +2890,10 @@ def load_environment(args: argparse.Namespace) -> dict[str, str]:
     if dnf := os.getenv("MKOSI_DNF"):
         env["MKOSI_DNF"] = dnf
 
-    for s in args.environment:
+    entries = [x for y in (f.read_text().strip().splitlines() for f in args.environment_files) for x in y]
```
DaanDeMeyer, line 2893: This still seems overly complex, what about the following:

```
entries = [l for f in args.environment_files for l in f.read_text().strip().splitlines()]
```

tests/test_config.py
```diff
@@ -753,3 +753,43 @@ def test_specifiers(tmp_path: Path) -> None:
...
+        expected = {
+            "TestValue1": "100",
+            "TestValue2": "300",
+            "TestValue3": "400"
+        }
+
+        assert {k: v for k, v in config.environment.items() if k in expected} == expected
+
+        expected_env_files = [Path.cwd() / "mkosi.env", Path.cwd() / "other.env"]
+        assert config.environment_files == expected_env_files
```
DaanDeMeyer, line 795: These two lines could just be merged without the variable

DaanDeMeyer (changes requested):

mkosi/config.py
hellozee: oo this works, thanks

mkosi/config.py
```diff
@@ -2879,8 +2890,10 @@ def load_environment(args: argparse.Namespace) -> dict[str, str]:
     if dnf := os.getenv("MKOSI_DNF"):
         env["MKOSI_DNF"] = dnf
 
-    for s in args.environment:
+    entries = [x for f in args.environment_files for x in f.read_text().strip().splitlines()]
```
behrmann, line 2893: Nit: `x` is a very generic choice, let's give them proper names, to make this clearer and not much longer
```suggestion
    entries = [line for envfile in args.environment_files for line in f.read_text().strip().splitlines()]
```
(or at least give it a more telling short name like `e` or `l`, but proper names are nicer)

tests/test_config.py
```diff
@@ -753,3 +753,42 @@ def test_specifiers(tmp_path: Path) -> None:
...
+        """
+    )
+
+    with chdir(d):
+        _, [config] = parse_config()
+
+        expected = {
+            "TestValue1": "100",
+            "TestValue2": "300",
+            "TestValue3": "400"
```
behrmann, line 789: Nit: missing comma
```suggestion
            "TestValue3": "400",
```

behrmann, line 789: To make this a bit clearer
```suggestion
            "TestValue1": "100", # from other.env
            "TestValue2": "300", # from mkosi.conf
            "TestValue3": "400", # from mkosi.conf
```

This also shows that this doesn't test a setting from `mkosi.env` surviving, which would be nice.

tests/test_config.py
```diff
@@ -753,3 +753,42 @@ def test_specifiers(tmp_path: Path) -> None:
...
+        }
+
+        assert {k: v for k, v in config.environment.items() if k in expected} == expected
```
behrmann, line 792: Why the `if k in expected`? All keys from the env are in expected, so the only thing this can do is hide something unexpected, which a test should catch
```suggestion
        assert {k: v for k, v in config.environment.items()} == expected
```

behrmann (commented):
Very nice! A few last nits from me

tests/test_config.py
hellozee: we have a couple more entries in the environment which is injected my mkosi, not in the the expected dict

DaanDeMeyer (approved):

tests/test_config.py
behrmann: Good point! Though I'd express this with a slightly different focus and add a comment for why this looks the way it does
```suggestion
        # Only check values for keys from expected, as config.environment contains other items as well
        assert {k: config.environment[k] for k in expected.keys()} == expected
```

behrmann (approved):

tests/test_config.py
hellozee: aye aye

DaanDeMeyer:
@hellozee For any following pull requests, please use `git commit --amend` to fix your existing commit instead of pushing new commits to fix review comments in earlier commits.

DaanDeMeyer:
Thanks for contributing!

hellozee:
> please use `git commit --amend` to fix your existing commit instead of pushing new commits to fix review comments in earlier commits.

ahh, ok, my bad, I assumed force pushes were not allowed

PR #2182, opened 2023-12-15
Use mkosi.key/mkosi.crt for SSH authentication

DaanDeMeyer:
Instead of using the user's SSH certificate and key, let's use the
X509 certificate and private key generated by 'mkosi genkey' instead.

This saves us from having to rely on ssh-agent to get the public key
or doing otherwise complicated logic to try and find the public and
private key. We also avoid always needing a separate public/private
key just for SSH by automatically converting the X509 certificate to
a SSH public key.


behrmann (approved):
LGTM implementatiion-wise. I'm not entirely sure about the UX of going for certs. We already rely on openssh being on the host and most people will likely already have an ssh agent running. I'm sure we'll confuse those users by "not just using the agent". On the other hand, I know how people can struggle with setting up an ssh agent, so telling them to run `mkosi genkey` once will lower the hurdle for them. That being said, if this is only in the NEWS it will be lost on users, let's add a paragraph to the ssh verb docs as well, that we don't use the agent, but certs and people need to run `genkey`.

PR #2183, opened 2023-12-15
Support booting Fedora default image on s390x

DaanDeMeyer:


mkosi/__init__.py
```diff
@@ -1412,6 +1414,24 @@ def build_kernel_modules_initrd(state: MkosiState, kver: str) -> Path:
...
+
+    seq = []
+    for p in initrds:
+        initrd = p.read_bytes()
+        n = len(initrd)
+        padding = b'\0' * (round_up(n, 4) - n)  # pad to 32 bit alignment
+        seq += [initrd, padding]
+
+    output.write_bytes(b''.join(seq))
+    return output
```
behrmann, line 1432: it doesn't have to be high performance and the usual case will probably be a single initrd, but this would still be nicer with a lot less copying around of the stuff by using [`io.BytesIO`](https://docs.python.org/3.11/library/io.html?highlight=bytesio#io.BytesIO), as a simple in memory file like buffer, that add the end produces a bytestring value that can be written out.

mkosi/architecture.py
```diff
@@ -121,11 +121,16 @@ def default_serial_tty(self) -> str:
         return {
             Architecture.arm   : "ttyAMA0",
             Architecture.arm64 : "ttyAMA0",
+            Architecture.s390  : "ttysclp0",
```
behrmann, line 124: Do we need s390? While s390x is alive, I remember s390 being a dead arch like itanium or superh or m68k

behrmann (commented):
Looks reasonable, but there should maybe be a separate tests for whether single und multiple match sections work as expected so we don't regress there in the future.

mkosi/architecture.py
DaanDeMeyer: It is a part of the discoverable partitions spec, so I don't think it hurts to do minimal effort to try and support it. I would treat it the same as s390x in all cases.

huoqifeng:
@DaanDeMeyer sorry for bothering you in a closed PR, I think this PR is to enable an image on s390x and make it launched via qemu? May I know an example mkosi.conf that worked?

PR #2186, opened 2023-12-17
Output file names now use `Output=` without version appended

NekkoDroid:
Previously when `Output=` was not specified its value was set to the value of `ImageId=` or `image` and only when actually writing the file was the version appended. This did not match the documentation, which states that the value of `Output=` is set to `ImageId=` and `ImageVersion=` if they exist. The main problem this causes is that you cannot specify `%o` to get the entire output name (without extension) and it is impossible supporting a format with and without version.

This came to my attention when I was adding a custom Initrd and wanting to reference it as `%O/%o.cpio.zst` when I had a version specified.

This change primarily adjust the implementation matching the documentation. A side-effect which appeared during the adjust is in how clearing of output files are handled: When specifying `clean -ff` or `-fff` elsewhere it now clears the entire output directory instead of the output name without version prefix.

The other incompatibility to previous versions to my knowledge is if someone explicitly specified `Output=` and referenced it with `%o_%v` as a workaround (would need to be updated to simply `%o`) and if you want preserve the output name with the version appended `Output=%i_%v` needs to be used (or `Output=` needs to be omitted)

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -525,6 +525,15 @@ def config_default_compression(namespace: argparse.Namespace) -> Compression:
         return Compression.none
 
 
+def config_default_output(namespace: argparse.Namespace) -> str:
+    output = namespace.image_id or namespace.image or "image"
+
+    if namespace.image_version:
+        output += f"_{namespace.image_version}"
+
+    return output
+
+
 def config_default_distribution(namespace: argparse.Namespace) -> Distribution:
     detected = detect_distribution()[0]
 
@@ -1553,6 +1562,8 @@ SETTINGS = (
         section="Output",
         specifier="o",
         parse=config_parse_output,
+        default_factory=config_default_output,
+        default_factory_depends=("image_id", "image_version"),
         help="Output name",
     ),
     MkosiConfigSetting(
@@ -3104,9 +3115,6 @@ def load_config(args: MkosiArgs, config: argparse.Namespace) -> MkosiConfig:
 
     if config.sign:
         config.checksum = True
-
-    if config.output is None:
-        config.output = config.image_id or config.image or "image"
 
     config.credentials = load_credentials(config)
     config.kernel_command_line_extra = load_kernel_command_line_extra(config)
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -1233,17 +1233,8 @@ class MkosiConfig:
         })
 
     @property
-    def output_with_version(self) -> str:
-        output = self.output
-
-        if self.image_version:
-            output += f"_{self.image_version}"
-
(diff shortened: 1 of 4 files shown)


DaanDeMeyer:
We create a symlink in the output directory without the version in it that points to the latest version. Doesn't that work for this use case?

NekkoDroid:
I just tested with setting `Initrds=%O/%o` and it complains `Initrd /.../mkosi.output/NekkOS.raw` not found. I assume the extension gets appended when non is specified?

Partitally another reason I did this change is to be able to output to `mkosi.output/%v/%o.ext`

DaanDeMeyer:
`Initrds=%O/%o` will point to the current output (so the image that's using the initrd, not the initrd output itself). You can't use `%o` to refer to the output of a previous image build generally.

You might want to look into the newly added `InitrdInclude=` which allows you to just extend the default initrd and you won't have to bother with `Initrds=`.

NekkoDroid:
I did know that `%O/%o` would point to the current image, the plan was for the initrd and the regular system having the same output name, with exception of their extension. And I am mostly providing a custom initrd is to just have a bit more minimal initrd (and somewhat because I just want to "have less magic" in the background happening).

As implied, this patch patch isn't exactly something that is 100% necessary for me, but being able to output the files without the version attached (since the files would be organized under a folder with the version name) is something I am somewhat looking for. And the documentation could use a bit of clarification if keeping current behavior.

mkosi/config.py
```diff
@@ -2909,6 +2900,9 @@ def load_config(args: MkosiArgs, config: argparse.Namespace) -> MkosiConfig:
     if config.output is None:
         config.output = config.image_id or config.image or "image"
 
+        if config.image_version:
+            config.output += f"_{config.image_version}"
```
DaanDeMeyer, line 2904: We should move this logic into a `default_factory` for the `output` setting. So add `config_default_output()`, put this logic in there and then use it as the `default_factory` field for the `output` setting. with `default_factory_depends` on the `image_id` and `image_version` settings.

DaanDeMeyer (changes requested):
So I think it makes sense to do this, except the first commit, we really cannot do that. So I would just accept that clearing the entire output directory is not something we can do automatically when dealing with versions.

Instead, just drop the `output_with_version` logic from `unlink_output()`, and only keep the logic that removes everything prefixed with the output.

NekkoDroid:
Sorry that it took a bit longer but it should all be done now. Also reorganized the commits a bit.

DaanDeMeyer (approved):

DaanDeMeyer:
Thanks!

PR #2191, opened 2023-12-19
Stop bind mounting /sys in chroot environments

DaanDeMeyer:
The only reason we do this is to make systemd's unit test suite pass. https://github.com/systemd/systemd/pull/30527 fixes systemd's test suite to not fail when /sys is not mounted, so let's drop this bit.


behrmann (approved):
I wonder if this will break anybody, but let's see if somebody complains to add a knob for this.

PR #2196, opened 2023-12-20
ci: Build dnf based distros with opensuse tools tree

DaanDeMeyer:
dnf-data on opensuse had the missing dependency on ln added so we can now use opensuse tools trees to build dnf based distro images.


DaanDeMeyer:
Or so I thought but not just yet it seems

PR #2197, opened 2023-12-20
Use virtio nic model on s390x

DaanDeMeyer:
Also refactor things a bit to move more into the Architecture enum.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -308,7 +308,6 @@ class Architecture(StrEnum):
             Architecture.loongarch64 : "loongarch64",
         }.get(self)
 
-
     def to_qemu(self) -> str:
         a = {
             Architecture.alpha       : "alpha",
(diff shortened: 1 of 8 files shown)


mkosi/config.py
```diff
@@ -353,6 +353,30 @@ def supports_smbios(self, firmware: QemuFirmware) -> bool:
...
+            Architecture.arm64    : "virt",
+            Architecture.s390     : "s390-ccw-virtio",
+            Architecture.s390x    : "s390-ccw-virtio",
+            Architecture.ppc      : "pseries",
+            Architecture.ppc64    : "pseries",
+            Architecture.ppc64_le : "pseries",
+        }
+
+        if self not in m:
+            die(f"No qemu machine defined for architecture {self}")
```
behrmann, line 370: Before, this defaulted to `q35`, while now it dies immediately. I assume the change is because it before qemu would have thrown an error instead. Explaining the reasoning behind this change in the commit message would be nice.

behrmann (approved):

PR #2204, opened 2023-12-21
Install centos/fedora packaging tools in default images

DaanDeMeyer:
Useful for doing Fedora/CentOS packaging work.


davide125:
We should add rpminspect too: `rpminspect`, `rpminspect-data-centos`, `rpminspect-data-fedora`

PR #2210, opened 2023-12-22
Mount package manager trees

DaanDeMeyer:
Now that /etc and /var are free game when running within bwrap() because we don't mount in the directories from the host anymore, let's take advantage of that by mounting all our package manager configuration to the canonical location in /etc instead of configuring the package managers via their CLI or config file to look in the right directory.

This also makes us look for rpm configuration in /etc/rpm instead of /usr/lib/rpm as that's now possible.

### DIFF
--- a/mkosi/installer/apt.py
+++ b/mkosi/installer/apt.py
@@ -14,25 +14,21 @@ def setup_apt(state: MkosiState, repos: Sequence[str]) -> None:
     (state.pkgmngr / "etc/apt/apt.conf.d").mkdir(exist_ok=True, parents=True)
     (state.pkgmngr / "etc/apt/preferences.d").mkdir(exist_ok=True, parents=True)
     (state.pkgmngr / "etc/apt/sources.list.d").mkdir(exist_ok=True, parents=True)
-    (state.pkgmngr / "var/log/apt").mkdir(exist_ok=True, parents=True)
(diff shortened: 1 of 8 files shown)


mkosi/installer/apt.py
```diff
@@ -72,19 +66,16 @@ def apt_cmd(state: MkosiState, command: str) -> list[PathString]:
         "-o", "APT::Get::Allow-Remove-Essential=true",
         "-o", "APT::Sandbox::User=root",
         "-o", f"Dir::Cache={state.cache_dir / 'apt'}",
-        "-o", f"Dir::State={state.pkgmngr / 'var/lib/apt'}",
-        "-o", f"Dir::State::status={state.root / 'var/lib/dpkg/status'}",
-        "-o", f"Dir::Etc::trusted={trustedkeys}",
-        "-o", f"Dir::Etc::trustedparts={trustedkeys_dir}",
-        "-o", f"Dir::Log={state.pkgmngr / 'var/log/apt'}",
-        "-o", f"Dir::Bin::dpkg={shutil.which('dpkg')}",
+        "-o", f"Dir::State={state.cache_dir / 'apt'}",
```
behrmann, line 69: Could cache and state dir in the same directory collide?

behrmann (approved):
LGTM, one question

mkosi/installer/apt.py
DaanDeMeyer: They don't seem to on my system, should be fine

PR #2223, opened 2024-01-01
Use bubblewrap to set up the tools tree instead of doing it ourselves

DaanDeMeyer:
The problem with overmounting the host's /usr (in a private mount
namespace) is that we have no control over the symlinks in the root
directory (/lib, /bin, /lib64) and if these symlinks don't match
between the host distribution and the tools tree distribution, all
kinds of weird breakage starts happening. For example, using Fedora
tools trees on Arch Linux is currently broken because /lib64 on Arch
Linux points to /usr/lib whereas on Fedora it points to /usr/lib64.

Because we can't (and shouldn't) modify the symlinks of the host's
root filesystem, we need to set up the tools tree in a sandbox that
we chroot into, so that we have full control over the rootfs of the
sandbox and can make sure the symlinks are correct. Luckily, we
already do just that with bubblewrap, except that currently we mount
the tools tree over /usr ourselves and then just carry that over into
the bubblewrap sandbox.

Instead, we stop mounting over the host's /usr ourselves and have
bubblewrap pick the right /usr itself. We also copy the symlinks from
the tools tree or the host if there is no tools tree.

Because we don't mount over the host's /usr anymore, we have to run
every tool that should come from the tools tree with bubblewrap now.
The side effect of this is that almost all of our tools now run
sandboxed. We also have to make use of find_binary() everywhere
instead of shutil.which() to make sure we look for binaries in the
tools tree when required. Various other codepaths that look into /usr
also have to be modified to look into the tools tree when needed.

Also, because we don't unshare the user namespace in the main mkosi
process anymore now, we can get rid of a lot of chown()'s in qemu.py
and opening the qemu device file descriptors can be moved into
run_qemu() itself.

We also don't have to make sure all python modules are loaded anymore
as the host's /usr is never overmounted so the required python modules
will be available for the entire runtime of the mkosi process.

Because virtiofsd is now executed with bubblewrap, we use bubblewrap
to set up the required uidmap instead of relying on virtiofsd to do it
with newuidmap/newgidmap. Note that this breaks RuntimeTrees= as
virtiofsd unconditionally tries to drop groups with setgroups() which
fails with EPERM in an unprivileged user namespace set up by bubblewrap.
This is fixed by https://gitlab.com/virtio-fs/virtiofsd/-/merge_requests/207
which is still awaiting review.

To make this work codewise, this commit renames the bwrap() function
to sandbox_cmd() (similar to chroot_cmd() and apivfs_cmd()) which now
returns a command line instead of executing the command itself. run()
is modified to take an extra "sandbox" arguments which is simply the
part of the full command that sets up the sandbox. Context and Config
both learn new sandbox() methods which set up the sandbox for each
object respectively (mostly by adding extra bind mounts).

Because almost every call to run() now takes a sandbox, this gives us
a lot of control over the individual environment for each tool we run.
We make use of this to restrict each tool we run to the minimal possible
sandbox that that tool needs to run. By specifically mounting in the
required paths for each tool we run, we also make sure these are always
available instead of relying that somewhere we mount a path that has the
input in it.

Because we allow passing arbitrary options to mkosi qemu, mkosi boot and
various other verbs, we run these verbs with a relaxed sandbox, where we
mount in most directories from the host. This means that whatever
directories users specify will be available.

In terms of CI, the extra sandboxing means that our previous approach of
building various systemd binaries from source and symlinking them to
/usr/bin doesn't work anymore. Instead, we opt to always use tools trees
and drop the host builds from the testing matrix. This also simplifies
and speeds up the github action as we don't have to compile systemd and
xfsprogs from source and we have to install fewer packages.

### DIFF
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -22,7 +22,7 @@ from pathlib import Path
 from types import TracebackType
 from typing import Any, Callable, NoReturn, Optional
 
-from mkosi.log import ARG_DEBUG, die
+from mkosi.log import ARG_DEBUG, ARG_DEBUG_SHELL, die
 from mkosi.types import _FILE, CompletedProcess, PathString, Popen
 from mkosi.util import INVOKING_USER, flock
 
@@ -251,11 +251,14 @@ def run(
     env: Mapping[str, str] = {},
     cwd: Optional[Path] = None,
     log: bool = True,
+    preexec_fn: Optional[Callable[[], None]] = None,
+    sandbox: Sequence[PathString] = (),
 ) -> CompletedProcess:
+    sandbox = [os.fspath(x) for x in sandbox]
     cmdline = [os.fspath(x) for x in cmdline]
 
     if ARG_DEBUG.get():
-        logging.info(f"+ {shlex.join(cmdline)}")
+        logging.info(f"+ {shlex.join(sandbox + cmdline)}")
 
     if not stdout and not stderr:
         # Unless explicit redirection is done, print all subprocess
@@ -281,6 +284,11 @@ def run(
     elif stdin is None:
         stdin = subprocess.DEVNULL
 
+    def preexec() -> None:
+        make_foreground_process()
+        if preexec_fn:
+            preexec_fn()
+
     try:
         # subprocess.run() will use SIGKILL to kill processes when an exception is raised.
         # We'd prefer it to use SIGTERM instead but since this we can't configure which signal
@@ -288,7 +296,7 @@ def run(
         # subprocess.run().
         with sigkill_to_sigterm():
             return subprocess.run(
-                cmdline,
+                sandbox + cmdline,
                 check=check,
                 stdin=stdin,
                 stdout=stdout,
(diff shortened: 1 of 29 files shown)


mkosi/run.py
```diff
@@ -374,11 +390,25 @@ def preexec() -> None:
...
     for name in names:
-        path = ":".join(os.fspath(p) for p in [root / "usr/bin", root / "usr/sbin"]) if root else os.environ["PATH"]
+        if Path(name).is_absolute():
+            name = root / Path(name).relative_to("/")
+        elif "/" in str(name):
+            name = root / name
+
         if (binary := shutil.which(name, path=path)):
-            return Path("/") / Path(binary).relative_to(root or "/")
+            if root != Path("/") and not Path(binary).is_relative_to(root):
```
behrmann, line 408: If the first clause is true, then path is root/usr/bin:root/usr/sbin and which will only look for name there. I don't think they can then not be relative to root or am I missing some edge case in which?

mkosi/run.py
```diff
@@ -304,9 +306,21 @@ def run(
...
+                user=user,
+                group=group,
+                env=env,
+                cwd=cwd,
+                preexec_fn=make_foreground_process,
+            )
         if log:
             log_process_failure(cmdline, e.returncode)
-        raise e
+        raise subprocess.CalledProcessError(e.returncode, cmdline, output=e.output, stderr=e.stderr)
```
behrmann, line 323: What's the reason for constructing a new error here?

behrmann (commented):
Another big one. I love that the try import stuff is going away. A few questions, though.

mkosi/run.py
DaanDeMeyer: @behrmann Absolute symlinks ;)

mkosi/run.py
DaanDeMeyer: To get rid of the sandboxing from the command line so we don't show it to the user by default.

mkosi/run.py
behrmann: How I always forget about the cosmic horror unleashed upon the world by symlinks.

mkosi/run.py
behrmann: Makes sense. We could construct a new horror here and also add a `from e`, this way one would see both stack traces and could scroll up to see the old context. Might be handy for debugging.

DaanDeMeyer: I opted to just change `e.cmd` instead of doing the chaining. I think that leads to a more readable stacktrace and if users want the full command they can run with `--debug`

malt3:
Tested 02b52c7ba82abd7499a0b304b05eb3d370da7ac9 on NixOS. Seems to work fine out of the box.

behrmann (approved):

behrmann:
The new sandboxing looks great, let's see where the new sandbox is too harsh.

PR #2224, opened 2024-01-01
ci: Add systemd-ukify to Debian packages

behrmann:
At least on trixie and sid it's a package of its own now. Let's see whether CI stops complaining with this.


DaanDeMeyer:
This needs to use apt search patterns to only install systemd-ukify if it's actually available

PR #2230, opened 2024-01-02
Preserve target directories stat when copying extra/skeleton trees

DaanDeMeyer:
When copying extra and skeleton trees, let's not touch the permissions
of directories that already exist in the image's root directory. 99%
of the time, the directories are only in the extra tree to make sure
the files go in the right directory in the image's root directory and
serve no other purpose so it makes sense to ignore their metadata in
this case.

Because cp does not support this natively (either all permissions are
copied for directories and files or none are copied), we implement this
ourselves by saving the necessary permissions before we call cp and
restoring them afterwards).

### DIFF
--- a/mkosi/tree.py
+++ b/mkosi/tree.py
@@ -1,8 +1,11 @@
 # SPDX-License-Identifier: LGPL-2.1+
 
+import contextlib
 import errno
 import shutil
 import subprocess
+import tempfile
+from collections.abc import Iterator
 from pathlib import Path
 
 from mkosi.config import ConfigFeature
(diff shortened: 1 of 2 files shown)


mkosi/tree.py
```diff
@@ -45,6 +48,21 @@ def cp_version() -> GenericVersion:
     return GenericVersion(run(["cp", "--version"], stdout=subprocess.PIPE).stdout.strip().splitlines()[0].split()[3])
 
 
+@contextlib.contextmanager
+def preserve_target_directories_stat(src: Path, dst: Path) -> Iterator[None]:
+    dirs = list(p for d in src.glob("**/") if (dst / (p := d.relative_to(src))).exists())
```
behrmann, line 53: Having the `list` call here makes this harder to read due to to the abundance of parens, let's make this a native list comprehension and visually a bit more distinct
```suggestion
    dirs = [p for d in src.glob("**/") if (dst / (p := d.relative_to(src))).exists()]
```

behrmann (approved):
LGTM, one nit

PR #2231, opened 2024-01-03
Don't use smm on aarch64

DaanDeMeyer:
This property is only supported on the x86 machines so stop trying to use it on arm.

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -537,7 +537,7 @@ def run_qemu(args: MkosiArgs, config: MkosiConfig, qemu_device_fds: Mapping[Qemu
 
     machine = f"type={config.architecture.default_qemu_machine()}"
     if firmware == QemuFirmware.uefi:
-        machine += f",smm={'on' if ovmf_supports_sb else 'off'}"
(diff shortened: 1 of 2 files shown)


mkosi/qemu.py
```diff
@@ -537,7 +537,7 @@ def run_qemu(args: MkosiArgs, config: MkosiConfig, qemu_device_fds: Mapping[Qemu
 
     machine = f"type={config.architecture.default_qemu_machine()}"
     if firmware == QemuFirmware.uefi:
-        machine += f",smm={'on' if ovmf_supports_sb else 'off'}"
+        machine += f",smm={'on' if ovmf_supports_sb and config.architecture.supports_smm() else 'off'}"
```
behrmann, line 540: Wouldn't it make sense to move the `config.architecture.supports_smm()` calls into the places where `find_ovmf_firmware` returns `True` for the second argument?

behrmann (approved):
LGTM, but one question about placement.

mkosi/qemu.py
DaanDeMeyer: I don't think so, these are two different concepts

PR #2234, opened 2024-01-03
Rename various symbols

DaanDeMeyer:
- Let's get rid of the Mkosi prefix everywhere. Python has namespaced modules for a reason, let's make use of that.
- Let's also rename State to Context, to match systemd where Context is generally used as well instead of State.

No changes in behavior, just renaming.


behrmann (approved):
I really like the renaming and also specifically the walrus-mirror change in the RPM distros.

PR #2236, opened 2024-01-03

DaanDeMeyer:


mkosi/__init__.py
```diff
@@ -312,7 +312,7 @@ def configure_autologin(context: Context) -> None:
 
 @contextlib.contextmanager
 def mount_cache_overlay(context: Context) -> Iterator[None]:
-    if not context.config.incremental or not any(context.root.iterdir()):
+    if not context.config.incremental or not context.config.base_trees or context.config.overlay:
```
behrmann, line 315: Just to be entirely clear, since I could read the commit message either way: This is supposed to return early in case there already is an overlay (since there is already one, no need for another) or no base tree is configured (since we don't need an overlay), right?

behrmann (approved):
LGTM. The wording of the commit message
> These are only intended for files that affect package manager
operation 

might not quite catch that this includes config a user would put in so that a postinst script becomes a noop, which one might not really read as package manager operation, since it might also be final config, but I agree with the change itself, because
> since if the skeleton tree is changed, users are likely going to want
to throw away their cache regardless.

mkosi/__init__.py
DaanDeMeyer: Exactly

PR #2238, opened 2024-01-03
Add serpentOS and moss support

hunger:
SerpentOS (https://serpentos.com/) is a new Linux distribution that is currently investing heavily into new tooling to build and package software.

It also uses a new package manager called `moss` to quickly install software.

This commit adds support for both serpentos as well as its package manager.

There are a couple of limitations though:

* I did not add tests as building moss needs to be build from rust sources at this time 
* mkosi can not create disk images from inside serpent OS (systemd-repart is too old there)
* You can not create bootable SOS images (systemd-ukify is missing)

The devs are focusing on the tooling right now and did not update the packages in a while.

This PR potentially enables doing tests using mkosi definitions, so I think this PR is valuable even at this time. I think the bit of code that this PR adds will also stay stable: None of the commands used changed in a long time (they were even stable across a rewrite of the entire packager in rust).

### DIFF
--- a/mkosi.conf.d/20-serpentos.conf
+++ b/mkosi.conf.d/20-serpentos.conf
@@ -1,0 +1,15 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+[Match]
+Distribution=serpentos
+
+[Content]
+ShimBootloader=none
+# Override the entire Packages thing:
+# Serpent OS has no attr, autoconf, automake, ca-certificates, ...
+Packages=
+Packages=
+        dash # provides /usr/bin/sh for now...
+        linux-kvm
+        moss
+        systemd
--- a/mkosi/distributions/__init__.py
+++ b/mkosi/distributions/__init__.py
@@ -21,6 +21,7 @@ class PackageType(StrEnum):
     deb    = enum.auto()
     pkg    = enum.auto()
     ebuild = enum.auto()
+    moss   = enum.auto()
 
 
 class DistributionInstaller:
@@ -93,6 +94,7 @@ class Distribution(StrEnum):
     centos       = enum.auto()
     rhel         = enum.auto()
     rhel_ubi     = enum.auto()
+    serpentos    = enum.auto()
     openmandriva = enum.auto()
     rocky        = enum.auto()
     alma         = enum.auto()
--- a/mkosi/distributions/serpentos.py
+++ b/mkosi/distributions/serpentos.py
@@ -1,0 +1,87 @@
+# SPDX-License-Identifier: LGPL-2.1+
+
+from collections.abc import Iterable, Sequence
+
+from mkosi.config import Architecture, Config
+from mkosi.context import Context
+from mkosi.distributions import Distribution, DistributionInstaller, PackageType
+from mkosi.installer import PackageManager
+from mkosi.installer.moss import Moss
+from mkosi.log import die
+
+
+class Installer(DistributionInstaller):
+    @classmethod
+    def pretty_name(cls) -> str:
+        return "iSerpent OS"
+
+    @classmethod
+    def filesystem(cls) -> str:
+        return "btrfs"
+
+    @classmethod
+    def package_type(cls) -> PackageType:
+        return PackageType.moss
+
+    @classmethod
+    def default_release(cls) -> str:
+        return "rolling"
+
+    @classmethod
+    def default_tools_tree_distribution(cls) -> Distribution:
+        return Distribution.serpentos
+
+    @classmethod
+    def package_manager(cls, config: "Config") -> type[PackageManager]:
+        return Moss
+
+    @classmethod
+    def createrepo(cls, context: Context) -> None:
+        pass
+
+    @classmethod
+    def setup(cls, context: Context) -> None:
+        Moss.setup(context, cls.repositories(context))
+
+    @classmethod
+    def sync(cls, context: Context) -> None:
+        pass
+
+    @classmethod
+    def install(cls, context: Context) -> None:
+        # needed for moss to copy the kernel later
+        (context.root / "boot").mkdir(exist_ok=True)
+
+    @classmethod
+    def install_packages(cls, context: Context, packages: Sequence[str], apivfs: bool = True) -> None:
+        Moss.invoke(
+            context,
+            "install",
+            [],
+            packages,
+            apivfs=apivfs,
+        )
+
+    @classmethod
+    def remove_packages(cls, context: Context, packages: Sequence[str]) -> None:
+        Moss.invoke(context, "remove", [], packages)
+
+    @classmethod
+    def repositories(cls, context: Context) -> Iterable[Moss.Repository]:
+        url = "https://dev.serpentos.com/volatile/x86_64/stone.index"
+        if context.config.local_mirror:
+            url = context.config.local_mirror
+
+        yield Moss.Repository("volatile", "Default repository", url, 1000)
+
+    @classmethod
+    def architecture(cls, arch: Architecture) -> str:
+        a = {
+            Architecture.x86_64 : "x86_64",
+        }.get(arch)
+
+        if not a:
+            die(f"Architecture {a} is not supported by {cls.pretty_name()}")
+
+        return a
+
--- a/mkosi/installer/__init__.py
+++ b/mkosi/installer/__init__.py
@@ -102,6 +102,7 @@ def clean_package_manager_metadata(context: Context) -> None:
     for tool, paths in (("rpm",      ["var/lib/rpm", "usr/lib/sysimage/rpm"]),
                         ("dnf5",     ["usr/lib/sysimage/libdnf5"]),
                         ("dpkg",     ["var/lib/dpkg"]),
+                        ("moss",     [".moss"]),
                         (executable, [f"var/lib/{subdir}", f"var/cache/{subdir}"])):
         if always or not find_binary(tool, root=context.root):
             rmtree(*(context.root / p for p in paths if (context.root / p).exists()), sandbox=context.sandbox)
--- a/mkosi/installer/moss.py
+++ b/mkosi/installer/moss.py
@@ -1,0 +1,133 @@
+# SPDX-License-Identifier: LGPL-2.1+
+
+import os
+from collections.abc import Iterable, Sequence
+from pathlib import Path
+from typing import NamedTuple
+
+from mkosi.config import Config
+from mkosi.context import Context
+from mkosi.installer import PackageManager
+from mkosi.mounts import finalize_source_mounts
+from mkosi.run import find_binary, run
+from mkosi.sandbox import apivfs_cmd
+from mkosi.types import PathString
+from mkosi.util import sort_packages
+
+
+def _moss_cache_dir(context: Context) -> Path:
+    moss_cache_dir = context.config.package_cache_dir_or_default() / "moss"
+    return moss_cache_dir
+
+
+class Moss(PackageManager):
+    class Repository(NamedTuple):
+        id: str
+        description: str
+        uri: str
+        priority: int
+
+    @classmethod
+    def executable(cls, config: Config) -> str:
+        # Allow the user to override autodetection with an environment variable
+        moss = config.environment.get("MKOSI_MOSS")
+        root = config.tools()
+
+        return Path(moss or find_binary("moss", root=root) or "moss").name
+
+    @classmethod
+    def subdir(cls, config: Config) -> Path:
+        return Path("moss")
+
+    @classmethod
+    def cache_subdirs(cls, cache: Path) -> list[Path]:
+        return [
+            p / "moss"
+            for p in cache.iterdir()
+            if p.is_dir() and "-" in p.name and "mkosi" not in p.name
+        ]
+
+    @classmethod
+    def scripts(cls, context: Context) -> dict[str, list[PathString]]:
+        return {
+            "moss": apivfs_cmd(context.root) + cls.cmd(context),
+            "mkosi-install": ["moss", "install"],
+            "mkosi-upgrade": ["moss", "sync"],
+            "mkosi-remove": ["moss", "remove"],
+        }
+
+    @classmethod
+    def mounts(cls, context: Context) -> list[PathString]:
+        return [
+            *super().mounts(context),
+            # Moss has all its cache in `,moss/cache`, mount that from the outside
+            "--bind",
+            _moss_cache_dir(context),
+            Path("/.moss/cache"),
+        ]
+
+    @classmethod
+    def setup(cls, context: Context, repositories: Iterable[Repository]) -> None:
+        # Create custom cache dir for moss
+        moss_cache_dir = _moss_cache_dir(context)
+        moss_cache_dir.mkdir(exist_ok=True)
+        (context.root / ".moss").mkdir(exist_ok=True)
+        (context.root / ".moss" / "cache").mkdir(exist_ok=True)
+
+        for repo in repositories:
+            cls.invoke(
+                context,
+                "repo",
+                [
+                    "add",
+                    "-c",
+                    repo.description,
+                    "-p",
+                    f"{repo.priority}",
+                    repo.id,
+                    repo.uri,
+                ],
+            )
+
+    @classmethod
+    def cmd(cls, context: Context) -> list[PathString]:
+        moss = cls.executable(context.config)
+        return [
+            moss,
+            "--directory",
+            context.root,
+            "--yes-all",
+        ]
+
+    @classmethod
+    def invoke(
+        cls,
+        context: Context,
+        operation: str,
+        options: Sequence[str] = (),
+        packages: Sequence[str] = (),
+        apivfs: bool = True,
+    ) -> None:
+        with finalize_source_mounts(
+            context.config,
+            ephemeral=os.getuid() == 0 and context.config.build_sources_ephemeral,
+        ) as sources:
+            run(
+                cls.cmd(context) + [operation, *options, *sort_packages(packages)],
+                sandbox=(
+                    context.sandbox(
+                        network=True,
+                        options=[
+                            "--bind",
+                            context.root,
+                            context.root,
+                            *cls.mounts(context),
+                            *sources,
+                            "--chdir",
+                            "/work/src",
+                        ],
+                    )
+                    + (apivfs_cmd(context.root) if apivfs else [])
+                ),
+                env=context.config.environment,
+            )
--- a/mkosi/resources/mkosi-initrd/mkosi.conf.d/10-serpentos.conf
+++ b/mkosi/resources/mkosi-initrd/mkosi.conf.d/10-serpentos.conf
@@ -1,0 +1,35 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+[Match]
+Distribution=serpentos
+
+[Content]
+Packages=
+        dash # for /bin/sh
+        # Various libraries that are dlopen'ed by systemd
+
+        # File system checkers for supported root file systems
+        e2fsprogs
+        xfsprogs
+
+        # fsck.btrfs is a dummy, checking is done in the kernel.
+
+        util-linux
+
+RemoveFiles=
+        /.moss
+
+        # Remove all files that are only required for development.
+        /usr/lib/*.a # static libs
+        /usr/lib/gconv # encoding conversion
+        /usr/lib/kernel # kernel-install files
+
+        /usr/share/defaults/*
+        # Remove docs
+        /usr/share/doc/*
+        /usr/share/i18n/*
+        /usr/share/info/*
+        /usr/share/man/*
+
+        # Remove locale config
+        /usr/share/locale/*
--- a/mkosi/resources/mkosi-initrd/mkosi.conf.d/20-stub.conf
+++ b/mkosi/resources/mkosi-initrd/mkosi.conf.d/20-stub.conf
@@ -3,6 +3,7 @@
 [Match]
 Format=uki
 Distribution=!arch
+Distribution=!serpentos
 
 [Content]
 Packages=systemd-boot


DaanDeMeyer:
Yeah don't worry about those I'm fighting them as we speak

DaanDeMeyer:
So playing around with this, first thing that seems to be missing is /bin/sh. Which package provides that in serpent os? I can install bash but it doesn't provide /bin/sh. This is important for mkosi's scripts as they tend to use /bin/sh in their shebang.

mkosi/installer/moss.py
```diff
@@ -0,0 +1,41 @@
...
+
+def moss_cmd(context: Context) -> list[PathString]:
+    cmdline: list[PathString] = [
+        "moss",
+        "--directory",
+        f"{context.root}",
+        "--yes-all",
+    ]
+
+    return cmdline
```
behrmann, line 19: No need for the extra variable and annotation
```suggestion
    return [
        "moss",
        "--directory",
        f"{context.root}",
        "--yes-all",
    ]
```

The same style is used e.g. in pacman.py

mkosi/installer/moss.py
```diff
@@ -0,0 +1,41 @@
...
+
+def setup_moss(
+    context: Context,
+) -> None:
+    (context.pkgmngr / "etc/moss/repo.d").mkdir(exist_ok=True, parents=True)
+    bwrap(context,
+        moss_cmd(context)
+        + ["repo", "add", "volatile", "https://dev.serpentos.com/volatile/x86_64/stone.index"],
+        network=True, env=context.config.environment,
+    )
```
behrmann, line 30:
```suggestion
    bwrap(
        context,
        moss_cmd(context) + ["repo", "add", "volatile", "https://dev.serpentos.com/volatile/x86_64/stone.index"],
        network=True, 
        env=context.config.environment,
    )
```
the hardcoded url should probably be configurable via the mirror, which is usually done in the distribution module.

mkosi/installer/moss.py
```diff
@@ -0,0 +1,41 @@
...
+
```
behrmann, line 31: missing newline
```suggestion


```

mkosi/installer/moss.py
```diff
@@ -0,0 +1,41 @@
...
+def invoke_moss(
+    context: Context,
+    operation: str,
+    packages: Sequence[str] = (),
+    apivfs: bool = False,
+) -> None:
+    cmd = apivfs_cmd(context.root) if apivfs else []
+    bwrap(context, cmd + moss_cmd(context) + [operation] + [*sort_packages(packages)],
+        network=True, env=context.config.environment,
+    )
```
behrmann, line 41: formatting and no need to unpack the list of packages and pack it up again.
```suggestion
    bwrap(
        context, 
        cmd + moss_cmd(context) + [operation] + sort_packages(packages),
        network=True, 
        env=context.config.environment,
    )
```

mkosi/installer/moss.py
behrmann: It might also make sense for moss_cmd to take a command argument, see apt_cmd, which would make some of the list additions further down unnecessary.

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,67 @@
...
+    @classmethod
+    def default_tools_tree_distribution(cls) -> Distribution:
+        return Distribution.serpentos
+
+    @classmethod
+    def tools_tree_packages(cls) -> list[str]:
+        return [
+            "systemd",
+            "zstd",
+        ]
```
behrmann, line 38: tools_tree_packages lives in Config. Is your expectation that these would be the default tools tree packages or additional ones?

mkosi/distributions/__init__.py
```diff
@@ -106,6 +108,9 @@ def is_dnf_distribution(self) -> bool:
     def is_apt_distribution(self) -> bool:
         return self in (Distribution.debian, Distribution.ubuntu)
 
+    def is_moss_distribution(self) -> bool:
+        return self == Distribution.serpentos
```
DaanDeMeyer, line 112: This isn't used anywhere so can be dropped

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,67 @@
...
+from mkosi.context import Context
+from mkosi.distributions import Distribution, DistributionInstaller, PackageType
+from mkosi.installer.moss import invoke_moss, setup_moss
+from mkosi.log import die
+
+
+class Installer(DistributionInstaller):
+    @classmethod
+    def pretty_name(cls) -> str:
+        return "serpent os"
```
DaanDeMeyer, line 15: This should be "Serpent OS" I guess

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,67 @@
...
+            "zstd",
+        ]
+
+    @classmethod
+    def setup(cls, context: Context) -> None:
+        (context.root / "etc").mkdir(exist_ok=True, parents=True)
+
+    @classmethod
+    def install(cls, context: "Context") -> None:
+        setup_moss(context)
```
DaanDeMeyer, line 46: This should be called in `setup()`

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,67 @@
...
+    @classmethod
+    def tools_tree_packages(cls) -> list[str]:
+        return [
+            "systemd",
+            "zstd",
+        ]
+
+    @classmethod
+    def setup(cls, context: Context) -> None:
+        (context.root / "etc").mkdir(exist_ok=True, parents=True)
```
DaanDeMeyer, line 42: Some clarification on why this is required would be nice

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,67 @@
...
+
+    @classmethod
+    def install(cls, context: "Context") -> None:
+        setup_moss(context)
+        (context.root / "boot").mkdir(exist_ok=True)
```
DaanDeMeyer, line 47: Also here, please add some clarification on why this is required.

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,67 @@
...
+
+    @classmethod
+    def install_packages(cls, context: Context, packages: Sequence[str], _apivfs: bool = True) -> None:
+        invoke_moss(context, "install", packages)
```
DaanDeMeyer, line 51: `apivfs` is not passed through here. Also please name it `apivfs` like everywhere else.

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,67 @@
...
+        invoke_moss(context, "remove", packages)
+
+    @classmethod
+    def architecture(cls, arch: Architecture) -> str:
+        a = {
+            Architecture.x86_64 : "x86_64",
+        }.get(arch)
+
+        if not a:
+            die(f"Architecture {a} is not supported by Serpent OS")
```
DaanDeMeyer, line 64: Please use the pretty_name() helper here

mkosi/installer/__init__.py
```diff
@@ -30,7 +31,8 @@ def clean_package_manager_metadata(context: Context) -> None:
     for tool, paths in (("rpm",    ["var/lib/rpm", "usr/lib/sysimage/rpm"]),
                         ("dnf5",   ["usr/lib/sysimage/libdnf5"]),
                         ("dpkg",   ["var/lib/dpkg"]),
-                        ("pacman", ["var/lib/pacman"])):
+                        ("pacman", ["var/lib/pacman"]),
+                        ("moss",   [".moss", "etc/moss"])):
```
DaanDeMeyer, line 35: Is `.moss` located in the root directory?

mkosi/installer/moss.py
```diff
@@ -0,0 +1,41 @@
...
+from mkosi.context import Context
+from mkosi.types import PathString
+from mkosi.util import sort_packages
+
+
+def moss_cmd(context: Context) -> list[PathString]:
+    cmdline: list[PathString] = [
+        "moss",
+        "--directory",
+        f"{context.root}",
```
DaanDeMeyer, line 15: No point in using a format string here

mkosi/installer/moss.py
```diff
@@ -0,0 +1,41 @@
...
+        "--yes-all",
+    ]
+
+    return cmdline
+
+
+def setup_moss(
+    context: Context,
+) -> None:
+    (context.pkgmngr / "etc/moss/repo.d").mkdir(exist_ok=True, parents=True)
```
DaanDeMeyer, line 25: You create `etc/moss/repo.d` in the `pkgmngr` directory but never configure `moss` to actually look for repos in that directory.

DaanDeMeyer (changes requested):
`moss` should really have an option to look for repositories and configuration in general in a directory outside of the root directory. All the other package managers we support have this and it allows us to keep the image itself clean of package manager specific details.

I'm also missing an option to specify a cache directory for `moss` to use so we can have `moss` use mkosi's cache directory.

I'm also curious what package signing looks like on Serpent OS. Is any signing done at all? If so, where is the keyring located that `moss` should use?

mkosi/installer/moss.py
DaanDeMeyer: We shouldn't use the `moss` command to add repositories. Instead, we should write the file ourselves to the right location (I hope a human readable format is used).

mkosi/installer/moss.py
DaanDeMeyer: Does `moss` look in the host's root directory or in the image's root directory for its configuration? If it's in the host's root directory then this is fine.

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,72 @@
...
+    def default_tools_tree_distribution(cls) -> Distribution:
+        return Distribution.serpentos
+
+    @classmethod
+    def tools_tree_packages(cls) -> list[str]:
+        return [
+            "dash", # for sh
+            "systemd",
+            "zstd",
+        ]
```
DaanDeMeyer, line 40: This should just be dropped as this method's been removed in git main

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @classmethod
+    def setup(cls, context: Context) -> None:
+        if not shutil.which("moss"):
+            die(f"moss binary not found but needed to install {cls.pretty_name()}")
```
DaanDeMeyer, line 45: No need for this since we don't do it for the other distributions either. We'll add proper checks at some point but we'll do it for all distributions at once

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,72 @@
...
+
+        # needed to normalize modification time of /boot later:
+        (context.root / "boot").mkdir(exist_ok=True)
+
+    @classmethod
+    def install(cls, context: "Context") -> None:
+        pass
```
DaanDeMeyer, line 52: So in `install()` we have to install the basic filesystem packages. Please check the other distributions for examples. Does Serpent OS have a package that provides basic directories such as /usr and /etc? We should also install the package that provides /etc/passwd and friends here.

mkosi/installer/moss.py
```diff
@@ -0,0 +1,29 @@
...
+from mkosi.util import sort_packages
+
+
+def moss_cmd(context: Context) -> list[PathString]:
+    return [
+        "moss",
+        "--directory",
+        context.root,
+        "--yes-all",
+    ]
```
DaanDeMeyer, line 17: This can be a single line

mkosi/installer/moss.py
```diff
@@ -0,0 +1,29 @@
...
+
```
DaanDeMeyer, line 18: Two newlines between functions please

DaanDeMeyer (changes requested):
Please make sure `mkosi -d serpentos -f qemu` works when executed from the mkosi repository.

mkosi/distributions/serpentos.py
DaanDeMeyer: @hunger Right but when does it run systemd-tmpfiles and such to populate the rest of the system? Does it intend to run these all at firstboot or are these run by the package manager?

DaanDeMeyer:
@hunger We should not use the SerpentOS initrd when building mkosi images. Our current policy is that if no initrd is provided by the user we build one ourselves. So you'd have to add definitions for SerpentOS to mkosi/resources/mkosi-initrd so we can build SerpentOS initrds.

DaanDeMeyer:
@hunger From what I read on their blog, they weren't even planning on using systemd-boot or UKIs though, so mkosi will produce something significantly different from the official image either way. At some point we'll add support for picking up signed prebuilt UKIs likely, though I don't really see the point in picking up prebuilt initrds.

mkosi/__init__.py
```diff
@@ -1790,7 +1790,6 @@ def install_uki(context: Context, partitions: Sequence[Partition]) -> None:
         microcode = build_microcode_initrd(context)
 
         initrds = [microcode] if microcode else []
-        initrds += context.config.initrds or [build_initrd(context)]
```
DaanDeMeyer, line 1793: This definitely can't be dropped.

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,112 @@
...
+                continue
+
+            kkind = match.group(1)
+            kver = match.group(2)
+
+            modules_dir = (context.root / f"usr/lib/modules/{kver}.{kkind}")
+            if modules_dir.is_dir():
+                shutil.move(k, (modules_dir / f"vmlinuz-{kver}.{kkind}"))
+                shutil.move((kernel_dir / f"initrd-com.serpentos.{kkind}.{kver}"),
+                            (modules_dir / f"initrd-{kver}.{kkind}"))
```
DaanDeMeyer, line 55: No need to move the initrd since we won't do anything with it.

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,112 @@
...
+    @classmethod
+    def default_release(cls) -> str:
+        return "rolling"
+
+    @classmethod
+    def default_tools_tree_distribution(cls) -> Distribution:
+        return Distribution.serpentos
+
+    @classmethod
+    def _move_kernel_and_initrd(cls, context: Context) -> None:
```
DaanDeMeyer, line 38: I don't think we really need regexp for this. Please just do this similarly to how we do it for Debian:

```
for d in context.root.glob("boot/vmlinuz-*"):
    kver = d.name.removeprefix("vmlinuz-")
    vmlinuz = context.root / "usr/lib/modules" / kver / "vmlinuz"
    if not vmlinuz.exists():
        shutil.copy2(d, vmlinuz)
```

mkosi/distributions/serpentos.py
DaanDeMeyer: @hunger Still need an answer to my question here.

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,112 @@
...
+            (repo_dir / "volatile.yaml").write_text(
+                textwrap.dedent(
+                    """\
+                    volatile:
+                      description: "Default repository"
+                      uri: https://dev.serpentos.com/volatile/x86_64/stone.index
+                      priority: 1000
+                    """
+                )
+            )
```
DaanDeMeyer, line 84: I'm confused. Where exactly does `moss` look for its configuration? In /etc/moss of the host system or in /etc/moss of the root directory it's operating on?

mkosi/installer/moss.py
```diff
@@ -0,0 +1,40 @@
...
+from mkosi.context import Context
+from mkosi.mounts import finalize_ephemeral_source_mounts
+from mkosi.run import run
+from mkosi.sandbox import apivfs_cmd, finalize_crypto_mounts
+from mkosi.types import PathString
+from mkosi.util import sort_packages
+
+
+def moss_cmd(context: Context) -> list[PathString]:
+    return [ "moss", "--directory", context.root, "--yes-all", ]
```
DaanDeMeyer, line 14: Please remove the trailing comma

DaanDeMeyer (changes requested):
I don't see how this makes mkosi handle the initrd? You have to modify mkosi/resources/mkosi-initrd to add definitions to handle serpentOS.

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,112 @@
...
+                # Remove now dead symlinks to current kernel/initrd
+                for f in kernel_dir.glob("current.*.kernel"):
+                    os.remove(f)
+                for f in kernel_dir.glob("current.*.initrd"):
+                    os.remove(f)
+
+    @classmethod
+    def setup(cls, context: Context) -> None:
+        # needed to normalize modification time of /boot later:
+        (context.root / "boot").mkdir(exist_ok=True)
```
DaanDeMeyer, line 70: This can be dropped after https://github.com/systemd/mkosi/pull/2284 is merged

ikeycode:
Hi @DaanDeMeyer - Ikey here, Serpent OS lead. I've just been made aware of this PR, so I thought I'd answer some of your queries.

- moss-rs doesn't *yet* take non `rootfs` configurations but it is planned (it uses the same usr/etc layering approach of systemd). It operates on the contextual root directory (`-D`)
- package signing will be introduced in a v2 bump using multisig (See https://github.com/serpent-os/moss-rs/issues/50)
- We plan to use systemd-boot (+shim) - we ship prebuilt initrds and the `dracut` package. Our `linux-kvm` package shows a simple example of creating an initrd, it's just not something we'd support in an *installed* system.
 - For now, we force `/bin/sh` via `dash` until an alternatives system is implemented. Note also we enforce `usr-merge`, so: `moss install 'binary(sh)'` will yield the correct candidate.
 - The team and I have mentioned to @hunger we do have our own plans for container builds, and can't ourselves throw support into this effort right now (volatility of the system internals)
 - We do currently issue nightly builds of `moss-rs` for testing, if its useful we could work out some signed releases on a specific branch.
 
 Like said, while we're not formally supporting this effort we're not going to block progress or innovation. Any questions , don't hesitate.

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,112 @@
...
+
+        repo_dir = (context.root / "etc/moss/repo.d")
+        if not repo_dir.exists():
+            repo_dir.mkdir(exist_ok=True, parents=True)
```
DaanDeMeyer, line 74: No point in checking if this exists if we're going to create it unconditionally. You should check instead if `volatime.yaml` already exists and only write it if it doesn't already exist.

DaanDeMeyer:
@ikeycode 

> moss-rs doesn't yet take non rootfs configurations but it is planned (it uses the same usr/etc layering approach of systemd). It operates on the contextual root directory (-D)

Where does it look for its configuration though? In the rootfs or outside of it?

> We do currently issue nightly builds of moss-rs for testing, if its useful we could work out some signed releases on a specific branch.

We can just build it from source probably, no big deal

> Like said, while we're not formally supporting this effort we're not going to block progress or innovation. Any questions , don't hesitate.

Yeah we don't really care about being anything official, as long as the package manager makes it easy enough and the whole thing is relatively stable, we can probably support it.

ikeycode:
@DaanDeMeyer right now it looks within the rootfs, though the plan is to support all required configs via CLI/clap.

mkosi/distributions/serpentos.py
DaanDeMeyer: To clarify, having to move the files at all is a hack. Having kernel related files in /usr/lib/modules is practically Linux API for all intents and purposes and SerpentOS should really be installing its kernels and such in there by default. Doing so will save them a lot of pain getting all kinds of tools working that will look for this stuff in /usr/lib/modules.

But sure, feel free to move all files if you care about this.

mkosi/distributions/serpentos.py
DaanDeMeyer: Right just leave it empty for now we can figure this out later.

mkosi/distributions/serpentos.py
DaanDeMeyer: This makes it harder to extend the "volatile" repo with other repos when that becomes possible. Please name the file `mkosi.yml` so that users can override the default repositories by simply creating an empty file `mkosi.yml` in their package manager tree.

mkosi/resources/mkosi-initrd/mkosi.conf.d/10-serpentos.conf
```diff
@@ -0,0 +1,38 @@
...
+RemoveFiles=
+        /.moss
+
+        # Remove all files that are only required for development.
+        /usr/bin/mkfs*
+        /usr/bin/mke2fs
+        /usr/lib/*.a
+        /usr/lib/kernel/*
+
+        /usr/share/bash-completion/*
```
DaanDeMeyer, line 31: How much does this actually save? We don't remove bash completions on any other distributions either

DaanDeMeyer: In that case please keep these in there for now.

> The resulting initrd is still 2.5x the original SOS one.

That's probably because we include all kernel modules by default which SOS might not do. The rest if probably simply due to moss not doing sufficient subpackages, causing lots of unnecessary stuff to be pulled into the initrd. This is something to solve within SOS itself.

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,110 @@
...
+                      uri: https://dev.serpentos.com/volatile/x86_64/stone.index
+                      priority: 1000
+                    """
+                )
+            )
+
+    @classmethod
+    def install(cls, context: Context) -> None:
+        # needed to copy files into later
+        (context.root / "boot").mkdir(exist_ok=True)
```
DaanDeMeyer, line 88: Where exactly do we fail if this is not added? Can you give me the full exception stacktrace?

mkosi/installer/moss.py
```diff
@@ -0,0 +1,40 @@
...
+        run(
+            moss_cmd(context) + [operation, *sort_packages(packages)],
+            sandbox=(
+                context.sandbox(
+                    network=True,
+                    options=[
+                        "--bind", context.root, context.root,
+                        *finalize_crypto_mounts(tools=context.config.tools()),
+                        *sources,
+                        "--chdir", "/work/src"
```
DaanDeMeyer, line 33: Missing trailing comma

mkosi.conf.d/20-serpentos.conf
```diff
@@ -0,0 +1,13 @@
...
+
+[Match]
+Distribution=serpentos
+
+[Content]
+ShimBootloader=none
+Packages=
+Packages=
+        systemd
+        dash
```
DaanDeMeyer, line 11: dash is useful for scripting but pretty useless as an emergency login shell so we should install bash as well

mkosi.conf.d/20-serpentos.conf
```diff
@@ -0,0 +1,13 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+[Match]
+Distribution=serpentos
+
+[Content]
+ShimBootloader=none
+Packages=
```
DaanDeMeyer, line 8: Instead of overriding the package list we should move the packages that aren't available in SerpentOS to the distribution specific configuration files.

Also, what about `less`, I'd be very surprised if that isn't packaged for SerpentOS already.

mkosi/resources/mkosi-initrd/mkosi.conf.d/10-serpentos.conf
```diff
@@ -0,0 +1,39 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+[Match]
+Distribution=serpentos
+
+[Content]
+# Completely override the default Packages
+Packages=
```
DaanDeMeyer, line 8: We shouldn't override the list of packages but instead move packages that aren't available on SerpentOS to the distribution specific configuration files

mkosi/resources/mkosi-initrd/mkosi.conf.d/10-serpentos.conf
```diff
@@ -0,0 +1,39 @@
...
+        dash
+        # Various libraries that are dlopen'ed by systemd
+
+        # File system checkers for supported root file systems
+        e2fsprogs
+        xfsprogs
+
+        # fsck.btrfs is a dummy, checking is done in the kernel.
+
+        util-linux
```
DaanDeMeyer, line 20: bash? less? `dash` is not a useful emergency login shell so we should definitely make sure `bash` is available.

DaanDeMeyer: @hunger I'm open to having conversations about all of these decisions, but please understand that we're not going to special case SerpentOS. We want to provide a consistent experience across distributions. We're not going to make exceptions for just SerpentOS. So if you want to see this changed, please do so in separate PRs and do so for all supported distributions

DaanDeMeyer (changes requested):
Which cache directory does `moss` use? Can you add that to `finalize_package_manager_mounts()`?

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,110 @@
...
+    def default_release(cls) -> str:
+        return "rolling"
+
+    @classmethod
+    def default_tools_tree_distribution(cls) -> Distribution:
+        return Distribution.serpentos
+
+    @classmethod
+    def _move_kernel_and_initrd(cls, context: Context) -> None:
+        regexp = re.compile("com\.serpentos\.([a-z][a-z0-9]+)\.(.*)")
```
keszybz, line 39: This isn't actually doing what it seems to be doing.
The one layer of escaping is "consumed" by python itself, so the actual regexp string that is used is
`com.serpentos.([a-z][a-z0-9]+).(.*)`. Of course it works, because `.` also matches `.`.
A good general rule is to always use r-strings for regexps.

Python `re` does regexp caching internally. So explicit calls to `re.compile` are generally not needed.

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,110 @@
...
+
+        # SerpentOS puts its kernels and initrds into /usr/lib/kernel
+        kernel_dir = context.root / "usr/lib/kernel"
+        for k in kernel_dir.glob("com.serpentos.*"):
+            match = regexp.fullmatch(str(k.name))
+            if not match:
```
keszybz, line 45: `if not (match := …)`

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,110 @@
...
+                continue
+
+            kkind = match.group(1)
+            kver = match.group(2)
```
keszybz, line 49: kkind, kver = match.groups()

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,110 @@
...
+
+            modules_dir = (context.root / f"usr/lib/modules/{kver}.{kkind}")
```
keszybz, line 51: Spurious parens.

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,110 @@
...
+            if modules_dir.is_dir():
+                shutil.move(k, (modules_dir / f"vmlinuz-{kver}.{kkind}"))
+                shutil.move((kernel_dir / f"initrd-com.serpentos.{kkind}.{kver}"),
+                            (modules_dir / f"initrd-{kver}.{kkind}"))
```
keszybz, line 55: Also here.

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,110 @@
...
+
+                # Move the config, etc. while at it anyway:
+                for f in kernel_dir.glob("*{kver}.{kkind}"):
+                    shutil.move(f, (modules_dir / f.name))
```
keszybz, line 59: ... and here.

mkosi/resources/mkosi-initrd/mkosi.conf.d/10-serpentos.conf
```diff
@@ -0,0 +1,37 @@
...
+
+RemoveFiles=
+        /.moss
+
+        # Remove all files that are only required for development.
+        /usr/lib/*.a # static libs
+        /usr/lib/gconv # encoding conversion
+        /usr/lib/kernel # kernelinstall files
```
keszybz, line 26: There is `installkernel` and `kernel-install`, but no `kernelinstall` that I know of. Typo?

keszybz (commented):
This doesn't look too bad. If it works, we should definitely merge it.

keszybz:
> * For now, we force `/bin/sh` via `dash` until an alternatives system is implemented. Note also we enforce `usr-merge`, so: `moss install 'binary(sh)'` will yield the correct candidate.

If this was already discussed above in one of the collapsed comment, apologies. Github doesn't make it easy to search the whole conversation.

Debian tried to switch to `dash-as-/bin/sh`, and it was a lengthy and painful failure. There are bazillions of scripts which just use `#!/bin/sh`, but actually depend on bash features or run incorrectly with dash. So implicitly using dash is a constant fight. On the other hand, dash may be a bit smaller and faster, but the differences are insignificant, and any non-toy installation will require bash anyway, so the highest possible benefits are tiny. This is just a waste of time in the long run.

ikeycode:
> Debian tried to switch to `dash-as-/bin/sh`, and it was a lengthy and painful failure. There are bazillions of scripts which just use `#!/bin/sh`, but actually depend on bash features or run incorrectly with dash. So implicitly using dash is a constant fight. On the other hand, dash may be a bit smaller and faster, but the differences are insignificant, and any non-toy installation will require bash anyway, so the highest possible benefits are tiny. This is just a waste of time in the long run.

I'm inclined to agree fwiw. Sure, CONF_SHELL can be overriden, and tbh we never pursued it for POSIX it was entirely to speed up package builds. It does do that, but it is an awful pain in the arse. Even using `dash` for the recipe scripts makes life harder, ie brace globs, single vs double square bracket, etc.

I've largely been waiting for someone to verbalise it better than myself as an excuse to nuke it from orbit. The maintenance story around `dash` isn't fantastic, and I'm aware of a few situations where it segfaults.

Conan-Kudo:
Please write this up somewhere so that I can take it back to the openSUSE guys who recently decided busybox-sh should be `/bin/sh` sometimes. 😦

bluca:
> > * For now, we force `/bin/sh` via `dash` until an alternatives system is implemented. Note also we enforce `usr-merge`, so: `moss install 'binary(sh)'` will yield the correct candidate.
> 
> If this was already discussed above in one of the collapsed comment, apologies. Github doesn't make it easy to search the whole conversation.
> 
> Debian tried to switch to `dash-as-/bin/sh`, and it was a lengthy and painful failure. There are bazillions of scripts which just use `#!/bin/sh`, but actually depend on bash features or run incorrectly with dash. So implicitly using dash is a constant fight. On the other hand, dash may be a bit smaller and faster, but the differences are insignificant, and any non-toy installation will require bash anyway, so the highest possible benefits are tiny. This is just a waste of time in the long run.

Yeah unless you are absolutely strapped for resources and counting milliseconds, don't bother with dash. Also I recommend staying away from 'alternatives' systems - they go behind the back of packages and make a mess of things, and don't really work for read-only rootfs images.

DaanDeMeyer:
@ikeycode Any chance you could put your cache/state directory somewhere in /var (/var/cache or /var/lib) as well? Currently moss is the odd one out by using a new top level directory instead of the standard FHS ones.

ikeycode:
> @ikeycode Any chance you could put your cache/state directory somewhere in /var (/var/cache or /var/lib) as well? Currently moss is the odd one out by using a new top level directory instead of the standard FHS ones.

Unfortunately I can't see a sane way to make this possible. Apart from relying on hardlinks for each produced transaction, we can't afford the penalty (or associated pains) with dealing with cross-device link scenarios. Even if we took a `linkat_or_copyat` approach, we'd still then have the issue of promoting transactions to `/usr` (achieved using `renameat2` w/ ATOMIC_EXCHANGE)

Long story short if anyone used `/var` as a mountpoint, it'd cause significant issues in our model. Open to suggestions but I don't see an easy way around it. In many ways `.moss` is the real OS store (and moss dbs/etc) with some dynamically generated links for usr-merge, and a dynamic `/usr` pivoted from within `/.moss` (cycling between staging transaction and archived transactions for rollback)

Conan-Kudo:
That sounds like moss shares a lot more with rpm-ostree than regular package managers...

ikeycode:
> That sounds like moss shares a lot more with rpm-ostree than regular package managers...

Likely it aims for the same outcome, but achieves it in a much different way while focusing on composition and granularity. It also uses containers for triggers, with a fully functioning build system. It doesn't suffer the performance
impact of ostree nor does it require `overlayfs` (in fact, without the whiteout options it *breaks* overlayfs*). It also has a read-optimised binary format (`.stone`) which is designed for skip-iteration and zstd (while not suffering some design flaws like encoding names in the archive header, for example). A well defined container of versioned containers with strongly typed + tagged keys, and deduplicated content addressable storage at the heart.

It allows promotion and composition without mandated reboots or virtual trees, though at some point we will introduce deferred activation of new `/usr` under certain conditions (such as kernel and libc). Internally its a strange beast, but the end user feels like they're using a normal package manager (Despite forced stateless by only allowing `/usr` in packages)

DaanDeMeyer:
@hunger Any update on this?

DaanDeMeyer:
> I do have an update to new mkosi syntax here, but I can not get the entire thing to boot in qemu right now, so I think it is a bit pointless to push at this time. I am not sure what is causing this: The initrd fails to mount the root FS, but I can manually mount it just fine after logging into the initrd rescue shell.

Are you building a disk image or trying to boot a directory image? Booting a directory image in qemu will only work properly with systemd v255 or newer if CONFIG_VIRTIOFS is not a builtin module in the kernel.

mkosi/installer/moss.py
```diff
@@ -0,0 +1,133 @@
...
+                [
+                    "add",
+                    "-c",
+                    repo.description,
+                    "-p",
+                    f"{repo.priority}",
+                    repo.id,
+                    repo.uri,
+                ],
+            )
```
DaanDeMeyer, line 90: Do we have to do this via the cli? I would prefer if we write the file ourselves as we do it like that for every other distribution as well.

mkosi/installer/moss.py
```diff
@@ -0,0 +1,133 @@
...
+
+    @classmethod
+    def cmd(cls, context: Context) -> list[PathString]:
+        moss = cls.executable(context.config)
+        return [
+            moss,
+            "--directory",
+            context.root,
+            "--yes-all",
+        ]
```
DaanDeMeyer, line 100: Please simplify this to `return ["moss", ...]`

mkosi/installer/moss.py
```diff
@@ -0,0 +1,133 @@
...
+                        network=True,
+                        options=[
+                            "--bind",
+                            context.root,
+                            context.root,
+                            *cls.mounts(context),
+                            *sources,
+                            "--chdir",
+                            "/work/src",
+                        ],
```
DaanDeMeyer, line 128: Please put related stuff on the same line in the same way that it's done for the other package manager implementations

mkosi/installer/moss.py
```diff
@@ -0,0 +1,133 @@
...
+                    )
+                    + (apivfs_cmd(context.root) if apivfs else [])
```
DaanDeMeyer, line 130: I expect that you just copy pasted this from another implementation, but it'd be great if we could figure out whether `moss` needs the APIVFS filesystems mounted at all when installing packages. I figure since it's trying to be modern it might be able to operate without these mounted.

mkosi/installer/moss.py
```diff
@@ -0,0 +1,133 @@
...
+
+    @classmethod
+    def invoke(
+        cls,
+        context: Context,
+        operation: str,
+        options: Sequence[str] = (),
+        packages: Sequence[str] = (),
+        apivfs: bool = True,
+    ) -> None:
```
DaanDeMeyer, line 110: Please update the arguments to match the other `invoke` methods

mkosi/distributions/serpentos.py
```diff
@@ -0,0 +1,87 @@
...
+        pass
+
+    @classmethod
+    def setup(cls, context: Context) -> None:
+        Moss.setup(context, cls.repositories(context))
+
+    @classmethod
+    def sync(cls, context: Context) -> None:
+        pass
+
```
DaanDeMeyer, line 49: I think moss supports this via `moss repo update`

DaanDeMeyer (changes requested):

DaanDeMeyer:
> The big question is: Can I just run moss without any sandboxing going on? Moss does set up user namespaces and such itself and that fails now when I try to run it from inside mkosi. I guess that is not how you intend mkosi to be used ;-)

@hunger The sandboxing should be made optional in `moss` so we can disable it in mkosi.

DaanDeMeyer:
@hunger We don't use sandboxing for security purposes, we use it for other stuff. For example replacing /usr so we can use tools from a different distribution to build the image, or to replace /etc so the user can provide their own package manager configuration files etc.

`moss` can do its own sandboxing, but it should make sure it still works from within an already sandboxed environment.

DaanDeMeyer:
@hunger Any update on this?

PR #2243, opened 2024-01-04
Copy over kernel images from /boot for opensuse as well

DaanDeMeyer:
On Tumbleweed they're already in /usr/lib/modules but on older releases they aren't so copy them over like we do for debian.


behrmann:
Looking at #2244 the CI failures should be unrelated. Merging.

PR #2245, opened 2024-01-04
Fix importlib usage

DaanDeMeyer:
We have to use as_file() on the final path, not the module path.


behrmann (approved):
As discussed, LGTM. It irks me greatly I didn't notice this earlier.

behrmann:
One nit: the commit message could mention zipapps, because that's the context where this broke.

behrmann (approved):

PR #2246, opened 2024-01-05

DaanDeMeyer:


mkosi/tree.py
```diff
@@ -76,7 +76,7 @@ def copy_tree(
         "cp",
         "--recursive",
         "--dereference" if dereference else "--no-dereference",
-        f"--preserve=mode,timestamps,links,xattr{',ownership' if preserve_owner else ''}",
+        f"--preserve=mode,timestamps,links{',ownership,xattr' if preserve else ''}",
```
behrmann, line 79: The xattr change should probably also be mentioned in the NEWS.

mkosi/qemu.py
```diff
@@ -702,12 +709,28 @@ def run_qemu(args: Args, config: Config, qemu_device_fds: Mapping[QemuDeviceNode
...
             target = Path("/root/src") / (tree.target or tree.source.name)
-            kcl += [f"systemd.mount-extra={sock.name}:{target}:virtiofs"]
+            kcl += [f"systemd.mount-extra={tag}:{target}:virtiofs"]
+
+        if want_scratch(config) or config.output_format in (OutputFormat.disk, OutputFormat.esp):
+            cmdline += ["-device", "virtio-scsi-pci,id=scsi"]
+
+        if want_scratch(config):
+            scratch = stack.enter_context(tempfile.NamedTemporaryFile(dir="/var/tmp", prefix="mkosi-scratch"))
+            scratch.truncate(1024**4)
```
behrmann, line 725: Thats a terabyte and while I think truncating the file to that size in tmpfs will not allocate a file this large, is this guaranteed for all FS or could this go haywire in some edge case and fill someone's space?

behrmann (approved):
LGTM, two nits.

mkosi/qemu.py
DaanDeMeyer: We do the same thing in systemd-repart and nobody's complained yet, so I think it should be fine.

mkosi/tree.py
DaanDeMeyer: I think we can probably get away without mentioning this. I'm 99% sure nobody relies on this.

PR #2252, opened 2024-01-08
ci: Enable Debian builds with tools trees again

DaanDeMeyer:
The required fix in systemd is now in Debian so let's try enabling
these again.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1292,7 +1292,11 @@ def prepare_grub_bios(context: Context, partitions: Sequence[Partition]) -> None
                 "search_fs_file",
             ],
             sandbox=context.sandbox(
-                options=["--ro-bind", context.root / "usr", "/usr", "--bind", context.root, context.root],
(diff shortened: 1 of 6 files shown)


behrmann (approved):

mkosi/__init__.py
```diff
@@ -164,7 +164,6 @@ def install_build_packages(context: Context) -> None:
     if not context.config.build_scripts or not context.config.build_packages:
         return
 
-    # TODO: move to parenthesised context managers once on 3.10
     pd = str(context.config.distribution).capitalize()
     with complete_step(f"Installing build packages for {pd}"), mount_build_overlay(context):
```
behrmann, line 168: The idea here was this, if I remember correctly
```suggestion
    with (
        complete_step(f"Installing build packages for {str(context.config.distribution).capitalize()}"), 
        mount_build_overlay(context)
    ):
```
because including it on the line without being able to break it ran afoul the line length limit.

PR #2258, opened 2024-01-09
Add policycoreutils to tools trees

DaanDeMeyer:
This package provides setfiles which is required for selinux
relabeling.


behrmann (approved):
LGTM, since this is only the tools tree, which doesn't install recommends and doesn't run services, but together with either it's a recipe for weird errors on Debian.

PR #2263, opened 2024-01-10
Fix invoked_as_root initialization

DaanDeMeyer:


bluca (approved):
fixes the issue for me - would be good to have CI coverage if it's missing

DaanDeMeyer:
@bluca No coverage but rather hard to test so let's do without coverage for now.

PR #2265, opened 2024-01-10
Copy nspawn settings to the output directory again

DaanDeMeyer:
machinectl pull-tar looks for a settings file so let's make sure the output directory can be used directly for this purpose by copying the nspawn settings file to the output directory again.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1911,6 +1911,14 @@ def copy_vmlinuz(context: Context) -> None:
         break
 
 
+def copy_nspawn_settings(context: Context) -> None:
+    if context.config.nspawn_settings is None:
+        return None
+
+    with complete_step("Copying nspawn settings file…"):
+        shutil.copy2(context.config.nspawn_settings, context.staging / context.config.output_nspawn_settings)
+
+
 def copy_initrd(context: Context) -> None:
     if (context.staging / context.config.output_split_initrd).exists():
         return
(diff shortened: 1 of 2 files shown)


mkosi/__init__.py
```diff
@@ -2993,9 +3003,12 @@ def run_shell(args: Args, config: Config) -> None:
         cmdline += [f"--set-credential={k}:{v}"]
 
     with contextlib.ExitStack() as stack:
+        # Make sure the latest nspawn settings are always used.
         if config.nspawn_settings:
+            exists = (config.output_dir_or_cwd() / f"{name}.nspawn").exists()
             shutil.copy2(config.nspawn_settings, config.output_dir_or_cwd() / f"{name}.nspawn")
-            stack.callback(lambda: (config.output_dir_or_cwd() / f"{name}.nspawn").unlink())
+            if not exists:
+                stack.callback(lambda: (config.output_dir_or_cwd() / f"{name}.nspawn").unlink())
```
behrmann, line 3011: Since the callback is executed only later, this could be turned around
```
if not (config.output_dir_or_cwd() / f"{name}.nspawn").exists():
    stack.callback(lambda: (config.output_dir_or_cwd() / f"{name}.nspawn").unlink())
shutil.copy2(config.nspawn_settings, config.output_dir_or_cwd() / f"{name}.nspawn")
```
which I think is a bit clearer that this is setting up cleanup before copying it into place.

A missing_ok=True might also make this a bit more robust, since we probably don't care if it's already gone.

behrmann (approved):
LGTM, one nit

PR #2275, opened 2024-01-12
Improve SELinux binary policy selection

DaanDeMeyer:
Let's deal with the possibility that there might be more than one policy in the binary policy directory. Let's also make sure that we consider other files in the directory that might not be policies.


mkosi/__init__.py
```diff
@@ -1974,18 +1974,18 @@ def calculate_signature(context: Context) -> None:
     if sys.stderr.isatty():
         env |= dict(GPGTTY=os.ttyname(sys.stderr.fileno()))
 
+    options: list[PathString] = ["--perms", "755", "--dir", home, "--bind", home, home]
+
+    # gpg can communicate with smartcard readers via this socket so bind mount it in if it exists.
+    if (p := Path("/run/pcscd//pcscd.comm")).exists():
```
behrmann, line 1980: That double // is equivalent to a single one in that position.

mkosi/config.py
```diff
@@ -3603,15 +3604,15 @@ def want_selinux_relabel(config: Config, root: Path, fatal: bool = True) -> Opti
...
+    # The policy file is named policy.XX where XX is the policy version that indicates what features are
+    # available.
     binpolicydir = root / "etc/selinux" / policy / "policy"
-
-    try:
-        # The policy file is named policy.XX where XX is the policy version that indicates what features are
-        # available. It's not expected for there to be more than one file in this directory.
-        binpolicy = next(binpolicydir.glob("*"))
-    except StopIteration:
+    policies = [p for p in binpolicydir.glob("*") if p.suffix and all(c in string.digits for c in p.suffix[1:])]
```
behrmann, line 3610: c.isdecimal would have probably been ok here, but the comparison with string.digits is stricter, which is good. A comment to that extent would probably be good so the reasoning is clear to readers.

PR #2283, opened 2024-01-14
Make sure /etc/mtab exists in sandbox

DaanDeMeyer:
Required for pacman's CheckSpace option. To avoid messing with the
package manager tree /etc too much, we bind mount individual
subdirectories of it instead of the entire directory.

### DIFF
--- a/mkosi/sandbox.py
+++ b/mkosi/sandbox.py
@@ -138,6 +138,9 @@ def sandbox_cmd(
         *options,
     ]
 
+    if not relaxed:
+        cmdline += ["--symlink", "../proc/self/mounts", "/etc/mtab"]
+
     # If we're using /usr from a tools tree, we have to use /etc/alternatives from the tools tree as well if it
     # exists since that points directly back to /usr. Apply this after the options so the caller can mount
     # something else to /etc without overriding this mount. In relaxed mode, we only do this if /etc/alternatives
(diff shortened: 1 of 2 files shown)


mkosi/sandbox.py
```diff
@@ -150,12 +155,11 @@ def sandbox_cmd(
...
-    if devices:
-        shm = ":"
-    else:
-        shm = "chmod 1777 /dev/shm"
+    # bubblewrap creates everything with a restricted mode so relax stuff as needed.
+    shm = "chmod 1777 /dev/shm && " if not devices else ""
+    etc = "chmod 755 /etc && " if not relaxed else ""
 
-    cmdline += ["sh", "-c", f"{shm} && exec $0 \"$@\""]
+    cmdline += ["sh", "-c", f"{shm} {etc} exec $0 \"$@\"".strip()]
```
behrmann, line 162: Though it's a bit longer I think something like this might be a bit easier to read like a top down script
```python
script = []
if not devices:
    script += ["chmod 1777 /dev/shm"]
if not relaxed:
    script += ["chmod 755 /etc"]
script += ['exec $0 "$@"']
return ["sh", "-c", *[" && ".join(script)]]
```
Also doesn't need manual management for the && to be in the right place.

behrmann (approved):

PR #2285, opened 2024-01-14
Run systemd-tmpfiles as part of the build

DaanDeMeyer:
Let's make sure we take user provided tmpfiles snippets into account
as well. Since systemd now mounts the initramfs read-only by default,
we need to make sure all tmpfiles snippets that copy to /etc have
already been processed during the image build itself as they won't be
able to run during the initramfs stage.


behrmann (approved):

septatrix:
> Since systemd now mounts the initramfs read-only by default, we need to make sure all tmpfiles snippets that copy to /etc have already been processed during the image build itself as they won't be able to run during the initramfs stage.

I assume this is relevant for tmpfiles used within the initramfs itself, not those for setting up stuff on the sysroot systemd will pivot into? Is there by chance somewhere where I can read up on this? I have not read anything about this on the mailing list and I cannot something on GH

PR #2296, opened 2024-01-17
Hash root password when a plaintext one is given

DaanDeMeyer:


septatrix:
Great, I think this should be satisfactory for everyone. The flag still works in all circumstances, it is hashed, and this who want more control over seed or algorithm can still use the pre hashed variant

PR #2298, opened 2024-01-17
Introduce config_make_dict_parser()

DaanDeMeyer:


mkosi/config.py
```diff
@@ -671,6 +671,48 @@ def config_match_version(match: str, value: str) -> bool:
     return True
 
 
+def config_make_dict_parser(delimiter: str,
+                            *,
+                            parse: Callable[[str], tuple[str, Any]],
```
behrmann, line 676: The function doesn't really make sense without passing parse to it, because it would always fail and the argument doesn't have a default, so it should be treated on the same footing as delimiter.
```suggestion
                            parse: Callable[[str], tuple[str, Any]],
                            *,
```

behrmann (approved):

mkosi/config.py
DaanDeMeyer: Hmm, I feel the callsite is more obvious if the keyword is specified in this case.

PR #2300, opened 2024-01-18
Don't mount /srv and /mnt read-only

DaanDeMeyer:
It seems there are use cases where users expect to write their output
to a directory in /srv or /mnt so let's make that writable. This should
be safe as we set up a custom sandbox now so none of the tools we run
will have access to /srv and /mnt in the first place.


Edu4rdSHL (approved):
Thank you!

PR #2310, opened 2024-01-19
Add MicrocodeHost= option

behrmann:


mkosi/__init__.py
```diff
@@ -1545,6 +1545,46 @@ def build_initrd(context: Context) -> Path:
...
+                family = int(value)
+            elif key == "model":
+                model = int(value)
+            elif key == "stepping":
+                stepping = int(value)
+
+        if vendor_id is not None and family is not None and model is not None and stepping is not None:
+            break
+    else:
+        return (None, None)
```
DaanDeMeyer, line 1568: We should probably warn here if we're missing something

mkosi/__init__.py
```diff
@@ -1545,6 +1545,46 @@ def build_initrd(context: Context) -> Path:
     return config.output_dir / config.output
 
 
+def identify_cpu(root: Path) -> tuple[Optional[Path], Optional[Path]]:
+    for entry in Path("/proc/cpuinfo").read_text().split("\n\n"):
```
DaanDeMeyer, line 1549: We should skip this if we're not building for x86

mkosi/config.py
```diff
@@ -2015,6 +2016,15 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
         parse=config_make_list_parser(delimiter=",", parse=make_path_parser(required=False)),
         help="Add a user-provided initrd to image",
     ),
+    ConfigSetting(
+        dest="initrd_hostonly_microcode",
```
DaanDeMeyer, line 2020: `MicrocodeHost=` or maybe `MicrocodeHostonly=`? I would leave any mention of the initrd out of the setting name.

DaanDeMeyer (changes requested):
These changes should also be applied to build_microcode_initrd() in the kernel install plugin. Except they should be done unconditionally there.

mkosi/config.py
behrmann: Went with `MicrocodeHostonly=`.

mkosi/__init__.py
behrmann: I moved the skip into `build_microcode_initrd`

mkosi/resources/mkosi.md
```diff
@@ -1133,6 +1133,11 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
   separated list of package specifications. This option may be used
   multiple times in which case the specified package lists are combined.
 
+`InitrdHostonlyMicrocode=`, `--initrd-hostonly-microcode=`
```
DaanDeMeyer, line 1136: Needs to be updated. Also I'd go for `MicrocodeHost=` after all because it is nicely consistent with `KernelModulesIncludeHost=`

mkosi/__init__.py
```diff
@@ -1546,6 +1546,9 @@ def build_initrd(context: Context) -> Path:
 
 
 def build_microcode_initrd(context: Context) -> Optional[Path]:
+    if context.config.architecture != Architecture.x86_64:
```
DaanDeMeyer, line 1590: Let's add x86 here as well for good measure.

mkosi/__init__.py
```diff
@@ -1545,9 +1546,49 @@ def build_initrd(context: Context) -> Path:
...
+        if (ucode := (uroot / f"{family:02x}-{model:02x}-{stepping:02x}.initramfs")).exists():
+            return (Path(f"{vendor_id}.bin"), ucode)
+
+    return (Path(f"{vendor_id}.bin"), None)
+
+
 def build_microcode_initrd(context: Context) -> Optional[Path]:
     if context.config.architecture != Architecture.x86_64:
-        return
+        return None
```
DaanDeMeyer, line 1591: This should go into the previous commit

mkosi/__init__.py
```diff
@@ -1545,9 +1546,49 @@ def build_initrd(context: Context) -> Path:
...
+            if key == "vendor_id":
+                vendor_id = value
+            elif key == "cpu family":
+                family = int(value)
+            elif key == "model":
+                model = int(value)
+            elif key == "stepping":
+                stepping = int(value)
+
+        if vendor_id is not None and family is not None and model is not None and stepping is not None:
```
DaanDeMeyer, line 1566: Do we need `is not None` here or is just checking for truthy good enough?

mkosi/resources/mkosi.md
behrmann: Ah, forgot about that one. Changed the name as well, although I think it's a bit less clear I have no strong preference.

mkosi/config.py
```diff
@@ -2015,6 +2016,15 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
...
         help="Add a user-provided initrd to image",
     ),
+    ConfigSetting(
+        dest="microcode_host",
+        metavar="BOOL",
+        nargs="?",
+        section="Content",
+        parse=config_parse_boolean,
+        default=False,
+        help="Only include the host CPU's microcode"
```
DaanDeMeyer, line 2026: Trailing comma

mkosi/__init__.py
```diff
@@ -1545,7 +1546,50 @@ def build_initrd(context: Context) -> Path:
...
+        if (ucode := (uroot / f"{family:02x}-{model:02x}-{stepping:02x}")).exists():
+            return (Path(f"{vendor_id}.bin"), ucode)
+        if (ucode := (uroot / f"{family:02x}-{model:02x}-{stepping:02x}.initramfs")).exists():
+            return (Path(f"{vendor_id}.bin"), ucode)
+
+    return (Path(f"{vendor_id}.bin"), None)
+
+
 def build_microcode_initrd(context: Context) -> Optional[Path]:
+    if context.config.architecture not in [Architecture.x86, Architecture.x86_64]:
```
DaanDeMeyer, line 1590: Use tuple instead of list?

kernel-install/50-mkosi.install
```diff
@@ -44,11 +45,14 @@ def mandatory_variable(name: str) -> str:
...
 
 def build_microcode_initrd(output: Path) -> Optional[Path]:
-    amd = Path("/usr/lib/firmware/amd-ucode")
-    intel = Path("/usr/lib/firmware/intel-ucode")
+    vendor, ucode = identify_cpu(Path("/"))
 
-    if not amd.exists() and not intel.exists():
-        logging.debug("/usr/lib/firmware/{amd-ucode,intel-ucode} not found, not adding microcode initrd")
+    if vendor is None:
+        logging.debug("Unable to determine the vendor of your CPU, not adding microcode initrd")
```
DaanDeMeyer, line 51: Warning?

kernel-install/50-mkosi.install
```diff
@@ -44,11 +45,14 @@ def mandatory_variable(name: str) -> str:
...
+        return None
+
+    if ucode is None:
+        logging.debug("Unable to find microcode for your CPU in /usr/lib/firmware, not adding microcode initrd")
```
DaanDeMeyer, line 55: Ditto

kernel-install/50-mkosi.install
DaanDeMeyer: Also, "not adding microcode" should be sufficient, no need to say that it's packaged up as an initrd

DaanDeMeyer:
Second commit message needs an update

mkosi/__init__.py
behrmann: In most cases that should be alright, since I don't think the values can actually be empty (though I added a guard for that now) and the numbers will probably not be zero, but I think having an explicit dummy type here makes this more robust.

mkosi/__init__.py
behrmann: It doesn't make much of a difference under the hood - the list can't really be changed - and I prefer the visual distinctiveness of square braces. With parens I'm always thinking twice whether it's a tuple or a parenthesised expression.

DaanDeMeyer: The thing is we use parentheses almost everywhere else, so this breaks from the established practice

mkosi/resources/mkosi.md
```diff
@@ -1133,6 +1133,11 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
   separated list of package specifications. This option may be used
   multiple times in which case the specified package lists are combined.
 
+`MicrocodeHost=`, `--microcode-host=`
+
+: When set to true only include microcode for the host's CPU in the built
+  initrd.
```
DaanDeMeyer, line 1139: in the built initrd => in the image?

DaanDeMeyer: When set to true => when enabled?

kernel-install/50-mkosi.install
behrmann: Made it a warning and left out the "initrd" at the end. Just mimicked what was there before. ;)

DaanDeMeyer (approved):

mkosi/__init__.py
behrmann: Changed it for uniformities sake (though there are some differing places still), but generally I think square braces would be more readable and tuples don't have any real upside in these positions.

PR #2313, opened 2024-01-22
Allow including builtin configs

DaanDeMeyer:
Let's make it possible to include builtin configurations using
e.g. Include=mkosi-initrd. This allows building the default initrd
or default tools tree independently whereas currently these can only
be built as part of another image.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -21,7 +21,6 @@ from collections.abc import Iterator, Mapping, Sequence
 from pathlib import Path
 from typing import Optional, TextIO, Union, cast
 
-import mkosi.resources
 from mkosi.archive import extract_tar, make_cpio, make_tar
 from mkosi.burn import run_burn
 from mkosi.config import (
@@ -79,7 +78,6 @@ from mkosi.util import (
     one_zero,
     read_env_file,
     read_os_release,
-    resource_path,
     round_up,
     scopedenv,
     umask,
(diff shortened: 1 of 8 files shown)


mkosi/__init__.py
```diff
@@ -3563,24 +3558,24 @@ def run_verb(args: Args, images: Sequence[Config]) -> None:
...
+
+        images[i] = config = dataclasses.replace(
+            config,
+            tools_tree=tools.output_dir_or_cwd() / tools.output if tools else config.tools_tree,
+        )
 
-            if tools and not (tools.output_dir_or_cwd() / tools.output_with_compression).exists():
-                fork_and_wait(lambda: run_build(args, tools)) # pyright: ignore
+        if tools and not (tools.output_dir_or_cwd() / tools.output_with_compression).exists():
+            fork_and_wait(lambda: run_build(args, tools, resources=resources)) # type:ignore
```
behrmann, line 3573: I think that should stay
```suggestion
            fork_and_wait(lambda: run_build(args, tools, resources=resources)) # pyright: ignore
```
since only pyright chokes on this (also if the check is made explicitly `tools is not None`).

mkosi/__init__.py
```diff
@@ -3543,14 +3537,15 @@ def run_verb(args: Args, images: Sequence[Config]) -> None:
...
     for config in images:
         if not args.verb.needs_build() and args.verb != Verb.clean:
             continue
 
-        if config.tools_tree and config.tools_tree.name == "default":
-            with finalize_default_tools(args, config) as tools:
-                fork_and_wait(lambda: run_clean(args, tools)) # pyright: ignore
+        if config.tools_tree and config.tools_tree == Path("default"):
+            tools = finalize_default_tools(args, config, resources=resources)
+            fork_and_wait(lambda: run_clean(args, tools)) # type: ignore
```
behrmann, line 3548:
```suggestion
            cleanup_tools = finalize_default_tools(args, config, resources=resources)
            fork_and_wait(lambda: run_clean(args, cleanup_tools))
```
The reason mypy complains about this now is because they try to reify the type for the whole function, which makes the possible `None` further down travel backward in time to this point. Changing the name allows this instance of `tools` to have a different type. This also make either type ignore comment unnecessary.

behrmann (approved):

Edu4rdSHL (approved):

PR #2315, opened 2024-01-22
sandbox: Deal correctly with unmerged-usr

### LINKED ISSUE #2314: `sh` unavailable in sandbox on non `usr`-merged systems
mkosi's sandbox [mounts `/usr:/usr`](https://github.com/systemd/mkosi/blob/14b07c74d5b3a8b0f44d9a5c733f7902c7859417/mkosi/sandbox.py#L88) if no ToolsTree is specified. On NixOS, which is not usr-merged, this causes problems when programs have runtime dependencies on absolute paths in `(/usr)/bin`. This is the case for `apt-get install`, that tries to find an  `sh` at runtime in `/usr/bin` or `/bin`. That fails on NixOS, due to only `/usr` being mounted, while `sh` is in `/bin/sh`.

Two possible workarounds that I have in mind:
1. (Preferred) Find a way to tell `apt-get` which shell to use. `dpkg`  [respects the `SHELL` environment variable](https://man7.org/linux/man-pages/man1/dpkg.1.html), but I couldn't find a similar option for `apt-get`.  If there would be such an option, we should simply tell `apt-get` to use the `sh` from the Nix store that's already mounted.
2. Loosen the sandbox and mount `/bin(/sh)` into the environment. 

If you have any input regarding the first proposed solution, let me know. Otherwise, I'd be happy to implement the second solution too.

Cheers, Moritz

DaanDeMeyer:
If /bin and such are directories instead of symlinks, bind mount them in to the sandbox.

Fixes #2314

### DIFF
--- a/mkosi/sandbox.py
+++ b/mkosi/sandbox.py
@@ -137,6 +137,8 @@ def sandbox_cmd(
     for d in ("bin", "sbin", "lib", "lib32", "lib64"):
         if (p := tools / d).is_symlink():
             cmdline += ["--symlink", p.readlink(), Path("/") / p.relative_to(tools)]
+        elif p.is_dir():
+            cmdline += ["--ro-bind", p, Path("/") / p.relative_to(tools)]
 
     path = "/usr/bin:/usr/sbin" if tools != Path("/") else os.environ["PATH"]
 


behrmann (approved):

msanft:
@DaanDeMeyer works. Thanks a ton!

PR #2316, opened 2024-01-22
Don't use a tools tree for verbs that don't need a build

DaanDeMeyer:
Currently we try to use a tools tree that isn't there for verbs
that don't need a build. Let's instead opt to not use one altogether
since building an entire image just to run ssh is kind of overkill.


mkosi/__init__.py
```diff
@@ -3557,12 +3550,24 @@ def run_verb(args: Args, images: Sequence[Config]) -> None:
...
-    build = False
+    assert args.verb.needs_build()
 
-    for i, config in enumerate(images):
-        if not args.verb.needs_build():
+    for config in images:
+        if not config.minimum_version or config.minimum_version <= __version__:
             continue
 
+        die(f"mkosi {config.minimum_version} or newer is required to build this configuration (found {__version__})")
```
behrmann, line 3559: I think this would be less roundabout
```suggestion
        if config.minimum_version and config.minimum_version > __version__:
            die(f"mkosi {config.minimum_version} or newer is required to build this configuration (found {__version__})")
```

This would also allow to move the check directly below into this loop.

PR #2322, opened 2024-01-24
Add FAQ entry on how to add regular users to image

### LINKED ISSUE #2320: Proper way to create an unpriviledged user?
Hi,

I'm wondering what the recommended way is to create an unprivileged user in my image? By default, only a root user is created as far as I can tell.

I see that [there is special handling for the `adduser` command](https://github.com/systemd/mkosi/blob/dc94177f3c57152770cb1c2316f971f6732d9525/mkosi/resources/mkosi.md?plain=1#L1947) which suggests that I should use that.

I've been using this in a `mkosi.postinst` script like so:

```bash
#!/bin/bash

useradd --user-group --create-home --shell=/usr/bin/bash --password='mypwd' user
```

which seems to work okay, but I'm wondering if this is the 'right'/idiomatic way to do it. I could not find more hints in the docs or example projects.

Might be related to #1930, but I'm not sure.

I'd appreciate a little docs snippet on that issue.

Thanks

DaanDeMeyer:
Fixes #2320

### DIFF
--- a/mkosi/resources/mkosi.md
+++ b/mkosi/resources/mkosi.md
@@ -2316,6 +2316,18 @@ Note that the minimum required Python version is 3.9.
   `/etc/tmpfiles.d/static-nodes-permissions.conf` and change the mode of
   `/dev/kvm` from `0660` to `0666`.
 
+- How do I add a regular user to an image?
+
+  You can use the following snippet in a post-installation script:
+
+  ```sh
+  useradd --create-home --user-group $USER --password "$(openssl passwd -stdin -6 <$USER_PASSWORD_FILE)"
+  ```
+
+  Note that from systemd v256 onwards, if enabled,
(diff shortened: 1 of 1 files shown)


mkosi/resources/mkosi.md
```diff
@@ -2316,6 +2316,18 @@ these settings across reboots, copy
...
+
+You can use the following snippet in a post-installation script:
+
+```sh
+useradd --create-home --user-group $USER --password "$(openssl passwd -stdin -6 <$USER_PASSWORD_FILE)"
+```
+
+Note that from systemd v256 onwards, if enabled,
+`systemd-homed-firstboot.service` will prompt to create a regular user
+on first boot if there are no regular users.
```
behrmann, line 2329: Can you indent the body by two spaces (same for the entry above) like this? This way the generated man page has the correct indentation as well.

behrmann (approved):

PR #2323, opened 2024-01-24
Use a subdirectory of TMPDIR as the sandbox's /var/tmp

DaanDeMeyer:
Let's take TMPDIR into account instead of always using /var/tmp. We
also make sure we override TMPDIR to /var/tmp in the sandbox.


DaanDeMeyer:
Replaces #2319

behrmann (approved):

PR #2328, opened 2024-01-25
Add CompressLevel= setting

DaanDeMeyer:
Let's allow configuring the compression level.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -545,10 +545,27 @@ def config_parse_source_date_epoch(value: Optional[str], old: Optional[int]) -> 
     try:
         timestamp = int(value)
     except ValueError:
-        raise ValueError(f"{value} is not a valid timestamp")
+        die(f"{value} is not a valid timestamp")
+
     if timestamp < 0:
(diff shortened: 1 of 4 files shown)


mkosi/config.py
```diff
@@ -1685,6 +1703,14 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
         default_factory_depends=("distribution", "release", "output_format"),
         help="Enable whole-output compression (with images or archives)",
     ),
+    ConfigSetting(
+        dest="compress_level",
+        metavar="LEVEL",
+        section="Output",
+        parse=config_parse_compress_level,
+        default=3,
```
behrmann, line 1711: It's a bit unfortunate that zstd is the odd one out here—the default level for gzip and xz is 6. Since I think that people should and probably will default to zstd nowadays, I think it's fine to default to that and not add the additional complexity of having different defaults for different compressors. It should be noted in the NEWS entry on this feature, though, that the default of 3 is lower than the gzip and xz default of 6.

behrmann (approved):

PR #2331, opened 2024-01-26
Run distribution upgrade automatically in some scenarios

DaanDeMeyer:
If we notice extra packages were added by the build script to the
local packages repository or if we're running an incremental build
and there are local packages, do a distribution upgrade so that local
packages can be upgraded without having to throw away the cached
images.

### DIFF
--- a/mkosi/installer/apt.py
+++ b/mkosi/installer/apt.py
@@ -154,14 +154,12 @@ def createrepo_apt(context: Context) -> None:
         run(["dpkg-scanpackages", context.packages],
             stdout=f, sandbox=context.sandbox(options=["--ro-bind", context.packages, context.packages]))
 
-    (context.pkgmngr / "etc/apt/sources.list.d").mkdir(parents=True, exist_ok=True)
-    (context.pkgmngr / "etc/apt/sources.list.d/mkosi-packages.sources").write_text(
-        f"""\
-        Enabled: yes
-        Types: deb
-        URIs: file:///work/packages
-        Suites: {context.config.release}
-        Components: main
-        Trusted: yes
-        """
+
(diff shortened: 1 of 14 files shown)


mkosi/installer/apt.py
```diff
@@ -39,11 +48,21 @@ def setup_apt(context: Context, repos: Sequence[str]) -> None:
...
             for repo in repos:
-                f.write(f"{repo}\n")
+                f.write(
+                    textwrap.dedent(
+                        f"""\
+                        Types: {" ".join(repo.types)}
+                        URIs: {repo.url}
+                        Suites: {repo.suite}
+                        Components: {" ".join(repo.components)}
+                        {"Signed-By" if repo.signedby else "Trusted"}: {repo.signedby or "yes"}
```
behrmann, line 62:
```suggestion
                        {f"Signed-By: {repo.signedby}" if repo.signedby else "Trusted: yes"}
```
contra: nested f-string
pro: easy to read what belongs where, tested to wrk on 3.9

mkosi/installer/apt.py
```diff
@@ -39,11 +48,21 @@ def setup_apt(context: Context, repos: Sequence[str]) -> None:
...
+                        """
```
behrmann, line 63: The whole thing would be great as the `__str__` method on `AptRepository`—would save the `textwrap.dedent`, too.

behrmann (approved):
Maybe we could make a function on context
```
def need_local_repo(self) -> bool:
    return any(self.packages.iterdir())
```
because somehow I stumbled over this twice when reading it and thought about packages to be installed and not the package dir.

DaanDeMeyer:
Will submit a new PR

PR #2351, opened 2024-02-07
Add support for oci-dir output (fixes #1865)

### LINKED ISSUE #1865: OCI as output format
mkosi is a nice tool for building (in addition to other things) nspawn images.
While nspawn can use a large number of different formats, OCI is a very convenient one since there is existing infrastructure for distribution (container registries).
Should mkosi learn how to generate the [on disk OCI format](https://github.com/opencontainers/image-spec/blob/main/image-layout.md)?
Additionally, should mkosi be able to reuse existing directory trees from OCI images / layers as inputs?
Not sure if this makes sense and if this belongs here (or should be implemented in systemd-repart instead).

septatrix:
A first draft for rudimentary OCI image outputs

PS: I intend to squash this once more questions are answered/resolved

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -20,7 +20,7 @@ import textwrap
 import uuid
 from collections.abc import Iterator, Mapping, Sequence
 from pathlib import Path
-from typing import Optional, TextIO, Union, cast
+from typing import Optional, Union, cast
 
 from mkosi.archive import extract_tar, make_cpio, make_tar
 from mkosi.burn import run_burn
@@ -75,6 +75,7 @@ from mkosi.util import (
     flock,
     flock_or_die,
     format_rlimit,
+    hash_file,
     make_executable,
     one_zero,
     parents_below,
@@ -2457,17 +2458,6 @@ def copy_initrd(context: Context) -> None:
         break
 
 
-def hash_file(of: TextIO, path: Path) -> None:
-    bs = 16 * 1024**2
-    h = hashlib.sha256()
-
-    with path.open("rb") as sf:
-        while (buf := sf.read(bs)):
-            h.update(buf)
-
-    of.write(h.hexdigest() + " *" + path.name + "\n")
-
-
 def calculate_sha256sum(context: Context) -> None:
     if not context.config.checksum:
         return
@@ -2478,7 +2468,7 @@ def calculate_sha256sum(context: Context) -> None:
     with complete_step("Calculating SHA256SUMS…"):
         with open(context.workspace / context.config.output_checksum, "w") as f:
             for p in context.staging.iterdir():
-                hash_file(f, p)
+                print(hash_file(p) + " *" + p.name, file=f)
 
         (context.workspace / context.config.output_checksum).rename(context.staging / context.config.output_checksum)
 
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3260,6 +3260,103 @@ def make_disk(
     return make_image(context, msg=msg, skip=skip, split=split, tabs=tabs, root=context.root, definitions=definitions)
 
 
+def make_oci(context: Context, root_layer: Path, dst: Path) -> None:
+    ca_store = dst / "blobs" / "sha256"
+    with umask(~0o755):
+        ca_store.mkdir(parents=True)
+
+    layer_diff_digest = hash_file(root_layer)
+    maybe_compress(
+        context,
+        context.config.compress_output,
+        context.staging / "rootfs.layer",
+        # Pass explicit destination to suppress adding an extension
+        context.staging / "rootfs.layer",
+    )
+    layer_digest = hash_file(root_layer)
+    root_layer.rename(ca_store / layer_digest)
+
+    creation_time = (
+        datetime.datetime.fromtimestamp(context.config.source_date_epoch, tz=datetime.timezone.utc)
+        if context.config.source_date_epoch is not None
+        else datetime.datetime.now(tz=datetime.timezone.utc)
+    ).isoformat()
+
+    oci_config = {
+        "created": creation_time,
+        "architecture": context.config.architecture.to_oci(),
+        # Name of the operating system which the image is built to run on as defined by
+        # https://github.com/opencontainers/image-spec/blob/v1.0.2/config.md#properties.
+        "os": "linux",
+        "rootfs": {
+            "type": "layers",
+            "diff_ids": [f"sha256:{layer_diff_digest}"],
+        },
+        "config": {
+            "Cmd": [
+                "/sbin/init",
+                *context.config.kernel_command_line,
+            ],
+        },
+        "history": [
+            {
+                "created": creation_time,
+                "comment": "Created by mkosi",
+            },
+        ],
+    }
+    oci_config_blob = json.dumps(oci_config)
+    oci_config_digest = hashlib.sha256(oci_config_blob.encode()).hexdigest()
+    with umask(~0o644):
+        (ca_store / oci_config_digest).write_text(oci_config_blob)
+
+    layer_suffix = context.config.compress_output.oci_media_type_suffix()
+    oci_manifest = {
+        "schemaVersion": 2,
+        "mediaType": "application/vnd.oci.image.manifest.v1+json",
+        "config": {
+            "mediaType": "application/vnd.oci.image.config.v1+json",
+            "digest": f"sha256:{oci_config_digest}",
+            "size": (ca_store / oci_config_digest).stat().st_size,
+        },
+        "layers": [
+            {
+                "mediaType": f"application/vnd.oci.image.layer.v1.tar{layer_suffix}",
+                "digest": f"sha256:{layer_digest}",
+                "size": (ca_store / layer_digest).stat().st_size,
+            }
+        ],
+        "annotations": {
+            "io.systemd.mkosi.version": __version__,
+            **({
+                "org.opencontainers.image.version": context.config.image_version,
+            } if context.config.image_version else {}),
+        }
+    }
+    oci_manifest_blob = json.dumps(oci_manifest)
+    oci_manifest_digest = hashlib.sha256(oci_manifest_blob.encode()).hexdigest()
+    with umask(~0o644):
+        (ca_store / oci_manifest_digest).write_text(oci_manifest_blob)
+
+        (dst / "index.json").write_text(
+            json.dumps(
+                {
+                    "schemaVersion": 2,
+                    "mediaType": "application/vnd.oci.image.index.v1+json",
+                    "manifests": [
+                        {
+                            "mediaType": "application/vnd.oci.image.manifest.v1+json",
+                            "digest": f"sha256:{oci_manifest_digest}",
+                            "size": (ca_store / oci_manifest_digest).stat().st_size,
+                        }
+                    ],
+                }
+            )
+        )
+
+        (dst / "oci-layout").write_text(json.dumps({"imageLayoutVersion": "1.0.0"}))
+
+
 def make_esp(context: Context, uki: Path) -> list[Partition]:
     if not (arch := context.config.architecture.to_efi()):
         die(f"Architecture {context.config.architecture} does not support UEFI")
@@ -3560,6 +3657,17 @@ def build_image(context: Context) -> None:
             context.root, context.staging / context.config.output_with_format,
             tools=context.config.tools(),
             sandbox=context.sandbox,
+        )
+    elif context.config.output_format == OutputFormat.oci:
+        make_tar(
+            context.root, context.staging / "rootfs.layer",
+            tools=context.config.tools(),
+            sandbox=context.sandbox,
+        )
+        make_oci(
+            context,
+            context.staging / "rootfs.layer",
+            context.staging / context.config.output_with_format,
         )
     elif context.config.output_format == OutputFormat.cpio:
         make_cpio(
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -173,6 +173,7 @@ class OutputFormat(StrEnum):
     sysext    = enum.auto()
     tar       = enum.auto()
     uki       = enum.auto()
+    oci       = enum.auto()
 
     def extension(self) -> str:
         return {
@@ -205,6 +206,7 @@ class Compression(StrEnum):
     xz   = enum.auto()
     bz2  = enum.auto()
     gz   = enum.auto()
+    gzip = "gz"
     lz4  = enum.auto()
     lzma = enum.auto()
 
@@ -215,6 +217,18 @@ class Compression(StrEnum):
         return {
             Compression.zstd: ".zst"
         }.get(self, f".{self}")
+
+    def oci_media_type_suffix(self) -> str:
+        suffix = {
+            Compression.none: "",
+            Compression.gz:   "+gzip",
+            Compression.zstd: "+zstd",
+        }.get(self)
+
+        if not suffix:
+            die(f"Compression {self} not supported for OCI layers")
+
+        return suffix
 
 
 class DocFormat(StrEnum):
@@ -376,6 +390,28 @@ class Architecture(StrEnum):
 
         if not a:
             die(f"Architecture {self} not supported by QEMU")
+
+        return a
+
+    def to_oci(self) -> str:
+        a = {
+            Architecture.arm         : "arm",
+            Architecture.arm64       : "arm64",
+            Architecture.loongarch64 : "loong64",
+            Architecture.mips64_le   : "mips64le",
+            Architecture.mips_le     : "mipsle",
+            Architecture.ppc         : "ppc",
+            Architecture.ppc64       : "ppc64",
+            Architecture.ppc64_le    : "ppc64le",
+            Architecture.riscv32     : "riscv",
+            Architecture.riscv64     : "riscv64",
+            Architecture.s390x       : "s390x",
+            Architecture.x86         : "386",
+            Architecture.x86_64      : "amd64",
+        }.get(self)
+
+        if not a:
+            die(f"Architecture {self} not supported by OCI")
 
         return a
 
@@ -634,6 +670,8 @@ def config_default_compression(namespace: argparse.Namespace) -> Compression:
             return Compression.xz
         else:
             return Compression.zstd
+    elif namespace.output_format == OutputFormat.oci:
+        return Compression.gz
     else:
         return Compression.none
 
--- a/mkosi/resources/mkosi.md
+++ b/mkosi/resources/mkosi.md
@@ -667,7 +667,8 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
   archive is generated), `disk` (a block device OS image with a GPT
   partition table), `uki` (a unified kernel image with the OS image in
   the `.initrd` PE section), `esp` (`uki` but wrapped in a disk image
-  with only an ESP partition), `sysext`, `confext`, `portable` or `none`
+  with only an ESP partition), `oci` (a directory compatible with the
+  OCI image specification), `sysext`, `confext`, `portable` or `none`
   (the OS image is solely intended as a build image to produce another
   artifact).
 
@@ -705,11 +706,12 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 : Configure compression for the resulting image or archive. The argument can be
   either a boolean or a compression algorithm (`xz`, `zstd`). `zstd`
   compression is used by default, except CentOS and derivatives up to version
-  8, which default to `xz`. Note that when applied to block device image types,
+  8, which default to `xz`, and OCI images, which default to `gzip`.
+  Note that when applied to block device image types,
   compression means the image cannot be started directly but needs to be
   decompressed first. This also means that the `shell`, `boot`, `qemu` verbs
   are not available when this option is used. Implied for `tar`, `cpio`, `uki`,
-  and `esp`.
+  `esp`, and `oci`.
 
 `CompressLevel=`, `--compress-level=`
 
--- a/mkosi/util.py
+++ b/mkosi/util.py
@@ -7,6 +7,7 @@ import enum
 import errno
 import fcntl
 import functools
+import hashlib
 import importlib
 import importlib.resources
 import itertools
@@ -298,3 +299,15 @@ def resource_path(mod: ModuleType) -> Iterator[Path]:
             p.parent.chmod(0o755)
 
         yield p
+
+
+def hash_file(path: Path) -> str:
+    # TODO Replace with hashlib.file_digest after dropping support for Python 3.10.
+    bs = 16 * 1024**2
+    h = hashlib.sha256()
+
+    with path.open("rb") as sf:
+        while (buf := sf.read(bs)):
+            h.update(buf)
+
+    return h.hexdigest()
--- a/tests/test_boot.py
+++ b/tests/test_boot.py
@@ -57,7 +57,7 @@ def test_format(config: Image.Config, format: OutputFormat) -> None:
         if image.config.distribution == Distribution.rhel_ubi:
             return
 
-        if format in (OutputFormat.tar, OutputFormat.none) or format.is_extension_image():
+        if format in (OutputFormat.tar, OutputFormat.oci, OutputFormat.none) or format.is_extension_image():
             return
 
         if format == OutputFormat.directory and not find_virtiofsd():


mkosi/archive.py
```diff
@@ -63,6 +66,72 @@ def make_tar(src: Path, dst: Path, *, tools: Path = Path("/"), sandbox: Sequence
         )
 
 
+def make_oci(context: Context, dst: Path) -> None:
+    ca_store = dst / "blobs" / "sha256"
+    ca_store.mkdir(parents=True)
+
+    layer_archive = context.staging / context.config.output_tar
+    with layer_archive.open("rb") as f:
+        layer_digest = hashlib.file_digest(f, "sha256").hexdigest()
```
septatrix, line 75: `file_digest` only exists in Python 3.11, I will change this in the final version similar to a loop similar to what is done for the `SHA256SUM` logic

mkosi/config.py
```diff
@@ -348,6 +352,28 @@ def to_qemu(self) -> str:
...
+            Architecture.mips_le     : "mipsle",
+            Architecture.ppc         : "ppc",
+            Architecture.ppc64       : "ppc64",
+            Architecture.ppc64_le    : "ppc64le",
+            Architecture.riscv32     : "riscv",
+            Architecture.riscv64     : "riscv64",
+            Architecture.s390x       : "s390x",
+            Architecture.x86         : "386",
+            Architecture.x86_64      : "amd64",
+        }.get(self)
```
septatrix, line 370: I am unable to check this with all the arches, I just took this list from the specification/documentation

septatrix (commented):
Please also note the several inline comments I left

mkosi/archive.py
septatrix: We could also backport this but I am not sure how to reference licenses in such cases, instead I moved the existing `hash_file` into utils and also use it here

mkosi/config.py
```diff
@@ -324,6 +326,8 @@ def to_efi(self) -> Optional[str]:
             Architecture.loongarch64 : "loongarch64",
         }.get(self)
 
+        # TODO should this have an explicit check+die?
```
behrmann, line 329: Why the comment here?

mkosi/config.py
```diff
@@ -157,6 +157,8 @@ class OutputFormat(StrEnum):
     sysext    = enum.auto()
     tar       = enum.auto()
     uki       = enum.auto()
+    # TODO Do we want to support oci-dir or oci-tar, or both?
+    oci       = enum.auto()
```
behrmann, line 161: Go with oci_dir, since that's it and it leaves more open
```suggestion
    # TODO Do we want to support oci-dir or oci-tar, or both?
    oci       = enum.auto()
```

mkosi/archive.py
```diff
@@ -63,6 +67,72 @@ def make_tar(src: Path, dst: Path, *, tools: Path = Path("/"), sandbox: Sequence
         )
 
 
+def make_oci(context: Context, dst: Path) -> None:
+    ca_store = dst / "blobs" / "sha256"
+    ca_store.mkdir(parents=True)
+
+    layer_archive = context.staging / context.config.output_tar
+    layer_digest = hash_file(layer_archive)
+    # Should we create a symlink at the old path such that the tar can still be used?
```
behrmann, line 76: I don't think we need to generate more output than people ask for, when they want tar, they can generate a tar.

mkosi/archive.py
```diff
@@ -63,6 +67,72 @@ def make_tar(src: Path, dst: Path, *, tools: Path = Path("/"), sandbox: Sequence
...
+    layer_archive.rename(ca_store / layer_digest)
+
+    oci_config = {
+        # TODO Add more metadata? Created timestamp, history, config...
+        "architecture": context.config.architecture.to_oci(),
+        "os": "linux",
```
behrmann, line 82: Is that always just "linux" or should that be the distro?

mkosi/archive.py
```diff
@@ -63,6 +67,72 @@ def make_tar(src: Path, dst: Path, *, tools: Path = Path("/"), sandbox: Sequence
...
+    ca_store = dst / "blobs" / "sha256"
+    ca_store.mkdir(parents=True)
+
+    layer_archive = context.staging / context.config.output_tar
+    layer_digest = hash_file(layer_archive)
+    # Should we create a symlink at the old path such that the tar can still be used?
+    layer_archive.rename(ca_store / layer_digest)
+
+    oci_config = {
+        # TODO Add more metadata? Created timestamp, history, config...
```
behrmann, line 80: timestamp makes sense, though that should probably use `SOURCE_DATE_EPOCH`. Are there any fields that could use the image version or image id? 

No idea what history and config are.

mkosi/util.py
```diff
@@ -258,3 +259,15 @@ def _write_contents(target, source):
...
+# TODO Replace with hashlib.file_digest after dropping support for Python 3.11.
+def hash_file(path: Path) -> str:
+    bs = 16 * 1024**2
+    h = hashlib.sha256()
+
+    with path.open("rb") as sf:
+        while (buf := sf.read(bs)):
+            h.update(buf)
+    
+    return h.hexdigest()
```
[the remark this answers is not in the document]
septatrix: Oh no, I have become the very thing I swore to destroy ^^ (While working I started to use the shortcut for save without formatting which also disables adding the trailing newline, after a while I simply disabled Black for this repo but that did not save this normally I guess).

mkosi/config.py
septatrix: All the other `to_*` methods have explicit checks if the architecture is supported and exit with a proper error message except for this one. That seemed like an oversight though on second look it seems like the `None` case is used somewhere and explicit checks are in some other places

mkosi/config.py
septatrix: I wanted to go with the nomenclature used by the containers tools which use [`oci` and `oci-archive`](https://github.com/containers/image/blob/278eaa3c1b629f0438e784cde703403c44898fc4/docs/containers-transports.5.md#ocipathreference)

mkosi/archive.py
septatrix: I was thinking of matching the current behavior for initrd/efi files which are also provided even when building disk images

mkosi/archive.py
septatrix: That is always linux (https://github.com/opencontainers/image-spec/blob/365fa41472a19107b0b4cbc17e141a9b97f9b66e/config.md?plain=1#L108)

mkosi/archive.py
septatrix: > timestamp makes sense, though that should probably use `SOURCE_DATE_EPOCH`.

Good idea.

> Are there any fields that could use the image version or image id?

There is `os.version` though that is for the host system on which this should run (which is also the case for `os`).

Otherwise we have labels/annotations, [many of which are standardized](https://github.com/opencontainers/image-spec/blob/365fa41472a19107b0b4cbc17e141a9b97f9b66e/annotations.md#pre-defined-annotation-keys). For us `org.opencontainers.image.version` would work. There is also `org.opencontainers.image.base.name` but that is actually for base images in the Docker sense, not our base distribution.

> No idea what history and config are.

Config can be used to preset information used when running the image. E.g. an entrypoint/cmd, working dir, environment variables. One idea would be to set `/sbin/init` here though I am not sure if that is the best idea. If we do not set anything at least the entrypoint/cmd always has to be given explicitly. Working dir seems to default to `/` and env is empty apart from what the container runtime sets.
(Users are still able to create a new image building upon the oci output from mkosi and set these afterwards. Though I am not sure about the UX for docker - might be that they have to explicitly load the oci image into the local registry first. With podman this is as simple as `FROM oci:mkosi.output/image\nCMD /usr/bin/bash`)

History is used to log the stuff applied to the image, e.g. commands run, files added etc. Dockerhub seem to show them when viewing details about specific image version, e.g. https://hub.docker.com/layers/library/fedora/40/images/sha256-e7c8351c0292634436207cd10654eff5361153f0617e71c8fbd6e8dcf8d6f826?context=explore. These mostly seem to use the syntax from Dockerfiles though the spec does not set any restrictions on them. I am not sure what would happen if one where to just write random stuff in there.
We could save all invoked commands and store them in the history but that seems rather cumbersome and not worthwhile.

septatrix: > Though I am not sure about the UX for docker - might be that they have to explicitly load the oci image into the local registry first. With podman this is as simple as `FROM oci:mkosi.output/image\nCMD /usr/bin/bash`)

I just checked how this works with docker. Unless I missed something one has to run `tar -c . | docker import - my-mkosi-oci` before being able to use the image. Docker (or buildx) does not support simply using `oci:...` as an image specifier but instead interprets it as an image with name "oci".

mkosi/config.py
septatrix: `docker image import` only seems to allow importing tars so supporting both might be a good idea (at least in the long run).

behrmann: > I wanted to go with the nomenclature used by the containers tools which use [oci and oci-archive](https://github.com/containers/image/blob/278eaa3c1b629f0438e784cde703403c44898fc4/docs/containers-transports.5.md#ocipathreference)

If that's the official nomenclature, then let's go with that. 

> docker image import only seems to allow importing tars so supporting both might be a good idea (at least in the long run).

Is docker still a thing? I thought all the cool kids these days used podman.

mkosi/archive.py
behrmann: That's true, but the tar might be quite a bit bigger and we have an explicit tar output. In the end I'm only -0 about it, though.

mkosi/archive.py
behrmann: What a terrible spec, they could link to whatever they reference. Could you add a comment explaining that this is (I suppose) the first half (i.e. non-arch) output of `go tool dist list`?

septatrix:
Should I add something like podman in the CI to test the OCI? A runtime-spec for `systemd-nspawn --oci-bundle` should also be possible but that would be a bit more complex

mkosi/archive.py
behrmann: History and config seem a bit pointless or too far out, but setting `org.opencontainers.image.base.name` might make sense. Wouldn't the images produced by mkosi be for all intents and purpses base images?

behrmann:
I guess. If we are to add this, it should be covered by a test.

mkosi/archive.py
septatrix: They do (and it works in the rendered markdown) though I linked the raw markdown where the link target is in the footnotes: https://golang.org/doc/install/source#environment. I will add the comment though

mkosi/archive.py
septatrix: Yes though the base name annotation is for the other way around. It should refer to the image upon which this image is based. E.g. if I create an image based on `registry.fedoraproject.org/fedora:latest` then that would saved in the annotations.

mkosi/config.py
septatrix: > If that's the official nomenclature, then let's go with that.

Not by the [opencontainers initiative](https://github.com/opencontainers) itself as they do not specify how image names should be resolved but its the name used by the [containers projects](https://github.com/containers) which actually support working directly with local OCIs.

> Is docker still a thing?

Not on anything I touch but yes, still alive and kicking.

septatrix:
Thinking about the tests: Unpacking the OCI bundle and creating the runtime spec is rather annoying and [`umoci`](https://github.com/opencontainers/umoci) which would be great for this is not package in most places. So while using `sd-nspawn --oci-bundle` would be neat it is rather annoying and unpacking the OCI and creating the runtime spec on our own would add places where we could hide our own bugs (apart from being annoying).

As a result I think using podman to test this is likely the simplest way for now. However, installing podman in the tools tree also has it problems as the tools tree is inaccessible from within the pytest tests as the proper config parsing etc is only done within mkosi itself. So using the tools tree version would only be possible if we add podman support to the boot/shell verbs. I am not sure if this is desired as this would also likely behave differently regarding stuff like `RuntimeTree`, network namespaces, etc...

Alternatively we could install podman outside of the tools tree.

Which variant do you prefer?

mkosi/archive.py
septatrix: I tried out setting a `Cmd` in config such that `/sbin/init` is started automatically, I think that might be a reasonable thing to do but we can also drop it. I am a bit toying around with what would make sense.

I also added a single history field for testing with a `comment` and `created_by` field. The comment does not seem to show up but the created_by field is at least shown in Docker Hub for my demo image: https://hub.docker.com/layers/septatrix/mkosi-oci-fedora-test/latest/images/sha256-a81676033c96a8de79a3441640fbfb3ca14da09276e1409281e7ce02ecc44e55?context=explore

septatrix:
I have added a basic tests using podman

mkosi/archive.py
septatrix: The tar is already created as it is used inside the OCI (also its currently not compressed, see comment further below in the code). So it would only be a symlink to an already existing file

septatrix (commented):
The CI failures for Arch and Debian testing seem to be due to https://github.com/systemd/systemd/issues/29860. No clue why CentOS fails and I did not find the motivation to lock through all the logs yet.

Regardless: This is not a problem with the way the OCI images are constructed but a bug/misconfiguration in either podman or systemd. Otherwise all CI runs would fail. I am actually inclined to add skips for the failing scenarios. People can still use the OCI generated by mkosi and start the tools directly instead of going via the init system. Merging this functionality in mkosi could even be beneficial for podman/systemd to better diagnose who exactly is at fault here.

septatrix:
> No clue why CentOS fails and I did not find the motivation to lock through all the logs yet.

It's an endless loop due to journald not starting properly (which is also why the timeout after 1h kicked in. The repeating log section is attached. I have found a few systemd issues which seem to reference what I would guess is the error cause (`PR_SET_MM_ARG_START failed...`). As this does not happen in the other distros I would assume that this has since been fixed in upstream systemd further making me lean towards skipping the OCI testing for centos.

<details>
<summary>Log extract</summary>

```txt
         Starting Journal Service...
systemd-tmpfiles-setup.service: starting held back, waiting for: systemd-journald.service
systemd-journald.service: Failed to add invocation ID to keyring, ignoring: Operation not permitted
Operating on architecture: x86
Operating on architecture: x32
Operating on architecture: x86-64
Operating on architecture: x86
Operating on architecture: x32
Operating on architecture: x86-64
Operating on architecture: x86
Operating on architecture: x32
Operating on architecture: x86-64
Restricting namespace to: n/a.
Operating on architecture: x86
Blocking cgroup.
Blocking ipc.
Blocking net.
Blocking mnt.
Blocking pid.
Blocking user.
Blocking uts.
Blocking time.
Operating on architecture: x32
Blocking cgroup.
Blocking ipc.
Blocking net.
Blocking mnt.
Blocking pid.
Blocking user.
Blocking uts.
Blocking time.
Operating on architecture: x86-64
Blocking cgroup.
Blocking ipc.
Blocking net.
Blocking mnt.
Blocking pid.
Blocking user.
Blocking uts.
Blocking time.
Operating on architecture: x86
Operating on architecture: x32
Operating on architecture: x86-64
Operating on architecture: x86-64
[**    ] (2 of 2) A start job is running for…odes in /dev (6min 41s / no limit)
M
[K[*     ] (2 of 2) A start job is running for…odes in /dev (6min 42s / no limit)
M
[K[**    ] (2 of 2) A start job is running for…odes in /dev (6min 42s / no limit)
M
[K[***   ] (1 of 2) A start job is running for Journal Service (3s / no limit)
M
[K[ ***  ] (1 of 2) A start job is running for Journal Service (4s / no limit)
M
[K[  *** ] (1 of 2) A start job is running for Journal Service (4s / no limit)
M
[K[   ***] (2 of 2) A start job is running for…odes in /dev (6min 44s / no limit)
M
[K[    **] (2 of 2) A start job is running for…odes in /dev (6min 45s / no limit)
M
[K[     *] (2 of 2) A start job is running for…odes in /dev (6min 45s / no limit)
M
[K[    **] (1 of 2) A start job is running for Journal Service (6s / no limit)
M
[K[   ***] (1 of 2) A start job is running for Journal Service (7s / no limit)
M
[K[  *** ] (1 of 2) A start job is running for Journal Service (7s / no limit)
M
[K[ ***  ] (2 of 2) A start job is running for…odes in /dev (6min 47s / no limit)
M
[K[***   ] (2 of 2) A start job is running for…odes in /dev (6min 48s / no limit)
M
[K[**    ] (2 of 2) A start job is running for…odes in /dev (6min 48s / no limit)
M
[K[*     ] (1 of 2) A start job is running for Journal Service (9s / no limit)
M
[K[**    ] (1 of 2) A start job is running for Journal Service (10s / no limit)
Received SIGCHLD from PID 64 (systemd-journal).
Child 64 (systemd-journal) died (code=exited, status=1/FAILURE)
systemd-journald.service: Child 64 belongs to systemd-journald.service.
systemd-journald.service: Main process exited, code=exited, status=1/FAILURE
systemd-journald.service: Failed with result 'exit-code'.
systemd-journald.service: Service will restart (restart setting)
systemd-journald.service: Changed start -> failed
systemd-journald.service: Job 293 systemd-journald.service/start finished, result=failed
M
[K[FAILED] Failed to start Journal Service.
[KSee 'systemctl status systemd-journald.service' for details.
systemd-journald.socket: Changed running -> listening
systemd-journald.service: Unit entered failed state.
systemd-journald.service: Consumed 15ms CPU time.
systemd-journald.service: Changed failed -> auto-restart
systemd-journald.service: Control group is empty.
systemd-journald.service: Service has no hold-off time (RestartSec=0), scheduling restart.
systemd-journald.service: Trying to enqueue job systemd-journald.service/restart/replace
systemd-journald.service: Installed new job systemd-journald.service/restart as 298
systemd-journald.service: Enqueued job systemd-journald.service/restart as 298
systemd-journald.service: Scheduled restart job, restart counter is at 42.
systemd-journald.socket: Incoming traffic
systemd-journald.socket: Changed listening -> running
sysinit.target: starting held back, waiting for: systemd-resolved.service
systemd-journal-flush.service: starting held back, waiting for: systemd-journald.service
systemd-journald.service: Changed auto-restart -> dead
systemd-journald.service: Job 298 systemd-journald.service/restart finished, result=done
[  OK  ] Stopped Journal Service.
systemd-journald.service: Converting job systemd-journald.service/restart -> systemd-journald.service/start
systemd-journald.service: Consumed 15ms CPU time.
sysinit.target: starting held back, waiting for: systemd-resolved.service
systemd-journal-flush.service: starting held back, waiting for: systemd-journald.service
systemd-journald.service: Will spawn child (service_enter_start): /usr/lib/systemd/systemd-journald
systemd-journald.service: Failed to set 'trusted.invocation_id' xattr on control group /system.slice/systemd-journald.service, ignoring: Operation not permitted
systemd-journald.service: Failed to remove 'trusted.delegate' xattr flag on control group /system.slice/systemd-journald.service, ignoring: Operation not permitted
systemd-journald.service: Passing 3 fds to service
systemd-journald.service: About to execute /usr/lib/systemd/systemd-journald
systemd-journald.service: Forked /usr/lib/systemd/systemd-journald as 65
PR_SET_MM_ARG_START failed: Operation not permitted
systemd-journald.service: Changed dead -> start
```

</details>

DaanDeMeyer:
We're trying to get `umoci` packaged in Fedora so then it should be OK to depend on it.

DaanDeMeyer:
@septatrix Looking at https://pkgs.org/download/umoci, it seems that only centos/fedora are missing and the rest is covered already. Are there any particular distributions you have in mind?

septatrix:
Oh in that case I guess not. I was looking at their readme but I guess that's quite outdated: https://github.com/opencontainers/umoci?tab=readme-ov-file#install

tests/test_boot.py
```diff
@@ -40,6 +40,28 @@ def test_boot(config: Image.Config, format: OutputFormat) -> None:
...
+            # This is not possible as overmounting is not permitted inside the container.
+            # Running the container with higher privileges using `--privileged` or `--cap-add SYS_ADMIN`
+            # only results in several other units misbehaving.
+            # We could alternatively add a machine-id file ourselves
+            # though that would skip the first-boot logic which we would also like to test.
+            # Masking these units is the simplest way out until we can figure out something better
+            # or systemd resolved this upstream.
+            # TODO it seems that mounting an empty file at /etc/machine-id also works
+            # but than systemd-machine-id-commit.service fails as /etc/ is not on a temporary file system...
+            # And that would also skip the first-boot logic.
```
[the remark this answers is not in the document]
septatrix: I marked these for skipping as I do not consider supporting containers with systemd inside them to be mandatory for this feature and it is unlikely that there is something we could do to improve this.

However, I am tempted to /cc someone like rhatdan who might know more about the podman side of things. Do you perhaps know anyone more suited for this before I ping him?

mkosi/resources/mkosi.md
```diff
@@ -666,11 +667,12 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 : Configure compression for the resulting image or archive. The argument can be
   either a boolean or a compression algorithm (`xz`, `zstd`). `zstd`
   compression is used by default, except CentOS and derivatives up to version
-  8, which default to `xz`. Note that when applied to block device image types,
+  8, which default to `xz`, and OCI images, which default to `gzip`.
```
septatrix, line 670: Before anyone asks, OC**I** **i**mages is correct, as the first abbreviation stands for open container initiative

septatrix:
Is there anything else which has to be addressed? The failing test was flaky at the time of my last push

mkosi/config.py
```diff
@@ -200,6 +202,18 @@ def extension(self) -> str:
             Compression.zstd: ".zst"
         }.get(self, f".{self}")
 
+    def oci_media_type_suffix(self) -> str:
+        suffix = {
+            Compression.none: "",
+            Compression.gz: "+gzip",
+            Compression.zstd: "+zstd",
```
keszybz, line 209: Please align the strings to a column.

mkosi/__init__.py
```diff
@@ -2806,6 +2796,104 @@ def make_disk(
...
+def make_oci(context: Context, root_layer: Path, dst: Path) -> None:
+    ca_store = dst / "blobs" / "sha256"
+    ca_store.mkdir(parents=True)
+
+    layer_diff_digest = hash_file(root_layer)
+    maybe_compress(
+        context,
+        context.config.compress_output,
+        context.staging / "rootfs.layer",
+        context.staging / "rootfs.layer",
```
keszybz, line 2808: If the same arg is given twice on purpose, maybe add a brief comment what it happenning here.

mkosi/__init__.py
```diff
@@ -2806,6 +2796,104 @@ def make_disk(
...
+        if context.config.source_date_epoch
+        else datetime.datetime.now(tz=datetime.timezone.utc)
+    ).isoformat()
+
+    oci_config = {
+        "created": creation_time,
+        "architecture": context.config.architecture.to_oci(),
+        # Name of the operating system which the image is built to run on.
+        # SHOULD use value listed in the golang documentation for $GOOS.
+        # https://github.com/opencontainers/image-spec/blob/v1.0.2/config.md#properties
```
keszybz, line 2826: I was confused by this comment. "SHOULD" makes it sound like something is not implemented here. Maybe just say "Name of the operating system which the image is built to run on as defined by https://github.com/opencontainers/image-spec/blob/v1.0.2/config.md#properties."

keszybz:
> We're trying to get `umoci` packaged in Fedora so then it should be OK to depend on it.

Link?

tests/test_boot.py
```diff
@@ -40,6 +40,31 @@ def test_boot(config: Image.Config, format: OutputFormat) -> None:
             args = ["systemd.mask=systemd-resolved.service"] if format == OutputFormat.directory else []
             image.boot(options=options, args=args)
 
+        if format == OutputFormat.oci:
+            if image.config.distribution in (Distribution.arch, Distribution.centos, Distribution.debian):
+                pytest.skip("These distro's systemd version does not properly run inside the podman")
+
+            # These services fail as /etc/machine-id cannot be properly initialized.
+            # This is not possible as overmounting is not permitted inside the container.
+            # Running the container with higher privileges using `--privileged` or `--cap-add SYS_ADMIN`
```
keszybz, line 49: Please don't use backticks. Backticks only make sense in markdown, and this is not markdown.

DaanDeMeyer:
> > We're trying to get `umoci` packaged in Fedora so then it should be OK to depend on it.
> 
> Link?

No link, Michel Lind has been working on it.

I don't think we want to merge this until we figure out whether we can test this with systemd-nspawn instead of podman. I'd much rather get more coverage on systemd-nspawn's OCI support than having to figure out various podman failures when these tests inevitably start failing.

tests/test_boot.py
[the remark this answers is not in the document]
septatrix: @rhatdan Are you aware of any changes in this area? It seems like the environment which podman provides is no longer sufficient for systemd to run. Should I create an issue for this in the podman repo?

septatrix:
> I don't think we want to merge this until we figure out whether we can test this with systemd-nspawn instead of podman. I'd much rather get more coverage on systemd-nspawn's OCI support than having to figure out various podman failures when these tests inevitably start failing.

Another issue with using umoci would be that it only generates the minimal OCI runtime bundle and [has no means to specify e.g. mounts](https://github.com/opencontainers/umoci/issues/504). The means we would have to set up [all the stuff required by systemd](https://systemd.io/CONTAINER_INTERFACE/) ourselves, likely by modifying the JSON spec generated by umoci. This means that we introduce a new step where we could hide existing errors or add our own ones

mkosi/__init__.py
```diff
@@ -3090,6 +3178,20 @@ def build_image(context: Context) -> None:
...
             sandbox=context.sandbox,
         )
+    elif context.config.output_format == OutputFormat.oci:
+        make_tar(
+            context.root, context.staging / "rootfs.layer",
+            tools=context.config.tools(),
+            sandbox=context.sandbox,
+            # We do not want SELinux xattrs in OCI layers
+            # as they seldom make sense inside containers.
+            options=["--xattrs-exclude", "security.selinux"],
```
DaanDeMeyer, line 3188: People make VMs now from OCI images, so I think excluding selinux is a bit presumptuous. I would prefer the selinux xattrs are stripped out when extracting the OCI image, not when creating it.

DaanDeMeyer (changes requested):
Can we drop the stuff for running OCI images with `podman`? I really don't think we should add that. I'm OK with the OCI output format though.

Also please squash commits, reviewing unsquashed commits is just pure pain.

mkosi/__init__.py
septatrix: umoci logged warnings when extracting an OCI with selinux labels but I can keep them in if that is preferred

septatrix:
I rebased the branch and squashed some commit. I hope the split are done in a way which makes review easier.

I also dropped the podman stuff. This means that only building is tested and not running but until we find a better solution (e.g. umoci, sd-nspawn with a manually created OCI bundle, lxc import with oci-template) this should be fine. We can still revisit this at a later time (e.g. when umoci is included everywhere and then include it in the tools tree).

mkosi/resources/mkosi.md
```diff
@@ -705,11 +706,12 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
```
DaanDeMeyer, line 709: Why default to gzip for OCI images? We generally don't compress by default, I wouldn't do so for OCI images either.

mkosi/__init__.py
```diff
@@ -3261,7 +3261,99 @@ def make_disk(
 
 
 def make_oci(context: Context, root_layer: Path, dst: Path) -> None:
-    raise NotImplementedError
+    ca_store = dst / "blobs" / "sha256"
+    ca_store.mkdir(parents=True)
```
DaanDeMeyer, line 3265: Use `with umask` here to make sure the mode is correct. Should be good with 755 here.

mkosi/__init__.py
```diff
@@ -3260,6 +3260,10 @@ def make_disk(
     return make_image(context, msg=msg, skip=skip, split=split, tabs=tabs, root=context.root, definitions=definitions)
 
 
+def make_oci(context: Context, root_layer: Path, dst: Path) -> None:
```
DaanDeMeyer, line 3263: Just name `root_layer` `root` or so

mkosi/__init__.py
```diff
@@ -3261,7 +3261,99 @@ def make_disk(
...
+            **({
+                "org.opencontainers.image.version": context.config.image_version,
+            } if context.config.image_version else {}),
+        }
+    }
+    oci_manifest_blob = json.dumps(oci_manifest)
+    oci_manifest_digest = hashlib.sha256(oci_manifest_blob.encode()).hexdigest()
+    (ca_store / oci_manifest_digest).write_text(oci_manifest_blob)
+
+    with (dst / "index.json").open("w") as f:
```
DaanDeMeyer, line 3339: Explicit mode here please

mkosi/__init__.py
```diff
@@ -3261,7 +3261,99 @@ def make_disk(
...
+        "annotations": {
+            "io.systemd.mkosi.version": __version__,
+            **({
+                "org.opencontainers.image.version": context.config.image_version,
+            } if context.config.image_version else {}),
+        }
+    }
+    oci_manifest_blob = json.dumps(oci_manifest)
+    oci_manifest_digest = hashlib.sha256(oci_manifest_blob.encode()).hexdigest()
+    (ca_store / oci_manifest_digest).write_text(oci_manifest_blob)
```
DaanDeMeyer, line 3337: `with umask` here to make sure the mode is correct.

mkosi/__init__.py
```diff
@@ -3261,7 +3261,99 @@ def make_disk(
...
+                        "mediaType": "application/vnd.oci.image.manifest.v1+json",
+                        "digest": f"sha256:{oci_manifest_digest}",
+                        "size": (ca_store / oci_manifest_digest).stat().st_size,
+                    }
+                ],
+            },
+            f,
+        )
+
+    with (dst / "oci-layout").open("w") as f:
```
DaanDeMeyer, line 3355: Explicit mode

mkosi/__init__.py
```diff
@@ -3261,7 +3261,99 @@ def make_disk(
...
+                "org.opencontainers.image.version": context.config.image_version,
+            } if context.config.image_version else {}),
+        }
+    }
+    oci_manifest_blob = json.dumps(oci_manifest)
+    oci_manifest_digest = hashlib.sha256(oci_manifest_blob.encode()).hexdigest()
+    (ca_store / oci_manifest_digest).write_text(oci_manifest_blob)
+
+    with (dst / "index.json").open("w") as f:
+        json.dump(
```
DaanDeMeyer, line 3340: Just use `json.dumps` + `Path.write_text()` here please

mkosi/__init__.py
```diff
@@ -3261,7 +3261,99 @@ def make_disk(
...
+                        "digest": f"sha256:{oci_manifest_digest}",
+                        "size": (ca_store / oci_manifest_digest).stat().st_size,
+                    }
+                ],
+            },
+            f,
+        )
+
+    with (dst / "oci-layout").open("w") as f:
+        json.dump({"imageLayoutVersion": "1.0.0"}, f)
```
DaanDeMeyer, line 3356: `Path.write_text()` here as well please

DaanDeMeyer (changes requested):
Looks great, but please make sure all output files and directories are created with explicit permissions. Also, please merge the second and third commit.

mkosi/__init__.py
septatrix: Do you mean `os.umask` or the mode parameter of mkdir?

DaanDeMeyer (approved):

DaanDeMeyer:
@septatrix This would be a separate PR, but given the "layers" stuff in OCI images, it'd be cool if we could integrate this properly with `Overlay=yes` and `BaseTrees=` somehow.

mkosi/__init__.py
```diff
@@ -3270,6 +3260,104 @@ def make_disk(
...
+        "os": "linux",
+        "rootfs": {
+            "type": "layers",
+            "diff_ids": [f"sha256:{layer_diff_digest}"],
+        },
+        "config": {
+            "Cmd": [
+                "/sbin/init",
+                *context.config.kernel_command_line,
+                *context.config.kernel_command_line_extra,
```
DaanDeMeyer, line 3299: Oops, I missed this, please drop this, KernelCommandLineExtra= is for extra stuff to be passed at runtime, not when building the image.

DaanDeMeyer (changes requested):

mkosi/__init__.py
DaanDeMeyer: Fixed this myself

mkosi/__init__.py
```diff
@@ -3270,6 +3260,103 @@ def make_disk(
...
+        context.staging / "rootfs.layer",
+        # Pass explicit destination to suppress adding an extension
+        context.staging / "rootfs.layer",
+    )
+    layer_digest = hash_file(root_layer)
+    root_layer.rename(ca_store / layer_digest)
+
+    creation_time = (
+        datetime.datetime.fromtimestamp(context.config.source_date_epoch, tz=datetime.timezone.utc)
+        if context.config.source_date_epoch
```
behrmann, line 3281: You probably want to check `if context.config.source_date_epoch is not None` here, since otherwise the timestamp 0 is illegal and would lead to the `now` branch.

mkosi/config.py
```diff
@@ -634,6 +670,8 @@ def config_default_compression(namespace: argparse.Namespace) -> Compression:
             return Compression.xz
         else:
             return Compression.zstd
+    elif namespace.output_format == OutputFormat.oci:
+        return Compression.gzip
```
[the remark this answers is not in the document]
septatrix: Zstd support is not yet required by the spec: https://github.com/opencontainers/image-spec/issues/803

behrmann (approved):

cyphar:
It's a bit of a shame I didn't see this thread back when you were discussing using umoci -- I would've been quite happy to help! :smile_cat:

umoci is packaged for most distributions, and it is an optional requirement for LXC and Incus (as well as for KIWI but that's something primarily used by (open)SUSE AFAIK) so anything that packages those probably also packages umoci. It's a very easy program to package (speaking as a package maintainer for openSUSE) so I wouldn't expect that to be a large hurdle to overcome. I guess I should add a reference to https://pkgs.org/download/umoci or similar to the README.

As for the discussion about setting runtime-spec `config.json` parameters -- I would like to know more about what in particular you would like to configure. Most of the things in the [systemd container interface specification](https://systemd.io/CONTAINER_INTERFACE/) are handled by OCI container runtimes and do not require explicit configuration AFAICS.

My original goal was that `umoci unpack` would give you a basic runtime configuration with everything from the image itself populated (such as resource limits, cmdline options, etc etc) and then users could add whatever extra bits they require to provide as much flexibility as possible. Notably [I wrote the specification for converting from image-spec configurations to runtime-spec configurations](https://github.com/opencontainers/image-spec/blob/main/conversion.md).

I felt that adding a bunch of extra configuration knobs would be out of place for a tool for images -- the runtime-spec is quite trivial to modify using `jq` or any other general-purpose JSON library (if you prefer a purpose-built-tool then there's [`oci-runtime-tools`](https://github.com/opencontainers/runtime-tools) but I would personally recommend modifying the JSON directly if what you need is trivial). There is a very large number of possible things that a user might want to configure that it would be unreasonable to add a knob for every one to umoci.

That being said, I am open to discussing what kinds of knobs you might need.

PR #2354, opened 2024-02-08
Add [TriggerMatch] section support

DaanDeMeyer:
These allow specifying that one of multiple sets of conditions should be satisfied for the configuration file to be included.

### DIFF
--- a/mkosi/resources/mkosi.md
+++ b/mkosi/resources/mkosi.md
@@ -313,15 +313,17 @@ configuration files that can still be overridden by specifying the
 setting explicitly via the CLI.
 
 To conditionally include configuration files, the `[Match]` section can
-be used. Matches can use a pipe symbol (`|`) after the equals sign
-(`…=|…`), which causes the match to become a triggering match. The
-config file will be included if the logical AND of all non-triggering
-matches and the logical OR of all triggering matches is satisfied. To
-negate the result of a match, prefix the argument with an exclamation
-mark. If an argument is prefixed with the pipe symbol and an exclamation
-mark, the pipe symbol must be passed first, and the exclamation second.
+be used. A `[Match]` section consists of invididual conditions.
+Conditions can use a pipe symbol (`|`) after the equals sign (`…=|…`),
+which causes the condition to become a triggering condition. The config
+file will be included if the logical AND of all non-triggering
+conditions and the logical OR of all triggering conditions is satisfied.
+To negate the result of a condition, prefix the argument with an
+exclamation mark. If an argument is prefixed with the pipe symbol and an
+exclamation mark, the pipe symbol must be passed first, and the
+exclamation second.
 
-Note that `[Match]` settings match against the current values of
+Note that `[Match]` conditions compare against the current values of
 specific settings, and do not take into account changes made to the
 setting in configuration files that have not been parsed yet. Also note
 that matching against a setting and then changing its value afterwards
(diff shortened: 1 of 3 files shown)


nilskattenbeck-bosch:
This only adds functionality which is already possible using two files and adds nothing new - is this only for deduplication then?

DaanDeMeyer:
@nilskattenbeck-bosch While it's possible to do this with multiple config files + Include= for avoiding duplication, that's rather complex so I'd like to add a simpler interface via TriggerMatch

tests/test_config.py
behrmann, line 1: Another test mixing `[Match]` and `[TriggerMatch]` would be good.

behrmann:
```python
        # Mixing both
        Path("mkosi.conf").write_text(
            """\
            [Match]
            Format=disk

            [TriggerMatch]
            Architecture=arm64

            [TriggerMatch]
            Architecture=x86-64

            [Output]
            ImageId=abcde
            """
        )

        # Match and first TriggerMatch sections match
        _, [config] = parse_config(["--format", "disk", "--architecture=arm64"])
        assert config.image_id == "abcde"

        # Match section matches, but no TriggerMatch section matches
        _, [config] = parse_config(["--format", "disk", "--architecture=s390x"])
        assert config.image_id != "abcde"

        # Second TriggerMatch section matches, but the Match section does not
        _, [config] = parse_config(["--format", "tar", "--architecture=x86-64"])
        assert config.image_id != "abcde"
```

behrmann (approved):
LGTM, but a few comments for more docs and more tests.

The unicode stuff in the docs might seem overkill, but to some people this speaks more than a thousand words. ;)

mkosi/resources/mkosi.md
```diff
@@ -350,6 +352,34 @@ Architecture=|x86-64
...
+
+The semantics of conditions in `[TriggerMatch]` sections is the same as
+in `[Match]`, i.e. all normal conditions are joined by a logical AND and
+all triggering conditions are joined by a logical OR. When mixing
+`[Match]` and `[TriggerMatch]` sections, a match is achieved when all
+`[Match]` sections match and at least one `[TriggerMatch]` section
+matches. No match sections are valued as true. Logically this means:
+
+```
+(∧ᵢ Matchᵢ) ∧ (∨ᵢ TriggerMatchᵢ)
```
septatrix, line 380: Usually one would use the bigger n-ary operators here (⋀ and ⋁) for the expressions in the parentheses.

We could also write this as a proper math section which would be rendered prettier in the markdown preview. I tried this out put sadly the roff output seems to have underlines then - no clue what is going on there:

Rendered man page

Rendered markdown

I just tried out several combinations be the underlines and tildes seem to always be generated by pandoc...

```md
$$
(\wedge_i Match_i) \wedge (\bigvee_i \text{TriggeringMatch}_i)
$$

$$
(⋀_\text{ᵢ} \text{Match}_{i}) ∧ (⋁ᵢ TriggerMatch_i)
$$

```txt
(⋀ᵢ Matchᵢ) ⋀ (⋁ᵢ TriggerMatchᵢ)
```
```

Comment (behrmann) [reply mkosi/resources/mkosi.md:380]: I went for working in the confines of unicode, since that should work in a man page ;) You're right though, that I missed `N-ARY LOGICAL AND` (and or) 
```
(⋀ᵢ Matchᵢ) ∧ (⋁ᵢ TriggerMatchᵢ)
```
they were too small on my screen to see the difference.

Comment (septatrix) [reply mkosi/resources/mkosi.md:380]:
```suggestion
(⋀ᵢ Matchᵢ) ∧ (⋁ᵢ TriggerMatchᵢ)
```

PR #2360, opened 2024-02-09
Have systemd-repart generate fstab and crypttab if requested

DaanDeMeyer:
If systemd-repart is new enough, let's specify --generate-fstab= and --generate-crypttab= so that these files are automatically generated and included in the disk image if the corresponding new settings are used in any partition definition files.

We also make sure systemd-repart always uses the same seed by
generating the random seed ourselves instead of leaving it up to
systemd-repart.

See https://github.com/systemd/systemd/pull/30636.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2653,6 +2653,7 @@ def make_image(
     msg: str,
     skip: Sequence[str] = [],
     split: bool = False,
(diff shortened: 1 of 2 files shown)


behrmann (approved):

keszybz (approved):
LGTM too.

PR #2362, opened 2024-02-09
opensuse: Add python3-pefile to tools tree and default image

DaanDeMeyer:
This dependency was dropped in opensuse's systemd packaging so add it ourselves.

See https://build.opensuse.org/request/show/1144939.


behrmann (approved):

behrmann:
Weird CI failure for CentOS, just shows error, but since it's CentOS it's unrelated to this PR. Merging.

PR #2363, opened 2024-02-09
docs: Add some extra useful defines to rpm build doc

DaanDeMeyer:
These speed up the build so let's make readers aware that they exist.

### DIFF
--- a/docs/building-rpms-from-source.md
+++ b/docs/building-rpms-from-source.md
@@ -148,6 +148,13 @@ env --chdir=mkosi \
     ${BUILDDIR:+--define} \
     ${BUILDDIR:+"_vpath_builddir $BUILDDIR"} \
     --define "_build_name_fmt %%{NAME}-%%{VERSION}-%%{RELEASE}.%%{ARCH}.rpm" \
+    --define "_binary_payload w.ufdio" \
+    --define "debug_package %{nil}" \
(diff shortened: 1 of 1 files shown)


docs/building-rpms-from-source.md
```diff
@@ -148,6 +148,13 @@ env --chdir=mkosi \
     ${BUILDDIR:+--define} \
     ${BUILDDIR:+"_vpath_builddir $BUILDDIR"} \
     --define "_build_name_fmt %%{NAME}-%%{VERSION}-%%{RELEASE}.%%{ARCH}.rpm" \
+    --define "_binary_payload w.ufdio" \
+    --define "debug_package %{nil}" \
+    --define "__brp_strip %{nil}" \
+    --define "__brp_compress %{nil}" \
+    --define "__brp_mangle_shebangs %{nil}" \
+    --define "__brp_strip_comment_note %{nil}" \
+    --define "__brp_strip_static_archive %{nil}" \
```
keszybz, line 157: This changes the output rpms a bit. In particular, `__brp_mangle_shebangs` has an important effect. But for well-written packages, it shouldn't matter. So I guess it's OK to do this.

I guess that if we don't build the debug packages, then it's better to *not* strip the binaries, so that debugging works nicely.

Issue #2367, opened 2024-02-10
Imcompatible image name with machinectl (v2)


iam-TJ:
Despite PR #1739 "Machine name fixes" (in response to Issue #1224) v20.2 still suffers the problem of using an underscore when suffixing `mkosi.version` to container names causing an invalid machine name:
```
$ cat mkosi.version
0.11
$ machinectl list-images | grep '\.10'
debian-testing-amd64_0.10 subvolume no       Sat 2024-02-10 16:13:07 GMT -
$ machinectl start debian-testing-amd64_0.10
Invalid machine name debian-testing-amd64_0.10.
```

PR #2368, opened 2024-02-10
Do not use underscore in image names

iam-TJ:
When suffixing image_version (mkosi.version) do not use an underscore since that breaks systemd-machined/machinectl because undescore are invalid in machine names.

This is related to commit 7eb4ea8

Closes: #2367

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -588,7 +588,7 @@ def config_default_output(namespace: argparse.Namespace) -> str:
     output = namespace.image_id or namespace.image or "image"
 
     if namespace.image_version:
-        output += f"_{namespace.image_version}"
+        output += f"-{namespace.image_version}"
 
     return output
 
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -896,7 +896,7 @@ def test_output_id_version(tmp_path: Path) -> None:
     with chdir(d):
         _, [config] = parse_config()
 
-    assert config.output == "output_1.2.3"
+    assert config.output == "output-1.2.3"
 
 
 def test_deterministic() -> None:


DaanDeMeyer (approved):

septatrix:
Now this is again incompatible with the [version format required by DDI](https://uapi-group.org/specifications/specs/discoverable_disk_image/#image-version) which mandates underscores for separating the version from the rest :/ Are we gonna ping-pong this back and forth?

behrmann:
That DDI names are incompatible with machined is something that should be fixed there, but generally we should produce valid DDI names.

DaanDeMeyer:
Right let's revert this.

iam-TJ:
So how to fix it so `systemd-machined` gets valid names?

septatrix:
> So how to fix it so `systemd-machined` gets valid names?

I am not sure about `machinectl` but I assume it allows you to override the machine name on import. Alternatively you can change the output name with `Output=` or simply rename the file

iam-TJ:
> > So how to fix it so `systemd-machined` gets valid names?
> 
> I am not sure about `machinectl` but I assume it allows you to override the machine name on import. Alternatively you can change the output name with `Output=` or simply rename the file

There is no "import" - `mkosi` generates the directory directly in `/var/lib/machines`  (as aBTRFS subvol) and when using `mkosi.version`  it uses an invalid name.

Originally this worked fine then around v12 or v14 it was broken by using the underscore; I reported bugs and fixes for it back then.

I've revised the patch to be conditional on `output_format == OutputFormat.directory` so it should not affect DDIs, only directory/subvol output.

septatrix:
> There is no "import" - `mkosi` generates the directory directly in `/var/lib/machines` (as aBTRFS subvol) and when using `mkosi.version` it uses an invalid name.

Moving the generated images directly into the machines folder seems like a very specific edge-case.

> I've revised the patch to be conditional on `output_format == OutputFormat.directory` so it should not affect DDIs, only directory/subvol output.

That would invalidate other assumptions such as globs matching all artifacts generated by mkosi for a specific image (e.g. `rm mkosi.output/MyOS_1.*`). Does using `Output=` not work for you?

DaanDeMeyer:
@iam-TJ Please use `Output=image-%v` or something similar to use a hyphen instead of an underscore

Issue #1993, opened 2023-10-19
Add way to easily persist CLI configuration


DaanDeMeyer:
It'd be very useful if we could "remember" configuration passed via CLI between running multiple commands.

DaanDeMeyer:
```
Daan De Meyer, [10/19/23 9:51 AM]
So this just came to mind

Daan De Meyer, [10/19/23 9:52 AM]
What about adding a —persist option to write the final configuration as JSON to mkosi.json in the cwd

Daan De Meyer, [10/19/23 9:53 AM]
And then until —persist is specified again, whenever mkosi runs again, we use the config from mkosi.json merged with CLI arguments

Daan De Meyer, [10/19/23 9:56 AM]
Wait forget about persisting the final configuration

Daan De Meyer, [10/19/23 9:56 AM]
We'd persist the CLI configuration

Daan De Meyer, [10/19/23 9:58 AM]
What this solves

Daan De Meyer, [10/19/23 9:58 AM]
Is that currently, when I share mkosi one liners

Daan De Meyer, [10/19/23 9:58 AM]
They have to keep the full CLI at hand

Daan De Meyer, [10/19/23 9:58 AM]
You can't build the image

Daan De Meyer, [10/19/23 9:59 AM]
And then run it with mkosi qemu

Daan De Meyer, [10/19/23 9:59 AM]
Because mkosi qemu needs the full config as well

Daan De Meyer, [10/19/23 9:59 AM]
But if we had a —persist

J B, [10/19/23 9:59 AM]
I'm good with that

Daan De Meyer, [10/19/23 9:59 AM]
You'd only need the full CLI once

Daan De Meyer, [10/19/23 10:00 AM]
I wonder if this should be a verb or a CLI option

J B, [10/19/23 10:00 AM]
Why not both?

J B, [10/19/23 10:01 AM]
Nah, scratch that, just a verb

Daan De Meyer, [10/19/23 10:02 AM]
The thing is you might want to either nuke or extend the existing mkosi.json

J B, [10/19/23 10:03 AM]
Nuking could be done via mkosi persist --force and extending would just be doing CLI arguments, no?

Daan De Meyer, [10/19/23 10:03 AM]
Yeah I was thinking we'd reuse the existing —force argument

J B, [10/19/23 10:04 AM]
and if you'd want to extend the json config via config files we could add a hydrate verb to write out a config from JSON

Daan De Meyer, [10/19/23 10:04 AM]
Do we have a better name available than "persist"?

[...]

Issue #1112, opened 2022-08-10
Support reproducible builds


Foxboron:
Looking into supporting `mkosi` for https://system-transparency.org/ but we need support for reproducible builds.

Currently `mkosi` support `--finalize-script` for image cleanups, but there are none distributed as part of `mkosi` project, but I'm wondering if we should contemplate having a set of cleanup scripts to help make images reproducible as part of the project? I think it could be done by having `--reproducible` as an alias for `--finalize-script=reproducible-builds.$distro`, or something along those lines.

`mmdebstrap` effectively has a little bit of code to just rm files which has timestamps and similar embedded.
https://gitlab.mister-muffin.de/josch/mmdebstrap/src/branch/main/mmdebstrap#L2948

Relevant issues around Reproducible Builds; https://github.com/systemd/mkosi/issues/700 https://github.com/systemd/mkosi/issues/687

behrmann:
Like the idea, but I guess an additional step in `build_image`, maybe just after `run_finalize_script`:
```diff
run_finalize_script
+if args.reproducible:
+    make_reproducible(args, root, do_run_build_script, for_cache)
```
for some implementation of `make_reproducible` would be more in line with the rest (and I think right now we only support a single finalize script).

Foxboron:
A little bit of work alter and I have gotten reproducible Debian images with `mkosi`. A couple of hacks and some stuff that should be polished, but hopefully it inspires a bit!

I'll send a few pull-requests :)

```bash
(.venv) λ mkosi-test » sudo mkosi -d debian --finalize-script=reproducible --no-manifest -o debian.cpio.xz -t cpio --compress-output=xz build
[.....[
‣   Running finalize script…
+ rm -f /var/tmp/mkosi-lgzfxrx8/root/var/log/dpkg.log
+ rm -f /var/tmp/mkosi-lgzfxrx8/root/var/log/bootstrap.log
+ rm -f /var/tmp/mkosi-lgzfxrx8/root/var/cache/apt/pkgcache.bin
+ rm -f /var/tmp/mkosi-lgzfxrx8/root/var/log/apt/history.log
+ rm -f /var/tmp/mkosi-lgzfxrx8/root/var/log/apt/term.log
+ rm -f /var/tmp/mkosi-lgzfxrx8/root/var/log/alternatives.log
+ rm -f /var/tmp/mkosi-lgzfxrx8/root/var/cache/ldconfig/aux-cache
[...]

PR #2372, opened 2024-02-10
Issue 1993 persist cli config

iam-TJ:
A first pass that extracts the valid command-line configuration items and reports it via `logging.info()` thus:
```
‣ CLI:  --distribution=debian --release=testing --image_id=debian-testing-amd64
```
If this is suitable then persisting to the file system could be added; possibly (optionally) writing  a `mkosi.conf.d/` file?


mkosi/config.py
```diff
@@ -3174,8 +3174,10 @@ def finalize_defaults(namespace: argparse.Namespace, defaults: argparse.Namespac
         finalize_defaults(namespace, defaults)
         images = [namespace]
 
+    cli_valid = dict({key: getattr(cli_ns, key) for key in vars(cli_ns)})
```
septatrix, line 3177: Maybe a bit cleaner:

```py
cli_valid = {
    key: value
    for key, value in vars(ns).items()
    if key not in SETTINGS_LOOKUP_BY_DEST
}
```

mkosi/config.py
```diff
@@ -3194,6 +3196,14 @@ def finalize_defaults(namespace: argparse.Namespace, defaults: argparse.Namespac
             hint="Prefix the setting with '@' in the image configuration file to allow overriding it from the command line.", # noqa: E501
         )
 
+    commandline=""
+    for key in cli_valid:
+        value = cli_valid[key]
+        commandline+=f" --{key}={value}"
```
septatrix, line 3202: The way this is written results in a double space `CLI:  --foob=bar`.
Alternative which eliminates the preceeding space: `" ".join(f"--{key}={value}" for key, value in cli_valid.items())`

septatrix:
> If this is suitable then persisting to the file system could be added; possibly (optionally) writing a `mkosi.conf.d/` file?

Maybe this could be combined with #1112 and they could somehow be stored in the manifest? This might be a bit overkill for only remembering the flags from the last invocation but we could kill two birds with one stone here

DaanDeMeyer:
I added the discussion me and @behrmann had about this a while ago to #1993. I hope that's sufficient to show that a basic implementation like this is not going to cut it.

mkosi/config.py
```diff
@@ -3194,6 +3196,14 @@ def finalize_defaults(namespace: argparse.Namespace, defaults: argparse.Namespac
...
+
+    if len(commandline) > 0:
```
keszybz, line 3204: Just `if commandline`.

DaanDeMeyer:
Let's close this for now until there's activity again from the author

PR #2373, opened 2024-02-10
Add support for signing with OpenSSL Engines

bluca:


mkosi/__init__.py
```diff
@@ -789,23 +789,27 @@ def sign_efi_binary(context: Context, input: Path, output: Path) -> None:
...
+                ]
+            options = [
                         "--ro-bind", context.config.secure_boot_certificate, context.config.secure_boot_certificate,
                         "--ro-bind", input, input,
                     ]
-                ),
-            )
+            if context.config.secure_boot_key_uri:
+                cmd += [
+                        "--engine", context.config.secure_boot_key_uri[:context.config.secure_boot_key_uri.find(":")],
```
behrmann, line 803:
```suggestion
                        "--engine", context.config.secure_boot_key_uri.split(":")[0],
```
though I'd maybe prefer splitting that beforehand and having some handling if there is no colon in there.

mkosi/__init__.py
```diff
@@ -789,23 +789,27 @@ def sign_efi_binary(context: Context, input: Path, output: Path) -> None:
...
+                        "--key", context.config.secure_boot_key_uri,
```
behrmann, line 804: That's the whole thing with what's before the colon?

behrmann (commented):
The Ci failure is mostly overly long lines, which we check with `ruff` now.

mkosi/__init__.py
bluca: yes, this needs to include the prefix, pkcs11:somethingsomething

mkosi/__init__.py
```diff
@@ -789,22 +789,31 @@ def sign_efi_binary(context: Context, input: Path, output: Path) -> None:
         find_binary("sbsign", root=context.config.tools()) is not None
     ):
         with open(output, "wb") as f:
-            run(
-                [
+            cmd: list[Union[Path, str]] = [
```
DaanDeMeyer, line 792: Use `PathString` here instead `Union[Path, str]`

mkosi/__init__.py
```diff
@@ -789,22 +789,31 @@ def sign_efi_binary(context: Context, input: Path, output: Path) -> None:
...
-                    "--key", context.config.secure_boot_key,
                     "--cert", context.config.secure_boot_certificate,
                     "--output", "/dev/stdout",
-                    input,
-                ],
-                stdout=f,
-                sandbox=context.sandbox(
-                    options=[
-                        "--ro-bind", context.config.secure_boot_key, context.config.secure_boot_key,
+                ]
```
DaanDeMeyer, line 796: Indentation is off

mkosi/__init__.py
```diff
@@ -789,22 +789,31 @@ def sign_efi_binary(context: Context, input: Path, output: Path) -> None:
...
+            options: list[Union[Path, str]] = [
```
DaanDeMeyer, line 797: Ditto for `PathString`

mkosi/__init__.py
DaanDeMeyer: One annoying thing is that if a uri used that takes a path this will fail as we won't mount it into the sandbox. But let's just require users to use the other option for that for now.

mkosi/__init__.py
```diff
@@ -789,22 +789,31 @@ def sign_efi_binary(context: Context, input: Path, output: Path) -> None:
...
+            elif context.config.secure_boot_key:
+                cmd += ["--key", context.config.secure_boot_key]
+                options += ["--ro-bind", context.config.secure_boot_key, context.config.secure_boot_key]
+            cmd += [input]
+            run(
+                cmd,
+                stdout=f,
+                sandbox=context.sandbox(
+                    options=options,
+                    devices=context.config.secure_boot_key_uri is not None
```
DaanDeMeyer, line 815: You can just use `bool(...)`

mkosi/__init__.py
```diff
@@ -922,26 +931,39 @@ def install_systemd_boot(context: Context) -> None:
             # We reuse the key for all secure boot databases to keep things simple.
             for db in ["PK", "KEK", "db"]:
                 with umask(~0o600), open(keys / f"{db}.auth", "wb") as f:
-                    run(
-                        [
+                    cmd: list[Union[Path, str]] = [
```
DaanDeMeyer, line 934: `PathString`

mkosi/__init__.py
```diff
@@ -922,26 +931,39 @@ def install_systemd_boot(context: Context) -> None:
...
                             "--output", "/dev/stdout",
-                            db,
-                            context.workspace / "mkosi.esl",
-                        ],
-                        stdout=f,
-                        sandbox=context.sandbox(
-                            options=[
-                                "--ro-bind", context.config.secure_boot_key, context.config.secure_boot_key,
+                    ]
+                    options: list[Union[Path, str]] = [
```
DaanDeMeyer, line 941: `PathString`

mkosi/__init__.py
```diff
@@ -922,26 +931,39 @@ def install_systemd_boot(context: Context) -> None:
...
                                 context.config.secure_boot_certificate,
                                 context.config.secure_boot_certificate,
                                 "--ro-bind", context.workspace / "mkosi.esl", context.workspace / "mkosi.esl",
-                            ],
+                    ]
+                    if context.config.secure_boot_key_uri:
+                        cmd += [
+                            "--engine", context.config.secure_boot_key_uri.split(":")[0],
+                            "--key", context.config.secure_boot_key_uri,
+                            ]
```
DaanDeMeyer, line 951: Indentation

mkosi/__init__.py
```diff
@@ -922,26 +931,39 @@ def install_systemd_boot(context: Context) -> None:
...
+                    elif context.config.secure_boot_key:
+                        cmd += [
+                            "--key", context.config.secure_boot_key,
+                        ]
```
DaanDeMeyer, line 955: Can be one line

mkosi/__init__.py
```diff
@@ -922,26 +931,39 @@ def install_systemd_boot(context: Context) -> None:
...
+                        options += ["--ro-bind", context.config.secure_boot_key, context.config.secure_boot_key]
+                    cmd += [
+                            db,
+                            context.workspace / "mkosi.esl",
+                    ]
```
DaanDeMeyer, line 960: Indentation

mkosi/__init__.py
```diff
@@ -922,26 +931,39 @@ def install_systemd_boot(context: Context) -> None:
...
+                    run(
+                        cmd,
+                        stdout=f,
+                        sandbox=context.sandbox(
+                            options=options,
+                            devices=context.config.secure_boot_key_uri is not None,
```
DaanDeMeyer, line 966: `bool(....)`

mkosi/__init__.py
```diff
@@ -1784,23 +1806,30 @@ def build_uki(
...
             ]
             options += [
-                "--ro-bind", context.config.secure_boot_key, context.config.secure_boot_key,
                 "--ro-bind", context.config.secure_boot_certificate, context.config.secure_boot_certificate,
             ]
+            if context.config.secure_boot_key_uri:
+                cmd += [
+                        "--signing-engine", context.config.secure_boot_key_uri.split(":")[0],
+                        "--secureboot-private-key", context.config.secure_boot_key_uri,
+                    ]
```
DaanDeMeyer, line 1827: Indentation

mkosi/__init__.py
```diff
@@ -1784,23 +1806,30 @@ def build_uki(
...
+            elif context.config.secure_boot_key:
+                cmd += [
+                        "--secureboot-private-key", context.config.secure_boot_key,
+                    ]
```
DaanDeMeyer, line 1831: Can be one line

mkosi/__init__.py
```diff
@@ -1835,7 +1865,7 @@ def build_uki(
         options += ["--ro-bind", initrd, initrd]
 
     with complete_step(f"Generating unified kernel image for kernel version {kver}"):
-        run(cmd, sandbox=context.sandbox(options=options))
+        run(cmd, sandbox=context.sandbox(options=options, devices=context.config.secure_boot_key_uri is not None))
```
DaanDeMeyer, line 1868: `bool(...)`

mkosi/__init__.py
```diff
@@ -2702,7 +2734,10 @@ def make_image(
                 cmdline,
                 stdout=subprocess.PIPE,
                 env=context.config.environment,
-                sandbox=context.sandbox(devices=not context.config.repart_offline, options=options),
+                sandbox=context.sandbox(
+                    devices=not context.config.repart_offline or context.config.verity_key_uri is not None,
```
DaanDeMeyer, line 2738: `bool(...)`

mkosi/__init__.py
```diff
@@ -2872,7 +2909,10 @@ def make_extension_image(context: Context, output: Path) -> None:
         run(
             cmdline + ["--definitions", r],
             env=env,
-            sandbox=context.sandbox(devices=not context.config.repart_offline, options=options),
+            sandbox=context.sandbox(
+                devices=not context.config.repart_offline or context.config.verity_key_uri is not None,
```
DaanDeMeyer, line 2913: `bool(...)`

DaanDeMeyer (changes requested):
Looks great, just code formatting nits

bluca:
Need to rework this, not all engines use URIs, the tpm engine for exmaple just takes files, but needs to be explicitly loaded

bluca:
Requires https://github.com/systemd/systemd/pull/31274

DaanDeMeyer (changes requested):
So let's expose the same interface that's used in systemd now: `VerityKeySource` and `SecureBootKeySource`. I'm fine with these being stored as strings that are just passed through directlly to the corresponding systemd tool.

bluca:
rearranged the options to match, but I don't know how to do the enum stuff

mkosi/__init__.py
```diff
@@ -1910,11 +1928,22 @@ def build_uki(
         )
 
         if sign_expected_pcr:
-            cmd += [
-                "--pcr-private-key", context.config.secure_boot_key,
-                "--pcr-banks", "sha1,sha256",
-            ]
-            options += ["--ro-bind", context.config.secure_boot_key, context.config.secure_boot_key]
+            cmd += ["--pcr-banks", "sha1,sha256"]
```
behrmann, line 1931: Didn't we just change swtpm to only have sha256 banks, could this be an issue?

mkosi/config.py
```diff
@@ -462,8 +464,12 @@ def parse_path(value: str,
...
         path = path.expanduser()
 
-    if required and not path.exists():
-        die(f"{value} does not exist")
+    if not path.exists():
+        if required:
+            die(f"{value} does not exist")
+        # If this is could be either a path or a string, and it doesn't exist, assume it's a string and return it.
+        if pathstring:
+            return Path(original)
```
behrmann, line 472: The name `pathstring` for this option seems a bit obscure, mostly because of our type `PathString`. Maybe a name like `assume_exists`? That being said the comment is a bit unclear. We don't need to assume something is a string, we know the value we have coming in and even use it as `original`. 

But looking at the only two users, I'm not sure I understand why this is needed. The users are:
```python
parse=config_make_path_parser(secret=True, required=False, pathstring=True),
```
but `secret=True` will not touch `path` if it doesn't exist, `absolute` is false by default and `resolve`, which is true by default, should be a noop for a non-existing path, so `required=False` should already suffice to skip this. The only difference then, is that this returns the path without expanding environment variables or the username, after expanding them and seeing that it doesn't exist. If it's about that a name a name for the option like `unexpand_vars_and_user_when_nonexistent` (or anything that's less of a mouthful)  might be better. 

Or just an option `passthrough` doing just that: returning `Path(value)` at the earliest possible moment.

mkosi/__init__.py
bluca: dunno, but the banks are not changed by this PR as you can see in the diff

mkosi/config.py
bluca: I've renamed to assume_exists - for the rest, this was necessary to stop all the type checking malarky from complaining at me. If you want it restructured in some way please feel free to push a change to the branch, after ensuring the CI is happy with it

mkosi/__init__.py
```diff
@@ -1910,11 +1928,22 @@ def build_uki(
...
+                    context.config.secure_boot_key_source.type == KeySource.Type.engine and
+                    context.config.secure_boot_certificate
+                ):
+                    cmd += [
+                        "--signing-engine", context.config.secure_boot_key_source.source,
+                        "--pcr-public-key", context.config.secure_boot_certificate,
+                    ]
+                    options += [
+                        "--ro-bind", context.config.secure_boot_certificate, context.config.secure_boot_certificate,
+                    ]
```
DaanDeMeyer, line 1944: This part confuses me. What does the public key have to do with the signing engine?

DaanDeMeyer: @bluca Ok I hadn't noticed that we interpret `--pcr-public-key` as a certificate if `--signing-engine` is specified.

I did notice in the ukify source code that you forgot to convert the certificate to a public key for storage in the `pcrpkey` section which expects a public key and not a certificate, quote from the docs:

>A .pcrpkey section with a public key in the PEM format matching the signature data in the the .pcrsig section.

bluca: That should be handled by this bit?

https://github.com/systemd/systemd/blob/main/src/boot/measure.c#L861

DaanDeMeyer: @bluca That's for systemd-measure, I'm talking about this bit:

```python
    pcrpkey = opts.pcrpkey
    if pcrpkey is None:
        if opts.pcr_public_keys and len(opts.pcr_public_keys) == 1:
            pcrpkey = opts.pcr_public_keys[0]
        elif opts.pcr_private_keys and len(opts.pcr_private_keys) == 1:
            from cryptography.hazmat.primitives import serialization
            privkey = serialization.load_pem_private_key(pathlib.Path(opts.pcr_private_keys[0]).read_bytes(), password=None)
            pcrpkey = privkey.public_key().public_bytes(
                encoding=serialization.Encoding.PEM,
                format=serialization.PublicFormat.SubjectPublicKeyInfo,
            )

    sections = [
        # name,      content,         measure?
        ('.osrel',   opts.os_release, True ),
        ('.cmdline', opts.cmdline,    True ),
        ('.dtb',     opts.devicetree, True ),
        ('.uname',   opts.uname,      True ),
        ('.splash',  opts.splash,     True ),
        ('.pcrpkey', pcrpkey,         True ),
        ('.initrd',  initrd,          True ),

        # linux shall be last to leave breathing room for decompression.
        # We'll add it later.
    ]
```

When using `--signing-engine`, `.pcrpkey` will now contain a certificate instead of a public key.

bluca: Ah yes you are right, will fix that

mkosi/config.py
behrmann: As far as I understand this piece, I see two reasonable things here
```python
def parse_path(value: str,
               *,
               required: bool = True,
               resolve: bool = True,
               expanduser: bool = True,
               expandvars: bool = True,
               secret: bool = False,
               absolute: bool = False,
               pass_through: bool = False,
               constants: Sequence[str] = ()) -> Path:
    if value in constants:
        return Path(value)

    if pass_through:
        return Path(value)

    # rest omitted
```
and
```python
def parse_path(value: str,
               *,
               required: bool = True,
               resolve: bool = True,
               expanduser: bool = True,
               expandvars: bool = True,
               secret: bool = False,
               absolute: bool = False,
               assume_exists: bool = False,
               constants: Sequence[str] = ()) -> Path:
    if value in constants:
        return Path(value)

    original = value

    if expandvars:
        value = os.path.expandvars(value)

    path = Path(value)

    if expanduser:
        if path.is_relative_to("~") and not INVOKING_USER.is_running_user():
            path = INVOKING_USER.home() / path.relative_to("~")
        path = path.expanduser()

    if assume_exists:
        return path
    # rest omitted
```

Depending on whether expanding environment variables or `~` is meant to be possible or not. Either type checked for me on a local checkout of your branch.

bluca: please pick your favorite and just push it

DaanDeMeyer (changes requested):
The documentation and help strings of `VerityKey` and `SecureBootKey` should be updated to mention that the format depends on the value of `VerityKeySource` and `SecureBootKeySource`

bluca:
> The documentation and help strings of `VerityKey` and `SecureBootKey` should be updated to mention that the format depends on the value of `VerityKeySource` and `SecureBootKeySource`

Updated

mkosi/__init__.py
```diff
@@ -858,22 +859,29 @@ def sign_efi_binary(context: Context, input: Path, output: Path) -> None:
...
+                "--output", "/dev/stdout",
+            ]
+            options: list[PathString] = [
+                "--ro-bind", context.config.secure_boot_certificate, context.config.secure_boot_certificate,
+                "--ro-bind", input, input,
+            ]
+            if context.config.secure_boot_key_source.type == KeySource.Type.engine:
+                cmd += ["--engine", context.config.secure_boot_key_source.source]
+            if context.config.secure_boot_key_source.type == KeySource.Type.file or \
+                    Path(context.config.secure_boot_key).exists():
```
DaanDeMeyer, line 875: Please drop the exists() check here. If the key source is configured as `file` we should fail if the file doesn't exist instead of silently not adding specific options.

mkosi/__init__.py
```diff
@@ -991,26 +999,35 @@ def install_systemd_boot(context: Context) -> None:
...
+                    options: list[PathString] = [
+                        "--ro-bind",
+                        context.config.secure_boot_certificate,
+                        context.config.secure_boot_certificate,
+                        "--ro-bind", context.workspace / "mkosi.esl", context.workspace / "mkosi.esl",
+                    ]
+                    if context.config.secure_boot_key_source.type == KeySource.Type.engine:
+                        cmd += ["--engine", context.config.secure_boot_key_source.source]
+                    if context.config.secure_boot_key_source.type == KeySource.Type.file or \
+                            Path(context.config.secure_boot_key).exists():
```
DaanDeMeyer, line 1019: Ditto please drop `exists()` check

mkosi/__init__.py
DaanDeMeyer: If mypy complains just add an assert

mkosi/__init__.py
```diff
@@ -991,26 +999,35 @@ def install_systemd_boot(context: Context) -> None:
...
+                        options += ["--ro-bind", context.config.secure_boot_key, context.config.secure_boot_key]
+                    cmd += [
+                        db,
+                        context.workspace / "mkosi.esl",
+                    ]
```
DaanDeMeyer, line 1024: Make this one line

mkosi/__init__.py
```diff
@@ -1887,9 +1904,13 @@ def build_uki(
                 context.config.secure_boot_certificate,
             ]
             options += [
-                "--ro-bind", context.config.secure_boot_key, context.config.secure_boot_key,
                 "--ro-bind", context.config.secure_boot_certificate, context.config.secure_boot_certificate,
             ]
+            if context.config.secure_boot_key_source.type == KeySource.Type.engine:
+                cmd += ["--signing-engine", context.config.secure_boot_key_source.source]
+            if context.config.secure_boot_key_source.type == KeySource.Type.file or \
+                    Path(context.config.secure_boot_key).exists():
```
DaanDeMeyer, line 1912: Ditto please drop `exists()` check

mkosi/config.py
```diff
@@ -2309,12 +2344,20 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
     ),
     ConfigSetting(
         dest="secure_boot_key",
-        metavar="PATH",
+        metavar="KEY",
         section="Validation",
-        parse=config_make_path_parser(secret=True),
+        parse=config_parse_key,
         paths=("mkosi.key",),
         help="UEFI SecureBoot private key in PEM format",
```
DaanDeMeyer, line 2352: Please update help text as well to be a bit more generic

mkosi/__init__.py
```diff
@@ -2817,7 +2817,10 @@ def make_image(
         options += ["--ro-bind", context.config.passphrase, context.config.passphrase]
     if context.config.verity_key:
         cmdline += ["--private-key", context.config.verity_key]
-        options += ["--ro-bind", context.config.verity_key, context.config.verity_key]
+        if context.config.verity_key_source.type != KeySource.Type.file:
+            cmdline += ["--private-key-source", str(context.config.verity_key_source)]
+        if Path(context.config.verity_key).exists():
```
DaanDeMeyer, line 2834: Please replace this check with a check if the key source type == file

mkosi/__init__.py
```diff
@@ -2995,7 +3004,10 @@ def make_extension_image(context: Context, output: Path) -> None:
...
```
DaanDeMeyer, line 3021: Ditto

mkosi/__init__.py
```diff
@@ -1931,11 +1931,23 @@ def build_uki(
...
+                ):
+                    cmd += [
+                        "--signing-engine", context.config.secure_boot_key_source.source,
+                        "--pcr-public-key", context.config.secure_boot_certificate,
+                    ]
+                    options += [
+                        "--ro-bind", context.config.secure_boot_certificate, context.config.secure_boot_certificate,
+                    ]
+                if context.config.secure_boot_key_source.type == KeySource.Type.file or \
+                        Path(context.config.secure_boot_key).exists():
```
DaanDeMeyer, line 1949: Please drop `exists()` check

DaanDeMeyer (changes requested):

mkosi/config.py
```diff
@@ -482,6 +482,13 @@ def parse_path(value: str,
     return path
 
 
+def config_parse_key(value: Optional[str], old: Optional[str]) -> Optional[PathString]:
+    if not value:
+        return None
+
+    return parse_path(value, secret=True) if Path(value).exists() else value
```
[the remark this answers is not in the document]
DaanDeMeyer: I actually don't like this. `PathString` is a much better type here since when an engine is used it is not a path but a string.

mkosi/__init__.py
bluca: The key might be a file even if the source is an engine, for example when using the tpm2 engine, you just pass a key file to it, so it needs to be bind mounted, otherwise it will fail

DaanDeMeyer: @bluca In that case please do `isinstance(context.config.secure_boot_key, os.PathLike)`  then instead

mkosi/config.py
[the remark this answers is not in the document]
behrmann: Ah, took me a while to grok this. You're right. I'd like something else than a string flying out of this, but for now a string is the correct thing.

mkosi/config.py
```diff
@@ -994,6 +1001,33 @@ def config_parse_minimum_version(value: Optional[str], old: Optional[GenericVers
...
+def config_parse_key_source(value: Optional[str], old: Optional[KeySource]) -> Optional[KeySource]:
+    if not value:
+        return old
+
+    try:
+        type = KeySource.Type(value.split(":")[0])
+    except ValueError:
+        die(f"'{value}' is not a valid key source")
+
+    source = value.split(":")[1] if len(value.split(":")) > 1 else ""
```
behrmann, line 1026:
```suggestion
    typ, _, source = value.partition(":")
    try:
        type = KeySource.Type(typ)
    except ValueError:
        die(f"'{value}' is not a valid key source")
```

behrmann (commented):
The needed `str()` calls look like a bug in mypy. I've opened https://github.com/python/mypy/issues/16985

DaanDeMeyer:
`PathString` json serialization is lossy so I changed back to `exists()` checks instead.

PR #2376, opened 2024-02-11
Fix base tree tar archive unpacking

### LINKED ISSUE #2374: Sysext example doesn't build with `Format=tar`
When attempting to build the sysext example, but changing to `Format=tar` (and `CompressOutput=false` and adjusting `BaseTrees` of `btrfs`) mkosi exits with the following error:

```
‣  Creating tar archive /var/tmp/mkosi-workspaceo_jtq2be/staging/base.tar…
‣  /home/nekko/Overlay/mkosi.output/base.tar size is 136.3M, consumes 136.3M.
‣ Building btrfs image
‣  Mounting base trees…
‣   Extracting tar archive /home/nekko/Overlay/mkosi.output/base.tar…
bwrap: Can't find source path /var/tmp/mkosi-workspacei7ho9ng6/bases/base.tar-e64a5ae4b49c47d4b299acd1ba3f3380: No such file or directory
bwrap: Can't find source path /var/tmp/mkosi-workspacei7ho9ng6/bases/base.tar-e64a5ae4b49c47d4b299acd1ba3f3380: No such file or directory
‣ "tar --extract --file - --directory /var/tmp/mkosi-workspacei7ho9ng6/bases/base.tar-e64a5ae4b49c47d4b299acd1ba3f3380 --keep-directory-symlink --no-overwrite-dir --same-permissions --numeric-owner --same-order --acls --selinux --xattrs --force-local --exclude './dev/*' --exclude './proc/*' --exclude './sys/*' --exclude './tmp/*' --exclude './run/*' --exclude './var/tmp/*'" returned non-zero exit code 1.
```

This does not happen when removing `Overlay=true`

DaanDeMeyer:
Fixes #2374

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -108,7 +108,7 @@ def mount_base_trees(context: Context) -> Iterator[None]:
                 extract_tar(
                     path, d,
                     tools=context.config.tools(),
-                    sandbox=context.sandbox(options=["--bind", d, d]),
+                    sandbox=context.sandbox(options=["--ro-bind", path, path, "--bind", d.parent, d.parent]),
                 )
                 bases += [d]
             elif path.suffix == ".raw":


mkosi/__init__.py
```diff
@@ -108,7 +108,7 @@ def mount_base_trees(context: Context) -> Iterator[None]:
                 extract_tar(
                     path, d,
                     tools=context.config.tools(),
-                    sandbox=context.sandbox(options=["--bind", d, d]),
+                    sandbox=context.sandbox(options=["--ro-bind", path, path, "--bind", d.parent, d.parent]),
```
behrmann, line 111: Why do we have to bind all bases writable here?

DaanDeMeyer: Because the target directory does not exist yet and is created by `tar` so we have to mount the parent directory so it can be created.

behrmann (approved):

PR #2377, opened 2024-02-11
tree-wide: Introduce SandboxProtocol

DaanDeMeyer:
Instead of passing a full sandbox command into the functions from
tree.py, archive.py, kmod.py and partition.py, let's instead pass
in a function that creates a sandbox, so we can pass in the required
options from the functions themselves. This reduces duplication a lot
as we don't have to specify all the sandbox options at each callsite.


mkosi/installer/pacman.py
```diff
@@ -49,7 +49,12 @@ def mounts(cls, context: Context) -> list[PathString]:
             # installed packages. The former should go in the cache directory, the latter should go in the image,
             # so we bind mount the local directory from the image to make sure that happens.
             "--bind", context.root / "var/lib/pacman/local", "/var/lib/pacman/local",
-        ] if (context.root / "var/lib/pacman/local").exists() else [])
+        ] if (context.root / "var/lib/pacman/local").exists() else []) + ([
+            "--ro-bind", context.config.tools() / "etc/makepkg.conf", "/etc/makepkg.conf",
+        ] if (
+            (context.config.tools() / "etc/makepkg.conf").exists() and
+            not (context.pkgmngr / "etc/makepkg.conf").exists()
+        ) else [])
```
behrmann, line 57: I think this might be too much chaining for readability, how about
```python
    @classmethod
    def mounts(cls, context: Context) -> list[PathString]:
        mounts = [
            *super().mounts(context),
            # pacman writes downloaded packages to the first writable cache directory. We don't want it to write to our
            # local repository directory so we expose it as a read-only directory to pacman.
            "--ro-bind", context.packages, "/var/cache/pacman/mkosi",
        ]
        if (context.root / "var/lib/pacman/local").exists():
            # pacman reuses the same directory for the sync databases and the local database containing the list of
            # installed packages. The former should go in the cache directory, the latter should go in the image,
            # so we bind mount the local directory from the image to make sure that happens.
            mounts += [
            "--bind", context.root / "var/lib/pacman/local", "/var/lib/pacman/local",
        ]
        if (
            (context.config.tools() / "etc/makepkg.conf").exists() and
            not (context.pkgmngr / "etc/makepkg.conf").exists()
        ):
            mounts += [
            "--ro-bind", context.config.tools() / "etc/makepkg.conf", "/etc/makepkg.conf",
        ]
        return mounts
```

behrmann (approved):

DaanDeMeyer:
Added one more commit to force ndb db rpm backend on opensuse

PR #2378, opened 2024-02-11
Do not use underscore in image names for directory

iam-TJ:
For OutputFormat.directory when suffixing image_version (mkosi.version) do not use an underscore since that breaks systemd-machined/machinectl because underscore is invalid in machine names.

This is related to commit 7eb4ea8

Closes: #2367

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -588,7 +588,13 @@ def config_default_output(namespace: argparse.Namespace) -> str:
     output = namespace.image_id or namespace.image or "image"
 
     if namespace.image_version:
-        output += f"_{namespace.image_version}"
+        # Discoverable Disk Image (DDI) compatible separator
+        separator = "_"
+        if namespace.output_format == OutputFormat.directory:
+            # systemd-machined compatible separator
+            separator = "-"
+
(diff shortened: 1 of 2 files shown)


DaanDeMeyer:
@iam-TJ This is not the way to go, as I mentioned in the other PR, you can set `Output=` to customize the output filename. Set it to `Output=image-%v` to get the image version in the output name without an underscore.

iam-TJ:
That doesn't work nicely when different Formats (via `--format`)  for the same configuration might be used since it special cases directory output names for machined.

My entire experience over this  hyphen/underscore issue since ~v14 suggests that mkosi no longer wants to support machined containers as first-class citizens.

behrmann:
> That doesn't work nicely when different Formats (via --format) for the same configuration might be used since it special cases directory output names for machined.

Is something preventing you from setting it for all output types?

> My entire experience over this hyphen/underscore issue since ~v14 suggests that mkosi no longer wants to support machined containers as first-class citizens.

As mentioned previously, mkosi is the wrong place to fix this, this needs fixing in machined. The problem is that there are two competing, mutually exclusive needs until this is fixed in machined. No one has gotten around to that yet, unfortunately.

PR #2383, opened 2024-02-13
Fail early with a clear error if keyring does not exist

DaanDeMeyer:


DaanDeMeyer:
CI failure is unrelated

behrmann (approved):
I'm not necessarily a fan of `listify`, but let's get this out the door first.

PR #2386, opened 2024-02-13
docs: Make "Building rpms from source" non-dnf specific

DaanDeMeyer:
Let's make the doc non-dnf specific by not relying on dnf builddep
and using mkosi-install to install packages. This allows using the
same logic for opensuse images.

We also simplify things by only installing --buildrequires since
trying to cache --requires from the rpm spec isn't very useful as
most of the --requires dependencies are automatically generated and
won't be listed by rpmspec --requires in the first place.

### DIFF
--- a/docs/building-rpms-from-source.md
+++ b/docs/building-rpms-from-source.md
@@ -11,11 +11,10 @@ If you want to build an RPM from source and install it within a mkosi
 image, you can do that with mkosi itself without using `mock`. The steps
 required are as follows:
 
-1. Install `Requires` dependencies in the image
-2. Install `BuildRequires` dependencies in the build overlay
-3. Install dynamic `BuildRequires` dependencies in the build overlay
-4. Build the RPM with `rpmbuild`
-5. Install the built rpms in the image
+1. Install `BuildRequires` dependencies in the build overlay
+1. Install dynamic `BuildRequires` dependencies in the build overlay
+1. Build the RPM with `rpmbuild`
+1. Install the built rpms in the image
 
 In the following examples, we'll use mkosi itself and its Fedora RPM
 spec as an example.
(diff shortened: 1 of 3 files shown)


docs/building-rpms-from-source.md
```diff
@@ -62,62 +61,76 @@ The prepare script `mkosi.prepare` then looks as follows:
...
-        EXIT_STATUS=$?
-        if [ $EXIT_STATUS -ne 11 ]; then
-            exit $EXIT_STATUS
-        fi
-
-        dnf builddep /var/tmp/SRPMS/mkosi-*.buildreqs.nosrc.rpm
-    done
-fi
+    --define "_sourcedir rpm" \
+    rpm/mkosi.spec | sort --unique >/tmp/buildrequires
```
behrmann, line 75: Just for completeness sake, since I don't do RPM: Does this query multiple specs here or can the same package be required multiple times? Asking because of the `--unique` here, not that it huts.

docs/building-rpms-from-source.md
```diff
@@ -62,62 +61,76 @@ The prepare script `mkosi.prepare` then looks as follows:
...
+    rpmbuild \
+    -bd \
+    --build-in-place \
+    --define "_topdir /var/tmp" \
+    --define "_sourcedir rpm" \
+    --define "_build_name_fmt %%{NAME}-%%{VERSION}-%%{RELEASE}.%%{ARCH}.rpm" \
+    rpm/mkosi.spec
+do
+    EXIT_STATUS=$?
+    if [ $EXIT_STATUS -ne 11 ]; then
```
behrmann, line 90: I assume 11 is the return code for missing build requirements, a comment for that here in the docs (but not necessarily in the actual prepare script) would be nice.

behrmann (approved):
Shell-wise LGTM, whether it makes sense RPM-wise is for someone more qualified to judge.

docs/building-rpms-from-source.md
DaanDeMeyer: This is just for safety, since the algorithm below breaks if the same dependency is added multiple times from the same rpm.

docs/building-rpms-from-source.md
DaanDeMeyer: We already mention this below

PR #2398, opened 2024-02-16
Fedora rawhide key kerfuffle workaround

keszybz:
This makes mkosi not fail for rawhide right now.

### DIFF
--- a/mkosi/distributions/fedora.py
+++ b/mkosi/distributions/fedora.py
@@ -14,7 +14,20 @@ from mkosi.installer import PackageManager
 from mkosi.installer.dnf import Dnf
 from mkosi.installer.rpm import RpmRepository, find_rpm_gpgkey, setup_rpm
 from mkosi.log import die
-from mkosi.util import listify
+from mkosi.util import listify, tuplify
+
+
+@tuplify
+def find_fedora_rpm_gpgkeys(context: Context) -> Iterable[str]:
+    key1 = find_rpm_gpgkey(context, key=f"RPM-GPG-KEY-fedora-{context.config.release}-primary")
+    key2 = find_rpm_gpgkey(context, key=f"RPM-GPG-KEY-fedora-{context.config.release}-secondary")
+
+    if key1:
+        yield key1
+    if key2:
+        yield key2
+    if not key1 and not key2:
+        yield "https://fedoraproject.org/fedora.gpg"
 
 
 class Installer(DistributionInstaller):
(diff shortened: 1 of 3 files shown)


mkosi/distributions/fedora.py
```diff
@@ -14,7 +16,33 @@
...
+
+@tuplify
+def find_fedora_rpm_gpgkeys(context: Context) -> Iterable[str]:
+    key1 = find_rpm_gpgkey(context, key=f"RPM-GPG-KEY-fedora-{context.config.release}-primary")
+    key2 = find_rpm_gpgkey(context, key=f"RPM-GPG-KEY-fedora-{context.config.release}-secondary")
+
+    if key1:
+        # During branching, there is always a kerfuffle with the key transition.
+        # For Rawhide, try to load the N+1 key, just in case our local configuration
+        # still indicates that Rawhide==N, but really Rawhide==N=1.
```
behrmann, line 30:
```suggestion
        # still indicates that Rawhide==N, but really Rawhide==N+1.
```

mkosi/distributions/fedora.py
```diff
@@ -14,7 +16,33 @@
...
+        if context.config.release == "rawhide" and key1.startswith("file://"):
+            path = Path(key1[7:]).resolve()
```
behrmann, line 32: I think
```suggestion
            path = Path(key1.removeprefix("file://")).resolve()
```
documents the intent more clearly than the `[7:]` and we don't need the performance here.

mkosi/distributions/fedora.py
```diff
@@ -14,7 +16,33 @@
...
+            if m := re.match(r"RPM-GPG-KEY-fedora-(\d+)-(primary|secondary)", path.name):
+                version = int(m.group(1))
+                if key3 := find_rpm_gpgkey(context, key=f"RPM-GPG-KEY-fedora-{version + 1}-primary"):
+                    # We yield the resolved path for key1, to make it clear that it's
+                    # for version N, and the other key is for version N+1.
+                    key1 = f"file://{path}"
```
behrmann, line 38:
```suggestion
                    key1 = path.as_uri()
```

mkosi/distributions/fedora.py
```diff
@@ -74,12 +102,8 @@ def remove_packages(cls, context: Context, packages: Sequence[str]) -> None:
     @classmethod
     @listify
     def repositories(cls, context: Context) -> Iterable[RpmRepository]:
-        gpgurls = (
-            find_rpm_gpgkey(
-                context,
-                key=f"RPM-GPG-KEY-fedora-{context.config.release}-primary",
-            ) or "https://fedoraproject.org/fedora.gpg",
-        )
+        gpgurls = find_fedora_rpm_gpgkeys(context)
```
behrmann, line 105: I think I'd forego introducing `tuplify` and just use `listify` and just call `tuple` on it here.

behrmann (commented):
LGTM, a few minor things

mkosi/distributions/fedora.py
keszybz: It'd be great if `removeprefix` returned None, because then one could collapse the whole thing into the conditional. Unfortunately they chose a different interface.

mkosi/distributions/fedora.py
keszybz: I think it's important to hide such ugly details… You define `tuplify` once and then can have nicer code in multiple places.

behrmann (approved):

PR #2402, opened 2024-02-18
Decouple UKI copying from UKI building

DaanDeMeyer:
Let's make sure that even if we don't build any UKIs, we still copy out any existing ones.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1902,6 +1902,9 @@ def install_uki(context: Context, partitions: Sequence[Partition]) -> None:
     if not stub.exists() and context.config.bootable == ConfigFeature.auto:
         return
 
+    if context.config.bootable == ConfigFeature.enabled and not gen_kernel_images(context):
+        die("A bootable image was requested but no kernel was found")
+
     roothash = finalize_roothash(partitions)
 
     for kver, kimg in gen_kernel_images(context):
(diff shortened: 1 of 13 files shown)


mkosi/__init__.py
```diff
@@ -363,11 +363,7 @@ def finalize_scripts(scripts: Mapping[str, Sequence[PathString]] = {}) -> Iterat
                     DIR="$(cd "$(dirname "$0")" && pwd)"
                     PATH="$(echo "$PATH" | tr ':' '\\n' | grep -v "$DIR" | tr '\\n' ':')"
                     export PATH
-                    if [ $# -gt 0 ]; then
-                        exec {shlex.join(str(s) for s in script)} "$@"
-                    else
-                        exec {shlex.join(str(s) for s in script)} sh -i
-                    fi
+                    exec {shlex.join(str(s) for s in script)} "$@"
```
behrmann, line 366: Maybe the functionality could be kept for mkosi-chroot
```suggestion
                    """ + ("""exec {shlex.join(str(s) for s in script)} "$@"""" if script != "mkosi-chroot" else """
                    if [ $# -gt 0 ]; then
                        exec {shlex.join(str(s) for s in script)} "$@"
                    else
                        exec {shlex.join(str(s) for s in script)} sh -i
                    fi)
```

behrmann (approved):

mkosi/__init__.py
DaanDeMeyer: Too much of a hack if you ask me

behrmann (approved):

PR #2418, opened 2024-02-22
Add support for sync scripts

DaanDeMeyer:
Sync scripts allow updating various sources automatically before
doing a build.


mkosi/__init__.py
```diff
@@ -2221,7 +2299,13 @@ def check_inputs(config: Config) -> None:
...
                 die(f"Initrd {p} is not a file")
 
-    for script in config.prepare_scripts + config.build_scripts + config.postinst_scripts + config.finalize_scripts:
+    for script in (
+        config.sync_scripts +
+        config.prepare_scripts +
+        config.build_scripts +
+        config.postinst_scripts +
+        config.finalize_scripts
+    ):
```
behrmann, line 2308: Possible alternative
```suggestion
    for script in itertools.chain(
        config.sync_scripts,
        config.prepare_scripts,
        config.build_scripts,
        config.postinst_scripts,
        config.finalize_scripts,
    ):
```

mkosi/resources/mkosi.md
```diff
@@ -1832,6 +1839,14 @@ are mounted into the current working directory before running the script
 in the current working directory. `$SRCDIR` is set to point to the
 current working directory. The following scripts are supported:
 
+* If **`mkosi.sync`** (`SyncScripts=`) exists, it is executed before the
+  image is built. This script may be used to update various sources that
+  are used to build the image. One use case it to run `git pull` on
```
behrmann, line 1844:
```suggestion
  are used to build the image. One use case is to run `git pull` on
```

mkosi/resources/mkosi.md
```diff
@@ -1832,6 +1839,14 @@ are mounted into the current working directory before running the script
...
+  various source repositories before building the image. Specifically,
+  the `BuildSourcesEphemeral=` setting does not apply to sync scripts,
+  and so sync scripts can be used to update build sources even if
```
behrmann, line 1847:
```suggestion
  so sync scripts can be used to update build sources even if
```
or
```suggestion
  and sync scripts can be used to update build sources even if
```
or
```suggestion
  so that sync scripts can be used to update build sources even if
```

mkosi/installer/apt.py
```diff
@@ -174,7 +175,10 @@ def invoke(
         mounts: Sequence[PathString] = (),
         stdout: _FILE = None,
     ) -> CompletedProcess:
-        with finalize_ephemeral_source_mounts(context.config) as sources:
+        with finalize_source_mounts(
+            context.config,
+            ephemeral=os.getuid() == 0 and context.config.build_sources_ephemeral,
```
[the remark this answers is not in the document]
DaanDeMeyer: I think we should probably share more logic between the invoke() methods since they're very similar, but I'll leave that for another PR

mkosi/__init__.py
```diff
@@ -351,47 +345,71 @@ def mount_build_overlay(context: Context, volatile: bool = False) -> Iterator[Pa
...
+            os.chmod(Path(d) / name, 0o755)
             os.utime(Path(d) / name, (0, 0))
 
         yield Path(d)
 
 
+GIT_COMMAND = (
+    "git",
+    "-c",
+    "safe.directory=*",
```
keszybz, line 382: Please put the argument to `-c` on the same line.

mkosi/__init__.py
```diff
@@ -351,47 +345,71 @@ def mount_build_overlay(context: Context, volatile: bool = False) -> Iterator[Pa
...
+        f"--reuid={INVOKING_USER.uid}",
+        f"--regid={INVOKING_USER.gid}",
+        "--clear-groups",
+    )
+
+
 def finalize_host_scripts(
     context: Context,
-    helpers: Mapping[str, Sequence[PathString]],
+    helpers: Mapping[str, Sequence[PathString]] = {},
```
keszybz, line 397: Hmm, using a mutable object as the default is considered bad practice. I think this should be None or ().

keszybz (approved):

mkosi/__init__.py
DaanDeMeyer: @keszybz The `Mapping` type is immutable. The reason using a mutable object is bad practice is because it could be mutated. However, the type is immutable here which means it cannot be mutated so it's fine. We use this pattern all over the place so let's not try to fix this here.

PR #2420, opened 2024-02-22
Reuse existing scripts for mkosi-install

DaanDeMeyer:
Otherwise we'll expand the full command line twice, once as part of the mkosi-install script, in which apt-get is expanded again.

### DIFF
--- a/mkosi/installer/apt.py
+++ b/mkosi/installer/apt.py
@@ -3,7 +3,7 @@ import os
 import textwrap
 from collections.abc import Iterable, Sequence
 from pathlib import Path
-from typing import NamedTuple, Optional
+from typing import NamedTuple, Optional, cast
 
 from mkosi.config import Config
 from mkosi.context import Context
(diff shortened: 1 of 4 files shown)


mkosi/installer/apt.py
```diff
@@ -62,12 +62,12 @@ def scripts(cls, context: Context) -> dict[str, list[PathString]]:
                 "apt-mark",
                 "apt-sortpkgs",
             )
-        } | {
-            "mkosi-install"  : apivfs_cmd(context.root) + cls.cmd(context, "apt-get") + ["install"],
-            "mkosi-upgrade"  : apivfs_cmd(context.root) + cls.cmd(context, "apt-get") + ["upgrade"],
-            "mkosi-remove"   : apivfs_cmd(context.root) + cls.cmd(context, "apt-get") + ["purge"],
-            "mkosi-reinstall": apivfs_cmd(context.root) + cls.cmd(context, "apt-get") + ["install", "--reinstall"],
-        }
+        } | cast(dict[str, list[PathString]], {
```
behrmann, line 65: Can we get a todo here noting that this is another workaround for pyright?

behrmann (approved):

mkosi/installer/apt.py
DaanDeMeyer: Will do a follow up

PR #2425, opened 2024-02-23
Use --keep-directory-symlink from cp 9.5 onwards

### LINKED ISSUE #2168: Copying in trees from build script fails with overwriting symlinks
This is the same issue as https://github.com/systemd/mkosi/issues/166 but as it reappeared, I thought I'd open a new bug report.

I'm trying to build `sbctl` in an mkosi container. The output provides `/usr/local/share/man/man8/sbctl.8`, which is outputted into `DESTDIR` in `mkosi.build.chroot` in my case. `mkosi` then tries to install that file to the regular tree, which fails:

```
+ tree /work/dest
/work/dest
├── local_scratch
│   └── aur_packages
│       ├── pacman-hook-kernel-install.pkg.tar.zst
│       └── paru.pkg.tar.zst
└── usr
    └── local
        ├── bin
        │   └── sbctl
        ├── lib
        │   └── kernel
        │       └── install.d
        │           └── 91-sbctl.install
        └── share
            ├── bash-completion
            │   └── completions
            │       └── sbctl
            ├── fish
            │   └── vendor_completions.d
            │       └── sbctl.fish
            ├── licenses
            │   └── sbctl
            │       └── LICENSE
            ├── man
            │   └── man8
            │       └── sbctl.8
            └── zsh
                └── site-functions
                    └── _sbctl

20 directories, 9 files
+ tree //usr/local
//usr/local
├── bin
├── etc
├── games
├── include
├── lib
├── man
├── sbin
├── share
│   └── man -> ../man
└── src

11 directories, 0 files
‣  Copying in build tree…
cp: cannot overwrite non-directory '/home/cnhof/.cache/mkosi-workspace62czz5gn/root/usr/local/share/man' with directory '/home/cnhof/.cache/mkosi-workspace62czz5gn/dest/usr/local/share/man'
‣ "cp --recursive --no-dereference --preserve=mode,timestamps,links,xattr,ownership --reflink=auto /home/cnhof/.cache/mkosi-workspace62czz5gn/dest /home/cnhof/.cache/mkosi-workspace62czz5gn/root --no-target-directory" returned non-zero exit code 1.
```

### LINKED ISSUE #2334: mkosi removes required iwlwifi firmware files
When filtering kernel modules with the following configuration:

```
KernelModulesExclude=.*
KernelModulesIncludeHost=true
KernelModulesInclude=
   iwlwifi
```

`mkosi` keeps the `iwlwifi` module and its firmware files, but still removes the `iwlwifi-*.pnvm.zst` files associated w/ the firmware files.  

I have no idea what these files are, and they don't appear in the firmware dependencies of `iwlwifi` (i.e. `modinfo iwlwifi` doesn't show them).  But `iwlwifi` needs them, and crashes when booting the generated image if those files aren't there.

I'm not sure it's possible to retain these files automatically given that they're not listed as firmware dependencies, but there doesn't seem to be any configuration option to exclude firmware files from the kernel module filter.

I'm currently working around this with

```
rsync -mar --include '*/' --include 'iwlwifi-*.pnvm.*' --exclude '*' \
            "${BUILDROOT}/usr/lib/firmware/" \
            "${BUILDROOT}/usr/lib/saved-firmware"
```

in `mkosi.prepare final`, and the reverse 

```
rsync -mar "${BUILDROOT}/usr/lib/saved-firmware/" \
            "${BUILDROOT}/usr/lib/firmware"
rm -rf "${BUILDROOT}/usr/lib/saved-firmware/"
```

in `mkosi.finalize`.

It'd be nice though if mkosi could retain these files automatically, or offer some customization option to retain certain firmware files.

That said, I'm fine with the workaround, so feel free to close this issue if it's out of scope for mkosi.

DaanDeMeyer:
--keep-directory-symlink instructs cp to not fail when trying to copy a directory onto a symlink but to follow the symlink instead.

The patch to introduce it was merged into coreutils and will be available from coreutils 9.5 onwards.

--copy-contents has to be added as well to make
--keep-directory-symlink work. --copy-contents is generally harmless for our use cases and won't change anything.

Fixes #2168

### DIFF
--- a/mkosi/tree.py
+++ b/mkosi/tree.py
@@ -14,6 +14,7 @@ from mkosi.run import find_binary, run
 from mkosi.sandbox import SandboxProtocol, nosandbox
 from mkosi.types import PathString
 from mkosi.util import flatten
+from mkosi.versioncomp import GenericVersion
 
 
 def statfs(path: Path, *, sandbox: SandboxProtocol = nosandbox) -> str:
@@ -23,6 +24,10 @@ def statfs(path: Path, *, sandbox: SandboxProtocol = nosandbox) -> str:
 
 def is_subvolume(path: Path, *, sandbox: SandboxProtocol = nosandbox) -> bool:
     return path.is_dir() and statfs(path, sandbox=sandbox) == "btrfs" and path.stat().st_ino == 256
+
+
+def cp_version() -> GenericVersion:
+    return GenericVersion(run(["cp", "--version"], stdout=subprocess.PIPE).stdout.splitlines()[0].split()[3])
 
 
 def make_tree(
@@ -92,8 +97,12 @@ def copy_tree(
         "--dereference" if dereference else "--no-dereference",
         f"--preserve=mode,links{',timestamps,ownership,xattr' if preserve else ''}",
         "--reflink=auto",
+        "--copy-contents",
         src, dst,
     ]
+    if cp_version() >= "9.5":
+        copy += ["--keep-directory-symlink"]
+
     options: list[PathString] = ["--ro-bind", src, src, "--bind", dst.parent, dst.parent]
 
     # If the source and destination are both directories, we want to merge the source directory with the
--- a/mkosi/sandbox.py
+++ b/mkosi/sandbox.py
@@ -203,6 +203,8 @@ def apivfs_cmd(root: Path) -> list[PathString]:
         "--dev", root / "dev",
         # Make sure /etc/machine-id is not overwritten by any package manager post install scripts.
         "--ro-bind-try", root / "etc/machine-id", root / "etc/machine-id",
+        # Nudge gpg to create its sockets in /run by making sure /run/user/0 exists.
+        "--dir", root / "run/user/0",
         *finalize_passwd_mounts(root),
         "sh", "-c",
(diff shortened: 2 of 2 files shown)


mkosi/tree.py
```diff
@@ -25,6 +26,10 @@ def is_subvolume(path: Path, *, sandbox: SandboxProtocol = nosandbox) -> bool:
     return path.is_dir() and statfs(path, sandbox=sandbox) == "btrfs" and path.stat().st_ino == 256
 
 
+def cp_version() -> GenericVersion:
+    return GenericVersion(run(["cp", "--version"], stdout=subprocess.PIPE).stdout.strip().splitlines()[0].split()[3])
```
behrmann, line 30: I think the first split is a bit superfluous, since it only strips the trailing newline. If the first line had whitespace at the beginning that would not affect the splitlines and would be cleaned up by the second split as well.
```suggestion
    return GenericVersion(run(["cp", "--version"], stdout=subprocess.PIPE).stdout.splitlines()[0].split()[3])
```

behrmann (approved):

Cornelicorn:
The commit message still mentions the incorrect issue :)

PR #2432, opened 2024-02-28
remove grub2 package on s390x fedora

lysliu:


mkosi/resources/mkosi-tools/mkosi.conf.d/10-centos-fedora/mkosi.s390x.conf
behrmann, line 1: Please don't double this file. Remove `grub2-tools` from the above file and don't add the additional matches there, but add a new file in `10-centos-fedora/mkosi.conf.d`. Have a look at `10-uefi.conf` there. If what you need is a `HostArchitetecture` match (which seems likely to me), you might just add `grub2-tools` there.

behrmann (changes requested):

mkosi/resources/mkosi-tools/mkosi.conf.d/10-centos-fedora/mkosi.s390x.conf
lysliu: Thanks, I moved the `grub2-tools` to the `10-uefi.conf` file.
since it already use the `HostArchitecture` match there.

DaanDeMeyer:
This needs to be a new file that only applies if the target architecture is x86.

DaanDeMeyer (changes requested):
Please remove the fedora specific dropin since the centos-fedora one applies to fedora as well already.

DaanDeMeyer (approved):

PR #2434, opened 2024-02-28
fix sync when running as sudo and finalize when image lacks packaging directories

richardmaw-codethink:
Recent-ish changes to package cache behaviour broke creating sysexts with finalize scripts for me and the repo sync changes made running with sudo fail for me.

These changes make things work, though it's possible it may be better to change finalize's context to have the same package cache dir as earlier steps rather than falling back to `root / "var"`, and while `rchown_package_manager_dirs` works to fix the permissions of directories created by the package manager it causes the logs about cleaning up permissions to be emitted multiple times and it may be strange to have it running nested inside itself.

### DIFF
--- a/mkosi/installer/__init__.py
+++ b/mkosi/installer/__init__.py
@@ -41,7 +41,7 @@ class PackageManager:
 
         for d in ("cache", "lib"):
             src = context.package_cache_dir / d / subdir
-            mounts += ["--bind", src, Path("/var") / d / subdir]
+            mounts += ["--bind-try", src, Path("/var") / d / subdir]
 
             # If we're not operating on the configured package cache directory, we're operating on a snapshot of the
             # repository metadata in the image root directory. To make sure any downloaded packages are still cached in
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3741,7 +3741,8 @@ def run_sync(args: Args, config: Config, *, resources: Path) -> None:
         install_package_manager_trees(context)
         context.config.distribution.setup(context)
 
-        sync_repository_metadata(context)
+        with rchown_package_manager_dirs(config):
+            sync_repository_metadata(context)
 
         src = config.package_cache_dir_or_default() / "cache" / subdir
         for p in config.distribution.package_manager(config).cache_subdirs(src):
--- a/mkosi/installer/dnf.py
+++ b/mkosi/installer/dnf.py
@@ -30,6 +30,8 @@ class Dnf(PackageManager):
 
     @classmethod
     def cache_subdirs(cls, cache: Path) -> list[Path]:
+        if not cache.exists():
+            return []
         return [
             p / "packages"
             for p in cache.iterdir()


mkosi/installer/__init__.py
```diff
@@ -41,6 +41,8 @@ def mounts(cls, context: Context) -> list[PathString]:
 
         for d in ("cache", "lib"):
             src = context.package_cache_dir / d / subdir
+            if not src.exists():
```
behrmann, line 44: I think `--bind-try` would be the thing to use for that case:
> Equal to `--bind` but ignores non-existant SRC

I vaguely remember @DaanDeMeyer mentioning, that CentOS8's bubblewrap was to old for that, but according to pkgs.org that has version 0.4 and I can find it in the man page for version 0.3.1, so it might be safe.

richardmaw-codethink: `--ro-bind-try` is already in use, so I'll rework this to use `--bind-try`.

richardmaw-codethink: It's specifically dnf assuming that the directory exists that causes the failure first. Other package managers only need it to be a virtual path and append a subdir to it.

The later part where it additionally binds the default package cache dir in checks whether the directory exists first, so shouldn't need `--bind-try`.


```diff
diff --git a/mkosi/installer/__init__.py b/mkosi/installer/__init__.py
index 3ee2f903..d86c17c8 100644
--- a/mkosi/installer/__init__.py
+++ b/mkosi/installer/__init__.py
@@ -41,9 +41,7 @@ class PackageManager:
 
         for d in ("cache", "lib"):
             src = context.package_cache_dir / d / subdir
-            if not src.exists():
-                continue
-            mounts += ["--bind", src, Path("/var") / d / subdir]
+            mounts += ["--bind-try", src, Path("/var") / d / subdir]
 
             # If we're not operating on the configured package cache directory, we're operating on a snapshot of the
             # repository metadata in the image root directory. To make sure any downloaded packages are still cached in
diff --git a/mkosi/installer/dnf.py b/mkosi/installer/dnf.py
index e74fd48c..0e5f0875 100644
--- a/mkosi/installer/dnf.py
+++ b/mkosi/installer/dnf.py
@@ -30,6 +30,8 @@ class Dnf(PackageManager):
 
     @classmethod
     def cache_subdirs(cls, cache: Path) -> list[Path]:
+        if not cache.exists():
+            return []
         return [
             p / "packages"
             for p in cache.iterdir()
```

richardmaw-codethink:
[CI / integration-test (debian, ubuntu)][] failed with `[  104.462358] systemd[1]: Failed to start systemd-machine-id-commit.service - Commit a transient machine-id on disk.`
I've seen `(debian, arch)` fail this way too so I'm assuming it's just a bit flaky.

[CI / integration-test (debian, ubuntu)]: https://github.com/systemd/mkosi/actions/runs/8113317119/job/22176414700?pr=2434

DaanDeMeyer:
@richardmaw-codethink I've opened #2444 as an alternative to your first commit. Can you try and see if that fixes the issue for you? I prefer running things as the right user when possible over chowning.

mkosi/installer/__init__.py
DaanDeMeyer: Sorry for the churn but I actually prefer the initial implementation in this case. The command lines are already long enough as they are no need to make them even longer.

DaanDeMeyer (changes requested):
So in `clean_package_manager_metadata()`, we copy the repository metadata from the image to the workspace to make sure it stays available. So I'm a bit confused about the second commit since when we run the finalize scripts we're supposed to use the repository metadata from the workspace and not from the image.

DaanDeMeyer:
Both issues have been addressed in #2444 and #2460 so closing this one

PR #2438, opened 2024-03-01
set default unknown os type

lysliu:
set default unknown os type

Fix issue: #2442

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -144,11 +144,17 @@ class KernelType(StrEnum):
             logging.warning("bootctl doesn't know kernel-identify verb, assuming 'unknown' kernel type")
             return KernelType.unknown
 
-        type = run(
-            ["bootctl", "kernel-identify", path],
-            stdout=subprocess.PIPE,
-            sandbox=config.sandbox(options=["--ro-bind", path, path]),
-        ).stdout.strip()
+        try:
+            type = run(
+                ["bootctl", "kernel-identify", path],
+                stdout=subprocess.PIPE,
+                sandbox=config.sandbox(options=["--ro-bind", path, path]),
+            ).stdout.strip()
+        except subprocess.CalledProcessError as e:
+            logging.warning(
+                f"'bootctl kernel-identify' returned a exit code '{e.returncode}' for '{path}', assume type 'unknown'"
+            )
+            return KernelType.unknown
 
         try:
             return cls(type)


mkosi/qemu.py
```diff
@@ -144,19 +144,22 @@ def identify(cls, config: Config, path: Path) -> "KernelType":
...
+            # If an error occurs, set a default value for 'type'
+            type = "unknown"
 
         try:
             return cls(type)
         except ValueError:
             logging.warning(f"Unknown kernel type '{type}', assuming 'unknown'")
             return KernelType.unknown
 
-
```
behrmann, line 159: Please don't remove this line.

mkosi/qemu.py
```diff
@@ -144,19 +144,22 @@ def identify(cls, config: Config, path: Path) -> "KernelType":
...
-            stdout=subprocess.PIPE,
-            sandbox=config.sandbox(options=["--ro-bind", path, path]),
-        ).stdout.strip()
+        try:
+            type = run(
+                ["bootctl", "kernel-identify", path],
+                stdout=subprocess.PIPE,
+                sandbox=config.sandbox(options=["--ro-bind", path, path]),
+            ).stdout.strip()
+        except Exception:
```
behrmann, line 153: No blanket `except Exception`, please. You probably want
```suggestion
        except subprocess. CalledProcessError:
```
or even better
```suggestion
        except subprocess. CalledProcessError as e:
```

mkosi/qemu.py
```diff
@@ -144,19 +144,22 @@ def identify(cls, config: Config, path: Path) -> "KernelType":
...
+            # If an error occurs, set a default value for 'type'
+            type = "unknown"
```
behrmann, line 155: You can directly return here
```suggestion
            return KernelType.unknown
```
or even better with
```suggestion
            logging.warning(f"'bootctl kernel-identify' returned a exit code {e.returncode} for '{path}', assuming type 'unknown'")
            return KernelType.unknown
```
assuming the exception above as saved with `as e`

lysliu: Break the logging warning to 2 lines to resolve 
```

mkosi/qemu.py:155:120: E501 Line too long (131 > 119)
```

mkosi/qemu.py
```diff
@@ -144,11 +144,17 @@ def identify(cls, config: Config, path: Path) -> "KernelType":
...
+        try:
+            type = run(
+                ["bootctl", "kernel-identify", path],
+                stdout=subprocess.PIPE,
+                sandbox=config.sandbox(options=["--ro-bind", path, path]),
+            ).stdout.strip()
+        except subprocess. CalledProcessError as e:
+            # If an error occurs, set a default value for 'type'
+            logging.warning(f"'bootctl kernel-identify' returned a exit code {e.returncode} for '{path}'")
+            logging.warning(f"'assuming type 'unknown'")
```
behrmann, line 156: Instead of logging to separate lines
```suggestion
            logging.warning(
                f"'bootctl kernel-identify' returned a exit code {e.returncode} for '{path}' assuming type 'unknown'"
            )
```

lysliu:
> Looks sensible now codewise, but, maybe I've missed it, I don't recall an issue that this solves in particular, i.e. does it make sense to default to `unknown` if bootctl can't tell what we're talking about? I assume this is some s390x issue and that architecture presumably not using EFI?

linked to issue: #2442

DaanDeMeyer:
@lysliu I don't see how this would fix #2442. I've prepared #2445 which will fix #2442. Please show the exact error that `bootctl kernel-identify` fails with.

DaanDeMeyer (changes requested):
I don't see how this would fix the issue. Please show exactly how `bootctl` fails. I've also prepared #2445 which will fix #2442.

DaanDeMeyer:
Let's close this, I don't think we should ignore every failure from bootctl. Please open an issue with the exact error you get from bootctl if you still see this happening.

PR #2440, opened 2024-03-01
Add support for embedding devicetree into UKI

### LINKED ISSUE #2439: Allow specifying a devicetree binary that is passed to ukify
To easily use the dtbs provided with kernel packages allow specifying a devicetree name in mkosi.

It can be passed as parameter to ukify. The dtb is typically installed by the kernel package.

manut:
Resolves #2439

To easily use the dtbs provided with kernel packages allow specifying a devicetree name in mkosi.

It can be passed as parameter to ukify.
The dtb is typically installed by the kernel package.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1872,6 +1872,11 @@ def build_uki(
         "--ro-bind", stub, stub,
     ]
 
+    if context.config.devicetree:
+        dtb = context.root / f"usr/lib/linux-image-{kver}" / context.config.devicetree
+        cmd += ["--devicetree", dtb]
+        options += ["--ro-bind", dtb, dtb]
+
     if context.config.secure_boot:
         assert context.config.secure_boot_key
         assert context.config.secure_boot_certificate
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -1255,6 +1255,7 @@ class Config:
     initrds: list[Path]
     initrd_packages: list[str]
     microcode_host: bool
+    devicetree: Optional[Path]
     kernel_command_line: list[str]
     kernel_modules_include: list[str]
     kernel_modules_exclude: list[str]
@@ -2133,6 +2134,12 @@ SETTINGS = (
         section="Content",
         parse=config_make_list_parser(delimiter=","),
         help="Add additional packages to the default initrd",
+    ),
+    ConfigSetting(
+        dest="devicetree",
+        section="Content",
+        parse=config_parse_string,
+        help="Devicetree included in UKI",
     ),
     ConfigSetting(
         dest="kernel_command_line",
@@ -3548,6 +3555,7 @@ def summary(config: Config) -> str:
                     Shim Bootloader: {config.shim_bootloader}
                             Initrds: {line_join_list(config.initrds)}
                     Initrd Packages: {line_join_list(config.initrd_packages)}
+                         Devicetree: {config.devicetree}
                 Kernel Command Line: {line_join_list(config.kernel_command_line)}
              Kernel Modules Include: {line_join_list(config.kernel_modules_include)}
              Kernel Modules Exclude: {line_join_list(config.kernel_modules_exclude)}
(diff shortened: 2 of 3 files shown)


mkosi/config.py
```diff
@@ -2134,6 +2135,12 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
         parse=config_make_list_parser(delimiter=","),
         help="Add additional packages to the default initrd",
     ),
+    ConfigSetting(
+        dest="devicetree",
+        section="Content",
+        parse=config_parse_string,
```
behrmann, line 2141: You probably want `config_make_path_parser(required=False)` here.

mkosi/__init__.py
```diff
@@ -1872,6 +1872,11 @@ def build_uki(
         "--ro-bind", stub, stub,
     ]
 
+    if context.config.devicetree:
+        dtb = context.root / f"usr/lib/linux-image-{kver}" / context.config.devicetree
```
behrmann, line 1876: There should probably be some checking here for the case that the file does not exist in the image.

behrmann (commented):
Looks mostly reasonable to me. I think this might also be possible by embedding a ukify config file in the package manager tree or the image. I think that was also our idea at some point, to not have to provide all the knobs from ukify's CLI in mkosi, although we've already exposed quite a few of them.

mkosi/config.py
manut: Thanks for the hint. Applied.

mkosi/__init__.py
manut: The "other" options like e.g. initrds also do not check the existence of the files.

If a non existing dtb will be configured the current output looks like this:

```
‣  Generating unified kernel image for kernel version 6.6.15-arm64
bwrap: Can't find source path /home/manut/.cache/mkosi/mkosi-workspaceuw1qmalv/root/usr/lib/linux-image-6.6.15-arm64/idonot/exist.dtb: No such file or directory
bwrap: Can't find source path /home/manut/.cache/mkosi/mkosi-workspaceuw1qmalv/root/usr/lib/linux-image-6.6.15-arm64/idonot/exist.dtb: No such file or directory
‣ "/usr/bin/ukify --cmdline @/home/manut/.cache/mkosi/mkosi-workspaceuw1qmalv/cmdline --os-release @/home/manut/.cache/mkosi/mkosi-workspaceuw1qmalv/root/usr/lib/os-release --stub /home/manut/.cache/mkosi/mkosi-workspaceuw1qmalv/root/usr/lib/systemd/boot/efi/linuxaa64.efi.stub --output /home/manut/.cache/mkosi/mkosi-workspaceuw1qmalv/root/boot/EFI/Linux/debian-6.6.15-arm64.efi --efi-arch aa64 --uname 6.6.15-arm64 --devicetree /home/manut/.cache/mkosi/mkosi-workspaceuw1qmalv/root/usr/lib/linux-image-6.6.15-arm64/idonot/exist.dtb --sign-kernel --signtool sbsign --secureboot-private-key /home/manut/qemu/mkosi.key --secureboot-certificate /home/manut/qemu/mkosi.crt --pcr-private-key /home/manut/qemu/mkosi.key --pcr-banks sha1,sha256 build --linux /home/manut/.cache/mkosi/mkosi-workspaceuw1qmalv/root/usr/lib/modules/6.6.15-arm64/vmlinuz --initrd /home/manut/.cache/mkosi/mkosi-workspaceuw1qmalv/initrd/image.raw-initrd" returned non-zero exit code 1.
```

Shall we fail with a nicer looking message?

DaanDeMeyer, line 1876: This seems like a completely arbitrary documentation to me for putting device trees. Is there any kind of spec mandating that this is where device trees should be installed? If not, we need such a spec first. See for example https://github.com/uapi-group/specifications/pull/91 where a standard location for UKI addons is proposed. We'd need such a proposal for device trees as well.

Until then, I'd prefer if you used a ukify config file in a package manager tree to accomplish this.

PR #2441, opened 2024-03-01
mailmap: use a signle spelling for Jörg's and Neal's contributions

keszybz:


behrmann (approved):
Thanks :) I've been meaning to fix that. Recently switched to the proper spelling, since my patience with system's not handling unicode properly has finally run out.

Issue #2430, opened 2024-02-25
Add ToolsTreeRepositores / ToolsTreePackageManagerTrees


jgouly:
For an image I'm building, I want to include the Asahi Mesa library in the ToolsTree so that I can run QEMU images with GL acceleration.

However there's currently no way to add extra repos to the ToolsTree.

I locally have the below hack which passes PackageManagerTrees to the ToolsTree build also:
```
diff --git a/mkosi/__init__.py b/mkosi/__init__.py
index 74e20fe3..3800fd16 100644
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3593,6 +3593,7 @@ def finalize_default_tools(args: Args, config: Config, *, resources: Path) -> Co
         *(["--package-cache-dir", str(config.package_cache_dir)] if config.package_cache_dir else []),
         "--incremental", str(config.incremental),
         "--acl", str(config.acl),
+        *([f"--package-manager-tree={package.source}:{package.target}" for package in config.package_manager_trees]),
         *([f"--package={package}" for package in config.tools_tree_packages]),
         "--output", f"{config.tools_tree_distribution}-tools",
         *(["--source-date-epoch", str(config.source_date_epoch)] if config.source_date_epoch is not None else []),
```

septatrix:
Maybe mkosi could allow for custom tools tree similar to initrds. Ship definitions for one which works for most and otherwise make dependencies/images for work them

DaanDeMeyer:
> Maybe mkosi could allow for custom tools tree similar to initrds. Ship definitions for one which works for most and otherwise make dependencies/images for work them

Maybe read the documentation for the ToolsTree option ;)

PR #2447, opened 2024-03-02
Add ToolsTreeRepositories= and ToolsTreePackageManagerTrees=

DaanDeMeyer:
Fixes https://github.com/systemd/mkosi/issues/2430

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -124,6 +124,9 @@ class ConfigTree:
     def with_prefix(self, prefix: Path = Path("/")) -> tuple[Path, Path]:
         return (self.source, prefix / os.fspath(self.target).lstrip("/") if self.target else prefix)
 
+    def __str__(self) -> str:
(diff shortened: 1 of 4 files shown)


mkosi/config.py
```diff
@@ -3431,23 +3452,11 @@ def none_to_default(s: Optional[object]) -> str:
...
-
-
-def line_join_tree_list(array: Sequence[ConfigTree]) -> str:
+def line_join_list(array: Iterable[object]) -> str:
     if not array:
         return "none"
 
-    items = [format_tree(tree) for tree in array]
+    items = (str(item) for item in array)
     return "\n                                     ".join(items)
```
behrmann, line 3460: This wasn't too long before either, I think this can be safely inlined
```suggestion
    return "\n                                     ".join(str(item) for item in array)
```

behrmann (approved):

PR #2450, opened 2024-03-02
Two opensuse improvements

DaanDeMeyer:


behrmann:
> As a workaround, the db can be rebuilt in BDB format in a finalize chroot
script by running "rpm --rebuilddb".

This should probably be mentioned in the NEWS.

behrmann (approved):

PR #2451, opened 2024-03-02
Add string strip prefix helper

keszybz:


DaanDeMeyer:
Can't we just use removeprefix() which is already in stdlib?

keszybz:
> Can't we just use removeprefix() which is already in stdlib?

This replaces the uses of removeprefix. Sadly, removeprefix is semi-useless because it returns "success" when the prefix was removed and also when it wasn't removed.

> mkosi/distributions/fedora.py:3:1: I001 [*] Import block is un-sorted or un-formatted

Hmm, can we get rid of this check? It's really forcing a bad change here.

DaanDeMeyer:
> Hmm, can we get rid of this check? It's really forcing a bad change here.

Even if the change is locally bad here I think overall this provides a benefit so I'd like to keep it.

behrmann:
> This replaces the uses of removeprefix. Sadly, removeprefix is semi-useless because it returns "success" when the prefix was removed and also when it wasn't removed.

It returns the string after removing the prefix that might or might not be there. I think it's a valid choice to design the API like that, since it allows to chain string operations. Returning the string might also not be a "success", since it might be empty, which is falsey.  To have our cake and eat it, too, this would need something more akin to a Rust result type, with a success and an error state wrapping the string.

That being said, I like the added function. It makes most of the code more readable in most cases, except the parsing of the source date epoch, where I prefer the original, which is—I think—less complex.

mkosi/util.py
```diff
@@ -59,6 +59,12 @@ def round_up(x: int, blocksize: int = 4096) -> int:
     return (x + blocksize - 1) // blocksize * blocksize
 
 
+def string_strip_prefix(s: str, prefix: str) -> Optional[str]:
+    if s.startswith(prefix):
+        return s.removeprefix(prefix)
+    return None
```
DaanDeMeyer, line 65: I would name this function `startswith` so it mimicks the same function we have in systemd

DaanDeMeyer (changes requested):

DaanDeMeyer:
Pushed an update to rename to startswith() and fixed the `ruff` error. Setting the green label.

PR #2452, opened 2024-03-03
Only enable sha256 PCR bank for swtpm

DaanDeMeyer:
Mimicks the same change in systemd-vmspawn
(https://github.com/systemd/systemd/commit/519bad6c2c23d3c2dc9558878becb485f3ae9057)


behrmann (approved):
LGTM, but maybe we could not send the output to dev null when we have debug mode enabled?

PR #2460, opened 2024-03-05
Package manager metadata cleaning fixes

DaanDeMeyer:


behrmann:
Added a label so we mention not cleaning up package manager metadata on `auto` for directory and tar output

PR #2463, opened 2024-03-05
Update NEWS

DaanDeMeyer:


NEWS.md
```diff
@@ -33,6 +33,36 @@
...
+  still change. As such `mkosi vmspawn` is also considered experimental.
+- Added `SyncScripts=` which can be used to update various build sources
+  before starting the image build.
+- The `DISTRIBUTION=` and `RELEASE=` environment variables are now set
+  when running scripts.
+- Added `ToolsTreeRepositories=` and `ToolsTreePackageManagerTrees=`.
+- Added `RuntimeNetwork=` to configure the networking used when booting
+  the image.
+- Added `SecureBootKeySource=` and `VerityKeySource=` to support signing
+  images with OpenSSL engines.
```
bluca, line 65: Should probably say that it requires systemd v256?

bluca (approved):

bluca (approved):

mkosi/__init__.py
```diff
@@ -1843,6 +1843,16 @@ def extract_pe_section(context: Context, binary: Path, section: str, output: Pat
...
 
 
+def want_signed_pcrs(config: Config) -> bool:
+    return bool(
+        config.sign_expected_pcr == ConfigFeature.enabled or
+        (
+            config.sign_expected_pcr == ConfigFeature.auto and
+            find_binary("systemd-measure", "/usr/lib/systemd/systemd-measure", root=config.tools())
+        )
+    )
```
behrmann, line 1853: Wrapping this in `bool` does the trick, but that this is anything that a bool is just an artifact from or returning the first non-False thing and and returning the last. Checking the return from `find_binary` makes this intent a bit clearer, I think.
```suggestion
    return (
        config.sign_expected_pcr == ConfigFeature.enabled or
        (
            config.sign_expected_pcr == ConfigFeature.auto and
            find_binary("systemd-measure", "/usr/lib/systemd/systemd-measure", root=config.tools()) is not None
        )
    )
```

NEWS.md
```diff
@@ -33,6 +33,40 @@
...
+- Made GPG key handling for Fedora rawhide more robust
+- If systemd-repart 256 or newer is available, mkosi will instruct it
+  to generate /etc/fstab and /etc/crypttab for the image if any
+  partition definitions contain the corresponding settings
+  (`MountPoint=` and `EncryptedVolume=`).
+- `bash` is now started in the debug shell instead of `sh`.
+- The default release for Ubuntu is now `noble`.
+- Ubuntu is now used as the default tools tree distribution for Ubuntu
+  instead of Debian.
+- Added `mkosi vmspawn` which boots the image with `systemd-vmspawn`.
```
behrmann, line 54: Since we're mentioning it below, let's also mention that it will be in 256.

NEWS.md
```diff
@@ -33,6 +33,40 @@
...
+- Added `TriggerMatch=` to specify multiple match sections of which only
+  one should be satisfied.
+- Added various packages to the default tools tree.
+- Added `mkosi-install`, `mkosi-upgrade`, `mkosi-remove` and
+  `mkosi-reinstall` scripts which allow writing scripts that are
+  independent of the package manager being used to build the image.
+- We now expand specifiers in `Match` section values
+- Made GPG key handling for Fedora rawhide more robust
+- If systemd-repart 256 or newer is available, mkosi will instruct it
+  to generate /etc/fstab and /etc/crypttab for the image if any
```
behrmann, line 47:
```suggestion
  to generate `/etc/fstab` and `/etc/crypttab` for the image if any
```

NEWS.md
```diff
@@ -33,6 +33,40 @@
   to install additional packages.
 - Implemented `CacheOnly=metadata`. Note that in the JSON output, the
   value of `CacheOnly=` will now be a string instead of a boolean.
+- Added `CompressLevel=` to set the compression level to use.
+- Dropped experimental Gentoo support.
+- Added `TriggerMatch=` to specify multiple match sections of which only
+  one should be satisfied.
+- Added various packages to the default tools tree.
```
behrmann, line 40: I think it would be clearer to enumerate them. If I didn't miss anything looking at it it was
- jq
- attr
- acl
- git
- sed
- grep
- findutils

NEWS.md
behrmann: Nevermind, missed I scrolled over the 255 heading in the NEWS.

PR #2468, opened 2024-03-06
Write fallback gnutls crypto policy when setting up apt

DaanDeMeyer:
This is required on Fedora to make sure gnutls (which is used by apt) works properly.


behrmann (approved):
CI failures looks unrelated.

PR #2473, opened 2024-03-09
Make config available as env variable

Cornelicorn:


DaanDeMeyer (changes requested):
Can you set `indent=0` and make sure that the environment variable contains minified json since environment variable size is limited.

DaanDeMeyer:
Actually we should do this slightly differently, environment variable size limit is 128k which will be too small to fit proper configs. Instead, set `$MKOSI_CONFIG=/work/config` for every script and write config to a temporary file which you then mount to `/work/config` in the sandbox.

behrmann:
Daan already mentioned my major gripe with this, a temporary file is a better choice for this than an environment variable, so the latter should just hold the path to the file in case we want to move it. With a file the json also doesn't need to be minified, which might be nice when using the debug shell and having a look a it (without needing to pipe it through jq for formatting). 

The new environment variable also needs to be mentioned in the doc section for environment variables and added to the table.

Cornelicorn:
This version now dumps the config into the workspace at it's creating and mounts it to `/work/config.json` for all the scripts.

Cornelicorn:
Should this also get a `NEWS.md` entry?

DaanDeMeyer (approved):

DaanDeMeyer:
@Cornelicorn I'll add it when I write news for the next release

PR #2476, opened 2024-03-10
Add grub for EFI support

DaanDeMeyer:
We also rework the grub setup to not copy the grub modules into the
ESP anymore. We do this as grub for EFI booted in secure boot mode
does not load any unsigned modules for security reasons so we opt
to include all necessary modules into the grub image itself.

### DIFF
--- a/mkosi.conf.d/20-opensuse.conf
+++ b/mkosi.conf.d/20-opensuse.conf
@@ -22,6 +22,7 @@ Packages=
         erofs-utils
         grep
         grub2-i386-pc
+        grub2-x86_64-efi
         iproute
         iputils
         kernel-kvmsmall
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -12,6 +12,7 @@ import os
 import resource
 import shlex
 import shutil
+import stat
 import subprocess
 import sys
 import tempfile
@@ -863,7 +864,8 @@ def sign_efi_binary(context: Context, input: Path, output: Path) -> Path:
         context.config.secure_boot_sign_tool == SecureBootSignTool.auto and
         find_binary("sbsign", root=context.config.tools()) is not None
     ):
-        with open(output, "wb") as f:
+        with tempfile.NamedTemporaryFile(dir=output.parent, prefix=output.name) as f:
+            os.chmod(f.name, stat.S_IMODE(input.stat().st_mode))
             cmd: list[PathString] = [
                 "sbsign",
                 "--key", context.config.secure_boot_key,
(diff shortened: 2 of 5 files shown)


mkosi/__init__.py
```diff
@@ -863,7 +864,8 @@ def sign_efi_binary(context: Context, input: Path, output: Path) -> Path:
         context.config.secure_boot_sign_tool == SecureBootSignTool.auto and
         find_binary("sbsign", root=context.config.tools()) is not None
     ):
-        with open(output, "wb") as f:
+        with tempfile.NamedTemporaryFile(dir=output.parent, prefix=output.name) as f:
```
behrmann, line 867: I think this can raise `FileNotFoundError`, since the file is missing when it's supposed to be deleted, at least this happens with
```python
with tempfile.NamedTemporaryFile(prefix="foo-", delete=False) as f:
     f.write(b"foobar\n")
     os.rename(f.name, "/tmp/foobar")
```
for me. `NamedTemporaryFile` has a `delete=` argument, that is `True` by default, setting it to false should make this safe.

mkosi/__init__.py
```diff
@@ -1278,18 +1275,31 @@ def grub_bios_install(context: Context, partitions: Sequence[Partition]) -> None
...
+
+    if want_grub_bios(context):
+        grub_mkimage(context, target="i386-pc", modules=("biosdisk",))
+
+    if want_grub_efi(context):
+        output = context.root / (
+            shim_second_stage_binary(context)
+            if context.config.shim_bootloader != ShimBootloader.none
+            else efi_boot_binary(context)
+        )
```
behrmann, line 1291: Shorter and easier on the eyes—in my opinion
```suggestion
        if context.config.shim_bootloader != ShimBootloader.none:
            output = context.root / shim_second_stage_binary(context)
        else:
            output = context.root / efi_boot_binary(context)
```

mkosi/__init__.py
DaanDeMeyer: Setting it to `false` means the file doesn't get cleaned up on failure either. Changed it to use `os.link()` instead.

behrmann (approved):

mkosi.conf.d/20-opensuse.conf
```diff
@@ -22,6 +22,7 @@ Packages=
         erofs-utils
         grep
         grub2-i386-pc
+        grub2-x86_64-efi
```
bluca, line 25: won't this break building !x86 images?

DaanDeMeyer: The opensuse image doesn't build on !x86 to begin with. I make these !x86 buildable on an as-needed basis and nobody's needed !x86 suse yet.

PR #2478, opened 2024-03-11
Boot improvements

### LINKED ISSUE #2472: Ability to build images that are directly bootable on hardware with the usual MS signing keys
I've been fiddling around with the new `ShimBootloader=signed` option and was sort of expecting images to be bootable with SecureBoot on machines from the usual vendors. 

After some chatting with @DaanDeMeyer - it became clear that even though all the pieces are available in mkosi, we need a bit more glue-logic to arrange such images.

We were talking about two main areas for improvement:

**mkosi build**
Afaik, what is missing here is the ability to detect and copy the signed bootloader and kernel directly from the distribution - mkosi should not try to build a unified kernel image as these cannot be verified

A solution could be to introduce `Bootloader=grub-signed` that should try copy the signed grub and copy signed kernels/or UKI's from the distribution itself.

**mkosi qemu**
It would be very nice if it was possible to test these images using qemu, i don't know if it should have a flag that makes it magicly use some OVMF vars with the Microsoft keys enrolled. To me the `QemuFirmwareVariables=` and accompanying documentation that mentions `virt-firmware` was enough.

DaanDeMeyer:
Fixes #2472

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -914,13 +914,14 @@ def install_systemd_boot(context: Context) -> None:
         return
 
     directory = context.root / "usr/lib/systemd/boot/efi"
-    if not directory.exists() or not any(directory.iterdir()):
+    signed = context.config.shim_bootloader == ShimBootloader.signed
+    if not directory.glob("*.efi.signed" if signed else "*.efi"):
         if context.config.bootable == ConfigFeature.enabled:
-            die("A EFI bootable image with systemd-boot was requested but systemd-boot was not found at "
-                f"{directory.relative_to(context.root)}")
+            die(f"An EFI bootable image with systemd-boot was requested but a {'signed ' if signed else ''}"
+                f"systemd-boot binary was not found at {directory.relative_to(context.root)}")
         return
 
-    if context.config.secure_boot:
+    if context.config.secure_boot and not signed:
         with complete_step("Signing systemd-boot binaries…"):
             for input in itertools.chain(directory.glob('*.efi'), directory.glob('*.EFI')):
                 output = directory / f"{input}.signed"
@@ -1129,9 +1130,10 @@ def want_grub_efi(context: Context) -> bool:
     if context.config.bootloader != Bootloader.grub:
         return False
 
-    have = find_grub_directory(context, target="x86_64-efi") is not None
-    if not have and context.config.bootable == ConfigFeature.enabled:
-        die("An EFI bootable image with grub was requested but grub for EFI is not installed")
+    if context.config.shim_bootloader != ShimBootloader.signed:
+        have = find_grub_directory(context, target="x86_64-efi") is not None
+        if not have and context.config.bootable == ConfigFeature.enabled:
+            die("An EFI bootable image with grub was requested but grub for EFI is not installed")
 
     return True
 
@@ -1191,9 +1193,13 @@ def prepare_grub_config(context: Context) -> Optional[Path]:
             f.write("set timeout=0\n")
 
     if want_grub_efi(context):
-        # Signed EFI grub shipped by distributions reads its configuration from /EFI/<distribution>/grub.cfg in
-        # the ESP so let's put a shim there to redirect to the actual configuration file.
-        earlyconfig = context.root / "efi/EFI" / context.config.distribution.name / "grub.cfg"
+        # Signed EFI grub shipped by distributions reads its configuration from /EFI/<distribution>/grub.cfg (except
+        # in OpenSUSE) in the ESP so let's put a shim there to redirect to the actual configuration file.
+        if context.config.distribution == Distribution.opensuse:
+            earlyconfig = context.root / "efi/EFI/BOOT/grub.cfg"
+        else:
+            earlyconfig = context.root / "efi/EFI" / context.config.distribution.name / "grub.cfg"
+
         with umask(~0o700):
             earlyconfig.parent.mkdir(parents=True, exist_ok=True)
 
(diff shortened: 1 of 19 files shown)


mkosi/resources/mkosi.md
```diff
@@ -1498,20 +1503,30 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
+
+: When set to `microsoft`, a firmware variables file with the Microsoft
+  secure boot certificates already enrolled will be used.
+
+: When set to `custom`, the secure boot certificate from
+  `SecureBootCertificate=` will be enrolled into the default firmware
+  variables file.
 
 : `virt-fw-vars` from the
   [virt-firmware](https://gitlab.com/kraxel/virt-firmware) project can
```
septatrix, line 1532: Maybe drop the link and simply mention that it is installed in the tools tree.

Also GH does not let me annotate it but I guess the below section about where Debian and Fedora ship ovmf variables with MS keys enrolled is now superfluous?

mkosi/qemu.py
```diff
@@ -660,16 +624,49 @@ def run_qemu(args: Args, config: Config) -> None:
...
+            assert ovmf
             ovmf_vars = stack.enter_context(tempfile.NamedTemporaryFile(prefix="mkosi-ovmf-vars"))
-            shutil.copy2(config.qemu_firmware_variables or find_ovmf_vars(config), Path(ovmf_vars.name))
-            cmdline += ["-drive", f"file={ovmf_vars.name},if=pflash,format=raw"]
-            if ovmf_supports_sb:
+            ovmf_vars_format = (
+                "raw"
+                if config.qemu_firmware_variables not in (None, Path("custom"), Path("microsoft"))
+                else ovmf.vars_format
+            )
```
behrmann, line 640: As soon as the ternary isn't a oneliner anymore, it is shorted as a proper if-statement.
```suggestion
            if config.qemu_firmware_variables not in (None, Path("custom"), Path("microsoft"))
                 ovmf_vars_format = "raw"
            else:
                 ovmf_vars_format = ovmf.vars_format
```

mkosi/__init__.py
```diff
@@ -941,13 +941,14 @@ def install_systemd_boot(context: Context) -> None:
         return
 
     directory = context.root / "usr/lib/systemd/boot/efi"
-    if not directory.exists() or not any(directory.iterdir()):
+    signed = context.config.shim_bootloader == ShimBootloader.signed
+    if not directory.glob("*.efi.signed" if signed else "*.efi"):
         if context.config.bootable == ConfigFeature.enabled:
-            die("A EFI bootable image with systemd-boot was requested but systemd-boot was not found at "
-                f"{directory.relative_to(context.root)}")
+            die(f"A EFI bootable image with systemd-boot was requested but a {'signed' if signed else ''} "
```
behrmann, line 947: Nit: 
```suggestion
            die(f"A EFI bootable image with systemd-boot was requested but a {'signed ' if signed else ''}"
```
otherwise there is a double space after "a" in the unsigned case.

mkosi/qemu.py
```diff
@@ -172,119 +172,79 @@ def find_qemu_binary(config: Config) -> str:
...
-        "usr/share/ovmf/OVMF_VARS.fd",
-        "usr/share/OVMF/OVMF_VARS_4M.fd",
-        "usr/share/OVMF/OVMF_VARS.fd",
-    ]
+        for target in j["targets"]:
+            if target["architecture"] != arch:
+                continue
+
+            for glob in target["machines"]:
+                if machine in glob:
```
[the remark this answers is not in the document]
DaanDeMeyer: Tried that first, doesn't work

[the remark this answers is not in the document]
behrmann: Ah, yes, the qemu targets I can see in that dir on my machine are `pc-q35-*`, `pc-i440fx-*` and `microvm`, so `fnmatch` can't match this. I think a comment like 
> # "in" is used here because we need to match, e.g. "q35" in "pc-q35-*"

would be helpful when reading this without having to consult the files.

mkosi/__init__.py
```diff
@@ -2098,29 +2155,39 @@ def install_uki(context: Context, kver: str, kimg: Path, token: str, partitions:
...
-        kver,
-        context.root / kimg,
-        initrds,
-        finalize_cmdline(context, roothash),
-        boot_binary,
-    )
+    if context.config.shim_bootloader == ShimBootloader.signed:
+        for p in (context.root / "usr/lib/modules" / kver).glob("*.efi"):
+            shutil.copy2(p, boot_binary)
+            break
```
behrmann, line 2165: What's the reason for only picking the first one we find and not all of them? In the future there might be multiple different distro-provided UKIs that people could install. A debug log line about which UKI is picked here might be helpful.

behrmann (approved):
LGTM, one question and I still think a comment documenting an example for the strings being matched with the firmware glob would be nice.

mkosi/__init__.py
DaanDeMeyer: We'd have to figure out naming first, let's do that in a separate PR

mkosi/tree.py
```diff
@@ -146,8 +148,10 @@ def copy_tree(
 
 def rmtree(*paths: Path, sandbox: SandboxProtocol = nosandbox) -> None:
     if paths:
+        parents = set(p.parent for p in paths)
```
behrmann, line 151: Let's make this
```suggestion
        parents = sorted(set(p.parent for p in paths))
```
since `set` is not guaranteed to be. This way we guarantee the output to look stable between different invocations.

behrmann (approved):

PR #2481, opened 2024-03-11
Only mount /etc/resolv.conf if it actually exists

DaanDeMeyer:


behrmann (approved):

mkosi/sandbox.py
```diff
@@ -171,7 +171,7 @@ def sandbox_cmd(
     if scripts:
         cmdline += ["--ro-bind", scripts, "/scripts"]
 
-    if network and not relaxed:
+    if network and not relaxed and Path("/etc/resolv.conf").exists():
         cmdline += ["--bind", "/etc/resolv.conf", "/etc/resolv.conf"]
```
septatrix, line 175: Couldn't this simply be a `--bind-try` (to prevent races)?

DaanDeMeyer: No the command lines in --debug mode are already too long as it is, I'm not going to make them longer to prevent a theoretical race.

PR #2484, opened 2024-03-12
Use --close with flock and drop --no-fork

DaanDeMeyer:
Let's make sure only flock holds the lock and it's not inherited by the child process it spawns. This fixes an SELinux denial on some systems.


behrmann (approved):

behrmann:
CI failures look unrelated
- arch-arch: got stuck and ran out of time
- arch-debian: too
- fedora-arch: a mirror failed to deliver a package
- opensuse-arch: caught a keyboard interrupt during a cp

Merging, since on Matrix we got told it works.

PR #2486, opened 2024-03-12
Catch ENODEV as well when checking if a device is available

DaanDeMeyer:
This triggers on kernel updates on Arch without rebooting first so let's make sure we handle it as well.


behrmann:
Having `ENODEV` in a separate branch there and suggestig a reboot might actually prove to be quite helpful for users.

DaanDeMeyer:
I think that's a bit too specific, we don't really know why the ENODEV could be popping up.

PR #2487, opened 2024-03-12
ci: print group section lines for GitHub workflow

behrmann:


septatrix:
My issue was focused towards people running mkosi in their CI (1-3 runs) and not about the selftests in this repo (20+ runs) but this seems like an improvement nonetheless

tests/__init__.py
```diff
@@ -163,3 +164,13 @@ def suspend_capture_stdin(pytestconfig: Any) -> Iterator[None]:
...
+
+
+@contextlib.contextmanager
+def ci_group(s: str) -> Iterator[None]:
+    if os.getenv("MKOSI_CI_SECTION") == "github":
+        print("", flush=True)
+        print(f"::group::{s}", flush=True)
+    yield
+    if os.getenv("MKOSI_CI_SECTION") == "github":
+        print("::endgroup::", flush=True)
```
DaanDeMeyer, line 176: try/finally here please

tests/test_initrd.py
```diff
@@ -40,13 +40,20 @@ def passphrase() -> Iterator[Path]:
...
-        options=[
-            "--directory", "",
-            "--include=mkosi-initrd/",
-        ],
-    ) as initrd:
+    with (
+        ci_group(
+            f"Initrd image {config.distribution}/{config.release} using {config.tools_tree_distribution} tools tree"
+            if config.tools_tree_distribution else
+            f"Initrd image {config.distribution}/{config.release} (no tools tree)"
```
DaanDeMeyer, line 47: I don't see the point of logging about whether we use a tools tree or not. Just add a summary dump before the build instead if you want that info logged.

tests/__init__.py
```diff
@@ -163,3 +164,13 @@ def suspend_capture_stdin(pytestconfig: Any) -> Iterator[None]:
 
     if pytestconfig.getoption("capture") == "no":
         capmanager.resume_global_capture()
+
+
+@contextlib.contextmanager
+def ci_group(s: str) -> Iterator[None]:
+    if os.getenv("MKOSI_CI_SECTION") == "github":
```
DaanDeMeyer, line 171: Just check for `$GITHUB_ACTIONS` here which is always set to true when running in github actions.

tests/conftest.py
```diff
@@ -50,3 +50,9 @@ def config(request: Any) -> Image.Config:
         tools_tree_distribution=cast(Distribution, request.config.getoption("--tools-tree-distribution")),
         debug_shell=request.config.getoption("--debug-shell"),
     )
+
+
+@pytest.fixture(autouse=True)
+def github_ci_sections(request: Any) -> Any:
```
DaanDeMeyer, line 56: This should be named `test_ci_sections`

DaanDeMeyer (changes requested):

tests/conftest.py
behrmann: I went with just `ci_sections`, since it's not a test.

DaanDeMeyer (approved):

tests/__init__.py
```diff
@@ -163,3 +164,15 @@ def suspend_capture_stdin(pytestconfig: Any) -> Iterator[None]:
...
     if pytestconfig.getoption("capture") == "no":
         capmanager.resume_global_capture()
+
+
+@contextlib.contextmanager
+def ci_group(s: str) -> Iterator[None]:
+    try:
+        if os.getenv("GITHUB_ACTIONS"):
+            print("", flush=True)
+            print(f"::group::{s}", flush=True)
```
DaanDeMeyer, line 174: This should be outside of the try

behrmann: I think that's visually a bit more noisy, but I don't have a strong preference.

behrmann:
Setting the "don't merge" label for the moment, because I want to check the CI logs before merging

.github/workflows/ci.yml
```diff
@@ -170,7 +170,7 @@ jobs:
 
     - name: Run integration tests
       run: |
-        sudo timeout -k 30 1h python3 -m pytest \
+        sudo --preserve-env=GITHUB_ACTIONS timeout -k 30 1h python3 -m pytest \
```
DaanDeMeyer, line 173: This can just be --preserve-env, we don't care about security in CI

behrmann: Changed it to a plain `--preserve-env`

behrmann:
Dropped the don't merge label, since the last run worked fine and passing more environment shouldn't change that.

PR #2489, opened 2024-03-13
Run sync scripts in strict sandbox again

DaanDeMeyer:
We can't use a relaxed sandbox as we need to be able to mount the directories containing certificates and keys from the tools tree (finalize_crypto_mounts()) which will fail with permission errors in a relaxed sandbox if the required mountpoints do not already exist in the host filesystem.

So let's switch back to a strict sandbox, except that we now always run as root in the strict sandbox so the overlay mount for /usr from the package manager tree won't fail anymore.

Partial revert of 9dd49ad22944bdb90a748d5b35d138ed1ce0ae2a

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -421,19 +421,19 @@ def finalize_config_json(config: Config) -> Iterator[Path]:
         yield Path(f.name)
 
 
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
Merging since this isn't covered by CI yet

PR #2506, opened 2024-03-14
Use Github Workflow commands in log formatting

behrmann:
Only a draft for now so I can see what this looks like.


behrmann:
I'll close this for now. As is this doesn't work as intended.

PR #2511, opened 2024-03-15
Filter and sort all mounts in sandbox_cmd()

DaanDeMeyer:
We don't want users of sandbox_cmd() to have to care about mount
ordering. Currently, if mounts with a more general destination are
ordered after mounts with a more specific destination, the earlier
mount is hidden by the later mount. By sorting by destination, we
avoid this issue.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -65,7 +65,7 @@ from mkosi.run import (
     log_process_failure,
     run,
 )
-from mkosi.sandbox import chroot_cmd, finalize_crypto_mounts, finalize_passwd_mounts
+from mkosi.sandbox import Mount, chroot_cmd, finalize_crypto_mounts, finalize_passwd_mounts
 from mkosi.tree import copy_tree, move_tree, rmtree
 from mkosi.types import PathString
 from mkosi.user import CLONE_NEWNS, INVOKING_USER, become_root, unshare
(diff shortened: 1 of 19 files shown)


mkosi/mounts.py
```diff
@@ -123,16 +124,16 @@ def mount_overlay(
...
 
-        options: list[PathString] = ["--dir", "/work/src"]
-        for src, target in sorted(mounts, key=lambda s: s[1]):
-            options += ["--bind", src, target]
+        mounts = []
+        for src, target in sorted(sources, key=lambda s: s[1]):
+            mounts += [Mount(src, target)]
 
-        yield options
+        yield mounts
```
behrmann, line 139:
```python
        yield [Mount(src, target) for src, target in sorted(sources, key=lambda s: s[1])]
```

mkosi/sandbox.py
```diff
@@ -5,18 +5,45 @@
...
+    required: bool = True
+
+    def __hash__(self) -> int:
+        return hash((Path(self.src), Path(self.dst), self.devices, self.ro, self.required))
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, Mount):
+            return False
+
+        return self.__hash__() == other.__hash__()
```
behrmann, line 29: Why not make this a dataclass and get this for free?

mkosi/__init__.py
```diff
@@ -608,21 +612,22 @@ def run_build_scripts(context: Context) -> None:
...
+                            *sources,
+                            Mount(script, "/work/build-script", ro=True),
+                            Mount(json, "/work/config.json", ro=True),
+                            Mount(cd, "/work/scripts", ro=True),
+                            Mount(context.root, context.root),
+                            Mount(context.install_dir, "/work/dest"),
+                            Mount(context.staging, "/work/out"),
                             *(
-                                ["--bind", os.fspath(context.config.build_dir), "/work/build"]
+                                [Mount(os.fspath(context.config.build_dir), "/work/build")]
```
behrmann, line 624: That's no longer needed
```suggestion
                                [Mount(context.config.build_dir, "/work/build")]
```

mkosi/__init__.py
```diff
@@ -608,21 +612,22 @@ def run_build_scripts(context: Context) -> None:
...
                         network=context.config.with_network,
-                        options=sources + [
-                            "--ro-bind", script, "/work/build-script",
-                            "--ro-bind", json, "/work/config.json",
-                            "--ro-bind", cd, "/work/scripts",
-                            "--bind", context.root, context.root,
-                            "--bind", context.install_dir, "/work/dest",
-                            "--bind", context.staging, "/work/out",
+                        mounts=[
+                            *sources,
```
behrmann, line 616: Nit: I preferred the `sources` instead of unpacking it here, but I'm only +0 on the issue.

behrmann:
Otherwise LGTM, much cleaner :)

mkosi/sandbox.py
DaanDeMeyer: I'm not sure if the dataclass `__eq__` would do what we need. 

> eq: If true (the default), an [__eq__()](https://docs.python.org/3/reference/datamodel.html#object.__eq__) method will be generated. This method compares the class as if it were a tuple of its fields, in order. Both instances in the comparison must be of the identical type.

We don't want a direct comparison, we want a comparison after converting the `PathString` fields to `Path`

PR #2513, opened 2024-03-15
Always mount context.root to /buildroot

DaanDeMeyer:
Let's leak fewer host specific details into the sandbox by always mounting the image root directory to /buildroot in the sandbox. This also simplifies debugging as the image rootfs will always be at /rootfs instead of in some host specific path.

### DIFF
--- a/mkosi/sandbox.py
+++ b/mkosi/sandbox.py
@@ -64,14 +64,14 @@ def have_effective_cap(capability: Capability) -> bool:
     return (int(hexcap, 16) & (1 << capability.value)) != 0
 
 
-def finalize_passwd_mounts(root: Path) -> list[Mount]:
+def finalize_passwd_mounts(root: PathString) -> list[Mount]:
     """
     If passwd or a related file exists in the apivfs directory, bind mount it over the host files while we
     run the command, to make sure that the command we run uses user/group information from the apivfs
     directory instead of from the host.
     """
     return [
-        Mount(root / "etc" / f, f"/etc/{f}", ro=True, required=False)
+        Mount(Path(root) / "etc" / f, f"/etc/{f}", ro=True, required=False)
         for f in ("passwd", "group", "shadow", "gshadow")
     ]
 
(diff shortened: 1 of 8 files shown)


mkosi/sandbox.py
```diff
@@ -239,43 +239,43 @@ def sandbox_cmd(
...
         "sh", "-c",
-        f"chmod 1777 {root / 'tmp'} {root / 'var/tmp'} {root / 'dev/shm'} && "
-        f"chmod 755 {root / 'run'} && "
+        "chmod 1777 /buildroot/tmp /buildroot/var/tmp /buildroot/dev/shm && "
+        "chmod 755 /buildroot/run && "
         # Make sure anything running in the root directory thinks it's in a container. $container can't always be
         # accessed so we write /run/host/container-manager as well which is always accessible.
-        f"mkdir -m 755 {root}/run/host && echo mkosi >{root}/run/host/container-manager && "
+        "mkdir -m 755 /buildroot/run/host && echo mkosi >/buildroot/run/host/container-manager && "
         "exec $0 \"$@\"",
```
behrmann, line 262: The warning is wrong, since the implicit concatenation is intended here, but maybe something like this
```python
        " && ".join(
            [
                "chmod 1777 /buildroot/tmp /buildroot/var/tmp /buildroot/dev/shm",
                "chmod 755 /buildroot/run",
                "mkdir -m 755 /buildroot/run/host && echo mkosi >/buildroot/run/host/container-manager",
                "exec $0 \"$@\"",
            ],
        )
```
or 
```python
        (
            "chmod 1777 /buildroot/tmp /buildroot/var/tmp /buildroot/dev/shm && " +
            "chmod 755 /buildroot/run && " +
            "mkdir -m 755 /buildroot/run/host && echo mkosi >/buildroot/run/host/container-manager && " +
            "exec $0 \"$@\""
        ),
```
(with or without the `+` operators) would make the inline nature of this script more apparent.

Also same below for the other warning.

behrmann (approved):

PR #2515, opened 2024-03-15
Fix kernel module glob

DaanDeMeyer:
When dkms modules are installed there will be .ko modules in directories other than "kernel" so make sure we only look for modules in the "kernel" directory.


behrmann (approved):

septatrix:
Does this mean that dkms modules are now never installed?

PR #2516, opened 2024-03-15
Virtiofsd fixes

DaanDeMeyer:


behrmann (commented):
LGTM, though the `logging.info` added in the third commit could probably also be a `logging.warning` for better visibility.

Setting the news label, because the needed exclusive access for virtio looks very plausible, but I think it's something we should mention somewhere.

PR #2518, opened 2024-03-16
Introduce VirtualMachineMonitor= and various cleanups

DaanDeMeyer:


tests/test_json.py
```diff
@@ -128,6 +129,7 @@ def test_config() -> None:
             "FinalizeScripts": [],
             "Format": "uki",
             "Hostname": null,
+            "Vmm": "qemu",
```
behrmann, line 132: This needs sorting.

tests/test_json.py
```diff
@@ -398,8 +401,8 @@ def test_config() -> None:
         qemu_gui = True,
         qemu_kernel = None,
         qemu_kvm = ConfigFeature.auto,
-        qemu_mem = "",
-        qemu_smp = "yes",
+        qemu_mem = 123,
+        qemu_smp = 2,
```
behrmann, line 405: These two are still the other value in the JSON above.

mkosi/firecracker.py
```diff
@@ -0,0 +1,167 @@
...
+    QemuFirmware,
+)
+from mkosi.log import die
+from mkosi.partition import finalize_root, find_partitions
+from mkosi.qemu import apply_runtime_size, copy_ephemeral, finalize_drive, generate_scratch_fs, want_scratch
+from mkosi.run import run
+
+
+def maybe_decompress_kernel(src: Path, dst: Path) -> Path:
+    b = src.read_bytes()
```
behrmann, line 29: This might be a bit much, since it reads the whole file. It's also a bit weird to do this and then open the file again later. The latter could be avoided by wrapping `b` in a a `io.BytesIO`, but maybe it would be best to open `src` at the beginning?

mkosi/resources/mkosi.md
```diff
@@ -1450,6 +1438,26 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 : List of colon-separated paths to look for tools in, before using the
   regular `$PATH` search path.
 
+`Vmm=`, `--vmm=`
```
behrmann, line 1441: Though explained in the next sentence, the name is a bit cryptic. Maybe `VmMonitor=`?

behrmann (commented):
The refactors are nice!

If we add firecracker support I'm not sure whether in addition to removing the `vmspawn` verb, we should also remove the `qemu` verb (kept around hidden for backwardscompat) and exchange it for a generic `vm` verb or whether maybe `boot` should be the unified frontend.

mkosi/resources/mkosi.md
DaanDeMeyer: Vmm seems to be the usual abbrevation used for this. I want to rename all `QemuXXX=` settings to `VmmXXX=` in the future, so keeping it short seems good here?

mkosi/firecracker.py
DaanDeMeyer: We have to read the whole file regardless to be able to search in it. Opening the file here means an extra level of indentation which I want to avoid.

mkosi/resources/mkosi.md
DaanDeMeyer: Renamed to `VirtualMachineMonitor=` but kept `--vmm`

mkosi/firecracker.py
behrmann: > We have to read the whole file regardless to be able to search in it. 

Yes, but we don't have to read the whole file it into into some buffer we're holding only to read the first four bytes.

> Opening the file here means an extra level of indentation which I want to avoid.

That's true, that's cumbersome, but since we're already opening and closing the file twice, this could just be two with statements on top of each other, adding one level of indentation for the first few lines, but not the whole function.

mkosi/resources/mkosi.md
behrmann: With that consideration in mind that seems seems sensible.

PR #2521, opened 2024-03-17
Fix erroneous negation leading to omitted firmware files

septatrix:


DaanDeMeyer (changes requested):
I don't think the extra directories cause any harm, please submit that part as a separate PR if you really care about that and keep just the bugfix in this PR.

septatrix:
> I don't think the extra directories cause any harm, please submit that part as a separate PR if you really care about that and keep just the bugfix in this PR.

I extracted the directory stuff into #2522

DaanDeMeyer (approved):

PR #2522, opened 2024-03-18
Omit empty kernel/firmware directories from initrd

septatrix:
Extracted into a separate PR as suggested in https://github.com/systemd/mkosi/pull/2521#pullrequestreview-1941599887

While the space savings are not huge they are measurable, especially when having a lot of modules in the main image and very few in the initrd. In my case I have the `@hardware-support` fedora group installed which pulls in pretty much every driver though I have excluded almost all of those from the initrd.

Additionally, this makes the directory structure a lot cleaner and easier to inspect which kernel modules and firmware files are actually included.

### DIFF
--- a/mkosi/kmod.py
+++ b/mkosi/kmod.py
@@ -172,30 +172,22 @@ def gen_required_kernel_modules(
         mods = set((modulesd / "kernel").rglob("*.ko*"))
         firmware = set()
 
-    yield modulesd.parent
-    yield modulesd
-    yield modulesd / "kernel"
+    # We do not want to yield directories lower than /usr/lib, especially none outside of our root namespace.
+    # Therefore we disregard the lowest few parent directores.
+    # This also improves performance as fewer set memberships have to be checked.
+    prefix_length = len((root / "usr/lib").parts)
+    required_paths = mods | firmware | set(
+        parent
+        for file in mods | firmware
+        for parent in file.parents[:-prefix_length]
+    )
+    yield from sorted(required_paths)
 
-    if (root / "usr/lib/firmware").exists():
-        yield root / "usr/lib/firmware"
-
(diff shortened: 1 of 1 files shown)


mkosi/kmod.py
```diff
@@ -172,30 +172,22 @@ def gen_required_kernel_modules(
...
-        yield p
-
-    for p in (root / modulesd).iterdir():
-        if p.name.startswith("modules"):
-            yield p
+    # We do not want to yield directories lower than /usr/lib, especially none outside of our root namespace.
+    # Therefore we disregard the lowest few parent directores.
+    # This also improves performance as fewer set memberships have to be checked.
+    prefix_length = len((root / "usr/lib").parts)
+    required_paths = mods | firmware | set(
```
keszybz, line 179: Please just use the set comprehension: `{ parent for … }`

keszybz (commented):
Looks reasonable.

mkosi/kmod.py
behrmann: The whole thing could also be written as
```python
firmwared = root / "usr/lib/firmware"
required_paths = mods | firmware | { 
    parent
    for file in mods | firmware
    for parent in file.parents
    if parent.is_relative_to(modulesd) or parent.is_relative_to(firmwared)
}
```
It's a bit clearer (no list index juggling) and performs the same. It's also a bit tighter in what it accepts.

behrmann (approved):
LGTM, same nit as @keszybz and another suggestion.

DaanDeMeyer:
I've just put up https://github.com/systemd/mkosi/pull/2523 which fixes this in a (subjectively) cleaner way

DaanDeMeyer:
Let's close this as the other PR was merged

PR #2523, opened 2024-03-18
Only add directories with modules in them to the kmods initrd

DaanDeMeyer:


DaanDeMeyer:
Replaces #2522

behrmann (approved):

mkosi/kmod.py
```diff
@@ -179,12 +188,10 @@ def gen_required_kernel_modules(
     if (root / "usr/lib/firmware").exists():
         yield root / "usr/lib/firmware"
 
-    for d in (modulesd, root / "usr/lib/firmware"):
-        for p in (root / d).rglob("*"):
-            if p.is_dir():
-                yield p
+    yield from {p for m in sorted(mods) for p in parents_relative_to(m, modulesd / "kernel")}
+    yield from {p for f in sorted(firmware) for p in parents_relative_to(f, root / "usr/lib/firmware")}
```
septatrix, line 192: As this yields from the set it loses sorting

mkosi/kmod.py
```diff
@@ -179,12 +188,10 @@ def gen_required_kernel_modules(
     if (root / "usr/lib/firmware").exists():
         yield root / "usr/lib/firmware"
```
septatrix, line 189: This yield and the three above could also be removed by adjusting the path below to just `root / "usr/lib"`

PR #2525, opened 2024-03-18
Simplify gen_required_kernel_modules()

DaanDeMeyer:
- Make sure we yield paths in sorted order
- Use yield from more
- Replace parents_relative_to() with a simpler solution

### DIFF
--- a/mkosi/kmod.py
+++ b/mkosi/kmod.py
@@ -150,12 +150,9 @@ def resolve_module_dependencies(
     return set(nametofile[m] for m in mods if m in nametofile), set(firmware)
 
 
-def parents_relative_to(path: Path, other: Path) -> Iterator[Path]:
-    for p in path.parents:
-        if p == other or not p.is_relative_to(other):
-            return
-
-        yield p
+def parents_below(path: Path, below: Path) -> list[Path]:
+    parents = list(path.parents)
+    return parents[:parents.index(below)]
 
 
 def gen_required_kernel_modules(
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
cc @septatrix

mkosi/kmod.py
```diff
@@ -181,28 +173,21 @@ def gen_required_kernel_modules(
...
+        )
+    )
 
-    if (root / modulesd / "vdso").exists():
+    if (modulesd / "vdso").exists():
         yield modulesd / "vdso"
-
-        for p in (root / modulesd / "vdso").iterdir():
-            yield p
+        yield from sorted((modulesd / "vdso").iterdir())
```
septatrix, line 190: I am not fully sure what these files to but I think they should be above the `if not mods or not de: return`?

septatrix: Also the sorting should be redundant as we do not return any subdirectories

DaanDeMeyer: Direction iteration order is not stable, so this is required for reproducible builds as well.

septatrix (approved):

mkosi/kmod.py
```diff
@@ -150,12 +150,9 @@ def resolve_module_dependencies(
     return set(nametofile[m] for m in mods if m in nametofile), set(firmware)
 
 
-def parents_relative_to(path: Path, other: Path) -> Iterator[Path]:
-    for p in path.parents:
-        if p == other or not p.is_relative_to(other):
-            return
-
-        yield p
+def parents_until(path: Path, until: Path) -> list[Path]:
```
behrmann, line 153: I think that name would be a bit clearer
```suggestion
def parents_below(path: Path, until: Path) -> list[Path]:
```
since—to me—until  doesn't imply whether the bound is part or not, while below makes it clear that all of them should be children.

behrmann (approved):

PR #2534, opened 2024-03-19
Further support for working with sudo and non-home directories

richardmaw-codethink:
I missed a couple of cases because the directories already existed.

If `INVOKING_USER.mkdir(path)` is used on a path in a shared working directory like

```
$ ls -ld /data
drwxrws--- 4 root devgroup 4096 Mar 12 17:11 /data
```

then the run call needs extra groups to also be set so that it can traverse /data.

This PR also includes a relaxing of the restriction that the invoking user's uid and gid are only used if the path is a subpath of home, but I'm not familiar with the need for that restriction in the first place so whether there's already a parent directory owned by the user is a guess at an alternative.


mkosi/user.py
```diff
@@ -41,6 +42,11 @@ def name(cls) -> str:
     def home(cls) -> Path:
         return Path(f"~{cls.name()}").expanduser()
 
+    @classmethod
+    @functools.lru_cache(maxsize=1)
+    def extra_groups(cls) -> Sequence[int]:
```
DaanDeMeyer, line 47: This should be called in `INVOKING_USER.init()` to initialize it.

DaanDeMeyer (changes requested):

DaanDeMeyer (approved):

mkosi/user.py
```diff
@@ -25,6 +26,7 @@ class INVOKING_USER:
     def init(cls) -> None:
         name = cls.name()
         home = cls.home()
+        extra_groups = cls.extra_groups()
```
DaanDeMeyer, line 29: The method should just be called, not assigned to a variable

DaanDeMeyer (changes requested):

DaanDeMeyer (approved):

PR #2539, opened 2024-03-20
Only set some environment variables when invoking package managers

DaanDeMeyer:
We really only want to set KERNEL_INSTALL_BYPASS and SYSTEMD_HWDB_UPDATE_BYPASS when we're invoking package managers so let's make sure those are only set when invoking package managers.

Let's also allow users to override both all of these and let's not set them when Bootable=no so distros can do whatever they want if mkosi's bootable image logic is not being used.


behrmann (approved):

behrmann:
CI failure is due to ancient dnf in Debian.

PR #2540, opened 2024-03-20
Mount pkgmngr/etc as a whole instead of individual files in it.

DaanDeMeyer:
We stopped doing this because bubblewrap would create the /etc/mtab in the package manager tree /etc and leave it there after exiting, which would cause bubblewrap to fail the next time we executed it as the symlink already existed.

To avoid reintroducing this issue, we only have bubblewrap create the symlink if nothing's going to be mounted to /etc and create the symlink ourselves in the pkgmngr/etc.

These changes also make sure that the mounts from finalize_passwd_mounts() take precedence over the passwd and group files from pkgmngr/etc.


keszybz:
Looks reasonable.

PR #2541, opened 2024-03-20
Set $HOME to the user's home directory in sync scripts

DaanDeMeyer:
Makes sure git can find the user's git configuration.


keszybz:
Looks reasonable.

PR #2543, opened 2024-03-21
Add proxy settings + various fixes

DaanDeMeyer:


behrmann (approved):
LGTM, but the docs could mention which tools are specifically configured to use the proxy (package managers) and which environment variables are set (the ones for git as well as the generic ones)

DaanDeMeyer:
I'll do a follow up PR for the docs

PR #2544, opened 2024-03-21
Remove setuid/setgid bits from build and workspace directory

DaanDeMeyer:
Both of these can be inherited so remove them from both the workspace and the build directory where inheriting these bits could end up leaking stuff from the host into the image.

Also remove INVOKING_USER.mkdir() while we're at it as only one user was remaining which we can do much more easily by doing the logic before we go into the user namespace.


richardmaw-codethink:
I spotted this branch existed while rebasing, and can confirm it also fixes my problem with setgid bits making their way into packages and causing directory conflicts with other packages.

Issue #2545, opened 2024-03-21
[Match] sections ignored in mkosi.conf.d directories nested in mkosi.images


JM1:
The `[Match]` section of a `mkosi.conf` file in a directory like `mkosi.images/system` or `mkosi.images/system/mkosi.conf.d/30-centos-fedora` shall apply to the entire directory, for example `mkosi.images/system/mkosi.conf.d` or `mkosi.images/system/mkosi.conf.d/30-centos-fedora/mkosi.conf.d`. However, inheritance of the `[Match]` section to nested directories breaks in the second case `mkosi.images/system/mkosi.conf.d/30-centos-fedora`.

To better illustrate the problem I have uploaded a reproducer at [JM1/mkosi-issue-match-inheritance](https://github.com/JM1/mkosi-issue-match-inheritance):

It produces two images `base` and `system` of type `directory` for distributions `centos`, `debian`, `fedora` and `ubuntu`. The `system` image will install package `shim` on `centos` and `fedora` which is not available on `debian` (but on `ubuntu` it is). It will install package
`shim-signed` on `debian` and `ubuntu` which is not available on `centos` and `fedora`.

In practice though, `mkosi` will try to install `shim` when building the `system` image for `debian`. The `[Match]` section in [mkosi.images/system/mkosi.conf.d/30-centos-fedora/mkosi.conf] is not applied to [mkosi.images/system/mkosi.conf.d/30-centos-fedora/mkosi.conf.d/example.conf] properly. The `system` image build fails 😱

Is this behavior expected? 🤔 Can mkosi be changed to include nested dirs in `mkosi.images` when evaluating `[Match]` sections? 🥺

[mkosi.images/system/mkosi.conf.d/30-centos-fedora/mkosi.conf]: https://github.com/JM1/mkosi-issue-match-inheritance/blob/main/mkosi.images/system/mkosi.conf.d/30-centos-fedora/mkosi.conf
[mkosi.images/system/mkosi.conf.d/30-centos-fedora/mkosi.conf.d/example.conf]: https://github.com/JM1/mkosi-issue-match-inheritance/blob/main/mkosi.images/system/mkosi.conf.d/30-centos-fedora/mkosi.conf.d/example.conf

DaanDeMeyer:
@JM1 This is definitely a bug, fix waiting in the linked PR.

JM1:
@DaanDeMeyer Thank you for fixing this bug so quickly 🤩

PR #2546, opened 2024-03-22
Make sure we yield an empty section marker at the end of parse_ini()

DaanDeMeyer:
Fixes https://github.com/systemd/mkosi/issues/2545


JM1 (approved):
Tested it against the [reproducer](https://github.com/JM1/mkosi-issue-match-inheritance) for #2545 and can confirm it now works as expected 🥳

PR #2551, opened 2024-03-26
Allow escaping in QemuArgs=

DaanDeMeyer:
Sometimes we need spaces inside the qemu argument so let's make sure that's possible.


behrmann (approved):

behrmann:
Merging, since the unit tests passed and the rest doesn't exercise the change. No point in burning cycles on it.

PR #2553, opened 2024-03-26
Capitalise inner JSON keys

behrmann:


mkosi/config.py
```diff
@@ -1194,6 +1194,20 @@ def __call__(
...
+
+
+def transform_dict_keys(obj: T, key_transformer: Callable[[S], S], exceptions: Sequence[type]) -> T:
+    if isinstance(obj, list):
+        return cast(T, [transform_dict_keys(e, key_transformer, exceptions) for e in obj])
+    elif isinstance(obj, dict) and type(obj) not in exceptions:
+        return cast(
+            T, {key_transformer(k): transform_dict_keys(v, key_transformer, exceptions) for k, v in obj.items()}
+        )
+    return cast(T, obj) # type: ignore
```
DaanDeMeyer, line 1208: Huh, why do we need this cast?

mkosi/config.py
```diff
@@ -1186,6 +1194,20 @@ def __call__(
...
 
 
+S = TypeVar("S", bound=Any)
+T = TypeVar("T", bound=Any)
+
+
+def transform_dict_keys(obj: T, key_transformer: Callable[[S], S], exceptions: Sequence[type]) -> T:
+    if isinstance(obj, list):
+        return cast(T, [transform_dict_keys(e, key_transformer, exceptions) for e in obj])
+    elif isinstance(obj, dict) and type(obj) not in exceptions:
```
DaanDeMeyer, line 1204: Why doesn't `type(obj) is dict` work here so we don't need the `exceptions` argument?

behrmann: That's sufficient at the moment. `exceptions` was more explicit, since those were cases were we explicitly don't want capitalisation, but I'm fine with excepting every dict for now.

mkosi/config.py
```diff
@@ -445,6 +445,23 @@ def native(cls) -> "Architecture":
...
+        for k, v in kwargs.items():
+            self[k] = v
+
+    def __setitem__(self, key: str, value: Any) -> None:
+        def key_transformer(k: str) -> str:
+            if (s := SETTINGS_LOOKUP_BY_DEST.get(k)) is not None:
+                return s.name
+            return "".join(p.capitalize() for p in k.split("_"))
+
+        super().__setitem__(key_transformer(key), value)
```
DaanDeMeyer, line 462: I wouldn't bother with subclassing dict. A factory function that simply initializes a regular dict with the key transformer is fine.

DaanDeMeyer (changes requested):

DaanDeMeyer (approved):

PR #2556, opened 2024-03-26
Call ukify with --microcode if supported

tfg13:
This is the companion change to https://github.com/systemd/systemd/pull/31872.

If both systemd-stub and ukify are version `256~devel` or greater, mkosi will put the microcode initrd in the new `.ucode` UKI section by using the `--ucode` flag in ukify. Otherwise it will put the microcode initrd as first regular initrd (same as before).

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -9,6 +9,7 @@ import itertools
 import json
 import logging
 import os
+import re
 import resource
 import shlex
 import shutil
@@ -1846,20 +1847,20 @@ def identify_cpu(root: Path) -> tuple[Optional[Path], Optional[Path]]:
     return (Path(f"{vendor_id}.bin"), None)
 
 
-def build_microcode_initrd(context: Context) -> Optional[Path]:
+def build_microcode_initrd(context: Context) -> list[Path]:
     if not context.config.architecture.is_x86_variant():
-        return None
+        return []
 
     microcode = context.workspace / "microcode.initrd"
     if microcode.exists():
-        return microcode
+        return [microcode]
 
     amd = context.root / "usr/lib/firmware/amd-ucode"
     intel = context.root / "usr/lib/firmware/intel-ucode"
 
     if not amd.exists() and not intel.exists():
         logging.warning("/usr/lib/firmware/{amd-ucode,intel-ucode} not found, not adding microcode")
-        return None
+        return []
 
     root = context.workspace / "microcode-root"
     destdir = root / "kernel/x86/microcode"
@@ -1871,7 +1872,7 @@ def build_microcode_initrd(context: Context) -> Optional[Path]:
         vendorfile, ucodefile = identify_cpu(context.root)
         if vendorfile is None or ucodefile is None:
             logging.warning("Unable to identify CPU for MicrocodeHostonly=")
-            return None
+            return []
         with (destdir / vendorfile).open("wb") as f:
             f.write(ucodefile.read_bytes())
     else:
@@ -1887,7 +1888,7 @@ def build_microcode_initrd(context: Context) -> Optional[Path]:
 
     make_cpio(root, microcode, sandbox=context.sandbox)
 
-    return microcode
+    return [microcode]
 
 
 def build_kernel_modules_initrd(context: Context, kver: str) -> Path:
@@ -1993,7 +1994,8 @@ def build_uki(
     stub: Path,
     kver: str,
     kimg: Path,
-    initrds: Sequence[Path],
+    microcodes: list[Path],
+    initrds: list[Path],
     cmdline: Sequence[str],
     output: Path,
 ) -> None:
@@ -2077,6 +2079,20 @@ def build_uki(
     cmd += ["build", "--linux", kimg]
     mounts += [Mount(kimg, kimg, ro=True)]
 
+    if microcodes:
+        ukify = find_binary("ukify", "/usr/lib/systemd/ukify", root=context.config.tools())
+        assert ukify is not None
+        # new .ucode section support?
+        if (
+            systemd_tool_version(context.config, ukify) >= "256~devel" and
+            systemd_stub_version(context, stub) >= "256~devel"
+        ):
+            for microcode in microcodes:
+                cmd += ["--microcode", microcode]
+                mounts += [Mount(microcode, microcode, ro=True)]
+        else:
+            initrds = microcodes + initrds
+
     for initrd in initrds:
         cmd += ["--initrd", initrd]
         mounts += [Mount(initrd, initrd, ro=True)]
@@ -2128,6 +2144,14 @@ def systemd_stub_binary(context: Context) -> Path:
     return stub
 
 
+def systemd_stub_version(context: Context, stub: Path) -> GenericVersion:
+    sdmagic = extract_pe_section(context, stub, ".sdmagic", context.workspace / "sdmagic")
+    sdmagic_text = sdmagic.read_text()
+    if version := re.match(r"#### LoaderInfo: systemd-stub (?P<version>[.~^a-zA-Z0-9-]+) ####", sdmagic_text):
+        return GenericVersion(version.group("version"))
+    die(f"Unable to determine systemd-stub version, found {sdmagic_text!r}")
+
+
 def want_uki(context: Context) -> bool:
     return want_efi(context.config) and (
             context.config.bootloader == Bootloader.uki or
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -2091,6 +2104,15 @@ def systemd_stub_binary(context: Context) -> Path:
     return stub
 
 
+def systemd_stub_version(context: Context, stub: Path) -> GenericVersion:
+    sdmagic = extract_pe_section(context, stub, ".sdmagic", context.workspace / "sdmagic")
+    sdmagic_text = sdmagic.read_text()
+    version = re.match("#### LoaderInfo: systemd-stub (.+) ####", sdmagic_text)
+    if not version:
+        die(f"Unable to determine systemd-stub version, found \"{sdmagic_text}\"")
+    return GenericVersion(version.group(1))
```
behrmann, line 2113:
```suggestion
    if version := re.match("#### LoaderInfo: systemd-stub (?P<version>.+) ####", sdmagic_text):
        return GenericVersion(version.group("version"))
        
    die(f"Unable to determine systemd-stub version, found \"{sdmagic_text}\"")        
```
The regex could also be made a bit more conservative, e.g. only with the character set `[\.\-~^a-zA-Z0-9]`

tfg13:
I will go over this and fix the merge conflicts once the primary PR moves forward

mkosi/__init__.py
```diff
@@ -2077,10 +2079,32 @@ def build_uki(
...
+                cmd += ["--microcode", microcode]
+                mounts += [Mount(microcode, microcode, ro=True)]
+        else:
+            initrds = microcodes + initrds
+
     for initrd in initrds:
         cmd += ["--initrd", initrd]
         mounts += [Mount(initrd, initrd, ro=True)]
 
+    logging.debug(f"final cmd before run: {cmd}")
```
behrmann, line 2106: With `ARG_DEBUG` `run` (or rather `spawn`) will already log the command.

mkosi/__init__.py
```diff
@@ -2077,10 +2079,32 @@ def build_uki(
...
+        initrds = initrds.copy()
+        ukify = find_binary("ukify", "/usr/lib/systemd/ukify", root=context.config.tools())
+        if ukify is None:
+            # this can't happen, we check ukify earlier
+            # but the linter doens't know and is concerned about None
+            die("ukify not found")
+        # new .ucode section support?
+        if (
+            systemd_tool_version(context.config, ukify) >= "256~devel"
+            and systemd_stub_version(context, stub) >= "256~devel"
```
behrmann, line 2094:
```suggestion
            systemd_tool_version(context.config, ukify) >= "256~devel" and
            systemd_stub_version(context, stub) >= "256~devel"
```

mkosi/__init__.py
```diff
@@ -2077,10 +2079,32 @@ def build_uki(
...
+        ):
+            for microcode in microcodes:
+                cmd += ["--microcode", microcode]
+                mounts += [Mount(microcode, microcode, ro=True)]
+        else:
+            initrds = microcodes + initrds
```
behrmann, line 2100:
```suggestion
            initrds += microcodes
```

mkosi/__init__.py
```diff
@@ -2077,10 +2079,32 @@ def build_uki(
     cmd += ["build", "--linux", kimg]
     mounts += [Mount(kimg, kimg, ro=True)]
 
+    if microcodes:
+        logging.debug("microcode codepath YES")
+        # might need modification
+        initrds = initrds.copy()
```
behrmann, line 2085: Why the copy here?

mkosi/__init__.py
```diff
@@ -3133,7 +3174,7 @@ def save_uki_components(context: Context) -> tuple[Optional[Path], Optional[str]
     stub = shutil.copy2(stub, context.workspace)
     microcode = build_microcode_initrd(context)
 
-    return stub, kver, kimg, microcode
+    return stub, kver, kimg, [microcode] if microcode else []
```
behrmann, line 3177: I think i would make sense to have `build_microcode_initrd` return a `list[Path]` now instead of an `Optional[Path]`. Even though it only ever returns an empty list or one with one element, it would make the whole thing more uniform.

behrmann (commented):
> Main PR landed, so I rebased this one. It works, but I am not 100% sure about all the things in copy_uki. There we exact the sections again, not sure if we want/need to handle separate .ucode sections there. (That would also mean a new config value for output_split_ucode? I left it out for now, anyone know?

Right now it would be copied out together with the initrd, right? I think we should copy it out to a separate file then in the split output.

DaanDeMeyer:
> Right now it would be copied out together with the initrd, right? I think we should copy it out to a separate file then in the split output.

Nah just make one big fat initrd in the output. That makes it easier to deal with.

mkosi/__init__.py
tfg13: sorry that is my personal debugging, should have removed that

tfg13:
> > Right now it would be copied out together with the initrd, right? I think we should copy it out to a separate file then in the split output.
> 
> Nah just make one big fat initrd with microcode in front in the output. That makes it easier to deal with.

You mean even if it is two separate sections in the UKI? So basically recombine them for easier output?

mkosi/__init__.py
tfg13: That would change the order though, no? In this (non .ucode) codepath, the microcode is supposed to be first

mkosi/__init__.py
tfg13: IIRC the previous version manipulated the input struct (instead of assigning a new list), so that messed with the caller's datastructure. I think this is no longer the case, I'll double check

DaanDeMeyer:
> > > Right now it would be copied out together with the initrd, right? I think we should copy it out to a separate file then in the split output.
> > 
> > 
> > Nah just make one big fat initrd with microcode in front in the output. That makes it easier to deal with.
> 
> You mean even if it is two separate sections in the UKI? So basically recombine them for easier output?

Yeah, for all intents and purposes we can do with a single split initrd.

mkosi/__init__.py
```diff
@@ -1846,20 +1847,20 @@ def identify_cpu(root: Path) -> tuple[Optional[Path], Optional[Path]]:
     return (Path(f"{vendor_id}.bin"), None)
 
 
-def build_microcode_initrd(context: Context) -> Optional[Path]:
+def build_microcode_initrd(context: Context) -> list[Path]:
```
keszybz, line 1850: Does the return type really need to be changed here? This function returns at most one item, and the previous signature was a better match for that.

mkosi/__init__.py
```diff
@@ -2174,20 +2204,24 @@ def finalize_cmdline(context: Context, roothash: Optional[str]) -> list[str]:
...
-        initrds += context.config.initrds
+        initrds = context.config.initrds
     elif any((context.artifacts / "io.mkosi.initrd").glob("*")):
-        initrds += sorted((context.artifacts / "io.mkosi.initrd").iterdir())
+        initrds = sorted((context.artifacts / "io.mkosi.initrd").iterdir())
     else:
-        initrds += [build_default_initrd(context)]
+        initrds = [build_default_initrd(context)]
 
     return initrds
```
keszybz, line 2226: Same here.

keszybz (changes requested):

mkosi/__init__.py
```diff
@@ -2128,6 +2150,14 @@ def systemd_stub_binary(context: Context) -> Path:
     return stub
 
 
+def systemd_stub_version(context: Context, stub: Path) -> GenericVersion:
+    sdmagic = extract_pe_section(context, stub, ".sdmagic", context.workspace / "sdmagic")
+    sdmagic_text = sdmagic.read_text()
+    if version := re.match(r"#### LoaderInfo: systemd-stub (?P<version>[\.\-~^a-zA-Z0-9]+) ####", sdmagic_text):
+        return GenericVersion(version.group("version"))
+    die(f"Unable to determine systemd-stub version, found \"{sdmagic_text}\"")
```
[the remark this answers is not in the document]
tfg13: TIL

mkosi/__init__.py
tfg13: No, but then we need to mangle it in the caller, which has to return `list[Path]` since it might have to return multiple. @behrmann suggested changing it here instead, above:

> I think i would make sense to have build_microcode_initrd return a list[Path] now instead of an Optional[Path]. Even though it only ever returns an empty list or one with one element, it would make the whole thing more uniform.

No strong opinion, happy to do it either way

keszybz (approved):
LGTM.

DaanDeMeyer (approved):

PR #2557, opened 2024-03-27
Stop measuring for SHA1 TPM bank

DaanDeMeyer:
Let's avoid having to deal with openssl configuration to re-enable SHA1 support by not measuring for the SHA1 TPM bank.


behrmann (approved):
Makes sense, considering the stub is dropping all support for TPM1.2 and TPM2 was the step to get rid of SHA1 being hardcoded, so there shouldn't be any TPM2 out there that doesn't at least support *one* of the others.

PR #2558, opened 2024-03-27
Fix PCR banks

DaanDeMeyer:
Fixes: d44e0f8eaa9bc4837d27bbbb55969f2fa5187897

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1981,9 +1981,6 @@ def build_uki(
         "--output", output,
         "--efi-arch", arch,
         "--uname", kver,
-        # SHA1 might be disabled in OpenSSL depending on the distro so we opt to not sign for SHA1 to avoid having to
-        # manage a bunch of configuration to re-enable SHA1.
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -2028,7 +2025,9 @@ def build_uki(
         if want_signed_pcrs(context.config):
             cmd += [
                 "--pcr-private-key", context.config.secure_boot_key,
-                "--pcr-banks", "sha1,sha256",
+                # SHA1 might be disabled in OpenSSL depending on the distro so we opt to not sign for SHA1 to avoid
+                # having to manage a bunch of configuration to re-enable SHA1.
+                "--pcr-banks", "sha256",
```
behrmann, line 2030: I just went back to 96ba1a479e6d5217a350ad3eb1734b0d235f9c79 when this was introduced (in some other form) and it was sha1 and sha256 ever since. Is there a good reason not to make it `sha256,sha384,sha512` now?

DaanDeMeyer: Not sure, let's make that a separate discussion

PR #2564, opened 2024-03-27
Support mkosi ssh for multiple running instances of the same image

DaanDeMeyer:
Let's add a stopgap solution until systemd-machined supports everything we need. We maintain a super basic JSON state file in the runtime directory that is used to map a machine name to the corresponding SSH proxy command.

We also store the path to the ssh key in there so that mkosi ssh can be run from every directory.

The new Machine= option allows selecting the machine name to use. Unless set explicitly, we also use the machine name as the hostname for the machine.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -1414,6 +1414,7 @@ class Config:
     runtime_network: Network
     ssh_key: Optional[Path]
     ssh_certificate: Optional[Path]
+    machine: Optional[str]
     vmm: Vmm
 
     # QEMU-specific options
@@ -1435,6 +1436,9 @@ class Config:
 
     def name(self) -> str:
         return self.image_id or self.image or "default"
+
+    def machine_or_name(self) -> str:
+        return self.machine or self.name()
 
     def output_dir_or_cwd(self) -> Path:
         return self.output_dir or Path.cwd()
@@ -2748,6 +2752,12 @@ SETTINGS = (
         help="Set the virtual machine monitor to use for mkosi qemu",
     ),
     ConfigSetting(
+        dest="machine",
+        metavar="NAME",
+        section="Host",
+        help="Set the machine name to use when booting the image",
+    ),
+    ConfigSetting(
         dest="qemu_gui",
         metavar="BOOL",
         nargs="?",
@@ -2794,7 +2804,7 @@ SETTINGS = (
         metavar="NUMBER|auto|hash",
         section="Host",
         parse=config_parse_vsock_cid,
-        default=QemuVsockCID.hash,
+        default=QemuVsockCID.auto,
         help="Specify the VSock connection ID to use",
     ),
     ConfigSetting(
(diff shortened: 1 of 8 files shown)


mkosi/resources/mkosi.md
```diff
@@ -101,6 +101,11 @@ The following command line verbs are known:
   arguments to the `ssh` invocation. To connect to a container, use
   `machinectl login` or `machinectl shell`.
 
+: The `Machine=` option can be used to give the machine a custom name
```
behrmann, line 104: Since the error message mentiones `Ephemeral=` this probably should, too.

mkosi/config.py
```diff
@@ -1436,6 +1437,9 @@ class Config:
     def name(self) -> str:
         return self.image_id or self.image or "default"
 
+    def machine_or_default(self) -> str:
```
behrmann, line 1440: Maybe
```suggestion
    def machine_or_name(self) -> str:
```
since `_or_default` leaves open what the default is, whether it's `name()` or just `"default"`.

mkosi/resources/mkosi.md
behrmann: Maybe this should also specifically say
```suggestion
: The `Machine=` option can be used to give the machine a custom hostname
```

mkosi/qemu.py
```diff
@@ -553,6 +553,41 @@ def finalize_drive(drive: QemuDrive) -> Iterator[Path]:
...
+@contextlib.contextmanager
+def finalize_state(config: Config, cid: int) -> Iterator[None]:
+    (INVOKING_USER.runtime_dir() / "machine").mkdir(parents=True, exist_ok=True)
+
+    if INVOKING_USER.is_regular_user():
+        os.chown(INVOKING_USER.runtime_dir(), INVOKING_USER.uid, INVOKING_USER.gid)
+        os.chown(INVOKING_USER.runtime_dir() / "machine", INVOKING_USER.uid, INVOKING_USER.gid)
+
+    with flock(INVOKING_USER.runtime_dir() / "machine"):
+        if (p := INVOKING_USER.runtime_dir() / "machine" / config.machine_or_default()).exists():
```
behrmann, line 565: Would be nice for this to have a `.json` extension.

behrmann (approved):

mkosi/config.py
```diff
@@ -3556,6 +3566,9 @@ def load_kernel_command_line_extra(args: argparse.Namespace) -> list[str]:
     if not any(s.startswith("SYSTEMD_SULOGIN_FORCE=") for s in args.kernel_command_line_extra):
         cmdline += ["SYSTEMD_SULOGIN_FORCE=1"]
 
+    if not any(s.startswith("systemd.hostname=") for s in args.kernel_command_line_extra):
```
septatrix, line 3569: Would this override `Hostname=` values?

DaanDeMeyer: Yes

mkosi/qemu.py
DaanDeMeyer: We would still have to flock it to not conflict with reading the file

mkosi/qemu.py
```diff
@@ -553,6 +547,113 @@ def finalize_drive(drive: QemuDrive) -> Iterator[Path]:
...
+    (INVOKING_USER.runtime_dir() / "machine").mkdir(parents=True, exist_ok=True)
+
+    if INVOKING_USER.is_regular_user():
+        os.chown(INVOKING_USER.runtime_dir(), INVOKING_USER.uid, INVOKING_USER.gid)
+        os.chown(INVOKING_USER.runtime_dir() / "machine", INVOKING_USER.uid, INVOKING_USER.gid)
+
+    with flock(INVOKING_USER.runtime_dir() / "machine"):
+        if (p := INVOKING_USER.runtime_dir() / "machine" / f"{config.machine_or_name()}.json").exists():
+            die(f"Another virtual machine named {config.machine_or_name()} is already running",
+                hint="Use Machine= or --machine to specify a different virtual machine name")
```
behrmann, line 561: We're generally config file option forward, but in this case, wouldn't mentioning `--machine` first make more sense?

behrmann (approved):

PR #2568, opened 2024-03-28
Pass --quiet when invoking busctl

DaanDeMeyer:
We don't care about the return value of the DBUS method so let's silence busctl so it doesn't show it.


behrmann (approved):
Making it not quiet on debug might make sense, but this is also fine.

PR #2574, opened 2024-04-01
Fix empty config.json caused by buffered write

### LINKED ISSUE #2573: $MKOSI_CONFIG file is empty
Just noticed that after pulling new changes that `$MKOSI_CONFIG` is containing the same file but that file is empty.
Bisecting reveals that this was introduced with 9dd49ad22944bdb90a748d5b35d138ed1ce0ae2a

To reproduce:
```bash
cd $(mktemp -d)
cat <<EOF >mkosi.sync
#!/bin/bash

cat \$MKOSI_CONFIG

exit 1
EOF
chmod +x mkosi.sync
mkosi
```

Cornelicorn:
Closes #2573


DaanDeMeyer:
Ugh I've lost count how many times we've had bugs due to a missing flush.

DaanDeMeyer (approved):

PR #2576, opened 2024-04-01
Allocate scopes for virtiofsd and swtpm as well

DaanDeMeyer:


mkosi/qemu.py
```diff
@@ -1002,7 +1012,13 @@ def run_qemu(args: Args, config: Config) -> None:
             for fd in qemu_device_fds.values():
                 os.close(fd)
 
-            allocate_scope(config, qemu.pid)
+            name = f"mkosi-{config.machine_or_name().replace('_', '-')}"
+            allocate_scope(
+                config,
+                name=name,
```
behrmann, line 1018:
```suggestion
            allocate_scope(
                config,
                name=f"mkosi-{config.machine_or_name().replace('_', '-')}",
```
would be a bit more consistent with the rest

behrmann (approved):

mkosi/qemu.py
DaanDeMeyer: We use the name in the description as well so we need a separate variable

PR #2583, opened 2024-04-03
Add specifiers for various paths

### LINKED ISSUE #2579: `ExtraTrees=` via `InitrdInclude=` incorrectly resolved
When using `ExtraTrees=` fom a *file* which is included via `InitrdInclude=` which itself was set in the config file of an image the paths are incorrectly resolved. More specifically they seem to be resolved relative to the project root instead of the config file (or the image root).

I would have expected that paths are resolved relative to the included config file. This is also the case when including a directory which contains a `mkosi.conf` file.

I have not yet had the time to check if this also happens with a normal include, without the image step, or with attributes which resolve paths.

DaanDeMeyer:
Fixes #2579

### DIFF
--- a/mkosi/resources/mkosi.md
+++ b/mkosi/resources/mkosi.md
@@ -1875,6 +1875,31 @@ use `%%`. The following specifiers are understood:
 | `ImageVersion=`    | `%v`      |
 | `Profile=`         | `%p`      |
 
+There are also specifiers that are independent of settings:
+
+| Specifier | Value                                   |
+|-----------|-----------------------------------------|
+| `%C`      | Parent directory of current config file |
+| `%P`      | Current working directory               |
+| `%D`      | Directory that mkosi was invoked in     |
+
+Note that the current working directory changes as mkosi parses its
+configuration. Specifically, each time mkosi parses a directory
+containing a `mkosi.conf` file, mkosi changes its working directory to
+that directory.
+
+Note that the directory that mkosi was invoked in is influenced by the
+`--directory=` command line argument.
+
+The following table shows example values for the directory specifiers
+listed above:
+
+|      | `$D/mkosi.conf` | `$D/mkosi.conf.d/abc/abc.conf` | `$D/mkosi.conf.d/abc/mkosi.conf` |
+|------|-----------------|--------------------------------|----------------------------------|
+| `%C` | `$D`            | `$D/mkosi.conf.d`              | `$D/mkosi.conf.d/qed`            |
+| `%P` | `$D`            | `$D`                           | `$D/mkosi.conf.d/qed`            |
+| `%D` | `$D`            | `$D`                           | `$D`                             |
+
 ## Supported distributions
 
 Images may be created containing installations of the following
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -920,6 +920,28 @@ def test_specifiers(tmp_path: Path) -> None:
                     ImageVersion=%v
                     OutputDirectory=%O
                     Output=%o
+                    ConfigRootDirectory=%D
+                    ConfigRootConfdir=%C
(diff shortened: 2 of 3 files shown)


tests/test_config.py
```diff
@@ -934,6 +950,9 @@ def test_specifiers(tmp_path: Path) -> None:
             "ImageVersion": "1.2.3",
             "OutputDirectory": str(Path.cwd() / "abcde"),
             "Output": "test",
+            "ConfigRootDirectory": os.fspath(d),
+            "ConfigCurrentWorkingDirectory": os.fspath(d / "mkosi.conf.d/qed"),
+            "ConfigParentDirectory": os.fspath(d / "mkosi.conf.d"),
```
behrmann, line 955: Adding some
```
ConfigRootDirectory=%D
ConfigRootConfdir=%C
ConfigRootPwd=%P
```
as well as
```
ConfigAbcDirectory=%D
ConfigAbcConfdir=%C
ConfigAbcPwd=%P
```
and
```
ConfigQedDirectory=%D
ConfigQedConfdir=%C
ConfigQedPwd=%P
```
in the relevant config snippets, this would need a 
```suggestion
            "ConfigRootDirectory": os.fspath(d),
            "ConfigRootConfdir": os.fspath(d),
            "ConfigRootPwd": os.fspath(d),
            "ConfigAbcDirectory": os.fspath(d),
            "ConfigAbcConfdir": os.fspath(d / "mkosi.conf.d"),
            "ConfigAbcPwd": os.fspath(d),
            "ConfigQedDirectory": os.fspath(d),
            "ConfigQedConfdir": os.fspath(d / "mkosi.conf.d/qed"),
            "ConfigQedPwd": os.fspath(d / "mkosi.conf.d/qed"),
```
which is rather instructive, though the names are somewhat less clearn. Still, I think it would be good to test all the combinations here and add a table like this to the docs, so that it's easy to look up.

DaanDeMeyer: The table seems a bit excessive, updated the test

mkosi/resources/mkosi.md
```diff
@@ -1875,6 +1875,22 @@ use `%%`. The following specifiers are understood:
 | `ImageVersion=`    | `%v`      |
 | `Profile=`         | `%p`      |
 
+There are also specifiers that are independent of settings:
+
+| Specifier | Value                                   |
+|-----------------------------------------------------|
+| `%C`      | Parent directory of current config file |
```
behrmann, line 1882:
```suggestion
| Specifier | Value                                   |
|-----------|-----------------------------------------|
| `%C`      | Parent directory of current config file |
```
Otherwise the table rendering is broken

mkosi/resources/mkosi.md
```diff
@@ -1875,6 +1875,22 @@ use `%%`. The following specifiers are understood:
...
+| `%P`      | Current working directory               |
+| `%D`      | Directory that mkosi was invoked in     |
+
+Note that the current working directory changes as mkosi parses its
+configuration. Specifically, each time mkosi parses a directory
+containing a `mkosi.conf` file, mkosi changes its working directory to
+that directory.
+
```
behrmann, line 1890: The table I had in mind would look roughly like this
```
|      | `D/mkosi.conf` | `D/mkosi.conf.d/abc/abc.conf` | `D/mkosi.conf.d/abc/mkosi.conf` |
|------|----------------|-------------------------------|---------------------------------|
| `%C` | `D`            | `D`                           | `D`                             |
| `%P` | `D`            | `D/mkosi.conf.d`              | `D/mkosi.conf.d/qed`            |
| `%D` | `D`            | `D`                           | `D/mkosi.conf.d/qed`            |
```

septatrix:
Thanks, looking good :)

PR #2587, opened 2024-04-03
Trim default image package lists

DaanDeMeyer:
Let's remove build tools and Fedora/CentOS packaging tools as these aren't getting used and pull in the kitchen sink.


mkosi.conf.d/30-debian-ubuntu/mkosi.conf
behrmann, line 1: I think we should maybe postpone this change while Ubuntu CI is still on the fritz.

behrmann (approved):

PR #2590, opened 2024-04-04
Implement run() on top of spawn()

DaanDeMeyer:


mkosi/tree.py
```diff
@@ -158,10 +158,10 @@ def rmtree(*paths: Path, tools: Path = Path("/"), sandbox: SandboxProtocol = nos
             stdout=subprocess.DEVNULL if not ARG_DEBUG.get() else None,
             stderr=subprocess.DEVNULL if not ARG_DEBUG.get() else None)
 
-    paths = tuple(p for p in paths if p.exists())
-    if paths:
-        run(["rm", "-rf", "--", *paths],
-            sandbox=sandbox(mounts=[Mount(p.parent, p.parent) for p in paths]))
+    filtered = {p for p in paths if p.exists()}
+    if filtered:
+        run(["rm", "-rf", "--", *filtered],
```
behrmann, line 163: Since we already stat every member once, we can also sort this.
```suggestion
        run(["rm", "-rf", "--", *sorted(filtered)],
```
It doesn't make it any more reproducible, but the shell output when shown during debug will look more similar between different invocations.

behrmann (approved):
What lovely deduplication!

PR #2592, opened 2024-04-04
Make sure we always sort when iterating over config files in a dir

DaanDeMeyer:


behrmann (approved):

behrmann:
The opensuse error is a repository error unrelated to this change, merging.

PR #2593, opened 2024-04-04
Check in spawn() whether the command we're trying to run is available

### LINKED ISSUE #2584: Installing bootable Debian sid fails with `depmod: 1: exec: depmod: not found`
I added *systemd-boot* and *kmod* explicitly, although, *kmod* is installed by default, so `/usr/sbin/depmod` should be in the image.

    $ git log --no-decorate --oneline -1
    c824325c Merge pull request #2572 from DaanDeMeyer/journal
    $ more mkosi.local.conf
    [Content]
    #Repositories=main,universe
    Packages=linux-image-cloud-amd64
             systemd
             systemd-boot
             kmod
             udev
             dbus
    $ mkosi -d debian -r sid
    […]
    Setting up file (1:5.45-3) ...
    Setting up kmod (32-1) ...
    sed: can't read /etc/modules: No such file or directory
    Setting up libassuan0:amd64 (2.5.6-1) ...
    Setting up libgomp1:amd64 (14-20240330-1) ...
    Setting up libfdt1:amd64 (1.7.0-2) ...
    Setting up amd64-microcode (3.20240116.2) ...
    amd64-microcode: initramfs support missing
    Setting up libexecs1:amd64 (1.4-2) ...
    Setting up libunwind8:amd64 (1.6.2-3) ...
    Setting up libsemanage-common (3.5-1) ...
    Setting up libsource-highlight-common (3.1.9-4.3) ...
    Setting up libsframe1:amd64 (2.42-4) ...
    Setting up libelf1t64:amd64 (0.191-1) ...
    Setting up libjansson4:amd64 (2.14-2+b2) ...
    Setting up libfsverity0:amd64 (1.5-1.1+b2) ...
    Setting up libeac3:amd64 (1.1.2+ds+git20220117+453c3d6b03a0-1.1+b2) ...
    Setting up libkrb5support0:amd64 (1.20.1-6+b1) ...
    Setting up libdw1t64:amd64 (0.191-1) ...
    Setting up systemd-boot (255.4-1+b1) ...
    Setting up ovmf (2024.02-2) ...
    Setting up tzdata (2024a-1) ...
    […]
    Setting up kmod (32-1) ...
    sed: can't read /etc/modules: No such file or directory
    […]
    Processing triggers for shim-signed:amd64 (1.40+15.7-1) ...
    ‣  Running prepare script /home/build/src/mkosi/mkosi.prepare.chroot…
    ‣ bootctl is not installed, assuming 'unknown' kernel type
    ‣  Running depmod for 6.7.9-cloud-amd64
    depmod: 1: exec: depmod: not found
    ‣ "depmod --all --basedir /buildroot 6.7.9-cloud-amd64" returned non-zero exit code 127.
    ‣  (Fixing ownership of package manager cache directory)

DaanDeMeyer:
Currently, if we try to run a command within a sandbox, we fail with
an unclear error if the program is not installed. This is because our
FileNotFoundError exception handler is never triggered as the program
we run via subprocess is almost always "sh" or "bwrap". Let's make sure
we also check for the actual program we're going to run in the sandbox
and show a clear error if it's not available.

Fixes #2584

### DIFF
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -244,6 +244,14 @@ def spawn(
             # expect it to pick.
             assert nfd == SD_LISTEN_FDS_START + i
 
+    # First, check if the sandbox works at all before executing the command.
+    if sandbox and (rc := subprocess.run(sandbox + ["true"]).returncode) != 0:
+        log_process_failure(sandbox, cmdline, rc)
+        raise subprocess.CalledProcessError(rc, sandbox + cmdline)
+
+    if subprocess.run(sandbox + ["sh", "-c", f"command -v {cmdline[0]}"], stdout=subprocess.DEVNULL).returncode != 0:
+        die(f"{cmdline[0]} not found.", hint=f"Is {cmdline[0]} installed on the host system?")
+
     if (
         foreground and
         sandbox and
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1462,7 +1462,6 @@ def grub_bios_setup(context: Context, partitions: Sequence[Partition]) -> None:
         # be able to do the mount and we don't know the pid beforehand.
         run(
             [
-                "sh", "-c", f"mount --bind {mountinfo.name} /proc/$$/mountinfo && exec $0 \"$@\"",
                 setup,
                 "--directory", "/grub",
                 context.staging / context.config.output_with_format,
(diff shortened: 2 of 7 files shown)


mkosi/run.py
```diff
@@ -244,6 +244,12 @@ def preexec() -> None:
             # expect it to pick.
             assert nfd == SD_LISTEN_FDS_START + i
 
+    if (
+        (not sandbox or subprocess.run(sandbox + ["true"]).returncode == 0) and
```
behrmann, line 248: Maybe it's a bit late, but this line confuses me. I assume the `true` check is to see that something can run successfully in the sandbox if one exists? But if that's not the case (the rc isn't 0), then the `die` will be skipped and the rest of the function will probably fail for other reasons. Wouldn't it be better to fail earlier? Also in that case the whole line could go, because it wouldn't matter whether sandbox were empty or not.

behrmann (commented):
LGTM, one question though

PR #2601, opened 2024-04-07
Various improvements

DaanDeMeyer:


mkosi/resources/mkosi.md
```diff
@@ -1777,7 +1777,7 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
   directory to mount into any machine (container or VM) started by
   mkosi. The second path refers to the target directory inside the
   machine. If the second path is not provided, the directory is mounted
-  below `/root/src` in the machine. If the second path is relative, it
+  at `/root/src` in the machine. If the second path is relative, it
```
behrmann, line 1780: This should probably be mentioned in the next NEWS.

mkosi/qemu.py
```diff
@@ -938,22 +941,44 @@ def run_qemu(args: Args, config: Config) -> None:
...
                     "-chardev", f"socket,id={sock.name},path={sock}",
                     "-device", f"vhost-user-fs-pci,queue-size=1024,chardev={sock.name},tag=root",
                 ]
                 kcl += ["root=root", "rootfstype=virtiofs", "rw"]
 
-        for tree in config.runtime_trees:
-            sock = stack.enter_context(start_virtiofsd(config, tree.source, uidmap=True))
-            tag = tree.target.name if tree.target else tree.source.name
+        def add_virtiofs_mount(sock: Path, dst: PathString, *, tag: str) -> None:
+            nonlocal cmdline, kcl
```
behrmann, line 958: It's a big uglier to pass in `cmdline` and kcl` as arguments and return them, but I think it's worth it, since it's easier to reason about than `nonlocal`.

behrmann (commented):
LGTM, but I'd be happier if we could do without the non-locality.

PR #2603, opened 2024-04-08
Add feature request template

behrmann:


behrmann:
Closing since going through the settings once wasn't necessary, but the bug report YAML was missing a label. For feature requests we probably want the same form based thing.

PR #2607, opened 2024-04-08
Log when we can't rename in move_tree()

DaanDeMeyer:
Can help explain why mkosi is slower than it should be.


DaanDeMeyer:
> LGTM, but this might actually be a place to use a `logging.warning`

That might get too noisy in cases where this is expected

PR #2609, opened 2024-04-08
Allow setting QemuSmp=0 to use all available CPUs

DaanDeMeyer:


mkosi/qemu.py
```diff
@@ -712,6 +712,10 @@ def register_machine(config: Config, pid: int, fname: Path) -> None:
     )
 
 
+def available_cpus(config: Config) -> int:
+    return int(run(["nproc"], stdout=subprocess.PIPE, sandbox=config.sandbox()).stdout.strip())
```
behrmann, line 716: Is there anything `nproc` does differently from `os.cpu_count`? The latter calls `os.sched_getaffinity` and I would assume `nproc` does something similar.

behrmann (approved):
LGTM, one small question

behrmann (approved):

mkosi/qemu.py
```diff
@@ -808,7 +808,7 @@ def run_qemu(args: Args, config: Config) -> None:
     cmdline: list[PathString] = [
         find_qemu_binary(config),
         "-machine", machine,
-        "-smp", str(config.qemu_smp),
+        "-smp", str(config.qemu_smp if config.qemu_smp != 0 else os.cpu_count()),
```
behrmann, line 811: One last nit. This could be
```suggestion
        "-smp", str(config.qemu_smp or os.cpu_count()),
```

PR #2618, opened 2024-04-10
Fix credential file and journal remote exit propagation

richardmaw-codethink:
Getting logs out with journal-forward and using a drop-in to make the target unit exit on success or failure found a couple of bugs.

I'm not 100% on the name of the `terminate` parameter and it's possible the process could exit on its own between poll and terminate which could still propagate an exception, but that would have to be a failure we might be interested in and the fix would just be to remove check from the journal-remote call.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -858,6 +858,8 @@ def config_make_dict_parser(delimiter: str,
             else:
                 die(f"{p} does not exist")
 
+            return new
+
         if unescape:
             lex = shlex.shlex(value, posix=True)
             lex.whitespace_split = True
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -166,6 +166,7 @@ def run(
 def spawn(
     cmdline: Sequence[PathString],
     check: bool = True,
+    terminate: bool = False,
     stdin: _FILE = None,
     stdout: _FILE = None,
     stderr: _FILE = None,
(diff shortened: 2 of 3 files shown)


mkosi/run.py
```diff
@@ -281,15 +282,23 @@ def preexec() -> None:
...
         ) as proc:
+            terminated = False
             try:
                 yield proc
             except BaseException:
                 proc.terminate()
                 raise
+            else:
+                if proc.poll() is None and terminate:
+                    proc.terminate()
```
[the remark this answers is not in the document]
richardmaw-codethink: `terminated` might need a better name. It's meant to flag that we explicitly terminated it ourselves at which point the only returncode we're expecting to handle is -SIGTERM.

If we're not checking if it's already exited then `terminate` means "allow `check` to ignore SIGTERM", but I think we only want to ignore SIGTERM if we were the ones to send the terminate signal.

mkosi/config.py
```diff
@@ -858,6 +858,8 @@ def config_parse_dict(value: Optional[str], old: Optional[dict[str, Any]]) -> Op
             else:
                 die(f"{p} does not exist")
 
+            return new
```
DaanDeMeyer, line 861: We also need an additional check that `value` is not an empty string to make sure we don't enter this `if` if we get an empty string.

DaanDeMeyer (changes requested):
@richardmaw-codethink How exactly was systemd-journal-remote failing when terminated? At least a simple test with SIGINT seems to exit with 0 for me.

DaanDeMeyer:
I cherry-picked the first commit into #2620

mkosi/config.py
DaanDeMeyer: Nvm don't bother with this I added it myself as part of #2620

DaanDeMeyer:
Let's close this in favor of #2621

PR #2619, opened 2024-04-10
Delete everything in output directory that we're going to replace

DaanDeMeyer:


jgouly:
Fixes the issue I was having.

PR #2629, opened 2024-04-14
Switch to reprepro for generating local apt repositories

DaanDeMeyer:
We don't install dpkg-dev in tools trees anymore to avoid pulling in perl, which means we don't have access to dpkg-scanpackages in tools trees anymore.

Instead of adding back dpkg-dev, let's instead switch to reprepro for generating our local apt repository. It's written in C, packaged everywhere and has hardly any dependencies.


DaanDeMeyer:
Merging so I can fix systemd CI

PR #2634, opened 2024-04-15
Allow using ExtraSearchPaths= with ToolsTree=

DaanDeMeyer:


behrmann (approved):
I think this is a great change that makes `ExtraSearchPaths=` nicely usable again. At the same time I'm not all that thrilled about the doubling of places where the first argument of a given cmdline needs to be repeated, since this seems like something easily forgotten (and not easily caught in CI if the thing is actually there). 

Although it does sound a bit scary, maybe we could to a self-modifying sandbox by default, where spawn makes a new copy of the sandbox with the first argument of its cmdline added (if not opted out via some means)?

DaanDeMeyer:
So I think we mostly have this covered by making `binary` a required argument for the `sandbox()` method. We could still fat finger the name but I think that's OK given the added complexity of not doubling the binary name would be quite large.

PR #2638, opened 2024-04-17
Drop --cache=always for virtiofsd

daandemeyer:
We want to support users writing to these directories from the host so --cache=always is not an option as that assumes virtiofsd has exclusive write access.


behrmann (approved):

behrmann:
CI failure is unrelated, the mirror was too slow.

PR #2653, opened 2024-04-23
Support SplitArtifacts= for extension images

DaanDeMeyer:


mkosi/__init__.py
```diff
@@ -3464,29 +3469,33 @@ def make_extension_image(context: Context, output: Path) -> None:
...
+                        context.config.verity_key_source.type != KeySource.Type.file
+                    ),
+                    mounts=mounts,
                 ),
-                mounts=mounts,
-            ),
+            ).stdout
         )
 
+    logging.debug(json.dumps(j, indent=4))
```
behrmann, line 3492: It might be nice to not load the json right away, but only further down when it comes to getting the partition. This way here the actual repart output could be logged without round tripping it here.

behrmann (approved):

mkosi/__init__.py
DaanDeMeyer: Ehh should be fine, let's do this in another PR as it needs to be done in make_image() as well

PR #2655, opened 2024-04-24
ci: remove packages.microsoft.com

bluca:
It is not needed, it publishes things like dotnet, and it is often broken, so just remove the sources

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,8 @@ jobs:
 
     - name: Install
       run: |
+        # This is added by default, and it is often broken, but we don't need anything from it
+        sudo rm -f /etc/apt/sources.list.d/microsoft-prod.list
         sudo apt-get update
         sudo apt-get install pandoc python3-pytest
         python3 -m pip install --upgrade setuptools wheel pip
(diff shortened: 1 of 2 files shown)


DaanDeMeyer (approved):

.github/workflows/ci.yml
```diff
@@ -19,6 +19,8 @@ jobs:
 
     - name: Install
       run: |
+        # This is added by default, and it is often broken, but we don't need anything from it
+        rm -f /etc/apt/sources.list.d/microsoft-prod.list
```
keszybz, line 23: Either this will fail or `sudo` is not needed below.

action.yaml
```diff
@@ -35,6 +35,8 @@ runs:
   - name: Dependencies
     shell: bash
     run: |
+      # This is added by default, and it is often broken, but we don't need anything from it
+      rm -f /etc/apt/sources.list.d/microsoft-prod.list
```
keszybz, line 39: Ditto.

.github/workflows/ci.yml
bluca: whops, copypasta

keszybz (approved):

PR #2658, opened 2024-04-24
 Drop redundant ukify lookup 

DaanDeMeyer:
Otherwise the extra search paths are not taken into account.


DaanDeMeyer:
(It was also using find_binary() instead of Config.find_binary() which was causing issues)

PR #2679, opened 2024-05-03
qemu: Configure caching properly

DaanDeMeyer:
For the scratch device, we can use unsafe caching as it's thrown away immediately after the machine shuts down. For the root disk, we can use unsafe caching only if --ephemeral is enabled.

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -1036,7 +1036,7 @@ def run_qemu(args: Args, config: Config) -> None:
 
         if want_scratch(config):
             scratch = stack.enter_context(generate_scratch_fs(config))
-            cache = "cache.writeback=on,cache.direct=on,cache.no-flush=yes"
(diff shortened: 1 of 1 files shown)


bluca (approved):

behrmann (approved):

mkosi/qemu.py
```diff
@@ -1036,7 +1036,7 @@ def add_virtiofs_mount(
 
         if want_scratch(config):
             scratch = stack.enter_context(generate_scratch_fs(config))
-            cache = "cache.writeback=on,cache.direct=on,cache.no-flush=yes"
+            cache = "cache.writeback=on,cache.direct=on,cache.no-flush=yes,aio=io_uring"
```
bluca, line 1039: does this require a particular kernel/qemu combination?

DaanDeMeyer: It's been supported since qemu 5.0

bluca: cool, didn't realize it was that old

PR #2680, opened 2024-05-03
Add RuntimeProperties= setting

DaanDeMeyer:
This allows configuring properties on the scopes spawned by
systemd-nspawn or systemd-run.


DaanDeMeyer:
Arch CI failing will be fixed by https://gitlab.archlinux.org/archlinux/packaging/packages/libpsl/-/merge_requests/3

behrmann:
Another gnarly refactor :) Looks reasonable, but maybe a better name than `RuntimeProperties=` might be `ScopeProperties=`. It would say more pointedly what it's for and we'd keep the more generic name for future use.

DaanDeMeyer:
> Another gnarly refactor :) Looks reasonable, but maybe a better name than `RuntimeProperties=` might be `ScopeProperties=`. It would say more pointedly what it's for and we'd keep the more generic name for future use.

I went for `UnitProperties=` because I don't exclude the scopes becoming transient services at some point.

behrmann:
> I went for `UnitProperties=` because I don't exclude the scopes becoming transient services at some point.

Even better.

behrmann (approved):

PR #2691, opened 2024-05-07
doc: move [Match] and [Config] section further down

behrmann:
Both topics are more advanced and since the entries in the [Match] section have the same names as the the things they match, they are the first hit when searching the document.


DaanDeMeyer (approved):

behrmann:
Merging, since this a docs only change and we don't need to burn CI minutes on that

PR #2692, opened 2024-05-07
Check if modules.builtin exists before reading it

DaanDeMeyer:


behrmann (approved):
LGTM, but CentOS CI seems unhappy because the initrd is too large. Is this from the dropped workaround? Looks like that was merged months ago. Did it only drop into CentOS now?

DaanDeMeyer:
No the new release of systemd in centos pulls in python just like Arch

PR #2696, opened 2024-05-08
Only use a single default tools tree per build

DaanDeMeyer:
Currently, if multiple images are defined, each with a default tools tree, if they use the same distribution, all the various ToolsTreeXXX= settings for the later images will be ignored since we'll only build one tools tree.

Also, if any of the images set Incremental=no, we will remove the default tools tree outputs even if Incremental=yes is enabled for other images.

To keep ourselves sane, let's avoid dealing with multiple defaults tools trees and only look at the last image to figure out whether we should build a default tools tree and what to put in it.


mkosi/__init__.py
```diff
@@ -4558,21 +4562,17 @@ def run_verb(args: Args, images: Sequence[Config], *, resources: Path) -> None:
...
     for i, config in enumerate(images):
-        tools = (
-            finalize_default_tools(args, config, resources=resources)
-            if config.tools_tree and config.tools_tree == Path("default")
-            else None
-        )
+        if tools and config.tools_tree == Path("default"):
+            tools_tree = tools.output_dir_or_cwd() / tools.output
+        else:
+            tools_tree = config.tools_tree
```
behrmann, line 4568: mypy can't synthesize the type properly because the two branches have different types
```suggestion
        tools_tree = (
            tools.output_dir_or_cwd() / tools.output
            if tools and config.tools_tree == Path("default")
            else config.tools_tree
        )
```

mkosi/__init__.py
```diff
@@ -4530,17 +4530,21 @@ def run_verb(args: Args, images: Sequence[Config], *, resources: Path) -> None:
         if args.verb == Verb.build and not args.force:
             check_outputs(config)
 
+    last = images[-1]
```
behrmann, line 4533: Since I can't comment below: `last = image[-1]` happens again directly after this block and I think it would be helpful to add a comment there, that it's still needed, since `last` and `image[-1]` will not be the same anymore after line 4570.

behrmann (approved):

PR #2697, opened 2024-05-09
opensuse: Install dnf5 in OpenSUSE tools tree

DaanDeMeyer:


mkosi/resources/mkosi-tools/mkosi.conf.d/10-opensuse.conf
```diff
@@ -11,6 +11,8 @@ Packages=
         curl
         distribution-gpg-keys
         dnf-plugins-core
+        dnf5
+        dnf5-plugins
```
keszybz, line 15: Does it make sense to have `dnf-plugins-core` too?

DaanDeMeyer: See the line just above dnf5?

keszybz: That's why I'm asking ;)

DaanDeMeyer: @keszybz We still install `dnf` as well (it's in another config file) so we should keep the plugins for dnf as well. We should keep installing dnf because dnf5 is still full of bugs so users should be able to switch back to dnf with `MKOSI_DNF`

PR #2698, opened 2024-05-10
Only run configure scripts for verbs that need a build

DaanDeMeyer:
In systemd, we want to use configure scripts to determine whether qemu was built with support for specific devices and skip running a test if it wasn't, or otherwise add the device to the qemu arguments.

To make this work, we need to run the configure scripts with the default tools tree available if one is configured.

Let's change the behavior of configure scripts to only run for verbs that need a build and run them after building the default tools tree so that they can be run with the tools tree mounted.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -462,7 +462,6 @@ def run_configure_scripts(config: Config) -> Config:
                     env=env | config.environment,
                     sandbox=config.sandbox(
                         binary=None,
-                        tools=False,
                         mounts=[*sources, Mount(script, "/work/configure", ro=True)],
                         options=["--dir", "/work/src", "--chdir", "/work/src"]
                     ),
@@ -4498,9 +4497,6 @@ def run_verb(args: Args, images: Sequence[Config], *, resources: Path) -> None:
         die("No configuration found",
             hint="Make sure you're running mkosi from a directory with configuration files")
 
-    for i, config in enumerate(images):
-        images[i] = run_configure_scripts(config)
-
     if args.verb == Verb.summary:
         if args.json:
             text = json.dumps(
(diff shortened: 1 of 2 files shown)


mkosi/resources/mkosi.md
```diff
@@ -2102,11 +2102,15 @@ in the current working directory. `$SRCDIR` is set to point to the
...
-  serialized as JSON on stdin and should output the modified
-  configuration serialized as JSON on stdout. Note that this script does
-  not use the tools tree even if one is configured.
+  before building the image. This script may be used to dynamically
+  modify the configuration. It receives the configuration serialized as
+  JSON on stdin and should output the modified configuration serialized
+  as JSON on stdout. Note that this script only runs when building or
+  booting the image (`build`, `qemu`, `boot` and `shell` verbs). If a
+  default tools tree is configured, it will be built before running the
+  configure scripts and the the configure scripts will run with the
```
behrmann, line 2111:
```suggestion
  configure scripts and the configure scripts will run with the
```

mkosi/__init__.py
```diff
@@ -4559,6 +4555,14 @@ def run_verb(args: Args, images: Sequence[Config], *, resources: Path) -> None:
 
         check_workspace_directory(config)
 
+    if tools and not (tools.output_dir_or_cwd() / tools.output).exists():
+        if args.verb == Verb.build or args.force > 0:
+            fork_and_wait(run_sync, args, tools, resources=resources)
+            fork_and_wait(run_build, args, tools, resources=resources)
+        else:
+            die(f"Default tools tree requested for image '{last.name()}' but it has not been built yet",
+                hint="Make sure to build the image first with 'mkosi build' or use --force")
```
behrmann, line 4564: I think it would make sense to put `--force` in single quotes, too, for consistency with `'mkosi build'`

behrmann (approved):

PR #2699, opened 2024-05-10
Chroot for depmod and modinfo

DaanDeMeyer:
modinfo cannot always work with output from newer or different depmod.

Specifically, this fixes the case where modinfo sch_fq_codel fails with
"module not found" on CentOS Stream 9 images built from Fedora 40. When
depmod from Fedora 40 is used, modinfo in the image fails with "module
not found". When depmod from inside the image is used, modinfo succeeds
as expected.

We'd rather not do this but in this case there's no other option.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -572,7 +572,7 @@ def run_prepare_scripts(context: Context, build: bool) -> None:
             arg = "final"
 
         for script in context.config.prepare_scripts:
-            chroot = chroot_cmd(resolve=True)
+            chroot = chroot_cmd(resolve=True, work=True)
 
             helpers = {
                 "mkosi-chroot": chroot,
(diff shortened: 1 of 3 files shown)


mkosi/__init__.py
marc-hb, line 1: After this exact commit, `mkosi` fails for me with `modinfo not found`. I'm building Fedora 40 on Fedora 40.

Note my build script (https://github.com/pmem/run_qemu) invokes `mkosi` with `sudo` in case that's relevant. I'm new to `run_qemu` and I don't know yet why it uses `sudo`. It could be explained in its documentation somewhere.

Is `sudo mkosi` supported and regularly tested in CI? If it's not tested then it does not work :-)

DaanDeMeyer: Is the kmod package installed in the image?

marc-hb: Digging deeper, I'm starting to realize https://github.com/pmem/run_qemu seems to be dependent on some pretty ancient version of mkosi. For instance, it creates a mkosi.default file that gets ignored....

DaanDeMeyer: Yeah that's an extremely old version. We've done an api redesign since then so multiple changes will be required to get it working with the latest version.

PR #2702, opened 2024-05-11
Remove workaround

DaanDeMeyer:
All these distros ship newer systemd now, so we can drop the workaround.


behrmann (approved):

behrmann:
I only looked at the Debian and Fedora version of the CentOS CI, but there the directory output still fails because of resolved.

PR #2704, opened 2024-05-12
Add optional file ID for qemu drives

DaanDeMeyer:
For testing multipath in systemd's integration tests, we need multiple qemu drives backed by the same file. Let's allow specifying an additional file ID to make this possible with QemuDrive=.

### DIFF
--- a/mkosi/util.py
+++ b/mkosi/util.py
@@ -17,7 +17,7 @@ import re
 import resource
 import stat
 import tempfile
-from collections.abc import Iterable, Iterator, Mapping, Sequence
+from collections.abc import Hashable, Iterable, Iterator, Mapping, Sequence
 from pathlib import Path
 from types import ModuleType
 from typing import Any, Callable, Optional, TypeVar, no_type_check
@@ -27,6 +27,7 @@ from mkosi.types import PathString
 
 T = TypeVar("T")
 V = TypeVar("V")
+S = TypeVar("S", bound=Hashable)
 
 
 def dictify(f: Callable[..., Iterator[tuple[T, V]]]) -> Callable[..., dict[T, V]]:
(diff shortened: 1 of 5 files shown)


mkosi/util.py
```diff
@@ -319,3 +320,15 @@ def try_or(fn: Callable[..., T], exception: type[Exception], default: T) -> T:
...
+
+
+def groupby(seq: Sequence[T], key: Callable[[T], S]) -> Sequence[tuple[S, list[T]]]:
+    grouped: dict[S, list[T]] = {}
+
+    for i in seq:
+        if key(i) not in grouped:
+            grouped[key(i)] = []
+
+        grouped[key(i)].append(i)
```
behrmann, line 332: Let's only call the key function once per element.
```suggestion
        k = key(i)
        if k not in grouped:
            grouped[k] = []

        grouped[k].append(i)
```

mkosi/util.py
```diff
@@ -319,3 +320,15 @@ def try_or(fn: Callable[..., T], exception: type[Exception], default: T) -> T:
         return fn()
     except exception:
         return default
+
+
+def groupby(seq: Sequence[T], key: Callable[[T], S]) -> Sequence[tuple[S, list[T]]]:
```
behrmann, line 325: We can't make the promise for the return value, so 
```suggestion
def groupby(seq: Sequence[T], key: Callable[[T], S]) -> list[tuple[S, list[T]]]:
```
would be a better return.

The name `groupby` is pretty fitting, but I think something like `group_by_function` would have lesser change to be mistaken by cursory reading with `itertools.groupby`.

behrmann: To be fair, `itertools.groupby` also takes a `key` function and this could be written using `itertools.groupby`. Giving this a better name is hard.

DaanDeMeyer: We should be fine with groupby(). I don't wanna bother with itertools.groupby() because then we have to sort and it doesn't end up more readable then the version without itertools.groupby().

behrmann (approved):

tests/test_json.py
```diff
@@ -437,7 +437,10 @@ def test_config() -> None:
         proxy_url="https://my/proxy",
         qemu_args=[],
         qemu_cdrom=False,
-        qemu_drives=[QemuDrive("abc", 200, Path("/foo/bar"), "abc,qed"), QemuDrive("abc", 200, None, "")],
+        qemu_drives=[
+            QemuDrive("abc", 200, Path("/foo/bar"), "abc,qed", "red"),
```
behrmann, line 441: Still needs accompanying changes in the JSON

Issue #2566, opened 2024-03-28
Specifier for the image name (as used in `Dependencies=`)


NekkoDroid:
Currently there is no way to refer to the name of a given image when using `mkosi.images`, which is a minor problem when `ImageVersion=` is set as the output name will have the version appended which isn't always desirable (e.g. grouping all output in a subdirectory that has the version embedded `mkosi.output/v1.0/...`).

With a specifier for the name you can reduce the duplication where the image name is explicitly named and needs to be kept up-to-date when making changes.

PR #2712, opened 2024-05-15
Prefer `config.image` over `config.image_id`

NekkoDroid:
As it currently stands, when using multiple images via `mkosi.images` all the images have the same (output) name if `ImageId=` is set, which results in all images needing to override the output name as to not clash.

This changes the default output name and the display name of those images to prefer their actual "sub-image" name, while still falling back to the `ImageId=` for the root image case.

This also somewhat has the sideeffect that it solves the original issue I had in https://github.com/systemd/mkosi/issues/2566 (but the requested specifier might still be useful, so I'll keep it open for now).

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -689,7 +689,7 @@ def config_default_compression(namespace: argparse.Namespace) -> Compression:
 
 
 def config_default_output(namespace: argparse.Namespace) -> str:
-    output = namespace.image_id or namespace.image or "image"
+    output = namespace.image or namespace.image_id or "image"
 
     if namespace.image_version:
         output += f"_{namespace.image_version}"
@@ -1523,7 +1523,7 @@ class Config:
     image: Optional[str]
 
     def name(self) -> str:
-        return self.image_id or self.image or "default"
+        return self.image or self.image_id or "default"
 
     def machine_or_name(self) -> str:
         return self.machine or self.name()


DaanDeMeyer (approved):

behrmann (approved):

ragazenta:
This broke my current setup with multiple profiles, I set different `ImageId` for each profile. Isn't the previous code more correct, prefer specific (`image_id`) over generic (`image`)?

<details>

<summary>My setup</summary>

```shell
mkosi.images/
├── base
│   ├── mkosi.conf
│   └── mkosi.configure
├── initrd
│   ├── mkosi.conf
│   ├── mkosi.conf.d
│   │   ├── 10-aaa
│   │   │   ├── mkosi.conf
│   │   │   └── mkosi.extra
│   │   └── 20-bbb
│   │       ├── mkosi.conf
│   │       └── mkosi.extra
│   └── mkosi.extra
└── system
```

This previously produced 3 initrd images, default `initrd`, `initrd-aaa`, `initrd-bbb`. After this commit, produces only 1 which is updated every build (with -f).
</details>

DaanDeMeyer:
No because you might want an initrd, main image and exitrd all with the same image id. @ragazenta Can you switch to Output= by any chance?

ragazenta:
> No because you might want an initrd, main image and exitrd all with the same image id. @ragazenta Can you switch to Output= by any chance?

Ah, I see. Thanks.

septatrix:
The docs for `Output=` still mention the old behaviour for determining the output name

PR #2713, opened 2024-05-15
mkosi-initrd: remove some unnecessary files from openSUSE initrd

aafeijoo-suse:
Initial clean-up, please let me know if this is not the way to go. In my tests, this reduces the size from ~100M to ~40M.

### DIFF
--- a/mkosi/resources/mkosi-initrd/mkosi.conf.d/10-opensuse.conf
+++ b/mkosi/resources/mkosi-initrd/mkosi.conf.d/10-opensuse.conf
@@ -23,5 +23,57 @@ Packages=
         util-linux
 
 RemoveFiles=
-        /usr/share/locale/*
+        /etc/bash_completion.d
+        /etc/man.conf
+        /srv
+        /usr/local/man
+        /usr/share/bash-completion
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
We can certainly add more `RemoveFiles=` but I'd prefer that anything which can be fixed with reasonable effort in packages is fixed there instead.

For example, you're removing `/usr/bin/rpm*`, but that means something is pulling in rpm as a dependency which really shouldn't be pulling in rpm as a required dependency, so I'd prefer you figure out what is pulling in rpm and fix that spec to make rpm an optional dependency. Same thing for other stuff.

`suse-module-tools` is another thing that should really be an optional dependency and not pulled in by default. And of course `zypper` should not be pulled in as a required dependency either. You get the idea.

For locales ideally it's done via a package as well. On Fedora there's `glibc-minimal-langpack` to only install C.UTF-8 for example.

Aside from those, I'd prefer that we only use `RemoveFiles=` for stuff that takes up significant space. For example I'm skeptical that /etc/zypp is going to be taking up a lot of space.

aafeijoo-suse:
Ok, understood.

PR #2718, opened 2024-05-17
mkosi-initrd: Don't remove sanitizer libraries from initrd

DaanDeMeyer:
Let's not unconditionally remove sanitizer libraries and their dependencies from the initrd as it turns out running software with sanitizers in the initrd isn't that far fetched.


bluca:
`FAILED tests/test_initrd.py::test_initrd_size - AssertionError: assert 635487...`

behrmann (approved):
The CI tests are expected, because CentOS changed a dependency a while ago, which blew up the initrd size. I'm rerunning the one Suse test, because I'm not sure what's the reason there.

behrmann:
Merging since Suse CI unborked itself and CentOS is known to have a too large initrd at the moment.

Issue #2719, opened 2024-05-20
AttributeError: 'NoneType' object has no attribute 'strip'


gdamjan:
### mkosi commit the issue has been seen with

c783b3b0718293387aa5c2b438fa17eebb84142b

### Used host distribution

Fedora rawhide

### Used target distribution

Debian 13 (trixie)

### Linux kernel version used

6.9.1-arch1-1

### CPU architectures issue was seen on

None

### Unexpected behaviour you saw

Host is ArchLinux, mkosi is from latest main branch as of today.

Exception thrown when running `mkosi build`:
```
$ ~/src/mkosi/bin/mkosi build
…
AttributeError: 'NoneType' object has no attribute 'strip'
```

### Used mkosi config

```ini
[Distribution]
Distribution=debian
Release=trixie

[Host]
ToolsTree=default
ToolsTreeDistribution=fedora
ToolsTreeRelease=rawhide
ToolsTreePackages=diffutils

[Output]
Format=disk

[Content]
Bootable=yes
Timezone=Etc/UTC
Locale=C
RootPassword=root
Packages=
        apt
        sudo
        ca-certificates
        linux-image-cloud-amd64
        systemd
        systemd-sysv
        systemd-boot
        systemd-resolved
        systemd-timesyncd
        systemd-cron
        dbus-broker
        openssh-server
        less
        vim
        iproute2
        curl

KernelCommandLine=console=tty0 console=ttyS0,115200n8 nvme_core.io_timeout=4294967295 rd.emergency=poweroff
```

### mkosi output

```sh
Traceback (most recent call last):
  File "/home/damjan/src/mkosi/mkosi/run.py", line 60, in uncaught_exception_handler
    yield
  File "/home/damjan/src/mkosi/mkosi/run.py", line 101, in fork_and_wait
    target(*args, **kwargs)
  File "/home/damjan/src/mkosi/mkosi/__init__.py", line 4414, in run_sync
    context = Context(
              ^^^^^^^^
  File "/home/damjan/src/mkosi/mkosi/context.py", line 39, in __init__
    make_tree(
  File "/home/damjan/src/mkosi/mkosi/tree.py", line 48, in make_tree
    if statfs(path.parent, sandbox=sandbox) != "btrfs":
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/damjan/src/mkosi/mkosi/tree.py", line 25, in statfs
    ).stdout.strip()
             ^^^^^
[...]

PR #2720, opened 2024-05-22
sandbox: correctly die instead of crashing if bwrap is not installed

keszybz:
Fixes https://github.com/systemd/mkosi/issues/2719.


DaanDeMeyer:
We have the check_tools() function for this. Please just add a check for bubblewrap there.

keszybz:
That works too.

behrmann (approved):

PR #2727, opened 2024-05-29
Add option to support systemd-sysupdate naming style

### LINKED ISSUE #1638: Make UKI naming configurable
I'd like to update the UKIs with systemd-sysupdate at a later point. Including the kernel version in the UKI name makes this tasks harder, as sysupdate does not support (actual asterisk) wildcards and its kernel version specifier is rather limited.  
Alternatively, support for kernel "flavours" could be useful.

I do not yet have an idea for an elegant configuration option for this though.

If there is an easier option to rename generated UKIs after the fact, please tell me.

mcassaniti:
When using `systemd-sysupdate` the names of output files need to contain the image version since this is the only version that is reasonable to use across all artifacts. This change makes sure the UKI stored in the ESP or XBOOTLDR partition has an appropriate name, along with the name if `SplitArtifacts=` is set.

Note: The UKI cannot contain the kernel version or other values that will change and cannot be hard-coded into a `sysupdate.d` file.

Fixes #1638.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2312,10 +2312,15 @@ def install_uki(context: Context, kver: str, kimg: Path, token: str, partitions:
         else:
             boot_binary = context.root / efi_boot_binary(context)
     else:
-        if roothash:
-            _, _, h = roothash.partition("=")
-            boot_binary = context.root / f"boot/EFI/Linux/{token}-{kver}-{h}{boot_count}.efi"
+        if context.config.sysupdate_naming:
+            # Use the image version as the authoritative version
+            # Only one kernel version is supported
+            name = f"{context.config.image_id}_{context.config.image_version}{boot_count}"
+            boot_binary = context.root / f"boot/EFI/Linux/{name}.efi"
         else:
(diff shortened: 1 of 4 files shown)


mcassaniti:
What do you recommend? I can guess when the naming should change but that likely to catch users by surprise. I know a finalize script can cover the ESP rename, but it won't work for the output file name. I may have the wrong attitude here, but I'd assume that a project under the systemd umbrella should be able to output artifacts in a format that is accepted by the rest of the project (i.e. mkosi should do the work rather than the end user).

I also don't know if I broke the CI results or if it's something else.

mcassaniti:
It seems that #2731 does in fact cover the requirements!!! I'm more than happy to be wrong.

The partial configuration below is required once the other patch lands. The output controls the UKI output while `UnifiedKernelImageName` controls the file name within the ESP/XBOOTLDR partition.

```INI
[Output]
ImageId = ...
ImageVersion = ...
Output = %i_%v

[Content]
UnifiedKernelImageName = %i_%v
```

PR #2731, opened 2024-05-29
Add `UnifiedKernelImageName=`

NekkoDroid:
This is a more scoped down version of #2727 to just be able to adjust the name of the UKI.

This possibly could be extended in the future to allow some form of templating for multiple kernels, but that effort is probably better put into adding the missing parts in kernel-install.

Fixes: #1638

### DIFF
--- a/mkosi/resources/mkosi.md
+++ b/mkosi/resources/mkosi.md
@@ -990,6 +990,12 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     Otherwise Type 1 entries as defined by the Boot Loader Specification
     will be used instead. If disabled, Type 1 entries will always be used.
 
+`UnifiedKernelImageName=`, `--unified-kernel-image-name=`
+
+: Specify the name to install the unified kernel image as. Takes a filename
+  without any path components or the `.efi` extension. This is only supported
+  when using a single unified kernel image.
+
 `Initrds=`, `--initrd`
 :   Use user-provided initrd(s). Takes a comma separated list of paths to initrd
     files. This option may be used multiple times in which case the initrd lists
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2312,7 +2312,9 @@ def install_uki(context: Context, kver: str, kimg: Path, token: str, partitions:
         else:
             boot_binary = context.root / efi_boot_binary(context)
     else:
-        if roothash:
+        if name := context.config.unified_kernel_image_name:
+            boot_binary = context.root / f"boot/EFI/Linux/{name}{boot_count}.efi"
+        elif roothash:
             _, _, h = roothash.partition("=")
             boot_binary = context.root / f"boot/EFI/Linux/{token}-{kver}-{h}{boot_count}.efi"
         else:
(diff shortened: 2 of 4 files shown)


michaelbeaumont:
Great! It's probably handled generically for all settings but can this setting in particular also be parameterized with specifiers like ImageId/ImageVersion?

mcassaniti:
I'm sorry to say that this doesn't fix #1638. The file name in the ESP and the output file name both need to be changed and they need to include the image version to fix the issue. File name templates maybe?

NekkoDroid:
> Great! It's probably handled generically for all settings but can this setting in particular also be parameterized with specifiers like ImageId/ImageVersion?

Yes, you can by setting it to `UnifiedKernelImageName=%i_%v`.

> The file name in the ESP and the output file name both need to be changed and they need to include the image version to fix the issue.

You can already set the output name for all the output files by changing the `Output=` to `Output=%i_%v`.

mcassaniti:
I'll test and get back to you. It's taken me too long to work out that specifiers are not just for things like `[Match]` blocks.

mcassaniti:
Tested successfully. See comments on #1638.

mcassaniti:
Can I suggest adding a piece to the documentation so other users don't get so lost looking at how to support `systemd-sysupdate`? Something like "One use case is setting the name to `%i_%v` for use with `systemd-sysupdate` and having `MatchPattern = %M_@v.efi` in your sysupdate.d kernel configuration."

Perhaps it's time mkosi had some documented examples where there is an explanation to go with the example. Bonus points if the examples go through CI. I'd happily write one for systemd-sysupdate.

septatrix:
Looks like a decent way to go forward with this while it gets figured out how this could be natively supported by `kernel-install`

DaanDeMeyer:
Adding don't-merge until we get to discuss this in the systemd biweekly.

DaanDeMeyer:
At the very least this is going to need a second level of specifiers, that are expanded just before we actually write the UKI. This option should also be made more generic so it applies to Type1 entries as well.

The second level of specifiers should include the following:

- Roothash specifier (%h)
- Entry token specifier (%e)
- Kernel version specifier (%k)
- Boot count specifier (%c)
- ...

DaanDeMeyer:
> @DaanDeMeyer any other specifiers to add? I can't think or see any other ones of relevance.
> 
> And would it be OK to use `@` as the specifier here as to not clash with the default specifiers? I was thinking of `@` since that is already what sysupdate for example uses for its secondary specifiers (or are those the primary?).
> 
> And regarding Type1: currently those kernel images are organized in `/boot/<token>/<kver>`, which with the current "no path components" limitation would make this sort of organizing impossible. Should those all then simply live in `/boot/<name>`?
> 
> And any specific wish for the renamed option name other than `KernelImageFormat=`?
> 
> I guess the specifiers are probably better handled/discussed in #2732?

> And would it be OK to use @ as the specifier here as to not clash with the default specifiers? I was thinking of @ since that is already what sysupdate for example uses for its secondary specifiers (or are those the primary?).

No because then we could conflict with sysupdate in the future. I'm partial to just using % as well, with the caveat that in the config file you'd have to escape the % so it doesn't get malparsed by the config specifiers (by doing %%h).

> And regarding Type1: currently those kernel images are organized in `/boot/<token>/<kver>`, which with the current "no path components" limitation would make this sort of organizing impossible. Should those all then simply live in `/boot/<name>`?

It's done like this because kernel-install does the same. I'd prefer to not deviate from what kernel-install does too much. See boot loader specification for type 1 entries.

> And any specific wish for the renamed option name other than `KernelImageFormat=`?

I think `BootloaderEntryFormat=` is better because the boot loader specification refers to both type 1 and type 2 as bootloader entries.

> I guess the specifiers are probably better handled/discussed in https://github.com/systemd/mkosi/pull/2732?

I think @behrmann is OK if we merge that logic into this PR?

behrmann:
Totally fine with that. Feel free to just cherry-pick it and rebase on top

I just pushed a change switching to the above letters for the specifiers and made it `&` instead of `%`. Not `@` so as not to confuse people further with overlap with sysupdate and to sidestep the specifiers we always expand. Also has the nice upside, that its a separate namespace for letters.

mkosi/resources/mkosi.md
```diff
@@ -990,6 +990,12 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     Otherwise Type 1 entries as defined by the Boot Loader Specification
     will be used instead. If disabled, Type 1 entries will always be used.
 
+`UnifiedKernelImageName=`, `--unified-kernel-image-name=`
+
+: Specify the name to install the unified kernel image as. Takes a filename
+  without any path components or the `.efi` extension. This is only supported
+  when using a single unified kernel image.
```
behrmann, line 997: I finally fixed this the other day ;)
```suggestion
`UnifiedKernelImageName=`, `--unified-kernel-image-name=`
:   Specify the name to install the unified kernel image as. Takes a filename
    without any path components or the `.efi` extension. This is only supported
    when using a single unified kernel image.
```

PR #2734, opened 2024-05-30
Add option to skip passing --seed to systemd-repart

michaelbeaumont:
The use case here is if we've already set `/etc/machine-id` in our image, for example if `/` is meant to be read-only. At the moment it's not possible to use the DDI with `SD_GPT_VAR`, because the partition UUID generated by `systemd-repart` won't match `/etc/machine-id`, `--seed` is always passed and overrides it. This PR allows setting `Seed=no` and adding a `mkosi.repart/var.conf` like:

```
[Partition]
Type=var
CopyFiles=/var:/
SizeMinBytes=1G
```

so that the generated image will auto mount that partition to `/var`.
The current default, where `mkosi` itself generates a UUID and passes it via `--seed`, is left as the default.

Unfortunately this only works with an already set `/etc/machine-id` because of [ordering between `gpt-auto-generator` and `systemd-repart` at start up](https://github.com/systemd/systemd/issues/31071). Also relevant: https://github.com/systemd/mkosi/discussions/2304

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3192,7 +3192,6 @@ def make_image(
         "--json=pretty",
         "--no-pager",
         f"--offline={yes_no(context.config.repart_offline)}",
-        "--seed", str(context.config.seed),
         context.staging / context.config.output_with_format,
     ]
     mounts = [Mount(context.staging, context.staging)]
@@ -3227,6 +3226,8 @@ def make_image(
             "--generate-fstab=/etc/fstab",
             "--generate-crypttab=/etc/crypttab",
         ]
+    if context.config.seed:
+        cmdline += ["--seed", str(context.config.seed)]
 
     for d in definitions:
         cmdline += ["--definitions", d]
@@ -3480,7 +3481,6 @@ def make_extension_image(context: Context, output: Path) -> None:
         "--dry-run=no",
         "--no-pager",
         f"--offline={yes_no(context.config.repart_offline)}",
-        "--seed", str(context.config.seed) if context.config.seed else "random",
         "--empty=create",
         "--size=auto",
         "--definitions", r,
@@ -3510,6 +3510,8 @@ def make_extension_image(context: Context, output: Path) -> None:
         cmdline += ["--sector-size", str(context.config.sector_size)]
     if context.config.split_artifacts:
         cmdline += ["--split=yes"]
+    if context.config.seed:
+        cmdline += ["--seed", str(context.config.seed)]
 
     with complete_step(f"Building {context.config.output_format} extension image"):
         j = json.loads(
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -4119,7 +4119,9 @@ def json_type_transformer(refcls: Union[type[Args], type[Config]]) -> Callable[[
     def path_list_transformer(pathlist: list[str], fieldtype: type[list[Path]]) -> list[Path]:
         return [Path(p) for p in pathlist]
 
-    def uuid_or_false_transformer(uuid_or_false_str: str, fieldtype: type[Union[uuid.UUID, Literal[False]]]) -> Union[uuid.UUID | Literal[False]]:
+    def uuid_or_false_transformer(
+        uuid_or_false_str: str, fieldtype: type[Union[uuid.UUID, Literal[False]]],
+    ) -> Union[uuid.UUID | Literal[False]]:
         if uuid_or_false_str in {"0", "false", "no", "n", "f", "off", "never"}:
             return False
         return uuid.UUID(uuid_or_false_str)
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -4119,8 +4119,10 @@ def json_type_transformer(refcls: Union[type[Args], type[Config]]) -> Callable[[
     def path_list_transformer(pathlist: list[str], fieldtype: type[list[Path]]) -> list[Path]:
         return [Path(p) for p in pathlist]
 
-    def uuid_transformer(uuidstr: str, fieldtype: type[uuid.UUID]) -> uuid.UUID:
-        return uuid.UUID(uuidstr)
+    def uuid_or_false_transformer(uuid_or_false_str: str, fieldtype: type[Union[uuid.UUID, Literal[False]]]) -> Union[uuid.UUID | Literal[False]]:
+        if uuid_or_false_str in {"0", "false", "no", "n", "f", "off", "never"}:
+            return False
+        return uuid.UUID(uuid_or_false_str)
 
     def root_password_transformer(
         rootpw: Optional[list[Union[str, bool]]], fieldtype: type[Optional[tuple[str, bool]]]
@@ -4189,7 +4191,7 @@ def json_type_transformer(refcls: Union[type[Args], type[Config]]) -> Callable[[
         Path: path_transformer,
         Optional[Path]: optional_path_transformer,
         list[Path]: path_list_transformer,
-        uuid.UUID: uuid_transformer,
+        Union[uuid.UUID, Literal[False]]: uuid_or_false_transformer,
         Optional[tuple[str, bool]]: root_password_transformer,
         list[ConfigTree]: config_tree_transformer,
         tuple[str, ...]: str_tuple_transformer,
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -29,7 +29,7 @@ import uuid
 from collections.abc import Collection, Iterable, Iterator, Sequence
 from contextlib import AbstractContextManager
 from pathlib import Path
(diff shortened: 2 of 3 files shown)


DaanDeMeyer:
Ugh, I'd prefer we didn't add support for this, shipping with a fixed machine ID really shouldn't happen. Can't you just derive the seed from the machine ID and pass that as a fixed seed?

michaelbeaumont:
> shipping with a fixed machine ID really shouldn't happen.
> Can't you just derive the seed from the machine ID and pass that as a fixed seed?

Why exactly? Maybe I'm missing background philosophy here but I don't see why this is such a bad thing. How else can read only systems be done cleanly? To be clear, I'm generating the machine id using `systemd-firstboot --root="${BUILDROOT}" --setup-machine-id`. So it's fixed at build time, not before build time.

In fact the working alternative is to just use the seed generated by `mkosi` as the machine ID. Though somehow I don't find the flow `mkosi seed`->`machine-id` and `mkosi seed`->`systemd-repart seed` very satisfying. Like isn't it cleaner to have this feature and have the flow `mkosi seed`->`machine-id`->`systemd-repart seed` where `systemd-repart` just does the right thing, without overriding it with `--seed`.

Maybe it's the default that tripped me up, if I don't set `Seed=` then why would I want _mkosi_ to generate a seed itself and pass that to `systemd-repart`?
The current behavior is just very surprising/basically buggy if you do end up setting `/etc/machine-id`. And using `--setup-machine-id` seems very valid given that it's an option to `systemd-repart`.

It seems to me that `systemd-firstboot`/`systemd-repart` is meant to handle the "generate my machine ID" & "use my machine ID for partition UUIDs" use case, or is `--root` & `--setup-machine-id`/not using `--seed` discouraged?

DaanDeMeyer:
> > shipping with a fixed machine ID really shouldn't happen.
> > Can't you just derive the seed from the machine ID and pass that as a fixed seed?
> 
> Why exactly? Maybe I'm missing background philosophy here but I don't see why this is such a bad thing. How else can read only systems be done cleanly? To be clear, I'm generating the machine id using `systemd-firstboot --root="${BUILDROOT}" --setup-machine-id`. So it's fixed at build time, not before build time.
> 
> In fact the working alternative is to just use the seed generated by `mkosi` as the machine ID. Though somehow I don't find the flow `mkosi seed`->`machine-id` and `mkosi seed`->`systemd-repart seed` very satisfying. Like isn't it cleaner to have this feature and have the flow `mkosi seed`->`machine-id`->`systemd-repart seed` where `systemd-repart` just does the right thing, without overriding it with `--seed`.
> 
> Maybe it's the default that tripped me up, if I don't set `Seed=` then why would I want _mkosi_ to generate a seed itself and pass that to `systemd-repart`? The current behavior is just very surprising/basically buggy if you do end up setting `/etc/machine-id`. And using `--setup-machine-id` seems very valid given that it's an option to `systemd-repart`.
> 
> It seems to me that `systemd-firstboot`/`systemd-repart` is meant to handle the "generate my machine ID" & "use my machine ID for partition UUIDs" use case, or is `--root` & `--setup-machine-id`/not using `--seed` discouraged?

You just let mkosi write "uninitialized" to /etc/machine-id and systemd will mount a transient machine-id over it during runtime. If you want a machine to have a fixed machine ID across reboots, you should set `systemd.machine_id=` on the kernel command line instead.

As for `SD_GPT_VAR`, you really shouldn't put a /var partition inside your image. Instead it should be generated on first boot by systemd-repart after switch-root.

michaelbeaumont:
> As for SD_GPT_VAR, you really shouldn't put a /var partition inside your image. Instead it should be generated on first boot by systemd-repart after switch-root.

[This doesn't work (well)](https://github.com/systemd/systemd/issues/31071#issuecomment-2089273177)

> If you want a machine to have a fixed machine ID across reboots, you should set systemd.machine_id= on the kernel command line instead.

Do you know why exactly? Why are there now two places to persist a machine id? Certainly "a fixed machine ID across reboots" is what setting `/etc/machine-id` does?
Just curious what the thinking is there.

And still... why always pass `--seed` to `systemd-repart` and by default?

DaanDeMeyer:
> > As for SD_GPT_VAR, you really shouldn't put a /var partition inside your image. Instead it should be generated on first boot by systemd-repart after switch-root.
> 
> [This doesn't work (well)](https://github.com/systemd/systemd/issues/31071#issuecomment-2089273177)

No it does not but that's what should be fixed instead of trying to do weird workarounds.

> > If you want a machine to have a fixed machine ID across reboots, you should set systemd.machine_id= on the kernel command line instead.

> Do you know why exactly? Why are there now two places to persist a machine id? Certainly "a fixed machine ID across reboots" is what setting `/etc/machine-id` does? Why is it _so_ important that `/etc/machine-id` is only written on boot. I mean mkosi runs `systemd-firstboot` itself, so it's already emulating parts of booting. Just curious what the thinking is there.

We call systemd-firstboot to configure other things such as root password, not to configure the machine ID..

See https://www.freedesktop.org/software/systemd/man/latest/machine-id.html for more information on the machine ID and the design behind it.

> And still... why always pass `--seed` to `systemd-repart` and by default?

Because otherwise systemd-repart uses random seeds and since we invoke repart multiple times during a build, the seed is different every time leading to very weird bugs. So mkosi calculates a random seed itself which it passes to repart to make sure the seed is random but the same for each call to systemd-repart in a single build.

To me it seems like the only reason you're adding this is to work around https://github.com/systemd/systemd/issues/31071#issuecomment-2089273177. Instead I think https://github.com/systemd/systemd/issues/31071#issuecomment-2089273177 should just be fixed in systemd. I don't have time to do that myself, but you're welcome to propose and contribute a fix to systemd so that `SD_GPT_VAR` works properly for your use case.

michaelbeaumont:
> We call systemd-firstboot to configure other things such as root password, not to configure the machine ID.

Yes. That is what happens. But _why_ is it then not right to call `systemd-firstboot` to configure the machine ID? mkosi is already 'faking' a first boot, and writing a machine ID is part of first boot, so why is it so unthinkable to write a machine ID with mkosi?

Indeed, from reading `INITIALIZATION` in `man machine-id` there's really nothing to say that `/etc/machine-id` shouldn't be written when generating an image and quite a bit to suggest that it's completely valid.

> For normal operating system installations, where a custom image is created for a specific machine, /etc/machine-id should be populated during installation.
> systemd-machine-id-setup(1) may be used by installer tools to initialize the machine ID at install time, but /etc/machine-id may also be written using any other means.


I dunno, that reads to me like exactly like my strategy here, depending on how you define "installation".

> since we invoke repart multiple times during a build,

OK, that's indeed a good reason for the default.

I mean I guess the workaround of extracting the Seed from the mkosi config where I write the machine ID works for me.

I'd still like to make a change to the docs because IMO it is really not clear that the when `Seed=` is unset, mkosi will still use its own even if `/etc/machine-id` is set.

DaanDeMeyer:
Documenting this sounds good to me.

michaelbeaumont:
I'll open a PR to update the `Seed=` docs or maybe where systemd-repart is introduced.

I think there's also an issue also with docs around machine-id in general, either in systemd or here in mkosi, because I _really_ don't see where machine-id shouldn't be set in an image, again given:

> For normal operating system installations, where a custom image is created for a specific machine, /etc/machine-id should be populated during installation.

> systemd-firstboot(1) may be used to initialize /etc/machine-id on mounted (but not booted) system images.

Sorry to stress this so much but it's quite confusing and unclear, in my opinion.

If mkosi is not meant for when "a custom image is created for a specific machine" then it should be called out.

DaanDeMeyer:
> I'll open a PR to update the `Seed=` docs or maybe where systemd-repart is introduced.
> 
> I think there's also an issue also with docs around machine-id in general, either in systemd or here in mkosi, because I _really_ don't see where machine-id shouldn't be set in an image, again given:
> 
> > For normal operating system installations, where a custom image is created for a specific machine, /etc/machine-id should be populated during installation.
> 
> > systemd-firstboot(1) may be used to initialize /etc/machine-id on mounted (but not booted) system images.
> 
> Sorry to stress this so much but it's quite confusing and unclear, in my opinion.
> 
> If mkosi is not meant for when "a custom image is created for a specific machine" then it should be called out.

We generally intend to build generic images that are not tied to a specific machine. This doesn't mean that it can't be used to build images for a specific machine, just that we're not going to merge very specific fixes specifically for this use case if another workaround is possible.

In this case, it's perfectly possible to set `Seed=` manually so I'm not willing to merge your PR to allow disabling it just for this use case.

michaelbeaumont:
> We generally intend to build generic images that are not tied to a specific machine
> if another workaround is possible.

OK, then I'll add this to the docs in some form as well.

DaanDeMeyer:
Let's close this, feel free to open a new PR to document this

PR #2735, opened 2024-05-31
fedora: use F40

keszybz:


DaanDeMeyer (approved):

DaanDeMeyer:
No need to run CI since it is already running Fedora 40

PR #2742, opened 2024-06-02
Have coredumpctl and journalctl operate on forwarded journal if available

DaanDeMeyer:
If ForwardJournal= is configured, have coredumpctl and journalctl operate on it instead of on the image itself. While this doesn't handle the edge case where the journal is forwarded but the coredumps are stored in the image, let's assume that users that enable ForwardJournal= will also configure coredumps to be stored in the journal.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -4029,12 +4029,13 @@ def run_shell(args: Args, config: Config) -> None:
 
 
 def run_systemd_tool(tool: str, args: Args, config: Config) -> None:
-    if config.output_format not in (OutputFormat.disk, OutputFormat.directory):
+    if config.output_format not in (OutputFormat.disk, OutputFormat.directory) and not config.forward_journal:
         die(f"{config.output_format} images cannot be inspected with {tool}")
 
     if (
         args.verb in (Verb.journalctl, Verb.coredumpctl)
         and config.output_format == OutputFormat.disk
+        and not config.forward_journal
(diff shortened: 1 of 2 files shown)


mkosi/__init__.py
```diff
@@ -4029,34 +4029,44 @@ def run_shell(args: Args, config: Config) -> None:
...
 def run_systemd_tool(tool: str, args: Args, config: Config) -> None:
-    if config.output_format not in (OutputFormat.disk, OutputFormat.directory):
+    if config.output_format not in (OutputFormat.disk, OutputFormat.directory) and not config.forward_journal:
         die(f"{config.output_format} images cannot be inspected with {tool}")
 
     if (
         args.verb in (Verb.journalctl, Verb.coredumpctl)
         and config.output_format == OutputFormat.disk
+        and not config.forward_journal
         and os.getuid() != 0
```
behrmann, line 4039: Looking at this, shouldn't we check whether the user is in `systemd-journal` group as an alternative to the root check (for the `journalctl` verb)?

behrmann (commented):
Looks good, but storing the coredumps in the journal should be mentioned in the man page in the `ForwardJournal=` section.

mkosi/__init__.py
DaanDeMeyer: @behrmann This check is about being able to mount the disk image

DaanDeMeyer:
Only documentation changes, merging

PR #2744, opened 2024-06-02
ci: Switch to Ubuntu Noble

DaanDeMeyer:


DaanDeMeyer:
The opensuse and ubuntu tools tree failures are because they're missing https://github.com/systemd/systemd/pull/32892 in their build of systemd. opensuse should get v255.7 with the fix soon. Will need to get this backported to ubuntu as well.

behrmann (approved):

PR #2747, opened 2024-06-05
Make sure we don't fail when there is no sdmagic section in sd-stub

DaanDeMeyer:
The sdmagic section in sd-stub was only introduced in systemd 250. Since Ubuntu Jammy ships systemd 249, let's make sure we gracefully handle the scenario where we can't find the sdmagic section.


behrmann (approved):
LGTM, but let's add a comment `extract_pe_section`, that this is a workaround for pre-250, so that we can drop it again at some point .

PR #2748, opened 2024-06-05
dnf: Enable versionlock plugin by default

DaanDeMeyer:
Let's allow users to make use of the versionlock plugin by enabling it by default. To make sure it doesn't fail, we write a noop configuration that makes the plugin do nothing at all which users can then override using PackageManagerTrees=.


behrmann (approved):

behrmann:
The failing CI is currently expected until their respective fixes move into their repos. Merging.

PR #2756, opened 2024-06-06
Introduce "default" and "host" for kernel modules include settings

DaanDeMeyer:


mkosi/resources/mkosi.md
```diff
@@ -1047,21 +1047,22 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
-    are included in the image as well. This setting takes priority over `KernelModulesExclude=` and only makes
+    are included in the image as well.
+
+:   If the special value `default` is used, the default kernel modules
+    defined in the `mkosi-initrd` configuration are included as well.
+
+:   If the special value `host` is used, the currently loaded modules on
+    the host system are included as well.
+
+:   This setting takes priority over `KernelModulesExclude=` and only makes
```
behrmann, line 1058:
```suggestion
    If the special value `default` is used, the default kernel modules
    defined in the `mkosi-initrd` configuration are included as well.

    If the special value `host` is used, the currently loaded modules on
    the host system are included as well.

    This setting takes priority over `KernelModulesExclude=` and only makes
```

mkosi/__init__.py
```diff
@@ -1897,6 +1897,17 @@ def build_microcode_initrd(context: Context) -> list[Path]:
     return [microcode]
 
 
+def finalize_kernel_modules_include(context: Context, *, include: Sequence[str], host: bool) -> list[str]:
```
behrmann, line 1900: The order of the includes shouldn't be important, since it only goes into a regex, right? Then we should maybe make this a set to deduplicate.

behrmann (commented):
Nice! Consolidating he options should make this easier to follow for new users. Should we maybe namespace the constants there, e.g. `mkosi.default`/`mkosi.host` (or with dashes instead of dots)? Just to be on the safe side, that a kernel module doesn't get one of the names by accident.

ruff is complaining about the imports not being sorted.

PR #2761, opened 2024-06-06
Make --tools-tree the same as --tools-tree=default

DaanDeMeyer:
Saves on typing and makes it easier to use overall.


behrmann (approved):

behrmann:
That's true, that's very unfortunate, since that's more niche.

PR #2771, opened 2024-06-09
Add `mkosi.postoutput` scripts

NekkoDroid:


mkosi/resources/mkosi.md
```diff
@@ -2023,6 +2029,13 @@ current working directory. The following scripts are supported:
   artifacts from `SplitArtifacts=yes` or RPMs built in a build script).
   Note that this script does not use the tools tree even if one is configured.
 
+* If **`mkosi.postoutput`** (`PostOutputScripts=`) exists, it is executed right
```
behrmann, line 2032: Nit: this point should maybe go above the `mosi.clean` script, since logically cleaning things will be a thing that happens after having generated outputs.

mkosi/config.py
```diff
@@ -2290,6 +2291,17 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
...
+    ConfigSetting(
+        dest="postoutput_scripts",
+        long="--postoutput-script",
+        metavar="PATH",
+        name="PostOutputScripts",
+        section="Content",
+        parse=config_make_list_parser(delimiter=",", parse=make_path_parser()),
+        paths=("mkosi.postoutput",),
+        path_default=False,
+        help="Output introspection script to run outside image",
```
behrmann, line 2303: I'm not sure introspection is the best word to use here, at least alone, since this is supposed to also modify outputs, right?

mkosi/resources/mkosi.md
behrmann: This also needs a column in the table for environment variables below.

mkosi/resources/mkosi.md
```diff
@@ -2023,6 +2029,13 @@ current working directory. The following scripts are supported:
...
+  after all the output files have been generated, before they are finally
+  moved into the output directory. A post-output script can be used to
+  introspect the generated output files to modify any `BuildSource=` mount
+  (e.g. writing a custom `SHA256SUMS` file). Note that this script does not use
```
behrmann, line 2036: Typo: `BuildSource=` → `BuildSources=`.

I'm not entirely sure I understand how the `BuildSources=` come in here, since the example is about `SHA256SUMS` and this seems a bit incongruent, i.e. If this is about the sums and signing, why would I want to modify the build sources at this point? On that note: signing should probably be mentioned here as well.

behrmann (commented):
Nice addition!

mkosi/config.py
NekkoDroid: Currently I have it that the `/work/out` (`$OUTPUTDIR`) is mounted read-only, as I would expect that the output shouldn't be modified after it was built, hence only introspect the output.

Although I can change that if its too strict.

mkosi/resources/mkosi.md
NekkoDroid: With the output dir mounted read-only that is somewhat the only option if you want to write some custom output persistently.

I guess the `SHA256SUMS` example was a bit vague and "me usecase" specific, but its to *not* use the integrated `SHA256SUMS` file as that always gets removed during clean, only contains the info for a specific image and is prepended with the image name.

My intended usecase is to append the sums to a common `SHA256SUMS` file (and only remove the lines for a specific image in `mkosi.clean`) so that it can be consumed by `sd-sysupdate`.

behrmann: > With the output dir mounted read-only that is somewhat the only option if you want to write some custom output persistently.

> Currently I have it that the /work/out ($OUTPUTDIR) is mounted read-only, as I would expect that the output shouldn't be modified after it was built, hence only introspect the output.

Mhm, I see the point about the outputs not being changed after building, but writing a different `SHA256SUMS` is, I think, still producing an output and staging that in `BuildSources=` to persist it, does not seem ideal to me, one of the reasons being interaction with the ephemeral build sources option.

NekkoDroid: > one of the reasons being interaction with the ephemeral build sources option.

This is similarly handled as with the `mkosi.sync` scripts, where the build sources are always mounted non-ephemeral https://github.com/NekkoDroid/mkosi/blob/c087b9cb1190ca754e2126596c4dda745c178eaf/mkosi/__init__.py#L856. If we want to keep it this way I probably should mention this in the docs.

Anyway, my usecase doesn't exactly hinge on the output being read-only or not, but I do need some way to be able to access 1 directory up from the output dir (on host) without it being ephemeral. And this would currently be the main way I'd see doing this without a rework how mkosi handles its `SHA256SUMS` file.

mkosi/config.py
DaanDeMeyer: The thing is that this script might want to add more outputs of its own. SBOM manifest or something could be a good example. So let's keep the output directory writable.

mkosi/resources/mkosi.md
DaanDeMeyer: Build sources definitely need to be ephemeral in the postoutput scripts if the setting is used. These scripts are specifically intended to operate on the outputs, not on the sources so it does not make sense for the sources to not be ephemeral when running output scripts.

mkosi/config.py
NekkoDroid: And suggestions on what to use for the description, my mind is blanking?

behrmann: How about
```suggestion
        help="Output postprocessing script to run outside image",
```

DaanDeMeyer (changes requested):

DaanDeMeyer (approved):

mkosi/__init__.py
```diff
@@ -830,6 +830,47 @@ def run_finalize_scripts(context: Context) -> None:
...
+
+    env = dict(
+        DISTRIBUTION=str(context.config.distribution),
+        RELEASE=context.config.release,
+        ARCHITECTURE=str(context.config.architecture),
+        SRCDIR="/work/src",
+        OUTPUTDIR="/work/out",
+        MKOSI_UID=str(INVOKING_USER.uid),
+        MKOSI_GID=str(INVOKING_USER.gid),
+        MKOSI_CONFIG="/work/config.json",
```
behrmann, line 845: These three are still missing in the environment variable table.

NekkoDroid: Dunno how I mist those

DaanDeMeyer:
We should really add some code reuse between all the scripts at some point

PR #2772, opened 2024-06-09
Set default output name for `mkosi-tools` to `mkosi.tools`

NekkoDroid:
This should allow one to invoke `mkosi --directory "" --include mkosi-tools` in the project root to create a tools tree that is automatically detected and used in subsequent invocations.

### DIFF
--- a/mkosi/resources/mkosi-tools/mkosi.conf
+++ b/mkosi/resources/mkosi-tools/mkosi.conf
@@ -2,6 +2,7 @@
 
 [Output]
 Format=directory
+@Output=mkosi.tools
 ManifestFormat=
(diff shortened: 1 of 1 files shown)


behrmann (approved):
Not really my workflow, I tend to write a config and just put in `ToolsTree=default` in there, but LGTM.

NekkoDroid:
The main reason I'd prefer not to use `ToolsTree=default` in the image I am currently working on is

1. the output is grouped in `mkosi.output/v%v`, so every new version would have its own copy
2. I'd prefer if I could just take `mkosi.output` and host it without too many extra files

PR #2779, opened 2024-06-12
Shell completion

behrmann:
That's a rough draft I had lying around for a while now and polished up on the train. It needs more completers for actions and I haven't thought about whether the assumptions for the bash completion are sound and the fish completion is written looking just at the docs.

The idea is to be able to generate a completion script that distros can "build" during packaging and ship pre-built.

### DIFF
--- a/mkosi/completion.py
+++ b/mkosi/completion.py
@@ -1,0 +1,201 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+import argparse
+import dataclasses
+import io
+import shlex
+from collections.abc import Iterable, Mapping
+from pathlib import Path
+from typing import Optional, Union
+
+from mkosi.config import SETTINGS, SETTINGS_LOOKUP_BY_DEST, Args, CompGen, Verb, create_argument_parser
+from mkosi.log import die
+
+
+@dataclasses.dataclass(frozen=True)
+class CompletionItem:
+    short: Optional[str]
+    long: Optional[str]
+    help: Optional[str]
+    nargs: Union[str, int]
+    choices: list[str]
+    compgen: CompGen
+
+
+def collect_completion_arguments() -> list[CompletionItem]:
+    parser = create_argument_parser()
+
+    options = [
+        CompletionItem(
+            short=next((s for s in action.option_strings if not s.startswith("--")), None),
+            long=next((s for s in action.option_strings if s.startswith("--")), None),
+            help=action.help,
+            nargs=action.nargs or 0,
+            choices=[str(c) for c in action.choices] if action.choices is not None else [],
+            compgen=CompGen.from_action(action),
+        )
+        for action in parser._actions
+        if (action.option_strings and
+            action.help != argparse.SUPPRESS and
+            action.dest not in SETTINGS_LOOKUP_BY_DEST)
+    ]
+
+    options += [
+        CompletionItem(
+            short=setting.short,
+            long=setting.long,
+            help=setting.help,
+            nargs=setting.nargs or 1,
+            choices=[str(c) for c in setting.choices] if setting.choices is not None else [],
+            compgen=setting.compgen,
+        )
+        for setting in SETTINGS
+    ]
+
+    return options
+
+
+def finalize_completion_bash(options: list[CompletionItem], resources: Path) -> str:
+    def to_bash_array(name: str, entries: Iterable[str]) -> str:
+        return f"declare -a {name.replace('-', '_')}=(" + " ".join(shlex.quote(str(e)) for e in entries) + ")"
+
+    def to_bash_hasharray(name: str, entries: Mapping[str, Union[str, int]]) -> str:
+        return (
+            f"declare -A {name.replace('-', '_')}=(" +
+            " ".join(f"[{shlex.quote(str(k))}]={shlex.quote(str(v))}" for k, v in entries.items()) + ")"
+        )
+
+    completion = resources / "completion.bash"
+
+    options_by_key = {o.short: o for o in options if o.short} | {o.long: o for o in options if o.long}
+
+    with io.StringIO() as c:
+        c.write("# SPDX-License-Identifier: LGPL-2.1-or-later\n\n")
+        c.write(to_bash_array("_mkosi_options", options_by_key.keys()))
+        c.write("\n\n")
+
+        nargs = to_bash_hasharray("_mkosi_nargs", {optname: v.nargs for optname, v in options_by_key.items()})
+        c.write(nargs)
+        c.write("\n\n")
+
+        choices = to_bash_hasharray(
+            "_mkosi_choices", {optname: " ".join(v.choices) for optname, v in options_by_key.items() if v.choices}
+        )
+        c.write(choices)
+        c.write("\n\n")
+
+        compgen = to_bash_hasharray(
+            "_mkosi_compgen",
+            {optname: v.compgen.to_bash() for optname, v in options_by_key.items() if v.compgen != CompGen.default},
+        )
+        c.write(compgen)
+        c.write("\n\n")
+
+        c.write(to_bash_array("_mkosi_verbs", [str(v) for v in Verb]))
+        c.write("\n\n\n")
+
+        c.write(completion.read_text())
+
+        return c.getvalue()
+
+
+def finalize_completion_fish(options: list[CompletionItem], resources: Path) -> str:
+    with io.StringIO() as c:
+        c.write("# SPDX-License-Identifier: LGPL-2.1-or-later\n\n")
+        c.write("complete -c mkosi -f\n")
+
+        c.write("complete -c mkosi -n '__fish_is_first_token' -a \"")
+        c.write(" ".join(str(v) for v in Verb))
+        c.write("\"\n")
+
+        for option in options:
+            if not option.short and not option.long:
+                continue
+
+            c.write("complete -c mkosi ")
+            if option.short:
+                c.write(f"-s {option.short.lstrip('-')} ")
+            if option.long:
+                c.write(f"-l {option.long.lstrip('-')} ")
+            if isinstance(option.nargs, int) and option.nargs > 0:
+                c.write("-r ")
+            if option.choices:
+                c.write("-a \"")
+                c.write(" ".join(option.choices))
+                c.write("\" ")
+            if option.help is not None:
+                help = option.help.replace("'", "\\'")
+                c.write(f"-d \"{help}\" ")
+            c.write(option.compgen.to_fish())
+            c.write("\n")
+        return c.getvalue()
+
+
+def finalize_completion_zsh(options: list[CompletionItem], resources: Path) -> str:
+    def to_zsh_array(name: str, entries: Iterable[str]) -> str:
+        return f"declare -a {name.replace('-', '_')}=(" + " ".join(shlex.quote(str(e)) for e in entries) + ")"
+
+    completion = resources / "completion.zsh"
+
+    with io.StringIO() as c:
+        c.write("#compdef mkosi\n")
+        c.write("# SPDX-License-Identifier: LGPL-2.1-or-later\n\n")
+
+        c.write(to_zsh_array("_mkosi_verbs", [str(v) for v in Verb]))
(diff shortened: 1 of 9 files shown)


bluca:
bash completion would be really nice, yes please

mkosi/config.py
```diff
@@ -3184,6 +3221,14 @@ def create_argument_only_parser(chdir: bool = True) -> argparse.ArgumentParser:
         help="The format to show documentation in",
         default=DocFormat.auto,
         type=DocFormat,
+        choices=list(DocFormat)
+    )
+    parser.add_argument(
+        "--shell-completion",
```
keszybz, line 3228: Hmm, `bin/mkosi --shell-completion` fails with `mkosi: error: argument --shell-completion: expected one argument`, and `bin/mkosi --shell-completion auto` starts a build. Oh, it's supposed to be used with `completion` verb. I think the verb is not needed, since we need the switch anyway. So I'd make `--shell-completion` standalone like `--help` or `--version`.

mkosi/resources/completion.bash
```diff
@@ -0,0 +1,45 @@
...
+        fi
+    else
+        # the preceding word wasn't an option, so we are doing position
+        # arguments now and all of them are verbs
+        COMPREPLY=( $(compgen -W "${_mkosi_verbs[*]}" -- "${completing_word}") )
+    fi
+}
+
+complete -o filenames -F _mkosi_completion mkosi
+complete -o filenames -F _mkosi_completion python -m mkosi
```
keszybz, line 45: Duplication here?

mkosi/config.py
```diff
@@ -4295,3 +4340,87 @@ def systemd_tool_version(config: Config, tool: PathString) -> GenericVersion:
...
+            f"declare -A {name.replace('-', '_')}=(" +
+            " ".join(f"[{shlex.quote(str(k))}]={shlex.quote(str(v))}" for k, v in entries.items()) + ")"
+        )
+
+    completion = resources / "completion.bash"
+
+    options_by_key = {o.short: o for o in options if o.short} | {o.long: o for o in options if o.long}
+
+    with io.StringIO() as c:
+        c.write(to_bash_array("_mkosi_options", options_by_key.keys()))
```
keszybz, line 4402: Hmm, why `io.StringIO`? Why not just use `print()` here?

mkosi/config.py
```diff
@@ -3184,6 +3221,14 @@ def create_argument_only_parser(chdir: bool = True) -> argparse.ArgumentParser:
         help="The format to show documentation in",
         default=DocFormat.auto,
         type=DocFormat,
+        choices=list(DocFormat)
+    )
+    parser.add_argument(
+        "--shell-completion",
+        help="The shell to print completion for",
```
keszybz, line 3229: "completion script" ?

keszybz (commented):
This is pretty nice. If only we could make `zsh` work too :)

Some issues I notice in quick testing:
- we only allow one verb, so if there's a verby-looking word already present, we shouldn't propose verbs.
- most verbs don't accept options after the verb. Thus, `mkosi build -` should not complete *our* options. Either nothing or a general command completion like for `sudo`.

behrmann:
> This is pretty nice. If only we could make zsh work too :)

I have a few long train rides upcoming this week and I've been reading the rather dense zsh completion documentation ;)

mkosi/resources/completion.bash
behrmann: No, that's actually here to complete both for the command `mkosi` and when calling mkosi via `python -m mkosi`. Not knowing fish I think that's still missing from the fish version.

mkosi/config.py
behrmann: Probably premature optimization. The idea is, that distro packages ship the already built completion script, but in case somebody really wants to generate it themselves, `io.StringIO()` is the fastest way to to generate a string. Also, this way the whole output gets built or not, it can't be generated and printed half. At least that was the idea.

keszybz: I guess not writing the partial string makes sense. But can it really fail?

At least use `print(…, file=out)` instead of `c.write()`.

behrmann:
Pushed an updated version, making this just an command line option instead of a verb, this also sidesteps the shell autodetection. This version adds zsh, but I've still not tested zsh nor fish, since I forgot to install them before boarding my train. 

I've kept the `.write` calls instead of `print`, since I find it nicer for the crufty strings this is building. 

I've not yet thought about not proposing verbs multiple times, because in the case of the bash completion this will necessitate changing the approach for the completion completely, since the whole argument string needs to be scanned and I think this can't use the "stateless" approach of looking at the current and preceding option. No idea how fish and zsh handle this, but generally all three need a few more completion generations.

I think the shell check things are mostly gunk, except maybe the thing about mapfile, this needs looking into.

Please, nobody ask for another shell, bash and zsh are scary enough, I do hope nobody is still using csh or ksh... :)

keszybz (commented):
I hammered the zsh part until is started to behave.

behrmann:
Another train ride another battle with shell completion. I tested the generated completions for bash, fish and zsh and at least they don't throw errors on me and complete the options I tried. The zsh version now has working choices, too, which didn't work for me during testing and the zsh version has the constant stuff factored out into a resource, like the bash version. Adding more completers might necessitate adding a resource for fish as well. 

There's still a lot of room for improvement (boolean options that can also take various things that we parse as booleans, options that take comma- or otherwise separated lists), but the basic stuff is there. 

The three completions are as similar as possible and as idiomatic as I was able to make them without being a user of fish or zsh (thanks @keszybz!). The bash version, while idiomatic and using what bash provides (it's O(1)!), will happily complete multiple verbs. That will need fixing, but that's a rewrite of the resource to parse the whole command line on every completion. I think I'd like to leave that for later. 

Please give it a try to see what breaks :)

DaanDeMeyer (commented):
Can't we pass the shell for which to generate completions as an argument and then make this a new verb `completion`? I don't like making this an option tbh.

DaanDeMeyer (commented):
if we made this a `verb`, we could also move all this logic into its own file `completion.py`

keszybz:
Ooh, nice, under zsh, completions for `-d` and `-t` work perfectly.
The same under bash.
So I think this is functionally complete.

I don't really have an opinion on verb vs. option choice. Previously, I asked for the verb to be removed because it was had both an option and the verb. But having the verb without an option is also fine.

In my local testing, I get this strange result:
```console
$ bin/mkosi --shell-completion=bash | less
...
declare -A _mkosi_compgen=([-C]=_mkosi_compgen_dirs [--directory]=_mkosi_compgen_dirs)

:[13]  + 2689831 suspended (tty input)  bin/mkosi --shell-completion=bash | 
        2689832 suspended (tty input)  less
```
It's stochastic, but happens ~80% of the time. I don't understand what is going on here, but I think it may be a bug in `less`. The output from `mkosi` seems to be just pure ASCII. Maybe zsh tries to do a read at exactly the wrong time and this causes … something?

behrmann:
I have no strong feeling regarding whether this is an option or a verb, I'll change it back to a verb.

behrmann:
Pushed a rework into a verb that also moves most of the stuff into a separate module. The `CompGen` enum, since it lives in `Args`, needs to stay in `config`, though, so as to not have a cyclic import.

mkosi/config.py
```diff
@@ -1167,6 +1212,7 @@ class ConfigSetting:
     path_secret: bool = False
     specifier: str = ""
     universal: bool = False
+    compgen: CompGen = CompGen.default
```
DaanDeMeyer, line 1215: Do we still need this in this struct?

mkosi/completion.py
```diff
@@ -0,0 +1,199 @@
...
+    completion = resources / "completion.zsh"
+
+    with io.StringIO() as c:
+        c.write("#compdef mkosi\n")
+        c.write("# SPDX-License-Identifier: LGPL-2.1-or-later\n\n")
+
+        c.write(to_zsh_array("_mkosi_verbs", [str(v) for v in Verb]))
+        c.write("\n\n")
+
+
```
DaanDeMeyer, line 145: Double newline

mkosi/completion.py
```diff
@@ -0,0 +1,199 @@
...
+            if option.choices:
+                c.write("-a \"")
+                c.write(" ".join(option.choices))
+                c.write("\" ")
+            if option.help is not None:
+                help = option.help.replace("'", "\\'")
+                c.write(f"-d \"{help}\" ")
+            c.write(option.compgen.to_fish())
+            c.write("\n")
+        return c.getvalue()
```
DaanDeMeyer, line 129: Missing newline

mkosi/completion.py
```diff
@@ -0,0 +1,199 @@
...
+
+def print_completion(args: Args, *, resources: Path) -> None:
+    if not args.cmdline:
+        die(
+            "No shell to generate completion script for specified",
+            hint="Please specify either one of: bash, fish, zsh"
+        )
+
+    shell = args.cmdline[0]
+    completion_args = collect_completion_arguments()
```
DaanDeMeyer, line 188: I would just inline this variable

DaanDeMeyer (changes requested):
Can we fix the shellchecks as well? I didn't review the generation stuff too closely but it all looks reasonable, the comments are mostly about decoupling this a little more so it's self contained.

Still need to test the fish completion

mkosi/resources/bash.completion
```diff
@@ -0,0 +1,44 @@
...
+    compgen -d -- "$1"
+}
+
+_mkosi_completion() {
+    # completing_program="$1"
+    local completing_word="$2"
+    local completing_word_preceding="$3"
+
+    if [[ "$completing_word" =~ ^- ]]; then  # completing an option
+        readarray -t COMPREPLY < <(compgen -W "${_mkosi_options[*]}" -- "${completing_word}")
```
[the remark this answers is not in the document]
keszybz: I thought that renaming the file would make shellcheck ignore it. But this clearly doesn't work. I'll drop the rename.

keszybz:
I force-pushed with the changes that were requested. I want to get this merged… I thought it'd be a few small cleanups, but then it turns out to be more work. I hope that's OK.

behrmann:
> I thought it'd be a few small cleanups, but then it turns out to be more work. I hope that's OK.

Thanks! That's why I didn't get around to it this week, a few deadlines to keep.

> Also, adjust the formatting in the bash resource to follow the usual style
with 'if something; then' on one line.

I gave you the style in zsh, give me bash. I strongly detest semicolons in shell scripts and the two line style is the way ALGOL intended it. ;) 

Joke aside, the style with semicolon is more common, but I do think the two-line style is more readable, makes visually nicer blocks and looks nicer when the predicate or what one wants to loop over takes more than one line, which happens easily.

I'm not married to it, though, and the further rework to make the different `CompGen` usable can happen later, that is why the annotation for them was in `config`. Daan and I spoke about trying to look up the types on the `Config` object, but we don't need to wait for this, the current form should be good enough as a start and stylistic quibbles can be changed later.

keszybz:
I undid the style change to `if`s.

keszybz:
OK, the tests pass and the comments have been addressed. Can we merge this and do follow-ups separately?

DaanDeMeyer:
shellcheck is still failing. I'm fine with disabling the checks but that should be fixed before merging.

DaanDeMeyer (approved):
LGTM but CI should be green

keszybz:
It is green, except for bogus comments from shellcheck that cannot be silenced.

keszybz:
https://github.com/koalaman/shellcheck/issues/2411

DaanDeMeyer:
@keszybz The differential-shellcheck has an `exclude-path` option. We can use that.

behrmann:
I added a commit with a forward definition of the (hash) arrays, that should hopefully appease the differential shell check

keszybz:
> The differential-shellcheck github action has an `exclude-path` option. We can use that.

I looked at docs, but I didn't see it. That would work.

> I added a commit with a forward definition of the (hash) arrays, that should hopefully appease the differential shell check

That also works…

PR #2784, opened 2024-06-13
centos: Fix GPG key search logic

DaanDeMeyer:
Let's only look for the release key of the release we're building for and add support for searching for the SHA256 key which is used by centos stream 10.

We can't use the symlinked names because those don't exist on centos.org/keys.


behrmann (approved):
The remaining CentOS error is due to our regex [not matching the stub version because it is a lot of null bytes](https://github.com/systemd/mkosi/actions/runs/9506835802/job/26205006217?pr=2784#step:7:7405)

Issue #2786, opened 2024-06-14
How to run mkosi inside chroot


asrinuva:
### mkosi commit the issue has been seen with

main

### Used host distribution

debian 13

### Used target distribution

chroot

### Linux kernel version used

 6.5.0-35-generic #35~22.04.1-Ubuntu

### CPU architectures issue was seen on

x86_64

### Unexpected behaviour you saw

Hi everyone,

                   My intention is to create standalone initrd via yocto recipe, created the recipe to chroot into the rootfs and installed the necessary packages and the calling mkosi to generate image/initrd

In the first time it says that "KeyError: 'getpwuid(): uid not found: 1000" this one gone after adding user with UID 1000 in chroot.
then i invoke the mkosi with the minimal configuration but still it says the below error:
No configuration found
‣ (Make sure you're running mkosi from a directory with configuration files)
‣ + tput cnorm
‣ + tput smam
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/mkosi/run.py", line 55, in uncaught_exception_handler
    yield
  File "/usr/lib/python3.11/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/mkosi/__main__.py", line 38, in main
    run_verb(args, images, resources=resources)
  File "/usr/lib/python3/dist-packages/mkosi/__init__.py", line 4098, in run_verb
    die("No configuration found",
  File "/usr/lib/python3/dist-packages/mkosi/log.py", line 31, in die
    sys.exit(1)
SystemExit: 1

Any help would be appreciated

### Used mkosi config

```ini
mkosi.conf file used in root directory:

[Output]
Format=cpio

[Content]
Packages=systemd
         udev
         kmod
```

### mkosi output

```sh
root@srinu-Precision-Tower-3430:/# ls
bin   dev  home        initrd.img.old  lib64       media       mnt  proc  run   srv  tmp  var      vmlinuz.old
boot  etc  initrd.img  lib             lost+found  mkosi.conf  opt  root  sbin  sys  usr  vmlinuz
root@srinu-Precision-Tower-3430:/# mkosi --debug
‣ Including configuration file /mkosi.conf
‣ Including configuration file /mkosi.conf
‣ No configuration found
‣ (Make sure you're running mkosi from a directory with configuration files)
‣ + tput cnorm
‣ + tput smam
Traceback (most recent call last):
[...]

PR #2787, opened 2024-06-15
chdir into /proc for determining default Config

DaanDeMeyer:
There are users that run mkosi from / so let's use /proc instead.

Fixes https://github.com/systemd/mkosi/issues/2786


behrmann (approved):

asrinuva:
Or shall i apply this patch on top of debian trixie mkosi to get this support?

DaanDeMeyer:
@asrinuva You don't need this patch if you just use any other directory than / to put the mkosi config files and do the build.

asrinuva:
Thanks! @DaanDeMeyer, Let me check.

PR #2789, opened 2024-06-16
Use dataclasses instead of NamedTuples, unnest code

keszybz:


mkosi/installer/apt.py
```diff
@@ -16,26 +18,27 @@
...
+            f"""\
+            Types: {" ".join(self.types)}
+            URIs: {self.url}
+            Suites: {self.suite}
+            Components: {" ".join(self.components)}
+            {"Signed-By" if self.signedby else "Trusted"}: {self.signedby or "yes"}
 
+            """
+        )
+
```
behrmann, line 40:
```suggestion
        )


```

mkosi/installer/pacman.py
```diff
@@ -16,12 +17,12 @@
 from mkosi.util import umask
 from mkosi.versioncomp import GenericVersion
 
+@dataclasses.dataclass
+class PacmanRepository:
+    id: str
+    url: str
```
behrmann, line 23:
```suggestion
    url: str

```

tests/__init__.py
```diff
@@ -19,15 +20,16 @@
...
 
 
+@dataclasses.dataclass
+class ImageConfig:
+    distribution: Distribution
+    release: str
+    tools_tree_distribution: Optional[Distribution]
+    tools_tree_release: Optional[str]
+    debug_shell: bool
+
```
behrmann, line 30:
```suggestion
    debug_shell: bool


```

behrmann (commented):
Looks reasonable to me, a few style nits and the imports need some adjustement.

I quite liked the approach of having the repositories be something like an "associated type" of the package mager, similiar to how on would have the type returned by iterator trait in Rust, but I don't think it makes all that much of a difference in terms of typing in Python.

keszybz:
Updated with the suggested changes and one more cleanup.

> I quite liked the approach of having the repositories be something like an "associated type" of the package mager

Yeah, but then the nested class is in the parent class namespace, which can get weird. It probably doesn't matter for those simple cases, but it's not a good pattern in general.

mkosi/installer/pacman.py
behrmann: This one is still missing the empty line.

keszybz:
Locally, the fedora integration tests fail with
```
  - nothing provides libperl.so.5.38()(64bit) needed by perf-6.10.0-0.rc3.32.fc41.x86_64
```

I guess it's it's the same in CI.

mkosi/installer/pacman.py
keszybz: I think it's good. I see two empty lines above and below `PacmanRepository`.

keszybz:
`perf` is from `kernel.src.rpm` and it was rebuilt and the latest version requires `libperl.so.5.40`. But I think updates to rawhide are stuck, I see some other packages not being tagged after a rebuild.

mkosi/installer/pacman.py
behrmann: You're right, somehow it's stuck for me on GitHub, but in the checkout it's fine.

behrmann (approved):

behrmann:
> Locally, the fedora integration tests fail with
> 
> ```
>   - nothing provides libperl.so.5.38()(64bit) needed by perf-6.10.0-0.rc3.32.fc41.x86_64
> ```
> 
> I guess it's it's the same in CI.

Yes, it's the same there as well, I guess we'll have to wait for that to be fixed, though it doesn't need to delay this PR.

keszybz:
There's some chatter in releng matrix channel that signing is stuck. We might have to wait until Monday, US West timezone.

septatrix (commented):
I do not agree with the statement that namedtuples are an outdated solution. They are still frequently used and are ideal for immutable record (which is basically all they serve in mkosi). We do not utilize the unwrapping they offer but neither do we use any of the automatically implemented stuff of dataclasses. If we want to treat them more like classes we should also consider moving methods into them (e.g. `we_are_wanted`)

The nested classes for the repos had the advantage that could in theory be defined in the abstract parent classes and thus mandated/checked my mypy but in practice we did not use that and treatment between installers/distros is not consistent so I have no strong opinion about that change.

+1 for unnesting the image config

DaanDeMeyer (approved):
I don't necessarily think dataclasses are better than named tuples either but consistency is a good thing so this LGTM.

PR #2797, opened 2024-06-22
SHA256SUMS fixes

NekkoDroid:
1. Move file generating after `mkosi.postoutput`
2. Skip directories in staging (user may have created some)
3. Remove check for directory output type (user may have added non-dir output that should be hashed)

I am a bit uncertain if I should check if there are any files to hash to skip generating the file but if the user did request creating the file they might unconditionally rely on it existing.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2645,12 +2645,13 @@ def calculate_sha256sum(context: Context) -> None:
     if not context.config.checksum:
         return
 
-    if context.config.output_format == OutputFormat.directory:
-        return
-
     with complete_step("Calculating SHA256SUMS…"):
         with open(context.workspace / context.config.output_checksum, "w") as f:
             for p in context.staging.iterdir():
+                if p.is_dir():
+                    logging.warning(f"Cannot checksum directory '{p}', skipping")
+                    continue
+
                 print(hash_file(p) + " *" + p.name, file=f)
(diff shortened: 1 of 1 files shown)


septatrix:
Intuitively I would expect to be able to access the sha sums in the postoutput script to e.g. upload them and the image to an s3 bucket for sysupdate.

DaanDeMeyer:
> Intuitively I would expect to be able to access the sha sums in the postoutput script to e.g. upload them and the image to an s3 bucket for sysupdate.

Indeed, I don't think we can just change this like this. The second commit looks good though

NekkoDroid:
> Intuitively I would expect to be able to access the sha sums in the postoutput script to e.g. upload them and the image to an s3 bucket for sysupdate.

I guess that does make sense. My thought process was more that since the files can still be created, removed and modified at that point it does make sense yet to create the file.

DaanDeMeyer:
> For uploading, wouldn't it make more sense to do that after everything has been moved to `OutputDirectory=`? or am I missing something

They're might be an appeal to having mkosi do the upload so one just has to run mkosi and not have to wrap it with something else. I used to recommend a wrapper script but now that we have postoutput scripts it's possible to have mkosi do it itself via a postoutput script.

DaanDeMeyer (approved):

PR #2805, opened 2024-06-26
[RFC] mkosi-initrd: provide a separate mkosi-initrd script

aafeijoo-suse:
Extract the basic functionality implemented on the kernel-install plugin into a separate script. This allows to simplify the initrd/UKI build auto-filling arguments of the `mkosi` call, and also auto-include user/distro configuration from `/{etc,usr/lib}/mkosi-initrd`.

### DIFF
--- a/kernel-install/50-mkosi.install
+++ b/kernel-install/50-mkosi.install
@@ -13,7 +13,7 @@ from mkosi import identify_cpu
 from mkosi.archive import make_cpio
 from mkosi.config import OutputFormat, __version__
 from mkosi.log import die, log_setup
-from mkosi.run import find_binary, run, uncaught_exception_handler
+from mkosi.run import run, uncaught_exception_handler
 from mkosi.types import PathString
 from mkosi.util import umask
 
@@ -79,30 +79,42 @@ def main() -> None:
         usage='50-mkosi.install COMMAND KERNEL_VERSION ENTRY_DIR KERNEL_IMAGE INITRD…',
     )
 
-    parser.add_argument("command",
-                        metavar="COMMAND",
-                        help="The action to perform. Only 'add' is supported.")
-    parser.add_argument("kernel_version",
-                        metavar="KERNEL_VERSION",
-                        help="Kernel version string")
-    parser.add_argument("entry_dir",
-                        metavar="ENTRY_DIR",
-                        type=Path,
-                        nargs="?",
-                        help="Type#1 entry directory (ignored)")
-    parser.add_argument("kernel_image",
-                        metavar="KERNEL_IMAGE",
-                        type=Path,
-                        nargs="?",
-                        help="Kernel image")
-    parser.add_argument("initrds",
-                        metavar="INITRD…",
-                        type=Path,
-                        nargs="*",
-                        help="Initrd files")
-    parser.add_argument("--version",
-                        action="version",
-                        version=f"mkosi {__version__}")
+    parser.add_argument(
+        "command",
+        metavar="COMMAND",
+        help="The action to perform. Only 'add' is supported.",
+    )
+    parser.add_argument(
+        "kernel_version",
+        metavar="KERNEL_VERSION",
+        help="Kernel version string",
+    )
+    parser.add_argument(
+        "entry_dir",
+        metavar="ENTRY_DIR",
+        type=Path,
+        nargs="?",
+        help="Type#1 entry directory (ignored)",
+    )
+    parser.add_argument(
+        "kernel_image",
+        metavar="KERNEL_IMAGE",
+        type=Path,
+        nargs="?",
+        help="Kernel image",
+    )
+    parser.add_argument(
+        "initrds",
+        metavar="INITRD…",
+        type=Path,
+        nargs="*",
+        help="Initrd files",
+    )
+    parser.add_argument(
+        "--version",
+        action="version",
+        version=f"mkosi {__version__}",
+    )
 
     context = Context(
         **vars(parser.parse_args()),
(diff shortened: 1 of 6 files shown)


mkosi/initrd/__main__.py
```diff
@@ -0,0 +1,102 @@
...
+    )
+
+    parser.add_argument("--kernel-version",
+                        metavar="KERNEL_VERSION",
+                        help="Kernel version string",
+                        default=platform.uname().release)
+    parser.add_argument("--format",
+                        choices=["cpio", "uki"],
+                        help="Output format (CPIO archive or UKI)",
+                        default="cpio")
```
DaanDeMeyer, line 34: Let's allow `directory` as well as that's very useful for inspecting the initrd.

mkosi/initrd/__main__.py
```diff
@@ -0,0 +1,102 @@
...
+    log_setup()
+
+    parser = argparse.ArgumentParser(
+        prog="mkosi-initrd",
+        description="Build initrds or Unified Kernel Images using mkosi",
+        allow_abbrev=False,
+        usage="mkosi-initrd [options...]",
+    )
+
+    parser.add_argument("--kernel-version",
```
DaanDeMeyer, line 27: Can you do the same formatting here and below as you did for `ArgumentParser` above?

mkosi/initrd/__main__.py
```diff
@@ -0,0 +1,102 @@
...
+                        default=platform.uname().release)
+    parser.add_argument("--format",
+                        choices=["cpio", "uki"],
+                        help="Output format (CPIO archive or UKI)",
+                        default="cpio")
+    parser.add_argument("--version",
+                        action="version",
+                        version=f"mkosi {__version__}")
+
+    args_initrd, args_other = parser.parse_known_args()
```
DaanDeMeyer, line 39: I would name these`args` and `extra`. The current names are a bit verbose

mkosi/initrd/__main__.py
```diff
@@ -0,0 +1,102 @@
...
+
+    cmdline: list[PathString] = [
+        "mkosi",
+        "--directory", "",
+        "--format", str(args_initrd.format),
+        "--workspace-dir=/var/tmp",
+        "--package-cache-dir=/var",
```
DaanDeMeyer, line 46: So doing this unconditionally means running `mkosi-initrd` as non-root is not going to work. It also means that if users pass `-fff` as the extra arguments we'll start removing stuff in /var which could be potentially dangerous.

So we should probably only do this when we're invoked as root and we should probably not allow passing arbitrary extra arguments to the script either..This means we should allow passing `--output` and `--output-dir` to this script but nothing else (unless there's an explicit use case).

mkosi/initrd/__main__.py
```diff
@@ -0,0 +1,102 @@
...
+            if (Path("/etc") / p).resolve().is_file():
+                shutil.copy2(Path("/etc") / p, Path(d) / "etc" / p)
+            else:
+                shutil.copytree(Path("/etc") / p, Path(d) / "etc" / p,
+                                ignore=shutil.ignore_patterns("S.*"), dirs_exist_ok=True)
+
+        cmdline += ["--package-manager-tree", d]
+
+        if args_other is not None:
+            cmdline += args_other
```
DaanDeMeyer, line 93: We should probably drop this, see above

DaanDeMeyer (changes requested):
Looks great but some comments

mkosi/initrd/__main__.py
```diff
@@ -0,0 +1,102 @@
...
+
+    for d in ("/usr/lib/mkosi-initrd", "/etc/mkosi-initrd"):
+        if Path(d).exists():
+            cmdline += ["--include", d]
+
+    with tempfile.TemporaryDirectory() as d:
+        # Make sure we don't use any of mkosi's default repositories.
+        for p in (
+            "yum.repos.d/mkosi.repo",
+            "apt/sources.list",
```
DaanDeMeyer, line 64: I updated this in 70aa901697f12182ccaa24e2325867d275479b55 which causes a merge conflict so this will need to be updated to match that commit

mkosi/initrd/__main__.py
DaanDeMeyer: @aafeijoo-suse We definitely should not allow specifying arbitrary arguments. Some of them might make sense, but others definitely don't.

For example, specifying `--distribution` or `--release` is a very bad idea, because we disable usage of mkosi's builtin repositories and always use the ones from the host. So using anything other than the host's distribution and release in this case is a very bad idea.

This also makes me realize that we should very clearly document this tool is for building an initrd for the current host system and that for building generic initrds you need to invoke mkosi directly.

aafeijoo-suse: ok, let's polish up the maybe-necessary missing arguments in later steps.

aafeijoo-suse:
> @aafeijoo-suse We definitely should not allow specifying arbitrary arguments. Some of them might make sense, but others definitely don't.

I'm realizing that `--force` would also be good.

kernel-install/50-mkosi.install
```diff
@@ -129,65 +141,19 @@ def main() -> None:
...
         "--output-dir", context.staging_area,
-        "--extra-tree", f"/usr/lib/modules/{context.kernel_version}:/usr/lib/modules/{context.kernel_version}",
-        "--extra-tree=/usr/lib/firmware:/usr/lib/firmware",
-        "--kernel-modules-exclude=.*",
-        "--kernel-modules-include=host",
-        "--include=mkosi-initrd",
     ]
 
     if context.verbose:
         cmdline += ["--debug"]
```
DaanDeMeyer, line 152: This is still missing from mkosi-initrd which means it'll fail when running kernel-install in verbose mode. We should add a --debug argument to mkosi-initrd as well.

kernel-install/50-mkosi.install
```diff
@@ -79,30 +79,42 @@ def main() -> None:
...
-                        type=Path,
-                        nargs="*",
-                        help="Initrd files")
-    parser.add_argument("--version",
-                        action="version",
-                        version=f"mkosi {__version__}")
+    parser.add_argument(
+        "command",
+        metavar="COMMAND",
+        help="The action to perform. Only 'add' is supported."
```
DaanDeMeyer, line 85: Can we add trailing commas here and everywhere else?

mkosi/initrd/__main__.py
```diff
@@ -0,0 +1,123 @@
...
+
+    parser.add_argument(
+        "--kernel-version",
+        metavar="KERNEL_VERSION",
+        help="Kernel version string",
+        default=platform.uname().release
+    )
+    parser.add_argument(
+        "-t", "--format",
+        choices=["cpio", "uki", "directory"],
```
DaanDeMeyer, line 35: Can you use `OutputFormat.cpio` and such here instead of the raw strings? That makes everything a bit more type safe.

mkosi/initrd/__main__.py
```diff
@@ -0,0 +1,123 @@
...
+from mkosi.types import PathString
+
+
+@uncaught_exception_handler()
+def main() -> None:
+    log_setup()
+
+    parser = argparse.ArgumentParser(
+        prog="mkosi-initrd",
+        description="Build initrds or Unified Kernel Images using mkosi",
```
DaanDeMeyer, line 22: "Build initrds or Unified Kernel Images for the current system using mkosi"? I want to explicitly say these are for the current system and not generic.

DaanDeMeyer (changes requested):
Can you also add a `resources/` folder in `mkosi/initrd` with `mkosi-initrd.md` and write a man page for this in there?

aafeijoo-suse:
@DaanDeMeyer I think I addressed all your requests, PTAL. Thanks!

mkosi/initrd/resources/mkosi-initrd.md
```diff
@@ -0,0 +1,59 @@
...
+# OPTIONS
+
+`--kernel-version=`
+:   Kernel version where to look for the kernel modules to include. Defaults to
+    the kernel version of the running system (`uname -r`).
+
+`--format=`, `-t`
+:   Output format. One of `cpio` (CPIO archive), `uki` (a unified kernel image
+    with the image in the `.initrd` PE section) or `directory` (for generating
+    an image directly in a local directory).
```
DaanDeMeyer, line 28: Should mention here that we default to `cpio`

mkosi/initrd/resources/mkosi-initrd.md
```diff
@@ -0,0 +1,59 @@
...
+    to `initrd`.
+
+    Note that this only specifies the output prefix, depending on the
+    specific output format and compression used, the full output name might be
+    `initrd.cpio.zst`
+
+`--output-dir=`, `-O`
+:   Path to a directory where to place all generated artifacts. If this is
+    not specified and the directory `mkosi.output/` exists in the local
+    directory, it is automatically used for this purpose.
```
DaanDeMeyer, line 41: This should mention we use the current working directory if this is not specified and `mkosi.output` does not exist.

mkosi/initrd/resources/mkosi-initrd.md
```diff
@@ -0,0 +1,59 @@
...
+    with the image in the `.initrd` PE section) or `directory` (for generating
+    an image directly in a local directory).
+
+`--output=`, `-o`
+:   Name to use for the generated output image file or directory. Defaults
+    to `initrd`.
+
+    Note that this only specifies the output prefix, depending on the
+    specific output format and compression used, the full output name might be
+    `initrd.cpio.zst`
```
DaanDeMeyer, line 36: Missing .

mkosi/initrd/__main__.py
```diff
@@ -0,0 +1,141 @@
...
+        metavar="DIR",
+        help="Output directory",
+        default="",
+    )
+    parser.add_argument(
+        "-f", "--force",
+        help="Remove existing image file before operation",
+        action="count",
+        default=0,
+    )
```
DaanDeMeyer, line 56: I'm not entirely sure we should allow configuring this just yet. Maybe just always add a single `--force` instead? I'm a bit scared about making this configurable since 3x `--force` would start nuking caches in /var when running as root.

DaanDeMeyer (changes requested):
Just a few more nits and we're good to go

mkosi/initrd/__main__.py
aafeijoo-suse: It sounds good enough for me.

mkosi/initrd/resources/mkosi-initrd.md
aafeijoo-suse: In fact using `mkosi.output` was not working, fixed.

mkosi/initrd/__main__.py
```diff
@@ -0,0 +1,134 @@
...
+        default=False,
+    )
+    parser.add_argument(
+        "--version",
+        action="version",
+        version=f"mkosi {__version__}",
+    )
+
+    args = parser.parse_args()
+    output_dir = "mkosi.output" if args.output_dir == "" and Path("mkosi.output").exists() else args.output_dir
```
DaanDeMeyer, line 64: Ah right because of `--directory ""` we ignore any local files. This makes me think that maybe it's simpler to not support automatic usage of `mkosi.output` and just always use the current working directory unless one is specified explicitly.

aafeijoo-suse: Sure, changed.

DaanDeMeyer (approved):

PR #2814, opened 2024-07-01
Support decompressing tar archives (+ use systemd-dissect long option)

NekkoDroid:
This does switch back to reading the input file directly from the mount instead of stdin as GNU tar doesn't support decompresssion from stdin.

As a sidenote: I noticed that before `src` was still mounted into the sandbox even though it was read from stdin, any specific reason for that?

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -108,8 +108,8 @@ def mount_base_trees(context: Context) -> Iterator[None]:
                 extract_tar(path, d, sandbox=context.sandbox)
                 bases += [d]
             elif path.suffix == ".raw":
-                run(["systemd-dissect", "-M", path, d])
-                stack.callback(lambda: run(["systemd-dissect", "-U", d]))
+                run(["systemd-dissect", "--mount", "--mkdir", path, d])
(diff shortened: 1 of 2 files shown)


mkosi/__init__.py
```diff
@@ -108,8 +108,8 @@ def mount_base_trees(context: Context) -> Iterator[None]:
                 extract_tar(path, d, sandbox=context.sandbox)
                 bases += [d]
             elif path.suffix == ".raw":
-                run(["systemd-dissect", "-M", path, d])
-                stack.callback(lambda: run(["systemd-dissect", "-U", d]))
```
DaanDeMeyer, line 112: These are not the same as `-M` and `-U`, please see the manpage.

NekkoDroid: Should I then just drop the commit or add `--mkdir` and `--rmdir`?

DaanDeMeyer (approved):

mkosi/__init__.py
DaanDeMeyer: Add --mkdir and --rmdir

PR #2820, opened 2024-07-02
Pick up GNUPGHOME from user's environment

### LINKED ISSUE #2815: Honor GNUPGHOME when signing checksum

main

### Used host distribution

Arch Linux

### Used target distribution

Arch Linux

### Linux kernel version used

6.9.7-arch1-1

x86_64

Hi :wave:

I’d like to sign checksums but mkosi does not honor my `GNUPGHOME` [environment variable](https://www.gnupg.org/documentation/manuals/gnupg/GPG-Configuration.html). I suggest either honor this env var or provide a way to choose an alternative gnupg directory?

Maybe such option already exists, but I didn’t see anything related in [the manual](https://github.com/systemd/mkosi/blob/main/mkosi/resources/mkosi.md).

### Used mkosi config

_No response_

### mkosi output

```sh
‣  Calculating SHA256SUMS…
‣ GPG home /home/me/.gnupg not found
```

DaanDeMeyer:
If not set in Environment=, we default to the user's home directory. Let's instead use GNUPGHOME from the user's environment if it is set before falling back to the user's home directory.

Fixes #2815.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -3830,6 +3830,8 @@ def load_environment(args: argparse.Namespace) -> dict[str, str]:
         env["GIT_PROXY_SSL_KEY"] = "/proxy.clientkey"
     if dnf := os.getenv("MKOSI_DNF"):
         env["MKOSI_DNF"] = dnf
+    if gnupghome := os.getenv("GNUPGHOME"):
+        env["GNUPGHOME"] = gnupghome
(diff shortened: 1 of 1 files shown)


bluca (approved):

DaanDeMeyer:
Not covered by CI

PR #2822, opened 2024-07-02
Add mirror to package cache key if one is specified

DaanDeMeyer:
On Arch Linux, pacman seems to only check if the sync db on the mirror is newer than the one available locally, which breaks when using e.g. the archive where the sync db is older by definition.

Let's add any specified mirror to the default package cache dir key to make sure that a different cache is used for explicitly specified mirrors.


bluca (approved):

greyltc:
I've tested this and it's working as expected.

PR #2825, opened 2024-07-02
Always rerun build if Format=none and don't remove previous outputs

DaanDeMeyer:
Let's always rerun the build if Format=none. Also, since we know Format=none won't touch any of the outputs we know about, let's keep the existing outputs intact. This allows using Format=none to rerun the build script without removing the existing output (e.g. directory or disk image).

We'll be able to make use of this in mkosi-kernel to rebuild the kernel modules without removing the disk image produced in a previous step.

We also simplify check_outputs() to only check the main output and not the auxiliary outputs.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2823,14 +2823,13 @@ def check_inputs(config: Config) -> None:
 
 
 def check_outputs(config: Config) -> None:
-    for f in (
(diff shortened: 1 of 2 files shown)


behrmann (approved):
LGTM, but should be mentioned in the man page.

PR #2828, opened 2024-07-03
add Arch Linux ARM repo

greyltc:
Arch Linux ARM needs its alarm repo

### DIFF
--- a/mkosi/distributions/arch.py
+++ b/mkosi/distributions/arch.py
@@ -84,6 +84,9 @@ class Installer(DistributionInstaller):
                 ) if repo in context.config.repositories
             ] + ["core", "extra"]
 
+            if context.config.architecture.is_arm_variant():
+                repos += ["alarm"]
+
             for repo in repos:
                 yield PacmanRepository(repo, url)
 


mkosi/distributions/arch.py
```diff
@@ -84,6 +84,10 @@ def repositories(cls, context: Context) -> Iterable[PacmanRepository]:
                 ) if repo in context.config.repositories
             ] + ["core", "extra"]
 
+            # the alarm repo needs to be at the top for Arch Linux ARM
+            if context.config.architecture.is_arm_variant():
+                repos.insert(0, "alarm")
```
DaanDeMeyer, line 89: Please just initialize the `repos` list with `alarm` if needed instead of inserting to the front at the end. Also, can you post the contents of the default pacman.conf on alarm? I'd like to know if there are more repos we might need to take into account.

greyltc: Here's alarm's pacman.conf: https://github.com/archlinuxarm/PKGBUILDs/blob/master/core/pacman/pacman.conf

DaanDeMeyer: Then adding it to the end is the way to go

DaanDeMeyer (changes requested):
This seems overly complicated, I think we can just add `alarm` at the end. I'd like to know the purpose of the `aur` repo? If it's AUR packages then I don't think we should add it.

mkosi/distributions/arch.py
```diff
@@ -84,6 +84,9 @@ def repositories(cls, context: Context) -> Iterable[PacmanRepository]:
                 ) if repo in context.config.repositories
             ] + ["core", "extra"]
 
+            if context.config.architecture.is_arm_variant():
+                repos.append("alarm")
```
DaanDeMeyer, line 88: We prefer `+= ["alarm"]` over `append` for consistency

DaanDeMeyer (changes requested):
Almost there!

DaanDeMeyer (approved):

PR #2830, opened 2024-07-03
use virtio console in x86

teknoraver:
Use the VirtIO console device instead of the emulated one on x86. The console output is ~16x faster, and it's very noticeable with scrolling output or interactive tools like `htop`

Emulated device:
```
time hexdump -vC -n200000 /dev/zero

real	0m6.756s
user	0m0.035s
sys	0m6.714s
```

VirtIO console
```
time hexdump -vC -n200000 /dev/zero

real	0m0.421s
user	0m0.040s
sys	0m0.376s
```

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -429,6 +429,8 @@ class Architecture(StrEnum):
             Architecture.ppc      : "hvc0",
             Architecture.ppc64    : "hvc0",
             Architecture.ppc64_le : "hvc0",
+            Architecture.x86      : "hvc0",
+            Architecture.x86_64   : "hvc0",
         }.get(self, "ttyS0")
 
     def supports_smbios(self, firmware: QemuFirmware) -> bool:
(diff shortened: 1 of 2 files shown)


DaanDeMeyer (commented):
So we used to do this but I got rid of it again because edk2 didn't have a virtio serial driver and a bunch of distributions didn't have a crucial fix in systemd that makes this nice to use. I'll test if those are gone now and then we can switch. Otherwise I prefer to stick with a regular serial console.

DaanDeMeyer:
Yeah so unfortunately on Fedora 40's edk2 I don't get any output from systemd-boot when using virtconsole. @kraxel I remember you were working on a virtio-serial driver for edk2, did that end up landing?

teknoraver:
I see, you disabled it in 2893ca9777e6a209c14a01957fc482d21aea0e44.
Is there some chance to make it configurable from the config file?
In this way users which aren't using EFI could benefit from the faster console.

DaanDeMeyer:
We can add a new setting `QemuSerial=` or so with values `isa` and `virtio` to allow choosing between the two.

DaanDeMeyer:
@kraxel The problem was on my end, it indeed works great, thanks for working on this!

PR #2833, opened 2024-07-04
Add support for Azure Linux

DaanDeMeyer:
Azure Linux looks a lot like Fedora Linux so we opt to share configuration
between Azure and Fedora/CentOS and inherit the Azure definition from
Fedora.

This should not be merged until Azure Linux 3.0 moves out of preview since the mirror urls will have to change when that happens.


mkosi.conf.d/30-azure-centos-fedora/mkosi.conf.d/20-x86-64.conf
```diff
@@ -7,5 +7,5 @@ Architecture=x86-64
 Packages=
         grub2-efi-x64-modules
         grub2-pc
-        kernel-uki-virt
+        # kernel-uki-virt
```
bluca, line 10: why is this commented out?

DaanDeMeyer: oops leftover

behrmann (approved):
Would it be possible to add it to CI in some way?

keszybz (approved):
LGTM too.

bluca:
The kconfigs we need have been added: https://github.com/microsoft/azurelinux/pull/9971
not sure when the next RC will happen, but these should be included

bluca (approved):

PR #2834, opened 2024-07-04
Use qemu's virtconsole for the serial console

DaanDeMeyer:
edk2 now has a virtio serial driver so let's switch to virtconsole
for the serial console as it's significantly faster compared to the
old ISA serial console.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,12 @@
 # mkosi Changelog
 
+## v24
+
+- The default kernel command line of `console=ttyS0` (or equivalent for
+  other architectures) has been removed. The required `console=`
+  argument to have the kernel output to the serial console has to be
+  added manually from `v24` onwards.
+
 ## v23.1
 
 - Respin due to git tag mismatch
--- a/mkosi.conf
+++ b/mkosi.conf
@@ -34,7 +34,7 @@ RemoveFiles=
         /usr/lib/kernel/install.d/50-dracut.install
 
 # Make sure that SELinux doesn't run in enforcing mode even if it's pulled in as a dependency.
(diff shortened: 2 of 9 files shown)


teknoraver:
I confirm it works:
```
[root@fedora ~]# w
 17:53:36 up 0 min,  3 users,  load average: 0.18, 0.06, 0.02
USER     TTY        LOGIN@   IDLE   JCPU   PCPU WHAT
root     hvc0      17:53    0.00s  0.03s  0.01s w
```

teknoraver:
And it's fast:
```
[root@fedora ~]# time hexdump -vC -n200000 /dev/zero

real	0m0.451s
user	0m0.029s
sys	0m0.418s
```

behrmann (approved):
Beautiful.

rpardini:
While this does indeed work, for images targeting `ttyS0` or `ttyAMA0` as (real/physical) consoles, this breaks the autologin feature on those consoles (`tty1` still works). I guess at a certain point (eg, Rockchip devices and their console on `ttyS2`) it makes more sense to make autologin consoles configurable somehow. Should I open an issue about this?

DaanDeMeyer:
> While this does indeed work, for images targeting `ttyS0` or `ttyAMA0` as (real/physical) consoles, this breaks the autologin feature on those consoles (`tty1` still works). I guess at a certain point (eg, Rockchip devices and their console on `ttyS2`) it makes more sense to make autologin consoles configurable somehow. Should I open an issue about this?

You can configure autologin with systemd credentials these days on a per tty basis. I would always do that instead of the autologin setting. I'll probably remove it at some point.

PR #2835, opened 2024-07-04
Replace root=PARTUUID and mount.usr=PARTUUID with the actual partition UUID

DaanDeMeyer:
These are not known upfront so let's treat these literals as placeholders which are to be replaced with the actual partition UUID when we know it.


greyltc:
this worked for me! :-)

behrmann (approved):

PR #2838, opened 2024-07-05
ci: Enable noble-proposed

DaanDeMeyer:
See https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/2067907


behrmann (approved):

DaanDeMeyer:
Update is not in proposed mirrors yet so we'll need to wait

PR #2839, opened 2024-07-05
Use mkosi dependencies in default image

DaanDeMeyer:
Let's reduce duplication by using mkosi dependencies to install
all required dependencies to build the default image. To make this
work we add support for tools trees on jammy/bookworm. We also drop
support for a focal default image as it's too much of a burden to
maintain.

We also remove the EPEL only packages from the CentOS tools tree so
that mkosi dependencies works on a CentOS Stream system without EPEL
enabled.

Finally, curl-minimal is replaced by curl because trying to install
curl-minimal on a system with curl leads to a dnf failure so we opt
for the larger curl in order to make mkosi dependencies useful in
more situations.


mkosi/resources/mkosi-tools/mkosi.conf.d/10-debian-ubuntu/mkosi.conf.d/ubuntu-keyring.conf
```diff
@@ -0,0 +1,9 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+[Match]
+Distribution=|!debian
+Release=|!bookworm
```
behrmann, line 5: `ubuntu-keyring` is also not available in bullseye, but Debian bullseye would match this, because (Debian != Debian) or (bullseye != bookworm) is true.

mkosi/resources/mkosi-tools/mkosi.conf.d/10-debian-ubuntu/mkosi.conf.d/virtiofsd.conf
```diff
@@ -0,0 +1,13 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+[Match]
+Distribution=|!ubuntu
+Release=|!jammy
+
+[Match]
+Distribution=|!debian
+Release=|!bookworm
```
behrmann, line 9: This is is a bit hard to reason about
```
((not ubuntu) or (not jammy)) and ((not debian) or (not bookworm)) =>
(not (ubuntu and jammy)) and (not (debian and bookworm)) =>
not ((ubuntu and jammy) or (debian and bookworm)
```
so I think this is correct, but it could do with a comment, so that readers don't have to do the boolean logic :)

behrmann (commented):
LGTM, but the [ubuntu-keyring](https://packages.debian.org/trixie/ubuntu-keyring) is not available in bullseye.

mkosi/resources/mkosi-tools/mkosi.conf.d/10-debian-ubuntu/mkosi.conf.d/ubuntu-keyring.conf
DaanDeMeyer: So bullseye is missing a ton of packages just like Focal, so I don't think we can easily support `mkosi dependencies` on it.

PR #2843, opened 2024-07-06
opensuse: Add support for aarch64

DaanDeMeyer:
This requires another rework for the repositories() method for OpenSUSE. Whereas before we picked up all the repositories from https://download.opensuse.org/tumbleweed/repo/, it turns out that the debug and source repositories can also be found at https://download.opensuse.org/debug/tumbleweed/repo/ and https://download.opensuse.org/source/tumbleweed/repo/ respectively. Furthermore, the latter locations are the only ones that are available for ports under https://download.opensuse.org/ports/ so we switch to those instead.

Additionally, openSUSE-current is not available for ports so we disallow using current, stable and release with architectures other than x86-64.

While there are more ports than just aarch64, for now let's just add aarch64 and wait for user demand before we add any others.

### DIFF
--- a/mkosi.conf.d/15-x86-64.conf
+++ b/mkosi.conf.d/15-x86-64.conf
@@ -2,6 +2,10 @@
 
 [Match]
 Architecture=x86-64
+# We cannot install the grub tools in the OpenSUSE tools tree due to
+# https://bugzilla.opensuse.org/show_bug.cgi?id=1227464.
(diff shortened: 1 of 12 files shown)


DaanDeMeyer:
Also including various tools tree and default image package fixes as extra commits on this PR

behrmann (approved):

PR #2844, opened 2024-07-06
run: Don't try to communicate inner pid for debug shell

DaanDeMeyer:
The pipe write end has already been closed by then so bash will fail with "bad file descriptor". Since there's no reason in having the inner pid in the debug shell let's make sure we don't try to communicate it there.


DaanDeMeyer:
Trivial fix so merging

PR #2850, opened 2024-07-09
Track local packages in cache manifest

DaanDeMeyer:
Let's track the modification timestamps of the packages from configured
package directories so we can rebuild the image if new packages appear
or are removed or changed.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -591,6 +591,7 @@ def run_prepare_scripts(context: Context, build: bool) -> None:
                             Mount(json, "/work/config.json", ro=True),
                             Mount(context.root, "/buildroot"),
                             Mount(context.artifacts, "/work/artifacts"),
+                            Mount(context.repository, "/work/packages"),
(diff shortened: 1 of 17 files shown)


mkosi/__init__.py
```diff
@@ -1787,7 +1780,9 @@ def finalize_default_initrd(
         "--acl", str(config.acl),
         *(f"--package={package}" for package in config.initrd_packages),
         *(f"--volatile-package={package}" for package in config.initrd_volatile_packages),
-        *(["--package-directory", str(package_dir)] if package_dir else []),
+        *(f"--package-directory={d}" for d in config.package_directories),
+        *(f"--volatile-package-directory={d}" for d in config.volatile_package_directories),
+        *([f"--volatile-package-directory={packages}"] if packages else []),
```
behrmann, line 1785: The `packages` parameter name is not ideal. It made me do a double take here, since `packages` here is a directory, but `package` above are actual package names.

behrmann (commented):
LGTM, one nit

PR #2854, opened 2024-07-10
Add `mkosi.pkgmngr` path to `PackageManagerTrees=`

NekkoDroid:
- Reword documentation to no relying on a reference later in the doc
- Test to ensure `paths=` and `default_factory=` work as expected


behrmann (approved):
That the package manager tree doesn't have a dir of its own was intentional when it was added, but I've run into expecting this to work out of the box myself a few times, so I think it's sensible to add this.

DaanDeMeyer:
I always thought the name was too ugly to make it a default thing, but let's go with this.

PR #2855, opened 2024-07-10
debian/ubuntu: Exclude docs in extract_tar() if WithDocs=no

### LINKED ISSUE #2851: WithDocs=false is broken on Debian

2f9e2cdde41c81161fce048e37e88743dff3709a

### Used host distribution

Ubuntu 24.04

### Used target distribution

Debian or Ubuntu

### Linux kernel version used

_No response_

None

`WithDocs=false` fails to remove the directories specified in apt.py: https://github.com/systemd/mkosi/blob/2f9e2cdde41c81161fce048e37e88743dff3709a/mkosi/installer/apt.py#L185-L193
```
$ mkosi --debug build
...
$ find image/usr/share/{doc,man,info} -type f | wc -l
767
$ du -hs image/usr/share/{doc,man,info}
5.6M    image/usr/share/doc
2.3M    image/usr/share/man
544K    image/usr/share/info
```
This happens because manual tar extraction part of the "debootstrap" step does not respect `config.with_docs`:
https://github.com/systemd/mkosi/blob/2f9e2cdde41c81161fce048e37e88743dff3709a/mkosi/distributions/debian.py#L177-L184

Changing the tar command to set `--exclude` fixes the issue:
```diff
diff --git a/mkosi/archive.py b/mkosi/archive.py
index 911e15e6..312a3f86 100644
--- a/mkosi/archive.py
+++ b/mkosi/archive.py
@@ -83,6 +83,11 @@ def extract_tar(
             "--xattrs",
             "--force-local",
             *tar_exclude_apivfs_tmp(),
+            "--exclude", "./usr/share/doc/*",
+            "--exclude", "./usr/share/man/*",
+            "--exclude", "./usr/share/groff/*",
+            "--exclude", "./usr/share/gtk-doc/*",
+            "--exclude", "./usr/share/info/*",
         ],
         sandbox=sandbox(
             binary="tar",
```

### Used mkosi config

```ini
[Output]
Format=directory

[Distribution]
Distribution=debian

[Content]
WithDocs=false
```

### mkosi output

https://gist.github.com/dzil123/c80fa6dfc8a156e8b209c64f3f2c66b2

DaanDeMeyer:
Fixes #2851

### DIFF
--- a/mkosi/installer/apt.py
+++ b/mkosi/installer/apt.py
@@ -40,6 +40,14 @@ class AptRepository:
 
 
 class Apt(PackageManager):
+    documentation_exclude_globs = [
+        "usr/share/doc/*",
+        "usr/share/man/*",
+        "usr/share/groff/*",
+        "usr/share/gtk-doc/*",
+        "usr/share/info/*",
+    ]
+
     @classmethod
     def executable(cls, config: Config) -> str:
         return "apt"
@@ -183,14 +191,8 @@ class Apt(PackageManager):
             ]
 
         if not context.config.with_docs:
-            cmdline += [
-                "-o", "DPkg::Options::=--path-exclude=/usr/share/doc/*",
-                "-o", "DPkg::Options::=--path-include=/usr/share/doc/*/copyright",
-                "-o", "DPkg::Options::=--path-exclude=/usr/share/man/*",
-                "-o", "DPkg::Options::=--path-exclude=/usr/share/groff/*",
-                "-o", "DPkg::Options::=--path-exclude=/usr/share/gtk-doc/*",
-                "-o", "DPkg::Options::=--path-exclude=/usr/share/info/*",
-            ]
+            cmdline += [f"--option=DPkg::Options::=--path-exclude=/{glob}" for glob in cls.documentation_exclude_globs]
+            cmdline += ["--option=DPkg::Options::=--path-include=/usr/share/doc/*/copyright"]
 
         if context.config.proxy_url:
             cmdline += [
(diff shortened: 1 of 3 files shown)


mkosi/installer/apt.py
```diff
@@ -40,6 +40,15 @@ def __str__(self) -> str:
 
 
 class Apt(PackageManager):
+    DOCUMENTATION_GLOBS = [
```
behrmann, line 43: PEP8 only advises uppercase for global variables. For class variables lower case is fine and I think it would nicer to keep uppercase for global things only.

But maybe it would also make sense to add this as a method so that all package managers can implement this.

mkosi/installer/apt.py
```diff
@@ -40,6 +40,15 @@ def __str__(self) -> str:
...
+        "usr/share/doc/*",
+        "usr/share/doc/*/copyright",
```
behrmann, line 45: That's already included in the line above, isn't it?

DaanDeMeyer: Ah shit this was an include to keep those

DaanDeMeyer:
We exclude the copyright as well when we extract the tars because tar does not have an `--include` option but this doesn't matter because when we install the packages again with `dpkg` later on it will extract all the tars again but keep the copyright intact this time.

mkosi/installer/apt.py
DaanDeMeyer: We don't really need this for any other package managers since for those we don't have to extract tars

behrmann (approved):

PR #2856, opened 2024-07-10
Downgrade SELinuxRelabel=yes to SELinuxRelabel=auto for the initrd

DaanDeMeyer:
Otherwise we fail if there's no policy installed in the initramfs. Putting a policy in the initramfs isn't really a thing so let's not fail if there is one available but still relabel if there is one installed.


bluca (approved):

DaanDeMeyer:
Trivial fix so let's not wait for CI

PR #2862, opened 2024-07-11
Add CacheKey config to override cache_tree_paths

dzil123:
My usecase doesn't fit in mkosi.images.

Currently, I'm managing my image variants with `include` and `image_id`. If I instead put my image variants in mkosi.images, each variant tries to build the same default-initrd image and invalidate the cache each time.

This doesn't work for me because each image variant should have a separate cache for itself and its unique default-initrd image.


DaanDeMeyer:
The thing is, you're introducing the assumption here that a different image ID equals a different set of packages that get installed, which is not always going to be true. With this change you force everyone that's using different image IDs without changing the packages that get installed per image ID to have a duplicate cache image per image ID.

I think introducing a `CacheKey=` setting might be a better approach. You could then set `CacheKey=%d-%r-%a-%i` whereas for everyone else it would still default to `CacheKey=%d-%r-%a`.

mkosi/__init__.py
```diff
@@ -1776,6 +1776,7 @@ def finalize_default_initrd(
         *(["--output-dir", str(output_dir)] if output_dir else []),
         *(["--workspace-dir", str(config.workspace_dir)] if config.workspace_dir else []),
         *(["--cache-dir", str(config.cache_dir)] if config.cache_dir else []),
+        *(["--cache-key", config.cache_key]),
```
dzil123, line 1779: This passes the finalized value with the specifiers replaced to the CLI call instead of the original template, but I'm not sure how else to implement this.

dzil123:
After further investigation, I'm not a fan of this solution because it evaluates the cache_key template at the moment that config is parsed, not at build time. This means that any config changes in mkosi.profiles, mkosi.images, etc will not affect cache_key and potentially break caching if the additional configs change the set of packages.

DaanDeMeyer:
@dzil123 We should do this via a second layer of specifiers, similar to what we do for `UnifiedKernelmageFormat=`, We can use `&` as the specifier character and only interpret the specifiers in the `cache_tree_paths()` function.

dzil123:
Thank you for the feedback. Unfortunately I'm out of time to work on this. I hope someone else will pick this up.

DaanDeMeyer:
Let's close this for now then, I filed https://github.com/systemd/mkosi/issues/2883 to track the feature request

PR #2876, opened 2024-07-16
Fail when trying to change universal settings in subimages

DaanDeMeyer:
Let's error out when users try to configure universal settings in subimages since these will always be overridden.


DaanDeMeyer:
cc @septatrix

behrmann (approved):

PR #2878, opened 2024-07-17
Various small improvements

DaanDeMeyer:


mkosi.postinst
```diff
@@ -0,0 +1,12 @@
...
+set -e
+
+mkosi-chroot \
+    useradd \
+    --user-group \
+    --create-home \
+    --password "$(openssl passwd -1 testuser)" \
+    --groups sudo,systemd-journal \
+    --shell /bin/bash \
+    testuser
```
behrmann, line 12: Let's be clear on where this comes from
```suggestion
    mkosiuser
```

behrmann (approved):

mkosi.postinst
DaanDeMeyer: I don't think it matters since it's the default image, this is intended for use by developers, not users

DaanDeMeyer: But we can name the user mkosi to make it shorter I guess

PR #2884, opened 2024-07-19
Make ParseContext a regular class

DaanDeMeyer:
Let's simplify the config parsing implementation by making ParseContext a regular class instead of a singleton. Additionally, we make ConfigAction a global class and slightly change the behavior of --include= on the command line by parsing all given includes after parsing all the other command line arguments.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -3358,19 +3358,42 @@ def resolve_deps(images: Sequence[argparse.Namespace], include: Sequence[str]) -
     return sorted(images, key=lambda i: order.index(i.image))
 
 
-def parse_config(argv: Sequence[str] = (), *, resources: Path = Path("/")) -> tuple[Args, tuple[Config, ...]]:
+class ConfigAction(argparse.Action):
+    def __call__(
+        self,
+        parser: argparse.ArgumentParser,
+        namespace: argparse.Namespace,
+        values: Union[str, Sequence[Any], None],
+        option_string: Optional[str] = None
+    ) -> None:
+        assert option_string is not None
 
-    class ParseContext:
+        if values is None and self.nargs == "?":
+            values = self.const or "yes"
+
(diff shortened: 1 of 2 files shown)


mkosi/config.py
```diff
@@ -3358,19 +3358,45 @@ def resolve_deps(images: Sequence[argparse.Namespace], include: Sequence[str]) -
...
+        assert option_string is not None
+
+        if values is None and self.nargs == "?":
+            values = self.const or "yes"
 
-    class ParseContext:
+        try:
+            s = SETTINGS_LOOKUP_BY_DEST[self.dest]
+        except KeyError:
+            die(f"Unknown setting {option_string}")
```
keszybz, line 3377: Shouldn't this be an assert?

mkosi/config.py
```diff
@@ -3358,19 +3358,45 @@ def resolve_deps(images: Sequence[argparse.Namespace], include: Sequence[str]) -
...
+
+        if values is None or isinstance(values, str):
+            setattr(namespace, s.dest, s.parse(values, getattr(namespace, self.dest, None)))
+        else:
+            for v in values:
+                assert isinstance(v, str)
+                setattr(namespace, s.dest, s.parse(v, getattr(namespace, self.dest, None)))
```
keszybz, line 3384: Save `getattr(namespace, self.dest, None)` to a temp var?

keszybz (commented):
Looks reasonable.

mkosi/config.py
DaanDeMeyer: We have to re-evaluate it every time anyway so I don't think this gets us much

mkosi/config.py
DaanDeMeyer: I just dropped the exception handling since we should never hit it in the first place

PR #2886, opened 2024-07-21
Allow multilib and multilib-testing repositories for arch distribution

gui-don:
I also noticed setting random values as repositories makes mkosi silently ignore them. Wouldn’t it be nice to have a small warning when a repository name is not expected for a given distribution? Helps debug typos.


DaanDeMeyer (approved):

DaanDeMeyer:
@gui-don Yeah this has been a bit of a pain point for a longer time. I haven't done it so far because it can't be implemented for every distribution. For every rpm based distribution we don't know all the repositories so we can't fail on unknown ones as those could be ones that were added by users via package manager trees

PR #2889, opened 2024-07-21
Make make_cpio() take a list of files relative to the root directory

DaanDeMeyer:
We operate on absolute paths all the time in kmod.py only for them to be made relative to the root directory just before they are passed to cpio. Let's save on the amount of allocations by always operating on paths relative to the root directory.

Because rglob() doesn't support returning paths relative to the given directory, we chdir() into the root directory before globbing instead.

### DIFF
--- a/mkosi/archive.py
+++ b/mkosi/archive.py
@@ -9,7 +9,7 @@ from mkosi.log import log_step
 from mkosi.run import run
 from mkosi.sandbox import Mount, SandboxProtocol, finalize_passwd_mounts, nosandbox
 from mkosi.types import PathString
-from mkosi.util import umask
+from mkosi.util import chdir, umask
 
 
 def tar_exclude_apivfs_tmp() -> list[str]:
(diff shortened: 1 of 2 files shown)


mkosi/archive.py
```diff
@@ -103,8 +103,8 @@ def make_cpio(
     sandbox: SandboxProtocol = nosandbox,
 ) -> None:
     if not files:
-        files = src.rglob("*")
-    files = sorted(files)
+        with chdir(src):
+            files = sorted(Path(".").rglob("*"))
```
behrmann, line 107: If files doesn't come in sorted already this the files in the cpio might not be sorted anymore and the whole thing less reproducible.

DaanDeMeyer: Yes but the only user of does does make sure the files are already sorted

behrmann: True, though it is something to keep in mind in addition. The overhead of sorting again should be negligible.

behrmann (approved):
Apart from the nit, LGTM.

PR #2895, opened 2024-07-22
Make sure mkosi works in the initramfs

DaanDeMeyer:
bubblewrap uses pivot_root() which doesn't work in the initramfs as pivot_root() requires / to be a mountpoint which is not the case in the initramfs. So, to make sure mkosi works from within the initramfs, let's make / a mountpoint by recursively bind-mounting / (the directory) to another location and then switching root into the bind mount directory.


behrmann (approved):
Awesome in both senses of the word :)

PR #2899, opened 2024-07-23
Resolve symlink for bind mount

akihikodaki:
Passing symlink as bind mount source fails so resolve it beforehand.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3,6 +3,7 @@
 import contextlib
 import dataclasses
 import datetime
+import errno
 import hashlib
 import io
 import itertools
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
@akihikodaki Can you give more information on exactly which error you get?

mkosi/__init__.py
```diff
@@ -4738,9 +4739,20 @@ def run_build(args: Args, config: Config, *, resources: Path, package_dir: Optio
             remount += ["/usr"]
 
         for d in remount:
-            if Path(d).exists():
-                options = "ro" if d in ("/usr", "/opt") else "ro,nosuid,nodev,noexec"
-                run(["mount", "--rbind", d, d, "--options", options])
```
DaanDeMeyer, line 4743: Please just call `.resolve()` on the source path here, that should be sufficient to fix the issue.

DaanDeMeyer (changes requested):

akihikodaki:
I'll close this issue because `mount` is fixed upstream: https://github.com/util-linux/util-linux/issues/2773
Please tell me if you think `mkosi` needs to have a workaround; I'll reopen with the suggested change.

PR #2905, opened 2024-07-24
Add `mkosi.bump` script

NekkoDroid:


mkosi/__init__.py
```diff
@@ -4368,24 +4368,37 @@ def generate_key_cert_pair(args: Args) -> None:
...
 
-    try:
-        m = int(v[-1])
-    except ValueError:
-        new_version = version + ".2"
-        logging.info(
-            "Last component of current version is not a decimal integer, "
-            f"appending '.2', bumping '{version}' → '{new_version}'."
-        )
+    if not (version_file := Path("mkosi.version")).exists():
```
behrmann, line 4376: No need for this to be defined inline, when it's used further below
```suggestion
    version_file = Path("mkosi.version")
    if not version_file.exists():
```

mkosi/__init__.py
```diff
@@ -4368,24 +4368,37 @@ def generate_key_cert_pair(args: Args) -> None:
...
+        die(f"Cannot bump image version, '{version_file}' not found")
+
+    version = version_file.read_text().strip()
+
+    if (script := Path("mkosi.bump").absolute()).exists():
```
behrmann, line 4381: All other scripts are configurable, so this should probably, too.

mkosi/__init__.py
```diff
@@ -4368,24 +4368,37 @@ def generate_key_cert_pair(args: Args) -> None:
...
+        if not os.access(script, os.X_OK):
+            die(f"{script} is not executable")
+
+        with complete_step(f"Running bump script {script}…"):
+            new_version = run([script, version], stdout=subprocess.PIPE).stdout.strip()
```
behrmann, line 4386: This should have some sandboxing so that it cannot write anything, with all the script running moved to a separate `run_bump_script` function.

mkosi/__init__.py
NekkoDroid: To be fair, `mkosi.version` in this case also isn't currently configurable for this to work.

I could possibly fix that by introducing a `image_version_file` config option and making `image_version` depend on that. Then the parameter for the script would likely be `image_version` and would just be written to `image_version_file` without needing to re-read `image_version_file`. Basically:

```py
with open(config.image_version_file, "w") as o:
	run([config.bump_script, config.image_version], stdout=o, ...)
```

Does this make sense?

And under what section should `bump_script` go, `[Host]`?

septatrix:
This gives me an idea for another related but not exclusive feature. Support for an executable mkosi.version which would just return a version to use, e.g. determined from the current date or git tag

NekkoDroid:
> Support for an executable mkosi.version which would just return a version to use, e.g. determined from the current date or git tag

I also had that idea a bit ago, just having that only determined by if the file is executable or not sounds like it easily could be unintentional.

It also makes IMO more sense, as you could then just `git tag` and have a build kick of when a new tag gets pushed, no need to commit a new version file.

NekkoDroid:
I just rebased and saw `IMAGE_ID` and `IMAGE_VERSION` were added to the env-vars of scripts. The thing is where are they set? They aren't set in the `env = dict(...)` of scripts.

behrmann:
> I just rebased and saw IMAGE_ID and IMAGE_VERSION were added to the env-vars of scripts. The thing is where are they set? They aren't set in the env = dict(...) of scripts.

`load_environment` prepopulates the environment for `config.environment`. We should probably also document the rest from there.

mkosi/__init__.py
```diff
@@ -4836,6 +4885,9 @@ def run_verb(args: Args, images: Sequence[Config], *, resources: Path) -> None:
         page(text, args.pager)
         return
 
+    if args.verb == Verb.bump:
```
behrmann, line 4888: Why move this down?

mkosi/__init__.py
```diff
@@ -4375,25 +4375,77 @@ def generate_key_cert_pair(args: Args) -> None:
     )
 
 
-def bump_image_version() -> None:
+def run_bump_script(config: Config, version: str) -> str:
```
behrmann, line 4378: Having it close to the bump function makes sense, but all the other `run_foo_script` are together, so let's add it there. We've been wanting to make them share code for a long time anyway.

mkosi/config.py
```diff
@@ -766,6 +766,14 @@ def config_default_dependencies(namespace: argparse.Namespace) -> Optional[list[
     return dependencies
 
 
+def config_default_image_version(namespace: argparse.Namespace) -> Optional[str]:
+    if not namespace.image_version_file:
+        return None
+
+    version_file = cast(Path, namespace.image_version_file)
+    return version_file.read_text().rstrip("\n")
```
behrmann, line 774: I think it makes more sense to just strip all whitespace to the right
```suggestion
    return version_file.read_text().rstrip()
```
Though it would probably be better to strip it all
```suggestion
    return version_file.read_text().strip()
```

mkosi/resources/mkosi.md
```diff
@@ -2054,6 +2065,12 @@ current working directory. The following scripts are supported:
   artifacts from `SplitArtifacts=yes` or RPMs built in a build script).
   Note that this script does not use the tools tree even if one is configured.
 
+* If **`mkosi.bump`** (`BumpScript=`) exists, it is executed when running
+  `mkosi bump` or after a successful build with the `--auto-bump` option.
+  A bump script can customize the logic used when incrementing from one version
+  to a newer version. The current version is passed as first argument to the
+  script and the new version to bump to is expected to be output to stdout.
```
behrmann, line 2072:
```suggestion
  to a newer version. The current version is passed as first and only argument to the
  script and the new version is expected on stdout.
```

mkosi/__init__.py
NekkoDroid: I wanted to move it at least after https://github.com/systemd/mkosi/blob/6710479778d03eb05ebe8dfe972c53737fa34f75/mkosi/__init__.py#L4821 just to ensure no default configs are used, as it now depends on a config being present. And after `summary` seemed like a fine position but I am unsure if it actually should now also be after `ensure_root_is_mountpoint()`.

behrmann:
Looks good to me, but it doesn't work for me, at the moment.

Give a version file that just contains `1.2.3`, `mkosi bump` does the job and increments it to `1.2.4`, but adding a bump script 
```bash
#!/bin/bash
set -x
echo '2.0.0'
```
gives
```
$ python -m mkosi bump
‣ Running bump script /home/drako/Documents/git/mkosi/mkosi.bump…
mount: /tmp/volatile-overlayt8jmgxml-mountpointkp_oxawh: must be superuser to use mount.
       dmesg(1) may have more information after failed mount system call.
‣ "mount --no-mtab overlay /tmp/volatile-overlayt8jmgxml-mountpointkp_oxawh --types overlay --options lowerdir=/home/drako/Documents/git/mkosi,upperdir=/tmp/volatile-overlayt8jmgxml,workdir=/tmp/volatile-overlayt8jmgxml-workdir7lj_91_a,index=off,metacopy=off,userxattr" returned non-zero exit code 32.
umount: /tmp/volatile-overlayt8jmgxml-mountpointkp_oxawh: must be superuser to unmount.
‣ "umount --no-mtab /tmp/volatile-overlayt8jmgxml-mountpointkp_oxawh" returned non-zero exit code 32.
```

So the bumping needs to move after a `become_root`.

NekkoDroid:
Weird, on my end that seem to work, so I can't test if I actually fixed the problem. ¯\\\_(ツ)_/¯

Anyway, I added a `become_root` before the `run_bump_script` as I didn't see any `become_root` that I could put it after and move it also after `ensure_root_is_mountpoint` as it does invoke bwrap.

behrmann:
I'm still getting:
```
$ python -m mkosi bump
‣ Running bump script ~/mkosi/mkosi.bump…
mount: /tmp/volatile-overlayv8bjfcsu-mountpointbg1t77ec: permission denied.
       dmesg(1) may have more information after failed mount system call.
‣ "mount --no-mtab overlay /tmp/volatile-overlayv8bjfcsu-mountpointbg1t77ec --types overlay --options lowerdir=/home/drako/Documents/git/mkosi,upperdir=/tmp/volatile-overlayv8bjfcsu,workdir=/tmp/volatile-overlayv8bjfcsu-workdir75tygpmx,index=off,metacopy=off,userxattr" returned non-zero exit code 32.
umount: /tmp/volatile-overlayv8bjfcsu-mountpointbg1t77ec: must be superuser to unmount.
‣ "umount --no-mtab /tmp/volatile-overlayv8bjfcsu-mountpointbg1t77ec" returned non-zero exit code 32.
```

NekkoDroid:
Oh wait, I just realized you were running this in the repo directory. There I am able to reproduce it. It seems a fix for this would be by changing to `finalize_source_mounts(config, ephemeral=False) as sources` in `run_bump_script`, but that doesn't seem like the right fix for this.

NekkoDroid:
After a while of testing I haven't found a way to get it to work with `config.build_sources_ephemeral`. I tried `fork_and_wait`, `preexec_fn=become_root` as well as different locations for the `become_root` and none of them fixed the issue.

It also seems that all of the `finalize_source_mounts` that don't set `ephemeral=False` all use `context.sandbox` (which doesn't exist yet) and `context.config.build_sources_ephemeral`. This includes `run_clean_scripts`, `run_configure_scripts`, `run_sync_scripts` and `run_shell`

NekkoDroid:
Anything else I could try? I am all out of ideas and all my testing resulted in the same outcome so for now I just set it to `ephemeral=False`.

I had the idea of replacing `BuildSourcesEphemeral=` with a `BuildSourcesMode=` that takes in `ReadOnly`, `ReadWrite` and `Ephemeral`. This would allow mounting the sources read-only without needing to mount an overlay.

behrmann:
> Anything else I could try?

Sorry, I didn't have much time to look further into this week, but I do want to see this land, since I have use for this :)

> It also seems that all of the finalize_source_mounts that don't set ephemeral=False all use context.sandbox (which doesn't exist yet) and context.config.build_sources_ephemeral. This includes run_clean_scripts, run_configure_scripts, run_sync_scripts and run_shell

We added #2930 yesterday and I expected this to trip the above described behaviour, but it didn't. What exactly did you run into there?

NekkoDroid:
> Sorry, I didn't have much time to look further into this week, but I do want to see this land, since I have use for this :)

I also had use for this until I actually sat down to implement https://github.com/systemd/mkosi/issues/2915. It's a relativly small patch that built ontop of this, but sadly no sandboxing since at parse time there isn't much you can rely on :(

> What exactly did you run into there?

It was more something I noticed, rather than weird/erronious behavior. All the scripts that can use overlays (via `build_source_ephemeral`) use the `context.sandbox()` instead of the `config.sandbox()` and those using `config.sandbox()` all set `ephemeral=` to `False`. At the point where the bump script is being invoked there is no `context` object I could invoke to test if that would solve the issue.

mkosi/resources/mkosi.md
```diff
@@ -625,6 +625,12 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     `/usr/lib/os-release` or similar, in particular the `IMAGE_VERSION=`
     field of it).
 
+`ImageVersionFile=`, `--image-version-file=`
+:   Takes a path to a file to read the `ImageVersion=` from if not
+    explicitly set. The file will also be used when running `mkosi bump`
+    or when building with the `--auto-bump` option enabled. This value
+    default to `mkosi.version` if it is found in the config directory.
```
DaanDeMeyer: I'm not a big fan of a separate setting for this. I'd prefer to just make `mkosi.version` the default path for `ImageVersion=`

mkosi/config.py
```diff
@@ -3075,6 +3076,15 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
...
         help="Set the path used to store forwarded machine journals",
     ),
+    ConfigSetting(
+        dest="bump_script",
+        metavar="PATH",
+        section="Host",
+        parse=config_make_path_parser(),
+        paths=("mkosi.bump",),
+        help="Bump script to increment the version number",
+        scope=SettingScope.local,
```
DaanDeMeyer, line 3086: This is the default so this can be dropped

mkosi/config.py
```diff
@@ -3075,6 +3076,15 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
         parse=config_make_path_parser(required=False),
         help="Set the path used to store forwarded machine journals",
     ),
+    ConfigSetting(
+        dest="bump_script",
+        metavar="PATH",
+        section="Host",
```
DaanDeMeyer, line 3082: Should be same section as the image version so `Output`

mkosi/resources/mkosi.md
NekkoDroid: Dropped the second commit.

mkosi/config.py
NekkoDroid: Done. Had it as `Host` mostly because it is something entirely run on the host and doesn't really have much to do with the actual output (just to explain my thought process).

mkosi/__init__.py
```diff
@@ -4386,23 +4386,26 @@ def generate_key_cert_pair(args: Args) -> None:
...
+    if not version_file.exists():
+        die(f"Cannot bump image version, '{version_file}' not found")
+
+    version = version_file.read_text().strip()
     v = version.split(".")
 
     try:
-        m = int(v[-1])
+        v[-1] = str(int(v[-1]) + 1)
+        logging.info("Increasing last component of version by one")
```
DaanDeMeyer: This is excessive, I'd drop it

NekkoDroid: Am I correct in assuming you mean the log message and not the commit?

behrmann (approved):
Just gave it a try, works for me now.

behrmann:
@NekkoDroid I thought about the executable `mkosi.version` and I do find the idea appealing, but I don't think it's a general solution, since the easiest case of just having a static version needs some state and having that just be (the output of) a script would make incrementing the value manual again, which `mkosi bump` was meant to solve (be that a good idea or not in hindsight).

I think we might have our cake and eat it, too, based on the work in this PR, by making the `mkosi.bump` into `mkosi.dynamicversion` (or rather a better name). This setup would leave us with a state file `mkosi.version` and a script `mkosi.dynamicversion`. The script would receive the contents of `mkosi.version` on stdin if the file exists and either the literal string `version` or `bump` as the first argument. 

During build when we need a version it would be called with the `version` argument and define the version used for the build. If the autobump option is used or mkosi is called with the `bump` verb, it would receive that as an argument and the output would be written to `mkosi.version`, if it exists, at the end. So, the version script would be called at least once and at most twice during a run. 

If no version script exists, the version file content would be passed through untouched and used as the version. If no version file exists only dynamically generated versions are possible and the version cannot be bumped (because there is no file to write it to).

I think this should cover all bases and be backwards compatible. Thoughts?

NekkoDroid:
That sounds more complicated than it needs to be.

I just don't see how it would make sense to have an executable `mkosi.version` when the version isn't already tracked by some other means (be it git, date or some other meta build), so us having another file that would track the version just doesn't make sense to me.

It could be as simple as if `mkosi.version` is exectuable run it and read stdout else read the file contents and when `mkosi bump` gets called run `mkosi.bump` (which then can modify external state via build sources) write the output to `mkosi.version`, unless it's executable.

This would cover the following cases:

- no `mkosi.version`: No bumping is possible
- static `mkosi.version` without `mkosi.bump`: Internal bump logic gets used and written back to `mkosi.version`
- static `mkosi.version` with `mkosi.bump`: `mkosi.bump` logic gets used and written back to `mkosi.version`
- dynamic `mkosi.version` without `mkosi.bump`: version is tracked externally and there is no way to bump (e.g. date based)
- dynamic `mkosi.version` with `mkosi.bump`: version is tracked externally and the external source can be modified by `mkosi.bump` and `BuildSource=` (e.g. `git tag`)

What would be a case where we would need to keep track of the version in `mkosi.version` but also have the version dynamically querried?

behrmann:
> I just don't see how it would make sense to have an executable mkosi.version when the version isn't already tracked by some other means (be it git, date or some other meta build), so us having another file that would track the version just doesn't make sense to me.

That assumes one would want to modify build sources. Having state separate from build sources one puts in, seems more sensible to me, but in the end, I mostly care about date-based versioning, so I can do without state.

That being said, we could probably do without adding another script to generate a version, since a version can also be added by a configure script.

NekkoDroid:
I think I'll just open a PR after this is merged, with how I intended to implement it and this can be discussed there.

NekkoDroid:
Anything I could do to move this along?

behrmann:
Really sorry for the holdup! 

> I think I'll just open a PR after this is merged, with how I intended to implement it and this can be discussed there.

I spoke to @DaanDeMeyer earlier and a `mkosi.version` script does seem like a more important/better thing to have. 

The bump feature in general might not have been the most well thought out thing in hindsight and it's probably better to leave it as is, since if we merge it, we'll have to continue supporting it for the future.

If you could open the PR that would be great.

PR #2908, opened 2024-07-24
Various kernel modules improvements

DaanDeMeyer:


mkosi/kmod.py
```diff
@@ -106,9 +106,14 @@ def resolve_module_dependencies(
...
                 key, sep, value = line.partition("=")
 
-            if key in ("depends", "softdep"):
+            if key == "depends":
                 depends += [normalize_module_name(d) for d in value.strip().split(",") if d]
 
+            elif key == "softdep":
+                # softdep is delimited by spaces and can contain strings like pre: and post: so discard anything that
+                # ends with a colon.
+                depends += [normalize_module_name(d) for d in value.strip().split(" ") if d and not d.endswith(":")]
```
behrmann, line 115: If the delimiter is just whitespace and not single spaces, then maybe like this, since it allows to drop the check for empty parts
```suggestion
                depends += [normalize_module_name(d) for d in value.strip().split() if not d.endswith(":")]
```

behrmann (approved):

PR #2909, opened 2024-07-24
kernel-install: Only build UKI if we're configured as the UKI generator

DaanDeMeyer:


DaanDeMeyer:
We don't cover this with CI yet and the reporter confirmed the issue is fixed so let's merge.

PR #2916, opened 2024-07-27
completion: fix bash completion script

behrmann:
Template the options definitions directly into the completion function, since for some weird scoping reasons even though the script is read fine and when running a shell with set -x one can see e.g. _mkosi_options being assigned the proper values, the completion function still uses '' for "${_mkosi_options[*]}".

This wasn't caught during development because the script works fine when sourced.


behrmann:
Thanks at Arch's gromit for catching this.

DaanDeMeyer (approved):

PR #2917, opened 2024-07-27
kali: A distribution based on Debian: https://www.kali.org/

mtdcr:
Kali includes many packages suitable for offensive security tasks. It follows a rolling release model and serves fewer architectures than Debian.


DaanDeMeyer:
@mtdcr Can you write a little more on why you'd like mkosi to support Kali? In other words, what's your specific use case for this?

mtdcr:
I use it to build custom tailored images for penetration testing. Kali contains a huge amount of useful packages that aren't available in Debian or any other supported distribution. Using mkosi makes it easy to share the build instructions with my team and allows them to customize their images easily.

mkosi/distributions/kali.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class Installer(debian.Installer):
+    @classmethod
+    def pretty_name(cls) -> str:
+        return "Kali"
+
+    @classmethod
+    def default_release(cls) -> str:
+        return "kali-rolling"
```
DaanDeMeyer, line 21: Please just make this `rolling`

mkosi/distributions/kali.py
```diff
@@ -0,0 +1,66 @@
...
+from mkosi.distributions import Distribution, debian
+from mkosi.installer.apt import AptRepository
+from mkosi.log import die
+from mkosi.util import listify
+
+
+class Installer(debian.Installer):
+    @classmethod
+    def pretty_name(cls) -> str:
+        return "Kali"
```
DaanDeMeyer, line 17: The official name is "Kali Linux" I think so let's use that here.

mkosi/distributions/kali.py
```diff
@@ -0,0 +1,66 @@
...
+        return "kali-rolling"
+
+    @classmethod
+    def default_tools_tree_distribution(cls) -> Distribution:
+        return Distribution.kali
+
+    @staticmethod
+    @listify
+    def repositories(context: Context, local: bool = True) -> Iterable[AptRepository]:
+        types = ("deb", "deb-src")
```
DaanDeMeyer, line 30: This can be inlined since it's only used once

mkosi/distributions/kali.py
```diff
@@ -0,0 +1,66 @@
...
+        components = ("main", *context.config.repositories)
```
DaanDeMeyer, line 31: Same here

mkosi/distributions/kali.py
```diff
@@ -0,0 +1,66 @@
...
+            yield AptRepository(
+                types=("deb",),
+                url=context.config.local_mirror,
+                suite=context.config.release,
+                components=("main",),
+                signedby=None,
+            )
+            return
+
+        mirror = context.config.mirror or "http://http.kali.org/kali"
```
DaanDeMeyer, line 43: Can also be inlined

mkosi/distributions/kali.py
```diff
@@ -0,0 +1,66 @@
...
+        signedby = Path("/usr/share/keyrings/kali-archive-keyring.gpg")
```
DaanDeMeyer, line 44: And so can this

mkosi/distributions/kali.py
```diff
@@ -0,0 +1,66 @@
...
+    def architecture(cls, arch: Architecture) -> str:
+        a = {
+            Architecture.arm64       : "arm64",
+            Architecture.arm         : "armhf",
+            Architecture.x86_64      : "amd64",
+            Architecture.x86         : "i386",
+        }.get(arch)
+
+        if not a:
+            die(f"Architecture {arch} is not supported by Kali")
```
DaanDeMeyer, line 64: Please use `cls.pretty_name()` in the error message

DaanDeMeyer (changes requested):
This is missing updates to the documentation, the tools tree definitions in `mkosi/resources/mkosi-tools`, initramfs definitions in `mkosi/resources/mkosi-initrd` and the default image in `mkosi.conf.d`. The documentation should be updated to mention Kali where we mention the other distributions. The definitions should be updated to make sure the shared Debian/Ubuntu configs also apply to Kali Linux where applicable.

The end result should be that I can run `mkosi -d kali --tools-tree -t disk -f qemu` from the mkosi repository and get a root shell in the booted Kali image just like I can do with the other distributions.

mkosi/distributions/kali.py
DaanDeMeyer: Ah in that case nvm

DaanDeMeyer:
@mtdcr Please rebase on the git main branch as well

mtdcr:
- `mkosi -d kali --tools-tree -t disk -f qemu` works now.
- Every spot where I could find a mention of Debian or Ubuntu was adjusted to include Kali, where it made sense to me.
- The example from docs/bootable.md can be built.
- `Release=testing` was removed from `mkosi.conf.d/20-debian/mkosi.conf`, because it was the default and it made the file reusable for kali.
- `mkosi/resources/mkosi-tools/mkosi.extra/usr/share/keyrings/kali-archive-keyring.gpg` had to be added, in order to enable Debian tools to install from Kali sources. Would there have been a better way?

mtdcr:
@DaanDeMeyer do you see any relation between the two failing tests and my patch?

bluca:
> * `mkosi/resources/mkosi-tools/mkosi.extra/usr/share/keyrings/kali-archive-keyring.gpg` had to be added, in order to enable Debian tools to install from Kali sources. Would there have been a better way?

Is the keyring not available in a package in debian/ubuntu? If not, we should get it uploaded

mkosi.conf.d/20-debian-kali/mkosi.conf
```diff
@@ -1,10 +1,10 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
 [Match]
-Distribution=debian
+Distribution=|debian
+Distribution=|kali
 
 [Distribution]
-Release=testing
 Repositories=non-free-firmware
```
DaanDeMeyer, line 8: Please keep this file debian specific and add a new directory for kali

mkosi/distributions/kali.py
```diff
@@ -0,0 +1,60 @@
...
+            signedby=Path("/usr/share/keyrings/kali-archive-keyring.gpg"),
+        )
+
+    @classmethod
+    def architecture(cls, arch: Architecture) -> str:
+        a = {
+            Architecture.arm64       : "arm64",
+            Architecture.arm         : "armhf",
+            Architecture.x86_64      : "amd64",
+            Architecture.x86         : "i386",
```
DaanDeMeyer, line 54: Reduce indentation here

DaanDeMeyer (changes requested):
Yeah we're not adding the kali archive keyring to the mkosi repository. Please install the relevant package instead.

mtdcr:
> Yeah we're not adding the kali archive keyring to the mkosi repository. Please install the relevant package instead.

Right. I can't reproduce the failure I previously ran into without the key file anymore. Probably something that was only relevant to a prior state of the patch. With your requested changes applied, it looks good to me now.

mkosi.conf.d/20-kali/mkosi.conf
```diff
@@ -0,0 +1,11 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+[Match]
+Distribution=kali
+
+[Distribution]
+Repositories=non-free-firmware
+
```
DaanDeMeyer, line 8: Is this an actual repository for Kali or did you just copy paste this?

behrmann:
We're running both Debian and Ubuntu and even thought it seems somewhat wasteful, we should run Kali as well, when we add it, otherwise we will regress, probably sooner rather than later.

DaanDeMeyer (changes requested):
`kali-archive-keyring` should be installed in the default Kali tools tree and this should be reflected in the tools tree installed packages documentation

DaanDeMeyer:
> We're running both Debian and Ubuntu and even thought it seems somewhat wasteful, we should run Kali as well, when we add it, otherwise we will regress, probably sooner rather than later.

I am not going to investigate Kali specific failures though. I'll fix any breakages introduced by the PR itself, but if Kali starts failing for any other reason not introduced by my PRs, I'll disable the CI again rather than investigating. Maintaining the existing distributions is enough work as it is. So I'm OK with adding Kali specific CI, but if it starts failing for reasons unrelated to changes in mkosi, I won't spend time investigating why.

mkosi.conf.d/20-kali/mkosi.conf
mtdcr: As Kali follows Debian (testing) closely, it is a real and required repository (for microcode packages).

mkosi/resources/mkosi-tools/mkosi.prepare
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+# Copy the user-installed keyring from the host to the buildroot to avoid
+# failing signature verification. Do not overwrite existing files.
+if [ "$1" = "final" ] && [ -f /usr/share/keyrings/kali-archive-keyring.gpg ] && [ -d "$BUILDROOT/usr/share/keyrings" ]; then
+    cp -anv /usr/share/keyrings/kali-archive-keyring.gpg "$BUILDROOT/usr/share/keyrings/kali-archive-keyring.gpg"
+fi
```
DaanDeMeyer, line 8: This should not be required, please drop it and rely on the package being installed instead.

DaanDeMeyer (changes requested):

mkosi/resources/mkosi-tools/mkosi.prepare
mtdcr: The package *is* installed on the host, but it is not available from Debian sources (or anywhere else than Kali itself). Thus the need to copy the file from the host. Otherwise the build fails because apt can not verify signatures without the keys, and then it reuses cached package lists without non-free-firmware.

I can probably remove it if it is ok to remove the microcode packages, too. But it would just be hiding a problem with the assumptions of the existing build logic.

mtdcr: @bluca: I haven't found any. You can try it, but this won't solve the problem for Fedora and others. Links to sources are in the commit message.

DaanDeMeyer: @mtdcr It's up to those distros to package the kali keyring if you want to build Kali images from those distros. Until that happens you'll need to be on a Kali host to build a Kali image or manually put the keyring in /usr/share/keyrings on the host machine used to build the Kali image. This is explicitly how mkosi chooses to operate, so please drop the prepare script.

mkosi.conf.d/20-kali/mkosi.conf
```diff
@@ -0,0 +1,14 @@
...
+[Host]
+ToolsTreeDistribution=kali
```
DaanDeMeyer, line 10: Please drop this. To build a Kali tools tree, the host system has to be Kali as well, otherwise the keyring won't be available since it's not shipped anywhere else (yet). But if the host system is Kali, the tools tree distribution is automatically chosen to be Kali already, so setting this is pointless.

DaanDeMeyer (changes requested):
One more small change and then it LGTM

mkosi.conf.d/20-kali/mkosi.conf
mtdcr: Because this was added just to fulfill your requirement of being able to run `mkosi -d kali --tools-tree -t disk -f qemu` from the mkosi repository, which I don't need for my use-case, I'm fine with breaking this command for people who installed the keyring manually on non-Kali setups.

A new version was pushed with this statement removed.

PR #2924, opened 2024-07-29
Set up proper environment variables for kernel-install

DaanDeMeyer:
If we're not explicitly disabling kernel-install during package manager invocations, let's set up the environment to make it do the right thing instead.

### DIFF
--- a/mkosi/installer/__init__.py
+++ b/mkosi/installer/__init__.py
@@ -47,6 +47,14 @@ class PackageManager:
             context.config.bootable != ConfigFeature.disabled
         ):
             env["KERNEL_INSTALL_BYPASS"] = "1"
+        else:
+            env |= {
+                "BOOT_ROOT": "/boot",
+                # Required to make 90-loaderentry.install put the right paths into the bootloader entry.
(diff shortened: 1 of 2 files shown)


mkosi/installer/__init__.py
```diff
@@ -47,6 +47,13 @@ def finalize_environment(cls, context: Context) -> dict[str, str]:
             context.config.bootable != ConfigFeature.disabled
         ):
             env["KERNEL_INSTALL_BYPASS"] = "1"
+        else:
+            env |= {
+                "SYSTEMD_XBOOTLDR_PATH": "/boot",
+                "SYSTEMD_ESP_PATH": "/efi",
+                # Required to make 90-loaderentry.install put the right paths into the bootloader entry.
+                "BOOT_MNT": "/boot",
```
behrmann, line 55: Instead of setting all these variables, shouldn't we just set `BOOT_ROOT`? That seems more in line what is actually advertised.

DaanDeMeyer: We will still need to set `$BOOT_MNT` regardless unfortunately.

PR #2926, opened 2024-07-29
Look for $USER for the username before reading /etc/passwd

DaanDeMeyer:
Let's take $USER into account if set before reading /etc/passwd for the username. This gives a way out for environments where the uid of the user does not have an entry in /etc/passwd.

### DIFF
--- a/mkosi/user.py
+++ b/mkosi/user.py
@@ -39,7 +39,7 @@ class INVOKING_USER:
     @classmethod
     @functools.lru_cache(maxsize=1)
     def name(cls) -> str:
-        return pwd.getpwuid(cls.uid).pw_name
+        return os.getenv("USER", pwd.getpwuid(cls.uid).pw_name)
 
     @classmethod
     @functools.lru_cache(maxsize=1)


mkosi/user.py
```diff
@@ -39,7 +39,7 @@ def is_running_user(cls) -> bool:
     @classmethod
     @functools.lru_cache(maxsize=1)
     def name(cls) -> str:
-        return pwd.getpwuid(cls.uid).pw_name
+        return os.getenv("USER", pwd.getpwuid(cls.uid).pw_name)
```
behrmann, line 42: I think asking first and falling back to the env would be better, but I'm fine either way.
```suggestion
        try:
            return pwd.getpwuid(cls.uid).pw_name
        except KeyError:
            if name := os.getenv("USER"):
                return name
            raise
```

behrmann (approved):
I think as a fix it's good enough for now, but I think

mkosi/user.py
DaanDeMeyer: I think it's more consistent to let environment win like we do elsewhere

PR #2930, opened 2024-08-01
ci: add nop versions of all scripts to our integration tests

behrmann:
This is to investigate @NekkoDroid's comment in https://github.com/systemd/mkosi/pull/2905#issuecomment-2260984613

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -162,6 +162,24 @@ jobs:
         DefaultDeviceTimeoutSec=180
         EOF
 
+        # fail if the script already exists, to avoid hard to debug CI errors
+        [[ -f mkosi.configure ]] && exit 1
+        tee mkosi.configure <<EOF
+        echo "Hello from inside mkosi.configure!" >&2
+        cat
+        EOF
+        chmod +x mkosi.configure
+
+        # prepare and postinst are already used in CI
+        for script in sync build finalize postoutput clean
+        do
+            [[ -f "mkosi.${script}" ]] && exit 1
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -162,6 +162,18 @@ jobs:
         DefaultDeviceTimeoutSec=180
         EOF
 
+        tee mkosi.configure <<EOF
+        echo "Hello from inside mkosi.configure!" >&2
+        cat
+        EOF
+
+        for script in sync prepare buuild postinst finalize postoutput clean
```
DaanDeMeyer, line 170: typo

.github/workflows/ci.yml
```diff
@@ -162,6 +162,19 @@ jobs:
         DefaultDeviceTimeoutSec=180
         EOF
 
+        tee mkosi.configure <<EOF
+        echo "Hello from inside mkosi.configure!" >&2
+        cat
+        EOF
+
+        for script in sync prepare build postinst finalize postoutput clean
```
DaanDeMeyer, line 170: This overrides the existing `mkosi.postinst` script no?

behrmann: And the prepare, too, dropped both from the list.

DaanDeMeyer (approved):

DaanDeMeyer (changes requested):
Actually can you check if each script already exists and fail the CI if it does? Otherwise someone's going to run into some hard to debug issue when they try to add a script for the default image.

DaanDeMeyer (approved):

Issue #757, opened 2021-07-08
Use only locally available GPG keys


calestyo:
Hey.

Not sure whether I've understood the code correctly, but it seems that `mkosi` quite easily falls back to disabling signature checks for at lest the DNF/YUM case:
https://github.com/systemd/mkosi/blob/28e9da0f0e822dada347cb3e4563bf1e5fe2d1e9/mkosi/__init__.py#L1825-L1826

which seems to be set here https://github.com/systemd/mkosi/blob/main/mkosi/__init__.py#L1850-L1856

My understanding is, that it first looks locally for the key, and if not there falls back to simply downloading it if there is a URL specified (in mkosi) itself, and if not it simply doesn't verify signatures at all.
It also doesn't seem to be obviously documented whether that key must be in the host OS or inside the container image already.

For security reasons, it should per default only use/trust locally installed keys. Additionally perhaps even ignore any `gpgcheck=0` or `gpgkey=` that use non-local key locations if possible.

Simply downloading a key via HTTPS doesn't seem really much more secure either. The usual CA bundles (e.g. Mozilla) come with some ~150 root CAs... (which themselves have thousands of intermediate CAs) many of them under the control of quite questionable organisations.

The same principles would obviously apply to `debootstrap`, etc. as well.. but at least for debootstrap I couldn't find that similar things are done (though I didn't look were closely).

So would it be possible to default to only using local keys and doing anything further like downloading/trusting keys or simply not checking sigs only with some `--i-really-know-what-i-am-doing-promised` option?

Another way would be that mkosi ships those keys, which obviously has some pros and cons.
Pros, that people using mkosi would at least get all the same keys (and not having to download them - more or less securely - on their own).

Cons, that you guys would have to maintain this... and also that e.g. in the case of bootstrapping Debian/Ubuntu on a Debian/Ubuntu system their own keyring packages should be rather used than what mkosi ships.

Cheers,
Chris.

DaanDeMeyer:
For dnf, we do explicitly warn when we're not using gpg checking. We can't really rely only on local keys by default since local gpg keys can only be relied on to be available if we're building the same distro as the host distro. I'm open to dropping the fallback that disables gpg checking though. Or we add it behind an option so that it's opt-in and not used by default

calestyo:
A warning is in practise probably useless,... in automatic deployments no one notices them anyway,.. and even if, possible rogue packages would have already been installed.

[...]

PR #2931, opened 2024-08-01
 Introduce RepositoryKeyFetch=

DaanDeMeyer:
This setting controls whether we'll fetch GPG keys remotely or not.
We disable it by default so that we only rely on locally available GPG
keys for checking package and repository metadata signatures.

This new setting only affects dnf/zypper based distributions as apt
and pacman do not support retrieving GPG keys remotely in the first
place.

zypper does not trust GPG keys listed in gpgkey= by default so we import
local GPG keys manually with rpm to work around that.

Fixes https://github.com/systemd/mkosi/issues/757


mkosi/resources/mkosi-tools/mkosi.conf.d/10-debian-ubuntu/mkosi.conf.d/grub.conf
```diff
@@ -1,8 +0,0 @@
-# SPDX-License-Identifier: LGPL-2.1-or-later
-
-[Match]
-Architecture=x86-64
```
behrmann, line 4: That's a slight change, because the package also exists for `x86`, but that shouldn't be a problem.

behrmann (approved):

mkosi/installer/rpm.py
```diff
@@ -22,8 +23,31 @@ class RpmRepository:
...
+) -> str: ...
+
+@overload
+def find_rpm_gpgkey(
+    context: Context,
+    key: str,
+    fallback: Optional[str] = None,
+    *,
+    required: Literal[False]
+) -> Optional[str]: ...
```
behrmann, line 42: More whitespace
```suggestion

@overload
def find_rpm_gpgkey(
    context: Context,
    key: str,
    fallback: Optional[str] = None,
    *,
    required: Literal[True] = True,
) -> str: ...

@overload
def find_rpm_gpgkey(
    context: Context,
    key: str,
    fallback: Optional[str] = None,
    *,
    required: Literal[False]
) -> Optional[str]: ...

```

behrmann (approved):

behrmann (approved):

PR #2932, opened 2024-08-01
ci: also ensure we can write in scripts

behrmann:


behrmann:
Doesn't obviously break anything, but this should move into the integration tests themselves.

PR #2940, opened 2024-08-05
Make `SYSEXT_SCOPE` and `CONFEXT_SCOPE` configurable

### LINKED ISSUE #2938: Not supporting ImageId= when building multiple images breaks system extension image building

24

### Used host distribution

Ubuntu 24.04

### Used target distribution

Ubuntu 24.04

### Linux kernel version used

6.8.0-39-generic

x86_64

Sub-images no longer allow `ImageId=`

Thinking logically, the main image that could be running on a laptop might have an image ID of laptop. If you want to build several system extension images which need their own image ID then you're now out of luck. It makes complete sense to use the laptop image as a base and build the extension images as an overlay.

### Used mkosi config

```ini
[Output]
ImageId = ...
```

### mkosi output

```sh
Setting ImageId cannot be configured in subimage ...
```

NekkoDroid:


mkosi/__init__.py
```diff
@@ -293,17 +293,20 @@ def configure_extension_release(context: Context) -> None:
...
-            f.write(f"{prefix}_VERSION_ID={context.config.image_version}\n")
+        if f"{prefix}_ID" not in extrelease:
+            if id := context.config.environment.get(f"{prefix}_ID", context.config.image_id):
+                f.write(f"{prefix}_ID={id}\n")
 
-        if f"{prefix}_SCOPE" not in extrelease:
-            f.write(f"{prefix}_SCOPE=initrd system portable\n")
+        if f"{prefix}_VERSION_ID" not in extrelease:
+            if version_id := context.config.environment.get(f"{prefix}_VERSION_ID", context.config.image_version):
+                f.write(f"{prefix}_VERSION_ID={version_id}\n")
```
DaanDeMeyer, line 305: I'm not sure about these two, I think we should stick to the existing settings here since we have them already. I don't think we gain much by also picking up the environment variables

mkosi/__init__.py
```diff
@@ -293,17 +293,20 @@ def configure_extension_release(context: Context) -> None:
...
 
-        if "ARCHITECTURE" not in extrelease:
-            f.write(f"ARCHITECTURE={context.config.architecture}\n")
+        if f"{prefix}_SCOPE" not in extrelease:
+            if scope := context.config.environment.get(f"{prefix}_VERSION_ID"):
+                f.write(f"{prefix}_SCOPE={scope}")
```
DaanDeMeyer, line 309: Copy paste bug? Also we should keep the existing default.

DaanDeMeyer (changes requested):

mkosi/__init__.py
NekkoDroid: Indeed was a copy-paste error. I do think having it by default use "upstream default" in this case ("system portable") does make more sense unless explicitly a different value is wanted.

But if you are adament on wanting to keep the existing default I can do that.

DaanDeMeyer (approved):

mkosi/__init__.py
```diff
@@ -300,7 +300,8 @@ def configure_extension_release(context: Context) -> None:
             f.write(f"{prefix}_VERSION_ID={context.config.image_version}\n")
 
         if f"{prefix}_SCOPE" not in extrelease:
-            f.write(f"{prefix}_SCOPE=initrd system portable\n")
+            if scope := context.config.environment.get(f"{prefix}_SCOPE", "initrd system portable"):
```
behrmann, line 303: Why guard this behind an if? The only case this condition can be false is if `{prefix}_SCOPE` is defined but empty.  If this is intended, than a comment would be good here, to say that this is indeed a case that's been considered and we want to allow people to explicitly unset this, otherwise it might make sense to write the empty value and leave the using software to deal with this.

DaanDeMeyer:
Should also be documented in "Environment Variables" in the readme I guess

mkosi/__init__.py
NekkoDroid: Was a leftover from before I added the default back

mkosi/__init__.py
```diff
@@ -300,7 +300,7 @@ def configure_extension_release(context: Context) -> None:
             f.write(f"{prefix}_VERSION_ID={context.config.image_version}\n")
 
         if f"{prefix}_SCOPE" not in extrelease:
-            f.write(f"{prefix}_SCOPE=initrd system portable\n")
+            f.write(f"{prefix}_SCOPE={context.config.environment.get(f"{prefix}_SCOPE", "initrd system portable")}\n")
```
DaanDeMeyer, line 303: Please use single quotes for the inner quotes to keep python 3.9 compat

DaanDeMeyer (changes requested):

behrmann:
LGTM now, apart from the quoting Daan mentioned that also trips our pyright CI

PR #2942, opened 2024-08-05
Clarify that inherited settings take priority over CLI/main settings

DaanDeMeyer:
If "inherited" settings are configured explicitly in subimages, those
values should take priority over values configured in the main config
and CLI.


DaanDeMeyer:
Actually I'm not entirely sure this is what we want, the current behavior seems better.

Issue #2874, opened 2024-07-15
`PackageManagerTree=` in image directory no longer falls back to `mkosi.skeleton/`


septatrix:
### mkosi commit the issue has been seen with

main

### Used host distribution

Fedora 40

### Used target distribution

Fedora 40

### Linux kernel version used

6.9.8-200.fc40.x86_64

### CPU architectures issue was seen on

None

### Unexpected behaviour you saw

My previously working build started to fail due to dnf not finding a match for packages from additional repos.
Checking with `mkosi summary` it reports `Package Manager Trees: none` instead of `.../mkosi.images/system/mkosi.skeleton` like it did previously.

Moving the skeleton tree to the main image, or moving it to `mkosi.pkgmngr/` results in it being correctly picked up.

DaanDeMeyer:
@septatrix I think this is fallout from the config parsing rework. Package manager trees are a univeral property so they're explicitly passed to the subimages which causes the default one to not be picked up even if there are none in the main image. I recommend either of the solutions you already came up with or explicitly configuring `PackageManagerTrees=` in the subimage which should also make it get picked up.

septatrix:
Maybe universal options should not be explicitly passed to subimages if they are none (unless explicitly set to none)? At the very least the current behavior contradicts what is written in the docs.

For the time being I moved it to `mkosi.pkgmngr`

PR #2944, opened 2024-08-06
Don't pass down empty lists to subimages unless explicitly configured

DaanDeMeyer:
This makes sure that subimages use default values for list based
settings unless they were explicitly configured in configuration or
on the command line or have a non-empty default value in the main
image.

Fixes https://github.com/systemd/mkosi/issues/2874

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -251,24 +251,29 @@ def test_parse_config(tmp_path: Path) -> None:
 
     (d / "mkosi.images").mkdir()
 
-    for n in ("one", "two"):
-        (d / "mkosi.images" / f"{n}.conf").write_text(
-            f"""\
-            [Distribution]
-            Repositories=append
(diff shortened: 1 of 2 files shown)


tests/test_config.py
```diff
@@ -297,6 +303,9 @@ def test_parse_config(tmp_path: Path) -> None:
     assert one.image_version == "1.2.3"
     assert two.image_version == "4.5.6"
 
+    # Default values from subimages should be picked up.
+    assert len(two.package_manager_trees) == 1 and two.package_manager_trees[0].source.name == "mkosi.skeleton"
```
behrmann, line 307: This should probably also assert something about `one`
```suggestion
    assert len(one.package_manager_trees == 0 and len(two.package_manager_trees) == 1
    assert two.package_manager_trees[0].source.name == "mkosi.skeleton"
```

PR #2945, opened 2024-08-06
Adds current mkosi PID to qemu JSON state file name

gdonval:
When mkosi crashes, Python does not always clean qemu state file. This patch appends current mkosi's PID to the name, solving this problem and allowing other instances of mkosi to use the same machine name without being confused.

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -702,14 +702,16 @@ def finalize_drive(drive: QemuDrive) -> Iterator[Path]:
 
 @contextlib.contextmanager
 def finalize_state(config: Config, cid: int) -> Iterator[None]:
-    (INVOKING_USER.runtime_dir() / "machine").mkdir(parents=True, exist_ok=True)
+    mkosi_pid = os.getpid()
+    machine_folder = INVOKING_USER.runtime_dir() / "machine"
+    machine_folder.mkdir(parents=True, exist_ok=True)
 
     if INVOKING_USER.is_regular_user():
         os.chown(INVOKING_USER.runtime_dir(), INVOKING_USER.uid, INVOKING_USER.gid)
-        os.chown(INVOKING_USER.runtime_dir() / "machine", INVOKING_USER.uid, INVOKING_USER.gid)
+        os.chown(machine_folder, INVOKING_USER.uid, INVOKING_USER.gid)
 
-    with flock(INVOKING_USER.runtime_dir() / "machine"):
-        if (p := INVOKING_USER.runtime_dir() / "machine" / f"{config.machine_or_name()}.json").exists():
+    with flock(machine_folder):
+        if (p := machine_folder / f"{config.machine_or_name()}-{mkosi_pid:d}.json").exists():
             die(f"Another virtual machine named {config.machine_or_name()} is already running",
                 hint="Use --machine to specify a different virtual machine name")
 
@@ -731,7 +733,7 @@ def finalize_state(config: Config, cid: int) -> Iterator[None]:
     try:
         yield
     finally:
-        with flock(INVOKING_USER.runtime_dir() / "machine"):
+        with flock(machine_folder):
             p.unlink(missing_ok=True)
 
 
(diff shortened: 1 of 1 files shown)


mkosi/qemu.py
```diff
@@ -702,14 +702,16 @@ def finalize_drive(drive: QemuDrive) -> Iterator[Path]:
...
 
     if INVOKING_USER.is_regular_user():
         os.chown(INVOKING_USER.runtime_dir(), INVOKING_USER.uid, INVOKING_USER.gid)
-        os.chown(INVOKING_USER.runtime_dir() / "machine", INVOKING_USER.uid, INVOKING_USER.gid)
+        os.chown(machine_folder, INVOKING_USER.uid, INVOKING_USER.gid)
 
-    with flock(INVOKING_USER.runtime_dir() / "machine"):
-        if (p := INVOKING_USER.runtime_dir() / "machine" / f"{config.machine_or_name()}.json").exists():
+    with flock(machine_folder):
+        if (p := machine_folder / f"{config.machine_or_name()}-{mkosi_pid:d}.json").exists():
```
behrmann, line 714: The `:d` is not necessary, numbers are by default formatted as decimal numbers. 
```suggestion
        if (p := machine_folder / f"{config.machine_or_name()}-{mkosi_pid}.json").exists():
```

gdonval: Locking is done on the machine folder (typically `/run/user/$uid/mkosi/machine`). This PR does not change that.

Is there any reason no two `mkosi` instances should not use the same machine name?
If the only reason is that it's later used by `mkosi` to ssh into the VM, then the current approach is fine.

---

> What kind of crashes sidestep the `finally`?

I'm not sure if it's just `ctrl+C` at the wrong time or if it's just when I try booting an image that crashes or if I quit the terminal without doing `ctrl+C` but I can't seem to just be able to `mkosi qemu` for a full day without having to manually clean the machine state directory. It's genuinely happening a lot.

DaanDeMeyer: @gdonval Are you on the latest commit on main? We recently added logic to handle `SIGHUP`, that could maybe fix your issue.

gdonval: No, I'm on 24.3.

At any rate, I'm a bit confused... Is the `if ...: die()` here a simple way to prevent spurious overwrites/deletions or is there any value in making sure machine names are unique across the machine?

If it's the first one, I think my approach is valid. If it's the second one, maybe let's just not use hostname as **default** machine name value?

DaanDeMeyer: > At any rate, I'm a bit confused... Is the if ...: die() here a simple way to prevent spurious overwrites/deletions or is there any value in making sure machine names are unique across the machine?

The `SIGHUP` handling we have on git main will likely reduce the number of times you run into this as the files in /run/ will be properly cleaned up when `mkosi qemu` crashes.

> If it's the first one, I think my approach is valid. If it's the second one, maybe let's just not use hostname as default machine name value?

@gdonval The machine name is used to identify the target for `mkosi ssh` so making it random would make that harder. We can probably pick a better name than `default` though and allow disabling "registering" the machine.

gdonval: 👍 for `SIGHUP`.

@DaanDeMeyer, so

1. The machine name is not a lock, therefore its existence should not make new instances die.
2. That's a `mjosi ssh` concern, therefore it should not affect new `mkosi qemu` calls.

I suggest:

1. either don't write a state file, unless `--machine` is specified;
2. or write machine named with a counter appended (similar to this PR) so that `mkosi qemu` doesn't crash.

In the latter case, `qemu ssh` can look for a precise machine name (with counter) or could look at the highest counter for sma specific machine name stem.

DaanDeMeyer: @gdonval I prefer `mkosi ssh` to work out of the box without having to specify `--machine`. What about using the working directory name as the machine name instead of `default`? That should drastically reduce the number of conflicts.

gdonval: @DaanDeMeyer, you prefer 

> `mkosi ssh` to work without having to specify machine

and suggest using the folder name. 

Am I right to assume you would call `mkosi ssh` only from the current `mkosi` folder then? 

In that case, why don't we write a softlink reference somewhere appropriate in the current `mkosi` project folder? State would still be written centrall (albeit with a convenient distinguisher) and `mkosi ssh` would follow the local link to that folder's default session.

DaanDeMeyer:
I figure we shouldn't touch this logic too much and instead just start registering the virtual machines with machined via the new varlink API instead once that's widely available (systemd 256). So closing this PR

PR #2946, opened 2024-08-08
os-release fixes for Debian and derivatives

bluca:


mkosi/__init__.py
```diff
@@ -258,6 +262,13 @@ def configure_os_release(context: Context) -> None:
                 elif context.config.image_version and line.startswith("IMAGE_VERSION="):
                     new.write(f'IMAGE_VERSION="{context.config.image_version}"\n')
                     image_version_written = True
+                elif (
+                    context.config.distribution == Distribution.debian and
+                    context.config.release in ("unstable", "sid") and
+                    line.startswith("VERSION_CODENAME=")
+                ):
+                    # Workaround for Debian brokeness: unstable sets VERSION_CODENAME=<testing codename>
+                    new.write('VERSION_CODENAME=sid\n')
```
DaanDeMeyer, line 271: Please move this to the debian installer class `install()` function. Also add a link to upstream bug report please, I don't like carrying distribution workarounds forever.

mkosi/__init__.py
```diff
@@ -278,6 +278,22 @@ def configure_os_release(context: Context) -> None:
...
+                "--quiet",
+                "--root",
+                context.root,
+                "--local",
+                "--add",
+                "--rename",
+                "--divert",
+                f"/{candidate}.mkosi",
+                f"/{candidate}"
+            ], stderr=subprocess.DEVNULL)
```
DaanDeMeyer, line 295: We've never pretended that running a package manager on mkosi produced images was going to do the right thing in the first place. If you put `IMAGE_ID` and `IMAGE_VERSION` in /etc/os-release, I think it's safe to assume we're doing a golden image that won't be updated via package manager anymore.

DaanDeMeyer (changes requested):

mkosi/__init__.py
bluca: Ok, but I'll need to completely duplicate the function then?

DaanDeMeyer: Yeah that's fine, just duplicate it but without all the image ID stuff

bluca: Well it would need that too? Because of the diversion

mkosi/__init__.py
DaanDeMeyer: Is dpkg going to overwrite /etc/os-release? If not can't we rework the os-release stuff to also replace /etc/os-release so we don't have to do the dpkg-divert stuff?

bluca: Yes, as it's a symlink in a package, so it would be overwritten. And also we could be building a usr-only image, but as above we can't know here, as it's only specified in repart.conf, so we have to change it in /usr/

DaanDeMeyer: I'm suggesting we change both /etc/os-release (replace symlink with our own file) and /usr/lib/os-release. Then on update /usr/lib/os-release gets overwritten but /etc/os-release won't be overwritten. If dpkg overwrites /etc/os-release on updates that's a bug to fix in debian

bluca: But what's the point of doing that? You need a divert in place to avoid it being replaced anyway in either case

behrmann: Wow, I didn't want to believe this, but just gave it a try and `/etc/os-release` is clobbered when `base-files` is installed. That behaviour is a bug that should be fixed, since that's config that is admin territory.

DaanDeMeyer: > But what's the point of doing that? You need a divert in place to avoid it being replaced anyway in either case

Policy is that /etc/os-release should always be preferred over /usr/lib/os-release, so there's no need for the divert if we also replace /etc/os-release since it won't matter what the contents of /usr/lib/os-release are.

Yes this doesn't handle the case where you're doing upgrades on a /usr only image but I don't think we need to support that use case. Either you do /usr only golden image or you do regular image with /etc included and this approach covers both those use cases.

bluca: > > But what's the point of doing that? You need a divert in place to avoid it being replaced anyway in either case
> 
> Policy is that /etc/os-release should always be preferred over /usr/lib/os-release, so there's no need for the divert if we also replace /etc/os-release since it won't matter what the contents of /usr/lib/os-release are.
> 
> Yes this doesn't handle the case where you're doing upgrades on a /usr only image but I don't think we need to support that use case. Either you do /usr only golden image or you do regular image with /etc included and this approach covers both those use cases.

ok that's probably good enough

mkosi/__init__.py
bluca: Is the cache discarded if the image id or version change? Because by moving it there, it will only run when the cache is discarded, as opposed to now where it always runs

bluca: looks like it is

mkosi/distributions/debian.py
```diff
@@ -190,6 +191,68 @@ def install(cls, context: Context) -> None:
...
+                    "--quiet",
+                    "--root",
+                    context.root,
+                    "--local",
+                    "--add",
+                    "--rename",
+                    "--divert",
+                    f"/{candidate}.mkosi",
+                    f"/{candidate}"
+                ], stderr=subprocess.DEVNULL)
```
DaanDeMeyer, line 252: This needs to run within a sandbox like all the other commands. Please specify a sandbox with `context.root` mounted to `/buildroot` and then have `dpkg-divert` operate on that.

DaanDeMeyer: Please also open a bug for /etc/os-release being overridden by base-files updates and then link it here (I don't care if it gets immediately closed, it'll serve as documentation that all we're doing here is working around's debian's brokenness.

mkosi/distributions/debian.py
```diff
@@ -190,6 +191,68 @@ def install(cls, context: Context) -> None:
...
+                run([
+                    "dpkg-divert",
+                    "--quiet",
+                    "--root",
+                    context.root,
+                    "--local",
+                    "--add",
+                    "--rename",
+                    "--divert",
+                    f"/{candidate}.mkosi",
```
DaanDeMeyer, line 250: Let's use .divert or so instead of .mkosi

mkosi/distributions/debian.py
```diff
@@ -190,6 +191,68 @@ def install(cls, context: Context) -> None:
...
+                        # Workaround for Debian brokeness: unstable sets VERSION_CODENAME=<testing codename>
+                        new.write('VERSION_CODENAME=sid\n')
+                    else:
+                        new.write(line)
+
+                # append if they were missing
+                if context.config.image_id and not image_id_written:
+                    new.write(f'IMAGE_ID="{context.config.image_id}"\n')
+                if context.config.image_version and not image_version_written:
+                    new.write(f'IMAGE_VERSION="{context.config.image_version}"\n')
```
DaanDeMeyer, line 233: So why can't we just fix up the `VERSION_CODENAME` here and set up the divert and then leave the fixing of `IMAGE_ID` and `IMAGE_VERSION` to the generic `configure_os_release()` method? I don't see why we have to do the `IMAGE_ID` stuff here as well.

DaanDeMeyer (changes requested):

DaanDeMeyer (approved):

PR #2947, opened 2024-08-08
debian: Fix up os-release for unstable/sid builds

DaanDeMeyer:
The version codename for unstable/sid builds is indistinguishable from
testing. Let's make sure we fix that up ourselves so that unstable image
builds can be properly distinguished from testing builds.

### DIFF
--- a/mkosi/distributions/debian.py
+++ b/mkosi/distributions/debian.py
@@ -190,6 +190,8 @@ class Installer(DistributionInstaller):
 
         cls.install_packages(context, [Path(deb).name.partition("_")[0].removesuffix(".deb") for deb in essential])
 
+        fixup_os_release(context)
+
     @classmethod
     def install_packages(cls, context: Context, packages: Sequence[str], apivfs: bool = True) -> None:
         # Debian policy is to start daemons by default. The policy-rc.d script can be used choose which ones to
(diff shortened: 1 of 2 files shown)


mkosi/distributions/debian.py
```diff
@@ -253,3 +255,48 @@ def install_apt_sources(context: Context, repos: Iterable[AptRepository]) -> Non
...
+        if candidate == "etc/os-release":
+            run([
+                "dpkg-divert",
+                "--quiet",
+                "--root=/buildroot",
+                "--local",
+                "--add",
+                "--rename",
+                "--divert",
+                f"/{candidate}.dpkg",
```
bluca, line 298: this is usually suffixed to indicate who changed it, but it doesn't matter too much

mkosi/distributions/debian.py
```diff
@@ -253,3 +255,48 @@ def install_apt_sources(context: Context, repos: Iterable[AptRepository]) -> Non
         with sources.open("w") as f:
             for repo in repos:
                 f.write(str(repo))
+
+
+def fixup_os_release(context: Context) -> None:
+    if context.config.release not in ("unstable", "sid"):
+        return
+
```
bluca, line 263: this should skip overlays/confext/sysext too

bluca (approved):

PR #2954, opened 2024-08-14
Add executable `mkosi.version` support for generating the version dynamically

### LINKED ISSUE #2915: Add an executable version of `mkosi.version`
Originally proposed by @septatrix in https://github.com/systemd/mkosi/pull/2905#issuecomment-2251456448

This is effectivly a version of `mkosi.version` that is executed and its output on `stdout` is used as `ImageVersion=`, instead of using the contents of the file for the version.

This would for example allow writing a script that roughly contains `exec git describe --tags --dirty --broken` to use a versioning scheme that is defined by `git tag`, without needing to run `mkosi bump` (with the linked PR) to update the version file before every build.

NekkoDroid:
The script is executed unsandboxed (not a fan of this, but unsure how this should be done before any configuration even exists) and reads `stdout` to figure out the version.

First commit allows more configurable handling of "standard config paths". Maybe this change should also be extended to `mkosi.rootpw` to allow generating it dynamically?

Closes #2915

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -1142,6 +1142,21 @@ def config_parse_minimum_version(value: Optional[str], old: Optional[GenericVers
     return max(old, new)
 
 
+def file_run_or_read(file: Path) -> str:
+    "Run the specified file and capture its output if it's executable, else read file contents"
+
+    if os.access(file, os.X_OK):
+        return run([file.absolute()], stdout=subprocess.PIPE, env=os.environ).stdout
+
+    content = file.read_text()
+
+    if content.startswith("#!/"):
+        die(f"{file} starts with a shebang ({content.splitlines()[0]})",
+            hint="This file should be executable")
+
+    return content
+
+
 @dataclasses.dataclass(frozen=True)
 class KeySource:
     class Type(StrEnum):
@@ -3677,7 +3692,7 @@ class ParseContext:
                             self.config,
                             s.dest,
                             s.parse(
-                                extra.read_text().rstrip("\n") if s.path_read_text else f,
+                                file_run_or_read(extra).rstrip("\n") if s.path_read_text else f,
                                 getattr(self.config, s.dest, None)
                             ),
                         )
--- a/mkosi/resources/mkosi.md
+++ b/mkosi/resources/mkosi.md
@@ -620,13 +620,14 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 `ImageVersion=`, `--image-version=`
 :   Configure the image version. This accepts any string, but it is
     recommended to specify a series of dot separated components. The
-    version may also be configured in a file `mkosi.version` in which
-    case it may be conveniently managed via the `bump` verb or the
-    `--auto-bump` option. When specified the image version is included
-    in the default output file name, i.e. instead of `image.raw` the
-    default will be `image_0.1.raw` for version `0.1` of the image, and
-    similar. The version is also passed via the `$IMAGE_VERSION` to any
-    build scripts invoked (which may be useful to patch it into
+    version may also be configured by reading a `mkosi.version` file (in
+    which case it may be conveniently managed via the `bump` verb or the
+    `--auto-bump` option) or by reading stdout if it is executable (see
+    the **Scripts** section below). When specified the image version is
+    included in the default output file name, i.e. instead of `image.raw`
+    the default will be `image_0.1.raw` for version `0.1` of the image,
+    and similar. The version is also passed via the `$IMAGE_VERSION` to
+    any build scripts invoked (which may be useful to patch it into
     `/usr/lib/os-release` or similar, in particular the `IMAGE_VERSION=`
     field of it).
 
@@ -1137,9 +1138,10 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 
 `RootPassword=`, `--root-password=`,
 :   Set the system root password. If this option is not used, but a `mkosi.rootpw` file is found in the local
-    directory, the password is automatically read from it. If the password starts with `hashed:`, it is treated
-    as an already hashed root password. The root password is also stored in `/usr/lib/credstore` under the
-    appropriate systemd credential so that it applies even if only `/usr` is shipped in the image. To create
+    directory, the password is automatically read from it or if the file is executable it is run as a script
+    and stdout is read instead (see the **Scripts** section below). If the password starts with `hashed:`, it is
+    treated as an already hashed root password. The root password is also stored in `/usr/lib/credstore` under
+    the appropriate systemd credential so that it applies even if only `/usr` is shipped in the image. To create
     an unlocked account without any password use `hashed:` without a hash.
 
 `Autologin=`, `--autologin`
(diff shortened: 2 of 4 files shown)


behrmann:
Having both `mkosi.version` and `mkosi.version.run` leaves the path open to reusing the former as input for the latter, as I described earlier, but if we decide we don't need that, this could be conveniently folded into one and use `mkosi.version` as a script if it's executable and as the regular `mkosi.version` otherwise.

> The script is executed unsandboxed (not a fan of this, but unsure how this should be done before any configuration even exists) and reads stdout to figure out the version.

Maybe that's not necessary? Could we run it just before the configure script?

behrmann:
> I theoretically might be possible, just it would be a bit bigger of a change since ImageVersion= is inherited to the subimage configs, which can override it if they want to and to my knowledge the %v specifier is expanded also during config parsing.

That's true, though thinking about it, that's a problem we already have with configure scripts, since they could potentially change the version. That's probably not a solvable problem, since expansions can't be unmade.

So, yes, for least surprise having the version as early as possible seems reasonable.

NekkoDroid:
Pushed the cleanup commit from #2905 since I had to add the error handling for when `mkosi.version` is executable anyway :)

tests/test_config.py
```diff
@@ -1275,3 +1275,15 @@ def test_environment(tmp_path: Path) -> None:
 
         assert sub.environment["PassThisEnv"] == "abc"
         assert "TestValue2" not in sub.environment
+
+
+def test_mkosi_version_run(tmp_path: Path) -> None:
+    d = tmp_path
+
+    version = d / "mkosi.version"
+    version.write_text("#!/bin/sh\necho '1.2.3'")
```
behrmann, line 1284: Nit:
```suggestion
    version.write_text("#!/bin/sh\necho '1.2.3'\n")
```

mkosi/config.py
```diff
@@ -1160,6 +1161,27 @@ def config_parse_minimum_version(value: Optional[str], old: Optional[GenericVers
...
+def config_path_read(file: Path) -> str:
+    return file.read_text().rstrip("\n")
+
+
+def config_path_run_or_read(file: Path) -> str:
+    "Run the specified file and capture its output if it's executable, else read file contents"
+    if os.access(file, os.X_OK):
+        return config_path_run(file)
+    else:
+        return config_path_read(file)
```
behrmann, line 1182: Nit:
```suggestion
    
    return config_path_read(file)
```

Also, this is in theory a backwards incompatible change, but I'm not sure we should read the whole file here. Having more than a single line in a version doesn't make sense conceptually. My point is, that it's easy to forget to make a file executable and ending up with the text of a script wouldn't be helpful, instead we should die if the text in this case starts with `#!/`, since neither character is allowed in the version format spec and I'm not aware it is used in any version string. If people really want that a multiline script as a version, they can write the script and just add a `head -n -1 $0` at the end.

At the very least we should log a warning.

NekkoDroid:
The CI failures seem unrelated?

behrmann:
I'm travelling over the weekend. Will have a closer look when I have time, Monday at the latest.

The CI failures for Fedora at least are unrelated. We have issues with the Fedora keys about once every release. Haven't looked at the Arch failures.

mkosi/config.py
```diff
@@ -1160,6 +1161,48 @@ def config_parse_minimum_version(value: Optional[str], old: Optional[GenericVers
...
+def config_make_path_handler(handler: ConfigPathCallback,
+                             strip_newlines: bool = True,
+                             allow_multiline: bool = True) ->ConfigPathCallback:
+    def path_handler(file: Path) -> str:
+        content = handler(file)
+
+        if strip_newlines:
+            content = content.rstrip("\n")
+
+        if not allow_multiline and content.count("\n") > 0:
```
behrmann, line 1198: We're only interested in the first one
```suggestion
        if not allow_multiline and "\n" in content:
```

mkosi/config.py
```diff
@@ -1160,6 +1161,48 @@ def config_parse_minimum_version(value: Optional[str], old: Optional[GenericVers
...
+def config_path_read(file: Path) -> str:
+    return file.read_text()
+
+
+def config_path_run_or_read(file: Path) -> str:
+    "Run the specified file and capture its output if it's executable, else read file contents"
+    if os.access(file, os.X_OK):
+        return config_path_run(file)
+
+    content = config_path_read(file)
```
behrmann, line 1180: Let's not make this overly complicated
```suggestion
    content = file.read_text()
```

mkosi/config.py
```diff
@@ -1160,6 +1161,48 @@ def config_parse_minimum_version(value: Optional[str], old: Optional[GenericVers
...
+    if content.startswith("#!/"):
+        die(f"{file} starts with a shebang ({content.splitlines()[0]})",
+            hint="This file should be executable")
+
+    return content
+
+
+def config_make_path_handler(handler: ConfigPathCallback,
+                             strip_newlines: bool = True,
+                             allow_multiline: bool = True) ->ConfigPathCallback:
```
behrmann, line 1191:
```suggestion
                             allow_multiline: bool = True) -> ConfigPathCallback:
```

tests/test_config.py
```diff
@@ -1275,3 +1275,15 @@ def test_environment(tmp_path: Path) -> None:
...
         assert sub.environment["PassThisEnv"] == "abc"
         assert "TestValue2" not in sub.environment
+
+
+def test_mkosi_version_run(tmp_path: Path) -> None:
+    d = tmp_path
+
+    version = d / "mkosi.version"
+    version.write_text("#!/bin/sh\necho '1.2.3'\n")
+    version.chmod(0o755)
```
behrmann, line 1285: Can you add another assert where this file is made non-executable to check that the error is raised correctly?

behrmann (commented):
Haven't tested it yet, but looks reasonable to me. A few last nits

mkosi/config.py
```diff
@@ -2641,7 +2688,11 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
         section="Content",
         parse=config_parse_root_password,
         paths=("mkosi.rootpw",),
-        path_read_text=True,
+        path_handler=config_make_path_handler(
+            handler=config_path_read,
```
DaanDeMeyer, line 2692: Let's make this executable as well, seems like it could be useful as well. Frankly, I wouldn't even allow configuring this. If we're reading the contents immediately during config parsing, then I think it'll always make sense to execute the file if it is executable to get the contents that way. So I'd remove the concept of a handler and always do the equivalent of run_or_read().

mkosi/config.py
```diff
@@ -1160,6 +1161,48 @@ def config_parse_minimum_version(value: Optional[str], old: Optional[GenericVers
...
+    if content.startswith("#!/"):
+        die(f"{file} starts with a shebang ({content.splitlines()[0]})",
+            hint="This file should be executable")
+
+    return content
+
+
+def config_make_path_handler(handler: ConfigPathCallback,
+                             strip_newlines: bool = True,
+                             allow_multiline: bool = True) -> ConfigPathCallback:
```
DaanDeMeyer, line 1191: Let's add these options when we actually need them. Since the values are always the same at the moment let's just drop them for now.

DaanDeMeyer (changes requested):

mkosi/resources/mkosi.md
```diff
@@ -2076,6 +2078,18 @@ current working directory. The following scripts are supported:
...
 
+* If **`mkosi.version`** exists and is executable, it is run during
+  configuration parsing and populates `ImageVersion=` with the output on stdout.
+  This can be used for external version tracking, e.g. with `git describe` or
+  `date '+%Y-%m-%d'` Note that this script is executed on the host system
+  without any sandboxing.
+
+* If **`mkosi.rootpw`** exists and is executable, it is run during
+  configuration parsing and populated `RootPassword=` with the output
+  on stdout. This can be used to randomily generate a password and can
```
behrmann, line 2089:
```suggestion
  on stdout. This can be used to randomly generate a password and can
```
Though this is now possible I'm not sure we should advertise throwing around cleartext password.

behrmann (approved):

DaanDeMeyer (approved):

septatrix:
Thanks for implementing this Nekko :tada:

PR #2956, opened 2024-08-16
Introduce mkosi-sandbox and stop using subuids for image builds

DaanDeMeyer:
Over the last years, we've accumulated a rather nasty set of workarounds
for various issues in bubblewrap:

- We contributed setpgid to util-linux and use it if available because
  bubblewrap does not support making its child process the foreground
  process.
- We added the innerpid logic to run() because bubblewrap does not forward
  signals to the separate child process it runs in the sandbox which meant
  they were getting SIGKILLed when we killed bubblewrap, preventing proper
  cleanup from happening.
- bubblewrap does not provide a proper way to detect whether the command
  was found in the sandbox or not, which meant we had to execute command -v
  within the sandbox separately to check whether the command exists or not.
- We had to add extra logic to make sure / was a mount in the initramfs to
  allow running mkosi in the initramfs as bubblewrap does not fall back to
  MS_MOVE if pivot_root() doesn't work.
- We had to stitch together shell invocations after bubblewrap but before
  executing the actual command we want to run to make sure directories had
  the correct mode as bubblewrap creates everything with mode 0700 which was
  too restrictive in many cases for us. This was fixed with new --perms and
  --chmod options in bubblewrap 0.5 but we had to keep compat with 0.4
  because that's what's shipped in CentOS Stream 9.
- We had to figure out a shell hack to do overlayfs mounts as these are not
  supported by bubblewrap (even though a PR for the feature has been open for
  years).
- We had to introduce a Mount struct to pass around mounts so we could deduplicate
  and sort them before passing them to bubblewrap as bubblewrap did not do this
  itself.
- Debugging all the above was made all the harder by the fact that bubblewrap's
  source code is full of tech debt from its history of being a setuid tool
  instead of using user namespaces. Getting any fixes into upstream is almost
  impossible as the tool is practically unmaintained.

Aside from bubblewrap, our other source of troubles has been newuidmap/newgidmap.
Running as a user within the subuid range configured in /etc/sub{u,g}id has
meant we're constantly fixing ownership and permissions issues where stuff needs
to be chowned and chmodded everywhere to make sure the current user and the
subuid user can access the proper files. Another unfortunate side effect is that
users end up with many files owned by the subuid root user in their home
directories when building images with mkosi;

Let's fix all these issues at once by getting rid of bubblewrap and
newuidmap/newgidmap.

bubblewrap is replaced with a new tool mkosi-sandbox. It looks and behaves a
lot like bubblewrap, except it's much less code and much more flexible to fit
our needs, allowing us to get rid of all the hacks we've built up over the years to
work around issues that didn't get fixed in bubblewrap.

To get rid of newuidmap/newgidmap, a rework of our user namespacing was needed.
The need to use newuidmap/newgidmap came from the assumption that we need a full
65k subuid range to do unprivileged image builds, as distributions ship packages
containing files and directories that are not owned by the root user. After some
investigation, it turns out that there's very few files and directories not owned
by root in distribution packages if you ignore /var. If we could temporarily
ignore the ownership on these files and directories until we can get distributions
to only ship root owned files in /usr and /etc of their packages, we could simply
map the current user to root in a user namespace and get rid of the subuid range
completely.

Turns out that's possible with a seccomp filter. seccomp allows you to make all
chown() syscalls succeed without actually doing anything. The files and directories
end up owned by the root user instead. If we assume this is OK and are OK with
instructing users to use tmpfiles to fix up the permissions on first boot if needed,
a seccomp filter like this is sufficient to allow us to get rid of doing image
builds within a subuid user namespace.

It turns out we can go one step further. It turns out that for the majority of
the image build, one doesn't actually need to be the root user. Only package
managers and systemd-repart need the current user to be mapped to root to do their
job correctly. The reason we did the entire build mapped to root until now was
that we need to do a few mounts as part of the image build process and for now
I was under the assumption that you needed to be root for that. It turns out that
when you unshare a user namespace, you get a full set of capabilities regardless
of whether you're root or some other uid in the user namespace. The only difference
is that when you exec a subprocess as root, the capabilities aren't lost, whereas
they are when you exec a subprocess as a non-root user. This can be avoided by
adding the capabilities of the non-root user to the inheritable and ambient set.
Once that's done, any subprocess exec'd by a non-root user in the user namespace
can mount as many bind and overlay mounts as they can think of.

The above allows us to run most of the image build under the current user uid
instead of root, only switching to root when running package managers, invoking
systemd-repart or systemd-tmpfiles, or when chroot-ing into the image. This allows
us to get rid of various hacks we had to look up the proper user name or home
directory.

Specifically, we can get rid of the following:

- mkosi-as-caller can become a noop since we now by default run the build as the
  caller.
- Lots of chmod()'s and chown()'s can be removed
- All uses of INVOKING_USER.uid/gid can be removed, and most can be replaced with
  simple os.getuid()/os.getgid()
- We can use /etc/passwd and /etc/group from the host instead of building our own
- We can get rid of the Acl= option as the user will now be able to remove (almost)
  all files written by mkosi.
- We don't have to rchown the package manager cache directory anymore after each
  build. Root user builds will now use the system cache instead of the per user
  cache.
- We can get rid of the Mount struct as mkosi-sandbox dedups and sorts operations
  itself.

One thing to note is that if we're invoked as root, none of the seccomp or capabilities
stuff applies and it is all skipped as it's not required in that case. This means that
when building as root it's still possible to have more than one user in the generated
image unlike when building unprivileged. Also note that users can still be added to
/etc/passwd and such, they just can't own any files or directories in the image itself
until the image is booted.


NekkoDroid:
I just attempted to build my image with this and immediatly ran into a problem:

```
‣ Copying in package manager file trees…
‣ Syncing package manager metadata for image-initrd image
  File "/home/nekko/.local/src/mkosi/mkosi/cage.py", line 543, in <module>
  File "/home/nekko/.local/src/mkosi/mkosi/cage.py", line 435, in main
  File "/home/nekko/.local/src/mkosi/mkosi/cage.py", line 188, in mount_rbind
  File "/home/nekko/.local/src/mkosi/mkosi/cage.py", line 82, in oserror
FileNotFoundError: [Errno 2] No such file or directory: '/mnt/wsl/resolv.conf'
‣ "pacman --root=/buildroot --logfile=/dev/null --dbpath=/var/lib/pacman --cachedir=/var/cache/pacman/mkosi --cachedir=/var/cache/pacman/pkg --hookdir=/buildroot/etc/pacman.d/hooks --arch x86_64 --color auto --noconfirm --sync --refresh" returned non-zero exit code 1.
```

Seems it doesn't copy/use the target of `/etc/resolve.conf`, which I assume might also fail when using sd-resolved.

DaanDeMeyer:
@NekkoDroid Should be fixed, please try again.

NekkoDroid:
Indeed the `resolve.conf` issue is gone. After I had some issues with pacman failing to lock the database, but I managed to fix it by using `mkosi -fff`.

Now I was able to build the image successfully.

septatrix:
I like the idea of not having to deal with bubblewrap but I am not so sure about dropping uid mapping. I fear that most distros are not there yet and for any larger image this would mean having to create dozens of sd-tmpfile configs (and likely still missing a few leading to hard to troubleshoot bugs). Based on the motivation these seem like independent things so maybe it would be better to first adopt mkosi-cage and leave the uid mapping in place until a later time

DaanDeMeyer:
@septatrix You can still run something like `unshare --map-auto --map-current-user --setuid 0 --setgid 0 mkosi --package-cache-dir mkosi.pkgcache -f` if you want to use subuids/subgids. Doing it within mkosi causes too much complexity and I'm sure we're better off keeping this outside of mkosi.

DaanDeMeyer:
> @septatrix You can still run something like `unshare --map-auto --map-current-user --setuid 0 --setgid 0 mkosi --package-cache-dir mkosi.pkgcache -f` if you want to use subuids/subgids. Doing it within mkosi causes too much complexity and I'm sure we're better off keeping this outside of mkosi.

I added a FAQ entry now to explain what's going on.

mkosi/cage.py
```diff
@@ -0,0 +1,664 @@
...
+libc.umount2.argtypes = (ctypes.c_char_p, ctypes.c_int)
+libc.capget.argtypes = (ctypes.c_void_p, ctypes.c_void_p)
+libc.capset.argtypes = (ctypes.c_void_p, ctypes.c_void_p)
+
+
+def oserror(arg: str = "") -> None:
+    if arg:
+        raise OSError(ctypes.get_errno(), os.strerror(ctypes.get_errno()), arg)
+    else:
+        raise OSError(ctypes.get_errno(), os.strerror(ctypes.get_errno()))
```
behrmann, line 90: Let's stay with the variable name that `OSError` uses and since the additional arguments default to `None`, we can shorten this a bit.
```suggestion
def oserror(filename: str = "") -> None:
    raise OSError(ctypes.get_errno(), os.strerror(ctypes.get_errno()), filename or None)    
```

mkosi/cage.py
```diff
@@ -0,0 +1,664 @@
...
+itself. To keep the runtime of this script to a minimum, please don't import any extra modules if it can be avoided.
+"""
+
+import ctypes
+import encodings.ascii  # noqa: F401 (loaded lazily by open() which happens too late)
+import os
+import sys
+import warnings  # noqa: F401 (loaded lazily by os.execvp() which happens too late)
+
+AT_EMPTY_PATH = 0x1000
```
behrmann, line 15:
```suggestion
# The following constants are taken from the Linux kernel headers
AT_EMPTY_PATH = 0x1000
```

mkosi/cage.py
```diff
@@ -0,0 +1,664 @@
...
+MS_SLAVE = 1 << 19
+NR_mount_setattr = 442
+NR_move_mount = 429
+NR_open_tree = 428
+OPEN_TREE_CLOEXEC = os.O_CLOEXEC
+OPEN_TREE_CLONE = 1
+PR_CAP_AMBIENT = 47
+PR_CAP_AMBIENT_RAISE = 2
+SCMP_ALLOW = 0x7FFF0000
+SCMP_ERRNO = 0x00050000
```
behrmann, line 50: That makes them much easier to grep for
```suggestion
SCMP_RET_ALLOW = 0x7FFF0000
SCMP_RET_ERRNO = 0x00050000
```

mkosi/cage.py
```diff
@@ -0,0 +1,664 @@
...
+
+class mount_attr(ctypes.Structure):
+    _fields_ = [
+        ("attr_set", ctypes.c_uint64),
+        ("attr_clr", ctypes.c_uint64),
+        ("propagation", ctypes.c_uint64),
+        ("userns_fd", ctypes.c_uint64),
+    ]
+
+class cap_user_header_t(ctypes.Structure):
```
behrmann, line 60: Let's add a comment with the original name, so this is easy to look for
```suggestion
class cap_user_header_t(ctypes.Structure):
    # __user_cap_header_struct
```

mkosi/cage.py
```diff
@@ -0,0 +1,664 @@
...
+    _fields_ = [
+        ("version", ctypes.c_uint32),
+        ("pid", ctypes.c_int),
+    ]
+
+class cap_user_data_t(ctypes.Structure):
```
behrmann, line 66:
```suggestion
class cap_user_data_t(ctypes.Structure):
    # __user_cap_data_struct
```

mkosi/cage.py
```diff
@@ -0,0 +1,669 @@
...
+MS_SLAVE = 1 << 19
+NR_mount_setattr = 442
+NR_move_mount = 429
+NR_open_tree = 428
+OPEN_TREE_CLOEXEC = os.O_CLOEXEC
+OPEN_TREE_CLONE = 1
+PR_CAP_AMBIENT = 47
+PR_CAP_AMBIENT_RAISE = 2
+SCMP_ACT_ALLOW = 0x7FFF0000
+SCMP_ACT_ERRNO = 0x00050000
```
behrmann, line 51: Using the `ACT` names it would make sense to change this slightly
```suggestion

# These definitions are taken from the libseccomp headers
SCMP_ACT_ALLOW = 0x7FFF0000
SCMP_ACT_ERRNO = 0x00050000
```

mkosi/cage.py
```diff
@@ -0,0 +1,669 @@
...
+    libseccomp.seccomp_release.restype = None
+    libseccomp.seccomp_syscall_resolve_name.argtypes = (ctypes.c_char_p,)
+    libseccomp.seccomp_rule_add_exact.argtypes = (ctypes.c_void_p, ctypes.c_uint32, ctypes.c_int, ctypes.c_uint)
+    libseccomp.seccomp_load.argtypes = (ctypes.c_void_p,)
+
+    seccomp = libseccomp.seccomp_init(SCMP_ACT_ALLOW)
+
+    try:
+        for syscall in ("chown", "chown32", "fchown", "fchown32", "fchownat", "lchown", "lchown32"):
+            id = libseccomp.seccomp_syscall_resolve_name(syscall.encode())
```
behrmann, line 193: Nit: sincce it's all static we can avoid the encode.
```suggestion
        for syscall in (b"chown", b"chown32", b"fchown", b"fchown32", b"fchownat", b"lchown", b"lchown32"):
            id = libseccomp.seccomp_syscall_resolve_name(syscall)
```

mkosi/cage.py
```diff
@@ -0,0 +1,667 @@
...
+
+    libseccomp.seccomp_init.argtypes = (ctypes.c_uint32,)
+    libseccomp.seccomp_init.restype = ctypes.c_void_p
+    libseccomp.seccomp_release.argtypes = (ctypes.c_void_p,)
+    libseccomp.seccomp_release.restype = None
+    libseccomp.seccomp_syscall_resolve_name.argtypes = (ctypes.c_char_p,)
+    libseccomp.seccomp_rule_add_exact.argtypes = (ctypes.c_void_p, ctypes.c_uint32, ctypes.c_int, ctypes.c_uint)
+    libseccomp.seccomp_load.argtypes = (ctypes.c_void_p,)
+
+    seccomp = libseccomp.seccomp_init(SCMP_ACT_ALLOW)
```
behrmann, line 187: Just to be safe, we should probably check whether this is not null.

mkosi/cage.py
```diff
@@ -0,0 +1,667 @@
...
+
+    try:
+        attr = mount_attr()
+        attr.attr_set = attrs
+
+        flags = AT_EMPTY_PATH|AT_RECURSIVE
+
+        try:
+            libc.mount_setattr.argtypes = (
+                ctypes.c_int, ctypes.c_char_p, ctypes.c_uint, ctypes.c_void_p, ctypes.c_size_t)
```
behrmann, line 226:
```suggestion
                ctypes.c_int, ctypes.c_char_p, ctypes.c_uint, ctypes.c_void_p, ctypes.c_size_t
            )
```

mkosi/cage.py
```diff
@@ -0,0 +1,667 @@
...
+            r = libc.mount_setattr(fd, b"", flags, ctypes.addressof(attr), MOUNT_ATTR_SIZE_VER0)
+        except AttributeError:
+            libc.syscall.argtypes = (
+                ctypes.c_long, ctypes.c_int, ctypes.c_char_p, ctypes.c_uint, ctypes.c_void_p, ctypes.c_size_t)
```
behrmann, line 230:
```suggestion
                ctypes.c_long, ctypes.c_int, ctypes.c_char_p, ctypes.c_uint, ctypes.c_void_p, ctypes.c_size_t
            )
```

mkosi/cage.py
```diff
@@ -0,0 +1,667 @@
...
+
+        if r < 0:
+            oserror(src)
+
+        try:
+            libc.move_mount.argtypes = (ctypes.c_int, ctypes.c_char_p, ctypes.c_int, ctypes.c_char_p, ctypes.c_uint)
+            r = libc.move_mount(fd, b"", AT_FDCWD, dst.encode(), MOVE_MOUNT_F_EMPTY_PATH)
+        except AttributeError:
+            libc.syscall.argtypes = (
+                ctypes.c_long, ctypes.c_int, ctypes.c_char_p, ctypes.c_int, ctypes.c_char_p, ctypes.c_uint)
```
behrmann, line 241:
```suggestion
                ctypes.c_long, ctypes.c_int, ctypes.c_char_p, ctypes.c_int, ctypes.c_char_p, ctypes.c_uint
            )
```

mkosi/cage.py
```diff
@@ -0,0 +1,667 @@
...
+
+
+def acquire_privileges(*, become_root: bool = False) -> bool:
+    if os.getuid() == 0 or (not become_root and have_effective_cap(CAP_SYS_ADMIN)):
+        return False
+
+    become_user(0 if become_root else os.getuid(), 0 if become_root else os.getgid())
+
+    if not become_root:
+        cap_permitted_to_ambient()
```
behrmann, line 311:
```suggestion
    if become_root:
        become_user(0, 0)
    else:
        become_user(os.getuid(), os.getgid())    
        cap_permitted_to_ambient()
```

mkosi/cage.py
```diff
@@ -0,0 +1,667 @@
...
+
+
+def userns_has_single_user() -> bool:
+    if not os.path.exists("/proc/self/uid_map"):
+        return False
+
+    with open("/proc/self/uid_map") as f:
+        lines = f.readlines()
+
+    return len(lines) == 1 and int(lines[0].split()[-1]) == 1
```
behrmann, line 323: Safes one check in the usual case of the file existing
```suggestion
    try:
        with open("/proc/self/uid_map") as f:
            lines = f.readlines()

        return len(lines) == 1 and int(lines[0].split()[-1]) == 1
    except FileNotFoundError:
        return False
```

mkosi/cage.py
```diff
@@ -0,0 +1,667 @@
...
+    if not os.path.exists(src) and operation.endswith("try"):
+        return
+
+    with umask(~0o755):
+        os.makedirs(os.path.dirname(dst), exist_ok=True)
+
+    if not os.path.exists(dst):
+        with umask(~0o644 if os.path.isfile(src) else ~0o755):
+            if os.path.isfile(src):
+                os.close(os.open(dst, os.O_CREAT|os.O_CLOEXEC))
```
behrmann, line 366: Let's hoist the check up a level
```suggestion
        isfile = os.path.isfile(src)
        with umask(~0o644 if isfile else ~0o755):
            if isfile:
                os.close(os.open(dst, os.O_CREAT|os.O_CLOEXEC))
```

mkosi/cage.py
```diff
@@ -0,0 +1,667 @@
...
+
+    return result
+
+
+def do_mount(operation: str, args: tuple[str, ...]) -> None:
+    src, dst = args
+    src = resolve("oldroot", src)
+    dst = resolve("newroot", dst)
+
+    if not os.path.exists(src) and operation.endswith("try"):
```
behrmann, line 357: I don't see it happening, that anohter option will grow that ends in try, but let's maybe add the word separator to be sure.
```suggestion
    if not os.path.exists(src) and operation.endswith("-try"):
```

mkosi/cage.py
```diff
@@ -0,0 +1,667 @@
...
+
+def do_proc(operation: str, args: tuple[str, ...]) -> None:
+    dst = resolve("newroot", args[0])
+    with umask(~0o755):
+        os.makedirs(dst, exist_ok=True)
+
+    mount_rbind("oldroot/proc", dst)
+
+
+ttyname = os.ttyname(2) if os.isatty(2) else None
```
behrmann, line 381: Since this is global "constant", let's maybe move it after the numerical cconstants and
```suggestion
TTYNAME = os.ttyname(2) if os.isatty(2) else None
```

mkosi/sandbox.py
```diff
@@ -196,7 +188,7 @@ def sandbox_cmd(
 
     for d in ("bin", "sbin", "lib", "lib32", "lib64"):
         if (p := tools / d).is_symlink():
-            cmdline += ["--symlink", p.readlink(), Path("/") / p.relative_to(tools)]
+            cmdline += ["--symlink", p.readlink(), Path('/') / p.relative_to(tools)]
```
behrmann, line 191: Nit: The change in quotes is the only change in this line. Changing that back makes the diff nicer
```suggestion
            cmdline += ["--symlink", p.readlink(), Path('/') / p.relative_to(tools)]
```

mkosi/config.py
```diff
@@ -2896,15 +2893,6 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
         parse=config_make_list_parser(delimiter=" "),
         help="Append extra entries to the kernel command line when booting the image",
     ),
-    ConfigSetting(
```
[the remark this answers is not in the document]
DaanDeMeyer: I think this option is rarely used so we should be OK with dropping it.

mkosi/resources/mkosi.md
```diff
@@ -2754,6 +2730,33 @@ Note that the minimum required Python version is 3.9.
...
+  can consider running `mkosi` as root which avoids this problem. Alternatively,
+  if running `mkosi` as root is not desired, you can use
+  `unshare --map-auto --map-current-user --setuid 0 --setgid 0` to become root in
+  a user namespace with more than one user assuming the UID/GID mappings in
+  `/etc/subuid` and `/etc/subgid` are configured correctly. Note that running mkosi
+  as root or with `unshare` means that all output files produced by mkosi will not
+  be owned by your current user anymore.
+
+  Note that for systemd services that need directories in `/var` owned by the service
+  user and group, an alternative to shipping these directories in rpm packages or
```
behrmann, line 2752: Let's drop the RPM there to make this more inclusive
```suggestion
  user and group, an alternative to shipping these directories in packages or
```

mkosi/resources/mkosi.md
```diff
@@ -2754,6 +2730,33 @@ Note that the minimum required Python version is 3.9.
...
+  creating them via systemd-tmpfiles is to use `StateDirectory=`, `CacheDirectory=` or
```
behrmann, line 2753: Creating them via tmpfiles.d config wouldn't be bad as long as the config is shipped in `/usr/` as well,  the problem are postinst scriptlets, so I'd write
```suggestion
  creating them via package postinstallation scripts, is to use `StateDirectory=`, `CacheDirectory=` or
```

mkosi/sandbox/resources/mkosi-sandbox.md
```diff
@@ -0,0 +1,151 @@
...
+`mkosi-sandbox [options…] command [arguments]`
+
+# DESCRIPTION
+
+`mkosi-sandbox` runs the given command in a custom sandbox. The sandbox is configured
+by specifying command line options that configure individual parts of the sandbox.
+
+If no command is specified, `mkosi-sandbox` will start `bash` in the sandbox.
+
+Note that this sandbox is not designed to be a security boundary. It's intended purpose
```
behrmann, line 20:
```suggestion
Note that this sandbox is not designed to be a security boundary. Its intended purpose
```

mkosi/sandbox/resources/mkosi-sandbox.md
```diff
@@ -0,0 +1,151 @@
...
+is to allow running commands in an isolated environment so they are not affected by the
+host system.
+
+# OPTIONS
+
+`--tmpfs`
+:   Takes a single argument. Mounts a new tmpfs in the sandbox at the given path.
```
behrmann, line 27: Not our usual style, but makes the sentences on arity unnecessary and therefore the whole man page if not shorter at least less repetitive and also drives home that `--arg=val` is not supported, since in our usualy style all CLI args are given as `--arg=`
```suggestion
`--tmpfs target`
:   Mounts a new tmpfs in the sandbox at the given path.
```

mkosi/sandbox/resources/mkosi-sandbox.md
```diff
@@ -0,0 +1,151 @@
...
+
+`--dev`
+:   Takes a single argument. Sets up a private /dev in the sandbox at the given path in
```
behrmann, line 30:
```suggestion
:   Takes a single argument. Sets up a private `/dev` in the sandbox at the given path in
```

mkosi/sandbox/resources/mkosi-sandbox.md
```diff
@@ -0,0 +1,151 @@
...
+    the sandbox. This private /dev will only contain the basic device nodes required
```
behrmann, line 31:
```suggestion
    the sandbox. This private `/dev` will only contain the basic device nodes required
```

mkosi/sandbox/resources/mkosi-sandbox.md
```diff
@@ -0,0 +1,151 @@
...
+    for a functioning sandbox (e.g. `/dev/null`) and no actual devices.
+
+`--proc`
+:   Takes a single argument. Mounts /proc from the host at the given path in the sandbox.
```
behrmann, line 35:
```suggestion
:   Takes a single argument. Mounts `/proc` from the host at the given path in the sandbox.
```

mkosi/sandbox/resources/mkosi-sandbox.md
```diff
@@ -0,0 +1,151 @@
...
+
+`--help`, `-h`
+:   Show brief usage information.
+
+# EXAMPLES
+
+Start `bash` in the current working directory in its own network namespace as the current user.
+
+```sh
+mkosi-sandbox --bind / / --chdir $PWD --unshare-net
```
behrmann, line 129:
```suggestion
mkosi-sandbox --bind / / --chdir "$PWD" --unshare-net
```
but probably better
```suggestion
mkosi-sandbox --bind / / --same-dir --unshare-net
```

behrmann (approved):

PR #2963, opened 2024-08-20
Set `DEFAULT_HOSTNAME` in os-release when specified (fixes #2673)

### LINKED ISSUE #2673: `DEFAULT_HOSTNAME=` in os-release is not set when `Hostname=` is set

main

### Used host distribution

Fedora 40

### Used target distribution

Fedora 40

### Linux kernel version used

6.8.7-300.fc40.x86_64

x86_64

On a usr/-only build the hostname reverted to "fedora" after losing a DHCP lease which assigned a hostname instead of the hostname set during image build by mkosi.  See https://github.com/systemd/systemd/issues/32541

As there is only one level of transient hostname "storage" there is no way to get the original hostname back afterwards. Therefore the only fallback possible is to use the default hostname from os-release (or the one compiled during systemd build). mkosi should therefore also update that variable to ensure the expected behavior.

### Used mkosi config

_No response_

### mkosi output

_No response_

septatrix:
When the user specifies a hostname in the mkosi config they intend this to be the hostname which the system should have by default. Setting this in the os-release file ensures that this will also be applied up after a DHCP lease expires and the previous transient hostname was lost (in case no hostname exists under /etc as is the case for a /usr-only system).

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -234,7 +234,7 @@ def check_root_populated(context: Context) -> None:
 
 def configure_os_release(context: Context) -> None:
     """Write IMAGE_ID and IMAGE_VERSION to /usr/lib/os-release in the image."""
-    if not context.config.image_id and not context.config.image_version:
+    if not (context.config.image_id or context.config.image_version or context.config.hostname):
         return
 
     if context.config.overlay or context.config.output_format in (OutputFormat.sysext, OutputFormat.confext):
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -234,21 +234,22 @@ def check_root_populated(context: Context) -> None:
 
 def configure_os_release(context: Context) -> None:
     """Write IMAGE_ID and IMAGE_VERSION to /usr/lib/os-release in the image."""
-    if not context.config.image_id and not context.config.image_version:
+    if not any ([context.config.image_id, context.config.image_version, context.config.hostname]):
```
behrmann, line 237: No space between name and argument
```suggestion
    if not any([context.config.image_id, context.config.image_version, context.config.hostname]):
```

Though I'd prefer
```suggestion
    if not (context.config.image_id or context.config.image_version or context.config.hostname):
```
since it's still shorter and less visual clutter (commas and braces).

DaanDeMeyer (approved):

PR #2964, opened 2024-08-22
Specify --renumber-inodes when calling cpio

DaanDeMeyer:
This helps with making initrds reproducible as the sequence of inodes will always be stable starting from 1.


behrmann:
No harm in setting it, but since `--reproducible` is supposed to set it already, we probably don't need this?

PR #2967, opened 2024-08-22
Two fixes

DaanDeMeyer:


behrmann (approved):
LGTM, but can we find a better name than `make_root`?  There's too many roots flying around. :) How about `make_tree_root`?

PR #2968, opened 2024-08-22
ci: Disable Arch Linux image with Ubuntu tools tree

DaanDeMeyer:
archlinux-keyring in Ubuntu Noble is already too old. Until it's updated, let's disable the CI build.

For the Github Action we add the kernel-utils ppa which does have an updated archlinux-keyring.


behrmann (approved):

DaanDeMeyer:
CI job didn't appear, no need to wait for the rest

PR #2969, opened 2024-08-22
Add missing directory checks in copy_tree()

DaanDeMeyer:


behrmann (approved):
Keeping the difference between iterdir and glob straight is a footgun.

PR #2973, opened 2024-08-23
Rework repository metadata handling

DaanDeMeyer:
- Stop copying repository metadata into the image.

This is too fragile to ever work properly. If the image is ever
used as a base tree, the caller would also need the exact same package
manager configuration for this to be remotely useful, as well as constantly
rebuild the image to keep the repository metadata up to date.

- Stop picking up repository metadata from the image

For the same reason, we can't use repository metadata from any
base trees automatically. It should be explicitly provided by the user
along with the required package manager configuration.

- Share the same repository metadata snapshot between the main and subimage
  builds

Let's ensure that the main image and all subimages are built of the
exact same repository metadata snapshot. Now that we enforce that all
subimages use the same distro, release, architecture, repositories and
everything else that's package manager related, we can use the same
metadata snapshot for every build.

- Rename package_cache_dir in Context to metadata_dir as there's only
metadata in this directory and never any packages.


mkosi/__init__.py
```diff
@@ -4750,108 +4743,126 @@ def run_verb(args: Args, images: Sequence[Config], *, resources: Path) -> None:
...
-                if config.output_format == OutputFormat.none and not config.build_scripts:
-                    return
+    if tools and not (tools.output_dir_or_cwd() / tools.output).exists():
+        with prepend_to_environ_path(tools):
+            check_tools(tools, Verb.build)
+            run_sync(args, tools, resources=resources)
+
+            with tempfile.TemporaryDirectory(
+                dir=tools.workspace_dir_or_default(),
+                prefix="mkosi-metadata-"
```
behrmann, line 4811:
```suggestion
                prefix="mkosi-metadata-",
```

mkosi/__init__.py
```diff
@@ -4644,7 +4616,14 @@ def run_sync(args: Args, config: Config, *, resources: Path) -> None:
...
 
 
-def run_build(args: Args, config: Config, *, resources: Path, package_dir: Optional[Path] = None) -> None:
+def run_build(
+    args: Args,
+    config: Config,
+    *,
+    resources: Path,
+    metadata_dir: Path,
+    package_dir: Optional[Path] = None
```
behrmann, line 4625:
```suggestion
    package_dir: Optional[Path] = None,
```

mkosi/resources/mkosi.md
```diff
@@ -488,30 +488,21 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
-    otherwise, it will use the configuration files from their canonical
-    locations in `/usr` or `/etc` in the package manager trees. For
-    example, it will look for `etc/dnf/dnf.conf` in the package manager
-    trees if `dnf` is used to install packages.
-
-    `SkeletonTrees=` and `PackageManagerTrees=` fulfill similar roles. Use
-    `SkeletonTrees=` if you want the files to be present in the final image. Use
-    `PackageManagerTrees=` if you don't want the files to be present in the final
-    image, e.g. when building an initrd or if you want to refer to paths outside
-    of the image in your repository configuration.
```
behrmann, line 514: With the rename and the removal of this paragraph we should consider scrapping the defaulting to the skeleton tree.

PR #2982, opened 2024-08-27
Use python3 in sandbox if host interpreter is not in /usr

DaanDeMeyer:
We only mount /usr into the sandbox, so if mkosi is invoked from a venv we'll fail to execute the apivfs script or chroot script in the sandbox as it will try to use an interpreter that isn't available.

Let's check if the used interpreter is relative to /usr and only use it to execute the chroot and apivfs scripts in the sandbox if it is.

### DIFF
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -559,8 +559,9 @@ def apivfs_options(*, root: Path = Path("/buildroot")) -> list[PathString]:
 
 
 def apivfs_script_cmd(*, tools: bool, options: Sequence[PathString] = ()) -> list[PathString]:
+    exe = Path(sys.executable)
     return [
-        "python3" if tools else sys.executable, "-SI", "/sandbox.py",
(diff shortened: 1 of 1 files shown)


evverx:
@DaanDeMeyer looks like it failed somewhere down the road
```
/var/tmp/repart-5ZRCxK successfully formatted as btrfs (label "root-x86-64", uuid 8d2a3dcb-0df0-4303-8367-4693165466e9)
Successfully formatted future partition 1.
Adding new partition 1 to partition table.
Writing new partition table.
All done.
‣ /usr/lib/firmware/{amd-ucode,intel-ucode} not found, not adding microcode
‣  Running modinfo to fetch kernel module dependencies
‣  Calculating required kernel modules and firmware
‣  Creating cpio archive /home/vagrant/.cache/mkosi/mkosi-workspace-0yjjrb93/kernel-modules-6.10.6-200.fc40.x86_64.initrd…
‣  Generating unified kernel image for kernel version 6.10.6-200.fc40.x86_64
‣ /home/vagrant/VENV/bin/python3 not found.
```

PR #2983, opened 2024-08-27
Use system python binary if interpreter is not in /usr

DaanDeMeyer:
Same fix as #2982, but for the python_binary() function.

cc @evverx

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2058,7 +2058,7 @@ def join_initrds(initrds: Sequence[Path], output: Path) -> Path:
     return output
 
 
-def python_binary(config: Config, *, binary: Optional[PathString]) -> str:
+def python_binary(config: Config, *, binary: Optional[PathString]) -> PathString:
     tools = (
         not binary or
         not (path := config.find_binary(binary)) or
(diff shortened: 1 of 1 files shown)


evverx:
Looks like it works in the sense that the image was built successfully this time:
```sh
All done.
‣  /home/vagrant/systemd/build/mkosi.output/image.raw size is 9.0G, consumes 2.5G.
```

I guess I still need to tweak the permissions somewhere before containers are launched:
```sh
fatal: detected dubious ownership in repository at '/work/src/pkg/fedora'
...
‣ "/work/sync final" returned non-zero exit code 128.
```

mkosi/__init__.py
```diff
@@ -2067,7 +2067,8 @@ def python_binary(config: Config, *, binary: Optional[PathString]) -> str:
 
     # If there's no tools tree, prefer the interpreter from MKOSI_INTERPRETER. If there is a tools
     # tree, just use the default python3 interpreter.
-    return "python3" if tools and config.tools_tree else sys.executable
+    exe = Path(sys.executable)
+    return "python3" if (tools and config.tools_tree) or not exe.is_relative_to("/usr") else exe
```
behrmann, line 2071:
```suggestion
    exe = sys.executable
    return "python3" if (tools and config.tools_tree) or not Path(exe).is_relative_to("/usr") else exe
```
and we can keep the return value a string or
```suggestion
    exe = Path(sys.executable)
    return Path("python3") if (tools and config.tools_tree) or not exe.is_relative_to("/usr") else exe
```
and make the return value a `Path`. I'd prefer the former.

behrmann (approved):

mkosi/__init__.py
DaanDeMeyer: I dunno PathString seems to convey best here what we want.

evverx:
@DaanDeMeyer Looks like
```
fatal: detected dubious ownership in repository at '/work/src/pkg/fedora'
...
‣ "/work/sync final" returned non-zero exit code 128.
```
can be get around by adding `-ff` everywhere but I wonder if the use cases where images are built as non-root and then containers are run as root are supposed to be supported? I think it was mentioned somewhere that it should all be rootless eventually but it isn't there yet as far as I understand.

PR #2986, opened 2024-08-28
sandbox: introduce the MKOSI_CHROOT_SUPPRESS_CHOWN variable

behrmann:
This allows suppressing chowns in scripts.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -49,6 +49,7 @@ from mkosi.config import (
     Verb,
     Vmm,
     format_bytes,
+    parse_boolean,
     parse_config,
     summary,
     systemd_tool_version,
@@ -66,8 +67,10 @@ from mkosi.pager import page
 from mkosi.partition import Partition, finalize_root, finalize_roothash
 from mkosi.qemu import KernelType, copy_ephemeral, run_qemu, run_ssh, start_journal_remote
 from mkosi.run import (
+    apivfs_options,
     chroot_cmd,
-    chroot_script_cmd,
+    chroot_options,
+    finalize_interpreter,
     finalize_passwd_mounts,
     find_binary,
     fork_and_wait,
@@ -567,9 +570,21 @@ def script_maybe_chroot_sandbox(
     network: bool,
 ) -> Iterator[list[PathString]]:
     options = ["--dir", "/work/src", "--chdir", "/work/src", *options]
+    suppress_chown = parse_boolean(context.config.environment.get("MKOSI_CHROOT_SUPPRESS_CHOWN", "0"))
 
     helpers = {
-        "mkosi-chroot": chroot_script_cmd(tools=bool(context.config.tools_tree), network=network, work=True),
+        "mkosi-chroot": [
+            finalize_interpreter(bool(context.config.tools_tree)), "-SI", "/sandbox.py",
+            "--bind", "/buildroot", "/",
+            "--bind", "/var/tmp", "/var/tmp",
+            *apivfs_options(root=Path("/")),
+            *chroot_options(),
+            "--bind", "/work", "/work",
+            "--chdir", "/work/src",
+            *(["--ro-bind-try", "/etc/resolv.conf", "/etc/resolv.conf"] if network else []),
+            *(["--suppress-chown"] if suppress_chown else []),
+            "--",
+        ],
         "mkosi-as-caller": mkosi_as_caller(),
         **context.config.distribution.package_manager(context.config).scripts(context),
     }
@@ -589,6 +604,9 @@ def script_maybe_chroot_sandbox(
             ) as sandbox:
                 yield sandbox
         else:
+            if suppress_chown:
+                options += ["--suppress-chown"]
+
             with chroot_cmd(
                 root=context.root,
                 network=network,
--- a/mkosi/resources/mkosi.md
+++ b/mkosi/resources/mkosi.md
@@ -2730,6 +2730,11 @@ Note that the minimum required Python version is 3.9.
   changing ownership to root will work but changing ownership to any other user
   or group will fail.
 
+  Note that chown calls are only suppressed when running package managers, but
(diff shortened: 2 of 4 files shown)


mkosi/resources/mkosi.md
```diff
@@ -2730,6 +2730,10 @@ Note that the minimum required Python version is 3.9.
   changing ownership to root will work but changing ownership to any other user
   or group will fail.
 
+  Note that chown calls are only suppressed when running package managers, but
+  not during scripts. If this should be required, e.g. for a build script, you
```
DaanDeMeyer, line 2734: "during scripts" => "when running scripts", "should be required"' => "is required"

mkosi/resources/mkosi.md
```diff
@@ -2730,6 +2730,10 @@ Note that the minimum required Python version is 3.9.
...
+  can set the `MKOSI_CHROOT_SUPPRESS_CHOWN` variable to a non-empty value.
```
DaanDeMeyer, line 2735: Please check for `1` or a boolean value instead of non-empty

mkosi/sandbox/__init__.py
```diff
@@ -619,7 +619,8 @@ def main() -> None:
     upperdir = ""
     workdir = ""
     chdir = None
-    become_root = suppress_chown = unshare_net = unshare_ipc = False
+    become_root = unshare_net = unshare_ipc = False
+    suppress_chown = bool(os.getenv("MKOSI_CHROOT_SUPPRESS_CHOWN", False))
```
DaanDeMeyer, line 623: This should be processed outside of mkosi-sandbox. mkosi should look for the environment variable and pass --suppress-chown to chroot_cmd() when running scripts if it is set.

DaanDeMeyer (changes requested):

mkosi/__init__.py
```diff
@@ -568,8 +571,23 @@ def script_maybe_chroot_sandbox(
...
     helpers = {
-        "mkosi-chroot": chroot_script_cmd(tools=bool(context.config.tools_tree), network=network, work=True),
+        "mkosi-chroot": [
+            choose_interpreter(bool(context.config.tools_tree)), "-SI", "/sandbox.py",
+            "--bind", "/buildroot", "/",
+            "--bind", "/var/tmp", "/var/tmp",
+            *apivfs_options(root=Path("/")),
+            *chroot_options(),
+            "--bind", "/work", "/work",
+            "--chdir", "/work/src",
```
DaanDeMeyer, line 586: You can pass `options` here as well and get rid of passing `--chdir` and `--suppress-chown` manually here

mkosi/run.py
```diff
@@ -616,15 +603,12 @@ def chroot_cmd(
...
-        "python3" if tools or not exe.is_relative_to("/usr") else exe, "-SI", "/sandbox.py",
-        "--bind", "/buildroot", "/",
-        "--bind", "/var/tmp", "/var/tmp",
-        *apivfs_options(root=Path("/")),
-        *chroot_options(),
-        *(["--bind", "/work", "/work", "--chdir", "/work/src"] if work else []),
-        *(["--ro-bind-try", "/etc/resolv.conf", "/etc/resolv.conf"] if network else []),
-        "--",
-    ]
+def choose_interpreter(tools: bool) -> str:
```
DaanDeMeyer, line 606: `finalize_interpreter`? Seems more consistent with how we usually name things

DaanDeMeyer (changes requested):
Looks good, two minor things

mkosi/__init__.py
behrmann: Split out options from the caller and the ones here, which then allows the reuse.

mkosi/__init__.py
```diff
@@ -566,10 +569,23 @@ def script_maybe_chroot_sandbox(
     options: Sequence[PathString],
     network: bool,
 ) -> Iterator[list[PathString]]:
-    options = ["--dir", "/work/src", "--chdir", "/work/src", *options]
+    localoptions = ["--dir", "/work/src", "--chdir", "/work/src"]
+
+    if parse_boolean(context.config.environment.get("MKOSI_CHROOT_SUPPRESS_CHOWN", "0")):
+        localoptions += ["--suppress-chown"]
```
DaanDeMeyer, line 575: So this approach actually means we pass --suppress-chown to the non-chroot script as well. So we should handle this separately after all.

So I'd drop `localoptions` again and handle it like you did earlier after all, my bad.

DaanDeMeyer (changes requested):

mkosi/__init__.py
```diff
@@ -568,8 +571,23 @@ def script_maybe_chroot_sandbox(
 ) -> Iterator[list[PathString]]:
     options = ["--dir", "/work/src", "--chdir", "/work/src", *options]
 
+    suppress_chown = parse_boolean(context.config.environment.get("MKOSI_CHROOT_SUPPRESS_CHOWN", "0"))
+    if suppress_chown:
+        options += ["--suppress-chown"]
```
DaanDeMeyer, line 576: This still passes `--suppress-chown` to the non-chroot command. This should only be added to the `mkosi-chroot` command options and the chroot script options.

DaanDeMeyer (changes requested):

mkosi/__init__.py
```diff
@@ -567,15 +570,30 @@ def script_maybe_chroot_sandbox(
...
+            "--",
+        ],
         "mkosi-as-caller": mkosi_as_caller(),
         **context.config.distribution.package_manager(context.config).scripts(context),
     }
 
     with finalize_host_scripts(context, helpers) as hd:
         if script.suffix != ".chroot":
+            if suppress_chown:
+                options += ["--suppress-chown"]
```
DaanDeMeyer, line 595: This needs to be in the other branch

DaanDeMeyer (approved):

PR #2987, opened 2024-08-29
Only run sync scripts if we're going to build the image

DaanDeMeyer:


bluca (approved):

DaanDeMeyer:
cc @evverx

evverx:
> good-to-merge/waiting-for-ci

Does the CI actually test those things?

I think it should break the kludge with `-ff` because it worked only because `pkg/fedora` was overriden with `-ff boot`. I'll run it a bit later.

evverx:
> I don't have the time to write extensive tests for all these features unfortunately

@DaanDeMeyer that's OK. I didn't mean to say that it isn't OK (even though this use case is explicitly mentioned in the systemd HACKING documentation). I'm not sure how stuff is reviewed so fast without actually running anything.

evverx:
So I took a look at what the systemd github mkosi action does and it seems it runs everything as root so I guess I'll just switch to that as well given that it's tested and should probably work most of the time. I think this PR makes sense but it probably makes things worse in the sense that `-ff` can no longer "fix" issues with permissions when `boot` is run as `root`. I'll leave it to someone else to be a guinea pig and test those use cases.

evverx:
For the record with `-ff` it just failed with
```sh
fatal: detected dubious ownership in repository at '/work/src/pkg/fedora'
...
‣ "/work/sync final" returned non-zero exit code 128.
```
then again it isn't tested so it isn't supported. Maybe it should be documented that everything should be run as root.

DaanDeMeyer:
@evverx We do not drop privileges anymore when running as root, so doing image builds as root sometimes and without root other times is going to lead to issues. `-f` implies a build is done.

Yes nspawn needs to be root for the foreseeable future. So if you want to build an image and then boot in nspawn you have to run two commands: `mkosi -f && sudo mkosi boot`

With this fix we won't run the sync script anymore when you're not doing a build so this will work.

I just get around all these issues by running `mkosi qemu` which does work unprivileged and has none of the limitations of a container.

DaanDeMeyer:
Switched to a cleaner approach

evverx:
> mkosi -f && sudo mkosi boot

That's exactly what I had been doing up until today. A few days ago it started failing with those errors so I switched to running both commands as `root` by analogy with that the systemd CI does.

> I just get around all these issues by running mkosi qemu

The problem is that I already run that in VMs so another layer of virtualization is noticeable.

DaanDeMeyer:
> > mkosi -f && sudo mkosi boot
> 
> That's exactly what I had been doing up until today. A few days ago it started failing with those errors so I switched to running both commands as `root` by analogy with that the systemd CI does.

Yes I broke that use case accidentally but it should be fixed by this PR.

> > I just get around all these issues by running mkosi qemu
> 
> The problem is that I already run that in VMs so another layer of virtualization is noticeable.

In a VM I would personally always just run everything as root.

evverx:
> In a VM I would personally always just run everything as root.

Can't argue with that. I think it was run as non-root because initially I didn't run `mkosi` to boot images I built. Then I wasn't able to boot things with `podman` anymore for some reason and added `sudo mkosi boot` instead and that's how I ended up with all that :-)

evverx:
> it should be fixed by this PR

It seems to be working.

The fedora repository seems to be down
```
‣ /home/vagrant/systemd/build/mkosi.cache/fedora~40~x86-64~build.cache does not exist, not reusing cached images                                              
‣ Running sync script /home/vagrant/systemd/mkosi.images/build/mkosi.sync…                                                                                    
Cloning into 'pkg/fedora'...                                                                                                                                  
fatal: unable to access 'https://src.fedoraproject.org/rpms/systemd.git/': Failed to connect to src.fedoraproject.org port 443 after 269644 ms: Couldn't conne
ct to server                                                                                                                                                  
‣ "/work/sync final" returned non-zero exit code 128. 
```
so I didn't test it with the main branch of systemd.

PR #2990, opened 2024-08-30
More docs at your finger tips

behrmann:
This adds the feature to pass an argument to the documentation command, so that the man page we've grown in addition can be viewed easily as well.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -247,6 +247,12 @@ class DocFormat(StrEnum):
     pandoc   = enum.auto()
     system   = enum.auto()
 
+    @classmethod
+    def all(cls) -> list["DocFormat"]:
+        # this excludes auto and encodes the order in which these should be
+        # checked when searching for docs
+        return [cls.man, cls.pandoc, cls.markdown, cls.system]
+
 
 class Bootloader(StrEnum):
     none         = enum.auto()
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -88,6 +88,7 @@ class Verb(StrEnum):
             Verb.coredumpctl,
             Verb.burn,
             Verb.completion,
+            Verb.documentation,
         )
 
     def needs_build(self) -> bool:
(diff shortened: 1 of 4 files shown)


keszybz:
Looks reasonable…

DaanDeMeyer:
We discussed this in chat but I'd prefer mkosi-initrd learns a --documentation switch to show its documentation.

Doing the same for mkosi-sandbox is harder without introducing dependencies on the rest of mkosi but it could be done if `show_docs()` is made independent of the rest of mkosi and is implemented using only stdlib stuff.

behrmann:
I'll rework it next week. I think it's fine to do this both for mkosi-initrd and mkosi-sandbox by lazily using the `show_docs` stuff, which in the case of `mkosi-sandbox` will just not work within the sandbox, but I think that's fine, people will usually have another terminal open.

mkosi/initrd.py
```diff
@@ -62,6 +65,12 @@ def main() -> None:
         action="store_true",
         default=False,
     )
+    parser.add_argument(
+        "--show-documentation",
```
DaanDeMeyer, line 69: `-D` as a short option?

mkosi/config.py
```diff
@@ -246,6 +247,10 @@ class DocFormat(StrEnum):
     pandoc   = enum.auto()
     system   = enum.auto()
 
+    @classmethod
+    def all(cls) -> list["DocFormat"]:
+        return [cls.man, cls.pandoc, cls.markdown, cls.system]
```
DaanDeMeyer, line 252: You can just do `list(DocFormat)` no?

DaanDeMeyer (changes requested):

mkosi/config.py
behrmann: No, because this would include `auto` and this also encodes their order, which otherwise would be by the string value, i.e. lexicographical. Added a comment to that extent.

PR #2993, opened 2024-08-31
Add cat-config verb

keszybz:


mkosi/config.py
```diff
@@ -4111,10 +4122,29 @@ def format_bytes_or_none(num_bytes: Optional[int]) -> str:
...
+    for n, config in enumerate(images):
+        if n > 0:
+            print(file=c)
+        print(bold(f"### IMAGE: {config.image or 'default'}"), file=c)
+        for path in config.config_paths:
+            # Display the paths as relative to ., if underneath.
+            try:
+                path = path.relative_to(Path.cwd())
+            except ValueError:
+                pass
```
DaanDeMeyer, line 4142: I'd just use `path.is_relative_to()` here, it'll be shorter and you get rid of the false positive from CodeQL at the same time. Sure there's the whole ask for forgiveness thing rather than permission but I think we can ignore that here.

mkosi/config.py
```diff
@@ -4111,10 +4122,29 @@ def format_bytes_or_none(num_bytes: Optional[int]) -> str:
...
+    return f"{Style.bold}{s}{Style.reset}"
+
+
+def cat_config(images: Sequence[Config]) -> str:
+    c = io.StringIO()
+    for n, config in enumerate(images):
+        if n > 0:
+            print(file=c)
+        print(bold(f"### IMAGE: {config.image or 'default'}"), file=c)
+        for path in config.config_paths:
```
DaanDeMeyer, line 4137: This could do with some more newlines

mkosi/config.py
```diff
@@ -1411,6 +1413,7 @@ class Config:
     """
 
     profile: Optional[str]
+    config_paths: list[Path]
```
DaanDeMeyer, line 1416: I would name this `files`. The class is already named `Config`, no need to repeat it here.

mkosi/resources/mkosi.md
```diff
@@ -51,17 +53,21 @@ that may generate disk images with a number of bells and whistles.
 The following command line verbs are known:
 
 `summary`
-:   Outputs a human-readable summary of all options used for building an
+:   Output a human-readable summary of all options used for building an
```
DaanDeMeyer, line 56: Might as well use `Show` instead of `Output`

mkosi/resources/mkosi.md
```diff
@@ -51,17 +53,21 @@ that may generate disk images with a number of bells and whistles.
...
     image. This will parse the command line and `mkosi.conf` file as it
-    would do on `build`, but only output what it is configured for and not
+    would do on `build`, but only print what it is configured for and not
```
DaanDeMeyer, line 58: "what it is" => "what is"

mkosi/config.py
```diff
@@ -4031,7 +4031,9 @@ def load_environment(args: argparse.Namespace) -> dict[str, str]:
     if gnupghome := os.getenv("GNUPGHOME"):
         env["GNUPGHOME"] = gnupghome
 
-    env |= dict(parse_environment(line) for f in args.environment_files for line in f.read_text().strip().splitlines())
+    env |= dict(parse_environment(line)
+                for f in args.environment_files
+                for line in f.read_text().strip().splitlines())
```
DaanDeMeyer, line 4036: I don't mind really but this is completely different from how we format stuff in the rest of the codebase. I know we disagree on proper formatting but let's keep things consistent please.

DaanDeMeyer (changes requested):
Looks good, just stylistic comments

mkosi/resources/mkosi.md
keszybz: I did that, but then I noticed that the text needs more updating.

mkosi/resources/mkosi.md
keszybz: "what it is … configured for". It is correct.

mkosi/config.py
keszybz: The previous version was a giant completely illegible line… So I think what I had was better. Please see if you like the new version.

DaanDeMeyer (approved):

DaanDeMeyer:
CentOS CI is broken because they increased their initrd size by adding selinux-policy as an unconditional dependency to their systemd package

OpenSUSE CI is flaky in test_initrd_luks.

mkosi/config.py
```diff
@@ -3701,7 +3708,10 @@ def parse_config_one(self, path: Path, profiles: bool = False, local: bool = Fal
                         )
 
         if path.exists():
-            logging.debug(f"Including configuration file {Path.cwd() / path}")
+            p = Path.cwd() / path
```
DaanDeMeyer, line 3711: Should use a different name here otherwise mypy gets angry about types because python for some insane reason doesn't do block scoping

mkosi/config.py
```diff
@@ -1411,6 +1413,7 @@ class Config:
     """
 
     profile: Optional[str]
+    files: list[Path]
```
DaanDeMeyer, line 1416: This needs to be added to test_json as well in the tests

mkosi/config.py
keszybz: Yep. I renamed it to `abs_path`.

behrmann:
Nice, the other day I thought that we needed this and there it is. Thanks!

Only nit: I would have exchanged the `dict` call to a literal when we're already touching the line.

keszybz:
> Only nit: I would have exchanged the `dict` call to a literal when we're already touching the line.

An explicit `dict()` is needed because the source is a generator.

PR #2996, opened 2024-09-03
Add 'login' to Debian/Ubuntu/Kali package list

bluca:
The login package as provided by util-linux is 'Protected' but no longer 'Essential' and that's intentional, so it will not be pulled in by default. Add it to the list.


DaanDeMeyer:
Should be added `to docs/bootable.md` as well

DaanDeMeyer (approved):

PR #2997, opened 2024-09-03
base64 encode mirror if we put it in package cache dir key

DaanDeMeyer:
Mirrors contain / characters. Paths can't contain / characters. Let's base64 encode the mirrors to avoid this issue.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -1600,7 +1600,7 @@ class Config:
     def package_cache_dir_or_default(self) -> Path:
         key = f"{self.distribution}~{self.release}~{self.architecture}"
         if self.mirror:
-            key += f"-{self.mirror}"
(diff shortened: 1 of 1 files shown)


behrmann:
Looks reasonable to me, but wouldn't running this through systemd-escape be nicer, since it would still be vaguely human readable? It's not really something a human has to look at, but if it ever comes up in debugging I can unmangle sd escape better in my head than base64 ;)

DaanDeMeyer:
> Looks reasonable to me, but wouldn't running this through systemd-escape be nicer, since it would still be vaguely human readable? It's not really something a human has to look at, but if it ever comes up in debugging I can unmangle sd escape better in my head than base64 ;)

I don't think it matters for the mirrors tbh

PR #3004, opened 2024-09-05
sandbox: Use separate variable name when we change types

DaanDeMeyer:


behrmann (approved):

behrmann:
It's just a renamed variable, no need to burn CI for this, since it type checks.

PR #3005, opened 2024-09-05
Various mypy fixes

DaanDeMeyer:


mkosi/sandbox.py
```diff
@@ -626,6 +628,8 @@ def main() -> None:
     chdir = None
     become_root = suppress_chown = unshare_net = unshare_ipc = False
 
+    ttyname = os.ttyname(2) if os.isatty(2) else ""
```
behrmann, line 631: While we're at it
```suggestion
    ttyname = os.ttyname(2) or ""
```
this way we only evaluate it once

behrmann: Oops, operational blindness.

PR #3008, opened 2024-09-06
Use more directories of sandbox trees in the sandbox

DaanDeMeyer:
Previously, we only picked up /usr and /etc from the sandbox trees.
Let's make this more generic and pick up a bunch of extra directories
as well. To avoid any changes persisting outside of the sandbox, let's
use overlayfs with a temporary writable directory as the upperdir of the
overlayfs (and make sure we use a tmpfs as the upperdir for /tmp and /run).


mkosi/run.py
```diff
@@ -514,36 +519,49 @@ def sandbox_cmd(
...
+    else:
+        cmdline += ["--dir", "/var/tmp", "--unshare-ipc"]
 
-    for d in ("bin", "sbin", "lib", "lib32", "lib64"):
-        if (p := tools / d).is_symlink():
-            cmdline += ["--symlink", p.readlink(), Path("/") / p.relative_to(tools)]
-        elif p.is_dir():
-            cmdline += ["--ro-bind", p, Path("/") / p.relative_to(tools)]
+        if devices:
+            cmdline += ["--bind", "/sys", "/sys", "--bind", "/dev", "/dev"]
```
behrmann, line 526: Previously in the `relaxed=False and devices=True` case, this would have bind mounted `/run` as well, but is not now or am I missing something?

mkosi/run.py
```diff
@@ -514,36 +519,49 @@ def sandbox_cmd(
...
 
     if scripts:
         cmdline += ["--ro-bind", scripts, "/scripts"]
 
-    if network and not relaxed and Path("/etc/resolv.conf").exists():
-        cmdline += ["--ro-bind", "/etc/resolv.conf", "/etc/resolv.conf"]
+    with contextlib.ExitStack() as stack:
+        for d in ("etc", "opt", "srv", "media", "mnt", "var", "run", "tmp"):
+            if not overlay and not relaxed:
+                continue
```
behrmann, line 543: this short-circuits the whole loop body, so maybe this could be hoisted up to yield immediately in that case?

mkosi/run.py
DaanDeMeyer: Yeah that's a bug fix, there's no need to mount /run when we need device access.

behrmann: Excellent! Could you add that to the commit message in case anybody is following along?

mkosi/qemu.py
```diff
@@ -496,14 +497,14 @@ def start_journal_remote(config: Config, sockfd: int) -> Iterator[None]:
         with spawn(
             [
                 bin,
-                "--output", config.forward_journal,
+                "--output", joinpath("/work", str(config.forward_journal)),
```
behrmann, line 500: Why is this (and the rest in this file not using `workdir`?

behrmann (commented):
Looks reasonable to me, though I only had quick look on what changed to the previous version.

mkosi/qemu.py
DaanDeMeyer: Good catch

PR #3010, opened 2024-09-09
Add bin/mkosi-initrd and bin/mkosi-sandbox wrappers and some cleanups

keszybz:


bin/mkosi
```diff
@@ -1,12 +1,12 @@
 #!/usr/bin/env bash
 # SPDX-License-Identifier: LGPL-2.1-or-later
 set -e
-PYTHONPATH="$(dirname "$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")")"
-export PYTHONPATH
+export PYTHONPATH="$(dirname "$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")")"
```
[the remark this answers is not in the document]
behrmann: That shellcheck warning is why this was separate.

bin/mkosi
```diff
@@ -1,12 +1,12 @@
...
+command="$(basename "${BASH_SOURCE[0]}" | tr - .)"
```
behrmann, line 5:
```suggestion
command="$(basename "${BASH_SOURCE[0]//-/.}")"
```

behrmann (commented):
Not opposed to the line-length changes, although that's probably something where we should go to 109 in general, since right now they are within the limit of 119. 

The generator change in sandbox will probably not fly, since that would probably also pull in `Generator` from typing, but I'll leave that to @DaanDeMeyer since he did all the import wrangling for sandbox.

mkosi/sandbox.py
```diff
@@ -6,6 +6,7 @@
 itself. To keep the runtime of this script to a minimum, please don't import any extra modules if it can be avoided.
 """
 
+import contextlib
```
DaanDeMeyer, line 11: This stuff is extremely sensitive to ms amount increases given how many times we invoke it. Please drop this, I don't think the drop in complexity outweights the extra import in this case.

DaanDeMeyer (changes requested):

mkosi/sandbox.py
DaanDeMeyer: Some non scientific benchmarks to show it does add overhead:

```

➜  mkosi git:(sandbox) time python3 -S -I -c "import ctypes; import os; import sys;"

________________________________________________________
Executed in   11.17 millis    fish           external
   usr time    7.13 millis  327.00 micros    6.80 millis
   sys time    3.95 millis   46.00 micros    3.90 millis

➜  mkosi git:(sandbox) time python3 -S -I -c "import ctypes; import os; import sys; import contextlib"

________________________________________________________
Executed in   20.48 millis    fish           external
   usr time   32.30 millis   15.97 millis   16.33 millis
   sys time    7.91 millis    4.87 millis    3.05 millis

➜  mkosi git:(sandbox) time python3 -S -I -c "import ctypes; import os; import sys;"

________________________________________________________
Executed in   11.40 millis    fish           external
   usr time    8.30 millis  370.00 micros    7.93 millis
   sys time    3.04 millis   52.00 micros    2.99 millis

➜  mkosi git:(sandbox) time python3 -S -I -c "import ctypes; import os; import sys; import contextlib"

________________________________________________________
Executed in   14.01 millis    fish           external
   usr time   12.56 millis    0.00 micros   12.56 millis
   sys time    1.44 millis  387.00 micros    1.05 millis
```

keszybz: I have my doubts whether this microbenchmark is useful. I expect that normally something else pulls in contextlib, so the addition "import" is just a lookup in `sys.modules`, i.e. effectively free. But it's not terribly important, so I dropped the commit.

behrmann (approved):

mkosi/sandbox.py
DaanDeMeyer: Well this microbenchmark imports everything that's imported by sandbox.py, so there's actually nothing pulling in contextlib. Upstream python goes through quite a bit of effort to keep the imports of core modules like these very minimal

PR #3012, opened 2024-09-10
Introduce History= setting

DaanDeMeyer:
Currently, to boot an image with mkosi qemu after building it with
mkosi build, various settings have to be identical to when mkosi build
was invoked to make sure that mkosi can find the outputs of the previous
build. Because this is rather error prone and annoying, let's introduce
a History= setting to allow mkosi to remember the configuration of the
last build which can then be read again when running a verb that operates
on a built image.

Another case where this is extremely useful is when some part of the
configuration changes every single time, for example if mkosi.version
is an executable script that outputs the current time, which is then
encoded in the output name, we have to remember the previous config,
otherwise mkosi wouldn't be able to find the outputs of the previous
build.

Note that while we load the configuration of the previous build, we
ignore all settings from the [Host] section which we read again from
the configuration files, as the user should be able to change these
without rebuilding the image.


mkosi/__init__.py
```diff
@@ -4100,8 +4098,12 @@ def run_verb(args: Args, images: Sequence[Config], *, resources: Path) -> None:
                         package_dir=Path(package_dir),
                     )
 
-            if args.auto_bump:
-                bump_image_version()
+        if args.auto_bump:
+            bump_image_version()
```
behrmann, line 4102: Is this changing indentation here because it doesn't need the temporary directories? That makes sense, but is a bit unexpected in the commit.

behrmann (commented):
Haven't run it, but codewise looks reasonable to me, it's mostly moves.

mkosi/__init__.py
DaanDeMeyer: Yeah exactly that, I was too lazy to make it a separate commit.

NEWS.md
```diff
@@ -39,6 +39,9 @@
 - Note to packagers: The manual pages have been moved to resources/man
   and now include man pages for mkosi-initrd and mkosi-sandbox as
   well.
+- `InitrdInclude=` was removed. If you're using `InitrdInclude=`, please
+  build your initrd via a subimage in `mkosi.images` containing
+  `Include=mkosi-initrd` and any customizations you wish to add.
```
septatrix, line 44: How exactly should one use that? Naming the image initrd and adding the include does not seem to work. I could add a postbuild step to move the generated initrd to `$ARTIFACTDIR/io.mkosi.initrd` but that seems like a hacky solution. Am I missing an easier solution?

DaanDeMeyer: `Initrds=` setting

septatrix: Ah yeah I forgot about that...I knew I was missing sth

PR #3017, opened 2024-09-11
Beef up tty checks for color handling

### LINKED ISSUE #3013: Detect pipe or dumb terminal for `cat-config`

main

### Used host distribution

Fedora 40

### Used target distribution

Fedora 40

### Linux kernel version used

6.10.7-200.fc40.x86_64

x86_64

Piping `mkosi cat-config` into another command or running it with `TERM=dumb` still outputs ANSI escape codes.
In my case I was piping the output to `wl-copy` (to paste it here for another bug report) but it did not detect it as text since it starts with ansi escape codes.

### Used mkosi config

_No response_

### mkosi output

_No response_

DaanDeMeyer:
Let's use the same checks used by systemd.

Fixes #3013

### DIFF
--- a/mkosi/log.py
+++ b/mkosi/log.py
@@ -14,13 +14,20 @@ ARG_DEBUG_SHELL = contextvars.ContextVar("debug-shell", default=False)
 LEVEL = 0
 
 
+def terminal_is_dumb() -> bool:
+    if not sys.stdout.isatty() and not sys.stderr.isatty():
+        return True
+
+    return os.getenv("TERM", "") == "dumb"
+
+
 class Style:
-    bold: Final[str] = "\033[0;1;39m" if sys.stderr.isatty() else ""
-    blue: Final[str] = "\033[0;1;34m" if sys.stderr.isatty() else ""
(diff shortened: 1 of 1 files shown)


mkosi/log.py
```diff
@@ -14,13 +14,17 @@
 LEVEL = 0
 
 
+def terminal_is_dumb() -> bool:
+    return not sys.stdout.isatty() and not sys.stderr.isatty() and os.getenv("TERM", "") != "dumb"
```
behrmann, line 18: Wouldn't this make more sense
```suggestion
    return not sys.stdout.isatty() or not sys.stderr.isatty() or os.getenv("TERM", "") == "dumb"
```
since if `TERM=dumb` we should probably hide colour even if stderr and stdout are attached to a TTY, also if either stderr or stdout are not a TTY assuming a dumb terminal might be safer.

behrmann (approved):

septatrix:
For piping stuff like `cat-config` and `summary` this still happens because those pipe to stdout (which is not a tty) but stderr is still a tty so they still try to color the output

PR #3018, opened 2024-09-11
qemu: enroll certificate in MOK for dm-verity and kernel mods

bluca:
Automatically enroll the secure boot certificate in MOK, and also set the boolean so that the kernel trusts it to verify kernel modules and dm-verity volumes. This requires secure boot to be enabled to be effective, otherwise the kernel will ignore it.

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -646,6 +646,54 @@ def finalize_firmware_variables(
     stack: contextlib.ExitStack,
 ) -> tuple[Path, str]:
     ovmf_vars = stack.enter_context(tempfile.NamedTemporaryFile(prefix="mkosi-ovmf-vars-"))
+    ovmf_json = stack.enter_context(tempfile.NamedTemporaryFile(prefix="mkosi-ovmf-vars-json-"))
+
+    if config.secure_boot_certificate:
+        # In order to make the kernel use mkosi.crt to verify dm-verity volumes, we need to
+        # create runtime and volatile EFI variables, MokListTrustedRT and MokListRT, as
+        # it's only from kernel 6.11 that db is used for this purpose. Note that, unlike
+        # certificates in db, the certificates in MOK are only used if secure boot is enabled.
+        # First create an efivar with the certificate, and then read it and convert to the
+        # json format that virt-fw-vars expects. We skip this in case the user provides their
+        # own custom firmware variables.
+        # attr=4 means EFI_VARIABLE_RUNTIME_ACCESS, and the GUID is the well-known MOK one.
+
+        mok_sigdb = stack.enter_context(tempfile.NamedTemporaryFile(prefix="mkosi-ovmf-vars-sigdb-"))
+        run(
+            [
+                "virt-fw-sigdb",
+                "--add-cert", "605dab50-e046-4300-abb6-3dd810dd8b23", config.secure_boot_certificate,
+                "--output", mok_sigdb.name,
+            ],
+            sandbox=config.sandbox(
+                binary=qemu,
+                options=[
+                    "--bind", mok_sigdb.name, mok_sigdb.name,
+                    "--ro-bind", config.secure_boot_certificate, config.secure_boot_certificate,
+                ],
+            ),
+        )
+
+        mok_json = {
+            "version": 2,
+            "variables": [
+                {
+                    "name": "MokListRT",
+                    "guid": "605dab50-e046-4300-abb6-3dd810dd8b23",
+                    "attr": 4,
+                    "data": mok_sigdb.read().hex(),
+                },
+                {
+                    "name": "MokListTrustedRT",
+                    "guid": "605dab50-e046-4300-abb6-3dd810dd8b23",
+                    "attr": 4,
+                    "data": "01",
+                },
+            ],
+        }
+        ovmf_json.write(json.dumps(mok_json).encode())
+        ovmf_json.flush()
+
     if config.qemu_firmware_variables in (None, Path("custom"), Path("microsoft")):
         ovmf_vars_format = ovmf.vars_format
     else:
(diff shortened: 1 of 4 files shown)


mkosi/qemu.py
```diff
@@ -646,20 +646,62 @@ def finalize_firmware_variables(
...
+                        data=bytes(mok_sigdb).hex(),
+                    ),
+                    dict(
+                        name="MokListTrustedRT",
+                        guid="605dab50-e046-4300-abb6-3dd810dd8b23",
+                        attr=4,
+                        data="01",
+                    ),
+                ],
+            )
```
behrmann, line 682: For the environment variables I do understand where wanting to use `dict` comes from, but let's not make this too obscure
```suggestion
            mok_json = {
                "version": 2,
                "variable" [
                    {
                        "name": "MokListRT",
                        "guid": "605dab50-e046-4300-abb6-3dd810dd8b23",
                        "attr": 4,
                        "data": bytes(mok_sigdb).hex(),
                    },
                    {
                        "name": "MokListTrustedRT",
                        "guid": "605dab50-e046-4300-abb6-3dd810dd8b23",
                        "attr": 4,
                        "data": "01",
                    },
                ],
            }
```

mkosi/qemu.py
```diff
@@ -646,20 +646,62 @@ def finalize_firmware_variables(
...
+            ovmf_json.write(json.dumps(mok_json).encode())
```
behrmann, line 683: Any reason against
```suggestion
            json.dump(mok_json, ovmf_json)
```

mkosi/qemu.py
```diff
@@ -646,20 +646,62 @@ def finalize_firmware_variables(
     stack: contextlib.ExitStack,
 ) -> tuple[Path, str]:
     ovmf_vars = stack.enter_context(tempfile.NamedTemporaryFile(prefix="mkosi-ovmf-vars-"))
+    ovmf_json = stack.enter_context(tempfile.NamedTemporaryFile(prefix="mkosi-ovmf-vars-json-"))
+    have_mok_json = False
+
+    if config.secure_boot_certificate:
+        try:
+            from virt.firmware.efi import guids, siglist
```
behrmann, line 654: Is there a sensible way to get this within the Python stdlib? Right now we don't have external dependencies. `guids.parse_str` only seems to generate a UUID object, which is from the stdlib, but `mok_sigdb.add_cert` seems to read the raw bytes out of the X509 cert, which is a tad more difficult.

mkosi/qemu.py
bluca:
```
  File "/home/bluca/git/mkosi/mkosi/qemu.py", line 683, in finalize_firmware_variables
    json.dump(mok_json, ovmf_json)
  File "/usr/lib/python3.11/json/__init__.py", line 180, in dump
    fp.write(chunk)
  File "/usr/lib/python3.11/tempfile.py", line 638, in func_wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
TypeError: a bytes-like object is required, not 'str'
```

mkosi/qemu.py
DaanDeMeyer: Shelling out to `virt-fw-vars` is much better than importing the module. Shelling out means we can use `virt-fw-vars` from a tools tree if there is one, whereas with importing we can't do that without going through extra hoops. Please rework this to use the `virt-fw-vars` command line tool, looking at its `--help` there seems to be options to do what you're doing here.

pyproject.toml
```diff
@@ -45,6 +45,8 @@ pythonVersion = "3.9"
 
 [tool.mypy]
 python_version = 3.9
+# Fails on external libraries like virt.firmware
+ignore_missing_imports = true
```
DaanDeMeyer, line 49: Think this can be dropped now?

mkosi/qemu.py
```diff
@@ -679,7 +728,26 @@ def finalize_firmware_variables(
...
+                    binary=qemu,
+                    options=[
+                        "--bind", ovmf_vars.name, ovmf_vars.name,
+                        "--ro-bind", ovmf_json.name, ovmf_json.name,
+                        "--ro-bind", vars, vars,
+                    ],
+                ),
+            )
+        else:
+            shutil.copy2(vars, Path(ovmf_vars.name))
```
DaanDeMeyer, line 750: This should be dropped, just because the image is built with a secure boot certificate shouldn't imply that we automatically add it to any given OVMF variables, especially ones provided by the user.

DaanDeMeyer (changes requested):

mkosi/qemu.py
```diff
@@ -646,6 +646,53 @@ def finalize_firmware_variables(
...
+                {
+                    "name": "MokListTrustedRT",
+                    "guid": "605dab50-e046-4300-abb6-3dd810dd8b23",
+                    "attr": 4,
+                    "data": "01",
+                },
+            ],
+        }
+        ovmf_json.write(json.dumps(mok_json).encode())
+        ovmf_json.flush()
```
DaanDeMeyer, line 694: Make this a function and call it from within the `custom` if block and `assert` on `secure_boot_certificate`

mkosi/qemu.py
bluca: it's now skipped if there are user-provided vars

mkosi/qemu.py
```diff
@@ -679,7 +729,27 @@ def finalize_firmware_variables(
             if config.qemu_firmware_variables == Path("microsoft") or not config.qemu_firmware_variables
             else config.qemu_firmware_variables
         )
-        shutil.copy2(vars, Path(ovmf_vars.name))
+        if (config.secure_boot_certificate and
+                (config.qemu_firmware_variables == Path("microsoft") or not config.qemu_firmware_variables)):
```
DaanDeMeyer, line 733: As mentioned in Telegram, I would really prefer we don't conditionalize based on whether there's a secure boot certificate configured or not as that's build time configuration and not runtime configuration. So if you'd like to have this for the microsoft keys as well please gate it behind a new constant for `QemuFirmwareVariables=` named `microsoft+mok`

PR #3026, opened 2024-09-13
Enable history for the default image

DaanDeMeyer:


behrmann (approved):
The changes to the integration tests are quite something, but they are easier to read now.

PR #3028, opened 2024-09-14
Remove references of bubblewrap in docs/tools configuration

jelly:
Since b3a3e7e7fcb2a4e8f mkosi no longer relies on bubblewrap for sandboxing.


keszybz (approved):

DaanDeMeyer:
These were left in intentionally. It's not because we don't use bubblewrap anymore that we wouldn't install it into tools trees anymore for usage by various user scripts and such

DaanDeMeyer:
@jelly Did you see any actual issues with bubblewrap getting installed into tools trees?

keszybz:
But why would you want to install bubblewrap? It was added for a specific purpose and that purpose is now gone…

behrmann:
I think it's fine to drop this, I don't think there will be many (any?) users using bwrap directly in scripts in the first place, if there are, they can probably move easily to `mkosi-sandbox` or can use `ToolsTreePackages=`. Not installing makes the default slightly slimmer, which is nice for most use like our CI.

DaanDeMeyer:
> But why would you want to install bubblewrap? It was added for a specific purpose and that purpose is now gone…

Sure but everything we install in the tools tree is available for use by users in `mkosi.postinst` and such as well. It's not because we don't use it anymore that users won't use it anymore in their scripts is what I'm saying

PR #3029, opened 2024-09-14
Fix relative path calculation in filter_kernel_modules()

DaanDeMeyer:
I'm not sure what possessed me when I last touched this, but to get the path relative to the kernel/ directory we have to strip of the first 5 parts, not just 1.


DaanDeMeyer:
cc @septatrix this should fix the issue as well, can you verify?

DaanDeMeyer:
@septatrix Ping

septatrix (approved):

PR #3030, opened 2024-09-14
Two improvements

DaanDeMeyer:


mkosi/run.py
```diff
@@ -435,11 +435,23 @@ def vartmpdir() -> Iterator[Path]:
...
+                def remove() -> None:
+                    acquire_privileges()
+                    shutil.rmtree(d)
+
+                fork_and_wait(remove)
+                return
+            elif e.errno != errno.ENOENT:
+                raise
 
         shutil.rmtree(d)
```
behrmann, line 456: I think by using the `else` on the try except the warning about the return in the except could be shut up as well. 
```suggestion
        except OSError as e:
            if e.errno == errno.ENOTEMPTY:
                def remove() -> None:
                    acquire_privileges()
                    shutil.rmtree(d)

                fork_and_wait(remove)
            elif e.errno != errno.ENOENT:
                raise
        else:
            shutil.rmtree(d)
```

behrmann (approved):
LGTM, one alternative for the warning. It' a bit of the trade off and I don't think it would be necessarily better, because it could be misread as being on the if-elif-block.

mkosi/run.py
DaanDeMeyer: Doesn't work, the else block is only executed if the try block finishes without raising an exception

behrmann: You're right, I wrongly assumed we could only fall through if there are not exceptions, but `ENOENT` is silently skipped. This shouldn't be bad, but it's a bit unexpected, so maybe this should be logged.

PR #3032, opened 2024-09-16
Add back MachineId= setting

DaanDeMeyer:
To build an image with a dm-verity protected root partition that has
a persistent machine ID, the machine ID has to be embedded in the image,
so let's add back the MachineId= setting to support this use case.

### DIFF
--- a/mkosi/resources/man/mkosi.md
+++ b/mkosi/resources/man/mkosi.md
@@ -1098,6 +1098,14 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     a host system with the same SELinux policy as the image you're
     building.
 
+`MachineId=`, `--machine-id=`
+
+:  Takes a UUID or the special value `random`. Sets the machine ID of the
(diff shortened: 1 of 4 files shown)


mkosi/resources/man/mkosi.md
```diff
@@ -1096,6 +1098,14 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     a host system with the same SELinux policy as the image you're
     building.
 
+`MachineId=`, `--machine-id=`
+
+:  Takes a UUID or the special value `random`. Sets the machine ID of the
+   image to the specified UUID. If set to `random`, a random UUID will be
+   written to `/etc/machine-id`. If not specified explicitly and the file
+   `mkosi.machine-id` exists in the local directory, the UUID to use it
```
behrmann, line 1106:
```suggestion
   `mkosi.machine-id` exists in the local directory, the UUID to use is
```

behrmann (approved):

PR #3034, opened 2024-09-17
Make sure we only use workdir() for keys if they're a path on disk

### LINKED ISSUE #3033: Can't pass pkcs11 url to sbsign

main (055ae7877465be4e9b2dce648ffc1d89963da45c)

### Used host distribution

Ubuntu Noble (using systemd built from 6c8e9ba2f0adb7a66fe520a3f58f36d69bc8db43)

### Used target distribution

Ubuntu Noble

### Linux kernel version used

6.8.0-41-generic

x86_64

I'm trying to use a Secure Boot key that is contained in a pkcs11 device. My Secure Boot configuration looks like:

```
SecureBoot=yes
SecureBootKey=pkcs11:object=sb-key
SecureBootCertificate=../cert.pem
SecureBootKeySource=engine:pkcs11
```

I expect mkosi to use `SecureBootKey` directly, but it is turning it into an invalid path in my work directory before passing it to `sbsign`.

### Used mkosi config

```ini
SecureBoot=yes
SecureBootKey=pkcs11:object=sb-key
SecureBootCertificate=../cert.pem
SecureBootKeySource=engine:pkcs11
```

### mkosi output

```sh
‣  Signing systemd-boot binaries…
‣ + sbsign --key /work/pkcs11:object=sb-key --cert /work/home/pwicks/Projects/test/os/cert.pem --output /work/var/tmp/mkosi-workspace-pl32mq2_/root/usr/lib/systemd/boot/efi/systemd-bootx64.efi.signed --engine pkcs11 /work/var/tmp/mkosi-workspace-pl32mq2_/root/usr/lib/systemd/boot/efi/systemd-bootx64.efi
Format not recognized!
The private key ID is not a valid PKCS#11 URI
The PKCS#11 URI format is defined by RFC7512
The legacy ENGINE_pkcs11 ID format is also still accepted for now
Format not recognized!
The private key ID is not a valid PKCS#11 URI
The PKCS#11 URI format is defined by RFC7512
The legacy ENGINE_pkcs11 ID format is also still accepted for now
The private key was not found at: /work/pkcs11:object=sb-key
PKCS11_get_private_key returned NULL
```

DaanDeMeyer:
Fixes #3033

### DIFF
--- a/mkosi/bootloader.py
+++ b/mkosi/bootloader.py
@@ -506,7 +506,6 @@ def sign_efi_binary(context: Context, input: Path, output: Path) -> Path:
     ):
         cmd: list[PathString] = [
             "sbsign",
-            "--key", workdir(context.config.secure_boot_key),
             "--cert", workdir(context.config.secure_boot_certificate),
             "--output", workdir(output),
         ]
@@ -518,7 +517,10 @@ def sign_efi_binary(context: Context, input: Path, output: Path) -> Path:
         if context.config.secure_boot_key_source.type == KeySourceType.engine:
             cmd += ["--engine", context.config.secure_boot_key_source.source]
         if context.config.secure_boot_key.exists():
+            cmd += ["--key", workdir(context.config.secure_boot_key)]
             options += ["--ro-bind", context.config.secure_boot_key, workdir(context.config.secure_boot_key)]
+        else:
+            cmd += ["--key", workdir(context.config.secure_boot_key)]
         cmd += [workdir(input)]
         run(
             cmd,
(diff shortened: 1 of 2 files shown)


DaanDeMeyer:
@pwicks86 can you give this a try?

bluca (approved):

mkosi/bootloader.py
```diff
@@ -518,7 +517,10 @@ def sign_efi_binary(context: Context, input: Path, output: Path) -> Path:
         if context.config.secure_boot_key_source.type == KeySourceType.engine:
             cmd += ["--engine", context.config.secure_boot_key_source.source]
         if context.config.secure_boot_key.exists():
+            cmd += ["--key", workdir(context.config.secure_boot_key)]
             options += ["--ro-bind", context.config.secure_boot_key, workdir(context.config.secure_boot_key)]
+        else:
+            cmd += ["--key", workdir(context.config.secure_boot_key)]
```
pwicks86, line 523: I think this needs to be `cmd += ["--key", context.config.secure_boot_key]` here.

pwicks86:
> @pwicks86 can you give this a try?

I think there's a typo here, line 523 in bootloader.py shouldn't have the call to `workdir`. If i remove that then `sbsign` is correctly talking to my pkcs11 provider.

pwicks86:
There's another related problem here which is that pkcs11 urls as defined by RFC7512 may have `;` characters (e.g.  `pkcs11:object=my-sign-key;type=private`), but the config file parser sees these as comment characters and so only the first part of a given url is passed through to `sbsign`.

PR #3040, opened 2024-09-17
Fixups for gpg

DaanDeMeyer:


dvzrv:
Thanks! These changes allow for successfully signing with a hardware token (Nitrokey in my case)!

PR #3045, opened 2024-09-18
Add mkosi.*.d for automatic script discovery

behrmann:
This adds `mkosi.$SCRIPT.d` from which scripts are picked up automatically. Recently I switched a project from using a single `mkosi.build` to using multiple scripts and that was a lot of churn, not only because of the renaming of files, but also because of a lot of repetitive adding of scripts to the config files. Picking this up from a well-known directory makes this more ergonomic.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -557,6 +557,36 @@ def parse_path(
     return path
 
 
+def parse_paths_from_directory(
+    value: str,
+    *,
+    required: bool = True,
+    resolve: bool = True,
+    expanduser: bool = True,
+    expandvars: bool = True,
+    secret: bool = False,
+    absolute: bool = False,
+    constants: Sequence[str] = (),
+) -> list[Path]:
+    path = parse_path(
+        value,
+        required=required,
+        resolve=resolve,
+        expanduser=expanduser,
+        expandvars=expandvars,
+        secret=secret,
+        absolute=absolute,
+        constants=constants,
+    )
+    if not path.exists():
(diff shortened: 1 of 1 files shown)


DaanDeMeyer (changes requested):
We should enforce that it's a directory and fail if it's not

mkosi/config.py
```diff
@@ -557,6 +557,36 @@ def parse_path(
     return path
 
 
+def parse_paths_and_dirs(
```
DaanDeMeyer, line 560: This does not seem appropriately named anymore. `parse_paths_from_directory()` maybe?

mkosi/config.py
```diff
@@ -3846,6 +3884,26 @@ def parse_config_one(self, path: Path, parse_profiles: bool = False, parse_local
...
+                        expanduser=False,
+                        expandvars=False,
+                    )
+                    for e in recursive_extras:
+                        if e.exists():
+                            setattr(
+                                self.config,
+                                s.dest,
+                                s.parse(
+                                    file_run_or_read(e).rstrip("\n") if s.path_read_text else os.fspath(e),
```
DaanDeMeyer, line 3902: We should never use `path_read_text` when more than one path can be specified so using `file_run_or_read()` on paths from a directory should never happen in the first place

DaanDeMeyer (changes requested):
Needs documentation updates

DaanDeMeyer (approved):

PR #3046, opened 2024-09-19
docs: Add doc on disk images with verity protected root partition

DaanDeMeyer:


docs/root-verity.md
```diff
@@ -0,0 +1,185 @@
...
+Label=%M_%A_verity_sig
+SplitName=%t.%U
+Type=root-verity-sig
+Verity=signature
+VerityMatchKey=root
+```
+
+Then, you'll need a dropin for systemd-repart in the initrd to make sure
+it runs after the root partition has been mounted, so let's create an
+initrd with mkosi.images where we customize systemd-repart to behave
```
behrmann, line 47:
```suggestion
initrd with `mkosi.images` where we customize systemd-repart to behave
```

docs/root-verity.md
```diff
@@ -0,0 +1,185 @@
...
+
+# mkosi.extra/usr/lib/repart.d/40-var.conf
+[Partition]
+Type=var
+Format=ext4
+Encrypt=tpm2
+SizeMinBytes=2G
+```
+
+Because in this setup /etc is immutable, we have to embed the machine ID
```
behrmann, line 119:
```suggestion
Because in this setup `/etc` is immutable, we have to embed the machine ID
```

docs/root-verity.md
```diff
@@ -0,0 +1,185 @@
...
+in the image itself at build time so generate a machine ID and persist
```
behrmann, line 120:
```suggestion
in the image itself at build time to generate a machine ID and persist
```

behrmann (approved):

PR #3049, opened 2024-09-19
Don't pick up definitions from the host when running burn

DaanDeMeyer:


behrmann (approved):
This should maybe be fixed in repart by adding `--dont-read-system-definitions` switch, because this solution is a bit unintuitive.

PR #3053, opened 2024-09-20
Add `OutputMode=` option

aafeijoo-suse:
Also, set output mode `600` by default to initrds generated using `mkosi-initrd`. Nowadays this may not make any sense, but it caused a  couple of CVEs in the past (CVE-2016-8637, CVE-2012-4453).


mkosi/config.py
```diff
@@ -682,6 +682,24 @@ def config_parse_compress_level(value: Optional[str], old: Optional[int]) -> Opt
     return level
 
 
+def config_parse_mode(value: Optional[str], old: Optional[int]) -> Optional[int]:
+    if not value:
+        return None
+
+    try:
+        mode = int(value, 8)
```
behrmann, line 690: Let's be nice here
```suggestion
        mode = int(value, base=8)
```

mkosi/config.py
```diff
@@ -4190,6 +4220,14 @@ def format_bytes_or_none(num_bytes: Optional[int]) -> str:
     return format_bytes(num_bytes) if num_bytes is not None else "none"
 
 
+def format_octal(oct_value: int) -> str:
+    return oct(oct_value).replace("0o","")
```
behrmann, line 4224:
```suggestion
    return f"{oct_value:o}
```
or better even
```suggestion
    return f"{oct_value:>04o}
```

behrmann (commented):
I don't think it hurts to add this, but if this is an issue, we should probably create the files with those permission already by setting the umask instead of chmoding it later. 

I understand the rationale of the linked CVEs, but file permissions are at the very best a fig leaf for something in `/boot` and if this ends up on an ESP or an XBOOTLDR partition, then those will be FAT and that doesn't support permissions anyway, does it?

aafeijoo-suse:
> I don't think it hurts to add this, but if this is an issue, we should probably create the files with those permission already by setting the umask instead of chmoding it later.

That was my first approach, but I had always `-x` for all u, g and o, not sure what causes this.

> I understand the rationale of the linked CVEs, but file permissions are at the very best a fig leaf for something in `/boot` and if this ends up on an ESP or an XBOOTLDR partition, then those will be FAT and that doesn't support permissions anyway, does it?

That's why I said that it doesn't make a lot of sense nowadays, but I would like to support old systems too.

mkosi/config.py
```diff
@@ -2074,6 +2093,17 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
         help="Output directory",
         scope=SettingScope.universal,
     ),
+    ConfigSetting(
+        dest="output_mode",
+        short="-M",
+        metavar="MODE",
+        name="OutputMode",
```
DaanDeMeyer, line 2100: Can be dropped, is automatically inferred

mkosi/config.py
```diff
@@ -2074,6 +2093,17 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
         help="Output directory",
         scope=SettingScope.universal,
     ),
+    ConfigSetting(
+        dest="output_mode",
+        short="-M",
```
DaanDeMeyer, line 2098: I don't think this needs one of our precious short option letters

mkosi/config.py
```diff
@@ -2074,6 +2093,17 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
...
+        metavar="MODE",
+        name="OutputMode",
+        section="Output",
+        specifier="M",
```
DaanDeMeyer, line 2102: I don't think this needs a specifier either, do you have a use case for this?

mkosi/config.py
```diff
@@ -2074,6 +2093,17 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
...
+        parse=config_parse_mode,
+        help="Output mode",
```
DaanDeMeyer, line 2104: This could be a bit more descriptive

mkosi/initrd.py
```diff
@@ -111,6 +111,8 @@ def main() -> None:
             "--package-cache-dir=/var",
             "--cache-only=metadata",
         ]
+        if args.format != str(OutputFormat.directory):
```
DaanDeMeyer, line 114: Please use `OutputFormat.directory.name`

mkosi/resources/man/mkosi.md
```diff
@@ -562,6 +562,11 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     not specified and the directory `mkosi.output/` exists in the local
     directory, it is automatically used for this purpose.
 
+`OutputMode=`, `--output-mode=`, `-M`
+:   File system access mode used when creating the output image file or
+    directory. Takes an access mode in octal notation. If not set, uses the
```
DaanDeMeyer, line 567: Using this for the directory seems wrong. When we're building a directory image this setting should not apply since it will directly affect the image itself as well

mkosi/__init__.py
```diff
@@ -3302,6 +3302,12 @@ def build_image(context: Context) -> None:
         output_base.unlink(missing_ok=True)
         output_base.symlink_to(context.config.output_with_compression)
 
+    if context.config.output_mode is not None and context.config.output_format != OutputFormat.directory:
+        os.chmod(output_base, context.config.output_mode)
+        if context.config.output_format in (OutputFormat.uki, OutputFormat.esp):
+            os.chmod(context.staging / context.config.output_split_kernel, context.config.output_mode)
+            os.chmod(context.staging / context.config.output_split_initrd, context.config.output_mode)
```
DaanDeMeyer, line 3309: We should blanket apply it to every output I think so I'd do this differently. In `finalize_staging()`, before we copy a file to the output directory, make sure it has the requested output mode.

aafeijoo-suse: ok, PTAL

DaanDeMeyer (approved):

behrmann (approved):

PR #3056, opened 2024-09-20
Use run0 instead of refusing to run commands that need root unprivileged

DaanDeMeyer:


mkosi/__init__.py
```diff
@@ -3637,7 +3642,7 @@ def run_systemd_tool(tool: str, args: Args, config: Config) -> None:
             f"Output {config.output_dir_or_cwd() / config.output} does not exist, cannot inspect with {tool}"
         )
 
-    cmd: list[PathString] = [tool_path]
+    cmd += [tool_path]
```
behrmann, line 3645: Is this maybe a holdover from prepending `run0`?

behrmann (approved):
LGTM, though the idiom `["run0"] if os.getuid() != 0 else []` would make for a good function in `util.py`.

PR #3057, opened 2024-09-21
Parse profiles after mkosi.conf.d

DaanDeMeyer:
Currently profiles can't depend on any of the configuration set in mkosi.conf.d as they are parsed before mkosi.conf.d is parsed. Let's parse the profile related configuration last instead so it can match on all the configuration set in mkosi.conf.d.

To set the distribution and release and such based on the profile, a dropin in mkosi.conf.d can match on the configured profile instead.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -3782,21 +3782,26 @@ class ParseContext:
         extras = path.is_dir()
 
         if path.is_dir():
-            path = path / "mkosi.conf"
+            path /= "mkosi.conf"
 
         if not self.match_config(path):
             return False
 
         if extras:
-            if local and (path.parent / "mkosi.local.conf").exists():
-                self.parse_config_one(path.parent / "mkosi.local.conf")
+            if local:
+                if (
+                    (localpath := path.parent / "mkosi.local/mkosi.conf").exists()
+                    or (localpath := path.parent / "mkosi.local.conf").exists()
+                ):  # fmt: skip
+                    self.parse_config_one(localpath)
 
-                # Configuration from mkosi.local.conf should override other file based configuration but not
-                # the CLI itself so move the finalized values to the CLI namespace.
-                for s in SETTINGS:
-                    if hasattr(self.config, s.dest):
-                        setattr(self.cli, s.dest, self.finalize_value(s))
-                        delattr(self.config, s.dest)
+                    # Local configuration should override other file based
+                    # configuration but not the CLI itself so move the finalized
+                    # values to the CLI namespace.
+                    for s in SETTINGS:
+                        if hasattr(self.config, s.dest):
+                            setattr(self.cli, s.dest, self.finalize_value(s))
+                            delattr(self.config, s.dest)
 
             for s in SETTINGS:
                 if (
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -3866,28 +3866,30 @@ class ParseContext:
                 setattr(self.config, s.dest, s.parse(v, getattr(self.config, s.dest, None)))
                 self.parse_new_includes()
 
+        profilepath = None
         if profiles:
             profile = self.finalize_value(SETTINGS_LOOKUP_BY_DEST["profile"])
             self.immutable.add("Profile")
 
             if profile:
-                for p in (profile, f"{profile}.conf"):
-                    p = Path("mkosi.profiles") / p
-                    if p.exists():
+                for p in (Path(profile), Path(f"{profile}.conf")):
+                    profilepath = Path("mkosi.profiles") / p
+                    if profilepath.exists():
                         break
                 else:
                     die(f"Profile '{profile}' not found in mkosi.profiles/")
 
                 setattr(self.config, "profile", profile)
 
-                with chdir(p if p.is_dir() else Path.cwd()):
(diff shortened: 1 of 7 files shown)


DaanDeMeyer:
Added the PR to read mkosi.local on top of this one as it depends on some stuff from the profile one

mkosi/config.py
```diff
@@ -3785,21 +3785,33 @@ def parse_config_one(self, path: Path, profiles: bool = False, local: bool = Fal
         extras = path.is_dir()
 
         if path.is_dir():
-            path = path / "mkosi.conf"
+            path /= "mkosi.conf"
```
behrmann, line 3788: I generally like in place operators, but for paths I think their readability is not really great.

mkosi/config.py
```diff
@@ -3869,29 +3881,31 @@ def parse_config_one(self, path: Path, profiles: bool = False, local: bool = Fal
...
             if profile:
-                for p in (profile, f"{profile}.conf"):
-                    p = Path("mkosi.profiles") / p
-                    if p.exists():
+                for p in (Path(profile), Path(f"{profile}.conf")):
+                    profilepath = Path("mkosi.profiles") / p
+                    if profilepath.exists():
                         break
                 else:
                     die(f"Profile '{profile}' not found in mkosi.profiles/")
```
behrmann, line 3895:
```suggestion
                if not any((Path("mkosi.profiles") / p).exists() for p in (Path(profile), Path(f"{profile}.conf")))
                    die(f"Profile '{profile}' not found in mkosi.profiles/")
```

mkosi/config.py
```diff
@@ -3785,21 +3785,33 @@ def parse_config_one(self, path: Path, profiles: bool = False, local: bool = Fal
...
+                    localpath = path.parent / p
 
-                # Configuration from mkosi.local.conf should override other file based configuration but not
-                # the CLI itself so move the finalized values to the CLI namespace.
-                for s in SETTINGS:
-                    if hasattr(self.config, s.dest):
-                        setattr(self.cli, s.dest, self.finalize_value(s))
-                        delattr(self.config, s.dest)
+                    if localpath.is_dir():
+                        localpath /= "mkosi.conf"
```
behrmann, line 3799: Looking at the code both `mkosi.local` and `mkosi.local.conf` can be either a file or directory, that seems a bit redundant. I don't think I understand the rationale.

DaanDeMeyer: Huh right let me fix that

mkosi/config.py
DaanDeMeyer: Does this work? We still need the profile path later.

mkosi/config.py
DaanDeMeyer: Ehh we use them in other places already as well, let's keep em for now

behrmann (approved):

mkosi/resources/man/mkosi.md
```diff
@@ -2169,7 +2169,8 @@ Scripts executed by mkosi receive the following environment variables:
 * `$DISTRIBUTION_ARCHITECTURE` contains the architecture from
   `$ARCHITECTURE` in the format used by the configured distribution.
 
-* `$PROFILE` contains the profile from the `Profile=` setting.
+* `$PROFILES` contains the profiles from the `Profiles=` setting as a
```
behrmann, line 2172: That also needs to be changed in the table.

behrmann (changes requested):

behrmann:
LGTM otherwise, but the change from `PROFILE` to `PROFILES` needs to be mentioned in the news otherwise this might be a too large footgun.

mkosi/config.py
```diff
@@ -3777,26 +3781,31 @@ def match_config(self, path: Path) -> bool:
...
-                # Configuration from mkosi.local.conf should override other file based configuration but not
-                # the CLI itself so move the finalized values to the CLI namespace.
-                for s in SETTINGS:
-                    if hasattr(self.config, s.dest):
-                        setattr(self.cli, s.dest, self.finalize_value(s))
-                        delattr(self.config, s.dest)
+            if parse_local:
+                if (
+                    (localpath := path.parent / "mkosi.local/mkosi.conf").exists() or
+                    (localpath := path.parent / "mkosi.local.conf").exists()
```
behrmann, line 3798: I assume you want the alignment, but ruff following PEP8 would put it like this
```suggestion
                    (localpath := path.parent / "mkosi.local/mkosi.conf").exists()
                    or (localpath := path.parent / "mkosi.local.conf").exists()
```

mkosi/resources/man/mkosi.md
```diff
@@ -299,28 +299,29 @@ grouped by section below.
 Configuration is parsed in the following order:
 
 * The command line arguments are parsed.
-* `mkosi.local.conf` is parsed if it exists. This file should be in
-  `.gitignore` (or equivalent) and is intended for local configuration.
+* `mkosi.local` or `mkosi.local.conf` is parsed if it exists. This file or
```
behrmann, line 302: This way file and directory are symmetric with the next sentence.
```suggestion
* `mkosi.local.conf` or `mkosi.local` is parsed if it exists. This file or
```

PR #3060, opened 2024-09-21
Add Incremental=strict

DaanDeMeyer:
This ensures cached images already exist and fails the build if they
don't exist yet.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -4224,7 +4224,13 @@ def run_verb(args: Args, images: Sequence[Config], *, resources: Path) -> None:
 
     output = last.output_dir_or_cwd() / last.output_with_compression
 
-    if args.verb == Verb.build and not args.force and output.exists() and not output.is_symlink():
+    if (
+        args.verb == Verb.build
+        and not args.force
+        and output.exists()
+        and not output.is_symlink()
+        and last.output_format != OutputFormat.none
+    ):
         logging.info(f"Output path {output} exists already. (Use --force to rebuild.)")
         return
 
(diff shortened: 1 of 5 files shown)


mkosi/__init__.py
```diff
@@ -4236,13 +4244,32 @@ def run_verb(args: Args, images: Sequence[Config], *, resources: Path) -> None:
 
     check_workspace_directory(last)
 
+    if last.incremental == Incremental.strict:
+        if args.force > 1:
+            die("Cannot remove incremental caches when building with Incremental=strict")
```
behrmann, line 4249: The usability here seems a bit weird, since if `Incremental=strict` is set in a config file, `mkosi -ff` doesn't work anymore without editing the file or setting `-i no`. Would doing
```suggestion
        if 1 < args.force < 3:
            die("Cannot remove incremental caches when building with Incremental=strict")
```
make sense here? Maybe something like `hint="Use --incremental=no"` would be enough.

mkosi/resources/man/mkosi.md
```diff
@@ -1309,16 +1309,20 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
-    cached image is first removed and then re-created.
+:   Takes either `strict` or a boolean value as its argument. Enables
+    incremental build mode. In this mode, a copy of the OS image is created
+    immediately after all OS packages are installed and the prepare scripts
+    have executed but before the `mkosi.build` scripts are invoked (or
+    anything that happens after it). On subsequent invocations of `mkosi`
+    with the `-i` switch this cached image may be used to skip the OS package
+    installation, thus drastically speeding up repetitive build times. Note
+    that while there is some rudimentary cache invalidation, it is definitely
+    not perfect. In order to forcerebuilding of the cached image, combine
```
behrmann, line 1320:
```suggestion
    not perfect. In order to force rebuilding of the cached image, combine
```

behrmann (approved):

mkosi/__init__.py
DaanDeMeyer: I think a hint is the way to go

PR #3061, opened 2024-09-21
docs: Fix partition order for root verity docs

DaanDeMeyer:
The generated root partition needs to be resized on first boot for the A/B update setup so we have to make sure it's the last one and give it the same size requirements as the B partition in the image so it's grown to the same size.

While we're at it we also reorder the settings a bit so things are more consistent.

### DIFF
--- a/docs/root-verity.md
+++ b/docs/root-verity.md
@@ -6,40 +6,42 @@ partition, put the following in mkosi.repart:
 ```conf
 # mkosi.repart/00-esp.conf
 [Partition]
+Type=esp
+Format=vfat
 CopyFiles=/efi:/
 CopyFiles=/boot:/
-Format=vfat
-SizeMinBytes=1024M
-Type=esp
+SizeMinBytes=1G
+SizeMaxBytes=1G
 
-# mkosi.repart/10-root.conf
+# mkosi.repart/10-root-verity-sig.conf
 [Partition]
-CopyFiles=/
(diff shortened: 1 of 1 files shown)


docs/root-verity.md
```diff
@@ -6,40 +6,41 @@ partition, put the following in mkosi.repart:
...
 [Partition]
+Type=esp
+Format=vfat
 CopyFiles=/efi:/
 CopyFiles=/boot:/
-Format=vfat
-SizeMinBytes=1024M
-Type=esp
+SizeMinBytes=1G
+SizeMinBytes=1G
```
behrmann, line 14: This line is doubled, I assume you mean
```suggestion
SizeMaxBytes=1G
SizeMinBytes=1G
```

docs/root-verity.md
```diff
@@ -68,43 +69,47 @@ existing partitions:
...
+SizeMaxBytes=2G
+SizeMinBytes=2G
 
 # mkosi.extra/usr/lib/repart.d/30-swap.conf
 [Partition]
 Type=swap
+Format=swap
+Encrypt=tpm2
 SizeMinBytes=4G
 SizeMaxBytes=4G
```
behrmann, line 114: Let's switch this around for consistency, too
```suggestion
SizeMaxBytes=4G
SizeMinBytes=4G
```

PR #3063, opened 2024-09-23
Fix signing with engine

DaanDeMeyer:
We need to make sure /run/pcscd/pcscd.comm is available to make
signing work so let's mount /run/pcscd into the sandbox every time
we're going to sign something if it exists.


behrmann (approved):
ruff's not happy about the first commit, otherwise LGTM.

PR #3066, opened 2024-09-25
Don't explicitly bind mount in keyrings from host into sandbox

DaanDeMeyer:
These are automatically mounted in as part of the tools tree. This change means they'll always be picked up from the tools tree and won't take ToolsTreeCertificates= into account anymore, but then keyrings weren't exactly certificates in the first place, and can still be picked up from the host by using SandboxTrees=.

This allows users to provide their own keyrings using sandbox trees. Currently we just overmount the user provided ones with the ones from the host.

### DIFF
--- a/mkosi/mounts.py
+++ b/mkosi/mounts.py
@@ -92,17 +92,18 @@ def finalize_crypto_mounts(config: Config) -> list[PathString]:
     mounts = [
         (root / subdir, Path("/") / subdir)
         for subdir in (
-            Path("usr/share/keyrings"),
-            Path("usr/share/distribution-gpg-keys"),
             Path("etc/pki"),
             Path("etc/ssl"),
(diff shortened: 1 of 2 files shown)


behrmann (approved):
Makes sense, that `/usr/share/keyrings` from the tools tree should win vs the sandbox tree is non-obvious for users.

mkosi/mounts.py
```diff
@@ -92,8 +92,6 @@ def finalize_crypto_mounts(config: Config) -> list[PathString]:
     mounts = [
         (root / subdir, Path("/") / subdir)
         for subdir in (
-            Path("usr/share/keyrings"),
```
behrmann, line 95: Looking at the docs for `ToolsTreeCertificates=` this is explicitly mentioned there, so that needs to be changed.

Issue #3054, opened 2024-09-20
Mkosi 25 is not compatible with linux-hardened


dvzrv:
### mkosi commit the issue has been seen with

158c2d1fd221e7dd689ea14a1b32617b22764b43

### Used host distribution

Arch Linux

### Linux kernel version used

6.10.10-hardened1-1-hardened

### CPU architectures issue was seen on

x86_64

### Unexpected behaviour you saw

When trying to use this repository's `mkosi` to run `mkosi build` on the setup available in this repository, it fails on [linux-hardened](https://github.com/anthraxx/linux-hardened) (user namespaces are enabled using `sysctl kernel.unprivileged_userns_clone=1`).

cc @anthraxx

### Used mkosi config

```ini
# SPDX-License-Identifier: LGPL-2.1-or-later
[Build]
CacheDirectory=mkosi.cache
History=yes

[Output]
# These images are (among other things) used for running mkosi which means we need some disk space available so
# default to directory output where disk space isn't a problem.
Format=directory
OutputDirectory=mkosi.output

[Content]
Autologin=yes
SELinuxRelabel=no
ShimBootloader=unsigned
BuildSources=.
BuildSourcesEphemeral=yes

Packages=
        bash
        binutils
        gdb
        less
        nano
        strace
        sudo
        systemd
        tmux
        udev
        zsh

InitrdPackages=
        less

RemoveFiles=
        # The grub install plugin doesn't play nice with booting from virtiofs.
        /usr/lib/kernel/install.d/20-grub.install
        # The dracut install plugin doesn't honor KERNEL_INSTALL_INITRD_GENERATOR.
        /usr/lib/kernel/install.d/50-dracut.install

# Make sure that SELinux doesn't run in enforcing mode even if it's pulled in as a dependency.
KernelCommandLine=
        enforcing=0
        systemd.log_ratelimit_kmsg=0
        systemd.crash_shell
        printk.devkmsg=on
        systemd.early_core_pattern=/core

KernelModulesInitrdExclude=.*
KernelModulesInitrdInclude=default

[Host]
QemuMem=4G
```

### mkosi output

```sh
PATH="/home/user/Downloads/mkosi/bin:$PATH" mkosi build
‣ Syncing package manager metadata
Traceback (most recent call last):
  File "/home/user/Downloads/mkosi/mkosi/sandbox.py", line 828, in <module>
  File "/home/user/Downloads/mkosi/mkosi/sandbox.py", line 793, in main
[...]

PR #3073, opened 2024-09-27
Don't use overlayfs for /etc and /opt in sandbox

DaanDeMeyer:
Unprivileged overlayfs isn't available everywhere (see #3054). So let's try to accomodate this a little by not using overlayfs for /etc and /opt from the sandbox tree and instead mounting them read-only into the sandbox. If required, scripts can still mount an overlayfs onto these if needed, we just don't do it by default anymore.

This does mean we need to set up /etc with mountpoints and symlinks beforehand in install_sandbox_trees(), but this shouldn't be a huge problem.


keszybz:
LGTM. CI failure is unrelated.

PR #3074, opened 2024-09-28
Add support for building UKI addons

NekkoDroid:
Not something I personally need anymore (so feel free to close), but I already implemented it and thought this might be useful to some.

This basically just reuses `ukify`s ability to take config files in which you can configure whatever (e.g. `Cmdline=`, `DeviceTree=`, ...) and passes the output target and secure boot parameters when needed. This should allow additions to `ukify` without needing to update the code here.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1901,6 +1901,45 @@ def install_uki(
             f.write("fi\n")
 
 
+def build_pe_addon(context: Context, arch: str, stub: Path, config: Path, output: Path) -> None:
+    arguments: list[PathString] = [
+        "--config", config,
+    ]  # fmt: skip
+
+    options: list[PathString] = [
+        "--ro-bind", config, config,
+    ]  # fmt: skip
+
+    with complete_step(f"Generating PE addon /{output.relative_to(context.root)}"):
+        run_ukify(context, arch, stub, output, arguments, options)
+
+
+def install_pe_addons(context: Context) -> None:
+    if not context.config.pe_addons:
+        return
+
+    if not (arch := context.config.architecture.to_efi()):
+        die(f"Architecture {context.config.architecture} does not support UEFI")
+
+    stub = systemd_addon_stub_binary(context)
+    if not stub.exists():
+        die(f"sd-stub not found at /{stub.relative_to(context.root)} in the image")
+
+    addon_dir = context.root / "boot/loader/addons"
+    with umask(~0o755):
+        addon_dir.mkdir(parents=True, exist_ok=True)
+
+    for addon in context.config.pe_addons:
+        output = addon_dir / addon.with_suffix(".addon.efi").name
+        build_pe_addon(context, arch, stub, config=addon, output=output)
+
+
+def systemd_addon_stub_binary(context: Context) -> Path:
+    arch = context.config.architecture.to_efi()
+    stub = context.root / f"usr/lib/systemd/boot/efi/addon{arch}.efi.stub"
+    return stub
+
+
 def install_kernel(context: Context, partitions: Sequence[Partition]) -> None:
     # Iterates through all kernel versions included in the image and generates a combined
     # kernel+initrd+cmdline+osrelease EFI file from it and places it in the /EFI/Linux directory of
@@ -3369,6 +3408,7 @@ def build_image(context: Context) -> None:
         install_systemd_boot(context)
         install_grub(context)
         install_shim(context)
+        install_pe_addons(context)
         run_sysusers(context)
         run_tmpfiles(context)
         run_preset(context)
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1442,49 +1442,36 @@ def want_signed_pcrs(config: Config) -> bool:
     )
 
 
-def build_uki(
+def run_ukify(
     context: Context,
+    arch: str,
     stub: Path,
-    kver: str,
-    kimg: Path,
-    microcodes: list[Path],
-    initrds: list[Path],
-    cmdline: Sequence[str],
     output: Path,
+    arguments: list[PathString],
+    options: list[PathString],
 ) -> None:
-    # Older versions of systemd-stub expect the cmdline section to be null terminated. We can't
-    # embed NUL terminators in argv so let's communicate the cmdline via a file instead.
-    (context.workspace / "cmdline").write_text(f"{' '.join(cmdline).strip()}\x00")
-
-    if not (arch := context.config.architecture.to_efi()):
-        die(f"Architecture {context.config.architecture} does not support UEFI")
-
-    if not (ukify := context.config.find_binary("ukify", "/usr/lib/systemd/ukify")):
+    ukify = context.config.find_binary("ukify", "/usr/lib/systemd/ukify")
+    if not ukify:
         die("Could not find ukify")
 
-    cmd: list[PathString] = [
+    cmd = [
         python_binary(context.config, binary=ukify),
         ukify,
-        *(["--cmdline", f"@{context.workspace / 'cmdline'}"] if cmdline else []),
-        "--os-release", f"@{context.root / 'usr/lib/os-release'}",
+        "build",
+        *arguments,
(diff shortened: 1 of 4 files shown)


mkosi/resources/man/mkosi.md
```diff
@@ -984,6 +984,14 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     | `&h`      | `roothash=` or `usrhash=` value of kernel argument |
     | `&c`      | Number of tries used for boot attempt counting     |
 
+`UnifiedKernelImageAddons=`, `--unified-kernel-image-addon`
+:   Build additional UKI addons. Takes a comma separated list of paths to
+    `ukify` config files. This option may be used multiple times in which case
+    each config gets built into a correspond addon. Each addon has the name
```
behrmann, line 990:
```suggestion
    each config gets built into a corresponding addon. Each addon has the name
```

mkosi/__init__.py
```diff
@@ -1442,6 +1442,38 @@ def want_signed_pcrs(config: Config) -> bool:
     )
 
 
+def add_ukify_secure_boot_arguments(
+    context: Context, out_cmd: list[PathString], out_options: list[PathString]
+) -> None:
```
behrmann, line 1447: Mutating input arguments makes things harder to reason about. I'd prefer this to return the extra cmd argument and options and adding them where they are used.

mkosi/__init__.py
```diff
@@ -1442,6 +1442,38 @@ def want_signed_pcrs(config: Config) -> bool:
...
+            "--secureboot-private-key", context.config.secure_boot_key,
+            "--secureboot-certificate", context.config.secure_boot_certificate,
+        ]  # fmt: skip
+        out_options += [
+            "--ro-bind", context.config.secure_boot_certificate, context.config.secure_boot_certificate,
+        ]  # fmt: skip
+        if context.config.secure_boot_key_source.type == KeySourceType.engine:
+            out_cmd += ["--signing-engine", context.config.secure_boot_key_source.source]
+            out_options += ["--bind-try", "/run/pcscd", "/run/pcscd"]
+        if context.config.secure_boot_key.exists():
```
behrmann, line 1463: Does it make sense to check the existence at this point? If the key is set and the key doesn't exist, then suppressing that seems like either a footgun or bound to blow up a bit later?

NekkoDroid: I just copied that code out to a separate function, I don't know the reasoning behind this.

DaanDeMeyer: The exists is to check if it's a pkcs11 uri or an actual file on disk

mkosi/__init__.py
```diff
@@ -1442,6 +1442,41 @@ def want_signed_pcrs(config: Config) -> bool:
     )
 
 
+def get_ukify_secure_boot_arguments(context: Context) -> tuple[list[PathString], list[PathString]]:
```
DaanDeMeyer, line 1445: This feels messy

Can we instead do a `ukify()` function that takes a `Context` parameter as well as extra arguments and options and all it does is add architecture and secure boot stuff before invoking ukify?

DaanDeMeyer (changes requested):

mkosi/__init__.py
```diff
@@ -1875,6 +1902,35 @@ def install_uki(
             f.write("fi\n")
 
 
+def build_uki_addon(context: Context, config: Path, output: Path) -> None:
+    cmd: list[PathString] = [
+        "build",
+        "--output", output,
```
DaanDeMeyer, line 1908: This can be moved into `run_ukify()`

mkosi/__init__.py
```diff
@@ -1875,6 +1902,35 @@ def install_uki(
...
+        "--config", config,
+    ]  # fmt: skip
+
+    options: list[PathString] = [
+        "--bind", output.parent, output.parent,
```
DaanDeMeyer, line 1913: This too

mkosi/__init__.py
```diff
@@ -1511,6 +1499,55 @@ def build_uki(
...
+    (context.workspace / "cmdline").write_text(f"{' '.join(cmdline).strip()}\x00")
+
+    if not (ukify := context.config.find_binary("ukify", "/usr/lib/systemd/ukify")):
+        die("Could not find ukify")
+
+    cmd: list[PathString] = [
+        "build",
+        *(["--cmdline", f"@{context.workspace / 'cmdline'}"] if cmdline else []),
+        "--os-release", f"@{context.root / 'usr/lib/os-release'}",
+        "--output", workdir(output),
```
DaanDeMeyer, line 1533: This can be moved into `run_ukify()`

mkosi/__init__.py
```diff
@@ -1511,6 +1499,55 @@ def build_uki(
...
+        "--uname", kver,
+        "--linux", kimg,
+    ]  # fmt: skip
+
+    options: list[PathString] = [
+        "--bind", output.parent, workdir(output.parent),
```
DaanDeMeyer, line 1539: And this too

DaanDeMeyer (changes requested):

mkosi/__init__.py
```diff
@@ -1511,6 +1499,55 @@ def build_uki(
...
+) -> None:
+    # Older versions of systemd-stub expect the cmdline section to be null terminated. We can't
+    # embed NUL terminators in argv so let's communicate the cmdline via a file instead.
+    (context.workspace / "cmdline").write_text(f"{' '.join(cmdline).strip()}\x00")
+
+    if not (ukify := context.config.find_binary("ukify", "/usr/lib/systemd/ukify")):
+        die("Could not find ukify")
+
+    cmd: list[PathString] = [
+        "build",
```
DaanDeMeyer, line 1530: Maybe move this into `run_ukify()` as well, I doubt we'll ever do something else than `build` and if needed we can just make it configurable

mkosi/__init__.py
```diff
@@ -1916,6 +1972,9 @@ def install_kernel(context: Context, partitions: Sequence[Partition]) -> None:
         if context.config.bootloader == Bootloader.uki:
             break
 
+    if want_uki(context):
+        install_uki_addons(context)
```
DaanDeMeyer, line 1976: I think we can do this earlier after `run_hwdb()` or so. I don't think we should gate this on `want_uki()`. If the addons are configured, let's generate them. That allows users to disable `Bootable=` but still generate addons in case they want to generate UKIs via their own logic

mkosi/config.py
```diff
@@ -2552,6 +2553,15 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
...
     ),
+    ConfigSetting(
+        dest="unified_kernel_image_addons",
+        long="--unified-kernel-image-addon",
+        metavar="PATH",
+        section="Content",
+        parse=config_make_list_parser(delimiter=",", parse=make_path_parser()),
+        recursive_paths=("mkosi.uki-addons/",),
+        help="Configuration files to generate UKI addons",
+    ),
```
DaanDeMeyer, line 2564: Should we name these PE addons? That's how they're named in the UAPI spec

DaanDeMeyer (approved):

mkosi/config.py
```diff
@@ -4479,6 +4489,7 @@ def summary(config: Config) -> str:
                     Shim Bootloader: {config.shim_bootloader}
               Unified Kernel Images: {config.unified_kernel_images}
         Unified Kernel Image Format: {config.unified_kernel_image_format}
+        Unified Kernel Image Addons: {line_join_list(config.pe_addons)}
```
behrmann, line 4492: A bit too late at this, but I think this should mention "PE" somehow, too, e.g. "Unified Kernel Image PE Addons", since the option is named `--pe-addon`, that is what I would grep for in the output of `mkosi summary`.

NekkoDroid: Woops, forgot to rename that. Probably should only be "PE Addons" since that is what they are referred to in the spec.

PR #3075, opened 2024-09-28
Clean cache directory in run_clean()

DaanDeMeyer:
Let's make sure we clean up the previous cache in run_clean() instead of in save_cache().


behrmann (approved):
Looking at `have_cache` and `not have_cache()` that's really an awful lot of double and triple negatives. :)

PR #3077, opened 2024-09-29
Fix `summary` name and reorder `PeAddons=`

NekkoDroid:


mkosi/config.py
```diff
@@ -4489,14 +4489,14 @@ def summary(config: Config) -> str:
                     Shim Bootloader: {config.shim_bootloader}
               Unified Kernel Images: {config.unified_kernel_images}
         Unified Kernel Image Format: {config.unified_kernel_image_format}
-        Unified Kernel Image Addons: {line_join_list(config.pe_addons)}
                             Initrds: {line_join_list(config.initrds)}
                     Initrd Packages: {line_join_list(config.initrd_packages)}
```
NekkoDroid, line 4493: Shouldn't these 2 options go down to the section with "Kernel Modules Initrd"? Seems a bit off to have them separate from all the other initrd options.

mkosi/config.py
behrmann, line 1: Nit: while GitHub does format the backticks in the commit message, I don't think we should do Markdown there.

behrmann (approved):

DaanDeMeyer (approved):

PR #3080, opened 2024-09-30
Don't create mountpoint in /var in sandbox

DaanDeMeyer:
/var will always be writable in the sandbox so no need to precreate the sandbox. This also saves us from using an overlayfs mount from /var in the sandbox as it will be completely empty now.


behrmann (approved):
LGTM, but let's wait for feedback from @dvzrv

dvzrv:
Thanks! That appears to work on linux-hardened as well! :tada:

Issue #3083, opened 2024-09-30
`--force` should not remove the output when it fails


aafeijoo-suse:
### mkosi commit the issue has been seen with

main

### Used host distribution

openSUSE Tumbleweed

### CPU architectures issue was seen on

None

### Unexpected behaviour you saw

Maybe for building OS images this is not critical, but for building initrds it does not look good... Maybe keep a backup copy of the current output image file or dir until mkosi exists with success, or delay removal until the end.

### mkosi output

```sh
# ls -l /boot/initrd-6.10.7-1-default
-rw------- 1 root root 29540805 Sep 30 16:10 /boot/initrd-6.10.7-1-default
# mkosi-initrd -O /boot -o initrd-6.10.7-1-default
‣ Removing output files of default image…
...
‣  Creating cpio archive /var/tmp/mkosi-workspace-0ihot0ak/staging/initrd-6.10.7-1-default.cpio…
‣  Compressing /var/tmp/mkosi-workspace-0ihot0ak/staging/initrd-6.10.7-1-default.cpio with zstd
‣ Could not rename /var/tmp/mkosi-workspace-0ihot0ak/staging/initrd-6.10.7-1-default.cpio.zst to /boot/initrd-6.10.7-1-default.cpio.zst as they are located on different devices, falling back to copying
cp: cannot create regular file '/work/boot/initrd-6.10.7-1-default.cpio.zst': Read-only file system
‣ "cp --recursive --no-dereference --preserve=mode,links,timestamps,ownership,xattr --reflink=auto --copy-contents /work/var/tmp/mkosi-workspace-0ihot0ak/staging/initrd-6.10.7-1-default.cpio.zst /work/boot/initrd-6.10.7-1-default.cpio.zst" returned non-zero exit code 1.
[...]

PR #3088, opened 2024-10-01
Don't remount directory read-only if output directory is located in it

DaanDeMeyer:
See #3083


aafeijoo-suse:
Tested and working. Thanks!

behrmann (approved):

PR #3091, opened 2024-10-01
Evaluate tools_tree earlier

DaanDeMeyer:
We now call have_cache() in run_clean() so let's make sure we evaluate the tools tree stuff earlier as have_cache() depends on it.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -1816,13 +1816,8 @@ class Config:
             "prepare_scripts": sorted(
                 base64.b64encode(script.read_bytes()).decode() for script in self.prepare_scripts
             ),
-            # We don't use the full path here since tests will often use temporary directories for the output
-            # directory which would trigger a rebuild every time.
-            "tools_tree": self.tools_tree.name if self.tools_tree else None,
-            "tools_tree_distribution": self.tools_tree_distribution,
-            "tools_tree_release": self.tools_tree_release,
-            "tools_tree_mirror": self.tools_tree_mirror,
-            "tools_tree_packages": sorted(self.tools_tree_packages),
+            # Statting the root directory of the tools tree isn't fool proof but should be good enough.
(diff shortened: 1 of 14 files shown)


mkosi/config.py
```diff
@@ -1817,13 +1819,8 @@ def cache_manifest(self) -> dict[str, Any]:
...
             ),
-            # We don't use the full path here since tests will often use temporary directories for the output
-            # directory which would trigger a rebuild every time.
-            "tools_tree": self.tools_tree.name if self.tools_tree else None,
-            "tools_tree_distribution": self.tools_tree_distribution,
-            "tools_tree_release": self.tools_tree_release,
-            "tools_tree_mirror": self.tools_tree_mirror,
-            "tools_tree_packages": sorted(self.tools_tree_packages),
+            # Statting the root directory of the tools tree isn't fool proof but should be good enough.
+            "tools_tree": [self.tools_tree, self.tools_tree.stat().st_mtime_ns] if self.tools_tree else None,
```
behrmann, line 1823: Nit: 
```suggestion
            "tools_tree": [self.tools_tree, self.tools_tree.stat().st_mtime_ns] if self.tools_tree else [],
```
Would keep the type of the value the same whether `self.tools_tree` is true or not.

mkosi/config.py
```diff
@@ -809,7 +808,10 @@ def config_default_repository_key_fetch(namespace: argparse.Namespace) -> bool:
         return cast(bool, namespace.distribution.is_rpm_distribution())
 
     if namespace.tools_tree != Path("default"):
-        return False
+        return (
+            detect_distribution(namespace.tools_tree)[0] == Distribution.ubuntu
+            and namespace.distribution.is_rpm_distribution()
+        )
```
behrmann, line 814: I think this needs an accompanying change in the section for `RepositoryKeyCheck=`

behrmann (approved):

Issue #3042, opened 2024-09-18
Extend OpenPGP signing support with Stateless OpenPGP


dvzrv:
### mkosi commit the issue has been seen with

https://github.com/systemd/mkosi/commit/baab5c5014e9332c6c85d4408d7b08d7907b40ce

### Used host distribution

Arch Linux

### Used target distribution

Arch Linux

### Linux kernel version used

6.10.10-arch1-1

### CPU architectures issue was seen on

x86_64

### Unexpected behaviour you saw

Currently it is only possible to sign the SHA256SUMS file using `gpg`.
However, gnupg is a complex thing to get working properly (see #3040) and also [starts to diverge from OpenPGP compatibility in >=2.4](https://wiki.archlinux.org/title/GnuPG#OpenPGP_compatibility).

Better, more simple (and stateless) alternatives for the "sign an artifact" use-case exist with [Stateless OpenPGP](https://wiki.archlinux.org/title/Stateless_OpenPGP) (SOP). With [`rsop`](https://codeberg.org/heiko/rsop) we even have a SOP implementation with smartcard support!

To that end, it would be great to extend the signing capabilities in mkosi by allowing to set a specific OpenPGP implementation (e.g. using an `OpenPGPTool` / `--openpgp-tool` option in the `[Validation]` section).
The `Key` option would then need to support setting either an [OpenPGP fingerprint](https://openpgp.dev/book/glossary.html#term-OpenPGP-Fingerprint) (in the case of `gpg`) or a path to a key (or certificate) in the case of SOP implementations.

Additionally, it would be good to rename the currently used options in the `[Validation]` section: `Sign` to `OpenPGPSign` and `Key` to `OpenPGPKey` (that way one could use several signing schemes in parallel, e.g. the ones discussed in https://github.com/systemd/mkosi/issues/624).

Closing, I would like to add, that `.gpg` is not a good signature suffix for OpenPGP signatures (e.g. `.sig` might be better as it is not OpenPGP implementation specific and indicates that it is a signature and not possibly a certificate or a keyring, etc.), but I guess that is currently somewhat fixed due to how [sysupdate.d](https://man.archlinux.org/man/sysupdate.d.5.en) expects it. :smiling_face_with_tear: 

### Used mkosi config

```ini
[Output]
Format=disk
SplitArtifacts=yes

[Content]
Bootable=yes
Bootloader=systemd-boot
Hostname=arch
Packages=
	base
	linux
	nftables
	openssh
	systemd
[...]

PR #3094, opened 2024-10-02
Allow specifying OpenPGP implementation to use for signing

wiktor-k:
Hi :wave: 

I'm pushing a first preview of SOP signing for SHA256SUMs. I've tested that locally and also added tests for rsop and old gpg (just in case) and will begin to refactor this a bit. Since I'm new to both mkosi and Python I think it's good to get feedback earlier in the process if something looks particularly wrong here.

Note that I've tried to make it as minimal as possible with *no* nice-to-have adjustments mentioned in #3042.

Fixes: https://github.com/systemd/mkosi/issues/3042
CC: @dvzrv


mkosi/config.py
```diff
@@ -4337,6 +4344,7 @@ def summary(config: Config) -> str:
                          Passphrase: {none_to_none(config.passphrase)}
                            Checksum: {yes_no(config.checksum)}
                                Sign: {yes_no(config.sign)}
+                       OpenPGP Tool: ({"gpg" if config.openpgp_tool is None else config.openpgp_tool})
```
behrmann, line 4550: This can't really be `None`, since `gpg` is the default. The worst thing would be, that it's empty
```suggestion
                       OpenPGP Tool: ({config.openpgp_tool or "gpg"})
```

mkosi/config.py
```diff
@@ -1535,6 +1535,7 @@ class Config:
     passphrase: Optional[Path]
     checksum: bool
     sign: bool
+    openpgp_tool: Optional[str]
```
behrmann, line 1633: Since this has a default, this could probably just be `str`.

mkosi/__init__.py
```diff
@@ -2157,41 +2157,68 @@ def calculate_signature(context: Context) -> None:
     if not context.config.sign or not context.config.checksum:
         return
 
-    cmdline: list[PathString] = ["gpg", "--detach-sign", "--pinentry-mode", "loopback"]
+    tool = context.config.openpgp_tool
```
behrmann, line 2160: Can we name this `pgptool` or something like that? Tools trees are most often referred to as `tools` and with only one letter difference, this might be confusing on such a long stretch.

DaanDeMeyer:
I'm curious about hardware tokens? Will those be supported with a unified CLI interface as well?

mkosi/__init__.py
```diff
@@ -2157,41 +2157,68 @@ def calculate_signature(context: Context) -> None:
...
+    if tool is None or tool == "gpg":
```
behrmann, line 2161: The two branches should probably be their own functions `sign_gpg` and `sign_stateless_pgp`.

This will also make the diff somewhat smaller, since the gpg code will only move and not change indentation level.

wiktor-k:
> I'm curious about hardware tokens. Will those be supported with a unified CLI interface as well?

Yes, I actually tested that with my Yubikey and found [a couple of issues I fixed](https://codeberg.org/heiko/rsop/pulls/14). This, of course, depends on the underlying SOP implementation that's used. `rsop` implements it.

wiktor-k:
Okay folks, thanks for the great feedback! I need to install the type-checking tools locally not to spam the CI. I'll be back with the adjustments (all valid points! :bow: )

mkosi/config.py
```diff
@@ -1632,7 +1632,8 @@ class Config:
     passphrase: Optional[Path]
     checksum: bool
     sign: bool
-    key: Optional[str]
+    openpgp_tool: str
+    key: Optional[Union[Path, str]]
```
behrmann, line 1636: We have a thing there
```suggestion
    key: Optional[PathString]
```

.github/workflows/ci.yml
```diff
@@ -156,7 +156,7 @@ jobs:
       - name: Install
         run: |
           sudo apt-get update
-          sudo apt-get install python3-pytest lvm2 cryptsetup-bin btrfs-progs
+          sudo apt-get install python3-pytest lvm2 cryptsetup-bin btrfs-progs sqop
```
DaanDeMeyer, line 159: This can be dropped

tests/test_signing.py
```diff
@@ -0,0 +1,97 @@
...
+
+from . import Image, ImageConfig
+
+pytestmark = pytest.mark.integration
+
+
+@pytest.mark.skipif(
+    not Path("/usr/bin/sqop").exists(),
+    reason="requires sqop for signing but not all distros package it (e.g. not in opensuse)",
+)
```
DaanDeMeyer, line 20: This doesn't actually work, sometimes tests run with a tools tree which means we pick up tools from the tools tree instead of from the host, so checking whether the binary is installed on the host like you do here doesn't work.

With https://github.com/systemd/mkosi/pull/3109 the tools tree path will be available in the `ImageConfig` object and you can use it together with `find_binary()` to check if `sqop` is available

wiktor-k: That's great! I tried checking for `config.distribution` and comparing that to `Distribution.opensuse` but it ended up not good :-/

tests/test_signing.py
```diff
@@ -0,0 +1,97 @@
...
+
+from mkosi.run import find_binary, run
+
+from . import Image, ImageConfig
+
+pytestmark = pytest.mark.integration
+
+
+def test_signing_checksums_with_sop(config: ImageConfig) -> None:
+    assert config.tools is not None, "Needs config.tools to check for SOP implementation"
```
wiktor-k, line 18: :thinking: I've read #3109 and maybe I'm holding something wrong but this assertion keeps failing (locally). Do I need to enable something for the test to get the `mkosi.tools`?

DaanDeMeyer: You just need to pass `root=config.tools` to `find_binary()` that's it, no need for assetions. It might or might not be set depending on whether mkosi.tools exists or not.

tests/test_signing.py
```diff
@@ -0,0 +1,97 @@
...
+    if find_binary("sqop", config.tools) is None:
+        print("Needs `sqop` binary in PATH to perform sop tests.")
+        return
```
DaanDeMeyer, line 20: Use `pytest.skip` please

tests/test_signing.py
```diff
@@ -0,0 +1,97 @@
...
+
+        signing_key.chmod(0o755)
+
+        # extract public key (certificate)
+        with open(signing_key, "rb") as i, open(signing_cert, "wb") as o:
+            run(cmdline=["sqop", "extract-cert"], stdin=i, stdout=o)
+
+        signing_cert.chmod(0o755)
+
+        with Image(config) as image:
```
DaanDeMeyer, line 40: This can be merged with the upper `with` to reduce the indentation

tests/test_signing.py
```diff
@@ -0,0 +1,96 @@
...
+        for p in gnupghome.rglob("*"):
+            p.chmod(0o777)
+
+        # export public key (certificate)
+        with open(signing_cert, "wb") as o:
+            run(cmdline=["gpg", "--export", signing_key], env=env, stdout=o)
+
+        signing_cert.chmod(0o755)
+
+        with open("/tmp/list", "wb") as o:
```
DaanDeMeyer, line 86: Please use `tempfile.TemporaryFile` here

DaanDeMeyer: I'm a bit confused what this actually does?

tests/test_signing.py
```diff
@@ -0,0 +1,96 @@
...
+            run(["ls", "-la", gnupghome], stdout=o)
+
+        with Image(config) as image:
```
DaanDeMeyer, line 89: Same thing, please merge with upper `with` to reduce indentation

DaanDeMeyer (changes requested):

tests/test_signing.py
wiktor-k: Whoops... this was my debugging code, I'll remove it altogether!

tests/test_signing.py
wiktor-k: Okay. I assumed `pytest.skip` wouldn't be able to access the `config` parameter of the function but I didn't actually test this :roll_eyes:

mkosi/config.py
```diff
@@ -4562,6 +4569,7 @@ def summary(config: Config) -> str:
                          Passphrase: {none_to_none(config.passphrase)}
                            Checksum: {yes_no(config.checksum)}
                                Sign: {yes_no(config.sign)}
+                       OpenPGP Tool: ({config.openpgp_tool or "gpg"})
```
behrmann, line 4572: Just so we don't forget: This also needs an entry in the man page.

tests/test_signing.py
```diff
@@ -0,0 +1,90 @@
...
+from mkosi.run import find_binary, run
+
+from . import Image, ImageConfig
+
+pytestmark = pytest.mark.integration
+
+
+def test_signing_checksums_with_sop(config: ImageConfig) -> None:
+    if find_binary("sqop", root=config.tools) is None:
+        pytest.skip("Needs `sqop` binary in PATH to perform sop tests.")
```
behrmann, line 19: Nit: let's not do backticks, since they can evaluate in a shell and people tend to copy stuff around, also, an empty line after this would be nice.
```suggestion
        pytest.skip("Needs 'sqop' binary in PATH to perform sop tests.")
        
```

tests/test_signing.py
```diff
@@ -0,0 +1,90 @@
...
+    with tempfile.TemporaryDirectory() as path, Image(config) as image:
+        tmp_path = Path(path)
+        tmp_path.chmod(0o755)
```
behrmann, line 22: Not that it matters a lot in the test, but why does the directory holding the key need to be world readable?

tests/test_signing.py
```diff
@@ -0,0 +1,90 @@
...
+
+        signing_key = tmp_path / "signing-key.pgp"
+        signing_cert = tmp_path / "signing-cert.pgp"
+
+        # create a brand new signing key
+        with open(signing_key, "wb") as o:
+            run(cmdline=["sqop", "generate-key", "--signing-only", "Test"], stdout=o)
+
+        signing_key.chmod(0o755)
```
behrmann, line 31: And why is the key executable (and the cert below, too)?

tests/test_signing.py
```diff
@@ -0,0 +1,90 @@
...
+        signature = image.output_dir / "image.SHA256SUMS.gpg"
+
+        with open(signed_file, "rb") as i:
+            run(cmdline=["sqop", "verify", signature, signing_cert], stdin=i)
+
+
+def test_signing_checksums_with_gpg(config: ImageConfig) -> None:
+    with tempfile.TemporaryDirectory() as path, Image(config) as image:
+        tmp_path = Path(path)
+        tmp_path.chmod(0o777)
```
behrmann, line 53: Let's really not have a world-writable thing, even in tests.

wiktor-k: The issue here is that `gpg` during its operation writes temporary files to that directory and since the tests starts as root but then switch user during `image.build` the directory must be readable to both root and that temporary build user.

I'll try to minimize the number of `chmod`s. With the `sop` case the key and cert files need to be just readable by both users but in GPG case it needs writeable access sadly :/

wiktor-k:
Okay, I think I minimized it as much as possible. Removed all dependencies to the point where removing anything else triggers CI failures.

One thing that is consistently failing is the opensuse/arch build with a rather peculiar message:

```
Failed:
    dbus-1-common-1.14.10-3.1.noarch                                              
    system-group-hardware-20170617-25.4.noarch                                    
    system-group-kvm-20170617-25.4.noarch                                         
    system-user-lp-20170617-25.4.noarch                                           
  
  Cleaning up.
  Plugins were unloaded.
  
    File "/usr/lib/python3.12/site-packages/dnf/cli/main.py", line 67, in main
      return _main(base, args, cli_class, option_parser_class)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/usr/lib/python3.12/site-packages/dnf/cli/main.py", line 106, in _main
      return cli_run(cli, base)
             ^^^^^^^^^^^^^^^^^^
    File "/usr/lib/python3.12/site-packages/dnf/cli/main.py", line 130, in cli_run
      ret = resolving(cli, base)
            ^^^^^^^^^^^^^^^^^^^^
    File "/usr/lib/python3.12/site-packages/dnf/cli/main.py", line 183, in resolving
      base.do_transaction(display=displays)
    File "/usr/lib/python3.12/site-packages/dnf/cli/cli.py", line 272, in do_transaction
      raise dnf.exceptions.Error(_('Transaction failed'))
  dnf.exceptions.Error: Transaction failed
```

If you have any ideas on what to optimize here I'm all ears. Sadly the GnuPG tests require world-writable dirs (since the dir is setup before the `image.build` thing is executed). I could also remove that test but I think it's a good idea to see if signing works. If there are other approaches I can try, I'm happy to adjust the code.

Thanks for your time! :wave:

mkosi/config.py
```diff
@@ -4597,6 +4604,7 @@ def summary(config: Config) -> str:
...
```
DaanDeMeyer, line 4607: You can drop the fallback here since the field will always be set as it has a default value

tests/test_signing.py
```diff
@@ -0,0 +1,89 @@
...
+from mkosi.run import find_binary, run
+
+from . import Image, ImageConfig
+
+pytestmark = pytest.mark.integration
+
+
+def test_signing_checksums_with_sop(config: ImageConfig) -> None:
+    if find_binary("sqop", root=config.tools) is None:
+        pytest.skip("Needs 'sqop' binary in PATH to perform sop tests.")
```
DaanDeMeyer, line 19: Add a newline here

tests/test_signing.py
```diff
@@ -0,0 +1,89 @@
...
+        # create a brand new signing key
+        run(cmdline=["gpg", "--quick-gen-key", "--batch", "--passphrase", "", signing_key], env=env)
+
+        # GnuPG will set 0o700 permissions so that the secret files are not available
+        # to other users. Since this is for tests only and we need that keyring for signing
+        # enable all permissions. We need write permissions since GnuPG creates temporary
+        # files in this directory during operation.
+        gnupghome.chmod(0o777)
+        for p in gnupghome.rglob("*"):
+            p.chmod(0o777)
```
DaanDeMeyer, line 76: So maybe we can just make sure the owner is correct here instead. You can chown the directory to Image.uid, Image.gid and run the gpg process as that user as well, then we don't need the world writable directory.

DaanDeMeyer (changes requested):

mkosi/config.py
wiktor-k: Entry added. See if looks okay.

wiktor-k:
Okay, I've adjusted the PR and included manpage changes. The CI failures don't look to be related (please correct me if I'm wrong). Thanks! :wave:

DaanDeMeyer (changes requested):

mkosi/resources/man/mkosi.1.md
```diff
@@ -1195,6 +1195,11 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 `Sign=`, `--sign`
 :   Sign the generated `SHA256SUMS` using `gpg` after completion.
 
+`OpenpgpTool=`, `--openpgp-tool`
+:   OpenPGP implementation to use for signing. `gpg` is the default.
+    Selecting a value different than the default will use a given Stateless
```
behrmann, line 1200:
```suggestion
    Selecting a value different than the default will use the given Stateless
```
Also, itt would be nice to add an example here, e.g. `sqop`.

wiktor-k: Good idea with the example. I could list two (say `sqop` and `rsop`). What do you think about linking to https://www.openpgp.org/about/sop/ (which lists an even wider roster of implementations)? Would that be a bit too much?

behrmann: I don't think that would be too much, how about
```suggestion
    OpenPGP (SOP) tool for signing the `SHA256SUMS` file.
    
    Examplary choices are `sqop` and `rsop`, but any implementation from 
    https://www.openpgp.org/about/sop/ that can be installed locally will work
```

wiktor-k: Excellent wording, thanks for the suggestion! I just appended a dot at the end and re-pushed. I've also checked it locally with `./tools/make-man-page.sh` and `man ./mkosi/resources/man/mkosi.1`. It looked fine.

tests/__init__.py
```diff
@@ -61,6 +61,7 @@ def mkosi(
         user: Optional[int] = None,
         group: Optional[int] = None,
         check: bool = True,
+        env: Optional[Mapping[str, str]] = None,
```
behrmann, line 64: Couldn't this be made a bit tighter?
```suggestion
        env: Mapping[str, str] = os.environ,
```
This would also make the fallback clear in the signature without having to scroll through the code.

tests/__init__.py
```diff
@@ -74,12 +75,18 @@ def mkosi(
...
-            env=os.environ,
+            env=env or os.environ,
         )  # fmt: skip
 
-    def build(self, options: Sequence[PathString] = (), args: Sequence[str] = ()) -> CompletedProcess:
+    def build(
+        self,
+        options: Sequence[PathString] = (),
+        args: Sequence[str] = (),
+        env: Optional[Mapping[str, str]] = None,
```
behrmann, line 88: As above.

behrmann (approved):
LGTM, apart from the nits.

wiktor-k:
> LGTM, apart from the nits.

Phew, thanks for your time! Nits resolved :sweat_smile:

tests/__init__.py
```diff
@@ -74,6 +74,7 @@ def mkosi(
             check=check,
             stdin=stdin,
             stdout=sys.stdout,
+            stderr=sys.stderr,
```
DaanDeMeyer, line 78: This is the default actually, see the implementation of `spawn()`

DaanDeMeyer, line 78: This is the default, no need to specify it explicitly

wiktor-k: Ack, dropping that commit then!

mkosi/__init__.py
```diff
@@ -2295,6 +2303,37 @@ def calculate_signature(context: Context) -> None:
...
+    if signing_key is None:
+        die("Signing key is mandatory when using SOP signing")
+
+    cmdline: list[PathString] = [pgptool, "sign", "/signing-key.pgp"]
+
+    options: list[PathString] = [
+        "--bind", signing_key, "/signing-key.pgp",
+        "--bind", context.staging, workdir(context.staging),
+        "--bind", "/run", "/run",
+    ]  # fmt: skip
```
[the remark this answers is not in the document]
wiktor-k: Inlined, the formatting changed even though `# fmt: skip` was there... maybe I used it wrong, or maybe it's not a problem? :thinking:

tests/test_signing.py
```diff
@@ -0,0 +1,90 @@
...
+        os.chown(tmp_path, image.uid, image.gid)
+
+        signing_key = tmp_path / "signing-key.pgp"
+        signing_cert = tmp_path / "signing-cert.pgp"
+
+        # create a brand new signing key
+        with open(signing_key, "wb") as o:
+            run(cmdline=["sqop", "generate-key", "--signing-only", "Test"], stdout=o)
+
+        signing_key.chmod(0o744)
```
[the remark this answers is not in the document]
wiktor-k: Tested again and it works without it so: removing.

mkosi/__init__.py
```diff
@@ -2258,6 +2258,14 @@ def calculate_signature(context: Context) -> None:
     if not context.config.sign or not context.config.checksum:
         return
 
+    pgptool = context.config.openpgp_tool
```
DaanDeMeyer, line 2261: Please drop this one as well

mkosi/__init__.py
```diff
@@ -2295,6 +2303,38 @@ def calculate_signature(context: Context) -> None:
         )
 
 
+def calculate_signature_sop(context: Context) -> None:
+    pgptool = context.config.openpgp_tool
```
DaanDeMeyer, line 2307: And this one too

mkosi/__init__.py
```diff
@@ -2295,6 +2303,38 @@ def calculate_signature(context: Context) -> None:
...
+            [pgptool, "sign", "/signing-key.pgp"],
+            env=context.config.environment,
+            stdin=i,
+            stdout=o,
+            sandbox=context.sandbox(
+                binary=pgptool,
+                options=[
+                    "--bind",
+                    context.config.key,
+                    "/signing-key.pgp",
```
DaanDeMeyer, line 2326: Please keep options and their arguments on the same line and disable auto formatting for this block, see rest of the code for examples

mkosi/config.py
```diff
@@ -1750,7 +1750,8 @@ class Config:
     passphrase: Optional[Path]
     checksum: bool
     sign: bool
-    key: Optional[str]
+    openpgp_tool: str
+    key: Optional[PathString]
```
DaanDeMeyer, line 1754: This is always a string, should keep the type as it was before

DaanDeMeyer (changes requested):

DaanDeMeyer (approved):

PR #3095, opened 2024-10-02
tools: Don't use apt pattern for ubuntu-keyring

behrmann:
Running

    mkosi --directory "" -d debian -r bookworm --include mkosi-tools --output mkosi.tools

on current Debian stable on gets an error that the package 'ubuntu-keyring' doesn't have any installable candidates. Moving the inclusion of the package out of the purview of apt and back into mkosi's fixes this issue.


behrmann:
@bluca Do you have an idea why this could be? First I assumed the pattern would break because `ubuntu-keyring` exists in `bookworm-backports`, but not `bookworm` proper, but the same is also true for `distribution-gpg-keys`. I don't really see a pattern for this.

bluca:
Dunno, maybe the pattern breaks due to `Provides: ubuntu-archive-keyring, ubuntu-cloud-keyring, ubuntu-dbgsym-keyring` ? Just guessing

DaanDeMeyer (approved):

PR #3096, opened 2024-10-02
mkosi-initrd: build using a temporary directory

### LINKED ISSUE #3083: `--force` should not remove the output when it fails

aafeijoo-suse:
Fixes #3083

### DIFF
--- a/mkosi/initrd.py
+++ b/mkosi/initrd.py
@@ -11,9 +11,10 @@ from pathlib import Path
 import mkosi.resources
 from mkosi.config import DocFormat, OutputFormat
 from mkosi.documentation import show_docs
-from mkosi.log import log_setup
+from mkosi.log import log_notice, log_setup
 from mkosi.run import find_binary, run, uncaught_exception_handler
-from mkosi.sandbox import __version__
+from mkosi.sandbox import __version__, umask
+from mkosi.tree import copy_tree
 from mkosi.types import PathString
 from mkosi.util import resource_path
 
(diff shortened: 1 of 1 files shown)


aafeijoo-suse:
wow github went crazy with the diff

behrmann:
It's fine when one hides the whitespace or looks at it locally `--color-moved=blocks` and `--color-moved-ws=allow-indentation-change` ;)

mkosi/initrd.py
```diff
@@ -88,89 +89,96 @@ def main() -> None:
...
+            # there will be a populated dnf cache directory.
+            run(
+                cmdline,
+                stdin=sys.stdin,
+                stdout=sys.stdout,
+                env={"MKOSI_DNF": dnf.name} if (dnf := find_binary("dnf", "dnf5")) else {},
+            )
+
+            if not args.output_dir:
+                args.output_dir = Path().resolve()
```
DaanDeMeyer, line 178: `Path.cwd()`

DaanDeMeyer (changes requested):
We should ignore the symlink mkosi creates in the temporary output directory. It'll be named identically to `args.output` in this case

aafeijoo-suse:
> It's fine when one hides the whitespace or looks at it locally `--color-moved=blocks` and `--color-moved-ws=allow-indentation-change` ;)

TIL, thanks! Although I expected github to be smarter by default ;)

aafeijoo-suse:
> We should ignore the symlink mkosi creates in the temporary output directory. It'll be named identically to `args.output` in this case

I agree with this. PTAL.

mkosi/initrd.py
```diff
@@ -88,89 +89,99 @@ def main() -> None:
...
-
-        # Prefer dnf as dnf5 has not yet officially replaced it and there's a much bigger chance that there
-        # will be a populated dnf cache directory.
-        run(
-            cmdline,
-            stdin=sys.stdin,
-            stdout=sys.stdout,
-            env={"MKOSI_DNF": dnf.name} if (dnf := find_binary("dnf", "dnf5")) else {},
-        )
+                shutil.copy2(Path(staging_dir) / args.output, Path(args.output_dir))
```
DaanDeMeyer, line 184: Doesn't this copy only the symlink instead?

DaanDeMeyer: Ah I guess it dereferences it

DaanDeMeyer: Can we make that explicit by adding a `resolve()` here and a comment?

DaanDeMeyer: And in that case, we can use `copy_tree()` in both cases as well

aafeijoo-suse: hmm, not on my tests... but I actually forgot to create the output directory if it does not exist

aafeijoo-suse: Done and tested ok.

mkosi/initrd.py
```diff
@@ -88,89 +89,102 @@ def main() -> None:
...
+            # there will be a populated dnf cache directory.
+            run(
+                cmdline,
+                stdin=sys.stdin,
+                stdout=sys.stdout,
+                env={"MKOSI_DNF": dnf.name} if (dnf := find_binary("dnf", "dnf5")) else {},
+            )
+
+            if args.output_dir:
+                Path(args.output_dir).mkdir(parents=True, exist_ok=True)
```
DaanDeMeyer, line 178: Please use the umask() helper here to make sure the mode is 700

mkosi/initrd.py
```diff
@@ -88,89 +89,102 @@ def main() -> None:
...
-            stdout=sys.stdout,
-            env={"MKOSI_DNF": dnf.name} if (dnf := find_binary("dnf", "dnf5")) else {},
-        )
+                args.output_dir = Path.cwd()
+
+            log_notice(f"Copying {staging_dir}/{args.output} to {args.output_dir}/{args.output}")
+            # mkosi symlinks the expected output image, so dereference it
+            copy_tree(
+                Path(f"{staging_dir}/{args.output}").resolve(),
+                Path(f"{args.output_dir}/{args.output}"),
```
DaanDeMeyer, line 186: This will rename the file when copying. I don't think we want that, we want the full name with extension and compression. Just specifying the output directory here will be sufficient

aafeijoo-suse: Hmm.. I don't think we want the full name here, if someone wants to produce an image called `initrd-<kver>`, he wants that name and not e.g. `initrd-<kver>.cpio.zst`

DaanDeMeyer: Makes sense, let's leave it like this

aafeijoo-suse: Ok, done.

mkosi/initrd.py
```diff
@@ -88,89 +89,103 @@ def main() -> None:
...
-            if not (Path("/etc") / p).exists():
-                continue
-
-            (Path(d) / "etc" / p).parent.mkdir(parents=True, exist_ok=True)
-            if (Path("/etc") / p).resolve().is_file():
-                shutil.copy2(Path("/etc") / p, Path(d) / "etc" / p)
+            if Path(d).exists():
+                cmdline += ["--include", d]
+
+        with tempfile.TemporaryDirectory() as d:
```
DaanDeMeyer, line 132: Actually this leads to unnecessary indentation, can you merge this with the `with` above to reduce the indentation?

DaanDeMeyer (changes requested):

DaanDeMeyer (approved):

PR #3103, opened 2024-10-04
Relax read-only mounts even more

DaanDeMeyer:
Turns out having home directories as a subdirectory of /usr is a thing. Let's relax the readonly mount requirements even more to make this use case work as well.

PR #3106, opened 2024-10-06
Define our own config for PE addons and UKI profiles

DaanDeMeyer:
ukify's config parser uses python's configparser module and as such
suffers from all its issues just like we used to in mkosi. Having ukify
parse the config file also means that we have to make sure any paths
configured in the profile are available in the sandbox.

Instead, let's define our own configs for the PE addons and UKI profiles
so we get to take advantage of our own config file parser and have full
knowledge of all the configured settings so we can mount extra stuff into
the sandbox if needed.

It also gets rid of the hack where we parse ukify's config file to figure
out the command line.

cc @NekkoDroid

### DIFF
--- a/mkosi/resources/man/mkosi.1.md
+++ b/mkosi/resources/man/mkosi.1.md
@@ -942,19 +942,25 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 
 `UnifiedKernelImageProfiles=`, `--uki-profile=`
 :   Build additional UKI profiles. Takes a comma separated list of paths
-    to `ukify` config files. This option may be used multiple times in
+    to UKI profile config files. This option may be used multiple times in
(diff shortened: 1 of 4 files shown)


DaanDeMeyer:
This is on top of #3103

mkosi/resources/man/mkosi.1.md
```diff
@@ -1932,6 +1938,41 @@ config file is read:
     each individual subimage as if they were "universal" settings. See
     the **Building multiple images** section for more information.
 
+### [Addon] Section
```
behrmann, line 1941: I think we should name this section `PeAddon` and the other `UkiProfile` (or `UnifiedKernelImageProfile`), it's more symmetrical with the options and doesn't give up the short names to these more specialised uses (also in the case of the Profile section it could be confusing with our profiles).

PR #3122, opened 2024-10-10
Update description of `ToolsTreeDistribution=`

septatrix:
It was changed to default to the host distro instead of the target distro a while a ago but this was forgotten.

PS: The code also supports falling back to `custom` but I am unsure if that makes a lot of sense?


DaanDeMeyer:
Probably does not but let's fix that in another PR

DaanDeMeyer (approved):

PR #3125, opened 2024-10-11
Fix repart definition for ESP builds

septatrix:
Currently ESP builds succeed but do not actually include any UKI as sd-repart silently ignore that they are missing. I am not sure if this is the correct place to add that fix. Maybe a `--copy-from` in the repart invocation would be cleaner.

Also, disk builds with UKI bootloaders are still broken - not sure if anything else is affected, too

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2986,6 +2986,7 @@ def make_image(
     verity: bool = False,
     root: Optional[Path] = None,
     definitions: Sequence[Path] = [],
+    options: Sequence[PathString] = (),
 ) -> list[Partition]:
     cmdline: list[PathString] = [
         "systemd-repart",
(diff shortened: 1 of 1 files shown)


DaanDeMeyer (approved):
Yikes, good catch

DaanDeMeyer (changes requested):
We also need to pass something like `--ro-bind, uki, workdir(uki)` to the sandbox options of `make_image()`

septatrix:
> We also need to pass something like `--ro-bind, uki, workdir(uki)` to the sandbox options of `make_image()`

Do we? The uki is implicitly mounted as part of the staging dir under `/work`. Or do you suggest that as an alternative to adjusting the repart definition?

I am really not sure where the best place to put this fix is.

septatrix:
And I took a look at the logs when building a disk image with a UKI as a bootloader but was unable to find the error in that case...

DaanDeMeyer:
> > We also need to pass something like `--ro-bind, uki, workdir(uki)` to the sandbox options of `make_image()`
> 
> Do we? The uki is implicitly mounted as part of the staging dir under `/work`. Or do you suggest that as an alternative to adjusting the repart definition?
> 
> I am really not sure where the best place to put this fix is.

You are correct that it's available in the sandbox, it's the "implicit" that's the problem. We mount in the staging directory for another unrelated reason and that might change in the future so I'd like to make it explicit. Just adding an `options` argument to `make_image()` and then passing it through is good enough.

DaanDeMeyer (approved):

PR #3129, opened 2024-10-16
Fix conf with settings in Content instead of Build

aafeijoo-suse:
Follow-up for 6912dc0ea0c7e20900a9b5d2c43343c1c782e91c

Seen on the CI for `Environment=`, but there are more wrong settings remaining.

```
2024-10-16T08:48:32.7586608Z tests/test_initrd.py::test_initrd_luks 
2024-10-16T08:48:32.7587441Z ##[group]test_initrd_luks
2024-10-16T08:48:32.8751669Z ‣ Loading configuration file /home/runner/work/mkosi/mkosi/mkosi.local.conf
2024-10-16T08:48:32.8753483Z ‣ /home/runner/work/mkosi/mkosi/mkosi.local.conf: Setting Environment should be configured in [Build], not [Content].
```

BTW, there is something wrong with `test_config`, because it should be failing before this PR and it isn't. If the section is valid the test pass (although the specific conf setting is not under the right section).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -169,6 +169,8 @@ jobs:
 
           [Content]
           KernelCommandLine=systemd.default_device_timeout_sec=180
+
+          [Build]
           Environment=SYSTEMD_REPART_MKFS_OPTIONS_EROFS="--quiet"
 
           [Host]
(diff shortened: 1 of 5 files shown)


behrmann (approved):

behrmann:
I don't think `test_config` is broken, since we only log a warning, but essentially only throw everything into one big bucket anyway. We should though, add something of a strict mode, maybe it's enough to run python with `-Werror` (though this would then break the test counting the warnings).

aafeijoo-suse:
> Yes, because we the only thing we do in that case is log a warning, because we have no conflicting keys between sections and throw everything into the same config object anyway.

Ahh ok, good to know. Thanks for the explanation.

PR #3131, opened 2024-10-17
Allow more granular control on which artifacts are output

NekkoDroid:
This PR repurposes `SplitArtifacts=` to be able to specify multiple artifact types that should be split out of the image. By specifying one or more of `uki`, `kernel`, `initrd` and `partitions` the respective components are extracted from the image. The truthy and falsy values still map to the equivalent of the old values to preserve backwards compatibility.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -50,6 +50,7 @@ from mkosi.completion import print_completion
 from mkosi.config import (
     PACKAGE_GLOBS,
     Args,
+    ArtifactOutput,
     Bootloader,
     Cacheonly,
     Compression,
@@ -2100,8 +2101,11 @@ def make_uki(
         output,
     )
 
-    extract_pe_section(context, output, ".linux", context.staging / context.config.output_split_kernel)
-    extract_pe_section(context, output, ".initrd", context.staging / context.config.output_split_initrd)
+    if ArtifactOutput.kernel in context.config.split_artifacts:
+        extract_pe_section(context, output, ".linux", context.staging / context.config.output_split_kernel)
+
+    if ArtifactOutput.initrd in context.config.split_artifacts:
+        extract_pe_section(context, output, ".initrd", context.staging / context.config.output_split_initrd)
 
 
 def compressor_command(context: Context, compression: Compression) -> list[PathString]:
@@ -2179,6 +2183,9 @@ def get_uki_path(context: Context) -> Optional[Path]:
 
 
 def copy_uki(context: Context) -> None:
+    if ArtifactOutput.uki not in context.config.split_artifacts:
+        return
+
     if (context.staging / context.config.output_split_uki).exists():
         return
 
@@ -2187,6 +2194,9 @@ def copy_uki(context: Context) -> None:
 
 
 def copy_vmlinuz(context: Context) -> None:
+    if ArtifactOutput.kernel not in context.config.split_artifacts:
+        return
+
     if (context.staging / context.config.output_split_kernel).exists():
         return
 
@@ -2202,6 +2212,9 @@ def copy_vmlinuz(context: Context) -> None:
 
 
 def copy_initrd(context: Context) -> None:
+    if ArtifactOutput.initrd not in context.config.split_artifacts:
+        return
+
     if not want_initrd(context):
         return
 
@@ -3379,7 +3392,7 @@ def make_extension_image(context: Context, output: Path) -> None:
         ]  # fmt: skip
     if context.config.sector_size:
         cmdline += ["--sector-size", str(context.config.sector_size)]
-    if context.config.split_artifacts:
+    if ArtifactOutput.partitions in context.config.split_artifacts:
         cmdline += ["--split=yes"]
 
     with complete_step(f"Building {context.config.output_format} extension image"):
@@ -3401,7 +3414,7 @@ def make_extension_image(context: Context, output: Path) -> None:
 
     logging.debug(json.dumps(j, indent=4))
 
-    if context.config.split_artifacts:
+    if ArtifactOutput.partitions in context.config.split_artifacts:
         for p in (Partition.from_dict(d) for d in j):
             if p.split_path:
                 maybe_compress(context, context.config.compress_output, p.split_path)
@@ -3646,7 +3659,7 @@ def build_image(context: Context) -> None:
     partitions = make_disk(context, msg="Formatting ESP/XBOOTLDR partitions")
     grub_bios_setup(context, partitions)
 
-    if context.config.split_artifacts:
+    if ArtifactOutput.partitions in context.config.split_artifacts:
         make_disk(context, split=True, msg="Extracting partitions")
 
     copy_nspawn_settings(context)
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2152,12 +2152,9 @@ def copy_nspawn_settings(context: Context) -> None:
         shutil.copy2(context.config.nspawn_settings, context.staging / context.config.output_nspawn_settings)
 
 
-def copy_uki(context: Context) -> None:
-    if (context.staging / context.config.output_split_uki).exists():
-        return
-
+def get_uki_path(context: Context) -> Optional[Path]:
     if not want_efi(context.config) or context.config.unified_kernel_images == ConfigFeature.disabled:
-        return
+        return None
 
     ukis = sorted(
         (context.root / "boot/EFI/Linux").glob("*.efi"),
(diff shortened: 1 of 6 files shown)


mkosi/__init__.py
```diff
@@ -2144,66 +2148,82 @@ def maybe_compress(
...
-        pass
-    elif ukis:
-        uki = ukis[0]
-    else:
-        return
+    if ukis:
+        return ukis[0]
 
-    shutil.copy(uki, context.staging / context.config.output_split_uki)
+    return None
```
behrmann, line 2180: Please, no `return None` at the end
 ```python
if not ukis:
    return None

return ukis[0]
```

File: mkosi/__init__.py [hunk 2]
```diff
@@ -2144,66 +2148,82 @@ def maybe_compress(
...
+        shutil.copy2(context.config.nspawn_settings, context.staging / context.config.output_nspawn_settings)
+
 
+def get_uki(context: Context) -> Optional[Path]:
     if not want_efi(context.config) or context.config.unified_kernel_images == ConfigFeature.disabled:
-        return
+        return None
+
+    uki = context.root / efi_boot_binary(context)
+    if uki.exists() and (KernelType.identify(context.config, uki) == KernelType.uki):
```
behrmann, line 2164: when moving code, please keep the style in tact (within reason), since that makes reviewing the diff easier, since git can then colour the move as a move instead of as removed and added lines.
```suggestion
    if (uki := context.root / efi_boot_binary(context)).exists() and (
        KernelType.identify(context.config, uki) == KernelType.uki
    ):
```

mkosi/__init__.py
```diff
@@ -2144,66 +2148,82 @@ def maybe_compress(
...
 
 
-def copy_nspawn_settings(context: Context) -> None:
-    if context.config.nspawn_settings is None:
-        return None
-
-    with complete_step("Copying nspawn settings file…"):
-        shutil.copy2(context.config.nspawn_settings, context.staging / context.config.output_nspawn_settings)
-
-
```
behrmann, line 2199: Nit: it makes sense to move the function from this somewhat random place, having that as the first commit would be nicer, since that obscures the refactor a bit and doesn't have any functional change.

mkosi/resources/man/mkosi.1.md
```diff
@@ -596,13 +596,18 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
 `SplitArtifacts=`, `--split-artifacts`
-:   If specified and building a disk image, pass `--split=yes` to systemd-repart
-    to have it write out split partition files for each configured partition.
-    Read the [man](https://www.freedesktop.org/software/systemd/man/systemd-repart.html#--split=BOOL)
+:   The artifact types to split out of the final image. A comma-delimited
+    list consisting of `uki`, `kernel`, `initrd` and `partitions`. When
+    building a bootable image `kernel` and `initrd` correspond to their
+    artifact found in the image (or in the UKI), while `uki` copies out the
+    entire UKI. When building a disk image and `partitions` is specified,
+    pass `--split=yes` to systemd-repart to have it write out split partition
```
behrmann, line 604:
```suggestion
    entire UKI. 
    
    When building a disk image and `partitions` is specified,
    `--split=yes` is passed to systemd-repart to have it write out split partition
```

behrmann (commented):
I like the refactor in the first commit (sans the style nits), but I'm a bit -0 on the second commit, since it breaks backwards compatibility and I'm not quite sure I understand why the superfluous artifacts cannot be deleted with a postoutput script, but I don't have too strong feelings there.

The backwards compat could probably be kept by making `ArtifactsOutput` a `enum.Flag` instead of a `StrEnum`, which would then probably look something like
```python
class ArtifactOutput(enum.Flag):
    false = 0
    uki = 1
    kernel = 2
    initrd = 4
    partitions = 8
    default = 7
```

NekkoDroid:
still looking into using `enum.Flag`, seems promising but I am still having a few problem with it :)

NekkoDroid:
I am unable to figure out how to serialize the value and actually make it assign the value when returning from the parse function. When setting any value it still keeps resulting in the default when printing the summary

mkosi/config.py
```diff
@@ -1291,6 +1309,25 @@ def config_parse_key_source(value: Optional[str], old: Optional[KeySource]) -> O
...
+    if boolean_value is not None:
+        return ArtifactOutput.yes if boolean_value else ArtifactOutput.no
+
+    try:
+        new = ArtifactOutput[value]
+    except KeyError:
+        die(f"'{value}' is not a valid {ArtifactOutput.__name__}")
+
+    return old | new if old else new
+
```
behrmann, line 1330: missing empty line

mkosi/config.py
```diff
@@ -4783,6 +4824,9 @@ def uki_profile_transformer(
     ) -> list[UKIProfile]:
         return [UKIProfile(profile=profile["Profile"], cmdline=profile["Cmdline"]) for profile in profiles]
 
+    def artifact_output_transformer(output: str, fieldtype: type[ArtifactOutput]) -> ArtifactOutput:
```
behrmann, line 4827: Is that one necessary? I think it should in principle just work with the existing `enum_transformer`, e.g. I get
```python
>>> ArtifactOutput(14)
<ArtifactOutput.kernel|initrd|partitions: 14>
```

mkosi/config.py
```diff
@@ -1291,6 +1309,25 @@ def config_parse_key_source(value: Optional[str], old: Optional[KeySource]) -> O
     return KeySource(type=type, source=source)
 
 
+def config_parse_artifact_output(
+    value: Optional[str],
+    old: Optional[ArtifactOutput]
```
behrmann, line 1314:
```suggestion
    old: Optional[ArtifactOutput],
```

NekkoDroid:
the JSON problem is "solved" (happy for any recommendation that makes it easier to understand than the raw flag number).

The issue with not being able to set the values in the config still persists though.

NekkoDroid:
Some CI fails due to dnf, but others seem to not be able to `in` an `IntFlag` :/ Unsure what is causing this

```
    File "/home/runner/work/mkosi/mkosi/mkosi/run.py", line 64, in uncaught_exception_handler
      yield
    File "/home/runner/work/mkosi/mkosi/mkosi/run.py", line 105, in fork_and_wait
      target(*args, **kwargs)
    File "/home/runner/work/mkosi/mkosi/mkosi/__init__.py", line 4416, in run_build
      build_image(
    File "/home/runner/work/mkosi/mkosi/mkosi/__init__.py", line 3661, in build_image
      if ArtifactOutput.partitions in context.config.split_artifacts:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  TypeError: argument of type 'int' is not iterable
```

behrmann:
I'm travelling this weekend and forgot my notebook, so will only be able to look at this again on Monday, sorry.

NekkoDroid:
Needing to use `ArtifactOutput(split_artifacts)` to get CI to be happy isn't the prettiest but I guess it gets the job done. If anyone knows what is causing it or how I could fix it other than what I did I'd like to hear.

mkosi/__init__.py
```diff
@@ -2168,30 +2165,36 @@ def copy_uki(context: Context) -> None:
...
-    extract_pe_section(context, uki, ".initrd", context.staging / context.config.output_split_initrd)
+    return ukis[0]
 
-    # ukify will have signed the kernel image as well. Let's make sure we put the signed kernel
-    # image in the output directory instead of the unsigned one by reading it from the UKI.
-    extract_pe_section(context, uki, ".linux", context.staging / context.config.output_split_kernel)
+
+def copy_uki(context: Context) -> None:
+    if (context.staging / context.config.output_split_uki).exists():
+        return
```
DaanDeMeyer, line 2189: Can't we just return the path here and then get rid of get_uki() and use copy_uki() everywhere instead? I'd prefer to have this be a single function instead of two.

tests/test_json.py
```diff
@@ -333,7 +334,7 @@ def test_config() -> None:
                 }
             ],
             "SourceDateEpoch": 12345,
-            "SplitArtifacts": true,
+            "SplitArtifacts": 3,
```
DaanDeMeyer, line 337: We should really serialize this as a list of strings, this is not exactly usable for users doing something with the JSON output

tests/test_config.py
```diff
@@ -1277,3 +1278,51 @@ def test_mkosi_version_executable(tmp_path: Path) -> None:
...
+
+    (d / "mkosi.conf").write_text(
+        """
+        [Output]
+        SplitArtifacts=yes
+        """
+    )
+
+    with chdir(d):
+        "mkosi.conf"
```
DaanDeMeyer, line 1326: This doesn't seem to add any value to me

mkosi/config.py
```diff
@@ -499,7 +499,18 @@ def native(cls) -> "Architecture":
...
-def parse_boolean(s: str) -> bool:
+class ArtifactOutput(enum.IntFlag):
+    uki = 1
+    kernel = 2
+    initrd = 4
+    partitions = 8
+
+    # Set a high bit to prevent these from showing up when only the previous values are set
+    no = 0b10000000000000000000000000000000 | uki | kernel | initrd
+    yes = 0b10000000000000000000000000000000 | uki | kernel | initrd | partitions
```
DaanDeMeyer, line 510: I wonder if this is just too much C for python and we're better off just having a set of enum values instead. That makes the JSON serialization easier as well.

DaanDeMeyer (changes requested):

mkosi/__init__.py
NekkoDroid: That would result in the uki always getting copied when it exists, even when it wasn't specified. Just conditioning the copy part and still return the path would feel a bit weird, but I could do that.

behrmann (approved):

mkosi/__init__.py
behrmann: I think it's good as is with the name change. I don't think `copy_uki` shouldn't copy the uki in some scenarios, that would just be false advertising. I could see just always copying out the UKI and having the kernel and initrd be options in addition to the uki. I think that would also be fair, but I lean more towards the way it is now.

DaanDeMeyer (approved):

PR #3138, opened 2024-10-23
Set qemu drive caching options for qemu drives as well

DaanDeMeyer:
We already set these for the scratch device, let's make sure we set them for any drives we allocate as well.

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -752,7 +752,7 @@ def finalize_drive(config: Config, drive: QemuDrive) -> Iterator[Path]:
         prefix=f"mkosi-drive-{drive.id}",
     ) as file:
         make_nocow(config, Path(file.name))
-        file.truncate(drive.size)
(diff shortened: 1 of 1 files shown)


behrmann (approved):
Looks reasonable, but in a bit further up these options are bound to
```python
direct = fname.stat().st_size % resource.getpagesize() == 0
ephemeral = config.ephemeral
```
Since the qemu drives might be used in a less scratchy way, could this make sense here, too?

DaanDeMeyer:
> Looks reasonable, but in a bit further up these options are bound to
> 
> ```python
> direct = fname.stat().st_size % resource.getpagesize() == 0
> ephemeral = config.ephemeral
> ```
> 
> Since the qemu drives might be used in a less scratchy way, could this make sense here, too?

We allocate these ourselves and remove them when the VM exits so they're always ephemeral. I pushed one more change to make sure the drives are aligned to the page size so we can always use cache.direct as well.

PR #3156, opened 2024-10-31
Map current user to root in subuid user namespace

DaanDeMeyer:
By mapping the current user to root in the subuid user namespace,
we don't have to change the ownership of all the files in the directory
tree to root in the subuid uid/gid range. This means that on btrfs
filesystems, we can do a subvolume snapshot instead of an expensive
full tree recursion to copy each file individually.

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -565,9 +565,11 @@ def copy_ephemeral(config: Config, src: Path) -> Iterator[Path]:
         yield src
         return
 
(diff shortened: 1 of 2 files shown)


septatrix (approved):
Boot time to init went down from ~6s to ~3s for the default Arch `mkosi-kernel` image (~2.2G, ~71k files) in some rudimentary testing

behrmann:
I looked at the fedora:fedora and centos:arch CI and the directory output fails at firstboot during boot.

PR #3158, opened 2024-10-31
Ensure user provided cmdline is always last

septatrix:
This is required for `--` to work such that one can stop the kernel from parsing further options and simply passing them as arguments to the init process

Open to suggestions on how to improve this. We could for example first collect all of mkosi's options and then directly before executing qemu append the user provided cmdline. This is just the quickest fix, not necessarily the cleanest ;) It also assumes that nothing depends on the order of the arguments which I am not sure is the case (e.g. what happens if two `systemd.set_credential_binary` with the same name defined - with this change the order would be reversed).

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -789,19 +789,7 @@ def finalize_state(config: Config, cid: int) -> Iterator[None]:
 
 
 def finalize_kernel_command_line_extra(config: Config) -> list[str]:
-    columns, lines = shutil.get_terminal_size()
-    term = finalize_term()
-
-    cmdline = [
-        "rw",
-        # Make sure we set up networking in the VM/container.
-        "systemd.wants=network.target",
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
@septatrix We can just move the `finalize_extra_command_line_extra()` call further down

septatrix:
> @septatrix We can just move the `finalize_extra_command_line_extra()` call further down

Then we would need to special case some things as e.g. `root=` currently depends on the finalized cmdline

DaanDeMeyer:
@septatrix You can generalize `finalize_kernel_command_line_extra()` to `finalize_kernel_command_line()` and move the if condition at the callsite into the function. Then you can evaluate against `finalize_kernel_command_line()` instead of `kcl` when determining `root=`.

septatrix:
Not sure if I understood your suggestion correctly. I feel like I might have done the exact opposite now...

PR #3159, opened 2024-10-31
Mount BuildDirectory= RW during mkosi.prepare

NekkoDroid:
This is to allow aquiring external resources which are not part of local sources (e.g. `makepkg --verifysource`) for use during a later build step in which internet connect is disabled.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -706,9 +706,6 @@ def run_prepare_scripts(context: Context, build: bool) -> None:
     if context.config.profiles:
         env["PROFILES"] = " ".join(context.config.profiles)
 
-    if context.config.build_dir is not None:
-        env |= dict(BUILDDIR="/work/build")
-
     env |= context.config.environment
 
     with (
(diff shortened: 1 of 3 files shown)


DaanDeMeyer:
So the problem with this is that if you nuke the build directory, the prepare scripts won't automatically be rerun when mkosi is invoked again. Last time I ran into this problem I found it to be much nicer to store this stuff in the build overlay of the image (this was rust stuff and I just used /usr/share/cargo in the buildroot to store all of the dependencies).

DaanDeMeyer:
I'm a bit confused why I even made the build directory accessible to prepare scripts in the first place, we should probably remove that

NekkoDroid:
> So the problem with this is that if you nuke the build directory, the prepare scripts won't automatically be rerun when mkosi is invoked again.

I assume you mean when building incrementally, right?

> I'm a bit confused why I even made the build directory accessible to prepare scripts in the first place, we should probably remove that

I can remove that in this PR then. Unsure how I am going to solve my problem but I'll maybe make a more detailed writedown in Matrix and it can be discussed there how I/we could proceed.

DaanDeMeyer:
@NekkoDroid Can you add a little blurb about this to the news file as well?

DaanDeMeyer (approved):

PR #3160, opened 2024-10-31
Fix qemu gui output on aarch64

davide125:
On aarch64 `-vga virtio` doesn't work, we need to use `-device virtio-gpu-pci` instead. However, unless we _also_ pass `console=tty0` the output only goes to the serial for some reason.

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -827,6 +827,8 @@ def finalize_kernel_command_line_extra(config: Config) -> list[str]:
             "console=hvc0",
             f"TERM={term}",
         ]
+    elif config.architecture.is_arm_variant():
+        cmdline += ["console=tty0"]
 
     for s in config.kernel_command_line_extra:
         key, sep, value = s.partition("=")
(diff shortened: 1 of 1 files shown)


davide125:
This still isn't quite right, as I'm getting systemd boot messages, but not the kernel ones.

mkosi/qemu.py
```diff
@@ -819,6 +819,9 @@ def finalize_kernel_command_line_extra(config: Config) -> list[str]:
         # CD-ROMs are read-only so tell systemd to boot in volatile mode.
         cmdline += ["systemd.volatile=yes"]
 
+    if config.qemu_gui and config.architecture.is_arm_variant():
+        cmdline += ["console=tty0"]
```
DaanDeMeyer, line 823: This should be moved to `finalize_kernel_command_line_extra()`

mkosi/qemu.py
```diff
@@ -1112,7 +1115,10 @@ def run_qemu(args: Args, config: Config) -> None:
     cmdline += ["-cpu", "max"]
 
     if config.qemu_gui:
-        cmdline += ["-vga", "virtio"]
+        if config.architecture.is_x86_variant():
+            cmdline += ["-vga", "virtio"]
```
DaanDeMeyer, line 1119: Let's always use this as the fallback, no clue if it works on s390 or power or not but we might as well try this instead of doing nothing at all

DaanDeMeyer (changes requested):

mkosi/qemu.py
davide125: It's already in that function

behrmann (approved):

DaanDeMeyer:
Pushed a small nit fix, setting green label

PR #3161, opened 2024-11-01
Add support for bootctl secure boot auto enrollment 

DaanDeMeyer:
Matching PR for https://github.com/systemd/systemd/pull/34948

### DIFF
--- a/mkosi/bootloader.py
+++ b/mkosi/bootloader.py
@@ -5,7 +5,6 @@ import subprocess
 import sys
 import tempfile
 import textwrap
-import uuid
 from collections.abc import Iterator, Sequence
 from pathlib import Path
 from typing import Optional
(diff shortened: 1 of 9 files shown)


mkosi/bootloader.py
```diff
@@ -730,7 +770,7 @@ def install_systemd_boot(context: Context) -> None:
                 run(
                     [
                         "sbsiglist",
-                        "--owner", str(uuid.uuid4()),
+                        "--owner", "00000000-0000-0000-0000-000000000000",
```
behrmann, line 773: Can this have any downside? Maybe we could generate this from our seed if one is set?

DaanDeMeyer: This seems to be purely information, from the UEFI spec:

> Each signature has an owner SignatureOwner, which is a GUID identifying the agent which inserted the signature in the database. Agents might include the operating system or an OEM-supplied driver or application. Agents may examine this field to understand whether they should manage the signature or not.
_images/Secure_Boot_and_Driver_Signing-6.png

behrmann: Makes sense, but then we could also just choose one randomly once and document it as the mkosi SignatureOwner GUID? Though, maybe it would also make sense to have it in systemd instead or to expose it as a knob. For now this is probably good enough.

PR #3164, opened 2024-11-02
Add sandbox verb

DaanDeMeyer:
In systemd, to run the integration tests, we need to run meson on
the host which will itself invoke mkosi to run the integration tests.
This means all the dependencies to run meson need to be installed on
the host. This doesn't just mean meson needs to be installed, but also
a compiler and various required build dependencies of systemd to allow
building tools invoked by mkosi to build the image.

To avoid having to install these dependencies on the host system, let's
introduce a sandbox verb which runs a command in the mkosi sandbox that's
also used by other verbs such as boot, qemu and shell. This then allows
extra required tools to be installed in the tools tree via ToolsTreePackages=
to allow running these commands without having to install them on the host
system.


mkosi/resources/man/mkosi.1.md
```diff
@@ -135,6 +137,13 @@ The following command line verbs are known:
     specified after the `sysupdate` verb are passed directly to
     `systemd-sysupdate` invocation.
 
+`sandbox`
+:   Run arbitrary commands inside of the same sandbox used to execute
+    other verbs such as `boot`, `shell`, `qemu` and more. This generally
```
behrmann, line 142:
```suggestion
    other verbs such as `boot`, `shell`, `qemu` and more. Thus generally
```
or
```suggestion
    other verbs such as `boot`, `shell`, `qemu` and more. Therefore generally
```
or
```suggestion
    other verbs such as `boot`, `shell`, `qemu` and more. Generally
```

mkosi/config.py
behrmann, line 1: The verb also needs to be added to the help output around line 3630.

mkosi/run.py
```diff
@@ -554,10 +563,23 @@ def sandbox_cmd(
...
         "--setenv",
         "PATH",
-        f"/scripts:{'/usr/bin:/usr/sbin' if tools != Path('/') else os.environ['PATH']}",
+        ":".join(
+            [
+                *(["/scripts"] if scripts else []),
+                "/usr/bin",
+                "/usr/sbin",
+                *(
+                    [s for s in os.environ["PATH"].split(":") if home and s.startswith(os.fspath(home))]
```
behrmann, line 577: I think it makes sense as written in the commit message, but it's also a potential footgun. We should definitely document this in the man page as well as in the NEWS.

behrmann (approved):

PR #3168, opened 2024-11-05
Add support for systemd-sbsign

DaanDeMeyer:
Matching PR for https://github.com/systemd/systemd/pull/35021

### DIFF
--- a/mkosi/bootloader.py
+++ b/mkosi/bootloader.py
@@ -5,15 +5,18 @@ import subprocess
 import sys
 import tempfile
 import textwrap
-from collections.abc import Iterator, Sequence
+from collections.abc import Iterator, Mapping, Sequence
 from pathlib import Path
 from typing import Optional
 
 from mkosi.config import (
     BiosBootloader,
     Bootloader,
+    CertificateSource,
+    CertificateSourceType,
     Config,
     ConfigFeature,
+    KeySource,
     KeySourceType,
     OutputFormat,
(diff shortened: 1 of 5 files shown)


mkosi/bootloader.py
```diff
@@ -501,17 +501,63 @@ def sign_efi_binary(context: Context, input: Path, output: Path) -> Path:
...
-    if (
+    sbsign = context.config.find_binary("systemd-sbsign", "/usr/lib/systemd/systemd-sbsign")
+    if context.config.secure_boot_sign_tool == SecureBootSignTool.systemd and not sbsign:
+        die("Could not find systemd-sbsign")
+
+    cmd: list[PathString]
+    options: list[PathString]
+
+    if context.config.secure_boot_sign_tool == SecureBootSignTool.systemd or (
+        context.config.secure_boot_sign_tool == SecureBootSignTool.auto and sbsign
```
behrmann, line 512: Haven't checked this, but
```suggestion
        context.config.secure_boot_sign_tool == SecureBootSignTool.auto and sbsign is not None
```
should make the assert below unnecessary and would be more symmetric with the predicate in the other branch.

behrmann (approved):

mkosi/bootloader.py
DaanDeMeyer: Doesn't make the assert unnecessary unfortunately

PR #3169, opened 2024-11-05
Add mkosi.initrd.extra files to initrd.

plastikos:
This adds the processing of mkosi.initrd.extra - similar to mkosi.extra but for adding files to the initrd.


DaanDeMeyer:
@plastikos We don't really want to duplicate more settings for the default initrd. What you can do instead is build the initrd as `mkosi.images/initrd` and then you can use all the existing mkosi settings for your initrd. For example we do this in the systemd repo: https://github.com/systemd/systemd/tree/main/mkosi.images/initrd. You can use the initrd in your main image with the `Initrds=` setting (see https://github.com/systemd/systemd/blob/main/mkosi.conf#L53).

plastikos:
@DaanDeMeyer, I'll check it out.

DaanDeMeyer:
> @DaanDeMeyer, I'll check it out. My concern is if I want a _small_ change to the built-in `initrd` build do I have to replicate all of the details or can I simply add the small difference.

@plastikos You add `Include=mkosi-initrd` to the config in mkosi.images/initrd and you get the default initrd almost the same as you would now.

plastikos:
Redundant to existing functionality.

PR #3170, opened 2024-11-05
Run FinalizeScripts= when using Format=none

NekkoDroid:
This is to allow copying any build files (like packages) to the output directory after the build, so that all the related logic is contained in a single image and this doesn't need to be done in a separate image.


DaanDeMeyer:
The thing about `Format=none` is that it can't touch the output directory so we don't have to clean it up beforehand. I don't think we want to get rid of that. Basically, `Format=none` cannot produce outputs no matter what. Anything it produces should be accessed via the build directory.

NekkoDroid:
I guess that does somewhat make sense.

PR #3172, opened 2024-11-08
[RFC] mkosi-initrd: inject breakpoints into the boot process

aafeijoo-suse:
Added three breakpoints that can stop the boot process and open a sulogin shell by passing the `rd.break=` option on the kernel command line:

- `rd.break=pre-udev`: at the very beginning, before systemd-udevd.service starts.
- `rd.break=pre-mount`: before the real root is mounted, i.e., before sysroot.mount starts.
- `rd.break=pre-switch-root`: at the end, before switching root.

---

Before discussing whether these options are too many, too few, or their names are appropriate, here is what other initrd generators are offering:

|  | option | values |
|-|-|-|
| dracut | rd.break | cmdline, pre-udev, pre-trigger, initqueue, pre-mount, mount, pre-pivot/cleanup (default) |
| initramfs-tools | break | top, modules, premount (default), mount, mountroot, bottom, init |
| mkinitcpio | break | premount (default), postmount |

### DIFF
--- a/mkosi/resources/mkosi-initrd/mkosi.extra/usr/lib/systemd/system-preset/10-mkosi-initrd.preset
+++ b/mkosi/resources/mkosi-initrd/mkosi.extra/usr/lib/systemd/system-preset/10-mkosi-initrd.preset
@@ -1,0 +1,5 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+enable mkosi-initrd-pre-mount.service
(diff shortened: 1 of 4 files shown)


DaanDeMeyer:
I think this is material for systemd instead of mkosi-initrd.

aafeijoo-suse:
Ok, I'll submit a similar patch there, I guess for v258.

PR #3178, opened 2024-11-11
GitHub Action: Install all required mkosi dependencies

davide125:
This is needed for `PackageDirectories` to work.

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -56,6 +56,10 @@ runs:
         sudo aa-teardown || true
         sudo apt-get remove apparmor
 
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
So we should go all the way and run `mkosi dependencies | xargs -d '\n' sudo apt-get install` to install all required dependencies to run mkosi.

davide125:
This accidentally dropped the `--assume-yes --no-install-recommends`. Fixing that in https://github.com/systemd/mkosi/pull/3185

PR #3179, opened 2024-11-11
Make sure user provided cmdline is always last

DaanDeMeyer:
Replaces #3158

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -1185,18 +1185,11 @@ def run_qemu(args: Args, config: Config) -> None:
 
         apply_runtime_size(config, fname)
 
-        if kernel and (
-            KernelType.identify(config, kernel) != KernelType.uki
-            or not config.architecture.supports_smbios(firmware)
-        ):
-            kcl = config.kernel_command_line + finalize_kernel_command_line_extra(config)
-        else:
-            kcl = finalize_kernel_command_line_extra(config)
-
+        kcl = []
         if kernel:
             cmdline += ["-kernel", kernel]
 
-            if any(s.startswith("root=") for s in kcl):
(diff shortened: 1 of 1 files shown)


septatrix (approved):

mkosi/qemu.py
```diff
@@ -788,11 +788,19 @@ def finalize_state(config: Config, cid: int) -> Iterator[None]:
             p.unlink(missing_ok=True)
 
 
-def finalize_kernel_command_line_extra(config: Config) -> list[str]:
+def finalize_kernel_command_line(config: Config, kernel: Optional[Path], firmware: QemuFirmware) -> list[str]:
```
behrmann, line 791: That's what ruff is complaining about.
```suggestion
def finalize_kernel_command_line(
    config: Config, 
    kernel: Optional[Path], 
    firmware: QemuFirmware,
) -> list[str]:
```

behrmann (approved):

mkosi/qemu.py
```diff
@@ -788,6 +788,22 @@ def finalize_state(config: Config, cid: int) -> Iterator[None]:
...
+def finalize_kernel_command_line(
+    config: Config,
+    kernel: Optional[Path],
+    firmware: QemuFirmware,
+) -> list[str]:
+    if kernel and (
+        KernelType.identify(config, kernel) != KernelType.uki
+        or not config.architecture.supports_smbios(firmware)
+    ):
+        cmdline = [*config.kernel_command_line]
```
septatrix, line 800: I actually wondered why this is done. It obviously is required for the eventual call to qemu but prepending it this early leads to a different behavior when directly booting a kernel vs normal boot if the normal cmdline contains a `root=` (as in one case it will be picked up by the check below and in the other case it will not)

PR #3182, opened 2024-11-12
Add %I specifier for subimages

NekkoDroid:
This expands to the name of the subimage (and an empty string for the main image) and is the same value that is used for `Dependencies=`. Also exports it as an env var to scripts for easier access.

Fixes: #2566

### DIFF
--- a/mkosi/resources/man/mkosi.1.md
+++ b/mkosi/resources/man/mkosi.1.md
@@ -2031,11 +2031,12 @@ use `%%`. The following specifiers are understood:
 
 There are also specifiers that are independent of settings:
 
-| Specifier | Value                                   |
-|-----------|-----------------------------------------|
-| `%C`      | Parent directory of current config file |
-| `%P`      | Current working directory               |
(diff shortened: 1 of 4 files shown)


mkosi/resources/man/mkosi.1.md
```diff
@@ -2036,6 +2036,7 @@ There are also specifiers that are independent of settings:
 | `%C`      | Parent directory of current config file |
 | `%P`      | Current working directory               |
 | `%D`      | Directory that mkosi was invoked in     |
+| `%I`      | The name of the current subimage        |
```
behrmann, line 2039: Shouldn't that go in the table above?

behrmann (commented):
LGTM.

mkosi/resources/man/mkosi.1.md
NekkoDroid: I wouldn't say so, as `config.image` isn't controllable via a setting

DaanDeMeyer (approved):

PR #3183, opened 2024-11-12
Fix man page reference to --tools-tree-packages

davide125:
This makes it match the man page.


behrmann:
The fix should actually be the other way around and fix the man page. We do singular for the command line option and plural for the config file options. I think the idea is, that it's more natural to do `--package --package bar`, even though we accept `--package foo,bar`. I don't know whether this is a very good rationale, but at least it's mostly consistent.

davide125:
Thanks for the context @behrmann, I've updated this to fix the manpage instead.

PR #3193, opened 2024-11-15
Add sound device when QemuGui=true

davide125:
Using `sdl` here as I think it's the most compatible option.


davide125:
Updated to use pipewire, it seems to work the same, except the default volume is a bit quieter compared to sdl for some reason.

DaanDeMeyer (approved):

Issue #3189, opened 2024-11-14
Ubuntu 24.04 images fail to build


dillona:
### mkosi commit the issue has been seen with

653adbeae37bc9bd3ef1884e3467646425aba472

### Used host distribution

Ubuntu 24.04

### Used target distribution

Ubuntu 24.04

### Linux kernel version used

6.8.0-1016-aws

### CPU architectures issue was seen on

None

### Unexpected behaviour you saw

Ubuntu 24.04 images fail to build.

In #16 `dbus` and `libpam-systemd` were added to the `debootstrap` invocation, however in #1442 this was effectively removed with the suggestion to add these packages manually to the `Packages` list. 

This does not seem to work. If I edit `mkosi/distributions/debian.py` to add those packages to the `debootstrap`-equivalent installation, my build completes successfully.

### Used mkosi config

```ini
[Distribution]
Distribution=ubuntu
Release=noble
Architecture=x86-64
Repositories=universe

[Content]
Bootable=no

[Output]
Format=tar
CompressOutput=zstd

[Content]
Bootable=no
Packages=ubuntu-desktop,dbus,libpam-systemd
WithRecommends=yes
```

### mkosi output

```sh
[ beginning omitted to fit within Github limits ]
Setting up libdbus-1-3:amd64 (1.14.10-4ubuntu4.1) ...
Setting up dbus-bin (1.14.10-4ubuntu4.1) ...
Setting up dbus-session-bus-common (1.14.10-4ubuntu4.1) ...
Setting up dbus-daemon (1.14.10-4ubuntu4.1) ...
/usr/lib/tmpfiles.d/dbus.conf:13: Failed to resolve user 'messagebus': No such process
Setting up dbus-system-bus-common (1.14.10-4ubuntu4.1) ...
Setting up dbus (1.14.10-4ubuntu4.1) ...
Setting up systemd-sysv (255.4-1ubuntu8.4) ...
Setting up libpam-systemd:amd64 (255.4-1ubuntu8.4) ...
Setting up dbus-user-session (1.14.10-4ubuntu4.1) ...
Setting up snapd (2.65.3+24.04) ...
Created symlink /etc/systemd/system/multi-user.target.wants/snapd.apparmor.service → /usr/lib/systemd/system/snapd.apparmor.service.
Created symlink /etc/systemd/system/multi-user.target.wants/snapd.autoimport.service → /usr/lib/systemd/system/snapd.autoimport.service.
[...]

PR #3200, opened 2024-11-15
Use passwd symlinks instead of bind mounts

DaanDeMeyer:
Bind mounts don't reflect changes to the original files if they're
replaced instead of modified. Let's use symlinks instead so that
changes to the original files are always reflected.

Fixes https://github.com/systemd/mkosi/issues/3189

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -1818,6 +1818,7 @@ class Config:
     tools_tree_repositories: list[str]
     tools_tree_sandbox_trees: list[ConfigTree]
     tools_tree_packages: list[str]
+    tools_tree_package_directories: list[Path]
     tools_tree_certificates: bool
     incremental: Incremental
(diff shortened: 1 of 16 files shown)


mkosi/config.py
```diff
@@ -3166,6 +3167,14 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
         parse=config_make_list_parser(delimiter=","),
         help="Add additional packages to the default tools tree",
     ),
+    ConfigSetting(
+        dest="tools_tree_package_directories",
+        long="--tools-tree-package-directory",
```
behrmann, line 3172: Depending on the direction of #3201, this should maybe be `--tools-tree-package-dir`.

behrmann (commented):
Typo in the second commit message, otherwise LGTM.

PR #3201, opened 2024-11-16
Rename some opts to be more consistent

NekkoDroid:
Just noticed some directory long opts used `dir=` while others used `directory=`, should all now be consistently `directory=`.


behrmann (approved):
I like the longer name a bit better and changing it from `-directory` to `-dir` leaves the `--directory` option, but it's still more consistent than the status quo.

NekkoDroid:
I am also fine with changing them all to use the longer `*-directory=`

behrmann:
Let's have @DaanDeMeyer weigh in, since he just added another. Either way this PR will need a rebase.

NekkoDroid:
Changed them to the longer `*-directory=`. I think I got most of them all with the exception of `--wipe-build-dir` as that (from what I see) doesn't have a way to keep backwards compat.

DaanDeMeyer (approved):

PR #3207, opened 2024-11-18
Assorted fixes for mkosi-initrd and kernel-install

bluca:


mkosi/initrd.py
```diff
@@ -38,6 +38,13 @@ def main() -> None:
         help="Kernel version string",
         default=platform.uname().release,
     )
+    parser.add_argument(
+        "--kernel-image",
+        metavar="KERNEL_IMAGE",
+        help="Kernel image",
+        type=Path,
+        nargs="?",
```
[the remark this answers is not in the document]
DaanDeMeyer: Or should be replaced with default=None.

DaanDeMeyer (changes requested):

mkosi/initrd.py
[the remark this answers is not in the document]
behrmann: I'd prefer the first option, since the implicit default is always `None`

mkosi/initrd.py
```diff
@@ -111,6 +118,12 @@ def main() -> None:
             "--include=mkosi-initrd",
         ]  # fmt: skip
 
+        if args.kernel_image:
+            cmdline += [
+                "--extra-tree",
+                f"{args.kernel_image}:/usr/lib/modules/{args.kernel_version}/vmlinuz",
+            ]
```
behrmann, line 125:
```suggestion
            cmdline += [
                "--extra-tree", f"{args.kernel_image}:/usr/lib/modules/{args.kernel_version}/vmlinuz",
            ]  # fmt: skip
```
since we usually keep flags and their arguments on the same line.

kernel-install/50-mkosi.install
```diff
@@ -153,20 +154,21 @@ def main() -> None:
...
-    (context.staging_area / output).unlink()
+    run(cmdline, stdin=sys.stdin, stdout=sys.stdout)
 
     if format == OutputFormat.cpio:
         build_microcode_initrd(context.staging_area / "microcode")
     else:
-        (context.staging_area / f"{output}.vmlinuz").unlink()
-        (context.staging_area / f"{output}.initrd").unlink()
+        (context.staging_area / f"{output}.vmlinuz").unlink(missing_ok=True)
+        (context.staging_area / f"{output}.initrd").unlink(missing_ok=True)
```
DaanDeMeyer, line 171: This can just be dropped and instead you can add `SplitArtifacts=` to `mkosi-initrd/mkosi.conf` in the `[Output]` section.

DaanDeMeyer (changes requested):

kernel-install/50-mkosi.install
bluca: but why should it fail if it's not there? It doesn't seem to be used

behrmann:
Can you squash the last to commits?

DaanDeMeyer (approved):

PR #3213, opened 2024-11-20
ci: make the output of the formatting job actionable

behrmann:
Since @bluca didn't know how to handle the error, let's make this clearer.

With this change, there will be no output by default and if e.g. `kmod.py` would be reformatted, it would look like this
```
Would reformat: mkosi/kmod.py
Please run 'ruff format' on the above files
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,11 @@ jobs:
       - name: Run ruff format
         run: |
           ruff --version
-          ruff format --check mkosi/ tests/ kernel-install/50-mkosi.install
+          if ! ruff format --check --quiet mkosi/ tests/ kernel-install/50-mkosi.install
+          then
+              echo "Please run 'ruff format' on the above files or apply the diffs below manually"
+              ruff format --check --quiet --diff mkosi/ tests/ kernel-install/50-mkosi.install
+          fi
 
       - name: Check that tabs are not used in code
         run: sh -c '! git grep -P "\\t" "*.py"'


bluca:
It doesn't work here:

```
$ ruff format mkosi
warning: `ruff format` is a work-in-progress, subject to change at any time, and intended only for experimentation.
ruff failed
  Cause: TOML parse error at line 79, column 1
   |
79 | lint.select = ["E", "F", "I", "UP"]
   | ^^^^
unknown field `lint`, expected one of `allowed-confusables`, `builtins`, `cache-dir`, `dummy-variable-rgx`, `exclude`, `extend`, `extend-exclude`, `extend-include`, `extend-ignore`, `extend-select`, `extend-fixable`, `extend-unfixable`, `external`, `fix`, `fix-only`, `fixable`, `output-format`, `force-exclude`, `ignore`, `ignore-init-module-imports`, `include`, `line-length`, `tab-size`, `logger-objects`, `required-version`, `respect-gitignore`, `select`, `show-source`, `show-fixes`, `src`, `namespace-packages`, `target-version`, `preview`, `task-tags`, `typing-modules`, `unfixable`, `flake8-annotations`, `flake8-bandit`, `flake8-bugbear`, `flake8-builtins`, `flake8-comprehensions`, `flake8-copyright`, `flake8-errmsg`, `flake8-quotes`, `flake8-self`, `flake8-tidy-imports`, `flake8-type-checking`, `flake8-gettext`, `flake8-implicit-str-concat`, `flake8-import-conventions`, `flake8-pytest-style`, `flake8-unused-arguments`, `isort`, `mccabe`, `pep8-naming`, `pycodestyle`, `pydocstyle`, `pyflakes`, `pylint`, `pyupgrade`, `format`, `per-file-ignores`, `extend-per-file-ignores`
```

can it show the full diff instead, so that it can be applied manually?

behrmann:
Geez... ruff in Debian is way behind the curve 0.0.291+dfsg1-4 vs 0.7.4 on Arch. Looking at the above error, it already fails at parsing it's config file.

I mean, we both love the pain of running Debian stable, but this is the point where I recommend using for `pipx install ruff` (because `uv` is not a thing yet in Debian…) for stuff like this.

behrmann:
With the version now the ouput would be something like this
```
Would reformat: mkosi/kmod.py
Please run 'ruff format' on the above files or apply the diffs below manually
--- mkosi/kmod.py
+++ mkosi/kmod.py
@@ -15,7 +15,8 @@

 def loaded_modules() -> list[str]:
     # Loaded modules are listed with underscores but the filenames might use dashes instead.
-    return [rf"/{line.split()[0].replace('_', '[_-]')}\.ko"
+    return [
+        rf"/{line.split()[0].replace('_', '[_-]')}\.ko"
         for line in Path("/proc/modules").read_text().splitlines()
     ]


```

bluca (approved):

bluca:
> Geez... ruff in Debian is way behind the curve 0.0.291+dfsg1-4 vs 0.7.4 on Arch. Looking at the above error, it already fails at parsing it's config file.
> 
> I mean, we both love the pain of running Debian stable, but this is the point where I recommend using for `pipx install ruff` (because `uv` is not a thing yet in Debian…) for stuff like this.

This is not even stable, it's from unstable. I see that it's a rust program, so that's probably why, rust stuff is always broken. Why oh why did they decide to write a python linter in rust?

DaanDeMeyer:
> > Geez... ruff in Debian is way behind the curve 0.0.291+dfsg1-4 vs 0.7.4 on Arch. Looking at the above error, it already fails at parsing it's config file.
> > I mean, we both love the pain of running Debian stable, but this is the point where I recommend using for `pipx install ruff` (because `uv` is not a thing yet in Debian…) for stuff like this.
> 
> This is not even stable, it's from unstable. I see that it's a rust program, so that's probably why, rust stuff is always broken. Why oh why did they decide to write a python linter in rust?

Everything written before in python was slow as molasses

behrmann:
> Why oh why did they decide to write a python linter in rust?

Because gotta go fast.

The Python dev tooling right now is fast become all-rust and `ruff` (for linting and formatting) and `uv` (for package management, virtual environments and now apparently even bootstrapping having a Python interpreter) seem to be entrenching themselves. 

The choice of Rust for this is or seems somewhat unnatural and makes the distros' lives way more difficult, but I do understand the value for "regular people".

behrmann:
Oh and the people behind ruff and uv are working on a type checker, too. That's actually something I'm looking forward to.

bluca:
As long as the CI outputs exactly what to do I don't mind, given that it's someone else's job to keep that stuff working :-P

PR #3215, opened 2024-11-20
Make ConfigSetting generic

septatrix:
This prevents one from accidentally using incompatible parse/match/default_factory callbacks. A few type ignores are necessary due to python/mypy#3737 and in `finalize_config` because any inference is impossible there

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -28,7 +28,7 @@ import uuid
 from collections.abc import Collection, Iterable, Iterator, Sequence
 from contextlib import AbstractContextManager
 from pathlib import Path
-from typing import Any, Callable, Optional, TypeVar, Union, cast
+from typing import Any, Callable, Generic, Optional, TypeVar, Union, cast
 
 from mkosi.distributions import Distribution, detect_distribution
 from mkosi.log import ARG_DEBUG, ARG_DEBUG_SANDBOX, ARG_DEBUG_SHELL, Style, die
@@ -48,11 +48,11 @@ from mkosi.util import (
 from mkosi.versioncomp import GenericVersion
 
 T = TypeVar("T")
+SE = TypeVar("SE", bound=StrEnum)
 
-ConfigParseCallback = Callable[[Optional[str], Optional[Any]], Any]
-ConfigMatchCallback = Callable[[str, Any], bool]
-ConfigDefaultCallback = Callable[[argparse.Namespace], Any]
-
+ConfigParseCallback = Callable[[Optional[str], Optional[T]], Optional[T]]
+ConfigMatchCallback = Callable[[str, T], bool]
+ConfigDefaultCallback = Callable[[argparse.Namespace], T]
 
 BUILTIN_CONFIGS = ("mkosi-tools", "mkosi-initrd", "mkosi-vm")
 
@@ -676,7 +676,7 @@ def config_match_build_sources(match: str, value: list[ConfigTree]) -> bool:
     return Path(match.lstrip("/")) in [tree.target for tree in value if tree.target]
 
 
-def config_make_list_matcher(parse: Callable[[str], T]) -> ConfigMatchCallback:
+def config_make_list_matcher(parse: Callable[[str], T]) -> ConfigMatchCallback[list[T]]:
     def config_match_list(match: str, value: list[T]) -> bool:
         return parse(match) in value
 
@@ -687,7 +687,7 @@ def config_parse_string(value: Optional[str], old: Optional[str]) -> Optional[st
     return value or None
 
 
-def config_make_string_matcher(allow_globs: bool = False) -> ConfigMatchCallback:
+def config_make_string_matcher(allow_globs: bool = False) -> ConfigMatchCallback[str]:
     def config_match_string(match: str, value: str) -> bool:
         if allow_globs:
             return fnmatch.fnmatchcase(value, match)
@@ -906,8 +906,8 @@ def config_default_proxy_url(namespace: argparse.Namespace) -> Optional[str]:
     return None
 
 
-def make_enum_parser(type: type[StrEnum]) -> Callable[[str], StrEnum]:
-    def parse_enum(value: str) -> StrEnum:
+def make_enum_parser(type: type[SE]) -> Callable[[str], SE]:
+    def parse_enum(value: str) -> SE:
         try:
             return type(value)
         except ValueError:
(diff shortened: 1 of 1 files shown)


mkosi/config.py
```diff
@@ -1010,12 +1008,12 @@ def config_match_version(match: str, value: str) -> bool:
...
     delimiter: Optional[str] = None,
-    parse: Callable[[str], tuple[str, Any]],
+    parse: Callable[[str], tuple[str, str]],
     unescape: bool = False,
     allow_paths: bool = False,
     reset: bool = True,
-) -> ConfigParseCallback:
-    def config_parse_dict(value: Optional[str], old: Optional[dict[str, Any]]) -> Optional[dict[str, Any]]:
+) -> ConfigParseCallback[dict[str, str]]:
+    def config_parse_dict(value: Optional[str], old: Optional[dict[str, str]]) -> Optional[dict[str, str]]:
```
septatrix, line 1016: The change from `Any` to `str` was done because otherwise the parser is unsound if `allow_paths=True`. If this should ever be required we would likely need to split `parse` into `splitter: str -> (key: str, val: str)` and `parser: (val: str) -> (val: T)`. This change only enforces it at the type level

mkosi/config.py
```diff
@@ -2160,7 +2159,7 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
         yield section, "", ""
 
 
-PE_ADDON_SETTINGS = (
+PE_ADDON_SETTINGS: tuple[ConfigSetting[Any], ...] = (
```
behrmann, line 2162: I don't think we win anything by this being tuples, so if we make these lists, the typing becomes a nicer
 ```suggestion
PE_ADDON_SETTINGS: list[ConfigSetting[Any]] = [
```

File: mkosi/config.py [hunk 3]
```diff
@@ -948,11 +946,11 @@ def config_match_enum(match: str, value: StrEnum) -> bool:
 def config_make_list_parser(
     *,
     delimiter: Optional[str] = None,
-    parse: Callable[[str], Any] = str,
+    parse: Callable[[str], T] = str,  # type: ignore # see mypy#3737
```
behrmann, line 949: I dislike type ignores and an explicit `cast(Callable[[str], T], str)` would work around this, but in this case this is probably not worth it here, but let's not link to the issue but make this
```suggestion
    parse: Callable[[str], T] = str,  # type: ignore # TODO: use PEP696 default type parameter once we are on 3.12
```
same in the other places.

mkosi/config.py
```diff
@@ -4062,19 +4061,19 @@ def parse_new_includes(self) -> None:
...
+                return (getattr(self.config, setting.dest, None) or []) + v  # type: ignore
             elif isinstance(v, dict):
-                return (getattr(self.config, setting.dest, None) or {}) | v
+                return (getattr(self.config, setting.dest, None) or {}) | v  # type: ignore
             elif isinstance(v, set):
-                return (getattr(self.config, setting.dest, None) or set()) | v
+                return (getattr(self.config, setting.dest, None) or set()) | v  # type: ignore
             else:
-                return v
+                return v  # type: ignore
```
behrmann, line 4076: I think this is a place where we can do explicit casts instead, since we can guarantee more than just ignoring this. Same in the next hunk.

behrmann:
One thing I forgot: This should probably be checked on Python 3.9 to ensure this doesn't introduce anything by chance that fails on 3.9.

mkosi/config.py
septatrix: Sadly PEP 696 does not solve our problem, the issue/pep name is a bit confusing. The PEP is about global defaults for the TypeVar whereas what we need is inference of T based on default values for parameters whose annotation contains T

mkosi/config.py
septatrix: I tried that but problem is we would need to cast to `list[something]` but we cannot extract that "something" from T (and casting to T also does not work). I could take another look but I played around with this a bit to no avail

septatrix:
> One thing I forgot: This should probably be checked on Python 3.9 to ensure this doesn't introduce anything by chance that fails on 3.9.

Mypy and pyright both assume Python 3.9 in CI

mkosi/config.py
```diff
@@ -1422,7 +1421,7 @@ class ConfigSetting:
     # settings for argparse
     short: Optional[str] = None
     long: str = ""
-    choices: Optional[Any] = None
+    choices: Optional[list[Any]] = None
```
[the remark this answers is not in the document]
septatrix: I am not sure I follow, `StrEnum.values()` does return a `list[str]`?

mkosi/config.py
septatrix: I got it a bit more strict but currently type checkers are still not really able to follow the code / Pythons type system is not capable enough to express what we are doing here. It _could_ be possible to do some trickery with `TypeVar`s, several `@overloads`, and putting the collection merging into helper functions but I failed to make that work in an acceptable amount of time so I will let this rest for now

mkosi/config.py
```diff
@@ -1634,11 +1633,13 @@ class UKIProfile:
     cmdline: list[str]
 
 
-def make_simple_config_parser(settings: Sequence[ConfigSetting], type: type[Any]) -> Callable[[str], Any]:
+def make_simple_config_parser(
+    settings: Sequence[ConfigSetting[Any]], valtype: type[T]
```
DaanDeMeyer, line 1637: I think `object` is better than `Any` here since it doesn't disable type checking, ditto elsewhere

mkosi/config.py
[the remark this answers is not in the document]
septatrix: This is just mirroring the argument of argparse which allows anything that can be turned into a string, e.g. ints

mkosi/config.py
```diff
@@ -1654,7 +1655,7 @@ def finalize_value(config: argparse.Namespace, setting: ConfigSetting) -> None:
 
         setattr(config, setting.dest, default)
 
-    def parse_simple_config(value: str) -> Any:
+    def parse_simple_config(value: str) -> T:
```
DaanDeMeyer, line 1658: This can be `object` as well instead of `Any` I think, same elsewhere

mkosi/config.py
```diff
@@ -2160,7 +2163,7 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
         yield section, "", ""
 
 
-PE_ADDON_SETTINGS = (
+PE_ADDON_SETTINGS: list[ConfigSetting[Any]] = [
```
DaanDeMeyer, line 2166: This can be `Any` as well I think same elsewhere

mkosi/config.py
```diff
@@ -1634,11 +1633,13 @@ class UKIProfile:
     cmdline: list[str]
 
 
-def make_simple_config_parser(settings: Sequence[ConfigSetting], type: type[Any]) -> Callable[[str], Any]:
+def make_simple_config_parser(
+    settings: Sequence[ConfigSetting[object]], valtype: type[T]
```
DaanDeMeyer, line 1637: Let's put these on separate lines by adding a trailing comma

behrmann:
> Mypy and pyright both assume Python 3.9 in CI

Yeah, but type checking and actually running without throwing an exception (e.g. just the help output), are two different things, unfortunately. We've had plenty of cases where successfully type-checked code crashed because of circular imports. :)

mkosi/config.py
behrmann: Mhm, but shouldn't this work here? e.g. defining `ParserFunc` as alias for `Callable[[str], T]` with a default of `Callable[[str], str]` (that would still break, because `str` is seen as type and not as a function, so we'd need an `idparser = lambda s: s`, haven't checked this yet, though). I haven't read the full issue, but from the most recent posts I got the impression it would be closed due to PEP696.

mkosi/config.py
septatrix: This is any. And using object here breaks because the members are not homogeneous and inference goes haywire

mkosi/config.py
septatrix: The issue was reopened because it is not sufficiently solved by PEP696 and it's my understanding that our use case if one of those not solved by the PEP

behrmann (approved):

DaanDeMeyer (approved):

PR #3218, opened 2024-11-21
mkosi-initrd: Include extra modules

DaanDeMeyer:
These are all modules that the kernel tries to load while in the initrd on my laptop. All of these seem generic enough to include by default.

### DIFF
--- a/mkosi/resources/mkosi-initrd/mkosi.conf
+++ b/mkosi/resources/mkosi-initrd/mkosi.conf
@@ -45,6 +45,8 @@ KernelModulesInclude=
                     /autofs4.ko
                     /binfmt_misc.ko
                     /btrfs.ko
+                    /cdrom.ko
+                    /cfg80211.ko
                     /configfs.ko
                     /dm-crypt.ko
                     /dm-integrity.ko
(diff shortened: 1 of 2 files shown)


mkosi/resources/mkosi-initrd/mkosi.conf
```diff
@@ -53,18 +55,28 @@ KernelModulesInclude=
                     /dm-raid.ko
                     /dm-verity.ko
                     /dmi-sysfs.ko
+                    /drm_buddy.ko
                     /efi-pstore.ko
                     /efivarfs.ko
                     /erofs.ko
                     /ext4.ko
+                    /i2c-smbus.ko
+                    /intel_vsec.ko
```
behrmann, line 64: Probably doesn't hurt, but this doesn't seem generic for AMD and probably only makes sense for x86_64?

behrmann (approved):

mkosi/resources/mkosi-initrd/mkosi.conf
DaanDeMeyer: On non x86-64 it simply won't be there and will be skipped. And yes this is intel specific but that just means we should also include the AMD modules, since this should be generic

PR #3219, opened 2024-11-21
sysupdate: Make it work in combination with a tools tree

DaanDeMeyer:
- We have to make sure systemd-sysupdate looks at the os-release from the host even when using a tools tree.
- systemd-sysupdate can't detect the root block device when running with a tools tree. Let's abuse /run/systemd/volatile-root to shortcut the detection logic instead.

### DIFF
--- a/mkosi/sysupdate.py
+++ b/mkosi/sysupdate.py
@@ -1,7 +1,9 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
 import os
+import subprocess
 import sys
+import tempfile
 from pathlib import Path
 
 from mkosi.config import Args, ArtifactOutput, Config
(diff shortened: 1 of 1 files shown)


mkosi/sysupdate.py
```diff
@@ -24,29 +26,50 @@ def run_sysupdate(args: Args, config: Config) -> None:
...
+                relaxed=True,
+                setup=["run0"] if os.getuid() != 0 else [],
+                options=[
+                    *(["--bind", "/boot", "/boot"] if Path("/boot").exists() else []),
+                    *(["--bind", "/efi", "/efi"] if Path("/efi").exists() else []),
+                    *(
+                        [
+                            # Make sure systemd-sysupdate parses os-release from the host and not the tools
+                            # tree.
+                            "--bind", "/usr/lib/os-release", "/usr/lib/os-release",
```
behrmann, line 66: Why not also bind the one from `/etc/` here?

behrmann (approved):

mkosi/sysupdate.py
DaanDeMeyer: Because we already use /etc from the host in a relaxed sandbox

PR #3220, opened 2024-11-22
Include machine name in virtiofsd unit name

DaanDeMeyer:
If the same directory is shared across multiple guests, we end up with duplicate unit names even though we need separate virtiofsd instances. Let's fix this by including the machine name in the virtiofsd unit name so that the unit names will be unique across guests.

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -328,9 +328,6 @@ def start_virtiofsd(
     name: Optional[str] = None,
     selinux: bool = False,
 ) -> Iterator[Path]:
-    if name is None:
-        name = systemd_escape(config, directory, path=True)
-
     virtiofsd = find_virtiofsd(root=config.tools(), extra=config.extra_search_paths)
     if virtiofsd is None:
         die("virtiofsd must be installed to boot directory images or use RuntimeTrees= with mkosi qemu")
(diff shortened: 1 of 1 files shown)


mkosi/qemu.py
```diff
@@ -384,8 +381,11 @@ def start_virtiofsd(
         # as root without the source directories necessarily being owned by root. We achieve this by running
         # virtiofsd as the owner of the source directory and then mapping that uid to root.
 
-        name = f"mkosi-virtiofsd-{name}"
-        description = f"virtiofsd for {directory}"
+        if name is None:
+            name = systemd_escape(config, directory, path=True)
+
+        name = f"mkosi-virtiofsd-{config.machine_or_name()}-{name}"
```
behrmann, line 387: Looking at this again, this seems a bit weird? We have five call sites for `start_virtiofsd`, four of them don't set `name`, which means it gets interpolated as `None`, and the one that does, sets it to `config.machine_or_name()`, which would mean that it gets here twice, once escaped and once unescaped.

mkosi/qemu.py
```diff
@@ -384,8 +381,11 @@ def start_virtiofsd(
         # as root without the source directories necessarily being owned by root. We achieve this by running
         # virtiofsd as the owner of the source directory and then mapping that uid to root.
 
+        if name is None:
+            name = f"{config.machine_or_name()}-{systemd_escape(config, directory, path=True)}"
```
behrmann, line 385: We should probably also escape the name to be on the safe side and falling back to the default name for an empty name would probably also make sense
```suggestion
        if not name:
            name = f"{config.machine_or_name()}-{systemd_escape(config, directory, path=True)}"
        else:
            name = systemd_escape(config, name)
```

behrmann (approved):

PR #3222, opened 2024-11-22
Clarify shell vs boot in man page

bjackman:
In retrospect it seems like this should have been obvious but it confused me. Thanks to Daan for pointing it out on Matrix. Reword to help out people like me in the future.

### DIFF
--- a/mkosi/resources/man/mkosi.1.md
+++ b/mkosi/resources/man/mkosi.1.md
@@ -76,17 +76,18 @@ The following command line verbs are known:
 
 `shell`
 :   This builds the image if it is not built yet, and then invokes
-    `systemd-nspawn` to acquire an interactive shell prompt in it. An optional
-    command line may be specified after the `shell` verb, to be invoked in place
-    of the shell in the container. Use `-f` in order to rebuild the image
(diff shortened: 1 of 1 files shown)


mkosi/resources/man/mkosi.1.md
```diff
@@ -76,17 +76,17 @@ The following command line verbs are known:
 
 `shell`
 :   This builds the image if it is not built yet, and then invokes
-    `systemd-nspawn` to acquire an interactive shell prompt in it. An optional
-    command line may be specified after the `shell` verb, to be invoked in place
-    of the shell in the container. Use `-f` in order to rebuild the image
-    unconditionally before acquiring the shell, see below. This command must be
-    executed as `root`.
+    `systemd-nspawn` run an interactive shell in the image, without fully
```
behrmann, line 79: A "to" or similar is missing here
```suggestion
    `systemd-nspawn` to run an interactive shell in the image, without fully
```
I'd also scratch the "fully" at the end, since there's not really a half-booted system. It's basically a better chroot, maybe noting that would be helpful to people?

bjackman: Thanks, sounds good to me.

PR #3227, opened 2024-11-23
Use tools tree again for ssh, coredumpctl and journalctl

DaanDeMeyer:
In effect these verbs are very similar to the sandbox verb except
they run a predefined command line instead of an arbitrary one. So
let's make them behave the same as the sandbox verb when it comes to
the tools tree.


mkosi/__init__.py
```diff
@@ -4730,9 +4703,14 @@ def run_verb(args: Args, images: Sequence[Config], *, resources: Path) -> None:
                 sync_repository_metadata(args, [tools], resources=resources, dst=Path(metadata_dir))
                 fork_and_wait(run_build, args, tools, resources=resources, metadata_dir=Path(metadata_dir))
 
-    if args.verb == Verb.sandbox:
+    if not args.verb.needs_build():
```
behrmann, line 4706: This is a bit too roundabout. There's other verbs that don't need a built, but which are not in the dict below, so if not handleded properly before, this could easily diverge. Maybe a new method `is_sandboxlike`?

DaanDeMeyer: I don't think this will be a big problem. In fact I think it'll be harder to follow if we introduce another method for this.

PR #3232, opened 2024-11-26
doc: make ToolsTreeDistribution default match easier to understand

behrmann:
/cc @septatrix


behrmann:
Excellent point, this has been wrong for a while.

septatrix (approved):
I think replacing "Default" in the heading with "Fallback" or better yet "When unset" might improve this further

PR #3235, opened 2024-11-27
Revert "Remove tools argument from finaiize_default_initrd()"

DaanDeMeyer:
This reverts commit 75a22e52602f18f887a0bac82eaf84e5983941c1.


DaanDeMeyer:
Turns out the assumption was wrong, if we pass the tools tree it also has to exist which isn't always the case when cleaning up

PR #3236, opened 2024-11-27
Add APK and postmarketOS support

septatrix:
This PR adds support for APK and postmarketOS/Alpine (postmarketOS works like a layer atop of Alpine). It is not yet ready to be merged as it contains a few hacks to get stuff working properly. Regardless, this already works and can be used as a common playground to avoid multiple people working on this independently and duplicating the work.

Some stuff needs to be fixed in mkosi, some in postmarketOS, and some in Alpine. And some stuff needs discussion. Current TODOs/problems:

- Populate all the vm/initrd/tools presets
- mkosi assumes that vmlinuz files have their version as a suffic. This is not true for Alpine. `file` is able to extract the version though - maybe this is something which should be supported by `bootctl`: https://github.com/systemd/systemd/issues/35370
- check support for other arches
- add support for mirrors,  local repo, etc.
- check that APK cache properly works
- set up keyring for apk (but `--keys-dir` is always relative to root destination...)
- mkosi always assumes that arches which support SMBIOS also have kernels which enable `CONFIG_DMI_SYSFS`. Alpine does not
- usr-merge is currently hacked in, should be dropped once this is merged upstream

Pinging @calebccff as this is based on his prototype

### DIFF
--- a/mkosi.conf.d/20-alpine.conf
+++ b/mkosi.conf.d/20-alpine.conf
@@ -8,9 +8,12 @@ Distribution=postmarketos
 RepositoryKeyCheck=no
 
 [Content]
+InitrdPackages=
+    systemd-networkd # required for systemd-network-generator
 Packages=
     bash
     systemd-boot
+    systemd-networkd # required for systemd-network-generator
     linux-virt
     # stubbyboot-efistub
     # gummiboot-efistub
--- a/mkosi.conf.d/20-alpine.conf
+++ b/mkosi.conf.d/20-alpine.conf
@@ -11,8 +11,9 @@ RepositoryKeyCheck=no
 Packages=
     bash
     systemd-boot
+    linux-virt
     # stubbyboot-efistub
     # gummiboot-efistub
 
-Bootable=no
+Bootable=yes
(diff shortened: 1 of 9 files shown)


calebccff:
thanks for picking this up and working on it!

> usr-merge is currently hacked in, should be dropped once this is merged upstream

worth noting that while this is fine for a basic poc, it's not very practical yet (e.g. try installing `tar` in that rootfs and you'll [run into trouble](https://pkgs.alpinelinux.org/contents?name=tar&repo=main&branch=edge&arch=x86_64)) -- this is being worked on upstream though and should be resolved after the next stable releases comes out next month.

Alpine tracking issue is https://gitlab.alpinelinux.org/alpine/aports/-/issues/16514

septatrix:
It also seems like Alpine does not use kernel-install, thus no `KERNEL_INSTALL_INITRD_GENERATOR` is set and mkinitfs is unconditionally invoked

DaanDeMeyer:
I'm going to close this PR for now, I don't see the point in having it open until Alpine lands usrmerge and it clutters the PR page.

PR #3255, opened 2024-12-04
qemu: run ssh client as root in sandbox

hundeboll:
Not all users are configured in /etc/passwd (e.g. LDAP users from sssd), and so the ssh client fails when run as non-root user in the sandbox. Simply work around the issue by running as fake-root in the sandbox.

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -1471,6 +1471,6 @@ def run_ssh(args: Args, config: Config) -> None:
             network=True,
             devices=True,
             relaxed=True,
-            options=["--same-dir"],
+            options=["--same-dir", "--become-root"],
         ),
     )


DaanDeMeyer:
Can you explain how ssh fails in the sandbox? Can we get some logs?

DaanDeMeyer:
> > Can you explain how ssh fails in the sandbox? Can we get some logs?
> 
> I've changed the ssh log-level from `ERROR` to `DEBUG`, but I think ssh exits on the failing user lookup before any debug is printed, so not much to show:
> ```
> % ./mkosi/bin/mkosi ssh
> No user exists for uid 511185
> ```
> 
> I'm on a corporate system with `sssd` integration for the company active directory. Turns out I can make user lookup work in the sandbox by installing `sssd-client` in the toolstree, so I guess we can close this PR, if you don't like running `ssh` as (fake) root?

If the package doesn't pull in the kitchen sink we can install it by default. Can you do a pr for that?

hundeboll:
> If the package doesn't pull in the kitchen sink we can install it by default. Can you do a pr for that?

On arch, the following are pulled in:
```
avahi
bind
c-ares
cifs-utils
cyrus-sasl-gssapi
ding-libs
dnssec-anchors
jemalloc
jose
ldb
libbsd
libcups
libdaemon
libmaxminddb
libuv
libwbclient
nfsidmap
smbclient
sssd
talloc
tdb
tevent
```

hundeboll:
Fedora is less intrusive:
```
libsss_idmap
libsss_nss_idmap
sssd-client
```

DaanDeMeyer:
I guess the other thing we could consider is always using the NSS modules from the host instead of the tools tree. I don't know if those ever do any dynamic linking against anything except libc but if they don't it might be viable to use the ones from the host.

DaanDeMeyer:
Nvm not an option. We can just add it to the tools tree

DaanDeMeyer:
Closing this PR as its not the way to go

PR #3258, opened 2024-12-06
fedora: Bump default release to 41

DaanDeMeyer:


behrmann (approved):
LGTM, but it looks like it needs it initrd size bumped (failed in the Arch Fedora combination)

DaanDeMeyer:
Also breaks building suse but that's a dependency cycle bug in opensuse so merging

PR #3263, opened 2024-12-07
Add mkosi-addon and kernel-install plugin

bluca:
Add new mkosi-addon and kernel-install plugin to build local customizations into an EFI addon.

This allows us to move closer to the desired goal of having universal UKIs, built by vendors, used together with locally built enhancements.


bluca:
Tested on my laptop, seems to work nicely. To install the addon requires a change to 90-uki-copy.install that I'll send once this is sorted out.

DaanDeMeyer:
I think this belongs in systemd, not in mkosi. Users should be able to use their vendor provided UKI with an addon without needing mkosi installed on their system.

The UKI is supposed to come with a set of kernel modules to satisfy 99% common use cases, so shipping the entirety of host loaded kernel modules as an addon doesn't make sense to me.

Also we created mkosi-initrd specifically to generate hostonly initrds.   That's it's only reason to exist, so adding a --host-only switch to it seems counter intuitive. I know you're trying to use mkosi-initrd and mkosi-addon together, but that means you're using a locally built initrd and then pretending it's not locally built. To me this is just overcomplicating the setup for the sake of pretending we're doing vendor built UKIs because distributions don't actually have vendor built UKIs yet.

So I agree that an addon generating ker el-install plugin makes sense, but it should live in systemd next to the ukify plugin, not in mkosi, and we shouldn't try to make mkosi-initrd generate vendor built initrds when it's specifically built to generate local initrds.

DaanDeMeyer:
cc @keszybz

bluca:
It needs the logic to copy the kernel modules and firmware from the host, which is in the mkosi library code, so while the script could be moved, it would still depend on mkosi anyway, so why not keep it here so both plugins can be kept in sync?

DaanDeMeyer:
I'm arguing it doesn't need that logic. The vendor uki should include a default set of modules and firmware that work for most use cases. And then in super special cases addons are used to include a few extras.

So it can live in systemd without having to depend on mkosi.

bluca:
That's unlikely to work though, you'd need a UKI with hundreds of MBs of stuff - the NVIDIA firmware alone is like 200M. Plus there's a gazillion different bluetooth/wifi firmwares, each in the tens of MBs each. There will always be demand for a minimal generic UKI + local-specific modifications. Otherwise why have that logic in mkosi-initrd either? Just build UKIs with everything, no?

DaanDeMeyer:
> That's unlikely to work though, you'd need a UKI with hundreds of MBs of stuff - the NVIDIA firmware alone is like 200M. Plus there's a gazillion different bluetooth/wifi firmwares, each in the tens of MBs each. There will always be demand for a minimal generic UKI + local-specific modifications. Otherwise why have that logic in mkosi-initrd either? Just build UKIs with everything, no?

We have that logic in mkosi-initrd precisely because the initrd is built locally. We're building locally anyway so we might as well make use of the advantages that brings us.

With your approach, you're using a vendor signed UKI and then requiring every user to enroll their own key since they need to add their own addon with kernel modules and firmware. So the entire point of the vendor signed UKI becomes moot because every single user still has to enroll their own key.

The entire goal of the vendor signed UKIs is that they work out of the box for most use cases or the vendor also provides signed addons that extend the UKI. As soon as you require every single user to enroll their own key, you might as well build and sign the entire UKI locally and not bother with vendor signing to begin with.

So a solution like this that implies every user enrolling their own key is not an option in my opinion.

DaanDeMeyer:
And if you want this just for the niche use case where a user wants a locally signed addon with kernel modules with a minimal signed vendor UKI, that's too niche to me to justify all this extra code in mkosi for that specific use case. I would much rather we focus on building the entire UKI ourselves and signing that locally then instead of focusing on building addons for vendor signed UKIs.

DaanDeMeyer:
What is attractive to add to mkosi is a way to build those signed vendor addons. So I am actually interested in the addon output format as a similar mechanism to sysexts you can use it together with Overlay=yes to package an overlay on top of the base initrd mage into an addon.

bluca:
Having the possibility of building local customizations will always be needed, it cannot possibly be preditected what all combinations one might need will be, especially including local configuration (eg: crypttab) and kernel command line. Yes that will require local keys, but that's always been the case, eg: out of tree kernel drivers or filesystems

DaanDeMeyer:
> Having the possibility of building local customizations will always be needed, it cannot possibly be preditected what all combinations one might need will be, especially including local configuration (eg: crypttab) and kernel command line. Yes that will require local keys, but that's always been the case, eg: out of tree kernel drivers or filesystems

Yes it will be needed, it's just not something I want to maintain the code for in mkosi. If the kernel modules logic is dropped I think it's perfectly fine to propose this for inclusion in systemd.

I think mkosi should focus on either building vendor signed UKIs + vendor signed addons or locally signed UKIs and not everything inbetween.

aafeijoo-suse:
> Also we created mkosi-initrd specifically to generate hostonly initrds. That's it's only reason to exist, so adding a --host-only switch to it seems counter intuitive.

So, the idea is having `mkosi-initrd` only for host specific initrds and use `mkosi --format=cpio --include=mkosi-initrd ...` for generic initrds?  IMHO, having a `--host-only` switch to generate initrds with and without local configuration on them is a must-have feature if `mkosi-initrd` is intended to be a complete initrd generator.

mkosi/addon.py
```diff
@@ -0,0 +1,212 @@
...
+            "--build-sources", "",
+            "--include=mkosi-addon",
+        ]  # fmt: skip
+
+        cmdline += [
+            "--extra-tree", f"/usr/lib/modules/{args.kernel_version}:/usr/lib/modules/{args.kernel_version}",
+            "--extra-tree=/usr/lib/firmware:/usr/lib/firmware",
+            "--remove-files=/usr/lib/firmware/*-ucode",
+            "--kernel-modules-exclude=.*",
+            "--kernel-modules-include=host",
```
DaanDeMeyer, line 107: I don't think we should do this by default, let's leave it for users to enable in their mkosi-addon configuration.

mkosi/resources/mkosi-addon/mkosi.conf
```diff
@@ -0,0 +1,22 @@
...
+
+[Output]
+Output=initrd.addon
+Format=addon
+ManifestFormat=
+SplitArtifacts=
+
+[Content]
+Bootable=no
+MakeInitrd=yes
```
DaanDeMeyer, line 14: I'm confident we don't need this, this creates the /init and /etc/initrd-release symlinks which shouldn't be in the addon.

mkosi/addon.py
```diff
@@ -0,0 +1,212 @@
...
+                shutil.copy2(Path("/etc") / p, Path(sandbox_tree) / "etc" / p)
+            else:
+                shutil.copytree(
+                    Path("/etc") / p,
+                    Path(sandbox_tree) / "etc" / p,
+                    ignore=shutil.ignore_patterns("gnupg"),
+                    dirs_exist_ok=True,
+                )
+
+        cmdline += ["--sandbox-tree", sandbox_tree]
```
DaanDeMeyer, line 165: No point in all this if the distribution is `custom`, you can drop it

kernel-install/51-mkosi-addon.install
```diff
@@ -0,0 +1,111 @@
...
+
+def mandatory_variable(name: str) -> str:
+    try:
+        return os.environ[name]
+    except KeyError:
+        die(f"${name} must be set in the environment")
+
+
+@uncaught_exception_handler()
+def main() -> None:
```
DaanDeMeyer, line 46: This should be opt-in somehow, we shouldn't generate an addon by default.

DaanDeMeyer: Always building an addon just if we're building an UKI shouldn't happen IMO.

mkosi/addon.py
```diff
@@ -0,0 +1,212 @@
...
+            ]
+
+        for d in (
+            "/usr/lib/mkosi-addon",
+            "/usr/local/lib/mkosi-addon",
+            "/run/mkosi-addon",
+            "/etc/mkosi-addon",
+        ):
+            if Path(d).exists():
+                cmdline += ["--include", d]
```
DaanDeMeyer, line 130: We could for example skip generating the addon if no local configuration is defined.so that users opt-in by writing a configuration file.

mkosi/addon.py
```diff
@@ -0,0 +1,212 @@
...
+                    and not entry[0].startswith("#")
+                    and "x-initrd.attach" in entry[3]
+                )
+            ]
+            print(crypttab)
+            if crypttab:
+                with (Path(staging_dir) / "crypttab").open("w") as f:
+                    f.write("# Automatically generated by mkosi-addon\n")
+                    f.write("\n".join(crypttab))
+                cmdline += ["--extra-tree", f"{staging_dir}/crypttab:/etc/crypttab"]
```
DaanDeMeyer, line 183: Please share this logic with `mkosi-initrd`

mkosi/addon.py
```diff
@@ -0,0 +1,212 @@
...
+        if Path("/etc/kernel/cmdline").exists():
+            cmdline += ["--kernel-command-line", Path("/etc/kernel/cmdline").read_text()]
+
+        # Prefer dnf as dnf5 has not yet officially replaced it and there's a much bigger chance that there
+        # will be a populated dnf cache directory.
+        run(
+            cmdline,
+            stdin=sys.stdin,
+            stdout=sys.stdout,
+            env={"MKOSI_DNF": dnf.name} if (dnf := find_binary("dnf", "dnf5")) else {},
```
DaanDeMeyer, line 194: This can be dropped

mkosi/addon.py
```diff
@@ -0,0 +1,212 @@
...
+            if crypttab:
+                with (Path(staging_dir) / "crypttab").open("w") as f:
+                    f.write("# Automatically generated by mkosi-addon\n")
+                    f.write("\n".join(crypttab))
+                cmdline += ["--extra-tree", f"{staging_dir}/crypttab:/etc/crypttab"]
+
+        if Path("/etc/kernel/cmdline").exists():
+            cmdline += ["--kernel-command-line", Path("/etc/kernel/cmdline").read_text()]
+
+        # Prefer dnf as dnf5 has not yet officially replaced it and there's a much bigger chance that there
```
DaanDeMeyer, line 188: This comment can be dropped

mkosi/addon.py
```diff
@@ -0,0 +1,212 @@
...
+                Path(args.output_dir).mkdir(parents=True, exist_ok=True)
+        else:
+            args.output_dir = Path.cwd()
+
+        log_notice(f"Copying {staging_dir}/{args.output} to {args.output_dir}/{args.output}")
+        # mkosi symlinks the expected output image, so dereference it
+        copy_tree(
+            Path(f"{staging_dir}/{args.output}").resolve(),
+            Path(f"{args.output_dir}/{args.output}"),
+        )
```
DaanDeMeyer, line 208: I think we can share this with `mkosi-initrd` as well

mkosi/resources/mkosi-addon/mkosi.conf
```diff
@@ -0,0 +1,22 @@
...
+CleanPackageMetadata=yes
```
DaanDeMeyer, line 15: This can be dropped

DaanDeMeyer (changes requested):
So we have the initrd-addon output format now so that logic can be removed from this PR.

After taking another look I see the value in having this, but with some changes, see other comments.

mkosi/addon.py
bluca: seems a bit overkill for one function

kernel-install/51-mkosi-addon.install
```diff
@@ -0,0 +1,111 @@
...
+        version=f"mkosi {__version__}",
+    )
+
+    context = Context(
+        **vars(parser.parse_args()),
+        staging_area=Path(mandatory_variable("KERNEL_INSTALL_STAGING_AREA")),
+        layout=mandatory_variable("KERNEL_INSTALL_LAYOUT"),
+        image_type=mandatory_variable("KERNEL_INSTALL_IMAGE_TYPE"),
+        verbose=int(os.getenv("KERNEL_INSTALL_VERBOSE", 0)) > 0,
+    )
```
DaanDeMeyer, line 91: Pretty sure all of this can be shared with mkosi-initrd as well

kernel-install/51-mkosi-addon.install
```diff
@@ -0,0 +1,111 @@
...
+def we_are_wanted(context: Context) -> bool:
+    return context.layout == "uki"
+
+
+def mandatory_variable(name: str) -> str:
+    try:
+        return os.environ[name]
+    except KeyError:
+        die(f"${name} must be set in the environment")
+
```
DaanDeMeyer, line 43: Can move this to util.py and import it

mkosi/__init__.py
```diff
@@ -283,7 +283,10 @@ def remove_packages(context: Context) -> None:
 
 
 def check_root_populated(context: Context) -> None:
-    if context.config.output_format.is_extension_image():
+    if (
+        context.config.output_format.is_extension_image()
+        or context.config.output_format == OutputFormat.addon
+    ):
```
DaanDeMeyer, line 289: Can be dropped, `is_extension_image()` handles `addon`

mkosi/__init__.py
```diff
@@ -303,7 +306,11 @@ def configure_os_release(context: Context) -> None:
     if not (context.config.image_id or context.config.image_version or context.config.hostname):
         return
 
-    if context.config.overlay or context.config.output_format.is_extension_image():
+    if (
+        context.config.overlay
+        or context.config.output_format.is_extension_image()
+        or context.config.output_format == OutputFormat.addon
+    ):
```
DaanDeMeyer, line 313: Ditto

mkosi/__init__.py
```diff
@@ -2084,7 +2091,7 @@ def install_kernel(context: Context, partitions: Sequence[Partition]) -> None:
     # single-file images have the benefit that they can be signed like normal EFI binaries, and can
     # encode everything necessary to boot a specific root device, including the root hash.
 
-    if context.config.output_format in (OutputFormat.uki, OutputFormat.esp):
+    if context.config.output_format in (OutputFormat.uki, OutputFormat.esp, OutputFormat.addon):
```
DaanDeMeyer, line 2094: Ditto

mkosi/__init__.py
DaanDeMeyer: And everywhere below

mkosi/util.py
```diff
@@ -257,3 +257,29 @@ def current_home_dir() -> Optional[Path]:
 
 def unique(seq: Sequence[T]) -> list[T]:
     return list(dict.fromkeys(seq))
+
+
+def initrd_copy_host_config(staging_dir: str) -> list[str]:
```
DaanDeMeyer, line 262: This is too generic for util.py, you can just keep it in initrd.py and import it from addon.py

mkosi/addon.py
DaanDeMeyer: I mean with the output dir shenanigans above as well

mkosi/addon.py
```diff
@@ -0,0 +1,151 @@
...
+            "--include=mkosi-addon",
+        ]  # fmt: skip
+
+        cmdline += [
+            "--extra-tree=/usr/lib/systemd/boot/efi:/usr/lib/systemd/boot/efi",
+            "--remove-files=/usr/lib/systemd/boot/efi/",
+            "--sandbox-tree",
+            f"/usr/lib/modules/{args.kernel_version}:/usr/lib/modules/{args.kernel_version}",
+            "--sandbox-tree=/usr/lib/firmware:/usr/lib/firmware",
+        ]  # fmt: skip
```
DaanDeMeyer, line 108: Just combine these, and move the systemd-stub stuff into the config file

DaanDeMeyer: Also the sandbox trees for modules and firmware should be extra trees

DaanDeMeyer (changes requested):

mkosi/__init__.py
```diff
@@ -2170,11 +2177,25 @@ def make_initrd_addon(context: Context, stub: Path, output: Path) -> None:
...
+            context.config,
+            version="256~devel",
+            reason="build addon with .ucode section support",
+            hint=("Use ToolsTree=default to download most required tools including ukify automatically"),
+        )
+
+        for microcode in microcodes:
+            arguments += ["--microcode", workdir(microcode)]
+            options += ["--ro-bind", microcode, workdir(microcode)]
+
```
DaanDeMeyer, line 2192: As mentioned in chat, I'd drop this until a proper use case appears, a vendor UKI without microcode is going to be rather broken.

mkosi/__init__.py
bluca: Why? It's a different format from uki/esp

mkosi/addon.py
bluca: extra-tree means the entire thing gets included - all the firmware and all the kernel modules

DaanDeMeyer: Not if you add `KernelModulesExclude=.*`

DaanDeMeyer: And then other stuff can be selectively included

mkosi/__init__.py
DaanDeMeyer: It's handled by the below block:

```
    if context.config.bootable == ConfigFeature.auto and (
        context.config.output_format == OutputFormat.cpio
        or context.config.output_format.is_extension_or_portable_image()
        or context.config.overlay
    ):
        return
```

(And you set `Bootable=no` anyway so it'll be skipped even earlier.

mkosi/addon.py
bluca: `KernelModulesExclude=.*` does not work, everything is included anyway

DaanDeMeyer: https://github.com/systemd/mkosi/pull/3381

bluca: Still doesn't work with that either

mkosi/config.py
```diff
@@ -843,7 +843,13 @@ def config_parse_mode(value: Optional[str], old: Optional[int]) -> Optional[int]
...
 
 def config_default_compression(namespace: argparse.Namespace) -> Compression:
-    if namespace.output_format in (OutputFormat.tar, OutputFormat.cpio, OutputFormat.uki, OutputFormat.esp):
+    if namespace.output_format in (
+        OutputFormat.tar,
+        OutputFormat.cpio,
+        OutputFormat.uki,
+        OutputFormat.esp,
+        OutputFormat.addon,
+    ):
```
DaanDeMeyer, line 852: I've split this out into https://github.com/systemd/mkosi/pull/3384

mkosi/initrd.py
```diff
@@ -21,43 +21,54 @@
...
-@uncaught_exception_handler()
-def main() -> None:
-    log_setup()
-
-    parser = argparse.ArgumentParser(
-        prog="mkosi-initrd",
-        description="Build initrds or unified kernel images for the current system using mkosi",
-        allow_abbrev=False,
-        usage="mkosi-initrd [options...]",
+def initrd_copy_host_config(staging_dir: str) -> list[str]:
```
DaanDeMeyer, line 24: Let's handle crypttab and kernel/cmdline separately. (Don't necessarily need a separate function for kernel/cmdline.

So I'd rename this to `process_crypttab()` or so

DaanDeMeyer (changes requested):

mkosi/addon.py
```diff
@@ -0,0 +1,95 @@
...
+            "--force",
+            "--directory", "",
+            f"--format={str(OutputFormat.addon)}",
+            "--output", args.output,
+            "--output-directory", staging_dir,
+            "--build-sources", "",
+            "--include=mkosi-addon",
+            "--sandbox-tree",
+            f"/usr/lib/modules/{args.kernel_version}:/usr/lib/modules/{args.kernel_version}",
+            "--sandbox-tree=/usr/lib/firmware:/usr/lib/firmware",
```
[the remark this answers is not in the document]
bluca: Still doesn't work, all modules are added

kernel-install/51-mkosi-addon.install
bluca: but they are different kernel-install scripts?

DaanDeMeyer: Yes but each kernel-install plugin receives the same arguments, so we can share the argument parser and the context object between them.

mkosi/addon.py
```diff
@@ -0,0 +1,97 @@
...
+
+        if args.debug:
+            cmdline += ["--debug"]
+        if args.debug_shell:
+            cmdline += ["--debug-shell"]
+
+        if os.getuid() == 0:
+            cmdline += [
+                "--workspace-dir=/var/tmp",
+                "--package-cache-dir=/var",
```
DaanDeMeyer, line 73: This one can be removed to, (look at which lines the comment is left on)

README.md
```diff
@@ -81,15 +81,16 @@ mkosi binary and is not to be considered a public API.
 
 ## kernel-install plugin
```
DaanDeMeyer, line 82: plugin => plugins

DaanDeMeyer: Actually I would document these separately and as a "one or the other" type of setup. We shouldn't recommend users use both. Either they use mkosi-initrd to build their own UKI with local modifications or they use a vendor provided UKI and mkosi-addon for local modifications. But it should be one or the other, not both at once.

mkosi/addon.py
```diff
@@ -0,0 +1,97 @@
...
+    args = parser.parse_args()
+
+    if args.show_documentation:
+        with resource_path(mkosi.resources) as r:
+            show_docs("mkosi-addon", DocFormat.all(), resources=r)
+        return
+
+    # No local configuration? Then nothing to do
+    if not Path("/etc/mkosi-addon").exists() and not Path("/run/mkosi-addon").exists():
+        return
```
DaanDeMeyer, line 48: Let's move this earlier, I wouldn't even try to parse arguments if no configuration exists

kernel-install/51-mkosi-addon.install
```diff
@@ -0,0 +1,105 @@
...
+        nargs="?",
+        help="Type#1 entry directory (ignored)",
+    )
+    parser.add_argument(
+        "kernel_image",
+        metavar="KERNEL_IMAGE",
+        type=Path,
+        nargs="?",
+        help="Kernel image",
+    )
```
DaanDeMeyer, line 72: We should check that a kernel image is provided and that it is a UKI (we have `KernelType.identify()` for that) and skip the plugin if a UKI was not provided on the kernel command line.

If kernel-install is building its own UKI this plugin should be skipped.

DaanDeMeyer (changes requested):
I'll try it out locally to see what's up with the kernel modules

mkosi/addon.py
bluca: showing --help or equivalent is useful even with no configuration, so that you can find out what configuration to add

mkosi/addon.py
```diff
@@ -0,0 +1,96 @@
...
+
+    with tempfile.TemporaryDirectory() as staging_dir:
+        cmdline: list[PathString] = [
+            "mkosi",
+            "--force",
+            "--directory", "",
+            f"--format={str(OutputFormat.addon)}",
```
DaanDeMeyer, line 55: This can be dropped

mkosi/util.py
```diff
@@ -257,3 +258,42 @@ def current_home_dir() -> Optional[Path]:
...
+
+
+def mandatory_variable(name: str) -> str:
+    try:
+        return os.environ[name]
+    except KeyError:
+        die(f"${name} must be set in the environment")
+
+
+def kernel_install_arg_parse(parser: argparse.ArgumentParser) -> None:
```
DaanDeMeyer, line 270: Same thing here, just keep it in initrd.py, makes more sense there for sharing than here

mkosi/addon.py
```diff
@@ -0,0 +1,96 @@
...
+
+    args = parser.parse_args()
+
+    if args.show_documentation:
+        with resource_path(mkosi.resources) as r:
+            show_docs("mkosi-addon", DocFormat.all(), resources=r)
+        return
+
+    # No local configuration? Then nothing to do
+    if not Path("/etc/mkosi-addon").exists() and not Path("/run/mkosi-addon").exists():
```
DaanDeMeyer, line 47: Actually let's move this to the kernel-install plugin, makes more sense to do the skip there rather than here, having the skip here makes local development and testing harder than necessary

kernel-install/51-mkosi-addon.install
```diff
@@ -0,0 +1,84 @@
...
+@dataclasses.dataclass(frozen=True)
+class Context:
+    command: str
+    kernel_version: str
+    entry_dir: Path
+    kernel_image: Path
+    staging_area: Path
+    layout: str
+    image_type: str
+    verbose: bool
```
DaanDeMeyer, line 33: This class should be shared as well

DaanDeMeyer: And add `kernel_install_arg_parse()` as a static `parse()` method to this class maybe

kernel-install/51-mkosi-addon.install
```diff
@@ -0,0 +1,84 @@
...
+    )
+
+    if context.command != "add" or not context.layout == "uki":
+        return
+
+    if not context.kernel_image or not context.kernel_image.exists():
+        die(f"Kernel image {context.kernel_image} does not exist")
+
+    if KernelType.identify(context.config, context.kernel_image) != KernelType.uki:
+        die(f"Kernel image {context.kernel_image} is not a Unified Kernel Image")
```
DaanDeMeyer, line 67: We should skip in this case, not fail loudly

mkosi/addon.py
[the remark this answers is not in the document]
DaanDeMeyer: With https://github.com/systemd/mkosi/pull/3387 and these local modifications, it works:

```
+            "--extra-tree", f"/usr/lib/modules/{args.kernel_version}:/usr/lib/modules/{args.kernel_version}",
+            "--extra-tree=/usr/lib/firmware:/usr/lib/firmware",
+            "--kernel-modules-exclude=.*",

```

mkosi/resources/mkosi-addon/mkosi.conf
```diff
@@ -0,0 +1,22 @@
...
+Bootable=no
+# Needs to be available for the addon stub, but don't want it in the initrd
+ExtraTrees=/usr/lib/systemd/boot/efi:/usr/lib/systemd/boot/efi
+RemoveFiles=/usr/lib/systemd/boot/efi/
+RemoveFiles=
+        # Including kernel images in the initrd is generally not useful.
+        # This also stops mkosi from extracting the kernel image out of the image as a separate output.
+        /usr/lib/modules/*/vmlinuz*
+        /usr/lib/modules/*/vmlinux*
+        /usr/lib/modules/*/System.map
```
DaanDeMeyer, line 22: You will want to add the following here:

```
+        # This is an addon so drop all modules files as these would override the ones from the base image.
+        /usr/lib/modules/*/modules.*
+        # Arch Linux specific file.
+        /usr/lib/modules/*/pkgbase
+        # Drop microcode directories explicitly as these are not dropped by the kernel modules processing
+        # logic.
+        /usr/lib/firmware/intel-ucode
+        /usr/lib/firmware/amd-ucode
```

DaanDeMeyer (changes requested):

DaanDeMeyer:
One more change that we should make, since we now only run the kernel-install plugin when passed a vendor provided UKI, we can extract the initrd from the vendor provided UKI and use it as a base tree.

That would make this a lot more powerful, since if the vendor provided UKI would include a package database, it'd become possible to install extra packages in the addon. It also makes sure we only add kernel modules to the addon that aren't already in the vendor UKI initrd after #3390 is merged.

We have a function `extract_pe_section()` to extract sections from a UKI. You'll also have to add `extract_cpio()` to `archive.py` to implement the CPIO extraction.

DaanDeMeyer:
> One more change that we should make, since we now only run the kernel-install plugin when passed a vendor provided UKI, we can extract the initrd from the vendor provided UKI and use it as a base tree.
> 
> That would make this a lot more powerful, since if the vendor provided UKI would include a package database, it'd become possible to install extra packages in the addon. It also makes sure we only add kernel modules to the addon that aren't already in the vendor UKI initrd after #3390 is merged.
> 
> We have a function `extract_pe_section()` to extract sections from a UKI. You'll also have to add `extract_cpio()` to `archive.py` to implement the CPIO extraction.

Actually nvm this doesn't account for UKI profiles, ignore my blabber

kernel-install/50-mkosi.install
```diff
@@ -80,45 +59,17 @@ def main() -> None:
...
-        nargs="?",
-        help="Kernel image",
-    )
     parser.add_argument(
         "initrds",
         metavar="INITRD…",
         type=Path,
         nargs="*",
         help="Initrd files",
     )
```
DaanDeMeyer, line 68: This can be moved into shared code as well, the command line interface used by plugins is identical across plugins

kernel-install/50-mkosi.install
```diff
@@ -80,45 +59,17 @@ def main() -> None:
...
-    parser.add_argument(
-        "--version",
-        action="version",
-        version=f"mkosi {__version__}",
-    )
 
-    context = Context(
+    KernelInstallContext.kernel_install_arg_parse(parser)
```
DaanDeMeyer, line 70: Please just name this `parse()`

kernel-install/50-mkosi.install
```diff
@@ -80,45 +59,17 @@ def main() -> None:
         allow_abbrev=False,
         usage="50-mkosi.install COMMAND KERNEL_VERSION ENTRY_DIR KERNEL_IMAGE INITRD…",
```
DaanDeMeyer, line 60: Please move the entire argument parser into the `parse()` method as well and pass in the name and description of the plugin as arguments.

kernel-install/50-mkosi.install
```diff
@@ -80,45 +59,17 @@ def main() -> None:
...
-    parser.add_argument(
-        "--version",
-        action="version",
-        version=f"mkosi {__version__}",
-    )
 
-    context = Context(
+    KernelInstallContext.kernel_install_arg_parse(parser)
+
+    context = KernelInstallContext(
```
DaanDeMeyer, line 72: Please return the context from the `parse()` static method

kernel-install/51-mkosi-addon.install
```diff
@@ -0,0 +1,66 @@
...
+
+    KernelInstallContext.kernel_install_arg_parse(parser)
+
+    context = KernelInstallContext(
+        **vars(parser.parse_args()),
+        staging_area=Path(mandatory_variable("KERNEL_INSTALL_STAGING_AREA")),
+        layout=mandatory_variable("KERNEL_INSTALL_LAYOUT"),
+        image_type=mandatory_variable("KERNEL_INSTALL_IMAGE_TYPE"),
+        verbose=int(os.getenv("KERNEL_INSTALL_VERBOSE", 0)) > 0,
+    )
```
DaanDeMeyer, line 40: Same here as above, please consolidate in `parse()` method

kernel-install/51-mkosi-addon.install
```diff
@@ -0,0 +1,66 @@
...
+
+    if context.command != "add" or not context.layout == "uki":
+        return
+
+    if not context.kernel_image or not context.kernel_image.exists():
+        die(f"Kernel image {context.kernel_image} does not exist")
```
DaanDeMeyer, line 46: Please skip in this case as well, let's leave validation of the kernel image argument to `kernel-install`

kernel-install/51-mkosi-addon.install
```diff
@@ -0,0 +1,66 @@
...
+
+    context = KernelInstallContext(
+        **vars(parser.parse_args()),
+        staging_area=Path(mandatory_variable("KERNEL_INSTALL_STAGING_AREA")),
+        layout=mandatory_variable("KERNEL_INSTALL_LAYOUT"),
+        image_type=mandatory_variable("KERNEL_INSTALL_IMAGE_TYPE"),
+        verbose=int(os.getenv("KERNEL_INSTALL_VERBOSE", 0)) > 0,
+    )
+
+    if context.command != "add" or not context.layout == "uki":
```
DaanDeMeyer, line 42: `context.layout != "uki"`

DaanDeMeyer: Let's also log an info message here when we skip and everywhere else

mkosi/resources/man/mkosi-addon.1.md
```diff
@@ -0,0 +1,49 @@
...
+mkosi-addon — Build addons for unified kernel images for the current system
+using mkosi
+
+# SYNOPSIS
+
+`mkosi-addon [options…]`
+
+# DESCRIPTION
+
+`mkosi-addon` is wrapper on top of `mkosi` to simplify the generation of
```
DaanDeMeyer, line 16: is a wrapper

mkosi/resources/man/mkosi-addon.1.md
```diff
@@ -0,0 +1,49 @@
...
+addons containing customizations for a Unified Kernel Images specific for the
```
DaanDeMeyer, line 17: addons => PE addons

DaanDeMeyer, line 17: For unified kernel images

DaanDeMeyer, line 17: specific for => specific to

mkosi/resources/man/mkosi-addon.1.md
```diff
@@ -0,0 +1,49 @@
...
+current running system. Will include entries in `/etc/crypttab` marked with
```
DaanDeMeyer, line 18: current running => running or local

mkosi/resources/man/mkosi-addon.1.md
```diff
@@ -0,0 +1,49 @@
...
+`x-initrd.attach`, `/etc/kernel/cmdline`, kernel modules, firmwares and microcode
```
DaanDeMeyer, line 19: microcode should be dropped

DaanDeMeyer (changes requested):

kernel-install/50-mkosi.install
bluca: this is specific for this one, the other doesn't accept this parameter

kernel-install/50-mkosi.install
bluca: why? it's just going to be more convoluted, as it's different across scripts anyway

kernel-install/50-mkosi.install
bluca: the parameters are different so can't do that

kernel-install/50-mkosi.install
DaanDeMeyer: @bluca It doesn't work like that, the plugins don't define the command line interface, kernel-install does. And kernel-install will pass all the initrds it receives as command line arguments to all plugins. So each plugin has to be able to deal with that.

`mkosi-addon` doesn't have to do anything with the passed initrds, it can just ignore them, but the parser should parse them all the same.

kernel-install/50-mkosi.install
DaanDeMeyer: It's not, it's 100% identical for each kernel-install plugin, only the description and name are different, all the rest is identical because kernel-install determines the interface, not an individual plugin.

kernel-install/50-mkosi.install
DaanDeMeyer: See above

README.md
```diff
@@ -79,17 +79,31 @@ when not installed as a zipapp.
...
+be built and installed.
 
 Once installed, the mkosi plugin can be enabled by writing
-`initrd_generator=mkosi-initrd` to `/usr/lib/kernel/install.conf` or to
-`/etc/kernel/install.conf`.
+`initrd_generator=mkosi-initrd` and `layout=uki` to `/usr/lib/kernel/install.conf`
+or to `/etc/kernel/install.conf`.
+
+## Addon plugin
+To enable this features, install `kernel-install/51-mkosi-addon.install` into
```
DaanDeMeyer, line 99: features => feature

mkosi/resources/man/mkosi-addon.1.md
```diff
@@ -0,0 +1,49 @@
...
+
+`mkosi-addon [options…]`
+
+# DESCRIPTION
+
+`mkosi-addon` is a wrapper on top of `mkosi` to simplify the generation of PE
+addons containing customizations for unified kernel images specific to the
+running or local system. Will include entries in `/etc/crypttab` marked with
+`x-initrd.attach`, `/etc/kernel/cmdline`, kernel modules and firmwares for the
+running hardware.
```
DaanDeMeyer, line 20: The "kernel modules and firmwares for the running hardware" should be dropped as that now only happens if users add `KernelModulesIncludeHost=` to their local configuration.

Maybe document that if users want to add kernel modules in use on their host system, they can add `KernelModulesIncludeHost=` to their local configuration.

DaanDeMeyer (changes requested):

DaanDeMeyer (approved):

PR #3264, opened 2024-12-07
mkosi-initrd: remove dpkg in newer ubuntu/debian releases

bluca:


DaanDeMeyer:
You're inventing Match syntax here. It needs to be one per line, not space delimited list.

bluca:
Being creative is an important part of life

bluca:
This works, sadly install-info is installed instead due to an old dependency in gzip, filed a bug https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1089532 and a patch, but it's much smaller so it's still better than nothing

DaanDeMeyer (approved):

Issue #3265, opened 2024-12-07
Issues with unprivileged userns disabled


bjackman:
### mkosi commit the issue has been seen with

c4bbf3b71a3e2cf947995caedf10f69da3c4957a

### Used host distribution

Ubuntu 24.04.1 LTS

### CPU architectures issue was seen on

None

### Unexpected behaviour you saw

[Recent Ubuntu versions](https://ubuntu.com/blog/ubuntu-23-10-restricted-unprivileged-user-namespaces) have an AppArmor policy to disable creation of unprivileged user namespaces. 

This is extremely annoying but unfortunately it's a pretty sensible policy given the current state of the kernel, I expect other general-purpose distros will do the same if they haven't already. 

For AppArmor specifically you can disable the restriction system-wide with:

```sh
sudo sysctl -w kernel.apparmor_restrict_unprivileged_unconfined=0
sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
```

Or you can create a profile for the binary:

```
❯❯  cat > /etc/apparmor.d/mkosi 
abi <abi/4.0>,
include <tunables/global>

# REPLACE THE PATH to point to where the binary is
profile mkosi /home/brendan/.local/share/pipx/venvs/mkosi/bin/mkosi flags=(unconfined) {
  userns,

  # Site-specific additions and overrides. See local/README for details.
  include if exists <local/mkosi>
}
❯❯  sudo systemctl reload apparmor
```

There is also a kernel patch floating around that does this with a sysctl separate from any LSM framework, in that case you remove the restriction system-wide with:

```
sudo sysctl -w kernel.unprivileged_userns_clone=1
```

(Note this has the opposite polarity of Ubuntu's AppArmor sysctl).

I dunno if other LSMs have something like Ubuntu's AppArmor thing. I assume they will because globally disabling unpriv userns must break a lot of useful stuff.

I wonder if there's any good fix for this? I guess if there are distro packages of mkosi they can just add the LSM profile where that's a possibility. For people installing from source.. I guess it just has to be docs? Maybe mkosi could spit out a suggested LSM config based on your distro and the path of its binary?

bjackman:
For searchability, here's what the error looks like with the AppArmor restriction- `unshare` returns `-EPERM`:

```
❯❯  mkosi -f
‣ Validating certificates and keys
‣ Syncing package manager metadata
Traceback (most recent call last):
[...]

PR #3266, opened 2024-12-08
sandbox: Spit out some info when unsharing userns gets EPERM

bjackman:
To try and minimise the pain of this issue
(https://github.com/systemd/mkosi/issues/3265), dump some info that might help users resolve it.

I had a quick look around expecting to find a document from Red Hat discussing this topic much like the Ubuntu one I've linked here, but I didn't find it. Hopefully if it exists someone else can add it later.

### DIFF
--- a/mkosi/resources/man/mkosi.1.md
+++ b/mkosi/resources/man/mkosi.1.md
@@ -2904,6 +2904,29 @@ images you want to build.
 
 Note that the minimum required Python version is 3.9.
 
+mkosi needs unrestricted abilities to create and act within namespaces. Some
+distros restrict creation of, or capabilities within, user namespaces, which
+breaks mkosi.
+
+For information about Ubuntu, that implements such restrictions using AppArmor, see
+https://ubuntu.com/blog/ubuntu-23-10-restricted-unprivileged-user-namespaces.
+For other systems, try researching the `kernel.unprivileged_userns_clone` or
+`user.max.user_namespace` sysctls.
+
+For Ubuntu systems, you can remove the restrictions for mkosi by
+adapting this snippet to point to your mkosi binary, copying it to
+`/etc/apparmor.d/path.to.mkosi`, and then running `systemctl reload apparmor`:
+
+```
+abi <abi/4.0>,
+
+include <tunables/global>
+
+/path/to/mkosi flags=(default_allow) {
+  userns,
+}
+```
+
 # Frequently Asked Questions (FAQ)
 
 - Why does `mkosi qemu` with KVM not work on Debian/Kali/Ubuntu?
--- a/mkosi/sandbox.py
+++ b/mkosi/sandbox.py
@@ -28,6 +28,7 @@ CLONE_NEWIPC = 0x08000000
 CLONE_NEWNET = 0x40000000
 CLONE_NEWNS = 0x00020000
 CLONE_NEWUSER = 0x10000000
+EPERM = 1
 ENOENT = 2
 LINUX_CAPABILITY_U32S_3 = 2
 LINUX_CAPABILITY_VERSION_3 = 0x20080522
@@ -342,6 +343,10 @@ def become_user(uid: int, gid: int) -> None:
 
     try:
         unshare(CLONE_NEWUSER)
+    except OSError as e:
+        if e.errno == EPERM:
+            print(UNSHARE_EPERM_MSG, file=sys.stderr)
+        raise
     finally:
         os.write(event, ctypes.c_uint64(1))
         os.close(event)
@@ -694,6 +699,16 @@ See the mkosi-sandbox(1) man page for details.\
 """
 
 
+UNSHARE_EPERM_MSG = """
+mkosi was forbidden to unshare namespaces.
+
+This probably means your distribution has restricted unprivileged user namespaces.
+
+Please consult the REQUIREMENTS section of the mkosi man page, e.g. via "mkosi
+documentation", for workarounds.
+"""
+
+
 def main() -> None:
     # We don't use argparse as it takes +- 10ms to import and since this is purely for internal
     # use, it's not necessary to have good UX for this CLI interface so it's trivial to write
@@ -806,7 +821,14 @@ def main() -> None:
     if suppress_chown and (userns or userns_has_single_user()):
         seccomp_suppress_chown()
 
-    unshare(namespaces)
+    try:
+        unshare(namespaces)
+    except OSError as e:
+        # This can happen here as well as in become_user, it depends on exactly
+        # how the userns restrictions are implemented.
+        if e.errno == EPERM:
+            print(UNSHARE_EPERM_MSG, file=sys.stderr)
+        raise
 
     # If we unshared the user namespace the mount propagation of root is changed to slave automatically.
     if not userns:


mkosi/sandbox.py
```diff
@@ -98,8 +99,19 @@ def oserror(filename: str = "") -> None:
...
+restricted unprivileged user namespace creation to reduce exposure to kernel bugs.
+For information about Ubuntu and AppArmor implementations of this restriction,
+see https://ubuntu.com/blog/ubuntu-23-10-restricted-unprivileged-user-namespaces.
+For other systems, research the kernel.unprivileged_userns_clone and
+user.max_user_namespaces sysctls.
+"""
+
 def unshare(flags: int) -> None:
     if libc.unshare(flags) < 0:
+        if ctypes.get_errno() == errno.EPERM:
```
bjackman, line 113: Oh only do this if unsharing userns. otherwise could lead to yet more pain & confusion

DaanDeMeyer (changes requested):
The check should be done in `become_user()`, not in `unshare()`. You can catch the exception there and handle it in the same way.

mkosi/sandbox.py
```diff
@@ -98,8 +99,19 @@ def oserror(filename: str = "") -> None:
...
+            sys.stderr.write(f"\033[0;1;39m{UNSHARE_EPERM_MSG}\033[0m")
```
behrmann, line 114:
```suggestion
            print(f"\033[0;1;39m{UNSHARE_EPERM_MSG}\033[0m", file=sys.stderr)
```
Also, I'd drop the formatting since I don't think it's really necessary.

mkosi/sandbox.py
```diff
@@ -98,8 +99,19 @@ def oserror(filename: str = "") -> None:
...
 
 
+UNSHARE_EPERM_MSG = """
+mkosi was forbidden to unshare the user namespace. This probably means your distro has
+restricted unprivileged user namespace creation to reduce exposure to kernel bugs.
+For information about Ubuntu and AppArmor implementations of this restriction,
+see https://ubuntu.com/blog/ubuntu-23-10-restricted-unprivileged-user-namespaces.
+For other systems, research the kernel.unprivileged_userns_clone and
+user.max_user_namespaces sysctls.
+"""
```
behrmann, line 109: That's a pretty dense block of text. Let's make this a bit lighter so it stands out visually.
```suggestion
UNSHARE_EPERM_MSG = """\
mkosi was unable to unshare the user namespace. 

This probably means your distribution has restricted the creation of unprivileged user namespaces.

Please consult the mkosi man page, e.g. via "mkosi documentation" for workarounds.
"""

```
The specifics for workarounds should go into the manual, where they can be looked up properly and where we can add more text and more examples, e.g. a snippet people can copy.

mkosi/sandbox.py
bjackman: Oh, interesting. It's not actually the `CLONE_NEWUSER` that fails. We're allowed to create userns but then inside that userns you don't get to create the actually interesting namespaces like the mount ns. 

I'm not gonna spend time digging into the exact details of this works since a) it'll be very Ubuntu-specific and b) it won't have much bearing on the ultimate action taken by the user. I'm just gonna be vaguer in the error message.

bjackman:
> The check should be done in become_user(), not in unshare(). You can catch the exception there and handle it in the same way.

Ah sorry didn't see this - see my other comment thread: https://github.com/systemd/mkosi/pull/3266#discussion_r1874779387

I haven't bothered to comprehend the Ubuntu/AppArmor details here but the high level (which I am already appraised of due due to my day job) is: creating user namespaces itself is harmless. The issue is that then you have `CAP_SYS_ADMIN` and that gives you access to buggy kernel code in resources that are within _other_ namespaces. So, we only _actually_ need to prevent the next step where you access that buggy code. So I **guess** Ubuntu is doing something like "you can create the userns, but you can't create the mount/net ns inside it". Ideally it would go more like "you can create the mount ns too, but you can't mount a block filesystem/create a netfilter rule", but AFAIK that's just not really possible with the kernel's namespace infra. 

I think there's _some_ risk that on some system this issue might not manifest in `unshare` at all, but I would say it's better to wait and address that issue if it actually arises.

DaanDeMeyer:
What I'm saying is that `unshare()` is a generic function and shouldn't contain such specific error handling. So please move the error handling to the call of `unshare()` that actually fails. Whether that's the initial user namespace unsharing of something after that (or both).

mkosi/sandbox.py
bjackman: Sounds good - I think "forbidden" is better than "unable" though since we are specifically talking about `EPERM` here? Given what I learned in [this thread](https://github.com/systemd/mkosi/pull/3266#discussion_r1874779387).

bjackman:
> So please move the error handling to the call of unshare() that actually fails. 

Sure, done

mkosi/resources/man/mkosi.1.md
```diff
@@ -2904,6 +2904,31 @@ images you want to build.
...
+```
+abi <abi/4.0>,
+
+include <tunables/global>
+
+/path/to/mkosi flags=(default_allow) {
+  userns,
+
+  # Site-specific additions and overrides. See local/README for details.
+  include if exists <local/opt.google.chrome.chrome>
```
bjackman, line 2928: nope

mkosi/sandbox.py
```diff
@@ -806,7 +817,11 @@ def main() -> None:
     if suppress_chown and (userns or userns_has_single_user()):
         seccomp_suppress_chown()
 
-    unshare(namespaces)
+    try:
+        unshare(namespaces)
+    except OSError as e:
+        if e.errno == EPERM:
+            print(UNSHARE_EPERM_MSG, file=sys.stderr)
```
DaanDeMeyer, line 825: Is this really the only one that can fail due to permission issues? Or is this only when apparmor fails? What about if the sysctls are enabled?

bjackman: Just noticed my Ubuntu system has the `kernel.unprivileged_userns_clone` patch, so I tried that out. And indeed if I set that to 0 the failure happens in `become_user`. So I've just duplicated the logic there.

I think this could easily happen in a third place too, we probably can't make this foolproof, even if we just do it inside `unshare`. (E.g. I know of a certain Linux version that "solves" this problem in a third way that is neither `kernel.unprivileged_userns_clone` nor an LSM policy, I dunno where mkosi would fail on that system, I can't easily test it or share the details either).

mkosi/sandbox.py
```diff
@@ -342,6 +343,9 @@ def become_user(uid: int, gid: int) -> None:
 
     try:
         unshare(CLONE_NEWUSER)
+    except OSError as e:
+        if e.errno == EPERM:
+            print(UNSHARE_EPERM_MSG, file=sys.stderr)
```
DaanDeMeyer, line 348: Needs a `raise` at the end here

DaanDeMeyer (changes requested):

mkosi/sandbox.py
bjackman: Oops thanks

mkosi/resources/man/mkosi.1.md
```diff
@@ -2904,6 +2904,31 @@ images you want to build.
 
 Note that the minimum required Python version is 3.9.
 
+mkosi needs unrestricted abilities to create and act within namespaces. Some
+distros restrict creation of, or capabilities within, user namespaces, which
+breaks mkosi. For information about Ubuntu and AppArmor implementations of this
+restriction, see
+https://ubuntu.com/blog/ubuntu-23-10-restricted-unprivileged-user-namespaces.
```
behrmann, line 2911:
```suggestion
breaks mkosi. 

For information about Ubuntu, that implements such restrictions using AppArmor, see
https://ubuntu.com/blog/ubuntu-23-10-restricted-unprivileged-user-namespaces.
```

mkosi/resources/man/mkosi.1.md
```diff
@@ -2904,6 +2904,31 @@ images you want to build.
...
+For other systems, try researching the `kernel.unprivileged_userns_clone` or
+`user.max.user_namespace` sysctls.
+
+For Ubuntu/AppArmor systems, you should be able to remove the restrictions by
```
behrmann, line 2915:
```suggestion
For Ubuntu systems, you can remove these restrictions for mkosi by
```

mkosi/resources/man/mkosi.1.md
```diff
@@ -2904,6 +2904,31 @@ images you want to build.
...
+adapting this snippet to point to your mkosi binary, copying it to
+/etc/apparmor.d/mkosi, and then running `systemctl reload apparmor`:
```
behrmann, line 2917:
```suggestion
`/etc/apparmor.d/mkosi`, and then running `systemctl reload apparmor`:
```

mkosi/sandbox.py
behrmann: Even though it's `EPERM` I think "unable" is a nicer way of putting it, since it's not mkosi being singled out, but this being a flat policy that mkosi wasn't yet exempted from. What I'm saying is that "forbidden" sounds meaner than being collateral damage of wider security policy.

mkosi/resources/man/mkosi.1.md
```diff
@@ -2904,6 +2904,31 @@ images you want to build.
...
+```
+abi <abi/4.0>,
+
+include <tunables/global>
+
+/path/to/mkosi flags=(default_allow) {
+  userns,
+
+  # Site-specific additions and overrides. See local/README for details.
+  include if exists <local/mkosi>
```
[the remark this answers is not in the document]
bjackman: Good point

mkosi/resources/man/mkosi.1.md
bjackman: [Seems you are right](https://ubuntu.com/server/docs/apparmor#profiles), only of a minority of the ones on my own Ubuntu system follow that style, dunno if that's because they are special or just from before the style emerged or what. Anyway, will do what the docs say, thanks.

bjackman:
Been a long time since I did any real work in the GitHub UI, it's got fancy new features (they still haven't fixed https://github.com/isaacs/github/issues/386 though??), LMK if I'm doing anything stupid here.

DaanDeMeyer (approved):

behrmann (approved):

PR #3270, opened 2024-12-12
Rework crypto-policies handling slightly

DaanDeMeyer:
Let's bind mount from the tools tree if it provides crypto policies and only write our own for rpm-sequoia if the tools tree doesn't provide its own crypto policies.


bluca (approved):
The bot says it's all good, so it must be true

behrmann (approved):

PR #3271, opened 2024-12-12
Use host model cpu

teknoraver:
mkosi currently calls qemu with `-cpu max`. This enables all features supported by the accelerator, which could be less than the ones supported by the host.

Switch to the default `-cpu host` and keep max only when using binary translation.


behrmann (approved):
LGTM, if I understand this correctly, i.e. use `-cpu host`, which is the default, in case of we're on the same arch (or binary compatible).

teknoraver:
@behrmann correct!

PR #3275, opened 2024-12-13
pacman: Make sure repositories from dropins take priority

DaanDeMeyer:
When multiple repositories ship the same package, the repository defined first in the pacman config file takes priority. Let's make sure user defined repositories take priority over the ones defined in mkosi by moving the Include= statement up a little in the config file.

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -45,6 +45,15 @@ runs:
         sudo sysctl --ignore --write kernel.apparmor_restrict_unprivileged_unconfined=0
         sudo sysctl --ignore --write kernel.apparmor_restrict_unprivileged_userns=0
 
+    - name: Create missing mountpoints
+      shell: bash
+      run: |
+        for p in /etc/pki /etc/pacman.d/gnupg /etc/ssl /etc/ca-certificates /var/lib/ca-certificates /etc/crypto-policies; do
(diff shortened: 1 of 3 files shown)


action.yaml
```diff
@@ -45,6 +45,15 @@ runs:
...
         sudo sysctl --ignore --write kernel.apparmor_restrict_unprivileged_userns=0
 
+    - name: Create missing mountpoints
+      shell: bash
+      run: |
+        for p in /etc/pki /etc/pacman.d/gnupg /etc/ssl /etc/ca-certificates /var/lib/ca-certificates /etc/crypto-policies; do
+          if [[ ! -e "$p" ]]; then
+            sudo mkdir -p "$p"
+          fi
+        done
```
behrmann, line 55: Why not
```suggestion
        sudo mkdir -p /etc/pki /etc/pacman.d/gnupg /etc/ssl /etc/ca-certificates /var/lib/ca-certificates /etc/crypto-policies
```
since `mkdir -p` is idempotent?

behrmann (approved):

action.yaml
DaanDeMeyer: It'll fail if any of these is a symlink and not a directory I think.

Issue #3276, opened 2024-12-13
`SplitArtifacts=` doesn't work for `Format=uki`


Winterhuman:
### mkosi commit the issue has been seen with

24.3

### Used host distribution

Arch Linux

### Used target distribution

Arch Linux

### Linux kernel version used

6.12.4-arch1-1

### CPU architectures issue was seen on

x86_64

### Unexpected behaviour you saw

`SplitArtifacts=`/`SplitArtifacts=no` still produces `.initrd` & `.vmlinuz` files under `mkosi.output/` when using `Format=uki`. Ideally, only the final `.efi` UKI file would be produced.

### Used mkosi config

```ini
[Distribution]
Distribution=arch

[Content]
Packages=
	base
	linux

[Output]
Format=uki
ImageId=repro
SplitArtifacts=
PackageCacheDirectory=mkosi.pkgcache
```

### mkosi output

```sh
‣ Including configuration file /home/user/Public/mkosi builds/repro/mkosi.conf
‣ + timedatectl show -p Timezone --value
‣ + timedatectl show -p Timezone --value
‣ + findmnt --target / --output FSTYPE --noheadings
‣ Acquiring lock on /tmp/mkosi-uidmap-lock-r85scrf4
‣ Acquired lock on /tmp/mkosi-uidmap-lock-r85scrf4
‣ + flock --exclusive --close /tmp/mkosi-uidmap-lock-r85scrf4 newuidmap 17257 0 100000 65436 65436 1000 1 65437 165437 99
‣ + flock --exclusive --close /tmp/mkosi-uidmap-lock-r85scrf4 newgidmap 17257 0 100000 65436 65436 1000 1 65437 165437 99
‣ Removing output files of repro image…
‣ Acquiring lock on /home/user/Public/mkosi builds/repro/mkosi.output/repro
‣ Acquired lock on /home/user/Public/mkosi builds/repro/mkosi.output/repro
‣ + rm -rf -- '/home/user/Public/mkosi builds/repro/mkosi.output/repro' '/home/user/Public/mkosi builds/repro/mkosi.output/repro.efi' '/home/user/Public/mkosi builds/repro/mkosi.output/repro.initrd' '/home/user/Public/mkosi builds/repro/mkosi.output/repro.vmlinuz'
‣ + stat --file-system --format %T /var/tmp/mkosi-workspace-rdma92cf
‣ + btrfs subvolume create /var/tmp/mkosi-workspace-rdma92cf/root
[...]

PR #3277, opened 2024-12-13
Fix `SplitArtifacts=` to not split extra artifacts

NekkoDroid:
Fixes: https://github.com/systemd/mkosi/issues/3276


mkosi/config.py
```diff
@@ -1381,7 +1381,7 @@ def config_parse_artifact_output_list(
     value: Optional[str], old: Optional[list[ArtifactOutput]]
 ) -> Optional[list[ArtifactOutput]]:
```
behrmann, line 1382: With the below change `None` is no longer an expected output.

NekkoDroid: The cast was technically also wrong on my end. We can still get `None` from `config_make_list_parser`

PR #3279, opened 2024-12-15
Polish command line only option descriptions

matmaer:


behrmann:
There's a few commits in here, that shouldn't be. Can you reset the branch to current main and the new changes on top of that? One waymight be
```bash
git checkout main
git pull <name for your upstream remote> main
# save this branch under a different name
git branch -m polish-man polish-man_v0
git switch -c polish-man
git cherry-pick --no-commit 282d428 14e610d
git commit -m 'man: polish command line only option descriptions'
git push --force-with-lease <name of your fork> polish-man
```

mkosi/resources/man/mkosi.1.md
```diff
@@ -204,80 +204,67 @@ The following command line verbs are known:
 
 ## Commandline-only Options
 
-Those settings cannot be configured in the configuration files.
+The following settings cannot be configured in the configuration files:
+
+`--help`, `-h`
+:   Show brief usage information.
```
behrmann, line 210: Moving it away from the middle makes sense, but first is too prominent and nobody looking at the manual will look there for this. Better move it to the end, just like in the actual usage info.

mkosi/resources/man/mkosi.1.md
```diff
@@ -204,80 +204,67 @@ The following command line verbs are known:
...
-    specifying this option twice will ensure the intermediary
-    cache files are removed, too, before the re-build is initiated. If a
-    package cache is used (also see the **Files** section below),
-    specifying this option thrice will ensure the package cache is
-    removed too, before the re-build is initiated. For the `clean`
-    operation this option has a slightly different effect: by default
-    the verb will only remove build artifacts from a previous run, when
-    specified once the incremental cache files are deleted too, and when
-    specified twice the package cache is also removed.
+:   Replace the output file if it already exists, by default `mkosi`refuses to overwrite an existing image.
```
behrmann, line 213: Missing space, also there's no need to join both sentences.
```suggestion
:   Replace the output file if it already exists. By default `mkosi` refuses to overwrite an existing image.
```

mkosi/resources/man/mkosi.1.md
```diff
@@ -204,80 +204,67 @@ The following command line verbs are known:
...
+    This option can be specified once, twice or trice to change the scope of artifact removals.
```
behrmann, line 214: Typo
```suggestion
    This option can be specified once, twice or thrice to change the scope of artifact removals.
```
but if we're touching this we should probably use something less poetic
```suggestion
    This option can be specified once, twice or three times to change the scope of artifact removals.
```
Same below.

mkosi/resources/man/mkosi.1.md
```diff
@@ -204,80 +204,67 @@ The following command line verbs are known:
...
+    Specify once to delete all build artifacts from a previous run. Twice to include removal of the
+    intermediary package cache from incremental builds. Thrice to remove all build artifacts and the the full
+    package cache. See the **Files** section for more context and details.
+    If `mkosi` is followed by the `clean` verb, the `-f` option can only be specified twice, see the
+    description of the command line verbs above.
```
behrmann, line 219: This does make it shorter, but I think we shouldn't "Mario, your commandline option description is in another castle." and describe the version of clean in a separate paragraph here as well.

mkosi/resources/man/mkosi.1.md
```diff
@@ -204,80 +204,67 @@ The following command line verbs are known:
...
 
 `--directory=`, `-C`
-:   Takes a path to a directory. `mkosi` switches to this directory before
-    doing anything. Note that the various configuration files are searched
-    for in this directory, hence using this option is an effective way to
-    build a project located in a specific directory.
+:   Specify a path to a directory. `mkosi` switches to this directory before doing anything. Note that the
+    various configuration files are searched for in this directory by default. Using this option can be used
+    to take advantage of defaults when executing mkosi inside of a profile directory or to execute `mkosi` in
+    any other directory deemed useful.
```
behrmann, line 225: I don't think this improves on the previous version. "Takes a path" is a common way of putting this, since functions take arguments and in a more abstract sense mkosi is a function and this is a way to pass such an argument. 

The "by default" at the end of third sentence is superfluous with respect to the rest of the sentence and the last sentence is a lot of words that say very little. It should also not mention profiles, since that's a different concept in mkosi that doesn't have anything to do with this.

mkosi/resources/man/mkosi.1.md
```diff
@@ -204,80 +204,67 @@ The following command line verbs are known:
...
 
 `--debug`
-:   Enable additional debugging output.
+:   Enable additional debugging output. Also useful to observe from time to time when everything just works.
+    This pro-active approach leads to a better understanding on how `mkosi` works for future usage.
```
behrmann, line 229: I would not advertise doing debug logging proactively. If things just work, there is little benefit in this. We have debug output in CI and it's a lot.

mkosi/resources/man/mkosi.1.md
```diff
@@ -204,80 +204,67 @@ The following command line verbs are known:
...
 
 `--debug-shell`
-:   When executing a command in the image fails, mkosi will start an interactive
-    shell in the image allowing further debugging.
+:   When executing a command that fails to build an image, mkosi will start an interactive shell inside of the
```
behrmann, line 232: The previous version is actually correct, since this triggers when a command fails that is being run in the image being built. That being said, we can probably just say "When executing failing command during an image build".

mkosi/resources/man/mkosi.1.md
```diff
@@ -204,80 +204,67 @@ The following command line verbs are known:
...
+    interrupted or failed image build. This feature helps to acquire additional insights for troubleshooting.
```
behrmann, line 233: Style: interrupted or failed, not both.
```suggestion
    interrupted or failed image build. This feature helps to acquire additional insights for troubleshooting.
```
The last sentence is too abstract, more concrete would be something like "This can be used to debug failing build scripts, see the **Scripts** section."

mkosi/resources/man/mkosi.1.md
```diff
@@ -204,80 +204,67 @@ The following command line verbs are known:
...
 
 `--debug-sandbox`
 :   Run `mkosi-sandbox` with `strace`.
 
 `--version`
-:   Show package version.
-
-`--help`, `-h`
-:   Show brief usage information.
+:   Show package version, build from source to have the latest version ahead of a new release.
```
behrmann, line 243: This should stay as is, but probably move to the end together with the help description.

mkosi/resources/man/mkosi.1.md
```diff
@@ -204,80 +204,67 @@ The following command line verbs are known:
...
 
 `--genkey-common-name=`
-:   Common name to be used when generating keys via mkosi's `genkey` command. Defaults to `mkosi of %u`, where
-    `%u` expands to the username of the user invoking mkosi.
+:   Common name to be used when generating keys via mkosi's `genkey` command. Defaults to `mkosi of %u`,
+    where `%u` expands to the username of the user invoking mkosi.
```
behrmann, line 247: This is not changed, is it? If it's not changed it should appear in the diff.

mkosi/resources/man/mkosi.1.md
```diff
@@ -204,80 +204,67 @@ The following command line verbs are known:
...
 :   Number of days that the keys should remain valid when generating keys via mkosi's `genkey` command.
     Defaults to two years (730 days).
 
 `--auto-bump=`, `-B`
-:   If specified, after each successful build the version is bumped in a
-    fashion equivalent to the `bump` verb, in preparation for the next
-    build. This is useful for simple, linear version management: each
-    build in a series will have a version number one higher then the
-    previous one.
+:   Automatically increment, or bump, the version know to `mkosi`, after a successful build. Each build in a
```
behrmann, line 254: Typo
```suggestion
:   Automatically increment, or bump, the version known to `mkosi`, after a successful build. Each build in a
```
The insert "or bump" should probably be dropped for brevity. This should also still mention that this is only really useful for simple linear versions.

mkosi/resources/man/mkosi.1.md
```diff
@@ -204,80 +204,67 @@ The following command line verbs are known:
...
+    series will have a version number one higher then the previous one. See also the description of the
+    related `bump` verb above.
```
behrmann, line 256:
```suggestion
    related `bump` verb above and `mkosi.version` in the **Files** section.
```

mkosi/resources/man/mkosi.1.md
```diff
@@ -204,80 +204,67 @@ The following command line verbs are known:
...
-    documentation is shown in the original Markdown format. `man` shows the
-    documentation in man page format, if it is available. `pandoc` will generate
-    the man page format on the fly, if `pandoc` is available. `system` will show
-    the system-wide man page for mkosi, which may or may not correspond to the
-    version you are using, depending on how you installed mkosi. `auto`, which is
-    the default, will try all methods in the order `man`, `pandoc`, `markdown`,
-    `system`.
+:   The format to view documentation. The default, `auto` will try the 4 supported methods in this order:
+    `man`, `pandoc`,`markdown` and `system`.  This is done in an effort to avoid unsupported formats on the
+    host. `system` will show the system-wide man page, `pandoc` can generate the man page format on the fly.
```
behrmann, line 261: That `auto` tries all of them is less important that describing all options and this loses the descriptions of `man` and `markdown`, also `system` has lost the somewhat important information, that while the other options come from something that this particular installation of mkosi controls, no matter how it arrived on your disk, the last one might not and therefore might not reflect reality. The previous version could have done with a few paragraphs, though.

mkosi/resources/man/mkosi.1.md
```diff
@@ -204,80 +204,67 @@ The following command line verbs are known:
...
 
 `--json`
 :   Show the summary output as JSON-SEQ.
 
 `--wipe-build-dir`, `-w`
-:   Wipe the build directory if one is configured before building the image.
+:   Unconditionally wipe the build directory if one is configured and found.
```
behrmann, line 267: I don't think this improves on the previous version.

behrmann (changes requested):

mkosi/resources/man/mkosi.1.md
matmaer: ok although this takes effect without building a new image, and is obviously used before building a new image.

mkosi/resources/man/mkosi.1.md
matmaer: I don't agree, I think it's better to make the man page shorter by referring to previously declared descriptions, when available. This principle can be applied to many repetitions throughout the document.  So it's rather important to continue on the same page going forward.

mkosi/resources/man/mkosi.1.md
matmaer: I would argue to look at it from a user's standpoint, that is not at your technical level.  Also usage that is "blind" to any CI as mkosi can be used as a simple command to build an image without considering CI.

mkosi/resources/man/mkosi.1.md
matmaer: ok I see, thanks for the insight

mkosi/resources/man/mkosi.1.md
matmaer: My goal is to start each description with the value and default for the option or setting at hand.  Currently this is inconsistent, hampering the ability to skim over the document. This is another "general principle" I'd like to defend.

matmaer: For less technical users, and by including "value", it's pretty obvious we are talking about what the option "takes", "accepts", "honors" or anything else.

mkosi/resources/man/mkosi.1.md
matmaer: I was in doubt about this, and it looks weird indeed, agreed.

mkosi/resources/man/mkosi.1.md
matmaer: I didn't change this so following your own correction for sure.

mkosi/resources/man/mkosi.1.md
matmaer: It is meant to offload support questions by users to their own capability, by guiding them to information that can educate them.

mkosi/resources/man/mkosi.1.md
matmaer: ok I see, thank for the guidance, can you commit the change? I don't see your suggestion in the proposed correction.

matmaer:
> There's a few commits in here, that shouldn't be. Can you reset the branch to current main and the new changes on top of that? One waymight be
> 
> ```shell
> git checkout main
> git pull <name for your upstream remote> main
> # save this branch under a different name
> git branch -m polish-man polish-man_v0
> git switch -c polish-man
> git cherry-pick --no-commit 282d428 14e610d
> git commit -m 'man: polish command line only option descriptions'
> git push --force-with-lease <name of your fork> polish-man
> ```

I didn't mean to include the previous commits, I tried to avoid this. Even after re-cloning, deleting and re-creating the polish-man branch, which I assumed had another hash, it still showed up.

matmaer:
> There's a few commits in here, that shouldn't be. Can you reset the branch to current main and the new changes on top of that? One waymight be
> 
> ```shell
> git checkout main
> git pull <name for your upstream remote> main
> # save this branch under a different name
> git branch -m polish-man polish-man_v0
> git switch -c polish-man
> git cherry-pick --no-commit 282d428 14e610d
> git commit -m 'man: polish command line only option descriptions'
> git push --force-with-lease <name of your fork> polish-man
> ```

OK trying your suggestion, thanks for the guidance.

matmaer (commented):
It's weird I did it before but somehow I messed up, please cherry pick on your end. Thanks and hopefully better git flow next time.

PR #3280, opened 2024-12-17
Add initrd-addon output format

DaanDeMeyer:
Similar to how we can produce sysext and confext extensions, let's also
support building initrd addons.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2112,11 +2112,19 @@ def install_kernel(context: Context, partitions: Sequence[Partition]) -> None:
 
 
 def make_uki(
-    context: Context, stub: Path, kver: str, kimg: Path, microcode: list[Path], output: Path
+    context: Context,
+    stub: Path,
+    kver: str,
+    kimg: Path,
+    microcode: list[Path],
+    output: Path,
 ) -> None:
     make_cpio(context.root, context.workspace / "initrd", sandbox=context.sandbox)
     maybe_compress(
-        context, context.config.compress_output, context.workspace / "initrd", context.workspace / "initrd"
+        context,
+        context.config.compress_output,
+        context.workspace / "initrd",
+        context.workspace / "initrd",
     )
 
     initrds = [context.workspace / "initrd"]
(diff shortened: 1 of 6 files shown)


mkosi/__init__.py
```diff
@@ -2142,6 +2147,34 @@ def make_uki(
...
+    )
+    arguments: list[PathString] = [
+        "--initrd",
+        workdir(context.workspace / "initrd"),
+    ]
+    options: list[PathString] = [
+        "--ro-bind",
+        context.workspace / "initrd",
+        workdir(context.workspace / "initrd"),
+    ]
```
behrmann, line 2166: Nit: since they are arguments for the options
```suggestion
    arguments: list[PathString] = [
        "--initrd", workdir(context.workspace / "initrd"),
    ]  # fmt: skip
    options: list[PathString] = [
        "--ro-bind", context.workspace / "initrd",  workdir(context.workspace / "initrd"),
    ]  # fmt: skip
```

mkosi/config.py
```diff
@@ -189,24 +189,36 @@ class OutputFormat(StrEnum):
...
     def use_outer_compression(self) -> bool:
-        return self in (OutputFormat.tar, OutputFormat.cpio, OutputFormat.disk) or self.is_extension_image()
+        return self in (
+            OutputFormat.tar,
+            OutputFormat.cpio,
+            OutputFormat.disk,
+            OutputFormat.sysext,
+            OutputFormat.confext,
+            OutputFormat.portable,
+        )
```
behrmann, line 215: That change should probably be in the first commit instead of the last.

mkosi/__init__.py
```diff
@@ -3057,9 +3090,16 @@ def reuse_cache(context: Context) -> bool:
     return True
 
 
-def save_uki_components(
+def save_efi_components(
```
behrmann, line 3093: The renaming seemed a bit weird to me upon first glance, but I understand why with a separate file you don't want to call it a UKI component. EFI component still sounds a bit weird to me, maybe ESP component?

behrmann (approved):

PR #3282, opened 2024-12-18
Drop PEAddons=

DaanDeMeyer:
Putting these inside the image isn't very useful except for the very specific use case of being able to skip copying the addon when copying the ESP to another disk to implement a "live" installer mode. For all other use cases, PE addons are either supposed to be generated locally or optionally downloaded alongside the image. For these use cases it makes much more sense to output the addons alongside the image, but not in it.

But if it makes more sense to output the addons alongside the image, then it becomes unclear whether this is something that should be implemented inside mkosi in the first place. Until we figure whether this makes sense and how to implement it, let's remove the functionality.

Addons can still be added to the image or generated alongside it by running ukify in a postinst script. The secure boot signing options will have to be passed manually but this isn't an unsolvable problem.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1996,30 +1996,6 @@ def install_uki(
             f.write("fi\n")
 
 
-def install_pe_addons(context: Context) -> None:
-    if not context.config.pe_addons:
-        return
-
-    stub = systemd_addon_stub_binary(context)
-    if not stub.exists():
-        die(f"sd-stub not found at /{stub.relative_to(context.root)} in the image")
-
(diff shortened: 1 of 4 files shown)


DaanDeMeyer:
cc @NekkoDroid

behrmann (approved):
This needs a news entry that people needing this functionality will need to either use a build script or stay on an older mkosi version.

behrmann:
Merging, because @NekkoDroid gave a :+1:  on the PR description.

behrmann:
Wasn't in a release yet, so dropping the label.

PR #3286, opened 2024-12-19
Run binaries from ExtraSearchPaths= within tools tree

DaanDeMeyer:
Until now, there was always the implicit assumption that any paths configured with ExtraSearchPaths= contained binaries built against the host's /usr. With that assumption, it made sense to execute binaries found in these paths outside of the tools tree as otherwise you might end up with missing or out of date libraries if these are not available within the tools tree.

However, with the introduction of "mkosi sandbox", what I want to do in systemd is to have contributors build systemd within the sandbox and then use those binaries in mkosi to build the image. This means that the build directory configured with ExtrasSearchPaths= suddenly contains binaries built against the tools tree's /usr (if one is configured) instead of the host's /usr.

Given this new use case, let's get rid of the logic to not use the tools tree for binaries in ExtraSearchPaths=, instead, users of ExtraSearchPaths= using a tools tree will have to make sure to use a tools tree that mostly matches their host's /usr.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -498,10 +498,7 @@ def finalize_host_scripts(
         if context.config.find_binary(binary):
             scripts[binary] = (binary, "--root", "/buildroot")
     if ukify := context.config.find_binary("ukify"):
-        # A script will always run with the tools tree mounted, so we pass binary=None to disable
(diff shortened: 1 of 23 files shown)


behrmann (approved):

behrmann:
Nice. A feature implemented by removing code, making both the code and the logic of the feature easier to understand.

PR #3293, opened 2024-12-20
burn: Fail if no arguments are provided

DaanDeMeyer:
If no arguments are provided systemd-repart will operate on the device backing the rootfs which in the case of burn we definitely don't want to do most of the time so let's fail unless a device is explicitly provided.

### DIFF
--- a/mkosi/burn.py
+++ b/mkosi/burn.py
@@ -12,6 +12,9 @@ def run_burn(args: Args, config: Config) -> None:
     if config.output_format not in (OutputFormat.disk, OutputFormat.esp):
         die(f"{config.output_format} images cannot be burned to disk")
 
+    if not args.cmdline:
+        die("Please specify a device to burn the image to", hint="For example /dev/disk/by-id/usb-foobar")
+
     fname = config.output_dir_or_cwd() / config.output
 
     if len(args.cmdline) != 1:


mkosi/burn.py
```diff
@@ -12,6 +12,9 @@ def run_burn(args: Args, config: Config) -> None:
     if config.output_format not in (OutputFormat.disk, OutputFormat.esp):
         die(f"{config.output_format} images cannot be burned to disk")
 
+    if not args.cmdline:
+        die("Please specify a device to burn the image to", hint="For example /dev/sdb")
```
behrmann, line 16: Our userbase will most likely not fall into the trap, but I'd rather not name a device that has chance of actually existing, how about `/dev/sdz`, which should be safe if there's not too many drives (under the assumption somebody sitting in front of machine with this many drives knows what they are doing) or `/dev/nvme1n1`, since I think the first number is a namespace and those are not in use in the wild.

behrmann (approved):
LGTM in general, one comment.

mkosi/burn.py
septatrix: Even better (though a bit more verbose) might be `/dev/disk/by-id/usb-foobar` which I have come to use since its even harder to mess those up (and I do not get confused when switching between laptop and desktop with different disk types/counts)

behrmann (approved):

PR #3294, opened 2024-12-20
Don't spawn a shell in sandbox if no command is provided

DaanDeMeyer:
We need https://github.com/systemd/systemd/issues/35000 to make this not confusing. Currently there's zero feedback that a user is in a sandbox shell.


behrmann (approved):
LGTM, though I prefer the current behaviour. The distinction is a bit whether mentally this is like `chroot` or like `bwrap`. Since it's closer to `bwrap`, being explicit in the command makes sense.

PR #3298, opened 2024-12-22
Wait until sandbox exec()'s specified command before continuing

DaanDeMeyer:
Let's wait in spawn() until all setup logic has completed before
continuing if a sandbox is used. When using spawn() without run(),
this helps us fail early if the setup logic fails instead of later
on if we're not checking the result of the spawn() command immediately
afterwards.

As a side effect, this also acts as a synchronization point when using
systemd-run --scope where when spawn() returns, we can be 100% sure that
the scope has been created, which is important when calling RegisterMachine
from systemd-machined which does the wrong thing if the scope for the specified
machine does not yet exist and ends up killing the parent unit (often the user
session) instead of just the virtual machine.

### DIFF
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -3,6 +3,7 @@
 import contextlib
 import errno
 import fcntl
+import functools
 import itertools
 import logging
 import os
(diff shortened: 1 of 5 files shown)


mkosi/run.py
```diff
@@ -219,44 +264,27 @@ def spawn(
...
-            assert nfd == SD_LISTEN_FDS_START + i
+            # Make sure the write end of the pipe (which we pass to the subprocess) is higher than all the
+            # file descriptors we'll pass to the subprocess, so that it doesn't accidentally get closed by
+            # the logic in preexec().
+            if pass_fds:
+                pwfd = fd_move_above(pwfd, list(pass_fds)[-1])
 
-    with sandbox as sbx:
-        prefix = [os.fspath(x) for x in sbx]
+            exec_prefix = ["--exec-fd", str(SD_LISTEN_FDS_START + len(pass_fds)), "--"]
```
behrmann, line 279: I think this would be slightly more readable.
```suggestion
            exec_prefix = ["--exec-fd", f"{SD_LISTEN_FDS_START + len(pass_fds)}", "--"]
```
Looking at this I had to remember that `SD_LISTEN_FDS_START` is 3 to remember that this expression makes sense, because the parentheses are a bit non-obvious. The f-string is a bit unnecessary, but sidesteps this.

behrmann (approved):

Issue #3176, opened 2024-11-11
Add a mkosi.init script to run before configuration files parsing happens


davide125:
### mkosi commit the issue has been seen with

main

### Used host distribution

Fedora Linux 41

### Used target distribution

Fedora Linux 41

### CPU architectures issue was seen on

None

### Unexpected behaviour you saw

I'd like to have a `mkosi.init` script to run at the very beginning (before config parsing/validation) in order to create/initialize the contents of things like `PackageDirectories`  and `ExtraSearchPaths`. The specific usecase I have in mind is building systemd as part of the mkosi build, and including the systemd packages into the image built by mkosi. My attempt at that involved cloning systemd in the mkosi directory, and then adding a `mkosi.sync` with `mkosi --directory=systemd --release=41 -t none` to build it. However, this runs too late, and mkosi ends up failing because the build directory doesn't exist yet.

### Used mkosi config

```ini
[Distribution]
Distribution=fedora
Release=41

[Content]
PackageDirectories=systemd/build/mkosi.builddir/%d~%r~%a

[Host]
ExtraSearchPaths=systemd/build/mkosi.builddir/%d~%r~%a
```

### mkosi output

```sh
‣ systemd/build/mkosi.builddir/fedora~41~x86-64 does not exist
```

davide125:
Obviously, a workaround for this is manually running `mkdir systemd/build/mkosi.builddir/fedora~41~x86-64` before attempting to build, but that's an additional manual step I'd like to avoid if possible.

DaanDeMeyer:
So I'm conflicted on this one but I also think this is a bit too niche to warrant introducing an extra script.

I'd suggest reworking your setup to not build systemd in mkosi.sync but instead have a wrapper script that builds systemd and then execs into mkosi, and passthrough all arguments to that script to mkosi. Then you keep full flexibility of passing CLI options to mkosi, and the only con is that you have to invoke a separate script instead of mkosi directly, but this is the same limitation as in the systemd repository, where you have to run `mkosi sandbox meson compile -C build` to build systemd on the host first before running mkosi.

PR #3300, opened 2024-12-22
Always check tools for build at the same place

DaanDeMeyer:
Let's only check non-build verb tools early and check build verb tools at the same place regardless of whether the verb is build or not to keep things more consistent.

As a side effect this allows building tools from source to do the build in sync scripts which is done in
https://github.com/davide125/arcadeos/blob/main/mkosi.sync.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -4725,7 +4725,9 @@ def run_verb(args: Args, images: Sequence[Config], *, resources: Path) -> None:
 
     for i, config in enumerate(images):
         with prepend_to_environ_path(config):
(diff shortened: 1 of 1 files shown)


behrmann (approved):
LGTM, but the link to Arcade OS in the commit should be to a commit instead of main, otherwise this will bitrot.

davide125:
Tested in an ubuntu 24.04 container and it seems to work as expected, thanks! Related to this, it might make sense to also push some path validation down to when it's actually needed (specifically `ExtraSearchPaths` and `PackageDirectories`, see #3176 for more details).

PR #3302, opened 2024-12-23
Drop support for pesign secure boot signtool

DaanDeMeyer:
sbsigntools is packaged in EPEL now and in the future we're looking to standardize solely on systemd-sbsign so let's already drop support for pesign which was only really useful on CentOS Stream because it didn't have sbsigntools which it does have now.


behrmann (approved):
LGTM, but should go in after #3300, since it includes it.

PR #3304, opened 2024-12-26
Add option to control whether to register with machined

bluca:
On an ephemeral test runner it is useless churn to register the guest, and can trigger nasty bugs like the one where machined kills the user session. Add a config knob, enabled by default, so that it can be disabled where it's not useful.

### DIFF
--- a/mkosi/resources/man/mkosi.1.md
+++ b/mkosi/resources/man/mkosi.1.md
@@ -1801,6 +1801,10 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     Note that `Ephemeral=` has to be enabled to start multiple instances
     of the same image.
 
+`Register=`, `--register=`
+:   Takes a boolean value. Enabled by default. Specifies whether to register
+    the vm/container with systemd-machined.
+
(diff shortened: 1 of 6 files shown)


bluca:
This also allows running nspawn where machined is not available, as currently it just fails since the default is that it tries to register and fails if it can't.

mkosi/resources/man/mkosi.1.md
```diff
@@ -1801,6 +1801,10 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     Note that `Ephemeral=` has to be enabled to start multiple instances
     of the same image.
 
+`MachinedRegister=`, `--machined-register=`
+:   Takes a boolean value, defaults to `true`. Specifies whether to register
```
behrmann, line 1805:
```suggestion
:   Takes a boolean value. Enabled by defaults. Specifies whether to register
```

behrmann (approved):

DaanDeMeyer (changes requested):
Looks great but I think we can just call the option `Register=`.

DaanDeMeyer (approved):

PR #3305, opened 2024-12-27
Rename various [Runtime] options and rename mkosi qemu to mkosi vm

DaanDeMeyer:
- Let's get rid of the "Qemu" prefix for runtime options as they apply both to the qemu and vmspawn verbs.
- Let's rename QemuGui= to Console= and make it take all the options that are accepted by vmspawn's --console= option.
- Let's rename the various runtime options to the corresponding vmspawn option where applicable.
- Let's rename mkosi qemu to mkosi vm as well while we're at it.


mkosi/config.py
```diff
@@ -3537,108 +3546,136 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
         help="Directory containing systemd-sysupdate transfer definitions",
     ),
     ConfigSetting(
-        dest="qemu_gui",
-        metavar="BOOL",
+        dest="console",
```
behrmann, line 3549: This needs a non-`None` default.

PR #3311, opened 2025-01-02
Make CDROM an upper case option

DaanDeMeyer:
It's an abbrevation so let's make it upper case similar to TPM and KVM.


behrmann (approved):
The shouty-case for `CDROM=` doesn't look great, but it is slightly more consistent.

PR #3312, opened 2025-01-02
Modules and firmware includes

hundeboll:
In order to include out-of-tree modules (and their required firmware) installed in an `updates` subfolder, a special case is added when a file is found below that folder.

Furthermore, a firmware file required by a module might be a symbolic link to the real firmware file. Add handling of this case too.

### DIFF
--- a/mkosi/kmod.py
+++ b/mkosi/kmod.py
@@ -10,6 +10,7 @@ from pathlib import Path
 
 from mkosi.log import complete_step, log_step
 from mkosi.run import chroot_cmd, run
+from mkosi.sandbox import chase
 from mkosi.util import chdir, parents_below
 
 
@@ -179,6 +180,15 @@ def gen_required_kernel_modules(
         with chdir(root):
             mods = set(modulesd.rglob("*.ko*"))
         firmware = set()
+
+    # Some firmware dependencies are symbolic links, so the targets for those must be included in the list
+    # of required firmware files too. Intermediate symlinks are not included, and so links pointing to links
+    # results in dangling symlinks in the final image.
+    for fw in firmware.copy():
+        if (root / fw).is_symlink():
+            target = Path(chase(os.fspath(root), os.fspath(fw)))
+            if target.exists():
+                firmware.add(target.relative_to(root))
 
     yield from sorted(
         itertools.chain(
--- a/mkosi/resources/man/mkosi.1.md
+++ b/mkosi/resources/man/mkosi.1.md
@@ -1016,10 +1016,10 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 
 `KernelModulesInclude=`, `--kernel-modules-include=`
 :   Takes a list of regex patterns that specify kernel modules to include in the image. Patterns should be
-    relative to the `/usr/lib/modules/<kver>/kernel` directory. mkosi checks for a match anywhere in the module
-    path (e.g. `i915` will match against `drivers/gpu/drm/i915.ko`). All modules that match any of the
-    specified patterns are included in the image. All module and firmware dependencies of the matched modules
-    are included in the image as well.
+    relative to `/usr/lib/modules/<kver>/<subdir>` paths. mkosi checks for a match anywhere in the module path
+    (e.g. `i915` will match against `drivers/gpu/drm/i915.ko`). All modules that match any of the specified
+    patterns are included in the image. All module and firmware dependencies of the matched modules are included
+    in the image as well.
 
     If the special value `default` is used, the default kernel modules
     defined in the `mkosi-initrd` configuration are included as well.
--- a/tests/test_initrd.py
+++ b/tests/test_initrd.py
@@ -225,9 +225,9 @@ def test_initrd_size(config: ImageConfig) -> None:
         maxsize = 1024**2 * {
             Distribution.fedora: 63,
             Distribution.debian: 62,
-            Distribution.ubuntu: 56,
+            Distribution.ubuntu: 57,
             Distribution.arch: 83,
             Distribution.opensuse: 64,
-        }.get(config.distribution, 57)
+        }.get(config.distribution, 58)
 
         assert (Path(image.output_dir) / "image.initrd").stat().st_size <= maxsize


mkosi/kmod.py
```diff
@@ -234,12 +238,27 @@ def process_kernel_modules(
                 p.rmdir()
 
         for fw in firmware:
+            # assume any files in /usr/lib/firmware/updates are required
+            if fw.is_relative_to(firmwared / "updates"):
+                continue
+
+            if fw in required and (root / fw).is_symlink():
+                # mopdules may depend on a symbolic link to the actual firmware file, so
```
behrmann, line 246:
```suggestion
                # modules may depend on a symbolic link to the actual firmware file, so
```
there's also typos in the commit messages, `lnk` instead of `link` in the first and `an` instead of `and` in the second.

mkosi/kmod.py
```diff
@@ -218,12 +218,16 @@ def process_kernel_modules(
...
 
         with chdir(root):
             modules = sorted(modulesd.rglob("*.ko*"), reverse=True)
             firmware = sorted(firmwared.rglob("*"), reverse=True)
 
         for m in modules:
+            if m.is_relative_to(modulesd / "updates"):
+                continue
+
             if m in required:
```
behrmann, line 231: This is a bit shorter and the comment below, would probably make more sense here. 
```suggestion
            if m in required or m.is_relative_to(modulesd / "updates"):
```
That being said, if files under `updates/` are required, that logic should be in `gen_required_kernel_modules`.

mkosi/kmod.py
```diff
@@ -234,12 +238,27 @@ def process_kernel_modules(
...
+            if fw in required and (root / fw).is_symlink():
+                # mopdules may depend on a symbolic link to the actual firmware file, so
+                # make sure those target files are included in the final image too.
+                p = (root / fw).resolve().relative_to(root)
+                required.add(p)
+                unneeded.discard(p)
+                continue
+
             if fw in required:
                 continue
```
behrmann, line 254: Since it tests the same thing again, this should probably be folded in, so that order between the blocks matters less.
```suggestion
            if fw in required
                if (root / fw).is_symlink():
                    # modules may depend on a symbolic link to the actual firmware file, so
                    # make sure those target files are included in the final image too.
                    p = (root / fw).resolve().relative_to(root)
                    required.add(p)
                    unneeded.discard(p)
                continue
```

DaanDeMeyer (changes requested):
This seems unnecessarily complicated, why not loop over `required`, calls `resolve()` on each entry and add all resolved paths to `required` as well?

hundeboll:
> This seems unnecessarily complicated, why not loop over `required`, calls `resolve()` on each entry and add all resolved paths to `required` as well?

Okay, spent some time to get to know `gen_required_kernel_modules()`. Better?

hundeboll:
Also removed the unconditional include of modules / firmwares below an `updates` folder, as the include / exlcude logic already handles those two. I've updated the documentation instead.

mkosi/kmod.py
```diff
@@ -180,6 +180,11 @@ def gen_required_kernel_modules(
             mods = set(modulesd.rglob("*.ko*"))
         firmware = set()
 
+    links = {
+        p.resolve().relative_to(root)
```
DaanDeMeyer, line 184: This will resolve the symlink relative to the host root directory which could lead to very wrong and very hard to debug issues depending on the host system.

Please use `chase()` from mkosi.sandbox to resolve the symlinks properly.

mkosi/kmod.py
```diff
@@ -180,6 +180,11 @@ def gen_required_kernel_modules(
...
+        for p in filter(lambda fw: fw.is_symlink() and fw.exists(), map(lambda fw: root / fw, firmware))
```
DaanDeMeyer, line 185: Please use list comprehensions or a regular for loop, we don't use `filter()` or `map()` anywhere else, let's keep it that way for consistency.

mkosi/kmod.py
```diff
@@ -189,6 +194,7 @@ def gen_required_kernel_modules(
             },
             mods,
             firmware,
+            links,
```
DaanDeMeyer, line 197: We should make sure the parent directories of these exist as well so please modify the first entry of the chain as well to make the parent directories for all these as well.

mkosi/kmod.py
```diff
@@ -180,6 +180,11 @@ def gen_required_kernel_modules(
             mods = set(modulesd.rglob("*.ko*"))
         firmware = set()
 
+    links = {
```
DaanDeMeyer, line 183: This is a bad variable name, these aren't links, they're resolved links.

I don't think we need a new variable name for this, just adding the resolved symlinks to the `firmware` variable should be good enough.

This can also use a general comment explaining why we're doing this.

DaanDeMeyer (changes requested):

mkosi/kmod.py
```diff
@@ -180,6 +181,14 @@ def gen_required_kernel_modules(
             mods = set(modulesd.rglob("*.ko*"))
         firmware = set()
 
+    # Some firmware dependencies are symbolic links, so the targets for those must be included in the list
+    # of required firmware files too.
+    for fw in firmware.copy():
+        if (root / fw).is_symlink():
+            target = Path(chase(root, fw))
+            if target.exists():
+                firmware.add(target.relative_to(root))
```
DaanDeMeyer, line 190: Maybe extend the comment and mention that we make an implicit assumption here that we only have a single level of symlinks. If we have a symlink to a symlink this approach won't work and we'll end up with a dangling symlink in the initrd.

DaanDeMeyer (approved):

mkosi/kmod.py
```diff
@@ -180,6 +181,15 @@ def gen_required_kernel_modules(
             mods = set(modulesd.rglob("*.ko*"))
         firmware = set()
 
+    # Some firmware dependencies are symbolic links, so the targets for those must be included in the list
+    # of required firmware files too. Links are not resolved recursively, and so links pointing to links
```
DaanDeMeyer, line 185: "links are not resolved recursively" => "intermediate symlinks are not included, ..."

DaanDeMeyer (changes requested):

DaanDeMeyer (approved):

mkosi/resources/man/mkosi.1.md
```diff
@@ -1016,10 +1016,11 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 
 `KernelModulesInclude=`, `--kernel-modules-include=`
 :   Takes a list of regex patterns that specify kernel modules to include in the image. Patterns should be
-    relative to the `/usr/lib/modules/<kver>/kernel` directory. mkosi checks for a match anywhere in the module
-    path (e.g. `i915` will match against `drivers/gpu/drm/i915.ko`). All modules that match any of the
-    specified patterns are included in the image. All module and firmware dependencies of the matched modules
-    are included in the image as well.
+    relative to the `/usr/lib/modules/<kver>/kernel` directory (note that they match against modules in the
+    `updates` and `extra` sibling directories too). mkosi checks for a match anywhere in the module path (e.g.
```
behrmann, line 1020: I think this would be clearer:
```suggestion
    relative to the `/usr/lib/modules/<kver>` directory. mkosi checks for a match anywhere in the module path (e.g.
```
mentioning the `updates` and `extra` directories explicitly only obscures that, we look relative to that.

behrmann (approved):

mkosi/kmod.py
```diff
@@ -180,6 +181,15 @@ def gen_required_kernel_modules(
             mods = set(modulesd.rglob("*.ko*"))
         firmware = set()
 
+    # Some firmware dependencies are symbolic links, so the targets for those must be included in the list
+    # of required firmware files too. Intermediate symlinks are not included, and so links pointing to links
+    # results in dangling symlinks in the final image.
+    for fw in firmware.copy():
+        if (root / fw).is_symlink():
+            target = Path(chase(root, fw))
```
behrmann, line 189:
```suggestion
            target = Path(chase(os.fspath(root), os.fspath(fw)))
```
is needed here to appease mypy.

mkosi/kmod.py
```diff
@@ -180,6 +181,15 @@ def gen_required_kernel_modules(
             mods = set(modulesd.rglob("*.ko*"))
         firmware = set()
 
+    # Some firmware dependencies are symbolic links, so the targets for those must be included in the list
+    # of required firmware files too. Intermediate symlinks are not included, and so links pointing to links
+    # results in dangling symlinks in the final image.
+    for fw in firmware.copy():
+        if (root / fw).is_symlink():
+            target = Path(chase(os.fspath(root), os.fspath(fw)))
+            if target.exists():
```
DaanDeMeyer, line 190: @hundeboll Is this check required? I don't think it's expected for the firmware directory to have dangling symlinks.

hundeboll: Yes, `linux-firmware` on fedora rawhide has a dangling symlink which makes `cpio` fail later in the `mkosi` process:
```sh
# find /usr/lib/firmware/ -xtype l
/usr/lib/firmware/mrvl/sd8688_helper.bin.xz
```

DaanDeMeyer:
@hundeboll The ubuntu and centos initrd sizes need to be bumped by 1 in tests/test_initrd.py

mkosi/kmod.py
```diff
@@ -180,6 +181,15 @@ def gen_required_kernel_modules(
             mods = set(modulesd.rglob("*.ko*"))
         firmware = set()
 
+    # Some firmware dependencies are symbolic links, so the targets for those must be included in the list
+    # of required firmware files too. Intermediate symlinks are not included, and so links pointing to links
+    # results in dangling symlinks in the final image.
+    for fw in firmware.copy():
```
septatrix, line 187: Is the call to `copy` even required here?

behrmann: Otherwise this is mutating the set while it is being iterated over, which throws a `RuntimeError: Set changed size during iteration`.

PR #3314, opened 2025-01-03
Centos tools tree fixes

DaanDeMeyer:


behrmann (approved):

mkosi.conf.d/40-tools/mkosi.conf
behrmann, line 1: Keeping that otherwise empty file around shouldn't be necessary or is it? `40-tools` being a dir and `mkosi.conf.d` existing should be enough.

PR #3323, opened 2025-01-04
Remove preference for dnf4 in mkosi-initrd

septatrix:
This effectively reverts 7b929049 as DNF 5 is now the new default in the most recent Fedora release.

### DIFF
--- a/mkosi/initrd.py
+++ b/mkosi/initrd.py
@@ -200,13 +200,14 @@ def main() -> None:
         if Path("/etc/kernel/cmdline").exists():
             cmdline += ["--kernel-command-line", Path("/etc/kernel/cmdline").read_text()]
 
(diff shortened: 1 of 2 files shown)


DaanDeMeyer:
We should check whether dnf symlinks to dnf5 or not and base the decision based on that.

septatrix:
In that case couldn't we just set it as `MKOSI_DNF=dnf`. Then the symlink will simply refer us to the desired one? Or are there systems where `dnf` does not exist and only `dnf5` is present?

DaanDeMeyer:
@septatrix If we set `MKOSI_DNF=dnf` then all the dnf4 specific logic in mkosi will be used, even if `dnf` points to `dnf5`, so this would need more work in mkosi.

DaanDeMeyer (approved):

PR #3336, opened 2025-01-08
Add option to save the BuildSourcesEphemeral overlay.

behrmann:
This systematises a trick Daan showed me a while back and that I've been using most everywhere since. Having it in mkosi itself makes it easier to use for everybody.

### DIFF
--- a/mkosi/mounts.py
+++ b/mkosi/mounts.py
@@ -6,9 +6,10 @@ import stat
 import tempfile
 from collections.abc import Iterator, Sequence
 from pathlib import Path
-from typing import Optional
+from typing import Optional, Union
 
-from mkosi.config import Config
+from mkosi.config import BuildSourcesEphemeral, Config
+from mkosi.log import die
 from mkosi.sandbox import OverlayOperation
 from mkosi.util import PathString, flatten
 
@@ -56,7 +57,11 @@ def mount_overlay(
 
 
 @contextlib.contextmanager
-def finalize_source_mounts(config: Config, *, ephemeral: bool) -> Iterator[list[PathString]]:
+def finalize_source_mounts(
+    config: Config,
+    *,
+    ephemeral: Union[BuildSourcesEphemeral, bool],
+) -> Iterator[list[PathString]]:
     with contextlib.ExitStack() as stack:
         options: list[PathString] = []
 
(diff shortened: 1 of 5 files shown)


DaanDeMeyer (commented):
So I'm not too sure about adding official support for my hack, how about we document it somewhere instead?

behrmann:
I'm also happy to call the option `voidwarranty-buildcache` instead.

behrmann:
Rebased and dropped the dont merge label, since it worked locally. I cherry-picked 62582f3d0d545c41264dfeed3f5880580b8dc29a from #3402 for now to make CI run.

DaanDeMeyer (changes requested):
Shouldn't we limit this to only when running build scripts and do regular ephemeral for the other scripts?

behrmann:
Good catch, I thought it was there, that's why it's mentioned in the man page. I must have lost this in a rebase.

mkosi/mounts.py
```diff
@@ -56,20 +56,30 @@ def mount_overlay(
...
     with contextlib.ExitStack() as stack:
         options: list[PathString] = []
 
         for t in config.build_sources:
             src, dst = t.with_prefix("/work/src")
 
             if ephemeral:
-                upperdir = Path(stack.enter_context(tempfile.TemporaryDirectory(prefix="volatile-overlay")))
-                os.chmod(upperdir, src.stat().st_mode)
+                if ephemeral == BuildSourcesEphemeral.buildcache and config.build_dir is not None:
```
DaanDeMeyer, line 71: I would fail if this is specified but there's no build directory configured

DaanDeMeyer (changes requested):

mkosi/mounts.py
```diff
@@ -56,20 +57,36 @@ def mount_overlay(
...
             if ephemeral:
-                upperdir = Path(stack.enter_context(tempfile.TemporaryDirectory(prefix="volatile-overlay")))
-                os.chmod(upperdir, src.stat().st_mode)
+                if ephemeral == BuildSourcesEphemeral.buildcache:
+                    if config.build_dir is None:
+                        die(
+                            "BuildSourcesEphemeral=buildcache was configured, but no build directory exists.",  # noqa: E501
+                            hint="Configure BuildDirectory= or create mkosi.builddir.",
+                        )
+                    assert config.build_dir
```
DaanDeMeyer, line 78: This shouldn't be needed anymore no?

DaanDeMeyer (approved):

PR #3339, opened 2025-01-09
Add missing continue if pandoc is not available

behrmann:
Without the continue, the run on the next line will call die if pandoc is not available, thus breaking the fallback for subsequent formats.


DaanDeMeyer:
If pandoc is the only format we'll fail silently if we add this

behrmann:
Running `mkosi --doc-format pandoc documentation` still prints the warning and since this will only ever be used interactively, I think this is fine.

PR #3344, opened 2025-01-11
Drop support for Focal

DaanDeMeyer:
It's going EOL in April, so I think it's fine if we drop support for it.

### DIFF
--- a/mkosi/distributions/ubuntu.py
+++ b/mkosi/distributions/ubuntu.py
@@ -25,11 +25,8 @@ class Installer(debian.Installer):
     def repositories(cls, context: Context, local: bool = True) -> Iterable[AptRepository]:
         types = ("deb", "deb-src")
 
-        # From kinetic onwards, the usr-is-merged package is available in universe and is required by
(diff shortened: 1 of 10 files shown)


mkosi/distributions/ubuntu.py
```diff
@@ -25,11 +25,8 @@ def default_tools_tree_distribution(cls) -> Distribution:
     def repositories(cls, context: Context, local: bool = True) -> Iterable[AptRepository]:
         types = ("deb", "deb-src")
 
-        # From kinetic onwards, the usr-is-merged package is available in universe and is required by
-        # mkosi to set up a proper usr-merged system so we add the universe repository unconditionally.
         components = (
             "main",
-            *(["universe"] if context.config.release not in ("focal", "jammy") else []),
```
behrmann, line 32: Why is the `jammy` thing being dropped here, too?

behrmann (approved):
LGTM, except for one question

mkosi/distributions/ubuntu.py
DaanDeMeyer: Not required anymore, `usr-is-merged` was moved to the `main` repository

PR #3346, opened 2025-01-12
Introduce support for overriding system's DeviceTree

### LINKED ISSUE #2439: Allow specifying a devicetree binary that is passed to ukify

andersson:
While experimenting with OS images on development boards where either the bootloader doesn't load a Flattened DeviceTree Blob, or if one wants to replace the one provided by the bootloader with a specific/new one for development/experimentation/testing purposes, it's convenient to use the OS-loaders ability to load a specific DeviceTree blob.

Add a new option to mkosi, to cause a specified DeviceTree blob to be baked into the UKI, or copied into /boot and added to the systemd loader entry.

As different distributions store the dtb files in different locations the requested file is searched for in the locations spotted in Debian, Fedora, Arch Linux packages (both from distro and kernel build system - as these differs as well).

Resolves #2439
Based on initial effort by Manuel Traut.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1460,6 +1460,21 @@ def build_kernel_modules_initrd(context: Context, kver: str) -> Path:
     return kmods
 
 
+def find_devicetree(context: Context, kver: str) -> Path:
+    assert context.config.devicetree
+
+    for d in (
+        context.root / f"usr/lib/firmware/{kver}/device-tree",
+        context.root / f"usr/lib/linux-image-{kver}",
+        context.root / f"usr/lib/modules/{kver}/dtb",
+    ):
+        dtb = d / context.config.devicetree
+        if dtb.exists():
+            return dtb
+
+    die(f"Requested devicetree {context.config.devicetree} not found")
+
+
 def join_initrds(initrds: Sequence[Path], output: Path) -> Path:
     assert initrds
 
@@ -1608,6 +1623,11 @@ def build_uki(
         "--ro-bind", kimg, workdir(kimg),
         *flatten(["--ro-bind", os.fspath(profile), os.fspath(workdir(profile))] for profile in profiles),
     ]  # fmt: skip
+
+    if context.config.devicetree:
+        dtb = find_devicetree(context, kver)
+        arguments += ["--devicetree", workdir(dtb)]
+        options += ["--ro-bind", dtb, workdir(dtb)]
 
     if context.config.secure_boot:
         assert context.config.secure_boot_key
@@ -1814,6 +1834,12 @@ def install_type1(
 
     kmods = build_kernel_modules_initrd(context, kver)
 
+    dtb = None
+    if context.config.devicetree:
+        dtb = dst / context.config.devicetree
+        with umask(~0o700):
+            dtb.parent.mkdir(parents=True, exist_ok=True)
+
     with umask(~0o600):
         if (
             want_efi(context.config)
@@ -1831,6 +1857,9 @@ def install_type1(
         ]
         initrds += [Path(shutil.copy2(kmods, dst / "kernel-modules.initrd"))]
 
+        if dtb:
+            shutil.copy2(find_devicetree(context, kver), dtb)
+
         with entry.open("w") as f:
             f.write(
                 textwrap.dedent(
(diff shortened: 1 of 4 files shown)


mkosi/config.py
```diff
@@ -4816,6 +4823,7 @@ def summary(config: Config) -> str:
                             Initrds: {line_join_list(config.initrds)}
                     Initrd Packages: {line_join_list(config.initrd_packages)}
            Initrd Volatile Packages: {line_join_list(config.initrd_volatile_packages)}
+                         Devicetree: {config.devicetree}
```
behrmann, line 4826:
```suggestion
                         Devicetree: {none_to_none(config.devicetree)}
```

mkosi/__init__.py
```diff
@@ -1460,6 +1460,21 @@ def build_kernel_modules_initrd(context: Context, kver: str) -> Path:
     return kmods
 
 
+def find_requested_devicetree(context: Context, kver: str) -> Path:
+    assert context.config.devicetree
+
+    for f in (
+        context.root / f"lib/firmware/{kver}/device-tree",
```
behrmann, line 1467:
```suggestion
        context.root / f"usr/lib/firmware/{kver}/device-tree",
```

mkosi/__init__.py
```diff
@@ -1460,6 +1460,21 @@ def build_kernel_modules_initrd(context: Context, kver: str) -> Path:
     return kmods
 
 
+def find_requested_devicetree(context: Context, kver: str) -> Path:
+    assert context.config.devicetree
+
+    for f in (
```
DaanDeMeyer, line 1466: Let's call this `d` or `p`, since it's a path/directory, but definitely not a file

mkosi/__init__.py
```diff
@@ -1460,6 +1460,21 @@ def build_kernel_modules_initrd(context: Context, kver: str) -> Path:
     return kmods
 
 
+def find_requested_devicetree(context: Context, kver: str) -> Path:
```
DaanDeMeyer, line 1463: Let's just name this `find_devicetree()`

mkosi/__init__.py
```diff
@@ -1814,6 +1834,14 @@ def install_type1(
 
     kmods = build_kernel_modules_initrd(context, kver)
 
+    dtb = None
+    if context.config.devicetree:
+        dtb_src = find_requested_devicetree(context, kver)
+
+        dtb = dst / context.config.devicetree
+        dtb.parent.mkdir(parents=True, exist_ok=True)
```
DaanDeMeyer, line 1842: This can be dropped, we already create the parent directory earlier in this function

mkosi/__init__.py
```diff
@@ -1814,6 +1834,14 @@ def install_type1(
 
     kmods = build_kernel_modules_initrd(context, kver)
 
+    dtb = None
+    if context.config.devicetree:
+        dtb_src = find_requested_devicetree(context, kver)
```
DaanDeMeyer, line 1839: This can be inlined into the `shutil.copy2()` call

DaanDeMeyer (changes requested):

mkosi/__init__.py
andersson: We create "dst" previously, but context.config.devicetree typically contains a vendor path in it (such as qcom/sc8280xp-crd.dtb, someone need to create the "qcom" part)

But looking at the code again, I think the mkdir and the copy should be done with umask...

DaanDeMeyer: Ahah, in that case, like you already said, please do it with umask

DaanDeMeyer:
@andersson You resolved some conversations but didn't push, maybe you forgot?

andersson:
@DaanDeMeyer no I didn't push, I was waiting for your confirmation on the install_type1() rework. I marked them as resolved as they where trivial and I wanted to de-clutter the PR view. Perhaps I'm using Github wrong?

I've updated and pushed a new version now. Thank you.

DaanDeMeyer (changes requested):
Missing documentation in man page

DaanDeMeyer (approved):

mkosi/__init__.py
```diff
@@ -1814,6 +1834,12 @@ def install_type1(
 
     kmods = build_kernel_modules_initrd(context, kver)
 
+    dtb = None
+    if context.config.devicetree:
+        dtb = dst / context.config.devicetree
+        with umask(~0o700):
```
behrmann, line 1840: Is there a reason for this strict umask? Can a devicetree file be sensitive or is this required by whatever is loading the devicetree file?

behrmann (approved):

mkosi/__init__.py
andersson: I picked the same umask as for dst (the parent) and systemd-boot loader entries directories. There's nothing sensitive and all these things typically ends up in a vfat partition.

PR #3354, opened 2025-01-14
Bind mount /dev/fuse in sandbox

hundeboll:
/dev/fuse is needed to support building container images from within the mkosi.build scripts. Make it avaiable in the sandbox just like /dev/null and its friends.

Since not all mkosi builds runs in host contexts where /dev/fuse is actually available (i.e. in a container), update the logic to skip /dev nodes if they are not present.

### DIFF
--- a/mkosi/sandbox.py
+++ b/mkosi/sandbox.py
@@ -524,11 +524,15 @@ class DevOperation(FSOperation):
         # /dev/null fails with EACCESS for unknown reasons.
         mount("tmpfs", dst, "tmpfs", 0, "mode=0755")
 
-        for node in ("null", "zero", "full", "random", "urandom", "tty"):
+        for node in ("null", "zero", "full", "random", "urandom", "tty", "fuse"):
+            nsrc = joinpath(oldroot, "dev", node)
+            if not os.path.exists(nsrc) and node == "fuse":
+                continue
+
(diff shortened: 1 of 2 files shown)


mkosi/sandbox.py
```diff
@@ -524,11 +524,15 @@ def execute(self, oldroot: str, newroot: str) -> None:
...
+        for node in ("null", "zero", "full", "random", "urandom", "tty", "fuse"):
+            nsrc = joinpath(oldroot, "dev", node)
+            if not os.path.exists(nsrc):
+                continue
+
             ndst = joinpath(dst, node)
             os.close(os.open(ndst, os.O_CREAT | os.O_CLOEXEC | os.O_EXCL))
 
-            mount(joinpath(oldroot, "dev", node), ndst, "", MS_BIND, "")
+            mount(nsrc, ndst, "", MS_BIND, "")
```
DaanDeMeyer, line 535: This means we silently ignore all missing device nodes instead of just /dev/fuse. We should fail loudly if any of the other device nodes are not available, so please rework this so that only /dev/fuse is skipped if it is not available.

DaanDeMeyer (changes requested):

DaanDeMeyer (approved):

behrmann (approved):

PR #3359, opened 2025-01-14
Make sure we use the same host distribution/release in sandbox

DaanDeMeyer:
If the distribution is not explicitly specified in the config file and the host distribution does not match the tools tree distribution, then currently mkosi -f sandbox mkosi -f and mkosi -f will build images for different distributions.

Let's make sure these commands use the same default distribution by communicating the host distribution and release to the sandbox via environment variables and using them to determine the default values to use if available.


mkosi/config.py
```diff
@@ -861,7 +861,10 @@ def config_default_output(namespace: argparse.Namespace) -> str:
 
 
 def config_default_distribution(namespace: argparse.Namespace) -> Distribution:
-    detected = detect_distribution()[0]
+    if d := os.getenv("MKOSI_HOST_DISTRIBUTION"):
+        detected = Distribution(d)
+    else:
+        detected = detect_distribution()[0]
```
behrmann, line 867: This sidesteps this here neatly
```suggestion
    if d := os.getenv("MKOSI_HOST_DISTRIBUTION"):
        return Distribution(d)
    
    detected = detect_distribution()[0]
```

mkosi/config.py
```diff
@@ -874,8 +877,12 @@ def config_default_distribution(namespace: argparse.Namespace) -> Distribution:
 
 
 def config_default_release(namespace: argparse.Namespace) -> str:
+    if (d := os.getenv("MKOSI_HOST_DISTRIBUTION")) and (r := os.getenv("MKOSI_HOST_RELEASE")):
+        hd, hr = Distribution(d), r
+    else:
+        hd, hr = detect_distribution()
```
behrmann, line 883: This is probably most easily squashed by
```suggestion
    hd: Optional[Distribution]
    hr: Optional[str]
    if (d := os.getenv("MKOSI_HOST_DISTRIBUTION")) and (r := os.getenv("MKOSI_HOST_RELEASE")):
        hd, hr = Distribution(d), r
    else:
        hd, hr = detect_distribution()
```

behrmann (approved):

PR #3362, opened 2025-01-15
mkosi-initrd: Two fixes

DaanDeMeyer:


behrmann (approved):

DaanDeMeyer:
cc @aafeijoo-suse for visibility though I don't think the credential stuff will affect you since it's only for compat with older systemd versions

aafeijoo-suse:
> cc @aafeijoo-suse for visibility though I don't think the credential stuff will affect you since it's only for compat with older systemd versions

Ok from my side.

DaanDeMeyer:
Guess we still need both these files, closing

PR #3364, opened 2025-01-15
Add missing parentheses to make sure we don't throw away scope for virtiofsd

DaanDeMeyer:


mkosi/qemu.py
```diff
@@ -401,7 +401,7 @@ def start_virtiofsd(
                     "--bind", directory, workdir(directory),
                     *(["--become-root"] if uidmap else []),
                 ],
-                setup=scope + become_root_in_subuid_range_cmd() if scope and not uidmap else [],
+                setup=scope + (become_root_in_subuid_range_cmd() if scope and not uidmap else []),
```
behrmann, line 404: Yikes, that's easy to miss. Let's put it on two lines to make the intention very clear
```suggestion
                setup=(
                   scope + 
                   (become_root_in_subuid_range_cmd() if scope and not uidmap else [])
                ),
```

behrmann (approved):

PR #3366, opened 2025-01-15
Add a keyring cache and use it for Arch Linux

DaanDeMeyer:
Currently, the Arch keyring from the host or tools tree is always
used directly without any option to modify it (except by modifying
it directly on the host or in the tools tree). We also don't support
RepositoryKeyFetch= for the Arch Linux keyring and it's one of the
sources of annoying mounts outside of /usr which complicates
"mkosi sandbox".

Let's improve the situation by switching to our own pacman keyring
instead of using the one from the tools tree or host. Instead, we'll
only use /usr/share/pacman/keyrings from the host or tools tree to
populate the keyring we maintain ourselves. Users can extend the
keyring created by mkosi with their own keys via sandbox trees. If
RepositoryKeyFetch= is enabled, we'll download the archlinux-keyring
package and extract the keyring from there into the sandbox.

The keyring cache is maintained together with the repository metadata
cache. The only difference is that sync_repository_metadata() will
update the global package cache whereas the keyring directory is always
maintained per cache directory instead of globally.


bluca (approved):

mkosi/config.py
```diff
@@ -905,23 +905,22 @@ def config_default_repository_key_fetch(namespace: argparse.Namespace) -> bool:
     if detect_distribution()[0] != Distribution.ubuntu:
         return False
 
+    def needs_repository_key_fetch(distribution: Distribution) -> bool:
+        return distribution == Distribution.arch or distribution.is_rpm_distribution()
```
behrmann, line 909: Let's move this to the beginning of the function, having function definitions in the middle always strikes me a bit unexpected.

mkosi/config.py
```diff
@@ -905,23 +905,22 @@ def config_default_repository_key_fetch(namespace: argparse.Namespace) -> bool:
...
-        return cast(bool, namespace.distribution.is_rpm_distribution())
+        return needs_repository_key_fetch(namespace.distribution)
 
     if namespace.tools_tree != Path("default"):
         return (
             detect_distribution(namespace.tools_tree)[0] == Distribution.ubuntu
-            and namespace.distribution.is_rpm_distribution()
-        )
+            and needs_repository_key_fetch(namespace.distribution)
+        )  # fmt: skip
```
behrmann, line 918: Wow the way it reformats this is insane, that can only be a bug in ruff. So it can be referenced in a bug report (ruff version 0.9.1):
```diff
         return needs_repository_key_fetch(namespace.distribution)
 
     if namespace.tools_tree != Path("default"):
-        return (
-            detect_distribution(namespace.tools_tree)[0] == Distribution.ubuntu
-            and needs_repository_key_fetch(namespace.distribution)
-        ) 
+        return detect_distribution(namespace.tools_tree)[
+            0
+        ] == Distribution.ubuntu and needs_repository_key_fetch(namespace.distribution)
 
     return (
         namespace.tools_tree_distribution == Distribution.ubuntu
```

DaanDeMeyer:
Just moved a function around so let's not wait for CI again

mkosi/config.py
```diff
@@ -902,26 +902,25 @@ def config_default_tools_tree_distribution(namespace: argparse.Namespace) -> Dis
...
```
behrmann, line 918: The bug report is https://github.com/astral-sh/ruff/issues/15527

PR #3377, opened 2025-01-19
Add scripts and config to support signing UKIs in OBS

bluca:
https://build.opensuse.org/package/show/home:bluca:branches:home:bluca:systemd/debian-image


DaanDeMeyer (changes requested):
Making the output directory available to build scripts is not an option, because we don't clean up the output directory when Format=none which causes a big mess if we make it available to build scripts.

Please use either the build directory or the source directories via `BuildSources=` to get stuff out of the build script.

bluca:
> Making the output directory available to build scripts is not an option, because we don't clean up the output directory when Format=none which causes a big mess if we make it available to build scripts.
> 
> Please use either the build directory or the source directories via `BuildSources=` to get stuff out of the build script.

But how? Everything in the build directory is lost after the build

DaanDeMeyer:
> But how? Everything in the build directory is lost after the build

Huh? Not at all, `$BUILDDIR` is persisted across builds. Just set `BuildDirectory=<some-path>` and it'll be created and persisted across builds. The annoying thing about `BuildDirectory=` for this use case is that we create a subdirectory in it, so you'll probably want to use `BuildSources=` instead.

bluca:
```
The following files have no copyright information:
* mkosi/resources/mkosi-obs/mkosi.build
* mkosi/resources/mkosi-obs/mkosi.postoutput
```

wat

bluca:
> > But how? Everything in the build directory is lost after the build
> 
> Huh? Not at all, `$BUILDDIR` is persisted across builds. Just set `BuildDirectory=<some-path>` and it'll be created and persisted across builds. The annoying thing about `BuildDirectory=` for this use case is that we create a subdirectory in it, so you'll probably want to use `BuildSources=` instead.

But I mean that whatever is in there is not then moved to the output directory, so the build finishes and there's nothing, it's all lost

DaanDeMeyer:
SPDX

bluca:
> SPDX

It's there

```
#!/bin/bash
```

DaanDeMeyer:
> > > But how? Everything in the build directory is lost after the build
> > 
> > 
> > Huh? Not at all, `$BUILDDIR` is persisted across builds. Just set `BuildDirectory=<some-path>` and it'll be created and persisted across builds. The annoying thing about `BuildDirectory=` for this use case is that we create a subdirectory in it, so you'll probably want to use `BuildSources=` instead.
> 
> But I mean that whatever is in there is not then moved to the output directory, so the build finishes and there's nothing, it's all lost

I mean that you make the output directory available to the build script via `BuildSources=`

mkosi/resources/mkosi-obs/mkosi.build
```diff
@@ -0,0 +1,107 @@
...
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Second stage of build:
+# - signed hashes are in cpio archive in SOURCES/ together with artifacts from previous build
+# - attach the signatures to the UKI(s) with pesign
+# - remove shasums of previous artifacts, given we are re-creating them after this step
+# - place artifacts from previous builds and signed UKI in output directory
+
+set -e
+set -x
```
behrmann, line 11: Nit:
```suggestion
set -ex
```

mkosi/resources/mkosi-obs/mkosi.build
```diff
@@ -0,0 +1,107 @@
...
+
+if [ ! -f /usr/src/packages/SOURCES/mkosi.cpio.rsasign.sig ]; then
+    exit 0
+fi
+
+cpio -t < /usr/src/packages/SOURCES/mkosi.cpio.rsasign.sig
```
behrmann, line 17: Nit here and elsewhere to be consistent with the style in the systemd repo
```suggestion
cpio -t </usr/src/packages/SOURCES/mkosi.cpio.rsasign.sig
```
Same for output redirection `>foo` instead of `> foo`.

mkosi/resources/mkosi-obs/mkosi.build
```diff
@@ -0,0 +1,107 @@
...
+nss_db="$PWD/nss-db"
+certutil -N -d sql:"$nss_db" --empty-password
+certutil -A -d sql:"$nss_db" -n cert -t CT,CT,CT -i /usr/src/packages/SOURCES/_projectcert.crt
+
+cp -r /usr/src/packages/SOURCES/"$IMAGE_ID"* "$OUTPUTDIR"
+rm -f "$OUTPUTDIR/mkosi.cpio.rsasign*" "$OUTPUTDIR"/*.sha*
+
+while read -r SIG
+do
+    export infile="${SIG%.sig}"
```
behrmann, line 39: Why is that exported it only seems to be used for interpolation.

mkosi/resources/mkosi-obs/mkosi.build
```diff
@@ -0,0 +1,107 @@
...
+    cpio -i --to-stdout "${infile}" < /usr/src/packages/SOURCES/*.cpio.rsasign > "$(basename "${infile}").sattrs"
+    test -s "$(basename "${infile}").sattrs"
+
+    DEST="$OUTPUTDIR/${SIG#*/}"
+    DEST="${DEST%%.sig}"
+
+    # ensure the EFI hash matches before and after attaching the signature
+    old_hash=$(pesign -n sql:"$nss_db" -h -P -i "/usr/src/packages/SOURCES/$infile" | cut -d ' ' -f1)
```
behrmann, line 47: Nit: choose your poison
```suggestion
    old_hash=$(pesign -n sql:"$nss_db" -h -P -i "/usr/src/packages/SOURCES/$infile" | cut -d' ' -f1)
```
or
```suggestion
    old_hash=$(pesign -n sql:"$nss_db" -h -P -i "/usr/src/packages/SOURCES/$infile" | cut -d ' ' -f 1)
```
same below

mkosi/resources/mkosi-obs/mkosi.build
```diff
@@ -0,0 +1,107 @@
...
+        exit 1
+    fi
+
+    rm -f "$(basename "${infile}").sattrs" "$SIG" "$infile"
+done < <(find hashes/ukis -type f \( -name '*efi.sig' -o -name 'vmlinu*.sig' \) -printf '%P\n')
+rm -rf nss-db "$OUTPUTDIR"/*.sig hashes/ukis
+
+while read -r SIG
+do
+    uki="$OUTPUTDIR/$(basename $(dirname ${SIG%.sig}))"
```
behrmann, line 64:
```suggestion
    uki="$OUTPUTDIR/$(basename "$(dirname "${SIG%.sig}")")"
```

mkosi/resources/mkosi-obs/mkosi.build
```diff
@@ -0,0 +1,107 @@
...
+    pcrs="${uki%.efi}.pcrs"
+    pol="$(basename ${SIG%.sig})"
```
behrmann, line 66:
```suggestion
    pol="$(basename "${SIG%.sig}")"
```

mkosi/resources/mkosi-obs/mkosi.build
```diff
@@ -0,0 +1,107 @@
...
+
+while read -r PCRS
+do
+    uki="${PCRS%.pcrs}.efi"
+    # Attach the content of the PCRS file to the uki with objcopy in a .pcrsig section
+    objcopy --add-section .pcrsig="$PCRS" "$uki" "$uki"
+    mkdir -p hashes/ukis
+    pesign --force -n sql:"$nss_db" -i "$uki" -E "hashes/ukis/$(basename "$uki")"
+done < <(find "$OUTPUTDIR" -type f -name '*.pcrs')
+rm -f "$OUTPUTDIR"/*.pcrs**
```
behrmann, line 98: Why the double star here? Does it do anything different without `shopt -s globstar`?

mkosi/resources/mkosi-obs/mkosi.postoutput
```diff
@@ -0,0 +1,54 @@
...
+UNSIGNED=( "$(find "$OUTPUTDIR" -type f \( -name "*.efi" -o -name "vmlinu*" \) -printf '%P\n')" )
+
+for f in "${UNSIGNED[@]}"
+do
+    if ukify --json=short inspect "${OUTPUTDIR}/${f}" | jq -e 'has(".linux")' > /dev/null && \
+        ! ukify --json=short inspect "${OUTPUTDIR}/${f}" | jq -e 'has(".pcrsig")' > /dev/null; then
+        mkdir -p "hashes/pcrs/$f"
+        cat "${OUTPUTDIR}/${f%.efi}.pcrs" | jq -r 'to_entries[] | .value[].pol' | while read -r pol; do
+            echo -n "$pol" | base64 --decode > "hashes/pcrs/${f}/${pol}"
+        done
```
behrmann, line 29: Maybe
```suggestion
        while read -r pol; do
            echo -n "$pol" | base64 --decode > "hashes/pcrs/${f}/${pol}"
        done < <(jq -r 'to_entries[] | .value[].pol' "${OUTPUTDIR}/${f%.efi}.pcrs")
```
or something like
```suggestion
        <"${OUTPUTDIR}/${f%.efi}.pcrs" jq -r 'to_entries[] | .value[].pol' | 
            while read -r pol; do
                echo -n "$pol" | base64 --decode > "hashes/pcrs/${f}/${pol}"
            done
```
maybe also without the input redirection

mkosi/resources/mkosi-obs/mkosi.postoutput
```diff
@@ -0,0 +1,54 @@
...
+# certutil will fail if it's called twice
+rm -rf "$nss_db"
+mkdir -p "$nss_db" hashes
+certutil -N -d sql:"$nss_db" --empty-password
+
+declare -a UNSIGNED
+UNSIGNED=( "$(find "$OUTPUTDIR" -type f \( -name "*.efi" -o -name "vmlinu*" \) -printf '%P\n')" )
+
+for f in "${UNSIGNED[@]}"
+do
```
behrmann, line 23: I greatly prefer this style, but for consistency with the rest
```suggestion
for f in "${UNSIGNED[@]}"; do
```

mkosi/__init__.py
```diff
@@ -1703,6 +1706,38 @@ def build_uki(
...
+                    "x509",
+                    "-pubkey",
+                    "-noout",
+                    "-in",
+                    context.config.sign_expected_pcr_certificate,
+                ],
+                stdout=public_key.open("wb"),
+            )
+            arguments += ["--pcr-public-key", workdir(public_key)]
+            options += ["--ro-bind", public_key, workdir(public_key)]
```
DaanDeMeyer, line 1733: Use `systemd-keyutil public` please, and pass in all the proper arguments so it works with providers as well

bluca:
How do I get past this false positive?

```
The following files have no copyright information:
* mkosi/resources/mkosi-obs/mkosi.build
* mkosi/resources/mkosi-obs/mkosi.postoutput
* ```

File: mkosi/resources/mkosi-obs/mkosi.postoutput [hunk 11]
```diff
@@ -0,0 +1,58 @@
...
+rm -rf "$nss_db"
+mkdir -p "$nss_db" hashes
+certutil -N -d sql:"$nss_db" --empty-password
+
+for f in "${UNSIGNED[@]}"; do
+    if [ -f "${OUTPUTDIR}/${f%.efi}.pcrs" ]; then
+        mkdir -p "hashes/pcrs/$f"
+        while read -r pol; do
+            echo -n "$pol" | tr '[:lower:]' '[:upper:]' | basenc --base16 --decode >"hashes/pcrs/${f}/${pol}"
+        done < <(jq -r 'to_entries[] | .value[].pol' <"${OUTPUTDIR}/${f%.efi}.pcrs")
```
behrmann, line 31: haven't checked this with actual data yet, but maybe the whole thing could be
```suggestion
        jq -r 'to_entries[] | .value[].pol | ascii_downcase' "${OUTPUTDIR}/${f%.efi}.pcrs" | 
            basenc --base16 --decode >"hashes/pcrs/${f}/${pol}"
```

mkosi/resources/mkosi-obs/mkosi.postoutput
```diff
@@ -0,0 +1,58 @@
...
+pushd hashes
+find . -type f | cpio -H newc -o >"$OUTPUTDIR/hashes.cpio.rsasign"
+popd
+rm -rf hashes "$nss_db"
+
+echo "Staging the following files for signing:"
+cpio -t <"$OUTPUTDIR/hashes.cpio.rsasign"
+
+# The second stage will not do a full rebuild, but only attach signatures to the existing UKI
+cat <<EOF >"$OUTPUTDIR/mkosi.conf"
```
behrmann, line 48: Nit:
```suggestion
cat >"$OUTPUTDIR/mkosi.conf" <<EOF 
```
It does make it a bit clearer, that the output redirection is not part of the heredoc.

mkosi/resources/mkosi-obs/mkosi.postoutput
bluca: $pol is not defined in that case

mkosi/__init__.py
```diff
@@ -1644,7 +1647,7 @@ def build_uki(
     cmdline: Sequence[str],
     profiles: Sequence[Path],
     output: Path,
-) -> None:
+) -> Union[str, None]:
```
DaanDeMeyer, line 1650: `Optional[str]`

mkosi/__init__.py
```diff
@@ -1619,20 +1619,23 @@ def run_ukify(
...
 
-    run(
+    result = run(
         cmd,
         stdin=(
             sys.stdin
             if context.config.secure_boot_key_source.type != KeySourceType.file
             else subprocess.DEVNULL
         ),
+        stdout=subprocess.PIPE,
```
DaanDeMeyer, line 1629: This should be optional and only done if we need the json output, so add a `json` argument to `run_ukify()` that translates to `"--json=short",` and only make stdout a pipe if that is passed.

You can then make the return type of the function `dict[str, Any]`, parse the json in `run_ukify()` itself if requested and return an empty dict if no json output was requested.

mkosi/__init__.py
```diff
@@ -1727,6 +1730,42 @@ def build_uki(
...
+            run(keyutil_arguments, stdout=public_key.open("wb"))
+            arguments += ["--pcr-public-key", workdir(public_key)]
+            options += ["--ro-bind", public_key, workdir(public_key)]
+        else:
+            arguments += [
+                "--pcr-public-key", workdir(context.config.sign_expected_pcr_certificate),
+            ]  # fmt: skip
+            options += [
+                "--ro-bind", context.config.sign_expected_pcr_certificate, workdir(context.config.sign_expected_pcr_certificate),  # noqa: E501
+            ]  # fmt: skip
```
DaanDeMeyer, line 1768: I don't see the point of having to do this inside of `mkosi`, `ukify` calls `systemd-keyutil` itself to extract the public key, why do it again here

mkosi/__init__.py
```diff
@@ -2751,6 +2793,15 @@ def check_tools(config: Config, verb: Verb) -> None:
             check_tool(config, "sbsiglist", reason="set up systemd-boot secure boot auto-enrollment")
             check_tool(config, "sbvarsign", reason="set up systemd-boot secure boot auto-enrollment")
 
+        if ArtifactOutput.pcrs in config.split_artifacts:
+            check_systemd_tool(
+                config,
+                "systemd-measure",
+                "/usr/lib/systemd/systemd-measure",
+                version="257.999",
```
DaanDeMeyer, line 2801: Let's not bother with the OBS versions in mkosi, just make this 257

DaanDeMeyer (changes requested):

mkosi/__init__.py
bluca: But 257 is not right? It will be added in 258, but if I make it 258, then it won't work with packages as the version is 257.999

mkosi/__init__.py
DaanDeMeyer: Sure but systemd-measure is invoked by ukify? What I'm saying is that ukify already knows how to invoke systemd-keyutil to get the public key. If it doesn't do that yet for --policy-digest, we should fix ukify to do it for that codepath as well.

mkosi/__init__.py
bluca: How do I solve this one:

```
mkosi/__init__.py:1645:9: error: Returning Any from function declared to return
"dict[str, Any]"  [no-any-return]
            return json.loads(result.stdout.strip())
```

DaanDeMeyer: Just use `cast()` to cast it to the right type.

mkosi/__init__.py
```diff
@@ -1751,7 +1802,9 @@ def build_uki(
         options += ["--ro-bind", initrd, workdir(initrd)]
 
     with complete_step(f"Generating unified kernel image for kernel version {kver}"):
-        run_ukify(context, stub, output, cmdline=cmdline, arguments=arguments, options=options)
+        return run_ukify(
+            context, stub, output, cmdline=cmdline, arguments=arguments, options=options, json_out=json_out
+        )
```
behrmann, line 1807: Nit:
```suggestion
        return run_ukify(
            context, 
            stub, 
            output, 
            cmdline=cmdline, 
            arguments=arguments, 
            options=options, 
            json_out=json_out,
        )
```

bluca: tried that first, but the linter rejects it and wants it this way

behrmann: Did you have the trailing command after `json_out`? Without it `ruff format` will make it a single line, with it, it comes out like in the suggestion.

mkosi/__init__.py
bluca: I've added a TODO comment to bump this to 258 once it's released and available

mkosi/__init__.py
bluca: that was probably it

mkosi/__init__.py
bluca: ok, stampity then https://github.com/systemd/systemd/pull/36321

bluca:
ukify change is merged, is this good to go?

DaanDeMeyer (approved):

PR #3386, opened 2025-01-19
addon: Don't add initrd section if buildroot is empty

DaanDeMeyer:


mkosi/__init__.py
```diff
@@ -193,6 +193,9 @@ def remove_files(context: Context) -> None:
     if context.config.output_format.is_extension_image():
         with complete_step("Removing empty directories…"):
             for d in reversed(sorted(context.root.glob("**/"))):
+                if d == context.root:
+                    continue
+
                 if not any(d.iterdir()):
                     d.rmdir()
```
septatrix, line 200: Using `context.root.walk(top_down=True)` would allow us to avoid the sort which eagerly consumes the iterator unnecessarily.

Also can't we simply break once we reached the root or simply add it as another check the the if below

PR #3390, opened 2025-01-20
Enforce that images with Overlay=yes only add files

DaanDeMeyer:
Any extension images built with Overlay=yes should never override files in the base image, so let's add some enforcement to make sure that's the case by automatically removing files that already exist in the base image.


mkosi/__init__.py
```diff
@@ -175,9 +175,22 @@ def mount_base_trees(context: Context) -> Iterator[None]:
...
+        with mount_overlay(bases, context.root, upperdir=context.root):
+            yield
+
+        stack.enter_context(mount_overlay(bases, context.workspace / "lower"))
+
+        for p in context.root.rglob("*"):
+            rel = p.relative_to(context.root)
+            q = context.workspace / "lower" / rel
+
+            if q.is_dir(follow_symlinks=False):
```
behrmann, line 187: That kwarg was only added in 3.13

mkosi/__init__.py
```diff
@@ -175,9 +175,22 @@ def mount_base_trees(context: Context) -> Iterator[None]:
...
+                if not p.is_dir(follow_symlinks=False):
+                    die(f"/{rel} is a directory in the base tree but not in the overlay")
+                shutil.copystat(q, p, follow_symlinks=False)
+            elif q.exists(follow_symlinks=False):
```
behrmann, line 191: This one was only added in 3.12 only unfortunately

PR #3395, opened 2025-01-21
kernel-install: Various fixes

DaanDeMeyer:


kernel-install/50-mkosi.install
```diff
@@ -52,19 +52,25 @@ def main() -> None:
...
     context = KernelInstallContext.parse(
-        "kernel-install plugin to build initrds or Unified Kernel Images using mkosi",
-        "50-mkosi.install COMMAND KERNEL_VERSION ENTRY_DIR KERNEL_IMAGE INITRD…",
+        name="50-mkosi.install",
+        description="kernel-install plugin to build initrds or Unified Kernel Images using mkosi",
     )
 
     if context.command != "add" or not we_are_wanted(context):
+        if context.verbose:
+            logging.info("mkosi-initrd is not enabled, skipping")
```
behrmann, line 61: It would be better to configure the log level to be e.g. warn by default and in case of verbose bump the log level once to info instead of prepending this everywhere.

behrmann (approved):

PR #3421, opened 2025-01-23
Fix VarLink Machine.Register invocation for v257+

labichn:
Ran into a couple of issues running with QEMU after b90dc2a50cc79f70cf1c919c82857c518c77cd43, and this branch fixes them.

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -1039,7 +1039,7 @@ def register_machine(config: Config, pid: int, fname: Path, cid: Optional[int]) 
                         "name": config.machine_or_name().replace("_", "-"),
                         "service": "mkosi",
                         "class": "vm",
-                        "leader": pid,
+                        "leader": {"pid": pid},
                         **({"rootDirectory": os.fspath(fname)} if fname.is_dir() else {}),
                         **({"vSockCid": cid} if cid is not None else {}),
                         **({"sshAddress": f"vsock/{cid}"} if cid is not None else {}),


DaanDeMeyer:
With your change, I get the following on systemd 256:

```
Method call io.systemd.Machine.Register() failed: org.varlink.service.InvalidParameter
‣ "varlinkctl call /run/systemd/machine/io.systemd.Machine io.systemd.Machine.Register '{"name": "default", "service": "mkosi", "class": "vm", "leader": {"pid": 2257407}, "rootDirectory": "/home/daandemeyer/projects/mkosi/mkosi.output/image-746e38fe3da34120", "vSockCid": 4049558221, "sshAddress": "vsock/4049558221", "sshPrivateKeyPath": "/home/daandemeyer/projects/mkosi/mkosi.key"}'" returned non-zero exit code 1.
```

Please give more information on what environment you're running in.

labichn:
Ah, I'm running pretty close to latest systemd@main, sorry for the lack of context. The PID structure change was introduced by https://github.com/systemd/systemd/pull/34703, and based on the PR comment I assume it was released with v257. (It's in v257-stable: https://github.com/systemd/systemd/blob/v257-stable/src/machine/machine-varlink.c#L48.)

Is there a standard way to encode functionality based on systemd version in the mkosi codebase so I can adjust the VarLink call for v257+? If you can point me to an example I'll make this change conditional.

Relevant bits from my `varlinkctl introspect /run/system/machine/io.systemd.Machine`:
```
# An object for referencing UNIX processes
type ProcessId(
        # Numeric UNIX PID value
        pid: int,
        # 64bit inode number of pidfd if known
        pidfdId: ?int,
        # Boot ID of the system the inode number belongs to
        bootId: ?int
)
...
method Register(
        name: string,
        id: ?string,
        service: ?string,
        class: string,
        leader: ?ProcessId,
        rootDirectory: ?string,
        ifIndices: ?[]int,
        vSockCid: ?int,
        sshAddress: ?string,
        sshPrivateKeyPath: ?string,
        # Controls whether to allocate a scope unit for the machine to register. If false, t>
        # client already took care of that and registered a service/scope specific to the
        # machine.
        allocateUnit: ?bool,
        # Controls whether interactive authentication (via polkit) shall be allowed. If
        # unspecified defaults to false.
        allowInteractiveAuthentication: ?bool
) -> ()
...
```

DaanDeMeyer:
@labichn This might be a bug in systemd-machined. It should accept the pid as specified here. Can you enable debug logging for systemd-machined (`run0 systemctl service-log-level systemd-machined debug`) and get the logs of systemd-machined (`journalctl -u systemd-machined -e`) after running `mkosi qemu`?

DaanDeMeyer:
@labichn This is a compat break between systemd 256 and systemd 257 so it should be fixed in systemd. Please open an issue there so we can track it.

PR #3422, opened 2025-01-24
Fix accessing "name" field in busctl json output

DaanDeMeyer:


behrmann (approved):
Sad this wasn't caught by a typechecker.

PR #3423, opened 2025-01-24
Only parse profiles from subimages and includes if those are dirs

DaanDeMeyer:


mkosi/config.py
```diff
@@ -4297,6 +4297,8 @@ def parse_config_one(self, path: Path, parse_profiles: bool = False, parse_local
         s: Optional[ConfigSetting[object]]  # Hint to mypy that we might assign None
         extras = path.is_dir()
 
+        assert path.is_absolute()
```
behrmann, line 4300: For typing assert is fine, but we shouldn't use it for errors.
```suggestion
        if not path.is_absolute():
            raise ValueError(f'{path} is not absolute')
```

behrmann (approved):

mkosi/config.py
DaanDeMeyer: @behrmann This is a programming error, not a user error, so an assert fits better imo

PR #3425, opened 2025-01-24
Use become_root_cmd() when running systemd-repart in run_shell() as well

DaanDeMeyer:
systemd-repart's --image switch requires root privileges as well, so let's use become_root_cmd() there as well.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -4010,6 +4010,7 @@ def run_shell(args: Args, config: Config) -> None:
                     network=True,
                     devices=True,
                     options=["--bind", fname, workdir(fname)],
+                    setup=become_root_cmd(),
                 ),
             )  # fmt: skip
 


behrmann (approved):

mkosi/__init__.py
```diff
@@ -4010,6 +4010,7 @@ def run_shell(args: Args, config: Config) -> None:
                     network=True,
                     devices=True,
                     options=["--bind", fname, workdir(fname)],
+                    setup=become_root_cmd(),
```
evverx, line 4013: As far as I understand the idea is to be able to run `mkosi boot` without `sudo` (and with the latest changes it works). Though it breaks when things like `--register=yes/Register=` are added because `mkosi` bails out with
```sh
Container registration requires root privileges
```
even though it should have no problem with that in theory. (I'm not sure how important it is though. It's just something I noticed when I poked mkosi in the context of HACKING.md)

evverx: It's of course possible to get it around with
```sh
mkosi boot --register=yes
```

PR #3434, opened 2025-01-27
zypper: set $releasever variable

aafeijoo-suse:
zypper allows to use the `$releasever` variable in repo URLs (see [zypper(8) man page](https://manpages.opensuse.org/Tumbleweed/zypper/zypper.8.en.html)). When mkosi uses zypper repos from the host instead of the default ones, this variable needs to be replaced with the host's `VERSION_ID`, otherwise repo URLs are incomplete.


DaanDeMeyer:
@aafeijoo-suse What about passing `--releasever` to zypper instead? At least for dnf that is equivalent to setting the var.

aafeijoo-suse:
> @aafeijoo-suse What about passing `--releasever` to zypper instead? At least for dnf that is equivalent to setting the var.

While I found this issue in Leap, that's where this variable is often used to ease distribution upgrades (e.g. `$releasever=15.6`), the only difference is that zypper always displays a warning with the `--releasever` option. E.g. in TW:

```
Warning: Enforced setting: $releasever=20250124
```

I'm fine with either way.

PR #3435, opened 2025-01-27
 opensuse: Make use of $releasever more 

DaanDeMeyer:


aafeijoo-suse:
Please, don't use `$releasever` on the default Tubleweed urls...

DaanDeMeyer:
Let's not bother in that case

PR #3437, opened 2025-01-27
Extension repart definitions

hundeboll:


DaanDeMeyer (changes requested):
First commit looks good, second one less so. Making the repart definitions for these output types a free for all does not seem like a good idea.

Can't you achieve what you need with `Format=disk` and your own repart definitions instead?

hundeboll:
> Can't you achieve what you need with `Format=disk` and your own repart definitions instead?

I wanted to benefit from the extension release file generated by `mkosi`... But if my use case is too niche, I can do with the `Format=disk` approach.

Another option would be a way to add `--exclude-partitions=root-verity-sig` to the `systemd-repart` command?

And while we're at it: Why not use `systemd-repart --make-ddi=sysext` instead of maintaining definition files?

hundeboll:
> Another option would be a way to add `--exclude-partitions=root-verity-sig` to the `systemd-repart` command?
> 
> And while we're at it: Why not use `systemd-repart --make-ddi=sysext` instead of maintaining definition files?

Okay, pushed a proposal using the `--make-ddi` option. This changes the output image to always contain a verity data hash partition. Let me know if this has interest...

DaanDeMeyer:
@hundeboll This is much worse, now the repart definitions have to be installed in the root filesystem or tools tree. That's exactly why we don't use --make-ddi= in the first place.

DaanDeMeyer:
@hundeboll Adding `Verity=hash` sounds good. You'll want to use `config_make_enum_parser_with_boolean()` for that.

hundeboll:
I added the `Verity=hash` option. Should be ready for review.

Thanks for the hint on `config_make_enum_parser_with_boolean` :+1:

mkosi/config.py
```diff
@@ -363,6 +363,13 @@ def __bool__(self) -> bool:
         return self != BuildSourcesEphemeral.no
 
 
+class Verity(StrEnum):
+    yes = enum.auto()
```
DaanDeMeyer, line 367: Let's name this `signed`

mkosi/resources/man/mkosi.1.md
```diff
@@ -1140,16 +1140,17 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
+    any verity partitions in the disk image produced by **systemd-repart**. If
+    disabled, verity partitions will be excluded from the extension images
+    produced by **systemd-repart**. If set to `hash`, **mkosi** configures
+    **systemd-repart** to create a verity hash partition, but no signature
+    partition. If set to `auto` and a verity key and certificate are present,
+    **mkosi** will pass them to systemd-repart and expects the generated disk
+    image to contain verity partitions, but the build won't fail if no verity
+    partitions are found in the disk image produced by **systemd-repart**.
 
     Note that explicitly disabling signed verity is not yet implemented
```
DaanDeMeyer, line 1155: This also needs to mention that `Verity=hash` is not supported for the `disk` output yet.

DaanDeMeyer (changes requested):

DaanDeMeyer:
Setting don't-merge until we get another stable release out

mkosi/config.py
hundeboll: It's okay to break existing users of `--verity=yes` ?

DaanDeMeyer: `parse_boolean()` will take care of the boolean values

septatrix (changes requested):
Thanks for working on this and implementing my proposal from #3121. However for consistency we should either reuse the nomenclature from either sd-repart or sd.image-policy. Those are off/hash/signature and off/verity/signed respectively

mkosi/resources/man/mkosi.news.7.md
```diff
@@ -4,6 +4,12 @@
 
 # mkosi Changelog
 
+## v26
+
+- Teach --verity a new new `hash` value, which skips the verity signature
+  partition for extension / portable images. To align the possible values,
+  `yes` is renamed to `signed`.
```
septatrix, line 11: signed -> signature

tests/test_json.py
```diff
@@ -388,7 +389,7 @@ def test_config() -> None:
             "UseSubvolumes": "auto",
             "VSock": "enabled",
             "VSockCID": -2,
-            "Verity": "enabled",
+            "Verity": "signed",
```
septatrix, line 392: Same here

septatrix (changes requested):

mkosi/resources/man/mkosi.1.md
```diff
@@ -1140,19 +1140,21 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
+    from the extension images produced by **systemd-repart**. If set to
+    `verity`, **mkosi** configures **systemd-repart** to create a verity hash
+    partition, but no signature partition. If set to `auto` and a verity key
+    and certificate are present, **mkosi** will pass them to systemd-repart and
+    expects the generated disk image to contain verity partitions, but the build
+    won't fail if no verity partitions are found in the disk image produced by
+    **systemd-repart**.
+
+    Note that explicitly disabling verity signature and/or hash is not yet
+    implemented for the `disk` output and only works for extension images at the
```
septatrix, line 1157: `defer` is not documented

mkosi/resources/man/mkosi.news.7.md
septatrix: Not resolved?

DaanDeMeyer:
Closing this one as I'm working on top of it (attribution for the first commits is of course preserved)

PR #3441, opened 2025-01-28
Move want_uki() check out of build_uki_profiles()

DaanDeMeyer:
Otherwise the function is skipped in make_uki() even though it should never be skipped in that case if there are UKI profiles defined.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2059,7 +2059,7 @@ def systemd_addon_stub_binary(context: Context) -> Path:
 
 
 def build_uki_profiles(context: Context, cmdline: Sequence[str]) -> list[Path]:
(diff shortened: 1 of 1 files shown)


behrmann (approved):
LGTM, but in the case of UKI profiles being defined, shouldn't then `want_uki` not always be true? What I mean, should `want_uki` grow 
```python
if context.config.unified_kernel_image_profiles:
    return True
```

bluca (approved):

DaanDeMeyer:
> LGTM, but in the case of UKI profiles being defined, shouldn't then want_uki not always be true? What I mean, should want_uki grow

No, there might be profiles defined for "if" a UKI is built, but they shouldn't case us to always build a uki

PR #3442, opened 2025-01-28
portable images require an os-release file

mt-at-mt:
This fixes:

root@mymachine:~# portablectl attach foo_3
(Matching unit files with prefix 'foo'.)
AttachImage failed: Image '/var/lib/portables/foo_3.raw' lacks os-release data, refusing.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -189,7 +189,7 @@ def mount_base_trees(context: Context) -> Iterator[None]:
                 if p.is_symlink() or not p.is_dir():
                     die(f"/{rel} is a directory in the base tree but not in the overlay")
                 shutil.copystat(q, p)
-            elif q.is_symlink() or q.exists():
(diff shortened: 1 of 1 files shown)


DaanDeMeyer:
The os-release file is supposed to come from your distribution. You seem to be using this with `Overlay=yes` which is almost always going to be wrong for portable services unless you're using extension images for your portable images. You probably want to not use `Overlay=yes`, but see the faq entry on `PORTABLE_PREFIXES` as well (https://github.com/systemd/mkosi/blob/main/mkosi/resources/man/mkosi.1.md#frequently-asked-questions-faq)

mt-at-mt:
> The os-release file is supposed to come from your distribution. You seem to be using this with `Overlay=yes`

yes, I expected that just the additional files are needed for a portable service.

> which is almost always going to be wrong for portable services unless you're using extension images for your portable images. 

Got it, portable means portable between systemd based distros.

PR #3443, opened 2025-01-28
mkosi-sandbox: Improve formatting of error messages

DaanDeMeyer:


mkosi/sandbox.py
```diff
@@ -100,17 +100,37 @@ class cap_user_data_t(ctypes.Structure):
...
+    reset: str  = "\033[0m"          if not terminal_is_dumb() else ""
+    # fmt: on
+
+
+ENOSYS_MSG = """\
+{red}mkosi was unable to invoke the {syscall}() system call.{reset}
 This probably means either the system call is not implemented by the running kernel version ({kver}) or the
-system call is prohibited via seccomp if mkosi is being executed inside a containerized environment.
+system call is prohibited via seccomp if mkosi is being executed inside a containerized environment.\
 """
```
behrmann, line 125: This could be formatted here (like below)
```python
ENOSYS_MSG = f"""\
{Style.red}mkosi was unable to invoke the {{syscall}}() system call.{Style.reset}
This probably means either the system call is not implemented by the running kernel version ({{kver}}) or the
system call is prohibited via seccomp if mkosi is being executed inside a containerized environment.\
"""

behrmann (approved):

PR #3446, opened 2025-01-28
Various cache fixes

DaanDeMeyer:
We have access to the config object in have_cache() and this argument is specifically intended to be used whenever we pass the default tools tree to have_cache(), so let's just do the check based on config.image in have_cache() itself.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3114,10 +3114,12 @@ def have_cache(config: Config) -> bool:
 
 
 def reuse_cache(context: Context) -> bool:
-    if not have_cache(context.config):
+    if not context.config.incremental or context.config.base_trees or context.config.overlay:
         return False
 
     final, build, _ = cache_tree_paths(context.config)
+    if not final.exists() or (need_build_overlay(context.config) and not build.exists()):
+        return False
 
     with complete_step("Copying cached trees"):
         copy_tree(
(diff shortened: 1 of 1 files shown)


behrmann (commented):
The first two look good to me, the third one I don't quite get, since it's mostly doing the same as `have_cache` with less logging interspersed. The only marked difference is that `have_cache` also looks at the manifest, which could maybe be optional with a switch or go into a separate function.

DaanDeMeyer:
> The first two look good to me, the third one I don't quite get, since it's mostly doing the same as `have_cache` with less logging interspersed. The only marked difference is that `have_cache` also looks at the manifest, which could maybe be optional with a switch or go into a separate function.

The thing is, by the time we call reuse_cache(), we've already called have_cache() and cleaned up if needed, so I think it makes more sense for reuse_cache() to behave based on the paths on disk and not check the manifest again.

behrmann (approved):

behrmann:
> The thing is, by the time we call reuse_cache(), we've already called have_cache() and cleaned up if needed, so I think it makes more sense for reuse_cache() to behave based on the paths on disk and not check the manifest again.

I don't disagree with the reasoning not check the manifest again, my point was that there's a lot of common logic in have_cache and reuse_cache now, but that can be fixed later, since the same idioms appear elsewhere, too.

PR #3448, opened 2025-01-28
Treat terminal as dumb if either stdout or stderr is not a tty

### LINKED ISSUE #3445: Dumb terminal detection for `cat-config` and `summary` is still broken as those only look at stderr but not stdout
For piping stuff like `cat-config` and `summary` this still happens because those pipe to stdout (which is not a tty) but stderr is still a tty so they still try to color the output

_Originally posted by @septatrix in https://github.com/systemd/mkosi/issues/3017#issuecomment-2435578640_

DaanDeMeyer:
Fixes #3445


mkosi/sandbox.py
```diff
@@ -101,7 +101,7 @@ class cap_user_data_t(ctypes.Structure):
 
 
 def terminal_is_dumb() -> bool:
-    if not sys.stdout.isatty() and not sys.stderr.isatty():
+    if not sys.stdout.isatty() or not sys.stderr.isatty():
         return True
 
     return os.getenv("TERM", "") == "dumb"
```
behrmann, line 107: Let's maybe shorten this down to this
```suggestion
    return not sys.stdout.isatty() or not sys.stderr.isatty() or os.getenv("TERM", "") == "dumb"
```
since setting `TERM=dumb` should be enough even if attached to a TTY, that would be consistent with e.g. `tput`.

PR #3451, opened 2025-01-29
Fix condition when removing duplicate files from the overlay

DaanDeMeyer:


behrmann (approved):

stgraber:
We (incus-os) were affected by this and can confirm that this PR fixes the issue for us.

PR #3461, opened 2025-02-03
genkey: also generate the pubkey in addition to the certificate

behrmann:
This is relevant for PCR signature workflow in ukify.

/cc @bluca

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ mkosi.local.conf
 mkosi.tools
 /mkosi.key
 /mkosi.crt
+/mkosi.pub
 __pycache__
 blog/output
 blog/pelicanconf.py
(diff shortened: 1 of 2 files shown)


DaanDeMeyer (commented):
Should we really do this inside of mkosi? It's trivial to generate outside of mkosi no?

behrmann:
I don't think it hurts to have another file around and it makes it more seemless. I just ran into this by trying to use `mkosi.crt` instead and had to ask @bluca for the necessary openssl invocation.

DaanDeMeyer:
What do you need the public key for?

behrmann:
The example in the ukify man page in [this PR](https://github.com/systemd/systemd/pull/36181)

bluca:
Some commands need a pub key instead of a cert, like measure

behrmann:
Yes, that's probably the way to go instead.

PR #3466, opened 2025-02-04
Extend Verity= to support hash-only verity and deferred verity-sig

DaanDeMeyer:


DaanDeMeyer:
@septatrix I've reverted the changes to use `signature` and `verity`. I disagree that these need to be the same as systemd-repart. The `Verity=` setting in systemd-repart indicates whether a partition is the verity data, hash or signature partition. The `Verity=` setting in mkosi indicates whether we should do signed, hash-only or no verity at all. The settings may have the same name, but they do completely different things, hence there's no reason to use the same names as the repart setting.

behrmann (approved):

septatrix:
> @septatrix I've reverted the changes to use `signature` and `verity`. I disagree that these need to be the same as systemd-repart. The `Verity=` setting in systemd-repart indicates whether a partition is the verity data, hash or signature partition. The `Verity=` setting in mkosi indicates whether we should do signed, hash-only or no verity at all. The settings may have the same name, but they do completely different things, hence there's no reason to use the same names as the repart setting.

I think that's fair and fine by me. One could even justify this by saying that this uses the names from sd-image-policy and "hash = verity - signed" ;)

mkosi/resources/man/mkosi.news.7.md
```diff
@@ -1037,7 +1043,7 @@
   configuration from the host system, instead of the internal list.
 - A new `SshAgent=` option configures the path to the ssh agent.
 - A new `SshPort=` option overrides the port used for ssh.
-- The `Verity=` setting supports a new value `signed`. When set, verity data
+- The `Verity=` setting supports a new value `signature`. When set, verity data
```
septatrix, line 1050: signature -> signed (also why is the changelog for v11 even touched?)

bluca (approved):
Tested this change in OBS for the deferred option and works nicely

hundeboll:
@DaanDeMeyer Thanks for finishing this. I was back late from FOSDEM, and then had to overcome the flu I picked up while there...

DaanDeMeyer:
> @DaanDeMeyer Thanks for finishing this. I was back late from FOSDEM, and then had to overcome the flu I picked up while there...

@hundeboll No worries, thanks for doing the initial work. You should have come say hi at FOSDEM, we (mkosi/systemd maintainers) were there as well.

PR #3468, opened 2025-02-04
Drop two unneeded calls to umask()

DaanDeMeyer:
Both these commands write temporary files to the workspace which are not written to the image, so no need to care about the umask.


behrmann (approved):
Looks reasonable. None of the the commands warns about permissions, right?

DaanDeMeyer:
> Looks reasonable. None of the the commands warns about permissions, right?

They're operating on certificates, which are public, so no

PR #3473, opened 2025-02-05
kmod: do not filter out user-defined firmware

aafeijoo-suse:
A user may want to force the addition of certain firmware files that cannot be determined as a dependecy of any added kernel module, so do not remove those.


mkosi/kmod.py
```diff
@@ -258,8 +258,25 @@ def process_kernel_modules(
...
+                    and fw.is_relative_to(firmwared)
+                )
+            ):
+                user_fw.add(fw)
+                for p in parents_below(context.root / fw, context.root / firmwared):
+                    user_fw.add(p.relative_to(context.root))
+
         for fw in firmware:
-            if fw in required:
+            if fw in required or fw in user_fw:
```
DaanDeMeyer, line 279: I'm a bit confused, in mkosi-initrd we do `"--extra-tree=/usr/lib/firmware:/usr/lib/firmware",`, doesn't this mean that we stop filtering firmware at all in that case?

aafeijoo-suse: The extra firmware added using `ExtraTrees=` in user conf is added, but the removed within this loop if it's not required by the any kernel modules.

DaanDeMeyer: Yeah I get that, but in mkosi-initrd all the user firmware comes from "user conf", so we'll stop removing any firmware at all.

DaanDeMeyer: @aafeijoo-suse Isn't https://github.com/systemd/mkosi/pull/3473 what you want instead of this? (educated guess from me, not sure if it's related)

aafeijoo-suse: > Yeah I get that, but in mkosi-initrd all the user firmware comes from "user conf", so we'll stop removing any firmware at all.

hmm, not on my tests... if I add extra firmware in conf and that firmware does not appear to be dependent of any kernel modules, it's removed here.

> Isn't https://github.com/systemd/mkosi/pull/3473

yeah, 3473 is what I've opened :)

DaanDeMeyer: oops I meant https://github.com/systemd/mkosi/pull/3474

aafeijoo-suse: > oops I meant #3474

that means initrds > 600 MB... it's not about the generic mode only, it's about adding what the user wants

aafeijoo-suse: btw, it's very common to force the addition of firmware files in the initrd

DaanDeMeyer: @aafeijoo-suse I get that, I just don't fully understand how we can distinguish between "forced" firmware file additions and regular ones with just --extra-tree. In the mkosi-initrd logic, All the firmware is added with --extra-tree, which I assume we want to be filtered based on which kernel modules are installed.

aafeijoo-suse: Yes, and the filter works well depending on the kernel modules added (loaded or generic), that's perfect.

> btw, it's very common to force the addition of firmware files in the initrd

Maybe I expressed myself badly with this sentence. For a regular use case, this should not happen, i.e., the firmware detected as dependent is already enough. But for some corner cases, users have to explicitly add 2,3,4... fw files. For that, we could have something like `FirmwareInclude=`, but I thought handling these files with `ExtraTrees=` was enough (for example, in dracut we use `install_items`, which is the generic way to add any type of file).

DaanDeMeyer: @aafeijoo-suse So doing this with `ExtraTrees=` won't work, since we need the firmware exclusion to work with `ExtraTrees=` to make mkosi-initrd work properly.

I think adding `FirmwareInclude=` is the way to go here.

aafeijoo-suse: > So doing this with ExtraTrees= won't work

At least on my tests this works...

> I think adding FirmwareInclude= is the way to go here.

But definitely having a specific option for this it's cleaner.

aafeijoo-suse:
Replaced by #3483

PR #3474, opened 2025-02-06
mkosi-initrd: Keep all kernel modules and firmware in generic mode

DaanDeMeyer:
Currently, we only keep what's in mkosi-initrd's builtin allowlist, which is rather limited and results in most drivers and firmware being removed. Let's instead copy dracut and keep all modules and drivers when building a generic initrd.

### DIFF
--- a/mkosi/initrd.py
+++ b/mkosi/initrd.py
@@ -251,13 +251,15 @@ def main() -> None:
             "--extra-tree", f"/usr/lib/modules/{args.kernel_version}:/usr/lib/modules/{args.kernel_version}",
             "--extra-tree=/usr/lib/firmware:/usr/lib/firmware",
             "--remove-files=/usr/lib/firmware/*-ucode",
-            "--kernel-modules-exclude=.*",
             "--build-sources", "",
             "--include=mkosi-initrd",
         ]  # fmt: skip
 
         if not args.generic:
-            cmdline += ["--kernel-modules-include=host"]
+            cmdline += [
+                "--kernel-modules-exclude=.*",
+                "--kernel-modules-include=host",
+            ]
 
         if args.kernel_image:
             cmdline += [


DaanDeMeyer:
@aafeijoo-suse So I understand this leads to 600MB initrds but it's kinda weird that in dracut if you do nohostonly mode you get more modules and firmware and in our case if you do --generic you get less drivers and firmware.

aafeijoo-suse:
> @aafeijoo-suse So I understand this leads to 600MB initrds but it's kinda weird that in dracut if you do nohostonly mode you get more modules and firmware and in our case if you do --generic you get fewer drivers and firmware.

The reason is every dracut module installs its specific kernel modules, and there is a lot of dracut modules, most of them included in no-hostonly. But here the picture is different, we have this common mkosi-initrd configuration with a relative good set of kernel modules installed, and that's all. So, IMO the size of the initrd should grow only depending on specific mkosi-initrd configurations providing some specific functionalities, but having everything from the start seems too much to me, and would make this --generic switch a bit unusable.

DaanDeMeyer:
@aafeijoo-suse Note that the default include list is very likely missing a huge amount of modules. It's being built piece by piece by me (and now Lennart as well) installing particleos on various hardware and seeing what's missing.

aafeijoo-suse:
> @aafeijoo-suse Note that the default include list is very likely missing a huge amount of modules. It's being built piece by piece by me (and now Lennart as well) installing particleos on various hardware and seeing what's missing.

Yes, I know, but it's a good start :)

septatrix:
We could also add all modules from the Ubuntu Core initrd: https://github.com/canonical/core-initrd/blob/main/modules/main/extra-modules.conf

Similarly other tools maintain module lists for host-independet initrds

PR #3475, opened 2025-02-06
Various rpm gpg key lookup fixes

DaanDeMeyer:


mkosi/distributions/fedora.py
```diff
@@ -66,11 +69,27 @@ def find_fedora_rpm_gpgkeys(context: Context) -> Iterable[str]:
...
 
-            keyurl = f"{keys}/{key}"
-        else:
-            keyurl = "https://fedoraproject.org/fedora.gpg"
+            yield f"{keys}/{key}"
 
-        yield keyurl
+            # Same as above, the symlink in distribution-gpg-keys might not have been updated yet to point to
+            # the new rawhide key when branching happens, so try to load the N+1 key as well.
+            if m := re.match(versionre, key):
```
keszybz, line 76: I think we never get here. Either `key1` or `key2` must be set.

keszybz: > GPG Keys are configured as: file:///usr/share/distribution-gpg-keys/fedora/RPM-GPG-KEY-fedora-rawhide-primary

Right. So we find the rawhide key on disk, but for some reason we don't find the adjacent key file.

mkosi/distributions/fedora.py
```diff
@@ -29,18 +30,20 @@ def find_fedora_rpm_gpgkeys(context: Context) -> Iterable[str]:
         context, key=f"RPM-GPG-KEY-fedora-{context.config.release}-secondary", required=False
     )
 
+    versionre = r"RPM-GPG-KEY-fedora-(\d+)-(primary|secondary)"
```
behrmann, line 33:
```suggestion
    versionre = re.compile(r"RPM-GPG-KEY-fedora-(\d+)-(primary|secondary)")
```
so that below you can do, e.g. 
```
versionre.match(Path(rhs).name)
```

behrmann: You could also do
```
versionre = re.compile(r"RPM-GPG-KEY-fedora-(?P<versionnum>\d+)-(primary|secondary)")
```
so that the group doesn't have to be looked up by number

PR #3477, opened 2025-02-06
Use verity repart definitions for extension/portable with hash and defer

hundeboll:
The --exlude-partitions and --defer-partitions options to systemd-repart makes no difference if none of the excluded or deferred partitions are defined. So make dure the verity definitions are used when Verity=hash or Verity=defer is configured.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3532,7 +3532,7 @@ def make_esp(context: Context, uki: Path) -> list[Partition]:
 
 
 def make_extension_or_portable_image(context: Context, output: Path) -> None:
-    if want_verity(context.config) or context.config.verity == Verity.signed:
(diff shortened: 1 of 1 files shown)


DaanDeMeyer (approved):

DaanDeMeyer:
@hundeboll Please fix `want_verity()` as well

DaanDeMeyer:
Actually we can just fix `want_verity()` and remove the or from this check

hundeboll:
> Actually we can just fix `want_verity()` and remove the or from this check

Sure about that? The `want_verity()` is also used to decide if a key/certificate argument should be passed to `systemd-repart`.

hundeboll:
But we can remove the `Verity.signed` from the tuple, as `want_verity()` checks for that case :)

DaanDeMeyer:
I prepped #3478 which is a more comprehensive fix, so let's close this one

PR #3479, opened 2025-02-06
tree: Don't copy xattrs to overlayfs

DaanDeMeyer:
Trying to copy the selinux xattrs to a directory in an overlayfs
filesystem will fail with "Operation not supported". There's no way
to instruct cp to not copy or ignore failures to copy selinux xattrs
so let's instead not try to copy xattrs at all when copying to directories
in overlayfs filesystems.

### DIFF
--- a/mkosi/tree.py
+++ b/mkosi/tree.py
@@ -123,6 +123,7 @@ def copy_tree(
         use_subvolumes == ConfigFeature.disabled
         or not preserve
         or not is_subvolume(src)
+        or statfs(str(dst.parent)) != BTRFS_SUPER_MAGIC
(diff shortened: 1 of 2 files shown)


mkosi/tree.py
```diff
@@ -123,6 +131,7 @@ def copy() -> None:
         use_subvolumes == ConfigFeature.disabled
         or not preserve
         or not is_subvolume(src)
+        or statfs(str(dst.parent)) != BTRFS_SUPER_MAGIC
```
behrmann, line 134:
```suggestion
        or statfs(os.fspath(dst.parent)) != BTRFS_SUPER_MAGIC
```

behrmann (approved):

PR #3480, opened 2025-02-06
Use `os.fspath` in more places

behrmann:
Should hopefully be all of them (outside of sandbox).

### DIFF
--- a/docs/CODING_STYLE.md
+++ b/docs/CODING_STYLE.md
@@ -1,0 +1,40 @@
+---
+title: Coding Style
+category: Contributing
+layout: default
+SPDX-License-Identifier: LGPL-2.1-or-later
+---
+
+# Coding Style
+
+## Python Version
+
+- The lowest supported Python version is CPython 3.9.
+
+## Formatting
+
+- Use the accompanying `.editorconfig` or `.dir-locals.el`.
+- For Python files we use the style from `ruff format` with a line length of 109 characters and four spaces
+  indentation.
+- Indentation with tabs is not allowed.
+- When it improves readability, judicious use of `# noqa: E501` comments is allowed.
+- Long lists, including argument lists, should have a trailing comma to force ruff to split all elements on a
+  line of their own.
+- List of commandline arguments should not split the argument of a commandline option and the option. This
+  needs to be enforced with `# fmt: skip` comments, e.g. do
+  ```python
+  cmd = [
+      "--option", "foo",
+  ]  # fmt: skip
+  ```
+  and do NOT do
+  ```python
+  cmd = [
+      "--option",
+      "foo",
+  ]
+  ```
+- When coercing Path-like objects to strings, use `os.fspath`, since this calls the `__fspath__` protocol
+  instead of `__str__`. It also ensures more type-safety, since every Python object supports `__str__`, but
+  not all support `__fspath__` and this gives the typechecker more information what is expected at this
+  point. It also signals the intent to the reader more than a blanket `str()`.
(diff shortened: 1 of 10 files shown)


docs/CODING_STYLE.md
```diff
@@ -0,0 +1,40 @@
...
+
+# Coding Style
+
+## Python Version
+
+- The lowest supported Python version is CPython 3.9.
+
+## Formatting
+
+- Use the accompanying `.editconfig` or `.dir-locals.el`.
```
DaanDeMeyer, line 16: Typo

docs/CODING_STYLE.md
```diff
@@ -0,0 +1,40 @@
...
+- Long lists, including argument lists, should have a trailing comma to force ruff to split all elements on a
+  line of their own.
+- List of commandline arguments should not split the argument of a commandline option and the option. This
+  needs to be enforced with `# fmt: skip` comments, e.g. do
+  ```python
+  cmd = [
+      "--option", "foo",
+  ]  # fmt: skip
+  ```
+  and do NOT
```
DaanDeMeyer, line 30: do not do

docs/CODING_STYLE.md
```diff
@@ -0,0 +1,40 @@
...
+  ```python
+  cmd = [
+      "--option",
+      "foo",
+  ]
+  ```
+- When coercing Path-like objects to strings, use `os.fspath`, since this calls the `__fspath__` protocol
+  instead of `__str__`. It also ensures more type-safety, since every Python object supports `__str__`, but
+  not all support `__fspath__` and this gives the typechecker more information what is expected at this. It
+  also signals the intent to the reader than a blanket `str()`.
```
DaanDeMeyer, line 40: more than

docs/CODING_STYLE.md
```diff
@@ -0,0 +1,40 @@
...
+  ]  # fmt: skip
+  ```
+  and do NOT
+  ```python
+  cmd = [
+      "--option",
+      "foo",
+  ]
+  ```
+- When coercing Path-like objects to strings, use `os.fspath`, since this calls the `__fspath__` protocol
```
DaanDeMeyer, line 37: Missing period

DaanDeMeyer (changes requested):

docs/CODING_STYLE.md
```diff
@@ -0,0 +1,40 @@
...
+# Coding Style
+
+## Python Version
+
+- The lowest supported Python version is CPython 3.9.
+
+## Formatting
+
+- Use the accompanying `.editconfig` or `.dir-locals.el`.
+- For Python files we use the style from `ruff format` with a line length of 109 characters and four spaces
```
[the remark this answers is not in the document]
hundeboll: Isn't the period already there after `indentation` on the following line?

docs/CODING_STYLE.md
```diff
@@ -0,0 +1,40 @@
...
+  ```python
+  cmd = [
+      "--option",
+      "foo",
+  ]
+  ```
+- When coercing Path-like objects to strings, use `os.fspath`, since this calls the `__fspath__` protocol
+  instead of `__str__`. It also ensures more type-safety, since every Python object supports `__str__`, but
+  not all support `__fspath__` and this gives the typechecker more information what is expected at this
+  point. It also signals the intent to the reader more than a blanket `str()`.
```
hundeboll, line 40: I tend to use `p.as_posix()` instead. Has that been considered?

docs/CODING_STYLE.md
hundeboll: Isn't the period already after `__str__` on the following line?

hundeboll (approved):

docs/CODING_STYLE.md
DaanDeMeyer: I don't see the point of using this in a codebase that only works on Linux in the first place

Issue #3376, opened 2025-01-17
`bootctl` fails to find `/buildroot/usr/lib/systemd/boot/efi`


rustydb:
### mkosi commit the issue has been seen with

7979fb772323db64b1c7a8267cd6dd6d8c8ab99f

### Used host distribution

Debian 12.8

### Used target distribution

RHEL 9.5

### CPU architectures issue was seen on

aarch64

### Unexpected behaviour you saw

I updated `mkosi` 1-2 days ago and my build started failing during the bootloader setup.

If I checkout f03944912a01573425def66ac99e40fd315113ea, the one preceding  7eab77bfb91fc5c8bf6f32da24a9cd76d9c150fe, I can build just fine.

For now I'm working around this using f03944912a01573425def66ac99e40fd315113ea.

Do I need to do something new now to provide `/usr/lib/systemd/boot/efi` or is this a bug in `bootloader.py`?

[mkosi_bootloader_build_failure.log](https://github.com/user-attachments/files/18459210/mkosi_bootloader_build_failure.log)

### Used mkosi config

```ini
### IMAGE: default
# mkosi.conf
[Config]
MinimumVersion=v24.3

[Distribution]
Distribution=rhel
Release=9

[Output]
Format=disk
ImageId=rhel
ManifestFormat=json,changelog
OutputDirectory=out

[Build]
ToolsTree=default
ToolsTreePackages=
    qemu-user
    qemu-user-static

[Content]

# "hashed:" with no hash produces a blank password.
RootPassword=hashed:

# Ignore SELinux (breaks boots, at least for VMs).
KernelCommandLine=
        enforcing=0

Packages=
    kernel-core
Bootable=true
WithRecommends=0
WithDocs=0
CleanPackageMetadata=true

# mkosi.conf.d/arm64.conf
[Match]
Architecture=|arm64

[Distribution]
LocalMirror=https://preserve.eag.rdlabs.hpecorp.net/mirrors/netinst/rhel/9.5/aarch64/latest/BaseOS?ssl_verify=no

[Build]
ToolsTreeCertificates=false
Environment=
    DEBUG=1
```

### mkosi output

```sh
See attachement; debug output too long.
```

DaanDeMeyer:
Need to install systemd-boot in the image. We were accidentally picking it up from the host.

rustydb:
Thanks.

9cel:
> Need to install systemd-boot in the image. We were accidentally picking it up from the host.

[...]

PR #3482, opened 2025-02-07
doc: Add a note about dependencies for `Bootloader=systemd-boot`

9cel:
Adds a small note to (hopefully) make it clear that `Bootloader=systemd-boot` requires `bootctl` in the image. See https://github.com/systemd/mkosi/issues/3376

### DIFF
--- a/mkosi/resources/man/mkosi.1.md
+++ b/mkosi/resources/man/mkosi.1.md
@@ -896,6 +896,11 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     `vmlinuz`. The `$version` is as produced by Kbuild's `kernelversion` make
     target.
 
+    Note: When using `systemd-boot` or `systemd-boot-signed`, `mkosi` expects
+    the `systemd-boot` EFI binaries to be present in the image. Depending on
+    your distribution, these may be packaged separately. For example, Debian-
+    based images will need `systemd-boot-efi`.
+
 `BiosBootloader=`, `--bios-bootloader=`
 :   Takes one of `none` or `grub`. Defaults to `none`. If set to `none`,
     no BIOS bootloader will be installed. If set to `grub`, grub is


DaanDeMeyer (changes requested):
We don't need `bootctl` in the image, we need `systemd-boot` in the image. Please mention that the `systemd-boot` package should be installed.

mkosi/resources/man/mkosi.1.md
```diff
@@ -888,6 +888,10 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     in the ESP for compatibility with signed versions of grub which load
     the grub configuration from this location.
 
+    Note: When using `systemd-boot` or `systemd-boot-signed`, the image
+    must include `bootctl`. This usually requires adding a package, e.g.
+    `systemd-boot` on Debian-based distributions or `systemd` on Arch.
```
behrmann, line 893: Isn't it hard not to have `systemd` on Arch? How about
```suggestion
    must include `bootctl`. This may require adding a package when it is packaged separately, e.g.
    `systemd-boot` on Debian-based distributions.
```

mkosi/resources/man/mkosi.1.md
```diff
@@ -888,6 +888,10 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     in the ESP for compatibility with signed versions of grub which load
     the grub configuration from this location.
 
+    Note: When using `systemd-boot` or `systemd-boot-signed`, the image
```
behrmann, line 891: Since this is more of a footnote type thing and more paragraphs follow, could you move this at the end of the option description, i.e. after the paragraph that starts with "Kernels need to be placed"?

9cel:
Sorry 😭 it worked after I added the `systemd-boot` package, and that only (directly) provides `bootctl` so I assumed that was the dependency. I guess it's really the EFI binaries that are necessary, which are pulled in indirectly since `systemd-boot` depends on `systemd-boot-efi`.

Hopefully it makes more sense now.

DaanDeMeyer (approved):

PR #3483, opened 2025-02-07
Add FirmwareInclude= and FirmwareExclude= options

aafeijoo-suse:
Rigth now, firmware added to the image is determined as a dependency of any added kernel modules. However, a common use case is that a user may need to force certain firmware to be included or excluded regardless of that dependency.

Replaces #3473 based on https://github.com/systemd/mkosi/pull/3473#discussion_r1944338648


mkosi/config.py
```diff
@@ -1860,6 +1860,9 @@ class Config:
     kernel_modules_initrd_exclude: list[str]
     kernel_modules_initrd_include_host: bool
```
DaanDeMeyer, line 1862: Please drop this newline to keep it together with the kernel module include/exclude settings

mkosi/config.py
```diff
@@ -2891,6 +2894,20 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
         parse=config_make_list_parser(delimiter=","),
         help="When building a kernel modules initrd, exclude the specified kernel modules",
     ),
+    ConfigSetting(
+        dest="firmware_include",
+        metavar="REGEX",
```
DaanDeMeyer, line 2899: Should we go for globs? I've long since realized that making the kernel modules include/exclude settings regex was a mistake.

mkosi/config.py
```diff
@@ -4909,6 +4926,9 @@ def summary(config: Config) -> str:
       Kernel Modules Initrd Exclude: {line_join_list(config.kernel_modules_initrd_exclude)}
  Kernel Modules Initrd Include Host: {yes_no(config.kernel_modules_initrd_include_host)}
```
DaanDeMeyer, line 4928: Please drop this newline as well

mkosi/kmod.py
```diff
@@ -195,6 +231,12 @@ def gen_required_kernel_modules(
             mods = set(modulesd.rglob("*.ko*"))
         firmware = set()
 
+    # Include or exclude firmware explicitly configured
+    if firmware_include or firmware_exclude:
+        firmware.update(
```
DaanDeMeyer, line 236: So this will only ever add more firmware to the list which I think isn't quite the right semantics. For example, at the moment, if a module depends on some firmware, even if that firmware is listed in `FirmwareExclude=`, it'll still be added to the image.

I think the right semantics are the following:

- FirmwareExclude: The firmware matched by this is always excluded from the image, even if it is required by a kernel module.
- FirmwareInclude: The firmware matched by this is always included in the image, even if it is not required by a kernel module and even if it is matched by FirmwareExclude.

In terms of implementation, I think this can be done in two phases. Phase 1 removes firmware from the list of firmware returned by `resolve_module_dependencies()` based on FirmwareExclude= and FirmwareInclude=. Phase 2 adds firmware not in the list of firmware based on FirmwareInclude=.

DaanDeMeyer (changes requested):
We should add `FirmwareIncludeInitrd=` and `FirmwareExcludeInitrd=` as well

DaanDeMeyer:
Actually let's not add FirmwareIncludeInitrd= and FirmwareExcludeInitrd= for now until someone requests them. And even then we might be better off requiring them to build the initrd via mkosi.images/ instead.

mkosi/config.py
behrmann: While that would be easier, having this disparate doesn't seem like a good idea, they should behave the same. We could break compat again and make the others globs as well.

mkosi/config.py
aafeijoo-suse: There is an empty line a bit before, should I remove that too?

https://github.com/systemd/mkosi/blob/c2b7730ea0246285149d924c6d467808981ded2e/mkosi/config.py#L1853-L1861

mkosi/config.py
aafeijoo-suse: Same as before:

https://github.com/systemd/mkosi/blob/c2b7730ea0246285149d924c6d467808981ded2e/mkosi/config.py#L4902-L4910

mkosi/config.py
aafeijoo-suse: Well, I just kept what it was done for kernel modules...

mkosi/kmod.py
aafeijoo-suse: > So this will only ever add more firmware to the list

Right, I missed this part, it should remove elements from the initial set, not only add. I'll rework that.

mkosi/config.py
DaanDeMeyer: Let's just keep it as regex, I don't want to break compat on this since it's likely actually used often

mkosi/config.py
DaanDeMeyer: Keep it separate from the initrd options but next to the regular include/exclude options. Maybe github messed up the formatting

mkosi/kmod.py
aafeijoo-suse: > I think the right semantics are the following:
> 
>     * FirmwareExclude: The firmware matched by this is always excluded from the image, even if it is required by a kernel module.
> 
>     * FirmwareInclude: The firmware matched by this is always included in the image, even if it is not required by a kernel module and even if it is matched by FirmwareExclude.

Actually this is already done within the `filter_firmware()` function, I don't think it is necessary to split the implementation in two phases, it's enough to replace the `firmware.update()` with `firmware =`.

mkosi/config.py
```diff
@@ -4909,6 +4925,8 @@ def summary(config: Config) -> str:
...
 
+                   Firmware Include: {line_join_list(config.firmware_include)}
+                   Firmware Exclude: {line_join_list(config.firmware_exclude)}
                              Locale: {none_to_default(config.locale)}
```
DaanDeMeyer, line 4930: There should be a newline between these two fields

tests/test_json.py
```diff
@@ -451,8 +457,10 @@ def test_config() -> None:
         extra_trees=[],
         files=[],
         finalize_scripts=[],
-        firmware_variables=Path("/foo/bar"),
```
DaanDeMeyer, line 454: This should be kept, at least in vscode's sorting command, "_" sorts before "="

mkosi/kmod.py
```diff
@@ -55,6 +55,39 @@ def filter_kernel_modules(
...
+def filter_firmware(
+    root: Path,
+    firmware: set[Path],
+    *,
+    include: Iterable[str],
+    exclude: Iterable[str],
+) -> list[Path]:
+    firmwared = Path("usr/lib/firmware")
+    with chdir(root):
+        all_firmware = set(firmwared.rglob("*"))
```
DaanDeMeyer, line 67: We should move this inside of the `if include:` block to optimize things if `FirmwareInclude=` is not set.

mkosi/kmod.py
```diff
@@ -195,6 +231,12 @@ def gen_required_kernel_modules(
             mods = set(modulesd.rglob("*.ko*"))
         firmware = set()
 
+    # Include or exclude firmware explicitly configured
+    if firmware_include or firmware_exclude:
```
DaanDeMeyer, line 235: Let's move this condition inside of `filter_firmware()`

mkosi/kmod.py
```diff
@@ -55,6 +55,39 @@ def filter_kernel_modules(
...
+        remove = set()
+        regex = re.compile("|".join(exclude))
+        for f in firmware:
+            rel = os.fspath(Path(*f.parts[3:]))
+            if rel not in keep and regex.search(rel):
+                remove.add(f)
+
+        firmware -= remove
+
+    return sorted(firmware)
```
DaanDeMeyer, line 88: Sorting here when we're immediately going to turn it into an (unordered) set afterwards is just wasteful. Let's immediately return a set from this function instead.

mkosi/kmod.py
```diff
@@ -55,6 +55,39 @@ def filter_kernel_modules(
...
+
+    if exclude:
+        remove = set()
+        regex = re.compile("|".join(exclude))
+        for f in firmware:
+            rel = os.fspath(Path(*f.parts[3:]))
+            if rel not in keep and regex.search(rel):
+                remove.add(f)
+
+        firmware -= remove
```
DaanDeMeyer, line 86: I would switch the operations around, first do the `exclude` logic, and then the `include` logic. It doesn't matter too much if we remove stuff first that we'll add back later since insert/remove from a set is a constant time operation anyway.

Switching this around allows you to get rid of the `keep` variable.

mkosi/resources/man/mkosi.1.md
```diff
@@ -1047,6 +1047,17 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 `KernelModulesInitrdExclude=`, `--kernel-modules-initrd-exclude=`
 :   Like `KernelModulesExclude=`, but applies to the kernel modules included in the kernel modules initrd.
 
+`FirmwareInclude=`, `--firmware-include=`
+:   Takes a list of regex patterns that specify firmware files to include in the image. Patterns should be
+    relative to `/usr/lib/firmware/<subdir>` paths. **mkosi** checks for a match anywhere in the firmware path
+    (e.g. `bcm8483` will match against `cxgb4/bcm8483.bin`). All firmware files that match any of the specified
+    patterns are included in the image.
+
+`FirmwareExclude=`, `--firmware-exclude=`
```
DaanDeMeyer, line 1056: We should mention here that firmware is excluded even if an included module depends on it.

DaanDeMeyer (changes requested):

mkosi/kmod.py
behrmann, line 88: While this is symmetric with `filter_kernel_modules`, this goes through `resolve_module_dependencies`, which makes `firmware` a set and with this here, we sort it, to directly throw this away again by making it a set at the call site and sorting it again a bit later. You can return `firmware` here and make the function return `set[Path]` instead.

mkosi/config.py
aafeijoo-suse: Sorry, I completely misunderstood https://github.com/systemd/mkosi/pull/3483#discussion_r1946223262. It should be ok now.

mkosi/kmod.py
```diff
@@ -55,6 +55,39 @@ def filter_kernel_modules(
...
+    firmware: set[Path],
+    *,
+    include: Iterable[str],
+    exclude: Iterable[str],
+) -> set[Path]:
+    if not include and not exclude:
+        return firmware
+
+    if exclude:
+        remove = set()
```
DaanDeMeyer, line 69: This isn't required, you can just remove from `firmware` directly instead of gathering in a separate set first

DaanDeMeyer (changes requested):

mkosi/kmod.py
aafeijoo-suse: You cannot remove elements from `firmware` while you're iterating it... see below

DaanDeMeyer: Ahah, indeed, nvm

DaanDeMeyer (approved):

aafeijoo-suse:
Thanks to both of you!

PR #3486, opened 2025-02-08
Start systemd-storagetm in mkosi serve as well if available

DaanDeMeyer:
Allows accessing a disk image over NVME-TCP instead of having to download it completely over HTTP.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -13,6 +13,7 @@ import re
 import resource
 import shlex
 import shutil
+import signal
 import socket
 import stat
 import subprocess
@@ -107,6 +108,7 @@ from mkosi.run import (
     finalize_passwd_symlinks,
     fork_and_wait,
     run,
+    spawn,
     workdir,
 )
 from mkosi.sandbox import (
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -4244,16 +4246,45 @@ def run_coredumpctl(args: Args, config: Config) -> None:
...
+                        setup=become_root_cmd(),
+                    ),
+                    foreground=True,
+                )
+            )
+
+            storagetm.wait()
+            http.send_signal(signal.SIGINT)
+        else:
+            http.wait()
```
behrmann, line 4287: http gets sent `SIGINT` in the if-branch, but everything gets torn right afterwards, would this maybe make more sense
```suggestion

        http.wait()
```

behrmann (approved):
LGTM, just one question.

mkosi/__init__.py
DaanDeMeyer: We always wait for the process to finish in spawn() unless an exception is thrown, so no need to wait in this case, though it doesn't hurt either

PR #3488, opened 2025-02-09
Add ToolsTreePrepareScripts=

DaanDeMeyer:
Let's give users a way to make arbitrary modifications to the default tools tree.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -4387,6 +4387,9 @@ def finalize_default_tools(config: Config, *, resources: Path) -> Config:
         "--incremental", str(config.incremental),
         *([f"--package={package}" for package in config.tools_tree_packages]),
         *([f"--package-directory={os.fspath(directory)}" for directory in config.tools_tree_package_directories]),  # noqa: E501
+        *([f"--build-sources={tree}" for tree in config.build_sources]),
+        "--build-sources-ephemeral", str(config.build_sources_ephemeral),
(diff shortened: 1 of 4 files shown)


mkosi/__init__.py
```diff
@@ -4387,6 +4387,9 @@ def finalize_default_tools(config: Config, *, resources: Path) -> Config:
         "--incremental", str(config.incremental),
         *([f"--package={package}" for package in config.tools_tree_packages]),
         *([f"--package-directory={os.fspath(directory)}" for directory in config.tools_tree_package_directories]),  # noqa: E501
+        *([f"--build-sources={tree}" for tree in config.build_sources]),
```
behrmann, line 4390: The build sources here are needed to have them be usable by the prepare script, right?

DaanDeMeyer: Yeah

septatrix:
Would that work for the use-case in matrix? The tools tree packages would still be installed before the prepare script ran so the certificates for the mirror are not yet installed. Wouldn't it rather require something akin to `ToolsTreeSkeleton=`?

DaanDeMeyer:
> Would that work for the usecase in matrix? The tools tree packages would still be installed before the prepare script ran so the certificates for the mirror are not yet installed. Wouldn't it rather require something akin to `ToolsTreeSkeleton=`?
> 
> Might also be that I did not completely grok his usecase

The default tools tree is always built with the host certificates, you can then use a combo of `ToolsTreeSandboxTrees=` and a prepare script to get the CA into the tools tree.

PR #3493, opened 2025-02-10
serve: Add StorageTargetMode= feature

DaanDeMeyer:
Let's make it configurable whether we run systemd-storagetm as part
of serve.


mkosi/qemu.py
```diff
@@ -1563,13 +1564,19 @@ def add_virtiofs_mount(
                 ),
             ),
         ) as proc:
+            # Immediately suspend the child process so we can optionally register the VM with
+            # systemd-machined before continuing.
+            proc.send_signal(signal.SIGSTOP)
```
bluca, line 1575: Isn't systemctl freeze/thaw better, so that the whole scope is frozen at the same time?

DaanDeMeyer: I don't think there's much point in freezing the scope when we know there's only one process in it. Even if there were more, the qemu one is the only one writing to stdout so it should be fine to just freeze that one.

bluca: There might be more in the future - but also it's a good chance to dogfood these APIs too

DaanDeMeyer: I had to rework this and now we don't send SIGSTOP from mkosi itself anymore, so there's no opportunity to use systemctl freeze anymore here.

mkosi/resources/man/mkosi.1.md
```diff
@@ -1850,6 +1850,15 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     of file if your workload produces more than `4G` worth of journal
     data.
 
+`StorageTargetMode=`, `--storage-target-mode=`
+:   Specifies whether the `serve` verb should start systemd-storagetm to
```
behrmann, line 1854:
```suggestion
:   Specifies whether the `serve` verb should start **systemd-storagetm** to
```

mkosi/resources/man/mkosi.1.md
```diff
@@ -1850,6 +1850,15 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
+    serve disk images over NVME-TCP. Takes a boolean value or `auto`. If
+    enabled, systemd-storagetm is always started and mkosi will fail it
```
behrmann, line 1856:
```suggestion
    enabled, systemd-storagetm is always started and mkosi will fail if
```

behrmann (approved):

PR #3496, opened 2025-02-11
mkosi-sandbox: Implement --pack-fds

DaanDeMeyer:
This allows us to move the file descriptor packing logic from spawn() to mkosi-sandbox. The main advantage here is that we can pass file descriptors now without necessarily packing them together, which we now only do for systemd-journal-remote which requires it.


mkosi/sandbox.py
```diff
@@ -438,6 +443,67 @@ def is_relative_to(one: str, two: str) -> bool:
     return os.path.commonpath((one, two)) == two
 
 
+def pack_file_descriptors() -> int:
+    fds = []
+
+    with os.scandir("/proc/self/fd") as it:
```
behrmann, line 449: The whole thing needs the list to be sorted, right? Maybe I'm too paranoid, but does scandir guarantee the order to be lexicographical? If not, maybe a `sorted` afterwards wouldn't hurt?

mkosi/sandbox.py
```diff
@@ -438,6 +443,67 @@ def is_relative_to(one: str, two: str) -> bool:
...
+
+            if fd < SD_LISTEN_FDS_START:
+                continue
+
+            fds.append(fd)
+
+    # os.scandir() either opens a file descriptor to the given path or dups the given file descriptor. Either
+    # way, there will be an extra file descriptor in the fds array that's not valid anymore now, so find out
+    # which one and drop it.
+    fds = [fd for fd in fds if libc.fcntl(fd, F_GETFD, 0) >= 0]
```
behrmann, line 467: Since I just looked at the scandir docs, it mentions that since 3.7 file descriptors are supported, so maybe we could open `/proc/self/fd` ourselves and then right away which fd not to add?

behrmann (approved):

mkosi/sandbox.py
DaanDeMeyer: @behrmann No that doesn't work, because scandir() will dup that file descriptor so we just move the problem

mkosi/sandbox.py
```diff
@@ -438,6 +443,67 @@ def is_relative_to(one: str, two: str) -> bool:
...
```
behrmann, line 467: Doing the `sorted` call here instead would sidestep the need for a key function.
```suggestion
    fds = sorted(fd for fd in fds if libc.fcntl(fd, F_GETFD, 0) >= 0)
```

PR #3499, opened 2025-02-12
Allow building "netesp" images with the esp output format.

DaanDeMeyer:
Coined by @poettering, netesp images only have a boot loader in them and nothing else and should use raw.img as their extension, so let's add the necessary knobs to enable this use case.


behrmann:
Looks reasonable, but the JSON serialisation is missing.

PR #3502, opened 2025-02-13
Stop using nargs="?" for options

DaanDeMeyer:
This allows us to get rid of the ambiguity when parsing the verb which
could be interpreted as the argument of the previous option by argparse.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -1517,7 +1517,6 @@ class ConfigSetting(Generic[T]):
     long: str = ""
     choices: Optional[list[str]] = None
     metavar: Optional[str] = None
-    nargs: Optional[str] = None
     const: Optional[Any] = None
     help: Optional[str] = None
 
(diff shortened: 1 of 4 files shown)


mkosi/config.py
```diff
@@ -4562,6 +4562,10 @@ def parse_config(
             continue
 
         if v_i > 0 and argv[v_i - 1] != "--":
+            # If -- is already in the command line after the verb, we have to remove it.
+            if "--" in argv[v_i:]:
+                del argv[v_i + argv[v_i:].index("--")]
+
```
behrmann, line 4568: I'm not sure I understand the semantics here. E.g. for `journalctl` we document since yesterday
> All arguments specified after the `journalctl` verb and separated by `--` from the regular options are appended to the **journalctl** invocation.

but with this change we'd effectively always move the `--` before the verb:
```
>>> argv = ["mkosi", "--mkosi1", "journalctl", "--mkosi2", "--", "--journalctl1"]
>>> v_i = argv.index("journalctl")
>>> del argv[v_i + argv[v_i:].index("--")]
>>> argv
['mkosi', '--mkosi1', 'journalctl', '--mkosi2', '--journalctl1']
>>> argv.insert(v_i, "--")
>>> argv
['mkosi', '--mkosi1', '--', 'journalctl', '--mkosi2', '--journalctl1']
```

PR #3505, opened 2025-02-13
Fix tty resize with vmspawn as virtual machine manager

hundeboll:
A default console is configured  on the kernel commandline by systemd-vmspawn, and the value depends on the booted architecture. The rows/columns configuration by mkosi must match that console device to have an effect.


DaanDeMeyer:
Let's not do this, I recently fixed vmspawn to use hvc0 as well

DaanDeMeyer:
Let's close this, we should standardize on hvc0 instead

PR #3508, opened 2025-02-13
docs: Various fixes

DaanDeMeyer:
- Make sure we suffix all long CLI options with '='
- Make sure whenever we use a boolean option that we pass an argument
- Make sure we prefix with '--' whenever we pass extra arguments


DaanDeMeyer:
Should all be uncontroversial so let's merge

PR #3509, opened 2025-02-13
Add BuildKey= and CacheKey= settings

DaanDeMeyer:
Let's give users more control over how many different cache and
build subdirectories we maintain by introducing CacheKey= and
BuildKey= with support for delayed specifiers.


mkosi/config.py
```diff
@@ -1975,6 +1977,49 @@ class Config:
...
+        if self.proxy_peer_certificate:
+            env["GIT_PROXY_SSL_CAINFO"] = "/proxy.cacert"
+        if self.proxy_client_certificate:
+            env["GIT_PROXY_SSL_CERT"] = "/proxy.clientcert"
+        if self.proxy_client_key:
+            env["GIT_PROXY_SSL_KEY"] = "/proxy.clientkey"
+        if dnf := os.getenv("MKOSI_DNF"):
+            env["MKOSI_DNF"] = dnf
+        if gnupghome := os.getenv("GNUPGHOME"):
+            env["GNUPGHOME"] = gnupghome
```
behrmann: If think this change i a good cleanup, but we should document the undocumented variables:
- the four variables from `env` should just be noted, that mkosi sets them for internal use, the same should maybe be done for `MKOSI_DNF` or maybe all `MKOSI_` variables? 
- `SUBIMAGE` is not yet documented (and used?)
- all proxy variables are not documented
- that we forward `GNUPGHOME` from the environment is also undocumented

mkosi/resources/man/mkosi.1.md
```diff
@@ -1441,6 +1440,29 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     found in the local directory it is automatically used for this
     purpose.
 
+`CacheKey=`, `--cache-key=`
+:   Specifies the subdirectory within the cache directory where to store
+    the cached image. This may include both the regular specifiers (see
+    **Specifiers**) and special delayed specifiers, that are expanded
+    after config parsing has finished instead of during config parsing,
```
behrmann, line 1447:
```suggestion
    after config parsing has finished, instead of during config parsing,
```

mkosi/resources/man/mkosi.1.md
```diff
@@ -1459,6 +1481,19 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     in the local directory it is automatically used for this purpose (also
     see the **Files** section below).
 
+`BuildKey=`, `--build-key=`
+:   Specifies the subdirectory within the build directory where to store
+    incremental build artifacts. This may include both the regular
+    specifiers (see **Specifiers**) and special delayed specifiers, that
+    are expanded after config parsing has finished instead of during
```
behrmann, line 1488:
```suggestion
    are expanded after config parsing has finished, instead of during
```

mkosi/config.py
```diff
@@ -2128,8 +2178,39 @@ def cache_manifest(self) -> dict[str, Any]:
...
+            if (specifier := specifiers.get(m)) is not None:
+                return specifier
+
+            logging.warning(f"Unknown specifier '&{m}' found in {text}, ignoring")
+            return ""
+
+        result = re.sub(r"&(?P<specifier>[&a-zA-Z])", replacer, text)
+
+        # Strip separator characters at the end and remove duplicate separate characters in the middle.
+        return re.sub(r"([~\-_])\1+", "\\1", result).strip("~-_")
```
behrmann, line 2203: If you move `-` to either end, it doesn't need to be escaped, since it's no longer a character range
```suggestion
        return re.sub(r"([~_-])\1+", "\\1", result).strip("~-_")
```

behrmann: That being said, what's the idea behind the stripping? It works as intended when people don't mix separators, but e.g. `~-~` would not get shortened. I think keeping the separators would actually be helpful, since then even if image ID or version expand to an empty string, this is preserved when resplitting the key, e.g.:
`distribution, image_id, image_version, release = 'foo---bar'.split('-')`

mkosi/config.py
DaanDeMeyer: Most of these are just for internal use and users don't need to know about them I think, let's leave this for another day.

mkosi/config.py
DaanDeMeyer: This actually isn't needed anymore, dropping

mkosi/resources/man/mkosi.1.md
```diff
@@ -1459,6 +1480,18 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     in the local directory it is automatically used for this purpose (also
     see the **Files** section below).
 
+`BuildKey=`, `--build-key=`
+:   Specifies the subdirectory within the build directory where to store
+    incremental build artifacts. This may include both the regular
+    specifiers (see **Specifiers**) and special delayed specifiers, that
+    are expanded after config parsing has finished, instead of during
+    config parsing, which are the same delayed specifiers that are
+    supported by `CacheKey=`The default format for this parameter is
```
behrmann, line 1489:
```suggestion
    supported by `CacheKey=`. The default format for this parameter is
```

mkosi/config.py
```diff
@@ -2107,6 +2164,16 @@ def outputs(self) -> list[str]:
...
 
+    @property
+    def build_subdir(self) -> Path:
+        assert self.build_dir
+        subdir = self.expand_key_specifiers(self.build_key)
+
+        if subdir == "-":
+            return self.build_dir
+
+        return self.build_dir / subdir
```
behrmann, line 2175: Do we actually need the extra semantic for `-` here? If subdir is the empty string, this would naturally fall back to the build directory and we don't prevent an empty build key, or do we?

mkosi/__init__.py
```diff
@@ -5086,15 +5086,24 @@ def run_verb(args: Args, images: Sequence[Config], *, resources: Path) -> None:
...
             )
 
         if not last.repart_offline and os.getuid() != 0:
-            die(f"Must be root to build {last.name()} image configured with RepartOffline=no")
+            die(f"Must be root to build {last.image} image configured with RepartOffline=no")
 
         check_workspace_directory(last)
 
+        if last.incremental:
+            for a, b in itertools.permutations(images, 2):
```
behrmann, line 5099: Nit: After checking `a == b` we don't need to check `b == a`
```suggestion
            for a, b in itertools.combinations(images, 2):
```

mkosi/__init__.py
```diff
@@ -5086,15 +5086,24 @@ def run_verb(args: Args, images: Sequence[Config], *, resources: Path) -> None:
...
+                if a.expand_key_specifiers(a.cache_key) == b.expand_key_specifiers(b.cache_key):
+                    die(
+                        f"Image {a.image} and {b.image} have the same cache key "
+                        f"'{a.expand_key_specifiers(a.cache_key)}'",
```
behrmann, line 5103: Nit:
```suggestion
                        f"Image {a.image} and {b.image} have the same cache key a.expand_key_specifiers(a.cache_key)}'",  # noqa: E501
```

mkosi/__init__.py
```diff
@@ -1330,11 +1334,22 @@ def finalize_default_initrd(
...
+
+    initrd = dataclasses.replace(initrd, image="default-initrd")
 
-    run_configure_scripts(config)
+    if initrd.incremental and initrd.expand_key_specifiers(initrd.cache_key) == config.expand_key_specifiers(
+        config.cache_key
+    ):
+        die(
+            f"Image '{config.image}' and its default initrd image have the same cache key "
+            f"'{config.expand_key_specifiers(config.cache_key)}'",
```
behrmann, line 1348: Nit:
```suggestion
            f"Image '{config.image}' and its default initrd image have the same cache key '{config.expand_key_specifiers(config.cache_key)}'",  # noqa: E501
```

behrmann (approved):

mkosi/config.py
DaanDeMeyer: Assigning the empty string to a setting resets it back to the default value. That's why I added this.

PR #3513, opened 2025-02-15
Couple of OBS fixes for ParticleOS

bluca:


mkosi/resources/mkosi-obs/mkosi.conf
```diff
@@ -2,8 +2,18 @@
 [Build]
 SandboxTrees=/usr/src/packages/SOURCES:/usr/src/packages/SOURCES
 
+[Build]
```
DaanDeMeyer, line 5: Don't repeat the section here

mkosi/resources/mkosi-obs/mkosi.conf
```diff
@@ -2,8 +2,18 @@
...
+ToolsTree=
+History=no
+CacheDirectory=
+Incremental=no
+
 [Output]
+OutputDirectory=
```
DaanDeMeyer, line 12: We should set `Checksum=yes` here as well, I think we want to move towards a world where all the OBS specific configuration is in here and not in OBS itself?

DaanDeMeyer (changes requested):

mkosi/__init__.py
```diff
@@ -2128,6 +2128,9 @@ def install_uki(
             boot_binary,
         )
 
+        if ArtifactOutput.pcrs in context.config.split_artifacts and pcrs:
+            (context.staging / context.config.output_split_pcrs).write_text(json.dumps(pcrs))
```
DaanDeMeyer, line 2132: This doesn't work if there's more than one UKI in the image, we'll just overwrite the pcrs file again and again (and we loop over kernels from highest to lowest version). We should handle this in `install_kernel()` and only keep the pcrs file for the UKI with the highest version (the first one) as that is the one we extract as a split artifact.

DaanDeMeyer (changes requested):

mkosi/__init__.py
bluca: If there are multiple ones and only one gets signed it's still problematic though - is this something that happens in practice?

mkosi/resources/mkosi-obs/mkosi.conf
DaanDeMeyer: And the other ones from obs-build as well:

```
        --repository-key-check=no \
        --with-network=never \
        --local-mirror file:///.build.binaries/ \
        --cache /.build.binaries/ \
```

mkosi/__init__.py
bluca: I'll start with processing the first one as suggested, and added a TODO to consider the case of multiple ukis in the future

mkosi/__init__.py
```diff
@@ -2230,16 +2234,23 @@ def install_kernel(context: Context, partitions: Sequence[Partition]) -> None:
     token = find_entry_token(context)
     cmdline = finalize_cmdline(context, partitions, finalize_roothash(partitions))
     profiles = build_uki_profiles(context, cmdline) if want_uki(context) else []
+    # The first processed UKI is the one that will be used as split artifact, so take pcrs from
+    # it and ignore the rest
+    # TODO: we should probably support signing pcrs for all built UKIs
+    pcrs = None
```
DaanDeMeyer, line 2240: Can be emtpy dict as well

DaanDeMeyer (approved):

PR #3520, opened 2025-02-17
mkosi-obs: use compression and update ESP when DDIs are built with UKIs

bluca:


mkosi/resources/mkosi-obs/mkosi.build
```diff
@@ -58,6 +58,23 @@ while read -r SIG; do
...
+    if [ -f "${DEST%efi}raw" ] || [ -f "${DEST%efi}raw.zst" ]; then
+        if [ -f "${DEST%efi}raw.zst" ]; then
+            unzstd "${DEST%efi}raw.zst"
+        fi
+        offset="$(systemd-repart --json=short "${DEST%efi}raw" | jq -r '.[] | select(.type == "esp") | .offset')"
+        if [ "$offset" = "null" ]; then
+            rm -f "${DEST%efi}raw"
+            continue
+        fi
+        mcopy -o -i "${DEST%efi}raw@@${offset}" "$DEST" "::EFI/Linux/$(basename "$DEST")"
```
DaanDeMeyer, line 72: So this is problematic. To make sysupdate work we need to follow the naming format from the match pattern of the sysupdate file. This looks as follows:

```
MatchPattern=%M_@v+@l-@d.efi \
             %M_@v+@l.efi \
             %M_@v.efi
```

I think for starters naming it `<image-id>_<image-version>.efi` in the ESP is fine. That's information which is easily available from the os-release file. We can figure out how to implement boot counting later.

DaanDeMeyer (changes requested):

mkosi/resources/mkosi-obs/mkosi.build
bluca: the name isn't decided here? It's what mkosi spits out

bluca: ie this script is just using whatever filename is decided by the mkosi build, it just looks at suffixes and the output dir

bluca: ie, this is not a new uki being added, it's replacing the existing one - hence the `-o` parameter to mcopy, to overwrite the existing file

DaanDeMeyer: Yes mkosi spits out according to `Output=` that's OK because that UKI is intended for consumption by systemd-sysupdate which will give it the proper name when installing it to the ESP. Here we're not using systemd-sysupdate to install the UKI to the ESP so you're responsible to give it the proper name.

bluca: IE before this change, the mkosi-built DDI `ParticleOS-x86-64_3.6.raw` contains `/efi/EFI/Linux/ParticleOS-x86-64_3.6.efi`

DaanDeMeyer: Are you sure you're on the latest version of mkosi and particleos?

```
-bash-5.2# ls /efi/EFI/Linux/
ParticleOS_20250217145022.efi
```

bluca: mkosi yes, particle maybe not, rebuilding

DaanDeMeyer (approved):

PR #3523, opened 2025-02-18
Drop boot counting support

DaanDeMeyer:
Setting up boot counting for the initial UKI in an image does not make sense as there's nothing to fall back to. The existing interface where we pick up the number of tries from a file in /etc inside the image was also rather bad so let's get rid of the boot counting specifier.


DaanDeMeyer:
cc @NekkoDroid

behrmann (approved):

bluca:
Boot counting makes sense for updates being installed on an existing system, but for an initial ESP yeah there's not much point

behrmann:
For posterity: I wondered the same, but it's handled by sysupdate, so mkosi doesn't need to.

PR #3527, opened 2025-02-19
mkosi-tools: Split off misc, package-manager and runtime profiles

DaanDeMeyer:
Let's make the dependencies verb a little more flexible by splitting
off various profiles in mkosi-tools, the misc, package-manager and
runtime profiles. The misc profile contains all the useful tools we
install to make sure they're available in scripts. The package manager
contains non-native package managers and related tools. The runtime
profile contains all the tools required to boot images. The regular
configuration now only includes the tools required to build native images
with mkosi.

The default tools tree still includes both the misc, package-manager and
runtime profiles by default but the list of profiles can be changed with
the new ToolsTreeProfiles= setting. The dependencies verb includes neither
profile, but we now allow passing arbitrary settings to the dependency verb
as arguments so that packages from additional profiles can be requested if
needed.


bluca (approved):

mkosi/__init__.py
```diff
@@ -4945,11 +4946,11 @@ def run_verb(args: Args, images: Sequence[Config], *, resources: Path) -> None:
 
     if args.verb == Verb.dependencies:
         _, [deps] = parse_config(
-            ["--directory", "", "--repositories", "", "--include=mkosi-tools", "build"],
+            ["--directory", "", "--repositories", "", *args.cmdline, "--include=mkosi-tools", "build"],
```
behrmann, line 4949: Looking at this from a completions point of view, this is slightly terrifying. :)

Do we need the full power of being able to pass in any mkosi argument or can we make this 
```suggestion
            ["--directory", "", "--repositories", "", *(f"--profile={p}" for p in args.cmdline), "--include=mkosi-tools", "build"],
```
This way one also doesn't need to pass `--profile` and it's just arguments.

behrmann (approved):

mkosi/__init__.py
DaanDeMeyer: OK makes sense let me change this

DaanDeMeyer: Actually, this would also prevent doing something like `mkosi dependencies -- -d debian` to get the dependencies of debian on Fedora. I figure we may want to support that as well.

PR #3529, opened 2025-02-19
Bump Fedora initrd size

DaanDeMeyer:
Fixes CI failure because the Fedora initrd increased in size again.


behrmann (approved):

behrmann:
In the fedora:ubuntu paing it still failed the initrd size test

PR #3531, opened 2025-02-19
Add new option syntax with globs

keszybz:
I think we want to make the switch. This is the first part, conversion of `KernelModules{Initrd,}{Include,Exclude}`. I didn't do the docs. I'll add that if we have general agreement that this makes sense. After doing the conversion, I find the new syntax much easier to write and easier to get correct.

There are some other options that take globs, those should probably be converted for consistency too.

I also think that the design of Include/Exclude should be changed to be like in systemd. If only Include is specified, then the default for Exclude is `*`. That way, we avoid one line in the common case. We can use the new rules for the new set of options. I didn't implement that part yet.

I think that having new options is a requirement to maintain backward compatibility. I think the old names should be deprecated and dropped after a few releases. After that, we can rename the options back, keeping the new extra-long names for compatibility.

I also considered `KernelModules{Initrd,}{Add,Remove}`. This would avoid the extra-long names, but it'd create inconsistency with naming in other places. Suggestions are welcome.


mkosi/kmod.py
```diff
@@ -92,6 +123,19 @@ def normalize_module_name(name: str) -> str:
     return name.replace("_", "-")
 
 
+def normalize_module_glob(name: str) -> str:
+    # We want to replace '_' by '-', except when used in […]
```
behrmann, line 127: That doesn't treat `_` and `-` equivalently, it only treats all `_` as `-`, but not the other way around? `intel-vsec` is a module where both `modinfo intel-vsec` and `modinfo intel_vsec` work, but the name is `intel_vsec`. I'm not sure it's a good idea to replace the correct name by something that also works, this seems like something that's bound to break. Maybe this should replace both `-` and `_` by `[-_]`?

mkosi/kmod.py
```diff
@@ -17,10 +18,25 @@
...
     # Loaded modules are listed with underscores but the filenames might use dashes instead.
-    return [
-        rf"/{line.split()[0].replace('_', '[_-]')}\.ko"
-        for line in Path("/proc/modules").read_text().splitlines()
-    ]
+    return [normalize_module_name(line.split()[0])
+            for line in Path("/proc/modules").read_text().splitlines()]
+
+
+def globs_match_module(name, globs):
```
behrmann, line 25: Just for completeness, did you consider simply postprocessing the globs to regex via `fnmatch.translate`? That would keep the rest more or less the same?

behrmann (commented):
I like moving this to globs! A few notes below. This also needs docs.

septatrix:
What would be the glob equivalents of the following includes?

```
^ide 
^drivers/net/(?!wireless/)
^fs/(?!(smb|nfs|nls)/)
^drivers/gpu/drm/(intel|amd)
```

DaanDeMeyer:
cc @aafeijoo-suse this might interest you as well

aafeijoo-suse:
ugh... I agree that this change itself makes sense, the glob suffix is more appropriate, but... at this point for v26? these type of breaking changes are more usual on early releases... anyway, if you are going to change this, the sooner the better.
IIUC after this first step, there will be another one that provides the current options (without the glob suffix) that do not accept globs, right?

aafeijoo-suse:
> I don't think that's a breaking change.

I disagree, with this PR the options without the `Glob` suffix stop working, so users will need to manually adapt their configs. Anyway, I'm not saying I'm against the change.

behrmann:
> I disagree, with this PR the options without the Glob suffix stop working, so users will need to manually adapt their configs. Anyway, I'm not saying I'm against the change.

I haven't run the code beyond the summary, but
```
python -m mkosi --kernel-modules-include-glob='foo*' --kernel-modules-include=^bar summary
```
produces
```
             Kernel Modules Include: foo*
                                     re:^bar
```
and the code checks both
```
            if (
                (patterns and regex.search(rel))
                or globs_match_module(normalize_module_name(rel), globs)
            ):
```
after sorting the include into globs and regex. Am I missing something?

septatrix:
Also if this gets added I think it should also be added for the `Firmware*=` regex options at the same time

mkosi/kmod.py
septatrix: It's also what Python would do behind the scene anyhow

keszybz:
> ugh... I agree that this change itself makes sense, the glob suffix is more appropriate, but... at this point for v26? these type of breaking changes are more usual on early releases...

Really? The first sentence of the commit message says that the change is backward compatible (and how the patch is made more complicated by the need to maintain compatibility).

The cover letter also discussed this…

septatrix:
> and how the patch is made more complicated by the need to maintain compatibility.

Shouldn't usage of `fnmatch.translate` pretty much eliminate any complications? I see that you want some special behavior (which I am not sure if I like that because it might be unexpected) but replacing `re.search` with `re.match` or simply prepending and appending `$/^` should also solve that

aafeijoo-suse:
> Really? The first sentence of the commit message says that the change is backward compatible (and how the patch is made more complicated by the need to maintain compatibility).
> 
> The cover letter also discussed this…

My bad, it's ok. I should've reviewed the whole config.py, not just the changes.

mkosi/config.py
```diff
@@ -3980,6 +4028,9 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
 ]
 SETTINGS_LOOKUP_BY_NAME = {name: s for s in SETTINGS for name in [s.name, *s.compat_names]}
 SETTINGS_LOOKUP_BY_DEST = {s.dest: s for s in SETTINGS}
+SETTINGS_LOOKUP_BY_OPTION = {name: s
+                             for s in SETTINGS
+                             for name in [s.long, *s.compat_longs, s.short] if name}
```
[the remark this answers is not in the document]
keszybz: That's __much__ harder to read :(

mkosi/kmod.py
keszybz: The "normalization" only needs to be done in one direction, as long as it's the same direction. We call `normalize_module_glob()` on the globs, and `normalize_module_name()` on module names. They both change `_` to `-`, and then we get a match. This change is not visible to the user, except that patterns with both characters match module names. This is what the kernel implements too.

The normalization happens in filtering, but the actual path that is used is the path found on the file system, i.e. the actual module name.

mkosi/kmod.py
keszybz: I don't see any particular reason to do this. This code runs fast enough to not be noticeable. And in the longer term I hope to get rid of all the regexp-based patterns.

keszybz:
> What would be the glob equivalents of the following includes?
> 
> ```
> ^ide 
> ^drivers/net/(?!wireless/)
> ^fs/(?!(smb|nfs|nls)/)
> ^drivers/gpu/drm/(intel|amd)
> ```

```
SomethingInclude=
  /ide/*
  /drivers/net/*
  /fs/*
  /drivers/gpu/drm/intel/*
  /drivers/gpu/drm/amd/*

SomethingExclude=
  /drivers/net/wireless/*
  /fs/smb/*
  /fs/nfs/*
  /fs/nls/*
```

Looking at this example, I guess we could change it so that the trailing `*` is not required on directories. Obviously, `foo/*` must be meant, since `foo/` cannot match a module name.

septatrix:
> > What would be the glob equivalents of the following includes?
> > ```
> > ^ide 
> > ^drivers/net/(?!wireless/)
> > ^fs/(?!(smb|nfs|nls)/)
> > ^drivers/gpu/drm/(intel|amd)
> > ```
> 
> ```
> SomethingInclude=
>   /ide/*
>   /drivers/net/*
>   /fs/*
>   /drivers/gpu/drm/intel/*
>   /drivers/gpu/drm/amd/*
> 
> SomethingExclude=
>   /drivers/net/wireless/*
>   /fs/smb/*
>   /fs/nfs/*
>   /fs/nls/*
> ```
> 
> Looking at this example, I guess we could change it so that the trailing `*` is not required on directories. Obviously, `foo/*` must be meant, since `foo/` cannot match a module name.

But includes override excludes don't they? So the fs and wireless stuff wouldn't work

mkosi/kmod.py
```diff
@@ -17,10 +18,25 @@
...
+    # Strip '.ko' suffix and an optional compression suffix
+    name = re.sub(r'\.ko(\.[a-z0-9]+)?$', '', name)
```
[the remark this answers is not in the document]
keszybz: I checked the kernel config, and only gzip, xz, and zstd are supported. So we can enumerate them without trouble.

keszybz:
Updated version.

To solve the issue that @septatrix described, the globs can include negative patterns. See the commit description for details.

Firmware list is given the same treatment as module list.

The man page and news are updated now too.

mkosi/config.py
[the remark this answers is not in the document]
behrmann: Ok, but then let's not deviate from the style of the rest totally and go with one of these version
```
SETTINGS_LOOKUP_BY_OPTION = {
    name: s 
    for s in SETTINGS 
    for name in [s.long, *s.compat_longs, s.short] if name
}  # fmt: skip
```
or
```
SETTINGS_LOOKUP_BY_OPTION = {
    name: s for s in SETTINGS for name in [s.long, *s.compat_longs, s.short] if name
}
```

mkosi/kmod.py
```diff
@@ -18,41 +20,101 @@
...
+        if (
+            # match the full path
+            (glob.startswith("/") and fnmatch.fnmatch(f"/{name}", glob))
+            or
+            # match a subset of the path, at path element boundary
+            ("/" in glob and fnmatch.fnmatch(f"/{name}", f"*/{glob}"))
+            or
+            # match the basename
+            fnmatch.fnmatch(name.split("/")[-1], glob)
+        ):
```
behrmann, line 49: This can be compressed by not having the comments between the or and the following predicate, that these are ored is already sufficiently highlighted from it leading the line.
```suggestion
        if (
            # match the full path
            (glob.startswith("/") and fnmatch.fnmatch(f"/{name}", glob))
            # match a subset of the path, at path element boundary            
            or ("/" in glob and fnmatch.fnmatch(f"/{name}", f"*/{glob}"))
            # match the basename            
            or fnmatch.fnmatch(name.split("/")[-1], glob)
        ):
```

mkosi/kmod.py
```diff
@@ -18,41 +20,101 @@
 def loaded_modules() -> list[str]:
     # Loaded modules are listed with underscores but the filenames might use dashes instead.
     return [
-        rf"/{line.split()[0].replace('_', '[_-]')}\.ko"
-        for line in Path("/proc/modules").read_text().splitlines()
+        normalize_module_name(line.split()[0]) for line in Path("/proc/modules").read_text().splitlines()
     ]
 
 
+def globs_match_filename(name, globs) -> Optional[bool]:
```
behrmann, line 27: The `Optional[bool]` return makes this a bit hard to understand, since one has to look at the users of the function to understand the difference between `False` and `None`. How about this instead
```suggestion
def globs_match_filename(name, globs, *, match_default: bool: False) -> bool:
```
and return `match_default` at the end. The option would need to be added to `glob_match_firmware`, but it would make this a simple yes/no predicate again and sidestep the question "Why is there an `is False` check there?" in the firmware inclusion code.

mkosi/kmod.py
```diff
@@ -18,41 +20,101 @@
...
+    logging.debug(f"Kernel modules exclude: {' '.join(exclude)}")
+
     modulesd = Path("usr/lib/modules") / kver
     with chdir(root):
-        modules = set(modulesd.rglob("*.ko*"))
+        # The glob may match additional paths.
+        # Narrow this down to *.ko, *.ko.gz, *.ko.xz, *.ko.zst.
+        modules = set(
+            m for m in modulesd.rglob("*.ko*") if m.name.endswith((".ko", ".ko.gz", ".ko.xz", ".ko.zst"))
+        )
```
behrmann, line 84: Nit: with the added logic this should be a set literal instead of a `set` call
```suggestion
        modules = {
            m for m in modulesd.rglob("*.ko*") if m.name.endswith((".ko", ".ko.gz", ".ko.xz", ".ko.zst"))
        }
```

mkosi/kmod.py
```diff
@@ -79,19 +148,38 @@ def filter_firmware(
...
+            if (patterns and regex.search(rel)) or globs_match_firmware(rel, globs):
                 firmware.add(f)
 
+    logging.debug(f"Including {len(firmware)} firmare files")
+
     return firmware
 
 
 def normalize_module_name(name: str) -> str:
+    # Replace '_' by '-'
```
behrmann, line 166: We generally should have more comments, but this one might be a tad too much. Maybe a few more comments could be added to the slightly less obvious function below instead? :)

mkosi/kmod.py
```diff
@@ -79,19 +148,38 @@ def filter_firmware(
...
+def normalize_module_glob(name: str) -> str:
+    # We want to replace '_' by '-', except when used in […]
+    ans = ""
+    while name:
+        i = (name + "[").index("[")
+        ans += name[:i].replace("_", "-")
+        name = name[i:]
+        i = (name + "]").index("]")
+        ans += name[:i + 1]  # fmt: skip
+        name = name[i + 1:]  # fmt: skip
```
behrmann, line 179: Please, let's not do this
```suggestion
        ans += name[: i + 1]
        name = name[i + 1 :]
```
Do I find the above a bit too airy, sure, but it's what [PEP8](https://peps.python.org/pep-0008/#pet-peeves) would recommend and we shouldn't proliferate `# fmt: skip`s beyond where they are really needed for better readability and I don't think this is it.

septatrix:
Using `~` (tilde) to indicate negative matches would be closer to how systemd does it (e.g. syscall and capabilities filters) - and their normal meaning of `$HOME` would not make sense here anyhow

mkosi/kmod.py
keszybz: This is another case where the automatic formatter does the wrong thing and we accept that. We don't actually follow PEP8 all that closely, almost nobody does. But OK, in this case the automatic version is just slightly worse, not egregious and completely unreadable like in many other cases.

keszybz:
> Using `~` (tilde) to indicate negative matches would be closer to how systemd does it (e.g. syscall and capabilities filters) - and their normal meaning of `$HOME` would not make sense here anyhow

Yeah, I know systemd uses the tilde. But that's because systemd didn't have a choice, since `-.mount` and other names starting with `-` were chosen to be valid names (causing cascading problems in usability)  and had to be taken into account when the exclusion syntax was added much later.  The tilde is not great, because it's quite inconvenient to type on various language keyboards, as you mention can possibly be confused with the home dir, and depending on the font, it can be rendered as a little small squiggle above the middle line of letters. We're not going to recreate the systemd semantics here, so we don't need to maintain bug-for-bug compat.

keszybz:
Updated with the suggested style and formatting fixes.

behrmann (approved):
I think it's fine to use `-` here instead of `~` for negation and I think this is more than good enough to iterate on if necessary. Just needs the last unneeded import removed (probably, can't check at the moment).

behrmann:
Needs a rebase, because I just merged your other PR. 

The CI failure is because if codespell, which I tend to forget we have, can you please add
```
ignore-words-list = ans
```
to `.codespellrc`? I think this is a perfectly fine shorthand to have.

keszybz:
zypper: error while loading shared libraries: libyaml-cpp.so.0.8: cannot open shared object file: No such file or directory

Non mea culpa.

behrmann (approved):

PR #3542, opened 2025-02-22
Add SplitArtifacts=metainfo

bluca:
GNOME Software parses .metainfo.xml files to provide nice metadata for artifacts it downloads/updates. Follow the spec and add parameter for it, filling the content from os-release.

https://www.freedesktop.org/software/appstream/docs/sect-Metadata-OS.html


mkosi/__init__.py
```diff
@@ -388,6 +388,64 @@ def configure_os_release(context: Context) -> None:
...
+                    if id:
+                        id_field += f".{id}"
+                    metainfo += f"  <id>{id_field}</id>\n"
+            else:
+                metainfo += f"  <id>{id}</id>\n"
+            if name:
+                metainfo += f"  <name>{name}</name>\n"
+            elif context.config.image_id:
+                metainfo += f"  <name>{context.config.image_id}</name>\n"
+            metainfo += f"  <summary>{id} image built with mkosi</summary>\n"
```
behrmann, line 425: When `id` is `None` this would give "None image built with mkosi"?

mkosi/__init__.py
```diff
@@ -388,6 +388,64 @@ def configure_os_release(context: Context) -> None:
...
+            timestamp = (
+                datetime.datetime.fromtimestamp(context.config.source_date_epoch, tz=datetime.timezone.utc)
+                if context.config.source_date_epoch is not None
+                else datetime.datetime.now(tz=datetime.timezone.utc)
+            ).isoformat()
+            metainfo += f'     <release version="{context.config.image_version}" date="{timestamp}" type="development">\n'  # noqa E501
+            metainfo += "       <description></description>\n"
+            metainfo += "     </release>\n"
+            metainfo += "  </releases>\n"
+            metainfo += "</component>\n"
```
behrmann, line 446:
```suggestion
            timestamp = (
                datetime.datetime.fromtimestamp(context.config.source_date_epoch, tz=datetime.timezone.utc)
                if context.config.source_date_epoch is not None
                else datetime.datetime.now(tz=datetime.timezone.utc)
            ).isoformat()
            metainfo += textwrap.dedent(
                """\
                  <icon type="remote">https://brand.systemd.io/assets/svg/systemd-logomark.svg</icon>
                  <metadata_license>FSFAP</metadata_license>
                  <releases>
                     <release version="{context.config.image_version}" date="{timestamp}" type="development">
                       <description></description>
                     </release>
                  </releases>
                </component>     
                """
            )
```

mkosi/resources/man/mkosi.1.md
```diff
@@ -624,7 +624,7 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 `SplitArtifacts=`, `--split-artifacts=`
 :   The artifact types to split out of the final image. A comma-delimited
     list consisting of `uki`, `kernel`, `initrd`, `os-release`, `prcs`, `partitions`,
-    `roothash` and `tar`. When building a bootable image `kernel` and `initrd`
+    `metainfo`, `roothash` and `tar`. When building a bootable image `kernel` and `initrd`
```
behrmann, line 627: `metainfo` sounds a bit generic, this is only used by appstream, isn't it? Maybe this should be `appstream-metainfo`?

behrmann (changes requested):

mkosi/__init__.py
```diff
@@ -388,6 +388,64 @@ def configure_os_release(context: Context) -> None:
...
+
+            # The <id> format is reverse DNS style with home_url + id
+            if home_url:
+                from urllib.parse import urlparse
+
+                url = urlparse(home_url)
+                if url.netloc:
+                    netloc = url.netloc.split(".")
+                    netloc.reverse()
+                    netloc.remove("www")
```
septatrix, line 414: This removes any www, not only one at the start

mkosi/__init__.py
bluca: ID is a mandatory field in os-release

mkosi/__init__.py
bluca: www is not a valid tld or domain

septatrix: But it is a valid subdomain (e.g. `distro.www.company.org`). Not sure who would do that stuff but there is no reason to not just do the correct thing.

```suggestion
                    netloc = url.netloc.removeprefix("www.").split(".")
```

mkosi/__init__.py
septatrix: No, the man page says it: "If not set, a default of "ID=linux" may be used." so not, it is not mandatory (or the man page is outdated. Either use `id or "linux"`, or add an assert

mkosi/__init__.py
```diff
@@ -385,9 +386,68 @@ def configure_os_release(context: Context) -> None:
...
+        description = f"{osrelease_content.get("PRETTY_NAME")} built with mkosi"
+    else:
+        description = "Image built with mkosi"
+    icon = context.config.appstream_icon or "https://brand.systemd.io/assets/svg/systemd-logomark.svg"
+    if context.config.appstream_homepage:
+        home_url = context.config.appstream_homepage
+    elif "HOME_URL" in osrelease_content:
+        home_url = osrelease_content.get("HOME_URL")
+    else:
+        home_url = None
```
behrmann, line 409: The get already does the same fallback.
```suggestion
    else:
        home_url = osrelease_content.get("HOME_URL")
```

mkosi/__init__.py
```diff
@@ -385,9 +386,68 @@ def configure_os_release(context: Context) -> None:
...
 
+    # https://www.freedesktop.org/software/appstream/docs/sect-Metadata-OS.html
+    osid = context.config.appstream_id or osrelease_content.get("ID", "linux")
+    name = context.config.appstream_name or osrelease_content.get("NAME") or context.config.image_id or osid
+    if context.config.appstream_description:
+        description = context.config.appstream_description
+    elif "PRETTY_NAME" in osrelease_content:
+        description = f"{osrelease_content.get("PRETTY_NAME")} built with mkosi"
+    else:
+        description = "Image built with mkosi"
```
behrmann, line 402:
```suggestion
    else:
        description = f"{osrelease_content.get('PRETTY_NAME', 'Image')} built with mkosi"
```

mkosi/config.py
```diff
@@ -575,6 +575,7 @@ class ArtifactOutput(StrEnum):
     pcrs = enum.auto()
     roothash = enum.auto()
     os_release = enum.auto()
+    metainfo = enum.auto()
```
behrmann, line 578: that doesn't match the documentation anymore
```suggestion
    appstream_metainfo = enum.auto()
```

mkosi/__init__.py
```diff
@@ -385,9 +386,68 @@ def configure_os_release(context: Context) -> None:
...
+        if not url.netloc:
+            home_url = None
+    if context.config.appstream_id:
+        id = context.config.appstream_id
+    elif home_url:
+        netloc = url.netloc.split(".")
+        netloc.reverse()
+        netloc.remove("www")
+        id = ".".join(netloc)
+        id += f".{osid}"
```
septatrix, line 421: Because it got marked as outdated due to refactorings:

> But it is a valid subdomain (e.g. distro.www.company.org). Not sure who would do that stuff but there is no reason to not just do the correct thing.

```suggestion
        netloc = url.netloc.removeprefix("www.").split(".")
        id = ".".join(reversed(netloc))
        id += f".{osid}"
```

mkosi/__init__.py
```diff
@@ -385,9 +386,68 @@ def configure_os_release(context: Context) -> None:
...
+    else:
+        id = osid
```
behrmann, line 423: This handles `home_url` in a single branch and hopefully also addresses @septatrix comment.
```suggestion
    if context.config.appstream_id:
        id = context.config.appstream_id
    elif home_url:
        url = urlparse(home_url)
        if url.netloc:
            netloc = url.netloc.removeprefix("www.").split(".")
            id = ".".join(reversed(netloc)) + f".{osid}"
        else:
            home_url = None
            id = osid
    else:
        id = osid
```

mkosi/__init__.py
bluca: there is no distribution doing anything like that

mkosi/__init__.py
bluca: there are various optional parts in the middle so line-by-line is better

mkosi/__init__.py
bluca: it's the same result, doesn't make any difference

mkosi/__init__.py
septatrix: But there is no reason to knowingly implement the wrong behavior instead of just doing it the right way (tm)

mkosi/__init__.py
behrmann: I think having the logic for `home_url` in a single place is cleaner and removing the leading `www.` with `removeprefix` is shorter, and I'd argue cleaner, with the added check whether `www` is actually in there, that is needed by `remove`.

That being said, why do you want to strip the `www` in the first place? Above you said it's not a valid tld or domain and that is true, but neither is the fallback `linux` and in the reverse url scheme, it would end up at the end, just before `osid`, anyway. Also, this is only a fallback path, people should set the appstream id in the first place (maybe using specifiers), and singling out `www` seems a bit odd. 

I generally agree that it wouldn't make sense to have `www` in an appstream ID, but neither would other subdomains like `download`, as when I grab [current systemd](https://download.opensuse.org/repositories/system:/systemd/Debian_Testing/amd64/systemd_257.999+1648+g8e0037fb3-0_amd64.deb) from OBS. Can we do without stripping `www` or what is the logic here that I don't see?

@septatrix `list.remove` only removes the first occurence, so this also behaves the same. I misremembered this as well.

septatrix: > Can we do without stripping `www` or what is the logic here that I don't see?

This is from the spec:

> This means the reversed-DNS HOME_URL value from /etc/os-release (without any https/www parts) combined with the ID of the operating system will produce the operating system's component id.

---

> @septatrix `list.remove` only removes the first occurence, so this also behaves the same. I misremembered this as well

No they do not behave the same.
It removes the first occurrence, not the occurrence iff it is at index 0. Therefore a "www" subdomain somewhere in the middle of a URL would also be stripped which I am pretty sure is not how the spec is supposed to be interpreted (because then it would likely also mean removing any "www" component, not only the first occurrence.

behrmann: Thanks @septatrix, I really shouldn't look at this before having coffee :)

PR #3543, opened 2025-02-22
Add short option for initrd kernel version

keszybz:


mkosi/initrd.py
```diff
@@ -97,6 +97,43 @@ def parse(*, name: str, description: str) -> "KernelInstallContext":
             verbose=int(os.getenv("KERNEL_INSTALL_VERBOSE", 0)) > 0,
         )
 
+def create_parser() -> argparse.ArgumentParser:
```
behrmann, line 100:
```suggestion

def create_parser() -> argparse.ArgumentParser:
```

mkosi/initrd.py
```diff
@@ -97,6 +97,43 @@ def parse(*, name: str, description: str) -> "KernelInstallContext":
...
+    parser.add_argument(
+        "-g",
+        "--generic",
+        help="Build a generic initrd without host-specific kernel modules",
+        action="store_true",
+        default=False,
+    )
+
+    initrd_common_args(parser)
+    return parser
```
behrmann, line 136:
```suggestion
    return parser

```

mkosi/initrd.py
```diff
@@ -142,6 +179,7 @@ def initrd_finalize(staging_dir: str, output: str, output_dir: str) -> None:
 def initrd_common_args(parser: argparse.ArgumentParser) -> None:
     parser.add_argument(
         "--kernel-version",
+        "-k",
```
behrmann, line 182: Could you move that before the long option? The short ones are first in the other parser definitions.

behrmann (approved):
I'm -0 to the first commit, since I find it a bit harder to read, but the other two changes are good, just a nit and a fix to CI needed.

keszybz:
Updated. I think the first commit makes sense because we already had a mix of both styles and I don't like the inconsistency. I agree that the "compressed" form can be harder to read for some forms of the value argument, but I think this is outweighed by the fact that the resulting invocation command is easier to read (and select&paste if desired).

mkosi/__init__.py
```diff
@@ -1298,57 +1298,57 @@ def finalize_default_initrd(
...
-        "--with-network", str(config.with_network),
-        "--cache-only", str(config.cacheonly),
-        *(["--output-directory", os.fspath(output_dir)] if output_dir else []),
-        *(["--workspace-directory", os.fspath(config.workspace_dir)] if config.workspace_dir else []),
-        *(["--cache-directory", os.fspath(config.cache_dir)] if config.cache_dir else []),
-        "--cache-key", config.cache_key or '-',
-        *(["--package-cache-directory", os.fspath(config.package_cache_dir)] if config.package_cache_dir else []),  # noqa: E501
-        *(["--local-mirror", str(config.local_mirror)] if config.local_mirror else []),
-        "--incremental", str(config.incremental),
+        *([f"--compress-output={p}"] if (p := config.compress_output) else []),
```
behrmann, line 1313:
```suggestion
        *([f"--compress-output={c}"] if (c := config.compress_output) else []),
```

mkosi/__init__.py
```diff
@@ -4501,35 +4501,35 @@ def finalize_default_tools(config: Config, *, resources: Path) -> Config:
...
+        f"--incremental={config.incremental}",
         *([f"--package={package}" for package in config.tools_tree_packages]),
         *([f"--package-directory={os.fspath(directory)}" for directory in config.tools_tree_package_directories]),  # noqa: E501
         *([f"--build-sources={tree}" for tree in config.build_sources]),
-        "--build-sources-ephemeral", str(config.build_sources_ephemeral),
+        f"--build-sources-ephemeral={config.build_sources_ephemeral}",
         *([f"--prepare-script={os.fspath(script)}" for script in config.tools_tree_prepare_scripts]),
         "--output=tools",
-        *(["--source-date-epoch", str(config.source_date_epoch)] if config.source_date_epoch is not None else []),  # noqa: E501
+        *([f"--source-date-epoch={p}"] if (p := config.source_date_epoch) is not None else []),
```
behrmann, line 4526:
```suggestion
        *([f"--source-date-epoch={e}"] if (e := config.source_date_epoch) is not None else []),
```

behrmann (commented):
I'm sorry you're seeing so much minor CI fallout. With this change mypy seems to be more aggressively want to unify the type of like-named variables. The below renamings fix that.

DaanDeMeyer:
@keszybz Please rebase and then we can merge this

keszybz:
Rebased with the requested cosmetic changes.

behrmann (approved):

PR #3544, opened 2025-02-23
debian: use libeatmydata with dpkg

bluca:
Even with --force-unsafe-io dpkg still does a lot of fsyncs. Pull libeatmydata1 together with the bootstrapping set, and LD_PRELOAD it. It makes all fsync and related syscalls no-ops.

Before: 2m23.104s

After: 1m8.610s

The package is ~8KB and installed is ~30KB so it should make no difference to have it installed, and it's inert unless preloaded.

### DIFF
--- a/mkosi.conf.d/30-debian-kali-ubuntu/mkosi.conf
+++ b/mkosi.conf.d/30-debian-kali-ubuntu/mkosi.conf
@@ -8,6 +8,7 @@ Distribution=|ubuntu
 [Content]
 Packages=
         fish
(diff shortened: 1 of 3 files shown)


DaanDeMeyer (changes requested):
Just add an option to mkosi-sandbox to make fsync and friends a noop like we already do with chown()

DaanDeMeyer:
Call it --supress-sync and just copy what nspawn does with its --supress-sync option.

DaanDeMeyer:
Implemented the alternative solution in #3545

PR #3549, opened 2025-02-24
Drop ConfigureScripts=

DaanDeMeyer:
These were added to allow generating configuration based on the tools tree, specifically QemuArgs= and Drives= for systemd's integration tests. Now that mkosi sandbox exists and we run meson within mkosi sandbox in systemd (https://github.com/systemd/systemd/pull/36500), there's no real need to keep these around anymore, and we don't know of any users outside of systemd, so let's drop support for them.

These were always somewhat of a hack to begin with, since we didn't do any proper validation on the JSON returned by configure scripts.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -71,10 +71,8 @@ from mkosi.config import (
     cat_config,
     expand_delayed_specifiers,
     format_bytes,
-    have_history,
     parse_boolean,
     parse_config,
-    resolve_deps,
     summary,
     systemd_tool_version,
     want_selinux_relabel,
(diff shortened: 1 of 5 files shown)


slarew:
@DaanDeMeyer We use a configuration script (`ConfigureScripts=`) to customize the `KernelCommandLine` config value embedded in our secure boot signed UKI. Our use case for the config script is to set `root=UUID=...`, where the UUID is the filesystem UUID (and not partition UUID) derived from the current build's config seed.

Would you consider keeping `ConfigureScripts=` for now (possibly marking it as deprecated)?

DaanDeMeyer:
Yes of course, thanks for letting me know there's a user of this feature.

PR #3553, opened 2025-02-25
Enable Hyperscale rpm COW plugin

DaanDeMeyer:
The CentOS Hyperscale SIG provides rpm/dnf plugins to enable rpm
COW which drastically speeds up image builds so let's ship the
necessary logic to make use of it when building images with the
Hyperscale experimental repositories.


behrmann:
The first commits needs the run and subprocess import removed

PR #3557, opened 2025-02-26
qemu: Switch to -blockdev instead of -drive where possible

DaanDeMeyer:
-drive is considered legacy whereas -blockdev is the new interface. It's slightly more verbose but not incredibly so, let's switch to it.

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -1415,9 +1415,18 @@ def run_qemu(args: Args, config: Config) -> None:
 
         if want_scratch(config):
             scratch = stack.enter_context(generate_scratch_fs(config))
-            cache = "cache.writeback=on,cache.direct=on,cache.no-flush=yes,aio=io_uring"
+            blockdev = [
+                "driver=raw",
+                "node-name=scratch",
+                "discard=unmap",
+                "file.driver=file",
+                f"file.filename={scratch}",
+                "file.aio=io_uring",
+                "cache.direct=on",
+                "cache.no-flush=on",
+            ]
             cmdline += [
-                "-drive", f"if=none,id=scratch,file={scratch},format=raw,discard=on,{cache}",
+                "-blockdev", ",".join(blockdev),
                 "-device", "virtio-blk-pci,drive=scratch",
             ]  # fmt: skip
             kcl += [f"systemd.mount-extra=LABEL=scratch:/var/tmp:{config.distribution.filesystem()}"]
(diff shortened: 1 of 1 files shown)


mkosi/qemu.py
```diff
@@ -1500,14 +1502,16 @@ def add_virtiofs_mount(
...
             for drive in drives:
-                arg = f"if=none,id={drive.id},file={file},format=raw,file.locking=off,cache.writeback=on,cache.direct=on,cache.no-flush=yes,aio=io_uring"  # noqa: E501
+                file = f"file.driver=file,file.filename={tmp},file.aio=io_uring,file.locking=off"
+                cache = "cache.direct=on,cache.no-flush=yes"
+                arg = f"driver=raw,node-name={drive.id},{file},{cache}"
                 if drive.options:
                     arg += f",{drive.options}"
 
-                cmdline += ["-drive", arg]
+                cmdline += ["-blockdev", arg]
```
behrmann, line 1514: That reads a bit backwrds, how about
```python
                arg = [
                    "driver=raw",
                    f"node-name={drive.id}",
                    "file.driver=file",
                    f"file.filename={tmp}",
                    "file.aio=io_uring",
                    "file.locking=off"
                    "cache.direct=on",
                    "cache.no-flush=yes",
                ]
                if drive.options:
                    arg += [drive.options]

                cmdline += ["-blockdev", ",".join(arg)]
```

behrmann (approved):

mkosi/qemu.py
kashyapc: That's quick.  To follow-up from the Matrix chat.  Although `-drive` works, its usage is considered "legacy" for about ~5 years now by the QEMY block layer folks.

Quoting the QEMU "block device options" section:

> "The most explicit way to describe disks is to use a combination of -device to specify the hardware device and -blockdev to describe the backend. The device defines what the guest sees and the backend describes how QEMU handles the data. It is the only guaranteed stable interface for describing block devices and as such is recommended for management tools and scripting.
>
> "The -drive option combines the device and backend into a single command line option which is a more human friendly. There is however no interface stability guarantee although some older board models still need updating to work with the modern blockdev forms."

I wouldn't read too much into "no interface stability guarantee" of `-drive` — as it has worked well for many years.  That said, `-blockdev` is the way forward, as `-drive` can get deprecated down the line.

I use two invocations of `-blockdev` is something like this (depending on what I'm doing).  A slightly verbose variant:

```
-blockdev node-name=node-Base,driver=qcow2,file.driver=file,file.filename=./base.qcow2,file.node-name=node-Base-file \
-device virtio-blk,drive=node-Base,id=virtio0
```

A shorter variant:
```
-blockdev driver=qcow2,node-name=n1,file.filename=cirros.qcow2,file.driver=file
-device virtio-blk,drive=n1,id=virtio0
```

I guess `mkosi` can get away with the shorter variant.  But please make sure all your existing use-cases continue to work before committing this :-)

bluca:
I think this breaks the systemd build in noble, but can't say for sure as at the moment there's a newer failure (https://github.com/systemd/mkosi/pull/3566 ) once that one is solved I'll check again

bluca:
```
5621s qemu-system-x86_64: -blockdev driver=raw,node-name=nvme0,file.driver=file,file.filename=/var/tmp/mkosi-drive-nvme0933av9lr,file.aio=io_uring,file.locking=off,cache.direct=on,cache.no-flush=yes,cache=unsafe: Parameters 'cache.*' used inconsistently
```

https://autopkgtest.ubuntu.com/results/autopkgtest-noble-upstream-systemd-ci-systemd-ci/noble/amd64/s/systemd-upstream/20250301_025258_75f65@/log.gz

looks related to me?

behrmann:
Looking at https://manpages.ubuntu.com/manpages/noble/en/man1/qemu-system-x86_64.1.html I think `cache=unsafe` is a `-drive`-only option and should be dropped, since we already set the other to `cache.` options.

behrmann:
These are not in mkosi, though, but in the systemd tests.

bluca:
yeah but this started with this change I think

behrmann:
Indeed, but if you grep for unsafe in the above man page, you'll find a table from which I gather, that it should be possible to just drop this option .

PR #3558, opened 2025-02-27
Add InitrdProfiles= with initial storage profile

DaanDeMeyer:
Let's allow not installing lvm2 in the initrd by introducing a storage profile which users can disable.


mkosi/config.py
```diff
@@ -600,6 +600,10 @@ class ToolsTreeProfile(StrEnum):
     runtime = enum.auto()
 
 
+class InitrdProfile(StrEnum):
+    storage = enum.auto()
```
behrmann, line 604: needs
```
list[InitrdProfile]: enum_list_transformer,
```
in the transformers dictionary

behrmann (approved):

aafeijoo-suse:
I do really like this, thanks!

PR #3564, opened 2025-02-28
Rename "storage" initrd profile to "lvm"

aafeijoo-suse:
Since it actually only provides support for LVM, it's best to keep the different specific functionalities separate.


DaanDeMeyer (approved):

behrmann:
This needs to actually change the name of the profile in `mkosi/resources/mkosi-initrd/mkosi.profiles/storage`

behrmann (approved):

PR #3568, opened 2025-03-01
mkosi-obs: Replace usage of pesign/efitools with systemd-sbsign

DaanDeMeyer:
Let's consolidate on systemd-sbsign across mkosi to do everything related to secure boot signing.

### DIFF
--- a/mkosi/resources/mkosi-obs/mkosi.build
+++ b/mkosi/resources/mkosi-obs/mkosi.build
@@ -4,7 +4,7 @@
 # Second stage of build:
 # - signed hashes are in cpio archive in SOURCES/ together with artifacts from previous build
 # - if there are PCR policies to attach, do so and then prepare the hashes of the UKIs themselves
-# - if not, attach the signatures to the UKI(s) with pesign
+# - if not, attach the signatures to the UKI(s) with systemd-sbsign
 # - remove shasums of previous artifacts, given we are re-creating them after this step
 # - place artifacts from previous builds and signed UKI in output directory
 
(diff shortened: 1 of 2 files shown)


DaanDeMeyer:
Requires https://github.com/systemd/systemd/pull/36574

bluca:
It's fine to use these new tools elsewhere, but I do not want them in the OBS scripts. Everything else in OBS uses pesign, including package builds, so I want to use the same thing, to guarantee it works. pesign has also been battle-tested on real hardware for 10 years, while this new stuff hasn't, I imagine it will take a few years of actual usage before it becomes compatible with all the firmwares and hardware out there

behrmann:
I mean, shouldn't we have both at least as an option? Assuming that it doesn't work in all cases, then if we don't dogfood it, it won't.

bluca:
Sure it's fine to use these other tools elsewhere, in the ci, for local builds, etc etc

DaanDeMeyer:
> It's fine to use these new tools elsewhere, but I do not want them in the OBS scripts. Everything else in OBS uses pesign, including package builds, so I want to use the same thing, to guarantee it works. pesign has also been battle-tested on real hardware for 10 years, while this new stuff hasn't, I imagine it will take a few years of actual usage before it becomes compatible with all the firmwares and hardware out there

This is a bit ridiculous, as long as we generate identical output to the other tools, of course we'll be compatible. I'll add integration tests in systemd that ensure we generate the same output.

bluca:
You cannot verify that you have the same output, because you don't have access to the same input - the certificate. And yes that matters. Hardware quirks matter too, as you found out the hard way with the padding bug in the stub. What the spec says and what happens in real life often diverge.

DaanDeMeyer:
Cool I'll just modify this PR to have it call both systemd-sbsign and the existing tool and verify the output is identical for both tools and fail the build if it isn't. Then you can be happy knowing we're using your supposedly stable better legacy tools and I can be happy knowing systemd-sbsign is dogfooded and if the comparison ever fails we'll automatically get a bug report so we can fix whatever is different in systemd-sbsign.

PR #3572, opened 2025-03-02
Improve parsing of local mkosi config

septatrix:
Clarify that mkosi.local shall be a directory with a mkosi.conf file inside of it.

### DIFF
--- a/mkosi/resources/man/mkosi.1.md
+++ b/mkosi/resources/man/mkosi.1.md
@@ -326,9 +326,9 @@ grouped by section below.
 Configuration is parsed in the following order:
 
 * The command line arguments are parsed.
-* `mkosi.local.conf` or `mkosi.local` is parsed if it exists. This file or
-  directory should be in `.gitignore` (or equivalent) and is intended for local
-  configuration.
+* `mkosi.local.conf` and `mkosi.local/` are parsed if they exists (in that order).
+  This file or directory should be in `.gitignore` (or equivalent)
+  and is intended for local configuration.
 * Any default paths (depending on the option) are configured if the
   corresponding path exists.
 * `mkosi.conf` is parsed if it exists in the directory configured with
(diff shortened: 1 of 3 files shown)


septatrix:
I think the cleaner fix is to also unconditionally parse `mkosi.local/` iff it is a dir to allow stuff like `mkosi.local/mkosi.conf.d/...` without requiring a `mkosi.conf` file for detection

septatrix:
To expand on the "and allows working around some limitations of the config system", the problem is when one wants to locally include something which uses matches or profiles that are set by `mkosi.conf`.

```ini
# File: mkosi.conf
[Distribution]
Distribution=arch

# File: mkosi.local/mkosi.conf.d/20-include-tools.conf
[Include]
Include=mkosi-tools
```

This can now be worked around by putting the distro configuration at `mkosi.local/mkosi.conf.d/10-default-distro.conf` and ensuring that this file **is** tracked by git. Not a very pretty solution but it enables a few things that are otherwise not possible. Every other solution (e.g. postpone parsing of configs with `[Matches]`) I could think of had other disadvantages or just moves that problem to other scenarios

mkosi/resources/man/mkosi.1.md
```diff
@@ -326,9 +326,9 @@ grouped by section below.
 Configuration is parsed in the following order:
 
 * The command line arguments are parsed.
-* `mkosi.local.conf` or `mkosi.local` is parsed if it exists. This file or
-  directory should be in `.gitignore` (or equivalent) and is intended for local
-  configuration.
+* `mkosi.local.conf` and `mkosi.local/` are parsed if they exists (in that order).
+  This file or directory should be in `.gitignore` (or equivalent)
+  and is intended for local configuration.
```
behrmann, line 331: Sorry, late to the game, the "this file or directory" makes sense logically, but should probably be an "and", since above we now say that both are parsed. In the last line it should it should also read "are" instead of "is", since both are intended for local configuration.

septatrix: -> 3581

PR #3573, opened 2025-03-02
Fix signing of grub boot binary

MihirLuthra:
When `sign_efi_binary` runs sandbox for signing, it will mount input path as read only. But in case of grub, the caller function passes input and output path to be the same. Due to this, command complains `image_write/open: Read-only file system` and exits.


DaanDeMeyer:
I don't think this is the proper fix, please test https://github.com/systemd/mkosi/pull/3574 instead

PR #3578, opened 2025-03-03
Add "raid" initrd profile

aafeijoo-suse:
This initrd profile provides support for assembling RAID arrays.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -602,6 +602,7 @@ class ToolsTreeProfile(StrEnum):
 
 class InitrdProfile(StrEnum):
     lvm = enum.auto()
+    raid = enum.auto()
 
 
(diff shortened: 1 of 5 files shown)


DaanDeMeyer:
So the local in "mkosi-initrd" until now was always limited to `initrd.py`, this PR would change that which I don't think we want. I'd prefer if we keep the /etc/mdadm.conf related changes localized to `initrd.py`.

aafeijoo-suse:
> So the local in "mkosi-initrd" until now was always limited to `initrd.py`, this PR would change that which I don't think we want. I'd prefer if we keep the /etc/mdadm.conf related changes localized to `initrd.py`.

Ok, it makes sense, changed.

DaanDeMeyer (approved):

PR #3582, opened 2025-03-05
Move package installation and removal to PackageManager interface

DaanDeMeyer:
There's no need for these to be implemented by the Distribution
interface as they don't need distribution specific knowledge so let's
move them to the PackageManager interface instead.

### DIFF
--- a/mkosi/distributions/debian.py
+++ b/mkosi/distributions/debian.py
@@ -143,6 +143,13 @@ class Installer(DistributionInstaller):
                     f"-oDPkg::Pre-Install-Pkgs::=cat >{workdir(Path(f.name))}",
                     "?essential",
                     "base-files",
+                    # Debian policy is to start daemons by default. The policy-rc.d script can be used choose
+                    # which ones to start. Let's install the necessary packages to deny all daemon startups.
+                    # Instead, systemd presets should be used to decide which daemons are enabled and which
+                    # are not. See https://people.debian.org/~hmh/invokerc.d-policyrc.d-specification.txt for
+                    # more information.
+                    "policy-rcd-declarative",
+                    "policy-rcd-declarative-deny-all",
                 ],
                 options=["--bind", f.name, workdir(Path(f.name))],
             )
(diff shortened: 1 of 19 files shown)


mkosi/distributions/debian.py
```diff
@@ -143,6 +144,13 @@ def install(cls, context: Context) -> None:
                     f"-oDPkg::Pre-Install-Pkgs::=cat >{workdir(Path(f.name))}",
                     "?essential",
                     "base-files",
+                    # Debian policy is to start daemons by default. The policy-rc.d script can be used choose
+                    # which ones to start. Let's install the necessary packages to deny all daemon startups.
+                    # Instead, systemd presets should be used to decide which daemons are enabled and which
+                    # are not. See https://people.debian.org/~hmh/invokerc.d-policyrc.d-specification.txt for
+                    # more information.
+                    "policy-rcd-declarative",
+                    "policy-rcd-declarative-deny-all",
```
behrmann, line 153: I'm not sure I see the value in this, since I don't think this uses presets at all. If I remember correctly, they are used by init-system-helpers during install, but policy-rc.d is also asked, so this would be an additional deny-all on top of presets? Also, this doesn't really do anything more than our one-line shell snippet does, but would leave that around in the system, which doesn't really seem ideal. 

I do agree that Debian should use presets, but [policy-rcd-declarative](https://salsa.debian.org/wouter/policy-decl/-/tree/master?ref_type=heads) doesn't really seem to lead anywhere.

behrmann (commented):
LGTM for the most part, one minor nit and a comment on `policy-rcd-declarative`, which I don't think is worth it.

mkosi/distributions/debian.py
DaanDeMeyer: @behrmann I think it's actually ideal to leave it around in the system. It effectively disables daemon enablement via the old legacy invoke-rc.d stuff.

bluca: This pulls in the entirety of perl, and it's pretty much abadonware - last updated 5 years ago

bluca: You can do the same by dropping in a two lines script btw:

```
mkdir -p $BUILD_ROOT/usr/sbin
cat > $BUILD_ROOT/usr/sbin/policy-rc.d <<EOF
#!/bin/sh
exit 101
EOF
```

bluca: The full one? in a minimal image? No absolutely not - only the "base" one, which is a few libraries

DaanDeMeyer: https://github.com/systemd/mkosi/pull/3584

PR #3583, opened 2025-03-05
Use target path as mount tags

septatrix:
This prevents collisions when mounting something to directories of the same name or root.

I noticed this when I wanted to mount something to `/root` (but not my homedir so `RuntimeHome=yes` was not an option). This conflicted with the root mount itself. Virtiofsd seems to be completely fine with slashes in tags so just using the target path is an easy fix.


septatrix:
Ah yes, 36 chars:

https://github.com/torvalds/linux/blob/848e076317446f9c663771ddec142d7c2eb4cb43/include/uapi/linux/virtio_fs.h#L12-L13

https://gitlab.com/virtio-fs/virtiofsd/-/blob/c0f7439d5382a4cf61439d12aa6eec3fb3ab6f1e/src/vhost_user.rs#L54

PR #3584, opened 2025-03-06
Revert 9f4d174daac954aaa9e18ee9bb58c6d7a946ec9d

DaanDeMeyer:
The declarative package pulls in perl which is not ideal, let's
switch back to the previous approach.

### DIFF
--- a/mkosi/distributions/debian.py
+++ b/mkosi/distributions/debian.py
@@ -144,13 +144,6 @@ class Installer(DistributionInstaller):
                     f"-oDPkg::Pre-Install-Pkgs::=cat >{workdir(Path(f.name))}",
                     "?essential",
                     "base-files",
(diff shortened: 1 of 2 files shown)


bluca:
There was actually the script added before, why not restore that too?

DaanDeMeyer:
Because I'd like to figure out if we still need it or not.

DaanDeMeyer:
nvm I'm not too interested in chasing this right now, let's switch back to the previous approach

behrmann:
> Because I'd like to figure out if we still need it or not. Even https://wiki.debian.org/Daemon does not mention invoke-rc.d anymore.

Speaking from experience, our `policy-rc.d` runs on every apt invocation that touches a running service. I don't think that interface will change.

PR #3585, opened 2025-03-06
Rework --run-build-scripts

DaanDeMeyer:
Instead of (optionally) building the image when this option is enabled,
let's insist that the image has already been built and cached. This allows
us to reuse the history of the previous build if History= is enabled. At
the same time, rename the option to --rerun-build-scripts to indicate more
clearly that the image needs to have been built once already.

This allows the option to be used in systemd's mkosi.clangd script to replace
"-t none" and "--incremental=strict" there.


mkosi/resources/man/mkosi.1.md
```diff
@@ -293,6 +293,12 @@ Those settings cannot be configured in the configuration files.
 `--wipe-build-dir`, `-w`
 :   Wipe the build directory if one is configured before building the image.
 
+`--rerun-build-scripts`, `-R`
+:   Rerun build scripts. Requires the `Incremental=` option to be
+    enabled and the image to have been built once already. If `History=`
+    is enabled, the config from the previous build will be reused and no
+    new history will be written.
```
behrmann, line 300: This should probably also clearly spell out that configure scripts will not be rerun.

mkosi/__init__.py
```diff
@@ -5114,50 +5118,64 @@ def run_verb(args: Args, images: Sequence[Config], *, resources: Path) -> None:
...
+    if last.is_incremental() and (last.incremental == Incremental.strict or args.rerun_build_scripts):
+        if args.force > 1:
+            die(
+                "Cannot remove incremental caches when building with Incremental=strict",
+                hint="Build once with '-i yes' to update the image cache",
+            )
+
+        for config in images:
+            if not config.is_incremental() or have_cache(config):
+                continue
```
behrmann, line 5163: That both branches below die tripped me for a second before I looked at the other side of the diff.

I think this could be written as
```suggestion
        if any((c := config).is_incremental() and not have_cache(c) for config in images:
```
to reuse the breaking config in the error message. A construction like this was recently used in ukify. That or a small comment above the loop would probably make this easier on the reader.

behrmann (approved):

mkosi/resources/man/mkosi.1.md
DaanDeMeyer: That should probably be part of the `History=` docs

behrmann (approved):

PR #3586, opened 2025-03-08
Add option to persist runtime drives

### LINKED ISSUE #3576: Offer persist flag for `Drives=`
Currently the `Drives=` option is very similar to a scratch disk as they are removed after each run. It would be useful to add a new flag `persist`, which, when set, suppresses removal of the respective file. This flag should require that `file-id` is set for it to function.

This is useful when having an otherwise ephemeral setup but wanting to persist some other stuff across starts/stops that cannot be covered by runtime trees/virtiofsd.

septatrix:
Fixes #3576

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -156,6 +156,7 @@ class Drive:
     directory: Optional[Path]
     options: Optional[str]
     file_id: str
+    persist: bool
 
 
 # We use negative numbers for specifying special constants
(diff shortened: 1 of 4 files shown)


behrmann (approved):

DaanDeMeyer:
Can we make this a little more extensible? Please read a comma delimited list of options from the new field, with initiall just persist supported. I don't like the current implementation at all.

septatrix:
> Can we make this a little more extensible? Please read a comma delimited list of options from the new field, with initiall just persist supported. I don't like the current implementation at all.

You mean like `Drives=...:persist,foo,bar`?

DaanDeMeyer:
> > Can we make this a little more extensible? Please read a comma delimited list of options from the new field, with initiall just persist supported. I don't like the current implementation at all.
> 
> You mean like `Drives=...:persist,foo,bar`?

Yes

PR #3589, opened 2025-03-13
mkosi-initrd: protect old initrd image against errors

aafeijoo-suse:
When `copy_tree()` ends up calling `cp`, if it fails because there is not enough space, it leaves and incomplete initrd image in the output directory. If we are writing the initrd image directly where a bootloader entry has an initrd configured, the system will fail to boot. So, backup the old initrd image before calling `copy_tree()`, and restore it if that function throws an exception.


DaanDeMeyer (changes requested):
This is very messy., there's no guarantee we can put the backup back in place when copying the new one fails. We should copy the new initrd to a temporary file next to the initrd we're replacing and then use rename() to atomically replace the old one.

mkosi/initrd.py
```diff
@@ -141,12 +142,32 @@ def initrd_finalize(staging_dir: str, output: str, output_dir: str) -> None:
     else:
         output_dir = os.fspath(Path.cwd())
 
+    # backup current image
+    if Path(f"{output_dir}/{output}").exists():
```
behrmann, line 146: The `Path` calls everywhere are quite ugly and the same string is made a `Path` many times over, maybe just make all of them `Path`s in the beginning.

behrmann (commented):
I agree with @DaanDeMeyer on how this should be implemented, but generally I do want to see this feature. I've had the exact thing this is supposed to prevent happen to me with dracut a few times. :)

aafeijoo-suse:
> We should copy the new initrd to a temporary file next to the initrd we're replacing and then use rename() to atomically replace the old one.

With this approach we are always invalidating an amount of space equal to the initrd size, but ok, I see your point.

mkosi/initrd.py
aafeijoo-suse: You're quite right...

DaanDeMeyer:
Let's merge https://github.com/systemd/mkosi/pull/3591 first and then rebase this on top of that and make use of Path more

mkosi/initrd.py
```diff
@@ -141,12 +142,22 @@ def initrd_finalize(staging_dir: str, output: str, output_dir: str) -> None:
...
     # mkosi symlinks the expected output image, so dereference it
-    copy_tree(
-        Path(f"{staging_dir}/{output}").resolve(),
-        Path(f"{output_dir}/{output}"),
-    )
+    try:
+        copy_tree(staging.resolve(), tmp)
+    except subprocess.CalledProcessError:
+        rmtree(tmp)
+        raise
```
behrmann, line 155: Since the except is here only to ensure cleanup, we can move it into a `finally` instead, which is a bit more concise and saves the `import subprocess`.
```suggestion
finally:
        rmtree(tmp)
```

behrmann (approved):
LGTM, #3591 has just been merged, so this just needs the rebase. One small nit

mkosi/initrd.py
behrmann: Scratch that nonsense, I'll get some coffee…

behrmann (approved):

mkosi/initrd.py
```diff
@@ -187,12 +188,22 @@ def initrd_finalize(staging_dir: Path, output: str, output_dir: Optional[Path])
...
+
+    log_notice(f"Moving {tmp} to {final}")
+    if final.is_dir():
+        rmtree(final)
```
DaanDeMeyer, line 205: I'd just do this unconditionally

PR #3592, opened 2025-03-14
Make `Drive=` flags more general

septatrix:
> Can we make this a little more extensible? Please read a comma delimited list of options from the new field, with initiall just persist supported. I don't like the current implementation at all.

_Originally posted by @DaanDeMeyer in https://github.com/systemd/mkosi/issues/3586#issuecomment-2716986154_


behrmann (approved):

septatrix:
A bit of bikeshedding but I think "persistent" would be preferable grammatically?

septatrix:
Note to self: man page is missing

septatrix:
I also changed it to always use a stable filename for drives. This makes it more consistent with most other places where we create files, sockets, or systemd unit names. However, this would now come at the disadvantage that two users on the same machine would fight over the same filename under `/var/tmp`. I thought about also adding `$USER` to the filename...any ideas?

Motivation was that I do not always need `persist` but instead only want to run some scripts from outside on the backing file. Adding an extra flag like `suppress-tmp-suffix` seems convoluted if we can already reasonably do this by default

mkosi/config.py
```diff
@@ -1382,13 +1386,15 @@ def parse_drive(value: str) -> Drive:
...
+    flag_list = p.split(",") if len(parts) > 5 and (p := parts[5]) else []
+
     return Drive(
         id=id,
         size=parse_bytes(parts[1]),
         directory=parse_path(p) if len(parts) > 2 and (p := parts[2]) else None,
         options=p if len(parts) > 3 and (p := parts[3]) else None,
         file_id=p if len(parts) > 4 and (p := parts[4]) else id,
-        persist=parse_boolean(p) if len(parts) > 5 and (p := parts[5]) else False,
+        flags=[DriveFlag(f) for f in flag_list],
```
DaanDeMeyer, line 1397: This will blow up spectacularly on unknown flags. Please use `make_enum_parser(DriveFlag)(f)` or so

DaanDeMeyer (changes requested):

mkosi/resources/man/mkosi.1.md
```diff
@@ -1822,7 +1822,7 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 
 `Drives=`, `--drive=`
 :   Add a drive. Takes a colon-delimited string of format
-    `<id>:<size>[:<directory>[:<options>[:<file-id>[:<persist>]]]]`. `id` specifies
```
DaanDeMeyer, line 1825: Should be in the first commit

mkosi/resources/man/mkosi.1.md
```diff
@@ -1834,11 +1834,12 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     backing the drive. If unset, this defaults to the drive ID.
     Drives with the same file ID will share the backing file.
     The directory and size of the file will be determined from the first drive with a given file ID.
-    `persist` takes a boolean value and determines whether the drive will be persisted across **qemu** invocations.
-    Enabling persistence also prevents suffixing the filename with a random string.
-    The file backing the drive will always be available under `/<directory>/mkosi-drive-<file-id>`
-    You can skip values by setting them to the empty string, specifying e.g. `myfs:1G::::yes`
-    will create a persistent drive under `/var/tmp/mkosi-drive-myfs`.
```
DaanDeMeyer, line 1841: This should be in the first commit as well

DaanDeMeyer:
Actually can you drop the last commit or make it a separate PR? I'm not a big fan. You can always just glob to get the full filename.

DaanDeMeyer (changes requested):

septatrix:
> Actually can you drop the last commit or make it a separate PR? I'm not a big fan. You can always just glob to get the full filename.

I updated it to only include the `machine_or_name` now but I agree, globbing should be sufficient so I would also be fine with completely dropping it. Adding at least machine/name though makes it at least consistent with the rest of the codebase.

---

I also noticed that `NamedTemporaryFile` keeps the file open so I adapted the persist logic to do the same which simplifies the code a bit. While at it: does `maybe_make_nocow` work if there are still active FDs to the file?

DaanDeMeyer (approved):

PR #3595, opened 2025-03-17
Add yubico-piv-tool in tools

p1gp1g:
OpenSC works fine when using the main signing slot of yubikeys, but doesn't work with "retired slots". Adding yubico-piv-tool allows to do it with other slots as well.

Retired slots for yubikeys can be used to store keys used for different purposes. For example, someone may use the main signing slot to sign packages or libraries, and want to use a different key for secure boot. This other key can be stored in Retired Slot 1.

I think Yubikeys are common hardware that can be used to sign images with mkosi, so it makes sense to have it in the _tools_ image.

### DIFF
--- a/mkosi/resources/mkosi-tools/mkosi.conf
+++ b/mkosi/resources/mkosi-tools/mkosi.conf
@@ -26,4 +26,5 @@ Packages=
         systemd
         tar
         xfsprogs
+        yubico-piv-tool
         zstd
--- a/mkosi/resources/mkosi-tools/mkosi.extra/usr/share/p11-kit/modules/libyksc11.module
+++ b/mkosi/resources/mkosi-tools/mkosi.extra/usr/share/p11-kit/modules/libyksc11.module
@@ -1,0 +1,3 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+module: /usr/lib64/libykcs11.so.2


DaanDeMeyer:
I don't think we should work around limitations in opensc in mkosi. Please work with opensc to make it work with retired slots.

p1gp1g:
There is an open issue for that: https://github.com/OpenSC/OpenSC/issues/3287 and I agree it would be better to have a proper fix in opensc, but until then it would be nice to just be able to sign the images. Is there anything blocking except it isn't the perfect solution ? Can it be included until the issue is fixed ?

p1gp1g:
Another solution would a way to set some extra files for the tools tree

DaanDeMeyer:
@p1gp1g There's ToolsTreePrepareScripts= now, you can use that to make whatever changes you want locally

p1gp1g:
For anyone seing this PR: it wasn't possible to use opensc because of the missing context and yubico manual to restore slots history wasn't correct.

It is possible to restore it with this:
```console
$ echo C10114C20100FE00 | xxd -r -p | ykman --device myserial piv objects import 0x5FC10C - --management-key  '[...]' --pin '[...]'
```

cf. https://github.com/OpenSC/OpenSC/issues/3287#issuecomment-3123551187

PR #3598, opened 2025-03-18
Fix KernelModulesExclude= settings

DaanDeMeyer:
With https://github.com/systemd/mkosi/commit/eecf8b3b5c43e4832a11c8718ae43e559a755829, exclude settings have
started taking priority over include or the new glob settings whereas
they should not. Fix the logic so the globs and include patterns always
take priority over the exclude patterns.


behrmann (approved):
Mea culpa, overlooked that in eecf8b3b5c43e4832a11c8718ae43e559a755829

PR #3599, opened 2025-03-19
initrd: Install gzip in all initrds, not just the Arch one

DaanDeMeyer:
Fedora also has keymaps compressed with gzip and doesn't pull in gzip by default so let's just install gzip everywhere to make the problem go away.


keszybz (approved):
Yeah, this is a pragmatic solution.

Issue #3338, opened 2025-01-09
Services with `ConditionNeedsUpdate=` are still executed at first boot


septatrix:
### mkosi commit the issue has been seen with

main

### Used host distribution

Fedora 41

### Used target distribution

Arch

### Linux kernel version used

6.12.7-200.fc41.x86_64

### CPU architectures issue was seen on

x86_64

### Unexpected behaviour you saw

On the first boot services which have `ConditionNeedsUpdate=/var` (or `/etc`) set are still started even though they are fresh images. This delays the bootup unnecessarily. For images with an immutable `/etc` this might even result in them being started each boot (as the `/etc/.updated` file cannot be created) though I did not explicitly test that.

One example of such a service is `systemd-journal-catalog-update` even though the catalog is freshly installed. `systemd-hwdb-update` would be a similar candidate, however, it has additional checks which do not trigger because mkosi manually invokes `systemd-hwdb --usr`. For other services these optimizations do not exist.

`mkosi` should likely start all services which use `ConditionNeedsUpdate=` as part of the build process and call `/usr/lib/systemd/systemd-update-done` at the very end. Doing it this way should be safe as either: The `/etc` and/or `/var` directories are included in the image in which case their generated files would also be included, or, they are not included in which case the `.updated` files would also be absent resulting in them being invoked again as expected.

---

(I noticed there is also another issue if a device reboots from v1 to v2 after v3 is built but not yet installed. In that case the `.updated` files will have timestamps later than v3's `/usr` resulting in `ConditionNeedsUpdate=` not being triggered after a subsequent update from v2 to v3. This, however, is a technical problem in the way these `.updated` files currently work and not a problem of mkosi.)

### Used mkosi config

```ini
Default config from this repo
```

DaanDeMeyer:
I'll certainly review a PR to run `systemd-journal-catalog-update` as part of the image build process. I'm not sure we should be calling systemd-update-done though, that seems like something to fix in systemd.

septatrix:
> Even if that were implemented there are still several other services which will get invoked on first boot unnecessarily.

One example is ldconfig to populate `/etc/ld.so.cache`. Here it also will not be possible to support the mechanism from hwdb/udev of checking if `/usr/lib/ld.so.cache` exists because it will be very hard to change to logic of `ld.so`.

[...]

PR #3602, opened 2025-03-19
Make /etc/ and /usr/ as updated after installation

keszybz:
This implements the "conservative approach" discussed in https://github.com/systemd/mkosi/issues/3338. If we know that /etc/ and /var/ were populated during the installation, we can opt out of running early boot services, making the first boot quicker.

C.f. https://bugzilla.redhat.com/show_bug.cgi?id=2348669.

### DIFF
--- a/mkosi.finalize
+++ b/mkosi.finalize
@@ -1,0 +1,5 @@
+#!/bin/bash
+# SPDX-License-Identifier: CC-0
+set -e
+
+touch -r "$BUILDROOT/usr" "$BUILDROOT/etc/.updated" "$BUILDROOT/var/.updated"


DaanDeMeyer:
Why do this in a finalize script? That will only run in the mkosi repo itself. Shouldn't we do this in python in build_image() for all images?

keszybz:
> Why do this in a finalize script? That will only run in the mkosi repo itself. Shouldn't we do this in python in build_image() for all images?

See the discussion in #3338. We don't *know* for certain if the image relies of those services being started or not. I think that in fact in most cases this is true, and starting them just wastes time, but in principle somebody could just drop in a `sysusers.d` file in `mkosi.extra` and rely on the user being created.

I think that there *are* good reasons to change the semantics and say that all such updates must be done during the build. This is in particular necessary if the image is to be used read-only. At least because of that I think it'd make sense to do this.

septatrix:
> Why do this in a finalize script? That will only run in the mkosi repo itself. Shouldn't we do this in python in build_image() for all images?

As discussed in the linked issue this would not always be sound. Think e.g. of a user which installs a library during `mkosi.build`. In this case the `ldconfig.service` should still be started. This also applies to other things installed manually like journal catalogs.

septatrix (changes requested):
Commit message: Make → Mark

keszybz:
This got merged into systemd as https://github.com/systemd/systemd/commit/858e59c82c.

PR #3604, opened 2025-03-20
kmod: Include all kernel modules if no include patterns were specified

### LINKED ISSUE #3603: Disk image does not boot (no output on hvc0 and hang on /dev/gpt-auto-root)

f11a4ba751cf287cba4f00b54779d4f7e25a5516

### Used host distribution

Arch

### Used target distribution

Arch

### Linux kernel version used

6.13.7-arch1-1

x86_64

I tried to create a bootable image and start a VM from it on Arch:
```
mkosi -p linux,systemd,bash --autologin=true --bootable=true
mkosi vm
```

I see no output in the terminal (hvc0).
When starting with `mkosi --console=gui vm` I also see that it is actually booting but it then hangs on:
```
A start job is running for /dev/gpt-auto-root
```

The same does work if I build the image with the older mkosi 25.3 from the Arch packages.

### Used mkosi config

```ini

```

### mkosi output

```sh

```

DaanDeMeyer:
Fixes #3603

### DIFF
--- a/mkosi/kmod.py
+++ b/mkosi/kmod.py
@@ -124,7 +124,7 @@ def filter_kernel_modules(
                 remove.add(m)
 
         modules -= remove
-    else:
+    elif include:
         # If no exclude patterns are specified, only keep the specified kernel modules.
         modules = {modulesd / m for m in keep}
 


dvob:
Thank you for the quick fix.
I tested it and it fixes the problems of #3603

PR #3605, opened 2025-03-20
Implement linux-noinitrd firmware

DaanDeMeyer:
Useful for direct kernel booting without an initrd even if one was built.


behrmann (commented):
Needs a `ruff check --fix` to sort imports and run the rest of CI, apart from that LGTM.

PR #3606, opened 2025-03-20
Scmversion

behrmann:
This commit changes the versioning scheme of non-release versions of mkosi to
align with the standard Python version scheme [1]. Unfortunately this scheme is
not fully compatible with the UAPI group version format specification.

Concretely, this changes

    26~devel

to

    25.3.post1.dev244+g0eac3718

The used format ensures that for prereleases the following inequalities still
hold

    25.3 < 25.3.post1.dev244+g0eac3718 < 25.4 < 26

Usage of this new format has the following benefits:

1. It is compatible with PyPI and allows publishing mkosi there in the future

2. Since it contains the number of commits since the last tag as well as part
   of the commit hash in the "local part" of the version, it becomes easier for
   bug reporters to show the version they are on no matter how mkosi has been
   installed.

3. The version no longer needs to be managed manually.

The current implementation moves the version from mkosi/sandbox.py to a new
file _version.py that obtains the version either via importlib metadata (in
case mkosi has been installed) or by calling setuptools-scm directly, to
support the bin/ shims. Since the version is no longer available in
mkosi/sandbox.py, it passed as an environment variable for scripts.

[1] https://packaging.python.org/en/latest/specifications/version-specifiers/


mkosi/_version.py
```diff
@@ -0,0 +1,32 @@
...
+import subprocess
+from importlib.metadata import PackageNotFoundError, version
+from pathlib import Path
+
+try:
+    __version__ = version("package-name")
+except PackageNotFoundError:
+    # This branch means, that mkosi was not installed as a package as Python is aware, this is most likely
+    # the case if mkosi was run via the bin/mkosi shim in mkosi's repository. Though discouraged, we fall
+    # back to importing setuptools_scm directly and having it try figure stuff out.
```
septatrix, line 14: You are not importing anything here?

mkosi/_version.py
```diff
@@ -0,0 +1,32 @@
...
+        )
+        version, *rest = p.stdout.strip().split("-")
+        version = version.lstrip("v")
+        if rest:
+            numcommits, commit = rest
+            __version__ = f"{version}.post1.dev{numcommits}+{commit}.d{datetime.datetime.now():%Y%m%d}"
+        else:
+            # we are exactly on a trag
+            __version__ = version
+    except (subprocess.CalledProcessError, NotADirectoryError):
```
septatrix, line 31: Can also fail with `FileNotFoundError` if git is not installed (e.g. someone just downloaded the archive via the browser)

mkosi/_version.py
```diff
@@ -0,0 +1,32 @@
...
+        __version__ = os.getenv("MKOSI_VERSION", "0")
```
septatrix, line 32: Maybe we should log a warning if we fall back to 0 here?

septatrix (commented):
Maybe the version fallback should happen in the `bin/` scripts directly (i.e. they would set the env var) if that is the only way we encourage people to use it without installing it?

septatrix:
Also, what happens when people have installed (an older version) of mkosi via pip (or their distro) but call a local mkosi clone via the `bin/` wrappers?

mkosi/_version.py
behrmann: That comment was leftover from the first version that called setuptools-scm directly, but I wanted to have the fallback work without having having to have that installed.

mkosi/_version.py
```diff
@@ -0,0 +1,32 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+import datetime
+import os
+import subprocess
+from importlib.metadata import PackageNotFoundError, version
+from pathlib import Path
+
+try:
+    __version__ = version("package-name")
```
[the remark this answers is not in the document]
behrmann: Made it explictly `"mkosi"`. Left over from testing the fallback path.

behrmann:
> Maybe the version fallback should happen in the bin/ scripts directly (i.e. they would set the env var) if that is the only way we encourage people to use it without installing it?

I think it's nicer to have it in the module, since it gives the power of both a proper language as well as making it unnecessery to implement this multiple times.

mkosi/_version.py
behrmann: I intended the 0 to be a warning of some sort, since it's a mostly a non-sensical value. That there is no warning at the moment is a feature to support the hack when generating a zipapp. In one local iteration I had changed the installation of the zipapp during build to pip, but this doesn't make the version available, so now the version just get's patched in rather ham-fistedly. if there were a warning here, 

I'm not opposed to adding war warning, I just don't know how to best add one (or make the current hack even more unpalatable).

behrmann: Made this slightly less ham-fisted to also support plain tar balls and it now it also warns for the ultimate fallback.

behrmann:
I've tested this with for mkosi, mkosi-addon, mkosi-initrd and mkosi-sandbox
- called as Python modules while in the repo
- via the shims while in the repo
- via the shims with a different working directory
- via the shims symlinked into `PATH`
- installed in a virtual environment generated with the standard library and installed via pip
- installed in a virtual environment generated with the standard library and installed via pip as an editable install
- installed in a virtual environment generated with uv and installed via uv pip
- installed in a virtual environment generated with uv and installed via uf pip as an editable install
- running the modules from an unpacked tarball
- running the shims from an unpacked tarball
- running a zipapp

All of them with a distribution mkosi installed and not installed. All output a version like `25.3.post1.dev244+g293af6b3.d20250404` except for the tarballs, which show the version from the static file.

mkosi/_version.py
```diff
@@ -0,0 +1,85 @@
...
+        tag, *rest = p.stdout.strip().split("-")
+        tag = tag.lstrip("v")
+        if rest:
+            numcommits, commit = rest
+            return f"{tag}.post1.dev{numcommits}+{commit}.d{datetime.datetime.now():%Y%m%d}"
+
+        # we are exactly on a tag
+        return tag
+    except (subprocess.CalledProcessError, NotADirectoryError, FileNotFoundError):
+        return None
```
DaanDeMeyer, line 62: So this completely goes against the convention we adopted in systemd to have the devel version be of the *next* version we've developing, which is fundamentally incompatible with `git describe` since that will always return the latest tag which is the *previous* version.

DaanDeMeyer (changes requested):

mkosi/_version.py
behrmann: To document the current state for others: 

We're a bit between the rock of UAPI version format spec and the hard place of [Python's public version identifiers](https://packaging.python.org/en/latest/specifications/version-specifiers/#public-version-identifiers). 

Both are compatible as far as the set of allowed characters for Python's scheme is (mostly) subset (disregarding the local part divided by + and the epoch divided by !, but `systemd-analyze compare-version` will happily treat them as separators, so that's fine), unfortunately the semantics are different. Whereas the UAPI spec follows RPM an .deb and has sigils that raise or lower a version, allowing not having to keep state and being able to compare version strings character by character, Python has multicharater substrings like `.dev` or `.rc` that take their place.

Using a version scheme anchored on the last release and counting commits since, allows to fit into both schemes, as described above, since this will hold
```
25 < 25.3 < 25.3.post1.dev244+g0eac3718 < 25.4 < 26
```
Unfortunately, there is no way to construct a version starting with the next release number, that compares smaller than that release under both the UAPI version spec and Python's, e.g. `26rc244` would compare smaller than 26 in under Python's scheme, but compares larger under the UAPI scheme, simply because it is shorter.

PR #3612, opened 2025-03-24
Always cache the default tools tree

DaanDeMeyer:
For the default tools tree, the cached image and the final image
are the same for all intents and purposes, so let's stop storing both
a cached image and a final image for the default tools tree and instead
only store the final image and use it as the cached image by always
writing a cache manifest next to it.

At the same time always use the name mkosi.tools to reduce the chance
of conflicts as "tools" is an incredibly common directory name.


behrmann (approved):
Looks reasonable to me, but I'll be surprised if the rework of `INVOKING_USER` won't break something, so far it always did.

PR #3613, opened 2025-03-24
tools: Add "devel" profile

DaanDeMeyer:
This profile contains tools required to build C/C++ projects. This is
useful to use mkosi sandbox to quickly build one off projects without
having to list a bunch of basic packages over and over again.


behrmann (approved):

septatrix:
Maybe explicitly add ninja, too?

PR #3616, opened 2025-03-25
Make sure tools tree recorded in history doesn't change when in sandbox

DaanDeMeyer:
Currently, when in the sandbox, the tools tree will always be recorded as "null" in the history. This causes problems if an image build is done inside of mkosi sandbox and mkosi is later invoked outside of the sandbox as the history will say no tools tree was used and so the tools tree won't be used, even if we're running outside of the sandbox.

To fix the issue, let's make sure we always record the proper tools tree, but let's not make use of it in the sandbox by moving the sandbox check to the Config object's tools() method.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -72,6 +72,7 @@ from mkosi.config import (
     expand_delayed_specifiers,
     format_bytes,
     have_history,
+    in_sandbox,
     parse_boolean,
     parse_config,
     resolve_deps,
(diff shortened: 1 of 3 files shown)


mkosi/__init__.py
```diff
@@ -4994,15 +4991,13 @@ def run_verb(args: Args, images: Sequence[Config], *, resources: Path) -> None:
...
         tools = None
 
     for i, config in enumerate(images):
-        if in_sandbox():
-            tools_tree = None
-        elif tools and config.tools_tree == Path("default"):
+        if tools and config.tools_tree == Path("default"):
             tools_tree = tools.output_dir_or_cwd() / tools.output
         else:
             tools_tree = config.tools_tree
```
behrmann, line 5003: mypy is complaining about this, but since it figures `tools_tree` should be `Path`. Adding `tools_tree: Optional[Path]` is one way to solve this, since it was obviously nullable before.

behrmann (approved):

PR #3620, opened 2025-03-25
Don't cache the package manager used in the tools tree cache manifest

DaanDeMeyer:
Doing this causes issues when using mkosi sandbox as have_cache(tools) will be different inside and outside of the sandbox when building Fedora and only dnf is available outside the sandbox but dnf5 is available inside the sandbox. Since we don't need to cache the package manager used anyway when building the default tools tree because we don't cache the package manager metadata either, don't cache it to avoid this problem.


mkosi/config.py
```diff
@@ -2251,7 +2251,16 @@ def cache_manifest(self) -> dict[str, Any]:
             "release": self.release,
             "mirror": self.mirror,
             "architecture": self.architecture,
-            "package_manager": self.distribution.package_manager(self).executable(self),
+            # Caching the package manager used does not matter for the default tools tree because we don't
+            # cache the package manager metadata for the tools tree either. In fact, it can cause issues when
+            # as the cache manifest for the tools tree will be different sometimes depending on whether we're
```
behrmann, line 2256:
```suggestion
            # cache the package manager metadata for the tools tree either. In fact, it can cause issues
            # as the cache manifest for the tools tree will sometimes be different depending on whether we're
```

Issue #3625, opened 2025-03-26
Arch build package metadata is added to base image


clarfonthey:
### mkosi commit the issue has been seen with

gfd76436f

### Used host distribution

Arch

### Used target distribution

Arch

### Linux kernel version used

6.12.17-1-lts

### CPU architectures issue was seen on

x86_64

### Unexpected behaviour you saw

Pacman keeps track of the metadata for installed packages in `/var/lib/pacman/local`. When installing build packages, the metadata for those packages is also included in the base image, rather than as part of the build overlay.

In the example config, you can see this as the directory `/var/lib/pacman/local/go-*` being present in the final image, despite the fact that the other files from the `go` package are not installed.

### Used mkosi config

```ini
# note: a mkosi.build script is required to actually generate the build overlay. I just set its contents to:
# #!/bin/bash
# echo "hi :)"

[Distribution]
Distribution=arch

[Output]
Format=directory
OutputDirectory=output/

[Build]
Incremental=true
BuildDirectory=build/
CacheDirectory=cache/
PackageCacheDirectory=pkgcache/
WorkspaceDirectory=workspace/
BuildSources=

[Content]
Packages=base
BuildPackages=go
```

### mkosi output

```sh
‣ Validating certificates and keys
‣ Populating pacman keyring
gpg: /etc/pacman.d/gnupg/trustdb.gpg: trustdb created
gpg: no ultimately trusted keys found
gpg: starting migration from earlier GnuPG versions
gpg: porting secret keys from '/etc/pacman.d/gnupg/secring.gpg' to gpg-agent
gpg: migration succeeded
==> Generating pacman master key. This may take some time.
gpg: Generating pacman keyring master key...
gpg: directory '/etc/pacman.d/gnupg/openpgp-revocs.d' created
gpg: revocation certificate stored as '/etc/pacman.d/gnupg/openpgp-revocs.d/E996ECFB05ACDBCD5170671257B76DA078EE3EE3.rev'
gpg: Done
==> Updating trust database...
gpg: marginals needed: 3  completes needed: 1  trust model: pgp
gpg: depth: 0  valid:   1  signed:   0  trust: 0-, 0q, 0n, 0m, 0f, 1u
==> Appending keys from archlinux.gpg...
==> Appending keys from archlinuxarm.gpg...
==> Appending keys from archzfs.gpg...
==> Locally signing trusted keys in keyring...
  -> Locally signed 10 keys.
==> Importing owner trust values...
[...]

PR #3631, opened 2025-03-27
pacman: Always bind mount /var/lib/pacman/local from sandbox

DaanDeMeyer:
We want any writes to /var/lib/pacman/local to go to any configured
overlayfs on /buildroot, instead of going directly to the
/var/lib/pacman/local directory which might be a lowerdir in the
overlayfs if one is used. Let's implement this by simply specifying
a path relative to the sandbox instead of specifying the path on the
host.

Fixes https://github.com/systemd/mkosi/issues/3625


DaanDeMeyer:
@clarfonthey Can you give this a try?

behrmann (approved):

clarfonthey (approved):

PR #3633, opened 2025-03-28
Allow combining --force and --rerun-build-scripts

DaanDeMeyer:
Until now we didn't allow this, but it turns out there's actually a use case for this, build the image if it doesn't exist yet, reuse the existing image otherwise.


behrmann (approved):
Not that it would have been covered, but these are the moments I miss the flowchart we had in the man page.

PR #3645, opened 2025-04-02
Don't chdir() to mkosi/ subdirectory

DaanDeMeyer:
Changing the working directory to mkosi/ is not intuitive. We want to parse configuration from there, but the working directory should not change (e.g. you don't want mkosi sandbox to execute inside the mkosi/ subdirectory when invoked in the top level directory).

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -4911,7 +4911,13 @@ def have_history(args: Args) -> bool:
     )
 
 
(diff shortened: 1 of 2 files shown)


behrmann (approved):
This indeed looks better, maybe the `imagedir` stuff could also be covered by a test to be on the safe side.

PR #3646, opened 2025-04-02
mkosi-initrd: fix vconsole configuration for Ubuntu

aafeijoo-suse:
Ubuntu (and Debian?) installs `/etc/vconsole.conf` as a dangling symlink to `/etc/default/keyboard`, and that causes the `mkosi-initrd` build to fail:

```
\u2023  Copying in extra file trees\u2026
cp: not writing through dangling symlink '/work/var/tmp/mkosi-workspace-us4prlhz/root/etc/vconsole.conf'
\u2023 "cp --recursive --no-dereference --preserve=mode,links --reflink=auto --copy-contents /work/etc/default/keyboard /work/var/tmp/mkosi-workspace-us4prlhz/root/etc/vconsole.conf" returned non-zero exit code 1.
bash-5.2# ls -l /work/var/tmp/mkosi-workspace-us4prlhz/root/etc/vconsole.conf
lrwxrwxrwx 1 0 0 16 Apr  2 14:49 /work/var/tmp/mkosi-workspace-us4prlhz/root/etc/vconsole.conf -> default/keyboard
bash-5.2# ls -l /work/var/tmp/mkosi-workspace-us4prlhz/root/etc/default/keyboard
ls: cannot access '/work/var/tmp/mkosi-workspace-us4prlhz/root/etc/default/keyboard': No such file or directory
```

Follow-up for 8856a2f1204fac775f8efe91bf155a91ec67e5fe

### DIFF
--- a/mkosi/initrd.py
+++ b/mkosi/initrd.py
@@ -181,14 +181,12 @@ def process_crypttab(staging_dir: Path) -> list[str]:
     return cmdline
 
 
-def add_raid_config() -> list[str]:
-    cmdline = []
-
-    for f in ("/etc/mdadm.conf", "/etc/mdadm.conf.d", "/etc/mdadm/mdadm.conf", "/etc/mdadm/mdadm.conf.d"):
-        if Path(f).exists():
-            cmdline += ["--extra-tree", f"{f}:{f}"]
-
-    return cmdline
+def raid_config() -> list[str]:
+    return [
+        f"--extra-tree={f}:{f}"
+        for f in ("/etc/mdadm.conf", "/etc/mdadm.conf.d", "/etc/mdadm/mdadm.conf", "/etc/mdadm/mdadm.conf.d")
+        if Path(f).exists()
+    ]
 
 
 def vconsole_config() -> list[str]:
@@ -315,7 +313,7 @@ def main() -> None:
         for p in args.profile:
             cmdline += ["--profile", p]
             if p == "raid":
-                cmdline += add_raid_config()
+                cmdline += raid_config()
 
         if args.kernel_image:
             cmdline += [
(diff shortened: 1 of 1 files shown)


behrmann (approved):
Urghs, I thought this hack had died already, but apparently no. Debian (and therefore Ubuntu?) use console-setup and I don't think what is pretending to be vconsole is anything but—@bluca will probably be able to tell the whole story.

mkosi/initrd.py
```diff
@@ -191,6 +191,16 @@ def add_raid_config() -> list[str]:
...
 
 
+def add_vconsole_config() -> list[str]:
+    cmdline = []
+
+    for f in ("/etc/default/keyboard", "/etc/vconsole.conf"):
+        if Path(f).exists():
+            cmdline += ["--extra-tree", f"{f}:{f}"]
+
+    return cmdline
```
keszybz, line 201: What about this:
```
def vconsole_config() -> Iterator[str]:
     for f in ("/etc/default/keyboard", "/etc/vconsole.conf"):  
            if Path(f).exists():
                yield from ["--extra-tree", f"{f}:{f}"]
```
and then
```
cmdline += list(vconsole_config())
```

behrmann: The call to list is actually not needed, `list += iterator` works, making this a bit prettier.

But staying with a plain list instead of an iterator would be clearer, in my opinion.
```python
def vconsole_config() -> list[str]:
     return [
        f"--extra-tree={f}:{f}"
        for f in ("/etc/default/keyboard", "/etc/vconsole.conf")
        if Path(f).exists()]
     ]
```

aafeijoo-suse: It makes sense, changed.

PR #3648, opened 2025-04-03
Set MKOSI_DEBUG instead of DEBUG

DaanDeMeyer:
DEBUG is very generic and can picked up by anything, but sometimes with drastically different meanings to what is supposed to be in the environment variable value.

For example the xfstests configure script expects DEBUG to contain compiler options, which makes it break in extremely hard to debug ways if we set DEBUG ourselves to a boolean value.

Let's namespace this to avoid conflicts.


DaanDeMeyer:
cc @aafeijoo-suse

behrmann (approved):
Flat namespaces are a bane :(

PR #3651, opened 2025-04-04
Add "plymouth" initrd profile

### LINKED ISSUE #3650: [RFC] plymouth initrd profile?
Does it make sense to consider an initrd profile for plymouth? I've been playing around with this to get the password to decrypt the rootfs and it works, but the packaging between distros is a mess (see below), maintaining this will be a pain... on the other hand, some users want a graphical interface at boot... opinions?

openSUSE Tumbleweed:
```
[Content]
Packages=
        cantarell-fonts
        distribution-logos-openSUSE-Tumbleweed
        plymouth
        plymouth-branding-openSUSE
```

Fedora rawhide:
```
[Content]
Packages=
        abattis-cantarell-fonts
        plymouth
        plymouth-system-theme
```

Ubuntu 21.10:
```
[Content]
Packages=
        kbd
        plymouth
        plymouth-themes
        plymouth-theme-ubuntu-text
```

aafeijoo-suse:
This initrd profile provides a graphical interface at boot (animation and password prompt). I only added support for what I've already tested successfully: openSUSE, Fedora and Ubuntu.

Closes #3650

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -627,6 +627,7 @@ class ToolsTreeProfile(StrEnum):
 class InitrdProfile(StrEnum):
     lvm = enum.auto()
     pkcs11 = enum.auto()
+    plymouth = enum.auto()
     raid = enum.auto()
 
 
--- a/mkosi/resources/man/mkosi-initrd.1.md
+++ b/mkosi/resources/man/mkosi-initrd.1.md
@@ -46,6 +46,8 @@ initrds and Unified Kernel Images for the current running system.
 
     The `lvm` profile enables support for LVM.
     The `pkcs11` profile enables support for PKCS#11.
+    The `plymouth` profile provides a graphical interface at boot (animation and
+    password prompt).
     The `raid` profile enables support for RAID arrays.
 
 `--debug`
--- a/mkosi/resources/man/mkosi.1.md
+++ b/mkosi/resources/man/mkosi.1.md
@@ -1043,6 +1043,8 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 
     The `lvm` profile enables support for LVM.
     The `pkcs11` profile enables support for PKCS#11.
+    The `plymouth` profile provides a graphical interface at boot (animation and
+    password prompt).
     The `raid` profile enables support for RAID arrays.
 
 `InitrdPackages=`, `--initrd-package=`
--- a/mkosi/resources/mkosi-initrd/mkosi.profiles/plymouth/mkosi.conf
+++ b/mkosi/resources/mkosi-initrd/mkosi.profiles/plymouth/mkosi.conf
@@ -1,0 +1,4 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+[Content]
+Packages=plymouth
--- a/mkosi/resources/mkosi-initrd/mkosi.profiles/plymouth/mkosi.conf.d/10-fedora.conf
+++ b/mkosi/resources/mkosi-initrd/mkosi.profiles/plymouth/mkosi.conf.d/10-fedora.conf
@@ -1,0 +1,9 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+[Match]
+Distribution=fedora
+
+[Content]
+Packages=
+        abattis-cantarell-fonts
+        plymouth-system-theme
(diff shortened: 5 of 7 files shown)


behrmann (commented):
Looks reasonable to me, it would be nice if we could tests this in CI

aafeijoo-suse:
> Looks reasonable to me, it would be nice if we could tests this in CI

No idea how to test this in the CI... We use to test GUIs in openQA, e.g.:

https://openqa.opensuse.org/tests/4973939#step/boot_encrypt/1

This is the test for a separate /boot not encrypted, / encrypted lvm. Performing a similar test, but building the initrd with mkosi instead of dracut, would be easy. Eventually, we will include tests like this one for mkosi-initrd. This whole openQA testsuite runs for every published Tumbleweed snapshot, e.g., for 20250405:

https://openqa.opensuse.org/tests/overview?distri=microos&distri=opensuse&version=Tumbleweed&build=20250405&groupid=1

DaanDeMeyer (changes requested):
I don't think the plymouth profile should be enabled by default yet.

aafeijoo-suse:
> I don't think the plymouth profile should be enabled by default yet.

Closing then.

DaanDeMeyer:
> > I don't think the plymouth profile should be enabled by default yet.
> 
> Closing then.

I didn't mean we couldn't add it, I was just saying for now it might be best if users have to explicitly enable it with `--profile plymouth` or `InitrdProfiles=plymouth`

aafeijoo-suse:
> I didn't mean we couldn't add it, I was just saying for now it might be best if users have to explicitly enable it with `--profile plymouth` or `InitrdProfiles=plymouth`

And this is how profiles work right now... what am I missing? I wouldn't like to enable any profiles by default either.

DaanDeMeyer:
> > I didn't mean we couldn't add it, I was just saying for now it might be best if users have to explicitly enable it with `--profile plymouth` or `InitrdProfiles=plymouth`
> 
> And this is how profiles work right now... what am I missing? I wouldn't like to enable any profiles by default either.

Ah right, indeed. My bad I forgot we changed this, can you reopen the PR? LGTM then actually.

aafeijoo-suse:
> Ah right, indeed. My bad I forgot we changed this, can you reopen the PR? LGTM then actually.

No problem :)

I'd understand if plymouth wasn't added, the packaging on each distro is terrible...

DaanDeMeyer (approved):

bluca:
I have tried to make plymouth work in Debian with the cryptsetup prompt, but it doesn't work. At best I get a black screen until I touch the fido key, and then a gray screen with three moving dots afterwards.

Do you need any particular kernel modules or something like that?

aafeijoo-suse:
> I have tried to make plymouth work in Debian with the cryptsetup prompt, but it doesn't work. At best I get a black screen until I touch the fido key, and then a gray screen with three moving dots afterwards.
> 
> Do you need any particular kernel modules or something like that?

I only tried this with Ubuntu, and there I had to append also `--profile=lvm`, otherwise the `device-mapper` udev rules are not added by any dependency, and cryptsetup hangs forever after entering the password with plymouth.

aafeijoo-suse:
> I only tried this with Ubuntu, and there I had to append also `--profile=lvm`, otherwise the `device-mapper` udev rules are not added by any dependency, and cryptsetup hangs forever after entering the password with plymouth.

FTR, this: https://gitlab.com/cryptsetup/cryptsetup/-/issues/869

bluca:
ie, the lvm2, kbd and plymouth-themes packages are already included

aafeijoo-suse:
> For me the unlock works, it's just the GUI integration that doesn't, it's just a black screen.

Ok, I just installed debian testing on a vm and this works for me:

```
[Content]
Packages=
        # plymouth default theme in debian is "emerald", which is shipped by desktop-base
        desktop-base
        kbd
        plymouth
        plymouth-themes

KernelModules=
        /drivers/gpu/drm
```

Please, no one ask me to try Arch... :)

bluca:
No difference on my laptop, still black screen. I guess the issue is graphics drivers, you need the full stack to make it work

aafeijoo-suse:
> No difference on my laptop, still black screen. I guess the issue is graphics drivers, you need the full stack to make it work

Yes, `--kernel-modules-include=host` helps here...

bluca:
This is a generic one built on OBS so can't use that. I think we need to wait for: https://fedoraproject.org/wiki/Changes/PlymouthUseSimpledrm

aafeijoo-suse:
Does it work if you set `KernelCommandLine=plymouth.use-simpledrm`? (you shouldn't need to add `KernelModules=` in this case)

bluca:
The issue is that debian does not build simpledrm, I've requested it

PR #3652, opened 2025-04-05
obs: fix sign-efi-sig-list parameter format

bluca:
It wants a date in the format Y-m-d H:M:S rather than unix seconds. Also if SOURCE_DATE_EPOCH is not defined, use actual epoch, as the timestamps must match when creating the signature and when attaching it


behrmann (approved):
Compared with [sign-efi-sig-list](https://git.kernel.org/pub/scm/linux/kernel/git/jejb/efitools.git/tree/sign-efi-sig-list.c#n166) that is the time string they are parsing, which is a bit confusing, since the example in the man page is `'Jul 21 09:39:37 BST 2012'`

PR #3660, opened 2025-04-07
Generate key-cert pair in config directory (fixes #3655)

### LINKED ISSUE #3655: `mkosi genkey` should generate keys under `mkosi/` if the config lies there

5c7e4bd7b146a1fd6d991263f46c14268a914dd9

### Used host distribution

_No response_

### Used target distribution

_No response_

### Linux kernel version used

_No response_

None

Running `mkosi genkey` when the mkosi config lies beneath `mkosi/mkosi.conf` generates the files in the top level dir instead

### Used mkosi config

```ini

```

### mkosi output

```sh

```

septatrix:
Returning the used configpath from `parse_config` seems like the easiest way to do this as they cannot be really attached to the image configs or args. If you have better suggestions I would be happy to change this

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -71,6 +71,7 @@ from mkosi.config import (
     cat_config,
     expand_delayed_specifiers,
     format_bytes,
+    get_configdir,
     have_history,
     in_sandbox,
     parse_boolean,
@@ -4478,8 +4479,10 @@ def generate_key_cert_pair(args: Args) -> None:
     keylength = 2048
     expiration_date = datetime.date.today() + datetime.timedelta(int(args.genkey_valid_days))
 
-    for f in ("mkosi.key", "mkosi.crt"):
-        if Path(f).exists() and not args.force:
+    configdir = get_configdir(args)
+
+    for f in (configdir / "mkosi.key", configdir / "mkosi.crt"):
+        if f.exists() and not args.force:
             die(
                 f"{f} already exists",
                 hint="To generate new keys, first remove mkosi.key and mkosi.crt",
@@ -4502,8 +4505,8 @@ def generate_key_cert_pair(args: Args) -> None:
             "-new",
             "-x509",
             "-newkey", f"rsa:{keylength}",
-            "-keyout", "mkosi.key",
-            "-out", "mkosi.crt",
+            "-keyout", configdir / "mkosi.key",
+            "-out", configdir / "mkosi.crt",
             "-days", str(args.genkey_valid_days),
             "-subj", f"/CN={args.genkey_common_name}/",
             "-nodes"
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -4964,6 +4964,20 @@ def finalize_default_tools(
     return Config.from_dict(context.finalize())
 
 
+def get_configdir(args: Args) -> Path:
+    """Allow locating all mkosi configuration in a mkosi/ subdirectory
+    instead of in the top-level directory of a git repository.
+    """
+    if (
+        args.directory is not None
+        and not (Path("mkosi.conf").exists() or Path("mkosi.tools.conf").exists())
+        and (Path("mkosi/mkosi.conf").is_file() or Path("mkosi/mkosi.tools.conf").exists())
+    ):
+        return Path.cwd() / "mkosi"
+
+    return Path.cwd()
+
+
 def parse_config(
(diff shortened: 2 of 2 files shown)


mkosi/__init__.py
```diff
@@ -4472,14 +4472,14 @@ def run_serve(args: Args, config: Config) -> None:
...
-def generate_key_cert_pair(args: Args) -> None:
+def generate_key_cert_pair(args: Args, configdir: Path) -> None:
     """Generate a private key and accompanying X509 certificate using openssl"""
 
     keylength = 2048
     expiration_date = datetime.date.today() + datetime.timedelta(int(args.genkey_valid_days))
 
     for f in ("mkosi.key", "mkosi.crt"):
-        if Path(f).exists() and not args.force:
+        if Path(configdir / f).exists() and not args.force:
```
behrmann, line 4482: since `configdir` is by construction already a `Path`.
```suggestion
        if (configdir / f).exists() and not args.force:
```

mkosi/__init__.py
```diff
@@ -4935,7 +4935,9 @@ def dump_history_json(tools: Optional[Config], images: Sequence[Config]) -> str:
     )
 
 
-def run_verb(args: Args, tools: Optional[Config], images: Sequence[Config], *, resources: Path) -> None:
+def run_verb(
+    args: Args, configdir: Path, tools: Optional[Config], images: Sequence[Config], *, resources: Path
```
behrmann, line 4939:
```suggestion
    args: Args, 
    configdir: Path, 
    tools: Optional[Config], 
    images: Sequence[Config], 
    *, 
    resources: Path,
```

mkosi/config.py
```diff
@@ -4952,7 +4952,7 @@ def parse_config(
     *,
     resources: Path = Path("/"),
     only_sections: Sequence[str] = (),
-) -> tuple[Args, Optional[Config], tuple[Config, ...]]:
+) -> tuple[Args, Path, Optional[Config], tuple[Config, ...]]:
```
behrmann, line 4955: At this point I wonder whether this function shouldn't instead return something structured instead and while I do see the logic of adding it in second place, adding it there is a weird bit of churn, fortunately we are the only users of this.

behrmann (commented):
Looks mostly reasonable to me, a few, nits. I think the reason this hadn't happen yet, was that @DaanDeMeyer wasn't sure as what to see the certs at this point. I do see the appeal of having everything in one place, but I think it's also a bit weird to generate something that's basically config slightly out of sight of a user, so I too am not entirely sure about this.

DaanDeMeyer (changes requested):
We can derive the config directory on demand. Just extract the logic from config.py into a function and reuse it in genkey to figure out the config directory.

DaanDeMeyer (approved):

septatrix:
> We can derive the config directory on demand. Just extract the logic from config.py into a function and reuse it in genkey to figure out the config directory.

Great idea, updated it now.

Regarding the message if the files already exist: Is the full path okay (they might be a bit long) or should I add a `relative_to(Path.cwd())`?

DaanDeMeyer:
Full is ok

septatrix:
I am not sure what these CI errors are telling me...?

mkosi/__init__.py
```diff
@@ -4478,11 +4479,16 @@ def generate_key_cert_pair(args: Args) -> None:
...
+
+    for f in (configdir / "mkosi.key", configdir / "mkosi.crt"):
+        if f.exists() and not args.force:
             die(
                 f"{f} already exists",
-                hint="To generate new keys, first remove mkosi.key and mkosi.crt",
+                hint=(
+                    "To generate new keys, first remove "
+                    f"{configdir / 'mkosi.key'} and {configdir / 'mkosi.crt'}"
+                ),
```
behrmann, line 4491: Let's not split this apart
```suggestion
                hint=(
                    f"To generate new keys, first remove {configdir / 'mkosi.key'} and {configdir / 'mkosi.crt'}"  # noqa: E501
                ),
```
or just keep it without configdir, since that can already be read of from the line above and having it twice there, this could become quite long on screen.

DaanDeMeyer:
@septatrix Just rebase on latest main to fix those

DaanDeMeyer:
I'll merge tomorrow since this needs a follow up PR in systemd

PR #3661, opened 2025-04-07
Add new Ssh=runtime option

### LINKED ISSUE #3654: Add option for `Ssh=` to not install custom units
New enough distros ship with `systemd-ssh-generator` themselves so mkosi does not need to add anything there anymore. To the contrary, it rather runs the risk of overwriting existing files by the distro or having slightly incompatible unit version (e.g. the distro does not use `sshd-keygen.target` leading to failures).

I can disable the generator via the command line but then I would only get mkosi's potentially broken units. I could also remove `/usr/lib/systemd/system-preset/80-mkosi-ssh.preset` but that does not solve the problem of mkosi potentially overwriting existing units (and generally leaving unnecessary files around)

Completely disabling `Ssh=` works in theory but has the disadvantage that one does not get [the warning if they forgot to generate ssh keys](https://github.com/systemd/mkosi/blob/de70e3f52a6982132097f67d83b9a383b48fb86e/mkosi/qemu.py#L949)

septatrix:


mkosi/qemu.py
```diff
@@ -1133,6 +1129,12 @@ def run_qemu(args: Args, config: Config) -> None:
     ):
         die("SecureBootCertificate= must be configured to use FirmwareVariables=custom|microsoft-mok")
 
+    if config.ssh in (Ssh.yes, Ssh.runtime) and (not config.ssh_key or not config.ssh_certificate):
+        die(
+            "SSH key and certificate must be configured to use SSH=yes|runtime",
```
behrmann, line 1134:
```suggestion
            "SSH key and certificate must be configured to use Ssh=yes or Ssh=runtime",
```

mkosi/resources/man/mkosi.1.md
behrmann, line 1: The `runtime` option should be explained.

mkosi/config.py
```diff
@@ -1979,7 +1985,7 @@ class Config:
 
     autologin: bool
     make_initrd: bool
-    ssh: bool
+    ssh: Ssh
```
behrmann, line 1988: This needs accompanying changes in the JSON test.

behrmann (commented):
I agree that we should handle the case of this being handled by the ssh-generator, I think exposing this to users is a reasonable choice, that might be useful for other things, but maybe it would be easier to just check for the existence of the generator in the image instead? `configure_ssh` is run after all packages are installed, so this information would be available at this point, also finalize scripts run after this and could still remove config we generate, so that might be a smaller change all in all.

septatrix:
> ...I think exposing this to users is a reasonable choice, that might be useful for other things, but maybe it would be easier to just check for the existence of the generator in the image instead?

I first and foremost wanted an option that bails out starting the VM when no SSH keys are present without putting files in the image. I also though about adding an "auto" option to `Ssh=` but it would either need to still default to "off", or not bail out on missing keys as that would otherwise break compat.

Now that I think about it, what would you say about `Ssh=always|never|detect` (i.e. not calling it "auto") and defaulting to "never". The first two are self explanatory, the last one would install units iff the ssh-generator binary does not exist, and always bail out on missing SSH keys?

septatrix:
Or do you suggest to just overwrite the behavior of `Ssh=yes`?

behrmann:
Can you also move the fixes in the commit message out of the subject?

mkosi/resources/man/mkosi.1.md
```diff
@@ -1941,10 +1947,10 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 `SshKey=`, `--ssh-key=`
 :   Path to the X.509 private key in PEM format to use to connect to a
     virtual machine started with `mkosi vm` and built with the `Ssh=`
-    option enabled via the `mkosi ssh` command. If not configured and
-    `mkosi.key` exists in the working directory, it will automatically be
-    used for this purpose. Run `mkosi genkey` to automatically generate
-    a key in `mkosi.key`.
+    option enabled (or **systemd-ssh-generator** installed) via the `mkosi ssh` command.
```
behrmann, line 1950:
```suggestion
    option enabled (or with **systemd-ssh-generator** installed) via the `mkosi ssh` command.
```

mkosi/resources/man/mkosi.1.md
```diff
@@ -1154,8 +1154,14 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     used as an initramfs.
 
 `Ssh=`, `--ssh=`
-:   If specified, an **sshd** socket unit and matching service are installed
-    in the final image that expose SSH over VSock. When building with this
+:   Specifies whether to install an **sshd** socket unit and matching service
+    in the final image. Takes one of `always`, `never`, `auto` or `runtime`.
+    Defaults to `auto`. If set to `auto` and `systemd-ssh-generator`
```
behrmann, line 1159: That's a bit dense, let's split this up slightly.
```suggestion
    in the final image. Takes one of `always`, `never`, `auto` or `runtime`.
    Defaults to `auto`. 
    
    If set to `auto` and `systemd-ssh-generator`
```

Also, to maybe make this clearer to users who might wondern about different packaging among distros
```suggestion
    in the final image. Takes one of `always`, `never`, `auto` or `runtime`.
    Defaults to `auto`.
    
    If set to `auto` and the generator binary `systemd-ssh-generator`
```

mkosi/resources/man/mkosi.1.md
```diff
@@ -1154,8 +1154,14 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
+    is not preset in the image, or set to `always`, mkosi will install **sshd** units
+    in the final image that expose SSH over VSock. If set to `never`,
+    mkosi will not install the **sshd* units. If the `runtime` value is used,
```
behrmann, line 1162: Nit: slightly less doubling.
```suggestion
    mkosi will not install these units. If the `runtime` value is used,
```

behrmann (approved):

DaanDeMeyer (approved):

PR #3662, opened 2025-04-08
Allow MinimumVersion to be a git sha

DaanDeMeyer:
In systemd we very often require newer mkosi git commits which means that setting MinimumVersion=25~devel doesn't work anymore. Let's allow the MinimumVersion= to be a git sha so that we can enforce the mkosi version to be new enough.

We skip git commit checks if we're in the sandbox since mkosi will be a zipapp there and we won't have access to the git repository to do the check.

### DIFF
--- a/mkosi/resources/man/mkosi.1.md
+++ b/mkosi/resources/man/mkosi.1.md
@@ -2166,6 +2166,12 @@ config file is read:
 :   The minimum **mkosi** version required to build this configuration. If
     specified multiple times, the highest specified version is used.
 
+    The minimum version can also be specified as a unabbreviated git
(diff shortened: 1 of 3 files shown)


mkosi/resources/man/mkosi.1.md
```diff
@@ -2166,6 +2166,12 @@ config file is read:
 :   The minimum **mkosi** version required to build this configuration. If
     specified multiple times, the highest specified version is used.
 
+    The minimum version can also be specified as a git commit hash, in
```
behrmann, line 2169:
```suggestion
    The minimum version can also be specified as a unabbreviated git commit hash, in
```

behrmann (approved):

PR #3664, opened 2025-04-08
action: Unshallow repository

DaanDeMeyer:
We want the full history so that MinimumVersion= with a git sha
works properly.


DaanDeMeyer:
I just double checked this works, so let's not wait for CI again

PR #3672, opened 2025-04-08
Don't use gen_kernel_images() in run_depmod()

### LINKED ISSUE #3670: mkosi-initrd: Ubuntu/Debian default initrd images contain all kernel modules

main

### Used host distribution

Ubuntu 24.10

### Used target distribution

Ubuntu 24.10

### Linux kernel version used

6.11.0-21-generic

x86_64

Ubuntu/Debian does not ship the kernel image under `/usr/lib/modules/<kver>`... so `vmlinuz` is not found, the "Applying kernel module filters" step is never reached, and the size of the initrd images is +600MB.

https://github.com/systemd/mkosi/blob/191e9486edd7b7c648e6b611010c036f09791719/mkosi/bootloader.py#L626-L635

This issue could be addressed manually using the `--kernel-image` option, but I think it should be fixed somehow by default.

### Used mkosi config

```ini

```

### mkosi output

```sh
root@dev:/home/dev# mkosi-initrd -o test.img
...
\u2023  /tmp/tmpd2v0jivu/test.img.cpio.zst size is 663.3M, consumes 663.3M.
\u2023 Copying /tmp/tmpd2v0jivu/test.img to /home/dev/test.img.new
\u2023 Moving /home/dev/test.img.new to /home/dev/test.img
...
root@dev:/home/dev# mkosi-initrd --kernel-image /boot/vmlinuz-$(uname -r) -o test.img
...
\u2023  /tmp/tmpjl6yicze/test.img.cpio.zst size is 42.2M, consumes 42.2M.
\u2023 Copying /tmp/tmpjl6yicze/test.img to /home/dev/test.img.new
\u2023 Moving /home/dev/test.img.new to /home/dev/test.img
```

DaanDeMeyer:
We don't actually care about kernel images in this function, so let's not depend on those being present.

Fixes #3670

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2967,10 +2967,13 @@ def run_depmod(context: Context, *, cache: bool = False) -> None:
         return
 
     if not cache:
-        for kver, _ in gen_kernel_images(context):
+        for modulesd in (context.root / "usr/lib/modules").glob("*"):
+            if not modulesd.is_dir():
+                continue
+
             process_kernel_modules(
                 context,
-                kver,
+                modulesd.name,
                 modules_include=finalize_kernel_modules_include(
                     context,
                     include=context.config.kernel_modules_include,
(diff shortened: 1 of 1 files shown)


behrmann (approved):
Nit: sprinkling in some `kver = modulesd.name` would make the diff smaller and I think the intentation allaround clearer.

PR #3674, opened 2025-04-09
Allow PCR signing settings to be overridden in sub-images

mcassaniti:
A sub-image is usually built from the same base image. That base image may be used for a rescue image and a primary workstation image. If the rescue image has PCR signing enabled then it could allow the disk to be automatically unlocked. This change allows specifying separate PCR signing keys which will support only the workstation sub-image performing an automatic unlock.

### DIFF
--- a/mkosi/resources/man/mkosi.1.md
+++ b/mkosi/resources/man/mkosi.1.md
@@ -2913,10 +2913,6 @@ following settings are universal and cannot be configured in subimages:
 - `VerityCertificateSource=`
 - `VerityKey=`
 - `VerityKeySource=`
-- `SignExpectedPcrCertificate=`
-- `SignExpectedPcrCertificateSource=`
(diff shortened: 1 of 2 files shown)


mkosi/resources/man/mkosi.1.md
```diff
@@ -2934,6 +2934,10 @@ down to subimages but can be overridden:
 - `CacheKey=`
 - `BuildKey=`
 - `CompressLevel=`
+- `SignExpectedPcrKey=`
+- `SignExpectedPcrKeySource=`
+- `SignExpectedPcrCertificate=`
+- `SignExpectedPcrCertificateSource=`
```
DaanDeMeyer, line 2940: These should be removed from the universal settings listed above.

DaanDeMeyer (approved):

PR #3676, opened 2025-04-09
zypper: do not fail if a package configured to be removed is not found

aafeijoo-suse:
`dnf` does not fail in this case:

```
‣  Removing 1 packages…
No matches were found for the following plugin name patterns while enabling libdnf plugins: builddep, reflink, versionlock
No matches were found for the following plugin name patterns while disabling libdnf plugins: *
No packages to remove for argument: shadow-utils

Nothing to do.
‣  Removing files…
```

But `zypper` fails by default, breaking the build. Found with mkosi-initrd in openSUSE, the `shadow` package is not installed when using `WithRecommends=yes`.

```
‣  Removing 2 packages…
Warning: Enforced setting: $releasever=20241001
Reading installed packages...
'shadow' not found in package names. Trying capabilities.
No provider of 'shadow' found.
Resolving package dependencies...

The following package is going to be REMOVED:
  sysuser-shadow

1 package to remove.

Package install size change:
              |         0 B    required by packages that will be installed
    -3.0 KiB  |  -    3.0 KiB  released by packages that will be removed

Backend:  classic_rpmtrans
Continue? [y/n/v/...? shows all options] (y): y
(1/1) Removing: sysuser-shadow-3.3-8.2.noarch ..........................................................................................[done]
Installation has completed with error.
‣ "zypper --installroot=/buildroot --cache-dir=/var/cache/zypp --non-interactive --no-refresh --releasever=20241001 remove --clean-deps shadow sysuser-shadow" returned non-zero exit code 104.
```

This behavior can be changed using the `--ignore-unknown` global option (which goes before the command).


mkosi/installer/zypper.py
```diff
@@ -126,7 +126,8 @@ def invoke(
         stdout: _FILE = None,
     ) -> CompletedProcess:
         return run(
-            cls.cmd(context) + [operation, *arguments],
+            cls.cmd(context)
+            + [*(["--ignore-unknown"] if operation == "remove" else []), operation, *arguments],
```
DaanDeMeyer, line 130: Please add an `options` argument and pass this in instead of putting it in `invoke()` itself.

DaanDeMeyer (changes requested):

mkosi/installer/zypper.py
aafeijoo-suse: Sure, changed.

PR #3678, opened 2025-04-10
Rework history <=> sandbox integration

DaanDeMeyer:
Until now, we read the tools tree configuration both inside and outside the sandbox. This is problematic for a few reasons:

- For a number of reasons, the tools tree configuration inside and outside the sandbox may be different. This is especially true when the history is used, as when invoking mkosi sandbox and mkosi is invoked again within the sandbox, we don't know when invoking mkosi sandbox whether the nested invocation will reuse the history or not. This can lead to situations where mkosi sandbox thinks the tools tree is up-to-date whereas mkosi inside the sandbox thinks the tools tree is out of date, which is a situation that is very hard to resolve for users.
- Even if we read the tools tree configuration inside the sandbox, there's no real point to it except maybe for showing the summary. Since the tools tree has already been built and mounted at that point, we can't build or use a new tools tree, or clean up an old one.

Let's fix these issues by *not* reading the tools tree configuration inside the sandbox, neither from configuration not from history. The one piece of tools tree configuration we do need to pass into the sandbox is the default tools tree path as that gets included in the history of the other images so we pass it in via an environment variable $MKOSI_DEFAULT_TOOLS_TREE_PATH.

As a consequence of this approach, we cannot write the tools tree history from within the sandbox anymore, so we write it to a separate file from outside the sandbox instead.

Additionally, we make more commands use the history. Up until now clean, summary and other verbs did not make use of the history. This doesn't really make sense, so let's make sure they make use of the history as well. summary in particular is important as it may be used to query information about the built image after building it which needs to take the history into account.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -108,6 +108,9 @@ class Verb(StrEnum):
             Verb.dependencies,
         )
 
+    def needs_tools(self) -> bool:
+        return self in (Verb.sandbox, Verb.journalctl, Verb.coredumpctl, Verb.ssh)
+
     def needs_build(self) -> bool:
         return self in (
             Verb.build,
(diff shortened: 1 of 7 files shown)


mkosi/config.py
```diff
@@ -4904,13 +4907,22 @@ def finalize(self) -> dict[str, Any]:
...
+        return False
+
+    if args.verb in (Verb.build, Verb.cat_config):
+        return False
+
+    if args.verb.needs_tools() and args.force > 0:
+        return False
+
+    if args.verb.needs_build() and args.force > 0 and not args.rerun_build_scripts:
+        return False
```
behrmann, line 4923: The lower one cannot be hit, because if `args.verb.needs_build()` return true and `args.force` is non-zero, the upper one will already have returned `False`.

mkosi/config.py
```diff
@@ -5007,8 +5019,11 @@ def parse_config(
     if args.debug_sandbox:
         ARG_DEBUG_SANDBOX.set(args.debug_sandbox)
 
+    if args.rerun_build_scripts and not args.verb.needs_build():
+        die(f"--rerun-build-scripts cannot be used with '{args.verb}' command")
+
     if args.cmdline and not args.verb.supports_cmdline():
-        die(f"Arguments after verb are not supported for {args.verb}.")
+        die(f"Arguments after verb are not supported for '{args.verb}' command")
```
behrmann, line 5026: This should either have an article
```suggestion
        die(f"Arguments after verb are not supported for the '{args.verb}' command")
```
or change order
```suggestion
        die(f"Arguments after verb are not supported for command '{args.verb}'")
```

mkosi/config.py
DaanDeMeyer: Is this right? The upper one uses needs_tools() and the lower one uses needs_build().

behrmann (approved):

PR #3681, opened 2025-04-10
Take shared lock in copy_ephemeral()

DaanDeMeyer:
We don't need an exclusive lock to make this copy, so let's take only a shared lock.

### DIFF
--- a/mkosi/util.py
+++ b/mkosi/util.py
@@ -139,9 +139,9 @@ def flock(path: Path, flags: int = fcntl.LOCK_EX) -> Iterator[int]:
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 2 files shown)


mkosi/util.py
```diff
@@ -139,9 +139,9 @@ def flock(path: Path, flags: int = fcntl.LOCK_EX) -> Iterator[int]:
 
 
 @contextlib.contextmanager
-def flock_or_die(path: Path) -> Iterator[Path]:
+def flock_or_die(path: Path, flags: int = fcntl.LOCK_EX) -> Iterator[Path]:
```
behrmann, line 142: The change to this function is only for symmetry with the definition of `flock`, right? Confused me a bit since it's in the same commit.

PR #3685, opened 2025-04-12
Fail if --rerun-build-scripts is used and tools is out of date

DaanDeMeyer:
If --rerun-build-scripts is used, we shouldn't rebuild the tools tree unless --force is specified as well, so make sure we check for that specific case.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -5030,8 +5030,7 @@ def run_verb(args: Args, tools: Optional[Config], images: Sequence[Config], *, r
     if tools and not have_cache(tools):
         if (args.rerun_build_scripts or args.verb != Verb.build) and args.force == 0:
             die(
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -5028,10 +5028,9 @@ def run_verb(args: Args, tools: Optional[Config], images: Sequence[Config], *, r
 
     # For the default tools tree have_cache() encompasses the "has the tools tree been built at all" check.
     if tools and not have_cache(tools):
-        if args.verb != Verb.build and args.force == 0:
+        if (args.rerun_build_scripts or args.verb != Verb.build) and args.force == 0:
             die(
-                f"Default tools tree requested for image '{last.image}' but it is out-of-date or has not "
-                "been built yet",
+                f"Default tools tree requested but it is out-of-date or has not been built yet",
```
behrmann, line 5033: ruff complains about this still being an f-string.

PR #3688, opened 2025-04-14
mkosi-obs: append certs from mkosi.uefi.db/ to 'db'

bluca:
Allow appending an arbitrary number of certs to the autoenrolled UEFI db. Expects list of PEM formatted files. Only implemented for OBS builds for now.

### DIFF
--- a/mkosi/resources/mkosi-obs/mkosi.postoutput
+++ b/mkosi/resources/mkosi-obs/mkosi.postoutput
@@ -110,6 +110,15 @@ if ((${#DDIS[@]} > 0)); then
     cert-to-efi-sig-list -g "$guid" /usr/src/packages/SOURCES/_projectcert.crt db.esl
     cp db.esl KEK.esl
     cp db.esl PK.esl
+
+    # If there are any additional certs, concatenate them
+    for cert in /usr/src/packages/SOURCES/*/mkosi.uefi.db/*.crt; do
+        test -f "$cert" || continue
+        cert-to-efi-sig-list -g "$guid" "$cert" tmp.esl
+        cat tmp.esl >>db.esl
+        rm -f tmp.esl
+    done
+
     for i in *.esl; do
         sign-efi-sig-list -o -g "$guid" -t "$(date -d "@${SOURCE_DATE_EPOCH:-0}" "+%Y-%m-%d %H:%M:%S")" "${i%.esl}" "$i" "${i%.esl}.auth"
     done


mkosi/resources/man/mkosi.1.md
```diff
@@ -1208,6 +1208,11 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     `secure-boot-enroll manual` in it. Auto enrollment is not supported on
     systemd versions older than v252. Defaults to `yes`.
 
+`SecureBootAutoEnrollCertificates=`, `--secure-boot-auto-enroll-certificates=`
+:   If `SecureBootAutoEnroll=` is enabled, appends additional certificates to
+    UEFI db. Expects comma-separated list of paths to PEM formatted X.509
+    certificates. Only implemented for images built on OBS for now.
```
behrmann, line 1214: For `SecureBootAutoEnroll` we have a pre-257 and a post-257 code path. The pre-257 codepath actually seems more amenable to adding this quickly, the post-257 might need some work in bootctl, since, I've only looked at the man page, it seems to require cert and key.

Generally I'd prefer to have this option hidden and undocumented as long as it cannot be used outside of OBS.

mkosi/resources/man/mkosi.1.md
```diff
@@ -1208,6 +1208,11 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     `secure-boot-enroll manual` in it. Auto enrollment is not supported on
     systemd versions older than v252. Defaults to `yes`.
 
+`SecureBootAutoEnrollCertificates=`, `--secure-boot-auto-enroll-certificates=`
+:   If `SecureBootAutoEnroll=` is enabled, appends additional certificates to
+    UEFI db. Expects comma-separated list of paths to PEM formatted X.509
```
behrmann, line 1213:
```suggestion
:   If `SecureBootAutoEnroll=` is enabled, appends additional certificates to the
    UEFI db. Expects a comma-separated list of paths to PEM formatted X.509
```

mkosi/resources/mkosi-obs/mkosi.postoutput
```diff
@@ -110,6 +110,15 @@ if ((${#DDIS[@]} > 0)); then
     cert-to-efi-sig-list -g "$guid" /usr/src/packages/SOURCES/_projectcert.crt db.esl
     cp db.esl KEK.esl
     cp db.esl PK.esl
+
+    # If there are any additional certs, concatenate them
+    while read -r cert; do
+        rm -f tmp.esl
```
behrmann, line 116: Removing this file both at the beginning and end of the loop seems gratuitous, since `tmp.esl` is not used elsewhere in the script.
```suggestion
```

mkosi/resources/man/mkosi.1.md
bluca: I don't think it makes sense to spend time to add support for something new to a legacy code path, so I've removed it from the manpage for now

mkosi/config.py
```diff
@@ -3345,6 +3346,14 @@ def parse_kernel_module_filter_regexp(p: str) -> str:
         default=True,
         help="Automatically enroll the secureboot signing key on first boot",
     ),
+    ConfigSetting(
+        dest="secure_boot_auto_enroll_certificates",
+        long="--secure-boot-auto-enroll-certificates",
+        metavar="PATH",
+        section="Validation",
+        parse=config_make_list_parser(delimiter=",", parse=make_path_parser()),
+        help="List of comma-separated paths to PEM certificates to be automatically enrolled in UEFI db",
```
behrmann, line 3355:
```suggestion
        help=,argparse.SUPPRESS
```
so it's not show in the help.

mkosi/config.py
```diff
@@ -5404,6 +5413,7 @@ def summary(config: Config) -> str:
          {bold("VALIDATION")}:
                     UEFI SecureBoot: {yes_no(config.secure_boot)}
          UEFI SecureBoot AutoEnroll: {yes_no(config.secure_boot_auto_enroll)}
+   UEFI SecureBoot AutoEnroll Certs: {line_join_list(config.secure_boot_auto_enroll_certificates)}
```
behrmann, line 5416: This could be optionally written out, like e.g. `config.kernel_modules_exclude`

mkosi/resources/man/mkosi.1.md
behrmann: Sorry if I was unclear, I didn't want to imply that it should be added to the legacy codepath, I'm totally fine with adding it to the post-257 codepath only. I just wanted to say that it seems to need a bit more work outside of mkosi. I do think this could be useful and you find the time to get around to it.

That being said, while that work is undergoing, I prefer this option not to be user-visible.

behrmann (approved):

DaanDeMeyer (changes requested):
Please add support to bootctl --secure-boot-auto-enroll first so we can support the feature both for local and OBS builds

bluca:
I don't really have a use for that for local builds, so I don't want to spend my time on it

behrmann:
If there's no immediate intention to add this feature to bootctl, maybe go with an environment variable with space-separated paths set in the profile to pass the extra certificates to the script instead.

DaanDeMeyer (approved):

PR #3692, opened 2025-04-15
Don't try to copy selinux xattrs on zfs either

### LINKED ISSUE #3684: Remove dependency on git

mkosi 26~devel

### Used host distribution

Fedora 41

### Used target distribution

N/A

### Linux kernel version used

6.13.9-200.fc41.x86_64

x86_64

```
❯  mkosi -f sandbox -- meson setup build

‣ Cannot check mkosi git version, not running mkosi from a git repository
```

Installed with README.md provided instructions
```
dnf config-manager addrepo --from-repofile=https://download.opensuse.org/repositories/system:systemd/Fedora_41/system:systemd.repo
dnf install mkosi
```

### Used mkosi config

```ini
N/A
```

### mkosi output

```sh
N/A
```

DaanDeMeyer:
This also fails with permission denied errors so let's skip it on zfs as well.

Fixes #3684

@eadwu this should fix your selinux issue

### DIFF
--- a/mkosi/tree.py
+++ b/mkosi/tree.py
@@ -13,7 +13,15 @@ from pathlib import Path
 from mkosi.config import ConfigFeature
 from mkosi.log import ARG_DEBUG, die
 from mkosi.run import SandboxProtocol, nosandbox, run, workdir
-from mkosi.sandbox import BTRFS_SUPER_MAGIC, FS_NOCOW_FL, OVERLAYFS_SUPER_MAGIC, chattr, lsattr, statfs
+from mkosi.sandbox import (
+    BTRFS_SUPER_MAGIC,
+    FS_NOCOW_FL,
+    OVERLAYFS_SUPER_MAGIC,
+    ZFS_SUPER_MAGIC,
+    chattr,
+    lsattr,
+    statfs,
+)
 from mkosi.util import PathString, flatten
 from mkosi.versioncomp import GenericVersion
 
@@ -113,7 +121,10 @@ def copy_tree(
         attrs += ",timestamps,ownership"
 
         # Trying to copy selinux xattrs to overlayfs fails with "Operation not supported" in containers.
-        if statfs(os.fspath(dst.parent)) != OVERLAYFS_SUPER_MAGIC or not tree_has_selinux_xattr(src):
+        if statfs(os.fspath(dst.parent)) not in (
+            OVERLAYFS_SUPER_MAGIC,
+            ZFS_SUPER_MAGIC,
+        ) or not tree_has_selinux_xattr(src):
             attrs += ",xattr"
 
     def copy() -> None:
--- a/mkosi/sandbox.py
+++ b/mkosi/sandbox.py
@@ -57,6 +57,7 @@ NR_open_tree = 428
 OPEN_TREE_CLOEXEC = os.O_CLOEXEC
 OPEN_TREE_CLONE = 1
 OVERLAYFS_SUPER_MAGIC = 0x794C7630
+ZFS_SUPER_MAGIC = 0x0CB1BA00
 PR_CAP_AMBIENT = 47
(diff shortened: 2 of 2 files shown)


mkosi/tree.py
```diff
@@ -113,7 +121,10 @@ def copy_tree(
         attrs += ",timestamps,ownership"
 
         # Trying to copy selinux xattrs to overlayfs fails with "Operation not supported" in containers.
-        if statfs(os.fspath(dst.parent)) != OVERLAYFS_SUPER_MAGIC or not tree_has_selinux_xattr(src):
+        if statfs(os.fspath(dst.parent)) not in (
+            OVERLAYFS_SUPER_MAGIC,
+            ZFS_SUPER_MAGIC,
```
behrmann, line 126: I'm not sure this is the best solution. I think the problem comes from zfs just being such a foreign object with terrible non-Linux defaults. It should work with the zfs property `xattr=sa` set, which, if I remember correctly puts the xattr data into the inodes. That means, if this works on a properly configured zfs, that should be fixed on user side and put into our FAQ instead of handling zfs things explicity here.

mkosi/tree.py
```diff
@@ -113,7 +121,10 @@ def copy_tree(
         attrs += ",timestamps,ownership"
 
         # Trying to copy selinux xattrs to overlayfs fails with "Operation not supported" in containers.
```
behrmann, line 123: The comment should probably be updated.

behrmann (commented):
LGTM codewise, one nit, but I think this should work on zfs as well.

mkosi/tree.py
eadwu: Oh ...

Mine is indeed not ```
xattr     on     inherited 
```

`xattr=on` is the default value though.

**Comment by eadwu** [comment]
I don't seem to be hitting the conditional?
```
particleos on  obs/fedora-41 [✘!+?] 
❯ ../mkosi/bin/mkosi -d fedora -r 41 --cache-only=never --console=gui -f
cp: setting attribute 'security.selinux' for 'security.selinux': Operation not permitted
‣ "cp --recursive --no-dereference --preserve=mode,links,timestamps,ownership,xattr --reflink=auto --copy-contents /work/usr/share/crypto-policies/back-ends/DEFAULT /work/var/home/sss/.cache/mkosi/mkosi-workspace-jd3094z7/sandbox/etc/crypto-policies/back-ends --no-target-directory" returned non-zero exit code 1.

particleos on  obs/fedora-41 [✘!+?] 
❯  rg ZFS_SUPER_MAGIC ../mkosi           
../mkosi/mkosi/tree.py
20:    ZFS_SUPER_MAGIC,
126:            ZFS_SUPER_MAGIC,

../mkosi/mkosi/sandbox.py
60:ZFS_SUPER_MAGIC = 0x0CB1BA00
```

File: mkosi/tree.py (hunk 1 above)
Comment (behrmann) [reply mkosi/tree.py:126]: Ok, so you seem to be running zfs 2.3, since `xattr=sa` is shown as `xattr=on`. 

Do you have any selinux labels in the first place? Probably, since you are running Fedora? In the issue you never mentioned what target distribution you were using. Does it support selinux? Are you using a tools tree? Honestly, zfs says it supports selinux (it even has some properties for it), if selinux-enabled tooling on a filesystem that purports to support selinux, apparently with the necessary options set to support selinux, is not able to set labels, that seems like a pretty outrageous bug on zfs side.

Comment (eadwu) [reply mkosi/tree.py:126]: I boot with `selinux=0` since first install and installed Fedora 41 through bootc.

It does not look like I have any selinux attributes to begin with.
```
❯  ls -laZ ~/.cache/.                    
total 508
drwxr-xr-x 32 sss users ?       40 Apr 17 12:45  .
drwx------ 38 sss users ?       56 Apr 17 12:32  ..
```

PR #3696, opened 2025-04-15
Only write CLI arguments to history instead of full config

DaanDeMeyer:
The huge downside of writing the full config to the history is that
any modifications to config files are ignored until the image is
rebuilt.

If we make the argument that changes to config files are usually
supposed to be permanent rather than ephemeral, it doesn't make sense
to not take them into account when reusing the history. Especially since
the primary purpose of the history is to avoid specifying the same CLI
arguments over and over again without encoding those CLI arguments into
a config file.

So instead of saving the full config to the history and not reparsing
the configuration when reusing the history, let's only save the CLI
arguments to the history and always reparse the configuration, even when
we end up reusing the history.

This keeps the crucial benefit of the history intact (not having to repeat
yourself endlessly) while still making sure any new settings added to the
configuration files are taken into account immediately.

It also simplifies the implementation quite a bit, as we can get away
again with maintaining a single history file instead of two.

Additionally, we opt to completely ignore the history when using the sandbox
verb. The idea here is that the history only applies to the final image (and
subimages). This was already the case as any --tools-tree-xxx CLI options aren't
saved in the history as they are not fields of the Config object.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3976,10 +3976,7 @@ def build_image(context: Context) -> None:
         check_root_populated(context)
         run_build_scripts(context)
 
-        if context.config.output_format == OutputFormat.none or (
-            context.args.rerun_build_scripts
-            and (context.config.output_dir_or_cwd() / context.config.output).exists()
-        ):
+        if context.config.output_format == OutputFormat.none or context.args.rerun_build_scripts:
             return
 
         if wantrepo:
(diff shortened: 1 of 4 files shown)


behrmann (commented):
Nice, this approach looks quite a bit cleaner.

mkosi/resources/man/mkosi.1.md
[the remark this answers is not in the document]
DaanDeMeyer: Not possible, we don't have a config object yet when we run it

mkosi/__init__.py
```diff
@@ -5008,6 +4987,10 @@ def run_verb(args: Args, tools: Optional[Config], images: Sequence[Config], *, r
     # The images array has been modified so we need to reevaluate last again.
     last = images[-1]
 
+    if args.verb == Verb.bump:
+        finalize_image_version(args, last)
```
septatrix, line 4991: Should this still check if `mkosi.version` is executable? Even worse, if it doesn't check but overwrite the file content it is still marked executable and the next invocation of mkosi will try to execute it (which fails)

DaanDeMeyer: We check if it's executable during config parsing

septatrix: But the `finalize_image_version` unconditionally writes the version string into it even if it is executable or am I missing something? I.e. `mkosi bump` with an executable `mkosi.version` just breaks now?

septatrix: Ah I think I get it now, please forget it

PR #3697, opened 2025-04-16
mkosi-obs: simplify generation of signed UEFI auth files

ddstreet:
We can use systemd-keyutil to generate the PKCS#7 signature from the OBS-provided PKCS#1 signature and OBS-provided x509 certificate, instead of the 'kernel-sign-file' script.

Also we can use the ESL files directly, instead of the perl statement that removes the auth header, as mkosi.postoutput leaves them for us to use.


bluca (approved):

DaanDeMeyer:
@bluca This is ridiculous you reject my code to use systemd stuff to do this instead of obs stuff but now it's suddenly fine???

bluca:
This was a hack using a kmod script in a way that wasn't intended, it has nothing to do with pesign

PR #3699, opened 2025-04-17
dnf: Stop messing around with plugins

DaanDeMeyer:
We disabled plugins by default a long time ago because the versionlock plugin's configuration from the host was interfering with image builds. Now that we have completely ownership over /etc, there shouldn't be a need to disable plugins anymore if we assume they won't do anything nasty by default without configuration.


teknoraver:
LGTM

PR #3702, opened 2025-04-18
mkosi-initrd: handle KMPs and symlinks under weak-updates

aafeijoo-suse:


mkosi/initrd.py
```diff
@@ -289,6 +289,12 @@ def main() -> None:
             show_docs("mkosi-initrd", DocFormat.all(), resources=r)
         return
 
+    modulesd = Path("/usr/lib/modules") / args.kernel_version
+    if not modulesd.is_dir() or not (
+        (modulesd / "modules.dep").exists() or (modulesd / "modules.dep.bin").exists()
+    ):
```
behrmann, line 295: Maybe factor this into a separate function `is_valid_modulesd` (maybe in `kmod`?), since it's used in three places in two files.

behrmann (commented):
Looks reasonable to me. 

Is weak updates a suse thing and is my understanding correct that it's only about reuse of compatible kernels via symlinks?

aafeijoo-suse:
> Is weak updates a suse thing

KMPs yes, but weak modules no: https://src.fedoraproject.org/rpms/kmod/blob/rawhide/f/weak-modules

>  and is my understanding correct that it's only about reuse of compatible kernels via symlinks?

Compatible kernel modules, yes.

mkosi/initrd.py
aafeijoo-suse: Right, changed.

behrmann (approved):

mkosi/kmod.py
```diff
@@ -16,6 +16,13 @@
 from mkosi.util import chdir, parents_below
 
 
+def is_valid_modulesd(modulesd: Path) -> bool:
+    # Check whether a provided kernel modules directory is valid
+    return modulesd.is_dir() and (
+        (modulesd / "modules.dep").exists() or (modulesd / "modules.dep.bin").exists()
+    )
```
DaanDeMeyer, line 23: Is this correct for use in run_demod()? Isn't it depmod that's going to write these files in the first place, so it's a chicken and egg problem to use this in run_depmod()?

DaanDeMeyer (changes requested):
I'm not sure about the last commit, isn't that a bug in the suse kernel packaging?

mkosi/kmod.py
aafeijoo-suse: Hmm, right... for depmod we're already on the sandbox, so it doesn't make sense... Should I keep this `is_valid_modulesd` function only for `initrd.py`, or write what I had before?

aafeijoo-suse:
> I'm not sure about the last commit, isn't that a bug in the suse kernel packaging?

No, the `weak-updates` directory with symlinks to previous kernels is also used by [Debian](https://manpages.debian.org/unstable/dkms/dkms.8.en.html#NO_WEAK_MODULES=), [RHEL](https://access.redhat.com/articles/9749), [Fedora](https://src.fedoraproject.org/rpms/kmod/blob/rawhide/f/weak-modules#_4), [Oracle Linux](https://docs.oracle.com/en/operating-systems/oracle-linux/8/boot/boot-AboutKernelModules.html#ol-weakmodules)...

mkosi/kmod.py
DaanDeMeyer, line 23: Can you move this into initrd.py as well?

aafeijoo-suse: Sure, done.

mkosi/kmod.py
```diff
@@ -419,6 +419,9 @@ def process_kernel_modules(
 
             p = context.root / m
             if p.is_file() or p.is_symlink():
+                if p.is_symlink():
+                    p_target = context.root / p.resolve().relative_to("/")
```
DaanDeMeyer, line 423: Let's just use `chase()` here, definitely going to be more robust than this

mkosi/initrd.py
```diff
@@ -300,15 +311,22 @@ def main() -> None:
...
             "--remove-files=/usr/lib/firmware/*-ucode",
             "--build-sources=",
             "--include=mkosi-initrd",
         ]  # fmt: skip
 
         if not args.generic:
-            cmdline += ["--kernel-modules-include=host"]
+            cmdline += ["--kernel-modules=host"]
+
+        for path, _, files in os.walk(modulesd / "weak-updates"):
```
DaanDeMeyer, line 324: Do all distributions use the same directory name?

DaanDeMeyer: Please use `Path.rglob()` instead of `os.walk()`

DaanDeMeyer (changes requested):

mkosi/kmod.py
aafeijoo-suse: After removing this check from `run_depmod()`, it fails for the kernel directory that contains only weak modules:

```
‣  Applying kernel module filters
‣   Running modinfo to fetch kernel module dependencies
modinfo: ERROR: Module nvidia-modeset not found.
modinfo: ERROR: Module nvidia not found.
‣ "modinfo --set-version 6.14.0-1-default --null nvidia-modeset nvidia" returned non-zero exit code 1.
localhost:/ # ls -l /usr/lib/modules/6.14.0-1-default/
total 0
drwxr-xr-x 1 root root 80 Apr 23 13:00 updates
localhost:/ # ls -l /usr/lib/modules/6.14.0-1-default/updates/nvidia-open-driver-G06-signed-570.133.07/
total 7360
-rw-r--r-- 1 root root 1137468 Apr 23 13:00 nvidia-modeset.ko.zst
-rw-r--r-- 1 root root 6395074 Apr 23 13:00 nvidia.ko.zst
```

DaanDeMeyer: strace modinfo to see where it fails to look? I'm not sure whether this is supposed to work or not.

aafeijoo-suse:
```
openat(AT_FDCWD, "/sys/module/compression", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib/modules/6.14.0-1-default/modules.softdep", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib/modules/6.14.0-1-default/modules.weakdep", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/proc/cmdline", O_RDONLY|O_CLOEXEC) = 3
read(3, "BOOT_IMAGE=/boot/vmlinuz-6.14.1-"..., 4095) = 174
read(3, "", 3921)                       = 0
close(3)                                = 0
openat(AT_FDCWD, "/usr/lib/modules/6.14.0-1-default/modules.dep.bin", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib/modules/6.14.0-1-default/modules.alias.bin", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib/modules/6.14.0-1-default/modules.builtin.bin", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib/modules/6.14.0-1-default/modules.builtin.alias.bin", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
write(2, "modinfo: ERROR: Module nvidia-mo"..., 49modinfo: ERROR: Module nvidia-modeset not found.
) = 49
openat(AT_FDCWD, "/usr/lib/modules/6.14.0-1-default/modules.dep.bin", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib/modules/6.14.0-1-default/modules.alias.bin", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib/modules/6.14.0-1-default/modules.builtin.bin", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib/modules/6.14.0-1-default/modules.builtin.alias.bin", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
write(2, "modinfo: ERROR: Module nvidia no"..., 41modinfo: ERROR: Module nvidia not found.
) = 41
exit_group(1)                           = ?
+++ exited with 1 +++
```

DaanDeMeyer: So should we skip directories that only contain an `updates` directory?

aafeijoo-suse: I'm not sure what's the best check... we had the `modules.dep{,.bin}` check [in dracut since forever](https://github.com/dracutdevs/dracut/commit/b29cb51654db13706c7356aa827e554456b25b0a), but it does not call `modinfo` directly, it has its custom binary using `libkmod`. On the other hand, running `depmod` internally in mkosi on such kernel directories with only `updates` is pointless. So, I'd say that the check in `initrd.py` is ok as it is now, but `modulesd.is_dir()` in `run_depmod` is not enough.

DaanDeMeyer: So let's skip directories in run_depmod() that only have "updates" in them then.

aafeijoo-suse: Done, not sure if the syntax for the check is the best in python tough.

behrmann: If that refers to the `len(list(glob))` thing, then a possible alternative would be `sum(1 for f in glob)`, which is slightly longer about as fast and maybe marginally less ugly, should use less memory, though, especially if the glob is an `rglob`. On that note, should it be glob or rglob?

aafeijoo-suse: > If that refers to the `len(list(glob))` thing

Yes.

> , then a possible alternative would be `sum(1 for f in glob)`, which is slightly longer about as fast and maybe marginally less ugly

Yes, it's another option. Definitely not so ugly, and we only have to sum ~30 elements. I can write that if you want.

> , should use less memory, though, especially if the glob is an `rglob`. On that note, should it be glob or rglob?

glob, we only need the 1st level, not the whole kernel tree.

behrmann: I don't think it's sufficiently better, so let's keep it like that. If it stays ugly enough, maybe somebody else will find a nicer solution.

PR #3704, opened 2025-04-20
Use a default tools tree by default if mkosi.tools.conf exists

DaanDeMeyer:


mkosi/resources/man/mkosi.1.md
```diff
@@ -1342,6 +1342,9 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     using the settings below or with `mkosi.tools.conf` which can either be a
     file or directory containing extra configuration for the default tools tree.
 
+    If `mkosi.tools.conf` exists in the directory mkosi is invoked in,
+    a default tools tree is used by default.
```
behrmann, line 1346: A lot of default, how about
```suggestion
    a default tools tree is used if no tools tree is specified.
```

mkosi/__init__.py
```diff
@@ -4504,7 +4504,9 @@ def generate_key_cert_pair(args: Args) -> None:
     keylength = 2048
     expiration_date = datetime.date.today() + datetime.timedelta(int(args.genkey_valid_days))
 
-    configdir = finalize_configdir(args)
+    configdir = finalize_configdir(args.directory)
+    if not configdir:
+        die("genkey cannot be used with --directory \"\"")
```
behrmann, line 4509: Ruff complains about this and suggest
```suggestion
        die('genkey cannot be used with --directory ""')
```
I'd suggest
```suggestion
        die("genkey cannot be used with --directory ''")
```
since I'd type that (no need to press shift and for literals I don't want interpolation anyway) or
```suggestion
        die("genkey cannot be used with empty --directory)
```

behrmann (approved):

PR #3705, opened 2025-04-21
mkosi: Override misconfigured gitconfig HTTP/HTTPS proxy with ProxyUrl

ryantimwilson:
When running mkosi on a server with an outgoing HTTP/HTTPS proxy, Build.ProxyUrl must be set. However, some repositories (e.g. systemd) call git in their mkosi scripts (e.g. cloning package repositories for distros) and .gitconfig can also contain different http.proxy and https.proxy settings. Thus, if .gitconfig is misconfigured with the wrong proxy, the user will get confusing errors related to git fetching repositories.

To ensure we use a consistent proxy across git and mkosi, we override git http proxy configuration via GIT_CONFIG_COUNT, GIT_CONFIG_KEY_{n}, GIT_CONFIG_VALUE{n} environment variables. Most of the code complexity is dealing with the case when these variables are set by the user via Environment / EnvironmentFiles.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -2131,6 +2131,7 @@ class Config:
             for line in f.read_text().strip().splitlines()
         )
         env |= self.environment
+        env |= finalize_git_config(self.proxy_url, env)
 
         return env
 
@@ -5242,6 +5243,27 @@ def finalize_term() -> str:
         term = "vt220" if sys.stderr.isatty() else "dumb"
 
     return term if sys.stderr.isatty() else "dumb"
+
+
+def finalize_git_config(proxy_url: Optional[str], env: dict[str, str]) -> dict[str, str]:
+    if proxy_url is None:
+        return {}
+
+    try:
+        cnt = int(env.get("GIT_CONFIG_COUNT", "0"))
+    except ValueError:
+        raise ValueError("GIT_CONFIG_COUNT environment variable must be set to a valid integer")
+
+    # Override HTTP/HTTPS proxy in case its set in .gitconfig to a different value than proxy_url.
+    # No need to override http.proxy / https.proxy if set in a previous GIT_CONFIG_* variable since
+    # the last setting always wins.
+    return {
+        "GIT_CONFIG_COUNT": str(cnt + 2),
+        f"GIT_CONFIG_KEY_{cnt}": "http.proxy",
+        f"GIT_CONFIG_VALUE_{cnt}": proxy_url,
+        f"GIT_CONFIG_KEY_{cnt + 1}": "https.proxy",
+        f"GIT_CONFIG_VALUE_{cnt + 1}": proxy_url,
+    }
 
 
 def yes_no(b: bool) -> str:
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1321,6 +1321,58 @@ def test_environment(tmp_path: Path) -> None:
         assert "TestValue2" not in sub.environment
 
 
+def test_proxy(tmp_path: Path) -> None:
+    d = tmp_path
+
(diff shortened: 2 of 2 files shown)


mkosi/config.py
```diff
@@ -5244,6 +5245,35 @@ def finalize_term() -> str:
...
 
 
+def finalize_git_config(proxy_url: Optional[str], env: dict[str, str]) -> dict[str, str]:
+    if proxy_url is None:
+        return {}
+
+    cnt_str = env.get("GIT_CONFIG_COUNT", "0")
+    if not cnt_str.isdigit():
+        raise ValueError("GIT_CONFIG_COUNT environment variable must be set to a valid integer")
+    cnt = int(cnt_str)
```
behrmann, line 5255: Nit: fewer named variables
```suggestion
    try:
        cnt = int(env.get("GIT_CONFIG_COUNT", 0))
    except ValueError:
        raise ValueError("GIT_CONFIG_COUNT environment variable must be set to a valid integer")
```

mkosi/config.py
```diff
@@ -5244,6 +5245,35 @@ def finalize_term() -> str:
...
+
+    env_override = {}
+    for i in range(cnt):
+        key = env.get(f"GIT_CONFIG_KEY_{i}")
+        if key is None:
+            raise ValueError(
+                f"GIT_CONFIG_KEY_{i} environment variable must be set with GIT_CONFIG_COUNT={cnt}"
+            )
```
behrmann, line 5263: What does git do here? i.e. is git fine if `GIT_CONFIG_COUNT` is nonzero but these variables are not set? If so, we shouldn't mess with these semantics. Maybe we shouldn't check for this, since we don't mess with these variables outside of this?

mkosi/config.py
```diff
@@ -5244,6 +5245,35 @@ def finalize_term() -> str:
...
+            env_override[f"GIT_CONFIG_VALUE_{i}"] = proxy_url
+
+    # Override HTTP/HTTPS proxy in case its set in .gitconfig to a different value than proxy_url.
+    return env_override | {
+        "GIT_CONFIG_COUNT": str(cnt + 2),
+        f"GIT_CONFIG_KEY_{cnt}": "http.proxy",
+        f"GIT_CONFIG_VALUE_{cnt}": proxy_url,
+        f"GIT_CONFIG_KEY_{cnt + 1}": "https.proxy",
+        f"GIT_CONFIG_VALUE_{cnt + 1}": proxy_url,
+    }
```
behrmann, line 5274: Is git first setting wins or last setting wins? If it's the latter, wouldn't just setting these last be enough?

ryantimwilson: Appears to be last setting wins from my local testing and docs: https://git-scm.com/docs/git-config#Documentation/git-config.txt-get

So my previous nonsense from before is indeed unnecessary!

mkosi/config.py
ryantimwilson: git will emit an error per docs: https://git-scm.com/docs/git-config#Documentation/git-config.txt-GITCONFIGCOUNT

>  Any missing key or value is treated as an error. 

Given below where this whole block is unneccessary, yes lets remove this check

mkosi/config.py
ryantimwilson: Fixed in newest version

mkosi/config.py
ryantimwilson: Fixed in newest version

mkosi/config.py
ryantimwilson: Fixed in newest version

DaanDeMeyer (approved):

behrmann (approved):

PR #3706, opened 2025-04-24
Don't use default value if optional settings are set to none

DaanDeMeyer:
Just like it's possible to disable usage of default values for collection based settings, we should also support it for optional settings. This can be used to disable usage of the tools tree even if mkosi.tools.conf exists by setting ToolsTree= in mkosi.local.conf.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -1792,8 +1792,8 @@ class Args:
 
     @classmethod
     @functools.lru_cache(maxsize=1)
-    def fields(cls) -> set[str]:
-        return {f.name for f in dataclasses.fields(cls)}
+    def fields(cls) -> dict[str, dataclasses.Field[Any]]:
+        return {f.name: f for f in dataclasses.fields(cls)}
 
     @classmethod
     def from_namespace(cls, ns: dict[str, Any]) -> "Args":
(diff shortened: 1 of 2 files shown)


mkosi/config.py
```diff
@@ -4684,8 +4684,18 @@ def finalize_value(self, setting: ConfigSetting[T]) -> Optional[T]:
...
+        fieldname = (
+            setting.dest.removeprefix("tools_tree_") if setting.scope == SettingScope.tools else setting.dest
+        )
+        field = Config.fields().get(fieldname)
+        origin = typing.get_origin(field.type) if field else None
+        args = typing.get_args(field.type) if field else []
+        if (
+            (setting.dest in self.cli or setting.dest in self.config)
+            and field
+            and (origin in (dict, list, str) or (origin is typing.Union and type(None) in args))
```
behrmann, line 4698: Thinking about this again, to only cover `Optional[foo]` this should probably also check `len(args) == 2`, because otherwise something like `None | str | int`, would fall under that, but I think that would be fine in that case as well, so maybe generally.

behrmann (approved):
Great addition!

bluca (approved):

PR #3707, opened 2025-04-24
repart: use --append-fstab=auto if available

aplanas:
When generating images there is a chance that a pre-existent /etc/fstab
is present in the mkosi buildroot.  systemd-repart can merge the already
existing fstab with the one generated, separating the sections in a way
that the systemd-repart can be reused several times.


behrmann:
No that the accompanying systemd PR is merged, d you want to undraft this?

aplanas:
@behrmann thanks for the ping! I updated the code and now is ready for review

DaanDeMeyer (approved):

PR #3717, opened 2025-05-15
Move ugly comparison to util

behrmann:
/cc @aafeijoo-suse

### DIFF
--- a/mkosi/kmod.py
+++ b/mkosi/kmod.py
@@ -441,3 +441,14 @@ def process_kernel_modules(
                     p.parent.rmdir()
             elif p.exists():
                 p.rmdir()
+
+
+def is_valid_kdir(kdir: Path) -> bool:
+    dircontent = list(kdir.glob("*"))
+
+    # kdir does not exist or is empty
+    if not dircontent:
+        return False
+
+    # check that kdir contains more than just updates
+    return dircontent != [kdir / "updates"]
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -87,7 +87,7 @@ from mkosi.context import Context
 from mkosi.distributions import Distribution, detect_distribution
 from mkosi.documentation import show_docs
 from mkosi.installer import clean_package_manager_metadata
-from mkosi.kmod import gen_required_kernel_modules, loaded_modules, process_kernel_modules
+from mkosi.kmod import gen_required_kernel_modules, is_valid_kdir, loaded_modules, process_kernel_modules
 from mkosi.log import ARG_DEBUG, complete_step, die, log_notice, log_step
 from mkosi.manifest import Manifest
 from mkosi.mounts import (
@@ -2983,9 +2983,7 @@ def run_depmod(context: Context, *, cache: bool = False) -> None:
 
     if not cache:
         for modulesd in (context.root / "usr/lib/modules").glob("*"):
-            if not modulesd.is_dir() or (
-                (modulesd / "updates").exists() and len(list(modulesd.glob("*"))) == 1
-            ):
+            if not is_valid_kdir(modulesd):
                 continue
 
             process_kernel_modules(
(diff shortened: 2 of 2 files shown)


mkosi/util.py
```diff
@@ -269,3 +269,7 @@ def mandatory_variable(name: str) -> str:
         return os.environ[name]
     except KeyError:
         die(f"${name} must be set in the environment")
+
+
+def dir_only_contains(dir: Path, tgt: PathString) -> bool:
+    return not dir.is_dir() or (dir / tgt).exists() and len(list(dir.glob("*"))) == 1
```
behrmann, line 275: Thinking about this some more, would it make sense for this the directory to also be skipped if it's empty?

septatrix, line 275: Instead of the list allocation and eager glob, the following should also do the trick and be more efficient:

```python
try:
    [child] = dir.iterdir()
except ValueError:
    return False # unwrapping failed, either 0 or 2+ children

Now that it is extracted into a function there is no need to keep it a oneliner

Comment (behrmann) [reply mkosi/util.py]: This is indeed prettier. The whole function becomes a bit unwieldy, but at least it's twice as fast!

Comment (aafeijoo-suse) [reply mkosi/util.py]: > Thinking about this some more, would it make sense for this the directory to also be skipped if it's empty?

It would, but... is realistic to think of a system with a `/usr/lib/modules/<kver>` directory with nothing within it? As a curiosity, some yeas ago I had to add a check with `/usr/lib/modules/<kver>/kernel` in dracut to fix a "real" customer case...

https://github.com/openSUSE/dracut/commit/24aa3f7488b0d668902c0d2bc2566e4fdc289ff4

File: mkosi/util.py [hunk 2]
```diff
@@ -269,3 +269,16 @@ def mandatory_variable(name: str) -> str:
         return os.environ[name]
     except KeyError:
         die(f"${name} must be set in the environment")
+
+
+def dir_only_contains(dir: Path, tgt: PathString) -> bool:
```
aafeijoo-suse, line 274: Shouldn't this be called something like `is_valid_kdir` or so?

`not dir.is_dir():` returing `True` does not mean that `dir` only contains `tgt`...

mkosi/util.py
behrmann: > It would, but... is realistic to think of a system with a /usr/lib/modules/<kver> directory with nothing within it? 

That's why I wondered. I mean the Debian system I just checked, (that's long-running and has been updating for over a year now, so quite different from the images we build, has about 20 directories, more than three quarters o which are empty.

mkosi/kmod.py
```diff
@@ -441,3 +441,16 @@ def process_kernel_modules(
                     p.parent.rmdir()
             elif p.exists():
                 p.rmdir()
+
+
+def is_valid_kdir(kdir: Path) -> bool:
+    if not kdir.is_dir():
+        return True
```
DaanDeMeyer, line 448: Shouldn't this return False? The function is called is_valid_kdir()

keszybz (commented):
Oh, I see I didn't press "submit". Sorry for any outdated comments.

mkosi/util.py
```diff
@@ -272,4 +272,13 @@ def mandatory_variable(name: str) -> str:
...
 def dir_only_contains(dir: Path, tgt: PathString) -> bool:
-    return not dir.is_dir() or (dir / tgt).exists() and len(list(dir.glob("*"))) == 1
+    if not dir.is_dir():
+        return True
+
+    # Ensure the directory only contains a single entry
+    try:
+        [child] = dir.iterdir()
+    except ValueError:
+        return False
```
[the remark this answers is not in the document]
keszybz: Please ignore the comments about the name. But I think the proposed code is better.

behrmann:
Reworked as suggested, though it's a bit unfortunate that the try-except-less code is slower by more than a factor of 2. I added skipping empty directories as well.

septatrix:
> Reworked as suggested, though it's a bit unfortunate that the try-except-less code is slower by more than a factor of 2. I added skipping empty directories as well.

It could also be unrolled into two `next(...)` calls to keep it lazy but avoid the try-except (even though the unpacking pattern is a quite common and idiomatic Python...)

PR #3721, opened 2025-05-20
Fix version autobumping

p1gp1g:
Auto bump was never used because image_version is not set when autobumping.

I'm not 100% certain of the patch, but the `and context.cli.get("image_version") is not None` must not be applied to auto_bump

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -5142,8 +5142,10 @@ def parse_config(
     configdir = finalize_configdir(args.directory)
 
     if (
-        ((args.auto_bump and args.verb.needs_build()) or args.verb == Verb.bump)
-        and context.cli.get("image_version") is not None
+        (
+          (args.auto_bump and args.verb.needs_build())
+          or (args.verb == Verb.bump and (args.auto_bump or context.cli.get("image_version") is not None))
+        )
         and configdir is not None
     ):
         context.cli["image_version"] = bump_image_version(configdir)


mkosi/config.py
```diff
@@ -5142,8 +5142,10 @@ def parse_config(
     configdir = finalize_configdir(args.directory)
 
     if (
-        ((args.auto_bump and args.verb.needs_build()) or args.verb == Verb.bump)
-        and context.cli.get("image_version") is not None
+        (
+          (args.auto_bump and args.verb.needs_build())
+          or (args.verb == Verb.bump and (args.auto_bump or context.cli.get("image_version") is not None))
+        )
```
behrmann, line 5148: Having `auto_bump` in two places here is a bit confusing, the logic can be rearranged into
```
        (
          (args.auto_bump and args.verb.needs_build()) or
          (args.verb == Verb.bump and args.auto_bump) or 
          (args.verb == Verb.bump and context.cli.get("image_version") is not None))
        )
```
and therefore
```suggestion
        (
          (args.auto_bump and (args.verb.needs_build() or args.verb == Verb.bump))
          or (args.verb == Verb.bump and context.cli.get("image_version") is not None)
        )
```
but since `context.cli["image_version"]` gets clobbered with whatever `bump_image_version(configdir)` returns, we probably need to drop that instead.

p1gp1g: I've been hesitating between the 2 forms, in one case `auto_bump` is in two places, in the other case it is `verb == Verb.bump`

Issue #3727, opened 2025-05-21
Allow providing extra configuration for reprepro via sandbox trees


whiteley:
### mkosi commit the issue has been seen with

4d031bc57d7ab662721d6be72be2b20748d4c223

### Used host distribution

Ubuntu noble

### Used target distribution

Ubuntu noble

### Linux kernel version used

6.1.89

### CPU architectures issue was seen on

x86_64

### Unexpected behaviour you saw

When using a local package that previously worked I see the following output

```
‣  Rebuilding local package repository
No section given for 'local-pkg', skipping.
```
(then after altering the package and rebuilding)
```
‣  Rebuilding local package repository
No priority given for 'local-pkg', skipping.
```

These fields are recommended not mandatory for a debian packages according to https://www.debian.org/doc/debian-policy/ch-controlfields.html#debian-binary-package-control-files-debian-control and worked with older versions of `mkosi` and `dpkg-scanpackages`. There are options available to `reprepro` that can override these files but they do potentially have unintended consequences.

I'm happy to make a PR for changing the execution of `reprepro` if desired but I'm curious what the maintainers viewpoint is on this. As a workaround, I tried the below change which worked until we can try to get those fields added to all of our internal packages.

```
diff --git a/mkosi/installer/apt.py b/mkosi/installer/apt.py
index 1d78f22c..b145b1f4 100644
--- a/mkosi/installer/apt.py
+++ b/mkosi/installer/apt.py
@@ -299,6 +299,8 @@ class Apt(PackageManager):
             [
                 "reprepro",
                 "--ignore=extension",
+                "--section=misc",
+                "--priority=optional",
                 "includedeb",
                 "mkosi",
                 *names,
```

### Used mkosi config

```ini
[Distribution]
Distribution=ubuntu
Release=noble

[Output]
Format=portable
ImageId=test

[Content]
Packages=local-pkg
PackageDirectories=mkosi.packages
```

### mkosi output

```sh
‣ Loading configuration file /home/mwhiteley/mkosi-test/mkosi.conf
‣ Loading configuration file /tmp/tmph4oq3q5w/resources/mkosi-initrd/mkosi.conf
‣ Loading configuration file /tmp/tmph4oq3q5w/resources/mkosi-initrd/mkosi.conf.d/10-debian-kali-ubuntu/mkosi.conf
[...]

PR #3730, opened 2025-05-22
apt: add a variable MKOSI__REPREPRO_EXTRA_ARGS to pass args to reprepro

behrmann:
This can be used to pass extra options to reprepro, e.g. to set section or priority when a package does not provide them, which reprepro complains. Clobbering these fields is generally not advisable, because e.g. priority interacts with the essential flag. Even though Debian policy only recommends these fields and dpkg therefore accepts packages without them, they are broken and should be fixed, alas some proprietary packages omit them and would need repackaging.

To make this easier on users, give them a knob they can use to try to work around this.

Fixes: #3727

@whiteley can you give this a try?

### DIFF
--- a/mkosi/installer/apt.py
+++ b/mkosi/installer/apt.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
 import dataclasses
+import shlex
 import textwrap
 from collections.abc import Sequence
 from pathlib import Path
@@ -299,6 +300,7 @@ class Apt(PackageManager):
             [
                 "reprepro",
                 "--ignore=extension",
+                *shlex.split(context.config.environment.get("MKOSI__REPREPRO_EXTRA_ARGS", "")),
                 "includedeb",
                 "mkosi",
                 *names,
--- a/mkosi/resources/man/mkosi.1.md
+++ b/mkosi/resources/man/mkosi.1.md
@@ -2999,6 +2999,14 @@ repository.
   value of the respective `extension-release` file when building a sysext
   or confext. By default the value is set to `initrd system portable`.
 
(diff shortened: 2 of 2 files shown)


DaanDeMeyer:
The tool supports config files (https://www.mankier.com/1/reprepro#Config_Files). Let's do this via that instead of environment variables so stuff can be plugged in via sandbox trees

whiteley:
> The tool supports config files (https://www.mankier.com/1/reprepro#Config_Files). Let's do this via that instead of environment variables so stuff can be plugged in via sandbox trees

Looks like adding a line

```
DebOverride: /path/to/mkosi.override
```

to the `conf/distributions` file generated in https://github.com/systemd/mkosi/blob/a0b4e1af9a35c0ce7e105c3c11b541f4a70b7b35/mkosi/installer/apt.py#L285 and then having that `mkosi.override` populated by the user with content like

```
pkg-name Priority required
pkg-name Section web
```

and the ability to provide it via sandbox trees does indeed solve the issue for us.

behrmann:
> DebOverride: /path/to/mkosi.override

I haven't gotten around to looking at the man page yet, so I'm not quite sure I see what this would look like, i.e. should we always stage an empty `mkosi.override` somewhere and copy one in from a well-known path, e.g. `/etc/mkosi/reprepro.override` in the tools tree?

DaanDeMeyer:
> > DebOverride: /path/to/mkosi.override
> 
> I haven't gotten around to looking at the man page yet, so I'm not quite sure I see what this would look like, i.e. should we always stage an empty `mkosi.override` somewhere and copy one in from a well-known path, e.g. `/etc/mkosi/reprepro.override` in the tools tree?

I'd just go for `/usr/lib/reprepro` and `/etc/reprepro` or something like that.

Issue #3687, opened 2025-04-13
Unable to copy files into sandbox


eadwu:
### mkosi commit the issue has been seen with

a7e90514fa298a03b5880bb4c9d79dc5e7d55eca

### Used host distribution

Fedora 41

### Used target distribution

Fedora 41/RawHide

### Linux kernel version used

6.13.9-200.fc41.x86_64

### CPU architectures issue was seen on

x86_64

### Unexpected behaviour you saw

Trying to inject third-party / external repository (ZFS) . All my commands are prefixed with `sudo` due to `cp` and `selinux` thing.

Running 
```
❯ sudo ../mkosi/bin/mkosi -f sandbox -- wc -l /etc/pki/rpm-gpg/RPM-GPG-KEY-openzfs-fedora-41
wc: /etc/pki/rpm-gpg//etc/pki/rpm-gpg/RPM-GPG-KEY-openzfs-fedora-41: No such file or directory
```

If `[Build]` section is removed, it is identical to host
```
❯  sudo rm -rf mkosi.tools/ && sudo ../mkosi/bin/mkosi -f sandbox -- wc -l /etc/pki/rpm-gpg/RPM-GPG-KEY-openzfs-fedora-41
57 /etc/pki/rpm-gpg/RPM-GPG-KEY-openzfs-fedora-41
```

If I directly copy the file into `mkosi.tools/` after it is generated, it works fine.
```
❯  sudo cp /etc/pki/rpm-gpg/RPM-GPG-KEY-openzfs-fedora-41 ./mkosi.tools/etc/pki/rpm-gpg
sudo ../mkosi/bin/mkosi -f sandbox -- wc -l /etc/pki/rpm-gpg/RPM-GPG-KEY-openzfs-fedora-41
57 /etc/pki/rpm-gpg/RPM-GPG-KEY-openzfs-fedora-41
```

Also why is it that the files in the toolchain can be used in the sandbox? The `zfs.repo` file defining the repository works with `SandboxTrees` but including the repository key into the `ToolsTree` makes it available during the image build process.

### Used mkosi config

```ini
[Distribution]
Distribution=fedora
Release=41

[Build]
ToolsTree=default
ToolsTreeDistribution=fedora
# ToolsTreeRelease=rawhide

SandboxTrees=RPM-GPG-KEY-openzfs-2022:/etc/pki/rpm-gpg/RPM-GPG-KEY-openzfs-2022
[...]

PR #3736, opened 2025-05-26
Fix for #3687

DaanDeMeyer:


mkosi/installer/rpm.py
```diff
@@ -52,16 +54,23 @@ def find_rpm_gpgkey(
...
+        f"/etc/pki/rpm-gpg/**/{key}",
+    ]
 
-    # ToolsTreeCertificates= only applies to certificates but the rpm gpg keys in /etc are located within the
-    # /etc/pki certificates directory so as a result the option has to apply to the rpm gpg keys in /etc as
-    # well
-    root = context.config.tools() if context.config.tools_tree_certificates else Path("/")
+    paths = (
+        run(
+            ["bash", "-c", rf"shopt -s nullglob && printf '%s\n' {' '.join(globs)}"],
```
behrmann, line 64: That construction did surprise me a bit a first, relying on the glob to be expanded and the expanded thing passed to `printf`. I think some `find /usr/share/distribution-gpg-keys /etc/pki -name <key> -type f` would be easier and we have it in the tools tree.

DaanDeMeyer: We have it in the tools tree but it's part of the misc profile which is not required to put in the dependency list of mkosi distro packages, so I'd prefer to use bash here to avoid that dependency.

PR #3741, opened 2025-05-26
sandbox: Support using mkosi-sandbox as a library

DaanDeMeyer:
Let's allow using mkosi-sandbox to create an in process sandbox by not taking a command line to execute if mkosi-sandbox was itself imported and not executed.

If we were imported, also don't do any print()'s to behave like a proper library.

### DIFF
--- a/mkosi/resources/man/mkosi-sandbox.1.md
+++ b/mkosi/resources/man/mkosi-sandbox.1.md
@@ -21,6 +21,20 @@ Note that this sandbox is not designed to be a security boundary. Its intended p
 is to allow running commands in an isolated environment so they are not affected by the
 host system.
 
+It is possible to use `mkosi-sandbox` to create an in process sandbox for python
+applications by importing it as a module and invoking its main function with only
+options. No command line to execute is needed in this case. As an example:
+
+```python
+import mkosi.sandbox
+
+mkosi.sandbox.main(["--become-root"])
+print(os.getuid())
+```
+
+Only the `main` function can be invoked. Invoking any other functions
+from `mkosi.sandbox` is not supported and may break in future versions.
+
 # OPTIONS
 
 `--tmpfs DST`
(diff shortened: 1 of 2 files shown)


mkosi/sandbox.py
```diff
@@ -129,7 +129,7 @@ class Style:
 
 
 def oserror(syscall: str, filename: str = "") -> None:
-    if ctypes.get_errno() == ENOSYS:
+    if ctypes.get_errno() == ENOSYS and __name__ == "__main__":
```
behrmann, line 132: Nit: Since this is used so often a simple function
```python
def is_main() -> bool:
    return __name__ == "__main__"
```
or a global variable `IS_MAIN = __name__ == "__main__"`, would make this a bit nicer.

mkosi/sandbox.py
```diff
@@ -871,11 +871,11 @@ def execute(self, oldroot: str, newroot: str) -> None:
 """
 
 
-def main() -> None:
+def main(argv: list[str] = []) -> None:
```
behrmann, line 874: Shouldn't it be fine to move `sys.argv` up here?
```suggestion
def main(argv: list[str] = sys.argv[1:]) -> None:
```

mkosi/resources/man/mkosi-sandbox.1.md
```diff
@@ -21,6 +21,20 @@ Note that this sandbox is not designed to be a security boundary. Its intended p
 is to allow running commands in an isolated environment so they are not affected by the
 host system.
 
+It is possible to use `mkosi-sandbox` to create an in process sandbox for python
+applications by importing it as a module and invoking its main function with only
+options. No command line to execute is needed in this case. As an example:
+
+```py
```
behrmann, line 28: Nit: For consistency with the only other Python code blacks we have
```suggestion
```python
```

behrmann (approved):

PR #3743, opened 2025-05-27
Some manual rework

behrmann:


mkosi/resources/man/mkosi.1.md
```diff
@@ -3232,6 +3232,11 @@ include <tunables/global>
 
   You can set `PORTABLE_PREFIXES=` in the `os-release` file yourself in a postinst script.
 
+- How can I make a tools tree that is not a directory?
+
+  The output format for tools trees cannot currently be changed via configuration files. Please use the `-t`
+  option on the commandline.
```
DaanDeMeyer, line 3238: If we add this we should also document how to build a non default tools tree in the first place

DaanDeMeyer (changes requested):

mkosi/resources/man/mkosi.1.md
behrmann: I reworked this to add a tools tree section.

mkosi/resources/man/mkosi.1.md
```diff
@@ -2858,6 +2799,84 @@ script. When all three are enabled together turn-around times for
...
+builds more reproducible, but also allows to use newer tooling, that is not yet available in the host
+distribution running mkosi.
+
+Tools trees can be provided via the `ToolsTree=` option, the `mkosi.tools` directory or built automatically
+by mkosi if set to `ToolsTree=default`. For most use cases setting it is sufficient to use the default tools
+trees and the use of a tools tree is recommended.
+
+Tools trees can be further customized via the different `ToolsTree*=` variables as well as the
+`mkosi.tools.conf` configuration file or directory. The output format for tools trees cannot currently be
+changed via configuration files. Please use the `-t` option on the commandline.
```
DaanDeMeyer, line 2814: This still comes out of nowhere. We have to mention first that it's possible at all to run something like `mkosi --include=mkosi-tools`

mkosi/resources/man/mkosi.1.md
```diff
@@ -2858,6 +2799,84 @@ script. When all three are enabled together turn-around times for
...
+
+Fully custom tools trees can be built like any other mkosi image—using the `mkosi-tools` include, see
+`Include=`, can be used to include the packages from the table below.
```
DaanDeMeyer, line 2817: This should probably be moved earlier and include the command line example

DaanDeMeyer (changes requested):

mkosi/resources/man/mkosi.1.md
```diff
@@ -2858,7 +2799,89 @@ script. When all three are enabled together turn-around times for
...
+Fully custom tools trees can be built like any other mkosi image, but mkosi provides a builtin include
+providing the default tools tree packages:
+
+```bash
+mkosi --include=mkosi-tools
+```
+
+Tools trees, including default tools trees, can be further customized via the different `ToolsTree*=`
+variables as well as the `mkosi.tools.conf` configuration file or directory. The output format for tools
+trees cannot currently be changed via configuration files. Please use the `-t` option on the command line.
```
DaanDeMeyer, line 2821: I would drop the sentence about the output format being unchangeable and `-t` and change the above example to `mkosi --include=mkosi-tools -t directory` instead

DaanDeMeyer (changes requested):

PR #3748, opened 2025-05-28
Generate initrds outside of umask context manager

septatrix:


septatrix:
Context: `apk` still seems to look for a lot of config under the buildroot so for PostmarketOS/Alpine we need to create some files in `Apk.setup`. Due to the context manager being active when calling `finalize_initrds` they inherit that mask resulting in dirs like `etc` becoming `drw-------`.

We could also manually reset the mask in `Apk.setup` but it seems like an oversight to me that `finalize_initrd` would be called with a mask applied. Happy though to apply that downstream only for apk though if you prefer

PR #3749, opened 2025-05-29
Bind mount /etc from tools tree into relaxed sandbox

DaanDeMeyer:
Config from /etc often references stuff in /usr. Two examples I've encountered are shell config from /etc/profile and dnf5 blowing up when there's plugin configuration in /etc without the corresponding plugin being installed.

To work around such issues, let's use /etc from the tools tree in the relaxed sandbox instead of /etc from the host. This also saves the user from having to create directories in their host's /etc to be able to use mkosi sandbox.


mkosi/__init__.py
```diff
@@ -4126,21 +4126,33 @@ def run_sandbox(args: Args, config: Config) -> None:
...
+            and not (config.tools() / "etc/resolv.conf").exists()
+        ):
+            (config.tools() / "etc/resolv.conf").touch()
+
+        # Since we reuse almost every top level directory from the host except /usr and /etc, the crypto
+        # mountpoints have to exist already in these directories or we'll fail with a permission error. Let's
+        # check this early and show a better error and a suggestion on how users can fix this issue. We use
+        # slice notation to get every 3rd item from the mounts list which is the destination path.
+        for dst in finalize_certificate_mounts(config, relaxed=True)[2::3]:
+            if not os.fspath(dst).startswith("/etc") and not Path(dst).exists():
```
```suggestion
        for dst in finalize_certificate_mounts(config, relaxed=True)[2::3]:
            p = Path(dst)
            if not p.is_relative_to("/etc") and not p.exists():
```

mkosi/__init__.py
```diff
@@ -4948,6 +4952,22 @@ def run_build(
...
 
+def ensure_tools_tree_has_etc_resolv_conf(config: Config) -> None:
+    if not config.tools_tree:
+        return
+
+    # We can't bind mount in the hosts's /etc/resolv.conf if this file doesn't exist without making the
+    # entirety of /etc writable or messing around with overlayfs, so let's just ensure it exists.
+    path = config.tools_tree / "etc/resolv.conf"
+
+    if not path.is_symlink() and not path.exists():
```
behrmann, line 4963: This bends my brain a bit. Logically this is equivalent to `not (path.is_symlink() or path.exists())`. `Path.exists()` by default follows symlinks, so this covers the case when `path` is a dangling symlink, since `Path.is_symlink` is also true for them, right? I think in that case this could just be
```suggestion
    if not path.exists(follow_symlinks=False):
```
to ensure that there's something to mount over.

daandemeyer: So as mentioned in chat, follow_symlinks unfortunately doesn't exist for exists() until 3.12

evverx:
With this PR merged `mkosi boot -- -n` (or `mkosi boot` with `RuntimeNetwork=interface`) fails with
```sh
Failed to generate predictable MAC address for container side: Package not installed
Parent died too early
```
because of `/etc/machine-id` in the systemd repository.

daandemeyer:
> With this PR merged `mkosi boot -- -n` (or `mkosi boot` with `RuntimeNetwork=interface`) fails with
> 
> ```shell
> Failed to generate predictable MAC address for container side: Package not installed
> Parent died too early
> ```
> 
> because of `/etc/machine-id` in the systemd repository.

Fix waiting in https://github.com/systemd/mkosi/pull/3783

PR #3750, opened 2025-05-29
mkosi-tools: Introduce debug and gui profiles to reduce the number of packages installed by default

DaanDeMeyer:
This saves roughly 33% (500M) of installed packages on the Fedora tools tree.


behrmann (approved):
CI failure is spurious, mypy 1.16.0 apparently makes a cast unnecessary.

Issue #3751, opened 2025-05-29
Initird stage fails with `Distribution=ubuntu` when creating raw images


SapphirusBeryl:
### mkosi commit the issue has been seen with

25.3

### Used host distribution

Arch Linux Host -> Docker -> Ubuntu 25.04 container

### Used target distribution

Ubuntu 24.04 LTS

### Linux kernel version used

6.12.31-1-lts

### CPU architectures issue was seen on

x86_64

### Unexpected behaviour you saw

When utilising `mkosi` to create Ubuntu-based virtual machine images, the process results in a failure during the `initrd` stage: `apt` exits with non-zero exit code 100.

Prior to this failure, the following output is returned by `apt`:

```sh
E: Unable to locate package erofs-utils
```

By contrast, a similar configuration with Debian as the set distribution using a bare-minimum set of packages, works without any trouble, whatsoever.

Full debug logs: [debug.txt](https://github.com/user-attachments/files/20511776/debug.txt)

### Used mkosi config

```ini
[Output]
Format=disk
OutputDirectory=build

[Distribution]
Distribution=ubuntu
Release=noble

# Package list obtained from Ubuntu noble cloud image with `$ apt-mark showmanual`
[Content]
Packages=
		base-files
		bash
		bsdutils
		cloud-init
		dash
		diffutils
		eatmydata
		findutils
		grep
		grub-pc
		gzip
		hostname
		init
		libeatmydata1
		libwrap0
		linux-virtual
		login
		ncurses-base
		ncurses-bin
		ncurses-term
		openssh-server
		openssh-sftp-server
		python-babel-localedata
		python3-babel
		python3-jinja2
		python3-json-pointer
		python3-jsonpatch
		python3-jsonschema
		python3-markupsafe
		python3-pyrsistent
		python3-serial
		python3-tz
		shim-signed
		ssh-import-id
		ubuntu-minimal
		ubuntu-server
		ubuntu-standard
		util-linux

[Content]
Bootable=yes
Bootloader=grub
RootPassword=password
```

### mkosi output

```sh
Copying in of '/var/tmp/.#repart0a5339ce8685d34f' on block level completed.
Syncing future partition 1 contents to disk.
Block level copying and synchronization of partition 1 complete in 2.482703s (542.2M/s).
Adding new partition 1 to partition table.
Writing new partition table.
All done.
‣  Running modinfo to fetch kernel module dependencies
‣  Calculating required kernel modules and firmware
[...]

PR #3752, opened 2025-05-30
fix: Remove erofs-utils from initrd

SapphirusBeryl:
Since Ubuntu now includes `erofs-utils` in the universe repository, add it to the default repositories. This fixes a bug causing the initrd stage of Ubuntu images to fail.

Closes https://github.com/systemd/mkosi/issues/3751.

### DIFF
--- a/mkosi/resources/mkosi-initrd/mkosi.conf.d/10-arch.conf
+++ b/mkosi/resources/mkosi-initrd/mkosi.conf.d/10-arch.conf
@@ -8,7 +8,6 @@ Packages=
         btrfs-progs
         e2fsprogs
         xfsprogs
-        erofs-utils
         dosfstools
 
         # Various libraries that are dlopen'ed by systemd
--- a/mkosi/resources/mkosi-initrd/mkosi.conf.d/10-centos/mkosi.conf.d/20-epel-packages.conf
+++ b/mkosi/resources/mkosi-initrd/mkosi.conf.d/10-centos/mkosi.conf.d/20-epel-packages.conf
@@ -6,4 +6,3 @@ Repositories=epel
 [Content]
 Packages=
         btrfs-progs
-        erofs-utils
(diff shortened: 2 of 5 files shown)


SapphirusBeryl:
> Maybe we can just modify those config files in `mkosi/resources/`?

Tried that in `resources/mkosi-initrd/mkosi.conf.d` and it didn't work.

DaanDeMeyer:
I don't think we should enable the universe repository by default, just like we don't enable epel by default for centos. If canonical at will moves stiff around, that's something to take up with canonical.

SapphirusBeryl:
> I don't think we should enable the universe repository by default, just like we don't enable epel by default for centos. If canonical at will moves stiff around, that's something to take up with canonical.

https://github.com/systemd/mkosi/blob/main/mkosi.conf.d/20-ubuntu/mkosi.conf#L8 Around 4 months ago, it was added to the tools tree by default.

If the solution otherwise is to remove erofs-utils, do we remove it from every other distribution during the initrd stage, as well? This seems like the more aggressive action than just enabling universe by default.

DaanDeMeyer:
Like I said, there's no point to it being in the initrd anymore, systemd won't generate fsck units for it anymore.

daandemeyer:
> > Like I said, there's no point to it being in the initrd anymore, systemd won't generate fsck units for it anymore.
> 
> There are LTS distributions, such as Ubuntu 24.04, for example, which ship a version of systemd from two years ago. For our use case in this instance, it shouldn't break anything, but I'm not confident there won't be regressions for use cases involving EROFS.
> 
> I'm happy to update this PR to remove these for whichever distributions applicable, though first I just want to be absolutely certain to clarify there won't be any regressions to distributions still shipping older versions of systemd.

Older versions of systemd will simply notice the erofs fsck program isn't installed and won't generate the fsck unit, whereas on newer systemd it won't be generated regardless of whether the erofs fsck program is installed or not. So we're not regressing anything.

behrmann (approved):

Issue #3715, opened 2025-05-07
APT sources file missing if `apt` is not installed on the base image


ragazenta:
### mkosi commit the issue has been seen with

main

### Used host distribution

Debian Testing

### Used target distribution

Debian 12 Bookworm

### CPU architectures issue was seen on

None

### Unexpected behaviour you saw

In my configuration `apt` is not installed on base image, but it will be installed on main image.

The following check, prevent APT sources file being installed to final image. It was introduced in 74eea36f34073114041f5d4d522ca0b7c393ea5e.

```python
if "apt" in itertools.chain(context.config.packages, context.config.volatile_packages):
```

### Used mkosi config

```ini

```

### mkosi output

```sh

```

behrmann:
Let me try to rephrase that, the way I understand it:
- You are building a base image without apt, but that image should carry the `.sources` file
- The `.sources` file does not get installed because of this check
- you build another image using the base image using `BaseTrees=` that installs apt, but skips the logic to generate the `.sources` files?

Is this correct?

ragazenta:
Yes, if I understand correctly, the previous behaviour was to always install `.sources` file unless `${BUILDROOT}/usr/bin/apt` is missing.

ragazenta:
Hmm, the problem introduced in 74eea36f34073114041f5d4d522ca0b7c393ea5e that `install_apt_sources` is called only once, before `apt` even being installed. 

Before 74eea36f34073114041f5d4d522ca0b7c393ea5e,  `install_apt_sources`  is called multiple times (it's called in every `install_packages`).

So I think `install_apt_sources` should be called again before final step.

daandemeyer:
> Hmm, the problem introduced in [74eea36](https://github.com/systemd/mkosi/commit/74eea36f34073114041f5d4d522ca0b7c393ea5e) that `install_apt_sources` is called only once, before `apt` even being installed.
> 
> Before [74eea36](https://github.com/systemd/mkosi/commit/74eea36f34073114041f5d4d522ca0b7c393ea5e), `install_apt_sources` was called multiple times (it's called in every `install_packages`).
> 
> So I think `install_apt_sources` should be called again before final step.

Why is it a problem that we only call it once before `apt` is installed? We now have an additional check to see if apt is going to be installed.

ragazenta:
Here is a simple reproducer: https://github.com/ragazenta/mkosi-bugrep/tree/main/missing-apt-sources

PR #3753, opened 2025-05-30
Introduce distribution installer finalize function

ragazenta:
This adds an ability to do something "after" installing is done. The current use case is for Debian based distribution to install APT sources files if needed.

Closes: #3715


daandemeyer:
Closing in favor of https://github.com/systemd/mkosi/pull/3963

PR #3754, opened 2025-05-30
Add /etc/xattr.conf in the sandbox to workaround cross-FS copy issue

### LINKED ISSUE #3732: mkosi is over-eager to use "cp --preserve=xattr", fails in cross-filesystem cases

main

### Used host distribution

archlinux

### Used target distribution

ubuntu

### Linux kernel version used

6.14.6-arch1-1

x86_64

`WorkspaceDirectory` set to XFS-supported directory (here `/tmp/mkosi.workspace`). All other `mkosi` dirs on BTRFS-supported directory.

```
...
‣ Copying repository metadata
cp: setting attributes for '/work/tmp/mkosi-gdonval-workspace/mkosi/mkosi-metadata-lwbvcpek/cache/apt/srcpkgcache.bin': Operation not supported
...
cp: setting attributes for '/work/tmp/mkosi-gdonval-workspace/mkosi/mkosi-metadata-lwbvcpek/cache/apt': Operation not supported
‣ "cp --recursive --no-dereference --preserve=mode,links,timestamps,ownership,xattr --reflink=auto --copy-contents '/work/home/gdonval/.cache/mkosi/ubuntu~noble~x86-64/cache/apt' /work/tmp/mkosi-gdonval-workspace/mkosi/mkosi-metadata-lwbvcpek/cache/apt --no-target-directory" returned non-zero exit code 1.
```

The reason it happens it because original files have attrs that cannot be copied, e.g:

```
$ lsattr -a /home/gdonval/Documents/Work/RVMI/Services/UbuntuRVMI/mkosi.cache/ubuntu~noble~x86-64~main.metadata.cache/cache/apt/srcpkgcache.bin
--------c------------- /home/gdonval/Documents/Work/RVMI/Services/UbuntuRVMI/mkosi.cache/ubuntu~noble~x86-64~main.metadata.cache/cache/apt/srcpkgcache.bin
$ lsattr -a /tmp/mkosi-gdonval-workspace/mkosi/mkosi-metadata-lwbvcpek/cache/apt/srcpkgcache.bin
---------------------- /tmp/mkosi-gdonval-workspace/mkosi/mkosi-metadata-lwbvcpek/cache/apt/srcpkgcache.bin
```

That `c` is invalid on XFS. 

### Used mkosi config

```ini
[Distribution]
Distribution=ubuntu
Release=noble
Repositories=universe multiverse

[Build]
WorkspaceDirectory=/tmp/mkosi.workspace

[Output]
Format=disk
```

### mkosi output

```sh

```

gdonval:
Fixes #3732.

This problem is only present when copying files across filesystem boundaries. Caveat: /etc/xattr.conf itself might be copied to destination if present at the source. This should be fine until coreutils bug 78623 is closed.

### DIFF
--- a/mkosi/tree.py
+++ b/mkosi/tree.py
@@ -148,7 +148,33 @@ def copy_tree(
         if src.is_dir():
             cmdline += ["--no-target-directory"]
 
-        run(cmdline, sandbox=sandbox(options=options))
+        # A valid /etc/xattr.conf is necessary to work around cross-filesystem copy failures (bug #3732).
+        # This can collapse to the following when cp implements a way to filter out specific xattr
+        # on the command line:
+        #
+        #     run(cmdline, sandbox=sandbox(options=options))
+        #
+        # cmdline, above, will need to be suitably adjusted.
+        with tempfile.TemporaryDirectory() as tmp:
+            import textwrap
+            p = Path(tmp) / "xattr.conf"
+            p.write_text(
+                textwrap.dedent(
+                    """
+                    *                             skip          # Default: ignore all xattr but
+                    system.*                      permissions   #   allow well-known 'system'
+                    trusted.*                     permissions   #   allow well-known 'trusted'
+                    user.*                        permissions   #   allow well-known 'user'
+                    security.*                    permissions   #   allow well-known 'security'
+                                                                # but 
+                    trusted.SGI_*                 skip          #   skip XFS-specific namespace (not portable)
+                    user.Beagle.*                 skip          #   skip Beagle index data (not portable)
+                    security.evm                  skip          #   skip evm (may only be written by kernel)
+                    """
+                )
+            )
+            local_options = options + ["--ro-bind", p, "/etc/xattr.conf"]
+            run(cmdline, sandbox=sandbox(options=local_options))
 
     # Subvolumes always have inode 256 so we can use that to check if a directory is a subvolume.
     if (


mkosi/tree.py
```diff
@@ -148,7 +148,34 @@ def copy() -> None:
         if src.is_dir():
             cmdline += ["--no-target-directory"]
 
-        run(cmdline, sandbox=sandbox(options=options))
+        # A valid /etc/xattr.conf is necessary to work around cross-filesystem copy failures (bug #3732).
```
behrmann: Nit: can you make this a full link to make this standalone? Some distros, e.g. Debian with salsa, have forks on their own forges. A link make this standalone.

mkosi/tree.py
```diff
@@ -148,7 +148,34 @@ def copy() -> None:
...
+        # This can collapse to the following when cp implements a way to filter out specific xattr
+        # on the command line:
+        #
+        #     run(cmdline, sandbox=sandbox(options=options))
+        #
+        # cmdline, above, will need to be suitably adjusted.
+        with tempfile.TemporaryDirectory() as tmp:
+            import textwrap
+
+            p = Path(tmp) / "xattr.conf"
```
behrmann, line 162: There's always more stuff to learn, I've never stumbled over that before. Just out of curiosity, where can one find documentation on this? `man xattr` is empty on this, coreutils only references this in a comment and also in other plumbing layer repos I don't see documentation and also cursory web searches don't yield any results.

mkosi/tree.py
gdonval: I am not sure what you mean. In doubt, I've just added full URLs to the corresponding bug reports though so that future us can read this, go to the bug report and act in consequence.

mkosi/tree.py
gdonval: >  I've never stumbled over that before. 

Me neither. :) The reason is that it is actually part of `libattr` and there is no documentation about it. The original git repo (https://cgit.git.savannah.gnu.org/cgit/attr.git) is slow and most mirrors are out of date but this shows where `xattr.conf` is being defined and parsed:

https://github.com/xmake-mirror/attr/blob/d43f8bcc0c31356146da3491a17273eeaf1e328a/libattr/attr_copy_action.c#L31

It looks like it works by building a linked-list of patterns and actions (`attr_parse_attr_conf`) and then later goes through the list, from the end, and return corresponding skip/permit action at first match (`attr_copy_action`). I might be wrong though: I definitely didn't spend enough time to wrap my head around all that.

gdonval:
BTW, alternative is to use `cp -a`, which ignores `xattr` copy failures but we probably want to know about those failures. I don't think it is fine to "hope for the best" without any control: e.g. RH-based OS depends on xattr for the system to work at all (SELinux) and it would not be fine for them to silently vanish because host has them disabled (e.g. `noxattr` mount option, FAT32 intermediary, some networked FS).

mkosi/tree.py
```diff
@@ -148,7 +148,35 @@ def copy() -> None:
...
+                    user.*            permissions   #   allow well-known 'user'
+                    security.*        permissions   #   allow well-known 'security'
+                                                    # but 
+                    trusted.SGI_*     skip          #   skip XFS-specific namespace (not portable)
+                    user.Beagle.*     skip          #   skip Beagle index data (not portable)
+                    security.evm      skip          #   skip evm (may only be written by kernel)
+                    """
+                )
+            )
+            local_options = options + ["--ro-bind", p, "/etc/xattr.conf"]
```
daandemeyer: Please just inline this into the call

mkosi/tree.py
```diff
@@ -148,7 +148,35 @@ def copy() -> None:
...
+                    trusted.*         permissions   #   allow well-known 'trusted'
+                    user.*            permissions   #   allow well-known 'user'
+                    security.*        permissions   #   allow well-known 'security'
+                                                    # but 
+                    trusted.SGI_*     skip          #   skip XFS-specific namespace (not portable)
+                    user.Beagle.*     skip          #   skip Beagle index data (not portable)
+                    security.evm      skip          #   skip evm (may only be written by kernel)
+                    """
+                )
+            )
```
daandemeyer, line 177: Can we make this a bit smarter and also copy the btrfs attrs if we detect the destination filesystem is also btrfs?

gdonval: The only `btrfs` xattr I know about is `btrfs.compression`, which is definitely something I think should be set by `systemd-repart` (`Compression=...`), not copied over from whatever host configuration is.

Normal file attributes (`chattr`) are **not** affected by this barred one: `btrfs` that calculates the `c`-flag on the fly depending on `btrfs.compression` xattr, which we've established is not something we should be copying over anyway.

daandemeyer: Have we established that? I don't remember agreeing to that. If a file is compressed on one btrfs filesystem, why shouldn't it be compressed after copying?

gdonval: > If a file is compressed on one btrfs filesystem, why shouldn't it be compressed after copying?

Because the one who "chose" to compress it was myself from 7 years ago, not the package manager. 

On my system, **I**, myself, chose to `btrfs set property set / compression "zstd"`, 7 years ago when I installed my local system. And now, this is inherited by mkosi's workspace folder (as xattr `btrfs.compression = zstd`).

Do we want to persist local arbitrary choices or do we want that choice to be explicit made in `repart.d` files?

If you want `btrfs.compression` to be copied, you need to make sure you start with a clean slate (i.e. clean file system) so that you can ensure this was an explicit packaging choice.

gdonval: I can make it smarter, but if we do, we might as well make a few checks on a single canary file before copying the tree and use `cp -a`, which will opportunistically keep as much metadata as possible.

gdonval: I see the problem... `systemd-repart` does not set `btrfs.compression`. `Compression=zstd` seems to actually do nothing with btrfs: it's supposed to use a `mkfs.btrfs` flag to force compression whilst copying files but `compsize` reports no compression was ever attempted. That definitely is suboptimal. Proof:

image

Debug shows the same:

image


`root-x86-64` is as generated by `systemd-repart` with `Compression=zstd` whereas `new-btrfs-root` was generated with `sudo mkfs.btrfs --compress=zstd -L "new-btrfs-root" -r ".../root-x86-64/" btrfs.img`.

Not a bug in `mkosi`, just a constraint added to a very hairy problem...

daandemeyer:
@gdonval I don't think this is the right fix since it only affects `cp` and no other tools that do copying.

A better way to handle this might be to use syscall interception. That probably sounds insane but I actually implemented this already to try and make selinux relabeling possible unprivileged (https://github.com/systemd/mkosi/compare/main...DaanDeMeyer:mkosi:seccomp-notify). That doesn't actually work because you can't set user xattrs on symlinks, but the same syscall interception mechanism could be used to prevent messing around with questionable xattrs possible.

daandemeyer:
Closing this for now since I doubt it's the right approach

gdonval:
@DaanDeMeyer I agree and I don't think what you suggest is insane. I was already unhappy with the current approach too: having to write into a hardcoded file, as "root", so that `cp` just behaves is insane.

I tried other stuff:

1. `[Fail]` get upstream to add a `xattr` filter flag to `cp` 
2. `[Fail]` use rust `cp` reimplementation
3. `[Fail]` use FUSE to generate and mount temporary file-backed workspaces
4. `[Fail]` use SUID mounters to achieve the same but with working drivers
5. `[Fail]` write a tiny overlay FUSE FS to preserve `xattr` in own namespace and passthrough file copy to the underlying FS.


For posterity...

## Upstream

Idea: adequate CLI options could make early failure detection work.

Upstream said no. The argument was to use `cp -a`, which does not fail if all `xattr` cannot be copied and, after the fact, check that sufficient `xattr` have been copied. Suggested CLI options would be clunky. This use-case is outside `cp`'s remit.

Does not solve the lossy `xattr` copy problem anyway: it would just detect it and that would just allow us to specify what it OK to ignore or not.

## Rust `cp`

Idea: a different implementation might have adequate machinery to preserve `xattr`.

Shares the same limitations as `cp`, not 1-1 feature/API-wise and `mkosi` fails pretty quickly with it.  

## File-backed FUSE workspace

Idea: create a workspace to install packages in that is on the same FS as mkosi target. `xattr` are guaranteed to be "as installed" and `mkosi` can be used to alter `xattr` at any point (neat!).

But FUSE drivers for ext4/btrfs/xfs, if they exist, are very limiting and slow, to put it kindly.

## SUID/Polkit mounter

Idea: same as above but with kernel drivers and privileged mounter (like the one Gnome or KDE use to make ext4/btrfs/xfs pendrives accessible without asking for root password).

Well, SUID is not neat, especially not when compounded with containers. And, for existing tools, naming conventions/restrictions of mount points with existing tools makes them a very bad fit (you don't get guaranteed-stable names and nested mounts are a nope). The new `systemd-mountfsd` helper does not make sense for this specific use-case (due to restrictions).

## Overlay FUSE FS

Idea: create a tiny fuse overlay FS that losslessly preserves `xattr` but is passthrough for everything else.

With python FUSE library... I couldn't get the passthrough to work (broke reflink copies) but even if I did, I'm not sure it would be sufficient: I'm pretty sure some package manager post-install scripts try very hard to determine what FS the files are installed on, this breaks that. Plus, we are still stuck with a unwieldy mess: `xattr` are *all* preserved until the final copy (neat) and we only care about what has been explicitly set (very neat) but we still end up with a final `cp -a` that is no better than `shutil.copy`.

daandemeyer:
@gdonval So I'm pretty sure my seccomp notify stuff would be able to handle this use case as well, I just need to come up with a decent storage alternative instead of user xattrs and then we could store all the problematic stuff in the sidecar storage instead of on the file itself.

gdonval:
> because you can't set user xattrs on symlinks

[Ah I see](https://lwn.net/ml/linux-kernel/20210902152228.665959-1-vgoyal@redhat.com/)... that explains why my FUSE stuff worked on simple files but not on a full tarball.


> I'm pretty sure my seccomp notify stuff would be able to handle this use case as well

Just skimmed through the code. I was not familiar with seccomp notifier. It looks nifty though, without the FUSE edge-cases and it seems to have been designed with container use-cases in mind. I don't know if sqlite would be performant enough for this but that makes much more sense to me than the `xattr.conf` file thing.

PR #3755, opened 2025-06-01
mkosi-initrd: Add atkbd and i8042 modules to the default initrd modules

DaanDeMeyer:
Required to get my laptop keyboard working in an initrd with just the default modules.

### DIFF
--- a/mkosi/resources/mkosi-initrd/mkosi.conf
+++ b/mkosi/resources/mkosi-initrd/mkosi.conf
@@ -46,6 +46,7 @@ KernelModules=
         amd_ctl
         amd-pmc
         amd64_edac
+        atkbd
         autofs4
         binfmt_misc
         btrfs
@@ -73,6 +74,7 @@ KernelModules=
         i2c_hid_acpi
         i2c-mux
         i2c-smbus
+        i8042
         intel-gtt
         intel_rapl_common
         intel-uncore-frequency-common
@@ -131,4 +133,3 @@ KernelModules=
         xhci-pci-renesas
         /fs/nls/
         crypto/
-


poettering:
Curious: which distro doesn't include those drivers in their kernels built-in? weird stuff...

daandemeyer:
> Curious: which distro doesn't include those drivers in their kernels built-in? weird stuff...

Arch Linux

poettering:
> Arch Linux

so if the kernel freezes at early boot and usually blinks the kbd leds, that's not going to work on arch? And if systemd freezes early, then ctrl-alt-del is not going to do a thing?

arch should really fix this, it's an amazingly bad idea

daandemeyer:
> > Arch Linux
> 
> so if the kernel freezes at early boot and usually blinks the kbd leds, that's not going to work on arch? And if systemd freezes early, then ctrl-alt-del is not going to do a thing?
> 
> arch should really fix this, it's an amazingly bad idea

https://gitlab.archlinux.org/archlinux/packaging/packages/linux/-/issues/142

HanabishiRecca:
> Curious: which distro doesn't include those drivers in their kernels built-in? weird stuff...

Most PCs use USB keyboards nowdays, so not so useful anyway. And even not all laptops actually connect the keyboard via PS/2 interface.
Or do you suggest to make the entire USB/HID stack built-in too?

HanabishiRecca:
Ok, Arch Linux made them built-in along with USB HID modules.

poettering:
> Most PCs use USB keyboards nowdays, so not so useful anyway. And even not all laptops actually connect the keyboard via PS/2 interface.
> Or do you suggest to make the entire USB/HID stack built-in too?

Yeah, that's precisely what I suggest: make USB + HID stuff built-in. It's what other distros tend to do, too. It's just so basic. In the very ver few cases where a system has no USB at all, you might not want a generic kernel anyway, and otherwise wasting a few K in memory is probably not the worst. The benefits on having something that "just works" and is faster really trump the corner cases where you have no USB at all, I'd say.

PR #3760, opened 2025-06-04
ci: Drop differential shellcheck

DaanDeMeyer:
We already run shellcheck on all scripts as part of the main CI job, so we don't really get any benefit of a differential shellcheck as all our code is already checked by shellcheck so let's just drop it.


behrmann:
No need to waste CI minutes on this.

PR #3762, opened 2025-06-04
mkosi-tools: move 'man' pkg from generic list to distro-specific lists

craftyguy:
Follow up to 94cc136fbe, this moves the man pkg to distro-specific pkg lists to support distros that use a different name for this package.


behrmann:
I'm not sure I understand this change. None of the distributions use a different name, so this is a noop?

daandemeyer:
This is for postmarketos but as mentioned in the chat, we won't merge it separately, only as part of the PR introducing postmarketos

craftyguy:
Yep, I'll close this issue and make a separate MR later with this patch and others to add support for postmarketOS. Sorry for the noise!

PR #3763, opened 2025-06-04
detect kernel ver from modules sub dir if unable to extract from filename

craftyguy:
Some distros don't include the kernel version in the filename of the kernel image,  so the normal version detection mechanisms don't work. Grabbing the first dir under this path that looks like a kernel version should work for us, since mkosi is always building a new image there should only ever be 1 modules dir and it should be for the only kernel that was installed.

For some additional context, I'm working on reviving/fixing patches to add support for postmarketOS. pmOS (and Alpine Linux) don't currently support multiple concurrent kernels being installed. I'm hoping that this change is acceptable, I don't intend to change existing behavior for other distros), and we can back it out some time in the future when we can shake this quirk of ours.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1259,6 +1259,20 @@ def gzip_binary(context: Context) -> str:
     return "pigz" if context.config.find_binary("pigz") else "gzip"
 
 
+def _kernel_get_ver(context: Context) -> str:
+    path = context.root / "usr/lib/modules"
+    # try to get version from the first dir under usr/lib/modules
+
+    if not path.exists():
+        return "unknown"
+
+    for item in path.iterdir():
+        if item.is_dir() and re.match(r'\d+\.\d+.*', item.name):
+            return item.name
+
+    return "unknown"
+
+
 def fixup_vmlinuz_location(context: Context) -> None:
     # Some architectures ship an uncompressed vmlinux (ppc64el, riscv64)
     for type in ("vmlinuz", "vmlinux"):
@@ -1266,7 +1280,11 @@ def fixup_vmlinuz_location(context: Context) -> None:
             if d.is_symlink():
                 continue
 
-            kver = d.name.removeprefix(f"{type}-")
+            # Extract kernel version pattern from filename
+            filename = d.name.removeprefix(f"{type}-")
+            match = re.search(r'\d+\.\d+[\w\.-]*', filename)
+            kver = match.group(0) if match else _kernel_get_ver(context)
+
             vmlinuz = context.root / "usr/lib/modules" / kver / type
             if not vmlinuz.parent.exists():
                 continue


mkosi/__init__.py
```diff
@@ -1259,14 +1259,32 @@ def gzip_binary(context: Context) -> str:
...
 
 
+def _kernel_get_ver(context: Context) -> str:
+    path = context.root / "usr/lib/modules"
+    # try to get version from the first dir under usr/lib/modules
+
+    if not path.exists():
+        return "unknown"
+
+    for item in path.iterdir():
```
behrmann, line 1269: This could be combined into
```suggestion
    for item in path.glob("*"):
```
and if you relax the regex (assuming any dir under `/usr/lib/modules` starting with a digit is a kernel) it could just be
```suggestion
    for item in path.glob("[0-9]*"):
```

mkosi/__init__.py
```diff
@@ -1259,14 +1259,32 @@ def gzip_binary(context: Context) -> str:
     return "pigz" if context.config.find_binary("pigz") else "gzip"
 
 
+def _kernel_get_ver(context: Context) -> str:
```
behrmann, line 1262: That name is a bit generic, how about
```suggestion
def _kernel_get_ver_from_modules(context: Context) -> str:
```

mkosi/__init__.py
```diff
@@ -1259,14 +1259,32 @@ def gzip_binary(context: Context) -> str:
...
+    path = context.root / "usr/lib/modules"
```
behrmann, line 1263: Also less generic:
```suggestion
    moddir = context.root / "usr/lib/modules"
```

mkosi/__init__.py
behrmann: Also, this does assume that the sorting order is reasonable and as you say, there will usually only be one, but people do weird stuff in scripts, so these assumptions might not hold generally, so it would be better to sort this by version (we have a version object for that).

septatrix:
I just noticed that another, maybe cleaner solution would be to update your Linux packages to ship the kernel under `/usr/lib/modules/<over>/` and put a symlink to it under `/boot`. That is after all pretty much everything this function does at all. If that already were the case it would simply bail out with the first if-condition

craftyguy:
> I just noticed that another, maybe cleaner solution would be to update your Linux packages to ship the kernel under `/usr/lib/modules/<over>/` and put a symlink to it under `/boot`. That is after all pretty much everything this function does at all. If that already were the case it would simply bail out with the first if-condition

Interesting idea, but maybe link it the other way around since /boot may be ESP (and some of those are vfat, never mind that the link would be broken at boot). I'd have to think about this, since it would mean changing how hundreds of kernels in pmOS are installed.


As discussed in chat, I'll close this issue and make a separate MR later with this patch and others to add support for postmarketOS. Sorry for the noise!

PR #3766, opened 2025-06-11
Add "nfs" (and "network") initrd profiles

aafeijoo-suse:
nfs-utils-2.8.4 will provide its own nfsroot-generator [1] to allow mounting the real rootfs via NFSv4, so we can create an "nfs" initrd profile to support it. Also, this requires some type of networking in the initrd, so I also created a "network" profile to enable systemd-networkd.

Tested in openSUSE Tumbleweed using a custom OBS build [2] and in Fedora rawhide, since it already ships nfs-utils-2-8-4-rc2 [3].

[1] http://git.linux-nfs.org/?p=steved/nfs-utils.git;a=commit;h=ed86ea08dadafbac948c6a45629a6f3282a77233
[2] https://build.opensuse.org/package/show/home:afeijoo:branches:openSUSE:Factory:nfsroot/nfs-utils
[3] https://src.fedoraproject.org/rpms/nfs-utils/c/e7f85470bc0ba53a3338468f50ffc62bf0f03cba?branch=rawhide

### DIFF
--- a/mkosi/resources/mkosi-initrd/mkosi.profiles/nfs/mkosi.conf
+++ b/mkosi/resources/mkosi-initrd/mkosi.profiles/nfs/mkosi.conf
@@ -1,0 +1,7 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+[Content]
+KernelModules=
+        fs/nfs/
+        net/sunrpc/
+        nfs_acl
--- a/mkosi/resources/mkosi-initrd/mkosi.profiles/nfs/mkosi.conf.d/fedora.conf
+++ b/mkosi/resources/mkosi-initrd/mkosi.profiles/nfs/mkosi.conf.d/fedora.conf
@@ -1,0 +1,9 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+[Match]
+Distribution=fedora
+
+[Content]
+Packages=
+        nfs-utils
+        libnfsidmap
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -637,6 +637,7 @@ class ToolsTreeProfile(StrEnum):
 class InitrdProfile(StrEnum):
     lvm = enum.auto()
     network = enum.auto()
+    nfs = enum.auto()
     pkcs11 = enum.auto()
     plymouth = enum.auto()
     raid = enum.auto()
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -636,6 +636,7 @@ class ToolsTreeProfile(StrEnum):
 
 class InitrdProfile(StrEnum):
     lvm = enum.auto()
+    network = enum.auto()
     pkcs11 = enum.auto()
     plymouth = enum.auto()
     raid = enum.auto()
--- a/mkosi/initrd.py
+++ b/mkosi/initrd.py
@@ -196,6 +196,20 @@ def process_crypttab(staging_dir: Path) -> list[str]:
     return cmdline
 
 
+def network_config() -> list[str]:
+    return [
+        f"--extra-tree={f}:{f}"
+        for f in (
+            "/etc/systemd/network",
+            "/etc/systemd/networkd.conf",
+            "/etc/systemd/networkd.conf.d",
+            "/etc/systemd/resolved.conf",
+            "/etc/systemd/resolved.conf.d",
+        )
+        if Path(f).exists()
+    ]
+
+
 def raid_config() -> list[str]:
     return [
         f"--extra-tree={f}:{f}"
@@ -333,7 +347,9 @@ def main() -> None:
 
         for p in args.profile:
             cmdline += ["--profile", p]
-            if p == "raid":
+            if p == "network":
+                cmdline += network_config()
+            elif p == "raid":
                 cmdline += raid_config()
 
         if args.kernel_image:
--- a/mkosi/initrd.py
+++ b/mkosi/initrd.py
@@ -210,6 +210,10 @@ def network_config() -> list[str]:
     ]
 
 
+def nfs_config() -> list[str]:
+    return [f"--extra-tree={f}:{f}" for f in ("/etc/idmapd.conf", "/etc/idmapd.conf.d") if Path(f).exists()]
+
+
 def raid_config() -> list[str]:
     return [
         f"--extra-tree={f}:{f}"
@@ -349,6 +353,8 @@ def main() -> None:
             cmdline += ["--profile", p]
             if p == "network":
                 cmdline += network_config()
+            elif p == "nfs":
+                cmdline += nfs_config()
             elif p == "raid":
                 cmdline += raid_config()
 
--- a/mkosi/resources/man/mkosi-initrd.1.md
+++ b/mkosi/resources/man/mkosi-initrd.1.md
@@ -46,6 +46,8 @@ initrds and Unified Kernel Images for the current running system.
 
     The `lvm` profile enables support for LVM.
     The `network` profile enables support for network via **systemd-networkd**.
+    The `nfs` profile enables support for NFS. It requires networking in the
+    initrd, using the `network` profile, or some other custom method.
     The `pkcs11` profile enables support for PKCS#11.
     The `plymouth` profile provides a graphical interface at boot (animation and
     password prompt).
--- a/mkosi/resources/man/mkosi-initrd.1.md
+++ b/mkosi/resources/man/mkosi-initrd.1.md
@@ -45,6 +45,7 @@ initrds and Unified Kernel Images for the current running system.
     disabled.
 
     The `lvm` profile enables support for LVM.
+    The `network` profile enables support for network via **systemd-networkd**.
     The `pkcs11` profile enables support for PKCS#11.
     The `plymouth` profile provides a graphical interface at boot (animation and
     password prompt).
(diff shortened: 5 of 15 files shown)


behrmann (approved):

mkosi/resources/mkosi-initrd/mkosi.profiles/nfs/mkosi.conf.d/fedora.conf
```diff
@@ -0,0 +1,12 @@
...
+[Match]
+Distribution=fedora
+
+[Content]
+Packages=
+        nfs-utils
+        libnfsidmap
+
+        # 60-nfs.rules calls sysctl
+        procps-ng
```
septatrix: It took me a while to figure out that this meant the udev rules files, maybe add that to the comment.
Also, maybe nfs-utils could be persuaded to instead call `systemd-sysctl` (which the udev docs also recommend)?

mkosi/resources/mkosi-initrd/mkosi.profiles/nfs/mkosi.conf
```diff
@@ -0,0 +1,8 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+[Content]
+KernelModules=
+        fs/nfs/
+        net/ipv6/
```
septatrix: Why not also ipv4? Also maybe this should rather be part of the network profile and not only the nfs profile?

mkosi/resources/mkosi-initrd/mkosi.profiles/nfs/mkosi.conf.d/fedora.conf
aafeijoo-suse: > It took me a while to figure out that this meant the udev rules files, maybe add that to the comment.

Ok

> Also, maybe nfs-utils could be persuaded to instead call systemd-sysctl (which the udev docs also recommend)?

I'd say this request is outside the scope of this PR :)

mkosi/resources/mkosi-initrd/mkosi.profiles/nfs/mkosi.conf
aafeijoo-suse: My bad, [ipv6 is only required for rpcbind](https://git.linux-nfs.org/?p=steved/rpcbind.git;a=blob;f=systemd/rpcbind.socket.in;h=5dd09a143e167ed8930704309c22cfb3a7428216;hb=HEAD#l12), hence we don't need those modules here since the generator is for NFSv4.

In any case, it wouldn't hurt to always add some network kernel modules with the network profile. We could use dracut as a reference to start: https://github.com/dracutdevs/dracut/blob/master/modules.d/90kernel-network-modules/module-setup.sh#L39-L44

mkosi/resources/mkosi-initrd/mkosi.profiles/nfs/mkosi.conf.d/fedora.conf
behrmann, line 1: Could you add a file for Arch and Debian as well? In Arch the required packages should be `nfs-utils` and `procps-ng` with the idmapping library being required by `nfs-utils`. In Debian the required packages should be `procps` (for `sysctl`) and `nfs-common` (for `mount.nfs4`, `libnfsidmap1` is also a hard dependency of this).

mkosi/resources/mkosi-initrd/mkosi.profiles/nfs/mkosi.conf
behrmann: That sounds reasonable.

mkosi/resources/mkosi-initrd/mkosi.profiles/nfs/mkosi.conf.d/fedora.conf
aafeijoo-suse: I could, but do you want to add this before the version of nfs-utils/nfs-common that provides this feature is released? Also, that would require some changes in the network profile, since at least Debian/Ubuntu ship systemd-networkd with the systemd package.

behrmann: Ah, forgot again that this is yet unreleased. 

I think I'd add them and note in the docs, that this needs at least version 2.8.4 or adding a generator of ones own. That seems easier than monitoring when these needed versions migrate into releases. For Debian this will probably be a while anyway, since they've frozen for trixie already and something newer than 2.8.3 will only make it into a release in two years. What do you think, @DaanDeMeyer?

mkosi/resources/mkosi-initrd/mkosi.profiles/network/mkosi.extra/usr/lib/systemd/system-preset/10-mkosi-initrd-network.preset
```diff
@@ -0,0 +1,4 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+enable systemd-networkd.service
+enable systemd-resolved.service
```
daandemeyer: So I'm OK with going with this approach for now but can I interest you by any chance to add support for /usr/lib/systemd/initrd-preset to systemd? I'm confident that sharing presets between initrd and main system is not going to scale in the long run.

aafeijoo-suse: I can only see problems with cases like the Debian/Ubuntu mentioned before, where systemd-networkd is shipped with the systemd package, so maybe they would enable it without wanting to do it, because the systemd package is always included. That said, IMO it should be on a separate package...

mkosi/resources/mkosi-initrd/mkosi.profiles/nfs/mkosi.conf.d/fedora.conf
daandemeyer: Yeah let's just add all of these right away, no need to wait until the distribution ships a recent enough version of nfsutils

mkosi/resources/mkosi-initrd/mkosi.profiles/nfs/mkosi.conf.d/fedora.conf
```diff
@@ -0,0 +1,12 @@
...
+[Match]
+Distribution=fedora
+
+[Content]
+Packages=
+        nfs-utils
+        libnfsidmap
+
+        # udev rules 60-nfs.rules calls sysctl
+        procps-ng
```
daandemeyer: These are useful debugging tools any dependencies that we don't already pull in, maybe lets' install procps-ng by default and not via a profile?

mkosi/resources/mkosi-initrd/mkosi.profiles/nfs/mkosi.conf.d/fedora.conf
aafeijoo-suse: Ok, added.

daandemeyer:
I am introducing initrd presets in https://github.com/systemd/systemd/pull/38028 btw

mkosi/resources/mkosi-initrd/mkosi.profiles/network/mkosi.extra/usr/lib/systemd/system-preset/10-mkosi-initrd-network.preset
aafeijoo-suse: Removed since https://github.com/systemd/systemd/pull/38028 was merged.

aafeijoo-suse: Hmm... https://github.com/systemd/systemd/pull/38059

daandemeyer: @aafeijoo-suse What's the problem?

aafeijoo-suse: Maybe I don't understand what the latest PR mean, so please correct me if I'm wrong. If `99-default-initrd.preset` contains `disable *` and is parsed after `90-systemd-initrd.preset`, doesn't that mean that everything is disabled, and therefore the former directives with `enable systemd-networkd.service` and `enable systemd-resolved.service`?

daandemeyer: @aafeijoo-suse In case of presets the first directive that applies to the service takes precedence, not the latest.

aafeijoo-suse: ok, thanks, I just found it: `If multiple lines apply to a unit name, the first matching one takes precedence over all others.`

mkosi/resources/mkosi-initrd/mkosi.profiles/network/mkosi.conf
```diff
@@ -0,0 +1,7 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+[Match]
+Distribution=!arch
+
+[Content]
+Packages=systemd-resolved
```
daandemeyer: Let's do this a bit more structured, I'd leave mkosi.conf empty and just have mkosi.conf.d/ with two files: systemd-networkd.conf and systemd-resolved.conf, each with the appropriate matches in them.

aafeijoo-suse: ok, ptal.

PR #3768, opened 2025-06-11
Bring back --default to allow OBS builds to select an entry point among many

bluca:
When building a mkosi repository called 'particleos', OBS calls mkosi as such:

mkosi --directory /usr/src/packages/SOURCES/particleos --default mkosi.conf

with mkosi.conf being the source file extracted at source time, ie: defined by the user, among the many files available in the repository. So this can be used to select the entry point, that defines what to build, when a repository provides many build types, profiles, images, etc.

Right now --default is a no-op, so make it add the passed conf file as a high priority config file, so that the entry point can have precedence over anything else.

The file is placed at:

/usr/src/packages/SOURCES/mkosi.conf

This cannot be controlled by the user, and it's fixed.

Config file parsing requires absolute paths, and aborts otherwise. Because it is called with a filename only, but mkosi executes from the --directory location, special case it and if a non-absolute path is passed, look in the parent of --directory so that the use case of providing an entry point can be enabled.

With this I can add entry points to hidden directories in a repo, or even outside of it, and select them in the `_service` on OBS and build multiple different images from the same git repo and branch, for example the particleos gnome image:

https://build.opensuse.org/projects/home:bluca:branches:home:bluca:systemd/packages/particleos/files/_service?expand=1


tests/test_json.py
```diff
@@ -61,6 +61,7 @@ def test_args(path: Optional[Path]) -> None:
             "DebugSandbox": false,
             "DebugShell": false,
             "DebugWorkspace": false,
+            "DefaultConfig": {f'"{os.fspath(path)}"' if path is not None else "null"},
```
behrmann: Nit: I know this is only a test, but `Directory` and `DefaultConfig` having the same non-null value will never make sense. If somebody looks at this to see how something is used (I did in the past), this would be confusing, so maybe just hard-coding
```suggestion
            "DefaultConfig": null,
```
would be better.

mkosi/config.py
```diff
@@ -5146,6 +5156,15 @@ def parse_config(
...
     context.config["directory"] = args.directory
 
+    # Config paths need to be absolute - as a special case, if it is not, look in the parent
+    # of --directory. This works nicely on OBS, where the --directory is the git repository
+    # and the --default is the mkosi.conf selected at source service time, placed in the parent.
+    if args.default_config is not None:
+        if args.directory is not None and not args.default_config.is_absolute():
+            context.default_config = args.directory.parent / args.default_config
+        else:
+            context.default_config = args.default_config
```
behrmann, line 5166: Since the `args.default_config` is `None` anyways, this could be shortened into
```suggestion
    if (
        args.default_config is not None:
        and args.directory is not None 
        and not args.default_config.is_absolute()
    ):
        context.default_config = args.directory.parent / args.default_config
```
(well, it's longer, but slightly less complex)

behrmann (commented):
I'm not sure having this undocumented is a good idea. Having surprise config pickup from a parent could be problematic, we might forget this exists, I sure did already forget `--default` still existed as a noop. Besides that I don't have strong objections.

bluca:
> I'm not sure having this undocumented is a good idea. Having surprise config pickup from a parent could be problematic, we might forget this exists, I sure did already forget `--default` still existed as a noop. Besides that I don't have strong objections.

I have no preference, given it's for a specific use case I thought you preferred to have it hidden, but I'll change that

mkosi/config.py
bluca: With that change, unless args.directory is set, args.default_config is ignored, even if it's an absolute path which would otherwise work

mkosi/resources/man/mkosi.1.md
```diff
@@ -256,6 +256,12 @@ Those settings cannot be configured in the configuration files.
     working directory. If the empty string is specified, all configuration in
     the current working directory will be ignored.
 
+`--default=`
+:   Takes an absolute path to a configuration file. **mkosi** parses this file
+    before doing anything. This allows a user to select an entry point to
+    programmatically decide among various available options (e.g.: profiles)
+    without having to select them on the command line.
+
```
septatrix, line 264: I do not quite understand how this is different from an `--include=`, especially if that include points to a directory with a `mkosi.local.conf` file

septatrix (commented):
Do I understand it correctly that this "default" file is generated dynamically by OBS?

mkosi/resources/man/mkosi.1.md
bluca: --include is not used, so it's not useful

bluca:
No, it's extracted from wherever, either the same repo or another

mkosi/resources/man/mkosi.1.md
septatrix: Wouldn't this rather be something to fix in the OBS integration of mkosi?

bluca: No, backward compatibility must be maintained, that is staying exactly as it is

daandemeyer:
This all looks extremely ugly to me.

Why don't we just special case OBS by looking for /usr/src/packages/SOURCES/mkosi.conf explicitly and parsing it if it exists? In this case I prefer a small OBS specific hack that doesn't leak into the interface rather than adding more ugly interfaces just to deal with OBS's shortcomings.

bluca:
That would be fine by me, but as discussed it requires another workaround, opened a new PR since it's a different approach from this: https://github.com/systemd/mkosi/pull/3790

PR #3784, opened 2025-06-26
Add support for Amazon Linux 2023

Vasu77df:
[Amazon Linux 2023](https://aws.amazon.com/linux/amazon-linux-2023/) is similar to Azure linux and Fedora.

We would like to start building System Extension images of Amazon Linux 2023 and adopting for aws workflows, and use mkosi for this process.

Some quirks are Amazon Linux 2023 is a new distribution in itself, it's not a version like a successor of AL2, so we only support building Amazon Linux 2023, as AL2 is going to be end of life by 2026 so it is not worth supporting AL2.

Also Amazon Linux 2023 is not supported for bare metal devices, we only focus on supporting for VMs, and system extension images, and nspawn containers.

Amazon Linux 2023 also does not have the packages available to be a tool tree for mkosi, you can use other distribution as tool tree to build Amazon Linux 2023

I have attached the build output as a test output as reference:
- [amazonlinux_2023_build_output.txt](https://github.com/user-attachments/files/20934324/amazonlinux_2023_build_output.txt)


mkosi/distributions/amazonlinux.py
```diff
@@ -0,0 +1,69 @@
...
+
+from mkosi.config import Architecture
+from mkosi.context import Context
+from mkosi.distributions import fedora, join_mirror
+from mkosi.installer.dnf import Dnf
+from mkosi.installer.rpm import RpmRepository, find_rpm_gpgkey, setup_rpm
+from mkosi.log import die
+
+
+class Installer(fedora.Installer):
```
behrmann: So, since this is something of a Fedora spin-off, should this get a Fedora tools tree by default, like CentOS or Ubuntu? If not, why not?

behrmann (commented):
First off: Could you squash the three commits into one, since the last three are only fixups?

More substantially: Will you be around to fix this, if this breaks? Also, to see whether this breaks, it would be preferable if CI were added to it. It can skip all the unsupported cases, if you only support building on a specific distro, that's also fine, but what's not covered by CI will never be noticed if it works or not (Mageia? Mandriva?) and it will bitrot.

mkosi/distributions/amazonlinux.py
Vasu77df: Squashed the commits, sorry about the miss there. 

While I am not a core maintainer of Amazon Linux 2023, I am employed by amazon, until then I will be around to support it and can work with maintainers in Amazon Linux 2023 to fix any issues if they arise.

I will add support for CI, and setting the default tools tree for this, and update this PR in a couple of days.

daandemeyer:
I actually have  a PR sitting locally for this as well, the reason I haven't pushed it yet is because of https://github.com/rpm-software-management/dnf5/issues/1321, when building amazon linux images dnf5 tried to do things with amazon cdn urls with '+'s in them and that did not go well. AFAIK that issue still happens, can you verify you can build amazon linux images with dnf5? If not I'd prefer that bug in dnf5 is fixed first since we're looking to only support dnf5 and not dnf anymore.

Vasu77df:
Got it, I have been using a debian tools tree to build amazon linux 2023, so I have did not hit this issue pattern, I guess the CI matrix would have caught this :). I am trying to find some spare time to close the loop on this. I will test if this works with dnf5 if it does not, I'll look into if what I can do to move this forward.

daandemeyer:
@Vasu77df The dnf5 issue was fixed, so I think we can move forward with adding Amazon Linux support now.

Please drop the merge commit.

Vasu77df:
That's great news, thanks for sharing the update, let me pick this back up again, add CI tests, and cleanup the PR.

pendo324:
Hey @DaanDeMeyer, did @Vasu77df happen to communicate why this PR was abandoned via some other channels? I'd be interested in seeing Amazon Linux 2023 support in mkosi and can try to take this over if there's no current owner

Vasu77df:
Sorry about not providing an update before closing. 
While the DNF5 issue was fixed it did not make it into Fedora Rawhide yet. I just checked and looks like, it's been packaged and available in Rawhide, so we can wrap this up now.
Due to some personal work, I have not been able to get to this, and my workspace/laptop was borked hence closed the CR. I might be able to get back to this, this weekend.

PR #3788, opened 2025-06-30
mkosi-tools: Make sure opensc module is registered with p11-kit

DaanDeMeyer:
Otherwise openssl will fail to load keys off a yubikey when using
pkcs11-provider. For more discussion and why this isn't the default
everywhere, see https://gitlab.archlinux.org/archlinux/packaging/packages/opensc/-/issues/2.


daandemeyer:
Pushed one more commit on top to revert the roothash= changes for root=dissect since in the end root=dissect still relies on roothash= and usrhash= to figure out the right partition to use.

mkosi/backport.py
```diff
@@ -28,7 +31,9 @@ def _tempfile(
     fd, raw_path = tempfile.mkstemp(suffix=suffix)
     try:
         try:
-            os.write(fd, reader())
+            os.write(fd, path.read_bytes())
+            if os.access(path, os.X_OK):
+                 os.fchmod(fd, path.stat().st_mode | stat.S_IEXEC)
```
behrmann: Hard to see, but ruff complains, that's indented five not four.

mkosi/resources/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+
+    return path
+
+
+@contextlib.contextmanager
+def temporary_file(path: Traversable, suffix: str = "") -> Iterator[Path]:
+    # Not using tempfile.NamedTemporaryFile as it leads to deeper 'try'
+    # blocks due to the need to close the temporary file to work on Windows
+    # properly.
+    fd, raw_path = tempfile.mkstemp(suffix=suffix)
```
behrmann: Since this moves away from the original, this comment can probably go, as this will never be used in Windows now.

mkosi/resources/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+from collections.abc import Iterator
+from contextlib import AbstractContextManager
+from importlib.abc import Traversable
+from pathlib import Path
+
+
+def make_executable_if_script(path: Path) -> Path:
+    with path.open() as f:
+        buf = f.read(3)
+        if buf == "#!/":
```
behrmann: This should fix the test failure.
```suggestion
    with path.open("rb") as f:
        buf = f.read(3)
        if buf == b"#!/":
```

mkosi/resources/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+# Based on code from https://github.com/python/cpython/blob/main/Lib/importlib/resources/_common.py
+
+import contextlib
+import functools
+import os
+import stat
+import tempfile
+from collections.abc import Iterator
+from contextlib import AbstractContextManager
+from importlib.abc import Traversable
```
behrmann: Since it does complain, this could be made, since we probably will have to add this anyway once 3.14 is released.
```python
import sys
if sys.version_info < (3, 13):
    from importlib.abc import Traversable
else:
    from importlib.resources.abc import Traversable
```

daandemeyer:
CI failures are unrelated

PR #3796, opened 2025-07-05
Fixes for arm64 images, add CI job

bluca:
Some fixes to build arm64 images, and add a build-only CI job for it

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,10 +100,10 @@ jobs:
         run: tools/make-man-page.sh
 
   integration-test:
-    runs-on: ubuntu-24.04
+    runs-on: ${{ matrix.runner }}
     needs: unit-test
     concurrency:
-      group: ${{ github.workflow }}-${{ matrix.distro }}-${{ matrix.tools }}-${{ github.ref }}
+      group: ${{ github.workflow }}-${{ matrix.distro }}-${{ matrix.tools }}-${{ matrix.runner }}-${{ github.ref }}
       cancel-in-progress: true
     strategy:
       fail-fast: false
@@ -122,6 +122,8 @@ jobs:
           - fedora
           - opensuse
           - ubuntu
+        runner:
+          - ubuntu-24.04
         exclude:
           # pacman is not packaged in EPEL.
           - distro: arch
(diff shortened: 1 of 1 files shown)


mkosi/qemu.py
```diff
@@ -700,9 +700,9 @@ def finalize_firmware(
     ):
         return Firmware.linux
 
-    # At the moment there are no qemu firmware descriptions for non-x86 architectures that advertise
-    # secure-boot support so let's default to no secure boot for non-x86 architectures.
-    if config.architecture.is_x86_variant():
+    # At the moment there are no qemu firmware descriptions for non-x86/arm architectures that advertise
+    # secure-boot support so let's default to no secure boot for non-x86/arm architectures.
+    if config.architecture.is_x86_variant() or config.architecture.is_arm_variant():
```
daandemeyer: Can you guarantee the necessary firmware descriptions are shipped for all distros? From my experience these can be wildly different across distros.

daandemeyer: Especially with older qemu versions in the mix

bluca: It is on debian/ubuntu so I'll limit this to that set

daandemeyer: Please do proper dilligence and check for various distros. Should be trivial to add CI jobs for every distro for debugging this and then removing them again once we've figured out which ones work and which ones don't.

bluca: It's really not as the CI doesn't cover it as it just times out, you have to build and boot it manually and check if the TPM is working or not. People who are interested in other distributions can send the fix, if they need the TPM to work on arm64 in their distro

tests/__init__.py
```diff
@@ -135,13 +135,15 @@ def boot(self, options: Sequence[str] = (), args: Sequence[str] = ()) -> Complet
...
 
     def vm(self, options: Sequence[str] = (), args: Sequence[str] = ()) -> CompletedProcess:
+        need_hyperv_workaround = os.uname().machine == "x86_64"
+
         result = self.mkosi(
             "vm",
             [
                 "--runtime-build-sources=no",
                 "--vsock=yes",
                 # TODO: Drop once both Hyper-V bugs are fixed in Github Actions.
```
behrmann, line 145: 7a36ebeecbee5464880662d65512c14bec8cb813 originally introduced this comment. Unfortunately the commit message has no further information, do you know what the status of these bugs is, @DaanDeMeyer?

behrmann (commented):
LGTM. Typo in the message of the third commit (enabled -> enable).

tests/__init__.py
bluca: Still broken, had disabled that by mistake and it immediately crashed

PR #3797, opened 2025-07-05
Do not try to install packages that are listed in RemovePackages=

DaanDeMeyer:
This allows using RemovePackages= in mkosi.local.conf to prevent certain packages listed in the regular configuration from being installed in the first place.

### DIFF
--- a/mkosi/installer/pacman.py
+++ b/mkosi/installer/pacman.py
@@ -198,7 +198,12 @@ class Pacman(PackageManager):
 
     @classmethod
     def remove(cls, context: Context, packages: Sequence[str]) -> None:
-        cls.invoke(context, "--remove", ["--nosave", "--recursive", *packages], apivfs=True)
+        installed = {
+            cls.parse_pkg_desc(i)[0] for i in (context.root / "var/lib/pacman/local").glob("*/desc")
+        }
+        remove = [p for p in packages if p in installed]
+        if remove:
+            cls.invoke(context, "--remove", ["--nosave", "--recursive", *remove], apivfs=True)
 
     @classmethod
     def keyring(cls, context: Context) -> None:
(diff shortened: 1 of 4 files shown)


cvlc12:
Apparently, the packages in `RemovePackages=` still appear in `Packages=` in `mkosi summary`. Is this desirable ?

cvlc12:
The problem is this now fails when trying to remove a package which has not been installed (since already removed):
```
‣  Removing 1 packages…
error: target not found: linux-firmware
‣ "pacman --root=/buildroot --logfile=/dev/null --dbpath=/var/lib/pacman --cachedir=/var/cache/pacman/mkosi --cachedir=/var/cache/pacman/pkg --hookdir=/buildroot/etc/pacman.d/hooks --arch x86_64 --color auto --noconfirm --remove --nosave --recursive linux-firmware" returned non-zero exit code 1.
```
Still wonder if using a new `SkipPackages=` would be nicer and avoid having to rework the `RemovePackages=` logic ?

daandemeyer:
RemovePackages= for pacman now avoids trying touninstall packages that aren't installed in the first place.

mkosi/installer/pacman.py
```diff
@@ -198,7 +198,11 @@ def install(
 
     @classmethod
     def remove(cls, context: Context, packages: Sequence[str]) -> None:
-        cls.invoke(context, "--remove", ["--nosave", "--recursive", *packages], apivfs=True)
+        installed = sorted((context.root / "var/lib/pacman/local").glob("*/desc"))
+        names = [cls.parse_pkg_desc(i)[0] for i in installed]
```
behrmann: the number of installed packages might be large, so this should probably be a set like in in `finalize_packages`
```suggestion
        names = {cls.parse_pkg_desc(i)[0] for i in installed}
```

cvlc12:
Seems to work !

[edit]
Except as noted above, the packages still appear in `summary` as if they will be installed

daandemeyer:
>  Except as noted above, the packages still appear in summary as if they will be installed

I think this is up for discussion. The summary is a description of the configuration, not necessarily what we'll end up doing. So I think still listing all packages in the summary is the way to go.

cvlc12:
Alright, thanks very much for adding this !

PR #3798, opened 2025-07-05
Add Skip packages

cvlc12:
Alternative to https://github.com/systemd/mkosi/pull/3797 which is currently broken because it tries to remove packages which have already been removed.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -244,18 +244,23 @@ def remove_files(context: Context) -> None:
                         t.rmdir()
 
 
+def finalize_packages(config: Config) -> list[str]:
+    s = set(config.remove_packages)
+    return [p for p in config.packages if p not in s]
+
+
 def install_distribution(context: Context) -> None:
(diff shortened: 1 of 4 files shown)


cvlc12:
Added docs. However `mkosi summary` still lists skipped packages in 'Packages=' I think it would be nicer not to.

cvlc12:
Hmm. Unsure how to fix the test

daandemeyer:
I seriously doubt this is the way to go. I mich prefer fixing RemovePackages= to not fail if the package is not installed. We already do this for various package managers, just not apt.

cvlc12:
> I seriously doubt this is the way to go. I mich prefer fixing RemovePackages= to not fail if the package is not installed. We already do this for various package managers, just not apt.

ok. However in my case tested https://github.com/systemd/mkosi/pull/3797 with pacman (not apt) and it failed

daandemeyer:
Sure, and that's something we should fix, as I now did in #3797 

Let's close this, I'm confident it's not the way to go

PR #3802, opened 2025-07-10
Add new Snapshot= setting

daandemeyer:
In systemd CI, we often run into issues caused by updates to third-party
components like the kernel package in rolling release distributions
like Arch Linux or Fedora Rawhide. When these happen, the corresponding
CI job starts failing on every PR and bisecting the distribution to figure
out when the breakage was introduced is rather tedious.

To mitigate this problem, we need to be able to pin the rolling release
distributions to a specific snapshot which we control. This allows us to
update the pinned snapshot in a PR created by a bot, so that any failures
introduced by moving to a newer snapshot will be limited to the PR that bumps
the snapshot. Any regressions can then be debugged and fixed before merging
the PR that switches us to the new snapshot.

To make this possible, let's introduce a new Snapshot= setting and implement
it for every distribution that has a snapshot concept or something that maps
to it. Per distribution:

- Debian => snapshot.debian.org (unlimited)
- Ubuntu => snapshot.ubuntu.com (unlimited)
- Arch => archive.archlinux.org (unlimited)
- OpenSUSE => download.opensuse.org/history (limited to a month of snapshots)
- CentOS => composes.stream.centos.org (limited to 3 weeks of snapshots)
- Fedora => https://kojipkgs.fedoraproject.org/ (limited to 2 weeks of snapshots)

Additionally, for CentOS, we also support using composes from mirror.facebook.net
which keeps them around forever so we get unlimited snapshots there as well for
CentOS Stream.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1344,6 +1344,7 @@ def finalize_default_initrd(
         f"--release={config.release}",
         f"--architecture={config.architecture}",
         *([f"--mirror={config.mirror}"] if config.mirror else []),
+        *([f"--snapshot={config.snapshot}"] if config.snapshot else []),
         f"--repository-key-check={config.repository_key_check}",
         f"--repository-key-fetch={config.repository_key_fetch}",
         *([f"--repositories={repository}" for repository in config.repositories]),
@@ -4243,6 +4244,10 @@ def run_box(args: Args, config: Config) -> None:
         )
 
 
+def run_latest_snapshot(args: Args, config: Config) -> None:
+    print(config.distribution.latest_snapshot(config))
+
+
 def run_shell(args: Args, config: Config) -> None:
     opname = "acquire shell in" if args.verb == Verb.shell else "boot"
     if config.output_format not in (OutputFormat.directory, OutputFormat.disk):
@@ -5101,6 +5106,10 @@ def run_verb(args: Args, tools: Optional[Config], images: Sequence[Config], *, r
 
     if args.verb == Verb.bump:
         finalize_image_version(args, last)
+        return
+
+    if args.verb == Verb.latest_snapshot:
+        run_latest_snapshot(args, last)
         return
 
     if args.verb == Verb.clean:
--- a/mkosi/distributions/arch.py
+++ b/mkosi/distributions/arch.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
+import datetime
 import tempfile
 from collections.abc import Iterable
 from pathlib import Path
@@ -8,7 +9,7 @@ from mkosi.archive import extract_tar
 from mkosi.config import Architecture, Config
 from mkosi.context import Context
 from mkosi.curl import curl
-from mkosi.distributions import DistributionInstaller, PackageType
+from mkosi.distributions import DistributionInstaller, PackageType, join_mirror
 from mkosi.installer.pacman import Pacman, PacmanRepository
 from mkosi.log import complete_step, die
 
@@ -44,7 +45,7 @@ class Installer(DistributionInstaller):
                 curl(
                     context.config,
                     "https://archlinux.org/packages/core/any/archlinux-keyring/download",
-                    Path(d),
+                    output_dir=Path(d),
                 )
                 extract_tar(
                     next(Path(d).iterdir()),
@@ -69,9 +70,24 @@ class Installer(DistributionInstaller):
             yield PacmanRepository("core", context.config.local_mirror)
         else:
             if context.config.architecture.is_arm_variant():
-                url = f"{context.config.mirror or 'http://mirror.archlinuxarm.org'}/$arch/$repo"
+                if context.config.snapshot and not context.config.mirror:
+                    die("There is no known public mirror for snapshots of Arch Linux ARM")
+
+                mirror = context.config.mirror or "http://mirror.archlinuxarm.org"
             else:
-                url = f"{context.config.mirror or 'https://geo.mirror.pkgbuild.com'}/$repo/os/$arch"
+                if context.config.mirror:
+                    mirror = context.config.mirror
+                elif context.config.snapshot:
+                    mirror = "https://archive.archlinux.org"
+                else:
+                    mirror = "https://geo.mirror.pkgbuild.com"
+
+            if context.config.snapshot:
+                url = join_mirror(mirror, f"repos/{context.config.snapshot}/$repo/os/$arch")
+            elif context.config.architecture.is_arm_variant():
+                url = join_mirror(mirror, "$arch/$repo")
+            else:
+                url = join_mirror(mirror, "$repo/os/$arch")
 
             # Testing repositories have to go before regular ones to to take precedence.
             repos = [
@@ -107,3 +123,10 @@ class Installer(DistributionInstaller):
             die(f"Architecture {a} is not supported by Arch Linux")
 
         return a
+
+    @classmethod
+    def latest_snapshot(cls, config: Config) -> str:
+        url = join_mirror(config.mirror or "https://archive.archlinux.org", "repos/last/lastsync")
+        return datetime.datetime.fromtimestamp(int(curl(config, url)), datetime.timezone.utc).strftime(
+            "%Y/%m/%d"
+        )
--- a/mkosi/distributions/centos.py
+++ b/mkosi/distributions/centos.py
@@ -4,6 +4,7 @@ from collections.abc import Iterable
 
 from mkosi.config import Architecture, Config
 from mkosi.context import Context
+from mkosi.curl import curl
 from mkosi.distributions import (
     Distribution,
     DistributionInstaller,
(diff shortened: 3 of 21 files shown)


mkosi/installer/dnf.py
```diff
@@ -25,9 +25,16 @@ def subdir(cls, config: Config) -> Path:
...
     @classmethod
-    def package_subdirs(cls, cache: Path) -> list[Path]:
+    def package_subdirs(cls, cache: Path) -> list[tuple[Path, Path]]:
+        dirs = [p for p in cache.iterdir() if p.is_dir() and "-" in p.name and "mkosi" not in p.name]
         return [
-            p / "packages" for p in cache.iterdir() if p.is_dir() and "-" in p.name and "mkosi" not in p.name
+            (
+                d.relative_to(cache) / "packages"
+                if cache == Path("/var")
+                else Path("packages") / d.name[: d.name.rfind("-")],
```
behrmann: A comment for the logic `d.name[: d.name.rfind("-")` would be nice, since it's not readily apparent from the commit message

mkosi/__init__.py
```diff
@@ -4836,8 +4843,23 @@ def sync_repository_metadata(
 
     subdir = last.distribution.package_manager(last).subdir(last)
 
+    # Pacman and Zypper are unable to cache repository metadata for multiple mirrors at the same time in the
+    # same cache directory as they both use only use the repository id as the cache key which will trivially
```
```suggestion
    # same cache directory as they both only use the repository id as the cache key which will trivially
```

mkosi/distributions/centos.py
```diff
@@ -169,43 +218,46 @@ def repository_variants(cls, context: Context, repo: str) -> Iterable[RpmReposit
...
                     f"{repo.lower()}-source",
                     f"{url}?arch=source&repo=centos-{repo.lower()}-source-$stream",
-                    cls.gpgurls(context),
+                    gpgurls,
                     enabled=False,
                 )
 
     @classmethod
     def repositories(cls, context: Context) -> Iterable[RpmRepository]:
+        gpgurls = cls.gpgurls(context)
```
behrmann: Instead of passing them around, wouldn't it work to decorate the method with `functool.lru_cache`? This could be a much smaller change.

mkosi/distributions/fedora.py
```diff
@@ -134,34 +133,68 @@ def install(cls, context: Context) -> None:
...
+                    f"{repo.lower()}-debuginfo", f"{url}/$basearch/debug/tree", gpgurls, enabled=False
+                )
+                yield RpmRepository(f"{repo.lower()}-source", f"{url}/source/tree", gpgurls, enabled=False)
+        elif mirror:
+            if mirror == "https://kojipkgs.fedoraproject.org":
+                subdir = f"compose/{context.config.release}"
+            else:
+                subdir = "linux/"
+                subdir += "development" if context.config.release == "rawhide" else "releases"
+                subdir += "/$releasever"
```
behrmann, line 182: How about
```suggestion
                subdir = f"linux/{'development' if context.config.release == 'rawhide' else 'releases'}/$releasever"
```

mkosi/resources/man/mkosi.1.md
```diff
@@ -224,6 +226,14 @@ The following command line verbs are known:
     See the documentation for `ToolsTreeProfiles=` for a list of
     available profiles.
 
+`latest-snapshot`
+:   Output the latest available snapshot in the configured mirror.
+
+    This verb is useful to automatically bump snapshots every so often.
+    Note that this verb only outputs the latest snapshot. It's up to the
+    caller to ensure that the snapshot is written to the intended configuration
+    file.
```
behrmann, line 235: The whole thing can be formulated much shorter ;)
```suggestion
    This verb is useful to automatically bump snapshots every so often,
    which is left as an exercise to the reader.
```
(not serious)

mkosi/distributions/centos.py
daandemeyer: I don't think we can, requires inputs to be properly hasheable.

mkosi/distributions/fedora.py
daandemeyer: too long, ruff complains

mkosi/distributions/centos.py
behrmann: True, but that's something that could be achieved, I think. All inputs to `Context` are hashable (Args and Config thanks to being frozen), the only thing that makes `Context` not hashable, should then be the lowerdir list, but a hash function like this might just work:
```python
def __hash__(self) -> int:
     return hash(
         (
             self.args,
             self.config,
             self.workspace,
             self.resources,
             self.keyring_dir,
             self.metadata_dir,
             self.package_dir,
             tuple(self.lowerdirs),
             self.upperdir,
             self.workdir,
         ) 
```         
since then all the ingredients are hashable.

daandemeyer: I think this is more complex and less maintainable than just calculating the list once and passing it around.

behrmann: I don't think I agree, since it's one more thing that's extra to carry around and manage ourselves, but I also don't think it's enough of a win to drag this out because of this.

mkosi/distributions/arch.py
```diff
@@ -69,9 +70,24 @@ def repositories(cls, context: Context) -> Iterable[PacmanRepository]:
...
-                url = f"{context.config.mirror or 'http://mirror.archlinuxarm.org'}/$arch/$repo"
+                if context.config.snapshot and not context.config.mirror:
+                    die("There is no known public mirror for snapshots of Arch Linux ARM")
+
+                mirror = context.config.mirror or "http://mirror.archlinuxarm.org"
+            else:
+                if context.config.mirror:
+                    mirror = context.config.mirror
+                elif context.config.snapshot:
+                    mirror = "https://archive.archlinux.org"
```
christian-heusel, line 81: We also do have mirrors for the archive, therefore it would be good to not only hit on the main mirror:
```ini
[archive_mirrors]
america.mirror.pkgbuild.com
asia.mirror.pkgbuild.com
europe.mirror.pkgbuild.com
```

We might need to introduce something like `geo.mirror.pkgbuild.com` for the archive though 🤔 

Sadly two of them have just now been decomissioned by their sponsors 😬

daandemeyer: @christian-heusel I don't think we want to hardcode mirrors like this in mkosi, a geo mirror for the archive would be much appreciated. For now I dropped the code that makes use of the archive in CI, but I'm going to merge this with the official mirror for now. Happy to switch to a geo mirror url when that becomes available.

mkosi/distributions/ubuntu.py
```diff
@@ -75,4 +83,26 @@ def repositories(cls, context: Context, local: bool = True) -> Iterable[AptRepos
...
+            if date := startswith(line, "Date: "):
+                # %a and %b parse the abbreviated day of the week and the abbreviated month which are both
+                # locale-specific so set the locale to C explicitly to make sure we try to parse the english
+                # abbreviations used in the Release file.
+                lc = locale.setlocale(locale.LC_TIME)
+                try:
+                    locale.setlocale(locale.LC_TIME, "C")
+                    return datetime.datetime.strptime(date, r"%a, %d %b %Y %H:%M:%S %Z").strftime(
+                        r"%Y%m%dT%H%M%SZ"
+                    )
```
behrmann: These don't need to be raw strings, since they don't contain any escapes 
```suggestion
                    return datetime.datetime.strptime(date, "%a, %d %b %Y %H:%M:%S %Z").strftime(
                        "%Y%m%dT%H%M%SZ"
                    )
```
(e.g. `r"%a, %d %b %Y %H:%M:%S %Z" == "%a, %d %b %Y %H:%M:%S %Z"` returns true, while `r"%a, %d %b %Y %H:%M:%S %Z\n" == "%a, %d %b %Y %H:%M:%S %Z\n"` does not)

mkosi/distributions/arch.py
```diff
@@ -107,3 +123,10 @@ def architecture(cls, arch: Architecture) -> str:
             die(f"Architecture {a} is not supported by Arch Linux")
 
         return a
+
+    @classmethod
+    def latest_snapshot(cls, config: Config) -> str:
+        url = join_mirror(config.mirror or "https://archive.archlinux.org", "repos/last/lastsync")
+        return datetime.datetime.fromtimestamp(int(curl(config, url)), datetime.timezone.utc).strftime(
+            r"%Y/%m/%d"
```
behrmann:
```suggestion
            "%Y/%m/%d"
```

mkosi/installer/apt.py
```diff
@@ -31,6 +32,7 @@ def __str__(self) -> str:
             Suites: {self.suite}
             Components: {" ".join(self.components)}
             {"Signed-By" if self.signedby else "Trusted"}: {self.signedby or "yes"}
+            {f"Snapshot: {self.snapshot}" if self.snapshot else ""}
```
behrmann, line 35: `Snapshot` only appears for Debian in the `sources.list` man page with trixie, but with Ubuntu I can find it as far back as 22.04, so using this should be fine.

PR #3805, opened 2025-07-15
ci: add couple more arm64 builds, add ppc64le builds

bluca:


behrmann (approved):
I'm fine with this, but it  will probably make CI even slower.

bluca:
I checked and it shouldn't, as these only build images, the slow part is booting the VMs for the integration tests. If the runners suffer too much from throttling we can remove some or all again

PR #3809, opened 2025-07-17
Build UKI with signed bootloader if custom firmware vars are used

bluca:
With custom firmware we enroll our keys in db, so local UKIs can be built and there's no need to fail the build. Many distributions ship signed bootloaders, but they still don't ship UKIs.


DaanDeMeyer (changes requested):
FirmwareVariables is a runtime setting, it can not affect image builds. Please explain what you're trying to do.

mkosi/__init__.py
```diff
@@ -2133,17 +2133,22 @@ def install_uki(
     with umask(~0o700):
         boot_binary.parent.mkdir(parents=True, exist_ok=True)
 
+    needs_build = True
```
behrmann: I think this extra conditional makes this a bit harder to understand. Let me check whether my understanding is correct: You want to hit the lines from 2147 to 2170 both if the bootloader is not signed or if it is signed but the firmware variables are custom? 

I think it would be a bit easier to follow if 2147 to 2170 were moved to their own function `build_and_measure_uki` (or some such) and this would then be
```python
    if context.config.bootloader.is_signed():
        for p in (context.root / "usr/lib/modules" / kver).glob("*.efi"):
            log_step(f"Installing prebuilt UKI at {p} to {boot_binary}")
            shutil.copy2(p, boot_binary)
            break
        else:
            if context.config.bootable != ConfigFeature.enabled:
                return pcrs

            # If we are enrolling our own keys in db, we can build a custom UKI if one is not found
            if context.config.firmware_variables != Path("custom"):
                die(f"Couldn't find a signed UKI binary installed at /usr/lib/modules/{kver} in the image")

            build_and_measure_uki(…)
    else:
        build_and_measure_uki(…)
```

Alternatively, perhaps the grub part could be moved above this block, the break made a return and the following else-branches moved one indentation-level in.

bluca:
> FirmwareVariables is a runtime setting, it can not affect image builds. Please explain what you're trying to do.

It's as written in the description - there's really no distro signed uki available, so the requirement that a distro signed bootloader must come with a uki doesn't work. I am building an image with the distro signed bootloader to test it, but I want to test it with a uki too, which has to be built locally since it doesn't exist. With the custom FW variables this works just fine with this change. Without it the build fails as no uki exists.

daandemeyer:
If we really need this in mkosi (if it's just for a one off test I wouldn't bother tbh) then the `UnifiedKernelImages=` setting should be changed to a custom enum with values `unsigned`, `none`, `signed` and `auto` so that in your case you can set `Bootloader=signed` and `UnifiedKernelImages=unsigned`

bluca:
Ok, changed to enum and added parser for backward compat

bluca:
> mkosi/__init__.py:2137:110: E501 Line too long (117 > 109)

one of these days I'm going to buy you all an ultrawide monitor

bluca:
Any suggestion on how to make this work?

```
  Drill down into differing attribute unified_kernel_images:
    unified_kernel_images: 'auto' != <UnifiedKernelImage.auto: 'auto'>
```

Tried the obvious ones but no luck

behrmann:
> Any suggestion on how to make this work?

There's a big dict in `config.py::json_type_transformer`, you need to add it there, then the string from JSON should get roundtripped into an enum again.

> one of these days I'm going to buy you all an ultrawide monitor

Joke's on you, already sitting in front of one :P

bluca:
Is this good to go now? Implemented as requested

mkosi/resources/man/mkosi.1.md
```diff
@@ -1001,13 +1001,19 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
+    `unsigned`, `signed` or `auto`. Defaults to `auto`. If `unsigned` or `signed`, unified
+    kernel images
     are always used and the build will fail if any components required to
     build unified kernel images are missing. If set to `auto`, unified
     kernel images will be used if all necessary components are available.
     Otherwise Type 1 entries as defined by the Boot Loader Specification
     will be used instead. If disabled, Type 1 entries will always be used.
+    If `Bootloader=` is set to one of the signed variant, a pre-built UKI
+    will be searched and the build will fail if it cannot be found, unless
+    it is set to `unsigned`, in which case it will be built locally. This
```
behrmann: The "it" here is a bit unclear, looking at the code you mean
```suggestion
    `UnifiedKernelImages=` is set to `unsigned`, in which case the UKI will be built locally. This
```

mkosi/config.py
```diff
@@ -951,6 +958,26 @@ def config_parse_mode(value: Optional[str], old: Optional[int]) -> Optional[int]
...
+        return UnifiedKernelImage.unsigned
+
+    if value == "signed":
+        return UnifiedKernelImage.signed
+
+    # For backward compatibility, this used to be a feature flag
+    if parse_feature(value) == ConfigFeature.enabled:
+        return UnifiedKernelImage.signed
+
+    return UnifiedKernelImage.none
```
daandemeyer: Please use `config_make_enum_parser_with_boolean(UnifiedKernelImage, yes=UnifiedKernelImage.signed), no=UnifiedKernelImage.none)` instead.

mkosi/__init__.py
```diff
@@ -2133,7 +2134,10 @@ def install_uki(
     with umask(~0o700):
         boot_binary.parent.mkdir(parents=True, exist_ok=True)
 
-    if context.config.bootloader.is_signed():
+    if (
+        context.config.bootloader.is_signed()
+        and context.config.unified_kernel_images != UnifiedKernelImage.unsigned
+    ):
```
daandemeyer: We should never go into the else branch here if `unified_kernel_images` is set to `signed`. 

So in this function I'd determine the final `UnifiedKernelImage` value by translating `auto` to `signed` if the bootloader is signed and `unsigned` if the bootloader is unsigned.

If not set to `auto` then we always use the configured value, so we'll install unsigned UKI with signed bootloader or signed UKI with unsigned bootloader. Might not make sense but if the user explicitly wants it we should do it.

mkosi/__init__.py
```diff
@@ -2807,7 +2811,10 @@ def check_tools(config: Config, verb: Verb) -> None:
                 reason="build unified kernel image profiles",
                 hint=("Use ToolsTree=default to download most required tools including ukify automatically"),
             )
-        elif want_efi(config) and config.unified_kernel_images == ConfigFeature.enabled:
+        elif want_efi(config) and config.unified_kernel_images in (
+            UnifiedKernelImage.signed,
+            UnifiedKernelImage.unsigned,
+        ):
```
daandemeyer: Can you add a method `enabled()` to the enum and use that here?

mkosi/__init__.py
bluca: ok, done

behrmann:
Merging since @bluca needs it for some follow-up and all comments are addressed.

PR #3811, opened 2025-07-17
qemu: Disable hpet for x86 VMs

DaanDeMeyer:
hpet is an emulated clocksource that is generally discouraged in favor of kvm-clock or tsc for virtual machines. While mkosi's virtual machines already use kvm-clock, leaving hpet enabled causes qemu on the host to consume a non-trivial amount of cpu, so let's disable the hpet feature since we're not making use of it anyway.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -554,6 +554,9 @@ class Architecture(StrEnum):
     def supports_smm(self) -> bool:
         return self.is_x86_variant()
 
+    def supports_hpet(self) -> bool:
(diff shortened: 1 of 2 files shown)


keszybz:
My testcase was running `meson test -C build --setup=shell -i TEST-89-RESOLVED-MDNS` in the systemd repo. Without this patch I see ~9% CPU constant use with an idle guest, with the patch ~0.6%, so this clearly helps.

PR #3813, opened 2025-07-17
Change ruff check and ruff format order

behrmann:


behrmann:
No point wasting more CI on this.

PR #3815, opened 2025-07-18
Various man page fixes

DaanDeMeyer:


mkosi/resources/man/mkosi.1.md
```diff
@@ -428,9 +428,9 @@ Architecture=|x86-64
 Architecture=|arm64
 ```
 
-The `[TriggerMatch]` section can be used to indicate triggering match
-sections. These are identical to triggering conditions except they apply
-to the entire match section instead of just a single condition. As an
+The `[TriggerMatch]` section can be used to indicate triggering matches.
+These are identical to triggering conditions except they apply to the
```
behrmann: How about 
```suggestion
These are identical to triggering conditions in systemd units except they apply to the
```
so that it's easier for people to make the connection, since we talk about conditions in the context of matches as well.

behrmann (approved):
LGTM, one little point

PR #3817, opened 2025-07-20
FirmwareVariables: allow generating during image build

bluca:
The build immediately fails if FirmwareVariables=%O/somefile is used, as the config parser won't be able to find it, so it is not possible to generate it during the image build itself (e.g: mkosi.postoutput) in order to add generated keys to MOK. Set required=False.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -4178,7 +4178,7 @@ SETTINGS: list[ConfigSetting[Any]] = [
         dest="firmware_variables",
         metavar="PATH",
         section="Runtime",
-        parse=config_make_path_parser(constants=("custom", "microsoft", "microsoft-mok")),
+        parse=config_make_path_parser(constants=("custom", "microsoft", "microsoft-mok"), required=False),
         help="Set the path to the firmware variables file to use",
         compat_longs=("--qemu-firmware-variables",),
         compat_names=("QemuFirmwareVariables",),
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -768,6 +768,8 @@ def finalize_firmware_variables(
             if config.firmware_variables == Path("microsoft") or not config.firmware_variables
             else config.firmware_variables
         )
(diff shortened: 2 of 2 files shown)


mkosi/config.py
```diff
@@ -4165,7 +4165,7 @@ def parse_kernel_module_filter_regexp(p: str) -> str:
         dest="firmware_variables",
         metavar="PATH",
         section="Runtime",
-        parse=config_make_path_parser(constants=("custom", "microsoft", "microsoft-mok")),
+        parse=config_make_path_parser(constants=("custom", "microsoft", "microsoft-mok"), required=False),
```
behrmann, line 4181: Switching this of generally seems like something that could trip up people, since things that would have been caught at config parsing time, will now be caught only later. Maybe we could warn in the required case only if the path starts with `%`?

bluca: How are other things that can use specifiers handled?

daandemeyer:
Isn't this solved by making sure `FirmwareVariables=` is defined after `OutputDirectory=` is set?

If we want to make this work regardless of which order the settings are defined in, we need to add a generic form of delayed specifiers, that are only expanded once the entire config has been parsed.

bluca:
> Isn't this solved by making sure `FirmwareVariables=` is defined after `OutputDirectory=` is set?
> 
> If we want to make this work regardless of which order the settings are defined in, we need to add a generic form of delayed specifiers, that are only expanded once the entire config has been parsed.

Ordering them did not help, I had [Output] first and [Runtime] later in the same mkosi.conf and it still barfed

bluca:
ok

behrmann (approved):

behrmann:
> assuming we support this for vmspawn, otherwise ignore the vmspawn part

Yep, not the case
```
    if config.firmware_variables and config.firmware_variables != Path("microsoft"):
        die("mkosi vmspawn does not support FirmwareVariables=")
```

PR #3818, opened 2025-07-21
Pass unexpanded values to config parsers

behrmann:
This is an alternative to #3817, making the unexpanded values available to config parsers, allowing the path parser to not require paths to exist that might be generated later because they are in the output directory.


bluca (approved):
Works for me, thanks

daandemeyer:
I really dislike this, since the original issue was fixed, let's close this.

PR #3822, opened 2025-07-25
Avoid duplicate files in cpio archives

sorenisanerd:
If you have `lib/firmware/dirA/hugeblob.bin` and `lib/firmware/dirB` is a symlink to `dirA`, you would be asking cpio to archive both `lib/firmware/dirA/hugeblob.bin` and `lib/firmware/dirB/hugeblob.bin`, i.e. two copies of the same file.  This particularly affects the kernel modules initrd.

Address this problem by filtering out any files that are below a symlink. Python 3.13 introduced a `recurse_symlinks` kwarg to `Path.rglob()` which I believe achieves the same thing.


mkosi/archive.py
```diff
@@ -119,6 +119,16 @@ def make_cpio(
     else:
         files = sorted(files)
 
+    # Filters out files that are below a symlink. If you have
+    # `lib/firmware/dirA/hugeblob.bin`, and `lib/firmware/dirB` is a
+    # symlink to `dirA`, we'd end up asking cpio to archive both
+    # `lib/firmware/dirA/hugeblob.bin` and
+    # `lib/firmware/dirB/hugeblob.bin`, i.e. two copies of the same
+    # file. Once Python 3.13 becomes the minimum, we can use
+    # the `recurse_symlinks=` argument to `Path.rglob()`,
```
behrmann: Can you make this
```suggestion
    # file. 
    # TODO: Once Python 3.13 becomes the minimum, we can use
    # the `recurse_symlinks=` argument to `Path.rglob()`,
```

`TODO` comments are much easier to grep for regularly.

mkosi/util.py
```diff
@@ -269,3 +269,10 @@ def mandatory_variable(name: str) -> str:
         return os.environ[name]
     except KeyError:
         die(f"${name} must be set in the environment")
+
+
+def is_below_symlink(path: Path) -> bool:
+    for parent in path.parents:
+        if parent.is_symlink():
+            return True
+    return False
```
behrmann: Come to think of it, this can be written as
```suggestion
    return any(parent.is_symlink() for parent in path.parents)
```

mkosi/archive.py
sorenisanerd: Good call, thanks.

sorenisanerd:
With this config:
```
[Distribution]
Distribution=ubuntu
Release=noble
Repositories=main,universe

[Include]
Include=mkosi-vm
```

These are the generated artifacts without this change:

```
soren@zinc:~/vm/before$ du -a .
0	./image
463408	./image.efi
14656	./image.vmlinuz
448688	./image.initrd
1812156	./image.raw
2738912	.
```

And with this patch:
```
soren@zinc:~/vm/after$ du -a .
0	./image
334576	./image.efi
14656	./image.vmlinuz
319852	./image.initrd
1683324	./image.raw
2352412	.
```

128MB reduction in size of the initrd.

sorenisanerd:
The opensuse failures are affecting other PR's, too, so I think they're unrelated to this change.

behrmann:
> 128MB reduction in size of the initrd.

Very good, but I wonder: If before cpio flattened symlinked paths into two separate copies, is the symlink now only an empty directory? Do we guarantee that no paths are broken?

sorenisanerd:
The only situation I can think of where a file might have been included previously but not with this patch, is if you have something like:

```
/blobs/fileA
/lib/firmware/extra_blobs -> /blobs
```

`make_cpio(src='/lib/firmware', ...)` would rummage through `/lib/firmware` and find `/lib/firmware/extra_blobs` as well as `/lib/firmware/extra_blobs/fileA`. The former would be archived as a symlink, the latter as a file. When unpacking it, it'll first create `/lib/firmware/extra_blobs` as a symlink to `/blobs` and then unpack `fileA` to `/lib/firmware/extra_blobs/fileA` which in reality is `/blobs/fileA` due to the symlink.

`tar c [...] /lib/firmware` will not traverse `/lib/firmware/extra_blobs`, nor will `find /lib/firmware -print0 | cpio -0 -o`. We're really working around the fact that `Path.rglob` descends into "directories" that are actually symlinks.

septatrix:
Hm based on the documentation I would have expected this to already be the case? I.e. it states that symlinks are not followed (for the `**` parts) - the new kwarg would also not change that, it rather does the opposite.

sorenisanerd:
I think you're right. I guess the dupes may be passed to `make_cpio` from `gen_required_kernel_modules`. I'll take a look tomorrow. Maybe this isn't the right way to fix it, even though the output ends up being the same.

sorenisanerd:
@septatrix You were absolutely right. During the kmod dependency resolution, we collect a list of all firmware files that are required by any included module. Example:
```
$ modinfo nouveau | grep 'tu10[24]/gsp'
firmware:       nvidia/tu104/gsp/gsp-535.113.01.bin
firmware:       nvidia/tu104/gsp/bootloader-535.113.01.bin
firmware:       nvidia/tu104/gsp/booter_unload-535.113.01.bin
firmware:       nvidia/tu104/gsp/booter_load-535.113.01.bin
firmware:       nvidia/tu102/gsp/gsp-535.113.01.bin
firmware:       nvidia/tu102/gsp/bootloader-535.113.01.bin
firmware:       nvidia/tu102/gsp/booter_unload-535.113.01.bin
firmware:       nvidia/tu102/gsp/booter_load-535.113.01.bin
$ ls -ld /usr/lib/firmware/nvidia/tu104/gsp
lrwxrwxrwx 1 root root 12 Mar 14 06:30 /usr/lib/firmware/nvidia/tu104/gsp -> ../tu102/gsp
```

End result is `tu102/gsp/*` getting included twice.


The solution remains the same, but I've removed the references to `Path.rglob()` as the culprit.

daandemeyer:
I think this probably should be fixed in gen_required_kernel_modules() rather than in make_cpio(). In gen_required_kernel_modules(), for all firmware we're going to add, we should remove all firmware below a directory symlink beneath /usr/lib/firmware and replace it with the directory symlink instead.

So traverse all parent paths of each firmware up until /usr/lib/firmware and if one if a parent path is a symlink, remove the firmware and add the parent path instead to the list of firmware to add.

septatrix:
And if you rebase on latest main you can drop the formatting change (see 9de34024adf58e0475c532f34bc8cd3f217218eb)

behrmann:
Thanks for the contribution @sorenisanerd! 

Replaced by #3872.

PR #3823, opened 2025-07-25
Include {sys,conf}ext info in metadata file

aplanas:


bluca (approved):
nice, thanks

mkosi/manifest.py
```diff
@@ -214,6 +248,28 @@ def record_pkg_packages(self) -> None:
...
+            self.ext = ExtManifest(
+                prefix,
+                extrelease["ID"],
+                extrelease.get(f"{prefix}_LEVEL"),
+                extrelease.get("VERSION_ID"),
+                extrelease.get(f"{prefix}_ID"),
+                extrelease.get(f"{prefix}_VERSION_ID"),
+                extrelease[f"{prefix}_SCOPE"],
+                extrelease["ARCHITECTURE"],
+            )
```
behrmann: This doesn't make sense to me. Why should anything be written if the extension release file not exist? Also `extrelease["ARCHITECTURE"]` will just crash when the file is empty.

mkosi/manifest.py
```diff
@@ -63,9 +64,42 @@ def report(self) -> str:
         return t
 
 
+@dataclasses.dataclass
+class ExtManifest:
```
behrmann: Let's not have this be too cryptic
```suggestion
class ExtensionManifest:
```

mkosi/manifest.py
```diff
@@ -63,9 +64,42 @@ def report(self) -> str:
...
+    """Sysext or Confext metadata"""
+
+    prefix: str
+    id: str
+    level: Optional[str]
+    version_id: Optional[str]
+    ext_id: Optional[str]
+    ext_version_id: Optional[str]
+    scope: str
+    architecture: str
```
behrmann: Nit: Could you sort the optional ones to the end?

mkosi/manifest.py
```diff
@@ -63,9 +64,42 @@ def report(self) -> str:
         return t
 
 
+@dataclasses.dataclass
+class ExtManifest:
+    """Sysext or Confext metadata"""
+
+    prefix: str
```
behrmann: One could move this to have the two be separate types, but I think that's overkill, but we can give more information here
```suggestion
    prefix: Union[Literal["CONFEXT"], Literal["SYSEXT"]]
```
Also, why not name this `type`?

mkosi/manifest.py
behrmann: Also, this will have already gone through `configure_extension_release`, will it not? Maybe we should just split that function in one doing the reading into a dictionary and one writing out the file, since this is to a large extent the same function only producing this extra output instead. I assume it was taken from there, that would then also explain the `prefix` name.

aplanas: > This doesn't make sense to me.

Yep. Removed the `else` part. If it is a SYSEXT the file will be there

> Maybe we should just split that function in one doing the reading into a dictionary and one writing out the file,

Where do you think that I can put the dict? In `context`?

mkosi/manifest.py
aplanas: Yes. Done.

mkosi/manifest.py
aplanas: Done, but I am not sure that I am happy. This order follows the one in the file, and the one in `configure_extension_release`, this consistency have some value where mapping the code.

mkosi/manifest.py
aplanas: I found that `Union["CONFEXT", "SYSEXT"]` also works. Done.

> Also, why not name this `type`?

`type` should be `prefix.lower()`, this is always use as a prefix: `f"{prefix}_XXX"`.  I can change it tho.

mkosi/manifest.py
behrmann: My idea was to just split out reading the extension release from `configure_extension_release`, but the function mixes the data with that from os-release. But looking at it now, currently it passes everything it reads from it initally through and I think that would be the sanest thing here as well, i.e. dropping the extra dataclass and just reading the dictionary from the extension release file and putting that in the manifest.

aplanas: Should be done. I added this change as a new commit, that I will squash into the original if this is the direction that we want to follow, but as is now can be useful to compare.

mkosi/manifest.py
```diff
@@ -214,6 +217,19 @@ def record_pkg_packages(self) -> None:
                 self.source_packages[source] = source_package
             source_package.add(package)
 
+    def record_extension_release(self) -> None:
+        if self.context.config.output_format not in (OutputFormat.sysext, OutputFormat.confext):
+            return
+
+        with complete_step("Recording sysext / confext information in manifest…"):
+            self.extension_type: Optional[Literal["sysext", "confext"]] = (
```
behrmann: This annotation is not correct, since it cannot be `None`, but it's also not necessary, neither mypy nor pyright complain for me locally
```suggestion
            self.extension_type = (
```

mkosi/manifest.py
```diff
@@ -68,6 +69,8 @@ class Manifest:
     context: Context
     packages: list[PackageManifest] = dataclasses.field(default_factory=list)
     source_packages: dict[str, SourcePackageManifest] = dataclasses.field(default_factory=dict)
+    extension: Optional[dict[str, str]] = None
+    extension_type: Optional[Literal["sysext", "confext"]] = None
```
behrmann: I don't think we should have the type separate from the thing itself, since now it's two independent things that need to be kept together and the information is also redundant, since I can tell from the extension release whether it is a sysext or confext, how about
```suggestion
    extension: list[dict[str, str]] = []
```
and just adding the dictionary there? This way, if we were to add multiple extensions in one go (not that we can do this at the moment) or would merge manifests for some reason, this would still fit under the same key.

behrmann:
@aplanas Could you rebase and drop the second commit?

mkosi/manifest.py
aplanas: `extension_type`, or something equivalent (as was in the dataclass) is required, as this information is not available in the moment of writing the dictionary, only in collect stage (here)

Another option is to do `extension: dict[str, dict[str, str]]` and store the type and the content together.

For multiple extensions, how will they look like?  For example, now the json generated is like this:

```json
{
  "manifest_version": 1,
  "config": {
    "name": "image",
    "distribution": "opensuse",
    "architecture": "x86-64",
    "version": "0.0",
    "release": "tumbleweed"
  },
  "packages": [
    {
      "type": "rpm",
      "name": "babeltrace",
      "version": "1.5.8-4.2",
      "architecture": "x86_64"
    },
  ],
  "sysext": {
    "ID": "opensuse-microos",
    "SYSEXT_LEVEL": "glibc-2.41",
    "VERSION_ID": "20250725",
    "SYSEXT_VERSION_ID": "0.0",
    "SYSEXT_SCOPE": "system",
    "ARCHITECTURE": "x86-64"
  }
}
```

How multiple "sysext" can be expressed?

aplanas:
> @aplanas Could you rebase and drop the second commit?

Dropped, but seems that it is still not in main

mkosi/manifest.py
behrmann: > extension_type, or something equivalent (as was in the dataclass) is required, as this information is not available in the moment of writing the dictionary, only in collect stage (here)

I'm not sure I understand this. From
```{
    "ID": "opensuse-microos",
    "SYSEXT_LEVEL": "glibc-2.41",
    "VERSION_ID": "20250725",
    "SYSEXT_VERSION_ID": "0.0",
    "SYSEXT_SCOPE": "system",
    "ARCHITECTURE": "x86-64"
  }
```
I can read of, that this is a sysext. so if the key in the manifest just were `extensions` and I'd get a list of dictionaries like that, I'd still know which ones are sysexts and which ones are confexts (or something else if that ever came to pass)?

thkukuk: No, from the keys you cannot determine if this is a sysext or a confext. The example here is an exception.

thkukuk:
```
{ "ID": "_any" }
```
would be valid but does not give you any pointer about what this is.

behrmann: You're absolutely right. Reading the `os-release` man page again, none of these keys I thought were mandatory, are—there's even an example (example 2).

> How multiple "sysext" can be expressed?

Instead of a dict, a list[tuple[str, dict[str, str]]], where the first element of the tuple encodes the type and the second the extension release dictionary.

aplanas: >> How multiple "sysext" can be expressed?
> Instead of a dict

But what is the expectation in the manifest file. Something like this?

```json
...
extensions: [
  "sysext": {
  },
  "confext": {
  }
]
```

For me this does not makes much sense, each extension image contains a single sysext or confext, not both. I am not sure that both together resolve anything. If the use case is multiple sysexts, that does not makes also sense to me, there is no decided ID that differentiate those sysexts (and this PR should not address this).

behrmann: A list of tuples in Python maps to a list of list in JSON. so it would be
```json
extensions: [
    [
        "sysext",
        {
            …
        }
    ] 
]
```
Not particularly pretty, but a straightforward map, easily mergeable, easy to iterate over, not to deep and it keeps the type information (sysext or confext) together with the data, without having to put them in the data with some key that might be clobbered (though unlikely) with what's in the extension release.

thkukuk:
I tested the PR and there is one problem:
it is common practice to modify `/buildroot/usr/lib/extension-release.d/extension-release.*` from `mkosi.finalize`.
As example, the sysext images for RKE2 and k3s, both kubernetes distributions shipped as single, self-contained, static go binary without any dependencies to the OS have as content (adjusted in `mkosi.finalize`), contain:
```
ID=_any
SYSEXT_VERSION_ID=35.12
SYSEXT_SCOPE=system
ARCHITECTURE=x86-64
EXTENSION_RELOAD_MANAGER=1
```
so that they can be installed on any x86-64 distribution.
But the sysext entry in the manifest has `"ID": "opensuse-microos"`, so seems to be generated before mkosi.finalize.

mkosi/manifest.py
thkukuk: @behrmann The name of the `/usr/lib/extension-release.d/extension-release.*` file must be identical to the image name, else `systemd-sysext` will refuse to mount it.
So with this restriction, it will never be possible that one image contains multiple extensions.
Does it really make sense to have this "extensions" construct if there can always be only one entry?

aplanas: Should be done. New output:

```json
{
  "manifest_version": 1,
  "config": {
    "name": "image",
    "distribution": "opensuse",
    "architecture": "x86-64",
    "version": "0.0",
    "release": "tumbleweed"
  },
  "packages": [
    {
      "type": "rpm",
      "name": "babeltrace",
      "version": "1.5.8-4.2",
      "architecture": "x86_64"
    }
  ],
  "extensions": [
    [
      "sysext",
      {
        "ID": "opensuse-microos",
        "SYSEXT_LEVEL": "glibc-2.41",
        "VERSION_ID": "20250727",
        "SYSEXT_VERSION_ID": "0.0",
        "SYSEXT_SCOPE": "system",
        "ARCHITECTURE": "x86-64"
      }
    ]
  ]
}
```

behrmann: > @behrmann The name of the `/usr/lib/extension-release.d/extension-release.*` file must be identical to the image name, else `systemd-sysext` will refuse to mount it.
> So with this restriction, it will never be possible that one image contains multiple extensions.
> Does it really make sense to have this "extensions" construct if there can always be only one entry?

You're right that one image will only ever contain a single extension, the direction I am coming from with this:
1. I don't think adding another dedicated object (the extension release dataclass) is necessary, when we just want to save a dictionary with varying sets of keys and values
2. Unfortunately, it's not just a set of keys and values, but also a type, since the extension release is not self-documenting in what kind of extension it is (that ship has probably sailed), but I don't think it's a good idea to save both of those pieces of information in two separate attributes in the manifest.

From this there's a few obvious shapes to record this data together:
- `tuple[str, dict[str, str]`, e.g `("sysext", {"ID": …})`, or 
- `dict[str, str | dict[str, str]]`, e.g. `{"type": "sysext", "data": {"ID": …}}`, or 
- `dict[str, dict[str, str]]`, e.g., as @aplanas had, `{"sysext": {"ID: …}}`.

The solution @aplanas came up with, is totally fine, but since we can have an extension or not, the key in the manifest must be either of these or null. So, when using the manifest we would always need to check whether the key is set. 

My thinking in such cases is, that it's probably likely, that one will need to work with multiple manifests at some point, e.g. when having built multiple sysexts from the same base. I usually find having a list (or other container) of things nicer than having multiple "one or nothing" of a thing, since they combine more easily. 

The third point in the list is already a container, but it doesn't combine well, since it's a 1:1 mapping. So the in all cases we could put either of these in a list. My preference is for the first, since it's flat and easy.

Does that reasoning make sense?

behrmann:
> so seems to be generated before mkosi.finalize.

That's true, so that hunk should probably move some 20 lines down.

thkukuk:
> > so seems to be generated before mkosi.finalize.
> 
> That's true, so that hunk should probably move some 20 lines down.

I learned meanwhile that you can do this changes in mkosi.postinst, too, since mkosi will merge the files.

mkosi/manifest.py
aplanas: > The solution @aplanas came up with, is totally fine, but since we can have an extension or not, the key in the manifest must be either of these or null. So, when using the manifest we would always need to check whether the key is set.

For completion, the initial solutions (including the one with a dataclass) was more `dict[str, str]`, so you can have 0 or 1 extension. To check if a manifest belong to a sysext was a `"sysext" in manifest` check, that IMHO is simpler.  Now you need something like `any((e == "sysext" for e,_ in manifest["extensions"])) if "extensions" in manifest`.

In the current form "extensions" will not appear in the manifest, should I make it always appear and be an empty list if there is no extension?

aplanas: > My thinking in such cases is, that it's probably likely, that one will need to work with multiple manifests at some point, e.g. when having built multiple sysexts from the same base. I usually find having a list (or other container) of things nicer than having multiple "one or nothing" of a thing, since they combine more easily.

the `extension-release.d` name suggest multiple files, but the current code only recognize and extend one, based on version (`extension-release.{self.context.config.output}`).  The decision now is what to do with the other files that the user injected, and if this one that match the config.output suffix should be special and what that means.

My worry is that if one of those decision change the schema.

aplanas:
@behrmann something else that needs to be addressed?

mkosi/__init__.py
```diff
@@ -4047,6 +4047,9 @@ def build_image(context: Context) -> None:
         configure_ssh(context)
         configure_clock(context)
 
+        if manifest:
+            manifest.record_extension_release()
```
behrmann, line 4051: I think the point that extension release might be modified in a finalize script is valid. At the moment I don't see what would speak against moving after `run_finalize_scripts`. Do you see anything?

mkosi/manifest.py
```diff
@@ -238,6 +251,12 @@ def as_dict(self) -> dict[str, Any]:
             "packages": [package.as_dict() for package in self.packages],
         }
 
+        # Optionally, describe the SYSEXT / CONFEXT metadata
+        if self.extensions:
+            d["extensions"] = self.extensions
```
behrmann: I think it would be fine to always add this key, in the worst case it's an empty dict. Makes it nice to assume it's always in there. 

Either way we should probably bump the manifest version.

mkosi/__init__.py
aplanas: Two notes:

* The expected way to interact with sysext files is the user creating then before they are collected. This is how it is working now
* If we move the record_extension_release outside the image creation, then we will have an inconsistency about when the package information is collected.  If mkosi.finalize can make changes for the extension information, then it can add or remove package

For me seems more fit mkosi.postinst, but if mkosi.finalize is a valid place too, then we need to move both collectors. That would be OK?

mkosi/manifest.py
aplanas: Done, also bumped to "2" the manifest version.

mkosi/manifest.py
```diff
@@ -229,15 +242,20 @@ def as_dict(self) -> dict[str, Any]:
...
-            "manifest_version": 1,
+            "manifest_version": 2,
             # Describe the image itself.
             "config": config,
             # Describe the image content in terms of packages.
             "packages": [package.as_dict() for package in self.packages],
         }
 
+        # Describe the SYSEXT / CONFEXT metadata
+        d["extensions"] = self.extensions
```
behrmann: this can be moved into the above dictionary now.

aplanas: Ouch, of course. Done.

behrmann:
LGTM. One last thing from my side: Could you fix the typo in the commit message ("Inlcude" → "Include")? :)

aplanas:
Fixed the commit message

aplanas:
~~@behrmann I added an extra commit to discuss the topic of late call of metadata recollection~~

Dropped, as the package information cannot be collected that late.

mkosi/manifest.py
```diff
@@ -214,6 +216,17 @@ def record_pkg_packages(self) -> None:
                 self.source_packages[source] = source_package
             source_package.add(package)
 
+    def record_extension_release(self) -> None:
+        if self.context.config.output_format not in (OutputFormat.sysext, OutputFormat.confext):
+            return
+
+        with complete_step("Recording sysext / confext information in manifest…"):
```
daandemeyer: No need for `sysext / confext` here, you already calculate which one just below, just use that in the step message instead of listing both.

mkosi/manifest.py
```diff
@@ -214,6 +216,17 @@ def record_pkg_packages(self) -> None:
...
+            type_ = "sysext" if self.context.config.output_format == OutputFormat.sysext else "confext"
```
daandemeyer: No need to use `type_`, you can just use `type`, it doesn't matter that it overrides a builtin.

mkosi/manifest.py
```diff
@@ -214,6 +216,17 @@ def record_pkg_packages(self) -> None:
...
+            d = "usr/lib" if self.context.config.output_format == OutputFormat.sysext else "etc"
+            p = self.context.root / d / f"extension-release.d/extension-release.{self.context.config.output}"
+
+            self.extensions.append((type_, read_env_file(p)))
```
daandemeyer: Please use `+= [...]` instead of `append()` like we do everywhere else.

mkosi/manifest.py
```diff
@@ -231,11 +244,13 @@ def as_dict(self) -> dict[str, Any]:
 
         return {
             # Bump this when incompatible changes are made to the manifest format.
-            "manifest_version": 1,
+            "manifest_version": 2,
```
daandemeyer: This is not an incompatible change to the manifest format so no need to bump the manifest version. Users simply need to check if the extensions field is in the manifest or not.

DaanDeMeyer (changes requested):

mkosi/manifest.py
aplanas: Done

mkosi/manifest.py
aplanas: Very ugly. But done.

mkosi/manifest.py
aplanas: Yes. Done.

mkosi/manifest.py
aplanas: Makes sense. Done.

behrmann: I think we should bump it nevertheless, because for prospective users of this, it us much easier to reason about "version 1 has these keys", "version 2 has the keys from v1 and this new key" compared to "version 1 has these keys and possibly any other from this set". We should make it easy to consume this. 

Assuming we wouldn't bump it on adding a new key, we would need to bump the version number anyway if we change any of the optional keys. I think it's therefore easier to also bump it on adding one and making a better guarantee about which keys exist.

mkosi/manifest.py
behrmann: This is still missing?

mkosi/manifest.py
aplanas: If I make "extensions" key again optional (now will always appear), I think @DaanDeMeyer's argument remains valid. But it is true that now this key will appears always, and this can be interpreted as an incompatible change.

Any preference?

mkosi/manifest.py
aplanas: Should be there now .`with complete_step(f"Recording {type} information in manifest…"):

mkosi/manifest.py
behrmann: What I meant by optional was in the context of the versioning of keys, i.e. the key can be part of a v1 manifest or not, not the Python type.

thkukuk: If you bump the manifest version because you add new, optional fields, every tool parsing the manifest and checking the manifest number needs to get adjusted only for the version check,  not for anything else. I don't think that's a good idea.
I think the manifest version should be handled like sonames+symbol versions of shared libraries: don't change the number if only new keys got added but nothing changed incompatible (so no keys removed, no meaning of a key changed). Tools should be able to scope and ignore new json entries, at least I hope.
If json parsers really suddenly fail if there is a new key, then we need to bump the manifest version.

behrmann: For me the argument is not about the parser failing, parsing the JSON will most likely never fail if new keys are added. 

For me the argument is about making the data easy to use. Two points on this:
1. The first one is a bit squishy, it maybe comes down to to dynamic vs static typing. Having an open-ended number of keys in one version manifest number (i.e. not bumping the number when adding keys) pushes the logic needed to get data out of the manifest to the user ("Is the data I want in there and does it have the correct shape?"). 
2. Adding keys can be an incompatible change depending on the usage, e.g. having a Python function and passing in the whole dictionary, `foo(**manifest)`), will break if `foo` is not equipped to handle arbitrary `kwargs`.  

I think it is more practical to bump the number on any change, since it gives a stronger guarantee on what the shape of the data is. Users are still free to ignore the version number and duck type the heck out of the manifest, but tracking the changes with the version number also allows for the use case of not having to do that.

aplanas:
Some final notes for the reviewers.  The current patch is:

* Keeping the metadata version to `v1`
* Adding a new 'extensions: []` empty array in the metadata for non sysext/confext images
* If the image is is a sysext(/confext), the array will contain a single tuple: `extensions: [["sysext", {"ID": "opensuse-microos", "SYSEXT_LEVEL": "glibc-2.41", ...}]]`
* Even if the data type `extensions` support multiple sysext/confext (is an array of tuples), the part that collect the code only support the current situation of a single extension file

thkukuk:
> * Even if the data type `extensions` support multiple sysext/confext (is an array of tuples), the part that collect the code only support the current situation of a single extension file

Even if the code would support collecting several extension files: the current design of sysext images and `systemd-sysext` makes it impossible to combine several sysext and/or confext images in one image. And except for one sysext with one confext image this doesn't even make any sense. To support this, the whole spec of sysext/confext and of systemd-sysext needs a redesign, I doubt that this will ever happen.

aplanas:
ping? any decision about manifest bumping?

mkosi/manifest.py
daandemeyer: > Adding keys can be an incompatible change depending on the usage, e.g. having a Python function and passing in the whole dictionary, foo(**manifest)), will break if foo is not equipped to handle arbitrary kwargs.

If users are doing this I very much doubt they're checking the manifest version in the first place.

Overall, I think we're spending wayyyy too much time on something that isn't really all that important when it comes down to it. Whatever way we go, I doubt it's going to make a substantial difference for users. If we don't bump they'll check for the key, if we do bump they might still check for the key or check for the version (assuming they want to support multiple manifest versions in the first place. I doubt anyone is going to complain regardless of the way we go.

I'm just going to merge this PR since I think merging this sooner rather than later allows everyone to focus on more important stuff than this.

daandemeyer:
> Even if the code would support collecting several extension files: the current design of sysext images and systemd-sysext makes it impossible to combine several sysext and/or confext images in one image. And except for one sysext with one confext image this doesn't even make any sense. To support this, the whole spec of sysext/confext and of systemd-sysext needs a redesign, I doubt that this will ever happen.

I agree, let's keep things simple and just support one extension metadata file, not a list of them. If we ever end up having to support more than one of these bumping the manifest version to make that possible is going to be the least of our worries.

aplanas:
@DaanDeMeyer I am back to a single "extension":

* Keeping the metadata version to v1
* Adding a new `extension: {}` empty dict in the metadata for non sysext/confext images
* If the image is is a sysext(/confext), the array will contain a single extension dict: `{"sysext", {"ID": "opensuse-microos", "SYSEXT_LEVEL": "glibc-2.41", ...}}`

behrmann (approved):
Let's get this off the ground. 

I still think we should bump the manifest version, since we're changing the schema, though.

mkosi/manifest.py
```diff
@@ -216,9 +218,20 @@ def record_pkg_packages(self) -> None:
...
+    def record_extension_release(self) -> None:
+        if self.context.config.output_format not in (OutputFormat.sysext, OutputFormat.confext):
+            return
+
+        type = "sysext" if self.context.config.output_format == OutputFormat.sysext else "confext"
+        with complete_step(f"Recording {type} information in manifest…"):
+            d = "usr/lib" if self.context.config.output_format == OutputFormat.sysext else "etc"
+            p = self.context.root / d / f"extension-release.d/extension-release.{self.context.config.output}"
+
+            self.extension = {type: read_env_file(p)}
```
daandemeyer: So we can already get the type from the `config` field in the manifest, so I don't think we need to duplicate it here. so I'd get rid of one level of dict and just have `extension` be a dict containing the metadata directly without nesting it in the type.

aplanas: Done. But I added a new field `extension["type"]` that contains the type (`sysext`, `confext`).  The nested key was used as a discriminant before.

behrmann: I understood Daan's comment as taking the type of the extension image from the value of `config` in the top-level of the manifest. This would be strictly different from my earlier assumption that the type of the extension image can be gleaned from the contents of the file, which we established doesn't work. When taking this information from `config`, adding it to the dict should be unnecessary and I really don't think we should changing that information and risk clobbering a key.

aplanas: We cannot deduce the type checking `config`. For example, this comes from one of the sysext:

```
{
  "manifest_version": 1,
  "config": {
    "name": "strace",
    "distribution": "opensuse",
    "architecture": "x86-64",
    "version": "0.0",
    "release": "tumbleweed"
  },
```

There is no entry today that that describe content type, not even optional: https://github.com/systemd/mkosi/blob/main/mkosi/manifest.py#L224.  `version` and `release` are the only two optional keys in `config`

Should I add the `type` in config? Maybe with three values: `sysconf`, `extconf`, `image`?

behrmann: > Should I add the type in config? Maybe with three values: sysconf, extconf, image? 

I'd be in favour of that.

aplanas: Done then. `config["type"]` will contain the type (`sysconf || extconf || image`)

daandemeyer: Ah sorry I got confused, I thought `config` already encoded the entire `Config` object into the manifest. I think that's the way to go actually. Just set `config` to `context.config.to_dict()`.

Then you can the `output_format` field in `config` as the discriminator.

behrmann: I think the way this is, is because the manifest predates the JSON serialization code, but it does have the upside of this part of the manifest being stable. Our config dictionary is not stable and changes whenever we change anything about the config. 

Switching this to the config dictionary has the upside of including all information, but the downside renaming all keys (even the existing corresponding ones, since they are lower case). I think having the keys in config reflect the entire config is more useful.

Not wanting to beat a dead horse, though, this should probably warrant bumping the manifest version.

daandemeyer: In that case, let's rename `type` to `output_format` for now to at least match the name in `Config`

aplanas: Yes, renamed.

mkosi/manifest.py
```diff
@@ -231,13 +243,22 @@ def as_dict(self) -> dict[str, Any]:
         if self.context.config.release is not None:
             config["release"] = self.context.config.release
 
+        if self.context.config.output_format == OutputFormat.sysext:
+            config["output_format"] = "sysext"
+        elif self.context.config.output_format == OutputFormat.confext:
+            config["output_format"] = "confext"
+        else:
+            config["output_format"] = "image"
```
daandemeyer: Please simplify this to

```
config["output_format"] = self.context.config.output_format.value
```

DaanDeMeyer (changes requested):

mkosi/manifest.py
aplanas: I moved it into the config dict definition itself.

Done

mkosi/manifest.py
```diff
@@ -238,6 +250,10 @@ def as_dict(self) -> dict[str, Any]:
             "config": config,
             # Describe the image content in terms of packages.
             "packages": [package.as_dict() for package in self.packages],
+            # Describe the image type (sysext, confext, image).
+            "output_format": self.context.config.output_format.value,
```
daandemeyer: The comment is outdated, I would just drop it, and please put this in `config`

aplanas: Ouch : ( I am sorry, I am a bit in automatic mode.

Moved, and instead of `.value` I used `str()` as the rest of the `StrEnum`

DaanDeMeyer (approved):

aplanas:
Thank @behrmann and @DaanDeMeyer for the patience and the reviews.

behrmann:
Thanks for the contribution and thank you your the patience! This took quite a bit longer than it usually takes us, because of holidays and other things keeping us busy. Sorry for that.

PR #3824, opened 2025-07-25
Run finalize scripts after installing kernel

sorenisanerd:
I needed a way to adjust the GRUB config, but there are no hooks between writing the bootloader config and building the disk image.

I considered adding a new hook for it, but I could not find any use of the finalize hook in the wild that wouldn't work just as well at this slightly later point in the build process.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -4048,13 +4048,14 @@ def build_image(context: Context) -> None:
 
     clean_package_manager_metadata(context)
     remove_files(context)
-    run_finalize_scripts(context)
 
     rmtree(context.root / "work")
 
     normalize_mtime(context.root, context.config.source_date_epoch)
     partitions = make_disk(context, skip=("esp", "xbootldr"), tabs=True, msg="Generating disk image")
     install_kernel(context, partitions)
+    run_finalize_scripts(context)
+
     normalize_mtime(context.root, context.config.source_date_epoch, directory=Path("boot"))
     normalize_mtime(context.root, context.config.source_date_epoch, directory=Path("efi"))
     partitions = make_disk(context, msg="Formatting ESP/XBOOTLDR partitions")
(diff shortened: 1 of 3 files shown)


behrmann:
> I considered adding a new hook for it, but I could not find any use of the finalize hook in the wild that wouldn't work just as well at this slightly later point in the build process.

I think the change as is breaks reproducibility of images, because the adjustment of mtimes is before the finalize script runs.

> I needed a way to adjust the GRUB config, but there are no hooks between writing the bootloader config and building the disk image.

Can you describe what this addresses? Otherwise it is hard to judge this change.

sorenisanerd:
I'm creating a custom grub config. I need to do so after the kernel is installed (which is after `mkosi.finalize`), but before the disk image is finalized.

I see what you're saying about mtimes. I only touch `/boot` and its `normalize_mtime` is called after this hook, so I conveniently avoid it. I just now realize that the disk image actually already has been created with /boot and /efi deferred, so changes outside /boot and /efi won't even make it into the image. That obviates the mtimes problem, but if you expect to make changes to the filesystem, I can see you're going to be frustrated. I've only ever used the `mkosi.finalize` step for boot loader configuration and building SBOMs.

Ideas for boot loader customization? It has to be after verity stuff has been produced and/or the kernel has been installed, but before /boot is embedded in the disk image. Do we need a new hook? Ugh.

behrmann:
> I'm creating a custom grub config. I need to do so after the kernel is installed (which is after mkosi.finalize), but before the disk image is finalized.

You mentioned that in the original post already, but could you be a bit more specific in what sense and what information you need at this point? This would make it easier to get an idea how to fit this in. I for one am wary of accidentally breaking some use by moving the point at which the finalize script runs around.

daandemeyer:
@sorenisanerd Can you explain your use case more? What exactly do you need to customize? Why does it have to happen after the kernel is installed and the verity stuff is finalized?

daandemeyer:
Closing due to lack of response

PR #3828, opened 2025-07-28
env: export $EFI_ARCHITECTURE in hook scripts on EFI arches

bluca:
Saves from having to do the translations manually a bunch of times

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -609,6 +609,9 @@ def run_configure_scripts(config: Config) -> Config:
         MKOSI_DEBUG=one_zero(ARG_DEBUG.get()),
     )
 
+    if config.architecture.to_efi() is not None:
+        env["EFI_ARCHITECTURE"] = str(config.architecture.to_efi())
+
     if config.profiles:
         env["PROFILES"] = " ".join(config.profiles)
 
@@ -655,6 +658,9 @@ def run_sync_scripts(config: Config) -> None:
         CACHED=one_zero(have_cache(config)),
         MKOSI_DEBUG=one_zero(ARG_DEBUG.get()),
     )
+
+    if config.architecture.to_efi() is not None:
+        env["EFI_ARCHITECTURE"] = str(config.architecture.to_efi())
 
     if config.profiles:
         env["PROFILES"] = " ".join(config.profiles)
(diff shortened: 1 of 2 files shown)


mkosi/__init__.py
```diff
@@ -609,6 +609,9 @@ def run_configure_scripts(config: Config) -> Config:
         MKOSI_DEBUG=one_zero(ARG_DEBUG.get()),
     )
 
+    if config.architecture.to_efi() is not None:
+        env["EFI_ARCHITECTURE"] = config.architecture.to_efi()
```
behrmann: Maybe we can eschew the fine difference between an empty and an unset variable and just do
```suggestion
    env["EFI_ARCHITECTURE"] = config.architecture.to_efi() or ""
```
and always define it. It's a a lot shorter and we can move it to the above definition of `env`.

bluca: I'd rather have it undefined instead of empty for non-efi arches

behrmann: Ok, my opinion is not too strong, in pure shell it's only a difference in bash which has `-v` for `[[`, in e.g. dash, this could not be told apart without `set -u` and in Python both `None` and `""` are falsey, so what's the upside of not having this defined? Maybe I'm missing something obvious.

bluca: It seems just clearer, an empty strings looks like it's unknown or something went wrong when looking it up

behrmann: Maybe my brain is just too pickled in shell, but I'd prefer the constant environment. Nevertheless, it's more in line with us defining `PROFILE` only if we have some.

mkosi/__init__.py
```diff
@@ -609,6 +609,9 @@ def run_configure_scripts(config: Config) -> Config:
         MKOSI_DEBUG=one_zero(ARG_DEBUG.get()),
     )
 
+    if config.architecture.to_efi() is not None:
+        env["EFI_ARCHITECTURE"] = str(config.architecture.to_efi())
```
behrmann, line 613: Just for posterity: This is a rake that's only sidestepped by the `is not None` check, since `str(None)` is equal to `"None"`.

behrmann (approved):

Issue #3827, opened 2025-07-27
add support for dtbauto for embedding multiple dtbs in UKI


craftyguy:
Ukify can embed a given list of dtbs into .dtbauto sections, and it would be helpful if mkosi exposed this functionality. See: https://github.com/systemd/systemd/pull/34158

craftyguy:
FYI I started working on a patch for this

PR #3832, opened 2025-07-28
Add devicetree-auto support for UKI

craftyguy:
This renames Devicetree --> Devicetrees, and adds support for listing
globs and files as currently supported for KernelModule* options.
When multiple devicetrees are found, they are added to dtbauto sections
when building a UKI.
Multiple dtbs are not supported for type 1 booting.

Fixes https://github.com/systemd/mkosi/issues/3827

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -90,7 +90,13 @@ from mkosi.documentation import show_docs
 from mkosi.installer import clean_package_manager_metadata
 from mkosi.installer.pacman import Pacman
 from mkosi.installer.zypper import Zypper
-from mkosi.kmod import gen_required_kernel_modules, is_valid_kdir, loaded_modules, process_kernel_modules
+from mkosi.kmod import (
+    filter_devicetrees,
+    gen_required_kernel_modules,
+    is_valid_kdir,
+    loaded_modules,
+    process_kernel_modules,
+)
 from mkosi.log import ARG_DEBUG, complete_step, die, log_notice, log_step
 from mkosi.manifest import Manifest
 from mkosi.mounts import (
@@ -1596,21 +1602,6 @@ def build_kernel_modules_initrd(context: Context, kver: str) -> Path:
     return kmods
 
 
-def find_devicetree(context: Context, kver: str) -> Path:
-    assert context.config.devicetree
-
-    for d in (
-        context.root / f"usr/lib/firmware/{kver}/device-tree",
-        context.root / f"usr/lib/linux-image-{kver}",
-        context.root / f"usr/lib/modules/{kver}/dtb",
-    ):
-        dtb = d / context.config.devicetree
-        if dtb.exists():
-            return dtb
-
-    die(f"Requested devicetree {context.config.devicetree} not found")
-
-
 def want_signed_pcrs(config: Config) -> bool:
     return config.sign_expected_pcr == ConfigFeature.enabled or (
         config.sign_expected_pcr == ConfigFeature.auto
@@ -1768,10 +1759,13 @@ def build_uki(
         *flatten(["--ro-bind", os.fspath(profile), os.fspath(workdir(profile))] for profile in profiles),
     ]  # fmt: skip
 
-    if context.config.devicetree:
-        dtb = find_devicetree(context, kver)
-        arguments += ["--devicetree", workdir(dtb)]
-        options += ["--ro-bind", dtb, workdir(dtb)]
+    if context.config.devicetrees:
+        dtbs = filter_devicetrees(context.root, kver, include=context.config.devicetrees)
+        switch = "--devicetree" if len(dtbs) == 1 else "--devicetree-auto"
+        for dtb_rel in dtbs:
+            dtb = context.root / dtb_rel
+            arguments += [switch, workdir(dtb)]
+            options += ["--ro-bind", dtb, workdir(dtb)]
 
     if context.config.splash:
         splash = context.root / os.fspath(context.config.splash).lstrip("/")
@@ -2015,8 +2009,16 @@ def install_type1(
         entry.parent.mkdir(parents=True, exist_ok=True)
 
     dtb = None
-    if context.config.devicetree:
-        dtb = dst / context.config.devicetree
+    source_dtb = None
+    if context.config.devicetrees:
+        dtbs = filter_devicetrees(context.root, kver, include=context.config.devicetrees)
+        if len(dtbs) != 1:
+            die(
+                "Type 1 boot entries support only single devicetree, use UKI builds for multiple devicetrees"
+            )
+
+        source_dtb = context.root / dtbs[0]
+        dtb = dst / dtbs[0].relative_to(f"usr/lib/modules/{kver}/dtb")
         with umask(~0o700):
             dtb.parent.mkdir(parents=True, exist_ok=True)
 
@@ -2039,8 +2041,8 @@ def install_type1(
             Path(shutil.copy2(initrd, dst.parent / initrd.name)) for initrd in microcode + initrds
         ] + [Path(shutil.copy2(kmods, dst / "kernel-modules.initrd"))]
 
-        if dtb:
-            shutil.copy2(find_devicetree(context, kver), dtb)
+        if dtb and source_dtb:
+            shutil.copy2(source_dtb, dtb)
 
         with entry.open("w") as f:
             f.write(
--- a/mkosi/kmod.py
+++ b/mkosi/kmod.py
@@ -510,3 +510,43 @@ def is_valid_kdir(kdir: Path) -> bool:
 
     # check that kdir contains more than just updates
     return dircontent != [kdir / "updates"]
+
+
+def filter_devicetrees(
+    root: Path,
+    kver: str,
+    *,
+    include: Iterable[str],
+) -> list[Path]:
+    if not include:
+        return []
+
+    logging.debug(f"Devicetrees include: {' '.join(include)}")
+
+    # Search standard DTB locations
+    dtb_dirs = [
+        Path("usr/lib/firmware") / kver / "device-tree",
+        Path(f"usr/lib/linux-image-{kver}"),
+        Path("usr/lib/modules") / kver / "dtb",
+    ]
+
+    matched_dtbs = []
+    globs = list(include)
+
+    with chdir(root):
+        for dtb_dir in dtb_dirs:
+            all_dtbs = [p for p in dtb_dir.rglob("*.dtb") if p.is_file() or p.is_symlink()]
+            logging.debug(f"Found {len(all_dtbs)} DTB files in {dtb_dir}")
+
+            for dtb in all_dtbs:
+                rel_path = os.fspath(dtb.relative_to(dtb_dir))
+                if globs_match_filename(rel_path, globs):
+                    logging.debug(f"Matched DTB: {rel_path} in {dtb_dir}")
+                    matched_dtbs.append(dtb)
+
+    if not matched_dtbs:
+        logging.warning(f"Devicetrees patterns '{globs}' matched 0 files")
+    else:
+        logging.debug(f"Including {len(matched_dtbs)} devicetree files")
+
+    return sorted(matched_dtbs)
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -2014,7 +2014,7 @@ class Config:
     initrd_packages: list[str]
     initrd_volatile_packages: list[str]
     microcode_host: bool
-    devicetree: Optional[str]
+    devicetrees: list[str]
     splash: Optional[Path]
     kernel_command_line: list[str]
     kernel_modules_include: list[str]
@@ -3165,10 +3165,11 @@ SETTINGS: list[ConfigSetting[Any]] = [
         help="Packages to install in the initrd that are not cached",
     ),
     ConfigSetting(
-        dest="devicetree",
+        dest="devicetrees",
         section="Content",
-        parse=config_parse_string,
-        help="Devicetree to be used by the booting kernel",
+        parse=config_make_list_parser(delimiter=","),
+        help="Devicetree(s) to be used by the booting kernel",
+        compat_names=("Devicetree",),
     ),
     ConfigSetting(
         dest="splash",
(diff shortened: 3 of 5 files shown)


mkosi/__init__.py
```diff
@@ -1585,6 +1585,27 @@ def find_devicetree(context: Context, kver: str) -> Path:
...
+    dtbs: list[Path] = []
+
+    # Search for any listed devicetrees
+    for d in (
+        context.root / f"usr/lib/firmware/{kver}/device-tree",
+        context.root / f"usr/lib/linux-image-{kver}",
+        context.root / f"usr/lib/modules/{kver}/dtb",
+    ):
+        if d.exists():
+            dtbs.extend(d.rglob("*.dtb"))
```
behrmann: I think you can skip the existence check, since if the glob is empty (even if nothing exists), it's empty
```suggestion
    dtbs = [
        dtb
        for dtbdir in [
            context.root / f"usr/lib/firmware/{kver}/device-tree",
            context.root / f"usr/lib/linux-image-{kver}",
            context.root / f"usr/lib/modules/{kver}/dtb",
        ]
        for dtb in dtbdir.rglob("*.dtb")
    ]
```

mkosi/__init__.py
```diff
@@ -1585,6 +1585,27 @@ def find_devicetree(context: Context, kver: str) -> Path:
...
+
+    # Search for devicetrees in a given directory
+    if directory:
+        dtb_dir = context.root / os.fspath(directory).lstrip("/")
+        if dtb_dir.exists():
+            dtbs.extend(dtb_dir.rglob("*.dtb"))
```
behrmann: Same
```suggestion
    if directory:
        dtb_dir = context.root / directory.relative_to("/")
        dtbs.extend(dtb_dir.rglob("*.dtb"))
```

mkosi/__init__.py
```diff
@@ -1585,6 +1585,25 @@ def find_devicetree(context: Context, kver: str) -> Path:
     die(f"Requested devicetree {context.config.devicetree} not found")
 
 
+def find_devicetrees_auto(context: Context, kver: str, directory: Optional[Path] = None) -> list[Path]:
```
daandemeyer: Why have a `directory` argument here? Just use `context.config.devicetree_auto_directory` directly in this function.

daandemeyer:
> The latter is useful for situations where the list of dtbs is not known at configure time (e.g. in postmarketOS, dtbs aren't known until package install time since we use packaging to configure the distro.. i.e. packages and config they provide influence which dtb(s) are needed to boot on particular devices)

This makes sense, but why not just make these packages install stuff to some canonical location where mkosi can look this stuff up inside the image after package installation? I'm not convinced we need a separate setting for adding an extra lookup directory for this stuff.

DaanDeMeyer (changes requested):

craftyguy:
> This makes sense, but why not just make these packages install stuff to some canonical location where mkosi can look this stuff up inside the image after package installation? I'm not convinced we need a separate setting for adding an extra lookup directory for this stuff.

Sounds good but I'm not sure how we'd want that implemented here though. kernel-install specifies `/usr/lib/modules/KERNEL_VERSION/dtb/`, so there's already a canonical location for these files... but the current `Devicetree=` option tells mkosi to pick 1 dtb under that dir.

So, maybe if `Devicetree=all` it just includes all dtbs in that dir? Or a new option to signal to mkosi to get everything in that dir (and ignore `Devicetree=` if that's also set) ?

daandemeyer:
`Devicetree=all` sounds like it could do the trick.

Or what if we make that setting take a glob and if it's a glob, we include all those device trees?

craftyguy:
Ya I think either works, though if it's a glob maybe people would try to use globbing to select more specific things and would be more difficult to implement (so implementing "all" would be easier?)

Which do you prefer?

daandemeyer:
@craftyguy I think we can go with a glob for the existing `Devicetree` setting. If it's a glob (a glob has one of `*`, `?`, `[` or `]`), then we use --devicetree-auto, otherwise we keep using --devicetree. 

And we should probably fail hard if it is a glob and type 1 entries are configured to be used since we need to be using UKIs for the devicetree-auto stuff I think.

craftyguy:
@DaanDeMeyer ok I think I got this implemented, and tested with some globs locally (expected dtbs were embedded in dtbauto sections)

daandemeyer:
Please pull in https://github.com/systemd/mkosi/pull/3881 or rebase on top of it when merged.

mkosi/__init__.py
```diff
@@ -1611,6 +1613,38 @@ def find_devicetree(context: Context, kver: str) -> Path:
     die(f"Requested devicetree {context.config.devicetree} not found")
 
 
+
+def find_devicetrees(context: Context, kver: str) -> list[Path]:
+    assert context.config.devicetree
+
+    if not glob.has_magic(os.fspath(context.config.devicetree)):
```
daandemeyer: Can you copy this function from https://github.com/python/cpython/blob/5ae8b97f6bdcadd0cc15e0046b0db09e4c6e0d6b/Lib/glob.py#L218 into util.py? This isn't actually public API

mkosi/__init__.py
```diff
@@ -1611,6 +1613,38 @@ def find_devicetree(context: Context, kver: str) -> Path:
...
+    dtb_dirs = [
+        context.root / f"usr/lib/firmware/{kver}/device-tree",
+        context.root / f"usr/lib/linux-image-{kver}",
+        context.root / f"usr/lib/modules/{kver}/dtb",
+    ]
+
+    matched_dtbs = []
+    for dtb_dir in dtb_dirs:
+        if not dtb_dir.exists():
+            continue
```
daandemeyer: No need for this check, glob will return nothing if the directory doesn't exist.

mkosi/__init__.py
```diff
@@ -1611,6 +1613,38 @@ def find_devicetree(context: Context, kver: str) -> Path:
...
+
+    # Glob pattern - search all directories
+    dtb_dirs = [
+        context.root / f"usr/lib/firmware/{kver}/device-tree",
+        context.root / f"usr/lib/linux-image-{kver}",
+        context.root / f"usr/lib/modules/{kver}/dtb",
+    ]
+
+    matched_dtbs = []
+    for dtb_dir in dtb_dirs:
```
daandemeyer: Please name this `d` instead of `dtb_dir`. loop variables can be single letter.

mkosi/__init__.py
```diff
@@ -1611,6 +1613,38 @@ def find_devicetree(context: Context, kver: str) -> Path:
...
+        if not dtb_dir.exists():
+            continue
+
+        all_dtbs = [p for p in dtb_dir.rglob("*.dtb") if p.is_file()]
```
daandemeyer: What about symlinks? If those could be used theoretically here we'll need handling for them.

mkosi/__init__.py
```diff
@@ -1611,6 +1613,38 @@ def find_devicetree(context: Context, kver: str) -> Path:
...
+        for dtb in all_dtbs:
+            rel_path = os.fspath(dtb.relative_to(dtb_dir))
+            if globs_match_filename(rel_path, [os.fspath(context.config.devicetree)], match_default=False):
```
daandemeyer: Hmm this makes me think we should probably rename the setting to `Devicetrees=`  and have it take a list of globs so we can take full advantage of the globbing logic we have for kernel modules here as well.

mkosi/__init__.py
```diff
@@ -1611,6 +1613,38 @@ def find_devicetree(context: Context, kver: str) -> Path:
...
+                matched_dtbs.append(dtb)
+
```
daandemeyer: Please debug log the amount of matched dtbs as well here

mkosi/__init__.py
```diff
@@ -1769,9 +1803,14 @@ def build_uki(
     ]  # fmt: skip
 
     if context.config.devicetree:
-        dtb = find_devicetree(context, kver)
-        arguments += ["--devicetree", workdir(dtb)]
-        options += ["--ro-bind", dtb, workdir(dtb)]
+        if glob.has_magic(os.fspath(context.config.devicetree)):
```
daandemeyer: If the setting turns into a list of globs then the logic here should change to use --devicetree if we matched a single device tree and --devicetree-auto otherwise.

craftyguy:
Picked to this branch

mkosi/__init__.py
daandemeyer: Given my other comments, we probably also don't want to special case a single devicetree and use the same finding logic for both scenarios.

mkosi/__init__.py
craftyguy: I think this is fine, `is_file` doc says:

> Return True if the path points to a regular file (or a symbolic link pointing to a regular file)


I'm not sure if this handles symlinks pointing to other symlinks... sooooo I'll just also test for `is_symlink()` here too 🤷‍♂️

craftyguy:
@DaanDeMeyer I tried to make this behave like the kernel modules regex/glob stuff, and in the latest patch I replaced `Devicetree=` with `Devicetrees=` (I asked in chat whether this is what you meant, but lmk if you think this should be a sep. option).

Anyways, it seems to work when I set `Devicetrees=rk3588-*,sc8280xp-lenovo-thinkpad-x13s-el2.dtb
`, with debug logging:


```
‣ Devicetrees include: rk3588-* sc8280xp-lenovo-thinkpad-x13s-el2.dtb
‣ Found 972 DTB files in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-jaguar-pre-ict-tester.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-armsom-sige7.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-rock-5b-pcie-srns.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-rock-5b.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-coolpi-cm5-genbook.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-edgeble-neu6a-io.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-toybrick-x0.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-jaguar.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-friendlyelec-cm3588-nas.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-tiger-haikou-haikou-video-demo.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-rock-5-itx.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-coolpi-cm5-evb.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-evb1-v10.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-turing-rk1.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-mnt-reform2.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-orangepi-5-ultra.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-rock-5b-plus.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-tiger-haikou.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-quartzpro64.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-edgeble-neu6b-io.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-orangepi-5-max.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-armsom-w3.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-nanopc-t6-lts.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-firefly-itx-3588j.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-h96-max-v58.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-nanopc-t6.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-rock-5b-pcie-ep.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-edgeble-neu6b-wifi.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-edgeble-neu6a-wifi.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-ok3588-c.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-orangepi-5-plus.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: rockchip/rk3588-evb2-v10.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Matched DTB: qcom/sc8280xp-lenovo-thinkpad-x13s-el2.dtb in usr/lib/modules/6.16.3-0-stable/dtb
‣ Including 33 devicetree files
```

mkosi/__init__.py
craftyguy: No longer relevant

mkosi/__init__.py
```diff
@@ -2039,8 +2044,8 @@ def install_type1(
             Path(shutil.copy2(initrd, dst.parent / initrd.name)) for initrd in microcode + initrds
         ] + [Path(shutil.copy2(kmods, dst / "kernel-modules.initrd"))]
 
-        if dtb:
-            shutil.copy2(find_devicetree(context, kver), dtb)
+        if dtb and source_dtb is not None:
```
daandemeyer: Simplify this to `if dtb and source_dtb`

mkosi/kmod.py
```diff
@@ -510,3 +510,48 @@ def is_valid_kdir(kdir: Path) -> bool:
...
+
+    matched_dtbs = []
+    patterns = [p[3:] for p in include if p.startswith("re:")]
+    regex = re.compile("|".join(patterns)) if patterns else None
+    globs = [p for p in include if not p.startswith("re:")]
+
+    with chdir(root):
+        for dtb_dir in dtb_dirs:
+            if not dtb_dir.exists():
+                continue
```
daandemeyer: This check can still be dropped I think, glob will return 0 results if the dir doesn't exist.

PR #3833, opened 2025-07-29
mkosi-vm: Drop tpm_tis change

DaanDeMeyer:
Follow up for 07c24a7d42b20d4fde14e6b0bf4b1d77ac299d13

Setting this for mkosi-vm means only tpm_tis is included and nothing else is included which is bogus so revert the change.

### DIFF
--- a/mkosi/resources/mkosi-vm/mkosi.conf
+++ b/mkosi/resources/mkosi-vm/mkosi.conf
@@ -15,6 +15,3 @@ Packages=
         strace
         systemd
         udev
-
-KernelInitrdModules=
-        tpm_tis


bluca:
Doesn't that mean "also add"? If not, how can that be expressed? Otherwise tpm support will be broken again

behrmann:
> Doesn't that mean "also add"?

No, not for empty exclude. I misremembered the semantics, sorry.

bluca:
Ok so what's the right way to add something?

daandemeyer:
> Ok so what's the right way to add something?

No need, all modules are added by default. Only if you override the setting to an explicit list of modules, then you need to ofc also add tpm_tis to that explicit list, but you already did that for mkosi-initrd's default modules so that should be sufficient.

behrmann:
The current semantics are in `gen_required_kernel_modules` and  `filter_kernel_modules`. Without includes or excludes everything matching the glob `*.ko*` is added, otherwise all of these are taken and only includes minus excludes are kept.

Issue #3837, opened 2025-08-05
Allow grub BIOS boot without ESP


X4mp:
Hi,
We are building an ubuntu image for an use-case where grub BIOS boot without ESP partition is necessary. 
We use `BiosBootloader=grub`, `Bootable=yes` and `Bootloader=grub` as well in our mkosi.conf.
Unfortunately, if we do not provide the repart config for an ESP partition (only bios grub, xbootldr and root) mkosi complains about the missing esp partition.
We have hacked a bit around and got it working. I'll supply the POC changes via an MR below.

First and foremost I'd like to generally chat about the potential changes necessary to sustainably implement support for this and if possible to have some maintainer have a look at our changes and help guiding us to get this implemented if possible.

## Repart configuration:
00-bios.conf
``` 
[Partition]
# UUID of the grub BIOS boot partition which grubs needs on GPT to
# embed itself into.
Type=21686148-6449-6e6f-744e-656564454649
SizeMinBytes=1M
SizeMaxBytes=1M
``` 
10-xbootldr.conf
```
[Partition]
Label=boot
Type=bc13c2ff-59e6-4262-a352-b275fd6f7172
SizeMinBytes=1024M
SizeMaxBytes=2048M
Format=ext4
CopyFiles=/boot:/
```
20-root.conf
```
# root partiton ( / )
[Partition]
Type=root
Label=root
Format=ext4
CopyFiles=/
Minimize=guess
```

mkosi.conf
```
[Distribution]
Distribution=ubuntu
Release=noble

[Output]
Format=disk

[Content]
Bootable=true
Bootloader=grub
BiosBootloader=grub
Packages=
             ...
```

We hope to contribute our changes to mkosi upstream if possible to help other people who might have similar use-cases.

behrmann:
Which mkosi version have you had this problem with?

behrmann:
Having had a look at your PR, let's maybe step back: I now see the ext4 in your xbootldr partition definition. Is your plan to have a separate `/boot` partition, not an xbootldr partition?

X4mp:
I'm sorry, I'm not an expert on the topic and get confused with naming and what is what quite easily. 
I assume you are correct and it's not strongly an xbootldr partition, but rather a separate `/boot` partition. On the specifics why, i'd have to talk to my colleagues here, but I'm now reading more into xbootldr and boot partition in general.

If it would be vfat formated, would it work the same way? I guess the change regarding the decision making where to put grub is still necessary.

[...]

PR #3838, opened 2025-08-05
Allow Grub BIOS boot without ESP partition

X4mp:
We need to build an image with following partition:
- Bios grub
- XBOOTLDR (label=boot, CopyFiles=/boot:/) 
- root

But mkosi doesn't support this yet. 
In addition we need grub to be installed to the XBOOTLDR partition instead of the BIOS partition with efi/ prefix.

This MR relates to https://github.com/systemd/mkosi/issues/3837


mkosi/bootloader.py
```diff
@@ -321,7 +332,7 @@ def install_grub(context: Context) -> None:
         return
 
     if want_grub_bios(context):
-        grub_mkimage(context, target="i386-pc", modules=("biosdisk",))
+        grub_mkimage(context, target="i386-pc", modules=("biosdisk","ext2"))
```
X4mp: We'd like to get rid of this specific changes, but we aren't aware how we should add ext4 FS support to grub otherwise? Without it, grub won't understand the ext4 XBOOTLDR disk.

mkosi/bootloader.py
```diff
@@ -105,10 +105,6 @@ def want_grub_bios(context: Context, partitions: Sequence[Partition] = ()) -> bo
     if partitions and not bios and context.config.bootable == ConfigFeature.enabled:
         die("A BIOS bootable image with grub was requested but no BIOS Boot Partition was configured")
 
-    esp = any(p.type == "esp" for p in partitions)
-    if partitions and not esp and context.config.bootable == ConfigFeature.enabled:
-        die("A BIOS bootable image with grub was requested but no ESP partition was configured")
```
behrmann: This seems incorrect to me in the light that later on you do note in a comment that there should be either esp or xbootldr. Having neither should still be an error condition.

mkosi/bootloader.py
```diff
@@ -169,8 +176,8 @@ def prepare_grub_config(context: Context) -> Optional[Path]:
         with umask(~0o700):
             earlyconfig.parent.mkdir(parents=True, exist_ok=True)
 
-        # Read the actual config file from the root of the ESP.
-        earlyconfig.write_text(f"configfile /{context.config.distribution.grub_prefix()}/grub.cfg\n")
+        # Point the shim to the correct grub.cfg location (xbootldr or esp)
+        earlyconfig.write_text(f"configfile {grub_prefix}/grub.cfg\n")
```
behrmann: This changes the ESP case from `/<grub prefix>/grub.cfg` to `/efi/<grub prefix>/grub.cfg`.

mkosi/bootloader.py
```diff
@@ -321,7 +328,7 @@ def install_grub(context: Context) -> None:
         return
 
     if want_grub_bios(context):
-        grub_mkimage(context, target="i386-pc", modules=("biosdisk",))
+        grub_mkimage(context, target="i386-pc", modules=("biosdisk", "ext2"))
```
behrmann: Why is `ext2` here?

mkosi/bootloader.py
behrmann: Sorry, overlooked this earlier. 

xbootldr with ext4 is, I think, contrary to the idea. xbootldr is supposed to be a partition the (EFI) firmware can read, so it is usually assumed to be fat, just like an ESP.

mkosi/bootloader.py
X4mp: yes you are correct. It's probably better to follow the specification and allow to have either, ESP or XBOOTLDR partition.

mkosi/bootloader.py
X4mp: This is not necessary anymore. We have switched to the correct partition format for xbootldr.

X4mp:
Can someone restart the failed integration test? I think it's not related to this PR.

behrmann:
Yes, the CI-failue is unrelated.

bluca:
Why is this needed? Can't you just set Bootable=no?

X4mp:
> Why is this needed? Can't you just set Bootable=no?

According to https://github.com/systemd/mkosi/blob/44161624/mkosi/bootloader.py#L88 setting `Bootable=no` would also break our grub setup.

bluca:
Sure, it just means you have to provide the setup yourself in your own hook script

X4mp:
> Sure, it just means you have to provide the setup yourself in your own hook script

I'm not familiar with the hook script you are refering to. Maybe you can give an example or link to the documentation so I can look it up.

Otherwise, @behrmann would you mind having another look at the updated code changes?

behrmann:
> I'm not familiar with the hook script you are refering to. Maybe you can give an example or link to the documentation so I can look it up.

What @bluca is referring to is doing the setup yourself in an `mkosi.finalize` script, i.e. sidestepping mkosi doing any setup (`Bootable=no`) and doing it yourself.

X4mp:
> > I'm not familiar with the hook script you are refering to. Maybe you can give an example or link to the documentation so I can look it up.
> 
> What @bluca is referring to is doing the setup yourself in an `mkosi.finalize` script, i.e. sidestepping mkosi doing any setup (`Bootable=no`) and doing it yourself.

I guess that's possible. Haven't tried it.
But isn't it be better to extend the existing grub/bios support in mkosi to allow avoiding esp, since it's not strictly needed with this combo?

X4mp:
We have decided, because of more complex reasons than I can detail here, to move the setup to something similar @bluca  was initially mentioning. We now build just an archive and use custom tooling to create the final image with grub and everything needed.
Therefore I am closing this MR.

PR #3840, opened 2025-08-05
ci: give a hint about possible fixes for failing reuse lints

behrmann:


.github/workflows/ci.yml
```diff
@@ -46,7 +46,10 @@ jobs:
       - name: License Checking (reuse)
         run: |
           reuse --version
-          reuse lint
+          if ! reuse lint
```
bluca, line 49: actually this means the test won't fail anymore no? given the if eats the error?

behrmann: Absolutely right. That's what I get for doing this right before going to bed.

PR #3841, opened 2025-08-06
Drop microsecond resolution for datetime.now()

aplanas:
The RPM INSTALLTIME attribute is an integer represetantion of the installation time of the package, and datetime.now is a date representation of a float timestamp.  This can produce some rounding errors is powerful build servers.

For example, if the variable `_init_timestamp` has a value XXXXX.1 but in the same sub-second the package gets installed, the registered installation time will be the integer representation (XXXXX), making the comparison done for exclusion of the package to be `True`.

This patch will remove the microsecond granularity of the datetime, converting the timestamp on its integer representation, instead of the default float one.  The comparison is still done in the datetime data type.

### DIFF
--- a/mkosi/manifest.py
+++ b/mkosi/manifest.py
@@ -69,7 +69,9 @@ class Manifest:
     packages: list[PackageManifest] = dataclasses.field(default_factory=list)
     source_packages: dict[str, SourcePackageManifest] = dataclasses.field(default_factory=dict)
 
-    _init_timestamp: datetime.datetime = dataclasses.field(init=False, default_factory=datetime.datetime.now)
+    _init_timestamp: datetime.datetime = dataclasses.field(
+        init=False, default_factory=lambda: datetime.datetime.now().replace(microsecond=0)
+    )
 
     def need_source_info(self) -> bool:
         return ManifestFormat.changelog in self.context.config.manifest_format


mkosi/manifest.py
```diff
@@ -69,7 +69,7 @@ class Manifest:
     packages: list[PackageManifest] = dataclasses.field(default_factory=list)
     source_packages: dict[str, SourcePackageManifest] = dataclasses.field(default_factory=dict)
 
-    _init_timestamp: datetime.datetime = dataclasses.field(init=False, default_factory=datetime.datetime.now)
```
behrmann, line 72: I think a smaller change could be
```diff
-   _init_timestamp: datetime.datetime = dataclasses.field(init=False, default_factory=datetime.datetime.now)
+   _init_timestamp: datetime.datetime = dataclasses.field(init=False, default_factory=lambda: datetime.datetime.now().replace(microsecond=0))
```
leaving the rest unchanged.

aplanas:
> I understand from the description that the problem seems to be a difference in resolution (time coming in seconds and datetime having microsecond resolution), but can you explain a bit more what the resulting problem is?

Ah sure.

The problem is that the package is collected in the metadata if the installed time is bigger or equal than the time that the metadata object was created.  This is used to separate packages that comes from the base image from the ones that comes from the extension, IIUC.  Because the information in the RPM has second as a resolution, can happen that this check become false: `datetime.datetime.fromtimestamp(int(installtime)) < self._init_timestamp` can be true, because XXXXX < XXXXXX.5, for example.  This will exclude packages from the metadata list, depending on the execution time.

mkosi/manifest.py
aplanas: I found comparing `int` more intuitive, but I will test this change tomorrow

aplanas: Yes. Dropped the time() approach and use this instead. Done

behrmann:
Thanks for the explanation! Can you put a (maybe shortened) version of this into the commit message and adjust it, since it still talks about comparing the times as integers? 

Apart from the commit message, LGTM.

aplanas:
> commit message

Updated.

PR #3843, opened 2025-08-07
README: clarify that companion tools can also be enabled from the git repo

aafeijoo-suse:


behrmann:
Not necessary to burn CI on this, since it's a documentation-only change.

PR #3848, opened 2025-08-09
Use apt-ftparchive as fallback for reprepro

tmuehlbacher:
Afaict, this should work quite similar to the previous dpkg-scanpackages but also doesn't use perl and is part of apt. Certain distros don't have reprepro packaged, e.g. nixpkgs.

Disclaimer: I haven't tested this yet and I am not good at python.


daandemeyer:
@tmuehlbacher apt-ftparchive seems like a better tool for the job than reprepro for our specific use case, so I'm fine with just replacing reprepro with it. No need for the fallback path, you can get rid of reprepro entirely. I'd have used apt-ftparchive instead of reprepro if I had known it existed when I replaced dpkg-scanpackages.

daandemeyer:
cc @whiteley this might affect if apt-ftparchive is more strict then dpkg-scanpackages like reprepro was.

tmuehlbacher:
@DaanDeMeyer, sure. I just wasn't sure if there is something useful there in terms of features with reprepro that I wasn't aware of.

Also don't know if there are additional things to do, like does this have to added to the list of `"apt-*"` commands in `.scripts()`?

daandemeyer:
@tmuehlbacher I've tested this and made the necessary changes I would have left as comments otherwise. Please double check the changes look OK to you.

behrmann:
>  I just wasn't sure if there is something useful there in terms of features with reprepro that I wasn't aware of.

One thing that can be useful is overriding config which can be used to override fields in the Debian package, see #3730. This might not be a problem with `apt-ftparchive`, but for completeness: Is there a way to achieve something similar with `apt-ftparchive`, e.g. override the section or suite field (specifically the `Suite: ./` field looks quite curious))?

daandemeyer:
> > I just wasn't sure if there is something useful there in terms of features with reprepro that I wasn't aware of.
> 
> One thing that can be useful is overriding config which can be used to override fields in the Debian package, see #3730. This might not be a problem with `apt-ftparchive`, but for completeness: Is there a way to achieve something similar with `apt-ftparchive`, e.g. override the section or suite field (specifically the `Suite: ./` field looks quite curious))?

I don't think there is, but I'm assuming it won't be required as I think apt-ftparchive is going to be less strict than reprepro is

tmuehlbacher:
The way I understand it, reprepro manages the actual repo and dir hierarchy and such but dpkg-scanpackages and apt-ftparchive just take the dir as it is and output `Packages` files. So I don't think there should be any issues, however I don't know for sure how malformed packages are handled.

The `./` part is same as how it was with dpkg-scanpackages. I think makes sense when you look here: https://wiki.debian.org/DebianRepository/Format#Overview

behrmann:
Merging since @tmuehlbacher gave a thumbs up on the latest iteration from Daan, so this should be good to go.

PR #3853, opened 2025-08-14
mkosi-initrd: add --workspace-dir option

### LINKED ISSUE #3852: mkosi-initrd does not respect WorkspaceDirectory

71e8404

### Used host distribution

Debian 13

### Used target distribution

Debian 13

### Linux kernel version used

6.12.38+deb13-amd64

x86_64

mkosi-initrd does not respect the use of WorkspaceDirectory because the default value of `/var/tmp` is explicitly set on the command line.

https://github.com/systemd/mkosi/blob/eeb4ce6302bbf742440d9b31f7813a79279d95ea/mkosi/initrd.py#L377

This is problematic for a number of reasons, but the primary one in my use case is that `/var/tmp` is often mounted `noexec` as part of hardening policies on secure systems. mkosi-initrd works perfectly fine with a dedicated workspace directory created in a less exposed location, but `initrd.py` must be locally modified to not pass the `--workspace-dir` option for it to properly function.

When `/var/tmp` is mounted `noexec`, mkosi fails to run post-installation scripts for distribution packages (and any executable in the build space), which results in an overall failed build.

### Used mkosi config

```ini
[Build]
WorkspaceDirectory=/var/lib/mkosi-initrd/build
```

### mkosi output

```sh

```

aafeijoo-suse:
Fixes #3852

### DIFF
--- a/mkosi/resources/man/mkosi-addon.1.md
+++ b/mkosi/resources/man/mkosi-addon.1.md
@@ -34,6 +34,10 @@ the running hardware can be included if a local configuration with the option
 :   Path to a directory where to place all generated artifacts. Defaults to the
     current working directory.
 
+`--workspace-dir=`
+:   Path to a directory where to store data required temporarily while
+    building the image. Defaults to `/var/tmp`.
+
 `--debug`
 :   Enable additional debugging output.
 
--- a/mkosi/addon.py
+++ b/mkosi/addon.py
@@ -67,7 +67,7 @@ def main() -> None:
 
         if os.getuid() == 0:
             cmdline += [
-                "--workspace-dir=/var/tmp",
+                f"--workspace-dir={args.workspace_dir}",
                 "--output-mode=600",
             ]
 
--- a/mkosi/initrd.py
+++ b/mkosi/initrd.py
@@ -269,6 +269,13 @@ def initrd_common_args(parser: argparse.ArgumentParser) -> None:
         type=Path,
     )
     parser.add_argument(
+        "--workspace-dir",
+        metavar="DIR",
+        help="Workspace directory",
+        default="/var/tmp",
+        type=Path,
+    )
+    parser.add_argument(
         "--debug",
         help="Turn on debugging output",
         action="store_true",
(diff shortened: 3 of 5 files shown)


mkosi/resources/man/mkosi-addon.1.md
```diff
@@ -34,6 +34,10 @@ the running hardware can be included if a local configuration with the option
 :   Path to a directory where to place all generated artifacts. Defaults to the
     current working directory.
 
+`--workspace-dir=`
+:   Path to a directory where to store data required temporarily while
+    building the image. Defaults to `/var/tmp`.
```
behrmann, line 39: Can you add a note like
```suggestion
:   Path to a directory where to store data required temporarily while
    building the image. Defaults to `/var/tmp`. For further details see **mkosi(1)**.
```
so that people have a hint that there's environment variables that affect this as well?

Also, looking at our documentation, I see that we do document `XDG_CACHE_DIR` affecting `WorkspaceDirectory=`, but not `CACHE_DIRECTORY`, although it does as well. Could you add this?

aafeijoo-suse: > so that people have a hint that there's environment variables that affect this as well?

Well this is not true with the current patch setting `default="/var/tmp"` to preserve the old behavior.
But, if this is not really necessary, I would just drop adding `--workspace-dir=` always to the `mkosi` call and just rely on config (if it is necessary to change the default value mentioned in mkosi(1)).

> Also, looking at our documentation, I see that we do document XDG_CACHE_DIR affecting WorkspaceDirectory=, but not CACHE_DIRECTORY, although it does as well. Could you add this?

Ok.

behrmann (approved):
Sorry, forgot about this

aafeijoo-suse:
> Sorry, forgot about this

No problem and no rush :)

I'd wait for @DaanDeMeyer before merging this, because he added `--workspace-dir=/var/tmp` to the mkosi call in bce6b59669494268a672f9aeddbf9e0887f22b5a

If the current default is good enough, I'd just remove that.

behrmann:
Merging. CI failures are the current Fedora SBAT PE header size issues and the Arch dDoS..

PR #3857, opened 2025-08-14
fedora: be more persistent when guessing what rawhide could be

behrmann:


bluca (approved):
Everything is broken and this fixes it, so let's go for it and adjust later if needed

PR #3860, opened 2025-08-17
mkosi-tools: include opensuse-release in suse builds

bluca:
Otherwise os-release is missing and the img can't be identified

### DIFF
--- a/mkosi/resources/mkosi-tools/mkosi.conf
+++ b/mkosi/resources/mkosi-tools/mkosi.conf
@@ -11,6 +11,7 @@ Packages=
         curl
         dosfstools
         e2fsprogs
+        jq
         keyutils
         kmod
         mtools
--- a/mkosi/resources/mkosi-tools/mkosi.conf.d/opensuse.conf
+++ b/mkosi/resources/mkosi-tools/mkosi.conf.d/opensuse.conf
@@ -12,6 +12,7 @@ Packages=
         erofs-utils
         grub2-common
         libseccomp2
+        openSUSE-release
         pkcs11-provider
         policycoreutils
         python3-pefile
--- a/mkosi/resources/mkosi-tools/mkosi.conf.d/efi.conf
+++ b/mkosi/resources/mkosi-tools/mkosi.conf.d/efi.conf
@@ -1,0 +1,8 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+[Match]
+Architecture=uefi
+
+[Content]
+Packages=
+        efitools
--- a/mkosi/resources/mkosi-tools/mkosi.profiles/misc/mkosi.conf
+++ b/mkosi/resources/mkosi-tools/mkosi.profiles/misc/mkosi.conf
@@ -10,7 +10,6 @@ Packages=
         findutils
         gawk
         grep
-        jq
         less
         man
         nano


mkosi/resources/mkosi-tools/mkosi.conf
```diff
@@ -11,6 +11,7 @@ Packages=
         curl
         dosfstools
         e2fsprogs
+        jq
```
behrmann, line 14: I don't think that should be changed. This was split off in #3527 with the intent of being pulled in by `ToolsTreeProfiles=` and ensuring that the tools tree as lean as possible and adding it back is a slippery slope for all the other stuff. I think pulling in the `misc` profile or adding jq explicitly with `ToolsTreePackages=` should be the way to go for OBS.

bluca: `mkosi dependencies` doesn't use that

behrmann: But can you do `mkosi dependencies -- --profile misc` instead of `mkosi dependencies`? That's the union of `mkosi dependencies` and the dependencies from the `misc` tools tree profile.

bluca: the fedora spec file doesn't do that: https://src.fedoraproject.org/rpms/mkosi/blob/rawhide/f/mkosi.spec#_108

behrmann: That spec seems a bit odd?
```bash
{
  bin/mkosi dependencies | sed -e 's/^/Recommends: /'
  echo "%package system-deps"
  echo "Summary:       Pull in additional dependencies needed to build images"
  bin/mkosi dependencies | sed -e 's/^/Requires: /'
  echo "Requires:      pesign"
  echo "%description system-deps"
  echo "This package pulls in all the dependencies needed to build images"
  echo "%files system-deps"
} >%{specpartsdir}/mkosi.specpart
```
The output `mkosi dependencies` is added both as `Recommends` and `Requires`? Wouldn't it make more sense to call it with some or all profiles for `Recommends`? @keszybz, what do you think?

bluca: recommends is not pulled in at build time by default, so also doesn't work unfortunately

behrmann: That makes sense (although having the same thing both as recommends and requires still seems weird to me), but let me try to put it on a different footing: There's a lot of checks `%if %{defined suse_version}` there and though the above block is not behind such a guard, there's a `Requires: pesign`, which we don't use anywhere anymore except for OBS. Would it be an option to add an OBS profile for the tools tree and calling mkosi with that profile? For that matter why doesn't that already work with the `misc` profile? 

Maybe I'm fundamentally misunderstanding how OBS works, but at the moment I fail to understand why the package cannot be installed into the tools tree in OBS, i.e. why `ToolsTreePackages=jq` or `ToolsTreeProfiles=misc` in the `mkosi/mkosi.conf` in the systemd repo wouldn't work.

bluca: How would that work? It's just a package dependency, it's either there or it's not

mkosi/resources/mkosi-tools/mkosi.conf.d/opensuse.conf
```diff
@@ -11,6 +11,7 @@ Packages=
         distribution-gpg-keys
         erofs-utils
         libseccomp2
+        openSUSE-release
```
daandemeyer, line 15: We add this by default now these days unless another one is specified explicitly, so this shouldn't be needed.

mkosi/resources/mkosi-tools/mkosi.conf
daandemeyer: Yeah this is wrong, `jq` is not required for core operation, and so should be in a profile that needs to be explicitly enabled. We don't want `mkosi dependencies` to pull in a bunch of random stuff that's not required for core operation, and `jq` certainly is not required for core operation.

If you need `jq` as a dep of mkosi because OBS is shit then it should be gated behind a conditional in the mkosi OBS spec.

mkosi/resources/mkosi-tools/mkosi.conf.d/opensuse.conf
bluca: `mkosi dependencies` doesn't list it

mkosi/resources/mkosi-tools/mkosi.conf
bluca: It is a dependency of mkosi, it's used in the signing scripts

daandemeyer: That's an optional part of mkosi that is only used in OBS. It's not sufficient to make this listed by `mkosi dependencies` by default. It's fine to have in the `misc` profile so it can be optionally pulled in via `mkosi dependencies -- --profile misc`.

mkosi/resources/mkosi-tools/mkosi.conf.d/opensuse.conf
daandemeyer: So we actually pull this in via the `patterns-base-minimal_base` package in the misc profile as well, so it's similar to `jq`. I don't think the mkosi package should have a direct dependency on a release package, but it's fine to pull one in via the `misc` profile.

mkosi/resources/mkosi-tools/mkosi.conf
bluca: ext4 is also optional, xfs is also optional, pkcs11 is also optional, etc etc

but then the mkosi specf needs to start using that profile, otherwise the package won't pull them in

mkosi/resources/mkosi-tools/mkosi.conf.d/opensuse.conf
bluca: As above, the mkosi package needs to use that profile then

bluca:
These dependencies can either go here, or directly in the fedora spec file at https://src.fedoraproject.org/rpms/mkosi/blob/rawhide/f/mkosi.spec - I have no opinion on which is better, and I am fine with either. But it needs to be in one of these, so that the dependency on the mkosi package pulls in everything required to build images. How that dependency comes to be could be one of many ways, but it has to be one.

mkosi/resources/mkosi-tools/mkosi.conf
keszybz: > The output `mkosi dependencies` is added both as `Recommends` and `Requires`

Recommends are attached to the main `mkosi` package, `Requires` are attached to the `mkosi-system-deps` subpackage.

> Wouldn't it make more sense to call it with some or all profiles for `Recommends`

I think it'd be reasonable to include more stuff in Recommends. But I'm not sure if the list from `--profile misc` is the  right source. I get this:
```
+acl
+attr
+bash
+diffutils
+file
+findutils
+gawk
+git-core
+grep
+jq
+less
+man
+man-pages
+nano
+sed
+strace
+which
```
This seems quite … random. Maybe we should just add `jq` to that list.
Also `Recommends: less` should be added, it's not useful for building stuff, but makes `summary` and `help` nicer.

A second question is what `mkosi-system-deps` should list as `Requires`. I think it'd be fine to just add `Requires: jq` there. @bluca, you added that package IIRC. Would that work for you?

keszybz: → https://src.fedoraproject.org/rpms/mkosi/pull-request/23 ?

bluca: > A second question is what mkosi-system-deps should list as Requires. I think it'd be fine to just add Requires: jq there. @bluca, you added that package IIRC. Would that work for you?

As a temporary workaround sure, this is what I currently use: https://github.com/bluca/mkosi-fedora/commit/f49ca26d2b449e44f2b2a75301008a4f6b209578

In general, the 'mkosi-system-deps' is not actually pulled in by anything, so it's not a complete solution. If the main mkosi package requires it, then sure it would work as a definitive solution for me.

keszybz: I updated the PR to include `efitools` too.

keszybz: > If the main mkosi package requires it, then sure it would work as a definitive solution for me.

I don't think we want to add such a dependency in the general case. If necessary, we could add it conditionalized on obs.

bluca: > I updated the PR to include `efitools` too.

ok then I'll close this thanks

bluca: > > If the main mkosi package requires it, then sure it would work as a definitive solution for me.
> 
> I don't think we want to add such a dependency in the general case. If necessary, we could add it conditionalized on obs.

Unfortunately that doesn't work, AFAIK RPM doesn't have "conditional" dependencies, either a pkg requires another, or it doesn't?

keszybz: There are a few different possibilities. Firstly, we could use `%if %{with obs} … %endif`. We already do that in various places. This would mean that the packages built in obs would have the extra dep in all cases, but I think this is OK. Secondly, rpm has "rich dependencies", in the form `Requires: (a if b)` to mean that `a` is required if `b` is installed. (`b` can be a package name or a virtual Provides on the package.) This could be made to work if there's a package which is always installed in the environment where the extra dep is wanted.

bluca: > Firstly, we could use %if %{with obs} … %endif. We already do that in various places. 

That's ok as a temporary workaround, but only works if you do custom package builds, which is a temporary thing until the next release. After that, RPMs provided by Fedora (and others) need to work as they are published, without manual, local rebuilds, otherwise it gets unwieldly.

> Secondly, rpm has "rich dependencies", in the form Requires: (a if b) to mean that a is required if b is installed.

I tried:

`Requires:       (%{name}-system-deps = %{version}-%{release} if kernel-obs-build)`

unfortunately this does not seem to work, maybe the resolver doesn't support that

keszybz: I checked here, and it works as expected, with dnf5. (I don't have `kernel-obs-build`, but with some other package.)

But I also filed https://github.com/rpm-software-management/dnf5/issues/2427.

daandemeyer: Instead of this eternal discussion about what should be Recommends and what should be Requires, can't we make tools trees work in OBS? Let's just have mkosi in OBS build its own userspace from the local mirror provided by OBS and then use that to build the image. Then we don't have to mess around with the rpm at all and any extra packages required to build an image can be declared in the image configuration and don't have to all be made `Requires` in the rpm spec.

bluca: I think the issue is the package - it's not a normal dependency, it's injected in every build (as it's the kernel used to boot the ephemeral build VM). That's the only one I could find that is special and always installed, everything else is just a normal dependency chain, so unfortunately I don't think there's any anchor package available for this

bluca: > Instead of this eternal discussion about what should be Recommends and what should be Requires, can't we make tools trees work in OBS? Let's just have mkosi in OBS build its own userspace from the local mirror provided by OBS and then use that to build the image. Then we don't have to mess around with the rpm at all and any extra packages required to build an image can be declared in the image configuration and don't have to all be made `Requires` in the rpm spec.

That doesn't solve the issue, as you still need a package dependency to pull in all the dependencies needed by the mkosi tools tree

daandemeyer: @bluca Why? That doesn't happen for the regular image packages either? If you can get those into the local mirror, why wouldn't you be able to get the tools tree packages into the local mirror?

bluca: Images have dependencies on the packages they need. There's no way to know at any given time what mkosi itself needs, that's what package dependencies exist for - so that you don't need to guess, and so that the build doesn't break every time mkosi changes, and the package itself pulls in what it requires, like any other package

PR #3862, opened 2025-08-18
mkosi-tools: virtiofsd is only available on a subset of architectures on debian/ubuntu

bluca:
It's a rust program, so of course it's a major pain to ship it, so it has been restricted to a few arches and it's not available on the rest, and likely won't

### DIFF
--- a/mkosi/resources/mkosi-tools/mkosi.profiles/runtime/mkosi.conf.d/debian-kali-ubuntu/mkosi.conf
+++ b/mkosi/resources/mkosi-tools/mkosi.profiles/runtime/mkosi.conf.d/debian-kali-ubuntu/mkosi.conf
@@ -7,7 +7,6 @@ Distribution=|ubuntu
 
 [Content]
 Packages=
-        ?exact-name(virtiofsd)
         openssh-client
         ovmf
         qemu-efi-aarch64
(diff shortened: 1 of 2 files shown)


daandemeyer:
Can you add the apt error to the commit message? I would expect the regex thing to automatically skip the package on arches where it doesn't exist.

bluca:
```
Note, selecting 'git' instead of 'git-core'
Note, selecting 'man-db' instead of 'man'
Note, selecting 'gnu-which' instead of 'which'
Package virtiofsd is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'virtiofsd' has no installation candidate
‣ "/work/prepare final" returned non-zero exit code 123.
```

bluca:
command line was:

```
mkosi --tools-tree= --architecture arm -d debian -r testing --bootable yes --package systemd-boot -f -i --secure-boot no --bootloader systemd-boot --shim-bootloader none --format disk --package apt --package git --package systemd-resolved build
```

daandemeyer:
Please include the error in the commit message, makes future spelunking easier

bluca:
`ERROR    root:log.py:21 'ppc64_le' is not a valid Architecture` GODDA

PR #3867, opened 2025-08-21
fedora: Rework rawhide GPG key logic

daandemeyer:
- Drop secondary key logic as looking at https://github.com/rpm-software-management/distribution-gpg-keys/tree/main/keys/fedora,
  this hasn't been used for a long time.
- If repository key fetching is enabled, always look up the key remotely
  as e.g. on CentOS 9 or so the rawhide symlink might be horribly outdated.
- If not using repository key fetching, Use all local keys newer than the
  rawhide key as well to maximize the chances of including the current rawhide
  key.
- Resolve symlinks within the sandbox in find_rpm_gpgkey() as we might not be
  able to resolve the symlinks outside of the sandbox.

### DIFF
--- a/mkosi/distributions/fedora.py
+++ b/mkosi/distributions/fedora.py
@@ -17,7 +17,7 @@ from mkosi.distributions import (
 from mkosi.installer.dnf import Dnf
 from mkosi.installer.rpm import RpmRepository, find_rpm_gpgkey, setup_rpm
 from mkosi.log import die
-from mkosi.util import startswith, tuplify
+from mkosi.util import tuplify
 
 DISTRIBUTION_GPG_KEYS_UPSTREAM = (
     "https://raw.githubusercontent.com/rpm-software-management/distribution-gpg-keys/main/keys/fedora"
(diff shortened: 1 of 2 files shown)


mkosi/distributions/fedora.py
```diff
@@ -35,95 +35,62 @@ def read_remote_rawhide_key_symlink(context: Context) -> str:
...
-                            Path(d),
-                            log=False,
-                        )
-
-                    yield f"{DISTRIBUTION_GPG_KEYS_UPSTREAM}/RPM-GPG-KEY-fedora-{version + 1}-primary"
-                except subprocess.CalledProcessError:
-                    pass
-        else:
-            yield "https://fedoraproject.org/fedora.gpg"
+    if context.config.release == "rawhide" and (rawhide_will_be := versionre.match(Path(key).name)):
```
behrmann, line 78: If `context.config.release` and `context.config.release` is false, `versionre` cannot match here if the rawhide key is not a symlink, which should never happen, but I saw it in CI here for some reason. I think a reasonable fallback would be either taking all Fedora keys, or taking all starting with an arbitrary cutoff (e.g. starting now with 42), or with what we define as default release here.

daandemeyer: > If context.config.release and context.config.release is false

Hmm? I don't follow this

behrmann: Sorry, that was missing "is rawhide" after release.

Let's assume `context.config.release` is `"rawhide"` and `context.config.repository_key_fetch` is false. Then the first block is skipped and we yield `"RPM-GPG-KEY-fedora-{context.config.release}-primary"`, which will have been canonicalised to the release number. But only if the file is a symlink. 

I know this doesn't make an awful lot of sense, since it is a symlink in the `distributon-gpgp-keys`, but when I wrote this check, this blew up in CI, which is why the `version_file.readlink()` is guarded behind a `version_file.is_symlink()`. When I print-debugged this, `version_file.is_symlink()` was false, so this would canonicalise to itself, the rawhide key

I think it would make sense to sidestep this issue and if `versionre.match(Path(key).name))` fails to fall back to some arbitrary version number we hardcode, e.g. 42, or to read it from the distribution-gpg-keys repo as well, since I think that's still ok, because we're not fetching the key, we're just fetching which locally existing key to use.

daandemeyer: I don't think that's necessarily required though, in that case we'll still have the rawhide key itself. I really don't want to hardcode release numbers in mkosi. And doing all the keys we can find is a bit too much. We can of course find the newest key and use up until N-X earlier keys but let's not bother until this actually turns out to be a problem in practice since that'd be a pain to implement.

Let's see if we run into any troubles with this approach first.

behrmann: > I don't think that's necessarily required though, in that case we'll still have the rawhide key itself.

Well, but these changes came from the time period where the rawhide key itself didn't match?

> I really don't want to hardcode release numbers in mkosi. And doing all the keys we can find is a bit too much. We can of course find the newest key and use up until N-X earlier keys but let's not bother until this actually turns out to be a problem in practice since that'd be a pain to implement.

Taking the last k keys would also be nice. I don't think it's that ugly even, e.g. the last 3 version numbers
```python
sorted(
    int(m.group(1)) 
    for p in Path("/usr/share/distribution-gpg-keys/fedora/").glob("RPM-GPG-KEY-fedora-*-primary") 
    if (m := versionre.match(p.name))
)[-3:]
```

daandemeyer: That would still maybe make us add older keys than the rawhide one as well, potentially even for EOL releases since those are shipped in distribution-gpg-keys. I don't want to do that.

behrmann: That could happen here as well, when counting keys up, if the distribution-gpg-keys package is sufficiently old. But instead of taking the highest version keys or hardcoding a release, we could still keep fetching the highest release number from the distribution-gpg-repo, even if we don't fetch the keys.

[Here's the debug print statement](https://github.com/systemd/mkosi/compare/7bfe763899b2beb3973d55f53d50a46ddaab1fe0..6cac7b44632ffe81afbf2bd4f4f87c7783d069a3#diff-7b02ca2c83f4e2f2b991f9a5b19f9d1175f1fc30f20d6a1d7a8e810237b5118aL44)  and [here's the CI output](https://github.com/systemd/mkosi/actions/runs/16975851922/job/48124601047#step:10:63). At that point the rawhide key was the wrong one and it wasn't pointing to a release to count on from in CI, so with this change we'd just have the rawhide key and wouldn't be able to build Fedora.

daandemeyer: @behrmann The reason you got that result is because you were resolving the symlink outside of the sandbox. You were doing the symlink resolution on the host and the rawhide key file isn't even installed on the host in the first place so naturally it couldn't be resolved.

This PR changes the symlink resolution so it's done in the sandbox so this situation can't occur anymore.

behrmann: Ah, that's it. Yes. :man_facepalming:

PR #3869, opened 2025-08-22
initrd: Stop passing --workspace-dir=

### LINKED ISSUE #3852: mkosi-initrd does not respect WorkspaceDirectory

daandemeyer:
The default value when running as root is /var/tmp now so there's no
need to specify --workspace-dir= explicitly anymore. This allows the
workspace directory to be changed in the configuration file in
/etc/mkosi-initrd.
    
Fixes #3852

### DIFF
--- a/mkosi/addon.py
+++ b/mkosi/addon.py
@@ -66,10 +66,7 @@ def main() -> None:
             cmdline += ["--debug-sandbox"]
 
         if os.getuid() == 0:
-            cmdline += [
-                "--workspace-dir=/var/tmp",
-                "--output-mode=600",
-            ]
+            cmdline += ["--output-mode=600"]
 
         cmdline += include_system_config("mkosi-addon")
         cmdline += process_crypttab(Path(staging_dir))
--- a/mkosi/initrd.py
+++ b/mkosi/initrd.py
@@ -374,7 +374,6 @@ def main() -> None:
 
         if os.getuid() == 0:
             cmdline += [
-                "--workspace-dir=/var/tmp",
                 "--package-cache-dir=/var",
                 "--cache-only=metadata",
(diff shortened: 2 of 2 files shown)


mkosi/addon.py
```diff
@@ -67,7 +67,7 @@ def main() -> None:
 
         if os.getuid() == 0:
             cmdline += [
-                f"--workspace-dir={args.workspace_dir}",
+                "--workspace-dir=/var/tmp",
```
behrmann: Wouldn't hardcoding `/var/tmp` here make `mkosi-addon` fail in the case it is `noexec`, since CLI overrides the config file?

PR #3871, opened 2025-08-22
rhel: Look up entitlement keys and certificates in sandbox

### LINKED ISSUE #3854: mkosi uses wrong path to keys and entitlements when setting up repos for RHEL

main

### Used host distribution

Ubuntu 24.10

### Used target distribution

RHEL 10

### Linux kernel version used

6.11.0-29-generic

x86_64

When mkosi creates the repo file, it won't use the proper path to the ssl certs and keys:

### Used mkosi config

```ini
No specific configuration needed, just a `mkosi.sandbox` folder with the certs.
```

### mkosi output

```sh
bash-5.2# cat /etc/yum.repos.d/mkosi.repo | grep -i ssl | sort | uniq
sslcacert=/var/tmp/mkosi-workspace-t1epxo7k/sandbox/etc/rhsm/ca/redhat-uep.pem
sslclientcert=/var/tmp/mkosi-workspace-t1epxo7k/sandbox/etc/pki/entitlement/7091014227886103379.pem
sslclientkey=/var/tmp/mkosi-workspace-t1epxo7k/sandbox/etc/pki/entitlement/7091014227886103379-key.pem
bash-5.2# ls /var/tmp/mkosi-workspace-t1epxo7k/sandbox
ls: cannot access '/var/tmp/mkosi-workspace-t1epxo7k/sandbox': No such file or directory
bash-5.2# ls /etc/pki/entitlement/7091014227886103379.pem
/etc/pki/entitlement/7091014227886103379.pem
```

DaanDeMeyer:
Similar to how we handle rpm GPG keys, let's also look up the extra stuff for RHEL in the sandbox in the same way.

Fixes #3854


daandemeyer:
@g-arjones Can you check if this works, I don't have a RHEL subscription at hand.

mkosi/distributions/rhel.py
```diff
@@ -30,45 +31,44 @@ def sslcacert(context: Context) -> Optional[Path]:
...
             return None
 
-        pattern = "etc/pki/entitlement/*.pem"
-
-        p = next((p for p in sorted(context.sandbox_tree.glob(pattern)) if "key" not in p.name), None)
-        if not p:
-            p = next((p for p in sorted(Path("/").glob(pattern)) if "key" not in p.name), None)
-        if not p:
-            die("Entitlement certificate not found in host system or sandbox tree")
+        glob = "/etc/pki/entitlement/*.pem"
```
g-arjones, line 63: You have to exclude `*-key.pem` otherwise the key will be picked instead of the cert (because of the sorting order)

g-arjones: See:

```console
bash-5.2# cat /etc/yum.repos.d/mkosi.repo | grep sslclient | sort | uniq
sslclientcert=/etc/pki/entitlement/7091014227886103379-key.pem
sslclientkey=/etc/pki/entitlement/7091014227886103379-key.pem
```

behrmann: Since we're doing this in bash anyway, we can also turn on extglob in `glob_in_sandbox` and do
```suggestion
        glob = "/etc/pki/entitlement/!(*-key).pem"
```

g-arjones: Isn't it easier and more elegant to do it in mkosi (host context) in Python and then use `Path.relative_to` to make it relative to the sandbox tree? Or maybe use `find` instead of `bash` if it really has to be done from the sandbox.

daandemeyer: Doing it like this is less fragile as regardless of where the certificates come from in the sandbox, this will find them. Doing it in python means we have to check every possible source of the /etc/pki/entitlement directory in the sandbox, which is a lot more fragile.

Using `find` means adding a dependency on `findutils` which I'd only add if it saves a substantial amount of complexity which I don't think we've reached yet.

g-arjones: The new glob produces a bash syntax error:

```console
‣ Copying in sandbox trees…
bash: -c: line 1: syntax error near unexpected token `('
bash: -c: line 1: `shopt -s nullglob extglob && printf '%s\n' /etc/pki/entitlement/!(*-key).pem | xargs -r readlink -f'
‣ "bash -c 'shopt -s nullglob extglob && printf '"'"'%s\n'"'"' /etc/pki/entitlement/!(*-key).pem | xargs -r readlink -f'" returned non-zero exit code 2.
```

You can test this by adding dummy empty `/etc/pki/entitlement/0.pem` and `/etc/pki/entitlement/0-key.pem` files to your sandbox. Run with `--debug-shell`... It will fail, but you will get a chance to inspect the generated `mkosi.repo` file to see if it's as expected. You don't really need a RHEL subscription to test this.

g-arjones: I suggest keeping the glob as simple as possible and filtering `*-key.pem` out in the calling method (`rhel.Installer.sslclientcert`) instead of just returning `paths[0]`

g-arjones:
It works now, thanks a lot!

g-arjones (approved):

behrmann (approved):

PR #3872, opened 2025-08-22
kmod: Make sure we include all intermediate firmware symlinks as well

DaanDeMeyer:
Replaces #3822

### DIFF
--- a/mkosi/kmod.py
+++ b/mkosi/kmod.py
@@ -10,10 +10,10 @@ from collections.abc import Iterable, Iterator, Reversible
 from pathlib import Path
 
 from mkosi.context import Context
-from mkosi.log import complete_step, log_step
+from mkosi.log import complete_step, die, log_step
 from mkosi.run import chroot_cmd, run
 from mkosi.sandbox import chase
-from mkosi.util import chdir, parents_below
+from mkosi.util import chdir, parents_below, path_and_parents_below
 
 
 def loaded_modules() -> list[str]:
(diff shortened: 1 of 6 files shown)


daandemeyer:
@sorenisanerd Can you give this a try?

daandemeyer:
Verified now that this works as expected.

mkosi/kmod.py
```diff
@@ -165,7 +165,7 @@ def filter_firmware(
     if include:
         firmwared = Path("usr/lib/firmware")
         with chdir(root):
-            all_firmware = set(firmwared.rglob("*"))
+            all_firmware = set(p for p in firmwared.rglob("*") if p.is_file() or p.is_symlink())
```
```suggestion
            all_firmware = {p for p in firmwared.rglob("*") if p.is_file() or p.is_symlink()}
```

mkosi/kmod.py
```diff
@@ -349,14 +349,56 @@ def gen_required_kernel_modules(
...
+        # /usr/lib/firmware/nvidia/gb102
+        # /usr/lib/firmware/nvidia/gb203
+        # /usr/lib/firmware/nvidia/gb205
+        # /usr/lib/firmware/nvidia/gb206
+        # /usr/lib/firmware/nvidia/gb207
+        # /usr/lib/firmware/nvidia/tu104/gsp
+        # /usr/lib/firmware/nvidia/tu106/gsp
+        # /usr/lib/firmware/nvidia/tu117/gsp
+
+        for p in (*reversed(parents_below(context.root / fw, context.root / firmwared)), context.root / fw):
```
behrmann: The unpacking makes this more difficult to understand, I think it would be easier to add a function
```python
def path_and_parents_below(path: Path, below: Path) -> list[Path]:
    return [path] + parents_below(path, below)
```
and use that here

behrmann (approved):
Just two minor comments

PR #3874, opened 2025-08-24
mkosi-obs fixlets

bluca:


behrmann (approved):
LGTM, since these are the two bug fixes split out off #3860 that only affect OBS.

PR #3875, opened 2025-08-24
Check whether a UKI is wanted in make_uki() like in other paths

bluca:
Otherwise it will be built for the ESP even if it's not wanted, like in the netesp case in particleos

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3700,7 +3700,7 @@ def make_esp(
     if not context.config.architecture.to_efi():
         die(f"Architecture {context.config.architecture} does not support UEFI")
 
-    if stub and kver and kimg:
+    if stub and kver and kimg and want_uki(context):
         token = find_entry_token(context)
         uki = context.root / finalize_uki_path(
             context, finalize_bootloader_entry_format(context, kver, token)


daandemeyer:
How will it be built for the netesp? We don't install a kernel in the netesp image and it's gated on a kernel being installed in the image.

bluca:
The UnifiedKernelmages= setting is just ignored when building ESPs, which seems like a bug to me. It can be reproduced by having a config file that lists a kernel package

daandemeyer:
The `UnifiedKernelImages=` isn't quite right for this since if set to `none`, we document that type1 entries will be used instead. This isn't quite hooked up for the `esp` output yet, but we might want to add something like that in the future.

Hooking this up to the `Bootable=` setting seems like a better option. I've opened https://github.com/systemd/mkosi/pull/3876 to implement it like that. Note that currently the `netesp` profile sets `Bootable=yes` in particleos and we'll need to change that to `no` to make this work. I double checked and it builds fine with `Bootable=no`.

bluca:
Dunno, it seems a bit strange to have a setting called UnifiedKernelImages that doesn't control whether UnifiedKernelImages are used for just one out of a number of cases

daandemeyer:
> Dunno, it seems a bit strange to have a setting called UnifiedKernelImages that doesn't control whether UnifiedKernelImages are used for just one out of a number of cases

I completely agree that we should implement support for it for the `esp` output, but the important thing is that if it is disabled, we should use type1 entries, just like we do for the `disk` output when the setting is disabled. So even if we implemented this, it wouldn't solve your problem, since you want neither UKIs or type1 entries. Hence why I implemented support for `Bootable=no` which actually solves your problem.

PR #3876, opened 2025-08-24
Make Bootable= determine whether we build a UKI for esp images

daandemeyer:
Currently, we build a UKI for ESP images only if a kernel is installed. Let's make this a bit more flexible by hooking it up to the Bootable= setting. If Bootable=no, then we won't add a UKI to the esp regardless if a kernel is installed in the image or not.


behrmann (commented):
Looks reasonable, but this should be mentioned in the man page as well.

behrmann (approved):

PR #3877, opened 2025-08-25
kmod: Beef up firmware symlink chasing

DaanDeMeyer:
- It turns out we need to handle absolutely symlinks after all as /usr/lib/firmware/regulatory.db is a symlink to /etc/alternatives in Debian and derivatives.
- Our previous implementation didn't handle cases where a symlink target consisted out of two parts e.g. if C -> A/B then we wouldn't try to resolve A.

Fix both issues by switching to a minimal implementation of the chase() function in systemd.

Follow up for 221293e33defe30b111bc13957a0d52d2ea1c45b

### DIFF
--- a/mkosi/kmod.py
+++ b/mkosi/kmod.py
@@ -10,10 +10,10 @@ from collections.abc import Iterable, Iterator, Reversible
 from pathlib import Path
 
 from mkosi.context import Context
-from mkosi.log import complete_step, die, log_step
+from mkosi.log import complete_step, log_step
 from mkosi.run import chroot_cmd, run
 from mkosi.sandbox import chase
-from mkosi.util import chdir, parents_below, path_and_parents_below
+from mkosi.util import chdir, parents_below
 
 
 def loaded_modules() -> list[str]:
@@ -281,8 +281,7 @@ def resolve_module_dependencies(
 
                 glob = "" if value.endswith("*") else ".*"
 
-                for fw in Path("usr/lib/firmware").glob(f"{value}{glob}"):
-                    firmware.add(fw)
+                firmware.update(Path("usr/lib/firmware").glob(f"{value}{glob}"))
 
             elif key == "name":
                 # The file names use dashes, but the module names use underscores. We track the names in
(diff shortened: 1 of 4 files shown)


mkosi/kmod.py
```diff
@@ -380,25 +380,34 @@ def gen_required_kernel_modules(
...
+                continue
+
+            firmware.add(current.relative_to(context.root))
 
-            while p.is_symlink():
-                target = p.readlink()
-                if target.is_absolute():
-                    die(f"Found unexpected absolute symlink {target} in {firmwared}")
+            target = current.readlink()
+            parts = list(target.parts)
```
behrmann: Making the tuple a list here shouldn't be necessary
```suggestion
            parts = target.parts
```

mkosi/kmod.py
```diff
@@ -380,25 +380,34 @@ def gen_required_kernel_modules(
...
-            # cpio and it interprets intermediary path components that are symlinks as regular directories.
-            # Additionally, intermediary path components that are symlinks also mess up the path
-            # deduplication we want to get by putting all the firmware to add in a set as the paths will hash
-            # differently even though they're the same paths after resolution.
-            p = p.parent.resolve(strict=True).joinpath(p.name)
+        todo = list(reversed(fw.parts))
+        current = context.root
+        while todo:
+            part = todo.pop()
+            if part == ".":
```
behrmann: `fw` is guaranteed to only contain relative paths, since `filter_kernel_modules` returns relative paths and `filter_firmware` keeps them that way, but adding
```suggestion
            if part == "." or part == "/":
```
might be a good idea anyway, since any stray `/` would otherwise clobber `current` to the real root otherwise. This also makes this easier to read, since I had to go back to the above functions to check they don't return absolute paths.

Handling this here, could also allow rewriting the check below as
```python
current = context.root if target.is_absolute() else current.parent
```

behrmann (approved):

mkosi/kmod.py
```diff
@@ -276,9 +276,13 @@ def resolve_module_dependencies(
...
-                glob = "" if value.endswith("*") else "*"
-                fw = [f for f in Path("usr/lib/firmware").glob(f"{value}{glob}")]
-                firmware.update(fw)
+                if (Path("usr/lib/firmware") / value).exists():
+                    firmware.add(Path("usr/lib/firmware") / value)
+
+                glob = "" if value.endswith("*") else ".*"
+
+                for fw in Path("usr/lib/firmware").glob(f"{value}{glob}"):
+                    firmware.add(fw)
```
```suggestion
                firmware.union(fw for fw in Path("usr/lib/firmware").glob(f"{value}{glob}"))
```

PR #3879, opened 2025-08-26
Replace RuntimeHome= with BindUser=

daandemeyer:
Let's replace our home grown RuntimeHome= option with BindUser= to
mimick systemd-nspawn's --bind-user= option. systemd-vmspawn will soon
learn to support --bind-user= as well and we'll enable the option for
it when that happens as well.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -2105,7 +2105,7 @@ class Config:
     runtime_scratch: ConfigFeature
     runtime_network: Network
     runtime_build_sources: bool
-    runtime_home: bool
+    bind_user: bool
     unit_properties: list[str]
     ssh_key: Optional[Path]
     ssh_certificate: Optional[Path]
@@ -4009,11 +4009,11 @@ SETTINGS: list[ConfigSetting[Any]] = [
         scope=SettingScope.main,
     ),
     ConfigSetting(
-        dest="runtime_home",
+        dest="bind_user",
         metavar="BOOL",
         section="Runtime",
         parse=config_parse_boolean,
-        help="Mount current home directory to /root when booting the image",
+        help="Bind current user from host into container or virtual machine",
         scope=SettingScope.main,
     ),
     ConfigSetting(
@@ -5610,7 +5610,7 @@ def summary(config: Config) -> str:
                     Runtime Scratch: {config.runtime_scratch}
                     Runtime Network: {config.runtime_network}
               Runtime Build Sources: {config.runtime_build_sources}
-  Runtime Home or Working Directory: {yes_no(config.runtime_home)}
+                          Bind User: {yes_no(config.bind_user)}
                     Unit Properties: {line_join_list(config.unit_properties)}
                     SSH Signing Key: {none_to_none(config.ssh_key)}
                     SSH Certificate: {none_to_none(config.ssh_certificate)}
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -54,7 +54,6 @@ from mkosi.user import INVOKING_USER, become_root_in_subuid_range, become_root_i
 from mkosi.util import (
     PathString,
     StrEnum,
-    current_home_dir,
     flock,
     flock_or_die,
     groupby,
@@ -1146,6 +1145,9 @@ def run_qemu(args: Args, config: Config) -> None:
     ):
         die("SecureBootCertificate= must be configured to use FirmwareVariables=custom|microsoft-mok")
 
+    if config.bind_user:
+        die("mkosi qemu does not support --bind-user=")
+
     # After we unshare the user namespace to sandbox qemu, we might not have access to /dev/kvm or related
     # device nodes anymore as access to these might be gated behind the kvm group and we won't be part of the
     # kvm group anymore after unsharing the user namespace. To get around this, open all those device nodes
@@ -1216,7 +1218,6 @@ def run_qemu(args: Args, config: Config) -> None:
     if (
         config.runtime_trees
         or config.runtime_build_sources
-        or config.runtime_home
         or config.output_format == OutputFormat.directory
     ):
         shm = ["-object", f"memory-backend-memfd,id=mem,size={config.ram // 1024**2}M,share=on"]
--- a/tests/test_json.py
+++ b/tests/test_json.py
@@ -108,6 +108,7 @@ def test_config() -> None:
             "BaseTrees": [
                 "/hello/world"
             ],
+            "BindUser": true,
             "BiosBootloader": "none",
             "Bootable": "disabled",
             "Bootloader": "grub",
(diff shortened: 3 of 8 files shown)


mkosi/qemu.py
```diff
@@ -1404,13 +1421,16 @@ def add_virtiofs_mount(
...
-                credentials["fstab.extra"] = ""
+            if not (credentials / "fstab.extra").exists():
+                fstab = ""
+            else:
+                fstab = (credentials / "fstab.extra").read_text()
 
-            if credentials["fstab.extra"] and not credentials["fstab.extra"][-1] == "\n":
-                credentials["fstab.extra"] += "\n"
+            if fstab and not fstab[-1] == "\n":
+                fstab += "\n"
```
behrmann, line 1427: Extra whitespace at the end shouldn't hurt or does it? The logic could be omitted in favour of
```python
    fstab = (credentials / "fstab.extra").read_text() + "\n"
```

mkosi/qemu.py
```diff
@@ -1503,25 +1519,31 @@ def add_virtiofs_mount(
...
+        smbiosdir = Path(stack.enter_context(tempfile.TemporaryDirectory(prefix="mkosi-smbios-")))
+
+        for p in credentials.iterdir():
+            payload = base64.b64encode(p.read_text().encode()).decode()
+
             if config.architecture.supports_smbios(firmware):
-                cmdline += ["-smbios", f"type=11,value=io.systemd.credential.binary:{k}={payload}"]
+                with (smbiosdir / p.name).open("w") as f:
+                    f.write(f"io.systemd.credential.binary:{p.name}=")
+                    f.write(payload)
```
behrmann: Fewer encoding trips
```python
            payload = base64.b64encode(p.read_binary())

            if config.architecture.supports_smbios(firmware):
                with (smbiosdir / p.name).open("wb") as f:
                    f.write(f"io.systemd.credential.binary:{p.name}=".encode())
                    f.write(payload)

File: mkosi/qemu.py [hunk 3]
```diff
@@ -1503,25 +1519,31 @@ def add_virtiofs_mount(
...
-                f.write(v)
-                f.flush()
-                cmdline += ["-fw_cfg", f"name=opt/io.systemd.credentials/{k},file={f.name}"]
+            elif config.architecture.supports_fw_cfg() and len(p.name) <= 55 - len(
+                "opt/io.systemd.credentials/"
+            ):
+                cmdline += ["-fw_cfg", f"name=opt/io.systemd.credentials/{p.name},file={p}"]
             elif kernel:
-                kcl += [f"systemd.set_credential_binary={k}:{payload}"]
+                kcl += [f"systemd.set_credential_binary={p.name}:{payload}"]
```
behrmann: This is a footgun:
```
>>> bar=b"bar"
>>> f"foo={bar}"
"foo=b'bar'"
```
i.e. when putting a byte into an f-string, this will get stringified in an unexpected way
```suggestion
                kcl += [f"systemd.set_credential_binary={p.name}:{payload.decode()}"]
```
(Obviously only in the context of the above comment)

tests/test_json.py
behrmann, line 1: This also needs the corresponding change in the JSON.

mkosi/qemu.py
daandemeyer: I'm pretty sure I added this for a reason back then, so I'd rather keep it like this

behrmann: I checked and it was added in #2826 and I didn't ask about it back then. Looking at `fstab-generator.c` it does a while loop over `getmnt` and looking at the implementation in glibc that's mostly a while loop over `get_mnt_entry` which skips empty and comment lines.

But I'm also fine with keeping it as is, since I only looked at the implementations and didn't run it.

mkosi/config.py
```diff
@@ -1225,17 +1228,11 @@ def config_parse_dict(value: Optional[str], old: Optional[dict[str, str]]) -> Op
                     if p.is_dir():
                         continue
 
-                    if os.access(p, os.X_OK):
-                        new[p.name] = run([p], stdout=subprocess.PIPE, env=os.environ).stdout
-                    else:
-                        new[p.name] = p.read_text()
+                    new[p.name] = p.absolute()
```
behrmann: with the credential content generation having moved this branch could be written as
```python
            if Path(value).is_dir():
                new.update(
                    {
                       p.name: p.absolute()
                       for p in sorted(Path(value).iterdir())
                       if p.is_dir()
                    }
                )
```

mkosi/qemu.py
```diff
@@ -1503,25 +1516,31 @@ def add_virtiofs_mount(
...
+                cmdline += ["-smbios", f"type=11,path={smbiosdir / p.name}"]
             # qemu's fw_cfg device only supports keys up to 55 characters long.
-            elif config.architecture.supports_fw_cfg() and len(k) <= 55 - len("opt/io.systemd.credentials/"):
-                f = stack.enter_context(tempfile.NamedTemporaryFile(prefix="mkosi-fw-cfg-", mode="w"))
-                f.write(v)
-                f.flush()
-                cmdline += ["-fw_cfg", f"name=opt/io.systemd.credentials/{k},file={f.name}"]
+            elif config.architecture.supports_fw_cfg() and len(p.name) <= 55 - len(
+                "opt/io.systemd.credentials/"
+            ):
```
behrmann: The way ruff breaks this makes this harder to read than necessary, how about
```suggestion
            elif config.architecture.supports_fw_cfg() and len(f"opt/io.systemd.credentials/{p.name}") <= 55:
```

This should probably also throw an error or warning if the combined name is longer than 55 characters, otherwise this is silently dropped and people will wonder why.

mkosi/config.py
```diff
@@ -3996,11 +3993,11 @@ def parse_kernel_module_filter_regexp(p: str) -> str:
         scope=SettingScope.main,
     ),
     ConfigSetting(
-        dest="runtime_home",
```
behrmann, line 4019: Is there a reason to omit `compat_longs` and `compat_names` (or at least the former)?

mkosi/config.py
```diff
@@ -3996,11 +3993,11 @@ def parse_kernel_module_filter_regexp(p: str) -> str:
...
+        dest="bind_user",
```
behrmann, line 4012: Also needs the corresponding change in `mkosi.md`, to change `RuntimeHome=` for `BindUser=`.

mkosi/qemu.py
daandemeyer: Let's fix the silent dropping in another PR, that's a preexisting problem.

mkosi/config.py
daandemeyer: The option does something totally different now, I don't think that warrants a compat name.

mkosi/config.py
```diff
@@ -1216,34 +1216,33 @@ def config_match_version(match: str, value: str) -> bool:
...
-                        new[p.name] = run([p], stdout=subprocess.PIPE, env=os.environ).stdout
-                    else:
-                        new[p.name] = p.read_text()
+                new.update(
+                    {
+                        p.name: p.absolute()
+                        for p in sorted(Path(value).iterdir())
+                        if not p.is_dir()
+                    }
+                )
```
behrmann: ruff's complaining about this now, since it's actually short enough to fit on a line, missed that.

PR #3887, opened 2025-09-01
Various package fixes

DaanDeMeyer:


mkosi/resources/mkosi-tools/mkosi.conf.d/debian-kali-ubuntu/mkosi.conf
```diff
@@ -17,6 +17,7 @@ Packages=
         libseccomp2
         libtss2-dev
         policycoreutils
+        python3
```
behrmann, line 20: We should probably also do `python-is-python3` here, since then we have the `python -> python3` symlink across all distros, I think.

behrmann (approved):

mkosi/resources/mkosi-tools/mkosi.conf.d/debian-kali-ubuntu/mkosi.conf
daandemeyer: That means we'd be messing a bit too much with system configuration when this is passed to `mkosi dependencies | apt ...` so let's not do that.

behrmann:
arch:debian, arch:fedora, centos:debian, centos:fedora, ubuntu:debian, ubuntu:fedora, debian:ubuntu, debian:arch, and fedora: ubuntu end up in emergency.target because the initrd tries to look for plymouth.

behrmann:
fedora:fedora on ppc64 fails because ncdu is not available.

PR #3888, opened 2025-09-02
apt: Generate Release file

DaanDeMeyer:
Pinning by origin can't quite work if we don't generate a release file, so let's make sure we generate one which was lost when we switched to apt-ftparchive.

### DIFF
--- a/mkosi/installer/apt.py
+++ b/mkosi/installer/apt.py
@@ -292,6 +292,27 @@ class Apt(PackageManager):
                 ),
             )  # fmt: skip
 
+        metadata = [
+            "Origin=mkosi",
+            f"Architectures={context.config.distribution.architecture(context.config.architecture)}",
+            "Description=mkosi local repository",
+        ]
+
+        with (context.repository / "Release").open("wb") as f:
+            run(
+                [
+                    "apt-ftparchive", "release", ".",
+                    *(f"-o=APT::FTPArchive::Release::{m}" for m in metadata),
+                ],
+                stdout=f,
+                sandbox=context.sandbox(
+                    options=[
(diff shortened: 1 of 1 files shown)


mkosi/installer/apt.py
```diff
@@ -292,6 +292,27 @@ def createrepo(cls, context: Context) -> None:
...
+            "Origin=mkosi",
+            f"Architectures={context.config.distribution.architecture(context.config.architecture)}",
+            "Description=mkosi local repository",
+        ]
+
+        with (context.repository / "Release").open("wb") as f:
+            run(
+                [
+                    "apt-ftparchive", "release", ".",
+                    *([f"-o=APT::FTPArchive::Release::{m}" for m in metadata]),
```
behrmann: Either alone will do already
```suggestion
                    *(f"-o=APT::FTPArchive::Release::{m}" for m in metadata),
```
or
```suggestion
                    *[f"-o=APT::FTPArchive::Release::{m}" for m in metadata],
```

daandemeyer:
> LGTM, but since we're writing a release file ourselves now, I wonder whether we could make the Suites: key in mkosi-local.sources something more recognizable, since Suite can also be passed in via APT::FTPArchive::Release

Is there a benefit to this?

behrmann:
> Is there a benefit to this?

I was about to say, being more obvious, since `./` is far from what one usually sees there. But after going through `man sources.list` again and finally seeing
> suite can specify an exact path, in which case the components must be omitted and suite must end with a slash (/).

the current form is preferable, since we don't have to change anything about the repo layout. 

I do wonder now, though, whether the suite `./` should also be specified for the generation of the release file.

PR #3889, opened 2025-09-02
apt: fix installing local packages when they are older than what the distro provides

bluca:
Need to allow downgrades in case the package is already installed (eg: libsystemd0), and also need to use "" as the origin so that it actually matches the local repository


behrmann:
Can you split that up? Allowing downgrades is reasonable on its own, but for the release #3888 is probably the better approach.

daandemeyer:
#3888 doesn't work on its own.

Anyway, I'm pretty sure the other distros have the same problem, we should implement it for all package managers. I have a branch for that lying around somewhere.

bluca:
> Can you split that up? Allowing downgrades is reasonable on its own, but for the release #3888 is probably the better approach.

That one doesn't work, the preferences is still wrong and doesn't match

PR #3890, opened 2025-09-03
Add support for downgrading when installing volatile packages

DaanDeMeyer:
Replaces #3889


bluca:
Have you tested this? It doesn't fix the preferences file so I'm pretty sure it won't work either

daandemeyer:
This is my branch from months ago which I did test back then with `reprepro`.

This is meant to go in together with https://github.com/systemd/mkosi/pull/3888 which should fix the issue you mentioned.

bluca:
When I tested that one the preferences still didn't match what actually is used

PR #3891, opened 2025-09-08
Don't add ncdu for ppc64-le on Fedora

behrmann:


mkosi.tools.conf/mkosi.conf.d/fedora-not-ppc64le.conf
```diff
@@ -0,0 +1,9 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+[Match]
+Distribution=fedora
+Architecture=!ppc64-le
+
+[Content]
+Packages=
```
daandemeyer: Please name the file ncdu.conf and put everything inside a fedora/ directory

mkosi.tools.conf/mkosi.conf.d/fedora-not-ppc64le.conf
```diff
@@ -0,0 +1,9 @@
...
+        ncdu
```
daandemeyer: This only installs ncdu on Fedora?

Please rename the file to ncdu.conf and change the match section to

```conf
[Match]
Distribution=|!fedora
Packages=|!ppc64-le
```

behrmann: Logic is hard. 

I pushed
```
[TriggerMatch]
Distribution=!fedora

[TriggerMatch]
Distribution=fedora
Architecture=!ppc64-le
``` 
which in effect is also `not fedora or not ppc64-le`, but a bit more explicit about the cases covered.

mkosi.tools.conf/mkosi.conf.d/fedora-not-ppc64le.conf
behrmann: I don't think it makes sense to put it into a subdir for fedora, since it installs ncdu for all distros, so having something affecting other distros in a fedora subdirectory would be a bit unexpected, so I kept it top-level as `ncdu.conf`.

PR #3893, opened 2025-09-11
ci: Disable PPC jobs temporarily

daandemeyer:
Something in the PPC runners is broken causing all CI jobs to get cancelled when trying to rerun CI jobs. Let's disable the PPC jobs temporarily until the issues with the underlying runners are fixed so the rest of CI can run again and we can merge PRs again.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,12 +159,13 @@ jobs:
           - distro: ubuntu
             tools: ubuntu
             runner: ubuntu-24.04-arm
-          - distro: fedora
-            tools: fedora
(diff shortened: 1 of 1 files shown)


daandemeyer:
The remaining failures are due to a missing backport of https://github.com/systemd/systemd/commit/ec1d031f3de02f84beca89e2b402d085fba62be4 in ubuntu, opensuse and centos. opensuse should move to 257.9 very soon (https://build.opensuse.org/request/show/1303345). ubuntu will backport that patch in a few weeks with the next SRU. No timeline yet on centos but hopefully won't take too long either.

Merging

PR #3894, opened 2025-09-12
Add note on Encrypt=yes to Passphrase= docs

### LINKED ISSUE #3886: Setting `Passphrase=` does not actually enable disk encryption

v25.3

### Used host distribution

Debian Trixie

### Used target distribution

Debian Trixie

### Linux kernel version used

6.12.41+deb13-amd64

x86_64

Setting `Passphrase=my_file_with_password.txt` does not actually enable disk encryption. The image is bootable.

The mkosi man page doesn't actually say what `Passphrase=` _does_. It says the following:

> Passphrase=, --passphrase= : Specify the path to a file containing the passphrase to use for LUKS encryption. It should contain the passphrase literally, and not end in a newline character (i.e. in the same format as cryptsetup and /etc/crypttab expect the passphrase files). The file must have an access mode of 0600 or less.

It says "the passphrase to use for LUKS encryption". The strings "LUKS" and "encryption" show up in the mkosi manpage in the paragraph quoted above and one more place:

> The mkosi.passphrase file provides the passphrase to use when LUKS encryption is selected. It should contain the passphrase literally, and not end in a newline character (i.e. in the same format as cryptsetup and /etc/crypttab expect the passphrase files). The file must have an access mode of 0600 or less.

That's it.

This is not enough information to know how to use mkosi to set up disk encryption.  Or to know what setting `Passphrase=` will end up doing.

The sad part is that I know how to set up disk encryption _without_ mkosi. I have created a bootable USB stick with hand-written shell scripts that set up LUKS disk encryption (using `/etc/crypttab`, `cryptsetup`, `systemd-cryptsetup` etc) and then install a full Debian Trixie using `debootstrap` which will correctly prompt for a disk password on boot (and successfully unlock and proceed to login). 

I've spent the last couple of days trying to replace this collection of shell scripts with mkosi and I can't get disk encryption to work. I know _what_ needs to be set up, but I don't know which magical mkosi config incantations I need.

So I think this is largely a mkosi documentation issue.

(For the record, I have _partially_ moved forward from this point; it's clear to me now that `Encrypt=key-file` in a systemd-repart file [plus other magic] is necessary, but please note that _nothing_ in mkosi's docs told me I needed to use that. I stumbled upon it while reading the `repart.d` man page. And I _still_ haven't figured out the right repart + mkosi config incantations to make mkosi build an encrypted image. And I _know_ what needs to be done! XD I can't imagine how someone who doesn't know how disk encryption on Linux works could set it it up with mkosi...)

Thank you for working on mkosi. There's so much that's awesome about this tool. :)

### Used mkosi config

```ini
[Distribution]
Distribution=debian
Release=stable

[Output]
Format=disk

[Validation]
# The contents of this file are "changeme", without the quotes
Passphrase=disk-enc-password.txt

[Content]
RootPassword=changeme
Bootable=yes
Packages=
  bash
  coreutils
  cryptsetup
  cryptsetup-initramfs
  grub*-
  libpam-runtime
  linux-image-amd64
  login
  systemd
  systemd-boot
  systemd-boot-efi
  systemd-boot-tools

[Runtime]
Firmware=uefi
Ephemeral=true
```

### mkosi output

```sh
GitHub says content is too big to be included here so I put it up as a public Gist at 
https://gist.github.com/Valloric/3c0df79bc818928240ab3e94d095ad6f
```

daandemeyer:
Fixes #3886

### DIFF
--- a/mkosi/resources/man/mkosi.1.md
+++ b/mkosi/resources/man/mkosi.1.md
@@ -1344,6 +1344,11 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     `/etc/crypttab` expect the passphrase files). The file must have an
     access mode of 0600 or less.
 
+    Note that this setting on its own does not enable any encryption.
+    You also have to add one or more partition definition files to
+    `mkosi.repart/` with `Encrypt=key-file` to add encrypted partitions
+    to the image.
+
 `Checksum=`, `--checksum=`
 :   Generate a `<output>.SHA256SUMS` file of all generated artifacts
     after the build is complete.


mkosi/resources/man/mkosi.1.md
```diff
@@ -1344,6 +1344,10 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     `/etc/crypttab` expect the passphrase files). The file must have an
     access mode of 0600 or less.
 
+    Note that this setting on its own does not enable any encryption. You
+    also have to add one or more partition definition to `mkosi.repart/`
```
behrmann: Missing the "files" or a plural s
```suggestion
    also have to add one or more partition definition files to `mkosi.repart/`
```

Valloric:
Thank you! 🙂

PR #3899, opened 2025-09-12
sandbox: Rework ctypes hack

daandemeyer:
Now that libffi marks the trampoline file descriptor opened by ctypes
as O_CLOEXEC, we can move file descriptor packing to the very end of
setting up the sandbox and close all file descriptors marked with
O_CLOEXEC already which will also close the libffi trampoline file
descriptor.

### DIFF
--- a/mkosi/sandbox.py
+++ b/mkosi/sandbox.py
@@ -34,6 +34,7 @@ ENOENT = 2
 ENOSYS = 38
 F_DUPFD = 0
 F_GETFD = 1
+FD_CLOEXEC = 1
 FS_IOC_GETFLAGS = 0x80086601
 FS_IOC_SETFLAGS = 0x40086602
 FS_NOCOW_FL = 0x00800000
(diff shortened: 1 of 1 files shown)


mkosi/sandbox.py
```diff
@@ -560,7 +510,44 @@ def is_relative_to(one: str, two: str) -> bool:
...
+        if flags < 0:
+            return -1
+
+        if flags & FD_CLOEXEC:
+            os.close(fd)
+            return -1
+
+        return fd
+
+    fds = sorted(fd for fd in fds if process_fd(fd) >= 0)
```
behrmann: This doesn't use the returned fd's, so this could be a bit simpler
```suggestion
    def process_fd(fd: int) -> bool:
        flags = libc.fcntl(fd, F_GETFD, 0)
        if flags < 0:
            return False

        if flags & FD_CLOEXEC:
            os.close(fd)
            return False

        return True

    fds = sorted(fd for fd in fds if process_fd(fd))
```

PR #3903, opened 2025-09-15
Fix if check in install_systemd_boot()

daandemeyer:
A generator object always evaluates to True, we have to wrap it with
any() to check if it contains anything.


behrmann (approved):

behrmann:
Good catch, what a footgun.

PR #3781, opened 2025-06-23
Add support for postmarketOS

craftyguy:
This adds initial support for postmarketOS in mkosi, continuing earlier work from @septatrix and @kcxt.

Merging /usr in Alpine Linux (which pmOS is based on) is still WIP but is moving along. I'm submitting these patches now for some initial feedback, but I don't expect them to be merged until /usr is merged in postmarketOS.

PR #3908, opened 2025-09-21
Add experimenta alpine/apk support

yuwata:


yuwata:
I did not notice there is #3781...

daandemeyer:
@yuwata Yeah that's unfortunate. I'm hesitant to merge alpine support in mkosi before systemd is packaged and at least nominally supported there, but we can use postmarketos in the meantime instead, which unfortunately makes this PR redundant with the postmarketos one :(.

fossdd:
> @yuwata Yeah that's unfortunate. I'm hesitant to merge alpine support in mkosi before systemd is packaged and at least nominally supported there

FTR I'm working on packaging systemd for Alpine but actually merging systemd into Alpine would need to be after the musl-PR got merged and released. (And there might be some communication with other developers I need to deal with first :))

yuwata:
Note, with this and https://github.com/systemd/systemd/pull/38825, TEST-01-BASIC, TEST-03-JOBS, and TEST-07-PID1 are green on nspawn. So, technically, very basic level of musl support is already completed.
I am also optimistic about 'political' concern. The musl support code in systemd is cleanly separated from the main code. Hence, if one dislike musl, he/she can just ignore it.

yuwata:
@fossdd If you'd like, see the last commit in https://github.com/systemd/systemd/pull/38825. Any suggestion and comments are welcome.

daandemeyer:
@yuwata Should we close this?

yuwata:
Yes. Sorry for the noise.

PR #3910, opened 2025-09-23
Add cxl config option support

gourryinverse:
x86 and ARM machines now support CXL devices, but these require cxl=on and maxmem arguments in the machine and memory configuration lines respectively.  Add support for these options.


mkosi/config.py
```diff
@@ -4113,6 +4118,17 @@ def parse_kernel_module_filter_regexp(p: str) -> str:
...
     ),
+    ConfigSetting(
+        dest="maxmem",
+        name="MAXMEM",
+        metavar="BYTES",
+        section="Runtime",
+        parse=config_parse_bytes,
+        help="Configure guest's MAXMEM size",
+        compat_longs=("--qemu-maxmem",),
+        compat_names=("QemuMaxMem",),
```
behrmann: `compat_` is for, well, compat. The other ones had different names in the past, but new ones, shouldn't get these.

mkosi/config.py
```diff
@@ -4124,6 +4140,17 @@ def parse_kernel_module_filter_regexp(p: str) -> str:
...
     ),
+    ConfigSetting(
+        dest="cxl",
+        name="CXL",
+        metavar="BOOLEAN",
+        section="Runtime",
+        parse=config_parse_boolean,
+        help="Enable CXL device support",
+        compat_longs=("--qemu-cxl",),
+        compat_names=("QemuCxl",),
```
behrmann: same

mkosi/config.py
```diff
@@ -4113,6 +4118,17 @@ def parse_kernel_module_filter_regexp(p: str) -> str:
         compat_names=("QemuMem",),
         scope=SettingScope.main,
     ),
+    ConfigSetting(
+        dest="maxmem",
+        name="MAXMEM",
```
daandemeyer: So we have `RAM=` to set memory, I figure that means this either needs to be `MaxRAM=` or `MaxMem=`, I'm leaning towards `MaxRAM=` to be consistent with the existing option name but I'm not sure if that really matches with CXL so I'd be fine with `MaxMem=` as well

mkosi/config.py
gourryinverse: removed

mkosi/config.py
gourryinverse: removed

gourryinverse:
> Both of these will also need to be added to the man page.

done.

mkosi/resources/man/mkosi.1.md
```diff
@@ -1744,10 +1744,18 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 `RAM=`, `--ram=`
 :   Configures the amount of RAM assigned to the guest when booting a virtual machine. Defaults to `2G`.
 
+`MAXMEM=`, `--maxmem=`
```
daandemeyer: This should be `MaxRAM=` or `MaxMem=`, not sure which one is preferred, `MaxRAM=` matches the existing `RAM=` option but might not fit CXL semantics.

mkosi/config.py
```diff
@@ -5359,6 +5382,8 @@ def line_join_list(array: Iterable[object]) -> str:
 
 
 def format_bytes(num_bytes: int) -> str:
+    if not num_bytes:
+        return ""
```
daandemeyer: Please drop this change and use `format_bytes_or_none()` instead.

mkosi/config.py
gourryinverse: the setting in qemu is literally `-m NG,maxmem=NG' and it doesn't necessarily map to RAM - it could be used by device memory.  So I will change it to MaxMem

PR #3913, opened 2025-09-24
Various postmarketos fixes

DaanDeMeyer:


yuwata (approved):
LGTM.

mkosi.conf.d/postmarketos.conf
```diff
@@ -5,12 +5,8 @@ Distribution=postmarketos
 
 [Content]
 Packages=
-        musl-utils
         openssh-server
         openssh-server-pam-systemd
-        postmarketos-base
-        postmarketos-base-systemd
-        util-linux
```
craftyguy, line 13: Why are these being moved? It seems like it would break building non-vm images, since these packages are required to install some very core things needed in pmOS (like systemd and other config)

yuwata:
BTW, no integration tests for postmarketos yet?

mkosi.conf.d/postmarketos.conf
daandemeyer: This is just for the default image in the mkosi repo. This default image always includes the mkosi-vm config, which is the config to build an image that can be booted in a VM. So this change makes sure these packages are always installed in more cases than they are now.

If you build an image without config outside of the mkosi repo, we don't install any packages at all except the base filesystem layout, this is by design. You don't want systemd in a sysext.

daandemeyer:
> BTW, no integration tests for postmarketos yet?

I already found out the alpine linux-virt package is missing required kernel modules so it's going to take a while before we can get those working.

mkosi.conf.d/postmarketos.conf
craftyguy: Ahhhh ok, thank you for clarifying :)

craftyguy:
> > BTW, no integration tests for postmarketos yet?
> 
> I already found out the alpine linux-virt package is missing required kernel modules so it's going to take a while before we can get those working.

I sent a patch to add this dependency: https://gitlab.postmarketos.org/postmarketOS/pmaports/-/merge_requests/7079

PR #3916, opened 2025-09-25
mkosi-initd: install libseccomp in debian/ubuntu

bluca:
It is now a dlopen optional dep so it's only recommended by the packages, so pull it in like libcryptsetup and libfido2.


daandemeyer:
Let's just list it everywhere as in my original PR.

PR #3917, opened 2025-09-26
mkosi-initrd: add 89-ethernet.network for network profile

bluca:
networkd ships with no configuration enabled, so we need to symlink one in for it to do anything


yuwata:
Why not to enable default network config option?

bluca:
Sorry what do you mean by that precisely?

yuwata:
I mean, why not to enable -Ddefault-network=true meson option? Or, do I misunderstand the point of this PR?

daandemeyer:
@yuwata That would have to be done in Fedora, where we obviously cannot make that change for backwards compat concerns

yuwata:
Ahh, I forgot configs in mkosi repository does not build systemd but use systemd provided by distribution. Sorry for the noise.

mkosi/resources/__init__.py
```diff
@@ -81,6 +81,8 @@ def write_contents(target: Path, source: Traversable) -> Path:
         child.mkdir()
         for item in source.iterdir():
             write_contents(child, item)
+    elif source.is_symlink():
+        os.symlink(source.readlink(), child)
```
daandemeyer: Please move this first, otherwise we'll descend into directory symlinks.

DaanDeMeyer (changes requested):

mkosi/resources/__init__.py
```diff
@@ -77,7 +77,9 @@ def temporary_dir(path: Traversable) -> Iterator[Path]:
 
 def write_contents(target: Path, source: Traversable) -> Path:
     child = target.joinpath(source.name)
-    if source.is_dir():
+    if source.is_symlink():
+        os.symlink(source.readlink(), child)
```
behrmann: Since `child` is a Path, because it's derived from `target`, we don't need to resort to `os`

```suggestion
        child.symlink_to(source.readlink())
```

mkosi/resources/mkosi-initrd/mkosi.profiles/network/mkosi.extra/usr/lib/systemd/networkd/89-ethernet.network
```diff
@@ -0,0 +1 @@
+89-ethernet.network.example
```
behrmann: I don't think bookworm is a target for this, but the file changed its name in commit 9b7a624267fddc5c20bd15480e7a393d7a3b270e from `80-` to `89-`, so there might be the unlikely case where this produces a dangling symlink (pre 255).

DaanDeMeyer (approved):

mkosi/resources/mkosi-initrd/mkosi.profiles/network/mkosi.extra/usr/lib/systemd/networkd/89-ethernet.network
bluca: pretty sure most things here won't work with 255, so a danglign symlink is the least problem

behrmann:
Looking at the CI failure, there is a certain problem: The thing coming in is a traversable, since the thing in resources need not be backed by a filesystem, the package could be whatever weird thing. I'm not sure that e.g. Python wheels support symlinks (they're zipfiles that have to work on windows). While that might not be relevant when running mkosi from source, this will break zipapps or possibly distro packages where building a wheel might be an intermediate step. 

A workaround might actually be to ship the `mkosi.extra` as a tar file, which can be adressed as a single file as a resouce.

bluca:
Python used to be fun, you could assign a banana to a bathtub and it would just do it

behrmann:
Well, you can still do that, but now we just get warned before trying to take a bath in the banana. :)

mkosi/resources/__init__.py
```diff
@@ -77,7 +77,9 @@ def temporary_dir(path: Traversable) -> Iterator[Path]:
 
 def write_contents(target: Path, source: Traversable) -> Path:
     child = target.joinpath(source.name)
-    if source.is_dir():
+    if Path(source).is_symlink():
```
daandemeyer: Ah fuck the Traversable interface from python doesn't actually expose symlinks, this isn't going to work after all. When the resources come from mkosi packaged up as a zipapp this would blow up.

Forget the mkosi.extra approach, please turn this into mkosi.postinst script that creates the symlink, that'll work.

behrmann:
In the context of the initrd profile the simpler workaround might even be to duplicate the file contents here, at the same time sidestepping the renaming issue mentioned above.

daandemeyer:
That means we don't get updates, I prefer the symlink

mkosi/resources/mkosi-initrd/mkosi.profiles/network/mkosi.postinst
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+mkdir -p "$BUILDROOT/usr/lib/systemd/networkd/"
+ln -sf 89-ethernet.network.sample "$BUILDROOT/usr/lib/systemd/networkd/89-ethernet.network"
```
daandemeyer: This is amazing and all but maybe double check the directory where networkd config files are stored ;)

bluca: how tf did this work

```
$ ls i/usr/lib/systemd/network/ -l
total 96
-rw-r--r-- 1 bluca bluca  71 Mar 26  2025 73-usb-net-by-mac.link
-rw-r--r-- 1 bluca bluca 569 Jul  5 21:11 80-6rd-tunnel.link
-rw-r--r-- 1 bluca bluca 542 Jul  5 21:11 80-6rd-tunnel.network
-rw-r--r-- 1 bluca bluca 719 Jul  5 21:11 80-auto-link-local.network.example
-rw-r--r-- 1 bluca bluca 664 Jul  5 21:11 80-container-host0.network
-rw-r--r-- 1 bluca bluca 601 Jul  5 21:11 80-container-host0-tun.network
-rw-r--r-- 1 bluca bluca 656 Jul  5 21:11 80-container-vb.link
-rw-r--r-- 1 bluca bluca 692 Jul  5 21:11 80-container-vb.network
-rw-r--r-- 1 bluca bluca 607 Jul  5 21:11 80-container-ve.link
-rw-r--r-- 1 bluca bluca 823 Jul 18 12:15 80-container-ve.network
-rw-r--r-- 1 bluca bluca 586 Jul  5 21:11 80-container-vz.link
-rw-r--r-- 1 bluca bluca 808 Jul 18 12:15 80-container-vz.network
-rw-r--r-- 1 bluca bluca 613 Jul  5 21:11 80-namespace-ns.link
-rw-r--r-- 1 bluca bluca 825 Jul 18 12:15 80-namespace-ns.network
-rw-r--r-- 1 bluca bluca 607 Jul  5 21:11 80-namespace-ns-tun.link
-rw-r--r-- 1 bluca bluca 822 Jul 18 12:15 80-namespace-ns-tun.network
-rw-r--r-- 1 bluca bluca 555 Jul  5 21:11 80-vm-vt.link
-rw-r--r-- 1 bluca bluca 776 Jul 18 12:15 80-vm-vt.network
-rw-r--r-- 1 bluca bluca 452 Jul  5 21:11 80-wifi-adhoc.network
-rw-r--r-- 1 bluca bluca 664 Jul  5 21:11 80-wifi-ap.network.example
-rw-r--r-- 1 bluca bluca 602 Jul  5 21:11 80-wifi-station.network.example
lrwxrwxrwx 1 bluca bluca  27 Aug 28 17:15 89-ethernet.network -> 89-ethernet.network.example
-rw-r--r-- 1 bluca bluca 636 Jul  5 21:11 89-ethernet.network.example
-rw-r--r-- 1 bluca bluca 504 Jul  5 21:11 99-default.link
```

mkosi/resources/mkosi-initrd/mkosi.profiles/network/mkosi.postinst
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+mkdir -p "$BUILDROOT/usr/lib/systemd/network/"
+ln -sf 89-ethernet.network.sample "$BUILDROOT/usr/lib/systemd/network/89-ethernet.network"
```
behrmann: Don't you mean `.example` instead?

mkosi/resources/mkosi-initrd/mkosi.profiles/network/mkosi.postinst
bluca: it didn't rebuild the profile nor clear out the old image so was looking at the wrong thing hhhh

PR #3918, opened 2025-09-26
Postmarketos fixes

DaanDeMeyer:


craftyguy:
thank you for figuring out the weird rootfs mounting issue! lgtm!

I filed a request with apk-tools about adding the ability to cache packages separately: https://gitlab.alpinelinux.org/alpine/apk-tools/-/issues/11144

mkosi/__init__.py
```diff
@@ -3962,24 +3962,28 @@ def copy_repository_metadata(config: Config, dst: Path) -> None:
...
                     subdst.mkdir(parents=True, exist_ok=True)
 
-                def sandbox(
-                    *,
-                    options: Sequence[PathString] = (),
-                ) -> AbstractContextManager[list[PathString]]:
-                    return config.sandbox(options=[*options, *exclude])
+                caches = config.distribution.package_manager(config).package_subdirs(cachedir)
+
+                # If there no package cache subdirectories, we'll always operate on the package cache
```
yuwata: there 'are' no package cache subdirectories

daandemeyer:
Added one more commit to stop storing repository metadata in the shared package cache which doesn't save too much and allows drastically simplifying the whole setup. Though apk still gets to store its repository metadata in the shared package cache until the issue to introduce a separate package cache directory option is resolved.

behrmann:
Ah, was still reviewing the version from before the last push, the typo's also fixed and the series easier in this order.

PR #3922, opened 2025-09-27
Treat default initrd as a regular subimage

DaanDeMeyer:
Currently, because we build the default initrd as a substep of building
a regular image, we have lots of special cased logic for it and we still
propagate settings manually from the regular image to its default initrd.

Let's streamline this by treating the default initrd as a regular image.
The only complication about making this change is that we used to build
the default initrd on demand only if a kernel was actually installed into
the image. Because we have to make the decision of whether to build the
default initrd or not way earlier now, we can't check if kernels were
installed into the image or not. Instead, we check if any known kernel
packages are listed to be installed which should be a decent enough
heuristic.

Another regression is that the default initrd won't have access to any
packages built as part of the main image build anymore. We used to rely
on this in systemd but now we build the systemd packages in a separate
build subimage and those will still be available to the default initrd
image build.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -10,6 +10,7 @@ import functools
 import getpass
 import graphlib
 import io
+import itertools
 import json
 import logging
 import math
@@ -1652,6 +1653,19 @@ class SettingScope(StrEnum):
     main = enum.auto()
     # Only passed down to tools tree, can only be configured in main image.
     tools = enum.auto()
+    # Only passed down to initrd, can only be configured in main image.
+    initrd = enum.auto()
+
+    def is_main_setting(self) -> bool:
+        return self in (SettingScope.main, SettingScope.tools, SettingScope.initrd, SettingScope.multiversal)
+
+    def removeprefix(self, setting: str) -> str:
+        if self == SettingScope.tools:
+            return setting.removeprefix("tools_tree_")
+        elif self == SettingScope.initrd:
+            return setting.removeprefix("initrd_")
+        else:
+            return setting
 
 
 @dataclasses.dataclass(frozen=True)
@@ -3165,6 +3179,7 @@ SETTINGS: list[ConfigSetting[Any]] = [
         choices=InitrdProfile.values(),
         default=[],
         help="Which profiles to enable for the default initrd",
+        scope=SettingScope.initrd,
     ),
     ConfigSetting(
         dest="initrd_packages",
@@ -3173,6 +3188,7 @@ SETTINGS: list[ConfigSetting[Any]] = [
         section="Content",
         parse=config_make_list_parser(delimiter=","),
         help="Add additional packages to the default initrd",
+        scope=SettingScope.initrd,
     ),
     ConfigSetting(
         dest="initrd_volatile_packages",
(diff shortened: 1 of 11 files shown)


mkosi/config.py
```diff
@@ -5235,19 +5326,28 @@ def parse_config(
...
         if in_box():
             config["tools_tree"] = Path(os.environ["MKOSI_DEFAULT_TOOLS_TREE_PATH"])
         else:
             tools = finalize_default_tools(context, config, configdir=configdir, resources=resources)
             config["tools_tree"] = tools.output_dir_or_cwd() / tools.output
 
+    if want_default_initrd(main):
+        initrd = finalize_default_initrd(context, config, resources=resources)
+        config["initrds"] = [
+            initrd.output_dir_or_cwd() / initrd.output if i == Path("default") else i for i in main.initrds
```
behrmann: Since this line lost some indentation now a linebreak got removed, let's add some parentheses so it's less glyph soup.
```suggestion
            (initrd.output_dir_or_cwd() / initrd.output if i == Path("default") else i) for i in main.initrds
```

mkosi/config.py
```diff
@@ -3145,7 +3148,10 @@ def parse_kernel_module_filter_regexp(p: str) -> str:
         long="--initrd",
         metavar="PATH",
         section="Content",
-        parse=config_make_list_parser(delimiter=",", parse=make_path_parser(required=False)),
+        parse=config_make_list_parser(
+            delimiter=",", parse=make_path_parser(required=False, constants=("default",))
```
```suggestion
            delimiter=",", 
            parse=make_path_parser(required=False, constants=("default",)),
```

mkosi/config.py
```diff
@@ -4618,7 +4627,8 @@ def setting_prohibited(self, setting: ConfigSetting[T]) -> bool:
         return (
             (not setting.tools and image == "tools")
             or (
-                setting.scope in (SettingScope.main, SettingScope.tools, SettingScope.multiversal)
+                setting.scope
+                in (SettingScope.main, SettingScope.tools, SettingScope.initrd, SettingScope.multiversal)
```
behrmann: This grouping could do with a name, maybe
```suggestion
                setting.scope.can_only_configured_on_main()
```

mkosi/config.py
```diff
@@ -4756,11 +4766,15 @@ def finalize_value(self, setting: ConfigSetting[T]) -> Optional[T]:
         ):
             return v
 
+        if setting.scope == SettingScope.tools:
+            fieldname = setting.dest.removeprefix("tools_tree_")
+        elif setting.scope == SettingScope.initrd:
+            fieldname = setting.dest.removeprefix("initrd_")
+        else:
+            fieldname = setting.dest
```
behrmann: Maybe too silly, but how about pushing this transformation into `SettingScope`?

mkosi/config.py
```diff
@@ -1652,6 +1653,19 @@ class SettingScope(StrEnum):
...
+    def is_main_setting(self) -> bool:
+        return self in (SettingScope.main, SettingScope.tools, SettingScope.initrd, SettingScope.multiversal)
+
+    def removeprefix(self, setting: str) -> str:
+        if self == SettingScope.tools:
+            return setting.removeprefix("tools_tree_")
+        elif self == SettingScope.initrd:
+            return setting.removeprefix("initrd_")
+        else:
+            return setting
```
behrmann, line 1668: Nit: now that it's inside a function
```suggestion

        return setting
```
would look a bit nicer.

behrmann (approved):

mkosi/config.py
daandemeyer: If there was more I'd fix this but I don't wanna wait for CI again so gonna leave this for another time

PR #3930, opened 2025-09-29
postmarketos: implement is_kernel_package

craftyguy:
There are many many kernel packages in pmOS+Alpine and the pkg names don't follow any specific pattern that sets them apart from some other non-kernel packages, so the implementation tries to exclude known pkgs that are not kernels and assume anything that doesn't match these package names/patterns are kernel packages.

### DIFF
--- a/mkosi/distributions/postmarketos.py
+++ b/mkosi/distributions/postmarketos.py
@@ -112,5 +112,21 @@ class Installer(DistributionInstaller):
 
     @classmethod
     def is_kernel_package(cls, package: str) -> bool:
-        # TODO: Cover all of postmarketos's kernel packages.
-        return package == "linux-virt"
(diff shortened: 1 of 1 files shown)


mkosi/distributions/postmarketos.py
```diff
@@ -112,5 +112,18 @@ def architecture(cls, arch: Architecture) -> str:
...
+            return False
+
+        if package in {
+            "linux-timemachine",
+            "linux-headers",
+            "linux-apfs-rw-src",
+        }:
+            return False
+
+        return True
```
daandemeyer, line 132: Should still check here that the package starts with "linux-"?

craftyguy: Oh derp, I didn't mean to leave that out... Resolved now

PR #3936, opened 2025-10-04
action: make it work when used as a submodule

davide125:
When consuming the mkosi GitHub action from a submodule (e.g. in https://github.com/davide125/arcadeos/pull/10), `.git` will be a text file, not a folder, so the check fails. Make it more generic to cover this scenario as well.


davide125:
CI failure looks unrelated

behrmann (approved):

PR #3942, opened 2025-10-05
test: add unit tests for parents_below() utility function

DaanDeMeyer:


tests/test_util.py
```diff
@@ -0,0 +1,56 @@
...
+import pytest
+
+from mkosi.util import parents_below
+
+
+def test_parents_below_basic() -> None:
+    path = Path("/a/b/c/d/e")
+    below = Path("/a/b")
+    result = parents_below(path, below)
+    assert result == [Path("/a/b/c/d"), Path("/a/b/c")]
```
behrmann: I think the `result` is a bit superfluous and makes it easier to read, since it's a repeated line that doesn't do anything
```suggestion
    assert parents_below(path, below) == [Path("/a/b/c/d"), Path("/a/b/c")]
```
The tests that raise break this pattern.

tests/test_util.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def test_parents_below_below_is_child_raises() -> None:
+    path = Path("/a/b")
+    below = Path("/a/b/c")
+    with pytest.raises(ValueError):
+        parents_below(path, below)
+
+
+def test_parents_below_root() -> None:
```
behrmann: This check should be further up, since it's basically the same as `_basic` and doesn't really fit the pattern after the raising tests.

behrmann:
Merging, since the tests have passed and this doesn't change anything but tests.

PR #3947, opened 2025-10-09
Don't unconditionally sync when PackageCacheDirectory=/var

### LINKED ISSUE #3946: partial system upgrades on Arch linux after commit fc2da84
After commit fc2da84 mkosi, when run as root, allows pacman to interact directly with /var/lib/pacman/sync instead of a copy of it as before.  This means the sync metadata will be updated every time mkosi is run as root.  This can lead to partial system upgrades on the host, which can lead to a broken system.

https://wiki.archlinux.org/title/System_maintenance#Partial_upgrades_are_unsupported

daandemeyer:
If PackageCacheDirectory=/var and a cache directory is not configured (like in mkosi-initrd), then we'd always sync repository metadata, regardless of the value of CacheOnly=. Let's fix that by disallowing CacheOnly=metadata|always if a cache directory is not configured and only syncing if CacheOnly=never|auto, regardless if a cache directory is configured or not.

Fixes #3946

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2651,6 +2651,13 @@ def check_inputs(config: Config) -> None:
                 hint="Use Profile= to configure the profile ID",
             )
 
+    if (
+        config.cacheonly not in (Cacheonly.never, Cacheonly.auto)
+        and not config.cache_dir
+        and config.package_cache_dir_or_default() != Path("/var")
+    ):
+        die(f"A cache directory must be configured in order to use CacheOnly={config.cacheonly}")
+
 
 def check_tool(config: Config, *tools: PathString, reason: str, hint: Optional[str] = None) -> Path:
     tool = config.find_binary(*tools)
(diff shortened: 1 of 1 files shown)


behrmann:
@chriscjs can you check whether this fixes your issue?

chriscjs:
I applied your patch and I get the following error:

sudo mkosi-initrd
[sudo] password for root:

Validating certificates and keys
‣ A cache directory must be configured in order to use CacheOnly=metadata

‣ "mkosi --force --directory= --format=cpio --output=initrd --output-directory=/tmp/tmp98orzyxo --extra-tree=/usr/lib/modules/6.17.1-arch1-1:/usr/lib/modules/6.17.1-arch1-1 --extra-tree=/usr/lib/firmware:/usr/lib/firmware '--remove-files=/usr/lib/firmware/*-ucode' --build-sources= --include=mkosi-initrd --kernel-modules=host --package-cache-dir=/var --cache-only=metadata --output-mode=600 --include /etc/mkosi-initrd --sandbox-tree=/tmp/tmpqza030tx --extra-tree /tmp/tmp98orzyxo/crypttab:/etc/crypttab --extra-tree=/etc/vconsole.conf:/etc/vconsole.conf" returned non-zero exit code 1.

CacheOnly is hardcoded on line 378 in initrd.py
--cache-only=metadata

daandemeyer:
@chriscjs Should work now

chriscjs:
Thanks that one worked.

PR #3950, opened 2025-10-13
Don't die when detecting multiple kernel versions

### LINKED ISSUE #3949: 5e38eec86 breaks multiple kernels

main

### Used host distribution

ArchLinux

### Used target distribution

ArchLinux

### Linux kernel version used

6.16.3-arch1-1

x86_64

I have a use-case where I install multiple (same version, different config) kernels into an image and generate different UKI:s for each one of them. This broke with commit 5e38eec86 which does not allow more than one directory in /lib/modules/.

### Used mkosi config

```ini

```

### mkosi output

```sh
5e38eec86 mkosi/__init__.py (Clayton Craft  2025-06-04 15:44:50 -0700 1303)  "Multiple kernel module directories found in /usr/lib/modules, unable to determine correct version to use"  # noqa: E501
```

ppira:
Fixes #3949

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1294,13 +1294,13 @@ def gzip_binary(context: Context) -> str:
 
 
 def kernel_get_ver_from_modules(context: Context) -> Optional[str]:
-    # Try to get version from the first dir under usr/lib/modules but fail if multiple versions are found
-    versions = [
+    # Try to get version from the first dir under usr/lib/modules, warn if multiple versions are found
+    versions = sorted(
         p.name for p in (context.root / "usr/lib/modules").glob("*") if KERNEL_VERSION_PATTERN.match(p.name)
-    ]
+    )
     if len(versions) > 1:
-        die(
-            "Multiple kernel module directories found in /usr/lib/modules, unable to determine correct version to use"  # noqa: E501
+        logging.warning(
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -1294,13 +1294,13 @@ def gzip_binary(context: Context) -> str:
...
-    # Try to get version from the first dir under usr/lib/modules but fail if multiple versions are found
+    # Try to get version from the first dir under usr/lib/modules, warn if multiple versions are found
     versions = [
         p.name for p in (context.root / "usr/lib/modules").glob("*") if KERNEL_VERSION_PATTERN.match(p.name)
     ]
     if len(versions) > 1:
-        die(
-            "Multiple kernel module directories found in /usr/lib/modules, unable to determine correct version to use"  # noqa: E501
+        logging.warning(
+            f"Multiple kernel module directories found in /usr/lib/modules, determining version from first: {versions[0]}"  # noqa: E501
```
behrmann, line 1303: That's not necessarily deterministic, since the iteration order is not guaranteed. If we want to go this way it should be
```python
    versions = sorted(
        p.name for p in (context.root / "usr/lib/modules").glob("*") if KERNEL_VERSION_PATTERN.match(p.name)
    )
````
above

daandemeyer:
This is the wrong fix, closing the PR and continuing discussion in the issue.

PR #3954, opened 2025-10-15
Add support for assert sections

daandemeyer:
Often you only want to support a single distribution. Adding a [Match] section to mkosi.conf will be very confusing for users as they will end up with an empty image. By using [Assert], they'll get a clear error that what they're doing is not supported.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -2566,7 +2566,7 @@ def parse_ini(path: Path, only_sections: Collection[str] = ()) -> Iterator[tuple
     if section and setting and value is not None:
         yield section, setting, value
 
-    if section:
+    if section and (not only_sections or section in only_sections):
         yield section, "", ""
 
 
@@ -4808,7 +4808,7 @@ class ParseContext:
 
         return default
 
-    def match_config(self, path: Path) -> bool:
+    def match_config(self, path: Path, asserts: bool = False) -> bool:
         condition_triggered: Optional[bool] = None
         match_triggered: Optional[bool] = None
         skip = False
@@ -4818,12 +4818,17 @@ class ParseContext:
         if not path.exists():
             return True
 
-        for section, k, v in parse_ini(path, only_sections=["Match", "TriggerMatch"]):
-            if not k and not v:
-                if section == "Match" and condition_triggered is False:
-                    return False
+        sections = ("Assert", "TriggerAssert") if asserts else ("Match", "TriggerMatch")
 
-                if section == "TriggerMatch":
+        for section, k, v in parse_ini(path, only_sections=sections):
+            if not k and not v:
+                if condition_triggered is False:
+                    if section == "Assert":
+                        die(f"{path.absolute()}: Trigger condition in [Assert] section was not satisfied")
+                    elif section == "Match":
+                        return False
+
+                if section in ("TriggerAssert", "TriggerMatch"):
                     match_triggered = bool(match_triggered) or condition_triggered is not False
 
                 condition_triggered = None
@@ -4833,6 +4838,7 @@ class ParseContext:
             if skip:
                 continue
 
+            raw = v
             trigger = v.startswith("|")
             v = v.removeprefix("|")
             negate = v.startswith("!")
@@ -4867,14 +4873,20 @@ class ParseContext:
             if negate:
                 result = not result
             if not trigger and not result:
-                if section == "TriggerMatch":
+                if section.startswith("Trigger"):
                     skip = True
                     condition_triggered = False
                     continue
 
+                if asserts:
+                    die(f"{path.absolute()}: {k}={raw} in [Assert] section was not satisfied")
+
                 return False
             if trigger:
                 condition_triggered = bool(condition_triggered) or result
+
+        if match_triggered is False and asserts:
+            die(f"{path.absolute()}: None of the [TriggerAssert] sections was satisfied")
 
         return match_triggered is not False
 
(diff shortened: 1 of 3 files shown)


mkosi/config.py
```diff
@@ -4867,15 +4873,21 @@ def match_config(self, path: Path) -> bool:
...
                 result = not result
             if not trigger and not result:
-                if section == "TriggerMatch":
+                if section.startswith("Trigger"):
                     skip = True
                     condition_triggered = False
                     continue
 
+                if asserts:
+                    die(f"{path}: {k}={raw} in [Assert] section was not satisfied")
```
behrmann: This should probably use `path.absolute()` as below.

mkosi/config.py
```diff
@@ -4818,12 +4818,17 @@ def match_config(self, path: Path) -> bool:
...
-        for section, k, v in parse_ini(path, only_sections=["Match", "TriggerMatch"]):
+        sections = ("Assert", "TriggerAssert") if asserts else ("Match", "TriggerMatch")
+
+        for section, k, v in parse_ini(path, only_sections=sections):
             if not k and not v:
-                if section == "Match" and condition_triggered is False:
-                    return False
+                if condition_triggered is False:
+                    if section == "Assert":
+                        die(f"{path}: Trigger condition in [Assert] section was not satisfied")
```
behrmann: This should probably also get `path.absolute()` as below.

tests/test_config.py
```diff
@@ -1578,3 +1578,66 @@ def test_subdir(tmp_path: Path) -> None:
...
+
+            [Assert]
+            Environment=ABC=QED
+            """
+        )
+
+        with pytest.raises(SystemExit):
+            parse_config()
+        with pytest.raises(SystemExit):
+            parse_config(["--image-id", "abcde"])
```
behrmann: For completeness: This should also check that 
```python
        with pytest.raises(SystemExit):
            parse_config(["--environment", "ABC=QED"])
```
raises and that
```python
        assert parse_config(["--image-id", "abcde", "--environment", "ABC=QED"])[2][-1].image_id == "abcde"
```
does not raise

tests/test_config.py
```diff
@@ -1578,3 +1578,66 @@ def test_subdir(tmp_path: Path) -> None:
...
+            """
+            [Assert]
+            ImageId=abcde
+            """
+        )
+
+        with pytest.raises(SystemExit):
+            parse_config()
+
+        parse_config(["--image-id", "abcde"])
```
behrmann, line 1598: Nit: it would be nice to not just check it doesn't raise, but to assert something positive, e.g.
```suggestion
        assert parse_config(["--image-id", "abcde"])[2][-1].image_id == "abcde"
```

mkosi/resources/man/mkosi.1.md
```diff
@@ -466,6 +466,10 @@ matches. The absence of match sections is valued as true. Logically this means:
 (⋀ᵢ Matchᵢ) ∧ (⋁ᵢ TriggerMatchᵢ)
 ```
 
+There is also support for `[Assert]` and `[TriggerAssert]` sections which
+behave identically to match sections except parsing configuration will
+fail if the assert sections are not satisfied.
```
behrmann: I think it helps to spell this out clearly
```suggestion
fail if the assert sections are not satisfied, i.e. all `[Assert]` sections 
in a file as well as at least one `[TriggertAssert]` section have to be 
satisfied or config parsing will fail. 
```

tests/test_config.py
```diff
@@ -1578,3 +1578,66 @@ def test_subdir(tmp_path: Path) -> None:
...
+        parse_config(["--environment", "ABC=QED"])
+
+        (d / "mkosi.conf").write_text(
+            """
+            [Assert]
+            ImageId=abcde
+
+            [TriggerAssert]
+            Environment=ABC=QED
```
behrmann, line 1641: This test is a bit more interesting when adding a second `TriggertAssert` section, e.g.
```ini
[TriggerAssert]
Environment=DEF=QEE
```
since it exemplifies the behaviour when mixing both quite well.

tests/test_config.py
daandemeyer: I don't think checking whether config parsing works at all belongs in this test

tests/test_config.py
```diff
@@ -1578,3 +1578,73 @@ def test_subdir(tmp_path: Path) -> None:
...
+            """
+        )
+
+        with pytest.raises(SystemExit):
+            parse_config()
+        with pytest.raises(SystemExit):
+            parse_config()
+        with pytest.raises(SystemExit):
+            parse_config([])
```
behrmann: Should this maybe be

```suggestion
        with pytest.raises(SystemExit):
            parse_config([])
        with pytest.raises(SystemExit):
            parse_config(["--image-id", "abcde"])
        with pytest.raises(SystemExit):
            parse_config(["--environment", "ABC=QED"])
```

tests/test_config.py
```diff
@@ -1578,3 +1578,73 @@ def test_subdir(tmp_path: Path) -> None:
...
+
+            [TriggerAssert]
+            Environment=DEF=QEE
+            """
+        )
+
+        with pytest.raises(SystemExit):
+            parse_config()
+
+        parse_config(["--image-id", "abcde", "--environment", "ABC=QED"])
```
behrmann, line 1651:
```suggestion
        parse_config(["--image-id", "abcde", "--environment", "ABC=QED"])
        parse_config(["--image-id", "abcde", "--environment", "DEF=QEE"])        
```

tests/test_config.py
behrmann: Ok, but let's add a comment above, e.g.
```
# Does not raise, i.e. parses successfully, but we don't care for the content
```
then it's also clear for the rest.

PR #3955, opened 2025-10-15
ubuntu: Switch to devel as the default release

daandemeyer:
Same reasoning as for the other distros, devel will always point to something recent, whereas hardcoding stable release names will get out of date very soon.


daandemeyer:
A bunch of CIs fixed, one new one broken because ubuntu's kernel images are identified as UKIs. That's something to fix in their stubble bootloader, so merging to get our CI greener again.

PR #3956, opened 2025-10-15
Don't fix up vmlinuz locations if not required

### LINKED ISSUE #3949: 5e38eec86 breaks multiple kernels

daandemeyer:
If we already have a kernel image for every directory in /usr/lib/modules then there's no need to do anything.

Fixes #3949

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1309,6 +1309,17 @@ def kernel_get_ver_from_modules(context: Context) -> Optional[str]:
 
 
 def fixup_vmlinuz_location(context: Context) -> None:
+    modulesd = Path("usr/lib/modules")
+
+    if not (context.root / modulesd).exists():
+        return
+
+    # Don't touch anything if all the modules directories contain a kernel image already.
+    if all(
+        (d / "vmlinuz").is_file() or (d / "vmlinux").is_file() for d in (context.root / modulesd).iterdir()
+    ):
+        return
+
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -1309,6 +1309,18 @@ def kernel_get_ver_from_modules(context: Context) -> Optional[str]:
...
+
+    if not (context.root / modulesd).exists():
+        return
+
+    # Don't touch anything if all the modules directories contain a kernel image already.
+    for d in (context.root / modulesd).iterdir():
+        if not (d / "vmlinuz").is_file() and not (d / "vmlinux").is_file():
+            break
+    else:
+        return
```
```suggestion
    if all(
        (d / "vmlinuz").is_file() or (d / "vmlinux").is_file() for d in (context.root / modulesd).iterdir()
    ):
        return
```

behrmann (approved):

PR #3957, opened 2025-10-15
Look for .sdmagic before we consider a PE binary a UKI/addon

daandemeyer:
In ubuntu devel they started using the stubble bootloader for their kernel images which includes a .linux and a .osrel section which means it identifies as a UKI so we skip it. Let's additionally look for .sdmagic as a workaround to fix this for now.

To implement this we can't make use of bootctl kernel-identify anymore so we reimplement the logic ourselves with pefile. While we're at it, we move KernelType to bootloader.py as it makes more sense to live there than in qemu.py.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -29,6 +29,7 @@ from typing import Any, Optional, Union, cast
 
 from mkosi.archive import can_extract_tar, extract_tar, make_cpio, make_tar
 from mkosi.bootloader import (
+    KernelType,
     efi_boot_binary,
     extract_pe_section,
     gen_kernel_images,
@@ -106,7 +107,6 @@ from mkosi.mounts import (
 from mkosi.pager import page
 from mkosi.partition import Partition, finalize_root, finalize_roothash
 from mkosi.qemu import (
-    KernelType,
     copy_ephemeral,
     finalize_credentials,
     finalize_kernel_command_line_extra,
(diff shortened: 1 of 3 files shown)


keszybz:
Hmm, but wouldn't a better solution be to make `bootctl kernel-identify` detect those stubbled images properly? Fixing it just here is a workaround that we'll need in other places too.

daandemeyer:
> Hmm, but wouldn't a better solution be to make `bootctl kernel-identify` detect those stubbled images properly? Fixing it just here is a workaround that we'll need in other places too.

I don't know what the proper way to identify those images is, for all intents and purposes they look exactly like a UKI. And since they use ukify it'd be on us to add extra hacks to make sure they can build images that do not identify as UKIs. We can't add any stubble specific code to systemd to detect them either, so I have no clue how to fix this in a non-hacky way.

behrmann:
Would it maybe make sense to add an `--assert-section` option to `bootctl kerne-identify`? Then we could call `bootctl kernel-identify --assert-section .sdmagic`

daandemeyer:
> Would it maybe make sense to add an `--assert-section` option to `bootctl kerne-identify`? Then we could call `bootctl kernel-identify --assert-section .sdmagic`

Looking for `.sdmagic` is the hack we don't want to do in the first place. There should be no requirement for UKIs to have a `sdmagic` section.

behrmann:
True, I was thinking if maybe stubble wanted to do add a different section, they could then reuse the tooling to do detection as `bootctl kernel-identify --assert-section .stubble` or some such.

behrmann (approved):
I think it makes sense to merge this as is t get Ubuntu CI working again in a reasonable manner, but we do need to clarify the spec to answer the question: What is a UKI?

PR #3963, opened 2025-10-16
Make sure apt sources are installed when BaseTrees= is in the mix

daandemeyer:
Let's make sure to install apt sources every time we install the apt package, even if BaseTrees= or volatile packages or such are in the mix.

To make this work we unfortunately have to reintroduce install_packages().

While we're at it, make sure the installed apt sources don't use the configured mirror or snapshot. We only install a default set of sources and for anything custom users should install their own sources.

### DIFF
--- a/mkosi/distribution/kali.py
+++ b/mkosi/distribution/kali.py
@@ -24,11 +24,15 @@ class Installer(debian.Installer, distribution=Distribution.kali):
         return Distribution.kali
 
     @classmethod
-    def repositories(cls, context: Context, local: bool = True) -> Iterable[AptRepository]:
-        if context.config.snapshot:
+    def repositories(cls, context: Context, for_image: bool = False) -> Iterable[AptRepository]:
+        mirror = None if for_image else context.config.mirror
+        if not mirror:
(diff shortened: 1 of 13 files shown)


mkosi/distribution/kali.py
```diff
@@ -24,11 +24,15 @@ def default_tools_tree_distribution(cls) -> Distribution:
         return Distribution.kali
 
     @classmethod
-    def repositories(cls, context: Context, local: bool = True) -> Iterable[AptRepository]:
-        if context.config.snapshot:
+    def repositories(cls, context: Context, for_image: bool = False) -> Iterable[AptRepository]:
+        mirror = None if for_image else context.config.mirror
+        if not mirror:
+            mirror = "http://http.kali.org/kali"
```
behrmann, line 30: That should be mentioned in the man page in the table for default mirrors (it should have been there before, but I just noticed).

daandemeyer: Huh? I can see it right there in the table already

behrmann: I guess I was blind yesterday.

behrmann (approved):

PR #3965, opened 2025-10-17
Simplify implementation of distribution installers

daandemeyer:
- Use __init_subclass__() to automatically have distribution installers register themselves with the DistributionInstaller class on import
- Remove duplicated methods from Distribution that only forward to the corresponding DistributionInstaller and instead have callers access the Installer class directly. Make it a property to make this less verbose

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -31,7 +31,7 @@ from contextlib import AbstractContextManager
 from pathlib import Path
 from typing import Any, Callable, ClassVar, Generic, Optional, Protocol, TypeVar, Union, cast
 
-from mkosi.distributions import Distribution, detect_distribution
+from mkosi.distribution import Distribution, detect_distribution
 from mkosi.log import ARG_DEBUG, ARG_DEBUG_SANDBOX, ARG_DEBUG_SHELL, complete_step, die
 from mkosi.pager import page
 from mkosi.run import SandboxProtocol, find_binary, nosandbox, run, sandbox_cmd, workdir
--- a/mkosi/distribution/__init__.py
+++ b/mkosi/distribution/__init__.py
@@ -71,7 +71,7 @@ class Distribution(StrEnum):
 
     @property
     def installer(self) -> type["DistributionInstaller"]:
-        importlib.import_module(f"mkosi.distributions.{self.name}")
+        importlib.import_module(f"mkosi.distribution.{self.name}")
         return DistributionInstaller.registry[self]
 
 
--- a/mkosi/installer/pacman.py
+++ b/mkosi/installer/pacman.py
@@ -9,7 +9,7 @@ from pathlib import Path
 
 from mkosi.config import Config
 from mkosi.context import Context
-from mkosi.distributions import detect_distribution
+from mkosi.distribution import detect_distribution
 from mkosi.installer import PackageManager
 from mkosi.log import complete_step
 from mkosi.run import CompletedProcess, run, workdir
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -215,7 +215,7 @@ jobs:
           EOF
 
           # TODO: Use $SYSTEMD_REPART_OVERRIDE_FSTYPE_ROOT once we drop support for Ubuntu Noble.
-          sed -i 's/return "btrfs"/return "ext4"/' mkosi/distributions/*.py
+          sed -i 's/return "btrfs"/return "ext4"/' mkosi/distribution/*.py
 
           # fail if the script already exists, to avoid hard to debug CI errors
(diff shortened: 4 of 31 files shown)


mkosi/installer/pacman.py
```diff
@@ -164,7 +165,7 @@ def cmd(cls, context: Context) -> list[PathString]:
             "--cachedir=/var/cache/pacman/mkosi",
             "--cachedir=/var/cache/pacman/pkg",
             "--hookdir=/buildroot/etc/pacman.d/hooks",
-            "--arch", context.config.distribution.architecture(context.config.architecture),
+            "--arch", context.config.distribution.installer.architecture(context.config.architecture),
```
behrmann: Let's be consistent here and move this up like in the hunk above.

mkosi/installer/dnf.py
```diff
@@ -188,7 +188,7 @@ def cmd(
 
         if not context.config.architecture.is_native():
             cmdline += [
-                f"--forcearch={context.config.distribution.architecture(context.config.architecture)}"
+                f"--forcearch={context.config.distribution.installer.architecture(context.config.architecture)}"
             ]
```
behrmann: This would also be shorter by moving this out
```python
arch = context.config.distribution.installer.architecture(context.config.architecture)
cmdline += [f"--forcearch={arch}]

File: mkosi/installer/apt.py [hunk 3]
```diff
@@ -165,7 +165,7 @@ def finalize_environment(cls, context: Context) -> dict[str, str]:
 
     @classmethod
     def cmd(cls, context: Context, command: str = "apt-get") -> list[PathString]:
-        debarch = context.config.distribution.architecture(context.config.architecture)
+        debarch = context.config.distribution.installer.architecture(context.config.architecture)
```
behrmann: We could be more consistent here by renaming this to `arch`

mkosi/installer/apk.py
```diff
@@ -72,7 +72,7 @@ def cmd(cls, context: Context) -> list[PathString]:
             "--cache-packages",
             "--cache-dir", "/var/cache/apk",
             "--cache-predownload",
-            "--arch", context.config.distribution.architecture(context.config.architecture),
+            "--arch", context.config.distribution.installer.architecture(context.config.architecture),
```
behrmann: Looking at this pattern: How about we add `arch` as a classmethod on `PackageManager`?

mkosi/config.py
```diff
@@ -3626,7 +3629,9 @@ def parse_kernel_module_filter_regexp(p: str) -> str:
         parse=config_parse_string,
         match=config_make_string_matcher(),
         default_factory_depends=("tools_tree_distribution",),
-        default_factory=lambda ns: d.default_release() if (d := ns["tools_tree_distribution"]) else None,
+        default_factory=lambda ns: d.installer.default_release()
+        if (d := ns["tools_tree_distribution"])
+        else None,
```
behrmann: Let's parenthesise this, then after formatting it becomes the eminently more readable
```python
default_factory=(
    lambda ns: d.installer.default_release() if (d := ns["tools_tree_distribution"]) else None
)
```

mkosi/distribution/__init__.py
```diff
@@ -123,54 +69,69 @@ def is_rpm_distribution(self) -> bool:
...
 
-    def keyring(self, context: "Context") -> None:
-        return self.installer().keyring(context)
+class DistributionInstaller:
+    registry: dict[Distribution, "type[DistributionInstaller]"] = {}
 
-    def setup(self, context: "Context") -> None:
-        return self.installer().setup(context)
+    def __init_subclass__(cls, distribution: Distribution):
+        cls.registry[distribution] = cls
```
behrmann, line 82: Since this is more of an implementation detail and not part of the interface, I think it would be nicer to make this attribute `_registry` instead of `registry`

daandemeyer: If we did this anywhere else I'd agree but we don't so I don't wanna start doing it now.

behrmann: I think it would be nice to do it here, since it makes it easier to read, e.g. when adding a new distribution it makes it clear that one doesn't need to think about this at all after setting `distribution=` on the subclass, since it's fully localised to this class and one doesn't need to check whether it's used anywhere else. 

Just because we've not done this elsewhere yet is not a reason to start with it somewhere.

daandemeyer: > Just because we've not done this elsewhere yet is not a reason to start with it somewhere.

If we decide to do this I want us to start doing it consistently, but I'm not convinced it's worth to do this in the first place. So I'd prefer to leave that discussion for another time/pr and not have it block merging this one.

PR #3967, opened 2025-10-17
Allow users to disable custom console background color

qdeslandes:
When mkosi creates a new console with systemd-pty-forward, it sets a custom (dark) color, which might not play nice depending on your terminal theme.

Introduce the WithConsoleBackground= [Runtime] option to disable with behaviour. Custom console background is enabled by default, so the current behaviour is not modified.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -6036,3 +6036,24 @@ def systemd_tool_version(*tool: PathString, sandbox: SandboxProtocol = nosandbox
     logging.debug(f"Version reported by {tool[-1]} is {version}")
 
     return version
+
+
+def systemd_pty_forward(config: Config, background: str, title: str) -> list[str]:
+    tint_bg = parse_boolean(config.environment.get("SYSTEMD_TINT_BACKGROUND", "1")) and parse_boolean(
+        os.environ.get("SYSTEMD_TINT_BACKGROUND", "1")
+    )
+    adjust_title = parse_boolean(
+        config.environment.get("SYSTEMD_ADJUST_TERMINAL_TITLE", "1")
+    ) and parse_boolean(os.environ.get("SYSTEMD_ADJUST_TERMINAL_TITLE", "1"))
+
+    if not tint_bg and not adjust_title:
+        return []
+    if not config.find_binary("systemd-pty-forward"):
+        return []
+
+    cmd = ["systemd-pty-forward"]
+    if tint_bg:
+        cmd += [f"--background={background}"]
+    if adjust_title:
+        cmd += ["--title=", title]
+    return cmd
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -80,6 +80,7 @@ from mkosi.config import (
     parse_config,
     resolve_deps,
     summary,
+    systemd_pty_forward,
     systemd_tool_version,
     want_kernel,
     want_selinux_relabel,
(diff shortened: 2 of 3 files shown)


behrmann:
I agree this would be useful, but mkosi is not the right place to fix this, but this should preferably go into `systemd-pty-forward`. `systemd-run` and friends already take an environment variable `SYSTEMD_RUN_SHELL_PROMPT_PREFIX`, so something like `SYSTEMD_PTY_FORWARD_BGCOLOR` would make sense, which we could then check for and not set our own colours.

daandemeyer:
https://github.com/systemd/systemd/pull/39382 will do the trick by allowing you to set the `SYSTEMD_TINT_BACKGROUND=0` environment variable globally to disable color tinting.

daandemeyer:
We can just check the same environment variables in mkosi, so check for $SYSTEMD_TINT_BACKGROUND and $SYSTEMD_ADJUST_TERMINAL_TITLE in `config.environment` and `os.environ`. If both are not set we can skip using `systemd-pty-forward` completely.

qdeslandes:
I've modified the logic to check apply `--background` if `SYSTEMD_TINT_BACKGROUND` is not 0 in mkosi's environment variable or the global environment variables. Same for `--title`. 

`systemd-pty-forward` is still used if both the custom title and custom background are disabled as the `--read-only` flag is passed in some cases, it makes the logic simpler to use it by default.

mkosi/config.py
```diff
@@ -2343,6 +2332,20 @@ def build_subdir(self) -> Path:
...
+    def shall_customize_pty_title(self) -> bool:
+        return (
+            self.environment.get("SYSTEMD_ADJUST_TERMINAL_TITLE", "1") == "1"
+            and os.environ.get("SYSTEMD_ADJUST_TERMINAL_TITLE", "1") == "1"
+        )
+
+    @property
+    def shall_customize_pty_bg(self) -> bool:
+        return (
+            self.environment.get("SYSTEMD_TINT_BACKGROUND", "1") == "1"
```
behrmann: Can you make this `try_parse_boolean`?  systemd accepts more than `1` and `0` for these values, so we will have to do the same.

mkosi/config.py
```diff
@@ -2343,6 +2332,20 @@ def build_subdir(self) -> Path:
 
         return self.build_dir / subdir
 
+    @property
+    def shall_customize_pty_title(self) -> bool:
+        return (
+            self.environment.get("SYSTEMD_ADJUST_TERMINAL_TITLE", "1") == "1"
+            and os.environ.get("SYSTEMD_ADJUST_TERMINAL_TITLE", "1") == "1"
```
behrmann: I'm not quite convinced we need to check both. Checking just the OS environment seems fine to me, but let's wait what Daan thinks on this. 

What's more interesting, if neither the background should be tinted nor the title set, pty-forward can just be skipped.

qdeslandes: > I'm not quite convinced we need to check both. Checking just the OS environment seems fine to me, but let's wait what Daan thinks on this.

I've checked both according to Daan's original comment. If any of them is set to disable title/background, I disable it. Happy to modify if needed.

> What's more interesting, if neither the background should be tinted nor the title set, pty-forward can just be skipped.

I've updated the PR to implement this. I'm not sure if it's the right way because it makes the logic more difficult to follow for little benefit to the user. Using `systemd-pty-forward` by default with the proper options according to the env is simpler to understand than using `systemd-pty-forward` or not based on the env (especially as it's required for `Console.read_only`.

behrmann: I think an implementation to skip pty-forward could be to factor a function to set up systemd-pty-forward and unpack the result:
```python
def systemd_pty_forward(config: Config, background: str, title: str) -> list[str]:
    tint_bg = parse_boolean(
        self.environment.get("SYSTEMD_TINT_BACKGROUND", "1")
    ) and parse_boolean(os.environ.get("SYSTEMD_TINT_BACKGROUND", "1"))
    adjust_title = parse_boolean(
        self.environment.get("SYSTEMD_ADJUST_TERMINAL_TITLE", "1")
    ) and parse_boolean(os.environ.get("SYSTEMD_ADJUST_TERMINAL_TITLE", "1"))

    if not tint_bg and not adjust_title:
        return []
    if not config.find_binary("systemd-pty-forward"):
        return []

    cmd = ["systemd-pty-forward"]
    if ting_bg:
        cmd += [f"--background={background}""]
    if adjust_title:
        cmd += ["--title=", title]
    return cmd
```

daandemeyer: Yeah this looks good and allows reusing the logic between the two callsites

qdeslandes: Applied

PR #3969, opened 2025-10-20
arch: Switch to the newly established fastly mirror

christian-heusel:
Since we can expect quite good caching results for CI-type setups it
might make sense to use this new mirror as a default for mkosi.

---

cc @antiz96 @pitastrudl

### DIFF
--- a/mkosi/distribution/arch.py
+++ b/mkosi/distribution/arch.py
@@ -80,7 +80,7 @@ class Installer(DistributionInstaller, distribution=Distribution.arch):
                 elif context.config.snapshot:
                     mirror = "https://archive.archlinux.org"
                 else:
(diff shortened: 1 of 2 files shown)


daandemeyer:
Seems to work fine, can we assume the same longevity for this mirror as for the geo one? 

Also, can't the geo one use the fastly one behind the scenes?

christian-heusel:
> Seems to work fine, can we assume the same longevity for this mirror as for the geo one?

Yes it is planned to support this setup over a longer time, if it turns out to not work out for mkosi you can always switch back to the geomirrors 😊 If desired you can also keep the geomirrors as fallback to the fastly one in mkosi 😊

 > Also, can't the geo one use the fastly one behind the scenes?

Layering multiple geo-dns does not sound like a good idea, so we intend to keep these two separate.

PR #3971, opened 2025-10-21
Drop support for RuntimeScratch= and CDROM=

daandemeyer:
systemd-vmspawn solves the problem solved by RuntimeScratch= by resizing the entire image instead. Let's align with vmspawn and drop RuntimeScratch= and require users to use RuntimeSize= instead.


behrmann (approved):

mkosi/resources/man/mkosi.news.7.md
behrmann, line 1: The behaviour change of making unknown options warn instead of fail should also be mentioned here.

PR #3973, opened 2025-10-22
Fix and refactor console escape codes

behrmann:


mkosi/log.py
```diff
@@ -24,6 +24,39 @@ def die(message: str, *, hint: Optional[str] = None) -> NoReturn:
     sys.exit(1)
 
 
+class ConsoleCodes:
+    # \033\\ is one possible ECMA-48 string terminators, BEL would be valid for most terminal emulators as
+    # well, but kitty, annoyingly, actually rings the bell in that case. Other string terminators would be
```
jcgl17, line 32: Sample size of one, but I had no problems with kitty actually ringing the bell when terminating with BEL. (And fwiw, I'm a pretty heavy bell user.) In the issue I had opened for Kitty, Kovid even noted that BEL should work. So \*shrug\*

mkosi/log.py
```diff
@@ -24,6 +24,39 @@ def die(message: str, *, hint: Optional[str] = None) -> NoReturn:
...
+    @staticmethod
+    def set_window_title(title: str) -> str:
+        # The title is set twice, once for all terminal emulators that are not konsole and once for konsole
+        return f"\033]0;mkosi: {title}\033\\\033]30;mkosi: {title}\033\\"
+
+    @staticmethod
+    def push_window_title() -> str:
+        """Push the window title on the window title stack."""
+        # not supported by konsole
+        return "\033[22t\033\\"
```
jcgl17: From what I've read, the `ST` code (`\033\`) isn't needed by `CSI` sequences (`\033[`). Is this right?

mkosi/log.py
behrmann: I tested this again and using BEL gives an audible bell for me and also adds the bell emoji to the title, so I think this comment is correct.

mkosi/log.py
behrmann: I thought I had seen a terminal emulator where it didn't, but when I went through the list again, this worked everywhere, except konsole, where it doesn't anyway. Maybe I got that mixed up. Fixed.

behrmann:
@jcgl17 I reworked this somewhat and included your work from #3974 and added you as a Co-Author. Can you check again that this works for you?

jcgl17:
Cool, thanks! This continues to work for me.

DaanDeMeyer (approved):

PR #3974, opened 2025-10-22
log: terminate OSC escape sequences with ST code

jcgl17:
Commit f26cb341 introduced some OSC escape sequences that resulted in all mkosi output being hidden when using the Kitty terminal: https://github.com/kovidgoyal/kitty/issues/9139

This commit rectifies the situation by correctly sending the `ST` code after `OSC` has been sent, thereby ending the OSC sequence.

image

Ref:
- https://en.wikipedia.org/wiki/ANSI_escape_code#Operating_System_Command_sequences
- https://sw.kovidgoyal.net/kitty/shell-integration/#notes-for-shell-developers
- https://ghostty.org/docs/vt/concepts/sequences#osc-sequences

### DIFF
--- a/mkosi/log.py
+++ b/mkosi/log.py
@@ -16,6 +16,11 @@ ARG_DEBUG_SHELL = contextvars.ContextVar("debug-shell", default=False)
 ARG_DEBUG_SANDBOX = contextvars.ContextVar("debug-sandbox", default=False)
 LEVEL = 0
 
+# Terminal control characters
+ESCAPE_CODE_OSC = "\033]"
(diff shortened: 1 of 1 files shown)


behrmann:
Thanks for the contribution. @DaanDeMeyer already saw your issue against kitty and I put up #3973 earlier to fix this. It also fixes a bit more, since during testing I noticed that konsole behaves a bit weirdly there. Could you see whether #3973 fixes this for you?

jcgl17:
I totally missed your fix! Looks like I should've checked before opening this PR. Yeah, your #3973 fixes the problem for me too. Thanks much.

PR #3983, opened 2025-10-25
sandbox: Drop all capabilities that don't make sense in userns

daandemeyer:
When unsharing a user namespace, we get a full set of capabilities, of which a ton don't make sense to keep. Why drop them? Because it's possible that other tools check if they have the required capabilities to run, like systing now checking if it is invoked with CAP_BPF. If we don't drop CAP_BPF, systing will think it's able to attach BPF programs even though in reality it can't as CAP_BPF in a user namespace doesn't actually allow you to attach BPF programs.

While we're at it, let's be a bit more thorough with the capability logic and make sure we modify all capability sets to only contain the capabilities we want to keep.


behrmann (approved):
CI failure looks unrelated, the TPM problems we see regularly.

PR #3987, opened 2025-11-03
installer: Only mount configured state subdirs into sandbox

### LINKED ISSUE #3985: Arch: mkosi kernel-install plugin breaks during pacman updates

main

### Used host distribution

Arch

### Used target distribution

_No response_

### Linux kernel version used

_No response_

None

```
(7/9) Installing kernel and initrd using kernel-install...
‣ Building main image
‣  Copying in sandbox trees…
‣  Installing Arch Linux
error: failed to init transaction (unable to lock database)
error: could not lock database: File exists
  if you're sure a package manager is not already
  running, you can remove /var/lib/pacman/db.lck
‣ "pacman --root=/buildroot --logfile=/dev/null --dbpath=/var/lib/pacman --cachedir=/var/cache/pacman/mkosi --cachedir=/var/cache/pacman/pkg --hookdir=/buildroot/etc/pacman.d/hooks --arch x86_64 --color auto --noconfirm --sync --needed --assume-installed initramfs filesystem" returned non-zero exit code 1.
‣ "mkosi --force --directory= --format=cpio --output=initrd --output-directory=/tmp/tmp3sbykupt --extra-tree=/usr/lib/modules/6.17.6-arch1-1:/usr/lib/modules/6.17.6-arch1-1 --extra-tree=/usr/lib/firmware:/usr/lib/firmware '--remove-files=/usr/lib/firmware/*-ucode' --build-sources= --include=mkosi-initrd --kernel-modules=host --extra-tree=/usr/lib/modules/6.17.6-arch1-1/vmlinuz:/usr/lib/modules/6.17.6-arch1-1/vmlinuz --package-cache-dir=/var --cache-only=metadata --output-mode=600 --include /etc/mkosi-initrd --sandbox-tree=/tmp/tmp16unkl4i --kernel-command-line 'rootflags=compress-force=zstd zswap.enabled=0 i915.force_probe=!9a49 xe.force_probe=9a49
' --extra-tree=/etc/vconsole.conf:/etc/vconsole.conf" returned non-zero exit code 1.
‣ "mkosi-initrd --kernel-version 6.17.6-arch1-1 --format cpio --output initrd --output-dir /tmp/kernel-install.staging.dqxuR5 --kernel-image /usr/lib/modules/6.17.6-arch1-1/vmlinuz" returned non-zero exit code 1.
/usr/lib/kernel/install.d/50-mkosi.install failed with exit status 1.
```

When thereafter running `kernel-install add-all`, things work properly. So it only seems to break when triggered by pacman.

### Used mkosi config

```ini

```

### mkosi output

```sh

```

daandemeyer:
This got lost somewhere with the countless refactorings. Let's not mount a state directory to /var/lib unless one is configured. Most package managers don't actually store anything there that we care about and if we use PackageCacheDirectory=/var, we might end up mounting too much state there, such as the pacman host db lock file.

Fixes #3985

### DIFF
--- a/mkosi/installer/apt.py
+++ b/mkosi/installer/apt.py
@@ -64,8 +64,8 @@ class Apt(PackageManager):
         return ["*.deb", "*.ddeb"]
 
     @classmethod
-    def state_subdirs(cls, state: Path) -> list[Path]:
-        return [state / "lists"]
+    def state_subdirs(cls) -> list[Path]:
+        return [Path("lists")]
 
     @classmethod
     def dpkg_cmd(cls, command: str) -> list[PathString]:
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -4726,6 +4726,10 @@ def sync_repository_metadata(
     for d in ("cache", "lib"):
         (metadata_dir / d / subdir).mkdir(parents=True, exist_ok=True)
 
+    src = metadata_dir / "lib" / subdir
+    for p in last.distribution.installer.package_manager(last).state_subdirs():
+        (src / p).mkdir(parents=True, exist_ok=True)
+
     (last.package_cache_dir_or_default() / "cache" / subdir).mkdir(parents=True, exist_ok=True)
 
     # Sync repository metadata unless explicitly disabled.
@@ -4753,7 +4757,6 @@ def sync_repository_metadata(
                     context,
                     force=context.args.force > 1 or context.config.cacheonly == Cacheonly.never,
                 )
-
         src = metadata_dir / "cache" / subdir
         dst = last.package_cache_dir_or_default() / "cache" / subdir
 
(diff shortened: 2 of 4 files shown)


mkosi/installer/apt.py
```diff
@@ -65,7 +65,7 @@ def package_globs(cls) -> list[str]:
 
     @classmethod
     def state_subdirs(cls, state: Path) -> list[Path]:
```
behrmann: Both users of this now don't use the `state` argument, so maybe this should be dropped.

daandemeyer: I thought about it but it mimicks `package_subdirs()` which also receives the cache directory even though it's only used in one scenario, so I'm inclined to keep it for consistency, but I don't really mind either.

behrmann: I do see the symmetry argument, but if it has one user, it's at least used, but if this one has none, I would be slightly more inclined to not have an argument that's not used anywhere. A reason to keep it around would be if it might grow a user soon, but I'm also not sure about that. If that's the case it can be added back.

PR #3991, opened 2025-11-07
config.py: add config_default_proxy_exclude()

marc-hb:
Automagically defaulting --proxy-url to a proxy value found in the environment is nice. But it backfires seriously when not ignoring the no_proxy= value in the same environment, because it effectively blocks access to internal mirrors with timeouts and/or confusing error messages.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -1087,6 +1087,21 @@ def config_default_proxy_url(namespace: dict[str, Any]) -> Optional[str]:
     return None
 
 
+def config_default_proxy_exclude(namespace: dict[str, Any]) -> Optional[list[str]]:
+    names = ("no_proxy", "NO_PROXY")
+
+    for env in namespace["environment"]:
+        k, _, v = cast(str, env).partition("=")
+        if k in names:
+            return v.split(",")
+
+    for k, v in os.environ.items():
+        if k in names:
+            return v.split(",")
+
+    return None
+
+
 def config_default_proxy_peer_certificate(namespace: dict[str, Any]) -> Optional[Path]:
     for p in (Path("/etc/pki/tls/certs/ca-bundle.crt"), Path("/etc/ssl/certs/ca-certificates.crt")):
         if p.exists():
@@ -3916,6 +3931,8 @@ SETTINGS: list[ConfigSetting[Any]] = [
     ConfigSetting(
         dest="proxy_exclude",
         section="Build",
+        default_factory=config_default_proxy_exclude,
+        default_factory_depends=("environment",),
         metavar="HOST",
         parse=config_make_list_parser(delimiter=","),
         help="Don't use the configured proxy for the specified host(s)",


mkosi/config.py
```diff
@@ -1087,6 +1087,30 @@ def config_default_proxy_url(namespace: dict[str, Any]) -> Optional[str]:
...
+NO_PROXY_SEP = ","
+
+
+def config_default_proxy_exclude(namespace: dict[str, Any]) -> Optional[list[str]]:
+    names = ("no_proxy", "NO_PROXY")
+
+    for env in namespace["environment"]:
+        k, _, v = env.partition("=")
+        if k in names:
+            return cast(str, v).split(NO_PROXY_SEP)
```
behrmann: The confusion comes from further up, since what goes in is a `dict[str, Any]`. So let's put the cast further up, since that better explains the assumption and still makes mypy happy.
```suggestion
        k, _, v = cast(str, env).partition("=")
        if k in names:
            return v.split(NO_PROXY_SEP)
```

mkosi/config.py
```diff
@@ -1087,6 +1087,30 @@ def config_default_proxy_url(namespace: dict[str, Any]) -> Optional[str]:
...
+
+    for k, v in os.environ.items():
+        if k in names:
+            return v.split(NO_PROXY_SEP)
```
behrmann: `os.environ` is a mapping, so no need to iterate over the whole environment, just grab the ones you're interested in.

mkosi/config.py
```diff
@@ -1087,6 +1087,30 @@ def config_default_proxy_url(namespace: dict[str, Any]) -> Optional[str]:
     return None
 
 
+# There are rumours about some implementations using a space separator
+# but the comma seems the most popular.
+#      https://about.gitlab.com/blog/we-need-to-talk-no-proxy/
+# As of versions 8.11.1 and 2.2.0, curl and wget do not support spaces,
+# only commas. The lack of space is certainly less likely to trigger
+# quoting issues.
+NO_PROXY_SEP = ","
```
behrmann: I don't think this is necessary, if it's only rumours. We never supported anything else and it's not settable for users.

marc-hb: You mean the comment is not necessary? Or should I drop the the NO_PROXY_SEP constant too?

EDIT: I agree the lack of standardization described in https://about.gitlab.com/blog/we-need-to-talk-no-proxy/ is not about the separator.

mkosi/config.py
marc-hb: I can but that would be inconsistent with the `config_default_proxy_url()` logic above that I preserved. That could lead to "fun" debugging sessions (a bit like the ones in ... https://about.gitlab.com/blog/we-need-to-talk-no-proxy/?)

`os.environ` preserves the order from the operating system (at least on Linux it does). The current logic gives precedence to the first match in the operating system list. Your suggestion gives precedence to the first match in the `names` list.

Considering these variables are always exported, it's not extraordinary for `no_proxy` and `NO_PROXY` to be different: they could simply come from different scripts. People hate proxies so I've seen very many people more or less randomly copy/paste configuration from various places into various other places until it works.

I could "optimize" `config_default_proxy_url()` too to keep them consistent with each other but that would then be "backwards-incompatible". Changing the precedence in `config_default_proxy_url()` would be much more likely to affect users not just because that function already exists (obviously) but also because having different URLs for `http` versus `https` is very common, much more common than having different `no_proxy` and `NO_PROXY` values.

`os.environ` is never huge so this optimization is IMHO not worth the disruption.

mkosi/config.py
behrmann: I think you can drop `NO_PROXY_SEP`.

mkosi/config.py
behrmann: What a mess. 

Having read the post again, it seems that `no_proxy` is preferred everywhere… except Go, because of course.

> Considering these variables are always exported, it's not extraordinary for no_proxy and NO_PROXY to be different: they could simply come from different scripts. People hate proxies so I've seen very many people more or less randomly copy/paste configuration from various places into various other places until it works.

Having these variables set to different values, sounds quite broken, but I'm not quite sure how this applies to us, since mkosi will be called in some environment defined environment, e.g. a users, and probably not by random scripts? And we always set both variables to the same value. Maybe I'm misunderstanding your point.

But, in my opinion, having both set to different values sounds pretty broken in the first place and then having the preference be defined by ordering in the environment sounds less predictable than choosing one precedence (probably lowercase) and sticking with that, although warning if both are defined would be nice.

That all being said. You're right, it's already handled the same way in `config_default_proxy_url` and consistency trumps this, so keep it as is.

marc-hb: > Having these variables set to different values, sounds quite broken, but I'm not quite sure how this applies to us, since mkosi will be called in some environment defined environment, e.g. a users, and probably not by random scripts? And we always set both variables to the same value. Maybe I'm misunderstanding your point.

Yes, having different, non-empty values in `no_proxy` and `NO_PROXY` is definitely broken. But in practice, proxy configuration is broken much more often than not. One of the likely reasons is that build and test systems stack layer upon layer upon layer and each layer exports its own proxy configuration. An image build tool like mkosi is not necessarily at the top, far from it.

This is specific to neither interactive nor automated use. In fact, these two types of environment should differ as little as possible from each other to avoid "CI secrets" and unreproducible failures - but I digress.

> That all being said. You're right, it's already handled the same way in config_default_proxy_url and consistency trumps this, so keep it as is.

Thank you!

marc-hb:
Just for "fun", this is a real-world error message that you can get if you try to use internal mirrors and build Arch without this fix:

```
‣ Syncing package manager metadata
‣ + pacman --root=/buildroot --logfile=/dev/null --dbpath=/var/lib/pacman --cachedir=/var/cache/pacman/mkosi --cachedir=/var/cache/pacman/pkg --hookdir=/buildroot/etc/pacman.d/hooks --arch x86_64 --color auto --noconfirm --sync --refresh
:: Synchronizing package databases...
 core.db failed to download
 extra.db failed to download
error: failed retrieving file 'core.db' from linux-mirror.intel.com : The requested URL returned error: 403
error: failed retrieving file 'extra.db' from linux-mirror.intel.com : The requested URL returned error: 403
error: failed to synchronize all databases (failed to retrieve some files)
```

Good luck with that!

No `mkosi --debug*` option will tell you more.


PS: yes, proxies are evil. Tell me something I don't know :-)
PPS: this happens when `pacman` tries to reach linux-mirror.intel.com through the proxy. Don't ask.

Issue #3948, opened 2025-10-10
Better systemd-repart logs on failure


marc-hb:
This is  an offshoot from:
- #3223 

While #3223 focuses on fixing the actual issue, I think it would be beneficial to provider better `--debug` logs for `systemd-repart` in general. But I'm not sure how. I'd be happy to submit a pull request if anyone has any suggestion.

We use `mkosi` as a part of something bigger, which means most users do not interact directly with `mkosi` and are not familiar with it. So when they are affected by the #3223 "disk full" failure, they are completely puzzled by the error below and stuck.

```
Formatting ESP/XBOOTLDR partitions
‣ + systemd-repart --empty=allow --size=auto --dry-run=no --json=pretty --no-pager --root=/buildroot --offline=yes --seed 24133db4-0b8b-4d5b-9fc7-67fe31430d30 /work/tmp/rq_mkosi_wspaces/mkosi-workspace-y3td3nlz/staging/root.img.raw --definitions /work/tmp/qbuild.tmp/mkosi.repart
Automatically determined minimal disk image size as 3.2G, current image size is 3.2G.
File '/work/tmp/rq_mkosi_wspaces/mkosi-workspace-y3td3nlz/staging/root.img.raw' already is of requested size or larger, not growing. (3.2G >= 3.2G)
Applying changes to /work/tmp/rq_mkosi_wspaces/mkosi-workspace-y3td3nlz/staging/root.img.raw.
Successfully wiped file system signatures from future partition 0.
Successfully discarded data from future partition 0.
Formatting future partition 0.
Preparing to populate vfat filesystem.
Ready to populate vfat filesystem.
mkfs.fat 4.2 (2021-01-31)
Disk full
(mcopy) failed with exit status 1.
[...]

PR #3992, opened 2025-11-10
run.py: always log the sandbox when a command fails

marc-hb:
The main focus of commit https://github.com/systemd/mkosi/commit/7e7a3c71696d66d1774abff3eff80d5032612c88 ("Don't log sandbox for every
command") was to reduce verbosity in the --debug case and it did just
that. To avoid losing the sandbox completely, that commit said "... but
still log the full sandbox if a command fails" and it did that too in
`__init__.py`; for both the --debug and non-debug cases.

However, what happened in run.py was a bit different: there, commit
https://github.com/systemd/mkosi/commit/7e7a3c71696d66d1774abff3eff80d5032612c88 showed the sandox on failure but only in the --debug
case!? Make run.py consistent with `__init__.py` and always show the
sandbox on failure in run.py, in _both_ --debug and non-debug case.

This helps with issue https://github.com/systemd/mkosi/issues/3948: hiding that bind mounts exist is especially
confusing on fatal "disk full" failures because that leaves only
misleading paths in the error message! On failure, showing such critical
sandbox information must not require --debug.

Not concealing the sandbox on failure is useful in any case, not
just https://github.com/systemd/mkosi/issues/3948

Signed-off-by: Marc Herbert <marc.herbert@intel.com>

PR #3993, opened 2025-11-11
make_image: log systemd-repart *.conf files at the --debug level

marc-hb:
As discussed in #3948, systemd-repart *.conf files have default values which is convenient until this fails with some dreaded "disk full" error \- then it becomes very mysterious. To considerably speed up the investigation about what exactly is full, show the configuration files in use when using --debug.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3390,6 +3390,15 @@ def make_image(
     ):
         env["SYSTEMD_REPART_MKFS_EXT4_OPTIONS"] = "-O ^orphan_file"
 
+    if ARG_DEBUG.get():
+        for dir in definitions:
+            for c in dir.glob("*.conf"):
+                # Do not spam the logs in case something goes wrong
+                logging.debug(f"# {c} (truncated to 100 lines)")
+                with open(c) as f:
+                    for line in itertools.islice(f, 100):
+                        logging.debug(line.strip())
+
     with complete_step(msg):
         output = json.loads(
             run_systemd_sign_tool(


marc-hb:
1 random QEMU timeout just like in unrelated #3992 .

It seems ToolsTreeDistribution is Fedora in most timeouts.

https://github.com/systemd/mkosi/actions/runs/19250690460/job/55034934798#step:11:5849

Now 2 timeouts after re-run:
https://github.com/systemd/mkosi/actions/runs/19252159958/job/55039247078?pr=3993
https://github.com/systemd/mkosi/actions/runs/19252159958/job/55039247127?pr=3993

Now just 1:
https://github.com/systemd/mkosi/actions/runs/19253565585/job/55043403913?pr=3993

Still not winning the lottery:
https://github.com/systemd/mkosi/actions/runs/19255156663/job/55048285518?pr=3993

1 failing again:
https://github.com/systemd/mkosi/actions/runs/19259212198/job/55060033981?pr=3993

Finally all green!
https://github.com/systemd/mkosi/actions/runs/19269008595

mkosi/__init__.py
```diff
@@ -3390,6 +3390,19 @@ def can_orphan_file(distribution: Optional[Distribution], release: Optional[str]
...
 
+    if ARG_DEBUG.get():
+        for dir in definitions:
+            for c in dir.glob("*.conf"):
+                logging.debug(f"\n === {c} ===")
+                with open(c) as f:
+                    lines = 0
+                    for line in f:
+                        logging.debug(line.strip())
+                        lines += 1
```
```suggestion
                    for lines, line in enumerate(f):
                        logging.debug(line.strip())
```
though `lines` is then no longer a good name.

mkosi/__init__.py
```diff
@@ -3390,6 +3390,19 @@ def can_orphan_file(distribution: Optional[Distribution], release: Optional[str]
...
+                        if lines > 100:
```
behrmann: That seems a bit arbitrary. Is the idea for it to be much longer than than all possible repart files? The number of directives for these files is about 40.

mkosi/__init__.py
behrmann: Also, you could write something like `truncated to X lines`  in the message above saying from which file this is and instead of enumerating lines do 
```python
for line in itertools.islice(f, 100):
```

mkosi/__init__.py
marc-hb: It is arbitrary. Yes, the idea is for this to be longer than all "regular" repart files, as a safety to protect the debug logs and keep them usable, triggered only when something went terribly wrong. For instance: a wrong .conf file landing there by accident. Or some other bug. I can summarize this as a one-line comment in the source.

> The number of directives for these files is about 40.

I don't get this sorry. Did you mean 40 lines? Is this an average?

behrmann: > I don't get this sorry. Did you mean 40 lines? Is this an average?

That's just the rough amount of different directives for repart files. Even if you put multiple partitions in a file, which I've not seen done, to be honest, some of these directives are exclusive or dependent on others, so in general these files will be below 10 lines long.

marc-hb:
Sample output:

```
=== /tmp/rq_mkosi_wspaces/mkosi-workspace-nkv888ul/repart-definitions/10-root.conf ===
‣ [Partition]
‣ Type=root
‣ Format=btrfs
‣ CopyFiles=/
‣ Minimize=guess
‣ 
 === /tmp/rq_mkosi_wspaces/mkosi-workspace-nkv888ul/repart-definitions/00-esp.conf ===
‣ [Partition]
‣ Type=esp
‣ Format=vfat
‣ CopyFiles=/boot:/
‣ CopyFiles=/efi:/
‣ === 00-esp.conf truncated to 5 lines ===  (after changing the constant in the script)
```

mkosi/__init__.py
marc-hb: Done.

-----

As usual, QEMU from the Fedora ToolsTreeDistribution is hanging in 2 places. 
https://github.com/systemd/mkosi/actions/runs/19315730637/job/55247012466?pr=3993

Last time I force-pushed and banged until they all passed but this is getting tedious :-(

behrmann: > Last time I force-pushed and banged until they all passed but this is getting tedious :-(

You really don't need to. Sometimes CI is flaky, e.g. due to distro-specific problems, and in a lot of cases it is clear that the CI failure is not related to the change in the PR.

mkosi/__init__.py
```diff
@@ -3390,6 +3390,20 @@ def can_orphan_file(distribution: Optional[Distribution], release: Optional[str]
...
+                logging.debug(f"\n === {c} ===")
+                with open(c) as f:
+                    # Do not spam the logs in case something goes wrong
+                    maxlines = 100
+                    for line in itertools.islice(f, maxlines):
+                        logging.debug(line.strip())
+
+                    for anything_left in f:
+                        logging.debug(f"=== {c.name} truncated to {maxlines} lines ===")
+                        break
```
behrmann: Let's shorten this somewhat, the truncation will never really be a thing
```suggestion
                logging.debug(f"# {c} (truncated to 100 lines)")
                with open(c) as f:
                    for line in itertools.islice(f, 100):
                        logging.debug(line.strip())
```
Using a `#` as the separator, makes these files still valid, when copied from screen.

marc-hb: Done.

PR #3994, opened 2025-11-11
opensuse: fix Leap 16.0 build

aafeijoo-suse:
Regarding the first commit, see https://en.opensuse.org/Package_repositories#Update

```
Update

Repository of official security and bugfix updates for OSS packages.

Version: Tumbleweed http://download.opensuse.org/update/tumbleweed/
Version: Leap 16.0 No dedicated update repository as we use repo-oss for updates as well.
Version: Leap 15.6 http://download.opensuse.org/update/leap/15.6/oss/
```

I confirmed the change of the 2nd commit internally, but I haven't found any announcement.

### DIFF
--- a/mkosi/distribution/opensuse.py
+++ b/mkosi/distribution/opensuse.py
@@ -56,7 +56,12 @@ class Installer(DistributionInstaller, distribution=Distribution.opensuse):
     def install(cls, context: Context) -> None:
         packages = ["filesystem"]
         if not any(p.endswith("-release") for p in context.config.packages):
-            packages += ["openSUSE-release"]
+            if context.config.release in ("current", "stable", "leap") or (
+                context.config.release != "tumbleweed" and GenericVersion(context.config.release) >= 16
+            ):
+                packages += ["Leap-release"]
+            else:
+                packages += ["openSUSE-release"]
 
         cls.install_packages(context, packages, apivfs=False)
 


mkosi/distribution/opensuse.py
```diff
@@ -213,30 +214,22 @@ def repositories(cls, context: Context) -> Iterable[RpmRepository]:
...
-                yield RpmRepository(
-                    id="oss-update",
-                    url=f"baseurl={url}",
-                    gpgurls=fetch_gpgurls(context, url) if not zypper else (),
-                )
+            if (
+                context.config.release in ("current", "stable", "leap")
+                or GenericVersion(context.config.release) >= 16
+            ):
+                subdir += f"/distribution/{release}/repo"
```
daandemeyer: No need for the leading /, `join_mirror()` handles that

mkosi/distribution/opensuse.py
```diff
@@ -213,30 +214,22 @@ def repositories(cls, context: Context) -> Iterable[RpmRepository]:
...
+            else:
+                subdir += f"/update/{release}"
```
daandemeyer: Ditto

mkosi/distribution/opensuse.py
```diff
@@ -213,30 +214,22 @@ def repositories(cls, context: Context) -> Iterable[RpmRepository]:
...
-            if context.config.release in ("current", "stable", "leap"):
-                url = join_mirror(mirror, f"{subdir}/update/openSUSE-current")
-                yield RpmRepository(
-                    id="oss-update",
-                    url=f"baseurl={url}",
-                    gpgurls=fetch_gpgurls(context, url) if not zypper else (),
-                )
+            if (
+                context.config.release in ("current", "stable", "leap")
+                or GenericVersion(context.config.release) >= 16
```
daandemeyer, line 224: Relying on the fact that tumbleweed compares lower than 16 here seems very opaque. Please double check that the release is a digit before doing version comparisons on it.

aafeijoo-suse: Ok, I hadn't thought about handling a fixed Tumbleweed release, but you're right, better safe than sorry.

mkosi/distribution/opensuse.py
```diff
@@ -55,7 +56,13 @@ def setup(cls, context: Context) -> None:
     def install(cls, context: Context) -> None:
         packages = ["filesystem"]
         if not any(p.endswith("-release") for p in context.config.packages):
-            packages += ["openSUSE-release"]
+            if context.config.release in ("current", "stable", "leap") or (
+                GenericVersion(context.config.release) >= 16
+                and GenericVersion(context.config.release) < 20000000
```
behrmann: You can shorten this down
```suggestion
                16 <= GenericVersion(context.config.release) < 20000000
```
also the `20000000` comes a bit out of the blue if one doesn't know it's a date (I assume). A comment would help, or maybe
```suggestion
                16 <= GenericVersion(context.config.release) < 2000_00_00
```
though that also only makes the slightly more obvious it's a date.

aafeijoo-suse: Thanks Jörg, I wasn't familiar with this notation...

behrmann: You're welcome!

I still think a comment like `# tumbleweed releases are date-based` would be good/better, though.

aafeijoo-suse: done

mkosi/distribution/opensuse.py
```diff
@@ -55,7 +56,14 @@ def setup(cls, context: Context) -> None:
     def install(cls, context: Context) -> None:
         packages = ["filesystem"]
         if not any(p.endswith("-release") for p in context.config.packages):
-            packages += ["openSUSE-release"]
+            # Tumbleweed releases are date-based
+            if (
+                context.config.release in ("current", "stable", "leap")
+                or 16 <= GenericVersion(context.config.release) < 2000_00_00
```
daandemeyer: Tumbleweed snapshots are covered by the `Snapshot=` feature, they shouldn't be specified as releases. Can you drop that logic again?

aafeijoo-suse: Then I completely misunderstood your comment in https://github.com/systemd/mkosi/pull/3994#discussion_r2531812476

daandemeyer: If `release == "tumbleweed"`, then we should not pass it to `GenericVersion()` and do version comparisons with it

aafeijoo-suse: ok

mkosi/distribution/opensuse.py
```diff
@@ -55,7 +56,12 @@ def setup(cls, context: Context) -> None:
     def install(cls, context: Context) -> None:
         packages = ["filesystem"]
         if not any(p.endswith("-release") for p in context.config.packages):
-            packages += ["openSUSE-release"]
+            if context.config.release in ("current", "stable", "leap") or (
+                context.config.release != "tumbleweed" and GenericVersion(context.config.release) >= 16
```
daandemeyer, line 60: Frankly I would just use `context.config.release.isdigit() && GenericVersion(context.config.release) >= 16 `. That way you never compare anything that's not a digit against 16.

mkosi/distribution/opensuse.py
```diff
@@ -213,30 +219,22 @@ def repositories(cls, context: Context) -> Iterable[RpmRepository]:
...
-                yield RpmRepository(
-                    id="oss-update",
-                    url=f"baseurl={url}",
-                    gpgurls=fetch_gpgurls(context, url) if not zypper else (),
-                )
+            if (
+                context.config.release in ("current", "stable", "leap")
+                or GenericVersion(context.config.release) >= 16
+            ):
+                subdir += f"distribution/{release}/repo"
```
daandemeyer, line 226: This seems wrong, if release is a digit, we need to look inside of https://download.opensuse.org/distribution/leap/

mkosi/distribution/opensuse.py
```diff
@@ -213,30 +219,22 @@ def repositories(cls, context: Context) -> Iterable[RpmRepository]:
...
-            if context.config.release in ("current", "stable", "leap"):
-                url = join_mirror(mirror, f"{subdir}/update/openSUSE-current")
-                yield RpmRepository(
-                    id="oss-update",
-                    url=f"baseurl={url}",
-                    gpgurls=fetch_gpgurls(context, url) if not zypper else (),
-                )
+            if (
+                context.config.release in ("current", "stable", "leap")
+                or GenericVersion(context.config.release) >= 16
```
daandemeyer, line 224: Also needs an `isdigit()` check

mkosi/distribution/opensuse.py
aafeijoo-suse: Ah, but the problem is Leap is `15.6`, `16.0`..., and `isdigit` does not work for decimals. So, I could use `isdecimal`, hoping that releases always follow this.

mkosi/distribution/opensuse.py
aafeijoo-suse: I tested this before submit it. See:

https://github.com/systemd/mkosi/blob/c43123398cb8e0941c94fbee1f3dec8f6514e1bf/mkosi/distribution/opensuse.py#L178

mkosi/distribution/opensuse.py
aafeijoo-suse: nvm, `isdecimal` does not work either.

daandemeyer: OK let's keep it like this then

aafeijoo-suse: `context.config.release.replace(".", "").isnumeric()` works, but I was expecting a bit more from python...

npmccallum:
Could we get a release of mkosi with this? Thanks!

daandemeyer:
@npmccallum We just released mkosi v26

PR #3997, opened 2025-11-15
Bundle manifest

### LINKED ISSUE #3995: Optionally include manifest in built image
As discussed with @DaanDeMeyer on Matrix, it would helpful to have the manifest inside a built image.

My own use-case is diffing my current ParticleOS version against an updated version to see what packages have changed. Currently, I save manifests of all installed images somewhere on my build machine's filesystem, and then match against IMAGE_VERSION from /etc/os-release to get the current version's manifest. From there, I can diff against a newly-built image's manifest. It would be nicer to have something like, for example, {/etc,/usr/lib}/os-manifest.json.

More generally, I think it makes sense for OS images to be able to self-describe their contents. /etc/os-release already contains some high-level self-description, but more granular SBOM-y stuff feels desirable too.

As far as implementation goes, Daan mentioned on Matrix that it could make sense to emit the manifest earlier in the build process, and then optionally write it somewhere in /usr. It could be conditioned behind something like `CleanPackageMetadata=` so that there's no additional configuration added for this feature.

Not sure if it makes sense to compress the manifest; with my fully-featured ParticleOS desktop image, an uncompressed manifest is 302k. gzip brings it down to 25k. (Changelog goes from 3.5M to 538k with gzip.) *Edit:* compression would of course only matter for output formats that don't have transparent compression.

jcgl17:
My attempt to close #3995. Writes the json manifest to /usr/lib/os-manifest. Maybe it's too presumptuous to claim that path, but naming is hard and I had to pick something.

Since the manifest takes up only ~300k with >2000 packages, compressing the manifest didn't seem to make sense.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -90,7 +90,7 @@ from mkosi.config import (
 from mkosi.context import Context
 from mkosi.distribution import Distribution, detect_distribution
 from mkosi.documentation import show_docs
-from mkosi.installer import clean_package_manager_metadata
+from mkosi.installer import clean_package_manager_metadata, package_manager_metadata_to_clean
 from mkosi.kmod import (
     filter_devicetrees,
     is_valid_kdir,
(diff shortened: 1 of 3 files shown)


jcgl17:
It's probably better to name this file something like `mkosi-manifest` instead of `os-manifest`. A more reasonable claim.

behrmann:
> It's probably better to name this file something like mkosi-manifest instead of os-manifest. A more reasonable claim.

I'd prefer this to be mkosi-named, otherwise the name is too generic and similar to `os-release` and for something more general a UAPI spec would be more fitting that should then also take all the existing SBOM stuff into consideration, let's keep this localised to mkosi.

jcgl17:
Sounds right to me. Just pushed a change to change from `os-manifest` to `mkosi-manifest`.

daandemeyer (changes requested):
We should come up with a design that allows us to store manifests in sysexts and other images as well.

Please store the image manifest in /usr/lib/mkosi/<image>.manifest and store extension manifests in /usr/lib/mkosi/extension-manifest.d/<image>.manifest

jcgl17:
@daandemeyer, sorry for taking so long to come back around to this! I've pushed changes that I believe should implement the behavior you outlined, namely basing the name manifest path on the output format:

1. If an extension image (per `OutputFormat.is_extension_image()`), then save to `/usr/lib/mkosi/extension-manifest.d/.manifest`
2. Otherwise, save to `/usr/lib/mkosi/.manifest`

How does this look to you?

PR #3999, opened 2025-11-18
Configure default tools tree with ToolsTree=yes

daandemeyer:
ToolsTree=default has always been rather weird, let's use ToolsTree=yes instead (but keep compat ofc)


behrmann:
In principle I like this change, but the special value `default` has the upside, that `yes` somewhat implies `no` to work and as is this leads to the somewhat confusing error message
```
mkosi --tools-tree=no build
‣ no does not exist
```

PR #4000, opened 2025-11-27
Silence swtpm SHA1 signature deprecation warning

LaurenceKiln:
When launching `mkosi vm`, the following warning appears:
```
swtpm: Warning: Profile-enabled algorithms contain disabled 'RSA-1024-sign(SHA1, pkcs1-pss)'
swtpm: Warning: Setting OPENSSL_ENABLE_SHA1_SIGNATURES=1
```

This is due to the default profile of swtpm including support for SHA1 signatures, which are deprecated in OpenSSL.

According to swtpm(8), the swtpm options
```
--profile-name=custom  --profile-remove-disabled=check
```
will do the following:
```
- remove camellia, tdes, and rsaes (RSA encryption with PKCS#1 v1.5 padding)
- disable signature support (RSA and EC) over SHA1
- disable unpadded RSA encryption
- set the minimum size for RSA keys to 2048 bits
- set the minimum size for EC keys to 224 bits
```

These options are sufficient to silence the warning, and has the benefit/risk of automatically removing similarly deprecated algorithms in the future, as libtpms/swtpm evolves.

Since the mkosi invocation of `swtpm setup` restricts PCR banks to SHA256, mkosi is already opinionated about the TPM provided to the TPM. It seems reasonable to go ahead and just prune dead crypto algorithms.

If in the future users need more control over the swtpm configuration, the sensible solution is probably the addition of a TPMProfile= option, which provides a suitable swtpm profile in JSON format.

But for now, just make the warning go away.

You can look at the built-in swtpm profiles using:
```
$ swtpm_setup --tpm2 --print-profiles | jq
```

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -6040,6 +6040,21 @@ def want_selinux_relabel(
     return setfiles, policy, fc, binpolicy
 
 
+def swtpm_setup_version(sandbox: SandboxProtocol = nosandbox) -> GenericVersion:
+    version = GenericVersion(
+        run(
+            ["swtpm_setup", "--version"],
+            stdout=subprocess.PIPE,
+            sandbox=sandbox(),
+            success_exit_status=(0, 1),
+        ).stdout.split()[-1]
+    )
+
+    logging.debug(f"Version reported by swtpm_setup is {version}")
+
+    return version
+
+
 def systemd_tool_version(*tool: PathString, sandbox: SandboxProtocol = nosandbox) -> GenericVersion:
     version = GenericVersion(
         run(
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -43,6 +43,7 @@ from mkosi.config import (
     VsockCID,
     finalize_term,
     format_bytes,
+    swtpm_setup_version,
     systemd_pty_forward,
     systemd_tool_version,
     want_selinux_relabel,
@@ -271,6 +272,11 @@ def start_swtpm(config: Config) -> Iterator[Path]:
                 "--pcr-banks",
                 "sha256",
                 "--config", "/dev/null",
+                *(
+                    ["--profile-name=custom", "--profile-remove-disabled=check"]
+                    if swtpm_setup_version() >= "0.10.0" 
+                    else []
+                 ),
             ],
             sandbox=config.sandbox(options=["--bind", state, workdir(Path(state))]),
             stdout=None if ARG_DEBUG.get() else subprocess.DEVNULL,


daandemeyer:
@LaurenceKiln These options seem rather new, I don't think we can assume they're always available. Please add some version parsing of the swtpm version and conditionalize adding these options based on the reported version.

mkosi/qemu.py
```diff
@@ -271,6 +271,7 @@ def start_swtpm(config: Config) -> Iterator[Path]:
                 "--pcr-banks",
                 "sha256",
                 "--config", "/dev/null",
+                "--profile-name=custom", "--profile-remove-disabled=check",
```
behrmann:
```suggestion
                "--profile-name=custom", 
                "--profile-remove-disabled=check",
```

behrmann:
> These options seem rather new, I don't think we can assume they're always available.

For the record, they're only available starting with version 10, which can be found in Arch, Fedora and Suse's Leap 16 and Tumbleweed. No released Debian or Ubuntu comes even close.

LaurenceKiln:
Would you prefer the safer alternative, which is to pass in `env=dict(OPENSSL_ENABLE_SHA1_SIGNATURES="1")` ?

Note: the warning itself was only [added](https://github.com/stefanberger/swtpm/commit/8d31d88c) between 0.9.0 and 0.10.0

behrmann:
As @DaanDeMeyer already mentioned, you can guard the addition of these new options behind a version check. You can split the output `swtpm_setup --version`, where the version is the last element (splitting on whitespace). If it parses as a version and compares larger or equal to `0.10` you can add these options. 

Personally I don't think we should reenable legacy algorithms.

LaurenceKiln:
Parsing version strings is often brittle :
```
$ swtpm --version
TPM emulator version 0.10.1, Copyright (c) 2014-2022 IBM Corp. and others
```

But I can do that.

behrmann:
The one for `swtpm_setup` is a bit easier:
```
$ swtpm_setup --version
TPM emulator setup tool version 0.10.1
```

daandemeyer:
> Parsing version strings is often brittle :
> 
> ```
> $ swtpm --version
> TPM emulator version 0.10.1, Copyright (c) 2014-2022 IBM Corp. and others
> ```
> 
> But I can do that.

Yes it is, but we don't have anything better. If an environment variable can silence the warning as well then please do that so we don't need a version check.

mkosi/config.py
```diff
@@ -6040,6 +6040,24 @@ def want_selinux_relabel(
     return setfiles, policy, fc, binpolicy
 
 
+def swtpm_tool_version(*tool: PathString, sandbox: SandboxProtocol = nosandbox) -> GenericVersion:
```
daandemeyer: Please check the `swtpm_setup` version, name this function `swtpm_setup_version()` and drop the tool argument.

mkosi/qemu.py
```diff
@@ -261,6 +262,11 @@ def find_ovmf_firmware(config: Config, firmware: Firmware) -> Optional[OvmfConfi
 
 @contextlib.contextmanager
 def start_swtpm(config: Config) -> Iterator[Path]:
+    extra_args = []
+    if swtpm_tool_version("swtpm") >= "0.10.0":
+        extra_args = ["--profile-name=custom",
+                      "--profile-remove-disabled=check",
+                     ]
```
daandemeyer: Please inline this into `run()` like so

```
*(["--profile-name=custom", "--profile-remove-disabled=check"] if swtpm_setup_version() >= "0.10.0" else []),
```

behrmann: The line's a bit long then and would be broken normally, but isn't because of the `fmt: skip` so the option values align with their options. You can break it this way
```
                *(
                    ["--profile-name=custom", "--profile-remove-disabled=check"]
                    if swtpm_tool_version("swtpm") >= "0.10.0"
                    else []
                )
```
we do this in a few other places.

mkosi/config.py
LaurenceKiln: Right

mkosi/qemu.py
```diff
@@ -271,7 +277,7 @@ def start_swtpm(config: Config) -> Iterator[Path]:
                 "--pcr-banks",
                 "sha256",
                 "--config", "/dev/null",
-            ],
+            ] + extra_args,
```
behrmann: You need to inline this here before the list is closed.

LaurenceKiln: Got it. Thanks.

mkosi/config.py
```diff
@@ -6040,6 +6040,24 @@ def want_selinux_relabel(
...
+    version = GenericVersion(
+        run(
+            ["swtpm_setup", "--version"],
+            stdout=subprocess.PIPE,
+            sandbox=sandbox(),
+        )
+        .stdout.split()[-1]
+        .strip("()")
+        .removeprefix("v")
+        .removesuffix(",")
```
behrmann: These are unnecessary now.

mkosi/qemu.py
```diff
@@ -271,6 +272,7 @@ def start_swtpm(config: Config) -> Iterator[Path]:
                 "--pcr-banks",
                 "sha256",
                 "--config", "/dev/null",
+                *(["--profile-name=custom", "--profile-remove-disabled=check"] if swtpm_setup_version() >= "0.10.0" else []),
```
behrmann:
```suggestion
                *(
                    ["--profile-name=custom", "--profile-remove-disabled=check"] 
                    if swtpm_setup_version() >= "0.10.0" 
                    else []
                 ),
```

LaurenceKiln:
Placing `OPENSSL_ENABLE_SHA1_SIGNATURES=1` in the env will silence the warning, but also hides it from the user, and may require occasional maintenance in the future if other algos are disabled. Do you still prefer that option @DaanDeMeyer ?

mkosi/qemu.py
```diff
@@ -45,6 +45,7 @@
     format_bytes,
     systemd_pty_forward,
     systemd_tool_version,
+    swtpm_setup_version,
```
behrmann: This is what CI complains about, `swtpm_setup_version` needs to go before `systemd_pty_forward`.

daandemeyer:
> Placing `OPENSSL_ENABLE_SHA1_SIGNATURES=1` in the env will silence the warning, but also hides it from the user, and may require occasional maintenance in the future if other algos are disabled. Do you still prefer that option @DaanDeMeyer ?

Let's stick with what you have now

mkosi/qemu.py
LaurenceKiln: I haven't replicated your tooling locally yet. sorry.

mkosi/qemu.py
behrmann: CI still complains about this.

mkosi/config.py
LaurenceKiln: Deliberate. for future-proofing?

mkosi/qemu.py
LaurenceKiln: ruff check != ruff format. I'll remember.

LaurenceKiln:
I'm looking at the failures.

mkosi/config.py
behrmann: That's only guesswork on what it might be in the future. Since the operations don't commute it would have to change in that particular way, but even if it did, the first step takes the last element of the whitespace split, i.e. if it changed to be, e.g. like the version output for `swtpm` itself, you'd end up with `others` as version. 

Please drop the unnecessary stuff and let's fix this once it breaks and keep the code clean.

mkosi/config.py
```diff
@@ -6040,6 +6040,24 @@ def want_selinux_relabel(
     return setfiles, policy, fc, binpolicy
 
 
+def swtpm_setup_version(sandbox: SandboxProtocol = nosandbox) -> GenericVersion:
+    version = GenericVersion(
+        run(
```
behrmann, line 6045: you will need to set `success_exit_status=(0, 1)` here, since old `swtpm_setup` (?) returns 1 for `--version` No idea who thought this a good idea.

mkosi/config.py
LaurenceKiln: `removesuffix` and `removeprefix` do commute, and unless you think the future version string will change to `version v(0.12.0,)` I don't see your point.

If you fear they'll change the order of the tokens, than these added cleanups are still inoffensive, but not helpful either. 

So I don't see any downsides, but potential upside if these compensate for minor version format changes in the future.

but :shrug: .... I'll remove it.

mkosi/config.py
LaurenceKiln: thank you!, I've been scratching my head over this one.

mkosi/config.py
LaurenceKiln:
```python
if m:=re.search(r"(\d+\.\d+\.\d+})",s): 
    version = m.group(1)
```
would also work and might be better overall

mkosi/qemu.py
```diff
@@ -271,6 +272,8 @@ def start_swtpm(config: Config) -> Iterator[Path]:
                 "--pcr-banks",
                 "sha256",
                 "--config", "/dev/null",
+                *(["--profile-name=custom", "--profile-remove-disabled=check"]
+                   if swtpm_setup_version() >= "0.10.0" else []),
```
behrmann: Please follow this style
```suggestion
                *(
                    ["--profile-name=custom", "--profile-remove-disabled=check"]
                    if swtpm_setup_version() >= "0.10.0" 
                    else []
                 ),
```
It is what `ruff` would format it to if we didn't have `fmt: skip` below, which we have so options and their values are on the same line.

LaurenceKiln:
Thanks for the help.

PR #4002, opened 2025-11-27
Rename mkosi 'ssh' service to 'mkosi-sshd'

LaurenceKiln:
Since It's `sshd` not `ssh`. Also distinguish it from system `sshd`, and from systemd's `sshd-vsock`.


daandemeyer:
I don't wanna touch this too much actually, this can be considered deprecated, I'll remove it at some point when the systemd stuff is more widespread

PR #4005, opened 2025-11-28
Improve manpage output using pandoc lua filter

LaurenceKiln:
pandoc treats `foo` as a fixed-with font change when converting to troff/manpage output, which has no visual effect  when reading manpages in the console.

pandoc supports lua scripts for manipulating the AST during conversion. The added script converts `code` to **code** in the manpage output.

Before:
mkosi_man_before
After:
mkosi_man_after

### DIFF
--- a/tools/make-man-page.sh
+++ b/tools/make-man-page.sh
@@ -1,11 +1,14 @@
 #!/bin/sh
 # SPDX-License-Identifier: LGPL-2.1-or-later
-# shellcheck disable=SC2086 # Allow word splitting on bare vars
 set -ex
 
-PANDOC_ARGS="--lua-filter=mkosi/resources/pandoc/md2man.lua -t man -s"
-pandoc $PANDOC_ARGS -o mkosi/resources/man/mkosi.1 mkosi/resources/man/mkosi.1.md
-pandoc $PANDOC_ARGS -o mkosi/resources/man/mkosi-addon.1 mkosi/resources/man/mkosi-addon.1.md
-pandoc $PANDOC_ARGS -o mkosi/resources/man/mkosi-initrd.1 mkosi/resources/man/mkosi-initrd.1.md
-pandoc $PANDOC_ARGS -o mkosi/resources/man/mkosi-sandbox.1 mkosi/resources/man/mkosi-sandbox.1.md
-pandoc $PANDOC_ARGS -o mkosi/resources/man/mkosi.news.7 mkosi/resources/man/mkosi.news.7.md
+MD_DIR=mkosi/resources/man
+OUTPUT_DIR=mkosi/resources/man
+
+for mdfile in "$MD_DIR"/*.?.md; do
+    pandoc \
+    --lua-filter=mkosi/resources/pandoc/md2man.lua \
+    -s -t man \
+    -o  "${OUTPUT_DIR}/$(basename "${mdfile}" .md)" \
+    "${mdfile}"
+done


mkosi/resources/man/mkosi.1.md
```diff
@@ -201,6 +201,7 @@ The following command line verbs are known:
     `mkosi/resources` directory of the Python package, if it is missing, as well as to install it in the
     appropriate search path for man pages. The man page can be generated from the markdown file
     `mkosi/resources/man/mkosi.1.md` e.g via `pandoc -t man -s -o mkosi.1 mkosi.1.md`.
+    See also `tools/make-man-page.sh`.
```
behrmann: Please drop this, the tools in `tools/` are not shipped.

mkosi/documentation.py
```diff
@@ -32,11 +32,15 @@ def show_docs(
                     logging.warn("pandoc is not available")
                     continue
                 pandoc = run(
-                    ["pandoc", "-t", "man", "-s", resources / f"man/{manual}.{man_chapter}.md"],
+                    [
+                        "pandoc", "-t", "man",
```
behrmann:
```suggestion
                        "pandoc", 
                        "-t", "man",
```

mkosi/documentation.py
```diff
@@ -32,11 +32,15 @@ def show_docs(
...
+                        "-s", resources / f"man/{manual}.{man_chapter}.md",
+                        "--lua-filter", resources / "/pandoc/md2man.lua",
```
behrmann: Without this change this does not work
```suggestion
                        "--lua-filter", resources / "pandoc/md2man.lua",
```

tools/make-man-page.sh
```diff
@@ -2,8 +2,15 @@
...
-pandoc -t man -s -o mkosi/resources/man/mkosi-initrd.1 mkosi/resources/man/mkosi-initrd.1.md
-pandoc -t man -s -o mkosi/resources/man/mkosi-sandbox.1 mkosi/resources/man/mkosi-sandbox.1.md
-pandoc -t man -s -o mkosi/resources/man/mkosi.news.7 mkosi/resources/man/mkosi.news.7.md
+EXTRA_ARGS=
+if [ -f mkosi/resources/pandoc/md2man.lua ]; then
+    EXTRA_ARGS="--lua-filter=mkosi/resources/pandoc/md2man.lua"
+else
+    echo "Warning: 'md2man.lua' was not found. Output may be degraded."
+fi
+
```
behrmann: This doesn't need handling that this is not available, since the script is only used during development and packaging. You can roll all the options into one variable.
```suggestion
PANDOC_ARGS='--lua-filter=mkosi/resources/pandoc/md2man.lua -t man -s'
```

tools/make-man-page.sh
```diff
@@ -2,8 +2,15 @@
...
+pandoc ${EXTRA_ARGS} -t man -s -o mkosi/resources/man/mkosi.1 mkosi/resources/man/mkosi.1.md
```
behrmann: Style nit: No unnecessary curly braces
```suggestion
pandoc $EXTRA_ARGS -t man -s -o mkosi/resources/man/mkosi.1 mkosi/resources/man/mkosi.1.md
```

behrmann:
With the above changes this works
- straight from the checkout called via `python -m mkosi`
- straight from the checkout called via the shim
- installed into a virtualenv
- installed into a virtualenv in editable mode
- as zipapp

LaurenceKiln:
Thanks for the review @behrmann

tools/make-man-page.sh
```diff
@@ -1,9 +1,11 @@
...
 # SPDX-License-Identifier: LGPL-2.1-or-later
+# shellcheck disable=SC2086 # Allow word splitting on bare vars
 set -ex
 
-pandoc -t man -s -o mkosi/resources/man/mkosi.1 mkosi/resources/man/mkosi.1.md
-pandoc -t man -s -o mkosi/resources/man/mkosi-addon.1 mkosi/resources/man/mkosi-addon.1.md
-pandoc -t man -s -o mkosi/resources/man/mkosi-initrd.1 mkosi/resources/man/mkosi-initrd.1.md
-pandoc -t man -s -o mkosi/resources/man/mkosi-sandbox.1 mkosi/resources/man/mkosi-sandbox.1.md
-pandoc -t man -s -o mkosi/resources/man/mkosi.news.7 mkosi/resources/man/mkosi.news.7.md
+PANDOC_ARGS="--lua-filter=mkosi/resources/pandoc/md2man.lua -t man -s"
```
daandemeyer: Use an array instead and do `"${PANDOC_ARGS[@]}"` and then you don't need to disable the shellcheck

DaanDeMeyer (changes requested):

tools/make-man-page.sh
LaurenceKiln: So #!/bin/bash is ok?
Pushed alternative via loop as extra commit.

daandemeyer: bash is ok

PR #4008, opened 2025-11-29
verity: copy signing certificate to /usr/lib/verity.d/

bluca:
If a signing certificate for verity is specified copy it in the image to /usr/lib/verity.d/ so that it can be used for userspace verification when not using secure boot

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -511,6 +511,25 @@ def configure_autologin(context: Context) -> None:
         )
 
 
+def configure_verity_certificate(context: Context) -> None:
+    if not context.config.verity_certificate:
+        return
+
+    # TODO: support providers after https://github.com/systemd/systemd/pull/39962 is merged
+    if context.config.verity_certificate_source.type != CertificateSourceType.file:
+        return
+
+    veritydir = context.root / "usr/lib/verity.d"
+    with umask(~0o755):
+        veritydir.mkdir(parents=True, exist_ok=True)
+
+    # dissect wants .crt and will ignore anything else
+    dest = veritydir / context.config.verity_certificate.with_suffix(".crt").name
+
+    with umask(~0o644):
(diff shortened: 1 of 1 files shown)


bluca:
As requested by @poettering

mkosi/__init__.py
```diff
@@ -511,6 +511,30 @@ def configure_autologin(context: Context) -> None:
...
+        veritydir.mkdir(parents=True, exist_ok=True)
+
+    # dissect wants .crt and will ignore anything else
+    if context.config.verity_certificate.suffix == ".crt":
+        dest = veritydir / context.config.verity_certificate.name
+    else:
+        dest = (
+            veritydir
+            / f"{context.config.verity_certificate.name[: -len(context.config.verity_certificate.suffix)]}.crt"  # noqa: E501
+        )
```
hundeboll: `verity_certificate` is a `Path` object, right? Then
```py
# dissect wants .crt and will ignore anything else
dest = veritydir / context.config.verity_certificate.with_suffix(".crt").name
```

mkosi/__init__.py
```diff
@@ -511,6 +511,30 @@ def configure_autologin(context: Context) -> None:
         )
 
 
+def configure_verity_certs(context: Context) -> None:
```
daandemeyer: There's just one so let's call it `configure_verity_certificate()`

mkosi/__init__.py
```diff
@@ -511,6 +511,24 @@ def configure_autologin(context: Context) -> None:
...
+    if not context.config.verity_certificate:
+        return
+
+    if context.config.verity_certificate_source.type != CertificateSourceType.file:
+        return
```
daandemeyer, line 520: https://github.com/systemd/systemd/pull/39962 will make this work with providers as well. Can you add a TODO to integrate that when it is merged?

PR #4010, opened 2025-11-29
Eliminate `btrfs subvolume delete` errors when building as non-root

LaurenceKiln:
Eliminate these misleading errors when non-root runs `mkosi build --debug`:
```
ERROR: Could not destroy subvolume/snapshot: Operation not permitted
WARNING: deletion failed with EPERM, you don't have permissions or send
may be in progress or the subvolume is set as default
```

The errors stem from `btrfs subvolume delete` requiring either running as root, or the `user_subvol_rm_allowed` mount option.
  
Since kernel 4.18+ (2018), btrfs has allowed deletion of subvolumes via the rmdir(2) syscall. So in practice, we almost always can just use `rm -rf` to remove a subolume.

~~Version 1:  sniff `sysfs` for the btrfs `rmdir_subvol` feature and if it's present, we don't have to special case the subvolume delete and just use `rm`.~~

Version 2: check if we have root privileges or volume has magic mount option before running `btrfs subvolume delete` .

### DIFF
--- a/mkosi/tree.py
+++ b/mkosi/tree.py
@@ -14,7 +14,7 @@ from mkosi.config import ConfigFeature
 from mkosi.log import ARG_DEBUG, die
 from mkosi.run import SandboxProtocol, nosandbox, run, workdir
 from mkosi.sandbox import BTRFS_SUPER_MAGIC, FS_NOCOW_FL, OVERLAYFS_SUPER_MAGIC, chattr, lsattr, statfs
-from mkosi.util import PathString, flatten
+from mkosi.util import PathString, flatten, have_host_uid0
 from mkosi.versioncomp import GenericVersion
 
 
@@ -32,6 +32,24 @@ def cp_version(*, sandbox: SandboxProtocol = nosandbox) -> GenericVersion:
         .stdout.splitlines()[0]
         .split()[3]
     )
+
+
+def dir_mount_has_option(path: Path, option: str, sandbox: SandboxProtocol = nosandbox) -> bool:
+    cmdline = ["findmnt", "-rn", "-O", option, "-T", workdir(path, sandbox)]
+    completed_process = run(
+        cmdline,
+        check=True,
+        sandbox=sandbox(
+            options=flatten([("--bind", path.parent, workdir(path.parent, sandbox))]),
+        ),
+        success_exit_status=(0, 1),
+        stdout=subprocess.DEVNULL if not ARG_DEBUG.get() else None,
+        stderr=subprocess.DEVNULL if not ARG_DEBUG.get() else None,
+    )
+
+    # found : returncode=0 and stdout contains mount line
+    # not found : returncode=1 and stdout is empty
+    return completed_process.stdout is not None and option in completed_process.stdout
 
 
 def make_tree(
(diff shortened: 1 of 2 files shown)


LaurenceKiln:
There **is** a performance hit in `rm -rf` vs `subvolume delete` , but it's less than on a second with a 1GB image as long as the kernel  dentry cache is hot, which it would be when building an image.

daandemeyer:
I care a lot about that one second unfortunately.

I don't think we have to use btrfs-progs at all for the subvolume stuff actually, the ioctl's aren't that hard, so if we want to get rid of the error, I'd prefer we switch to using the ioctl()'s directly to create and remove subvolumes.

LaurenceKiln:
> I care a lot about that one second unfortunately.

Builds aren't fast enough here to make this seem consequential. But it turns out doing the correct thing here isn't hard to do.

> I don't think we have to use btrfs-progs at all for the subvolume stuff actually, the ioctl's aren't that hard, so if we want to get rid of the error, I'd prefer we switch to using the ioctl()'s directly to create and remove subvolumes.

Since `btrfs subvolume delete` itself uses `BTRFS_IOC_SNAP_DESTROY` to delete the subvolume, I don't see that calling the ioctl ourselves gets us past the privileges issue.

Pushed updated version, which explicitly checks if we're running with host root privs or the volume is mounted with the magic option.

Review note:
- At least on fedora, `findmnt` is already available via mkosi dependencies via: 
`systemd->(util-linux-core or util-linux) -> findmnt`

Tested with and without sudo, and checked the mount option detection inside and outside non-/root container.

daandemeyer:
> Since btrfs subvolume delete itself uses BTRFS_IOC_SNAP_DESTROY to delete the subvolume, I don't see that calling the ioctl ourselves gets us past the privileges issue.

No but it does allow us to check `errno` and just not log if we fail because of privileges

LaurenceKiln:
> No but it does allow us to check errno and just not log if we fail because of privileges

Right. Hopefully this latest version makes that unnecessary.

mkosi/util.py
```diff
@@ -252,3 +252,14 @@ def mandatory_variable(name: str) -> str:
         return os.environ[name]
     except KeyError:
         die(f"${name} must be set in the environment")
+
+
+def have_host_uid0() -> bool:
+    result = False
+    regex = rf"\s*{os.geteuid()}\s*{0}\s*(\d+)$"
```
behrmann, line 259: please use the proper regex quoting function and compile the regex

mkosi/util.py
```diff
@@ -252,3 +252,14 @@ def mandatory_variable(name: str) -> str:
...
+    with open("/proc/self/uid_map") as f:
+        for line in f.read().split("\n"):
+            if re.match(regex, line):
+                result = True
+
+    return result
```
behrmann, line 265:
```suggestion
        for line in f.read().split("\n"):
        if any(re.match(regex, line) for line in f.read().splitlines()):
            return True
            
        return False
```
or just read the whole string from the file and match on it instead of matching for every file, since it won't be gigabytes in size.

mkosi/tree.py
```diff
@@ -34,6 +34,24 @@ def cp_version(*, sandbox: SandboxProtocol = nosandbox) -> GenericVersion:
     )
 
 
+def dir_mount_has_option(path: Path, option: str, sandbox: SandboxProtocol = nosandbox) -> bool:
+    cmdline = ["findmnt", "-rn", "-O", option, "-T", workdir(path, sandbox)]
+    completed_process = run(
+        cmdline,
```
behrmann, line 40:
```suggestion
    completed_process = run(
        ["findmnt", "-rn", "-O", option, "-T", workdir(path, sandbox)],
```

LaurenceKiln:
> I find this very hacky and lots of added complexity just to get rid of some harmless debug logging.
> 
> I'd really prefer we switch to the ioctls to fix this rather than hacking around it like this

Inside the btrfs kernel code, exactly these checks are made in all likelihood. I don't see what's hacky about it. 

The reason to get rid of this "harmless" warnings, is that you don't know they're harmless until you spend time looking into them. Which I had to do because they seemed genuine. When users sit down to debug an issue with their build, they shouldn't have to wade through red herrings.

Can I add an "Please Ignore these if you're running as non-root" message? that's also a solution from my pov.

mkosi/util.py
LaurenceKiln: Not sure I follow. Do you mean I should `re.escape(os.geteuid())`? when would this break?

Since you suggested below it's better to search once through the whole file, I skipped the compilation.

daandemeyer:
I just implemented the ioctl stuff myself, let's continue in https://github.com/systemd/mkosi/pull/4018

PR #4016, opened 2025-12-02
Don't discard ordering of include/exclude entries in KernelModules=

LaurenceKiln:
Since eecf8b3b5 `KernelModules=` supports both inclusion and exclusion of kmods. When resolving kmod presets like "default" and "host" (and perhaps others in the future) it's important to retain the ordering of entries, so that the user can add a preset and then subtract only specific modules from the working set.

`filter_kernel_modules` relies on this ordering to do its job.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1514,7 +1514,8 @@ def finalize_kernel_modules_include(context: Context, *, include: Sequence[str],
         else:
             final.append(p)
 
-    return final
+    # deduplicate while maintaining ordering
+    return list({k: None for k in final})
 
 
 def build_kernel_modules_initrd(context: Context, kver: str) -> Path:


mkosi/__init__.py
```diff
@@ -1491,19 +1491,21 @@ def build_microcode_initrd(context: Context) -> list[Path]:
...
+        if "default" in patt:
+            with chdir(context.resources / "mkosi-initrd"):
+                # TODO: figure out a way to propagate all relevant settings, not just arch
+                _, _, [initrd] = parse_config(
+                    ["--architecture", str(context.config.architecture)],
+                    resources=context.resources,
+                )
+            final.extend(initrd.kernel_modules_include)
+        elif host or "host" in patt:
+            final.extend(loaded_modules())
```
behrmann: This adds `loaded_modules()` on every iteration if `host` is `True`.

mkosi/__init__.py
```diff
@@ -1491,19 +1491,21 @@ def build_microcode_initrd(context: Context) -> list[Path]:
...
-                ["--architecture", str(context.config.architecture)],
-                resources=context.resources,
-            )
-        final.update(initrd.kernel_modules_include)
-    if host or "host" in include:
-        final.update(loaded_modules())
-
+def finalize_kernel_modules_include(context: Context, *, include: Sequence[str], host: bool) -> list[str]:
+    final = []
+    for patt in include:
```
daandemeyer: Please use `p`

mkosi/__init__.py
LaurenceKiln: Well spotted.

LaurenceKiln:
@DaanDeMeyer , I noticed that when you merge PRs the original commit message is lost and there's also no reference to the Issue/PR. Is this deliberate? Should I just skip the commit messages?

daandemeyer:
> @DaanDeMeyer , I noticed that when you merge PRs the original commit message is lost and there's also no reference to the Issue/PR. Is this deliberate? Should I just skip the commit messages?

I just press merge when > 1 commit or rebase when 1 commit, same policy as in systemd. No clue what github does under the hood

LaurenceKiln:
Looks like "Rebase and merge" will not add a link to the PR, possibly because the rebase changes the commit hashes. That's idiotic. 

https://github.com/orgs/community/discussions/62118

mkosi/__init__.py
```diff
@@ -1491,18 +1491,28 @@ def build_microcode_initrd(context: Context) -> list[Path]:
...
+def finalize_kernel_modules_include(context: Context, *, include: Sequence[str], host: bool) -> list[str]:
+    final = []
+
+    host_included = False
+    if host:
+        final.extend(loaded_modules())
+        host_included = True
+
+    for p in include:
+        if "default" in p:
```
daandemeyer: This should be an equality check, not a substring check

mkosi/__init__.py
```diff
@@ -1491,18 +1491,28 @@ def build_microcode_initrd(context: Context) -> list[Path]:
...
+            with chdir(context.resources / "mkosi-initrd"):
+                # TODO: figure out a way to propagate all relevant settings, not just arch
+                _, _, [initrd] = parse_config(
+                    ["--architecture", str(context.config.architecture)],
+                    resources=context.resources,
+                )
+            final.extend(initrd.kernel_modules_include)
+        elif "host" == p and not host_included:
```
daandemeyer: Don't do yoda style conditions, put the variable first and the constant second

mkosi/__init__.py
```diff
@@ -1491,18 +1491,28 @@ def build_microcode_initrd(context: Context) -> list[Path]:
...
-        with chdir(context.resources / "mkosi-initrd"):
-            # TODO: figure out a way to propagate all relevant settings, not just arch
-            _, _, [initrd] = parse_config(
-                ["--architecture", str(context.config.architecture)],
-                resources=context.resources,
-            )
-        final.update(initrd.kernel_modules_include)
-    if host or "host" in include:
-        final.update(loaded_modules())
+def finalize_kernel_modules_include(context: Context, *, include: Sequence[str], host: bool) -> list[str]:
```
behrmann, line 1494: With the switch from `set[str]` to `list[str]` deduplication is lost. A way to achieve both deduplication and order would be to exchange the set for a dictionary, with e.g. only `None` keys. This way the tracking of `host_included` would also not be needed.

LaurenceKiln: I *think* downstream code is indifferent to this, and I know the PR I'm working on definitely does. But you're right, and neatness counts.

daandemeyer:
@LaurenceKiln Please amend commits instead of pushing updates. We want to merge perfect commits and I don't have the muscle memory to remember I need to squash yours.

LaurenceKiln:
understood. not too late to force-push a correction.

daandemeyer:
> not too late to force-push a correction.

People frown upon it so I don't do it, I also hope we have branch protection on.

Also, massive thanks for all the quality contributions! Great to see someone picking up the pace as I haven't been doing too much lately.

LaurenceKiln:
Thanks for the kudos. I came across the project it and it just grabbed me.

PR #4017, opened 2025-12-02
Optimize kmod.py::resolve_module_dependencies()

LaurenceKiln:
Instead of retrieving the modinfo for all modules on the system
(around 4700 modules for 6.17.8-300.fc43.x86_64), retrieve only
the modules specified by the user, and then transitively fetch
the `modinfo` for dependencies.

When all modules are included, it makes no difference.

But, when the user significantly reduces the set of included modules,
the cost of multiple calls to `modinfo` while chasing the dependency
chains is much lower than the cost of querying and parsing all the
kernel modules on the system, for data that is never used.

For example, `KernelModules=default` pulls in around 570 modules.
With these changes the wall time for this build step is much reduced:
- 9sec -> 5sec on ramdisk.
- ~~85~~ 25sec  -> 5sec (!) on a slow HDD.

### DIFF
--- a/mkosi/kmod.py
+++ b/mkosi/kmod.py
@@ -7,7 +7,7 @@ import logging
 import os
 import re
 import subprocess
-from collections.abc import Iterable, Iterator, Reversible
+from collections.abc import Generator, Iterable, Iterator, Reversible
 from pathlib import Path
 
 from mkosi.context import Context
@@ -18,10 +18,28 @@ from mkosi.util import chdir, flatten, parents_below
 
 
 @dataclasses.dataclass(frozen=True)
-class ModuleDependencyInfo:
-    name: str
-    modules: list[str]
-    firmware: list[str]
+class Dependencies:
+    hard: dict[str, set[str]]
+    soft: dict[str, set[str]]
+    fw: dict[str, set[Path]]
+
+    def update(self, deps_info: "Dependencies") -> None:
+        self.hard.update(deps_info.hard)
+        self.soft.update(deps_info.soft)
+        self.fw.update(deps_info.fw)
+
+    def all_mods(self) -> set[str]:
+        return self.hard.keys() | self.soft.keys()
+
+    def all_moddeps(self) -> set[str]:
+        return set(flatten(self.hard.values())) | set(flatten(self.soft.values()))
+
+    def all_moddeps_typed(self) -> Generator[tuple[str, str, str]]:
+        for label, collection in [("soft", self.soft), ("hard", self.hard)]:
+            yield from ((m, label, dep) for m, deps in collection.items() for dep in deps)
+
+    def all_fwdeps(self) -> set[Path]:
+        return set(flatten(self.fw.values()))
 
 
 def loaded_modules() -> list[str]:
@@ -242,11 +260,14 @@ def modinfo(context: Context, kver: str, modules: Iterable[str]) -> str:
 def parse_modinfo_output(
     context: Context,
     info: str,
-) -> Iterable[ModuleDependencyInfo]:
-    result = []
+) -> Dependencies:
+    """Parse modinfo output into structured dependency data"""
+    result = Dependencies(hard={}, soft={}, fw={})
 
     depends: set[str] = set()
+    soft_depends: set[str] = set()
     firmware: set[Path] = set()
+    missing_firmware: set[str] = set()
 
     with chdir(context.root):
         for line in info.split("\0"):
@@ -262,22 +283,32 @@ def parse_modinfo_output(
             elif key == "softdep":
                 # softdep is delimited by spaces and can contain strings like pre: and post: so discard
                 # anything that ends with a colon.
-                depends.update(normalize_module_name(d) for d in value.split() if not d.endswith(":"))
+                soft_depends.update(normalize_module_name(d) for d in value.split() if not d.endswith(":"))
 
             elif key == "firmware":
-                if (Path("usr/lib/firmware") / value).exists():
-                    firmware.add(Path("usr/lib/firmware") / value)
-
                 glob = "" if value.endswith("*") else ".*"
+                found = set(Path("usr/lib/firmware").glob(f"{value}{glob}"))
+                if (Path("usr/lib/firmware") / value).exists():
+                    found.add(Path("usr/lib/firmware") / value)
 
-                firmware.update(Path("usr/lib/firmware").glob(f"{value}{glob}"))
+                firmware.update(found)
+                if not found:
+                    missing_firmware.add(value)
 
             elif key == "name":
                 name = normalize_module_name(value)
-                result.append(ModuleDependencyInfo(name, depends, firmware))
+
+                for fw in missing_firmware:
+                    logging.warning(f"{fw} is a fw dependency of {name} but is not installed, ignoring")
+
+                result.hard[name] = depends
+                result.soft[name] = soft_depends
+                result.fw[name] = firmware
 
                 depends = set()
+                soft_depends = set()
                 firmware = set()
+                missing_firmware = set()
 
     return result
 
@@ -302,31 +333,21 @@ def resolve_module_dependencies(
             allmodules = set(modulesd.rglob("*.ko*"))
         nametofile = {module_path_to_name(m): m for m in allmodules}
 
-        moddep: dict[str, set[str]] = {}
-        firmwaredep: dict[str, set[Path]] = {}
-
         modules = {normalize_module_name(m) for m in modules}  # ensure normalized names
         available = builtin | nametofile.keys()
         todo = modules & available
-        done = set()
+        deps = Dependencies(hard={m: set() for m in todo}, soft={}, fw={})
 
         while todo:
-            info = modinfo(context, kver, todo)
+            deps.update(parse_modinfo_output(context, modinfo(context, kver, todo)))
+            todo = (deps.all_moddeps() - deps.all_mods()) & available
 
-            for record in parse_modinfo_output(context, info):
-                moddep[record.name] = record.modules
-                firmwaredep[record.name] = record.firmware
+        mods = (deps.all_mods() - builtin) & available
+        firmware = deps.all_fwdeps()
 
-            done.update(todo)
-            todo = (set(flatten(moddep.values())) - done) & available
-
-        mods = (moddep.keys() - builtin) & available
-        firmware = {fw for fws in firmwaredep.values() for fw in fws}
-
-    for m, depends in moddep.items():
-        for d in depends:
-            if d not in available:
-                logging.warning(f"{d} is a dependency of {m} but is not installed, ignoring.")
+    for m, typ, dep in deps.all_moddeps_typed():
+        if dep not in available:
+            logging.warning(f"{dep} is a {typ} dependency of {m} but is not installed, ignoring.")
 
     logging.debug(f"A total of {len(mods)} kernel modules will be included in the image")
     logging.debug(f"A total of {len(firmware)} firmware files were detected as kmod dependencies")


mkosi/kmod.py
```diff
@@ -235,82 +235,78 @@ def resolve_module_dependencies(
...
     firmware: set[Path] = set()
 
-    with chdir(context.root):
-        for line in info.split("\0"):
-            key, sep, value = line.partition(":")
-            if not sep:
-                key, sep, value = line.partition("=")
+    todo = list(modules) + list(builtin)
+    done = set()
+    MAX_CHUNK_SIZE = 1024
```
daandemeyer: Why not keep the 8500 chunk size?

daandemeyer: Also only used once, no point in having a constant, just inline it.

mkosi/kmod.py
```diff
@@ -235,82 +235,78 @@ def resolve_module_dependencies(
...
+    while todo:
+        chunksize = min(len(todo), MAX_CHUNK_SIZE)
+        chunk, todo = list(todo)[:chunksize], todo[chunksize:]
```
daandemeyer: No need for the `list()` here I think, you already initialize `todo` as a list above

mkosi/kmod.py
```diff
@@ -235,82 +235,78 @@ def resolve_module_dependencies(
...
 
     depends: set[str] = set()
     firmware: set[Path] = set()
 
-    with chdir(context.root):
-        for line in info.split("\0"):
-            key, sep, value = line.partition(":")
-            if not sep:
-                key, sep, value = line.partition("=")
+    todo = list(modules) + list(builtin)
```
daandemeyer: Why bother with a list actually? I'd make this a set

mkosi/kmod.py
```diff
@@ -235,82 +235,78 @@ def resolve_module_dependencies(
...
+    done = set()
```
daandemeyer: Not sure why we have to maintain `done`? Just subtract `chunk` from todo every step and you'll eventually end up empty.

mkosi/kmod.py
```diff
@@ -235,82 +235,78 @@ def resolve_module_dependencies(
...
+                    firmware = set()
 
-    while todo:
-        m = todo.pop()
-        if m in mods:
-            continue
+        done.update(chunk)
+        todo = list(set(todo) - done)
+
+    mods = set(moddep) | set(todo) | set(builtin)
```
daandemeyer: You'll only get here if `todo` is empty, so no bother including it

mkosi/kmod.py
```diff
@@ -235,82 +235,78 @@ def resolve_module_dependencies(
...
+    firmware = {_ for fw_files in firmwaredep.values() for _ in fw_files}
```
daandemeyer: Please use a regular variable name here, even if we don't use it. `fw` should suffice

mkosi/kmod.py
daandemeyer: `modinfo` will return results even for builtin module, so I don't think you need to add `builtin` here either.

DaanDeMeyer (changes requested):
So we have to be slightly careful here because setting up the sandbox with python isn't exactly free so if we invoke modinfo too many times that'll start slowing us down. But since you did the measurements I think we should be fine.

LaurenceKiln:
I've been working on properly testing this, so I haven't addressed the review notes yet.

(Everything below is on fedora)

- I think testing found a bug in the old version. When modules include a non-existent module, the old version silently ignores it (it only checks the deps for missing modules)
- Wrote a POC test to thoroughly compare input/output pairs for old and new implementations. You can run it directly via "./test_new_dep_res.py". It also shows runtime for old/new versions.
- I've verified that:
1. The function outputs match in all cases of interest I could identify (See test). Including testing hundreads of randomly selected groups of modules (including some with firmware)
2. That it outputs the same warnings about dependencies being missing (e.g. cifs2 softdep on `aead2`.
3. That missing modules (not deps) trigger a warning (for now, see `Questions`)

Questions:
- Original version treated soft-dep and deps equally, which I don't like. It converted missing modules into a warning. Shouldn't a missing hard-dep stop the build?
- What's the correct behavior for `KernelModules=no_such_kmod`? warn or fail?

Notes:
- On fedora, `modinfo crc32` returns "crc32-cryptoapi" which is **aliased** to `crc32`, while `modinfo -m crc32` only shows the built-in crc32. The old `modinfo` didn't use `-m` but it doesn't matter because the code looked at builtin's first. To keep this behavior with the new impl, I switched to `modinfo -m`.
- Much of runtime now seems to be  spend in bringing up and tearing down the sandbox for `modinfo`, which comes in at about 2 seconds on my archaic laptop HDD. Probably less noticeable  SSD, but perhaps there's room to optimize sandbox. In any case, `modinfo` seems pretty innocuous to justify spending a majority of its time building a sandbox.

mkosi/kmod.py
LaurenceKiln: We need to maintain done to remove dependencies that have more than one parent from being visited more than once via modinfo. I'll add the missing code.

mkosi/kmod.py
LaurenceKiln: 8500 seemed oddly specific and I didn't understand where it came from.

AFAIK the limit is not on num of args as the orig comment suggested, but on the  size of combined env+cmdline.

`getconf ARG_MAX -> 2MiB `. Fedora ships with about 5000 modules, average name length is around 12. `12*5000=60kb` so we're not even close to half the limit. 

I already upped this to 10000 once I realized the sandbox was expensive. Anything larger than the number of modules (with future headroom) would be exactly the same. 

It's behind a constant because bare magic numbers in code just look wrong to me. But I'll inline it.

mkosi/kmod.py
LaurenceKiln: Because we need to do set difference, and slicing, and neither type can do both.  It's a set now.

mkosi/kmod.py
```diff
@@ -220,6 +220,66 @@ def modinfo(context: Context, kver: str, modules: Iterable[str]) -> str:
...
+) -> tuple[dict[str, set[str]], dict[str, set[Path]], set[str]]:
+    moddep = {}
+    firmwaredep = {}
+    skipped = set()
+
+    depends: set[str] = set()
+    firmware: set[Path] = set()
+
+    todo = set(modules)
+    done = set()
```
behrmann: Needs a typo annotation.

test_new_dep_res.py
```diff
@@ -0,0 +1,129 @@
+#!/usr/bin/env python3
+
+# This is a throwaway. Tell mypy to ignore
```
behrmann: If this is a throaway, it doesn't really need to be in the PR, but if it's a test that establishes correctness, then it should be a proper test.

LaurenceKiln:
> So we have to be slightly careful here because setting up the sandbox with python isn't exactly free so if we invoke modinfo too many times that'll start slowing us down. But since you did the measurements I think we should be fine.

Right.  From random testing, the dependency chain can get up to around 5. 

At least here that's still much faster than querying all modules on the system. Plus, for the default case with all modules (with a large enough chunk size) there's still just one modinfo call.

The obvious solution is to lift the sandbox to encompass all the modinfo calls, since sandbox creation is now the bottleneck.

Do run the test/bench and see if the tradeoff is different on your machine. My laptop is old and I can't rely on it to predict perf for common user.

mkosi/kmod.py
```diff
@@ -239,6 +299,51 @@ def resolve_module_dependencies(
...
+
+    for name in skipped:
+        if name not in builtin:
+            # FIXME: should abort?
+            logging.warning(f"{name} was requested but is not installed, ignoring.")
+
+    logging.debug(f"The following {len(mods)} kernel modules will be included in image:")
+    logging.debug(str(mods))
+    logging.debug(f"The following {len(mods)} firmware files will be included in image:")
+    logging.debug(str(firmware))
```
behrmann: I don't think this needs separate coercion to string.
```suggestion
    logging.debug(mods)
    logging.debug(f"The following {len(mods)} firmware files will be included in image:")
    logging.debug(firmware)
```

mkosi/kmod.py
```diff
@@ -239,6 +299,51 @@ def resolve_module_dependencies(
         allmodules = set(modulesd.rglob("*.ko*"))
     nametofile = {module_path_to_name(m): m for m in allmodules}
 
+    log_step("Resolving dependencies for kernel modules and firmware")
+
+    moddep, firmwaredep, skipped = get_deps(context, kver, modules, builtin, set(nametofile))
+
+    mods = set(normalize_module_name(_) for _ in moddep) - set(builtin)
```
```suggestion
    mods = {normalize_module_name(m) for m in moddep} - set(builtin)
```

mkosi/kmod.py
```diff
@@ -239,6 +299,51 @@ def resolve_module_dependencies(
         allmodules = set(modulesd.rglob("*.ko*"))
     nametofile = {module_path_to_name(m): m for m in allmodules}
 
+    log_step("Resolving dependencies for kernel modules and firmware")
+
+    moddep, firmwaredep, skipped = get_deps(context, kver, modules, builtin, set(nametofile))
```
behrmann: dict keys already behave like a set, so this should work, probably by typing `get_deps` as taking a `collections.abc.Set` there.
```suggestion
    moddep, firmwaredep, skipped = get_deps(context, kver, modules, builtin, nametofile.keys())
```

mkosi/kmod.py
```diff
@@ -220,6 +220,66 @@ def modinfo(context: Context, kver: str, modules: Iterable[str]) -> str:
     ).stdout.strip()
 
 
+def get_deps(
+    context: Context, kver: str, modules: Iterable[str], builtin: set[str], loadable_modules: set[str]
+) -> tuple[dict[str, set[str]], dict[str, set[Path]], set[str]]:
```
behrmann: The return type for this is completely unwieldy, this could really do with some type aliases.

mkosi/kmod.py
```diff
@@ -220,6 +220,66 @@ def modinfo(context: Context, kver: str, modules: Iterable[str]) -> str:
     ).stdout.strip()
 
 
+def get_deps(
+    context: Context, kver: str, modules: Iterable[str], builtin: set[str], loadable_modules: set[str]
```
behrmann: Please add a comma after the last argument, so ruff splits all the arguments.

mkosi/kmod.py
```diff
@@ -220,6 +220,66 @@ def modinfo(context: Context, kver: str, modules: Iterable[str]) -> str:
...
+        skipped.update(todo - loadable_modules - builtin)
+        todo = todo - done - skipped
+        if not todo:
+            break
+        chunksize = min(len(todo), 10000)
+        chunk = set(list(todo)[:chunksize])
+        info = modinfo(context, kver, chunk)
+
+        with chdir(context.root):
+            for line in info.split("\x00"):
```
behrmann: The majority of times we use `\0` for null.

test_new_dep_res.py
LaurenceKiln: It's not meant to be merged since if the PR is accepted the old code will go away anyway. I wanted to share it with you so you can check performance for yourself, and to convince you (and me) that the code is correct, since there's some subtle corner cases involved.

mkosi/kmod.py
LaurenceKiln: It needs to be a set because I use it for set difference. And I only need the keys, so this is the right place to convert type.

mkosi/kmod.py
LaurenceKiln: not pretty. ruff says Type aliases were added in 3.12 and pyproject.toml says "requires-python = ">=3.9".

daandemeyer:
I'm not gonna bother reviewing the version with the changes to make your testing script work. I'll trust your measurements and verify myself with a quick image build after your stuff is merged. So ping me when you're happy with your measurements and I'll be happy to review

LaurenceKiln:
Ok. I wouldn't like to merge this until I've figured out a way to share a chroot between multiple calls.

daandemeyer:
> Ok. I wouldn't like to merge this until I've figured out a way to share a chroot between multiple calls.

That's going to be non-trivial change. I'm not opposed if the numbers show a big win, but let's please get this working without the chroot reuse and then add the chroot reuse in a second PR.

mkosi/kmod.py
behrmann: But views of dictionary keys (`dict.keys()`) support set operations and are an `collections.abc.Set`.

mkosi/kmod.py
behrmann: True, add a docstring explaining what is returned then. 

Related `get_deps` is a very generic name, `get_module_deps` would be more concrete.

LaurenceKiln:
- Improve performance when KernelModules= restricts the set of included modules
- Warn when a requested module is missing
- Report final module/firmware count and list

Local benchmark on laptop with HDD with btrfs (cached build):
```
% time mkosi -f --kernel-modules=default build 

workspace         Change[sec]
tmpfs (ram)         -8 
HDD                -75
```

mkosi/kmod.py
LaurenceKiln: Got it.

LaurenceKiln:
@behrmann , @DaanDeMeyer, If there are no further review notes, I consider this ready to merge. Removed the testing commit.

mkosi/kmod.py
```diff
@@ -239,78 +303,28 @@ def resolve_module_dependencies(
...
+                skipped.discard(d)
+                logging.warning(f"{d} is a dependency of {m} but is not installed, ignoring.")
+
+    for name in skipped:
+        if name not in builtin:
+            # FIXME: should abort?
+            logging.warning(f"{name} was requested but is not installed, ignoring.")
+
+    logging.debug(f"The following {len(mods)} kernel modules will be included in the image:")
+    logging.debug(sorted(mods))
```
daandemeyer: Same as on the other PR, we used to do this and it was very very noisy.

I added SplitArtifacts=kernel-modules-initrd to split out the kernel modules initrd as a separate artifact which can then be inspected using the usual tools. So I don't think we should log each individual module we're including in the debug logs. Feel free to debug log a message informing users about SplitArtifacts=kernel-modules-initrd instead.

mkosi/kmod.py
```diff
@@ -220,6 +220,70 @@ def modinfo(context: Context, kver: str, modules: Iterable[str]) -> str:
     ).stdout.strip()
 
 
+def get_module_deps(
```
daandemeyer: I don't see a point in splitting this out into its own function (except to accomodate your now removed testing script). So lets please keep this in resolve_module_dependencies()

mkosi/kmod.py
```diff
@@ -203,7 +203,7 @@ def module_path_to_name(path: Path) -> str:
 
 
 def modinfo(context: Context, kver: str, modules: Iterable[str]) -> str:
-    cmdline = ["modinfo", "--set-version", kver, "--null"]
+    cmdline = ["modinfo", "-m", "--set-version", kver, "--null"]
```
daandemeyer: Please use the long option name

DaanDeMeyer (changes requested):

mkosi/kmod.py
LaurenceKiln: it had nothing to do with the test, the function just got too big for my taste. Let me try and improve this by breaking out just the parsing bit.

mkosi/kmod.py
LaurenceKiln: I'd like these to stay. How about a cutoff? 1000 modules is one page of text on a full screen.  Would that be acceptable?

mkosi/kmod.py
LaurenceKiln: Done.

mkosi/kmod.py
```diff
@@ -284,8 +262,8 @@ def resolve_module_dependencies(
                 firmware.update(Path("usr/lib/firmware").glob(f"{value}{glob}"))
 
             elif key == "name":
-                # The file names use dashes, but the module names use underscores. We track the names in
-                # terms of the file names, since the depends use dashes and therefore filenames as well.
+                # The names in `modules` argument have been previously normalized in
+                # `filter_kernel_modules`. We normalize everything to match.
```
behrmann: Nit:
```suggestion
                # filter_kernel_modules. We normalize everything to match.
```

mkosi/kmod.py
```diff
@@ -294,23 +272,70 @@ def resolve_module_dependencies(
...
+    modules: Iterable[str],
+) -> tuple[set[Path], set[Path]]:
+    """
+    Returns a tuple of lists containing the paths to the module and firmware dependencies of the given list
+    of module names (including the given module paths themselves). The paths are returned relative to the
+    root directory.
+    """
+    modulesd = Path("usr/lib/modules") / kver
+    if (p := context.root / modulesd / "modules.builtin").exists():
+        builtin = set(module_path_to_name(Path(m)) for m in p.read_text().splitlines())
```
```suggestion
        builtin = {module_path_to_name(Path(m)) for m in p.read_text().splitlines()}
```

mkosi/kmod.py
```diff
@@ -294,23 +272,70 @@ def resolve_module_dependencies(
...
-        depends = moddep.get(m, set())
+    moddep: dict[str, set[str]] = {}
+    firmwaredep: dict[str, set[Path]] = {}
+    skipped = set()
+
+    todo: set[str] = set(modules)
+    done: set[str] = set()
+    while True:
+        skipped.update(todo - builtin - nametofile.keys())
+        todo = todo - done - skipped
```
```suggestion
        todo.difference_update(done, skipped)
```

mkosi/kmod.py
```diff
@@ -294,23 +272,70 @@ def resolve_module_dependencies(
...
 
-        depends = moddep.get(m, set())
+    moddep: dict[str, set[str]] = {}
+    firmwaredep: dict[str, set[Path]] = {}
+    skipped = set()
+
+    todo: set[str] = set(modules)
+    done: set[str] = set()
+    while True:
+        skipped.update(todo - builtin - nametofile.keys())
```
behrmann: `builtin` and `nametofile` are not assigned in the loop so it should be possible to lift the last line out of the loop and into the definition of `skipped`.
```suggestion
    skipped = builtin | nametofile.keys()

    todo: set[str] = set(modules)
    done: set[str] = set()
    while True:
```

mkosi/kmod.py
```diff
@@ -294,23 +272,70 @@ def resolve_module_dependencies(
...
+
+        moddep_chunk, firmwaredep_chunk = parse_modinfo_output(context, info)
+
+        moddep.update(moddep_chunk)
+        firmwaredep.update(firmwaredep_chunk)
+        done.update(chunk)
+        todo.update(set(flatten(moddep.values())) - done)
+
+    mods = set(moddep) - set(builtin)
+    firmware = {_ for fws in firmwaredep.values() for _ in fws}
```
behrmann: Please, no `_`, give this variable a name like `fw`

mkosi/kmod.py
```diff
@@ -294,23 +272,70 @@ def resolve_module_dependencies(
...
+        info = modinfo(context, kver, chunk)
+
+        moddep_chunk, firmwaredep_chunk = parse_modinfo_output(context, info)
+
+        moddep.update(moddep_chunk)
+        firmwaredep.update(firmwaredep_chunk)
+        done.update(chunk)
+        todo.update(set(flatten(moddep.values())) - done)
+
+    mods = set(moddep) - set(builtin)
```
```suggestion
    mods = set(moddep.keys() - builtin)
```
though
```suggestion
    mods = moddep) - set(builtin
```
might also just work

mkosi/kmod.py
```diff
@@ -294,23 +272,70 @@ def resolve_module_dependencies(
...
-
-        mods.add(m)
-        todo += depends
-        firmware.update(firmwaredep.get(m, set()))
+                skipped.discard(d)
+                logging.warning(f"{d} is a dependency of {m} but is not installed, ignoring.")
+
+    for name in skipped:
+        if name not in builtin:
+            # FIXME: should abort?
```
behrmann: I don't think this should abort, since it doesn't now and it would be a backward compat break.

mkosi/kmod.py
LaurenceKiln: no.

```
#grep -PR "#.+\`"  
resources/man/mkosi.1.md:## Different ways to boot with `vm`
config.py:        # The default value is set in `__init__.py` in `install_uki`.
```

mkosi/kmod.py
LaurenceKiln: nice to know it exists, I don't see the need to change.

mkosi/kmod.py
LaurenceKiln: I don't see it. `todo` is updated by `flatten(moddep.values())`, and we need to remove `builtin` and `nametofile`  from newly learned dependencies so as not to revisit modules. But `skipped` has to start off empty, or:
```
skipped = builtin | nametofile.keys()
while true:
    todo = todo - done - skipped
```
... would leave `todo` empty before each chunk is processed. Remember , `nametofile` is every module found under /lib/modules. 

Am I missing something?

mkosi/kmod.py
LaurenceKiln: `set(moddep) - set(builtin)` is far more readable, and for ~5000 entries it doesn't matter if we use set(dict) instead of `dict.keys()`

Please wind down the style-golfing.

mkosi/kmod.py
LaurenceKiln: It might break backcompat by _alerting_ users they have a typo in their config instead of letting them think the module was found.

The question is whether there's a legitimate use case for users to ask for a module but want to ignore it if missing. Can you think of one?

LaurenceKiln:
- Warn early about missing requested modules.
- No need to explicitly keep track of skipped.
- Removed chunking logic. cmdline limit allows well over 100k+ typical-length names. 
- Simplified loop.

LaurenceKiln:
- Distinguish soft and hard dependencies for warnings
- parse_modinfo_output returns a TypedDict

LaurenceKiln:
- warn about missing firmware files. This can be quite noisy since` no firmware packages are installed unless the user explicitly included them in `Packages=`, but I think we do want to be loud about missing firmware. Previously, there was no indication that firmware might be missing.

mkosi/kmod.py
behrmann: I don't think the grep is a good metric for this, since of all comprehensions, set comprehensions are the rarest cases and of all builtin datatype constructors set is probably called the most, since empty sets don't have a constructor, though there is a PEP to change that.

Generally, comprehensions should call the syntax for the comprehension instead of calling the constructor on a generator, because it leads to better performing code (they do produce different Python bytecode). Besides, using the proper syntax being more common and in most cases more readable. 

The exception to the rule is our use of `dict` quite often for environments, since the keys don't need to be stringified and environment variables are valid Python variable identifier.

mkosi/kmod.py
behrmann: This too is a performance thing, the version `todo = todo - done - skipped` produces an intermediary set and throws the initial `todo` away, where the `.difference_update` version doesn't do extra allocations of set objects.

mkosi/kmod.py
LaurenceKiln: I agree that set comprehensions make more sense,  but this is just a nit.
But you've very diligent (good) about style issues, and it can be aggravating to get ever more corrections for "violations" that are common in existing code or were actually copied from from the previous version of the code.

The perf diff here is completely negligible ("premature optimization..."). This PR code cuts off 8 seconds off runtime on ramdisk. I'd be amazed if this change matters for perf beyond millisecs.

I made the change everywhere so we don't need to argue the point anymore. Please resolve if there's nothing else.

mkosi/kmod.py
LaurenceKiln: Which again is correct in a strict sense, but not relevant here. I had to look up `difference_update`. A-B-C does not require explanation.

calculating difference with `-` operator is just... Jörg, come on.

LaurenceKiln:
Updated to fix 4053

mkosi/kmod.py
```diff
@@ -268,51 +251,102 @@ def resolve_module_dependencies(
...
+    nametofile = {module_path_to_name(m): m for m in allmodules}
+
+    log_step("Resolving dependencies for kernel modules and detecting needed firmware")
+
+    moddep: dict[str, set[tuple[str, str]]] = {}
+    firmwaredep: dict[str, set[Path]] = {}
+
+    modules = {normalize_module_name(m) for m in modules}  # ensure normalized names
+    available = builtin | nametofile.keys()
+    todo = set(modules) & available
```
behrmann: `modules` is a set already.
```suggestion
    todo = modules & available
```

mkosi/kmod.py
```diff
@@ -8,12 +8,18 @@
...
 from mkosi.context import Context
 from mkosi.log import complete_step, log_step
 from mkosi.run import chroot_cmd, run
 from mkosi.sandbox import chase
-from mkosi.util import chdir, parents_below
+from mkosi.util import chdir, flatten, parents_below
+
+
+class DepsDict(TypedDict):
+    moddep: dict[str, set[tuple[str, str]]]  # name: {(h)ard/(s)oft, name),...}
```
behrmann: Let's not make this more nested and more obscure, i.e. please don't just make the keys `"h"` and `"s"`, as this makes this harder to reason about, just call it `"hard"` and `"soft"`, also, this would probably benefit from flattening this out, i.e. having a key `moddep_hard`, whose values is a `dict[str, set[str]]`, and one `moddep_soft` whose values are also that.

mkosi/kmod.py
```diff
@@ -268,51 +251,102 @@ def resolve_module_dependencies(
...
+
+    moddep: dict[str, set[tuple[str, str]]] = {}
+    firmwaredep: dict[str, set[Path]] = {}
+
+    modules = {normalize_module_name(m) for m in modules}  # ensure normalized names
+    available = builtin | nametofile.keys()
+    todo = set(modules) & available
+    done = set()
+
+    for m in set(modules) - available:
```
```suggestion
    for m in modules - available:
```

mkosi/kmod.py
```diff
@@ -268,51 +251,102 @@ def resolve_module_dependencies(
...
+    mods = (moddep.keys() - builtin) & available
+    firmware = {fw for fws in firmwaredep.values() for fw in fws}
 
-        depends = moddep.get(m, set())
-        for d in depends:
-            if d not in nametofile and d not in builtin:
-                logging.warning(f"{d} is a dependency of {m} but is not installed, ignoring ")
+    for m, depends in moddep.items():
+        for typ, name in depends:
+            typ = dict(s="soft", h="hard").get(typ, "?")
```
behrmann: As written above, the typed dict would probably benefit from splitting the `moddep` key and I really think the type shouldn't just be `"s"` and `"h"`, since it's only ever used here, but for completeness
```suggestion
            typ = {"s": "soft", "h": "hard"}[typ]
```
since it's a clear constant and could be more easily optimised as such from a smarter compiler (currently it also gets constructed on every loop iteration, but it's still 30% faster) and since you construct the dict, and only ever fill it with soft and hard deps, there is nochance of something else coming out.

mkosi/kmod.py
```diff
@@ -8,12 +8,18 @@
...
 
 from mkosi.context import Context
 from mkosi.log import complete_step, log_step
 from mkosi.run import chroot_cmd, run
 from mkosi.sandbox import chase
-from mkosi.util import chdir, parents_below
+from mkosi.util import chdir, flatten, parents_below
+
+
+class DepsDict(TypedDict):
```
daandemeyer: Please use frozen dataclasses like we use everywhere else.

behrmann: Also, let's make this a dataclass, since we generally use dataclasses for stuff like this.

mkosi/kmod.py
LaurenceKiln: Agreed on the labels. But I disagree about the restructuring. The deps should be grouped by module, not by their type. I could replace `tuple[str, str]` with a namedtuple "record" type if you think that's more readable. or frozen dataclass if that's idiomatic in mkosi.

behrmann: I don't see how grouping this adds value. modinfo itself treats them as two different keys, so there is no point in throwing them together, since the set is also unordered. 

Since this should be moved to a dataclass anyway as well, you could add a method that returns the union of the hard and soft dependencies, if needed.

LaurenceKiln: Yes, I like that idea.

mkosi/kmod.py
daandemeyer: As discussed in the other PR, we can add this behind a `-trace-modules` switch or so. But let's not do that as part of this PR. Either split this out into its own PR or add it to the other logging PR.

mkosi/kmod.py
```diff
@@ -16,6 +17,12 @@
 from mkosi.util import chdir, parents_below
 
 
+@dataclasses.dataclass(frozen=True)
+class DepsInfo:
+    moddep: dict[str, set[str]]
+    firmwaredep: dict[str, set[Path]]
```
daandemeyer: You're leaking how the data is used by the caller into the dataclass.

`parse_modinfo_output()` should return the most simple form representing the data parsed from `modinfo`. It's up to the caller to transform the result into the required data structures that it needs to be able to work with the data.

What this means in practice is that I think `parse_modinfo_output()` should return `list[ModuleDependencyInfo`` where `ModuleDependencyInfo` looks as follows:

```
@dataclasses.dataclass(frozen=True)
class ModuleDependencyInfo:
    name: str
    modules: list[str]
    firmware: list[str]
```

And resolve_module_dependencies() stores the information from that list into its two lookup dicts.

mkosi/kmod.py
```diff
@@ -293,53 +293,51 @@ def resolve_module_dependencies(
...
+    modules = {normalize_module_name(m) for m in modules}  # ensure normalized names
+    available = builtin | nametofile.keys()
+    todo = modules & available
+    done = set()
 
-    moddep = {}
-    firmwaredep = {}
+    for m in set(modules) - available:
+        # FIXME: should abort?
+        logging.warning(f"{m} was requested but is not installed, ignoring.")
```
daandemeyer: Can this actually happen? The modules passed to this function are the result of the globs. If the glob didn't match, the module will not be passed to this function so every module passed to this function should by definition be available?

PR #4018, opened 2025-12-03
tree: Use btrfs ioctls instead of btrfs tool

daandemeyer:
The btrfs ioctls are trivial to use so let's avoid
the overhead of spawning sandboxes and processes
and just use the ioctls directly.

### DIFF
--- a/mkosi/sandbox.py
+++ b/mkosi/sandbox.py
@@ -21,6 +21,10 @@ AT_FDCWD = -100
 AT_NO_AUTOMOUNT = 0x800
 AT_RECURSIVE = 0x8000
 AT_SYMLINK_NOFOLLOW = 0x100
+BTRFS_IOC_SNAP_CREATE_V2 = 0x50009417
+BTRFS_IOC_SUBVOL_CREATE_V2 = 0x50009418
+BTRFS_IOC_SNAP_DESTROY_V2 = 0x5000943F
+BTRFS_SUBVOL_NAME_MAX = 4039
 BTRFS_SUPER_MAGIC = 0x9123683E
 CAP_CHOWN = 0
 CAP_DAC_OVERRIDE = 1
@@ -42,6 +46,7 @@ CLONE_NEWNET = 0x40000000
 CLONE_NEWNS = 0x00020000
 CLONE_NEWUSER = 0x10000000
 EBADF = 9
+ENAMETOOLONG = 36
 EPERM = 1
 ENOENT = 2
 ENOSYS = 38
@@ -81,6 +86,16 @@ SCMP_ACT_ALLOW = 0x7FFF0000
 SCMP_ACT_ERRNO = 0x00050000
 SD_LISTEN_FDS_START = 3
 SIGSTOP = 19
+
+
+class btrfs_ioctl_vol_args_v2(ctypes.Structure):
+    _fields_ = [
+        ("fd", ctypes.c_int64),
+        ("transid", ctypes.c_uint64),
+        ("flags", ctypes.c_uint64),
+        ("unused", ctypes.c_uint64 * 4),
+        ("name", ctypes.c_char * (BTRFS_SUBVOL_NAME_MAX + 1)),
+    ]
 
 
 class mount_attr(ctypes.Structure):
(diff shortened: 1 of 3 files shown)


LaurenceKiln:
BTRFS_IOC_SNAP_DESTROY is marked [obsolete](https://github.com/kdave/btrfs-progs/blob/devel/Documentation/btrfs-ioctl.rst#id25), in favor of BTRFS_IOC_SNAP_DESTROY_V2.

LaurenceKiln:
This is the original. Looks like the naive version doesn't cover nested subvolumes?

https://github.com/kdave/btrfs-progs/blob/745e510b6c82829b9345699db323b9a615a9f539/libbtrfsutil/subvolume.c#L1309-L1342

daandemeyer:
> This is the original. Looks like the naive version doesn't cover nested subvolumes?
> 
> https://github.com/kdave/btrfs-progs/blob/745e510b6c82829b9345699db323b9a615a9f539/libbtrfsutil/subvolume.c#L1309-L1342

Neither did btrfs-progs until recently. Shouldn't be a problem, we don't create nested subvolumes

mkosi/sandbox.py
```diff
@@ -368,6 +381,69 @@ def chattr(path: str, attr: int) -> None:
...
+def btrfs_subvol_create(path: str) -> None:
+    parent = os.path.dirname(path)
+    name = os.path.basename(path)
+    validate_subvol_name(name)
+    fd = os.open(parent, os.O_CLOEXEC | os.O_RDONLY | os.O_DIRECTORY)
+
+    try:
+        args = btrfs_ioctl_vol_args()
+        args.fd = 0
+        args.name = name.encode()
```
behrmann: We get a constructor from ctypes, so we can shorten this
```suggestion
        args = btrfs_ioctl_vol_args(0, name.encode())
```

mkosi/sandbox.py
```diff
@@ -368,6 +381,69 @@ def chattr(path: str, attr: int) -> None:
         raise OSError(r, os.strerror(r), path)
 
 
+def validate_subvol_name(name: str) -> None:
+    if len(name) > BTRFS_PATH_NAME_MAX:
+        raise OSError(ENAMETOOLONG, os.strerror(ENAMETOOLONG), name)
+
+
+def btrfs_subvol_create(path: str) -> None:
```
behrmann: It would be nice to have a common helper, since except for the constant create and delete are the same function.

mkosi/tree.py
```diff
@@ -188,18 +195,20 @@
...
+            btrfs_subvol_delete(os.fspath(p))
+        except OSError as e:
+            if e.errno in (errno.EPERM, errno.EACCES):
+                logging.debug(
+                    f"Could not delete subvolume {p} due to permission issues.\n"
+                    "Consider adding user_subvol_rm_allowed to your btrfs filesystem mount options."
+                )
+            else:
+                logging.debug(f"Failed to delete subvolume {p}: {e}")
+            pass
```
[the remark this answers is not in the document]
behrmann: This indeed needs to be removed.

hundeboll:
Why use `str` for paths in the sandbox module instead of `Path` objects?

`os.open()` takes path-like objects just fine, and you get `Path.name` and `Path.parent` for free?

Then `btrfs_subvol_ioctl()` can be simplified to:
```py
def btrfs_subvol_ioctl(path: Path, cmd: int, src_fd: int = 0) -> None:
    if len(path.name) > BTRFS_SUBVOL_NAME_MAX:
        raise OSError(ENAMETOOLONG, os.strerror(ENAMETOOLONG), path.name)

    fd = os.open(path.parent, os.O_CLOEXEC | os.O_RDONLY | os.O_DIRECTORY)

    # ...
```

daandemeyer:
@hundeboll from the top of sandbox.py:

> Note that this is invoked many times while building the image and as a result, the performance of this script has a substantial impact on the performance of mkosi itself. To keep the runtime of this script to a minimum, please don't import any extra modules if it can be avoided.

behrmann (approved):
Nice, very neat now.

PR #4020, opened 2025-12-04
Improve logging messages for kmod/fw resolution

LaurenceKiln:
- Disambiguate use of the term "include" (include directive/include in image)
- Add logged steps with clearer descriptions: inclusion/exclusion, dep resolution, final count, and debuglog sorted lists whee appropriate

To make the log clearer, and provide information in --debug to help someone fine-tuning KernelModules=+- to see what's happening along the pipeline, without waiting for the final image to build.

Complements #4017.

### DIFF
--- a/mkosi/kmod.py
+++ b/mkosi/kmod.py
@@ -85,10 +85,11 @@ def filter_kernel_modules(
     include: Iterable[str],
     exclude: Iterable[str],
 ) -> list[str]:
+    log_step("Applying kernel modules include/exclude configuration")
     if include:
-        logging.debug(f"Kernel modules include: {' '.join(include)}")
+        logging.debug(f"Kernel modules include directives: {' '.join(include)}")
     if exclude:
-        logging.debug(f"Kernel modules exclude: {' '.join(exclude)}")
+        logging.debug(f"Kernel modules exclude directives: {' '.join(exclude)}")
 
     modulesd = Path("usr/lib/modules") / kver
     with chdir(root):
@@ -137,7 +138,7 @@ def filter_kernel_modules(
         # If no exclude patterns are specified, only keep the specified kernel modules.
         modules = {modulesd / m for m in keep}
 
-    logging.debug(f"Including {len(modules)}/{n_modules} kernel modules.")
+    logging.debug(f"Passing {len(modules)}/{n_modules} kernel modules on to dependency resolution.")
 
     return sorted(module_path_to_name(m) for m in modules)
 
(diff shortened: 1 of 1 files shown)


mkosi/kmod.py
```diff
@@ -175,7 +177,8 @@ def filter_firmware(
             if (patterns and regex.search(rel)) or globs_match_firmware(rel, globs):
                 firmware.add(f)
 
-    logging.debug(f"Including {len(firmware)} firmware files")
+    logging.debug(f"The following {len(firmware)} firmware files will be included in the image:")
+    logging.debug(sorted(firmware))
```
daandemeyer: How noisy is this in practice

LaurenceKiln: It's debug only and not unlike what's already done at other points.

behrmann: The problem is, that CI runs with debug logs and the debug logs are already quite unwieldy, so it does make a difference whether this adds a few dozen, hundred or thousand lines.

LaurenceKiln: The debug log already shows the list of all modules for dep resolution **twice**, once as `Kernel modules include (directives)` and then as part of the modinfo command line, regardless of list size. 

Why do *that*, but require the user to dig up some obscure option that generates an extra initrd file, and then find out how to inspect its contents for modules, instead of just logging this highly relevant info? it's a debug log - it should be useful, not pretty. I needed this when I was figuring out  https://github.com/systemd/mkosi/pull/4016

The use case I have in mind is a user tweaking KernelModules, in which case they can iterate on a small list of modules, and it's convenient to just have everything in the log while finding the right globs. That's why limiting to reasonable number of modules seems like good solution to me.

daandemeyer: I think it's not too much harder to have a user analyze the resulting kernel modules initrd instead of looking at the debug logs.

We used to do this before and it just turned out to be too noisy. It's nice to have to have when you're doing kernel modules stuff. It really gets in the way when you're doing anything else. And while our other debug logs are generally not too invasive, this stuff is.

LaurenceKiln: This extra logging has been **SO** useful to me already on multiple occasions. I feel like it's worth fighting for.

And in any case, the final outcome would be be **less** output not more.

the firmware list for the default build is just 2/7/24 files depending. And the rest of this PR  changes is mostly wording and some added steps for structuring the logs. So this PR isn't verbose at all.

And the big wall of modules that currently appears in the log is only there because #4017 hasn't been merged yet. With it, the working set of modules is drastically reduced for the default build and so is the debug output.

I just don't see the problem.

daandemeyer: Sorry but I'm not convinced. If we add this it would have to go behind a separate --trace-modules option or such. I don't want it to be done by default when debug logging is enabled.

If you're OK with adding `--trace-modules` and gating this behind that then we can go with that approach

LaurenceKiln: Fair.  I'll move the listings out of both PRs and into a separate `--trace-modules` PR.

PR #4023, opened 2025-12-06
Fix truthy condition

LaurenceKiln:


behrmann (approved):
Good catch.

LaurenceKiln:
copilot in unrelated convo, actually

PR #4025, opened 2025-12-06
Use Path.relative_to instead of Path.parts

LaurenceKiln:


bluca:
This broke booting on my pi3 as it stops https://github.com/systemd/mkosi/pull/4003 from working, I'll file a revert

LaurenceKiln:
my bad.

PR #4026, opened 2025-12-06
Update doc for mkosi ssh to mention sd v256 sshd-vsock

LaurenceKiln:


LaurenceKiln:
one is `--ssh` the other is `mkosi ssh`, they are different things.  A user may read the revised option docs and understand that systemd already provides the sshd. But It's not 100% obvious from that  `mkosi ssh` will know how to connect when systemd provides the service, since the credentials aspect isn't explained.

Better to explicitly document and not require users to piece the puzzle together.

PR #4028, opened 2025-12-06
portable: support split roothash

bluca:
Portable images are DDIs too, support writing a split roothash as with OS images


mkosi/__init__.py
```diff
@@ -3807,6 +3807,11 @@ def make_extension_or_portable_image(context: Context, output: Path) -> None:
             if p.split_path:
                 maybe_compress(context, context.config.compress_output, p.split_path)
 
+    if ArtifactOutput.roothash in context.config.split_artifacts and (
```
daandemeyer: Let's split this out in a separate function `write_split_roothash()` or so that takes context and the list of partitions

PR #4030, opened 2025-12-07
initd: always compress kmod initrd

bluca:
Compress with level '-0' on newer distributions where modules are individually compressed.
Concatenating compressed and uncompressed cpio archives in the same file makes it impossible to extract them again to inspect the content, as zstd chokes when a zstd and non-zstd content is in the same file, and cpio chokes when the opposite happens.

By using -0 and always having zstd content, one can just pipe unzstd into cpio and it works fine.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1541,19 +1541,19 @@ def build_kernel_modules_initrd(context: Context, kver: str) -> Path:
         sandbox=context.sandbox,
     )
 
-    if context.config.distribution.is_apt_distribution():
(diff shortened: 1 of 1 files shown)


daandemeyer:
I really don't think this is the way to go. The existing tools for working with initrds from mkinitcpio and dracut (or maybe a new tool) should just learn to deal with these kind of setups properly.

Let's close this for now as one can specify SplitArtifacts=kernel-modules-initrd as a workaround to get the kernel modules initrd as a split artifact on image builds. That doesn't cover everything but should be good enough until a comprehensive solution can be implemented.

PR #4032, opened 2025-12-07
Revert "Use Path.relative_to instead of Path.parts"

bluca:
This stopped https://github.com/systemd/mkosi/pull/4003/commits/42e31875e31a20ad21776b290ac99d2029d31b72 from working, which breaks booting on the PI3

Reverts systemd/mkosi#4025

### DIFF
--- a/mkosi/kmod.py
+++ b/mkosi/kmod.py
@@ -107,7 +107,7 @@ def filter_kernel_modules(
         globs = [normalize_module_glob(p) for p in include if not p.startswith("re:")]
 
         for m in modules:
-            rel = os.fspath(m.relative_to(modulesd))
+            rel = os.fspath(Path(*m.parts[5:]))
 
             if (patterns and regex.search(rel)) or globs_match_module(normalize_module_name(rel), globs):
                 keep.add(rel)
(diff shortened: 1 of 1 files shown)


LaurenceKiln:
hang on. can you try 
`os.fspath(f.relative_to(firmwared.parent))`
instead of 
`os.fspath(f.relative_to(firmwared))`

for some reason, the logic is different for modules and firmware.

LaurenceKiln:
does changing 
```
/drivers/mmc/host/
```
to 
```
/kernel/drivers/mmc/host/
# or
drivers/mmc/host/
```
fix it?

I didn't realize the old code was dropping the `kernel/` part from `/usr/lib/module/KVER/kernel/<RELA_PATH>`.

So #4025 breaks the glob that relies on absolute paths, e.g  `/drivers/mmc/host` (see [`globs_match_filename`](https://github.com/systemd/mkosi/blob/cbb1daeb76ed662db08f2253a347d3dd8c1621d6/mkosi/kmod.py#L26)) that are now identified as `/kernel/drivers`. 

This was not a deliberate break, but the old behavior is problematic since only in-tree modules live under `kernel/`. Out-of-tree modules can live under `KVER/update/` or `KVER/extra` for example. There would be no way I think to include `/extra` for example, since that would wrongly look for such a dir it under `kernel/`.

bluca:
Existing configuration cannot be broken, it needs to continue working

LaurenceKiln:
suggested PR fix in #4035, as alternative to reverting.

daandemeyer:
Merging this for now because I need time to think about the proper fix, feel free to add this back to your other PR

PR #4033, opened 2025-12-07
addon: search for stub in tools tree or host if there is no image

bluca:
When building a kernel cmdline addon there's no image and no packages, so there's no stub to be found. Search for it on the tools tree first, and if not there either, on the host.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2174,7 +2174,25 @@ def install_uki(
 
 def systemd_addon_stub_binary(context: Context) -> Path:
     arch = context.config.architecture.to_efi()
+
     stub = context.root / f"usr/lib/systemd/boot/efi/addon{arch}.efi.stub"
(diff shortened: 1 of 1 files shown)


bluca:
That doesn't work, there's no tools tree on obs builds so currently building addons there is broken

daandemeyer:
Let's take a step back, what's the problem with just installing systemd-stub in the image when using the addon output to make sure the stub is available?

bluca:
There's no image, it's just a kernel cmdline

daandemeyer:
> There's no image, it's just a kernel cmdline

Sure we'll throw away the image at the end but that doesn't prevent you from adding `Packages=systemd-boot-efi` to fetch the stub so we can pick it up before we throw away the image.

bluca:
Adding packages means an initrd image is built into the addon, which is wrong when I just want a cmdline addon

PR #4035, opened 2025-12-07
Match KernelModules= globs relative to modules root dir

LaurenceKiln:
**Update:** folded in the reverted changes from #4025

#4025 broke back-compat and was reverted. My bad.

The old code, matched globs relative to `/lib/modules/KVER/kernel`. I didn't realize #4025 changed this behavior and now matches relative to `/lib/modules/KVER` . 
This broke absolute globs like `/drivers/mmc/host`.

One could argue that the behavior is #2025 is more correct.
Although the old convention saves us from prepending `/kernel` to every absolute globs, it ignores the fact that `kernel/` only holds *in-tree* modules. out-of-tree modules (DKMS, BSPs for embedded)  generally live in sibling directories like `updates/` or `extra/` or `vendor/`. 

Also, If you cat `/lib/modules/$(uname -r)/modules.dep`,  you'll see that it treats modules on the system relative to  `/lib/modules/KVER`, not `/lib/modules/KVER/kernel`. That's the text equivalent of what `modprobe` looks at. 

This patch reapplies #4025, originally meant as a code style issue, and adds a condition to `globs_match_filename` that tries to match absolute globs under `kernel/` first so that existing configs continue to work without modification. This should be fine since generally there will be no collisions (e.g. there is no `kernel/updates` or `kernel/extra`).

Note: in this case as well, adding the extra logging from #4017 and #4020 were very helpful.

### DIFF
--- a/mkosi/kmod.py
+++ b/mkosi/kmod.py
@@ -42,8 +42,10 @@ def globs_match_filename(
             glob += "*"
 
         if (
-            # match the full path
-            (glob.startswith("/") and fnmatch.fnmatch(f"/{name}", glob))
+            # Match globs starting with / relative to kernel/ first, since in-tree module are the common case
+            (glob.startswith("/") and fnmatch.fnmatch(f"/{name}", f"/kernel{glob}"))
+            # Now match absolute globs relative to lib/modules/KVER/
+            or (glob.startswith("/") and fnmatch.fnmatch(f"/{name}", glob))
             # match a subset of the path, at path element boundary
             or ("/" in glob and fnmatch.fnmatch(f"/{name}", f"*/{glob}"))
             # match the basename
@@ -107,9 +109,13 @@ def filter_kernel_modules(
         globs = [normalize_module_glob(p) for p in include if not p.startswith("re:")]
 
         for m in modules:
-            rel = os.fspath(Path(*m.parts[5:]))
+            rel = os.fspath(m.relative_to(modulesd))
+            # old regexes match relative to modulesd/subdir/ not modulesd/
+            legacy_rel = os.fspath(Path(*m.parts[5:]))
 
-            if (patterns and regex.search(rel)) or globs_match_module(normalize_module_name(rel), globs):
+            if (patterns and regex.search(legacy_rel)) or globs_match_module(
+                normalize_module_name(rel), globs
+            ):
                 keep.add(rel)
 
     if exclude:
@@ -119,8 +125,11 @@ def filter_kernel_modules(
 
         remove = set()
         for m in modules:
-            rel = os.fspath(Path(*m.parts[5:]))
-            if rel not in keep and regex.search(rel):
+            rel = os.fspath(m.relative_to(modulesd))
+            # old regexes match relative to modulesd/subdir/ not modulesd/
+            legacy_rel = os.fspath(Path(*m.parts[5:]))
+
+            if rel not in keep and regex.search(legacy_rel):
                 remove.add(m)
 
         modules -= remove
@@ -145,6 +154,8 @@ def filter_firmware(
     if exclude:
         logging.debug(f"Firmware exclude: {' '.join(exclude)}")
 
+    firmwared = Path("usr/lib/firmware")
+
     # globs can be also used to exclude firmware, so we we need to apply them
     # to the inherited list of firmware files too.
     globs = [p for p in include if not p.startswith("re:")]
@@ -156,14 +167,13 @@ def filter_firmware(
         regex = re.compile("|".join(patterns))
 
         for f in firmware:
-            rel = os.fspath(Path(*f.parts[3:]))
+            rel = os.fspath(f.relative_to(firmwared))
             if (patterns and regex.search(rel)) or not globs_match_firmware(rel, globs, match_default=True):
                 remove.add(f)
 
         firmware -= remove
 
     if include:
-        firmwared = Path("usr/lib/firmware")
         with chdir(root):
             all_firmware = {p for p in firmwared.rglob("*") if p.is_file() or p.is_symlink()}
 
@@ -171,7 +181,7 @@ def filter_firmware(
         regex = re.compile("|".join(patterns))
 
         for f in all_firmware:
-            rel = os.fspath(Path(*f.parts[3:]))
+            rel = os.fspath(f.relative_to(firmwared))
             if (patterns and regex.search(rel)) or globs_match_firmware(rel, globs):
                 firmware.add(f)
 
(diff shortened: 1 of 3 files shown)


LaurenceKiln:
@bluca , I'd appreciate it if you could verify this works for you.

bluca:
Please test it yourself by building with `--architecture arm64` and ensuring those arm-specific modules are there

LaurenceKiln:
I verified that `/drivers/mmc/host` glob works. If you don't want to test, don't.

mkosi/kmod.py
```diff
@@ -42,8 +42,10 @@ def globs_match_filename(
             glob += "*"
 
         if (
+            # for backcompat, allow absolute paths relative to resolve relative /kernel/ dir
```
daandemeyer: Let's not say this is for backwards compat. Out of tree modules are a relatively niche use case. I think it makes sense to look in the kernel/ directory by default.

daandemeyer: This also needs to be documented in the `KernelModules=` documentation that paths are now searched relative to /usr/lib/modules/KVER and /usr/lib/modules/KVER/kernel.

mkosi/kmod.py
```diff
@@ -107,7 +109,7 @@ def filter_kernel_modules(
         globs = [normalize_module_glob(p) for p in include if not p.startswith("re:")]
 
         for m in modules:
-            rel = os.fspath(Path(*m.parts[5:]))
+            rel = os.fspath(m.relative_to(modulesd))
```
daandemeyer, line 112: You're still breaking the regex path by making this change without changing the regex logic to deal with /kernel.

LaurenceKiln: Since regexes are soft-deprecated, I added code that simpy preserves the old behavior for regex verbatim.

So `/^drivers\/foo/` keeps working, and `/^vendor\/foo/` keeps **not** working.

LaurenceKiln:
I verified that the firmware paths are not changed by the switch to `relative_to`.

mkosi/resources/man/mkosi.1.md
```diff
@@ -1168,15 +1163,29 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     The modules that were last matched by a positive pattern are included in the image,
     as well as their module and firmware dependencies.
 
-    The module paths are taken relative to the `/usr/lib/modules/<kver>/<subdir>/kernel/` directory,
-    and the `.ko` suffix and compression suffix are ignored during matching.
-    The patterns may include just the basename (e.g. `loop`),
+    The `.ko` suffix and compression suffix are ignored during matching.
+
+    globs are matched against module paths relative to `/usr/lib/module/<kver>`,
```
```suggestion
    Globs are matched against module paths relative to `/usr/lib/module/<kver>`,
```

mkosi/resources/man/mkosi.1.md
```diff
@@ -1168,15 +1163,29 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
+    e.g. the module at `/usr/lib/module/<kver>/kernel/foo/bar.ko.xz`
+    becomes `kernel/foo/bar` for matching purposes.
+
+    globs beginning with "/" are treated specially. The glob is *first* matched
```
```suggestion
    Globs beginning with "/" are treated specially. The glob is *first* matched
```

mkosi/kmod.py
```diff
@@ -42,8 +42,10 @@ def globs_match_filename(
             glob += "*"
 
         if (
-            # match the full path
-            (glob.startswith("/") and fnmatch.fnmatch(f"/{name}", glob))
+            # Match abs globs relto kernel/ first, since in-tree module are the common case
```
```suggestion
            # Match globs starting with / relative to kernel/ first, since in-tree module are the common case
```

mkosi/kmod.py
```diff
@@ -42,8 +42,10 @@ def globs_match_filename(
...
+            (glob.startswith("/") and fnmatch.fnmatch(f"/{name}", "/kernel" + glob))
```
```suggestion
            (glob.startswith("/") and fnmatch.fnmatch(f"/{name}", f"/kernel{glob}"))
```

mkosi/kmod.py
```diff
@@ -42,8 +42,10 @@ def globs_match_filename(
...
+            # Now match abs globs relto lib/modules/KVER/
```
```suggestion
            # Now match absolute globs relative to lib/modules/KVER/
```

mkosi/kmod.py
```diff
@@ -107,9 +109,12 @@ def filter_kernel_modules(
         globs = [normalize_module_glob(p) for p in include if not p.startswith("re:")]
 
         for m in modules:
-            rel = os.fspath(Path(*m.parts[5:]))
+            rel = os.fspath(m.relative_to(modulesd))
+            legacy_rel = os.fspath(Path(*m.parts[5:]))
```
behrmann, line 114: At this point the intent of making this more readable by using `Path.relative_to` instead of `Path.parts` is somewhat lost, because now there's both, but this is handling the actual module paths, that include the `kernel/` prefix, and not the match patterns, so can you point to an example where `rel` and `legacy_rel` would be different, because I can't think of one and couldn't find one on a couple of machines either.  If there is an example, there should be a comment explaining why there is both.

LaurenceKiln: Exactly right. Original PR was meant to be a style fix, which unintentionally revealed a limitation that this PR now addresses. 

Basically, we want to support glob matching for out-of-tree modules (i.e. modules not under `kernel/`).

> can you point to an example where rel and legacy_rel would be different

They are *always* different.
If `m` is `usr/lib/modules/KVER/kernel/drivers/foo.ko.xz` then
- `rel` gives `kernel/drivers/foo` , and 
- `legacy_rel` gives `drivers/foo`  (like the original code did)

legacy_rel is needed to keep existing regex of the form '^drivers/` behaving the same way. See my https://github.com/systemd/mkosi/pull/4035#discussion_r2603272978.

Corrections welcome.

LaurenceKiln: note there's **no leading /** in `m`. I originally missed this when testing the change in a terminal, which is why I thought I wasn't changing anything. Looks like maybe you did too?

behrmann: >> can you point to an example where rel and legacy_rel would be different
>
> They are always different.

Took me a moment to see where my thinking was wrong, but this depends on all paths in `modules` being relative paths. `os.fspath(m.relative_to(modulesd))` and `os.fspath(Path(*m.parts[5:]))` will come out the same for absolute paths because of the leading `/` in the parts.

LaurenceKiln: right, but the code is `for m in Path("usr/lib/modules/<kver>") .rglob("*.ko*")`, so they **are** all relative. Easy to miss and I also did, originally.

Anyway, I added  a comment as you suggested.

PR #4036, opened 2025-12-08
mkosi-obs: do not prepare authvars if there is no ESP to install them in

bluca:


mkosi/resources/mkosi-obs/mkosi.postoutput
```diff
@@ -57,12 +57,19 @@ for f in "${KERNELS[@]}"; do
 done
 
 repart_dir="$(jq -r '.RepartDirectories[-1]' "$MKOSI_CONFIG")"
+if [ -z "$repart_dir" ] || [ "$repart_dir" = "null" ]; then
+    # If there are no specific definitions, try to figure out if any of the predefined ones exist based on the image format
+    format="$(jq -r '.Format' "$MKOSI_CONFIG")"
+    if [ -n "$format" ] && [ "$format" != "null" ] && [ -d "/usr/lib/python3/dist-packages/mkosi/resources/repart/definitions/$format.repart.d" ]; then
+        repart_dir="/usr/lib/python3/dist-packages/mkosi/resources/repart/definitions/$format.repart.d"
+    fi
+fi
```
daandemeyer: Please implement `SplitArtifacts=repart-definitions` instead and use it in mkosi-obs.

bluca: that's way too much effort, I'll just drop this and duplicate the definitions in my project so that they appear in the config

PR #4037, opened 2025-12-08
 Add wireless-regdb package to initrd default conf for fedora

LaurenceKiln:
Debugging logs in #4017 warn that cfg80211 needs the wifi regulatory database as a firmware dependency (to set transmit strengths and such).
```
 6328 Oct 11 03:00 /usr/lib/firmware/regulatory.db
1085 Oct 11 03:00 /usr/lib/firmware/regulatory.db.p7s
```
It's a 50k package. Add it.

Why does early init need wifi? it isn't in /boot/initramfs on fedora.

### DIFF
--- a/mkosi/resources/mkosi-initrd/mkosi.conf
+++ b/mkosi/resources/mkosi-initrd/mkosi.conf
@@ -16,6 +16,7 @@ Packages=
         bash                      # for emergency logins
         less                      # this makes 'systemctl' much nicer to use ;)
         gzip                      # For compressed keymap unpacking by loadkeys
+        wireless-regdb            # fw dep of kmod cfg80211
 
 RemoveFiles=
         # we don't need this after the binary catalogs have been built


daandemeyer:
Please install the package across all distributions

daandemeyer:
@LaurenceKiln So as explained in the other PR, the firmware comes from the rootfs, so this shouldn't be needed. Now wireless-regdb does ship a udev rule and binary which are invoked when the cfg80211 module is loaded, but I would need to see some conclusive evidence that it is not sufficient for wireless-regdb to be installed in the rootfs and that it has to be installed in the initrd for the regdb stuff to work properly.

LaurenceKiln:
I looked into it. Wifi should still work if the db is on the filesystem when the user scans/connects to an AP (a "regulatory hint" which triggers a `require_firmware`). 
If the db is missing entirely, wifi will work but possibly with some features disabled.

If that's your criteria for inclusion, it doesn't need to be in the initrd.

IIUC, If the user install the package in the roots, as a fw dep it will end up in the kmod initrd anyway, no? but if it's not installed, the user won't know until a warning shows up in the journal. Why is it so important to protect the case where the module is installed, but the firmware package isn't?

daandemeyer:
Let's close this for now as it's unclear whether the package needs to be in the initrd or not

LaurenceKiln:
Ok. Note that 4017 adds a warning about missing firmware during kmodinitrd build. Since cfg80211 is part of the default conf, and not part of the main image packages (for each distro) by default, this missing firmware result in a warning every time. Arguably, as it should.

PR #4040, opened 2025-12-08
Add linux-firmware package to default initrd conf

LaurenceKiln:
When installing a kernel, we generally want the firmware to be there as well. mkosi prunes unused blobs during build anyway (in `process_kernel_modules`).

### DIFF
--- a/mkosi/resources/mkosi-initrd/mkosi.conf
+++ b/mkosi/resources/mkosi-initrd/mkosi.conf
@@ -16,6 +16,7 @@ Packages=
         bash                      # for emergency logins
         less                      # this makes 'systemctl' much nicer to use ;)
         gzip                      # For compressed keymap unpacking by loadkeys
+        linux-firmware            # mkosi will remove unused blobs during build
 
 RemoveFiles=
         # we don't need this after the binary catalogs have been built


daandemeyer:
This is completely wrong, the default initrd conf is for the userspace bits, not the modules and firmware (though we do put the default kernel modules list in there as well for convenience). The modules and firmware are bundled separately in the kernel modules initrd. which is assembled from the kernel modules and firmware installed into the actual image, not the initrd image.

LaurenceKiln:
The default initrd conf isn't what generates the actual initrd (what's used before switchroot)? that's really confusing.

daandemeyer:
> The default initrd conf isn't what generates the actual initrd (what's used before switchroot)? that's really confusing.

Sure it does, just not the modules and firmware. Those are put in a separate cpio that's concatenated to the one containing the userspace.

daandemeyer:
> The default initrd conf isn't what generates the actual initrd (what's used before switchroot)? that's really confusing.

Sure it does, just not the modules and firmware. Those are put in a separate cpio that's concatenated to the one containing the userspace.

LaurenceKiln:
I read up on it and I get it now. 

The extra initrd doesn't show up in lsinitrd, or 3cpio, or unkinitramfs either. because they are stricter than the kernel is about extracting the initramfs.

It's been quite difficult to piece all this together: the multi-segment initrd, the tooling issues, how modules end up in the initrd, how firmware files are found,  the file pruning after dep resolution, which conf affects what, availability of `SplitArtifacts=kernel-modules-initrd`, etc'. Half of it I only learned by getting in trouble and getting corrections from you, which doesn't scale.

I'd like to document all this better. Where would you suggest I do it? manpage? Q&A? is there a mkosi blog?

daandemeyer:
@LaurenceKiln I think a docs/ARCHITECTURE.md would be a decent place for this. That'll automatically get rendered on the website as well.

PR #4041, opened 2025-12-08
Warn user when installing a kernel package without linux-firmware package

LaurenceKiln:
When users build a main image, they may use a blank config. If they install kernel packages, `run_depmod` detects a valid module directory and kicks off processing of kmod and dw dependencies. The user almost certainly wants the firmware to be available for whatever kmods they will include, and mkosi will remove any unused blobs anyway.

Detect this situation and warn the user.

This doesn't cover all firmware  since some of it is scattered in distro-specific packages. But this sufficiently informs them that they need to think about installing firmware packages.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -5450,6 +5450,15 @@ def parse_config(
 
     subimages = resolve_deps(subimages, main.dependencies)
 
+    for config in tuple(subimages + [main]):
+        if "linux-firmware" not in config.packages and any(
+            config.distribution.installer.is_kernel_package(p)
+            for p in itertools.chain(config.packages, config.volatile_packages)
+        ):
+            logging.warning(
+                "This config installs a kernel in the image. You probably want to add `linux-firmware` to Packages=."
+            )
+            break
     return args, tools, tuple(subimages + [main])
 
 


daandemeyer:
So for any virtual machine based use cases, which is a substantial amount of what mkosi is used for day to day, there's no need for firmware at all so this is way too opinionated unfortunately.

Also, we try very hard to keep mkosi independent of individual package names, which this also goes against. We do it for kernel package names because there's no other way to figure out if we need to build a default initrd or not, but I don't want to hardcode any more.

Sorry but I think this is just something we're gonna have to live with. mkosi expects its users to be familiar with the distribution that they're building.

LaurenceKiln:
https://packages.fedoraproject.org/pkgs/linux-firmware/
https://archlinux.org/packages/core/any/linux-firmware/
https://launchpad.net/ubuntu/+source/linux-firmware
https://packages.debian.org/trixie/firmware-linux (not linux-firmware)

There is currently no warning during build when a firmware dependency is missing. And you don't want this warning either? I think mkosi could be more helpful about this.

> So for any virtual machine based use cases, which is a substantial amount of what mkosi is used for day to day, there's no need for firmware at all so this is way too opinionated unfortunately.

If a module is installed, it should have its firmware depndencies installed. And for modules that aren't included, the blobs are removed during the build, aren't they?

PR #4046, opened 2025-12-09
Use proper constants for ansi colors

daandemeyer:
I was playing around with mypyc again and it did not like class attributes like the Style ones. While mypyc ended up not working for other reasons, let's switch to something it is happy with which is just regular constants named identically to how they're named in systemd. It's arguably not uglier than the Style namespace class.

### DIFF
--- a/mkosi/sandbox.py
+++ b/mkosi/sandbox.py
@@ -142,19 +142,18 @@ def terminal_is_dumb() -> bool:
     return not sys.stdout.isatty() or not sys.stderr.isatty() or os.getenv("TERM", "") == "dumb"
 
 
-class Style:
-    # fmt: off
-    bold: str   = "\033[0;1;39m"     if not terminal_is_dumb() else ""
-    blue: str   = "\033[0;1;34m"     if not terminal_is_dumb() else ""
-    gray: str   = "\033[0;38;5;245m" if not terminal_is_dumb() else ""
-    red: str    = "\033[31;1m"       if not terminal_is_dumb() else ""
-    yellow: str = "\033[33;1m"       if not terminal_is_dumb() else ""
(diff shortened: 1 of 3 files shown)


mkosi/sandbox.py
```diff
@@ -142,19 +142,18 @@ def terminal_is_dumb() -> bool:
...
-    yellow: str = "\033[33;1m"       if not terminal_is_dumb() else ""
-    reset: str  = "\033[0m"          if not terminal_is_dumb() else ""
-    # fmt: on
+# fmt: off
+ANSI_BOLD:   str = "\033[0;1;39m"     if not terminal_is_dumb() else ""
+ANSI_BLUE:   str = "\033[0;1;34m"     if not terminal_is_dumb() else ""
+ANSI_GRAY:   str = "\033[0;38;5;245m" if not terminal_is_dumb() else ""
+ANSI_RED:    str = "\033[31;1m"       if not terminal_is_dumb() else ""
+ANSI_YELLOW: str = "\033[33;1m"       if not terminal_is_dumb() else ""
+ANSI_RESET:  str = "\033[0m"          if not terminal_is_dumb() else ""
```
behrmann: I just checked and this seems to allow us to drop the annotation
```suggestion
ANSI_BOLD   = "\033[0;1;39m"     if not terminal_is_dumb() else ""
ANSI_BLUE   = "\033[0;1;34m"     if not terminal_is_dumb() else ""
ANSI_GRAY   = "\033[0;38;5;245m" if not terminal_is_dumb() else ""
ANSI_RED    = "\033[31;1m"       if not terminal_is_dumb() else ""
ANSI_YELLOW = "\033[33;1m"       if not terminal_is_dumb() else ""
ANSI_RESET  = "\033[0m"          if not terminal_is_dumb() else ""
```

PR #4050, opened 2025-12-10
run: Set up sandbox with a preexec function

daandemeyer:
There's no need to initialize a new python interpreter every single
time we need a sandbox when we can use the one we already have by
setting up the sandbox with a preexec_fn. A preexec_fn executes before
execve(), so we can reuse the same python interpreter we're already
running in instead of having to spawn a new one.

If we're debugging the sandbox or running a setup command, we stick to
the old approach of invoking a separate python interpreter.

### DIFF
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -18,8 +18,9 @@ from pathlib import Path
 from types import TracebackType
 from typing import TYPE_CHECKING, Any, Callable, Generic, NoReturn, Optional, Protocol, TypeVar
 
+import mkosi.sandbox
 from mkosi.log import ARG_DEBUG, ARG_DEBUG_SANDBOX, ARG_DEBUG_SHELL, die
-from mkosi.sandbox import acquire_privileges, joinpath, umask
+from mkosi.sandbox import SD_LISTEN_FDS_START, acquire_privileges, joinpath, umask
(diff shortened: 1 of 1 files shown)


mkosi/run.py
```diff
@@ -209,21 +213,56 @@ def spawn(
...
+        prefix: list[str]
+        if sbx and not apply_sandbox_in_preexec:
+            module = stack.enter_context(resource_path(sys.modules[__package__ or __name__]))
+            prefix = [
+                *(["strace", "--detach-on=execve"] if ARG_DEBUG_SANDBOX.get() else []),
+                sys.executable, "-SI", os.fspath(module / "sandbox.py"),
+                *sbx,
+            ]  # fmt: skip
+        else:
+            prefix = []
```
behrmann: Since the typing needs to stay outside anyway, we can use this to get rid of the else branch and make this a bit easier on the eyes. 
```suggestion
        prefix: list[str] = []
        if sbx and not apply_sandbox_in_preexec:
            module = stack.enter_context(resource_path(sys.modules[__package__ or __name__]))
            prefix = [
                *(["strace", "--detach-on=execve"] if ARG_DEBUG_SANDBOX.get() else []),
                sys.executable, "-SI", os.fspath(module / "sandbox.py"),
                *sbx,
            ]  # fmt: skip
```

PR #4053, opened 2025-12-10
Treat broken symlinks as missing in kmod/fw resolution

LaurenceKiln:
on fedora `linux-firmware` creates a symlink to a module which is provided by another package, without depending on that package.

If you set a `FirmwareFiles=` glob that matches, or a `KernelModules=` glob that brings in the modules that depends on this firmware, the non-existent path can reach `make_cpio` and cause it to die when called by `build_kernel_modules_initrd`

### DIFF
--- a/mkosi/kmod.py
+++ b/mkosi/kmod.py
@@ -276,12 +276,13 @@ def resolve_module_dependencies(
                 depends.update(normalize_module_name(d) for d in value.split() if not d.endswith(":"))
 
             elif key == "firmware":
-                if (Path("usr/lib/firmware") / value).exists():
+                if (Path("usr/lib/firmware") / value).resolve().exists():
                     firmware.add(Path("usr/lib/firmware") / value)
 
                 glob = "" if value.endswith("*") else ".*"
-
-                firmware.update(Path("usr/lib/firmware").glob(f"{value}{glob}"))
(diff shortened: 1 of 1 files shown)


mkosi/kmod.py
```diff
@@ -276,12 +276,13 @@ def resolve_module_dependencies(
...
-                if (Path("usr/lib/firmware") / value).exists():
+                if (Path("usr/lib/firmware") / value).resolve().exists():
                     firmware.add(Path("usr/lib/firmware") / value)
 
                 glob = "" if value.endswith("*") else ".*"
-
-                firmware.update(Path("usr/lib/firmware").glob(f"{value}{glob}"))
+                found = list(Path("usr/lib/firmware").glob(f"{value}{glob}"))
+                found = [fw for fw in found if fw.resolve().exists()]
+                firmware.update(found)
```
```suggestion
                firmware.update(fw for fw in Path("usr/lib/firmware").glob(f"{value}{glob}") if fw.resolve().exists())
```

DaanDeMeyer (changes requested):
This isn't the right way to fix this, we have to make sure we don't add danling symlink targets when we're resolving symlinks.

https://github.com/systemd/mkosi/pull/4054 is the proper fix, I'll close this one.

PR #4058, opened 2025-12-14
pacman: Make sure hookdir exists before running pacman

daandemeyer:
The latest pacman release seems to fail if the passed hookdir does not exist, so let's make sure it exists before invoking pacman.


daandemeyer:
This is a regression in pacman, I probably won't merge this since a fix has been merged in pacman already: https://gitlab.archlinux.org/pacman/pacman/-/merge_requests/328

daandemeyer:
Yeah Arch already updated to the latest stable branch commit in their pacman package so closing this one, this will fix itself shortly

PR #4061, opened 2025-12-15
Measure and report log_step duration with --debug

LaurenceKiln:
```
‣ Removing output files of default-initrd image… 
...
‣ 0.1s for step "Removing output files of default-initrd image…"
‣ Removing cache entries of default-initrd image…
...
‣ 0.2s for step "Removing cache entries of default-initrd image…"
```

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1527,42 +1527,41 @@ def build_kernel_modules_initrd(context: Context, kver: str) -> Path:
     if kmods.exists():
         return kmods
 
-    log_step("Building kernel modules initrd")
-
-    make_cpio(
-        context.root,
-        kmods,
-        files=gen_required_kernel_modules(
-            context,
-            kver,
-            modules_include=finalize_kernel_modules_include(
+    with complete_step("Building kernel modules initrd"):
+        make_cpio(
+            context.root,
+            kmods,
+            files=gen_required_kernel_modules(
                 context,
-                include=context.config.kernel_modules_initrd_include,
-                host=context.config.kernel_modules_initrd_include_host,
+                kver,
+                modules_include=finalize_kernel_modules_include(
+                    context,
+                    include=context.config.kernel_modules_initrd_include,
+                    host=context.config.kernel_modules_initrd_include_host,
+                ),
+                modules_exclude=context.config.kernel_modules_initrd_exclude,
+                firmware_include=context.config.firmware_include,
+                firmware_exclude=context.config.firmware_exclude,
             ),
-            modules_exclude=context.config.kernel_modules_initrd_exclude,
-            firmware_include=context.config.firmware_include,
-            firmware_exclude=context.config.firmware_exclude,
-        ),
-        sandbox=context.sandbox,
-    )
+            sandbox=context.sandbox,
+        )
 
-    if context.config.distribution.is_apt_distribution():
-        # Older Debian and Ubuntu releases do not compress their kernel modules, so we compress the
-        # initramfs instead. Note that this is not ideal since the compressed kernel modules will
-        # all be decompressed on boot which requires significant memory.
-        if context.config.distribution == Distribution.debian and context.config.release in (
-            "sid",
-            "testing",
-        ):
-            compression = Compression.none
-        else:
-            compression = Compression.zstd
+        if context.config.distribution.is_apt_distribution():
+            # Older Debian and Ubuntu releases do not compress their kernel modules, so we compress the
+            # initramfs instead. Note that this is not ideal since the compressed kernel modules will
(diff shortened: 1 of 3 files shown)


DaanDeMeyer (changes requested):
If you want to measure time taken by a step please use `complete_step()` and put the logic in that context manager instead of introducing globals

LaurenceKiln:
I'd like the debug log to include duration for every step taken during the build. As a way to track where time is spent, and to make it easier for users with different  setups (non-btrfs) to report performance issues, without having to dig into the code.

complete_step itself calls log_step, and most individual step, including some lengthy ones use log_step directly. adding this only to complete_step won't achieve the goal.

like the existing use of `LEVEL` by the logging code, I think module-scoped global are a common way to maintain state between calls, without resorting to classes and instances.

slightly reworked.

behrmann:
> like the existing use of LEVEL by the logging code, I think module-scoped global are a common way to maintain state between calls, without resorting to classes and instances.

Currently we have two globals and it would be preferable to have fewer of them, not more. Globals make things harder to reason about, since you always have to consider the whole history of everything instead have only reasoning about a local thing.

> complete_step itself calls log_step, and most individual step, including some lengthy ones use log_step directly. adding this only to complete_step won't achieve the goal.

I assume, based on your other PRs, your issue is mostly with the calls in `kmod.py`. It would probably make more sense to move these to use `complete_step` instead and keep the timing logic you want to add in `complete_step`.

daandemeyer:
Like @behrmann said, please move all the callsites you want measured to use `complete_step()` instead of `log_step()`

LaurenceKiln:
I've been focusing on kmod, that doesn't mean optimization opportunities can't be found elsewhere. There's value in having a cross-sectional view of everything in order to identify the bottlenecks, and even as a way to track performance over time, if only to pinpoint regressions.

As I said, I would like all steps which merit a call to `log_step` to be measured by default. 

If adding one more global is so unthinkable, we can:
1. switch log_step to be a class instance with a `__call__` method.
2. get rid of  LEVEL and replace it with `len(TIMESTAMP_RECORDS)-1`. That way, there's no change in the number of globals.

behrmann:
> As I said, I would like all steps which merit a call to log_step to be measured by default.

There are currently 16 users of `log_step` that are not `complete_step`. Of these two print the size of outputs, which probably doesn't really need measuring and one would be a tiny bit more involved to move to `complete_step`, because it is conditional (extracting a tar file in `archive.py`), the others should be easily movable to `complete_step`.

LaurenceKiln:
Done. You were right, this makes more sense.

```
$ grep -P 'for step|Building' log2
‣ 0.0s for step "Removing output files of default-initrd image…"
‣ 0.2s for step "Removing output files of main image…"
‣ 0.2s for step "Validating certificates and keys"
‣ Building default-initrd image
‣  0.4s for step "Copying cached trees"
‣  0.1s for step "Copying in extra file trees…"
‣  0.0s for step "Generating system users"
‣  0.1s for step "Generating volatile files"
‣  0.3s for step "Applying presets…"
‣  0.0s for step "Applying first boot settings"
‣  0.5s for step "Generating hardware database"
‣  0.1s for step "Removing files…"
‣  0.4s for step "Creating cpio archive initrd.cpio…"
‣  0.4s for step "Compressing initrd.cpio with zstd"
‣ 2.9s for step "Building default-initrd image"
‣ Building main image
‣  1.9s for step "Copying cached trees"
‣  0.1s for step "Copying in extra file trees…"
‣  0.0s for step "Setting up autologin…"
‣  0.0s for step "Installing systemd-boot…"
‣  0.1s for step "Generating system users"
‣  0.1s for step "Generating volatile files"
‣  0.3s for step "Applying presets…"
‣  9.2s for step "Running modinfo to fetch kernel module dependencies"
‣  0.0s for step "Applying firmware include/exclude configuration"
‣  0.3s for step "Applying kernel module filters"
‣  0.1s for step "Applying first boot settings"
‣  0.5s for step "Generating hardware database"
‣  0.5s for step "Relabeling files using targeted policy"
‣  4.7s for step "Generating disk image"
‣  Building kernel modules initrd
‣   0.5s for step "Running modinfo to fetch kernel module dependencies"
‣   0.0s for step "Applying firmware include/exclude configuration"
‣   0.1s for step "Creating cpio archive kernel-modules-6.17.9-300.fc43.x86_64.initrd…"
‣  0.7s for step "Building kernel modules initrd"
‣  0.9s for step "Generating unified kernel image for kernel version 6.17.9-300.fc43.x86_64"
‣  0.2s for step "Formatting ESP/XBOOTLDR partitions"
‣ 24.4s for step "Building main image"
```

Issue #4039, opened 2025-12-08
Error building - "No such file or directory: 'newroot/tmp'"


JohnMoon-Voyager:
### mkosi commit the issue has been seen with

main

### Used host distribution

Fedora CoreOS 42

### Used target distribution

Ubuntu Noble

### Linux kernel version used

6.15.6-200.fc42.x86_64

### CPU architectures issue was seen on

None

### Unexpected behaviour you saw

During the mkosi.sync phase, I get the following exception:
```
‣ + /work/sync final
Traceback (most recent call last):
  File "/tmp/tmpfcx_mkfw/mkosi/sandbox.py", line 1270, in <module>
  File "/tmp/tmpfcx_mkfw/mkosi/sandbox.py", line 1225, in main
  File "/tmp/tmpfcx_mkfw/mkosi/sandbox.py", line 987, in execute
  File "/tmp/tmpfcx_mkfw/mkosi/sandbox.py", line 197, in mount
  File "/tmp/tmpfcx_mkfw/mkosi/sandbox.py", line 172, in oserror
FileNotFoundError: [Errno 2] No such file or directory: 'newroot/tmp'
```

At the same time, I see a log in dmesg:
```
overlayfs: failed to resolve 'oldroot/var/home/myuser/myproject/tmpfs': -2
```

I added some debug to the `mount` method in `sandbox.py` (just to print the args going into `libc.mount`) and I see:
```
‣ + /work/sync final
None, /, None, 540672, None
b'tmpfs', /tmp, b'tmpfs', 0, None
b'newroot', newroot, None, 20480, None
b'tmpfs', newroot/dev, b'tmpfs', 0, b'mode=0755'
b'oldroot/dev/null', newroot/dev/null, None, 4096, None
b'oldroot/dev/zero', newroot/dev/zero, None, 4096, None
b'oldroot/dev/full', newroot/dev/full, None, 4096, None
b'oldroot/dev/random', newroot/dev/random, None, 4096, None
b'oldroot/dev/urandom', newroot/dev/urandom, None, 4096, None
b'oldroot/dev/tty', newroot/dev/tty, None, 4096, None
b'oldroot/dev/fuse', newroot/dev/fuse, None, 4096, None
b'devpts', newroot/dev/pts, b'devpts', 0, b'newinstance,ptmxmode=0666,mode=620'
b'oldroot/dev/pts/0', newroot/dev/console, None, 4096, None
[...]

PR #4062, opened 2025-12-16
sandbox: Don't make overlay an absolute path if tmpfs

daandemeyer:
Fixes https://github.com/systemd/mkosi/issues/4039


behrmann:
LGTM, but the first commit is already on main

PR #4068, opened 2025-12-16
run.py: stop hiding sandbox and recommend --debug on sandbox failure

marc-hb:
As discussed in #3948 and #3992, logging commands failing in a sandbox without ever mentioning that a sandbox was in use is misleading and very time-consuming; notably because paths look real but they are not.

To remove this confusion and cost, simply add one short line mentioning that a sandbox was involved when the failure happened + recommend to re-run with --debug to display bind mounts and other important information.

Also separate the command and error message on different lines for better readability and re-usability.

Just for the record: before commit 7e7a3c71696d ("Don't log sandbox for every command"), `run.py` was logging sandbox commands every time which was too verbose and unreadable. That commit said "... but still log the full sandbox if a command fails" and it did that do in `__init__.py` - but not always in `run.py`. After that commit, the sandbox was logged by `run.py` only in --debug mode and became completely invisible otherwise.

### DIFF
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -106,14 +106,22 @@ def fork_and_wait(target: Callable[..., None], *args: Any, **kwargs: Any) -> Non
         raise subprocess.CalledProcessError(rc, ["self"])
 
 
+def log_sandbox_issue(sandbox: Sequence[str], cmdline: Sequence[str], errmsg: str) -> None:
+    if ARG_DEBUG.get():
+        # Complete but very long and hard to read. It could be reusable.
+        logging.error(f"{shlex.join(['mkosi-sandbox', *sandbox, *cmdline])}")
+    else:
+        logging.error("sandbox command omitted, rerun with --debug to see full command with bind mounts")
+        logging.error(f"{shlex.join(cmdline)}")
+
+    logging.error(errmsg)
+
+
 def log_process_failure(sandbox: Sequence[str], cmdline: Sequence[str], returncode: int) -> None:
     if -returncode in (signal.SIGINT, signal.SIGTERM):
         logging.error(f"Interrupted by {signal.Signals(-returncode).name} signal")
     elif returncode < 0:
-        logging.error(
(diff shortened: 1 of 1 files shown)


mkosi/run.py
```diff
@@ -106,14 +106,22 @@ def fork_and_wait(target: Callable[..., None], *args: Any, **kwargs: Any) -> Non
         raise subprocess.CalledProcessError(rc, ["self"])
 
 
+def log_sandbox_issue(sandbox: Sequence[str], cmdline: Sequence[str], errmsg: str) -> None:
+    if ARG_DEBUG.get():
+        # Complete but very long and hard to read. It could be reusable.
+        logging.error(f"{shlex.join(['mkosi-sandbox', *sandbox, *cmdline])}")
+    else:
+        logging.error("sandbox command omitted, rerun with --debug to see full command with bind mounts")
+        logging.error(f"{shlex.join(cmdline)}")
```
daandemeyer, line 115: Please stick to a single call to `logging.error()` and use newlines instead. I would keep the original log message the same and if `--debug` is not set, simply append another line that says the following:

"sandbox details omitted, rerun with --debug to see the full sandbox command."

marc-hb: > ... and use newlines instead. I would keep the original log message the same...

So are you ok with printing the error message on a separate line? Adding just one newline. Not 100% clear to me from the above, sorry.

I tested this PR heavily (especially compared to how small of a code change it is...) and NOT having the error message trailing a 5-10 lines-long command made a massive difference. It's quite painful to find before this PR.

I don't mind the code style and the number of logging statements, the terminal output and "user experience" is really the purpose of this.

marc-hb (commented):
Thanks for the prompt review which... I missed! Sorry for the belated reply.

daandemeyer:
@marc-hb I merged a variation of this PR in the linked pr. Let me know if that doesn't address your issue

marc-hb:
That should do it, thx!

PR #4077, opened 2025-12-18
Add support for SplitArtifacts=microcode-initrd

LaurenceKiln:


daandemeyer:
Needs justification

LaurenceKiln:
symmetry. consistency. documentation. inspection.

daandemeyer:
> symmetry. consistency. documentation. inspection.

Not really convinced tbh. We should only add this if there's a clear use case for it.

LaurenceKiln:
It's part of the initrd file (or should be #4078), but its generation/existence wasn't previously logged (until #4076), it's not mentioned as an artifact in the manpage, nor is it built as a separate image. Why treat it differently than the userspace and kmod initrd? 

With the official recommendation in the manpage being to avoid traditional tools to deal with mkosi initrd, plus the difficulties around unpacking them (#4030), it seems like the simplest thing to do is to treat the microcode initrd just like the other components that go into the final initrd. Plus, extending `SplitArtifacts=` this way serves a documentary role, by at least mentioning the extra initrd alongside the kmod initrd.

Will this option see intense use? probly not. Neither do a lot of niche options/features.

behrmann:
> Will this option see intense use? probly not. Neither do a lot of niche options/features.

These options were added in the past because there was at least one use for it, i.e. things were added as needed. While I do like symmetry, the question is whether there is an actual use for this or whether it is just cosmetic.

LaurenceKiln:
I don't see the cost, but the consistency alone is enough for me. Similar things should be behave similarly, for least surprise.

Still, how are users supposed to inspect the microcode initrd? For example, I used this to check the contents of the microcode initrd, and discovered the AMD microcode is not being included for some reason, even though it's installed. Isn't that scenario an "actual use"?

PR #4084, opened 2025-12-22
qemu: Register with systemd-machined in user session

daandemeyer:
Now that machine registration works unprivileged
since systemd v259, let's switch to unconditionally registering machines with the user session
systemd-machined instance.

This breaks compat but the previous implementation arguably wasn't useful or used, since registration would only be done when running as root or if the
Register= feature was explicitly enabled. And if
not running as root, you'd have to authenticate
every time when booting the image to register it
which is arguably too annoying that anyone actually bothered with it.

As vmspawn doesn't yet support registering with the user machined instance, we stop registering vmspawn machines for now. https://github.com/systemd/systemd/pull/40185 will add support for user machined regisration to
vmspawn.

For nspawn we stick with system machined registration for now.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -111,7 +111,6 @@ from mkosi.qemu import (
     copy_ephemeral,
     finalize_credentials,
     finalize_kernel_command_line_extra,
-    finalize_register,
     join_initrds,
     run_qemu,
     run_ssh,
@@ -4228,7 +4227,7 @@ def run_shell(args: Args, config: Config) -> None:
 
     # Underscores are not allowed in machine names so replace them with hyphens.
     name = config.machine_or_name().replace("_", "-")
-    cmdline += ["--machine", name, "--register", yes_no(finalize_register(config))]
+    cmdline += ["--machine", name, "--register", yes_no(config.register != ConfigFeature.disabled)]
 
     with contextlib.ExitStack() as stack:
         for f in finalize_credentials(config, stack).iterdir():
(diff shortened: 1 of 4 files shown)


mkosi/resources/man/mkosi.1.md
```diff
@@ -2029,6 +2029,10 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     systemd-machined. If `auto`, mkosi will register the vm/container
     with systemd-machined if it is available. Defaults to `auto`.
 
+    Note that mkosi will try to register the machine with
+    systemd-machined in the user session. systemd v259 or newer is
+    required to have systemd-machined be available in user sessions.
```
behrmann, line 2034:
```suggestion
    required to have systemd-machined be available in the user sessions.
```

mkosi/__init__.py
```diff
@@ -4228,7 +4227,7 @@ def run_shell(args: Args, config: Config) -> None:
 
     # Underscores are not allowed in machine names so replace them with hyphens.
     name = config.machine_or_name().replace("_", "-")
-    cmdline += ["--machine", name, "--register", yes_no(finalize_register(config))]
+    cmdline += ["--machine", name, "--register", config.register != ConfigFeature.disabled]
```
```suggestion
    cmdline += ["--machine", name, "--register", yes_no(config.register != ConfigFeature.disabled)]
```

mkosi/resources/man/mkosi.1.md
daandemeyer: This is a multiple so no need for "the"

PR #4085, opened 2025-12-22
Add pyrightconfig.json

daandemeyer:
Let's configure the files on which pyright should run to avoid long startup times where it tries to check every single file in the workspace directory.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,6 +54,11 @@ py_version = "39"
 
 [tool.pyright]
 pythonVersion = "3.9"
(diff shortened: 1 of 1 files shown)


behrmann:
Let's keep this config in pyproject.toml. This should look like this
```toml
[tool.pyright]
include = [
    "mkosi/**/*.py",
    "tests/**/*.py",
    "kernel-install/*.install"
]
```

daandemeyer:
Ah, amazing, didn't know that worked

PR #4086, opened 2025-12-23
run: Call execvpe() from preexec function

daandemeyer:
Python does its own executable lookup in $PATH before executing the preexec function, and
hence before we have set up the sandbox which influences the lookup results. To get around
that, let's call execvpe() ourselves inside the preexec() function, and not give Python the
chance to do it itself. This ensures we can do the proper executable lookup after setting
up the sandbox. If we can't find the executable, do nothing, and let Python do its own
search logic so it can return a proper error, which we cannot do from the preexec function.
Note that by doing this we also skip Python closing all open file descriptors except the
ones specified by the user in pass_fds, but since Python opens all file descriptors with
O_CLOEXEC anyway, we'll assume we're good and don't need to close open file descriptors
explicitly.

### DIFF
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -222,9 +222,13 @@ def spawn(
                 preexec()
 
             if sbx and apply_sandbox_in_preexec:
-                # The env passed to subprocess.Popen() replaces the environment wholesale so any
-                # modifications made by mkosi-sandbox would be overridden if we used that. Hence process the
-                # environment in the preexec function.
+                # if we get here we should have neither a prefix nor a setup command to execute.
+                assert not prefix
+                assert not setup
+
+                # mkosi.sandbox.main() updates os.environ but the environment passed to Popen() is not yet in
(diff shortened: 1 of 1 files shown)


mkosi/run.py
```diff
@@ -244,7 +257,7 @@ def _preexec() -> None:
 
         try:
             proc = subprocess.Popen(
-                [*setup, *prefix, *cmdline],
+                [*setup, *prefix, *cmd],
```
behrmann, line 264: If `cmd` is already `exec`'d in the `preexec_fn`, then won't `setup` and `prefix` no longer be called there?

daandemeyer: `setup` and `prefix` will both be empty if we call `execvp` from the preexec function.

behrmann: Would be good to note that in a comment, simply to remember there are two different paths for `cmd` to be executed here.

PR #4089, opened 2025-12-24
Various fixes

daandemeyer:


mkosi/sandbox.py
```diff
@@ -446,76 +446,89 @@ def join_new_session_keyring() -> None:
...
+        fd = libc.syscall(NR_open_tree, dirfd, path.encode(), flags)
 
     if fd < 0:
-        oserror("open_tree", src)
+        oserror("open_tree", path)
 
-    with close(fd):
-        attr = mount_attr()
-        attr.attr_set = attrs
+    return fd
```
behrmann, line 462: The complaint mypy bubbles up now is because this function returns something, where it didn't before. Looking at the same position before, the revealed type for `fd` is `Any`, too. Just add a cast here.

mkosi/sandbox.py
```diff
@@ -446,76 +446,89 @@ def join_new_session_keyring() -> None:
...
-def mount_rbind(src: str, dst: str, attrs: int = 0) -> None:
-    """
-    When using the old mount syscall to do a recursive bind mount, mount options are not
-    applied recursively. Because we want to do recursive read-only bind mounts in some cases, we
-    use the new mount API for that which does allow recursively changing mount options when doing
-    bind mounts.
-    """
-    flags = AT_NO_AUTOMOUNT | AT_RECURSIVE | AT_SYMLINK_NOFOLLOW | OPEN_TREE_CLONE
-
+def open_tree(dirfd: int, path: str, flags: int = 0) -> int:
```
behrmann, line 449: This mollifies mypy without additional imports
```sugge-stion
def open_tree(dirfd: int, path: str, flags: int = 0) -> int:
    fd: int
```

PR #4091, opened 2025-12-25
Fix reproducibility for microcode initrd and gzip compression

p5:
Some minor improvements to reproducibility, impacting the UKIs and compressed archive outputs (i.e. OCIs).  

I have not confirmed whether this fixes reproducibility for other output formats or other use-cases, but it does in my particular case.

For the compression, I don't believe xz or zstd require changes as their default configurations are reproducible.  gzip is the only one that stores the timestamps in the headers.

With the updated configuration from this MR, plus some additional generalisation required in dnf environments (see below), I was able to produce fully reproducible OCI images built via mkosi.

```
RPMDB="/usr/lib/sysimage/rpm/rpmdb.sqlite"
if [ -f "$RPMDB" ]; then
    sqlite3 "$RPMDB" "PRAGMA journal_mode = DELETE;"
    rm -f /usr/lib/sysimage/rpm/.rpm.lock
fi

rm -rf /usr/lib/sysimage/libdnf5 \
 /var/cache/* \
 /etc/machine-id \
 /etc/group- \
 /etc/gshadow- \
 /etc/passwd- \
 /etc/shadow-
```

---

When building the microcode initrd, files are created in a temporary
directory with current timestamps. These timestamps are then embedded
in the CPIO archive, causing non-reproducible builds even when
SourceDateEpoch is set.

Fix this by normalizing the modification times of all files in the
microcode root directory to source_date_epoch before creating the
CPIO archive.

---

The gzip format includes an MTIME field in its header that stores the
modification time of the original file. This causes compressed archives
to differ between builds even when the uncompressed content is identical.

Add the --no-name flag to gzip which suppresses storing the original
filename and timestamp, making gzip output reproducible.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2349,7 +2349,7 @@ def compressor_command(context: Context, compression: Compression) -> list[PathS
     """Returns a command suitable for compressing archives."""
 
     if compression == Compression.gz:
-        return [gzip_binary(context), f"-{context.config.compress_level}", "--stdout", "-"]
+        return [gzip_binary(context), "--no-name", f"-{context.config.compress_level}", "--stdout", "-"]
     elif compression == Compression.xz:
         return ["xz", "--check=crc32", f"-{context.config.compress_level}", "-T0", "--stdout", "-"]
     elif compression == Compression.zstd:


p5:
CI failures appear to be a network flakes, so hopefully rerunning at another time will resolve these.

This MR doesn't touch anything I would expect to cause regressions at this stage in CI.

```
‣ Downloading Arch Linux keyring
curl: (35) Recv failure: Connection reset by peer
‣ "curl --location --output-dir /work/var/tmp/mkosi-workspace-bq5r66_g/tmp/tmpm2on3hdz --remote-name --no-progress-meter --fail --proxy-capath /proxy.cacert https://archlinux.org/packages/core/any/archlinux-keyring/download" returned non-zero exit code 35.
Error: Process completed with exit code 35.
```

mkosi/__init__.py
```diff
@@ -1500,6 +1500,12 @@ def build_microcode_initrd(context: Context) -> list[Path]:
                     for p in intel.iterdir():
                         f.write(p.read_bytes())
 
+        # Normalize timestamps for reproducible builds before creating cpio
+        if context.config.source_date_epoch is not None:
+            clamp_mtime(root, context.config.source_date_epoch)
+            for p in root.rglob("*"):
+                clamp_mtime(p, context.config.source_date_epoch)
```
daandemeyer: You can replace this with the `normalize_mtime()` function instead.

mkosi/__init__.py
```diff
@@ -2352,7 +2352,7 @@ def compressor_command(context: Context, compression: Compression) -> list[PathS
     """Returns a command suitable for compressing archives."""
 
     if compression == Compression.gz:
-        return [gzip_binary(context), f"-{context.config.compress_level}", "--stdout", "-"]
+        return [gzip_binary(context), "-n", f"-{context.config.compress_level}", "--stdout", "-"]
```
daandemeyer: Please use the long option name instead of the short one.

DaanDeMeyer (changes requested):

mkosi/__init__.py
p5: Thanks!  Replaced with a single call to `normalize_mtime()`.

(No rush getting this merged - spend the holidays with the family :) )

mkosi/__init__.py
```diff
@@ -1500,6 +1500,10 @@ def build_microcode_initrd(context: Context) -> list[Path]:
                     for p in intel.iterdir():
                         f.write(p.read_bytes())
 
+        # Normalize timestamps for reproducible builds before creating cpio
+        if context.config.source_date_epoch is not None:
+            normalize_mtime(root, context.config.source_date_epoch)
```
daandemeyer: No need to gate this behind the if condition, `normalize_mtime()` will do that internally.

DaanDeMeyer (changes requested):

PR #4092, opened 2025-12-25
kmod: Stop retrieving dependency info of all modules

daandemeyer:
Instead of running modinfo once to retrieve the dependency information
of all modules, let's only retrieve the dependency information of the
modules that are to be included in the image and their transitive
dependencies. This means we have to run modinfo multiple times, but with
far fewer modules than before. This ends up being faster than retrieving
the dependency information of all modules, especially after the optimization
from https://github.com/systemd/mkosi/commit/e276dac87a530efac4376a5059b980f2d43460f5.

For the mkosi default image build on Arch Linux this reduces the time for
calculating the required kernel modules and firmware on my laptop from 5s
to 0.5s.

### DIFF
--- a/mkosi/kmod.py
+++ b/mkosi/kmod.py
@@ -27,7 +27,7 @@ def loaded_modules() -> list[str]:
 def all_modules(modulesd: Path) -> Iterator[Path]:
     # The glob may match additional paths.
     # Narrow this down to *.ko, *.ko.gz, *.ko.xz, *.ko.zst.
-    return (m for m in modulesd.rglob("*.ko*") if m.name.endswith((".ko", ".ko.gz", ".ko.xz", ".ko.zst")))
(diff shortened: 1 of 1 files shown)


daandemeyer:
This is my attempt at #4017

mkosi/kmod.py
```diff
@@ -297,30 +255,60 @@ def resolve_module_dependencies(
...
+) -> tuple[set[Path], set[Path]]:
+    """
+    Returns a tuple of lists containing the paths to the module and firmware dependencies of the given list
+    of module names (including the given module paths themselves). The paths are returned relative to the
+    root directory.
+    """
+    modulesd = Path("usr/lib/modules") / kver
+
+    if (p := context.root / modulesd / "modules.builtin").exists():
+        builtin = set(module_path_to_name(Path(m)) for m in p.read_text().splitlines())
```
behrmann: Since we're touching this anyway.
```suggestion
        builtin = {module_path_to_name(Path(m)) for m in p.read_text().splitlines()}
```

PR #4093, opened 2025-12-26
arch: Download archlinux-keyring with pacman

daandemeyer:
curl-ing the Arch Linux website fails quite often due to connection issues. Let's try downloading the archlinux-keyring package with Pacman so we go directly to a mirror and avoid hitting the Arch Linux website.

### DIFF
--- a/mkosi/distribution/arch.py
+++ b/mkosi/distribution/arch.py
@@ -1,9 +1,7 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
 import datetime
-import tempfile
 from collections.abc import Iterable
-from pathlib import Path
 
 from mkosi.archive import extract_tar
 from mkosi.config import Architecture, Config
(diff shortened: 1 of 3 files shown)


mkosi/distribution/arch.py
```diff
@@ -38,22 +36,45 @@ def package_manager(cls, config: "Config") -> type[Pacman]:
...
-                curl(
-                    context.config,
-                    "https://archlinux.org/packages/core/any/archlinux-keyring/download",
-                    output_dir=Path(d),
+            with complete_step(f"Downloading {cls.pretty_name()} keyring"):
+                # We can't override pacman configuration options from the command line so we fall back to
+                # appending to the configuration file instead.
+                config = context.sandbox_tree / "etc/pacman.conf"
+                lines = config.read_text().splitlines()
+                config.write_text("\n".join(lines + ["[options]", "SigLevel = Never"]))
```
behrmann: A bit more wordy, but less splitting and joining of strings and the config snippet is a bit more explicit (and the file has a newline at the end)
```python
                config = context.sandbox_tree / "etc/pacman.conf"
                content = config.read_text()
                with config.open("a") as c:
                    c.write(
                        textwrap.dedent(
                            """
                            [options]
                            SigLevel = Never
                            """
                        )
                    )
```
or more compactly
```python
                config = context.sandbox_tree / "etc/pacman.conf"
                content = config.read_text()
                with config.open("a") as c:
                    c.write("\n[options]\n")
                    c.write("SigLevel = Never\n")
```
The last line would also only be `config.write_text(content)`

behrmann (approved):

PR #4094, opened 2025-12-26
kmod: Only use --modname if available

daandemeyer:
--modname is only available since 2022, and not available in kmod in CentOS Stream 9, so query if it is available and only use it if it is available.

Querying --help every time isn't the most efficient but adds negligible time to the modinfo step in practice.

### DIFF
--- a/mkosi/kmod.py
+++ b/mkosi/kmod.py
@@ -8,13 +8,14 @@ import os
 import re
 import subprocess
 from collections.abc import Iterator, Sequence
+from contextlib import AbstractContextManager
 from pathlib import Path
 
 from mkosi.context import Context
 from mkosi.log import complete_step
 from mkosi.run import chroot_cmd, run
 from mkosi.sandbox import chase
-from mkosi.util import chdir, parents_below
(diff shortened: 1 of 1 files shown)


mkosi/kmod.py
```diff
@@ -211,17 +212,25 @@ class ModuleDependencyInfo:
...
 
     if context.config.output_format.is_extension_image() and not context.config.overlay:
         cmdline += ["--basedir", "/buildroot"]
-        sandbox = context.sandbox(options=context.rootoptions(readonly=True))
+
+        def sandbox() -> AbstractContextManager[list[PathString]]:
+            return context.sandbox(options=context.rootoptions(readonly=True))
     else:
-        sandbox = chroot_cmd(root=context.rootoptions)
+
```
behrmann, line 223: Nit: the empty line here is a bit weird.

behrmann (approved):

mkosi/kmod.py
daandemeyer: ruff wants it

PR #4095, opened 2025-12-27
docs: Update unprivileged user namespace docs

daandemeyer:
Let's not recommend fiddling around with apparmor profiles, but instead recommend enabling unprivileged user namespaces unconditionally. Users that care about security can figure out apparmor profiles on their own.

Also reorder and reword the REQUIREMENTS section in general.


daandemeyer:
Merging so I can link to it from the systemd docs, I'll address any review comments in a follow up PR

PR #4100, opened 2025-12-29
Patchset to enable running `mkosi` on NixOS to hack on systemd

Ma27:
I'm aware that some of those changes are potentially controversial, so I recommend to review this commit-by-commit. For instance, the first commit should be rather uncontroversial, the rest needs more discussion.

----

These patches alongside with a small wrapper (https://git.mbosch.me/ma27/systemd-dev/) are what I built to get a working systemd dev-environment on NixOS. I still use Fedora for the OS image itself and the tools tree. A long-term goal I have is to get NixOS into a state where it can be supported by mkosi natively, but this is the first step to enable NixOS folks (myself included) to hack on the systemd codebase itself.

The issues being fixed were:

* NixOS doesn't have a global `/usr` with some tools inside that are required up-front (e.g. libseccomp or dnf).
* `/bin/bash` not existing[1]
* `--repository_key_fetch` being ignored for the tools-tree.

The wrapper I linked above is mostly a bit of polish to make it easier to work with this setup, but in principle `mkosi` without any additional wrappers is sufficient to hack on systemd from NixOS.

[1] This is also an issue with the systemd tree. Right now i did patch the shebangs inside systemd (e.g. `mkosi.sync`), I may file PRs later if you are OK with such a change here.

### DIFF
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -262,7 +262,7 @@ def spawn(
     if "TMPDIR" in os.environ:
         env["TMPDIR"] = os.environ["TMPDIR"]
 
-    for e in ("SYSTEMD_LOG_LEVEL", "SYSTEMD_LOG_LOCATION"):
+    for e in ("SYSTEMD_LOG_LEVEL", "SYSTEMD_LOG_LOCATION", "LD_LIBRARY_PATH"):
         if e in os.environ:
             env[e] = os.environ[e]
 


daandemeyer:
These all look reasonable, I'll take a closer look later.

mkosi/config.py
```diff
@@ -2743,7 +2743,7 @@ def parse_kernel_module_filter_regexp(p: str) -> str:
         default_factory=config_default_repository_key_fetch,
         parse=config_parse_boolean,
         help="Controls whether distribution GPG keys can be fetched remotely",
-        scope=SettingScope.universal,
+        scope=SettingScope.multiversal,
```
daandemeyer: Whether we use repository key fetching for the tools tree or the target image is not always the same. Imagine if you're using a fedora tools tree on nixos to build a fedora image. You want repository key fetching for the tools tree, but not for the final image since the tools tree will have the necessary keys.

I've opened https://github.com/systemd/mkosi/pull/4101 instead which will make sure we use repository key fetching by default when building an image from nixos.

mkosi/run.py
```diff
@@ -718,7 +732,7 @@ def chroot_options() -> list[PathString]:
         "--unshare-ipc",
         "--setenv", "container", "mkosi",
         "--setenv", "HOME", "/",
-        "--setenv", "PATH", "/usr/bin:/usr/sbin",
+        "--setenv", "PATH", ":".join(mk_path()),
```
daandemeyer: This seems wrong, when chrooted we're chrooted into the image we're building, the nixos path won't matter there. I think this can just be dropped.

mkosi/run.py
```diff
@@ -38,6 +38,16 @@
 T = TypeVar("T")
 
 
+@functools.cache
```
daandemeyer: Don't bother unless you're profiled this slows things down noticeably.

mkosi/run.py
```diff
@@ -38,6 +38,16 @@
...
+def mk_path() -> Sequence[str]:
+    return (
+        "/usr/bin",
+        "/usr/sbin",
+        # NixOS
+        *(el for el in os.environ.get("PATH", "").split(":") if el.startswith("/nix/store")),
+    )
```
daandemeyer: You can inline this into `finalize_path()` since we don't need it in `chroot_options()`.

mkosi/resources/mkosi-tools/mkosi.prepare
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
daandemeyer, line 1: Let's do this thoroughly and use this variant everywhere across mkosi.

daandemeyer:
3rd commit message needs to be updated. 1st commit needs to be dropped

mkosi/config.py
Ma27: I tried out #4101 and the problem is that there are numerous places where mkosi still tries to call `Distribution(d)` for the host distribution which is now nixos, e.g.

```
  File "/mkosi/mkosi/mkosi/run.py", line 53, in uncaught_exception_handler
    yield
  File "/usr/lib64/python3.14/contextlib.py", line 85, in inner
    return func(*args, **kwds)
  File "/mkosi/mkosi/mkosi/__main__.py", line 38, in main
    args, tools, images = parse_config(sys.argv[1:], resources=resources)
                          ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mkosi/mkosi/mkosi/config.py", line 5345, in parse_config
    config = context.finalize()
  File "/mkosi/mkosi/mkosi/config.py", line 5033, in finalize
    ns[s.dest] = copy.deepcopy(self.finalize_value(s))
                               ~~~~~~~~~~~~~~~~~~~^^^
  File "/mkosi/mkosi/mkosi/config.py", line 4804, in finalize_value
    default = setting.default_factory(factoryns)
  File "/mkosi/mkosi/mkosi/config.py", line 1033, in config_default_release
    hd, hr = Distribution(d), r
             ~~~~~~~~~~~~^^^
  File "/usr/lib64/python3.14/enum.py", line 707, in __call__
    return cls.__new__(cls, value)
           ~~~~~~~~~~~^^^^^^^^^^^^
  File "/usr/lib64/python3.14/enum.py", line 1192, in __new__
    raise ve_exc
ValueError: 'nixos' is not a valid Distribution
```

I (re)considered adding a NixOS-type that's empty for now, but this kinda defeats the purpose of this PR, being able to hack on systemd since we'd now have to add `nixos`-definitions to systemd's mkosi configuration that are essentially the same as `Distribution=custom` (and thus not really valid for NixOS).

The other option would be to assume that for `Distribution.custom` the keys also don't exist by default and set `repository_key_fetch` always to True for Distribution=custom.

Wdyt?

daandemeyer: That's just some bugs in https://github.com/systemd/mkosi/pull/4101, I updated it now to address those. Please try again, I don't have a nixos installation at hand to test.

Ma27: Looks good now, thanks!

PR #4103, opened 2025-12-30
Add support for locale-gen

daandemeyer:
If locales are configured in /etc/locale.gen, let's run locale-gen.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3123,6 +3123,20 @@ def run_tmpfiles(context: Context) -> None:
         )  # fmt: skip
 
 
+def run_locale_gen(context: Context) -> None:
+    if not (context.root / "etc/locale.gen").exists():
+        return
+
+    if all(
+        line.strip().startswith("#") or not line.strip()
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -3123,6 +3123,17 @@ def run_tmpfiles(context: Context) -> None:
         )  # fmt: skip
 
 
+def run_locale_gen(context: Context) -> None:
+    if not (context.root / "etc/locale.gen").exists():
+        return
+
+    if all(line.startswith("#") for line in (context.root / "etc/locale.gen").read_text().splitlines()):
```
behrmann: I've only looked at the `locale-gen` in Arch, but that also [skips empty lines](https://gitlab.archlinux.org/archlinux/packaging/packages/glibc/-/blob/main/locale-gen?ref_type=heads#L31)
```suggestion
    if all(line.startswith("#") or not line.strip() for line in (context.root / "etc/locale.gen").read_text().splitlines()):
```

PR #4104, opened 2025-12-30
opensuse: Import GPG keys for all repositories

daandemeyer:
Let's not just import GPG keys for our own repositories, but for all repositories.

### DIFF
--- a/mkosi/distribution/opensuse.py
+++ b/mkosi/distribution/opensuse.py
@@ -6,7 +6,7 @@ from pathlib import Path
 from typing import Union
 from xml.etree import ElementTree
 
-from mkosi.config import Architecture, Config
+from mkosi.config import Architecture, Config, parse_ini
 from mkosi.context import Context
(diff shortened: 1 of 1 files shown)


mkosi/distribution/opensuse.py
```diff
@@ -261,6 +256,24 @@ def is_kernel_package(cls, package: str) -> bool:
...
+                continue
+
+            keys = value.splitlines()
+            for key in keys:
+                if not key.startswith("file://"):
+                    continue
+
+                files.add(Path(key.removeprefix("file://")))
+
+    return list(files)
```
behrmann: sets still don't guarantee order
```suggestion
    return sorted(files)
```
or make `files` a dict and do
```python
files[Path(key.removeprefix("file://"))] = None
```
then `list(files.keys())` will be in insertion order.

behrmann (approved):

PR #4108, opened 2026-01-02
distribution: fix epel repo metalinks

l-austenfeld:
This commit fixes the EPEL repository meta-links.

There where two problems:
- The link is different when a minor version is specified as well (i.e. `10.1` instead of just `10`)
- The names for the testing repos was wrong

The links can be confirmed by looking in the `/etc/yum.repos.d/epel.repo` and `/etc/yum.repos.d/epel-testing.repo` files on a system with the `epel-release` package installed.

The changes can be tested with this `mkosi.conf`:
```
[Distribution]
Distribution=rocky
Release=10.1
RepositoryKeyFetch=yes
Repositories=epel epel-testing
```

### DIFF
--- a/mkosi/distribution/centos.py
+++ b/mkosi/distribution/centos.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
 from collections.abc import Iterable
+import re
 
 from mkosi.config import Architecture, Config
 from mkosi.context import Context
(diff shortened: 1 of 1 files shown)


mkosi/distribution/centos.py
```diff
@@ -308,6 +308,11 @@ def epel_repositories(cls, context: Context) -> Iterable[RpmRepository]:
         else:
             url = "metalink=https://mirrors.fedoraproject.org/metalink?arch=$basearch"
 
+            if GenericVersion(context.config.release) >= 10:
+                repo_extra_part = "-z"
+            else:
+                repo_extra_part = ""
```
behrmann: We can shorten this a bit
```suggestion
            repo_extra_part = "-z" if GenericVersion(context.config.release) >= 10 else ""
```

mkosi/distribution/centos.py
```diff
@@ -308,6 +308,11 @@ def epel_repositories(cls, context: Context) -> Iterable[RpmRepository]:
         else:
             url = "metalink=https://mirrors.fedoraproject.org/metalink?arch=$basearch"
 
+            if GenericVersion(context.config.release) >= 10:
```
daandemeyer: What happens when `11` is released? It'll use the `-z` suffix as well then, which it probably shouldn't?

mkosi/distribution/centos.py
```diff
@@ -316,38 +321,38 @@ def epel_repositories(cls, context: Context) -> Iterable[RpmRepository]:
...
                 gpgurls,
                 enabled=False,
             )
             yield RpmRepository(
                 "epel-testing-source",
-                f"{url}&repo=testing-source-epel{release}",
+                f"{url}&repo=epel{repo_extra_part}-testing-source-{release}",
                 gpgurls,
                 enabled=False,
             )
```
daandemeyer: Have you double checked this also works with centos 9?

mkosi/distribution/centos.py
l-austenfeld: Good point, I forgot about ternary expressions. Will fix that.

mkosi/distribution/centos.py
l-austenfeld: I think there are some version parsing functions already in use in the code somewhere, I will see if I can just use those to check for the presence of a minor version.  
I think this could already be a problem with just version 10 alone, but that never worked in my tests before (it tries to mix Rocky's 10.1 and EPEL's 10.2 repos and fails) so I can't really confirm.

mkosi/distribution/centos.py
l-austenfeld: I did not think to test older versions, that is a good point. Is there and existing test suite I can just run to check this specific thing or should I write my own? Also if I have to write my own (I am assuming a handful of mkosi.conf files is enough) is there a good place to put that, so that future problems can be detected sooner?

daandemeyer: The integration test system is in dire need of a revamp, so a manual test will suffice for now.

l-austenfeld:
The GitHub diff is a bit confusing, so I will try to explain the changes:

I moved all existing metalink entries into an if that checks if the major release is below 10. I removed the two smaller checks for the same, both for epel-next, since they are now redundant.

I then added a separate else branch for release 10 and above that yields the new metalinks. I renamed the repo_extra_part variable to minor_version_flag to better reflect the meaning and used a small regex to detect if a minor version is present.

The minor version regex is the part I am most unsure about.  
I could not find any existing minor version detection code in the project, so I had to write my own. I first thought about adding the functionality to GenericVersion. Seeing the different version formats that can handle, I decided that was above my skill level.  
I decided that checking if the version string starts with a number followed by a period was enough to detect if a minor version is present.

l-austenfeld:
Replaced by #4161

PR #4111, opened 2026-01-06
qemu: Start vsock notify handler thread after starting qemu

daandemeyer:


bluca (approved):
fixes the issue with running without root

PR #4112, opened 2026-01-07
ci: add job to verify simple GHA workflow

bluca:
https://github.com/systemd/mkosi/commit/e276dac87a530efac4376a5059b980f2d43460f5 introduced a regression that broke using the mkosi action in a simple workflow that builds and boots an image.

Add a new job that provides such a smoke test, to avoid these regressions

### DIFF
--- a/.github/workflows/gha.yml
+++ b/.github/workflows/gha.yml
@@ -1,0 +1,44 @@
+name: mkosi
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  boot:
+    runs-on: ${{ matrix.runner }}
+    concurrency:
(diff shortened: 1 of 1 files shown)


behrmann:
This looks like a good addition, but couldn't we add just the smoke test in the integration test and skip building an additional image? In the grand scheme of things it probably doesn't matter, since we're building so many already.

bluca:
I thought about that, but the goal here was to try and replicate a simple enough workflow file one might use in another repo to avoid such regressions, having it mixed with the complex and custom existing one risks masking some issues

daandemeyer:
I would like to figure out why this one triggers the issue and the other ones don't tbh before we merge this.

bluca:
Wasn't it sudo?

bluca:
Can we merge this please? I'd like to avoid further regressions sneaking in - having an extra couple of CI jobs shouldn't hurt, they are fairly quick anyway

PR #4115, opened 2026-01-08
config: Fix InitrdPackages not being added with default

### LINKED ISSUE #4114: InitrdPackages silently ignored when using the default initrd

main

### Used host distribution

openSUSE Leap 16

### Used target distribution

openSUSE Tumbleweed

### Linux kernel version used

6.12.0-160000.8-default

None

The packages listed from `InitrdPackages` are ignored when using the default initrd.

## Explanation

In the function `finalize_default_initrd`, https://github.com/systemd/mkosi/blob/6f4684c7775b584e3c1a742510aa8aa6bdd473a7/mkosi/config.py#L5142-L5145
The `ns` referred to the `context.defaults`. The value for "initrd_packages" is then copied into it with the dest "package":
https://github.com/systemd/mkosi/blob/6f4684c7775b584e3c1a742510aa8aa6bdd473a7/mkosi/config.py#L5147

And just after as we used the default initrd it will parse the config for the `mkosi-initrd` as followed:
https://github.com/systemd/mkosi/blob/6f4684c7775b584e3c1a742510aa8aa6bdd473a7/mkosi/config.py#L5160-L5161

After the parsing we got:
``` 
context.cli      = {}  (empty - not specified on CLI) 
context.config   = {"packages": ["systemd", "udev", "bash", ...]}
context.defaults = {"packages": ["systemd-experimental"]} 
``` 

But later it call `finalize()` that calls `finalize_values()`:
https://github.com/systemd/mkosi/blob/6f4684c7775b584e3c1a742510aa8aa6bdd473a7/mkosi/config.py#L5023
It will only fetch the context.config because it exists, and the context.default is ignored.

## Proposition

In the `finalize_default_initrd` if mkosi-initrd is parsed, can we add the context.default to not loose data ?

### Used mkosi config

```ini

```

### mkosi output

```sh

```

val4oss:
The check `s.dest in main.config` in finalize_default_initrd() is unreliable - initrd_packages is often not present in main.config even when specified in configuration files, causing values to fall through to context.defaults.

When mkosi-initrd/mkosi.conf is parsed, the parser only checks context.config (not context.defaults), so user packages are not seen and cannot be merged. Additionally, during finalization, context.config takes priority over context.defaults, causing user packages to be lost.

Fix by always using context.config for non-CLI initrd settings, ensuring they are visible to the parser and get properly merged with the mkosi-initrd default packages via the list parser.

Fixes #4114

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -5149,10 +5149,10 @@ def finalize_default_initrd(
                 ns = context.cli
                 if f"{s.dest}_was_none" in main.cli:
                     ns[f"{dest}_was_none"] = main.cli[f"{s.dest}_was_none"]
-            elif s.dest in main.config:
-                ns = context.config
             else:
-                ns = context.defaults
+                # Always use context.config for non-CLI initrd settings to ensure they
+                # merge with mkosi-initrd config during parsing
+                ns = context.config
 
             ns[dest] = copy.deepcopy(finalized[s.dest])
 


daandemeyer:
Not quite correct, my pr has the correct fix

val4oss:
Thank you for the correct fix.

PR #4116, opened 2026-01-08
Make sure we pass the right context to finalize_default_initrd()

### LINKED ISSUE #4114: InitrdPackages silently ignored when using the default initrd

daandemeyer:
We mess around with the context object to make it for for use when reading the subimages. But we need the full context again for parsing the default initrd later on, so make a copy before we delete stuff and pass that to finalize_default_initrd()

Fixes #4114

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1652,3 +1652,34 @@ def test_assert(tmp_path: Path) -> None:
 
         parse_config(["--image-id", "abcde", "--environment", "ABC=QED"])
         parse_config(["--image-id", "abcde", "--environment", "DEF=QEE"])
+
+
+def test_initrd_packages(tmp_path: Path) -> None:
+    d = tmp_path
+
+    (d / "mkosi.conf").write_text(
+        """\
+        [Content]
+        InitrdPackages=package1
+        InitrdPackages=package2
+
+        [Content]
+        Bootable=yes
+        """
+    )
+
+    with chdir(d), resource_path(mkosi.resources) as resources:
+        _, _, [initrd, _] = parse_config(resources=resources)
+
+    assert "package1" in initrd.packages
+    assert "package2" in initrd.packages
+
+    # Make sure the InitrdPackages= are also picked up when a subimage is defined.
+    (d / "mkosi.images").mkdir()
+    (d / "mkosi.images/subimage.conf").touch()
+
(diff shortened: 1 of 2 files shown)


tests/test_config.py
```diff
@@ -1652,3 +1652,24 @@ def test_assert(tmp_path: Path) -> None:
 
         parse_config(["--image-id", "abcde", "--environment", "ABC=QED"])
         parse_config(["--image-id", "abcde", "--environment", "DEF=QEE"])
+
+
+def test_initrd_packages(tmp_path: Path) -> None:
```
behrmann, line 1657: The test doesn't seem to test what it's supposed to test, i.e. when I remove the above changes, the test still passes.

daandemeyer: oops fixed

PR #4119, opened 2026-01-13
Various small fixes

elmarco:


README.md
```diff
@@ -33,7 +33,7 @@ might have to add `~/.local/bin` to your user's `PATH`.
 
 ```shell
 git clone https://github.com/systemd/mkosi
-ln -s $PWD/mkosi/bin/mkosi ~/.local/bin/mkosi
+ln -s $PWD/bin/mkosi ~/.local/bin/mkosi
```
behrmann, line 36: That's not correct, since the `git clone` above will not chdir into the just created `mkosi` directory.

mkosi/tree.py
```diff
@@ -98,6 +98,12 @@ def tree_has_selinux_xattr(path: Path) -> bool:
     )
 
 
+def tree_has_ima_xattr(path: Path) -> bool:
+    return any(
+        "security.ima" in os.listxattr(p, follow_symlinks=False) for p in (path, *path.rglob("*"))
+    )
```
behrmann, line 104: ruff complains

```suggestion
    return any("security.ima" in os.listxattr(p, follow_symlinks=False) for p in (path, *path.rglob("*")))
```

mkosi/tree.py
```diff
@@ -118,9 +124,16 @@ def copy_tree(
...
     if preserve:
         attrs += ",timestamps,ownership"
+        with_xattrs = True
 
         # Trying to copy selinux xattrs to overlayfs fails with "Operation not supported" in containers.
         if statfs(os.fspath(dst.parent)) != OVERLAYFS_SUPER_MAGIC or not tree_has_selinux_xattr(src):
+            with_xattrs = False
+
+        if tree_has_ima_xattr(src):
+            with_xattrs = False
```
behrmann, line 134: This seems problematic, since even if we decided to e.g. copy xattr to keep selinux attributes around, we'd drop them if a file also has these xattrs. 

Despite being a bit annoying is there any issue with the warning from cp?

mkosi/resources/mkosi-tools/mkosi.conf.d/fedora/mkosi.conf
```diff
@@ -12,3 +12,4 @@ Packages=
         erofs-utils
         pkcs11-provider
         python3-pefile
+        systemd-boot-unsigned
```
daandemeyer, line 15: This is wrong, systemd-boot has to be installed inside the image, not in the tools tree.

README.md
elmarco: oh that's right, my bad! It's a bit uncommon though.

mkosi/tree.py
elmarco: c&p from tree_has_selinux_xattr. We may want to generalize the function instead.

mkosi/tree.py
elmarco: cp fails with exit 1

mkosi/resources/mkosi-tools/mkosi.conf.d/fedora/mkosi.conf
elmarco: ah, where should it be ? mkosi.conf.d/fedora/mkosi.conf ?

mkosi/tree.py
behrmann: Ruff complains here because `security.ima` is shorter than `security.selinux` and therefore this fits on a single line. Either way, this needs to be addressed.

mkosi/tree.py
behrmann: Ok, but since this is the same as `security.selinux` then, shouldn't it be be
```python
statfs(os.fspath(dst.parent)) != OVERLAYFS_SUPER_MAGIC 
or (not tree_has_selinux_xattr(src) and not tree_has_ima_xattr(src))
```
then?

mkosi/resources/mkosi-tools/mkosi.conf.d/fedora/mkosi.conf
daandemeyer: @elmarco Nowhere? You have to install it yourself in your image definition

PR #4121, opened 2026-01-13
opensuse: Fetch remote keys as well if RepositoryKeyFetch= is enabled

daandemeyer:
Also improve error handling for if local keys do not exist.

### DIFF
--- a/mkosi/distribution/opensuse.py
+++ b/mkosi/distribution/opensuse.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
+import os
 import tempfile
 from collections.abc import Iterable
 from pathlib import Path
@@ -15,7 +16,8 @@ from mkosi.installer.rpm import RpmRepository, find_rpm_gpgkey, setup_rpm
 from mkosi.installer.zypper import Zypper
 from mkosi.log import die
 from mkosi.mounts import finalize_certificate_mounts
-from mkosi.run import run
+from mkosi.run import run, workdir
+from mkosi.util import flatten
 from mkosi.versioncomp import GenericVersion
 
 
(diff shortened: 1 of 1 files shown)


mkosi/distribution/opensuse.py
```diff
@@ -266,10 +269,21 @@ def fetch_gpgkeys(context: Context) -> list[Path]:
...
-                    continue
-
-                files.add(Path(key.removeprefix("file://")))
+                if key.startswith("file://"):
+                    path = key.removeprefix("file://")
+                    if not (context.config.tools() / path.lstrip("/")).exists():
+                        die(f"Local repository GPG key specified ({key}) but not found at {path}")
+
+                    files.add(context.config.tools() / path.lstrip("/"))
+                elif context.config.repository_key_fetch:
```
behrmann: So since we're passing this through to curl and curl is such a swiss army knife, it would be nice to ensure that `key` starts with `https://` here.

mkosi/distribution/opensuse.py
```diff
@@ -266,10 +269,21 @@ def fetch_gpgkeys(context: Context) -> list[Path]:
...
-                if not key.startswith("file://"):
-                    continue
-
-                files.add(Path(key.removeprefix("file://")))
+                if key.startswith("file://"):
+                    path = key.removeprefix("file://")
+                    if not (context.config.tools() / path.lstrip("/")).exists():
+                        die(f"Local repository GPG key specified ({key}) but not found at {path}")
+
+                    files.add(context.config.tools() / path.lstrip("/"))
```
behrmann: Minimally less cluttered
```suggestion
                    path = key.removeprefix("file://").lstrip("/")
                    if not (context.config.tools() / path).exists():
                        die(f"Local repository GPG key specified ({key}) but not found at /{path}")

                    files.add(context.config.tools() / path)
```

PR #4126, opened 2026-01-18
Allow specifying "default" value for Initrds=

daandemeyer:
Also use it as the default value for Initrds= when we decide a default initrd is needed. This allows both using the default initrd alongside other initrds as well as disabling building the default initrd by assigning the empty string to Initrds=.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -1116,6 +1116,31 @@ def config_default_proxy_peer_certificate(namespace: dict[str, Any]) -> Optional
     return None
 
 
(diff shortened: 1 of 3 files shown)


mkosi/config.py
```diff
@@ -5225,22 +5265,6 @@ def want_kernel(config: Config) -> bool:
     return True
 
 
-def want_default_initrd(config: Config) -> bool:
```
behrmann, line 5228: I do like that this function is quite descriptive with its name. You could keep this function around and do 
```python
    return Path("default") in config.initrds
```
This would also make the diff below smaller.

PR #4130, opened 2026-01-20
dnf: Give advanced users some control over plugins

daandemeyer:
Let's add some environment variables to control plugins for cases where users have some dnf plugin they can't touch on their host system which doesn't behave properly in mkosi-sandbox and which they can't remove themselves.

### DIFF
--- a/mkosi/installer/dnf.py
+++ b/mkosi/installer/dnf.py
@@ -169,6 +169,11 @@ class Dnf(PackageManager):
             "--setopt=persistdir=/buildroot/var/lib/dnf",
         ]
 
+        if plugins := context.config.environment.get("MKOSI_DNF_DISABLE_PLUGINS"):
+            cmdline += [f"--disableplugin={plugins}"]
+        if plugins := context.config.environment.get("MKOSI_DNF_ENABLE_PLUGINS"):
+            cmdline += [f"--enableplugin={plugins}"]
+
         if ARG_DEBUG.get():
             cmdline += ["--setopt=debuglevel=10"]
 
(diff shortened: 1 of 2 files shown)


mkosi/installer/dnf.py
```diff
@@ -169,6 +169,11 @@ def cmd(
             "--setopt=persistdir=/buildroot/var/lib/dnf",
         ]
 
+        if (plugins := context.config.environment.get("DNF_ENABLE_PLUGINS")):
+            cmdline += [f"--enableplugin={plugins}"]
+        if (plugins := context.config.environment.get("DNF_DISABLE_PLUGINS")):
+            cmdline += [f"--disableplugin={plugins}"]
```
davide125: Convention is usually to do disable before enable (as that allows one to do `--disableplugin=* --enableplugin=whatever`).

mkosi/installer/dnf.py
```diff
@@ -169,6 +169,11 @@ def cmd(
             "--setopt=persistdir=/buildroot/var/lib/dnf",
         ]
 
+        if (plugins := context.config.environment.get("DNF_DISABLE_PLUGINS")):
+            cmdline += [f"--disableplugin={plugins}"]
+        if (plugins := context.config.environment.get("DNF_ENABLE_PLUGINS")):
```
behrmann: 'tis what ruff's complaining about.
```suggestion
        if plugins := context.config.environment.get("DNF_DISABLE_PLUGINS"):
            cmdline += [f"--disableplugin={plugins}"]
        if plugins := context.config.environment.get("DNF_ENABLE_PLUGINS"):
```

behrmann (approved):
These should get a note in the `ENVIRONMENT VARIABLES` section otherwise they are quite hard to find out about.

PR #4133, opened 2026-01-23
Support resizing output image

pothos:
When the final image does not contain the rootfs it must be grown before being able to boot it. Since the exact size to grow is not easily known to the user, it's best if the image is already grown to the right size. Also, when flashing to disk it helps to have the image be the minimum size needed because otherwise flashing on a too small disk succeeds but boot will fail. With the right image size it's easier for the user to know whether the disk is large enough.


mkosi/__init__.py
```diff
@@ -4128,6 +4128,13 @@ def build_image(context: Context) -> None:
     elif context.config.output_format == OutputFormat.directory:
         context.root.rename(context.staging / context.config.output_with_format)
 
+    if context.config.output_size and context.config.output_format == OutputFormat.disk:
+        output = context.staging / context.config.output_with_format
+        current_size = output.stat().st_size
+        if context.config.output_size > current_size:
```
behrmann, line 4135: Looking at the implementation, it is clear, that this can only grow the image, and the documentation does mention it, but logging warning in the case we cannot grow the image, would be a nicer UX.

pothos: Thanks, like that with `log_notice`?

mkosi/__init__.py
```diff
@@ -4129,6 +4128,15 @@ def build_image(context: Context) -> None:
...
+            with complete_step(f"Resizing output image to {format_bytes(context.config.output_size)}…"):
+                os.truncate(output, context.config.output_size)
+        else:
+            log_notice(f"Can't change output size because image is already larger ({format_bytes(current_size)}) than requested")
```
behrmann: Nit: 
```suggestion
            log_notice(f"Can't change the output size because the image is already larger ({format_bytes(current_size)}) than requested")
```

pothos: Thanks

pothos: Ok, finally fixed the ruff formatting

PR #4136, opened 2026-01-24
ssh: user port forwarding if Vsock is not enabled

gourryinverse:
Some sandboxes don't allow Vsock access but still allow local host port forwarding.

If `VSock=no` and `Ssh=yes|always`, add a manual port forward to 22->2222 to the host (fairly standard port choice).

### DIFF
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -1083,7 +1083,11 @@ def run_qemu(args: Args, config: Config) -> None:
     ]  # fmt: skip
 
     if config.runtime_network == Network.user:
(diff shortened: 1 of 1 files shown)


gourryinverse:
the failing CI test does not look related to the changes in this patch

daandemeyer:
Can you explain your environment more? With https://stefano-garzarella.github.io/posts/2026-02-11-vsock-netns/ we're getting vsock namespacing very soon so vsock will be allowed sooner or later, hence I'm a bit reluctant to merge this.

gourryinverse:
this may not actually be a requirement, we're automating some testing with claude it was having issues connecting without an explicit port.  I think we can kill this one.

PR #4142, opened 2026-01-28
Add support for building UKIProfiles initrds with extra packages

bluca:
If Packages= is specifed under a UKI profile, take the 'main' initrd and add the new packages on top of it, and use that as the profile's initrd. Profile's sections are complete replacements, they are not additive, so it needs to be the full thing.

### DIFF
--- a/mkosi/resources/man/mkosi.1.md
+++ b/mkosi/resources/man/mkosi.1.md
@@ -2266,6 +2266,13 @@ settings can be specified in the `UKIProfile` section:
     `.cmdline` section and the extra kernel command line arguments
     specified with this setting.
 
+`Packages=`
+:   Install the specified distribution packages (i.e. RPM, deb, …) in the
+    initrd, on top of the base profile's initrd. Takes a comma-separated list
+    of package specifications. This option may be used multiple times in which
+    case the specified package lists are combined. The types and syntax is the
+    same as the `Content` section's `Packages=` setting.
+
 `SignExpectedPcr=`
 :   Sign expected PCR measurements for this UKI profile. Takes a boolean.
     Enabled by default.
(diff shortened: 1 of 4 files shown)


mkosi/resources/man/mkosi.1.md
```diff
@@ -2266,6 +2266,13 @@ settings can be specified in the `UKIProfile` section:
     `.cmdline` section and the extra kernel command line arguments
     specified with this setting.
 
+`Packages=`
```
behrmann, line 2269: So while this does end in it's own structure in the config, I think calling this `Packages=` has the (only very slight!) potential for hard to debug errors, since during config parsing we do pass around dictionaries. 

More importantly, though, we already have `InitrdPackages=` which would make this a bit confusing, since that's for the packages in the initrd. 

How about `ExtraInitrdPackages=` to clearly communicate that this is different and that this is additive?

bluca: I'd rather leave it as-is, given it's a different section and it has the same syntax as the other it seems clearer. More importantly OBS won't parse a new setting for dependency resolution, so that would be a massive pain to change

daandemeyer:
I doubt we should do this until we have a mechanism in systemd to make these overlays on top of the main initrd given the size increases you get otherwise, closing for now, feel free to reopen if you disagree.

bluca:
Lennart doesn't want to have the initrd profile overlay thingy so that won't happen. Given it's optional and one has to actually configure it, it should be fine to just add it? There's no cost to anybody not configuring packages in the profile

daandemeyer:
I'd prefer to not have to maintain code that nobody will end up using due to various limitations

PR #4143, opened 2026-01-28
ci: add nodejs-npm to tools tree packages for Fedora

behrmann:


mkosi.tools.conf/mkosi.prepare.chroot
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
-if ! command -v pyright &>/dev/null; then
-    npm install -g pyright
+if command -v pyright >/dev/null; then
```
bluca: Is the `!` removal intentional? Now it will only install it if it's already available?

bluca: fixed as requested over IM

PR #4144, opened 2026-01-28
Add MakeScriptsExecutable= setting to optionally try to make scripts executable before bailing out

bluca:
If it fails, it was going to die() anyway.

OBS sources defined inline (ie, not in a tarball) cannot have the mode preserved, so it's not possible to have mkosi.build or so as a bare script in an OBS project, one needs to tar it up and extract it again later, which means it cannot be edited by the inline editor, which is very convenient for small and trivial builds like an addon.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -649,13 +649,21 @@ def finalize_config_json(config: Config) -> Iterator[Path]:
         yield Path(f.name)
 
 
+def check_script(config: Config, script: Path) -> None:
+    if not os.access(script, os.X_OK):
+        if config.make_scripts_executable:
+            logging.warning(f"{script} is not executable, attempting to chmod it")
+            os.chmod(script, os.stat(script).st_mode | stat.S_IXUSR)
+        else:
+            die(f"{script} is not executable")
+
+
 def run_configure_scripts(config: Config) -> Config:
     if not config.configure_scripts:
         return config
 
     for script in config.configure_scripts:
-        if not os.access(script, os.X_OK):
-            die(f"{script} is not executable")
+        check_script(config, script)
 
     env = dict(
         DISTRIBUTION=str(config.distribution),
@@ -703,8 +711,7 @@ def run_sync_scripts(config: Config) -> None:
         return
 
     for script in config.sync_scripts:
-        if not os.access(script, os.X_OK):
-            die(f"{script} is not executable")
+        check_script(config, script)
 
     env = dict(
         DISTRIBUTION=str(config.distribution),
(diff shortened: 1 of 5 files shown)


behrmann:
I feel a bit uneasy about, since I'm not a fan of running stuff that I didn't make executable—even if I just forgot. I think what irks me about this, is that it makes changes to my config somewhat magically.

Would it work for you to make this opt in with an option?

bluca:
Sure

mkosi/__init__.py
```diff
@@ -655,7 +655,11 @@ def run_configure_scripts(config: Config) -> Config:
 
     for script in config.configure_scripts:
         if not os.access(script, os.X_OK):
-            die(f"{script} is not executable")
+            if config.make_hooks_executable:
+                logging.warning(f"{script} is not executable, attempting to chmod it")
+                os.chmod(script, os.stat(script).st_mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
+            else:
+                die(f"{script} is not executable")
```
daandemeyer: Please abstract this into a function and reuse it everywhere

daandemeyer (changes requested):

mkosi/resources/man/mkosi.1.md
```diff
@@ -1753,6 +1753,10 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     Currently, setting a proxy client key is only supported when **dnf** or
     **dnf5** is used to build the image.
 
+`MakeHooksExecutable=`, `--make-hooks-executable=`
```
daandemeyer: Actually, we don't use the hook terminology anywhere, can we just name this `MakeScriptsExecutable=` and use the script terminology everywhere instead of hook?

bluca: done

mkosi/__init__.py
```diff
@@ -649,13 +649,21 @@ def finalize_config_json(config: Config) -> Iterator[Path]:
         yield Path(f.name)
 
 
+def check_script(config: Config, script: Path) -> None:
+    if not os.access(script, os.X_OK):
+        if config.make_scripts_executable:
+            logging.warning(f"{script} is not executable, attempting to chmod it")
+            os.chmod(script, os.stat(script).st_mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
```
daandemeyer: S_IXUSR should be sufficient? Making these group and world executable seems wrong.

PR #4151, opened 2026-02-05
action: Run apt-get update before installing packages

daandemeyer:
The repository metadata in the image can get out of date. Let's run apt-get update to make sure it is fresh.


behrmann:
CI failures are because it's that time of the year again and therefore unrelated.

PR #4152, opened 2026-02-05
fedora: Use N-1 key as well when querying rawhide GPG key

daandemeyer:
We now run into the opposite problem where distribution-gpg-keys is updated before rawhide is updated, so query the N-1 key as well to avoid issues in the other direction.


keszybz (approved):
LGTM.

PR #4157, opened 2026-02-11
oci: allow user defined labels and annotations

bketelsen:
Adds `OciLabels=` and `OciAnnotations=` to `[Output]` section of mkosi config.

OciLabels map to `.config.Labels` in the OCI config blob (Equiv to LABEL in Dockerfile)
OciAnnotations maps to `.annotations` in the OCI manifest.

Auto-set `org.opencontainers.image.ref.name` to mkosi ImageID as a convenience feature.

Example:

```
[Output]
OciLabels=containers.bootc=1
OciAnnotations=org.opencontainers.image.vendor=frostyard
```

If useful, can follow up PR with cli flags for --oci-label and --oci-annotation.


mkosi/config.py
```diff
@@ -2900,6 +2902,22 @@ def parse_kernel_module_filter_regexp(p: str) -> str:
         help="Set ID for image",
         scope=SettingScope.inherit,
     ),
+    ConfigSetting(
+        dest="oci_labels",
+        metavar="KEY=VALUE",
+        section="Output",
+        parse=config_make_dict_parser(delimiter=" ", parse=parse_key_value, unescape=True),
+        match=config_match_key_value,
```
daandemeyer: I don't think we need matches for these immediately unless you have a clear use case for that

mkosi/config.py
```diff
@@ -2900,6 +2902,22 @@ def parse_kernel_module_filter_regexp(p: str) -> str:
...
+        help="Set OCI config labels (visible in podman/docker inspect)",
+    ),
+    ConfigSetting(
+        dest="oci_annotations",
+        metavar="KEY=VALUE",
+        section="Output",
+        parse=config_make_dict_parser(delimiter=" ", parse=parse_key_value, unescape=True),
+        match=config_match_key_value,
```
daandemeyer: Ditto

mkosi/__init__.py
```diff
@@ -3678,6 +3678,7 @@ def make_oci(context: Context, root_layer: Path, dst: Path) -> None:
                 "/sbin/init",
                 *context.config.kernel_command_line,
             ],
+            **({"Labels": dict(context.config.oci_labels)} if context.config.oci_labels else {}),
```
behrmann, line 3681: The `dict` call here is unnecessary, since `context.config.oci_labels` already is a dictionary.

mkosi/__init__.py
```diff
@@ -3733,6 +3735,15 @@ def make_oci(context: Context, root_layer: Path, dst: Path) -> None:
...
                             "size": (ca_store / oci_manifest_digest).stat().st_size,
+                            **(
+                                {
+                                    "annotations": {
+                                        "org.opencontainers.image.ref.name": context.config.image_id,
+                                    },
+                                }
+                                if context.config.image_id
+                                else {}
+                            ),
```
behrmann, line 3746: Sorry if this question is silly, I'm not an OCI expert by any measure, but looking at the spec both the `annotations` for the image manifest spec and the image index spec seem to use the same pre-defined keys. If this is correct, is there a reason to set `org.opencontainers.image.ref.name` here, but not above, and vice versa for `org.opencontainers.image.version`? Also, would it make more sense to name this `oci_manifest_annotations` instead, since it might be necessary to have other annotations for the `index.json`?

PR #4158, opened 2026-02-11
Bump minimum python version to 3.10

daandemeyer:
- CentOS Stream 9 has python 3.9 by default, but python 3.12 is packaged
- Ubuntu Jammy has python 3.10.

So we'll require CentOS Stream 9 users to install the python3.12
package to keep using mkosi, which shouldn't be a problem.

Bumping version allows us to switch to the Union operator among other
improvements. This commit gets rid of Union and Optional, we'll adopt
other 3.10 features later.

### DIFF
--- a/mkosi/sandbox.py
+++ b/mkosi/sandbox.py
@@ -132,7 +132,6 @@ libc = ctypes.CDLL(None, use_errno=True)
 libc.syscall.restype = ctypes.c_long
 libc.unshare.argtypes = (ctypes.c_int,)
 libc.statfs.argtypes = (ctypes.c_char_p, ctypes.c_void_p)
-libc.eventfd.argtypes = (ctypes.c_int, ctypes.c_int)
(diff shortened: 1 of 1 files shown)


bin/mkosi
```diff
@@ -10,10 +10,24 @@ if [ -z "$MKOSI_INTERPRETER" ]; then
...
+        candidate="$(
+            IFS=:
+            for dir in $PATH; do
+                for bin in "$dir"/python3*; do
+                    [ -x "$bin" ] && basename "$bin"
+                done
+            done | sort --unique --version-sort --reverse | head --lines=1
+        )"
+
+        if [ -n "$candidate" ] && "$candidate" -c "import sys; sys.exit(sys.version_info < (3, 9))"; then
```
behrmann: Since this is a static string, single quotes are better, also this should check for being smaller than `(3, 10)`, since this is true until 3.9 and will exit non-zero because of that and otherwise it would already exit with zero for 3.9
```suggestion
        if [ -n "$candidate" ] && "$candidate" -c 'import sys; sys.exit(sys.version_info < (3, 10))'; then
```

PR #4159, opened 2026-02-12
Support glob expansion for path list config options

### LINKED ISSUE #4155: Support for custom mkosi.*.d / mkosi.images folders?
Hi! Thanks so much for your work on mkosi! At Flashbots, we're using it to create reproducible Linux images for use in Trusted Execution Environments like Intel TDX and AMD SEV. It's made the image creation process 100x easier for everyone involved and drastically sped up our development feedback loop.

We've built up a pretty big monorepo with a large hierarchy of shared modules and configuration. The way we manage this is by having multiple top level `mkosi.conf` entrypoints in different folders of our repo, so different teams within the company can use the same repo to build their products without having to port shared infrastructure back and forth.

This works really well for our use case, but some features in mkosi have a minor issue that prevents them from working with this approach. Mkosi enforces that directories like `mkosi.sync.d` and `mkosi.postinst.d` as well as `mkosi.images` need to be in the root of the repo, and there's no way to specify multiple of them. Because of this, we can't have per-module mkosi.postinst.d directories or mkosi.images directories.

Would it be possible to add some options like `PostinstDirectories=` for this functionality?

Thanks!

alexhulbert:
Adds support for configuration like `PostInstallationScripts=my-module/mkosi.postinst.d/*`

Resolves #4155

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -8,6 +8,7 @@ import enum
 import fnmatch
 import functools
 import getpass
+import glob
 import graphlib
 import io
 import itertools
@@ -1204,6 +1205,7 @@ def config_make_list_parser(
     unescape: bool = False,
     reset: bool = True,
     key: Optional[Callable[[T], Any]] = None,
+    expand_globs: bool = False,
 ) -> ConfigParseCallback[list[T]]:
     def config_parse_list(value: Optional[str], old: Optional[list[T]]) -> Optional[list[T]]:
         new = old.copy() if old else []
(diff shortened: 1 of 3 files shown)


daandemeyer:
@alexhulbert Actually please update the docs and add some tests to test_config.py

alexhulbert:
All set!

behrmann:
Looks reasonable to me and the sorting behaviour with globs is properly documented. Could you rebase and squash this? We don't usually do back merges from main into feature branches and the docs and tests belong to the feature.

alexhulbert:
No problem :+1: Squashed & rebased

PR #4160, opened 2026-02-12
Add support for using the foreign UID range for directory images

daandemeyer:
systemd recently introduced the foreign UID range to enable unprivileged
nspawn containers using directory rootfs trees. Let's get rid of all our
hacks to get directory images booting in nspawn and qemu and instead insist
on the foreign UID range when booting directory images.

Because the foreign UID range is rather new, we don't insist on using it
to build images. The new ForeignUIDRange= option has to be enabled to make
use of the foreign UID range. When enabled, we'll enter a user namespace
with the foreign UID range mapped into it when building an image and use
systemd-dissect at the end of the image build to recursively chown the
directory image to the foreign UID range.

run_clean() now needs access to the foreign UID range to remove directory
trees owned by the foreign UID range so we run it with fork_and_wait() as
well now.

To get access to the foreign UID range, we use a few extensions to
AllocateUserRange() from systemd-nsresourced which means we add a minimal
varlink implementation to sandbox.py.

To boot directory images with systemd-nspawn and systemd-vmspawn, we have
to delegate a transient UID range to the mkosi sandbox so that it can be
allocated to systemd-nspawn as a transient UID range by systemd-nsresourced.

For mkosi qemu, we have to set things up for virtiofsd ourselves which we
do with the new --bind-foreign option for mkosi-sandbox in which case it does
the same setup that vmspawn does for virtiofsd itself.

All the foreign UID range functionality will require systemd v260 or newer,
specifically the following PRs:
- https://github.com/systemd/systemd/pull/40212
- https://github.com/systemd/systemd/pull/40564
- https://github.com/systemd/systemd/pull/40415

### DIFF
--- a/mkosi/resources/man/mkosi-sandbox.1.md
+++ b/mkosi/resources/man/mkosi-sandbox.1.md
@@ -66,6 +66,17 @@ from `mkosi.sandbox` is not supported and may break in future versions.
 `--ro-bind-try SRC DST`
 :   Like `--bind-try`, but does a recursive readonly bind mount.
 
+`--bind-foreign SRC DST`
+:   Like `--bind`, but `mkosi-sandbox` will allocate a user namespace and provision it with a transient UID
+    range using `systemd-nsresourced` and then idmap the foreign UID range to this transient UID range using
+    `systemd-mountfsd` for each `--bind-foreign` mount. Finally, `mkosi-sandbox` will join the user namespace
+    just before executing the configure command in the sandbox.
+
+    Using this option requires `systemd-nsresourced` and `systemd-mountfsd` v260 or newer.
+
+`--ro-bind-foreign SRC DST`
+:   Like `--bind-foreign`, but does a readonly mount.
+
 `--symlink SRC DST`
 :   Creates a symlink at `DST` in the sandbox pointing to `SRC`. If `DST` already
     exists and is a file or symlink, a temporary symlink is created and mounted on
@@ -112,6 +123,21 @@ from `mkosi.sandbox` is not supported and may break in future versions.
     sandbox will be able to do bind mounts and other operations.
 
     If `mkosi-sandbox` is invoked as the root user, this option won't do anything.
+
+`--map-foreign`
+:   Map the foreign UID range in the sandbox. For more information on the foreign UID range,
+    see [Users, Groups, UIDs and GIDs on systemd Systems](https://systemd.io/UIDS-GIDS/).
+
+    When running `mkosi-sandbox` unprivileged, this option requires `systemd-nsresourced`
+    v260 or newer.
+
+`--map-delegate N`
+:   Map N additional 64K UID/GID ranges in the sandbox.
+
+    This is needed to execute `systemd-nspawn` or any program that tries to allocate a
+    transient UID range via `systemd-nsresourced` from within the `mkosi-sandbox` environment.
+
+    This option requires `systemd-nsresourced` v260 or newer.
 
 `--suppress-chown`
 :   Specifying this option causes all calls to `chown()` or similar system calls to become a
--- a/mkosi/resources/man/mkosi.1.md
+++ b/mkosi/resources/man/mkosi.1.md
@@ -92,12 +92,21 @@ The following command line verbs are known:
     shell in the container. To pass extra options to nspawn, separate them
     from regular options with `--`.
 
+    To use `directory` images with `shell`, the `ForeignUIDRange=` option has to
+    be enabled when building the image.
+
 `boot`
 :   Similar to `shell`, but instead of spawning a shell, it boots systemd in the
     image using **systemd-nspawn**. Extra arguments may be specified after
     the `boot` verb, which are passed as the *kernel command line* to the
     init system in the image. To pass extra options to nspawn, separate them
     from regular options with `--`.
+
+    To use `directory` images with `boot`, the `ForeignUIDRange=` option has to
+    be enabled when building the image.
+
+    To use `disk` images with `boot`, systemd-nsresourced and systemd-mountfsd
+    v260 or newer are required on the host.
 
 `vm`
 :   Similar to `boot`, but uses the configured virtual machine monitor (by
@@ -107,6 +116,9 @@ The following command line verbs are known:
     virtual machine monitor. See `VirtualMachineMonitor=` for more
     information. To pass extra options to the configured virtual machine
     monitor, separate them from regular options with `--`.
+
+    To use `directory` images with `vm`, the `ForeignUIDRange=` option has to
+    be enabled when building the image.
 
 `ssh`
 :   When the image is built with the `Ssh=always` option or if systemd's
@@ -1797,6 +1809,16 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 `MakeScriptsExecutable=`, `--make-scripts-executable=`
 :   If one of the hook scripts (see `SCRIPTS` section) is not marked as executable, attempt to chmod it
     instead of failing outright. Defaults to `no`.
+
+`ForeignUIDRange=`, `--foreign-uid-range=`
+:   If enabled, directory images will be owned by the foreign UID range on disk
+    instead of the user's UID.
+
+    Enabling this option is required to use directory images with `mkosi shell`,
+    `mkosi boot` and `mkosi vm`.
+
+    systemd-nsresourced and systemd-mountfsd v260 or newer are required on the host
+    to make use of this option.
 
 ### [Runtime] Section (previously known as the [Host] section)
 
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -261,12 +261,8 @@ def _preexec(
     cmd: list[str],
     env: dict[str, Any],
     sandbox: list[str],
-    preexec: Optional[Callable[[], None]],
 ) -> None:
     with uncaught_exception_handler(exit=os._exit, fork=True, proceed=True):
-        if preexec:
-            preexec()
-
         if not sandbox:
             return
 
@@ -306,7 +302,6 @@ def spawn(
     pass_fds: Collection[int] = (),
     env: Mapping[str, str] = {},
     log: bool = True,
-    preexec: Optional[Callable[[], None]] = None,
     success_exit_status: Sequence[int] = (0,),
     setup: Sequence[PathString] = (),
     sandbox: AbstractContextManager[Sequence[PathString]] = nosandbox(),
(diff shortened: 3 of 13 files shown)


mkosi/sandbox.py
```diff
@@ -587,7 +591,152 @@ def __exit__(self, *args: object, **kwargs: object) -> None:
...
 
 
-def become_user(uid: int, gid: int) -> None:
+if sys.version_info >= (3, 10):
+    Json = dict[str, "Json"] | list["Json"] | tuple["Json"] | str | int | float | bool | None
+    FD = int | tuple[str, int]
+else:
+    from typing import Union
+
+    Json = Union[dict[str, "Json"], list["Json"], tuple["Json"], str, int, float, bool, None]
```
behrmann: Having the `tuple[Json]` here only really makes sense for input/dumping to JSON, but not for reading, since tuples become lists, so a function receiving JSON will never get a tuple. I think it would make sense to use lists there for input, too, since it looks closer to the actual JSON.

It would also make sense to add `JsonDict`, since that's the things we mostly care about.

mkosi/sandbox.py
```diff
@@ -587,7 +591,152 @@ def __exit__(self, *args: object, **kwargs: object) -> None:
...
+
+        for fd in fds:
+            if isinstance(fd, int):
+                opened_fds += [fd]
+            else:
+                path, flags = fd
+                opened = os.open(path, flags | os.O_CLOEXEC)
+                opened_fds += [stack.enter_context(close(opened))]
+
+        message: dict[str, object] = {"method": method}
```
behrmann: Why is this typed as `object` instead of `Json`?

mkosi/sandbox.py
```diff
@@ -587,7 +591,152 @@ def __exit__(self, *args: object, **kwargs: object) -> None:
...
+        socket.send_fds(sock, [request], opened_fds)
+        response, reply_fds, _, _ = socket.recv_fds(sock, bufsize=10 * 1024 * 1024, maxfds=10)
+        reply = json.loads(response.decode().rstrip("\0"))
+
+        if "error" in reply:
+            for fd in reply_fds:
+                os.close(fd)
+
+            if "errnoName" in reply.get("parameters", {}):
+                num = getattr(errno, reply["parameters"]["errnoName"])
```
behrmann: if `errno` is not an attribute, then `getattr` will make this safely return `None`, but then `os.strerror` will throw an exception.

mkosi/sandbox.py
```diff
@@ -587,7 +591,152 @@ def __exit__(self, *args: object, **kwargs: object) -> None:
...
+                opened = os.open(path, flags | os.O_CLOEXEC)
+                opened_fds += [stack.enter_context(close(opened))]
+
+        message: dict[str, object] = {"method": method}
+        if parameters:
+            message["parameters"] = parameters
+
+        request = json.dumps(message).encode() + b"\x00"
+        socket.send_fds(sock, [request], opened_fds)
+        response, reply_fds, _, _ = socket.recv_fds(sock, bufsize=10 * 1024 * 1024, maxfds=10)
```
behrmann: The error that this didn't type properly is fortunately no longer
```suggestion
        response, reply_fds, _, _ = socket.recv_fds(sock, bufsize=10 * 1024**2, maxfds=10)
```

mkosi/sandbox.py
```diff
@@ -587,7 +591,157 @@ def __exit__(self, *args: object, **kwargs: object) -> None:
...
+        if "error" in reply:
+            for fd in reply_fds:
+                os.close(fd)
+
+            if "errnoName" in reply.get("parameters", {}):
+                num = getattr(errno, reply["parameters"]["errnoName"], None)
+                if num is not None:
+                    raise OSError(num, os.strerror(num))
+                else:
+                    raise OSError(num)
```
behrmann: Nit: Let's turn this around, since the case is unlikely. While at it, we can give it a reasonable error message 
```suggestion
                if num is None:
                    raise OSError(f"Method {method} returned an error but no errnoName")

                raise OSError(num, os.strerror(num))
```

mkosi/run.py
```diff
@@ -369,9 +364,10 @@ def spawn(
                     _preexec,
                     cmd=cmd,
                     env=env,
-                    sandbox=sbx if apply_sandbox_in_preexec else [],
-                    preexec=preexec,
-                ),
+                    sandbox=sbx,
+                )
+                if apply_sandbox_in_preexec and sbx
+                else None,
```
behrmann: What the ternary expression applies to is really hart to discern because it's the same indenation as the keyword arguments. You can wrap the argument to `preexec_fn` in parentheses and give it an extra layer of indentation, that's much easier to read (and doesn't even need a `# fmt: skip`).

mkosi/run.py
```diff
@@ -406,9 +402,10 @@ def spawn(
                         _preexec,
                         cmd=["bash"],
                         env=env,
-                        sandbox=sbx if apply_sandbox_in_preexec else [],
-                        preexec=preexec,
-                    ),
+                        sandbox=sbx,
+                    )
+                    if apply_sandbox_in_preexec and sbx
+                    else None,
```
behrmann: Same here.

mkosi/resources/man/mkosi-sandbox.1.md
```diff
@@ -113,6 +124,21 @@ from `mkosi.sandbox` is not supported and may break in future versions.
...
 
+`--map-foreign`
+:   Map the foreign UID range in the sandbox. For more information on the foreign UID range,
+    see [Users, Groups, UIDs and GIDs on systemd Systems](https://systemd.io/UIDS-GIDS/).
+
+    When running `mkosi-sandbox` unprivileged, this option requires `systemd-nsresourced`
+    v260 or newer.
+
+`--map-delegate N`
+:   Map N additional 64K UID/GID ranges in the sandbox.
```
behrmann, line 135: Silly nit: The official prefix for 1024² is Ki, although I've only ever seen it used as a prefix in units, but k as a suffix is only an extension of usage of k as a prefix, so this could be
```suggestion
:   Map N additional 64Ki UID/GID ranges in the sandbox.
```

mkosi/resources/man/mkosi.1.md
behrmann, line 1: While it is mentioned in the `mkosi-sandbox` man page, I think the need for nsresourced and mountfsd 260 or newer should also be mentioned in the description of the verbs that need it.

mkosi/sandbox.py
```diff
@@ -587,7 +591,157 @@ def __exit__(self, *args: object, **kwargs: object) -> None:
...
 
-def become_user(uid: int, gid: int) -> None:
+if sys.version_info >= (3, 10):
+    Json = dict[str, "Json"] | list["Json"] | str | int | float | bool | None
+    FD = int | tuple[str, int]
+else:
+    from typing import Union
+
+    Json = Union[dict[str, "Json"], list["Json"], str, int, float, bool, None]
+    FD = Union[int, tuple[str, int]]
```
behrmann, line 607: I assume the benefit of guarding the `typing` import is to not pay the import price on 3.10 and newer? Is the difference all that big? Running `python -c 'import sys' vs `python -c 'import sys, type'` under time gives a difference of 0.08(5)s vs 0.07(5)s for me, in a running interpreter, the difference is slightly starker with 185ns vs 319ns, though not in any order of magnitude one would notice.

mkosi/sandbox.py
```diff
@@ -587,7 +591,157 @@ def __exit__(self, *args: object, **kwargs: object) -> None:
...
+    delegate: int = 0,
+    become_root: bool = False,
+) -> None:
+    import uuid
+
+    varlink(
+        path="/run/systemd/io.systemd.NamespaceResource",
+        method="io.systemd.NamespaceResource.AllocateUserRange",
+        parameters={
+            "name": f"mkosi-{uuid.uuid4().hex[:4]}",
```
behrmann, line 691: Why only 4 here? Usually it's either 16 or 32?

mkosi/sandbox.py
```diff
@@ -618,10 +774,18 @@ def become_user(uid: int, gid: int) -> None:
                 os.close(event)
                 with open(f"/proc/{ppid}/setgroups", "wb") as f:
                     f.write(b"deny\n")
+
+                gidmap = f"{gid} {os.getgid()} 1\n"
+                if foreign:
+                    gidmap += f"{FOREIGN_UID_MIN} {FOREIGN_UID_MIN} 65536\n"
                 with open(f"/proc/{ppid}/gid_map", "wb") as f:
-                    f.write(f"{gid} {os.getgid()} 1\n".encode())
+                    f.write(gidmap.encode())
```
behrmann, line 761: I don't quite remember anymore why these files are opened `"wb"`, I vaguely remember that there was a reason. If there's not, there's little reason in doing the encoding oneself, if there is a reason, adding a comment at the top of the file might be helpful for the future and also `range_is_mapped` uses the only non-binary `open` call in the whole file.

mkosi/sandbox.py
```diff
@@ -1209,7 +1466,26 @@ def main(argv: list[str] = sys.argv[1:]) -> None:
...
 
-    userns = acquire_privileges(become_root=become_root, network=bool(namespaces & CLONE_NEWNET))
+    network = bool(namespaces & CLONE_NEWNET)
+    foreign = any(fsop.foreign for fsop in fsops if isinstance(fsop, BindOperation))
+
+    userns = acquire_privileges(
+        identity=True,
+        foreign=map_foreign,
+        delegate=map_delegate + int(foreign),
+        become_root=not foreign and become_root,
```
behrmann: Nit:

```suggestion
        delegate=(map_delegate + int(foreign)),
        become_root=(not foreign and become_root),
```

mkosi/resources/man/mkosi-sandbox.1.md
daandemeyer: I'm gonna stick to the systemd naming here which is `K`

mkosi/sandbox.py
daandemeyer: The limit on nsresourced side is really low

mkosi/sandbox.py
daandemeyer: Sure but we don't know if it's more on slower systems, and we don't know if it'll stay that way. Given it's trivial to avoid the import, I say let's just avoid it.

mkosi/sandbox.py
daandemeyer: I'm pretty sure there was a reason, but I can't recall either. Let's not handle this in the same PR. I don't want to accidentally introduce a hard to debug regression.

mkosi/resources/man/mkosi.1.md
```diff
@@ -1798,6 +1810,16 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
     instead of failing outright. Defaults to `no`.
 
+`ForeignUIDRange=`, `--foreign-uid-range=`
+:   If enabled, directory images will be owned by the foreign UID range on disk
+    instead of the user's UID.
+
+    Enabling this option is required to use directory images with `mkosi shell`,
+    `mkosi boot` and `mkosi vm`.
+
+    systemd-nsresourced and systemd-mountfsd 260 or newer are required on the host
```
```suggestion
    systemd-nsresourced and systemd-mountfsd v260 or newer are required on the host
```

daandemeyer:
Just doc fixes in the latest push, merging

PR #4164, opened 2026-02-13
Various stuff

daandemeyer:


mkosi/config.py
```diff
@@ -1124,7 +1125,7 @@ def config_default_initrds(namespace: dict[str, Any]) -> list[Path]:
         return []
 
     if namespace["bootable"] == ConfigFeature.auto and (
-        namespace["output_format"] == OutputFormat.cpio
+        namespace["output_format"] in (OutputFormat.cpio, OutputFormat.directory)
```
behrmann, line 1128: This should be noted in the man page, because at the moment we write
> **mkosi** will automatically build a default initrd.

without any qualifier

mkosi/__init__.py
```diff
@@ -4757,7 +4745,9 @@ def run_clean(args: Args, config: Config) -> None:
         remove_image_cache = args.force > 0
         remove_package_cache = args.force > 1
     else:
-        remove_outputs = args.force > 0 or (config.is_incremental() and not have_cache(config))
+        remove_outputs = args.force > int(config.incremental == Incremental.relaxed) or (
```
behrmann: While I'm probably guilty of writing something like this myself, this is really obscure, since it uses that bools used to be ints in Python's far past and since they never could break compat on this, True is 1. Let's add a comment to that effect
```suggestion
        # Use that True is 1 and False is 0 in numeric contexts
        remove_outputs = args.force > (config.incremental == Incremental.relaxed) or (
```
and we can drop the int call as well, since for the above reason it still type checks.

behrmann (approved):

PR #4165, opened 2026-02-13
Rework sandbox globbing and exist checks

daandemeyer:
Using bash to glob in the sandbox is rather primitive. Let's do better by taking advantage of the fact that we can run mkosi-sandbox without executing another binary. We beef up fork_and_wait() to allow passing in a sandbox and use a pipe to send the pickled result of the target function back to the parent process so we can return it from fork_and_wait(). This allows us to rewrite glob_in_sandbox() and exists_in_sandbox() to be much simpler.

At the same time, make sure we show proper stacktraces when using uncaught_exception_handler() after fork. To achieve this, we have to make sure it logs all frames, including the parent's ones, as we have to log the exception from the forked process as we can't pickle exceptions and send them to the parent.

Additionally, try to populate the line cache as early as possible as we might not be able to access source files after sandboxing ourselves.

While we're at it, we switch _preexec() over to using uncaught_exception_handler() as well.

### DIFF
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -3,7 +3,10 @@
 import contextlib
 import errno
 import functools
+import inspect
+import linecache
 import logging
+import multiprocessing
 import os
 import queue
 import shlex
@@ -16,9 +19,10 @@ import uuid
 from collections.abc import Awaitable, Collection, Iterator, Mapping, Sequence
 from contextlib import AbstractContextManager
 from pathlib import Path
-from types import TracebackType
-from typing import TYPE_CHECKING, Any, Callable, Generic, NoReturn, Protocol, TypeVar
+from types import FrameType, TracebackType
+from typing import TYPE_CHECKING, Any, Callable, Generic, NoReturn, Protocol, TypeVar, cast
 
+import mkosi
 import mkosi.sandbox
 from mkosi.log import ARG_DEBUG, ARG_DEBUG_SANDBOX, ARG_DEBUG_SHELL, die
 from mkosi.sandbox import acquire_privileges, joinpath, umask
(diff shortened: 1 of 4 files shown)


mkosi/run.py
```diff
@@ -46,21 +50,70 @@ def ensure_exc_info() -> tuple[type[BaseException], BaseException, TracebackType
     return (exctype, exc, tb)
 
 
+# Make sure the line cache is primed as the mkosi source files might become inaccessible if we sandbox
+# ourselves.
+for path in Path(mkosi.__file__).parent.rglob("*.py"):
+    linecache.getlines(os.fspath(path))
```
behrmann, line 56: I wasn't sure this would run in a zipapp, but it does.

mkosi/run.py
```diff
@@ -46,21 +50,70 @@ def ensure_exc_info() -> tuple[type[BaseException], BaseException, TracebackType
...
+        # Start from the caller's frame (skip uncaught_exception_handler() itself and
+        # the wrapping context manager).
+        frame = inspect.currentframe()
+        if frame is not None:
+            frame = frame.f_back
+        if frame is not None:
+            frame = frame.f_back
+        while frame is not None:
+            frames.append(frame)
+            frame = frame.f_back
```
behrmann: The comment above that we skip two frames made this clear, but when first looking at this, I wondered if this is not a copy paste error. I'm not a 100% sure, but I'd say the checks for None are superfluous, since we should be guaranteed to be in *a* frame when this is called, also if we don't even have two frames `frames` will be empty anyway, since in the while loop nothing will be appended to it. How about
```suggestion
        # Start from the caller's frame
        frame = inspect.currentframe()
        while frame is not None:
            frames.append(frame)
            frame = frame.f_back
        # skip uncaught_exception_handler() itself and the wrapping context manager)
        frames = frames[2:]
```

PR #4166, opened 2026-02-13
Use python3.12 as interpreter in CentOS Stream 9 zipapp

daandemeyer:


behrmann:
As mentioned via sidechannel, I'm good with this change (also, the blast radius is rather small, since it's a tools tree thing), though I still think we should just bump to 3.12. :) 

We should mention this in the news, though, since it could affect Python things used in tools trees.

PR #4168, opened 2026-02-14
Tracking ty compatibility

behrmann:
ty is a new checker from the people that brought us ruff and uv. Since it is written in Rust, it is vastly faster than mypy or pyright. It is not yet fully features, but the recent release has removed about 80% of the false positives it had before. This allows us to annotate the remaining false positives with ignore comments for ty specifically and add two type annotations that also make the intent clearer to the human reader.

At the same time, ty complains about type ignore comments, that are needed for either mypy or pyright. Unfortunately there is no syntax for type checker specific ignores, but we can silence those errors for ty. This also adds comments on which type checker makes a specific type ignore comment necessary.

Unfortunately the tests cannot yet be properly type checked with ty due to [1].

[1] https://github.com/astral-sh/ty/issues/2797

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -5199,12 +5199,12 @@ def run_verb(args: Args, tools: Optional[Config], images: Sequence[Config], *, r
         if any((c := config).is_incremental() and not have_cache(config) for config in images):
             if args.rerun_build_scripts:
                 die(
-                    f"Cannot use --rerun-build-scripts as the cache for image {c.image} is out-of-date",
+                    f"Cannot use --rerun-build-scripts as the cache for image {c.image} is out-of-date",  # ty: ignore[unresolved-reference]
                     hint="Rebuild the image to update the image cache",
                 )
             else:
                 die(
-                    f"Strict incremental mode is enabled and cache for image {c.image} is out-of-date",
+                    f"Strict incremental mode is enabled and cache for image {c.image} is out-of-date",  # ty: ignore[unresolved-reference]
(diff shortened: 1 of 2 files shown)


mkosi/__init__.py
```diff
@@ -5187,12 +5194,12 @@ def run_verb(args: Args, tools: Config | None, images: Sequence[Config], *, reso
         if any((c := config).is_incremental() and not have_cache(config) for config in images):
```
daandemeyer, line 5199: You can write this as a regular for loop with a check at the beginning to continue to not have to add the comments I think

mkosi/config.py
```diff
@@ -4639,7 +4641,7 @@ def resolve_deps(images: Sequence[Config], include: Sequence[str]) -> list[Confi
     graph = {config.image: config.dependencies for config in images}
 
     if any((missing := i) not in graph for i in include):
```
daandemeyer, line 4645: Same here

behrmann:
Drafting this now to keep this around as a tracking PR what would be needed to make mkosi check successfully with ty. This should be merged as is, since ty should learn the things that are currently missing and adding the ignores here could silence them in their CI, which also checks mkosi.

mkosi/__init__.py
behrmann: That's tracked in [this issue](https://github.com/astral-sh/ty/issues/162) and already has [a PR](https://github.com/astral-sh/ruff/pull/23167)

PR #4169, opened 2026-02-14
Revert "Bump minimum python version to 3.10"

daandemeyer:
This reverts commit https://github.com/systemd/mkosi/commit/22b2f0bf18ac98f62ef92745fde5dd3f8369d4bf.

Turns out using python3.12 on CentOS causes more issues than
we thought it would, so let's revert the move to python 3.10.
Instead, we'll conditionally import Union in sandbox.py only on
python 3.9 and use the Union operator otherwise.


behrmann (approved):
Sad. :(

PR #4175, opened 2026-02-16
Remove all cached images if repository metadata will be synced

daandemeyer:
If we don't do this, and packages are installed in postinstall scripts, we can end up in partial upgrade situations, where we install packages that depend on newer versions of libraries that are installed. This is particularly a problem on Arch Linux, which explicitly does not support partial upgrades.

To solve the issue, if we decide we need to sync repository metadata, remove all cached images.

To avoid removing syncing repository metadata (and thus removing all cached images) when a cached image is invalidated, CacheOnly=metadata can be configured for the image.


mkosi/__init__.py
```diff
@@ -4825,6 +4828,17 @@ def ensure_directories_exist(config: Config) -> None:
             config.build_subdir.chmod(stat.S_IMODE(st.st_mode) & ~(stat.S_ISGID | stat.S_ISUID))
 
 
+def repository_metadata_needs_sync(images: Sequence[Config]) -> bool:
+    for config in images:
+        if config.cacheonly == Cacheonly.never:
+            return True
```
behrmann, line 4834: This should be added to the man page under `CacheOnly=`, because currently it only mentions the original behaviour below.

PR #4185, opened 2026-02-20
kmod: actually ignore missing module deps

craftyguy:
Prior to f9e7527e, IIUC modules in todo were ignored when they were popped and `moddep.get(m)` returned None for any modules that were missin. After f9e7527e, missing modules aren't actually ignored, which leads to e.g.:

```
‣ aes is a dependency of cifs but is not installed, ignoring
‣ nls is a dependency of cifs but is not installed, ignoring
‣ aead2 is a dependency of cifs but is not installed, ignoring
modinfo: ERROR: Module name nls not found.
```

(nls.ko doesn't exist, and shouldn't exist in recent (6.19-ish) kernel trees)

### DIFF
--- a/mkosi/kmod.py
+++ b/mkosi/kmod.py
@@ -317,7 +317,7 @@ def resolve_module_dependencies(
 
             mods.add(name)
             firmware.update(depinfo.firmware)
-            todo += [m for m in depinfo.modules if m not in mods]
+            todo += [m for m in depinfo.modules if m not in mods and m in nametofile]
 
     return set(nametofile[m] for m in mods if m in nametofile), set(firmware)
 


daandemeyer:
I don't think we need to look at builtin, I've had the same patch sitting in the buildstream PR (https://github.com/systemd/mkosi/pull/4149/changes/47fa51e31773f40041d4afbdf46b7bda7ddd5478) so if you copy that we can merge it unless there's a good reason to go with this version.

craftyguy:
Ahh ok, yeah I was just copying the condition that lead to the "ignoring" message being printed. It seems weird to me to have that message printed from a different set of conditions from the actual behavior.

craftyguy:
Updated to drop the builtin condition

PR #4187, opened 2026-02-23
mkosi-tools: Install dnf5 on debian when available

daandemeyer:


mkosi/resources/mkosi-tools/mkosi.profiles/package-manager/mkosi.conf.d/debian-kali-ubuntu/mkosi.conf.d/dnf5.conf
```diff
@@ -0,0 +1,13 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+[TriggerMatch]
+Distribution=debian
+Release=|testing
```
bluca, line 7: might want to add `forky` and `duke` already so that when they ship it works

PR #4190, opened 2026-02-24
Allow nesting mkosi box both sessions and indicate nesting in prompt

pothos:
So far nesting mkosi box was forbidden because it's confusing if there is no indicator. For scripts running commands it would not be an issue, though, and one would force MKOSI_IN_BOX=0 for that. But with SHELL_PROMPT_PREFIX becomming common we can now indicate the current level of nesting.
Allow mkosi box nesting for both interactive and scripted usage without requring a MKOSI_IN_BOX overwrite. Indicate the current nesting level in the shell prompt.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -4172,12 +4172,6 @@ def build_image(context: Context) -> None:
 
 
 def run_box(args: Args, config: Config) -> None:
-    if in_box():
-        die(
-            "mkosi box cannot be invoked from within another mkosi box environment",
-            hint="Exit the current mkosi box environment and try again",
-        )
-
     if not args.cmdline:
         die("Please specify a command to execute in the sandbox")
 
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -4197,6 +4191,16 @@ def run_box(args: Args, config: Config) -> None:
     hd, hr = detect_distribution()
 
     env = {"MKOSI_IN_BOX": "1"}
+
+    prefix = os.getenv("SHELL_PROMPT_PREFIX", "")
+    m = re.search(r"\(box(?::(\d+))?\)", prefix)
```
behrmann: Making the group named would be nice to see what it is we're handling
```suggestion
    m = re.search(r"\(box(?::(?P<level>\d+))?\)", prefix)
```
then you can use `m.group("level")` below. Also I'd go for
```suggestion
    m = re.search(r"\(box(?::([1-9][0-9]+))?\)", prefix)
```
since `\d` matches arbitrary unicode number things and we're only really equipped to handle ordinary numbers here.

pothos: Thanks, taken :) (It's `*` instead of `+` for the second digit)

Manually tested going to 21 and all seemed good.

PR #4195, opened 2026-02-26
debian: Install mount explicitly

daandemeyer:
Not pulled in by default anymore.

### DIFF
--- a/mkosi.tools.conf/mkosi.conf
+++ b/mkosi.tools.conf/mkosi.conf
@@ -6,3 +6,4 @@ Packages=
         pandoc
         gnupg
         lvm2
+        udev


behrmann (approved):

mkosi.conf.d/debian-kali-ubuntu/mkosi.conf
```diff
@@ -10,3 +10,4 @@ Packages=
         openssh-client
         openssh-server
         python3
+        passwd
```
bluca, line 13: I'm actually trying to get rid of this, so we can build images without setuid - I think it's not in by default because I managed to fix the last one in the set that pulled it in

bluca: is there any reason you need it?

daandemeyer: useradd?

bluca: do we need that for mkosi? or for manual usage?

daandemeyer: We use it to create a test user for the mkosi repo image

bluca: userdbd snippet?

daandemeyer: Old distros

bluca: it's from 245 in 2019, should be ok by now? https://github.com/systemd/mkosi/pull/4202

PR #4197, opened 2026-03-01
mkosi-cmd: Add VM-side command channel daemon for mkosi cmd

gourryinverse:
Some automation cannot use typical network sockets (vsock and standard sockets), but can use unix domain sockets.  This requires a QEMU serial device attached to a host domain socket and a daemon running in the guest which provides the terminal access.

The 'mkosi cmd' feature provides this interface for VMs running with Console=headless mode, allowing automation to interact with the guest without running into network jailing issues.

Along with the command addition (in config) we ship:
    - The daemon script
    - systemd service setup
    - preset to launch the service by default

The daemon exist cleanly after 30s if no serial device appears (e.g. Console != headless).

When in headless mode, add two serial devices:
    - One writes kmsg/dmesg output to /tmp/mkosi-serial.log
    - One is the mkosi-cmd unix domain socket for interactions

Signed-off-by: Gregory Price <gourry@gourry.net

### DIFF
--- a/mkosi.extra/usr/lib/systemd/system-preset/00-mkosi.preset
+++ b/mkosi.extra/usr/lib/systemd/system-preset/00-mkosi.preset
@@ -23,3 +23,6 @@ disable auditd.service
 
 # systemd-timesyncd is not enabled by default in the default systemd preset so enable it here instead.
 enable systemd-timesyncd.service
+
+# Command channel daemon for host interaction via mkosi cmd
+enable mkosi-cmd-daemon.service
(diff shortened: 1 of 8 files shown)


gourryinverse:
> Sorry but this is very much out of scope for mkosi. Even `mkosi ssh` is slowly on the way out and will be replaced by `ssh machine/xxx` in the future via systemd-ssh-proxy. I would much rather we discuss ways to make `ssh` work for your use case.
> 
> qemu's vsock support can also be done via unix sockets on the host instead of vsock. I'd rather we enable that.

hm, i'll see if i can rework the vsock option from it's current yes/no to maybe (yes/no) for compat but maybe + "domain" ? 

Either way, ok, will see if i can rework this.

gourryinverse:
replaced with ssh machine/* fixup for TCP loopback

PR #4199, opened 2026-03-02
Improve log output

pothos:
- Add context to log output
    
    Currently it's not really easy to know which log output belongs to which
    mkosi image when there's the tools tree image, the main image, and any
    sub images.
    Add a log prefix that remembers the current image so that it's not just
    "‣  Installing Fedora Linux" but  "‣ [tools]  Installing Fedora Linux",
    "‣ [default-initrd]  Installing Fedora Linux", and
    "‣ [main]  Installing Fedora Linux".
- Allow printing of step durations with CLI flag
    
    The printing of the duration for each step was behind the --debug flag
    which adds more noise when one is just interested in the build times.
    Also, the printing of the step duration serves as good marker during
    log reading to understand where the above log output belongs to.
    Add a new --log-timing flag to always print the step durations.

I first thought about prefixing all output lines but with the two changes above it works well enough to frame long output. Prefixing all output including the scripts would mean to emulate a TTY so that colors and so on would still be used. That could be a good idea as well if one needs timestamps for every line but I think that would be something for another PR.


mkosi/log.py
```diff
@@ -112,24 +122,31 @@ def complete_step(text: str, text2: Optional[str] = None) -> Iterator[list[Any]]
...
         }  # fmt: skip
 
         super().__init__(fmt, *args, **kwargs)
 
     def format(self, record: logging.LogRecord) -> str:
+        image = IMAGE.get()
+        if image:
+            record.image_prefix = f"{ANSI_BLUE}[{image}]{ANSI_RESET} "
+        else:
+            record.image_prefix = ""
```
behrmann: That we can shorten down a bit
```suggestion
        record.image_prefix = f"{ANSI_BLUE}[{image}]{ANSI_RESET} " if IMAGE.get() else ""
```
Though this is unlikely to clash, since this is our addition, I wonder whether we should make this `._image_prefix` instead.

pothos:
Had to rebase but since I moved the IMAGE.set to the build_image function I wanted to test again but ran into some other issue with latest main.

pothos:
Should be good now

mkosi/config.py
```diff
@@ -4534,6 +4535,12 @@ def create_argument_parser(chdir: bool = True) -> argparse.ArgumentParser:
         action="store_true",
         default=False,
     )
+    parser.add_argument(
+        "--log-timing",
```
daandemeyer: Let's name this `--debug-perf` or `--debug-timing` or so?

daandemeyer: Needs to be added to the man page as well

pothos: Thanks! Went with `debug-timing`.

PR #4200, opened 2026-03-02
Story outputs in config directory by default

daandemeyer:
If no output location is explicitly configured and we have a mkosi/ subdirectory, let's store the .mkosi-private directory and the tools tree and everything else there instead of in the top level directory.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -2251,7 +2251,7 @@ class Config:
         return self.machine or self.image
 
     def output_dir_or_cwd(self) -> Path:
-        return self.output_dir or Path.cwd()
+        return self.output_dir or finalize_configdir(Path.cwd()) or Path.cwd()
 
     def workspace_dir_or_default(self) -> Path:
         if self.workspace_dir:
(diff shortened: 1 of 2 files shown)


mkosi/config.py
```diff
@@ -5370,8 +5370,11 @@ def parse_config(
     if not args.verb.needs_config():
         return args, None, ()
 
-    if have_history(args):
-        history = Config.from_partial_json(Path(".mkosi-private/history/latest.json").read_text())
+    configdir = finalize_configdir(args.directory)
+    historydir = configdir or Path.cwd()
+
+    if have_history(args, historydir):
+        history = Config.from_partial_json((historydir / ".mkosi-private/history/latest.json").read_text())
```
behrmann: That seems a bit odd. Wouldn't `.mkosi-private/history` make more sense as `historydir`? 

Having all the outputs go to `.mkosi-private/output` instead of `.mkosi-private` would also make this nicely sorted and prevent any clashes.

daandemeyer: Makes sense, introduced `finalize_historydir()` and use that everywhere now instead.

PR #4201, opened 2026-03-02
Fix error checks for nsresource_allocate_user_range()

daandemeyer:
We have to check for VarlinkError(org.varlink.service.InvalidParameter) instead of errno.EINVAL. While we're at it, check for FileNotFoundError instead of OSError(ENOENT) as well.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2,7 +2,6 @@
 
 import contextlib
 import datetime
-import errno
 import functools
 import getpass
 import hashlib
@@ -131,6 +130,7 @@ from mkosi.run import (
 )
 from mkosi.sandbox import (
     CLONE_NEWNS,
+    VarlinkError,
     __version__,
     acquire_privileges,
     join_new_session_keyring,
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
```diff
@@ -4986,10 +4986,14 @@ def run_verb(args: Args, tools: Optional[Config], images: Sequence[Config], *, r
...
+    except Exception as e:
         # Don't fail if systemd-nsresourced is too old or not installed, use a regular unpriv user namespace
         # instead.
-        if isinstance(e, OSError) and e.errno not in (errno.EINVAL, errno.ENOENT):
+
+        if not isinstance(e, (FileNotFoundError, VarlinkError)):
+            raise
+
+        if isinstance(e, VarlinkError) and e.error != "org.varlink.service.InvalidParameter":
             raise
```
behrmann: The instance checks are basically the same as the native syntax
```suggestion
    # Don't fail if systemd-nsresourced is too old or not installed, use a regular unpriv user namespace
    # instead.
    except (FileNotFoundError, VarlinkError) as e:
        if isinstance(e, VarlinkError) and e.error != "org.varlink.service.InvalidParameter":
            raise
        logging.debug(
            f"Could not provision user namespace via systemd-nsresourced ({e}), falling back to "
            "unprivileged user namespace via unshare(CLONE_NEWUSER) and writing /proc/self/uid_map directly"
        )
    except Exception:
        raise
```

Issue #4173, opened 2026-02-14
Adding repositories with local config broken on Arch


leoratte:
I added `Profiles=obs-repos` to my mkosi.local.conf and the build broke with:
```
mkosi -B -f
‣ Running bump script /home/leo/git/particleos/mkosi.bump
‣ Bumping version: 'n/a' → '20260214201224'
‣ Copying in sandbox trees…
‣ Syncing package manager metadata
:: Synchronizing package databases...
 system_systemd_Arch                                      6.8 KiB   103 KiB/s 00:00 [################################################] 100%
 core                                                   123.6 KiB  1585 KiB/s 00:00 [################################################] 100%
 extra                                                    7.9 MiB  11.1 MiB/s 00:01 [################################################] 100%
warning: Public keyring not found; have you run 'pacman-key --init'?
error: system_systemd_Arch: key "713BF72732820778" is unknown
error: keyring is not writable
error: failed to synchronize all databases (unexpected error)
‣ "pacman --root=/buildroot --logfile=/dev/null --dbpath=/var/lib/pacman --cachedir=/var/cache/pacman/mkosi --cachedir=/var/cache/pacman/pkg --hookdir=/buildroot/etc/pacman.d/hooks --arch x86_64 --color auto --noconfirm --sync --refresh" returned non-zero exit code 1.
‣ (sandbox command omitted, rerun with --debug to see the full sandbox command.)
```

I think this is caused by [mkosi commit ](https://github.com/systemd/mkosi/pull/4093), which moves pacman keyring initialization after database synchronization in mkosi/__init__.py.

mkosi.local.conf for reference:
```
[Distribution]
Distribution=arch

[Config]
Profiles=obs-repos
Profiles=desktop,kde,flathub

[Validation]
SecureBoot=no
SignExpectedPcr=no
```
I' m trying this on manjaro.

bluca:
Regression introduced by https://github.com/systemd/mkosi/pull/4093 as hinted above, minimal reproducer:

```
[Build]
SandboxTrees=systemd.gpg:/usr/share/pacman/keyrings/systemd.gpg
             systemd-trusted:/usr/share/pacman/keyrings/systemd-trusted
             arch.conf:/etc/pacman.d/systemd.conf
```

With arch.conf systemd-trusted and systemd.gpg from https://github.com/systemd/particleos/tree/main/mkosi.profiles/obs-repos

JiwaniZakir:
[...]

PR #4203, opened 2026-03-04
initialise keyring earlier in the sandbox lifecycle

arcayr:
#4093 moved the keyring initialisation later in the setup lifecycle as part of normalising arch/pacman setup. if any custom repositories are used, pacman requires the keys to exist for all configured repositories from the get-go or exits immediately. let's set this up earlier in the sandbox lifecycle to provide the requisite keys.

i haven't verified if this breaks any distro other than arch, but i couldn't find any rationale for moving it in the git commit logs so i don't know if it was moved intentionally or incidentally.

original cut of this simply had the pacman installer run `pacman-key` twice (once at setup, and once where it currently runs), but that's is basically begging for hysteresis and i couldn't find any other example of that sort of special casing in the codebase.

marked as draft because i'll check other distributions this weekend. if i'm operating on a bad assumption here i can rework this as required.

### DIFF
--- a/mkosi/distribution/arch.py
+++ b/mkosi/distribution/arch.py
@@ -82,6 +82,7 @@ class Installer(DistributionInstaller, distribution=Distribution.arch):
     @classmethod
     def setup(cls, context: Context) -> None:
         Pacman.setup(context, list(cls.repositories(context)))
+        Pacman.keyring(context)
 
     @classmethod
     def install(cls, context: Context) -> None:
--- a/mkosi/distribution/arch.py
+++ b/mkosi/distribution/arch.py
@@ -82,6 +82,9 @@ class Installer(DistributionInstaller, distribution=Distribution.arch):
     @classmethod
     def setup(cls, context: Context) -> None:
         Pacman.setup(context, list(cls.repositories(context)))
+        # pacman's keyring needs to be initialized and pre-seeded with any drop-in keys prior
+        # to its first sync, otherwise it'll refuse to continue when using custom repositories
+        # due to missing signatures.
         Pacman.keyring(context)
 
     @classmethod


arcayr:
simplest reproducer of the original problem can be found from ~bluca [here](https://github.com/systemd/mkosi/issues/4173#issuecomment-3904726765), however the setup i had was adding cachyos repositories.

while i can see _why_ it's happening after the fact, i cannot replicate the build failure even on a completely fresh config, so i'll need to spend some time reverse-investigating to find a middle ground.

daandemeyer:
This does not work because we need to sync the repository metadata to be able to download the Arch keyring using pacman if --repository-key-fetch= is enabled.

arcayr:
posted on matrix but copying it here because it's easier to follow.

i guess my core issue is that i need to call `pacman-key --init` and `pacman-key --populate --populate-from /usr/share/pacman/keyrings` before the [first sync](https://github.com/systemd/mkosi/blob/main/mkosi/__init__.py#L4873), but because of how pacman-key and pacman expect to work with this keyring i need to do this inside the sandbox. this worked prior to #4093 because the keyring package download and import happened out of band, which indirectly picked up any other signing keys that existed in the directory.

i patched my local to simply call `Pacman.keyring` a second (technically a first, as it will fire first) time [at setup](https://github.com/systemd/mkosi/blob/main/mkosi/distribution/arch.py#L84). this doesn't appear to have had any adverse effects, but i'm not sure this is the approach you'd want to take. if that makes sense to you i'm happy to update the pr.

i have a few alternatives in mind but all of them would require a bigger rework of the `DistributionInstaller`, and it's only really used for arch and specific flavours of opensuse at the moment.

arcayr:
alright double keyring calling it is. unsetting `RepositoryKeyFetch` and preloading all keys into `mkosi.sandbox/usr/share/pacman/keyrings` doesn't solve this for arch due to it being unable to import those into the keyring at sync time.

running `pacman-key --init` when a keyring exists isn't harmful to the existing keyring, and this is isolated to arch so doesn't affect opensuse's keyring components.

cvlc12:
Seems to solve it for me !

cvlc12:
Any update on this ? This does seem to make it possible to use `obs-repos` with particleos on Arch.

PR #4208, opened 2026-03-07
Add workflow and script to automatically update distribution snapshots

daandemeyer:
Add a GitHub Actions workflow that runs every three days to pin each distribution to its latest snapshot. The workflow runs tools/update-snapshot.py for each distribution (arch, centos, debian, fedora, opensuse, ubuntu) which fetches the latest snapshot via "mkosi latest-snapshot", updates the corresponding mkosi config file with the new Snapshot= setting, and commits the result.

The workflow then creates a pull request for the update using github-script. Each distribution gets its own dedicated branch (update-snapshot/<distro>) so that only a single pull request is open per distribution at a time. If an existing pull request is already open, it is updated in place by force-pushing to the branch.

Auto-merge is enabled on each pull request via the GraphQL enablePullRequestAutoMerge mutation so that the snapshot update lands automatically once CI is green.


.github/workflows/update-snapshot.yml
```diff
@@ -0,0 +1,114 @@
...
+on:
+  schedule:
+    # Update snapshots every three days at midnight
+    - cron: "0 0 */3 * *"
+  workflow_dispatch:
+
+jobs:
+  update-snapshot:
+    runs-on: ubuntu-24.04
+    if: github.repository == 'daandemeyer/mkosi'
```
behrmann: shouldn't that be 
```suggestion
    if: github.repository == 'systemd/mkosi'
```

tools/update-snapshot.py
```diff
@@ -0,0 +1,138 @@
...
+    args = parse_args()
+
+    path = find_config(args.distribution)
+    release = read_release(path)
+
+    cmd = [
+        "mkosi",
+        "-d", args.distribution,
+        "latest-snapshot",
+    ]
```
behrmann:
```suggestion
    ]  # fmt: skip
```
just the be on the safe side.

.github/workflows/update-snapshot.yml
keszybz: backdoor detected!!

keszybz:
Looks reasonable…

mkosi/curl.py
```diff
@@ -1,61 +1,55 @@
...
+        if pattern == "*":
+            return False
+
+        if host == pattern or host.endswith(f".{pattern}"):
+            return False
+
+    return True
+
+
+def curl(config: Config, url: str) -> str:
```
behrmann: What's the idea behind changing this to stdlib? I think that's one of the less-used parts of the stdlib because more people just use requests.

The function and module are now also something of a misnomer.

daandemeyer: I wanted to be able to run `latest-snapshot` without needing a tools tree.

daandemeyer: Happy to rename

daandemeyer: Renamed to fetch

tools/mkosi-manifest-diff.py
```diff
@@ -0,0 +1,72 @@
...
+        old = json.load(f)
+
+    with open(sys.argv[2]) as f:
+        new = json.load(f)
+
+    old_packages = {(p["name"], p["architecture"]): p["version"] for p in old.get("packages", [])}
+    new_packages = {(p["name"], p["architecture"]): p["version"] for p in new.get("packages", [])}
+
+    old_keys = set(old_packages.keys())
+    new_keys = set(new_packages.keys())
```
behrmann: The `set()` call around them is not really necessary, `dict_keys` already support all set operations by themselves.

tools/mkosi-manifest-diff.py
```diff
@@ -0,0 +1,72 @@
...
+        print()
+
+    counts = []
+    if changed:
+        counts.append(f"{len(changed)} upgraded")
+    if added:
+        counts.append(f"{len(added)} added")
+    if removed:
+        counts.append(f"{len(removed)} removed")
+    print(f"**Summary:** {', '.join(counts)}")
```
behrmann: Nit: You could just go with
```python
print(f"**Summary:** {len(changed)} upgraded, {len(added)} added, {len(removed)} removed")
```
This way the last line always has the same structure, which, at least I, find visually easier when comparing output.

.github/actions/mkosi-update-snapshot/action.yml
```diff
@@ -0,0 +1,202 @@
...
+      shell: bash
+      env:
+        DISTRIBUTION: ${{ inputs.distribution }}
+        RELEASE: ${{ inputs.release }}
+        CONFIG: ${{ inputs.config }}
+      run: |
+        branch="update-snapshot/${DISTRIBUTION:-default}"
+        git checkout -B "$branch"
+
+        mkosi ${DISTRIBUTION:+-d "$DISTRIBUTION"} ${RELEASE:+-r "$RELEASE"} latest-snapshot -- --update "$CONFIG" --commit
```
behrmann, line 55: The `latest-snapshot` command is not documented. I think it should be, since it's quite useful for the snapshot functionality.

It's also a bit of a break with the rest of the config parsing, that this implements it's own argument parser. That should be documented. It seems a bit peculiar, since so far every subcommand's options just went into the big option bag.

mkosi/snapshot.py
```diff
@@ -0,0 +1,97 @@
...
+                    # Insert before the blank line.
+                    result.pop()
+                    result.append(f"Snapshot={snapshot}")
+                    result.append(line)
+                    added = True
+                    in_distribution = False
+
+        if not added:
+            # No [Distribution] section exists; add one.
+            result = new + ["", "[Distribution]", f"Snapshot={snapshot}"]
```
behrmann: This has the edge case, that if there is just an empty distribution section
```
[Distribution]
```
add the end of the file, it will add a second distribution section. So adding the section header should be contingent on not being `in_distribution`.

PR #4238, opened 2026-03-25
run_locale_gen: noop if output_format is confext

257:
Signed-off-by: Paymon MARANDI <paymon@encs.concordia.ca>

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3135,6 +3135,10 @@ def run_tmpfiles(context: Context) -> None:
 
 
 def run_locale_gen(context: Context) -> None:
+    # locale should be generated for the base root tree, makes no sense to generate it for extensions
+    if context.config.overlay or context.config.output_format.is_extension_image():
+        return
+
     if not (context.root / "etc/locale.gen").exists():
         return
 


behrmann (approved):

daandemeyer (commented):
Please use the same pattern for skipping as the other similar functions. It should be skipped for all extensions, not just confext

257:
> Please use the same pattern for skipping as the other similar functions. It should be skipped for all extensions, not just confext

do you consider `oci` and `addon` as extensions too?

257:
> do you consider `oci` and `addon` as extensions too?

`oci` needs locale, so i just added the `addon` to the check.

257:
> > do you consider `oci` and `addon` as extensions too?
> 
> `oci` needs locale, so i just added the `addon` to the check.

just came across the right call; disregard the comment.

mkosi/__init__.py
```diff
@@ -3135,6 +3135,10 @@ def run_tmpfiles(context: Context) -> None:
 
 
 def run_locale_gen(context: Context) -> None:
+    # locale should be generated for the base root tree, makes no sense to generate it for extensions
+    if context.config.overlay or context.config.output_format.is_extension_or_portable_image():
```
behrmann, line 3139: Actually, thinking more about this, I think this should just be
```suggestion
    if context.config.overlay or context.config.output_format.is_extension_image():
```
since portable images are full OS trees in the sense that they need to contain everything the portable service might need. That *might* include locale data, so in that case that should be guarded by the check for `/etc/locale.gen` below.

keszybz (approved):
Makes sense.

PR #4247, opened 2026-03-27
feature: Add Bootloader=uki_prebuilt for distro-shipped UKIs

### LINKED ISSUE #4174: Support distro UKI + shim direct boot via BOOTX64.CSV
We’re building images that use distro UKIs with the distro shim and ran into the following; we’d like to propose a small extension.

- In our use case, we rely on distro shipped UKI and distro shim(e.g Fedora kernel-uki-virt + shim-x64).
- For that we need the UKI on the ESP and BOOTX64.CSV so shim boots the UKI directly (no grub).
- Currently we work around this by using a post-install script (copy UKI to e.g. $BUILDROOT/efi/EFI/Linux/, write efi/EFI/<vendor>/BOOTX64.CSV); repart CopyFiles then puts it on the ESP.)
- It would be helpful if mkosi had an option that does this (discover distro UKI , put it where repart copies to ESP, write BOOTX64.CSV) so we can drop the script.

src-up:
Some distributions (like Fedora) ship their own prebuilt Unified Kernel Images (UKIs). In some use cases, it's helpful to use mkosi to build images that consume these distro-shipped UKIs rather than building a UKI from scratch.

One such use case is testing [cvmutils](https://gitlab.com/vkuznets/cvmutils/-/tree/cvmutils), where the goal is to replace [ad-hoc shell scripts](https://gitlab.com/vkuznets/cvmutils-tests-e2e/-/blob/master/mkimage.sh?ref_type=heads) that handle image building with mkosi. However, doing so currently requires [extra scripts](https://gitlab.com/simsingh/cvmutils/-/tree/mkosi/tests/integration/mkosi/config/fedora?ref_type=heads) to bridge the gap.

This PR fills that gap by allowing mkosi to natively consume prebuilt UKIs, eliminating the need for any additional scripts.

Fixes https://github.com/systemd/mkosi/issues/4174

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -83,6 +83,7 @@ from mkosi.config import (
     summary,
     systemd_tool_version,
     want_kernel,
+    want_prebuilt_uki,
     want_selinux_relabel,
     yes_no,
 )
@@ -2144,10 +2145,7 @@ def install_uki(
     with umask(~0o700):
         boot_binary.parent.mkdir(parents=True, exist_ok=True)
 
-    if (
-        context.config.bootloader.is_signed()
-        and context.config.unified_kernel_images == UnifiedKernelImage.auto
-    ) or context.config.unified_kernel_images == UnifiedKernelImage.signed:
+    if want_prebuilt_uki(context.config):
         for p in (context.root / "usr/lib/modules" / kver).glob("*.efi"):
             log_step(f"Installing prebuilt UKI at {p} to {boot_binary}")
             copyfile2(p, boot_binary)
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -30,6 +30,7 @@ from typing import Optional
 from mkosi.bootloader import KernelType
 from mkosi.config import (
     Args,
+    Bootloader,
     Config,
     ConfigFeature,
     ConsoleMode,
@@ -1314,9 +1315,16 @@ def run_qemu(args: Args, config: Config) -> None:
         if kernel and (kerneltype != KernelType.uki or not config.architecture.supports_smbios(firmware)):
             cmdline += ["-append", " ".join(config.kernel_command_line + kcl)]
         elif config.architecture.supports_smbios(firmware):
+            # With Bootloader=uki-signed, a UKI built by the distro is used, and we cannot embed
+            # config.kernel_command_line in the UKI. Instead, pass those options through SMBIOS type#11. We
+            # know that this will work because all UEFI systems support SMBIOS and UKIs by construction use
+            # systemd-stub, which reads SMBIOS type#11.
+            stub_kcl = (
+                config.kernel_command_line if config.bootloader == Bootloader.uki_signed else []
+            ) + kcl
             cmdline += [
                 "-smbios",
-                f"type=11,value=io.systemd.stub.kernel-cmdline-extra={' '.join(kcl).replace(',', ',,')}",
+                f"type=11,value=io.systemd.stub.kernel-cmdline-extra={' '.join(stub_kcl).replace(',', ',,')}",  # noqa: E501
                 "-smbios",
                 f"type=11,value=io.systemd.boot.kernel-cmdline-extra={' '.join(kcl).replace(',', ',,')}",
             ]
--- a/tests/test_boot.py
+++ b/tests/test_boot.py
@@ -5,7 +5,7 @@ import subprocess
 
 import pytest
 
-from mkosi.config import Bootloader, Firmware, OutputFormat
+from mkosi.config import Architecture, Bootloader, Firmware, OutputFormat
 from mkosi.distribution import Distribution
 from mkosi.run import find_binary, run
 from mkosi.versioncomp import GenericVersion
@@ -59,11 +59,37 @@ def test_format(config: ImageConfig, format: OutputFormat) -> None:
 
 @pytest.mark.parametrize("bootloader", Bootloader)
 def test_bootloader(config: ImageConfig, bootloader: Bootloader) -> None:
-    if config.distribution == Distribution.rhel_ubi or bootloader.is_signed():
+    if config.distribution == Distribution.rhel_ubi or (
+        bootloader.is_signed() and bootloader != Bootloader.uki_signed
+    ):
+        return
+
+    # TODO: want_prebuilt_uki() also fires for UnifiedKernelImage=signed with a non-signed bootloader,
+    # but there is no integration test for that path yet.
+    # uki-signed test matrix:
+    #   x86-64 Fedora  → supports_smbios(uefi)=True,  kernel-uki-virt available → runs
+    #   arm64  Fedora  → supports_smbios(uefi)=True,  kernel-uki-virt available → runs
+    #   ppc64le Fedora → supports_smbios(uefi)=False                             → skipped
+    #   non-Fedora     → no kernel-uki-virt equivalent                           → skipped
+    if bootloader == Bootloader.uki_signed and (
+        config.distribution != Distribution.fedora
+        or not Architecture.native().supports_smbios(Firmware.uefi)
+    ):
         return
 
     firmware = Firmware.linux if bootloader == Bootloader.none else Firmware.auto
 
     with Image(config) as image:
-        image.build(["--format=disk", "--bootloader", str(bootloader)])
+        image.build(
+            [
+                "--format=disk",
+                "--bootloader",
+                str(bootloader),
+                *(
+                    ["--incremental=no", "--cache-key=&d~&r~&a~&I~prebuilt"]
+                    if bootloader == Bootloader.uki_signed
+                    else []
+                ),
+            ]
+        )
         image.vm(["--firmware", str(firmware)])
--- a/tests/test_boot.py
+++ b/tests/test_boot.py
@@ -5,7 +5,7 @@ import subprocess
 
 import pytest
 
-from mkosi.config import Bootloader, Firmware, OutputFormat
+from mkosi.config import Architecture, Bootloader, Firmware, OutputFormat
 from mkosi.distribution import Distribution
 from mkosi.run import find_binary, run
 from mkosi.versioncomp import GenericVersion
@@ -59,11 +59,33 @@ def test_format(config: ImageConfig, format: OutputFormat) -> None:
 
 @pytest.mark.parametrize("bootloader", Bootloader)
 def test_bootloader(config: ImageConfig, bootloader: Bootloader) -> None:
-    if config.distribution == Distribution.rhel_ubi or bootloader.is_signed():
+    if config.distribution == Distribution.rhel_ubi or (
+        bootloader.is_signed() and bootloader != Bootloader.uki_signed
+    ):
+        return
+
+    # TODO: want_prebuilt_uki() also fires for UnifiedKernelImage=signed with a non-signed bootloader,
+    # but there is no integration test for that path yet.
+    # uki-signed test matrix:
+    #   x86-64 Fedora  → supports_smbios(uefi)=True,  kernel-uki-virt available → runs
+    #   arm64  Fedora  → supports_smbios(uefi)=True,  kernel-uki-virt available → runs
+    #   ppc64le Fedora → supports_smbios(uefi)=False                             → skipped
+    #   non-Fedora     → no kernel-uki-virt equivalent                           → skipped
+    if bootloader == Bootloader.uki_signed and (
+        config.distribution != Distribution.fedora
+        or not Architecture.native().supports_smbios(Firmware.uefi)
+    ):
         return
 
     firmware = Firmware.linux if bootloader == Bootloader.none else Firmware.auto
 
     with Image(config) as image:
-        image.build(["--format=disk", "--bootloader", str(bootloader)])
+        image.build(
+            [
+                "--format=disk",
+                "--bootloader",
+                str(bootloader),
+                *(["--incremental=no"] if bootloader == Bootloader.uki_signed else []),
+            ]
+        )
         image.vm(["--firmware", str(firmware)])
--- a/tests/test_boot.py
+++ b/tests/test_boot.py
@@ -85,7 +85,7 @@ def test_bootloader(config: ImageConfig, bootloader: Bootloader) -> None:
                 "--format=disk",
                 "--bootloader",
                 str(bootloader),
-                *(["--incremental=no"] if bootloader == Bootloader.uki_signed else []),
+                *(["--incremental=yes"] if bootloader == Bootloader.uki_signed else []),
             ]
         )
         image.vm(["--firmware", str(firmware)])
--- a/mkosi.conf.d/fedora/mkosi.conf.d/arm64-uki-signed.conf
+++ b/mkosi.conf.d/fedora/mkosi.conf.d/arm64-uki-signed.conf
@@ -1,0 +1,13 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+[Match]
+Architecture=arm64
+Bootloader=uki-signed
+
+[Content]
+# kernel-uki-virt ships a pre-built UKI for virtual machines, needed to test
+# Bootloader=uki-signed. Not available on all Fedora architectures (e.g. ppc64le),
+# so kept in arch-specific configs and gated on Bootloader=uki-signed. A separate
+# file is needed to match both Architecture= and Bootloader= simultaneously.
+Packages=
+        kernel-uki-virt
--- a/mkosi.conf.d/fedora/mkosi.conf.d/x86_64-uki-signed.conf
+++ b/mkosi.conf.d/fedora/mkosi.conf.d/x86_64-uki-signed.conf
@@ -1,0 +1,13 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+[Match]
+Architecture=x86-64
+Bootloader=uki-signed
+
+[Content]
+# kernel-uki-virt ships a pre-built UKI for virtual machines, needed to test
+# Bootloader=uki-signed. Not available on all Fedora architectures (e.g. ppc64le),
+# so kept in arch-specific configs and gated on Bootloader=uki-signed. A separate
+# file is needed to match both Architecture= and Bootloader= simultaneously.
+Packages=
+        kernel-uki-virt
--- a/mkosi/bootloader.py
+++ b/mkosi/bootloader.py
@@ -25,6 +25,7 @@ from mkosi.config import (
     SecureBootSignTool,
     ShimBootloader,
     systemd_tool_version,
+    want_prebuilt_uki,
 )
 from mkosi.context import Context
 from mkosi.distribution import Distribution
@@ -680,15 +681,21 @@ def gen_kernel_images(context: Context) -> Iterator[tuple[str, Path]]:
         # scripts in the kernel source tree sometimes do weird stuff. But let's make sure we're not returning
         # UKIs as the UKI on Fedora is named vmlinuz-virt.efi. Also look for uncompressed images (vmlinux) as
         # some architectures ship those. Prefer vmlinuz if both are present.
-        for kimg in kver.glob("vmlinuz*"):
-            if KernelType.identify(context.config, kimg) != KernelType.uki:
-                yield kver.name, kimg
-                break
+        if want_prebuilt_uki(context.config):
+            for kimg in kver.glob("vmlinuz*.efi"):
+                if KernelType.identify(context.config, kimg) == KernelType.uki:
+                    yield kver.name, kimg
+                    break
         else:
-            for kimg in kver.glob("vmlinux*"):
+            for kimg in kver.glob("vmlinuz*"):
                 if KernelType.identify(context.config, kimg) != KernelType.uki:
                     yield kver.name, kimg
                     break
+            else:
+                for kimg in kver.glob("vmlinux*"):
+                    if KernelType.identify(context.config, kimg) != KernelType.uki:
+                        yield kver.name, kimg
+                        break
 
 
 def install_systemd_boot(context: Context) -> None:
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -3299,6 +3299,7 @@ SETTINGS: list[ConfigSetting[Any]] = [
         dest="bootloader",
         section="Content",
         parse=config_make_enum_parser(Bootloader),
+        match=config_make_enum_matcher(Bootloader),
         choices=Bootloader.choices(),
         default=Bootloader.systemd_boot,
         help="Specify which UEFI bootloader to use",
@@ -5443,6 +5444,15 @@ def want_kernel(config: Config) -> bool:
 
 def want_default_initrd(config: Config) -> bool:
     return Path("default") in config.initrds
+
+
+def want_prebuilt_uki(config: Config) -> bool:
+    # Returns True when mkosi should use a distro-pre-built signed UKI rather than building one itself.
+    # This happens when a signed bootloader is selected (implying distro UKIs) or when
+    # UnifiedKernelImages=signed is set explicitly.
+    return (
+        config.bootloader.is_signed() and config.unified_kernel_images == UnifiedKernelImage.auto
+    ) or config.unified_kernel_images == UnifiedKernelImage.signed
 
 
 def finalize_historydir(args: Args, output_dir: Optional[Path] = None) -> Path:


src-up:
Tested on Fedora 43 with kernel-uki-virt and ShimBootloader=none — boots successfully via UEFI → EFI/BOOT/BOOTX64.EFI (UKI).

Shim integration (ShimBootloader=signed) is not yet working and will be addressed in a follow-up.

MoltenArmor:
This is a bit confusing. Looking at `install_uki()`, doesn't mkosi already copy a prebuilt UKI from `/usr/lib/modules/<kver>/` directly when `Bootloader=uki-signed` is set with `UnifiedKernelImages=auto`, or when `UnifiedKernelImages=signed` is set explicitly?

```
    if (
        context.config.bootloader.is_signed()
        and context.config.unified_kernel_images == UnifiedKernelImage.auto
    ) or context.config.unified_kernel_images == UnifiedKernelImage.signed:
        for p in (context.root / "usr/lib/modules" / kver).glob("*.efi"):
            log_step(f"Installing prebuilt UKI at {p} to {boot_binary}")
            copyfile2(p, boot_binary)
            break
```

src-up:
> Looking at install_uki(), doesn't mkosi already copy a prebuilt UKI from /usr/lib/modules/<kver>/ directly when Bootloader=uki-signed is set with UnifiedKernelImages=auto, or when UnifiedKernelImages=signed is set explicitly?

i think , currently with Bootloader=uki-signed , stub check fails...

build_image -> install_kernel -> want_uki = True -> stub check fires → DIES: `systemd-stub not found`
              
systemd-stub is used to build UKIs from scratch. so i think the uki-signed assumes systemd-stub would be present, which is not needed for prebuilt UKIs.

Even if we add systemd-stub to list of packages, we get 
`A bootable image was requested but no kernel was found`

I believe this is because, install_uki is only called if gen_kernel_images returns something and assumption there is that both a raw vmlinuz AND a signed .efi alongside it in /usr/lib/modules/<kver>/ will exist. Since they dont exist for distros like fedora, gen_kernel_images returns empty and code never reaches install_uki..

- install_uki() is only called from the gen_kernel_images() loop, which looks for vmlinuz* that is not a UKI
- So the "find distro-signed UKI" code inside install_uki() is designed for distros that ship both a raw vmlinuz AND a signed .efi alongside it in /usr/lib/modules/<kver>/
- kernel-uki-virt (fedora prebuilt UKI) only ships the .efi — no separate vmlinuz — so gen_kernel_images() yields nothing and that path is never reached

uki-prebuilt does not make this assumption..

mkosi/__init__.py
```diff
@@ -2247,6 +2247,26 @@ def install_kernel(context: Context, partitions: Sequence[Partition]) -> None:
...
+        # With shim, place the UKI at shim's second-stage slot (e.g. grubx64.EFI) so the
+        # boot chain is: shim (BOOTX64.EFI) → UKI (grubx64.EFI). Without shim, place it
+        # at the UEFI fallback path (BOOTX64.EFI) for direct firmware boot.
+        if context.config.shim_bootloader != ShimBootloader.none:
+            boot_binary = context.root / shim_second_stage_binary(context)
+        else:
+            boot_binary = context.root / efi_boot_binary(context)
+        with umask(~0o700):
+            boot_binary.parent.mkdir(parents=True, exist_ok=True)
+        for kver, kimg in gen_kernel_images(context):
```
```suggestion
        for _, kimg in gen_kernel_images(context):
```
this will appease ruff and since it's not used, it doesn't need a name

behrmann: Nit: give this some space for readability

```suggestion
        if context.config.shim_bootloader != ShimBootloader.none:
            boot_binary = context.root / shim_second_stage_binary(context)
        else:
            boot_binary = context.root / efi_boot_binary(context)

        with umask(~0o700):
            boot_binary.parent.mkdir(parents=True, exist_ok=True)

        for kver, kimg in gen_kernel_images(context):
```

mkosi/__init__.py
```diff
@@ -2247,6 +2247,26 @@ def install_kernel(context: Context, partitions: Sequence[Partition]) -> None:
...
+            log_step(f"Installing prebuilt distro UKI {kimg} to {boot_binary}")
+            copyfile2(kimg, boot_binary)
+            break
+        else:
+            if context.config.bootable == ConfigFeature.enabled:
+                die("Bootloader=uki-prebuilt was set but no pre-built UKI was found in /usr/lib/modules")
+        return
```
behrmann: Less keyword soup
```suggestion
            return

        if context.config.bootable == ConfigFeature.enabled:
            die("Bootloader=uki-prebuilt was set but no pre-built UKI was found in /usr/lib/modules")
```

mkosi/resources/man/mkosi.1.md
```diff
@@ -1051,6 +1051,12 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
     The `signed` variants will only install pre-signed EFI binaries
     shipped by the distribution.
 
+    If set to `uki-prebuilt`, **mkosi** copies a unified kernel image already
+    shipped by the distribution from `/usr/lib/modules/<version>/` (as `*.efi`)
+    to `EFI/BOOT/BOOT*.EFI` in the ESP and does not invoke `ukify` or require
+    systemd-stub. The highest kernel version is chosen; the first PE file in that
+    directory that identifies as a UKI is used.
```
behrmann: Nit: the PE file doesn't identify as anything, we identify it as UKI
```suggestion
    directory that is identified as a UKI is used.
```

behrmann (commented):
If Fedora ships a prebuilt UKI, then this should be covered by the tests as well and an image with a prebuilt UKI should be built.

src-up:
@behrmann Thanks for the review. 

Tried two approaches for test coverage:

- Injecting --package=kernel-uki-virt at test time — fails because --incremental=strict rejects the cache manifest mismatch.
- Adding kernel-uki-virt to mkosi.conf.d/fedora/mkosi.conf — fails with `No match for argument: kernel-uki-virt` on rawhide


Open to suggestions on how to handle this. 

Rest of the comments were addressed.

src-up:
@behrmann 
moved kernel-uki-virt to the arch-specific x86_64.conf and arm64.conf configs (where the package exists), and excluded ppc64le from the test since it doesn't support SMBIOS anyway.

One other thing needed: the CI test framework injects params like systemd.unit= and systemd.firstboot=no via kernel_command_line at build time — but for a prebuilt UKI those can't be baked in. Fixed by passing them at runtime via SMBIOS type-11 (io.systemd.stub.kernel-cmdline-extra), which systemd-stub reads from the prebuilt UKI.

Please take a look when you get a chance.

src-up:
@daandemeyer any thoughts on this ?

src-up:
@behrmann please review

behrmann:
> @behrmann please review

On vacation currently, will review sometime in the next few days.

src-up:
> On vacation currently, will review sometime in the next few days.

@behrmann no worries. I have rebased from latest main

For the new uki-prebuilt path, the sequence is:

- mkosi/bootloader.py: gen_kernel_images() switches to gen_prebuilt_uki_images() when uki-prebuilt is selected, and that path picks a UKI .efi from /usr/lib/modules/<kver>/.

- mkosi/__init__.py (install_kernel): mkosi copies that selected UKI to the EFI boot target (boot_binary, i.e. shim second stage or fallback EFI path) via copyfile2(...).

CI log matches this flow, with the raw line: Installing prebuilt distro UKI .../usr/lib/modules/.../vmlinuz-virt.efi to .../efi/EFI/BOOT/grubx64.EFI.

Relevant CI job: integration-test (fedora, fedora, ubuntu-24.04)

mkosi/qemu.py
behrmann, line 1: The alternative would be to provide them a profile and that could be mentioned as a TODO, but I think implementing this is out of scope for this PR.

behrmann (approved):

src-up:
@behrmann can we get these merged ?

behrmann:
@src-up Oh, geez, I thought I had already. I was talking to Vitaly (?) about this two weeks ago. I'm terribly sorry. 

I've just went through the logic again and notice I had misunderstood @MoltenArmor's initial comment about [the codepath for prebuilt UKIs](https://github.com/systemd/mkosi/blob/b9d4c0623975d75788627c114147be394939c7cb/mkosi/__init__.py#L2152). 

Having gone through the code again, your analysis why that code isn't used is correct (`install_uki` not being called if `gen_kernel_images` doesn't return anything). Having looked at this for the past hour, I don't think this ever worked, because the the logic to not pick up EFI binaries in `gen_kernel_images` predates (2023, in #2166) the logic of installing those files in `install_uki` (b68a3ff10fb95c835c541897fc807e312e7351bc in 2024).

I think this should work if `gen_kernel_images` tried to look for UKIs in `gen_kernel_images` when `uki-signed` is used. The addition of `gen_prebuilt_uki_images` and in `install_kernel` should then not be necessary. The added test is still very valuable, since, as I think, if this never worked, this is due to it not being tested.

I would be grateful if you could have a look whether the above idea makes this work. Sorry for bungling up the communication on this so much.

src-up:
> @src-up Oh, geez, I thought I had already. I was talking to Vitaly (?) about this two weeks ago. I'm terribly sorry.
> 
Yes, Vitaly mentioned he met and spoke to you @behrmann 

> I would be grateful if you could have a look whether the above idea makes this work. Sorry for bungling up the communication on this so much.

No worries :=) . Thanks for reviewing this again. Much appreciated. 

I tried putting the changes together like you suggested and came up with this commit https://github.com/src-up/mkosi/commit/2d3c99b3d657f94b2ab405b5536bc65217d5adc8. Let me know if this is more in line with what you were thinking. 

I tested the below commands on a fedora box and they pass
```
bin/mkosi -f box -- true
tools/integration-test-setup.sh fedora fedora
bin/mkosi box -- pytest -m integration --distribution fedora --capture=no --verbose     'tests/test_boot.py::test_bootloader[uki-signed]'
```

behrmann:
@src-up Looks good, but I have to think this through more fully, I'll come back to you about this tomorrow.

behrmann:
@src-up Sorry, late again. I left you a few commits on the commit. Thank you!

src-up:
@behrmann hey, so I have moved that commit https://github.com/src-up/mkosi/commit/2d3c99b3d657f94b2ab405b5536bc65217d5adc8 as part of this PR. Let me know what u think of the changes now ...

- **Guard**: replaced is_signed() with a want_prebuilt_uki() helper. Also added match=config_make_enum_matcher(Bootloader) to make Bootloader= usable in [Match] sections.
- **Glob**: narrowed vmlinuz* → vmlinuz*.efi in the UKI-yielding branch.
- **kernel-uki-virt**: Since multiple [Match] sections in the same file are ANDed, and we have packages that are needed for other Bootloader modes, moved to new x86_64-uki-signed.conf / arm64-uki-signed.conf with [Match] Bootloader=uki-signed.

mkosi/qemu.py
```diff
@@ -1314,9 +1315,24 @@ def add_virtiofs_mount(
...
+            # these cannot be baked in at build time.
+            # Chosen solution: inject them at runtime via SMBIOS type-11, which
+            # systemd-stub reads. stub_kcl includes config.kernel_command_line for
+            # uki-signed so it gets passed to systemd-stub via this mechanism.
+            # We can safely assume stub_kcl will be invoked for all uki-signed use
+            # cases because: (1) uki-signed always boots via UEFI so supports_smbios()
+            # is always true for it, and (2) systemd-stub is guaranteed to be present
+            # in the uki-signed UKI since .sdmagic — a PE section that systemd-stub
+            # adds when wrapping a kernel into a UKI — is required by mkosi to qualify
+            # the .efi as a UKI and therefore be copied into the ESP as a uki-signed image.
```
keszybz: Was this generated by AI?

Maybe something like this:
> With Bootloader=uki-signed, a UKI built by the distro is used, and we cannot embed config.kernel_command_line in the UKI.
> Instead, pass those options through SMBIOS type#11.
> We know that this will work because all UEFI systems support SMBIOS and
> UKIs by construction use systemd-stub, which reads SMBIOS type#11.

keszybz (commented):
Looks reasonable, but the commit has not commit message. Please add a proper commit message.

mkosi/qemu.py
src-up: > Was this generated by AI?

:-) I think I had divided my notes into Need, Obstacle and Chosen because i spent significant amount of time figuring the SMBIOS path and justifying it to myself that we need it. I wanted to remember next time what i did.

But yeah too long. I have reduced it as per your suggestion.

src-up:
@keszybz thanks for the review. I added the commit message and updated the comment. Let me know if this can be merged now.

mkosi/qemu.py
```diff
@@ -1314,9 +1315,14 @@ def add_virtiofs_mount(
         if kernel and (kerneltype != KernelType.uki or not config.architecture.supports_smbios(firmware)):
             cmdline += ["-append", " ".join(config.kernel_command_line + kcl)]
         elif config.architecture.supports_smbios(firmware):
+            # With Bootloader=uki-signed, a UKI built by the distro is used, and we cannot embed
+            # config.kernel_command_line in the UKI. Instead, pass those options through SMBIOS type#11. We
+            # know that this will work because all UEFI systems support SMBIOS and UKIs by construction use
+            # systemd-stub, which reads SMBIOS type#11.
+            stub_kcl = (config.kernel_command_line if config.bootloader == Bootloader.uki_signed else []) + kcl
```
behrmann: This line still needs to be formatted with ruff

tests/test_boot.py
```diff
@@ -59,7 +59,21 @@ def test_format(config: ImageConfig, format: OutputFormat) -> None:
 
 @pytest.mark.parametrize("bootloader", Bootloader)
 def test_bootloader(config: ImageConfig, bootloader: Bootloader) -> None:
-    if config.distribution == Distribution.rhel_ubi or bootloader.is_signed():
+    if (config.distribution == Distribution.rhel_ubi or
+        bootloader.is_signed() and bootloader != Bootloader.uki_signed):
+        return
```
behrmann: This also still needs to be formatted, also, `and` does bind stronger than `or`, but it would be more obvious with parentheses

```suggestion
    if (
        config.distribution == Distribution.rhel_ubi or
        (bootloader.is_signed() and bootloader != Bootloader.uki_signed)
    ):
        return
```

behrmann (commented):
Almost done

tests/test_boot.py
src-up: Thanks. Done.

mkosi/qemu.py
src-up: Oh i thought i had take care of this. Doing it now.

tests/test_boot.py
```diff
@@ -59,7 +59,23 @@ def test_format(config: ImageConfig, format: OutputFormat) -> None:
 
 @pytest.mark.parametrize("bootloader", Bootloader)
 def test_bootloader(config: ImageConfig, bootloader: Bootloader) -> None:
-    if config.distribution == Distribution.rhel_ubi or bootloader.is_signed():
+    if (
+        config.distribution == Distribution.rhel_ubi
+        or (bootloader.is_signed() and bootloader != Bootloader.uki_signed)
+    ):
```
behrmann: Shouldn't have blindly guessed what it would look like
```suggestion
    if config.distribution == Distribution.rhel_ubi or (
        bootloader.is_signed() and bootloader != Bootloader.uki_signed)
    ):
```
This is what CI is complaining about.

src-up:
@behrmann Thanks for mentioning about CI and your patience on this PR. I should have checked that myself first. 

Post the ruff fix, the fedora CI tests were also failing due to cache manifest not matching the saved one. 

With Packages=kernel-uki-virt, the package appears in the cache manifest. Since the CI cache was built without it, the manifests differ and --incremental=strict aborts the build. I changed it to VolatilePackages= so that kernel-uki-virt is installed on top of the cache without affecting the manifest. 

The fedora/fedora CI test passes now, but one of them (opensuse) ais still failing due to openSUSE GPG key verification error - perhaps unrelated changes.

behrmann:
> @behrmann Thanks for mentioning about CI and your patience on this PR. I should have checked that myself first.

No worries. Your work on this has been very diligent, thank you.

> The fedora/fedora CI test passes now, but one of them (opensuse) ais still failing due to openSUSE GPG key verification error - perhaps unrelated changes.

That is unrelated. The key is not yet updated in a release version of distribution-gpg-keys 

> Post the ruff fix, the fedora CI tests were also failing due to cache manifest not matching the saved one.
>
> With Packages=kernel-uki-virt, the package appears in the cache manifest. Since the CI cache was built without it, the manifests differ and --incremental=strict aborts the build. I changed it to VolatilePackages= so that kernel-uki-virt is installed on top of the cache without affecting the manifest.

The workaround seems like a hack. Maybe for this particular piece of the matrix `--incremental=yes` or `--incremental=no` can be set—we do have `--incremental=no` for the extension image tests. You can try that in CI or I'll try to test this locally once I find time. It's just my first idea, I might be wrong.

src-up:
Hey @behrmann With "incremental=no", it is showing 7 failed tests and i think "default-initrd.cache does not exist" in logs means the cache is getting removed.

Here is [link to the raw log output ](https://productionresultssa0.blob.core.windows.net/actions-results/0de1e242-a9e0-4889-8a0e-ad887ab89044/workflow-job-run-cdb4750a-162a-53f3-97ef-ad984165a2f9/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-07-21T19%3A05%3A51Z&sig=O1vU8GiwjsnLbslb58xzWslY28zkUnnFDVz5WFHSolk%3D&ske=2026-07-21T21%3A24%3A00Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-07-21T17%3A24%3A00Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-07-21T18%3A55%3A46Z&sv=2025-11-05)of one of the 7 failing CI checks, specifically these lines 

```
2026-07-20T19:50:51.6024838Z ‣ /home/runner/work/mkosi/mkosi/mkosi.cache/fedora~rawhide~x86-64~default-initrd.cache does not exist, not reusing cached images
2026-07-20T19:50:51.6026197Z ‣ Strict incremental mode is enabled and cache for image default-initrd is out-of-date
2026-07-20T19:50:51.6026874Z ‣ (Build once with '-i yes' to update the image cache)
2026-07-20T19:50:51.6246787Z Traceback (most recent call last):
2026-07-20T19:50:51.6253892Z   File "/home/runner/work/mkosi/mkosi/mkosi/run.py", line 104, in uncaught_exception_handler
2026-07-20T19:50:51.6254536Z     yield
2026-07-20T19:50:51.6254913Z   File "/usr/lib64/python3.15/contextlib.py", line 116, in inner
2026-07-20T19:50:51.6255582Z     return func(*args, **kwds)
2026-07-20T19:50:51.6256086Z   File "/home/runner/work/mkosi/mkosi/mkosi/__main__.py", line 44, in main
2026-07-20T19:50:51.6256675Z     run_verb(args, tools, images, resources=resources)
2026-07-20T19:50:51.6257204Z     ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2026-07-20T19:50:51.6257764Z   File "/home/runner/work/mkosi/mkosi/mkosi/__init__.py", line 5251, in run_verb
2026-07-20T19:50:51.6258302Z     die(
2026-07-20T19:50:51.6258540Z     ~~~^
2026-07-20T19:50:51.6259100Z         f"Strict incremental mode is enabled and cache for image {c.image} is out-of-date",
2026-07-20T19:50:51.6259767Z         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2026-07-20T19:50:51.6260305Z         hint="Build once with '-i yes' to update the image cache",
2026-07-20T19:50:51.6260790Z         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2026-07-20T19:50:51.6261164Z     )
2026-07-20T19:50:51.6261399Z     ^
2026-07-20T19:50:51.6261767Z   File "/home/runner/work/mkosi/mkosi/mkosi/log.py", line 35, in die
2026-07-20T19:50:51.6262608Z     sys.exit(1)
2026-07-20T19:50:51.6262877Z     ~~~~~~~~^^^
2026-07-20T19:50:51.6263138Z SystemExit: 1
```

I can try "incremental=yes" also but wanted to show you these first.

src-up:
@behrmann I tried with both incremental=yes and incremental=no. 

Where things stand 

1. **incremental=strict CI** (default used by all tests) -  test_bootloader[uki-signed] fails. Cache is built without kernel-uki-virt, but when test_bootloader[uki-signed] runs with Packages=kernel-uki-virt , the manifest mismatches and build aborted. 
2. **incremental=no** test_bootloader[uki-signed] passes, but both default-initrd and main caches are unconditionally deleted (no manifest check at all). The 10 subsequent tests fail with "does not exist". Deletion of cache happens because the way code is written, if force flag is on (which is for all tests), cache is deleted if cache is ignored which is always true for incremental=no. 
3. **incremental=yes**  test_bootloader[uki-signed] passes, but rebuilds and saves both caches with kernel-uki-virt. The 10 subsequent tests fail with "Cache manifest mismatch" — the saved cache now has kernel-uki-virt but those tests build without Bootloader=uki-signed.
4. **VolatilePackages=** works but i do see ur point, its a hack.


Honestly, not sure how to proceed here.

behrmann:
Thanks for trying this out. I have a couple of deadlines I need to handle, so I won't be looking at this seriously before the week after next (the first of August that is).

src-up:
@behrmann Hey , Let me know if this week is a good time to figure out what path to take forward. Thanks !

behrmann:
I just opened #4410 to look into this.

behrmann:
@src-up after playing around with this for a bit, I think the most concise solution is to set a different cache key for this particular test. You can find this as the last commit on #4410. The changes before that are unnecessary, I only factored out that particular instance from the test initially, because I was looking at the ordering of the tests.

src-up:
Thanks @behrmann for working on this. I got distracted so working on this now. So i should test this with that single last commit ?

behrmann:
No worries! Please do. I think it should be enough to add `"--cache-key=&d~&r~&a~&I~prebuilt",` (or similar) for the signed UKI case.

src-up:
@behrmann it seems all checks passed with
`["--incremental=no", "--cache-key=&d~&r~&a~&I~prebuilt"]`

src-up:
@behrmann i believe this can be merged now. Please let me know if i am missing something here. Thanks!

tests/test_boot.py
```diff
@@ -59,11 +59,37 @@ def test_format(config: ImageConfig, format: OutputFormat) -> None:
...
 
     with Image(config) as image:
-        image.build(["--format=disk", "--bootloader", str(bootloader)])
+        image.build(
+            [
+                "--format=disk",
+                "--bootloader",
+                str(bootloader),
+                *(
+                    ["--incremental=no", "--cache-key=&d~&r~&a~&I~prebuilt"]
```
behrmann, line 89: Last question from my side: Doesn't this work without `--incremental=no`? Since the new cache key separated the cache it should work to have a separate cache for the prebuilt ones, no?

behrmann (approved):
LGTM. There's a few smaller followups to do (`Bootloader=` match should be documented, some formatting), but I don't think this necessarily belongs here.

tests/test_boot.py
src-up: I tried with only the cache-key prior to this. I believe the tests use --incremental=strict, and the new ~prebuilt cache does not exist yet, so the build fails.

src-up:
> LGTM. There's a few smaller followups to do (`Bootloader=` match should be documented, some formatting), but I don't think this necessarily belongs here.

@behrmann feel free to list them here for posterity. I can address them in a follow up PR soon.

behrmann:
Talked to Daan out of band and he said it looks reasonable as well, so finally merging this. Thanks a lot for your patience in seeing this through, @src-up.

src-up:
> Talked to Daan out of band and he said it looks reasonable as well, so finally merging this. Thanks a lot for your patience in seeing this through, @src-up.

Many thanks.

PR #4250, opened 2026-03-29
docs: Add example for installing AUR packages in ArchLinux images

### LINKED ISSUE #2578: Example on building Arch based image with packages from AUR?
Would be nice to have some example on how to build Arch based image that have packages from AUR (when host that builds image is also Arch linux and of same platform, in case it matters).

estebanpuyanas:
# Add example for installing AUR packages in Arch-based images

Adds a documentation file under `mkosi/docs`explaining how to install AUR packages at image build time using `aurutils` and `mkosi`'s `$PACKAGEDIR` mechanism.

The example uses `neovim-git` as a worked case because it has both official-repo `makedepends` and AUR dependencies, which exercises the full dependency resolution pipeline.

The approach covers:

1. Bootstrapping `aurutils` from the AUR at build time.
2. Using `aur depends` + `tsort` for dependency resolution.
3. Building packages as the `nobody` user to satisfy `makepkg`'s refusal to run as root.
4. Using `$PACKAGEDIR` to pass built packages from the build phase to `mkosi.postinst.chroot`.
5. Configuring mirrors via `reflector` since the build sandbox starts with none.

Tested end-to-end on Arch Linux (EndeavourOS) with QEMU. The image builds successfully and `nvim --version` confirms correct installation after booting.

Closes #2578

### DIFF
--- a/docs/building-aur-packages.md
+++ b/docs/building-aur-packages.md
@@ -1,0 +1,264 @@
+---
+title: Installing AUR Packages in Arch Linux Images
+category: Tutorials
+layout: default
+SPDX-License-Identifier: LGPL-2.1-or-later
+---
+
+# Installing AUR Packages in Arch Linux Images:
+
+`mkosi` does not natively support installing packages from the [Arch User Repository
+(AUR)](https://aur.archlinux.org/) since AUR packages must be built from source using `mkpkg` rather than
+installed through `pacman`. This section describes how to install AUR packages at image build time using
+[`aurutils`](https://aur.archlinux.org/packages/aurutils) and a local pacman repository.
+
+## Overview:
+
+The approach involes using three `mkosi` hook scripts that run in sequence:
+
+1. `mkosi.prepare.chroot`: Initialises the pacman keyring, uses `reflector` to fetch and rank mirrors, bootstraps `aurutils` from the AUR (since it is an AUR package itself and therefore cannot be listed in
+   the `BuildPackages=` section of an `mkosi.conf` file), and then sets up a local pacman repository backed by
+   `$PACKAGEDIR`.
+
+2. `mkosi.build.chroot`: Resolves the full AUR dependency graph, installs official-repo makedepends, fetches `PKGBUILD`s, builds packages as the unprivileged [nobody](https://wiki.ubuntu.com/nobody) user, and registers the results in the local repository inside `$PACKAGEDIR`.
+
+3. `mkosi.postinst.chroot`: Configures pacman in the final image to read from `$PACKAGEDIR` and installs all built AUR packages.
+
+`BuildSourcesEphemeral=yes` ensures that build tools (`base-devel`, `git`, etc.) installed via
+`BuildPackages=` are present during the build phase but [are not included in the final image](/mkosi/resources/man/mkosi.1.md#build-section). `$PACKAGEDIR` is a mkosi-managed directory that persists between the build phase and the postinst phase.
+
+### Why `nobody`?
+
+Given that `makepkg` refuses to run as root as a safety measure, inside mkosi's build sandbox there is no persistent user database, so the existing `nobody` system user (`uid 65534`) is a convenient unprivileged identity that is always available.
+
+### Why `aur depends` and `tsort`?
+
+AUR packages can depend on other AUR packages. With `aur depends -r` emits the full
+transitive dependency graph. `tsort` linearises the dependencies into a build order that guarantees each package is built after the packages it depends on. For neovim-git, the AUR package used in this example, this means any AUR dependencies are built and registered before neovim-git itself attempts to build against them.
+
+# Example installing `neovim-git`:
+
+## `mkosi.conf`:
+
+```sh
+[Distribution]
+Distribution=arch
+
+[Build]
+BuildSourcesEphemeral=yes
+
+# Enables network connectivity while mkosi.build build scripts are invoked.
+# Required to fetch AUR sources and dependencies.
+WithNetwork=yes
+
+[Content]
+# Build-time only tools.
+BuildPackages=
+    base-devel
+    git
+    jq
+    reflector
+
+# AUR packages are installed via mkosi.postinst.chroot.
+Packages=
+    base
+    linux
+    linux-firmware
+    systemd
+    bash
+    reflector
+```
+
+## `mkosi.prepare.chroot`:
+
+```sh
+#!/usr/bin/env bash
+
+set -eux
+
+if [[ "${1}" != "build" ]]; then
+    exit 0
+fi
+
+# Pacman initialization operations:
+pacman-key --init
+pacman-key --populate archlinux
+
+reflector --latest 10 --sort rate --save /etc/pacman.d/mirrorlist
+
+pacman -Sy
+
+# Bootstrap aurutils from AUR, with user nobody into /tmp, then install with pacman.
+install -d /tmp/aurutils-build -o nobody
+
+cd /tmp/aurutils-build
+sudo -u nobody git clone https://aur.archlinux.org/aurutils.git
+cd aurutils
+
+( source ./PKGBUILD
+  pacman -S --needed --noconfirm \
+      "${makedepends[@]}" "${depends[@]}"
+)
+
+sudo -u nobody PKGDEST="/tmp" PKGEXT=".pkg.tar" makepkg --clean --cleanbuild --noconfirm
+pacman -U --noconfirm /tmp/aurutils-*.pkg.tar
+
+cd /
+
+# Create the local repo directory owned by nobody.
+install -d /aur -o nobody
+
+# Initialise an empty pacman database for the [aur] repo.
+sudo -u nobody tar -ca -f /aur/aur.db.tar.xz -T /dev/null
+sudo -u nobody ln -sf /aur/aur.db.tar.xz /aur/aur.db
+
+mkdir -p /etc/pacman.d
+
+# Sentinel file so /etc/pacman.d is never empty (pacman requires Include targets exist).
+cat > /etc/pacman.d/dummy.conf << 'EOF'
+# This file exists so /etc/pacman.d is never an empty directory.
+EOF
+
+# Drop the [aur] repo fragment.
+cat > /etc/pacman.d/aur.conf << 'EOF'
+[aur]
+SigLevel = Optional TrustAll
+Server = file:///aur
+EOF
+
+# Add the Include directive to pacman.conf if not already present.
+if ! grep -q 'Include = /etc/pacman.d/\*.conf' /etc/pacman.conf; then
+    echo -e '\nInclude = /etc/pacman.d/*.conf' >> /etc/pacman.conf
+fi
+
+# Sync so pacman knows about [aur] before mkosi.build.chroot runs.
+pacman -Sy
+```
+
+## `mkosi.build.chroot`:
+
+```sh
+# mkosi.build.chroot
+#
+# This runs after mkosi.prepare.chroot "build" phase.
+# Still inside the BUILD overlay, so base-devel, aurutils, git, are all available here.
+
+set -eux
+
+PACKAGES=(
+    neovim-git
+    # foo
+    # bar
+    # and any other AUR packages you want to build here
+)
+
+# Set up the local pacman repo inside $PACKAGEDIR (not /aur) to pass
+# built packages from the build script to postinst.
+
+install -d "$PACKAGEDIR" -o nobody
+
+sudo -u nobody tar -ca -f "$PACKAGEDIR/aur.db.tar.xz" -T /dev/null
+sudo -u nobody ln -sf "$PACKAGEDIR/aur.db.tar.xz" "$PACKAGEDIR/aur.db"
+
+# Point pacman at $PACKAGEDIR as the [aur] repo.
+mkdir -p /etc/pacman.d
+cat > /etc/pacman.d/aur.conf << EOF
+[aur]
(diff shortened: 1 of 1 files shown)


docs/building-aur-packages.md
```diff
@@ -0,0 +1,264 @@
...
+title: Installing AUR Packages in Arch Linux Images
+category: Tutorials
+layout: default
+SPDX-License-Identifier: LGPL-2.1-or-later
+---
+
+# Installing AUR Packages in Arch Linux Images:
+
+`mkosi` does not natively support installing packages from the [Arch User Repository
+(AUR)](https://aur.archlinux.org/) since AUR packages must be built from source using `mkpkg` rather than
```
behrmann:
```suggestion
(AUR)](https://aur.archlinux.org/) since AUR packages must be built from source using `makepkg` rather than
```

docs/building-aur-packages.md
```diff
@@ -0,0 +1,264 @@
...
+    jq
+    reflector
+
+# AUR packages are installed via mkosi.postinst.chroot.
+Packages=
+    base
+    linux
+    linux-firmware
+    systemd
+    bash
```
behrmann: These two are dependencies of `base`

docs/building-aur-packages.md
```diff
@@ -0,0 +1,264 @@
...
+2. `mkosi.build.chroot`: Resolves the full AUR dependency graph, installs official-repo makedepends, fetches `PKGBUILD`s, builds packages as the unprivileged [nobody](https://wiki.ubuntu.com/nobody) user, and registers the results in the local repository inside `$PACKAGEDIR`.
+
+3. `mkosi.postinst.chroot`: Configures pacman in the final image to read from `$PACKAGEDIR` and installs all built AUR packages.
+
+`BuildSourcesEphemeral=yes` ensures that build tools (`base-devel`, `git`, etc.) installed via
+`BuildPackages=` are present during the build phase but [are not included in the final image](/mkosi/resources/man/mkosi.1.md#build-section). `$PACKAGEDIR` is a mkosi-managed directory that persists between the build phase and the postinst phase.
+
+### Why `nobody`?
+
+Given that `makepkg` refuses to run as root as a safety measure, inside mkosi's build sandbox there is no persistent user database, so the existing `nobody` system user (`uid 65534`) is a convenient unprivileged identity that is always available.
```
behrmann: A very thin safety measure, you can also set `EUID` to something else.

docs/building-aur-packages.md
```diff
@@ -0,0 +1,264 @@
...
+
+The approach involves using three `mkosi` hook scripts that run in sequence:
+
+1. `mkosi.prepare.chroot`: Initialises the pacman keyring, uses `reflector` to fetch and rank mirrors, bootstraps `aurutils` from the AUR (since it is an AUR package itself and therefore cannot be listed in
+   the `BuildPackages=` section of an `mkosi.conf` file), and then sets up a local pacman repository backed by
+   `$PACKAGEDIR`.
+
+2. `mkosi.build.chroot`: Resolves the full AUR dependency graph, installs official-repo makedepends, fetches `PKGBUILD`s, builds packages as the unprivileged [nobody](https://wiki.ubuntu.com/nobody) user, and registers the results in the local repository inside `$PACKAGEDIR`.
+
+3. `mkosi.postinst.chroot`: Configures pacman in the final image to read from `$PACKAGEDIR` and installs all built AUR packages.
```
behrmann: The usage of `.chroot` scripts implies that all of this is done in the image being built and that the image should be updated later. That is not something mkosi necessarily guarantees is possible. If you built this inside an Arch tools tree, you could also piggyback off the setup that mkosi already does.

docs/building-aur-packages.md
```diff
@@ -0,0 +1,264 @@
...
+    reflector
+
+# AUR packages are installed via mkosi.postinst.chroot.
+Packages=
+    base
+    linux
+    linux-firmware
+    systemd
+    bash
+    reflector
```
behrmann: Why not just use the fastly mirror that mkosi defaults to?

docs/building-aur-packages.md
```diff
@@ -0,0 +1,264 @@
...
+install -d /tmp/aurutils-build -o nobody
+
+cd /tmp/aurutils-build
+sudo -u nobody git clone https://aur.archlinux.org/aurutils.git
+cd aurutils
+
+( source ./PKGBUILD
+  pacman -S --needed --noconfirm \
+      "${makedepends[@]}" "${depends[@]}"
+)
```
behrmann: Why is this necessary? Why not use `--syncdeps` from `makepkg`?

behrmann (commented):
Thanks! This is an interesting writeup, though I'm not entirely sure the upstream docs are the best place for it in this form, since advocating for any AUR helper is really mkosi's place. 

A few thoughts I had while skimming it.

docs/building-aur-packages.md
estebanpuyanas: Fair point. I ran into mirror timeout issues during local testing (likely network-specific), and `reflector` resolved them. Rather than including it as a hard dependency, I'll remove it from the scripts and add a note saying that if you encounter mirror timeouts, adding `reflector` to `BuildPackages=` and calling it like it currently is in line 87,234 before `pacman -Sy` is a reliable workaround.

docs/building-aur-packages.md
estebanpuyanas: From my understanding, `--syncdeps` calls `pacman` via `sudo` internally, which won't work since `nobody` has no `sudo` privileges in the build sandbox. The intended fix would be to install deps as root first using `makepkg --nodeps --nobuild` to parse them, then install, then build. Alternatively, since `aurutils` itself has a well-known and stable dependency list, I can just hardcode the deps install and drop the `PKGBUILD` sourcing entirely. Which approach would you prefer?

As a sidenote, the discussion in #2578 has these two gists, which I largely used to base the work done here:
https://gist.github.com/michaelbeaumont/1b2b58d63ccdc446496f0156d6aca4ad
https://gist.github.com/Strykar/ffafe1fca0ce777a048ec44984d7aaef

docs/building-aur-packages.md
estebanpuyanas: Looking at the existing [building-rpms-from-source.md](https://github.com/systemd/mkosi/blob/main/docs/building-rpms-from-source.md) example, I can see it uses `mkosi-chroot` and `mkosi-install` from the host side rather than relying on `.chroot` scripts.

I'd be happy to rework the example to follow that pattern if that's the preferred direction. Alternatively I can add an explicit note scoping this example to Arch hosts building Arch images and flag the tools tree approach as a follow-up?

docs/building-aur-packages.md
estebanpuyanas: @behrmann, I tested that approach by replacing the highlighted lines to:

```sh
sudo -u nobody \
    PKGDEST="/tmp" \
    PKGEXT=".pkg.tar" \
    makepkg --clean --cleanbuild --noconfirm --syncdeps
```
However, it resulted in a failed build, as it does seem like `syncdeps` calls `pacman` internally via `sudo` as I said above. I am attaching a text file with the build output in case I missed anything else that I could try. [failed-syncdeps-build.txt](https://github.com/user-attachments/files/26364542/failed-syncdeps-build.txt)

docs/building-aur-packages.md
behrmann: Sticking with the style of the RPM docs doing it in the image being built is fine.

docs/building-aur-packages.md
```diff
@@ -0,0 +1,244 @@
...
+```
+
+## `mkosi.prepare.chroot`:
+
+```sh
+#!/usr/bin/env bash
+
+set -eux
+
+if [[ "${1}" != "build" ]]; then
```
behrmann: style: no unnecessary curly braces
```suggestion
if [[ "$1" != "build" ]]; then
```

docs/building-aur-packages.md
```diff
@@ -0,0 +1,244 @@
...
+    exit 0
+fi
+
+# Pacman initialization operations:
+pacman-key --init
+pacman-key --populate archlinux
```
behrmann: This shouldn't be necessary, since mkosi will have already installed the base system

docs/building-aur-packages.md
```diff
@@ -0,0 +1,244 @@
...
+
+cat > /etc/pacman.d/mirrorlist << 'EOF'
+Server = https://fastly.mirror.pkgbuild.com/$repo/os/$arch
+EOF
```
behrmann, line 92: Since mkosi has already installed things at this point, why change the mirror now?

docs/building-aur-packages.md
```diff
@@ -0,0 +1,244 @@
...
+
+pacman -Sy
```
behrmann, line 94: This should also not be necessary, since packages have just been installed.

docs/building-aur-packages.md
```diff
@@ -0,0 +1,244 @@
...
+sudo -u nobody tar -ca -f "$PACKAGEDIR/aur.db.tar.xz" -T /dev/null
+sudo -u nobody ln -sf "$PACKAGEDIR/aur.db.tar.xz" "$PACKAGEDIR/aur.db"
+
+mkdir -p /etc/pacman.d
+
+cat > /etc/pacman.d/aur.conf << EOF
+[aur]
+SigLevel = Optional TrustAll
+Server = file://$PACKAGEDIR
+EOF
```
behrmann: Style:
```suggestion
cat >/etc/pacman.d/aur.conf <<EOF
[aur]
SigLevel = Optional TrustAll
Server = file://$PACKAGEDIR
EOF
```

That being said, why install this in the build step of the prepare script, when you add it back in the postinst script? 

That being said, is there a reason to make a repo? `pacman -U` installs packages from paths just fine, so maybe this is a complexity that could be avoided.

docs/building-aur-packages.md
```diff
@@ -0,0 +1,244 @@
...
+    makepkg --clean --cleanbuild --noconfirm
+
+pacman -U --noconfirm /tmp/aurutils-*.pkg.tar
+
+cd /
+
+# Initialize the local pacman repo inside $PACKAGEDIR.
+install -d "$PACKAGEDIR" -o nobody
+sudo -u nobody tar -ca -f "$PACKAGEDIR/aur.db.tar.xz" -T /dev/null
+sudo -u nobody ln -sf "$PACKAGEDIR/aur.db.tar.xz" "$PACKAGEDIR/aur.db"
```
behrmann: Is there a reason not to use `repo-add` here? It comes with pacman.

docs/building-aur-packages.md
```diff
@@ -0,0 +1,244 @@
...
+# This runs after mkosi.prepare.chroot "build" phase.
+# Still inside the BUILD overlay, so base-devel, aurutils, git, are all available here.
+
+set -eux
+
+PACKAGES=(
+    neovim-git
+    # foo
+    # bar
+    # and any other AUR packages you want to build here
```
behrmann: Since your example is neovim, you can just stick with that

```suggestion
    neovim-git
```

docs/building-aur-packages.md
```diff
@@ -0,0 +1,244 @@
...
+
+pacman -S --needed --noconfirm \
+    git pacutils curl perl perl-json-xs bash
+
+sudo -u nobody \
+    PKGDEST="/tmp" \
+    PKGEXT=".pkg.tar" \
+    makepkg --clean --cleanbuild --noconfirm
+
+pacman -U --noconfirm /tmp/aurutils-*.pkg.tar
```
behrmann: Is there a deeper reason to build aurutils in the prepare script? It could also be built in the build script together with the rest. Alternatively, you could do all the building in the prepare script and would not require `WithNetwork=yes`

docs/building-aur-packages.md
estebanpuyanas: Without these lines, `pacman -S` fails with "no servers configured for repository" and "database file does not exist" errors. The `postinst` script has the same requirement. If `mkosi` already handles this in a way I am missing, I'd appreciate guidance on the correct approach.

I am also attaching logs of a failed build after removing these lines where the server failed to sync databases.

[pacman-sy-failed-build.txt](https://github.com/user-attachments/files/26660960/pacman-sy-failed-build.txt)

docs/building-aur-packages.md
estebanpuyanas: All building has been moved into the prepare script as you suggested, which removes the need for a separate build step. The `mkosi.build.chroot` is now a stub that exists only to trigger `mkosi`'s build phase so that `mkosi.prepare.chroot` is called with the build argument. `WithNetwork=yes` is still required since it controls network access for the prepare script as well.

docs/building-aur-packages.md
estebanpuyanas: This block has been removed in the latest version. The solution now uses `pacman -U $PACKAGEDIR/*.pkg.tar.zst` directly in the `postinst` script, avoiding the need for a local `pacman` repo entirely.

docs/building-aur-packages.md
estebanpuyanas: Removing these from either script causes `pacman` to fail with "Public keyring not found" and "keyring is not writable" errors when attempting to verify package signatures. Happy to remove them if there's a `mkosi` setting that ensures the keyring is already initialized at this point, but I wasn't able to find one.

Again, adding logs of the failed build I got when removing these lines.

[pacman-keyring-failed-build.txt](https://github.com/user-attachments/files/26660983/pacman-keyring-failed-build.txt)

PR #4268, opened 2026-04-04
Fix #4233: `mkosi` runs into permission errors even when running as root

### LINKED ISSUE #4233: `mkosi` runs into permission errors even when running as root
After debugging, most likely change to introduce this was #4160.

Essentially: I run `mkosi` as root for various reasons, but some time between v26 and the latest git commit caused this to fail. Kind of going mad, actually, trying to debug it, because it boils down to `config.build_dir.exists()` being false when run via `mkosi`, but not when I run it via a separate Python terminal referencing the same path.

What *appears* to be the issue is somehow, the sandboxing called before `ensure_directories_exist` is preventing access to my home directory, which contains the directory with all the mkosi configurations. This is the only possible explanation I can get, seeing as this is the debug output:

```
‣ Loading configuration file /home/ltdk/code/codeberg/cfg-system/mkosi.conf.d/0-base.conf
‣ Loading configuration file /home/ltdk/code/codeberg/cfg-system/mkosi.conf.d/1-x86-64.conf
‣ /home/ltdk/code/codeberg/cfg-system/build/incremental.x86-64/arch~rolling~x86-64~main.cache does not exist, not reusing cached images
‣ /home/ltdk/code/codeberg/cfg-system/build/incremental.x86-64/arch~rolling~x86-64~main.cache does not exist, not reusing cached images
‣ 0.0s for step "Validating certificates and keys"
‣ + tput cnorm
‣ + tput smam
  File "/usr/lib/python3.14/site-packages/mkosi/run.py", line 104, in uncaught_exception_handler
    yield
  File "/usr/lib/python3.14/contextlib.py", line 85, in inner
    return func(*args, **kwds)
  File "/usr/lib/python3.14/site-packages/mkosi/__main__.py", line 44, in main
    ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/mkosi/__init__.py", line 5195, in run_verb
    ensure_directories_exist(last)
    ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/usr/lib/python3.14/site-packages/mkosi/__init__.py", line 4787, in ensure_directories_exist
    p.mkdir(parents=True, exist_ok=True)
    ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/pathlib/__init__.py", line 1011, in mkdir
    os.mkdir(self, mode)
    ~~~~~~~~^^^^^^^^^^^^
PermissionError: [Errno 13] Permission denied: '/home/ltdk/code/codeberg/cfg-system/build'
```

When I did more testing, it appeared that the path existence check was failing… at my home directory, which is currently inaccessible to anything but my user. Which should not matter to root, but it does matter in the sandbox, presumably.

I get that running as root is an undesirable situation, but as far as I'm aware, your sandboxing isn't even good enough to run `sudo` inside the chroot, so, I can't even make the sandbox pretend that it's running as root for the sake of copying a few files. So, it would be nice if the sandboxing didn't totally break here.

JiwaniZakir:
Closes #4233

Skip user namespace entry in `acquire_privileges` when running as real root with a non-zero `delegate` value.

## Changes

**`mkosi/sandbox.py` — `acquire_privileges()`**
The condition on line 802 that gates early-exit (returning `False`, meaning "don't enter a user namespace") previously blocked that exit whenever `delegate` was truthy. It now allows the early exit when both `os.getuid() == 0` and `os.getgid() == 0`, so a real-root process is not forced into a user namespace solely because of a non-zero `delegate` argument.

```python
# before
and not delegate
# after
and (not delegate or (os.getuid() == 0 and os.getgid() == 0))
```

**`tests/test_sandbox.py`** (new file)
Adds `test_acquire_privileges_root_with_delegate_skips_userns`, which mocks `have_effective_cap` to return `True`, patches `os.getuid`/`os.getgid` to return `0`, and asserts that `acquire_privileges(foreign=True, delegate=3)` returns `False`.

## Motivation

When mkosi is invoked as root, the sandbox set up before `ensure_directories_exist()` was entering a user namespace because `delegate` was non-zero. Inside that namespace the process no longer held host-root privileges, making directories under paths like `/home/user/...` (mode `0700`) inaccessible. The result was a `PermissionError: [Errno 13] Permission denied` on a `mkdir` call for the build directory, even though the caller was UID 0 on the host. The regression was introduced in #4160, which added the `delegate` parameter path through `acquire_privileges`.

Real root does not benefit from a user namespace for privilege escalation purposes — it already has the capabilities it needs — so skipping namespace entry when `uid == gid == 0` is both correct and safe.

## Testing

Unit test added in `tests/test_sandbox.py` validates the fixed code path directly:

```
$ python -m pytest tests/test_sandbox.py -v
tests/test_sandbox.py::test_acquire_privileges_root_with_delegate_skips_userns PASSED
```

Manual verification: running `mkosi` as root with a build directory inside a `0700` home directory no longer raises `PermissionError` and proceeds through `ensure_directories_exist()` successfully.

---
*This PR was created with AI assistance (Claude). The changes were reviewed by quality gates and a critic model before submission.*

### DIFF
--- a/tests/test_sandbox.py
+++ b/tests/test_sandbox.py
@@ -1,0 +1,22 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+from unittest import mock
+
+from mkosi.sandbox import acquire_privileges
+
+
+def test_acquire_privileges_root_with_delegate_skips_userns() -> None:
+    """Running as root with delegate>0 should not enter a user namespace.
+
+    Regression test for https://github.com/systemd/mkosi/issues/4233:
+    when mkosi runs as root, acquire_privileges(foreign=True, delegate=3)
+    must return False (skip namespace entry) so that subsequent calls to
+    ensure_directories_exist() retain host-root filesystem access.
+    """
+    with (
+        mock.patch("mkosi.sandbox.have_effective_cap", return_value=True),
+        mock.patch("os.getuid", return_value=0),
+        mock.patch("os.getgid", return_value=0),
+    ):
+        result = acquire_privileges(foreign=True, delegate=3)
+    assert result is False
--- a/mkosi/sandbox.py
+++ b/mkosi/sandbox.py
@@ -799,7 +799,7 @@ def acquire_privileges(
         have_effective_cap(CAP_SYS_ADMIN)
         and identity
         and (not foreign or have_effective_cap(CAP_CHOWN))
-        and not delegate
+        and (not delegate or (os.getuid() == 0 and os.getgid() == 0))
         and (not become_root or (os.getuid() == 0 and os.getgid() == 0))
     ):
         return False


tests/test_sandbox.py
```diff
@@ -0,0 +1,22 @@
...
+    must return False (skip namespace entry) so that subsequent calls to
+    ensure_directories_exist() retain host-root filesystem access.
+    """
+    with (
+        mock.patch("mkosi.sandbox.have_effective_cap", return_value=True),
+        mock.patch("os.getuid", return_value=0),
+        mock.patch("os.getgid", return_value=0),
+    ):
+        result = acquire_privileges(foreign=True, delegate=3)
+    assert result is False
```
behrmann, line 22:
```suggestion
        assert not acquire_privileges(foreign=True, delegate=3)
```

tests/test_sandbox.py
behrmann, line 1: The regression test is a nice thought, but I don't see it surviving the next refactor. The internal API of `sandbox.py` changes as requirements for it change. The relevant thing is the behaviour as seen from the outside.

MoltenArmor:
Hi, just wondering if there is any progress on this PR?

ToolmanP:
this is still broken, is this PR still ongoing?

behrmann:
I've put up #4462 as an alternative to this. If people affected by this could have a look, that would be great.

PR #4271, opened 2026-04-06
Add SparseOutput= setting to convert disk images to Android sparse format

craftyguy:
This uses img2simg on the raw disk image after partition splitting but before compression, so split partition artifacts remain usable and the sparse image can still be compressed with CompressOutput=. This is useful for producing images that can be flashed via fastboot. Requires img2simg from android-tools.

I decided to use a new separate option for toggling this, rather than using Format=, because the build pipeline is identical to Format=disk and adding a new format would require updating every OutputFormat.disk check across the codebase. It's also not a CompressOutput= value since sparse is a format conversion, not compression, and users may want to compress the sparse image afterwards.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2779,6 +2779,9 @@ def check_inputs(config: Config) -> None:
             ),
         )
 
+    if config.sparse_output and config.output_format != OutputFormat.disk:
+        die("SparseOutput=yes can only be used with Format=disk")
+
     if config.is_incremental() and config.cacheonly == Cacheonly.never:
         die("Incremental=yes cannot be used with CacheOnly=never")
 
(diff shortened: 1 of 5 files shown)


craftyguy:
I'll fix the ruff format failure in CI, but first I'd like feedback on this approach :)

behrmann:
> I'll fix the ruff format failure in CI, but first I'd like feedback on this approach :)

I'll talked to @daandemeyer, since I figured that's something that would usually go into repart, but he thinks it's too invasive to reimplement. 

Personally, I also think it's somewhat niche as in "it targets something very specific" (notwithstanding, that these are probably by orders of magnitudes the most-widely deployed devices), but I also think it's a good idea to have this be built in, since it's small enough and makes it self-contained to built pmOS images without having to always carry a script around to fix the image up after the fact.

craftyguy:
@behrmann Thanks for the feedback.

If I could have implemented this in a mkosi script, I totally would have... the problem was that there was no mkosi script that fit exactly where this would need to run. It has to happen *after* mkosi.finalize (so the image is "complete"), but before mkosi.postoutput so that things like Compression= can be used on it. Basically it falls between steps 30 and 31 (from the manpage):

>       30. Generate final output format
> 
>       31. Run post‐output scripts (mkosi.postoutput)


I guess I *could* use a postoutput scrit and extract-->img2simg-->compress again but that's not ideal :(

behrmann:
Sorry if my post was a bit obtuse: I agree that it would be better to have it in mkosi and Daan at least mentioned that he thought it was too invasive for repart.

craftyguy:
@behrmann Ohhh ok, ya I misunderstood 😅 Is there anything I should change about this approach?

behrmann:
I think the approach looks reasonable.

Would be nice if we had tests for it, but it would be nice if we had pmOS tests in the first place :)

craftyguy:
I learned recently that some OEMs have different definitions of "sparse image" and it requires running a different tool with different options... So maybe instead of this param taking a bool, it should take a value that indicates the "type" of sparse image to generate?

behrmann:
If there's multiple implementations of sparse outputs, then it would make sense to name them, yes.

PR #4272, opened 2026-04-09
Avoid invalidating all caches on mid-build-failure recovery

daandemeyer:
Previously, repository_metadata_needs_sync() returned True whenever any Cacheonly=auto image lacked a cache, which caused run_clean() to wipe every image cache to avoid partial-upgrade scenarios. As a result, if mkosi was building several images for the first time and one of them failed mid-way, the next run would re-sync metadata, wipe the caches of the images that had already succeeded, and rebuild everything from scratch.

Use the manifest file as a "previously cached" marker — it's written when an image builds successfully and survives even after the cache contents go stale — and only re-sync metadata (and invalidate all caches) when every incremental image was previously cached and at least one of them is now out of date. In the mid-build-failure case, the image that failed never wrote a manifest, so we don't trigger the re-sync and the surviving caches are preserved.

For all-non-incremental setups we keep the previous behavior of always syncing — there are no caches to preserve and every build downloads packages. In a mixed setup, the non-incremental images are ignored by the heuristic and may eventually fail because of stale metadata, but that's the cost of mixing them with incremental images we want to preserve.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -4798,14 +4798,38 @@ def ensure_directories_exist(config: Config) -> None:
 
 
 def repository_metadata_needs_sync(images: Sequence[Config]) -> bool:
-    for config in images:
-        if config.cacheonly == Cacheonly.never:
-            return True
+    if any(c.cacheonly == Cacheonly.never for c in images):
+        return True
 
-        if config.cacheonly == Cacheonly.auto and not have_cache(config):
-            return True
+    if not (auto := [c for c in images if c.cacheonly == Cacheonly.auto]):
+        return False
 
-    return False
+    # If there are no incremental images at all, we have no caches to preserve, so
+    # always sync to avoid failures from stale metadata. In a mixed setup, ignore
+    # the non-incremental images and apply the heuristic to the incremental ones only.
+    # Non-incremental images in a mixed setup may eventually fail because of stale
+    # metadata, but that's the cost of mixing them with incremental images we want
+    # to preserve.
+    if not (incremental := [c for c in auto if c.is_incremental()]):
+        return True
+
+    # Use the manifest file as the "previously cached" marker: it's written when an
+    # image builds successfully and survives even after the image cache content goes
+    # stale. If no incremental image has a manifest, no image has ever been built
+    # before, so there's nothing to be consistent with and we have to sync to get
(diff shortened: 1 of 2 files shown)


mkosi/__init__.py
```diff
@@ -4798,14 +4798,35 @@ def ensure_directories_exist(config: Config) -> None:
...
-        if config.cacheonly == Cacheonly.never:
-            return True
+    if any(c.cacheonly == Cacheonly.never for c in images):
+        return True
 
-        if config.cacheonly == Cacheonly.auto and not have_cache(config):
-            return True
+    auto = [c for c in images if c.cacheonly == Cacheonly.auto]
+    if not auto:
+        return False
```
```suggestion
    if not (auto := [c for c in images if c.cacheonly == Cacheonly.auto]):
        return False
```

mkosi/__init__.py
```diff
@@ -4798,14 +4798,35 @@ def ensure_directories_exist(config: Config) -> None:
...
 
-    return False
+    # If there are no incremental images at all, we have no caches to preserve, so
+    # always sync to avoid failures from stale metadata. In a mixed setup, ignore
+    # the non-incremental images and apply the heuristic to the incremental ones only.
+    # Non-incremental images in a mixed setup may eventually fail because of stale
+    # metadata, but that's the cost of mixing them with incremental images we want
+    # to preserve.
+    incremental = [c for c in auto if c.is_incremental()]
+    if not incremental:
```
```suggestion
    if not (incremental := [c for c in auto if c.is_incremental()]):
```

mkosi/__init__.py
```diff
@@ -4798,14 +4798,40 @@ def ensure_directories_exist(config: Config) -> None:
...
+    # everything to avoid partial upgrades. Otherwise, some images have manifests
+    # and some don't (e.g. recovering from a mid-build failure), and the existing
+    # metadata is by construction consistent with the caches that do exist, so we
+    # don't need to re-sync.
+    if not any(cache_tree_paths(c)[2].exists() for c in incremental):
+        return True
+
+    return all(cache_tree_paths(c)[2].exists() for c in incremental) and any(
+        not have_cache(c) for c in incremental
+    )
```
behrmann: This one too would be easier to read as
```suggestion
    return all(cache_tree_paths(c)[2].exists() for c in incremental) and not all(
        have_cache(c) for c in incremental
    )
```

PR #4273, opened 2026-04-09
Revert "Add Incremental=relaxed"

daandemeyer:
This does not work as a per distribution CacheKey= means that we can switch between distributions without invalidating caches, meaning the output files won't be removed yet will still be those of the previous distribution rather than the new one.

Revert for now until I figure out a better approach.

This reverts commit a97c50044ce19d72e5119cd11af6993c3944ab65.


behrmann (approved):
Since the main user is already merged.

PR #4274, opened 2026-04-11
apk: Fix EBUSY in sync() when context.root is a mount point

craftyguy:
sync() temporarily moved context.root out of the way to get an empty directory to run apk against without touching the image root. This fails when context.root is a mount point, e.g. when Overlay= is enabled:

    OSError: [Errno 16] Resource busy: '...root' -> '...saved-root'

This uses a throwaway directory in the workspace instead, since apk doesn't need to run against context.root specifically.


behrmann (approved):

behrmann:
@craftyguy can you rebase please?

craftyguy:
@behrmann  done! please have a look :)

PR #4275, opened 2026-04-12
feat: added confirmation prompt to mkosi burn before burning image to device

### LINKED ISSUE #3821: Add confirmation prompt for mkosi burn
`mkosi burn` needs a confirmation dialog to avoid nuking a system drive. Additionally it should print out the device size and vendor, and the sizes, file system types, mount points, and names of all partitions. This will add an additional layer of protection. All the required information could be retrieved via this command `lsblk -o PATH,LABEL,PARTLABEL,FSTYPE,SIZE,HOTPLUG,MOUNTPOINTS </path/to/device>`

estebanpuyanas:
# Explanation: 

The confirmation prompt runs `lsblk` to retrieve device information and then requests the user types `y` or `yes` to proceed and `n` or `no` to abort. If the user chooses to proceed, it will continue the burning process as usual, otherwise it prints a message saying the burning process was aborted.

Closes #3821


mkosi/burn.py
```diff
@@ -21,6 +22,25 @@ def run_burn(args: Args, config: Config) -> None:
...
+
+    logging.info(f"About to burn image to device: {device}")
+    logging.info("The following block device layout will be overwritten:")
+    run(lsblk_command)
+
+    sys.stderr.write("Do you want to continue? [y/N] ")
+    sys.stderr.flush()
+    reply = sys.stdin.readline().strip().lower()
+    if reply not in ("y", "yes"):
+        die("Aborting burning image to device.")
```
behrmann: Why not use `input`?

mkosi/burn.py
```diff
@@ -21,6 +22,25 @@ def run_burn(args: Args, config: Config) -> None:
     if len(args.cmdline) != 1:
         die("Expected device argument.")
 
+    device = args.cmdline[0]
+    lsblk_command = [
+        "lsblk",
+        "-o",
+        "PATH,LABEL,PARTLABEL,FSTYPE,SIZE,HOTPLUG,MOUNTPOINTS",
```
behrmann: In our codebase options and their arguments go on one line

```suggestion
        "-o", "PATH,LABEL,PARTLABEL,FSTYPE,SIZE,HOTPLUG,MOUNTPOINTS",
```
This requires a `# fmt: skip` comment on the closing line of the list.

behrmann (commented):
This functionality should have gone into repart, as mentioned in the issue, though I'm not sure whether that is still the case with the installer work that's ongoing. I think that would still make more sense. 

That being said, a few comments

behrmann:
I'm somewhat of two minds about this. This operation already requires elevated privileges, and none of the other commands usually used to get images onto devices (`cp`, `dd`, `cat`) will ask by default . Also, having this be always interactive makes it more involved to run this non-interactively, though I see that happening more rarely with this command. 

I'm ±0 on this, but I see that there are people who legitimately want to be on the safe side here. If it's not to go into repart, maybe an `--interactive` switch would make sense here.

PR #4279, opened 2026-04-21
ukify: Support multiple PCR signing keys to sign phases separately

### LINKED ISSUE #4109: Support for Phases= from systemd-measure

main

### Used host distribution

Arch

### Used target distribution

Arch

### Linux kernel version used

6.18.2-arch2-1

x86_64

the mkosi-initrd configuration doesn't seem to accept a `Phases=` switch like ukify does.

Thus secrets from the initrd phase are accessible from the booted system.

### Used mkosi config

```ini
[Validation]
SecureBoot=yes
SignExpectedPcr=yes
SecureBootKey=/etc/kernel/secure-boot-private-key.pem
SecureBootCertificate=/etc/kernel/secure-boot-certificate.pem
SignExpectedPcrKey=/etc/systemd/tpm2-pcr-private-key.pem
SignExpectedPcrCertificate=/etc/systemd/tpm2-pcr-public-key.pem
```

### mkosi output

```sh

```

pothos:
To be able to bind a secret against a signed PCR policy so that it only unlocks in the initrd but not the final system one has to sign the policies with separate keys because that's what one actually binds to. This is supported in ukify as "phases" and ukify accepts the PCR signing keys specified multiple times together with the phase they are for. We can't extend the existing config setting for this scheme, so instead we have to add a new setting which can hold the configuration entries. Also, we need a new setting to specify which PCR pub key goes into the .pcrpkey UKI section because when we have many, ukify won't add any by default.
Add SignExpectedPcrWithPhases= as alternative to SignExpectedPcrKey=/ SignExpectedPcrCertificate=/*Source= where we can now specify the phase signing configuration one line each. A line contains KEY CERT PHASES [KEYSRC [CERTSRC]] where KEYSRC/CERTSRC defaults to "file". One can also set "-" for PHASES to omit --phases for ukify so that it chooses the defaults. Also add SignExpectedPcrUKIPublicKey= to be able to specify the key for the .pcrpkey UKI section.

Closes https://github.com/systemd/mkosi/issues/4109

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -63,6 +63,7 @@ from mkosi.config import (
     ManifestFormat,
     Network,
     OutputFormat,
+    PcrSigner,
     SecureBootSignTool,
     ShimBootloader,
     Ssh,
@@ -1583,8 +1584,10 @@ def want_signed_pcrs(config: Config) -> bool:
     return config.sign_expected_pcr == ConfigFeature.enabled or (
         config.sign_expected_pcr == ConfigFeature.auto
         and config.find_binary("systemd-measure", "/usr/lib/systemd/systemd-measure") is not None
-        and bool(config.sign_expected_pcr_key)
-        and bool(config.sign_expected_pcr_certificate)
+        and (
+            (bool(config.sign_expected_pcr_key) and bool(config.sign_expected_pcr_certificate))
+            or bool(config.sign_expected_pcr_with_phases)
+        )
     )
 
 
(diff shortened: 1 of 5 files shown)


mkosi/__init__.py
```diff
@@ -1755,73 +1758,105 @@ def build_uki(
...
+        need_run_bind = False
+
+        for signer in signers:
+            # If we're providing the private key via an engine or provider, we have to pass in
+            # a X.509 certificate via --pcr-certificate as well.
+            if signer.key_source.type != KeySourceType.file:
+                if signer.certificate_source.type == CertificateSourceType.provider:
+                    arguments += [
+                        "--certificate-provider",
+                        f"provider:{signer.certificate_source.source}",
```
behrmann: Preferably put these on the same line as well, like below

mkosi/__init__.py
```diff
@@ -1755,73 +1758,105 @@ def build_uki(
...
+                    ]
+
+                need_run_bind = True
+
+                if signer.certificate.exists():
+                    arguments += [cert_parameter, workdir(signer.certificate)]
+                    if signer.certificate not in seen_bind_paths:
+                        options += [
+                            "--ro-bind", signer.certificate, workdir(signer.certificate),
+                        ]  # fmt: skip
```
behrmann: You could move this below and add this once in a loop, then you could also skip the checks whether the path is in the set, since adding it again will be a noop.

mkosi/config.py
```diff
@@ -1733,6 +1733,88 @@ def config_parse_certificate_source(
...
+    # quoted with " or '). Line continuation with trailing backslash is folded by the list parser
+    # before we see the value, so each invocation sees exactly one logical entry.
+    try:
+        parts = shlex.split(value)
+    except ValueError as e:
+        die(f"Failed to parse SignExpectedPcrWithPhases= entry '{value}': {e}")
+
+    if not 3 <= len(parts) <= 5:
+        die(
+            f"SignExpectedPcrWithPhases= entry must have 3 to 5 whitespace-separated fields "
```
behrmann: This one doesn't need to be an f-string
```suggestion
            "SignExpectedPcrWithPhases= entry must have 3 to 5 whitespace-separated fields "
```

mkosi/config.py
```diff
@@ -6144,6 +6312,27 @@ def credential_transformer(
...
 
+    def pcr_signer_transformer(
+        signers: list[dict[str, Any]],
+        fieldtype: type[list[PcrSigner]],
+    ) -> list[PcrSigner]:
+        return [
+            PcrSigner(
+                key=Path(d["Key"]),
+                certificate=Path(d["Certificate"]),
+                phases=list(d.get("Phases", [])),
```
behrmann: The `list` call is not necessary.
```suggestion
                phases=d.get("Phases", []),
```

mkosi/resources/man/mkosi.1.md
```diff
@@ -1456,10 +1456,93 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
+    SignExpectedPcrWithPhases=
+        /keys/tpm2-pcr-initrd-private-key.pem /keys/tpm2-pcr-initrd.crt enter-initrd
+        /keys/tpm2-pcr-system-private-key.pem /keys/tpm2-pcr-system.crt enter-initrd:leave-initrd,enter-initrd:leave-initrd:sysinit,enter-initrd:leave-initrd:sysinit:ready
+    SignExpectedPcrUKIPublicKey=/keys/tpm2-pcr-initrd-public-key.pem
+    ```
+
+`SignExpectedPcrUKIPublicKey=`, `--sign-expected-pcr-uki-public-key=`
+:   Path to a PEM-encoded public key to embed in the UKI's `.pcrpkey` PE
+    section (forwarded to **ukify** as `--pcrpkey=`). The file is embedded
+    verbatim by **ukify**; an X.509 certificate is *not* accepted here
```
behrmann: Here, as well as further below in `config.py` in an error message: Instead of semicolons, make more sentences, I think that makes more sense for readability instead of multiple concatenated sentences. The same probably goes for most parenthesized parts, that would go better into a sentence or a subclause with a comma.

mkosi/resources/man/mkosi.1.md
```diff
@@ -1456,10 +1456,93 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
+    (convert with: `openssl x509 -in mkosi.crt -pubkey -noout`).
+    When not set, **ukify** auto-derives the section contents from the
```
```suggestion
    When not set, **ukify** derives the section contents from the
```

mkosi/config.py
```diff
@@ -5430,6 +5531,64 @@ def finalize_historydir(args: Args) -> Path:
...
+            if (
+                config.sign_expected_pcr_certificate is not None
+                and config.sign_expected_pcr_certificate.name == "mkosi.crt"
+            ):
+                clears.append("SignExpectedPcrCertificate=")
+            hint = None
+            if clears:
+                hint = (
+                    "mkosi.key/mkosi.crt in the config directory are picked up automatically; "
+                    "delete the file(s) or clear the auto-pickup by adding to [Validation]: "
```
```suggestion
                    "delete the file(s) or clear the pickup by adding to [Validation]: "
```

mkosi/resources/man/mkosi.1.md
```diff
@@ -1456,10 +1456,93 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
+    to explicitly control what goes into the `.pcrpkey` section.
+
+    Because **ukify** relies on a single global `--signing-engine`/
+    `--signing-provider` per invocation, all signer entries must use the
+    same mechanism for their `KEYSOURCE` (and `CERTIFICATESOURCE`) fields,
+    and the same mechanism as `SecureBootKeySource=`/`SecureBootCertificateSource=`
+    when Secure Boot signing is also enabled.
+
+    Note that `mkosi.key` and `mkosi.crt` in the config directory are
+    auto-picked-up by `SignExpectedPcrKey=` and `SignExpectedPcrCertificate=`
```
```suggestion
    automatically picked up by `SignExpectedPcrKey=` and `SignExpectedPcrCertificate=`
```

mkosi/resources/man/mkosi.1.md
```diff
@@ -1456,10 +1456,93 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
...
+    Entries are separated by newlines. Fields within an entry are
+    whitespace-separated with shell-style quoting, so values
+    containing spaces can be quoted with `"` or `'`. A trailing `\`
+    continues the entry on the next line.
+
+    Requires **ukify** version 253 or newer.
+
+    When multiple entries are configured, **ukify** will not embed a
+    `.pcrpkey` PE section in the UKI by default (it only auto-embeds when
+    exactly one signing key is supplied). Set `SignExpectedPcrUKIPublicKey=`
```
```suggestion
    `.pcrpkey` PE section in the UKI by default, it only does so when
    exactly one signing key is supplied. Set `SignExpectedPcrUKIPublicKey=`
```

behrmann (commented):
Arghs, forgot to press the submit button

mkosi/resources/man/mkosi.1.md
pothos: Thanks, done

mkosi/__init__.py
pothos: Good idea, done

bluca:
The general feature looks ok, as long as the existing configs keep working. I'm not 100% convinced on the format of the configuration, giving semantic meaning to line breaks, but I'll leave that decision to @daandemeyer and @behrmann

behrmann:
> I'm not 100% convinced on the format of the configuration, giving semantic meaning to line breaks

That is indeed a bit hairy, but the problem is that we're running out of separators at this interplay, since between mkosi and ukify we're already using whitespace, commas, and colons. Could this maybe be better served by an indirection putting this into the native config files for ukify and passing them to ukify from some well-known place, e.g. some `mkosi.ukify`?

pothos:
I think we won't need this anymore because the single-key phase signing works now with the policy references and we don't need separate signing keys per phase. Not sure if all this is wired up already, though.

PR #4282, opened 2026-04-23
postmarketOS: add all supported architectures

Gelbpunkt:
This omits armhf, which is armv6, since mkosi only knows a single architecture for all 32-bit ARM variants. It is eventually going to be removed upstream anyways, so let's not bother with it.


behrmann:
@craftyguy I see a :+1: from you, so you think this is good to go?

craftyguy:
@behrmann yep, looks good to me!

behrmann (approved):

PR #4283, opened 2026-04-25
Compare contents before deduplicating against base tree

craftyguy:
The overlay deduplication unconditionally removed any non-directory file
present in both the upper layer and the base tree, even when the overlay
file had different contents. This is visible on Alpine/postmarketOS
where busybox and coreutils both install symlinks at the same paths in
/usr/bin but with different targets. Installing coreutils in a sysext
replaces the busybox symlinks in the overlay, but the deduplication
removed them, leaving the old busybox targets visible at runtime.

This compares symlink targets and file contents before removing them.

Signed-off-by: Clayton Craft <clayton@craftyguy.net>

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2,6 +2,7 @@
 
 import contextlib
 import datetime
+import filecmp
 import functools
 import getpass
 import hashlib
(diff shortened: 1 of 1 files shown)


daandemeyer:
In my opinion extensions overriding files from the base system is rather broken so I disagree with allowing extensions that override the base system.

craftyguy:
> In my opinion extensions overriding files from the base system is rather broken so I disagree with allowing extensions that override the base system.

Hmm yeah I see your point. I thought extensions were basically a "diff" against some base image.

I think I can solve my specific problem another way (just make sure coreutils pkg is in the base image), so I'm totally fine with closing this if you disagree with it.

PR #4288, opened 2026-05-02
config: Don't export MKOSI_DNF into a tools tree

### LINKED ISSUE #4280: MKOSI_DNF propagates to tools tree

main

### Used host distribution

silverblue 44

### Used target distribution

fedora 44

### Linux kernel version used

_No response_

None

If `MKOSI_DNF` is set (to `dnf4` for example), mkosi will correctly use it to create the tools tree when ToolsTree=yes. However, the build of the image then also tries to use that executable (so again `dnf4`) when the tools tree only contains `dnf`. `MKOSI_DNF` should only be effective in the host context.

### Used mkosi config

```ini

```

### mkosi output

```sh

```

src-up:
When a tools tree is in use, the dnf executable runs from inside the tools tree sandbox, not from the host. Propagating $MKOSI_DNF from the host environment into the tools tree leads to unpredictable results as the dnf executable may not exist inside the tools tree. 

Fixes #4280

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -2310,7 +2310,7 @@ class Config:
             env["GIT_PROXY_SSL_CERT"] = "/proxy.clientcert"
         if self.proxy_client_key:
             env["GIT_PROXY_SSL_KEY"] = "/proxy.clientkey"
-        if dnf := os.getenv("MKOSI_DNF"):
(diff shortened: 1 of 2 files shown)


keszybz (approved):
LGTM.

PR #4289, opened 2026-05-08
apk: skip removal of packages that aren't installed

craftyguy:
This fixes a failure when RemovePackages= listed package that wasn't installed. The installed package database is queried first, and filter to only packages that are actually present before calling apk del. The query matches against both name and provides fields so packages specified by a provider name are also found.

### DIFF
--- a/mkosi/installer/apk.py
+++ b/mkosi/installer/apk.py
@@ -1,6 +1,8 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
 import dataclasses
+import json
+import subprocess
 from collections.abc import Sequence
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


mkosi/installer/apk.py
```diff
@@ -125,7 +127,33 @@ def install(
 
     @classmethod
     def remove(cls, context: Context, packages: Sequence[str]) -> None:
-        cls.invoke(context, "del", packages, apivfs=True)
+        # Match against both name and provides so that packages specified
+        # by a provider name are also found.
```
behrmann, line 131: From context pretty clear, but just to be safe, the result is guaranteed to be a JSON list? Would be nice to add a comment to that regard.

mkosi/installer/apk.py
```diff
@@ -125,7 +127,33 @@ def install(
...
+            stdout=subprocess.PIPE,
+        )  # fmt: skip
+
+        # Build a lookup set from both real package names and their provides entries
+        # (stripped of version suffixes) so we can match against the original input names.
+        installed: set[str] = set()
+        for pkg in json.loads(result.stdout or "[]"):
+            installed.add(pkg["name"])
+            for p in pkg.get("provides", []):
+                installed.add(p.split("=", 1)[0])
```
behrmann:
```suggestion
            installed.update(p.split("=", 1)[0] for p in pkg.get("provides", [])
```

behrmann (approved):

mkosi/installer/apk.py
craftyguy: ok tried to improve the comment :)

mkosi/installer/apk.py
behrmann: Is it intentional, that the `installed.add(pkg["name"])` went missing as well?

craftyguy: ugh no, I messed up. thanks for catching that and sorry :)

behrmann:
CI failures are unrelated, since this is APK/pmOS-specific.

PR #4291, opened 2026-05-08
Make OUTPUT variable available in postoutput scripts

Andre-85:
This patch makes the variable OUTPUT available in postoutput scripts.

In my use case i wanted to add an symlink to the generated image. But in the outputdir are two entries: The image itself (with the output extension) and the symlink to it (without the extention). So a simple: 
`ln -s ${OUTPUTDIR}/* <location>` will not work since * matches two files. So i ended up doing `ln -s ${OUTPUTDIR}/*.img <location>`. But that's a foot gun: As soon as i change the configured OutputExtension from "img" to something elses this script would fail.

With the new OUTPUT variable i can simply write  `ln -s ${OUTPUTDIR}/$OUTPUT <location>` (for using the symlink) or `ln -s ${OUTPUTDIR}/${OUTPUT}.* <location>` (for the image itself) .

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1130,6 +1130,7 @@ def run_postoutput_scripts(context: Context) -> None:
         ),
         SRCDIR="/work/src",
         OUTPUTDIR="/work/out",
+        OUTPUT=context.config.output,
         MKOSI_UID=str(os.getuid()),
         MKOSI_GID=str(os.getgid()),
         MKOSI_CONFIG="/work/config.json",


behrmann:
I don't think this is necessary, since you can get the output file name via e.g. `jq .Output $MKOSI_CONFIG`.

Assuming this should be added, this would be missing:
- documentation in the man page
- an answer to why this should only be added to the postoutput script and not other scripts, like the build script, as well.

Andre-85:
> I don't think this is necessary, since you can get the output file name via e.g. `jq .Output $MKOSI_CONFIG`.
> 
> Assuming this should be added, this would be missing:
> 
>     * documentation in the man page
> 
>     * an answer to why this should only be added to the postoutput script and not other scripts, like the build script, as well.

I am aware of this possibility to do this with jq. It was just about to use an extra tool at all. So in theory it would also be a possibility not to expose any environment variables to the postoutput (and perhaps to the other scripts) except MKOSI_CONFIG at all since all environment variables available in the scripts right now, can be extracted from $MKOSI_CONFIG as well. 

I thought that the main purpose of the environment variables right now, is to get easy and fast access to the information that is needed in most cases. In the case of the postoutput script i would say it is doing further manipulation to the output image, but as I stated that the name of the output image needs to be guessed somehow or the jq tool needs to be used.

To your question why the output variable should be only being exposed to postoutput: As far as I understand is postoutput hook script is the only hook script type where the output image already exists, so i guess it would be pointless for the other scripts.

I case you are ok with my explanation I would of course also add info about this to the manpage.

What do you think?

behrmann:
> I thought that the main purpose of the environment variables right now, is to get easy and fast access to the information that is needed in most cases. In the case of the postoutput script i would say it is doing further manipulation to the output image, but as I stated that the name of the output image needs to be guessed somehow or the jq tool needs to be used.

It's a bit of a historical thing. Environment variables are easy to expose. Later on we grew the feature of serialising the config and started exposing the whole config as JSON, so we no longer needed to add an environment variable for every single thing people might want to have exposed.

PR #4294, opened 2026-05-10
Force white foreground color

glemco:
By default, mkosi sets a dark green background color but doesn't touch
the foreground color. This is completely unreadable on light themed
terminals where the foreground is black.

Add the foreground color specifier to white.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -4253,7 +4253,7 @@ def run_box(args: Args, config: Config) -> None:
     cmdline = [*args.cmdline]
 
     if sys.stdin.isatty() and sys.stdout.isatty():
-        cmdline = systemd_pty_forward(config, background="48;2;12;51;51", title="mkosi-sandbox") + cmdline
+        cmdline = systemd_pty_forward(config, background="48;2;12;51;51;38;5;15", title="mkosi-sandbox") + cmdline
 
     with contextlib.ExitStack() as stack:
         if config.tools() != Path("/"):
--- a/mkosi/qemu.py
+++ b/mkosi/qemu.py
@@ -1014,7 +1014,7 @@ def run_qemu(args: Args, config: Config) -> None:
     if config.console in (ConsoleMode.interactive, ConsoleMode.read_only):
         cmdline += systemd_pty_forward(
             config,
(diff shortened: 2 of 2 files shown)


behrmann:
While I agree, that the current colour choice is unreadable on e.g. black on white (default xterm), I think forcing white foreground makes this worse on dark terminals.

You can disable the tinting by setting `SYSTEMD_TINT_BACKGROUND=1` in your environment.

glemco:
Thanks for the comment, but why would it make it worse on dark terminals? Isn't white the default foreground there?

I'm aware SYSTEMD_TINT_BACKGROUND would help here, but it took me a while to figure it out (it isn't documented directly in mkosi) and anyway I don't think the command should require an additional setting to be at all readable.

I thought of implementing some kind of dynamic background selection like systemd does (https://github.com/systemd/systemd/blob/0bf094b7636471d4c25df5b60d1e8fe2512601e5/src/shared/pretty-print.c#L509), that's probably the cleanest way if foreground is not viable on dark terminals.

What do you think?

behrmann:
> Thanks for the comment, but why would it make it worse on dark terminals? Isn't white the default foreground there?

Not sure it is? It's definitely not for me. If memory serves me well, in Ubuntu the default was a light grey? My point is rather, that changing the foreground is a bit unexpected considering the option is called `--background`. I do sympathize with your plight, though, I'm just wondering whether there's not a better way for ptyforward to grow a way where users can better control the colour—better than just switching the behaviour off.

glemco:
Well, you make a fair point about it being called --background..
It would indeed be much better to extend ptyforward to take something like a hue parameter that would then simply call the `terminal_tint_color()` I linked above.
Now I don't think that can be defined as ANSI Escape Code though, so the --background option would need to be extended to parse a custom syntax (e.g. h;44), or another option like --background-hue .
I can draft a PR directly to systemd and see what the folks there think about it.

Maybe even better: some flag like --preserve, which extracts the hue of the passed background and flips the saturation if that's too much for the theme.

behrmann:
Thanks, @glemco. Since the background tinting got dropped entirely in #4300 I'm gonna close this. Please resubmit if we need to tinker with the tinting some more.

PR #4296, opened 2026-05-11
nspawn: Use --forward-journal instead of running journal-remote ourselves

daandemeyer:
systemd-nspawn now supports --forward-journal with matching --forward-journal-max-use, --forward-journal-keep-free, --forward-journal-max-file-size and --forward-journal-max-files options, so let it run systemd-journal-remote itself instead of doing it ourselves via a bind-mounted unix socket.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -14,7 +14,6 @@ import resource
 import shlex
 import shutil
 import signal
-import socket
 import stat
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


behrmann (commented):
This needs a sufficiently new nspawn, so this should probably bail with a nice error message if nspawn is too old?

daandemeyer:
nspawn itself will bail out, should be sufficient.

PR #4297, opened 2026-05-11
ci: add postmarketOS to integration testing

craftyguy:
apk-tools isn't packaged in Ubuntu, so the scope of testing is limited quite a bit to only a handful of distros/tools trees that have apk available. This also means we're not able to test building images for other distros in a pmOS tools tree since the base distro is... 🥁... Ubuntu and it's unable to build a pmOS tools tree because it doesn't have apk-tools.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,7 @@ jobs:
           - debian
           - fedora
           - opensuse
+          - postmarketos
           - ubuntu
         tools:
           - arch
@@ -80,6 +81,13 @@ jobs:
             tools: opensuse
           - distro: ubuntu
             tools: opensuse
+          # apk is not packaged in Debian, Ubuntu, or EPEL.
+          - distro: postmarketos
+            tools: centos
+          - distro: postmarketos
+            tools: debian
+          - distro: postmarketos
+            tools: ubuntu
         include:
           # low rate limit on s390x/ppc64le/arm64 workers
           - distro: debian
@@ -103,6 +111,10 @@ jobs:
           - distro: debian
             tools: debian
             runner: ubuntu-24.04-s390x
+          # build pmOS using pmOS tools tree
+          - distro: postmarketos
+            tools: postmarketos
+            runner: ubuntu-24.04
 
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
--- a/action.yaml
+++ b/action.yaml
@@ -95,11 +95,14 @@ runs:
     - name: Install apk
       # ubuntu-24.04 doesn't package apk, so fetch the official statically-linked binary from
       # upstream so mkosi can use it to populate the postmarketOS tools tree.
+      #
+      # NOTE: apk is dropped into /usr/bin and not /usr/local/bin because later CI
+      # scripts (like .github/workflows/ci.yml) clean up /usr/local
       shell: bash
       run: |
-        sudo curl -fsSL -o /usr/local/bin/apk https://gitlab.alpinelinux.org/api/v4/projects/5/packages/generic/v3.0.6/x86_64/apk.static
-        echo 'f1489e05bace7d7dd0a687fcd38d50b585ac660af4231668b123649bef3718c4  /usr/local/bin/apk' | sha256sum --check
-        sudo chmod +x /usr/local/bin/apk
+        sudo curl -fsSL -o /usr/bin/apk https://gitlab.alpinelinux.org/api/v4/projects/5/packages/generic/v3.0.6/x86_64/apk.static
+        echo 'f1489e05bace7d7dd0a687fcd38d50b585ac660af4231668b123649bef3718c4  /usr/bin/apk' | sha256sum --check
+        sudo chmod +x /usr/bin/apk
 
     - name: Dependencies
       shell: bash
--- a/mkosi.tools.conf/mkosi.conf.d/postmarketos.conf
+++ b/mkosi.tools.conf/mkosi.conf.d/postmarketos.conf
@@ -6,4 +6,5 @@ Distribution=postmarketos
 [Content]
 Packages=
         py3-codespell
+        py3-pytest
         ruff
--- a/mkosi/resources/mkosi-tools/mkosi.conf.d/postmarketos.conf
+++ b/mkosi/resources/mkosi-tools/mkosi.conf.d/postmarketos.conf
@@ -8,14 +8,22 @@ Packages=
         abuild
         apk-tools
         btrfs-progs
+        cryptsetup
         distribution-gpg-keys
         erofs-utils
-        # NOTE: grub disabled because package trigger fails
-        # grub
+        grub
+        grub-efi
         libfido2
         libseccomp
         p11-kit
         sbsigntool
+        sfdisk
         squashfs-tools
         ukify
         xz
+
+        # These replace the busybox applets, which don't support the full set of
+        # features that mkosi needs
+        losetup
+        mount
+        umount
--- a/mkosi/resources/mkosi-tools/mkosi.profiles/package-manager/mkosi.conf.d/postmarketos.conf
+++ b/mkosi/resources/mkosi-tools/mkosi.profiles/package-manager/mkosi.conf.d/postmarketos.conf
@@ -6,6 +6,7 @@ Distribution=postmarketos
 [Content]
 Packages=
         apt
+        archlinux-keyring
         pacman
         debian-archive-keyring
         ubuntu-archive-keyring
--- a/mkosi.conf.d/postmarketos.conf
+++ b/mkosi.conf.d/postmarketos.conf
@@ -5,6 +5,8 @@ Distribution=postmarketos
 
 [Content]
 Packages=
+        grub
+        grub-efi
         openssh-server
         openssh-server-pam-systemd
 
--- a/mkosi/resources/mkosi-initrd/mkosi.conf.d/postmarketos.conf
+++ b/mkosi/resources/mkosi-initrd/mkosi.conf.d/postmarketos.conf
@@ -6,9 +6,11 @@ Distribution=postmarketos
 [Content]
 Packages=
         cryptsetup-libs
+        device-mapper
         kbd
         libseccomp
         util-linux
+        util-linux-login
 
         # systemd-repart dependency
         libfdisk
--- a/mkosi/resources/mkosi-tools/mkosi.profiles/package-manager/mkosi.conf.d/arch.conf
+++ b/mkosi/resources/mkosi-tools/mkosi.profiles/package-manager/mkosi.conf.d/arch.conf
@@ -5,6 +5,7 @@ Distribution=arch
 
 [Content]
 Packages=
+        apk-tools
         apt
         debian-archive-keyring
         createrepo_c
--- a/mkosi/resources/mkosi-tools/mkosi.profiles/package-manager/mkosi.conf.d/opensuse.conf
+++ b/mkosi/resources/mkosi-tools/mkosi.profiles/package-manager/mkosi.conf.d/opensuse.conf
@@ -5,5 +5,6 @@ Distribution=opensuse
 
 [Content]
 Packages=
+        apk-tools
         dnf5
         dnf5-plugins
--- a/mkosi/resources/mkosi-vm/mkosi.conf.d/postmarketos.conf
+++ b/mkosi/resources/mkosi-vm/mkosi.conf.d/postmarketos.conf
@@ -14,6 +14,7 @@ Packages=
         postmarketos-base
         postmarketos-base-systemd
         systemd-boot
+        systemd-efistub
         systemd-networkd
         systemd-udevd
         tpm2-tools


craftyguy:
This opensuse tools tree failure seems unrelated to my changes, it's also failing when building images for other distros:

```
[32](https://github.com/systemd/mkosi/actions/runs/25703288918/job/75468214832?pr=4297#step:10:4133)
+ /usr/lib/systemd/systemd-keyutil extract-public --private-key /work/home/runner/work/mkosi/mkosi/mkosi.key
+ /usr/lib/systemd/systemd-measure sign --osrel=/work/var/tmp/mkosi-workspace-ovxxn95g/root/usr/lib/os-release --cmdline=/work/var/tmp/mkosi-workspace-ovxxn95g/cmdline --uname=/tmp/tmp.uname0_1a3qma --pcrpkey=/tmp/tmp.pcrpkey7wvwhp5e --linux=/work/var/tmp/mkosi-workspace-ovxxn95g/root/usr/lib/modules/6.18.29-0-virt/vmlinuz --initrd=/tmp/tmp.initrdxwyk1xfg --sbat=/tmp/tmp.sbatetn4_75z --bank=sha256 --private-key=/work/home/runner/work/mkosi/mkosi/mkosi.key
Could not calculate PolicyPCR digest: Operation not supported

**Comment by behrmann** [comment]
Thanks for this! 

Yeah, opensuse is borked at the moment and nobody had time to investigate yet.

**Comment by daandemeyer** [comment]
@craftyguy We can mess around a bit, you can add a step to the mkosi action to download a static apk (check the hash ofc) and together with #4313 we'll be able to build from ubuntu

**Comment by daandemeyer** [comment]
If you rebase we now install apk in the github action

**Comment by behrmann** [comment]
@craftyguy could you rebase this? I think it would be great if we could land this before any larger work on the integration tests, like #4357, happens.

**Comment by craftyguy** [comment]
@behrmann rebased, and got pmOS-related tests passing now

**Comment by martinpitt** [comment]
This fails opensuse's `test_extension[sysext]` on both [debian](https://github.com/systemd/mkosi/actions/runs/27593456992/job/81579049842?pr=4297) and [archi](https://github.com/systemd/mkosi/actions/runs/27593456992/job/81579049802?pr=4297), I haven't seen that breakage yet. That needs a closer look, smells like and actual regression? [this run](https://github.com/systemd/mkosi/actions/runs/27593456992/job/81579049749?pr=4297) could be an opensuse mirror flake, worth retrying (I can't trigger retries myself)

**Update**: The two failures above are the same reason, suse repo download timeouts/failures. :heavy_check_mark:

File: .github/workflows/ci.yml [hunk 1]
```diff
@@ -53,17 +53,22 @@ jobs:
           - debian
           - fedora
           - opensuse
```
martinpitt, line 55: nit: The "ci: enable postmarketOS in integration testing" commit should appear *last* in the series, not first (think "each commit on main should work"). All the others are preparation for making the pOS tests work.

.github/workflows/ci.yml
```diff
@@ -53,17 +53,22 @@ jobs:
...
+          - postmarketos
           - ubuntu
         tools:
           - arch
           - centos
           - debian
           - fedora
           - opensuse
+          - postmarketos
```
martinpitt: Is pOS actually meant to be a dev platform where you would run things like `mkosi` to build e.g. ubuntu? If not, testing with a pOS tools tree feels like a waste (we already run into serious github runner capacity issues). If so, then fine of course.

If this gets dropped, the excludes from below can be reduced.

mkosi/resources/mkosi-tools/mkosi.profiles/package-manager/mkosi.conf.d/postmarketos.conf
```diff
@@ -6,6 +6,7 @@ Distribution=postmarketos
 [Content]
 Packages=
         apt
+        archlinux-keyring
```
martinpitt, line 9: Similar question as above wrt. if pOS is meant to be a developer platform.

mkosi.tools.conf/mkosi.conf.d/postmarketos.conf
```diff
@@ -6,4 +6,5 @@ Distribution=postmarketos
 [Content]
 Packages=
         py3-codespell
+        py3-pytest
```
martinpitt, line 9: Can you please squash these micro-commits into logically related commits? Like grouping all the missing packages in the pOS tools tree.

mkosi/resources/mkosi-tools/mkosi.conf.d/postmarketos.conf
```diff
@@ -16,9 +16,11 @@ Packages=
         libfido2
         libseccomp
         losetup
+        mount
```
martinpitt: It would be helpful to add the "Otherwise busybox versions are used, which don't have the same features" from the commit message also as a comment here. Easier to understand in a few months.

action.yaml
```diff
@@ -97,9 +97,9 @@ runs:
       # upstream so mkosi can use it to populate the postmarketOS tools tree.
       shell: bash
       run: |
-        sudo curl -fsSL -o /usr/local/bin/apk https://gitlab.alpinelinux.org/api/v4/projects/5/packages/generic/v3.0.6/x86_64/apk.static
-        echo 'f1489e05bace7d7dd0a687fcd38d50b585ac660af4231668b123649bef3718c4  /usr/local/bin/apk' | sha256sum --check
-        sudo chmod +x /usr/local/bin/apk
+        sudo curl -fsSL -o /usr/bin/apk https://gitlab.alpinelinux.org/api/v4/projects/5/packages/generic/v3.0.6/x86_64/apk.static
+        echo 'f1489e05bace7d7dd0a687fcd38d50b585ac660af4231668b123649bef3718c4  /usr/bin/apk' | sha256sum --check
+        sudo chmod +x /usr/bin/apk
```
martinpitt, line 105: This is a bit evil -- `/usr` is package manager territory.  What's the error message? Can this be fixed with running with fixing `$PATH` somewhere, or calling the configure with `PATH=/usr/local/bin:$PATH` or so?

martinpitt (commented):
Thanks @craftyguy ! I have some questions, and the commit series could use some squashing and reordering, but by and large this looks great!

action.yaml
behrmann: `apk not found` sounds a bit like a PATH issue?

craftyguy: yeah I thought maybe since this is ephemeral CI that it might be "ok" to just toss it into /usr/bin 🤣

It's almost certainly a PATH issue, I just need to figure out which 'layer' of mkosi it needs to be set, which is kinda painful to do since I cannot reproduce this locally and have to debug it in the wonderful github CI.

.github/workflows/ci.yml
craftyguy: In theory, yes, it's a general purpose distro that happens to run on phones (among many other things). In practice, I have no idea who is doing dev work on other distros using mkosi on pmOS. We do have users running other distros in containers with podman, docker, and VMs.

I enabled these because it's what was done for other distros, but I'd be OK with removing them if it makes your CI load lighter and re-adding them IIF people start reporting issues e.g. building ubuntu on pmOS with mkosi.

wdyt?

action.yaml
craftyguy: was adding some debug prints to ci.yml and noticed this... https://github.com/systemd/mkosi/blob/main/.github/workflows/ci.yml#L116

So I guess /usr/local is being deleted to "free space"

craftyguy: "solved" it by moving the apk install step from actions.yml so that it can happen after the "free space" step in ci.yaml.

martinpitt: Ah, I see! Sorry, that moving doesn't work. `actions.yml` is meant to be reusable from other projects which use mkosi to build their images in a workflow. So I retract my objection, it's not more important than all that fallout. Please just install it into `/usr` then. Sorry for the detour!

.github/workflows/ci.yml
```diff
@@ -122,6 +139,16 @@ jobs:
         run: sudo ln -svf $PWD/bin/mkosi /usr/bin/mkosi
         working-directory: ./
 
+      - name: Install apk
```
martinpitt: Sorry, that moving doesn't work. actions.yml is meant to be reusable from other projects which use mkosi to build their images in a workflow. So I retract my objection, it's not more important than all that fallout. Please just install it into /usr then. Sorry for the detour!

.github/workflows/ci.yml
martinpitt: I'm on the fence, and as I haven't yet earned the right to make large directional decisions for mkosi: @behrmann @daandemeyer WDYT?

martinpitt:
I had another look. If you revert the moving of `apk` install and put it back into actions.yml in /usr, I have no further objections. (Pending the "tools test or not" question from the maintainers above). Thanks!

.github/workflows/ci.yml
craftyguy: Oh no! lol ok I'll remove this patch, and re-add my other one that dumps it into /usr/bin (along with a comment about "why?")

.github/workflows/ci.yml
behrmann: Generally, I think, that a lot of the matrix of our build tests covers test cases people will probably never do, since e.g. I assume there's little overlap between Arch and CentOS target audiences. I kinda don't want to make pmOS a second class citizen, but it's also clear that this grows quadratically, which isn't great either.

That being said. Even if we only test pmOS as a target from various distros and don't build other distros from pmOS, I really would always want to see that pmOS builds from a pmOS tools tree.

behrmann:
Modulo the concerns about the size of the build matrix this LGTM.

.github/workflows/ci.yml
craftyguy: > Even if we only test pmOS as a target from various distros and don't build other distros from pmOS, I really would always want to see that pmOS builds from a pmOS tools tree.

Sounds good to me, and we can always adjust later if necessary. So IIUC, only enable:

- build pmOS on pmOS
- build pmOS on everything

And I would *not* enable:

- build everything on pmOS

> I kinda don't want to make pmOS a second class citizen

I (a pmOS dev) definitely don't feel like a second class citizen here, so don't worry :P

craftyguy: @behrmann  ok I think I got this implemented. Now pmOS is built in other distro tools trees (that have apk available) and no distros are built with a pmOS tools tree (other than pmOS). I had to add to the "include:" list to get pmOS building with pmOS tools tree.

mkosi/resources/mkosi-tools/mkosi.profiles/package-manager/mkosi.conf.d/postmarketos.conf
craftyguy: given the other thread/question above, now that we are *not* explicitly testing "build Arch on pmOS tools tree", should I drop this or is there any harm in leaving it even if this combo isn't tested in CI? It would save someone from having to add it later.

martinpitt: This can be kept IMO.

martinpitt (approved):
Excellent! I'm looking into [these](https://github.com/systemd/mkosi/actions/runs/27721694120/job/82008778934?pr=4297) ( :arrow_right: #4363) [two](https://github.com/systemd/mkosi/actions/runs/27721694120/job/82008778902?pr=4297) ( :arrow_right:  #4362 ) failures. They are unrelated to this PR, but annoying.

action.yaml
```diff
@@ -95,11 +95,14 @@ runs:
...
       # upstream so mkosi can use it to populate the postmarketOS tools tree.
+      #
+      # NOTE: apk is dropped into /usr/bin and not /usr/local/bin because later CI
+      # scripts (like .github/workflows/ci.yml) clean up /usr/local
       shell: bash
       run: |
-        sudo curl -fsSL -o /usr/local/bin/apk https://gitlab.alpinelinux.org/api/v4/projects/5/packages/generic/v3.0.6/x86_64/apk.static
-        echo 'f1489e05bace7d7dd0a687fcd38d50b585ac660af4231668b123649bef3718c4  /usr/local/bin/apk' | sha256sum --check
-        sudo chmod +x /usr/local/bin/apk
+        sudo curl -fsSL -o /usr/bin/apk https://gitlab.alpinelinux.org/api/v4/projects/5/packages/generic/v3.0.6/x86_64/apk.static
```
behrmann, line 103: You didn't introduce this, but I wonder whether we have a good way of good way of keeping this up to date.

behrmann (approved):
Lovely, thank you! One last question

action.yaml
craftyguy: The best idea I have is to use the gitlab API to get the latest release of apk-tools, get the checksum for it, then download/verify the apk.static binary for x86_64. the API is json, so it would be a handful of curl | jq lines.

If that seems like an OK approach, I can take a shot at implementing it

behrmann: That would be lovely, but that's for a later PR.

craftyguy: ok I gave it a go over here: https://github.com/systemd/mkosi/pull/4364

PR #4300, opened 2026-05-14
box: Drop background tinting

daandemeyer:
Now that we have $SHELL_PROMPT_PREFIX integration,
let's drop the background tinting. Especially since it
doesn't quite nest all that well yet and we often invoke
nspawn and such from within mkosi box which does its own
background tinting.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -80,7 +80,6 @@ from mkosi.config import (
     parse_config,
     resolve_deps,
     summary,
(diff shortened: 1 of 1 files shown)


mkosi/__init__.py
behrmann: This import needs to be dropped to make ruff happy.

behrmann (approved):
This doesn't touch qemu, but since that's on the chopping block anyway this should also close #4294

PR #4302, opened 2026-05-15
Add ansi module for ANSI definitions.

jakobkukla:
Also removes duplicate ANSI definitions which were identical to `ANSI_BOLD` and `ANSI_RESET` respectively, but used `os.isatty(2)` instead of `terminal_is_dumb()` for terminal detection.

### DIFF
--- a/mkosi/ansi.py
+++ b/mkosi/ansi.py
@@ -1,0 +1,16 @@
+import os
+import sys
+
+
+def terminal_is_dumb() -> bool:
+    return not sys.stdout.isatty() or not sys.stderr.isatty() or os.getenv("TERM", "") == "dumb"
+
+
+# fmt: off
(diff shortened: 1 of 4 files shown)


daandemeyer:
sandbox.py needs to be standalone and can't depend on any other modules from mkosi, hence we can't merge this as is.

jakobkukla:
For the reason given in the docstring at the beginning?

daandemeyer:
Yeah we mount it into sandboxes and such. It needs to be standalone so we can rely on being able to run it after mounting it into a sandbox without having to figure out which files it imports and also mounting those.

behrmann:
@jakobkukla Thanks for the contribution, but for the reason discussed above I'm closing this. Feel free to resubmit if you have an ansatz that is compatible with sandbox needing to be standalone.

PR #4308, opened 2026-05-17
mkosi-initrd: Also add cryptsetup-libs explicitly to the initrd

daandemeyer:
And fix the name of the package on opensuse in the tools tree.

### DIFF
--- a/mkosi/resources/mkosi-initrd/mkosi.conf.d/azure-centos-fedora.conf
+++ b/mkosi/resources/mkosi-initrd/mkosi.conf.d/azure-centos-fedora.conf
@@ -11,6 +11,7 @@ Distribution=|azure
 [Content]
 Packages=
         # Various libraries that are dlopen'ed by systemd
+        cryptsetup-libs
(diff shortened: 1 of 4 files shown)


mkosi/resources/mkosi-tools/mkosi.conf.d/azure-centos-fedora/mkosi.conf
```diff
@@ -11,7 +11,7 @@ Distribution=|azure
 [Content]
 Packages=
         createrepo_c
-        cryptsetup-libs
+        libcryptsetup12
```
behrmann: This doesn't seem right, shouldn't that only be fore suse?

At least centos/centos fails on this, saying it doesn't know `libcryptsetup12`.

daandemeyer: woops fixed

behrmann (approved):

PR #4311, opened 2026-05-19
sandbox: fix console-script entrypoint not being able to execute commands

### LINKED ISSUE #4303: mkosi-sandbox console-script wrapper cannot execute commands

main

### Used host distribution

_No response_

### Used target distribution

_No response_

### Linux kernel version used

_No response_

None

`pyproject.toml` declares `mkosi-sandbox = "mkosi.sandbox:main"` as a
console-script entrypoint, but the resulting script cannot be used to execute
`mkosi-sandbox` with a command when installed with `pip install`.

`mkosi-sandbox` raises a `ValueError` when given a trailing
command unless `__name__ = __main__`, which is never true for the pip-generated
wrapper. This also leads to the `vm`, `shell` and `journalctl` subcommands
failing on NixOS (as I believe Nix internally uses the `pyproject.toml` script
entrypoints). See https://github.com/NixOS/nixpkgs/issues/510068 for the related issue.

Reproduction:
```
cd mkosi
python -m venv .venv && .venv/bin/pip install .
.venv/bin/mkosi-sandbox -- /bin/true
```

I think the core issue here is that `sandbox::main` is used for both: cli entrypoint and library function.
I would propose splitting these two responsibilities into two different functions. Introduce a `setup_sandbox(...)` library function and make `main` just the cli entrypoint which calls the library function internally. Possibly even pulling all the cli logic out into its own module (which does argument parsing and prints pretty errors). This would make the current `is_main` check redundant (solving this kind of bugs) and also provides better UX when used as a library function. I would be happy to contribute this change if that's desired.

### Used mkosi config

```ini

```

### mkosi output

```sh
  File "/home/jakob/Documents/mkosi/.venv/bin/mkosi-sandbox", line 6, in <module>
    sys.exit(main())
             ~~~~^^
  File "/home/jakob/Documents/mkosi/.venv/lib/python3.13/site-packages/mkosi/sandbox.py", line 1407, in main
    raise ValueError(f"A command line to execute can only be provided if {__name__} is executed")
ValueError: A command line to execute can only be provided if mkosi.sandbox is executed
```

jsrkamal2002:
When installed via pip/pipx/Nix, the mkosi-sandbox console-script calls main() directly, so __name__ is "mkosi.sandbox" rather than "__main__". This causes is_main() to return False, which raises a ValueError when a command is passed after -- and prevents os.execvp() from running. Fix this by introducing an entrypoint() wrapper that sets a module-level _entrypoint flag before calling main(), and update is_main() to also check that flag. Point the console-script at entrypoint() instead of main(). Library calls from run.py are unaffected since they call main() directly without going through entrypoint().
Fixes: https://github.com/systemd/mkosi/issues/4303

### DIFF
--- a/mkosi/sandbox.py
+++ b/mkosi/sandbox.py
@@ -167,9 +167,11 @@ This probably means either the system call is not implemented by the running ker
 system call is prohibited via seccomp if mkosi is being executed inside a containerized environment.\
 """
 
+_entrypoint: bool = False
+
 
 def is_main() -> bool:
-    return __name__ == "__main__"
+    return __name__ == "__main__" or _entrypoint
 
 
 def oserror(syscall: str, filename: str = "", errno: int = 0) -> None:
@@ -1573,6 +1575,12 @@ def main(argv: list[str] = sys.argv[1:]) -> None:
                 sys.exit(127)
 
             raise
+
+
+def entrypoint() -> None:
+    global _entrypoint
+    _entrypoint = True
+    main()
 
 
 if is_main():
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ bootable = [
 [project.scripts]
 mkosi = "mkosi.__main__:main"
 mkosi-initrd = "mkosi.initrd:main"
-mkosi-sandbox = "mkosi.sandbox:main"
+mkosi-sandbox = "mkosi.sandbox:entrypoint"
 mkosi-addon = "mkosi.addon:main"
 
 [tool.setuptools]


behrmann (changes requested):
So that doesn't work as advertised, I think your missing your actual entrypoint
```console
$ uv run --with-requirements <(echo git+https://github.com/systemd/mkosi@e25ce82ad39af3446315fa27064b7ed72498bac8) mkosi-sandbox bash
    Updated https://github.com/systemd/mkosi (e25ce82ad39af3446315fa27064b7ed72498bac8)
      Built mkosi @ git+https://github.com/systemd/mkosi@e25ce82ad39af3446315fa27064b7ed72498bac8
Installed 1 package in 25ms
  File "/home/user/.cache/uv/builds-v0/.tmp6PbhpQ/bin/mkosi-sandbox", line 4, in <module>
    from mkosi.sandbox import entrypoint
ImportError: cannot import name 'entrypoint' from 'mkosi.sandbox' (/home/user/.cache/uv/archive-v0/2_uwuIJT_ubOYao2/lib/python3.13/site-packages/mkosi/sandbox.py)
```

behrmann (approved):

behrmann:
@jsrkamal2002 Can you rebase this instead? We usually don't do back merges here.

daandemeyer:
I put up https://github.com/systemd/mkosi/pull/4315 as an alternative

daandemeyer:
Let's close since https://github.com/systemd/mkosi/pull/4315 was merged

jsrkamal2002:
Hi all,
Thank you for the updates on PR #4311.
I understand that the PR has been closed because Daan De Meyer's
alternative, #4315, was merged.
I also noted Jörg Behrmann's request to rebase instead of using back merges.
I will keep that in mind for future contributions.
Thanks for taking care of the fix.

Regards,
Raj Kamal

behrmann:
Thanks you for the bug report and the contribution, even though we went with the slightly larger rework! :)

PR #4313, opened 2026-05-19
apk: Implement repository_key_fetch for the postmarketOS distribution

daandemeyer:
When repository_key_fetch is enabled, fetch the alpine-keys and postmarketos-keys
packages with --allow-untrusted, extract them, and copy the trusted public keys
into /etc/apk/keys/ in the sandbox tree so subsequent installs verify normally.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/mkosi/distribution/postmarketos.py
+++ b/mkosi/distribution/postmarketos.py
@@ -8,6 +8,7 @@ from mkosi.distribution import Distribution, DistributionInstaller, PackageType
 from mkosi.installer import PackageManager
 from mkosi.installer.apk import Apk, ApkRepository
 from mkosi.log import complete_step, die
+from mkosi.tree import rmtree
 from mkosi.util import copyfile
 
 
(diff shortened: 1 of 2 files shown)


mkosi/distribution/postmarketos.py
```diff
@@ -36,6 +37,40 @@ def default_tools_tree_distribution(cls) -> Distribution:
...
+            # /usr/share/apk/keys/<arch>/, and their post-install script copies the arch-specific
+            # subset into /etc/apk/keys/. We do the same copy ourselves since post-install scripts
+            # don't run for extracted (uninstalled) packages.
+            for search in (extract / "usr/share/apk/keys" / arch, extract / "etc/apk/keys"):
+                if not search.is_dir():
+                    continue
+                for key in search.glob("*.rsa.pub"):
+                    dest = keys / key.name
+                    if dest.exists():
+                        continue
```
behrmann, line 68: I've not looked at the postinstall scriplets, but assuming they are shell, do they go to the extra length of not overwriting existing keys or why doesn't clobbering the files not make sense?

behrmann (commented):
Looks reasonable to me.

/cc @craftyguy so you're aware.

PR #4315, opened 2026-05-19
sandbox: split CLI and library entry points to fix console-script execution

### LINKED ISSUE #4303: mkosi-sandbox console-script wrapper cannot execute commands

daandemeyer:
When installed via pip/pipx/Nix, the mkosi-sandbox console-script calls
main() with __name__ == "mkosi.sandbox" rather than "__main__", so the
old is_main() check returned False and refused to execvp a trailing
command. Fix this by separating the two responsibilities: enter() is the
library function that sets up the sandbox in the current process and
returns any trailing argv, while main() is the CLI wrapper that calls
enter(), prints friendly error messages carried on a new SandboxOSError,
and execvp's the command. The console-script entry point points at
main() and works regardless of how it is invoked.

Fixes: https://github.com/systemd/mkosi/issues/4303

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/mkosi/sandbox.py
+++ b/mkosi/sandbox.py
@@ -168,14 +168,22 @@ system call is prohibited via seccomp if mkosi is being executed inside a contai
 """
 
 
-def is_main() -> bool:
-    return __name__ == "__main__"
+class SandboxOSError(OSError):
+    # OSError carrying a human-readable message that the CLI prints to stderr. Library callers can
+    # catch OSError as usual; only the CLI wrapper inspects .message.
+    def __init__(self, errno: int, message: str, filename: str = "") -> None:
+        super().__init__(errno, os.strerror(errno), filename or None)
+        self.message = message
 
 
 def oserror(syscall: str, filename: str = "", errno: int = 0) -> None:
     errno = abs(errno) or ctypes.get_errno()
-    if errno == ENOSYS and is_main():
-        print(ENOSYS_MSG.format(syscall=syscall, kver=os.uname().version), file=sys.stderr)
+    if errno == ENOSYS:
+        raise SandboxOSError(
+            errno,
+            ENOSYS_MSG.format(syscall=syscall, kver=os.uname().version),
+            filename,
+        )
 
     raise OSError(errno, os.strerror(errno), filename or None)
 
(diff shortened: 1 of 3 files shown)


mkosi/sandbox.py
```diff
@@ -1563,17 +1566,29 @@ def main(argv: list[str] = sys.argv[1:]) -> None:
...
+
+def main() -> None:
+    try:
+        argv = enter()
+    except SandboxOSError as e:
+        print(e.message, file=sys.stderr)
+        raise
+
+    if not argv:
+        argv = ["bash"]
```
behrmann:
```suggestion
    argv = argv or ["bash"]
```

behrmann (approved):
This looks very nice. Tested and works via uv.

PR #4316, opened 2026-05-19
fedora: allow Snapshot= for any kojipkgs-style mirror

mvo5:
The Snapshot= setting was gated to Release=rawhide and Mirror=https://kojipkgs.fedoraproject.org with two hard die() checks, but the URL builder underneath works for any release and any mirror that mimics the koji compose layout.

So this commit drops the restrictions and just assuems a koji style layout when snapshot is used. With that any koji style snapshot can be used.

### DIFF
--- a/mkosi/distribution/fedora.py
+++ b/mkosi/distribution/fedora.py
@@ -151,18 +151,14 @@ class Installer(DistributionInstaller, distribution=Distribution.fedora):
     def repositories(cls, context: Context) -> Iterable[RpmRepository]:
         gpgurls = find_fedora_rpm_gpgkeys(context)
 
-        if context.config.snapshot and context.config.release != "rawhide":
-            die(f"Snapshot= is only supported for rawhide on {cls.pretty_name()}")
-
         mirror = context.config.mirror
         if not mirror and context.config.snapshot:
             mirror = "https://kojipkgs.fedoraproject.org"
 
-        if context.config.snapshot and mirror != "https://kojipkgs.fedoraproject.org":
-            die(
-                f"Snapshot= is only supported for {cls.pretty_name()} if Mirror=https://kojipkgs.fedoraproject.org"
-            )
-
+        # Snapshot= and the koji compose URL layout work for any release and
+        # against any mirror that mimics the kojipkgs.fedoraproject.org tree.
+        # The inverse (kojipkgs without Snapshot=) is still rejected because
+        # that mirror has no usable non-snapshot URL surface.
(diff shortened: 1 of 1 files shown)


mkosi/distribution/fedora.py
```diff
@@ -183,22 +179,37 @@ def repositories(cls, context: Context) -> Iterable[RpmRepository]:
...
+            # carry post-release updates, so users pin via Snapshot= and rely
+            # on a live updates layer on top — keep that behaviour. Any other
+            # kojipkgs-compatible mirror is assumed to expose a *complete*
+            # compose per snapshot (e.g. an internal rpmrepo gateway that
+            # merges GA + updates at snapshot time), in which case the
+            # separate live-updates fetch is redundant and would 404.
+            if context.config.release != "rawhide" and (
+                not context.config.snapshot
+                or mirror == "https://kojipkgs.fedoraproject.org"
+            ):
```
daandemeyer: Are we sure we don't want to do the same for the kojipkgs mirror here?

daandemeyer (approved):

mkosi/distribution/fedora.py
```diff
@@ -183,22 +179,28 @@ def repositories(cls, context: Context) -> Iterable[RpmRepository]:
...
-                subdir = f"compose/{context.config.release}"
+            # Snapshot= implies the koji compose URL layout (compose/<release>/
+            # Fedora-<Release>-<snapshot>/compose/Everything/...), regardless
+            # of the actual mirror host, so any kojipkgs-compatible mirror
+            # (including our internal rpmrepo gateway) works.
+            if context.config.snapshot:
+                subdir = (
+                    f"compose/{context.config.release}/"
+                    f"Fedora-{context.config.release.capitalize()}-{context.config.snapshot}/compose"
+                )
```
behrmann, line 190: Nit: That seems a bit prone to break and is a bit harder to grep for, I'd just go for the overly long line and ignore ruff there
```suggestion
                subdir = f"compose/{context.config.release}/Fedora-{context.config.release.capitalize()}-{context.config.snapshot}/compose"  # noqa: E501
```

behrmann (approved):

mkosi/distribution/fedora.py
daandemeyer: Ehh I think the current form is fine

daandemeyer (approved):

mkosi/distribution/fedora.py
mvo5: Thanks! I like it. The PR got merged now, I can do a followup with just this tiny change if desired.

PR #4322, opened 2026-05-21
mkosi-initrd: Add libfdisk1

martinpitt:
systemd-repart otherwise crashes on Debian with

> libfdisk.so.1 is not available

For consistency, add it to the other OSes, too.

---

Fixes first boot after installing Debian:

### DIFF
--- a/mkosi/resources/mkosi-initrd/mkosi.conf.d/debian-ubuntu.conf
+++ b/mkosi/resources/mkosi-initrd/mkosi.conf.d/debian-ubuntu.conf
@@ -1,0 +1,8 @@
+[Match]
+Distribution=|debian
+Distribution=|ubuntu
+
+[Content]
+Packages =
+        # systemd-repart dependency
+        libfdisk1
--- a/mkosi/resources/mkosi-initrd/mkosi.conf.d/arch.conf
+++ b/mkosi/resources/mkosi-initrd/mkosi.conf.d/arch.conf
@@ -15,6 +15,9 @@ Packages=
         libseccomp
         tpm2-tss
 
+        # systemd-repart needs libfdisk.so
+        util-linux-libs
+
         procps-ng
         util-linux
 
--- a/mkosi/resources/mkosi-initrd/mkosi.conf.d/azure-centos-fedora.conf
+++ b/mkosi/resources/mkosi-initrd/mkosi.conf.d/azure-centos-fedora.conf
@@ -15,6 +15,9 @@ Packages=
         libseccomp
         tpm2-tss
 
+        # systemd-repart dependency
+        libfdisk
+
         # File system checkers for supported root file systems
         e2fsprogs
         xfsprogs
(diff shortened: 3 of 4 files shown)


daandemeyer:
Please add it everywhere to make things consistent

martinpitt:
@daandemeyer Updated. I didn't find that package in [postmarketos](https://pkgs.postmarketos.org/packages?name=*fdisk*&branch=main&repo=&arch=&origin=) or [arch](https://archlinux.org/packages/?sort=&q=fdisk&maintainer=&flagged=) though.

mkosi/resources/mkosi-initrd/mkosi.conf.d/debian-ubuntu.conf
```diff
@@ -0,0 +1,8 @@
+[Match]
+Distribution=|debian
+Distribution=|ubuntu
```
daandemeyer, line 3: Missing file for Arch Linux?

Also this file should be named debian-ubuntu.conf

martinpitt: Ack, will rename. As I said above, didn't find this package for arch, I need to check more closely. But that will have to wait until next week, $dayjob is *really* calling now . Moving to draft for now.

behrmann:
```
$ pacman -Ql | grep libfdisk
lib32-util-linux /usr/lib32/libfdisk.so
lib32-util-linux /usr/lib32/libfdisk.so.1
lib32-util-linux /usr/lib32/libfdisk.so.1.1.0
util-linux-libs /usr/include/libfdisk/
util-linux-libs /usr/include/libfdisk/libfdisk.h
util-linux-libs /usr/lib/libfdisk.so
util-linux-libs /usr/lib/libfdisk.so.1
util-linux-libs /usr/lib/libfdisk.so.1.1.0
```

martinpitt: Thanks @behrmann ! Added and renamed the file.

daandemeyer (approved):

behrmann:
Seems a bit weird, that postmarket doesn't have this package, even though alpine has it (libfdisk), but that's something we can add later.

@craftyguy Could you shed some light here maybe>

behrmann (approved):

craftyguy:
sorry I couldn't respond in time before this was merged, but in postmarketos/Alpine Linux it comes from `libfdisk`: https://pkgs.alpinelinux.org/contents?file=libfdisk.so.*&path=&name=&branch=edge&repo=&arch=x86_64


> that postmarket doesn't have this package, even though alpine has it (libfdisk), but that's something we can add later.

pmOS is like 85% Alpine Linux (the remaining diff is mostly systemd and other config), packages like libfdisk come from the Alpine binary repos :)

martinpitt:
@craftyguy ah! that explains why [that search didn't find it](https://pkgs.postmarketos.org/packages?name=*fdisk*&branch=main&repo=&arch=&origin=). Updated in #4328

PR #4329, opened 2026-05-22
finalize_scripts: tighten the PATH-strip condition to actual self-exec

katexochen:
The PATH-strip prelude exists to prevent `/scripts/<name>` from recursing into itself when the script execs a binary named `<name>` via PATH lookup. The current heuristic, `config.find_binary(name)`, checks a broader condition: whether *some* binary called `<name>` exists in the search path. That produces false positives for scripts that exec a different binary entirely.

In particular, `mkosi-install` execs `dnf install $@`, expecting `dnf` to resolve to the wrapped `/scripts/dnf` (which adds `--installroot=…`, `--use-host-config`, `--setopt=…`). When the prelude gets emitted for `mkosi-install` too, it strips `/scripts` from PATH before the exec, so `dnf` resolves to the raw system binary and is invoked with none of the wrapper's overrides.

Tighten the check: emit the prelude only when one of the exec'd argv tokens is the unqualified string `name`, which is the only case where PATH lookup could resolve back to `/scripts/<name>`.

Reproducer:

```
$ mkdir /tmp/repro && cd /tmp/repro
$ cat > mkosi.conf <<'EOF'
[Distribution]
Distribution=fedora
EOF
$ cat > mkosi.prepare <<'EOF'
mkosi-install --dump-main-config bash 2>&1 | grep '^installroot ' EOF
$ chmod +x mkosi.prepare
$ mkosi -ff build
```

Before: installroot = /
After: installroot = /buildroot

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -594,10 +594,21 @@ def finalize_scripts(config: Config, scripts: Mapping[str, Sequence[PathString]]
         for name, script in scripts.items():
             # Make sure we don't end up in a recursive loop when we name a script after the binary
             # it execs by removing the scripts directory from the PATH when we execute a script.
+            #
+            # Recursion is only possible when some unqualified token in the exec'd command line is
+            # exactly "name" -- in that case the shell would resolve it back to "/scripts/<name>"
+            # via PATH. Checking "config.find_binary(name)" instead is a broader condition: it
+            # triggers PATH-strip whenever the script name happens to exist somewhere in the
+            # search path, including for scripts that exec a different binary entirely. For
+            # example, "mkosi-install" execs "dnf install", but the strip causes the wrapped
+            # "/scripts/dnf" to be hidden from PATH, so the raw system dnf runs without the
+            # wrapper's "--installroot", "--use-host-config", and "--setopt=" flags.
+            same_name_recursion = any("/" not in (s := os.fspath(arg)) and s == name for arg in script)
+
             with (Path(d) / name).open("w") as f:
                 f.write("#!/bin/sh\n")
 
-                if config.find_binary(name):
+                if same_name_recursion:
                     f.write(
                         textwrap.dedent(
                             """\


mkosi/__init__.py
```diff
@@ -594,10 +594,21 @@ def finalize_scripts(config: Config, scripts: Mapping[str, Sequence[PathString]]
...
+            #
+            # Recursion is only possible when some unqualified token in the exec'd command line is
+            # exactly `name` -- in that case the shell would resolve it back to `/scripts/<name>`
+            # via PATH. Checking `config.find_binary(name)` instead is a broader condition: it
+            # triggers PATH-strip whenever the script name happens to exist somewhere in the
+            # search path, including for scripts that exec a different binary entirely. For
+            # example, `mkosi-install` execs `dnf install`, but the strip causes the wrapped
+            # `/scripts/dnf` to be hidden from PATH, so the raw system dnf runs without the
+            # wrapper's `--installroot`, `--use-host-config`, and `--setopt=` flags.
+            same_name_recursion = any("/" not in (s := os.fspath(arg)) and s == name for arg in script)
```
behrmann, line 606: So, `script` here is a `Sequence[PathString]` that's the command line we'll exec. My question is, why do we need to look at all arguments (that don't contain a `/`), but all of them? I probably missed something, but the only thing I can see where this would be relevant, is ukify, which is being called with the python interpreter as arg0. But in that case this wouldn't make a difference, no?

mkosi/__init__.py
```diff
@@ -594,10 +594,21 @@ def finalize_scripts(config: Config, scripts: Mapping[str, Sequence[PathString]]
...
             # it execs by removing the scripts directory from the PATH when we execute a script.
+            #
+            # Recursion is only possible when some unqualified token in the exec'd command line is
+            # exactly `name` -- in that case the shell would resolve it back to `/scripts/<name>`
+            # via PATH. Checking `config.find_binary(name)` instead is a broader condition: it
+            # triggers PATH-strip whenever the script name happens to exist somewhere in the
+            # search path, including for scripts that exec a different binary entirely. For
+            # example, `mkosi-install` execs `dnf install`, but the strip causes the wrapped
+            # `/scripts/dnf` to be hidden from PATH, so the raw system dnf runs without the
+            # wrapper's `--installroot`, `--use-host-config`, and `--setopt=` flags.
```
behrmann: Nit: This is not markdown
```suggestion
            # exactly "name" -- in that case the shell would resolve it back to "/scripts/<name>"
            # via PATH. Checking "config.find_binary(name)" instead is a broader condition: it
            # triggers PATH-strip whenever the script name happens to exist somewhere in the
            # search path, including for scripts that exec a different binary entirely. For
            # example, "mkosi-install" execs "dnf install", but the strip causes the wrapped
            # "/scripts/dnf" to be hidden from PATH, so the raw system dnf runs without the
            # wrapper's "--installroot", "--use-host-config", and "--setopt=" flags.
```

behrmann (commented):
Really great find! Nice work.

mkosi/__init__.py
katexochen: The dnf wrapper expands to roughly `python3 -SI /sandbox.py ... env VAR=val ... dnf --installroot=/buildroot ... "$@"`, based on 

```
    def scripts(cls, context: Context) -> dict[str, list[PathString]]:
        return {
            "dnf": cls.apivfs_script_cmd(context) + cls.env_cmd(context) + cls.cmd(context),
```

The shell resolves arg0 (python3) but sandbox.py later does `execpv("env", ...)` and env later does its own `execvp("dnf", ...)`, and if /scripts is still in PATH at this point we re-enter the wrapper again.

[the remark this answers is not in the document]
katexochen: Agreed, that's a valid concern. But as described, while we would be doing unnecessary work, it wouldn't cause harm. Checking only script[0] won't suffice due to the exec nesting.

katexochen:
I thought about this a bit more. While this patch is a strict improvement, it isn't perfect.

[Claude pointed out](https://github.com/systemd/mkosi/pull/4329#discussion_r3288109385):
There could be cases where we find some arg that matches the script name, but is actually an argument and we would never try to actually resolve it via PATH further down the exec chain.

Other thing that came to my mind:
There could be multiple arguments that should actually resolve to /scripts. If we detect a loop, we remove /scripts from PATH, and as a result, other args might not be resolved correctly. Example:

  `/scripts/foo` invokes `foo dnf`, where `dnf` should resolve to `/scripts/dnf`.
  We detect `foo` might loop back to` /scripts/foo`.
  We remove `/scripts` from `PATH`.
   `dnf` now resolves to something else than `/scripts/dnf`, or nothing at all.
   
My instant idea was to instantly resolve `foo` in the example from a `PATH` that doesn't contain `/scripts`, but keep `/scripts` in the `PATH` for the actual invocation. However, that wouldn't work out well if we hit a an arg that isn't actually meant to be resolved via PATH (claude's point from before).

behrmann:
> this patch is a strict improvement,

Agreed.

> There could be cases where we find some arg that matches the script name

That was also the main point of my question, since I figured only doing this with argv0 could sidestep the issue.

> My instant idea was to instantly resolve foo in the example from a PATH that doesn't contain /scripts, but keep /scripts in the PATH for the actual invocation. However, that wouldn't work out well if we hit a an arg that isn't actually meant to be resolved via PATH (claude's point from before).

I'm not sure this is sensibly solvable algorithmically. Maybe we could add some token (`MKOSI_RESOLVE_IN`) in our wrappers to document what we actually mean and resolve it correctly depending on that.

PR #4333, opened 2026-05-26
Mount /etc/resolv.conf symlink into sandbox

daandemeyer:
Currently, if /etc/resolv.conf is a symlink, we
bind mount the actual file it points to into the 
sandbox. Problem with this approach is that if the 
file it points to is replaced on the host, creating
nested sandboxes will fail because bind mounting a 
file whose source does not exist anymore fails with
ENOENT when you call mount. 

To fix this (partially), make sure we bind mount the 
/etc/resolv.conf symlink if it one. That way, on
resolved systems, we'll bind mount the /etc/resolv.conf
symlink and the /run/systemd/resolve directory into the
sandbox, with the latter containing the symlink target of
/etc/resolv.conf. Because we don't mount the target file
directly but its parent directory, if the file is replaced,
the sandbox will see the new file.

To make this work we add new --bind-nofollow and 
--ro-bind-nofollow options to mkosi-sandbox.

### DIFF
--- a/mkosi/resources/man/mkosi-sandbox.1.md
+++ b/mkosi/resources/man/mkosi-sandbox.1.md
@@ -77,6 +77,13 @@ from `mkosi.sandbox` is not supported and may break in future versions.
 `--ro-bind-foreign SRC DST`
 :   Like `--bind-foreign`, but does a readonly mount.
 
+`--bind-nofollow SRC DST`
+:   Like `--bind`, but does not follow symlinks for `SRC`. If `SRC` is itself a symlink, the
+    symlink is bind mounted as-is rather than its target.
+
+`--ro-bind-nofollow SRC DST`
+:   Like `--bind-nofollow`, but does a readonly mount.
+
 `--symlink SRC DST`
 :   Creates a symlink at `DST` in the sandbox pointing to `SRC`. If `DST` already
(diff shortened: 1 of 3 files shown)


mkosi/resources/man/mkosi-sandbox.1.md
```diff
@@ -77,6 +77,14 @@ from `mkosi.sandbox` is not supported and may break in future versions.
 `--ro-bind-foreign SRC DST`
 :   Like `--bind-foreign`, but does a readonly mount.
 
+`--bind-nofollow SRC DST`
+:   Like `--bind`, but does not follow symlinks for `SRC`. If `SRC` is itself a symlink, the
+    symlink is bind mounted as-is rather than its target. This is useful when the symlink at
+    `SRC` may change on the host and the sandbox should resolve it at access time instead.
```
behrmann: Just to be sure: From the commit message this is more about the target of `SRC` changing, not `SRC` itself changing, no?

behrmann (approved):

mkosi/resources/man/mkosi-sandbox.1.md
daandemeyer: I'm just gonna drop that part

PR #4335, opened 2026-05-29
tests: Move unit tests from GitHub workflow into pytest

martinpitt:
Extract all CI linter/formatter/install checks from the GitHub workflow
into pytest tests. Workflows are hard to reproduce locally, and it's
much more convenient to reproduce what CI does with a simple `pytest`
command. Update the documentation and also stop recommending locally
installed `mypy` and friends -- let's always use the version from
`mkosi box` to get something halfway reproducible.

Keep the `sudo mkosi -h` test in the workflow, though -- it cannot run
inside `mkosi box` as that doesn't have `sudo`.

---

I'm still not 100% happy about this, e.g. it's mysterious what the `- uses: ./` step does. I'd also like to get rid of the `Install` step, but I have a gut feeling it interacts with the `uses: ./` magic. But anyway, this is enough to make the linter failures reproducible and consistent. I'd like to get your (in particular @daandemeyer 's) opinion what you think about the general direction.

After this, I'd like to do the same treatment to the integration test (I didn't yet try to run them locally), and then finally towards actually fixing the two current failures (arch distro+debian tools, and arch+fedora).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,10 +2,6 @@
 *.cache-pre-dev
 *.cache-pre-inst
 .cache
-.mkosi.1
-.mkosi-addon.1
-.mkosi-initrd.1
-.mkosi-sandbox.1
 .mypy_cache/
 .project
 .pydevproject
@@ -14,6 +10,8 @@
 /SHA256SUMS
 /SHA256SUMS.gpg
 /build
+# Generated zipapp
+/builddir/
 /dist
 /mkosi.build
 /mkosi.egg-info
@@ -23,6 +21,7 @@
 /mkosi.rootpw
 mkosi.local
 mkosi.local.conf
+mkosi/resources/man/*.[17]
 mkosi.tools
 mkosi.tools.manifest
 /mkosi.key
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -235,6 +235,7 @@ def run(
     success_exit_status: Sequence[int] = (0,),
     setup: Sequence[PathString] = (),
     sandbox: AbstractContextManager[Sequence[PathString]] = nosandbox(),
+    cwd: Optional[PathString] = None,
 ) -> CompletedProcess:
     if input is not None:
         assert stdin is None  # stdin and input cannot be specified together
@@ -251,6 +252,7 @@ def run(
         success_exit_status=success_exit_status,
         setup=setup,
         sandbox=sandbox,
+        cwd=cwd,
     ) as process:
         out, err = process.communicate(input)
 
@@ -305,6 +307,7 @@ def spawn(
     success_exit_status: Sequence[int] = (0,),
     setup: Sequence[PathString] = (),
     sandbox: AbstractContextManager[Sequence[PathString]] = nosandbox(),
+    cwd: Optional[PathString] = None,
 ) -> Iterator[Popen]:
     cmd = [os.fspath(x) for x in cmdline]
 
@@ -362,6 +365,7 @@ def spawn(
                 group=group,
                 pass_fds=pass_fds,
                 env=env if not sbx or not apply_sandbox_in_preexec else None,
+                cwd=os.fspath(cwd) if cwd is not None else None,
                 preexec_fn=(
                     functools.partial(
                         _preexec,
--- a/mkosi/sandbox.py
+++ b/mkosi/sandbox.py
@@ -880,6 +880,10 @@ def userns_acquire_empty() -> int:
 
 
 def chase(root: str, path: str, *, nofollow: bool = False) -> str:
+    # pyright hack around `reportPossiblyUnboundVariable`; it doesn't understand
+    # that it's defined/used only if `nofollow` is True
+    parent = ""
+    base = ""
     if nofollow:
         parent = os.path.dirname(path) or "/"
         base = os.path.basename(path)
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -1,0 +1,65 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+import os
+import tempfile
+from pathlib import Path
+
+from mkosi.run import run
+
+REPO_ROOT = Path(__file__).parent.parent
+
+
+def test_mkosi_help_direct() -> None:
+    """Test mkosi can be run from current directory."""
+    run(["python3", "-m", "mkosi", "-h"], cwd=REPO_ROOT)
+
+
+def test_venv_installation() -> None:
+    """Test mkosi can be installed in a venv."""
+    with tempfile.TemporaryDirectory() as tmpdir:
+        venv = Path(tmpdir) / "testvenv"
+
+        # Create venv
+        run(["python3", "-m", "venv", os.fspath(venv)], cwd=REPO_ROOT)
+        pip = venv / "bin/python3"
+
+        # Upgrade pip, setuptools, wheel
+        run(
+            [os.fspath(pip), "-m", "pip", "install", "--upgrade", "setuptools", "wheel", "pip"],
+            cwd=REPO_ROOT,
+        )
+
+        # Install mkosi
+        run([os.fspath(pip), "-m", "pip", "install", "."], cwd=REPO_ROOT)
+
+        # Test that mkosi works
+        run([os.fspath(venv / "bin/mkosi"), "-h"], cwd=REPO_ROOT)
+
+
+def test_editable_venv_installation() -> None:
+    """Test mkosi can be installed in editable mode."""
+    with tempfile.TemporaryDirectory() as tmpdir:
+        venv = Path(tmpdir) / "testvenv"
+
+        run(["python3", "-m", "venv", os.fspath(venv)], cwd=REPO_ROOT)
+        pip = venv / "bin/python3"
+
+        # Upgrade pip, setuptools, wheel
+        run(
+            [os.fspath(pip), "-m", "pip", "install", "--upgrade", "setuptools", "wheel", "pip"],
+            cwd=REPO_ROOT,
+        )
+
+        # Install mkosi in editable mode
+        run([os.fspath(pip), "-m", "pip", "install", "--editable", "."], cwd=REPO_ROOT)
+
+        # Test that mkosi works
+        run([os.fspath(venv / "bin/mkosi"), "-h"], cwd=REPO_ROOT)
+
+
+def test_zipapp_creation() -> None:
+    """Test zipapp generation."""
+    run(["./tools/generate-zipapp.sh"], cwd=REPO_ROOT)
+
+    run(["./builddir/mkosi", "-h"], cwd=REPO_ROOT)
+    run(["./builddir/mkosi", "documentation"], cwd=REPO_ROOT)
--- a/tests/test_linters.py
+++ b/tests/test_linters.py
@@ -1,0 +1,90 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+import os
+import shutil
+import subprocess
+from pathlib import Path
+
+import pytest
+
+from mkosi.run import run
+
+# tolerate missing tools in some random dev environment, but enforce them in CI/mkosi box
+SKIP_MISSING_TOOLS = "MKOSI_IN_BOX" not in os.environ
+
+REPO_ROOT = Path(__file__).parent.parent
+
+
+def kernel_install_files() -> list[str]:
+    """Get list of kernel-install/*.install files."""
+    return [os.fspath(p) for p in (REPO_ROOT / "kernel-install").glob("*.install")]
+
+
+def _skip_if_missing(tool: str) -> bool:
+    """Return True if we should skip the test because tool is missing."""
+    return SKIP_MISSING_TOOLS and shutil.which(tool) is None
+
+
+@pytest.mark.skipif(_skip_if_missing("ruff"), reason="ruff not found")
+def test_ruff_format_check() -> None:
+    """Check that code is formatted with ruff format."""
+    run(["ruff", "format", "--check", "--diff", "mkosi/", "tests/", *kernel_install_files()], cwd=REPO_ROOT)
+
+
+@pytest.mark.skipif(_skip_if_missing("ruff"), reason="ruff not found")
+def test_ruff_check() -> None:
+    """Check code quality with ruff."""
+    run(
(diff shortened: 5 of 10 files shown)


daandemeyer (changes requested):
Let's use `run()` from mkosi.run rather than raw `subprocess`

General approach is fine, though same applies as particleos, I want to switch mkosi to barrage sooner rather than later (damn fibers in systemd keeping me busy). For now moving these to pytest is fine though

mkosi/sandbox.py
```diff
@@ -880,6 +880,8 @@ def userns_acquire_empty() -> int:
 
 
 def chase(root: str, path: str, *, nofollow: bool = False) -> str:
+    parent = ""
+    base = ""
```
behrmann, line 886: With what version of pytest is that? I don't see it with `pyright 1.1.410`

tests/test_install.py
```diff
@@ -0,0 +1,62 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+import subprocess
+import tempfile
+from pathlib import Path
+
+
+def test_mkosi_help_direct() -> None:
+    """Test mkosi can be run from current directory."""
+    subprocess.check_call(["python3", "-m", "mkosi", "-h"])
```
behrmann: All of these checks do assume that the current working directory is the root of the repo. Is this ensured in some way?

Also, having this in a test framework, it would be nice to skip the tests instead of failing if something like mypy or pyright is not installed.

tests/test_install.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def test_venv_installation() -> None:
+    """Test mkosi can be installed in a venv."""
+    with tempfile.TemporaryDirectory() as tmpdir:
+        venv = Path(tmpdir) / "testvenv"
+
+        # Create venv
+        subprocess.check_call(["python3", "-m", "venv", str(venv)])
```
behrmann: Nit: (below too)
```suggestion
        subprocess.check_call(["python3", "-m", "venv", os.fspath(venv)])
```

When coercing Path-like objects to strings, use `os.fspath`, since this calls the `__fspath__` protocol
  instead of `__str__`. It also ensures more type-safety, since every Python object supports `__str__`, but
  not all support `__fspath__` and this gives the typechecker more information what is expected at this
  point. It also signals the intent to the reader more than a blanket `str()`.

tests/test_install.py
```diff
@@ -0,0 +1,62 @@
...
+
+        pip = venv / "bin" / "python3"
```
behrmann: Nit:
```suggestion
        pip = venv / "bin/python3"
```
It's a bit shorter and has a few fewer unnecessary glyphs. It does assume that `/` is the path separator, but we do so all over, it's not like we're gonna run this on Windows.

martinpitt:
@daandemeyer understood, changed to `run()`. I also dropped the `_in_box()` wrapper, that was a thinko -- the whole unit test already runs inside the box.

tests/test_linters.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def _in_box(cmd: list[str]) -> list[str]:
+    """Wrap command to run in mkosi box."""
+    return ["bin/mkosi", "box", "--", *cmd]
+
+
+def _kernel_install_files() -> list[str]:
+    """Get list of kernel-install/*.install files."""
+    return glob.glob("kernel-install/*.install")
```
behrmann: We ship most of our resources in `mkosi` itself. It might make sense to bundle the kernel-install files as well.

mkosi/sandbox.py
martinpitt:
```
❱❱❱ bin/mkosi box -- pyright --version
pyright 1.1.410

❱❱❱ bin/mkosi box -- pytest --version
pytest 9.0.3
```

This is with the (default) Fedora box, i.e. running `bin/mkosi -f box -- true` *without* a `mkosi.local.conf`. I don't have `pacman` installed locally, so I can't create an Arch one like in CI, and running the tests inside of `distrobox` doesn't work either because of `mkosi.sandbox.VarlinkError: io.systemd.NamespaceResource.DynamicRangeUnavailable: {}`.

However, I can take out that commit. It's not going to break CI, "just" local test runs.

behrmann: That's weird, since you've got the same pyright version. 

I'm wondering whether pyright is not picking up something properly, but that wouldn't really explain why it works in CI already.

I'd say leave it in, I think it's sensible to be able to run this on other distros with stock tools. Add a comment though, that it is a pyrightism. I'm hopeful we can get rid off pyright in the not too distant future.

tests/test_install.py
martinpitt: The tests run in `mkosi box`. Personally I prefer failing the tests over silently skipping them, as otherwise you very easily end up with accidentally skipping tests in CI if the box config changes. That decision is up to you, of course.

How do you run tests outside of the repo root? If I try e.g. `cd mkosi; ../bin/mkosi box -- pytest` it fails right away with `pytest not found`, so I guess there's some extra magic/options/path/env settings? Thanks!

tests/test_linters.py
martinpitt: I'm sorry, I don't know what you mean. Still an mkosi beginner here :see_no_evil:

mkosi/sandbox.py
martinpitt: Added a "hack" comment.

tests/test_install.py
martinpitt: TIL, thanks! Done throughout.

tests/test_install.py
martinpitt: Done.

tests/test_install.py
behrmann: > The tests run in mkosi box. Personally I prefer failing the tests over silently skipping them, as otherwise you very easily end up with accidentally skipping tests in CI if the box config changes. That decision is up to you, of course.

Yes, that makes sense. The direction I'm coming from, I really like being able to run a project's tests as much as possible. If I run tests and they crash because I don't have e.g. some other specific tool installed, I consider that an unnecessary roadblock (in a lot of cases).

We sidestep this mostly with mkosi box, but I'd like to be able to run the tests outside of mkosi box as well.

I think we can split the difference. Both GitHub actions but also other CI runners (like GitLab) set the `CI` variable. I'd wager this one is pretty standard nowadays, so a check could be

```python
if not (shutil.which("some_test_binary") or os.getenv("CI")):
    pytest.skip("some_test_binary is missing")
```

> How do you run tests outside of the repo root? 

You can run `pytest path/to/repo` or `mypy path/to/repo/mkosi`. I actually do do that sometimes (when I open a fresh terminal), although I grant that might be niche. I haven't checked, but I'd assume the cwd for the former pytest invocation to not be the repo root. 

Maybe I'm being to paranoid here.

tests/test_linters.py
behrmann: Actually not a mkosi thing. We ship most resources in `mkosi/resources`, which makes them accesible via `importlib.resources`. This question was mostly born from the above question of the working directory. It might not be worth it depending on how this goes, i.e. if we say you have to be in the top-level dir of the repo to run the tests, but it would be a way to reference the files from inside the tests.

tests/test_install.py
martinpitt: wrt. the skips: How about `which("mypy") and "MKOSI_IN_BOX" not in os.environ`? That has a similar behaviour, but less "magic" wrt. GitHub actions, but under control of `mkosi box`, and should make both of us happy. WDYT?

Wrt. calling from a different dir, I'll look at that next week (still have some $work things to do today). Should not be hard to support surely, with `__file__`.

behrmann: Yes, checking for `MKOSI_IN_BOX` is surely the safest.

martinpitt: I implemented the test skips as discussed above. Will push in a bit.

Wrt. different directories: this doesn't work with `box`, as the "run in target project" assumption is already built in:

```
❱❱❱ ~/upstream/mkosi/bin/mkosi box -- pytest upstream/mkosi/
‣ pytest not found.
```

(happens on main). But it does work without box, i.e. "random dev env", `pytest -k lint upstream/mkosi/tests/` from my home dir works on main, but indeed fails the new linter/install tests.

I updated the tests to pass a `cwd` kwarg to `run()`, now it works as you expect. Tested with 

```
❱❱❱ cd ~; distrobox enter sid -- pytest -v upstream/mkosi/tests/
[...]
upstream/mkosi/tests/test_linters.py::test_ruff_format_check SKIPPED (ruff not found)                            [ 42%]
upstream/mkosi/tests/test_linters.py::test_ruff_check SKIPPED (ruff not found)                                   [ 43%]
upstream/mkosi/tests/test_linters.py::test_no_tabs_in_code PASSED                                                [ 43%]
upstream/mkosi/tests/test_linters.py::test_codespell SKIPPED (codespell not found)                               [ 44%]
upstream/mkosi/tests/test_linters.py::test_reuse_lint SKIPPED (reuse not found)                                  [ 44%]
upstream/mkosi/tests/test_linters.py::test_mypy PASSED                                                           [ 45%]
upstream/mkosi/tests/test_linters.py::test_mypy_python310 PASSED                                                 [ 45%]
upstream/mkosi/tests/test_linters.py::test_pyright SKIPPED (pyright not found)                                   [ 46%]
upstream/mkosi/tests/test_linters.py::test_shellcheck SKIPPED (shellcheck not found)                             [ 47%]
upstream/mkosi/tests/test_linters.py::test_man_page_generation SKIPPED (pandoc not found)                        [ 47%]
[...]
```

My sid container has mypy, but none of the others.

tests/test_linters.py
martinpitt: Fixed as part of the above to make the tests work when run from any dir.

tests/test_linters.py
```diff
@@ -0,0 +1,98 @@
...
+
+# tolerate missing tools in some random dev environment, but enforce them in CI/mkosi box
+SKIP_MISSING_TOOLS = "MKOSI_IN_BOX" not in os.environ
+
+REPO_ROOT = Path(__file__).parent.parent
+
+
+def _kernel_install_files() -> list[str]:
+    """Get list of kernel-install/*.install files."""
+    return glob.glob(os.fspath(REPO_ROOT / "kernel-install/*.install"))
```
behrmann: Since you're already importing pathlib, you can omit the glob import

```suggestion
    return [os.fspath(p) for p in (REPO_ROOT / "kernel-install).glob(*.install")]
```

mkosi/run.py
```diff
@@ -362,6 +365,7 @@ def spawn(
                 group=group,
                 pass_fds=pass_fds,
                 env=env if not sbx or not apply_sandbox_in_preexec else None,
+                cwd=os.fspath(cwd) if cwd is not None else None,
```
behrmann, line 368: I feel a bit uneasy, because an empty string for `cwd` would lead to `FileNotFoundError` here.
```suggestion
                cwd=os.fspath(cwd) if cwd else None,
```
would solve this, but this is generally just a shortcoming of the API and I'm not sure whether this is worthwhile to do. It is a bit shorter though.

behrmann (approved):

mkosi/run.py
martinpitt: I find that getting an ENOENT for `""` is correct. `cwd=""` looks like an error, and a type confusion. If you don't mind, I'd rather keep it as-is.

tests/test_linters.py
martinpitt: Yes, good idea, done. I also changed the other place, so the `glob` import is gone now.

martinpitt:
@daandemeyer you still have a :-1: on this (use `mkosi.run()`), I did that. Can you PTAL?

mkosi/sandbox.py
martinpitt: @behrmann @davide125 This also [affects main now](https://github.com/systemd/mkosi/actions/runs/26755419982/job/78853917143?pr=4338#step:12:7), so this "spread" to arch.

tests/test_install.py
```diff
@@ -0,0 +1,59 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+import tempfile
+from os import fspath
```
daandemeyer: We do `import os` everywhere else so that as well here please

tests/test_linters.py
```diff
@@ -0,0 +1,98 @@
...
+
+from mkosi.run import run
+
+# tolerate missing tools in some random dev environment, but enforce them in CI/mkosi box
+SKIP_MISSING_TOOLS = "MKOSI_IN_BOX" not in os.environ
+
+REPO_ROOT = Path(__file__).parent.parent
+
+
+def _kernel_install_files() -> list[str]:
```
daandemeyer: We don't bother with `_` prefixing "private" functions anywhere else, so let's not start doing that here

tests/test_linters.py
```diff
@@ -0,0 +1,98 @@
...
+def test_ruff_check() -> None:
+    """Check code quality with ruff."""
+    run(
+        ["ruff", "check", "--output-format=github", "mkosi/", "tests/", *_kernel_install_files()],
+        cwd=REPO_ROOT,
+    )
+
+
+def test_no_tabs_in_code() -> None:
+    """Ensure no tabs are used in Python code."""
```
daandemeyer: This comment and all the ones that follow in this file are pointless, the test names already tell you the same thing so please drop those

daandemeyer (approved):

.gitignore
```diff
@@ -31,3 +27,7 @@ __pycache__
 blog/output
 blog/pelicanconf.py
 blog/publishconf.py
+# Generated man pages
+mkosi/resources/man/*.[17]
+# Generated zipapp
+builddir/
```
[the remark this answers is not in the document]
martinpitt: Done.

tests/test_linters.py
martinpitt: Ack, dropped the prefix.

tests/test_install.py
martinpitt: Changed.

tests/test_linters.py
martinpitt: OK -- wasn't sure about conventions etc., as the workflow also had `name:`s. I agree it's just noise, dropped.

PR #4338, opened 2026-06-01
Don't add El Torito boot catalog for BIOS/grub images by default

### LINKED ISSUE #4337: Fedora buils fail to install grub because of blocklists
Recently, CI has [started failing all Fedora builds](https://github.com/systemd/mkosi/actions/runs/26709370410/job/78718470948?pr=4335#step:11:7326):

> /usr/sbin/grub2-bios-setup: warning: Attempting to install GRUB to a disk with multiple partition labels.  This is not supported yet..
> /usr/sbin/grub2-bios-setup: warning: Embedding is not possible.  GRUB can only be installed in this setup by using blocklists.  However, blocklists are UNRELIABLE and their use is discouraged..
> /usr/sbin/grub2-bios-setup: error: will not proceed with blocklists.
> ‣ [main] "/usr/sbin/grub2-bios-setup --directory /grub /work/var/tmp/mkosi-workspace-kb5y5mxx/staging/image.raw" returned non-zero exit code 1.

This effortlessly reproduces locally in a clean tree with
```sh
bin/mkosi --distribution fedora --format disk --force build
```

#4323 (May 21) still "worked" (well, the two "known" failures), #4334 (May 26) shows that grub failure already. But interestingly, both have `grub2-common-1:2.12-60.fc45` and `grub2-pc-1:2.12-45.fc45`, and this also does not coincide (at least not obviously) with a [recent Fedora grub update](https://bodhi.fedoraproject.org/updates/?packages=grub2). I also checked [dist-git](https://src.fedoraproject.org/rpms/grub2/commits/rawhide) and nothing stands out. So this will need some more digging into.

mkosi.conf.d/x86-64.conf statically configures this, so I suppose CI only tests Bios, not EFI?

martinpitt:
systemd-repart v261 adds an El Torito boot catalog, which writes an ISO9660 header to the image. grub-bios-setup interprets that header as a filesystem on the bare disk and fails:

> warning: Attempting to install GRUB to a disk with multiple partition labels. This is not supported yet
> Embedding is not possible. GRUB can only be installed in this setup by
. using blocklists. However, blocklists are UNRELIABLE and their use is . discouraged
> error: will not proceed with blocklists

Adjust auto mode to only add the El Torito catalog when grub is not installed as the BIOS bootloader (`Explicit ElTorito=yes` still forces it on). This goes back to the status quo ante until systemd 260 for grub/BIOS.

Fixes #4337

---

See the issue above for detailed analysis and debugging notes. I expect all the Fedora tests to pass again with this. There are still two unrelated failures that already happened before (arch distro+debian tools: missing keyring, and arch+fedora: hangs when installing lsof), these will be treated separately.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -3478,9 +3478,16 @@ def make_image(
     if tabs and systemd_tool_version("systemd-repart", sandbox=context.sandbox) >= 258:
         cmdline += ["--append-fstab=auto"]
 
+    # grub-bios-setup interprets El Torito/ISO9660 header as a filesystem on the bare
+    # disk and then refuses to embed itself. So only enable it by default when we are
+    # not installing grub for BIOS.
+    el_torito_wanted = context.config.el_torito == ConfigFeature.enabled or (
+        context.config.el_torito == ConfigFeature.auto and not want_grub_bios(context)
+    )
+
     if (
         el_torito
-        and context.config.el_torito != ConfigFeature.disabled
+        and el_torito_wanted
         and systemd_tool_version("systemd-repart", sandbox=context.sandbox) >= "261~devel"
     ):
         cmdline += ["--el-torito=yes"]
--- a/mkosi/resources/man/mkosi.1.md
+++ b/mkosi/resources/man/mkosi.1.md
@@ -801,7 +801,9 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 
 `ElTorito=`, `--el-torito=`
 :   Whether to add an El Torito boot catalog to boot the ESP when building a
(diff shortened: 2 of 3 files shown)


martinpitt:
The [pyright failure](https://github.com/systemd/mkosi/actions/runs/26755419982/job/78853917143?pr=4338#step:12:7) is fixed in #4335, [here](https://github.com/systemd/mkosi/pull/4335/changes/0444a651358615814ca71c77108e88e4f781c33b#r3334236417). As that is "almost" approved, let's land that first?

behrmann (approved):

martinpitt:
I temporarily cherry-picked the pyright fix from #4335 to get a CI run. But let's queue this after that PR.

PR #4339, opened 2026-06-01
Fix unit tests at package build time

bluca:
Unit tests are broken as they expect a CI environment, but they are also ran at package build time. Skip gracefully when the CI tools are not available.

[   59s] ERROR    root:run.py:162 "/tmp/tmp1x492nlw/testvenv/bin/python3 -m pip install --upgrade setuptools wheel pip" returned non-zero exit code 1.
...
[   59s] ERROR    root:log.py:32 git not found.
...
[   59s] FAILED tests/test_install.py::test_venv_installation - subprocess.CalledProce...
[   59s] FAILED tests/test_install.py::test_editable_venv_installation - subprocess.Ca...
[   59s] FAILED tests/test_linters.py::test_no_tabs_in_code

Follow-up for b0f9525c2c74401fc4280c48233aa9831ced60d5

### DIFF
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -1,12 +1,45 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
+import importlib.util
 import os
+import shutil
+import socket
 import tempfile
 from pathlib import Path
+
+import pytest
 
 from mkosi.run import run
 
 REPO_ROOT = Path(__file__).parent.parent
+
+# tolerate missing tools/prerequisites in some random dev or packaging environment, but enforce them
+# in CI/mkosi box
+SKIP_MISSING_TOOLS = "MKOSI_IN_BOX" not in os.environ
+
+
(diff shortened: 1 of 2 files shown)


tests/test_install.py
```diff
@@ -1,19 +1,53 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
+import importlib.util
```
martinpitt, line 3: How about we mark these as `pytestmark = pytest.mark.install` and explicitly run them in CI only? That might be easier, more robust, and more predictable. As I "broke" it, I'm happy to propose a PR.

martinpitt (commented):
Oops, thanks @bluca !

tests/test_install.py
bluca: Sure, go for it

martinpitt: I sent #4341 with my alternate proposal.

PR #4341, opened 2026-06-02
Fix unit tests for package builds

martinpitt:
My alternate proposal for #4339

I kept the `test_linters.py` as the first commit from @bluca, and handled the install tests through a marker. I think this is simpler, more robust, and more predictable. WDYT?

### DIFF
--- a/tests/test_linters.py
+++ b/tests/test_linters.py
@@ -40,6 +40,7 @@ def test_ruff_check() -> None:
     )
 
 
+@pytest.mark.skipif(skip_if_missing("git"), reason="git not found")
 def test_no_tabs_in_code() -> None:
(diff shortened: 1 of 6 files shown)


tests/test_linters.py
```diff
@@ -40,6 +40,7 @@ def test_ruff_check() -> None:
     )
 
 
+@pytest.mark.skipif(_skip_if_missing("git"), reason="git not found")
```
behrmann: Any reason `skip_if_missing` didn't lose it's underscore before?
```suggestion
@pytest.mark.skipif(skip_if_missing("git"), reason="git not found")
```

behrmann (approved):

bluca (approved):

tests/test_linters.py
martinpitt: Just forgetfulness, sorry! Fixed in a separate commit.

PR #4343, opened 2026-06-02
action: skip install of unavailable packages

katexochen:
Some packages are not available on ubuntu-22.04 runners, skip them.


behrmann (approved):
Makes sense and shouldn't be an imposition, since people can (and should) use a tools tree anyway.

Issue #4344, opened 2026-06-03
distro=arch / tools=fedora failure: pacman fork deadlock with curl 8.20


martinpitt:
## Symptom

The [distro: arch, tools: fedora integration test](https://github.com/systemd/mkosi/actions/runs/26827453110/job/79099443526) consistently times out. Last known success on April 9 in #4273; the next PR #4274 failed on April 16.

Reproducible locally with

```sh
mkosi box -- python3 -m pytest -m integration --distribution arch -k 'test_extension[confext]' tests
```

## Debugging

While hanging, there are **two** pacman processes:

- the parent blocks in `poll()` on a socketpair to its child
- the child (forked by libalpm to run `ldconfig` in the buildroot) blocks forever in `FUTEX_WAIT`

Userspace backtrace of the child (gdb):

```
#3  pthread_cond_wait
#4  thrdq_unlink                      (libcurl.so.4 — Fedora curl 8.20.0)
#5  Curl_async_thrdd_multi_destroy
#6  curl_multi_cleanup
#7  _alpm_handle_free                 (libalpm.so.16 — pacman 7.1.0)
#8  _alpm_run_chroot
#9  _alpm_ldconfig
#10 alpm_trans_commit
```

**Root cause:** after extracting the package, libalpm's `_alpm_run_chroot()` forks to run `ldconfig` in the
buildroot. In the child it calls `_alpm_handle_free()` → `curl_multi_cleanup()`. curl 8.20.0 replaced the
per-handle resolver threads with a pooled threaded resolver (`thrdq`, curl commit `39036c9021`), and its
cleanup waits on a condition variable for resolver worker threads — which **do not exist in the forked
child**. The child deadlocks; the parent waits on the socketpair forever. (libcurl is documented as not
fork-safe; libalpm has always done this, it only became fatal with curl 8.20.)

The reliable trigger is a transaction that downloads a package and immediately commits (here: the tiny
`lsof` install in `test_extension`), so the resolver thread pool is still live at fork time.

## Why exactly this test

The deadlock needs **pacman** (→ `distro=arch`) linked against **curl 8.20** with an **unpatched libalpm**:

| Tools tree | pacman | curl | Result |
|---|---|---|---|
| fedora (Rawhide) | 7.1.0, **unpatched** (no patches in `pacman.spec`) | 8.20.0 | **hangs** |
[...]

PR #4345, opened 2026-06-03
tests: Run distro=arch / tools=fedora on Fedora 44 to evade pacman bug

martinpitt:
Fedora Rawhide's pacman 7.1.0 deadlocks with curl 8.20.0 when installing packages. Pin the tools tree to the latest stable Fedora release for this test run combination until the pacman fix lands in Rawhide.

curl issue (closed, libcurl is not fork-safe by design): https://github.com/curl/curl/issues/21466

pacman fix (not yet merged):
https://gitlab.archlinux.org/pacman/pacman/-/merge_requests/360

See issue #4344 for the full analysis.

---

This is clearly a hack, but I don't really want to push for adding an unmerged upstream pacman patch into the Fedora package. All in all I think this is better than a prolonged red integration test, though.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,6 +81,12 @@ jobs:
           - distro: ubuntu
             tools: opensuse
         include:
+          # Fedora rawhide's pacman 7.1.0 deadlocks with curl 8.20 when installing packages
+          # (https://github.com/curl/curl/issues/21466, https://gitlab.archlinux.org/pacman/pacman/-/merge_requests/360)
+          # use the latest stable release instead, until https://src.fedoraproject.org/rpms/pacman/pull-request/5# lands
+          - distro: arch
+            tools: fedora
+            tools_release: "44"
           # low rate limit on s390x/ppc64le/arm64 workers
           - distro: debian
             tools: debian
@@ -136,6 +142,7 @@ jobs:
 
           [Build]
           ToolsTreeDistribution=${{ matrix.tools }}
+          ToolsTreeRelease=${{ matrix.tools_release }}
           Environment=SYSTEMD_REPART_MKFS_OPTIONS_EROFS="--quiet"
 
           [Runtime]


martinpitt:
[arch/fedora passed](https://github.com/systemd/mkosi/actions/runs/26898232399/job/79344908569?pr=4345) :+1: . The [opensuse failures](https://github.com/systemd/mkosi/actions/runs/26898232399/job/79344908553?pr=4345) seem to happen often, this is a classic case of "*nnng* needs retry on transient errors". I'll see to that tomorrow, but it's independent.

that leaves the last [permanent failure on arch/debian](https://github.com/systemd/mkosi/actions/runs/26898232399/job/79344908514?pr=4345), which is still on my todo list.

@daandemeyer WDYT?

behrmann (approved):
I guess it's fine to add this, since it's only that one combination, that we ignored so far. We'll need to remember to clean this up again once this gets resolved (or 44 becomes too old and things start breaking because of that :) )

That being said, I do wonder why this is breaking on rawhide, but not on Arch itself, since pacman in Arch depends on curl as well and pacman in Arch is 7.1 and curl is 8.20.

martinpitt:
@behrmann it's mentioned in #4344 -- arch's pacman carries a [reduced fix](https://gitlab.archlinux.org/archlinux/packaging/packages/pacman/-/commit/387ebf06060fd5f4ccd2b5281789b11d30c09) which is sufficient to avoid the cleanup threading hang.

martinpitt:
transient opensuse failure addessed in #4346

behrmann:
> @behrmann it's mentioned in #4344 -- arch's pacman carries a [reduced fix](https://gitlab.archlinux.org/archlinux/packaging/packages/pacman/-/commit/387ebf06060fd5f4ccd2b5281789b11d30c09) which is sufficient to avoid the cleanup threading hang.

I had missed that, those bot descriptions are somewhat wordy. :)

Wouldn't it make more sense for Fedora to pull that patch as well?

martinpitt:
@behrmann It's not a "nice" patch (memleak), but indeed Arch applied it, so "good enough". I filed https://src.fedoraproject.org/rpms/pacman/pull-request/5 and validated it works. As it happens, we have a good relationship with the Fedora maintainer (hello @keszybz ! :wave: ), so indeed this has a chance to land soon. We can stall this PR for a little while.

martinpitt:
Rebased to add the Fedora PR reference. Inofficially, I just want to get a new test run to see if Debian's new arch keyring package now works as expected.

martinpitt:
Converting to Draft. I can prod @keszybz this week if we can land the Fedora fix soon.

martinpitt:
My PR landed, and we have a build: https://bodhi.fedoraproject.org/updates/FEDORA-2026-c4544289d1

So this is obsolete now.

Incidentally, Debian's pacman now has exactly the same bug, I'll send the corresponding fix there.

PR #4346, opened 2026-06-04
installer: Auto-retry package index download failures

martinpitt:
These are notoriously prone to transient network errors, and every CI
system in the world eventually grows an auto-retry for these. Now it's
mkosi's time!

The local repository set up by createrepo() never fails transiently, so
opt out of retrying there. apk's createrepo() keeps retrying, as its
sync refreshes all configured repositories including remote ones.

Don't apply this to other operations yet -- e.g. downloading actual
packages is much more expensive, and at least `dnf` has some basic retry
already built in (trying different mirrors). Let's add more of these
carefully only when they become a problem. E.g. zypper's
`--download in-advance` is a good candidate.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

----

I saw [this failure](https://github.com/systemd/mkosi/actions/runs/26898232399/job/79344908553?pr=4345#step:10:15) at least twice now.

### DIFF
--- a/mkosi/installer/apk.py
+++ b/mkosi/installer/apk.py
@@ -186,6 +186,8 @@ class Apk(PackageManager):
             + ["update", *(["--update-cache"] if force else [])],
             sandbox=context.sandbox(network=True, options=options),
             env=env,
+            # Downloading repository metadata often fails with transient network errors
+            num_retries=3,
         )
         rmtree(tmp)
 
--- a/mkosi/installer/apt.py
+++ b/mkosi/installer/apt.py
@@ -227,6 +227,7 @@ class Apt(PackageManager):
         apivfs: bool = False,
         options: Sequence[PathString] = (),
         stdout: _FILE = None,
+        retry: bool = False,
     ) -> CompletedProcess:
         with umask(~0o755):
             # TODO: Drop once apt 2.5.4 is widely available.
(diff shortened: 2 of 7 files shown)


martinpitt:
Hmm, so [this shows the retry is working](https://github.com/systemd/mkosi/actions/runs/26934555522/job/79461431681?pr=4346), but it seems opensuse "oss" is currently genuinely kaputt? But locally it works, maybe it's somehow rate-limited when being called from GitHub? These days, many sites have scraper protection which might interfere. Moving to draft for the time being.

daandemeyer:
I very much think mkosi is the wrong place to handle this. We have no clue whether a failure is transient or not, the package managers do, so I think retries should be handled by the package manager rather than us.

martinpitt:
@daandemeyer I agree in principle. In practice they don't though, so these retries get added everywhere. But okay!

martinpitt:
In the light of the current GPG sign/mirror failures that are plaguing us, I'd like to at least do some more test runs here. The previous round [all failed](https://github.com/systemd/mkosi/actions/runs/26935443706/job/79464269080) on a different issue:

> Run sudo curl -fsSL -o /usr/local/bin/apk https://gitlab.alpinelinux.org/api/v4/projects/5/packages/generic/v3.0.6/x86_64/apk.static
> curl: (56) Send failure: Connection reset by peer

and that is 100% us ("Install apk" step in our workflow). That could use a retry option (curl has plenty), but that isn't what's currently plaguing us. I want to see how green I get stuff with that retry magic, just to collect data (not necessarily to land it).

martinpitt:
I collected some more data, mostly on my fork (to avoid notification noise). Conclusion: There are about 4 things wrong currently (ugh a rabbit hole), but auto-retry does not help with *any* of them. So this really can be closed.

martinpitt:
:arrow_right: #4361 addresses the root cause.

PR #4348, opened 2026-06-04
config: Pull repository keys on Debian by default

### LINKED ISSUE #4347: Building Arch images with a Debian tools tree fails: "required key missing from keyring"
The [`Distribution=arch`/`ToolsTreeDistribution=debian` scenario](https://github.com/systemd/mkosi/actions/runs/26898232399/job/79344908514?pr=4345) has failed in CI for about a month now. This is the only remaining (persistent) failure. Building Arch images with an Arch or Fedora tools tree works fine.

## Reproducer

```ini
# mkosi.local.conf
[Distribution]
Distribution=arch

[Build]
ToolsTreeDistribution=debian
```

```console
$ mkosi -f build
```

The first pacman transaction (installing `filesystem` and `iana-etc` into the
`default-initrd` buildroot) fails:

```
‣ Building default-initrd image
‣ [default-initrd]  Installing Arch Linux
resolving dependencies...
looking for conflicting packages...

Packages (2) iana-etc-20260530-1  filesystem-2025.10.12-1

checking keyring...
downloading required keys...
error: keyring is not writable
error: required key missing from keyring
error: failed to commit transaction (unexpected error)
Errors occurred, no packages were upgraded.
‣ [default-initrd] "pacman --root=/buildroot ... --sync --needed --assume-installed initramfs filesystem" returned non-zero exit code 1.
```

## Root cause

The "required key missing" message is misleading — the key is present but **expired**:

1. `filesystem-2025.10.12-1` is signed by subkey `9B7A287D9A2EC608` of David Runge's
   packager key `991F6E3F0765CF6295888586139B09DA5BF0D338`, which carried an expiry
   date of **2026-05-09**.

2. Upstream extended the key's validity to 2030-01-08 in `archlinux-keyring`
   `20260420`. However, mkosi populates pacman's keyring from the **tools tree's**
   `archlinux-keyring` package, and Debian's is a stale snapshot:
   - sid/forky: `0~20251116-1` (2025-11-16)
   - trixie: `0~20250430-1`

   Both still contain the old, now-expired copy of the key. The populated keyring
   cache confirms it:

   ```console
   $ gpg --homedir mkosi.cache/arch~rolling~x86-64~main.keyring.cache --list-keys 9B7A287D9A2EC608
   pub   ed25519 2022-05-10 [SC] [expired: 2026-05-09]
         991F6E3F0765CF6295888586139B09DA5BF0D338
   uid           [ expired] David Runge <dvzrv@archlinux.org>
   ```

3. When pacman encounters an unusable signing key it tries to re-fetch a fresh copy
   via WKD ("downloading required keys..."). But mkosi bind-mounts the keyring
   **read-only** during package installation (`mkosi/installer/pacman.py`, only the
   initial `pacman-key --init/--populate` step gets a writable mount), so the import
   fails with "keyring is not writable" and pacman falls back to the generic
   "required key missing from keyring" error.

## Why only this combination fails

- **Arch tools tree**: `archlinux-keyring` is always current → key valid.
- **Fedora tools tree**: Fedora updates its `archlinux-keyring` package frequently →
  the expiry extension was picked up in time.
- **Debian tools tree**: the package lags upstream by ~6 months → expired key.

## Fix

The immediate fix is on the Debian side, I filed <https://bugs.debian.org/1138819>.
But this is structural: building Arch images with a non-Arch tools tree breaks
whenever *any* Arch packager key expires or rotates faster than the distribution
refreshes its `archlinux-keyring` snapshot.

I'll look into how we can avoid this in mkosi. It seems commit ddea81d8 already
addressed this somehow, but doesn't apply to this scenario.

martinpitt:
Debian's `archlinux-keyring` is lagging behind by more than half a year by now, and contains expired developer keys
(https://bugs.debian.org/1138819). This has broken the build of arch with a Debian tool box since May 09.

Even if/when the above Debian update happens, this is a recurring problem. Give in, and download the keys at build time just like we do on Ubuntu.

Fixes #4347

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -1087,11 +1087,14 @@ def config_default_repository_key_fetch(namespace: dict[str, Any]) -> bool:
         if d == "nixos":
             return True
 
-        return d == Distribution.ubuntu and needs_repository_key_fetch(namespace["distribution"])
+        return d in (Distribution.ubuntu, Distribution.debian) and needs_repository_key_fetch(
+            namespace["distribution"]
+        )
 
-    return namespace["tools_tree_distribution"] == Distribution.ubuntu and needs_repository_key_fetch(
-        namespace["distribution"]
-    )
+    return namespace["tools_tree_distribution"] in (
+        Distribution.ubuntu,
+        Distribution.debian,
+    ) and needs_repository_key_fetch(namespace["distribution"])
 
 
 def config_default_source_date_epoch(namespace: dict[str, Any]) -> Optional[int]:
--- a/mkosi/resources/man/mkosi.1.md
+++ b/mkosi/resources/man/mkosi.1.md
@@ -588,8 +588,9 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 
 `RepositoryKeyFetch=`, `--repository-key-fetch=`
 :   Controls whether **mkosi** will fetch distribution GPG keys remotely. Enabled by
-    default on Ubuntu when not using a tools tree or when using Ubuntu tools trees to build
-    Arch Linux or RPM-based distributions. Disabled by default on all other distributions.
+    default on Debian and Ubuntu when not using a tools tree or when using Debian or Ubuntu
+    tools trees to build Arch Linux or RPM-based distributions. Disabled by default on all
+    other distributions.
     When disabled, the distribution GPG keys for the target distribution have to be installed
     locally on the host system alongside the package manager for that distribution.
 


daandemeyer (approved):

behrmann:
Beautiful, now the test fails because it runs into a timeout.

martinpitt:
Uh, what -- local build was quite happy. Okay, this is putting up a fight then -- I'm afraid I need to do something else now, but I'll keep it on my todo list.

martinpitt:
Hm, that timeout suspiciously reeks of the pacman hang #4344

martinpitt:
The [last arch/debian run](https://github.com/systemd/mkosi/actions/runs/27186507766/job/80256777951?pr=4345) pulls in the updated keyring package, so that bug is solved. It's not going to fix the systematic problem, but we can limp along, and keep this on the shelf.

But now it runs into the timeout similar to how it failed here, but that's an unrelated problem.

martinpitt:
I reported the pacman hang on Debian to https://bugs.debian.org/1139622 and sent a package update patch there.

PR #4349, opened 2026-06-04
ubuntu: Don't treat linux-tools-xxx as a kernel package

daandemeyer:


bluca (approved):

behrmann:
Looking at the Debian package search, shouldn't we exclude [`linux-base`](https://packages.debian.org/trixie/all/linux-base/filelist) as well?

bluca:
and linux-libc-dev and linux-bpf-dev and linux-perf and linux-sysctl-defaults and...

bluca:
but this fixes an immediate blocker for my systemd PR so I'll merge it, more can be added (or rather maybe we can figure out a more robust way)

PR #4269, opened 2026-04-06
Fix #4206: cp: setting attribute 'security.ima' for 'security.ima': O...

JiwaniZakir:
Closes #4206

## Motivation

When building as non-root, `cp --preserve=xattr` fails with `Operation not permitted` if source files carry `security.ima` xattrs (e.g. `/etc/crypto-policies/back-ends/*` on Fedora with IMA enabled). The existing xattr guard in `copy_tree` only suppresses xattr copying on overlayfs destinations to work around SELinux issues, leaving the IMA case unhandled.

## Changes

**`mkosi/tree.py`**

- Adds `tree_has_ima_xattr(path: Path) -> bool` (mirrors the existing `tree_has_selinux_xattr`): walks `path` and all descendants via `rglob`, checking `os.listxattr(..., follow_symlinks=False)` for `"security.ima"`.
- In `copy_tree`, the condition that appends `",xattr"` to the `attrs` string (line ~134) is extended with a second guard: `os.getuid() == 0 or not tree_has_ima_xattr(src)`. xattr copying is now skipped when running as non-root and the source tree contains IMA xattrs, regardless of the filesystem type of the destination.

**`tests/test_tree.py`** (new file)

- `test_tree_has_ima_xattr_present` / `test_tree_has_ima_xattr_absent`: verify `tree_has_ima_xattr` returns the correct boolean when `os.listxattr` is mocked to include or exclude `"security.ima"`.
- `test_tree_has_selinux_xattr_present` / `test_tree_has_selinux_xattr_absent`: equivalent coverage for the pre-existing `tree_has_selinux_xattr` function, which had no unit tests.

## Testing

Unit tests added in `tests/test_tree.py` cover both the positive and negative cases for `tree_has_ima_xattr` and `tree_has_selinux_xattr` using `unittest.mock.patch` on `os.listxattr` and `Path.rglob`.

Manual verification: running `mkosi build` with `ToolsTreeDistribution=debian` / `ToolsTreeRelease=trixie` as a non-root user on a Fedora 43 host with IMA xattrs present on `/etc/crypto-policies/back-ends/*` no longer produces `cp: setting attribute 'security.ima' ... Operation not permitted` errors and completes successfully.

---
*This PR was created with AI assistance (Claude). The changes were reviewed by quality gates and a critic model before submission.*

PR #4353, opened 2026-06-05
Skip security.ima xattrs when copying tree as non-root

### LINKED ISSUE #4206: cp: setting attribute 'security.ima' for 'security.ima': Operation not permitted

v26

### Used host distribution

Fedora 43

### Used target distribution

Debian trixie

### Linux kernel version used

6.18.7-200.fc43.x86_64

x86_64

Building a tools tree as non-root fails with obscure xattr errors.

This is probably because this specific directory has `security.ima` xattrs set on my host.

```
> getfattr -m - -d /etc/crypto-policies/back-ends/*
# file: etc/crypto-policies/back-ends/bind.config
security.ima=0sAwIE0EMcMgBIMEYCIQCRq5r6QOdBuWm1/RTRZe0q4fdFbizOuxL5D6lU2iyLvQIhAL0ZI8qawMP+n2VpaxpZ8NSWSeSJZV/RPDcl3Rm834Ei
security.selinux="system_u:object_r:usr_t:s0"

...
```

Commenting out https://github.com/systemd/mkosi/blob/e9abfab744340dd2f608b589a9252a3e53b071c3/mkosi/tree.py#L128-L129 solves the problem for me.

### Used mkosi config

```ini
[Build]
ToolsTreeDistribution=debian
ToolsTreeRelease=trixie
```

### mkosi output

```sh
> mkosi build
cp: setting attribute 'security.ima' for 'security.ima': Operation not permitted
cp: setting attribute 'security.ima' for 'security.ima': Operation not permitted
cp: setting attribute 'security.ima' for 'security.ima': Operation not permitted
cp: setting attribute 'security.ima' for 'security.ima': Operation not permitted
cp: setting attribute 'security.ima' for 'security.ima': Operation not permitted
cp: setting attribute 'security.ima' for 'security.ima': Operation not permitted
cp: setting attribute 'security.ima' for 'security.ima': Operation not permitted
cp: setting attribute 'security.ima' for 'security.ima': Operation not permitted
cp: setting attribute 'security.ima' for 'security.ima': Operation not permitted
cp: setting attribute 'security.ima' for 'security.ima': Operation not permitted
cp: setting attribute 'security.ima' for 'security.ima': Operation not permitted
cp: setting attribute 'security.ima' for 'security.ima': Operation not permitted
cp: setting attribute 'security.ima' for 'security.ima': Operation not permitted
‣ "cp --recursive --no-dereference --preserve=mode,links,timestamps,ownership,xattr --reflink=auto --copy-contents /work/usr/share/crypto-policies/back-ends/DEFAULT /work/home/fina/.cache/mkosi/mkosi-workspace-c6epmdug/sandbox/etc/crypto-policies/back-ends --no-target-directory" returned non-zero exit code 1.
```

behrmann:
This is a resubmission of #4269 without the tests, that mocked out everything except for `any`, and with the ruff failure fixed. 

Fixes: #4206

### DIFF
--- a/mkosi/tree.py
+++ b/mkosi/tree.py
@@ -130,8 +130,11 @@ def copy_tree(
 
         # Trying to copy selinux xattrs to overlayfs fails with "Operation not supported" in containers.
         # Trying to copy security.ima xattrs fails with "Operation not permitted" when non-root.
-        if (statfs(os.fspath(dst.parent)) != OVERLAYFS_SUPER_MAGIC or not tree_has_selinux_xattr(src)) and (
-            os.getuid() == 0 or not tree_has_ima_xattr(src)
+        # Copying SELinux attributes also needs root and it seems unlikely, that copying IMA attributes to
+        # overlayfs should work when SELinux does not.
+        if not (
+            (statfs(os.fspath(dst.parent)) == OVERLAYFS_SUPER_MAGIC or os.getuid() != 0)
+            and (tree_has_selinux_xattr(src) or tree_has_ima_xattr(src))
         ):
             attrs += ",xattr"
 
--- a/mkosi/tree.py
+++ b/mkosi/tree.py
@@ -103,6 +103,10 @@ def tree_has_selinux_xattr(path: Path) -> bool:
     )
 
 
+def tree_has_ima_xattr(path: Path) -> bool:
+    return any("security.ima" in os.listxattr(p, follow_symlinks=False) for p in (path, *path.rglob("*")))
+
+
 def copy_tree(
     src: Path,
     dst: Path,
@@ -125,7 +129,10 @@ def copy_tree(
         attrs += ",timestamps,ownership"
 
         # Trying to copy selinux xattrs to overlayfs fails with "Operation not supported" in containers.
-        if statfs(os.fspath(dst.parent)) != OVERLAYFS_SUPER_MAGIC or not tree_has_selinux_xattr(src):
+        # Trying to copy security.ima xattrs fails with "Operation not permitted" when non-root.
+        if (statfs(os.fspath(dst.parent)) != OVERLAYFS_SUPER_MAGIC or not tree_has_selinux_xattr(src)) and (
+            os.getuid() == 0 or not tree_has_ima_xattr(src)
+        ):
             attrs += ",xattr"
 
     def copy() -> None:


behrmann:
@jonas2515 does this address your issue?

jonas2515:
Hah, just tested this, and I noticed the tree doesn't actually have `security.ima` xattrs set. What triggers the failure instead seems to be a logic mistake in the if condition: Replacing the `or` with an `and` makes it work for me.

```diff
-        if statfs(os.fspath(dst.parent)) != OVERLAYFS_SUPER_MAGIC or not tree_has_selinux_xattr(src):
+        if statfs(os.fspath(dst.parent)) != OVERLAYFS_SUPER_MAGIC and not tree_has_selinux_xattr(src):
```

behrmann:
The original intent there is expressed by the comment
>  Trying to copy selinux xattrs to overlayfs fails with "Operation not supported" in containers.

i.e. what we want is to only copy extended attributes, when the target is not
overlayfs, for which this fails, or when the source doesn't have SELinux
attributes, and this operation cannot fail (except for ima).

That means we want to exclude the case of "target FS is overlayfs" and "source
has SELinux attributes", that is
```python
   not (target_is_overlay_fs and src_has_selinux_attr)
=> not target_is_overlay_fs or not src_has_selinux_attr
```
which is what we have in the code.

Your change would make this
```python
   not target_is_overlay_fs and not src_has_selinux_attr 
=> not (target_is_overlay_fs or src_has_selinux_attr)
```

Shortening those two conditions down to A and B we currently have
|       | A | not A |
|-------|---|-------|
| B     |   | X     |
| not B | X | X     |

and with your change this would be
|       | A | not A |
|-------|---|-------|
| B     |   |       |
| not B |   | X     |

which is already included in ours. Considering that "not B" i.e. the source not
having SELinux attributes is not a problem, could you be in the upper right
quadrant and your target, while not being an overlayfs, does not support SELinux
(or other extended) attributes?

jonas2515:
> could you be in the upper right quadrant and your target, while not being an overlayfs, does not support SELinux
(or other extended) attributes?

Not sure what exactly the target in my case is, but I'm on a freshly installed Fedora 44 (so btrfs root partition) on x86, trying to build systemd by using `mkosi -f box`, so should be pretty much the most standard use of mkosi.

bluca (approved):

behrmann:
@jonas2515 I talked to @keszybz who knows more about SELinux than me and reworked the check slightly, assuming that not being root was the issue in your case. It would be nice if you could give this version another go.

keszybz (approved):
LGTM.

jonas2515:
Only got to it now, but can confirm this fixed the issue, thanks everyone!

PR #4358, opened 2026-06-11
tests: Make integration tests runnable locally

martinpitt:
Move integration test setup out of the workflow into a separate script, and document how to run them.


tools/integration-test-setup.sh
```diff
@@ -0,0 +1,62 @@
...
+    echo "Usage: $0 <distribution> <tools-tree-distribution>" >&2
+    exit 1
+fi
+
+distribution="$1"
+tools="$2"
+
+cat >mkosi.local.conf <<EOF
+[Distribution]
+Distribution=$distribution
```
behrmann: Nit: can we get curly braces here (and in the other heredocs that are interpolated)?
```suggestion
Distribution=${distribution}
```

tools/integration-test-setup.sh
```diff
@@ -0,0 +1,62 @@
...
+EOF
+
+# TODO: Remove once all distros have recent enough systemd that knows systemd.default_device_timeout_sec.
+mkdir -p mkosi-initrd/mkosi.extra/usr/lib/systemd/system.conf.d
+cat >mkosi-initrd/mkosi.extra/usr/lib/systemd/system.conf.d/device-timeout.conf <<EOF
+[Manager]
+DefaultDeviceTimeoutSec=180
+EOF
+
+cat >mkosi.configure <<'EOF'
```
behrmann: Non-interpolating heredocs are a bit of a deep cut, I'm not sure most readers understand that this one will not interpolate. While I generally prefer non-interpolating strings when possible, for heredocs it's always a bit unexpected. Since the one above doesn't do it either, I think it would be better to not do it here, too.

tools/integration-test-setup.sh
martinpitt: Done, same for tools.

tools/integration-test-setup.sh
martinpitt: I was going back and forth with this, but as you mentioned it as well, I maded it consistently interpolating.

martinpitt:
Given how slow github is today, I'll let [this transient opensuse mirror error](https://github.com/systemd/mkosi/actions/runs/27335446966/job/80773799230?pr=4358) slide. I can't re-run individual tests (needs project admin).

tools/integration-test-setup.sh
behrmann, line 1: As a general comment: Making this usable for regular development is great, but it's not unheard of for people to directly start writing mkosi config into their git checkout (as much as one can discourage that), so this script should not clobber existing files. since they will then likely not be checked in (and after this series, ignored).

.github/workflows/ci.yml
```diff
@@ -122,54 +122,24 @@ jobs:
...
-          echo "Hello from inside mkosi.configure!" >&2
-          cat
-          EOF
-          chmod +x mkosi.configure
+      - name: Configure and build
+        run: sudo tools/integration-test-setup.sh ${{ matrix.distro }} ${{ matrix.tools }}
 
-          # prepare and postinst are already used in CI
-          for script in sync build finalize postoutput clean
+      - name: Show generated configuration
```
behrmann: What's the idea behind this? There is a call to `mkosi summary` below, already. Granted, it doesn't show the script contents, but maybe that's not necessary?

.github/workflows/ci.yml
```diff
@@ -122,54 +122,24 @@ jobs:
...
+        if: always()
+        run: |
+          for f in \
+              mkosi.local.conf \
+              mkosi.configure \
+              mkosi.sync mkosi.build mkosi.finalize mkosi.postoutput mkosi.clean \
```
behrmann: Nit: not breaking the line below makes it harder to see that these are multiple elements
```suggestion
              mkosi.sync \
              mkosi.build \
              mkosi.finalize \
              mkosi.postoutput \
              mkosi.clean \
```

.github/workflows/ci.yml
```diff
@@ -178,12 +148,6 @@ jobs:
       - name: Summary
         run: sudo mkosi summary
```
behrmann: This should maybe also go into the script.

tools/integration-test-setup.sh
```diff
@@ -0,0 +1,62 @@
...
+for script in sync build finalize postoutput clean; do
+    cat >"mkosi.$script" <<EOF
+#!/bin/bash
+echo "Hello from inside mkosi.$script" >&2
+EOF
+    chmod +x "mkosi.$script"
+done
+
+# Build the tools tree and the image
+bin/mkosi -f box -- true
```
behrmann, line 95: It might be nice to add sections here, because what were multiple steps gets collated into one.
```suggestion
[ -n "GITHUB_ACTIONS" ] && echo '::group::build_tools_tree'
bin/mkosi -f box -- true
[ -n "GITHUB_ACTIONS" ] && echo '::endgroup::'
```
(same for steps below)

tools/integration-test-setup.sh
```diff
@@ -0,0 +1,62 @@
...
+EOF
+chmod +x mkosi.configure
+
+# prepare and postinst are already used in CI.
+for script in sync build finalize postoutput clean; do
+    cat >"mkosi.$script" <<EOF
+#!/bin/bash
+echo "Hello from inside mkosi.$script" >&2
+EOF
+    chmod +x "mkosi.$script"
```
behrmann: Nit:
```suggestion
    cat >"mkosi.${script}" <<EOF
#!/bin/bash
echo "Hello from inside mkosi.${script}" >&2
EOF
    chmod +x "mkosi.${script}"
```

behrmann (commented):
Really nice!

.github/workflows/ci.yml
martinpitt: We do it now, and I didn't want to regress the logging. Seems useful to fish it out of the log, as it will change over time. But I'm not wed to that, it should always be reproducible by running the script locally. I.e. I'm fine with dropping it.

daandemeyer (approved):

.github/workflows/ci.yml
behrmann: Right, didn't see the forest for all the teas.

.github/workflows/ci.yml
```diff
@@ -122,54 +122,24 @@ jobs:
...
-              tee "mkosi.${script}" <<TOK
-          #!/bin/bash
-          echo "Hello from inside mkosi.${script}" >&2
-          TOK
-              chmod +x "mkosi.${script}"
+              echo "### $f"
+              cat "$f"
           done
 
       - name: Generate key
```
behrmann: Ah, this, too should go into the script, because the key will be needed by some of the tests.

tools/integration-test-setup.sh
martinpitt: Right. I added a CI marker comment to the generated config and a safety check to the setup script that bails if the marker isn't there. That should hopefully prevent the worst of it.

.github/workflows/ci.yml
martinpitt: Most of these files are absolutely boring and static, so I dropped them. The only vaguely interesting one is mkosi.local.conf, but even that only differs in two words that are already visible from the test name and can also easily be reproduced locally. The configuration is also shown by `mkosi summary`, so even more redundant. So in the end I just dropped that "dump config" step entirely. Let me know if you'd rather have it back (@behrmann already mentioned the "why is this here?" above..)

.github/workflows/ci.yml
martinpitt: It's very noisy, and locally even involves a pager. I kept it in the workflow for now, there it is interesting as it also shows the installed packages (which may change over time).

tools/integration-test-setup.sh
martinpitt: OK, applied.

.github/workflows/ci.yml
martinpitt: Right, done.

tools/integration-test-setup.sh
martinpitt: Done.

tools/integration-test-setup.sh
```diff
@@ -0,0 +1,90 @@
...
+bin/mkosi -f box -- true
+endgroup
+
+group "build_image"
+bin/mkosi -f
+endgroup
+
+group "genkey"
+[ -e mkosi.key ] || bin/mkosi genkey
+endgroup
```
behrmann: This should go before the build of the tools tree.

tools/integration-test-setup.sh
```diff
@@ -0,0 +1,90 @@
...
+mkdir -p mkosi-initrd/mkosi.extra/usr/lib/systemd/system.conf.d
+cat >mkosi-initrd/mkosi.extra/usr/lib/systemd/system.conf.d/device-timeout.conf <<EOF
+[Manager]
+DefaultDeviceTimeoutSec=180
+EOF
+
+cat >mkosi.configure <<EOF
+#!/bin/bash
+echo "Hello from inside mkosi.configure!" >&2
+cat
```
keszybz, line 65: What does this `cat` do?

.github/workflows/ci.yml
behrmann: Mhm, this is a bit of a catch 22. So the reason I was thinking that this should go into the script is that it makes little sense to me to show the summary after already building the thing, i.e. I always found it useful to check the inputs before looking at the failure in CI. 

Since the build output is already rather noisy I think it might make more sense to run this and always connect stdin to `/dev/null` so the pager is skipped.

README.md
```diff
@@ -158,6 +158,30 @@ bin/mkosi box -- pytest -m install
...
 
+# Integration tests
+
+Integration tests build and boot full images. They are marked with the
+`integration` marker, and are skipped by default. They need a tools tree and an
+image to be built first. `tools/integration-test-setup.sh` writes a local
+configuration for the given image and tools tree distribution and builds both:
+
+```sh
+tools/integration-test-setup.sh <distribution> <tools-tree-distribution>
```
keszybz: I think it's better to use $DISTRIBUTION and $TOOLS_TREE_DISTRIBUTION as the placeholders. `<` and `>` are valid shell operators…

keszybz (approved):

tools/integration-test-setup.sh
```diff
@@ -0,0 +1,90 @@
...
+    exit 1
+fi
+
+distribution="$1"
+tools="$2"
+
+group "create_config"
+
+# safety check: don't clobber local config
+if [ -e mkosi.local.conf ] && ! head -n1 mkosi.local.conf | grep -q 'integration-test-setup.sh'; then
```
behrmann: It's a small file, checking it whole seems fine and easier to understand
```suggestion
if [ -e mkosi.local.conf ] && grep -q '# CI config generated by tools/integration-test-setup.sh' mkosi.local.conf; then
```

tools/integration-test-setup.sh
martinpitt: You know cats, they go where they want! :cat: 

Thanks for spotting! Fixed.

.github/workflows/ci.yml
martinpitt: See below, I dropped this complete thing now.

README.md
martinpitt: Sure! Done.

tools/integration-test-setup.sh
martinpitt: Ack, moved.

tools/integration-test-setup.sh
martinpitt: Done, without DRY.

.github/workflows/ci.yml
martinpitt: OK, done. No strong opinion.

tools/integration-test-setup.sh
martinpitt: Actually, this was already in main, it was just moved code. It's somehow load bearing, when I dropped it, a local `tools/integration-test-setup.sh arch fedora` fails with a `json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`.

This [happens in CI](https://github.com/systemd/mkosi/actions/runs/27487430795/job/81246358377?pr=4358#step:7:5798) as well, so I put the cat back.

behrmann: Yeah, the `mkosi.configure` script receives the mkosi config on stdin, can change it, and then needs to put it out on stdout. I didn't see it was `mkosi.configure` when @keszybz was asking, so I didn't have the answer on the ready. When you drop the `cat`, mkosi does not receive any config, hence the error.

tools/integration-test-setup.sh
```diff
@@ -0,0 +1,100 @@
...
+
+# safety check: don't clobber local config
+CI_MARKER='# CI config generated by tools/integration-test-setup.sh'
+if [ -e mkosi.local.conf ] && ! grep -qF "${CI_MARKER}" mkosi.local.conf; then
+    echo "There is an existing mkosi.local.conf, aborting" >&2
+    exit 1
+fi
+
+cat >mkosi.local.conf <<EOF
+$CI_MARKER
```
behrmann: Nit: The style I'm trying to follow is curly braces in strings that are anything more than the plain variable, so the variable is visible in a longer interpolation, but no extraneous curly braces when it's just the plain variable
```suggestion
CI_MARKER='# CI config generated by tools/integration-test-setup.sh'
if [ -e mkosi.local.conf ] && ! grep -qF "$CI_MARKER" mkosi.local.conf; then
    echo "There is an existing mkosi.local.conf, aborting" >&2
    exit 1
fi

cat >mkosi.local.conf <<EOF
${CI_MARKER}
```

behrmann (approved):

tools/integration-test-setup.sh
martinpitt: Thanks @behrmann !

tools/integration-test-setup.sh
martinpitt: Got it! Thanks for the explanation, makes sense. I'll try to keep that in mind in the future. Pushed.

martinpitt:
Thanks @behrmann, @daandemeyer , and @keszybz for your reviews! This looks quite a bit better now indeed. I can't press Ze Button, can one of you please do? Cheers!

PR #4360, opened 2026-06-14
tests: Fix parallel runs

martinpitt:
Pulled out of #4357 to review/land them separately. These apply to pytest as well in principle, if you run two tests in one terminal each, or use pytest-xdist.

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -2565,11 +2565,22 @@ def calculate_signature_gpg(context: Context) -> None:
     ]  # fmt: skip
 
     with complete_step("Signing SHA256SUMS…"):
-        run(
-            cmdline,
-            env=env,
-            sandbox=context.sandbox(options=options),
-        )
+        try:
+            run(
+                cmdline,
+                env=env,
+                sandbox=context.sandbox(options=options),
+            )
+        finally:
+            # gpg autostarts a gpg-agent to sign and, as the sandbox has no PID namespace, that agent is
+            # leaked when the sandbox goes away. Note this will also kill a "real" user agent, but
+            # gpg auto-starts a new one.
+            run(
+                ["gpgconf", "--kill", "gpg-agent"],
+                env=env,
+                sandbox=context.sandbox(options=options),
+                check=False,
+            )
 
 
 def calculate_signature_sop(context: Context) -> None:
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -5445,7 +5445,15 @@ def want_default_initrd(config: Config) -> bool:
     return Path("default") in config.initrds
 
 
-def finalize_historydir(args: Args) -> Path:
+def finalize_historydir(args: Args, output_dir: Optional[Path] = None) -> Path:
+    # When an output dir is given on the CLI, store the build history there so that concurrent builds with
+    # different output dirs don't clobber a shared history. Don't check the finalized OutputDirectory=
+    # config, only the CLI value: the former isn't known yet here (config files and includes are
+    # parsed later) and vm/boot can't see it anyway since they recover the config from the history instead of
+    # parsing it. An output dir set only in config files keeps the history in the config dir.
+    if output_dir is not None:
+        return output_dir / ".mkosi-private/history"
+
     configdir = finalize_configdir(args.directory)
     return (configdir or Path.cwd()) / ".mkosi-private/history"
 
@@ -5502,7 +5510,7 @@ def parse_config(
         return args, None, ()
 
     configdir = finalize_configdir(args.directory)
-    historydir = finalize_historydir(args)
+    historydir = finalize_historydir(args, context.cli.get("output_dir"))
 
     if have_history(args, historydir):
         history = Config.from_partial_json((historydir / "latest.json").read_text())
--- a/mkosi/installer/__init__.py
+++ b/mkosi/installer/__init__.py
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
-from collections.abc import Sequence
-from contextlib import AbstractContextManager
+import contextlib
+from collections.abc import Iterator, Sequence
 from pathlib import Path
 
 from mkosi.config import Config, ConfigFeature, OutputFormat
@@ -9,7 +9,7 @@ from mkosi.context import Context
 from mkosi.mounts import finalize_certificate_mounts
 from mkosi.run import apivfs_options, finalize_interpreter, finalize_passwd_symlinks, find_binary
 from mkosi.tree import rmtree
-from mkosi.util import PathString, flatten, startswith
+from mkosi.util import PathString, flatten, flock, startswith
 
 
 class PackageManager:
@@ -153,22 +153,32 @@ class PackageManager:
         ]  # fmt: skip
 
     @classmethod
+    @contextlib.contextmanager
     def sandbox(
         cls,
         context: Context,
         *,
         apivfs: bool,
         options: Sequence[PathString] = (),
-    ) -> AbstractContextManager[list[PathString]]:
-        return context.sandbox(
-            network=True,
-            options=[
-                *context.rootoptions(),
-                *cls.mounts(context),
-                *cls.options(root=context.root, apivfs=apivfs),
-                *options,
-            ],
-        )  # fmt: skip
+    ) -> Iterator[list[PathString]]:
+        # The package cache directory is shared between all mkosi builds of the same distribution (see
+        # Config.package_cache_dir_or_default()) and is bind mounted read-write into every package manager
+        # sandbox by mounts(). Avoid concurrent downloads, as those result in corruption/truncation and
+        # unnecessary duplicate fetches. Builds of different distributions use different cache directories
+        # and so don't contend.
+        with (
+            flock(context.config.package_cache_dir_or_default()),
+            context.sandbox(
+                network=True,
+                options=[
+                    *context.rootoptions(),
+                    *cls.mounts(context),
+                    *cls.options(root=context.root, apivfs=apivfs),
+                    *options,
+                ],
+            ) as sandbox,
+        ):
+            yield sandbox
 
     @classmethod
     def install(
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -129,7 +129,12 @@ class Image:
 
         return result
 
-    def vm(self, options: Sequence[str] = (), args: Sequence[str] = ()) -> CompletedProcess:
+    def vm(
+        self,
+        options: Sequence[str] = (),
+        args: Sequence[str] = (),
+        ram: str = "1536M",
+    ) -> CompletedProcess:
         need_hyperv_workaround = os.uname().machine == "x86_64"
 
         result = self.mkosi(
@@ -139,7 +144,7 @@ class Image:
                 "--vsock=yes",
                 # TODO: Drop once both Hyper-V bugs are fixed in Github Actions.
                 *(["--qemu-args=-cpu max,pcid=off"] if need_hyperv_workaround else []),
-                "--ram=2G",
+                f"--ram={ram}",
                 "--ephemeral=yes",
                 "--register=no",
                 *options,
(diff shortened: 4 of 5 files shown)


martinpitt:
Working through the failures (ignoring the opensuse repo FUBAR):

- [centos on ubuntu](https://github.com/systemd/mkosi/actions/runs/27493239915/job/81262500204?pr=4360) fails `test_initrd_luks_lvm`, due to a failed unit inside the image:
```
mkosi-check-and-shutdown.sh: + systemctl --failed --no-legend | tee /failed-services
mkosi-check-and-shutdown.sh: ● lvm2-monitor.service masked failed failed lvm2-monitor.service
mkosi-check-and-shutdown.sh: + [[ ! -s /failed-services ]]          ← /failed-services non-empty → fails
mkosi-check-and-shutdown.service: Main process exited, code=exited, status=1/FAILURE
Failed to start mkosi-check-and-shutdown.service
Doing "poweroff" action instead of an "exit" emergency action
```
The unit is masked, but also required by `sysinit.target.wants`, hence the :boom: . This is almost surely an unrelated flake.

- [ubuntu-on-fedora](https://github.com/systemd/mkosi/actions/runs/27493239915/job/81262500021?pr=4360#step:11:47422) failed with a kernel panic "Initramfs unpacking failed: write error". I think this happens in `test_initrd`. This smells like the same class of bugs like https://bugzilla.redhat.com/show_bug.cgi?id=2305968 , where the initrd just became too big. This smells like fallout from the RAM reduction.

martinpitt:
Works now. The remaining three failures are the known "opensuse mirror" SNAFU, plus the arch/debian timeout [fixed in unstable](https://tracker.debian.org/news/1762094/accepted-pacman-package-manager-710-3-source-into-unstable/) (should go into testing next Tuesday).

@behrmann @daandemeyer can you please take a look? Danke! :pray:

martinpitt:
Added yet another concurrency fix -- now the tests in #4357 are green :tada:

mkosi/__init__.py
behrmann, line 1: So this doesn't really fill me with joy. It looks a bit racy, because what if something kills the agent before (the calling gpgconf will probably start it to kill it) or maybe it doesn't get killed at all, if something crashes and the cmdline step or somebody presses C-c? This would probably be better set in a try except block, but I also wondern: Wouldn't it make sense to just *always* kill the agent, since GPG always restarts one anyway? 

I also thought about whether it would be better to just switch the default to e.g. the sequoia sop interface, but I'm not sure that's available in CentOS.

tests/__init__.py
```diff
@@ -129,7 +136,9 @@ def boot(self, options: Sequence[str] = (), args: Sequence[str] = ()) -> Complet
 
         return result
 
-    def vm(self, options: Sequence[str] = (), args: Sequence[str] = ()) -> CompletedProcess:
+    def vm(
+        self, options: Sequence[str] = (), args: Sequence[str] = (), ram: str = "1536M"
```
behrmann: When this type of break happens we always force 
```suggestion
        self, options: Sequence[str] = (), 
        args: Sequence[str] = (), 
        ram: str = "1536M",
```
via trailling comma

tests/__init__.py
```diff
@@ -139,9 +148,13 @@ def vm(self, options: Sequence[str] = (), args: Sequence[str] = ()) -> Completed
...
                 # TODO: Drop once both Hyper-V bugs are fixed in Github Actions.
                 *(["--qemu-args=-cpu max,pcid=off"] if need_hyperv_workaround else []),
-                "--ram=2G",
+                f"--ram={ram}",
                 "--ephemeral=yes",
                 "--register=no",
+                "--machine",
+                self.machine,
+                "--output-directory",
+                self.output_dir,
```
behrmann: We avoid splitting command line options and their arguments
```suggestion
                f"--machine={self.machine}",
                f"--output-directory={self.output_dir}",
```
Since they will usually change together, this makes diffs smaller.

tests/__init__.py
```diff
@@ -117,6 +120,10 @@ def boot(self, options: Sequence[str] = (), args: Sequence[str] = ()) -> Complet
                 "--runtime-build-sources=no",
                 "--ephemeral=yes",
                 "--register=no",
+                "--machine",
+                self.machine,
+                "--output-directory",
+                self.output_dir,
```
behrmann: Same here
```suggestion
                f"--machine={self.machine}",
                f"--output-directory={self.output_dir}",
```

mkosi/config.py
```diff
@@ -5502,7 +5507,7 @@ def parse_config(
         return args, None, ()
 
     configdir = finalize_configdir(args.directory)
-    historydir = finalize_historydir(args)
+    historydir = finalize_historydir(args, context.cli.get("output_dir"))
```
behrmann, line 5513: I've not checked this yet, but I'm not sure this works generally. The `context` here is a `ParseContext` and there is more parsing having later (`context.parse_new_includes`). I think the output directory could be changed later so this might be brittle and the history file end up in places that are unexpected. I think the point why the history dir is in the top-level because that is the one thing safely known at this point. @daandemeyer will remember the design better than me.

mkosi/installer/__init__.py
```diff
@@ -153,22 +153,32 @@ def apivfs_script_cmd(cls, context: Context) -> list[PathString]:
...
+                network=True,
+                options=[
+                    *context.rootoptions(),
+                    *cls.mounts(context),
+                    *cls.options(root=context.root, apivfs=apivfs),
+                    *options,
+                ],
+            ) as sandbox,
+        ):
+            yield sandbox  # fmt: skip
```
behrmann: You can drop the `# fmt: skip` here. It's a noop in the original as well already.

mkosi/__init__.py
martinpitt: Right, I could make it a try:/finally. My main worry was that when *always* killing the GPG agent, running the test on a developer laptop would also kill their "production" keyring, i.e. afterwards you have to unlock your key again. Not the worst thing in the world, and maybe it's worth to just do that, avoid the race, and document that. I wouldn't personally mind, but it felt a bit rude..

tests/__init__.py
martinpitt: I fully agree, and it's also more natural to read. That's what I had -- but `ruff format` insists on this form :cry:

behrmann: Yeah, that's a downside of ruff, but we use `# fmt: skip` judiciously for these cases. I've also written this down in `docs/CODING_STYLE.md` ;)

mkosi/__init__.py
martinpitt: I simplified and robustified it, I think it's okay to always kill the agent.

tests/__init__.py
martinpitt: Ack, done.

tests/__init__.py
martinpitt: Ah, got it now, sorry! I was a bit confused as that `fmt: skip` has to be placed so far off. Done.

tests/__init__.py
martinpitt: See above, done.

mkosi/installer/__init__.py
martinpitt: Indeed! done.

mkosi/__init__.py
```diff
@@ -2565,11 +2565,22 @@ def calculate_signature_gpg(context: Context) -> None:
...
-        )
+        try:
+            run(
+                cmdline,
+                env=env,
+                sandbox=context.sandbox(options=options),
+            )
+        finally:
+            # gpg autostarts a gpg-agent to sign and, as the sandbox has no PID namespace, that agent is
+            # leaked when the sandbox goes away. Note this will also call a "real" user agent, but
```
```suggestion
            # leaked when the sandbox goes away. Note this will also kill a "real" user agent, but
```

martinpitt: Sort of my proof that I also wrote code and comments, not just Claude :see_no_evil: 

I'll fix that of course, but let's first wait for [the other thread](https://github.com/systemd/mkosi/pull/4360#discussion_r3411577693), I don't want to burden CI unnecessarily.

martinpitt: Fixed. I rebased on top of yesterday's rpmrepo signature PR.

mkosi/config.py
martinpitt: I still believe this is consistent. Note that this *only* applies to the CLI option. Includes write to `self.config`, not `self.cli.config`. If -O is on the CLI, the value at here is already final; later parsing cannot change it.

The one exception is https://github.com/systemd/mkosi/blob/41fea1dd8da8dc6ffc45ae9d4fec5158f30505a4/mkosi/config.py#L5122 - that moves the finalized values from `config` to `cli`, and at that point `historydir` is already computed.

However: If the output dir comes from `mkosi.conf` or `mkosi.local.conf` or an include, `context.cli.get("output_dir")` is `None`, so history stays in `configdir`  for both write and subsequent read, so it still works. I updated the comment and commit message to explain this better, I hope it is clearer now?

The current design prevents *any* parallel builds, i.e. `--output-dir` is only half-implemented and I'd argue that writing a build specific history into a global dir is also broken. This commit is vital for parallel tests, so I'd rather keep it in this PR and finish the discussion about it.

Thanks!

behrmann: > However: If the output dir comes from mkosi.conf or mkosi.local.conf or an include, context.cli.get("output_dir") is None, so history stays in configdir for both write and subsequent read, so it still works. I updated the comment and commit message to explain this better, I hope it is clearer now?

Mhm, but our `mkosi.conf` does set `OutputDirectory=mkosi.output`, so doesn't that kinda throw wrench into things? Even if we didn't have that, we document that we default to `mkosi.output`, if it exists, which would just move the problem with a single history file to a different directory or am I misunderstanding something?

What about a different ansatz and we add an option to just not write history. We could use that in CI and would sidestep the history problem, since I think that multiple images being built from the same config in the same directory is maybe only a problem in CI—at least at `$DAYJOB` we solve it by templating our config into different trees and for a repo this would equally work with different git worktrees.

martinpitt: Hmm, I understood the `OutputDirectory=mkosi.output` in mkosi.conf as the default if you don't specify anything, but a setting in mkosi.local.conf trumps that, and `--output-directory` on the CLI trumps both.

> which would just move the problem with a single history file to a different directory 

I'm afraid I don't understand this. The purpose is precisely that builds with different output directories would stop writing different histories into the same global dir.

I actually had `History=no` in an intermediate version. I discarded it as it felt too much like a hack: i.e. not testing what mkosi does by default. We could document that if you use `--output-directory` you need to disable history. If you are fine with that approach, I can give this another go (I don't remember the details how well this worked, as I discarded that path).

Using worktrees would also work for CI with some extra hacks -- we *do* want to re-use the main tools and image build across tests, it's just that some tests build their own additional image. So with work trees, the test setup would have to copy these directories too, which is quite some extra work (these are large).

Perhaps this warrants a high-bandwidth gmeet discussion? I'm on PTO for the rest of the week, just occasionally look into my laptop. But could do that next week. And perhaps in the meantime put up a workaround with `History=no` or `--history=no`?

martinpitt: I revisited this, and `--history=no` in tests' `Image.build()` really *can't work*: It only disables *writing* the history, but that doesn't stop `.vm()` and `.boot()` from *reading* an existing history file. That will have entirely the wrong information then -- e.g. the "main" build has format "directory" (via `mkosi.conf` default, `latest.json` is empty), but `test_initrd_luks` uses `--format=disk`; similar for any other specific option. This would only work if history was disabled for *all* builds and all options get copied between `build()` and `vm()`.

The only other option that I see is to add an explicit `--history-dir` to the CLI to make this more explicit. But honestly I don't like that -- it's conceptually redundant with `--output-dir`, like when would you ever set this to a different dir?

So I put back the original commit again for now.

martinpitt: ... but I reworded the commit message to be clearer, and include more reasoning.

behrmann: Okay, I follow your reasoning now and agree that it's necessary, but @daandemeyer should give his thumbs up on this one, too.

daandemeyer (approved):

bluca:
This breaks the systemd CI, so I'll file a revert: https://github.com/systemd/systemd/actions/runs/28126107457/job/83290221148?pr=42739

PR #4361, opened 2026-06-15
dnf: Tighten repo_gpgcheck, work around librepo bug

martinpitt:
For a while now we've had a lot of GPG repo signature test failures with the `opensuse` tools image, like [this](https://github.com/systemd/mkosi/actions/runs/27507702208/job/81302020690?pr=4360) or [this](https://github.com/systemd/mkosi/actions/runs/27507702208/job/81302020722?pr=4360). Initially this looked transient, so I first tried to mitigate that with retries (#4346). But turns out this doesn't help at all. The root cause is a bug in openSUSE's librepo, triggered by updating gnupg to a development release. I reported that to https://bugzilla.opensuse.org/show_bug.cgi?id=1268271 with a standalone reproducer and added a workaround.

This also uncovered some unpredictability with how we handle repository signatures in the first place -- mostly, it depended on the defaults in the *tools* image, not the *target distro*.

This series fixes these cases.

There is one more failure mode left, like [this](https://github.com/martinpitt/mkosi/actions/runs/27535189841/job/81383028780#step:7:3039):

> Repository 'oss' is invalid.
> [oss|https://download.opensuse.org/tumbleweed/repo/oss] Failed to retrieve new repository metadata. History:
>  Timeout exceeded when accessing 'https://cdn.opensuse.org/tumbleweed/repo/oss/repodata/29117db410d8ea1fdb5f93c866f048524e7ee8e4458df9db5389b4f02c9227d922fc963f5d8df558997e769713c96fdd78d66fe674c08ba74067352216a22315-appdata.xml.gz'.
>   Timeout reached Curl error (28)

This is "throw hands into the air". This URL *absolutely works* when I download it from my laptop (in Germany) even *while the test is hanging*. I suspect that either the SUSE CDN rate-limits GitHub (and possibly other origins) for scraper defense, or there is some geolocation going on and one of the mirrors is actually broken.

### DIFF
--- a/mkosi/distribution/fedora.py
+++ b/mkosi/distribution/fedora.py
@@ -149,6 +149,7 @@ class Installer(DistributionInstaller, distribution=Distribution.fedora):
 
     @classmethod
     def repositories(cls, context: Context) -> Iterable[RpmRepository]:
+        # Fedora doesn't sign its repository metadata (no repomd.xml.asc), so disable repo_gpgcheck.
         gpgurls = find_fedora_rpm_gpgkeys(context)
 
         mirror = context.config.mirror
@@ -165,7 +166,9 @@ class Installer(DistributionInstaller, distribution=Distribution.fedora):
             )
 
         if context.config.local_mirror:
-            yield RpmRepository("fedora", f"baseurl={context.config.local_mirror}", gpgurls)
+            yield RpmRepository(
+                "fedora", f"baseurl={context.config.local_mirror}", gpgurls, repo_gpgcheck=False
+            )
             return
 
         if context.config.release == "eln":
@@ -173,11 +176,21 @@ class Installer(DistributionInstaller, distribution=Distribution.fedora):
 
             for repo in ("AppStream", "BaseOS", "Extras", "CRB"):
                 url = f"baseurl={join_mirror(mirror, repo)}"
-                yield RpmRepository(repo.lower(), f"{url}/$basearch/os", gpgurls)
+                yield RpmRepository(repo.lower(), f"{url}/$basearch/os", gpgurls, repo_gpgcheck=False)
                 yield RpmRepository(
-                    f"{repo.lower()}-debuginfo", f"{url}/$basearch/debug/tree", gpgurls, enabled=False
+                    f"{repo.lower()}-debuginfo",
+                    f"{url}/$basearch/debug/tree",
+                    gpgurls,
+                    enabled=False,
+                    repo_gpgcheck=False,
                 )
-                yield RpmRepository(f"{repo.lower()}-source", f"{url}/source/tree", gpgurls, enabled=False)
+                yield RpmRepository(
+                    f"{repo.lower()}-source",
+                    f"{url}/source/tree",
+                    gpgurls,
+                    enabled=False,
+                    repo_gpgcheck=False,
+                )
         elif mirror:
             # Snapshot= implies the koji compose URL layout (compose/<release>/
             # Fedora-<Release>-<snapshot>/compose/Everything/...), regardless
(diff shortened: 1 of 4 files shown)


martinpitt:
FTR: Another way to fix this would be to just globally disable repo gpg checks for all of dnf. It's not being used a lot (basically only CentOS, and possibly RHEL if you use mkosi for it), it was effectively broken until now anyway due to the combination of `skip_unavailable` and the unpredictable `repo_gpgcheck`, and I'm not sure how much security it actually adds. (Remember, the *rpms* still get validated, that's `gpgcheck=1`). If you'd rather keep it simple, I'm happy to rework.

martinpitt:
Meh, [centos/arch](https://github.com/systemd/mkosi/actions/runs/27537658400/job/81400133285?pr=4361) and  [centos/debian](https://github.com/systemd/mkosi/actions/runs/27537658400/job/81400133381?pr=4361) have the same librepo issue. I didn't spot that in my reduced test matrix on my fork. I now created an upstream librepo bug report at https://github.com/rpm-software-management/librepo/issues/376 .

At this point it really doesn't make sense any more to restrict the hack to a particular tools distro --- over time this will creep everywhere. I'll disable repo gpg checks globally for "cross-distro" builds, but keep it as a separate commit with the bug pointer. We can re-enable this if/when this gets fixed and released. I'll keep it enable for "native-distro" builds, that makes sense, is useful, and really ought to work.

martinpitt:
Cool, all centos/opensuse tests green!

[arch/ubuntu failure](https://github.com/systemd/mkosi/actions/runs/27545058250/job/81416924068?pr=4361) is something entirely different:

> Could not receive response to TPM2_PCR_Allocate from swtpm: No such file or directory
> A error occurred. Authoring the TPM state failed.

and I'd like to keep that yak out of this PR. arch/debian is the usual pacman hang. Tomorrow the fix will go into Debian testing.

martinpitt:
@daandemeyer @behrmann WDYT? (Also see https://github.com/systemd/mkosi/pull/4361#issuecomment-4706609512)

mkosi/distribution/fedora.py
```diff
@@ -149,6 +150,12 @@ def install(cls, context: Context) -> None:
 
     @classmethod
     def repositories(cls, context: Context) -> Iterable[RpmRepository]:
+        # Fedora doesn't sign its repository metadata (no repomd.xml.asc).
+        for repo in cls._repositories(context):
+            yield dataclasses.replace(repo, repo_gpgcheck=False)
```
daandemeyer: This seems like a really roundabout way to go about setting the field. Why not just set it directly when constructing the dataclass instance?

martinpitt: I mostly added this indirection as there are umpteen instances of `RpmRepository` instantiation in repositories(), and adding this flag to all of them felt a bit DRY. I agree though that this is a bit ugly, and I'm fine to sprinkle two dozen `repo_gpgcheck=False` over the function -- that's at least explicit.

Option 2 would be to default to `False`, and only opt into the known-good distros (most centos repos and opensuse). But when adding new RPM distros it's too likely to forget that, and not validate repo sigs even though it may add it; felt better to fail if you forget to turn it off.

Option 3, as [mentioned above](https://github.com/systemd/mkosi/pull/4361#issuecomment-4707731933) , would be to just throw in the towel and disable repo sigs everywhere. That would replace the two topmost commits to basically an one-liner, but disable repo validation on centos/centos.

daandemeyer: To be honest I had no clue we were using it anywhere except suse. So I'm fine with disabling it everywhere on centos/fedora tbh.

martinpitt: I have another idea: Make `RpmRepository` signed by default, and introduce a subclass `UnsignedRpmRepository` with `repo_gpgcheck=False`. That avoids the 22 `repo_gpgcheck=False` repetitions and (more annoyingly) lots of line reflows as the line lengths would then exceed the 109 limit.

martinpitt: Done. Now the diff is larger, but the code stays (almost) the same size, and it's more explicit.

behrmann: At this point we're well into bikeshedding territory, but I'd prefer not subclassing, but calling `RpmRepository` with a `rpg_gpgcheck=False` keyword argument.

behrmann (approved):
Thanks for looking at this so closely, tremendous work. Thanks!

martinpitt:
Sadly, [the](https://github.com/systemd/mkosi/actions/runs/27552635716/job/81448281878?pr=4361) [three](https://github.com/systemd/mkosi/actions/runs/27552635716/job/81448282464?pr=4361) [failures](https://github.com/systemd/mkosi/actions/runs/27552635716/job/81448282235?pr=4361) are now all the "CDN timeout" bug. I'm out of ideas what to do about that one..

mkosi/distribution/fedora.py
martinpitt: OK - Green with pink stripes it is! :grin:

mkosi/installer/dnf.py
```diff
@@ -107,6 +108,15 @@ def setup(
...
             )
 
+        # TODO: librepo rejects valid repository metadata GPG signatures with gpgme 2.1 ("Bad GPG
+        # signature"): https://github.com/rpm-software-management/librepo/issues/376
+        # This happens a lot when building a distro which signs its repos (CentOS) on newer distros like
+        # Arch, Debian testing, or openSUSE Tumbleweed, with their newer gpgme.
+        # Thus until this is fixed, only validate repo GPG signatures when the distributions match. We can
+        # assume that distros test their own dnf repos :-)
+        # Revert once librepo gets fixed and into the distros.
+        repo_gpgcheck_broken = detect_distribution(context.config.tools())[0] != context.config.distribution
```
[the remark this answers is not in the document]
martinpitt: Sure! Added.

mkosi/installer/rpm.py
```diff
@@ -23,6 +23,11 @@ class RpmRepository:
     sslclientcert: Optional[Path] = None
     priority: Optional[int] = None
 
+    # True if the repository metadata (repomd.xml) is GPG-signed with a key we configure (e.g. CentOS).
+    # False for repositories that don't publish a repomd signature (e.g. Fedora, EPEL) or sign it with
+    # a key we don't have. rpm signatures (gpgcheck) are always verified.
+    repo_gpgcheck: bool = True
```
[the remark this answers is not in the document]
martinpitt: Oops yes, fixed.

daandemeyer (approved):

PR #4362, opened 2026-06-18
curl: Retry on failures

martinpitt:
`curl()` is being used for things like fetching Fedora rawhide's GPG key, opensuse's `/history/latest` or CentOS' `.composeinfo`. These are all volatile and thus hard to pre-download/cache. But we've seen several transient download failures like

> curl: (35) Recv failure: Connection reset by peer

Make these more robust with curl's retry (with exponential back-off) mechanism.

----

Should help with [these three flakes](https://github.com/systemd/mkosi/actions/runs/27721694120/job/82008778902?pr=4297#step:8:58344)

### DIFF
--- a/mkosi/curl.py
+++ b/mkosi/curl.py
@@ -39,6 +39,10 @@ def curl(config: Config, url: str, *, output_dir: Optional[Path] = None, log: bo
             *(["--remote-name"] if output_dir else []),
             "--no-progress-meter",
             "--fail",
+            # be resilient against transient network failures
+            "--retry", "5",
+            # include mid-transfer connection reset
+            "--retry-all-errors",
             *(["--silent"] if not log else []),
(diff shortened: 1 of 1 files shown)


mkosi/curl.py
```diff
@@ -39,6 +39,10 @@ def curl(config: Config, url: str, *, output_dir: Optional[Path] = None, log: bo
             *(["--remote-name"] if output_dir else []),
             "--no-progress-meter",
             "--fail",
+            # be resilient against transient network failures
+            "--retry", "5",
+            # include mid-transfer connection reset
+            "--retry-all-errors",
```
martinpitt, line 45: Wrt. the warning in the curl manpage: This never gets any "input" (no POST), these are all stateless GET calls.

daandemeyer (approved):

behrmann (approved):

PR #4364, opened 2026-06-18
action: download latest version of apk

craftyguy:
This uses the gitlab API to fetch the latest release of apk from the upstream apk-tools repo, instead of hardcoding a version.

Since gitlab API can be flaky, I put it in a retry loop. set -x in a subshell in the loop because you can never have too much info if/when things break. I also made the whole thing non-fatal if it still fails to get apk after retrying, because not all of them need apk (even if it means other later jobs using apk might fail). I added a warning for that situation so hopefully it makes it a bit more obvious why a later job using apk might fail if that was the case.

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -96,13 +96,25 @@ runs:
       # ubuntu-24.04 doesn't package apk, so fetch the official statically-linked binary from
       # upstream so mkosi can use it to populate the postmarketOS tools tree.
       #
+      # Since gitlab API can be unreliable, this is retried a few times if there
+      # is a failure when fetching. Since all jobs might use this action file,
+      # a failure to fetch APK is not treated as fatal, even if it means some
+      # jobs that use apk might fail later.
+      #
       # NOTE: apk is dropped into /usr/bin and not /usr/local/bin because later CI
       # scripts (like .github/workflows/ci.yml) clean up /usr/local
       shell: bash
       run: |
-        sudo curl -fsSL -o /usr/bin/apk https://gitlab.alpinelinux.org/api/v4/projects/5/packages/generic/v3.0.6/x86_64/apk.static
-        echo 'f1489e05bace7d7dd0a687fcd38d50b585ac660af4231668b123649bef3718c4  /usr/bin/apk' | sha256sum --check
-        sudo chmod +x /usr/bin/apk
+        (
+            set -ex
+            APK_TOOLS_API="https://gitlab.alpinelinux.org/api/v4/projects/5"
+            APK_VERSION=$(curl -fsSL --retry 5 --retry-all-errors "$APK_TOOLS_API/repository/tags?per_page=1" | jq -r '.[0].name')
+            APK_PKG_ID=$(curl -fsSL --retry 5 --retry-all-errors "$APK_TOOLS_API/packages?package_name=${APK_VERSION}&package_type=generic" | jq '.[] | select(.version == "x86_64") | .id')
+            APK_SHA256=$(curl -fsSL --retry 5 --retry-all-errors "$APK_TOOLS_API/packages/${APK_PKG_ID}/package_files" | jq -r '.[0].file_sha256')
+            sudo curl -fsSL --retry 5 --retry-all-errors -o /usr/bin/apk "$APK_TOOLS_API/packages/generic/${APK_VERSION}/x86_64/apk.static"
+            echo "${APK_SHA256}  /usr/bin/apk" | sha256sum --check
+            sudo chmod +x /usr/bin/apk
+        ) || echo "WARNING: Failed to fetch latest apk!"
 
     - name: Dependencies
       shell: bash


action.yaml
```diff
@@ -96,13 +96,29 @@ runs:
...
+      #
       # NOTE: apk is dropped into /usr/bin and not /usr/local/bin because later CI
       # scripts (like .github/workflows/ci.yml) clean up /usr/local
       shell: bash
       run: |
-        sudo curl -fsSL -o /usr/bin/apk https://gitlab.alpinelinux.org/api/v4/projects/5/packages/generic/v3.0.6/x86_64/apk.static
-        echo 'f1489e05bace7d7dd0a687fcd38d50b585ac660af4231668b123649bef3718c4  /usr/bin/apk' | sha256sum --check
-        sudo chmod +x /usr/bin/apk
+        for _ in 1 2 3; do
+            (
```
martinpitt: This is a bit too coarse. This shouldn't retry e.g. the /tags page if the apk download fails. Instead, use curl's builtin `--retry*` options like in commit bb87e48401b80ca87f0a62fc362a1564d9eb5765

martinpitt (changes requested):
Thanks!

action.yaml
craftyguy: ohhh yeah this is much nicer, thanks for letting me know :)

craftyguy: ok I reworked this to drop the for loop and use curl's --retry param

action.yaml
```diff
@@ -96,13 +96,25 @@ runs:
...
+        (
+            set -ex
+            APK_TOOLS_API="https://gitlab.alpinelinux.org/api/v4/projects/5"
+            APK_VERSION=$(curl -fsSL --retry 5 --retry-all-errors "$APK_TOOLS_API/repository/tags?per_page=1" | jq -r '.[0].name')
+            APK_PKG_ID=$(curl -fsSL --retry 5 --retry-all-errors "$APK_TOOLS_API/packages?package_name=${APK_VERSION}&package_type=generic" | jq '.[] | select(.version == "x86_64") | .id')
+            APK_SHA256=$(curl -fsSL --retry 5 --retry-all-errors "$APK_TOOLS_API/packages/${APK_PKG_ID}/package_files" | jq -r '.[0].file_sha256')
+            sudo curl -fsSL --retry 5 --retry-all-errors -o /usr/bin/apk "$APK_TOOLS_API/packages/generic/${APK_VERSION}/x86_64/apk.static"
+            echo "${APK_SHA256}  /usr/bin/apk" | sha256sum --check
+            sudo chmod +x /usr/bin/apk
+        ) || echo "WARNING: Failed to fetch latest apk!"
```
martinpitt, line 117: I really don't like ignoring errors in workflows and tests, it makes things unpredictable and unnecessarily hard. It hasn't hurt us much before, so the simplest way would be to just keep the previous "must succeed" behaviour. Alternatively, this could introduce an option whether to install apk, and for backwards compatibility it could default to on. Then users can switch it off to skip the step if they don't need `apk`.

action.yaml
```diff
@@ -96,13 +96,25 @@ runs:
...
       shell: bash
       run: |
-        sudo curl -fsSL -o /usr/bin/apk https://gitlab.alpinelinux.org/api/v4/projects/5/packages/generic/v3.0.6/x86_64/apk.static
-        echo 'f1489e05bace7d7dd0a687fcd38d50b585ac660af4231668b123649bef3718c4  /usr/bin/apk' | sha256sum --check
-        sudo chmod +x /usr/bin/apk
+        (
+            set -ex
+            APK_TOOLS_API="https://gitlab.alpinelinux.org/api/v4/projects/5"
+            APK_VERSION=$(curl -fsSL --retry 5 --retry-all-errors "$APK_TOOLS_API/repository/tags?per_page=1" | jq -r '.[0].name')
+            APK_PKG_ID=$(curl -fsSL --retry 5 --retry-all-errors "$APK_TOOLS_API/packages?package_name=${APK_VERSION}&package_type=generic" | jq '.[] | select(.version == "x86_64") | .id')
```
behrmann, line 112: While testing this out this returned two lines for me, you probably want to ensure that only the latest pkg ID wins.

action.yaml
```diff
@@ -96,13 +96,25 @@ runs:
...
       # scripts (like .github/workflows/ci.yml) clean up /usr/local
       shell: bash
       run: |
-        sudo curl -fsSL -o /usr/bin/apk https://gitlab.alpinelinux.org/api/v4/projects/5/packages/generic/v3.0.6/x86_64/apk.static
-        echo 'f1489e05bace7d7dd0a687fcd38d50b585ac660af4231668b123649bef3718c4  /usr/bin/apk' | sha256sum --check
-        sudo chmod +x /usr/bin/apk
+        (
+            set -ex
+            APK_TOOLS_API="https://gitlab.alpinelinux.org/api/v4/projects/5"
+            APK_VERSION=$(curl -fsSL --retry 5 --retry-all-errors "$APK_TOOLS_API/repository/tags?per_page=1" | jq -r '.[0].name')
```
behrmann, line 111: I'm not sure what the GitLab API guarantees here, but the endpoint returns also rc releases and in between the v3.0.0 rc releases there was v2.14.10, so I don't assume any sorting.

Maybe something like this will do
```suggestion
            APK_VERSION=$(curl -fsSL --retry 5 --retry-all-errors "$APK_TOOLS_API/repository/tags?per_page=1" | jq -r '.[].name' | sort -V | tail -1) 
```

action.yaml
behrmann: I think hardcoding the current version as baseline might be fine.

behrmann (commented):
Thanks @craftyguy!

bluca:
superseded by https://github.com/systemd/mkosi/pull/4375

Issue #4365, opened 2026-06-19
openSUSE Tumbleweed CI failures: US mirror sync-lag → 404/timeout on fresh snapshots; persistent individual rpm download failures


martinpitt:
## Summary

mkosi CI openSUSE Tumbleweed jobs fail on metadata/RPM downloads. This happens only on the US-hosted GitHub Actions runners, never from a German laptop.

`download.opensuse.org` and `cdn.opensuse.org` are both MirrorCache geo-redirectors. They route US runners to US community mirrors that lag the rolling Tumbleweed publish. Right after a snapshot publishes, `repomd.xml` references checksum-named files the US mirrors haven't rsync'd yet (or already pruned). `zypper`/`dnf` try every routed mirror, get 404/timeout on all, and abort. EU mirrors near the Nuremberg origin sync fast enough that Germany never hits the window.

This started a few weeks ago, and almost every PR has had failures since then. Example runs: [during image build](https://github.com/systemd/mkosi/actions/runs/27666386674/job/81821375512), [during tests](https://github.com/systemd/mkosi/actions/runs/27737329026/job/82057068488) -- happens anywhere openSUSE packages get resolved/installed.

### 1. `zypper refresh` — metadata timeout (exit 4)

```
Retrieving: https://mirror.us.leaseweb.net/.../14be8958...-appdata.xml.gz [..not found]
Retrieving: https://codingflyboy.mm.fcix.net/.../14be8958...-appdata.xml.gz [..not found]
Retrieving: https://mirror.umd.edu/.../14be8958...-appdata.xml.gz [not found]
Retrieving: https://mirror.clarkson.edu/.../14be8958...-appdata.xml.gz [..not found]
Retrieving: https://cdn.opensuse.org/.../14be8958...-appdata.xml.gz [..error]
Repository 'oss' is invalid.
[oss|https://download.opensuse.org/tumbleweed/repo/oss] Failed to retrieve new repository metadata.
History:
 - Timeout exceeded when accessing '.../14be8958...-appdata.xml.gz'.
   Timeout reached Curl error (28)
‣ "zypper ... refresh" returned non-zero exit code 4.
```

### 2. `dnf makecache` — metadata 404 (exit 1)

```
error: Status code: 404 for https://slc-mirror.opensuse.org/tumbleweed/repo/oss/repodata/21cabd02...-primary.xml.zst
Errors during downloading metadata for repository 'oss':
[...]

PR #4366, opened 2026-06-19
ci: Pin openSUSE to second-to-last Tumbleweed snapshot; use metalink for the default mirror

martinpitt:
See #4365 for details. It will fix [this kind of failure](https://github.com/systemd/mkosi/actions/runs/27666386674/job/81821375512) and also [this kind](https://github.com/martinpitt/mkosi/actions/runs/27806672930/job/82287872691).


martinpitt:
Wow, a green CI run! That's been a while :tada:

martinpitt:
It works, but it's quite a hack. I'm currently testing something which will hopefully work better, so draft → don't land that just yet.

martinpitt:
See https://github.com/systemd/mkosi/issues/4365#issuecomment-4752581988 -- sadly, only using the metalink is structurally broken. We need both, I'm afraid :cry: and dear openSUSE, pleeease fix your mirror situation. :pray:

martinpitt:
The [opensuse extension tests all failed](https://github.com/systemd/mkosi/actions/runs/27833079144?pr=4366), they don't use the snapshot of the main image build. Fixed.

martinpitt:
Quo vadis CI? Pushed some commit message cleanups to hopefully tickle actions.

martinpitt:
two failures, both with the same cause:

*  [`test_extension[confext]`](https://github.com/systemd/mkosi/actions/runs/27864917310/job/82467366909?pr=4366): correct snapshot now, but failed to connect to slc-mirror.opensuse.org
*  [image build](https://github.com/systemd/mkosi/actions/runs/27862117430/job/82459855004?pr=4366): zypper fetches metainfo from the single top metalink mirror (slc-mirror, as above) and doesn't fail over for repomd. This mirror keeps being advertised as top US one, and keeps being faulty :cry: 

On the good side, the rpm download failover and the consistent snapshot selection now work!

Neither me nor Claude have any idea what else to try now. So I think that's about as good as it gets. At least slc-mirror is reachable again, so if @behrmann or @daandemeyer could retry that failed test, I'd appreciate -- it should hopefully go green now.

Also, ready for review! @Firstyear @cyphar I think this does what we discussed in #4365 -- but second pair of eyes appreciated!

tools/integration-test-setup.sh
```diff
@@ -52,6 +54,35 @@ Environment=SYSTEMD_REPART_MKFS_OPTIONS_EROFS="--quiet"
 KVM=yes
 EOF
 
+# openSUSE Tumbleweed mirrors frequently lag behind freshly published snapshots, especially to US based mirrors.
+# Pin the second-to-last snapshot (see https://github.com/systemd/mkosi/issues/4365). Snapshot= is only
+# supported for openSUSE on x86-64, and is not propagated to the tools tree, so set it separately for both.
+if [ "$(uname -m)" = "x86_64" ] && { [ "$distribution" = "opensuse" ] || [ "$tools" = "opensuse" ]; }; then
+    snapshot="$(curl -fsSL --retry 5 https://download.opensuse.org/history/ |
+        grep -oE 'href="\./20[0-9]{6}/"' | grep -oE '20[0-9]{6}' | sort -u | tail -2 | head -1)"
```
behrmann: As somebody guilty of regularly parsing out things with grep, this would be the shorter version
```suggestion
        grep -oP 'href="\./\K20[0-9]{6}(?=/") | sort -u | tail -2 | head -1)"
```

I do like the optimism of your regex, though. :)

tools/integration-test-setup.sh
```diff
@@ -52,6 +54,35 @@ Environment=SYSTEMD_REPART_MKFS_OPTIONS_EROFS="--quiet"
...
+    if [ -z "$snapshot" ]; then
+        echo "Failed to determine an openSUSE snapshot to pin" >&2
+        exit 1
+    fi
+    echo "Pinning openSUSE snapshot $snapshot" >&2
```
behrmann: Nit:
```suggestion
    echo "Pinning openSUSE snapshot ${snapshot}" >&2
```

behrmann (commented):
Thanks! I've not seen CI this green in a long time.

tools/integration-test-setup.sh
martinpitt: Clever, thanks! TIL about the `\K` "keep out" flag!

tools/integration-test-setup.sh
martinpitt: Argh! Done.

tools/integration-test-setup.sh
martinpitt: optimism> you mean that this will "only" break in 84 years? I do hope by then there will be something better than this ugly hack :sweat_smile:

behrmann: Only 74 years, we could still see that. ;)

behrmann:
So, LGTM in principle, but opensuse/arch and opensuse/opensuse are still running into timeouts. fedora/arch is running into a cooldown it can't handle, so that's different, but I'll leave looking at the suse failures for your, by now, trained eye.

martinpitt:
@behrmann that's the [same issue as above](https://github.com/systemd/mkosi/pull/4366#issuecomment-4757215647): slc-mirror.opensuse.org doesn't respond:

> zypper fetches metainfo from the single top metalink mirror (slc-mirror, as above) and doesn't fail over for repomd. This mirror keeps being advertised as top US one, and keeps being faulty

sadly, zypper/opensuse repos seem to have no solution for that, I'm at my wit's end :cry: It seems to be working again now.

fedora/arch cooldown: same, no idea -- we can't immediately retry a 429 error, that would only put even more load on the mirror.

martinpitt:
I sent #4368 with a version without the metalink change. Since both approaches are broken in different ways, I suppose when in doubt use the less intrusive one.

martinpitt:
I am shelving this. metalink would be better (that's what zypper is *designed* to handle well), but with that being so unreliable, there's no point in this.

PR #4368, opened 2026-06-21
ci: Pin openSUSE to second-to-last Tumbleweed snapshot

martinpitt:
This is a version of #4366 without the third commit that moves to metalink. metalink is good for using more mirrors, and auto-falling back away from broken ones (of which there are plenty). However, the metalink entry point `slc-mirror.opensuse.org` is *itself* unreliable: [example](https://github.com/systemd/mkosi/actions/runs/27864917310/job/82467366909?pr=4366), [example](https://github.com/systemd/mkosi/actions/runs/27862117430/job/82459855004?pr=4366), [example](https://github.com/systemd/mkosi/actions/runs/27894304758/job/82543218529?pr=4366).

So we now need to pick our poison: Fail on`slc-mirror` as above, or fail on individual broken mirrors [like this](https://github.com/martinpitt/mkosi/actions/runs/27806672930/job/82287872691).

Both this PR and #4366 improve the current situation where almost *every* opensuse run fails, but neither can bring reliable test runs :cry: that really needs fixing the opensuse mirror situation.

See #4365 for details.


martinpitt:
As expected, [that failure](https://github.com/systemd/mkosi/actions/runs/27908910169/job/82586924138?pr=4368) is due to a broken/hanging mirror and zypper not falling back to a different one. Without mirrorlink zypper has no other fallbacks, so that's now in "we have to live with this" territory. @behrmann can we land this one for now, and I shelve #4366?

behrmann (approved):

behrmann:
It's already a very decent improvement over the status quo ante. Thanks, Martin!

PR #4371, opened 2026-06-24
action: don't fail if apk cannot be downloaded

behrmann:
@craftyguy @bluca requested this, because this was failing CI elsewhere, I assume you don't mind, since  #4364 handles this similary.

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -100,9 +100,13 @@ runs:
       # scripts (like .github/workflows/ci.yml) clean up /usr/local
       shell: bash
       run: |
-        sudo curl -fsSL -o /usr/bin/apk https://gitlab.alpinelinux.org/api/v4/projects/5/packages/generic/v3.0.6/x86_64/apk.static
-        echo 'f1489e05bace7d7dd0a687fcd38d50b585ac660af4231668b123649bef3718c4  /usr/bin/apk' | sha256sum --check
-        sudo chmod +x /usr/bin/apk
+        if sudo curl -fsSL -o /usr/bin/apk https://gitlab.alpinelinux.org/api/v4/projects/5/packages/generic/v3.0.6/x86_64/apk.static
+        then
+            echo 'f1489e05bace7d7dd0a687fcd38d50b585ac660af4231668b123649bef3718c4  /usr/bin/apk' | sha256sum --check
+            sudo chmod +x /usr/bin/apk
+        else
+            echo "WARNING: Failed to fetch latest apk!"
+        fi
 
     - name: Dependencies
       shell: bash


bluca (approved):
thanks

behrmann:
@martinpitt You we're opposed to that kind of error handling. I just quickly put this up, but if you have a better idea that would be lovely as well.

bluca:
It can always be further fixed up, but a lot of CI jobs are failing now due to this, so let's merge it

martinpitt:
Yeah, this makes the whole thing rather useless.  I proposed to make it an action config knob to opt out:

```yaml
 - uses: mkosi/...
   with:
       apk: false
```

Since this PR landed, the new accepted default is "apk is not installed" - so we may just as well default to "false" now, and people who actually *do* want to build an arch tools image in a workflow would enable it. For them, an unpredictable "apk might be there" is not helpful, and for any others that install step is just an unnecessarily failure point and delay.

I'm happy to send a PR, but I didn't yet as there was no reply to https://github.com/systemd/mkosi/pull/4364#discussion_r3445457068 so far.

behrmann:
Happy to take the PR.

bluca:
> Yeah, this makes the whole thing rather useless. I proposed to make it an action config knob to opt out:
> 
> ```yaml
>  - uses: mkosi/...
>    with:
>        apk: false
> ```
> 
> Since this PR landed, the new accepted default is "apk is not installed" - so we may just as well default to "false" now, and people who actually _do_ want to build an arch tools image in a workflow would enable it. For them, an unpredictable "apk might be there" is not helpful, and for any others that install step is just an unnecessarily failure point and delay.
> 
> I'm happy to send a PR, but I didn't yet as there was no reply to [#4364 (comment)](https://github.com/systemd/mkosi/pull/4364#discussion_r3445457068) so far.

Please make it so that it's a per-job option, and not per-CI, so that if it fails, only the job actually wanting to use it fails, and not the entire set of CIs jobs using mkosi

martinpitt:
@bluca Realistically it's a per-step option (on the particular `uses:`). I think that's what you mean, just want to confirm? Thanks!

bluca:
Unfortunately that means it can fail all jobs regardless of whether they need it, so it needs to be more granular than that or we are back to square one - about 1/3 of CI runs are failing because of this right now

martinpitt:
@bluca oh, I see, matrix job. It think you want something like `apk: {{ distro == 'arch' ? true : false }}` (that's most probably not the correct jinja2 syntax, but you get the idea)

bluca:
Yeah with the conditional for the job actually using it, so 'pmos' or so I imagine

martinpitt:
@bluca  @behrmann FYI #4374

PR #4373, opened 2026-06-24
Revert "Put build history into the output directory"

bluca:
Breaks the systemd CI, and also seems to make passing --output-directory mandatory on the command line if it is defined in the config, which is just horrible UX:

‣ /home/runner/work/systemd/systemd/mkosi/mkosi.local.conf: Setting OutputDirectory specified by specifier '%O' in %O/ovmf_vars_shim.fd is not yet set, ignoring
‣ /home/runner/work/systemd/systemd/mkosi/mkosi.local.conf: Setting OutputDirectory specified by specifier '%O' in %O/ovmf_vars_shim.fd is not yet set, ignoring
‣ Firmware variables file /ovmf_vars_shim.fd does not exist

https://github.com/systemd/systemd/actions/runs/28126107457/job/83290221148?pr=42739

This reverts commit da49fe976c7398524202cfedf93c8ff332d4eaf2.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -5445,15 +5445,7 @@ def want_default_initrd(config: Config) -> bool:
     return Path("default") in config.initrds
 
 
-def finalize_historydir(args: Args, output_dir: Optional[Path] = None) -> Path:
-    # When an output dir is given on the CLI, store the build history there so that concurrent builds with
-    # different output dirs don't clobber a shared history. Don't check the finalized OutputDirectory=
-    # config, only the CLI value: the former isn't known yet here (config files and includes are
-    # parsed later) and vm/boot can't see it anyway since they recover the config from the history instead of
-    # parsing it. An output dir set only in config files keeps the history in the config dir.
-    if output_dir is not None:
-        return output_dir / ".mkosi-private/history"
-
+def finalize_historydir(args: Args) -> Path:
     configdir = finalize_configdir(args.directory)
     return (configdir or Path.cwd()) / ".mkosi-private/history"
 
@@ -5510,7 +5502,7 @@ def parse_config(
         return args, None, ()
 
     configdir = finalize_configdir(args.directory)
-    historydir = finalize_historydir(args, context.cli.get("output_dir"))
+    historydir = finalize_historydir(args)
 
     if have_history(args, historydir):
         history = Config.from_partial_json((historydir / "latest.json").read_text())
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -126,13 +126,12 @@ class Image:
                 "--register=no",
                 "--machine",
                 self.machine,
-                "--output-directory", self.output_dir,
                 *options,
             ],
             args,
             stdin=sys.stdin if sys.stdin.isatty() else None,
             check=False,
-        )  # fmt: skip
+        )
 
         if result.returncode != 123:
             raise subprocess.CalledProcessError(result.returncode, result.args, result.stdout, result.stderr)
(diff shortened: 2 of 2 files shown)


martinpitt:
Thanks @bluca -- I'll put that on my todo list then, check how systemd uses that, and have another go at that.

behrmann:
Unfortunate.

martinpitt:
@bluca I'd like to reproduce this on my fork, but want to understand it first. systemd pins mkosi by SHA. The second-last time happened in https://github.com/systemd/systemd/commit/936a79084719de0cb9f5afc4da48b795b06cb109 on June 4, which didn't include any of #4360. The next commit was https://github.com/systemd/systemd/commit/45f561bf12aef708eeac5520ae3a137dca2205d0 yesterday, which included both #4360 *and* this revert. So how did landing 4360 spontaneously break systemd CI? Thanks in advance!

martinpitt:
@bluca unping -- of course that happened in a PR, specifically https://github.com/systemd/systemd/pull/42739 . Sorry for the noise! :hugs:

martinpitt:
Got it -- sadly this is already quite fragile, and the change made that latent bug explode. `mkosi.local.conf` is parsed before `mkosi.conf`, so a `%O` in `mkosi.local.conf` cannot see the `OutputDirectory=` defined in mkosi.conf. It only resolves because a prior build's history pre-populates output_dir into context.cli before config parsing. On a clean systemd checkout this is visible even without the problematic commit (i.e. on current main). reproducing what CI configures:

```
❱❱❱ cat mkosi/mkosi.local.conf 
[Distribution]
Distribution=fedora

[Content]
Packages=shim-signed
ShimBootloader=signed

[Runtime]
Firmware=uefi-secure-boot
FirmwareVariables=%O/ovmf_vars_shim.fd

❱❱❱ mkosi summary
‣ /home/martin/upstream/systemd/mkosi/mkosi.local.conf: Setting OutputDirectory specified by specifier '%O' in %O/ovmf_vars_shim.fd is not yet set, ignoring
Firmware Variables: /ovmf_vars_shim.fd
```

:arrow_up: see the wrong path  there.

The subsequent `mkosi build` would then "seed" `--output-directory` into the build history file, so that `%O` resolves it for the following `mkosi vm/boot`.

[systemd's meson.build](https://github.com/systemd/systemd/blob/d2a704388f846f2b3d3550c19ce7af0e9dc360b0/meson.build#L2434)  target passes `--output-dir <builddir>/mkosi.output` on the CLI, while every consume step (mkosi summary, the shim/VM runs) passes no `-O` and relies on `OutputDirectory=../build/mkosi.output` in mkosi.conf. That's pretty close to what the original commit advertised: 

> As a consequence, `mkosi vm` (and the other verbs that consume a previous build) now require `-O`/`--output-directory` when the build used one. This is a behaviour change, but unbreaks having more than one output dir.

... except I didn't consider *mixing* `OutputDirectory=` in the config and using `-O` on the CLI. AFAICS, they really *have* to coincide, otherwise calling `mkosi` in systemd sometimes with `-O` and sometimes without would just result in a mess.

It would be nice to only do one *or* the other in systemd to avoid the redundancy, but I think both are useful: the config option for human developers running `mkosi vm` outside of meson after a build; and the meson CLI option for making sure the images go into the configured build tree. Now, this relies on the build tree literally being `build`, as otherwise `OutputDirectory=../build/mkosi.output` would disagree with the -O option. (That's the redundancy).

But I think we can fix mkosi for this specific case.

martinpitt:
I sent #4376 with a new version of this. This time with unit tests, and confirmed to work against both mkosi/barrage and also systemd's own CI (ran on my fork).

PR #4374, opened 2026-06-25
action: Make apk installation configurable and reliable again

martinpitt:
Commit 5ef262bc53c was a quick-fix to unbreak systemd's CI, but it's not a long-term fix: it makes the apk installation unreliable and unpredictable.

Revert it, and instead add an explicit `apk` action option to install it explicitly. As 5ef262bc53c landed, let it default to false (as arguably it's not needed for most test runs).

Enable it in the test matrix when building a postmarketos tools tree, as we need `apk` on the Ubuntu host to bootstrap it. For any other test scenario, apk will be installed into and run from the tools tree.

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -1,6 +1,12 @@
 name: setup-mkosi
 description: Install mkosi
 
+inputs:
+  apk:
+    description: Install the apk package manager (required to bootstrap a postmarketOS tools tree)
+    required: false
+    default: "false"
+
 runs:
   using: composite
   steps:
@@ -98,15 +104,12 @@ runs:
       #
       # NOTE: apk is dropped into /usr/bin and not /usr/local/bin because later CI
       # scripts (like .github/workflows/ci.yml) clean up /usr/local
+      if: ${{ inputs.apk == 'true' }}
       shell: bash
       run: |
-        if sudo curl -fsSL -o /usr/bin/apk https://gitlab.alpinelinux.org/api/v4/projects/5/packages/generic/v3.0.6/x86_64/apk.static
-        then
-            echo 'f1489e05bace7d7dd0a687fcd38d50b585ac660af4231668b123649bef3718c4  /usr/bin/apk' | sha256sum --check
-            sudo chmod +x /usr/bin/apk
-        else
-            echo "WARNING: Failed to fetch latest apk!"
-        fi
+        sudo curl -fsSL -o /usr/bin/apk https://gitlab.alpinelinux.org/api/v4/projects/5/packages/generic/v3.0.6/x86_64/apk.static
+        echo 'f1489e05bace7d7dd0a687fcd38d50b585ac660af4231668b123649bef3718c4  /usr/bin/apk' | sha256sum --check
+        sudo chmod +x /usr/bin/apk
 
     - name: Dependencies
       shell: bash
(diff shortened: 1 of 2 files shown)


martinpitt:
@bluca FYI. This also demonstrates how to enable the option conditionally on the matrix. I think that should work the same way in https://github.com/systemd/systemd/blob/main/.github/workflows/mkosi.yml#L92 -- however, I don't see postmarketos/pmos or alpine there at all? You suggested pmos was in there.

bluca:
> @bluca FYI. This also demonstrates how to enable the option conditionally on the matrix. I think that should work the same way in https://github.com/systemd/systemd/blob/main/.github/workflows/mkosi.yml#L92 -- however, I don't see postmarketos/pmos or alpine there at all? You suggested pmos was in there.

It's in another job (that still uses mkosi for the tools tree): https://github.com/systemd/systemd/blob/364bc151db6ae5e7410366843be860c7b60599c7/.github/workflows/unit-tests.yml#L65

The integration test job is WIP at https://github.com/systemd/systemd/pull/39823

martinpitt:
@bluca ah, thanks. That doesn't even need to be conditional there, just a static `apk: true` (as it has a static `ToolsTreeDistribution=postmarketos`)

bluca:
For that one yes, for the WIP ones it's mixed though

action.yaml
```diff
@@ -98,15 +104,12 @@ runs:
       #
       # NOTE: apk is dropped into /usr/bin and not /usr/local/bin because later CI
       # scripts (like .github/workflows/ci.yml) clean up /usr/local
+      if: ${{ inputs.apk == 'true' }}
```
behrmann, line 107: The vagaries of using YAML for programming. The default for `inputs.apk` above is the string `"false"` and we compare it here to the string `'true'` (I think double quotes support escape sequences and single quotes don't?) and in `ci.yaml` we pass it `${{ matrix.tools == 'postmarketos' }}`, which I think (hope?) would be a YAML bool? It all seems to work somehow, as all balls of YAML tend to do at least after a while, but is it possible to just do it all with bools here? That's more a curio question, I'm not quite convinced it's a sane question.

behrmann (approved):
Thanks a lot @martinpitt!

From the sheer looks of it, this lgtm, since I can see it working in CI, but a question just to be on the safe side:

bluca (approved):

action.yaml
martinpitt: Ah, gotta love yaml/GitHub. Honestly the "true" is sort of fake. Composite action inputs are always strings, [as they are stored as env vars](https://docs.github.com/en/actions/reference/workflows-and-actions/metadata-syntax#inputs). there is sadly no boolean type. So coercing `{{ inputs.apk }}` into a boolean would always be true. Even "0" is truthy as a string, so that doesn't help either.

Another possibility would be "something" vs. "" (empty string). But a comparison like `${{ matrix.tools == 'postmarketos' }}` evaluates to `false`, which is coerced into the string "false", which is truthy. What a mess..

(I asked claude too, it doesn't have a better idea either..)

behrmann: Well, if it works, it works and we'll just hold our noses, close our eyes and pray.

martinpitt: @behrmann no no -- we call that "carefully designed and QEed"!!1! :see_no_evil:

PR #4375, opened 2026-06-25
action: download latest version of apk (with fallback)

behrmann:
This is a version of #4364 with my comments worked in. 

/cc @craftyguy 

Supersedes: #4364

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -102,13 +102,30 @@ runs:
       # ubuntu-24.04 doesn't package apk, so fetch the official statically-linked binary from
       # upstream so mkosi can use it to populate the postmarketOS tools tree.
       #
+      # Since GitLab's API can be unreliable, this is retried a few times if there
+      # is a failure when fetching.
+      #
       # NOTE: apk is dropped into /usr/bin and not /usr/local/bin because later CI
       # scripts (like .github/workflows/ci.yml) clean up /usr/local
       if: ${{ inputs.apk == 'true' }}
       shell: bash
       run: |
-        sudo curl -fsSL -o /usr/bin/apk https://gitlab.alpinelinux.org/api/v4/projects/5/packages/generic/v3.0.6/x86_64/apk.static
-        echo 'f1489e05bace7d7dd0a687fcd38d50b585ac660af4231668b123649bef3718c4  /usr/bin/apk' | sha256sum --check
+        set +e
+        APK_TOOLS_API="https://gitlab.alpinelinux.org/api/v4/projects/5"
+        APK_VERSION=$(curl -fsSL --retry 5 --retry-all-errors "${APK_TOOLS_API}/repository/tags" | jq -r '.[].name' | grep -v '_rc' | sort -V | tail -1)
+        APK_PKG_ID=$(curl -fsSL --retry 5 --retry-all-errors "${APK_TOOLS_API}/packages?package_name=${APK_VERSION}&package_type=generic" | jq '.[] | select(.version == "x86_64") | .id' | sort -n | tail -1)
+        APK_SHA256=$(curl -fsSL --retry 5 --retry-all-errors "${APK_TOOLS_API}/packages/${APK_PKG_ID}/package_files" | jq -r '.[0].file_sha256')
+        set -e
+
+        # If the API fails, fall back to a known-good version
+        if [ -z "$APK_VERSION" ] || [ -z "$APK_SHA256" ] || [ "$APK_SHA256" = null ]
+        then
+            APK_VERSION='v3.0.6'
+            APK_SHA256='f1489e05bace7d7dd0a687fcd38d50b585ac660af4231668b123649bef3718c4'
+        fi
+
+        sudo curl -fsSL --retry 5 --retry-all-errors -o /usr/bin/apk "$APK_TOOLS_API/packages/generic/${APK_VERSION}/x86_64/apk.static"
+        echo "${APK_SHA256}  /usr/bin/apk" | sha256sum --check
         sudo chmod +x /usr/bin/apk
 
     - name: Dependencies
--- a/action.yaml
+++ b/action.yaml
@@ -112,8 +112,8 @@ runs:
       run: |
         set +e
         APK_TOOLS_API="https://gitlab.alpinelinux.org/api/v4/projects/5"
-        APK_FALLBACK_VERSION='v3.0.6'
-        APK_FALLBACK_SHA256='f1489e05bace7d7dd0a687fcd38d50b585ac660af4231668b123649bef3718c4'
+        APK_FALLBACK_VERSION='v3.0.7'
+        APK_FALLBACK_SHA256='c07bf5356eacc9dd7a8c56bc537f46702f007170287403299d52a04264e74b3c'
 
         APK_VERSION=$(curl -fsSL --retry 5 --retry-all-errors --retry-max-time 30 "${APK_TOOLS_API}/repository/tags" | jq -r '.[].name' | grep -v '_rc' | sort -V | tail -1)
         if [[ "$APK_VERSION" != "$APK_FALLBACK_VERSION" ]]
--- a/action.yaml
+++ b/action.yaml
@@ -112,16 +112,22 @@ runs:
       run: |
         set +e
         APK_TOOLS_API="https://gitlab.alpinelinux.org/api/v4/projects/5"
-        APK_VERSION=$(curl -fsSL --retry 5 --retry-all-errors "${APK_TOOLS_API}/repository/tags" | jq -r '.[].name' | grep -v '_rc' | sort -V | tail -1)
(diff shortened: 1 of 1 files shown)


action.yaml
```diff
@@ -102,13 +102,29 @@ runs:
       # ubuntu-24.04 doesn't package apk, so fetch the official statically-linked binary from
       # upstream so mkosi can use it to populate the postmarketOS tools tree.
       #
+      # Since GitLab's API can be unreliable, this is retried a few times if there
+      # is a failure when fetching. Since all jobs might use this action file,
+      # a failure to fetch APK is not treated as fatal, even if it means some
```
bluca: outdated comment?

action.yaml
```diff
@@ -102,13 +102,29 @@ runs:
...
       shell: bash
       run: |
-        sudo curl -fsSL -o /usr/bin/apk https://gitlab.alpinelinux.org/api/v4/projects/5/packages/generic/v3.0.6/x86_64/apk.static
-        echo 'f1489e05bace7d7dd0a687fcd38d50b585ac660af4231668b123649bef3718c4  /usr/bin/apk' | sha256sum --check
+        APK_TOOLS_API="https://gitlab.alpinelinux.org/api/v4/projects/5"
+        APK_VERSION=$(curl -fsSL --retry 5 --retry-all-errors "${APK_TOOLS_API}/repository/tags?per_page=1" | jq -r '.[].name' | sort -V | tail -1)
+        APK_PKG_ID=$(curl -fsSL --retry 5 --retry-all-errors "${APK_TOOLS_API}/packages?package_name=${APK_VERSION}&package_type=generic" | jq '.[] | select(.version == "x86_64") | .id' | sort | tail -1)
+        APK_SHA256=$(curl -fsSL --retry 5 --retry-all-errors "${APK_TOOLS_API}/packages/${APK_PKG_ID}/package_files" | jq -r '.[0].file_sha256')
+        # If the API fails, fall back to a known-good version
+        if [ -z "$APK_VERSION" ] || [ -z "$APK_SHA256" ]
```
martinpitt: This fallback path can never happen. [shell snippets use `set -e` and `pipefail` by default](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#exit-codes-and-error-action-preference), so failure to download apk from above would just abort the script right there.

action.yaml
```diff
@@ -102,13 +102,29 @@ runs:
...
       # scripts (like .github/workflows/ci.yml) clean up /usr/local
       if: ${{ inputs.apk == 'true' }}
       shell: bash
       run: |
-        sudo curl -fsSL -o /usr/bin/apk https://gitlab.alpinelinux.org/api/v4/projects/5/packages/generic/v3.0.6/x86_64/apk.static
-        echo 'f1489e05bace7d7dd0a687fcd38d50b585ac660af4231668b123649bef3718c4  /usr/bin/apk' | sha256sum --check
+        APK_TOOLS_API="https://gitlab.alpinelinux.org/api/v4/projects/5"
+        APK_VERSION=$(curl -fsSL --retry 5 --retry-all-errors "${APK_TOOLS_API}/repository/tags?per_page=1" | jq -r '.[].name' | sort -V | tail -1)
+        APK_PKG_ID=$(curl -fsSL --retry 5 --retry-all-errors "${APK_TOOLS_API}/packages?package_name=${APK_VERSION}&package_type=generic" | jq '.[] | select(.version == "x86_64") | .id' | sort | tail -1)
+        APK_SHA256=$(curl -fsSL --retry 5 --retry-all-errors "${APK_TOOLS_API}/packages/${APK_PKG_ID}/package_files" | jq -r '.[0].file_sha256')
```
martinpitt: An absent .file_sha256 gets rendered as string `"null"`, failing the `[ -z ]` check below. jq has an `// empty` guard for that.

Also, is it safe to blindly take the first (`[0]`) file? I suppose good enough if they only ever publish one file.

action.yaml
```diff
@@ -102,13 +102,29 @@ runs:
...
+      #
       # NOTE: apk is dropped into /usr/bin and not /usr/local/bin because later CI
       # scripts (like .github/workflows/ci.yml) clean up /usr/local
       if: ${{ inputs.apk == 'true' }}
       shell: bash
       run: |
-        sudo curl -fsSL -o /usr/bin/apk https://gitlab.alpinelinux.org/api/v4/projects/5/packages/generic/v3.0.6/x86_64/apk.static
-        echo 'f1489e05bace7d7dd0a687fcd38d50b585ac660af4231668b123649bef3718c4  /usr/bin/apk' | sha256sum --check
+        APK_TOOLS_API="https://gitlab.alpinelinux.org/api/v4/projects/5"
+        APK_VERSION=$(curl -fsSL --retry 5 --retry-all-errors "${APK_TOOLS_API}/repository/tags?per_page=1" | jq -r '.[].name' | sort -V | tail -1)
```
martinpitt: This gets the *most recent tag*, not the *highest version*. I.e. if "2.0.2" is the current highest version, and someone pushes an "1.0.3" maintenance update, the latter will the be most recent tag.

Also, the `tail -1` should be redundant with `per_page=1`?

action.yaml
```diff
@@ -102,13 +102,29 @@ runs:
...
+        APK_PKG_ID=$(curl -fsSL --retry 5 --retry-all-errors "${APK_TOOLS_API}/packages?package_name=${APK_VERSION}&package_type=generic" | jq '.[] | select(.version == "x86_64") | .id' | sort | tail -1)
```
martinpitt: This sorts IDs by strings, not numeric -- needs `-n`.

martinpitt (changes requested):

action.yaml
behrmann: Good catch, the `per_page=1` actually still needed to go. Getting the latest version is done by `sort -V`. I've also added a filter to get out the rc tags.

action.yaml
behrmann: I hope that should be safe, since it's only a single package ID.

action.yaml
behrmann: Hoisted by my own petard. Having [SC2155](https://www.shellcheck.net/wiki/SC2155) always in my mind I thought the return value would get clobbered. I wrapped it in `set +e/set -e`. (I'm actually no fan of `set -e`, I think it makes error handling in shell in harder than it already is—people don't just panic in Rust everywhere either.)

action.yaml
behrmann: I went for adding another check for `APK_SHA256` being the actual string `null`. Unlikely, but then it'll cover the key not being there and the value being empty.

action.yaml
behrmann: Good catch. Copying is hard.

action.yaml
martinpitt: Heh, I actually think quite the opposite.. Error handling *without* `set -e` is unbearably hard. With the new `set +e`, if the `APK_VERSION` curl call fails, you then do a curl call with ` package_name=` (empty), instead of aborting. I think a safer pattern would be

```sh
if X=$(curl ...) && 
   Y=$(curl ... $X ...) &&
   ...; then
      echo I got it!
else
     echo I am so sad now
```

martinpitt: I.e I'm concerned that `package_name=` might actually be interpreted as "not set"/"no package name filter" instead of "searching for empty package name", and you actually do get something. That's how most web services work?

behrmann: Yes, for `package_name=` you get all the packages and then basically a random `APK_PKG_ID` and therefore sha256sum. This hinges on `APK_VERSION` being empty, though, which will trigger the first check in the fallback check.

action.yaml
```diff
@@ -102,13 +102,30 @@ runs:
       # ubuntu-24.04 doesn't package apk, so fetch the official statically-linked binary from
       # upstream so mkosi can use it to populate the postmarketOS tools tree.
       #
+      # Since GitLab's API can be unreliable, this is retried a few times if there
```
craftyguy, line 105: I think this comment is stale, the retry mechanism was dropped a while ago in some previous rev, this just tries once then falls back

craftyguy: oh derp, it's referring to the curl --retry below

craftyguy:
Thanks @behrmann! Overall I think this looks good. I don't know how long the 'last known good' version will be hosted on gitlab, so that might need to be bumped at some point. The curl calls for gitlab/API stuff are also retried multiple times on failure so this seems a lot more robust at handling gitlab shenanigans

yuwata (approved):

yuwata:
Hmm, alpinelinux.org seems to be so unstable...
https://github.com/systemd/mkosi/actions/runs/32977707415/job/98210714702?pr=4375

bluca (approved):

bluca:
I guess this needs a shorter timeout and a retry loop or something, as the apk infrastructure seems the be really unreliable, even rerunning a couple of times the job still gets stuck

craftyguy:
It might be failing the anubis/bot check thing that's in front of gitlab.alpinelinux.org

bluca (approved):

PR #4376, opened 2026-06-26
Isolate build history per output directory, redux

martinpitt:
The original fixes for parallel tests were in #4360, but one of the key commits broke system and was reverted in #4373. I examined the regression in https://github.com/systemd/mkosi/pull/4373#issuecomment-4807496843 ,  captured the regression (and systemd-desired behaviour) in a unit test, and reworked the fix to not break existing behaviour (again including a unit test).

I already pushed these commits to #4357 to test them with barrage and actual parellelism, and they [work well enough](https://github.com/systemd/mkosi/actions/runs/28230032794/job/83632378425?pr=4357) (i.e. sans the opensuse mirror and `apk` download woes).

I also tested this on systemd: [this commit on my fork](https://github.com/systemd/systemd/commit/990a6957d2f6d05f93b7d8ed2404a3711b8b3974) switches mkosi over to this branch, and [the tests work](https://github.com/martinpitt/systemd/actions/runs/28231100691/job/83634782959) in the sense of "all image builds pass". There are some individual test failures, but they also happen in e.g. https://github.com/systemd/systemd/pull/42755

Details are in the individual commit messages.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -5445,7 +5445,12 @@ def want_default_initrd(config: Config) -> bool:
     return Path("default") in config.initrds
 
 
-def finalize_historydir(args: Args) -> Path:
+def finalize_historydir(args: Args, output_dir: Optional[Path] = None) -> Path:
+    # When an output directory is given, the build history is also stored there so that builds into
+    # distinct output directories don't read each other's history. Otherwise it lives in the config dir.
+    if output_dir is not None:
+        return output_dir / ".mkosi-private/history"
+
     configdir = finalize_configdir(args.directory)
     return (configdir or Path.cwd()) / ".mkosi-private/history"
 
@@ -5502,7 +5507,13 @@ def parse_config(
         return args, None, ()
 
     configdir = finalize_configdir(args.directory)
-    historydir = finalize_historydir(args)
+    config_historydir = finalize_historydir(args)
+    # Prefer the history stored in the output directory (when --output-directory is given on the CLI) so a
+    # consumer reads back exactly the build that wrote into that directory. Fall back to the config dir for
+    # consumers that don't pass --output-directory and for history written by older mkosi versions.
+    historydir = finalize_historydir(args, context.cli.get("output_dir"))
+    if not (historydir / "latest.json").exists():
+        historydir = config_historydir
 
     if have_history(args, historydir):
         history = Config.from_partial_json((historydir / "latest.json").read_text())
@@ -5555,8 +5566,15 @@ def parse_config(
     maincontext = copy.deepcopy(context)
 
     if config["history"] and want_new_history(args):
-        historydir.mkdir(parents=True, exist_ok=True)
-        (historydir / "latest.json").write_text(dump_json(Config.to_partial_dict(cli)))
+        # Store the history in the config dir (so consumers that don't pass --output-directory find it) and,
+        # when an output directory is configured, in the output directory too (so builds into distinct
+        # output directories stay isolated). These coincide when no output directory is set. This keys on
+        # the finalized output dir (config or CLI), unlike the read above which can only use the CLI value
+        # (the configuration isn't parsed yet there), so don't collapse the two into one variable.
+        latest_json = dump_json(Config.to_partial_dict(cli))
+        for hd in [config_historydir, finalize_historydir(args, config.get("output_dir"))]:
+            hd.mkdir(parents=True, exist_ok=True)
+            (hd / "latest.json").write_text(latest_json)
 
     tools = None
     if config.get("tools_tree") in (Path("default"), Path("yes")):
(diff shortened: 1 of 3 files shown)


martinpitt:
@behrmann @daandemeyer friendly review ping? Thanks!

tests/test_config.py
```diff
@@ -1992,3 +1992,64 @@ def test_history_empty_list(tmp_path: Path) -> None:
...
+        OutputDirectory={out}
+
+        [Build]
+        History=yes
+        """
+    )
+
+    with chdir(d):
+        # The build passes -O explicitly, pointing at the same directory the config already configures.
+        parse_config(["--output-directory", str(out), "--image-id", "img-x", "build"])
```
behrmann: Nit: (below as well)
```suggestion
        parse_config(["--output-directory", os.fspath(out), "--image-id", "img-x", "build"])
```

tests/test_config.py
```diff
@@ -1992,3 +1992,64 @@ def test_history_empty_list(tmp_path: Path) -> None:
...
+    # by ImageId so we can tell which build a later consumer reads back.
+    with chdir(d):
+        parse_config(["--output-directory", str(out_a), "--image-id", "img-a", "build"])
+        parse_config(["--output-directory", str(out_b), "--image-id", "img-b", "build"])
+
+        # A verb that consumes a previous build, pointed at output directory A, must read back the
+        # configuration of the build that wrote into A -- not whichever build happened to run last.
+        _, _, [config] = parse_config(["--output-directory", str(out_a), "summary"])
+
+    assert config.image_id == "img-a"
```
behrmann: I think it would be nice to test the full range here
```suggestion
        # A verb that consumes a previous build, pointed at output directory A, must read back the
        # configuration of the build that wrote into A -- not whichever build happened to run last.
        _, _, [config_last] = parse_config(["summary"])
        _, _, [config_a] = parse_config(["--output-directory", os.fspath(out_a), "summary"])
        _, _, [config_b] = parse_config(["--output-directory", os.fspath(out_b), "summary"])
        
    assert config_last.image_id == "img-b"
    assert config_a.image_id == "img-a"
    assert config_b.image_id == "img-b"        
```

mkosi/config.py
```diff
@@ -5555,8 +5566,14 @@ def parse_config(
...
 
     if config["history"] and want_new_history(args):
-        historydir.mkdir(parents=True, exist_ok=True)
-        (historydir / "latest.json").write_text(dump_json(Config.to_partial_dict(cli)))
+        # Store the history in the config dir (so consumers that don't pass --output-directory find it) and,
+        # when an output directory is configured, in the output directory too (so builds into distinct
+        # output directories stay isolated). These coincide when no output directory is set. This keys on
+        # the finalized output dir (config or CLI), unlike the read above which can only use the CLI value
+        # (the configuration isn't parsed yet there), so don't collapse the two into one variable.
+        for hd in {config_historydir, finalize_historydir(args, config.get("output_dir"))}:
```
behrmann: I don't trust iteraton over sets. While dicts preserve insertion order nowadays, I don't think this is guaranteed for sets.

Since the actions taken are idempotent (the mkdir would mkdir the same directory and skip it subsequently as a noop and the config does not change between writes), you can just do it twice
```suggestion
        for hd in (config_historydir, finalize_historydir(args, config.get("output_dir"))):
```

Please also pull the `dump_json` out of the loop, since we only need to create the string once.

behrmann (commented):
Thanks! Sorry, had absolutely forgotten about having to look at this.

tests/test_config.py
martinpitt: Argh, sorry. Fixed! (The other instances as well).

tests/test_config.py
martinpitt: Good idea! Done.

mkosi/config.py
martinpitt: Uh, what's wrong about set iteration? It would avoid writing the same file twice. But oh well, that's cheap and it's not important to argue about, so I just turned it into a list.

I pulled out the dump_json, thanks for pointing out.

behrmann: > Uh, what's wrong about set iteration? 

The order isn't guaranteed. I just got
```
>>> for i in {'a', 1, 'b', 2}:
>>>     print(i)
1
2
b
a
```
on my machine. 

This isn't really an issue here, but I'm always wary of things that are not deterministic, since this way the order of writes is always the same between different setups. (This is most likely deterministic, it's always the same order when I test this and I'm pretty sure it's in the order of the hash key.)

behrmann:
In the light of #4381 can you drop the `Co-Authored-By:` tag on the commit? An `Assisted-By: LLM` (yes, entirely generic), is fine.

behrmann (approved):

mkosi/config.py
martinpitt: Ah yes. I don't care about the order here, just wanted to avoid writing the same file twice. But anyway, good 'nuff!

bluca (approved):

PR #4377, opened 2026-06-30
Add support for signing PCR policies for initializing NvPCRs

chrisccoulson:
The way that NvPCRs are initialized and anchored in systemd is changing,
and to support this, the UKI needs to include signed PCR policies that
can authorize NvPCR initialization in early boot. This is enabled with a
new `--sign-nvpcr-init` option for ukify.

This adds a new "SignNvPCRInit=" option to control this. The default is
"auto" which will turn on this option if PCR signing is enabled and
ukify is new enough.

The corresponding systemd PR is https://github.com/systemd/systemd/pull/42796

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1787,6 +1787,17 @@ def build_uki(
         else:
             cert_parameter = "--pcr-public-key"
 
+        if context.config.sign_nvpcr_init == ConfigFeature.enabled or (
+            context.config.sign_nvpcr_init == ConfigFeature.auto
+            and systemd_tool_version(
+                python_binary(context.config),
+                ukify,
+                sandbox=context.sandbox,
+            )
+            >= "262~devel"
+        ):
+            arguments += ["--sign-nvpcr-init"]
+
         # If we're providing the private key via an engine or provider, we have to pass in a X.509
         # certificate via --pcr-certificate as well.
         if context.config.sign_expected_pcr_key_source.type != KeySourceType.file:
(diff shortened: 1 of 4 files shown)


mkosi/config.py
```diff
@@ -3769,6 +3770,15 @@ def parse_kernel_module_filter_regexp(p: str) -> str:
...
         scope=SettingScope.inherit,
     ),
+    ConfigSetting(
+        dest="sign_nvpcr_init",
+        metavar="FEATURE",
+        section="Validation",
+        name="SignNvPCRInit",
+        parse=config_parse_feature,
+        help="Generate a signed PCR policy specifically for authorizing the initialization of "
+        "NvPCRs and embed this into the UKI",
```
behrmann, line 3780: Please don't break that, but add a `noqa` comment

mkosi/__init__.py
```diff
@@ -1787,6 +1787,17 @@ def build_uki(
...
             cert_parameter = "--pcr-public-key"
 
+        if context.config.sign_nvpcr_init == ConfigFeature.enabled or (
+            context.config.sign_nvpcr_init == ConfigFeature.auto
+            and systemd_tool_version(
+                python_binary(context.config),
+                ukify,
+                sandbox=context.sandbox,
+            )
+            >= "262~devel"
```
behrmann, line 1797: This particular formatting has always disgusted me. Fortunately, We already need to test the ukify version a few lines above, so if you add 
```python
ukify_version = systemd_tool_version(python_binary(context.config), ukify, sandbox=context.sandbox)
```
before line 1778, you can make it
```suggestion
            and ukify_version >= "262~devel"
```
here and there (with the appripriate version there, of course).

behrmann (commented):
I'm setting the do not merge label, while the PR in systemd is in flight.

chrisccoulson:
I've re-opened this as a new PR (https://github.com/systemd/mkosi/pull/4404) as the systemd PR changed a bit and it's a different option now. It does address the comments in this PR.

PR #4379, opened 2026-07-01
sandbox: Split a chroot() context manager out of chase()

daandemeyer:
Extract the chroot/restore dance from chase() into a reusable chroot()
context manager so it can be used on its own.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/mkosi/sandbox.py
+++ b/mkosi/sandbox.py
@@ -935,6 +935,23 @@ def userns_acquire_empty() -> int:
     return userns_fd
 
 
+class chroot:
+    def __init__(self, root: str) -> None:
+        self.root = root
+
+    def __enter__(self) -> None:
+        self.cwd = os.getcwd()
+        self.fd = os.open("/", os.O_CLOEXEC | os.O_PATH | os.O_DIRECTORY)
+        os.chroot(self.root)
+        os.chdir("/")
+
+    def __exit__(self, *args: object, **kwargs: object) -> None:
+        os.fchdir(self.fd)
+        os.close(self.fd)
+        os.chroot(".")
+        os.chdir(self.cwd)
+
+
 def chase(root: str, path: str, *, nofollow: bool = False) -> str:
     # pyright hack around `reportPossiblyUnboundVariable`; it doesn't understand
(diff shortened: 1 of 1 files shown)


mkosi/sandbox.py
```diff
@@ -935,6 +935,27 @@ def userns_acquire_empty() -> int:
     return userns_fd
 
 
+class chroot:
```
behrmann, line 938: It being a class it should have an uppercase name by convention.
```suggestion
class Chroot:
```

mkosi/sandbox.py
```diff
@@ -935,6 +935,27 @@ def userns_acquire_empty() -> int:
...
+
+    def __enter__(self) -> None:
+        self.cwd = os.getcwd()
+        self.fd = os.open("/", os.O_CLOEXEC | os.O_PATH | os.O_DIRECTORY)
+        try:
+            os.chroot(self.root)
+            os.chdir("/")
+        except BaseException:
+            os.close(self.fd)
+            raise
```
behrmann: I think you can shorten this to
```suggestion
        os.chroot(self.root)
        os.chdir("/")
```
since context manager types unroll this
```python
with EXPR as VAR:
    BLOCK
```
to (something similar to) this
```python
mgr = (EXPR)
exit = type(mgr).__exit__  # Not calling it yet
value = type(mgr).__enter__(mgr)
exc = True
try:
    try:
        VAR = value  # Only if "as VAR" is present
        BLOCK
    except:
        # The exceptional case is handled here
        exc = False
        if not exit(mgr, *sys.exc_info()):
            raise
        # The exception is swallowed if exit() returns true
finally:
    # The normal and non-local-goto cases are handled here
    if exc:
        exit(mgr, None, None, None)
```
meaning `__exit__` will be called as a `finally` anyway.

Also: TIL context manager types can suppress exceptions by returning `True` from `__exit__`.

behrmann (approved):

mkosi/sandbox.py
daandemeyer: We do the context managers lowercase, the close stuff is the same

PR #4381, opened 2026-07-01
Forbid AI attribution in commit messages

daandemeyer:
Adopt systemd's policy: only human beings can be credited within commit
messages. Drop the previous AI Contribution Disclosure section that
required a Co-Authored-By trailer.


behrmann:
I wouldn't say it forbids AI attribution, it forbids the legally dubious `Co-{Authored,Developed}-By:` and it drops *the requirement* for attribution. Personally I think `Assisted-By:` is a good idea, but that's for elsewhere to discuss, this just aligns us with systemd proper.

PR #4386, opened 2026-07-03
Fix tar exclude patterns for apivfs paths

daradib:
Add `--anchored` and non-prefixed variants of exclude patterns so apivfs directories are properly excluded regardless of whether tar entries are prefixed with `./`.

To reproduce the issue, try using BaseTrees with a tarball that doesn't have the `./` prefix, for example, from [Ubuntu Cloud Images](https://cloud-images.ubuntu.com/):
```
[Content]
BaseTrees=cloud-images/ubuntu-26.04-server-cloudimg-amd64v3-root.tar.xz
```

It will not exclude /dev/* and fail to build.

### DIFF
--- a/mkosi/archive.py
+++ b/mkosi/archive.py
@@ -13,12 +13,13 @@ from mkosi.util import PathString, chdir
 
 def tar_exclude_apivfs_tmp() -> list[str]:
     return [
-        "--exclude", "./dev/*",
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Thank you!

behrmann:
Neat, TIL `--anchored`. Thanks!

PR #4388, opened 2026-07-03
debian: add new split-out systemd packages

bluca:
Tools and services related to tpm, imds and report were split out in new subpackages to reduce the default dependencies. Adjust the package lists accordingly.


behrmann (approved):

bluca:
Looks like we need to wait for the new version to migrate from stonking-proposed to stonking

behrmann (approved):

PR #4390, opened 2026-07-03
fix(ubuntu-kali-debian, tools-trees): add `passwd` into debian/ubuntu tools trees

tulilirockz:
This fixes an issue where `/usr/lib/rpm/sysusers.sh` would fail to run on a debian/ubuntu system due to `usermod`, `groupadd` and `useradd` not being present in the tools tree. RPM seems to run this specific script on the "host" (tools tree) and this missing makes it so fedora-based image builds that include any package with sysupdate scripts fails to build.

## Reproduction

```bash
tee mkosi.conf <<EOF
[Config]
MinimumVersion=27~devel

[Distribution]
Distribution=fedora
Release=44

[Build]
ToolsTree=default
ToolsTreeDistribution=debian

[Content]
Packages=dbus-common # this package includes sysupdate scripts as of f44

[Output]
OutputDirectory=mkosi.output
EOF
mkosi build
```

This should output normal build output for any fedora image, then it should fail with the following RPM scriptlet errors:

```bash
>>> Running sysusers scriptlet: setup-0:2.15.0-28.fc44.noarch
>>> Finished sysusers scriptlet: setup-0:2.15.0-28.fc44.noarch
>>> Scriptlet output:
>>> /usr/lib/rpm/sysusers.sh: line 107: groupadd: command not found
>>> /usr/lib/rpm/sysusers.sh: line 107: groupadd: command not found
>>> /usr/lib/rpm/sysusers.sh: line 107: groupadd: command not found
>>> /usr/lib/rpm/sysusers.sh: line 107: groupadd: command not found
[7/9] Installing fedora-release-0:44-18.noarch                                  100% |   1.0 KiB/s | 124.0   B |  00m00s
>>> Running sysusers scriptlet: setup-0:2.15.0-28.fc44.noarch
>>> Finished sysusers scriptlet: setup-0:2.15.0-28.fc44.noarch
>>> Scriptlet output:
>>> /usr/lib/rpm/sysusers.sh: line 107: groupadd: command not found
>>> /usr/lib/rpm/sysusers.sh: line 107: groupadd: command not found
>>> /usr/lib/rpm/sysusers.sh: line 107: groupadd: command not found
>>> /usr/lib/rpm/sysusers.sh: line 107: groupadd: command not found
>>> /usr/lib/rpm/sysusers.sh: line 107: groupadd: command not found
>>> /usr/lib/rpm/sysusers.sh: line 107: groupadd: command not found
>>> /usr/lib/rpm/sysusers.sh: line 107: groupadd: command not found
>>> /usr/lib/rpm/sysusers.sh: line 107: groupadd: command not found
>>> /usr/lib/rpm/sysusers.sh: line 107: groupadd: command not found
>>> /usr/lib/rpm/sysusers.sh: line 107: groupadd: command not found
>>> /usr/lib/rpm/sysusers.sh: line 107: groupadd: command not found
>>> /usr/lib/rpm/sysusers.sh: line 107: groupadd: command not found
>>> /usr/lib/rpm/sysusers.sh: line 107: groupadd: command not found
>>> /usr/lib/rpm/sysusers.sh: line 107: groupadd: command not found
# ...
```

## Real world 

- https://github.com/zirconium-dev/zirconium/actions/runs/28683154628
- https://github.com/zirconium-dev/zirconium/actions/runs/28459381411
- https://github.com/zirconium-dev/zirconium/actions/runs/28683154628

These image builds have been failing on the ubuntu runners because they depend on a NetworkManager package that fails if the sysupdate scripts do not work.

### DIFF
--- a/mkosi/resources/mkosi-tools/mkosi.profiles/misc/mkosi.conf.d/debian-kali-ubuntu.conf
+++ b/mkosi/resources/mkosi-tools/mkosi.profiles/misc/mkosi.conf.d/debian-kali-ubuntu.conf
@@ -7,6 +7,7 @@ Distribution=|ubuntu
 
 [Content]
 Packages=
+        passwd
         git-core
         libnss-resolve
         libnss-myhostname


tulilirockz:
CI seems to be failing due to unrelated issues

behrmann:
Sorry, somehow I had overlooked this PR. This should probably have been handled in #4195 already. 

@bluca You had opinions on `useradd` back then. I think adding it to the tools tree should be okay.

behrmann (approved):

bluca:
Doesn't rpm just use sysusers these days? @keszybz ?

bluca:
Gah fat fingered the update button on the phone

tulilirockz:
> Doesn't rpm just use sysusers these days?

It does, but in order to do that it has a `/usr/lib/rpm/sysusers.sh` script that gets called at `rpm -i` (postinstall script) time, that does that for Whatever Reason:tm:

tulilirockz:
> This should probably have been handled in https://github.com/systemd/mkosi/pull/4195 already.

Oh awesome! Feel free to close if this is already handled or something. Will try out with that other PR as well

bluca:
> > Doesn't rpm just use sysusers these days?
> 
> It does, but in order to do that it has a `/usr/lib/rpm/sysusers.sh` script that gets called at `rpm -i` (postinstall script) time, that does that for Whatever Reason:tm:

please try and figure that out, because suid packages were removed intentionally

tulilirockz:
I mean, I dont think theres any way to actually do this without `groupadd` and others? You can check from the reproduction that the build just straight up fails to complete because of that. I'm not sure exactly what we could do here to be completely honest

bluca:
Again, I'm pretty sure passwd should not be necessary anymore, as sysusers is used instead for fedora packages

keszybz:
> > Doesn't rpm just use sysusers these days?
> 
> It does, but in order to do that it has a `/usr/lib/rpm/sysusers.sh` script that gets called at `rpm -i` (postinstall script) time, that does that for Whatever Reason:tm:

In Fedora, we have a patch on rpm to use `systemd-sysusers` instead of that script (https://src.fedoraproject.org/rpms/rpm/blob/rawhide/f/rpm-4.20-sysusers.patch). It seems that the rpm upstream default is still to use the script. Normally, I'd say that we should support the unpatched version, but in this case, upstream is using this workaround script, and using the script is problematic. It may not implement all sysusers features (e.g. `r` support), and as discussed above, it pulls into those external setuid tools. And we really want to dogfood the systemd stuff in our config. I think that the RH people working on this applied the downstream patch and mostly forgot about this in the upstream.

In my `trixie` system, `/usr/lib/rpm/macros` has `%__systemd_sysusers %{_rpmconfigdir}/sysusers.sh` :(

So please ask Debian to patch this. The upstream default makes little sense on systemd systems.

bluca:
I don't really want to patch it downstream, why can't it be just fixed upstream?

bluca:
> I don't really want to patch it downstream, why can't it be just fixed upstream?

---> https://github.com/rpm-software-management/rpm/pull/4280

PR #4392, opened 2026-07-05
kmod: eliminate duplication in calls to modinfo

BroadlyWhitaker:
The original guard when building the new todo is:

```
todo += [m for m in depinfo.modules if m not in mods and m in nametofile]
```

`mods` only accumulates modules from previous iterations of the outer while loop. Within a single iteration, modules in the current batch are added to mods one at a time as `moddep.items()` is iterated. This means two distinct failure modes:

1. Cross-dependency within a batch: if modules A and B are both in the current todo and both depend on X, X passes the m not in mods check when processing A's deps and again when processing B's (since X hasn't been added to mods yet). X ends up in todo twice.

2. Intra-batch back-edge: if A depends on B and both are already in the current todo/moddep, B still passes m not in mods when processing A's dep list, queuing B for a redundant second modinfo call in the next iteration.

Both boil down to the same root cause: the check doesn't account for modules currently being processed.

The fix addresses both issues:

1. `m not in moddep.keys()` prevents modules from the current batch being re-queued into the next iteration's `todo`.

2. Turning`todo` into a set prevents a module from appearing *multiple times* in the same next `todo`, when two modules in the current batch share a dependency

Because these duplicates frequently occur in practice, eliminating them provides an incremental speedup on top of the original optimization made in GH4092/GH4017.

### DIFF
--- a/mkosi/kmod.py
+++ b/mkosi/kmod.py
@@ -292,8 +292,8 @@ def resolve_module_dependencies(
     with chdir(context.root):
         nametofile = {module_path_to_name(m): m for m in all_modules(modulesd)}
 
-    todo = [*builtin, *modules]
-    mods = set()
+    todo = {*builtin, *modules}
+    mods: set[str] = set()
     firmware = set()
 
     while todo:
@@ -304,20 +304,20 @@ def resolve_module_dependencies(
         # build a map that maps the module name to both its module dependencies and its firmware
         # dependencies. Because there's more kernel modules than the max number of accepted CLI
         # arguments, we split the modules list up into chunks if needed.
-        for i in range(0, len(todo), 8500):
-            chunk = todo[i : i + 8500]
+        ittodo = iter(todo)
+        while chunk := tuple(itertools.islice(ittodo, 8500)):
             moddep |= modinfo(context, kver, chunk)
 
-        todo = []
+        todo = set()
+        mods |= moddep.keys()
 
         for name, depinfo in moddep.items():
             for d in depinfo.modules:
                 if d not in nametofile and d not in builtin:
                     logging.warning(f"{d} is a dependency of {name} but is not installed, ignoring ")
 
-            mods.add(name)
             firmware.update(depinfo.firmware)
-            todo += [m for m in depinfo.modules if m not in mods and m in nametofile]
+            todo.update(m for m in depinfo.modules if m not in mods and m in nametofile)
 
     return set(nametofile[m] for m in mods if m in nametofile), set(firmware)
 


behrmann:
Thanks for the contribution! I will only have time to look at this later this week, but I started the CI run to see whether it works.

mkosi/kmod.py
```diff
@@ -292,23 +292,15 @@ def resolve_module_dependencies(
     with chdir(context.root):
         nametofile = {module_path_to_name(m): m for m in all_modules(modulesd)}
 
-    todo = [*builtin, *modules]
+    todo = set(builtin) | set(modules)
```
behrmann: Let's use a set literal here.
```suggestion
    todo = {*builtin, *modules}
```

mkosi/kmod.py
```diff
@@ -292,23 +292,15 @@ def resolve_module_dependencies(
...
     mods = set()
     firmware = set()
 
     while todo:
         moddep: dict[str, ModuleDependencyInfo] = {}
+        moddep |= modinfo(context, kver, tuple(todo))
```
behrmann: How are you preventing here, that the number of CLI arguments grows out of bounds, which is what the code currently does?

You could lift the tuplification of `todo` out and keep the loop, no?
```suggestion
        tochunk = tuple(todo)
        # We could run modinfo once for each module but that's slow. Luckily we can pass multiple modules
        # to modinfo and it'll process them all in a single go. We get the modinfo for all modules to
        # build a map that maps the module name to both its module dependencies and its firmware
        # dependencies. Because there's more kernel modules than the max number of accepted CLI
        # arguments, we split the modules list up into chunks if needed.
        for i in range(0, len(todo), 8500):
            chunk = tochunk[i : i + 8500]
            moddep |= modinfo(context, kver, chunk)    
```

behrmann (commented):
Thank you! Sorry for the lag on my end.

behrmann:
I've sat down an ran a tiny bit of benchmarking on this and running this against my host system, with empty inputs as
```python
required_kernel_modules(kver, modules_include=[], modules_exclude=[], firmware_include=[], firmware_exclude=[], resolve_func=resolve_module_dependencies)
```
after making the function to resolve module dependencies pluggable to easily compare this, I get roughly the same run time for both functions, with the version from this PR actually a tiny bit, but consistently, slower.

At first I thought it was the chunking and I changed that to
```
        ittodo = iter(todo)
        while chunk := tuple(itertools.islice(ittodo, 8500)):
            moddep |= modinfo(kver, chunk)
```
which is roughly `itertools.batch`, which we don't yet have because it was only added in 3.12.

That helps a bit, but doesn't get one all the way, but the real trick is in lifting the todo update
```
        todo = set()
        mods |= moddep.keys()

        for name, depinfo in moddep.items():
            for d in depinfo.modules:
                if d not in nametofile and d not in builtin:
                    logging.warning(f"{d} is a dependency of {name} but is not installed, ignoring ")

            firmware.update(depinfo.firmware)
            todo.update(m for m in depinfo.modules if m not in mods and m in nametofile)
```
Since `mods` just gets the `name`s from `moddep` added we can just add all keys from `moddep` to `mods`  in one go. This then also makes the `mods | moddep.keys()` unnecessary. This shaves of 30% of the runtime off of `required_kernel_modules` for me.

Could you check that this makes sense and works in your usecase?

BroadlyWhitaker:
Hi and thanks for looking at my PR.

How much benefit this does in practice depends both on the specifics of your `KernelModules` and on how fast your SSD is and It's difficult to give a meaningful general number. It's most impact comes when you specify a subset of modules, that have many common deps, and modinfo is called multiple times to resolve children-of-children. This isn't a common case, and the benefits are generally modest. 

With your change I see smaller 0-10% faster here, and your version is also cleaner so I'm happy to include your suggestion. I also added a Co-Authored-By.

mkosi/kmod.py
```diff
@@ -292,32 +292,32 @@ def resolve_module_dependencies(
     with chdir(context.root):
         nametofile = {module_path_to_name(m): m for m in all_modules(modulesd)}
 
-    todo = [*builtin, *modules]
-    mods = set()
+    todo = {*builtin, *modules}
+    mods : set[str] = set()
```
behrmann: Is this annotation necessary? I don't see mypy (or pyright or ty) complaning about this. My mypy version is 2.1. We do avoid unneeded annotations.

If it were necessary, there would be this small nit


```suggestion
    mods: set[str] = set()
```

mkosi/kmod.py
```diff
@@ -292,32 +292,32 @@ def resolve_module_dependencies(
...
+        tochunk = tuple(todo)
         # We could run modinfo once for each module but that's slow. Luckily we can pass multiple modules
         # to modinfo and it'll process them all in a single go. We get the modinfo for all modules to
         # build a map that maps the module name to both its module dependencies and its firmware
         # dependencies. Because there's more kernel modules than the max number of accepted CLI
         # arguments, we split the modules list up into chunks if needed.
         for i in range(0, len(todo), 8500):
-            chunk = todo[i : i + 8500]
+            chunk = tochunk[i : i + 8500]
             moddep |= modinfo(context, kver, chunk)
```
behrmann: Could you also incorporate this?
```python
        # We could run modinfo once for each module but that's slow. Luckily we can pass multiple modules
        # to modinfo and it'll process them all in a single go. We get the modinfo for all modules to
        # build a map that maps the module name to both its module dependencies and its firmware
        # dependencies. Because there's more kernel modules than the max number of accepted CLI
        # arguments, we split the modules list up into chunks if needed.
        # TODO: use itertools.batched starting with 3.12
        ittodo = iter(todo)
        while chunk := tuple(itertools.islice(ittodo, 8500)):
            moddep |= modinfo(kver, chunk)
```

mkosi/kmod.py
BroadlyWhitaker: I removed the space but yes I get mypy error.

```
mypy --version
mypy 2.1.0 (compiled: yes)

mypy mkosi
pyproject.toml: [mypy]: python_version: Python 3.9 is not supported (must be 3.10 or higher). You may need to put quotes around your Python version
mkosi/kmod.py:296:5: error: Need type annotation for "mods" (hint: "mods: set[<type>] = ...")  [var-annotated]
        mods = set()
        ^~~~
Found 1 error in 1 file (checked 52 source files)
```

mkosi/kmod.py
BroadlyWhitaker: you probably meant 
```
moddep |= modinfo(context, kver, chunk)
```

mkosi/kmod.py
```diff
@@ -292,32 +292,28 @@ def resolve_module_dependencies(
...
-        # We could run modinfo once for each module but that's slow. Luckily we can pass multiple modules
-        # to modinfo and it'll process them all in a single go. We get the modinfo for all modules to
-        # build a map that maps the module name to both its module dependencies and its firmware
-        # dependencies. Because there's more kernel modules than the max number of accepted CLI
-        # arguments, we split the modules list up into chunks if needed.
-        for i in range(0, len(todo), 8500):
-            chunk = todo[i : i + 8500]
+        # modinfo uses todo values as command line arguments, so we process in batches.
+        ittodo = iter(todo)
+        while chunk := tuple(itertools.islice(ittodo, 8500)):
```
martinpitt: My reading of [islice()](https://docs.python.org/3/library/itertools.html#itertools.islice) is different. Like this, it would skip the first 8500 items that way. Don't you mean setting the `stop` argument? `slice(ittodo, None, 8500)`?

martinpitt (commented):
Thanks! I am like 80% sure the islice() is wrong. There are also a fair amount of test failures, I didn't go through them (slooow internet in the train), but they might be related

behrmann:
I think the failures are something different, because, the opensuse/centos pairing notwithstanding, they're all Debian. I've only quickly looked at debian/debian, but there it's
```
2026-07-18T13:12:58.3668431Z ‣ losetup not found.
2026-07-18T13:12:58.4373304Z FAILED
```
which is a message from `run` not finding `losetup` during `test_initrd_lvm` and `test_initrd_lvm_luks`. That seems unrelated to me.

mkosi/kmod.py
behrmann: I think that should be okay, the code is [adapted from the itertools docs](https://docs.python.org/3/library/itertools.html#itertools.batched). Without a third argument the second argument is `stop`:
```python
>>> todo = set(range(7))
>>> ittodo = iter(todo)
>>> while chunk := tuple(itertools.islice(ittodo, 3)):
...     print(chunk)
(0, 1, 2)
(3, 4, 5)
(6,)
```
is as it should be, I think.

martinpitt: Got it. Yay (not) for magic argument parsing..

martinpitt (approved):
Thanks! I cannot resolve my own threads, but all good now.

mkosi/kmod.py
```diff
@@ -292,32 +292,28 @@ def resolve_module_dependencies(
...
         moddep: dict[str, ModuleDependencyInfo] = {}
 
-        # We could run modinfo once for each module but that's slow. Luckily we can pass multiple modules
-        # to modinfo and it'll process them all in a single go. We get the modinfo for all modules to
-        # build a map that maps the module name to both its module dependencies and its firmware
-        # dependencies. Because there's more kernel modules than the max number of accepted CLI
-        # arguments, we split the modules list up into chunks if needed.
-        for i in range(0, len(todo), 8500):
-            chunk = todo[i : i + 8500]
+        # modinfo uses todo values as command line arguments, so we process in batches.
```
behrmann: Since you have to rebase anyway: It would be nice if you could add the old comment back. It makes the diff smaller and it's still topical to what's happening.

behrmann (approved):
One last thing: Could you rebase against main so that we can run CI again? There was a change in Debian that broke CI (unrelated to your change).

mkosi/kmod.py
BroadlyWhitaker: IMHO the comment is far too verbose and unneeded.

behrmann:
Thanks @BroadlyWhitaker!

PR #4398, opened 2026-07-12
util: don't drop single-character variable names in read_env_file

prownd:
The regexp used to parse environment/os-release files required the variable name to consist of at least two characters ([a-zA-Z_][a-zA-Z_0-9]+), so a line with a single-character name such as "A=1" or "_=3" did not match and was silently discarded as a bad line.

Single-character names are valid both in os-release files and in the shell, so allow them by using '*' instead of '+' for the trailing character class.

Add a regression test covering single-character names.

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -4,7 +4,7 @@ from pathlib import Path
 
 import pytest
 
-from mkosi.util import parents_below
+from mkosi.util import parents_below, read_env_file
 
 
 def test_parents_below_basic() -> None:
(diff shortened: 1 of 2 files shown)


tests/test_util.py
```diff
@@ -50,3 +50,11 @@ def test_parents_below_below_is_child_raises() -> None:
     below = Path("/a/b/c")
     with pytest.raises(ValueError):
         parents_below(path, below)
+
+
+def test_read_env_file_single_char_name(tmp_path: Path) -> None:
```
martinpitt: Would you mind generalizing this to `test_read_env_file`, and adding a value with multiple characters, too?

martinpitt (commented):
Thank you! That function isn't tested at all right now, can we do one more round?

tests/test_util.py
martinpitt: Ah sorry, I mis-looked: this is about the *key* names, and there's already a `BC` test. So please only rename to `test_read_env_file`.

behrmann (approved):

tests/test_util.py
martinpitt: Cheers!

martinpitt (approved):
Thank you!

tests/test_util.py
prownd: Okay, I'll rename and modify it according to your suggestion.

PR #4399, opened 2026-07-12
distribution: report the actual architecture in unsupported-arch errors

prownd:
When an architecture is not supported by a distribution, the error message interpolated 'a', which is the result of dict.get(arch) and is therefore always None on this code path (we only get here when the lookup missed). As a result the message always read "Architecture None is not supported by ..." instead of naming the architecture the user actually requested.

Interpolate 'arch' instead, matching what debian.py and kali.py already do. Affects arch, azure, centos, fedora, mageia, openmandriva, opensuse and postmarketos.


behrmann (approved):
Geez, what a silly mistake. Thank you!

PR #4404, opened 2026-07-28
Add support for signing PCR policies for the initrd

chrisccoulson:
The way that NvPCRs are initialized and anchored in systemd is changing,
and to support this, the UKI needs to include signed PCR policies that
can only be satisfied during the initrd, as these are used to authorize
NvPCR initialization. This is enabled with a new `--sign-initrd-pcrs`
option for ukify.

This adds a new "SignInitrdPCRs=" option to control this. The default is
"auto" which will turn on this option if PCR signing is enabled and
ukify is new enough.

The corresponding systemd PR is https://github.com/systemd/systemd/pull/42796

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1775,17 +1775,17 @@ def build_uki(
             "--pcr-banks", "sha256",
         ]  # fmt: skip
 
-        if (
-            systemd_tool_version(
-                python_binary(context.config),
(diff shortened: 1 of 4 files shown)


behrmann:
I'll be looking at this next week, since I'm a bit swamped this one.

behrmann (approved):

PR #4419, opened 2026-08-15
disable grub trigger for postmarketOS CI

behrmann:


yuwata (approved):

behrmann:
@yuwata I had not really looked at #4422 before as I was travelling. I think that approach makes more sense, so that not everybody has to reinvent the wheel, since everybody would need to drop this into their sandbox tree otherwise. Unfortunately I couldn't reopen the PR, so I pushed this on top.

yuwata:
Hmm, personally I prefer your previous implementation. If your previous version does not work, then maybe better to create the .conf file in prepare script?

The reason I do not like #4422 is that the .conf file is created and removed every time. At least, I think it is better to keep the .conf file.

But, not a strong opinion. If it is approved by other maintainers, then that's fine.

behrmann:
> The reason I do not like https://github.com/systemd/mkosi/pull/4422 is that the .conf file is created and removed every time. At least, I think it is better to keep the .conf file.

I think that's fine. It's basically lifted straight from the code in `mkosi/installer/apt.py` that does the same for `/usr/sbin/policy-rc.d`. I thought about changing the commit and dropping the removal of the file, but thought it better to keep the symmetry.

yuwata:
I see. OK.

bluca (approved):

PR #4421, opened 2026-08-16
Add ToolsTreeSnapshot= for pinning the default tools tree

### LINKED ISSUE #4420: Snapshot= silently does not apply to the default tools tree (no ToolsTreeSnapshot=)

main (75fdb9b; also reproduces on v26)

### Used host distribution

Ubuntu 24.04 (GitHub Actions runner)

### Used target distribution

Ubuntu resolute

Setting `Snapshot=` in `[Distribution]` pins the image build but not the default tools tree. The tools tree is still built from live mirrors, and nothing warns about it.

As far as I can tell from config.py, `Snapshot` is declared `scope=SettingScope.universal` with `tools=True`. Universal propagates to subimages but not to the tools tree, and `tools=True` only permits the setting inside a tools tree config, it doesn't copy the main config's value over. The ToolsTree* settings that do cross the boundary (ToolsTreeMirror, ToolsTreeRelease, ...) have no ToolsTreeSnapshot counterpart.

How we hit it: we had `ToolsTreeMirror=https://snapshot.ubuntu.com/ubuntu/<ts>` and replaced it with `Snapshot=`, since Snapshot also covers the -security pocket, which a mirror URL cannot (ubuntu.py hardcodes security.ubuntu.com). That made things worse instead of better: the tools tree lost its only pin and quietly went back to archive.ubuntu.com, current-day packages. We build measured images, so tools tree drift changes the output while every config looks pinned.

I'd be fine with any of: a ToolsTreeSnapshot= setting, Snapshot= applying to the default tools tree, or a documentation note that mkosi.tools.conf is the intended way to pin it. Happy to send a PR for whichever direction you prefer.

Longer writeup with full CI logs: https://github.com/confidential-dot-ai/confidential-os-builder/issues/96

### Used mkosi config

```ini
[Distribution]
Distribution=ubuntu
Release=resolute
Snapshot=20260405T000000Z

[Build]
ToolsTree=default
ToolsTreeDistribution=ubuntu
ToolsTreeRelease=resolute
```

### mkosi output

```sh
‣ Syncing package manager metadata
Get:1 http://security.ubuntu.com/ubuntu resolute-security InRelease [137 kB]
Get:2 http://archive.ubuntu.com/ubuntu resolute InRelease [136 kB]
Get:3 http://security.ubuntu.com/ubuntu resolute-security/universe Sources [45.8 kB]
...
‣ Building tools image
# the later image phase fetches from snapshot.ubuntu.com as expected;
# only the tools tree ignores the snapshot
```

joaosa:
Fixes #4420.

`Snapshot=` has universal scope, so it reaches subimages but not the default tools tree, and there was no `ToolsTree*` counterpart to hand the value across. Replacing `ToolsTreeMirror=` with `Snapshot=` therefore silently dropped the tools tree back to live mirrors.

This adds `ToolsTreeSnapshot=`, modelled directly on `ToolsTreeMirror=`: a tools-scope setting whose default factory takes the image's snapshot when the tools tree distribution matches the image's, so the common same-distribution case is pinned without any extra configuration. Docs get the new setting, a note under `Snapshot=` about the tools tree, and the entry in the main-image-only settings list.

Ran `pytest tests/test_config.py tests/test_json.py` (74 passed). No serialization fixtures change since tools-scope settings aren't part of the image Config.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -3875,6 +3875,21 @@ SETTINGS: list[ConfigSetting[Any]] = [
         scope=SettingScope.tools,
     ),
     ConfigSetting(
+        dest="tools_tree_snapshot",
+        metavar="SNAPSHOT",
+        section="Build",
+        default_factory_depends=("distribution", "snapshot", "tools_tree_distribution"),
+        default_factory=(
+            lambda ns: (
+                ns["snapshot"]
+                if ns["snapshot"] and ns["distribution"] == ns["tools_tree_distribution"]
+                else None
+            )
+        ),
+        help="Set the distribution snapshot to use for the default tools tree",
+        scope=SettingScope.tools,
+    ),
+    ConfigSetting(
         dest="tools_tree_repositories",
         long="--tools-tree-repository",
         metavar="REPOS",
--- a/mkosi/resources/man/mkosi.1.md
+++ b/mkosi/resources/man/mkosi.1.md
@@ -572,6 +572,9 @@ boolean argument: either `1`, `yes`, or `true` to enable, or `0`, `no`,
 
     For any distribution not listed above, snapshots are not supported.
 
+    Note that this setting does not apply to the default tools tree; use
+    `ToolsTreeSnapshot=` for that.
+
 `LocalMirror=`, `--local-mirror=`
 :   The mirror will be used as a local, plain and direct mirror instead
     of using it as a prefix for the full set of repositories normally supported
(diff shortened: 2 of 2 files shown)


joaosa:
The postmarketos failures here seem to be the grub 2.14-r1 trigger breaking every PR's postmarketos leg. Sent a fix in #4422. The opensuse leg looks like a transient zypper repo error.

joaosa:
I think it would be good to merge https://github.com/systemd/mkosi/pull/4419 to make progress here 🙏🏼

behrmann:
Looks reasonable to me, fixing pmOS to merge this is not necessary, as that's unrelated.

bluca (approved):

PR #4422, opened 2026-08-16
apk: disable grub's update-grub trigger during package installation

joaosa:
All postmarketOS integration legs currently fail on every open PR: Alpine's grub 2.14-r1 ships a /boot trigger that execs update-grub, whose grub-probe cannot find a device for / inside the buildroot and fails the whole apk transaction:

```
/usr/sbin/grub-probe: error: cannot find a device for / (is /dev mounted?).
ERROR: lib/apk/exec/grub-2.14-r1.trigger: exited with error 1
```

The trigger has its own opt-out (`disable_trigger=1` in `/etc/update-grub.conf`), and mkosi manages bootloader configuration itself, so write the opt-out around installation and remove it afterwards unless the image shipped its own — same idea as the policy-rc.d file the apt installer writes on Debian.

Ran ruff format/check and the config unit tests (74 passed).

### DIFF
--- a/mkosi/installer/apk.py
+++ b/mkosi/installer/apk.py
@@ -11,6 +11,7 @@ from mkosi.config import Cacheonly, Config
 from mkosi.context import Context
 from mkosi.installer import PackageManager
 from mkosi.run import CompletedProcess, run, workdir
+from mkosi.sandbox import umask
 from mkosi.tree import rmtree
 from mkosi.util import _FILE, PathString, copyfile, flatten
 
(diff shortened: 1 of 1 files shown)


yuwata:
I think #4419 is better. I think it is not a good idea to create the file every time when a package is installed.

joaosa:
> I think #4419 is better. I think it is not a good idea to create the file every time when a package is installed.

Alright that is a good point

joaosa:
Closing in favor of https://github.com/systemd/mkosi/pull/4419

PR #4430, opened 2026-08-26
action: support container-based self-hosted runners

katexochen:
The action assumes a GitHub-hosted Ubuntu VM: sysfs is writable, udev is running, and apt-get exists. On a self-hosted runner inside a container none of that holds: sysfs is read-only, so udevadm trigger fails writing uevents, and apt-get doesn't exist on non-Debian hosts.

Run the udevadm calls only when /sys is writable, mirroring udevd's own ConditionPathIsReadWrite=/sys. In containers the device nodes are provided by the container manager, and the direct chmod still applies the permissions. Skip the apt-get calls when apt-get is not available, removing apparmor only exists to work around the broken Ubuntu runner images.

On GitHub-hosted runners all conditions hold, so nothing changes there.

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -22,14 +22,18 @@ runs:
         KERNEL=="vhost-vsock", GROUP="kvm", MODE="0666", OPTIONS+="static_node=vhost-vsock"
         KERNEL=="vhost-net", GROUP="kvm", MODE="0666", OPTIONS+="static_node=vhost-net"
         EOF
-        sudo udevadm control --reload-rules
+        if sudo udevadm control --ping &>/dev/null; then
+          sudo udevadm control --reload-rules
+        fi
         # kvm/vhost might not be available (e.g.: s390x, ppc64le)
         sudo modprobe kvm || true
         sudo modprobe vhost_vsock || true
         sudo modprobe vhost_net || true
-        [[ -e /dev/kvm ]] && sudo udevadm trigger --name-match=kvm
-        [[ -e /dev/vhost-vsock ]] && sudo udevadm trigger --name-match=vhost-vsock
-        [[ -e /dev/vhost-net ]] && sudo udevadm trigger --name-match=vhost-net
+        if sudo test -w /sys; then
+          [[ -e /dev/kvm ]] && sudo udevadm trigger --name-match=kvm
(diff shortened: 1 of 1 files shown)


action.yaml
```diff
@@ -22,14 +22,18 @@ runs:
...
+          sudo udevadm control --reload-rules
+        fi
         # kvm/vhost might not be available (e.g.: s390x, ppc64le)
         sudo modprobe kvm || true
         sudo modprobe vhost_vsock || true
         sudo modprobe vhost_net || true
-        [[ -e /dev/kvm ]] && sudo udevadm trigger --name-match=kvm
-        [[ -e /dev/vhost-vsock ]] && sudo udevadm trigger --name-match=vhost-vsock
-        [[ -e /dev/vhost-net ]] && sudo udevadm trigger --name-match=vhost-net
+        if sudo test -w /sys; then
```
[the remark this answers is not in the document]
katexochen: i guess we can assume sudo exists, also out of scope

behrmann, line 32: Between the first version and this one the first check changed from `test -w /sys` to `udevadm control --ping`, I assume to appease the remark about udev not running, but this check did not? The above check could set a variable and that be checked here as well or you could ping again here.

action.yaml
```diff
@@ -22,14 +22,18 @@ runs:
         KERNEL=="vhost-vsock", GROUP="kvm", MODE="0666", OPTIONS+="static_node=vhost-vsock"
         KERNEL=="vhost-net", GROUP="kvm", MODE="0666", OPTIONS+="static_node=vhost-net"
         EOF
-        sudo udevadm control --reload-rules
+        if sudo udevadm control --ping >/dev/null 2>&1; then
```
behrmann: Since this uses `bash`, we can make this slightly easier on the eyes

```suggestion
        if sudo udevadm control --ping &>/dev/null; then
```

behrmann (approved):

action.yaml
katexochen: I think it's right the way it is: `udevadm control` speaks to udevd, never touches sysfs. `udevadm trigger` writes changes to `/sys/.../uevent` but doesn't speak to udevd. 
`udevadm control` will succeed on a system with ro sysfs, and `udevadm trigger` will pass on a system without udevd running (but the triggers won't have effect there).

bluca (approved):

PR #4436, opened 2026-09-01
Allow using SignInitrdPCRs with SplitArtifacts=pcrs

bluca:
Needs https://github.com/systemd/systemd/pull/43609 to work end-to-end

### DIFF
--- a/mkosi/__init__.py
+++ b/mkosi/__init__.py
@@ -1721,6 +1721,7 @@ def build_uki(
         die("Could not find ukify")
 
     json_out = False
+    ukify_version = systemd_tool_version(python_binary(context.config), ukify, sandbox=context.sandbox)
 
     arguments: list[PathString] = [
         "--os-release", f"@{workdir(context.root / 'usr/lib/os-release')}",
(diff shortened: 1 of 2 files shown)


mkosi/__init__.py
```diff
@@ -1776,15 +1777,13 @@ def build_uki(
...
-        ukify_version = systemd_tool_version(python_binary(context.config), ukify, sandbox=context.sandbox)
-
         if ukify_version >= "258":
             cert_parameter = "--pcr-certificate"
         else:
             cert_parameter = "--pcr-public-key"
 
         if context.config.sign_initrd_pcrs == ConfigFeature.enabled or (
-            context.config.sign_initrd_pcrs == ConfigFeature.auto and ukify_version >= "261.999"
+            context.config.sign_initrd_pcrs == ConfigFeature.auto and ukify_version >= "262~"
```
behrmann, line 1786: So this was only changed from `262~devel` to `261.999` in a few weeks back, so it built in OBS. If the version on OBS is already e.g. 262.rc1now, then this could be bumped to
```suggestion
            context.config.sign_initrd_pcrs == ConfigFeature.auto and ukify_version >= "262"
```
since we'll bump it to that after release anyway, otherwise, it would make sense to just keep it the way it is currently, because otherwise the OBS packages won't fulfil this anymore.

bluca: It would break current builds as we have `262~rc1` right now, and `262 > 262~rc1`, so the trailing tilde is correct

behrmann (approved):

behrmann:
CI failures are unrelated and will be fixed by #4440

PR #4441, opened 2026-09-03
Add BuildStream support

abderrahim:
The GNOME OS folks are looking into mkosi to build their images instead
of BuildStream. While BuildStream will still take care of providing the
rootfs tree, mkosi would take over the responsibility of packaging that
directory tree into a disk image.

Let's add support for BuildStream to mkosi to make this possible. Unlike
the other supported distributions, BuildStream is not intended to be consumed
by installing individual packages. Instead, BuildStream elements should be
exposed which provide the full rootfs that should go into the image. That's
why we limit the number of packages that can be specified to a single one, which
should always provide all contents that should go into the image.

### DIFF
--- a/mkosi/distribution/buildstream.py
+++ b/mkosi/distribution/buildstream.py
@@ -1,0 +1,60 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+from mkosi.config import Architecture, Config
+from mkosi.context import Context
+from mkosi.distribution import (
+    Distribution,
+    DistributionInstaller,
+    PackageType,
+)
+from mkosi.installer.bst import BST
+from mkosi.log import die
+
+
+class Installer(DistributionInstaller, distribution=Distribution.buildstream):
+    @classmethod
+    def pretty_name(cls) -> str:
+        return "BuildStream"
+
+    @classmethod
+    def filesystem(cls) -> str:
+        return "btrfs"
+
+    @classmethod
+    def package_type(cls) -> PackageType:
+        return PackageType.none
+
+    @classmethod
+    def default_release(cls) -> str:
+        return "snapshot"
+
+    @classmethod
+    def package_manager(cls, config: "Config") -> type[BST]:
+        return BST
+
+    @classmethod
+    def setup(cls, context: Context) -> None:
+        pass
+
+    @classmethod
+    def install(cls, context: Context) -> None:
+        pass
+
+    @classmethod
+    def architecture(cls, arch: Architecture) -> str:
+        a = {
+            Architecture.x86_64: "x86_64",
+        }.get(arch)  # fmt: skip
+
+        if not a:
+            die(f"Architecture {a} is not supported by {cls.pretty_name()}")
+
+        return a
+
+    @classmethod
+    def latest_snapshot(cls, config: Config) -> str:
+        die(f"Latest snapshot not supported by {cls.pretty_name()}")
+
+    @classmethod
+    def is_kernel_package(cls, package: str) -> bool:
+        return False
--- a/mkosi/installer/bst.py
+++ b/mkosi/installer/bst.py
@@ -41,6 +41,8 @@ class BST(PackageManager):
         options: list[PathString] = [
             "--same-dir",
             *context.rootoptions(),
+            # bst will need write access to the current directory as it might write to .bst
+            "--bind", Path.cwd(), Path.cwd(),
             # bst might need to lookup files/paths across the user's home directory so make sure it is
             # available.
             "--bind", Path.home(), Path.home(),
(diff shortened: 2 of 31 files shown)


mkosi/installer/bst.py
```diff
@@ -0,0 +1,82 @@
...
+            Path.cwd(),
+            # bst might need to lookup files/paths across the user's home directory so make sure it is
+            # available.
+            "--bind",
+            Path.home(),
+            Path.home(),
+            "--setenv",
+            "HOME",
+            Path.home(),
+        ]
```
behrmann: Style nit:

```suggestion
        options: list[PathString] = [
            "--same-dir",
            *context.rootoptions(),
            # bst will need write access to the current directory as it might write to .bst
            "--bind", Path.cwd(), Path.cwd(),
            # bst might need to lookup files/paths across the user's home directory so make sure it is
            # available.
            "--bind", Path.home(), Path.home(),
            "--setenv", "HOME", Path.home(),
        ]  # fmt: skip
```

behrmann (commented):
Coming along nicely, thank you!

mkosi/installer/bst.py
abderrahim: It's what Daan had initially, but ruff decided to reformat it. I'll add the "fmt: skip" as suggested.

Skyb0rg007:
I'm trying this out, and ran into an issue where the current shell's umask was influencing the output OS image's file permissions. I think that `bst.py` needs to set the umask before calling `bst artifact checkout`.

Edit: #4443 might solve this; not sure

bluca:
Does the fact that a bunch of existing config files are getting `Distribution=!buildstream` added mean that every user of mkosi will also have to add the same boilerplate everywhere to their existing configs? Because if so I don't think that's a good solution, and it really should not be necessary to change existing configs just because a new backend is added

abderrahim:
> Does the fact that a bunch of existing config files are getting Distribution=!buildstream added mean that every user of mkosi will also have to add the same boilerplate everywhere to their existing configs? Because if so I don't think that's a good solution, and it really should not be necessary to change existing configs just because a new backend is added

In general, no. The files here are used by every user of mkosi. So essentially, the `Distribution=!buildstream` added there means that default package names don't get in the way if someone wants to use a buildstream distribution for their initrd or tools tree (but at the same time, they need to bring their own package for the initrd or tools tree).

So the only reason I can see of why a project author that uses mkosi would want to have something like this is if they want to support using buildstream as a tools tree alongside a different distribution, but also provide additional packages for their tools tree. A pretty niche use case if you ask me.

mkosi/distribution/buildstream.py
```diff
@@ -0,0 +1,60 @@
...
+        pass
+
+    @classmethod
+    def install(cls, context: Context) -> None:
+        pass
+
+    @classmethod
+    def architecture(cls, arch: Architecture) -> str:
+        a = {
+            Architecture.x86_64: "x86_64",
```
Skyb0rg007, line 46: Buildstream supports more than just x86_64; should at least list aarch64 too.

mkosi/installer/bst.py
```diff
@@ -0,0 +1,76 @@
...
+            # available.
+            "--bind", Path.home(), Path.home(),
+            "--setenv", "HOME", Path.home(),
+        ]  # fmt:skip
+
+        # We don't really want to run bst as (fake) root but it uses bubblewrap which stubbornly refuses to
+        # run when invoked unprivileged but with capabilities. We get around this by running as fake root but
+        # still setting $HOME to the user's home to reuse the buildstream cache directory.
+        run(
+            ["bst", "build", *packages],
```
Skyb0rg007, line 56: You should pass `--option target_arch <arch>` here, so that cross-compilation works.

mkosi/distribution/buildstream.py
abderrahim: I'm trying not to make large changes here. I've mostly taken Daan's https://github.com/systemd/mkosi/pull/4149 and fixed obvious problems.

The issue with architecture is more complex, and I'll reply to your other comment.

mkosi/installer/bst.py
abderrahim: Nope, that wouldn't work :-)

The issue is that architecture is a project option, and not a buildstream option. The convention used in buildstream's documentation (and in GNOME OS) is to use an option named `arch`.

Freedesktop-sdk uses two different architecture options since it supports cross-compiling the bootstrap, so you have `bootstrap_build_arch` and `target_arch`.

This isn't something that can easily be fixed here, so I'd rather leave that to a follow up.

Skyb0rg007: Oof. Alright, probably worth waiting for a follow-up then.

mkosi/distribution/buildstream.py
Skyb0rg007: I don't think listing BuildStream’s supported architectures here requires any cross-compilation complexities, but if there’s a planned follow-up I guess it’s also fine to wait.

PR #4443, opened 2026-09-03
sandbox: reset umask to 0022 before exec

jsrkamal2002:
A restrictive host umask (e.g. 0027) would otherwise leak into all sandboxed processes — package managers, build scripts — causing files installed into the rootfs to have wrong permissions (e.g. directories ending up 0750 instead of 0755).

Reset to 0022 at the end of enter(), after the sandbox is fully set up and right before the sandboxed command is exec'd. Both execution paths (preexec fork and standalone sandbox.py prefix) go through enter(), so one change covers both.

### DIFF
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -2245,6 +2245,7 @@ class Config:
     proxy_client_key: Optional[Path]
     make_scripts_executable: bool
     foreign_uid_range: bool
+    umask: Optional[int]
 
     nspawn_settings: Optional[Path]
     ephemeral: bool
@@ -2630,6 +2631,9 @@ class Config:
         options: Sequence[PathString] = (),
     ) -> AbstractContextManager[list[PathString]]:
         opt: list[PathString] = [*options]
+
+        if self.umask is not None:
+            opt += ["--umask", str(self.umask)]
 
         if not relaxed:
             opt += flatten(("--ro-bind", d, d) for d in self.extra_search_paths)
@@ -4193,6 +4197,15 @@ SETTINGS: list[ConfigSetting[Any]] = [
         parse=config_parse_boolean,
         help="Use the foreign UID range",
         scope=SettingScope.main,
+    ),
+    ConfigSetting(
+        dest="umask",
+        name="UMask",
+        metavar="MASK",
+        section="Build",
+        parse=config_parse_mode,
+        help="Set umask for processes running in the sandbox",
+        scope=SettingScope.multiversal,
     ),
     # Runtime section
     ConfigSetting(
@@ -6037,6 +6050,7 @@ def summary(config: Config) -> str:
                    Proxy Client Key: {none_to_none(config.proxy_client_key)}
 
     Automatically set +x on scripts: {yes_no(config.make_scripts_executable)}
+                      Sandbox UMask: {format_octal_or_default(config.umask)}
 
     {bold("HOST CONFIGURATION")}:
                     NSpawn Settings: {none_to_none(config.nspawn_settings)}
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -2633,7 +2633,7 @@ class Config:
         opt: list[PathString] = [*options]
 
         if self.umask is not None:
-            opt += ["--umask", str(self.umask)]
+            opt += ["--umask", f"{self.umask:o}")]
 
         if not relaxed:
             opt += flatten(("--ro-bind", d, d) for d in self.extra_search_paths)
--- a/mkosi/config.py
+++ b/mkosi/config.py
@@ -2633,7 +2633,7 @@ class Config:
         opt: list[PathString] = [*options]
 
         if self.umask is not None:
-            opt += ["--umask", f"{self.umask:o}")]
+            opt += ["--umask", f"{self.umask:o}"]
 
         if not relaxed:
             opt += flatten(("--ro-bind", d, d) for d in self.extra_search_paths)
--- a/mkosi/sandbox.py
+++ b/mkosi/sandbox.py
@@ -1712,6 +1712,10 @@ def enter(argv: list[str]) -> list[str]:
             os.environ["LISTEN_FDS"] = str(nfds)
             os.environ["LISTEN_PID"] = str(os.getpid())
 
+    # Reset umask so a restrictive host umask (e.g. 0027) doesn't leak into sandboxed processes
+    # and cause package managers/scripts to create files in the rootfs with wrong permissions.
+    os.umask(0o022)
+
     return argv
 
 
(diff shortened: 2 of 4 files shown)


behrmann:
What problem specifically does this address? If a a package manager doesn't properly set the permissions of a file that seems more an issue with the package manager than the umask?

jsrkamal2002:
I think the build should produce the same result regardless of the host's `umask`.

User-provided build scripts are probably the clearest example here. If someone writes a `BuildScript=` with a few `mkdir()` calls, they shouldn't have to care whether CI happens to be running with a hardened `umask` or not. Otherwise, the exact same build script can behave differently depending on the host configuration.

I actually ran into this on a hardened machine where the `umask` is set to `027`. For example, `/etc/profile.d/` and the scripts under it ended up with permissions that were too restrictive (`750` instead of the expected `755`/`644`), which meant other users inside the resulting rootfs couldn't properly use the profile scripts.

That's why I think resetting the `umask` when entering the sandbox makes sense. It keeps the build environment predictable and avoids having the host's security settings accidentally leak into the permissions of files created by the build.

behrmann:
While that is true, this cuts both ways. If anybody relied on the umask of their build environment for the permissions of the file they generated (maybe without them realizing), changing this to a less strict value will result in files having more open permissions than previously.

While it's generally wrong to use implicit assumptions for this I'd personally be more inclined towards *not* weakening permissions, since too strict is still better than the opposite. Therefore I don't think this is a good solution.

Better solutions would be a setting `ScriptUmask=` to be applied to user run scripts, that defaults to the current umask or just a documentation fix highlighting that people should set the umask.

jsrkamal2002:
That's a fair point. Resetting to a less restrictive `umask` could potentially open up permissions
that someone was implicitly relying on, and I agree that, as a default,
**"too strict is safer than too open."**

One thing worth clarifying on the `/etc/profile.d/` case: it's not purely a package manager bug —
the mechanism is more subtle.

On the affected machine:

- `/etc/profile` and `/etc/bash.bashrc` both set `umask 027` system-wide
- My personal `~/.bashrc` overrides that back to `0022`, so my interactive shell looks fine
- But mkosi's sandboxed processes are **non-interactive** — they don't source `~/.bashrc` — so
  they inherit `0027` from the system defaults

That's how the package manager ends up creating files with too-restrictive permissions: it sees a
stricter umask than expected, without either of us realizing it.

Since the issue affects more than just user build scripts — it hits **everything the sandbox runs**,
including package managers — I'd suggest considering a broader `Umask=` setting (or
`SandboxUmask=`) that applies to the entire sandbox, not just user scripts. It would:

- Default to the **current host umask** (no behavior change by default)
- Let users on hardened machines explicitly set it to `0022` to get predictable results

`ScriptUmask=` scoped to user scripts would also help, but wouldn't cover the package manager case.

Happy to implement either direction — a global sandbox umask setting, a script-scoped one, or just
a documentation note. Let me know what you'd prefer.

behrmann:
I stand by package manager not setting permissions to what they expect and relying on umask to be broken. Also, relying for what your bashrc sets to be honoured by some script to be borderline. The flowchart for what shell rc gets loaded by bash under what circumstances is nightmarish. 

That being said, a general `UMask=` setting seems okay to me. It's something one could set, but if this is so important for some project, I think it's fine to have it documented on config.

jsrkamal2002:
Agreed on both counts. I'll implement a `UMask=` setting that defaults to the current host umask (so no behavior change unless explicitly set) and document it in the config reference. I'll send a revised patch.

jsrkamal2002:
Replaces the earlier unconditional umask reset with a configurable `UMask=` setting (suggested by the maintainer). When set, the specified umask is applied to all sandbox processes just before `exec`. Unset by default, so existing behavior is unchanged.

Example:

```ini
[Build]
UMask=0022
```

jsrkamal2002:
The **(fedora, fedora, ubuntu-24.04-ppc64le)** integration tests are currently failing due to dependency issues.

mkosi/sandbox.py
```diff
@@ -1552,6 +1554,8 @@ def enter(argv: list[str]) -> list[str]:
             suppress_chown = True
         elif arg == "--suppress-sync":
             suppress_sync = True
+        elif arg == "--umask":
+            umask_value = int(argv.pop())
```
behrmann:
```suggestion
            umask_value = int(argv.pop(), 8)
```

mkosi/config.py
```diff
@@ -2631,6 +2632,9 @@ def sandbox(
     ) -> AbstractContextManager[list[PathString]]:
         opt: list[PathString] = [*options]
 
+        if self.umask is not None:
+            opt += ["--umask", str(self.umask)]
```
```suggestion
            opt += ["--umask", f"{self.umask:o}")]
```

behrmann (commented):
Maybe I misunderstood, what we were talking about, but why is it not enough to to set the umask once after config parsing and depend on umask being passed to child processes?

jsrkamal2002:
The `sandbox()` method is called per-image (it's on the Config object, not globally). mkosi supports building multiple images in a single invocation, each with its own config and potentially a different `UMask=` value. Passing `--umask` as a sandbox argument is the right granularity — setting the umask once globally after config parsing would only work for single-image builds and would hard-code whichever image's config you read first.

jsrkamal2002:
Hi @behrmann 👋

Just wanted to give you a quick update — all 30 CI checks are now passing on the latest commit (`305b5ea`)

The implementation is ready for review:

* `UMask=` is opt-in via the `[Build]` section / `--umask=` CLI flag
* The host's `umask` is inherited by default, so there’s no behavior change unless explicitly configured
* Per-sandbox granularity is maintained for multi-image builds

Would you mind taking one final look and approving if everything looks good? Happy to make any further adjustments if needed.

Thanks for your time!

jsrkamal2002:
Hey, just wanted to flag that the 6 failing integration-test (`fedora, *, ubuntu-24.04`) jobs aren't related to this PR's changes.

All checks were green on my last commit (`305b5ea`). The failures showed up after I merged the latest `main` into the branch (`33dee78`), which brought in `568b1bce` (`dnf: Only read repositories and variables from the sandbox tree`).

It looks like the combination of `--use-host-config` and the newly added `--setopt=reposdir=` flags for `dnf5` is causing Fedora image rebuilds to fail during integration tests on x86-64/KVM.

The `UMask=` changes here don't touch any of that — `umask` defaults to `None`, so there is no change in behaviour unless it is explicitly configured.

jsrkamal2002:
Hi @behrmann 👋
Could you please take a look at this when you get a chance? Thanks! 😊

PR #4446, opened 2026-09-06
Rename sysupdate configuration files to .transfer

johnp789:
Updated configuration file names in docs to use '.transfer' extension. There is a message: `As of v257, transfer definitions should have the '.transfer' extension.`


behrmann (approved):
Thank you!

PR #4455, opened 2026-09-11
run: pass through TERMINFO and TERMINFO_DIRS environment variables

### LINKED ISSUE #4448: `run.py` should pass through `TERMINFO` and `TERMINFO_DIRS` environment variables

main

### Used host distribution

NixOS 26.11

### Used target distribution

freedesktop-sdk (using #4441)

### Linux kernel version used

7.1.12

x86_64

I use kitty as my terminal program, which sets `TERM=xterm-kitty`. When running `mkosi vm`, I was shown `tput` errors when the VM exited: `unknown terminal "xterm-kitty"`.

This can be fixed by simply adding `TERMINFO` and `TERMINFO_DIRS` to the list of saved environment variables here:

https://github.com/systemd/mkosi/blob/938d6a1c21b7e6bb6e716aa7d66b6865aa6a6798/mkosi/run.py#L337

### Used mkosi config

```ini
# This isn't super relevant, but here goes
[Distribution]
Distribution=buildstream

[Content]
Packages=myos/filesystem.bst
Bootable=yes
Bootloader=uki
Initrds=build/mkosi/initrd.cpio.zst

[Output]
Format=disk
SplitArtifacts=uki
OutputDirectory=build/mkosi

[Runtime]
RAM=2G
Console=interactive
```

### mkosi output

```sh
...
tput: unknown terminal "xterm-kitty"
tput: unknown terminal "xterm-kitty"
```

Visio-Vanitas:
Fixes #4448

## Summary
Pass through `TERMINFO` and `TERMINFO_DIRS` environment variables in `mkosi/run.py` subprocess spawning alongside `SYSTEMD_LOG_LEVEL`, `SYSTEMD_LOG_LOCATION`, and `LD_LIBRARY_PATH`.

## Motivation
When using terminal emulators such as `xterm-kitty` that rely on non-standard terminfo definitions, commands and VMs spawned via `run()` would fail terminal capabilities detection (`tput: unknown terminal "xterm-kitty"`) when exiting because `TERMINFO` / `TERMINFO_DIRS` were dropped from the sanitized subprocess environment.

## Changes
- In `mkosi/run.py`, added `"TERMINFO"` and `"TERMINFO_DIRS"` to the list of environment variables preserved from `os.environ`.
- Added unit test in `tests/test_run.py` asserting that `TERMINFO` and `TERMINFO_DIRS` are properly passed through by `run()`.

## Validation
- Verified with unit test regression in `tests/test_run.py`.

### DIFF
--- a/mkosi/run.py
+++ b/mkosi/run.py
@@ -334,7 +334,7 @@ def spawn(
     if "TMPDIR" in os.environ:
         env["TMPDIR"] = os.environ["TMPDIR"]
 
-    for e in ("SYSTEMD_LOG_LEVEL", "SYSTEMD_LOG_LOCATION", "LD_LIBRARY_PATH"):
+    for e in ("SYSTEMD_LOG_LEVEL", "SYSTEMD_LOG_LOCATION", "LD_LIBRARY_PATH", "TERMINFO", "TERMINFO_DIRS"):
         if e in os.environ:
             env[e] = os.environ[e]
 
--- a/tests/test_run.py
+++ b/tests/test_run.py
@@ -3,11 +3,12 @@
 import contextlib
 import os
 import subprocess
+import sys
 from pathlib import Path
 
 import pytest
 
-from mkosi.run import fork_and_wait
+from mkosi.run import fork_and_wait, run
 from mkosi.sandbox import EPERM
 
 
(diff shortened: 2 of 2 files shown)


Skyb0rg007:
Another option is to only pass through the environment variables to the `mkosi vm` subcommand, since the error arises there due the fact that the subcommand calls `tput`; it could still be worth unsetting those variables in most other cases.

Visio-Vanitas:
Thanks for the review and suggestion.

Looking closer at the execution path, `tput` is actually invoked from the `finally:` block in `mkosi/__main__.py` to restore cursor visibility (`cnorm`) and line wrapping (`smam`) on exit. The error surfaced after running interactive commands like `mkosi vm` because `run()` sanitized `TERMINFO` and `TERMINFO_DIRS` before executing those exit cleanup calls.

Rather than whitelisting these variables globally in `run.py` or passing them through `mkosi vm`, we narrowed the passthrough strictly to the `tput` invocations in `mkosi/__main__.py`. This fixes the `tput: unknown terminal "xterm-kitty"` failure on exit while ensuring terminal environment variables do not leak into build sandboxes or other subcommands.

PR #4462, opened 2026-09-16
Fix rootfull builds

### LINKED ISSUE #4233: `mkosi` runs into permission errors even when running as root

### LINKED ISSUE #4452: Running with sudo/run0 fails at permissions check of configure script

v27 and main

### Used host distribution

Arch

### Used target distribution

Arch

### Linux kernel version used

7.2.3-zen1-3-zen

x86_64

When running `mkosi` with `sudo` or `run0`, a `mkosi.configure` script below a directory with permissions that don't allow other users to descend into that directory will appear as not executable and mkosi will fail.

A git bisect marks 7a1754109b571657fdeeb6795eac5a55c0426d9a as the culprit commit.

I guess this stems from the checks now also running inside a sandbox, so the actual problem is probably that the restrictive directory doesn't have `o+x`, mkosi can't open `mkosi.configure` at all and fails with an ambiguous error message. Changing the directory permissions to `0701` from `0700` solves this issue, so I'm fine if this is intended/acceptable behaviour, hopefully if someone has the same problem, they'll find this issue then.

But I think at least the error message could be a bit more on the point.

### Used mkosi config

```ini
# create mkosi.configure with executable bit inside a directory with restrictive permissions
cd "$(mktemp -d)"
chmod 0700 .
touch mkosi.configure
chmod +x mkosi.configure
run0 mkosi
# This is of course not a sensible example in general, but it's the easiest for reproducing the error
```

### mkosi output

```sh
‣ Not building default initrd for image 'main': Bootable=auto and no kernel package found in Packages= or VolatilePackages= for distribution arch
‣ Not building default initrd for image 'main': Bootable=auto and no kernel package found in Packages= or VolatilePackages= for distribution arch
‣ /tmp/tmp.wXDqPmEIJQ/mkosi.configure is not executable
‣ + tput cnorm
‣ + tput smam
  File "/usr/lib/python3.14/site-packages/mkosi/run.py", line 104, in uncaught_exception_handler
    yield
  File "/usr/lib/python3.14/contextlib.py", line 85, in inner
    return func(*args, **kwds)
  File "/usr/lib/python3.14/site-packages/mkosi/__main__.py", line 44, in main
    ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/mkosi/__init__.py", line 5278, in run_verb
    images[i] = config = run_configure_scripts(config)
                         ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/usr/lib/python3.14/site-packages/mkosi/__init__.py", line 668, in run_configure_scripts
    check_script(config, script)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/mkosi/__init__.py", line 660, in check_script
    die(f"{script} is not executable")
    ~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/mkosi/log.py", line 35, in die
    sys.exit(1)
    ~~~~~~~~^^^
SystemExit: 1
```

behrmann:
This is a followup to https://github.com/systemd/mkosi/commit/7a1754109b571657fdeeb6795eac5a55c0426d9a and https://github.com/systemd/mkosi/pull/4444 and adds a few knobs.

The default remains unchanged when mkosi is run as an unprivileged user and we try to acquire a delegation from nsresourced. When run as root, the default number of delegated subranges is set to zero and the foreign UID range is only used when explicitly requested, which leads to an early exit in `acquire_privileges` which is akin to the early exit pre 7a1754109b571657fdeeb6795eac5a55c0426d9a. To have root request delegated subranges as well, a new option `DelegateRanges=` is added.

The opt out for nsresourced via the `MKOSI_FORCE_USERNS_FALLBACK"` is reworked so that it is equivalent to getting a connection refused from nsresourced and trigger the fallback path in the same manner.

There is some more assorted fixlets and cosmetic changes around, `ForeignUIDRange=` is added to the summary, an unused variable is dropped, and an argument for `acquire_privileges` is renamed, so I can see it's an int instead of a bool without looking at the function signature.

Fixes: #4452 
Fixes: #4233
Closes: #4268


bluca (approved):

Cornelicorn:
Can confirm this fixes #4452 for me and my whole workflow (building AUR packages) also then runs through without further problems when on this branch.

daandemeyer (approved):

PR #4465, opened 2026-09-21
Feature/add oci output format

Andre-85:
Hi @all,

this patch adds oci-archive as output format to mkosi. These images can be directly imported to podman with cmdline `podman load` and especially  imported with podman-py to podman. Should be also useful for tools which operate with an REST API, since these tools normally are not supporting the simple folder layout produced by the oci output format of mkosi.

I tested on Fedora Workstation 44 with podman 5.8.1 and podman-py 5.6.0-4.fc44.noarch.


Andre-85:
my tests where:
```
mkosi --format=oci-archive build
=> produces image.tar
podman load -i image.tar
podman images
=> shows the imported image
```
and
```
mkosi --format=oci-archive build

import podman
with podman.PodmanClient() as client:
    client.images.load(file_path="image.tar")

podman images
=> shows the imported image
```

mkosi/__init__.py
```diff
@@ -4183,13 +4183,19 @@ def build_image(context: Context) -> None:
 
     if context.config.output_format == OutputFormat.tar:
         make_tar(context.root, context.staging / context.config.output_with_format, sandbox=context.sandbox)
-    elif context.config.output_format == OutputFormat.oci:
+    elif context.config.output_format in (OutputFormat.oci, OutputFormat.oci_archive):
```
behrmann: Please add an `is_oci` method to the `OutputFormat` enum.

mkosi/__init__.py
```diff
@@ -4183,13 +4183,19 @@ def build_image(context: Context) -> None:
...
         make_tar(context.root, context.staging / "rootfs.layer", sandbox=context.sandbox)
-        make_oci(
-            context,
-            context.staging / "rootfs.layer",
-            context.staging / context.config.output_with_format,
+        oci = (
+            context.staging / context.config.output_with_format
+            if context.config.output_format == OutputFormat.oci
+            else context.workspace / "oci"
```
behrmann, line 4191: put this in `context.staging` as well.

mkosi/__init__.py
```diff
@@ -4183,13 +4183,19 @@ def build_image(context: Context) -> None:
...
         )
+        make_oci(context, context.staging / "rootfs.layer", oci)
+
+        if context.config.output_format == OutputFormat.oci_archive:
+            make_tar(oci, context.staging /
+                     context.config.output_with_format, sandbox=context.sandbox)
+            rmtree(oci, sandbox=context.sandbox)
```
behrmann: The cleanup is unnecessary, it's not like we clean up the `rootfs.layer`.

mkosi/__init__.py
```diff
@@ -4183,13 +4183,19 @@ def build_image(context: Context) -> None:
...
+        oci = (
+            context.staging / context.config.output_with_format
+            if context.config.output_format == OutputFormat.oci
+            else context.workspace / "oci"
         )
+        make_oci(context, context.staging / "rootfs.layer", oci)
+
+        if context.config.output_format == OutputFormat.oci_archive:
+            make_tar(oci, context.staging /
+                     context.config.output_with_format, sandbox=context.sandbox)
```
```suggestion
            make_tar(
                oci, 
                context.staging / context.config.output_with_format, 
                sandbox=context.sandbox,
            )
```

mkosi/__init__.py
Andre-85: This is unclear to be. Here i need to check if i create the uncompressed or the tar compressed variant of the oci format. But with is_oci() i cannot differenciate that, since it returns for both formats (oci and oci-archive) true.

mkosi/__init__.py
Andre-85: I removed it

mkosi/__init__.py
Andre-85: Fixed that.

mkosi/config.py
```diff
@@ -1009,7 +1014,7 @@ def config_default_compression(namespace: dict[str, Any]) -> Compression:
         OutputFormat.addon,
     ):
         return Compression.zstd
-    elif namespace["output_format"] == OutputFormat.oci:
+    elif namespace["output_format"] in (OutputFormat.oci, OutputFormat.oci_archive):
```
behrmann: You can now use `.is_oci` here.

